PR #8558, opened 2017-04-10
Update autotuned_index.h

grundman:
Add missing sstream header for decl of std::ostringstream in line 104


sovrasov:
@grundman do you really have a build error without `sstream`? This header has been already included in 
pre-compiled header [precomp.hpp](https://github.com/opencv/opencv/blob/master/modules/flann/src/precomp.hpp) for flan module.

alalek:
This is a public header, so it can't rely on internal `precomp.hpp`

sovrasov:
:+1:

PR #8559, opened 2017-04-11
Initialize the empty InputOutputArray parameter in findTransformECC

liquidmetal:
Currently, `findTransformECC` expects (at least) three parameters. The third parameter is the warp matrix (2x3 or 3x3 depending on the type of warp specified) that aligns the other two parameters.

If this parameter is provided, it is used for initializing the optimization. Otherwise, the user currently needs to [initialize it to identity themselves](https://github.com/opencv/opencv/blob/master/samples/cpp/image_alignment.cpp#L286-L290). This can be seen in the `image_alignment.cpp` sample.

This PR changes this behavior. If the function `findTransformECC` notices that the `InputOutputArray` is empty, it initializes it to the correct sized identity matrix.

### DIFF
--- a/modules/video/include/opencv2/video/tracking.hpp
+++ b/modules/video/include/opencv2/video/tracking.hpp
@@ -297,7 +297,7 @@ row is ignored.
 Unlike findHomography and estimateRigidTransform, the function findTransformECC implements an
 area-based alignment that builds on intensity similarities. In essence, the function updates the
 initial transformation that roughly aligns the images. If this information is missing, the identity
-warp (unity matrix) should be given as input. Note that if images undergo strong
+warp (unity matrix) is used as an initialization. Note that if images undergo strong
 displacements/rotations, an initial transformation that roughly aligns the images is necessary
 (e.g., a simple euclidean/similarity transform that allows for the images showing the same image
 content approximately). Use inverse warping in the second image to take an image close to the first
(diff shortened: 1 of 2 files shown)


sovrasov:
@liquidmetal could you also add a note about new behavior in the documentation?

StevenPuttemans:
Good job on your first PR @liquidmetal !

liquidmetal:
@sovrasov I'm not sure how - could you point me to the file that generates documentation for this function?

sovrasov:
Documentation is [here](https://github.com/opencv/opencv/blob/c17afe0fab61bb11452b36fb94eaeb405293e086/modules/video/include/opencv2/video/tracking.hpp#L311), before the prototype.

sovrasov:
:+1:

liquidmetal:
The [error log](http://pullrequest.opencv.org/buildbot/builders/precommit_opencl_linux/builds/10594) suggests that the build break isn't related to this PR (it is related to the superres module).

StevenPuttemans:
> The error log suggests that the build break isn't related to this PR (it is related to the superres module).

Yes and as you can see, builds have all passed now, since buildbots have been restarted. Now it is just waiting for some admin to assign the PR to him and give his go, then buildbot will pick it up and merge it!

PR #8563, opened 2017-04-11
Fix to build for Android with cmake-toolchains

Slonegg:
Latest CMake has Android toolchain, so special toolchain file is not required.
https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-android-with-the-ndk

But PR fixes build when using integrated cmake toolchain.


alalek:
Inserted code in CMakeLists.txt doesn't work, because native CMake compiler detection is invoked under `project()` command. So used variables (like `CMAKE_ANDROID_ARCH_ABI`) are not initialized.

CMake provides support for NDK only (detection of C/C++ compilers and tools).
OpenCV uses Android SDK and support for Android SDK should be provided/ported too.

Slonegg:
Thanks @alalek, didn't notice that project command is down below. OpenCV builds just fine with just NDK and with SDK without Java. On Windows I did the following:
```
    cmake .. -G "Ninja" -DCMAKE_SYSTEM_NAME=Android -DCMAKE_SYSTEM_VERSION=21 -DCMAKE_ANDROID_ARCH_ABI=arm64-v8a -DCMAKE_ANDROID_NDK=/c/Android/sdk/ndk-bundle/ -DBUILD_opencv_java=OFF
    ninja all
```
[cmake_log.txt](https://github.com/opencv/opencv/files/917587/cmake_log.txt)
[build_log.txt](https://github.com/opencv/opencv/files/917585/build_log.txt)

Unfortunately OpenCV doesn't build with java with latest version of SDK and it seems unrelated to this PR. I did what I could to fix it, but my Android toolchain knowledge is not enough for that.
[build_log_with_java.txt](https://github.com/opencv/opencv/files/917586/build_log_with_java.txt)

Slonegg:
@alalek I've tested changes with older Android SDK, works fine, removed changes for new Android SDK
[cmake_log.txt](https://github.com/opencv/opencv/files/962039/cmake_log.txt)
[build_log.txt](https://github.com/opencv/opencv/files/962052/build_log.txt)

vpisarev:
@Slonegg, thank you for the contribution and sorry for the looong delay with reply. Looks like some parts of the patch (use of `__ANDROID__` in some macros) has been already integrated. As for the CMake part - we decided not to merge it in the current form. It adds some code into the root CMakeLists.txt instead of the OpenCVDetectAndroidSDK.cmake, it does not support Java bindings. Also, there is dependency of fresh CMake versions that should be in-synch with frequently updated Google toolchains. Besides, recently Google added their own CMake-friendly ways to build libraries and apps; we are going to migrate to it eventually

PR #8564, opened 2017-04-12
Remove stdout log when destructor called.

orisano:
### This pullrequest changes

We tried to execute opencv video capture using Python bindings, we have get “Clean up camera” log. I checked this logs so I faced this commented out log https://github.com/opencv/opencv/blob/05b15943d6a42c99e5f921b7dbaa8323f3c042c6/modules/videoio/src/cap_avfoundation.mm#L255

### Question
Is this message really necessary?
If no, I would like to remove this message in the destructure method …


sovrasov:
I think this line can be completely removed.

orisano:
fix it. removed line.

sovrasov:
:+1:

Issue #8532, opened 2017-04-06
Compile with OpenCV 3.2 VS2015 Win10 Failing on i7-7700HQ


turmansky:
##### System information (version)
- OpenCV => 3.2
- Operating System / Platform => Windows10 64 Bit
- Compiler => Visual Studio 2015 Update 3
Processor => i7-7700HQ

##### Detailed description
After downloading the tip of tree for the OpenCV-master and OPenCV_contrib-master, CMake 3.7..2 was used to setup the solution and then a Release build x64 was attempted.  
Following packages are on the laptop

- Intel Parallel studio Xe 2017 
- Python 3.6.1 
- CUDA 8.0 

Failure is always in opencv_core_AVX while compiling mathfuncs_core.avx.cpp
It reports syntax errors on the CV_CPU_OPTIMIZATION_HAL_NAMESPACE_BEGIN entry on line 76
CMake added for CPU_DISPATCH: SSE4_1;AVX;FP16;AVX2 at the configure step as the HW supports it
Tried removing that and going to SSE4_1 option for CPU_DISPATCH, did not help
Precompiled headers option is off.
WITH_TBB option is off.

The macros are not working correctly it appears for the dispatching
Any ideas as to what could be wrong ?

turmansky:
Removing CV_ENABLE_INTRINSICS solves the build issue
 bug only when using instrinsics . will investigate further but is  anyone has feedback please advise

peters:
I'm facing the same issue, only able to build when disabling intrinsics.

turmansky:
WITH_OPENCL also causes issues
Failure in openCV_core  due to ocl:core at linker step. A whole bunch of unresolved externals

turmansky:
Working CMAKE config
-------------------------
Could NOT find PythonInterp (missing:  PYTHON_EXECUTABLE) (Required is at least version "2.7")
Could NOT find PythonInterp (missing:  PYTHON_EXECUTABLE) (Required is at least version "2.6")
Could NOT find Jasper (missing:  JASPER_LIBRARIES JASPER_INCLUDE_DIR) 
found IPP (ICV version): 9.0.1 [9.0.1]
at: C:/Code/SrcExt/OpenCV/Buildv3.2_x64/3rdparty/ippicv/ippicv_win
CUDA detected: 8.0
CUDA NVCC target flags: -gencode;arch=compute_30,code=sm_30;-gencode;arch=compute_61,code=sm_61;-D_FORCE_INLINES
[...]

PR #8571, opened 2017-04-12
cmake: fix optimization filter

alalek:
Check file name only:
```
Excluding from source files list (optimization is disabled):
C:/Code/SrcExt/OpenCV/Buildv3.2_x64/modules/core/opencl_kernels_core.cpp
```

Related #8532


alalek:
:+1:

PR #8574, opened 2017-04-13
fix creation homogeneous affine matrix when constructing from 4x3 cv::Mat

elmewo:
### This pullrequest changes

Constructing an cv::Affine3 from cv::Mat with dimensionss 4x3 results in the 16th element of the matrix not being 1 but 0. Since this would not be a homogeneous matrix, this behaviour is not desired.
This PR fixes the issue by preventing a return in the case of a 4x3 matrix before the 4th row is initialized with [0,0,0,1].


sovrasov:
:+1:

PR #8580, opened 2017-04-14
Extended set of existing performance test to OpenVX HAL suitable modes

terfendail:
WIP
### This pullrequest changes

Added a few tests that evaluate performance with parameters suitable to process data with OpenVX

depends on extra [PR331](https://github.com/opencv/opencv_extra/pull/331)


vpisarev:
:+1:

PR #8582, opened 2017-04-15
cmake: disallow in-source builds

alalek:
Currently even CMake process modifies source tree (rewrites 3rdparty/ippicv).

There is another approach:
```
set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
set(CMAKE_DISABLE_SOURCE_CHANGES ON)
```

but these variables are not documented in CMake. Also they prevent using of in-source `.cache` directory

BTW, You still need to delete `CMakeFiles` / `CMakeCache.txt` from source tree after in-source build attempt.

Related CMake issues:
- https://gitlab.kitware.com/cmake/cmake/issues/11343
- https://gitlab.kitware.com/cmake/cmake/issues/6672

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,13 @@
 #
 # ----------------------------------------------------------------------------
 
+# Disable in-source builds to prevent source tree corruption.
+if(" ${CMAKE_SOURCE_DIR}" STREQUAL " ${CMAKE_BINARY_DIR}")
+  message(FATAL_ERROR "
+FATAL: In-source builds are not allowed.
+       You should create separate directory for build files.
+")
+endif()
 
 
 include(cmake/OpenCVMinDepVersions.cmake)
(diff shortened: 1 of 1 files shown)


vpisarev:
:+1: very useful patch! I think, our buildbot should also disable "in-source" patches when PR's are made from master branch :)

xanview:
Hmm,

```
$ mkdir build
$ cd build
$ pwd
/data/tmp/opencv/build
$ cmake -D BUILD_opencv_java=OFF -D CMAKE_BUILD_TYPE=RELEASE ..
CMake Error at CMakeLists.txt:11 (message):
  

  FATAL: In-source builds are not allowed.

         You should create separate directory for build files.



-- Configuring incomplete, errors occurred!
CMake Error: Unable to open check cache file for write. /data/tmp/opencv/CMakeFiles/cmake.check_cache
```

Any ideas?

alalek:
```
$ cd /data/tmp/opencv
$ ls -al CMake*
```

You should not have CMakeCache.txt here.

xanview:
Ah, that was it, thank you!

PR #8588, opened 2017-04-16
Set Mac OS X deployment target to 10.9

kylefleming:
### This pullrequest changes

Sets the MACOSX_DEPLOYMENT_TARGET to 10.9. Without it set, the build will choose the current OS version as the deployment target. Libraries should be built with a deployment target of the earliest compatible version. 10.9 is the earliest usable, since that is when OS X introduced libc++.

Tested on OS X 10.12.4 using opencv master (10787c6) with a test project that builds for a deployment target of 10.9.

### DIFF
--- a/platforms/osx/build_framework.py
+++ b/platforms/osx/build_framework.py
@@ -18,6 +18,7 @@ class OSXBuilder(Builder):
     def getBuildCommand(self, archs, target):
         buildcmd = [
             "xcodebuild",
+            "MACOSX_DEPLOYMENT_TARGET=10.9",
             "ARCHS=%s" % archs[0],
             "-sdk", target.lower(),
             "-configuration", "Release",


remia:
I think libc++ is available on 10.8.5 also but you have specify the `-stdlib=libc++` manually.

But the build fails in opencv_core module when setting 10.8 deployment target, currently looking into it.

alalek:
Does setting of environment variable works for you?

```
$ export MACOSX_DEPLOYMENT_TARGET=10.9
$ python platforms/osx/build_framework.py ...
```

kylefleming:
@alalek I think it makes more sense to include it in the build script. The library won't be able to be compiled below the current minimum version (10.9 at the moment) and there's no reason for someone to build the library with a higher deployment target since they could still use the library with an app that has a higher deployment target than the library is compiled for.

For comparison, the iOS build script includes the deployment target [here](https://github.com/opencv/opencv/blob/fe21487f91d4daba4ac55204b26b3a5f21223c34/platforms/ios/build_framework.py#L151)

alalek:
:+1:

PR #8593, opened 2017-04-17
fixing segfaults occuring when launching those unit tests

devnexen:
### This pullrequest changes

Some unit tests segfaults due to empty data.

### DIFF
--- a/modules/shape/test/test_shape.cpp
+++ b/modules/shape/test/test_shape.cpp
@@ -83,6 +83,9 @@ protected:
 
     vector<PointType> convertContourType(const Mat& currentQuery) const
     {
+        if (currentQuery.empty()) {
+            return vector<PointType>();
+        }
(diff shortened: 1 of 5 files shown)


modules/shape/test/test_shape.cpp
```diff
@@ -104,8 +107,8 @@ class ShapeBaseTest : public cvtest::BaseTest
 
         // Uniformly sampling
         random_shuffle(contoursQuery.begin(), contoursQuery.end());
-        int nStart=NP;
         vector<PointType> cont;
+        int nStart=NP;
```
alalek: This change looks unnecessary.

devnexen: it is not indeed ... I was trying an alternative solution and forgot to remake it original state.

alalek:
:+1:

PR #8596, opened 2017-04-17
Don't blow out the stack.  Use a smaller buffer and prevent buffer overruns with snprintf.

nnorwitz:


alalek:
1) remove non-ASCII symbols from source files
2) eliminate unnecessary commits from patch (use "git rebase -i opencv/master" instead of "git merge opencv/master")

sovrasov:
:+1:

PR #8524, opened 2017-04-05
Java Algorithm.save and Javadoc generation

mshabunin:
### This pullrequest changes

- replaces #8472 - fixes Algorithm.save and other some other function calls
- adds ML train/save/load test
- adds javadoc generation during the build (will be created in the `<build>/docs/javadoc` directory)

PR #8598, opened 2017-04-18
java: allow to wrap functions which take cv::Ptr

sovrasov:
This PR enables wrapping of functions with prototypes like `dummy(cv::Ptr<cv::Algorithm>)` in java bindings generator.

### DIFF
--- a/modules/java/generator/gen_java.py
+++ b/modules/java/generator/gen_java.py
@@ -649,6 +649,7 @@ public class $jname {
     protected final long nativeObj;
     protected $jname(long addr) { nativeObj = addr; }
 
+    public long getNativeObjAddr() { return nativeObj; }
 """
 
 T_JAVA_START_MODULE = """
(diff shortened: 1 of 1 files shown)


modules/java/generator/gen_java.py
```diff
@@ -649,6 +649,7 @@ def libVersionBlock():
     protected final long nativeObj;
     protected $jname(long addr) { nativeObj = addr; }
 
+    public long getNativeObj() { return nativeObj; }
```
alalek: OpenCV already uses [getNativeObjAddr()](https://github.com/opencv/opencv/blob/c757d2863efb71035468de49c83a4ec22203cf8e/modules/core/misc/java/src/java/core%2BMat.java#L1094-L1096).

sovrasov: But it's for Mat only.

vpisarev:
:+1:

abratchik:
Did you manage to test this fix? I did some analysis of similar issue [here](http://answers.opencv.org/question/146713/java-wrappers-for-opencv-text-module/). I mean adding a public method to return nativeObj is as good as making nativeObj public ) It is public for Mat and the likes so why not to make it public for Algorithm and all other orphans as well?

The other issue is - the way this pointer is wrapped when passed to a function is also an issue. Please kindly check the link I provided above and let me know if it makes sense to update as suggested.

mshabunin:
@abratchik , this fix has been tested in opencv/opencv_contrib#1115, which adds Java wrappers for `MultiTracker` class, which has methods like `bool add(Ptr<Tracker> tracker, ...)`, where generated `Tracker` and `MultiTracker` classes are in different packages. Although there is no Java test for this functionality, the compilation step has [passed](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_linux64/builds/10342).

>adding a public method to return nativeObj is as good as making nativeObj public

Public `get` method is better, because it does not allow user to modify `nativeObj` field.

>I did some analysis of similar issue here

This problem should've been fixed in #8524

PR #8602, opened 2017-04-18
cmake: update FFmpeg Windows wrapper

alalek:
ffmpeg 3.2.1 -> 3.3
libvpx 1.6.0 -> 1.6.1

https://github.com/opencv/opencv_3rdparty/pull/14


peters:
@alalek Hmm, I might hold one merging this one. This might only be affecting our internal test suite, but video encoding is now failing randomly and `vp9` videos end up corrupted (does not contain any frames at all). Downgrading solves the problem. Any points on what might be causing this?

peters:
Tried downgrading libvpx to `1.6.0` but the problem still persists.

alalek:
@peters OK

Are you able to reproduce problem with OpenCV encoding test [case](https://github.com/opencv/opencv/blob/cea0e94376eff179115ee30fa67c4c4745375e7b/modules/videoio/test/test_ffmpeg.cpp#L99) (add line with FOURCC for VP9 codecs)?

peters:
@alalek Yolo! Old binary 🛴 🛴 Works just fine! 🥇

vpisarev:
:+1:

PR #8609, opened 2017-04-19
Fix pyrlk.cl's wrong output of calcOpticalFlowPyrLK function's output vect…

### LINKED ISSUE #8608: The calcOpticalFlowPyrLK function's output err has calculation error between the cpu and gpu code.
Platform:
CPU: Intel 6th generation i5-6260U
GPU: Intel Iris Graphics 540 (Skylake GT3e)
OS: Linux Ubuntu 16.04 LTS x64
Compiler:
gcc: gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609
OpenCL: intel-opencl-2.0-r4.0.59481
OpenCV:
OpenCV => 3.1.0
WITH_OPENCL true

##### Detailed description
The output vector of errors - err - in calcOpticalFlowPyrLK function has different result between the cpu and gpu code. Although both the cpu and gpu code follow the same algorithm, the implementation details are differnet. This reason cause the big calculation error of function output err between cpu and gpu. As 01org/libxcam project request, they need the result of err in their project. The gpu's invalid result of err cause the wrong result of libxcam project.

Furthermore, reviewing calcOpticalFlowPyrLK ocl test code in /modules/video/test/ocl/test_optflowpyrlk.cpp, there is also no test for the result of err.  

Therefore, the calculation error of the result of the output err between the cpu and gpu is needed to be fixed.
 
##### Steps to reproduce
Just print the result of err from cpu and gpu code can find this problem.

With the patch showed below and running ./build/bin/opencv_test_video --gtest_filter=OCL_Video/PyrLKOpticalFlow*, this problem can also be reproduced.

```.diff
diff --git a/modules/video/test/ocl/test_optflowpyrlk.cpp b/modules/video/test/ocl/test_optflowpyrlk.cpp
index 3c264a5..9387f9c 100644
--- a/modules/video/test/ocl/test_optflowpyrlk.cpp
+++ b/modules/video/test/ocl/test_optflowpyrlk.cpp
@@ -44,7 +44,7 @@
 
 #include "../test_precomp.hpp"
 #include "opencv2/ts/ocl_test.hpp"
-
+#include <iostream>
 
 #ifdef HAVE_OPENCL
 
@@ -104,6 +104,7 @@ OCL_TEST_P(PyrLKOpticalFlow, Mat)
     ASSERT_EQ(cpuStatusCPU.size(), status.size());
 
     size_t mistmatch = 0;
+    size_t errmatch = 0;
     for (size_t i = 0; i < nextPts.size(); ++i)
     {
         if (status[i] != cpuStatusCPU[i])
@@ -122,11 +123,15 @@ OCL_TEST_P(PyrLKOpticalFlow, Mat)
 
             if (!eq || errdiff > 1e-1)
                 ++mistmatch;
+
+            if(std::abs(cpuErr[i] - err[i]) > 0.01) {
+                ++errmatch;
+            }
         }
     }
 
     double bad_ratio = static_cast<double>(mistmatch) / (nextPts.size());
-
+    std::cout << "misnumber: " << errmatch << std::endl;
     ASSERT_LE(bad_ratio, eps);
 }
```

LukeZheZhu:
resolves #8608 

I set the max error of the output err between cpu and gpu as 0.01, which means the test points with std::abs(cpuErr[i] - err[i]) < 0.01 are correct. As the different implementation details between the cpu and gpu code, only 90% test points can pass the test.

Furthermore, the gpu code only support window size within `16*16` to `24*24`. Therefore this patch cannot work correctness when the window size out of this range.

In summary, with max differnece of output err between cpu and gpu as 0.01 and window size in the range of `16*16` and `24*24`, this patch can provide 90% correctness.

### This pullrequest changes
Modified the modules/video/src/opencl/pyrlk.cl to fix the output err
Add relative test in the modules/video/test/ocl/test_optflowpyrlk.cpp.

### DIFF
--- a/modules/video/test/ocl/test_optflowpyrlk.cpp
+++ b/modules/video/test/ocl/test_optflowpyrlk.cpp
@@ -77,6 +77,7 @@ OCL_TEST_P(PyrLKOpticalFlow, Mat)
 {
     static const int npoints = 1000;
     static const float eps = 0.03f;
+    static const float erreps = 0.1f;
 
     cv::Mat frame0 = readImage("optflow/RubberWhale1.png", cv::IMREAD_GRAYSCALE);
     ASSERT_FALSE(frame0.empty());
@@ -104,6 +105,8 @@ OCL_TEST_P(PyrLKOpticalFlow, Mat)
     ASSERT_EQ(cpuStatusCPU.size(), status.size());
 
     size_t mistmatch = 0;
+    size_t errmatch = 0;
+
     for (size_t i = 0; i < nextPts.size(); ++i)
     {
         if (status[i] != cpuStatusCPU[i])
@@ -121,13 +124,22 @@ OCL_TEST_P(PyrLKOpticalFlow, Mat)
             float errdiff = 0.0f;
 
             if (!eq || errdiff > 1e-1)
+            {
                 ++mistmatch;
+                continue;
+            }
+
+            eq = std::abs(cpuErr[i] - err[i]) < 0.01;
+            if(!eq)
+                ++errmatch;
         }
     }
 
     double bad_ratio = static_cast<double>(mistmatch) / (nextPts.size());
+    double err_ratio = static_cast<double>(errmatch) / (nextPts.size());
 
     ASSERT_LE(bad_ratio, eps);
+    ASSERT_LE(err_ratio, erreps);
 }
 
 OCL_INSTANTIATE_TEST_CASE_P(Video, PyrLKOpticalFlow,
--- a/modules/video/src/lkpyramid.cpp
+++ b/modules/video/src/lkpyramid.cpp
@@ -849,6 +849,9 @@ namespace
                 return false;
             if (maxLevel < 0 || winSize.width <= 2 || winSize.height <= 2)
                 return false;
+            if (winSize.width < 16 || winSize.height < 16 ||
+                winSize.width > 24 || winSize.height > 24)
+                return false;
             calcPatchSize();
             if (patch.x <= 0 || patch.x >= 6 || patch.y <= 0 || patch.y >= 6)
                 return false;
--- a/modules/video/src/opencl/pyrlk.cl
+++ b/modules/video/src/opencl/pyrlk.cl
@@ -260,7 +260,7 @@ inline void GetPatch(image2d_t J, float x, float y,
 
 inline void GetError(image2d_t J, const float x, const float y, const float* Pch, float* errval)
 {
-    float diff = read_imagef(J, sampler, (float2)(x,y)).x-*Pch;
+    float diff = (((read_imagef(J, sampler, (float2)(x,y)).x * 16384) + 256) / 512) - (((*Pch * 16384) + 256) /512);
     *errval += fabs(diff);
 }
 
@@ -526,6 +526,6 @@ __kernel void lkSparse(image2d_t I, image2d_t J,
         nextPts[gid] = prevPt;
 
         if (calcErr)
-            err[gid] = smem1[0] / (float)(c_winSize_x * c_winSize_y);
+            err[gid] = smem1[0] / (float)(32 * c_winSize_x * c_winSize_y);
     }
 }


modules/video/test/ocl/test_optflowpyrlk.cpp
```diff
@@ -121,13 +124,24 @@ OCL_TEST_P(PyrLKOpticalFlow, Mat)
...
             if (!eq || errdiff > 1e-1)
+            {
                 ++mistmatch;
+                continue;
+            }
+
+            eq = std::abs(cpuErr[i] - err[i]) < 0.01;
+            if(!eq &&
+               winSize.width >= 16 && winSize.height >=16 &&
+               winSize.width <=24 && winSize.height <=24)
```
alalek: This test is intended to test optical flow implementation (any "optimized" version with passed parameters).
So this `winSize` limitation looks illogical here.
Especially, these `winSize` values are not covered by test parameters [below](https://github.com/opencv/opencv/pull/8609/files#diff-12bd0224b222e84bc79818257f400787R149)

BTW, Please ignore "OCL_" prefix here - it is for legacy reasons here (to compare with OpenCV 2.4 tests). It doesn't force OpenCL code.

LukeZheZhu: Hi alalek, 
I am totally get your point. This test not only works for OpenCL, but also works for the CUDA and any other optimizied version code. Therefore, this modification may badly influence the relevant code using this test.
In modules/video/src/opencl/pyrlk.cl file, however, the kernel can only supports the winSize within 16*16 to 24*24, although its input accepts the winSize outside this range. In this kernel, the input winSize bigger than 24*24 will be refered as 24*24, the input winSize smaller than 16*16 will be refered as 16*16. In the test case 2 and 3, with winSize 25*25, although the kernel can output the correct result of nextPts, the output of err is wrong even with the patch I pulled. This is the reason why I modify this test adding the winSize limitation in it.
So, are there any way to resolve this problem? How about that I add a new test specifically for OpenCL optical flow? Or would you mind that give me any other ideas?

Best,

alalek: > This test not only works for OpenCL, but also works for the CUDA and any other optimized version code.

CUDA has own API.
Available code path for this function is: CPU, CPU optimized with intrinsic (SSE/NEON), OpenCL, OpenVX (disabled due accuracy).

> How about that I add a new test specifically for OpenCL

No, you should not do this. At least you can't force OpenCL code path, especially if there is no OpenCL device or something is not supported (like `ocl::Image2D::isFormatSupported(CV_32F, 1, false)`).
This test is intended to check function output. It is launched on many platforms and we check that result is valid on all of them.

> In this kernel, the input winSize bigger than 24*24 will be referred as 24*24, the input winSize smaller than 16*16 will be referred as 16*16.

If OpenCL kernel doesn't support some `winSize` values (or other parameters)), then "host" code should check for these cases and bailout to other implementations.

LukeZheZhu: So, How about that I add a new OpenCL kernel file in modules/video/src/opencl/ directory? When the host code check winSize within `16*16` to `24*24`, the new kernel will be called. Otherwise, the old kernel - pyrlk.cl - will be used. Is that OK?

alalek: 1)
I mean there is no reason to call some implementation (for example, OpenCL kernel) with unsupported parameters.

You state this:
> In this kernel, the input winSize bigger than 24x24 will be referred as 24x24, the input winSize smaller than 16x16 will be referred as 16x16.

if this is true, then appropriate parameters check should be added.

2)
All OpenCV tests should not rely on concrete implementation.
So this strange condition should be eliminated, because its justification is related to concrete OpenCL implementation and it will not be accepted.

vpisarev:
@LukeZheZhu, I think, @alalek means that the check for window size is better be placed to checkParam() method (line 843) that checks whether algorithm parameters are appropriate for OpenCL implementation. BTW, the lower limit (16 x 16) for the window size is quite constraining. People often use window size like 11x11 or 15x15. So it would be nice if someone fixed the kernel to support smaller window size

LukeZheZhu:
Hi @vpisarev @alalek ,
Thanks for your replies. I have modified my patch and placed the check for window size to the checkParam() function. The build bot report failed while building. I, however, can successfully build the code on my computer. I have no idea about this failed.
BTW, I will try to work further on the pyrlk.cl to see if I can let the kernel support smaller window size, such as `11*11` or `15*15`.

LukeZheZhu:
Hi @vpisarev @alalek ,
Thanks for your replies. I have modified my patch and placed the check for window size to the checkParam() function. The buildbot report failed while building. I, however, can successfully build the code on my computer. I have no idea about this failed.
BTW, I will try to work further on the pyrlk.cl to see if I can let the kernel support smaller window size, such as 11*11 or 15*15.

vpisarev:
thank you! :+1:

PR #7935, opened 2016-12-28
highgui: preserve waitKey -1 return value

alalek:
https://github.com/opencv/opencv/pull/7866#issuecomment-269322846

PR #8610, opened 2017-04-19
Update Drawing_2.cpp

altindagtugba:
waitKey( DELAY ) >= 0  should be lower than zero


alalek:
Why?

For loop draws lines/objects until any key is pressed ( `>= 0` - some key is pressed, `< 0` - timeout).
Did you use OpenCV 3.2? There is bug fix: #7935

altindagtugba:
I used waitKeyEx(DELAY) and it returned -1 instead of 255

PR #8612, opened 2017-04-19
highgui: add ROI selection functionality

sovrasov:


modules/highgui/src/roiSelector.cpp
```diff
@@ -0,0 +1,193 @@
...
+        selectorParams.drawFromCenter = fromCenter;
+
+        // show the image and give feedback to user
+        imshow(windowName, img);
+
+        // copy the data, rectangle should be drawn in the fresh image
+        selectorParams.image = img.clone();
+
+        // select the object
+        setMouseCallback(windowName, mouseHandler, (void *)&selectorParams);
```
alalek: 1) Reliable callback cleanup process is required. Probably with save/restore of previous callback state.
2) Passed `(void *)&selectorParams`, but callback wrapper expects `(void*)this` - add just one "virtual" function or extra field and you would have interesting debug session of crash dumps.
3) `param` should be eliminated in this non-static method:
```
    void opencv_mouse_callback(int event, int x, int y, int, void *param)
    {
        handlerT *data = (handlerT *)param;
```
4. 'ESC' key inconsistency. Some highgui backends may return capital 'C' symbol.
5. 1000 FPS processing is not required (`waitKey(1)`). 20 fps is enough.
6. `& 0xff` is not required in OpenCV 3.2+ .

alalek:
:+1:

PR #8616, opened 2017-04-19
dnn refactoring

vpisarev:
**WIP**
some changes in the core to assist refactoring of opencv_contrib's dnn module; see https://github.com/opencv/opencv_contrib/pull/1102

### DIFF
--- a/modules/core/include/opencv2/core/mat.hpp
+++ b/modules/core/include/opencv2/core/mat.hpp
@@ -1215,6 +1215,9 @@ public:
     /** @overload */
     Mat reshape(int cn, int newndims, const int* newsz) const;
 
+    /** @overload */
+    Mat reshape(int cn, const std::vector<int>& newshape) const;
+
     /** @brief Transposes a matrix.
 
     The method performs matrix transposition by means of matrix expressions. It does not perform the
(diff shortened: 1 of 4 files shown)


modules/core/include/opencv2/core/mat.hpp
```diff
@@ -1717,6 +1720,12 @@ class CV_EXPORTS Mat
      */
     size_t total() const;
 
+    /** @brief Returns the total number of array elements.
+
+     The method returns the number of elements within a certain sub-array slice with startDim <= dim < endDim
+     */
+    size_t total(int startDim, int endDim=INT_MAX) const;
```
alalek: This functionality looks misnamed. Probably `sliceSize` would be good.
Also my suggestion is to move this out of `cv::Mat` class to prevent exploding of cv::Mat with rarely used/specific functionality.

vpisarev: `sliceSize` is no better than `total`, `sliceTotal` does not sound right either. The method is small, but may be useful when you operate on tensors, in dnn this method is used several times

vpisarev:
:+1:

PR #8618, opened 2017-04-19
Expose UMat OpenCL handles to python

pwuertz:
This is a suggestion for adding OpenCL interoperability to the python UMat interface. The pull request exposes the `UMat::handle` method for obtaining native OpenCL buffer handles. I also added new static methods for obtaining the context and queue handles.

This allows running custom OpenCL kernels on UMat buffers using PyOpenCV (zero copy). Example usage:
```python
import numpy as np
import cv2
import pyopencl as cl
import pyopencl.reduction
import pyopencl.array

# Create OpenCL buffer via OpenCV UMat
a_np = np.arange(512, dtype=np.float32)
a_um = cv2.UMat(a_np)

# Obtain native OpenCL handles from OpenCV
ctx_handle = cv2.UMat.context()
queue_handle = cv2.UMat.queue()
a_um_handle = a_um.handle(cv2.ACCESS_READ)

# Access context, queue and buffer using PyOpenCL
ctx = cl.Context.from_int_ptr(ctx_handle)
queue = cl.CommandQueue.from_int_ptr(queue_handle)
a_cl = cl.Buffer.from_int_ptr(a_um_handle)

# Run PyOpenCL kernel on UMat data
a_cl_ary = pyopencl.array.Array(queue, a_np.shape, a_np.dtype, data=a_cl)
krnl = pyopencl.reduction.ReductionKernel(ctx, np.float32,
    neutral="0", reduce_expr="a+b", map_expr="x[i]",
    arguments="__global float *x")
sum_cl = krnl(a_cl_ary).get()
assert np.allclose(sum_cl, a_np.sum())
```

### DIFF
--- a/modules/core/include/opencv2/core/mat.hpp
+++ b/modules/core/include/opencv2/core/mat.hpp
@@ -2464,6 +2464,10 @@ public:
     UMat& operator = (UMat&& m);
 #endif
 
+    /*! Returns the OpenCL buffer handle on which UMat operates on.
+        The UMat instance should be kept alive during the use of the handle to prevent the buffer to be
+        returned to the OpenCV buffer pool.
+     */
     void* handle(int accessFlags) const;
     void ndoffset(size_t* ofs) const;
 
--- a/modules/python/src2/cv2.cpp
+++ b/modules/python/src2/cv2.cpp
@@ -508,13 +508,71 @@ static PyObject * UMatWrapper_get(cv2_UMatWrapperObject* self)
     return pyopencv_from(m);
 }
 
+// UMatWrapper.handle() - returns the OpenCL handle of the UMat object
+static PyObject * UMatWrapper_handle(cv2_UMatWrapperObject* self, PyObject *args, PyObject *kwds)
+{
+    const char *kwlist[] = {"accessFlags", NULL};
+    int accessFlags;
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "i", (char**) kwlist, &accessFlags))
+        return 0;
+    return PyLong_FromVoidPtr(self->um->handle(accessFlags));
+}
+
+// UMatWrapper.isContinuous() - returns true if the matrix data is continuous
+static PyObject * UMatWrapper_isContinuous(cv2_UMatWrapperObject* self)
+{
+    return PyBool_FromLong(self->um->isContinuous());
+}
+
+// UMatWrapper.isContinuous() - returns true if the matrix is a submatrix of another matrix
+static PyObject * UMatWrapper_isSubmatrix(cv2_UMatWrapperObject* self)
+{
+    return PyBool_FromLong(self->um->isSubmatrix());
+}
+
+// UMatWrapper.context() - returns the OpenCL context used by OpenCV UMat
+static PyObject * UMatWrapper_context(cv2_UMatWrapperObject*)
+{
+    return PyLong_FromVoidPtr(cv::ocl::Context::getDefault().ptr());
+}
+
+// UMatWrapper.context() - returns the OpenCL queue used by OpenCV UMat
+static PyObject * UMatWrapper_queue(cv2_UMatWrapperObject*)
+{
+    return PyLong_FromVoidPtr(cv::ocl::Queue::getDefault().ptr());
+}
+
+static PyObject * UMatWrapper_offset_getter(cv2_UMatWrapperObject* self, void*)
+{
+    return PyLong_FromSsize_t(self->um->offset);
+}
+
 static PyMethodDef UMatWrapper_methods[] = {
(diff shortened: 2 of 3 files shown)


alalek:
1) Required lifetime control policies for these handles: specification and implementation.
2) Tests are required.

pwuertz:
@alalek 
Thanks for reviewing this!

2) I was tempted to use the example I posted above as a test, but I wasn't sure about the dependency on `pyopencl` I would force on the test suite. The alternative is to test the API only (just obtain the handles) and not the functionality itself to avoid additional dependencies. Which solution is preferred?

1) Could you clarify what you mean by that? I'm just granting access to the existing OpenCV C++ API. The python interface does not modify the specification, behavior and implementation of those methods in any way. The handles obtained from the C++ API are just raw pointers where the user alone is responsible for correct handling and lifetime considerations.

That being said, the use of PyOpenCL makes the handling of foreign resources very convenient. All of its from-int-pointer methods are calling the OpenCL reference counting functions by default, so OpenCL won't delete a buffer that is dropped by OpenCV but is still in use by PyOpenCL.

Enforcing such a lifetime control by OpenCV would require a hard dependency on a specific target library like PyOpenCL. My understanding of the existing OpenCV API is that it is the intent to support generic OpenCL interoperability, which comes with all the risks of providing raw pointers.

alalek:
- **Tests**
  Yes, it is enough to have just OpenCV API calls without pyopencl requirement, to guarantee that these API calls doesn't crash user program at least.

- **Handle lifetime / usage scope**
    I mean we should track cases like these if it is possible. Or specify that UMat lifetime should be guarantee to cover pyopencl buffer creation (and probably usage of this constructed Buffer, see below): 

```.diff
 a_um_handle = a_um.handle(cv2.ACCESS_READ)
+a_um = None # clear UMat and related OpenCL handle

 ...
 a_cl = cl.Buffer.from_int_ptr(a_um_handle) # will not work with invalid 'a_um_handle'
```

OpenCV uses OpenCL buffer pool, so returned handle can be reused and owned by another UMat instance if the current UMat is released.

IMHO, Instead of "get handle" design, we may use more strong "move ownership" semantic to avoid such problems: 

```
a_um_handle = a_um.detach_opencl_handle()
# a_um becomes empty/invalid
# a_um_handle is not used by OpenCV anymore and lifetime control is fully transferred to "user code"
```

- **ROI**
    UMat is just a "view" over some storage (clBuffer in this case). Information about handle is not enough - you need "layout" information too. At least, we should check that it is not required in case of whole buffer ROI.

pwuertz:
- __Tests__
    I added the suggested test for obtaining the OpenCL pointer from OpenCV.

---

The other points you mentioned apply to the current OpenCV API in general, not the python wrapper in particular. Just to clarify, I don't intend to open up a huge can of worms, i.e. redesigning the OpenCL interoperability API (at least not in this PR). I just want to access the existing OpenCV API for low level OpenCL access from python, including all its pitfalls and dangers.

The danger of discarded data due to buffer pool reassignments is certainly an important piece of information. If the OpenCV memory pool only tracks UMat refcounts and not OpenCL buffer refcounts it should indeed be advised to keep UMat instances alive when using the native buffers. The best place for this is probably the (not yet existent) `UMat::handle` documentation. 

The other point you mentioned is the memory layout information used by UMat for slicing/viewing multidimensional arrays. Of course there is no way of mapping the UMat superstructure to plain OpenCL, which a user that requests a raw buffer should be aware of. It is up to the user to take memory layout information (dimensions, strides, offsets) into account when interpreting the data (no safety net in `UMat::handle` either). Do you think I should add `UMat::isContinuous` and `UMat::offset` to the python interface as well so a user can check certain layout assumptions?

True, a direct UMat to PyOpenCL Array mapping would be way easier and safer to use, but I'm not planning on defining a new API. Also, someone might want to use a different library one day and require the low level API anyways.

alalek:
This make sense. Lets start with something that just works.

Could you take a look on Win32 builder problem?

pwuertz:
Currently I don't have a win32 system to check on this. I was trying to experiment by accepting long int instead of plain int just in case there is a problem with the accessFlags int type. But now the Win32 tests are not_queued by the build bot anymore. Is there a way to trigger this platform?

pwuertz:
Ok, apparently the win32 tests were being delayed. Accepting long int instead of int didn't fix the problem. I'm reverting it back.

I have no clue why PyArg_ParseTupleAndKeywords wouldn't accept `cv2.ACCESS_READ` as int argument on Win32. I don't think I can investigate this without a testing platform.

pwuertz:
All right, I managed to set up a 32bit build on Windows and ran the tests successfully now. The problem was a missing `METH_KEYWORDS` flag when registering the `handle` method. For some reason python is letting this slip on the other platforms/builds.

alalek:
Thank you! :+1:

PR #8620, opened 2017-04-20
Issue 8457: Updated comments for cv::ellipse and cv::ellipse2Poly to clarify some…

### LINKED ISSUE #8457: Unexpected behavior of ellipse()
- OpenCV => 3.2 (not verified on other versions)
- Operating System / Platform => Windows 64 Bit / Linux 32-bit arm
- Compiler => Visual Studio 2015 / g++ 4.8.4

If ellipse segment is drawn such that start and end angles "stradle" 0 degrees, arc is drawn correctly if start angle is negative (e.g.  -5), differently if start angle is large positive (e.g., 355).

If drawing an ellipse from, for example 355 degrees to 45 degrees, if the start angle is expressed as -5 degrees, the arc goes from -5 to +45 as expected.  If the start angle is expressed as 355, the arc goes "around the long way", i.e., it is drawn from 45 to 355 degrees.  Happens with line or filled.

Trivial example:

	cv::Mat img1(360, 640, CV_8UC1, cv::Scalar(255));
        cv::Mat img2(360, 640, CV_8UC1, cv::Scalar(255));

	cv::Point center(320, 180);
	cv::Size axes(100, 100);

	cv::ellipse(img1, center, axes, 0.0, -5.0, 45.0, cv::Scalar(0), -1);
	cv::ellipse(img2, center, axes, 0.0, 355.0, 45.0, cv::Scalar(0), -1);

	cv::imshow("As expected", img1);
        cv::imshow("Different", img2);
	cv::waitKey(0);

Not sure if I missed something in the documentation or this is intended.  I suspect a note in the documentation would be enough otherwise the behavior would be changed.

saskatchewancatch:
… behaviour that has confused some users. 
resolves #8457 

### This pullrequest changes
1. The docs for:
1.1. cv::ellipse: it appears "less" should have been "more,"  since the other cv::ellipse function takes less parameters than the first one and draws an ellipse specified by a bounding rectangle. Also makes clear as a user has noticed, that start and end angles are swapped if start angle is greater than end angle.
1.2 cv::ellipse2Poly: Similar change, as cv::ellipse uses cv::ellipse2Poly.

### DIFF
--- a/modules/imgproc/include/opencv2/imgproc.hpp
+++ b/modules/imgproc/include/opencv2/imgproc.hpp
@@ -4220,13 +4220,14 @@ CV_EXPORTS_W void circle(InputOutputArray img, Point center, int radius,
 
 /** @brief Draws a simple or thick elliptic arc or fills an ellipse sector.
 
-The function cv::ellipse with less parameters draws an ellipse outline, a filled ellipse, an elliptic
+The function cv::ellipse with more parameters draws an ellipse outline, a filled ellipse, an elliptic
 arc, or a filled ellipse sector. The drawing code uses general parametric form.
 A piecewise-linear curve is used to approximate the elliptic arc
 boundary. If you need more control of the ellipse rendering, you can retrieve the curve using
 cv::ellipse2Poly and then render it with polylines or fill it with cv::fillPoly. If you use the first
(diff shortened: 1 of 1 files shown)


sovrasov:
Please, squash commits into one before merge.

sovrasov:
:+1:

PR #8640, opened 2017-04-24
doc: change MathJax URL

### LINKED ISSUE #8630: MathJax CDN is shutting down
OpenCV uses MathJax in its Doxygen documentation. Like most users, OpenCV links to an external copy of the library hosted at the official CDN, cdn.mathjax.com. MathJax is [shutting down][] cdn.mathjax.org on April 31st. 

OpenCV needs to either update the doxygen config to point at an [alternative CDN][]  or self-host the mathjax libraries.

All the old documentation will need to be rebuilt as well.

https://github.com/opencv/opencv/blob/383559c2285baaf3df8cf0088072d104451a30ce/doc/Doxyfile.in#L177

[shutting down]: https://www.mathjax.org/cdn-shutting-down/
[alternative cdn]: https://www.mathjax.org/cdn-shutting-down/#alternative-cdn-providers

alalek:
resolves #8630


alalek:
:+1:

PR #8645, opened 2017-04-25
fix ByteStreamBuffer

brian-armstrong-discord:
The changes to make this handle all seek directions broke the ByteStreamBuffer. This patch fixes it so that inmemory exif orientation works.


alalek:
:+1:

PR #8655, opened 2017-04-26
core: fix persistence bug in RAW I/O code

alalek:
- persistence.cpp code expects special sizeof value for passed structures
- this assumption is lead to memory corruption problems
- fixed/workarounded test to prevent memory corruption on Linux 32-bit systems


alalek:
:+1:

PR #8656, opened 2017-04-26
cmake: fix POPCNT & SSE4.2 detection code

alalek:
Fix problem with Linux 32-bit builds:
```
-- Performing Test HAVE_CXX_MPOPCNT (check file: cmake/checks/cpu_popcnt.cpp)
-- Performing Test HAVE_CXX_MPOPCNT - Failed
-- POPCNT is not supported by C++ compiler
-- Performing Test HAVE_CXX_MSSE4_2 (check file: cmake/checks/cpu_sse42.cpp)
-- Performing Test HAVE_CXX_MSSE4_2 - Failed
-- SSE4_2 is not supported by C++ compiler
```

### DIFF
--- a/cmake/checks/cpu_popcnt.cpp
+++ b/cmake/checks/cpu_popcnt.cpp
@@ -1,8 +1,22 @@
-#include <nmmintrin.h>
-#ifndef _MSC_VER
-#include <popcntintrin.h>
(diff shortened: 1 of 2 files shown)


alalek:
:+1:

PR #8668, opened 2017-04-28
SoftFloat integrated 

savuor:
### This PR changes

[SoftFloat](http://www.jhauser.us/arithmetic/SoftFloat.html) is a software implementation of floating-point calculations according to IEEE 754 standard.
All calculations are done in integers, that's why they are machine-independent and bit-exact.
This library can be useful in accuracy-critical parts like look-up tables generation, tests, etc.

Data types (others were removed from the code):
* **softfloat32_t** binary compatible with **float**
* **softfloat64_t** binary compatible with **double**

Features:
* Addition, subtraction, multiplication, division, square root
* Conversion to/from int32/int64/float/double
* exp(), log() and cbrt() (for float32_t only) were taken from OpenCV and rewritten using SoftFloat primitives


vpisarev:
I would put it under cv namespace, but renamed float32_t to softfloat32_t etc.

savuor:
@vpisarev I've refactored the types completely, please review that

vpisarev:
@savuor, I'm looking at the very latest version of softfloat.hpp (https://github.com/savuor/opencv/blob/c0210151736dfbe0425a6c3e28e5f51d67d3552e/modules/core/include/opencv2/core/softfloat.hpp) and, as I see, many of my questions/requests are still unaddressed. Could you please make another set of changes?

modules/core/include/opencv2/core/softfloat.hpp
```diff
@@ -0,0 +1,325 @@
...
+    softfloat32_t( const uint64_t );
+    softfloat32_t( const int32_t );
+    softfloat32_t( const int64_t );
+    softfloat32_t( const float a ) { Cv32suf s; s.f = a; v = s.u; }
+
+    uint_fast32_t toUI32_minMag( bool exact = false ) const;
+    uint_fast64_t toUI64_minMag( bool exact = false ) const;
+    int_fast32_t   toI32_minMag( bool exact = false ) const;
+    int_fast64_t   toI64_minMag( bool exact = false ) const;
+    uint_fast32_t toUI32( uint_fast8_t roundingMode = round_near_even, bool exact = false ) const;
```
savuor: `operator int() const` and `operator float() const` create ambiguity when casting to int, which is needed, for example, for GTEST checks. The problem could be solved with explicit casts but they are not allowed until C++11.
That is why there will be no `operator int() const`. There will be separate `trunc()` function outside the classes.

modules/core/include/opencv2/core/softfloat.hpp
```diff
@@ -0,0 +1,312 @@
...
+    tininess_beforeRounding = 0,
+    tininess_afterRounding  = 1
+};
+//fixed to make softfloat code stateless
+const uint_fast8_t globalDetectTininess = tininess_afterRounding;
+
+/*----------------------------------------------------------------------------
+| Software floating-point rounding mode.
+*----------------------------------------------------------------------------*/
+enum {
```
savuor: Now round_near_even is used by in round() function and round_minMag which is supposed to be faster is used in trunc() function.

savuor:
@vpisarev I've fixed all the things you mentioned in review. Can we merge this PR now?

savuor:
@vpisarev I've fixed all the things you mentioned in review.

modules/core/include/opencv2/core/softfloat.hpp
```diff
@@ -0,0 +1,240 @@
...
+    static const softfloat fromRaw( const uint32_t a ) { softfloat x; x.v = a; return x; }
+
+    explicit softfloat( const uint32_t );
+    explicit softfloat( const uint64_t );
+    explicit softfloat( const int32_t );
+    explicit softfloat( const int64_t );
+    explicit softfloat( const float a ) { Cv32suf s; s.f = a; v = s.u; }
+
+    operator softdouble() const;
+    operator float() const { Cv32suf s; s.u = v; return s.f; }
```
vpisarev: operator int() is missing

modules/core/include/opencv2/core/softfloat.hpp
```diff
@@ -0,0 +1,240 @@
...
+    static softdouble  nan() { return softdouble::fromRaw( CV_BIG_INT(0x7FFFFFFFFFFFFFFF) ); }
+    static softdouble  one() { return softdouble::fromRaw( (uint_fast64_t)( 1023) << 52 ); }
+
+    uint64_t v;
+};
+
+/*----------------------------------------------------------------------------
+*----------------------------------------------------------------------------*/
+
+CV_EXPORTS int_fast64_t round( const softfloat&  a );
```
vpisarev: round and trunc should return int, not int_fast64_t. instead of trunc(), implement "operator int() const" inside the classes. floor() and ceil() would be also useful

vpisarev:
@savuor, now the patch looks great! please, fix a few remaining issues and then it can be merged in.

savuor:
@vpisarev  Everything's done.
* As I said earlier, adding `operator int()` will cause ambiguity with `operator float()` and break GTEST compilation. This could be solved with explicit cast operators but they are not allowed until C++11.
* Since `floor`, `round` and `ceil` are conflicting with the functions from `std`, they are named `cvRound`, `cvFloor` and `cvTrunc`. They are moved outside `cv` namespace (as well as other overloads are) to avoid compilation issues.

vpisarev:
:+1:

javierBarandiaran:
opencv_core does not build with this change in Visual Studio 12 2013 Win64. I have to comment 
`#include "opencv2/core/softfloat.hpp"`
these types uint32_t, uint64_t are not defined.

modules/ts/include/opencv2/ts/ts_gtest.h
```diff
@@ -10334,7 +10334,12 @@ class TypeWithoutFormatter<T, kConvertibleToInteger> {
   // T is not an enum, printing it as an integer is the best we can do
   // given that it has no user-defined printer.
   static void PrintValue(const T& value, ::std::ostream* os) {
+    // MSVC warns about implicitly converting from double and float to int for
+    // possible loss of data, so we need to temporarily disable the
+    // warning.
+    GTEST_DISABLE_MSC_WARNINGS_PUSH_(4244)
```
alalek: This should be reverted. Implicit conversions in tests should be avoided. Use explicit conversions instead like other tests.

alalek:
@javierBarandiaran Please check: #8868

modules/core/include/opencv2/core/softfloat.hpp
```diff
@@ -0,0 +1,271 @@
...
+// warranties of merchantability and fitness for a particular purpose are disclaimed.
+// In no event shall the Intel Corporation or contributors be liable for any direct,
+// indirect, incidental, special, exemplary, or consequential damages
+// (including, but not limited to, procurement of substitute goods or services;
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
```
alalek: Please avoid using of legacy OpenCV license headers.
Original file license (below) should be placed unchanged.

Example:
```
// of this distribution and at http://opencv.org/license.html

// This file is based on files from package issued with the following license:
... original license ...
```

modules/core/src/softfloat.cpp
```diff
@@ -0,0 +1,4793 @@
...
+    *------------------------------------------------------------------------*/
+    shiftDist = 0xBE - exp;
+    if ( 64 <= shiftDist ) {
+        if ( exact && (exp | sig) ) {
+            raiseFlags(flag_inexact);
+        }
+        return 0;
+    }
+    /*------------------------------------------------------------------------
+    *-
```
alalek: This mess of floating-point numbers can't guarantee bit-exactness at all.
All floating-point numbers should be converted to `int64_t`/`int32_t` (refer to original softfloat code).

P.S. GitHub links comment to wrong line, comment is for line [4400](https://github.com/opencv/opencv/pull/8668/files#diff-af9fdef372de6d687734904cf428479dR4400).

savuor:
@alalek Please take a look at PR #8925 which should fix the issues

PR #8671, opened 2017-04-29
fix test error on VS2012

### LINKED ISSUE #8670: test FileStorage_DMatch series fails on Visual Studio 2012

- OpenCV => master ( cea0e94376eff179115ee30fa67c4c4745375e7b )
- Operating System / Platform => Windows 7 Pro 64bit
- Compiler => Visual Studio 2012 Update 5

##### Detailed description
- ```opencv_test_core``` fails 
  - ```FileStorage_DMatch```
  - ```FileStorage_DMatch_vector```
  - ```FileStorage_DMatch_vector_vector```
```
[ RUN      ] Core_InputOutput.FileStorage_DMatch
C:\opencv\modules\core\test\test_io.cpp(1114): error:       Expected: fs_result.c_str()
      Which is: "%YAML:1.0\n---\nd: [ 1, 2, 3, -1.5000000000000000e+000 ]\n"
To be equal to: "%YAML:1.0\n---\nd: [ 1, 2, 3, -1.5000000000000000e+00 ]\n"
With diff:
@@ -1,3 +1,3 @@
 %YAML:1.0
 ---
-d: [ 1, 2, 3, -1.5000000000000000e+000 ]\n
+d: [ 1, 2, 3, -1.5000000000000000e+00 ]\n

[  FAILED  ] Core_InputOutput.FileStorage_DMatch (1 ms)
```
- The string is self is different, but the meaning is OK. It's just that the comparison is too strict.
- I looked in the code, and this line calls [```cv::write```](https://github.com/opencv/opencv/blob/931b32d1022dbb999ede6143307720910177d2c3/modules/core/include/opencv2/core/persistence.hpp#L1066) -> [```cv::writeScalar```](https://github.com/opencv/opencv/blob/931b32d1022dbb999ede6143307720910177d2c3/modules/core/src/persistence.cpp#L7174) -> [```icvYMLWriteReal```](https://github.com/opencv/opencv/blob/931b32d1022dbb999ede6143307720910177d2c3/modules/core/src/persistence.cpp#L7174) and at the end the call arrives to [```icvDoubleToString```](https://github.com/opencv/opencv/blob/931b32d1022dbb999ede6143307720910177d2c3/modules/core/src/persistence.cpp#L973)
- in this ```icvDoubleToString```, there is a ```sprintf```, and this behaviour seems to differ from VS2012 and VS2015.
- Thus, for VS2012 and older, the expectation string should be modified.
- I'll make a patch as a PR later

##### Steps to reproduce
- ```cmake``` and build ```opencv_test_core``` from master
- run ```opencv_test_core --gtest_filter=*Dmatch*```

tomoaki0705:
resolves #8670

### DIFF
--- a/modules/core/test/test_io.cpp
+++ b/modules/core/test/test_io.cpp
@@ -1059,7 +1059,11 @@ TEST(Core_InputOutput, FileStorage_DMatch)
 
     EXPECT_NO_THROW(fs << "d" << d);
     cv::String fs_result = fs.releaseAndGetString();
+#if defined _MSC_VER && _MSC_VER <= 1700 /* MSVC 2012 and older */
+    EXPECT_STREQ(fs_result.c_str(), "%YAML:1.0\n---\nd: [ 1, 2, 3, -1.5000000000000000e+000 ]\n");
+#else
     EXPECT_STREQ(fs_result.c_str(), "%YAML:1.0\n---\nd: [ 1, 2, 3, -1.5000000000000000e+00 ]\n");
+#endif
 
     cv::FileStorage fs_read(fs_result, cv::FileStorage::READ | cv::FileStorage::MEMORY);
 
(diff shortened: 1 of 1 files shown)


alalek:
Thank you! :+1:

Issue #4542, opened 2015-07-27
Inpainting support for float32 data


opencv-pushbot:
Transferred from http://code.opencv.org/issues/2944

```
|| David Shean on 2013-04-08 06:12
|| Priority: Normal
|| Affected: None
|| Category: photo
|| Tracker: Feature
|| Difficulty: None
|| PR: 
|| Platform: None / None
```
## Inpainting support for float32 data

```
The geospatial data community often works with single-band int16 or float32 raster data (e.g. digital surface models [DSM], spectrometer data) that contain "holes" of nodata pixels (e.g., bad pixels, data drops during transfer from spacecraft, occlusions in photogrammetric DSM products).  For some analyses, it is desirable to preserve these nodata values, but often, a continuous data product is required (e.g. removal of buildings to produce a digital terrain model, using a DSM for image orthorectification, etc.).  

Contemporary inpainting methods offer highly effective hole-filling capabilities.  If the OpenCV inpainting functionality offered support for float32 inputs, it would definitely be used by new communities outside of traditional CV/photography.  Members of the astronomy community also face similar issues (http://astrolitterbox.blogspot.com/2012/03/healing-holes-in-arrays-in-python.html).

A quick glance at https://github.com/Itseez/opencv/blob/master/modules/photo/src/inpaint.cpp suggests that the output of both single-band inpainting methods is recast to 8-bit:

icvTeleaInpaintFMM
sat = (float)((Ia/s+(Jx+Jy)/(sqrt(Jx*Jx+Jy*Jy)+1.0e-20f)+0.5f));
CV_MAT_ELEM(*out,uchar,i-1,j-1) = cv::saturate_cast<uchar>(sat);

icvNSInpaintFMM
Ia += (float)w * (float)(CV_MAT_ELEM(*out,uchar,km,lm));
CV_MAT_ELEM(*out,uchar,i-1,j-1) = cv::saturate_cast<uchar>((double)Ia/s); #s is also float

Seems like it would be relatively straightforward to support CV_32F?  I'm still learning, and I'm not quite ready to tackle on my own at this point.  Thanks.
```
##### François-Xavier Thomas on 2015-05-04 14:50

```
I second this. I got a somewhat working version by pre-scaling my raster array to 8-bit, then scaling it back after calling @cv2.inpaint@, however it would be great if that use case were covered by OpenCV directly.
```

pstack12:
I also have a use for the inpainting function to support floats or doubles.

PR #8674, opened 2017-04-29
Inpainting support for all 1-channel input images

belgraviton:
Solution for #4542 issue

### This pullrequest changes
New version of inpaint function is able to process all 1-channel and 8-bit 3-channel input images.
Old function can inpaint only 8-bit 1-channel and 3-channel images.

All 1-channel images are processing with double data type.
Inpaint function returns image with source image data type.

Processing of 8-bit 3-channel images is not changed.

### DIFF
--- a/modules/photo/src/inpaint.cpp
+++ b/modules/photo/src/inpaint.cpp
@@ -67,6 +67,10 @@ min4( float a, float b, float c, float d )
 #define INSIDE 2  //unknown
 #define CHANGE 3  //servise
 
+// Processing data types
+typedef double data_type;
+int data_type_cv = CV_64F;
+
 typedef struct CvHeapElem
 {
     float T;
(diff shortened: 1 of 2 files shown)


belgraviton:
Mean test time over  100 runs on inpaint test image from contrib_extra (release, 64bit, Win7, i7 3770):
double data type: 0.0632 (NS), 0.0972 (Telea)
uchar data type: 0.0624 (NS), 0.0948 (Telea)

modules/photo/src/inpaint.cpp
```diff
@@ -463,31 +467,31 @@ icvTeleaInpaintFMM(const CvMat *f, CvMat *t, CvMat *out, int range, CvPriorityQu
 
                               if (CV_MAT_ELEM(*f,uchar,k,l+1)!=INSIDE) {
                                  if (CV_MAT_ELEM(*f,uchar,k,l-1)!=INSIDE) {
-                                    gradI.x=(float)((CV_MAT_ELEM(*out,uchar,km,lp+1)-CV_MAT_ELEM(*out,uchar,km,lm-1)))*2.0f;
+                                    gradI.x=(float)((CV_MAT_ELEM(*out,data_type,km,lp+1)-CV_MAT_ELEM(*out,data_type,km,lm-1)))*2.0f;
```
sovrasov: I think `data_type` can be introduced in `icvTeleaInpaintFMM` and etc. as a template parameter, so conversion to `double` will become unnecessary.

PR #8676, opened 2017-04-29
Add default parameters for SGBM::create

jveitchmichaelis:
### This pullrequest changes

Adds some default parameters to `SGBM::create`, namely mindisparity, disparity range and SAD window size.

Fixes a bug where `cv::StereoSGBM::load<cv::StereoSGBM>` would not compile because `SGBM::create` has a few parameters with no default values.

Previously the code below would fail to compile on MSVC2015 (Error C2660):

`auto matcher = cv::StereoSGBM::load<cv::StereoSGBM>("./some_path/stereo_sgbm_params.xml");`

### DIFF
--- a/modules/calib3d/include/opencv2/calib3d.hpp
+++ b/modules/calib3d/include/opencv2/calib3d.hpp
@@ -1917,7 +1917,7 @@ public:
     set StereoSGBM::numDisparities at minimum. The second constructor enables you to set each parameter
     to a custom value.
      */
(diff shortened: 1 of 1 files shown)


sovrasov:
@jveitchmichaelis please, squash commits into one before merge.

jveitchmichaelis:
@sovrasov squashed :)

sovrasov:
:+1:

PR #8688, opened 2017-05-02
Use all available cpu cores in iOS/OSX build procedure

ysolovyov:
### This pullrequest changes

Use all available cpu cores in iOS/OSX build procedure


vpisarev:
:+1:

PR #8694, opened 2017-05-04
Parallelize Canny with custom gradient

mschoeneck:
### Parallelizing Canny with custom gradient and restructuring source code

I expanded the Canny algorithm to parallelize Canny with custom gradient. I have also restructured the source code to avoid unnecessary checks and aligned internal arrays (mag buffer, map image and maximum gradient buffer in case of a color image input).

I measured a performance gain of up to 15% compared to the actual implementation.


modules/imgproc/src/canny.cpp
```diff
@@ -285,422 +286,510 @@ static bool ocl_Canny(InputArray _src, const UMat& dx_, const UMat& dy_, OutputA
...
-    ~parallelCanny()
+    parallelCanny(const Mat &_dx, const Mat &_dy, Mat &_map, std::deque<uchar*> &borderPeaksParallel,
+                  int _low, int _high, bool _L2gradient) :
+        src(_dx), src2(_dy), map(_map), _borderPeaksParallel(borderPeaksParallel),
+        low(_low), high(_high), aperture_size(0), L2gradient(_L2gradient)
     {
+#if CV_SIMD128
+        haveSIMD = hasSIMD128();
+        if(haveSIMD)
+            _map.create(src.rows + 2, src.cols + CV_MALLOC_SIMD128 + (CV_MALLOC_SIMD128 - (src.cols & (CV_MALLOC_SIMD128 - 1))), CV_8UC1);
```
alalek: Probably the second arg is equal to:
```
alignSize(src.cols + 2 * CV_MALLOC_SIMD128, CV_MALLOC_SIMD128);
```

modules/core/include/opencv2/core/hal/intrin.hpp
```diff
@@ -433,6 +433,29 @@ template <> struct V_RegTrait128<double> {
 };
 #endif
 
+inline unsigned int trailingZeros32(unsigned int value) {
```
alalek: Could you remove duplicate of this function from [`contours.cpp`](https://github.com/opencv/opencv/blob/17eef4d8a967a3541ebeefa44b80cf8001b0d5a5/modules/imgproc/src/contours.cpp#L55) file and use this one?

alalek:
Thank you! :+1:

PR #8697, opened 2017-05-05
imgproc: fix BGRA2BGRA conversion

### LINKED ISSUE #8696: cvtColor() sometimes swaps R and B channel when not asked

- OpenCV => 3.2.0 (as extracted from OpenCV-3.2.0-vc14.exe)
- Operating System / Platform =>windows 7, 64-bit
The problem was originally seen by a self-compiled OpenCV 3.1 for Android.
The example program is for Windows as shown.

##### Detailed description

The call `cv::cvtColor(image, image, cv::COLOR_BGR2BGRA);`, when the Mat is already 4 channels, swaps R and B channels.

##### Steps to reproduce

```.cpp
// This small demo shows that RGB to RGBA conversion (or BGR to BGRA, that is the same)
// has a bug when the input matrix already is 4 channels.
// Tried with OpenCV-3.2.0-vc14.exe (extracted)

#include "stdafx.h" // silly windows stuff, generated by Visual Studio
#include <iostream>

int _tmain(int argc, _TCHAR* argv[])
{
	// Choose an image that has clear red and/or blue objects.
	std::string inputImage = "any red-blue picture really.jpg";

	cv::Mat image = cv::imread(inputImage);
	if (image.empty())
	{
		std::cerr << "Cannot read image file: " << inputImage << std::endl;
		return -1;
	}

	// Make sure to have 4 channels
	if (image.channels() == 3) {
		cv::cvtColor(image, image, cv::COLOR_BGR2BGRA);
	}

	// This image is ok
	cv::imshow("ok", image);

	// This conversion should do nothing visible
	cv::cvtColor(image, image, cv::COLOR_BGR2BGRA);

	// Now R and B are swapped.
	cv::imshow("r-and-b-swapped", image);

	// And here they are swapped again (back to the way they should be).
	cv::cvtColor(image, image, cv::COLOR_BGR2BGRA);
	cv::imshow("r-and-b-swapped-back", image);

	cv::waitKey();
}
```

sovrasov:
resolves #8696

### DIFF
--- a/modules/imgproc/test/test_color.cpp
+++ b/modules/imgproc/test/test_color.cpp
@@ -2190,3 +2190,28 @@ TEST(ImgProc_Bayer2RGBA, accuracy)
         }
     }
 }
+
+TEST(ImgProc_BGR2RGBA, regression_8696)
+{
+    Mat src(Size(80, 10), CV_8UC4);
+    src.setTo(Scalar(255, 0, 200, 100));
+
+    Mat dst;
+    cvtColor(src, dst, COLOR_BGR2BGRA);
+
+    EXPECT_DOUBLE_EQ(norm(dst - src, NORM_INF), 0.);
+}
+
+TEST(ImgProc_BGR2RGBA, 3ch24ch)
+{
+    Mat src(Size(80, 10), CV_8UC3);
+    src.setTo(Scalar(200, 0, 200));
+
+    Mat dst;
+    cvtColor(src, dst, COLOR_BGR2BGRA);
+
+    Mat expected(Size(80, 10), CV_8UC4);
+    expected.setTo(Scalar(80, 0, 200, 255));
+
+    EXPECT_DOUBLE_EQ(norm(expected - dst, NORM_INF), 0.);
+}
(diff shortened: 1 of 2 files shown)


Rhialto:
I'm looking at this code for the first time. Should a similar fix be applied to the code starting about line 800 of the same color.cpp file?

sovrasov:
Yes, it should be fixed too.

modules/imgproc/test/test_color.cpp
```diff
@@ -2190,3 +2190,14 @@ TEST(ImgProc_Bayer2RGBA, accuracy)
         }
     }
 }
+
+TEST(ImgProc_BGR2RGBA, regression)
+{
+    Mat src(Size(10, 10), CV_8UC4);
```
alalek: 1. Please use issue number in regression test: `regression_8696`
2. It is better to use bigger size to enable checks for vectorized code too: `Size(80,10)`

alalek:
BGRA2BGRA is noop operation in case of 4 channel images. Probably it should just make checks and bailout without any actual image processing. In this case changes with `bidx` should be dropped (or asserted via `CV_Assert` macro) because they are not too fast.

Rhialto:
It depends; the initial value of the A channel is ambiguously documented: "If conversion adds the alpha channel, its value will set to the maximum of corresponding channel range". In case of a 3-channel image in a 4-channel matrix, one can argue whether this constitutes "adding" the alpha channel. (I tend towards saying that it does, even if this is less efficient than considering the operation a no-op).

alalek:
@Rhialto Thanks!
Probably we need new test for BGR -> BGRA (with alpha channel addition) too.

vpisarev:
:+1:

PR #8698, opened 2017-05-05
fix Mat::release() resets channels info for Mat_

### LINKED ISSUE #8680: Mat::release() resets channels info for Mat_
In debug Mat::release() resets fixed info for Mat_ (pls see https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/mat.inl.hpp#L780).
For instance if Mat_ is compiled with 3 channels, release() sets it to 1.

cdcseacave:
resolves #8680 

test code:
```
void main() {
   cv::Mat_<cv::Point2i> mat(3,1);
   assert(mat.channels() == 2);
   mat.release();
   assert(mat.channels() == 2);
}
```

### DIFF
--- a/modules/core/include/opencv2/core/mat.hpp
+++ b/modules/core/include/opencv2/core/mat.hpp
@@ -2185,6 +2185,8 @@ public:
     void create(Size _size);
     //! equivalent to Mat::create(_ndims, _sizes, DatType<_Tp>::type)
     void create(int _ndims, const int* _sizes);
+    //! equivalent to Mat::release()
+    void release();
     //! cross-product
     Mat_ cross(const Mat_& m) const;
     //! data type conversion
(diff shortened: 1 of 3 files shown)


modules/core/include/opencv2/core/mat.hpp
```diff
@@ -2138,6 +2138,7 @@ template<typename _Tp> class Mat_ : public Mat
     void create(Size _size);
     //! equivalent to Mat::create(_ndims, _sizes, DatType<_Tp>::type)
     void create(int _ndims, const int* _sizes);
+	void release();
```
sovrasov: Bad indentation in this line (probably tab has been used instead of spaces).
Also the new method should be commented as the previous one:
```C++
//! equivalent to Mat::release()
  void release();
```

sovrasov:
:+1:

PR #8708, opened 2017-05-10
Fixed out of bound reading in makeAgastOffsets

### LINKED ISSUE #7406: Illegal array access in makeAgastOffsets

- OpenCV => d3930cdee1a0c7124045883b3062f25b3027fa97
- Operating System / Platform => OSX / 10.10
- Compiler => clang
##### Detailed description

In makeAgastOffsets, there is a 16-long for-loop that indexes past the end of an 8- or 12-element array.

In particular, this part:
  for( ; k < 16; k++ ) 
               pixel[k] = offsets[k][0] + offsets[k][1] \* rowStride;

When makeAgastOffsets is called with a value of type that is NOT AgastFeatureDetector::OAST_9_16, such as AgastFeatureDetector::AGAST_5_8, the "offsets" array is 8 or 12 elements long, not 16, and the code will attempt to read past the end off "offsets"

##### Steps to reproduce

Build opencv with "-fsanitize=address", and then call something that eventually goes into makeAgastOffsets, such as cv::BRISK_Impl::detectAndCompute and watch what the values going into pixels[k] are.

terfendail:
resolves #7406

### This pullrequest changes
Restricted pixel offset estimation in makeAgastOffset to actually described amount of pixels

### DIFF
--- a/modules/features2d/src/agast_score.cpp
+++ b/modules/features2d/src/agast_score.cpp
@@ -82,10 +82,15 @@ void makeAgastOffsets(int pixel[16], int rowStride, int type)
                               type == AgastFeatureDetector::AGAST_7_12s ? offsets12s :
                               type == AgastFeatureDetector::AGAST_5_8 ? offsets8  : 0;
 
+    const int offsets_len = type == AgastFeatureDetector::OAST_9_16 ? 16 :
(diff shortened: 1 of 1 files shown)


vpisarev:
:+1:

Issue #8709, opened 2017-05-10
Can't build current head on MSVC


PhilLab:
##### System information (version)

- OpenCV =>HEAD
- Operating System / Platform => Windows 8.1 x64bit
- Compiler => Visual Studio 12 Win64

##### Detailed description

Current HEAD compilation fails with
```
modules\core\src\mathfuncs_core.simd.hpp(553): error C2440: 'type cast' : cannot convert from '__m128i' to '__m128' 
modules\core\src\mathfuncs_core.simd.hpp(553): error C2660: '_mm256_insertf128_ps' : function does not take 2 arguments
```
on Visual Studio. This happens [in this line](https://github.com/opencv/opencv/blob/1e6ce1d2f827b5c04b0e7797e88c37a955563e31/modules/core/src/mathfuncs_core.simd.hpp#L553) which was committed in #8441 . [Stack Overflow](http://stackoverflow.com/a/13632812/1531708) says that MSVC needs ``_mm_castps_si128``

##### Steps to reproduce
Configure your solution with the following command (of course provided you have the third party libs organized as we do)
```
cmake -G "Visual Studio 12 Win64" 
  -DCMAKE_INSTALL_PREFIX=C:/opencv.install
  -DBUILD_DOCS=OFF
  -DBUILD_PERF_TESTS=OFF
  -DBUILD_TESTS=OFF 
  -DOPENCV_EXTRA_MODULES_PATH="C:/OpenCV/workspace/contrib/modules" 
  -DEIGEN_INCLUDE_PATH=C:/3rdparty.git/Eigen/3.3.2 
  -DWITH_OPENGL=ON
   -DWITH_OPENMP=ON
   -DGLOG_INCLUDE_DIR=C:/3rdparty.git/glog/0.3.4/include
  -DGLOG_LIBRARY=C:/3rdparty.git/glog/0.3.4/Release/libglog_static.lib 
  -DGlog_LIBS=C:/3rdparty.git/glog/0.3.4/Release/libglog_static.lib 
  -DCeres_DIR=C:/3rdparty.git/Ceres/1.12/CMake 
  -DVTK_DIR=C:/3rdparty.git/VTK/7.1.1/lib/cmake/vtk-7.1  
  "C:/OpenCV/workspace"
```
which generates this configuration:
```
 General configuration for OpenCV 3.2.0-dev =====================================
   Version control:               3.0.0-4748-g06b0fe3
 
   Extra modules:
[...]

PR #8710, opened 2017-05-10
Fixed exp32f() compilation on MSVC

PhilLab:
Resolving issue #8709


PhilLab:
@alalek Seems that none of the builds will start: ``not_queued``
image

alalek:
Thank you! :+1:

PR #8711, opened 2017-05-10
Aravis: Use of std::fabs, added support for 16-bit mono files and exposure compensation

### LINKED ISSUE #8707: OpenCV 3.2: Problems with The Aravis SDK Installation
Hello Guys,

I've found an issue on the OpenCV 3.2 release, if we run the installation with the Aravis SDK, certainly a problem is going to appear. The following file, [cap_aravis.cpp](https://github.com/opencv/opencv/blob/master/modules/videoio/src/cap_aravis.cpp), on the line 393, has a problem, because it is subtracting two floats and using [abs](http://www.cplusplus.com/reference/cstdlib/abs/?kw=abs) function. The correct approach is to use the [fabs](http://www.cplusplus.com/reference/cmath/fabs/?kw=fabs) function, in the [math.h](http://www.cplusplus.com/reference/cmath/?kw=math.h) header, to subtract floats.

The currently code is the following:
```c++
 if(exposureAvailable) {
            if(abs(exposure - ne) > 2) {
                // priority 2 - control of exposure time
                arv_camera_set_exposure_time(camera, (exposure = ne) );
                return;
            }
        }
```
The correct is:
```c++
 if(exposureAvailable) {
            if(fabs(exposure - ne) > 2) {
                // priority 2 - control of exposure time
                arv_camera_set_exposure_time(camera, (exposure = ne) );
                return;
            }
        }
```
@matheusns

ArkadiuszRaj:
resolves #8707 

### This pullrequest changes

Corrects the code to use only std::fabs() function.
Adds support for 16-bit mono files.

Additionally reduces number of  internal buffers.

### DIFF
--- a/modules/videoio/src/cap_aravis.cpp
+++ b/modules/videoio/src/cap_aravis.cpp
@@ -77,11 +77,13 @@
 //      video/x-raw, fourcc:'GREY'  -> 8bit, 1 channel
 //      video/x-raw, fourcc:'Y800'  -> 8bit, 1 channel
 //      video/x-raw, fourcc:'Y12 '  -> 12bit, 1 channel
+//      video/x-raw, fourcc:'Y16 '  -> 16bit, 1 channel
 //
 
 #define MODE_GREY   CV_FOURCC_MACRO('G','R','E','Y')
 #define MODE_Y800   CV_FOURCC_MACRO('Y','8','0','0')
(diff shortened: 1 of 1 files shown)


ArkadiuszRaj:
There is a fault in priority2, missing control if new exposure reaches maximum value.

matheusns:
The changes are correct.

alalek:
:+1:

PR #8712, opened 2017-05-10
Update cap_mjpeg_encoder.cpp

grundman:
Avoid shifting by 32bits; undefined behavior according to c - standard

### DIFF
--- a/modules/videoio/src/cap_mjpeg_encoder.cpp
+++ b/modules/videoio/src/cap_mjpeg_encoder.cpp
@@ -384,7 +384,7 @@ public:
         }
         else
         {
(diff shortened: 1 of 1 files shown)


terfendail:
It looks like shifting by 32 bits could occur only if one tries to `put` 0 bits (i.e. `len==0`). Call of the function with such parameters looks strange, so IMHO it would be better to check this situation explicitly rather than fix the consequences.

vpisarev:
yes, putting 0 bits is strange and that should be fixed, but the proposed fix provides nice extra protection. Since this is rare situation, the branch should be well-predicted and thus should not add a lot of overhead :+1:

PR #8713, opened 2017-05-10
Fix two race conditions in LazyInit and Matrix default allocator

grundman:


alalek:
1) Use CV_SINGLETON_LAZY_INIT macro, without creation of global objects (you can't guarantee initialization order of them - so `Mutex g_matAllocatorMutex;` doesn't work because it may not be initialized when you want to use them). See "static initialization order fiasco" for more info.
2) Don't change CV_SINGLETON_LAZY_INIT macro. It is well known pattern: https://en.wikipedia.org/wiki/Double-checked_locking

vpisarev:
@alalek, let me assign it to you

alalek:
@grundman Patch is rebased and updated (race condition in `Mat::getDefaultAllocator()` is resolved, other changes were removed).

alalek:
:+1:

PR #8717, opened 2017-05-11
Set LUT for Robertson weights function to zero on borders

### LINKED ISSUE #8706: HDR MergeRobertson: weights should be 0 at limits
The weights (calculated by RobertsonWeights()) used for merging the images using MergeRobertson should be 0 for input values of 0 and 255. This is also described in the article referenced in the documentation for the algorithm.

By not setting them to 0, image parts which are overexposed (and thus have values of 255) will have influence on the merged image although they should not.

sovrasov:
resolves #8706
Merge with https://github.com/opencv/opencv_extra/pull/335
Need to be merged manually because of bit patch to extra.


alalek:
:+1:

PR #8718, opened 2017-05-11
Added MatSize::dims() function

arrybn:
### This pullrequest changes
Added dims() function to MatSize struct. Enabled usage of it inside MatSize


arrybn:
wip

PR #8724, opened 2017-05-12
issue 8723: Changed doc for cv::imwrite to reflect correct default fl…

### LINKED ISSUE #8723: Doc states wrong default flag for png compression strategy for cv::imwrite

- OpenCV => :3.2:

##### Detailed description

The doc says the default png compression strategy is `IMWRITE_PNG_STRATEGY_DEFAULT` (0) when really cv:imwrite will use `IMWRITE_PNG_STRATEGY_RLE ` (3) if no flags or used or compression strategy is not specified.

##### Steps to reproduce
Code inspection around call to [`png_set_compression_strategy`](https://github.com/opencv/opencv/blob/master/modules/imgcodecs/src/grfmt_png.cpp#L407) reveals problem.

You can confirm with a debugger that `compression_strategy` is getting `3` when no compression strategy is specified.

saskatchewancatch:
…ag used for png compression strategy.

Correcting doc for cv::imwrite to note the actual flag it uses for default png compression strategy. Please see issue 8723 for context.

resolves #8723

### This pullrequest changes
The doc for cv::imwrite


berak:
could you add a 

    resolves: #8723 

line to your 1st comment, so the issue can get closed automatically, once the pr is merged ?

sovrasov:
:+1:

mshabunin:
As I can see, if one set `PNG_COMPRESSION` parameter, the strategy will be reset to `DEFAULT`: https://github.com/opencv/opencv/blob/ea911d34c60d3868debc9dc3a5dc1179d93d5baf/modules/imgcodecs/src/grfmt_png.cpp#L377-L382

saskatchewancatch:
@mshabunin That is true. But only if the compression level is explicitly set. If it is not set, then the compression level is set to constant 1 which is the minimal compression level and the compression strategy is set to RLE. With this PR, it will be clear from the docs which compression strategy will be used whether compression level is specified or not as a flag.

PR #8726, opened 2017-05-12
Enhance GigE support: Add BayerGR8 support for aravis

Neio:
Issue:
(no issue number)
Currently Aravis GigE only support mono color. We are not able to get RGB color space image.

Resolution:
Adding raw Bayer GR 8 Format support for Aravis (GigE) plugin. In this we we can convert raw Bayer GR 8 Format to RGB color space.

### DIFF
--- a/modules/videoio/src/cap_aravis.cpp
+++ b/modules/videoio/src/cap_aravis.cpp
@@ -77,11 +77,12 @@
 //      video/x-raw, fourcc:'GREY'  -> 8bit, 1 channel
 //      video/x-raw, fourcc:'Y800'  -> 8bit, 1 channel
 //      video/x-raw, fourcc:'Y12 '  -> 12bit, 1 channel
-//
+//      video/x-raw, fourcc:'BGR8'  -> 8bit, 1 channel
 
-#define MODE_GREY   CV_FOURCC_MACRO('G','R','E','Y')
-#define MODE_Y800   CV_FOURCC_MACRO('Y','8','0','0')
-#define MODE_Y12    CV_FOURCC_MACRO('Y','1','2',' ')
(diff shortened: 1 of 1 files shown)


alalek:
@ArkadiuszRaj Please take look on this patch.

ArkadiuszRaj:
This patch adds only passing new format code to Aravis library. The received image is not "debayerized' hence in my humble opinion for clarity it shall use code BYR1 and not BGR8.

From the other hand, BGR8 suggest layout of the filter array used - why not to use COLOR_BayerBG2BGR mode of cvtColor(), and enhance CvCaptureCAM_Aravis::retrieveFrame(int) to return RGB image?

Neio:
@ArkadiuszRaj I see.  I have updated the code to output RGB directly.

BTW, could you explain the meaning of code "BYR1"?

ArkadiuszRaj:
According to fourcc BYR1 andBYR2 means Bayer RAW 8 and 16 bits respectively. 

The code you provided seems to be OK. Hope you have verified conversion to RGB by your own. 

One comment. Your second commit adds plenty of unnecessary files to this pull request. @alalek what do you think on this?

alalek:
I think, it is better to "squash" commits into one (probably with "rebase" on upstream master branch) to remove unnecessary intermediate changes from git history.

Neio:
@ArkadiuszRaj Thanks for clarify. Yes, I have verified with my GigE camera (MindVision MV-GE130C-T).

@alalek I never tried. It seems like I don't need to anything but GitHub already provided an option to "squash" commits? ( https://github.com/blog/2141-squash-your-commits)

ArkadiuszRaj:
@alalek, this PR was closed by submitter but it is not merged.

PR #8731, opened 2017-05-14
CUDA median_filter.cu: remove #include precomp.hpp

### LINKED ISSUE #8730: CUDA median_filter.cu: redundant #include precomp.hpp

- OpenCV => 3.2 20170512
- Compiler => VS2017 (VS2015 for nvcc)

[+ Forum(VS2013)](http://answers.opencv.org/question/122286/vs-2013-cannot-open-include-file-tbbtbbh-no-such-file-or-directory/)

##### Detailed description

[/modules/cudafilters/src/cuda/median_filter.cu](/opencv/opencv/tree/master/modules/cudafilters/src/cuda/median_filter.cu) includes `precomp.hpp` which includes `opencv2/core/private.hpp` which includes `tbb/tbb.h`(when `WITH_TBB` is set), which is redundant and seems unclear why precompiled header is included in the first place.

And since TBB is not included in CUDA build scripts, when trying to generate `cuda_compile_generated_median_filter.cu.obj` it produces following error: `fatal error C1083: Cannot open include file: 'tbb/tbb.h': No such file or directory`

##### Steps to reproduce
Build opencv `WITH_CUDA` and `WITH_TBB`.

ylemkimon:
resolves #8730

### This pullrequest changes

Removed redundant precompiled header include and using-namespaces from `modules/cudafilters/src/cuda/median_filter.cu`

### DIFF
--- a/modules/cudafilters/src/cuda/median_filter.cu
+++ b/modules/cudafilters/src/cuda/median_filter.cu
@@ -42,11 +42,6 @@
 
 #if !defined CUDA_DISABLER
 
-#include "precomp.hpp"
-
-using namespace cv;
-using namespace cv::cuda;
-
 #include "opencv2/core/cuda/common.hpp"
 #include "opencv2/core/cuda/vec_traits.hpp"
 #include "opencv2/core/cuda/vec_math.hpp"


alalek:
`precomp.hpp` files have special purpose. They should not be removed.

ylemkimon:
@alalek But in this specific case, it seems redundant and other CUDA objects(*.cu) doesn't include it.

It builds successfully but not tested yet.

ylemkimon:
@alalek Code below seems to work fine.

'''.cpp
#include <iostream>

#include "opencv2/core.hpp"
#include "opencv2/video.hpp"
#include "opencv2/cudafilters.hpp"
#include "opencv2/cudaimgproc.hpp"

using namespace std;
using namespace cv;
using namespace cv::cuda;

int main(int argc, const char** argv)
{
	VideoCapture cap;
	cap.open(0);
	if (!cap.isOpened()) {
		cerr << "can not open camera" << endl;
		return -1;
	}

	Mat frame, dst;
	cap >> frame;

	GpuMat d_frame(frame), d_src, d_dst;
	cuda::cvtColor(d_frame, d_src, CV_RGB2GRAY);

	Ptr<Filter> median = createMedianFilter(d_src.type(), 10);
	median->apply(d_src, d_dst);

	namedWindow("grayscale", WINDOW_NORMAL);
	namedWindow("filtered", WINDOW_NORMAL);

	for (;;)
	{
		cap >> frame;
		if (frame.empty())
			break;
		d_frame.upload(frame);

		cuda::cvtColor(d_frame, d_src, CV_RGB2GRAY);
		median->apply(d_src, d_dst);

		d_dst.download(dst);

		imshow("grayscale", frame);
		imshow("filtered", dst);

		char key = (char)waitKey(30);
		if (key == 27)
			break;
	}

	return 0;
}
'''

vpisarev:
I think, we can accept the patch since it does not affect the main build with CUDA turned off :+1:

PR #8735, opened 2017-05-15
Fix icvJSONParseKey function for case big JSON file with only one line inside

vbystricky:
If we try to load big JSON file then our internal buffer can split keyname. We have problem with loading JSON file in these cases.

### This pullrequest changes 

icvJSONParseKey function - Try to load next part of the file into internal buffer if we obtain end of buffer when parse key name.

### DIFF
--- a/modules/core/src/persistence.cpp
+++ b/modules/core/src/persistence.cpp
@@ -3373,6 +3373,27 @@ static char* icvJSONParseKey( CvFileStorage* fs, char* ptr, CvFileNode* map, CvF
     do ++ptr;
     while( cv_isprint(*ptr) && *ptr != '"' );
 
+    if (*ptr == 0)
+    {
+        if (fs->buffer_start < beg)
+        {
+            int sz = static_cast<int>(ptr - beg);
+            memcpy(fs->buffer_start, beg, sz);
+            beg = fs->buffer_start;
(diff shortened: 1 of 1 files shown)


sovrasov:
Could you add a test for the solved issue?

vbystricky:
Something like this:

```
#include "opencv2/core.hpp"

#include <stdio.h>
#include <fstream>

int main(int argc, const char** argv)
{
    static std::string temp_path = cv::tempfile("test.json");
    {
        std::ofstream ofs(temp_path);
        ofs << "{     ";
        for (int i = 0; i < 52500; i++)
        {
            std::string str = cv::format("\"KEY%05d\" : \"VALUE\", ", i);
            ofs << str;
        }
        ofs << "}";
    }

    {
        cv::FileStorage fs(temp_path, cv::FileStorage::READ | cv::FileStorage::FORMAT_JSON);
    }

    remove(temp_path.c_str());
    return 0;
}
```

sovrasov:
Yes, but in-memory file is preferable in such tests. `FileStorage` has a special mode for this. [Example of in-memory FileStorage mode usage].(https://github.com/opencv/opencv/blob/master/modules/core/test/test_io.cpp#L1084)
The test can be placed [here](https://github.com/opencv/opencv/blob/master/modules/core/test/test_io.cpp#L1238).

PR #8739, opened 2017-05-16
Fix for accumulate performance test in case of multiple iterations

terfendail:
### This pullrequest changes

Reset dst on each accumulate performance test iteration in order to retain the same result for sanity


vpisarev:
:+1:

alalek:
I believe it is better to drop sanity check here (via `SANITY_CHECK_NOTHING`), instead of measuring of something, but definitely not "accumulate" call.

PR #8741, opened 2017-05-17
A new feature matching algorithm is added

JiawangBian:
I add a feature matching algorithm "GMS: Grid-based Motion Statistics for Fast, Ultra-robust Feature Correspondence IEEE, CVPR 2017"  to the feature2d module.

The project page of the paper is: [ProjectPage](http://jwbian.net/gms )
The source code is on the [Github](https://github.com/JiawangBian/GMS-Feature-Matcher)

### DIFF
--- a/modules/features2d/include/opencv2/features2d.hpp
+++ b/modules/features2d/include/opencv2/features2d.hpp
@@ -1103,6 +1103,28 @@ protected:
     int addedDescCount;
 };
 
+
+/** @brief GMS feature matching strategy
+
+@param size1 Input size of image1.
+@param size2 Input size of image2.
+@param keypoints1 Input keypoints of image1.
+@param keypoints2 Input keypoints of image2.
+@param matches_all Input neighborhood matches.
+@param InlierMask Output Mask. 1 for inlier, 0 for outlier
+@return the number of inliers
+
+@note 
+	Since GMS works well when features number is large, we recommend that you use ORB feature and set FastThreshold to 0 to get as many as possible features quickly.
+	If matching results are not satisfied, please add more features.(We use 10000 for images with 640 X 480).
+	If your images have big rotation and scale changes, please set rotation or scale true.
+ */
+ 
+CV_EXPORTS int gms_matching( const cv::Size size1, const cv::Size size2, const std::vector<KeyPoint>& keypoints1, const std::vector<KeyPoint>& keypoints2,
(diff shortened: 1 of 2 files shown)


alalek:
Hi,
Thank you for contribution!

1) Please provide tests. There are some tests for feature matchers, but required API is different.
2) In general, completely new algorithms should go into opencv_contrib if they are not highly integrated into existed code.
3) OpenCV uses BSD license for code. What is the license of this code?

JiawangBian:
Just now added a BSD license for the code. 
Will put the code into opencv_contrib and provide tests. thanks@alalek

alalek:
There are [entry-points](https://github.com/opencv/opencv/blob/3.2.0/modules/features2d/test/test_matchers_algorithmic.cpp#L532-L544) for existed feature matching tests.
Similar code (need to check consistency) is existed in `xfeatures2d` tests of opencv_contrib.
But we need to wrap this algorithm into `DescriptorMatcher` class to use them.

JiawangBian:
Those codes may be used to check the nearest neighbor matching. This algorithm regards nearest neighbor matches as input and labels them as inlier or outlier, so we may write another test code. Should I make a new module called "featurematching" to put it?

alalek:
> This algorithm regards nearest neighbor matches as input and labels them as inlier or outlier

> `matches_all` Input neighborhood matches.

OK, this is not classical feature matcher algorithm.
So existed tests for matchers are no applicable here, we need some other tests to check code.

> make a new module

New module is not needed. I believe we can put this into "xfeatures2d" module or `ccalib3d` (if it is similar to `findHomography` function).

JiawangBian:
I think this is a good position: "https://github.com/opencv/opencv/blob/master/modules/stitching/src/matchers.cpp"
Because I find ratio-test(BestOf2NearestMatcher) here and the new algorithm is an alternative to ratio-test.

Can I put the code here?

catree:
@JiawangBian If you are ok, I would like to port this pull request into the [xfeatures2d module](https://github.com/opencv/opencv_contrib/tree/master/modules/xfeatures2d). 
This way, the algorithm will be integrated in OpenCV contrib to be usable and later can be moved or refactored if needed.

JiawangBian:
OK，thanks very much。and i will keep updating if needed.

catree:
@JiawangBian For your information, I have created the corresponding pull request in opencv/opencv_contrib#1532.

PR #6447, opened 2016-04-21
add feature to convert FP32(float) to FP16(half) using HW instructions

tomoaki0705:
### What does this PR change?
Allow user to pack 32 bit float data in to 16bit float data (half)
  * check compiler support
  * check HW support before executing
  * add test doing round trip conversion from / to FP32
  * treat array correctly if size is not multiple of 4
  * add declaration to prevent warning

PR #8743, opened 2017-05-17
core: add OpenCL version of convertFp16 and test

tomoaki0705:
### This pullrequest changes
UMat version of convertFp16. This PR is  a follow up PR of #6447
Now all CPU (SW/SIMD), CUDA and OpenCL version is available.

### DIFF
--- a/modules/core/src/convert.cpp
+++ b/modules/core/src/convert.cpp
@@ -5371,6 +5371,34 @@ static bool ocl_convertScaleAbs( InputArray _src, OutputArray _dst, double alpha
     return k.run(2, globalsize, NULL, false);
 }
 
+static bool ocl_convertFp16( InputArray _src, OutputArray _dst, int ddepth )
+{
+    int type = _src.type(), cn = CV_MAT_CN(type);
+
+    _dst.createSameSize( _src, CV_MAKETYPE(ddepth, cn) );
+    int kercn = 1;
+    int rowsPerWI = 1;
+    String build_opt = format("-D HALF_SUPPORT -D dstT=%s -D srcT=%s -D rowsPerWI=%d%s",
+                           ddepth == CV_16S ? "half" : "float",
+                           ddepth == CV_16S ? "float" : "half",
+                           rowsPerWI,
+                           ddepth == CV_16S ? " -D FLOAT_TO_HALF " : "");
+    ocl::Kernel k("convertFp16", ocl::core::halfconvert_oclsrc, build_opt);
+    if (k.empty())
+        return false;
+
+    UMat src = _src.getUMat();
+    UMat dst = _dst.getUMat();
+
+    ocl::KernelArg srcarg = ocl::KernelArg::ReadOnlyNoSize(src),
+            dstarg = ocl::KernelArg::WriteOnly(dst, cn, kercn);
+
+    k.args(srcarg, dstarg);
+
+    size_t globalsize[2] = { (size_t)src.cols * cn / kercn, ((size_t)src.rows + rowsPerWI - 1) / rowsPerWI };
+    return k.run(2, globalsize, NULL, false);
+}
+
 #endif
 
 }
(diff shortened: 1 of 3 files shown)


modules/core/src/convert.cpp
```diff
@@ -5371,6 +5371,52 @@ static bool ocl_convertScaleAbs( InputArray _src, OutputArray _dst, double alpha
     return k.run(2, globalsize, NULL, false);
 }
 
+static bool ocl_convertFp16( InputArray _src, OutputArray _dst )
+{
+    int type = _src.type(), depth = CV_MAT_DEPTH(type), cn = CV_MAT_CN(type);
+
+    int nSChannels = _src.getMat().channels();
```
alalek: `.getMat()` is expensive operation (here and below) - it maps (or downloads) OpenCL buffer into CPU memory.
There is `cn` variable above for this value.

`_dst.create()` -> `_dst.createSameSize()`, so dims is not needed.

modules/core/src/convert.cpp
```diff
@@ -5371,6 +5371,52 @@ static bool ocl_convertScaleAbs( InputArray _src, OutputArray _dst, double alpha
...
+    switch( depth )
+    {
+    case CV_32F:
+        ddepth = CV_16S;
+        break;
+    case CV_16S:
+        ddepth = CV_32F;
+        break;
+    default:
+        CV_Error(Error::StsUnsupportedFormat, "Unsupported input depth");
```
alalek: Probably this check should be before `CV_OCL_RUN` call.
Arguments check code is common in general, so it should not be duplicated between "implementations".

modules/core/src/convert.cpp
```diff
@@ -5371,6 +5371,52 @@ static bool ocl_convertScaleAbs( InputArray _src, OutputArray _dst, double alpha
...
+                           ddepth == CV_16S ? "half" : "float",
+                           ddepth == CV_16S ? "float" : "half",
+                           rowsPerWI,
+                           ddepth == CV_16S ? " -D FLOAT_TO_HALF " : "");
+    ocl::Kernel k("convertFp16", ocl::core::halfconvert_oclsrc, build_opt);
+    if (k.empty())
+        return false;
+
+    UMat src = _src.getUMat();
+    UMat dst = _dst.getUMat();
```
alalek: `_src.getUMat(ACCESS_READ)`
`_dst.getUMat(ACCESS_WRITE)`

alalek:
@tomoaki0705 Well done! Please address comments above.

modules/core/src/convert.cpp
tomoaki0705: Thank you.  I agree for this modification

modules/core/src/convert.cpp
tomoaki0705: Thank you.  I'll put this in before.

modules/core/src/convert.cpp
tomoaki0705: I'm not sure about this point.
The ```ReadOnlyNoSize``` and ```WriteOnly``` function follows after here, so I'm not sure if ```ACCESS_READ``` and ```ACCESS_WRITE``` is necessary here.
If ```_src``` or ```_dst``` is already ```UMat```, this ```getUMat(ACCESS_READ)``` would try to grab a row of the image out of the range.

I believe it's OK to use the default ```-1``` for this function.
How is that ?

tomoaki0705:
@alalek , thank you for your review ! 
The implementation become much mature now.

About the third comment, I'm keep on facing run time error, so I'm not sure about the ```ACCESS_READ``` and ```ACCESS_WRITE```
Could you let me know your opinion, please ?

Thanks !

modules/core/src/convert.cpp
alalek: Right, my fail.
These methods doesn't have `ACCESS` parameter, there is `idx` only.

tomoaki0705: I appreciate your comment.

tomoaki0705:
I squashed the commits to one.
I'm happy to update this PR if anything is wrong.
Best.

alalek:
Thank you! :+1:

PR #8750, opened 2017-05-19
[photo] CalibrateDebevec fix and updates

amroamroamro:
The code had `3` hard-coded as number of channels when accessing image pixels. I changed it to use `img.channels()`, that way it works even for grayscale images. This is the first commit.

The other two commits are minor updates (fix typos, add comments, add extra checks, etc.)

Note: I checked the code against the reference paper, which includes a MATLAB implementation in the appendix: http://www.pauldebevec.com/Research/HDR/debevec-siggraph97.pdf

### DIFF
--- a/modules/photo/src/calibrate.cpp
+++ b/modules/photo/src/calibrate.cpp
@@ -43,7 +43,6 @@
 #include "precomp.hpp"
 #include "opencv2/photo.hpp"
 #include "opencv2/imgproc.hpp"
-//#include "opencv2/highgui.hpp"
 #include "hdr_common.hpp"
 
 namespace cv
@@ -57,7 +56,7 @@ public:
         samples(_samples),
         lambda(_lambda),
         random(_random),
-        w(tringleWeights())
+        w(triangleWeights())
     {
     }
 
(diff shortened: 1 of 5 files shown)


modules/photo/src/calibrate.cpp
```diff
@@ -57,77 +56,104 @@ class CalibrateDebevecImpl : public CalibrateDebevec
...
 
+        // check inputs
         std::vector<Mat> images;
         src.getMatVector(images);
         Mat times = _times.getMat();
 
         CV_Assert(images.size() == times.total());
         checkImageDimensions(images);
         CV_Assert(images[0].depth() == CV_8U);
+        CV_Assert(times.depth() == CV_32F);
```
sovrasov: It would be nice to check number of channels here.

amroamroamro: yep, I'll change it to `CV_Assert(times.type() == CV_32FC1);`

sovrasov: Sorry, actually I meant number of channels in input image.

amroamroamro: Oh there's no need. Although I expect images to be 1 or 3 channels, the code should work for any number of channels (each processed separately).
The output will be 256x1 matrix with the same number of channels as the input images.

sovrasov:
@amroamroamro squash commits into one before merge.

amroamroamro:
@sovrasov done

sovrasov:
:+1:

PR #8753, opened 2017-05-19
Fix cpu features detection on android

RyuheiMori:
/proc/self/auxv will fail from regular application processes on some Android platform versions.

use cpufeatures library provided by ndk instead.

### DIFF
--- a/3rdparty/cpufeatures/CMakeLists.txt
+++ b/3rdparty/cpufeatures/CMakeLists.txt
@@ -1,0 +1,29 @@
+
+ocv_include_directories(${CMAKE_CURRENT_SOURCE_DIR})
+file(GLOB cpuf_s *.c)
+file(GLOB cpuf_h *.h)
+
+set(lib_srcs  ${cpuf_s})
+set(lib_hdrs  ${cpuf_h})
+
+set(CPUFEATURES_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR} CACHE INTERNAL "")
+set(CPUFEATURES_LIBRARIES cpufeatures  CACHE INTERNAL "")
+
+add_library(cpufeatures STATIC ${lib_srcs} ${lib_hdrs})
+
+set_target_properties(cpufeatures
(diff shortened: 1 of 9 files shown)


alalek:
What Android NDK do you use to extract `cpufeatures` sources?

RyuheiMori:
r15 Beta 2

RyuheiMori:
This issue happens when running 32bit binary on 64 bit platform. I have confirmed the issue with sony z5 (android 7.0)/ samsung s6 /s7 (android 6.0).

https://android.googlesource.com/platform/ndk/+/882ac8f3392858991a0e1af33b4b7387ec856bd2

https://github.com/opencv/opencv/blob/master/modules/core/src/system.cpp#L490
During initialization, opencv will check whether baseline features exist or not.
If not, it abort immediately.

alalek:
@RyuheiMori Hi! I'm sorry for huge delay.

Unfortunately, `cpufeatures` detection uses `dlopen` call.
Currently OpenCV features detection code is processed at library initialization stage (under other `dlopen` call). This pattern is not supported well: recursive dlopen calls hangs on some devices (It hangs on Android 2.2.3, but Android 4.4.3 is fine).

Could you try to remove dlopen call and try again on your device:
```.diff
-void* libc_handle = dlopen("libc.so", RTLD_NOW);
+void* libc_handle = 0;
```

Does other fallback detections work for you?

alalek:
@RyuheiMori Thank you for your effort.

In general, I prefer "cpufeatures" way instead of self-written code.
I have updated library integration code and added hang workaround to cpufeatures.
Could you check this code: https://github.com/opencv/opencv/compare/master...alalek:pr8753 ?
You can fetch my updates via these commands:
```
git fetch https://github.com/alalek/opencv pr8753:alalek_pr8753
git checkout alalek_pr8753
```

RyuheiMori:
OK, it works on my devices

alalek:
@RyuheiMori Thanks for check!

I have updated this patch.
Your last changes are saved here: https://github.com/alalek/opencv/commits/pr8753_origin

PR #8754, opened 2017-05-20
fix py_svm_opencv sample

berak:
* fix some [code errors](http://answers.opencv.org/question/150063/python-svm-function-svmpredicttestdata-returns-tuple-instead-of-numpy-array/) (train responses must be integer, predict() returns a result/responses tuple)
* move code to samples/python and use @snippet and @include to reference it

### DIFF
--- a/samples/python/tutorial_code/ml/py_svm_opencv/hogsvm.py
+++ b/samples/python/tutorial_code/ml/py_svm_opencv/hogsvm.py
@@ -1,0 +1,71 @@
+import cv2
+import numpy as np
+
+SZ=20
+bin_n = 16 # Number of bins
+
+
+affine_flags = cv2.WARP_INVERSE_MAP|cv2.INTER_LINEAR
+
+## [deskew]
+def deskew(img):
+    m = cv2.moments(img)
+    if abs(m['mu02']) < 1e-2:
+        return img.copy()
+    skew = m['mu11']/m['mu02']
+    M = np.float32([[1, skew, -0.5*SZ*skew], [0, 1, 0]])
+    img = cv2.warpAffine(img,M,(SZ, SZ),flags=affine_flags)
+    return img
+## [deskew]
+
+## [hog]
+def hog(img):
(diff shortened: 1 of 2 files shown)


samples/python/tutorial_code/ml/py_svm_opencv/hogsvm.py
```diff
@@ -0,0 +1,68 @@
...
+    mag, ang = cv2.cartToPolar(gx, gy)
+    bins = np.int32(bin_n*ang/(2*np.pi))    # quantizing binvalues in (0...16)
+    bin_cells = bins[:10,:10], bins[10:,:10], bins[:10,10:], bins[10:,10:]
+    mag_cells = mag[:10,:10], mag[10:,:10], mag[:10,10:], mag[10:,10:]
+    hists = [np.bincount(b.ravel(), m.ravel(), bin_n) for b, m in zip(bin_cells, mag_cells)]
+    hist = np.hstack(hists)     # hist is a 64 bit vector
+    return hist
+## [hog]
+
+img = cv2.imread('digits.png',0)
```
berak: should it point to the data folder ?

sovrasov: It's a usual problem for samples. We have no mechanism to provide a path to the  `opencv/samples/data` folder. You can leave it as is, but at least help message in case of the bad image path should be printed. Also user can provide path to data via command line.

sovrasov:
:+1:

samples/python/tutorial_code/ml/py_svm_opencv/hogsvm.py
mshabunin: Most of the samples use `../data/` path by default (for example [this one](https://github.com/opencv/opencv/blob/master/samples/cpp/demhist.cpp#L67), [another one](https://github.com/opencv/opencv/blob/master/samples/python/calibrate.py#L38)).

PR #8757, opened 2017-05-20
correct bug in fastmeans (ref #7899)

### LINKED ISSUE #7899: wrong doc for fast means
http://docs.opencv.org/2.4.10/modules/photo/doc/denoising.html?highlight=denoise#fastnlmeansdenoisingcolored

The parameter is listed as hColor in the signature, but later referenced as hForColorComponents.  Also enought is mispeeled and should be enough.

I didn't check other versions of the doc to see if there was the same problem.

krishraghuram:
resolves #7899 

### This pullrequest changes

Corrected hForColorComponents to hColor in fastNlMeansDenoisingColored and fastNlMeansDenoisingColoredMulti


mshabunin:
:+1:

Issue #7197, opened 2016-08-30
T-API introduction has not been added to master


mshabunin:
##### Detailed description

There is no page in master documentation with general information about OpenCL (T-API). Corresponding 2.4 page: http://docs.opencv.org/2.4/modules/ocl/doc/introduction.html

Some data structures and utility functions can be documented too using information from 2.4 docs: http://docs.opencv.org/2.4/modules/ocl/doc/ocl.html

ankitaggarwal011:
@mshabunin, I would like to start contributing to OpenCV with this issue. As I understand, (OpenCL module)[https://github.com/opencv/opencv/tree/master/modules/core/include/opencv2/core] is part of core module in OpenCV 3.0+, compared to being a (separate module)[https://github.com/opencv/opencv/tree/2.4/modules/ocl] in OpenCV 2.4+ and there might be certain changes in its T-API going from 2.4 to 3.0.

Thus the corresponding documentation for OpenCL module in OpenCV 3.0+ has to be changed accordingly while using information from 2.4 documentation. Now, I've not worked a lot on the OpenCL module myself. Please let me know if this is right way to approach this issue. Thanks.

mshabunin:
@ankitaggarwal011 , yes it should become a part of _core_ module documentation (http://docs.opencv.org/master/dc/d83/group__core__opencl.html) and should introduce T-API to users. You can use information related to T-API from http://opencv.org/platforms/opencl.html and some parts from http://docs.opencv.org/2.4/modules/ocl/doc/introduction.html.

If introduction will be small (3-5 sections of text), then it can be placed to the _Detailed Description_ part (in [this](https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/ocl.hpp) file), otherwise separate page should be created (in [this](https://github.com/opencv/opencv/tree/master/modules/core/doc) folder) and referred from the description. You can insert code snippets (use `@snippet` doxygen command) from cpp code samples and corresponding tapi samples to illustrate differences.
As I can see, some functions are actually documented but with wrong formatting:
```.cpp
/*
//! @brief Convert OpenCL image2d_t to UMat
//
//! @note Note:
//   OpenCL image2d_t (cl_mem_image), should be compatible with OpenCV
//   UMat formats.
//   Memory content is copied from image to UMat with
//   clEnqueueCopyImageToBuffer function.
//
// Params:
//! @param  cl_mem_image - source image2d_t handle
//! @param  dst          - destination UMat
*/
[...]

PR #8758, opened 2017-05-20
Correct the existing documented T-API functions to match the doxygen format

krishraghuram:
resolves part of #7197

### This pullrequest changes

Corrected the existing documented T-API functions to match the doxygen format

### DIFF
--- a/modules/core/include/opencv2/core/ocl.hpp
+++ b/modules/core/include/opencv2/core/ocl.hpp
@@ -278,46 +278,38 @@ protected:
     Impl* p;
 };
 
-/**
-@brief Attaches OpenCL context to OpenCV
+/** @brief Attaches OpenCL context to OpenCV
 @note
-  OpenCV will check if available OpenCL platform has platformName name,
-  then assign context to OpenCV and call clRetainContext function.
-  The deviceID device will be used as target device and new command queue
-  will be created.
-@param platformName - name of OpenCL platform to attach,
-                     this string is used to check if platform is available
-                     to OpenCV at runtime
-@param platformID   - ID of platform attached context was created for
-@param context      - OpenCL context to be attached to OpenCV
-@param deviceID     - ID of device, must be created from attached context
+  OpenCV will check if available OpenCL platform has platformName name, then assign context to
+  OpenCV and call `clRetainContext` function. The deviceID device will be used as target device and
+  new command queue will be created.
+@param platformName name of OpenCL platform to attach, this string is used to check if platform is available to OpenCV at runtime
+@param platformID ID of platform attached context was created for
+@param context OpenCL context to be attached to OpenCV
+@param deviceID ID of device, must be created from attached context
 */
 CV_EXPORTS void attachContext(const String& platformName, void* platformID, void* context, void* deviceID);
 
-/**
-@brief Convert OpenCL buffer to UMat
+/** @brief Convert OpenCL buffer to UMat
 @note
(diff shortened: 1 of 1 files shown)


modules/core/include/opencv2/core/ocl.hpp
```diff
@@ -278,55 +278,52 @@ class CV_EXPORTS Platform
     Impl* p;
 };
 
-/*
-//! @brief Attaches OpenCL context to OpenCV
+/**
+// @brief Attaches OpenCL context to OpenCV
```
mshabunin: There is no need in `//` at the beginning of each line of multi-line comment, please use one of the following formats:
```.cpp
/**
  .....
  .....
  .....
*/
```
or
```.cpp
//! ...
//! ...
//! ...
//! ...
```
Usually we use the first variant: [example](http://docs.opencv.org/master/d4/db1/tutorial_documentation.html#tutorial_documentation_quick_start_2).

alalek: Full list of doxygen supported comments is here (for reference): https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html

modules/core/include/opencv2/core/ocl.hpp
```diff
@@ -569,7 +560,7 @@ class CV_EXPORTS Kernel
         i = set(i, a6); i = set(i, a7); i = set(i, a8); i = set(i, a9); i = set(i, a10); i = set(i, a11);
         i = set(i, a12); i = set(i, a13); i = set(i, a14); set(i, a15); return *this;
     }
-    /*
+    /**
     Run the OpenCL kernel.
```
mshabunin: This line can be used as a brief description (add `@brief` at the beginning).

modules/core/include/opencv2/core/ocl.hpp
```diff
@@ -278,55 +278,46 @@ class CV_EXPORTS Platform
...
+@brief Attaches OpenCL context to OpenCV
+@note
+  OpenCV will check if available OpenCL platform has platformName name,
+  then assign context to OpenCV and call clRetainContext function.
+  The deviceID device will be used as target device and new command queue
+  will be created.
+@param platformName - name of OpenCL platform to attach,
+                     this string is used to check if platform is available
+                     to OpenCV at runtime
+@param platformID   - ID of platform attached context was created for
```
mshabunin: It is better to remove `-` symbol and manual alignment in all parameter descriptions:
```.cpp
/**
...
@param platformName name of OpenCL platform to attach, this string is used to check if platform is available to OpenCV at runtime
@param platformID ID of platform attached context was created for
@param context OpenCL context to be attached to OpenCV
@param deviceID ID of device, must be created from attached context
...
*/
```

modules/core/include/opencv2/core/ocl.hpp
krishraghuram: I did notice the extra ```//!``` in the comments, but left it as is, assuming its a opencv coding style.
@mshabunin Thanks for the [example](http://docs.opencv.org/master/d4/db1/tutorial_documentation.html#tutorial_documentation_quick_start_2)

krishraghuram:
Dumb question : 
why does it say "krishraghuram committed with sovrasov" in the first commit of this PR?

mshabunin:
@krishraghuram , because @sovrasov has squashed your commits into one and this new commit has you recorded as the author and him as committer. You can also note that we've added two more commits to your branch with minor fixes, so now this PR is ready to be merged.
:+1:

PR #8769, opened 2017-05-22
core: issues found by static analysis

mshabunin:
### This pullrequest changes

Added definitions, checks and assertions to help analysis tools.

PR #8771, opened 2017-05-22
build: fix snprintf() usage

alalek:


vpisarev:
:+1:

PkLab:
It compile but I'm not sure this will works correctly

### Collect some docs 
#### `vsnprintf` is ANSI C++99
on [cppreference](http://en.cppreference.com/w/c/io/vfprintf)

> in case of truncation it returns the total number of characters (not including the terminating null-byte) which would have been written, if the limit was not imposed. Negative value if an error occurred


on VS2013 & VS2015

> return the number of characters written if the number of characters to write is less than or equal to count; if the number of characters to write is greater than count, these functions return -1 indicating that output has been truncated. The return value does not include the terminating null, if one is written.


on [VS2013](https://msdn.microsoft.com/en-us/library/d3xd30zz(v=vs.120).aspx)

> If there is room at the end (that is, if the number of characters to write is less than count), the buffer will be null-terminated.


on [VS2015](https://msdn.microsoft.com/en-us/library/d3xd30zz.aspx)

> The `vsnprintf` function always writes a null terminator...

#### `vsnprintf_s` is ANSI C++11
on cppreference

> returns number of characters not including the terminating null character (which is always written as long as buffer is not a null pointer and bufsz is not zero and not greater than RSIZE_MAX), which would have been written to buffer if bufsz was ignored, or a negative value if a runtime constraints violation or an encoding error occurred


#### `_vsnprintf_s` is MSVC specific maybe be C++11

on VS2013 & VS2015

> return the number of characters written, not including the terminating null,... unless count is _TRUNCATE, in which case as much of the string as will fit in buffer is written and -1 returned


on [VS2013](https://msdn.microsoft.com/en-us/library/d3xd30zz(v=vs.120).aspx) 

> If count is _TRUNCATE, then these functions write as much of the string as will fit in buffer while leaving room for a terminating null.

on VS2015(https://msdn.microsoft.com/en-us/library/d3xd30zz.aspx)

> writes up to count characters of the given data to the memory pointed to by buffer and appends a terminating null.

### Summary
- Return value: On MSVC number of characters written or -1 in case of truncation or error
- Return value: On cppreference returns the total number of characters which would have been written or -1 in case of error 

- `vsnprintf` works as expected on cppreference and VS2015... always writes a null terminator. On VS2013 the buffer will be null-terminated If there is room at the end

### My proposl
Use `vsnprintf` witch is ANSI C++99. On MSVC check for -1 and ensure null terminating on VS2013
```.cpp
static bool cv_snprintf(char* buf, int len, const char* fmt, ...)
{
    va_list va;
    va_start(va, fmt);
    int res = vsnprintf((char *)buf, len, fmt, va);
    va_end(va);
#if defined _MSC_VER
    // maybe truncation maybe error 
    if(res < 0)
        check for last errno 
    res = len -1;
    // ensure null terminating on VS2013	
#if def_MSC_VER<=1800
    buf[res] = 0; 
#endif
#endif
    return res >= 0 && res < len;
}
```

PkLab:
Still merging `snprintf` #8769... sure It was only a small oversight.
btw What about the final solution ?

PR #8772, opened 2017-05-22
Update drawing.cpp

grundman:
Init struct to zero to prevent uninitialized reads


sovrasov:
@grundman C++11 is disabled in OpenCV by default. BTW do you observe any uninitialized reads? It seems like all the fields are filled in [these](https://github.com/opencv/opencv/blob/45f0cf0d41170398d44612f69b574bae23f57e2f/modules/imgproc/src/drawing.cpp#L1168) lines.

wjkoh:
Thanks for the clarification, @sovrasov. We didn't pull this fix (https://github.com/opencv/opencv/commit/fa36b9d3455d62e98554a996dea3ebf7201109c2) and encountered uninitialized reads.

PR #8773, opened 2017-05-23
Fix frame timestamp in VideoCapture::get

### LINKED ISSUE #8763: VideoCapture Frame Timestamp

- OpenCV => 3.2

##### Detailed description

Hello,

I used [VideoCapture::get](http://docs.opencv.org/2.4/modules/highgui/doc/reading_and_writing_images_and_video.html#videocapture-get) with CV_CAP_PROP_POS_MSEC property identifier to read the timestamp of the last retrieved frame, but the return value was zero. There is a bug in [cap_v4l.cpp](https://github.com/opencv/opencv/blob/3.2.0/modules/videoio/src/cap_v4l.cpp) file the causes this problem. The value of the timestamp is copied from a buffer in [this line](https://github.com/opencv/opencv/blob/3.2.0/modules/videoio/src/cap_v4l.cpp#L868), but the the data in the buffer in not valid because we have already called VIDIOC_QBUF ioctl [here](https://github.com/opencv/opencv/blob/3.2.0/modules/videoio/src/cap_v4l.cpp#L864). The solution is to flip the two lines of the code, so we read the timestamp before calling VIDIOC_QBUF ioctl.

Thank You,
Ahmad

kvaghel1:
resolves #8763

### DIFF
--- a/modules/videoio/src/cap_v4l.cpp
+++ b/modules/videoio/src/cap_v4l.cpp
@@ -861,11 +861,11 @@ static int read_frame_v4l2(CvCaptureCAM_V4L* capture) {
    //printf("got data in buff %d, len=%d, flags=0x%X, seq=%d, used=%d)\n",
    //	  buf.index, buf.length, buf.flags, buf.sequence, buf.bytesused);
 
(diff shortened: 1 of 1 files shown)


alalek:
Thank you! :+1:

PR #8776, opened 2017-05-23
photo: add support of all types of 1c images in inpaint

### LINKED ISSUE #4542: Inpainting support for float32 data

sovrasov:
replaces #8674 
resolves #4542

### DIFF
--- a/modules/photo/include/opencv2/photo.hpp
+++ b/modules/photo/include/opencv2/photo.hpp
@@ -89,7 +89,7 @@ enum
 
 /** @brief Restores the selected region in an image using the region neighborhood.
 
-@param src Input any 1-channel or 8-bit 3-channel image.
+@param src Input 8-bit, 16-bit unsigned or 32-bit float 1-channel or 8-bit 3-channel image.
 @param inpaintMask Inpainting mask, 8-bit 1-channel image. Non-zero pixels indicate the area that
 needs to be inpainted.
 @param dst Output image with the same size and type as src .
--- a/modules/photo/include/opencv2/photo.hpp
+++ b/modules/photo/include/opencv2/photo.hpp
@@ -89,7 +89,7 @@ enum
 
 /** @brief Restores the selected region in an image using the region neighborhood.
 
-@param src Input 8-bit 1-channel or 3-channel image.
+@param src Input any 1-channel or 8-bit 3-channel image.
 @param inpaintMask Inpainting mask, 8-bit 1-channel image. Non-zero pixels indicate the area that
 needs to be inpainted.
 @param dst Output image with the same size and type as src .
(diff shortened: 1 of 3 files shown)


vpisarev:
@sovrasov, inpaint is not a basic nor very small function; such aggressive template expansion increases the binary size, about which we already get quite a few complains. I suggest to retain 8u, 16u and 32f; and remove everything else

sovrasov:
@vpisarev done.

vpisarev:
thanks! :+1:

PR #8778, opened 2017-05-23
V4L: do not return frame if it can not be retrieved

mshabunin:
rebased and modified #6853

### This pullrequest changes

Do not return frame if `ioctl(VIDIOC_DQBUF)` returned error.


vpisarev:
:+1:

Issue #7792, opened 2016-12-05
Missing ASSERT / bug in arithm.cpp


ArkadiuszRaj:
##### System information (version)
- OpenCV => 3.1 ( 380cc52)
- Operating System / Platform => Linux 64 Bit
- Compiler => gdb

##### Detailed description

Missing ASSERT /  bug in arithm.cpp

> size_t esz = src1.elemSize();
> size_t blocksize0 = (size_t)(BLOCK_SIZE + esz-1)/esz;

allows division by zero...

```
Thread 1 "pfnmc" received signal SIGFPE, Arithmetic exception.
0x00007ffff435146e in cv::compare (_src1=..., _src2=..., _dst=..., op=1) at /home/op/opencv/modules/core/src/arithm.cpp:1280
1280	    size_t blocksize0 = (size_t)(BLOCK_SIZE + esz-1)/esz;
(gdb) bt
#0  0x00007ffff435146e in cv::compare (_src1=..., _src2=..., _dst=..., op=1) at /home/op/opencv/modules/core/src/arithm.cpp:1280
#1  0x00007ffff44dd4fc in cv::MatOp_Cmp::assign (this=0x7ffff4fb8000 <cv::g_MatOp_Cmp>, e=..., m=..., _type=-1) at /home/op/opencv/modules/core/src/matop.cpp:1456
```

Excerpt from the code causing this behavior
```
Mat temp;
subtract(src(interior), mean[0], temp);
meanStdDev(temp, mean, stddev);
mask = temp > 5*mean[0];
```

alalek:
`elemSize()` should be non-zero. Please provide complete minimal reproducer of this problem.

ArkadiuszRaj:
That is complete code.

Before SIGPFE that is the state of the variables

*src* Mat content
```
[ 51,  50,  52,  51,  54,  58,  48,  53,  47,  66,  50,  46,  37,  44,  28,  20;
  52,  54,  48,  54,  50,  55,  56,  60,  47,  50,  50,  51,  43,  24,  28,  16;
  50,  55,  42,  49,  49,  54,  50,  56,  49,  37,  46,  39,  28,  22,  12,   7;
  53,  55,  53,  55,  54,  54,  50,  40,  48,  48,  45,  28,  21,  16,   8,   7]
```
*interior* rect definition (faulty, but...)
```
[12 x 0 from (2, 2)]
```
*temp* size() & elemSize()
```
[0 x 0] 0
```

[...]

PR #7874, opened 2016-12-15
Assert matrix dim > 0 in compare

addisonElliott:
resolves #7792 

Asserts that the matrix dimension is greater than zero (i.e. not an empty matrix) within the compare function. This resolves the issue of a segmentation fault occuring when the matrix is empty because the elemSize returns 0 and thus dividing by the elemSize results in a divide by zero error.

PR #8779, opened 2017-05-23
resolves https://github.com/opencv/opencv/issues/7792

vpisarev:
replaces https://github.com/opencv/opencv/pull/7874

### DIFF
--- a/modules/core/test/test_arithm.cpp
+++ b/modules/core/test/test_arithm.cpp
@@ -1912,3 +1912,13 @@ TEST(Subtract, scalarc4_matc4)
 
     ASSERT_EQ(0, cv::norm(cv::Mat(5, 5, CV_8UC4, cv::Scalar::all(250)), destImage, cv::NORM_INF));
 }
+
+TEST(Compare, empty)
+{
+    cv::Mat temp, dst1, dst2;
+    cv::compare(temp, temp, dst1, cv::CMP_EQ);
+    dst2 = temp > 5;
+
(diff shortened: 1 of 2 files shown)


modules/core/test/test_arithm.cpp
```diff
@@ -1912,3 +1912,13 @@ TEST(Subtract, scalarc4_matc4)
 
     ASSERT_EQ(0, cv::norm(cv::Mat(5, 5, CV_8UC4, cv::Scalar::all(250)), destImage, cv::NORM_INF));
 }
+
+TEST(Compare, empty)
+{
+    cv::Mat temp, dst1, dst2;
+    cv::compare(temp, temp, dst1, cv::CMP_EQ);
+    dst2 = temp > 5;
```
alalek: I still believe that these cases are not normal, because most of them is a result of program logic error. So we should not eat such errors silently.
See comment of original issue: https://github.com/opencv/opencv/issues/7792#issuecomment-265123281

vpisarev: @alalek, this is rather philosophical question. I believe that some cases can be "eaten". In this particular case the error will likely be signalled later when user will try to analyze the comparison result. And it will likely be signalled as exception, not as a program crash.

vpisarev:
:+1:

modules/core/test/test_arithm.cpp
ArkadiuszRaj: :+1:

PR #8783, opened 2017-05-23
change tutorial to use enum name instead of value

adhintz:
### This pullrequest changes

Change tutorial to use the enum name instead of its value.


vpisarev:
:+1:

PR #8794, opened 2017-05-25
Added Halide backend support for deep learning layers

dkurt:
### This pullrequest changes

Added looking for Halide library. Require flags `WITH_HALIDE` and `HALIDE_ROOT_DIR`.
If Halide was found, defines `HAVE_HALIDE` to 1. Otherwise defines to 0.
In code we should use `#if HAVE_HALIDE /*...*/ #endif` instead `#ifdef HAVE_HALIDE /*...*/ #endif`.

Installation guide might be found at dnn/tutorials/tutorial_dnn_halide.markdown.

TODO: Change `HALIDE_ROOT_DIR` to something more suitable. It might be `HALIDE_BUILD_DIR`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,6 +112,12 @@ endif()
 
 include(cmake/OpenCVUtils.cmake)
 
+if(OPENCV_CXX11)
+  #cmake_minimum_required(VERSION 3.1.0 FATAL_ERROR)
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
+endif()
+
 # must go before the project command
 ocv_update(CMAKE_CONFIGURATION_TYPES "Debug;Release" CACHE STRING "Configs" FORCE)
 if(DEFINED CMAKE_BUILD_TYPE)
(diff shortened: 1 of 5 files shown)


alalek:
Could you please rebase commit from this branch instead: https://github.com/alalek/opencv/commits/halide_cmake

Usage (requires CMake 3.x):
```
HALIDE_ROOT_DIR=<halide-path>/build cmake -DOPENCV_CXX11=ON -DWITH_HALIDE=ON <other regular options>
```

`#ifdef HAVE_HALIDE` is preferable (more compatible way to integrate compile-dependent features).

Usage of `HALIDE_BUILD_DIR` should be avoided (we should not clone variables and confuse developers: HALIDE_ROOT_DIR / HALIDE_BUILD_DIR / HALIDE_INSTALL_DIR / etc - `HALIDE_ROOT_DIR` is enough).

dkurt:
@alalek , thanks! I think I've done it correctly.
I'll replace my `#if` to `#ifdef` and use something like `bool haveHalide();` following https://github.com/opencv/opencv/blob/master/modules/core/src/ovx.cpp#L49

vpisarev:
:+1:

PR #8796, opened 2017-05-26
fix the mser issue

liurs1990:
### This pullrequest changes
fix the mser detectRegions bug,  detectRegions would miss some msers sometimes.
I hava found the mser do not work correctly for special image file.

The comp1's history size may be never updated after merging two comps, so one history's parent history size maybe smaller than its child history.
and when the history var is smaller than 0,f, it parent history will be ignored, so it will miss some usefull msers.

it need to update the comp1's history size when its size changed.

### DIFF
--- a/modules/features2d/src/mser.cpp
+++ b/modules/features2d/src/mser.cpp
@@ -335,6 +335,8 @@ public:
             head = comp1->head;
             tail = comp2->tail;
             size = comp1->size + comp2->size;
+            // update the history size
+            history->size =size;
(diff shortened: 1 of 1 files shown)


vpisarev:
thanks! :+1:

gylns:
I think that doesn't fix the issue, When a component grew to a new gray level, its history's size should equal to the component's size. 
@vpisarev, I have create a new PR #8910, would you like to give a review?

PR #8797, opened 2017-05-26
objdetect: fix in SSE code

### LINKED ISSUE #8795: HOGDescriptor::computeGradient produces an access violation for some windowed inputs

OpenCV =>  3.2
OS => Windows 64 bit
Compiler => Visual Studio 2017

##### Detailed description
With 0 padding and other parameters that work for other images, HOGDescriptor::computeGradient will produce an access violation error on line 374 of hog.cpp
This is caused by the value at xmap[-1] being changed to 65533 in line 328 of hog.cpp, which then means that p00 is set to an invalid value. This doesn't happen immediately but halfway through the image.
I don't know why this happens for some inputs and not others, but either there's a bug to fix or an assertion that should be made that prevents us reaching this step.

##### Steps to reproduce
run the code below
```.cpp
#include "opencv2/core.hpp"
#include "opencv2/objdetect.hpp"
using namespace cv;
int main()
{
   Mat test= Mat::zeros(576,704,CV_8UC3);
   Rect r(68,380,112,192);
   HOGDescriptor hog;
   std::vector<float> desc;
   hog.compute(test(r), desc, Size(8,8), Size(0,0)); //it throws during the execution of this line
   return 0;
}
```

sovrasov:
probably resolves #8795

### DIFF
--- a/modules/objdetect/src/hog.cpp
+++ b/modules/objdetect/src/hog.cpp
@@ -325,8 +325,14 @@ void HOGDescriptor::computeGradient(const Mat& img, Mat& grad, Mat& qangle,
 #if CV_SSE2
         __m128i ithree = _mm_set1_epi32(3);
         for ( ; x <= end - 4; x += 4)
-            _mm_storeu_si128((__m128i*)(xmap + x), _mm_mullo_epi16(ithree,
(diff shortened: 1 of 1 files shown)


vpisarev:
:+1:

PR #8798, opened 2017-05-26
fix estimateGlobalMotionRansac;

BKNio:
The problem arise when `ninliersMax < params.size` is true (line 454).  `subset0best` and `subset1best` already contain normalized points, but in case `motionModel` equal, for example, `MM_SIMILARITY` `estimateGlobMotionLeastSquaresSimilarity` performs normalizing again (line 283), end returns incorrect corrections `T0` and `T1` because data is already normalized. This is extremely rare case, but it happens sometimes

### DIFF
--- a/modules/videostab/src/global_motion.cpp
+++ b/modules/videostab/src/global_motion.cpp
@@ -402,8 +402,8 @@ Mat estimateGlobalMotionRansac(
     std::vector<Point2f> subset1(params.size);
 
     // best hypothesis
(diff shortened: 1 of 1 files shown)


vpisarev:
thanks, Dinar! I'm approving this PR, but if you have time, it would be nice to have a regression test to catch this case, so that someone does not "optimize" this code back to the wrong version sometimes later
 :+1:

PR #8799, opened 2017-05-26
Removed CMake configuration files from gitignore.

terfendail:
### This pullrequest changes
Since builds should be performed out of the source tree now it looks useful to be able to find out accidentally created cmake configs with `git status`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,8 +17,6 @@ bin/
 *.rule
 *.tmp
 */debug
-*/CMakeFiles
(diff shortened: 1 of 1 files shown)


terfendail:
Does it make sense to exclude all build related artifacts(`bin/`,`*/debug`,`build`, maybe `*.obj`) as well?

alalek:
:+1:

build folders don't confuse or break anything, so we can leave them "as is".

PR #8800, opened 2017-05-26
Filestorage node read default and std::string

### LINKED ISSUE #8767: Why not use the default in FileNode::operator >> ?
- OpenCV => 3.2

##### Detailed description

The `cv::read(const FileNode& node, value, default )` [here](https://github.com/opencv/opencv/blob/3.2.0/modules/core/include/opencv2/core/persistence.hpp#L713) uses a default value but not the related `FileNode::operator >>` 

It would be useful to keep same behaviour on the operator side and quite easy to implement.
For example the operator defined [here](https://github.com/opencv/opencv/blob/3.2.0/modules/core/include/opencv2/core/persistence.hpp#L1166) might become:
```.cpp
template<typename _Tp> static inline
void operator >> (const FileNode& n, _Tp& value)
{
    //OLD read( n, value, _Tp());
    read( n, value, value);   // NEW
}
```
I can agree that `_Tp()` object constructor should create the object in its default state but what with simple type like int and others ?

Using suggested change should help to use application based defaults both for objects and simple types.

Same for the [FileNodeIterator version](https://github.com/opencv/opencv/blob/3.2.0/modules/core/include/opencv2/core/persistence.hpp#L1142) and [std::string version](https://github.com/opencv/opencv/blob/3.2.0/modules/core/include/opencv2/core/cvstd.inl.hpp#L152) 

In addiciton, about the std::string version it would be also useful to add a cv::read version for it [here](https://github.com/opencv/opencv/blob/3.2.0/modules/core/include/opencv2/core/persistence.hpp#L716) because of unsupported cast of `value` by reference 
```.cpp
...
// available
CV_EXPORTS void read(const FileNode& node, String& value, const String& default_value);
// to add
CV_EXPORTS void read(const FileNode& node, std::string& value, const std::string& default_value);
```

##### Steps to reproduce
For example 

```.cpp
class MyClass
{
    Init()
    {
        m_myParam(5); // typical class initialization
    }
    bool ReadConfig()
    {
        Init();
        ...
        // IF node["myParam"] IS NOT DEFINED

        // this is bad because  m_myParam will become 0
        node["myParam"] >> m_myParam;

        // to keep application default value:
        if(!node["myParam"].empty())
                node["myParam"] >> m_myParam;

        // or
        cv::read(node["myParam"], m_myParam, m_myParam);
        ...
    }
}
```

PkLab:
resolves #8767

### This pullrequest changes
in cv::read(FileNode, ...) 
- use default_value even if data type mismatch
- add std::string overload
- little changes in the related doc

### DIFF
--- a/modules/core/src/persistence.cpp
+++ b/modules/core/src/persistence.cpp
@@ -7374,26 +7374,26 @@ void read(const FileNode& node, int& value, int default_value)
 {
     value = !node.node ? default_value :
     CV_NODE_IS_INT(node.node->tag) ? node.node->data.i :
-    CV_NODE_IS_REAL(node.node->tag) ? cvRound(node.node->data.f) : 0x7fffffff;
+    CV_NODE_IS_REAL(node.node->tag) ? cvRound(node.node->data.f) : default_value;
 }
 
 void read(const FileNode& node, float& value, float default_value)
 {
     value = !node.node ? default_value :
-        CV_NODE_IS_INT(node.node->tag) ? (float)node.node->data.i :
-        CV_NODE_IS_REAL(node.node->tag) ? (float)node.node->data.f : 1e30f;
+        CV_NODE_IS_REAL(node.node->tag) ? saturate_cast<float>(node.node->data.f):
+        CV_NODE_IS_INT(node.node->tag) ? (float)node.node->data.i :default_value;
 }
 
 void read(const FileNode& node, double& value, double default_value)
 {
     value = !node.node ? default_value :
-        CV_NODE_IS_INT(node.node->tag) ? (double)node.node->data.i :
-        CV_NODE_IS_REAL(node.node->tag) ? node.node->data.f : 1e300;
+        CV_NODE_IS_REAL(node.node->tag) ? node.node->data.f :
+        CV_NODE_IS_INT(node.node->tag) ? (double)node.node->data.i : default_value;
 }
 
 void read(const FileNode& node, String& value, const String& default_value)
 {
-    value = !node.node ? default_value : CV_NODE_IS_STRING(node.node->tag) ? String(node.node->data.str.ptr) : String();
+    value = !node.node ? default_value : CV_NODE_IS_STRING(node.node->tag) ? String(node.node->data.str.ptr) : default_value;
 }
 
 }
--- a/modules/core/src/persistence.cpp
+++ b/modules/core/src/persistence.cpp
@@ -7396,6 +7396,13 @@ void read(const FileNode& node, String& value, const String& default_value)
     value = !node.node ? default_value : CV_NODE_IS_STRING(node.node->tag) ? String(node.node->data.str.ptr) : default_value;
 }
 
+#ifndef OPENCV_NOSTL
+void read(const FileNode& node, std::string& value, const std::string& default_value)
+{
+    value = !node.node ? default_value : CV_NODE_IS_STRING(node.node->tag) ? std::string(node.node->data.str.ptr) : default_value;
+}
+#endif
+
 }
 
 
--- a/modules/core/include/opencv2/core/cvstd.inl.hpp
+++ b/modules/core/include/opencv2/core/cvstd.inl.hpp
@@ -151,9 +151,7 @@ FileNode::operator std::string() const
 template<> inline
 void operator >> (const FileNode& n, std::string& value)
 {
-    String val;
-    read(n, val, val);
-    value = val;
+    read(n, value, std::string());
 }
 
 template<> inline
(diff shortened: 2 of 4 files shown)


modules/core/src/persistence.cpp
```diff
@@ -7374,28 +7374,35 @@ void read(const FileNode& node, int& value, int default_value)
 {
     value = !node.node ? default_value :
     CV_NODE_IS_INT(node.node->tag) ? node.node->data.i :
-    CV_NODE_IS_REAL(node.node->tag) ? cvRound(node.node->data.f) : 0x7fffffff;
+    CV_NODE_IS_REAL(node.node->tag) ? cvRound(node.node->data.f) : default_value;
```
PkLab: Ok even if `cvRound` is as it was. 
- Is this means that any type <>`int` should raise a CV_Error or just in case of Real  ?
- Do you have any preferred error code / msg  for CV_Error  ?
- is saturate_cast in the float version ok ?

sovrasov:
Conversion real->int should be prohibited.
StsParseError seems to be suitable.
Saturate cast for floats is OK.

PkLab:
> Conversion real->int should be prohibited.

I've found similar behaviour here https://github.com/opencv/opencv/blob/master/modules/core/src/persistence.cpp#L5018

What about silent conversion done by 
- [reading string](https://github.com/PkLab/opencv/blob/ff901519741eb98791920ee2956307ec46e97da7/modules/core/src/persistence.cpp#L7396) ...if node data is not string ?
- and ... reading float or double... if node data is not int and not real ?

PkLab:
Sure The OpenCV team has good reason to prohibit conversion real -> int but I can't fully understand. 

1st, it was a previous behaviour... modifications can break previous code

2nd, suppose  I've a config file where a value was double. Later I would like to change that member to `int`. As soon as I'll run the application it will rise the CV_Error... this means that new coder have to manage error while using cv::read

Is not clear to me if the idea is:  _everting different than expected is an error_ .
1. What is the difference in the logic between `int<-round(MAX_DOUBLE)` and `float<-saturate_cast<float>(MAX_DOUBLE)`
2. Why I can't round a double but if a string is found while reading double the default is used ?

Can you show me some point to understand your intent ?
Thank you

PkLab:
any news on this ?

sovrasov:
If user gets `MAX_FLOAT` after `saturate_cast<float>(MAX_DOUBLE)` it won't be so harmful as to get integer 5 instead o float 5.5. I see the rising of Error is too big concern for users, so default value is also suitable in this case.

PkLab:
If I've right understood the following table should apply (no more CV_Error in case of real->int)

CASES       | read int| read float| read double | read string
------------|---------|-----------|-------------|------------
found int   | ok      | ok        | ok          | default 
found float | default | ok        | ok          | default
found double| default | stauration | ok          | default
found str   | default | default   | default     | ok
found obj   | default | default   | default     | default
not found   | default | default   | default     | default

sovrasov:
Ok, you'r right.

PkLab:
What about all other cases ? For example 
- read of [Point_<_Tp>](https://github.com/opencv/opencv/blob/3ea02e4fcc6ccc68305a1b1c709861bdfa972efb/modules/core/include/opencv2/core/persistence.hpp#L722 ) (Point3,Size,...) makes implicit rounding when read `Point2i<-Point2d`
- already cited [cvReadRawDataSlice](https://github.com/opencv/opencv/blob/master/modules/core/src/persistence.cpp#L5018)
- maybe others

Till now conversion int<-real is supported and done using round everywhere. 
I suppose you want to change all of them even if this can have some impact on existent user code.

sovrasov:
I'll look closer at these cases. We've decided to make changes slowly. For now it's better to merge only first commit of this PR.

PkLab:
I don't know... the overload of cv::read() for std::string is not so relevant.
At the end I prefer to leave the section as is waiting for a final decision. 
For that time please consider also #8767 .

PkLab:
@sovrasov 

> I'll look closer at these cases. We've decided to make changes slowly. 

is there any chance to see changes in 3.3.0 ?

sovrasov:
I think in 3.3 will be integrated changes from #8975 only.

vpisarev:
ok, let's close it. Implicit conversions may yield hard-to-debug bugs, especially in big applications developed by big teams. If the "data scheme" of a file storage is known, the proper types should be used to read values. Very safe conversions can be done implicitly, of course, like `float->double` or `int->float` (even though I'm not even sure about the latter). Type of the node can be requested with `FileNode::type()` method; so convenience wrapper functions that do implicit conversions can always be added by the user if needed

PR #8814, opened 2017-05-29
Use num_threads clause of #pragma omp parallel for OpenMP

woodychow:
to avoid calling omp_set_num_threads for the entire application.


sovrasov:
:+1:

woodychow:
Thanks.

The precommit test failed. I think there is something wrong with the test as my MR only changes OpenMP, and the tested binary is compiled with pthreads not OpenMP.

sovrasov:
There is a problem with the arm boards, don't worry about these failures.

woodychow:
Thanks!

PR #8816, opened 2017-05-29
Fixed snprintf for VS 2013

mshabunin:
See #8771 for details

### This pullrequest changes

Should fix VS2013 build.

### DIFF
--- a/modules/core/include/opencv2/core/private.hpp
+++ b/modules/core/include/opencv2/core/private.hpp
@@ -157,7 +157,6 @@ static inline cv::Size cvGetMatSize( const CvMat* mat )
 namespace cv
 {
 CV_EXPORTS void scalarToRawData(const cv::Scalar& s, void* buf, int type, int unroll_to = 0);
-bool cv_snprintf(char* buf, int len, const char* fmt, ...);
 }
 
 // property implementation macros
--- a/modules/core/include/opencv2/core/private.hpp
+++ b/modules/core/include/opencv2/core/private.hpp
@@ -157,6 +157,7 @@ static inline cv::Size cvGetMatSize( const CvMat* mat )
 namespace cv
 {
 CV_EXPORTS void scalarToRawData(const cv::Scalar& s, void* buf, int type, int unroll_to = 0);
+bool cv_snprintf(char* buf, int len, const char* fmt, ...);
 }
 
 // property implementation macros
--- a/modules/core/src/lpsolver.cpp
+++ b/modules/core/src/lpsolver.cpp
@@ -41,7 +41,6 @@
 #include "precomp.hpp"
 #include <climits>
 #include <algorithm>
-#include <cstdarg>
 
 #define dprintf(x)
 #define print_matrix(x)
--- a/modules/core/src/out.cpp
+++ b/modules/core/src/out.cpp
@@ -103,7 +103,7 @@ namespace cv
             }
             else
             {
-                cv_snprintf(floatFormat, 8, "%%.%dg", std::min(precision, 20));
+                cv_snprintf(floatFormat, sizeof(floatFormat), "%%.%dg", std::min(precision, 20));
             }
 
             switch(mtx.depth())
--- a/modules/core/src/out.cpp
+++ b/modules/core/src/out.cpp
@@ -103,7 +103,7 @@ namespace cv
             }
             else
             {
-                snprintf(floatFormat, 8, "%%.%dg", std::min(precision, 20));
+                cv_snprintf(floatFormat, 8, "%%.%dg", std::min(precision, 20));
             }
 
             switch(mtx.depth())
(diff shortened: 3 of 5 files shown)


mshabunin:
@PkLab , could you please check whether this PR fixes VS2013 compilation for you?

PkLab:
It builds on VS2013 even if I've doubt about 
see [A value of predefined macro __cplusplus is still 199711L](https://connect.microsoft.com/VisualStudio/feedback/details/763051/a-value-of-predefined-macro-cplusplus-is-still-199711L)

PkLab:
The code doesn't work correctly on VS2013. Try this test code
```.cpp
#define BUF_LEN 4
int main()
{
  char buf[BUF_LEN];
  bool res;
  res = cv_snprintf(buf,BUF_LEN,"%s","abc");
  cout << "buf=" << buf << endl;
  if (res)
    cout << "cv_snprintf: OK" << endl;
  else
    cout << "cv_snprintf: Error " << endl;
  char ch = cin.get();
  return 0;
}
```
On VS2013 cv_snprintf returns false and `buf="ab"`
while on GCC returns true and `buf="abc"` which is correct

alalek:
@PkLab What is result with `BUF_LEN=5`?

PkLab:
@alalek of course with BUF_LEN=5 the result is ok but it shouyld work also with BUF_LEN=5 because to write to write "abc" you need just 4 byte including null terminator.

I would write more simply
```.cpp
bool pk_snprintf(char* buf, int len, const char* fmt, ...)
{
    if (len <= 0) return false;
    va_list va;
    va_start(va, fmt);
#if defined _MSC_VER
    int res = _vsnprintf_s((char *)buf, len, _TRUNCATE, fmt, va);
    // ensure null terminating on VS
    if (res < 0) buf[len - 1] = 0;
    else buf[res] = 0;
#else
    int res = vsnprintf((char *)buf, len, fmt, va);
#endif
    va_end(va);
    return res >= 0 && res < len;
}
```

modules/core/include/opencv2/core/private.hpp
```diff
@@ -157,6 +157,7 @@ static inline cv::Size cvGetMatSize( const CvMat* mat )
 namespace cv
 {
 CV_EXPORTS void scalarToRawData(const cv::Scalar& s, void* buf, int type, int unroll_to = 0);
+bool cv_snprintf(char* buf, int len, const char* fmt, ...);
```
vpisarev: non-exported functions should not be declared anywhere in include/opencv2/.

mshabunin:
I've removed declaration from `private.hpp` and replaced implementation with the one proposed by @PkLab

alalek:
@PkLab With `BUF_LEN=3` what is `res` value? If it is 3, then statement `buf[res] = 0;` will corrupt memory outsize of buffer.
Porbably MSVS2012 do this: http://demin.ws/blog/english/2013/01/28/use-snprintf-on-different-platforms/

PkLab:
@alalek With `BUF_LEN`=3 `res=-1` 

- snprintf is not same as vsnprintf nor _vsnprintf_s
- also on VS2012 `_vsnprintf_s` returns -1 if `count= _TRUNCATE` [see remarks](https://msdn.microsoft.com/en-us/library/d3xd30zz(v=vs.110).aspx)

I have understood that  news with VS2015 is about that the null terminating is guaranteed. With previous versions you get null terminator only if there is left space in the buffer.

PS: should we check also [String format( const char* fmt, ... )](https://github.com/opencv/opencv/blob/master/modules/core/src/system.cpp#L746) ?

alalek:
@PkLab OK, It seems valid. Thank you!

PkLab:
btw we can stay safer using `if ((res < 0) || (res>=len)) buf[len - 1] = 0;`

PkLab:
or better, 
```.cpp
bool ok = res >= 0 && res < len;
if (!ok)  buf[len - 1] = 0;
else buf[res] = 0; // this line is useless
...
return ok;
```
just in case in the future _vsnprintf_s would follow same rules as gcc returning " total number of characters which would have been written" ;-)

mshabunin:
@PkLab , I've updated the VS branch, please take a look. 

BTW, what do you think about adding `&& _MSC_VER <= 1800` condition to match only VisualStudio versions <= 2013?

PkLab:
@mshabunin  It works here. 

About adding `&& _MSC_VER <= 1800`, `snprintf` is available since VS1015 but they suggest to use "More secure versions of these..._snprintf_s...". 

Because, at the end, our solution uses `_vsnprintf_s` 
- we don't have to take care to MSC_VER.
- and I think that `&& __cplusplus < 201103L` should be removed in [cv::format](https://github.com/opencv/opencv/blob/master/modules/core/src/system.cpp#L746)

alalek:
@PkLab Right, `__cplusplus < 201103L` condition doesn't have any sense to MSVC (MSVS 2017 still has `199711L` value)

PkLab:
@alalek right, and although `vsnprintf` is defined since MSV2105 its behaviour is still different from gcc and MS suggests to use more secure version. At the end `#if defined _MSC_VER` is enough.

alalek:
Added update to this patch.
I have added wrapper over `vsnprintf` too, because I will need it soon.

alalek:
:+1:

PR #8818, opened 2017-05-29
Do not clean src and gen directories when java is disabled

mshabunin:
### This pullrequest changes

When _java_ is disabled with `-DBUILD_opencv_java=OFF`, but _ant_ and _JNI_ are installed in the system cmake script removes `<build>/src` and `<build>/gen` directories. this behaviour could lead to build directory corruption in some configurations.

This is just a workaround, it should be better to use `CMAKE_CURRENT_BINARY_DIR` instead of hardcoded paths.


alalek:
:+1:

PR #8819, opened 2017-05-29
Fix build with CUDA on windows with Ninja and JOM (#6372)

### LINKED ISSUE #6372: OpenCV 3.1.0 build fails on Windows with Ninja Generator
This is a template helping you to create an issue which can be processes as quickly as possible. Feel free to add additional information or remove not relevant points if you do not need them.

If you have a question rather than reporting a bug please go to http://answers.opencv.org where you get much faster responses.
### Please state the information for your system
- OpenCV version: 3.1.0
- Host OS: Windows 10
- Compiler & CMake: MSVC 120 & CMake 3.5
### In which part of the OpenCV library you got the issue?
- CMakeLists.txt
### Expected behaviour

Build succeeds
### Actual behaviour

LINK : fatal error LNK1104: cannot open file 'Files\NVIDIA.obj'
### Additional description

This [line](https://github.com/Itseez/opencv/blob/master/CMakeLists.txt#L638) breaks the build with Ninja under Windows. I think the line should be:

``` CMake
# this
set(OPENCV_LINKER_LIBS ${OPENCV_LINKER_LIBS} ${CMAKE_LIBRARY_PATH_FLAG}$<SHELL_PATH:${p}>)
# instead of that
# set(OPENCV_LINKER_LIBS ${OPENCV_LINKER_LIBS} -L${p})
```

But even with this change the build fails because of the spaces in the path which Ninja does not like. Maybe `$<SHELL_PATH:...>` should return a short path but that's a CMake issue.

As a workaround one can use:

``` CMake
if(NOT CMAKE_GENERATOR MATCHES "Ninja" AND NOT MSVC)
  set(OPENCV_LINKER_LIBS ${OPENCV_LINKER_LIBS} -L${p})
endif()   
```

This problem also affects consuming projects as the link flags end up in the `OpenCVModules*.cmake` files.

I find Ninja much faster than msbuild on Windows especially when building with CUDA so it would be nice if OpenCV supported this generator too.

Robzz:
Fixed #6372. No idea if it's the proper way to fix it, but it works for me.

### This pullrequest changes

Quote the CUDA related flags passed to the linker in the main CMakeLists.txt


Robzz:
I just tried this and it fails building with visual studio. I'll try to come up with something that works for both.

Robzz:
It should be good now. I only quoted the linker flags when using JOM or Ninja with the Visual Studio compiler.

alalek:
Looks good to me.
Thank you! :+1:

PR #8820, opened 2017-05-30
Add function to sort keypoints that are calculated in parallel

woodychow:
to ensure stable output

Needed by opencv/opencv_contrib#1195

### DIFF
--- a/modules/features2d/src/keypoint.cpp
+++ b/modules/features2d/src/keypoint.cpp
@@ -221,4 +221,44 @@ void KeyPointsFilter::removeDuplicated( std::vector<KeyPoint>& keypoints )
     keypoints.resize(j);
 }
 
+struct KeyPoint12_LessThan
+{
+    bool operator()(const KeyPoint &kp1, const KeyPoint &kp2) const
+    {
+        if( kp1.pt.x != kp2.pt.x )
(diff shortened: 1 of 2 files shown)


modules/features2d/src/keypoint.cpp
```diff
@@ -221,4 +221,44 @@ void KeyPointsFilter::removeDuplicated( std::vector<KeyPoint>& keypoints )
...
+
+    std::sort(keypoints.begin(), keypoints.end(), KeyPoint12_LessThan());
+
+    for( i = 0, j = 1; j < n; ++j )
+    {
+        const KeyPoint& kp1 = keypoints[i];
+        const KeyPoint& kp2 = keypoints[j];
+        if( kp1.pt.x != kp2.pt.x || kp1.pt.y != kp2.pt.y ||
+            kp1.size != kp2.size || kp1.angle != kp2.angle ) {
+            keypoints[++i] = keypoints[j];
```
vpisarev: should not we use postfix increment here? (and replace keypoints.resize(i+1) with keypoints.resize(i) below)

vpisarev:
:+1:

alalek:
This should be merged after the contrib patch or it would require appropriate tests.

PR #8823, opened 2017-05-30
fixed typo in pyopencv_from Point3d

### LINKED ISSUE #8822: Typo in Point3d python converter
There is a typo in
https://github.com/opencv/opencv/blob/master/modules/python/src2/cv2.cpp

```
template<>
PyObject* pyopencv_from(const Point3d& p)
{
    return Py_BuildValue("(ddd)", p.x, p.y, p.y);
}
```

last line should be
`    return Py_BuildValue("(ddd)", p.x, p.y, p.z);`

wpalfi:
resolves #8822


alalek:
Nice catch! :+1:

PR #8828, opened 2017-05-31
Multithreaded IPP gaussian blur

woodychow:


modules/imgproc/src/smooth.cpp
```diff
@@ -2272,7 +2319,26 @@ static bool ipp_GaussianBlur(InputArray _src, OutputArray _dst, Size ksize,
         if(!ippBorder.m_borderType)
             return false;
 
-        CV_INSTRUMENT_FUN_IPP(::ipp::iwiFilterGaussian, &iwSrc, &iwDst, ksize.width, (float)sigma1, ippBorder);
+        // IW 2017u2 has bug which doesn't allow use of partial inMem with tiling
+        if((((ippBorder.m_borderFlags)&ippBorderInMem) && ((ippBorder.m_borderFlags)&ippBorderInMem) != ippBorderInMem)) {
+            return false;
```
alalek: 1) Should we fallback to single-threaded version?
```
CV_INSTRUMENT_FUN_IPP(::ipp::iwiFilterGaussian, &iwSrc, &iwDst, ksize.width, (float)sigma1, ippBorder);
```

2) Should we check version macro: `IW_VERSION_COMPLEX <= 20170002` (in case of fix in the next release)?

3) Should we add new macro for disabled IPP feature [here](https://github.com/opencv/opencv/blob/0d86681b88d4605b49ea06e0e6218c1bcc36925c/modules/core/include/opencv2/core/private.hpp#L189-L208)?
```
#define IPP_DISABLE_FILTERING_INMEM_PARTIAL 1 // IW 2017u2 has bug which doesn't allow use of partial inMem with tiling
```

woodychow: I agree with 1. Actually I just copied that logic from ipp_bilateralFilter.

I am not sure about 2 as there is no guarantee that it will be fixed in the next version. Has anyone reported the bug to Intel yet? It seems to me that 3 is the way to go for now.

woodychow:
@alalek, how does the code look now?

alalek:
@woodychow Looks good to me! Thank you :+1:

woodychow:
Thanks!

PR #8829, opened 2017-05-31
destructor virtual for interface BufferPoolController

Ubiquite:


alalek:
Lifetime management of this object is not assumed via this interface, so "virtual destructor" is not needed here.

PR #8833, opened 2017-05-31
Added performance test for nearest neighbor resize

terfendail:
### This pullrequest changes
Added performance test for resizeNN

### DIFF
--- a/modules/imgproc/perf/perf_resize.cpp
+++ b/modules/imgproc/perf/perf_resize.cpp
@@ -120,3 +120,31 @@ PERF_TEST_P(MatInfo_Size_Scale_Area, ResizeArea,
     //difference equal to 1 is allowed because of different possible rounding modes: round-to-nearest vs bankers' rounding
     SANITY_CHECK(dst, 1);
 }
+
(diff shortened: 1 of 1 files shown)


modules/imgproc/perf/perf_resize.cpp
```diff
@@ -120,3 +120,31 @@ PERF_TEST_P(MatInfo_Size_Scale_Area, ResizeArea,
...
+    Size to(cvRound(from.width * scale), cvRound(from.height * scale));
+    cv::Mat dst(to, matType);
+
+    declare.in(src, WARMUP_RNG).out(dst);
+    declare.time(100);
+
+    TEST_CYCLE() resize(src, dst, dst.size(), 0, 0, INTER_NEAREST);
+
+    //difference equal to 1 is allowed because of different possible rounding modes: round-to-nearest vs bankers' rounding
+    SANITY_CHECK(dst, 1);
```
alalek: Actually max difference here is 255: values between neighbor pixels. "1" is just pixel coordinate difference caused by rounding error.

My suggestion for checks is: `countNonZero(dst) > 0` + `SANITY_CHECK_NOTHING()`.

alalek:
:+1:

PR #8835, opened 2017-06-01
use pkg-config for finding gstreamer approprietely in Windows.

stnk20:
### This pullrequest changes
Use pkg-config for finding gstreamer approprietely in Windows.


alalek:
Validation is required for this PR. Also pkg-config installation is huge dependency on Windows.

stnk20:
I found simple workaround not relying on pkg-config. https://github.com/opencv/opencv/issues/8836#issuecomment-308625466

PR #8841, opened 2017-06-01
Fix a typo in OpenCVUtils.cmake

bouffa:
### This pullrequest changes

To test for Visual Studio generator, MSVC and not MSVS should be used


bouffa:
There are still other problems to be able to install pdb files. I'm closing this PR and will make another PR with full fixes

mshabunin:
@bouffa , you can update this PR later by pushing more commits to the same branch (`patch-1`).

alalek:
@bouffa BTW, CMake has different support for PDB files for CMake versions 2.8.11.x, 2.8.12.x and >=3.0.0. Probably we should target for modern CMake versions.

PR #8847, opened 2017-06-03
GSoC - OpenCV Java Highgui Class

Cartucho:
The objective is to:

    *Reduce greatly the number of lines of code in the Java codes;
    *Make it easy for Java users to add a trackbar and show the results;
    *Get the code more similar between C++, Java and Python, making the tutorials more uniform.

### DIFF
--- a/modules/highgui/CMakeLists.txt
+++ b/modules/highgui/CMakeLists.txt
@@ -1,5 +1,9 @@
 set(the_description "High-level GUI")
-ocv_add_module(highgui opencv_imgproc opencv_imgcodecs OPTIONAL opencv_videoio WRAP python)
+if(ANDROID)
+  ocv_add_module(highgui opencv_imgproc opencv_imgcodecs OPTIONAL opencv_videoio WRAP python)
+else()
+  ocv_add_module(highgui opencv_imgproc opencv_imgcodecs OPTIONAL opencv_videoio WRAP python java)
+endif()
 
 # ----------------------------------------------------------------------------
 #  CMake file for highgui. See root CMakeLists.txt
--- a/modules/highgui/misc/java/src/java/highgui+HighGui.java
+++ b/modules/highgui/misc/java/src/java/highgui+HighGui.java
@@ -1,0 +1,193 @@
+package org.opencv.highgui;
+
+import org.opencv.core.Mat;
+
+import javax.swing.*;
+import java.awt.*;
+import java.awt.event.KeyEvent;
+import java.awt.event.KeyListener;
+import java.awt.image.BufferedImage;
+import java.awt.image.DataBufferByte;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * This class was designed for use in Java applications
+ * to recreate the OpenCV HighGui functionalities.
+ */
+public final class HighGui {
+
+    // Constants for namedWindow
+    public final static int WINDOW_NORMAL = ImageWindow.WINDOW_NORMAL;
+    public final static int WINDOW_AUTOSIZE = ImageWindow.WINDOW_AUTOSIZE;
+
+    // Control Variables
+    public static int n_closed_windows = 0;
+    public static int pressedKey = -1;
+    public static CountDownLatch latch = new CountDownLatch(1);
+
+    // Windows Map
+    public static Map<String, ImageWindow> windows = new HashMap<String, ImageWindow>();
+
+    public static void namedWindow(String winname) {
+        namedWindow(winname, HighGui.WINDOW_AUTOSIZE);
+    }
+
+    public static void namedWindow(String winname, int flag) {
+        ImageWindow newWin = new ImageWindow(winname, flag);
+        if (windows.get(winname) == null) windows.put(winname, newWin);
+    }
+
+    public static void imshow(String winname, Mat img) {
+        if (img.empty()) {
+            System.err.println("Error: Empty image in imshow");
+            System.exit(-1);
+        } else {
+            ImageWindow tmpWindow = windows.get(winname);
+            if (tmpWindow == null) {
+                ImageWindow newWin = new ImageWindow(winname, img);
+                windows.put(winname, newWin);
+            } else {
+                tmpWindow.setMat(img);
+            }
+        }
+    }
+
+    public static Image toBufferedImage(Mat m) {
+        int type = BufferedImage.TYPE_BYTE_GRAY;
+
+        if (m.channels() > 1) {
+            type = BufferedImage.TYPE_3BYTE_BGR;
+        }
+
+        int bufferSize = m.channels() * m.cols() * m.rows();
+        byte[] b = new byte[bufferSize];
+        m.get(0, 0, b); // get all the pixels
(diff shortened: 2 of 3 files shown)


Cartucho:
In this version I'm only adding the code for showing images in a window, the trackbar and click methods will be added later.

The methods I added for now were:

- imshow
- waitKey
- namedWindow
- destroyWindow
- destroyAllWindows
- resizeWindow
- moveWindow

Cartucho:
I also did a bunch of tests you can find the (code here)[https://github.com/Cartucho/OpencvPRs/blob/master/HighGui%20Java/HelloWorld.java] you just need to uncomment each of the tests.

Cartucho:
@vrabaud @bmagyar @StevenPuttemans @mshabunin

karlphillip:
Why did it took so long for someone to write a HighGui for Java?!? This is great.
I wrote a few tests as well to see how it would behave and it seems robust enough.
Once the coding style is adjusted to match OpenCV's style I believe this can be merged into OpenCV's repository.

alalek:
> it took so long for someone to write a HighGui for Java

Primary Java bindings goal (for OpenCV developers) is supporting of Android development (with own custom UI implementation - there are no original Java AWT, Swing on Android).

Cartucho:
@alalek Anyway this code could be used for Java right? At least for the tutorials?

alalek:
@Cartucho Right!

StevenPuttemans:
Seems nice, just one remark, why are we pushing this into tutorial code? This seems the wrong location for it?

alalek:
It will be great to put these files into `modules/highgui/misc/java/src/java`, like this folder: https://github.com/opencv/opencv/tree/3.2.0/modules/core/misc/java/src/java
`+` sing in filename is used for package name separation (no subdirectories support yet), please keep using it.

Also you need to enable Java bindings generator for [highgui](https://github.com/opencv/opencv/blob/3.2.0/modules/highgui/CMakeLists.txt#L2):
```.diff
-ocv_add_module(highgui opencv_imgproc opencv_imgcodecs OPTIONAL opencv_videoio WRAP python)
+ocv_add_module(highgui opencv_imgproc opencv_imgcodecs
+               OPTIONAL opencv_videoio
+               WRAP python java)
```

("common" Java support is not tested for this module, - some errors may occurs)

modules/highgui/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 set(the_description "High-level GUI and Media I/O")
-ocv_add_module(highgui opencv_imgproc opencv_imgcodecs OPTIONAL opencv_videoio WRAP python)
+ocv_add_module(highgui opencv_imgproc opencv_imgcodecs OPTIONAL opencv_videoio WRAP python java)
```
vrabaud: As this fails on Android, you should do:
```
if(ANDROID)
 ocv_add_module(highgui opencv_imgproc opencv_imgcodecs OPTIONAL opencv_videoio WRAP python)
else()
 ocv_add_module(highgui opencv_imgproc opencv_imgcodecs OPTIONAL opencv_videoio WRAP python java)
endif()
```

modules/highgui/misc/java/src/java/highgui+ImageWindow.java
```diff
@@ -0,0 +1,123 @@
...
+    int flag;
+    int x = -1;
+    int y = -1;
+    int width = -1;
+    int height = -1;
+
+    public ImageWindow(String name, Mat img) {
+        this.name = name;
+        this.img = img;
+        this.flag = HighGui.WINDOW_NORMAL;
```
vrabaud: shouldn't you import the opencv HighGui module to be able to use that ?

modules/highgui/misc/java/src/java/highgui+HighGui.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+public class HighGui {
+
+    // Constants for namedWindow
+    static int WINDOW_NORMAL = 0; // enables you to resize the window
```
vrabaud: that should be: public static final int WINDOW_NORMAL

modules/highgui/misc/java/src/java/highgui+HighGui.java
```diff
@@ -0,0 +1,186 @@
...
+import java.awt.event.*;
+import java.awt.image.BufferedImage;
+import java.awt.image.DataBufferByte;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+public class HighGui {
```
vrabaud: you want those to be final most likely: public final class HighGui

modules/highgui/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 set(the_description "High-level GUI and Media I/O")
-ocv_add_module(highgui opencv_imgproc opencv_imgcodecs OPTIONAL opencv_videoio WRAP python)
+if(ANDROID)
+ ocv_add_module(highgui opencv_imgproc opencv_imgcodecs OPTIONAL opencv_videoio WRAP python)
```
vrabaud: needs a two-space indentation, not one

Cartucho:
@alalek do you know why the compilation is failing in Windows?

StevenPuttemans:
@Cartucho this is the reason of the fail

```
[javac] C:\build\precommit_windows64\build\src\org\opencv\highgui\ImageWindow.java:102: error: cannot find symbol
             [javac]         if (flag == HighGui.WINDOW_NORMAL) {
             [javac]                     ^
             [javac]   symbol:   variable HighGui
             [javac]   location: class ImageWindow
             [javac] 1 error
```

It seems HighGui is not knozn to the class?

vrabaud:
Yeah, we found the error but we think there is something missing. @alalek , do we need to create a "filelist" file like done in other modules ?

alalek:
@vrabaud `filelist` is not needed in general (until you will have module's specific JNI C++ code) - defaults should work fine.

vrabaud:
thx @alalek. Any idea on why it is failing on Windows then ? @Cartucho copied what was done for core but  static public constants do not seem to propagate

alalek:
Probably, there is missing:
```
import org.opencv.highgui.HighGui
```

Cartucho:
@alalek @StevenPuttemans . Vincent found the error, a circular dependency that got fixed. Thank you for your time!

Cartucho:
@alalek The [unsuccessful builds](http://pullrequest.opencv.org/buildbot/builders/precommit_linux64_no_opt/builds/11092/steps/test_videoio/logs/stdio) are not related to my code

vrabaud:
LGTM

vrabaud:
@alalek, this code looks good to me imho.

Later PR could simplify code like in here: 
https://github.com/opencv/opencv/blob/master/samples/java/tutorial_code/ImgProc/tutorial_template_matching/MatchTemplateDemo.java#L148
https://github.com/opencv/opencv/blob/master/samples/java/tutorial_code/core/mat_mask_operations/MatMaskOperations.java#L119
(basically no more explicit JFrames).

Should we actually modify those tutorials in this PR ?

alalek:
:+1:

PR #8852, opened 2017-06-05
Tests for videostab

BKNio:
Add regression tests for videostab module

### DIFF
--- a/modules/videostab/test/test_motion_estimation.cpp
+++ b/modules/videostab/test/test_motion_estimation.cpp
@@ -1,0 +1,171 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html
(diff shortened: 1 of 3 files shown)


modules/videostab/test/test_motion_estimation.cpp
```diff
@@ -0,0 +1,210 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
sovrasov: This license header is outdated. Please use
```
// of this distribution and at http://opencv.org/license.html
```
instead.

sovrasov:
:+1:, but squash commits into one before merge.

BKNio:
@sovrasov done

@vpisarev 
btw I have c++ [mosse](https://pdfs.semanticscholar.org/70c3/c9b9a40ca55264e454586dca2a6cf416f6e0.pdf) implimintation, does opencv need it?

PR #8860, opened 2017-06-06
cmake: java: use module's public headers only

### LINKED ISSUE #8849: Java wrappers for modules, which use protobuf, are not generated properly

- OpenCV => 3.2 master (fdba54885894aa875028ad08f53318d799509532)
- Operating System / Platform => OSX El Capitain 10.11.6 (15G1421)
- Compiler => Xcode Version 8.1 (8B62)

##### Detailed description

Java wrappers are not generated for dnn module. Enabling java for dnn leads to compilation errors related to the following issues:

1. Google protobuf library headers are not visible to java cmake scripts so building dnn with java leads to errors like below:
 
```
fatal error: google/protobuf/stubs/common.h: No such file or directory
compilation terminated.

```
2. Some constants are defined using cross-references to constants in other namespace, which is not interpreted properly by java generator. Example:
`
CONST DataType_INT_MIN_SENTINEL_DO_NOT_USE_=::google::protobuf::kint32min
`

3. CMake generates a lot of  warnings when building dnn with java support. These warnings are harmless (ex non-used variable in protobuf headers) and can be skipped to ensure clean logs  

##### Steps to reproduce

1. Enable contrib modules
2. Enable java wrappers generation for dnn module 
3.  build opencv

alalek:
resolves #8849


alalek:
Checked - no regressions are introduced into Java generated files.
:+1:

PR #8864, opened 2017-06-07
calib3d: add a calibration flag to fix tangential distortion

### LINKED ISSUE #8845: Feature Request: CV_CALIB_FIX_P1 and CV_CALIB_FIX_P2 flag for calibrateCamera

- OpenCV => Latest
- Operating System / Platform =>  Ubuntu 16.04
- Compiler => GCC 5.4

##### Detailed description

When specifying various flags when using the function calibrateCamera, according to http://docs.opencv.org/3.2.0/d9/d0c/group__calib3d.html#ga3207604e4b1a1758aa66acb6ed5aa65d there are options to fix K1 to K6, S1 to S4, but for some reason not P1 and P2. The only option is set them to zero. Is it possible to add in flags so we can provide an initial P1 and P2 and it will stay fixed during the optimization like all the other distortion coefficients?

sovrasov:
resolves #8845

### DIFF
--- a/modules/calib3d/src/calibration.cpp
+++ b/modules/calib3d/src/calibration.cpp
@@ -1319,7 +1319,7 @@ static double cvCalibrateCamera2Internal( const CvMat* objectPoints,
         (npoints->rows != 1 && npoints->cols != 1) )
         CV_Error( CV_StsUnsupportedFormat,
             "the array of point counters must be 1-dimensional integer vector" );
-    if(flags & CV_CALIB_TILTED_MODEL)
+    if(flags & CALIB_TILTED_MODEL)
     {
         //when the tilted sensor model is used the distortion coefficients matrix must have 14 parameters
         if (distCoeffs->cols*distCoeffs->rows != 14)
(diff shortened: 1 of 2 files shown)


modules/calib3d/src/calibration.cpp
```diff
@@ -1498,24 +1498,26 @@ static double cvCalibrateCamera2Internal( const CvMat* objectPoints,
...
+    if( !(flags & CALIB_THIN_PRISM_MODEL))
         flags |= CALIB_FIX_S1_S2_S3_S4;
-    if( !(flags & CV_CALIB_TILTED_MODEL))
+    if( !(flags & CALIB_TILTED_MODEL))
         flags |= CALIB_FIX_TAUX_TAUY;
 
     mask[ 4] = !(flags & CALIB_FIX_K1);
     mask[ 5] = !(flags & CALIB_FIX_K2);
+    mask[6]  =
+    mask[7]  = !(flags & CALIB_FIX_TANGENT_DIST);
```
alalek: Looks like this line overrides changes from line 1508, so function behavior is changed.
What is about to guard with "if" condition?
```
if (flags & CALIB_FIX_TANGENT_DIST)
{
    mask[6]  = mask[7]  = 1;
}
```

sovrasov: @alalek fixed.

alalek:
:+1:

jasjuang:
@sovrasov @alalek I didn't realize this PR until now. I tested the CALIB_FIX_TANGENT_DIST but it is not working. I initialized my distortion parameters to `[0, 0, 0, 0, 0]`, and regardless of whether I set my flags to 

``````
cv::CALIB_FIX_K1 + cv::CALIB_FIX_K2 + cv::CALIB_FIX_K3 + cv::CALIB_FIX_TANGENT_DIST
``````
or
``````
cv::CALIB_FIX_K1 + cv::CALIB_FIX_K2 + cv::CALIB_FIX_K3
``````

The outcome is always `[0, 0, 0.003178797775587784, 0.003211421837904917, 0]` after I call `cv::calibrateCamera`

Were you guys able to get this flag to work?

PR #8867, opened 2017-06-07
[GSoC 2017] Photometric Calibration

Yelen719:
### First attempt for PR with dummy function
This pull request changes:
1. Add two new interfaces in modules/calib3d/include/opencv2/calib3d.cpp
2. Add a new file modules/calib3d/src/photo_calbr_nonpara.cpp which implements the the dummy function
3. Add the reference of the paper from J. Engel et al. in doc/opencv.bib

@grace-

### DIFF
--- a/doc/opencv.bib
+++ b/doc/opencv.bib
@@ -904,3 +904,12 @@
   year = {2017},
   organization = {IEEE}
 }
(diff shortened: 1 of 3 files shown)


alalek:
https://github.com/opencv/opencv_contrib/pull/1219

PR #8868, opened 2017-06-07
build: fix build for MSVS2010-2013, remove softfloat from default scope

alalek:
"softfloat" is new experimental feature, so no need to force it as part of public API.


alalek:
:+1:

PR #8874, opened 2017-06-08
Enforce --std=c++98 flags for NVCC CUDA compiler

archenroot:
### This pull request changes

This fixes incompatibility of CUDA NVCC compiler with GCC 6.* and later by adding std++98 flag, CUDA NVCC compiler (8.0) doesn't support in the moment c++11 and later.

### DIFF
--- a/cmake/OpenCVDetectCUDA.cmake
+++ b/cmake/OpenCVDetectCUDA.cmake
@@ -123,7 +123,7 @@ if(CUDA_FOUND)
   endif()
 
   # NVCC flags to be set
-  set(NVCC_FLAGS_EXTRA "")
+  set(NVCC_FLAGS_EXTRA "--std=c++98")
 
   # These vars will be passed into the templates
   set(OPENCV_CUDA_ARCH_BIN "")


alalek:
This is very strange, because almost all distributions doesn't force C++11 mode for GCC. It is usually c++98 or gnu++98.

Your patch doesn't pass [CUDA builder](http://pullrequest.opencv.org/buildbot/builders/precommit_cuda_linux64/builds/10070) compilation.

You can try to add this flag for your custom build environment via [`CUDA_NVCC_FLAGS`](https://github.com/opencv/opencv/blob/3.2.0/cmake/FindCUDA.cmake#L86) CMake parameter.

archenroot:
Actually on GCC 4.9 or 5.4 all is fine. But when switch to 6.3 or 7. the
compilation always fails on NVCC commands with CUDA only. This is known in
GPU world, I am not 100% sure, this fix is common override.

Actually I am on Gentoo system, so I will create a patch which is applied
before configuration time..

Thanks anyway.

https://devtalk.nvidia.com/default/topic/949770/cuda-8-0rc-supporting-gcc6-/

In simple words, by default they don't work well together.

2017-06-08 17:35 GMT+02:00 Alexander Alekhin <notifications@github.com>:

PR #8877, opened 2017-06-09
photo: fix integer overflow

alalek:
There is no cast to wide integer type:
```
    std::numeric_limits<ST>::max() * std::numeric_limits<ST>::max()
```


mshabunin:
:+1:

PR #8880, opened 2017-06-09
fix clang/c2 error "Unexpected atomic instruction -- use Windows inte…

fo40225:
fix clang/c2 build fail "Unexpected atomic instruction -- use Windows interlock intrinsics"

ref:

https://blogs.msdn.microsoft.com/vcblog/2015/12/04/clang-with-microsoft-codegen-in-vs-2015-update-1/

### DIFF
--- a/modules/core/include/opencv2/core/operations.hpp
+++ b/modules/core/include/opencv2/core/operations.hpp
@@ -54,7 +54,7 @@
 /////// exchange-add operation for atomic operations on reference counters ///////
 #ifdef CV_XADD
   // allow to use user-defined macro
-#elif defined __GNUC__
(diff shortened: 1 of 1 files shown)


alalek:
This toolset defines _MSC_VER, __GNUC__, _WIN32 and __clang__ together (and doesn't compatible with none of original defines).
I believe currently we haven't any resources to support this toolset. Feel free to create fork with clang/c2 support or try to contribute into https://github.com/Microsoft/opencv (but looks like this fork is not active anymore).

BTW, For unsupported strange cases there is existed code to support custom `CV_XADD` macro - you should provide right value via C++ defines, your custom `CMAKE_CXX_FLAGS`, etc.

Issue #8628, opened 2017-04-21
Documentation error: gpu::reduce always outputs a row vector (OpenCV 2.4)


t-pollington:
**Summary**

The destination vector that [`gpu::reduce()`](http://docs.opencv.org/2.4.8/modules/gpu/doc/matrix_reductions.html#gpu-reduce) outputs is always a row vector. This was discovered when reading [Feng](https://stackoverflow.com/users/2429379/feng) & [jet47](https://stackoverflow.com/users/2451683/jet47)'s [discussion on Stack Overflow](https://stackoverflow.com/questions/21919990/possible-bug-in-opencv-matrix-reduction-of-gpumat/21926689#21926689) and demonstrated by the following tests on a rowsum operation:

**Test 1 (fails)**

```
// CPU init
double alpha = 1.8;
double x[977][1]; // also fill x array with values, lines hidden 
cv::Mat alpha_h(977,60,DataType<double>::type,alpha);
cv::Mat x_h(977,1,DataType<double>::type,x);

// GPU init & upload
gpu::GpuMat alpha_g(977,60,DataType<double>::type);
gpu::GpuMat rowtotal_g(977,1,DataType<double>::type,0.0);
gpu::GpuMat x_g(977,1,DataType<double>::type);
alpha_g.upload(alpha_h); 
x_g.upload(x_h);

// GPU operations
gpu::reduce(alpha_g,rowtotal_g,1,CV_REDUCE_SUM,-1); // rowsum reduction
gpu::add(rowtotal_g,x_g,rowtotal_g); // causes runtime error, see next line

"OpenCV Error: Assertion failed (src2.type() == src1.type() && src2.size() == src1.size()) in add, file /.../opencv-2.4/modules/gpu/src/element_operations.cpp, line 423"
```
This is because the documentation says I should expect `rowtotal_g` to be a 977x1 column vector, however it's really a 1x977 row vector, so [`gpu::add()`](http://docs.opencv.org/2.4.8/modules/gpu/doc/per_element_operations.html#gpu-add) can't add it to the 977x1 `x_g` column vector.

**Test 2 (succeeds)**

But with these code changes making `x_g` a row vector solves it:

```
cv::Mat x_h(1,977,DataType<double>::type,x);
gpu::GpuMat x_g(1,977,DataType<double>::type);
```
so that this line now runs fine

`gpu::add(rowtotal_g,DKAOarraycorr_g,rowtotal_g);
`

[...]

PR #8886, opened 2017-06-10
Gpu reduce doc

krishraghuram:
### This pullrequest changes
Fixes issue #8628 
Modifed the doc of gpu::reduce(only in 2.4 docs) as follows,
- "vec – Destination vec. Its size and type is defined by dim and dtype parameters." to
  "vec – Destination row vector. Its type is defined by dtype parameter."

- "0 means that the matrix is reduced to a single row. 1 means that the matrix is reduced to a single column." to
  "0 means that the matrix is reduced to a single row(of length equal to number of matrix columns). 1 means that the matrix is reduced to a single column(of length equal to the number of matrix rows). In either case, the output is always stored as a row vector of appropriate length."

- "performing the specified operation on the vectors until a single row/column is obtained" to
  "performing the specified operation on the vectors until a single column/row is obtained. However, the result is always stored as a row vector."


krishraghuram:
Sorry, made PR to master instead of 2.4. Closing it myself.

PR #8887, opened 2017-06-10
Modified doc for gpu::reduce(fixes issue 8628)

krishraghuram:
### This pullrequest changes
Fixes issue #8628 
Modified doc of gpu::reduce as follows(only for 2.4 docs, not for master)
- "vec – Destination vec. Its size and type is defined by dim and dtype parameters." to
"vec – Destination row vector. Its type is defined by dtype parameter."

- "0 means that the matrix is reduced to a single row. 1 means that the matrix is reduced to a single column." to
"0 means that the matrix is reduced to a single row(of length equal to number of matrix columns). 1 means that the matrix is reduced to a single column(of length equal to the number of matrix rows). In either case, the output is always stored as a row vector of appropriate length."

- "performing the specified operation on the vectors until a single row/column is obtained" to
"performing the specified operation on the vectors until a single column/row is obtained. However, the result is always stored as a row vector."

### DIFF
--- a/modules/gpu/doc/matrix_reductions.rst
+++ b/modules/gpu/doc/matrix_reductions.rst
@@ -186,9 +186,9 @@ Reduces a matrix to a vector.
 
     :param mtx: Source 2D matrix.
 
-    :param vec: Destination vector. Its size and type is defined by  ``dim``  and  ``dtype``  parameters.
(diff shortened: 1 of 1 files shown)


alalek:
Thank you! :+1:

PR #8890, opened 2017-06-11
cleanup: remove jinja2 source code

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1220

It used for Matlab binding only (checks added).


vpisarev:
:+1:

PR #8895, opened 2017-06-12
Add Python Docstrings From Doxygen Comments

lewisjb:
### This pull request changes:
The automatic generation of Python bindings.
Specifically, it parses the Doxygen comments; and adds them as the docstrings for python.

This allows OpenCV Python users to see information about a function without having to use Google:
```
>>> help(cv2.copyMakeBorder)
```
previously, this would just show different ways of calling this function; but now will show that _and_ [the Doxygen comments](http://docs.opencv.org/master/d2/de8/group__core__array.html#ga2ac1049c2c3dd25c2b41bffe17658a36).

There is the slight issue that the comments use C++ examples; but if the user were to Google the function, they would get that anyway.

Another benefit to this is that it could be used to help make Python-specific docs.

### Notable code changes
Previously, a declaration generated by hdr_parser.py was 4 elements long, with an undocumented optional 5th (original return type). I have changed this so that it is always 6 elements long, with no optionals, to keep things simple.

### Current issues
_List of issues I have found and am currently working on fixing_
None, if you find any issues, please leave a comment

### DIFF
--- a/modules/python/src2/gen2.py
+++ b/modules/python/src2/gen2.py
@@ -405,7 +405,9 @@ class FuncVariant(object):
         self.name = self.wname = name
         self.isconstructor = isconstructor
 
-        self.rettype = decl[4] if len(decl) >=5 else handle_ptr(decl[1])
+        self.docstring = decl[5]
+
+        self.rettype = decl[4] or handle_ptr(decl[1])
         if self.rettype == "void":
             self.rettype = ""
         self.args = []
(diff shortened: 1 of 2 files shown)


mshabunin:
@lewisjb , looks great! Do you plan to add support for [other doxygen comments](http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html)?
```.cpp
/*! doc */
//! doc
/// doc
some_identifier //!< doc
```

lewisjb:
@mshabunin Thanks! To be honest, I had forgotten to check if they were being used; I will add support for them now.

mshabunin:
:+1:

PR #8899, opened 2017-06-12
core: suppress unreachable code warning

### LINKED ISSUE #8898: unreachable code warnings happen with BASELINE=FP16
- OpenCV => master ( 125abe2fe40a5a249a9432f1e9d99ccf2326e784 )
- Operating System / Platform => Windows 7 Professional
- Compiler => Visual Studio 2012 update 5

##### Detailed description
 - the dispatch code will call the optimized version under certain situation
 - for example, ```cvtScaleHalf_``` function will be like following under ```dispatch mode``` after preprocess
```.cpp
template<> void
cvtScaleHalf_<float, short>( const float* src, size_t sstep, short* dst, size_t dstep, Size size )
{
    if(cv::checkHardwareSupport(CV_CPU_FP16))
        return cvtScaleHalf_SIMD32f16f, (src, sstep, dst, dstep, size);

    sstep /= sizeof(src[0]);
    dstep /= sizeof(dst[0]);

    for( ; size.height--; src += sstep, dst += dstep )
    {
        for ( int x = 0; x < size.width; x++ )
        {
            dst[x] = convertFp16SW(src[x]);
        }
    }
}
```
 - this makes sense
 - same function becomes following under ```non-dispatch mode```
```.cpp
template<> void
cvtScaleHalf_<float, short>( const float* src, size_t sstep, short* dst, size_t dstep, Size size )
{
    return (cvtScaleHalf_SIMD32f16f(src, sstep, dst, dstep, size));
    /**** under this line will be never reached ****/
    sstep /= sizeof(src[0]);
    dstep /= sizeof(dst[0]);

    for( ; size.height--; src += sstep, dst += dstep )
    {
        for ( int x = 0; x < size.width; x++ )
        {
            dst[x] = convertFp16SW(src[x]);
        }
    }
}
```
 - I'm sure that this isn't the only place, but it could happen in many places/situations
 - For FP16, I'll make a PR later

##### Steps to reproduce
 - build OpenCV with ```CPU_BASELINE=FP16```

tomoaki0705:
### This pullrequest changes
 - resolves #8898 
 - there might be a better way to fix, and I'm happy to have discussion

### DIFF
--- a/modules/core/src/convert.cpp
+++ b/modules/core/src/convert.cpp
@@ -4582,6 +4582,7 @@ cvtScaleHalf_<float, short>( const float* src, size_t sstep, short* dst, size_t 
 {
     CV_CPU_CALL_FP16(cvtScaleHalf_SIMD32f16f, (src, sstep, dst, dstep, size));
 
+#if !defined(CV_CPU_COMPILE_FP16)
     sstep /= sizeof(src[0]);
     dstep /= sizeof(dst[0]);
 
@@ -4592,6 +4593,7 @@ cvtScaleHalf_<float, short>( const float* src, size_t sstep, short* dst, size_t 
             dst[x] = convertFp16SW(src[x]);
         }
     }
+#endif
 }
 
 template<> void
@@ -4599,6 +4601,7 @@ cvtScaleHalf_<short, float>( const short* src, size_t sstep, float* dst, size_t 
 {
     CV_CPU_CALL_FP16(cvtScaleHalf_SIMD16f32f, (src, sstep, dst, dstep, size));
 
+#if !defined(CV_CPU_COMPILE_FP16)
     sstep /= sizeof(src[0]);
     dstep /= sizeof(dst[0]);
 
@@ -4609,6 +4612,7 @@ cvtScaleHalf_<short, float>( const short* src, size_t sstep, float* dst, size_t 
             dst[x] = convertFp16SW(src[x]);
         }
     }
+#endif
 }
 
 #ifdef HAVE_OPENVX


alalek:
@tomoaki0705 Condition should be changed. Take a look on dispatch macro [sources](https://github.com/opencv/opencv/blob/772a8188c08f12182328835026cf7faa7bbff653/modules/core/include/opencv2/core/cv_cpu_helper.h#L99-L101).

Could you check this code:
```
#if CV_CPU_HAS_SUPPORT_FP16 != 1
...
#endif
```

BTW, This approach will not work well with multiple dispatched entries - for example: AVX2 + AVX + SSE3 + generic with CPU_BASELINE=AVX2. We can't "drop" AVX code here, because it will break dispatched function with this pattern: AVX+SSE3+generic.

tomoaki0705:
Thank you for the comment @alalek 
I fixed quickly for the build error.

For this approach, you're right.  It doesn't work for other combinations. (at least it works for this specific situation)


I was thinking of "cascaded approach" as following.

```.cpp
void function(Mat& img, int param1, double param2)
{
    int x = 0;
    x = CV_CPU_CALL_AVX2(function_opt_avx2, (img, param1, param2, x));
    x = CV_CPU_CALL_AVX(function_opt_avx, (img, param1, param2, x));
    x = CV_CPU_CALL_SSE4(function_opt_avx, (img, param1, param2, x));

    // generic part
    for( ; x < img.width ; x++ )
    {
        // SW implementation
    }
}
```
each optimized function will receive the beginning of horizontal start point ```x``` and returns how much it processed.
The basic idea are already implemented in some functions.
In this example, the last generic part can exist without ```ifdef``` tricks.
Current helpers (```CV_CPU_CALL_AVX``` and others) directly ```return```s, so this approach requires more than a little modification, but I think this approach can help, too.

alalek:
This would work (probably with other macro name: `CV_CPU_PROCESS_AVX2`), but there it another problem here: it blocks compiler "dead code elimination".

tomoaki0705:
@alalek, thanks for your review.
I think, this should work.

```.cpp
void function(Mat& img, int param1, double param2)
{
    int x = 0;
    CV_CPU_CALL_AVX2(function_opt_avx2, (img, param1, param2, x));
    CV_CPU_CALL_AVX(function_opt_avx, (img, param1, param2, x));
    CV_CPU_CALL_SSE4(function_opt_sse4, (img, param1, param2, x));

    // generic part
    for( ; x < img.width ; x++ )
    {
        // SW implementation
    }
}
```
```.cpp
void function_opt_avx2(Mat& img, int param1, double param2, int& x0)
{
    int x = 0;
    for( int y = 0 ; y < img.rows ; y++ )
    {
        for( x = x0 ; x <= img.width-32 ; x++ )
        {
        }
    }
    x0 = x;
}
```

Probably, this won't cause a dead code elimination.
The variable ```x``` is cascaded, so compiler will not remove the code.

tomoaki0705:
@alalek, I was thinking to wait for this PR to be merged, and then work on the "cascaded version".

But I had some time to implement the and push the cascaded version.
How do you think about the cascaded approach ? (If it passes the build test)

alalek:
Patch looks good. Thank you!

Modern compilers are able to generate more efficient code even for generic C++ loops (without explicit intrinsics) with enabled AVX code generation.

So, dead code elimination is important thing - it prevents binaries size exploding.
In the cascaded approach we probably need to "cut" code via some "#if" statements.

BTW, Don't try to pass or use OpenCV datatypes into "optimized" code (like cv::Mat, InputArray, Size, etc) - to prevent unintended mix of AVX/SSE code with different instructions level (linker will leave only one "sample" of non-inlined functions/methods). It is quite dangerous. We need some additional work to allow this. The same note for std::vector and other "std" types - currently only C++ types are safe (all code is inlined).

tomoaki0705:
Thank you for the suggestion.

I think I started to catch up what you mean.

Standing at your point of view, cascaded version seems too big modification for just this FP16 fix.
I'd like to roll back this PR to previous ```ifdef``` approach.

alalek:
:+1:

BTW, saved cascaded approach [link](https://github.com/tomoaki0705/opencv/commit/e7161d4484ffd94afc72acbe30dce63103d4f4c1)

PR #8903, opened 2017-06-13
fixing models to resolve XML violation issue

### LINKED ISSUE #8894: Some XML cascade file have comments which violate XML syntax

- OpenCV => 3.1
- Operating System / Platform => n/a
- Compiler => n/a

##### Detailed description
Several of the cascade.xml files (e.g. data/haarcascades/haarcascade_frontalcatface.xml)
contain improper XML comments in the header:

&lt;!--------------------------------
...
----------------------------------&gt;

This does not cause problems when reading the data using OpenCV, but other readers (e.g. Python ElementTree) will not read these files. This is easily fixed by changing the lines to &lt;!-- and --&gt;.

XML comments start with &lt;!-- and end with --&gt;
The text in between must be at least one character, may not end in a hyphen, and may not contain two consecutive hyphens.

See https://www.w3.org/TR/REC-xml/#sec-comments

[ Some informal references miss this detail -- e.g. https://www.tutorialspoint.com/xml/xml_comments.htm ; not only fails to mention this, actually gives examples which violate it. ]

StevenPuttemans:
resolves #8894
As suggested in issue #8894, this solves issues with XML style violation.

### DIFF
--- a/data/haarcascades/haarcascade_frontalcatface.xml
+++ b/data/haarcascades/haarcascade_frontalcatface.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<!----------------------------------------------------------------------------
+<!--
  A frontal cat face detector using the basic set of Haar features, i.e.
  horizontal and vertical features but not diagonal features.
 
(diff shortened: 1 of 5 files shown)


alalek:
> - A 3rdparty/ippicv/downloads/linux-808b791a6eac9ed78d32a7666804320e/ippicv_linux_20151201.tgz (0)

Please remove this file from PR

StevenPuttemans:
@alalek done! seems I accidently included it. The 3rdparty folder should be in the gitignore file :)

alalek:
These files are moved into `.cache` folder in current master.
`3rdparty/ippicv/downloads/` was used in previous releases.

Current workaround is to create `.gitignore` file with "*" content in `3rdparty/ippicv/downloads/` directory.

alalek:
Thanks! :+1:

PR #8909, opened 2017-06-13
More information from Hough Circles

varunagrawal:
Added extra element to object in circle CvSeq to record the support of each circle for ranking

### This pullrequest changes

The `icvHoughCirclesGradient` function in `modules/imgproc/src/hough.cpp` has been updated to now return the number of supporting pixels as part of each circle object to allow for more analysis of each detected circle that may be required on a per-application basis.

### DIFF
--- a/modules/imgproc/src/hough.cpp
+++ b/modules/imgproc/src/hough.cpp
@@ -1259,7 +1259,7 @@ cvHoughCircles( CvArr* src_image, void* circle_storage,
     if( CV_IS_STORAGE( circle_storage ))
     {
         circles = cvCreateSeq( CV_32FC3, sizeof(CvSeq),
-            sizeof(float)*3, (CvMemStorage*)circle_storage );
+            sizeof(float)*4, (CvMemStorage*)circle_storage );
     }
     else if( CV_IS_MAT( circle_storage ))
     {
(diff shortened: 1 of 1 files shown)


modules/imgproc/src/hough.cpp
```diff
@@ -1210,10 +1210,11 @@ icvHoughCirclesGradient( CvMat* img, float dp, float min_dist,
         // Check if the circle has enough support
         if( max_count > acc_threshold )
         {
-            float c[3];
+            float c[4];
```
sovrasov: These changes have no effect because `cvSeqPush` copies only first 3 elements of `c`. To fix this make sufficient changes [here](https://github.com/opencv/opencv/pull/8909/files#diff-f3240fa1ccc6896a4c9e16db598360f0L1260) and [here](https://github.com/opencv/opencv/pull/8909/files#diff-f3240fa1ccc6896a4c9e16db598360f0L1272).

varunagrawal: @sovrasov thank you for the recommendation. I've updated the code.

savuor:
You'd better add corresponding changes to documentation of `cv::HoughTransform()` about output vector.

vpisarev:
@varunagrawal, thanks for the contribution! this change may break compatibility with existing code because it changes format of the output sequence. It's better to add extra function that does it. Also, it would be nice to get some evidence (e.g. some test cases and the way how you use the support to filter out the false circles) that the addition is useful and not just a temporary thing needed in some research. Feel free to reopen the PR with the requested changes made

PR #8910, opened 2017-06-13
fix the MSER history's size issue

gylns:
### This pullrequest changes

The component's history's size is not be updated correctly. see also #8796.

### DIFF
--- a/modules/features2d/src/mser.cpp
+++ b/modules/features2d/src/mser.cpp
@@ -325,7 +325,7 @@ public:
                 wp.pix0[comp1->tail].setNext(comp2->head);
                 tail = comp2->tail;
             }
-            
+
             size = comp1->size + comp2->size;
             history = comp1->history;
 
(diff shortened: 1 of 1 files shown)


gylns:
@alalek, Here is a image I used to test. 

mser_test

gylns:
I also saved some er images as a comparison. 
[mser_test.zip](https://github.com/opencv/opencv/files/1079873/mser_test.zip)

The images in folder er_branch1 is the newest results. the extreme regions are colorized, and the red color indicates the region is a MSER. The er's gray level and size can be got from the image's name, and the variation is labeled on the bottom of the image.
Comparing the two folders, we can see a MSER with gray level 68 is missed in folder er_branch2, since the variation is computed by wrong region size.

liurs1990:
@alalek  Sorry for replying you so late.
Here is the image I use to test mser.

the mser test code: 
`
    Ptr<MSER> mser = MSER::create(1, 56, (int)(gray.cols * gray.rows * 0.2));
    mser->setPass2Only(true);
    mser->detectRegions(gray, mserContours, boxRects);
`
Before I fixing the history size, the contours size with minArea 56 is smaller than it with minArea 57. It missed some areas.

test1

liurs1990:
@gylns  Could you share your test code? I do not quite understand the "val" and "size".

gylns:
@liurs1990, Here is the test [code](https://gist.github.com/gylns/10ddce082dca917e6d1902272a13092f).
I extracted the code from opencv, and add `drawERs` method to draw the ER history of one seed point (or more points) that is selected by user when the program starts.

sovrasov:
👍

PR #8912, opened 2017-06-14
(C++11) flann: add normal assignment operator for cvflann::any

alalek:
Fix tests on MacOS with enabled C++11 mode (memory corruption).


alalek:
:+1:

PR #8913, opened 2017-06-14
i8792: Proposed (incomplete) implementation for being able to write multipage images.

saskatchewancatch:
This makes the changes to `loadsave.cpp` that give an idea of the direction I plan to take this.

Changes made to `imwrite` and `_imwrite`.


saskatchewancatch:
Any reference to issue 8792 should actually be to issue 8729. Sorry for the confusion.

modules/imgcodecs/src/loadsave.cpp
```diff
@@ -591,32 +591,47 @@ bool imreadmulti(const String& filename, std::vector<Mat>& mats, int flags)
...
+static bool imwrite_( const String& filename, const std::vector<Mat>& img_vec,
                       const std::vector<int>& params, bool flipv )
 {
+    bool isMultiImg = img_vec.size() > 1;
+    std::vector<Mat> write_vec;
     Mat temp;
-    const Mat* pimage = &image;
-
-    CV_Assert( image.channels() == 1 || image.channels() == 3 || image.channels() == 4 );
+    const Mat* pimage = &img_vec[0]; //to suppress compiler warning
```
savuor: Maybe to suppress the warning it would be better to move this variable into the following `for` loop?

modules/imgcodecs/src/grfmt_tiff.cpp
```diff
@@ -565,90 +530,136 @@ bool  TiffEncoder::writeLibTiff( const Mat& img, const std::vector<int>& params)
...
             TIFFClose(pTiffHandle);
             return false;
         }
+
+        if (compression != COMPRESSION_NONE && !TIFFSetField(pTiffHandle, TIFFTAG_PREDICTOR, predictor))
+        {
+            TIFFClose(pTiffHandle);
+            return false;
+        }
+
```
savuor: Do we also need to set fields `TIFFTAG_SUBFILETYPE` (to `FILETYPE_PAGE`) and `TIFFTAG_PAGENUMBER` (with args `currentPageNumber`, `numberOfPages`)?

savuor:
@saskatchewancatch, thank you for your contribution! Can you resolve the conflicts?

saskatchewancatch:
@savuor Thank you for your feedback, I will address them.

modules/imgcodecs/src/grfmt_tiff.cpp
saskatchewancatch: It seems to work fine without these fields. But, I believe these fields will be helpful nonetheless.

vpisarev:
ok, since #10367 has been merged, this one can be closed now. @saskatchewancatch, thank you for the contribution!

PR #6201, opened 2016-03-02
GStreamer: allow reading raw MJPEG data

paroj:

PR #5489, opened 2015-10-11
cap_v4l: implement PROP_CONVERT_RGB

paroj:
allows disabling automatic conversion to RGB for certain formats. If conversion is disabled the returned image just points to the underlying buffer, so no memcpy is performed.
depends on #5474.

PR #8914, opened 2017-06-14
Add gstreamer capture capability for some YUV formats

stnk20:
### This pullrequest changes

Add gstreamer capture capability for some YUV formats:
* UYVY
* YUY2
* YVYU
* NV12
* NV21
* YV12
* I420

This change enables direct YUV capturing and avoids redundant color conversions.
This change works only for gstreamer-1.0, not for gstreamer-0.10.

### DIFF
--- a/modules/videoio/src/cap_gstreamer.cpp
+++ b/modules/videoio/src/cap_gstreamer.cpp
@@ -833,7 +833,7 @@ bool CvCapture_GStreamer::open( int type, const char* filename )
 
     caps = gst_caps_from_string("video/x-raw, format=(string){BGR, GRAY8}; video/x-bayer,format=(string){rggb,bggr,grbg,gbrg}; image/jpeg");
 
-    {
+    if(manualpipeline){
         GstPad* sink_pad = gst_element_get_static_pad(sink, "sink");
         GstCaps* peer_caps = gst_pad_peer_query_caps(sink_pad,NULL);
         if (!gst_caps_can_intersect(caps, peer_caps)) {
--- a/modules/videoio/src/cap_gstreamer.cpp
+++ b/modules/videoio/src/cap_gstreamer.cpp
@@ -287,6 +287,8 @@ IplImage * CvCapture_GStreamer::retrieveFrame(int)
         }
 
         int depth = 3;
+        bool height_extend = false;
+
 #if GST_VERSION_MAJOR > 0
         depth = 0;
         const gchar* name = gst_structure_get_name(structure);
@@ -295,9 +297,16 @@ IplImage * CvCapture_GStreamer::retrieveFrame(int)
         if (!name)
             return 0;
 
-        // we support 3 types of data:
+        // we support 11 types of data:
         //     video/x-raw, format=BGR   -> 8bit, 3 channels
         //     video/x-raw, format=GRAY8 -> 8bit, 1 channel
+        //     video/x-raw, format=UYVY  -> 8bit, 2 channel
+        //     video/x-raw, format=YUY2  -> 8bit, 2 channel
+        //     video/x-raw, format=YVYU  -> 8bit, 2 channel
+        //     video/x-raw, format=NV12  -> 8bit, 1 channel (height is 1.5x larger than true height)
+        //     video/x-raw, format=NV21  -> 8bit, 1 channel (height is 1.5x larger than true height)
+        //     video/x-raw, format=YV12  -> 8bit, 1 channel (height is 1.5x larger than true height)
+        //     video/x-raw, format=I420  -> 8bit, 1 channel (height is 1.5x larger than true height)
         //     video/x-bayer             -> 8bit, 1 channel
         //     image/jpeg                -> 8bit, mjpeg: buffer_size x 1 x 1
         // bayer data is never decoded, the user is responsible for that
@@ -310,6 +319,13 @@ IplImage * CvCapture_GStreamer::retrieveFrame(int)
 
             if (strcasecmp(format, "BGR") == 0) {
                 depth = 3;
+            }
+            else if( (strcasecmp(format, "UYVY") == 0) || (strcasecmp(format, "YUY2") == 0) || (strcasecmp(format, "YVYU") == 0) ){
+                depth = 2;
+            }
+            else if( (strcasecmp(format, "NV12") == 0) || (strcasecmp(format, "NV21") == 0) || (strcasecmp(format, "YV12") == 0) || (strcasecmp(format, "I420") == 0) ){
+                depth = 1;
+                height_extend = true;
             }
             else if(strcasecmp(format, "GRAY8") == 0){
                 depth = 1;
@@ -324,7 +340,11 @@ IplImage * CvCapture_GStreamer::retrieveFrame(int)
         }
 #endif
         if (depth > 0) {
-            frame = cvCreateImageHeader(cvSize(width, height), IPL_DEPTH_8U, depth);
+            if(height_extend){
+                frame = cvCreateImageHeader(cvSize(width, height*3/2), IPL_DEPTH_8U, depth);
+            }else{
+                frame = cvCreateImageHeader(cvSize(width, height), IPL_DEPTH_8U, depth);
+            }
         } else {
             gst_caps_unref(buffer_caps);
             return 0;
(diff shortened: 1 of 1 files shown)


arpu:
any idea wha i get this build error on this branch? 

> `[ 49%] Building CXX object modules/videoio/CMakeFiles/opencv_videoio.dir/src/cap_gstreamer.cpp.o
cc1plus: warning: command line option ‘-Wmissing-prototypes’ is valid for C/ObjC but not for C++
cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++
In file included from /home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/imgproc/include/opencv2/imgproc/types_c.h:46:0,
                 from /home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/imgproc/include/opencv2/imgproc/imgproc_c.h:46,
                 from /home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/videoio/src/precomp.hpp:53,
                 from /home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/videoio/src/cap_gstreamer.cpp:50:
/home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/videoio/src/cap_gstreamer.cpp: In member function ‘virtual bool CvCapture_GStreamer::open(int, const char*)’:
/home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/core/include/opencv2/core/core_c.h:2758:39: error: jump to label ‘exit’ [-fpermissive]
 #define __CV_END__         goto exit; exit: ; }
                                       ^
/home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/videoio/src/precomp.hpp:81:18: note: in expansion of macro ‘__CV_END__’
 #define __END__  __CV_END__
                  ^~~~~~~~~~
/home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/videoio/src/cap_gstreamer.cpp:911:5: note: in expansion of macro ‘__END__’
     __END__;
     ^
/home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/core/include/opencv2/core/core_c.h:2759:33: note:   from here
 #define __CV_EXIT__        goto exit
                                 ^
/home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/core/include/opencv2/core/core_c.h:2723:5: note: in expansion of macro ‘__CV_EXIT__’
     __CV_EXIT__;                                                   \
     ^~~~~~~~~~~
/home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/videoio/src/cap_gstreamer.cpp:779:13: note: in expansion of macro ‘CV_ERROR’
             CV_ERROR(CV_StsError, "GStreamer: cannot find appsink in manual pipeline\n");
             ^
/home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/videoio/src/cap_gstreamer.cpp:837:14: note:   crosses initialization of ‘GstCaps* peer_caps’
     GstCaps* peer_caps = gst_pad_peer_query_caps(sink_pad,NULL);
              ^~~~~~~~~
/home/arpu/Work/githubsources/opencv_gstreamer_yuv/modules/videoio/src/cap_gstreamer.cpp:836:13: note:   crosses initialization of ‘GstPad* sink_pad’
     GstPad* sink_pad = gst_element_get_static_pad(sink, "sink");
             ^~~~~~~~
At global scope:
cc1plus: warning: unrecognized command line option ‘-Wno-unnamed-type-template-args’
modules/videoio/CMakeFiles/opencv_videoio.dir/build.make:158: recipe for target 'modules/videoio/CMakeFiles/opencv_videoio.dir/src/cap_gstreamer.cpp.o' failed
make[2]: *** [modules/videoio/CMakeFiles/opencv_videoio.dir/src/cap_gstreamer.cpp.o] Error 1
CMakeFiles/Makefile2:1648: recipe for target 'modules/videoio/CMakeFiles/opencv_videoio.dir/all' failed
make[1]: *** [modules/videoio/CMakeFiles/opencv_videoio.dir/all] Error 2
Makefile:162: recipe for target 'all' failed
make: *** [all] Error 2
`

stnk20:
I reproduced your build error on Ubuntu 16.04.  @arpu .
This cross initialization error is caused by goto statement used in \_\_END\_\_ macro.
I commited workaround: enclosing definition statements  into block around line 836 :
https://github.com/opencv/opencv/pull/8914/commits/2931d667b4425c556c376115111340eb97d54c5a

arpu:
the build check error does nothing todo with this change right? 
http://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/11229/steps/test_videoio/logs/stdio

alalek:
Right, I have filled an issue: #8992

arpu:
@alalek  will this be merged?

alalek:
More work is required for this patch.

VideoCapture should return BGR (8UC3) image to user by default. Implicit return in other formats should be prohibited ( /cc @vpisarev ).
Some backends return frames in Gray (8UC1) format - but it is not a critical problem for many applications.

User may ask for other formats via setting of these properties (see #5489 /cc @paroj ):
1. turn off `CAP_PROP_CONVERT_RGB` (default value is true, that means BGR)
2. query or select `CAP_PROP_FOURCC` (if required)
3. select `CAP_PROP_FORMAT` of required (for example, to select 16-bit images)

Unfortunately, current support for `video/x-bayer` and `image/jpeg` (#6201 /cc @paroj ) in GStreamer backend is implemented with violation of this VideoCapture interface assumption. So it is not a good example.

P.S. Probably `CAP_PROP_MODE` will be deprecated.

paroj:
the gstreamer backend is a bit special as it allows gstreamer pipelines. Those duplicate the `CONVERT_RGB` "protocol" above. (via autovideoconvert and caps)
Enforcing these two to be in sync has little gain as we can assume people passing pipelines know what they are doing.

To enforce implicit conversion the default caps should be just `"video/x-raw, format=(string){BGR, GRAY8}"` for the default created pipeline.

This patch does not change the situation as it only adds the yuv formats when the default caps fail. This should never happen with the default pipeline as it contains autovideoconvert.

To make things clear you should however just do `if(manualpipeline)` though.

stnk20:
Thanks for your support. @paroj 

> This patch does not change the situation as it only adds the yuv formats when the default caps fail. This should never happen with the default pipeline as it contains autovideoconvert.

This is what I was considering. 
This patch should not affect running legacy code that expecting BGR format. 
I understand that there are some problem in the setProperty/getProperty part, but I was taking a policy to minimize the change.

> To make things clear you should however just do if(manualpipeline) though.

Simple solution. 
Is this enough? @alalek  or should I change setProperty/getProperty function?

alalek:
`if (manualpipeline)` should be fine (because it should not break general behavior of VideoCapture - it is very GStreamer-specific).

alalek:
:+1:

IMG-PRCSNG:
In his [comment](https://github.com/opencv/opencv/pull/8914#issuecomment-312521161) @paroj had mentioned -  
> To enforce implicit conversion the default caps should be just "video/x-raw, format=(string){BGR, GRAY8}" for the default created pipeline.

> This patch does not change the situation as it only adds the yuv formats when the default caps fail. This should never happen with the default pipeline as it contains autovideoconvert.

I noticed a weird issue when using uridecodebin on a Jetson TX2 - it chooses `omxh264dec` as decoder and somehow the image format before the autovideoconvert ended up being `NV12`. And if you see this [line: 860](https://github.com/opencv/opencv/blob/0c07836fbafabbcdb8c3393049d9820184429e45/modules/videoio/src/cap_gstreamer.cpp#L860), only few types are accepted - thus it ended up converting NV12 to GRAY8.  Here is the graphviz png file corresponding to the same. It will be helpful if someone can point out what is going wrong / what is the solution.
dotfile

I am currently hardcoding NV12 into the accepted types. Is this ok?

arpu:
best is to use a costum gstreamer pipe  
and cv::VideoCapture open

IMG-PRCSNG:
Agreed. But if you look at the documentation,
```cpp
/*
 *  When dealing with a file, CvCapture_GStreamer will not drop frames if the grabbing interval
 *  larger than the framerate period. (Unlike the uri or manual pipeline description, which assume
 *  a live source)
 */
```
For a `filesrc` in a custom pipeline, this will be a problem as it will treated as a live source and the frames will be skipped if the grabbing interval is larger than framerate period. Please correct me if I am wrong.

arpu:
sorry i cannot help on this :/ no idea

PR #8307, opened 2017-03-02
Modified code to work with universal build.

AhiyaHiya:
### This pullrequest changes

Modified Obj-C++ code (.mm file), so that it can compile for both 32-bit and 64-bit macOS; in the code's current state, you cannot build a Universal Intel binary for macOS.

PR #7193, opened 2016-08-29
Cleans up declarations in window_cocoa.mm

valeriyvan:
Cleans up declarations in window_cocoa.mm removing unnecessary iVars declarations and changing retain property attribute to strong.

PR #8916, opened 2017-06-14
macOS cocoa window autosize could not access image dimensions

### LINKED ISSUE #8885: OpenCV 3.2 + Python 3.6.1: imshow() not worked even with waitKey(0)
- OpenCV => 3.2
- Operating System / Platform => MacOS Sierra 10.12.5
- Compiler => Terminal

I followed the documentation with: 

```
import cv2
import numpy as np

img = cv2.imread('1.jpg')
print(img)
cv2.imshow('Image', img)
cv2.waitKey(0)
cv2.destroyAllWindows()
```

But I got empty content window:

[Image](https://i.stack.imgur.com/DxsAq.png)

Is this related to OpenCV?

For readability I posted this question to stackoverflow: https://stackoverflow.com/q/44469973/7813604

mattmyne:
resolves #8885 

### This pullrequest changes

Window autosize was not working for macOS using cocoa. The window's image dimensions could not be found. This has been fixed by removing __IP64__ specific synthesize in window_cocoa.mm that was causing null reference for the window's contentView image property in cvShowImage (image reference was not linked to _image). The cause of this was code previously added as part of #8307 to allow 32-bit OSX builds. It is possible similar modifications may be required should functions need to access CVWindow and CVSlider properties.

### DIFF
--- a/modules/highgui/src/window_cocoa.mm
+++ b/modules/highgui/src/window_cocoa.mm
@@ -862,11 +862,7 @@ void cv::setWindowTitle(const String& winname, const String& title)
 @end
 
 @implementation CVView
-#if defined(__LP64__)
-@synthesize image;
-#else // 32-bit Obj-C does not have automatic synthesize
 @synthesize image = _image;
-#endif
 
 - (id)init {
     //cout << "CVView init" << endl;


alalek:
@vpisarev Please take a look.
My suggestion is to revert patch #8307 completely and just drop `.mm` files from 32-bit builds (Apple doesn't support 32-bit ObjC compiler anymore) with corresponding functionality.

alalek:
@mattmyne Thank you for digging into this issue!

mattmyne:
I'd agree @alalek that reverting #8307 would cleanly resolve this too. In the unlikely event someone needs a 32-bit OSX compile they can always reference #8307 as a self-help guide..

mattmyne:
ok, having done some more testing #8307 actually broke not only image display but sliders and pretty much everything else in highgui for the majority of mac users (anyone using 64-bit and cocoa). Not ideal.

I've confirmed reverting #8307 fixes everything for 64-bit cocoa. I've also confirmed reverting #7193 still works for 64-bit cocoa too, so if it was the removal of ivals in #7193 that broke 32-bit cocoa compile then it's an option to revert this too so it works for everyone again. The code will retain some legacy structure but perhaps that's a price worth paying for everyone to be happy. I'd very much recommend a revert to either of these from the current situation.

vpisarev:
@mattmyne, thank you for the investigation! Can you please then prepare the patch that reverts #8307 and a part of #7193 with clear explanation what's happening with links to the history of such patches. We will then point people who try to revert it once again to the relevant discussion? Of course, the best solution would be to introduce some regression tests that would provide automatic protection, but I'm not sure if it's possible to do in this very case

mattmyne:
@vpisarev sure, though I'm not sure which part of #7193 you're referring to? If you want to revert more than just #8307 I assume this is to keep 32-bit OS X compile support so I think I can keep the "changing retain property attribute to strong" aspect of #7193 but restore the ivars? Do let me know if that's what you're after and I'll update this PR.

mattmyne:
Actually it seems there'd be little point to changing to "strong" since this is an ARC property that would be ignored by 32-bit OS X compiler anyway. So you want me to revert all of #7193 too?

mattmyne:
OK, I've submitted this as new pull request #8953 due to it being out-of-scope for this pull request. @vpisarev do let me know if #8953 is ok, and this pull request #8916 can be closed.

PR #8920, opened 2017-06-15
video: add a constructor with API preference parameter for VideoWriter

### LINKED ISSUE #8859: Unable to use gstreamer videowriter
System information (version)

OpenCV => 3.2
Operating System / Platform => Mac OS 10.12.5
Compiler => g++
Wrapper => Java/C++

It is impossible to instantiate a cvCreateVideoWriter_GStreamer because TRY_OPEN detects that a writer has already been created by ffmpeg (or possibly other cap).    FFMpeg jsut creates a file with the name of my gstreamer pipeline construction..

My call to VideoWriter():
writer.open("appsrc ! autovideoconvert ! vtenc_h264 ! h264parse ! mpegtsmux ! rtpmp2tpay ! udpsink host=127.0.0.1 port=5000", fourcc, capture.get(Videoio.CAP_PROP_FPS), frameSize);

I can temporarily fix the issue by moing the HAVE_GSTREAMER check above ffmpeg, but I supose the real fix would be to have ffmpeg fail if it thinks that it should not be a valid file.

sovrasov:
resolves #8859

### DIFF
--- a/modules/videoio/include/opencv2/videoio/videoio_c.h
+++ b/modules/videoio/include/opencv2/videoio/videoio_c.h
@@ -124,7 +124,8 @@ enum
     CV_CAP_FFMPEG = 1900,    // FFMPEG
     CV_CAP_IMAGES = 2000,    // OpenCV Image Sequence (e.g. img_%02d.jpg)
 
-    CV_CAP_ARAVIS = 2100     // Aravis GigE SDK
+    CV_CAP_ARAVIS = 2100,     // Aravis GigE SDK
+    CV_CAP_OCV_MJPEG = 2200   // Built-in MotionJPEG codec
 };
 
 /** @brief start capturing frames from camera: index = camera_index + domain_offset (CV_CAP_*)
--- a/modules/videoio/src/cap.cpp
+++ b/modules/videoio/src/cap.cpp
@@ -371,47 +371,59 @@ CV_IMPL CvCapture * cvCreateFileCapture (const char * filename)
 CV_IMPL CvVideoWriter* cvCreateVideoWriter( const char* filename, int fourcc,
                                             double fps, CvSize frameSize, int is_color )
 {
-    // If none of the writers is used
-    // these statements suppress 'unused parameter' warnings.
+    return cvCreateVideoWriterWithPreference(CV_CAP_ANY, filename, fourcc, fps, frameSize, is_color);
+}
+
+CvVideoWriter* cvCreateVideoWriterWithPreference(int apiPreference, const char* filename, int fourcc,
+                                            double fps, CvSize frameSize, int is_color )
+{
     CV_UNUSED(frameSize);
     CV_UNUSED(is_color);
-
-    //CV_FUNCNAME( "cvCreateVideoWriter" );
 
     CvVideoWriter *result = 0;
 
     if(!fourcc || !fps)
         TRY_OPEN(result, cvCreateVideoWriter_Images(filename))
 
-#ifdef HAVE_FFMPEG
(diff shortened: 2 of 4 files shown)


modules/videoio/include/opencv2/videoio/videoio_c.h
```diff
@@ -560,6 +560,14 @@ CVAPI(CvVideoWriter*) cvCreateVideoWriter( const char* filename, int fourcc,
                                            double fps, CvSize frame_size,
                                            int is_color CV_DEFAULT(1));
 
+/**
+
+*/
+CVAPI(CvVideoWriter*) cvCreateVideoWriterWithPreference(int api, const char* filename, int fourcc,
+                                           double fps, CvSize frame_size,
+                                           int is_color CV_DEFAULT(1));
```
alalek: Probably we should not add new exported `cv*` functions. Please make it local "static".

sovrasov: I've moved it to the `precomp.hpp` where other similar functions are declared.

modules/videoio/src/cap.cpp
```diff
@@ -371,45 +371,56 @@ CV_IMPL CvCapture * cvCreateFileCapture (const char * filename)
...
+{
     CV_UNUSED(frameSize);
     CV_UNUSED(is_color);
 
-    //CV_FUNCNAME( "cvCreateVideoWriter" );
-
     CvVideoWriter *result = 0;
 
     if(!fourcc || !fps)
         TRY_OPEN(result, cvCreateVideoWriter_Images(filename))
```
mshabunin: Maybe we could define additional API IDs for built-in MotionJPEG and Images backends? Or maybe we should add new set of IDs for _VideoWriter_ specifically: _WRI_ANY_, _WRI_FFMPEG_, etc.? What do you think?

modules/videoio/src/cap.cpp
```diff
@@ -371,45 +371,56 @@ CV_IMPL CvCapture * cvCreateFileCapture (const char * filename)
...
-
-#ifdef HAVE_GSTREAMER
-    TRY_OPEN(result, cvCreateVideoWriter_GStreamer(filename, fourcc, fps, frameSize, is_color))
-#endif
+    switch(apiPreference)
+    {
+        default:
+            //exit if the specified API is unavaliable
+            if (apiPreference != CV_CAP_ANY) break;
+        case(CV_CAP_FFMPEG):
```
alalek: BTW, Why "case" values are in brackets?

alalek:
Looks good to me. Well done! :+1:

modules/videoio/src/cap.cpp
mshabunin: In my opinion wrapping whole _case_ blocks into _#ifdef..#endif_ will look more clean:
```.cpp
#ifdef HAVE_FFMPEG
        case CV_CAP_FFMPEG:
            TRY_OPEN(result, cvCreateVideoWriter_FFMPEG_proxy (filename, fourcc, fps, frameSize, is_color))
            if (apiPreference != CV_CAP_ANY) break;
#endif

#ifdef HAVE_VFW
        case CV_CAP_VFW:
            TRY_OPEN(result, cvCreateVideoWriter_VFW(filename, fourcc, fps, frameSize, is_color))
            if (apiPreference != CV_CAP_ANY) break;
#endif
```

sovrasov: Don't know, just a habit.

modules/videoio/src/cap.cpp
sovrasov: I've defined `CAP_OCV_MJPEG` and we already have `CAP_IMAGES`.

mshabunin:
:+1:

PR #8925, opened 2017-06-15
SoftFloat: fixed constants, warnings, licence

savuor:
### This pullrequest changes
* Warning suppression for MSVC removed (implicit cast produced this warning, now the cast is explicit)
* Licence fixed in 2 files
* Float-point constants replaced by their hex representation


alalek:
:+1:

PR #8934, opened 2017-06-16
Refactor Python Classes

lewisjb:
_"Refactor Python Classes" isn't the best description, but couldn't think of a better one_
### This pull request changes the automatic python bindings
The purpose of this PR is to make the Python classes of OpenCV types more 'pythonic'.
Currently, custom types are instantiated via a constructor function of the same name as the type, and any class methods are given the name in the format: `<type_name>_<function_name>`.
Example:
```
>>> cv2.KeyPoint
<built-in function KeyPoint>
>>> type(cv2.KeyPoint(1,1,1))
<type 'cv2.KeyPoint'>
>>> cv2.KeyPoint_overlap
<built-in function KeyPoint_overlap>
>>> cv2.KeyPoint.overlap
  File "<stdin>", line 1, in <module>
AttributeError: 'builtin_function_or_method' object has no attribute 'overlap'
```
As you can see above, `cv2.KeyPoint` is actually a constructor function that returns an instance of a type (with the same name). And the classmethods are only accessible via the underscore convention.

This pull-request changes this so custom types are accessible, the constructor functions are the custom types' `__init__`s, and classmethods are also members of the type (the underscore convention is still usable due to backwards compatibility).

Example:
```
>>> cv2.KeyPoint
<type 'cv2.KeyPoint'>
>>> cv2.KeyPoint.overlap
<built-in method overlap of type object at 0xdeadbeef>
>>> cv2.KeyPoint_overlap
<built-in function KeyPoint_overlap>
```
Due to these changes, you can now inherit from these custom types:
```python
class MyKeyPoint(cv2.KeyPoint)
  def __init__(self, x, y):
    super(MyKeyPoint, self).__init__(x, y, 5)

x = MyKeyPoint(1, 2)
y = cv2.KeyPoint(1, 1, 3)
print cv2.KeyPoint.overlap(x, y)
```
_(Not a very practical example)_


berak:
quick question: would that also apply to `classx::create()` -> `classx_create()` functions ?

i sincerely like your idea, but this would break **everyone's** current code, all docs & tutorials.

lewisjb:
@berak Thanks for your feedback!

Yes, this change does apply to `classx::create()` functions; however, as far as I can tell, it won't break any existing code, docs, or tutorials (it would be nice if docs and tutorials changed to use this method though).
Even though you can now do `classx.create()`, you can _still_ do `classx_create()`.

Example:
```
>>> cv2.FastFeatureDetector.create()
<FastFeatureDetector 0xdeadbeef>
>>> cv2.FastFeatureDetector_create()
<FastFeatureDetector 0xcafebabe>
```

This change should _not_ affect existing code, but should allow new ways of writing that code.

sovrasov:
:+1:

PR #8939, opened 2017-06-18
Synthesize & property & explicit property declaration are referring p…

hwiesmann:
Synthesize & property & explicit property declaration are referring partially to different attributes

To solve this issue the @synthesize statement should always be used.


alalek:
Related #8916

alalek:
@hwiesmann Thanks for your effort!
This PR is replaced by #8953.

PR #8940, opened 2017-06-18
batch-oriented mutex locking in haar detector parallel loop body

678098:
### This pullrequest changes

In haar detector parallel loop body insertions into resulting vectors are made in batches.

### DIFF
--- a/modules/objdetect/src/haar.cpp
+++ b/modules/objdetect/src/haar.cpp
@@ -1281,6 +1281,8 @@ cvRunHaarClassifierCascade( const CvHaarClassifierCascade* _cascade,
 namespace cv
 {
 
+const size_t PARALLEL_LOOP_BATCH_SIZE = 100;
+
 class HaarDetectObjects_ScaleImage_Invoker : public ParallelLoopBody
 {
 public:
@@ -1318,6 +1320,10 @@ public:
 
         Size ssz(sum1.cols - 1 - winSize0.width, y2 - y1);
         int x, y, ystep = factor > 2 ? 1 : 2;
+
+        std::vector<Rect> vecLocal;
+        std::vector<int> rejectLevelsLocal;
+        std::vector<double> levelWeightsLocal;
 
 #ifdef HAVE_IPP
         if(CV_IPP_CHECK_COND && cascade->hid_cascade->ipp_stages )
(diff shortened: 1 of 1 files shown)


sovrasov:
Nice trick! :+1:

sovrasov:
@678098 unfortunately we have some problems with parallel frameworks.
In this example
```c++
cv::parallel_for_(cv::Range(0, 128), TestBody());
```
`operator()` of the `TestBody()` will be called 100 times with ranges (0,1), (1,2), ... , (99,100),
so your optimization will drastically decrease performance.

678098:
It's not transparent how initial range will be split (I looked into parallel.cpp). If only we could know how many subranges will be created (and their unique ids, such as start/end), we could allocate enough vectors beforehand without mutexes at all.

I will think about that and will write performance tests.

678098:
I made proxy class over vector, which implements push_back operation with or without mutex inside.

There are 2 new classes now: ParallelAccessVectorProxy (with or without mutex) and overall HaarParallelResultsStorage, which can construct ParallelAccessVectorProxy.

I made delayed initialization (with mutex) of  ParallelAccessVectorProxy with first detection, so the chance of multiple threads locking on initialization is not higher than chance of locking on first detection before.

HaarParallelResultsStorage is not using map as in 1st commit. Instead It contains 2 vectors and 1 vector of vectors:
1)One vector is used for ending subrange.
2)Vector of vectors is initializing on first detection, and Its size is predicted in suggestion of uniform range splitting. Positions of subvectors are calculated in suggestion of uniform range splitting too. If subrange starts not as we expected, we use following:
3)One vector is used in every unpredicted situation like before (with mutex).

vpisarev:
@678098, thank you for the contribution! I don't think we need parallel vector class inside OpenCV. Just do the merging of per-thread detections into the single output vector in a batch fashion, like after every 100 detections or so (and in the end as well). We used such trick in the CascadeClassifier and it works really well, no need to introduce extra generic classes that we don't want to support

678098:
Ok, will do, if you say so. It even makes things simplier.

678098:
@vpisarev, done.
What about performance tests? Do I need to make them?

modules/objdetect/src/haar.cpp
```diff
@@ -1472,13 +1521,27 @@ class HaarDetectObjects_ScaleCascade_Invoker : public ParallelLoopBody
...
+                        mtx->unlock();
+
+                        vecLocal.clear();
+                    }
                 }
                 ixstep = result != 0 ? 1 : 2;
             }
         }
+
+        if (vecLocal.size())
```
vpisarev: implicit conversion of size_t to bool is fine in C/C++, but it would be more stylish to use `!vecLocal.empty()` instead

vpisarev:
@678098, sorry for a looong delay, it has been release preparations and many other activities. Formal performance tests are not necessary, I think. But it would be nice if you publish the results that you obtained on your machine in different scenarios (no face, single face, many faces (10+ or 100+ or whatever)).

678098:
@vpisarev thanks for advice, I will get rid of implicit conversion.

vpisarev:
:+1:

PR #8942, opened 2017-06-19
Fixed gray window for gpu stereo BP and CSBP

zwsu:
### This pullrequest changes

compute() for BP and CSBP output 32-bit floating-point mat, and in cv::imshow() 32-bit floating-point is recognized as [0,1] and maped to [0,255], that causes gray window for BP and CSBP.


alalek:
:+1:

PR #8950, opened 2017-06-21
cmake: guard scanning of default MKL system-wide paths

alalek:
- `WITH_MKL` option is enabled
- user doesn't specify `MKLROOT`/`MKL_ROOT_DIR` variables

MKL introduces unwanted huge impact on regular static builds ([2 hours of compilation on i7+SSD](http://pullrequest.opencv.org/buildbot/builders/master-win64-vc14-static/builds/10131)).
Also we have active workaround for "WinPack" about MKL turning off.

### DIFF
--- a/cmake/OpenCVFindMKL.cmake
+++ b/cmake/OpenCVFindMKL.cmake
@@ -7,10 +7,10 @@
 #
 # On return this will define:
 #
-# HAVE_MKL          - True if Intel IPP found
-# MKL_ROOT_DIR      - root of IPP installation
-# MKL_INCLUDE_DIRS  - IPP include folder
-# MKL_LIBRARIES     - IPP libraries that are used by OpenCV
+# HAVE_MKL          - True if Intel MKL found
+# MKL_ROOT_DIR      - root of MKL installation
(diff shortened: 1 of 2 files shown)


cmake/OpenCVFindMKL.cmake
```diff
@@ -46,23 +46,26 @@ if(NOT DEFINED MKL_USE_MULTITHREAD)
...
-    set(mkl_root_paths ${MKL_ROOT_DIR})
+if(NOT MKL_ROOT_DIR OR NOT EXISTS "${MKL_ROOT_DIR}/include/mkl.h")
+    set(mkl_root_paths "${MKL_ROOT_DIR}")
     if(DEFINED ENV{MKLROOT})
-        list(APPEND mkl_root_paths $ENV{MKLROOT})
+        list(APPEND mkl_root_paths "$ENV{MKLROOT}")
     endif()
-    if(WIN32)
+
+    if(WITH_MKL AND NOT mkl_root_paths)
```
alalek: Main change is here

vpisarev:
yes, this makes sense, since now dnn module does not rely (performance-wise) on an external BLAS/Lapack. :+1:

PR #8959, opened 2017-06-21
(2.4) .gitignore: add .cache directory

alalek:
to be consistent with master branch and prevent unintended removals of download cache.


mshabunin:
:+1:

PR #8960, opened 2017-06-21
Catch SkipTestException in performance tests

dkurt:
### This pullrequest changes

Skip performance tests when `SkipTestException` have thrown.


alalek:
Thanks! :+1:

PR #8970, opened 2017-06-22
(2.4) cmake: fix libname for pkg-config configuration

alalek:


bkuhls:
Thanks, this patch fixes the problem I reported in https://github.com/opencv/opencv/pull/8611

alalek:
:+1:

PR #8975, opened 2017-06-23
core: forbid conversion real->int in some cases in FileStorage

sovrasov:


modules/core/include/opencv2/core/cvstd.inl.hpp
```diff
@@ -156,9 +156,7 @@ FileNode::operator std::string() const
 template<> inline
 void operator >> (const FileNode& n, std::string& value)
 {
-    String val;
-    read(n, val, val);
-    value = val;
+    read(n, value, std::string());
```
PkLab: should be managed the case `#ifndef OPENCV_NOSTL` has header

sovrasov: Thanks, I've forgot to check this.

modules/core/src/persistence.cpp
```diff
@@ -7375,28 +7375,35 @@ void read(const FileNode& node, int& value, int default_value)
 {
     value = !node.node ? default_value :
     CV_NODE_IS_INT(node.node->tag) ? node.node->data.i :
-    CV_NODE_IS_REAL(node.node->tag) ? cvRound(node.node->data.f) : 0x7fffffff;
+    CV_NODE_IS_REAL(node.node->tag) ? std::numeric_limits<int>::max() : std::numeric_limits<int>::max();
```
terfendail: Looks like INT_MAX will be returned in both cases. If this is intended why we even bother to check whether node is REAL?

sovrasov: You're right, thanks.

alalek:
:+1:

PR #7242, opened 2016-09-06
[2.4] types_c.h: Fix compiling VFP assembler code

bkuhls:
Replace asm by __asm__ according to
https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html#Alternate-Keywords
as suggested by Arnout Vandecappelle:
http://lists.busybox.net/pipermail/buildroot/2016-September/171491.html

to fix build errors in ffmpeg with opencv2 support detected by
buildroot autobuilders:
http://autobuild.buildroot.net/results/c32/c32a21240a9933796ee850349a62ff3c2314f25c/build-end.log

Signed-off-by: Bernd Kuhls <bernd.kuhls@t-online.de>

PR #8976, opened 2017-06-23
fast_math.hpp: Use __asm__ rather than asm; fixes including with -std=c99

jrtc27:
This is very similar to #7242; MRPT was failing to build on armhf, as it is compiled with -std=c99.


alalek:
:+1:

Issue #8974, opened 2017-06-23
build: AVX512 support


alalek:
Currently AVX512 support is broken and builds fail with enabled compiler option. Message is:
```
opencv/modules/core/src/system.cpp(488): error: identifier "CV_CPU_AVX512" is undefined
          int baseline_features[] = { CV_CPU_BASELINE_FEATURES };
```

On AVX512 capable machine:
```
cmake -DCPU_BASELINE=DETECT <...>
```
or on other x86 machine:
```
cmake -DCPU_BASELINE=AVX512 <...>
```

---

## Problem explanation

Current CMake scripts are designed for support "atomic" instruction sets (like, AVX512DQ) which has 1:1 mapping to compiler flags / processor features.

Groups, like AVX512 are not properly supported for now. Perhaps AVX512 can be replaced with AVX512-KNL (Knights Landing), AVX512-SKL (Skylake with BW, DQ, VL), AVX512-CNL (CannonLake with additional IFMA, VBMI) groups. But I'm not sure that it is a good idea.
BTW, Intel compiler defined this flag "COMMON-AVX512", "MIC-AVX512", "CORE-AVX512" (DQ is here)

So probably we should start from atomic features. Groups can be added later.

alalek:
CMake: `-DCPU_BASELINE=AVX512_SKX`
OpenCV runtime (dispatched): `OPENCV_CPU_DISABLE=AVX512-SKX`

PR #8986, opened 2017-06-25
build: disable AVX512

alalek:
Currently it is not supported (and actually it is not used in OpenCV).
All builds are broken with enabled AVX512 option or with `march=native` option on AVX512 capable CPU.

Workaround for #8974


mshabunin:
:+1:

PR #8352, opened 2017-03-10
core(test): InputArray invariant problem

alalek:
**WIP**

Regression test only.
There is no fix yet.

Related #8341 #8149

PR #8988, opened 2017-06-26
core: forbid handling of the case when src=dst in cv::repeat

### LINKED ISSUE #8972: Bug: Memory corruption in cv::repeat when src = dst
Original Post: http://answers.opencv.org/question/161823/cvrepeat-after-update-24-32-bug-or-feature/
------
#### 1.
- OpenCV => 3.2 (Compiled directly out of Git-Tag 3.2 without standard config)
- Operating System / Platform => Ubuntu 14.04 64 Bit
- Compiler => c++ (Ubuntu 5.4.1)

#### 2.
- OpenCV => 3.2 (Homebrew)
- Operating System / Platform =>OSX 10.11.6
- Compiler => Apple LLVM version 8.0.0 (clang-800.0.42.1)

### Detailed description

Consider this Code:
```
std::vector<int> lin = {1,2,3,4,5};
cv::Mat linmat = cv::Mat(lin).reshape(0,1);
```

At this point, we should have the Mat
````
[1,2,3,4,5]
````

So far so good. When applying `cv::repeat` to it in 2.4 like
```
cv::repeat(linmat,5,1,linmat);
```
The result in 2.4.8 was (like it should be):
```
    [1,2,3,4,5,
     1,2,3,4,5,
     1,2,3,4,5,
     1,2,3,4,5,
     1,2,3,4,5]
```
But **in 3.2, some memory-corruption happens when src = dst**. The result is some memory-garbage.
When I use a second `Mat` as dst, everything works fine.
Is this supposed to be like this or is this a bug?

### Full Code for 2.4.X
```
std::vector<int> lin = {1,2,3,4,5};
cv::Mat linmat = cv::Mat(lin).reshape(0,1);
cv::repeat(linmat,5,1,linmat);
```

### "Workaround" Code for 3.2
```
std::vector<int> lin = {1,2,3,4,5};
cv::Mat temp = cv::Mat(lin).reshape(0,1), linmat;
cv::repeat(temp,5,1,linmat);
```

### PR DESCRIPTION
**Pull request by sovrasov**
resolves #8972

### DIFF
--- a/modules/core/src/copy.cpp
+++ b/modules/core/src/copy.cpp
@@ -856,6 +856,7 @@ void repeat(InputArray _src, int ny, int nx, OutputArray _dst)
 {
     CV_INSTRUMENT_REGION()
 
+    CV_Assert(_src.getObj() != _dst.getObj());
     CV_Assert( _src.dims() <= 2 );
     CV_Assert( ny > 0 && nx > 0 );
 
--- a/modules/core/test/test_misc.cpp
+++ b/modules/core/test/test_misc.cpp
@@ -145,3 +145,12 @@ TEST(Core_String, end_method_regression)
     cv::String new_string(old_string.begin(), old_string.end());
     EXPECT_EQ(6u, new_string.size());
 }
+
+TEST(Core_Copy, repeat_regression_8972)
+{
+    Mat src = (Mat_<int>(1, 4) << 1, 2, 3, 4);
+
+    ASSERT_ANY_THROW({
+                         repeat(src, 5, 1, src);
+                     });
+}

### CODE REVIEW

File: modules/core/src/copy.cpp [hunk 1]
```diff
@@ -860,14 +860,15 @@ void repeat(InputArray _src, int ny, int nx, OutputArray _dst)
     CV_Assert( ny > 0 && nx > 0 );
 
     Size ssize = _src.size();
+    Mat src = _src.getMat();
     _dst.create(ssize.height*ny, ssize.width*nx, _src.type());
 
 #if !defined __APPLE__
     CV_OCL_RUN(_dst.isUMat(),
```
alalek: It is bad idea to fetch data into `cv::Mat` before `CV_OCL_RUN` statements.

alalek:
"Inplace" destination in this case doesn't make sense, so I believe it is better to throw exceptions.

sovrasov:
But it can be convenient in some cases to avoid redundant local variables.

modules/core/src/copy.cpp
```diff
@@ -859,15 +861,16 @@ void repeat(InputArray _src, int ny, int nx, OutputArray _dst)
     CV_Assert( _src.dims() <= 2 );
     CV_Assert( ny > 0 && nx > 0 );
 
-    Size ssize = _src.size();
-    _dst.create(ssize.height*ny, ssize.width*nx, _src.type());
-
 #if !defined __APPLE__
-    CV_OCL_RUN(_dst.isUMat(),
+    CV_OCL_RUN(_src.isUMat(),
```
alalek: `_dst.isUMat()` check is required for `CV_OCL_RUN` macro.

Also OpenCL(UMat) path doesn't support "inplace" mode.
Please take a look here: #8352 - it can't be resolved in general.

vpisarev:
:+1:

PR #8990, opened 2017-06-26
Fix several issues found by static analysis

mshabunin:
- fixed warnings about uninitialized variables or class members
- enabled `-Wuninitialized` compilation flag for GCC


alalek:
:+1:

PR #8998, opened 2017-06-27
add java wrappers to dnn module

abratchik:
### This pullrequest changes

Added java wrappers to dnn module.
Added autotests based on Google Inception 5 model graph. The autotest depends on OPENCV_DNN_TEST_DATA_PATH  and OPENCV_TEST_DATA_PATH environment variables pointing at correct folders where the following files should exist:

- {OPENCV_DNN_TEST_DATA_PATH }/dnn/tensorflow_inception_graph.pb
- {OPENCV_TEST_DATA_PATH}/dnn/space_shutte.jpg

If the OPENCV_DNN_TEST_DATA_PATH is not set the test will be disabled. If OPENCV_TEST_DATA_PATH is not set it will fail. If both env vars are set but one of the files or both are missing the test will fail.

### DIFF
--- a/modules/dnn/misc/java/test/DnnTensorFlowTest.java
+++ b/modules/dnn/misc/java/test/DnnTensorFlowTest.java
@@ -1,0 +1,113 @@
+package org.opencv.test.dnn;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import org.opencv.core.Core;
+import org.opencv.core.Mat;
+import org.opencv.core.Scalar;
+import org.opencv.core.Size;
+import org.opencv.dnn.DictValue;
+import org.opencv.dnn.Dnn;
+import org.opencv.dnn.Importer;
+import org.opencv.dnn.Layer;
+import org.opencv.dnn.Net;
+import org.opencv.imgcodecs.Imgcodecs;
+import org.opencv.imgproc.Imgproc;
+import org.opencv.test.OpenCVTestCase;
+
+public class DnnTensorFlowTest extends OpenCVTestCase {
+
+    private final static String ENV_OPENCV_DNN_TEST_DATA_PATH = "OPENCV_DNN_TEST_DATA_PATH";
+
+    private final static String ENV_OPENCV_TEST_DATA_PATH = "OPENCV_TEST_DATA_PATH";
+
+    String modelFileName = "";
+    String sourceImageFile = "";
+
+    Net net;
+
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+
+        String envDnnTestDataPath = System.getenv(ENV_OPENCV_DNN_TEST_DATA_PATH);
+
+        if(envDnnTestDataPath == null){
+            isTestCaseEnabled = false;
+            return;
+        }
+
+        File dnnTestDataPath = new File(envDnnTestDataPath);
+        modelFileName =  new File(dnnTestDataPath, "dnn/tensorflow_inception_graph.pb").toString();
+
+        String envTestDataPath = System.getenv(ENV_OPENCV_TEST_DATA_PATH);
+
+        if(envTestDataPath == null) throw new Exception(ENV_OPENCV_TEST_DATA_PATH + " has to be defined!");
+
+        File testDataPath = new File(envTestDataPath);
+
+        File f = new File(testDataPath, "dnn/space_shuttle.jpg");
+        sourceImageFile = f.toString();
+        if(!f.exists()) throw new Exception("Test image is missing: " + sourceImageFile);
+
+        net = new Net();
+        if(net.empty()) {
+            Importer importer = Dnn.createTensorflowImporter(modelFileName);
+            importer.populateNet(net);
+        }
+
+    }
+
+    public void testGetLayerTypes() {
+        List<String> layertypes = new ArrayList();
+        net.getLayerTypes(layertypes);
+
+        assertFalse("No layer types returned!", layertypes.isEmpty());
+    }
+
+    public void testGetLayer() {
+        List<String> layernames = net.getLayerNames();
+
+        assertFalse("Test net returned no layers!", layernames.isEmpty());
+
+        String testLayerName = layernames.get(0);
+
+        DictValue layerId = new DictValue(testLayerName);
+
+        assertEquals("DictValue did not return the string, which was used in constructor!", testLayerName, layerId.getStringValue());
+
+        Layer layer = net.getLayer(layerId);
+
+        assertEquals("Layer name does not match the expected value!", testLayerName, layer.get_name());
+
+    }
+
+    public void testTestNetForward() {
+        Mat rawImage = Imgcodecs.imread(sourceImageFile);
+
+        assertNotNull("Loading image from file failed!", rawImage);
+
+        Mat image = new Mat();
+        Imgproc.resize(rawImage, image, new Size(224,224));
+
+        Mat inputBlob = Dnn.blobFromImage(image);
+        assertNotNull("Converting image to blob failed!", inputBlob);
+
+        Mat inputBlobP = new Mat();
+        Core.subtract(inputBlob, new Scalar(117.0), inputBlobP);
+
+        net.setInput(inputBlobP, "input" );
+
+        Mat result = net.forward();
+
+        assertNotNull("Net returned no result!", result);
+
+        Core.MinMaxLocResult minmax = Core.minMaxLoc(result.reshape(1, 1));
+
+        assertTrue("No image recognized!", minmax.maxVal > 0.9);
+
+
+    }
+
+}
--- a/modules/dnn/CMakeLists.txt
+++ b/modules/dnn/CMakeLists.txt
@@ -9,7 +9,7 @@ endif()
 
 set(the_description "Deep neural network module. It allows to load models from different frameworks and to make forward pass")
 
-ocv_add_module(dnn opencv_core opencv_imgproc WRAP python matlab)
+ocv_add_module(dnn opencv_core opencv_imgproc WRAP python matlab java)
 ocv_warnings_disable(CMAKE_CXX_FLAGS -Wno-shadow -Wno-parentheses -Wmaybe-uninitialized -Wsign-promo
                                      -Wmissing-declarations -Wmissing-prototypes
 )
--- a/modules/dnn/include/opencv2/dnn/dict.hpp
+++ b/modules/dnn/include/opencv2/dnn/dict.hpp
@@ -39,14 +39,14 @@
 //
 //M*/
 
-#ifndef OPENCV_DNN_DNN_DICT_HPP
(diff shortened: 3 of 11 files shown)


modules/dnn/cmake/download_google_inception5.cmake
```diff
@@ -0,0 +1,30 @@
+function(download_google_inception5 dst_dir status_var)
```
alalek: No need to download these files, because they are not external to OpenCV.
- space_shuttle.jpg : opencv sources
- tensorflow_inception_graph.pb: opencv_extra testdata directory, check for `$ENV{OPENCV_TEST_DATA_PATH}`

abratchik: Hi, Alex, thank you, this helps. Need your advice though: $ENV{OPENCV_TEST_DATA_PATH} - is this cmake variable or environment? I did not find it in CMakeVars.txt. The closest match was OPENCV_TEST_DATA_INSTALL_PATH=share/OpenCV/testdata. As far as I can see it is used currently only by text module.
The other question - do I have to copy space_shuttle to one of the build/ subfolders during cmake? Which one I need to use?

dkurt: Hi, @abratchik ! `OPENCV_TEST_DATA_PATH` and `OPENCV_DNN_TEST_DATA_PATH` both works only as environment variables.

abratchik: Hi, Dmitry, thank you. This makes sense. I checked on my Mac, OPENCV_TEST_DATA_PATH points at contrib_extra/testdata i.e. part of contrib_extra repository. I'm not sure if it was sent automatically or I did it manually, don't remember. This is fine but here are my questions:
1) I'd like to use space_shuttle.jpg, which is in the opencv/dnn/tutorials/images. If I copy it there it means I'm duplicating the file from opencv to opencv_extra, which is not good. On the other hand, when I'm running unit test in java, I need to find this file somehow. If it is somewhere under build/ it is easy because this is current directory for java autotests. I can also use  OPENCV_TEST_DATA_PATH but this is less attractive because this variable may be not defined - do I have to skip test in this case?
2) OPENCV_DNN_TEST_DATA_PATH is empty on my Mac. If I want my tests to be included in auto-builds, I need to where this var is pointing to on auto-build environments. 
Sorry if my questions are a bit unclear.

alalek: `OPENCV_TEST_DATA_PATH` is specified manually before OpenCV test launch.

1.
> this variable may be not defined - do I have to skip test in this case?

OpenCV C++ tests fail in this case. They are "skipped" in case of external data only.
Unfortunately we don't have Java tests that uses `OPENCV_TEST_DATA_PATH` (in C++ tests test data is managed by "opencv_ts" module - via `cvtest::findDataFile()` call).

2.
`OPENCV_DNN_TEST_DATA_PATH` points to additional optional data (large files with DNN models).
They can be downloaded by this [script](https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/download_models.py) into separate directory ("OPENCV_DNN_TEST_DATA_PATH/dnn", size is 2.1Gb).

abratchik: Thank you Alex. So would that be safe to assume that OPENCV_DNN_TEST_DATA_PATH is set on the auto-build environments? May I rely on that?

Lat question - is it possible to move space_shuttle to testdata/dnn? I mean since there is a general rule to store test data there then I guess no point to have any test images in the main repository?

alalek: `OPENCV_TEST_DATA_PATH` - it must be set
`OPENCV_DNN_TEST_DATA_PATH` - optional, especially in case of embedded device.

> is it possible to move space_shuttle to testdata/dnn

Yes, you can copy this file to opencv_extra.

abratchik: Great, just [added](https://github.com/opencv/opencv_extra/pull/348), please kindly check and merge. I'll be waiting for your merge to complete the dnn wrappers pull request.

abratchik:
Hi, @alalek , with your latest patch java wrappers are generated successfully but it seems like CV__DNN_EXPERIMENTAL_ is also creating issues for python wrappers:

```
In file included from /build/precommit_linux64/build/modules/python3/pyopencv_generated_include.h:42:0,
                 from /build/precommit_linux64/opencv/modules/python/src2/cv2.cpp:12:
/build/precommit_linux64/opencv/modules/dnn/include/opencv2/dnn/dict.hpp:51:1: error: 'CV__DNN_EXPERIMENTAL_NS_BEGIN' does not name a type
 CV__DNN_EXPERIMENTAL_NS_BEGIN
 ^
In file included from /build/precommit_linux64/build/modules/python3/pyopencv_generated_include.h:42:0,
                 from /build/precommit_linux64/opencv/modules/python/src2/cv2.cpp:12:
```

alalek:
Order of includes is matter, please add [this line](https://github.com/opencv/opencv/blob/64b483c65be0eca72599bb5aa9c85160a339e740/modules/dnn/include/opencv2/dnn/all_layers.hpp#L44) into `dict.hpp`.

modules/python/src2/hdr_parser.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+##!/usr/bin/env python
```
alalek: Please revert this change

abratchik: Sure, will do. Please also let know where I can find tensorflow_inception_graph.pb on the autobuild servers? :) I'm using OPENCV_DNN_TEST_DATA_PATH variable and it gives me /opt/build-testdata. You can check the junit report [here](http://pullrequest.opencv.org/buildbot/builders/precommit_macosx/builds/11255/steps/test_java/logs/junit-report-html#DnnTensorFlowTest). But it is not there ))

alalek:
BTW, I have moved my fixes here: #9050

alalek:
@abratchik Current failures are related to non-ASCII symbols in source code comments. I'm investigating workaround for this problem.

alalek:
non-ASCII messages are related to #8895 and currently they don't cause build failure.

abratchik:
@alalek it seems like the issue with python3 wrappers. What is strange: I have both python2 and python 3 enabled on my Mac, all builds up well without any issue. One warning pops out on python3 like this:
```

In file included from /Users/abratchik/netbeansprojects/opencv/modules/python/src2/cv2.cpp:10:
In file included from /opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/numpy/core/include/numpy/ndarrayobject.h:18:
In file included from /opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/numpy/core/include/numpy/ndarraytypes.h:4:
/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/site-packages/numpy/core/include/numpy/npy_common.h:17:5: warning: 
      'NPY_INTERNAL_BUILD' is not defined, evaluates to 0 [-Wundef]
#if NPY_INTERNAL_BUILD

```

 But the end all builds and compiles successfully on my machine but not on autobuild servers. I also rebased on the latest master build (e5aa2135544cd3d1cf36521a77559588b292fd32) so all your fixes should be there already. Please kindly help me as I'm out of ideas how to fix it. Seems to be some old error as if it would be before your fixes but I cannot reproduce it locally.

alalek:
> all builds up well without any issue

I have the same problem - can't reproduce current failures on my machine locally. Will take a look on build containers.

alalek:
@abratchik Please apply [this patch](https://github.com/alalek/opencv/commit/pr8998_r)

abratchik:
ok, trying ...

abratchik:
@alalek now it seems to be ok but again, space_shuttle is missing :) Kindly merge [this one](https://github.com/opencv/opencv_extra/pull/349) please.

modules/dnn/misc/java/test/DnnTensorFlowTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+        if(path == null){
+            isTestCaseEnabled = false;
+            return;
+        }
+
+        File currentRelativePath = new File(System.getenv("OPENCV_DNN_TEST_DATA_PATH"));
+        modelFileName =  new File(currentRelativePath, "dnn/tensorflow_inception_graph.pb").toString();
+
+        File f = new File(currentRelativePath, "dnn/space_shuttle.jpg");
```
alalek: opencv_extra testdata files are located under
`OPENCV_TEST_DATA_PATH`
env variable instead of
`OPENCV_DNN_TEST_DATA_PATH`

abratchik: @alalek now tensorflow_inception_graph.pb is also missing whereas it was previously there for OPENCV_DNN_TEST_DATA_PATH :)

alalek: There are two environment variables:
- for general OpenCV testdata  from **opencv_extra**: `OPENCV_TEST_DATA_PATH`
- optional large DNN models: `OPENCV_DNN_TEST_DATA_PATH`
You should use both.

abratchik:
@alalek Alex, there is also one warning on the Mac environment, not sure what that means. Please also let me know the right location of tensorflow_inception_graph.pb and space_shutte.jpg. Teh first one seems to be found in the   OPENCV_DNN_TEST_DATA_PATH/dnn folder but the last one is not there. Should I try to search the model in OPENCV_DNN_TEST_DATA_PATH/dnn and the jpg in OPENCV_TEST_DATA_PATH/dnn?

alalek:
For Mac build try to suppress warning [here](https://github.com/opencv/opencv/blob/e5aa2135544cd3d1cf36521a77559588b292fd32/modules/java/CMakeLists.txt#L389):
```.diff
-ocv_warnings_disable(CMAKE_CXX_FLAGS -Wunused-const-variable)
+ocv_warnings_disable(CMAKE_CXX_FLAGS -Wunused-const-variable -Wundef)
```

abratchik:
@alalek done, the compilation warning seems to be addressed. Now the only remaining issue is placement of test files )

alalek:
https://github.com/opencv/opencv/pull/8998#discussion_r125158569 :

There are two environment variables:
- for general OpenCV testdata  from **opencv_extra**: `OPENCV_TEST_DATA_PATH`
- optional large DNN models: `OPENCV_DNN_TEST_DATA_PATH`

You should use both.

abratchik:
@alalek thank you Alex, it seems like worked this time around. Sorry it took a bit long for me to figure out since it is first time I'm writing the test case ))

abratchik:
@alalek this is ready for merge, no more commits )

alalek:
:+1:

mshabunin:
:+1:

PR #9003, opened 2017-06-27
Fixed some bugs from Halide tests

dkurt:
### This pullrequest changes

* Fixed ReLU Halide function with non-zero negative slope
* Keep origin fully-connected weights (Halide buffer wrap pointer to data. So in case of vector alignment it would be shifted)
* Modified Concat and Element-wise (sum, mult, max) tests considering layers order


vpisarev:
:+1:

PR #9005, opened 2017-06-27
AVX optimized implementation of accumulate functions migrated

terfendail:
**WIP**
### This pullrequest changes
Migrated accumulate AVX optimization to separate file

### DIFF
--- a/modules/imgproc/src/accum.cpp
+++ b/modules/imgproc/src/accum.cpp
@@ -86,318 +86,7 @@ struct AccW_SIMD
     }
 };
 
-#if CV_AVX
-template <>
-struct Acc_SIMD<float, float>
-{
-    int operator() (const float * src, float * dst, const uchar * mask, int len, int cn) const
-    {
-        int x = 0;
-        if (!mask)
-        {
-            len *= cn;
-            for ( ; x <= len - 8 ; x += 8)
-            {
-                __m256 v_src = _mm256_loadu_ps(src + x);
-                __m256 v_dst = _mm256_loadu_ps(dst + x);
-                v_dst = _mm256_add_ps(v_src, v_dst);
-                _mm256_storeu_ps(dst + x, v_dst);
-            }
-        }
-        return x;
-    }
-};
-
-template <>
-struct Acc_SIMD<float, double>
-{
-    int operator() (const float * src, double * dst, const uchar * mask, int len, int cn) const
-    {
-        int x = 0;
-        if (!mask)
-        {
-            len *= cn;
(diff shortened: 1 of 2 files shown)


modules/imgproc/src/accum.cpp
```diff
@@ -1493,10 +1182,24 @@ struct AccW_SIMD<ushort, double>
...
+template <typename T, typename AT> int AccProd_AVX(const T *, const T *, AT *, const uchar *, int, int);
+template <typename T, typename AT> int AccW_AVX(const T *, AT *, const uchar *, int, int, AT);
+#endif
+
 template<typename T, typename AT> void
 acc_( const T* src, AT* dst, const uchar* mask, int len, int cn )
 {
-    int i = Acc_SIMD<T, AT>()(src, dst, mask, len, cn);
+    int i;
+
```
mshabunin: Doc builder produces warning here:
```
modules/imgproc/src/accum.cpp:1196: trailing whitespace.
+    
```

modules/imgproc/src/accum.cpp
```diff
@@ -1493,10 +1182,24 @@ struct AccW_SIMD<ushort, double>
...
 template<typename T, typename AT> void
 acc_( const T* src, AT* dst, const uchar* mask, int len, int cn )
 {
-    int i = Acc_SIMD<T, AT>()(src, dst, mask, len, cn);
+    int i;
+    
+#if CV_AVX
+    if (checkHardwareSupport(CPU_AVX))
+        i = Acc_AVX<T, AT>(src, dst, mask, len, cn);
```
alalek: Please rebase and make changes similar to this patch:
https://github.com/opencv/opencv/pull/9007/files

Currently "dispatching" doesn't work at all.

tomoaki0705:
@terfendail , could you have a look on #9024 , please ?
I tried to add dispatch mode for these accumulate function series, and my patch will make a conflict with yours.  I really didn't do this on purpose, but I thought I should notice you.

I'm OK to drop my PR, but since we want to add a dispatch mode for each optimization, it might be worth to add SSE2/AVX/AVX2/NEON at the same time. Still, my PR is not yet finished.
I just wanted to know your opinion ( and from others ) which way to go (yours or mine or other choice), before either (or both) of us waste some time writing duplicated codes.

Again, I really didn't do this by any thought to heart you or over write your patch.

terfendail:
@tomoaki0705 , your [patch](https://github.com/opencv/opencv/pull/9024) looks more consistent and IMO is a better way to go. So I'll suspend my work on accumulate functions.

alalek:
#9024

PR #9013, opened 2017-06-28
Removed usage of std::map in DetectionOutput layer

arrybn:
### This pullrequest changes
Removed usage of std::map in DetectionOutput layer


vpisarev:
thanks, looks good! did you measure the performance before and after?

arrybn:
Performance for the SSD sample (averaged on 50 forward passes):

- Before: 431
- After: 426

vpisarev:
cool! :+1:

PR #9014, opened 2017-06-28
core: fix infinite recursion in compare

### LINKED ISSUE #8999: core: infinite recursion in compare()
- OpenCV => 3.2 (master)
- Operating System / Platform => Windows 64 Bit (win10)
- Compiler => mingw64

##### Detailed description
if both InputArrays can be interpreted as a Scalar (checkScalar() returns true), compare() goes into [an infinite recursion here](https://github.com/opencv/opencv/blob/master/modules/core/src/arithm.cpp#L1247)

##### Steps to reproduce

   
	Mat_<double> A(4,1); A << 1,3,2,4; // triggers checkScalar()
	Mat_<double> B(1,1); B << 3.4;     // also.
	Mat C;
	compare(A, B, C, CMP_LT);
	cerr << C << endl;

sovrasov:
resolves #8999

### DIFF
--- a/modules/core/src/arithm.cpp
+++ b/modules/core/src/arithm.cpp
@@ -1239,7 +1239,10 @@ void cv::compare(InputArray _src1, InputArray _src2, OutputArray _dst, int op)
             || !_src1.sameSize(_src2)
             || _src1.type() != _src2.type())
     {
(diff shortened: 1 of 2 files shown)


vpisarev:
:+1:

PR #9017, opened 2017-06-28
core: add an ability to use c++ 11 lambda expression as a parallel_for_ body

sovrasov:


modules/core/src/parallel.cpp
```diff
@@ -455,6 +455,29 @@ void cv::parallel_for_(const cv::Range& range, const cv::ParallelLoopBody& body,
     }
 }
 
+#if __cplusplus >= 201103L
+class ParallelLoopBodyLambdaWrapper : public cv::ParallelLoopBody
```
alalek: I believe we can put this into header files (without requirement for new CV_EXPORTS functions)

modules/core/include/opencv2/core/utility.hpp
```diff
@@ -56,6 +56,10 @@
 #include "opencv2/core.hpp"
 #include <ostream>
 
+#if __cplusplus >= 201103L
```
alalek: MSVC defines `__cplusplus` to 199711
Probably we should target for MSVS 2013+ too.
Check [this condition](https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/cvdef.h#L366).

modules/core/src/parallel.cpp
sovrasov: But in this case user should use strange wrapper instead of directly pass lambda to `parallel_for_`. Most users can write such a dummy wrapper by themselves, so we don't need this overload at all.
@vpisarev what do you think?

modules/core/include/opencv2/core/utility.hpp
sovrasov: May be we should replace separate features detection with common `CV_CXX11` in `cvdef.h`

modules/core/include/opencv2/core/utility.hpp
```diff
@@ -56,6 +56,10 @@
 #include "opencv2/core.hpp"
 #include <ostream>
 
+#if CV_CXX_11
```
alalek: `#ifdef` should be used here with current macro implementation.

vpisarev:
looks cool! :+1:

PR #9020, opened 2017-06-28
Fixed python sample for googlenet in dnn

arrybn:
### This pullrequest changes
Updated python googlenet sample to make it working


vpisarev:
:+1:

PR #9022, opened 2017-06-28
Keep origin convolution layer weights for Halide

dkurt:
### This pullrequest changes
Return `.clone()` call after #9011 changes.


vpisarev:
:+1:

PR #9024, opened 2017-06-28
imgproc: dispatch the accumulate series

tomoaki0705:
### This pullrequest changes
 - implemented dispatch version of accumulate function series.
 - optimize using ```SSE2```, ```NEON``` and ```AVX```


vpisarev:
@tomoaki0705 (adding @terfendail and @alalek as well to the conversation). This patch looks great, thanks! There is one important note though. We don't want to make all the optimized code dispatched. SSE2 and ARMv7 NEON are considered as baseline, i.e. they can always be used (enclosed with `#ifdef CV_SSE2`, `#ifdef CV_NEON`). To cover them both we can use universal intrinsics (preferable option) or use them directly. Only AVX/AVX2 and SSE4.x optimizations should be dispatched.  ARM v8 NEON intrinsics should be dispatched on 32-bit ARM platforms, but can freely be used on 64-bit ARM platforms. In other words, we don't want to introduce extra complexity where it's not needed. So, we can accept your patch instead of the one done by @terfendail, but do not create separate sse2 files, it's unnecessary complexity

alalek:
I believe, no need to split/duplicate code that based on universal intrinsics (SSE2/NEON/etc and even generic C++ code).
Please put this "common" code into ".simd.hpp" file and include it from .cpp files.
BTW, CMake scripts is able to [generate ".opt.cpp" files](https://github.com/opencv/opencv/blob/27c7ef28a756707bea87605d5ef0e4e64095d280/modules/core/CMakeLists.txt#L3) for your - please take a look on "mathfuncs_core" file from "core" module. It is smart version, so you will not see `.sse2.cpp` file on x64 platform by default.

> ARM v8 NEON intrinsics should be dispatched on 32-bit ARM platforms, but can freely be used on 64-bit ARM platforms

It should be handled via using of introduced dispatching macros. Manual complex "#ifdefs" code must be avoided.
BTW, sometimes it will raise "unreachable code" warning which is hard to eliminate especially in case of multiple dispatching paths.

Also these macros are useful for code validation of optimized code paths (we can turn off all dispatched from the single point - `CV_DISABLE_OPTIMIZATION` CMake option)

vpisarev:
I would not even bother making .simd.hpp for universal intrinsics, it's too much complexity for too little benefit

alalek:
`.simd.hpp` is a solution:
- to prevent code copy-paste
- for effective support of universal intrinsics
- to put all code to single place including "tail processing"
- to process existed `#if CV_AVX` correctly without additional changes

So, It is very similar to current OpenCV code structure, but you need to wrap "optimized code" into standalone functions (that don't use other code parts except universal intrinsics).

tomoaki0705:
@alalek, @vpisarev , I appreciate your nice review.

I was feeling it's a waste of time copy-pasting the universal intrinsic codes from on to another.

the mathfunc_core file was a good example for me.

I'll keep working cleaning up this file.

Thank you

alalek:
@tomoaki0705 BTW, There is simple "how to" example for code dispatching: #9074 . Perhaps it will be useful.

tomoaki0705:
@alalek, thanks for the notice.
I came up with the idea to give up for the ```AVX2``` version for now.
I'll just add ```AVX``` version and clean up soon.

BTW, why is the build bot always kicked when I push ?
In my recognition, if I write WIP at the beginning of the PR, the build bot would not be recognized.
Am I doing something wrong ?

alalek:
You need to squash your commits and rebase to latest master:

> This branch has conflicts that must be resolved
> Conflicting files
> modules/core/CMakeLists.txt

modules/imgproc/src/accum.simd.hpp
```diff
@@ -0,0 +1,2740 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
```
alalek: Please add `#include "opencv2/core/hal/intrin.hpp"` here.

Warning comes from these files:
> In file included from /build/precommit_macosx/build/modules/imgproc/accum.sse2.cpp

BTW, This file is autogenerated by CMake (`.sse2.cpp`, `.avx.cpp` and `.neon.cpp`) in build directory. This may confuse because you have files with similar names in "src" directory.

tomoaki0705: Thanks for your suggestion.
I think I'm almost there.  I hope I can find more time tomorrow

Best.

alalek:
Thank you! :+1:

PR #9028, opened 2017-06-28
dnn: added "hidden" experimental namespace

alalek:
Current public API of DNN module is not stabilized yet (especially on the "binary compatibility" level).
- there are some interfaces that require changes for better "const ref" parameters:
```
static Ptr<Layer> createLayerInstance(const String &type, LayerParams& params);
typedef Ptr<Layer>(*Constuctor)(LayerParams &params);
virtual void forward(std::vector<Mat*> &input, ...)
virtual int inputNameToIndex(String inputName);
...
```
- many "virtual" methods
- TODOs in public headers

Main purpose of this namespace is to avoid using of incompatible binaries that will cause applications crashes.

This additional namespace will not impact "Source code API". This change allows to continue maintaining of ABI checks for other OpenCV modules (with easy filtering out of DNN code).

This patch doesn't require any changes in tests or samples.

### DIFF
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -43,10 +43,9 @@
 #define OPENCV_DNN_DNN_ALL_LAYERS_HPP
 #include <opencv2/dnn.hpp>
 
-namespace cv
-{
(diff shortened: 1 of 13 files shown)


mshabunin:
:+1:

PR #9029, opened 2017-06-28
dnn: cleanup torch integration code

alalek:
- put functions into namespace (instead of extern "C")
- support file reading only
- remove unused code


vpisarev:
thanks! :+1:

arrybn:
:+1:

PR #9031, opened 2017-06-29
gtk: check NULL before unref

huangqinjin:
Calling `g_object_unref()` on a `NULL` pointer results in
```
GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
```
so check it before.


mshabunin:
:+1:

Issue #7599, opened 2016-11-02
Incorrect element size of cv::Mat when initialised with an instance of cv::Mat_<uint32_t>.


markcox80:
##### System information (version)
- OpenCV => 2.4.13
- Operating System / Platform => OSX 10.10.5
- Compiler => Apple LLVM version 7.0.2 (clang-700.1.81)

##### Detailed description

The example below initialises an instance of cv::Mat with an instance of cv::Mat_<uint32_t>. Unexpectedly, the element size of the cv::Mat is different to the original cv::Mat_<uint32_t>.

The problem appears to stem from CV_ELEM_SIZE and/or CV_ELEM_SIZE1.

This problem impacts iterators returned cv::Mat_<uint32_t>::begin() and cv::Mat_<uint32_t>::end().

##### Steps to reproduce
```.cpp
#include <opencv2/core.hpp>
#include <iostream>

using namespace std;
using namespace cv;

int
main(int argc, char **argv)
{
  const size_t rows = 100;
  const size_t cols = 200;

  Mat_<uint32_t> labels(rows, cols);
  Mat generic(labels);

  cout << "labels.elemSize(): " << labels.elemSize() << endl;
  cout << "generic.elemSize(): " << generic.elemSize() << endl;

  cout << "labels.elemSize1(): " << labels.elemSize1() << endl;
  cout << "generic.elemSize1(): " << generic.elemSize1() << endl;

  auto gtype = generic.type();
  auto dtype = DataType<uint32_t>::type;
  cout << "generic.type(): " << gtype << endl;
  cout << "DataType<uint32_t>::type: " << dtype << endl;

  cout << "CV_ELEM_SIZE(DataType<uint32_t>::type): " << CV_ELEM_SIZE(DataType<uint32_t>::type) << endl;
  cout << "CV_ELEM_SIZE1(DataType<uint32_t>::type): " << CV_ELEM_SIZE1(DataType<uint32_t>::type) << endl;

  return 0;
}

// $ c++ --std=c++11 blah.cpp -I/tmp/opencv-install/include -L/tmp/opencv-install/lib -lopencv_core -lz
// $ ./a.out
// labels.elemSize(): 4
// generic.elemSize(): 8
// labels.elemSize1(): 4
// generic.elemSize1(): 8
// generic.type(): 7
// DataType<uint32_t>::type: 7
// CV_ELEM_SIZE(DataType<uint32_t>::type): 8
[...]

PR #9034, opened 2017-06-29
Add constructors taking initializer_list for some of OpenCV data types

sovrasov:


doc/tutorials/core/mat_the_basic_image_container/mat_the_basic_image_container.markdown
```diff
@@ -192,10 +192,12 @@ object in multiple ways:
 
     ![](images/MatBasicContainerOut3.png)
 
--   For small matrices you may use comma separated initializers:
+-   For small matrices you may use comma separated initializers or initializer lists (C++ 11 support is required in the last case):
```
alalek: Could you please use "C++11" without space (reference: https://en.wikipedia.org/wiki/C%2B%2B11)
It will be nice to rename `CV_CXX_11` -> `CV_CXX11` (google for `"HAVE_CXX11"` vs `"HAVE_CXX_11"`)

alalek:
:+1:

de-vri-es:
This just horribly broke our existing code.

Consider the following snippet:
```
cv::Mat image = cv::imread("image.png");
cv::Mat copy{image};
```

This used to call the copy constructor `cv::Mat::Mat(const cv::Mat &)`. Now it calls `cv::Mat::Mat(const std::initializer_list<cv::Mat>)` which resulted in a 1x1 matrix with type 7 (instead of 8UC3 of the original). This is a pretty serious breakage.

Why should we have this `cv::Mat(const std::initializer_list<cv::Mat>)` constructor in the first place? What is it even supposed to do?

alalek:
@de-vri-es Problem is related to buggy default value of type trait: #7599

Anyway, after fix of mentioned problem, this code would be probably become incorrect and cause compiler error. Consider to use these statements:
- `cv::Mat copy(image);`
- `cv::Mat copy = image;`

de-vri-es:
Breaking `cv::Mat copy{image}` is really dangerous. `cv::Mat copy(image)` and `cv::Mat copy{image}` should do the same.

The whole point of uniform initialization syntax [1] was to make sure the programmer doesn't have to worry about which type of bracket to use for object initialization. But if suddenly the new syntax doesn't invoke the copy constructor anymore, you definitely have to worry.

[1] http://www.stroustrup.com/C++11FAQ.html#uniform-init

It can be fixed without removing the `std::initializer_list` constructor though, by restricting the types for _Tp. See #9507

PR #9037, opened 2017-06-29
Rewrote googlenet tests

arrybn:
### This pullrequest changes
Rewrote googlenet test: devided one compicated case to simpler ones;
Fixed issue in convolution layer with saving relu slopes after clear;


arrybn:
FYI I've measured performance and it remains the same

alalek:
:+1:

PR #9039, opened 2017-06-29
Disabled logging in caffe parser in release

arrybn:
### This pullrequest changes
Disabled logging in caffe parser in release. Parser still prints messages in Debug builds


alalek:
:+1:

PR #9041, opened 2017-06-29
AVX optimized implementation of separable filters migrated

terfendail:
### This pullrequest changes
AVX implementation of SIMD row and column processors migrated to separate file

### DIFF
--- a/modules/imgproc/src/filter.avx2.cpp
+++ b/modules/imgproc/src/filter.avx2.cpp
@@ -1,0 +1,200 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+//
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                           License Agreement
+//                For Open Source Computer Vision Library
+//
+// Copyright (C) 2000-2008, Intel Corporation, all rights reserved.
+// Copyright (C) 2009, Willow Garage Inc., all rights reserved.
+// Third party copyrights are property of their respective owners.
+//
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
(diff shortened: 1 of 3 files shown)


modules/imgproc/src/filter.avx2.cpp
```diff
@@ -0,0 +1,199 @@
...
+            s1 = _mm256_fmadd_ps(x0, f, s1);
+#else
+            s1 = _mm256_add_ps(s1, _mm256_mul_ps(x0, f));
+#endif
+        }
+
+        _mm256_storeu_ps(dst + i, s0);
+        _mm256_storeu_ps(dst + i + 8, s1);
+    }
+    _mm256_zeroupper();
```
alalek: This instruction must be placed in the end of function only. There is no reason to put it in the middle.

https://stackoverflow.com/questions/18319488/sse-and-avx-intrinsics-mixture#comment26889057_18321764
>  If you use SSE intrinsics with the AVX compiler flags, then the SSE intrinsics will compile to VEX-encoded SSE. It is perfectly fine to mix VEX-encoded SSE with VEX-encoded AVX.

https://stackoverflow.com/a/28356924

mschoeneck:
Correct me if I'm wrong, but I think [_mm256_zeroupper()](https://github.com/terfendail/opencv/blob/4d0f789e0a4c884381ac3a89cf37b4bcf850a293/modules/imgproc/src/filter.avx2.cpp#L140) hast to be moved between AVX and SSE code. So it should be in line 121. Same here [here](https://github.com/terfendail/opencv/blob/4d0f789e0a4c884381ac3a89cf37b4bcf850a293/modules/imgproc/src/filter.avx2.cpp#L194) --> L179

alalek:
> If you use SSE intrinsics with the AVX compiler flags, then the SSE intrinsics will compile to VEX-encoded SSE

See [comment](https://github.com/opencv/opencv/pull/9041#pullrequestreview-47213703) above.

So there are no "SSE" instructions in AVX-generated binaries.
"Next" SSE instructions are possible after AVX function leave only.

alalek:
:+1:

PR #9042, opened 2017-06-29
AVX optimized implementation of haar migrated to separate file

terfendail:
### This pullrequest changes
Haar AVX implementation migrated


alalek:
:+1:

PR #9048, opened 2017-06-30
imgproc: fix MORPH_HITMISS operation when kernel has no negative values

### LINKED ISSUE #8957: MORPH_HITMISS with all zero kernels
I am running a morphological hit and miss with the kernel center element set to 0 (don't care):

`cv2.morphologyEx(np.asarray([[0, 255, 0], [0, 0, 0], [0, 255, 0]], dtype=np.uint8), cv2.MORPH_HITMISS, np.asarray([[0, 1, 0], [0, 0, 0], [0, 1, 0]], dtype=np.int8))
`
I am getting: [[0, 0, 0], [0, 0, 0], [0, 0, 0]]

While I am expecting to get: [[0, 0, 0], [0, 255, 0], [0, 0, 0]]

I think this is because:
`e2 = src;` 

should be:
e2 = `src_complement;`

edit:
i think for the all-zero cases e1 and/or e2 should get ones_like(src)*255

- OpenCV => :grey_question:
- Compiler => :grey_question:

sovrasov:
resolves #8957

### DIFF
--- a/modules/imgproc/src/morph.cpp
+++ b/modules/imgproc/src/morph.cpp
@@ -2012,8 +2012,6 @@ void cv::morphologyEx( InputArray _src, OutputArray _dst, int op,
     CV_IPP_RUN_FAST(ipp_morphologyEx(op, src, dst, kernel, anchor, iterations, borderType, borderValue));
 #endif
 
-    Mat k1, k2, e1, e2; //only for hit and miss op
-
     switch( op )
     {
     case MORPH_ERODE:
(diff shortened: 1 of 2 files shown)


alalek:
@sovrasov Did you take a look on OpenCL and IPP backends about this case?

/cc @eyaler dan-masek Please take a look on this patch.

sovrasov:
IPP backend is disabled, HITMISS operation is not implemented in OpenCL function.

eyaler:
@alalek as i wrote in the issue, i know think that for the cases where countNonZero<=0 we should take an all on pattern instead or src/~src. the idea is that if the kernel is all zeros we dont care and should short circuit the and operator.

sovrasov:
@eyaler do you mean that if `countNonZero(k1) == 0 && countNonZero(k1) == 0` we should assign `dst=src`?

eyaler:
@sovrasov  my c++ is rusty but in python we should do 
`dst = np.ones_like(src)*255`
the idea is that if the kernel is all zeros we do not care and we want to pass through the AND condition

sovrasov:
Ok, I see. In this case the algorithm will do `src.copyTo(dst)` without making `&`.

alalek:
:+1:

PR #9051, opened 2017-06-30
Remove debugging code from DirectShow video capture

### LINKED ISSUE #8926: '***** VIDEOINPUT LIBRARY - 0.1995 - TFW07 *****' message displayed during startup in debug mode
When migrating a project from OpenCV 2.4.10 to 3.2.0 on a Windows 10/x64 project, I noticed that the following string was always printed in the console window (which confuses my testing framework):

> ***** VIDEOINPUT LIBRARY - 0.1995 - TFW07 *****

I did some digging, and it turns out this behaviour was changed in commit ec93d103b72cb9437cce11e1b93a1522c710d70d - here, the DirectShow debug message mechanism was restructured so that such messages are never printed in release builds, making things a bit more efficient.

However, the 'gs_verbose' flag is always enabled with no way to switch it off. In 2.4.x, a struct 'SupressVideoInputMessages' is used to disable this flag, and you could call 'videoInput::setVerbose(true);' if you really want this functionality (thus by default, it was always disabled and you wouldn't see such messages)

My suggestion is thus to:

- Initialize gs_verbose to false to get rid of the message by default
- Add a function to enable these messages - you can't call videoInput::setVerbose() from the outside as you don't have the class definition.

If people agree, I'm willing to code up the change and submit a pull request to make this happen.

zhmu:
This code is impossible to enable or disable at runtime, and that wouldn't be very useful either due to the static VideoCapture_DShow::g_VI declaration.

As this code is unavailable in release mode anyway and only seemed to have served purpose when creating this code (it was present in the original version in
github ofTheo/videoInput), I suggest it be removed.

resolves #8926 

### This pullrequest changes

This code removes the debugging code from the DirectShow video input, which is always enabled in debugging mode. As the code has been remarkably stable for the last 6 years, I hope we can do without it.

### DIFF
--- a/modules/ml/src/data.cpp
+++ b/modules/ml/src/data.cpp
@@ -403,7 +403,7 @@ public:
             Mat(tempCatMap).copyTo(catMap);
         }
 
-        if( varType.at<uchar>(ninputvars) == VAR_CATEGORICAL )
+        if( noutputvars > 0 && varType.at<uchar>(ninputvars) == VAR_CATEGORICAL )
         {
             preprocessCategorical(responses, &normCatResponses, labels, &counters, sortbuf);
             Mat(labels).copyTo(classLabels);
--- a/modules/videoio/src/cap_dshow.cpp
+++ b/modules/videoio/src/cap_dshow.cpp
@@ -334,35 +334,6 @@ interface ISampleGrabber : public IUnknown
 
 //STUFF YOU CAN CHANGE
 
-#ifdef _DEBUG
-#include <strsafe.h>
-
-//change for verbose debug info
-static bool gs_verbose = true;
-
-static void DebugPrintOut(const char *format, ...)
-{
-    if (gs_verbose)
-    {
-        va_list args;
-        va_start(args, format);
-        if( ::IsDebuggerPresent() )
-        {
-            CHAR szMsg[512];
-            ::StringCbVPrintfA(szMsg, sizeof(szMsg), format, args);
-            ::OutputDebugStringA(szMsg);
-        }
-        else
-        {
-            vprintf(format, args);
-        }
-        va_end (args);
-    }
-}
-#else
-#define DebugPrintOut(...) void()
-#endif
-
 //if you need VI to use multi threaded com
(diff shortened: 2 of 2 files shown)


alalek:
This debug code doesn't harm anything (and completely disabled in Release), so I believe we can keep it here.

alalek:
> remarkably stable for the last 6 years

This is not a criteria for code removal.
There are [several bugs](https://github.com/opencv/opencv/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20dshow%20) filled for "DShow", so ability to enable this debug code is useful for future improvements/fixes/refactoring.

zhmu:
Perhaps that is not a criteria, but as the linked issue remarks: having this code adds debugging output (the ***** VIDEOINPUT LIBRARY - 0.1995 - TFW07 ***** startup banner) to debug builds. This breaks things like attempting to run unit tests from Resharper C/C++, which is extremely annoying (especially as you often do that with debug builds)

Would you settle for a change which only removes that specific line from the output?

alalek:
Perhaps this should work:

```.diff
-static bool gs_verbose = true;
+static bool gs_verbose = false;

 static void DebugPrintOut(const char *format, ...)
 {
+    static bool force_debug = getenv("OPENCV_DSHOW_DEBUG") ? true : false;
-    if (gs_verbose)
+    if (gs_verbose || force_debug)
     {
```

zhmu:
That would work, but there is no way to change gs_verbose as you cannot call videoInput::setVerbose() from outside cap_dshow.cpp (and even if you could, there is still the static VideoCapture_DShow::g_VI to worry about)

Making these messages opt-in with an environment variable or similar would definitely have my support. I'd say you can just get rid of the gs_verbose variable completely.

alalek:
I agree, that `setVerbose` is internal API only and it is not exposed into public OpenCV API - so there is no access to it.

> static VideoCapture_DShow::g_VI 
Using of this global variable is not a good practice. This should be definitely refactored somehow.

zhmu:
Let's close this - I will submit a new pull request shortly.

PR #9052, opened 2017-06-30
a fix for open issue 4772

### LINKED ISSUE #4772: Static build error because of libstdc++ "multiple definitions"
Transferred from http://code.opencv.org/issues/3759

```
|| Christoph Spanknebel on 2014-06-18 18:50
|| Priority: Low
|| Affected: 2.4.9 (latest release)
|| Category: build/install
|| Tracker: Bug
|| Difficulty: Easy
|| PR: 
|| Platform: x86 / Windows
```
## Static build error because of libstdc++ "multiple definitions"

```
I tried to build OpenCV 2.4.9 static libraries with sources from Windows download and compiler TDM-GCC (MinGW32, GCC 4.8.1). In CMake configuration i just unchecked BUILD_SHARED_LIBS (from default).
While compiling the libraries only works, executing the "all" build tries to compile applications which results in errors because of multiple definitions in libstdc++.a and libstdc++.dll.a.
As background: TDM-GCC links runtimes statically by default. The compilation of OpenCV applications shows the flag "-libstdc++", which forces dynamic linking, while the linked libraries have the static runtime linked -> conflict!
```
##### Maksim Shabunin on 2015-04-27 09:11

```
-   Target version changed from 3.0-alpha to 3.0
```
##### Maksim Shabunin on 2015-04-27 14:38

```
-   Target version changed from 3.0 to 2.4.11
```

jqly:
resolves #4772

### This pullrequest changes

As described in open issue #4772 (Static build error because of libstdc++ "multiple definitions"), this error still exists in current opencv 3.2.0.

When `BUILD_SHARED_LIBS` is turned off, `-lstdc++` flag will be added to g++. In this case, `libstdc++.a` and `libstdc++.dll.a` will be all linked, as described in the issue 4772, resulting "multiple definitions" error to occur in several places.

However, the default settings of g++ compiler in TDM-GCC behaves just rightly, which links runtimes statically by default. So I think this line of code should be ommited at least in MINGW case. In my case, it works in windows 10, TDM-GCC Compiler Suite for Windows GCC 5 Series MinGW-w64 64/32-bit Edition.

As has such a fix, I successfully compiled out static libraries of opencv 3.2.0 without any error. And I have tried `opencv_test_core.exe` produced in the process of the whole building, passing all the tests.

### DIFF
--- a/cmake/OpenCVCompilerOptions.cmake
+++ b/cmake/OpenCVCompilerOptions.cmake
@@ -248,7 +248,9 @@ endif()
 # Extra link libs if the user selects building static libs:
 if(NOT BUILD_SHARED_LIBS AND CMAKE_COMPILER_IS_GNUCXX AND NOT ANDROID)
   # Android does not need these settings because they are already set by toolchain file
-  set(OPENCV_LINKER_LIBS ${OPENCV_LINKER_LIBS} stdc++)
+  if(NOT MINGW)
+    set(OPENCV_LINKER_LIBS ${OPENCV_LINKER_LIBS} stdc++)
+  endif()
   set(OPENCV_EXTRA_FLAGS "-fPIC ${OPENCV_EXTRA_FLAGS}")
 endif()
 


alalek:
:+1:

mshabunin:
:+1:

PR #8341, opened 2017-03-08
Fixed #8149 - Parallel canny on multichannel images

schwittmann:
resolves #8149

### This pullrequest changes

As suggested by @mschoeneck in #8149 moving of _dst.create() fixes the problem - if dst gets updated to the newly created matrix, too. opencv_test_imgproc runs fine.

PR #9060, opened 2017-06-30
canny: disallow broken inplace arguments

### LINKED ISSUE #8149: Race condition in Canny(), opencv 3
- OpenCV => 3.1.0-7
- Operating System / Platform => 4.9.2-1-MANJARO x86-64
- Compiler => 6.3.1 20170109 (GCC)
- No OpenCl, Tegra or IPP

##### Detailed description

We wanted our opencv based source code to be compatible with both opencv 2.4 and 3.x. However, our program yielded unexpected results with 3.x. After some rigorous debugging we discovered the bug in Canny(). While the documentation states that the output matrix ["can be the same as the input"](http://docs.opencv.org/3.0-beta/doc/tutorials/imgproc/imgtrans/canny_detector/canny_detector.html) this causes the problem. If we use a dedicated output matrix the results are as expected.

We assume that parallelization optimizations in 3.x causes this problem. Multiple threads operate read and write on the same matrix causing race conditions.

##### Steps to reproduce

    ```.cpp
    Canny( detected_edges, detected_edges, lowThreshold, lowThreshold*ratio, kernel_size );
    ```
On a machine with multiple cpu cores

##### Workaround
    ```.cpp
    Canny( detected_edges, tmp, lowThreshold, lowThreshold*ratio, kernel_size );
    ```

alalek:
resolves #8149

Replaces #8341
Related #8352

### DIFF
--- a/modules/imgproc/src/canny.cpp
+++ b/modules/imgproc/src/canny.cpp
@@ -972,6 +972,9 @@ void Canny( InputArray _src, OutputArray _dst,
 
     const Size size = _src.size();
 
+    // we don't support inplace parameters in case with RGB/BGR src
+    CV_Assert((_dst.getObj() != _src.getObj() || _src.type() == CV_8UC1) && "Inplace parameters are not supported");
+
     _dst.create(size, CV_8U);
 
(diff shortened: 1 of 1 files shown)


sovrasov:
👍

PR #9062, opened 2017-06-30
AVX optimized norm function implementations migrated to separate file

terfendail:
### This pullrequest changes
Moved AVX optimized implemetation of norm to separate file


alalek:
#9074

PR #9067, opened 2017-07-01
Fix error message fisheye CALIB_CHECK_COND

D-Alex:
The old error message was not giving any hint which input array (image)
led to an ill conditioned matrix. This made it near impossible to
identify poor images in a larger set.

A better approach would be to implement a checker function which gives
each image a rating before the real calibration is performed. This could
also include some image properties like sharpness, etc.


sovrasov:
👍

PR #9070, opened 2017-07-02
Add 64-bit imshow behavior in the documentation

### LINKED ISSUE #9010: Documentation of  imshow is not clear about how 64-bit floats are handled.
While the documentation of [`imshow`](http://docs.opencv.org/3.2.0/d7/dfc/group__highgui.html#ga453d42fe4cb60e5723281a89973ee563) is quite clear about what happens with majority of OpenCV datatypes, it omits 64-bit floating point values.

Looking at the [source](https://github.com/opencv/opencv/blob/master/modules/imgcodecs/src/utils.cpp#L622) of the conversion function, it's apparent that the behaviour is identical to 32-bit floats. 

Therefore I suggest the documentation to be updated to reflect this.

catree:
### This pullrequest changes

resolves #9010 
Document the behavior of `imshow` with 64-bit floating point image.


sovrasov:
:+1:

Issue #9065, opened 2017-07-01
Non-ASCII symbols in source code files


alalek:
Example (marked by `--> <--`):
```
202:In the figure below, the Delaunay-->’<<--s triangulation is marked with black lines and the Voronoi
856://! Guil, N., Gonz-->á<--lez-Linares, J.M. and Zapata, E.L. (1999). Bidimensional shape detection using an invariant approach. Pattern Recognition 32 (6): 1025-1038.
958:    @param rect -->–<-- Rectangle that includes all of the 2D points that are to be added to the subdivision.
```

About 68 lines.

Some files has BOM header again (64 files).

TonyLianLong:
I think the BOM appears in the WinRT files and WP files. Actually Microsoft created the BOM and editors on Windows needs those stuffs so it might be okay to have them on those files.

mshabunin:
Looks like this issue can be closed.

mshabunin:
Some issues still left in contrib:
```
[ 73%] Generating gen/tracking+TrackerMedianFlow.java, gen/tracking+Tracking.java, gen/tracking+Tracker.java, gen/tracking+TrackerTLD.java, gen/tracking+TrackerGOTURN.java, gen/tracking+TrackerKCF.java, gen/tracking+MultiTracker.java, gen/tracking+TrackerMIL.java, gen/tracking+TrackerBoosting.java, gen/tracking.cpp
--- Logging error ---
Traceback (most recent call last):
  File "/usr/lib/python3.4/logging/__init__.py", line 980, in emit
    stream.write(msg)
UnicodeEncodeError: 'ascii' codec can't encode character '\u2019' in position 478: ordinal not in range(128)
Call stack:
  File "/build/precommit_linux64/opencv/modules/java/generator/gen_java.py", line 1111, in <module>
    generator.gen(srcfiles, module, dstdir, common_headers)
  File "/build/precommit_linux64/opencv/modules/java/generator/gen_java.py", line 519, in gen
    logging.info("\n--- Incoming ---\n%s", pformat(decl, 4))
Message: '\n--- Incoming ---\n%s'
[...]

PR #9075, opened 2017-07-02
Remove unnecessary Non-ASCII characters from source code

TonyLianLong:
Remove unnecessary Non-ASCII characters and replace them with ASCII
characters
See https://github.com/opencv/opencv/issues/9065


modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -993,9 +993,9 @@ class CV_EXPORTS_W Subdiv2D
 
     /** @brief Returns the location of a point within a Delaunay triangulation.
 
-    @param pt – Point to locate.
-    @param edge – Output edge that the point belongs to or is located to the right of it.
-    @param vertex – Optional output vertex the input point coincides with.
+    @param pt - Point to locate.
+    @param edge - Output edge that the point belongs to or is located to the right of it.
+    @param vertex - Optional output vertex the input point coincides with.
```
alalek: BTW, Dashes "-" in parameters are not necessary (we don't use them in general).

samples/cpp/tutorial_code/calib3d/real_time_pose_estimation/src/PnPProblem.cpp
```diff
@@ -20,7 +20,7 @@ cv::Point3f SUB(cv::Point3f v1, cv::Point3f v2);
 cv::Point3f get_nearest_3D_point(std::vector<cv::Point3f> &points_list, cv::Point3f origin);
 
 
-/* Functions for Möller–Trumbore intersection algorithm */
+/* Functions for Möller-Trumbore intersection algorithm */
```
alalek: Do you have any ideas about "ö" from the name?

TonyLianLong: I think we'd better ask the authors of those files whether they want to change their names into English.

modules/imgproc/include/opencv2/imgproc.hpp
TonyLianLong: Do you want me to replace those dashes with ":" or spaces or just leave them there?

alalek: It would be nice to have single "space" (to keep single code style). [Example](https://github.com/opencv/opencv/blob/3.3.0-rc/modules/core/include/opencv2/core.hpp#L305-L315).

TonyLianLong: Okay, I removed the dashes and replaced them with single spaces.

alalek:
:+1:

PR #9077, opened 2017-07-03
cmake: update CXX11 compiler flag

alalek:
- rename `OPENCV_CXX11` => `ENABLE_CXX11`
- added autodetection of C++11 mode


mshabunin:
:+1:

Issue #8782, opened 2017-05-23
solvePnPRansac behavior wrt to the documentation


catree:
In OpenCV 3.2 (and since OpenCV 3.0), `solvePnPRansac` does not behave as we could expect after reading the corresponding [documentation](http://docs.opencv.org/3.2.0/d9/d0c/group__calib3d.html#ga50620f0e26e02caa2e9adc07b5fbf24e).

The function prototype is: `bool cv::solvePnPRansac(objectPoints, imagePoints, cameraMatrix, 	distCoeffs, rvec, tvec, useExtrinsicGuess = false, iterationsCount = 100, reprojectionError = 8.0, 	confidence = 0.99, inliers = noArray(), flags = SOLVEPNP_ITERATIVE)`.

The parameter `useExtrinsicGuess` will not be used and the choice of the PnP method can be misleading.

The [RANSAC](http://old.vision.ece.ucsb.edu/~zuliani/Research/RANSAC/docs/RANSAC4Dummies.pdf) algorithm implemented currently can be summarized as (when the number of points is `> 4`):

- for a certain number of iterations:
  - select randomly 5 points to construct the "Minimal Sample Sets" (MSSs) and estimate the camera pose using `SOLVEPNP_EPNP` (see [here](https://github.com/opencv/opencv/blob/3.2.0/modules/calib3d/src/solvepnp.cpp#L250)) **[Hypothesize]**
  - compute the consensus set, an inlier is a 3D point with a reprojection error (using `rvec` and `tvec` estimated with MSSs) below a threshold **[Test]**
  - "terminates when the probability of finding a better ranked CS drops below a certain threshold" ([here](https://github.com/opencv/opencv/blob/3.2.0/modules/calib3d/src/ptsetreg.cpp#L237) I think) or after reaching the maximum number of iterations
- estimate the final `rvec` and `tvec` using the inliers from the best consensus set with `flags` provided by the user (if `SOLVEPNP_P3P` it is `SOLVEPNP_EPNP` that is used instead), `useExtrinsicGuess` is ignored and sets to false ([see](https://github.com/opencv/opencv/blob/3.2.0/modules/calib3d/src/solvepnp.cpp#L287))

The documentation should be updated to reflect this behavior (`useExtrinsicGuess` is ignored and about the choice of the PnP method for the two "stages") and maybe we should add a quick description of the implementation of the RANSAC algorithm.

Better, I think the implementation can be improved:

[...]

PR #9086, opened 2017-07-03
Improve solvePnPRansac

### LINKED ISSUE #9085: solvePnPRansac() returns the "wrong" rvec and tvec

- OpenCV => 3.3.0-rc

##### Detailed description

From what I have understood of the code, `solvePnPRansac` returns the best `rvec` and `tvec` estimated during the Minimal Sample Sets step (estimated `rvec` and `tvec` that produce the best (bigger) consensus set) and not `rvec` and `tvec` estimated using all the inliers.
The corresponding faulty code is [here](https://github.com/opencv/opencv/blob/3.3.0-rc/modules/calib3d/src/solvepnp.cpp#L316-L330):

```
if( result <= 0 || _local_model.rows <= 0)
{
    _rvec.assign(rvec);    // output rotation vector
    _tvec.assign(tvec);    // output translation vector

    if( _inliers.needed() )
        _inliers.release();

    return false;
}
else
{
    _rvec.assign(_local_model.col(0));    // output rotation vector
    _tvec.assign(_local_model.col(1));    // output translation vector
}
```

When `result > 0` is ok, `_rvec` and `_tvec` are assigned with the `_local_model` insted of `rvec` and `tvec` estimated using all the inliers.

catree:
### This pullrequest changes

This PR is an attempt to improve and clarify the "algorithm" behind `solvePnPRansac`, see also #8782.
From what I have understood of the code:
 - when the number of input points is equal to the number of model points, `rvec` and `tvec` are estimated two times: once [here](https://github.com/opencv/opencv/blob/3.3.0-rc/modules/calib3d/src/ptsetreg.cpp#L198) and another time [here](https://github.com/opencv/opencv/blob/3.3.0-rc/modules/calib3d/src/solvepnp.cpp#L311)
 - this should be unnecessary and this PR "short-circuits" by calling directly `solvePnP`
 - the `useExtrinsicGuess` parameters has no effect as `ransac_kernel_method` is set by [default](https://github.com/opencv/opencv/blob/3.3.0-rc/modules/calib3d/src/solvepnp.cpp#L269) to `SOLVEPNP_EPNP` or `SOLVEPNP_P3P`, `SOLVEPNP_AP3P` by the user and is not used for the final call to `solvePnP`
 - with this PR, `useExtrinsicGuess` is used with `solvePnP` with all the inliers
 - document a little the behavior of `solvePnPRansac`
 - resolves #9085 there is an issue for me as the function returns the best `rvec` and `tvec` estimated during the Minimal Sample Sets step and not `rvec` and `tvec` estimated using all the inliers

Before this PR:
```
[----------] 3 tests from Calib3d_SolvePnPRansac
[ RUN      ] Calib3d_SolvePnPRansac.accuracy
mode: 0, method: 0 -> 100% (err < 5.08272e-06)
mode: 0, method: 1 -> 100% (err < 1.90336e-05)
mode: 0, method: 2 -> 100% (err < 6.83566e-05)
mode: 0, method: 3 -> 100% (err < 1.48587e-06)
mode: 0, method: 4 -> 100% (err < 3.88479e-05)
mode: 0, method: 5 -> 100% (err < 0.000988152)
mode: 1, method: 0 -> 100% (err < 0.000211219)
mode: 1, method: 1 -> 100% (err < 1.80233e-06)
mode: 1, method: 2 -> 100% (err < 1.70374e-05)
mode: 1, method: 3 -> 100% (err < 5.7117e-06)
mode: 1, method: 4 -> 100% (err < 5.93684e-06)
mode: 1, method: 5 -> 100% (err < 0.00151996)
[       OK ] Calib3d_SolvePnPRansac.accuracy (68 ms)
```

With this PR:
```
[----------] 3 tests from Calib3d_SolvePnPRansac
[ RUN      ] Calib3d_SolvePnPRansac.accuracy
mode: 0, method: 0 -> 100% (err < 1.15535e-07)
mode: 0, method: 1 -> 100% (err < 0.000260042)
mode: 0, method: 2 -> 100% (err < 0.00487855)
mode: 0, method: 3 -> 100% (err < 9.49898e-08)
mode: 0, method: 4 -> 90% (err < 0.0643757)
mode: 0, method: 5 -> 80% (err < 0.063708)
mode: 1, method: 0 -> 100% (err < 5.12943e-07)
mode: 1, method: 1 -> 100% (err < 1.98415e-07)
mode: 1, method: 2 -> 100% (err < 1.5598e-07)
mode: 1, method: 3 -> 100% (err < 0.00293191)
mode: 1, method: 4 -> 100% (err < 0.00293468)
mode: 1, method: 5 -> 100% (err < 0.000291018)
[       OK ] Calib3d_SolvePnPRansac.accuracy (76 ms)
```

At first sight, the accuracy is worse. But before the function returned the best `rvec` and `tvec` estimated during the Minimal Sample Sets step and now `rvec` and `tvec` estimated using all the inliers, the 
RANSAC parameters could be not optimal here.

### DIFF
--- a/modules/calib3d/src/solvepnp.cpp
+++ b/modules/calib3d/src/solvepnp.cpp
@@ -249,7 +249,7 @@ bool solvePnPRansac(InputArray _opoints, InputArray _ipoints,
         ipoints = ipoints0;
 
     int npoints = std::max(opoints.checkVector(3, CV_32F), opoints.checkVector(3, CV_64F));
-    CV_Assert( npoints >= 0 && npoints == std::max(ipoints.checkVector(2, CV_32F), ipoints.checkVector(2, CV_64F)) );
+    CV_Assert( npoints >= 4 && npoints == std::max(ipoints.checkVector(2, CV_32F), ipoints.checkVector(2, CV_64F)) );
 
     CV_Assert(opoints.isContinuous());
     CV_Assert(opoints.depth() == CV_32F || opoints.depth() == CV_64F);
@@ -279,6 +279,31 @@ bool solvePnPRansac(InputArray _opoints, InputArray _ipoints,
         ransac_kernel_method = SOLVEPNP_P3P;
     }
 
+    if( model_points == npoints )
+    {
+        bool result = solvePnP(opoints, ipoints, cameraMatrix, distCoeffs, _rvec, _tvec, useExtrinsicGuess, ransac_kernel_method);
+
+        if(!result)
+        {
+            if( _inliers.needed() )
+                _inliers.release();
+
+            return false;
+        }
+
+        if(_inliers.needed())
+        {
+            _inliers.create(npoints, 1, CV_32S);
+            Mat _local_inliers = _inliers.getMat();
+            for(int i = 0; i < npoints; i++)
+            {
+                _local_inliers.at<int>(i) = i;
+            }
+        }
+
+        return true;
+    }
+
     Ptr<PointSetRegistrator::Callback> cb; // pointer to callback
     cb = makePtr<PnPRansacCallback>( cameraMatrix, distCoeffs, ransac_kernel_method, useExtrinsicGuess, rvec, tvec);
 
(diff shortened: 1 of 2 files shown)


modules/calib3d/src/solvepnp.cpp
```diff
@@ -279,6 +279,31 @@ bool solvePnPRansac(InputArray _opoints, InputArray _ipoints,
...
+
+            return false;
+        }
+
+        if(_inliers.needed())
+        {
+            Mat _local_inliers;
+            for(int i = 0; i < npoints; i++)
+            {
+                _local_inliers.push_back(i);
```
sovrasov: Don't use `push_back` here, required size is known.

modules/calib3d/src/solvepnp.cpp
```diff
@@ -279,6 +279,31 @@ bool solvePnPRansac(InputArray _opoints, InputArray _ipoints,
...
+            }
+            _local_inliers.copyTo(_inliers);
```
sovrasov: Copy can be avoided:
```c++
_inliers.create(...);
Mat _local_inliers = _inliers.getMat()
//fill _local_inliers
```

catree:
Fix also an issue when `_local_model` is empty after the first step.
I did some tests, unfortunately there should be (almost) no improvement between returning `rvec` and `tvec` estimated from the Minimal Sample Sets step and estimated using all the inliers, at least on my test case ([video](https://youtu.be/Vx46XXifEzA)).
Same thing if I tweak the code to use `SOLVEPNP_ITERATIVE` initialized with the camera pose estimated from the MSS step for the final pose estimation ([video](https://youtu.be/7Tfz8uClg3I)). I am still tempted to say that the best logic should be to use for the final pose estimation an iterative method initialized with the results of the MSS step.

tompollok:
@catree is the source code + dataset that you used for the video available somewhere?

sovrasov:
:+1:

catree:
> @catree is the source code + dataset that you used for the video available somewhere?

@tompollok No, it is my own data.

HagegeR:
could this make it into 3.3?

PR #9088, opened 2017-07-04
core: get rid of OPENCV_NOSTL definition

sovrasov:
`OPENCV_NOSTL` is completely useless since OpenCV uses `std::vector` and etc. and thus can't be built without STL.


vpisarev:
I like it; the less obsure configurations, the better! :+1:

PR #9090, opened 2017-07-04
optimize out scaleLayer & concatLayer whenever possible

vpisarev:
n/t


dkurt:
Skip input blob to exclude it from concatenation layer optimizations.
Please add three lines to `dnn.cpp` from https://github.com/dkurt/opencv/commit/96d5b7ec5aad5bf2c6c48a9a78f54eea7172bd1d.
Also Halide test was fixed.

vpisarev:
@dkurt, thank you, I've applied your patch and it works!

alalek:
:+1:

PR #9091, opened 2017-07-04
ffmpeg: update 3.3.2

### LINKED ISSUE #8812: Allow openh264-x.x.x-win??msvc.dll to be found via %PATH%

- OpenCV => 3.2.0-vc14
- Operating System / Platform => Windows 7 SP1 64 Bit
- Compiler => Visual Studio 2015/2017

### Detailed description

When loading `openh264-1.6.0-win64msvc.dll` OpenCV engine doesn't use `%PATH%` when searching for this binary.
It expects (according to `\sources\3rdparty\ffmpeg\readme.txt`) that ...

> ... binary file can be placed into global system path (System32 or SysWOW64) or near application binaries.
>  You can also specify location of binary file via `OPENH264_LIBRARY_PATH` environment variable.

### Actual situation

1. `OPENH264_LIBRARY` (and not `OPENH264_LIBRARY_PATH` - fixed in #6080) allows to specify either full path for the binary to load (a kind of side effect, I guess), or overridden name for the binary.
2. The binary is loaded either from application directory (checked - working), or from system directory (I haven't checked this).
3. Directories from `%PATH%` are ignored when searching for the binary, thus causing inconvenience for non-EXE applications using OpenCV (like Python scripts - [example](https://stackoverflow.com/questions/41972503/could-not-open-codec-libopenh264-unspecified-error)).

### Expected situation

OpenCV should follow "standard" search path strategy for the binary file loading:

1. Application directory.
2. System directory.
3. Directories specified in `%PATH%`.

### Side-effects of `OPENH264_LIBRARY`

I guess, it is possible to have/run 32-bit and 64-bit applications with OpenCV on the same 64-bit PC, so `OPENH264_LIBRARY` will present a conflicting DLL name for such applications.

alalek:
- ffmpeg 3.3 -> 3.3.2
- added VERSIONINFO

https://github.com/opencv/opencv_3rdparty/pull/17

resolves #8812

### DIFF
--- a/3rdparty/ffmpeg/ffmpeg.cmake
+++ b/3rdparty/ffmpeg/ffmpeg.cmake
@@ -1,9 +1,9 @@
-# Binary branch name: ffmpeg/master_20170418
-# Binaries were created for OpenCV: b993b9b7c7f6f5f37d10acacb2962812228410ba
-set(FFMPEG_BINARIES_COMMIT "86c4a841055f2612774e85b4292bb20e5fe8a783")
-set(FFMPEG_FILE_HASH_BIN32 "3dea5f7f009b44601fe95728328e0f9e")
-set(FFMPEG_FILE_HASH_BIN64 "9debe701975ef074bd6661981f3f0716")
-set(FFMPEG_FILE_HASH_CMAKE "208c00f03d2f6f39fa6262649e0bfc8d")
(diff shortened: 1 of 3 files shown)


peters:
@alalek 👍

alalek:
:+1:

PR #9093, opened 2017-07-05
ocl_calcHist1: Use proper local size for merge_histogram kernel

wzw-intel:
merge_histogram kernel only need "BINS" theads to accumulate the
histgrams, it is not efficient to directly use maxGroupSize as
local size if maxGroupSize is far greater then BINS.


alalek:
:+1:

PR #9095, opened 2017-07-05
videoio: fixes for GStreamer support

alalek:
- emulated frame counter (with autodetection of GStreamer broken behavior)
- skip 'seek' tests if seeking is not supported by backend
- update 'fps' and total frames checks (increase error tolerance)
- update synthetic image generation


alalek:
:+1:

PR #9097, opened 2017-07-05
Suggests the cv::Rect() to be the identity under | operation

piotr-semenov:
Please find the minimal nonworking example below:

```
#include <opencv2/core.hpp>
#include <cstdlib>
#include <random>
#include <cassert>

#include <iostream>


int main(int argc, char** argv) {
    using std::rand;

    // Check the monoid's identity laws for rect "union".
    cv::Rect id;
    cv::Rect x(rand(), rand(), rand(), rand());

    assert(("(id | x) == x", (id | x) == x));  // Failed.
    assert(("(x | id) == x", (x | id) == x));  // Failed.

    return EXIT_SUCCESS;
}
```
This commit fixes this bug.

### DIFF
--- a/modules/core/include/opencv2/core/types.hpp
+++ b/modules/core/include/opencv2/core/types.hpp
@@ -1803,12 +1803,17 @@ Rect_<_Tp>& operator &= ( Rect_<_Tp>& a, const Rect_<_Tp>& b )
 template<typename _Tp> static inline
 Rect_<_Tp>& operator |= ( Rect_<_Tp>& a, const Rect_<_Tp>& b )
 {
-    _Tp x1 = std::min(a.x, b.x);
-    _Tp y1 = std::min(a.y, b.y);
-    a.width = std::max(a.x + a.width, b.x + b.width) - x1;
-    a.height = std::max(a.y + a.height, b.y + b.height) - y1;
-    a.x = x1;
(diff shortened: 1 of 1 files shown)


sovrasov:
It seems not to be a bug but a pre-defined behavior. Let we have the rect1(0,0,0,0), it's actually a point (0,0). And than we are trying to do `|` operation with some random rectangle rect2(1,1, 2, 5). The result by definition from the docs should contain both the point (0,0) and rect2, but not only rect2 as you suppose.

sovrasov:
In terms of the abstract algebra of course ti's better to have the identity element to make `|` operation satisfy group axioms but we are not digging so deep.

piotr-semenov:
@sovrasov According to OpenCV implementation, `cv::Rect(x,y,w,h).contains(cv::Point(a,b))` is true iff a is from [x.x+w) and b is from [y,y+h). So `cv::Rect(0,0,0,0).contains(cv::Point(0,0))` is false as well as `cv::Rect(x,y,0,0)` contains no points for all x and y. In another words, `cv::Rect(x,y,0,0)` denotes the empty point set.
I suggest `cv::Rect` to satisfy the well-known intuitive fact that the union of any set with the empty set is the set itself.

sovrasov:
:+1:

PR #9104, opened 2017-07-06
Added support for thresholding CV_16U images.

bottiger:
I work a lot with images of type CV_64F and CV_16U. Unfortunatly many of OpenCV's functions do not have support for these types, which requires a lot of extra work.

This patch includes support for thresholding CV_16U images. There is no support for HAVE_TEGRA_OPTIMIZATION or HAVE_IPP, but CV_SIMD128 is implemented.

This pull request is mostly a test of how easy it is to get patches merged upstream :)

### DIFF
--- a/modules/imgproc/src/thresh.cpp
+++ b/modules/imgproc/src/thresh.cpp
@@ -269,6 +269,204 @@ thresh_8u( const Mat& _src, Mat& _dst, uchar thresh, uchar maxval, int type )
     }
 }
 
+static void
+thresh_16u(const Mat& _src, Mat& _dst, ushort thresh, ushort maxval, int type)
+{
+	Size roi = _src.size();
+	roi.width *= _src.channels();
+	size_t src_step = _src.step;
+	size_t dst_step = _dst.step;
+
+	if (_src.isContinuous() && _dst.isContinuous())
+	{
+		roi.width *= roi.height;
+		roi.height = 1;
+		src_step = dst_step = roi.width;
+	}
+
+	// HAVE_TEGRA_OPTIMIZATION not supported
+
+	// HAVE_IPP not supported
+
+	int j = 0;
+	const ushort* src = _src.ptr<ushort>();
+	ushort* dst = _dst.ptr<ushort>();
+
+	// CV_SIMD128 not supported
+#if CV_SIMD128
+	bool useSIMD = checkHardwareSupport(CV_CPU_SSE2) || checkHardwareSupport(CV_CPU_NEON);
+	if (useSIMD)
+	{
+		int i;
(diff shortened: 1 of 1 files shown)


alalek:
Thank your for contribution!

1. "no tabs" please. Use "4 spaces" for indentation.
2. Use templates in case of multiple types to avoid code copy-paste of common parts.
3. Please keep single commit in PR (squash commits) and don't add "merge" commits (use git rebase instead).
4. Add tests for new functionality (at least accuracy).

bottiger:
@alalek Thanks for the feedback. I don't have time for that right now, but I'll do it at some point. The only question I have: Where should I add the tests?

alalek:
Data type for threshold tests is selected [here](https://github.com/opencv/opencv/blob/f670a9927026629a4083e05a1612f0adcad7727e/modules/imgproc/test/test_thresh.cpp#L78-L80).

bottiger:
@alalek one more question. You say I should use templates. However, the existing functions are not templated - even though I believe they could take advantage of it (to some degree). Is there a reason for this?

alalek:
Templates are used to remove duplicates of code. For example, this statement "dst[j] = src[j] <= thresh ? maxval : 0;" is used 8 times in this file.
There is [example](https://github.com/opencv/opencv/blob/master/modules/core/src/stat.cpp#L2300-L2317) of used template implementation. Also it is possible to do this via "switch" statement instead of function pointer.

modules/imgproc/src/thresh.cpp
```diff
@@ -269,6 +269,204 @@ thresh_8u( const Mat& _src, Mat& _dst, uchar thresh, uchar maxval, int type )
...
+
+	// HAVE_IPP not supported
+
+	int j = 0;
+	const ushort* src = _src.ptr<ushort>();
+	ushort* dst = _dst.ptr<ushort>();
+
+	// CV_SIMD128 not supported
+#if CV_SIMD128
+	bool useSIMD = checkHardwareSupport(CV_CPU_SSE2) || checkHardwareSupport(CV_CPU_NEON);
```
tomoaki0705: There's a [hasSIMD128()]( https://github.com/opencv/opencv/blob/32d4af36e273dad3d603ba983669d76da946845f/modules/core/include/opencv2/core/hal/intrin_sse.hpp#L1808 ) which checks the feature in better way

vpisarev:
@bottiger, thank you for the patch! Could you please fix some indentation issues found by our buildbot?

bottiger:
@vpisarev Maybe. But unfortunately I am not sure if I have the time at the moment.

vpisarev:
superseded with #9653

PR #9106, opened 2017-07-06
Eliminated protobuf dependency from Caffe importer

dkurt:
### This pullrequest changes
* Caffe importer no longer requires protobuf dependency
* Proposed OpenCV's proto parser
* Removed generated `caffe.pb.cc` and `caffe.pb.h` which takes ~2.0MB. Added `caffe_proto.pb` (20KB) that is binary version of `caffe.proto` (taken from textual file using protobuf compiler by `protoc --descriptor_set_out`). CMake script generates a header file contains binary array (would it be OK or replace static array to vector?).


vpisarev:
@dkurt, I suggest you to remove the word "w i p" from the PR description. We all know that this is experimental patch; but without this keyword your PR will be automatically checked by buildbot as soon as you put new commits.

dkurt:
@vpisarev, would it be OK if I made this usage as a single module @ opencv_contrib for a while? It let us:
* add more parser-specific tests (origin protobuf will be required for a part of tests)
* make softer migration: use parser if found
* reduce protobuf dependency in other modules (currently read mode only)

vpisarev:
@dkurt, I'm fine with it

vpisarev:
@dkurt, let's merge the patch. But could you please fix the merge conflicts first?

modules/dnn/CMakeLists.txt
```diff
@@ -6,16 +6,36 @@ if(DEFINED BUILD_opencv_dnn AND NOT BUILD_opencv_dnn)
   return()
 endif()
 
-include(${OpenCV_SOURCE_DIR}/cmake/OpenCVFindLibProtobuf.cmake)
-if(NOT Protobuf_FOUND)
-  ocv_module_disable(opencv_dnn)
+ocv_option(${the_module}_USE_PROTOBUF "Force use of Protobuf library instead OpenCV's protobuf parser" OFF)
+
+# One and only way if dnn module will be disabled - failed to link to system protobuf.
+if (DEFINED BUILD_PROTOBUF AND NOT BUILD_PROTOBUF)
```
alalek: `if` should be used without space in CMake here and below

modules/dnn/CMakeLists.txt
```diff
@@ -45,36 +65,50 @@ if(ANDROID)
...
-#supress warnings in autogenerated caffe.pb.* files
-add_definitions(-DHAVE_PROTOBUF=1)
-ocv_warnings_disable(CMAKE_CXX_FLAGS
-    -Wunused-parameter -Wundef -Wignored-qualifiers -Wno-enum-compare
-    -Wdeprecated-declarations
-    /wd4125 /wd4267 /wd4127 /wd4244 /wd4512 /wd4702
-    /wd4456 /wd4510 /wd4610 /wd4800
-    -wd858 -wd2196
-)
+if (${USE_PROTOBUF})
```
alalek: `if(USE_PROTOBUF)`

modules/dnn/CMakeLists.txt
```diff
@@ -6,16 +6,36 @@ if(DEFINED BUILD_opencv_dnn AND NOT BUILD_opencv_dnn)
   return()
 endif()
 
-include(${OpenCV_SOURCE_DIR}/cmake/OpenCVFindLibProtobuf.cmake)
-if(NOT Protobuf_FOUND)
-  ocv_module_disable(opencv_dnn)
+ocv_option(${the_module}_USE_PROTOBUF "Force use of Protobuf library instead OpenCV's protobuf parser" OFF)
```
alalek: This name is "_USE_PROTOBUF" because "the_module" is not defined here (before `ocv_add_module()`).

modules/dnn/src/caffe/caffe_importer_v2.cpp
```diff
@@ -0,0 +1,579 @@
...
+        if (node.has(name))
+        {
+            std::vector<T> param(node[name].size());
+            for (int i = 0; i < param.size(); ++i)
+            {
+                node[name][i] >> param[i];
+            }
+            DictValue dict;
+            if (typeid(T) == typeid(int32_t) || typeid(T) == typeid(uint32_t) ||
+                typeid(T) == typeid(int64_t) || typeid(T) == typeid(uint64_t))
```
alalek: Use template specializations / type traits instead.
I believe this "if" must be handled by `DictValue` itself.

modules/dnn/src/caffe/caffe_importer_v2.cpp
```diff
@@ -0,0 +1,579 @@
...
+namespace cv {
+namespace dnn {
+CV__DNN_EXPERIMENTAL_NS_BEGIN
+
+namespace
+{
+
+using namespace pb;
+
+class CaffeImporter : public Importer
```
alalek: Do we really need another one Caffe importer?

modules/dnn/src/caffe/caffe_importer_v2.cpp
```diff
@@ -0,0 +1,579 @@
...
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2017, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective owners.
+
+#ifndef HAVE_PROTOBUF
+#include <typeinfo>
+
+#include "../precomp.hpp"
```
alalek: `precomp.hpp` must be the first include without any `#if`s

modules/dnn/src/caffe/caffe_importer_v2.cpp
dkurt: Thanks! Replaced to `is_integer` and `is_specialized` for non-integers but floats and doubles.

modules/dnn/src/caffe/caffe_importer_v2.cpp
dkurt: Fixed

modules/dnn/CMakeLists.txt
dkurt: Replaced

modules/dnn/CMakeLists.txt
dkurt: Replaced

modules/dnn/CMakeLists.txt
dkurt: Replaced under the `ocv_add_module`.

May I ask you to show version of system protobuf to reproduce it?

Added text descriptors parsing. So now we generate `.pb` from `.proto` using protoc. Then print it as text. I've found only one way to use python:
```python
from google.protobuf import descriptor_pb2

msg = descriptor_pb2.FileDescriptorSet()

with open('/path/to/example.pb', 'rb') as f:
    msg.ParseFromString(f.read())
    print msg
```

modules/dnn/src/caffe/caffe_importer_v2.cpp
dkurt: @alalek, Thanks! It's possible to use a single Caffe importer. I'll do it.

ribalda:
Hi, have you taken a look to: https://github.com/google/protobuf/issues/1489 ?

PR #7797, opened 2016-12-06
Added VS2017 support (MSVC 1910)

patrikhuber:
### This pullrequest changes

Add support for VS2017 (RC is out, final will be very soon). It's identifier is MSVC 1910 and it's vc15.

Probably there are some changes necessary to the CMake `find_package` command: VS2017 is supposedly binary-compatible with VS2015, which means that when using the VS2015 generator, `find_package` should accept binaries created with vc14 as well as vc15, and the VS2017 generator should accept binaries from vc14 as well as vc15 too, but prefer binaries from vc15 (its own version). This can be done in a separate PR.

Even though they're binary compatible it still makes sense to merge this PR which outputs binaries generated with VS2017 into a vc15 folder, since vc15 is its new version number and I think I read in a blog-post that the binaries _do_ differ and some newer features are used if all components that are linked are compiled from VS2017, and no "older" libraries take part in the linking.

So in my opinion it makes sense to output binaries from VS2015 into vc14, and binaries from VS2017 into vc15 - even though they may well be binary compatible.

PR #9110, opened 2017-07-06
Added detection of MSVC1911

patrikhuber:
MSVC1911 is VS2017 Preview (which is an "early preview" of the next changes). The PR makes sure the version is detected properly and the INSTALL target then correctly installs to x64/vc15 (the same as for MSVC1910).

Same changes as done in #7797 for MSVC1910.


patrikhuber:
This could be back-ported to 2.4.x I guess if anyone can be bothered. I'm not using 2.4.x anymore ;-)

alalek:
Thanks! Do you have any information about binary compatibility between MSVS2015 (1900) and MSVS2017 (1910/1911)?

patrikhuber:
Hi @alalek, Yep, I linked this in #7797 already at the end, as per your request. Was the information there not sufficient?
(TLDR: Yes they're binary compatible).

alalek:
I found [this link](https://blogs.msdn.microsoft.com/vcblog/2017/03/07/binary-compatibility-and-pain-free-upgrade-why-moving-to-visual-studio-2017-is-almost-too-easy/):
> VC Runtime in MSVC 2017 is binary compatible with 2015

:+1:

patrikhuber:
@alalek Yep this link is even newer - I guess since December last year MS wrote a couple of blog posts with a bit more official information about it! :-)

PR #9111, opened 2017-07-06
reuse AVX2-optimized kernels for AVX1 CPUs

vpisarev:
E.g. it may be useful on Sandy Bridge or Ivy Bridge. On Ivy Bridge it accelerates SSD network forward pass by ~20%.


alalek:
:+1:

PR #9114, opened 2017-07-07
add libdnn acceleration to dnn module 

pengli:
libdnn provide ocl acceleration to current dnn module.  Currently it provide ocl kernels for 5 layers including convolution, softmax, LRN, fully connect and pooling.  In convolution it implements an auto-tuning mechanism to enumerate all possible convolution kernels and find the best performance one at first run time. So application can use the tuned kernel to achieve best performance.


modules/dnn/src/libdnn/include/greentea_math_functions.hpp
```diff
@@ -0,0 +1,54 @@
+/*
```
vpisarev: Thank you for the contribution, I believe, it will be very valuable!

Just a quick note while you are debugging the OpenCL kernels. We now have much stricter legal requirements, especially for the main repository. This heading comment has no license, has no copyright, we cannot include it into OpenCV in such a form. Besides, we are now regularly running tools to check the code cleanness. So, before the patch is integrated, we need to be sure that the proper header with the license and copyright is in place and the code in the file does not match with non-approved 3rd-party open-source software.

vpisarev:
@pengli, not sure if my comment for one of the files is visible. In brief, we need to make sure that every single contributed file has a proper license, copyright. And the code should be absolutely clean, it should not match code from other 3rd-party open-source software.

modules/dnn/src/libdnn/include/greentea_math_functions.hpp
pengli: @vpisarev , Thanks for your comments, we will check every single file to make sure it is clean and has proper license and copyright.

pengli:
@vpisarev , @alalek  We have updated the source code license and copyright, we also fixed some bugs reported in the buildbot test.  Please take a look at the patchset and provide your comments. Thanks.

vpisarev:
@pengli, thank you a lot! That was a great fight with buildbot and you won! :)

vpisarev:
Since it's quite a big patch in one of most important modules, we will check and test it thoroughly.

modules/dnn/CMakeLists.txt
```diff
@@ -68,7 +71,7 @@ ocv_source_group("Src\\protobuf" FILES ${Protobuf_SRCS} ${Protobuf_HDRS})
 ocv_module_include_directories(include ${Protobuf_INCLUDE_DIRS})
 
 ocv_glob_module_sources(${Protobuf_SRCS} ${Protobuf_HDRS} ${CBLAS_H_PROXY_PATH})
-ocv_create_module(${Protobuf_LIBRARIES} ${LAPACK_LIBRARIES})
+ocv_create_module(${Protobuf_LIBRARIES} ${LAPACK_LIBRARIES} ${OPENCL_LIBRARIES})
```
vpisarev: In OpenCV there is T-API that loads OpenCL runtime dynamically. You should not link OpenCL explicitly. Everything, including vendor extensions, is loaded dynamically.

modules/dnn/src/dnn.cpp
```diff
@@ -638,6 +638,10 @@ struct Net::Impl
         fusion = true;
         preferableBackend = DNN_BACKEND_DEFAULT;
         preferableTarget = DNN_TARGET_CPU;
+#ifdef HAVE_OPENCL
+        if (ocl::useOpenCL())
+            preferableTarget = DNN_TARGET_OPENCL;
+#endif
```
vpisarev: Until it's not tested well on different hardware, I suggest not to make OpenCL the preferable target. Let's have user to set it explicitly.

modules/dnn/src/dnn.cpp
```diff
@@ -1262,7 +1266,8 @@ struct Net::Impl
             allocateLayer(lid, layersShapes);
         }
 
-        fuseLayers(blobsToKeep_);
+        if (0)
+            fuseLayers(blobsToKeep_);
```
vpisarev: this change will seriously affect performance of CPU branch. By "seriously" I mean 50-100%. If you want to disable fusion in the case of OpenCL target, please, do so.

modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -150,6 +150,10 @@ class ConvolutionLayerImpl : public BaseConvolutionLayerImpl
     Ptr<BatchNormLayer> bnorm;
     Ptr<ScaleLayer> scaleLayer;
 
+#ifdef HAVE_OPENCL
+    Ptr<LibDNNConvSpatial<float>> convolutionOp;
```
vpisarev: There should be a space between >>, otherwise some compiler may produce error

modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -646,6 +650,57 @@ class ConvolutionLayerImpl : public BaseConvolutionLayerImpl
...
 
+#ifdef HAVE_OPENCL
+    bool forward_ocl(std::vector<Mat*> &inputs, std::vector<Mat> &outputs, std::vector<Mat> &internals)
+    {
+        int group = inputs[0]->size[1] / blobs[0].size[1];
+
+        if (convolutionOp.empty())
+        {
+            LibDNNConvConfig config;
+            config.in_shape = {inputs[0]->size[0], inputs[0]->size[1], inputs[0]->size[2], inputs[0]->size[3]};
```
vpisarev: it's another piece of C++ 11 code. @alalek, do we check anywhere that the code is C++ 98 compliant?

modules/dnn/src/libdnn/include/math_functions.hpp
```diff
@@ -0,0 +1,85 @@
...
+#ifndef _OPENCV_GREENTEA_MATH_FUNCTIONS_HPP_
+#define _OPENCV_GREENTEA_MATH_FUNCTIONS_HPP_
+#include "../../precomp.hpp"
+#include "common.hpp"
+
+
+#ifdef HAVE_OPENCL
+enum CBLAS_TRANSPOSE {CblasNoTrans=111, CblasTrans=112, CblasConjTrans=113};
+
+template<typename Dtype>
```
vpisarev: I suggest to put all the stuff from libdnn into libdnn namespace. As we have a good chance here, can we, perhaps, change the name to something that would indicate that it's DNN acceleration for OpenCL? There is clDNN, but the name is taken, maybe something similar? opencl4dnn? opencl_dnn_kernels?

modules/dnn/src/libdnn/include/common.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+#ifndef _OPENCV_LIBDNN_COMMON_HPP_
+#define _OPENCV_LIBDNN_COMMON_HPP_
+#include "../../precomp.hpp"
+#include "../../caffe/glog_emulator.hpp"
+#include <opencv2/core/opencl/runtime/opencl_core.hpp>
+
+#ifdef HAVE_OPENCL
+#ifdef USE_INDEX_64
+#define int_tp int64_t
```
vpisarev: could you please do a global rename within your directory instead of introducing aliases for popular types? it looks really confusing

modules/dnn/src/libdnn/src/common.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+void allocateMemory(void** ptr, uint_tp size, int_tp flags)
+{
+    ocl::Context ctx = ocl::Context::getDefault();
+    *ptr = (void*)clCreateBuffer((cl_context)ctx.ptr(),    // NOLINT
+                                 flags,
+                                 size, nullptr, nullptr);
+}
+
+const char* clGetErrorString(cl_int error)
```
vpisarev: @alalek, I believe, we have such functionality already in T-API, right?

modules/dnn/src/libdnn/src/common.cpp
```diff
@@ -0,0 +1,157 @@
...
+#ifdef HAVE_OPENCL
+bool isBeignet()
+{
+    return ocl::Device::getDefault().OpenCL_C_Version().find("beignet") != std::string::npos;
+}
+
+void allocateMemory(void** ptr, uint_tp size, int_tp flags)
+{
+    ocl::Context ctx = ocl::Context::getDefault();
+    *ptr = (void*)clCreateBuffer((cl_context)ctx.ptr(),    // NOLINT
```
vpisarev: Can we use UMat's insted of manual memory handling?

modules/dnn/src/layers/convolution_layer.cpp
alalek: This code is under `#ifdef HAVE_OPENCL`.
We currently have no "OpenCL with C++ 98" precommit builder.
Precommit Linux OpenCL builder forces C++11 (for Halide), similar case for Windows OpenCL. Nightly builder would detect this problem only (after merge).

modules/dnn/src/libdnn/src/common.cpp
alalek: Nope. Prehaps this should go to `core/src/ocl.cpp` and exported via `cv::ocv::getOpenCLErrorString(int errorCode)`.

modules/core/src/ocl.cpp
```diff
@@ -1351,7 +1351,7 @@ struct Context::Impl
                     const String& buildflags, String& errmsg)
     {
         size_t limit = getProgramCountLimit();
-        String key = Program::getPrefix(buildflags);
+        String key = Program::getPrefix(buildflags) + cv::format("%08llx", src.hash());
```
alalek: Nice catch! Thank you!

modules/dnn/src/libdnn/include/benchmark.hpp
```diff
@@ -0,0 +1,73 @@
...
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+#ifndef _OPENCV_LIBDNN_BENCHMARK_HPP_
+#define _OPENCV_LIBDNN_BENCHMARK_HPP_
+#include "../../precomp.hpp"
+#include "common.hpp"
+
+#ifdef HAVE_OPENCL
+class Timer {
```
alalek: Please reuse/adapt [`cv::TickMeter`](https://github.com/opencv/opencv/blob/bdb6b45c91803b6eb1a14d82f42e9a10c6d46487/modules/core/include/opencv2/core/utility.hpp#L277-L391).

alalek: Well, this interface is for OpenCL events. Probably we should integrate this into `cv::ocl::Kernel::run()`.

pengli:
Hi, @vpisarev @alalek ,  

we  have updated the patchset with changes to address the comments.  major changes including
1.  remove int_tp type
2.  change to ocl4dnn namespace
3.  manage ocl4dnn  internal cl buffer with UMat
4.  move getOpenCLErrorString() into cv::ocl
5.  move cl_event  based timer into cv::ocl
6.  make "DNN_TARGET_CPU" as default dnn target, restore layer fusion for CPU target
7.  Add prefer target property in Layer class to indicate derivative layer to call either OCL or CPU path based on the prefer target. (Note. it break ABI test in the buildbot). 

please review patchset  v2 and provide your comments. Thanks

modules/dnn/src/opencl/gemm_buffer.cl
```diff
@@ -0,0 +1,1478 @@
...
+
+#if defined(cl_intel_subgroups)
+#pragma OPENCL EXTENSION  cl_intel_subgroups : enable
+#endif
+
+#define VEC_SIZE        4
+#define LWG_HEIGHT      4
+#define TILE_M          8
+#define TILE_K          16
+#define TILE_N          32
```
alalek: We already have similar code here: https://github.com/opencv/opencv/blob/3.3.0/modules/core/src/opencl/intel_gemm.cl

Can we avoid code duplication?

modules/dnn/src/opencl/math.cl
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+}
+
+__kernel void TEMPLATE(sign,Dtype)(const int n, __global const Dtype* x,
+                                   const int offx, __global Dtype* y,
+                                   const int offy) {
+  for (int index = get_global_id(0); index < n; index += get_global_size(0)) {
+    y[index + offy] = (0.0 < x[index + offx])
+        - (x[index + offx] < 0.0);
```
alalek: What is about [sign()](https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/sign.html) ?

modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -648,6 +656,53 @@ class ConvolutionLayerImpl : public BaseConvolutionLayerImpl
...
+            config.bias_term = (hasBias()) ? true : false;
+            config.weights_backward = false;
+            config.bias_backward = false;
+
+            convolutionOp = Ptr<OCL4DNNConvSpatial<float> >(new OCL4DNNConvSpatial<float>(config));
+        }
+
+        UMat weights, biases;
+        weights = blobs[0].getUMat(ACCESS_READ);
+        if (hasBias()) biases = blobs[1].getUMat(ACCESS_READ);
```
alalek: Probably constant values can be pre-allocated and filled into `cv::UMat` during "preparation" stage.

modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -648,6 +656,53 @@ class ConvolutionLayerImpl : public BaseConvolutionLayerImpl
...
+            UMat inpMat, outMat;
+            inpMat = inputs[ii]->getUMat(ACCESS_READ);
+            outMat = outputs[ii].getUMat(ACCESS_WRITE);
+
+            int batch_size = inpMat.size[0];
+            cl_mem in_mem = (cl_mem)inpMat.handle(ACCESS_READ);
+            cl_mem out_mem = (cl_mem)outMat.handle(ACCESS_WRITE);
+
+            if (!convolutionOp->Forward((float *)in_mem, (float *)weight_mem, (float *)bias_mem,
+                                        (float *)out_mem, batch_size))
```
alalek: Please replace passing of cv_mem via `float*` type to `const cv::UMat&`.

modules/dnn/src/layers/elementwise_layers.cpp
```diff
@@ -468,18 +569,18 @@ struct ChannelsPReLUFunctor
...
                 x2 = v_select(x2 >= z, x2, x2*s4);
                 x3 = v_select(x3 >= z, x3, x3*s4);
-                v_store(ptr + i, x0);
-                v_store(ptr + i + 4, x1);
-                v_store(ptr + i + 8, x2);
-                v_store(ptr + i + 12, x3);
+                v_store(dstptr + i, x0);
+                v_store(dstptr + i + 4, x1);
+                v_store(dstptr + i + 8, x2);
+                v_store(dstptr + i + 12, x3);
```
alalek: Nice catch!

wzw-intel:
Hi @alalek ,
Thanks for your comments.
The cause why so many benchmarks are launched and so long time to run auto-tuning is "Failed to create cache directory".

The auto-tuning mechanism need to store the fine-tuned configuration to local storage (the path is $HOME/.cache/clCaffe/spatialKernels/ ). When the next time to run the convolution with same parameters, the cached configuration will be loaded and there is no need to tune again. 

In your case, “Failed to create cache directory” happens (I think it is because there is no write permission for the path), auto-tuning will be triggered every time when calling forward(). That is why you saw so many benchmarks. 

To fix this issue, we need to find a suitable place, with write permission allowed, to store the fine-tuned configurations. And this solution should be compatible with Linux, windows and MacOS.  Do you know which path in OpenCV is proper to store these tuned configurations?

For the Timer based on OpenCL event + dummy kernels, it makes sure any OpenCL commands sent between ocl::Timer::start() and ocl::Timer::stop() completed. So that this timer is needed to measure the OpenCL kernel execution time.

For other issues, I am still investigating :)

pengli:
Hi, @alalek 

We have pushed new commits to address your comments,  

1) By default convolution auto-tuning is disabled,  user can open it by exporting "OPENCV_OCL4DNN_ENABLE_AUTO_TUNING"  explicitly,  this could reduce "opencv_test_dnn" running time on buildbot dramatically,  a good convolution kernel is created and used by ocl4dnn,  but it may not be the best performance kernel because convolution  auto-tuning is disabled. 
2) user can still enable auto-tuning and run the dnn test by export the global environment above,  it will find the best GFLOPS kernel and save it to $HOME/spatialkernels/,  but this will take a long time at first time.
3)  As @wzw-intel said in above comment,  the ocl::timer implementation is to get ocl kernel run time precisely, auto-tuning mechanism could use it to calculate GFLOPS number of each convolution kernel, cv::TickMeter couldn't meet our requirement. 
4)  "Estimated GFLOPS/S: inf" is fixed. 

Please try the new patchset and provide your feedback. Thanks

gongzg:
@alalek I just pushed a patch to reduce about 75% of the kernel tuning candidates with less than 2% performance drop for the final result. Could you check whether this is acceptable now or we still need to further refine it. One possible further tuning method is to use multipe thread to build those kernel candidates.

gongzg:
@alalek As to the stable performance measuring issue for the auto-tuning mechanism, it is a valid issue but not really a big deal here in my opinion. The way we measure the time is to use the profiling event. The source of those unstable is the system dynamic envinronment. For example, both the CPU side cache or the GPU side cache situation may have some impact. The situation is worse for smaller work load, as a large portion of the time is the system overhead. The real execution time of this type of work load is really very dynamic. So as you already observed, this issue leads to unstable tuning result. However, in practice, for most of the real workload, the performance is dominate by some relatively large convolution layer. For those larger convolution layer, the tuning result get much more stable GFLOPS than tuning the smaller workload. And in practice, we can always get the almost best performance with this method. What's do you think?

gongzg:
@alalek  The warning message is from the OCL compiler, "warning: Linking two modules of different data layouts". Basiclly, it's harmless. Maybe we can report this issue to the compiler team.

alalek:
@gongzg Yes, we observe this OpenCL compiler message with other [tests](http://pullrequest.opencv.org/buildbot/builders/master-lin64/builds/10207/steps/test_core-ippicv-opencl/logs/stdio) too.

pengli:
@alalek ,  we have updated new patches, there is no kernel verification failure anymore in the convolution benchmarking,  and total tuning time is shortened.  As @gongzg said, the compiler warning can be ignored because it doesn't impact the final result.  Please review the new patchset. Thanks

modules/dnn/src/opencl/fillbuffer.cl
```diff
@@ -0,0 +1,63 @@
...
+  for (int_tp index = get_global_id(0); index < n; index += get_global_size(0)) {
+    x[index + offx] = alpha;
+  }
+}
+
+__kernel void TEMPLATE(fill,Dtype)(const int_tp n, const Dtype alpha, __global Dtype* x,
+                                   const int_tp offx) {
+  for (int_tp index = get_global_id(0); index < n; index += get_global_size(0)) {
+    x[index + offx] = alpha;
+  }
```
alalek: `cv::UMat::setTo(cv::Scalar::all(value))` should handle this better (including `clEnqueueFillBuffer` code path).

modules/dnn/src/opencl/math.cl
```diff
@@ -0,0 +1,154 @@
...
+
+__kernel void TEMPLATE(log,Dtype)(const int n, __global const Dtype* a,
+                                  const int offa, __global Dtype* y,
+                                  const int offy) {
+  for (int index = get_global_id(0); index < n; index += get_global_size(0)) {
+    y[offy + index] = log((Dtype)(a[offa + index]));
+  }
+}
+
+__kernel void TEMPLATE(powx,Dtype)(const int n, __global const Dtype* a,
```
alalek: There is already OpenCV functions for functionality of these kernels. There is no strong reason to add duplicates.

modules/dnn/src/opencl/ocl4dnn_lrn.cl
```diff
@@ -0,0 +1,604 @@
...
+      }
+          Dtype tmp_aveval = aveval / pool_size;
+          top_data[index] = bottom_data[index] * pow(tmp_aveval * scale + 1, power);
+    }
+  }
+}
+
+
+// !!!!!!!!
+// The following macros are duplicated in lrn_optimized_layer.cu since we currently
```
alalek: Probably this comment is not relevant anymore

modules/dnn/src/opencl/softmax.cl
```diff
@@ -65,11 +65,13 @@ __kernel void kernel_channel_sum(const int num, const int channels,
...
     const int num, const int channels,
-    const int spatial_dim, __global const T* channel_sum, __global T* data) {
+    const int spatial_dim, const int logsoftmax, __global const T* channel_sum, __global T* data) {
   int index = get_global_id(0);
   if(index < count) {
     int n = index / channels / spatial_dim;
     int s = index % spatial_dim;
-    data[index] /= channel_sum[n * spatial_dim + s];
+    T v = data[index] / channel_sum[n * spatial_dim + s];
+    if (logsoftmax) v = log(v);
```
alalek: `logsoftmax` should be passed via macro definition in build options

modules/dnn/src/opencl/softmax_loss.cl
```diff
@@ -0,0 +1,303 @@
...
+  LossParameter_NormalizationMode_BATCH_SIZE = 2,
+  LossParameter_NormalizationMode_NONE = 3
+};
+#endif
+// Copied from softmax_loss_layer.cpp, must keep consistent with the original implementation
+Dtype TEMPLATE(get_normalizer, Dtype)(
+    enum LossParameter_NormalizationMode normalization_mode, int valid_count,
+    int outer_num_, int inner_num_) {
+  Dtype normalizer;
+  switch (normalization_mode) {
```
alalek: `normalization_mode` should be passed via build macro too

modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp
```diff
@@ -0,0 +1,2000 @@
...
+}
+
+template<typename Dtype>
+ocl::Program OCL4DNNConvSpatial<Dtype>::compileKernel()
+{
+    String errmsg;
+    ocl::Context ctx = ocl::Context::getDefault();
+    ocl::ProgramSource src(kernel_.c_str());
+    ocl::Program program = ctx.getProg(src, options_, errmsg);
+    CV_Assert(!kernel_name_.empty());
```
alalek: Program compilation may fail due unsupported option: `-cl-no-subgroup-ifp` (it is described in OpenCL 2.1 only, OpenCV supports 1.2+ and should not fail on OpenCL 1.1).
1. Using of this option should be optional (or detected separately).
2. OpenCL program compilation errors should be handled properly. (It fails in another place with "Basic kernel failed verification" message)

Problem is observed on this configuration:
```
OpenCL Platforms:
    Intel(R) OpenCL
        iGPU: Intel(R) HD Graphics 520 (OpenCL 2.0 )
        CPU: Intel(R) Core(TM) i5-6300U CPU @ 2.40GHz (OpenCL 2.0 (Build 10094))
Current OpenCL device:
    Type = iGPU
    Name = Intel(R) HD Graphics 520
    Version = OpenCL 2.0
    Driver version = 20.19.15.4463
    Compute units = 24
    Max work group size = 256
    Local memory size = 64 kB
    Max memory allocation size = 805 MB 264 kB
    Double support = No
    Host unified memory = Yes
    Has AMD Blas = No
    Has AMD Fft = No
    Preferred vector width char = 1
    Preferred vector width short = 1
    Preferred vector width int = 1
    Preferred vector width double = 0
```

modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp
```diff
@@ -0,0 +1,2000 @@
...
+        int result = _mkdir(cache_path_.str().c_str());
+#else
+        int result = mkdir(cache_path_.str().c_str(), 0755);
+#endif
+        hasCacheDir = (result == -1) ? false : true;
+    }
+    if (!hasCacheDir) {
+        std::cout << "Failed to create cache directory: "
+                  << cache_path_.str()
+                  << ", disable auto-tuning." << std::endl;
```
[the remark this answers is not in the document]
alalek: BTW, In previous patch it creates 'spatialkernels' in the top-level directory of current disk on Windows.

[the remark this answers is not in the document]
pengli: "spatialkernels" is created to cache the tuned convolution kernel in the benchmarking,  once the tuning is done, convolution kernel is saved under this directory and ocl4dnn can load these kernels directly.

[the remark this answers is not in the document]
alalek: I know this, the problem is **in the top-level directory of current disk** (`C:\spatialkernels`).

modules/dnn/CMakeLists.txt
```diff
@@ -72,7 +75,11 @@ ocv_source_group("Src\\protobuf" FILES ${Protobuf_SRCS} ${Protobuf_HDRS})
 ocv_module_include_directories(include ${Protobuf_INCLUDE_DIRS})
 
 ocv_glob_module_sources(${Protobuf_SRCS} ${Protobuf_HDRS} ${CBLAS_H_PROXY_PATH})
-ocv_create_module(${Protobuf_LIBRARIES} ${LAPACK_LIBRARIES})
+if(APPLE)
+  ocv_create_module(${Protobuf_LIBRARIES} ${LAPACK_LIBRARIES} ${OPENCL_LIBRARIES})
```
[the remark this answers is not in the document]
pengli: Yes, dynamic loading OpenCL library is true for Linux and Windows,  But for Mac, the cmake log on buildbot  is below:
--   OpenCL:                        <Link with OpenCL library>
--     Link libraries:              -framework OpenCL

That's a way for static link,  so if I remove ${OPENCL_LIBRARIES} for APPLE,  the mac build on buildbot will be failed.

modules/core/src/ocl.cpp
```diff
@@ -1846,7 +1847,7 @@ struct Queue::Impl
         if( !dh )
             dh = (cl_device_id)pc->device(0).ptr();
         cl_int retval = 0;
-        handle = clCreateCommandQueue(ch, dh, 0, &retval);
+        handle = clCreateCommandQueue(ch, dh, CL_QUEUE_PROFILING_ENABLE, &retval);
```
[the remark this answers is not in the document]
pengli: yes, this flag is optional, in ocl4dnn we need this flag to get cl profiling info, but it looks that we don't have a way to set this flag through any interface.  Before we use getenv("ENABLE_OPENCL_PROFILING") to control this flag,  but it doesn't look like a good way because buildbot  doesn't recognize this flag.

modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp
```diff
@@ -0,0 +1,2000 @@
...
+                kernel.set(argIdx++, (uint16_t)pad_h_);
+                ocl::Queue queue = ocl::Queue::getDefault();
+                if (config->use_null_local) {
+                    err = clEnqueueNDRangeKernel((cl_command_queue)queue.ptr(),
+                                                 (cl_kernel)kernel.ptr(), 3,
+                                                 NULL,
+                                                 config->global_work_size, NULL, 0, NULL,
+                                                 NULL);
+                } else {
+                    err = clEnqueueNDRangeKernel((cl_command_queue)queue.ptr(),
```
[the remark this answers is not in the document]
pengli: One concern to use kernel.run() is that global work size may be changed in this function, so the ocl kernels may not work properly due to global work size change.

modules/dnn/src/ocl4dnn/src/math_functions.cpp
```diff
@@ -0,0 +1,674 @@
...
+    }
+    delete arg;
+}
+
+// Create and copy buffer to image for GEMM's matrix A and B.
+// Will return image to caller if the input image is NULL. Otherwise,
+// will use the image directly. It's caller's responsibility to
+// release the created image.
+template<typename Dtype>
+void ocl4dnnGEMMCopyBufferToImage(int32_t ctx_id,
```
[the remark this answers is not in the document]
pengli: ocl4dnnGEMMCopyBufferToImage isn't just copy the umat data as is, it may transpose the matrix and add extra padding for purpose.  So cv::ocl::Image2D isn't suitable in this case.

tslgithub:
stop


来自 魅蓝 Note5

-------- 原始邮件 --------
发件人：pengli <notifications@github.com>
时间：周二 9月5日 15:28
收件人：opencv/opencv <opencv@noreply.github.com>
抄送：Subscribed <subscribed@noreply.github.com>
主题：Re: [opencv/opencv] add libdnn acceleration to dnn module (#9114)

modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp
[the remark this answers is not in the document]
pengli: with latest update, The cache dir path on windows is "$USERPROFILE\spatialkernels",  for buildbot, the path is "C:\Users\build\spatialkernels\", so this issue should be fixed.

modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp
```diff
@@ -0,0 +1,2000 @@
...
+template<>
+std::string OCL4DNNConvSpatial<float>::programEntryToString(ocl::ProgramSource& src)
+{
+    return src.source();
+}
+
+template<>
+bool OCL4DNNConvSpatial<float>::createGEMMLikeConvKernel(int32_t blockM,
+                                                         int32_t blockK,
+
```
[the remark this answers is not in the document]
pengli: @alalek , I don't understand this comment,  could you please elaborate more ?

modules/dnn/CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ ocv_warnings_disable(CMAKE_CXX_FLAGS -Wno-shadow -Wno-parentheses -Wmaybe-uninit
 )
 ocv_warnings_disable(CMAKE_CXX_FLAGS /wd4701 /wd4100)
 
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src/ocl4dnn/include ${OPENCL_INCLUDE_DIRS})
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```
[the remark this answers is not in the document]
pengli: For Linux and windows, remove "-std=c++11" in CMAKE_CXX_FLAGS is ok,  code compiles works.  For Mac,  It causes compile failure.  so compiler may not support C++11 well on Mac.  So I removed this flag for windows/Linux and keep this flag for Mac.

modules/dnn/src/layers/concat_layer.cpp
```diff
@@ -166,11 +167,58 @@ class ConcatLayerImpl : public ConcatLayer
         }
     };
 
+#ifdef HAVE_OPENCL
+    bool forward_ocl(std::vector<Mat*> &inputs, std::vector<Mat> &outputs, std::vector<Mat> &internals)
```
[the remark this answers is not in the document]
pengli: yes, we should change the interface later to better support umat.

modules/dnn/src/opencl/softmax_loss.cl
pengli: fixed in new updates.

modules/dnn/src/opencl/softmax.cl
pengli: fixed in new updates.

modules/dnn/src/opencl/math.cl
pengli: fixed in new updates. have removed these kernels and switch to opencv

modules/dnn/src/opencl/fillbuffer.cl
pengli: removed fillbuffer.cl and switch to opencv setTo function

modules/dnn/src/opencl/ocl4dnn_lrn.cl
pengli: comments are cleaned.

modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp
pengli: yes, we have checked this build option based on opencl version and carefully check the program compilation result if it fails.

modules/core/src/ocl.cpp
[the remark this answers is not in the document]
pengli: we add one build option for enabling OpenCL profiling event, it gives user option to control this flag.

modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp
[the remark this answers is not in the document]
alalek: Code context is buggy (it is GitHub bug observed on large patches):
This is for lines like these:
```
float fastestTime = 999999990000000000000000000.0f;
```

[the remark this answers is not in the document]
pengli: Thanks for the info. we have fixed this in the update patch.

pengli:
@alalek , we have provided our feedback on the review comments and update the patchset to address most of them. please review. Thanks.

modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp
```diff
@@ -0,0 +1,2000 @@
...
+    auto_tuning_ = false;
+
+    if (std::getenv("OPENCV_OCL4DNN_KERNEL_CONFIG_PATH")) {
+        cache_path_ << std::getenv("OPENCV_OCL4DNN_KERNEL_CONFIG_PATH");
+    }
+
+    bool hasCacheDir = false;
+#if defined WIN32 || defined _WIN32
+    if (std::getenv("HOME"))
+        cache_path_ << std::getenv("USERPROFILE");
```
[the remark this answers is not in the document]
pengli: fixed in the new update

modules/dnn/src/ocl4dnn/src/math_functions.cpp
```diff
@@ -0,0 +1,674 @@
...
+                 const int32_t M, const int32_t N, const Dtype alpha,
+                 const cl_mem A, const int32_t offA, const cl_mem x,
+                 const int32_t offx, const Dtype beta, cl_mem y,
+                 const int32_t offy)
+{
+    ocl::Context ctx = ocl::Context::getDefault();
+
+    if (ocl::Device::getDefault().type() == CL_DEVICE_TYPE_CPU)
+    {
+        LOG(FATAL) << "Not Support CPU device";
```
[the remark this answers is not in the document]
pengli: fixed in the new update

modules/dnn/src/ocl4dnn/src/math_functions.cpp
```diff
@@ -0,0 +1,674 @@
...
+        // For matrix B with transpose, we need to handle them differently.
+        // As we can't use the sub group block read to get a row easily,
+        // we have to use CL_FLOAT type with read_imagef to get the row.
+        cl_int err;
+        format.image_channel_data_type = CL_FLOAT;
+        desc.image_type = CL_MEM_OBJECT_IMAGE2D;
+        desc.image_width = width;
+        format.image_channel_order = CL_R;
+
+        desc.image_height = height;
```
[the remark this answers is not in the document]
pengli: fixed in the update

modules/dnn/src/layers/fully_connected_layer.cpp
```diff
@@ -55,6 +60,11 @@ class FullyConnectedLayerImpl : public InnerProductLayer
 public:
     enum { VEC_ALIGN = 8 };
 
+#ifdef HAVE_OPENCL
+    Ptr<OCL4DNNInnerProduct<float>> innerProductOp;
```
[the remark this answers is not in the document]
pengli: fixed in the new update

modules/dnn/src/layers/concat_layer.cpp
```diff
@@ -166,11 +167,58 @@ class ConcatLayerImpl : public ConcatLayer
...
+        return true;
+    }
+#endif
+
     void forward(std::vector<Mat*> &inputs, std::vector<Mat> &outputs, std::vector<Mat> &internals)
     {
         CV_TRACE_FUNCTION();
         CV_TRACE_ARG_VALUE(name, "name", name.c_str());
 
+        CV_OCL_RUN((preferableTarget == DNN_TARGET_OPENCL) && ocl::Device::getDefault().isIntel(),
```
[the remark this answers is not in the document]
pengli: fixed in the new update

modules/dnn/src/layers/concat_layer.cpp
```diff
@@ -166,11 +167,58 @@ class ConcatLayerImpl : public ConcatLayer
...
+        int cAxis = clamp(axis, inputs[0]->dims);
+        if (!(cAxis == 1 && outputs[0].dims == 4))
+            return false;
+
+        int bottom_concat_axis;
+        int concat_size = inputs[0]->size[2] * inputs[0]->size[3];
+        int top_concat_axis = outputs[0].size[1];
+        int offset_concat_axis = 0;
+        UMat inpMat, outMat;
+        outMat = outputs[0].getUMat(ACCESS_READ);
```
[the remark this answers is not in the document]
pengli: fixed in the new update

modules/dnn/src/ocl4dnn/src/ocl4dnn_pool.cpp
```diff
@@ -0,0 +1,202 @@
...
+            oclk_max_pool_forward.set(argIdx++, (cl_mem) top_mask.handle(ACCESS_WRITE));
+
+            oclk_max_pool_forward.run(1, global, local, false);
+        }
+        break;
+    case LIBDNN_POOLING_METHOD_AVE:
+        {
+            ocl::Kernel oclk_ave_pool_forward(CL_KERNEL_SELECT("ave_pool_forward"),
+                                              cv::ocl::dnn::ocl4dnn_pooling_oclsrc);
+
```
alalek: Error checks are mandatory (see MacOS builds):
```
if (kernel.empty())
    return false;
```

modules/dnn/src/ocl4dnn/src/ocl4dnn_pool.cpp
```diff
@@ -0,0 +1,202 @@
...
+            oclk_ave_pool_forward.set(argIdx++, pooled_width_);
+            oclk_ave_pool_forward.set(argIdx++, kernel_h_);
+            oclk_ave_pool_forward.set(argIdx++, kernel_w_);
+            oclk_ave_pool_forward.set(argIdx++, stride_h_);
+            oclk_ave_pool_forward.set(argIdx++, stride_w_);
+            oclk_ave_pool_forward.set(argIdx++, pad_h_);
+            oclk_ave_pool_forward.set(argIdx++, pad_w_);
+            oclk_ave_pool_forward.set(argIdx++, (cl_mem) top.handle(ACCESS_WRITE));
+
+            oclk_ave_pool_forward.run(1, global, local, false);
```
alalek: Error checks are mandatory for `kernel.run()` calls

modules/dnn/src/ocl4dnn/src/ocl4dnn_pool.cpp
pengli: Fixed in the update patch. Checks on ocl4dnn kernel create and run are added.

modules/dnn/src/ocl4dnn/src/ocl4dnn_pool.cpp
pengli: Fixed in the update patch.

pengli:
@alalek , new comments have been addressed.  pls review.

pengli:
@alalek , any feedback ?

pengli:
@alalek ,  we have add pushed new patches to fix bugs and address the review comments. please review

alalek:
I pushed several fixes about C++11 and OpenCL kernels generation.

There is a problem with InnerProduct implementation, currently it doesn't work properly:
- there is image objects lifetime problem (`ocl::Image2D` destroys image handle, so it becomes invalid)
- strange conversion to 32S -> 8U
- "transpose" case probably doesn't handled correctly
- many code paths are not covered by tests. Probably we should remove dead code ("transpose" is always false in `OCL4DNNInnerProductConfig`)
Pending fixes are here: https://github.com/alalek/opencv/commits/pr9114_20170919

I would continue to work on Convolution pass, feel free to provide fixes for other parts.

pengli:
@alalek , The InnerProduct bug has been fixed. I have pushed new patch in the commit "replace cl_image with ocl::Image2D".  Also thanks for your patches.

modules/dnn/src/ocl4dnn/src/ocl4dnn_pool.cpp
```diff
@@ -0,0 +1,213 @@
...
+            oclk_max_pool_forward.set(argIdx++, kernel_h_);
+            oclk_max_pool_forward.set(argIdx++, kernel_w_);
+            oclk_max_pool_forward.set(argIdx++, stride_h_);
+            oclk_max_pool_forward.set(argIdx++, stride_w_);
+            oclk_max_pool_forward.set(argIdx++, pad_h_);
+            oclk_max_pool_forward.set(argIdx++, pad_w_);
+            oclk_max_pool_forward.set(argIdx++, ocl::KernelArg::PtrWriteOnly(top));
+            oclk_max_pool_forward.set(argIdx++, mask_idx_.empty() ? 0 : 1);
+            if (mask_idx_.empty())
+                oclk_max_pool_forward.set(argIdx++, NULL);
```
alalek: `(void*)NULL`, because just NULL has `int` type.

pengli: fixed in the update

alalek:
Something wrong was happened with PR - some commits are dropped.
You could grab my backup from September 22 here: https://github.com/alalek/opencv/commits/pr9114_origin

pengli:
@alalek , have restored back to the latest commit.

modules/dnn/src/ocl4dnn/include/default_kernel_config.hpp
```diff
@@ -0,0 +1,290 @@
+#ifndef _OPENCV_OCL4DNN_DEFAULT_KERNEL_CONFIG_HPP_
+#define _OPENCV_OCL4DNN_DEFAULT_KERNEL_CONFIG_HPP_
+#define CONFIG_NUM (sizeof(default_kernel_config) / sizeof(char *) / 2)
+const char *default_kernel_config[] = {
+// Configs for 72 EUs device
+"631d9d8eea647618","6 3 16 2 1 1 16 1 0 ",
```
alalek: Could you expand these hashes? (avoid using of hashes here)

Something like this (probably without driver version until we support version masks):
```
Intel(R) Corporation: Intel(R) HD Graphics (VER=r5.0.63503) CU=24 k5x1_cn32_g1_s1x1_d1x1_b0_in64x64_p2x0_num1_M32
```

How did you handle different names for the same devices on different platforms?
For example, we have this for i7-6700K with Intel OpenCL runtime:
- [Windows](http://pullrequest.opencv.org/buildbot/builders/precommit_opencl/builds/12523/steps/test_dnn-ippicv-opencl/logs/stdio): "Intel(R) HD Graphics 530"
- [Linux](): "Intel(R) HD Graphics"

Are there configurations for "beignet" runtime?

Could you also add OpenCL information for mentioned devices (into comments) from OpenCV tests header?

wzw-intel: -- Could you expand these hashes?
Yes, I will use more readable string to replace hashes.

-- How did you handle different names for the same devices on different platform?
Good question. I will remove the device name and vendor name from the the signature.

--Are there configurations for "beignet" runtime?
No. We recommend to use  "Intel OpenCL Driver" (https://software.intel.com/en-us/articles/opencl-drivers#latest_linux_driver) for the performance consideration. Beignet lacks of some key extension supports that need by convolution kernel.

--Could you also add OpenCL information for mentioned devices (into comments) from OpenCV tests header?
Ok.

alalek:
I have pushed several updates, including changes with kernel cache directory handling:
- don't create any directories - lets user to specify them explicitly
- allow two use configurations from memory and from external directory

There are several confusing messages from tests on Windows:
```
[ RUN      ] OCL_Torch_Importer.ENet_accuracy
Failed setup kernel by config: x = 4 y = 7 z = 16 type = 2
Failed setup kernel by config: x = 4 y = 7 z = 16 type = 2
Failed setup kernel by config: x = 4 y = 7 z = 16 type = 2
Failed setup kernel by config: x = 4 y = 7 z = 16 type = 2
[       OK ] OCL_Torch_Importer.ENet_accuracy (22401 ms)
```

Looks like it is related to preferred kernel size checks:
```
workgroupSize_used = kernel.preferedWorkGroupSizeMultiple();
if (workgroupSize_used != simd_size)
{
    unloadProgram(kernel_name_);
    return false;
}
```

It is very strange because Windows/Linux machines have similar hardware.

wzw-intel:
@alalek Thanks for your code refine. For the message "Failed setup kernel by config: x = 4 y = 7 z = 16 type = 2", as you suspected, it is related to preferred kernel size checks. OpenCL compiler chose an simd size that is not the one kernel source required. Although the hardware is the same , the Windows OpenCL compiler may have different behavior on this case compared with Linux OpenCL compiler. 
@gongzg for more comments.

gongzg:
@alalek For the same computing kernel, different OCL compiler may generate different instructions with different simd size. This is a typical situation we have observed on both Linux system and Windows system. As the auto-tuning mechanism will push the register pressure to the limit. Some times, slightly change in the compiler side may lead to register spilling with wider SIMD size, thus the compiler will fall back to a shorter SIMD width. The root cause of this dynamic check is at that time, not all compilers support the feature to force generate specified simd width code even with register spilling. Now I think it's better to add the following attribution to force the compiler to generate spefified simd width code:

__attribute__((intel_reqd_sub_group_size(SIMD_WIDTH)))

@wzw-intel please change the code a little bit to avoid generating incorrect simd width of code.

modules/dnn/src/ocl4dnn/src/ocl4dnn_softmax.cpp
```diff
@@ -0,0 +1,135 @@
...
+{
+    scale_data_.release();
+}
+
+template<typename Dtype>
+bool OCL4DNNSoftmax<Dtype>::Forward(const UMat& bottom, UMat& top)
+{
+    bool ret = false;
+    ocl::Queue queue = ocl::Queue::getDefault();
+    bool intel_subgroup = 0 && ocl::Device::getDefault().intelSubgroupsSupport();
```
alalek: Looks like, softmax is completely disabled. What is the problem here?

modules/dnn/src/ocl4dnn/src/common.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+using namespace cv;
+
+#ifdef HAVE_OPENCL
+
+bool build_option_check()
+{
+    int major = ocl::Device::getDefault().deviceVersionMajor();
+    int minor = ocl::Device::getDefault().deviceVersionMinor();
+    return (major >= 2) && (minor >= 1);
```
alalek: This condition is not valid for "=>2.1": "3.0" would not pass.

Also SRB5.0 supports "ifp" flag (at least don't fail) but it is still "OpenCL 2.0".
Perhaps we should try build with "ifp" flag and then fallback to build without this problematic flag in case of errors.

modules/dnn/src/ocl4dnn/src/ocl4dnn_inner_product.cpp
```diff
@@ -0,0 +1,115 @@
...
+    N_ = num_output_ = config.num_output;
+    M_ = config.M;
+    K_ = config.K;
+    phase_test_ = config.phase_test;
+    image_copied_ = false;
+
+    // Set up the bias multiplier
+    if (bias_term_)
+    {
+        bias_multiplier_ = UMat(1, M_, 1.0f, CV_32FC1);
```
alalek: Not used.

modules/dnn/src/ocl4dnn/src/ocl4dnn_softmax.cpp
pengli: The subgroup version softmax kernel is enabled in the new commit.

modules/dnn/src/ocl4dnn/src/common.cpp
wzw-intel: A more universal check method is implemented in new commit.

modules/dnn/src/ocl4dnn/src/ocl4dnn_inner_product.cpp
pengli: dropped unused variable in new commit

modules/core/src/ocl.cpp
```diff
@@ -5129,4 +5228,175 @@ bool internal::isCLBuffer(UMat& u)
...
+#ifdef HAVE_OPENCL
+        if (!running())
+        {
+            clWaitForEvents(1, &start_gpu_cl_);
+            clReleaseEvent(start_gpu_cl_);
+            ocl::Kernel kernel("null_kernel_float", ocl::core::benchmark_oclsrc);
+            float arg = 0;
+            clSetKernelArg((cl_kernel)kernel.ptr(), 0, sizeof(arg), &arg);
+            clEnqueueTask((cl_command_queue)queue.ptr(), (cl_kernel)kernel.ptr(), 0,
+                          NULL, &start_gpu_cl_);
```
alalek: Did you try to use events from [clEnqueueMarker()](https://www.khronos.org/registry/OpenCL/sdk/1.1/docs/man/xhtml/clEnqueueMarker.html) (OpenCL 1.1) or preferable [clEnqueueBarrierWithWaitList()](https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clEnqueueBarrierWithWaitList.html) (OpenCL 1.2+) instead of kernel task submitting?

wzw-intel: Thanks for your advice, I will try this in next commit

PR #9116, opened 2017-07-07
(2.4) Added detection of MSVC1911

alalek:
This is VS2017 Preview. It makes sure the version is detected properly and the INSTALL target then correctly installs to x64/vc15 (the same as MSVC1910).

Backport #9110


alalek:
:+1:

PR #9122, opened 2017-07-08
Fix msvc virtual destructor

pthom:
BufferPoolController has a non virtual protected destructor (which is legitimate)

However, Visual Studio sees this as a bug, if you enable more warnings, like below (example with cmake)
```
add_compile_options(/W3)     # level 3 warnings
add_compile_options(/we4265) # warning about missing virtual destructors
```

This is a proposition in order to silence this warning.

See https://github.com/ivsgroup/boost_warnings_minimal_demo for a demo of the same problem
with boost/exception.hpp and Visual Studio 2015.

Hoping this will be useful


alalek:
This destructor is not accessible (it is placed under `protected:`).
Warning suppression is fine.
Thank you! :+1:

Issue #5485, opened 2015-10-10
videoio: enum CAP_MODE_* should be removed


paroj:
`CAP_MODE_*` enum, is supposed to be used as `set(CAP_PROP_MODE, CAP_MODE_YUYV)`.

However this makes `CAP_PROP_MODE` ambiguous:
- the `cap_libv4l` implementation interprets it as "output mode" and  returns YUYV images.
- the `cap_*1394*` implementations interpret it as "transmission mode" and always return BGR images. (even converting gray to BGR)
- the `cap_v4l` implementation also interprets it as "transmission mode", but treats it read only
- other `cap_*` implementations do not support `CAP_PROP_MODE`

looking at the documentation `CAP_PROP_MODE` we find:

> Backend-specific value indicating the current capture mode.

therefore I think the `cap_libv4l` behavior wrong:
1. the call should result in YUYV transmission frames, but still return BGR
2. the currently available API for selecting output format is `CAP_PROP_CONVERT_RGB`, which allows to disable conversion and get the transmission format
3. the user actually should not be able to specify the output format at all - even though it may save a memcpy
   1. this would mean we have to move half of the `cvtColor` API to VideoCapture. `CAP_MODE_YUYV` implies `COLOR_XXXX2YUYV` - so `CAP_MODE_Lab` would make sense as well
   2. lets consider the transmission format is Bayer and the output is RGB. Which demosaicing algorithm should be used? Edge Aware, VNG, the fastest? 
   3. Lets consider the user wants GRAY images and the possible transmission formats are MJPEG and YUYV. Which one would be more efficient for VideoCapture to select?

allowing to specify the output format requires a large amount of logic inside VideoCapture which does not belong there.

However there are advantages in adding a API similar to `cv::IMREAD_GRAYSCALE` to VideoCapture - yet the `CAP_MODE_*` values are the wrong way to do so.

I suggest removing or deprecating the `CAP_MODE_*` enum as long as `cap_libv4l` is still the only implementation understanding it.

JoeHowse:
The point of `CAP_PROP_MODE` is to provide optimized capture of frames in a *known, specified format*.  Importantly, when using this API, the caller does not need to know the camera's capture format, and `set()` may reconfigure the capture format to a better optimized format.  Depending on the camera, the format conversion may be performed in the camera's firmware, which will be a better optimized approach than doing a conversation later in software.

[...]

PR #5458, opened 2015-10-06
fix zero length std::string in putText()

berak:
fix issue #5451

PR #9123, opened 2017-07-08
Support CV_CAP_MODE_GRAY in FFMPEG backend

cristiklein:
### This pullrequest changes

This pullrequests adds support for `CV_CAP_MODE_GRAY` in the FFMPEG videoio backend. This is particularly beneficial for applications that only care about the luminance, such as feature detection. Without this features wasteful conversion might happen, e.g., FFMPEG backend converts video from YUV to RGB then application uses `cvtColor` to convert frame to gray, although this information is already available in the Y channel.


peters:
@alalek 👍

paroj:
you should rather use `CAP_PROP_FORMAT`. see #5485

cristiklein:
Thanks @paroj for your feedback. I went through #5458 and I am confused. The proposal mentions `CAP_PROP_CONVERT` (not `CAP_PROP_FORMAT`), but there is no mention thereof in the nightly documentation: http://docs.opencv.org/master/d4/d15/group__videoio__flags__base.html#gaeb8dd9c89c10a5c63c139bf7c4f5704d

Also, I looked at `CAP_PROP_FORMAT`, but the documentation does not explain how a user is supposed to use this property to set grayscale output. I quickly checked the V4L2 source code and setting this property is not supported.

Can you clarify how I should proceed?

paroj:
yes, `CAP_PROP_CONVERT` has unfortunately been only discussed so far and not implemented.
1. turn off `CAP_PROP_CONVERT_RGB` (default value is true, that means BGR)
    * if source file is grayscale this should be sufficient
2. set `CAP_PROP_FORMAT` to `AV_PIX_FMT_GRAY8` for grayscale conversion
    * alternatively set `CAP_PROP_FOURCC` to (G,R,E,Y) to be backend agnostic. This is what cap_v4l uses. However with ffmpeg it will require some logic to map to `AV_PIX_FMT_GRAY8`. As a bonus this would match the VideoWriter API.

cristiklein:
Thank you for the clarification. Let me give you my opinion about your suggestion:

1. Setting `CAP_PROP_FORMAT` to `AV_PIX_FMT_GRAY8` is sub-optimal for two reasons: First, it requires the user to import an FFMPEG header to have access to the constant. Second, as somebody already mentioned in #5458, grayscale is very common in computer vision, hence it would be desirable to make this setting backend-neutral. As a user, I would very much appreciate writing the same code both for V4L2 in production and FFMPEG in unit testing.

2. `CAP_PROP_FOURCC` is already used by FFMPEG to retrieve the codec of the input video. Giving it different semantics for setting than for getting would be confusing, whereas changing the API and using it for setting "desired output format" feels wrong to me.

Do you feel there are other good options except `CAP_PROP_CONVERT`? I could certainly dedicate some time and implement that for the FFMPEG backend, but given the larger effort, I would rather wait for some momentum to build around it.

paroj:
unfortunately there is no backend neutral property to achieve this right now. That is exactly why `CAP_PROP_CONVERT` is needed.

You could let `CAP_PROP_CONVERT_RGB=false` imply grayscale for FFMPEG, but that would obviously not work for V4L and generally make the situation worse. So I am afraid `CAP_PROP_CONVERT` is the only clean solution.

vpisarev:
@cristiklein, thanks for the contribution! I do not see how the newly added FFMPEG enum values (which are declared in the internal header) relate to the public API. It's quite unsafe solution because  we can easily end up with unsynchronized values and then we get very weird results or crashes

cristiklein:
@vpisarev, I couldn't agree more! However, if you look at the FFMPEG enum values, they are all unrelated to the public API and can easily end up unsynchronised. Hence, I simply opted for maintaining the existing "style" hoping that that would make my patch more acceptable.

I think we both agree that a proper solution would entail making all FFMPEG enum values relate to their public counterpart.

alalek:
I believe because **standalone** ffmpeg wrapper can't use OpenCV's public API headers without additional refactoring (split enums from others things).

asenyaev:

PR #9129, opened 2017-07-10
Fix draw contours

alalek:
Restored old behavior without throwing an exception.

Perhaps we should warn user/developer about invalid data and throw exceptions anyway. So this PR should be closed.


peters:
@alalek I'd rather hit an assert then dealing with silent errors. The static analysis work you guys have done has actually helped me in finding multiple bugs :)

Issue #9113, opened 2017-07-07
[cross-compile][armv7]'ANDROID_LOG_INFO' was not declared in this scope 


samsun639:
- OpenCV => the latest,  git clone https://github.com/opencv/opencv.git
- Operating System / Platform => Ubuntu 14.04.5 LTS
- Compiler => apt-get install gcc-arm-linux-gnueabi gcc-arm-linux-gnueabihf

##### Steps to reproduce
1. cd ~/opencv/platforms/linux
2. mkdir -p build_hardfp
3. cd build_hardfp
4. cmake -DCMAKE_TOOLCHAIN_FILE=../arm-gnueabi.toolchain.cmake ../../..
5. make 

##### ERROR happen#######

> [  3%] Built target libtiff
[  8%] Built target libjpeg
[ 19%] Built target libwebp
[ 22%] Built target libjasper
[ 23%] Built target libpng
[ 30%] Built target IlmImf
[ 37%] Built target libprotobuf
Scanning dependencies of target opencv_core
[ 37%] Building CXX object modules/core/CMakeFiles/opencv_core.dir/src/system.cpp.o
/root/opencv/opencv/modules/core/src/system.cpp: In member function 'void cv::HWFeatures::initialize()':
/root/opencv/opencv/modules/core/src/system.cpp:462:29: error: 'ANDROID_LOG_INFO' was not declared in this scope
         __android_log_print(ANDROID_LOG_INFO, "OpenCV", "cpufeatures library is not avaialble for CPU detection");
                             ^
/root/opencv/opencv/modules/core/src/system.cpp:462:113: error: '__android_log_print' was not declared in this scope
         __android_log_print(ANDROID_LOG_INFO, "OpenCV", "cpufeatures library is not avaialble for CPU detection");
                                                                                                                 ^
/root/opencv/opencv/modules/core/src/system.cpp: In function 'void cv::error(const cv::Exception&)':
/root/opencv/opencv/modules/core/src/system.cpp:929:29: error: 'ANDROID_LOG_ERROR' was not declared in this scope
         __android_log_print(ANDROID_LOG_ERROR, "cv::error()", "%s", buf);
                             ^
/root/opencv/opencv/modules/core/src/system.cpp:929:72: error: '__android_log_print' was not declared in this scope
         __android_log_print(ANDROID_LOG_ERROR, "cv::error()", "%s", buf);
                                                                        ^
[...]

PR #9130, opened 2017-07-10
build: detect Android via '__ANDROID__' macro

alalek:
https://sourceforge.net/p/predef/wiki/OperatingSystems

related #9113


alalek:
:+1:

PR #9131, opened 2017-07-10
Eltwise layer fixes

dkurt:
### This pullrequest changes
* Replace type of element-wise summation weights from int to float
* Fix negative case of global pooling flag
* Halide test for all kinds of element wise layers (sum, product, max)


vpisarev:
@alalek, I think, the patch is fairly important to be put to R2 :+1:

PR #9139, opened 2017-07-11
GSoC - Improving code match_template.py

Cartucho:
### Easy to merge

Fixing Bad indentation. (line 18 to 96)
Adding _ to the unused variables for clarity. (line 77)

PR #9140, opened 2017-07-11
GSoC - Using Pylint to Test the python tutorial codes

Cartucho:
Testing Python codes

* Adding CMake script to check if pylint is installed
* Adding Pylint config file (to choose the tests that are enabled)
* Adding CMake script to samples/python/tutorial_code

Testing: bad-indentation, mixed-indentation, unnecessary-semicolon, unused-variable

Depends on: #9139


Cartucho:
Other tests can be added, like:

- max-line-length (Maximum number of characters on a single line.)

- not-in-loop (Used when break or continue keywords are used outside a loop.)

- function-redefined (Used when a function / class / method is redefined.)

- duplicate-argument-name (Duplicate argument names in function definitions are syntax errors.)

Check out the other possible tests [here](https://docs.pylint.org/en/1.6.0/features.html#id13), they just need to be added to the Pylint config file (pylintrc)!

Cartucho:
@alalek do you think this PR could get reviewed?
I think it would be a great add because this way we would grantee that the codes get tested!

alalek:
@Cartucho I have pushed changes with updated Pylint checks integration (to make it more generic and extend for python samples too). Please take a look.

Cartucho:
We just need a separate PR fixing the current codes. For example, add _ before the unused variable:
 
image

image

Cartucho:
@alalek Should I do a PR fixing the samples codes with the current pylintrc config file?
Do you agree with the current tests (want to add another one)?

alalek:
> separate PR fixing the current codes

Agree.
We need to fix problems before we would enable this check on our CI (currently there is no such check enabled).

> agree with the current tests

Current set of tests looks good to me.

Cartucho:
@alalek I added the fixes:
- samples/python/tutorial_code: #9139 
- samples/python: #9460

Cartucho:
We also need to apply this to samples/dnn, right?
image

alalek:
Well done. Thank you! :+1:

PR #9143, opened 2017-07-11
Fix wrong patterns generation in gen_pattern tool

### LINKED ISSUE #9132: gen_pattern.py generates incorrect checkerboard pattern

- OpenCV => 3.1

##### Detailed description
[gen_pattern.py](https://github.com/opencv/opencv/blob/master/doc/pattern_tools/gen_pattern.py) generates an incorrect checkerboard pattern (which cannot be fixed by tunning the rest of the parameters).
Indeed the squares overlap :
image
But they shouldn't.

I traced back the error to the svg header (produced in svgfig.py) where stroke-width has a non zero value resulting in a border added to each square:
```.svg
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg style="stroke-linejoin:round; stroke:black; stroke-width:0.5pt ....
```
Setting the border to 0
```.svg
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg style="stroke-linejoin:round; stroke:black; stroke-width:0.0pt ....
```
Results in no overlap and further eliminates the round edges caused by 

image
Note that this also eliminates the roundness of the border caused by 
```.svg
<svg style="stroke-linejoin:round; ...
```

It looks like stroke-width is hardcoded several times throughout svgfig.py.
Note: I haven't investigated if that affects the generation of other patterns.

##### Steps to reproduce
Generate a pattern of checkerboard type:
```.bat
python gen_pattern.py -o output.svg -T checkerboard
 ```

sovrasov:
resolves #9132

### DIFF
--- a/doc/pattern_tools/gen_pattern.py
+++ b/doc/pattern_tools/gen_pattern.py
@@ -38,7 +38,7 @@ class PatternMaker:
     r = spacing / self.radius_rate
     for x in range(1,self.cols+1):
       for y in range(1,self.rows+1):
-        dot = SVG("circle", cx=x * spacing, cy=y * spacing, r=r, fill="black")
+        dot = SVG("circle", cx=x * spacing, cy=y * spacing, r=r, fill="black", stroke="none")
         self.g.append(dot)
 
   def makeACirclesPattern(self):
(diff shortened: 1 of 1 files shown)


alalek:
:+1:

PR #9147, opened 2017-07-12
imgproc: use matlab-style fftShift in phaseCorrelate

### LINKED ISSUE #9144: fftShift in Phase correlation does not work properly when faced with odd dimensions (non even)
- OpenCV => 3.1

##### Detailed description
hey 
i was struggling with phase correlation for a while when i tried to get translation shift for some test images i am working with where they gave me irrational results (1214 , -936) instead of (0,0) while it gave perfect result on Matlab for the exact same images
for the sake of defending the opencv legacy i got my hand dirty into the source code  

I noticed that P and C have a reasonable result but C only get "Corrupted" after applying `fftShift `
where **when it try to swap the rectangles  it does not take into account when the dimension of the Mat is not even so the last column and the last row (depend on which dimension is even ) stay in place** 
and there was my problem where my highest response was in the first column Last row which stay in place but now considered in different coordinate and that is what gave me the stupid result 
```C++
                        Mat tmp;
			Mat q0(planes[i], Rect(0, 0, xMid, yMid));
			Mat q1(planes[i], Rect(xMid, 0, xMid, yMid));
			Mat q2(planes[i], Rect(0, yMid, xMid, yMid));
			Mat q3(planes[i], Rect(xMid, yMid, xMid, yMid));

			q0.copyTo(tmp);
			q3.copyTo(q0);
			tmp.copyTo(q3);
			
			q1.copyTo(tmp);
			q2.copyTo(q1);
			tmp.copyTo(q2);
```

##### Steps to reproduce
i applied phase correlation on 2 image with same dimenstion of (1868x2416) which when resized and padded for the optimal DFTSize  using (getOptimalDFTSize) become (1875x2430)
and my highest response (peak shift location) for C after the idft was in lowest row first column (i,j)=>(1874 , 0)

i did not upload the image because i think the bug is clear 

Best Regard
Sabaa

sovrasov:
resolves #9144

### DIFF
--- a/modules/imgproc/src/phasecorr.cpp
+++ b/modules/imgproc/src/phasecorr.cpp
@@ -372,17 +372,18 @@ static void fftShift(InputOutputArray _out)
 
     if(is_1d)
     {
+        int is_odd = (xMid > 0 && out.cols % 2 == 1) || (yMid > 0 && out.rows % 2 == 1);
         xMid = xMid + yMid;
 
         for(size_t i = 0; i < planes.size(); i++)
         {
             Mat tmp;
-            Mat half0(planes[i], Rect(0, 0, xMid, 1));
-            Mat half1(planes[i], Rect(xMid, 0, xMid, 1));
+            Mat half0(planes[i], Rect(0, 0, xMid + is_odd, 1));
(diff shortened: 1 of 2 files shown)


vpisarev:
:+1:

PR #9149, opened 2017-07-12
core(test): fix input data for OCL FP16 test

### LINKED ISSUE #9148: Incorrect input data in OCL convertFp16 tests
In case of FP16 (S16) => FP32 conversion.
Numbers from range `0xfc01 - 0xffff` / `0x7c01 - 0x7fff` are not valid for tests input (NaNs).

alalek:
resolves #9148


alalek:
:+1:

PR #9157, opened 2017-07-13
dshow: Only show debugging messages if environment variable is set

### LINKED ISSUE #8926: '***** VIDEOINPUT LIBRARY - 0.1995 - TFW07 *****' message displayed during startup in debug mode

zhmu:
This will change the DirectShow input module so that it will only display debugging messages if the environment variable OPENCV_DSHOW_SHOW_DEBUG is explicitly set to non-zero.

Only debug mode is affected by this change; the messages are never visible in release mode anyway.

Based on discussion at: https://github.com/opencv/opencv/pull/9051

resolves #8926

### DIFF
--- a/modules/videoio/src/cap_dshow.cpp
+++ b/modules/videoio/src/cap_dshow.cpp
@@ -337,11 +337,17 @@ interface ISampleGrabber : public IUnknown
 #ifdef _DEBUG
 #include <strsafe.h>
 
-//change for verbose debug info
-static bool gs_verbose = true;
-
 static void DebugPrintOut(const char *format, ...)
 {
+    static int gs_verbose = -1;
+    if (gs_verbose < 0)
+    {
+        // Fetch initial debug state from environment - defaults to disabled
+        const char* s = getenv("OPENCV_DSHOW_DEBUG");
+        gs_verbose = s != NULL && atoi(s) != 0;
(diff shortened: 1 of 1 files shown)


alalek:
:+1:

PR #9168, opened 2017-07-16
Open cv fast filter flow

yyxiongzju:
### Adding the fastfilterflow algorithm implementation for detection the motion of objects

I add file README.md for giving introduction and the reference and the implementation has been tested on two flow dataset, MPI Singtel dataset and Middlebury 2011 dataset.
Please make sure glpk library(optimization library for SDP) and OpenMP library installed before running.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,12 +7,14 @@
 * Q&A forum: <http://answers.opencv.org>
 * Issue tracking: <https://github.com/opencv/opencv/issues>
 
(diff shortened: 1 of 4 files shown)


vpisarev:
I believe, this PR is very far from being ready. GLPK is, as far as I understand, is GPL library. Adding dependency of it is not an option. Try to use solveLP() function instead (if it can substitute GLPK for this algorithm). This is new algorithm, so it should rather go to opencv_contrib, not opencv. There should be a regression test. Eigen should not be used unconditionally.

PR #9181, opened 2017-07-18
stitching(perf): check for available OpenCL memory

alalek:
Don't run tests if there is no enough memory.

```
[ RUN      ] OCL_stitch_boat.boat/0
OpenCV Error: Assertion failed (retval == CL_SUCCESS) in cv::ocl::OpenCLBufferPoolImpl::_allocateBufferEntry, file C:\build\master_noICV-win32-vc14\opencv\modules\core\src\ocl.cpp, line 2934
OpenCV Error: Insufficient memory (Failed to allocate 173924352 bytes) in cv::OutOfMemoryError, file C:\build\master_noICV-win32-vc14\opencv\modules\core\src\alloc.cpp, line 52
```


alalek:
:+1:

PR #9188, opened 2017-07-19
MobileNet SSD sample

arrybn:
### This pullrequest changes
Added C++ and python samples for MobileNet-based SSD model

### DIFF
--- a/modules/dnn/src/layers/detection_output_layer.cpp
+++ b/modules/dnn/src/layers/detection_output_layer.cpp
@@ -234,7 +234,7 @@ public:
 
         if (numKept == 0)
         {
-            CV_ErrorNoReturn(Error::StsError, "Couldn't find any detections");
+            return;
         }
         int outputShape[] = {1, 1, (int)numKept, 7};
         outputs[0].create(4, outputShape, CV_32F);
--- a/samples/dnn/mobilenet_ssd_python.py
+++ b/samples/dnn/mobilenet_ssd_python.py
@@ -1,0 +1,87 @@
+import numpy as np
+import argparse
+
+try:
+    import cv2 as cv
+except ImportError:
+    raise ImportError('Can\'t find OpenCV Python module. If you\'ve built it from sources without installation, '
+                      'configure environemnt variable PYTHONPATH to "opencv_build_dir/lib" directory (with "python3" subdirectory if required)')
+
+inWidth = 300
+inHeight = 300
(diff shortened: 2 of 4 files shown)


samples/dnn/ssd_mobilenet_object_detection.cpp
```diff
@@ -0,0 +1,162 @@
...
+        {
+            cout << "Couldn't find camera" << endl;
+            return -1;
+        }
+    }
+
+    Size inVideoSize = Size((int) cap.get(CV_CAP_PROP_FRAME_WIDTH),    //Acquire input size
+                            (int) cap.get(CV_CAP_PROP_FRAME_HEIGHT));
+
+    float WHRatio = inWidth / (float)inHeight;
```
terfendail: Wouldn't it be better to define it near other consts?

vpisarev:
:+1: @alalek, this PR needs to be merged in before 3.3

modules/dnn/src/layers/detection_output_layer.cpp
```diff
@@ -234,7 +234,8 @@ class DetectionOutputLayerImpl : public DetectionOutputLayer
 
         if (numKept == 0)
         {
-            CV_ErrorNoReturn(Error::StsError, "Couldn't find any detections");
+//            CV_ErrorNoReturn(Error::StsError, "Couldn't find any detections");
```
alalek: Perhaps we should remove this line completely.

samples/dnn/mobilenet_ssd_python.py
```diff
@@ -0,0 +1,87 @@
+from __future__ import print_function
+import numpy as np
+import sys
+sys.path.append('/home/arrybn/build/opencv/lib')
```
alalek: Please remove this hard-coded path too. Use `PYTHONPATH` environment variable instead to `cv2.pyd` location.

samples/dnn/ssd_mobilenet_object_detection.cpp
arrybn: fixed

modules/dnn/src/layers/detection_output_layer.cpp
arrybn: fixed

samples/dnn/mobilenet_ssd_python.py
arrybn: fixed

PR #9190, opened 2017-07-19
java: update source files processing, maven stuff

### LINKED ISSUE #9182: Fix OpenCVNativeLoader.java with the correct lib name

- OpenCV => 3.3.0-rc
- Compiler => Any (Java generation)

##### Detailed description

The current opencv/modules/java/generator/src/java/osgi+OpenCVNativeLoader.java file hardcodes the library name to opencv_java320. 
```.java
System.loadLibrary("opencv_java320");
```
  This will be incorrect when 3.3.0 makes actual release, and is incorrect now for anyone building this branch.

This is particularly important because without some code from *within* the package itself calling loadLibrary, OpenCV cannot be used within most OSGi setups (including part of a Eclipse RCP app) because of the way classloading is segregated therein.

##### Steps to reproduce

Observe the source file. Note hardcoded - and currently incorrect - libname
When built, cause OpenCVNativeLoader.init() to be invoked and observe the wrong library loaded
(or an exception to be thrown if the older 3.2.0 library isn't found)

alalek:
resolves #9182

### DIFF
--- a/modules/java/CMakeLists.txt
+++ b/modules/java/CMakeLists.txt
@@ -18,6 +18,12 @@ if(EXISTS ${CMAKE_BINARY_DIR}/gen)
   execute_process(COMMAND ${CMAKE_COMMAND} -E remove_directory "${CMAKE_BINARY_DIR}/gen")
 endif()
 
+if(ANDROID)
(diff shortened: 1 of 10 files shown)


alalek:
:+1:

PR #9192, opened 2017-07-19
Add ability to control bitrate of the video produced by VideoWriter w…

### LINKED ISSUE #8961: Need ability to control bitrate in VideoWriter with H264 codec

- OpenCV => 3.2
- Operating System / Platform => osx

##### Detailed description

There is no bitrate param in constructor of ```VideoWriter``` class in videoio module, however there is ```cv::VIDEOWRITER_PROP_QUALITY``` property in range ```[0...100]``` but it is used only for ```MotionJpegWriter```.  Atm bitrate inside ```CvVideoWriter_FFMPEG``` is hardcoded

https://github.com/opencv/opencv/blob/master/modules/videoio/src/cap_ffmpeg_impl.hpp#L1967

and later is actually reset to 0 for H264

https://github.com/opencv/opencv/blob/master/modules/videoio/src/cap_ffmpeg_impl.hpp#L1533

So the question is what is the best way to control quality of output video file - use ```VIDEOWRITER_PROP_QUALITY``` or pass bitrate directly in constructor?

@alalek I can implement this, but need your suggestions what it the best way to do it.

ysolovyov:
…ith FFMEG

resolves #8961

### This pullrequest changes

Add support to control bitrate of the output video produced by `cv::VideoWriter` with `FFMPEG`


modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -1473,7 +1473,6 @@ static AVStream *icv_add_video_stream_FFMPEG(AVFormatContext *oc,
 
     /* put sample parameters */
     int64_t lbit_rate = (int64_t)bitrate;
-    lbit_rate += (bitrate / 2);
```
ysolovyov: this seems was some hack

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -2036,7 +2032,6 @@ bool CvVideoWriter_FFMPEG::open( const char * filename, int fourcc,
     }
 
     int64_t lbit_rate = (int64_t)c->bit_rate;
-    lbit_rate += (bitrate / 2);
```
ysolovyov: same here

alalek:
Thank you for your effort!

I believe the proper way is to do such changes via [properties](https://github.com/opencv/opencv/blob/3.2.0/modules/videoio/include/opencv2/videoio.hpp#L899) to prevent bloating of methods parameters.
Modern codecs has a lot of parameters and someone would prefer to use qscale / crf / profiles / etc parameter instead of constant bitrate.

Unfortunately these props doesn't work well for changing - bitrate value is required in constructor. This is one of main problems of current VideoIO interfaces design.

ysolovyov:
ye, this only will work in case like that

```
VideoWriter writer;
writer.setProperty(BITRATE);
writer.open(...);
```

setting property after `open` called will not have effect, so does it make sense to rewrite like that or leave as is for now?

There is

```
Ptr<CvVideoWriter> writer;
Ptr<IVideoWriter> iwriter;
```

and `cvCreateVideoWriter_FFMPEG_proxy` comes from `CvVideoWriter` which does not have `setProperty` method, so this should be redone.

vpisarev:
I'd suggest a bit different method that we use in `cv::imwrite()`. Namely, we do have properties there, but we pass them as `std::vector<int>& params=std::vector<int>()`, i.e. a vector of even size `{ param1, value1, param2, value2, ... }`. I believe, it's universal enough solution, more natural and less error-prone - as you correctly noticed, setting a property may not have any effect and that would be a "quiet error".

ysolovyov:
@vpisarev so interface should look like in this case

```
CV_WRAP VideoWriter(const String& filename, int fourcc, double fps,
                Size frameSize, bool isColor = true, const std::vector<int>& params=std::vector<int>());
```
etc

my previous commit failed on buildserver because there were changes in interfaces of high-level functions, there is some limit for that as i understand

PR #9209, opened 2017-07-20
Fix persistence format

### LINKED ISSUE #9198: persistence, all compressed formats generate YAML

- OpenCV => 3.2
- Operating System / Platform => linux
- Compiler => gcc
##### Detailed description

Using CV_STORAGE_FORMAT_AUTO to write a file generates compressed YAML regardless of extension. E.g.  "filename.json.gz" generates a compressed YAML file with that name.

##### Steps to reproduce

Use 'xml.gz' or 'json.gz' on any opencv program that uses the persistance API and uses the default CV_STORAGE_FORMAT_AUTO parameter.

This is caused by the face that the logic determining the format a) doesn't take a potential '.gz' extension into account and b) uses cv_strcasecmp which requires equal length for matching.

The patch below is a minimal invasive fix for the issue.

##### patch
```
diff --git a/modules/core/src/persistence.cpp b/modules/core/src/persistence.cpp
index ccdb135..0ccb784 100644
--- a/modules/core/src/persistence.cpp
+++ b/modules/core/src/persistence.cpp
@@ -126,7 +126,7 @@ static char* icv_itoa( int _val, char* buffer, int /*radix*/ )
     return ptr;
 }
 
-static inline bool cv_strcasecmp(const char * s1, const char * s2)
+static inline bool cv_strcasecmp(const char * s1, const char * s2, bool submatch=false)
 {
     if ( s1 == 0 && s2 == 0 )
         return true;
@@ -135,10 +135,12 @@ static inline bool cv_strcasecmp(const char * s1, const char * s2)
 
     size_t len1 = strlen(s1);
     size_t len2 = strlen(s2);
-    if ( len1 != len2 )
+    if ( !submatch && len1 != len2 )
+        return false;
+    else if ( submatch && len1 < len2 )
         return false;
 
-    for ( size_t i = 0U; i < len1; i++ )
+    for ( size_t i = 0U; i < len2; i++ )
         if ( tolower( static_cast<int>(s1[i]) ) != tolower( static_cast<int>(s2[i]) ) )
             return false;
 
@@ -4225,10 +4227,13 @@ cvOpenFileStorage( const char* query, CvMemStorage* dststorage, int flags, const
         if( fmt == CV_STORAGE_FORMAT_AUTO && filename )
         {
             const char* dot_pos = strrchr( filename, '.' );
+            if (isGZ && cv_strcasecmp( dot_pos, ".gz" ) ) {
+               while(dot_pos-->filename && *dot_pos!='.');
+            }
             fs->fmt
-                = cv_strcasecmp( dot_pos, ".xml" )
+                = cv_strcasecmp( dot_pos, ".xml", true )
                 ? CV_STORAGE_FORMAT_XML
-                : cv_strcasecmp( dot_pos, ".json" )
+                : cv_strcasecmp( dot_pos, ".json", true )
                 ? CV_STORAGE_FORMAT_JSON
                 : CV_STORAGE_FORMAT_YAML
                 ;
```

alalek:
resolves #9198

### DIFF
--- a/modules/core/include/opencv2/core/persistence.hpp
+++ b/modules/core/include/opencv2/core/persistence.hpp
@@ -462,6 +462,11 @@ public:
      */
     static String getDefaultObjectName(const String& filename);
 
+    /** @brief Returns the current format.
+     * @returns The current format, see FileStorage::Mode
+     */
+    CV_WRAP int getFormat() const;
+
     Ptr<CvFileStorage> fs; //!< the underlying C FileStorage structure
     String elname; //!< the currently written element
     std::vector<char> structs; //!< the stack of written structures
--- a/modules/core/src/persistence.cpp
+++ b/modules/core/src/persistence.cpp
@@ -4270,11 +4270,25 @@ cvOpenFileStorage( const char* query, CvMemStorage* dststorage, int flags, const
 
         if( fmt == CV_STORAGE_FORMAT_AUTO && filename )
         {
-            const char* dot_pos = strrchr( filename, '.' );
+            const char* dot_pos = NULL;
+            const char* dot_pos2 = NULL;
+            // like strrchr() implementation, but save two last positions simultaneously
+            for (const char* pos = filename; pos[0] != 0; pos++)
+            {
+                if (pos[0] == '.')
+                {
+                    dot_pos2 = dot_pos;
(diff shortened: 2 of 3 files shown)


alalek:
:+1:

PR #9210, opened 2017-07-21
Fixing explicitly call download method for cuda::GpuMat object

### LINKED ISSUE #6231: Stitcher examples crash with CUDA.
### Please state the information for your system
- OpenCV version: 3.1.0
- Host OS: Linux (Ubuntu 14.04)  / Mac OS X 10.11.3 
- Compiler & CMake: GCC 4.8.4 & CMake 2.8.12.2 / AppleClang 7.0.2.7000181 & CMake 3.5.0
- CUDA: 7.5
### In which part of the OpenCV library you got the issue?
- stitching
- cpp-example-stitching, cpp-example-stitching_detailed
### Issue

On both Linux and OSX stitcher examples crash when gpu is enabled.
### Code example to reproduce the issue / Steps to reproduce the issue

Linux:

```
build/bin/cpp-example-stitching ~/Downloads/k1.jpg ~/Downloads/k2.jpg --try_use_gpu yes
OpenCV Error: The function/feature is not implemented (You should explicitly call download method for cuda::GpuMat object) in getMat_, file /home/paul/opencv/opencv/modules/core/src/matrix.cpp, line 1224
  what():  /home/paul/opencv/opencv/modules/core/src/matrix.cpp:1224: error: (-213) You should explicitly call download method for cuda::GpuMat object in function getMat_

Aborted (core dumped)
```

OSX:

```
bin/cpp-example-stitching ~/Downloads/k1.jpg ~/Downloads/k2.jpg --try_use_gpu yes
OpenCV Error: The function/feature is not implemented (You should explicitly call download method for cuda::GpuMat object) in getMat_, file /Users/paul/work/opencv/opencv_git/modules/core/src/matrix.cpp, line 1224
libc++abi.dylib: terminating with uncaught exception of type cv::Exception: /Users/paul/work/opencv/opencv_git/modules/core/src/matrix.cpp:1224: error: (-213) You should explicitly call download method for cuda::GpuMat object in function getMat_

Abort trap: 6
```

whizzzkid:
resolves #6231

### This pullrequest changes

As mentioned by @sadeqbillah

> There is a bug in the implementation at matchers.cpp GpuMats has been used in cvtColor function directly.


modules/stitching/src/matchers.cpp
```diff
@@ -583,7 +583,10 @@ void SurfFeaturesFinderGpu::find(InputArray image, ImageFeatures &features)
     image_.upload(image);
 
     ensureSizeIsEnough(image.size(), CV_8UC1, gray_image_);
-    cvtColor(image_, gray_image_, COLOR_BGR2GRAY);
+
+    Mat gray_image;
+    cvtColor(image, gray_image, COLOR_BGR2GRAY);
+    gray_image_.upload(gray_image);
```
alalek: Are you sure that `cv::cuda::cvtColor` was called here?

Usage of `cv::Mat` should be avoided in GPU pipelines without strong reason.

sadeqbillah: HI,

In fact the line number 583 has no impact on the code and wasting additional transfer time. 
LN 583:` image_.upload(image);`

Since, the surf has nothing to do with the colored version of the image, we are sending it to GPU unnecessarily. So, I would recommend removing this line as well. In the end, we are compromising the execution time without any benefits.

whizzzkid: Though this solves the problem, thinking about it more now I feel that since surf explicitly needs a grayscale image, converting image to grayscale should not be the job of this function. The best it can do is to assert if the image is grayscale or not. So the change should be in other places which want to use surf matchers.

whizzzkid: Thinking more about it, this just does not feels right. @sadeqbillah this approach might work, but this is wrong. Check this logic:

```
// Colored image uploaded
<cv::cuda::gpuMat> image_.upload(upload)
// converted grayscale on the gpu itself.
cv::cuda::cvtColor(image_, gray_image_, COLOR_BGR2GRAY)
// we do not need to convert to grayscale and upload again.
```

Agreed the problem is here, but this is not the right solution.

whizzzkid: @sadeqbillah while this does not produce any visible errors, it's also not working. For images which work fine on CPU, CUDA gives:

`Camera parameters adjusting failed.`

This definitely not a solution.

sadeqbillah: HI,

I would like to separate the issues in to two:
1. Fix build issue with opencv non-free or community modules (In this case XFeatures2D) [Build TIme]
2. Camera Parameter adjustment failing. [Run Time]

For the first one, trying to change color of a GPU mat with cvtColor definitely stops the build. So, the fix is proper.

For the second case,  CPU and GPU implementation of most of the functions can never be exactly the same due to CUDA nature. From my experience, I have found different outputs for the same set of images with CPU and GPU. This can be due to either different parameters or different implementation. Sometimes algorithms are written differently to avoid performance penalty. I believe the descriptor implementation is different between CPU and GPU. You can try with adjusting some parameters and check.

You can refer to: http://answers.opencv.org/question/72530/difference-between-surf_gpu-and-non-gpu-surf/ or http://opencv-users.1802565.n2.nabble.com/GPU-Differences-in-the-descriptors-of-GPU-and-CPU-implementations-of-SURF-td7519924.html

whizzzkid:
@alalek @sadeqbillah 

after debugging with gdb, I realized that cvtColors being referred to is from imageproc module where as in this case it should be from cudaimageproc.

This should fix this. This might not be the right way to do it or maybe using wrong conventions or maybe wrong style-guide, please correct me and I'll fix the PR. I usually do not work with c++ projects.

Hope it's not a problem
Thanks

whizzzkid:
@alalek can you please review this?

alalek:

PR #9221, opened 2017-07-23
 Code standard

wowotoupi:
The last else should add "{" and "}".


alalek:
Closing as malformed patch.

PR #9225, opened 2017-07-24
calib3d: rewrite reprojectImageTo3D using Matx expressions

paroj:
about the same speed, but significatly more readable

### DIFF
--- a/modules/calib3d/src/calibration.cpp
+++ b/modules/calib3d/src/calibration.cpp
@@ -2780,15 +2780,16 @@ void cv::reprojectImageTo3D( InputArray _disparity,
     Mat _3dImage = __3dImage.getMat();
 
     const double bigZ = 10000.;
-    double q[4][4];
-    Mat _Q(4, 4, CV_64F, q);
+    Matx44d _Q;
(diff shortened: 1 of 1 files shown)


sovrasov:
:+1:

modules/calib3d/src/calibration.cpp
```diff
@@ -2821,42 +2821,36 @@ void cv::reprojectImageTo3D( InputArray _disparity,
...
         if( dtype == CV_16SC3 )
         {
-            short* dptr0 = _3dImage.ptr<short>(y);
-            for( x = 0; x < cols*3; x++ )
+            Vec3s* dptr0 = _3dImage.ptr<Vec3s>(y);
+            for( x = 0; x < cols; x++ )
             {
-                int ival = cvRound(dptr[x]);
-                dptr0[x] = cv::saturate_cast<short>(ival);
+                dptr0[x] = dptr[x];
```
alalek: What is about `cvRound` + `cv::saturate_cast` operations?

sovrasov: They are hidden in conversion Vec3f->Vec3s.

PR #9226, opened 2017-07-24
Bit-exact version of RGB2Luv_b

savuor:
### This pullrequest changes

1. RGB2Luv_b is now calculated in integers which provides bit-exactness across architectures, compilers, etc. Trilinear interpolation is used with LUT calculated using softfloat.
2. Added bit-exactness test for RGB2Luv_b
3. All XYZ-related constants in `color.cpp` are converted to softfloat.

### Performance

Measured 100 runs on random 257x257 image
Repeated 100 times and then averaged

Version | Linux x64 Core i5-6600 @ 3.3 GHz |  Win x64 Core i5-6300U @ 2.4 GHz
------------ | ------------- | -------------
Bit-exact | 0.053 s | 0.133 s
Original | 0.055 s | 0.108 s

### Accuracy

Range of values is 0 to 255

- max abs error: 1
- mean by Luv channels: [0.45, 0.12, 0.08]
- stddev by Luv channels: [0.50, 0.33, 0.27]


vpisarev:
@savuor, thank you! Since we claim that this is bit-exact implementation, we should change our tests as well. They should also use softfloat for initialization and fixed-point in the inner loops. There should be zero difference between reference results and your code in the library.

savuor:
@vpisarev Done. Now there are tests for bit-exactness that calculate CRC of resulting images.
N.B. This should be merged after PR #9374

vpisarev:
:+1:

PR #9232, opened 2017-07-25
Fix FileStorage JSON format

dkurt:
### This pullrequest changes
Parsed JSON nodes were unnamed because of missed flag


alalek:
:+1:

PR #9236, opened 2017-07-26
Fix JSON booleans without quotes

dkurt:
### This pullrequest changes
JSON used to fail to parse bool values without quotes.


alalek:
:+1:

PR #9237, opened 2017-07-26
Fix WinRT build

alalek:
- WinRT compiler doesn't define `WIN32` anymore (`_WIN32` is defined). `WIN32` is not conformant to C/C++ standard, so it was eliminated/replaced.
- added options to build with MSVS 2017 / SDK 10.0


mshabunin:
:+1:

PR #9238, opened 2017-07-26
Valgrind

alalek:
- Split Valgrind suppression file:
  - **valgrind.supp**: OpenCV-related issues (usually, it is non-critical single-time allocated global objects)
  - **valgrind_3rdparty.supp**: This file is used to make Valgrind builder green. It refers to issues from external 3rdparty libraries/runtime/environment/etc. We use here very wide patterns, so be careful with using of these suppressions in your project.
- Install valgrind suppression files into "share/OpenCV" directory (UNIX targets, "dev" component)
- Valgrind runs OpenCV binaries compiled with "-O1 -g" compiler flags (`-O0` is very slow) with disabled IPP/OpenCL optimizations. Also we disable AVX2 code execution: `-DCPU_DISPATCH= -DCPU_BASELINE=AVX`


mshabunin:
:+1:

mshabunin:
:+1:

PR #9241, opened 2017-07-26
core(tls): don't use tlsSlots without synchronization

### LINKED ISSUE #9239: TlsStorage::getData and TlsStorage::setData are not thread safe
- OpenCV => 3.2

##### Detailed description

TlsStorage::getData and TlsStorage::setData are not thread safe. Access to tlsSlots in CVX_Assert is not protected by a mutex lock.

    void* getData(size_t slotIdx) const
    {
        CVX_Assert(tlsSlots.size() > slotIdx);
        ...
    }

    void setData(size_t slotIdx, void* pData)
    {
        CVX_Assert(tlsSlots.size() > slotIdx && pData != NULL);
        ...
    }

##### Steps to reproduce

This bug can lead to data race if we concurrently call two CV functions in which parallel_for_ is used with pthread implementation.

alalek:
resolves #9239

### DIFF
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -328,6 +328,17 @@ Cv64suf;
 #endif
 
 /****************************************************************************************\
+*                                    Thread sanitizer                                    *
+\****************************************************************************************/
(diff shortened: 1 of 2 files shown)


alalek:
:+1:

PR #9247, opened 2017-07-27
Python: wrap Algorithm::read and Algorithm::write

paroj:

### DIFF
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -3093,13 +3093,18 @@ public:
     */
     virtual void write(FileStorage& fs) const { (void)fs; }
 
+    /** @brief simplified API for language bindings
+     * @overload
+     */
+    CV_WRAP void write(const Ptr<FileStorage>& fs, const String& name = String()) const;
+
     /** @brief Reads algorithm parameters from a file storage
     */
-    virtual void read(const FileNode& fn) { (void)fn; }
+    CV_WRAP virtual void read(const FileNode& fn) { (void)fn; }
 
     /** @brief Returns true if the Algorithm is empty (e.g. in the very beginning or after unsuccessful read
      */
-    virtual bool empty() const { return false; }
+    CV_WRAP virtual bool empty() const { return false; }
 
     /** @brief Reads algorithm from the file node
 
--- a/modules/core/src/algorithm.cpp
+++ b/modules/core/src/algorithm.cpp
@@ -55,6 +55,19 @@ Algorithm::~Algorithm()
     CV_TRACE_FUNCTION();
 }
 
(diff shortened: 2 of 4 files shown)


vpisarev:
thank you! it would be useful to provide some test for this functionality

paroj:
I could not find a testing unit for the Algorithm API to add such a test to. It seems the only the Feature2D tests are currently using the Algorithm API at all. The HOGDescriptor provides a similar API, but does not derive from Algorithm.

Creating a new testing unit is beyond the scope of this PR though.

paroj:
or do you mean a pure python test? It would still have to depend on a random non-core algorithm implementing the API.

paroj:
bump

paroj:
@alalek @sovrasov

paroj:
bump

alalek:
Problem of this patch (actually in bindings generator) is here:
> must rename it as it is overloaded in e.g Feature2D

Renaming "read => read2" in base class (not in derived) class doesn't look as appropriate change.
I believe we should fix bindings generator for this problem somehow.

paroj:
> Renaming "read => read2" in base class (not in derived) class doesn't look as appropriate change.

yes, but manual renaming is necessary as python does not support overloading. However we can not rename "read" in derived class as it would break the python API - therefore we must rename it in the base class.

Which alternative behavior would you have in mind? An dispatcher in the bindings that emulates overloading for python?

alalek:
> dispatcher in the bindings that emulates overloading

Emulation already exists in Python bindings, but it doesn't look into base classes.
Actually C++ [doesn't](https://isocpp.org/wiki/faq/strange-inheritance#hiding-rule) look on [overloads](https://stackoverflow.com/questions/1896830/why-should-i-use-the-using-keyword-to-access-my-base-class-method) from [base](https://stackoverflow.com/questions/14212190/c-issue-with-function-overloading-in-an-inherited-class) class too. To enable this we need "using base_class::method_name;" in derived classes.

Also there is another question about Java bindings.

I have some [POC](https://github.com/opencv/opencv/compare/master...alalek:pr9247_r), but it is still very questionable.
Tested on Linux (with python2):
```
(opencv_build)$ (export PYTHONPATH=`pwd`/lib; cd opencv_src_dir/dev/modules/python/test; OPENCV_PYTEST_FILTER=test_algorithm_rw python2 test.py -v)
```

alalek:
@paroj Try to apply this patch: https://github.com/alalek/opencv/commit/pr9247_r2

paroj:
works for me/ python3, linux. Thats an elegant way to trick the parser into doing the right thing. :+1:

alalek:
@paroj Thank you for checking that!
Could you apply these changes to this PR?

PR #9248, opened 2017-07-27
core(alloc): drop unused code, use memalign()

alalek:
Valgrind provides better detection without memory buffer hacks.


alalek:
:+1:

PR #9252, opened 2017-07-27
Support in-memory TIFF encoding and decoding

### LINKED ISSUE #8483: [Feature Request] In-memory TIFF decoding
`TiffDecoder` sets `m_buf_supported` to `false`, but this is unnecessary. `libtiff` supports a `TIFFStreamOpen` which takes a `std::istream *` which could be used to decode tiffs. This would make the use of a tmp file unncessary and should generally be cleaner.

jbms:
Previously, only file-based encoding and decoding were supported with
the libtiff library, leading to the possible use of temporary files.

fixes #8483

### DIFF
--- a/modules/imgcodecs/src/grfmt_tiff.cpp
+++ b/modules/imgcodecs/src/grfmt_tiff.cpp
@@ -160,7 +160,7 @@ public:
                 break;
         }
         new_pos = std::min(new_pos, size);
-        decoder->m_buf_pos = new_pos;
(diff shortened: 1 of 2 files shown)


vpisarev:
@jbms, thank you for the contribution! please, fix the problems found by our buildbot (click "Details" above, or go to http://pullrequest.opencv.org)

jbms:
Errors and warnings are now fixed.

alalek:
:+1:

PR #9253, opened 2017-07-27
loadsave: check for errors when using temp file

### LINKED ISSUE #9251: Tempfile-based image decoding does not properly check for errors

- OpenCV => 06407b4d1446334307db93b3c09a22b27e4b57ce
- Operating System / Platform => n/a
- Compiler => n/a

##### Detailed description

When modules/imgcodecs/src/loadsave.cpp uses a temporary file for decoding an image, it does not properly check the return values of fwrite and fclose.  In the case that there is an error writing, such as if there is insufficient disk space, the file is silently truncated, which in the best case leads to an error decoding the image, and in the worst case, silent data corruption.

jbms:
Previously, the return value of fwrite and fclose were not properly
checked, leading to possible silent truncation of the data if writing
failed, e.g. due to lack of disk space.

Fixes #9251.

### DIFF
--- a/modules/imgcodecs/src/loadsave.cpp
+++ b/modules/imgcodecs/src/loadsave.cpp
@@ -651,8 +651,15 @@ imdecode_( const Mat& buf, int flags, int hdrtype, Mat* mat=0 )
         if( !f )
             return 0;
         size_t bufSize = buf.cols*buf.rows*buf.elemSize();
-        fwrite( buf.ptr(), 1, bufSize, f );
-        fclose(f);
+        if( fwrite( buf.ptr(), 1, bufSize, f ) != bufSize )
+        {
+            fclose( f );
+            CV_Error( CV_StsError, "failed to write image data to temporary file" );
+        }
+        if( fclose(f) != 0 )
+        {
+            CV_Error( CV_StsError, "failed to write image data to temporary file" );
+        }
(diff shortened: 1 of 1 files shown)


modules/imgcodecs/src/loadsave.cpp
```diff
@@ -651,8 +651,14 @@ imdecode_( const Mat& buf, int flags, int hdrtype, Mat* mat=0 )
         if( !f )
             return 0;
         size_t bufSize = buf.cols*buf.rows*buf.elemSize();
-        fwrite( buf.ptr(), 1, bufSize, f );
-        fclose(f);
+        if( fwrite( buf.ptr(), 1, bufSize, f ) != bufSize )
+        {
+            CV_Error( CV_StsError, "failed to write image data to temporary file" );
```
alalek: Please add `fclose(f)` call here (probably without check). `CV_Error()` doesn't return (it throws exception).

What is about to unify errors processing (see line 652 - "return 0;" statement)?

jbms: Good catch, thanks.  I see that this function both returns 0 to indicate an error in places, but elsewhere calls CV_Error, so I think this change at least doesn't make the situation worse.

alalek:
Thank you! :+1:

PR #9254, opened 2017-07-27
Maven build updated package dependency list & package check script.

### LINKED ISSUE #9211: Update package dependencies list used by Maven build.
Attempting to build OpenCV via Maven on Ubuntu 17.04 currently fails due to missing dependencies:

libpng12-dev
libjasper-dev

These packages have also been removed/renamed.

**To reproduce:**
On Ubuntu 17.04 or later checkout  master and from the Maven platform sub directory execute 'mvn clean install' command.
The build will fail with the above packages missing. Searching the repositories yields no results (e.g. aptitude search libpng12-dev).

**Proposed solution:**
Update the package dependency list but maintain backwards compatibility as required.

**Proposed Assignee:**
Myself - PR to be submitted in the coming days.

jtkb:
Resolves #9211 

Updates the package dependencies and allows packages to be marked as optional e.g. python dependencies.

### DIFF
--- a/platforms/maven/opencv/pom.xml
+++ b/platforms/maven/opencv/pom.xml
@@ -67,6 +67,10 @@
                             <workingDirectory>${project.basedir}/scripts</workingDirectory>
                             <arguments>
                                 <argument>deb_package_check</argument>
+                                <argument>-olibpng-dev|libpng12-dev</argument>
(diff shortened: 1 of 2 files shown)


alalek:
:+1:

PR #9255, opened 2017-07-28
Removed _gpu-suffix from usage messages to align with executable name

marting87:
### This pullrequest changes

Removed _gpu-suffix in various GPU samples to align with executable name.


alalek:
:+1:

PR #9259, opened 2017-07-28
Remove unused method declarations in hog.cpp

eyolfson:
The class `App` appears to have two unused methods: `message` and `checkRectSimilarity`. The is no definition or use of either of these methods. This appears to be dead code.


alalek:
Thank you! :+1:

PR #9260, opened 2017-07-28
GSoC - Add Python function signatures to docs

Cartucho:
Using Python help() command to add the Python signatures.

**WIP**

```
buildworker:Docs=linux-2
force_builders=Docs
allow_multiple_commits=1
```


doc/tools/python_signatures.py
```diff
@@ -0,0 +1,152 @@
+# TODO: clarify when there are several C++ signatures and only one Python one for the same function.
+#       i.e. calcHist() - http://docs.opencv.org/trunk/d6/dc7/group__imgproc__hist.html#ga4b2b5fd75503ff9e6844cc4dcdaed35d
+#       special case - http://docs.opencv.org/trunk/db/de0/group__core__utils.html#ga4910d7f86336cd4eff9dd05575667e41
+import codecs
+import re
+import sys
+import cv2
+
+from bs4 import BeautifulSoup
```
vrabaud: surround this with a try catch. If it fails with an ImportError, print something and return right away.

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,152 @@
+# TODO: clarify when there are several C++ signatures and only one Python one for the same function.
+#       i.e. calcHist() - http://docs.opencv.org/trunk/d6/dc7/group__imgproc__hist.html#ga4b2b5fd75503ff9e6844cc4dcdaed35d
```
vrabaud: use links to a stable version like 3.2

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,152 @@
...
+    for link in soup.findAll('a'):
+        tmp_href = link.get('href')
+        if is_module_link(tmp_href):
+            href_list.append(tmp_href)
+    return href_list
+
+
+def add_item(new_row, color, text):
+    new_item = soup.new_tag('td')
+    if color == "red":
```
vrabaud: why is that ? Maybe "red" will change in the future. Insert a comment at least

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,152 @@
...
+
+def add_signatures(soup, tmp_dir):
+    print tmp_dir
+    sign_counter = 0
+    python_sign_counter = 0
+
+    for function in soup.findAll("h2", {"class": "memtitle"}):
+        function_name = function.getText()
+
+        if "()" not in function_name:
```
vrabaud: comment why that is

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,152 @@
...
+            continue
+        else:
```
vrabaud: no need for an else here

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,152 @@
...
+            if "[" in function_name:
```
vrabaud: comment what is happening here.

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,152 @@
...
+                python_sign_counter+=1
+            except AttributeError:
+                continue
+
+    with open(tmp_dir, "w") as file:
+        file.write(str(soup))
+        file.close()
+    print "Added ["+str(python_sign_counter)+\
+          "/"+str(sign_counter)+"] Python signatures"
+    pass
```
vrabaud: why pass ?

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+ROOT_DIR = sys.argv[1]
+soup = load_html_file(ROOT_DIR + "index.html")
+href_list = get_links_list(soup)
+
+for link in href_list:
+    # add python singatures to the module
```
vrabaud: typo: Python signature

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,152 @@
...
+
+    with open(tmp_dir, "w") as file:
+        file.write(str(soup))
+        file.close()
+    print "Added ["+str(python_sign_counter)+\
+          "/"+str(sign_counter)+"] Python signatures"
+    pass
+
+
+ROOT_DIR = sys.argv[1]
```
vrabaud: get anything from here to below in and "if __name__ == "__main__":)

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,152 @@
...
+soup = load_html_file(ROOT_DIR + "index.html")
+href_list = get_links_list(soup)
+
+for link in href_list:
+    # add python singatures to the module
+    soup = load_html_file(ROOT_DIR + link)
+    sub_href_list = get_links_list(soup)
+    add_signatures(soup, ROOT_DIR + link)
+
+    # add python sigantures to the sub-modules
```
vrabaud: same typo

doc/tools/python_signatures.py
vrabaud: Use this instead: http://docs.opencv.org/3.2.0/d6/dc7/group__imgproc__hist.html#ga4b2b5fd75503ff9e6844cc4dcdaed35d

doc/CMakeLists.txt
```diff
@@ -163,14 +163,31 @@ if(BUILD_DOCS AND DOXYGEN_FOUND)
...
   )
+  
+  if(PYTHON2_EXECUTABLE)
+    add_custom_target(doxygen_python    
+      COMMAND ${PYTHON2_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/tools/python_signatures.py" "${CMAKE_CURRENT_BINARY_DIR}/doxygen/html/"
+      DEPENDS doxygen_cpp
+    )
+  endif()
+
+  if(PYTHON2_EXECUTABLE)
```
vrabaud: merge that part with if(PYTHON2_EXECUTABLE) above.

vrabaud:
@vpisarev , this looks pretty good to me: do you want us to push hard to get this into OpenCV 3.3 ? That would help document the Python API.
@Cartucho  is also almost done with the similar Java version.

@Cartucho , can you please add a screenshot on how it looks like now for a random function ? Thx

Cartucho:
Example:
image

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,159 @@
...
+            python_sign_counter+=1
+        except AttributeError:
+            continue
+
+    with open(tmp_dir, "w") as file:
+        file.write(str(soup))
+        file.close()
+    logging.debug("Added ["+str(python_sign_counter)+\
+          "/"+str(sign_counter)+"] Python signatures")
+    pass
```
vrabaud: no need for pass

doc/tools/python_signatures.py
```diff
@@ -0,0 +1,159 @@
...
+        try:
+            method = getattr(cv2, str(function_name))
+            signature = str(method.__doc__)
+            cpp_table = append_table_to(cpp_table, soup, "Python:", signature, function_name)
+            python_sign_counter+=1
+        except AttributeError:
+            continue
+
+    with open(tmp_dir, "w") as file:
+        file.write(str(soup))
```
vrabaud: please use the prettify() function

vrabaud:
The build now works but it would require to have numpy and beautifulsoup installed to get it to work.

woodknight:
Does this work with Python 3? Because in CMakeLists.txt I only see entry for Python 2.

Cartucho:
@woodknight I will add the Python 3 entry as soon as possible, thank you!

Cartucho:
@woodknight Now it should work fine for Python3. Give it a try!

woodknight:
@Cartucho hmmm I just tried, not sure what I did wrong, but the built doc does not have Python signatures.

Cartucho:
@woodknight Did you install BeautifulSoup (bs4)?

woodknight:
@Cartucho Yes. I use anaconda and it has beautifulsoup4 (4.6.0) installed.

woodknight:
@Cartucho Just tried with the recent update, and it works perfectly on my Windows machine.
py

Cartucho:
@alalek could you review this PR?

alalek:
@Cartucho Thank you! This works.

Lets make doc generation process more robust and minimize set of required dependencies.
- `__doc__` strings via `import cv2`:
  - cv2 is heavy dependency itself for documentation build
  - class methods information is not available
  - perhaps it is better to export and use Python signatures information directly instead of parsing them
- Python generator should be idempotent
- generator is rely on internal doxygen structure (this may be a problem in the future, probably we should add detection for unsupported layouts)

I have added PR with python signatures exporting into file (#9566) and added a prototype with updated doc generator based on mentioned and your PR (it still contains "dead code" and "debug statements"): https://github.com/alalek/opencv/commits/pr9260_r

Could you take a look on this and improve proposed changes:
- function signatures selection should be improved (correct namespace extraction)
- support class methods
- support `CV_WRAP_AS()`, like "FileStorage.getNode".
- fix duplication of Python signatures for all overloads or split them
- Java code paths are probably broken

alalek:
@Cartucho Any thoughts?

Cartucho:
@alalek Tomorrow I will be back home, I will have a look then!

Cartucho:
@alalek Maybe a way to deal with the Classes & selection would be to separate the JSON file into multiple files in a folder. For example:

- pyopencv_signatures_**cv**.json (would contain, i.e: **cv**::solveP3P, **cv**::undistort, ...)
- pyopencv_signatures_cv_**dnn**.json (i.e: cv::**dnn**::Layer, cv::**dnn**:Net::setParam)
- pyopencv_signatures_cv_**ml**.json (i.e: cv::**ml**:SVM::setGamma)
- ...

Then in the `python_signatures.py` we could add the [links of the classes](http://docs.opencv.org/trunk/annotated.html) that exist in the generated files (cv::dnn, ...).

About the constants `CV_WRAP_AS()`, are they being added to the JSON file I couldn't find them? If so, we could also add a pyopencv_constants.json file.

In `html_functions.py` in the function `add_signatures` if we keep track of the previous `function_name` we can avoid overloads with a `continue` if the name remains the same. (The functions are ordered alphabetically in the HTML's)

The Java signatures will get broken, but if we generate a Java JSON file just like the Python one the code would be the same.

What do you think?

alalek:
I believe we could use single file - the current JSON is not large (~200kb), so there is no problem to handle all this information. Also "dynamic" set of files would introduce unnecessary complexity into build scripts. So file split is not required here, but we could add additional information (for example, module name).

Constants are not handled at all in the current version.
`CV_WRAP_AS()` is added for functions/methods like [this](https://github.com/opencv/opencv/blob/3.3.0/modules/core/include/opencv2/core/persistence.hpp#L411).

The same note is for classes - they are not added yet. Classes has two forms:
- old (probably deprecated and would not be available in documentation): `BRISK_create(...)`
- new: `BRISK()`


> we can avoid overloads

Agree. Alternatively we could mark "already" used signatures on the same document page.

Cartucho:
So lets generate 3 files:
pyopencv_signatures.json
pyopencv_constants.json ([here](https://github.com/opencv/opencv/blob/master/modules/python/src2/gen2.py#L847))
pyopencv_classes.json ([here](https://github.com/opencv/opencv/blob/597b3f4456255ad87f774fdf9dfca3cfec252961/modules/python/src2/gen2.py#L827))

I believe avoiding overloads by keeping track of the previous signature will be faster.

Cartucho:
@alalek do you agree?

alalek:
Agree. I still believe that we can use single JSON file, but this question is not critical and multiple files would work well.

Cartucho:
@alalek you are right, lets keep it to a single file.
One more thing, don't you think that bs4 is also a strong dependency?

alalek:
> bs4 is also a strong dependency

bs4 is external dependency, but I don't see appropriate easy ways how to replace it. So I believe we could keep use it.

PR #9566 has been merged.
I have rebased experimental patches on the current master: https://github.com/alalek/opencv/commits/pr9260_r
No extra changes are done (I have no spare time to work on it in the next two weeks).
Feel free to grab and adopt it with discussed changes (including json file changes).

alalek:
BTW, Take a look on the "tags" file generated with #9773 . Probably it could be useful for scanning.

alalek:
I suggest to avoid scanning and use information provided by Doxygen in "opencv.tag" file (need to rebase this patch on latest master branch).

For example, this information is available for class `cv::ml::TrainData`:
```
  <compound kind="class">
    <name>cv::ml::TrainData</name>
    <filename>dc/d32/classcv_1_1ml_1_1TrainData.html</filename>
    <member kind="function" virtualness="virtual">
      <type>virtual</type>
      <name>~TrainData</name>
      <anchorfile>dc/d32/classcv_1_1ml_1_1TrainData.html</anchorfile>
      <anchor>a698c4ad95ae9cace589e361f16e9dc83</anchor>
      <arglist>()</arglist>
    </member>
    <member kind="function" virtualness="pure">
      <type>virtual int</type>
      <name>getCatCount</name>
      <anchorfile>dc/d32/classcv_1_1ml_1_1TrainData.html</anchorfile>
      <anchor>a7e687b7ee8325380bced49f5cd5baf15</anchor>
      <arglist>(int vi) const =0</arglist>
    </member>
    ...
```

For global enums / functions there is:
```
  <compound kind="namespace">
    <name>cv</name>
    <filename>d2/d75/namespacecv.html</filename>
    ...
    <member kind="enumeration">
      <type></type>
      <name>ImreadModes</name>
      <anchorfile>d4/da8/group__imgcodecs.html</anchorfile>
      <anchor>ga61d9b0126a3e57d9277ac48327799c80</anchor>
      <arglist></arglist>
    </member>
    <member kind="enumvalue">
      <name>IMREAD_UNCHANGED</name>
      <anchorfile>d4/da8/group__imgcodecs.html</anchorfile>
      <anchor>gga61d9b0126a3e57d9277ac48327799c80aeddd67043ed0df14f9d9a4e66d2b0708</anchor>
      <arglist></arglist>
    </member>
    <member kind="enumvalue">
      <name>IMREAD_GRAYSCALE</name>
      <anchorfile>d4/da8/group__imgcodecs.html</anchorfile>
      <anchor>gga61d9b0126a3e57d9277ac48327799c80ae29981cfc153d3b0cef5c0daeedd2125</anchor>
      <arglist></arglist>
    </member>
    <member kind="enumvalue">
      <name>IMREAD_COLOR</name>
      <anchorfile>d4/da8/group__imgcodecs.html</anchorfile>
      <anchor>gga61d9b0126a3e57d9277ac48327799c80af660544735200cbe942eea09232eb822</anchor>
      <arglist></arglist>
    </member>
    ...
    <member kind="function">
      <type>void</type>
      <name>Canny</name>
      <anchorfile>dd/d1a/group__imgproc__feature.html</anchorfile>
      <anchor>ga04723e007ed888ddf11d9ba04e2232de</anchor>
      <arglist>(InputArray image, OutputArray edges, double threshold1, double threshold2, int apertureSize=3, bool L2gradient=false)</arglist>
    </member>
    <member kind="function">
      <type>void</type>
      <name>Canny</name>
      <anchorfile>dd/d1a/group__imgproc__feature.html</anchorfile>
      <anchor>ga2a671611e104c093843d7b7fc46d24af</anchor>
      <arglist>(InputArray dx, InputArray dy, OutputArray edges, double threshold1, double threshold2, bool L2gradient=false)</arglist>
    </member>
   ...
  </compound>
```

I have prepared [sample script doxygen_scan.py](https://github.com/opencv/opencv/files/1442875/doxygen_scan.py.txt) (drop `.txt` extension - GitHub limitation) to extract `C++ name => file#anchor` information.
Need to add filtering for available C++ names only (from python_signatures.json).
Optionally we should match function args somehow on python "variants" (or try to store C++ arguments in python_signatures.json file).

alalek:
This PR is rebased on the latest master branch.

Cartucho:
Don't review/merge yet, I still have to deal with the methods, constants and classes.

Cartucho:
@alalek do you know if for example the:  class | cv::CLAHE in (here)[https://docs.opencv.org/master/d7/dbd/group__imgproc.html], is being scanned by `classes = root.findall("./compound[@kind='class']")` ?

alalek:
Hi,

class `cv::CLAHE` is found during doxygen scan, but looks like name is malformed for some reason ("NcvCTprep::cv::CLAHE"):
```
Class: NcvCTprep::cv::CLAHE => d6/db6/classcv_1_1CLAHE.html
    Method: cv::CLAHE::apply => d6/db6/classcv_1_1CLAHE.html#a4e92e0e427de21be8d1fae8dcd862c5e
    Method: cv::CLAHE::collectGarbage => d6/db6/classcv_1_1CLAHE.html#acfabd6cd7e533ddd11fa5b3591c07286
```
(this is output of doxygen_scan script)

Source `opencv.tag` file looks good:
```
  <compound kind="class">
    <name>cv::CLAHE</name>
    <filename>d6/db6/classcv_1_1CLAHE.html</filename>
    <base>cv::Algorithm</base>
    <member kind="function" virtualness="pure">
      <type>virtual void</type>
      <name>apply</name>
      <anchorfile>d6/db6/classcv_1_1CLAHE.html</anchorfile>
      <anchor>a4e92e0e427de21be8d1fae8dcd862c5e</anchor>
      <arglist>(InputArray src, OutputArray dst)=0</arglist>
    </member>
    <member kind="function" virtualness="pure">
      <type>virtual void</type>
      <name>collectGarbage</name>
      <anchorfile>d6/db6/classcv_1_1CLAHE.html</anchorfile>
      <anchor>acfabd6cd7e533ddd11fa5b3591c07286</anchor>
      <arglist>()=0</arglist>
    </member>
```

Looks like there is bug:
```.diff
for c in classes:
    c_name = c.find("./name").text
-    name = ns_name + '::' + c_name
+    name = c_name
```

Cartucho:
@alalek yeah, I noticed that.

Based on your proposed structure I'm going through all the anchors of a file and adding the functions/methods/..., one by one, that have a Python equivalent to the documentation.

Since there are no real anchors for the classes I was thinking of using the `<filename>.../.html</filename>` as an "anchor". The problem is that, for example, there is no direct relation between [group__imgproc.html](https://docs.opencv.org/master/d7/dbd/group__imgproc.html) and the Class `cv::CLAHE` .

alalek:
@Cartucho I pushed updates for this patch, results are available here: http://pullrequest.opencv.org/buildbot/export/pr/9260/docs/

Cartucho:
@alalek Nice!

Now we only need to match the Python and C++ arguments like for example in [VideoWriter](http://pullrequest.opencv.org/buildbot/export/pr/9260/docs/dd/d9e/classcv_1_1VideoWriter.html#ad59c61d8881ba2b2da22cff5487465b5). Right?

alalek:
Right, but currently I have no idea how to do this properly (robust way).
My suggestion is to merge this "as is" - this patch is large enough. And then we could try preparing fixes via separate PRs.

`opencv.tag` has some information about arguments, but it should be parser properly (including templates, default values, etc):
```
    <member kind="function">
      <type></type>
      <name>VideoWriter</name>
      <anchorfile>dd/d9e/classcv_1_1VideoWriter.html</anchorfile>
      <anchor>ac3478f6257454209fa99249cc03a5c59</anchor>
      <arglist>(const String &amp;filename, int fourcc, double fps, Size frameSize, bool isColor=true)</arglist>
    </member>
```

Cartucho:
Sounds good! I agree with the merging.

PR #9261, opened 2017-07-29
Add tested codecs for OSX to Python Docs

puqeko:
### This pullrequest changes

Add a list of tested codec's on OSX to python docs.
Also formated inline code.


alalek:
Thank you! :+1:

PR #9274, opened 2017-07-31
core(stat): disable IPP optimization in meanStdDev (cn > 1)

alalek:


vpisarev:
:+1: @lupustr3, looks like this is correct fix, right?

lupustr3:
@vpisarev, @alalek  right, thanks!

PR #9284, opened 2017-08-01
Make floating point literals to be of type "float" in dnn module OpenCL code.

ipuustin:
In `activations.cl`, make the type of floating point literals to be `float`. Otherwise the literals will be interpreted as type `double`, causing Beignet (https://wiki.freedesktop.org/www/Software/Beignet/) to have type conversion issues.


alalek:
:+1:

PR #9285, opened 2017-08-01
Added ELU and test for it

### LINKED ISSUE #9223: Exponential linear unit's support for cv.dnn
Hello!
During trying to convert my tensorflow-keras model to cv2.dnn I found that there is no support of [ELU activation](https://arxiv.org/pdf/1511.07289.pdf) function.
I got following error:
`OpenCV Error: Unspecified error (Unknown layer type Elu in op conv2d_127/Elu) in populateNet, file /home/user/Documents/projects/opencv/modules/dnn/src/tensorflow/tf_importer.cpp, line 375
conv2d_127/Elu:Elu(conv2d_127/convolution)
...
`
So, I kindly ask to add that activation function to cv2.dnn module

arrybn:
resolves #9223


dkurt:
:+1:

PR #9287, opened 2017-08-01
Unit tests for TensorFlow importer

dkurt:
### This pullrequest changes
Several unit tests with generated inputs/outputs/networks from TensorFlow framework.
* Fix padding mode
* Element-wise addition, multiplication
* One more concatenation layer type
* Padding layer (very limited usage: only one side padding, no more than one padding per layer)
* Some activation layers (Abs, Sigmoid, TanH)
* Batch normalization

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/358

### DIFF
--- a/modules/dnn/src/layers/convolution_layer.cpp
+++ b/modules/dnn/src/layers/convolution_layer.cpp
@@ -183,7 +183,7 @@ public:
         }
         else
         {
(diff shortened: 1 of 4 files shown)


arrybn:
:+1:

PR #9288, opened 2017-08-01
Honor fixing principal point in camera calib tutorial.

rpavlik:
### This pullrequest changes

the c++ camera calibration tutorial/sample, to honor the `calibFixPrincipalPoint` option that may be passed in the config file for a fisheye model calibration, which was previously only honored for non-fisheye calibrations despite being available in the fisheye model.


sovrasov:
:+1:

PR #9292, opened 2017-08-01
Using environment variable to store capture options

### LINKED ISSUE #9290: ffmpeg capture forced to TCP, even if UDP is desired
- OpenCV => latest(master)
- Operating System / Platform => Windows 10 Creators 64 Bit
- Compiler => Visual Studio 2015 64

##### Detailed description

I have an rtsp server which crashes after prolonged TCP connection. I have no way of avoiding this device, and I don't have control to fix the crash. UDP is needed. However, after bug "fix" [http://code.opencv.org/issues/2235#note-18](http://code.opencv.org/issues/2235#note-18) ffmpeg capture is hardcoded to TCP.

##### Steps to reproduce
Normal opening of RTSP stream using cv::VideoCapture

```.cpp
this->Capture = std::make_shared<cv::VideoCapture>(this->VideoURL, this->RequestedCaptureAPI);
this->Capture->read(*this->Frame)
```

adamrankin:
String is parsed by av_dict_parse_string(ENV{OPENCV_FFMPEG_CAPTURE_OPTIONS}, ";", "|")

resolves #9290 

### This pullrequest changes

Allows configuration of ffmpeg capture options via environment variable. If variable is not set or OpenCV cannot read env variables, defaults to TCP as per previous behaviour.

### DIFF
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -780,7 +780,7 @@ bool CvCapture_FFMPEG::open( const char* _filename )
     ic->interrupt_callback.opaque = &interrupt_metadata;
 #endif
 
-#if LIBAVFORMAT_BUILD >= CALC_FFMPEG_VERSION(52, 111, 0)
+#if LIBAVFORMAT_BUILD > CALC_FFMPEG_VERSION(51,11,0)
 #ifndef NO_GETENV
     char* options = getenv("OPENCV_FFMPEG_CAPTURE_OPTIONS");
     if(options == NULL)
@@ -789,7 +789,11 @@ bool CvCapture_FFMPEG::open( const char* _filename )
     }
     else
     {
+#if LIBAVUTIL_BUILD > CALC_FFMPEG_VERSION(52,6,0)
         av_dict_parse_string(&dict, options, ";", "|", 0);
+#else
+        av_dict_set(&dict, "rtsp_transport", "tcp", 0);
+#endif
     }
 #else
     av_dict_set(&dict, "rtsp_transport", "tcp", 0);
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -780,7 +780,7 @@ bool CvCapture_FFMPEG::open( const char* _filename )
     ic->interrupt_callback.opaque = &interrupt_metadata;
 #endif
 
-#if LIBAVFORMAT_BUILD > CALC_FFMPEG_VERSION(51,11,0)
+#if LIBAVFORMAT_BUILD >= CALC_FFMPEG_VERSION(52, 111, 0)
 #ifndef NO_GETENV
     char* options = getenv("OPENCV_FFMPEG_CAPTURE_OPTIONS");
     if(options == NULL)
@@ -789,7 +789,7 @@ bool CvCapture_FFMPEG::open( const char* _filename )
     }
     else
     {
-#if LIBAVUTIL_BUILD > CALC_FFMPEG_VERSION(52,6,0)
+#if LIBAVUTIL_BUILD >= (LIBAVUTIL_VERSION_MICRO >= 100 ? CALC_FFMPEG_VERSION(52, 17, 100) : CALC_FFMPEG_VERSION(52, 7, 0))
         av_dict_parse_string(&dict, options, ";", "|", 0);
 #else
         av_dict_set(&dict, "rtsp_transport", "tcp", 0);
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -789,7 +789,7 @@ bool CvCapture_FFMPEG::open( const char* _filename )
     }
     else
     {
-        av_dict_parse_string(&dict, options, ";", "|");
+        av_dict_parse_string(&dict, options, ";", "|", 0);
     }
 #else
     av_dict_set(&dict, "rtsp_transport", "tcp", 0);
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -781,7 +781,19 @@ bool CvCapture_FFMPEG::open( const char* _filename )
 #endif
 
 #if LIBAVFORMAT_BUILD >= CALC_FFMPEG_VERSION(52, 111, 0)
+#ifndef NO_GETENV
+    char* options = getenv("OPENCV_FFMPEG_CAPTURE_OPTIONS");
+    if(options == NULL)
+    {
+        av_dict_set(&dict, "rtsp_transport", "tcp", 0);
+    }
+    else
+    {
+        av_dict_parse_string(&dict, options, ";", "|");
+    }
+#else
     av_dict_set(&dict, "rtsp_transport", "tcp", 0);
+#endif
     int err = avformat_open_input(&ic, _filename, NULL, &dict);
 #else
     int err = av_open_input_file(&ic, _filename, NULL, 0, NULL);


adamrankin:
Hello, I'm having trouble identifying which version of ffmpeg is used for the build. I set up a linux build of FFMPEG/OpenCV and I don't get any compile issues.

Any help would be appreciated.

berak:
[from here](http://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/11829/steps/cmake/logs/stdio) (quite at the bottom):

```
--   Video I/O:
--     DC1394 1.x:                  NO
--     DC1394 2.x:                  NO
--     FFMPEG:                      YES
--       avcodec:                   YES (ver 54.35.1)
--       avformat:                  YES (ver 54.20.4)
--       avutil:                    YES (ver 52.3.0)
--       swscale:                   YES (ver 2.1.1)
--       avresample:                NO
--     GStreamer:                   NO
```

alalek:
These builders are based on Ubuntu 14.04 (failed) and Ubuntu 16.04 (passed).
You need to determine related API changes from [ffmpeg](https://github.com/FFmpeg/FFmpeg/blob/master/doc/APIchanges) and libav and specify proper "#ifdef" guards to use `av_dict_parse_string`.

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -780,8 +780,24 @@ bool CvCapture_FFMPEG::open( const char* _filename )
     ic->interrupt_callback.opaque = &interrupt_metadata;
 #endif
 
-#if LIBAVFORMAT_BUILD >= CALC_FFMPEG_VERSION(52, 111, 0)
+#if LIBAVFORMAT_BUILD > CALC_FFMPEG_VERSION(51,11,0)
```
alalek: Why is this change here?
- Condition must be ">=". ">" should not be used at all for version checks.
- Perhaps we should check for [53.4.0](https://github.com/FFmpeg/FFmpeg/blob/c79e7534712fa9e95a9ddde2c5e8c6815f5f2f6f/doc/APIchanges#L1986) (avformat_open_input)
- These change must be propagated to many places with "dict" (av_dict_free / declaration / etc)

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -780,8 +780,24 @@ bool CvCapture_FFMPEG::open( const char* _filename )
...
+#ifndef NO_GETENV
+    char* options = getenv("OPENCV_FFMPEG_CAPTURE_OPTIONS");
+    if(options == NULL)
+    {
+        av_dict_set(&dict, "rtsp_transport", "tcp", 0);
+    }
+    else
+    {
+#if LIBAVUTIL_BUILD > CALC_FFMPEG_VERSION(52,6,0)
```
alalek: ["lavu 52.17.100 / 52.7.0"](https://github.com/FFmpeg/FFmpeg/blob/c79e7534712fa9e95a9ddde2c5e8c6815f5f2f6f/doc/APIchanges#L1462)
```
#if LIBAVUTIL_BUILD >= (LIBAVUTIL_VERSION_MICRO >= 100 ? CALC_FFMPEG_VERSION(52, 17, 100) : CALC_FFMPEG_VERSION(52, 7, 0))
```

modules/videoio/src/cap_ffmpeg_impl.hpp
adamrankin: Ah crap, this is a mistake. Sorry about the sloppy commits.

modules/videoio/src/cap_ffmpeg_impl.hpp
adamrankin: Thanks, I am unfamiliar with the ffmpeg versioning. Fixes incoming.

modules/videoio/src/cap_ffmpeg_impl.hpp
adamrankin: With regards to 53.4.0, that makes sense to me. However, I am already making mistakes with regards to ffmpeg version checks, so ideally I would make as few changes as possible.

adamrankin:
New code usage could be as follows:

````.cpp
#if WIN32
  _putenv_s("OPENCV_FFMPEG_CAPTURE_OPTIONS", "rtsp_transport;udp");
#else
  setenv("OPENCV_FFMPEG_CAPTURE_OPTIONS", "rtsp_transport;udp");
#endif

auto capture = cv::VideoCapture(someURL, cv::CAP_FFMPEG);

#if WIN32
  _putenv_s("OPENCV_FFMPEG_CAPTURE_OPTIONS", "");
#else
  unsetenv("OPENCV_FFMPEG_CAPTURE_OPTIONS");
#endif
````

**Comment by PhilLab** [comment]
Definitely interesting 👍 Good finding! Is there a place in the documentation where your latest comment could be added?

**Comment by adamrankin** [comment]
Hmm, that does seem cleaner. The only pro to this method (I was inspired by another bug report or a similar topic) is that now you can pass any option(s) to ffmpeg.

I'm happy to write it up, but I'm not familiar with the docs. I'd appreciate any pointer to the right spot!

**Comment by PhilLab** [comment]
Hm, both ways have their pros/cons. I'm unsure which one to prefer.
Regarding the enum way, I would place it like the [device-specific flags](docs.opencv.org/master/dc/dfc/group__videoio__flags__others.html) which are documented by ``//!<`` (cf. the [videoio.hpp](https://github.com/opencv/opencv/blob/c95a97389db5d511c5ef8b7269d0f37ca5677a6d/modules/videoio/include/opencv2/videoio.hpp))

**Comment by adamrankin** [comment]
I think I would stay with the current implementation. Is this mergeable?

**Comment by vpisarev** [comment]
how robust is the solution? Assume that there are 2 apps on a system, one just uses opencv + ffmpeg to read videos, another one reads videos via network and so the options are relevant. This solution is global and so the first app will stop working. Shall we come up with some better solution?

**Comment by adamrankin** [comment]
@vpisarev I'm open to other solutions, but this solution is only global if the env variable is set permanently. Given my code example above, the options can be set in the code (more advanced code could cache any pre-existing options, and then restore them after connecting) without applying to other programs.

**Comment by alalek** [comment]
Environment variables have process scope, so solution/workaround is fine until we have some configuration API in OpenCV.
Looks good to me :+1:

**Comment by golden0080** [comment]
Hey, Guys, I'm happy to find out that TCP issue is fixed in this patch. But only by reading code, I'm a little bit confused how to tell OpenCV to use UDP whenever reading RTSP streams. Can you guys kindly provide a simple example?

**Comment by adamrankin** [comment]
You'll have to set an environment variable when creating the device

```
#if WIN32
  _putenv_s("OPENCV_FFMPEG_CAPTURE_OPTIONS", "rtsp_transport;udp");
#else
  setenv("OPENCV_FFMPEG_CAPTURE_OPTIONS", "rtsp_transport;udp");
#endif

auto capture = cv::VideoCapture(someURL, cv::CAP_FFMPEG);

#if WIN32
  _putenv_s("OPENCV_FFMPEG_CAPTURE_OPTIONS", "");
#else
  unsetenv("OPENCV_FFMPEG_CAPTURE_OPTIONS");
#endif
```

think-free:
This solution doesn't work on windows when using a dynamic library. Something like [this](https://stackoverflow.com/questions/25164478/setting-environmrnt-variables-for-dlls-in-ms-windows) seems to occure.
Ffmpeg parameters should be passed as parameters or configured by calling a function before the open.

The use case is when mixing several rtsp stream with mixed protocol (udp,tcp) in the same executable. The environment variable set by code isn't reflected in the opencv library.

Works fine on linux ...

adamrankin:
What a pain, thank you for catching this. I recommend opening a new issue to change the env var approach!

osmantoker:
> You'll have to set an environment variable when creating the device
> 
> ```
> #if WIN32
>   _putenv_s("OPENCV_FFMPEG_CAPTURE_OPTIONS", "rtsp_transport;udp");
> #else
>   setenv("OPENCV_FFMPEG_CAPTURE_OPTIONS", "rtsp_transport;udp");
> #endif
> 
> auto capture = cv::VideoCapture(someURL, cv::CAP_FFMPEG);
> 
> #if WIN32
>   _putenv_s("OPENCV_FFMPEG_CAPTURE_OPTIONS", "");
> #else
>   unsetenv("OPENCV_FFMPEG_CAPTURE_OPTIONS");
> #endif
> ```

hi, how can I set this env. on Java OpenCV?

PR #9294, opened 2017-08-02
Added function to get timings for layers

arrybn:
### This pullrequest changes
Added function to get tick counts for layers in model inference

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -47,6 +47,7 @@
 #include <iostream>
 #include <sstream>
 #include <iterator>
+#include <numeric>
 #include <opencv2/dnn/shape_utils.hpp>
 #include <opencv2/imgproc.hpp>
 
@@ -633,7 +634,7 @@ struct Net::Impl
         inpl.layerInstance = netInputLayer;
         layerNameToId.insert(std::make_pair(inpl.name, inpl.id));
 
-        lastLayerId = 1;
+        lastLayerId = 0;
         netWasAllocated = false;
         fusion = true;
         preferableBackend = DNN_BACKEND_DEFAULT;
@@ -656,6 +657,7 @@ struct Net::Impl
 
     bool netWasAllocated;
     bool fusion;
+    std::vector<int64> layersTimings;
 
     void compileHalide()
     {
@@ -716,6 +718,8 @@ struct Net::Impl
         it = layers.find(0);
         CV_Assert(it != layers.end());
         it->second.skipFlags[DNN_BACKEND_DEFAULT] = true;
+
+        layersTimings.clear();
     }
 
     void setUpNet(const std::vector<LayerPin>& blobsToKeep_ = std::vector<LayerPin>())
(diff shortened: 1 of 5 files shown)


modules/dnn/src/dnn.cpp
```diff
@@ -1269,6 +1273,8 @@ struct Net::Impl
             allocateLayer(lid, layersShapes);
         }
 
+        layersTimings.resize(lastLayerId + 1, 0);
```
dkurt: We have some bug with `lastLayerId` value - it initializes with 1 but must be initialized by 0 (because initially we have single input layer with id == 0). Please, replace `lastLayerId = 1;` to `lastLayerId = 0;` at `Net::Impl` constructor.

modules/dnn/src/dnn.cpp
```diff
@@ -1299,6 +1308,10 @@ struct Net::Impl
             }
         }
 
+        tm.stop();
+
+        layersTimings[ld.id] = tm.getTimeTicks();
```
dkurt: I think we should zero it manually for skipped layers as mentioned in documentation.

modules/dnn/src/dnn.cpp
arrybn: fixed

modules/dnn/src/dnn.cpp
arrybn: fixed

modules/dnn/src/dnn.cpp
dkurt: The last one thing, method `getLayerNames` returns all layers except data layer but `getPerfProfile` not. So, it seems better for me if we skip data layer in both methods.

arrybn: fixed

dkurt:
:+1: 

python sample:
```python
import cv2 as cv
import numpy as np

net = cv.dnn.readNetFromCaffe('bvlc_alexnet.prototxt', 'bvlc_alexnet.caffemodel')

inp = np.random.standard_normal([1, 3, 227, 227]).astype(np.float32)
net.setInput(inp)
net.forward()

freq = cv.getTickFrequency() * 1e-3
t, timings = net.getPerfProfile()
print t / freq
for name, t in zip(net.getLayerNames(), timings / freq):
    print name, t
```

```bash
17.713957
conv1 [ 1.92816]
relu1 [ 0.]
norm1 [ 0.424336]
pool1 [ 0.061396]
conv2 [ 1.817547]
relu2 [ 0.]
norm2 [ 0.416599]
pool2 [ 0.066611]
conv3 [ 1.478516]
relu3 [ 0.]
conv4 [ 1.069023]
relu4 [ 0.]
conv5 [ 0.93981]
relu5 [ 0.]
pool5 [ 0.059967]
fc6 [ 6.085307]
relu6 [ 0.]
drop6 [ 0.000408]
fc7 [ 2.671139]
relu7 [ 0.]
drop7 [ 0.000192]
fc8 [ 0.674861]
prob [ 0.020085]
```

PR #9302, opened 2017-08-03
Resolves #9191: Grabcut relearns GMM when called with GC_EVAL

### LINKED ISSUE #9191: Providing the GC_EVAL flag to cv::grabcut and providing already trained models always performs GMM learning
- OpenCV => 3.2.0.dev
- Compiler => g++6

##### Detailed description

This issue originates from this StackOverflow question: [Reusing models from grabcut in OpenCV](https://stackoverflow.com/questions/45195532/reusing-models-from-grabcut-in-opencv)

The issue that the original question poser (OP) asked was to be able to reuse the trained GMM models after performing an initial segmentation with `cv::grabcut`.  The OP is currently using the Python wrappers.  The OP noticed that the input foreground and background GMM models are supposed to be initialized to zero arrays and are updated once the segmentation is complete.  The OP would like to reuse these foreground and background GMM models so that they do not need to be learned again and the segmentation can be performed without the GMM learning.  

The OP has tried to use the `GC_EVAL` flag without success as the model gets cleared upon invocation of the `cv2.grabcut` function, which is a wrapper for `cv::grabcut`.  

We investigated the source: https://github.com/opencv/opencv/blob/master/modules/imgproc/src/grabcut.cpp#L574 and noticed that the GMM learning is still performed regardless of what flag is provided.  If `GC_EVAL` is provided, there should not be any GMM learning involved and the models should be used as is.  Additionally, if a pre-trained model is provided, there shouldn't be a requirement to specify a `cv::Rect` or mask to denote what the foreground model is as the models are already provided into the function.

##### Steps to reproduce

This is the source using Python from the StackOverflow post.  This is a slight modification to the source code seen in the tutorial, but using pre-trained models: http://docs.opencv.org/3.2.0/d8/d83/tutorial_py_grabcut.html.  I have not made a pull request as I am unsure as to whether the `learnGMMs` issue with `GC_EVAL` is the only thing that needs fixing.  Otherwise, one can simply append a simple `if` statement checking for `GC_EVAL` prior to calling `learnGMMs`.  Someone on StackOverflow experienced this issue in the past however, and this is what they eventually did as well: [OpenCV - GrabCut with custom foreground/background models
](https://stackoverflow.com/questions/23806064/opencv-grabcut-with-custom-foreground-background-models)

```.py
from matplotlib import pyplot as plt
import numpy as np
import cv2

img = cv2.imread('usimg1.jpg')
mask = np.zeros(img.shape[:2], np.uint8)
bgdModel = np.load('bgdmodel.npy')
fgdModel = np.load('fgdmodel.npy')

cv2.grabCut(img, mask, None, bgdModel, fgdModel, 5, cv2.GC_EVAL)

mask = np.where((mask==2) | (mask==0), 0, 1).astype('uint8') 
img = img * mask[:, :, np.newaxis]

plt.imshow(img)
plt.show()
```

ipa:
**Merge with extra:** https://github.com/opencv/opencv_extra/pull/359

### Resolves #9191: Grabcut relearns GMM when called with GC_EVAL

Issue #9191: When grabcut is called with GC_EVAL it should not relearn the GMM provided but rather use them to run the algorithm. 

The code was changed as discussed in the [Stackoverflow question](https://stackoverflow.com/questions/45195532/reusing-models-from-grabcut-in-opencv), such that if GC_EVAL is used the learnGMM method is not called.

Tests were also modified to check that after calling grabcut with GC_EVAL, the background and foreground models are not changed.

### DIFF
--- a/modules/imgproc/test/test_grabcut.cpp
+++ b/modules/imgproc/test/test_grabcut.cpp
@@ -99,12 +99,12 @@ void CV_GrabcutTest::run( int /* start_from */)
 	bgdModel.copyTo(exp_bgdModel);
 	fgdModel.copyTo(exp_fgdModel);
     grabCut( img, mask, rect, bgdModel, fgdModel, 2, GC_EVAL );
-	
+
     // Multiply images by 255 for more visuality of test data.
     if( mask_prob.empty() )
     {
 		mask.copyTo( mask_prob );
-        imwrite(string(ts->get_data_path()) + "grabcut/mask_prob1.png", mask_prob);
+        imwrite(string(ts->get_data_path()) + "grabcut/mask_prob.png", mask_prob);
     }
     if( exp_mask1.empty() )
     {
--- a/modules/imgproc/test/test_grabcut.cpp
+++ b/modules/imgproc/test/test_grabcut.cpp
@@ -96,25 +96,35 @@ void CV_GrabcutTest::run( int /* start_from */)
     mask = Scalar(0);
     Mat bgdModel, fgdModel;
     grabCut( img, mask, rect, bgdModel, fgdModel, 0, GC_INIT_WITH_RECT );
+	bgdModel.copyTo(exp_bgdModel);
+	fgdModel.copyTo(exp_fgdModel);
     grabCut( img, mask, rect, bgdModel, fgdModel, 2, GC_EVAL );
-
+	
     // Multiply images by 255 for more visuality of test data.
     if( mask_prob.empty() )
     {
-        mask.copyTo( mask_prob );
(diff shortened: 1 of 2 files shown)


modules/imgproc/test/test_grabcut.cpp
```diff
@@ -96,25 +96,35 @@ void CV_GrabcutTest::run( int /* start_from */)
     mask = Scalar(0);
     Mat bgdModel, fgdModel;
     grabCut( img, mask, rect, bgdModel, fgdModel, 0, GC_INIT_WITH_RECT );
+	bgdModel.copyTo(exp_bgdModel);
```
terfendail: Could you please use spaces instead of tabs for indentation

vpisarev:
replaced with #11339

PR #9303, opened 2017-08-03
AKAZE: update

alalek:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/360


alalek:
:+1:

hrnr:
thanks a lot for fixing all these issues. Your fix for uninitialized memory is far better than mine.

BTW isn't it quite hard to maintain all those descriptors for AKAZE? Maybe droping upright descriptors, which are not rotation invariant (and therefore hardly useful) might make things easier in the future.

PR #9304, opened 2017-08-03
perf_tests: disable sanity check in tests without generated data

mshabunin:
__WIP__

### This pullrequest changes

Some performance tests do not have sanity check data, hence they fail when option `--perf_verify_sanity` is provided.
I've disabled sanity check for such tests.


mshabunin:
Invalid PR

PR #9305, opened 2017-08-03
Make dnn networks creation by Importer is deprecated

dkurt:
### This pullrequest changes
`CV_DEPRECATED` added for `createTorchImporter`, `createCaffeImporter`, `createTensorflowImporter` functions. `Importer` class was marked too until we replace it internally.

**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1310

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -598,23 +598,27 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         Ptr<Impl> impl;
     };
 
-    /** @brief Small interface class for loading trained serialized models of different dnn-frameworks. */
+    /**
+     * @deprecated Deprecated as external interface. Will be for internal needs only.
+     * @brief Small interface class for loading trained serialized models of different dnn-frameworks. */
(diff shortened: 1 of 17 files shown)


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -583,23 +583,27 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
     class CV_EXPORTS_W Importer : public Algorithm
     {
     public:
 
         /** @brief Adds loaded layers into the @p net and sets connections between them. */
-        CV_WRAP virtual void populateNet(Net net) = 0;
+        CV_DEPRECATED CV_WRAP virtual void populateNet(Net net) = 0;
 
-        virtual ~Importer();
+        CV_DEPRECATED virtual ~Importer();
```
vpisarev: I think, destructors, which are called implicitly most of the time, should not be declared as deprecated

dkurt: Thanks! Fixed.

vpisarev:
thank you! :+1:

PR #9310, opened 2017-08-04
Parallel SVM::trainAuto: uses parallel_for_ to dispatch all parameters combination …

r2d3:
SVM::trainAuto is just 6 loops. I create a list of tasks and compute them using parallel_for_

### This pullrequest changes

Add parallel functionality to SVM::trainAuto

### DIFF
--- a/modules/ml/src/svm.cpp
+++ b/modules/ml/src/svm.cpp
@@ -1636,6 +1636,99 @@ public:
         return true;
     }
 
+    class TrainAutoBody : public ParallelLoopBody
+    {
+    public:
+        TrainAutoBody(const vector<SvmParams>& _parameters,
+                      const cv::Mat& _samples,
+                      const cv::Mat& _responses,
+                      const cv::Mat& _labels,
+                      const vector<int>& _sidx,
+                      bool _is_classification,
+                      int _k_fold,
+                      std::vector<double>& _result) :
+        parameters(_parameters), samples(_samples), responses(_responses), labels(_labels),
+        sidx(_sidx), is_classification(_is_classification), k_fold(_k_fold), result(_result)
+        {}
+
+        void operator()( const cv::Range& range ) const
+        {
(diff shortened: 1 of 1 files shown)


modules/ml/src/svm.cpp
```diff
@@ -1713,12 +1808,11 @@ class SVMImpl : public SVM
 
         int sample_count = samples.rows;
         var_count = samples.cols;
-        size_t sample_size = var_count*samples.elemSize();
 
         vector<int> sidx;
         setRangeVector(sidx, sample_count);
 
-        int i, j, k;
+        int i, k;
```
sovrasov: The same thing about loop variables.

modules/ml/src/svm.cpp
```diff
@@ -1636,6 +1636,101 @@ class SVMImpl : public SVM
...
+            svm->class_labels = labels;
+
+            int rtype = responses.type();
+
+            Mat temp_train_samples(train_sample_count, var_count_, CV_32F);
+            Mat temp_test_samples(test_sample_count, var_count_, CV_32F);
+            Mat temp_train_responses(train_sample_count, 1, rtype);
+            Mat temp_test_responses;
+
+            int i,j,k,p;
```
r2d3: Just copy/pasted the original code. I was thinking to follow the coding rules doing like this

r2d3: If only my 3 loops are using C++ style loop declaration and ALL the rest of SVM code is using C-style, this will look weird

sovrasov: SVM code already contains the both styles, take a look at `static bool solve_*` functions.

r2d3: Corrected.
If it is OK for you, I will merge the 2 commits

sovrasov: It's ok.

sovrasov:
:+1:

PR #9324, opened 2017-08-07
Nested parallel region fix;

lupustr3:
### This pullrequest changes
Fix for: single thread call doesn't reset flagNestedParallelFor counter after completion.

In following usage case:
cv::setNumThreads(0);
func_mt - due to error this counts as a parallel region
cv::setNumThreads(>0);
func_mt - function treats this as a nested region and forces single thread.


alalek:
@lupustr3 Thanks for report! Could you please take a look on this fix: #9342 ?

PR #9326, opened 2017-08-07
videoio: don't pass zero fps value to 3rd-party codecs

### LINKED ISSUE #9321: Argument checking for `VideoWriter.open()` (Do not make it fail silently)

- OpenCV => **3.2.0**
- Operating System / Platform => **mixed** (WinX 64 Bit, Ubuntu 16.04)
- Compiler => **mixed**

##### Detailed description
We followed the [Video Writer Basic Example](http://docs.opencv.org/master/d5/d57/videowriter_basic_8cpp-example.html#a6) on capturing videos from camera input. It took us a bunch of hard work to get the example working. For some scenarios the following code worked. For others, it could _never_ opened the writer:

~~~cpp
    // use default camera as video source
    VideoCapture cap(0);
    // ...
    writer.open(filename, cap.get(CAP_PROP_FOURCC), cap.get(CAP_PROP_FPS), src.size(), isColor);
    // check if we succeeded
    if (!writer.isOpened()) {
        cerr << "Could not open the output video file for write\n";
        return -1;
    }
~~~

We compared OS (Win vs. Linux), installed codecs, C++ and Python API (_this_ made a difference) until we figured out a simple root cause: Depending on some factors, `cap.get(CAP_PROP_FPS)` returned an acutal frame rate or just `0`.

##### What to do?
In this case, a basic argument checking in `VideoWriter.open()` would helped me _a lot_. Probably I am not the only one struggling over this point?

Is it possible to add basic argument checking for [`VideoWriter::open()`](https://github.com/opencv/opencv/blob/b500ae2d21e858bebd68b34b6cfef352979c269a/modules/videoio/src/cap.cpp#L787) and `VideoWriter::VideoWriter` to cover these "easy" issues?

##### Steps to reproduce

_N/A_

sovrasov:
resolves #9321

### DIFF
--- a/modules/videoio/src/cap.cpp
+++ b/modules/videoio/src/cap.cpp
@@ -382,6 +382,8 @@ static CvVideoWriter* cvCreateVideoWriterWithPreference(const char* filename, in
     if(!fourcc || !fps)
         TRY_OPEN(result, cvCreateVideoWriter_Images(filename))
 
+    CV_Assert(result || fps != 0);
+
     switch(apiPreference)
     {
         default:


alalek:
:+1:

PR #9327, opened 2017-08-07
core: fix FS doesn't close input file after parse error

### LINKED ISSUE #9312: Persistence: Reading a broken file does not close the file handle
- OpenCV => 3.2 + X, SHA1 dcd8589b678824beeba4f5741c975ec78f943c8a

##### Detailed description
cv::FilesStorage does not correctly close the file handle if reading threw an error. This leads to other processes / calls not being able to modify the file until the process is closed.

The reason might be that if the [parsing](https://github.com/opencv/opencv/blob/87c27a074d/modules/core/src/persistence.cpp#L4521) fails, [cvReleaseFileStorage](https://github.com/opencv/opencv/blob/87c27a074d/modules/core/src/persistence.cpp#L4528) is called which then calls icvClose. But since [fs->is_opened is false](https://github.com/opencv/opencv/blob/87c27a074d/modules/core/src/persistence.cpp#L669), this does not seem to close the file. This is because the flag is only set [after successful parsing](https://github.com/opencv/opencv/blob/87c27a074d/modules/core/src/persistence.cpp#L4537)

##### Steps to reproduce

    #include "stdafx.h"
    #include <opencv2/core.hpp>
    #include <boost/filesystem.hpp>
    #include <iostream>

    int main(int argc, char* argv[])
    {
        const std::string filePath = "test.yml"; // This is a YML with wrong indentation
        try {
            cv::FileStorage fs(filePath, cv::FileStorage::READ);
        }
        catch (const std::exception& e) {
            std::cerr << e.what() << std::endl;
            // This will fail due to "cannot access the file because it is being used by another process"
            boost::filesystem::rename(boost::filesystem::path(filePath), boost::filesystem::path(filePath + "_broken.yml"));
            return 1;
        }
        return 0;
    }

sovrasov:
resolves #9312

### DIFF
--- a/modules/core/test/test_io.cpp
+++ b/modules/core/test/test_io.cpp
@@ -1352,7 +1352,7 @@ TEST(Core_InputOutput, FileStorage_free_file_after_exception)
     const std::string content = "%YAML:1.0\n cameraMatrix;:: !<tag:yaml.org,2002:opencv-matrix>\n";
 
     fstream testFile;
-    testFile.open(fileName, std::fstream::out);
+    testFile.open(fileName.c_str(), std::fstream::out);
     if(!testFile.is_open()) FAIL();
     testFile << content;
     testFile.close();
@@ -1360,9 +1360,10 @@ TEST(Core_InputOutput, FileStorage_free_file_after_exception)
     try
     {
         FileStorage fs(fileName, FileStorage::READ + FileStorage::FORMAT_YAML);
+        FAIL();
     }
     catch (const std::exception&)
     {
-        ASSERT_EQ(std::remove(fileName.c_str()), 0);
     }
+    ASSERT_EQ(std::remove(fileName.c_str()), 0);
 }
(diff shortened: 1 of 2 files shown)


modules/core/test/test_io.cpp
```diff
@@ -1345,3 +1345,24 @@ TEST(Core_InputOutput, FileStorage_json_bool)
...
+    fstream testFile;
+    testFile.open(fileName.c_str(), std::fstream::out);
+    if(!testFile.is_open()) FAIL();
+    testFile << content;
+    testFile.close();
+
+    try
+    {
+        FileStorage fs(fileName, FileStorage::READ + FileStorage::FORMAT_YAML);
+    }
```
PhilLab: I would add an assertion failure after the constructor, e.g. assert that ``!fs.isOpened()`` to be sure that future changes (i.e. not throwing an exception) don't render this test useless.

Similarly, I would move the ``std::remove`` to after the catch block and leave the block empty

vpisarev:
cool! thanks! :+1:

PR #9330, opened 2017-08-08
[GSOC] Enable OCL for AKAZE

hrnr:
This re-enables OCL for AKAZE.

* ocl path is only used when image is passed as UMat
* CPU path does not use any UMats
* added new ocl test that checks results are the same CPU version

cc: @bmagyar @alalek

### DIFF
--- a/modules/features2d/src/kaze/AKAZEFeatures.cpp
+++ b/modules/features2d/src/kaze/AKAZEFeatures.cpp
@@ -15,10 +15,6 @@
 
 #include <iostream>
 
-#ifdef HAVE_OPENCL // OpenCL is not well supported
-#undef HAVE_OPENCL
-#endif
-
 // Namespaces
 namespace cv
 {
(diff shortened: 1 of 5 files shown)


hrnr:
Hi @alalek , could you please provide some advice how I should integrate OCL in AKAZE?

alalek:
Please follow these statements:
- don't enable OpenCL code path silently for `cv::Mat`. Enable OpenCL code path if user passes cv::UMat images explicitly.
- using cv::UMat internally for other OpenCV functions forces OpenCL code paths for them. Don't do this unconditionally.
- `cv::UMat::getMat()` calls should be used for local processing only (not more than function code scope). Minimize lifetime of these derived objects as mush as possible. If you want long-life data download - use `copyTo()`.
- unfortunately, current cv::UMat implementation/design is not very "universal". Data is forced be stored in OpenCL buffers (there is no "on-demand" behavior).
- use passed InputArray / OutputArray as mush as possible - they are "universal pointers". Don't call getMat()/getUMat() without a strong reason (to just pass these buffers to next functions).
- add separate tests for OpenCL (T-API) code paths (use `test/opencl` / `perf/opencl` sub-directories).

bmagyar:
Thanks for the heads up @alalek !

hrnr:
thanks. Then it's probably better to split OCL/CPU paths completely.                                                                                                                                                                                  ‎

hrnr:
@alalek What do you think about current approach?

alalek:
@hrnr Looks better. Could you add "OCL" (UMat) test for AKAZE?

modules/features2d/src/kaze/AKAZEFeatures.cpp
```diff
@@ -455,42 +481,84 @@ void AKAZEFeatures::Create_Nonlinear_Scale_Space(InputArray img)
...
+  }
+}
+
+/**
+ * @brief This method creates the nonlinear scale space for a given image
+ * @param image Input image for which the nonlinear scale space needs to be created
+ */
+void AKAZEFeatures::Create_Nonlinear_Scale_Space(InputArray image)
+{
+  if (image.isUMat()) {
```
alalek: `&& cv::ocl::useOpenCL()`

hrnr:
Thanks for review.

I've added test checking results are the same between CPU and GPU. For me this is good to go in, I don't plan to add anything more.

hrnr:
there seems to be failure on buildbot. otherwise the build and tests are OK.

alalek:
Don't worry - there are sporadic hang failures of Mac builders. This build is restarted.

alalek:
Looks good to me. Thank you! :+1:

bjquinn:
If I understand correctly, it looks like this merge missed the deadline for 3.3.0.  If I want to use AKAZE with OpenCL, I would either need to build OpenCV from source or wait until the next release? (probably ~6 months if history is a guide)

Thanks!!

bjquinn:
For what it's worth, I cloned opencv-python, updated opencv to master, and used Appveyor to build it for Windows.  GPU acceleration ended up being slower for our purposes than CPU, so I don't think we'll use it at the moment, but for what it's worth, anyone interested can see the below.

https://github.com/bjquinn/opencv-python

https://ci.appveyor.com/api/buildjobs/wrdwimqmaidxu4qi/artifacts/dist%2Fopencv_python-3.3.0%2Bccba2c6-cp36-cp36m-win_amd64.whl

hrnr:
Thanks for sharing the results. The OCL path is expected to win over CPU only on large images, so your results makes sense. After the changes in CPU path the threshold is quite high (but that of cause depends also on your CPU vs GPU performace).

bjquinn:
@hrnr our images are relatively large -- letter size pages at 200 DPI, but of course being documents, they're mostly empty space.  At any rate, it seems that for our purpose, adding CPU cores will be easier/cheaper to scale than GPUs given the performance characteristics or our data.  But the improved CPU performance of AKAZE is much appreciated.

PR #9331, opened 2017-08-08
build: suppress warnings on Visual Studio

tomoaki0705:
### This pullrequest changes
 - suppress warnings on Visual Studio
 - Many warnings of ```C4512``` happens on my local machine
 - The build farm disables this ```C4512```, but I think it's better to write a workaround rather than disabling it.
 - BTW, I can't figure out at where the build farm is disabling the ```C4512```. Anyone have an idea ?


alalek:
Fixing source code make sense in case of external impact, like public headers.
Could you just disable this warning [here](https://github.com/opencv/opencv/blob/3.3.0/cmake/OpenCVCompilerOptions.cmake#L324-L329) ?
I saw this warning with MSVS 2012. We use MSVS 2015 on build farm, so probably some non-critical alarms are removed in newer MSVS.

tomoaki0705:
Thank you for your review.
I followed your comment.  Looks better.

alalek:
:+1:

PR #9335, opened 2017-08-08
Maven build improvements

jtkb:
### This pullrequest changes

Documented how to change the version of Maven to match OpenCV core and modified bash scripts accordingly.

Changed from WARN to ERROR the environment variable checks - I found that without these set the build would fail on Linux Mint and less experienced individuals may be mystified by the CMake fail message. This change hopefully makes the failure more obvious.

Extracted integration test module and placed into own profile.

Configured CMake Maven plugin to use native CMake by default.

Updated documentation.

### DIFF
--- a/platforms/maven/opencv/pom.xml
+++ b/platforms/maven/opencv/pom.xml
@@ -50,7 +50,6 @@
                             <workingDirectory>${project.basedir}/scripts</workingDirectory>
                             <arguments>
                                 <argument>properties</argument>
-                                <argument>../../../../modules/core/include/opencv2/core/version.hpp</argument>
                                 <argument>${build.directory}</argument>
                                 <argument>build.properties</argument>
                             </arguments>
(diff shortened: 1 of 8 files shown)


mshabunin:
:+1:

platforms/maven/opencv/pom.xml
```diff
@@ -175,14 +176,14 @@
                         <configuration>
                             <rules>
                                 <requireEnvironmentVariable>
-                                    <level>WARN</level>
+                                    <level>ERROR</level>
                                     <variableName>ANT_HOME</variableName>
                                     <message>$ANT_HOME is not set. Build may fail.</message>
```
alalek: There is no requirement for ANT_HOME in Ubuntu 16.04 with system Apache ant package.
The same note about JAVA_HOME (it is not required with system java jdk package)

jtkb: I agree with you but this assumes that the user is building using the system packages (I am assuming you mean the repo provided ones). I for one do not use the system JDK packages and install the Oracle JDK manually. I also do not set the JAVA_HOME environment variable set, mainly because Debian alternatives handle determination of which binaries to execute and OpenCv is one of a very few instances I need to have it set. Without JAVA_HOME set I see this build error:

`[INFO] 
[INFO] --- cmake-maven-plugin:3.4.1-b2:compile (cmake-compile) @ opencv ---
[INFO] Configured to use native CMake
make: *** No rule to make target 'opencv_java'. Stop.
`

Which to someone with a similar installation as mine i.e. using Oracle JDK may not link to JAVA_HOME not being set.

Consequently I decided to change these two checks to Error in an attempt to provide clear failure reason.

PR #9338, opened 2017-08-09
Fixed some OpenCL interface bugs

dkurt:
### This pullrequest changes

* Removed constructor without body
* Keep arguments `wscale` and `iwscale` by default (ones). Otherwise it gives floating point exception because of 0 by 0 division at https://github.com/opencv/opencv/blob/master/modules/core/src/ocl.cpp#L2204.


alalek:
Related ABI issues are added into ignore list.
:+1:

PR #9342, opened 2017-08-10
core: fix flag processing for nested regions in cv::parallel_for_()

alalek:
Similar to #9324
- more robust processing of nested flag (exception-safe)
- decreased number of atomic operations


lupustr3:
@alalek this should work.

alalek:
:+1:

PR #9347, opened 2017-08-11
Fix doxygen comment in BFMatcher

kschwan:
A missing asterisk caused doxygen to not generate documentation for the
BFMatcher::create() method. This PR has a minor patch which remedies that.


sovrasov:
:+1:

PR #9360, opened 2017-08-13
calib3d: suppress warning message

tomoaki0705:
### This pullrequest changes
 - suppress warning message
 - ```bigZ``` is **only** used [here ](https://github.com/opencv/opencv/blob/800f656402740948a498023202cc597f96cc140b/modules/calib3d/src/calibration.cpp#L2816)
```cpp
dptr[x][2] = bigZ;
```
 - where ```dptr``` is an array of ```float```
 - I guess ```const float``` should be good enough for ```bigZ```


alalek:
Thank you! :+1:

PR #9366, opened 2017-08-14
brisk add detection threshold for custom sampling pattern

martinsey:
### This pullrequest changes

It enables the opportunity to set a detection threshold while passing a sampling pattern to the BRISK feature detector. So far this detection threshold was fixed to 20. 

I found it quite useful to be able to vary this value while playing with it and i thought it might be useful for others, too. I also considered writing a test for this addition, but i did not find an appropriate place for this or rather, i did not find a test for the sampling pattern with brisk. If requested i can still add it!

Feedback of any kind is highly appreciated.

### DIFF
--- a/modules/features2d/include/opencv2/features2d.hpp
+++ b/modules/features2d/include/opencv2/features2d.hpp
@@ -250,6 +250,23 @@ public:
     @param indexChange index remapping of the bits. */
     CV_WRAP static Ptr<BRISK> create(const std::vector<float> &radiusList, const std::vector<int> &numberList,
         float dMax=5.85f, float dMin=8.2f, const std::vector<int>& indexChange=std::vector<int>());
+
+    /** @brief The BRISK constructor for a custom pattern, detection threshold and octaves
+
+    @param thresh AGAST detection threshold score.
+    @param octaves detection octaves. Use 0 to do single scale.
+    @param radiusList defines the radii (in pixels) where the samples around a keypoint are taken (for
+    keypoint scale 1).
+    @param numberList defines the number of sampling points on the sampling circle. Must be the same
+    size as radiusList..
+    @param dMax threshold for the short pairings used for descriptor formation (in pixels for keypoint
+    scale 1).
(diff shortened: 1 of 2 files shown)


modules/features2d/include/opencv2/features2d.hpp
```diff
@@ -247,9 +247,10 @@ class CV_EXPORTS_W BRISK : public Feature2D
     scale 1).
     @param dMin threshold for the long pairings used for orientation determination (in pixels for
     keypoint scale 1).
-    @param indexChange index remapping of the bits. */
+    @param indexChange index remapping of the bits.
+    @param thresh AGAST detection threshold score. */
     CV_WRAP static Ptr<BRISK> create(const std::vector<float> &radiusList, const std::vector<int> &numberList,
-        float dMax=5.85f, float dMin=8.2f, const std::vector<int>& indexChange=std::vector<int>());
+        float dMax=5.85f, float dMin=8.2f, const std::vector<int>& indexChange=std::vector<int>(), int thresh=20);
```
sovrasov: Unfortunately you can't change signatures of the public API methods due to binary compatibility restrictions. You can make an overloaded version of `create` without default values for the last two arguments and with reversed order of these ones (`thresh, indexChange`).

modules/features2d/src/brisk.cpp
```diff
@@ -88,6 +88,8 @@ class BRISK_Impl : public BRISK
                      OutputArray descriptors,
                      bool useProvidedKeypoints );
 
+    void setThreshold(int threshold_);
```
martinsey: Removed it.

modules/features2d/include/opencv2/features2d.hpp
martinsey: I found it to be straight forward to create a combination of the two constructors in this case, so that you can additionally set the number of octaces. I then kept the order, which i thought is the most intuitive one. Is that fine for you?

sovrasov:
:+1:

PR #9369, opened 2017-08-14
highgui: refactor GTK backend

alalek:
- eliminate legacy C code, unused fields
- fix mutex synchronization issues
- set trackbar min/max without unnecessary truncation; allow negative values


alalek:
:+1:

PR #9374, opened 2017-08-15
Lab tests: bit-exactness and other fixes

savuor:
### This pullrequest changes

* Lab2RGB_b:
   - bit-exactness enabled for all modes
   - Non-vectorized code calculates the same as vectorized code
* sRGB support added to tests
* Bit-exact test written for Lab
* ColorLab_full test fixed, error threshold raised
* Luv tests: rewritten to actual algorithm, comments added

### DIFF
--- a/modules/imgproc/src/color.cpp
+++ b/modules/imgproc/src/color.cpp
@@ -6771,7 +6771,10 @@ struct Lab2RGBinteger
 
         //float fxz[] = { ai / 500.0f + fy, fy - bi / 200.0f };
         int adiv, bdiv;
-        adiv = aa*BASE/500 - 128*BASE/500, bdiv = bb*BASE/200 - 128*BASE/200;
+        //adiv = aa*BASE/500 - 128*BASE/500, bdiv = bb*BASE/200 - 128*BASE/200;
+        //approximations with reasonable precision
+        adiv = ((5*aa*53687 + (1 << 7)) >> 13) - 128*BASE/500;
+        bdiv = ((  bb*41943 + (1 << 4)) >>  9) - 128*BASE/200+1;
 
         int ifxz[] = {ify + adiv, ify - bdiv};
 
@@ -7104,8 +7107,6 @@ struct Lab2RGB_b
                const float* _whitept, bool _srgb )
     : fcvt(3, _blueIdx, _coeffs, _whitept, _srgb ), icvt(_dstcn, _blueIdx, _coeffs, _whitept, _srgb), dstcn(_dstcn)
     {
-        useBitExactness = (!_coeffs && !_whitept && _srgb && enableBitExactness);
-
         #if CV_NEON
         v_scale_inv = vdupq_n_f32(100.f/255.f);
         v_scale = vdupq_n_f32(255.f);
@@ -7162,7 +7163,7 @@ struct Lab2RGB_b
 
     void operator()(const uchar* src, uchar* dst, int n) const
     {
-        if(useBitExactness)
+        if(enableBitExactness)
         {
             icvt(src, dst, n);
             return;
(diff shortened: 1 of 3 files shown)


alalek:
+1 for comparing with reference data (golden values) or their hash for algorithms declared bit-exact:
- make sure that bit-exactness really work between different platforms, configurations, compiler versions, overflow policies
- get rid of reference data calculation from test running time
- we can still use "reference" algorithm in test for detailed error reporting or for reference data/hash generation.

savuor:
@mshabunin @alalek Done, take a look.
Now checksums are compared with gold ones, reference algorithms are used for error reporting.

modules/imgproc/src/color.cpp
```diff
@@ -6767,7 +6767,10 @@ struct Lab2RGBinteger
 
         //float fxz[] = { ai / 500.0f + fy, fy - bi / 200.0f };
         int adiv, bdiv;
-        adiv = aa*BASE/500 - 128*BASE/500, bdiv = bb*BASE/200 - 128*BASE/200;
+        //adiv = aa*BASE/500 - 128*BASE/500, bdiv = bb*BASE/200 - 128*BASE/200;
+        //approximations with reasonable precision
+        adiv = ((5*aa*53687 + (1 << 7)) >> 13) - 128*BASE/500;
+        bdiv = ((  bb*41943 + (1 << 4)) >>  9) - 128*BASE/200+1;
```
alalek: Looks like these code lines are not equal to their comments.

`adiv = aa*BASE/500 - 128*BASE/500`  ==> `adiv = (aa-128)*BASE/500` ?

modules/imgproc/test/test_color.cpp
```diff
@@ -1935,20 +2090,342 @@ TEST(Imgproc_ColorLab_Full, accuracy)
...
+            Mat probe(256, 256, CV_8UC3), result;
+            rng.fill(probe, RNG::UNIFORM, 0, 255, true);
+
+            cvtColor(probe, result, codes[c]);
+
+            uint32_t h = adler32(result);
+
+            if(h != hashes[c*nIterations + iter])
+            {
+                initLabTabs();
```
alalek: We should dump information about mode, iteration, actual hash - at least to make updates of hash table above.

modules/imgproc/test/test_color.cpp
```diff
@@ -1935,20 +2090,342 @@ TEST(Imgproc_ColorLab_Full, accuracy)
...
+                            ts->printf(cvtest::TS::SUMMARY, "Error in: (%d, %d)\n", x,  y);
+                            ts->printf(cvtest::TS::SUMMARY, "Conversion code: %s\n", names[c].c_str());
+                            ts->printf(cvtest::TS::SUMMARY, "Reference value: %d %d %d\n", gx[0], gx[1], gx[2]);
+                            ts->printf(cvtest::TS::SUMMARY, "Actual value: %d %d %d\n", rx[0], rx[1], rx[2]);
+                            ts->printf(cvtest::TS::SUMMARY, "Src value: %d %d %d\n", px[0], px[1], px[2]);
+                            ts->printf(cvtest::TS::SUMMARY, "Size: (%d, %d)\n", probe.rows, probe.cols);
+
+                            ts->set_failed_test_info(cvtest::TS::FAIL_BAD_ACCURACY);
+                            ts->set_gtest_status();
+                            break;
```
alalek: Using of "ts" is deprecated. You should use GTest `FAIL() << message` form.

modules/imgproc/src/color.cpp
savuor: The goal was to make non-vectorized code produce the same values as vectorized one.
This code is an approximation of the calculation which is done w/o integer division operation since there's no integer division available in SIMD.

savuor:
@alalek Done, take a look

vpisarev:
let's merge it in :+1:

savuor:
Created [PR #375](https://github.com/opencv/opencv_extra/pull/375) in extra repo containing updated test data

PR #9379, opened 2017-08-16
imgproc: fix winSize in createHanningWindow()

### LINKED ISSUE #9378: createHanningWindow returns -nan values if size is 1 in any dimension

- OpenCV => 3.1
- Operating System / Platform => Linux Mint 18.1 64 Bit
- Compiler => gcc

##### Detailed description

This is with the `Anaconda` OpenCV package. For example, the following code always creates a Hanning window with `-nan` values. Swapping dimensions does the same. Making both dims `> 1` returns numerical values, as expected. I also tried this on `Android` and `OpenCV 3.2`, with the same results. 

##### Steps to reproduce

```
 Mat hann;
 createHanningWindow(hann, Size(100, 1), CV_32F);
 printf("First element of %d hanning window is %f", numSamples, hann.at<float>(0, 0));
```

berak:
resolves #9378

add a note in docs, that winSize has to be > 1. (a winSize of 2 will still produce an all zero output)
add an assertion in createHanningWindow

### DIFF
--- a/modules/imgproc/include/opencv2/imgproc.hpp
+++ b/modules/imgproc/include/opencv2/imgproc.hpp
@@ -2723,7 +2723,7 @@ An example is shown below:
     createHanningWindow(hann, Size(100, 100), CV_32F);
 @endcode
 @param dst Destination array to place Hann coefficients in
-@param winSize The window size specifications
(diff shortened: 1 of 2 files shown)


sovrasov:
:+1:

PR #9382, opened 2017-08-16
fix_9121

### LINKED ISSUE #9121: Tutorial "Making your own linear filters!" is misleading

- OpenCV => Latest

##### Detailed description
The tutorial [Making your own linear filters!](http://docs.opencv.org/3.2.0/d4/dbd/tutorial_filter_2d.html)
is misleading, which says the function `cv::filter2D` does convolution. But it actually
implements the correlation semantics, though correlation and convolution
give the same result when the kernel is symmetric

StevenPuttemans:
resolves #9121: as suggested, changes convolution to correlation


alalek:
:+1:

PR #9394, opened 2017-08-17
Fix Halide OpenCL buffer deallocation

dkurt:
### This pullrequest changes
Unset device allocation from Halide buffers that just reuse it. Newer Halide frees it twice.


alalek:
:+1:

PR #9406, opened 2017-08-18
GSoC - Updating Core Tutorials

Cartucho:
Using the java Highgui, toggle buttons and other tools to update the Core tutorials.
Each commit will contain a separate tutorial.

### DIFF
--- a/doc/tutorials/core/adding_images/adding_images.markdown
+++ b/doc/tutorials/core/adding_images/adding_images.markdown
@@ -1,13 +1,16 @@
 Adding (blending) two images using OpenCV {#tutorial_adding_images}
 =========================================
 
+@prev_tutorial{tutorial_mat_operations}
+@next_tutorial{tutorial_basic_linear_transform}
+
 Goal
 ----
 
 In this tutorial you will learn:
 
 -   what is *linear blending* and why it is useful;
--   how to add two images using @ref cv::addWeighted
+-   how to add two images using **addWeighted()**
 
 Theory
 ------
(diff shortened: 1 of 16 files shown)


Cartucho:
@sturkmen72 The reason why we were changing to the general one is to have a more general purpose markdown that can adapt to all languages.

PR #9260 will be adding the Python signatures to the docs. When people click addWeighted() the three signatures will show up, C++, Java and Python.

doc/tutorials/core/adding_images/adding_images.markdown
```diff
@@ -1,13 +1,16 @@
...
+@next_tutorial{tutorial_basic_linear_transform}
+
 Goal
 ----
 
 In this tutorial you will learn:
 
 -   what is *linear blending* and why it is useful;
--   how to add two images using @ref cv::addWeighted
+-   how to add two images using **addWeighted()**
```
csukuangfj: I would suggest to add another method
```cpp
// addWeighted( src1, alpha, src2, beta, 0.0, dst);
dst = alpha * src1 + beta * src2 + 0.0;
```
which is more intuitive and user friendly, as we do not need to remember which function to call.

samples/python/tutorial_code/core/mat_mask_operations/mat_mask_operations.py
```diff
@@ -70,7 +68,7 @@ def main(argv):
     dst0 = sharpen(src)
 
     t = (time.time() - t) / 1000
-    print "Hand written function time passed in seconds: %s" % t
+    print ("Hand written function time passed in seconds: %s" % t)
```
alalek: Please add this line to the beginning of file (before any imports):
```
```
This will allow to remove space after `print` (for Python 2.x)

Cartucho: does it also work for Python 3.x?

alalek: Yes

Cartucho:
@alalek I added the `print_function` import to all the tutorials

vpisarev:
cool! let's merge it in :+1:

PR #9228, opened 2017-07-25
Update OpenCVCompilerOptions.cmake

neok-m4700:
misplaced else in `cmake/OpenCVCompilerOptions.cmake` results in configure errors

PR #9408, opened 2017-08-18
(2.4) Update OpenCVCompilerOptions.cmake (backport 9228)

alalek:
#9228


zbeekman:
Thanks, I was trying to figure out the correct way to shallow clone this (since I use a case-sensitive HFS+ disk image for my development, that is not big enough with my other projects to hold the entire git history of OpenCV) and then cherry pick that commit from 2.4 to master, but you beat me to it after I commented.

CC: @ilovezfs

https://github.com/Homebrew/homebrew-science/issues/6151#issuecomment-323384645

ilovezfs:
:+1:

PR #9426, opened 2017-08-21
Extending CPU dispatch to the tests; fixing a typo

borisfom:
### This pullrequest changes
Extending CPU dispatch mechanism to the test sources.  Also fixing a typo in macro name.


borisfom:
This PR enables running tests that depend on CPU dispatched extensions. As POC, I have moved FP16 conversion tests into its own source that gets FP16 tests compiled when FP16 extensions are present.
Previously, the extension support code would be compiled for the library, but corresponding tests would be skipped (as activation flags would not be set for the test sources).

alalek:
We should add code with dispatching too (detect hardware, make check, and call or skip the actual code).
Currently default build of tests is crashed on non-AVX/FP16 capable hardware.
If you don't have mentioned hardware, you can check this via [Intel software development emulator](https://software.intel.com/en-us/articles/intel-software-development-emulator). Example:

```
# emulate Intel Merom CPU (Core 2)
$ .../sde-external-8.9.0-2017-08-06-lin/sde64 -mrm -env OPENCV_OPENCL_DEVICE disabled -env OPENCV_IPP disabled -- /home/alalek/projects/opencv/build/opencv/bin/opencv_test_core --gtest_filter=*hal_intrin*
WARNING: IPP was disabled by OPENCV_IPP environment variable
CTEST_FULL_OUTPUT
OpenCV version: 3.3.0-dev
OpenCV VCS version: 3.3.0-55-g76f7fb523
Build type: release
Parallel framework: pthreads
CPU features: mmx sse sse2 sse3
OpenCL is disabled
Note: Google Test filter = *hal_intrin*
[==========] Running 11 tests from 1 test case.
[----------] 11 tests from hal_intrin
[ RUN      ] hal_intrin.uint8x16
[       OK ] hal_intrin.uint8x16 (118 ms)
[ RUN      ] hal_intrin.int8x16
[       OK ] hal_intrin.int8x16 (88 ms)
[ RUN      ] hal_intrin.uint16x8
[       OK ] hal_intrin.uint16x8 (98 ms)
[ RUN      ] hal_intrin.int16x8
[       OK ] hal_intrin.int16x8 (81 ms)
[ RUN      ] hal_intrin.uint32x4
[       OK ] hal_intrin.uint32x4 (78 ms)
[ RUN      ] hal_intrin.int32x4
[       OK ] hal_intrin.int32x4 (80 ms)
[ RUN      ] hal_intrin.uint64x2
[       OK ] hal_intrin.uint64x2 (23 ms)
[ RUN      ] hal_intrin.int64x2
[       OK ] hal_intrin.int64x2 (25 ms)
[ RUN      ] hal_intrin.float32x4
[       OK ] hal_intrin.float32x4 (64 ms)
[ RUN      ] hal_intrin.float64x2
[       OK ] hal_intrin.float64x2 (45 ms)
[ RUN      ] hal_intrin.float16x4
TID 0 SDE-ERROR: Executed instruction not valid for specified chip (CORE2): 0x6c1937: vpxor xmm0, xmm0, xmm0
Image: /home/alalek/projects/opencv/build/opencv/bin/opencv_test_core+0x2c1937 (in multi-region image, region# 0)
Function: _ZN6cvtest3hal7TheTestIN2cv8hal_FP1611v_float16x4EE19test_loadstore_fp16Ev
Instruction bytes are: c5 f9 ef c0 
```

borisfom:
@alalek : thanks for the input! 
Right, I was too quick to remove the guards around the test in fp16 module, and it has extra condition on SIMD128. I have restored the guard, should be clean now.

alalek:
These guards are `true` during compilation of `.fp16` files. I will add a fix.

borisfom:
@alalek : that means it was broken before - thanks for fixing it!

alalek:
The fix was added into this PR.

In general I would like to avoid dispatched code in tests. There are a lot of shared functions in .o files with same names and different code generation flags ("inline" or "template" code from GTest and/or C++ `std` library) - these symbols duplicates are processed by linker with undefined algorithm (it grabs only one instantiation and use it for "both" code branches), so code crash may occur in some cases.

borisfom:
Thanks @alalek ! Going forward, how should I organize FP16 tests - is there a better way ?

alalek:
I believe we should use OpenCV builds with `-DCPU_BASELINE=NATIVE` CMake option in that case instead of dispatching. This flag enables all available/supported tests for current hardware/software(compiler) setup (without care for other configurations with old hardware/etc).

OpenCV is cross-platform library and supports many optimization paths / platforms / etc via single entry point from public API.
HAL intrinsic is internal stuff for OpenCV development. Users should not use it directly (or via C++ preprocessor guards).

vpisarev:
@alalek, let me assign this one to you

alalek:
:+1:

PR #9428, opened 2017-08-21
fix cv::CommandLineParser.

csukuangfj:
### This pullrequest changes
Fix commandline parser.

It should handle bool value not only of "true" but also of "TRUE" and "True".

### DIFF
--- a/modules/core/include/opencv2/core/utility.hpp
+++ b/modules/core/include/opencv2/core/utility.hpp
@@ -808,7 +808,7 @@ public:
 
     This method returns the path to the executable from the command line (`argv[0]`).
 
-    For example, if the application has been started with such command:
+    For example, if the application has been started with such a command:
     @code{.sh}
     $ ./bin/my-executable
     @endcode
@@ -914,7 +914,7 @@ public:
     */
     void printMessage() const;
 
-    /** @brief Print list of errors occured
+    /** @brief Print list of errors occurred
 
     @sa check
     */
--- a/modules/core/src/command_line_parser.cpp
+++ b/modules/core/src/command_line_parser.cpp
@@ -69,6 +69,15 @@ static const char* get_type_name(int type)
     return "unknown";
 }
 
+static bool parse_bool(std::string str)
+{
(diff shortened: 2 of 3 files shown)


terfendail:
AFAIK parser always know the type of the value, so IMHO it would be useful to handle 0 and 1 as well.

dkurt:
I've always wanted to use OpenCV's parser with key-value pairs separated by space. That means
`--key value` equals to `--key=value`. Now `.get<std::string>` produces `true` in the first case and `value` in the last one. We could write something like `--key ~/myfile` instead `--key=/home/$(whoami)/myfile` and autocomplete paths in shell.
It would be very cool if you add mentioned functionality here.

csukuangfj:
Hi, thank you for the comment.

Currently, the commandline parser supports only "--key=value" or "-key=value"

If you provide "--key value", it will set "key" to "true" and consider "value" as
a positional argument.

It would be impossible to distinguish key value pairs from positional argument without the "equal" sign.

catree:
Maybe something like [this](https://stackoverflow.com/a/3613424) to be able to parse generic boolean representation?

```
bool parse_bool(std::string str)
{        
    std::transform(str.begin(), str.end(), str.begin(), ::tolower);
    std::istringstream is(str);
    bool b;
    is >> (str.size() > 1 ? std::boolalpha : std::noboolalpha) >> b;
    return b;
}
```

[Demo code](http://cpp.sh/6wr7j).

csukuangfj:
@catree Thank you. It has been updated.

terfendail:
The change looks fine for me. Since space separated parameter handling IMO will require deeper rework of `CommandLineParser` I prefer to handle it with separate PR and merge existing change as is.
@csukuangfj @dkurt What's your opinion?

alalek:
It would be nice to add new test cases for `CommandLineParser` [here](https://github.com/opencv/opencv/blob/3.3.0/modules/core/test/test_utils.cpp) .

csukuangfj:
@alalek  Done.

alalek:
@csukuangfj Thank you! Looks good to me.

terfendail:
👍

PR #9430, opened 2017-08-21
cmake: fix gcc version detection

sovrasov:


alalek:
@sovrasov Could you take a look on #9431 ?

PR #9431, opened 2017-08-21
cmake: use CMAKE_CXX_COMPILER_VERSION (CMake 2.8.8+)

### LINKED ISSUE #7565: "-fvisibility=hidden" should not be forced when building a library like OpenCV
- OpenCV => 3.1 & current git master
- Operating System / Platform => Mac OS X
- Compiler => Xcode / clang

##### Detailed description

In cmake/OpenCVCompilerOptions.cmake there is the following block:

```
# set default visibility to hidden
if(CMAKE_COMPILER_IS_GNUCXX AND CMAKE_OPENCV_GCC_VERSION_NUM GREATER 399)
  add_extra_compiler_option(-fvisibility=hidden)
  add_extra_compiler_option(-fvisibility-inlines-hidden)
endif()
```

Forcing an -fvisibility choice like this seems to me incorrect when building a _library_ like OpenCV.  The problem is that the -fvisibility choice must be the same for all code built, that is, my app, library1, library2, etc. must all be built with the same -fvisibility flag.

Yes, in general, -fvisibility=hidden is better than -fvisibility=default, but sometimes one needs to use -fvisibility=default, as my app currently does.  Since OpenCV forces -fvisibility=hidden, I get link warnings like this:

```
ld: warning: direct access in cv::ocl::selectOpenCLDevice() to global weak symbol std::__1::char_traits<char>::eq(char, char) means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```

If I add -fvisibility=default to my CMAKE_CXX_FLAGS, it ends up earlier on the command list and thus clobbered:

```
--     C++ flags (Release):         -fvisibility=default <...> -fvisibility=hidden -fvisibility-inlines-hidden
```

I propose simply removing the hardcoded -fvisibility choice, and leave it to the user to specify any needed build options, but maybe there are other solutions...

alalek:
- '-fvisibility' flag was updated
- PVAPI library detection was changed

resolves #7565 (allow to specify custom values for visibility option)


alalek:
:+1:

PR #9441, opened 2017-08-23
ocl: Add a function to delete a cached program.

wzw-intel:
This is the counterpart of "Context::getProg". With this function,
user have chance to delete a program he don't need anymore from
global cached programs, and save resource.


alalek:
Thank you for contribution!

Probably proper name for this functionality is "unload".
We have plans to add external OpenCL programs cache (on persistent storage, like HDD), so probably "delete" name should go for these calls (to prevent users confusing).

Also we can preserve "key" (and "build" flags) in `Program::Impl`.

alalek:
:+1:

PR #9401, opened 2017-08-18
[GSOC] Speeding-up AKAZE, part #4

hrnr:
WIP

Speed-up non linear diffusion. Doing all steps in one pass and merging addition and diffusion.

cc: @bmagyar

PR #9249, opened 2017-07-27
[GSOC] Speeding-up AKAZE, part #3

hrnr:
cc: @bmagyar 

This is currently based on #8951. I will rebase on master, once it gets in.

Currently this contains reworked version of finding scale space extremas. Speed up for images with lots of keypoints is ~3.6x.

```

Geometric mean

                                                   Name of Test                                                        perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf      perf      perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf   
                                                                                                                    8200996b1  b12facf48  aa5a72b46  8cc0b286c  1d3f7fe9e  c13351891  76151e566  ea089a8ab  f9c2951fa  ba071d1ad  09c7288de  d71718dea  61a35d7a6  1d42c1fa8  70b66d1b6 1a4c8989d b12facf48  aa5a72b46  8cc0b286c  1d3f7fe9e  c13351891  76151e566  ea089a8ab  f9c2951fa  ba071d1ad  09c7288de  d71718dea  61a35d7a6  1d42c1fa8  70b66d1b6  1a4c8989d 
                                                                                                                                                                                                                                                                                                      vs         vs         vs         vs         vs         vs         vs         vs         vs         vs         vs         vs         vs         vs         vs    
                                                                                                                                                                                                                                                                                                     perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf       perf   
                                                                                                                                                                                                                                                                                                  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1  8200996b1 
                                                                                                                                                                                                                                                                                                  (x-factor) (x-factor) (x-factor) (x-factor) (x-factor) (x-factor) (x-factor) (x-factor) (x-factor) (x-factor) (x-factor) (x-factor) (x-factor) (x-factor) (x-factor)
[...]

PR #8951, opened 2017-06-21
[GSOC] Speeding-up AKAZE, part #2

hrnr:
This part focuses on performance improvements for AKAZE on CPU and implementing basic OCL support.

cc: @bmagyar

OpenCL:
=======

some parts (mainly construction of the pyramid) execute OpenCL paths.

* `Create_Nonlinear_Scale_Space` almost all operatiins are OpenCl enabled
    * GaussianBlur - executes OCL, but less effective path through sepFilter2D
    * Scharr - has ocl
    * resize - has ocl
    * kfactor estimation - needs custom ocl kernel, might reuse some parts for computing histogram
    * diffussion + conductivity - needs custom ocl kernel
    * `Compute_Determinant_Hessian_Response` executes OpenCL general sepFilter2D, computing determinant would need custom kernel, but is not currently blocker
* `Feature_Detection` no OCL
* `Compute_Keypoints_Orientation` no OCL
* `Compute_Descriptors` no OCL


CPU status:
===========

* `Create_Nonlinear_Scale_Space` reworked, some intrinsics might help with diffusion
    * `Compute_Determinant_Hessian_Response` reworked, needs specialized fine tuning for kernels 5x5, 7x7, 9x9
* `Feature_Detection` this will be reworked together with GPU part, we might want the same format of keypoints on GPU and CPU
* `Compute_Keypoints_Orientation` reworked
* `Compute_Descriptors` not a blocker

The main parts (and largest bottlenecks) are:
=============================================

* `Create_Nonlinear_Scale_Space` ~19% ~21ms
    * needs to be reworked
* `Feature_Detection` ~52% ~60ms, of which:
    * `Compute_Determinant_Hessian_Response` 83% (over 43% of the whole algorithm)
        * reworked slightly -> reduced Feature_Detection to ~58ms
    * `Find_Scale_Space_Extrema` 16% (globally just 8%)
        * this might be worth to look into when other parts will be optimized
    * `Do_Subpixel_Refinement` 1%
        * not a bottleneck in the current state
* `Compute_Keypoints_Orientation` ~21% ~24ms
     * reworked completely, replaced by a faster implementation, parallelized -> reduced to 3.69ms
* `Compute_Descriptors` ~4% ~4.8ms
    * not a bottleneck in the current state

Improvement in `Compute_Determinant_Hessian_Response` is of cause not very
satisfying. The main problem there is that AKAZE uses Scharr with non-standard
size kernel (different from 3x3). Implementing Scharr with sepFilter2D is much
[...]

PR #8869, opened 2017-06-08
[GSOC] Speeding-up AKAZE, part #1

hrnr:
Speeading-up AKAZE features project.

cc: @bmagyar

This part focuses on creating a reliable test suite for AKAZE to support perf refactoring of AKAZE.

tests:
* refactored perf tests in feature2d. now all tests uses the common implementation. Now I can reliably compare results for different detectors.
* added perf tests for AKAZE, BRISK, KAZE, MSER
* refactored ocl test in feature2d to be similar to cpu perf tests
* ~~fix bug in CV_ENUM (macro expansion issue)~~ This does not seem very wise after all, since we have lots of flags implemented as macros and we don't want to expand those. I have solved this the other way around (expand before passing to CV_ENUM)
* _reworked invariance tests_
* _added test for stitching pipeline with AKAZE feature finder_

AKAZE:
* make AKAZE and KAZE for with UMat descriptors (this was broken, probably nobody uses that)
* fix crash in AKAZE when using KAZE descriptors
* _do not compute keypoint orientation twice and allow detesting detector phase separately_
* _fix crash (out-of-bound read) near edges of image when computing MLDB descriptors_
* _fix AKAZE wrapper for stitching pipeline, fixes:_
```
! OPENCV warning: getUMat()/getMat() call chain possible problem.
!                 Base object is dead, while nested/derived object is still alive or processed.
!                 Please check lifetime of UMat/Mat objects!
```

performace we have right now (before this project improvements), based on the new refactord tests:

```

Geometric mean for
detectAndExtract::feature2d::(Y, X)

                                 X\Y                                  AKAZE_DEFAULT AKAZE_DESCRIPTOR_KAZE ORB_1500_13_1 ORB_DEFAULT
"cv/detectors_descriptors_evaluation/images_datasets/leuven/img1.png"   143.51 ms         167.04 ms          8.97 ms     17.68 ms  
"stitching/a3.png"                                                      113.28 ms         132.09 ms          7.23 ms     15.05 ms  
```
```

Geometric mean for
detect::feature2d::(Y, X)

[...]

PR #9444, opened 2017-08-23
[GSOC] Speeding-up AKAZE, tracking tutorial

hrnr:
This PR adds FPS stats to tracking demo with AKAZE and ORB. This allows to easily compare performace of AKAZE vs ORB, note that fps are measured for whole pipeline of features + matching + homography estimation.

This is the last PR in series implementing GSoC project "Speeding-up AKAZE features":

mentor: @bmagyar

# Previous PRs

#8869 extends tests for AKAZE, reworks perf tests and fixes several bugs
#8951 first version of OCL implementation, many CPU improvements
#9249 reworks feature detection with the new faster algorithm
#9330 rework OCL implementation after feedback from @alalek, new test for OCL

# Project proposal

AKAZE features are scale-invariant binary features. They are licence-free and
currently available in main OpenCV repository. AKAZE features show for many
applications better accuracy than widely used ORB features (also implemented
in OpenCV) [0]. However currently the implementation of AKAZE features in
OpenCV is 4-5 times slower than implementation of ORB features and the
implementation of AKAZE features lacks GPU acceleration. I’d like to address
both issues with current implementation of AKAZE features.

There already exists improvements of original Pablo F. Alcantarilla code [1],
which was the base of the current OpenCV code. Most notably there is a fast
parallel CPU version [2], and CUDA-based GPU implementation [3], [4]. Both
implementations promise a major speedup. I’d like to integrate this work to
OpenCV.

## CPU version

CPU version should probably build upon [2], but much integration work and
improvement is needed. CPU version should use common techniques in OpenCV, for
example existing parallel framework and not custom parallelism. I’d like to
extend the code to use vector intrinsics (probably through the new dispatched
intrinsics framework), IPP and OpenCV hal functions where applicable to get
the further performance benefit. Careful integration (backed by tests) must be
done to preserve improvements made to OpenCV code over the time.

* base on [2]
* get some more performance using vectorization, IPP, hal
* new tests

## GPU version

GPU version should be transparently available through the T-API. We should probably prefer OpenCl version to integrate seamlessly into existing infrastructure in OpenCV. Although we might use some ideas from CUDA version [3], [4], it might be necessary to write the GPU kernel from scratch.

* new OpenCl kernel

## Benefits to OpenCV

* faster AKAZE feature detector/descriptor
* GPU version of AKAZE

## References

[0] http://ieeexplore.ieee.org/document/7804434/
[1] https://github.com/pablofdezalc/akaze
[2] https://github.com/h2suzuki/fast_akaze
[3] https://github.com/nbergst/akaze
[4] https://arxiv.org/abs/1607.06178

# Implemented features (all from proposal implemented)

* speed-up 2.2x - 3.77x on CPU measured on Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz

chart

* extended tests coverage, added new tests for AKAZE (including stitching tests with AKAZE)
* reworked invariance tests in features2d
* reworked perf tests for features2d module
* fixed several bugs in AKAZE implementation
* improved CPU implementation on many places
* new algorithm for finding features - extensively tested to ensure it gives comparable (or even better) results that original implementation
* added OCL version for creation of scale space pyramid. Added tests to ensure OCL results are the same as CPU results.

Most of these improvements have been included in OpenCV 3.3 release (the first 3 PRs).

# Video

[short video presenting this project](https://www.youtube.com/watch?v=k8nKPc4044A)

# Other work

Some code that hasn't been merged. This includes performance improvements that
wasn't worth the trade-off and explored approaches that didn't brought the
performance benefit.

[test_scharr](https://github.com/hrnr/opencv/tree/test_scharr) I have tried to
[replace Scharr operator in `Compute_Determinant_Hessian_Response` with Scharr
[with fixed 3x3 kernel.

[akaze_octaves](https://github.com/hrnr/opencv/tree/akaze_octaves) Reworked
[non-linear scale space pyramid so that diffusivity is propagated only inside
[octaves. Probably not worth it, since it damages accuracy.

[test_l1](https://github.com/hrnr/opencv/tree/test_l1) Using L1 metric for keypoints pruning is
surprisingly slower that computing L2 metric.

#9401 Reworked non-linear diffusion in cache optimal manner, but it didn't brought a nice speed-up.
After all it is better to stick to current (already reworked in #8951) parallel approach.

# Blogs

[Speeding-up OpenCV Features](https://medium.com/@hrnr_/speeding-up-opencv-features-2ae1bc700b9a)

### DIFF
--- a/samples/cpp/tutorial_code/features2D/AKAZE_tracking/planar_tracking.cpp
+++ b/samples/cpp/tutorial_code/features2D/AKAZE_tracking/planar_tracking.cpp
@@ -62,8 +62,11 @@ void Tracker::setFirstFrame(const Mat frame, vector<Point2f> bb, string title, S
 
 Mat Tracker::process(const Mat frame, Stats& stats)
 {
+    TickMeter tm;
     vector<KeyPoint> kp;
     Mat desc;
+
+    tm.start();
     detector->detectAndCompute(frame, noArray(), kp, desc);
     stats.keypoints = (int)kp.size();
 
@@ -85,6 +88,8 @@ Mat Tracker::process(const Mat frame, Stats& stats)
         homography = findHomography(Points(matched1), Points(matched2),
                                     RANSAC, ransac_thresh, inlier_mask);
     }
+    tm.stop();
+    stats.fps = 1. / tm.getTimeSec();
 
     if(matched1.size() < 4 || homography.empty()) {
         Mat res;
--- a/samples/cpp/tutorial_code/features2D/AKAZE_tracking/stats.h
+++ b/samples/cpp/tutorial_code/features2D/AKAZE_tracking/stats.h
@@ -7,11 +7,13 @@ struct Stats
     int inliers;
     double ratio;
     int keypoints;
+    double fps;
 
     Stats() : matches(0),
         inliers(0),
         ratio(0),
-        keypoints(0)
+        keypoints(0),
+        fps(0.)
     {}
 
     Stats& operator+=(const Stats& op) {
@@ -19,6 +21,7 @@ struct Stats
         inliers += op.inliers;
         ratio += op.ratio;
         keypoints += op.keypoints;
+        fps += op.fps;
         return *this;
     }
     Stats& operator/=(int num)
@@ -27,6 +30,7 @@ struct Stats
         inliers /= num;
         ratio /= num;
         keypoints /= num;
+        fps /= num;
         return *this;
     }
 };
--- a/samples/cpp/tutorial_code/features2D/AKAZE_tracking/utils.h
+++ b/samples/cpp/tutorial_code/features2D/AKAZE_tracking/utils.h
@@ -25,15 +25,17 @@ void drawBoundingBox(Mat image, vector<Point2f> bb)
 void drawStatistics(Mat image, const Stats& stats)
 {
     static const int font = FONT_HERSHEY_PLAIN;
-    stringstream str1, str2, str3;
+    stringstream str1, str2, str3, str4;
 
     str1 << "Matches: " << stats.matches;
     str2 << "Inliers: " << stats.inliers;
     str3 << "Inlier ratio: " << setprecision(2) << stats.ratio;
+    str4 << "FPS: " << std::fixed << setprecision(2) << stats.fps;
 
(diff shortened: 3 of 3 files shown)


tompollok:
@hrnr as you worked a lot on AKAZE in the past weeks/months - can you say anything about the speed difference for your OpenCL enabled AKAZE version vs the current implementation of ORB?

I have seen videos about a comparison between AKAZE and ORB, which mainly focused on comparing the feature detection and matching performance. But it would be also nice to know sth about the speed.

hrnr:
Here is comparison of all non-contrib features on Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz:

```

Geometric mean for
detectAndExtract::feature2d::(Y, X)

                                 X\Y                                  AKAZE_DEFAULT AKAZE_DESCRIPTOR_KAZE BRISK_DEFAULT KAZE_DEFAULT ORB_1500_13_1 ORB_DEFAULT
"cv/detectors_descriptors_evaluation/images_datasets/leuven/img1.png"   33.39 ms          39.50 ms          68.83 ms     154.16 ms      5.20 ms     10.52 ms  
"stitching/a3.png"                                                      26.21 ms          31.11 ms          39.94 ms     127.28 ms      4.44 ms      8.69 ms  
"stitching/s2.jpg"                                                      74.79 ms          102.70 ms         508.10 ms    338.16 ms     11.30 ms     27.36 ms  
```

As a rule of thumb AKAZE is now 3 times slower than ORB (or even less on bigger images). This is improvement from being 7 times slower than ORB on smaller images and 10 times slower on biger images. This is for default configurations.

Note that you can make ORB much more faster if you can accept lower accuracy.

OpenCL is currently useful mostly for big images (2K, 4K) e.g. top accuracy stitching etc. Depends on your needs and your hardware. You can evaluate for yourself if OCL will bring benefit for your pipeline on your hardware with new perf tests `OCL_feature2d_detectAndExtract.detectAndExtract`.

tompollok:
@hrnr thank you very much for this information! Also thank you for your contributions to opencv :-)

hrnr:
This PR is ready for merge. It adds FPS info to tracking tutorial, which I have used for making the video.

vpisarev:
thanks, great job (I mean, the whole project went well and there are some very useful results!) :+1:

PR #9445, opened 2017-08-23
Adds cv::resizeWindow() overload taking a single cv::Size argument.

adishavit:
Adds `cv::resizeWindow()` overload taking a single `cv::Size` argument instead of 2 separate `width` and `height`.

### DIFF
--- a/modules/highgui/include/opencv2/highgui.hpp
+++ b/modules/highgui/include/opencv2/highgui.hpp
@@ -401,13 +401,7 @@ CV_EXPORTS_W void imshow(const String& winname, InputArray mat);
  */
 CV_EXPORTS_W void resizeWindow(const String& winname, int width, int height);
 
-/** @brief Resizes window to the specified size
-
-@note
-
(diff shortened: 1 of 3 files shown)


modules/highgui/include/opencv2/highgui.hpp
```diff
@@ -401,6 +401,18 @@ CV_EXPORTS_W void imshow(const String& winname, InputArray mat);
  */
 CV_EXPORTS_W void resizeWindow(const String& winname, int width, int height);
 
+/** @brief Resizes window to the specified size
```
sovrasov: It's better to use `/** @overload` command here instead of useless copy of description.

adishavit: Good point - it's been a while since I used doxygen. Fixing.
Also, originally, I wanted to make it an inline function calling the one above, but I wasn't sure how that plays with the `CV_EXPORTS_W` macro.

Issue #9402, opened 2017-08-18
moveWindow adds offset on second call


dustsigns:
##### System information (version)
- OpenCV => 3.3
- Operating System / Platform => Ubuntu 16.04 (x64-64) with QT 5
- Compiler => g++ 5.4.0

##### Detailed description

`cv::moveWindow` places the window to a different X and Y position after the second call. This seems to be loosely related to #6650, but is different since the issue at hand also happens in a single-threaded environment. Consider the following minimum working example:

##### Steps to reproduce

```
#include <iostream>
#include <opencv2/highgui.hpp>

using namespace std;
using namespace cv;

int main()
{
  imshow("Test window", Mat(100, 100, CV_8UC1, Scalar(0)));
  moveWindow("Test window", 300, 0);
  waitKey(0);
  moveWindow("Test window", 300, 0);
  waitKey(0);
  return 0;
}
```

After pressing a key, the test window is moved slightly to the right, even though there are no other threads involved and the exact same coordinates are used. Note: I have only tested this with QT enabled.
The same issue happens in the Y direction, albeit with a larger offset:

```
#include <iostream>
#include <opencv2/highgui.hpp>

using namespace std;
using namespace cv;

int main()
{
  imshow("Test window", Mat(100, 100, CV_8UC1, Scalar(0)));
  moveWindow("Test window", 300, 200);
  waitKey(0);
  moveWindow("Test window", 300, 200);
  waitKey(0);
  return 0;
}
```

dkurt:
@dustsigns, I can't reproduce it. OpenCV 3.3, g++ 5.4.0, Qt 5.5.1.
Can you check a version of linked OpenCV?
```cpp
std::cout << CV_VERSION << std::endl;
```
May be it depends on initial position of the window? Could you try to create more than one window and move them together?

dustsigns:
The output is

> 3.3.0

I compile version 3.3 from source with QT 5.5.1 installed. I compile it with the contrib modules, but disable quite a lot of them. I also disable a few of opencv's (regular) modules, but this should not make any difference. I can post the complete command line that I use for `cmake` (which is quite long), if desired, but I don't think that the disabled modules are to blame.

If I move two windows together, the problem persists. I can make a video (or series of images) of this effect, if it helps.

dkurt:
I've also built only `core imgproc imgcodecs videoio highgui python2` modules.
[...]

PR #9447, opened 2017-08-23
Added 1ms delay to let system decorate Qt windows

dkurt:
### This pullrequest changes
resolves https://github.com/opencv/opencv/issues/9402
After creation of Qt windows, force execution of all events to let OS decorate title bars.


alalek:
Validated
:+1:

PR #9449, opened 2017-08-23
ocl: Fix OpenCL library detection in Linux

ribalda:
OpenCL runtime does not require OpenCL development file (libOpenCL.so),
just the "run" library (so.1).

This patch searches for the run library (so.1) if the dev library (.so)
is not found.

Web search shows that this error has been present since at least 2015
http://answers.opencv.org/question/80532/haveopencl-return-false/

Signed-off-by: Ricardo Ribalda Delgado <ricardo.ribalda@gmail.com>

### DIFF
--- a/modules/core/src/opencl/runtime/opencl_core.cpp
+++ b/modules/core/src/opencl/runtime/opencl_core.cpp
@@ -134,6 +134,24 @@ static void* WinGetProcAddress(const char* name)
 #include <dlfcn.h>
 #include <stdio.h>
 
+static void *GetHandle(const char *file)
+{
+    void *handle;
+
+    handle = dlopen(file, RTLD_LAZY | RTLD_GLOBAL);
+    if (!handle)
+        return NULL;
+
+    if (dlsym(handle, OPENCL_FUNC_TO_CHECK_1_1) == NULL)
+    {
+        fprintf(stderr, ERROR_MSG_INVALID_VERSION);
+        dlclose(handle);
+        return NULL;
+    }
+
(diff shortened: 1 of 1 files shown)


modules/core/src/opencl/runtime/opencl_core.cpp
```diff
@@ -143,21 +161,19 @@ static void* GetProcAddress(const char* name)
...
-            const char* path = "libOpenCL.so";
             const char* envPath = getenv("OPENCV_OPENCL_RUNTIME");
             if (envPath)
-                path = envPath;
-            handle = dlopen(path, RTLD_LAZY | RTLD_GLOBAL);
-            if (handle == NULL)
             {
-                if (envPath)
+                handle = GetHandle(envPath);
+		if (!handle)
```
alalek: please fix bad indentation

ribalda: Hi, I was trying to follow the file's indentation policy. 

Do you prefer to switch to tabs instead of 4 spaces?

alalek: Currently there is indentation with 2 tabs (try to select that via GitHub Web UI).

ribalda: You are absolutely right. Sorry about that.

I have sent a fixed version.

alalek:
Thank you for contribution! Looks good to me :+1:

ribalda:
Thanks for your review

PR #9456, opened 2017-08-24
Correctly restore (registry saved) window position on multi-monitor systems. Fixes #8840

### LINKED ISSUE #8840: HighGUI windows on Windows do not remember position on multiple screens
On Windows, the position of HighGUI windows is saved in the registry so that on the next run they are opened in the same position (if they have the same name). This is a _very_ convenient feature.

However, it does not work correctly when the windows is on 2nd/external screen/display/monitor.

I guess this is somewhere between a bug report and a feature request.

adishavit:
Replaces arbitrary position and size snapping by snapping to nearest monitor.
Fixes issue #8840
Windows only.

### DIFF
--- a/modules/highgui/src/window_w32.cpp
+++ b/modules/highgui/src/window_w32.cpp
@@ -324,15 +324,34 @@ icvLoadWindowPos( const char* name, CvRect& rect )
         RegQueryValueEx(hkey, "Width", NULL, &dwType, (BYTE*)&rect.width, &dwSize);
         RegQueryValueEx(hkey, "Height", NULL, &dwType, (BYTE*)&rect.height, &dwSize);
 
-        if( rect.x != (int)CW_USEDEFAULT && (rect.x < -200 || rect.x > 3000) )
-            rect.x = 100;
-        if( rect.y != (int)CW_USEDEFAULT && (rect.y < -200 || rect.y > 3000) )
-            rect.y = 100;
+        // Snap rect into closest monitor in case it falls outside it. // Adi Shavit
+        // set WIN32 RECT to be the loaded size
+        POINT tl_w32 = { rect.x, rect.y };
+        POINT tr_w32 = { rect.x + rect.width, rect.y };
 
-        if( rect.width != (int)CW_USEDEFAULT && (rect.width < 0 || rect.width > 3000) )
-            rect.width = 100;
-        if( rect.height != (int)CW_USEDEFAULT && (rect.height < 0 || rect.height > 3000) )
-            rect.height = 100;
+        // find monitor containing top-left and top-right corners, or NULL
+        HMONITOR hMonitor_l = MonitorFromPoint(tl_w32, MONITOR_DEFAULTTONULL);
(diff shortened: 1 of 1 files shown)


sturkmen72:
according to my understanding and experience ( tested with two monitors ) no need to add new lines to code. i prefer only modifications with values on existing code

adishavit:
@sturkmen72 : Sorry, I don't understand what you mean.
I have 2 monitors:

```
                           +--------------------------+
                           |     +-----+              |
                           |     |     |              |
                           |     |     |              |
+--------------------------+     +-----+              |
|                         ||            2             |
|                         ||                          |
|                         ||                          |
|           1             ||                          |
|                         ||                          |
|                         +---------------------------+
|                         |
|                         |
+-------------------------+
```
The window on monitor 2 will be restored in the wrong place if the offset is too big.

adishavit:
The old code did allow a -200 negative offset only to the left and up.
It's possible to add that 200px slack for partially out of screen.

adishavit:
You can use this program to test the existing and new PR by dragging the window and hitting ESC in each new position:

```
#include <opencv2/opencv.hpp>
#include <opencv2/highgui/highgui.hpp>

//////////////////////////////////////////////////////////////////////////


int main(int argc, char* argv[])
{
   using namespace cv;

   
   Mat img(Size{ 400,400 }, CV_8UC3);

   imshow("img", img);
   while ('x' != waitKey())
   {
      cv::destroyAllWindows();
      imshow("img", img);
   }


   return EXIT_SUCCESS;
}
```

sturkmen72:
sorry @adishavit i respect your work and idea.
This is my humble thought as a common OpenCV user
 but i still think same. no need to extra code. did you try to change values ( -200, 3000 )

```
       if( rect.x != (int)CW_USEDEFAULT && (rect.x < -200 || rect.x > 3000) )
           rect.x = 100;
        if( rect.y != (int)CW_USEDEFAULT && (rect.y < -200 || rect.y > 3000) )
            rect.y = 100;

         if( rect.width != (int)CW_USEDEFAULT && (rect.width < 0 || rect.width > 3000) )
             rect.width = 100;
         if( rect.height != (int)CW_USEDEFAULT && (rect.height < 0 || rect.height > 3000) )
             rect.height = 100;
```

adishavit:
I still do not understand what you mean by "no extra code".
_All_ I did is change the _arbitrarily chosen, magic numbers_, `-200, 3000` and `100` into more reasonable values that depend on your actual monitor(s).

sturkmen72:
i get screencaptures
(restored window position before patch )
page01
(restored window position after patch)
page02

i think patch should not move the window

adishavit:
OK, now I understand. 
Technically, the previous version would have moved your window too if you moved it on the right screen below the bottom of the left window. 
But, I will try to make the conditions for moving stricter - basically, only if the top bar (which you can use to move) is completely outside the screen.

sturkmen72:
@adishavit thanks. ( sorry for lacking explanation in the beginning )

adishavit:
There's still a hidden bug here.

adishavit:
Fixed bug.

vpisarev:
thanks! let's merge it in and let people test it :) :+1:

PR #9457, opened 2017-08-24
core: fix type traits

### LINKED ISSUE #7599: Incorrect element size of cv::Mat when initialised with an instance of cv::Mat_<uint32_t>.

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1352

Removed incorrect default values from type traits.

resolves #7599


modules/core/include/opencv2/core/traits.hpp
```diff
@@ -49,11 +49,23 @@
 namespace cv
 {
 
+//#define OPENCV_TRAITS_ENABLE_DEPRECATED
+
+// Hack for CUDA modules (refactoring is required for CUDA modules)
+#ifndef OPENCV_TRAITS_ENABLE_DEPRECATED
+# include "opencv2/opencv_modules.hpp"
+# ifdef HAVE_OPENCV_CUDAFEATURES2D
+#   define OPENCV_TRAITS_ENABLE_DEPRECATED
```
mshabunin: Is it possible to turn this on via cmake?

alalek: via CMake we should turn on this in two places at least:
- OpenCV compilation itself
- user applications which use OpenCV

mshabunin: @alalek , sure. Maybe something like the following lines will work in both cases:
```.cmake
if (HAVE_CUDA)
    target_compile_definitions(opencv_core PUBLIC OPENCV_TRAITS_ENABLE_DEPRECATED)
endif()
```

alalek: Replaced

mshabunin:
:+1:

PR #9458, opened 2017-08-24
improve the documentation.

csukuangfj:

### DIFF
--- a/samples/cpp/tutorial_code/snippets/core_merge.cpp
+++ b/samples/cpp/tutorial_code/snippets/core_merge.cpp
@@ -1,0 +1,36 @@
+/**
+ * @file core_merge.cpp
+ * @brief It demonstrates the usage of cv::merge.
+ *
+ * It shows how to merge 3 single channel matrices into a 3-channel matrix.
+ *
+ * @author KUANG Fangjun
+ * @date August 2017
+ */
+
+#include <iostream>
+#include <opencv2/core.hpp>
+
+using namespace std;
+using namespace cv;
+
+int main()
+{
+    //! [example]
+    Mat m1 = (Mat_<uchar>(2,2) << 1,4,7,10);
+    Mat m2 = (Mat_<uchar>(2,2) << 2,5,8,11);
+    Mat m3 = (Mat_<uchar>(2,2) << 3,6,9,12);
+
+    Mat channels[3] = {m1, m2, m3};
+    Mat m;
+    merge(channels, 3, m);
+    /*
+    m =
+    [  1,   2,   3,   4,   5,   6;
+       7,   8,   9,  10,  11,  12]
+    m.channels() = 3
+    */
+    //! [example]
+
+    return 0;
+}
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -847,6 +847,13 @@ obtained. For example, the function can be used to compute horizontal and vertic
 raster image. In case of REDUCE_MAX and REDUCE_MIN , the output image should have the same type as the source one.
 In case of REDUCE_SUM and REDUCE_AVG , the output may have a larger element bit-depth to preserve accuracy.
 And multi-channel arrays are also supported in these two reduction modes.
+
+The following code demonstrates its usage for a single channel matrix.
+@snippet snippets/core_reduce.cpp example
+
+And the following code demonstrates its usage for a two-channel matrix.
+@snippet snippets/core_reduce.cpp example2
+
 @param src input 2D matrix.
 @param dst output vector. Its size and type is defined by dim and dtype parameters.
 @param dim dimension index along which the matrix is reduced. 0 means that the matrix is reduced to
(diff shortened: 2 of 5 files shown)


mshabunin:
@csukuangfj , it will be better to put all code snippets into separate cpp-file (in samples/cpp/tutorial_code folder) and include them with doxygen command `@snippet`: http://docs.opencv.org/3.3.0/d4/db1/tutorial_documentation.html#tutorial_documentation_commands_include
This way code snippets will be compiled along with OpenCV and will always be in sync with the codebase.

csukuangfj:
@mshabunin Thank you for the hint. It makes the documentation much clearer in this way.

I find that most of the OpenCV APIs lack of demo code. To ensure that I understand
the APIs correctly,  I have to write some small tests before using them. Therefore, I think
such small code snippets with output would be very helpful for users.

csukuangfj:
@alalek Why are there some warnings (http://pullrequest.opencv.org/buildbot/builders/precommit_docs/builds/11966/steps/Pylint/logs/warnings%20%2863%29) ?

alalek:
@csukuangfj Don't worry. It is not related to your patch. We will fix these problems soon.

sturkmen72:
just an idea, 
1.what about using `namespace cv` and `namespace std`
2.what about gathering all code in a single cpp ? ( maybe we will add all this kind of simple usage code snippets in it)

csukuangfj:
@sturkmen72  Thank you for the comment.

1. Yes, you are right. It could write `using namespace cv;` at the beginning instead
of prefixing every OpenCV API with `cv`. But it is not considered as a good practice.
For example, [Why is “using namespace std” considered bad practice?][1] at stackoverflow.com

2. One of the advantage of scattering the demos over multiple files is that it does only one thing,
is self contained and easier to understand. Moreover, it could be served as demo code for newcomers.

[1]: https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-considered-bad-practice

hrnr:
Just a note to Mat initialization: I think it is a more elegant and recommended to either use Matx constructors or use the solution with Mat_ and `operator<<` to initialize matrices rather than c-style arrays.

csukuangfj:
@hrnr Thank you for the suggestion.

I have updated to use `cv::Mat_<>` for single channel matrix initialization. It is cumbersome
to use it for multi-channel initialization.

mshabunin:
@csukuangfj , 
- I suggest creating a `samples/cpp/tutorial_code/snippets` folder and moving files there with the following naming scheme: `<module>_<function>.cpp` or `<module>_<function_group>.cpp` (for example `core_merge.cpp` or `core_mat_init.cpp`). Then snippets can be included into documentation using shorter and more obvious path: `@snippet snippets/core_merge.cpp example` (doxygen does not need the full path).
- I agree with @sturkmen72 , with namespaces code will become more clean.
- I think it is better to avoid very long snippets (like in cv::reduce). For example you can illustrate just one or two operations with rows and columns.
- It is better to make code more compact by splitting initialization, calculation and output, choosing different data layouts (less rows) and omitting some fancy formatting:
    ```.cpp
    //! [example]
    Mat m = (Mat_<uchar>(2,3) << 1,2,3, 4,5,6), col_sum, row_average;
    reduce(m, col_sum, 0, CV_REDUCE_SUM, CV_32F);
    reduce(m, row_average, 1, CV_REDUCE_AVG, CV_32F);
    // col_sum = [5, 7, 9]
    // row_average = [2;
    //  5]
    //! [example]

    cout << "col_sum = " << col_sum << endl;
    cout << "row_average = " << row_average << endl;
    ```
- You can add local scopes to allow reuse of short variable names instead of adding `m2`, `m3`, etc.

csukuangfj:
@mshabunin 

Thank you for the advices. I have changed the commit as you suggested.

A preview can be found at the following address:
http://pullrequest.opencv.org/buildbot/export/pr/9458/docs/d2/de8/group__core__array.html#ga4b78072a303f29d9031d56e5638da78e

samples/cpp/tutorial_code/snippets/core_merge.cpp
```diff
@@ -0,0 +1,40 @@
...
+    //! [example]
+    Mat m1 = (Mat_<uchar>(2,2) << 1,4,7,10);
+    Mat m2 = (Mat_<uchar>(2,2) << 2,5,8,11);
+    Mat m3 = (Mat_<uchar>(2,2) << 3,6,9,12);
+
+    Mat channels[3] = {m1, m2, m3};
+    Mat m;
+    merge(channels, 3, m);
+
+    cout << "m =" << std::endl << m << endl;
```
mshabunin: Could you please move all `cout << ...` lines out of documentation snippet block to reduce its size.
Same for other files.

csukuangfj: DONE.

vpisarev:
@mshabunin, can you please check the updated version?

mshabunin:
Looks good to me, thank you!
:+1:

PR #9469, opened 2017-08-25
Fix #9053 frame index calculation for in VideoCapture::set (CAP_PROP_…

### LINKED ISSUE #9053: VideoCapture::set (CAP_PROP_POS_FRAMES, frameNumber) not exact in opencv 3.2 with ffmpeg
Example and confirmed effect:

http://answers.opencv.org/question/162781/videocaptureset-cap_prop_pos_frames-framenumber-not-exact-in-opencv-32-with-ffmpeg/

chcao052:
…POS_FRAMES,) with ffmeg

av_seek_frame() is seeking with DTS time, hence the initial time should be first_dts instead of start_time.

Why:  with my limit understanding of ffmpeg, start_time is the +ve offset from first_dts. 
For most video files, start_time and first_dts are the same hence the code seem to work. 
When they are different, using start_time to seek will result in the next key frame(eg. 30 for me).


chcao052:
Found new video samples, the fix is not not good enough. Need rework.

PR #9470, opened 2017-08-25
Bit-exact version of Luv2RGB_b

savuor:
### This pullrequest changes

1. Luv2RGB_b is now calculated in integers which provides bit-exactness across architectures, compilers, etc. All lookup tables are calculated in float
2. OCL code is done in correspondence with CPU's, docs fixed about X, Y, Z range
3. Added bit-exactness test for Luv2RGB_b

### Performance

Measured 100 runs on random 257x257 image
Repeated 100 times and then averaged

Version | Linux x64 Core i5-6600 @ 3.3 GHz |  Win x64 Core i5-6300U @ 2.4 GHz
------------ | ------------- | -------------
Bit-exact | 0.091 s | 0.195 s
Original | 0.051 s | 0.083 s

### Accuracy

Range of values is 0 to 255

- max abs error: 2 **ONLY FOR VALUES WHICH LIE IN WHITE POINT RANGE (WHEN 0 <= X, Y, Z < 2)**
- mean by RGB channels: mean [0.02, 0.01, 0.008]
- stddev by RGB channels: [0.13, 0.12, 0.09]


savuor:
Perf data updated by [PR #376](https://github.com/opencv/opencv_extra/pull/376) in extra repo

modules/imgproc/src/color.cpp
```diff
@@ -6013,6 +6017,43 @@ static void initLabTabs()
...
+                softfloat up = softfloat(9)*(u + L*un);
+                LuToUp_b[LL*256+uu] = cvRound(up*softfloat(BASE/1024));//1024 is OK, 2048 gave maxerr 3
+            }
+            for(int vv = 0; vv < 256; vv++)
+            {
+                softfloat v = softfloat(vv)*vRange/f255 + vLow;
+                softfloat vp = oneof4/(v + L*vn);
+                if(vp >  oneof4) vp =  oneof4;
+                if(vp < -oneof4) vp = -oneof4;
+                int ivp = cvRound(vp*softfloat(BASE*1024));
```
vpisarev: is there cvRound() overload for softfloat?

modules/imgproc/src/color.cpp
```diff
@@ -8329,22 +8384,235 @@ struct RGB2Luv_b
...
+        useBitExactness = (!_whitept && enableBitExactness);
+
         #if CV_NEON
-        v_scale_inv = vdupq_n_f32(100.f/255.f);
-        v_coeff1 = vdupq_n_f32(1.388235294117647f);
-        v_coeff2 = vdupq_n_f32(1.027450980392157f);
-        v_134 = vdupq_n_f32(134.f);
-        v_140 = vdupq_n_f32(140.f);
+        static const softfloat f255(255);
+        v_scale_inv = vdupq_n_f32(softfloat(100)/f255);
```
vpisarev: probably we should replace NEON and SSE intrinsics with universal intrinsics

modules/imgproc/src/color.cpp
savuor: Sure, see [softfloat.cpp](https://github.com/opencv/opencv/blob/e70b75b51a40f2716b65d3b70e0ee56966267f4d/modules/core/include/opencv2/core/softfloat.hpp#L389)

modules/imgproc/src/color.cpp
savuor: Done, take a look

vpisarev:
looks great! :+1:

PR #9474, opened 2017-08-26
Basic Haiku support

miqlas:
### This pullrequest changes

This PR adds a basic Haiku support.


vpisarev:
thank you! looks like the patch cannot break anything and thus can be safely integrated :+1:

Issue #5003, opened 2015-07-27
Support for Android Studio


opencv-pushbot:
Transferred from http://code.opencv.org/issues/4366

```
|| Philipp Hasper on 2015-05-28 11:31
|| Priority: Normal
|| Affected: None
|| Category: android
|| Tracker: Feature
|| Difficulty: 
|| PR: 
|| Platform: None / None
```
## Support for Android Studio

```
Android Studio is now the official supported way for developing Android apps - the Eclipse integration is deprecated.

We need to migrate the Android tutorials / Android build chain to Android Studio.
```

PhilLab:
Is this planned for OpenCV 3.0.1?

hauptmech:
I have been having a very unpleasant experience getting opencv and android development to work. I'm new to android and the java toolchain. I'd be really happy with a commandline walkthrough of getting any opencv native sample running with gradle+ndk-build. An android studio example would be ok too but that seems to be a moving target right now.

PhilLab:
Yeah, the "these tutorials are for Eclipse but are easily reproducable with Android Studio"-sentence at the beginning of many new Android OpenCV tutorials seems rather cynical :wink: It does work but it is by no means easy if you are new to Gradle.

hauptmech:
12 hours to get tutorial-2-mixedprocessing to build and run. Having OpenCV 2.4 and an Arm64 device did not help things, but most of my time was figuring out a working combination of android studio, android_ndk, gradle-experimental vs gradle plugins, and viable imports of the samples (or gradle layout remappings).

The most useful thing you guys could do is add gradle files to the samples. At least one could build them and start learning from them then.

PhilLab:
@hauptmech: is it possible for you to create a pull request containing that?

hauptmech:
Don't have time. Also anything I came up with would be nothing close to best practice which is what you want for a sample.

You will need to decide whether to support ndk-build or android-studio with gradle-experimental plugin.

hauptmech:
I revisited the OpenCV samples as prep to tackle the ROS android stuff. Using gradle-experimental and an in-situ build.gradle file for the library worked well for me. Possibly neither are best practice but after a couple days of working with this as a beginner, I can say that this approach is much easier to wrap my head around than the Import-to-android-studio way which has generally unreliable documentation and seems to change every few months.

I jotted down what I did here: http://hauptmech.github.io/OpenCV1/

PhilLab:
[...]

PR #9476, opened 2017-08-26
Android sample for DNN module

dkurt:
### This pullrequest changes
* MobileNet object detection+classification sample ported onto Android device.
* Tutorial for DNN module on Android and generally how to use OpenCV in Android Studio (existing examples consider Eclipse).

Sample isn't included into CMakeLists.txt because it won't work without huge `.caffeproto` and `.prototxt` anyway.


PhilLab:
Awesome! On a site note: this contains a small section which would also satisfy the long-awaited AndroidStudio tutorial (discussed together with the fill migration of samples here: https://github.com/opencv/opencv/issues/5003)

samples/android/mobilenet-objdetect/src/org/opencv/samples/opencv_mobilenet/MainActivity.java
```diff
@@ -0,0 +1,193 @@
...
+                    break;
+                }
+            }
+        }
+    };
+
+    @Override
+    public void onResume() {
+        super.onResume();
+        OpenCVLoader.initAsync(OpenCVLoader.OPENCV_VERSION_3_3_0, this, mLoaderCallback);
```
alalek: Perhaps we want to support the current OpenCV version only:
`OPENCV_VERSION_3_3_0` -> `OPENCV_VERSION`

samples/android/mobilenet-objdetect/src/org/opencv/samples/opencv_mobilenet/MainActivity.java
```diff
@@ -0,0 +1,193 @@
...
+
+public class MainActivity extends AppCompatActivity implements CvCameraViewListener2 {
+
+    // Initialize OpenCV manager.
+    private BaseLoaderCallback mLoaderCallback = new BaseLoaderCallback(this) {
+        @Override
+        public void onManagerConnected(int status) {
+            switch (status) {
+                case LoaderCallbackInterface.SUCCESS: {
+                    Log.i("MyTag", "OpenCV loaded successfully");
```
alalek: Declare this in the main class:
```
private static final String  TAG = "OpenCV::Sample::MobileNet";
```
and use it instead of "MyTag" (multiple locations).

samples/android/mobilenet-objdetect/src/org/opencv/samples/opencv_mobilenet/MainActivity.java
```diff
@@ -0,0 +1,193 @@
...
+        return "";
+    }
+
+    // Remove alpha channel from image.
+    private static Mat RGBA2RGB(Mat img) {
+        ArrayList<Mat> channels = new ArrayList<Mat>();
+        Core.split(img, channels);
+        channels.remove(3);
+        Mat res = new Mat(img.size(), CvType.CV_8UC3);
+        Core.merge(channels, res);
```
alalek: cvtColor?

dkurt: Indeed! Thanks!

samples/android/mobilenet-objdetect/src/org/opencv/samples/opencv_mobilenet/MainActivity.java
dkurt: Replaced to `OpenCV/Sample/MobileNet` because of restriction with 23 characters of tag length by default.

alalek:
Looks good to me :+1:

PR #9479, opened 2017-08-27
videoio(v4l): move MJPEG/JPEG to the end of probe list

### LINKED ISSUE #9477: corrupt JPEG data: # extraneous bytes before marker 0x##
- OpenCV => 3.3
- Operating System / Platform => Ubuntu 16.04.3 LTS (x86_64)
- Compiler => gcc 5.4.0 20160609

##### Detailed description
When accessing a Logitech C310 webcam with Java OpenCV 3.3 the following error message is repeatedly printed to the terminal (with varying values for number of bytes and marker)
`Corrupt JPEG data: 2 extraneous bytes before marker 0xd3`

The error message comes from 3rdparty/libjpeg/jdmarker.c
usage of JWRN_EXTRANEOUS_DATA

The problem appears to be caused by a bug with the libjpeg that ships with OpenCV. A fix is to build OpenCV with the option `WITH_JPEG` switched OFF causing OpenCV to use the platform's copy of libjpeg.

I recommend updating the instructions [here](http://docs.opencv.org/2.4/doc/tutorials/introduction/desktop_java/java_dev_intro.html)
to include `WITH_JPEG=OFF`

##### Steps to reproduce
Build OpenCV for Java and run in Linux using the settings recommended in the [Introduction to Java Development](http://docs.opencv.org/2.4/doc/tutorials/introduction/desktop_java/java_dev_intro.html)

alalek:
Workaround to hide JPEG warnings.

resolves #9477

### DIFF
--- a/modules/videoio/src/cap_v4l.cpp
+++ b/modules/videoio/src/cap_v4l.cpp
@@ -460,18 +460,18 @@ static int autosetup_capture_mode_v4l2(CvCaptureCAM_V4L* capture) {
     }
     __u32 try_order[] = {
             V4L2_PIX_FMT_BGR24,
+            V4L2_PIX_FMT_RGB24,
             V4L2_PIX_FMT_YVU420,
             V4L2_PIX_FMT_YUV411P,
+            V4L2_PIX_FMT_YUYV,
+            V4L2_PIX_FMT_UYVY,
+            V4L2_PIX_FMT_SBGGR8,
+            V4L2_PIX_FMT_SGBRG8,
+            V4L2_PIX_FMT_SN9C10X,
 #ifdef HAVE_JPEG
             V4L2_PIX_FMT_MJPEG,
             V4L2_PIX_FMT_JPEG,
 #endif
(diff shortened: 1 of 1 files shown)


paroj:
this merely hides the issue, but does not resolves it. moreover we actually want to prefer MJPEG over YUV etc. here as cameras offer higher resolution/ frame-rate options in MJPEG mode.

vpisarev:
@alalek, @paroj, let's merge it in. For now it can be considered as a workaround, which fixes a particular problem of concrete user. Perhaps it should be some configuration option, or proper implementation of `cap.set(CAP_PROP_FOURCC, CV_FOURCC('M', 'J', 'P', 'G'))` so that user can configure camera to use mjpeg instead of yuv. :+1:

paroj:
`cap.set(CAP_PROP_FOURCC, CV_FOURCC('M', 'J', 'P', 'G')` actually does already work.
My point was that a warning in libjpeg is preferable to not being able to select certain resolutions by default.

PR #9492, opened 2017-08-28
videoio: ximea - allow opening capture by serial number

paroj:


alalek:
Looks good to me :+1:

PR #9498, opened 2017-08-29
 i9403: Removing checks from trackbar min/max set functions.

### LINKED ISSUE #9403: Allow negative values for setTrackbarMin
- OpenCV => 3.3
- Operating System / Platform => Ubuntu 16.04 (x86-64) with QT 5
- Compiler => g++ 5.4.0

##### Detailed description

`setTrackbarMin` currently ignores negative parameters, making it impossible to set a negative minimum for a trackbar when using QT. The latter's `QSlider` which OpenCV relies on does not seem to have such a restriction (or, if so, it is also undocumented, cf. http://doc.qt.io/qt-5/qabstractslider.html#minimum-prop). If OpenCV keeps this restriction (`minval` needs to be positive) for compatibility (or other reasons), it should be mentioned in the documentation.

##### Steps to reproduce

There is currently no code path for the opposite case of
https://github.com/opencv/opencv/blob/316fb26f5badb2cbb3cff3dfbc284dcc29115759/modules/highgui/src/window_QT.cpp#L691

saskatchewancatch:
resolves #9403

### DIFF
--- a/modules/highgui/src/window_QT.cpp
+++ b/modules/highgui/src/window_QT.cpp
@@ -673,30 +673,20 @@ CV_IMPL void cvSetTrackbarPos(const char* name_bar, const char* window_name, int
 
 CV_IMPL void cvSetTrackbarMax(const char* name_bar, const char* window_name, int maxval)
 {
-    if (maxval >= 0)
+    QPointer<CvTrackbar> t = icvFindTrackBarByName(name_bar, window_name);
+    if (t)
     {
-        QPointer<CvTrackbar> t = icvFindTrackBarByName(name_bar, window_name);
-        if (t)
-        {
-            int minval = t->slider->minimum();
-            maxval = (maxval>minval)?maxval:minval;
-            t->slider->setMaximum(maxval);
-        }
+        t->slider->setMaximum(maxval);
     }
 }
 
 
 CV_IMPL void cvSetTrackbarMin(const char* name_bar, const char* window_name, int minval)
 {
-    if (minval >= 0)
+    QPointer<CvTrackbar> t = icvFindTrackBarByName(name_bar, window_name);
+    if (t)
     {
-        QPointer<CvTrackbar> t = icvFindTrackBarByName(name_bar, window_name);
-        if (t)
-        {
-            int maxval = t->slider->maximum();
-            minval = (maxval<minval)?maxval:minval;
-            t->slider->setMinimum(minval);
-        }
+        t->slider->setMinimum(minval);
     }
 }
 


modules/highgui/src/window_QT.cpp
```diff
@@ -673,30 +673,24 @@ CV_IMPL void cvSetTrackbarPos(const char* name_bar, const char* window_name, int
...
     {
-        QPointer<CvTrackbar> t = icvFindTrackBarByName(name_bar, window_name);
-        if (t)
-        {
-            int minval = t->slider->minimum();
-            maxval = (maxval>minval)?maxval:minval;
-            t->slider->setMaximum(maxval);
-        }
+        int minval = t->slider->minimum();
+        maxval = (maxval>minval)?maxval:minval;
```
alalek: This check/truncation is buggy too.

saskatchewancatch: @alalek That check and the similar [one](https://github.com/opencv/opencv/blob/master/modules/highgui/src/window_QT.cpp#L697) in `cvSetTrackbarMin` are trying to check that the range is valid. If the `new_max < min` then it just sets the `new_max` to the same as `min`.

I am not sure if this is what you meant.

May I suggest removing that check altogether and the similar [check](https://github.com/opencv/opencv/blob/master/modules/highgui/src/window_QT.cpp#L697) in `cvSetTrackbarMin`? The reason is that Qt5 manages range checking under the hood quite nicely. It seems what Qt5 does under the hood is more intuitive than what those checks are trying to do.

What do you think?

alalek: I believe OpenCV should do that user asks from OpenCV - without unclear internal logic like min/max values truncation.

This users code should not confuse users:
```
setTrackbarMax(..., -5);
setTrackbarMin(..., -10);
// expected range is [-10..-5] , not [-10..0]
```

In GTK backend I [leave](https://github.com/opencv/opencv/commit/9e239c7a1e86bc76d066c165559df32bacc44dd3#diff-922827782cb034256bf426c7e8d1fe59R1470) this code:
```
trackbar->maxval = maxval;
if (trackbar->maxval >= trackbar->minval)
    gtk_range_set_range(GTK_RANGE(trackbar->widget), trackbar->minval, trackbar->maxval);
```

trackbar->minval / trackbar->maxval is fields from OpenCV trackbar wrapper (need to add them into QT `cvTrackbar`).

If QT handles this internally then we could just call QT directly without extra logic.

BTW,
Perhaps we should have `setTrackbarMinMax` API to avoid inconsistent usage, but it is quite large set of changes for now.

saskatchewancatch: Excellent point @alalek.

In the case of the QT4/5, the backend handling appears to do exactly what you desire. So I think it should be fine to remove those checks.

With those checks removed, when I try
```
setTrackbarMax(..., -5);
setTrackbarMin(..., -10);
// expected range is [-10..-5] , not [-10..0]
```

I get the expected range of [-10..-5].

modules/highgui/src/window_QT.cpp
```diff
@@ -673,30 +673,22 @@ CV_IMPL void cvSetTrackbarPos(const char* name_bar, const char* window_name, int
...
     {
-        QPointer<CvTrackbar> t = icvFindTrackBarByName(name_bar, window_name);
-        if (t)
-        {
-            int minval = t->slider->minimum();
-            maxval = (maxval>minval)?maxval:minval;
-            t->slider->setMaximum(maxval);
-        }
+        int minval = t->slider->minimum();
+        t->slider->setMaximum(maxval);
```
alalek: `int minval` and maxval in the function below are not used please remove them and squash git commits into one before merge.

alalek:
Well done! Thank you :+1:

PR #9503, opened 2017-08-29
imgproc: disable buggy inplace processing in convexHull

### LINKED ISSUE #9472: usage of cv::convexHull()
- OpenCV => 3.3

##### Detailed description

`convexHull()` don't support for inplace parameters usage like `convexHull(points, points, true)`
when the function called like `convexHull(points, points, true)` it gives sometimes correct sometimes wrong results ( can be seen in the image i get running the following code )

the function either need to be fixed for supporting inplace parameters usage or should give an error when the input and output is same. 

hull

##### Steps to reproduce

```
#include <iostream>

using namespace cv;
using namespace std;

int main( int argc, char** argv )
{
    Mat img(500, 500, CV_8UC3);
    RNG& rng = theRNG();

    for(;;)
    {
        int i, count = (unsigned)rng%100 + 1;

        vector<Point> points;

        for( i = 0; i < count; i++ )
        {
            Point pt;
            pt.x = rng.uniform(img.cols/4, img.cols*3/4);
            pt.y = rng.uniform(img.rows/4, img.rows*3/4);

            points.push_back(pt);
        }

        img = Scalar::all(0);
        for( i = 0; i < count; i++ )
            circle(img, points[i], 3, Scalar(0, 0, 255), FILLED, LINE_AA);

		convexHull(points, points, true);
		polylines(img, points, true, Scalar(255, 0, 0), 1);
		imshow("hull", img);

        char key = (char)waitKey();
        if( key == 27 || key == 'q' || key == 'Q' ) // 'ESC'
            break;
    }

    return 0;
}
```
-->

sovrasov:
resolves #9472

### DIFF
--- a/modules/imgproc/include/opencv2/imgproc.hpp
+++ b/modules/imgproc/include/opencv2/imgproc.hpp
@@ -3977,6 +3977,8 @@ returns convex hull points. Otherwise, it returns indices of the convex hull poi
 output array is std::vector, the flag is ignored, and the output depends on the type of the
 vector: std::vector\<int\> implies returnPoints=false, std::vector\<Point\> implies
 returnPoints=true.
+
+@note `points` and `hull` should be different arrays, inplace processing isn't supported.
  */
 CV_EXPORTS_W void convexHull( InputArray points, OutputArray hull,
                               bool clockwise = false, bool returnPoints = true );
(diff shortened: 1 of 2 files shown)


modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -3977,6 +3977,8 @@ returns convex hull points. Otherwise, it returns indices of the convex hull poi
 output array is std::vector, the flag is ignored, and the output depends on the type of the
 vector: std::vector\<int\> implies returnPoints=false, std::vector\<Point\> implies
 returnPoints=true.
+
+@note `points` and `hull` should be different arrays, inplace processing doesn't supported.
```
lpea: "doesn't supported" => "isn't supported"

alalek:
:+1:

PR #9506, opened 2017-08-29
OCL: imgproc: eliminate div by zero in Canny

### LINKED ISSUE #9496: OpenCL: Divide by zero in `stage1_without_sobel` kernel

- OpenCV => 3.3.0-dev

##### Detailed description

On lines #224 and #345 of [modules/imgproc/src/opencl/canny.cl](https://github.com/opencv/opencv/blob/3.3.0/modules/imgproc/src/opencl/canny.cl) I'm seeing undefined behaviour with
regards to divide by zero when initalizing variables a & b.

E.g the line
```
int a = (y / (float)x) * TG22;
```

Can be broken down as such:
```
float div = y / (float)x;

// If x is 0 and y is +ve non-zero, div will be +INFINITY

float mul = div * TG22;

// Mul is INFINITY

int a = (int)mul;

// converting infinity to an integer is undefined behaviour
```

Rewriting this as below should resolve this, can file a pull request if it's acceptable.
```
int a = 0, b = 0;
if (x != 0)
{
   a = (y / (float)x) * TG22;
   b = (y / (float)x) * TG67;
}
```

##### Steps to reproduce
This was discovered using an internal OpenCL implementation, so may not be easily reproducible.

alalek:
resolves #9496

### DIFF
--- a/modules/imgproc/src/opencl/canny.cl
+++ b/modules/imgproc/src/opencl/canny.cl
@@ -221,11 +221,11 @@ __kernel void stage1_with_sobel(__global const uchar *src, int src_step, int src
     int value = 1;
     if (mag0 > low_thr)
     {
-        int a = (y / (float)x) * TG22;
-        int b = (y / (float)x) * TG67;
+        float x_ = abs(x);
+        float y_ = abs(y);
 
-        a = min((int)abs(a), 1) + 1;
-        b = min((int)abs(b), 1);
+        int a = (y_ * TG22 >= x_) ? 2 : 1;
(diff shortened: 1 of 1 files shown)


EwanC:
I think you need to do this for the case on line [345](https://github.com/alalek/opencv/blob/080ebf076405b11e6ece7d06c5d2762246bee659/modules/imgproc/src/opencl/canny.cl#L345) as well.
With the change in both locations, this fixes https://github.com/opencv/opencv/issues/9496 for me locally

alalek:
@EwanC Right!

vpisarev:
:+1:

PR #9507, opened 2017-08-29
Restrict std::initializer_list constructors to arithmetic types.

de-vri-es:
This PR restricts the `cv::Mat::Mat(const std::initializer_list<_Tp>)` constructor to arithmetic types. That way, it won't be used in place of the copy constructor in this code:

```
cv::Mat image = cv::imread("image.png");
cv::Mat copy{image};  // Currently calls the initializer_list constructor which turns it into a 1x1 matrix.
```

Or:
```
struct Foo {
  cv::Mat image;
  Foo(cv::Mat const & mat) : image{mat} {} // also calls wrong cv::Mat constructor 
}
```

The `initializer_list` constructor broke existing code where the copy constructor was invoked with curly braces. The code worked fine with opencv 3.2, but not anymore with 3.3. I think it is quite important to fix quickly.

The initializer list constructor doesn't make sense for non-scalar values anyway.

### DIFF
--- a/modules/core/include/opencv2/core/mat.hpp
+++ b/modules/core/include/opencv2/core/mat.hpp
@@ -53,6 +53,10 @@
 
 #include "opencv2/core/bufferpool.hpp"
 
+#ifdef CV_CXX11
+#include <type_traits>
+#endif
+
 namespace cv
 {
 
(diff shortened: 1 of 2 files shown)


vpisarev:
thanks! :+1:

PR #9509, opened 2017-08-29
Addressing CUDA9 shfl deprecation

borisfom:
### This pullrequest changes

In CUDA9, non-sync __shfl's are deprecated - they generate warnings and may not work properly.


alalek:
:+1:

PR #9510, opened 2017-08-30
Update stitching.cpp

sturkmen72:
### This pullrequest changes

Updates stitching.cpp
a new flag "--d3" added. 
when this flag used the program creates three chunks of each image to increase stitching success
you can find more detail about the changes at [this question on answers.opencv.org](http://answers.opencv.org/question/98110)


sovrasov:
:+1:

PR #9518, opened 2017-08-30
flann: std::vector<size_t> can't be converted into cv::Mat properly

alalek:
`size_t` is not supported `cv::Mat` type


sovrasov:
:+1:

PR #9521, opened 2017-08-31
Don't truncate label for tracknbar name (Qt)

### LINKED ISSUE #9489: Allow longer trackbar names
- OpenCV => 3.3
- Operating System / Platform => Ubuntu 16.04 (x86-64) with QT 5.5
- Compiler => g++ 5.4.0

##### Detailed description

Trackbar names are truncated after the tenth character, regardless of their actual length on-screen. The following example illustrates the effect:

##### Steps to reproduce

```

using namespace std;
using namespace cv;

static constexpr auto trackbar_name = "X angle [deg.]"; //Use "X angle [°]" to see the same effect

int main()
{
  namedWindow("Test window");
  imshow("Test window", Mat(600, 600, CV_8UC1, Scalar(128)));
  int dummy;
  createTrackbar(trackbar_name, "Test window", &dummy, 360, [](const int, void * const){});
  waitKey(0);
  return 0;
}
```

This limitation does not seem to be documented. But even so, it would be more desirable to use names as long as one pleases, i.e., without arbitrary character limitations. Being limited to ten characters is not really meaningful when using narrow characters/symbols or wide windows.

saskatchewancatch:
resolves #9489 

### This pullrequest changes

Setting truncate option for QString leftJustified method to false fixes 10 character truncation issue noted in issue 9489. This will allow trackbar names > 10 character in length to show up.

### DIFF
--- a/modules/highgui/src/window_QT.cpp
+++ b/modules/highgui/src/window_QT.cpp
@@ -1409,7 +1409,7 @@ void CvTrackbar::update(int myvalue)
 
 void CvTrackbar::setLabel(int myvalue)
 {
-    QString nameNormalized = name_bar.leftJustified( 10, ' ', true );
+    QString nameNormalized = name_bar.leftJustified( 10, ' ', false );
     QString valueMaximum = QString("%1").arg(slider->maximum());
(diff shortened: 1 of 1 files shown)


sovrasov:
:+1:

dustsigns:
Works for me.

PR #9534, opened 2017-09-01
cmake: cl2cpp change main OUTPUT to .hpp file

alalek:
To reduce number of unnecessary rebuilds on updates of OpenCL kernels

Without patch we have this rule in Makefile:
```
modules/core/opencl_kernels_core.hpp: modules/core/opencl_kernels_core.cpp
	@$(CMAKE_COMMAND) -E touch_nocreate modules/core/opencl_kernels_core.hpp
```
It always update timestamp of `.hpp` file even it is not changed, so many .cpp files are got rebuild.


mshabunin:
:+1:

PR #9551, opened 2017-09-04
Add multi-channel mask support to mean, meanStdDev and setTo

ChristofKaufmann:
This adds the possibility to use multi-channel masks for the functions `cv::mean`, `cv::meanStdDev` and the method `Mat::setTo`. The tests have now a probability to use multi-channel masks for operations that support them. This also includes `Mat::copyTo`, which supported multi-channel masks before, but there was no test confirming this.

### DIFF
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -608,7 +608,7 @@ CV_EXPORTS_W void findNonZero( InputArray src, OutputArray idx );
 
 The function cv::mean calculates the mean value M of array elements,
 independently for each channel, and return it:
-\f[\begin{array}{l} N_c =  \sum _{I: \; {\texttt{mask} (I)_c} \ne 0} 1 \\ M_c =  \left ( \sum _{I: \; {\texttt{mask} (I)_c} \ne 0}{ \texttt{src} (I)_c} \right )/N_c \end{array}\f]
+\f[\begin{array}{l} N =  \sum _{I: \; \texttt{mask} (I) \ne 0} 1 \\ M_c =  \left ( \sum _{I: \; \texttt{mask} (I) \ne 0}{ \texttt{mtx} (I)_c} \right )/N \end{array}\f]
 When all the mask elements are 0's, the function returns Scalar::all(0)
 @param src input array that should have from 1 to 4 channels so that the result can be stored in
 Scalar_ .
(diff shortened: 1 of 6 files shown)


alalek:
Thank you for contribution.

Are there any real use-cases for this functionality?
It make sense to mask image region via whole pixels. Are there any sense to mask "green" channel from image region?

ChristofKaufmann:
I guess. For example when you have an int16 image with multiple channels, but you know that only values in the range [0, 10000] are valid. This is common for satellite images. Then values might be invalid in a single channel only. Depending on the bands you select for the multi-channel image, some bands might suffer from clouds, others not.

vpisarev:
why cannot users just split the images into separate channels (using cv::split) and then apply whatever processing they want to those channels? I guess, mean, meanStdDev and setTo are not the only operations that are need for such images? The thing is that each such operation, especially replicated for all the supported datatypes increases the library size. I would think of using a workaround for such rarely used functionality instead of putting it in.

ChristofKaufmann:
I understand this argument. However the three modified functions seem to be the most useful ones. Especially `Mat::setTo` is helpful and requires the least code changes (non-templated code). Also the partner function `Mat::copyTo` supports "colorMasks" with similar code. With the ability to set values to zero with a multi-channel mask all "operations on arrays" can be used indirectly with multi-channel masks by setting the masked-out values to zero, except `mean` and `meanStdDev`. For these two functions the count would be wrong and therefore indeed a split of the image and the mask would be required. This is why I changed these two to support multi-channel masks directly.

However, if the code size increases too much by the modified `mean` and `meanStdDev` (templated code), I can make a new pull request which only changes `Mat::setTo` and adds the tests for `Mat::setTo` and `Mat::copyTo`.

vpisarev:
@ChristofKaufmann, great! I think, extending `Mat::setTo()` and leaving mean and meanStdDev untouched would be a good compromise! There is no need to make a new PR, just put new commits to this one. I will squash and merge it in as soon as it's ready and passes our buildbot tests

vpisarev:
thank you! :+1:

alalek:
There are no tests for new functionality, especially for OpenCL/IPP code path.

ChristofKaufmann:
There are tests for the new functionality. They are integrated in `test_arithm.cpp` and choose with a probability of 0.5 multi-channel masks instead of single-channel masks for `copyTo` and `setTo`. Are there special tests required for IPP code path? The IPP code should be used in `copymask`and thus tested already with the usual tests.

I am not sure about OpenCL, but the new code make multi-channel images with multi-channel masks look like larger single-channel images with a single-channel mask. So extra treatment should not be required.

Issue #8693, opened 2017-05-04
UMat refcount Error when setting UseOpenCL to flase


theaniolad:
##### System information (version)
- OpenCV => 3.2 
- Operating System / Platform => Windows 64 Bit
- Compiler => Visual Studio 2015

##### Detailed description

I'm currently changing all Mats in my program to UMats and have a crash with some functions (cv::convertTo and cv::reduce) which I don't think should happen. 

1. Code runs fine with Mats
2. Code runs fine with UMats if UseOpenCL set to true
3. Code crashes with **UMats if UseOpenCL set to false**

##### Steps to reproduce
```
	ocl::setUseOpenCL(false);
	auto src = UMat::zeros({ 10,10 }, CV_32FC1);
	reduce(src, src, 0, CV_REDUCE_SUM);  //crashes with refcount Error
       src.convertTo(src, CV_64FC1); //crashes with refcount Error

````
**The error is:**  
_**OpenCV Error: Assertion failed (u->refcount == 0) in cv::StdMatAllocator::deallocate, file opencv\sources\modules\core\src\matrix.cpp, line 214**_

Now you might say that I should simply use a different output UMat, but I would also like to write functions like this:

````
     void foo(cv::InputArray src, cv::OutputArray dst)
   {
  	cv::reduce(src, dst, 0, CV_REDUCE_SUM);
       //do other stuff
   }
   
  int main()
  {
	ocl::setUseOpenCL(false);
	auto src = UMat::zeros({ 10,10 }, CV_32FC1);
        UMat dst;
       	foo(src, src);
}
```

Currently I'm using the following fix to make the program run for all three cases:

```
 void foo(cv::InputArray src, cv::OutputArray dst)
{
     if (src.isUMat() && !ocl::useOpenCL())
    {
        cv::reduce(src.getUMat(), dst, 0, CV_REDUCE_SUM;
    }
     else 
    {
        cv::reduce(src, dst, 0, CV_REDUCE_SUM;
    }
}
```

dkurt:
Minimal example to reproduce it (summarized from `cv::reduce` code):
```cpp
void func(InputArray _src, OutputArray _dst) {
  Mat src = _src.getMat();
  _dst.create(10, 1, src.type());
}

int main(int argc, char** argv) {
  ocl::setUseOpenCL(false);
  UMat src(10, 10, CV_32FC1);
  func(src, src);
}
```

alalek:
Related: #8352

PR #9561, opened 2017-09-05
Fix UMat deallocation

dkurt:
### This pullrequest changes
resolves https://github.com/opencv/opencv/issues/8693


dkurt:
Looks like `testTempObjects_Mat` is an opposite test case.

alalek:
Workaround is fine to me. But I believe, "Inplace" processing doesn't make sense for `reduce()` operation (it is not "scalar multiplication" or similar) - perhaps we should just forbid them.

/cc @vpisarev

vpisarev:
let's merge it in for now. Fortunately, the fix does not impact performance when the operation is not in-place. :+1:

PR #9564, opened 2017-09-06
Allow TIFF min=white PHOTOMETRIC interpretation images to be read unc…

### LINKED ISSUE #9550: Incorrect handling of 16 bit tiff images with photometric interpretation 0

- OpenCV => 3.3

##### Detailed description

cv::imread seems to handle 16 bit grayvalue tif images incorrectly when photometric interpretation is 0 ("WhiteIsZero"). The resulting cv::Mat has type CV_8UC1. When the photometric interpretation is 1, however, the image is loaded correctly as CV_16UC1.

I suspect the error to be in the condition starting from line 154 in modules/imgcodecs/src/grfmt_tiff.cpp, where photometric==0 leads to setting bpp to 8.

##### Steps to reproduce

Test images were created using ImageJ using File->New->Image... with 16-bit. One image was inverted using Image->Color->Edit LUT...->Invert,,. 

tiffdump showed the photometric interpretation to be the only tag being different in the two images.

Images were loaded using the following code:

```cpp
cv::Mat image = cv::imread("16bit.tif", cv::IMREAD_UNCHANGED);
cv::Mat image_inv = cv::imread("16bit_inv.tif", cv::IMREAD_UNCHANGED);

std::cout << image.type() << std::endl;			// prints '2'
std::cout << image_inv.type() << std::endl;		// prints '0'
```

saskatchewancatch:
…hanged

resolves #9550

### DIFF
--- a/modules/imgcodecs/src/grfmt_tiff.cpp
+++ b/modules/imgcodecs/src/grfmt_tiff.cpp
@@ -235,7 +235,7 @@ bool TiffDecoder::readHeader()
             m_hdr = false;
 
             if( bpp > 8 &&
-               ((photometric != 2 && photometric != 1) ||
+               ((photometric > 2) ||
                 (ncn != 1 && ncn != 3 && ncn != 4)))
                 bpp = 8;
 


sovrasov:
:+1:

PR #9566, opened 2017-09-06
export Python signatures information

alalek:
related #9260

### DIFF
--- a/modules/python/common.cmake
+++ b/modules/python/common.cmake
@@ -48,17 +48,28 @@ set(cv2_generated_hdrs
     "${CMAKE_CURRENT_BINARY_DIR}/pyopencv_generated_funcs.h"
     "${CMAKE_CURRENT_BINARY_DIR}/pyopencv_generated_types.h"
     "${CMAKE_CURRENT_BINARY_DIR}/pyopencv_generated_type_reg.h"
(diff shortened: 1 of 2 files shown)


Cartucho:
@alalek I get the following error:
image

alalek:
Looks like you observe conflict with this patch: #8895
Did you rebase on latest OpenCV master?
I don't observe similar issue on OpenCV builders.

Cartucho:
Yes, you are right, now it worked

Cartucho:
So the generated json file contains the functions with the name, ret and args:
image

But now we still need to add the signatures to the generated docs, right?

alalek:
I have some experiments with adoption here: https://github.com/alalek/opencv/commits/pr9260_r (please check [comment](https://github.com/opencv/opencv/pull/9260#issuecomment-327382462) )

alalek:
JSON file is a dictionary with:
- key: C++ function/method name
- value: array of function "variants" (overloads):
  - element: dictionary of
    - arg: function/method arguments (there is no "self" for methods, probably we should add it later)
    - dst: rhs expression (output results)
    - name: Python name (useful for handling of CV_WRAP_AS).

vpisarev:
cool! :+1:

PR #9389, opened 2017-08-17
Advanced memory management for DNN blobs

dkurt:
### This pullrequest changes
There are two methods: optimal for host memory that will be processed on CPU and a less optimal for device memory (OpenCL). The last one follows the same `reuse or create` strategy as we use now, but it makes more dense packing. Achieved results:

Model | with PR, host memory | with PR, device memory | before PR, host / dev | No reusing
-- | -- | -- | -- | --
AlexNet | 2.32MB | 2.6MB | 3.72MB | 6.14MB
Inception-5h | 5.2MB | 8.65MB | 15.62MB | 33.13MB
GoogLeNet | 5.21MB | 8.76MB | 15.74MB | 33.43MB
ENet, 512x256 | 23.59MB | 26.24MB | 72.22MB | 137.9MB
SqueezeNet v1.1 | 4.87MB | 5.07MB | 11.71MB | 20.71MB
ResNet-50 | 9.63MB | 10.43MB | 31.51MB | 66.64MB

* Despite at the same time is actual only one memory block (host or device) and we can skip some allocations of host memory, we can't make a decision about it at the allocation stage. So, in case of OpenCL backend, we allocate both CPU and GPU memory (as it was before PR).
* Outputs with no references (like indices of MaxPooling) are allocated too. Otherwise it will be more complicated for maintaining.
* Required changes: https://github.com/opencv/opencv/pull/9384.

PR #9569, opened 2017-09-06
MobileNet-SSD and VGG-SSD topologies in Halide

dkurt:
### This pullrequest changes

* MobileNet- and VGG- based SSD models in Halide. The key difference is that there are layers from different backends and target devices (i.e. part of layers works on GPU with Halide backend and the rest on CPU with default backend).
* Parallel Halide layers compilation.

This PR implements part of https://github.com/opencv/opencv/pull/9389. So the last one will be lighter after it.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/372


vpisarev:
:+1:

PR #9574, opened 2017-09-07
I9482

### LINKED ISSUE #9482: NormTypes incomplete documentation
The documentation for NormTypes does not explain all options, in particular the difference between NORM_L2 and NORM_L2SQR:
http://docs.opencv.org/3.3.0/d2/de8/group__core__array.html#gad12cefbcb5291cf958a85b4b67b6149f

It wasn't obvious for me at first glance, that SQR means squared. I would add either
1) A case with the formula like for NORM_INF, _L1 and _2
2) Add a sentence describing the difference: "NORM_LF2SQR calculates the same norm as NORM_L2, but without taking the square root of the result."

saskatchewancatch:
### This pullrequest changes

Commit messages explain what's been done. Documentation for cv::norm has been slightly clarified by explicating NORM_L2SQR, and 2 array overload now has its own documentation.

NORM_TYPE_MASK is now masked out of doc generation since it's use is internal.

This is work in progress, but I've PR'd this to get some initial feedback. 

Please feel free to comment on this and add your own additions. Work needing to be done:

1) Clarify NORM_L2SQR for relative norm case. 
2) Clarify the Hamming norms.

### DIFF
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -640,34 +640,44 @@ Scalar_ 's.
 CV_EXPORTS_W void meanStdDev(InputArray src, OutputArray mean, OutputArray stddev,
                              InputArray mask=noArray());
 
-/** @brief Calculates an absolute array norm, an absolute difference norm, or a
-relative difference norm.
+/** @brief Calculates an absolute array norm.
 
-The function cv::norm calculates an absolute norm of src1 (when there is no
-src2 ):
+This version of cv::norm calculates the absolute norm of src1. The type of norm to calculate is specified using cv::NormTypes.
 
-\f[norm =  \forkthree{\|\texttt{src1}\|_{L_{\infty}} =  \max _I | \texttt{src1} (I)|}{if  \(\texttt{normType} = \texttt{NORM_INF}\) }
+\f[norm =  \forkfour{\|\texttt{src1}\|_{L_{\infty}} =  \max _I | \texttt{src1} (I)|}{if  \(\texttt{normType} = \texttt{NORM_INF}\) }
 { \| \texttt{src1} \| _{L_1} =  \sum _I | \texttt{src1} (I)|}{if  \(\texttt{normType} = \texttt{NORM_L1}\) }
-{ \| \texttt{src1} \| _{L_2} =  \sqrt{\sum_I \texttt{src1}(I)^2} }{if  \(\texttt{normType} = \texttt{NORM_L2}\) }\f]
+{ \| \texttt{src1} \| _{L_2} =  \sqrt{\sum_I \texttt{src1}(I)^2} }{if  \(\texttt{normType} = \texttt{NORM_L2}\) }
+{ \| \texttt{src1} \| _{L_2} ^{2} = \sum_I \texttt{src1}(I)^2} {if  \(\texttt{normType} = \texttt{NORM_L2SQR}\)}\f]
+
+If normType is not specified, NORM_L2 is used.
 
 or an absolute or relative difference norm if src2 is there:
 
-\f[norm =  \forkthree{\|\texttt{src1}-\texttt{src2}\|_{L_{\infty}} =  \max _I | \texttt{src1} (I) -  \texttt{src2} (I)|}{if  \(\texttt{normType} = \texttt{NORM_INF}\) }
-{ \| \texttt{src1} - \texttt{src2} \| _{L_1} =  \sum _I | \texttt{src1} (I) -  \texttt{src2} (I)|}{if  \(\texttt{normType} = \texttt{NORM_L1}\) }
-{ \| \texttt{src1} - \texttt{src2} \| _{L_2} =  \sqrt{\sum_I (\texttt{src1}(I) - \texttt{src2}(I))^2} }{if  \(\texttt{normType} = \texttt{NORM_L2}\) }\f]
 
-or
-
-\f[norm =  \forkthree{\frac{\|\texttt{src1}-\texttt{src2}\|_{L_{\infty}}    }{\|\texttt{src2}\|_{L_{\infty}} }}{if  \(\texttt{normType} = \texttt{NORM_RELATIVE | NORM_INF}\) }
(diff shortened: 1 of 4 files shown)


saskatchewancatch:
As commit comment above says, a bit of work is left to be done.

modules/core/include/opencv2/core.hpp
```diff
@@ -676,7 +686,24 @@ the results for all channels are combined.
...
+{ \| \texttt{src1} - \texttt{src2} \| _{L_2} =  \sqrt{\sum_I (\texttt{src1}(I) - \texttt{src2}(I))^2} }{if  \(\texttt{normType} = \texttt{NORM_L2}\) }
+{ \| \texttt{src1} - \texttt{src2} \| _{L_2} ^{2} =  \sum_I (\texttt{src1}(I) - \texttt{src2}(I))^2 }{if  \(\texttt{normType} = \texttt{NORM_L2SQR}\) }
+\f]
+
+or
+
+\f[norm =  \forkthree{\frac{\|\texttt{src1}-\texttt{src2}\|_{L_{\infty}}    }{\|\texttt{src2}\|_{L_{\infty}} }}{if  \(\texttt{normType} = \texttt{NORM_RELATIVE | NORM_INF}\) }
+{ \frac{\|\texttt{src1}-\texttt{src2}\|_{L_1} }{\|\texttt{src2}\|_{L_1}} }{if  \(\texttt{normType} = \texttt{NORM_RELATIVE | NORM_L1}\) }
+{ \frac{\|\texttt{src1}-\texttt{src2}\|_{L_2} }{\|\texttt{src2}\|_{L_2}} }{if  \(\texttt{normType} = \texttt{NORM_RELATIVE | NORM_L2}\) }\f]
+
```
sovrasov: Relative `NORM_L2SQR` is the same as relative `NORM_L2` but squared.
All the relative norms are computed as follows:
```c++
if( normType & CV_RELATIVE )
    {
        return norm(_src1, _src2, normType & ~CV_RELATIVE, _mask)/(norm(_src2, normType, _mask) + DBL_EPSILON);
    }
```

sovrasov:
Hamming norms are valid only for `CV8U` depth.
In case of `NORM_HAMMING` output of the `norm` function will be number of ones in binary sequence representing input array.
In case of `NORM_HAMMING2` each two bits of input sequence will be added and threated as one bit in simple hamming norm, so if we would have `NORM_HAMMING8` it would be equivalent to number of nonzero `8U` elements in input array.

modules/core/include/opencv2/core/base.hpp
```diff
@@ -151,47 +151,16 @@ enum DecompTypes {
...
- */
+//! norm types
 enum NormTypes { NORM_INF       = 1,
                  NORM_L1        = 2,
                  NORM_L2        = 4,
                  NORM_L2SQR     = 5,
                  NORM_HAMMING   = 6,
                  NORM_HAMMING2  = 7,
-                 NORM_TYPE_MASK = 7,
+                 #ifndef CV_DOXYGEN
```
mshabunin: I suggest removing this wrapper to avoid cluttering, maybe this value can be documented:
```.cpp
NORM_TYPE_MASK = 7, //!< bit-mask which can be used to separate norm type from norm flags
```

modules/core/include/opencv2/core/base.hpp
```diff
@@ -151,47 +151,16 @@ enum DecompTypes {
...
-and for \f$r(0.5) = \begin{pmatrix} 0.5 \\ 0.5 \end{pmatrix}\f$ the calculation is
-\f{align*}
-    \| r(0.5) \|_{L_1} &= |0.5| + |0.5| = 1 \\
-    \| r(0.5) \|_{L_2} &= \sqrt{(0.5)^{2} + (0.5)^{2}} = \sqrt{0.5} \\
-    \| r(0.5) \|_{L_\infty} &= \max(|0.5|,|0.5|) = 0.5.
-\f}
-The following graphic shows all values for the three norm functions \f$\| r(x) \|_{L_1}, \| r(x) \|_{L_2}\f$ and \f$\| r(x) \|_{L_\infty}\f$.
-It is notable that the \f$ L_{1} \f$ norm forms the upper and the \f$ L_{\infty} \f$ norm forms the lower border for the example function \f$ r(x) \f$.
-![Graphs for the different norm functions from the above example](pics/NormTypes_OneArray_1-2-INF.png)
- */
```
mshabunin: Why did you move this documentation to `norm` function? This enumeration is used in many different places and now users will not be able to see it when they jump here from functions other than `norm`. For example [BFMatcher::BFMatcher](http://docs.opencv.org/3.3.0/d3/da1/classcv_1_1BFMatcher.html#abe0bb11749b30d97f60d6ade665617bd).

I even suggest moving each block to corresponding enumeration item.

modules/core/include/opencv2/core/base.hpp
saskatchewancatch: Ok

modules/core/include/opencv2/core/base.hpp
saskatchewancatch: It seemed very specific to `cv::norm`.

My logic was that any other function that takes in `cv::NormTypes` as a parameter could possibly treat usage of `cv::NormTypes` differently and it would be better if individual method's treatment using `cv::NormTypes` parameter could be documented in that function's doc.

I agree with your suggestion though, let's just move it all to `cv::NormTypes`. Moving to corresponding enumeration item makes more sense.

saskatchewancatch:
@mshabunin As you suggested, the descriptions of the enums has been moved from cv::norm to cv::NormTypes.

This kind of layout as you suggested makes things neater.

I have to adjust the descriptions of cv::norm to remove the descriptions of these flags and clean up the grammar there.

sovrasov:
:+1: 
but please fix the remained issues with doc builder.

saskatchewancatch:
That should do it. Are those with merge permissions able to squash this on their end and keep the history of this discussion?

Thanks all for your input.

vpisarev:
now it's all green :+1:

PR #9576, opened 2017-09-07
Import and convert FP16 weights from TensorFlow

dkurt:
### This pullrequest changes
Experimental import of float16 weights from TensorFlow models and convert it into float32.
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/373

### DIFF
--- a/modules/dnn/src/tensorflow/tf_importer.cpp
+++ b/modules/dnn/src/tensorflow/tf_importer.cpp
@@ -63,15 +63,57 @@ void blobShapeFromTensor(const tensorflow::TensorProto &tensor, MatShape& shape)
     {
         const tensorflow::TensorShapeProto &_shape = tensor.tensor_shape();
         int i, n = _shape.dim_size();
-        shape.resize(n);
+        if (n)
+        {
+            shape.resize(n);
 
-        for (i = 0; i < n; i++)
-            shape[i] = (int)_shape.dim(i).size();
+            for (i = 0; i < n; i++)
+                shape[i] = (int)_shape.dim(i).size();
+        }
+        else
+            shape.resize(1, 1);  // Scalar.
     }
     else
     {
         CV_Error(Error::StsError, "Unknown shape of input tensor");
     }
+}
+
+static Mat getTensorContent(const tensorflow::TensorProto &tensor)
+{
+    std::string content = tensor.tensor_content();
+    switch (tensor.dtype())
+    {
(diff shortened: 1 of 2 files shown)


modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -63,17 +63,76 @@ void blobShapeFromTensor(const tensorflow::TensorProto &tensor, MatShape& shape)
...
+
+            Mat res(1, n, CV_32FC1);
+            float* data = (float*)res.data;
+            for (int i = 0; i < n; ++i)
+            {
+                int32_t fp16 = tensor.half_val(i);
+                int32_t sign = fp16 & 0x00008000;
+                int32_t expn = fp16 & 0x00007c00;
+                int32_t frac = fp16 & 0x000003ff;
+                int32_t fp32 = (sign << 16);
```
tomoaki0705: Why not using [ConvertFp16](https://github.com/opencv/opencv/blob/2ac57a2b1fc0b49ddd07f3e3633461fcab5fc3f0/modules/core/src/convert.cpp#L5092) which already converts from/to FP16 ?

dkurt: @tomoaki0705, thank you! I just haven't found it.

modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -63,17 +63,48 @@ void blobShapeFromTensor(const tensorflow::TensorProto &tensor, MatShape& shape)
...
+        case tensorflow::DT_DOUBLE:
+            return Mat(1, content.size() / sizeof(double), CV_64FC1, (void*)content.c_str()).clone();
+        case tensorflow::DT_HALF:
+        {
+            const RepeatedField<int32_t>& field = tensor.half_val();
+            Mat halfs(1, field.size(), CV_32SC1, (void*)field.data());
+            halfs.convertTo(halfs, CV_16UC1);
+
+            Mat values(halfs.size(), CV_16SC1, halfs.data);
+            convertFp16(values, values);
```
alalek: Be careful with pseudo-inplace processing. A lot of functions don't support this well, also there is almost no tests for such functionality.

dkurt: Thanks! Fixed.

vpisarev:
very useful stuff :+1:

PR #9577, opened 2017-09-07
Enabled forEach for const Mats

### LINKED ISSUE #8531: forEach crashes on const Mat&

- OpenCV => 3.2

##### Detailed description

`forEach` throws the following exception when trying to use it on a constant `cv::Mat` reference:

...\opencv\include\opencv2\core\mat.inl.hpp:1041: error: (-215) elemSize() == (((((DataType<_Tp>::type) & ((512 - 1) << 3)) >> 3) + 1) << ((((sizeof(size_t)/4+1)*16384|0x3a50) >> ((DataType<_Tp>::type) & ((1 << 3) - 1))*2) & 3)) in function cv::Mat::at

It works fine when the matrix reference is not constant.

##### Steps to reproduce

```c++
cv::Mat test(512, 512, CV_32FC3);
const cv::Mat& bar = test;

bar.forEach<const cv::Vec3f>([](const cv::Vec3f&, const int*) {

});
```

If you remove the `const` from all three uses (except `const int*`), then the code works fine. The exception is only triggered in debug mode.

mshabunin:
resolves #8531

### This pullrequest changes

After merging PR #9457, the problem described in #8531 become visible at compilation time. Since there were no template specialization for const types (`DataType<const Point3i>` for example), the default values for type and channels has been used causing invalid access to Mat elements.

### DIFF
--- a/modules/core/include/opencv2/core/mat.inl.hpp
+++ b/modules/core/include/opencv2/core/mat.inl.hpp
@@ -1214,7 +1214,7 @@ void Mat::forEach(const Functor& operation) {
 template<typename _Tp, typename Functor> inline
 void Mat::forEach(const Functor& operation) const {
     // call as not const
-    (const_cast<Mat*>(this))->forEach<const _Tp>(operation);
+    (const_cast<Mat*>(this))->forEach<_Tp>(operation);
 }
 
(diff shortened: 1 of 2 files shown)


vpisarev:
:+1:

PR #9579, opened 2017-09-07
OCL: imgproc: don't use doubles to process float data

alalek:
Changed functions: `warpAffine`/`warpPerspective`


vpisarev:
thanks! :+1:

PR #9580, opened 2017-09-07
Fixed some issues found by static analysis

mshabunin:
### This pullrequest changes

Several new issues and several existing with lower severity.

Notes:
- Contains definition change in HAL interface
- Use modern enum values instead of C-definitions in color.cpp

### DIFF
--- a/modules/core/src/array.cpp
+++ b/modules/core/src/array.cpp
@@ -374,7 +374,7 @@ cvGetMatND( const CvArr* arr, CvMatND* matnd, int* coi )
 // returns number of dimensions to iterate.
 /*
 Checks whether <count> arrays have equal type, sizes (mask is optional array
-that needs to have the same size, but 8uC1 or 8sC1 type).
+that needs to have the same size, but 8uC1 or 8sC1 type - feature has been disabled).
 Returns number of dimensions to iterate through:
 0 means that all arrays are continuous,
 1 means that all arrays are vectors of continuous arrays etc.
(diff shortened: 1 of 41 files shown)


modules/dnn/src/dnn.cpp
```diff
@@ -1333,7 +1333,7 @@ struct Net::Impl
 
         //forward parents
         MapIdToLayerData::iterator it;
-        for (it = layers.begin(); it->second.id < ld.id; it++)
+        for (it = layers.begin(); it != layers.end() && (it->second.id < ld.id); it++)
```
alalek: `it++` => `++it`

modules/core/src/array.cpp
```diff
@@ -399,17 +399,16 @@ cvInitNArrayIterator( int count, CvArr** arrs,
     if( !iterator )
         CV_Error( CV_StsNullPtr, "Iterator pointer is NULL" );
 
-    for( i = 0; i <= count; i++ )
+    if (mask)
+        CV_Error( CV_StsBadArg, "Iterator with mask is broken" );
```
alalek: "Iterator with mask is not supported"

modules/dnn/src/torch/THDiskFile.cpp
```diff
@@ -460,8 +453,9 @@ static long THDiskFile_readString(THFile *self, const char *format, char **str_)
   return 0;
 }
 
+using namespace std;
```
alalek: Looks like it is not needed.

modules/dnn/src/torch/THDiskFile.cpp
```diff
@@ -3,6 +3,7 @@
 #include "THGeneral.h"
 #include "THDiskFile.h"
 #include "THFilePrivate.h"
+#include <string>
```
alalek: I believe it should be located in `.hpp` file, because `std::string` is part of external interface.

modules/photo/src/inpaint.cpp
```diff
@@ -79,6 +79,8 @@ CvHeapElem;
 
 class CvPriorityQueueFloat
 {
+    CvPriorityQueueFloat(const CvPriorityQueueFloat & ); // copy disabled
```
alalek: `private:`

alalek:
:+1:

PR #9585, opened 2017-09-07
Adds two new more modern elliptical fit methods.

Jazmann:
The Approximate Mean Square (AMS)  proposed by Taubin 1991, and the Direct least square (Direct) method by Fitzgibbon1999.
To solve the non-symetrical eigenvalue problem an additional eigensystem solver is added but not integrated into the core.
The Eigensystem solver is similar to the one in core/lda.cpp but more stable. The routines in lda.cpp were tried but resulted in system crashes.

### This pullrequest changes

Adds fitEllipseAMS and fitEllipseDirect methods to imgproc

### DIFF
--- a/doc/opencv.bib
+++ b/doc/opencv.bib
@@ -296,73 +296,64 @@
   organization = {BMVA Press}
 }
 @ARTICLE{fitzgibbon1999,
-  abstract	= {This work presents a new efficient method for fitting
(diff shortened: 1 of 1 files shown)


alalek:
Thank you for contribution!

Please remove unrelated `.orig` files from the patch.

PR #9590, opened 2017-09-08
core: fix OpenCL runtime compilation with HAVE_OPENCL_STATIC

alalek:
OpenCV modules should call OpenCL function via `opencv_core` proxy functions.


alalek:
:+1:

PR #9591, opened 2017-09-08
Import and export FP16 Caffe models

dkurt:
### This pullrequest changes

* Function that converts weights of Caffe models to FP16 ([NVidia Caffe](https://github.com/NVIDIA/caffe) format)
* Import converted models into DNN.

Reduction of models is ~x2. Degradation of accuracy is ~1e-2.

GoogLeNet: 52MB &rarr; 26MB
AlexNet: 233MB &rarr; 117MB

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -701,6 +701,19 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
     CV_EXPORTS_W Mat blobFromImages(const std::vector<Mat>& images, double scalefactor=1.0,
                                     Size size = Size(), const Scalar& mean = Scalar(), bool swapRB=true);
 
+    /** @brief Convert all weights of Caffe network to half precision floating point.
+     * @param src Path to origin model from Caffe framework contains single
+     *            precision floating point weights (usually has `.caffemodel` extension).
+     * @param dst Path to destination model with updated weights.
+     *
+     * @note Shrinked model has no origin float32 weights so it can't be used
+     *       in origin Caffe framework anymore. However the structure of data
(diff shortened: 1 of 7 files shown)


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -696,6 +696,18 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+    /** @brief Convert all weights of Caffe network to half precision floating point.
+     * @param src Path to origin model from Caffe framework contains single
+     *            precision floating point weights (usually has `.caffemodel` extension).
+     * @param dst Path to destination model with updated weights.
+     *
+     * @note Shrinked model has no origin float32 weights so it can't be used
+     *       in Caffe framework anymore. That is why use a modified extension
+     *       for the destination model (i.e. `.caffemodel_fp16`).
+     */
+    CV_EXPORTS_W void shrinkNetFromCaffe(const String& src, const String& dst);
```
vpisarev: shrinkCaffeModel would be a better name that resembles `.caffemodel` extension.

modules/dnn/src/caffe/caffe.proto
```diff
@@ -62,6 +62,12 @@ message BlobProto {
   repeated double double_data = 8 [packed = true];
   repeated double double_diff = 9 [packed = true];
 
+  enum DataType {
```
vpisarev: it would be nice to put something like that into official Caffe tree, or maybe find a branch where it's already done

dkurt: NVidia's Caffe supports models stored this way, https://github.com/NVIDIA/caffe/blob/caffe-0.16/src/caffe/proto/caffe.proto#L9. So `shrinkCaffeModel` designed now to store float16 weights in field with the same tag.

vpisarev:
thank you! :+1:

PR #9592, opened 2017-09-08
CUDA: cudafilters: don't modify "const" data

### LINKED ISSUE #9589: Crash in Cuda Laplacian filter
OpenCV => 3.2 and 3.3
Operating System / Platform => Windows 64 Bit
Compiler => Visual Studio 2017 + cuda 8 and cuda 9

When using cv::cuda::createLaplacianFilter(), the only supported "scale" parameter seems to be 1, otherwise a crash occurs.

##### Steps to reproduce
`  try {

    int type = CV_32FC1;//same crash with 8UC1
    cv::Mat src(cv::Size(640, 480), type);
    cv::Mat dst(cv::Size(640, 480), type);
    int dx = 1;
    int dy = 0;
    int k = 1;
    double scale = 2;

    cv::Laplacian(src, dst, dst.depth(), k, scale);
    printf("ok\n");

    cv::cuda::GpuMat srcGpu(cv::Size(640, 480), type);
    cv::cuda::GpuMat dstGpu(cv::Size(640, 480), type);
    cv::cuda::Stream stream;

    cv::Ptr<cv::cuda::Filter> filter = cv::cuda::createLaplacianFilter(srcGpu.type(), dstGpu.type(), k, scale);//won't crash with scale = 1
    filter->apply(srcGpu, dstGpu, stream);//crash in cvtScale_SIMD<float, float, float> for OpenCV 3.2, iwiScale() for OpenCV 3.3
    printf("gpu : ok\n");//not reached

  }
  catch(std::exception& e){
    printf("%s\n", e.what());
  }`

alalek:
resolves #9589

### DIFF
--- a/modules/cudafilters/src/filtering.cpp
+++ b/modules/cudafilters/src/filtering.cpp
@@ -312,9 +312,8 @@ Ptr<Filter> cv::cuda::createLaplacianFilter(int srcType, int dstType, int ksize,
         {2.0f, 0.0f, 2.0f, 0.0f, -8.0f, 0.0f, 2.0f, 0.0f, 2.0f}
     };
 
-    Mat kernel(3, 3, CV_32FC1, (void*)K[ksize == 3]);
-    if (scale != 1)
-        kernel *= scale;
+    Mat kernel1(3, 3, CV_32FC1, (void*)K[ksize == 3]);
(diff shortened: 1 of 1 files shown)


vpisarev:
:+1:

PR #9596, opened 2017-09-08
calib3d: export parameters of CALIB_CB_CLUSTERING

paroj:


modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -781,6 +781,9 @@ struct CV_EXPORTS_W_SIMPLE CirclesGridFinderParameters
       SYMMETRIC_GRID, ASYMMETRIC_GRID
     };
     GridType gridType;
+
+    CV_PROP_RW float squareSize; //!< Distance between two adjacent points. Used by CALIB_CB_CLUSTERING.
```
paroj: increases size of `CirclesGridFinderParameters`

savuor:
The change to `CirclesGridFinderParameters` struct breaks ABI compatibility by adding 2 extra fields.
If there's no way to add these parameters then we should decline this PR

paroj:
ah.. I did not pay attention that there is no ::create() method for the Params struct. We could inherit CirclesGridFinderParameters2 from it to add the parameters though..

savuor:
@paroj, what if someone else decides to add more parameters in the future? Creating 3rd version of structure is not appropriate.
To my mind, using pImpl pattern with non-virtual methods for `CirclesGridFinderParameters2` with get/set methods would work better even though it complicates the structure.

paroj:
wrapped `CirclesGridFinderParameters2` inside `Ptr<>`. This way changing the size of `CirclesGridFinderParameters2` will not break ABI.

There is still an issue with the Java bindings though: They skip `CirclesGridFinderParameters`, but `CirclesGridFinderParameters2` refernces them.

paroj:
`CirclesGridFinderParameters` is blacklisted, so blacklist `CirclesGridFinderParameters2` as well.

vpisarev:
@paroj, can we rename the copy of findCirclesGrid that takes `CirclesGridFinderParameters2` to findCirclesGrid2? And use the parameter structure as-is, not `Ptr<>`. This is more future-proof solution, I believe.

paroj:
@vpisarev I had this behaviour in the first revision, however as @savuor noted:
> what if someone else decides to add more parameters in the future? Creating 3rd version of structure is not appropriate.

therefore I put the structure on the heap using `Ptr<>`. This way the ABI does not change in the above case. (except for inheriting from `CirclesGridFinderParameters2` which is neglectible)

vpisarev:
yes, we discussed it with `savuor` and agreed that `CirclesGridFinderParameters` is not as fast evolving as, say, VideoCapture, and thus a separate property setter/getter would be an overkill. So I'd suggest to get back to the previous solution with the API like:

<pre>
CV_EXPORTS_W bool findCirclesGrid2( InputArray image, Size patternSize,
                                   OutputArray centers, int flags,
                                   const Ptr<FeatureDetector> &blobDetector,
                                   CirclesGridFinderParameters2 parameters);
</pre>

and we will merge it in.

alalek:
It is better to use const reference for input parameters:
```.diff
-CirclesGridFinderParameters2 parameters
+const CirclesGridFinderParameters2& parameters
```

Old `CirclesGridFinderParameters parameters` would not be changed for now (but it should).

paroj:
> It is better to use const reference for input parameters:

the struct is modified inside findCirclesGrid2 (overlapping options with flags) - therefore I would leave is is.

savuor:
:+1:

PR #9604, opened 2017-09-11
OCL: support kernel profiling

alalek:


modules/core/include/opencv2/core/ocl.hpp
```diff
@@ -331,8 +331,12 @@ class CV_EXPORTS Queue
     void* ptr() const;
     static Queue& getDefault();
 
+    /// @brief Returns OpenCL command queue with enable profiling mode support
+    const Queue& getProfilingQueue() const;
+
+    struct Impl; friend struct Impl;
```
vpisarev: is this part really necessary? looks like we break encapsulation, but usually we find a way to keep it

alalek: This is still inaccessible via public API. Alternate way is to extend public API with new "private" methods that I want to avoid via this construction.

vpisarev:
:+1:

PR #9618, opened 2017-09-13
Added gradiantSize param into goodFeaturesToTrack API

vipinanand4:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/383

Removed hardcode value 3 in goodFeaturesToTrack API, and
added new param 'gradinatSize' in this API so that user can
pass any gradiant size as 3, 5 or 7.

Signed-off-by: Vipin Anand <anand.vipin@gmail.com>
Signed-off-by: Nilaykumar Patel<nilay.nilpat@gmail.com>
Signed-off-by: Prashanth Voora <prashanthx85@gmail.com>

### This pullrequest changes

This pull request removes hardcoding of gradiantSize and adds gradiantSize as parameter in goodFeaturesToTrack API to make it compliant with ShiToamsi and Harris Corners detector specification to support gradiant window from application.


modules/features2d/include/opencv2/features2d.hpp
```diff
@@ -516,7 +516,7 @@ class CV_EXPORTS_W GFTTDetector : public Feature2D
 {
 public:
     CV_WRAP static Ptr<GFTTDetector> create( int maxCorners=1000, double qualityLevel=0.01, double minDistance=1,
-                                             int blockSize=3, bool useHarrisDetector=false, double k=0.04 );
+                                             int blockSize=3, int gradiantSize=3, bool useHarrisDetector=false, double k=0.04 );
```
sovrasov: We can't change the public API due to binary compatibility obligations. You can create an overloaded version of the function.

modules/features2d/include/opencv2/features2d.hpp
```diff
@@ -529,6 +529,9 @@ class CV_EXPORTS_W GFTTDetector : public Feature2D
     CV_WRAP virtual void setBlockSize(int blockSize) = 0;
     CV_WRAP virtual int getBlockSize() const = 0;
 
+    CV_WRAP virtual void setGradiantSize(int gradiantSize) = 0;
```
sovrasov: New virtual methods also break binary compatibility. You can declare them as non-virtual and implicitly cast `this` pointer to `GFTTDetector_Impl` to call the real implementation.

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -1904,7 +1904,8 @@ or cornerMinEigenVal.
 CV_EXPORTS_W void goodFeaturesToTrack( InputArray image, OutputArray corners,
                                      int maxCorners, double qualityLevel, double minDistance,
                                      InputArray mask = noArray(), int blockSize = 3,
-                                     bool useHarrisDetector = false, double k = 0.04 );
```
sovrasov: The same thing with this function: we need to keep an overload with the original signature.

modules/imgproc/include/opencv2/imgproc/imgproc_c.h
```diff
@@ -934,6 +934,7 @@ CVAPI(void)  cvGoodFeaturesToTrack( const CvArr* image, CvArr* eig_image,
                                     double  min_distance,
                                     const CvArr* mask CV_DEFAULT(NULL),
                                     int block_size CV_DEFAULT(3),
+                                    int gradiant_size CV_DEFAULT(3),
```
sovrasov: C API deprecated, so it's better to leave it as is.

savuor:
To pass all the checks, you should fix the following:
1. Do not remove functions with old signatures because it breaks API and ABI compliance. For example, you should left `Ptr<GFTTDetector> GFTTDetector::create( maxCorners, qualityLevel, minDistance, blockSize, useHarrisDetector, k)` and call the basic function with `gradient` argument.
2. In `featureselect.cpp` you should fix `ocl_goodFeaturesToTrack()`: its calls to `cornerHarris()` and `cornerSubPix()` do not use `gradient` argument.
2. Perf tests look good and work on my PC but we need to find out what's wrong happened with them.
3. I recommend you to check spelling: rename `gradiant` to `gradient` everywhere.

modules/imgproc/include/opencv2/imgproc/imgproc_c.h
savuor: Moreover, it breaks ABI compatibility, so we don't need this

modules/features2d/include/opencv2/features2d.hpp
```diff
@@ -529,11 +531,16 @@ class CV_EXPORTS_W GFTTDetector : public Feature2D
...
+    CV_WRAP virtual void setGradientSize(int gradientSize);
+    CV_WRAP virtual int getGradientSize();
+
     CV_WRAP virtual void setHarrisDetector(bool val) = 0;
     CV_WRAP virtual bool getHarrisDetector() const = 0;
 
     CV_WRAP virtual void setK(double k) = 0;
     CV_WRAP virtual double getK() const = 0;
+protected:
+    int gradientSize;
```
savuor: This breaks ABI again.
You need to add this field to `GFTTDetector_Impl`, not to `GFTTDetector`

modules/features2d/include/opencv2/features2d.hpp
```diff
@@ -529,11 +531,16 @@ class CV_EXPORTS_W GFTTDetector : public Feature2D
     CV_WRAP virtual void setBlockSize(int blockSize) = 0;
     CV_WRAP virtual int getBlockSize() const = 0;
 
+    CV_WRAP virtual void setGradientSize(int gradientSize);
+    CV_WRAP virtual int getGradientSize();
```
savuor: As it's said before, new virtual methods break ABI.
You need to fix this as it's described [earlier](https://github.com/opencv/opencv/pull/9618#discussion_r138858648)

modules/imgproc/perf/perf_goodFeaturesToTrack.cpp
```diff
@@ -6,15 +6,16 @@ using namespace perf;
 using std::tr1::make_tuple;
 using std::tr1::get;
 
-typedef std::tr1::tuple<string, int, double, int, bool> Image_MaxCorners_QualityLevel_MinDistance_BlockSize_UseHarris_t;
-typedef perf::TestBaseWithParam<Image_MaxCorners_QualityLevel_MinDistance_BlockSize_UseHarris_t> Image_MaxCorners_QualityLevel_MinDistance_BlockSize_UseHarris;
+typedef std::tr1::tuple<string, int, double, int, int, bool> Image_MaxCorners_QualityLevel_MinDistance_BlockSize_gradientSize_UseHarris_t;
+typedef perf::TestBaseWithParam<Image_MaxCorners_QualityLevel_MinDistance_BlockSize_gradientSize_UseHarris_t> Image_MaxCorners_QualityLevel_MinDistance_BlockSize_gradientSize_UseHarris;
 
-PERF_TEST_P(Image_MaxCorners_QualityLevel_MinDistance_BlockSize_UseHarris, goodFeaturesToTrack,
+PERF_TEST_P(Image_MaxCorners_QualityLevel_MinDistance_BlockSize_gradientSize_UseHarris, goodFeaturesToTrack,
```
savuor: You need to update sanity perf data as well: the name of test case was changed, that's why perf tests are failed.
There's need to change `testdata/perf/imgproc.xml` file. [Here](https://github.com/opencv/opencv/wiki/HowToUsePerfTests#how-to-update-perf-data) is the description of how to do that or you can change it manually.
Then you need to create a branch with updated file in your `opencv_extra` fork with the same name as current branch has (i.e. `goodFeaturesToTrack_added_gradiantSize`) and provide a PR to opencv_extra project.

savuor:
:+1:

PR #9623, opened 2017-09-13
Disable SSD on CPU Halide test

dkurt:
### This pullrequest changes

Segmentation fault appears from time to time at SSD network compilation (CPU target only, mbox_conf concat layer with `1 754 328` output channels and `1x1` `width x height`).


vpisarev:
:+1:

PR #9626, opened 2017-09-14
fisheye: fix initUndistortRectifyMap()

T0ny0:
### This pullrequest changes

There is a bug in `cv::fisheye::initUndistortRectifyMap()` when you undistort fish-eye image with such `RectifyMap` that part of fisheye remains behind new camera view (`_w <= 0` in code) undistortion work incorrectly (see the photos). Also there is division by `_w` with no zero check. The fixed `initUndistortRectifyMap()` works fine in such situations.

bad

good

### DIFF
--- a/modules/calib3d/src/fisheye.cpp
+++ b/modules/calib3d/src/fisheye.cpp
@@ -475,17 +475,26 @@ void cv::fisheye::initUndistortRectifyMap( InputArray K, InputArray D, InputArra
 
         for( int j = 0; j < size.width; ++j)
         {
-            double x = _x/_w, y = _y/_w;
+            double u, v;
+            if( _w <= 0)
+            {
+                _x > 0 ? u = -std::numeric_limits<double>::infinity() : u = std::numeric_limits<double>::infinity();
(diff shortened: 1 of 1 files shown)


sovrasov:
:+1:

modules/calib3d/src/fisheye.cpp
```diff
@@ -475,17 +475,26 @@ void cv::fisheye::initUndistortRectifyMap( InputArray K, InputArray D, InputArra
 
         for( int j = 0; j < size.width; ++j)
         {
-            double x = _x/_w, y = _y/_w;
+            double u, v;
+            if( _w <= 0)
+            {
+                _x > 0 ? u = -std::numeric_limits<double>::infinity() : u = std::numeric_limits<double>::infinity();
+                _y > 0 ? v = -std::numeric_limits<double>::infinity() : v = std::numeric_limits<double>::infinity();
```
alalek: Static analyzers doesn't like assignments in expressions. Could your use this form instead:
```
u = (_x > 0) ? -std::numeric_limits<double>::infinity() : std::numeric_limits<double>::infinity();
v = (_y > 0) ? ... : ...;
```

T0ny0: Done

PR #9627, opened 2017-09-14
async bugfix

dtmoodie:
### This pullrequest changes

I ran into a problem where if the input GPU stream was sufficiently ahead of the CPU, then the upload would occur late enough that the local variables in loadConstants would be lost.  Thus this passes in pointers to members of PyrLKBase which will not go out of scope.

### DIFF
--- a/modules/cudaoptflow/src/pyrlk.cpp
+++ b/modules/cudaoptflow/src/pyrlk.cpp
@@ -103,9 +103,12 @@ namespace
     };
 
     PyrLKOpticalFlowBase::PyrLKOpticalFlowBase(Size winSize, int maxLevel, int iters, bool useInitialFlow) :
-        winSize_({winSize.width, winSize.height}), halfWinSize_({(winSize.width - 1) / 2, (winSize.height - 1) / 2}),
         maxLevel_(maxLevel), iters_(iters), useInitialFlow_(useInitialFlow)
     {
+        winSize_[0] = winSize.width;
+        winSize_[1] = winSize.height;
+        halfWinSize_[0] = (winSize.width - 1) / 2;
+        halfWinSize_[1] = (winSize.height - 1) / 2;
         pyrlk::loadWinSize(winSize_, halfWinSize_, 0);
         pyrlk::loadIters(&iters_, 0);
     }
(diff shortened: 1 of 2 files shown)


modules/cudaoptflow/src/pyrlk.cpp
```diff
@@ -88,7 +90,8 @@ namespace
         void dense(const GpuMat& prevImg, const GpuMat& nextImg, GpuMat& u, GpuMat& v, Stream& stream);
 
     protected:
-        Size winSize_;
+        int winSize_[2];
```
alalek: This is a bad practice, please avoid this.

modules/cudaoptflow/src/pyrlk.cpp
```diff
@@ -339,8 +354,13 @@ namespace
...
         }
 
-        virtual Size getWinSize() const { return winSize_; }
-        virtual void setWinSize(Size winSize) { winSize_ = winSize; }
+        virtual Size getWinSize() const { return cv::Size(winSize_[0], winSize_[1]); }
+        virtual void setWinSize(Size winSize) {
+            winSize_[0] = winSize.width;
+            winSize_[1] = winSize.height;
+            halfWinSize_[0] = (winSize.width - 1) / 2;
+            halfWinSize_[1] = (winSize.height -1) / 2;
```
alalek: `halfWinSize_` is updated in three locations, but used in `loadWinSize` only.
Consider to drop this field completely and calculate it on-the-fly.

dtmoodie:
The issue that this pull request fixes is that if the cv::cuda::Stream that is passed into calc is sufficiently loaded that the upload of the constants does not occur until after the function finishes; then the upload can upload garbage because the local variables for winSize and halfWinSize have gone out of scope.  Thus it was necessary to make the upload in loadWinSize access pointers to members inside the optical flow object.  This is at least the reasoning behind adding a halfWinSize member variable.

vpisarev:
@dtmoodie, ok. Could you please fix the compile warnings reported by our CUDA build slave: http://pullrequest.opencv.org/#/summary/opencv

vpisarev:
great, looks like the PR passes all the tests! :+1:

PR #9633, opened 2017-09-15
i9629 - Added actual documentation for cv::PSNR function

saskatchewancatch:
resolves #9629 

### This pullrequest changes

Added meaningful doc for cv::PSNR. Makes clear the restrictions on input images.

PR #9642, opened 2017-09-16
Add some useful input assertions to PSNR function

saskatchewancatch:
Input arrays must be depth CV_8U and of identical size.

Intended this to be part of PR #9633 but that PR got merged and closed.

Original assertion was insufficient.

Now asserts that both input arrays are depth CV_8U and have same sizes.

### DIFF
--- a/modules/core/src/stat.cpp
+++ b/modules/core/src/stat.cpp
@@ -4150,7 +4150,10 @@ double cv::PSNR(InputArray _src1, InputArray _src2)
 {
     CV_INSTRUMENT_REGION()
 
-    CV_Assert( _src1.depth() == CV_8U );
+    //Input arrays must have depth CV_8U and be of identical size
+    CV_Assert( _src1.depth() == CV_8U && _src2.depth() == CV_8U );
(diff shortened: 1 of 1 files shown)


modules/core/src/stat.cpp
```diff
@@ -4150,7 +4150,10 @@ double cv::PSNR(InputArray _src1, InputArray _src2)
 {
     CV_INSTRUMENT_REGION()
 
-    CV_Assert( _src1.depth() == CV_8U );
+    //Input arrays must have depth CV_8U and be of identical size
+    CV_Assert( _src1.depth() == CV_8U && _src2.depth() == CV_8U );
+    CV_Assert( _src1.rows() == _src2.rows() && _src1.cols() == _src2.cols() && _src1.dims() == _src2.dims() );
```
vpisarev: There is no need in the second assert. The function calls cv::norm() which does the same check. When writing high-level functions that are based on lower-level functions, we should rely on the checks that those lower-level functions do, otherwise the code will be flooded with duplicate assert's. The first assert still makes sense because the normalization in PSNR assumes that the arrays are 8-bit.

saskatchewancatch:
Thanks for catching that @vpisarev

vpisarev:
:+1:

PR #9644, opened 2017-09-16
imgproc: disable IPP gaussian blur call with inplace arguments

### LINKED ISSUE #9628: IwException from KAZE

- OpenCV => 3.3.0
- Operating System / Platform => Windows 10 Pro, 64 Bit

##### Detailed description

I'm seeing exceptions thrown by Intel's IPP while doing a feature detection using a KAZE algorithm object's detect() call.  The exception occurs while executing \modules\features2d\src\kaze\KAZEFeatures.cpp's KAZEFeatures::Create_Nonlinear_Scale_Space(), where it starts building Gaussian blurs for its scale space, specifically this line (currently line 104 in rev 73c3d14):

` gaussian_2D_convolution(evolution_[0].Lt, evolution_[0].Lt, 0, 0, options_.soffset);`

This makes its way down into the imgproc module and then IPP where iwiFilterGaussian() returns an IppStatus of 'ippStsInplaceModeNotSupportedErr (-9997)' which causes it to throw an IwException w/that value.  I *believe* that status is iwiFilterGaussian() signaling that its pSrcImage and pDstImage inputs can't be the same (for an 'in-place' blur in pSrcImage).

My KAZE output appears reasonable despite this--I'm getting keypoints, and they look good--I'm just fearful that one or more failed blurs could be introducing unnecessary noise and keeping the algorithm from its full potential.

Here's everything I can find in my cmake setup related to IPP...most/all should be at default:

BUILD\BUILD_IPP_IW=true

BUILD\BUILD_WITH_DYNAMIC_IPP=false

##### Steps to reproduce

Run the tutorial at:

https://github.com/opencv/opencv/blob/master/samples/cpp/tutorial_code/features2D/AKAZE_match.cpp

Note that no exceptions occur.  Next replace usage of AKAZE with KAZE at line 25:

`	Ptr<KAZE> akaze = KAZE::create();`

Build/run, observe the exception (5 occurrences of the following):

` Exception thrown at (address) in opencv_test.exe: Microsoft C++ exception: ipp::IwException at memory location (address).`

Note this is related to the question I posted at:

http://answers.opencv.org/question/173245/iwexception-from-kaze/

sovrasov:
resolves #9628

### DIFF
--- a/modules/imgproc/src/smooth.cpp
+++ b/modules/imgproc/src/smooth.cpp
@@ -2283,6 +2283,8 @@ static bool ipp_GaussianBlur(InputArray _src, OutputArray _dst, Size ksize,
     {
         Mat src = _src.getMat();
         Mat dst = _dst.getMat();
+        if(src.data == dst.data)
+            return false;
         ::ipp::IwiImage       iwSrc      = ippiGetImage(src);
         ::ipp::IwiImage       iwDst      = ippiGetImage(dst);
         ::ipp::IwiBorderSize  borderSize = ::ipp::iwiSizeToBorderSize(ippiGetSize(ksize));


modules/imgproc/src/smooth.cpp
```diff
@@ -2360,7 +2360,7 @@ void cv::GaussianBlur( InputArray _src, OutputArray _dst, Size ksize,
                _src.rows() > ksize.height && _src.cols() > ksize.width);
     (void)useOpenCL;
 
-    CV_IPP_RUN(!useOpenCL, ipp_GaussianBlur( _src,  _dst,  ksize, sigma1,  sigma2, borderType));
+    CV_IPP_RUN(!useOpenCL && _src.getObj() != _dst.getObj(), ipp_GaussianBlur( _src,  _dst,  ksize, sigma1,  sigma2, borderType));
```
alalek: This condition is not accurate.
We should compare values of `cv::Mat::data` fields (in `ipp_GaussianBlur` function to hide implementation details).
```
Mat a = ...
Mat b = a;
GaussianBlur(a, b , ...); // .getObj() is different, but `.data` is the same.
```

Perhaps OpenCL path may not work properly for inplace calls too.

alalek:
@sovrasov Are you able to reproduce original problem? IPP calls has try/catch guards, so it is very strange.

sovrasov:
I've tried to reproduce it on windows with the latest `master`, but without success.

DenisPryt:
Why this fix wasn't merged? I have same problem with `KAZEFeatures`. Inside `KAZEFeatures::Create_Nonlinear_Scale_Space` method exist code, that call `gaussian_2D_convolution` with equals src and dst Mats. As result `::ipp::iwiFilterGaussian` method throw `::ipp::IwException` every time.

PR #9649, opened 2017-09-18
Remove data permutation from Reshape layer

dkurt:
### This pullrequest changes

Removed `reorder_dims` attribute from Reshape layer. This flag is `true` for TensorFlow models but [tf.reshape](https://www.tensorflow.org/versions/master/api_docs/python/tf/reshape) doesn't make a permutation. However [tf.transpose](https://www.tensorflow.org/versions/master/api_docs/python/tf/transpose) does.

Import of a `Transpose` op is better to do with a some real network.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/377

### DIFF
--- a/modules/dnn/src/layers/permute_layer.cpp
+++ b/modules/dnn/src/layers/permute_layer.cpp
@@ -132,6 +132,7 @@ public:
 
         for (size_t i = 0; i < inputs.size(); i++)
         {
+            CV_Assert(inputs[i].size() == 4);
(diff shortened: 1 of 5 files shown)


vpisarev:
thanks! :+1:

vpisarev:
@dkurt, looks like there is merge conflict. Could you please fix it?

dkurt:
@vpisarev, thank you! Resolved.

PR #9650, opened 2017-09-18
Adds two new more modern elliptical fit methods.

Jazmann:
### This pullrequest changes

Adds fitEllipseAMS to imgproc: The Approximate Mean Square (AMS) proposed by Taubin 1991.
Adds fitEllipseDirect to imgproc: The Direct least square (Direct) method by Fitzgibbon1999.

New Tests are included for the methods.
fitEllipseAMS Tests
fitEllipseDirect Tests

Comparative examples are added to fitEllipse.cpp in Samples.

To solve the non-symetrical eigenvalue problem an additional eigensystem solver is added but not integrated into the core.
The Eigensystem solver is similar to the one in core/lda.cpp but more stable. The routines in lda.cpp were tried but resulted in system crashes.


Jazmann:
Good to go? I did wonder about integrating with fitEllipse via a 'method' specifier but thought I'd keep it simple for now. If you want me to do this I can but I would appreciate a name suggestion for the original Fitzgibbon method.

modules/imgproc/src/shapedescr.cpp
```diff
@@ -454,6 +455,369 @@ cv::RotatedRect cv::fitEllipse( InputArray _points )
...
+    double *Ad = _Ad;
+    Mat A( n, 6, CV_64F, Ad );
+
+    AutoBuffer<double> _DMd(6*6);
+    double *DMd = _DMd;
+    Mat DM( 6, 6, CV_64F, DMd );
+
+    AutoBuffer<double> _Md(3*3);
+    double  *Md = _Md;
+    Mat M( 3, 3, CV_64F, Md );
```
alalek: Consider to use `Matx33d` in case of fixed dimensions.

For 6x6 matrix use this: `Matx<double, 6, 6>`

Also replace access expressions:
```.diff
-Md[1*3+0]
+M(1, 0)
```

modules/imgproc/src/shapedescr.cpp
```diff
@@ -454,6 +455,369 @@ cv::RotatedRect cv::fitEllipse( InputArray _points )
...
+    Md[1*3+2] = -DMd[1*6+2] - (DMd[1*6+3]*TMd[0*3+2] + DMd[1*6+4]*TMd[1*3+2] + DMd[1*6+5]*TMd[2*3+2])/Ts;
+    Md[2*3+0] = (DMd[0*6+0] + (DMd[0*6+3]*TMd[0*3+0] + DMd[0*6+4]*TMd[1*3+0] + DMd[0*6+5]*TMd[2*3+0])/Ts)/2.;
+    Md[2*3+1] = (DMd[0*6+1] + (DMd[0*6+3]*TMd[0*3+1] + DMd[0*6+4]*TMd[1*3+1] + DMd[0*6+5]*TMd[2*3+1])/Ts)/2.;
+    Md[2*3+2] = (DMd[0*6+2] + (DMd[0*6+3]*TMd[0*3+2] + DMd[0*6+4]*TMd[1*3+2] + DMd[0*6+5]*TMd[2*3+2])/Ts)/2.;
+
+    if (fabs(cv::determinant(M)) > 1.0e-10) {
+        Mat eVal, eVec;
+        EigenvalueDecomposition::eigensystem(M, eVal, eVec);
+
+    // Select the eigen vector {a,b,c} which satisfies 4ac-b^2 > 0
```
alalek: please fix code indentation to keep code readable

modules/imgproc/src/shapedescr.cpp
```diff
@@ -454,6 +455,369 @@ cv::RotatedRect cv::fitEllipse( InputArray _points )
...
+    double cond[3];
+    cond[0]=(4.0 * eVec.at<double>(0,0) * eVec.at<double>(2,0) - eVec.at<double>(1,0) * eVec.at<double>(1,0));
+    cond[1]=(4.0 * eVec.at<double>(0,1) * eVec.at<double>(2,1) - eVec.at<double>(1,1) * eVec.at<double>(1,1));
+    cond[2]=(4.0 * eVec.at<double>(0,2) * eVec.at<double>(2,2) - eVec.at<double>(1,2) * eVec.at<double>(1,2));
+    if (cond[0]<cond[1]) {
+        (cond[1]<cond[2]) ? i=2 : i=1;
```
alalek: Code static analyzers dislike assignments in expressions. Please use this form instead:
```
i = (cond[1]<cond[2]) ? 2 : 1;
```

modules/core/include/opencv2/core/utils/logger.hpp
```diff
@@ -16,6 +16,7 @@
 //
 //! @{
 
+namespace cv {
 namespace utils {
 namespace logging {
```
alalek: Good catch!

samples/cpp/fitellipse.cpp
```diff
@@ -56,28 +96,54 @@ int main( int argc, char** argv )
...
+
+    RotatedRect box, boxAMS, boxDirect;
     vector<vector<Point> > contours;
     Mat bimage = image >= sliderPos;
 
     findContours(bimage, contours, RETR_LIST, CHAIN_APPROX_NONE);
 
     Mat cimage = Mat::zeros(bimage.size(), CV_8UC3);
 
+    Size textsize1 = getTextSize("openCV", FONT_HERSHEY_COMPLEX, 1, 1, 0);
```
alalek: "OpenCV" here and below.

modules/core/include/opencv2/core/utils/logger.hpp
Jazmann: I'm surprised that one hasn't caused problems for other peoples builds.

Jazmann:
Is there anything in that pull request that you would like me to look at in particular? It implements one of the two methods I implemented but the code is significantly different. I implemented both methods using a similar structure to the method in the two papers and it has proved to be both  robust and quick in my project.

modules/imgproc/src/shapedescr.cpp
alalek: `AutoBuffer` and following pointer should be removed too (they are not used).

modules/imgproc/src/shapedescr.cpp
```diff
@@ -454,6 +455,361 @@ cv::RotatedRect cv::fitEllipse( InputArray _points )
...
+        } else {
+            box.angle = (float)(fmod(theta*180/CV_PI,180.0));
+        };
+
+
+    } else {
+        box = cv::fitEllipseDirect( points );
+
+    }    } else {
+        box = cv::fitEllipse( points );
```
alalek: There is something wrong with indentation

Jazmann:
@alalek should I update fitEllipse to remove pointer reference [] syntax also?

vpisarev:
@Jazmann, thank you for the contribution! Could you please make one parameterised test out of two - they look almost identical except for the different algorithms used? Also, could you please remove the test cases from the sample. The sample's main purpose is to serve an illustration - how to use a particular function, not be a test

Jazmann:
The tests were generated using Mathematica. The points are the same but the expected output is different for each algorithm. The 1% difference allowed in the fit is because Mathematica's numerics extend beyond double precision. 
Can I retain 3 of the test output routines in the example and add commentary regarding the difference between the three routines. This will help people choose between the three. It is difficult to see the difference on the processed image because there are alot of ellipses and points.

modules/imgproc/src/shapedescr.cpp
```diff
@@ -454,6 +455,362 @@ cv::RotatedRect cv::fitEllipse( InputArray _points )
...
+    const Point* ptsi = points.ptr<Point>();
+    const Point2f* ptsf = points.ptr<Point2f>();
+
+    AutoBuffer<double> _Ad(n*6);
+    double *Ad = _Ad;
+    Mat A( n, 6, CV_64F, Ad );
+
+    AutoBuffer<double> _DMd(6*6);
+    double *DMd = _DMd;
+    Matx<double, 6, 6> DM = Matx<double, 6, 6>(DMd);
```
alalek: Replace these 3 lines to this one:
```
Matx<double, 6, 6> DM;
```

Apply similar changes to other "fixed size" cases.

vpisarev:
@Jazmann, I understand. But even though the tests have been autogenerated, we can still refactor them and make them more maintainable. Also, the example is too extensive in my opinion; if you want to show the difference between algorithms, you may find some picture (instead of stuff.jpg) that shows such a difference. Big manually specified point sets are good for the tests but they make the example look ugly.

Jazmann:
OK, I'm working on finding an image which will demonstrate the point that the AMS method is better for finding partially obscured ellipses but that the Direct method is more reliable. 
When I get the time I'll do the maths to see if the Original method is worth keeping.

Jazmann:
Thinking about it, the current fall through (AMS->Direct->Original) is optimal with current methods. Do you want me to consolidate the three in a fitEllipse method moving the current fitEllipse to fitEllipseOrig or similar?

vpisarev:
@Jazmann, for now we can have separate functions for different algorithms. In order to merge PR in, I'd like to have the tests consolidated and the sample simplified. Subsequent PRs can improve API and implementation, we have a plenty of time (till December) to do it.

Jazmann:
Please help! I don't understand the build bot error.

alalek:
Buildbot blames on patch size.
1. Image is too large (800x600 color image). Please reduce image size and convert image to grayscale (colors are not needed for threshold+findContours).
2. Squash your commits into one to eliminate unused changes/data from patch history.

alalek:
To squash commits into one you could use something like this:
```
git rebase -i HEAD~10    # 10 here is the number of pending commits (see this PR header)
# 'pick' your first commit and 'squash' others commits below your first.
# save & exit
git push --force origin ellipseFit
```

More info/guides:
https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

Jazmann:
Thanks alalek. rebased with all the image commits squashed

Jazmann:
Build bot is not building the code now. Is it because of the rebase?

modules/core/include/opencv2/core/eigensystem.hpp
```diff
@@ -0,0 +1,674 @@
...
+  }
+
+  // Nonsymmetric reduction from Hessenberg to real Schur form.
+
+  void hqr2() {
+
+    //  This is derived from the Algol procedure hqr2,
+    //  by Martin and Wilkinson, Handbook for Auto. Comp.,
+    //  Vol.ii-Linear Algebra, and the corresponding
+    //  Fortran subroutine in EISPACK.
```
alalek: Please avoid duplication of code and reuse/fix existed one:
https://github.com/opencv/opencv/blob/3.3.0/modules/core/src/lda.cpp#L319-L322

Jazmann: Is there an appropriate include statement to access the functions in ida.cpp?

Jazmann:
So I think the Windows build needs the methods made public in the main header. I refactored because I am building on a mac, for which everything was fine, and I needed to use build bot to find the problem.

Jazmann:
Anything else needed guys?

vpisarev:
@Jazmann, there is one more request to you before we merge the patch in. We've added a dedicated function to solve non-symmetric eigenvalue problem, see https://github.com/opencv/opencv/pull/9752. Could you please refactor your patch - remove the duplicated code and use `cv::eigenNonSymmetric()`?

samples/cpp/fitellipse.cpp
```diff
@@ -71,13 +223,36 @@ int main( int argc, char** argv )
 // draw it and approximate it by ellipses.
 void processImage(int /*h*/, void*)
 {
+
+    RotatedRect box, boxAMS, boxDirect;
     vector<vector<Point> > contours;
     Mat bimage = image >= sliderPos;
 
-    findContours(bimage, contours, RETR_LIST, CHAIN_APPROX_NONE);
+    findContours(bimage, contours, CV_RETR_EXTERNAL, CHAIN_APPROX_NONE);
```
alalek: `CV_` macros are deprecated (legacy C-API).

Jazmann:
Ok I've refactored and created a new pull request. I guess this request can be closed now?

Jazmann:
Nice work on eigenNonSymmetric BTW

PR #7308, opened 2016-09-18
fixed unnecessary reconstruction in seamlessclone

stalin18:
resolves [#5045](https://github.com/opencv/opencv/issues/5045)
### This pullrequest changes

Fixed unnecessary reconstruction in seamlessClone() by passing only the rectangular ROI area of input images to Cloning::normalClone().

Earlier, Cloning::normalClone() was called with full-size images; now destinatinoROI, sourceROI, maskROI and recoveredROI basically denote the rectangular ROI regions in destination, source, mask and output image.

PR #9654, opened 2017-09-18
Seamless clone: performance fixed

### LINKED ISSUE #5045: seamlessClone takes time proportional to the size of the image
Transferred from http://code.opencv.org/issues/4445

```
|| David Doria on 2015-06-27 23:31
|| Priority: Normal
|| Category: None
|| Tracker: Bug
|| Difficulty: 
|| PR: 
|| Platform: x64 / Linux
```
## seamlessClone takes time proportional to the size of the image

```
I've found that with a fixed hole size, when the size of the image increases the runtime increases substantially. For example, if I fill a 200x200 hole in a 500x500 image, it takes < 1 second. Then if I fill a 200x200 hole in a 3000x3000 image, it takes 20 seconds. The 500x500 was a cropped version of the 3000x3000 so that the hole and source could be identical for testing to make sure it wasn't a function of the data in the hole, etc.

My guess is that some processing is happening (derivatives being computed, etc.) over the whole image, even though they are only needed inside the source or hole region.
```
##### Philipp Hasper on 2015-07-02 07:47

```
Thanks for reporting. Could you add a minimal source code example?
```

savuor:
resolves #5045 

The patch is taken from PR #7308.
To pass tests, reference images were updated.

### DIFF
--- a/modules/photo/src/seamless_cloning.cpp
+++ b/modules/photo/src/seamless_cloning.cpp
@@ -56,15 +56,13 @@ void cv::seamlessClone(InputArray _src, InputArray _dst, InputArray _mask, Point
     const Mat mask = _mask.getMat();
     _blend.create(dest.size(), CV_8UC3);
     Mat blend = _blend.getMat();
+    dest.copyTo(blend);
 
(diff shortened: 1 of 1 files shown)


savuor:
:+1:

Issue #6496, opened 2016-05-02
Please consider supporting RelWithDebInfo


bjornpiltz:
I don't see the purpose of the following line in the top level CMake script:
`set(CMAKE_CONFIGURATION_TYPES "Debug;Release" CACHE STRING "Configs" FORCE)`
What are the reasons for disabling RelWithDebInfo?

PkLab:
I would see available also `MinSizeRel`

mshabunin:
These restrictions were made to explicitly limit the number of supported build configurations. If you need to build a custom variant of OpenCV with advanced parameters, you can patch this file locally.

PkLab:
It was to be aligned to `CMAKE_C[XX]_FLAGS_[DEBUG|RELEASE|RELWITHDEBINFO|MINSIZEREL]`

PR #9656, opened 2017-09-18
cmake: map RelWithDebInfo and MinSizeRel configurations to Release

### LINKED ISSUE #5564: buildPyramid crashes on VS 2013 when running in RelWithDebInfo mode.
When I call buildPyramid with an image as source and an empty std::vector as destination, I get the following exception (only in RelWithDebInfo): 

> Unhandled exception at 0x00007FFF1A97512D (msvcp120d.dll) in hole_filling.exe: 0xC0000005: Access violation reading location 0xFFFFFFFFFFFFFFFF.

In other modes (Debug, Release), everything works as excepted. This is the stack trace I get:

```
msvcp120d.dll!00007fff1a97512d()    Unknown
opencv_core300d.dll!std::vector<cv::Mat,std::allocator<cv::Mat> >::_Reallocate(unsigned __int64 _Count) Line 1606   C++
opencv_core300d.dll!std::vector<cv::Mat,std::allocator<cv::Mat> >::_Reserve(unsigned __int64 _Count) Line 1619  C++
opencv_core300d.dll!std::vector<cv::Mat,std::allocator<cv::Mat> >::resize(unsigned __int64 _Newsize) Line 1106  C++
opencv_core300d.dll!cv::_OutputArray::create(int d, const int * sizes, int mtype, int i, bool allowTransposed, int fixedDepthMask) Line 2387    C++
opencv_core300d.dll!cv::_OutputArray::create(int _rows, int _cols, int mtype, int i, bool allowTransposed, int fixedDepthMask) Line 2202    C++
opencv_imgproc300d.dll!cv::buildPyramid(const cv::_InputArray & _src, const cv::_OutputArray & _dst, int maxlevel, int borderType) Line 1507    C++
```

bblanchon:
resolves #5564

This Pull Request allows a user to compile her project with configurations RelWithDebInfo and MinSizeRel.
It was previously not possible as CMake was incorrectly linking with the Debug version of the library when using RelWithDebInfo  or MinSizeRel.

See https://github.com/opencv/opencv/pull/9161#issuecomment-330226465 for initial motivation.
@alalek, I didn't added a condition on `MSVC`, as this feature is not specific to Visual Studio.

Note that, this is not the same as #6496 because this Pull Request doesn't add new configurations to OpenCV, it only allows all configuration in the user's project.

### DIFF
--- a/cmake/templates/OpenCVConfig.cmake.in
+++ b/cmake/templates/OpenCVConfig.cmake.in
@@ -199,6 +199,14 @@ foreach(__cvcomponent ${OpenCV_FIND_COMPONENTS})
       )
     endif()
   endif()
+  # OpenCV supports Debug and Release only.
+  # RelWithDebInfo and MinSizeRel are mapped to Release
+  if(TARGET ${__cvcomponent})
+      set_target_properties(${__cvcomponent} PROPERTIES
+          MAP_IMPORTED_CONFIG_MINSIZEREL "Release"
+          MAP_IMPORTED_CONFIG_RELWITHDEBINFO "Release"
+      )
+  endif()
 endforeach()
 
 # ==============================================================


vpisarev:
@bblanchon, thanks, that's a nice feature! @alalek, could you review it, please? Is it compatible with all the CMake versions that we support?

bblanchon:
@vpisarev `MAP_IMPORTED_CONFIG_` is present in [CMake 2.6's documentation](https://cmake.org/cmake/help/cmake2.6docs.html#prop_tgt:MAP_IMPORTED_CONFIG_CONFIG), but I cannot find it in 2.4.
However, since it's just a property that we set, it'll be ignored by earlier versions.

alalek:

TobiasLangnerFUB:
It looks like a MSVC condition is necessary after all. On my machine with Ubuntu 16.04 and OpenCV 3.3.1 the changes of @bblanchon cause compile errors , when my build target is set to DebWithRelInfo. 

`make[2]: *** No rule to make target 'opencv_calib3d-NOTFOUND'`

If I build my project as Debug or Release it compiles. If I remove the changes from the pull request, it will also compile for RelWithDebInfo. 

I don't know if it is relevant, but It seems like there are no Imported Configurations available. I have added this to OpenCVConfig.cmake:

`get_target_property(foo ${__cvcomponent} IMPORTED_CONFIGURATIONS)`
`MESSAGE("IMPORTED_CONFIGURATIONS ${foo}")`

Then CMake will output: "IMPORTED_CONFIGURATIONS NONE"

But shouldn't the MAP_IMPORTED_CONFIG_ property be ignored in this case?

alalek:
@TobiasLangnerFUB The latest discussion is here: #10105

TobiasLangnerFUB:
I have seen that issue, but it doesn't fit my scenario. I haven't build OpenCV on my own, but use the opencv3 package for ROS Kinetic. The website http://wiki.ros.org/opencv3 lists the changes they have made to the CMake configuration. It doesn't look like it was built as RelWithDebInfo.

PR #9658, opened 2017-09-18
Upgrading OpenCL to 1.2.7

whizzzkid:
OpenCL was upgraded to 1.2.7 from 1.2.5 in cuda 9. These files have been shamelessly copied from the OpenCL source that comes along with cuda 9.


alalek:
1. OpenCL is Khronos standard. Proper files should be "grabbed" from Khronos registry: https://www.khronos.org/registry/OpenCL/
2. What kind of problems is observed?

alalek:
0. All Pull Request should be clean from unnecessary/unrelated commits.
Run commands like this:
```
git rebase -i opencv/master
# leave only required commits and save&exit
git push -f origin upgrading-opencl
```

vpisarev:
@whizzzkid, sorry, we cannot accept any sources from a proprietary CUDA distribution. As @alalek said, we can only use the standard headers from Khronos

PR #9668, opened 2017-09-19
Added warning when MSVC_VERSION is not recognized

Erotemic:
This PR is a simple changes that adds a warning when `MSVC_VERSION` is not equal to an expected value. This was causing issues on an older version of opencv (3.1.0), where it did not recognize `MSVC_VERSION 1911`, and it took a lot of digging to figure out what the issue was. 

By adding this warning the hope is that when a new visual studio version is published in the future, the poor souls who try to build an older OpenCV with their shiny new visual studio will encounter this warning.


alalek:
:+1:

PR #9669, opened 2017-09-20
Do not enable asm with clang

kraj:
clang pretends to be gcc 4.2.0 which means we will
use inline asm for no reason, instead of builtins
on clang when possible.

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/3rdparty/carotene/src/channel_extract.cpp
+++ b/3rdparty/carotene/src/channel_extract.cpp
@@ -231,7 +231,7 @@ void extract4(const Size2D &size,
                  srcStride == dst2Stride && \
                  srcStride == dst3Stride &&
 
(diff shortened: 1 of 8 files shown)


alalek:
Thank you for contribution!

There are still other cases with inline asm code: https://github.com/opencv/opencv/blob/8755a19e02a9a4c4080ea17766a9a0ed50e70b36/3rdparty/carotene/src/colorconvert.cpp#L675-L677

Please verify this output and apply similar changes for appropriate lines:
```
grep -Rn GNUC 3rdparty/carotene/ | grep -v clang
```

kraj:
@alalek the intention of patch is to match clang with latest gcc, the lines pointed above will work
identical with gcc or clang.

PR #9674, opened 2017-09-20
cmake: fix build process of OpenCL kernels

alalek:
Patch #9534 is not good.
- `cl2cpp` script may not update timestamp of output `.hpp` file (same contents)
- in this case "make" doesn't consider this file updated and rerun cl2cpp script multiple times.

This patch removes `.hpp` file from the `OUTPUT` list.


vpisarev:
:+1:

PR #9678, opened 2017-09-21
features2d: use universal intrinsic in FAST

tomoaki0705:
### This pullrequest changes
 - switch to universal intrinsic implementation from pure ```SSE2``` implementation


vpisarev:
looks great, thank you! :+1:

PR #9681, opened 2017-09-21
Update OpenCVDownload.cmake unpack option

jodrescher:
### This pullrequest changes

Added new "f" flag to the tar unpack, because without it doesn't complete, an compile fails


alalek:
Thank you for contribution.
Could your add CMake version / error message information?

jodrescher:
I would like to install it with Homebrew. **cmake is 3.9.3** in Homebrew.
I can't reproduce the error after changed and build works correct.

Something cached?

**But this was the error:** https://github.com/jodrescher/opencv/blob/50b651655d5ceaf86537511545f8ef46d9733584/cmake/OpenCVDownload.cmake#L223

I think the issue depends on the **tar implementation**: https://unix.stackexchange.com/questions/268273/tar-without-f-option-in-extraction
I tested it locally and without "f" tar hangs so to say forever, waits for input.

alalek:
Thank you for information!
:+1:

PR #9686, opened 2017-09-21
Update planar_tracking.cpp

sturkmen72:
### This pullrequest changes

Current code uses the first captured frame to select a bounding box.Some web cams returns a dark or blurry image in first capture.

Proposed code waits user's key press to stop video and select a bounding box.


savuor:
:+1:

PR #9691, opened 2017-09-22
Refactored Padding layer

dkurt:
### This pullrequest changes
Extend Padding layer to process multidimensional padding that means it pads several blob axes from different sides (from the left/right) simultaneously.

PR helps to use single `tf.pad` in case of spatial padding:
```python
padded = tf.pad(inp, [[0, 0], [3, 3], [3, 3], [0, 0]])
```
And `nn.SpatialZeroPadding` layer from Torch.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/384


vpisarev:
thanks! :+1:

PR #9692, opened 2017-09-22
dnn: perf tests

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1377


vpisarev:
@alalek, can you briefly describe what's the patch purpose? it looks like besides dnn tests, there is a bunch of changes in the testing engine as well.

dkurt:
@vpisarev, one of the things is that we can now replace performance tests bodies in single function. Before we used testing cycles at every PERF_TEST.

vpisarev:
:+1:

PR #9696, opened 2017-09-22
faster computation of pointcloud from disparity

tompollok:
Many coefficients of Q are 1.0 or 0.0. Here the non zero or one coefficients are extracted from the Q matrix and the matrix vector multiplication is performed manually. The pointcloud calculation is faster as it takes much less multiplications and additions.

The matrix Q is composed in [cvStereoRectify](https://github.com/opencv/opencv/blob/39e742765afd4de1852e4bd25cbac6d61d8c968a/modules/calib3d/src/calibration.cpp#L2450-L2463)

### This pullrequest changes

Results in much faster computation of a pointcloud from a disparity map
Saves 15 out of 16 multiplications and 9 out of 12 additions per pixel.

### DIFF
--- a/modules/calib3d/src/calibration.cpp
+++ b/modules/calib3d/src/calibration.cpp
@@ -2761,7 +2761,14 @@ void cv::reprojectImageTo3D( InputArray _disparity,
     const float bigZ = 10000.f;
     Matx44d _Q;
     Q.convertTo(_Q, CV_64F);
-
+    
+    // get all necessary coefficients of matrix Q
+    const double q03 = _Q(0,3);
+    const double q13 = _Q(1,3);
+    const double q23 = _Q(2,3);
+    const double q32 = _Q(3,2);
+    const double q33 = _Q(3,3);
+        
     int x, cols = disparity.cols;
     CV_Assert( cols >= 0 );
 
@@ -2808,10 +2815,15 @@ void cv::reprojectImageTo3D( InputArray _disparity,
         for( x = 0; x < cols; x++)
         {
             double d = sptr[x];
-            Vec4d homg_pt = _Q*Vec4d(x, y, d, 1.0);
-            dptr[x] = Vec3d(homg_pt.val);
-            dptr[x] /= homg_pt[3];
-
+            
+            //multiplication of _Q*Vec4d(x, y, d, 1.0) divided by scale W
+            const double WX = (x + q03);
+            const double WY = (y + q13);
+            const double WZ = q23;
+            const double W  = d*q32 + q33;
+            
+            dptr[x] = Vec3d(WX/W, WY/W, WZ/W);
+            
             if( fabs(d-minDisparity) <= FLT_EPSILON )
                 dptr[x][2] = bigZ;
         }


tompollok:
I guess the reason the test fails is because the matrix Q in test_reproject_image_to_3d.cpp does not have the same properties as in cvStereoRectify.

[see test_reproject_image_to_3d.cpp](https://github.com/opencv/opencv/blob/39e742765afd4de1852e4bd25cbac6d61d8c968a/modules/calib3d/test/test_reproject_image_to_3d.cpp#L126)

savuor:
@tompollok As far as I understood, the Q matrix can be created by `stereoRectify()` or can be created some other way (for example it can be built manually from rotation matrices multiplication, etc.).
That is why your assumption about zeros and ones of Q is not correct.

tompollok:
@savuor I will try to figure that out. reprojectImageTo3D needs defenitely some speedup as it is quite slow. Ill also provide some measurements.

savuor:
@tompollok Great! There can be several ways of how to speed the function up, like to rewrite it with [universal intrinsics](http://docs.opencv.org/master/df/d91/group__core__hal__intrin.html), etc. You are welcome to do this.
But this PR is not correct, so I close it.

tompollok:
@savuor I think in case someone uses a matrix Q' = T*Q where T is a transformation matrix that transforms the pointcloud from camera coordinate space to some world coordintante space, then my assumption is not valid.

I was thinking about two possible solution: 
- I could check if the Matrix Q is of the form as specified in cvStereoRectify with a lot of elements being zeros or ones and test the matrix if it does.
if(Q has this special property)
{
  //compute it the fast way (is way faster)
}
else
{
  //do it the default way, where a matrix vector multiplication is performed.
}

This way, it would be guaranteed, that the test succeed and nobodys code will be broken.

Or the second way would be to create an opencl version of reprojectImageTo3D(...) that always performs the full matrix vector multiplication.
This would take a bit longer for me to provide a pullrequest for, but i could try to provide an implementation for that.

savuor:
@tompollok I'm not sure that processing special cases of matrix would really improve the speed (I can be mistaken though).
The second variant is more universal and easy to maintain even though it's a bit harder to implement.

tompollok:
@savuor @seiko2plus @terfendail @tomoaki0705 by your contributions i have seen that you are familiar with universal intrinsics. I'd love to hear your advice if you dont mind.

I'd finally like continue to improve the performance of the conversion from a depth map to a 3d pointcloud that is computed using the function reprojectImageTo3D.

Its basically multiplying a 4x4 matrix with a homogeneous 4x1 vector in the general case, while sometimes the depth value is invalid such that the 3d point has to be discarded. This has to be done in for loops over all elements of the 2d depth map.

Vec4d homg_pt = _Q*Vec4d(x, y, d, 1.0);
            dptr[x] = Vec3d(homg_pt.val);
            dptr[x] /= homg_pt[3];

Has anyone a good idea how to start working on that? Should i rather go for opencl or does using universal intrinsics make more sense here? I have never used universal intrinsics so id like to know if the effort/time here is worth it.

Also im not sure how optimized a matrix vector multiplication in opencv already is. Looking at the source files i cant easily find the location where cv::Mat and Vec4d operator overload is defined.
Thanks in advance for your help and time.

PR #9700, opened 2017-09-23
Added iterator implementation to Matx

jakubstanecki:

### DIFF
--- a/modules/core/include/opencv2/core/matx.hpp
+++ b/modules/core/include/opencv2/core/matx.hpp
@@ -210,7 +210,170 @@ public:
     Matx(const Matx<_Tp, n, m>& a, Matx_TOp);
 
     _Tp val[m*n]; //< matrix elements
+
+    /**
+     * Iterator
+     */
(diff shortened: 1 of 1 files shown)


alalek:
Thank you for contribution!

Are there examples with real usage of this? Why this is better than two loops? Does it compatible with [C++11 for-each construction](http://en.cppreference.com/w/cpp/language/range-for)?

Tests are required for new functionality.

I believe we don't need reverse iterator at all.

jakubstanecki:
@alalek I don't know if it's better than two loops. Maybe it's a matter of more elegant code. Do you believe or are you sure that reverse iterator is not needed? Some STL containers implements reverse iterator as well. I will add examples and tests.

vpisarev:
I think, since there is `Mat_<T>(const Matx<T, m, n>& m)` constructor, and for `Mat_<T>` we have STL-compatible iterator, this functionality looks unnecessary for now. Let's close it. Feel free to reopen this PR if there is some extra justification provided.

PR #9705, opened 2017-09-24
Added DNN Darknet Yolo v2 for object detection

AlexeyAB:
```
opencv_extra=dnn_model_darknet_yolo_v2
```

### This pullrequest changes

Added neural network Darknet Yolo v2 for object detection: https://pjreddie.com/darknet/yolo/
Added example of usage: `yolo_object_detection.cpp` / `example_dnn-yolo_object_detection.exe`

* **Supported networks**: `yolo.cfg`, `yolo-voc.cfg`, `tiny-yolo.cfg`, `tiny-yolo-voc.cfg` - can be downloaded: https://drive.google.com/drive/folders/0BwRgzHpNbsWBN3JtSjBocng5YW8
* **Unsupported networks: `yolo9000.cfg`**

Supported layers: 
* route (as concat-layer)
* reorg (as an addition to the reshape-layer)
* maxpool
* convolutional (conv+bn+relu)
* region (detection_out) - added layer

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/385

----

#### Comparison of use:

* original Darknet-Yolo-v2: `darknet.exe detector test data/voc.data yolo-voc.cfg yolo-voc.weights -i 0 -thresh 0.24 data/dog.jpg`

* OpenCV Yolo example: `example_dnn-yolo_object_detection.exe -cfg=yolo/yolo.cfg -model=yolo/yolo.weights -image=yolo/dog.jpg -min_confidence=0.24`

----

#### Comparison of results OpenCV-example vs original Darknet: https://github.com/pjreddie/darknet

For `cfg`, `weights` and `jpg`-s from: https://drive.google.com/drive/folders/0BwRgzHpNbsWBN3JtSjBocng5YW8

* **Network resolution: 416 x 416**
* **threshold = 0.24**
* **nms-threshold = 0.4**

1. **`yolo.cfg` & `yolo.weights`**

    * `dog.jpg` using `yolo.cfg`
coco_dog

    * `eagle.jpg` using `yolo.cfg`
coco_eagle

    * `giraffe.jpg` using `yolo.cfg`
coco_giraffe

----

2. **`yolo-voc.cfg` & `yolo-voc.weights`**

    * `dog.jpg` using `yolo-voc.cfg`
voc_dog

    * `eagle.jpg` using `yolo-voc.cfg`
voc_eagle

    * `giraffe.jpg` using `yolo-voc.cfg`
voc_giraffe

----

### How to train (to detect your custom objects): https://github.com/AlexeyAB/darknet#how-to-train-to-detect-your-custom-objects

----

#### Accuracy-speed:

68747470733a2f2f6873746f2e6f72672f66696c65732f6132342f3231652f3036382f61323432316530363839666234336630383538346465396434346332323135662e6a7067

68747470733a2f2f6873746f2e6f72672f66696c65732f3361362f6664662f6235332f33613666646662353333663334636565396235326264643962623062313964392e6a7067

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -611,6 +611,14 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         virtual ~Importer();
     };
 
+    /** @brief Reads a network model stored in <a href="https://pjreddie.com/darknet/">Darknet</a> model files.
+    *  @param cfgFile      path to the .cfg file with text description of the network architecture.
+    *  @param darknetModel path to the .weights file with learned network.
+    *  @returns Network object that ready to do forward, throw an exception in failure cases.
+    * @details This is shortcut consisting from DarknetImporter and Net::populateNet calls.
+    */
+    CV_EXPORTS_W Net readNetFromDarknet(const String &cfgFile, const String &darknetModel = String());
+
     /**
      *  @deprecated Use @ref readNetFromCaffe instead.
      *  @brief Creates the importer of <a href="http://caffe.berkeleyvision.org">Caffe</a> framework network.
--- a/modules/dnn/src/darknet/darknet_importer.cpp
+++ b/modules/dnn/src/darknet/darknet_importer.cpp
@@ -1,0 +1,195 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+//
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                           License Agreement
+//                For Open Source Computer Vision Library
+//                        (3-clause BSD License)
+//
+// Copyright (C) 2017, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective owners.
+//
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
+//
+// * Neither the names of the copyright holders nor the names of the contributors
+// may be used to endorse or promote products derived from this software
+// without specific prior written permission.
+//
+// This software is provided by the copyright holders and contributors "as is" and
+// any express or implied warranties, including, but not limited to, the implied
+// warranties of merchantability and fitness for a particular purpose are disclaimed.
+// In no event shall copyright holders or contributors be liable for any direct,
+// indirect, incidental, special, exemplary, or consequential damages
+// (including, but not limited to, procurement of substitute goods or services;
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+
+#include "../precomp.hpp"
+
+#include <iostream>
+#include <algorithm>
+#include <vector>
+#include <map>
+
+#include "darknet_io.hpp"
+
+
+namespace cv {
+namespace dnn {
+CV__DNN_EXPERIMENTAL_NS_BEGIN
+
+namespace
+{
+
+class DarknetImporter : public Importer
+{
+    darknet::NetParameter net;
+
+public:
+
+    DarknetImporter() {}
+
+    DarknetImporter(const char *cfgFile, const char *darknetModel)
+    {
+        CV_TRACE_FUNCTION();
+
+        ReadNetParamsFromCfgFileOrDie(cfgFile, &net);
+
+        if (darknetModel && darknetModel[0])
+            ReadNetParamsFromBinaryFileOrDie(darknetModel, &net);
+    }
+
+    struct BlobNote
+    {
+        BlobNote(const std::string &_name, int _layerId, int _outNum) :
+            name(_name), layerId(_layerId), outNum(_outNum) {}
+
+        std::string name;
+        int layerId, outNum;
+    };
+
+    std::vector<BlobNote> addedBlobs;
+    std::map<String, int> layerCounter;
+
+    void populateNet(Net dstNet)
+    {
+        CV_TRACE_FUNCTION();
+
+        int layersSize = net.layer_size();
+        layerCounter.clear();
+        addedBlobs.clear();
+        addedBlobs.reserve(layersSize + 1);
+
+        //setup input layer names
+        {
+            std::vector<String> netInputs(net.input_size());
+            for (int inNum = 0; inNum < net.input_size(); inNum++)
+            {
+                addedBlobs.push_back(BlobNote(net.input(inNum), 0, inNum));
+                netInputs[inNum] = net.input(inNum);
+            }
+            dstNet.setInputsNames(netInputs);
+        }
+
+        for (int li = 0; li < layersSize; li++)
+        {
+            const darknet::LayerParameter &layer = net.layer(li);
+            String name = layer.name();
+            String type = layer.type();
+            LayerParams layerParams = layer.getLayerParams();
+
+            int repetitions = layerCounter[name]++;
+            if (repetitions)
+                name += cv::format("_%d", repetitions);
+
+            int id = dstNet.addLayer(name, type, layerParams);
+
+            // iterate many bottoms layers (for example for: route -1, -4)
+            for (int inNum = 0; inNum < layer.bottom_size(); inNum++)
+                addInput(layer.bottom(inNum), id, inNum, dstNet, layer.name());
+
+            for (int outNum = 0; outNum < layer.top_size(); outNum++)
+                addOutput(layer, id, outNum);
+        }
+
+        addedBlobs.clear();
+    }
+
+    void addOutput(const darknet::LayerParameter &layer, int layerId, int outNum)
+    {
+        const std::string &name = layer.top(outNum);
+
+        bool haveDups = false;
+        for (int idx = (int)addedBlobs.size() - 1; idx >= 0; idx--)
+        {
+            if (addedBlobs[idx].name == name)
+            {
+                haveDups = true;
+                break;
+            }
+        }
+
+        if (haveDups)
+        {
+            bool isInplace = layer.bottom_size() > outNum && layer.bottom(outNum) == name;
+            if (!isInplace)
+                CV_Error(Error::StsBadArg, "Duplicate blobs produced by multiple sources");
+        }
+
+        addedBlobs.push_back(BlobNote(name, layerId, outNum));
+    }
+
+    void addInput(const std::string &name, int layerId, int inNum, Net &dstNet, std::string nn)
+    {
+        int idx;
+        for (idx = (int)addedBlobs.size() - 1; idx >= 0; idx--)
+        {
+            if (addedBlobs[idx].name == name)
+                break;
+        }
+
+        if (idx < 0)
+        {
+            CV_Error(Error::StsObjectNotFound, "Can't find output blob \"" + name + "\"");
+            return;
+        }
+
+        dstNet.connect(addedBlobs[idx].layerId, addedBlobs[idx].outNum, layerId, inNum);
+    }
+
+    ~DarknetImporter()
+    {
+
+    }
+
+};
+
+}
+
+Net readNetFromDarknet(const String &cfgFile, const String &darknetModel /*= String()*/)
+{
+    DarknetImporter darknetImporter(cfgFile.c_str(), darknetModel.c_str());
+    Net net;
+    darknetImporter.populateNet(net);
+    return net;
+}
+
+CV__DNN_EXPERIMENTAL_NS_END
+}} // namespace
--- a/modules/dnn/src/darknet/darknet_io.cpp
+++ b/modules/dnn/src/darknet/darknet_io.cpp
@@ -1,0 +1,624 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+//
+//  By downloading, copying, installing or using the software you agree to this license.
+//  If you do not agree to this license, do not download, install,
+//  copy or use the software.
+//
+//
+//                           License Agreement
+//                For Open Source Computer Vision Library
+//                        (3-clause BSD License)
+//
+// Copyright (C) 2017, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective owners.
+//
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice,
+// this list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
+//
+// * Neither the names of the copyright holders nor the names of the contributors
+// may be used to endorse or promote products derived from this software
+// without specific prior written permission.
+//
+// This software is provided by the copyright holders and contributors "as is" and
+// any express or implied warranties, including, but not limited to, the implied
+// warranties of merchantability and fitness for a particular purpose are disclaimed.
+// In no event shall copyright holders or contributors be liable for any direct,
+// indirect, incidental, special, exemplary, or consequential damages
+// (including, but not limited to, procurement of substitute goods or services;
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+
+/*M///////////////////////////////////////////////////////////////////////////////////////
+//MIT License
+//
+//Copyright (c) 2017 Joseph Redmon
+//
+//Permission is hereby granted, free of charge, to any person obtaining a copy
+//of this software and associated documentation files (the "Software"), to deal
+//in the Software without restriction, including without limitation the rights
+//to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//copies of the Software, and to permit persons to whom the Software is
+//furnished to do so, subject to the following conditions:
+//
+//The above copyright notice and this permission notice shall be included in all
+//copies or substantial portions of the Software.
+//
+//THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+//SOFTWARE.
+//
+//M*/
+
+#include <opencv2/core.hpp>
+
+#include <iostream>
+#include <fstream>
+#include <sstream>
+
+#include "darknet_io.hpp"
+
+namespace cv {
+    namespace dnn {
+        namespace darknet {
+
+            template<typename T>
+            T getParam(const std::map<std::string, std::string> &params, const std::string param_name, T init_val)
+            {
+                std::map<std::string, std::string>::const_iterator it = params.find(param_name);
+                if (it != params.end()) {
+                    std::stringstream ss(it->second);
+                    ss >> init_val;
+                }
+                return init_val;
+            }
+
+            class setLayersParams {
+
+                NetParameter *net;
+                int layer_id;
+                std::string last_layer;
+                std::vector<std::string> fused_layer_names;
+
+            public:
+                setLayersParams(NetParameter *_net, std::string _first_layer = "data") :
+                    net(_net), layer_id(0), last_layer(_first_layer)
+                {}
+
+                void setLayerBlobs(int i, std::vector<cv::Mat> blobs)
+                {
+                    cv::dnn::experimental_dnn_v1::LayerParams &params = net->layers[i].layerParams;
+                    params.blobs = blobs;
+                }
+
+                cv::dnn::experimental_dnn_v1::LayerParams getParamConvolution(int kernel, int pad,
+                    int stride, int filters_num)
+                {
+                    cv::dnn::experimental_dnn_v1::LayerParams params;
+                    params.name = "Convolution-name";
+                    params.type = "Convolution";
+
+                    params.set<int>("kernel_size", kernel);
+                    params.set<int>("pad", pad);
+                    params.set<int>("stride", stride);
+
+                    params.set<bool>("bias_term", false);	// true only if(BatchNorm == false)
+                    params.set<int>("num_output", filters_num);
+
+                    return params;
+                }
+
+
+                void setConvolution(int kernel, int pad, int stride,
+                    int filters_num, int channels_num, int use_batch_normalize, int use_relu)
+                {
+                    cv::dnn::experimental_dnn_v1::LayerParams conv_param =
+                        getParamConvolution(kernel, pad, stride, filters_num);
+
+                    darknet::LayerParameter lp;
+                    std::string layer_name = cv::format("conv_%d", layer_id);
+
+                    // use BIAS in any case
+                    if (!use_batch_normalize) {
+                        conv_param.set<bool>("bias_term", true);
+                    }
+
+                    lp.layer_name = layer_name;
+                    lp.layer_type = conv_param.type;
+                    lp.layerParams = conv_param;
+                    lp.bottom_indexes.push_back(last_layer);
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+
+                    if (use_batch_normalize)
+                    {
+                        cv::dnn::experimental_dnn_v1::LayerParams bn_param;
+
+                        bn_param.name = "BatchNorm-name";
+                        bn_param.type = "BatchNorm";
+                        bn_param.set<bool>("has_weight", true);
+                        bn_param.set<bool>("has_bias", true);
+                        bn_param.set<float>("eps", 1E-6);	// .000001f in Darknet Yolo
+
+                        darknet::LayerParameter lp;
+                        std::string layer_name = cv::format("bn_%d", layer_id);
+                        lp.layer_name = layer_name;
+                        lp.layer_type = bn_param.type;
+                        lp.layerParams = bn_param;
+                        lp.bottom_indexes.push_back(last_layer);
+                        last_layer = layer_name;
+                        net->layers.push_back(lp);
+                    }
+
+                    if (use_relu)
+                    {
+                        cv::dnn::experimental_dnn_v1::LayerParams activation_param;
+                        activation_param.set<float>("negative_slope", 0.1f);
+                        activation_param.name = "ReLU-name";
+                        activation_param.type = "ReLU";
+
+                        darknet::LayerParameter lp;
+                        std::string layer_name = cv::format("relu_%d", layer_id);
+                        lp.layer_name = layer_name;
+                        lp.layer_type = activation_param.type;
+                        lp.layerParams = activation_param;
+                        lp.bottom_indexes.push_back(last_layer);
+                        last_layer = layer_name;
+                        net->layers.push_back(lp);
+                    }
+
+                    layer_id++;
+                    fused_layer_names.push_back(last_layer);
+                }
+
+                void setMaxpool(size_t kernel, size_t pad, size_t stride)
+                {
+                    cv::dnn::experimental_dnn_v1::LayerParams maxpool_param;
+                    maxpool_param.set<cv::String>("pool", "max");
+                    maxpool_param.set<int>("kernel_size", kernel);
+                    maxpool_param.set<int>("pad", pad);
+                    maxpool_param.set<int>("stride", stride);
+                    maxpool_param.set<cv::String>("pad_mode", "SAME");
+                    maxpool_param.name = "Pooling-name";
+                    maxpool_param.type = "Pooling";
+                    darknet::LayerParameter lp;
+
+                    std::string layer_name = cv::format("pool_%d", layer_id);
+                    lp.layer_name = layer_name;
+                    lp.layer_type = maxpool_param.type;
+                    lp.layerParams = maxpool_param;
+                    lp.bottom_indexes.push_back(last_layer);
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+                    layer_id++;
+                    fused_layer_names.push_back(last_layer);
+                }
+
+                void setConcat(int number_of_inputs, int *input_indexes)
+                {
+                    cv::dnn::experimental_dnn_v1::LayerParams concat_param;
+                    concat_param.name = "Concat-name";
+                    concat_param.type = "Concat";
+                    concat_param.set<int>("axis", 1);	// channels are in axis = 1
+
+                    darknet::LayerParameter lp;
+
+                    std::string layer_name = cv::format("concat_%d", layer_id);
+                    lp.layer_name = layer_name;
+                    lp.layer_type = concat_param.type;
+                    lp.layerParams = concat_param;
+                    for (int i = 0; i < number_of_inputs; ++i)
+                        lp.bottom_indexes.push_back(fused_layer_names.at(input_indexes[i]));
+
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+
+                    layer_id++;
+                    fused_layer_names.push_back(last_layer);
+                }
+
+                void setIdentity(int bottom_index)
+                {
+                    cv::dnn::experimental_dnn_v1::LayerParams identity_param;
+                    identity_param.name = "Identity-name";
+                    identity_param.type = "Identity";
+
+                    darknet::LayerParameter lp;
+
+                    std::string layer_name = cv::format("identity_%d", layer_id);
+                    lp.layer_name = layer_name;
+                    lp.layer_type = identity_param.type;
+                    lp.layerParams = identity_param;
+                    lp.bottom_indexes.push_back(fused_layer_names.at(bottom_index));
+
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+
+                    layer_id++;
+                    fused_layer_names.push_back(last_layer);
+                }
+
+                void setReorg(int stride)
+                {
+                    cv::dnn::experimental_dnn_v1::LayerParams reorg_params;
+                    reorg_params.name = "Reorg-name";
+                    reorg_params.type = "Reorg";
+                    reorg_params.set<int>("reorg_stride", stride);
+
+                    darknet::LayerParameter lp;
+                    std::string layer_name = cv::format("reorg_%d", layer_id);
+                    lp.layer_name = layer_name;
+                    lp.layer_type = reorg_params.type;
+                    lp.layerParams = reorg_params;
+                    lp.bottom_indexes.push_back(last_layer);
+                    last_layer = layer_name;
+
+                    net->layers.push_back(lp);
+
+                    layer_id++;
+                    fused_layer_names.push_back(last_layer);
+                }
+
+                void setPermute()
+                {
+                    cv::dnn::experimental_dnn_v1::LayerParams permute_params;
+                    permute_params.name = "Permute-name";
+                    permute_params.type = "Permute";
+                    int permute[] = { 0, 2, 3, 1 };
+                    cv::dnn::DictValue paramOrder = cv::dnn::DictValue::arrayInt(permute, 4);
+
+                    permute_params.set("order", paramOrder);
+
+                    darknet::LayerParameter lp;
+                    std::string layer_name = cv::format("premute_%d", layer_id);
+                    lp.layer_name = layer_name;
+                    lp.layer_type = permute_params.type;
+                    lp.layerParams = permute_params;
+                    lp.bottom_indexes.push_back(last_layer);
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+
+                    layer_id++;
+                    fused_layer_names.push_back(last_layer);
+                }
+
+                void setRegion(float thresh, int coords, int classes, int anchors, int classfix, int softmax, int softmax_tree, float *biasData)
+                {
+                    cv::dnn::experimental_dnn_v1::LayerParams region_param;
+                    region_param.name = "Region-name";
+                    region_param.type = "Region";
+
+                    region_param.set<float>("thresh", thresh);
+                    region_param.set<int>("coords", coords);
+                    region_param.set<int>("classes", classes);
+                    region_param.set<int>("anchors", anchors);
+                    region_param.set<int>("classfix", classfix);
+                    region_param.set<bool>("softmax_tree", softmax_tree);
+                    region_param.set<bool>("softmax", softmax);
+
+                    cv::Mat biasData_mat = cv::Mat(1, anchors * 2, CV_32F, biasData).clone();
+                    region_param.blobs.push_back(biasData_mat);
+
+                    darknet::LayerParameter lp;
+                    std::string layer_name = "detection_out";
+                    lp.layer_name = layer_name;
+                    lp.layer_type = region_param.type;
+                    lp.layerParams = region_param;
+                    lp.bottom_indexes.push_back(last_layer);
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+
+                    layer_id++;
+                    fused_layer_names.push_back(last_layer);
+                }
+            };
+
+            std::string escapeString(const std::string &src)
+            {
+                std::string dst;
+                for (size_t i = 0; i < src.size(); ++i)
+                    if (src[i] > ' ' && src[i] <= 'z')
+                        dst += src[i];
+                return dst;
+            }
+
+            template<typename T>
+            std::vector<T> getNumbers(const std::string &src)
+            {
+                std::vector<T> dst;
+                std::stringstream ss(src);
+
+                for (std::string str; std::getline(ss, str, ',');) {
+                    std::stringstream line(str);
+                    T val;
+                    line >> val;
+                    dst.push_back(val);
+                }
+                return dst;
+            }
+
+            bool ReadDarknetFromCfgFile(const char *cfgFile, NetParameter *net)
+            {
+                std::ifstream ifile;
+                ifile.open(cfgFile);
+                if (ifile.is_open())
+                {
+                    bool read_net = false;
+                    int layers_counter = -1;
+                    for (std::string line; std::getline(ifile, line);) {
+                        line = escapeString(line);
+                        if (line.empty()) continue;
+                        switch (line[0]) {
+                        case '\0': break;
+                        case '#': break;
+                        case ';': break;
+                        case '[':
+                            if (line == "[net]") {
+                                read_net = true;
+                            }
+                            else {
+                                // read section
+                                read_net = false;
+                                ++layers_counter;
+                                const size_t layer_type_size = line.find("]") - 1;
+                                CV_Assert(layer_type_size < line.size());
+                                std::string layer_type = line.substr(1, layer_type_size);
+                                net->layers_cfg[layers_counter]["type"] = layer_type;
+                            }
+                            break;
+                        default:
+                            // read entry
+                            const size_t separator_index = line.find('=');
+                            CV_Assert(separator_index < line.size());
+                            if (separator_index != std::string::npos) {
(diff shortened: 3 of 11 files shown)


vpisarev:
@AlexeyAB, thank you, this is very valuable contribution! Could you please add some regression test(s) for this functionality?

AlexeyAB:
@vpisarev I added: `modules/dnn/test/test_darknet_importer.cpp`
Also added test data and models (cfg, weights) for DNN Darknet Yolo v2: https://github.com/opencv/opencv_extra/pull/385

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,626 @@
...
+
+        int layers_counter = -1;
+
+        setLayersParams setParams(net);
+
+        typedef std::map<int, std::map<std::string, std::string> >::iterator it_type;
+        for (it_type i = net->layers_cfg.begin(); i != net->layers_cfg.end(); ++i) {
+            ++layers_counter;
+            std::map<std::string, std::string> &layer_params = i->second;
+            std::string layer_type = layer_params["type"];
```
dkurt: Please, add an assertion for unknown layer type to prevent unexpected errors. In example, I can't read any model now because an every `layer_type` ends with `]` character (`convolutional]`, `maxpool]`).

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -611,6 +611,18 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         virtual ~Importer();
     };
 
+    /** @brief Creates the importer of <a href="https://pjreddie.com/darknet/">Darknet</a> framework network.
+    *  @param cfgFile      path to the .cfg file with text description of the network architecture.
+    *  @param darknetModel path to the .weights file with learned network.
+    *  @returns Pointer to the created importer, NULL in failure cases.
+    */
+    CV_EXPORTS_W Ptr<Importer> createDarknetImporter(const String &cfgFile, const String &darknetModel = String());
```
dkurt: We defined methods like `createCaffeImporter` as deprecated. Please keep only `readNetFromDarknet`.

samples/dnn/yolo_object_detection.cpp
```diff
@@ -0,0 +1,123 @@
...
+        cerr << "cfg-file:     " << modelConfiguration << endl;
+        cerr << "weights-file: " << modelBinary << endl;
+        cerr << "Models can be downloaded here:" << endl;
+        cerr << "https://pjreddie.com/darknet/yolo/" << endl;
+        exit(-1);
+    }
+
+    cv::Mat frame = cv::imread(parser.get<string>("image"), -1);
+
+    if (frame.channels() == 4)
```
dkurt: It isn't necessary: just use `imread` with default argument (http://docs.opencv.org/master/d4/da8/group__imgcodecs.html#ga288b8b3da0892bd651fce07b3bbd3a56).

samples/dnn/yolo_object_detection.cpp
```diff
@@ -0,0 +1,123 @@
...
+        cvtColor(frame, frame, cv::COLOR_BGRA2BGR);
+    //! [Prepare blob]
+    Mat preprocessedFrame = preprocess(frame, network_width, network_height);
```
dkurt: Please, use `blobFromImage`'s arguments to make preprocessing (http://docs.opencv.org/3.3.0/d6/d0f/group__dnn.html#ga0507466a789702eda8ffcdfa37f4d194).

samples/dnn/yolo_object_detection.cpp
dkurt: @AlexeyAB, thanks, but I meant that `cv::imread` can read images with alpha into 24bit, http://docs.opencv.org/master/d4/da8/group__imgcodecs.html#gga61d9b0126a3e57d9277ac48327799c80af660544735200cbe942eea09232eb822.

AlexeyAB: @dkurt I fixed it. 
Initially I did it as in `ssd_object_detection.cpp` and I thought that this has some hidden meaning :)

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,638 @@
...
+                for (size_t k = 0; k < layers_vec.size(); ++k) {
+                    layers_vec[k] += layers_counter;
+                    current_shape.input_channels += net->out_shape_vec[layers_vec[k]].input_channels;
+                }
+                current_shape.input_w = net->out_shape_vec[layers_vec[0]].input_w;
+                current_shape.input_h = net->out_shape_vec[layers_vec[0]].input_h;
+
+                setParams.setConcat(layers_vec.size(), layers_vec.data());
+            }
+            else if (layer_type == "reorg")
```
dkurt: I'm a bit confused about reorg layer. Let the input is:
```
channel_0  channel_1  channel_2  channel_3
0 1        4 5        8 9        c d
2 3        6 7        a b        e f
```
and reorgStride = 2. So an output shape is 4x4x1 and values are:
```
output
1 4 1 5
8 c 9 d
2 6 3 7
a e b f
```
?

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,638 @@
...
+            {
+                int stride = getParam<int>(layer_params, "stride", 2);
+                current_shape.input_w = current_shape.input_w / stride;
+                current_shape.input_h = current_shape.input_h / stride;
+                current_shape.input_channels = current_shape.input_channels * (stride*stride);
+
+                setParams.setReshape(stride, current_shape.input_channels, current_shape.input_h, current_shape.input_w);
+
+                current_shape.input_channels = 256;
```
dkurt: Magic number?

modules/dnn/src/darknet/darknet_io.cpp
dkurt: It works now but `Reproducibility_TinyYoloVoc` and `Reproducibility_YoloVoc` tests are failed for me. Do them passed locally?

dkurt:
@AlexeyAB, Thank you for the valuable contribution! We need to test all the new usage carefully. Can you add some unit tests with small few-layer networks like we do for the other importers? (https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_torch_importer.cpp and https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/torch/torch_gen_test_data.lua, https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_tf_importer.cpp and https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/tensorflow/generate_tf_models.py). In example, write simple configs, run darknet to initialize the weights, pass some random input and get the output, put configs/weights/inputs/outputs into the opencv_extra/testdata/dnn darknet subfolder?

modules/dnn/src/darknet/darknet_io.cpp
AlexeyAB: I left unchanged a bit strange original implementation of this layer. It increases the field of view of each final activation.
Reshape: **26 x 26 x 64 -> 13 x 13 x 256**
reorg

```
input
    0, 1, 2, 3,
    4, 5, 6, 7, 
    8, 9, a, b,
    c, d, e, f
```

```
output
channel_0  channel_1  channel_2  channel_3
0 2        1 3        4 6        5 7
8 a        9 b        c e        d f
```
* OpenCV C++ example: http://coliru.stacked-crooked.com/a/eb13942be083fa3d
* Darknet C++ example: http://coliru.stacked-crooked.com/a/225d7fb1f25b286c

----
* Param `reverse` usually absent in the cfg-file of model, so `reverse = 0` by default https://github.com/pjreddie/darknet/blob/master/src/parser.c#L387

* Then `l.reverse = reverse;` https://github.com/pjreddie/darknet/blob/master/src/reorg_layer.c#L28

* So the function `reorg()` is called with `forward=0`: https://github.com/pjreddie/darknet/blob/8215a8864d4ad07e058acafd75b2c6ff6600b9e8/src/reorg_layer.c#L108

* `reorg_cpu()`-implementation uses `out[in_index] = x[out_index];`: https://github.com/pjreddie/darknet/blob/master/src/blas.c#L25

modules/dnn/src/layers/concat_layer.cpp
```diff
@@ -93,7 +95,10 @@ class ConcatLayerImpl : public ConcatLayer
             axisSum += curShape[cAxis];
         }
         outputs[0][cAxis] = axisSum;
-        return false;
+
+        // Darknet ROUTE-layer
+        if (useRoute) return true;
```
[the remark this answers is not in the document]
AlexeyAB: I don't know why, but it doesn't work for Yolo if `getMemoryShapes` returns `false`.
Route-layer simply copies unchanged outputs from several layers: https://github.com/pjreddie/darknet/blob/master/src/route_layer.c#L83

Used `copy_cpu()` with `INCX=1` and `INCY=1`: https://github.com/pjreddie/darknet/blob/master/src/blas.c#L208

AlexeyAB:
@dkurt So I already added testdata and models for object detection using DNN Darknet Yolo v2 to the `opencv_extra`: https://github.com/opencv/opencv_extra/pull/385
* `testdata/dnn/dogr.jpg` - test image resized to the network size 416x416, to eliminate the side effects of resizing
* `testdata/dnn/tiny-yolo-voc.cfg` - tiny model of Yolo v2 for Pascal VOC dataset
* `testdata/dnn/yolo-voc.cfg` - full model of Yolo v2 for Pascal VOC dataset
* Changed `testdata/dnn/download_models.py`:
    * Downloads `https://pjreddie.com/media/files/yolo-voc.weights` - full model of Yolo v2 trained for Pascal VOC dataset
    * Downloads `https://pjreddie.com/media/files/tiny-yolo-voc.weights` - tiny model of Yolo v2 trained for Pascal VOC dataset

In this pull-request there is: `modules/dnn/test/test_darknet_importer.cpp`
* `TEST(Test_Darknet, read_tiny_yolo_voc)`
* `TEST(Test_Darknet, read_yolo_voc)`
* `TEST(Reproducibility_TinyYoloVoc, Accuracy)` - **test passed**, results compared with original Darknet results: https://github.com/opencv/opencv/pull/9705/files#diff-9360334ee3b0098a7deae1fe9d8ba25dR106
* `TEST(Reproducibility_YoloVoc, Accuracy)` - **test failed**, results compared with original Darknet results: https://github.com/opencv/opencv/pull/9705/files#diff-9360334ee3b0098a7deae1fe9d8ba25dR158

```
[----------] 2 tests from Test_Darknet
[ RUN      ] Test_Darknet.read_tiny_yolo_voc
[       OK ] Test_Darknet.read_tiny_yolo_voc (2 ms)
[ RUN      ] Test_Darknet.read_yolo_voc
[       OK ] Test_Darknet.read_yolo_voc (2 ms)
[----------] 2 tests from Test_Darknet (4 ms total)

[----------] 1 test from Reproducibility_TinyYoloVoc
[ RUN      ] Reproducibility_TinyYoloVoc.Accuracy
[       OK ] Reproducibility_TinyYoloVoc.Accuracy (233 ms)
[----------] 1 test from Reproducibility_TinyYoloVoc (233 ms total)

[----------] 1 test from Reproducibility_YoloVoc
[ RUN      ] Reproducibility_YoloVoc.Accuracy
c:\opencv_source\opencv\modules\dnn\test\test_common.hpp(57): error: Expected: (
normInf) <= (lInf), actual: 0.00010556 vs 0.0001
[  FAILED  ] Reproducibility_YoloVoc.Accuracy (441 ms)
[----------] 1 test from Reproducibility_YoloVoc (441 ms total)
```

dkurt:
@AlexeyAB, yeah, it's great, but I meant tests for separate layers. First of all it's necessary to protect your work done from the bugs that might appear in future development. Next thing is that BuildBot doesn't test these models for now because there aren't there. My local tests are failed and I think we can solve a problem by small checks for separate layers.
```
[----------] 1 test from Reproducibility_TinyYoloVoc
[ RUN      ] Reproducibility_TinyYoloVoc.Accuracy
unknown file: Failure
C++ exception with description "/home/dkurtaev/opencv/modules/ts/src/ts_func.cpp:1374: error: (-215) src1.type() == src2.type() && src1.size == src2.size in function norm
" thrown in the test body.
[  FAILED  ] Reproducibility_TinyYoloVoc.Accuracy (101 ms)
[----------] 1 test from Reproducibility_TinyYoloVoc (101 ms total)

[----------] 1 test from Reproducibility_YoloVoc
[ RUN      ] Reproducibility_YoloVoc.Accuracy
/home/dkurtaev/opencv/modules/dnn/test/test_common.hpp:54: Failure
Expected: (normL1) <= (l1), actual: 0.000232658 vs 1e-05
/home/dkurtaev/opencv/modules/dnn/test/test_common.hpp:57: Failure
Expected: (normInf) <= (lInf), actual: 0.00485086 vs 0.0001
[  FAILED  ] Reproducibility_YoloVoc.Accuracy (317 ms)
[----------] 1 test from Reproducibility_YoloVoc (317 ms total)
```
I referenced how we write unit tests for different frameworks. The binary size of required data is not so huge (i.e. less than 0.5MB for TensorFlow layers) and you can add it in a single PR @ opencv_extra.

modules/dnn/src/darknet/darknet_io.cpp
dkurt: Thanks! Anyway I suggest to replace it as a single layer or think how we can do the same transformations using existing ones (Permute, Reshape). Reshape layer doesn't change the data by definition neither in one of the frameworks.

AlexeyAB:
@dkurt I changed test image from `dogr.jpg` to the lossless [`dog416.png`](https://github.com/opencv/opencv_extra/pull/385/files) in the `opencv_extra` and now it works.
Tested on both OS: Windows 7 x64 and Linux Debian 8.2 x64
```
[----------] 2 tests from Test_Darknet
[ RUN      ] Test_Darknet.read_tiny_yolo_voc
[       OK ] Test_Darknet.read_tiny_yolo_voc (0 ms)
[ RUN      ] Test_Darknet.read_yolo_voc
[       OK ] Test_Darknet.read_yolo_voc (1 ms)
[----------] 2 tests from Test_Darknet (3 ms total)

[----------] 1 test from Reproducibility_TinyYoloVoc
[ RUN      ] Reproducibility_TinyYoloVoc.Accuracy
[       OK ] Reproducibility_TinyYoloVoc.Accuracy (134 ms)
[----------] 1 test from Reproducibility_TinyYoloVoc (135 ms total)

[----------] 1 test from Reproducibility_YoloVoc
[ RUN      ] Reproducibility_YoloVoc.Accuracy
[       OK ] Reproducibility_YoloVoc.Accuracy (475 ms)
[----------] 1 test from Reproducibility_YoloVoc (475 ms total)
```
Results for comparison with OpenCV version are obtained on Linux Debian 8.2 using the current last commit of Darknet Yolo v2 compiled with GPU=0, OPENMP=1 and OpenCV=1: https://github.com/pjreddie/darknet

Using commands: 
* `./darknet detector test ./cfg/voc.data  ./cfg/tiny-yolo-voc.cfg ./tiny-yolo-voc.weights -thresh 0.24 ./dog416.png`

* `./darknet detector test ./cfg/voc.data  ./cfg/yolo-voc.cfg ./yolo-voc.weights -thresh 0.24 ./dog416.png`

----

About the new layer  `region_layer`. I will see how you write unit tests for separate layers, and how can I do this.

modules/dnn/src/darknet/darknet_io.cpp
AlexeyAB: @dkurt I added reorg as separate layer `reorg_layer.cpp`

modules/dnn/src/layers/concat_layer.cpp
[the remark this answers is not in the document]
AlexeyAB: @dkurt Yes, route with a single input (bottom layer) is used as identity layer.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,637 @@
...
+            ifile.read(reinterpret_cast<char *>(&net->seen), sizeof(uint64_t));
+        }
+        else {
+            int32_t iseen = 0;
+            ifile.read(reinterpret_cast<char *>(&iseen), sizeof(int32_t));
+            net->seen = iseen;
+        }
+        net->transpose = (net->major_ver > 1000) || (net->minor_ver > 1000);
+
+        layerShape current_shape;
```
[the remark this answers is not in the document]
AlexeyAB: Yes, weights file doesn't contain kernels shapes.
Darknet tracks shapes while parsing a cfg-file:
* `parse_network_cfg()` https://github.com/pjreddie/darknet/blob/master/src/parser.c#L630
    * `parse_convolutional()` https://github.com/pjreddie/darknet/blob/master/src/parser.c#L169
        * `make_convolutional_layer()` https://github.com/pjreddie/darknet/blob/master/src/convolutional_layer.c#L166
             * `convolutional_out_height()` https://github.com/pjreddie/darknet/blob/master/src/convolutional_layer.c#L66
```
int convolutional_out_height(convolutional_layer l)
{
    return (l.h + 2*l.pad - l.size) / l.stride + 1;
}
```

AlexeyAB:
@dkurt Yes, route layer with a single input (bottom layer) is used as identity layer.

modules/dnn/src/layers/reorg_layer.cpp
```diff
@@ -0,0 +1,139 @@
...
+            for (int k = 0; k < channels; ++k) {
+                for (int j = 0; j < height; ++j) {
+                    for (int i = 0; i < width; ++i) {
+                        int in_index = i + width*(j + height*k);
+                        int c2 = k % out_c;
+                        int offset = k / out_c;
+                        int w2 = i*reorgStride + offset % reorgStride;
+                        int h2 = j*reorgStride + offset / reorgStride;
+                        int out_index = w2 + width*reorgStride*(h2 + height*reorgStride*c2);
+                        dstData[in_index] = srcData[out_index];
```
dkurt: Is there no typo in `in`<->`out` indices place?

modules/dnn/src/layers/reorg_layer.cpp
```diff
@@ -0,0 +1,139 @@
...
+        CV_Assert(inputs.size() > 0);
+        outputs = std::vector<MatShape>(inputs.size(), shape(
+            inputs[0][0],
+            inputs[0][1] * reorgStride * reorgStride,
+            inputs[0][2] / reorgStride,
+            inputs[0][3] / reorgStride));
+
+        CV_Assert(outputs[0][0] > 0 && outputs[0][1] > 0 && outputs[0][2] > 0 && outputs[0][3] > 0);
+
+        return true;
```
dkurt: It seems to me Reorg layer can't work in-place. `getMemoryShapes` returns true if layer can do it.

modules/dnn/src/layers/region_layer.cpp
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    bool getMemoryShapes(const std::vector<MatShape> &inputs,
+                         const int requiredOutputs,
+                         std::vector<MatShape> &outputs,
+                         std::vector<MatShape> &internals) const
+    {
+        CV_Assert(inputs.size() > 0);
+        outputs = std::vector<MatShape>(inputs.size(), shape(inputs[0][1] * inputs[0][2] * anchors, inputs[0][3] / anchors));
+        return true;
```
dkurt: The same as Reorg layer: it should returns false.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,645 @@
...
+        void setConvolution(int kernel_h, int kernel_w, int pad_h, int pad_w, int stride_h, int stride_w,
+            int filters_num, int channels_num, int use_batch_normalize, int use_relu)
+        {
+            cv::dnn::experimental_dnn_v1::LayerParams conv_param =
+                getParamConvolution(kernel_h, kernel_w, pad_h, pad_w, stride_h, stride_w,
+                    filters_num, channels_num);
+
+            darknet::LayerParameter lp;
+            std::string layer_name = toString(layer_id);
+            if (use_batch_normalize || use_relu) layer_name = "conv_" + layer_name;
```
dkurt: It's better to name layers using type prefix every time. Moreover some of layers just named with numbers and it's hard to debug them.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,637 @@
...
+                net->layers.push_back(lp);
+            }
+
+            layer_id++;
+        }
+
+        void setMaxpool(size_t kernel, size_t pad, size_t stride, size_t channels_num)
+        {
+            cv::dnn::experimental_dnn_v1::LayerParams maxpool_param;
+            maxpool_param.set<cv::String>("pool", "max");
```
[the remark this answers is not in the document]
AlexeyAB: Ok. Also required `maxpool_param.set<cv::String>("pad_mode", "SAME");` for odd sizes of layers.

modules/dnn/src/layers/reorg_layer.cpp
AlexeyAB: No, there is no typo, initially I left unchanged a bit strange original implementation of this layer.
But now I have changed this place so that there is no confusion.

modules/dnn/src/darknet/darknet_io.cpp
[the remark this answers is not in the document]
dkurt: However only one way of padding strategy is used: manual values or padMode ("SAME", "VALID") from TensorFlow. Please take a look on the "ceil_mode" flag instead: https://github.com/opencv/opencv/blob/master/modules/dnn/src/layers/pooling_layer.cpp#L629.

modules/dnn/src/layers/concat_layer.cpp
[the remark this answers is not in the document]
dkurt: @AlexeyAB, could you add an extra branch during route layer creation: add Concat layer if number of inputs more than 1 or Identity layer otherwise?

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,637 @@
...
+        }
+
+        void setLastLayerName(std::string layer_name)
+        {
+            if(net->layers.size() > 0)
+                net->layers.rbegin()->layer_name = layer_name;
+        }
+
+        cv::dnn::experimental_dnn_v1::LayerParams getParamConvolution(int kernel, int pad,
+            int stride, int filters_num, int channels_num)
```
dkurt: Unused variable `channels_num`

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,637 @@
...
+                lp.bottom_indexes.push_back(last_layer);
+                last_layer = layer_name;
+                net->layers.push_back(lp);
+            }
+
+            layer_id++;
+            fused_layer_names.push_back(last_layer);
+        }
+
+        void setMaxpool(size_t kernel, size_t pad, size_t stride, size_t channels_num)
```
dkurt: Unused variable `channels_num`

modules/dnn/src/darknet/darknet_io.cpp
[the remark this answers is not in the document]
dkurt: @AlexeyAB, May be we can remove at least width/height sizes tracking? As I can see only `current_shape.input_channels` is used to read convolutional layer weights.

modules/dnn/src/darknet/darknet_io.cpp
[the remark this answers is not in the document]
AlexeyAB: * Accuracy test passed for Tiny-Yolo if `padMode="SAME"` with any `ceil_mode` value.
* Accuracy test can't be passed for Tiny-Yolo for any values of padMode (`padMode="VALID"` or if `padMod` isn't set) with any `ceil_mode` value.

modules/dnn/src/layers/concat_layer.cpp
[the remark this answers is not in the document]
AlexeyAB: @dkurt I added identity layer for the 1 input. But why concat layer can't work with 1 input, and why there is no CV_Assert for this case?

modules/dnn/src/layers/reorg_layer.cpp
```diff
@@ -0,0 +1,139 @@
...
+    {
+        CV_Assert(inputs.size() > 0);
+        outputs = std::vector<MatShape>(inputs.size(), shape(
+            inputs[0][0],
+            inputs[0][1] * reorgStride * reorgStride,
+            inputs[0][2] / reorgStride,
+            inputs[0][3] / reorgStride));
+
+        CV_Assert(outputs[0][0] > 0 && outputs[0][1] > 0 && outputs[0][2] > 0 && outputs[0][3] > 0);
+
```
dkurt: Please, add an assertion that `total(outputs[0]) == total(inputs[0])`.

modules/dnn/src/layers/reorg_layer.cpp
```diff
@@ -0,0 +1,139 @@
...
+            Mat srcBlob = *inputs[i];
+            MatShape inputShape = shape(srcBlob), outShape = shape(outputs[i]);
+            float *dstData = outputs[0].ptr<float>();
+            const float *srcData = srcBlob.ptr<float>();
+
+            int channels = inputShape[1], height = inputShape[2], width = inputShape[3];
+
+            int out_c = channels / (reorgStride*reorgStride);
+
+            for (int k = 0; k < channels; ++k) {
```
dkurt: Please make it more clear: iterate over output dimensions and map them to input ones.

modules/dnn/test/test_darknet_importer.cpp
```diff
@@ -0,0 +1,175 @@
...
+        resize(sample, sample, inputSize);
+
+    net.setInput(blobFromImage(sample, 1 / 255.F), "data");
+    Mat out = net.forward("detection_out");
+
+    Mat detection;
+    const float confidenceThreshold = 0.24;
+
+    for (int i = 0; i < out.rows; i++) {
+        float const*const prob_ptr = &out.at<float>(i, 5);
```
dkurt: `float const*const` is a bit confusing (there are 4 places with it).

May I ask you to use named constant variables or place a comments because it's hard to understand the magic numbers? Especially it's about samples and tests.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,638 @@
...
+
+
+        void setConvolution(int kernel, int pad, int stride,
+            int filters_num, int channels_num, int use_batch_normalize, int use_relu)
+        {
+            cv::dnn::experimental_dnn_v1::LayerParams conv_param =
+                getParamConvolution(kernel, pad, stride, filters_num);
+
+            darknet::LayerParameter lp;
+            std::string layer_name = "conv_" + toString(layer_id);
```
dkurt: Please try to use `cv::format("conv_%d", layer_id)` instead `toString` here and in other places.

modules/dnn/src/darknet/darknet_io.hpp
```diff
@@ -0,0 +1,119 @@
...
+//M*/
+
+#ifndef __OPENCV_DNN_DARKNET_IO_HPP__
+#define __OPENCV_DNN_DARKNET_IO_HPP__
+
+#include <opencv2/dnn/dnn.hpp>
+
+namespace darknet {
+
+    class LayerParameter {
```
dkurt: I hope we can emit this structure. Layers are connected sequentially or using explicit numeric offsets starting from the newly added layer. So I think it's possible to use single vector of layers during network building. May I ask you to try it?

modules/dnn/src/layers/reorg_layer.cpp
AlexeyAB: Most likely there was a logical mistake in the original version of Darknet in the reorg layer: https://github.com/pjreddie/darknet/blob/master/src/blas.c#L9

It works as I described if called `reorg(input, output, out_w, out_h, _out_c, ...);` https://github.com/opencv/opencv/pull/9705#discussion_r141307309

But in original Darknet version function is called in this way `reorg(input, output, in_w, in_h, ...);`, so the one-to-one correspondence of the input and output parameters is preserved, but very strange permutations occur.

But because original Darknet works with this implementation of reorg and all models trained using it, then we can't fix logical mistake.

Why the author has not found and corrected this error? I think:
* Perhaps this logical error does not spoil the detection accuracy, so it was not detected. 
* Theoretically, we can assume that this error even increased the accuracy - so the author found it and left it.

----

> iterate over output dimensions and map them to input ones.

So I can implement it in a such way, 
but it works correctly only **`if (in_w % 2 == 0 && in_h % 2 == 0 && in_c % 4 == 0)`** : http://coliru.stacked-crooked.com/a/b962d20938362d4f
```
void reorg_my(const float*const srcData,  float *const dstData, int width, int height, int channels, int reorgStride)
{
	int outChannels = channels * reorgStride * reorgStride;
	int outHeight = height / reorgStride;
	int outWidth = width / reorgStride;

	for (int y = 0; y < outHeight; ++y) {
		for (int x = 0; x < outWidth; ++x) {
			for (int c = 0; c < outChannels; ++c) {
				int out_index = x + outWidth*(y + outHeight*c);

				int step = c / channels;
				int x_offset = step % reorgStride;
				int y_offset = reorgStride * ((step / reorgStride) % reorgStride);

				int in_x = x * reorgStride + x_offset;
				
				int out_seq_y = y + c*outHeight;
				int in_intermediate_y = out_seq_y*2 - out_seq_y%2;
				in_intermediate_y = in_intermediate_y % (channels*height);
				int in_c = in_intermediate_y / height;
				int in_y = in_intermediate_y % height + y_offset;
						
				int in_index = in_x + width*(in_y + height*in_c);
				dstData[out_index] = srcData[in_index];
			}
		}
	}
}
```

modules/dnn/src/darknet/darknet_io.hpp
AlexeyAB: Do you mean that I should to try use `cv::dnn::experimental_dnn_v1::LayerParams` instead of `darknet::LayerParameter`?

dkurt: Yeah, I think we can just parse text and binary files simultaneously: for an every entry in config we create a new one LayerParams and fill it depends on layer type. If specific layer has weights - read them from opened binary file. Then add a layer to final network (addLayerToPrev or addLayer with multiple connections based on id of the new layer and offsets i.e. -1, -4 of route).

modules/dnn/test/test_darknet_importer.cpp
AlexeyAB: I removed `const*const`, also added named constant variables and described the format of network output that compared to the reference in the tests.

But why is `const*const` confusing, is this contrary to the code style conventions that is accepted in OpenCV? 
1st `const` forbids modification of values pointed to by this pointer, 2nd `const`  forbids modification of this pointer.

modules/dnn/src/darknet/darknet_io.hpp
dkurt: @AlexeyAB, on the other hand, let's keep it as is now. I'll just install darknet and compare it with PR and we can merge it.

modules/dnn/src/darknet/darknet_importer.cpp
```diff
@@ -0,0 +1,202 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
dkurt: I think we need to replace it. Please take a look on more fresh versions of the licence.

modules/dnn/src/darknet/darknet_importer.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+namespace cv {
+namespace dnn {
+CV__DNN_EXPERIMENTAL_NS_BEGIN
+
+namespace
+{
+
+template<typename T>
+static cv::String toString(const T &v)
```
dkurt: Can we remove it? There is only one entry of `toString` that might be replaced to `cv::format`. Thanks!

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,623 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
dkurt: The same, please update the licence.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -0,0 +1,623 @@
...
+
+#include <opencv2/core.hpp>
+
+#include <iostream>
+#include <fstream>
+#include <sstream>
+
+#include "darknet_io.hpp"
+
+namespace darknet {
```
dkurt: Is it possible to nest it into the `cv::dnn` namespace?

modules/dnn/src/darknet/darknet_io.hpp
```diff
@@ -0,0 +1,115 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
dkurt: One again: the licence update.

modules/dnn/src/darknet/darknet_io.hpp
```diff
@@ -0,0 +1,115 @@
...
+//SOFTWARE.
+//
+//M*/
+
+#ifndef __OPENCV_DNN_DARKNET_IO_HPP__
+#define __OPENCV_DNN_DARKNET_IO_HPP__
+
+#include <opencv2/dnn/dnn.hpp>
+
+namespace darknet {
```
dkurt: namespace `cv::dnn::darknet`

modules/dnn/test/test_darknet_importer.cpp
```diff
@@ -0,0 +1,185 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
dkurt: The licence update.

modules/dnn/test/test_darknet_importer.cpp
```diff
@@ -0,0 +1,185 @@
...
+// (including, but not limited to, procurement of substitute goods or services;
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+
+#include "test_precomp.hpp"
+#include "npy_blob.hpp"
```
dkurt: An extra include directive.

modules/dnn/src/darknet/darknet_importer.cpp
dkurt: @AlexeyAB, I don't think that we should place past years copyrights to newly added source.
@vpisarev, please tell us if the following header is enough?
```

//
// Copyright (C) 2017, Intel Corporation, all rights reserved.
// Third party copyrights are property of their respective owners.
```
Or the same as above but with a current date.

dkurt:
👍 
@AlexeyAB, thank you!

adamhrv:
Will there be a Python example for CV2 Darknet DNN?

When running `readNetFromDarknet` --> `net.forward()` in Python, the YoloV2 `yolo-voc.cfg` and `yolo-voc.weights`  predictions result doesn't seem to provide any detection info.

This works OK:
```
net = cv2.dnn.readNetFromDarknet(path_to_prototxt, path_to_model)
```
This seems to work OK:
```
imw,imh = (416,416)
blob = cv2.dnn.blobFromImage(cv2.resize(im, (416, 416)))  
net.setInput(blob)
detections = net.forward()
```
But the detection result doesn't make sense:
```
print('Detections len: {}'.format(len(detections)))
Detections len: 845
```
```
print('Detections 0: {}'.format(detections[0]))
Detections 0: [  5.25983423e-02   2.70044785e-02   9.04742535e-03   2.62199971e-03
   2.96895425e-10   0.00000000e+00   0.00000000e+00   0.00000000e+00
   0.00000000e+00   0.00000000e+00   0.00000000e+00   0.00000000e+00
   0.00000000e+00   0.00000000e+00   0.00000000e+00   0.00000000e+00
   0.00000000e+00   0.00000000e+00   0.00000000e+00   0.00000000e+00
   0.00000000e+00   0.00000000e+00   0.00000000e+00   0.00000000e+00
   0.00000000e+00]
```

Referring to [yolo_object_detection.cpp](https://github.com/opencv/opencv/blob/639b800787a37a8a560ff8006f8e46e2e7b43868/samples/dnn/yolo_object_detection.cpp#L132)

Thanks for your work porting to opencv.

fabito:
This [repo](https://github.com/the-house-of-black-and-white/opencv-dnn-demo) shows a face detection demo. The models were trained on the Widerface dataset and are available for download (weights and cfgs).

cansik:
It seems that sometimes the implementation does not return the correct detection class type. Even with the default pre-trained 80 class model, sometimes the kind of class is unknown (only zeros in the `5` to `85` column).

Also when I train my own model with two different classes, it only tells me the confidence, but not the class itself. Here is the dump of the result matrix:

```
[0.0401995, 0.031487927, 0.031374719, 0.056794502, 0.0012705148, 0, 0;
 0.04043559, 0.049086317, 0.22208165, 0.25339442, 0.00040254797, 0, 0;
 0.037753381, 0.036353372, 0.60338461, 0.96344143, 0.00046182834, 0, 0;
 0.034598812, 0.039104842, 0.82008636, 0.50834161, 0.0003691405, 0, 0;
 0.038062122, 0.038740713, 1.2965844, 0.84108508, 0.00030294483, 0, 0;
 0.11062718, 0.023617726, 0.083281159, 0.038289066, 0.00061990606, 0, 0;
 0.11572818, 0.046723619, 0.24326266, 0.19994149, 0.00070420129, 0, 0;
 0.11441542, 0.03014034, 0.48347056, 0.84074426, 0.00024233655, 0, 0;
 0.11178426, 0.035311423, 0.71218336, 0.46975818, 0.00027817892, 0, 0;
 0.11641605, 0.038460143, 1.1835779, 0.68950963, 0.00037787345, 0, 0;
 0.19388326, 0.02510196, 0.091353334, 0.028100489, 0.00042861109, 0, 0;
 0.19292459, 0.048351433, 0.22517532, 0.21802522, 0.00040042991, 0, 0;
 0.18578193, 0.031620376, 0.63702989, 0.94969654, 0.0002352587, 0, 0;
 0.18794204, 0.038929399, 0.74781221, 0.44536978, 0.00019949637, 0, 0;
 0.19130239, 0.03961565, 0.9100669, 0.58336502, 0.00025648749, 0, 0;
 0.27135891, 0.02685423, 0.090977632, 0.032328393, 0.00058053283, 0, 0;
 0.27235663, 0.050213691, 0.25882462, 0.26637048, 0.00038077682, 0, 0;
 0.25689453, 0.0360987, 0.69405353, 0.82805032, 0.00038927642, 0, 0;
 0.25953323, 0.045636557, 0.58337092, 0.38006222, 0.00038867182, 0, 0;
 0.2683081, 0.042552318, 0.89848542, 0.72007042, 0.00038684186, 0, 0;
...
```

When using the darkent to detect the objects, it is always able to tell what kind of object it is (with the confidence). Do you experience the same behaviour?

AlexeyAB:
@cansik  The fact is that for values that less than threshold: Darknet zeroes the `scale`, but OpenCV zeroes the `prob`. And Darknet zeroes `prob` only if `scale` isn't zeroed.


You can get **the same bounded boxes with the same probability in both OpenCV and Darknet** (but not the same probs which less than threshold) only with:
* 1 year old Darknet
* default Yolo v2 model 
* the same `threshold`
* the same `nms` threshold
* on image that whose size is equal to the size of network (416x416) due to different resize approaches: https://github.com/AlexeyAB/darknet/issues/232#issuecomment-336955485

**Note**: if in the original Darknet `scale < thresh`, then `objectness=0` and **prob will not be zeroed** even if `prob < thresh`. But in OpenCV-dnn **output** there isn't `scale`, so **prob will be zeroed** if `prob < thresh`.


----

What `thresh` did you use for Darknet and for OpenCV-dnn-yolo?
There should be zeros in both implementations: Darknet and OpenCV-dnn-yolo if prob or scale less than threshold.

1. OpenCV-dnn-yolo: https://github.com/opencv/opencv/blob/7ae83df8aa176d8402172c2d4d1cb5ba07d6c048/modules/dnn/src/layers/region_layer.cpp#L258-L261

----

2. Darknet: 
    * https://github.com/pjreddie/darknet/blob/508381b37fe75e0e1a01bcb2941cb0b31eb0e4c9/src/region_layer.c#L404
    * https://github.com/pjreddie/darknet/blob/508381b37fe75e0e1a01bcb2941cb0b31eb0e4c9/src/region_layer.c#L426-L432

**Note**: if in the original Darknet `scale < thresh`, then `objectness=0` and **prob will not be zeroed** even if `prob < thresh`. But in OpenCV-dnn **output** there isn't `scale`, so **prob will be zeroed** if `prob < thresh`.
```cpp
dets[index].objectness = scale > thresh ? scale : 0;
```
```cpp
                if(dets[index].objectness){
                    for(j = 0; j < l.classes; ++j){
                        int class_index = entry_index(l, 0, n*l.w*l.h + i, l.coords + 1 + j);
                        float prob = scale*predictions[class_index];
                        dets[index].prob[j] = (prob > thresh) ? prob : 0;
                    }
                }
```

cansik:
@AlexeyAB As far as I know, the probability gets cleared by opencv. That is ok for me if the confidence is under the threshold.

How do I set the threshold in opencv? Is it possible to lower it zero, to get the probabilities of all predictions? Or do you mean with *threshold* the threshold defined in the `cfg` file?

Here is an example which is really strange:

image

On this image the trained network finds three characters (in opencv). All of them have a confidence higher than `80%`, but the probability for these classes is zeroed. Do you know why this happens?

```
Class: none, Confidence: 0.8652
Class: none, Confidence: 0.8734
Class: none, Confidence: 0.8448
```

Second example: 

image

For this picture I have exported the result matrix: [yolo_results.sheets](https://docs.google.com/spreadsheets/d/17TzheAjits93Fn5RpQMk_ehDyN76SnkrciQ2vCGXtHw/edit?usp=sharing)

Why are there only three probabilities, and not more for each item? Or do I understand the result matrix wrongly?

dkurt:
@cansik, Have you tried to vary `thresh` parameter of `[region]` layer in `.cfg` file?

cansik:
@dkurt Yes that helped, thank you. I was not sure where to set the threshold, and a (stupid) bug in my result evaluation let to no difference, even when I played with this param.

Now everything works as expected!

dkurt:
@cansik, OpenCV parses `.cfg` file and extracts this threshold for non-maximum suppression procedure. I think the best solution is to set `thresh` to zero but post-process output detections in your application. You may try [object_detection.py](https://github.com/opencv/opencv/blob/master/samples/dnn/object_detection.py) sample. There is a slider you can use to change a confidence threshold and see the difference.

ahmadfaizan1990:
@AlexeyAB can we visualize our model on tensorboard after training?
if yes kindly show me the steps.
thanks!!!

AlexeyAB:
@ahmadfaizan1990 TensorBoard can visualize only data from models which were trained by using TensorFlow.

For Darknet your can see Loss & mAP (accuracy) chart during training: https://github.com/AlexeyAB/darknet#when-should-i-stop-training

68747470733a2f2f6873746f2e6f72672f776562742f79642f766c2f61672f7964766c616775746f66327a636e6a6f64737467726f656e3861632e6a706567

ahmadfaizan1990:
> @ahmadfaizan1990 TensorBoard can visualize only data from models which were trained by using TensorFlow.
> 
> For Darknet your can see Loss & mAP (accuracy) chart during training: https://github.com/AlexeyAB/darknet#when-should-i-stop-training
> 
> 68747470733a2f2f6873746f2e6f72672f776562742f79642f766c2f61672f7964766c616775746f66327a636e6a6f64737467726f656e3861632e6a706567

@AlexeyAB thanks for your reply, i have one more question if i change some convolutional layer in cfg file or want to minimize the layers, because after change in cfg file i am getting this error.
Untitled

AlexeyAB:
@ahmadfaizan1990 Ask here: https://github.com/AlexeyAB/darknet/issues

modules/dnn/src/layers/reorg_layer.cpp
Wulingtian: > Most likely there was made a logical mistake in the original version of Darknet in the reorg layer: https://github.com/pjreddie/darknet/blob/master/src/blas.c#L9
> 
> It works as I described if called `reorg(input, output, out_w, out_h, _out_c, ...);` [#9705 (comment)](https://github.com/opencv/opencv/pull/9705#discussion_r141307309)
> 
> But in the original Darknet version function is called in this way `reorg(input, output, in_w, in_h, in_c, ...);`, so the one-to-one correspondence of the input and output parameters is preserved, but very strange permutations occur.
> 
> But because original Darknet works with this implementation of reorg and all models trained using it, then we can't fix this logical mistake.
> 
> Why the author has not found and corrected this error? I think:
> 
>     * Perhaps this logical error does not spoil the detection accuracy, so it was not detected.
> 
>     * Theoretically, we can assume that this error even increased the accuracy - so the author found it and left it.
> 
> 
> > iterate over output dimensions and map them to input ones.
> 
> So I can implement it in a such way,
> but it works correctly only **`if (in_w % 2 == 0 && in_h % 2 == 0 && in_c % 4 == 0)`** : http://coliru.stacked-crooked.com/a/b962d20938362d4f
> 
> ```
> void reorg_my(const float*const srcData,  float *const dstData, int width, int height, int channels, int reorgStride)
> {
> 	int outChannels = channels * reorgStride * reorgStride;
> 	int outHeight = height / reorgStride;
> 	int outWidth = width / reorgStride;
> 
> 	for (int y = 0; y < outHeight; ++y) {
> 		for (int x = 0; x < outWidth; ++x) {
> 			for (int c = 0; c < outChannels; ++c) {
> 				int out_index = x + outWidth*(y + outHeight*c);
> 
> 				int step = c / channels;
> 				int x_offset = step % reorgStride;
> 				int y_offset = reorgStride * ((step / reorgStride) % reorgStride);
> 
> 				int in_x = x * reorgStride + x_offset;
> 				
> 				int out_seq_y = y + c*outHeight;
> 				int in_intermediate_y = out_seq_y*2 - out_seq_y%2;
> 				in_intermediate_y = in_intermediate_y % (channels*height);
> 				int in_c = in_intermediate_y / height;
> 				int in_y = in_intermediate_y % height + y_offset;
> 						
> 				int in_index = in_x + width*(in_y + height*in_c);
> 				dstData[out_index] = srcData[in_index];
> 			}
> 		}
> 	}
> }
> ```

Is there a GPU version of reorg？

PR #9706, opened 2017-09-24
imgcodecs(tiff): drop legacy code

alalek:
Drop code under *undefined* HAVE_TIFF
(violates several policies, like a way to turn off tiff support)


vpisarev:
makes sense! :+1:

PR #9707, opened 2017-09-25
Simplifying instructions in initUndistortRectifyMapLine_AVX

woodychow:
I didn't know _mm256_cvtpd_epi32 can be used to cast double to int32 with rounding when I first wrote the AVX2 code

### DIFF
--- a/modules/imgproc/src/undistort.avx2.cpp
+++ b/modules/imgproc/src/undistort.avx2.cpp
@@ -161,16 +161,8 @@ int initUndistortRectifyMapLine_AVX(float* m1f, float* m2f, short* m1, ushort* m
             __u = _mm256_mul_pd(__u, _mm256_set1_pd(INTER_TAB_SIZE));
             __v = _mm256_mul_pd(__v, _mm256_set1_pd(INTER_TAB_SIZE));
 
-            __m128 __u_float = _mm256_cvtpd_ps(__u);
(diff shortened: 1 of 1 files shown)


modules/imgproc/src/undistort.avx2.cpp
```diff
@@ -69,7 +69,7 @@ int initUndistortRectifyMapLine_AVX(float* m1f, float* m2f, short* m1, ushort* m
     const __m256d __matTilt_12 = _mm256_set1_pd(matTilt[5]);
     const __m256d __matTilt_22 = _mm256_set1_pd(matTilt[8]);
 
-    for (; j <= width - 4; j += 4, _x += 4 * ir[0], _y += 4 * ir[3], _w += 4 * ir[6])
+    for (; j < width - 4; j += 4, _x += 4 * ir[0], _y += 4 * ir[3], _w += 4 * ir[6])
```
alalek: > Should be less than width - 4. If width = 4, the for loop should be executed once only.

Actually if width == 4 then this loop will not be executed at all (`0 < 4 - 4` is false).

woodychow: Sorry I was dreaming. I will change in a bit as I don't have access to a dev machine.

alalek:
Thank you! :+1:

PR #9714, opened 2017-09-26
imgproc: use universal intrinsic as much as possible

tomoaki0705:
### This pullrequest changes
  * make ```SSE3``` part as common as possible with universal intrinsic implementation
  * put the reducing part out of the main loop

### DIFF
--- a/modules/imgproc/src/smooth.cpp
+++ b/modules/imgproc/src/smooth.cpp
@@ -3198,12 +3198,10 @@ public:
     {
         int i, j, cn = dest->channels(), k;
         Size size = dest->size();
-        #if CV_SSE3
+#if CV_SIMD128
         int CV_DECL_ALIGNED(16) buf[4];
-        float CV_DECL_ALIGNED(16) bufSum[4];
-        static const unsigned int CV_DECL_ALIGNED(16) bufSignMask[] = { 0x80000000, 0x80000000, 0x80000000, 0x80000000 };
-        bool haveSSE3 = checkHardwareSupport(CV_CPU_SSE3);
-        #endif
+        bool haveSIMD128 = hasSIMD128();
+#endif
 
         for( i = range.start; i < range.end; i++ )
         {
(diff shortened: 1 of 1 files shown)


modules/imgproc/src/smooth.cpp
```diff
@@ -3265,58 +3276,73 @@ class BilateralFilter_8u_Invoker :
...
+
+                        wsum  += bufFloat[0];
+                        sum_b += bufFloat[1];
+                        sum_g += bufFloat[2];
+                        sum_r += bufFloat[3];
+#else
+                        wsum += v_reduce_sum(vsumw);
+                        sum_b += v_reduce_sum(vsumb);
+                        sum_g += v_reduce_sum(vsumg);
+                        sum_r += v_reduce_sum(vsumr);
```
alalek: There is `v_reduce_sum4` with SSE3 replacement.

Could we store (or calculate) "wsum, sum_b, sum_g, sum_r" as `v_float32x4` ?

modules/imgproc/src/smooth.cpp
```diff
@@ -3217,35 +3215,48 @@ class BilateralFilter_8u_Invoker :
...
-                             _sw = _mm_hadd_ps(_w, _cw);
-                             _sw = _mm_hadd_ps(_sw, _sw);
-                             _mm_storel_pi((__m64*)bufSum, _sw);
-
-                             sum += bufSum[1];
-                             wsum += bufSum[0];
+                            v_float32x4 _valF = v_float32x4(sptr[j + space_ofs[k]],
+                                sptr[j + space_ofs[k + 1]],
+                                sptr[j + space_ofs[k + 3]],
+                                sptr[j + space_ofs[k + 2]]);
```
savuor: Is it a typo? Looks like there should be `space_ofs[k], space_ofs[k+1], space_ofs[k+2], space_ofs[k+3]`

modules/imgproc/src/smooth.cpp
```diff
@@ -3198,12 +3198,10 @@ class BilateralFilter_8u_Invoker :
     {
         int i, j, cn = dest->channels(), k;
         Size size = dest->size();
-        #if CV_SSE3
+#if CV_SIMD128
         int CV_DECL_ALIGNED(16) buf[4];
-        float CV_DECL_ALIGNED(16) bufSum[4];
-        static const unsigned int CV_DECL_ALIGNED(16) bufSignMask[] = { 0x80000000, 0x80000000, 0x80000000, 0x80000000 };
-        bool haveSSE3 = checkHardwareSupport(CV_CPU_SSE3);
```
savuor: We need to ask about SSE3 presence for the cases when a user has no SSE3 on their CPU.

modules/imgproc/src/smooth.cpp
```diff
@@ -3217,35 +3215,48 @@ class BilateralFilter_8u_Invoker :
...
+                                color_weight[buf[2]],
+                                color_weight[buf[3]]);
+                            v_float32x4 _sw = v_load(space_weight+k);
+                            v_float32x4 _w = _cw * _sw;
+                            _cw = _w * _valF;
+
+                            vsumw += _w;
+                            vsumc += _cw;
                         }
+#if CV_SSE3
```
savuor: The same, we should also check haveSSE3 to be sure that it's presented on user's CPU.

modules/imgproc/src/smooth.cpp
alalek: Not anymore.

OpenCV checks available features and [fails on problems](https://github.com/opencv/opencv/blob/3.3.0/modules/core/src/system.cpp#L506-L519).

modules/imgproc/src/smooth.cpp
```diff
@@ -3265,58 +3276,73 @@ class BilateralFilter_8u_Invoker :
...
+                            _b *=  _w;
+                            _g *=  _w;
+                            _r *=  _w;
+
+                            vsumw += _w;
+                            vsumb += _b;
+                            vsumg += _g;
+                            vsumr += _r;
+                        }
+#if CV_SSE3
```
savuor: The same about `haveSSE3` flag.

modules/imgproc/src/smooth.cpp
savuor: Thanks, other my comments about SSE3 are not actual anymore

modules/imgproc/src/smooth.cpp
tomoaki0705: Thanks! I really wonder why the test passed....

modules/imgproc/src/smooth.cpp
tomoaki0705: ```v_reduce_sum4``` !!!  I dreamed about this implementation and it exists !! wow !!

tomoaki0705:
Thank you for the comments.  I followed 2 comments.

modules/imgproc/src/smooth.cpp
savuor: You can also use `v_reduce_sum4` in that place (and, in fact, in other blocks which use `_mm_hadd_ps` in order to get rid of SSE3-dependant code).

tomoaki0705: Indeed, your right.  I followed your comment and the code looks very nice with no more ```CV_SSE3```
Thanks !

savuor:
:+1:

PR #9726, opened 2017-09-27
Fix out of bounds write

### LINKED ISSUE #9723: Out of bounds write causes Segmentation Fault
**System information (version)**

* OpenCV => 3.3 (the latest commit: 7475d23 2017-09-26)
* Operating System / Platform => Linux
* Compiler => gcc

**Detailed description**

An invalid writing occurs in the FillUniColor and FillUniGray functions in opencv/modules/imgcodecs/src/utils.cpp

**POC**
https://github.com/blendin/pocs/blob/master/opencv/0.OOB_Write_FillUniColor

**Steps to reproduce**
```c++
#include <opencv2/opencv.hpp>

int main(void) {
    cv::imread("0.OOB_Write_FillUniColor")
}
```

**Crash Details**
```
ASAN:DEADLYSIGNAL
=================================================================
==23351==ERROR: AddressSanitizer: SEGV on unknown address 0x7f9f027530fd (pc 0x7f9f01405bd0 bp 0x000000006060 sp 0x7fff2f321208 T0)
==23351==The signal is caused by a WRITE memory access.
    #0 0x7f9f01405bcf in FillUniColor(unsigned char*, unsigned char*&, int, int, int&, int, int, PaletteEntry) (/usr/local/lib/libopencv_imgcodecs.so.3.3+0x73bcf)
    #1 0x7f9f01414895 in cv::BmpDecoder::readData(cv::Mat&) (/usr/local/lib/libopencv_imgcodecs.so.3.3+0x82895)
    #2 0x7f9f013fc642 in cv::imread_(cv::String const&, int, int, cv::Mat*) (/usr/local/lib/libopencv_imgcodecs.so.3.3+0x6a642)
    #3 0x7f9f013fbe6b in cv::imread(cv::String const&, int) (/usr/local/lib/libopencv_imgcodecs.so.3.3+0x69e6b)
    #4 0x4f118b in main /lz/targets/opencv.cc:7:15
    #5 0x7f9f0011682f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #6 0x41a6e8 in _start (/lz/targets/verify+0x41a6e8)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (/usr/local/lib/libopencv_imgcodecs.so.3.3+0x73bcf) in FillUniColor(unsigned char*, unsigned char*&, int, int, int&, int, int, PaletteEntry)
==23351==ABORTING
```

blendin:
resolves #9723 

Analysis
===
**rle8** is missing a check when in encoding mode to see if we are writing past the height, `y`, of the image. This leads to an oob write. 

Fix
===
Same fix that is in place for **rle4** and **rle8** in default mode.

### DIFF
--- a/modules/imgcodecs/src/grfmt_bmp.cpp
+++ b/modules/imgcodecs/src/grfmt_bmp.cpp
@@ -375,6 +375,9 @@ decode_rle4_bad: ;
                                                 gray_palette[code] );
 
                         line_end_flag = y - prev_y;
+
+                        if( y >= m_height )
+                            break;
                     }
                     else if( code > 2 ) // absolute mode
                     {


savuor:
By the way, do we need to add the same check in other places, for example at [line 396](https://github.com/opencv/opencv/pull/9726/files#diff-236c3294512b9353649b64ad8385e984R396)?

savuor:
:+1:

blendin:
@savuor I haven't dug too deep into these functions. It doesn't look like `y` is used for anything meaningful in the `FillColorRow` functions or the if blocks they are in. There might be something bad in these functions because they don't really use the `y` variable. Checking that `y >= m_height` after every point it is incremented might be enough to stop a crash in the `FillColorRow` functions. I will audit these though with this patch in place and see if anything turns up.

I would merge this first though, and I'll submit another patch if something turns up.

Issue #4440, opened 2015-07-27
hough circle detection not symmetric, misses obvious circle


opencv-pushbot:
Transferred from http://code.opencv.org/issues/2461

```
|| Walter Blume on 2012-10-19 17:20
|| Priority: Normal
|| Affected: None
|| Category: imgproc, video
|| Tracker: Bug
|| Difficulty: None
|| PR: None
|| Platform: None / None
```
## hough circle detection not symmetric, misses obvious circle

```
For some reason the Hough circle detection misses an obvious circle, but if I flip the image, the circle is detected.  One would think that results would be similar with flipped image, but they are not.

Attached is source code and test input image that exhibited the problem.  Also screenshot with arrow showing the obvious circle miss.

Using OpenCV 2.4.2, built with TBB on Windows7 64 bit machine
```
##### Daniil Osokin on 2012-10-22 06:34

```
Thank you, we will check it.
-   Category set to imgproc, video
```
##### Jason Harper on 2014-02-22 23:04

```
The problem is in the non-maxima suppression. It is only checking if a value that passes the threshold is also larger (but not equal) than all of it's neighbors.  In the first case there are two pixels with equal values next to each other.  When that happens neither pixel is set as a possible center and the circle is ignored.  When the image is flipped one of these pixels gets an extra count leaving a single peak which gets labeled as a possible center.  

I changed the check to >= and it fixed the problem and will submit a pull request soon.
```
##### leonardo bocchi on 2015-06-01 15:59

```
I faced the same problem, and found the same solution.
To avoid duplication, I changed lines 1112-1113 of hough.cpp to read:

           adata[base] >= adata[base-1] && adata[base] > adata[base+1] &&
           adata[base] >= adata[base-acols-2] && adata[base] > adata[base+acols+2] )

It seems working and no duplicate detections.
```

savuor:
Fixed in #9730

PR #9730, opened 2017-09-27
Fix bug: non-maximum suppression for hough circle

yanlend:
The non-maximum suppression in the Hough accumulator incorrectly ignores maxima that extend over more than one cell, i.e. two neighboring cells both have the same accumulator value. This maximum is dropped completely instead of picking at least one of the entries. This frequently results in obvious circles being missed.

The behavior is now changed to be the same as for hough_lines.

See also https://github.com/opencv/opencv/issues/4440


yanlend:
The cuda hough circles code is correctly using >= already.

https://github.com/opencv/opencv/blob/7475d23fec4bb7c7e2b662a0fa022d706dd2b520/modules/cudaimgproc/src/cuda/hough_circles.cu#L145

savuor:
:+1:

PR #9734, opened 2017-09-27
Swap dimensions of deconvolution kernel

dkurt:
### This pullrequest changes
* Update deconvolution layer's kernel shape: from OIHW to IOHW.
  source: http://answers.opencv.org/question/175165/dnn-from-caffe-deconvolution-layer-assert-fails/
* Groups mode for deconvolution layer in Halide.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/386


vpisarev:
thanks! :+1:

PR #9736, opened 2017-09-28
Const and class to json

Cartucho:
### This pullrequest changes

Adding python constants and classes to the `pyopencv_signatures.json` file.

### DIFF
--- a/modules/python/src2/gen2.py
+++ b/modules/python/src2/gen2.py
@@ -811,7 +811,9 @@ class FuncInfo(object):
             cname = classinfo.cname + '::' + cname
         py_signatures = codegen.py_signatures.setdefault(cname, [])
         for v in self.variants:
-            compat_name = self.namespace.replace("cv", "cv2", 1) + "." + v.name
+            compat_name = v.name
+            if not self.isclassmethod:
+                compat_name = self.namespace.replace("cv", "cv2", 1) + "." + compat_name
             s = dict(name=compat_name, arg=v.py_arg_str, ret=v.py_return_str)
             for old in py_signatures:
                 if s == old:
--- a/modules/python/src2/gen2.py
+++ b/modules/python/src2/gen2.py
@@ -857,9 +857,14 @@ class PythonWrapperGenerator(object):
         self.classes[classinfo.name] = classinfo
 
         # Add Class to json file.
+        namespace, classes, name = self.split_decl_name(name)
+        namespace = '.'.join(namespace)
+        name = '_'.join(classes+[name])
+
+        compat_name = namespace.replace("cv", "cv2", 1) + "." + name
         py_signatures = self.py_signatures.setdefault(classinfo.cname, [])
-        py_signatures.append(dict(name=classinfo.name))
-        #print(classinfo.name)
+        py_signatures.append(dict(name=compat_name))
+        #print(compat_name)
 
     def split_decl_name(self, name):
         chunks = name.split('.')
--- a/modules/python/src2/gen2.py
+++ b/modules/python/src2/gen2.py
@@ -811,9 +811,12 @@ class FuncInfo(object):
             cname = classinfo.cname + '::' + cname
         py_signatures = codegen.py_signatures.setdefault(cname, [])
         for v in self.variants:
-            compat_name = v.name
-            if not self.isclassmethod:
-                compat_name = self.namespace.replace("cv", "cv2", 1) + "." + compat_name
+            compat_name = self.namespace.replace("cv", "cv2", 1) + "."
+            if self.isclassmethod:
+                compat_name = compat_name + self.classname.replace("ml_", "").replace("flann_", "") + "_" + v.name
+            else:
+                compat_name = compat_name + v.name
+            print(compat_name)
             s = dict(name=compat_name, arg=v.py_arg_str, ret=v.py_return_str)
             for old in py_signatures:
                 if s == old:
--- a/modules/python/src2/gen2.py
+++ b/modules/python/src2/gen2.py
@@ -855,6 +855,11 @@ class PythonWrapperGenerator(object):
             sys.exit(-1)
         self.classes[classinfo.name] = classinfo
 
+        # Add Class to json file.
+        py_signatures = self.py_signatures.setdefault(classinfo.cname, []) 
+        py_signatures.append(dict(name=classinfo.name)) 
+        #print(classinfo.name)
+
     def split_decl_name(self, name):
         chunks = name.split('.')
         namespace = chunks[:-1]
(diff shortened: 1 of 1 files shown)


Cartucho:
@alalek what do you think of this approach?

alalek:
Looks good!

I believe we should provide more accurate names:
```
    "cv::dnn::DNN_BACKEND_DEFAULT": [
        {
            "name": "DNN_BACKEND_DEFAULT",
            "ret": 0
        }
    ]
```
but there is lost additional "sub-module":
```
>>> cv2.DNN_BACKEND_HALIDE
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'DNN_BACKEND_HALIDE'
>>> cv2.dnn.DNN_BACKEND_HALIDE
1
```

Actually the same problem with functions too (existed before this patch):
```
>>> cv2.readNetFromTensorflow
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'readNetFromTensorflow'
>>> cv2.dnn.readNetFromTensorflow
<built-in function readNetFromTensorflow>

    "cv::dnn::readNetFromTensorflow": [
        {
            "arg": "model",
            "name": "readNetFromTensorflow",
            "ret": "retval"
        }
    ]
```

Need to find how sub-modules are handled (or ignore this for future changes with filing issue).


Also It is quite unclear to me how "expanded" values of enums could help OpenCV Users. I believe proper "Python" names is enough.

Cartucho:
@alalek Sorry for the delay! I've been busy in the University.
If we use the namespace we can do a quick fix to add the sub-module info.

Let me know what you think, I can do the same for the functions.

alalek:
Looks good to me!
Please add namespace for the functions and for the classes.

Cartucho:
@alalek I added the namespace to the functions and classes.

I noticed that in some of the classes a `_create(params...)` should be added
I.e:  `cv2.ml.ParamGrid` should be: `cv2.ml.ParamGrid_create()`

alalek:
Perhaps we should avoid adding namespaces to methods. Example:
```
{"cv::dnn::Layer::forward": [{"name": "cv2.dnn.forward" ...
```

`_create` had special processing. They can wrap "static" method "create()":
```
{"cv::ml::ParamGrid::create": [{"name": "cv2.ml.ParamGrid_create"
```
or wrap class constructor (it is buggy sometimes yet):
```
{"cv::KeyPoint::cv::KeyPoint::KeyPoint": [{"name": "cv2.KeyPoint" ...
```


There is attempt to eliminate `_create` functions (use ctor directly): #8934
So I believe we don't need "_create" suffix.

Cartucho:
@alalek is this correct? `cv2.dnn.forward`
Or do we need to create the net first and only then do the forward? I couldn't find any docs on this.

But it makes sense to have the namespace in the methods for example:
`cv::dnn::blobFromImage` -> `cv2.dnn.blobFromImage`


How do you want to proceed? We can either:

a) ignore for now the specific modules (dnn, ...) and then add them one by one checking if everything is alright.

b) try to figure out a way to cover all the specific modules corner cases

alalek:
> cv2.dnn.forward

No, `forward` is a [method name](https://docs.opencv.org/master/d3/d6c/classcv_1_1dnn_1_1Layer.html#a73b2805c23c1ed07dea8e6a99d96feaf) of `Layer`. So it is probably to keep it without any namespace (can't be used without class object: usage is `mylayer.forward(...)`).

`blobFromImage` is function, so namespace is applicable here.

Cartucho:
@alalek I understand now! I will remove the namespace from methods!

Cartucho:
@alalek I removed the namespace from methods. But, for example, the method `fourcc` (now in the .json file) should be `cv2.VideoWriter_fourcc()`.

alalek:
I believe because `fourcc()` is "static" method, right?

Cartucho:
@alalek from the brief:

VideoWriter_fourcc(c1, c2, c3, c4) -> retval

This static method constructs the fourcc code of the codec to be used in the constructor VideoWriter::VideoWriter or VideoWriter::open

alalek:
For static methods it is make sense to provide information in this format:
```
<namespace>.<class_name>_<method_name> = cv2.VideoWriter_fourcc
```

Cartucho:
@alalek How do you suggest me to check if it's static?
When I print the proto from line 624 they all appear to be static:

image

Cartucho:
@alalek I checked and I believe all the methods are static, i.e:

HOGDescriptor_getDaimlerPeopleDetector()
HOGDescriptor_getDefaultPeopleDetector()
VideoWriter_fourcc(c1, c2, c3, c4)
CascadeClassifier_convert(...)
SparsePyrLKOpticalFlow_create(winSize=None, ...)
...

alalek:
We need to check for "/S" attribute:
https://github.com/opencv/opencv/blob/3.3.0/modules/python/src2/hdr_parser.py#L360-L361

Cartucho:
@alalek actually there is a parameter to see if it's static (isclassmethod):

https://github.com/opencv/opencv/blob/3.3.0/modules/python/src2/gen2.py#L846-L848

For users to better understand that their dealing with a class object I added a my<Class> before, what do you think?

image

alalek:
> I added a my before

I don't understand. These static methods are already a part of Python bindings interface. No need to rename this.

```
>>> cv2.ml.TrainData_create
<built-in function TrainData_create>
```

Cartucho:
@alalek 
I gave you the wrong example.

I meant I wanted to add the my before the non static methods.
For example: `mylayer.forward(...)` instead of writing just `forward`.

According to these lines all the methods are static:
https://github.com/opencv/opencv/blob/3.3.0/modules/python/src2/gen2.py#L846-L848

Cartucho:
@alalek I commited the most recent version.

I don't know why some "regular methods" from sub-modules. For example the `objLayer.forward` are not showing up

alalek:
I have pushed changes:
- proper handling for constructors, classmethods (static methods), regular methods
- fix methods C++ names (`cv::ml::TrainData::cv::ml::TrainData`)
- removed magic values: bindings generator doesn't know: 'ml', 'flann' and even 'cv2'
- removed magic translations for constant names (reuse existed variables instead)
- removed evaluation of constants (it is useless thing: it is better to know that `ACCESS_RW` is `ACCESS_READ | ACCESS_WRITE` than some number)

Cartucho:
@alalek 
Just one thing, now it is generating for example 4 functions with the same name (`obj.load`), so if we could add the class as well (`obj<Classname>.load`) would be better for distinguishing. Do you agree?

alalek:
> now it is generating for example 4 functions with the same name (obj.load)

This is Python part, it is not required to be unique. We should map C++ entities => Python entities. I don't expect reverse mapping at this moment. So probably there is no serious problem.
Also there is no direct access to class methods without pre-created object instances.

Did you have a chance to take a look on opencv.tag file from doxygen?

Cartucho:
@alalek Sorry, I've been really busy! I will get back to this this weekend, what's left to do?

alalek:
I believe we can merge this patch and use current results in #9260 (check for more details).

PR #9742, opened 2017-09-28
Fix wrong power of teta in fisheye distortPoints

D-Alex:
The fisheye model uses only power of teta devidable by two

### DIFF
--- a/modules/calib3d/src/fisheye.cpp
+++ b/modules/calib3d/src/fisheye.cpp
@@ -294,10 +294,10 @@ void cv::fisheye::distortPoints(InputArray undistorted, OutputArray distorted, I
         // Angle of the incoming ray:
         double theta = atan(r);
 
-        double theta2 = theta*theta, theta3 = theta2*theta, theta4 = theta2*theta2, theta5 = theta4*theta,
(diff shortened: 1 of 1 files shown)


alalek:
Please add article references.
What is about (6) from [here](http://www.ee.oulu.fi/~jkannala/publications/tpami2006.pdf)?

D-Alex:
[opencv](http://docs.opencv.org/3.3.0/db/d58/group__calib3d__fisheye.html)
[github opencv](https://github.com/opencv/opencv/blob/a6d634b5b8b57408a0908eeb9e64ae78ab29d758/modules/calib3d/src/fisheye.cpp#L394)

There is an inconsistency and I would assume because teta is always positive there is no much different as long as it is consistent across all functions.

alalek:
> theta = theta_d / (1 + k[0] * theta2 + k[1] * theta4 + k[2] * theta6 + k[3] * theta8);

This is iterative solver for this:
```
(1 + k[0] * theta2 + k[1] * theta4 + k[2] * theta6 + k[3] * theta8) * theta = theta_d
```

Still odd power numbers.

D-Alex:
Fair - I probably got biased by this paper which does not change the power for the inverse function for none fish lenses http://ieeexplore.ieee.org/document/879788/ (eq. 12).

Thx for catching it.

PR #9743, opened 2017-09-28
TBB Debug Release

bradfordbarr:
### This pullrequest changes

OpenCV fails to detect tbb on a debug build if the platform has only tbb debug libraries.  This PR adds an additional check to the tbb detect logic for systems that only install tbb debug.


alalek:
Thank you for contribution!
:+1:

PR #9754, opened 2017-09-30
Update OpenCVFindLibsGrfmt.cmake

sturkmen72:
Updates OpenCVFindLibsGrfmt.cmake


sturkmen72:
@alalek i am not sure this is correct but i faced a problem when compiling with GDAL support and this fix solved the problem.

vpisarev:
@alalek, could you please review it?

alalek:
This looks very strange: patch changes code for case with disabled (not-available) GDAL, so it should not have any effect on code path which "enables" GDAL.

Did you tried to build OpenCV with clean CMake cache (clean build directory) without this patch?

sturkmen72:
ok let me try again and inform

sturkmen72:
image
@alalek  i tried again with clean built dir. without patch i can't get the result above.

alalek:
Well, I understand that you want. :+1:

BTW, In case of "missing" lines in cmake-gui, "Add entry" button should help. With command-line it is more easy =)

sturkmen72:
@alalek thanks.

 just a question. i compiled GDAL what should i do for enabling `cmake  find_package` find it ?

alalek:
It depends on package. I don't have GDAL build on Windows.
Replace "QUIET"/"REQUIRED" to receive verbose messages from CMake / "find script":
```.diff
-find_package(GDAL QUIET)
+find_package(GDAL REQUIRED)
```

sturkmen72:
for example i set environment variable OPENCV_DIR and cmake finds OpenCV. but i don't know what to do for other libs like GDAL , leptonica, tessseract etc.
is there a general rule about this

alalek:
`GDAL_DIR` would help only if you have `GDALConfig.cmake` somewhere in your build dir.
Perhaps you should have FindGDAL.cmake file in CMake folder - refer to it contents or [documentation](https://cmake.org/cmake/help/v3.9/module/FindGDAL.html).
"leptonica, tessseract" tutorial is [here](http://docs.opencv.org/master/db/d4c/tutorial_install_tesseract.html)

PR #6561, opened 2016-05-20
fix Core_MulSpectrum test on 64bit ARM + release mode

tomoaki0705:
resolves 6536

### What does this PR change?
  * fix Core_MulSpectrum test on 64bit ARM + release mode  (issue #6536)
  * increase threshold in particular tests due to the difference of fmsub and fsub

### the input
  - For the input, I didn't change, since only this modification was sufficient for the issue 6536

PR #9758, opened 2017-10-02
core: Increased maximum abs_error for hal_intrin.float* tests

barkovv:
### This pullrequest changes
This commit fixes failing of tests on PowerPC platform.

The maximum error of floating point comparations on PCC64 platform is 1.5259999999961416E-05 for hal_intrin.float32x4 test. This is a bit greater than standard allowed deviation for gtest's float comparations (1.0E-05, according to float.h man). This pull request affects PPC64 only since the other architectures fit the 1.0E-05 abs_error value.

### DIFF
--- a/modules/core/test/test_intrin_utils.hpp
+++ b/modules/core/test/test_intrin_utils.hpp
@@ -156,15 +156,19 @@ template <typename R> std::ostream & operator<<(std::ostream & out, const Data<R
     return out;
 }
 
+#ifndef EXPECT_COMPARE_EQ_FLOAT_ABS_ERROR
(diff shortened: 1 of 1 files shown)


tomoaki0705:
I went through this before (#6561)
If you are talking a specific test on ```ppc64```, I recommned you to follow the same way.
Changing the behaviour of ```EXPECT_COMPARE_EQ_``` effect too large compared to the goal you need.

Does **every** test from ```v_float32x4``` fails?

alalek:
@barkovv Please guard introduced changes via `#if` with some platform-related macro (`#if __ppc64__` or similar) and leave existed checks untouched.

barkovv:
This problem is resolved in #9763 .  So theese changed are not needed. Closing.

Issue #9224, opened 2017-07-24
derivAperture, histogramNormType not used in HOGDescriptor


gadcam:
OpenCV => 3.2 & master

If I am right <code>derivAperture</code> and <code>histogramNormType</code> parameters are not used in the <code>HOGDescriptor</code> class.
There are references of them in [hog.cpp](https://github.com/opencv/opencv/blob/master/modules/objdetect/src/hog.cpp) and [objdetect.hpp](https://github.com/opencv/opencv/blob/master/modules/objdetect/include/opencv2/objdetect.hpp).
Are they here for future development or should they be removed ?
In the first case I think something should be added to the documentation to explain their future role and that they are not currently used.

rdabane:
I'm also looking for clarification on this.
Also  not sure about exact normalization method is used(L1, L1-sqrt, L2-norm etc)

gadcam:
@rdabane These parameters are not used since years so I offered to remove them in #9767.

The L2-Norm is used : 

- If you check the code you will see there are a lot of references to [L2-Norm](https://github.com/opencv/opencv/blob/2a2537725acb483bcd82628edc17c18e7128d414/modules/objdetect/include/opencv2/objdetect.hpp#L608)
- The normalization procedure is implemented in [normalizeBlockHistogram](https://github.com/opencv/opencv/blob/master/modules/objdetect/src/hog.cpp#L1120)
- You can check it by looking at the code between [lines 1146 and 1164](https://github.com/opencv/opencv/blob/master/modules/objdetect/src/hog.cpp#L1146)

There is another thing that is not clear : what is the role of <code>free_coef</code>? It is not documented.

PR #9767, opened 2017-10-03
Cleanup HOGDescriptor

gadcam:
Before this pull request I opened #9224 and no-one could find what these parameters should do in two months.
Remove unused parameters since several years <code>derivAperture</code> and <code>histogramNormType</code> (it was not used in ["atomic bomb" commit](https://github.com/opencv/opencv/blob/127d6649a1c83397bf42799ac807af41aa507b30/modules/objdetect/src/hog.cpp)).

### DIFF
--- a/modules/js/test/test_objdetect.js
+++ b/modules/js/test/test_objdetect.js
@@ -145,9 +145,7 @@ QUnit.test('Cascade classification', function(assert) {
         assert.equal(hog.winSize.height, 128);
         assert.equal(hog.winSize.width, 64);
         assert.equal(hog.nbins, 9);
-        assert.equal(hog.derivAperture, 1);
         assert.equal(hog.winSigma, -1);
-        assert.equal(hog.histogramNormType, 0);
         assert.equal(hog.nlevels, 64);
 
         hog.nlevels = 32;
--- a/modules/objdetect/include/opencv2/objdetect.hpp
+++ b/modules/objdetect/include/opencv2/objdetect.hpp
@@ -372,8 +372,6 @@ http://www.learnopencv.com/handwritten-digits-classification-an-opencv-c-python-
 struct CV_EXPORTS_W HOGDescriptor
 {
 public:
-    enum { L2Hys = 0 //!< Default histogramNormType
-         };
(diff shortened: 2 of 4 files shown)


gadcam:
The only build to fail was <code>Linux x64</code> with <code>Failed compare abi dumps</code> but it is a normal effect of my pull request because I removed two parameters from <code>HOGDescriptor</code>.
What should I do now ?

sovrasov:
We can't change signatures of public methods and add/delete members of classes, because OpenCV preserves binary compatibility between intermediate releases. This PR can't be accepted before the next major OpenCV release. The best you can do now is to add a note to the documentation.

gadcam:
Ok I did not about that, thank you for the explanation.
In the meantime could someone review it and, if it is mergeable, tag it as mergeable for the next release ? So I could make the necessary changes if needed.

sturkmen72:
in my opinion you can do `histogramNormType `and `derivAperture `variables functional instead of removing them. i have no idea about `derivAperture `but for `histogramNormType ` you can add other NormTypes.

gadcam:
I am neither familiar with C++ nor with "SSE instructions" so I am not able to do that today. If someone has the skills and the time to do that I would pleased.
However it is like this since 2010 at least so do we really need that ?

sovrasov:
@gadcam the changes are ok, this patch can wait for the next big release.

vpisarev:
well, nowadays deep learning-based methods are used instead of HOG and they are almost as fast (and with specialized hardware they will become even faster). So let's consider HOG algorithm as legacy and therefore keep everything as-is. I suggest to just update the parameter descriptions that histogramNormType and derivAperture are currently not used. That's it.

vpisarev:
let me close it

gadcam:
@vpisarev I understand that you do not want to put more work into methods which are no more the state of the art. However, I do not understand why you do not want to merge a cleaning pull request like this one : these methods can be useful for students but also good enough in some situation for teams that can not afford to use a deep learning approach. There is almost no cost to merge it in the next big release and you can save time for people using the library. (You can also consider it as a gateway to use OpenCV for beginners)

PR #9773, opened 2017-10-04
Configuring Doxygen to generate a tag file.

shoeffner:
I added tag file generation for Doxygen so that I can reference OpenCV from my own project documentations without having my own generated tag file lying around in my repository, but instead can just download it from OpenCV's documentation in the future.

### This pullrequest changes

It configures Doxygen to generate a tag file inside the documentation's html root directory.

Additionally it adds a tutorial on how to use the tag file in other projects to link back to the OpenCV documentation. The tutorial is the only place which links to the tag file as of now, which is why it was added.

For more information on tag files and external documentation, check
https://www.stack.nl/~dimitri/doxygen/manual/external.html


mshabunin:
:+1:

PR #9776, opened 2017-10-04
Fix memory leak in the matrix copying constructor

### LINKED ISSUE #9732: Memory leak in rowRange ?

- OpenCV => :grey_question:
- Compiler => :grey_question:

##### Detailed description

If the final rowRange parameter is less than then "start rowRange" an exception occurs (in the constructor of the Mat object). For this, not properly created Mat object destructor is not called. But refCount is already increased. As a result, we get a memory leak, see sample code below. Is it a bug or desired behavior? Thanks. 

##### Steps to reproduce

```
  for (int k = 0; k < 500; k++)
  {
    try
    {
      std::cout << k << std::endl;
      Sleep(200);
      cv::Mat sourceMat(2000, 2000, CV_8UC1,cv::Scalar(0));
      cv::Mat copyheader = sourceMat.rowRange(19, 9);

    }
    catch (...)
    {
      std::cout << "error " << k << std::endl;

    }
  }    
```

IgWod:
### This pullrequest changes

As reported in the #9732 issue exception thrown inside the constructor cased a memory leak. This pull request adds a workaround that fixes this issue. I applied a workaround suggested in the issue. But as it was mentioned it's not fully robust solution and probably should be updated in the future. However I believe for now it's enough.

I looked for more issues like that in this file, but I haven't found any more. All other CV_Assert seem to be called beofre refcount is increased or no refcount incrementation is invloved.

### DIFF
--- a/modules/core/src/matrix.cpp
+++ b/modules/core/src/matrix.cpp
@@ -279,21 +279,31 @@ Mat::Mat(const Mat& m, const Range& _rowRange, const Range& _colRange) : size(&r
     }
 
     *this = m;
-    if( _rowRange != Range::all() && _rowRange != Range(0,rows) )
+    try
     {
-        CV_Assert( 0 <= _rowRange.start && _rowRange.start <= _rowRange.end && _rowRange.end <= m.rows );
-        rows = _rowRange.size();
-        data += step*_rowRange.start;
-        flags |= SUBMATRIX_FLAG;
-    }
+        if( _rowRange != Range::all() && _rowRange != Range(0,rows) )
+        {
+            CV_Assert( 0 <= _rowRange.start && _rowRange.start <= _rowRange.end
+                       && _rowRange.end <= m.rows );
+            rows = _rowRange.size();
+            data += step*_rowRange.start;
+            flags |= SUBMATRIX_FLAG;
+        }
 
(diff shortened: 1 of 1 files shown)


modules/core/src/matrix.cpp
```diff
@@ -279,21 +279,31 @@ Mat::Mat(const Mat& m, const Range& _rowRange, const Range& _colRange) : size(&r
...
-        CV_Assert( 0 <= _rowRange.start && _rowRange.start <= _rowRange.end && _rowRange.end <= m.rows );
-        rows = _rowRange.size();
-        data += step*_rowRange.start;
-        flags |= SUBMATRIX_FLAG;
-    }
+        if( _rowRange != Range::all() && _rowRange != Range(0,rows) )
+        {
+                CV_Assert( 0 <= _rowRange.start && _rowRange.start <= _rowRange.end
+                           && _rowRange.end <= m.rows );
+                rows = _rowRange.size();
```
alalek: Please keep indentation 4 spaces.

IgWod:
I fixed the indentation and amended the existing commit to avoid an extra commit that only fixes some white characters.

I see this PR was labeled with "port into master is needed". Should I cherry pick my change onto master and submit another pull request or is it done by you?

vpisarev:
looks good! :+1:

PR #9778, opened 2017-10-05
Deep learning colorization model

dkurt:
### This pullrequest changes
Colorization model from https://github.com/richzhang/colorization into DNN.

Source | Gray | Colorized
----------|--------------|------------
 |  |  |
 |  | 
 |  | 
 |  | 
 |  |

### DIFF
--- a/modules/dnn/src/layers/scale_layer.cpp
+++ b/modules/dnn/src/layers/scale_layer.cpp
@@ -33,6 +33,7 @@ public:
                          std::vector<MatShape> &outputs,
                          std::vector<MatShape> &internals) const
     {
+        CV_Assert(blobs.size() == 1 + hasBias);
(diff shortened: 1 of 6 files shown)


modules/dnn/src/layers/scale_layer.cpp
```diff
@@ -26,17 +26,37 @@ class ScaleLayerImpl : public ScaleLayer
     {
         setParamsFrom(params);
         hasBias = params.get<bool>("bias_term", false);
+
+        if (blobs.empty())
```
dkurt: We can add an extra layers in `.prototxt` that were not used during training. So they have no weights. In this model it's a Scale layer.

vpisarev:
very good! :+1:

PR #9779, opened 2017-10-05
videoio: remove AssetsLibrary dependency.

StatusReport:
### This pullrequest changes

The entire AssetsLibrary framework is deprecated since iOS 8.0. The code
used in the camera example code can use UIKit to save videos to the
camera instead, which allows to avoid linking with PhotoKit instead to
prevent increasing the iOS deployment target.


vpisarev:
:+1:

PR #9790, opened 2017-10-06
Small fix for using Qt 5.x with opencv_world

mevatron:


alalek:
Thank you for contribution!
Fix works :+1:

PR #9792, opened 2017-10-07
Port into master 9776

### LINKED ISSUE #9732: Memory leak in rowRange ?

alalek:
#9776 Fix a memory leak in the Mat copying constructor
resolves #9732

### DIFF
--- a/modules/core/src/matrix.cpp
+++ b/modules/core/src/matrix.cpp
@@ -484,21 +484,31 @@ Mat::Mat(const Mat& m, const Range& _rowRange, const Range& _colRange)
     }
 
     *this = m;
-    if( _rowRange != Range::all() && _rowRange != Range(0,rows) )
+    try
     {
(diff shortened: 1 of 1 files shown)


vpisarev:
:+1:

PR #9795, opened 2017-10-07
Extract code from scalarToRawData

IgWod:
### This pullrequest changes

This pull request extracts code that is repeated for many cases with only one difference. Use of templated function is more suitable in this case as it improves readability and maintainability of the code.

It's not an important change, but I believe it increases quality of the code and makes the code less prone for bugs as if someone wants to change a behaviour of the function it has to only be done in one place and it's also easier to add new cases.


sovrasov:
:+1:

PR #9796, opened 2017-10-08
in python, false should be False

lzx1413:
### This pullrequest changes
in samples/dnn/mobilenet_ssd_python.py, false should be False

### DIFF
--- a/samples/dnn/googlenet_python.py
+++ b/samples/dnn/googlenet_python.py
@@ -11,7 +11,7 @@ def get_class_list():
     with open('synset_words.txt', 'rt') as f:
         return [x[x.find(" ") + 1:] for x in f]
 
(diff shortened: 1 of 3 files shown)


berak:
yea, that's `True`, hehe (nice catch !)

and don't get scared from the buildbot failure, 

it's an internal infrastructure failure,  not related to your commit **at all**.

again, no need to do anything. just wait. the bot will restart (and resolve this).

alalek:
I added additional similar fixes.
BTW, Problem was introduced in this [commit](https://github.com/opencv/opencv/commit/c67ad4937846ca026257fcf529d3c93678beaefa).

alalek:
@lzx1413 Thank you for contribution! :+1:

PR #9798, opened 2017-10-08
dnn: fix_resnet_ssd_face_python.py_sample, too !

berak:
related to: #9796

another sample to fix


berak:
ah wait, based on wrong (local) branch

alalek:
@berak Thank you for contribution!
Actually I have pushed similar changes for additional two `.py` files into original patch, so additional fixes are not required.

berak:
yea, again, you were faster than me, simply ;)

PR #9802, opened 2017-10-09
minEnclosingCircle from three point. Issue #9797

### LINKED ISSUE #9797: calculating minEnclosingCircle from three point.
##### Detailed description

there is a problem when calculating minEnclosingCircle from three point.
yellow circle calculated with three point
green circle calculated with four point
minenclosingcircle

##### Steps to reproduce

```

#include <iostream>

using namespace cv;
using namespace std;

int main(int /*argc*/, char** /*argv*/)
{

    Mat img(400, 400, CV_8UC3,Scalar(0,0,0));

    vector<Point> points;
    points.push_back(Point(200, 100));
    points.push_back(Point(240, 150));
    points.push_back(Point(200, 250));

    // Find the minimum area enclosing circle
    Point2f center;
    float radius = 0;
    minEnclosingCircle(points, center, radius);

    // Draw the circle
    circle(img, center, cvRound(radius), Scalar(0, 255, 255), 1, LINE_AA);

    points.push_back(Point(230, 190));
    minEnclosingCircle(points, center, radius);

    // Draw the points
    for (int i = 0; i < 4; i++)
        circle(img, points[i], 3, Scalar(0, 0, 255), FILLED, LINE_AA);

    // Draw the circle
    circle(img, center, cvRound(radius), Scalar(0, 255, 0), 1, LINE_AA);

    imshow("minEnclosingCircle", img);
    waitKey();
    return 0;
}
```

Nickolays:
resolves #9797 
### Problem

If you give 3 points to the minEnclosingCircle function, it checks two vectors (from these points) for colinearity. And if they are colinear, they will find the maximum distance and take it for a diameter. Otherwise, construct a circle around three points. But this is not true. if we take points almost on the same line, then the vectors will not be collinear and the function will calculate a large circle.

### This pullrequest changes

Find the maximum distance and take it for a diameter. if the other point (not the one that made up the diameter) is inside the circle, then everything is fine. Otherwise, construct a circle around three points.

P.S. I can do tests if necessary
Sorry, the first time I did not change the branch

### DIFF
--- a/modules/imgproc/src/shapedescr.cpp
+++ b/modules/imgproc/src/shapedescr.cpp
@@ -43,11 +43,7 @@
 namespace cv
 {
 
-// inner product
-static float innerProduct(Point2f &v1, Point2f &v2)
-{
-    return v1.x * v2.y - v1.y * v2.x;
-}
+const float EPS = 1.0e-4f;
 
 static void findCircle3pts(Point2f *pts, Point2f &center, float &radius)
 {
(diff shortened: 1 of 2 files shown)


vpisarev:
@Nickolays, thank you! Can you please refactor your code - combine `findCircle3pts` and `findEnclosingCircle3pts` into a single function?

terfendail:
@Nickolays, it looks like collinearity check performed by `innerProduct(...)` is used to ensure the determinant(which is exactly `v1.x * v2.y - v1.y * v2.x`) is not zero. Though I can't say such a check is necessary there still is a small reason to retain it.
Also it looks like the maximum distance selection for 3 points could be performed by the generic branch of the code implemented by `findMinEnclosingCircle(...)`, `findSecondPoint(...)` and `findThirdPoint(...)`.
Could you please check whether it would solve the issue if the generic branch will be called for 3 points case as well?

Nickolays:
@terfendail, Ok I save innerProduct. You are absolutely right. If the generic branch will be called for 3 points, the code will work correctly. So I delete `findCircle3pts(...)`, `findEnclosingCircle3pts(...)`. How can I edit pull reguest? Or I need close it, and open again?

alalek:
> How can I edit pull reguest? Or I need close it, and open again?

No need to close PR for new updates. Refer to GitHub help for details.

Nickolays:
In `test_convhull.cpp`,`int CV_MinCircleTest::validate_test_results( int test_case_idx )`(965). There is a condition for checking the point belongs to the circle (the distance from the center is not more than the radius), but all the calculations are done in float, so they are not very accurate, I replaced the data type by double in the intermediate computations in `void cv :: minEnclosingCircle(...)`(and his descendants) . This increased the accuracy, but in some tests, the situation can still happen when ` 0 <distance minus the radius <1`. Therefore, I suggest simply formatting the code as said @vpisarev, because otherwise the real numbers accumulate inaccuracy. Do you agree?

terfendail:
@Nickolays 
The root cause of the test failure is related to difference in radius evaluation performed by `findEnclosingCircle3pts_orLess_32f` and `findCircle3pts`. The latter evaluate exact circle radius, while the former get the exact circle radius and then increase it by `EPS` to ensure all points fall into the circle regardless of the computation precision.
I can't say for sure which version of the result could be considered better or more convenient but the evaluation is already implemented the way it is so IMHO it would be better to retain original behavior with `EPS` increased radius.

Nickolays:
Yes, I'm forgot to add `radius += EPS;`

Bhanudutta:
Why don't retain them as 2 separate functions and allow user to decide
which to use according to the conditions , or better test both of the
function with few test cases and use the functions which gives better
results

I am new to openCV so forgive me if any mistake

On Oct 10, 2017 5:19 PM, "Nickolays" <notifications@github.com> wrote:

Yes, I'm forgot to add radius += EPS;

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/opencv/opencv/pull/9802#issuecomment-335447195>, or mute
the thread

modules/imgproc/src/shapedescr.cpp
```diff
@@ -43,64 +43,39 @@
 namespace cv
 {
 
-// inner product
+/* inner product
+unused
```
terfendail: I think it would be better to remove the unused code completely

Nickolays: Ok.

modules/imgproc/test/test_convhull.cpp
```diff
@@ -1017,6 +1017,71 @@ int CV_MinCircleTest::validate_test_results( int test_case_idx )
...
+}
+
+int CV_MinCircleTest2::validate_test_results( int test_case_idx )
+{
+    float eps = 1.0F;
+    int code = CV_BaseShapeDescrTest::validate_test_results( test_case_idx );
+    Point2f calcCenter;
+    float calcRadius;
+    // Find the minimum area enclosing circle
+    minEnclosingCircle(points, calcCenter, calcRadius);
```
terfendail: While the test designed such way could detect errors IMHO it's more obvious(and at least more common) to perform result evaluation in `run_func()` function and just verify the result quality in `validate_test_results()`.
Also added test reverify once again the generic case though the original issue is related to specific 3-points case.
Could you please add test or tests to verify specific cases with less than 4 points?

Nickolays: Ok, I will edit this. But I didn't understand. Do you want me to change this test so it works only for 3 points or less, or do you want me to add one more?

Nickolays: And why do we need to test the case with 2 and 1 points? Maybe it's worth the test for only 3?

Nickolays:
Maybe it is time to merge?

terfendail:
👍

Nickolays:
Why nobody merge it?

PR #9803, opened 2017-10-09
ocl: simplify ocl::Timer

wzw-intel:
Use clFinish to gurantee commands completed, instead of waiting for events.


alalek:
Thank you!
Looks good to me :+1:

PR #9804, opened 2017-10-09
Use Eigen::SelfAdjointEigenSolver in cv::eigen

woodychow:
CPU: Intel i7 6700k
Eigen version: 3.3.2
Memory: 16GB

With master
Core_Eigen.vector_32: ~1024ms
Core_Eigen.vector_64: ~1512ms

With my branch
Core_Eigen.vector_32: ~62ms
Core_Eigen.vector_64: ~91ms

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/core/src/lapack.cpp
+++ b/modules/core/src/lapack.cpp
@@ -43,6 +43,12 @@
 #include "precomp.hpp"
 #include <limits>
 
+#ifdef HAVE_EIGEN
+#include <Eigen/Core>
+#include <Eigen/Eigenvalues>
+#include "opencv2/core/eigen.hpp"
+#endif
+
 #if defined _M_IX86 && defined _MSC_VER && _MSC_VER < 1700
 #pragma float_control(precise, on)
 #endif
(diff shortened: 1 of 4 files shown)


woodychow:
Automated tests use Eigen 3.2.0. With Eigen 3.3.2, accuracy of Eigen::SelfAdjointEigenSolver is higher, and only Core_PCA.accuracy fails; threshold is 1e-6 and the error is 1.2e-6

SelfAdjointEigenSolver is already the most accurate eigensolver provided by the Eigen library according to this page.
https://eigen.tuxfamily.org/dox/group__TopicLinearAlgebraDecompositions.html

modules/core/src/lapack.cpp
```diff
@@ -1396,6 +1405,47 @@ bool cv::eigen( InputArray _src, OutputArray _evals, OutputArray _evects )
         v = _evects.getMat();
     }
 
+#ifdef HAVE_EIGEN
```
alalek: If some eigen versions doesn't pass accuracy checks then you should avoid using of them here:
```
#if defined HAVE_EIGEN && \
    (EIGEN_MAJOR_VERSION > 3 || (EIGEN_MAJOR_VERSION == 3 && EIGEN_MINOR_VERSION >= 2))
```

modules/core/src/lapack.cpp
```diff
@@ -43,6 +43,15 @@
 #include "precomp.hpp"
 #include <limits>
 
+#ifdef HAVE_EIGEN
+#  if defined __GNUC__ && defined __APPLE__
+#    pragma GCC diagnostic ignored "-Wshadow"
+#  endif
```
alalek: Why we need to suppress this warning?
Perhaps related code can be refactored.

woodychow: OK let me look into this. I see this in every single include of Eigen in opencv; I just copied it. I also don't have a MAC.

modules/core/src/lapack.cpp
woodychow: When I made the changes, there was no Core_PCA test. With Eigen 3.3.2, this branch now passes all Core_Eigen tests but fail in Core_PCA.

So unless we increase the error threshold, there is no way to merge this for now.

alalek: Old code uses double type for internal calculations. Currently there are two code branches: `float` and `double`. Probably we should increase tolerance for float-based calculations or use double-only code.

woodychow: Alright. Please let me know if you decide to increase the tolerance. I don't think there is anything that I can do before that. Thanks

woodychow:
Thanks very much for updating the tests. I appreciate it.

PR #9816, opened 2017-10-09
fix StereoBM disparity map right margin truncation when minDisparities > 0

### LINKED ISSUE #9783: StereoBM truncates right edge of disparity map by minDisparities
- OpenCV => 3.3 (git tag 3.3.0)
- Operating System / Platform => Ubuntu 14.04 amd64
- Compiler => GNU CC 4.8.4

##### Detailed description

I found a problem with StereoBM computing disparity maps when minDisparities is larger than zero.

This problem is particularly frustrating for users needing to resolve disparity for distances relatively close to the two imagers, and needing the optimized performance of StereoBM.

Example: Cropped disparity image, input is half size aloe image, with minDisparities=32, SADWindowSize=17 and numberOfDisparities=96, showing right margin of **40**, when I expected it to be **8**:

When minDisparities=32, right margin is 40, not SADWindowSize/2=8

StereoBM's _compute_ function uses _getValidDisparityROI_ (see [line 1186ff. of modules/calib3d/src/stereobm.cpp](https://github.com/opencv/opencv/blob/3.3.0/modules/calib3d/src/stereobm.cpp#L1186)) to determine the region of the disparity map over which disparities shall be calculated. _getValidDisparityROI_ sets the right edge of the ROI at:

```.cpp
    x = width - max(minDisparities, 0) - SADWindowSize / 2
```

In contrast, the right margin of the StereoSGBM disparity map (and the contrib StereoBinaryBM/SGBM disparity map) is unaffected by different values of minDisparities. StereoBM should behave similar to the other algorithms, and calculate disparity out to:

```.cpp
    x = width - SADWindowSize / 2
```

This issue was [discussed](http://answers.opencv.org/question/175572/stereobm-truncates-right-edge-of-disparity-by-mindisparities/) with @LaurentBerger on the OpenCV Q&A site.

##### Steps to reproduce

The following example uses the [aloe images](https://github.com/opencv/opencv/blob/3.3.0/samples/data/aloeL.jpg) from samples/data.

The first command line argument configures StereoBM minimum disparity. Compare the output of the script run with arguments of 0 (minDisparities=0), 16, and 32. At 0 you'll see the margin at the right is 8 pixels (half the window size of 17) - _correct_, then at 16 and 32 it grows with the higher minDisparities setting - _error_.

```.py
    #!/usr/bin/env python
    # Python 2/3 compatibility
    import sys
    import numpy as np
    import cv2
    
    print('load and downscale images')
    imgL = cv2.pyrDown( cv2.imread('../data/aloeL.jpg') )
    imgR = cv2.pyrDown( cv2.imread('../data/aloeR.jpg') )
    
    min_disp = 16
    if sys.argv[1] > 0:
        min_disp = int(sys.argv[1])
    num_disp = 112 - min_disp
    window_size = 17
    
    stereo = cv2.StereoBM_create(numDisparities = num_disp, blockSize = window_size)
    stereo.setMinDisparity(min_disp)
    stereo.setNumDisparities(num_disp)
    stereo.setBlockSize(window_size)
    stereo.setDisp12MaxDiff(0)
    stereo.setUniquenessRatio(10)
    stereo.setSpeckleRange(32)
    stereo.setSpeckleWindowSize(100)
    
    print('compute disparity')
    grayL = cv2.cvtColor(imgL, cv2.COLOR_BGR2GRAY)
    grayR = cv2.cvtColor(imgR, cv2.COLOR_BGR2GRAY)
    disp = stereo.compute(grayL, grayR).astype(np.float32) / 16.0
    disp_map = (disp - min_disp)/num_disp
    
    print('display')
    cv2.imshow('left', imgL)
    cv2.imshow('disparity', disp_map)
    cv2.waitKey()
    cv2.destroyAllWindows()
```

C++ behaves similarly. See the Q&A link above for more details and images.

##### Corrective action

The likely problematic line of code is in _getValidDisparityROI_, [line 2279 of modules/calib3d/src/stereosgbm.cpp](https://github.com/opencv/opencv/blob/3.3.0/modules/calib3d/src/stereosgbm.cpp#L2279):

```.cpp
    int xmax = std::min(roi1.x + roi1.width, roi2.x + roi2.width - minD) - SW2;
```    

When it is corrected to this:

```.cpp
    int xmax = std::min(roi1.x + roi1.width, roi2.x + roi2.width) - SW2;
```

then the StereoBM disparity map shows valid data all the way out to x = width - SADWindowSize / 2.

opalmirror:
…s > 0

resolves #9783 for the opencv master branch

This pullrequest modifies StereoSGBM::getValidDisparityROI() to return the correct disparity map region of interest, the right edge of which should be at x = width - SADWindowSize / 2.

The major effect of this is that the StereoBM::compute algorithm, which calls this function, will correctly calculate disparity to SADWindowSize / 2 of the right margin of the disparity map, instead of only to minDisparities + SADWindowSize / 2.

This allows StereoBM to compute a disparity map covering an ROI roughly the same size as StereoSGBM and the StereoBinary algorithms do, instead of being reduced in width by minDisparities on the right edge.

### DIFF
--- a/modules/calib3d/src/stereosgbm.cpp
+++ b/modules/calib3d/src/stereosgbm.cpp
@@ -2273,10 +2273,10 @@ Rect getValidDisparityROI( Rect roi1, Rect roi2,
                           int SADWindowSize )
 {
     int SW2 = SADWindowSize/2;
-    int minD = minDisparity, maxD = minDisparity + numberOfDisparities - 1;
+    int maxD = minDisparity + numberOfDisparities - 1;
 
     int xmin = std::max(roi1.x, roi2.x + maxD) + SW2;
-    int xmax = std::min(roi1.x + roi1.width, roi2.x + roi2.width - minD) - SW2;
+    int xmax = std::min(roi1.x + roi1.width, roi2.x + roi2.width) - SW2;
     int ymin = std::max(roi1.y, roi2.y) + SW2;
     int ymax = std::min(roi1.y + roi1.height, roi2.y + roi2.height) - SW2;
 


vpisarev:
thanks! :+1:

opalmirror:
@vpisarev Thank you!

alalek:
BTW, It would be nice to have some tests for mentioned parameters set.

opalmirror:
@alalek Great idea, thanks for the encouragement. I'll prepare some tests for different positive values of minimum disparity and add them to the opencv_test_calib3d set of tests.

PR #9818, opened 2017-10-10
fix#9570: implement mat ptr for generic types

### LINKED ISSUE #9570: Mat::ptr not defined for more than 3 dimensions

- OpenCV =>  3.3, git repository as of 9/6/17
- Operating System / Platform => probably all, tested on Linux
- Compiler => probably all C++ compilers, tested with g++ 5.4

##### Detailed description
There is no definition for the templated `ptr` method declared in `core/mat.hpp`:
 ```.cpp
template<typename _Tp> const _Tp* ptr(const int* idx) const;
```
The definition is missing in `core/mat.inl.hpp` (where it probably belongs), and the example code below generates linker's error:
    `mat4.cpp:(.text+0x113): undefined reference to `float* cv::Mat::ptr<float>(int const*)'`

##### Steps to reproduce

Try building an executable for this source code:

```.cpp
#include <opencv2/core.hpp>
int main (int argc, const char *argv[])
{
  int dims4[4] = {2,3,4,5}, idx4[4];
  cv::Mat_<float> m4 (4, dims4, 4.5);
  idx4      [3] = 0;
  for     (int w=0; w<dims4[0]; w++) {
    idx4    [0] = w;
    for     (int z=0; z<dims4[0]; z++) {
      idx4  [1] = z;
      for   (int y=0; y<dims4[1]; y++) {
	idx4[2] = y;
	float* yline = m4.ptr<float>(idx4);
	for (int x=0; x<dims4[2]; x++)
	  yline[x] = x+y+z;
      }
    }
  }
  return 0;
}
```

tz70s:
resolves #9570 
### This pullrequest changes

The original template based mat ptr for indexing is not implemented,
add the similar implementation as uchar type, but cast to
user-defined type from the uchar pointer.

However, due to the discussion of #9570, not sure whether the `mat::ptr(const int *index)` will be deprecated or not?

### DIFF
--- a/modules/core/test/test_mat.cpp
+++ b/modules/core/test/test_mat.cpp
@@ -1804,4 +1804,24 @@ TEST(Mat_, from_initializer_list)
     ASSERT_DOUBLE_EQ(norm(A, B, NORM_INF), 0.);
 }
 
+
+TEST(Mat, template_based_ptr)
+{
+    Mat mat = (Mat_<float>(2, 2) << 11.0f, 22.0f, 33.0f, 44.0f);
+    int idx[2] = {1, 0};
+    ASSERT_FLOAT_EQ(33.0f, *(mat.ptr<float>(idx)));
+    idx[0] = 1;
+    idx[1] = 1;
+    ASSERT_FLOAT_EQ(44.0f, *(mat.ptr<float>(idx)));
+}
+
+TEST(Mat_, template_based_ptr)
+{
+    int dim[4] = {2, 2, 1, 2};
+    Mat_<float> mat = (Mat_<float>(4, dim) << 11.0f, 22.0f, 33.0f, 44.0f,
+                                              55.0f, 66.0f, 77.0f, 88.0f);
+    int idx[4] = {1, 0, 0, 1};
+    ASSERT_FLOAT_EQ(66.0f, *(mat.ptr<float>(idx)));
+}
+
 #endif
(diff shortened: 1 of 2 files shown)


alalek:
Thank you for contribution!
This is dangerous API without controlling of input data, so it should be used very carefully.
Anyway, tests are required (implementation doesn't look correct - step is in bytes, not elements).

modules/core/test/test_mat.cpp
```diff
@@ -1804,4 +1804,24 @@ TEST(Mat_, from_initializer_list)
...
 
+
+TEST(Mat, template_based_ptr)
+{
+    Mat mat = (Mat_<float>(2, 2) << 0.11f, 0.22f, 0.33f, 0.44f);
+    int idx[2] = {1, 0};
+    ASSERT_DOUBLE_EQ(0.33, *(mat.ptr<float>(idx)));
+    idx[0] = 1;
+    idx[1] = 1;
+    ASSERT_DOUBLE_EQ(0.44, *(mat.ptr<float>(idx)));
```
alalek: Perhaps we could use integers here to prevent floating-point rounding: `0.xyf` to `xy.0f` (`44.0f`)

tz70s: Oh yes, thanks!

tz70s:
@alalek , are the implementation and tests correct or anything else that I should improve? Thanks!

alalek:
Thank you! :+1:

PR #9819, opened 2017-10-10
Add video example for Face Detection in Video Capture tutorial

ganwenyao:
```
docker_image-Docs=docs-js
```

@huningxin I add video example for Face Detection in Video Capture tutorial


vpisarev:
@ganwenyao, thank you for the PR! I think, putting video clip to OpenCV tree is not very good for 2 reasons: 1) it increases the space occupied by OpenCV even more and 2) there are possible issues with IP. Can we instead use some web link to an externally hosted video instead?

huningxin:
@vpisarev , the new video file (Megamind.mp4) is actually converted from https://github.com/opencv/opencv/blob/master/samples/data/Megamind.avi. The avi (divx) video is not supported by web browsers, so we convert it to mp4 (h.264). 

So for your concerns, the original video is already in opencv source tree, we assume it doesn't have IP issue. For space increasing, are you OK we replace the original Megamind.avi to Megamind.mp4 and use it in js tutorials? We will change other tutorials using this video to new file, like https://github.com/opencv/opencv/blob/19270103ec21aa788f11f8fa0ea9b7ceff561957/doc/tutorials/videoio/video-input-psnr-ssim/video_input_psnr_ssim.markdown.

For the option to use external video source, I found some videos on wikimedia, like this one https://commons.wikimedia.org/wiki/File:George_Monbiot_interview_with_The_Green_Interview.webm:  and this one https://commons.wikimedia.org/wiki/File:Drill.ogv. It says it is free to use. But it also has warnings of personality right issues. I am not sure whether it is appropriate to use in js tutorials. 

What do you think?

huningxin:
And I think allowing users to try external videos in the live example would be a great feature. To allow that, we need to set `crossOrigin` attribute of video element to `anonymous`. 

@ganwenyao @vpisarev , thoughts?

alalek:
Source branch is removed. Closing...

PR #9820, opened 2017-10-10
dnn: modify priorBox layer

sovrasov:
Make sufficient changes in PriorBox layer to be able to run model from https://github.com/MhLiao/TextBoxes

Required for https://github.com/opencv/opencv_contrib/pull/1399


modules/dnn/src/caffe/caffe.proto
```diff
@@ -145,6 +145,8 @@ message PriorBoxParameter {
   optional float step_w = 12;
   // Offset to the top left corner of each cell.
   optional float offset = 13 [default = 0.5];
+  // If true, two additional boxex for each center will be generated. Their centers will be shifted by y coordinate.
```
vpisarev: 1) boxex =&gt; boxes;
2) is `additional_y_offset` name taken from the original patch?

sovrasov: The original [patch](https://github.com/MhLiao/TextBoxes/commit/f615ff6d929d55d8d97dc7c0ece0ca22eea34117) doesn't contain any parameters, all changes are hardcoded.

vpisarev:
:+1:

PR #9823, opened 2017-10-10
dnn(halide): don't compile Halide via parallel_for_()

alalek:
To avoid problem with reduced stack size of inner threads (TBB backend).

```
with_tbb=ON
```


vpisarev:
:+1:

PR #9828, opened 2017-10-11
fix c++17 namespace issues

### LINKED ISSUE #9572: Issues found when building with MSVC + /std:c++17
The C++ compiler team is currently doing daily RWC (Read World Code) testing which includes OpenCV.
We are also working with internal partners and the community to fix source issues exposed by "/std:c++17".

In C++17, STL adds two new functions, std::clamp and std::reduce. Because some source files in OpenCV use 'using namespace std' and OpenCV also has functions with the same name, they can conflict and cause errors.

There are two issues found (both can be fixed if you remove 'using namespace std' :-) ):
1. OpenCV\src\samples\tapi\tvl1_optical_flow.cpp(40): error C2668: 'clamp': ambiguous call to overloaded function

Possible fix:
Use ‘::clamp’ instead of ‘clamp’ below:

	template <typename T> inline T mapValue(T x, T a, T b, T c, T d)
	{
		x = clamp(x, a, b);
		return c + (d - c) * (x - a) / (b - a);
	}

2. OpenCV\src\modules\core\test\test_mat.cpp(316): error C2794: 'value_type': is not a member of any direct or indirect base class of 'std::iterator_traits<_InIt>'

Possible fix:
Replace ‘reduce’ with ‘cv::reduce’. Looks that 'cv::reduce' is already used elsewhere in this file.

berak:
resolves #9572
### This pullrequest changes

add explicit namespace prefixes to the offending cases.

### DIFF
--- a/modules/core/test/test_mat.cpp
+++ b/modules/core/test/test_mat.cpp
@@ -133,7 +133,7 @@ int Core_ReduceTest::checkOp( const Mat& src, int dstType, int opType, const Mat
 
     assert( opRes.type() == CV_64FC1 );
     Mat _dst, dst, diff;
-    reduce( src, _dst, dim, opType, dstType );
+    cv::reduce( src, _dst, dim, opType, dstType );
     _dst.convertTo( dst, CV_64FC1 );
 
     absdiff( opRes,dst,diff );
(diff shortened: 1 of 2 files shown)


vpisarev:
thank you! :+1:

PR #9829, opened 2017-10-11
dnn(ocl4dnn): support log softmax in ocl4dnn

pengli:
dnn(ocl4dnn): support log softmax in ocl4dnn

### DIFF
--- a/modules/dnn/src/ocl4dnn/include/ocl4dnn.hpp
+++ b/modules/dnn/src/ocl4dnn/include/ocl4dnn.hpp
@@ -445,11 +445,12 @@ class OCL4DNNLRN
 
 struct OCL4DNNSoftmaxConfig
 {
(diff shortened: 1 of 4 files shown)


modules/dnn/src/ocl4dnn/include/ocl4dnn.hpp
```diff
@@ -450,6 +450,7 @@ struct OCL4DNNSoftmaxConfig
     MatShape in_shape;
     int axis;
     int channels;
+    bool logsoftmax;
```
alalek: Please initialize members in ctor to make happy static code analyzers.

pengli: Thanks for the review,  added member initialization in the constructor

alalek:
:+1:

PR #9834, opened 2017-10-11
MediaSDK video backend: Windows support

mshabunin:
### This pullrequest changes

Adds support for Windows version of MediaSDK. Specific `DeviceHandler` instance is empty, internal MediaSDK DirectX device will be used. 

Several warnings have been fixed.

### DIFF
--- a/modules/videoio/src/cap_mfx_common.hpp
+++ b/modules/videoio/src/cap_mfx_common.hpp
@@ -168,7 +168,7 @@ inline void cleanup(T * &ptr)
 
 //==================================================================================================
 
-struct Plugin
+class Plugin
 {
 public:
     static Plugin * loadEncoderPlugin(MFXVideoSession &session, mfxU32 codecId)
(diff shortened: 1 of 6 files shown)


modules/videoio/src/cap_mfx_common.hpp
```diff
@@ -302,7 +303,25 @@ class VAHandle : public DeviceHandler {
     int file;
 };
 
-// TODO: Windows specific
+#endif
+
+// Windows specific
+#ifdef _MSC_VER
```
alalek: The same problem: detection of platform via compiler is wrong.
`_WIN32` should be here.

modules/videoio/src/cap_mfx_common.hpp
```diff
@@ -302,7 +303,25 @@ class VAHandle : public DeviceHandler {
...
+
+#include <Windows.h>
+inline void sleep(unsigned long msec) { Sleep(msec); }
```
alalek: It is not "msec" in Unix: https://linux.die.net/man/3/sleep

mshabunin:
Fixed.
Thanks for comments.

vpisarev:
looks good to me! :+1:

PR #9835, opened 2017-10-11
dnn: add crop flag to blobFromImage

sovrasov:
Some models require direct resize instead of cropping on preprocessing stage, so corresponding option was introduced into `blobFromImage` function.


vpisarev:
thanks! :+1:

Issue #9622, opened 2017-09-13
Compilation and linking errors with Intel IPP/MKL 2018


sergiud:
##### System information (version)

- OpenCV => 3.3-dev (31348f8)
- Operating System / Platform => Windows 10
- Compiler => Visual Studio 15 2017

##### Detailed description

OpenCV does not compile with the standalone IPP and MKL 2018 (initial release). Some functionality in IPP IW has been either renamed or removed making it incompatible with previous releases. There are also linking errors related to MKL CBLAS:

```
"D:\Projects\opencv\build-vc141-x64\ALL_BUILD.vcxproj" (default target) (1) ->
"D:\Projects\opencv\build-vc141-x64\apps\annotation\opencv_annotation.vcxproj" (default target) (12) ->
"D:\Projects\opencv\build-vc141-x64\modules\core\opencv_core.vcxproj" (default target) (13) ->
(Link target) ->
  hal_internal.obj : error LNK2019: unresolved external symbol cblas_sgemm referenced in function "int __cdecl lapack_gemm<float>(float const *,unsigned __int64,float const *,unsigned __int64,float,float const *,unsigned __int64,float,f
loat *,unsigned __int64,int,int,int,int)" (??$lapack_gemm@M@@YAHPEBM_K01M01MPEAM1HHHH@Z) [D:\Projects\opencv\build-vc141-x64\modules\core\opencv_core.vcxproj]
  hal_internal.obj : error LNK2019: unresolved external symbol cblas_dgemm referenced in function "int __cdecl lapack_gemm<float>(float const *,unsigned __int64,float const *,unsigned __int64,float,float const *,unsigned __int64,float,f
loat *,unsigned __int64,int,int,int,int)" (??$lapack_gemm@M@@YAHPEBM_K01M01MPEAM1HHHH@Z) [D:\Projects\opencv\build-vc141-x64\modules\core\opencv_core.vcxproj]
  hal_internal.obj : error LNK2019: unresolved external symbol cblas_cgemm referenced in function "int __cdecl lapack_gemm_c<float>(float const *,unsigned __int64,float const *,unsigned __int64,float,float const *,unsigned __int64,float
,float *,unsigned __int64,int,int,int,int)" (??$lapack_gemm_c@M@@YAHPEBM_K01M01MPEAM1HHHH@Z) [D:\Projects\opencv\build-vc141-x64\modules\core\opencv_core.vcxproj]
[...]

PR #9836, opened 2017-10-11
Standalone IPP with IW fix;

lupustr3:
### This pullrequest changes

Fixes OpenCV IW configuration with standalone IPP package.


alalek:
Works well :+1:

fmigneault:
@alalek @lupustr3 
It looks like this PR will break further more IPP IW 2018 configuration for Windows platform (see #9622). 
Is it possible to plan ahead for v2018 that does use `iw/iw_version.h` instead of `iw/iw_ll.h` ?

alalek:
@fmigneault This fix workarounds [problem](https://github.com/opencv/opencv/issues/9622#issuecomment-335672978) with `Cannot open include file: 'iw/iw_ll.h': No such file or directory`.
Currently OpenCV rely on low-level functions from IPP IW which is not a part of public API, so "standalone" IPP IW can't be used directly for now.

PR #9839, opened 2017-10-11
Python wrappers for Vec and Matx

terfendail:
### This pullrequest changes
Added python wrappers for Vec and Matx

### DIFF
--- a/modules/python/src2/cv2.cpp
+++ b/modules/python/src2/cv2.cpp
@@ -394,6 +394,24 @@ bool pyopencv_to(PyObject* o, Mat& m, const char* name)
     return pyopencv_to(o, m, ArgInfo(name, 0));
 }
 
+template<typename _Tp, int m, int n>
+bool pyopencv_to(PyObject* o, Matx<_Tp, m, n>& mx, const ArgInfo info)
+{
+    Mat tmp;
+    if (!pyopencv_to(o, tmp, info)) {
+        return false;
+    }
+
+    tmp.copyTo(mx);
+    return true;
+}
+
+template<typename _Tp, int m, int n>
+bool pyopencv_to(PyObject* o, Matx<_Tp, m, n>& mx, const char* name)
+{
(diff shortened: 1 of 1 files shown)


modules/python/src2/cv2.cpp
```diff
@@ -681,26 +699,25 @@ PyObject* pyopencv_from(const UMat& m) {
...
         return true;
     if (PySequence_Check(o)) {
-        PyObject *fi = PySequence_Fast(o, name);
+        PyObject *fi = PySequence_Fast(o, (const char *)info);
         if (fi == NULL)
             return false;
         if (4 < PySequence_Fast_GET_SIZE(fi))
         {
-            failmsg("Scalar value for argument '%s' is longer than 4", name);
+            failmsg("Scalar value for argument '%s' is longer than 4", (const char *)info);
```
alalek: `info.name` is more clear and safe.

"Hack" operator should not be used at all (it is here for some build problems only) - it candidate for removal.

modules/python/src2/cv2.cpp
```diff
@@ -1022,14 +1045,140 @@ PyObject* pyopencv_from(const Point3f& p)
...
+template<>
+bool pyopencv_to(PyObject* obj, Vec2i& v, const char* name)
+{
+    return pyopencv_to(obj, v, ArgInfo(name, 0));
+}
+
+template<>
+PyObject* pyopencv_from(const Vec4d& v)
+{
+    return Py_BuildValue("(dddd)", v[0], v[1], v[2], v[3]);
```
alalek: We should start from use-cases including samples in Python language.
As @vpisarev said, Python tuples are very limited for Python processing:
- how to calculate "vector + vector"?
- how to apply "matrix * vector" operations?
- does `numpy.asarray` works fine?
- how to convert `numpy.array` to Vec4d-compatible form?

vpisarev:
Let's merge it in. Some modules use this feature, so they will be the implicit test :+1:

PR #9842, opened 2017-10-12
dnn: don't use "experimental_dnn_v1" namespace directly

alalek:
This is a technical namespace to guard ABI compatibility.


alalek:
:+1:

PR #9845, opened 2017-10-13
Layers for fast-neural-style models

dkurt:
### This pullrequest changes
Layers and changes required to use style transfer models from https://github.com/jcjohnson/fast-neural-style.

* Reflection padding
* Slice layer with begin-end ranges
* `nn.TotalVariation` loss layer mapping to an Identity layer.
* `nn.ShaveImage` using Slice layer
* Instance normalization as a sequence of MVN + Batch normalization layers

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/393

Images received using script from PR:


vpisarev:
cool! :+1:

PR #9847, opened 2017-10-13
dnn(ocl4dnn): add fusion support

wzw-intel:
ocl4dnn supports following fusion styles for now:
- Conv + [BN] + [Scale] + [ReLU/PReLU]


alalek:
Thank you! Works well! :+1:

PR #9848, opened 2017-10-13
Made flann dependency for features2d optional

jet47:
It will allow to build features2d even if flann module is not available.

### DIFF
--- a/modules/features2d/include/opencv2/features2d.hpp
+++ b/modules/features2d/include/opencv2/features2d.hpp
@@ -43,8 +43,12 @@
 #ifndef OPENCV_FEATURES_2D_HPP
 #define OPENCV_FEATURES_2D_HPP
 
+#include "opencv2/opencv_modules.hpp"
 #include "opencv2/core.hpp"
+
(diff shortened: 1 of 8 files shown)


modules/features2d/include/opencv2/features2d.hpp
```diff
@@ -1099,6 +1103,7 @@ class CV_EXPORTS_W BFMatcher : public DescriptorMatcher
     bool crossCheck;
 };
 
+#ifdef HAVE_OPENCV_FLANN
```
alalek: `#if defined(HAVE_OPENCV_FLANN) || defined(CV_DOXYGEN)`

jet47: Fixed.

alalek:
```.diff
-#ifdef
+#if
```

jet47: Sorry, my mistake.

mshabunin:
:+1:

PR #9850, opened 2017-10-13
fix the possible bug for key mapping in Linux GTK library

### LINKED ISSUE #9641: A possible bug in modules/highgui/src/roiSelector.cpp

- OpenCV => 3.3.0
- Operating System / Platform => Fedora 21 x64
- Compiler => g++

##### Detailed description

When running the example "opencv_contrib/modules/tracking/samples/tutorial_introduction_to_tracker.cpp" (opencv_contrib 3.3.0) , the prompt output gives me a hint - "Select a ROI and then press SPACE or ENTER button!
Cancel the selection process by pressing c button!
", but pressing ENTER button gives no response for me, while pressing SPACE is OK. So I write a python script to print the value that the function waitKey() returns when I press ENTER button. The result is 10. Then I traversed the related source codes, and found it in "modules/highgui/src/roiSelector.cpp: line 38"
` while (!(key == 32 || key == 27 || key == 13))`

I think this may be related to keyboard key mapping in Fedora 21. To make this code more robust, my proposed workaround is to change the code above to 
` while (!(key == 32 || key == 27 || key == 13 || key == 10))`

After changing the code locally, I recompile opencv 3.3.0 and opencv_contrib 3.3.0, and test the tutorial code 'tutorial_introduction_to_tracker.cpp' again. It is OK now.

klchang:
resolves #9641

### DIFF
--- a/modules/highgui/src/window_gtk.cpp
+++ b/modules/highgui/src/window_gtk.cpp
@@ -1632,7 +1632,7 @@ static gboolean icvOnKeyPress(GtkWidget* widget, GdkEventKey* event, gpointer us
         break;
     case GDK_Return:
     case GDK_Linefeed:
-        code = '\n';
+        code = 13;
         break;
     case GDK_Tab:
         code = '\t';


alalek:
:+1:

PR #9853, opened 2017-10-13
Fix some DNN samples for compatibility with Python 3

catree:
### This pullrequest introduces
small changes to be able to run some DNN samples with Python 3:

- integer division returns float in Python 3, [see](https://stackoverflow.com/questions/1282945/python-integer-division-yields-float)
- long renamed to int in Python 3

I hope these changes do not break the compatibility with Python 2.

### DIFF
--- a/samples/dnn/colorization.py
+++ b/samples/dnn/colorization.py
@@ -1,4 +1,6 @@
-# Script is based on https://github.com/richzhang/colorization/colorize.py
+# Script is based on https://github.com/richzhang/colorization/blob/master/colorize.py
+# To download the caffemodel and the prototxt, see: https://github.com/richzhang/colorization/tree/master/models
+# To download pts_in_hull.npy, see: https://github.com/richzhang/colorization/blob/master/resources/pts_in_hull.npy
 import numpy as np
 import argparse
 import cv2 as cv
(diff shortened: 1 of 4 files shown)


samples/dnn/colorization.py
```diff
@@ -27,8 +29,8 @@ def parse_args():
 
     # populate cluster centers as 1x1 convolution kernel
     pts_in_hull = pts_in_hull.transpose().reshape(2, 313, 1, 1)
-    net.getLayer(long(net.getLayerId('class8_ab'))).blobs = [pts_in_hull.astype(np.float32)]
```
dkurt: @catree, It'll break the script at https://github.com/opencv/opencv/blob/master/modules/dnn/misc/python/pyopencv_dnn.hpp#L9. Could you check it and  modify the function to be more flexible?

samples/dnn/mobilenet_ssd_python.py
```diff
@@ -99,7 +99,7 @@
         y2 = y1 + cropSize[1]
         x1 = (cols - cropSize[0]) / 2
         x2 = x1 + cropSize[0]
-        frame = frame[y1:y2, x1:x2]
+        frame = frame[int(y1):int(y2), int(x1):int(x2)] #for Python 3 compatibility (2/2=1.0)
```
dkurt: I think `y1 = int((rows - cropSize[1]) / 2)`, `x1 = int((cols - cropSize[0]) / 2)` is even better.

samples/dnn/colorization.py
catree: It works now for both Python version.

dkurt: @catree, So, we can emit the casting here?

dkurt:
:+1: It works for me! Thank you!

PR #9856, opened 2017-10-14
Update OpenCVCompilerOptimizations.cmake

GregoryMorse:
Neon (GNU C style command line) not supported on MSVC ARM breaking build fix


GregoryMorse:
ARM builds are not working with MSVC 2017 (version 15):

CMake Error at cmake/OpenCVCompilerOptimizations.cmake:448 (message):
  Compiler doesn't support baseline optimization flags: -mfpu=neon
Call Stack (most recent call first):
  cmake/OpenCVCompilerOptions.cmake:263 (ocv_compiler_optimization_options)
  CMakeLists.txt:529 (include)


-- Configuring incomplete, errors occurred!

GregoryMorse:
For documentation purposes for those who might try to build: To successfully build ARM MSVC on VS2017 with latest 3.3 also need obsolete patch (it was fixed in a very recently accepted PR to change to universal intrinsics):
merge pull request, plus sources\modules\imgproc\src\smooth.cpp change line: C:\Users\Gregory\Documents\My Libraries\opencv330\opencv\sources\modules\imgproc\src\smooth.cpp(3961): error C2664: '__n128 __neon_QdQnQm(unsigned int,__n128,__n128)': cannot convert argument 3 from 'volatile const float32x4_t' to '__n128'
REM C:\Users\Gregory\Documents\My Libraries\opencv330\opencv\sources\modules\imgproc\src\smooth.cpp(3961): note: Cannot copy construct union '__n128' due to ambiguous copy constructors or no available copy constructor
REM add: *(float32x4_t*)& thereby: float32x4_t _alpha = vsubq_f32(_val, *(float32x4_t*)&_val0);

GregoryMorse:
ARM MSVC builds on MSVC2017 are successful and this PR can be merged.

cmake/OpenCVCompilerOptimizations.cmake
```diff
@@ -257,7 +257,9 @@ elseif(ARM OR AARCH64)
   if(NOT AARCH64)
     ocv_update(CPU_KNOWN_OPTIMIZATIONS "VFPV3;NEON;FP16")
     ocv_update(CPU_VFPV3_FLAGS_ON "-mfpu=vfpv3")
-    ocv_update(CPU_NEON_FLAGS_ON "-mfpu=neon")
+    if(NOT MSVC)
+      ocv_update(CPU_NEON_FLAGS_ON "-mfpu=neon")
+    endif()
```
alalek: > Neon not supported on MSVC ARM

This is [not correct](https://msdn.microsoft.com/en-us/library/hh875058.aspx#Anchor_0).
MSVC doesn't support GNU-like compiler options. So other "-mfpu" flags should be fixed too to keep code consistent.

Do you have CMake toolchain file for MSVC ARM build?

modules/core/src/hal_replacement.hpp
```diff
@@ -717,8 +717,12 @@ inline int hal_ni_gemm64fc(const double* src1, size_t src1_step, const double* s
 
 #if defined __GNUC__
 #  pragma GCC diagnostic pop
+#define EXPAND
 #elif defined _MSC_VER
 #  pragma warning( pop )
+#define EXPAND( x ) x
+#else
+#define EXPAND
```
alalek: Macros with generic names should be avoided, especially in public headers (to prevent possible conflicts). For example, use `CV_` prefix.

Consider to reuse [this macro](https://github.com/opencv/opencv/blob/3.3.0/modules/core/include/opencv2/core/cvdef.h#L73).

cmake/OpenCVCompilerOptimizations.cmake
GregoryMorse: I am using the WinRT ARM toolchain file as I do not know of a non-WinRT Windows ARM toolchain file though it would likely be simply modifying the flag in the toolchain to support the full Win API.

You are right.  In fact I got it compiling with all the neon codes too as can be noted from my comment below where a fix to already replaced by universal framework code had to be fixed specifically for neon.  Only this command line was the problem.  I will attempt to generalize it for this file.

modules/core/src/hal_replacement.hpp
GregoryMorse: Yes its absolutely the correct design to reuse this one.

alalek:

PR #9860, opened 2017-10-16
Several uninitialized class fields

mshabunin:
Addresses two issues detected by static analysis.


modules/dnn/src/darknet/darknet_importer.cpp
```diff
@@ -64,7 +64,7 @@ class DarknetImporter : public Importer
 
 public:
 
-    DarknetImporter() {}
+    DarknetImporter() { net.channels = 0; net.width = 0; net.height = 0; }
```
alalek: It is better to create initialization ctor for `darknet::NetParameter` (darknet_io.hpp), than initializing its members here.

mshabunin: Thank you, I've updated the commit.

alalek:
:+1:

PR #9862, opened 2017-10-16
dnn: make NMS function public

sovrasov:
Merge with https://github.com/opencv/opencv_extra/pull/396


modules/dnn/include/opencv2/dnn/nms.inl.hpp
```diff
@@ -0,0 +1,114 @@
...
+struct NMSOverlap
+{
+    float operator() (const BoxType& a, const BoxType& b);
+};
+
+template <>
+inline float NMSOverlap<Rect>::operator() (const Rect& a, const Rect& b)
+{
+    float rectIntersectionArea = (float)(a & b).area();
+    return rectIntersectionArea / (a.area() + b.area() - rectIntersectionArea);
```
alalek: `1.0 - jaccardDistance(a, b)`

`jaccardDistance` is [here](https://github.com/opencv/opencv/blob/3.3.0/modules/core/include/opencv2/core/types.hpp#L1881-L1900)

modules/dnn/include/opencv2/dnn/nms.inl.hpp
```diff
@@ -0,0 +1,114 @@
...
+    {
+        if (scores[i] > threshold)
+        {
+            score_index_vec.push_back(std::make_pair(scores[i], i));
+        }
+    }
+
+    // Sort the score pair according to the scores in descending order
+    std::stable_sort(score_index_vec.begin(), score_index_vec.end(),
+                     SortScorePairDescend<int>);
```
alalek: stable_sort: "The order of equal elements is guaranteed to be preserved". Do we really need this?

Update: looks like it is just to be consistent with original implementation.

modules/dnn/include/opencv2/dnn/nms.inl.hpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Do non maximum suppression given bboxes and scores.
+// Inspired by Piotr Dollar's NMS implementation in EdgeBox.
+// https://goo.gl/jV3JYS
+//    bboxes: a set of bounding boxes.
+//    scores: a set of corresponding confidences.
+//    score_threshold: a threshold used to filter detection results.
+//    nms_threshold: a threshold used in non maximum suppression.
+//    top_k: if not -1, keep at most top_k picked indices.
```
alalek: -1 is quite strange value for threshold. `0` - is really still valid value?
Suggested:
```
> 0 - apply limit
else - use all values
```

modules/dnn/include/opencv2/dnn/nms.inl.hpp
```diff
@@ -0,0 +1,114 @@
...
+        const int idx = score_index_vec.front().second;
+        bool keep = true;
+        for (int k = 0; k < (int)indices.size() && keep; ++k) {
+            const int kept_idx = indices[k];
+            float overlap = computeOverlap(bboxes[idx], bboxes[kept_idx]);
+            keep = overlap <= adaptive_threshold;
+        }
+        if (keep)
+            indices.push_back(idx);
+        score_index_vec.erase(score_index_vec.begin());
```
alalek: Erase operation can be eliminated:
```
for (size_t i = 0; i < score_index_vec.size(); ++i)
{
    const int idx = score_index_vec[i].second;
    ...
}
```

dkurt:
TensorFlow may help with test samples: https://www.tensorflow.org/versions/master/api_docs/python/tf/image/non_max_suppression .

alalek:
It would be nice to have accuracy tests too.

alalek:
Looks good to me :+1:

modules/dnn/include/opencv2/dnn/nms.inl.hpp
```diff
@@ -0,0 +1,100 @@
+// This file is part of OpenCV project.
```
alalek: Please move this file into "src" directory (to keep exported interfaces clean).

PR #9864, opened 2017-10-16
Use CMake's ${CMAKE_INSTALL_LIBDIR} variable to decide where to place the built binaries

mapreri:
The typical situation where the usual, plain 'lib' is not wanted is in a
multiarch system, that allows installation of the same library built for
multiple architectures.
Use the standard ${CMAKE_INSTALL_LIBDIR} variable (coming from the
GNUInstallDirs module) to figure where to put the built libraries.

Most probably other variables from GNUInstallDirs should be used, but that's something for the future, when the need arises.


alalek:
Looks good to me. Patch will be merged after the nearest OpenCV release (there is some changes introduced which we should properly test).

BTW, You could override this value via CMake command-line parameters: `cmake -DOPENCV_LIB_INSTALL_PATH=...`

mapreri:
> BTW, You could override this value via CMake command-line parameters:

Sure, but less options and more ~~magic~~ automation is better :)

PR #9866, opened 2017-10-17
doc: Make MATHJAX_RELPATH configurable through cmake

mapreri:
This comes from Debian, where there is a patch changing MATHJAX_RELPATH to a local installation of MathJax.

I'd prefer to use a build-time option -DOCV_MATHJAX_RELPATH (or whatever, I picked that name, but it's really arbitrary), over a local patch.

Thank you for the consideration!


doc/CMakeLists.txt
```diff
@@ -37,6 +37,8 @@ if(BUILD_DOCS AND DOXYGEN_FOUND)
   unset(CMAKE_DOXYGEN_TUTORIAL_CONTRIB_ROOT)
   unset(CMAKE_DOXYGEN_TUTORIAL_JS_ROOT)
 
+  set(OCV_MATHJAX_RELPATH "http://cdn.mathjax.org/mathjax/latest" CACHE STRING "URI to a MathJax installation")
```
sturkmen72: "URI to a MathJax installation"
URI -> URL ?

alalek:
Thank you for contribution!

Consider to use these names:
`DOXYGEN_MATHJAX_RELPATH`
or
`OPENCV_MATHJAX_RELPATH` (to be consistent with other OpenCV's specific variables)

Please take a look on this issue about mathjax URL: #8630
Also we would like to make URL protocol-independent (avoid mix of http/https), so please use this:
```
"//cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0"
```

mapreri:
> Also we would like to make URL protocol-independent (avoid mix of http/https), so please use this:
> `"//cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0"`

Are you sure Doxygen supports this?  I've never saw it myself, but then again, it's not like I can claim to know Doxygen.

doc/CMakeLists.txt
mapreri: Because it can also point to a local path (i.e., in Debian we make it point to /usr/share/javascript/mathjax/), therefore 'URL' would be too limited.
What do you think?

alalek:
This doesn't work well (http/https - works, but `file://` - doesn't), so please force `https://` (it works in all cases):
"https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0"

mapreri:
Done, I picked OPENCV_MATHJAX_RELPATH as a name, and switched to the cloudlfare CDN as it was before (sorry about this, I copy-pasted the old URL by accident…).

alalek:

PR #9877, opened 2017-10-17
Fix build for non-linux architectures but still glibc-based

mapreri:
Exampls of these are gnu/kfreebsd and gnu/hurd, both available as
unofficial Debian ports.

They don't define `__linux__` (as they are non-linux…) but still define
`__GLIBC__`, so check on that.

### DIFF
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -52,7 +52,7 @@
     #undef abs
 #endif
 
-#if defined __linux__ || defined __APPLE__
+#if defined __linux__ || defined __APPLE__ || defined __GLIBC__
     #include <unistd.h>
     #include <stdio.h>
     #include <sys/types.h>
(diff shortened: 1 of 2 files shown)


alalek:
Does this really robust solution? For example, MinGW uses GLIBC/GLIBCXX internally, but we don't want to call "sysconf" in this case.

mapreri:
uhh, no idea really.  I'm not a windows dev at all, and know next to nothing about it.
But I see a windows build on your CI, doesn't that cover what you just said?

alalek:
We doesn't have MinGW builds actually (we test more native way: MSVS)

vpisarev:
@mapreri, thank you for the patch! FreeBSD flavors seem to be handled already. For HURD you can use more specific __GNU__ macro, which will be consistent with other OS'es, such as Linux (we use __linux__), Mac (we use __APPLE__) etc.

mapreri:
@vpisarev 
> FreeBSD flavors seem to be handled already

Note that here I'm talking about kFreeBSD, which does not define `__FreeBSD__` (besides, FreeBSD doesn't use GLIBC).  In kFreeBSD `__FreeBSD_kernel__` is defined though.

> For HURD you can use more specific `__GNU__` macro

Yes, `__GNU__` would work great for Hurd, but wouldn't cover our (admittedly weird) kFreeBSD case.

vpisarev:
@mapreri, ok, I see. @mshabunin checked this with MinGW and seems to work well. Let's put it in :+1:

mapreri:
Great, thank you!

If it turns out this breaks some other arch/system combination, feel free to ping me to see about solving it differently :)

PR #9886, opened 2017-10-19
Removed cmake submodules and contrib_world references

mshabunin:
**Merge with**: https://github.com/opencv/opencv_contrib/pull/1418

This feature has not been used for a long time


vpisarev:
@alalek, can you please review it? From my side I support any simplification of our CMake scripts, including this patch

alalek:
:+1:

PR #9889, opened 2017-10-19
Add const to getViewerPose()

### LINKED ISSUE #9393: Make Viz3d::getViewerPose const
- OpenCV => 3.3
- Compiler => g++ 5.4.0

##### Detailed description

`Viz3d::getViewerPose()` is not declared as const, although it returns a new `Affine3d` instance which does not change the `Viz3d` object: https://github.com/opencv/opencv/blob/a5b5684670ebc8d5dffa6499949ab1a573c04b8b/modules/viz/src/vizimpl.cpp#L512
https://github.com/opencv/opencv/blob/a5b5684670ebc8d5dffa6499949ab1a573c04b8b/modules/viz/src/vizimpl.cpp#L524

This leads to problems, e.g., it is not possible to ask a `const Viz3d` object for its viewer pose.

##### Steps to reproduce

Suggestion: Change the signature of [Viz3d::getViewerPose](http://docs.opencv.org/3.3.0/d6/d32/classcv_1_1viz_1_1Viz3d.html#a6eb3b162253dfefbdf8c965d74c796ef) from `cv::Affine3d cv::viz::Viz3d::VizImpl::getViewerPose()` to `cv::Affine3d cv::viz::Viz3d::VizImpl::getViewerPose() const`.

IgWod:
### This pullrequest changes
As described in the issue: #9393 method is not defined as `const cv::Affine3d cv::viz::Viz3d::VizImpl::getViewerPose()` however it doesn't modify the instance of the class. This PR adds the const qualifier to the method. Beside this function not being const may cause some problems it's also not consistent with other getters as they are defined as const. Hence making this method const is a good idea.

resolves #9393

### DIFF
--- a/modules/viz/src/vizimpl.cpp
+++ b/modules/viz/src/vizimpl.cpp
@@ -509,7 +509,7 @@ void cv::viz::Viz3d::VizImpl::setViewerPose(const Affine3d &pose)
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////////
-cv::Affine3d cv::viz::Viz3d::VizImpl::getViewerPose()
+cv::Affine3d cv::viz::Viz3d::VizImpl::getViewerPose() const
 {
     vtkCamera& camera = *renderer_->GetActiveCamera();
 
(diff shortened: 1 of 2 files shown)


savuor:
:+1:

PR #9890, opened 2017-10-20
Binary data for batch normalization test from Torch

dkurt:
### This pullrequest changes

* Binary data for batch norm test.
* Resize face detector input without keeping an aspect ratio.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/398


alalek:
Thank you! :+1:

PR #9895, opened 2017-10-20
Adding threading in calibrate.py

art049:
### This pullrequest changes

This pull request add threading in the calibration.py sample script. The old script can take a lot of time to be executed on large images.

### DIFF
--- a/samples/python/calibrate.py
+++ b/samples/python/calibrate.py
@@ -18,7 +18,6 @@ from __future__ import print_function
 
 import numpy as np
 import cv2
-import threading
 
 # local modules
 from common import splitfn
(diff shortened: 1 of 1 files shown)


alalek:
Thank you for your effort!
But lets keep samples as simple as possible.

art049:
Do you think adding this multi-threading could be useful in interactive calibration ?
I've been using it and it's quite annoying to wait a long time when images are large

alalek:
I believe, effective interactive calibration application should cache results of chess-board detection for previous frames. In this case parallel processing of images is not necessary (can't be applied, because we have one new image only).

BTW, Did you have a chance to take a look on this application (C++): https://github.com/opencv/opencv/tree/master/apps/interactive-calibration ?

art049:
Yes, I had a look. I understand your point when processing a video. But
when the input is a sequence of images It could be a lot faster just with
threading. I'm just asking because I did the modifications for myself and I
was interested in sharing these.

alalek:
I have pushed update with changes:
- replace magic THREADS number with command-line argument
- drop manual locks
- use Python thread [Pool](https://docs.python.org/2/library/multiprocessing.html#module-multiprocessing.dummy)
- other fixes (allow to disable debug mode, incorrect image pattern, etc)

PR #7064, opened 2016-08-08
Implementation of PAM encoder/decoder

DmtKats:
**Merge with**: https://github.com/opencv/opencv_extra/pull/306
**Original PR**: https://github.com/opencv/opencv/pull/6912

This pull request implements the PAM image format as defined at:
http://netpbm.sourceforge.net/doc/pam.html

The PAM format provides a generic means for storing 2 dimensional information.
This is useful for opencv since there are cases where data gets translated into
non standardized formats, which makes it difficult to store and load this information.

I have reopened the PR for opencv and opencv_extra as per the suggestion of alalek. This was done since the testing system ignores branches with names like "master" and "2.4"

Issue #6567, opened 2016-05-21
GdalDecoder::checkSignature has type-o


shawks03:
Discovered this bug when I built opencv without jpg support but with gdal support. Then trying to open a jpg the gdal decoder ran, and I got an exception from substr. Changing 140 to 0 made the exception go away.

Here's the code within modules/imgcodecs/src/grfmt_gdal.cpp, function GdalDecoder::checkSignature:

``` cpp
bool GdalDecoder::checkSignature( const String& signature )const{
    std::string str = signature.c_str();
    if( str.substr(0,4).find("NITF") != std::string::npos ){
        return true;
    }
    if( str.substr(140,4) == "DTED" ){
        return true;
    }
    return false;
}
```

kevin-coder:
@shawks03 , could you please provide code snip in order to re-produce this issue?

shawks03:
I don't have the original code, but if I remember correctly, I was calling cv::imread on a jpg. And I built my opencv without jpg support, but with gdal support. Therefore, GdalDecoder::checkSignature was called on the jpg, which caused the second substr to throw because the length of signature was only 4. From looking at the code, it's pretty obvious to me that 140 is a mistake.

kevin-coder:
Got you, and seems that it still cannot invoking checkSignature in GdalDecoder, could you please double confirm my re-produce environment as well as steps? Thanks!
**Environment:**
Windows 8.1, CMake 3.5.2 and VS2012
**Reproduce steps:**
1. Clone the latest master branch code and configure in CMake. The key steps in CMake configuration is:
1). Disable the MACRO "WITH_JPEG"
2). Enable the MACRO "WITH_GDAL"
2. After Configure and Generate successfully, build the OpenCV source code.
3. Write sample code which only include calling of function imread, passing the classic "lena.jpg" as its parameter.
4. Debug this sample code and find it cannot calling checkSignature in GdalDecoder.

In this procedure, it called checkSignature functions in other modules exclude GdalDecoder. I think my result was different from yours probably caused by:
1). My reproduce environment is different from yours.
2). My CMake configuration settings might be different from yours.
So, could you please double confirm above 2 steps? Thanks!

shawks03:
Can you confirm that gdal is installed and found by the cmake step for
opencv?

I'm using Ubuntu, but that shouldn't matter.

mshabunin:
I've reproduced the issue. Problem is in this line:
`std::string str = signature.c_str();`
[...]

PR #9906, opened 2017-10-22
Updates grfmt_gdal.cpp

sturkmen72:
#6567
additional fix to #7064


alalek:
@sturkmen72 Thank you!
Could you share some small image for this (for regression test)?

sturkmen72:
i have no valid image but a code snippet like https://github.com/opencv/opencv/blob/master/samples/cpp/train_HOG.cpp#L72-L81
should work without crash with this fix.

sturkmen72:
indeed according the documentation i think `GdalDecoder ` should be active only when `imread( files[i],IMREAD_LOAD_GDAL)`

alalek:
:+1:

PR #9753, opened 2017-09-30
core: remove raw SSE2/NEON implementation from matmul.cpp

tomoaki0705:
**Merge with extra**: opencv/opencv_extra#400

### This pullrequest changes
  * use universal intrinsic instead of raw intrinsic
  * add test ```Core_TransformLargeTest``` for 8u version of transform
    * 8u SIMD version requires at least length of 16 elements  (bytes) but it seems the maximum was 4 bytes
  * may be adding test in too much, waiting for advice

PR #9907, opened 2017-10-23
core: POWER/VSX Improvements

seiko2plus:
POWER/VSX Improvements

Improvements are related to #7207

- VSX compatibility in vsx_utils.hpp among compilers gcc, clang, xlc becomes more mature and organized.
 
- As result of supporting IBM Advance toolchain, now we can build and test Opencv for powerpc in x86_64,
  For more information about IBM Advance toolchain and how to install it checkout this [link](https://www.ibm.com/developerworks/community/wikis/home?lang=en#!/wiki/W51a7ffcf4dfd_4b40_9d82_446ebc23c550/page/IBM%20Advance%20Toolchain%20for%20PowerLinux%20Documentation).
  
  Options that are provided by toolchain
    
  * AT_PATH : Advance Toolchain directory, i.e. -DAT_PATH=/opt/at11.0

  * AT_RPATH : Add new directories to runtime search path, i.e. -DAT_RPATH=/path/to/opencv/lib,/path/to/opencl/runtime.
    Note:
    > By default, the Advance Toolchain will prepend its own lib directory to the RPATH/RUNPATH. 
      However, in case the user changes this behavior, he or she must ensure the Advance Toolchain 
      lib directory will always appear before a directory with another libc, i.e. /usr/lib, /usr/lib64, /lib or /lib64. 
      This will guarantee you won't mix parts of different glibc versions.

  * AT_HOST_LINK: instead of testing opencv on the target machine, 
    you can turn this option ON to link against host advance toolchain runtime 
    so you can perform tests directly in the host using QEMU user emulation.
    i.e. -DAT_HOST_LINK=ON , qemu-ppc64le ./bin/opencv_test_core

  * AT_NO_AUTOVEC: For some kind of reasons QEMU instruction emulation 
    fails to load unaligend addresses correctly especially with floating-point if vectorization optimization is enabled.
    Vectorization in GCC is enabled by the flag -ftree-vectorize and by default at -O3,
    so you should enable this option if you going to build opencv in release mode.
    Note: Auto vectorization optimization works without any issues in PowerKVM/QEMU. 

Current issues:
  - xlc fails to compile googletest
  - clang 5 aborted during parse googletest
  - There are several issues on big-endian mode and
    I'm willing to fix them later but I still have no intention to support vsx on it.

```
force_builders=Custom
```

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin_vsx.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_vsx.hpp
@@ -521,24 +521,25 @@ OPENCV_HAL_IMPL_VSX_BIN_FUNC(v_add_wrap, vec_add)
 OPENCV_HAL_IMPL_VSX_BIN_FUNC(v_sub_wrap, vec_sub)
 
 /** Bitwise shifts **/
-#define OPENCV_HAL_IMPL_VSX_SHIFT_OP(_Tpuvec, splfunc)   \
-inline _Tpuvec operator << (const _Tpuvec& a, int imm)   \
-{ return _Tpuvec(vec_sl(a.val, splfunc(imm))); }         \
-inline _Tpuvec operator >> (const _Tpuvec& a, int imm)   \
-{ return _Tpuvec(vec_sr(a.val, splfunc(imm))); }         \
-template<int imm> inline _Tpuvec v_shl(const _Tpuvec& a) \
-{ return _Tpuvec(vec_sl(a.val, splfunc(imm))); }         \
-template<int imm> inline _Tpuvec v_shr(const _Tpuvec& a) \
-{ return _Tpuvec(vec_sr(a.val, splfunc(imm))); }
+#define OPENCV_HAL_IMPL_VSX_SHIFT_OP(_Tpvec, shr, splfunc)   \
+inline _Tpvec operator << (const _Tpvec& a, int imm)         \
+{ return _Tpvec(vec_sl(a.val, splfunc(imm))); }              \
+inline _Tpvec operator >> (const _Tpvec& a, int imm)         \
+{ return _Tpvec(shr(a.val, splfunc(imm))); }                 \
+template<int imm> inline _Tpvec v_shl(const _Tpvec& a)       \
+{ return _Tpvec(vec_sl(a.val, splfunc(imm))); }              \
+template<int imm> inline _Tpvec v_shr(const _Tpvec& a)       \
+{ return _Tpvec(shr(a.val, splfunc(imm))); }
 
-OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_uint8x16, vec_uchar16_sp)
-OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_int8x16, vec_uchar16_sp)
-OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_uint16x8, vec_ushort8_sp)
-OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_int16x8, vec_ushort8_sp)
-OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_uint32x4, vec_uint4_sp)
-OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_int32x4, vec_uint4_sp)
-OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_uint64x2, vec_udword2_sp)
-OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_int64x2, vec_udword2_sp)
+OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_uint8x16, vec_sr, vec_uchar16_sp)
+OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_uint16x8, vec_sr, vec_ushort8_sp)
+OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_uint32x4, vec_sr, vec_uint4_sp)
+OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_uint64x2, vec_sr, vec_udword2_sp)
+// algebraic right shift
+OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_int8x16, vec_sra, vec_uchar16_sp)
+OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_int16x8, vec_sra, vec_ushort8_sp)
+OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_int32x4, vec_sra, vec_uint4_sp)
+OPENCV_HAL_IMPL_VSX_SHIFT_OP(v_int64x2, vec_sra, vec_udword2_sp)
 
 /** Bitwise logic **/
 #define OPENCV_HAL_IMPL_VSX_LOGIC_OP(_Tpvec)    \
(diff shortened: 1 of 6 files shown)


alalek:
Looks good to me :+1:

seiko2plus:
@alalek  I implemented the new universal intrinsics and I think it's ready for merge
but I prefer to wait until @tomoaki0705 finish #9753

tomoaki0705:
Thank you very much @seiko2plus . I really appreciate your kind help offer.
However ,the #9753 didin't end happily, I had to revert the ```dotprod``` version back to pure ```NEON``` version. (the story is written on #9753 )
This ends up in removing "accumulate" version of ```v_dotprod```.
Still, ```v_rotate_left```, ```v_rotate_right``` and ```v_matmuladd``` are there, but I had to remove new version of ```v_dotprod```

looking in your PR, it seems you already implemented this new version of ```v_dotprod```
I really appreciate your work, and I do respect your work, but since this intrinsic doesn't help, I removed it.

I'm sorry but to say this, but could you follow my modification, please ?

Again, nice work and sorry for disturbing because of my mistake.

seiko2plus:
@tomoaki0705 It's alright, I removed the accumulate version of v_dotprod and I kept v_matmuladd, v_rotate_left/right.

seiko2plus:
@alalek I changed my mind, please leave this PR open until I finish the implementation of universal intrinsics (exp, log, sin, cos) for VSX and @tomoaki0705 finishes #9753 in case if he is gonna need any new intrinsics.
thank you.

mshabunin:
I've been able to build OpenCV using toolchain provided by Ubuntu 17.10.

Config:
```.Dockerfile
FROM ubuntu:17.10

RUN apt-get update && apt-get install -y \
    curl git python3 cmake ninja-build \
    gcc-7-powerpc64le-linux-gnu \
    g++-7-powerpc64le-linux-gnu \
    binutils-powerpc64le-linux-gnu \
    libc6-ppc64-powerpc-cross \
    pkg-config-powerpc64le-linux-gnu
```

Toolchain file:
```.cmake
set(CMAKE_SYSTEM_NAME Linux)
set(CMAKE_SYSTEM_PROCESSOR powerpc64le)

set(CMAKE_C_COMPILER /usr/bin/powerpc64le-linux-gnu-gcc-7)
set(CMAKE_CXX_COMPILER /usr/bin/powerpc64le-linux-gnu-g++-7)

set(CMAKE_FIND_ROOT_PATH /usr/powerpc64le-linux-gnu)

SET(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM ONLY)
SET(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
SET(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
```

mshabunin:
I tried to run the _core_ test in same container (need to install `qemu-user` package):
```.sh
qemu-ppc64le -L /usr/powerpc64le-linux-gnu -E OPENCV_TEST_DATA_PATH=/testdata /build/bin/opencv_test_core
```
**Note:** options `-fno-tree-vectorize -fno-tree-slp-vectorize` were needed to fix some tests.

Unfortunately the _hal_intrin.int8x16_ test crashes (added `--gtest_filter=*:-hal_intrin.8x16` to filter it out):
```.sh
$ qemu-ppc64le -L /usr/powerpc64le-linux-gnu -E OPENCV_TEST_DATA_PATH=/testdata /build/bin/opencv_test_core
 --gtest_filter=*intrin*                                                                                                           
CTEST_FULL_OUTPUT                                                                                                                  
OpenCV version: 3.3.1-dev                                                                                      
OpenCV VCS version: unknown                                                                                                        
Build type: debug                                                                                          
Parallel framework: pthreads                                                                                                       
CPU features: VSX                                                                                              
OpenCL is disabled                                                                                                                 
Note: Google Test filter = *intrin*                                                                                       
[==========] Running 11 tests from 1 test case.                                                                                    
[----------] 11 tests from hal_intrin                                                                                              
[ RUN      ] hal_intrin.uint8x16                                                                                                   
[       OK ] hal_intrin.uint8x16 (8 ms)                                                                    
[ RUN      ] hal_intrin.int8x16                                                                                                    
Segmentation fault (core dumped) 
```

alalek:
@mshabunin Add "-cpu POWER9" option to QEMU (it is "Illegal instruction" fault).

seiko2plus:
@mshabunin Did you used this current patch during tests?

if not, please use this current patch for the following reasons

when we use one of these VSX instructions that truncate double-precision to fixed-point word or
round doubleword to single-precision

Qemu instruction emulation return: 
 even-numbered elements of the result vector are zero and the other elements contains conversion result
 ex: vec_ctu((vec_double2){1, 2}) return (vec_uint4) {0, 1, 0, 2}

PowerKVM/QEMU return: even and odd numbered elements contains conversion result
 ex: vec_ctu((vec_double2){1, 2}) return (vec_uint4) {1, 1, 2, 2}

I didn't notice this issue because all my previous tests was only on PowerKVM.
I worked around this issue by use odd-numbered elements instead of even-numbered elements.

alalek:
[This](http://www-01.ibm.com/support/docview.wss?uid=swg27048883&aid=7) says about `vec_ctu` (page 385):

> Note: Elements 1 and 3 of the result vector are undefined when a is a double
vector

mshabunin:
Yes, `-cpu POWER9` helped, now tests are passing. Thank you.

@seiko2plus , I used this patch + #10078 + master

seiko2plus:
@alalek vec_ctu(vec_double2, 0) maps to "xvcvdpuxws" which is affected by endian-mode and xlc implements it without any extra instraction.
this note "Elements 1 and 3 of the result vector are undefined when a is a double vector" must concern big-endian.

In other words, results of VSX conversion instractions are affected by endian mode for instance,

any VSX conversion instraction that converts 4 lanes to 2 lanes on big-endian mode it returns as following
{A, B, C, D} -> {A, C}
and vice versa on little-endian
{A, B, C, D} -> {B, D}

also any VSX conversion instraction that converts 2 lanes to 4 lanes on big-endian mode it returns as following
{A, B} -> {A, undefined, B, undefined}
and vice versa on little-endian
{A, B} -> {undefined, A, undefined, B}

note: 
- during tests on little-endian using powerkvm/power8 vm, I realised different behavior than expected
 {A, B} -> {A, A, B, B}

- QEMU sets undefined results to zero and that makes sense.
 {A, B} -> {0, A, 0, B}

- unfortunately I don't have a direct access to power8 machine to make tests on it. 


On the other hand, xlc(little-endian) somehow tends to add compatibility with big-endian,
for example when we use vec_cvf(vec_float4) which converts single-precision to double-precision, 
it converts the types of elements 0 and 2 in the vector.
i.e. vec_cvf((vec_float4){A, B, C, D}) -> (vec_double2) {A, C}

but that is going to cost us an extra instraction in little-endian and that's why
I ended up with implemented conversions intrinsics of odd-numbered elements which it maps directly to VSX instractions,
and also we still support conversion of even-numbered elements for gcc, clang.
for more explanation please check following intrinsics in vsx_utils.hpp
vec_cvfo, vec_ctdo, vec_ctslo, vec_ctulo.

------------------

@mshabunin old versions of qemu use power7 as default cpu and newest versions use power8 as default cpu.
That gives us two possibilities:

- The default optimization of your gcc cross compiler is targeting power9, you can change default using -mtune flag
  i.e. -mtune=power8

- Qemu version is old, starting from version 1.6 the default cpu is power8, so changing default cpu to power8 should fix the issue
  i.e. -cpu power8

Anyway I prefer to use IBM Advance toolchain which I support in this patch, Did you try using it?

mshabunin:
@seiko2plus , thanks for clarification. I did not try IBM Advance toolchain, the way with standard Ubuntu packages looks easier and works well enough, probably we will use it for testing on the buildbot.

As IBM documentation says:

>In some scenarios you will not get many advantages using the Advance Toolchain, and it will be easier to use the distro's toolchain that usually performs quite well. Such cases include when you are building applications that are not CPU-bound and on I/O constrained environments.

seiko2plus:
@alalek

**Results of VSX conversion instructions are affected by endianness:**

- any VSX conversion instruction that converts 4 to 2 elements on big-endian mode it returns as following
  {A, B, C, D} -> {A, C}
  and vice versa on little-endian
  {A, B, C, D} -> {B, D}

- any VSX conversions instruction that convert 2 to 4 elements on big-endian mode it returns as following
  {A, B} -> {A, undefined, B, undefined}
  and vice versa on little-endian
  {A, B} -> {undefined, A, undefined, B}

On the other hand, xlc(little-endian) prefer the compatibility with big-endian
over the performance, since supports this behavior cost an extra instruction in little-endian.

I ended up with implemented conversions intrinsics of odd-numbered elements which it maps directly to VSX instructions,
and also we still support conversion of even-numbered elements for gcc, clang.
for more explanation please check following intrinsics in vsx_utils.hpp
vec_cvfo, vec_ctdo, vec_ctslo, vec_ctulo.

**note**:
current compatibility in xlc is broken with conversions between 2 to 4 elements,
I was going to fix it in this patch but since it had merge, 
I'm going to release another patch fix this problem by the same way I did with 4 to 2 conversions,
also it will contain a fix for googletest issue in xlc.

mshabunin:
@seiko2plus , I thought we want to support only LE platforms:
```
#if defined(__VSX__) && defined(__PPC64__) && defined(__LITTLE_ENDIAN__)
#  define CV_VSX 1
#endif
```

seiko2plus:
@mshabunin Sorry for not being clear enough, you misunderstood me
I didn't mean that, I was just trying to explain the effects of little/big endian mode on the results of VSX conversion instructions.
However the next patch is going to add further clarification.

PR #9921, opened 2017-10-23
PReLU layer for multidimensional input

dkurt:
### This pullrequest changes

Fix PReLU activation after fully-connected layer: both in-place and standalone cases.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/399


vpisarev:
:+1:

PR #9922, opened 2017-10-23
ts(misc): support tables exporting in markdown format

alalek:
Basic support only (no symbol escapes/sanitizing)

Usage:
```
python table_formatter.py -o markdown test_report.xml
python summary.py -o markdown base.xml patch.xml
```


vpisarev:
:+1:

PR #9931, opened 2017-10-24
Reduce cvtColor memory footprint

mshabunin:
### This pullrequest changes

- several static tables have been moved from .bss section to heap
- fixed error in splineBuild function which caused uninitialized memory usage
- added a mutex guard to the tables initialization function

```
```

### DIFF
--- a/modules/core/include/opencv2/core/private.hpp
+++ b/modules/core/include/opencv2/core/private.hpp
@@ -162,7 +162,7 @@ CV_EXPORTS void scalarToRawData(const cv::Scalar& s, void* buf, int type, int un
 
 //! Allocate all memory buffers which will not be freed, ease filtering memcheck issues
 template <typename T>
-CV_EXPORTS T* allocSingleton(size_t count) { return fastMalloc(sizeof(T) * count); }
+CV_EXPORTS T* allocSingleton(size_t count) { return static_cast<T*>(fastMalloc(sizeof(T) * count)); }
 }
 
 // property implementation macros
--- a/modules/core/include/opencv2/core/private.hpp
+++ b/modules/core/include/opencv2/core/private.hpp
@@ -162,7 +162,7 @@ CV_EXPORTS void scalarToRawData(const cv::Scalar& s, void* buf, int type, int un
 
 //! Allocate all memory buffers which will not be freed, ease filtering memcheck issues
 template <typename T>
-CV_EXPORTS T* allocSingleton(size_t count) { return new T[count]; }
+CV_EXPORTS T* allocSingleton(size_t count) { return fastMalloc(sizeof(T) * count); }
 }
 
 // property implementation macros
--- a/modules/core/include/opencv2/core/private.hpp
+++ b/modules/core/include/opencv2/core/private.hpp
@@ -159,6 +159,10 @@ static inline cv::Size cvGetMatSize( const CvMat* mat )
 namespace cv
 {
 CV_EXPORTS void scalarToRawData(const cv::Scalar& s, void* buf, int type, int unroll_to = 0);
+
+//! Allocate all memory buffers which will not be freed, ease filtering memcheck issues
+template <typename T>
+CV_EXPORTS T* allocSingleton(size_t count) { return new T[count]; }
 }
 
 // property implementation macros
--- a/modules/imgproc/src/color.cpp
+++ b/modules/imgproc/src/color.cpp
@@ -5852,7 +5852,6 @@ enum
     trilinear_shift = 8 - lab_lut_shift + 1,
     TRILINEAR_BASE = (1 << trilinear_shift)
 };
-static int16_t *RGB2LabLUT_s16;
 static int16_t trilinearLUT[TRILINEAR_BASE*TRILINEAR_BASE*TRILINEAR_BASE*8];
 static ushort LabToYF_b[256*2];
 static const int minABvalue = -8145;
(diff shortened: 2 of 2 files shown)


modules/imgproc/src/color.cpp
```diff
@@ -5917,9 +5893,12 @@ static inline softfloat applyInvGamma(softfloat x)
                 pow(xd, softdouble::one()/gammaPower)*(softdouble::one()+gammaXshift) - gammaXshift);
 }
 
+static Mutex tablesInitMtx;
+
 static void initLabTabs()
 {
     static bool initialized = false;
+    AutoLock tablesInitLock(tablesInitMtx);
     if(!initialized)
```
alalek: Please apply this pattern:
https://en.wikipedia.org/wiki/Double-checked_locking

modules/imgproc/src/color.cpp
```diff
@@ -5917,9 +5893,12 @@ static inline softfloat applyInvGamma(softfloat x)
                 pow(xd, softdouble::one()/gammaPower)*(softdouble::one()+gammaXshift) - gammaXshift);
 }
 
+static Mutex tablesInitMtx;
```
alalek: It is not safe to use global static variable with constructor in general. Refer to 'static initialization order fiasco' problem. The most dangerous issues are:
- use before object constructor call
- use after object destroying

modules/imgproc/src/color.cpp
```diff
@@ -5845,10 +5820,11 @@ static const softdouble D65[] = {softdouble::fromRaw(0x3fee6a22b3892ee8),
                                  softdouble::fromRaw(0x3ff16b8950763a19)};
 
 enum { LAB_CBRT_TAB_SIZE = 1024, GAMMA_TAB_SIZE = 1024 };
-static float LabCbrtTab[LAB_CBRT_TAB_SIZE*4];
+static float *LabCbrtTab;
```
alalek: 1. `const`
2. `= NULL`
```
static const float* LabCbrtTab = NULL;
```
3. alignment requirements?
4. we need to make valgrind happy (suppress these hits). To make it easier, probably we should create new function for memory allocation for such "singleton" objects/buffers.
```
float* LabCbrtTab_ = cv::allocSingleton<float>(LAB_CBRT_TAB_SIZE * 4);
splineBuild(f, LAB_CBRT_TAB_SIZE, LabCbrtTab_); // without static_cast<float*>
LabCbrtTab = LabCbrtTab_;
```

vpisarev:
@mshabunin, could you add explicit "=0" or "=NULL", as @alalek suggests, in order to make the code cleaner.

modules/imgproc/src/color.cpp
```diff
@@ -173,7 +148,7 @@ static void splineBuild(const softfloat* f, int n, float* tab)
     int i;
     tab[0] = tab[1] = 0.0f;
 
-    for(i = 1; i < n-1; i++)
+    for(i = 1; i <= n-1; i++)
```
alalek: "<= n" ? (this doesn't look similar to line 150)

mshabunin: This is a fix of an uninitialized memory access due to an error in the _splineBuild_ function. It was not visible when initialization was static, but it is detected by valgrind when initialization is dynamic.

modules/core/include/opencv2/core/private.hpp
```diff
@@ -159,6 +159,10 @@ static inline cv::Size cvGetMatSize( const CvMat* mat )
 namespace cv
 {
 CV_EXPORTS void scalarToRawData(const cv::Scalar& s, void* buf, int type, int unroll_to = 0);
+
+//! Allocate all memory buffers which will not be freed, ease filtering memcheck issues
+template <typename T>
+CV_EXPORTS T* allocSingleton(size_t count) { return static_cast<T*>(fastMalloc(sizeof(T) * count)); }
```
alalek: It would be nice to add entry into vallgrind.supp file too.

mshabunin:
Binary statistics (without IPP): `bloaty -n 7 -d symbols lib/libopencv_imgproc.so`
Before:
```
     VM SIZE                                FILE SIZE
 --------------                          --------------
  40.5%  3.50Mi [Other]                   2.69Mi  66.6%
  19.0%  1.65Mi _ZN2cvL14RGB2LabLUT_s16E       0   0.0%
  19.0%  1.65Mi _ZN2cvL14RGB2LuvLUT_s16E       0   0.0%
   9.9%   877Ki [Unmapped]                1.35Mi  33.4%
   5.8%   512Ki _ZN2cvL9LvToVpl_bE             0   0.0%
   2.9%   256Ki _ZN2cvL13Lanczos4Tab_fE        0   0.0%
   2.9%   256Ki _ZN2cvL8LuToUp_bE              0   0.0%
   0.0%     709 [None]                         0   0.0%
 100.0%  8.65Mi TOTAL                     4.04Mi 100.0%
```
After:
```
     VM SIZE                                               FILE SIZE
 --------------                                         --------------
  66.7%  2.80Mi [Other]                                  2.68Mi  65.9%
  20.5%   880Ki [Unmapped]                               1.35Mi  33.2%
   6.0%   256Ki _ZN2cvL13Lanczos4Tab_fE                       0   0.0%
   3.0%   128Ki _ZN2cvL13Lanczos4Tab_iE                       0   0.0%
   1.5%  64.0Ki _ZN2cvL12BicubicTab_fE                        0   0.0%
   1.5%  64.0Ki _ZN2cvL12trilinearLUTE                        0   0.0%
   0.8%  36.1Ki _ZNK2cv15ThresholdRunnerclERKNS_5RangeE  36.1Ki   0.9%
   0.0%     609 [None]                                        0   0.0%
 100.0%  4.19Mi TOTAL                                    4.06Mi 100.0%
```

vpisarev:
@mshabunin, thanks! Let's merge it in :+1:

PR #9933, opened 2017-10-24
Adding new edge detection

fiammante:
This is an example for a new algorithm to detect edges based on the absolute difference of a gaussian blur and a blur, followed by histogram equalization. It also adds the computation of local minimum to contrast the edges. 
Change consists of 3 files: brightedges.cpp with the function, brightedges.h with the correspondign header file, and Edges.cpp for testing with  data\butterfly.jpg as default.
Here is the result for data\butterfly.jpg. 
All tested locally with Visualstudio. The code in brightedges.cpp and brightedges.h could be added as a function in imgproc.
sampleedgedetection


dkurt:
@fiammante, Do you know some applications to this edge detection algorithm? In example, that seems to me hard to find contours because of detailzation. Any comparison to well known methods?

fiammante:
@dkurt the detailization depends on the image. Butterfly images has a lot of contours so you get a lot of contours. You can tune the two blur kernel sizes parameters and the contrast parameter to reduce the number of details, if you have a small image keep the defaults, if you have a large image increase both kernel sizes. Reducing the contrast parameter in the funciton call reduces the details.
I use it successfully with low resolution images in one of my project to identify objects.
See another example in my answer here http://opencv.org/question/164533/how-to-smooth-the-edges-of-a-low-quality-image/
Here is a comparison with Canny and Sobel based on a picture of me:
comparison

alalek:
Thank you for contribution!

Edge detector is usually not used as standalone algorithm.
Its output is passed for forward processing - for example, it is used as input for `findContours` (returns contours information, not image itself).
Also "strong" edges look broken (with holes).

In case of low quality / noisy images, `bilateralFilter()` may be applied (hide high-frequency noise, make "edges" sharpen) or `medianFilter()` (faster, but usually enough in some cases).
Also pre-processing based on "sharpening" filters can be used (but it increases noise).

BTW, Also you may look on thinning algorithm: https://github.com/opencv/opencv_contrib/pull/751

From technical side, requirements for contributions are:
- new algorithms, experiments should go into [opencv_contrib](https://github.com/opencv/opencv_contrib) repository.
- algorithm should be packages into module, "samples" is not a good place for this
- some simple tests should be added (at least to check that provided algorithm doesn't crash)

fiammante:
@alalek Thank you for your comments. I am not using this algorithm stand alone. In fact in a project with medical ultrasound like images, I tried all of the others including the ones listed above,, cartoonifcation  with pyrDown and bilateral filters plus various other and also CNNs. All details are important in any diagnostic from an image even small ones. Using this one with further processing I succeeded where the others did not work. Strong edges can be recovered by further convolutions not part of the algorithm.
So a different algorithm for a somewhat different purpose. 
Part of what I included is a test program.
However if there is no interest here I'll drop it and maybe share it somewhere else. Given the results I had in my projects I thought it could interest other people.

PR #9939, opened 2017-10-25
dnn: fix sporadic crashes in getUMat()

alalek:
Incorrect "total" buffer size is calculated in StdMatAllocator::allocate() due wrong step values.

Problem is [observed](http://pullrequest.opencv.org/buildbot/builders/precommit_opencl/builds/12965/steps/test_dnn-ippicv-opencl/logs/stdio) via these sporadic failures:
```
[ RUN      ] Test_TensorFlow.conv
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
[  FAILED  ] Test_TensorFlow.conv (10 ms)
```


vpisarev:
:+1:

PR #9941, opened 2017-10-25
Improve C++ DNN object detection samples

catree:
### This pullrequest changes

- Add camera/video/image input for C++ DNN object detection samples. 
- Add nice display and computation time.
- Add some information about the default network.

### DIFF
--- a/samples/dnn/ssd_mobilenet_object_detection.cpp
+++ b/samples/dnn/ssd_mobilenet_object_detection.cpp
@@ -23,23 +23,25 @@ const char* classNames[] = {"background",
                             "motorbike", "person", "pottedplant",
                             "sheep", "sofa", "train", "tvmonitor"};
 
-const char* about = "This sample uses Single-Shot Detector "
-                    "(https://arxiv.org/abs/1512.02325)"
-                    "to detect objects on image.\n"
-                    ".caffemodel model's file is avaliable here: "
-                    "https://github.com/chuanqi305/MobileNet-SSD\n";
+const char* about = "This sample uses MobileNet Single-Shot Detector "
+                    "(https://arxiv.org/abs/1704.04861) "
+                    "to detect objects on camera/video/image.\n"
+                    ".caffemodel model's file is available here: "
+                    "https://github.com/chuanqi305/MobileNet-SSD\n"
+                    "Default network is 300x300 and 20-classes VOC.\n";
 
 const char* params
     = "{ help           | false | print usage         }"
       "{ proto          | MobileNetSSD_deploy.prototxt | model configuration }"
       "{ model          | MobileNetSSD_deploy.caffemodel | model weights }"
-      "{ video          |       | video for detection }"
+      "{ camera_device  | 0     | camera device number }"
+      "{ video          |       | video or image for detection}"
       "{ out            |       | path to output video file}"
       "{ min_confidence | 0.2   | min confidence      }";
 
 int main(int argc, char** argv)
 {
-    cv::CommandLineParser parser(argc, argv, params);
+    CommandLineParser parser(argc, argv, params);
 
     if (parser.get<bool>("help"))
     {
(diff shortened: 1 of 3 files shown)


samples/dnn/ssd_mobilenet_object_detection.cpp
```diff
@@ -55,19 +58,40 @@ int main(int argc, char** argv)
...
+        cerr << "prototxt:   " << modelConfiguration << endl;
+        cerr << "caffemodel: " << modelBinary << endl;
+        cerr << "Models can be downloaded here:" << endl;
+        cerr << "https://github.com/chuanqi305/MobileNet-SSD" << endl;
+        exit(-1);
+    }
+
+    Mat img;
+    VideoCapture cap;
+    if (parser.get<String>("image").empty()) {
```
alalek: `parser.get<String>("image")` used here 5 times in this example. Probably we could save this value into some variable:
```
String imagePath = parser.get<String>("image");
if (imagePath.empty())
    ...
```

samples/dnn/ssd_mobilenet_object_detection.cpp
```diff
@@ -55,19 +58,40 @@ int main(int argc, char** argv)
...
+            }
         }
     }
+    else
+        img = imread(parser.get<String>("image"));
 
-    Size inVideoSize = Size((int) cap.get(CV_CAP_PROP_FRAME_WIDTH),    //Acquire input size
-                            (int) cap.get(CV_CAP_PROP_FRAME_HEIGHT));
+    Size inVideoSize;
+    if (parser.get<String>("image").empty())
```
alalek: or we could use this condition here and below: `if (img.empty())`

samples/dnn/ssd_mobilenet_object_detection.cpp
```diff
@@ -55,19 +58,40 @@ int main(int argc, char** argv)
...
+            cap = VideoCapture(cameraDevice);
+            if(!cap.isOpened())
+            {
+                cout << "Couldn't find camera: " << cameraDevice << endl;
+                return -1;
+            }
         }
     }
+    else
+        img = imread(parser.get<String>("image"));
```
alalek:
```
if (img.empty())
{
    error message: missing file or unsupported image format
}
```

samples/dnn/ssd_mobilenet_object_detection.cpp
dkurt: We also can  pass image path to VideoCapture. Extra waitKey might be after the last frame processed.

catree:
I have updated the samples:
- capture is also used to read an image
- I have removed some console output I don't find useful

samples/dnn/ssd_mobilenet_object_detection.cpp
```diff
@@ -86,16 +106,24 @@ int main(int argc, char** argv)
               cropSize);
 
     VideoWriter outputVideo;
-    outputVideo.open(parser.get<String>("out") ,
-                     static_cast<int>(cap.get(CV_CAP_PROP_FOURCC)),
-                     cap.get(CV_CAP_PROP_FPS), cropSize, true);
+    bool readSingleImage = !parser.get<String>("image").empty() && (int) cap.get(CV_CAP_PROP_FRAME_COUNT) == 1;
```
catree: I cannot do `bool readSingleImage = (int) cap.get(CV_CAP_PROP_FRAME_COUNT) == 1` as I got `VIDEOIO ERROR: V4L2: getting property #7 is not supported` when opening a camera stream.

dkurt: @catree, it seems to me a kind of complicated. Please try to use a single argument `video` as input for VideoCapture: `VideoCapture cap(parser.get<String>("video"));`. That works fine as for single image, set of images or for a video.

catree: Ok, I have removed the image option.

samples/dnn/ssd_mobilenet_object_detection.cpp
```diff
@@ -55,19 +57,32 @@ int main(int argc, char** argv)
...
+    cap.open(parser.get<String>("video"));
     if(!cap.isOpened()) // check if we succeeded
     {
-        cap = VideoCapture(0);
+        int cameraDevice = parser.get<int>("camera_device");
+        cap = VideoCapture(cameraDevice);
         if(!cap.isOpened())
         {
-            cout << "Couldn't find camera" << endl;
+            cout << "Couldn't find camera: " << cameraDevice << endl;
```
dkurt: @catree, If user pass a wrong path to video file he has an unexpected error `Couldn't find camera: 0`. What do you think about the following logic: if parser has no `video` argument - open a camera stream. Otherwise, if video file didn't open - print a error like `cout << "Couldn't open image or video: " << parser.get<String>("video") << endl;`

catree: @dkurt I am ok with that.

vpisarev:
thanks! :+1:

PR #9945, opened 2017-10-26
add nvidia hardware encode/decode support in videoio with ffmpeg

yaochx:
```
force_builders=CUDA
```

### DIFF
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -168,6 +168,16 @@ extern "C" {
 #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


alalek:
No need to re-open pull requests just update your source branch (with `--force`).

mshabunin:
@yaochx , as I can see this PR contains changes to allow building with latest _master_ FFMpeg, but where is actual HW processing support? Are you sure it will work without `av_hwdevice_*`, `av_hwframe_*` calls? ([av_hwdevice_ctx_init](https://ffmpeg.org/doxygen/3.1/hwcontext_8c.html#a11d8c9101c95cdcab3dbc5b4c03dfd4d))

yaochx:
@mshabunin Thank you for advising. It doesn't work, I will try to work on it.

PR #9946, opened 2017-10-26
OCL: increase SURF stitching test tolerance

alalek:
OpenCL SURF implementation is unstable

Message example:
```
C:\build\precommit_opencl\opencv\modules\stitching\perf\opencl\perf_stitch.cpp(96): error: The difference between pano.size().width and 1124 is 58, which exceeds 50, where
pano.size().width evaluates to 1066,
1124 evaluates to 1124, and
50 evaluates to 50.
```


alalek:
:+1:

PR #9950, opened 2017-10-27
Fix CMake issues on OpenBSD

sizeofvoid:
### This pullrequest changes
Will fix some build issues on OpenBSD. Patches are made by the OpenBSD OpenCV maintainer.

### DIFF
--- a/cmake/OpenCVPCHSupport.cmake
+++ b/cmake/OpenCVPCHSupport.cmake
@@ -16,6 +16,11 @@ IF(CMAKE_COMPILER_IS_GNUCXX)
 
     IF(NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.2.0")
         SET(PCHSupport_FOUND TRUE)
+        IF(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+            SET(PCHSupport_FOUND FALSE)
+        ELSE(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+            SET(PCHSupport_FOUND TRUE)
+        ENDIF()
     ENDIF()
 
     SET(_PCH_include_prefix "-I")
--- a/modules/java/CMakeLists.txt
+++ b/modules/java/CMakeLists.txt
@@ -485,13 +485,25 @@ if(ANDROID)
           ARCHIVE DESTINATION ${OPENCV_LIB_INSTALL_PATH} COMPONENT java)
 else()
   if(NOT INSTALL_CREATE_DISTRIB)
-    ocv_install_target(${the_module} OPTIONAL EXPORT OpenCVModules
-            RUNTIME DESTINATION ${OPENCV_JAR_INSTALL_PATH} COMPONENT java
-            LIBRARY DESTINATION ${OPENCV_JAR_INSTALL_PATH} COMPONENT java)
+    if(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+      ocv_install_target(${the_module}
+              RUNTIME DESTINATION ${OPENCV_JAR_INSTALL_PATH} COMPONENT java
+              LIBRARY DESTINATION ${OPENCV_JAR_INSTALL_PATH} COMPONENT java)
+    else(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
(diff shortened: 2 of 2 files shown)


cmake/OpenCVPCHSupport.cmake
```diff
@@ -16,6 +16,11 @@ IF(CMAKE_COMPILER_IS_GNUCXX)
 
     IF(NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.2.0")
         SET(PCHSupport_FOUND TRUE)
+        IF(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+            SET(PCHSupport_FOUND FALSE)
+        ELSE(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
```
vpisarev: I believe, else section is not needed here

modules/java/CMakeLists.txt
```diff
@@ -485,13 +485,25 @@ if(ANDROID)
...
 else()
   if(NOT INSTALL_CREATE_DISTRIB)
-    ocv_install_target(${the_module} OPTIONAL EXPORT OpenCVModules
-            RUNTIME DESTINATION ${OPENCV_JAR_INSTALL_PATH} COMPONENT java
-            LIBRARY DESTINATION ${OPENCV_JAR_INSTALL_PATH} COMPONENT java)
+    if(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+      ocv_install_target(${the_module}
+              RUNTIME DESTINATION ${OPENCV_JAR_INSTALL_PATH} COMPONENT java
+              LIBRARY DESTINATION ${OPENCV_JAR_INSTALL_PATH} COMPONENT java)
+    else(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
```
vpisarev: we usually omit duplication of the condition inside else(). it really looks very confusing

vpisarev:
@alalek, could you review it? looks reasonable to me

cmake/OpenCVPCHSupport.cmake
```diff
@@ -16,6 +16,11 @@ IF(CMAKE_COMPILER_IS_GNUCXX)
 
     IF(NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.2.0")
         SET(PCHSupport_FOUND TRUE)
+        IF(${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
```
alalek: Precompiled headers is compiler/build system feature, not a specific platform. Does OpenBSD use GCC? What is the problem?

modules/java/CMakeLists.txt
```diff
@@ -485,13 +485,25 @@ if(ANDROID)
           ARCHIVE DESTINATION ${OPENCV_LIB_INSTALL_PATH} COMPONENT java)
 else()
   if(NOT INSTALL_CREATE_DISTRIB)
-    ocv_install_target(${the_module} OPTIONAL EXPORT OpenCVModules
```
alalek: I believe we should drop `EXPORT OpenCVModules` for "bindings" by default (it is really confusing).
But we need to test Android build more carefully (they like to link with libopencv_java.so).

cmake/OpenCVPCHSupport.cmake
sizeofvoid: yes, your right. Thanks

cmake/OpenCVPCHSupport.cmake
sizeofvoid: We build ports with clang and GCC depending on the arch. Thus PCH is a nogo.

alalek: > depending on the arch

Do you use CMake cross-compiling mode? CMake toolchain? Do you see CMAKE_CROSSCOMPILING variable?

Probably you could turn off [`ENABLE_PRECOMPILED_HEADERS`](https://github.com/opencv/opencv/blob/3.3.1/CMakeLists.txt#L295) in your build scripts.

alalek:
opencv_java issue is resolved here: #10001
Pre-compiled headers can be turned off via build scripts (CMake option/`CONFIGURE_ARGS`):
```
-DENABLE_PRECOMPILED_HEADERS=OFF
```

PR #9952, opened 2017-10-27
Add C++ DNN face detection sample: resnet_ssd_face.cpp

catree:
### This pullrequest changes
Translate the Python sample [resnet_ssd_face_python.py](https://github.com/opencv/opencv/blob/master/samples/dnn/resnet_ssd_face_python.py) to C++.


vpisarev:
thanks! :+1:

Issue #9953, opened 2017-10-27
Building against OpenBLAS complains about missing lapacke.h


FlorianFranzen:
##### System information (version)
- OpenCV => 3.3.1
- Operating System / Platform => Ubuntu 16.04. (xenial)

##### Detailed description

It is not possible to build openCV against libOpenBLAS unless one also installs an additional LAPACK implementation that includes the ```lapacke.h``` header, e.g. ```liblapacke-dev``` on Ubuntu.

To me that does not look like the intended behavior and it might be explained by the fact, that OpenBLAS on Ubuntu only include ```cblas.h``` and ```f77blas.h``` as an header. 

On the other hand ```OpenCVFindOpenBLAS.cmake``` does not check for ```lapacke.h```, so it might just be the case that ```ocv_lapack_check()``` is too strict when it comes to OpenBLAS.

@alalek You seem to have some experience with this. Any thoughts on this?

##### Steps to reproduce

If I only install ```libopenblas-dev```, I get the following cmake output once ```ocv_lapack_check()``` inside ```OpenCVFindLAPACK.cmake``` is called:

```
-- LAPACK(OpenBLAS): LAPACK_LIBRARIES: /usr/lib/libopenblas.so
CMake Warning at cmake/OpenCVFindLAPACK.cmake:29 (message):
  LAPACK(OpenBLAS): CBLAS/LAPACK headers are not found in '/usr/include'
Call Stack (most recent call first):
  cmake/OpenCVFindLAPACK.cmake:155 (ocv_lapack_check)
  CMakeLists.txt:596 (include)

```

If I now install any LAPACK implementation that includes a ```lapacke.h```, e.g.: ```liblapacke-dev```, everything passes and OpenBLAS will be used:

```
-- LAPACK(OpenBLAS): LAPACK_LIBRARIES: /usr/lib/libopenblas.so
-- LAPACK(OpenBLAS): Support is enabled.
```

alalek:
BLAS is API with functions for low-level matrix/vector operations (like gemm).
OpenCV requires LAPACK API too (high-level functionality, like matrix factorization, eigenvalues, etc): http://www.netlib.org/lapack/

For some reason, CMake's [FindLAPACK](https://cmake.org/cmake/help/v2.8.12/cmake.html#module:FindLAPACK) module doesn't provide any information about header files (they are required for C/C++ programs compilation).

Ubuntu's OpenBLAS binaries support LAPACK API symbols.
```
[...]

PR #9955, opened 2017-10-28
cmake: update OpenBLAS support

alalek:
- reorder path HINTS to avoid conflicts with other LAPACK libraries
- add `extern "C" { }` to avoid generation of C++ symbols

related #9953


vpisarev:
:+1:

Issue #9958, opened 2017-10-29
 outputs remain the same bug in detection_ouput


lzx1413:
##### System information (version)

- OpenCV => 3.1
- Operating System / Platform => Windows 64 Bit
- Compiler => Visual Studio 2015

##### Detailed description

If the detection result keepnum in detection_output_layer.cpp is 0, it will return directly. Because the net did not clear the output blob after every callback. The result of last image will remain as this image's result.
##### Steps to reproduce
set the confidence_threshold in mobilenet_ssd.prototxt to 0.5, we want the detection_output_layer's keepnum to be 0.
run the mobilenet_ssd_python.py with a normal image and a black image continuously. 
The black image will show the same result on the last image.
#####suggestion
we can reshape the outputmat to be 0 or  release it directly.

PR #9959, opened 2017-10-30
Reset confidences to zeros in case of no detections

dkurt:
### This pullrequest changes

* Test for MobileNet-SSD model from Caffe
* Reset out of date detections' confidences to zero in case of no detections on the frame

resolves https://github.com/opencv/opencv/issues/9958

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/402

### DIFF
--- a/modules/dnn/src/layers/detection_output_layer.cpp
+++ b/modules/dnn/src/layers/detection_output_layer.cpp
@@ -240,6 +240,9 @@ public:
 
         if (numKept == 0)
         {
+            // Set confidences to zeros.
+            Range ranges[] = {Range::all(), Range::all(), Range::all(), Range(2, 3)};
+            outputs[0](ranges).setTo(0);
             return;
         }
         int outputShape[] = {1, 1, (int)numKept, 7};
--- a/modules/dnn/test/test_caffe_importer.cpp
+++ b/modules/dnn/test/test_caffe_importer.cpp
@@ -139,6 +139,35 @@ TEST(Reproducibility_SSD, Accuracy)
     normAssert(ref, out);
 }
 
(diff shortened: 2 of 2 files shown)


lzx1413:
This will increase the time of SSD in fact, not suggest to do this

dkurt:
@lzx1413, please correct me if I missed something. This layer has no fixed output size, so it reallocates memory as needed according to number of detections. In case of no detections we can't return just an empty Mat because (1) there are applications that don't check it (2) in general words it's just a layer and output of it may be used as input for another one.

We can reshape it to `1x1x1x7` but what should we do with confidence? Setting it to zero seems to be the best way. This way it doesn't matter how much detections was lately. So there is no reallocation in this patch.

You can prepare efficiency measurements: run [mobilenet_ssd_python.py](https://github.com/opencv/opencv/blob/master/samples/dnn/mobilenet_ssd_python.py) with and without PR's changes and check the best observed time.

lzx1413:
https://github.com/opencv/opencv/blob/fd1eb63feeb7df1dae8d2b8c584d775c1da23709/modules/dnn/src/nms.inl.hpp#L67 because this function is controlled both by topk and the confidence_threshold, if we set confidence_threshold to 0, the nms will always have to process all of the topk bboxes constantly, which will be time consuming considering the image with no objects.

dkurt:
@lzx1413, no problem. Did you mean that it fills zeros every forward pass?

lzx1413:
no, I thought you changed the confidence_threshold in the layer parameter, this will lead to the result that the nms has to process all of the topk results everytime

vpisarev:
looks good! :+1:

PR #9963, opened 2017-10-30
Specify layers types at FP32->FP16 Caffe models converter

dkurt:
### This pullrequest changes

Specify layers types for FP32->FP16 conversion of Caffe models' weights. Otherwise we convert all the floats including sensitive values like PriorBox layer parameters (aspect ratio, variance etc.).


vpisarev:
Since we do not have ABI compatibility requirements for dnn module, I would simple add optional parameter, which is `vector<String>()` by default, and when it's empty, convert it to the default 2-element vector.

dkurt:
@vpisarev, thanks for review, done!

vpisarev:
thanks! :+1:

PR #9970, opened 2017-10-30
MediaSDK: fixed Linux build, improved BGR<->NV12 conversions

mshabunin:
### This pullrequest changes

- fixed compilation on Linux
- use `cv::hal::` functions to convert between BGR and NV12 formats in MediaSDK backend

**Notes:**
- performance improvement ~3-5% for `opencv_test_videoio --gtest_filter=*MFX*` (encode-decode test)
- -2 KiB to binary size

### DIFF
--- a/modules/imgproc/include/opencv2/imgproc/hal/hal.hpp
+++ b/modules/imgproc/include/opencv2/imgproc/hal/hal.hpp
@@ -192,6 +192,12 @@ CV_EXPORTS void cvtTwoPlaneYUVtoBGR(const uchar * src_data, size_t src_step,
                                     int dst_width, int dst_height,
                                     int dcn, bool swapBlue, int uIdx);
 
+//! Separate Y and UV planes
+CV_EXPORTS void cvtTwoPlaneYUVtoBGR(const uchar * y_data, const uchar * uv_data, size_t src_step,
+                                    uchar * dst_data, size_t dst_step,
+                                    int dst_width, int dst_height,
+                                    int dcn, bool swapBlue, int uIdx);
+
 CV_EXPORTS void cvtThreePlaneYUVtoBGR(const uchar * src_data, size_t src_step,
                                       uchar * dst_data, size_t dst_step,
                                       int dst_width, int dst_height,
@@ -201,6 +207,12 @@ CV_EXPORTS void cvtBGRtoThreePlaneYUV(const uchar * src_data, size_t src_step,
                                       uchar * dst_data, size_t dst_step,
                                       int width, int height,
                                       int scn, bool swapBlue, int uIdx);
+
+//! Separate Y and UV planes
+CV_EXPORTS void cvtBGRtoTwoPlaneYUV(const uchar * src_data, size_t src_step,
+                                    uchar * y_data, uchar * uv_data, size_t dst_step,
+                                    int width, int height,
+                                    int scn, bool swapBlue, int uIdx);
 
 CV_EXPORTS void cvtOnePlaneYUVtoBGR(const uchar * src_data, size_t src_step,
                                     uchar * dst_data, size_t dst_step,
--- a/modules/imgproc/src/color.cpp
+++ b/modules/imgproc/src/color.cpp
@@ -9050,44 +9050,60 @@ inline void cvtYUV420p2RGBA(uchar * dst_data, size_t dst_step, int dst_width, in
 
 ///////////////////////////////////// RGB -> YUV420p /////////////////////////////////////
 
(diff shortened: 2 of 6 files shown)


modules/imgproc/include/opencv2/imgproc/hal/hal.hpp
```diff
@@ -192,6 +192,12 @@ CV_EXPORTS void cvtTwoPlaneYUVtoBGR(const uchar * src_data, size_t src_step,
                                     int dst_width, int dst_height,
                                     int dcn, bool swapBlue, int uIdx);
 
+//! Separate Y and UV planes
+CV_EXPORTS void cvtTwoPlaneYUVtoBGR(const uchar * y_data, const uchar * uv_data, size_t src_step,
```
vpisarev: In IPP they use convention like cvtYUVtoBGR_P2. Probably, we should use it as well.

modules/imgproc/src/color.cpp
```diff
@@ -9105,40 +9113,40 @@ struct RGB888toYUV420pInvoker: public ParallelLoopBody
...
     {
-        return (src_width * src_height >= 320*240);
+        if( src_width * src_height >= 320*240 )
+            parallel_for_(Range(0, src_height/2), *this);
+        else
+            operator()(Range(0, src_height/2));
     }
 
 private:
     RGB888toYUV420pInvoker& operator=(const RGB888toYUV420pInvoker&);
```
vpisarev: do we need hidden = operator? probably, "const int scn" can be changed to "int scn". Since `operator()(const Range& r)` is `const` method, all the class members are automatically treated as constants

modules/imgproc/src/color.cpp
```diff
@@ -9105,40 +9113,40 @@ struct RGB888toYUV420pInvoker: public ParallelLoopBody
...
 
     const uchar * src_data;
     size_t src_step;
-    uchar * dst_data;
+    uchar *y_data, *uv_data;
     size_t dst_step;
     int src_width;
     int src_height;
     const int scn;
```
vpisarev: as mentioned above, `const` is unnecessary here

modules/imgproc/src/color.cpp
```diff
@@ -10777,25 +10785,36 @@ void cvtLabtoBGR(const uchar * src_data, size_t src_step,
...
     {
-    case 300: cvtYUV420sp2RGB<0, 0> (dst_data, dst_step, dst_width, dst_height, src_step, src_data, uv); break;
-    case 301: cvtYUV420sp2RGB<0, 1> (dst_data, dst_step, dst_width, dst_height, src_step, src_data, uv); break;
-    case 320: cvtYUV420sp2RGB<2, 0> (dst_data, dst_step, dst_width, dst_height, src_step, src_data, uv); break;
-    case 321: cvtYUV420sp2RGB<2, 1> (dst_data, dst_step, dst_width, dst_height, src_step, src_data, uv); break;
-    case 400: cvtYUV420sp2RGBA<0, 0>(dst_data, dst_step, dst_width, dst_height, src_step, src_data, uv); break;
-    case 401: cvtYUV420sp2RGBA<0, 1>(dst_data, dst_step, dst_width, dst_height, src_step, src_data, uv); break;
-    case 420: cvtYUV420sp2RGBA<2, 0>(dst_data, dst_step, dst_width, dst_height, src_step, src_data, uv); break;
-    case 421: cvtYUV420sp2RGBA<2, 1>(dst_data, dst_step, dst_width, dst_height, src_step, src_data, uv); break;
+    case 300: cvtYUV420sp2RGB<0, 0> (dst_data, dst_step, dst_width, dst_height, src_step, y_data, uv_data); break;
```
vpisarev: as we are trying to make OpenCV smaller, perhaps using real function parameters instead of template parameters is probably preferable.

modules/imgproc/src/color.cpp
alalek: This is for static code analyzers. There is no implementation. This is just a BKM.

In C++11 this should look like this:
```
RGB888toYUV420pInvoker& operator=(const RGB888toYUV420pInvoker&) = delete;
```
The same for copy constructor.

Without this compiler would generate default operator (with invalid imlementation), which may be used by mistake.

mshabunin:
@vpisarev , I've refactored this function to reduce binary size, performance seem to be the same (at least on my machine).
The function is called so to be uniform with other `hal::cvt*` functions, I think it will be better to rename them all at once some time later.

alalek:
:+1:

PR #9986, opened 2017-11-01
Add more details to drawContour description

### LINKED ISSUE #5256: drawContours, strange behavior
I have a list of contours to draw. Some of these contours intersect themselves.

When I want to draw them with OpenCV, I simply use the `cv::drawContours` function.

However, the behavior is quite strange.

Here is a quote of the official [documentation](http://docs.opencv.org/modules/imgproc/doc/structural_analysis_and_shape_descriptors.html#drawcontours)

```
C++: void drawContours(InputOutputArray image, InputArrayOfArrays contours, int contourIdx, const Scalar& color, int thickness=1, int lineType=8, InputArray hierarchy=noArray(), int maxLevel=INT_MAX, Point offset=Point() )
Parameters: 
contourIdx – Parameter indicating a contour to draw. If it is negative, all the contours are drawn.
```

So, from the documentation, if I want to draw all my areas, filled in black, I just have to do:

```
cv::drawContours(this->mask.raw,
                 this->areas, -1,
                 cv::Scalar(0,0,0),
                 cv::FILLED);
```

However, this gives me the following ouput:

[enter image description here](http://i.stack.imgur.com/IwYMe.png)

Here, we can clearly see that all my areas are NOT filled in black.

But if I loop over my areas list and call `cv::drawContours` for each area:

```
unsigned int i = 0;
for (const auto& area : this->areas)
  cv::drawContours(this->mask.raw,
                   this->areas, i++,
                   cv::Scalar(0,0,0),
                   cv::FILLED);
```

I got the good ouput which is quite different from the first one:

[enter image description here](http://i.stack.imgur.com/6oYV9.png)

Have I missed something from the documentation? Could someone explain me the behavior of `cv::drawContours` and what is the different of calling it one time for all areas and calling it each time for each area?

terfendail:
resolves #5256
### This pullrequest changes
Added `drawContours` interrior filling details description

### DIFF
--- a/modules/imgproc/include/opencv2/imgproc.hpp
+++ b/modules/imgproc/include/opencv2/imgproc.hpp
@@ -4584,6 +4584,8 @@ draws the contours, all the nested contours, all the nested-to-nested contours, 
 parameter is only taken into account when there is hierarchy available.
 @param offset Optional contour shift parameter. Shift all the drawn contours by the specified
 \f$\texttt{offset}=(dx,dy)\f$ .
+@note In case function is called to draw all contours with filled interior (negative contourIdx and thickness) it process nested contours as shapes with holes.
+In this mode function fill areas inner to odd amount of contours and avoid filling areas inner to even amount of contours using even-odd rule over all edges of all contours.
  */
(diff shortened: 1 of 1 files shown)


vpisarev:
:+1:

PR #9994, opened 2017-11-02
load DNN from memory buffer

### LINKED ISSUE #9765: Feature Request : ability to load Caffe/Tensorflow from FileStorage or from similar in-memory

- OpenCV =>  3.3
- Operating System / Platform => macOS 10.11
- Compiler => Xcode 8.2.1 / clang-800.0.42.1

##### Detailed description

Like we could load cv::ml::SVM from a cv::FileStorage :

~~~
cv::FileStorage fs;
m_svm = cv::Algorithm::read<cv::ml::SVM>(fs["SVM"]);
~~~

It will look like : `cv::dnn::Net = readNetFromTensorflow(fs["Model"]);`

r2d3:
resolves #9765

### This pullrequest changes

add an overloaded readNetFromTensorflow/readNetFromCaffe functions to read a TF/Caffe Net from data already in memory

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -634,10 +634,32 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
       */
     CV_EXPORTS_W Net readNetFromCaffe(const String &prototxt, const String &caffeModel = String());
 
+    /** @brief Reads a network model stored in Caffe model in memory.
+      * @details This is an overloaded member function, provided for convenience.
+      * It differs from the above function only in what argument(s) it accepts.
+      * @param bufferProto buffer containing the content of the .prototxt file
+      * @param lenProto length of bufferProto
+      * @param bufferModel buffer containing the content of the .caffemodel file
+      * @param lenModel length of bufferModel
+      */
+    CV_EXPORTS Net readNetFromCaffe(const char *bufferProto, size_t lenProto,
+                                    const char *bufferModel = NULL, size_t lenModel = 0);
+
     /** @brief Reads a network model stored in Tensorflow model file.
       * @details This is shortcut consisting from createTensorflowImporter and Net::populateNet calls.
       */
     CV_EXPORTS_W Net readNetFromTensorflow(const String &model, const String &config = String());
+
+    /** @brief Reads a network model stored in Tensorflow model in memory.
+      * @details This is an overloaded member function, provided for convenience.
+      * It differs from the above function only in what argument(s) it accepts.
+      * @param bufferModel buffer containing the content of the pb file
+      * @param lenModel length of bufferModel
+      * @param bufferConfig buffer containing the content of the pbtxt file
+      * @param lenConfig length of bufferConfig
+      */
+    CV_EXPORTS Net readNetFromTensorflow(const char *bufferModel, size_t lenModel,
+                                         const char *bufferConfig = NULL, size_t lenConfig = 0);
 
     /** @brief Reads a network model stored in Torch model file.
       * @details This is shortcut consisting from createTorchImporter and Net::populateNet calls.
--- a/modules/dnn/src/tensorflow/tf_io.cpp
+++ b/modules/dnn/src/tensorflow/tf_io.cpp
@@ -23,6 +23,7 @@ Implementation of various functions which are related to Tensorflow models readi
 
 #include "graph.pb.h"
 #include "tf_io.hpp"
+#include "../caffe/caffe_io.hpp"
 #include "../caffe/glog_emulator.hpp"
 
 namespace cv {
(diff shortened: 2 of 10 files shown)


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -639,6 +639,11 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
       */
     CV_EXPORTS_W Net readNetFromTensorflow(const String &model, const String &config = String());
 
+    /** @brief Reads a network model stored in Tensorflow model in memory
+     */
+    CV_EXPORTS_W Net readNetFromTensorflowBuffer(const char *bufferModel, size_t lenModel,
```
dkurt: Let's keep the same name for it, `readNetFromTensorflow`.

I think we can use brief description like at https://docs.opencv.org/master/d3/d63/classcv_1_1Mat.html#a2ec3402f7d165ca34c7fd6e8498a62ca: `This is an overloaded member function...` where only parameters are described.

Have you tested python / Java bindings for it?

modules/dnn/src/tensorflow/tf_io.cpp
```diff
@@ -52,6 +52,18 @@ bool ReadProtoFromBinaryFileTF(const char* filename, Message* proto) {
     return success;
 }
 
+bool ReadProtoFromBinaryBufferTF(const char* data, size_t len, Message* proto) {
```
dkurt: I think it's better to take existing two methods, remove `std::ifstream` creation from them and replace `const char* filename` argument to `std::istream& is` one.
In result, we'll have something like 

```cpp
bool ReadProtoFromTextFileTF(std::istream& s, Message* proto) {
    IstreamInputStream input(&is);
    return google::protobuf::TextFormat::Parse(&input, proto);
}
```
```cpp
bool ReadProtoFromTextFileTF(const char* buf, size_t len, Message* proto) {
    std::istringstream s(std::string(buf, len));
    return ReadProtoFromTextFileTF(s, proto);
}
```
```cpp
bool ReadProtoFromTextFileTF(const char* filename, Message* proto) {
    std::ifstream fs(filename, std::ios::in);
    CHECK(fs.is_open()) << "Can't open \"" << filename << "\"";
    return ReadProtoFromTextFileTF(fs, proto);
}
```

r2d3: The problem with std::istringstream s(std::string(buf, len)) is that, in some cases, it creates a copy of the buffer into a std::string.

However ArrayInputStream input(data, len) guarantees that there is no copy
https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.io.zero_copy_stream_impl_lite#ArrayInputStream

dkurt: @r2d3, we just vote to reduce a code duplication. There are two methods `ReadProtoFromBinaryFileTF` and `ReadProtoFromBinaryBufferTF` are different only at the beginning for now.

r2d3: OK I will reduce code duplication

modules/dnn/test/test_tf_importer.cpp
```diff
@@ -69,20 +69,61 @@ TEST(Test_TensorFlow, inception_accuracy)
     normAssert(ref, out);
 }
 
+static bool readFileInMemory(const string& filename, char** data, size_t* len)
```
dkurt: @r2d3, do you mind if we return a content by `std::string`? In example,
```cpp
static bool readFileInMemory(const string& filename, std::string* content) {
  std::ifstream ifs(filename.c_str());
  CV_Assert(ifs.is_open());

  ifs.seekg(0, std::ios::end);
  content->resize(ifs.tellg());
  ifs.seekg(0, std::ios::beg);
  ifs.read(&content->operator[](0), content->size());
}
```

r2d3: No, I was thinking about it as it is how I do file loading in personal tests

r2d3: Done

modules/dnn/src/tensorflow/tf_io.cpp
r2d3: I put the common code in `ReadProtoFromBinaryTF`.

I simplified the code in other functions. `raw_input` and `coded_input` are now created on stack to avoid new/delete.

modules/dnn/test/test_tf_importer.cpp
```diff
@@ -69,20 +69,55 @@ TEST(Test_TensorFlow, inception_accuracy)
...
+        string dataModel;
+        ASSERT_TRUE(readFileInMemory(netPath, dataModel));
+
+        string dataConfig;
+        if (hasText)
+            ASSERT_TRUE(readFileInMemory(netConfig, dataConfig));
+
+        net = readNetFromTensorflow(dataModel.c_str(), dataModel.size(),
+                                    dataConfig.c_str(), dataConfig.size());
+        ASSERT_FALSE(net.empty());
```
dkurt: I think this check is useful for both ways of importing.

r2d3: You are right, test is now done for both ways of importing

r2d3:
Hello @dkurt 

As TensorFlow from memory is now working. I began working on Caffe one. I began moving all the ReadProtoFrom* functions to caffe_io.cpp and using them in tf_io.cpp.

r2d3:
So readNetFromCaffe is now also implemented with a corresponding test (I had to duplicate the `readFileInMemory` function, there is place to put these test utility functions)

I do not know if we could implement the same mechanism for DarkNet and Torch ?!

Regards

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -634,11 +634,33 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
 
+    /** @brief Reads a network model stored in Caffe model in memory.
+      * @details This is an overloaded member function, provided for convenience.
+      * It differs from the above function only in what argument(s) it accepts.
+      * @param bufferProto buffer containing the content of the .prototxt file
+      * @param lenProto length of bufferProto
+      * @param bufferModel buffer containing the content of the .caffemodel file
+      * @param lenModel length of bufferModel
+      */
+    CV_EXPORTS_W Net readNetFromCaffe(const char *bufferProto, size_t lenProto,
```
dkurt: I had a problem with using it in python because of null characters in serialized protocol buffers. I think we should use just `CV_EXPORTS`.

r2d3: If we use only CV_EXPORTS, the function will not be exposed to Python/Java interface.

I should look at how the binding generator works so the function accept Python str instead of string

alalek: It can't be exposed, because bindings don't work properly with native C++ pointers (at least they don't support buffers handling).

r2d3: In `modules/python/src2/hdr_parser.py`, I find a `CV_CARRAY` that seems to be able to wrap arrays. But the corresponding `#define` has disappeared from OpenCV headers.

modules/dnn/test/test_caffe_importer.cpp
```diff
@@ -55,6 +55,43 @@ static std::string _tf(TString filename)
     return (getOpenCVExtraDir() + "/dnn/") + filename;
 }
 
+static bool readFileInMemory(const string& filename, string& content)
```
dkurt: Do you mind if we replace it to `test_common.hpp` to reduce duplicated code? (see `test_tf_importer.cpp`)

r2d3: No problem. I was looking for this kind of common file and did not find it

modules/dnn/test/test_caffe_importer.cpp
```diff
@@ -55,6 +55,43 @@ static std::string _tf(TString filename)
...
+    content.reserve(ifs.tellg());
+    ifs.seekg(0, std::ios::beg);
+
+    content.assign((std::istreambuf_iterator<char>(ifs)),
+                    std::istreambuf_iterator<char>());
+
+    return true;
+}
+
+TEST(Test_Caffe, memory_read)
```
dkurt: I think it'd be better to test an accuracy of one of the models too.

r2d3: in the `memory_read` test case (as I have done for `Test_TensorFlow`) or in a separate test case ?

dkurt: In example, replace `TEST(Reproducibility_AlexNet, Accuracy)` to
```cpp
typedef TestWithParam<tuple<bool> > Reproducibility_AlexNet;
TEST_P(Reproducibility_AlexNet, Accuracy) {
  float readFromMemory = get<0>(GetParam());
  ...
}

INSTANTIATE_TEST_CASE_P(Test_Caffe, Reproducibility_AlexNet, Values(
  Bool()
));
```

r2d3: done

dkurt:
:+1:
@r2d3, thank you! Looks good for me.

PR #10001, opened 2017-11-02
cmake: exclude opencv_java (JNI) from OpenCV C++ modules list

alalek:
for non-ANDROID platforms

related #9950


mshabunin:
:+1:

PR #10003, opened 2017-11-02
Documentation updates

sturkmen72:
### This pullrequest changes

Some improvements on documentation and samples

### DIFF
--- a/modules/core/include/opencv2/core/saturate.hpp
+++ b/modules/core/include/opencv2/core/saturate.hpp
@@ -58,8 +58,8 @@ namespace cv
 
 /** @brief Template function for accurate conversion from one primitive type to another.
 
- The functions saturate_cast resemble the standard C++ cast operations, such as static_cast\<T\>()
- and others. They perform an efficient and accurate conversion from one primitive type to another
+ The function saturate_cast resembles the standard C++ cast operations, such as static_cast\<T\>()
+ and others. It perform an efficient and accurate conversion from one primitive type to another
  (see the introduction chapter). saturate in the name means that when the input value v is out of the
  range of the target type, the result is not formed just by taking low bits of the input, but instead
  the value is clipped. For example:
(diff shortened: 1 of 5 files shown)


modules/core/include/opencv2/core/saturate.hpp
```diff
@@ -58,7 +58,7 @@ namespace cv
 
 /** @brief Template function for accurate conversion from one primitive type to another.
 
- The functions saturate_cast resemble the standard C++ cast operations, such as static_cast\<T\>()
+ The function saturate_cast resemble the standard C++ cast operations, such as static_cast\<T\>()
```
terfendail: I can't say for sure whether it is better to use singular or plural here. But anyway it should be uniform with the latter text, so either there should be plural or `they` should be replaced with `it`in the latter text.

samples/cpp/falsecolor.cpp
```diff
@@ -10,24 +13,18 @@ struct ParamColorMar {
...
+    if (x == COLORMAP_PARULA + 1)
     {
-        if (!lutRND)
-        {
-            RNG ra;
-            lutRND = makePtr<Mat>(256, 1, CV_8UC3);
-            ra.fill(*lutRND, RNG::UNIFORM, 0, 256);
-        }
-        applyColorMap(p->img, dst, *lutRND.get());
+        randu(lutRND, Scalar(0, 0, 0), Scalar(255, 255, 255));
```
terfendail: That's the only place where `lutRND` Mat is used. I think `lutRND' declaration should be moved here as well. There is no reason to retain it global since it will anyway be repopulated here on each call.

terfendail:
Changes look fine for me 👍

alalek:
@sturkmen72 Please fix [whitespace](http://pullrequest.opencv.org/buildbot/builders/precommit_docs/builds/12985/steps/whitespace%20opencv/logs/stdio) issues and squash commits into one. Thank you!

PR #9984, opened 2017-11-01
Matrices in colorizeSegmentation were not initialized.

Smorodov:
Matrices in colorizeSegmentation of  fcn_semsegm.cpp were not initialized. It caused bad segmentation results.

### This pullrequest changes

It initializes matrices before use. In previous version uninitialized matrices gave random segmentation mask result.

PR #10008, opened 2017-11-02
Matrices in colorizeSegmentation were not initialized.

Smorodov:
### This pullrequest changes

It initializes matrices before use. In previous version uninitialized matrices gave random segmentation mask result.

#9984


terfendail:
👍

PR #10018, opened 2017-11-03
OCL: add cache for OpenCL compiled programs

alalek:
Default cache locations is under:
- Linux: `/home/<user>/.cache/opencv/` (in general)
- Windows: `C:\Users\<user>\AppData\Local\Temp\opencv\`
- MacOSX: `$TMPDIR/opencv/`
- Android: no defaults, not used, but can be enabled via environment configuration
- FreeBSD/other BSD: not tested
- other(iOS/EMSCRIPTEN/etc): not supported / disabled

Verified on these configuraions:
- with Intel iGPU (Linux/Windows/MacOSX)
- with AMD dGPU (Windows)

```
docker_image-Docs=docs-js
```

### DIFF
--- a/modules/core/include/opencv2/core/ocl.hpp
+++ b/modules/core/include/opencv2/core/ocl.hpp
@@ -261,6 +261,8 @@ public:
     void setUseSVM(bool enabled);
 
     struct Impl;
+    inline Impl* getImpl() const { return (Impl*)p; }
+//protected:
     Impl* p;
 };
 
--- a/modules/core/include/opencv2/core/utils/filesystem.hpp
+++ b/modules/core/include/opencv2/core/utils/filesystem.hpp
@@ -1,0 +1,38 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef OPENCV_UTILS_FILESYSTEM_HPP
+#define OPENCV_UTILS_FILESYSTEM_HPP
+
+namespace cv { namespace utils { namespace fs {
+
+
+CV_EXPORTS bool exists(const cv::String& path);
+CV_EXPORTS bool isDirectory(const cv::String& path);
+
+
+CV_EXPORTS cv::String getcwd();
+
+
+CV_EXPORTS bool createDirectory(const cv::String& path);
+CV_EXPORTS bool createDirectories(const cv::String& path);
+
+#ifdef __OPENCV_BUILD
+// TODO
+//CV_EXPORTS cv::String getTempDirectory();
+
+/**
+ * @brief Returns directory to store OpenCV cache files
+ * Create sub-directory in common OpenCV cache directory if it doesn't exist.
+ * @param sub_directory_name name of sub-directory. NULL or "" value asks to return root cache directory.
+ * @param configuration_name optional name of configuration parameter name which overrides default behavior.
+ * @return Path to cache directory. Returns empty string if cache directories support is not available. Returns "disabled" if cache disabled by user.
+ */
+CV_EXPORTS cv::String getCacheDirectory(const char* sub_directory_name, const char* configuration_name = NULL);
+
+#endif
+
+}}} // namespace
+
+#endif // OPENCV_UTILS_FILESYSTEM_HPP
--- a/modules/core/include/opencv2/core/utils/filesystem.private.hpp
+++ b/modules/core/include/opencv2/core/utils/filesystem.private.hpp
@@ -1,0 +1,64 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef OPENCV_UTILS_FILESYSTEM_PRIVATE_HPP
+#define OPENCV_UTILS_FILESYSTEM_PRIVATE_HPP
+
+// TODO Move to CMake?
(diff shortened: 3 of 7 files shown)


modules/core/include/opencv2/core/utils/filesystem.hpp
```diff
@@ -0,0 +1,38 @@
...
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef OPENCV_UTILS_FILESYSTEM_HPP
+#define OPENCV_UTILS_FILESYSTEM_HPP
+
+namespace cv { namespace utils { namespace fs {
+
+
+CV_EXPORTS bool isExist(const cv::String& path);
```
vpisarev: it's wrong grammar. I suggest `doesExist()`, `isValidPath()`, `pathExist()`, `isFileOrDirectory()` etc.

modules/core/include/opencv2/core/ocl.hpp
```diff
@@ -261,6 +261,8 @@ class CV_EXPORTS Context
     void setUseSVM(bool enabled);
 
     struct Impl;
+    inline Impl* getImpl() const { return (Impl*)p; }
+//protected:
```
vpisarev: why `Impl* p` is not protected anymore if we have getImpl()?

modules/core/include/opencv2/core/utils/filesystem.hpp
```diff
@@ -0,0 +1,38 @@
...
+//CV_EXPORTS cv::String getOpenCVTempDirectory();
+
+/**
+ * @brief Returns directory to store OpenCV cache files
+ * Create sub-directory in common OpenCV cache directory if it doesn't exist.
+ * @param sub_directory_name name of sub-directory. NULL or "" value asks to return root cache directory.
+ * @param configuration_name optional name of configuration parameter name which overrides default behavior.
+ * @return Path to cache directory. Returns empty string if cache directories support is not available. Returns "disabled" if cache disabled by user.
+ */
+CV_EXPORTS cv::String getOpenCVCacheDirectory(const char* sub_directory_name, const char* configuration_name = NULL);
```
vpisarev: should we use getOCLCacheDirectory or getOpenCLCacheDirectory name? I think, using `OpenCV` word in OpenCV function name is unnecessary

modules/core/include/opencv2/core/utils/filesystem.private.hpp
```diff
@@ -0,0 +1,64 @@
...
+ */
+class CV_EXPORTS FileLock {
+public:
+    explicit FileLock(const char* fname);
+    ~FileLock();
+
+    void lock(); //< acquire exclusive (writer) lock
+    void unlock(); //< release exclusive (writer) lock
+
+    void lock_sharable(); //< acquire sharable (reader) lock
```
vpisarev: We are trying to use mixed case names for methods everywhere in OpenCV. Can we do it here as well?

modules/core/include/opencv2/core/utils/lock.hpp
```diff
@@ -0,0 +1,119 @@
...
+ * @sa lock_guard
+ */
+template <class _Mutex>
+class optional_lock_guard {
+public:
+    typedef _Mutex Mutex;
+
+    explicit inline optional_lock_guard(Mutex* m) : mutex_(m) { if (mutex_) mutex_->lock(); }
+
+    inline ~optional_lock_guard() { if (mutex_) mutex_->unlock(); }
```
vpisarev: can we use mixed case name for the class?

modules/core/include/opencv2/core/utils/filesystem.hpp
alalek: Changed to `exists()`: http://en.cppreference.com/w/cpp/filesystem/exists

modules/core/include/opencv2/core/ocl.hpp
alalek: Because "p" is not protected initially. It is "TODO" mark for future releases.
Accurate testing of libva / directx / opengl integrations is required before apply this.

modules/core/include/opencv2/core/utils/filesystem.hpp
alalek: This function is designed not for OpenCL programs cache only, it can be reused in multiple places (ocl4dnn tuning results cache, probably DL MO cache, etc).
Thanks, renamed to `getCacheDirectory` (without OpenCV, because we already in `cv` namespace).

modules/core/include/opencv2/core/utils/filesystem.private.hpp
alalek: This is not OpenCV specific name. Also this is not a part of external public API (not available on some platforms).

Renamed to `lock_shared()` (typo is here) which is used in:
- [C++14](http://en.cppreference.com/w/cpp/thread/shared_lock/lock)
- [Boost](http://www.boost.org/doc/libs/1_51_0/doc/html/thread/synchronization.html#thread.synchronization.mutex_concepts.shared_lockable.lock_shared)

modules/core/include/opencv2/core/utils/lock.hpp
alalek: This is not OpenCV specific name and not a part of external public API.

Name is selected to be consistent with declarations above: `lock_guard`, `shared_lock_guard` (both have alternative implementation in modern C++ / boost library under the same names).

File is renamed to "lock.private.hpp".

vpisarev:
:+1:

PR #10024, opened 2017-11-04
Clearing data into the detect method of the class cv::LineSegmentDete…

### LINKED ISSUE #10023: LSD SIGABRT when you process multiple images
- OpenCV => 3.3.1-74-gbccbec7
- Operating System / Platform => Ubuntu 16.04 LTS (x_86_64)
- Compiler => GCC c++  (ver 5.4.0)

##### Detailed description

The LSD algorithm is a powerful method to detect line segments, in OpenCV it can be found in the file https://github.com/opencv/opencv/blob/7267e94be78b3c8be4f1f08b912ab57a52cf0954/modules/imgproc/include/opencv2/imgproc.hpp#L1202

With the implementation in the file 
https://github.com/opencv/opencv/blob/7267e94be78b3c8be4f1f08b912ab57a52cf0954/modules/imgproc/src/lsd.cpp#L163

This algorithm uses to work fine, but I've found a problem when you process several images of different sizes with the same `LineSegmentDetectorImpl` object. This problem comes from the usage of the field list:
https://github.com/opencv/opencv/blob/7267e94be78b3c8be4f1f08b912ab57a52cf0954/modules/imgproc/src/lsd.cpp#L270

That is a vector of coordinate points that are pseudo ordered by magnitude. During a normal execution you call to [LineSegmentDetectorImpl::detect](https://github.com/opencv/opencv/blob/7267e94be78b3c8be4f1f08b912ab57a52cf0954/modules/imgproc/src/lsd.cpp#L410), after that the method call to [LineSegmentDetectorImpl::flsd](https://github.com/opencv/opencv/blob/7267e94be78b3c8be4f1f08b912ab57a52cf0954/modules/imgproc/src/lsd.cpp#L435) and after that there is a call to [LineSegmentDetectorImpl::ll_angle](https://github.com/opencv/opencv/blob/7267e94be78b3c8be4f1f08b912ab57a52cf0954/modules/imgproc/src/lsd.cpp#L522). In this last function is the first usage of `list`:

https://github.com/opencv/opencv/blob/7267e94be78b3c8be4f1f08b912ab57a52cf0954/modules/imgproc/src/lsd.cpp#L568

This first usage resize the vector, supposing that it's empty and all this points after resize are (0, 0). But when you've processed previous frames, the list field can contain other points and this points are going to be used later, causing an undefined behaviour in the algorithm.

##### Steps to reproduce
The following code makes the LSD crash:

```C++
#include <iostream>

#include <opencv2/opencv.hpp>

using namespace cv;
using namespace std;

#define FIRST_IMG "../resources/020.jpg"
#define SECONG_IMG "../resources/074.jpg"

int
main(int argc, char *argv[])
{
  // Create the detector
  Ptr<LineSegmentDetector> detector = createLineSegmentDetector(LSD_REFINE_NONE);
  
  vector<Vec4f> segments;
  cv::Mat image, results;
  
  // Detect segments in the first image (1152x768)
  image = imread(FIRST_IMG, IMREAD_GRAYSCALE);
  detector->detect(image, segments);
  
  // Show the results
  cv::cvtColor(image, results, CV_GRAY2BGR);
  detector->drawSegments(results, segments);
  imshow("LSD lines", results);
  cv::waitKey();
  segments.clear();
  
  // Detect segments in the second image (1024x768)
  image = imread(SECONG_IMG, IMREAD_GRAYSCALE);
  detector->detect(image, segments);
  
  //Show the results
  cv::cvtColor(image, results, CV_GRAY2BGR);
  detector->drawSegments(results, segments);
  imshow("LSD lines", results);
  cv::waitKey();
  segments.clear();
  
}
```
I also attach a .zip with image examples:
[demo-bug-lsd.zip](https://github.com/opencv/opencv/files/1443241/demo-bug-lsd.zip)

iago-suarez:
close #10023 

### Clearing data into the detect method of the class cv::LineSegmentDetectorImpl

This pull request adds a last line into the method:
https://github.com/opencv/opencv/blob/7267e94be78b3c8be4f1f08b912ab57a52cf0954/modules/imgproc/src/lsd.cpp#L410
that clears the vector `list` avoiding the crash detailed in the issue #10023 .

### DIFF
--- a/modules/imgproc/src/lsd.cpp
+++ b/modules/imgproc/src/lsd.cpp
@@ -430,6 +430,9 @@ void LineSegmentDetectorImpl::detect(InputArray _image, OutputArray _lines,
     if(w_needed) Mat(w).copyTo(_width);
     if(p_needed) Mat(p).copyTo(_prec);
     if(n_needed) Mat(n).copyTo(_nfa);
+
+    // Clear used structures
+    list.clear();
 }
 
 void LineSegmentDetectorImpl::flsd(std::vector<Vec4f>& lines,


modules/imgproc/src/lsd.cpp
```diff
@@ -430,6 +430,9 @@ void LineSegmentDetectorImpl::detect(InputArray _image, OutputArray _lines,
     if(w_needed) Mat(w).copyTo(_width);
     if(p_needed) Mat(p).copyTo(_prec);
     if(n_needed) Mat(n).copyTo(_nfa);
+
+    // Clear used structures
+    list.clear();
```
terfendail: It looks like `list` vector isn't used in any meaningful way. IMHO it could be completely removed.

alalek: What do you mean?
It is used in other methods:
- ll_angle()
- flsd()

terfendail: `ll_angle` fills it as a row/column ordered list of all image points(actually almost all) and link it in special order, however the list head and tail pointers are local variables and are discarded as soon as list is constructed and the link order isn't used anywhere also.
`flsd` function use the list as an array to iterate over all stored points in a straight way and doesn't consider list ordering. So actually it just iterates over all image points.

vpisarev:
ok, it would not hurt to merge it in :+1:

PR #10038, opened 2017-11-07
(2.4) git: .gitignore update

alalek:
ignore all "dot" files/directories by default.
'ignored' files can be added via 'git add -f' command if necessary.

Backport #10037


alalek:
:+1:

PR #10039, opened 2017-11-07
java/CMakeLists: Fix build error with OpenEmbedded

ribalda:
Latest OE version fails to build the java module with a bunch of this error:

| CMake Error: File /home/ricardo/curro/qt5022/build-qt5022-open/build/tmp/work/bobcat_64-poky-linux/opencv/3.3+gitAUTOINC+87c27a074d_2a9d1b22ed_a62e20676a_34e4206aef_fccf7cd6a4-r0/build/modules/java/gen/core+Core.java does not exist.
| CMake Error at modules/java/CMakeLists.txt:295 (configure_file):
|   configure_file Problem configuring file

This patch makes sure that the .java file exists before copying it
(otherwise it is generated).

Signed-off-by: Ricardo Ribalda Delgado <ricardo.ribalda@gmail.com>

### DIFF
--- a/modules/java/CMakeLists.txt
+++ b/modules/java/CMakeLists.txt
@@ -291,7 +291,7 @@ foreach(java_file ${step3_input_files})
   endif()
   if(__configure)
     configure_file("${java_file}" "${java_src_dir}/${output_name}" @ONLY)
-  elseif(NOT "${java_file}" MATCHES "${OpenCV_BINARY_DIR}/")
+  elseif(EXISTS "${java_file}" AND NOT "${java_file}" MATCHES "${OpenCV_BINARY_DIR}/")
     configure_file("${java_file}" "${java_src_dir}/${output_name}" COPYONLY)
   else()
     add_custom_command(OUTPUT "${java_src_dir}/${output_name}"


alalek:
> 3.3+gitAUTOINC+87c27a074d_2a9d1b22ed_a62e20676a_34e4206aef_fccf7cd6a4-r0

Consider to remove special symbols from the path instead: "+" in this case.
OpenCV's CMake scripts doesn't support regex symbols in source/build(binary) paths.

ribalda:
Hi @alalek 

Thanks for your flashing fast response. In this case the path follows Open Embedded's naming : package-Version+git+gitversion.

I would really like to continue with that naming. I think I am close to doing that :). It is mainly this change,and another one for protobuf and the code seems to build (running is a different issue). 

Will ping you when I have a running system .

Thanks again

alalek:
There is common pattern for sub-directory checks via "MATCHES" condition. This pattern is used in **several places**.
- special symbols from regex syntax may not work
- double plus ('++') just breaks CMake completely
- there is no way to [escape](https://stackoverflow.com/questions/13569660/cmake-escaping-symbols-inside-a-variable-in-regular-expressions) symbols from variable in CMake.
- CMake doesn't support "real" functions (with return values, only procedures exist in CMake)
- so there is no CMake's one line equivalent for using with `if()`.

Current patch is not valid anyway - MATCHES is **always** FALSE for your build directory.

BTW, You could turn off Java build via `-DBUILD_opencv_java=OFF` CMake option.

ribalda:
I will add the patch you OE and mark it as Upstream:inappropriate (like most of: http://git.openembedded.org/meta-openembedded/tree/meta-oe/recipes-support/opencv/opencv )

Thanks anyway!

PR #10050, opened 2017-11-08
Add Android Mat constructor with support for native buffer

floe:
As a prerequisite for properly supporting the Camera2 API,  this PR adds an Android Mat constructor with support for passing native buffers.

```
force_builders=Android pack
```


alalek:
What part of [Android API](https://developer.android.com/reference/android/hardware/camera2/package-summary.html) provides access to native buffers?
How to control lifetime of this native buffer?

floe:
Apologies in advance, I just realized this won't work as intended. Needs to pass a `jobject` instead of a `jlong` and then call `env->getDirectByteBuffer(..)`.

@alalek The Camera2 API (https://developer.android.com/reference/android/hardware/camera2/package-summary.html) allows you to select an `ImageReader` as a preview surface, and from there, you can get a _direct_ mapped `ByteBuffer`.

floe:
@alalek side question: I haven't yet managed to build the Android OpenCV package locally. Can I download the results from the buildbot somewhere?

floe:
Just for the record, I have a (mostly) finished implementation of a `JavaCamera2View`, but I'd like to get this changeset merged separately first, as it changes a core component.

floe:
Just for the record, I have a (mostly) finished implementation of a `JavaCamera2View`, but I'd like to get this changeset merged separately first, as it changes a core component.

alalek:
Could you please add some simple [Java test](https://github.com/opencv/opencv/tree/master/modules/core/misc/java/test) for new functionality?

Something like this:
```
1. Create ByteBuffer with direct memory access
2. Fill it with some values
3. Create `Mat m` (2D is preferable)
4. Call `norm(m, NORM_L2)` or `countNonZero(m)`
5. check result
6. Modify Mat contents (cv::add)
7. m.release()
8. Check ByteBuffer updated values (first / last is enough)
```

modules/java/generator/src/cpp/Mat.cpp
```diff
@@ -52,6 +52,31 @@ JNIEXPORT jlong JNICALL Java_org_opencv_core_Mat_n_1Mat__
...
 
 //
+//   Mat::Mat(int rows, int cols, int type, void* data)
+//
+
+JNIEXPORT jlong JNICALL Java_org_opencv_core_Mat_n_1Mat__IIIO
+  (JNIEnv* env, jclass, jint rows, jint cols, jint type, jobject data);
+
+JNIEXPORT jlong JNICALL Java_org_opencv_core_Mat_n_1Mat__IIIO
+  (JNIEnv* env, jclass, jint rows, jint cols, jint type, jobject data)
```
alalek: Proper signature generated by "javah" is:
```
/*
 * Class:     org_opencv_core_Mat
 * Method:    n_Mat
 * Signature: (IIILjava/nio/ByteBuffer;)J
 */
JNIEXPORT jlong JNICALL Java_org_opencv_core_Mat_n_1Mat__IIILjava_nio_ByteBuffer_2
  (JNIEnv *, jclass, jint, jint, jint, jobject);
```

floe:
Thanks, updated. I still can't build/test locally, since I have a newer version of the Android SDK installed that doesn't support the `android` command anymore - so I have to rely on the buildbot ATM.

alalek:
Check "Android pack" builder artifacts (step 14) from here: http://pullrequest.opencv.org/#/summary/
(file is too large (~300Mb) - please let know if there are some issues - there is [build](http://pullrequest.opencv.org/buildbot/builders/precommit_pack_android/builds/10263) for armv7/aarch64 arch only (32Mb)).

floe:
Seems to work so far, all tests have passed and I can build my `JavaCamera2Source` with the new pack. I haven't tested the source itself yet, but I would suggest to do that in a new PR.

alalek:
BTW, There is some information why memory pointer from `GetDirectBufferAddress()` can't be moved by JVM garbage collector for `ByteBuffer.allocateDirect()` objects: https://stackoverflow.com/a/5671880

PR #10055, opened 2017-11-08
update digits_video.py

mabdullahrafique:
Following were the errors in the digits_video.py
     1 ) The code was not working because data type of x was float however in "cv2.rectangle" we require integer .
     2 ) After pressing the "esc" button the image windows did not destroy
So I amended following things:
     1 ) converted data type of x to int.
     2 ) used cv2.destroyAllWindows() to close all windows after the press of "esc" by user.

### DIFF
--- a/samples/python/digits_video.py
+++ b/samples/python/digits_video.py
@@ -55,7 +55,7 @@ def main():
             if not (16 <= h <= 64  and w <= 1.2*h):
                 continue
             pad = max(h-w, 0)
-            x, w = x-pad/2, w+pad
+            x, w = x - (pad // 2), w + pad
             cv2.rectangle(frame, (x, y), (x+w, y+h), (0, 255, 0))
 
             bin_roi = bin[y:,x:][:h,:w]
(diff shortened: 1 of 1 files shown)


samples/python/digits_video.py
```diff
@@ -56,6 +56,7 @@ def main():
                 continue
             pad = max(h-w, 0)
             x, w = x-pad/2, w+pad
+            x= int(x)
```
alalek: Please use integer division instead on the line above:
```.diff
-pad/2
+pad//2
```

samples/python/digits_video.py
```diff
@@ -98,3 +99,4 @@ def main():
 
 if __name__ == '__main__':
     main()
+cv2.destroyAllWindows()
```
alalek: This line should be places into `main()` function.

Looks like this observed effect is from IPython or similar. Normal standalone Python applications works well.

BTW, You could use `destroyWindow(name)` call instead:
https://stackoverflow.com/questions/15144442/when-is-an-opencv-imshow-window-destroyed-c

mabdullahrafique: Yes destroyAllWindows() can be called in main() function . Thanks

alalek:
I pushed update for this PR.

PR #10058, opened 2017-11-09
boxFilter and GaussianBlur HAL interface

ElenaGvozdeva:
### This pullrequest changes

Add HAL interface for boxFilter and GaussianBlur

Sanity data for boxFilter CV8U to CV16U test  https://github.com/opencv/opencv_extra/pull/405

`test_modules=imgproc`

### DIFF
--- a/modules/imgproc/perf/perf_blur.cpp
+++ b/modules/imgproc/perf/perf_blur.cpp
@@ -43,6 +43,9 @@ typedef perf::TestBaseWithParam<Size_MatType_BorderType3x3_t> Size_MatType_Borde
 typedef std::tr1::tuple<Size, MatType, BorderType> Size_MatType_BorderType_t;
 typedef perf::TestBaseWithParam<Size_MatType_BorderType_t> Size_MatType_BorderType;
 
+typedef std::tr1::tuple<Size, int, BorderType3x3> Size_ksize_BorderType_t;
+typedef perf::TestBaseWithParam<Size_ksize_BorderType_t> Size_ksize_BorderType;
+
 PERF_TEST_P(Size_MatType_BorderType3x3, gaussianBlur3x3,
             testing::Combine(
                 testing::Values(szODD, szQVGA, szVGA, sz720p),
@@ -130,6 +133,28 @@ PERF_TEST_P(Size_MatType_BorderType3x3, box3x3,
     declare.in(src, WARMUP_RNG).out(dst);
 
     TEST_CYCLE() boxFilter(src, dst, -1, Size(3,3), Point(-1,-1), false, btype);
+
+    SANITY_CHECK(dst, 1e-6, ERROR_RELATIVE);
+}
+
+PERF_TEST_P(Size_ksize_BorderType, box_CV8U_CV16U,
+            testing::Combine(
+                    testing::Values(szODD, szQVGA, szVGA, sz720p),
+                    testing::Values(3, 5, 15),
+                    BorderType3x3::all()
+                    )
+            )
+{
+    Size size = get<0>(GetParam());
+    int ksize = get<1>(GetParam());
+    BorderType3x3 btype = get<2>(GetParam());
+
+    Mat src(size, CV_8UC1);
+    Mat dst(size, CV_16UC1);
+
+    declare.in(src, WARMUP_RNG).out(dst);
+
+    TEST_CYCLE() boxFilter(src, dst, CV_16UC1, Size(ksize, ksize), Point(-1,-1), false, btype);
 
     SANITY_CHECK(dst, 1e-6, ERROR_RELATIVE);
 }
--- a/modules/imgproc/src/hal_replacement.hpp
+++ b/modules/imgproc/src/hal_replacement.hpp
@@ -670,12 +670,12 @@ inline int hal_ni_threshold(const uchar* src_data, size_t src_step, uchar* dst_d
    @param src_data,src_step Source image
    @param dst_data,dst_step Destination image
    @param width,height Source image dimensions
+   @param cn Number of channels
    @param margins Margins for source image
    @param ksize Size of kernel
    @param anchor Anchor point
    @param normalize If true then result is normalized
    @param border_type Border type
-   @param cn Number of channels
 */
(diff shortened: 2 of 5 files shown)


modules/imgproc/src/smooth.cpp
```diff
@@ -2108,6 +2114,22 @@ void cv::GaussianBlur( InputArray _src, OutputArray _dst, Size ksize,
                _src.rows() > ksize.height && _src.cols() > ksize.width);
     (void)useOpenCL;
 
+    Mat src = _src.getMat();
+    Mat dst = _dst.getMat();
```
alalek: These Mats should be released before `CV_OCL_RUN` and/or `sepFilter2D` call

modules/imgproc/src/hal_replacement.hpp
```diff
@@ -615,6 +615,43 @@ inline int hal_ni_integral(int depth, int sdepth, int sqdepth, const uchar * src
...
+   @param dst_data,dst_step Destination image
+   @param width,height Source image dimensions
+   @param cn Number of channels
+   @param margins Margins for source image
+   @param ksize Size of kernel
+   @param anchor Anchor point
+   @param normalize If true then result is normalized
+   @param border_type Border type
+*/
+inline int hal_ni_boxFilter(int src_depth, int dst_depth, const uchar* src_data, size_t src_step, uchar* dst_data, size_t dst_step, int width, int height, int cn, CvRect margins, CvSize ksize, CvPoint anchor, bool normalize, int border_type) { return CV_HAL_ERROR_NOT_IMPLEMENTED; }
```
mshabunin: We tried to avoid all complex data structs in HAL interface, please use only types defined int `hal/interface.h` headers.
Do we need all provided values to run the filter?

modules/imgproc/src/smooth.cpp
```diff
@@ -1552,19 +1552,25 @@ void cv::boxFilter( InputArray _src, OutputArray _dst, int ddepth,
...
     if(!(borderType&BORDER_ISOLATED))
         src.locateROI( wsz, ofs );
     borderType = (borderType&~BORDER_ISOLATED);
 
+    CvRect margin = cvRect(ofs.x, ofs.y, wsz.width - src.cols - ofs.x, wsz.height - src.rows - ofs.y);
+
+    CALL_HAL(boxFilter, cv_hal_boxFilter, sdepth, ddepth, src.ptr(), src.step, dst.ptr(), dst.step, src.cols, src.rows, cn,
+             margin, (CvSize)(ksize), (CvPoint)(anchor), normalize, borderType);
+
+#ifdef HAVE_TEGRA_OPTIMIZATION
```
mshabunin: This block can be removed, it is not compatible with OpenCV 3.x.

modules/imgproc/src/hal_replacement.hpp
ElenaGvozdeva: Yes, I need all values. Some of them needs only for to checking, I can check that anchor == (-1, -1) in OpenCV but what if another HAL could work with anchor?

mshabunin: It will be better to split each function into `Init`, `run` and `Free` stages, like it is done for _Filter2d_, _SepFilter2d_ and others.

ElenaGvozdeva:
@mshabunin what is the right order of accelerator calls?  `OCL`, `HAL`, `IPP`?

mshabunin:
Yes: `OCL, HAL, OVX, IPP, fallback`

modules/imgproc/perf/perf_blur.cpp
```diff
@@ -134,6 +137,28 @@ PERF_TEST_P(Size_MatType_BorderType3x3, box3x3,
...
+    BorderType3x3 btype = get<2>(GetParam());
+
+    Mat src(size, CV_8UC1);
+    Mat dst(size, CV_16UC1);
+
+    declare.in(src, WARMUP_RNG).out(dst);
+
+    TEST_CYCLE() boxFilter(src, dst, CV_16UC1, Size(ksize, ksize), Point(-1,-1), false, btype);
+
+    SANITY_CHECK(dst, 1e-6, ERROR_RELATIVE);
```
mshabunin: You have to generate sanity check data and put it to _opencv_extra_ if you want to use this function. I'd suggest using `SANITY_CHECK_NOTHING()` instead.

ElenaGvozdeva: I have already done it opencv/opencv_extra#405

alalek:
Looks like "OCL_RUN" should be moved back.

modules/imgproc/src/smooth.cpp
```diff
@@ -2093,29 +2096,43 @@ void cv::GaussianBlur( InputArray _src, OutputArray _dst, Size ksize,
...
+    Point ofs;
+    Size wsz(src.cols, src.rows);
+    if(!(borderType & BORDER_ISOLATED))
+        src.locateROI( wsz, ofs );
+
+    CALL_HAL(gaussianBlur, cv_hal_gaussianBlur, src.ptr(), src.step, dst.ptr(), dst.step, src.cols, src.rows, sdepth, cn,
+             ofs.x, ofs.y, wsz.width - src.cols - ofs.x, wsz.height - src.rows - ofs.y, ksize.width, ksize.height,
+             sigma1, sigma2, borderType&~BORDER_ISOLATED);
+
+    src.release();
```
mshabunin: I think you can remove these calls here and pass `src` and `dst` to all following calls.

modules/imgproc/src/smooth.cpp
```diff
@@ -2093,29 +2096,40 @@ void cv::GaussianBlur( InputArray _src, OutputArray _dst, Size ksize,
...
+        src.locateROI( wsz, ofs );
+
+    CALL_HAL(gaussianBlur, cv_hal_gaussianBlur, src.ptr(), src.step, dst.ptr(), dst.step, src.cols, src.rows, sdepth, cn,
+             ofs.x, ofs.y, wsz.width - src.cols - ofs.x, wsz.height - src.rows - ofs.y, ksize.width, ksize.height,
+             sigma1, sigma2, borderType&~BORDER_ISOLATED);
+
+    CV_OVX_RUN(true,
+               openvx_gaussianBlur(_src, _dst, ksize, sigma1, sigma2, borderType))
+
+    CV_IPP_RUN_FAST(ipp_GaussianBlur(_src, _dst, ksize, sigma1, sigma2, borderType));
```
mshabunin: Please, replace `_src, _dst` to `src, dst` to avoid unnecessary image copying in underlying functions.

PR #10061, opened 2017-11-09
DNN Torch - workaround when torch importer is disabled

Sahloul:
### This pullrequest changes
`readNetFromTorch()` utilizes `TorchImporter` which might not be defined if `ENABLE_TORCH_IMPORTER` is not set


alalek:
Thank you! :+1:

PR #10066, opened 2017-11-10
Opened up for core/traits/DataDepth class, making members public.

allannk:
### This pullrequest changes
This pullrequest makes members of core/traits/TypeDepth public and usable. 
This is very useful for templated serialization of vectors/matrices, and as far as I can tell the only way to convert OpenCV types (`CV_32F`) into system types (`float`)

Beware: Typo in title, involved class is TypeDepth, not DataDepth! Sorry for the confusion

### DIFF
--- a/modules/core/include/opencv2/core/traits.hpp
+++ b/modules/core/include/opencv2/core/traits.hpp
@@ -282,6 +282,7 @@ public:
 
 template<int _depth> class TypeDepth
 {
+public:
 #ifdef OPENCV_TRAITS_ENABLE_LEGACY_DEFAULTS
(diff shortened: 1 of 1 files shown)


alalek:
I would reject this patch because it touches unused, deprecated (and even broken) structure.

allannk:
That is fair enough. Could you suggest an alternative method to get the native datatype of an OpenCV type? I.e. a way to convert `CV_64F` to `double`

alalek:
There is no way to do this.
Use types as first class citizens in "upstream" templates instead of some typeid numbers. This is more consistent practice.

PR #10068, opened 2017-11-10
FileStorage: use copydoc to add base64 info to constructor

paroj:


mshabunin:
:+1:

alalek:
BTW, bindings generators try to re-use these doc strings somehow.

PR #10070, opened 2017-11-10
Fix #7084 Thin Plate Spline warpImage() error

### LINKED ISSUE #7084: Bug in ThinPlateSplineShapeTransformer::wrapImage

- OpenCV => 3.1

##### Detailed description
Hi,

I'm working with the ThinPlateSplineShapeTransformer. I tried the methods **estimateTransformation**, **applyTransformation** and **warpImage**. It seems there is a bug in the ThinPlateSplineShapeTransformer::warpImage implementation where transformed X values are located in the oposite side. The same behavior for the Y values. 

##### Steps to reproduce

First, I tried only the **estimateTransformation** and **applyTransformation** methods.

```cpp
    auto tps = cv::createThinPlateSplineShapeTransformer();
    std::vector<cv::Point2f> sourcePoints, targetPoints;
    sourcePoints.push_back(cv::Point2f(0, 0));
    targetPoints.push_back(cv::Point2f(100, 0));
    sourcePoints.push_back(cv::Point2f(799, 0));
    targetPoints.push_back(cv::Point2f(799, 0));
    sourcePoints.push_back(cv::Point2f(0, 599));
    targetPoints.push_back(cv::Point2f(0, 599));
    sourcePoints.push_back(cv::Point2f(799, 599));
    targetPoints.push_back(cv::Point2f(799, 599));

    std::vector<cv::DMatch> matches;
        for (unsigned int i = 0; i < sourcePoints.size(); i++)
            matches.push_back(cv::DMatch(i, i, 0));

    tps->estimateTransformation(sourcePoints, targetPoints, matches);
    std::vector<cv::Point2f> transPoints;
    tps->applyTransformation(sourcePoints, transPoints);

    cout << "sourcePoints = " << endl << " " << sourcePoints << endl << endl;
    cout << "targetPoints = " << endl << " " << targetPoints << endl << endl;
    cout << "transPos = " << endl << " " << transPoints << endl << endl;
```
So, we want to transform 4 points (sourcePoints) into 4 targetPoints (targetPoints). In line 4 the source point (0,0) should be transformed to the target point (100, 0). In other words: The target point is 100 units moved to the right. We assume that we have a x axis which will be positive to the right. 
Then we match the 4 source to the target points with cv::DMatch. And then we call **estimateTransformation** to perform the thin plate spline implementation. And finally we prove the tranformation with the method **applyTransformation**. The transformed points "transPoints" should be equal to the targetPoints.
And thats true. The output is:
```
sourcePoints =
 [0, 0;
 799, 0;
 0, 599;
 799, 599]

targetPoints =
 [100, 0;
 799, 0;
 0, 599;
 799, 599]

transPos =
 [100.00003, 0;
 799, 0;
 4.196167e-005, 599;
 799, 599]
```
Hurrah. The **estimateTransformation** and **applyTransformation** methods work as expected.

But now we want to apply the transformation for a "real" image with **warpImage**. Be aware that the origin of an "image" is in the upper left corner. So the y axis goes from up to down and the x axis goes from left to right.

I added only these few lines of code to the above code

```cpp
    cv::Mat img = cv::imread("C:/tests/ThinPlateSplineShapeTransformer/t.png");
    cv::imshow("input", img);
    cv::Mat res;
    tps->warpImage(img, res);
    cv::imshow("warp result", res);
```

Here is the input image:
t
But the wrapped image is not correct. It seems that the point (0,0) is transformed to (-100, 0) or somewhere else.
wrapresult

A work-around is to switch sourcePoints and targetPoints in **estimateTransformation**
```cpp
// **** Switch source and target points so that wrapImage is working correctly. ****
tps->estimateTransformation(targetPoints, sourcePoints, matches); 
```
After the "switch" the wrap result is correct:
wrapresult
In this image you see that the point (0,0) is transformed to (100,0).

I didn't show the images for the Y transformation. But there is the same problem like the X transformation.

The work-around is very dangerous. If someone else is reading my code the other programmer could think that it is wrong and switch the source and target points back.

Greetings,
Thomas

lzhbrian:
### Resolves issue with warpImage() in Thin Plate Spline mentioned in #7084
* As suggested in #7084 by @WIStudent, the `ThinPlateSplineShapeTransformer` is working fine with `estimateTransformation()` and `applyTransformation()`. 
But `warpImage()` has got wrong with how `remap()` works.

* `remap()` performs:  ``` dst(x, y) = src(map_x(x, y), map_y(x, y)) ```
But what we really want is:  __``` dst(map_x(x, y), map_y(x, y)) = src(x, y) ```__

### Some math in the code
* Given that TPS is actually learning a translation 
`(x, y) -> (map_x(x, y), map_y(x, y)) = (x + \delta x, y + \delta y)`
* let `x1 = map_x(x, y)`, `y1 = map_y(x, y)`
* Then:`\delta x = x1 - x`, `\delta y = y1 - y`, `x = x1 - \delta x`, `y = y1 - \delta y`
* So: `dst(x1, y1) = dst(map_x(x, y), map_y(x, y)) = src(x, y) = src(x1 - \delta x, y1 - \delta y)`
* See x1, y1 as new x, y: `dst(x, y) = src(x - \delta x, y - \delta y)`
which is going to be:
```cpp
// deltax = pt.x - col;
// deltay = pt.y - row;
mapX.at<float>(row, col) = col - (pt.x - col);
mapY.at<float>(row, col) = row - (pt.y - row);
```

### P.S. Why don't we directly using something like the following?
```cpp
mapX.at<float>(pt.y, pt.x) = col;
mapY.at<float>(pt.y, pt.x) = row;
```
The above code is definitely fine to represent `dst(map_x(x, y), map_y(x, y)) = src(x, y)`
But using it would cause so many pixels with unknown value.

### DIFF
--- a/modules/shape/src/tps_trans.cpp
+++ b/modules/shape/src/tps_trans.cpp
@@ -161,8 +161,9 @@ void ThinPlateSplineShapeTransformerImpl::warpImage(InputArray transformingImage
         for (int col = 0; col < theinput.cols; col++)
         {
             Point2f pt = _applyTransformation(shapeReference, Point2f(float(col), float(row)), tpsParameters);
-            mapX.at<float>(row, col) = pt.x;
-            mapY.at<float>(row, col) = pt.y;
+
+            mapX.at<float>(row, col) = col - (pt.x - col);
+            mapY.at<float>(row, col) = row - (pt.y - row);
         }
     }
     remap(transformingImage, output, mapX, mapY, flags, borderMode, borderValue);


WIStudent:
`remap()` performs `dst(x,y) = src(map_x(x,y), map_y(x,y))`, not  `dst(x, y) = src(map(x), map(y))`.

TPS doesn't lern a transformation `(x, y) -> (map(x), map(y))` but `(x,y) -> map(x,y) = (x1, y1)`. So it must be `x1 = map_x(x,y)` and `y1 = map_y(x,y)`.

Also `dst(x,y) = src(map_x(x,y), map_y(x,y))` is exactly what we want. For each pixel of the output image we want the position in the input image to sample its value at that coordinate. If we would calculate the coordinate in the output image for each pixel of the input image and then "move" the input pixel to that coordinate, our output image would have pixels with unknown value.

lzhbrian:
Thanks @WIStudent, map_x(x,y), map_y(x,y) are actually what I really meant, I have edited them.

For the __unknown value issue__, If you directly give `dst(map_x(x, y), map_y(x, y)) = src(x, y)`, which is
```
mapX.at<float>(pt.y, pt.x) = col;
mapY.at<float>(pt.y, pt.x) = row;
```
then it would surely have many pixels with unknown value.

But as I have written in the above comment, I did some trick to make sure every pixel in the `dst` got a place to sample from the `src`.

I think the main issue here is __what `warpImage()` expects__, as @alalek said [here](https://github.com/opencv/opencv/issues/7084#issuecomment-343693842) in #7084. Though I do not see in what circumstances will we need a function to warp an image from a transformed one to source ...

WIStudent:
Let's ignore for a moment which image is the source image and which is the target image because that can be changed easily by switching parameters.

Instead let's first take a look at how to transform an image `I` using a function `f: R^2 -> R^2` (it maps a point in a 2D space to a point in another 2D space). There are two options.

### Forward Transformation
Let `(x_I, y_I)` be a point in image `I` and `(x_T, y_T)` a point in the transformed image `T`. For each pixel in `I` the coordinate in `T` is calculated by `(x_T,y_T) = f_forward(x_I, y_I)`. Then the pixel is "moved" to the coordinate  `(x_T,y_T)`. This has the flaw that there will be pixels in `T` where no value was assigned.
forward transformation
(Replace `I_O` with `I` and `I_T` with `T` in that image. The red dots are pixels where the values are unknown.)

### Backward Transformation
For each pixel in `T` the coordinate in `I` is calculated by `(x_I, y_I) = f_backward(x_T, y_T)`. To get the value for pixel `(x_T, y_T)` the value at `f_backward(x_T, y_T)` is sampled.
backward transformation
(again, replace `I_O` with `I` and `I_T` with `T`)
Background transformation has the advantage that no "holes" appear in the transformed image `T`. This is exactly what `warpImage()` does.


Now no matter which option we choose, we still need to calculate that function `f: R^2 -> R^2`, which will be estimated by a Thin-Plate-Spline function. If we want to use the backward transformation/`warpImage()`, we need to calculate the function `f_tps(x_T,y_T) = (x_I,y_I)`.

If `warpImage()` doesn't transform you image correctly, you most likely switched the parameters when calculating the TPS transformation and calculated `f_tps(x_I, y_I) = (x_T, y_T)` instead.

modules/shape/src/tps_trans.cpp
```diff
@@ -161,8 +161,9 @@ void ThinPlateSplineShapeTransformerImpl::warpImage(InputArray transformingImage
         for (int col = 0; col < theinput.cols; col++)
         {
             Point2f pt = _applyTransformation(shapeReference, Point2f(float(col), float(row)), tpsParameters);
-            mapX.at<float>(row, col) = pt.x;
-            mapY.at<float>(row, col) = pt.y;
+
+            mapX.at<float>(row, col) = col - (pt.x - col);
```
terfendail: This estimation is wrong. According to math evaluations you provide in description here should be
`mapX(x1, y1) = x1 - deltaX` where `deltaX = x1 - x`. You could consider `x1` as new `x`, but in this case you should somehow rename old `x` to retain it in the formula. Therefore if you substitute `deltaX` it would be `mapX(x1, y1) = x1 - (x1 -x)`. So after renaming you get `mapX(x, y) = x - (x - oldX)` instead of `mapX(x, y) = x - (x1 - x)`
For example the following set of points will be evaluated by the current code incorrectly:
src(0,0)->dst(1,0)
src(1,0)->dst(1,1)
src(1,1)->dst(0,1)
src(0,1)->dst(0,0)
In accordance with the formula dst(0,0) will be back projected to src(-pt.x, -pt.y)=src(-1, 0)(in accordance with `_applyTransformation` behaviour) but it should be src(0,1)

vpisarev:
ok, let's close it for now

Issue #8923, opened 2017-06-15
CMakeCompiler Options Silently change -O3 to -O2


adishavit:
On certain platforms, `OpenCVCompilerOptions.cmake` [changes optimization flag `-O3` to `-O2`](https://github.com/opencv/opencv/blame/5b63399197fdb408dc53bcdf29dbc7b1e0782258/cmake/OpenCVCompilerOptions.cmake#L31).  

This might have been important 5 years ago. s is still required?
It was originally supposed to be just for 32-bit MINGW, but that check is gone.

Specifically, it prevents -O3 when building OpenCV with emscripten, which may significantly reduce the optimizations. It's possible to just check for `NOT EMSCRIPTEN` but it may be worth reconsidering this whole silent change as it may not be relevant anymore.

adishavit:
This is especially strange since [`OpenCVDetectCXXCompiler.cmake`](https://github.com/opencv/opencv/blob/master/cmake/OpenCVDetectCXXCompiler.cmake) has the following lines:  
```
if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
  set(CMAKE_COMPILER_IS_GNUCXX 1)
  set(CMAKE_COMPILER_IS_CLANGCXX 1)
endif()
if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
  set(CMAKE_COMPILER_IS_GNUCC 1)
  set(CMAKE_COMPILER_IS_CLANGCC 1)
endif()
```
So if it's Clang it sets it also to be gcc!

adishavit:
cc: @taka-no-me

adishavit:
Apparently resolved by #10073.

PR #10073, opened 2017-11-12
cmake: drop amending of compilation options

alalek:
Allow to specify optimization level via CMake.

related #8923


mshabunin:
:+1:

PR #10076, opened 2017-11-13
ocl kernel run performance improvement

pengli:
don't set cl event callback after kernel enqueue

This is not efficient if there are many ocl kernels to be enqueued,
and will impact application performance for example the dnn module.
So the patch just cleanup umats and images after kernel is enqueued.

### DIFF
--- a/modules/core/src/ocl.cpp
+++ b/modules/core/src/ocl.cpp
@@ -2121,6 +2121,7 @@ struct Kernel::Impl
             }
         nu = 0;
         haveTempDstUMats = false;
+        haveTempSrcUMats = false;
     }
 
     void addUMat(const UMat& m, bool dst)
@@ -2131,6 +2132,8 @@ struct Kernel::Impl
         nu++;
         if(dst && m.u->tempUMat())
             haveTempDstUMats = true;
+        if(!dst && m.u->tempUMat() && m.u->urefcount == 2)
+            haveTempSrcUMats = true;
     }
 
     void addImage(const Image2D& image)
@@ -2168,6 +2171,7 @@ struct Kernel::Impl
     int nu;
     std::list<Image2D> images;
     bool haveTempDstUMats;
+    bool haveTempSrcUMats;
 };
 
 }} // namespace cv::ocl
@@ -2474,9 +2478,17 @@ bool Kernel::Impl::run(int dims, size_t globalsize[], size_t localsize[],
     }
     else
     {
-        addref();
-        isInProgress = true;
-        CV_OCL_CHECK(clSetEventCallback(asyncEvent, CL_COMPLETE, oclCleanupCallback, this));
+        if (haveTempSrcUMats)
+        {
+            addref();
+            isInProgress = true;
+            CV_OCL_CHECK(clSetEventCallback(asyncEvent, CL_COMPLETE, oclCleanupCallback, this));
+        }
+        else
+        {
+            cleanupUMats();
+            images.clear();
+        }
     }
     if (asyncEvent)
         CV_OCL_DBG_CHECK(clReleaseEvent(asyncEvent));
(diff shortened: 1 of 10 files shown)


alalek:
UMat can be allocated over user data.
1. Sometimes OpenCV can't use `CL_MEM_USE_HOST_PTR` feature to do this - it depends on platform restrictions (non-unified memory between, unsupported alignment, etc) or just not supported in OpenCL runtime at all. In this case OpenCV simulate this behaviour: on allocation - makes "copy" of user data (via `CL_MEM_COPY_HOST_PTR` or via separate "copy" process) and after deallocation synchronize data back. So deallocation process should be done very accurate - at least on the same OpenCL command queue (we don't use out-of-order queues yet).

2. OpenCV uses buffer pool to reduce number of allocations/deallocations. So "freed" UMat can be reused again while its kernel still processing.

pengli:
Is there a proper way to handle this situation  ? we observed "clSetEventCallback" cost some time in executing dnn net model like googlenet and we want to minimize this overhead.

alalek:
Looks like, there is some handling via `haveTempDstUMats` flag (but this force full synchronization).
Perhaps we need something like this for source UMats.

pengli:
@alalek , I added "haveTempSrcUMats" flag to check if src UMats requires async deallocation.  If yes,  cl event callback will be set, otherwise just call cleanupUMats directly. please check the new patch.

alalek:
I'm observing 40% performance degradation on this test case (48 -> 79ms): `ENet::DNNTestNetwork::(DNN_BACKEND_DEFAULT, DNN_TARGET_OPENCL)`.
Could you check this on your platforms?

pengli:
I run the test on  both Skylake (48eu) and Broadwell (24eu) ,  

1) on Skylake

without this PR:

[ PERFSTAT ]    (samples = 100, mean = 63.58, median = 63.25, stddev = 2.48 (3.9%))
[       OK ] DNNTestNetwork.ENet/1 (20120 ms)
[----------] 1 test from DNNTestNetwork (20120 ms total)

with this PR:

[ PERFSTAT ]    (samples = 100, mean = 71.73, median = 68.78, stddev = 7.82 (10.9%))
[       OK ] DNNTestNetwork.ENet/1 (19297 ms)
[----------] 1 test from DNNTestNetwork (19297 ms total)

2)  on Broadwell

without this PR:

[ PERFSTAT ]    (samples = 10, mean = 123.05, median = 121.77, stddev = 3.60 (2.9%))
[       OK ] DNNTestNetwork.ENet/1 (60136 ms)
[----------] 1 test from DNNTestNetwork (60136 ms total)

with this PR:

[ PERFSTAT ]    (samples = 38, mean = 116.94, median = 115.76, stddev = 3.29 (2.8%))
[       OK ] DNNTestNetwork.ENet/1 (57252 ms)
[----------] 1 test from DNNTestNetwork (57252 ms total)

So on skylake, the performance is down about 12%, on broadwell, the performance is improved.  one thing is that the torch enet model isn't fully enabled on GPU, there are many layers fallback called.  So I think it is better to check the OCL backend performance again when torch enet is fully enabled on GPU.

pengli:
I also test googlenet on skylake which is now fully enabled on OCL path 

without this PR

[ PERFSTAT ]    (samples = 10, mean = 24.93, median = 24.78, stddev = 0.43 (1.7%))
[       OK ] DNNTestNetwork.GoogLeNet/1 (12785 ms)
[----------] 1 test from DNNTestNetwork (12785 ms total)

with this PR

[ PERFSTAT ]    (samples = 10, mean = 20.72, median = 20.62, stddev = 0.46 (2.2%))
[       OK ] DNNTestNetwork.GoogLeNet/1 (12531 ms)
[----------] 1 test from DNNTestNetwork (12531 ms total)

This PR clearly improve the performance.

alalek:
Could you try this patch on your devices: https://github.com/alalek/opencv/commit/pr10076_r ?

PR #10078, opened 2017-11-13
core:ppc Fix 2 interleave logic errors in vsx_utils.hpp

justdoitqd:
When elements are 64 bits, the vec_st_interleave()/vec_ld_deinterleave()
doesn't interleave 4 elements correctly.

For vec_st_interleave(), following is saved into mem:
	a0 b0 a1 b1 c0 d0 c1 d1
     -> we expected:
	a0 b0 c0 d0 a1 b1 c1 d1

for vec_ld_deinterleave(), following is loaded into a b c d for memory
string { 1 2 3 4 5 6 7 8 }:
	a: 1 3
	b: 2 4
	c: 5 7
	d: 6 8
   -> we expected:
   	a: 1 5
	b: 2 6
	c: 3 7
	d: 4 8

This patch corrects this behavior.

Signed-off-by: Simon Guo <wei.guo.simon@gmail.com>

```
```

### DIFF
--- a/modules/core/include/opencv2/core/vsx_utils.hpp
+++ b/modules/core/include/opencv2/core/vsx_utils.hpp
@@ -797,8 +797,8 @@ FORCE_INLINE(void) vec_st_interleave(const Tvec& a, const Tvec& b,          \
                                      const Tvec& c, const Tvec& d, Tp* ptr) \
 {                                                                           \
     st_func(vec_mergeh(a, b), 0, ptr);                                      \
-    st_func(vec_mergel(a, b), 2, ptr);                                      \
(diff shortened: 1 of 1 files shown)


seiko2plus:
@justdoitqd What a mistake, I'm so embarrassed. can I combine your patch into #9907? since I already did many improvements on vsx_utils.hpp and also if I'm not mistaken this issue isn't going to be affected by the current implementation because universal intrinsics de&interleave supports all types except 64-bit so no need to immediate merge.

Thank you so much.

justdoitqd:
@seiko2plus 
Sure. Please go ahead to merge my patch.
thxs
- Simon

Issue #9269, opened 2017-07-30
opencv 3.2 Android: CameraBridgeViewBase.java camera preview does not respect orientation


stuaxo:
- OpenCV => 3.2
- Operating System / Platform => Ubuntu 17.04 64 bit
- Compiler => n/a (used sdk download)

This may be similar to #8866.

Symptons:   If the Android applications orientation is fixed to landscape then everything is displayed correctly, if not then the preview will be in the wrong orientation except when the phone is held in landscape.

There are various solutions in https://stackoverflow.com/questions/16669779/opencv-camera-orientation-issue with different levels of performance - some rotate the image after receiving it, others patch CameraBridgeViewBase.

The versions that patch CameraBridgeViewBase tend to keep track of the orientation and patch deliverAndDrawFrame.

I'm using Henry Doans answer  https://stackoverflow.com/questions/16669779/opencv-camera-orientation-issue/34541529#34541529  it seemed the most reasonable on a really quick look, but no idea if this is 100% efficient.

Phone seems to be a bit more laggy, but not as slow as some of the other solutions.

rajkumarmishra:
solution discussed here:
http://answers.opencv.org/question/20325/how-can-i-change-orientation-without-ruin-camera-settings/
https://stackoverflow.com/questions/16669779/opencv-camera-orientation-issue
has solved the problem of orientation, working fine for all the devices. but the fps is dropped.
can someone give solution to increase fps.

stuaxo:
@rajkumarmishra fixing this bug is the only way to get better fps, all the other solutions are workarounds that add extra work.

Issue #8866, opened 2017-06-07
Android: Camera2Renderer ignores display and sensor rotation


PhilLab:
The [Camera2Renderer](https://github.com/opencv/opencv/blob/master/modules/java/generator/src/java/android%2BCamera2Renderer.java) Android class does not rotate the preview image if the camera's sensor is rotated (this can happen with some smartglasses, e.g. with the Vuzix M300 where the sensor is mounted upside-down). The result is that the preview image is rotated by 180 degrees.

The [Camera2Sample](https://developer.android.com/samples/Camera2Basic/src/com.example.android.camera2basic/Camera2BasicFragment.html#l524) by Android uses

     characteristics.get(CameraCharacteristics.SENSOR_ORIENTATION);
to query the sensor rotation and

    activity.getWindowManager().getDefaultDisplay().getRotation();
to query the display rotation but I did not succeed in integrating this into OpenCV's Camera2Renderer.

PR #10081, opened 2017-11-13
[RFC] Java Camera2 View for Android

floe:
Based on #10050, here's a first alpha implementation of the Camera2 adapter class which should allow zero-copy access to preview image data. Comments very welcome.

Known issues at the moment:

- ~~tutorial2-mixedprocessing works with this class as a drop-in replacement for `JavaCameraView`, but quits after ~ 20-30 seconds (no crash, just exit due to internal camera queue error, perhaps some race condition?)~~ - fixed, was a memory leak
- lots of code overlap with `JavaCameraView` and `Camera2Renderer` (e.g. `JavaCameraFrame` internal class), needs cleanup/refactoring
- `cacPreviewSize(...)` currently buggy, selects smaller preview size than feasible

```
force_builders=Android pack
```


modules/java/generator/src/java/android+JavaCamera2View.java
```diff
@@ -0,0 +1,366 @@
...
+
+            mImageReader = ImageReader.newInstance(w,h,ImageFormat.YUV_420_888,2);
+            mImageReader.setOnImageAvailableListener(new ImageReader.OnImageAvailableListener() {
+                @Override
+                public void onImageAvailable(ImageReader reader) {
+                    Image image = reader.acquireLatestImage();
+                    if (image == null) return;
+                    ByteBuffer plane = image.getPlanes()[0].getBuffer();
+                    Mat yuv = new Mat( h * 3/2, w, CvType.CV_8UC1, plane );
+                    JavaCamera2Frame tempFrame = new JavaCamera2Frame(yuv,w,h);
```
alalek: [This](https://developer.android.com/reference/android/media/Image.html#getFormat()) says that all three planes should be used for `YUV_420_888`.

BTW, Please add additional sanity checks (assertions) - for actual `Image` format, `ByteBuffer` size, etc.

modules/java/generator/src/java/android+JavaCamera2View.java
```diff
@@ -0,0 +1,366 @@
...
+            }
+        } catch (InterruptedException e) {
+            throw new RuntimeException("Interrupted while trying to lock camera closing.", e);
+        } finally {
+            mCameraOpenCloseLock.release();
+            stopBackgroundThread();
+        }
+    }
+
+    boolean cacPreviewSize(final int width, final int height) {
```
alalek: There is typo (in original code too): `calcPreviewSize`

alalek:
Please update files filtering for old Android versions [here](https://github.com/opencv/opencv/blob/3.3.1/modules/java/CMakeLists.txt#L240)

modules/java/generator/src/java/android+JavaCamera2View.java
floe: Hmm... OTOH, you're correct (there are 3 planes), but obviously, the planes are contiguous in memory and also mapped in one single chunk, otherwise, the `cvtColor` call would cause a segfault. 

To implement this in a way that references all 3 planes separately would AFAICT require a new implementation of `cvtColor` that can handle 3 separate `Mat` objects. Otherwise, you'd need to copy the planes together into a single `Mat` and lose the advantage of the zero-copy native buffer.

modules/java/generator/src/java/android+JavaCamera2View.java
floe: Aaah. This is supposed to mean "calc". :man_facepalming: Will fix that.

modules/java/generator/src/java/android+JavaCamera2View.java
mshabunin: We have `hal::` conversion functions accepting two separate pointers for Y and UV planes ([example](https://github.com/opencv/opencv/blob/ba940dec38983cb35de6342d5e770eb0f054549e/modules/imgproc/include/opencv2/imgproc/hal/hal.hpp#L195-L199), so you can try to use them. You can add more overloaded functions if needed. But it is hard to say how it will work taking in account that each plane has its own `PixelStride` parameter which is not supported by current implementations.

floe: According to https://developer.android.com/reference/android/graphics/ImageFormat.html#YUV_420_888 the Y plane will always have a stride of 1, and the U/V planes will have equal stride. I may simply assert that U/V stride is also 1 for the moment (seems to be true for all implementations I've seen so far).

mshabunin: Are you sure U and V have PixelStride=1? Probably it is 2. If your current code has worked with YUV_BGRA_NV12 conversion by providing only _Y_ pointer, then planes should've been located in continuous memory block like this:
```
YYYY
YYYY
UVUV
```
Then: 
```
Uptr = Yptr + row_stride * rows
Upixel_stride = 2
Vptr = Uptr + 1
Vpixel_stride = 2
```

mshabunin: Corrections to my previous comment: the conversion you used is YUV2RGB2_NV21 and in memory layout the V channel stands before U.

floe: But I'm using COLOR_YUV2RGB_YV12 (see below), and that is for 3 sequential planes, isn't it?

mshabunin: As I can see from the code:
```.cpp
private int mPreviewFormat = ImageFormat.NV21;
```
This field is not modified anywhere and is used to determine conversion mode.

floe: My mistake, you're right. So it's one separate Y plane and one interleaved U/V plane.

floe: But in that case, `cvtTwoPlaneYUVtoBGR` should probably work?

alalek: Probably it should, but we should check these assumptions (via `CV_Assert`) before calling any processing code.

mshabunin: @floe , the problem with `cvtTwoPlaneYUVtoBGR` is that it does not get wrapped automatically, because it accepts raw C++ pointers. So you will need to write JNI wrapper by hand or wait_for/propose extended `cvtColor(InputArray ...)` interface.

floe: Is there any way to check whether the planes are contiguous in memory before passing them to `Mat()`? Then the current approach would already be sufficient...

floe:
I've updated this PR:

- fixed memory leak
- fixed some typos
- fixed incorrect format handling
- added lots of asserts

The only thing that remains open IMHO is the handling of the RGB conversion here: https://github.com/opencv/opencv/pull/10081/files#diff-980f99834f5223fab7fde83308fc9930R353 

AFAICT the cleanest way would be to store Y and U/V plane as two `Mat` objects in `JavaCamera2Frame` and add a wrapper to `ImgProc.cvtColor` that calls through to `cvtTwoPlaneYUVtoBGR`. Opinions?

floe:
Side question: where are the Java wrappers for `cvtColor` defined?

alalek:
Almost all wrapper for Java/Python bindings are generated automatically from public headers of modules (functions with `CV_EXPORTS_W`).

BTW, `cv::Mat` doesn't support multiple planes (you need 2 `cv::Mat` to store Y and UV planes).

floe:
Another side question: I added
```
force_builders=Android pack
```
to the initial PR comment, but that still doesn't seem to trigger the Android pack build?

floe:
*bump* sorry for the noise, but I do need the Android pack to work for testing...

floe:
Nevermind, seems to work now :+1:

floe:
I think this should now be ready for merging. I still can't get the "Android pack" build to trigger reliably, though. Any comments welcome /cc @alalek @mshabunin @vpisarev

floe:
Update: fixed one more bug in the C wrapper; seems to work quite smoothly now. I've tested on an Nexus 5 with Android 7.1.2, and got a reduction in CPU load of about 25% with tutorial-2-mixedprocessing.

floe:
One more question: I've called the C wrapper function `cvtColorTwoPlane`. To avoid namespace pollution, would it perhaps be better to just overload `cvtColor`?

modules/imgproc/src/color.cpp
```diff
@@ -11050,6 +11050,41 @@ inline bool isFullRange(int code)
...
+// helper function for dual-plane modes
+void cv::cvtColorTwoPlane( InputArray _ysrc, InputArray _uvsrc, OutputArray _dst, int code, int dcn )
+{
+    // only YUV420 is currently supported
+    switch (code) {
+        case COLOR_YUV2BGR_NV21:  case COLOR_YUV2RGB_NV21:  case COLOR_YUV2BGR_NV12:  case COLOR_YUV2RGB_NV12:
+        case COLOR_YUV2BGRA_NV21: case COLOR_YUV2RGBA_NV21: case COLOR_YUV2BGRA_NV12: case COLOR_YUV2RGBA_NV12:
+            break;
+        default:
+            return;
```
alalek: > return;

This is an error case. You should use `CV_Error()` here.

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -3673,6 +3673,8 @@ channels is derived automatically from src and code.
  */
 CV_EXPORTS_W void cvtColor( InputArray src, OutputArray dst, int code, int dstCn = 0 );
 
+CV_EXPORTS_W void cvtColorTwoPlane( InputArray src1, InputArray src2, OutputArray dst, int code, int dcn = 0 );
```
alalek: > dcn

Why? RGB/RGBA codes are different, so we can properly detect "dcn" value like `cvtColor` does.

modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
```diff
@@ -0,0 +1,393 @@
...
+
+        @Override
+        public Mat rgba() {
+            if (mPreviewFormat == ImageFormat.NV21)
+                Imgproc.cvtColor(mYuvFrameData, mRgba, Imgproc.COLOR_YUV2RGBA_NV21, 4);
+            else if (mPreviewFormat == ImageFormat.YV12)
+                Imgproc.cvtColor(mYuvFrameData, mRgba, Imgproc.COLOR_YUV2RGB_I420, 4);  // COLOR_YUV2RGBA_YV12 produces inverted colors
+            else if (mPreviewFormat == ImageFormat.YUV_420_888) {
+                assert(mUVFrameData != null);
+                Imgproc.cvtColorTwoPlane(mYuvFrameData,mUVFrameData,mRgba,Imgproc.COLOR_YUV2RGB_NV21, 4);
```
alalek:
```.diff
-COLOR_YUV2RGB_NV21
+COLOR_YUV2RGBA_NV21
```

modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
```diff
@@ -0,0 +1,393 @@
...
+                    assert(planes[0].getPixelStride() == 1);
+                    assert(planes[1].getPixelStride() == 2);
+                    assert(planes[2].getPixelStride() == 2);
+
+                    ByteBuffer y_plane = planes[0].getBuffer();
+                    ByteBuffer uv_plane = planes[1].getBuffer();
+                    Mat y_mat = new Mat( h, w, CvType.CV_8UC1, y_plane );
+                    Mat uv_mat = new Mat( h/2, w, CvType.CV_8UC1, uv_plane );
+                    JavaCamera2Frame tempFrame = new JavaCamera2Frame(y_mat,uv_mat,w,h);
+                    deliverAndDrawFrame(tempFrame);
```
alalek: Original `JavaCameraView` interface is asynchronous.
Image processing must be performed in a separate thread, see `CameraWorker`. This callback should just update the latest frame for next processing (non-processed frames are dropped).

modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
```diff
@@ -0,0 +1,393 @@
...
+                    // see also https://developer.android.com/reference/android/graphics/ImageFormat.html#YUV_420_888
+                    // Y plane (0) non-interleaved => stride == 1; U/V plane interleaved => stride == 2
+                    assert(planes[0].getPixelStride() == 1);
+                    assert(planes[1].getPixelStride() == 2);
+                    assert(planes[2].getPixelStride() == 2);
+
+                    ByteBuffer y_plane = planes[0].getBuffer();
+                    ByteBuffer uv_plane = planes[1].getBuffer();
+                    Mat y_mat = new Mat( h, w, CvType.CV_8UC1, y_plane );
+                    Mat uv_mat = new Mat( h/2, w, CvType.CV_8UC1, uv_plane );
```
alalek:
```.diff
-Mat( h/2, w, CvType.CV_8UC1, uv_plane );
+Mat(h/2, w/2, CvType.CV_8UC2, uv_plane);
```

modules/imgproc/include/opencv2/imgproc.hpp
floe: Ah, right. How about the name? Should I just overload `cvtColor`? After thinking about it a bit more, that actually seems cleaner to me...

modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
floe: Yes, but the `onImageAvailable` listener will already be called within the `mBackgroundHandler` thread. Isn't this equivalent?

alalek: Thread is "background" but it may block camera frames fetching and cause camera timeout/disconnection.
I have no experience with new Camera2 API, could you add some documentation link about this (callback processing/requirements)?

modules/imgproc/src/color.cpp
floe: OK, fixed.

modules/imgproc/include/opencv2/imgproc.hpp
floe: OK, fixed. Method name still as-is.

modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
floe: OK, fixed.

modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
floe: OK, fixed.

modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
floe: See here: https://developer.android.com/reference/android/media/ImageReader.html#acquireLatestImage()
Since I'm calling `acquireLatestImage`, the queue will be emptied and only the most recent image will be passed on for processing. The documentation also mentions that "This function is recommended to use over acquireNextImage() for most use-cases, as it's more suited for real-time processing." 
To be on the safe side, I'll test what happens if the processing takes a really long time (let's say 500 ms).

modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
```diff
@@ -0,0 +1,393 @@
...
+                        }
+                    }, mBackgroundHandler);
+        } catch (CameraAccessException e) {
+            Log.e(LOGTAG, "createCameraPreviewSession");
+        } catch (InterruptedException e) {
+            throw new RuntimeException(
+                    "Interrupted while createCameraPreviewSession", e);
+        }
+        finally {
+            //mCameraOpenCloseLock.release();
```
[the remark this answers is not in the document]
floe: It was already part of the original `JavaCameraView`, so I've kept it. However, it's quite possible that this is a kludge from the Camera1 API that's no longer needed for Camera2. I'll see if I can just get rid of it (I'd assume yes ;-)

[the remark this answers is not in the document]
alalek: Agreed.
Looks like locks are no longer needed (frames are processed in "sync" with callback with Camera2 API).

floe:
Removed the old semaphore and ran another test; seems to work just as well. Ready for merging IMHO.

floe:
Hm. The Linux OpenCL build failed, looks like a transient build error?

alalek:
Build failure is not related to your patch.

I run some tests with `JavaCamera2View` interface via OpenCV "tutorial2" sample and observing problem with application pause/resume events processing (pressing Android 'Home' button). Does it work for you (pause/restore handling)?

floe:
I'll have to test that again. What would be the expected behaviour with the old `JavaCameraView`?

alalek:
> expected behaviour

Application activity pause/resume handling should work with app crash/closing

alalek:
@floe I pushed several updates to this patch. Please take a look.

floe:
Thanks for your help!

PhilLab:
@floe Awesome, can't wait to try it. Do you know if the current implementation respects the camera orientation? https://github.com/opencv/opencv/issues/9269 and https://github.com/opencv/opencv/issues/8866

floe:
@PhilLab sorry for the late reply, have in fact not tried this yet. But since the code is based on `CameraBridgeViewBase` too, I suspect it will have the same issues.

ValkA:
I have an Android device that only works with `ImageReader` of `ImageFormat.YV12`.
The output is 3 planes as following:

```
Mat y_mat = new Mat(h, w, CvType.CV_8UC1, image.getPlanes()[0].getBuffer());
Mat u_mat = new Mat(h/2, w/2, CvType.CV_8UC1, image.getPlanes()[1].getBuffer());
Mat v_mat = new Mat(h/2, w/2, CvType.CV_8UC1, image.getPlanes()[2].getBuffer());
```

Do you have an idea how to convert all these into a full sized RGBA without copying/merging matrices?

ValkA:
ok inspired by JavaCameraView, this one listener handles both image formats :)
```
ImageReader.OnImageAvailableListener listener = new ImageReader.OnImageAvailableListener {
        int w,h;
        public OnImageAvailableListener(int w, int h){
            this.w = w;
            this.h = h;
        }

        final Mat rgba = new Mat(h,w,CvType.CV_8UC4);

        @Override
        public void onImageAvailable(ImageReader imageReader) {
            Image image = imageReader.acquireLatestImage();
            if(image == null) return;

            Mat yuv420 = new Mat(h+h/2, w, CvType.CV_8UC1, image.getPlanes()[0].getBuffer());
            switch (imageReader.getImageFormat()){
                case ImageFormat.YV12:
                    Imgproc.cvtColor(yuv420, rgba, Imgproc.COLOR_YUV2RGBA_YV12, 4);
                    break;
                case ImageFormat.YUV_420_888:
                    Imgproc.cvtColor(yuv420, rgba, Imgproc.COLOR_YUV2BGR_NV12, 4);
                    break;
            }

            //Do your thing here...

            Utils.matToBitmap(rgba, bitmap);
            image.close();
            cameraView.setImageBitmap(bitmap);
        }
    }
```

ValkA:
To fix color conversion bugs (which I had in various devices/emulators) I would propose to do the following 

in JavaCamera2View I would change JavaCamera2Frame to:

```
    private class JavaCamera2Frame implements CvCameraViewFrame {
        @Override
        public Mat gray() {
            return new Mat(mImage.getHeight(), mImage.getWidth(), CvType.CV_8UC1, mImage.getPlanes()[0].getBuffer());
        }

        @Override
        public Mat rgba() {
            Mat yuv420 = new Mat((3*mImage.getHeight())/2, mImage.getWidth(), CvType.CV_8UC1, mImage.getPlanes()[0].getBuffer());
            switch (mImage.getFormat()){
                case ImageFormat.YV12:
                    Imgproc.cvtColor(yuv420, mRgba, Imgproc.COLOR_YUV2RGBA_YV12, 4);
                    break;
                case ImageFormat.NV21:
                    Imgproc.cvtColor(yuv420, mRgba, Imgproc.COLOR_YUV2RGBA_NV21, 4);
                    break;
                case ImageFormat.YUV_420_888:
                    Imgproc.cvtColor(yuv420, mRgba, Imgproc.COLOR_YUV2RGBA_I420, 4);
                    break;
                default:
                    throw new IllegalArgumentException("Preview Format can be NV21 or YV12");
            }

            return mRgba;
        }

        public JavaCamera2Frame(Image image) {
            super();
            mImage = image;
            mRgba = new Mat();
        }

        public void release() {
            mRgba.release();
        }

        private Mat mRgba;
        private Image mImage;
    };
```

and OnImageAvailableListener to:
```
mImageReader.setOnImageAvailableListener(new ImageReader.OnImageAvailableListener() {
                @Override
                public void onImageAvailable(ImageReader reader) {
                    long t0 = System.currentTimeMillis();

                    Image image = reader.acquireLatestImage();
                    if (image == null)
                        return;

                    JavaCamera2Frame tempFrame = new JavaCamera2Frame(image);
                    deliverAndDrawFrame(tempFrame);
                    tempFrame.release();
                    image.close();

                    long t1 = System.currentTimeMillis();
                    Log.d(LOGTAG, "" + 1000/(t1-t0) + "fps");
                }
            }, mBackgroundHandler);
```

ansman:
@ValkA Your code only uses the Y plane which would lead to a gray scale picture.

ValkA:
It takes also UV in the extra h/2 rows

ansman:
Ah, I see. It does seem to be continuous but that is probably a dangerous assumption that can lead to strange issues down the line.

ValkA:
Yeah, also when lines are padded it has issues

ansman:
@ValkA, @floe, @alalek I've done some experiments and it seems that if you use a different size for your image reader than your surface (or perhaps it's just different from the standard 1920x1080) you don't actually get NV21 because of padding between the rows.

Furthermore it seems that V always comes before U in memory so just taking the second plane as an `uv` buffer seems incorrect (if anything it should take the third plane and treat it as `vu`)

alalek:
@ansman Could you please check this patch on your device: #14505 ?

ansman:
@alalek I added a review for the PR

PR #5512, opened 2015-10-14
load multi/hyperspectral images with using gdal

edgarriba:
Since now GDAL driver only supports to embed data in Mats with 1 or 3 channels.

With this patch it is able to parse data into N channels Mat in case you want to load e.g. multispectral and hyperspectral images.

Moreover, added small optimization for data injection into Mat by using ptr access.

PR #10093, opened 2017-11-15
Fix GDAL image decoding color problems identified by issue #10089, by:

### LINKED ISSUE #10089: IMREAD_LOAD_GDAL does not correctly handle color table gif or color jpeg files
- OpenCV => 3.3.1
##### Detailed description

The GDAL decoder in imgcodecs/src/grfmt_gdal.cpp does not correctly handle gif files that use a color table. The image created will be greyscale. In addition, color images of other formats, for example jpeg, may have their color components mixed up, as the GDAL decoder does not respect the color interpretation of the bands, as given by GDAL. In fact, the colors are taken to be blue, green, red, when they are usually red, green, blue (resulting in the problem described in issue #9809).

These problem can be reproduced using program code below and the attached images.

I have code fixes to suggest that I will create a pull request for.

colorchips
colorchips
colorchips

##### Steps to reproduce

```.cpp
/*
 * Test GDAL image read
 */

#include <iostream>

#include <opencv2/highgui/highgui.hpp>

void printUsage(char *progName)
{
    std::cout << "Usage: " << progName << " [option(s)] image1 ...\n";
    std::cout << "Options:\n"
              << "  -h   Help\n";
}

int main( int argc, char** argv )
{
  int firstArg = 1;
  
  if (argc > 1)
  {
    if (strcmp(argv[1], "-h") == 0)
    {
      printUsage(argv[0]);
      return -1;
    }
  }

  if((argc-firstArg) == 0)
  {
    printUsage(argv[0]);
    return -1;
  }

  // Create a window for display.
  cv::namedWindow( "Display window", cv::WINDOW_AUTOSIZE );

  for (size_t i=firstArg; i < argc; i++)
  {
    std::cout << "Loading image file " << argv[i] << "...\n";

    cv::Mat image;

    // Read the file
    int readFlags = CV_LOAD_IMAGE_COLOR | cv::IMREAD_LOAD_GDAL;
    image = cv::imread(argv[i], readFlags);

    // Check for invalid input
    if(image.empty())
    {
      std::cerr <<  "Could not open or find the image\n";
    }
    else
    {
      std::cout << "Read " << image.cols << " x " << image.rows << " x "
                << image.elemSize() << " image\n";

      // Show our image inside it.
      cv::imshow( "Display window", image );
    }

    // Wait for a keystroke in the window
    cv::waitKey(0);
  }
  return 0;
}
```

gdkessler:
Fixing CV_8UC1 symbol, which should be CV_8UC3 for RGB GDAL color table images.

Fixing image.ptr<VecX>(row,col)[] to be (*image.ptr<VecX>(row,col))[] to correctly access VecX array elements, as ptr<VecX>() returns a pointer to the VecX, not the first element of VecX. This fixes the color problem with color table gif images, and avoids out-of-bounds memory access.

Respecting the color identification of raster bands provided by the GDAL image driver, and produce BGR or BGRA images. Note that color bands of images using the HSL, CMY, CMYK, or YCbCr color space are ignored, rather than converting them to BGR.

resolves #10089

### This pullrequest changes

Changes modules/imgcodecs/src/grfmt_gdal.cpp to correct image read using the GDAL decoder.
Test images:
[GdalTestImages.zip](https://github.com/opencv/opencv/files/1475472/GdalTestImages.zip)

Code to test changes:
```.cpp
/*
 * Test GDAL image read
 */

#include <iostream>

#include <opencv2/highgui/highgui.hpp>

void printUsage(char *progName)
{
    std::cout << "Usage: " << progName << " [option(s)] image1 ...\n";
    std::cout << "Options:\n"
              << "  -h   Help\n";
}

int main( int argc, char** argv )
{
  int firstArg = 1;
  
  if (argc > 1)
  {
    if (strcmp(argv[1], "-h") == 0)
    {
      printUsage(argv[0]);
      return -1;
    }
  }

  if((argc-firstArg) == 0)
  {
    printUsage(argv[0]);
    return -1;
  }

  // Create a window for display.
  cv::namedWindow( "Display window", cv::WINDOW_AUTOSIZE );

  for (size_t i=firstArg; i < argc; i++)
  {
    std::cout << "Loading image file " << argv[i] << "...\n";

    cv::Mat image;

    // Read the file
    int readFlags = CV_LOAD_IMAGE_COLOR | cv::IMREAD_LOAD_GDAL;
    image = cv::imread(argv[i], readFlags);

    // Check for invalid input
    if(image.empty())
    {
      std::cerr <<  "Could not open or find the image\n";
    }
    else
    {
      std::cout << "Read " << image.cols << " x " << image.rows << " x "
                << image.elemSize() << " image\n";

      // Show our image inside it.
      if ((image.depth() == CV_16S) || (image.depth() == CV_16U))
      {
        cv::convertScaleAbs(image, image, 1.0/256.0);
      }
      else if (image.depth() == CV_32S)
      {
        cv::convertScaleAbs(image, image, 1.0/(256.0*256.0*256.0));
      }
      else if ((image.depth() == CV_32F) || (image.depth() == CV_64F))
      {
        cv::normalize(image, image, 255, 0);
      }
      cv::imshow( "Display window", image );
    }

    // Wait for a keystroke in the window
    cv::waitKey(0);
  }
  return 0;
}
```

```
```

### DIFF
--- a/modules/imgcodecs/src/grfmt_gdal.cpp
+++ b/modules/imgcodecs/src/grfmt_gdal.cpp
@@ -78,7 +78,7 @@ int  gdalPaletteInterpretation2OpenCV( GDALPaletteInterp const& paletteInterp, G
 
         /// RGB
         case GPI_RGB:
-            if( gdalType == GDT_Byte    ){ return CV_8UC1;  }
+            if( gdalType == GDT_Byte    ){ return CV_8UC3;  }
             if( gdalType == GDT_UInt16  ){ return CV_16UC3; }
             if( gdalType == GDT_Int16   ){ return CV_16SC3; }
             if( gdalType == GDT_UInt32  ){ return CV_32SC3; }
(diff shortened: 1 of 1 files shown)


modules/imgcodecs/src/grfmt_gdal.cpp
```diff
@@ -256,35 +256,35 @@ void write_pixel( const double& pixelValue,
...
     // input: 3 channel, output: 3 channel
     else if( gdalChannels == 3 && image.channels() == 3 ){
-        if( image.depth() == CV_8U ){  image.at<Vec3b>(row,col)[channel] = newValue;  }
-        else if( image.depth() == CV_16U ){  image.ptr<Vec3s>(row,col)[channel] = newValue;  }
-        else if( image.depth() == CV_16S ){  image.ptr<Vec3s>(row,col)[channel] = newValue;  }
-        else if( image.depth() == CV_32S ){  image.ptr<Vec3i>(row,col)[channel] = newValue;  }
-        else if( image.depth() == CV_32F ){  image.ptr<Vec3f>(row,col)[channel] = newValue;  }
-        else if( image.depth() == CV_64F ){  image.ptr<Vec3d>(row,col)[channel] = newValue;  }
+        if( image.depth() == CV_8U ){  (*image.ptr<Vec3b>(row,col))[channel] = newValue;  }
+        else if( image.depth() == CV_16U ){  (*image.ptr<Vec3s>(row,col))[channel] = newValue;  }
```
alalek: This doesn't looks as useful change. Why do you need it?

modules/imgcodecs/src/grfmt_gdal.cpp
```diff
@@ -388,38 +388,66 @@ bool GdalDecoder::readData( Mat& img ){
...
+            break;
+        case GCI_RedBand:
+            color = 2;
+            break;
+        case GCI_AlphaBand:
+            color = 3;
+            break;
+        default:
+            color = -1;
+            break;
```
alalek: Probably we should replace code to this in default case:
```
CV_Error(Error::StsNotImplemented, "unsupported/invalid image format");
```

Also we could remove `if (color >= 0)` below.

modules/imgcodecs/src/grfmt_gdal.cpp
gdkessler: image.ptr<Vec3b>(row,col) is not the same as image.at<Vec3b>(row,col). The image.ptr<Vec3b>() method returns a pointer to a Vec3b block, whereas image.at<Vec3b>() returns the Vec3b block itself.
In practical terms, (*image.ptr<Vec3b>(0,0))[1] will give you the 2nd byte in the data block, which is what you want, but image.ptr<Vec3b>(0,0)[1] will give you the 4th byte in the data block (the start of the 2nd Vec3b block), which is not what you want. In addition, image.ptr<Vec3b>(maxRow, maxCol)[1] will be an out-of-bounds memory access.

modules/imgcodecs/src/grfmt_gdal.cpp
gdkessler: That sounds like a good idea to me.

modules/imgcodecs/src/grfmt_gdal.cpp
gdkessler: The alternative fix would have been to change image.ptr<VecX>(row,col)[channel] to image.at<VecX>(row,col)[channel], but I assumed that image.ptr() was used to avoid additional bounds checks for each access that are done by image.at().

gdkessler: Sorry, I noticed that the template block are missing in my comments.  Here's another try:

image.ptr&lt;Vec3b&gt;(row,col) is not the same as image.at&lt;Vec3b&gt;(row,col). The image.ptr&lt;Vec3b&gt;() method returns a pointer to a Vec3b block, whereas image.at&lt;Vec3b&gt;() returns the Vec3b block itself.
In practical terms, (*image.ptr&lt;Vec3b&gt;(0,0))[1] will give you the 2nd byte in the data block, which is what you want, but image.ptr&lt;Vec3b&gt;(0,0)[1] will give you the 4th byte in the data block (the start of the 2nd Vec3b block), which is not what you want. In addition, image.ptr&lt;Vec3b&gt;(maxRow, maxCol)[1] will be an out-of-bounds memory access.

alalek: I see, there is corrent CV_8U case only.
BTW, `(*image.ptr<Vec3s>(row,col))` should be equal to `image.at<Vec3s>(row,col)` semantically.

gdkessler: Right. The difference is that the CV_DbgAssert() calls in at() are more expensive:

```.cpp
    CV_DbgAssert((unsigned)(i1 * DataType<_Tp>::channels) < (unsigned)(size.p[1] * channels()));
    CV_DbgAssert(CV_ELEM_SIZE1(traits::Depth<_Tp>::value) == elemSize1());
```

versus this in ptr():

```.cpp
    CV_DbgAssert((unsigned)i1 < (unsigned)size.p[1]);
```

It's a performance issue rather than a semantic issue. I would note that the original code, before commit 1e9bd59f on 10/14/2015, used at() methods. This is likely where the bug was introduced in the interest of performance improvement.

alalek: `CV_DbgAssert()` are only for debug build mode. So there is no performance question at all.

Agree #5512 introduced bug here.

modules/imgcodecs/src/grfmt_gdal.cpp
gdkessler: I've pushed this update. Let me know if it is what you had in mind.

PR #10095, opened 2017-11-15
imgproc(Canny): eliminate unnecessary operations

alalek:
drop manual loop unrolling:
- don't block compiler optimizations
- no effect on i5

BTW, `v_load_halves()` is not used anymore in OpenCV code.


alalek:
:+1:

PR #10099, opened 2017-11-16
js: fix tests (qunit -> node-qunit)

alalek:
https://www.npmjs.com/package/node-qunit:
> Up until version 1.0.0, this package was published under the name qunit. That name is now used by the official QUnit package and CLI, and this package will be published as node-qunit from version 1.0.0 onward.

```
```


huningxin:
Looks good. Thanks for fixing!

alalek:
:+1:

PR #10107, opened 2017-11-17
medianBlur hal

ElenaGvozdeva:
### This pullrequest changes

1. Change border type for medianBlur to BORDER_ISOLATED
1. Add hal interface for medianBlur

```
force_builders=ARMv7,ARMv8
```

### DIFF
--- a/modules/imgproc/src/hal_replacement.hpp
+++ b/modules/imgproc/src/hal_replacement.hpp
@@ -617,14 +617,14 @@ inline int hal_ni_integral(int depth, int sdepth, int sqdepth, const uchar * src
 
 /**
    @brief Calculate medianBlur filter
-   @param depth Depths of source and destination image
    @param src_data,src_step Source image
    @param dst_data,dst_step Destination image
    @param width,height Source image dimensions
+   @param depth Depths of source and destination image
    @param cn Number of channels
    @param ksize Size of kernel
 */
-inline int hal_ni_medianBlur(int depth, const uchar* src_data, size_t src_step, uchar* dst_data, size_t dst_step, int width, int height, int cn, int ksize) { return CV_HAL_ERROR_NOT_IMPLEMENTED; }
+inline int hal_ni_medianBlur(const uchar* src_data, size_t src_step, uchar* dst_data, size_t dst_step, int width, int height, int depth, int cn, int ksize) { return CV_HAL_ERROR_NOT_IMPLEMENTED; }
 
 //! @cond IGNORED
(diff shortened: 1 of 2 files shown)


modules/imgproc/src/hal_replacement.hpp
```diff
@@ -615,6 +615,21 @@ inline int hal_ni_integral(int depth, int sdepth, int sqdepth, const uchar * src
...
+/**
+   @brief Calculate medianBlur filter
+   @param depth Depths of source and destination image
+   @param src_data,src_step Source image
+   @param dst_data,dst_step Destination image
+   @param width,height Source image dimensions
+   @param cn Number of channels
+   @param ksize Size of kernel
+*/
+inline int hal_ni_medianBlur(int depth, const uchar* src_data, size_t src_step, uchar* dst_data, size_t dst_step, int width, int height, int cn, int ksize) { return CV_HAL_ERROR_NOT_IMPLEMENTED; }
```
alalek: BTW, depth and cn:
- may be combined into "int type"
- or placed on closer parameters positions

(Anyway looks like this is grabbed from other functions design)

vpisarev:
@ElenaGvozdeva, I agree with @alalek, depth and cn should be combined together. In warpAffine and other such functions the parameter is placed first, I think, we can do the same here. Otherwise, I'm fine with the patch

alalek:
There is few lines above already existed "integral" interface with the same "problem":
> inline int hal_ni_integral(**int depth**, int sdepth, int sqdepth, const uchar * src_data, size_t src_step, uchar * sum_data, size_t sum_step, uchar * sqsum_data, size_t sqsum_step, uchar * tilted_data, size_t tilted_step, int width, int height, **int cn**)

Perhaps @mshabunin can clarify idea of this approach.

mshabunin:
We did not expose `CV_MAT_CN` and other macros needed to retrieve depth and channels from type to the HAL interface (https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/hal/interface.h).

`cn` is put to the end of the parameter list of _integral_ function to match internal implementation interfaces (see https://github.com/opencv/opencv/commit/3d5c0f1faf8223a9778faf588a75a393a6231b2e for deails). I 
suggest the following parameters order to make it similar to `cvt...` functions:
```.cpp
int hal_ni_medianBlur(const uchar* src_data, size_t src_step, uchar* dst_data, size_t dst_step, int width, int height, int depth, int cn, int ksize)
```

vpisarev:
:+1:

PR #10124, opened 2017-11-21
Fix customPreviewLayer layout issue when rotateVideo is NO in iOS

zhongwuzw:
We should not adjust the `bounds` and `transform` when `rotateVideo` is NO.


vpisarev:
:+1:

PR #10126, opened 2017-11-21
dnn: drop OpenCL code path for DetectionOutputLayer

### LINKED ISSUE #10125: DNN_TARGET_OPENCL for MobileNet-SSD

- OpenCV => 3.3.1-dev
- Compiler => /usr/bin/c++  (ver 5.4.0)

##### Detailed description
OpenCL target for MobileNet-SSD object detection model produces an assertion.

##### Steps to reproduce

Add
```cpp
net.setPreferableTarget(DNN_TARGET_OPENCL);
```
into [samples/dnn/ssd_mobilenet_object_detection.cpp](https://github.com/opencv/opencv/blob/master/samples/dnn/ssd_mobilenet_object_detection.cpp):

```
OpenCV Error: Assertion failed (u->refcount == 0 && "UMat deallocation error: some derived Mat is still alive") in deallocate, file /home/dkurtaev/opencv/modules/core/src/ocl.cpp, line 3619
  what():  /home/dkurtaev/opencv/modules/core/src/ocl.cpp:3619: error: (-215) u->refcount == 0 && "UMat deallocation error: some derived Mat is still alive" in function deallocate
```

alalek:
getUMat()/getMat() calls are scope based. Results of these calls can't be stored somewhere for future usage.

Also DetectionOutputLayer implementation doesn't have OpenCL processing so separate OpenCL call is not needed (no benefits).

resolves #10125

/cc @pengli @wzw-intel

### DIFF
--- a/modules/core/include/opencv2/core/mat.hpp
+++ b/modules/core/include/opencv2/core/mat.hpp
@@ -357,6 +357,9 @@ public:
 
     void assign(const UMat& u) const;
     void assign(const Mat& m) const;
+
+    void assign(const std::vector<UMat>& v) const;
+    void assign(const std::vector<Mat>& v) const;
 };
 
 
--- a/modules/core/src/matrix.cpp
+++ b/modules/core/src/matrix.cpp
@@ -3053,6 +3053,82 @@ void _OutputArray::assign(const Mat& m) const
 }
 
 
+void _OutputArray::assign(const std::vector<UMat>& v) const
+{
+    int k = kind();
+    if (k == STD_VECTOR_UMAT)
+    {
+        std::vector<UMat>& this_v = *(std::vector<UMat>*)obj;
+        CV_Assert(this_v.size() == v.size());
+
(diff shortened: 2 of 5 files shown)


dkurt:
@alalek, I think we need to add a code similar to the following lines into `Layer::forward_fallback` at dnn.cpp:
```cpp
this->forward(inputs, outputs, internals);

if (outputs_arr.isUMatVector())
{
    for (int i = 0; i < outputs.size(); ++i)
    {
        outputs[i].getUMat(ACCESS_READ).copyTo(outputs_arr.getUMatRef(i));
    }
}
```
The thing is detection_output_layer call `create` for `cv::Mat` connected to `cv::UMat` that had been allocated only for single object (`outputs.resize(1, shape(1, 1, 1, 7));`). So if there are more objects actual Mat will be asynchronous to origin UMat. You may reproduce it using the same sample with `confidence_threshold: 0.01` argument at corresponding `.prototxt`.

dkurt:
:+1: Thanks! That works now.

PR #10133, opened 2017-11-22
 Python: wrap Algorithm::read and Algorithm::write

paroj:


alalek:
Lets keep it here: #9247

PR #10092, opened 2017-11-15
dnn: rename caffe.proto

alalek:
Should resolve caffe issues like [these](https://www.google.com/search?q=File+already+exists+in+database%3A+caffe.proto):
> [libprotobuf ERROR google/protobuf/descriptor_database.cc:57] File already exists in database: caffe.proto

PR #10139, opened 2017-11-22
dnn: rename caffe proto package

alalek:
related #10092


vpisarev:
@dkurt, could you please review it?

dkurt:
@vpisarev, @alalek ,  I'm fine with it.

PR #10141, opened 2017-11-22
Add ReLU and LeakyReLU activation function in ml module

LaurentBerger:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/409

Add ReLU and LeakyReLU activation function in ml module

### DIFF
--- a/modules/ml/test/test_mltests2.cpp
+++ b/modules/ml/test/test_mltests2.cpp
@@ -85,6 +85,22 @@ int str_to_ann_train_method( String& str )
     return -1;
 }
 
+int str_to_ann_activation_function(String& str)
+{
+    if (!str.compare("IDENTITY"))
+        return ANN_MLP::IDENTITY;
+    if (!str.compare("SIGMOID_SYM"))
+        return ANN_MLP::SIGMOID_SYM;
+    if (!str.compare("GAUSSIAN"))
+        return ANN_MLP::GAUSSIAN;
+    if (!str.compare("RELU"))
+        return ANN_MLP::RELU;
+    if (!str.compare("LEAKYRELU"))
+        return ANN_MLP::LEAKYRELU;
+    CV_Error(CV_StsBadArg, "incorrect ann activation function string");
+    return -1;
+}
+
 void ann_check_data( Ptr<TrainData> _data )
 {
     CV_TRACE_FUNCTION();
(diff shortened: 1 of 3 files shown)


alalek:
@LaurentBerger Please add tests for new functionality.

LaurentBerger:
@alalek I try to understand what is a test.
In opencv_extra/testdata/ml/legacy there is a [Neural Network](https://github.com/opencv/opencv_extra/blob/master/testdata/ml/legacy/ann_waveform.xml) trains with [data](https://github.com/opencv/opencv_extra/blob/master/testdata/ml/waveform.data).
I think I must train a NN using same data with RELU and LeakyRelu and writ NN in opencv_extra. is it right?

alalek:
I'm assume something like these steps ("/" below means separation between two modes: testdata generation mode / algorithm verification (normal) mode): 
- create simple model (or reuse existed ANN model if it is applicable) with new activation functions
- generate train data + save to opencv_extra (or nothing in case of reusing existed) / load train data from opencv_extra
- model.train()
- save train result to opencv_extra / load reference train result from opencv_extra and verify (somehow) trained result from previous step (norm L2 relative?)
- get reference trained model
- model.predict(inputs  from <train data>)
- save predict() result to opencv_extra / load reference predict() result from opencv_extra and verify (somehow) predicted result from previous step (norm L2 relative?)

Split testdata generation mode and normal verification mode via `#if`:
```
// Uncomment line below to re-generate testdata only.
// Keep line commented to run verification mode.
//#define GENERATE_TESTDATA 
...
result <== model.train(<traindata>)
#ifdef GENERATE_TESTDATA
... save trained "result" model to opencv_extra (model.save(fileStorage) ?) ...
#else
... "reference" = load reference model from opencv_extra ...
... make some checks between "result" model and "reference" ...
#endif
...
```

vpisarev:
:+1:

modules/ml/test/test_mltests2.cpp
```diff
@@ -177,6 +193,62 @@ float ann_calc_error( Ptr<StatModel> ann, Ptr<TrainData> _data, map<int, int>& c
...
+        Ptr<ml::ANN_MLP> y = Algorithm::load<ANN_MLP>(dataname + activationName[i] + ".yml");
+        ASSERT_TRUE(y != NULL) << "Could not load   " << dataname + activationName[i] + ".yml";
+        Mat testSamples = tdata->getTestSamples();
+        Mat rx, ry, dst;
+        x->predict(testSamples, rx);
+        y->predict(testSamples, ry);
+        absdiff(rx, ry, dst);
+        double minVal, maxVal;
+        minMaxLoc(dst, &minVal, &maxVal);
+        ASSERT_TRUE(maxVal<FLT_EPSILON) << "Predict are not equal for " << dataname + activationName[i] + ".yml and " << activationName[i];
```
alalek: BTW, absdiff()+minMaxLoc() => `[cvtest::]norm(rx, ry, NORM_INF)` ([documentation](https://docs.opencv.org/3.3.0/d2/de8/group__core__array.html#gad12cefbcb5291cf958a85b4b67b6149f))

`cvtest` namespace is recommended for test checks.

Riyuzakii:
@LaurentBerger @alalek , Is there some problem with using the relu function currently?
```
    mlp->setActivationFunction(ANN_MLP::SIGMOID_SYM, 0.5, 1);
    mlp->setActivationFunction(ANN_MLP::RELU, 0.5, 1);
```
the first line for sigmoid function works but for relu it says:
```
‘RELU’ is not a member of ‘cv::ml::ANN_MLP’
```
Can you think of any reason as to why this is happening?

LaurentBerger:
ANN_MLP::RELU was introduced in opencv 3.4.0 (Dec 2017)

PR #10145, opened 2017-11-23
Create avi container for mjpeg

allnes:


modules/videoio/src/cap_mjpeg_decoder.cpp
```diff
@@ -888,8 +377,9 @@ bool MotionJpegCapture::parseRiff(MjpegInputStream& in_str)
             //RiffList::m_size includes fourCC field which we have already read
             next_riff += (riff_list.m_size - 4);
 
-            AviMjpegStream mjpeg_video_stream;
-            bool is_parsed = mjpeg_video_stream.parseAvi(in_str, m_mjpeg_frames);
+            AVIReadContainer<MjpegInputStream> mjpeg_video_stream;
```
mshabunin: It will be better to move _MjpegInputStream_ to the same place where _AVIReadContainer_ is located, then you will be able to avoid templates in the code.

allnes: @mshabunin And unless the task was not to separate mjpeg codec and avi container, it turns out they will be together to one file again

mshabunin: I don't see a problem here: _MjpegInputStream_ is just a convenient object for reading AVI files, it does not implement MJPEG specific behavior.

modules/videoio/src/cap_avi_container.cpp
```diff
@@ -0,0 +1,799 @@
...
+
+    if(in_str && strh.m_four_cc == STRH_CC)
+    {
+        uint64_t next_strl_list = in_str->tellg();
+        next_strl_list += strh.m_size;
+
+        AviStreamHeader strm_hdr;
+        *in_str >> strm_hdr;
+
+        if(strm_hdr.fccType == VIDS_CC && strm_hdr.fccHandler == MJPG_CC)
```
mshabunin: Probably we should pass FourCC code (_MJPG_CC_ here) as a parameter into this method.

modules/videoio/src/cap_mjpeg_decoder.cpp
allnes: @mshabunin I already remade, you can see

modules/videoio/src/cap_avi_container.cpp
allnes: @mshabunin corrected

modules/videoio/src/cap_mjpeg_decoder.cpp
```diff
@@ -40,652 +40,11 @@
 //M*/
 
 #include "precomp.hpp"
-#include <deque>
-#include <stdint.h>
+#include "opencv2/core/private.cap_avi_container.hpp"
```
mshabunin: Let's rename this header to `container_avi.private.hpp` and it should be moved to the _videoio_ module.

modules/core/include/opencv2/core/private.cap_avi_container.hpp
```diff
@@ -0,0 +1,479 @@
...
+    uint64_t   m_movi_start;
+    uint64_t   m_movi_end;
+    frame_list m_frame_list;
+    uint32_t   m_width;
+    uint32_t   m_height;
+    double     m_fps;
+    bool       m_is_indx_present;
+};
+
+static const int AVIH_STRH_SIZE = 56;
```
mshabunin: All variables and functions marked `static` should be moved to the `.cpp` file.

modules/videoio/test/test_avi_container.cpp
```diff
@@ -0,0 +1,190 @@
...
+// loss of use, data, or profits; or business interruption) however caused
+// and on any theory of liability, whether in contract, strict liability,
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+
+#include "test_precomp.hpp"
+#include "opencv2/core/private.cap_avi_container.hpp"
+#include "opencv2/highgui.hpp"
```
mshabunin: This include should be removed.

mshabunin:
@allnes , we have updated test writing guidelines, please take a look: https://github.com/opencv/opencv/wiki/Coding_Style_Guide#implementing-tests

You have to wrap your test into namespace `opencv_test`.

modules/videoio/test/test_container_avi.cpp
```diff
@@ -0,0 +1,86 @@
...
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include "test_precomp.hpp"
+#include "opencv2/videoio/container_avi.private.hpp"
+#include <cstdio>
+
+using namespace cv;
+using namespace std;
+using namespace opencv_test;
```
mshabunin: All source code should be put into this namespace.

modules/videoio/test/test_container_avi.cpp
```diff
@@ -0,0 +1,86 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include "test_precomp.hpp"
+#include "opencv2/videoio/container_avi.private.hpp"
+#include <cstdio>
+
+using namespace cv;
+using namespace std;
```
mshabunin: Avoid using namespace std. Use std:: if necessary (common types are imported into opencv_test namespace).

modules/videoio/test/test_precomp.hpp
```diff
@@ -41,4 +41,25 @@ inline void generateFrame(int i, int FRAME_COUNT, cv::Mat & frame)
 #endif
 }
 
+class BunnyParameters
+{
+private:
+    int bunny_width, bunny_height, bunny_fps, bunny_count;
```
mshabunin: It is not necessary to store values in the object:
```.cpp
class BunnyParameters {
public:
  inline static int getWidth() { return 672; }
  inline static getCount() { return cvRound(getFps() * getTime()); }
  ...

// call BunnyParameters::getCount()
```

Filename generation can be provided by this class too:
```.cpp
inline static String getFilename(const String &ext) { 
  return cvtest::TS::ptr()->get_data_path() + "video/big_buck_bunny" + ext; 
}
```

allnes:
@mshabunin I'm fixed

modules/videoio/test/test_video_io.cpp
```diff
@@ -151,12 +151,12 @@ typedef tuple<string, int> Backend_Type_Params;
...
 class Videoio_Bunny : public Videoio_Test_Base, public testing::TestWithParam<Backend_Type_Params>
 {
+    BunnyParameters bunny_param;
 public:
     Videoio_Bunny()
     {
         ext = get<0>(GetParam());
         apiPref = get<1>(GetParam());
-
         video_file = cvtest::TS::ptr()->get_data_path() + "video/big_buck_bunny." + ext;
```
mshabunin: Please use `bunny_param.getFilename(String(".") + ext)` here.

modules/videoio/test/test_container_avi.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include "opencv2/videoio/container_avi.private.hpp"
+#include <cstdio>
+
+using namespace cv;
+
+namespace opencv_test
+{
+
+TEST(videoio_avi, good_MJPG) {
+    String filename = cvtest::TS::ptr()->get_data_path() + "video/big_buck_bunny.mjpg.avi";
```
mshabunin: You should add `BunnyParameters::getFilename(".mjpg.avi");` here for consistency.

allnes:
@mshabunin fixed

modules/videoio/include/opencv2/videoio/container_avi.private.hpp
```diff
@@ -0,0 +1,192 @@
...
+#define CONTAINER_AVI_HPP
+
+#ifndef __OPENCV_BUILD
+#  error this is a private header which should not be used from outside of the OpenCV library
+#endif
+
+#include "opencv2/core/cvdef.h"
+#include "opencv2/videoio/videoio_c.h"
+#include <deque>
+#include <stdint.h>
```
alalek: This is C++11/C11 header: http://www.cplusplus.com/reference/cstdint/ like uint32_t and other types.

allnes:
@alalek fixed types, review please

PR #10150, opened 2017-11-23
dnn(ocl4dnn): refactor pooling

alalek:
/cc @pengli


pengli:
@alalek , patch looks good to me 👍

vpisarev:
:+1:

PR #10152, opened 2017-11-24
bindings(python): fix handling of 'isalgorithm'

alalek:


vpisarev:
can you please put here (or put a link to) the problem description that the patch solves?

alalek:
Problem is incorrect implementation. It doesn't properly handle multiple inheritance levels.

For example, `cv::ml::Boost` currently has invalid `isalgorithm = False`, however it is derived from cv::Algorithm:
```
class CV_EXPORTS_W Boost : public DTrees
class CV_EXPORTS_W DTrees : public StatModel
class CV_EXPORTS_W StatModel : public Algorithm
```

mshabunin:
BTW, we have a test for that: https://github.com/opencv/opencv/blob/master/modules/python/test/test_misc.py

vpisarev:
ok, thanks! :+1:

PR #10155, opened 2017-11-24
OCL: update OpenCL runtime loader

alalek:
- fix compilation on Apple (undefined 'oclpath')
- don't warn on `OPENCV_OPENCL_RUNTIME=disabled`

```
```


vpisarev:
:+1:

PR #10159, opened 2017-11-25
Convert Keypoint and DMatch objects into Python `dict`s

### LINKED ISSUE #10158: Increasing efficiency of the Keypoint and DMatch Python wrappers
##### Detailed description
Python wrapper for `Keypoint` and `DMatch` is totally inefficient in which we need to access each object separately in order to extract its attributes (center, radius, trainingIdx and queryIdx), rather than using Numpy arrays.

Is it possible to change that behavior to something similar to `connectedComponentsWithStats` wrapper of the `stats` variable?  That wrapper is a Numpy array, smoothly used in conjugation with an exported enum, which allows vectorized operations without any needs for loops.

Sahloul:
### This pullrequest changes
resolves #10158
- Convert DMatch objects into Python dicts
- Convert Keypoint objects into Python dicts

Although current python code will get broken, but fixing it is very easy
Just replace the dots with the brackets (e.g. `m['queryIdx']` instead of
`m.queryIdx`)
Alternatively, you can recover the legacy behaviour using namedtuple. Example:
```
from collections import namedtuple
m = namedtuple("DMatch", m.keys())(*m.values())
print(m.queryIdx)
```

The great news is that we can now vectorize the code and avoid using
loops to access the attributes, by converting the list of dicts into
numpy arrays. Example:

```
import pandas
matches = pandas.DataFrame(matches).values #queryIdx, trainIdx, imgIdx, distance
src_kps[matches[0]]
```

### DIFF
--- a/modules/core/include/opencv2/core/types.hpp
+++ b/modules/core/include/opencv2/core/types.hpp
@@ -704,7 +704,7 @@ keypoint neighborhood is then analyzed by another algorithm that builds a descri
 represented as a feature vector). The keypoints representing the same object in different images
 can then be matched using cv::KDTree or another method.
 */
-class CV_EXPORTS_W_SIMPLE KeyPoint
+class CV_EXPORTS_W_MAP KeyPoint
 {
 public:
     //! the default constructor
@@ -801,7 +801,7 @@ public:
 query descriptor index, train descriptor index, train image index, and distance between
 descriptors.
 */
-class CV_EXPORTS_W_SIMPLE DMatch
+class CV_EXPORTS_W_MAP DMatch
 {
 public:
     CV_WRAP DMatch();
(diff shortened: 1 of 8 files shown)


alalek:
I would like to rejects this change, because
- it breaks Python API without fixing any bug or adding serious improvement.
- parameter type check doesn't work well for dicts. We require this for proper error messages.
- I see `pyopencv_from()`, bit where is `pyopencv_to()`?
- what is about `m.__dict__` / `vars(m)` / etc?
- what is about to expand existed objects classes to support Python dictionaries interfaces too instead of replacing them?
- Python is not very efficient by design - performance is not a primary goal here.

Sahloul:
@alalek Expanding existed objects classes sounds feasible.

- pyopencv_to does not seem necessary! Somehow I was able to pass the list of dicts to functions like `drawMatches`, and it was just fine.. I cannot tell the reason though!
- `m.__dict__`  and others still require looping through the list, we cannot utilize a C extension such as `DataFrame` to do that..

PR #10164, opened 2017-11-27
 Add ocl accuracy test for a few dnn nets

pengli:
They are alexnet, mobilenet-ssd, resnet50, squeezeNet_v1_1,
yolo and fast_neural_style.

Also fix a layer fusion bug exposed by new ocl accuracy tests.


vpisarev:
@alalek, is it ready to be merged? I'm fine with the patch

PR #10173, opened 2017-11-27
Update PriorBox layer

dkurt:
### This pullrequest changes

* Explicit `width` and `height` attributes of PriorBox layer
* General offsets parameters

**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1468


vpisarev:
:+1:

PR #10176, opened 2017-11-27
Update train_HOG.cpp

sturkmen72:
### This pullrequest changes

Updates train_HOG.cpp

1.adds ability of testing trained detector with webcam
2. adds ability of using positive images when they have not exact same size ( when width and height are specified )
3.adds a new cmd line parameter "-f" indicates if the program will generate and use mirrored samples or not

### DIFF
--- a/samples/cpp/train_HOG.cpp
+++ b/samples/cpp/train_HOG.cpp
@@ -10,14 +10,14 @@ using namespace cv;
 using namespace cv::ml;
 using namespace std;
 
-void get_svm_detector( const Ptr< SVM > & svm, vector< float > & hog_detector );
+vector< float > get_svm_detector( const Ptr< SVM >& svm );
 void convert_to_ml( const std::vector< Mat > & train_samples, Mat& trainData );
 void load_images( const String & dirname, vector< Mat > & img_lst, bool showImages );
 void sample_neg( const vector< Mat > & full_neg_lst, vector< Mat > & neg_lst, const Size & size );
-void computeHOGs( const Size wsize, const vector< Mat > & img_lst, vector< Mat > & gradient_lst );
-int test_trained_detector( String obj_det_filename, String test_dir, String videofilename );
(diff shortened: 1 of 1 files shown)


terfendail:
`computeHOG` use cropping to reduce input images to the size acceptable by `hog.compute`. Location of the region to crop is calculated as absolute value that fit to the center of the first image. If all images are of the same size this region fits all other images as well, but in case image sizes differ estimated region could crop different parts for different images. IMHO such behavior wouldn't be obvious for the user. I think it make sense to evaluate crop region for each image separately and clearly state in the help message the part of the image that will be processed in case detector size differs from the image size.
Also IMHO it make sense to check that all images are greater or at least equal to the size of the detector.

sturkmen72:
@terfendail thank you for your comment. let me try to improve the PR according to your comment when i find a free time.

sturkmen72:
@terfendail i tried to improve the code according to your comment with the commit : update1
if the code looks OK i will add some explanation about usage.

samples/cpp/train_HOG.cpp
```diff
@@ -345,7 +349,7 @@ int main( int argc, char** argv )
         clog << "...[done]" << endl;
 
         labels.clear();
-        labels.assign( pos_lst.size(), +1 );
+        labels.assign( positive_count, +1 );
         labels.insert( labels.end(), neg_lst.size(), -1);
```
terfendail: If there are small images in the negative set they will be discarded during `computeHOG` so the training set will be shorter than `neg_lst.size()`. Could you please use evaluation approach similar to the one in the first training step

terfendail: Other parts of the code looks fine for me. It would be great to describe details of improved sample behavior

terfendail:
Could you please squash commits into one

sturkmen72:
@terfendail could you review updated code and merge if it is OK.
I am postponing adding the description for now. i have still some ideas to improve training process. i want to make some trials later and describe details of improved sample behavior

terfendail:
👍

PR #10180, opened 2017-11-28
OCL: avoid unnecessary loading/initializing OpenCL subsystem

alalek:
If there are no OpenCL/UMat methods calls from application.

OpenCL subsystem is initialized:
- haveOpenCL() is called from application
- useOpenCL() is called from application
- access to OpenCL allocator: UMat is created (empty UMat is ignored) or UMat <-> Mat conversions are called

Don't call OpenCL functions if OPENCV_OPENCL_RUNTIME=disabled
(independent from OpenCL linkage type)

### DIFF
--- a/modules/calib3d/src/stereobm.cpp
+++ b/modules/calib3d/src/stereobm.cpp
@@ -1108,7 +1108,7 @@ public:
         int FILTERED = (params.minDisparity - 1) << disp_shift;
 
 #ifdef HAVE_OPENCL
-        if(ocl::useOpenCL() && disparr.isUMat() && params.textureThreshold == 0)
+        if(ocl::isOpenCLActivated() && disparr.isUMat() && params.textureThreshold == 0)
         {
             UMat left, right;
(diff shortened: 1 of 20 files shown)


vpisarev:
@alalek, looks like there are many changes where good old cv::ocl::useOpenCL() is replaced with cv::ocl::isOpenCLActivated(). Can we somehow provide better backward compatibility, i.e. adjust cv::ocl::useOpenCL() to suit the same purpose as cv::ocl::isOpenCLActivated()?

alalek:
> provide better backward compatibility

According the current semantic of `useOpenCL()` external applications should continue to use `useOpenCL()`.

We just need to replace calls inside library itself only.

The difference is that `isOpenCLActivated()` doesn't try to load/initialize OpenCL runtime and returns 'false' in cases when OpenCL is available/allowed theoretically (but it is not loaded yet because application doesn't try to use UMat or other OpenCL calls - see description).

vpisarev:
:+1:

PR #10184, opened 2017-11-28
Allow to use 3 points in SolvePnP if iterative method and initial guess

catree:
### This pullrequest changes

Relax the constraint about the number of input points for `SolvePnP`: add the possibility to use 3 points when using `SOLVEPNP_ITERATIVE` with an initial estimate ([see](http://answers.opencv.org/question/177753/cvsolvepnp-with-cv_iterative-under-opencv-33/)).


savuor:
:+1:

PR #10188, opened 2017-11-29
my modest contribution, dnn in real time

cabelo:
### This pullrequest changes

My small and modest contribution, dnn in real time, see video bellow..

https://www.youtube.com/watch?v=NHtRlndE2cg


samples/dnn/yolo_object_cam.cpp
```diff
@@ -0,0 +1,184 @@
...
+			const int probability_index = 5;
+			const int probability_size = detectionMat.cols - probability_index;
+			float *prob_array_ptr = &detectionMat.at<float>(i, probability_index);
+
+			size_t objectClass = std::max_element(prob_array_ptr, prob_array_ptr + probability_size) - prob_array_ptr;
+			float confidence = detectionMat.at<float>(i, (int)objectClass + probability_index);
+
+			if (confidence > confidenceThreshold)
+			{
+					string sLabel = returnLabel(objectClass,mylist );
```
alalek: Please replace tabs to 4 spaces.

samples/dnn/yolo_object_cam.cpp
```diff
@@ -0,0 +1,184 @@
...
+					cv::Mat color(roi.size(), CV_8UC3, cv::Scalar(0, 0, 0)); 
+					double alpha = 0.3;
+					cv::addWeighted(color, alpha, roi, 1.0 - alpha , 0.0, roi); 
+		 
+					putText(frame, sLabel, Point(object.x, object.y+12),FONT_HERSHEY_PLAIN, 1 ,Scalar::all(255)); 
+			
+			}
+		}
+		 
+		imshow("GPU", frame); //waitKey();
```
alalek: "GPU" -> "Result" / "YOLO"
(by default GPU acceleration is not enabled)

samples/dnn/yolo_object_cam.cpp
```diff
@@ -0,0 +1,184 @@
...
+		if (cv::waitKey(30) >= 0) grabFrame = false;
```
alalek:
```.diff
-if (cv::waitKey(30) >= 0) grabFrame = false;
+if (cv::waitKey(30) > 0)
+    break;
```

samples/dnn/yolo_object_cam.cpp
```diff
@@ -0,0 +1,184 @@
...
+const char* params
+    = "{ help           | false | print usage         }"
+      "{ cam            |       | device index        }"
+      "{ cfg            |       | model configuration }"
+      "{ model          |       | model weights       }"
+      "{ labels         |       | label of the object }"
+      "{ min_confidence | 0.24  | min confidence      }";
+
+
+void assetRoi(Rect &_roi, Mat &_frame  )
```
alalek: Add "static" for local functions "assetRoi" / "addLabels" / "returnLabel"

vpisarev:
@cabelo, thanks for the contribution!

what's the reason of populating the same code with little changes? Can we instead modify yolo_object_detection.cpp and add extra features there?

samples/dnn/yolo_object_cam.cpp
cabelo: Done

samples/dnn/yolo_object_cam.cpp
cabelo: done

samples/dnn/yolo_object_cam.cpp
cabelo: Done

samples/dnn/yolo_object_cam.cpp
cabelo: Done

cabelo:
screenshot_20171129_211955
Add image example.

alalek:
> perf_video-ippicv-opencl

This is sporadic failure, you need to amend commit and make a force push (or just wait when OpenCV maintainers reschedule these failed builds).

PNG image is too large. Lossless format is not necessary for samples (it is not a strict test), so try to use JPEG format instead.

Please merge/replace your changes into [yolo_object_detection.cpp](https://github.com/opencv/opencv/blob/3.3.1/samples/dnn/yolo_object_detection.cpp) file (we should avoid samples with similar/extended functionality).

vpisarev:
@cabelo, let me close this PR. Open the new one where you modify the existing yolo_object_detection.cpp. Also, we would rather not to add any new images with people into opencv repository, because it may violate legal rules. Let's only leave the live mode of the demo where the data is captured from camera (or let the user specify video file or still image). Thanks for understanding.

cabelo:
@vpisarev  if i crop my face in picture?

Issue #4599, opened 2015-07-27
BORDER_WRAP not available in FilterEngine


opencv-pushbot:
Transferred from http://code.opencv.org/issues/3168
```
|| Tama McGlinn on 2013-07-20 23:44
|| Priority: Low
|| Affected: None
|| Category: imgproc, video
|| Tracker: Feature
|| Difficulty: Easy
|| PR: 
|| Platform: None / None
```

BORDER_WRAP not available in FilterEngine
-----------
```
modules/imgproc/src/filter.cpp:160 - CV_Assert( columnBorderType != BORDER_WRAP );

This line prevents me from making a linear filter with BORDER_WRAP type. I have no idea why this is, nor how to fix it.
```

History
-------
##### Steven Puttemans on 2013-07-21 12:25
```
This is due to no code in the filter algorithm being provided for solving problems with the cv_borderwrap type of borderstyle. To fix it what you can do is create a pull request with the assert gone, then see how it influences the problem and where it goes wrong. But I am guessing some algorithms beneath it will generate errors which need to be fixed then.
```
##### Ivan Korolev on 2013-07-22 10:35
```
Hello Tama McGlinn,
Thank you for submitting this ticket. Unfortunately, FilterEngine doesn't support BORDER_WRAP type of border. If you could add this feature on your side, a pull request in our GitHub repo (http://www.code.opencv.org/projects/opencv/wiki/How_to_contribute for more details) would be highly appreciated! If you decide to continue with introducing this feature then please explain what is purpose of this feature? 
-   Priority changed from Normal to Low
-   Status changed from New to Open
-   Tracker changed from Bug to Feature
```
##### Alexander Smorkalov on 2013-11-27 10:53
```
-   Target version changed from 2.4.7 to Next Hackathon
-   Assignee deleted (Vadim Pisarevsky)
```
##### isscandar isscandar on 2014-09-30 18:10
```
Ivan Korolev wrote:

> If you decide to continue with introducing this feature then please explain what is purpose of this feature?

Hello Ivan,
[...]

PR #10197, opened 2017-11-29
imgproc: allow BORDER_WRAP in filter engine

hannometer:
resolves opencv/opencv#4599

### This pullrequest changes

- allow BORDER_WRAP as border type in FilterEngine
- add tests to make sure it works as expected; tested with GaussianBlur and Sobel

### DIFF
--- a/modules/imgproc/src/filter.cpp
+++ b/modules/imgproc/src/filter.cpp
@@ -124,8 +124,6 @@ void FilterEngine::init( const Ptr<BaseFilter>& _filter2D,
     rowBorderType = _rowBorderType;
     columnBorderType = _columnBorderType;
 
-    CV_Assert( columnBorderType != BORDER_WRAP );
-
     if( isSeparable() )
     {
         CV_Assert( rowFilter && columnFilter );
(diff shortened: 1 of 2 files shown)


vpisarev:
@hannometer, thank you for the patch. I suspect, the patch to enable this functionality is not that simple. There is OpenCL-accelerated branch, there is IPP code; there is parallel implementation. FilterEngine is quite obsolete and will likely be replaced with something else at some point. Patching just it in a hope that other branches will not be executed (whereas the opposite thing happens all the time) is not enough, of course. Can you provide much more elaborated test (or expand existing tests) to check other functions, other resolutions, GPU acceleration etc. We also need to check the whole result image, not just 2 elements of it.

hannometer:
Ok, I understand the problems... I'll try to find the time to write some more exhaustive tests, also for the OpenCL/IPP/parallel branches - or disable BORDER_WRAP just for them. I just think it's a shame that BORDER_WRAP for filtering is categorically denied although it works in the "normal" (i.e. non-OpenCL/IPP/...) setting. Also, because there's no mention of it in the documentation.

vpisarev:
ok, let's close it for now. We need some more complete solution, covering CPU, including parallel implementations, and GPU

PR #10212, opened 2017-12-01
OCL: fix usage of invalid OpenCL cache on mixed 64/32-bit platforms

alalek:
Observed during launch of 32/64-bit applications on Windows.
Added '32-bit' prefix for 32-bit OpenCL devices. No prefix on 64-bit configurations.

```
```


alalek:
:+1:

Issue #10128, opened 2017-11-21
dnn::readNetFromTorch some layers not implemented 


bill2239:
##### System information (version)
<
- OpenCV => 3.3.1
- Operating System / Platform => Windows 64 Bit
- Compiler => Visual Studio 2015
>

##### Detailed description
I try to run use dnn:Net module to load torch model, and it get error in this line 
dnn::Net net = dnn::readNetFromTorch(modelfile);
error
OpenCV Error: The function/feature is not implemented (Unknown nn class "nn.BatchNormalization") in cv::dnn::experimental_dnn_v2::TorchImporter::readTorchObject,
##### Steps to reproduce
dnn::Net net = dnn::readNetFromTorch(modelfile);
I did convert the net to cpumode by net= net:float(), it seem like some layers are not implemented 
nn.Dropout also have the same error.

dkurt:
@bill2239, please provide a minimal working example: a whole model definition or a single batch normalization layer test case.

bill2239:
okay i see
model definition
local function ConvBNReLU(nInputPlane, nOutputPlane)
      vgg:add(backend.SpatialConvolution(nInputPlane, nOutputPlane, 3,3, 1,1, 1,1))
      vgg:add(nn.SpatialBatchNormalization(nOutputPlane,1e-3))
      vgg:add(backend.ReLU(true))
      return vgg
    end
    ConvBNReLU(1,64)--:add(nn.Dropout(0.3,nil,true))
    ConvBNReLU(64,64)
    vgg:add(MaxPooling(2,2,2,2):ceil())

    ConvBNReLU(64,128)--:add(nn.Dropout(0.4,nil,true))
    ConvBNReLU(128,128)
    vgg:add(MaxPooling(2,2,2,2):ceil())

    ConvBNReLU(128,256)--:add(nn.Dropout(0.4,nil,true))
    ConvBNReLU(256,256)--:add(nn.Dropout(0.4,nil,true))
    ConvBNReLU(256,256)
    vgg:add(MaxPooling(2,2,2,2):ceil())

    ConvBNReLU(256,512)--:add(nn.Dropout(0.4,nil,true))
    ConvBNReLU(512,512)--:add(nn.Dropout(0.4,nil,true))
    ConvBNReLU(512,512)
    vgg:add(MaxPooling(2,2,2,2):ceil())

    -- In the last block of convolutions the inputs are smaller than
    -- the kernels and cudnn doesn't handle that, have to use cunn
[...]

PR #10221, opened 2017-12-04
nn.BatchNormalization and nn.Dropout layers from Torch

dkurt:
### This pullrequest changes

Non-spatial versions of batch normalization and dropout layers from Torch: `nn.BatchNormalization` and `nn.Dropout`.

resolves https://github.com/opencv/opencv/issues/10128

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/413


vpisarev:
:+1:

PR #10222, opened 2017-12-04
Fix pointPolygonTest for large coordinate values

reunanen:
Problem: `pointPolygonTest` gives incorrect results for large coordinate values (because multiplication overflows `int`).

Solution: use a wider type (`long long`).


vpisarev:
@reunanen, thanks!

The patch looks reasonable, but please fix the compile errors. Essentially, we did not ratify the use of C++ 11 in OpenCV yet. Please, specify the types explicitly. Also, people using older versions of Visual Studio may complain about "long long" use. Can you replace it with OpenCV-defined type "int64"?

modules/imgproc/test/test_pointpolygontest.cpp
```diff
@@ -0,0 +1,97 @@
...
+Point2f generate_point()
+{
+    return Point2f(40000, 40000);
+}
+
+template <typename T> bool CV_PointPolygonTestTest::require_is_inside()
+{
+    const auto contour = generate_contour<T>();
+    const auto point = generate_point();
+    const auto result = cv::pointPolygonTest(contour, point, false);
```
alalek: 1. Please replace 'auto' (current version of OpenCV should be compiled with C++98 compiler).
2. Please use native Google tests without legacy `cvtest::ArrayTest`:
```
TEST(Imgproc_PointPolygonTest, accuracy)
{
    Point contour[] = { Point(0, 0), ... };
    Point2f point = Point2f(40000, 40000);
    bool result = cv::pointPolygonTest(contour, point, false);
    EXPECT_TRUE(result) <<  "Desired result: point is inside polygon - actual result: point is not inside polygon";
}
```
3. No need to create separate file for this test, put in near existed tests - "test_contours.cpp".

vpisarev:
:+1:

Issue #10214, opened 2017-12-02
createsamples info message


DKurilo:
##### System information (version)
OpenCV: commit 7ae19467b57a19facc724444ac2d1d6cd609c726 and current

##### Detailed description

In CreateTestSample mode:
Now verbose message: "Open background image" points on wrong image.

Can you move:
```
#ifdef CV_VERBOSE
            printf( "Open background image: %s\n", data->filename[data->last] );
#endif /* CV_VERBOSE */
```
after 
```
            data->last %= data->count;
```
please?

It's difficult to find wrong negative image now.

vermashresth:
@DKurilo @alalek I am a beginner and would like to work on this isuue. Is the bug in utilty.cpp file for the createsamples folder in apps?

berak:
@vermashresth , that's a nice idea !

and yes, it's in [utility.cpp, here](https://github.com/opencv/opencv/blob/master/apps/createsamples/utility.cpp#L998-L1003)

PR #10226, opened 2017-12-04
fixed createsample info message 

vermashresth:
resolves https://github.com/opencv/opencv/issues/10214

Changed the verbose call to be after
data->last %= data->count;


apps/createsamples/utility.cpp
```diff
@@ -995,12 +995,13 @@ void icvGetNextFromBackgroundData( CvBackgroundData* data,
         {
             round = data->round;
 
+            data->last = rand() % data->count;
+            data->last %= data->count;
```
terfendail: IMHO the second line do nothing since the result of the first line is already a remainder. I think it would be better to remove it.

vermashresth: @terfendail  Sure, I will remove the second line.

vpisarev:
:+1:

PR #10228, opened 2017-12-05
dnn layers ocl implementation

pengli:
enable ocl path for layers including : detection_output, region, reorg, reshape and permute.


modules/dnn/src/layers/detection_output_layer.cpp
```diff
@@ -211,11 +212,154 @@ class DetectionOutputLayerImpl : public DetectionOutputLayer
...
+        if (numKept == 0)
+        {
+            // Set confidences to zeros.
+            Range ranges[] = {Range::all(), Range::all(), Range::all(), Range(2, 3)};
+            outputs[0](ranges).setTo(0);
+            return true;
+        }
+        int outputShape[] = {1, 1, (int)numKept, 7};
+        UMat umat = UMat(4, outputShape, CV_32F);
+        float* outputsData = umat.getMat(ACCESS_READ).ptr<float>();
```
alalek: This line is not correct:
- umat.getMat() maps OpenCL buffer into `cv::Mat`
- get ptr from `cv::Mat`
- destroy temporary `cv::Mat` and unmap all data back to OpenCL buffer
- pointer is invalid and can't be used.

ACCESS_READ => WRITE

modules/dnn/src/layers/detection_output_layer.cpp
```diff
@@ -211,11 +212,154 @@ class DetectionOutputLayerImpl : public DetectionOutputLayer
...
+            kernel.set(8, (int)clip);
+            kernel.set(9, ocl::KernelArg::PtrWriteOnly(outmat));
+
+            if (!kernel.run(1, &nthreads, NULL, false))
+                return false;
+        }
+
+        all_decode_bboxes.clear();
+        all_decode_bboxes.resize(num);
+        const float* decode_data = outmat.getMat(ACCESS_READ).ptr<float>();
```
alalek: This line is not correct:
- umat.getMat() maps OpenCL buffer into `cv::Mat`
- get ptr from `cv::Mat`
- destroy temporary `cv::Mat` and unmap all data back to OpenCL buffer
- pointer is invalid and can't be used.

modules/dnn/src/layers/region_layer.cpp
```diff
@@ -114,11 +115,82 @@ class RegionLayerImpl : public RegionLayer
...
+                softmax_kernel.set(7, (int)cols);
+                softmax_kernel.set(8, (int)anchors);
+                softmax_kernel.set(9, (float)thresh);
+                softmax_kernel.set(10, ocl::KernelArg::PtrWriteOnly(outBlob));
+                if (!softmax_kernel.run(1, &nthreads, NULL, false))
+                    return false;
+            }
+
+            if (nmsThreshold > 0) {
+                float *dstData = outBlob.getMat(ACCESS_WRITE).ptr<float>();
```
alalek: Same problem

modules/dnn/src/layers/reshape_layer.cpp
```diff
@@ -182,11 +182,39 @@ class ReshapeLayerImpl : public ReshapeLayer
...
+        for (size_t i = 0; i < inputs.size(); i++)
+        {
+            UMat srcBlob = inputs[i];
+            void *src_handle = inputs[i].handle(ACCESS_READ);
+            void *dst_handle = outputs[i].handle(ACCESS_WRITE);
+            if (src_handle != dst_handle)
+            {
+                MatShape outShape = shape(outputs[i]);
+                UMat umat = srcBlob.reshape(1, (int)outShape.size(), &outShape[0]);
+                umat.copyTo(outputs[i]);
```
alalek: `.assign()`?

modules/dnn/src/opencl/detection_output.cl
```diff
@@ -0,0 +1,181 @@
...
+    for (int index = get_global_id(0); index < nthreads; index += get_global_size(0))
+    {
+        Dtype bbox_xmin, bbox_ymin, bbox_xmax, bbox_ymax;
+        const int i = index % 4;
+        const int p = ((index / 4 / num_loc_classes) % num_priors) * 4;
+
+        const int c = (index / 4) % num_loc_classes;
+        int label = share_location ? -1 : c;
+        if (label == background_label_id)
+            return; // Ignore background class.
```
alalek: continue?

modules/dnn/src/opencl/detection_output.cl
```diff
@@ -0,0 +1,181 @@
...
```
alalek: continue?

modules/dnn/src/layers/detection_output_layer.cpp
```diff
@@ -211,11 +212,154 @@ class DetectionOutputLayerImpl : public DetectionOutputLayer
...
+    {
+        int shape[] = { numClasses, numPredsPerClass };
+        for (int i = 0; i < num; i++)
+            confPreds.push_back(Mat(2, shape, CV_32F));
+
+        UMat umat = inp1.reshape(1, num * numPredsPerClass);
+        for (int i = 0; i < num; ++i)
+        {
+            Range ranges[] = { Range(i * numPredsPerClass, (i + 1) * numPredsPerClass), Range::all() };
+            umat(ranges).t().copyTo(confPreds[i]);
```
alalek: It is better to use `cv::transpose()` directly.

modules/dnn/src/layers/detection_output_layer.cpp
pengli: +        Mat mat = outmat.getMat(ACCESS_READ);
+        const float* decode_data = mat.ptr<float>();

split this code into two line
first to call getMat to maps OpenCL buffer into cv::Mat
second to get the ptr from Mat

modules/dnn/src/opencl/detection_output.cl
pengli: it should be "return"  here,   the "nthreads"  in the loop is less than or equal to the step "get_global_size(0)",  so it actually only loop once in the kernel. nthreads is the total amounts of loc predictions we should compute.  so it just return if "label == background_label_id"

for (int index = get_global_id(0); index < nthreads; index += get_global_size(0))
{
    ...
}

modules/dnn/src/layers/detection_output_layer.cpp
pengli: code is updated  to use tranpose() here.

modules/dnn/src/layers/reshape_layer.cpp
pengli: Do you mean assign the "umat" to OutputArrayOfArrays  ?  here OutputArrayOfArrays is actually a vector of umat, so I can't assign one umat to a vector.  It would be convenient if OutputArrayOfArrays  provides an interface like "assign(UMat umat, int index)" to assign one umat into a vector with  a index.

modules/dnn/src/layers/detection_output_layer.cpp
alalek: You need to create code scope for mapped map, to sync data back before populating of UMat's outputs.
```
{
    Mat mat = umat.getMat(ACCESS_WRITE);
    ... work with mat (umat is blocked)
} // sync data mat->umat, unlock umat
... continue to work with umat ...
```

modules/dnn/src/layers/reshape_layer.cpp
alalek: There are `.assign()` [overloads](https://github.com/opencv/opencv/pull/10126/files) with `std::vector<UMat>`.

modules/dnn/src/layers/detection_output_layer.cpp
pengli: Thanks for the review,  have updated the code with this comments.

modules/dnn/src/layers/reshape_layer.cpp
pengli: added  .assign() to overload std::vector<UMat>

modules/dnn/src/layers/detection_output_layer.cpp
```diff
@@ -211,11 +212,160 @@ class DetectionOutputLayerImpl : public DetectionOutputLayer
...
+            {
+                count += outputDetections_(i, &outputsData[count * 7],
+                                           allDecodedBBoxes[i], allConfidenceScores[i],
+                                           allIndices[i]);
+            }
+            CV_Assert(count == numKept);
+        }
+        std::vector<UMat> & outputvec = *(std::vector<UMat> *)outs.getObj();
+        outputvec.clear();
+        outputvec.push_back(umat);
```
alalek:
```
outputs.clear();
outputs.push_back(umat);
outs.assign(outputs);
```

pengli: yeah，this is better，code is updated

PR #10232, opened 2017-12-05
Hough many circles

### LINKED ISSUE #10227: Hough Circles performance finding many small circles
- OpenCV => 3.3.1 (latest main)

##### Detailed description
HoughCircles performance is very slow finding circles in an image with many small circles. Our application needs to find approximately 9,000 to 10,000 circles in an image. 

The primary problem is the EstimateRadius code does a linear search for all non-zero edge points (the `nz` list) within range of each center point. See: [hough.cpp#L1220](https://github.com/opencv/opencv/blob/a3ec2ac3c5ae04da704b377d58527dacdfeb2926/modules/imgproc/src/hough.cpp#L1220). This can be optimized by changing the `nz` list to a matrix. The matrix is the same size as the image, initially set to zeros. An element is set to 1 if there is a non-zero edge at that position. Finding non-zero edge points within range of a center point requires iterating over a range of elements +/- max_radius from the center point. Because the matrix approach may be optimal only for small circles, I wrapped it in a class that hides the implementation details, and provided a vector-based implementation with the same API. This makes it easy to switch data structures. It would be good to have some input on what would work best for other developers. 

A second performance problem is the code for checking the minimum distance between circles. It is in a ParallelLoopBody but the distance check requires exclusive access to the common `circles` list. It's a little more complicated than that because the distance check is performed in two places, but basically the code was spending a lot of time waiting on the mutex and not really running in parallel. The second performance problem can be fixed by performing the distance check after the parallel EstimateRadius loop. The approach I am using is single-threaded. It compacts the `circles` list in place. 

Another issue with the minimum distance check is that which of the two circles was preserved depended on order of evaluation, and was essentially random. I changed the EstimateRadius output format to include a `Vec3f` for the circle and an `int` for the accumulator value. The circles are sorted with the highest accumulator values first. This way the minimum distance check preserves the circles with higher accumulator values. 

The HoughCircles API documentation says "Circles, corresponding to the larger accumulator values, will be returned first." But the code has not worked that way for a while. Now it will. 

The `cvHoughCircles` wrapper function specified a `circles_max` limit for the size of the output array. The EstimateRadius loop was checking if the max number of circles had been reached. That seemed like unnecessary overhead in the common case when `circles_max` is `INT_MAX`. And it conflicted with the requirement to return the best circles first. I changed the code so it applies the `circles_max` limit at the end, after sorting the circles and the minimum distance check, when copying the circles to the output array. 

There is a what looks like a typo at [hough.cpp#L1441](https://github.com/opencv/opencv/blob/a3ec2ac3c5ae04da704b377d58527dacdfeb2926/modules/imgproc/src/hough.cpp#L1441). The code returns a list of centers only, if `maxCircles == 0`. The HoughCircles API documentation says "you may set maxRadius to 0 to return centers only without radius search". I'm pretty sure the code meant to check `maxRadius` and that `maxCircles` is a typo. However, there is [other code](https://github.com/opencv/opencv/blob/a3ec2ac3c5ae04da704b377d58527dacdfeb2926/modules/imgproc/src/hough.cpp#L1507) that sets `maxRadius` to a default if it is `<= 0`. I have not tried to fix it. It would be simple to change, but there could be existing code that depends on the current behavior, even though the current behavior is not as documented. Maybe the best thing would be to update the API documentation. 

The proposed fixes are in [hough.cpp](https://github.com/TomBecker-BD/opencv/blob/hough-many-circles/modules/imgproc/src/hough.cpp) on a branch. If and when they seem okay, I can submit a PR. 

##### Steps to reproduce
I didn't see any unit tests for HoughCircles, so I added some. [test_houghCircles.cpp](https://github.com/TomBecker-BD/opencv/blob/ff449267713d4d995a9c704378edd427187c5e96/modules/imgproc/test/test_houghCircles.cpp#L151)

The `regression` test finds circles in the "stuff.jpg" image. 

The `MaxRadiusZeroCentersOnly` test demonstrates the bug where setting `maxRadius` to 0 does not stop the code from doing the radius search. 

The `Beads` test is for an image with many small circles. The test image is 1/4 size of the actual images used for our application. 

I don't have a unit test for the bug where the circles are not returned in order with higher accumulator values first. 

There should also be a performance test, but I haven't written one yet. 

Test data is in on a branch in [testdata/cv/imgproc](https://github.com/TomBecker-BD/opencv_extra/tree/TestHoughManyCircles/testdata/cv/imgproc).

TomBecker-BD:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/418

resolves #10227

### This pullrequest changes

1. Optimize for finding many small circles in an image. Uses a matrix of flags to hold the non-zero edge positions. If the number of non-zero edges is small it switches to a vector. 

2. Improve performance checking the minimum distance between circles. Move the code out of the EstimateRadius parallel loop body so there is not lock contention. 

3. Sort the circles so the circles with higher accumulator value are first. The API documentation says that it did this, but the code was not there. 

4. The maxCircles limit is applied only when copying the circles to the output array. Otherwise circles with higher accumulator values could be excluded. 

5. Set maxRadius < 0 to return the center points without the radius search. Preserves the existing behavior where setting maxRadius = 0 defaults maxRadius to the image size. 

6. Add unit and performance tests. 

Thanks to Hui Liu for the original performance optimization. 

```
docker_image:Docs=docs-js
```


modules/imgproc/src/hough.cpp
```diff
@@ -1212,15 +1455,14 @@ class HoughCircleEstimateRadiusInvoker : public ParallelLoopBody
...
                         }
+
+                        nnz = 0;
                     }
                 }
-#endif
 
-                // Estimate best radius
-                for(; j < nzSz; ++j)
+                for(int iz = 0; iz < nnz; ++iz)
```
alalek: This loop is a "tail processing" of SIMD optimized code. It can't start from zero again.

SIMD loop above is broken too. Also SIMD code is totally optional and can be excluded from sources.

savuor:
As I said in issue description, the typo was in the documentation, this is the `maxCircles` argument which signalizes that user doesn't ask for radiuses.
Therefore the condition `if(maxCIrcles == 0)` is correct and should be left with one more call to `RemoveOverlaps()`

modules/imgproc/perf/perf_houghCircles.cpp
```diff
@@ -0,0 +1,60 @@
+#include "perf_precomp.hpp"
+#include "opencv2/imgproc.hpp"
+#include "opencv2/imgproc/types_c.h"
+
+using namespace std;
+using namespace cv;
+using namespace perf;
+
+PERF_TEST(Basic, PerfHoughCircles)
```
savuor: `PerfHoughCircles` should be the first argument, `Basic` should be the second one like that:
`PERF_TEST(test_case, specific_test)`

modules/imgproc/perf/perf_houghCircles.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+    TEST_CYCLE()
+    {
+        HoughCircles(img, circles, CV_HOUGH_GRADIENT, dp, minDist, edgeThreshold, accumThreshold, minRadius, maxRadius);
+    }
+
+    SANITY_CHECK_NOTHING();
+}
+
+PERF_TEST(ManySmallCircles, PerfHoughCircles)
```
savuor: The same as above

savuor:
I made performance measurements and found out that (it was easy to predict though) using NZPointList is faster for small circles search but slower for big range of radiuses (for example like [that](https://user-images.githubusercontent.com/10323511/32551991-7c147c4c-c4a3-11e7-8572-300c63a7172e.jpg))
My suggestions are the following:
1. When iterating inside NZPointSet use both `minRadius` and `maxRadius` to skip inner radiuses too. This will make calculations faster also for cases when `minRadius` is quite big. In that case inner rectangle will be `sqrt(2)` times smaller than `minRadius`.
2. We can make a decision about what to use, `NZPointSet` or `NZPointList` at runtime. Since we reduce amount of iterations per circle center we just need to calculate the following comparison:
`(maxRadius*maxRadius - minRadius*minRadius) < nz.size()`
Based on what is smaller we decide what method to use. To be able to do that we need to keep both structures, which means that during accumulator build NZ we should fill both Set and List parts with the same pooints.

modules/imgproc/perf/perf_houghCircles.cpp
TomBecker-BD: I thought that was the right way to do it, but I got this error when I tried to run the tests: 
> Failed
> All tests in the same test case must use the same test fixture
> class.  However, in test case PerfHoughCircles,
> you defined test Basic and test ManySmallCircles
> using two different test fixture classes.  This can happen if
> the two classes are from different namespaces or translation
> units and have the same name.  You should probably rename one
> of the classes to put the tests into different test cases.

Maybe the problem is because I am using the `PERF_TEST` macro. I will look into it further.

modules/imgproc/src/hough.cpp
TomBecker-BD: Correct me if I'm wrong. I had to study the code to figure out how it worked. The previous code was loading data directly from the `nz` vector. It looped in increments of 4 for SIMD processing. This would leave a remainder of 0 to 3 points in `nz` that still needed to be processed. This is conveniently done by falling through to the non-SIMD loop. 

The new code can't do that, because `nz` is a container that is accessed only through a forward iterator. If the container is matrix-based, it contains flags, not points. The first loop fills the `nzx` and `nzy` arrays. When `nnz == 4` the arrays are loaded into vector registers and processed. When the iterator reaches the end, there will be 0 to 3 remaining points in `nzx` and `nzy`. The second loop is there to process them. The loop starts from 0 because it is an index into the `nzx` and `nzy` arrays. 

I also added an `else` to prevent the SIMD code from falling through into the non-SIMD loop. 

There still is a non-SIMD loop for the case where SIMD is disabled either at compile time or run time. To be honest, I have only been testing the SIMD code. The non-SIMD code should work, but it needs to be tested. 

I don't know why you are saying the SIMD loop is broken. The design was changed to work with a forward iterator instead of a random-addressable vector. It still produces the same results. Maybe there is something I did wrong or could do better.

TomBecker-BD:
Unfortunately, there does not seem to be a way the user can use the `maxCircles` argument to ask for no radius search. The `maxCircles` argument is not in the external API. 

The exported `cv::HoughCircles` function does not have a `maxCircles` argument. It calls the internal `HoughCircles` function with a hard-coded `-1` for `maxCircles`. 

The `cvHoughCircles` wrapper does not have a `maxCircles` argument. It calls the internal `HoughCircles` function with either `INT_MAX` or `circles->total` for `maxCircles`. 

The only way it might be possible to get a `maxCircles` value of 0 is by calling the `cvHoughCircles` wrapper with a zero-length `circle_storage`. But then there is no room to return any center coordinates. 

If the functionality is important, it could be implemented by adding a new function argument where it can be set by the user. I looked for any issues requesting that it be fixed, but I could not find any.

TomBecker-BD:
Good idea about skipping the pixels inside `minRadius / sqrt(2)`. 

If you want to choose between `NZPointList` or `NZPointSet` dynamically, it would be easy to copy from one data structure to the other. I would lean towards using `NZPointSet` for the accumulator build because the memory is pre-allocated, and because it automatically eliminates duplicates. I would move the code that gets `int nzSz = int(nz.size());` into the main function. That way there is no added overhead if the code decides to stay with the same data structure, and the cost of a single pass over the matrix if it decides to switch. The `HoughCircleEstimateRadiusInvoker` can get the `NZPoints` type as a template parameter.

modules/imgproc/perf/perf_houghCircles.cpp
TomBecker-BD: I was assuming `PERF_TEST` works like the googletest `TEST` macro. Unfortunately it only allows one test per test case. I tried `PERF_TEST_F` and it had the same limitation. I worked around it by changin the test case names to "PerfHoughCircles" and "PerfHoughCircles2".

savuor:
Sorry, I forgot that `maxCircles` is not a part of API.
There should be another way to turn off radius search through API.
Remember that `maxRadius=0` is a default argument with which the algorithm shouldn't require any post-processing to get circles.
That's why let's use negative values of `maxRadius` to turn off radius search. When building accumulator we need to use image size instead of `maxRadius`.

TomBecker-BD:
Thanks for the feedback. I am going to work on: 

1. Fixing build errors and warnings. 
2. Skip radius search if `maxRadius < 0`. 
3. Dynamically switch to `NZPointList` if it will be faster than `NZPointSet`.

TomBecker-BD:
There are two build issues: 
1. The "ManySmallCircles" unit and performance tests fail on the Mac build with an error -11. That is probably a FAIL_MISMATCH error. I don't know why the tests are passing on other platforms and failing only on the Mac build. There are no messages about unmet expectations. The tests that are failing are almost identical to other tests that pass on the Mac build. It's not a timeout because the performance test fails in less than a second. There are no exception messages or stack traces. 
2. The docs build has a warning "opencv_extra: 2263 KiB". This is probably because I added files. 

I am going to remove the "ManySmallCircles" tests. That should get the build to pass. It is disappointing because these tests demonstrate the performance issue that this PR fixes. I am hoping that maybe I (or someone else) can get them working on all platforms and then they can be added back.

TomBecker-BD:
There is one remaining build warning, on the docs branch: patch size opencv_extra opencv_extra: 2264 KiB

However, there are only two files added on the branch: 

- HoughCircles.xml 298 Bytes
- stuff.jpg 28.7 KB

Previously there was also a beads.png file for the "ManySmallCircles" tests. I tried replacing it with a smaller .jpg version. Then I gave up and deleted it. 

The patch size reported by the build bot looks like it is including the sizes of the files that were added and then deleted. If the branch is squash merged the patch size will be much smaller.

alalek:
Please create PR into opencv_extra (it is not necessary for testing, but necessary to review and merge).

TomBecker-BD:
opencv_extra [#418](https://github.com/opencv/opencv_extra/pull/418)

modules/imgproc/test/test_houghCircles.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+#if DEBUG_IMAGES
+    string imgProc = string(cvtest::TS::ptr()->get_data_path()) + "imgproc/";
+    highlightCircles(filename, circles, imgProc + "HoughCirclesTest_CentersOnly.png");
+#endif
+
+    EXPECT_GT(circles.size(), size_t(0)) << "Should find at least some circles";
+    for (size_t i = 0; i < circles.size(); ++i)
+    {
+        EXPECT_EQ(circles[0][2], 0.0f) << "Did not ask for radius";
```
savuor: did you forget `i` index?

modules/imgproc/test/test_houghCircles.cpp
```diff
@@ -0,0 +1,233 @@
...
+    highlightCircles(filename, circles, imgProc + "HoughCirclesTest_DefaultMaxRadius.png");
+#endif
+
+    int maxDimension = std::max(src.rows, src.cols);
+
+    EXPECT_GT(circles.size(), size_t(0)) << "Should find at least some circles";
+    for (size_t i = 0; i < circles.size(); ++i)
+    {
+        EXPECT_GE(circles[0][2], minRadius) << "Radius should be >= minRadius";
+        EXPECT_LE(circles[0][2], maxDimension) << "Radius should be <= max image dimension";
```
savuor: `i` index seems to be forgotten here

TomBecker-BD: Thanks for catching that. I just pushed a fix.

modules/imgproc/src/hough.cpp
savuor: The SIMD code is correct (even it has some non-evident logic) but it's arguable whether it gives us additional performance. I think you can freely exclude it from source.

savuor:
And by the way, have you found out what was wrong on Mac build?
Maybe removing that SIMD part will work.

TomBecker-BD:
I don't know why it was failing on the Mac. It was failing only in the "ManySmallCircles" test. I'm pretty sure the SIMD path is executed on the passing tests. It could be a compilation issue with NZPointSet, but that code was well exercised on other platforms. It could have been a glitch in pulling "Beads.png" or in loading it, due to the size of the image. But that's just guessing - I don't know. I removed the failing test as a workaround. I'm not happy about that, but it was the only way I could think of to stop the build from failing. 

I have a Mac at home I could use to debug the failure. If I can find the problem and fix it, I will submit a new PR with the "ManySmallCircles" test added back. But it's going to be a while before I have the time to work on it at home.

TomBecker-BD:
Should have looked at the build results. The Mac test timed out in `ImgProc/HoughCirclesTestFixture.regression/0, where GetParam() = ("imgproc/stuff.jpg", 20, 20, 30, 20, 200)`.

savuor:
There could be build system problems with Mac build, I just re-ran tests and everything works fine (except patch size for extra which we've fixed once and this wil be no problem).
Sure, you can add the test for ManySmallCircles later as a separate PR.

TomBecker-BD:
OK, thanks.

modules/imgproc/test/test_houghCircles.cpp
```diff
@@ -0,0 +1,233 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
alalek: Please avoid using of Capitalized letters in file names: test_houghCircles.cpp

savuor:
:+1:

alalek:
> Mac build with an error -11

This means that code tries to read memory out of buffer. This usually leads to crashes like this or non-predictable results of this algorithm or other algorithms because their memory was corrupted (worst case because these things very hard to debug).

Does this problem really get fixed (or failed tests are just disabled)?

TomBecker-BD:
I tried converting all the pointer references to go through `at` so access is range-checked in the debug build, and added range checking code around other arrays. So far I have not found any buffer errors. 

I am doing a Mac build now. I want to try running with AddressSanitizer.

TomBecker-BD:
I found a bug that was causing an EXC_BAD_ACCESS exception on the Mac. The cmpAccum function was not strictly weak ordered and it caused std::sort to go out of bounds. It should be safe to add back the ManySmallCircles tests. Unless the "beads.jpg" file is too big. It's 668 KB.

TomBecker-BD:
I ran the Hough Circles tests under Address Sanitizer, Thread Sanitizer, Undefined Behavior Sanitizer, Malloc Scribble, Malloc Guard Edges, and Guard Malloc. Thread Sanitizer reported a data race in ipp_gaussianBlurParallel. Undefined Behavior Sanitizer reported warnings about bit shifting in libjpeg. Otherwise no issues.

savuor:
@TomBecker-BD Thank you for a huge amount of your work!
I think now we can merge it

savuor:
:+1:

alalek:
@TomBecker-BD I pushed additional patches for this PR (simplify code, SIMD intrinsics).
Please take a look.

TomBecker-BD:
The new code looks good. It's faster. The results are not identical, but they look okay.

alalek:
> The results are not identical

There is an issue in NZPointSet iterator. It lost synchronization between "current" and "mi" - it returns points with "zero" value sometimes:
```
++current.x;
if ((current.y >= yInner.start && current.y < yInner.end) &&
    (current.x >= xInner.start && current.x < xInner.end))
{
    current.x = xInner.end;
    // FIXIT lost proper "mi" update like
    // mi = positions.ptr<uchar>(current.y, current.x - 1) // -1 see below: "++mi;"
}
if (current.x < xOuter.end)
{
    ++mi;
}
else
...
```

TomBecker-BD:
Good. That explains it. In retrospect excluding the inner range was too much complexity. The iterator needed its own suite of unit tests.

PR #10258, opened 2017-12-07
kmeans: number of channels in _centers fixed

### LINKED ISSUE #10174: samples/cpp/kmeans.cpp: Invalid layout of "center"
Sample code expects `CV_32FC2` type, but it is CV_32FC1 with 2 cols.

Reproduced in Debug build only (catched by `CV_DbgAssert()`).

```
        for (i = 0; i < centers.rows; ++i)
        {
            Point2f c = centers.at<Point2f>(i);     // <=== problem is here
            circle( img, c, 40, colorTab[i], 1, LINE_AA );
        }
```

Changing declaration of centers:
```.diff
-Mat centers;
+std::vector<Point2f> centers;
```
doesn't work (exception during `kmeans()` call).
Perhaps we should support this implicit conversion (via reshape).

savuor:
resolves #10174

### DIFF
--- a/modules/core/src/kmeans.cpp
+++ b/modules/core/src/kmeans.cpp
@@ -458,7 +458,14 @@ double cv::kmeans( InputArray _data, int K,
         {
             best_compactness = compactness;
             if( _centers.needed() )
-                centers.copyTo(_centers);
+            {
+                Mat reshaped = centers;
+                if(_centers.rows() == 1 && _centers.channels() == dims)
+                    reshaped = centers.reshape(dims, 1);
+                else if(_centers.cols()*_centers.channels() == dims)
+                    reshaped = centers.reshape(_centers.channels(), K);
+                reshaped.copyTo(_centers);
+            }
             _labels.copyTo(best_labels);
         }
     }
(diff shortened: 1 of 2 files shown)


modules/core/src/kmeans.cpp
```diff
@@ -458,7 +458,11 @@ double cv::kmeans( InputArray _data, int K,
         {
             best_compactness = compactness;
             if( _centers.needed() )
+            {
+                if(_centers.channels() == dims)
```
terfendail: It looks like the function allow to pass complex configuration of input points since in L239 `int dims = (isrow ? 1 : data0.cols)*data0.channels()`. IMHO it should be possible for the user to provide the same configuration for the `OutputArray _centers`

savuor: This makes sense, will be implemented

modules/core/src/kmeans.cpp
```diff
@@ -458,7 +458,14 @@ double cv::kmeans( InputArray _data, int K,
         {
             best_compactness = compactness;
             if( _centers.needed() )
-                centers.copyTo(_centers);
+            {
+                Mat reshaped = centers;
+                if(_centers.rows() == 1 && _centers.channels() == dims)
```
vpisarev: the fix is incorrect. The function behavior should not be dependent on the current shape of the output array. Instead, you should check whether the output array (`_centers`) has the fixed type (e.g. when `_centers` is `std::vector<someT>`) or not.

savuor: Done.
But we need to check actual `channels()` number (which shouldn't depend on current shape of array).

PR #10260, opened 2017-12-07
FFMPEG isn't loaded from the package directory sometimes

native-api:
As per https://msdn.microsoft.com/en-us/library/windows/desktop/ms683197(v=vs.85).aspx , `GetModuleFileName` isn't required to reset `GetLastError` on success, the return value alone is the error indicator.

I actually hit this in XP where the value still was the `ERROR_MOD_NOT_FOUND` from failed `PATH` search.

### DIFF
--- a/modules/videoio/src/cap_ffmpeg.cpp
+++ b/modules/videoio/src/cap_ffmpeg.cpp
@@ -124,8 +124,10 @@ private:
             if (m)
             {
                 wchar_t path[MAX_PATH];
-                size_t sz = GetModuleFileNameW(m, path, sizeof(path));
-                if (sz > 0 && ERROR_SUCCESS == GetLastError())
+                const size_t path_size = sizeof(path)/sizeof(*path);
+                size_t sz = GetModuleFileNameW(m, path, path_size);
+                /* Don't handle paths longer than MAX_PATH until that becomes a real issue */
+                if (sz > 0 && sz < path_size)
                 {
                     wchar_t* s = wcsrchr(path, L'\\');
                     if (s)


modules/videoio/src/cap_ffmpeg.cpp
```diff
@@ -125,7 +125,7 @@ class icvInitFFMPEG
             {
                 wchar_t path[MAX_PATH];
                 size_t sz = GetModuleFileNameW(m, path, sizeof(path));
-                if (sz > 0 && ERROR_SUCCESS == GetLastError())
+                if (sz > 0)
```
alalek: [MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683197(v=vs.85).aspx) states this:
> If the buffer is too small to hold the module name, the string is truncated to nSize characters including the terminating null character, the function returns nSize, and the function sets the last error to ERROR_INSUFFICIENT_BUFFER.

Perhaps check should be:
```
if (sz > 0 && sz < MAX_PATH)
```

but there are additional notes (including Windows XP) ...

So could you try to force reset of error code before function call in your configuration:
```
SetLastError(ERROR_SUCCESS);
```

BTW, Line above is buggy too:
```.diff
-size_t sz = GetModuleFileNameW(m, path, sizeof(path));
+size_t sz = GetModuleFileNameW(m, path, MAX_PATH);
```

because `sizeof(path)` is in bytes.

Also there is other important thing for Windows XP:
> Windows XP:  The string is truncated to nSize characters and is not null-terminated.

so we need to add `path[sz] = 0` into "if" body.

native-api:
Here. The "not null-terminated" remark only relates to the case where the path doesn't fit. So checking for MAX_PATH is enough. A valid C string cannot be of `len` size in a `len` size buffer, after all :-)

PR #10264, opened 2017-12-08
updated createsamples app utility functions to new c++ API from old c

terfendail:
### This pullrequest changes
createsamples app utility functions changed to use new C++ OpenCV API instead of old C


StevenPuttemans:
Great! Something that was on my TODO list for ages ^_^

Issue #7862, opened 2016-12-15
Why cv::resize with CV_8S and CV_32S type was disabled?


logic1988:
Using cv::resize function with cv::Mat in `CV_8S` and `CV_32S` type will cause an error:
`opencv/modules/imgproc/src/imgwarp.cpp:3368: error: (-215) func != 0 in function resize`
After review of the source code I found:
```
    static ResizeFunc linear_tab[] =
    {
        resizeGeneric_<
            HResizeLinear<uchar, int, short,
                INTER_RESIZE_COEF_SCALE,
                HResizeLinearVec_8u32s>,
            VResizeLinear<uchar, int, short,
                FixedPtCast<int, uchar, INTER_RESIZE_COEF_BITS*2>,
                VResizeLinearVec_32s8u> >,
        0,
        resizeGeneric_<
            HResizeLinear<ushort, float, float, 1,
                HResizeLinearVec_16u32f>,
            VResizeLinear<ushort, float, float, Cast<float, ushort>,
                VResizeLinearVec_32f16u> >,
        resizeGeneric_<
            HResizeLinear<short, float, float, 1,
                HResizeLinearVec_16s32f>,
            VResizeLinear<short, float, float, Cast<float, short>,
                VResizeLinearVec_32f16s> >,
        0,
        resizeGeneric_<
            HResizeLinear<float, float, float, 1,
                HResizeLinearVec_32f>,
            VResizeLinear<float, float, float, Cast<float, float>,
                VResizeLinearVec_32f> >,
        resizeGeneric_<
            HResizeLinear<double, double, float, 1,
                HResizeNoVec>,
            VResizeLinear<double, double, float, Cast<double, double>,
                VResizeNoVec> >,
        0
    };
```
This array corresponds to the type of Mat:
```
#define CV_8U   0
#define CV_8S   1
#define CV_16U  2
#define CV_16S  3
#define CV_32S  4
#define CV_32F  5
#define CV_64F  6
#define CV_USRTYPE1 7
```
The array value associated with `CV_8S` and `CV_32S` is 0, which led to error.
`cv::resize` is an important function, but the bug seems to exist for a long time, which have any special reasons?

alalek:
Do you have real use cases for these formats, especially with linear interpolation?
There are more relevant formats:
32S -> 32F
8S -> 16S

logic1988:
Yes, but I am very surprised why 16S can, but 8S and 32S but not.

HGuillemet:
[...]

PR #10266, opened 2017-12-08
doc: Add type restriction note for geometric transformations

ChristofKaufmann:
This is added according to #7862.

### This pullrequest changes

Although I think it would be better to make geometric image transformations work for `CV_8S` or `CV_32S` images, it should at least be documented that it does not.


vpisarev:
:+1:

PR #10268, opened 2017-12-08
Scale layer in case of 2D inputs

dkurt:
### This pullrequest changes

Test case with scale layer right after a fully-connected one: `getPlane` has [an assertion](https://github.com/opencv/opencv/blob/master/modules/dnn/include/opencv2/dnn/shape_utils.hpp#L114) that number of dimensions is greater than 2.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/417


vpisarev:
:+1:

PR #10269, opened 2017-12-08
Implementation of softdouble rounding to int64_t

terfendail:
### This pullrequest changes
Added implementation of softdouble rounding to int64_t and test for rounding and saturation_cast to int and int64

### DIFF
--- a/modules/core/test/test_math.cpp
+++ b/modules/core/test/test_math.cpp
@@ -3736,4 +3736,130 @@ TEST(Core_SoftFloat, sincos64)
     }
 }
 
+TEST(Core_SoftFloat, CvRound)
+{
+    struct
+    {
+        uint64_t inVal;
+        int64_t out64;
+        int32_t out32;
+    } _values[] =
+    {
+        { 0x0123456789abcdefU,                     0,             0 }, // 3.51270056408850369812238561681E-303
+        { 0x0000000000000000U,                     0,             0 }, // 0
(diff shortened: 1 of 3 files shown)


modules/core/test/test_math.cpp
```diff
@@ -3736,4 +3736,130 @@ TEST(Core_SoftFloat, sincos64)
...
+    struct testvalues
+    {
+        softdouble inVal;
+        int64_t out64;
+        int32_t out32;
+    } *values = (testvalues*)_values;
+
+    for (int i = 0, maxi = sizeof(_values) / sizeof(_values[0]); i < maxi; i++)
+    {
+        ASSERT_EQ(values[i].out64, cvRound64(values[i].inVal));
```
alalek: ASSERT_EQ is for fatal errors, like empty input image.
These checks don't look fatal, so please use `EXPECT_EQ` here.

savuor: @alalek What about ASSERTS at other SoftFloat tests in the same file? Are they to be replaced by EXPECT_* ?

savuor:
:+1:

modules/core/test/test_math.cpp
alalek: Rule is the same: If problem is not critical and doesn't break other checks completely and these checks are independent then `EXPECT` is preferable.
But be careful with large cycles (for each pixel from image) - we don't want 10Gb log files of errors: use `ASSERT` or `break` from cycle loop.

PR #10274, opened 2017-12-09
3rdparty: update dependencies

alalek:
based on #9781 by @sturkmen72 

- zlib 1.2.1
- libpng 1.6.34
- libjpeg 9b
- libtiff 4.0.9
- webp 0.6.1

Jasper library is not updated (likely will be replaced to [OpenJPEG](http://www.openjpeg.org/))

OpenEXR update commits are excluded (too large with autogenerated data): https://github.com/alalek/opencv/commits/update_openexr

CMake output (Win64 builder):
```
--   Media I/O: 
--     ZLib:                        build (ver 1.2.11)
--     JPEG:                        build (ver 90)
--     WEBP:                        build (ver encoder: 0x020e)
--     PNG:                         build (ver 1.6.34)
--     TIFF:                        build (ver 42 - 4.0.9)
--     JPEG 2000:                   build (ver 1.900.1)
--     OpenEXR:                     build (ver 1.7.1)
```

```
allow_multiple_commits=1
```


sturkmen72:
@alalek there is some issue about libpng see https://github.com/glennrp/libpng/issues/183

alalek:
Actually I observe similar problem with my "system" libpng (default mode in OpenCV) on mentioned image:
```
--     PNG:                         /lib64/libpng.so (ver 1.6.28)
```

> libpng error: IDAT: invalid distance too far back

I believe these images should be fixed instead of "supporting" of malformed inputs (which leads to applications vulnerabilities).

"pngfix" utility states this:

> Various known problems in PNG files are fixed while the file is being read
The exit code says what problems were fixed.  In particular the zlib error:
    "invalid distance too far back"
caused by an incorrect optimization of a zlib stream is fixed in any
compressed chunk in which it is encountered.  An integrity problem of the
PNG stream caused by a bug in libpng which wrote an incorrect chunk length
is also fixed.  Chunk CRC errors are automatically fixed up.


BTW, Even after "pngfix" image data is still looks broken (bottom lines are lost). The same image result is with png 1.6.24 which eats this problem.

alalek:
:+1:

PR #10287, opened 2017-12-11
Update createsamples app

### LINKED ISSUE #10267: createsamples: reproducible results
Currently there is no way to reproduce results of createsamples tool. Launching with the same parameters doesn't return the same results.

- This tool uses `srand((unsigned int)time(0));`. Need to add command-line option to specify seed with some fixed default (like 12345).
- This tool uses `rand()`. It is not a thread-safe call. Need to replace these calls on `cv::RNG` API.

sturkmen72:
resolves #10267

### This pullrequest changes
adds command-line option -rngseed
replaces rand() -> theRNG()

*WIP*

```
```

### DIFF
--- a/apps/createsamples/createsamples.cpp
+++ b/apps/createsamples/createsamples.cpp
@@ -45,15 +45,14 @@
  * Create test/training samples
  */
 
+#include "opencv2/core.hpp"
+#include "utility.hpp"
 #include <cstdio>
 #include <cstring>
 #include <cstdlib>
 #include <cmath>
-#include <ctime>
 
 using namespace std;
-
-#include "utility.hpp"
 
 int main( int argc, char* argv[] )
 {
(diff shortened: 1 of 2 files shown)


sturkmen72:
@alalek could you take a look at the code. i will test it later but let me know if i am on the right path.

alalek:
@sturkmen72 It seems good! Thank you!

sturkmen72:
@alalek please review it later. i am still working on it.

sturkmen72:
@terfendail do you have time to check #10264 there is a problem when calling the program like below ( it runs good when the commit reverted )

`opencv_createsamples -img face.jpg -num 200 -vec positives.vec -w 88 -h 48`

 it generates vec file width and height swapped

`opencv_createsamples -img face.jpg -num 200 -vec positives.vec -w 148 -h 48`

raises error

`OpenCV Error: Assertion failed (0 <= roi.x && 0 <= roi.width && roi.x + roi.width <= m.cols && 0 <= roi.y && 0 <= roi.height && roi.y + roi.height <= m.rows) in cv::Mat::Mat, file E:\git\opencv\modules\core\src\matrix.cpp, line 538`

Issue #10229, opened 2017-12-05
OpenCV doesn't link on Android x86/x86_64 with ndk 16.0.4442984 with R_386_GOTOFF error.


erikman:
##### System information (version)
- OpenCV => 3.3.0
- Operating System / Platform => Android x86/x86_64
- Compiler => Android toolchain (OpenCV binaries downloaded from github 3.3.0/3.3.1 release, My binary built from Android Studio, ndk-bundle-16.0.4442984 darwin-x86_64-gcc-i686-linux-android-4.9.x)

##### Detailed description
Can no longer link application for x86 or x86_64, get `relocation R_386_GOTOFF against preemptible symbol` in symbols from `libippiw.a` and `libippicw.a`. Rebuilding OpenCV for Android without IPP support fixes the problem. IPP are only used on x86/x86_64 so this is not a problem for arm builds, but the x86 builds are useful for debugging on HAX accelerated emulator.

Problem appeared after Android ndk upgrade (don't know my old version). The IPP binaries seems to be a binary only release compiled by Intel so I can't test rebuilding those libraries, otherwise searching the net suggests that the binaries should be compiled with `-fPIC` to fix this issue.

airman00:
I experienced the same issue when I upgraded Android Studio on MacOSX.

Downgrading from NDK16 to NDK14 temporarily solved the issue for me

To downgrade:
1. Download the older NDK (use 14) from here:
https://developer.android.com/ndk/downloads/older_releases.html
2. Unzip and rename entire NDK folder to "ndk-bundle"
3. Replace NDK 16 with NDK 14 by replacing the "ndk-bundle" in your Android home folder (for me it was /Library/Android/sdk) with the "ndk-bundle" folder from Step 2

hederguerreiro:
I had the same issue and your solution @airman00 solved my problem. Thanks!

sergiomarquesmoura:
Any update on this? Have the same issue and downgrading is not really an option because I'm building on an external server with the latest NDK (v16) installed

alalek:
Current workaround is disabling IPP (affects x86/x86_64 platforms): add "-DWITH_IPP=OFF" CMake option into build script.

sergiomarquesmoura:
@alalek I see, but in practice we currently have to remove the x86 and x86_64 ABIs right?

alalek:
> OpenCV binaries downloaded

[...]

PR #10288, opened 2017-12-12
Make it possible to build with Android NDK r16

erikman:
This pull request makes it possible to build OpenCV with Android NDK r16. Android have switched to unified include directories, and deprecated the `android` executable.

This patch makes it possible to compile the native libraries. On NDK r16 the `android` executable is not able to create `build.xml` files for use with `ant` so it will not build any `.apk` files.

### DIFF
--- a/cmake/OpenCVDetectAndroidSDK.cmake
+++ b/cmake/OpenCVDetectAndroidSDK.cmake
@@ -17,6 +17,7 @@ find_host_program(AVDMANAGER_EXECUTABLE
   NAMES avdmanager
   PATH_SUFFIXES tools/bin
     PATHS
+      ${ANDROID_SDK}
       ENV ANDROID_SDK
     DOC "Android AVD Manager location"
     NO_DEFAULT_PATH
@@ -27,6 +28,7 @@ find_host_program(ANDROID_EXECUTABLE
   NAMES android.bat android
   PATH_SUFFIXES tools
   PATHS
+    ${ANDROID_SDK}
     ENV ANDROID_SDK
   DOC "Android SDK location"
   NO_DEFAULT_PATH
--- a/cmake/OpenCVDetectAndroidSDK.cmake
+++ b/cmake/OpenCVDetectAndroidSDK.cmake
@@ -13,6 +13,15 @@ else()
   set(ANDROID_SDK_OS linux)
 endif()
 
+find_host_program(AVDMANAGER_EXECUTABLE
+  NAMES avdmanager
+  PATH_SUFFIXES tools/bin
+    PATHS
+      ENV ANDROID_SDK
+    DOC "Android AVD Manager location"
+    NO_DEFAULT_PATH
+    )
+
 #find android SDK: search in ANDROID_SDK first
 find_host_program(ANDROID_EXECUTABLE
   NAMES android.bat android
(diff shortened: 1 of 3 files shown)


alalek:
Consider using Android CMake toolchain file instead (it shipped with NDK 15+, so it is supported natively and should be well tested):
```
android-ndk-r15c/build/cmake/android.toolchain.cmake
```
Options are similar, but not exact. Something like this:
> cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK}/build/cmake/android.toolchain.cmake -DANDROID_NATIVE_API_LEVEL=android-23 -DBUILD_SHARED_LIBS=OFF -DENABLE_CXX11=ON -DANDROID_STL=c++_static -DANDROID_ABI="armeabi-v7a" -DANDROID_TOOLCHAIN=clang -DANDROID_ARM_NEON=ON <opencv_src>

Lets leave OpenCV's Android toolchain "as is" for legacy NDKs (and old devices).

erikman:
Thanks @alalek, that sounds like a good idea, I'll have a look

hannesa2:
In my personal Android point of view I want to see it **merged** !

With this spirit "let things untouched although a improvement/modernization is there" let OpenCV midterm (current) **unusable**.
And why ? Because of 3% of all users ! 
Please tell them "Sorry 3.4.0 is your last version support api 9, we have to got to the future. Urgent !" 

Look at current Android situation, it's not usable out of the box:
* outdated Eclipse sample project, support ended one year ago. since 2015 we know Android Studio is the future
* you can only build artifacts with archeological build tools, it's not specified which one
* nowhere any adaption to recent version is done, new user needs days to use it
* ... and a lot of more reasons

alalek:
@hannesa2 We can't merge any patches, because who would maintain this a few months later?

1.
The first two commits just hides "android" tool problem, without resolving them (creates .apk files with zero size, including Android tests - do we really need this?). Only complete migration to gradle builds can help here. Also ANDROID_SDK variable is not in recommended set of environment variables: https://developer.android.com/studio/command-line/variables.html

2.
The third 90cc3f6 commit is useful, but it is not complete: CMake is not able to properly call "find_file()" on headers.

3.
I'm not able to reproduce problem resolved by 3135af8 commit. It should not break anything, but I have no idea that problem related to NDK16 is resolved here. Is it a build problem or runtime issue (.apk crash on launch)?

erikman:
@alalek The problem is a build problem, I ran into this when I needed to recompile due to #10229 so my only goal was to be able to build libraries with r16 with my own build flags. I did not need the `.apk` at all so I was fine with not building it. So yes this hides the `android` tool problem, but will allow the build to produce libraries that can be used.

I was not aware of `ANDROID_SDK_ROOT` being available, I'll update to using that if available.

3135af8 and 90cc3f6 are not needed when using `android.toolchain.cmake` from ndk.

With the two first patches I'm able to build with the toolchain from ndk. Still doing some final testing.

erikman:
I looked into `ANDROID_SDK_ROOT` and it is not set as an environment variable when gradle runs cmake (I find it strange that we only get the ndk location and not the sdk location, see https://developer.android.com/ndk/guides/cmake.html ).

Locally I get around this by passing it from my `gradle.build`:

```
android {
    defaultConfig {
        externalNativeBuild {
            cmake {
                arguments "-DANDROID_SDK=${project.android.sdkDirectory}"
            }
       }
   }
}
```

my point is that I would like to pass the path as a cmake variable and not just an environment variable (which is harder to pass into cmake from gradle). So what I was doing in my patch was only to expose the `ANDROID_SDK` that already was used by OpenCV to also be exposed as a cmake variable.

Now I'm a bit hesitant to rename it since it is used by many of the support scripts: `test_cmake_build.py`, `samples/android/hello-android/run.cmd`, `test_ant_build.py`, `android/build_sdk.py` and `modules/ts/misc/run_android.py` and `modules/ts/misc/run.py`. @alalek: What do you think?

vpisarev:
@erikman, thank you! Some parts of your patch have been already re-used in  https://github.com/opencv/opencv/pull/10515. Let's close this one

PR #10299, opened 2017-12-13
cmake: rework MAP_IMPORTED_CONFIG to prevent unintended build failures

### LINKED ISSUE #10105: OpenCVConfig.cmake: issue with imported target configuration remapping
For development and debugging on linux, I tend to build OpenCV in ```RelWithDebInfo``` mode for use with my projects, which are also build in ```RelWithDebInfo``` mode.

Since commit 2d0611bc01b7c51bff8567c332077bc5c3f64c10, the above setup is broken; the projects that use OpenCV configure without errors, but the build fails with:
```
make[2]: *** No rule to make target 'opencv_core-NOTFOUND', needed by 'program-name'.  Stop
```

This is because ```RelWithDebInfo``` imported targets are now mapped to ```Release```, which does not exist in this case (as OpenCV was also built in ```RelWithDebInfo``` mode).

The comment added to ```OpenCVConfig.cmake``` by 2d0611bc01b7c51bff8567c332077bc5c3f64c10 states that "OpenCV supports Debug and Release only". This is not the always the case, as building OpenCV in ```RelWithDebInfo``` on linux results in build that generates ```OpenCVModules-relwithdebinfo.cmake```.

Therefore, the remapping should be done only if the requested configuration does not actually exist, or it should be limited to platforms/toolchains where it is known that ```RelWithDebInfo``` and ```MinSizeRel``` are not supported (presumably Visual Studio?).

@bblanchon, as the author of the change, can you comment?

alalek:
Introduce CMake variable OPENCV_MAP_IMPORTED_CONFIG which controls this behaviour
Added defaults for MSVS targets

resolves #10105

### DIFF
--- a/cmake/templates/OpenCVConfig.cmake.in
+++ b/cmake/templates/OpenCVConfig.cmake.in
@@ -126,6 +126,50 @@ if(NOT CMAKE_VERSION VERSION_LESS "2.8.11")
   endforeach()
 endif()
 
+
+if(NOT DEFINED OPENCV_MAP_IMPORTED_CONFIG)
+  if(CMAKE_GENERATOR MATCHES "Visual Studio" OR MSVC)
+    # OpenCV supports Debug and Release builds only.
+    # But MSVS has 'RelWithDebInfo' 'MinSizeRel' configurations for applications.
+    # By default CMake maps these configuration on the first available (Debug) which is wrong.
+    # Non-Debug build of Application can't be used with OpenCV Debug build (ABI mismatch problem)
+    # Add mapping of RelWithDebInfo and MinSizeRel to Release here
+    set(OPENCV_MAP_IMPORTED_CONFIG "RELWITHDEBINFO=!Release;MINSIZEREL=!Release")
+  endif()
+endif()
+set(__remap_warnings "")
+macro(ocv_map_imported_config target)
+  if(DEFINED OPENCV_MAP_IMPORTED_CONFIG) # list, "RELWITHDEBINFO=Release;MINSIZEREL=Release"
+    get_target_property(__available_configurations ${target} IMPORTED_CONFIGURATIONS)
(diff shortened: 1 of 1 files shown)


alalek:
@TobiasLangnerFUB @tfoote @pjreed @rokm @bblanchon 
Could you please check this patch with your build configurations?

bblanchon:
@alalek I'm not able to test right now as I'm working on a different project.

However, it looks very complicated, so I think it would be more reasonable to simply revert my commit.
Users who want to use RelWithDebugInfo would just need to build OpenCV themselves; apparently, most of them are already doing that anyway.

TobiasLangnerFUB:
@alalek The patch works for my configuration.

pauldinh:
I'm also having success with this patch in my builds.

tdenewiler:
My use case is building and using ROS packages that depend on OpenCV (my current environment is Ubuntu 16.04 with ROS Kinetic). I have OpenCV installed via ROS apt packages from their buildfarm (specifically `ros-kinetic-opencv3`). I normally build the entire ROS workspace with the `-DCMAKE_BUILD_TYPE=RelWithDebInfo` flag set. To test if this change works I naively applied the diff to my existing `/opt/ros/kinetic/share/OpenCV-3.3.1/OpenCVConfig.cmake` file. With the changes in this PR I was able to build my workspace and get the expected output from OpenCV calls.

This works for me.

alalek:
Thanks all for testing that!

@bblanchon I tested this patch with MSVS CMake generator, Debug+Release OpenCV installation and `RelWithDebugInfo` config for simple application (without Qt) - it works.

vpisarev:
:+1:

alex-petrenko:
This does not seem to work anymore. opencv_dnn module depends on some other 3rd party libs (such as protobuf) and my external project in RelWithDebInfo configuration tries to link with Debug build of libprotobuf. Is there a workaround?

PR #10303, opened 2017-12-13
core:ppc Fix several issues for VSX

seiko2plus:
## This pullrequest changes
- fix conversion intrinsics compatibility with xlc
- implement odd-elements 2 to 4 conversion intrinsics
- improve implementation of universal intrinsic v_popcount

````
force_builders=Custom
````

### CODE REVIEW

**Comment by alalek** [comment]
Thank you for contribution!

There is `FORCE_INLINE` definition in "vsx_utils.hpp" file:
```
#define FORCE_INLINE(tp) extern inline tp __attribute__((always_inline))
```

Could you please rename it to `VSX_FORCE_INLINE` or add `#undef FORCE_INLINE` to the end of this file (it is not used in other files)?

seiko2plus:
@alalek yes, it isn't used in other files but I think it would be better to rename it.

PR #10304, opened 2017-12-13
Add a tutorial about the homography

catree:
### This pullrequest changes

I would like to add a tutorial about the homography concept. While this is mainly a theoretical subject, the approach here is more to illustrate the basic concept of the homography with some demonstration codes. This is also a good way for me to summarize all the information I could have gathered by written in one location:

- pose from homography
- perspective correction
- homography from camera displacement
- homography decomposition

I hope it could be useful to other people.

[Demo](http://pullrequest.opencv.org/buildbot/export/pr/10304/docs/d9/dab/tutorial_homography.html)


samples/cpp/tutorial_code/features2D/Homography/tutorial_homography_ex4_decompose_homography.cpp
```diff
@@ -0,0 +1,186 @@
+#include <iostream>
+#include <opencv2/opencv.hpp>
+#include <opencv2/aruco.hpp>
```
alalek:
```
#ifdef HAVE_OPENCV_ARUCO // defined in opencv_modules.hpp
#include <opencv2/aruco.hpp>

... all other sample code ...
#else
{
std::cerr << "FATAL ERROR: This sample requires opencv_aruco module (from opencv_contrib)";
return 0;
}
#endif
```

catree:
Thanks! I have replaced the png images by jpg ones.

vpisarev:
excellent contribution!

PR #10307, opened 2017-12-14
SoftFloat tests: ASSERT* to EXPECT*

savuor:
Assertions replaced by expectations outside of cycles in SoftFloat tests


vpisarev:
:+1:

PR #10316, opened 2017-12-15
Added universal intrinsics based implementations for multichannel bit-exact resize

terfendail:
### This pullrequest changes
Added universal intrinsics based implementations for CV_8UC2, CV_8UC3 and CV_8UC4


vpisarev:
excellent! :+1:

PR #10329, opened 2017-12-16
remove VTK legacy functions

jasjuang:
### This pullrequest changes

Recently VTK removes some legacy functions that causes OpenCV to stop compiling.

https://github.com/Kitware/VTK/commit/fc2ac65dc798d932974abc500f14b94e2a5208f9

The fix is applied according to the suggestion of the VTK devs.

I tested and it compiles and runs without a problem on Ubuntu 16.04 latest OpenCV.

```
force_builders=Custom
```

### DIFF
--- a/modules/viz/src/widget.cpp
+++ b/modules/viz/src/widget.cpp
@@ -91,7 +91,9 @@ cv::viz::Widget cv::viz::Widget::fromPlyFile(const String &file_name)
 
     vtkSmartPointer<vtkDataSetMapper> mapper = vtkSmartPointer<vtkDataSetMapper>::New();
     mapper->SetInputConnection( reader->GetOutputPort() );
+#if VTK_MAJOR_VERSION < 8
     mapper->ImmediateModeRenderingOff();
+#endif
 
     vtkSmartPointer<vtkActor> actor = vtkSmartPointer<vtkActor>::New();
     actor->GetProperty()->SetInterpolationToFlat();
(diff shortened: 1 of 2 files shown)


modules/viz/src/widget.cpp
```diff
@@ -113,7 +112,6 @@ void cv::viz::Widget::setRenderingProperty(int property, double value)
         case POINT_SIZE:          actor->GetProperty()->SetPointSize(float(value)); break;
         case OPACITY:             actor->GetProperty()->SetOpacity(value);          break;
         case LINE_WIDTH:          actor->GetProperty()->SetLineWidth(float(value)); break;
-        case IMMEDIATE_RENDERING: actor->GetMapper()->SetImmediateModeRendering(int(value)); break;
```
alalek: Please use appropriate message (probably on "stderr") here that this property has no effect.

Silent removal cause old app got broken on this line:
```
CV_Assert("setPointCloudRenderingProperties: Unknown property");
```

jasjuang:
@alalek update as you requested

jasjuang:
@alalek I learned from the PCL folks that updating the condition to `#if VTK_RENDERING_BACKEND_OPENGL_VERSION < 2` is more precise. I just pushed the change.

vpisarev:
@jasjuang, we require all the builders to run without any compile warnings, let alone compile errors. So, please, find some solutions or workarounds to make all the builders green. Also, I would recommend you to create your PR out of a dedicated branch, not master. This way you will get your builds to rerun automatically with each your commit.

jasjuang:
@vpisarev my bad `VTK_RENDERING_BACKEND_OPENGL_VERSION` is actually a pcl defined macro, I switched back to `VTK_MAJOR_VERSION` and it compiles on my machine. 

As for the travis builds, what VTK is in it? I believe starting from VTK 8 it needs to be compiled with c++11 otherwise there will be `nullptr` unrecognized error from VTK headers. I had to manually add c++11 flags in ccmake in order for OpenCV to compile.

alalek:
Windows - VTK is not available
Linux - VTK 6.2 (system package from Ubuntu 16.04 LTS), 6.3 (Ubuntu 17.10)
Mac - VTK 8.0.1 (via homebrew)

jasjuang:
@alalek @vpisarev It looks like the build passes, so you guys must've handle the c++11 flags. I think this is ready to merge.

PR #10215, opened 2017-12-02
JavaScript bindings for dnn module

dkurt:
```
force_builders=Custom
docker_image:Docs=docs-js
docker_image:Custom=javascript
```

PR #10331, opened 2017-12-16
Changed wrapping mode for dnn::Net::forward

arrybn:
### This pullrequest changes
Changed wrapping mode for cv::dnn::Net::forward to avoid conflicts in python wrappers

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -423,8 +423,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
          *  @param outputBlobs contains all output blobs for each layer specified in @p outBlobNames.
          *  @param outBlobNames names for layers which outputs are needed to get
          */
-        void forward(std::vector<std::vector<Mat> >& outputBlobs,
-                     const std::vector<String>& outBlobNames);
+        CV_WRAP_AS(forwardAndRetrieve) void forward(CV_OUT std::vector<std::vector<Mat> >& outputBlobs,
+                                                    const std::vector<String>& outBlobNames);
 
         //TODO:
         /** @brief Optimized forward.


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -410,7 +410,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
          *  @param outputName name for layer which output is needed to get
          *  @details If @p outputName is empty, runs forward pass for the whole network.
          */
-        CV_WRAP void forward(OutputArrayOfArrays outputBlobs, const String& outputName = String());
+        CV_WRAP void forward(CV_OUT OutputArrayOfArrays outputBlobs, const String& outputName = String());
```
dkurt: @arrybn, thank you! Python bindings generator can do it by `OutputArrayOfArrays`: https://docs.opencv.org/master/da/d49/tutorial_py_bindings_basics.html

> Header parser can understand the input and output arguments from keywords like InputArray, OutputArray etc. But sometimes, we may need to hardcode inputs and outputs. For that, macros like CV_OUT, CV_IN_OUT etc. are used.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -423,7 +423,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
          *  @param outputBlobs contains all output blobs for each layer specified in @p outBlobNames.
          *  @param outBlobNames names for layers which outputs are needed to get
          */
-        void forward(std::vector<std::vector<Mat> >& outputBlobs,
+        void forward(CV_OUT std::vector<std::vector<Mat> >& outputBlobs,
```
dkurt: We've removed `CV_WRAP` macro here because it's wrapped similarly to a previous one method. So there are two ways  to do `forward` pass in python: with a single string or with a list of strings. May be it's better to wrap it by `CV_WRAP_AS` using different name?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -423,7 +423,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
                      const std::vector<String>& outBlobNames);
```
alalek: @arrybn Hi! Perhaps there is missing `CV_WRAP` keyword.

modules/dnn/include/opencv2/dnn/dnn.hpp
alalek: Agree with @dkurt . CV_OUT is not necessary here.
Perhaps there is some bug/issue with bindings generator itself.
Looks like nothing was changed by this patch in Python signatures:
- [master](https://docs.opencv.org/master/db/d30/classcv_1_1dnn_1_1Net.html#a8d248f50efec70dffa785de18a4b2170)
- [patch](http://pullrequest.opencv.org/buildbot/export/pr/10331/docs/db/d30/classcv_1_1dnn_1_1Net.html#a8d248f50efec70dffa785de18a4b2170)

Workaround is to pass None:
```
outputBlobs = new.forward(outputBlobs=None, outputName='name')
```

BTW, There is this method: `CV_WRAP Mat forward(const String& outputName = String());`. Python binding of this method has strong conflict with method above (can't be resolved properly without renaming like `CV_WRAP_AS`).

arrybn:
@alalek @dkurt I used CV_WRAP_AS , thanks for advice. It seems that we need OutputArrayOfArraysOfArrays

vpisarev:
@arrybn, glad to hear from you and thanks for the patch! ) I'd suggest to choose some better name than 'verboseForward'; to me verbose sounds like the method prints some internal information while proceeding with the forward pass. Perhaps, you should reflect that you want to retrieve the results from the specified intermediate layers.

arrybn:
I've come up with few options for the function wrapped name:

- forward2layers (or forwardToLayers)
- getLayersOutputs
- getOutputsForLayers
- passLayers

What do you think? Which is the best?

vpisarev:
@arrybn, since the method is the variation of "forward()" it should contain "forward" in the name. Out of the proposed variants, I think, "forwardToLayers" is the best. From my side I would suggest "forwardAndRetrieve" or "forwardPartial"

alalek:
"forwardMultiple"/"forwardMultipleLayers"

vpisarev:
"forwardMultiple" is not good, because it sounds like using `batch_size > 1`. "forwardMultipleLayers" is better, but still a bit confusing, in my opinion.

arrybn:
Changed to forwardAndRetrieve. Thanks, guys!

dkurt:
@arrybn, May I ask you to check JavaScript bindings too? Here is a tutorial of how to do it, https://docs.opencv.org/master/d4/da1/tutorial_js_setup.html. I afraid we've removed wrapper at https://github.com/opencv/opencv/pull/10215/files#diff-8670560265f0e0820497bbec415b9452L426 because of it.

dkurt:
:+1:

PR #10337, opened 2017-12-17
Fix default pooling layer type

dkurt:
### This pullrequest changes

Fixes case with default pooling layer type from Caffe:
```yaml
layer {
  name: "conv1_pool"
  type: "Pooling"
  bottom: "conv1_h"
  top: "conv1_pool"
  pooling_param {
    kernel_size: 3
    stride: 2
  }
}
```


vpisarev:
:+1:

PR #10340, opened 2017-12-17
core(logger): add log level configuration option

alalek:
Example why we need here two .hpp files - some .cpp module:
```
// want to enable "debug" messages in release builds too
#include <opencv2/core/utils/logger.defines.hpp>
#undef CV_LOG_STRIP_LEVEL
#define CV_LOG_STRIP_LEVEL CV_LOG_LEVEL_VERBOSE
#include <opencv2/core/utils/logger.hpp>
```
Without the first include, compiler warns about undefined symbols.

### DIFF
--- a/modules/core/include/opencv2/core/utils/logger.defines.hpp
+++ b/modules/core/include/opencv2/core/utils/logger.defines.hpp
@@ -1,0 +1,22 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef OPENCV_LOGGER_DEFINES_HPP
+#define OPENCV_LOGGER_DEFINES_HPP
+
+//! @addtogroup core_logging
+//! @{
+
+// Supported logging levels and their semantic
+#define CV_LOG_LEVEL_SILENT 0          //!< for using in setLogLevel() call
+#define CV_LOG_LEVEL_FATAL 1           //!< Fatal (critical) error (unrecoverable internal error)
+#define CV_LOG_LEVEL_ERROR 2           //!< Error message
+#define CV_LOG_LEVEL_WARN 3            //!< Warning message
+#define CV_LOG_LEVEL_INFO 4            //!< Info message
+#define CV_LOG_LEVEL_DEBUG 5           //!< Debug message. Disabled in the "Release" build.
+#define CV_LOG_LEVEL_VERBOSE 6         //!< Verbose (trace) messages. Requires verbosity level. Disabled in the "Release" build.
+
+//! @}
+
+#endif // OPENCV_LOGGER_DEFINES_HPP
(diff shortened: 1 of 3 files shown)


vpisarev:
@alalek, do you plan to insert some test use of this thing somewhere in OpenCV code? Just to see how it looks like in practice and whether it's convenient enough?

modules/core/include/opencv2/core/utils/logger.configuration.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#ifndef OPENCV_LOGGER_CONFIGURATION_HPP
+#define OPENCV_LOGGER_CONFIGURATION_HPP
+
+//! @addtogroup core_logging
+//! @{
+
+namespace cv {
+namespace utils {
+namespace logging {
```
mshabunin: Probably namespace statements can be omitted here.

modules/core/include/opencv2/core/utils/logger.configuration.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+// Supported logging levels and their semantic
+#define CV_LOG_LEVEL_SILENT 0          //!< for using in setLogLevel() call
+#define CV_LOG_LEVEL_FATAL 1           //!< Fatal (critical) error (unrecoverable internal error)
+#define CV_LOG_LEVEL_ERROR 2           //!< Error message
+#define CV_LOG_LEVEL_WARN 3            //!< Warning message
+#define CV_LOG_LEVEL_INFO 4            //!< Info message
+#define CV_LOG_LEVEL_DEBUG 5           //!< Debug message. Disabled in the "Release" build.
+#define CV_LOG_LEVEL_VERBOSE 6         //!< Verbose (trace) messages. Requires verbosity level. Disabled in the "Release" build.
+
```
mshabunin: I think `CV_LOG_STRIP_LEVEL` macro definition can be moved here.

modules/core/include/opencv2/core/utils/logger.configuration.hpp
alalek: Removed

modules/core/include/opencv2/core/utils/logger.configuration.hpp
alalek: See updated PR description.

alalek:
@vpisarev This can't be tested via current GoogleTests based OpenCV tests. Except useless `setLogLevel()`/`getLogLevel()` pair (which should not corrupt "external" explicit settings).
Tracking real process stdout/stderr output requires additional testing infrastructure which I don't want to introduce by this patch.

Try to run this and check the difference:
```
$./bin/opencv_test_core --gtest_filter=*OpenCL*programs
$ OPENCV_LOG_LEVEL=WARN ./bin/opencv_test_core --gtest_filter=*OpenCL*programs
```

modules/core/include/opencv2/core/utils/logger.configuration.hpp
mshabunin: It should work even if we define _level_ macros after _strip_ macro:
```.cpp
#define CV_LOG_STRIP_LEVEL CV_LOG_LEVEL_DEBUG
#include <opencv2/core/utils/logger.hpp>
```

vpisarev:
@alalek, ok, thanks!

PR #10341, opened 2017-12-18
add ocl accleration for prior box layer and depthwise convolution

pengli:
with depthwise convolution kernel, we can dramatically improve mobilenet ssd performance

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/src/layers/prior_box_layer.cpp
+++ b/modules/dnn/src/layers/prior_box_layer.cpp
@@ -45,6 +45,7 @@
 #include <float.h>
 #include <algorithm>
 #include <cmath>
+#include "opencl_kernels_dnn.hpp"
 
 namespace cv
 {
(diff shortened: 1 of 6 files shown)


vpisarev:
@alalek, I think, it would be nice to squeeze it into 3.4

dkurt:
On my machine performance of MobileNet-SSD from Caffe (20 classes) is changed from 172.13ms to 26.66ms (22.71ms on CPU). MobileNet-SSD from TensorFlow for 90 classes is boosted from 203.47ms to 45.11ms (25.15ms on CPU). See https://github.com/opencv/opencv/wiki/DNN-Efficiency.

@pengli, May I aks you to add a test for the last model like a https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_tf_importer.cpp#L189?

modules/dnn/src/layers/prior_box_layer.cpp
```diff
@@ -245,6 +246,21 @@ class PriorBoxLayerImpl : public PriorBoxLayer
...
+#ifdef HAVE_OPENCL
+        Mat offsetsX(1, _offsetsX.size(), CV_32FC1, &_offsetsX[0]);
+        Mat offsetsY(1, _offsetsX.size(), CV_32FC1, &_offsetsY[0]);
+        Mat aspectRatios(1, _aspectRatios.size(), CV_32FC1, &_aspectRatios[0]);
+        Mat variance(1, _variance.size(), CV_32FC1, &_variance[0]);
+
+        offsetsX.copyTo(umat_offsetsX);
+        offsetsY.copyTo(umat_offsetsY);
+        aspectRatios.copyTo(umat_aspectRatios);
+        variance.copyTo(umat_variance);
```
alalek: BTW, I believe we should not increase memory usage for non-OpenCL backends.
This should be moved into OpenCL specific code (for example, in `forward_ocl()` and called under some "initialization" flag once per layer instance).

vpisarev: agree. Such extensive consumption of OpenCL-only memory should not happen by default

pengli:
@dkurt,  I pushed one more commit  in this PR : "add ocl accuracy test for tf mobilenet ssd" ,  please check and see if this is what you want.

modules/dnn/src/layers/prior_box_layer.cpp
pengli: Fixed in the new update, moved the code into forward_ocl().

pengli:
@alalek, convolution verification failure has been fixed in the new update.

vpisarev:
@alalek, shall we merge it now? looks like the PR brings some significant performance improvements!

PR #10343, opened 2017-12-18
3rdparty: suppress warning on Visual Studio

tomoaki0705:
* remove non ASCII character from the comment
```
  C:\opencv\3rdparty\libwebp\src\enc\histogram_enc.c : warning C4819: The file contains a character that cannot be represented in the current code page (932). Save the file in Unicode format to prevent data loss [c:\opencv\build\3rdparty\libwebp\libwebp.vcxproj]
```


vpisarev:
:+1:

PR #10344, opened 2017-12-18
fix compilation error on Visual Studio 2013 and earlier

### LINKED ISSUE #10342: 3rdparty: Building OpenCV 3.4-rc fails on Visual Studio 2013 and older
Building OpenCV 3.4-rc on Visual Studio 2012 and 2013 fails for two reasons

1. ```inline``` is not supported in C
2. fall safe version of ```snprintf``` is not included in ```libtiff```

- OpenCV => 3.4.0-rc
- Operating System / Platform => Windows 7 Pro 64bit
- Compiler => Visual Studio 2012 and 2013

##### Detailed description
Libtiff problem on VS2013 and earlier

When building OpenCV 3.4.0-rc on Visual Studio 2012 and 2013, following error happens

```
C:\opencv\3rdparty\libtiff\tif_fax3.c(844): error C2054: expected '(' to follow 'inline' [C:\opencv\build\3rdparty\libtiff\libtiff.vcxproj]
C:\opencv\3rdparty\libtiff\tif_fax3.c(845): error C2085: 'find0span' : not in formal parameter list [C:\opencv\build\3rdparty\libtiff\libtiff.vcxproj]
C:\opencv\3rdparty\libtiff\tif_fax3.c(845): error C2143: syntax error : missing ';' before '{' [C:\opencv\build\3rdparty\libtiff\libtiff.vcxproj]
C:\opencv\3rdparty\libtiff\tif_fax3.c(906): error C2054: expected '(' to follow 'inline' [C:\opencv\build\3rdparty\libtiff\libtiff.vcxproj]
C:\opencv\3rdparty\libtiff\tif_fax3.c(907): error C2085: 'find1span' : not in formal parameter list [C:\opencv\build\3rdparty\libtiff\libtiff.vcxproj]
C:\opencv\3rdparty\libtiff\tif_fax3.c(907): error C2143: syntax error : missing ';' before '{' [C:\opencv\build\3rdparty\libtiff\libtiff.vcxproj]
```

I looked in the source code and realized [these](https://github.com/opencv/opencv/blob/3.4.0-rc/3rdparty/libtiff/tif_fax3.c#L843-L844) [two functions](https://github.com/opencv/opencv/blob/3.4.0-rc/3rdparty/libtiff/tif_fax3.c#L905-L906) are violating strict C.  Visual Studio 2013 and earlier requires "strict C" for file which has extension ".c", so, the [```libtiff/tif_fax3.c```](https://github.com/opencv/opencv/blob/3.4.0-rc/3rdparty/libtiff/tif_fax3.c) is treated as "C" file, not "C++".
Thus the ```inline``` keyword is violating the compilation.

For a workaround, I tried two approaches, but either of them end up in another problem of linker.
* Remove inline (i.e. ```inline static int32``` to ```static int32```)
* Replace by ```__inline``` (i.e. ```__inline static int32```)

Now, new problem is the linker saying

```error LNK2019: unresolved external symbol _TIFF_snprintf_f referenced in function _TIFFCreateAnonField```

I googled up a bit and found following points
  * ```_TIFF_snprintf_f``` are defined as ```extern``` [here](https://github.com/opencv/opencv/blob/3.4.0-rc/3rdparty/libtiff/tiffiop.h#L60-L64) in [tiffiop.h](https://github.com/opencv/opencv/blob/3.4.0-rc/3rdparty/libtiff/tiffiop.h)
  * ```_TIFF_snprintf_f``` seems a fail safe implementation when compiler doesn't support API ```snprintf```
  * Unfortunately, the implementation of the fail safe version is missing in OpenCV
     * Looking back in libtiff mirror, [it does exists in the original libtiff](https://gitlab.com/libtiff/libtiff/blob/Release-v4-0-9/port/snprintf.c#L26), but seems it disappeared during the migration.

Before I write a patch, I'd like to ask that I'm doing the write thing.  Are Visual Studio 2012 and 2013 still in the supported compiler list ?

##### Steps to reproduce
* Use Visual Studio 2012 or 2013
* Get OpenCV 3.4-rc source code
* Build with no option, it should 
  * try to build [```libtiff```](https://gist.github.com/tomoaki0705/67fcbafadd96fbe0f936a7bb262721b5#file-log-txt-L225)
  * fail to find [```snprintf```](https://gist.github.com/tomoaki0705/67fcbafadd96fbe0f936a7bb262721b5#file-log-txt-L130)
* Build target ```ALL_BUILD```

### PR DESCRIPTION
**Pull request by tomoaki0705**
close   #10342 
### This pullrequest changes
  - fix compilation error about "inline"
  - fix linker error about "_TIFF_snprintf_f"

### DIFF
--- a/3rdparty/libtiff/CMakeLists.txt
+++ b/3rdparty/libtiff/CMakeLists.txt
@@ -34,31 +34,27 @@ if(NOT MSVC)
   check_include_file(unistd.h    HAVE_UNISTD_H)
 endif()
 
-if(MSVC)
-  set(INLINE_KEYWORD "inline")
-else()
-  # Inspired from /usr/share/autoconf/autoconf/c.m4
-  foreach(inline_keyword "inline" "__inline__" "__inline")
-    if(NOT DEFINED C_INLINE)
-      set(CMAKE_REQUIRED_DEFINITIONS_SAVE ${CMAKE_REQUIRED_DEFINITIONS})
-      set(CMAKE_REQUIRED_DEFINITIONS ${CMAKE_REQUIRED_DEFINITIONS}
-          "-Dinline=${inline_keyword}")
-      check_c_source_compiles("
-          typedef int foo_t;
-          static inline foo_t static_foo() {return 0;}
-          foo_t foo(){return 0;}
-          int main(int argc, char *argv[]) {return 0;}"
-        C_HAS_${inline_keyword})
-      set(CMAKE_REQUIRED_DEFINITIONS ${CMAKE_REQUIRED_DEFINITIONS_SAVE})
-      if(C_HAS_${inline_keyword})
-        set(C_INLINE TRUE)
-        set(INLINE_KEYWORD "${inline_keyword}")
-      endif()
-   endif()
-  endforeach()
+# Inspired from /usr/share/autoconf/autoconf/c.m4
+foreach(inline_keyword "inline" "__inline__" "__inline")
   if(NOT DEFINED C_INLINE)
-    set(INLINE_KEYWORD)
-  endif()
+    set(CMAKE_REQUIRED_DEFINITIONS_SAVE ${CMAKE_REQUIRED_DEFINITIONS})
+    set(CMAKE_REQUIRED_DEFINITIONS ${CMAKE_REQUIRED_DEFINITIONS}
+        "-Dinline=${inline_keyword}")
+    check_c_source_compiles("
+        typedef int foo_t;
+        static inline foo_t static_foo() {return 0;}
+        foo_t foo(){return 0;}
+        int main(int argc, char *argv[]) {return 0;}"
+      C_HAS_${inline_keyword})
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: 3rdparty/libtiff/tif_fax3.c [hunk 1]
```diff
@@ -840,7 +840,7 @@ static	int32 find1span(unsigned char*, int32, int32);
  * table.  The ``base'' of the bit string is supplied
  * along with the start+end bit indices.
  */
-inline static int32
+__inline static int32
```
alalek: Could you fix it [here](https://github.com/opencv/opencv/blob/9de87d96ccfa2f0aeb2ea74e110565e6dfc3f047/3rdparty/libtiff/CMakeLists.txt#L39) ?

tomoaki0705: Thanks for your reply.  Let me do it.
You always surprise me by showing a better solution!

tomoaki0705:
I confirmed that c057d29 fix the build error on both Visual Studio 2012 and 2013

Issue #10110, opened 2017-11-17
About calcErrors and sampleWeights in module ML


LaurentBerger:
Hi,

In trainData a weight per sample can be defined in [create ](https://docs.opencv.org/master/dc/d32/classcv_1_1ml_1_1TrainData.html#a5e0c052f9aadce1f75cddbdbbf9c4f4d)method. Sample Weights are used in backprop, rprop method.
When you call [calcError ](https://docs.opencv.org/master/db/d7d/classcv_1_1ml_1_1StatModel.html#aa6a71b1ee5b7fa0b07b55e77106cda13)sample weigths are not used.
I think that [calcError ](https://github.com/opencv/opencv/blob/master/modules/ml/src/inner_functions.cpp#L92-L124)must use sample weights to give a good error value. Am I wrong?

My problem is that I use simulated annealing to optimize node weight and bias, and I need a global error : calcError. Rprop and backprop use sample weights. Hence if I use calcError and I will not optimize same neural network than in rprop because traindata weights are not the same.

I think calcErrormust be modified. Is it right?

LaurentBerger:
CalcErrors function uses result of predict function. It's OK when ANN_MLP is a classifier but not really good when ANN_MLP is use for regression.  I Think it could be practise to add function setClassifier true nothing is changed, false ANN_MLP is used for regression  and calcerror must use norm

PR #10346, opened 2017-12-18
Calcerror uses now weighted samples

LaurentBerger:
Solves #10110

delete some useless coments in ANN_MLPImpl::settrainmethod


vpisarev:
looks good to me! :+1:

PR #10352, opened 2017-12-19
add imwrite for pbm

### LINKED ISSUE #9791: imwrite doesn't handle pbm

- OpenCV => 3.2.0

##### Detailed description

If you try to save some monochrome image as bitmap ( .pbm P1/P4) you get a greymap (P2/P5) picture at outcome.

##### Steps to reproduce

A simple example:
```python
import numpy as np
import cv2
img = np.array([0, 0, 0, 0]).reshape((2,2))
cv2.imwrite('/tmp/out.pbm', img, (32, 0))
```
And out.pbm looks like this:
```
P2
2 2
255
   0   0
   0   0
```

vinay0410:
resolves #9791

### This pullrequest changes

1. Adds two parameters to imwrite as described below:
a) IMWRITE_PBM_MONOCHROME : Flag to choose, whether to write the image as a BitMap i.e. consisting of only 0's and 1's. Default is 0.
b) IMWRITE_PBM_THRESH_MONOCHROME : Flag to choose the threshold for converting a grayscale image to   a BitMap(Binary Image), which will be written as PBM. It would be ignored if above flag is not set to 1. Defaults to 128 for CV_8U and 32,768 for CV_16U.

2. Now images can be written in PBM format (P1/P4).

Sample Code:
```import cv2```
```import numpy as np```
```img = np.array([0,100,230,30,0,0]).reshape((3,2))```
```cv2.imwrite("out.pbm", img, (33,1))```

where 33 is the number of IMWRITE_PBM_MONOCHROME.

### CODE REVIEW

**Comment by alalek** [comment]
Tests are required for new functionality: take a look in write/read tests.

- rename bitmap => monochrome to avoid naming confusion
- BTW, mode can be handled by extension ".pbm" (".pgm" - gray images; ".ppm" - color)

**Comment by vinay0410** [comment]
Yes, I did use extension ".pbm" without the parameter earlier, But while running test for imgcodecs, a segmentation fault occurred while running read and write tests for ppm.

**Comment by vinay0410** [comment]
@alalek, I would be very happy, if could at least suggest something.
If I use .pbm to decide whether to out put monochrome, then imencode will have to pass an extra argument to for extension to encoder->write as defined [here](https://github.com/opencv/opencv/blob/master/modules/imgcodecs/src/loadsave.cpp#L897)

It's fine if writing to files, but if writing to buffer, then it causes a problem.

File: modules/imgcodecs/test/test_grfmt.cpp [hunk 1]
```diff
@@ -208,6 +216,88 @@ INSTANTIATE_TEST_CASE_P(All, Imgcodecs_ExtSize,
...
+typedef testing::TestWithParam<Size_thresh> Imgcodecs_Image_pbm_thresh;
+
+
+TEST_P(Imgcodecs_Image_pbm_thresh, read_write_pbm)
+{
+    const String ext = "pbm";
+    const Size size = get<0>(GetParam());
+    const int thresh = get<1>(GetParam());
+    const string full_name = cv::tempfile(ext.c_str());
+    const string _name = cvtest::TS::ptr()->get_data_path() + "readwrite/marbles.pbm";
```
alalek: Test data is missing.
You should put testdata into opencv_extra repository with same branch name (`write_pbm`) and create PR

modules/imgcodecs/test/test_grfmt.cpp
```diff
@@ -152,8 +152,15 @@ TEST_P(Imgcodecs_ExtSize, write_imageseq)
 
         Mat img_gt(size, CV_MAKETYPE(CV_8U, cn), Scalar::all(0));
         circle(img_gt, center, radius, Scalar::all(255));
-        ASSERT_TRUE(imwrite(filename, img_gt));
-
+        if (ext == ".pbm")
+        {
+          ASSERT_TRUE( imwrite(filename, img_gt, vector<int>(IMWRITE_PBM_MONOCHROME,1) ) );
```
alalek: > vector<int>(IMWRITE_PBM_MONOCHROME,1)

It is wrong. Refer to C++ documentation.

modules/imgcodecs/include/opencv2/imgcodecs.hpp
```diff
@@ -79,19 +79,21 @@ enum ImreadModes {
...
+       IMWRITE_JPEG_LUMA_QUALITY    = 5,  //!< Separate luma quality level, 0 - 100, default is 0 - don't use.
+       IMWRITE_JPEG_CHROMA_QUALITY  = 6,  //!< Separate chroma quality level, 0 - 100, default is 0 - don't use.
+       IMWRITE_PNG_COMPRESSION      = 16, //!< For PNG, it can be the compression level from 0 to 9. A higher value means a smaller size and longer compression time. If specified, strategy is changed to IMWRITE_PNG_STRATEGY_DEFAULT (Z_DEFAULT_STRATEGY). Default value is 1 (best speed setting).
+       IMWRITE_PNG_STRATEGY         = 17, //!< One of cv::ImwritePNGFlags, default is IMWRITE_PNG_STRATEGY_RLE.
+       IMWRITE_PNG_BILEVEL          = 18, //!< Binary level PNG, 0 or 1, default is 0.
+       IMWRITE_PXM_BINARY           = 32, //!< For PPM, PGM, or PBM, it can be a binary format flag, 0 or 1. Default value is 1.
+       IMWRITE_PBM_MONOCHROME       = 33, //!< For PBM, it can be a BitMap Flag, i.e. whether to write image in a BitMap Format, 0 or 1. Default value is 0.
+       IMWRITE_PBM_THRESH_MONOCHROME= 34, //!< For PBM, threshold above which pixels will be written as white(0's) and black(1's). Default value is 128 CV_8U and 32768 for CV_16U.
```
alalek: This definitely should not be in imgcodecs library. Lets keep it as simple as possible.
Use values equal to 0 as binary 0, and values non equal to 0 as binary 1.
OpenCV already has special processing functions. For example, for threshold 128:
```
imwrite("image.pbm", image > 128);
```

alalek:
@vinay0410 I have pushed commit to support PBM without additional parameters (by extension) + several fixes. Please take a look.

vinay0410:
Well, Thank you, for helping me, since I'm just a newbie in this repository. It works great now, Thanks again

Issue #7263, opened 2016-09-09
Suggestions for future GSoCs


StevenPuttemans:
Just want to point out that it might be interesting to get the opinion of the community into new subjects for GSoC 2017 and collect them somewhere central. We could use this issue for a growing list of suggestions.

Let me start by adding one

- Fixing the terribly broken down video backend (both writer and reader). There is simply no guarantee that stuff works on different systems for now, codecs are poorly supported and the huge amount of issues on the Q&A clearly show we need some fixing in this part of the library.

PkLab:
:+1: @StevenPuttemans (did you see the [new doc](http://docs.opencv.org/master/d0/da7/videoio_overview.html) ?)

- Another common question on the Q&A is about GUI. In addiction most of Windows user is moving to From Win32 MFC to .NET/WPF or Managed code. More in general the HighGui module is just to play and some wrapper is always needed to get a minimal usable interface, even if Qt, wxWin,...

  Sure GUI is not the core task for OpenCV but GUI is terribly useful or needed. In addiction I think that a comfortable way to use OpenCV with GUI from C#, VB, Delphi,...should exists.

  A crude idea on the fly: instead to open OpenCV to others, encapsulate it and user calling code code into any standard process provided with common interface.

  Thinking at `user application = (opencv + user vision code) + user GUI code` , we could have:
  - a context (interface, .a/.dll ,...) where user can encapsulate his/her pure C++ code that calls/uses OpenCV
  - any easy/standard way (pipe, shared mem, memfile, socket,...) to exchange informations as params or images with user GUI.

kinchungwong:
To second @PkLab, I have used multiple processes and shared memory to workaround webcam video issues. I start up one child process which grabs frame from webcam, copy that into shared memory, and use Win32 event to communicate to the parent process that a frame is ready for processing.

With some modest logic, it should be able to implement double buffering, page flipping, and swap chain, etc.

However, I doubt that beginning users will be able to implement or use such mechanisms on their own. There is a minimum level of required knowledge just to use it. In order to make a dedicated helper process, there needs to be a separate executable file (exe) which needs to be bundled with OpenCV and/or the executable that the OpenCV user is developing. This may be too complicated or not acceptable to some users though.

Finally, whenever using blocking synchronization, remember to implement it in a way that abandonment (e.g. the helper process was killed by user or crashed on its own) results in an exception thrown at the waiter, so that it will not become a zombie thread.

[...]

PR #10354, opened 2017-12-19
Add Python sample to show extrinsics from camera calibration

catree:
### This pullrequest changes

It could be useful to have something similar to Matlab [`showExtrinsics`](https://www.mathworks.com/help/vision/ref/showextrinsics.html) function to display the position of the calibration board w.r.t to the camera during the calibration step. See also this [feature suggestion](https://github.com/opencv/opencv/issues/7263#issuecomment-314168492).

When calibrating:
`./cpp-example-calibration left_imagelist.xml -w=9 -h=6 -s=0.025 -oe`

To display the extrinsics:
`python camera_calibration_show_extrinsics.py --calibration ../data/left_intrinsics.yml`
extrinsics_1

`python camera_calibration_show_extrinsics.py --calibration ../data/left_intrinsics.yml --patternCentric`
extrinsics_2

Note: 

- this work currently only for the `calibration.cpp` sample and not with `stereo_calib.cpp` sample.
- Matplotlib and Numpy are required


vpisarev:
:+1:

PR #10357, opened 2017-12-19
Resolved fatal error during initial compilation

Xenos5282:
fatal error is encountered when compiling OpenCV for Python (in Ubuntu 16.04) which is 'fatal error hdf5.h no such file or directory'.
Changes made looks for the HDF5 package and finds and loads settings from the external project.


alalek:
"#include <hdf5.h>" must be removed from public header of OpenCV module instead.
Please provide complete error message from compiler.

PR #10358, opened 2017-12-19
Add images stitching demo in the homography tutorial

catree:
### This pullrequest changes

Show how the homography can also be applied in the case of a pure rotational movement of the camera for a generic scene.

See also #10304


vpisarev:
:+1:

PR #10362, opened 2017-12-20
add HAL for FAST

ElenaGvozdeva:
### This pullrequest changes

Add HAL interface for FAST algorithm for corner detection. Call only for FAST_9_16 and threshold <= 20.

### DIFF
--- a/modules/features2d/src/fast.cpp
+++ b/modules/features2d/src/fast.cpp
@@ -45,6 +45,7 @@ The references are:
 #include "fast.hpp"
 #include "fast_score.hpp"
 #include "opencl_kernels_features2d.hpp"
+#include "hal_replacement.hpp"
 #include "opencv2/core/hal/intrin.hpp"
 
 #include "opencv2/core/openvx/ovx_defs.hpp"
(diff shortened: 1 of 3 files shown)


modules/features2d/src/fast.cpp
```diff
@@ -417,6 +418,65 @@ static bool openvx_FAST(InputArray _img, std::vector<KeyPoint>& keypoints,
 
 #endif
 
+namespace hal
+{
+inline int FAST(cv::Mat& src, std::vector<KeyPoint>& keypoints, int threshold, bool nonmax_suppression, int type)
+{
+    if (type != FastFeatureDetector::TYPE_9_16 || threshold > 20)
+        return CV_HAL_ERROR_NOT_IMPLEMENTED;
```
ElenaGvozdeva: @mshabunin, maybe we need to add `FAST_5_8` and `FAST_7_12` or FAST with parameter `type`, and check it inside HAL. What do you think?

mshabunin: @ElenaGvozdeva , yes it will be better to define available types in `features2d/hal/interface.h` file (see [imgproc/hal/interface.h](https://github.com/opencv/opencv/blob/master/modules/imgproc/include/opencv2/imgproc/hal/interface.h) for example).

modules/features2d/src/fast.cpp
```diff
@@ -417,6 +418,65 @@ static bool openvx_FAST(InputArray _img, std::vector<KeyPoint>& keypoints,
 
 #endif
 
+namespace hal
+{
+inline int FAST(cv::Mat& src, std::vector<KeyPoint>& keypoints, int threshold, bool nonmax_suppression, int type)
```
mshabunin: It will be better to remove this function from the `hal::` namespace, we use it for HAL public interface (defined in [hal.hpp](https://github.com/opencv/opencv/blob/master/modules/imgproc/include/opencv2/imgproc/hal/hal.hpp)).

You can mark it `static` and rename to something like `hal_FAST`.

modules/features2d/src/fast.cpp
```diff
@@ -425,6 +485,9 @@ void FAST(InputArray _img, std::vector<KeyPoint>& keypoints, int threshold, bool
     CV_OCL_RUN(_img.isUMat() && type == FastFeatureDetector::TYPE_9_16,
                ocl_FAST(_img, keypoints, threshold, nonmax_suppression, 10000));
 
+    cv::Mat img = _img.getMat();
+    CALL_HAL(fast, hal::FAST, img, keypoints, threshold, nonmax_suppression, type);
```
mshabunin: Is it possible to integrate [carotene implementation](https://github.com/opencv/opencv/blob/dcdd6af5a856826fe62c95322145731e702e54c5/3rdparty/carotene/include/carotene/functions.hpp#L1913-L1916) using this interface?

ElenaGvozdeva: No, we can only do something like this:

* add one more interface for FAST

```cpp
inline int hal_ni_FAST(const uchar* src_data, size_t src_step, int width, int height,
                       uchar* keypoints_data, size_t* keypoints_size,
                       int threshold, bool nonmax_suppression, int type) { return CV_HAL_ERROR_NOT_IMPLEMENTED; }

//! @cond IGNORED
#define cv_hal_FAST hal_ni_FAST
//! @endcond
```
* try to call this interface one by one
```cpp
    cv::Mat img = _img.getMat();
    CALL_HAL(fast_dense, hal::FAST, img, keypoints, threshold, nonmax_suppression, type);

    size_t keypoints_size;
    CALL_HAL(fast, cv_hal_FAST, img.data, img.step, img.cols, img.rows,
             (uchar*)(keypoints[0]), &keypoints_size, threshold, nonmax_suppression, type);
```

@vpisarev, if it isn't acceptable, could we discuss it one more time?

vpisarev:
@ElenaGvozdeva, this solution is probably fine for now. But I wonder what happens if you raise the threshold above 20? Your function becomes slower? Does it have anything to do with the multi-pass approach that you've described before NY? Can the HAL function be adjusted so that it calls some callback to dynamically reallocate std::vector instead of using mask? I suspect that in the case of a sparse feature set a lot of time is spent on the mask traversal

vpisarev:
:+1:

PR #8511, opened 2017-04-03
export img decoders/encoders

brian-armstrong-discord:
This pull request changes the image codecs API so that `findDecoder` and `findEncoder` are available to users. This allows users to inspect the header of an image before decoding it so that they can decide whether they want to decode the image before they do so. For example, the image might describe itself as being 1000000 x 1000000 pixels, in which case openCV will go ahead and allocate an extremely large buffer to hold this image. This new API gives the user the ability to decide they don't want to do that.

In addition, this patch attempts to comment the API a little and make it safer to use by doing bounds checking in `readData` so that a too-small buffer given to `readData` will not yield some kind of bad write outside array bounds.

PR #10367, opened 2017-12-20
imwrite multipage TIFF files

### LINKED ISSUE #8729: Create write version of of cv::imreadmulti, cv:imwritemulti
Currently we have [cv::imreadmulti](https://github.com/opencv/opencv/blob/master/modules/imgcodecs/src/loadsave.cpp#L585) which allows for multi-image Tiff files to be read. I very much like this. Probably the most common use case is to read in a bunch of related images for processing.

It is also useful to write out a vector of images to a single Tiff file. 

My feature request is simple. Signature of proposed addition to [loadsave.cpp](https://github.com/opencv/opencv/blob/master/modules/imgcodecs/src/loadsave.cpp):

`bool imwritemulti(const String& filename, std::vector<Mat>& mats, const std::vector<int>& params)`

`mats` would be written out as a multi-image file. Currently the only multi-image format supported by OpenCV is TIff.

savuor:
resolves #8729

### This pullrequest changes
The old PR #8913 re-created, merge conflicts are fixed

### DIFF
--- a/modules/imgcodecs/src/grfmt_base.cpp
+++ b/modules/imgcodecs/src/grfmt_base.cpp
@@ -127,6 +127,11 @@ bool BaseImageEncoder::setDestination( std::vector<uchar>& buf )
     return true;
 }
 
+bool BaseImageEncoder::writemulti(const std::vector<Mat>&, const std::vector<int>& )
+{
+    return false;
+}
+
 ImageEncoder BaseImageEncoder::newEncoder() const
 {
     return ImageEncoder();
--- a/modules/imgcodecs/src/grfmt_base.hpp
+++ b/modules/imgcodecs/src/grfmt_base.hpp
@@ -101,6 +101,7 @@ public:
     virtual bool setDestination( const String& filename );
     virtual bool setDestination( std::vector<uchar>& buf );
     virtual bool write( const Mat& img, const std::vector<int>& params ) = 0;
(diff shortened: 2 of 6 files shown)


alalek:
I believe OpenCV should handle most universal cases and "multipage" images case is not one of this. Also it is TIFF only (and perhaps animated GIFs). Image by definition is not multipage (see wiki).
For sequence of images we have VideoCapture/VideoWriter interface, so it is better to deprecate existed imreadmulti API and move multipage TIFF stuff there.

/cc @vpisarev 
see #8511

vpisarev:
actually, I like how it's handled. Without introducing extra function we added extra functionality. Of course, it needs to be properly documented. But the same technique can be used even for animated gifs

vpisarev:
:+1:

rounakskm:
Hi all,
I am fairly new to opencv and am having trouble finding this function.
Could anyone please tell me how to write  a `vector<cv::Mat> `-- where each Mat is a Tiff image -- to one stacked Tiff file?
An example would be much appreciated.

savuor:
@rounakskm Although this place is not for questions and answers (please use answers.opencv.org) I can help you as an exception. You can find an example of usage [here](https://github.com/opencv/opencv/blob/master/modules/imgcodecs/test/test_tiff.cpp#L265)

PR #10370, opened 2017-12-21
update dnn convolution auto-tuning

pengli:
```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/src/ocl4dnn/include/ocl4dnn.hpp
+++ b/modules/dnn/src/ocl4dnn/include/ocl4dnn.hpp
@@ -258,6 +258,12 @@ class OCL4DNNConvSpatial
                                  int lx, int ly, int lz,
                                  bool swizzle, bool nullLocal);
         void generateTunerItems(std::vector< cv::Ptr<tunerParam> > &tunerItems);
+        void generate_dwconv_tuneritems(std::vector< cv::Ptr<tunerParam> > &tunerItems,
+                                        int blockM, int blockK, int blockN);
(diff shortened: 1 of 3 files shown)


pengli:
@vpisarev , @alalek , Please help review the code and I hope it can catch up 3.4 release. Thanks.

samples/dnn/ssd_mobilenet_object_detection.cpp
```diff
@@ -122,6 +122,27 @@ int main(int argc, char** argv)
...
+                                      Size(inWidth, inHeight), meanVal, false); //Convert Mat to batch of images
+        //! [Prepare blob]
+
+        //! [Set input blob]
+        net.setInput(inputBlob, "data"); //set the network input
+        //! [Set input blob]
+
+        //! [Make forward pass]
+        Mat detection = net.forward("detection_out"); //compute output
+        //! [Make forward pass]
```
alalek: Not sure why we need this, but this can be simplified and placed into the main loop:
```.diff
-for(;;)
+for(size_t N = 0; ; N++)
 {
    ...
     Mat detection = net.forward("detection_out"); //compute output
     //! [Make forward pass]

+    if (N == 0) continue; // warmup on first frame
     ...
 }
...
```

BTW, Comments like these "//! [Set input blob]" are special and should be unique in file

pengli: yes，my purpose is to ignore first forward and not count its inference time，this seems not necessary in this sample, forward time is count frame by frame, so I just reverted this change

PR #10371, opened 2017-12-21
Optimize 2 opencv 2 files (merge.cpp/split.cpp) with PowerPC VSX intrinsics.

justdoitqd:
### This pullrequest changes

Optimize powerpc opencv 2 files (merge.cpp/split.cpp) with VSX intrinsics.

```
force_builders=Custom
```


alalek:
Thank you for contribution!
Adding direct usage of raw intrinsics in "processing" code is not allowed anymore.
This code should be ported on OpenCV's universal intrinsics (SIMD128): see this: #9753
Missing intrinsics should be added for C++, x86, ARM, VSX backends (with tests).

vpisarev:
@justdoitqd, I agree with @alalek, PowerPC is quite exotic for us and many other people platform, so those platform-specific branches are at the high risk to become abandoned and poorly tested. We extensively use and promote universal intrinsics nowadays, which is the right way to write vectorized code once and then compile and run it everywhere.

modules/core/src/merge.cpp
```diff
@@ -285,6 +346,16 @@ merge_( const T** src, T* dst, int len, int cn )
             int inc_i = (sizeof(T) == 8)? 1: 16/sizeof(T);
             int inc_j = 2 * inc_i;
 
+            VMerge2<T> vmerge;
+            for( ; i < len - inc_i; i += inc_i, j += inc_j)
+                vmerge(src0 + i, src1 + i, dst + j);
+        }
+#elif CV_VSX
```
seiko2plus: replace VMerge with universal intrinsics
```
#elif CV_VSX
if (hasSIMD128() && cn == 2)
{
    int inc_i = 16 / sizeof(T);
    int inc_j = 2 * inc_i;

    for (; i < len - inc_i; i += inc_i, j += inc_j)
        v_store_interleave(dst + j, v_load(src0 + i), v_load(src1 + i));
}
```
add int64 support for v_store_interleave in [intrin_vsx.hpp](https://github.com/justdoitqd/opencv/blob/20722a992215d46496fbf5abfed75dc03e18c4fa/modules/core/include/opencv2/core/hal/intrin_vsx.hpp#L334)
````
OPENCV_HAL_IMPL_VSX_INTERLEAVE(uint64, v_uint64x2)
OPENCV_HAL_IMPL_VSX_INTERLEAVE(int64, v_int64x2)
````

File: modules/core/src/merge.cpp [hunk 2]
```diff
@@ -105,6 +105,67 @@ MERGE4_KERNEL_TEMPLATE(VMerge4, ushort,  uint16x8x4_t, vld1q_u16, vst4q_u16);
 MERGE4_KERNEL_TEMPLATE(VMerge4, int   ,   int32x4x4_t, vld1q_s32, vst4q_s32);
 MERGE4_KERNEL_TEMPLATE(VMerge4, int64 ,   int64x1x4_t, vld1_s64 , vst4_s64 );
 
+#elif CV_VSX
```
seiko2plus: remove VMerge, no need to add extra templates over universal intrinsics

justdoitqd:
@seiko2plus  Ah.. My bad... I forgot to pull the related change regarding "OPENCV_HAL_IMPL_VSX_INTERLEAVE(int64, v_int64x2)" in my branch in.  And I should rebuilt on the commit branch.  Thanks for pointing it out.  I added the missing commit.

@alalek @vpisarev 
As seiko2plus mentioned, the change is already using general intrinsics v_store_interleave/v_load_deinterleave/v_load/v_store. They are already implemented for SSE/NEON.

Please let me know if I still miss anything.

vpisarev:
@justdoitqd, @seiko2plus, most of the time when we add branches with universal intrinsics, the corresponding native NEON/SSE branches can be removed. Of course, in some rare cases that may result in some performance degradation, but I would rather solve it by introducing new, higher-level universal intrinsics, or even by introducing small chunks of SSE/NEON-specific code (like here: https://github.com/opencv/opencv/blob/master/modules/dnn/src/layers/pooling_layer.cpp#L418). But while the latter approach might be fine for x86-specific code, because we and our users test it all the time, it may be not a good solution for PPC, because of the much smaller user base.

seiko2plus:
@alalek,  @vpisarev, Thank you for the clarification.

vpisarev:
@justdoitqd, as I see, the latest version still uses the macro `CV_VSX`, that is, even if the actual code is using universal intrinsics, the branches are not included into the regular x86/x64/ARM builds. To make those branches universally available, I suggest to use `CV_SIMD128` macro instead.

justdoitqd:
@vpisarev.  I see. I will try to find some timeslot to get this done to use CV_SIMD128 covering NEON/VSX.  However my problem is that I don't have a ARM server to test whether ARM will work. Is it OK for me to only test PowerPC since the code will be common?

vpisarev:
@justdoitqd, yes, it's ok. Also, you likely have x86/x64 machine somewhere, e.g. your laptop, so you can test it as well. Then, as soon as you switch from CV_VSX to CV_SIMD128, our buildbot will automatically test the code on Intel/AMD and ARM CPUs that we have in our battery of test machines.

edelsohn:
@vpisarev What is the procedure to add a VM at OSUOSL to the OpenCV buildbot?

vpisarev:
sorry for a long delay with reply; there is no well-defined procedure of adding new build slaves to our buildbot, because each new build slave will take extra resources from the OpenCV core team to support it. For now you may use the "Custom" build slave with special magic commands, see the top-most comment:

<pre>
force_builders=Custom
</pre>

vpisarev:
as we agreed to go the universal intrinsics route, let's close this PR

PR #10381, opened 2017-12-21
Update resize.cpp

sturkmen72:
workaround for https://github.com/opencv/opencv_contrib/issues/1490

### This pullrequest changes
i know this might be not a correct solution but may be it is better than  current before the release ( maybe developers have no time for correct solution )


vpisarev:
@sturkmen72, thank you for the patch! @terfendail has just submitted 2 patches that solve this problem in a better way: 1) we switched back from `INTER_LINEAR_EXACT` to `INTER_LINEAR` in some places in contrib, including dpm module and 2) now `INTER_LINEAR_EXACT` automatically falls back to `INTER_LINEAR` in the case of floating-point (32F, 64F) input.

PR #10386, opened 2017-12-22
Disabled universal intrinsic implementation for bit-exact 3-chan resize

terfendail:
### This pullrequest changes
Disabled universal intrinsic based implementation for bit-exact resize of 3-channel images


vpisarev:
:+1:

PR #10387, opened 2017-12-22
Added performance tests for linear resize of 2 and 3-channel images

terfendail:
### This pullrequest changes
Added performance tests for linear resize of 2 and 3-channel images


vpisarev:
:+1:

PR #10388, opened 2017-12-22
Update resize inline comments

cezheng:
Reading through the implementation, I feel this line of comment is not consistent with the actually code, so this is for correcting it.

### This pullrequest changes

Just one inline comment.

### DIFF
--- a/modules/imgproc/src/resize.cpp
+++ b/modules/imgproc/src/resize.cpp
@@ -3945,7 +3945,7 @@ void resize(int src_type,
         if( interpolation == INTER_LINEAR && is_area_fast && iscale_x == 2 && iscale_y == 2 )
             interpolation = INTER_AREA;
 
-        // true "area" interpolation is only implemented for the case (scale_x <= 1 && scale_y <= 1).
+        // true "area" interpolation is only implemented for the case (scale_x >= 1 && scale_y >= 1).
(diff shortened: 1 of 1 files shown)


vpisarev:
nope, that's wrong change. when destination image is smaller than the source one, which is where area interpolation is mostly usable and applicable, both scale_x and scale_y are less than or equal to 1. So the original comment is correct whereas the new one is wrong

cezheng:
After reading your reply, I'm more convinced the comment is wrong. @vpisarev 

If you look at the if clause under the comments, it is actually

```cpp
if( interpolation == INTER_AREA && scale_x >= 1 && scale_y >= 1 )
```
and only inside this if, are the actually area resize implementation called [here](https://github.com/opencv/opencv/blob/master/modules/imgproc/src/resize.cpp#L3861) and [here](https://github.com/opencv/opencv/blob/master/modules/imgproc/src/resize.cpp#L3901).

and if you look at scale_x and scale_y's definitions, it is ssize / dsize(inv_scale_x and inv_scale_y definitions [here](https://github.com/opencv/opencv/blob/master/modules/imgproc/src/resize.cpp#L3678)), which means if the destination image is smaller than the source one, scale_x and scale_y will be greater than 1.0.

Therefor the original comment is wrong. But I don't find a way to reopen this PR.

PR #10389, opened 2017-12-22
Remove redundant return variable in tutorials

wxzs5:
### This pullrequest changes

cv.solvePnP() can return three variables instead of four.
https://docs.opencv.org/master/d9/d0c/group__calib3d.html#ga549c2075fac14829ff4a58bc931c033d


wxzs5:
OK, got it!

PR #10395, opened 2017-12-22
Suggestion: add link to Android Studio package

floe:
I've spent a bit of work to make a "proper" Android Studio package from the Android SDK, which is available at https://github.com/floe/opencv-android. Not sure if the README is actually the right place for this, just wanted to suggest that this link may be useful. I've also enabled JitPack for this repo, so now the OpenCV SDK can be pulled into any Android Studio project by adding two lines to the Gradle build files.


alalek:
Root README file is not a right place for this.
I believe it is better to post your link here: #5003

floe:
OK, will do that, thanks.

PR #10398, opened 2017-12-22
ml: simplify interfaces of SimulatedAnnealingSolver

alalek:
- no need to wrap it into cv::Algorithm
- it uses custom non-cv::Mat objects (optimization system)

### DIFF
--- a/modules/ml/include/opencv2/ml.hpp
+++ b/modules/ml/include/opencv2/ml.hpp
@@ -1912,74 +1912,48 @@ public:
 *                                   Simulated annealing solver                             *
 \****************************************************************************************/
 
-/** @brief The class defines interface for system state used in simulated annealing optimization algorithm.
+#ifdef CV_DOXYGEN
+/** @brief This class declares example interface for system state used in simulated annealing optimization algorithm.
 
-@cite Kirkpatrick83 for details
+@note This class is not defined in C++ code and can't be use directly - you need your own implementation with the same methods.
 */
-class CV_EXPORTS SimulatedAnnealingSolverSystem
+struct SimulatedAnnealingSolverSystem
 {
-protected:
-    inline SimulatedAnnealingSolverSystem() {}
-public:
-    virtual ~SimulatedAnnealingSolverSystem() {}
-
     /** Give energy value for a state of system.*/
-    virtual double energy() const = 0;
+    double energy() const;
     /** Function which change the state of system (random pertubation).*/
-    virtual void changeState() = 0;
+    void changeState();
     /** Function to reverse to the previous state. Can be called once only after changeState(). */
-    virtual void reverseState() = 0;
+    void reverseState();
 };
+#endif // CV_DOXYGEN
 
 /** @brief The class implements simulated annealing for optimization.
- *
+
 @cite Kirkpatrick83 for details
+
+@param solverSystem optimization system (see SimulatedAnnealingSolverSystem)
+@param initialTemperature initial temperature
+@param finalTemperature final temperature
+@param coolingRatio temperature step multiplies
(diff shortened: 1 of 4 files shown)


modules/ml/include/opencv2/ml.hpp
```diff
@@ -1912,75 +1912,47 @@ class CV_EXPORTS_W ANN_MLP_ANNEAL : public ANN_MLP
...
-    /** Set initial temperature of simulated annealing procedure.
-    @param x new initial temperature. x\>0
-    */
-    void setInitialTemperature(double x);
-    /** Set final temperature of simulated annealing procedure.
-    @param x new final temperature value. 0\<x\<initial temperature
-    */
-    void setFinalTemperature(double x);
-    /** Get final temperature of simulated annealing procedure. */
-    double getFinalTemperature();
```
LaurentBerger: Can we get final temperature after stopped algorithm? To restart algorithm final temperature is necessary

alalek: yes, see "lastTemperature" optional CV_OUT parameter

BTW, There are no automatic bindings, because custom class SimulatedAnnealingSolverSystem is used.

alalek: @LaurentBerger Do you have any objections about interface update? Does existed interface mentioned in your [book](https://opencv.org/books.html)?

LaurentBerger: That's ok with CV_OUT. I don't use Simulated Annealing in my book so there is no problem.

There is only problem it is about regression in class ann_mlp. RPROP optimize networks : it is regression. SimulatedAnnealingANN_MLP uses calcError to optimize network and it is not regression, it is a classifier (except when there is only neuron in output layer) . It is not possible in interface to set classifier to true  (=regression).
When it is regression and there is more than 2 neurons in output layer SimulatadeAnnealingANN_MLP calcerror must give this results : 

float val = s.predict(sample, respRow);
err += norm(respRow - responses.row(si));

modules/ml/include/opencv2/ml/ml.inl.hpp
```diff
@@ -0,0 +1,60 @@
...
+     cv::RNG& rngEnergy
+)
+{
+    CV_Assert(finalTemperature > 0);
+    CV_Assert(initialTemperature > finalTemperature);
+    CV_Assert(iterationsPerStep > 0);
+    CV_Assert(coolingRatio < 1.0f);
+    double Ti = initialTemperature;
+    double previousEnergy = solverSystem.energy();
+    int exchange = 0;
```
LaurentBerger: May be it is better to set exchange to 1 because when network is already trained a second call to run will set trained to false if there is no exchange

modules/ml/src/ann_mlp.cpp
```diff
@@ -1075,14 +944,9 @@ class ANN_MLPImpl : public ANN_MLP_ANNEAL
...
-        t.setFinalTemperature(params.finalT);
-        t.setInitialTemperature(params.initialT);
-        t.setCoolingRatio(params.coolingRatio);
-        t.setIterPerStep(params.itePerStep);
+        SimulatedAnnealingANN_MLP s(*this, trainData);
         trained = true; // Enable call to CalcError
-        int iter =  t.run();
+        int iter = simulatedAnnealingSolver(s, params.initialT, params.finalT, params.coolingRatio, params.itePerStep, NULL, params.rEnergy);
         trained =false;
         return iter;
```
alalek: @LaurentBerger I will replace this return to:
```
return iter + 1; // ensure that 'train()' call is always successful
```

alalek:
:+1:

PR #10404, opened 2017-12-23
build: fix MSVS2010 build error

alalek:
> opencv\modules\dnn\src\layers\prior_box_layer.cpp(311): error C2668: 'cv::pow' : ambiguous call to overloaded function


dkurt:
:+1:

PR #10407, opened 2017-12-23
Added a Hough transformation with point data set as input

take1014:
I am developing an automatic driving technique at an automobile company.
I am developing image recognition technology that requires robustness, so I am receiving the benefits of opencv when doing the verification. Hough transform is a good example.
In addition to performing Hough transform from the input image, if we can perform Hough transform from the set of points, we thought that we can verify the effect of Hough transform more flexibly.

Thank you for reading the entire message.

### DIFF
--- a/modules/imgproc/include/opencv2/imgproc.hpp
+++ b/modules/imgproc/include/opencv2/imgproc.hpp
@@ -2150,6 +2150,12 @@ Must fall between 0 and max_theta.
 @param max_theta For standard and multi-scale Hough transform, maximum angle to check for lines.
 Must fall between min_theta and CV_PI.
  */
+// Classical Hough Transform
+struct LinePolar
+{
+    float rho;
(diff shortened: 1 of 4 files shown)


alalek:
Thank you for contribution!

1.
> cvHoughUsingSetOfPoints

Legacy OpenCV C API  (cv* calls) is not supported anymore. It is deprecated and will be removed. Adding of new functions should be avoided.

2. New functionality should have tests.

take1014:
Thank you for your reply.

>New functionality should have tests.
Sorry, I was pushing something different from the tested source code
(I am not accustomed to using github and made a mistake.)

I have a tested source code, so I want to commit , but
> OpenCV C API (cv * calls) is not supported anymore
->Oh....I was short of studying....

After studying opencv's rules and github firmly, I retry.

I appreciate your kind response.

Thank you.

PR #10412, opened 2017-12-25
Update to add window position and size retrieval to HighGUI

GregoryMorse:
Add read only property retrieval (window width, height, position x and y) for enhanced rendering capabilities and more sophisticated research tools

### DIFF
--- a/modules/highgui/include/opencv2/highgui.hpp
+++ b/modules/highgui/include/opencv2/highgui.hpp
@@ -197,11 +197,7 @@ enum WindowPropertyFlags {
        WND_PROP_AUTOSIZE     = 1, //!< autosize property      (can be WINDOW_NORMAL or WINDOW_AUTOSIZE).
        WND_PROP_ASPECT_RATIO = 2, //!< window's aspect ration (can be set to WINDOW_FREERATIO or WINDOW_KEEPRATIO).
        WND_PROP_OPENGL       = 3, //!< opengl support.
-       WND_PROP_VISIBLE      = 4, //!< checks whether the window exists and is visible
-       WND_PROP_RO_WIDTH     = 5, //!< read only field returning pixel width of window
-       WND_PROP_RO_HEIGHT    = 6, //!< read only field returning pixel height of window
-       WND_PROP_RO_XPOS      = 7, //!< read only field returning absolute X position of window
-       WND_PROP_RO_YPOS      = 8  //!< read only field returning absolute Y position of window
+       WND_PROP_VISIBLE      = 4  //!< checks whether the window exists and is visible
      };
 
 //! Mouse Events see cv::MouseCallback
@@ -445,6 +441,16 @@ The function getWindowProperty returns properties of a window.
 @sa setWindowProperty
  */
 CV_EXPORTS_W double getWindowProperty(const String& winname, int prop_id);
+
+/** @brief Provides rectangle of image in the window.
+
+The function getWindowImageRect returns the client screen coordinates, width and height of the image rendering area.
+
+@param winname Name of the window.
+
+@sa resizeWindow moveWindow
+ */
+CV_EXPORTS_W Rect cv::getWindowImageRect(const String& winname);
 
 /** @brief Sets mouse handler for the specified window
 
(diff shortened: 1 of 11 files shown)


alalek:
Thank you for contribution!
Could you add some test for this functionality in highgui module?

BTW, Do we need to process/flush UI messages queue (in terms of Win32) to put window on its final position?

GregoryMorse:
As for Windows, since this must be called from the UI thread (at least in the spirit of the rest of the current code, as only GTK appears to generalize this to arbitrary thread), then GetWindowRect will cause a SendMessage with a WM_RECT, which would cause the message queue to clear/flush while it blocked to process so there is no issue.  I am setting up for the test of the code now.  But I realize a problem with this design.  It is not useful without the trackbar height.  The idea of drawing an image to the exact window size to avoid the scaling requires precise knowledge of the image size, not just the effective window size.  And the trackbar height is the only missing link here if one is present.  I added code for Carbon and Cocoa, and WinRT would be relatively easy to add, though several other calls and features for WinRT were also not added so its questionable how useful this would be without them.

alalek:
> GetWindowRect

Perhaps GetClientRect may help:
https://stackoverflow.com/questions/7561049/what-is-the-difference-between-getclientrect-and-getwindowrect-in-winapi

I believe, for the first version it is fine to support windows without additional controls only. And return -1 in cases of custom controls.

GregoryMorse:
Yes you are absolutely right - further the StretchBlt call which gives the computation of the image size is using this.  In fact the toolbar should be irrelevant in all contexts if this is done properly.  The screen coordinates of the upper left of the image display location is largely a novelty to include but is possible.  The dimensions are very useful.

GregoryMorse:
Next attempt at working and careful solution for cross platform proper computation of the image screen position and size on the window.

GregoryMorse:
By changing the test code to call the new code, the automatic build checks are failing, but it seems to have nothing to do with the changes (Android and iOS only):
POST git-upload-pack (947 bytes)
error: RPC failed; result=56, HTTP code = 0
fatal: The remote end hung up unexpectedly

alalek:
@GregoryMorse Does this patch ready for final review?

GregoryMorse:
Yes it appears ready though I must admit not fetching the width and height atomically is largely already a disadvantage for the very precautious app developer.  But the change has even been hand verified in the debugger on Windows.  Also the X and Y positions of the upperleft screen position of the image are hard to reason or test about their exact correctness and their usefulness is questionable.  But the change compiles, and only is used for those who query such info, with proper error checking at any rate.  So final review is okay.

alalek:
> width and height atomically

I believe it is natural to have `bool cv::getWindowRect(const cv::String& window_name, CV_OUT cv::Rect& rect)` function instead of read-only properties.

BTW, C++  way is to use `cv::Rect cv::getWindowRect(const cv::String& window_name)` with exception in case of error/not supported (like trackbar) - exception with clear message is more useful than "bool" result.

getWindowRect (whole image area with title) or getWindowImageRect (client area only for image painting)?

P.S. What is about multi-monitor configurations? (need to check implementation of `cv::moveWindow()` if it is supported)

vpisarev:
@GregoryMorse, thanks for the contribution! looks useful!

GregoryMorse:
It is final now with the change made, returning errors as cv::Rect(-1, -1, -1, -1) for now in consistency with the other code.  A width and height of 0 means the window is minimized at least on Windows and should probably be documented as such.

GregoryMorse:
I updated my initial comment, as GetWindowRect/GetClientRect very well do not at least in a documented way use SendMessage and may very well use the data structure as specified in CreateWindow arguments for retrieving the window rectangle (client rect almost certainly just uses the window rect and origins to 0, 0 and adjusts width and height appropriately as right-left+1, bottom-right+1).  I am not sure if it flushes the queue or not.  But its likely thread safe in general regardless.  As to desiring a flush of the queue, since generally the context is for resizing live video with overlays, the issue works itself out over a frame as the old frame would be scaled (undesirably) for a very short period.

Multiple monitor support is partially working in HighGUI.  If the window gets too wide (more than a monitor), I cannot shrink it again without manually editing the registry values where the width is persisted.  Otherwise it works fully, even showing negative top or left values when the origin screen is aligned as such.  GetClientRect has already abstracted this away so for this change its not important.

GregoryMorse:
It is final now with the change made, returning errors as cv::Rect(-1, -1, -1, -1) for now in consistency with the other code. A width and height of 0 means the window is minimized at least on Windows and should probably be documented as such.

PR #10436, opened 2017-12-27
Test threads number command-line parameter

alalek:
OpenCV test system has "--perf_threads=N" option to control number of used threads in perf tests.
This patch adds similar option for accuracy tests ("--test_threads=N")

Also patch fixes issues with DNN tests which don't respect external settings.


alalek:
:+1:

PR #10446, opened 2017-12-28
add style draw in yolo

cabelo:
### This pullrequest changes 

Added style in the Yolo drawing.

Example below:

### DIFF
--- a/samples/dnn/yolo_object_detection.cpp
+++ b/samples/dnn/yolo_object_detection.cpp
@@ -152,7 +152,7 @@ int main(int argc, char** argv)
                 }
                 else
                 {
-                    line(frame, Point(xLeftBottom,yLeftBottom), Point(xLeftBottom+((width / 2)*frame.cols),yLeftBottom+((height / 2)*frame.rows )  ),Scalar(0, 255, 0),1);				
+                    line(frame, Point(xLeftBottom,yLeftBottom), Point(xLeftBottom+((width / 2)*frame.cols),yLeftBottom+((height / 2)*frame.rows )  ),Scalar(0, 255, 0),1);
                 }
 
                 if (objectClass < classNamesVec.size())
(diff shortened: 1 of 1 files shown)


cabelo:
Example :
example_style

samples/dnn/yolo_object_detection.cpp
```diff
@@ -156,8 +164,9 @@ int main(int argc, char** argv)
                     int baseLine = 0;
                     Size labelSize = getTextSize(label, FONT_HERSHEY_SIMPLEX, 0.5, 1, &baseLine);
                     rectangle(frame, Rect(Point(xLeftBottom, yLeftBottom ),
-                                          Size(labelSize.width, labelSize.height + baseLine)),
-                              Scalar(255, 255, 255), CV_FILLED);
+                              Size(labelSize.width, labelSize.height + baseLine)),
```
alalek: This indentation change does not correct: `Size()` is argument of `Rect()` ctor, not `rectangle()` call.

cabelo: Okay, but the code was already in the previous version.

I will fix

cabelo: ok, I understand sorry

cabelo: done

dkurt:
@cabelo, do we really need it? It's hard to understand an accuracy of predictions this way. I mean the same line might be drawn as for good bounding box as well for a poor one.
screenshot from 2017-12-28 12-35-19

cabelo:
@dkurt  I think that sometimes the boxes pollute the image a lot to analyze.

But you decide what's best.

alalek:
Rectangle diagonal line is enough to restore the original rectangle (with lines parallel X/Y axis), so I believe this line may be enough.

dkurt:
@alalek, or corners.
screenshot from 2017-12-28 17-04-16

alalek:
@dkurt Right.

BTW, In general, this is the sample, not a research tool. We should keep samples as simple as possible, just enough to demonstrate OpenCV functionality.

cabelo:
I'll make corners ok?

alalek:
I believe the current state of this patch is fine.

cabelo:
Ok @alalek  thank you.

Until the next patch

Bye

alalek:
I have pushed minor code refactoring update (see commit comments) including `cv::format` changes from #10483

dkurt:
:+1:

Issue #10439, opened 2017-12-27
DNN: need build option to disable memory manager


alalek:
To properly detect issues with valgrind tool.

related: #9160

PR #10456, opened 2017-12-28
Allocate new memory for optimized concat to prevent collisions.

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/10439

* Add a flag `opencv_dnn_REUSE_MEMORY` to disable memory reusing.
* Allocate memory for optimized concat (concat is skipped but it's inputs are computed directly into the concat's output) out of pool of shared blobs because it causes collisions like the following:

First 10 layers of SqueezeNet v1.1:

screenshot from 2017-12-28 16-10-00

Memory manager lets `fire2/squeeze1x1` (`fire2/squeeze1x1` + in-place `fire2/relu_squeeze1x1`) reuse output of `conv1`. However it also lets `fire2/concat` reuse the same blob of `conv1`. Generally it's correct but concat layer's optimization assign `fire2/expand1x1` and `fire2/expand3x3` place their outputs to concat's memory that is `conv1` blob. This way `fire2/expand3x3`'s input and output blobs are collided.

screenshot from 2017-12-28 16-10-01

As the simplest solution we can always allocate concat layers' memory but not reuse it.

topology visualized by http://ethereon.github.io/netscope

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -367,43 +367,42 @@ public:
         }
     }
 
-    void reuseOrCreate(const MatShape& shape, const LayerPin& lp, Mat& dst, bool force)
+    void reuseOrCreate(const MatShape& shape, const LayerPin& lp, Mat& dst)
     {
+#ifdef REUSE_DNN_MEMORY
         Mat bestBlob;
         LayerPin bestBlobPin;
 
-        if( !force )
+        std::map<LayerPin, Mat>::iterator hostIt;
+        std::map<LayerPin, int>::iterator refIt;
+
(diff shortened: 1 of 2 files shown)


modules/dnn/src/dnn.cpp
```diff
@@ -1415,6 +1410,9 @@ struct Net::Impl
 
                     if( i >= ninputs )
                     {
+                        // Allocate new memory to prevent collisions during memory
+                        // reusing (see https://github.com/opencv/opencv/pull/10456).
+                        output = output.clone();
```
alalek: :+1:

PR #10466, opened 2017-12-29
Wrap UMat dnn blobs

dkurt:

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -183,29 +183,34 @@ Mat blobFromImages(const std::vector<Mat>& images_, double scalefactor, Size siz
 class OpenCLBackendWrapper : public BackendWrapper
 {
 public:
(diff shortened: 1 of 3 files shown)


modules/dnn/src/dnn.cpp
```diff
@@ -180,6 +180,80 @@ Mat blobFromImages(const std::vector<Mat>& images_, double scalefactor, Size siz
     return blob;
 }
 
+class OpenCLBackendWrapper : public BackendWrapper
+{
+public:
+    OpenCLBackendWrapper(const Mat& m) : BackendWrapper(DNN_BACKEND_DEFAULT, DNN_TARGET_OPENCL)
+    {
+        umat = m.getUMat(ACCESS_RW);
```
alalek: `getMat()` / `getUMat()` are scope-based calls. These calls are not designed for using with long non-determined lifetime.

Consider using `copyTo()` instead (there is no "moveTo()" semantic between Mat/UMat yet).

dkurt: @alalek, Thank you for a review! `getUMat` has been replaced to `copyTo`.

PR #10473, opened 2017-12-30
cmake: avoid timestamp change of version_string.inc file

alalek:
To avoid unnecessary rebuilding of OpenCV binaries.

related #10300


alalek:
:+1:

PR #10489, opened 2018-01-02
Adding capability to parse subsections of a byte array in Java bindings

SarenT:
Adding capability to parse subsections of a byte array in Java bindings. (Because Java lacks pointers. Therefore, reading images within a subsection of a byte array is impossible by Java's nature and limitations. Because of this, many IO functions in Java require additional parameters offset and length to define, which section of an array to be read.) If these parameters are unavailable, such as with the current version of Java bindings of OpenCV, a solution would be to copy subsection of the array into new byte array object. This is unnecessarily inefficient. OpenCV anyhow copies the array again to create Mat object. I wanted to avoid this unnecessary copying for a project of mine by adding above mentioned parameters (offset + length) to the related functions.


modules/core/misc/java/src/java/core+Mat.java
```diff
@@ -1001,7 +1001,7 @@ public int put(int row, int col, short[] data) {
     }
 
     // javadoc:Mat::put(row,col,data)
-    public int put(int row, int col, byte[] data) {
+    public int put(int row, int col, byte[] data, int offset, int length) {
```
alalek: Please avoid changing of existed interfaces (making already existed programs incompatible) without strict requirements.
You should add new one overload here and below for native method (or use new name).

modules/core/misc/java/src/java/core+MatOfByte.java
```diff
@@ -30,22 +30,22 @@ public MatOfByte(Mat m) {
         //FIXME: do we need release() here?
     }
 
-    public MatOfByte(byte...a) {
+    public MatOfByte(int offset, int length, byte...a) {
```
alalek: The same note for this ctor and methods below.

modules/core/misc/java/test/MatTest.java
```diff
@@ -668,14 +668,14 @@ public void testPutIntIntByteArray() {
         byte[] buff0 = new byte[] { 10, 20, 30, 40, 50, 60 };
         byte[] buff1 = new byte[] { -1, -2, -3, -4, -5, -6 };
 
-        int bytesNum = m.put(1, 2, buff0);
+        int bytesNum = m.put(1, 2, buff0, 0, buff0.length);
```
alalek: Don't touch unrelated tests.
Add new one for this added feature.

modules/java/generator/src/cpp/Mat.cpp
```diff
@@ -1863,7 +1863,7 @@ namespace {
 #undef JOCvT
 }
 
-template<typename T> static int mat_put(cv::Mat* m, int row, int col, int count, char* buff)
+template<typename T> static int mat_put(cv::Mat* m, int row, int col, int count, int offset, char* buff)
```
alalek: Changing of internal interfaces are OK.

modules/core/misc/java/src/java/core+MatOfByte.java
```diff
@@ -30,24 +30,37 @@ public MatOfByte(Mat m) {
         //FIXME: do we need release() here?
     }
 
-    public MatOfByte(byte...a) {
+	public MatOfByte(byte...a) {
```
alalek: Please restore indentation (tabs -> 4 spaces) here and below.

modules/java/generator/src/cpp/Mat.cpp
```diff
@@ -1919,12 +1919,12 @@ template<class ARRAY> static jint java_mat_put(JNIEnv* env, jlong self, jint row
 extern "C" {
 
 JNIEXPORT jint JNICALL Java_org_opencv_core_Mat_nPutB
-    (JNIEnv* env, jclass, jlong self, jint row, jint col, jint count, jbyteArray vals);
+    (JNIEnv* env, jclass, jlong self, jint row, jint col, jint count, jint offset, jbyteArray vals);
```
alalek: This original interface should be restored too (it is JNIEXPORT , not internal).
Please add tests to ensure that this really works and these changes will not be broken in the future (somewhere [here](https://github.com/opencv/opencv/blob/3.4.0/modules/core/misc/java/test/MatTest.java))

modules/core/misc/java/src/java/core+MatOfByte.java
SarenT: Oops sorry.

alalek:
@SarenT Looks like, tests are not added into the latest commit.

SarenT:
@alalek I am trying to figure out that right now. I have created a MatOfByteTest.java class file and 2 image files (derivatives of lena.png and chessboard.jpg with 10 byte margins in front and at the end). Git reported it to be untracked files. Now they should be there. I am a beginner with git, I forgot the git add -A.

modules/core/misc/java/test/MatOfByteTest.java
```diff
@@ -0,0 +1,94 @@
...
+    Mat rgbLena2;
+    Mat grayChess2;
+    Mat rgbLenawMargin;
+    Mat grayChesswMargin;
+    
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+        LENA_W_10BYTEMARGIN_PATH = OpenCVTestRunner.LENA_PATH.replace("lena.png", "lena-10byte_margin.png");
+        CHESS_W_10BYTEMARGIN_PATH = OpenCVTestRunner.CHESS_PATH.replace("chessboard.jpg", "chessboard-10byte_margin.jpg");
```
alalek: These new files are malformed.
Please simplify test without imdecode() calls. We should test this functionality, not imdecode().

Something like this:
```
byte[] inputBytes = new byte[5] { 1,2,3,4,5 };
May m = new Mat(inputBytes);
May m0 = new Mat(0, 5, inputBytes);
assertArrayEquals(m, m0, EPS);
```

```
May m1 = new Mat(2, 2, inputBytes);
assertEquals(3, m1.get(0));
```

modules/core/misc/java/src/java/core+MatOfByte.java
```diff
@@ -48,6 +53,14 @@ public void fromArray(byte...a) {
         put(0, 0, a); //TODO: check ret val!
     }
 
+    public void fromArray(int offset, int length, byte...a) {
+        if(a==null || a.length==0 || a.length - (length + offset) < 0)
+            return;
```
alalek: > a.length - (length + offset) < 0

I believe we should have here the OutOfRange exception.

P.S. Add "offset >= 0" check.

SarenT: Hi, I agree with offset >= 0 (or "offset < 0" to return) but the original function doesn't throw any exception. Should I throw the exception or skip the check to let it naturally fail with OutOfRange exception?

modules/core/misc/java/test/MatOfByteTest.java
SarenT: I see, tests need to be as independent as possible. I will fix this.

SarenT: The files are malformed for a reason. My initial goal was to extract images embedded in RAW files coming from my DSLR but giving an offset and length of the JPG file. I added random bytes before and after the files to see if I can extract the png or jpg encoded files out. But IMO your example should be sufficient to test this. So, I will simplify it.

modules/core/misc/java/src/java/core+MatOfByte.java
alalek: `offset < 0` should throw OutOfRangeException too instead of return.

There is missing exception throwing on "a.length - (length + offset) < 0" condition in the last commit.

SarenT: Should we actually have something like `throw new OutOfRangeException()`? I am a bit confused, because in the original function a.length=0 returns nothing (no exception is thrown). Conceptially, `offset < 0` is similar to a `a=null` or `a.length=0`. First should have thrown NullPointerException and the second should have thrown IllegalArgumentException. Instead the original function returns. I did not want to change that design. What do you think?

alalek: Negative "offset" or "length" doesn't make any sense - it is programming error in general. Good library should point on this by raising an error message (detect errors early).
Bad libraries silently "eat" such errors and provide wrong results in the end of complex processing pipeline - hours of hard debugging are guaranteed.

I mean this:
```
if (offset < 0)
    throw new OutOfRangeException(...);
if (length < 0)
    throw new OutOfRangeException(...);
if (a==null || a.length==0)
    return; // keep original behavior, but NPE is preferable for a==null
if (a.length - (length + offset) < 0)
    throw new OutOfRangeException(...);
```

PR #10492, opened 2018-01-03
add ocl implementation of proposal layer

pengli:
```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/src/layers/proposal_layer.cpp
+++ b/modules/dnn/src/layers/proposal_layer.cpp
@@ -148,10 +148,88 @@ public:
         deltasPermute->finalize(layerInputs, layerOutputs);
     }
 
+#ifdef HAVE_OPENCL
+    bool forward_ocl(InputArrayOfArrays inputs_, OutputArrayOfArrays outputs_, OutputArrayOfArrays internals_)
+    {
+        std::vector<UMat> inputs;
+        std::vector<UMat> outputs;
+        std::vector<UMat> internals;
+
+        inputs_.getUMatVector(inputs);
+        outputs_.getUMatVector(outputs);
+        internals_.getUMatVector(internals);
+
+        CV_Assert(inputs.size() == 3);
+        CV_Assert(internals.size() == 3);
+        const UMat& scores = inputs[0];
+        const UMat& bboxDeltas = inputs[1];
+        const UMat& imInfo = inputs[2];
+        UMat& priorBoxes = internals[0];
+        UMat& permuttedScores = internals[1];
+        UMat& permuttedDeltas = internals[2];
+
+        CV_Assert(imInfo.total() >= 2);
+        // We've chosen the smallest data type because we need just a shape from it.
+        Mat szMat;
+        imInfo.copyTo(szMat);
+        int rows = (int)szMat.at<float>(0);
+        int cols = (int)szMat.at<float>(1);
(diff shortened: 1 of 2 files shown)


modules/dnn/src/layers/proposal_layer.cpp
```diff
@@ -148,11 +148,86 @@ class ProposalLayerImpl : public ProposalLayer
...
+#ifdef HAVE_OPENCL
+    bool forward_ocl(InputArrayOfArrays inps, OutputArrayOfArrays outs, OutputArrayOfArrays intnals)
+    {
+        std::vector<UMat> inputs;
+        std::vector<UMat> outputs;
+        std::vector<UMat> internals;
+
+        inps.getUMatVector(inputs);
+        outs.getUMatVector(outputs);
+        intnals.getUMatVector(internals);
```
alalek: I think it is better to use "intnals" => "internals_"

modules/dnn/src/layers/proposal_layer.cpp
```diff
@@ -148,11 +148,86 @@ class ProposalLayerImpl : public ProposalLayer
...
+        const UMat& imInfo = inputs[2];
+        UMat& priorBoxes = internals[0];
+        UMat& permuttedScores = internals[1];
+        UMat& permuttedDeltas = internals[2];
+
+        CV_Assert(imInfo.total() >= 2);
+        // We've chosen the smallest data type because we need just a shape from it.
+        float rows = imInfo.getMat(ACCESS_READ).at<float>(0);
+        float cols = imInfo.getMat(ACCESS_READ).at<float>(1);
+        umat_fakeImageBlob.create(shape(1, 1, (int)rows, (int)cols), CV_8UC1);
```
alalek: Need to initialize values of this "input".

modules/dnn/src/layers/proposal_layer.cpp
```diff
@@ -148,11 +148,86 @@ class ProposalLayerImpl : public ProposalLayer
...
+        const UMat& bboxDeltas = inputs[1];
+        const UMat& imInfo = inputs[2];
+        UMat& priorBoxes = internals[0];
+        UMat& permuttedScores = internals[1];
+        UMat& permuttedDeltas = internals[2];
+
+        CV_Assert(imInfo.total() >= 2);
+        // We've chosen the smallest data type because we need just a shape from it.
+        float rows = imInfo.getMat(ACCESS_READ).at<float>(0);
+        float cols = imInfo.getMat(ACCESS_READ).at<float>(1);
```
alalek: I think more efficient way is to select ROI (assume 2D layout - or use reshape()) and to copy it once (instead of two maps).

```
Mat szMat;
imInfo(Rect(0, 0, 1, 2)).copyTo(szMat);
float rows = (int)szMat.at<float>(0);
float cols = (int)szMat.at<float>(1);
```

modules/dnn/src/layers/proposal_layer.cpp
pengli: umat_fakeImageBlob is initialized with zero

modules/dnn/src/layers/proposal_layer.cpp
pengli: yeah, variable name is changed.

modules/dnn/src/layers/proposal_layer.cpp
pengli: ok, replaced two maps with one copyTo

PR #10502, opened 2018-01-04
Save video file

cabelo:
I posted this patch for when we need it shows the result of processing.


samples/dnn/yolo_object_detection.cpp
```diff
@@ -92,7 +105,6 @@ int main(int argc, char** argv)
 
     for(;;)
     {
-        Mat frame;
         cap >> frame; // get a new frame from camera/video or read image
```
alalek: The first frame is dropped/lost here (from the line 90). Different number of frames in input and output files will confuse users.

cabelo: Fix, thank you

PR #10504, opened 2018-01-04
modules/videoio: fix PS3 Eye camera property window

sthalik:
### PS3 Eye dshow dialog fix

The PS3 Eye camera is unable to display its property page dialog. It's a high-speed camera good for vision processing, up to 187 Hz at 320x240 and 75 Hz at 640x480. Users of the opentrack project rely on it heavily for infrared blob detection, and no doubt there are many other users of this hardware.

The solution came when running "amcap", the Windows SDK sample. The camera requires the video preview in order to display the dialog correctly. Turns out, calling `IMoniker::BindToObject` for `IID_IBaseFilter` resets the preview filter, confusing the DirectShow filter.

I changed all the places where the `pVideoInputFilter` value was freed and only initialized that field if it's `NULL` to begin with.

#### Camera dialog count

Also added is the ability for checking the amount of property windows opened for the camera. The dialog will crash if opened twice. I reused `cv::CAP_PROP_SETTINGS` for `cv::VideoCapture::get()` as not to bloat the API. Since that enum value isn't used otherwise, there are no regressions here.

#### Caveats

1) There's a comment about MPEG captures not showing the property window unless `IID_BaseFilter` is re-initialized. That comment is over seven years old though. If the issue is still present, perhaps I should amend the commit to check for MPEG fourcc's.
2) For PS3 Eye it's necessary to grab at least one frame anyway.

#### Room for improvement

1) We can use smart pointers to avoid the need for keeping the capture alive manually. Do you want it?
2) We could introduce another enum value for opening a blocking camera dialog. This is another way to keep the capture alive. User can run the property window thread on his own.
3) Need testing/confirmation with other drivers and dshow filters. A4 TECH cameras are another offender with all-around problems.
4) Maybe the `cv::VideoCapture::get(cv::CAP_PROP_SETTINGS)` change should go into a separate commit.

### DIFF
--- a/modules/videoio/src/cap_dshow.cpp
+++ b/modules/videoio/src/cap_dshow.cpp
@@ -434,6 +434,8 @@ class videoDevice{
         int  myID;
         long requestedFrameTime; //ie fps
 
+        LONG volatile property_window_count;
+
         char  nDeviceName[255];
         WCHAR wDeviceName[255];
 
@@ -499,7 +501,7 @@ class videoInput{
 
         //Launches a pop up settings window
         //For some reason in GLUT you have to call it twice each time.
-        void showSettingsWindow(int deviceID);
+        bool showSettingsWindow(int deviceID);
 
         //Manual control over settings thanks.....
         //These are experimental for now.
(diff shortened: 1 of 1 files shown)


alalek:
Verified with HP laptop camera - works well.

Do we really need multiple property windows? They are not synchronized. I believe we should block opening of the second property window.

sthalik:
@alalek you're right, a boolean flag will do.

PR #10505, opened 2018-01-04
Rewrite polar transformations

### LINKED ISSUE #9346: Set manually the destination image size of LinearPolar / LogPolar
##### Detailed description

The methods linearPolar and logPolar return the image of the same size as the source.

I consider this behavior strange since the dimensions of the result image are ρ and φ and have nothing to do with the original image width and height.

Ideally, the radial and angular dimensions should be set manually with some reasonable defaults depending on the _maxRadius_ parameter (say, _maxRadius_ × _π∙maxRadius_ to match the average resolution of the source for LinearPolar). I don’t think, however, that changing the default behavior (and thus breaking the compatibility) is the right thing to do.

So, is it possible to just add the method overloads with the destination image size as an argument to have the ability to set it manually?

PkLab:
resolves #9346

From the issue: 

> The methods linearPolar and logPolar return the image of the same size as the source.

The basic idea is to let user an option to set the destination image size or calculate it automatically so that the destination image size has same area of the bounding circle.

### This pullrequest changes

- A new `warpPolar`  function encapsulate both `linearPolar` and `logPolar` which are merged and (a bit) optimized
  - Destination size is a parameter or is calculated automatically to maintain objects area (or ratio) between source and destination
  - The ocl version hasn't been provided because the function (like old ones) uses `remap` which already has `ocl_remap`.
  - The function avoids to use `InputArray::GetMat` and `Mat::create` to get maximum benefit using `remap | ocl_remap`  
  - The accuracy test `Imgproc_warpPolar` has been added
- The functions `linearPolar` and `logPolar` has been deprecated.
  - The ocl version `ocl_linearPolar` and `ocl_logPolar` would be useless ?
  - The accuracy tests `Imgproc_logPolar` and `Imgproc_linearPolar` would be useless ?
- C version of the functions `cvLinearPolar` and `cvLogPolar` already use the size of destination therefore they are coherent with new `warpPolar`
- The related doc and the example `polar_transforms.cpp` has been updated

**Enjoy with comments and suggestions**


Felix-neko:
Well, just now I have written the next in Python as a replacement.

It is used to interpolate the radar images from sector radars and height matrices on my simulator. The hint is that polar image can have some offset from center and it can have angles not 0...360 but given in some interval.

```
class Polar2CartInterpolator(object):
    def __init__(self, ang_min, ang_max, r_min, r_max, height_matrix, add_one_more_angle=True, take_nearest=False):
        """
        Args:
            ang_min(float): minimal angle in radians
            ang_max(float): maximal angle in radians
            r_min(float): minimal distance
            r_max(float): maximal distance
            height_matrix(array-like): a 2D matrix whose shape is interpreted as ``n_angles``-by-``n_radius``
            add_one_more_angle(bool): if `True`, we pad the height matrix with extra row taken from its beginning
                (for example, if we have height matrix with 360 angles from 0 to 359 degrees, now has covers 361 angles
                from 0 to 360)
            take_nearest(bool): if `True`, we will interpolate with nearest value, if `False`, we'll use linear
                interpolation
        """
        self._ang_min = float(ang_min)
        self._ang_max = float(ang_max)
        assert self._ang_max > self._ang_min
        self._r_min = float(r_min)
        self._r_max = float(r_max)
        assert self._r_max > self._r_min
        assert self._r_min >= 0

        self._height = np.asarray(height_matrix)
        self._take_nearest = bool(take_nearest)
        assert len(self._height.shape) == 2
        self._n_ang, self._n_r = self._height.shape

        if self._take_nearest:
            method = "nearest"
        else:
            method = "linear"

        if bool(add_one_more_angle):
            ang_step = (self._ang_max - self._ang_min) / (self._n_ang - 1)
            self._ang_max += ang_step
            self._n_ang += 1
            self._height = np.concatenate((self._height, self._height[None, 0, :]), axis=0)

        ang_grid_src = np.linspace(self._ang_min, self._ang_max, self._n_ang)
        r_grid_src = np.linspace(self._r_min, self._r_max, self._n_r)

        self._interpolator = sp.interpolate.RegularGridInterpolator(
            points=(ang_grid_src, r_grid_src), values=self._height, bounds_error=False, method=method)

    def generate_cartesian_pic(self, x_min, x_max, n_x, y_min, y_max, n_y):
        """
        Args:
            x_min(float):
            x_max(float):
            n_x(float): number of X points
            y_min(float):
            y_max(float):
            n_y(float): number of Y points
        Returns:
            :class:`numpy.ndarray`: a 2D matrix with shape ``n_y``-by-``n_x``

        """
        x_grid = np.linspace(float(x_min), float(x_max), int(n_x))
        y_grid = np.linspace(float(y_min), float(y_max), int(n_y))
        x_mesh, y_mesh = np.meshgrid(x_grid, y_grid)

        complex_coords = np.zeros(shape=x_mesh.shape, dtype=np.complex128)
        complex_coords.real = x_mesh
        complex_coords.imag = y_mesh

        ang_mesh = (np.pi / 2. - np.angle(complex_coords) + 2 * np.pi) % (2 * np.pi)  # Azimuth: clockwisee from north
        r_mesh = np.abs(complex_coords)

        out_pic = self._interpolator(np.stack((ang_mesh, r_mesh), axis=-1))
        return out_pic
```

PkLab:
@sovrasov please can you (or someone else) check the build error ? it's only coming from `Linux OpenCL: test_dnn-ippicv-opencv` but I haven't touched dnn module. It looks this error is present also in other PRs

1000Thanks!

PkLab:
@Felix-neko nice options maybe the parameters list being too long/complicate

An alternative would be to use a ROI on output image to select the sector of annulus you want
```
wrapPolar(src,dst,center,maxRadius,false);
Klin = dst.cols / maxRadius;
Kangle_rad = dst.rows/ CV_2PI;
Kangle_deg = dst.rows/ 360.0;
Rect roi( Point(r_min * Klin, ang_min_rad * Kangle_rad),
          Point(r_max * Klin, ang_max_rad * Kangle_rad) );
Mat myAnnularSector = dst(roi);
```
But of course using your idea would avoid to calculate mapping for unused part of the polar image

PS: it would be nice if you can post here some sample image from radar **free to use in the doc**

catree:
Maybe `warpPolar()` (as `warpAffine()`, `warpPerspective()`) looks more appropriate than `wrapPolar()`?

PkLab:
@catree yes of course that's was my intention :( I'm going to commit the update

vpisarev:
@PkLab, thank you for the contribution! I think, I understand the rationale of making a single internal function that handles both cases (log and linear), but do we need to introduce another external functions and deprecate two existing ones? Maybe, we should retain the external API as-is?

PkLab:
@vpisarev I can do it as you prefer. 
Two existing one works differently, more in details they cover a single case of the new one (when dst.size == src.size). New function offers the possibilities to set `dst.size` and more important it can calculate automatically keeping same object areas between src and dst.
In addiction old functions requires OCL version but not new one.
Definitely new function has new parameters for different approach and different esult that's why I think old function should be replaced by new one that has new name to avoid backward compatibility.

We can have 2 functions instead 1 ... can you suggest a sample function interface for the external API ?

vpisarev:
@PkLab, I weighted various options and since we now have the binary API compatibility restriction, so that we cannot alter the existing `logPolar()` and `linearPolar()` functions anyway, I think that the solution with the new function should be fine. The only thing I'd suggest to change is to replace the boolean `semiLog` parameter with the 1-bit flag that is added to `flags`. That is, do something like this:

<pre>
enum WarpPolarMode
{
      WARP_POLAR_LINEAR = 0,
      WARP_POLAR_LOG = 256
};
CV_EXPORTS_W void warpPolar(InputArray src, OutputArray dst, Size dsize,
     Point2f center, double maxRadius, int flags);
</pre>

we should let the user to explicitly specify the mapping mode, so we retain all the parameters as non-optional.

PkLab:
It should be done as request by @vpisarev . I hope commits are fine.

I would like to show the function applied to a nice picture of our (Italy) Pantheon dome 2000 year old 👍 

pantheon2

Linear Polar than rotated right
pantheon2-rectified

PkLab:
@vpisarev What do you think if I change the function so that angles would be in cols instead in rows? To me look much more intuitive. I've a commit almost ready for this.

alalek:
@PkLab I believe we could add flag to specify side corresponding to warped singularity point (center), left (current), top (your pictures, also this transform used in eye iris recognition articles), bottom or right (probably last two are not necessary).

PkLab:
@alalek  It's not so easy to manage singularity point as "parameter" because mapping matrix would change. Ot would be required many if with different implementation of same concept function. It's cost/benefit. Is the "warped singularity point as parameter" so needed ?

alalek:
Not a separate parameter, just some enum values for flag (something like this, `MAP_ANGLE_VERTICAL` / `MAP_ANGLE_HORIZONTAL` - with some default (`=0`))

PkLab:
The point is not parameter or flag. Different singularity requires different mapping matrix that require different code

alalek:
Sorry, I don't understand. What are additional things required except swapping of rows/cols (in polar coords: angle / radius)? (yes, swapped/transposed result requires different mappings)

PkLab:
Not so hard I've already done here https://gist.github.com/PkLab/49a6e882806a62ed4315566733e4748b
(check //ANGLE IN COLS //ANGLE IN ROWS). By rows or by cols require 2 different implementation. This also for the inverse mapping... again its It's cost/benefit to maintain .

vpisarev:
@PkLab, I think that it should not be very difficult to merge the 2 implementations for different layouts into one. Just manually compare them and you will see that they are very similar

vpisarev:
replaced with #11323

snowman907:
Hey guys,
When I run the warpPolar I get the result I was expecting. But when I run it on a larger image dftcv_shft.shape, (19222, 19418, 2), I get the following error:

`error: OpenCV(4.2.0) D:\Build\OpenCV\opencv-4.2.0\modules\imgproc\src\imgwarp.cpp:1724: error: (-215:Assertion failed) dst.cols < SHRT_MAX && dst.rows < SHRT_MAX && src.cols < SHRT_MAX && src.rows` < SHRT_MAX in function 'cv::remap'`

According to warpPolar documentation the array needs be smaller than 32767x32767, which it is but seems to me the error is linked to the size of the array. But it seems the error is from remap!?

Any help would be appreciated.

PkLab:
The error is related to the destination image size , you are referring to the source image size.

```
dst.rows = radius
dst.cols = 2*radius*3.14
```
If you pick the center in the src and full `radius = src.rows/2` you have  `dst.cols = src.rosw*3.14 `. In your case you will have image size of 9611 x 60357

snowman907:
I am working with a digital elevation map, that is why the "image" is so big. I assume that 60357 is too big as the limit of the destination size is 32767!? If that is the problem, any idea how to overcome it?

PkLab:
The `Mat` object is defined using `int rows, int cols` hence you cant have images larger than `max int`
You might split the destination image by 4 using start and stop angle param.

I would suggest to visit https://answers.opencv.org you will find nice community where to ask "howto" questions

PR #10512, opened 2018-01-05
Update documentation

sturkmen72:
### This pullrequest changes

Updates documentation

### DIFF
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -140,7 +140,7 @@ public:
 
 By default the function prints information about the error to stderr,
 then it either stops if cv::setBreakOnError() had been called before or raises the exception.
-It is possible to alternate error processing by using cv::redirectError().
+It is possible to alternate error processing by using #redirectError().
 @param exc the exception raisen.
 @deprecated drop this version
  */
@@ -175,7 +175,7 @@ enum CovarFlags {
     /**The output covariance matrix is calculated as:
         \f[\texttt{scale}   \cdot  [  \texttt{vects}  [0]-  \texttt{mean}  , \texttt{vects}  [1]-  \texttt{mean}  ,...]  \cdot  [ \texttt{vects}  [0]- \texttt{mean}  , \texttt{vects}  [1]- \texttt{mean}  ,...]^T,\f]
         covar will be a square matrix of the same size as the total number of elements in each input
-        vector. One and only one of COVAR_SCRAMBLED and COVAR_NORMAL must be specified.*/
+        vector. One and only one of #COVAR_SCRAMBLED and #COVAR_NORMAL must be specified.*/
     COVAR_NORMAL    = 1,
     /** If the flag is specified, the function does not calculate mean from
         the input vectors but, instead, uses the passed mean vector. This is useful if mean has been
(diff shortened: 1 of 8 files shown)


samples/cpp/tutorial_code/snippets/imgcodecs_imwrite.cpp
```diff
@@ -0,0 +1,37 @@
...
+int main()
+{
+    // Create mat with alpha channel
+    Mat mat(480, 640, CV_8UC4);
+    createAlphaMat(mat);
+    vector<int> compression_params;
+    compression_params.push_back(IMWRITE_PNG_COMPRESSION);
+    compression_params.push_back(9);
+    try {
+        imwrite("alpha.png", mat, compression_params);
```
alalek: BTW, this is not directly related to this great patch, but `imwrite()` return "bool" result.
So the last message is confusing in case of "false" result.
```
bool result = false;
try
{
    result = imwrite("alpha.png", mat, compression_params);
}
catch (const cv::Exception& ex)
{
    fprintf(stderr, "Exception converting image to PNG format: %s\n", ex.what());
}
if (result)
    printf("Saved PNG file with alpha data.\n");
else
    printf("ERROR: Can't save PNG file.\n");
return result ? 0 : 1;
```

samples/cpp/tutorial_code/snippets/imgcodecs_imwrite.cpp
```diff
@@ -0,0 +1,37 @@
+#include <opencv2/opencv.hpp>
+
+using namespace cv;
+using namespace std;
+
+void createAlphaMat(Mat &mat)
```
alalek: to eliminate build warning use "static" here:
```
static void createAlphaMat(Mat &mat)
```

sturkmen72:
@alalek thank you for your review. let's keep the PR open. i will add similar changes later.

sturkmen72:
need help about CLAHE
http://pullrequest.opencv.org/buildbot/export/pr/10512/docs/d6/db6/classcv_1_1CLAHE.html#a3b96ad7aa2dbea70310d74d717bc5c65

modules/core/include/opencv2/core.hpp
```diff
@@ -862,8 +862,8 @@ CV_EXPORTS void minMaxLoc(const SparseMat& a, double* minVal,
 The function cv::reduce reduces the matrix to a vector by treating the matrix rows/columns as a set of
 1D vectors and performing the specified operation on the vectors until a single row/column is
 obtained. For example, the function can be used to compute horizontal and vertical projections of a
-raster image. In case of REDUCE_MAX and REDUCE_MIN , the output image should have the same type as the source one.
-In case of REDUCE_SUM and REDUCE_AVG , the output may have a larger element bit-depth to preserve accuracy.
+raster image. In case of cv::REDUCE_MAX and cv::REDUCE_MIN , the output image should have the same type as the source one.
+In case of cv::REDUCE_SUM and cv::REDUCE_AVG , the output may have a larger element bit-depth to preserve accuracy.
```
alalek: if you want, you can replace "cv::" prefix here to "#" symbol ("%" for classes): https://www.stack.nl/~dimitri/doxygen/manual/autolink.html

modules/core/include/opencv2/core/types.hpp
```diff
@@ -181,8 +181,8 @@ template<typename _Tp> class Point_
     double cross(const Point_& pt) const;
     //! checks whether the point is inside the specified rectangle
     bool inside(const Rect_<_Tp>& r) const;
-
-    _Tp x, y; //< the point coordinates
+    //! the point coordinates
+    _Tp x, y;
```
alalek: Original same-line comment should be "//!<" (missing '!')

See the same change below (line 266)

Issue #10510, opened 2018-01-04
protobuf: build issue


alalek:
Problem is fixed in protobuf upstream (3.2.0+).

With Android NDK 16.1 (gcc):

> FAILED: 3rdparty/protobuf/CMakeFiles/libprotobuf.dir/src/google/protobuf/util/internal/type_info.cc.o 
[...]

Issue #10448, opened 2017-12-28
Explaination for build_sdk.py


hannesa2:
##### System information (version)
- OpenCV => 3.4
- Operating System / Platform => Mac

##### Detailed description

I try to build OpenCV for Android by my own on my Mac, inspired by this [log](http://pullrequest.opencv.org/buildbot/builders/master-contrib_pack-contrib-android/builds/10317/steps/build%20sdk/logs/stdio) I installed the exact NDK `android-ndk-r10e` 

and I was forced to install `ninja` by `brew install ninja`
but currently I see 

> CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set

But I need some explanation for 
`buildenv  python ../opencv/platforms/android/build_sdk.py --build_doc --extra_pack 2.4.13:/opt/android/pack_2.4/ . ../opencv`

What is
* `buildenv` how does it looks like ?
* `--extra_pack` what is the content of `/opt/android/pack_2.4/` ?
* how to get rid of  `CMAKE_MAKE_PROGRAM is not set` ?

alalek:
- buildenv: standard ANDROID_SDK / ANDROID_NDK / NDKROOT / Java env variables
- ignore `--extra-pack`: it is for bundling OpenCV 2.4 into OpenCV Manager for Android package (deprecated component)
- try cleaning build directory, check that "ninja" command works from terminal (adjust PATH env variable if required)

hannesa2:
Thank you for the clarification. I only works on command line, and `ninja` works on any directory which contains a `build.ninja`file.
With `python ../opencv/platforms/android/build_sdk.py . ../opencv` I get 

[...]

Issue #8460, opened 2017-03-25
SDK Tools, Revision 25.3.0 and OpenCVDetectAndroidSDK.cmake


mishurov:
##### System information (version)

- OpenCV => 3.2.0
- Operating System / Platform => Gentoo / Android
- Compiler => Android NDK toolchain gcc

##### Detailed description
OpenCVDetectAndroidSDK.cmake doesn't detect an Android SDK after updating to the SDK Tools, Revision 25.3.0.
```bash
--     SDK target:                  android_sdk_target_status-NOTFOUND
--     android tool:                NO
```
```bash
~/Android/Sdk/tools/android
The android command is no longer available.
For manual SDK and AVD management, please use Android Studio.
For command-line tools, use tools/bin/sdkmanager and tools/bin/avdmanager
```
An older version of the tools just werks.

https://developer.android.com/studio/releases/sdk-tools.html
> SDK Tools, Revision 25.3.0 (March 2017)
>  ...
> Obsolete/deprecated tools have been removed:
> android
> ddms (instead see Using DDMS)
> draw9patch (instead see Draw 9-patch)
> hierarchyviewer (instead see Profile Your Layout with Hierarchy Viewer)
> traceview (instead see Profiling with Traceview and dmtracedump)
> ant scripts
> Project and activity templates

##### Steps to reproduce

1. Update SDK tools using an Android Studio
2. Try to build an OpenCV for Android with Java wrappers and examples support.

---
***alalek***:

Current solution is to downgrade SDK Tools to the latest available version prior 25.3.0: [25.2.5](https://developer.android.com/studio/releases/sdk-tools.html). Direct download links are:
- https://dl.google.com/android/repository/tools_r25.2.5-linux.zip (265Mb)
- https://dl.google.com/android/repository/tools_r25.2.5-macosx.zip (191Mb)
- https://dl.google.com/android/repository/tools_r25.2.5-windows.zip (293Mb)

diroitman:
Hi

I'm also interested - so how do you build opencv for android platform?
I tried platform/android/build_sdk.py script and it gives me the same error

Thank you

mishurov:
Gentoo has the old versions of the SDK with the working "android" tool. I set environment variable with the path to the old version instead of the user directory where Android Studio installs recent versions of SDK.
https://packages.gentoo.org/packages/dev-util/android-sdk-update-manager

victorv:
👍

adamajohnston:
[...]

PR #10515, opened 2018-01-05
Android: support NDK16

### LINKED ISSUE #10229: OpenCV doesn't link on Android x86/x86_64 with ndk 16.0.4442984 with R_386_GOTOFF error.

alalek:
- cherry-pick commit 90cc3f6 from #10288
- resolves #10229
- related #10510
- build_sdk.py still requires additional configuration changes (bump NDK API >= 14, GCC 4.9 anywhere, etc)
- old Android SDK with "android" tool is still required (ver 25.2.5 see #8460)
- resolves CMake 3.10 problem with missing `CMAKE_MAKE_PROGRAM` in case of (`-GNinja`): #10448

BTW, Binaries size of OpenCV Android SDK with NDK r16 is about ~3Gb (NDK r10 ~800Mb). Most part of this impact is related to static .a files.

```
allow_multiple_commits=1
```


PhilLab:
👍 Thank you for this, much appreciated !

PR #10517, opened 2018-01-05
video(perf): add Mog2/KNN tests, fixed bug in prepareData()

alalek:
prepareData() bug feeds the same image (the latest).
OCL perf test doesn't pass accuracy(!) check now, so this check is disabled.

related #10432


alalek:
:+1:

PR #10522, opened 2018-01-05
Issue10519

### LINKED ISSUE #10519: cvFindExtrinsicCameraParams2 is unsafe when count=3 points

- OpenCV => 3.4
- Compiler => GCC 5.4

##### Detailed description

cvFindExtrinsicCameraParams2 is unsafe when 3 points are used and useExtrinsicGuess is false. Code handling the case of 3 points should be removed and an associated assertion should be added preventing its use when 3 points are used. 

There reason why this is unsafe is that this uses LM optimisation that returns only a *single* solution. However pnp when n=3 (p3p) can have up to four real-valued solutions. A user should not call  cvFindExtrinsicCameraParams2 when n=3 and when no initial solution is provided, because there is no guarantee that it will converge on the correct solution. 

The case of n=3 is not handled in solvePnP for this exact reason (which calls cvFindExtrinsicCameraParams2) .

tobycollins:
resolves #10519

### This pullrequest changes

Added assertion to fail if (count == 3 && !useExtrinsicGuess)
Removed condition count < 4 when useExtrinsicGuess == false

### DIFF
--- a/modules/calib3d/src/calibration.cpp
+++ b/modules/calib3d/src/calibration.cpp
@@ -252,7 +252,7 @@ CV_IMPL int cvRodrigues2( const CvMat* src, CvMat* dst, CvMat* jacobian )
 {
     int depth, elem_size;
     int i, k;
-    double J[27];
+    double J[27] = {0};
     CvMat matJ = cvMat( 3, 9, CV_64F, J );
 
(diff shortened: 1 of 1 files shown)


vpisarev:
looks good to me! :+1:

PR #10529, opened 2018-01-06
Add a parameter labels to command line

LaurentBerger:
This pullrequest changes

Path to label name file can be give using command line


samples/dnn/caffe_googlenet.cpp
```diff
@@ -169,7 +171,11 @@ int main(int argc, char **argv)
     //! [Gather output]
 
     //! [Print results]
-    std::vector<String> classNames = readClassNames();
+    std::vector<String> classNames;
+    if (classNameFile.empty())
+        classNames = readClassNames();
```
alalek: Does this condition really pass? I mean there is default value for this parameter, so it should not be empty.

sturkmen72:
is it better if std:: and cv:: prefixes removed

PR #10544, opened 2018-01-07
cmake: allow BUILD_FAT_JAVA_LIB for non-Android targets too

### LINKED ISSUE #10535: Allow defining CVAPI_EXPORTS even when BUILD_SHARED_LIBS is OFF
I would like to submit a change that will allow the optional building of a distribution that with  CVAPI_EXPORTS defined even when not BUILD(ing)_SHARED_LIBS. This will allow the building of a single  artifact that can be used in a polyglot development environment.

In my case, I'm building the Java wrappers with the JNI library statically linked to the rest of OpenCV with the intention of distributing it with (actually inside of) a jar file. This artifact is also supposed to be able to have C/C++ and additional JNI code written against it. Currently it's impossible to do this because compiling statically hides the main APIs visibility from the resulting JNI jar file.

See also http://answers.opencv.org/question/181751/java-jni-symbol-lookup-error-from-custom-native-extension-using-opencv-c/

I can submit a PR with the appropriate change if it's something the team would be interested in.

Thanks

alalek:
resolves #10535


alalek:
:+1:

Issue #8300, opened 2017-03-01
Uniform empty image handling


StevenPuttemans:
As suggested by @sovrasov in PR#8288 we definately need to unify the way that empty images are handled inside the imgproc module.

- OPTION 1: generate an assert from the moment an empty input is given
- OPTION 2: silently cope with empty data and make output also empty

The second option will require the user to do all the checking on data validity himself.

alalek:
Comment for collecting "Option 1" votes (please use GitHub emoji)

alalek:
Comment for collecting "Option 2" votes (please use GitHub emoji)

darbedar:
Why not use empty image and assert a warning?

StevenPuttemans:
> Why not use the empty image and assert a warning?

Thats actually what option 1 suggests :P

sturkmen72:
i think this can be closed down

Sahloul:
> The second option will require the user to do all the checking on data validity himself.

Well, `option one (not two)` will require the user to check every input before passing it, or an exception will occur! On the other hand, the second option gives the user a uniform execution path, in which the input checking is not needed. The output as well might not need a check, as per the scenario described below. In fact, I find the first option overwhelming the user with input checks!

Sometimes, the user need to split the input image into several regions, and process them individually for region-specific transforms. In that case, the user is mostly aware that some regions might be empty and will not complain if they receive empty output as well. That is, they avoid introducing extra logic for empty regions filtration, expecting OpenCV to behave the same with all regions (`option two`). At a later stage, they might just join all outputs (e.g. keypoints) into a single collection. While such logic is totally valid, option one will just break it.

mpolicki:
@StevenPuttemans _Both_ options require the user to check the validity of input data; it would be nice if you updated your OP to say so.

But actually, the second option only requires validation in some cases since working with empty images has legitimate uses; therefore the second option is strictly better.

Also, look at it from an algebraic point of view. The equivalent of option 1 would be if mathematics threw an exception if you tried to perform set union with an empty set. In the matrix world, an empty matrix is the equivalent of 0, and it's a perfectly legitimate value.

vrabaud:
BTW, why limit this to imgproc ? Why isn't it possible right to fill an empty matrix with random numbers ? https://github.com/opencv/opencv/blob/4b6047e746f06ce3e595c886cf6c0266498c6a67/modules/core/src/rand.cpp#L394
[...]

PR #10549, opened 2018-01-08
`FLANN::knnSearch` kNN assertion

### LINKED ISSUE #10548: `FLANN::knnSearch` does not assert the kNN value
- OpenCV => 3.4.0-dev
- Compiler => ALL

##### Detailed description
FLANN returns garbage indices and distances when kNN is larger than its training dataset size.
More importantly, when kNN is slightly larger than the dataset size (e.g. `kNN=3, size()=2`), the returned indices beyond the dataset size are zeros, which creates silent, misleading correspondences.

##### Steps to reproduce
This is not a Python bug, it is also reproducible using C++
```.py
import numpy as np
import cv2

kNN = 7 #Any number bigger that len(index_points) will do
index_points = np.array([[0, 1, 2], [3, 4, 5]], np.float32)
query_points = np.array([[1, 2, 3]], np.float32)

flann_index = cv2.flann.Index if cv2.__version__.startswith('2') else cv2.flann_Index
flann = flann_index(index_points, dict(algorithm=0, trees=5))
idx, dist = flann.knnSearch(query_points, kNN, params = {})

np.set_printoptions(formatter={'all':lambda x: '{0:.3g},'.format(x)})
print(idx)
print(dist)
```

##### Expected result
```
[[0, 1, nan, nan, nan, nan, nan,]]
[[3, 12, nan, nan, nan, nan, nan,]]
```
OR to throw an exception!

##### Actual result
```
[[0, 1, 1.63e+09, 1.7e+09, 0, 0, 15,]]
[[3, 12, 7.26e-39, 8.72e-39, 7.35e-39, 7.26e-39, 6.7e-39,]]
```

Sahloul:
resolves #10548
### This pullrequest changes
- Prevent the user from using kNN values larger than the dataset size

### DIFF
--- a/modules/flann/src/miniflann.cpp
+++ b/modules/flann/src/miniflann.cpp
@@ -482,6 +482,9 @@ void runKnnSearch_(void* index, const Mat& query, Mat& indices, Mat& dists,
     typedef typename Distance::ResultType DistanceType;
     int type = DataType<ElementType>::type;
     int dtype = DataType<DistanceType>::type;
+    IndexType* index_ = (IndexType*)index;
+
+    CV_Assert((size_t)knn <= index_->size());
     CV_Assert(query.type() == type && indices.type() == CV_32S && dists.type() == dtype);
     CV_Assert(query.isContinuous() && indices.isContinuous() && dists.isContinuous());
 
@@ -489,8 +492,8 @@ void runKnnSearch_(void* index, const Mat& query, Mat& indices, Mat& dists,
     ::cvflann::Matrix<int> _indices(indices.ptr<int>(), indices.rows, indices.cols);
     ::cvflann::Matrix<DistanceType> _dists(dists.ptr<DistanceType>(), dists.rows, dists.cols);
 
-    ((IndexType*)index)->knnSearch(_query, _indices, _dists, knn,
-                                   (const ::cvflann::SearchParams&)get_params(params));
+    index_->knnSearch(_query, _indices, _dists, knn,
+                      (const ::cvflann::SearchParams&)get_params(params));
 }
 
 template<typename Distance>


modules/flann/src/miniflann.cpp
```diff
@@ -482,15 +482,18 @@ void runKnnSearch_(void* index, const Mat& query, Mat& indices, Mat& dists,
     typedef typename Distance::ResultType DistanceType;
     int type = DataType<ElementType>::type;
     int dtype = DataType<DistanceType>::type;
+    IndexType* index_ = (IndexType*)index;
+
+    CV_Assert(knn <= index_->size());
```
Sahloul: There are two main strategies:
- Failure-avoidance strategies
    - Set `knn = MIN(knn, index_->size());`
    - Swallow it and return NaNs for the additional kNNs.
- Raise and exception to reject the user input

Is it my imagination, or this is also related to #8300? In that case, I guess the vote for this matter is meaningless, as there are cases where one strategy is better than the other, and vise versa.

In this specific case, the user is expecting a matrix of kNN width, returning smaller one is not a sane choice.
Moreover, while returning NaNs in the distance makes sense, the indices matrix needs to be `int`-type. Accordingly, while I am a fan of failure-avoidance strategies, it just does not make sense here, and an exception is the best choice.

modules/flann/src/miniflann.cpp
```diff
@@ -482,15 +482,18 @@ void runKnnSearch_(void* index, const Mat& query, Mat& indices, Mat& dists,
     typedef typename Distance::ResultType DistanceType;
     int type = DataType<ElementType>::type;
     int dtype = DataType<DistanceType>::type;
+    IndexType* index_ = (IndexType*)index;
+
+    CV_Assert((size_t)knn <= index_->size());
```
Sahloul: There are two main strategies:
- Failure-avoidance strategies
    - Set `knn = MIN(knn, index_->size());`
    - Swallow it and return NaNs for the additional kNNs.
- Raise and exception to reject the user input

Is it my imagination, or this is also related to #8300? In that case, I guess the vote for this matter is meaningless, as there are cases where one strategy is better than the other, and vise versa.

In this specific case, the user is expecting a matrix of kNN width, returning smaller one is not a sane choice.
Moreover, while returning NaNs in the distance makes sense, the indices matrix needs to be `int`-type. Accordingly, while I am a fan of failure-avoidance strategies, it just does not make sense here, and an exception is the best choice.

alalek: #8300 is about an empty cv::Mat as input parameter. Is empty cv::Mat an invalid parameter or not?

This parameters check is fine here.

Issue #10530, opened 2018-01-06
Can no longer load yahoo nsfw resnet model in 3.4.0 release


TechnikEmpire:
##### System information (version)

- OpenCV => 3.4.0
- Operating System / Platform => Windows 10
- Compiler => MSVC 2017

##### Detailed description
Used to be able to load and use [this model](https://github.com/yahoo/open_nsfw/tree/master/nsfw_model).

Since 3.4.0 release, fails with:
```
OpenCV Error: Unspecified error (FAILED: ReadProtoFromBinaryFile(param_file, param). Failed to parse NetParameter file: resnet_50_1by2_nsfw_new.caffemodel) in cv::dnn::ReadNetParamsFromBinaryFileOrDie, file C:\dev\lib\opencv-3.4.0\modules\dnn\src\caffe\caffe_io.cpp, line 1159
```

##### Steps to reproduce
Use DNN module to load the above linked module. It'll fail to load. Worked fine before 3.4.0. Just to be sure, rebuilt 3.3.1 from source just now and it does function in 3.3.1 but not 3.4.0 from release source.

TechnikEmpire:
Just to add a bit more info. I first thought maybe protobuf version had changed or something, so I used latest `caffe-cuda` package in Ubuntu 17.10 and the upgrade utilities for it to upgrade the prototxt and model. Wasn't the issue.

dkurt:
@TechnikEmpire, thank you! I think the problem came in https://github.com/opencv/opencv/pull/10356 where we've introduced a `psroi_pooling_param` field with a tag `10001`. Yahoo's Caffe has a field [source_class](https://github.com/yahoo/caffe/blob/807ee665699421759f8ae9f27da436a2de69fede/src/caffe/proto/caffe.proto#L355) with the same tag. So it's a conflict of tags.

dkurt:
@TechnikEmpire, Would it be too much trouble for you to test changes from PR https://github.com/opencv/opencv/pull/10550?

TechnikEmpire:
@dkurt for sure, I'll build it in a bit here and report back asap.

TechnikEmpire:
@dkurt as of c08a0ed, I still get the same failure and same error message.

TechnikEmpire:
Oops, what a dummy I can be, I assumed the PR was merged. Will rebuild and test lol sorry.

PR #10550, opened 2018-01-08
Replace Caffe's psroi_pooling_param tag from 10001 to 10002

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/10530


TechnikEmpire:
@dkurt This solves it. I didn't just load the model, I ran it through a classifier I've written using the model and I got identical results to cv 3.3.1 against a data set of 2K images. Thanks for looking into this!

PR #10556, opened 2018-01-09
ORB support for SIFT keypoints

### LINKED ISSUE #10555: ORB descriptor cannot use SIFT keypoints
- OpenCV => 3.4.0-dev
- Compiler => ALL

##### Detailed description
Although ORB works with FAST or AKAZE keypoints, it fails to use the SIFT ones!

##### Steps to reproduce
```.py
import cv2

sift = cv2.xfeatures2d.SIFT_create()
orb = cv2.AKAZE_create()

rgb = cv2.imread('any_image.png')
kps = sift.detect(rgb)
kps, descrs = orb.compute(rgb, kps)
```

##### Actual result
Memory allocation error, and a crash!

Open3D-admin:
resolves #10555 
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1545

### This pullrequest changes
- Ensures that octave value remains less than 128
- Update the first level value and levels count accommodatingly

### DIFF
--- a/modules/features2d/src/orb.cpp
+++ b/modules/features2d/src/orb.cpp
@@ -650,6 +650,19 @@ static inline float getScale(int level, int firstLevel, double scaleFactor)
     return (float)std::pow(scaleFactor, (double)(level - firstLevel));
 }
 
+static inline int getLevel(KeyPoint &kp)
+{
+    int level = kp.octave;
+    CV_Assert(0 <= level);
+    //SIFT encodes layer and octave together
+    if (level >= 255)
+    {
+        level &= 255;
+        //extract SIFT negative octave
+        if (level >= 128) level |= -128;
+    }
+    return level;
+}
 
 class ORB_Impl : public ORB
 {
@@ -995,15 +1008,23 @@ void ORB_Impl::detectAndCompute( InputArray _image, InputArray _mask,
         // In short, ultimately the descriptor should
         // ignore octave parameter and deal only with the keypoint size.
         nLevels = 0;
+        int prevLevel = firstLevel;
         for( i = 0; i < nkeypoints; i++ )
         {
-            level = keypoints[i].octave;
-            CV_Assert(level >= 0);
-            if( i > 0 && level < keypoints[i-1].octave )
+            level = getLevel(keypoints[i]);
+            if (level < 0)
+            {
+                scaleFactor = 2; //SIFT scale factor
+                sortedByLevel = false; //Need to update octave IDs
+                //Use the negative level to adjust the firstLevel parameter
+                firstLevel = std::max(firstLevel, -level);
+            }
+            if(sortedByLevel && i > 0 && level < prevLevel)
                 sortedByLevel = false;
             nLevels = std::max(nLevels, level);
+            prevLevel = level;
         }
-        nLevels++;
+        nLevels += 1 + firstLevel;
     }
 
     std::vector<Rect> layerInfo(nLevels);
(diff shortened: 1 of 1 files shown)


Sahloul:
I suspected that this patch is not correct, therefore I closed it. Upon investigation, however, I found that this patch is just fine, but there is another bug described in #10561

modules/features2d/src/orb.cpp
```diff
@@ -650,6 +650,19 @@ static inline float getScale(int level, int firstLevel, double scaleFactor)
...
+static inline int getLevel(KeyPoint &kp)
+{
+    int level = kp.octave;
+    CV_Assert(0 <= level);
+    //SIFT encodes layer and octave together
+    //See issue #10555
+    if (level >= 255)
+    {
+        level &= 255;
+        if (level >= 128) level |= -128;
```
vpisarev: I believe, this line makes the level index negative when it's &gt; 128. Is that intended behaviour? We have tests for detectors and descriptors that check how well they preserve scale and rotation variance. Could you add such tests for SIFT + ORB combination?

modules/features2d/src/orb.cpp
```diff
@@ -995,15 +1008,22 @@ void ORB_Impl::detectAndCompute( InputArray _image, InputArray _mask,
...
         // ignore octave parameter and deal only with the keypoint size.
         nLevels = 0;
+        int prevLevel = firstLevel;
         for( i = 0; i < nkeypoints; i++ )
         {
-            level = keypoints[i].octave;
-            CV_Assert(level >= 0);
-            if( i > 0 && level < keypoints[i-1].octave )
+            level = getLevel(keypoints[i]);
+            if (level < 0) //issue #10555
```
Sahloul: @vpisarev As you can see here, the negative level value is used to adjust the `firstLevel` offset.

modules/features2d/src/orb.cpp
```diff
@@ -1109,9 +1129,10 @@ void ORB_Impl::detectAndCompute( InputArray _image, InputArray _mask,
             nkeypoints = (int)keypoints.size();
             for( i = 0; i < nkeypoints; i++ )
             {
-                level = keypoints[i].octave;
-                CV_Assert(0 <= level);
-                allKeypoints[level].push_back(keypoints[i]);
+                level = getLevel(keypoints[i]);
+                keypoints[i].octave = level + firstLevel;
+                CV_Assert(keypoints[i].octave >= 0);
```
Sahloul: @vpisarev Finally, all negative values are corrected using the `firstLevel` offset, and the assertion here ensures that the final value is just OK

modules/features2d/src/orb.cpp
Sahloul: @vpisarev Yes, this is intended behavior. Please check my other comments.
As for the any potential tests, unfortunately they will fail, because of #10561, and we cannot add them now!
However, I have carefully debugged this patch, and it is sound and correct.

Sahloul: @vpisarev The code has now some additional comments to explain the intended behavior.
Further, [this repo](https://github.com/Sahloul/opencv_contrib/commits/bug_fix/ORB) contains invaraince tests, and should be merge with this PR.

BTW: While the buildbot loaded my opencv_contrib repo, it did not execute its [perf_xfeatures2d](http://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/14690) tests...
Although I have locally confirmed that tests are passing, is your development cycle really alright?

alalek: @Sahloul You should create PR into opencv_contrib to run contrib tests.

Sahloul: @alalek All done!

modules/features2d/src/orb.cpp
```diff
@@ -650,6 +650,19 @@ static inline float getScale(int level, int firstLevel, double scaleFactor)
     return (float)std::pow(scaleFactor, (double)(level - firstLevel));
 }
 
+static inline int getLevel(KeyPoint &kp)
+{
+    int level = kp.octave;
+    CV_Assert(0 <= level);
+    //SIFT encodes layer and octave together
```
alalek: I see here this design problem:
- ORB should not have any knowledge about special cases of other descriptors/detectors, like SIFT.

We can try placing this logic into KeyPoint directly (some unified behavior with documentation).

Sahloul: @alalek
- Can we introduce the attribute `KeyPoint.layer`?
- Can we allow `KeyPoint.octave` to have a negative value?

alalek:
Source branch was removed, closing this.

PR #10568, opened 2018-01-09
core: fix unresolved symbols from utils::fs

### LINKED ISSUE #10565: glob.cpp causes build error when compiling ios --dynamic framework
- OpenCV => master
- Platform =>  OSX 10.12 (target: iOS)
- Tools =>  CMake 3.4-3.10.1 XCode 8.3-9.2
- Compiler => AppleClang 8.1.0.8020042 - 9.0.0.9000039

##### Detailed description
PR [#10154](https://github.com/opencv/opencv/pull/10154) makes a breaking change when building ios --dynamic.

The changes to `modules/core/src/glob.cpp` break dynamic builds of the ios framework..

##### steps to reproduce
from master:  
`python ./platforms/ios/build_framework.py  --dynamic`  

`opencv_world` fails to build:  

		Undefined symbols for architecture arm64:  
	  "cv::utils::fs::join(cv::String const&, cv::String const&)", 
	        referenced from:
	      glob_rec(cv::String const&, cv::String const&, std::__1::vector<cv::String, std::__1::allocator<cv::String> >&, bool, bool, cv::String const&) 
	      in glob.o

There seems to be a signature mismatch with `fs::join`

`cv::String const& ` used in glob.cpp  
`const cv::String&`  required in `fs::join`

Static framework builds are not affected.

As a temporary workaround I can revert the changes to `glob.cpp` - but fixing the signature mismatch would be an improvement.

alalek:
resolves #10565

### DIFF
--- a/modules/core/src/utils/filesystem.cpp
+++ b/modules/core/src/utils/filesystem.cpp
@@ -41,6 +41,8 @@
 #include <errno.h>
 #endif
 
+#endif // OPENCV_HAVE_FILESYSTEM_SUPPORT
+
 namespace cv { namespace utils { namespace fs {
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


alalek:
:+1:

PR #10569, opened 2018-01-09
Pass RANSAC parameters as function input

branka-plateiq:
### This pullrequest changes

This pull request changes estimateRigidTransform so it takes as input some RANSAC parameters instead of having them hardcoded in the function.

### DIFF
--- a/modules/video/include/opencv2/video/tracking.hpp
+++ b/modules/video/include/opencv2/video/tracking.hpp
@@ -250,8 +250,9 @@ when fullAffine=false.
 @sa
 estimateAffine2D, estimateAffinePartial2D, getAffineTransform, getPerspectiveTransform, findHomography
  */
-CV_EXPORTS_W Mat estimateRigidTransform( InputArray src, InputArray dst, bool fullAffine, int RANSAC_MAX_ITERS=500, double RANSAC_GOOD_RATIO=0.5,
-                                         int RANSAC_SIZE0=3);
+CV_EXPORTS_W Mat estimateRigidTransform( InputArray src, InputArray dst, bool fullAffine);
+CV_EXPORTS_W Mat estimateRigidTransform( InputArray src, InputArray dst, bool fullAffine, int ransacMaxIters, double ransacGoodRatio,
+                                         int ransacSize0);
 
 
 enum
--- a/modules/video/include/opencv2/video/tracking.hpp
+++ b/modules/video/include/opencv2/video/tracking.hpp
@@ -250,7 +250,8 @@ when fullAffine=false.
 @sa
 estimateAffine2D, estimateAffinePartial2D, getAffineTransform, getPerspectiveTransform, findHomography
  */
-CV_EXPORTS_W Mat estimateRigidTransform( InputArray src, InputArray dst, bool fullAffine );
+CV_EXPORTS_W Mat estimateRigidTransform( InputArray src, InputArray dst, bool fullAffine, int RANSAC_MAX_ITERS=500, double RANSAC_GOOD_RATIO=0.5,
+                                         int RANSAC_SIZE0=3);
 
 
 enum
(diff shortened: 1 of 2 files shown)


hrnr:
please check `estimateAffine2D` and `estimateAffinePartial2D`. I think we should deprecate `estimateRigidTrasform` in video module.

modules/video/src/lkpyramid.cpp
```diff
@@ -1607,8 +1605,8 @@ cv::Mat cv::estimateRigidTransform( InputArray src1, InputArray src2, bool fullA
     for( k = 0; k < RANSAC_MAX_ITERS; k++ )
     {
         int idx[RANSAC_SIZE0];
-        Point2f a[RANSAC_SIZE0];
-        Point2f b[RANSAC_SIZE0];
+        Point2f *a =new Point2f[RANSAC_SIZE0];
```
csukuangfj: `std::vector<Point2f> a(RANSAC_SIZE0);` may be better, since it
abstracts away new/delete.

vpisarev: I agree; also, those vectors should better be placed outside of the loop.

modules/video/include/opencv2/video/tracking.hpp
```diff
@@ -250,7 +250,8 @@ when fullAffine=false.
 @sa
 estimateAffine2D, estimateAffinePartial2D, getAffineTransform, getPerspectiveTransform, findHomography
  */
-CV_EXPORTS_W Mat estimateRigidTransform( InputArray src, InputArray dst, bool fullAffine );
+CV_EXPORTS_W Mat estimateRigidTransform( InputArray src, InputArray dst, bool fullAffine, int RANSAC_MAX_ITERS=500, double RANSAC_GOOD_RATIO=0.5,
+                                         int RANSAC_SIZE0=3);
```
vpisarev: In OpenCV 3.x we preserve binary compatibility. So you need to add another function with those extra parameters.

modules/video/src/lkpyramid.cpp
```diff
@@ -1487,17 +1487,15 @@ getRTMatrix( const Point2f* a, const Point2f* b,
...
+cv::Mat cv::estimateRigidTransform( InputArray src1, InputArray src2, bool fullAffine, int RANSAC_MAX_ITERS, double RANSAC_GOOD_RATIO,
+                                    int RANSAC_SIZE0)
 {
     CV_INSTRUMENT_REGION()
 
     Mat M(2, 3, CV_64F), A = src1.getMat(), B = src2.getMat();
 
     const int COUNT = 15;
     const int WIDTH = 160, HEIGHT = 120;
-    const int RANSAC_MAX_ITERS = 500;
```
vpisarev: Since you now pass parameters instead of using constants, there should be some code added to check the parameters for reasonable value ranges

hrnr:
@vpisarev: What do you think about `estimateAffine2D` and `estimateAffinePartial2D` in calib3d module? I think they provide all this functionality (+ some more) and are much cleaner. 

We might want to remove this function in OpenCV 4.0.

branka-plateiq:
@hrnr I will check estimateAffine2D and estimateAffinePartial2D and see if they give me the functionality I need. If they do, I will close this, if not I will make the required changes. @vpisarev @csukuangfj  thanks for the review

branka-plateiq:
@hrnr estimateAffine2D and estimateAffinePartial2D do not have the option of setting the minimum acceptable percentage of inliers for a model to be valid. I would like to go ahead with this PR because that is what I need and it was easy to add it here.

vpisarev:
let's finally merge it in. It's for 3.4 branch; in OpenCV 4.0 we will probably revise RANSAC kernels and remove redundant functionality :+1:

PR #10570, opened 2018-01-10
Multichannel Mat::ones and UMat::ones initialization

dkurt:
### This pullrequest changes

Before PR
```cpp
std::cout << cv::Mat::ones(3, 4, CV_8UC3) << std::endl;
```
produced
```
[  1,   0,   0,   1,   0,   0,   1,   0,   0,   1,   0,   0;
   1,   0,   0,   1,   0,   0,   1,   0,   0,   1,   0,   0;
   1,   0,   0,   1,   0,   0,   1,   0,   0,   1,   0,   0]
```

### DIFF
--- a/modules/core/test/test_mat.cpp
+++ b/modules/core/test/test_mat.cpp
@@ -1754,3 +1754,55 @@ TEST(Mat_, template_based_ptr)
 }
 
 #endif
+
+namespace cvtest {
+
+typedef testing::TestWithParam<tuple<int, Size, int> > Mat_init;
+TEST_P(Mat_init, ones)
+{
+    int depth = get<0>(GetParam());
+    Size size = get<1>(GetParam());
+    int ch = get<2>(GetParam());
+
+    Mat m = Mat::ones(size, CV_MAKETYPE(depth, ch));
+    ASSERT_EQ(countNonZero(m != 1), 0);
+}
+
+TEST_P(Mat_init, eye)
+{
+    int depth = get<0>(GetParam());
+    Size size = get<1>(GetParam());
+    int ch = get<2>(GetParam());
+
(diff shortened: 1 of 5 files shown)


modules/core/test/test_mat.cpp
```diff
@@ -1754,3 +1754,24 @@ TEST(Mat_, template_based_ptr)
...
+    int ch = get<2>(GetParam());
+
+    Mat m = Mat::ones(size, CV_MAKETYPE(depth, ch));
+    ASSERT_EQ(countNonZero(m != 1), 0);
+}
+
+INSTANTIATE_TEST_CASE_P(Core, Mat_init, Combine(
+        Values(CV_8U, CV_8S, CV_16S, CV_32S, CV_32F, CV_64F), // depth
+        Values(Size(3, 3), Size(16, 8)),  // input size
+        Values(1, 2, 3, 4)  // number of channels
```
csukuangfj: Does it work for matrices of more than 4 channels?

dkurt: @csukuangfj, it doesn't because of https://github.com/opencv/opencv/blob/7349b8f5cecfc6eb649b224e58da16d137c3c578/modules/core/src/matrix.cpp#L1168 .

alalek:
BTW, There is also "eye()" call with similar convention.

vpisarev:
it's not a trivial decision. While I'd vote for your change, @dkurt, in Mat::ones, I think that for Mat::eye() I would retain the current behaviour, because people may want to create a complex-valued matrix with 1's on the diagonal. :+1:

dkurt:
We've decided to close it to keep complex matrices initialization: `Mat::ones` fills by ones only real parts of values.

PR #10574, opened 2018-01-11
Newton's method can be more efficient

razerhell:
when we get the result of function distortPoint with a point (0, 0) and then undistortPoint with  the result, we get the point not (0, 0). and then we discovered that the old method is not convergence sometimes. finally we have gotten the right values by Newton's method.

### DIFF
--- a/modules/calib3d/src/fisheye.cpp
+++ b/modules/calib3d/src/fisheye.cpp
@@ -388,19 +388,20 @@ void cv::fisheye::undistortPoints( InputArray distorted, OutputArray undistorted
         {
             // compensate distortion iteratively
             double theta = theta_d;
-	    const double EPS = 1e-8;
-           
-	    for (int j=0; j<10; j++)
-	    {
-		double theta2 = theta*theta, theta4 = theta2*theta2, theta6 = theta4*theta2, theta8 = theta6*theta2;
-		double k0_theta2 = k[0] * theta2, k1_theta4 = k[1] * theta4, k2_theta6 = k[2] * theta6, k3_theta8 = k[3] * theta8;
-		double _theta = theta - (theta * (1 + k0_theta2 + k1_theta4 + k2_theta6 + k3_theta8) - theta_d) /
-					 (1 + 3*k0_theta2 + 5*k1_theta4 + 7*k2_theta6 + 9*k3_theta8);
-		if (EPS > fabs(_theta - theta))
-		{ break; }
-		theta = _theta;
-	    }
-            
+
+            const double EPS = 1e-8; // or std::numeric_limits<double>::epsilon();
+            for (int j = 0; j < 10; j++)
+            {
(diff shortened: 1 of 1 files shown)


modules/calib3d/src/fisheye.cpp
```diff
@@ -388,12 +388,19 @@ void cv::fisheye::undistortPoints( InputArray distorted, OutputArray undistorted
...
-
+            int j;
+			double eps0 = 1e-8, eps = 1;
+           
+			for (j=0; j<10 && eps > eps0; j++)
+			{
+				double theta2 = theta*theta, theta4 = theta2*theta2, theta6 = theta4*theta2, theta8 = theta6*theta2;
+				double theta3 = theta*theta2, theta5 = theta4*theta, theta7 = theta6*theta, theta9 = theta8*theta;
+				double _theta = theta - (theta + k[0]*theta3 + k[1]*theta5 + k[2]*theta7 + k[3]*theta9 - theta_d) /
+						 (1 + 3*k[0]*theta2 + 5*k[1]*theta4 + 7*k[2]*theta6 + 9*k[3]*theta8);
```
alalek: 1. Please use 4 spaces [instead](http://pullrequest.opencv.org/buildbot/builders/precommit_docs/builds/13870/steps/whitespace%20opencv/logs/stdio) of tabs.
2. Eliminate odd powers of theta by replacing of
```.diff
-double theta3 = theta*theta2, theta5 = theta4*theta, theta7 = theta6*theta, theta9 = theta8*theta;
- ... theta + k[0]*theta3 + k[1]*theta5 + k[2]*theta7 + k[3]*theta9 ...
+ ... theta * (1 + k[0]*theta2 + k[1]*theta4 + k[2]*theta6 + k[3]*theta8) ...
```
3. don't declare j outside of for loop
4. eps check is better to perform in the for body; `eps0` is constant and should be UPPERCASE:
```
const double EPS = 1e-8;
for (int j = 0; j < 10; j++)
{
    double theta2 = theta*theta, theta4 = theta2*theta2, theta6 = theta4*theta2, theta8 = theta6*theta2;
    double theta_diff = (theta * (1 + k[0]*theta2 + k[1]*theta4 + k[2]*theta6 + k[3]*theta8) - theta_d) /
                        (1 + 3*k[0]*theta2 + 5*k[1]*theta4 + 7*k[2]*theta6 + 9*k[3]*theta8);
    theta = theta - theta_diff;
    if (fabs(theta_diff) <= EPS)
        break;
}
```
5. You could extract common sub-expressions:
```
double k0_theta2 = k[0]*theta2, k1_theta4 = k[1]*theta4, k2_theta6 = k[2]*theta6, k3_theta8 = k[3]*theta8;
```

alalek:
@razerhell I pushed an update to this patch. Could you check this on your dataset?

PR #10575, opened 2018-01-11
Added getFontScaleFromHeight()

PhilLab:
Adding a function to compute font scale for a desired font height in pixels

### DIFF
--- a/modules/imgproc/include/opencv2/imgproc.hpp
+++ b/modules/imgproc/include/opencv2/imgproc.hpp
@@ -4730,6 +4730,20 @@ CV_EXPORTS_W Size getTextSize(const String& text, int fontFace,
                             double fontScale, int thickness,
                             CV_OUT int* baseLine);
 
+
+/** @brief Calculates the font-specific size to use to achieve a given height in pixels.
+
+@param fontFace Font to use, see cv::HersheyFonts.
+@param pixelHeight Pixel height to compute the fontScale for
+@param thickness Thickness of lines used to render the text.See putText for details.
+@return The fontSize to use for cv::putText
+
+@see cv::putText
+*/
+CV_EXPORTS_W double getFontScaleFromHeight(const int fontFace,
+                                           const int pixelHeight,
+                                           const int thickness = 1);
+
 /** @brief Line iterator
 
 The class is used to iterate over all the pixels on the raster line
--- a/modules/imgproc/src/drawing.cpp
+++ b/modules/imgproc/src/drawing.cpp
@@ -2364,6 +2364,17 @@ Size getTextSize( const String& text, int fontFace, double fontScale, int thickn
     return size;
 }
 
+double getFontScaleFromHeight(const int fontFace, const int pixelHeight, const int thickness)
+{
+    // By https://stackoverflow.com/a/27898487/1531708
+    const int* ascii = getFontData(fontFace);
+
+    int base_line = (ascii[0] & 15);
+    int cap_line = (ascii[0] >> 4) & 15;
+
+    return static_cast<double>(pixelHeight - static_cast<double>((thickness + 1)) / 2.0) / static_cast<double>(cap_line + base_line);
+}
+
 }
 
 


alalek:
@PhilLab What is kind of the problem with test?

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4730,6 +4730,20 @@ CV_EXPORTS_W Size getTextSize(const String& text, int fontFace,
                             double fontScale, int thickness,
                             CV_OUT int* baseLine);
 
+
+/** @brief Calculates the font-specific size to use to achieve a given height in pixels.
```
rgolovanov: Could you please add usage sample as we have for ```getTextSize```

modules/imgproc/src/drawing.cpp
```diff
@@ -2364,6 +2364,17 @@ Size getTextSize( const String& text, int fontFace, double fontScale, int thickn
     return size;
 }
 
+double getFontScaleFromHeight(const int fontFace, const int pixelHeight, const int thickness)
+{
+    // By https://stackoverflow.com/a/27898487/1531708
```
rgolovanov: I don't think this link is helpfull here since stakoverflow is pointing to opencv:
> I have found out that the cv::getTextSize in opencv looks as follows

modules/imgproc/src/drawing.cpp
```diff
@@ -2364,6 +2364,17 @@ Size getTextSize( const String& text, int fontFace, double fontScale, int thickn
...
+    const int* ascii = getFontData(fontFace);
+
+    int base_line = (ascii[0] & 15);
+    int cap_line = (ascii[0] >> 4) & 15;
+
+    return static_cast<double>(pixelHeight - static_cast<double>((thickness + 1)) / 2.0) / static_cast<double>(cap_line + base_line);
```
rgolovanov: a bit shorter:
```
   return static_cast<double>(pixelHeight - 0.5 * (thickness + 1)) / static_cast<double>(cap_line + base_line);
```

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4730,6 +4730,20 @@ CV_EXPORTS_W Size getTextSize(const String& text, int fontFace,
...
+
+@param fontFace Font to use, see cv::HersheyFonts.
+@param pixelHeight Pixel height to compute the fontScale for
+@param thickness Thickness of lines used to render the text.See putText for details.
+@return The fontSize to use for cv::putText
+
+@see cv::putText
+*/
+CV_EXPORTS_W double getFontScaleFromHeight(const int fontFace,
```
rgolovanov: no need to pass values by const

vpisarev:
:+1:

PhilLab:
@alalek: I did not know how to properly use a RNG suitable for tests

alalek:
@PhilLab So just avoid RNG and try using some fixed numbers: 20, 50, 100

modules/imgproc/src/drawing.cpp
PhilLab: The link points to the answer containing the code for ``getFontScaleFromHeight()``

modules/imgproc/include/opencv2/imgproc.hpp
PhilLab: Personally, I like to keep things const whenever possible. Can change it if you advise, though

modules/imgproc/include/opencv2/imgproc.hpp
PhilLab: Thanks for reviewing. Yes, I can. Should I push to this branch or open a new one as it is already merged?

rgolovanov: Yes, I think we need this in separate pull request.

modules/imgproc/src/drawing.cpp
rgolovanov: Yep, but this code is pretty similar to getTextSize code. So there is no need to external links.

modules/imgproc/include/opencv2/imgproc.hpp
rgolovanov: It is minor issue, but it leads to inconsistency between public functions, e.g. comparing to getTextSize, where only String parameter is passed by const ref.

rgolovanov:
@PhilLab sorry, I somehow missed your replies.

PR #10579, opened 2018-01-12
Propagate calculated Gaussian kernel size(ref)

woodychow:
Otherwise, ipp_GaussianBlur will fail if the user doesn't specify a kernel size. Not sure if this is the best fix though.

### DIFF
--- a/modules/imgproc/src/smooth.cpp
+++ b/modules/imgproc/src/smooth.cpp
@@ -1763,7 +1763,7 @@ cv::Mat cv::getGaussianKernel( int n, double sigma, int ktype )
 
 namespace cv {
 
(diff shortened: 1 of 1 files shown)


mshabunin:
What do you mean by word _fail_? Will IPP implementation be skipped or crash or produce an error?

woodychow:
Sorry I should have been more specific. The IPP C++ wrapper will throw IwException with status code = ippStsBadArgErr because the kernel size is 0. The exception will be caught in ipp_gaussianBlurParallel, and the IPP implementation will be skipped.

https://software.intel.com/en-us/ipp-dev-reference-filtergaussianborder

With this change, the calculated kernel size is sent to the IPP path.

PR #10581, opened 2018-01-12
A bunch of minor modifications

### LINKED ISSUE #10168: Multiple conditions with commas at CV_Assert
- OpenCV => 3.3.1
- Compiler => /usr/bin/c++ (ver 5.4.0)

CV_Assert with multiple conditions defined as separate code lines.

##### Steps to reproduce

```cpp
if (true)
  CV_Assert(1 != 2);
else
  CV_Assert(1 == 2, 2 == 3);
```
produces
```
error: (-215) 2 == 3
```

mshabunin:
resolves #10168

### This pullrequest changes

- fixes _CV_Assert_ macro for static analyzers
- disabled one warning in protobuf library
- added [ThinLTO](https://clang.llvm.org/docs/ThinLTO.html) compilation option
- some minor issues

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/calib3d/src/fisheye.cpp
+++ b/modules/calib3d/src/fisheye.cpp
@@ -762,7 +762,7 @@ double cv::fisheye::calibrate(InputArrayOfArrays objectPoints, InputArrayOfArray
 
 
     //-------------------------------Optimization
-    for(int iter = 0; iter <= std::numeric_limits<int>::max(); ++iter)
+    for(int iter = 0; iter < std::numeric_limits<int>::max(); ++iter)
     {
         if ((criteria.type == 1 && iter >= criteria.maxCount)  ||
             (criteria.type == 2 && change <= criteria.epsilon) ||
--- a/modules/core/include/opencv2/core/base.hpp
+++ b/modules/core/include/opencv2/core/base.hpp
@@ -409,13 +409,14 @@ CV_INLINE CV_NORETURN void errorNoReturn(int _code, const String& _err, const ch
 #endif
 
 #ifdef CV_STATIC_ANALYSIS
+
 // In practice, some macro are not processed correctly (noreturn is not detected).
 // We need to use simplified definition for them.
 #define CV_Error(...) do { abort(); } while (0)
-#define CV_Error_(...) do { abort(); } while (0)
-#define CV_Assert(cond) do { if (!(cond)) abort(); } while (0)
+#define CV_Error_( code, args ) do { cv::format args; abort(); } while (0)
 #define CV_ErrorNoReturn(...) do { abort(); } while (0)
 #define CV_ErrorNoReturn_(...) do { abort(); } while (0)
+#define CV_Assert_1( expr ) do { if (!(expr)) abort(); } while (0)
 
 #else // CV_STATIC_ANALYSIS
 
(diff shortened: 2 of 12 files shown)


modules/core/include/opencv2/core/base.hpp
```diff
@@ -445,17 +449,16 @@ for example:
...
-*/
+/** same as CV_Error(code,msg), but does not return */
+#define CV_ErrorNoReturn( code, msg ) cv::errorNoReturn( code, msg, CV_Func, __FILE__, __LINE__ )
 
-#define CV_VA_NUM_ARGS_HELPER(_1, _2, _3, _4, _5, _6, _7, _8, _9, _10, N, ...)    N
-#define CV_VA_NUM_ARGS(...)      CV_VA_NUM_ARGS_HELPER(__VA_ARGS__, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0)
+/** same as CV_Error_(code,args), but does not return */
+#define CV_ErrorNoReturn_( code, args ) cv::errorNoReturn( code, cv::format args, CV_Func, __FILE__, __LINE__ )
 
 #define CV_Assert_1( expr ) if(!!(expr)) ; else cv::error( cv::Error::StsAssert, #expr, CV_Func, __FILE__, __LINE__ )
```
csukuangfj: Should `CV_Assert_1` to 10 be wrapped inside `do {...} while(0)`

alalek: It is from here: #10168

modules/calib3d/src/fisheye.cpp
```diff
@@ -762,7 +762,7 @@ double cv::fisheye::calibrate(InputArrayOfArrays objectPoints, InputArrayOfArray
 
 
     //-------------------------------Optimization
-    for(int iter = 0; iter <= std::numeric_limits<int>::max(); ++iter)
+    for(int iter = 0; iter < std::numeric_limits<int>::max(); ++iter)
```
alalek:
```
for(int iter = 0; ; ++iter)
```
?

modules/core/src/dxt.cpp
```diff
@@ -2162,9 +2162,7 @@ static bool ocl_dft(InputArray _src, OutputArray _dst, int flags, int nonzero_ro
     // if output format is not specified
     if (complex_output + real_output == 0)
     {
-        if (real_input)
-            real_output = 1;
-        else
+        if (!real_input)
             complex_output = 1;
```
alalek: Looks like this code is not equivalent.

Does this warning is "set but not used"?
My suggestion is to drop "real_input" and "real_output" flags (replace them to "!complex_input" and "!complex_output") - we don't support [quaternions](http://ieeexplore.ieee.org/document/615636/) or similar here.

modules/core/include/opencv2/core/base.hpp
mshabunin: I've added mentioned construction to the `CV_Assert` macro.

modules/core/src/dxt.cpp
mshabunin: I've refactored this part of function to be more readable.

mshabunin:
Rebased on top of current _master_ branch.

PR #10586, opened 2018-01-13
Added functionality to get frames per second in utility.hpp

Bhanudutta:

### DIFF
--- a/modules/core/include/opencv2/core/utility.hpp
+++ b/modules/core/include/opencv2/core/utility.hpp
@@ -380,7 +380,29 @@ public:
     sumTime = 0;
     counter = 0;
     }
-
+    /**
+    returns frames per second.
+    @param reset_point is value of counter after which TickMeter should reset.
+    */
+    CV_WRAP double getFps(int reset_point = 10)
+    {
+    static double stat_fps = 0;
+    if(counter >= reset_point)
+        reset();
+    double fps = getCounter()/getTimeSec();
+    if(std::isnan(fps))
+        return stat_fps;
+        else stat_fps = fps;
+    return fps;
+    }
+    /**
+    increases counter
+    @param by value by which counter should be increased
+    */
+    CV_WRAP void incrementCounter(int by = 1)
+    {
+    counter += by;
+    }
 private:
     int64 counter;
     int64 sumTime;
--- a/samples/cpp/tickmeter.cpp
+++ b/samples/cpp/tickmeter.cpp
@@ -1,0 +1,32 @@
+//TickMeter sample program
+#include "opencv2/opencv.hpp"
+#include <iostream>
+
+using namespace cv;
+using namespace std;
+int main(int argc, char** argv)
+{
+    VideoCapture cap;
+    if(argc>1)
+    {
+        cap.open(argv[1]);
+    }
+    else
+    {
+        cap.open(0);
+    }
+    Mat frame;
+    TickMeter t;
+    namedWindow("TickMeter",1);
+    while(true)
+    {
+        t.start();
+        cap>>frame;
+        imshow("TickMeter",frame);
+        waitKey(1);
+        t.stop();
+        int n = 30;
+        cout<<"Current FPS averaged over "<<n<<" frame = "<<t.getFps(n)<<endl;
+    }
+    return 0;
+}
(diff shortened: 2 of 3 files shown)


samples/cpp/tickmeter.cpp
```diff
@@ -0,0 +1,31 @@
...
+    TickMeter t;
+    namedWindow("TickMeter",1);
+    while(true)
+    {
+        t.start();
+        cap>>frame;
+        imshow("TickMeter",frame);
+        waitKey(1);
+        t.stop();
+        cout<<"Current FPS = "<<t.getFps()<<endl;
```
alalek: > Current FPS

This is not true.

You calculate here an average FPS for all processed frames.

BTW, This metric is not very useful:
- long initialization of the first frame (buffers allocation/initialization) affects metric for the large number of the next frames
- in classical object detection apps processing time depends on number of objects on the input frame

Usually FPS is counted as an average for the last N frames:
- N=1: "true" value of the last processed frame (this is "current FPS"). Cons of this approach is very non-stable value (it changes very quickly, so it is not usable with UI).
- N=10: more stable result; corresponding to the contents of current frames instead of "global" average (for real-time processing app (like 20 FPS), lag is about 0.5 sec).

Bhanudutta:
Thanks for your feeback , but if do it other way , we can calculate total frames in a second and print that each second 
The only issue is that than fps will be provided at an interval of 1 second so if some process completes in time less than a second it will don't give us any result so I used this method , but I think you are right , the metric is not so useful , so I should think more for some better solution

samples/cpp/tickmeter.cpp
Bhanudutta: Thanks for your feeback , but if do it other way , we can calculate total frames in a second and print that each second
The only issue is that than fps will be provided at an interval of 1 second so if some process completes in time less than a second it will don't give us any result so I used this method , but I think you are right , the metric is not so useful , so I should think more for some better solution

dkurt: Sometimes algorithms can process several frames at the same time. I mean between `tm.start()` and `tm.stop()` we can process more than one frame. In example, deep learning networks can work with a batch of frames (a time series or just to improve a whole efficiency). However `counter` is incremented just at `tm.stop()`.

Bhanudutta: I think 2 solutions are possible : 
1. increment counter each time frame is read from VideoCapture , and not at the time of calling t.stop() but the issue will be then that if frames are read from any other source for example an array of Mat then it will become senseless.
2.add a public function called increaseCounter() in TickMeter class and then user can independently call it . 

I think solution 2 might work also I will change the algorithm to take average over 10 frames as told by @alalek

Bhanudutta:
I have made the required changes in the commit , 3 extra commits were made accidently

modules/core/include/opencv2/core/utility.hpp
```diff
@@ -380,7 +380,29 @@ class CV_EXPORTS_W TickMeter
     sumTime = 0;
     counter = 0;
     }
-
+    /**
+    returns frames per second.
+    @param reset_point is value of counter after which TickMeter should reset.
+    */
+    CV_WRAP double getFps(int reset_point = 10)
```
vpisarev: can you move implementations to some .cpp file? We would like to keep .hpp files as clean as possible

modules/core/include/opencv2/core/utility.hpp
```diff
@@ -380,7 +380,29 @@ class CV_EXPORTS_W TickMeter
...
+    if(std::isnan(fps))
+        return stat_fps;
+        else stat_fps = fps;
+    return fps;
+    }
+    /**
+    increases counter
+    @param by value by which counter should be increased
+    */
+    CV_WRAP void incrementCounter(int by = 1)
```
vpisarev: same thing

samples/cpp/tickmeter.cpp
```diff
@@ -0,0 +1,32 @@
...
+        cap.open(argv[1]);
+    }
+    else
+    {
+        cap.open(0);
+    }
+    Mat frame;
+    TickMeter t;
+    namedWindow("TickMeter",1);
+    while(true)
```
vpisarev: for infinite loops it's better to use `for(;;)` to avoid any compiler warnings

vpisarev:
let's close it for now

vpisarev:
:+1:

PR #10588, opened 2018-01-13
dnn::blobFromImage with OutputArray

dkurt:
### This pullrequest changes

Let call `blobFromImage` with pre-allocated blob.


dkurt:
@alalek, There is a copying at `net.setInput(blob)`:
https://github.com/opencv/opencv/blob/0c00652f6b4a7e48a4aab2d766e9dcc1dca920d1/modules/dnn/src/dnn.cpp#L1922:L1930 anyway. This patch might be useful to use `blobFromImage` with a single blob object to reduce number of memory allocations.

PR #10591, opened 2018-01-14
Make DNN crop layer match Caffe behavior when offset parameters unspecified

### LINKED ISSUE #10577: DNN crop layer behavior when "offset" parameter is unspecified
- OpenCV => 3.4.0

##### Detailed description

In the recent versions of Caffe, the crop layer offset parameters default to 0 if unspecified, which crops to the upper-left corner. However, the OpenCV dnn module crop layer implementation does a "crop to center" operation instead if there are no offset parameters. This results in different behavior for OpenCV vs. Caffe when importing Caffe networks that use a crop layer with no offset parameter specified, such as:

```
layer {
    name: "myCropLayer"
    type: "Crop"
    bottom: "imageToCrop"
    bottom: "imageRef"
    top: "imageCropped"
}
```

drkoller:
This resolves issue #10577.

Before this PR, the OpenCV dnn Crop layer implemented a "crop to center" effect if no `offset` parameter was specified. This change now matches the behavior of recent versions of Caffe, which use a default value of zero if no `offset` parameter is specified, resulting in a "crop to upper left" effect.

Some older versions of the Caffe Crop layer did do a "crop to center". It would be nice if a standardized parameter to select that behavior emerges, in which case it could be added back into the OpenCV dnn module as appropriate.

### DIFF
--- a/modules/dnn/src/layers/crop_layer.cpp
+++ b/modules/dnn/src/layers/crop_layer.cpp
@@ -110,26 +110,12 @@ public:
         }
 
         crop_ranges.resize(dims, Range::all());
-        for (int i = 0; i < dims; i++)
+        for (int i = start_axis; i < dims; i++)
         {
-            if( i < start_axis )
-                continue;
+            if (offset_final[i] < 0 || offset_final[i] + inpSzBlob.size[i] > inpBlob.size[i])
+                CV_Error(Error::StsBadArg, "invalid crop parameters or blob sizes");
 
-            if (!offset.empty()) //normal case
-            {
-                if (offset_final[i] < 0 || offset_final[i] + inpSzBlob.size[i] > inpBlob.size[i])
-                    CV_Error(Error::StsBadArg, "invalid crop parameters");
-
-                crop_ranges[i] = Range(offset_final[i], offset_final[i] + inpSzBlob.size[i]);
-            }
-            else //detect offset automatically so that cropped image is center of original one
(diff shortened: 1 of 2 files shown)


dkurt:
@drkoller, Thank you! It'd be nice to have some unit tests for it:
1. With different values of `axis` parameter (let both inputs are 4D so axis value in range [0, 3])
2. With no offsets
3. With a single offset value. This way this value should be applied for all cropped axes
4. With number of offsets equals to number of cropped axes

See https://github.com/BVLC/caffe/blob/2cbc1bba0922c29241e2474dd43b180be265229f/src/caffe/proto/caffe.proto#L623

I don't think that we need to obtain reference outputs from Caffe because layer is pretty simple. As for me it looks similarly to changes from https://github.com/opencv/opencv/pull/10585 so I'd like to suggest you write a parametric test.

drkoller:
@dkurt, Okay, I added a parametric unit test for the DNN crop layer, as you suggested.

dkurt:
@drkoller, thank you! Looks good for me. Let's just resolve a merge conflict and squash all commits into one.

drkoller:
@dkurt, Okay, I resolved the merge conflict and squashed the commits.

dkurt:
@alalek, may I ask you to run tests manually because of `master` source branch's name? Thanks!

dkurt:
:+1:

Issue #10321, opened 2017-12-15
RelWithDebInfo and MinSizeRel build types only work if opencv_world exists (patch included)


Earthwings:
##### System information (version)
- OpenCV => 3.3.1 (3.3.1-0xenial-20171106-101742-0800)
- Operating System / Platform => Ubuntu 16.04.3 LTS 64 bit
- Compiler => gcc (Ubuntu 5.4.0-6ubuntu1~16.04.5) 5.4.0 20160609

##### Detailed description

When using cmake and RelWithDebInfo as build type, I get errors like this:
```
make[2]: *** No rule to make target 'opencv_calib3d-NOTFOUND', needed by '...'.  Stop.
```

The reason seems to be that MAP_IMPORTED_CONFIG_RELWITHDEBINFO is set even if the previous target properties IMPORTED_LOCATION_RELEASE is not evaluated (no opencv_world here). Moving the scope of the MAP_IMPORTED_CONFIG_RELWITHDEBINFO in the upper section fixes it.

```
--- /opt/ros/kinetic/share/OpenCV-3.3.1/OpenCVConfig.cmake	2017-12-15 15:42:16.134636914 +0100
+++ OpenCVConfig.cmake	2017-12-15 15:39:03.629682333 +0100
@@ -198,14 +198,14 @@
         IMPORTED_LOCATION_RELEASE "${__location_release}"
       )
     endif()
-    # OpenCV supports Debug and Release only.
-    # RelWithDebInfo and MinSizeRel are mapped to Release
-    if(TARGET ${__cvcomponent})
-        set_target_properties(${__cvcomponent} PROPERTIES
-            MAP_IMPORTED_CONFIG_MINSIZEREL "Release"
-            MAP_IMPORTED_CONFIG_RELWITHDEBINFO "Release"
-        )
-    endif()
+  endif()
+  # OpenCV supports Debug and Release only.
[OpenCVConfig.cmake.diff.txt](https://github.com/opencv/opencv/files/1563073/OpenCVConfig.cmake.diff.txt)

+  # RelWithDebInfo and MinSizeRel are mapped to Release
+  if(TARGET ${__cvcomponent})
+      set_target_properties(${__cvcomponent} PROPERTIES
+          MAP_IMPORTED_CONFIG_MINSIZEREL "Release"
+          MAP_IMPORTED_CONFIG_RELWITHDEBINFO "Release"
+      )
   endif()
 endforeach()
```

##### Steps to reproduce

[...]

PR #10597, opened 2018-01-15
Fix mapping RelWithDebInfo/MinSizeRel to Release

Robzz:
### This pullrequest changes

When building in RelWithDebInfo mode, the build completes successfully but any other library trying to link with OpenCV will fail with the same errors as described in #10321. The cause seems to be that the configs to map stored in `OPENCV_MAP_IMPORTED_CONFIG` are written in all caps. I suspect this is a mistake, as the variable is matched against a regex, and the captured match in then converted to capitals. (`string(TOUPPER "${__remap_config}" __remap_config_upper)`).

In any case, this PR fixes the problem for me.

I'm on Windows + VS2017 (using VS2015 toolset). I only fixed this for my setup, and looking at the code, I think other setups (e.g. Linux/OS X) would encounter the same issue. I'll get someone to try it out so that I can push fixes for the other platforms if it turns out they need it, so please don't merge this until I had the time to do that :)

### DIFF
--- a/cmake/templates/OpenCVConfig.cmake.in
+++ b/cmake/templates/OpenCVConfig.cmake.in
@@ -134,7 +134,7 @@ if(NOT DEFINED OPENCV_MAP_IMPORTED_CONFIG)
     # By default CMake maps these configuration on the first available (Debug) which is wrong.
     # Non-Debug build of Application can't be used with OpenCV Debug build (ABI mismatch problem)
     # Add mapping of RelWithDebInfo and MinSizeRel to Release here
-    set(OPENCV_MAP_IMPORTED_CONFIG "RELWITHDEBINFO=!Release;MINSIZEREL=!Release")
+    set(OPENCV_MAP_IMPORTED_CONFIG "RelWithDebInfo=!Release;MinSizeRel=!Release")
   endif()
 endif()
 set(__remap_warnings "")


alalek:
What is value of "__available_configurations" on your build configuration?
CMake version?

Robzz:
Modifying the installed `OpenCVConfig.cmake` (from building with this PR) file as such:
```cmake
macro(ocv_map_imported_config target)
  if(DEFINED OPENCV_MAP_IMPORTED_CONFIG) # list, "RELWITHDEBINFO=Release;MINSIZEREL=Release"
    get_target_property(__available_configurations ${target} IMPORTED_CONFIGURATIONS)
    message("__available_configurations: ${__available_configurations}")
```
Outputs a whole lot of `__available_configurations: RELWITHDEBINFO`, I'm guessing one per built OpenCV module.

My build configuration is CMake 3.10.1, Visual Studio 15.5.3 using the VS2015 toolset and CUDA 9.1. Feel free to ask for any additional relevant information.

alalek:
Thank you for information! But this patch is wrong.
1. OpenCV supported builds are Debug and Release only. Other build configurations are not supported at all and proper compiler flags are not applied. 
2. Left side of expression is correct and should match to `__available_configurations`.
3. In your case (non-default OpenCV build) add this CMake variable:
```
OPENCV_MAP_IMPORTED_CONFIG=""
```

PR #10599, opened 2018-01-15
Add a new interface for hough transform

take1014:
There is an interface for Hough transform with image as input, but there is no interface with using set of points as input.

I thought that if we can use Hough transform with using set of points as input, we can develop image recognition more efficiently.

### This pullrequest changes
1:Add an interface for Hough transformation with set of points as input.
2：Add regression tests.

```
**WIP**

force_builders=Docs

```


alalek:
> 100644 → 100755

Don't enable executable bit for source files.

> HoughDetectParam *rho_param

Pointer here is useless. function doesn't expect NULL or several parameters. Unpack fields of this struct into regular function params.
The same note for `theta_param`.

> int point_cnt_max, const Point2f point[]

Use InputArray (with CV_32FC2 or CV_32SC2 type). "C" pointers are not accessible from Python/Java bindings and so they should be avoided.

> int polar_cnt_max, CV_OUT HoughLinePolar hough_polar[]

Use OutputArray instead. For details check `cv::HoughLines` implementation.

> ocl/test_houghlines.cpp

There is no OpenCL code in this function, you should place this test in non-ocl test file.

take1014:
Thank you for your comment. 

I fixed the source code #10599 .

It would be greatly appreciated if you could check it even when you have time.

PR #10602, opened 2018-01-16
ocl support for new layers in dnn module

pengli:
new layers including : mvn layer,  power layer , tanh layer and channels ReLU layer
also update batch_norm layer with a new ocl kernel.

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/src/layers/elementwise_layers.cpp
+++ b/modules/dnn/src/layers/elementwise_layers.cpp
@@ -267,7 +267,6 @@ struct ReLUFunctor
 
     bool applyOCL(InputArrayOfArrays inps, OutputArrayOfArrays outs, OutputArrayOfArrays internals)
     {
-        size_t wgSize = ocl::Device::getDefault().maxWorkGroupSize();
         std::vector<UMat> inputs;
         std::vector<UMat> outputs;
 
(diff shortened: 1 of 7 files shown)


modules/dnn/src/layers/elementwise_layers.cpp
```diff
@@ -395,8 +395,29 @@ struct TanHFunctor
...
+            UMat& src = inputs[i];
+            UMat& dst = outputs[i];
+
+            ocl::Kernel kernel("TanHForward", ocl::dnn::activations_oclsrc, buildopt);
+            kernel.set(0, (int)src.total());
+            kernel.set(1, ocl::KernelArg::PtrReadOnly(src));
+            kernel.set(2, ocl::KernelArg::PtrWriteOnly(dst));
+
+            size_t gSize = src.total();
+            CV_Assert(kernel.run(1, &gSize, &wgSize, false));
```
alalek: I believe we should allow to OpenCL runtime to select appropriate group size:
```.diff
-&wgSize
+NULL
```

Any issues with this?

pengli: No specific reason to use "wgSize", NULL is also OK in this patch. code is updated.

PR #10604, opened 2018-01-16
fix: use CXX_STANDARD when extracting compiler flags for PCH with GNUCXX

csabakeszegh:
When compiling with cmake using `-DCMAKE_CXX_STANDARD=11` use `-std=gnu++11`
for PCH compiler flags, otherwise it triggers an error:

    opencv_core_Release.gch: not used because `__cplusplus' defined as ` 201103L' not ` 199711L' [-Winvalid-pch]


cmake/OpenCVPCHSupport.cmake
```diff
@@ -56,6 +56,10 @@ MACRO(_PCH_GET_COMPILE_FLAGS _out_compile_flags)
           endforeach()
         endif()
 
+        GET_TARGET_PROPERTY(_cxx_standard ${_PCH_current_target} CXX_STANDARD)
+        if (_cxx_standard)
+            LIST(APPEND ${_out_compile_flags} "-std=gnu++${_cxx_standard}")
```
alalek: CMake has these variables, so hard-coding "gnu++" is not a good idea:
```
CMAKE_CXX11_EXTENSION_COMPILE_OPTION=-std=gnu++11
CMAKE_CXX11_STANDARD_COMPILE_OPTION=-std=c++11
CMAKE_CXX14_EXTENSION_COMPILE_OPTION=-std=gnu++14
CMAKE_CXX14_STANDARD_COMPILE_OPTION=-std=c++14
CMAKE_CXX98_EXTENSION_COMPILE_OPTION=-std=gnu++98
CMAKE_CXX98_STANDARD_COMPILE_OPTION=-std=c++98
```

csabakeszegh: Thank you for the quick response I was not aware of `CMAKE_CXX11_EXTENSION_COMPILE_OPTION` it's definitely better than hard coding `-std=gnu++`.  I will shortly update the PR.

csabakeszegh:
CUDA support also invalidates PCHs. In https://github.com/opencv/opencv/blob/master/modules/core/CMakeLists.txt#L67  `OPENCV_TRAITS_ENABLE_DEPRECATED` not seen when compiling PCHs.  Any hint how to solve it?  Workaround would be to check `HAVE_CUDA` in `OpenCVPCHSupport.cmake` but not sure this is the right way to do it.

cmake/OpenCVPCHSupport.cmake
```diff
@@ -56,6 +56,10 @@ MACRO(_PCH_GET_COMPILE_FLAGS _out_compile_flags)
           endforeach()
         endif()
 
+        GET_TARGET_PROPERTY(_cxx_standard ${_PCH_current_target} CXX_STANDARD)
+        if (_cxx_standard)
+            LIST(APPEND ${_out_compile_flags} "${CMAKE_CXX${_cxx_standard}_EXTENSION_COMPILE_OPTION}")
```
alalek: > _EXTENSION

Why not `_STANDARD`?

> CMAKE_CXX11_EXTENSION_COMPILE_OPTION=-std=gnu++11
> CMAKE_CXX11_STANDARD_COMPILE_OPTION=-std=c++11

alalek:
PCH file should not know anything about CUDA or other specific flags (by design it should be universal).

There is no serious regression in build time on modern hardware (code parse time is usually less than 10-20% in release builds).
It is better to install "ccache" tool for active development (not sure about CUDA support).

cmake/OpenCVPCHSupport.cmake
csabakeszegh: I haven't checked cmake source when it uses `EXTENSION`, and when `STANDARD`.  On the g++ command line its `-std=gnu++11` this is why I choosed `EXTENSION`.

alalek: CMake's logic is [here](https://gitlab.kitware.com/cmake/cmake/blob/b4dfe1d8163055f9f40a4c18f5eccccb28f06559/Source/cmLocalGenerator.cxx#L1527-1557)

csabakeszegh: Updated the PR, now I check `CXX_EXTENSIONS` property to select the correct variable. On build command line I use `-DCMAKE_CXX_STANDARD=11`, together with `-DENABLE_CXX11=1`. It turns `CMAKE_CXX_EXTENSIONS=OFF` (cmake's default is `ON` and complains that `__STRICT_ANSI__` not defined)

PR #10607, opened 2018-01-16
dnn : Added an imagesFromBlob method to the dnn module

Pandinosaurus:
Added an std::vector < cv::Mat > imagesFromBlob( cv::Mat blob )  method to the dnn module. 

Since the 4 dimensional cv::Mat may not be trivial to understand and parse, this method helps to retrieve the images contained in a blob via a more classical data structure. The retrieved images could be easily displayed and stored using the usual cv::imshow and cv::imwrite methods.


dkurt:
@Pandinosaurus, that's excited! Thank you for the start of this activity! Let me just review it a little bit.

dkurt:
@Pandinosaurus, Am I right that proposed `imagesFromBlob` just splits a blob by channels? However it's name looks like an inversion to `blobFromImages`. I mean we can expect that `imagesFromBlob(blobFromImages(images)) == images`.
I was wondering to see some usage opposite to `blobFrom*` like described above. There are some networks that return blob with an RGB image. So it'd be nice to have some method to make an interleaved image from it.
Actually I can't think up applications for retrieved channels.

Pandinosaurus:
You are right.
The idea was to provide an easy way to gather the outputs of a given layer after a forward pass. 

Maybe a name like "channelsFromBlob" would be more appropriate?

Maybe we could use the merge() method to get a RGB image once the channels have been gathered as 2 dimensional arrays?

dkurt:
@Pandinosaurus, I agree. Let's highlight two possible ways:
1. `std::vector<Mat> imageFromBlob(Mat blob)` takes 4D blob and returns separate channels. To make an interleaved image we can call `cv::merge`. The one problem is in batch dimension. In example, if we have blob of size `2x3xHxW` it returns x6 `cv::Mat`.
2. `std::vector<Mat> imageFromBlob(Mat blob)` that returns number of `cv::Mat` equals to batch size and an every Mat is an interleaved image with corresponding number of channels. We can access to separate channels by `cv::split` this way.

As you can see both points have an extra call of cv-function out the `imageFromBlob` depends on what we want to use further. However the first one requires access to blob's batch size.

Do you mind to implement `imageFromBlob` in a second way?

Pandinosaurus:
@dkurt, Sure, I will implement it the second way. It seems way more natural!

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -732,6 +732,33 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+    *  @param outputSize_ the size of the output image. Set to cv::Size(-1,-1) for no resizing operation.
+    *  @param scaleFactor_ diviser for @p images values. Should be the same value used with the blboFromImage(s) 
+    *         methods. the scaleFactor_ Will not be applied if equal to 0 (= default).
+    *  @param mean_ scalar with mean values which are added from channels. Expected to be used with 3 channel output 
+    *               images (RGB, BGR). Should be the same value used with the blboFromImage(s) methods. Set to 
+    *               cv::Scalar(-1,-1,-1) for not use it.
+    *  @returns vector of mat containing the multi channels images extracted from the blob in floating point
+    *           precision (CV_32F).
+    */
+    CV_EXPORTS_W std::vector<Mat> imagesFromBlob(const cv::Mat& blob_, const bool normalize_ = true,
```
dkurt: Let's go from the following test case:
```cpp
Mat img(rows, cols, CV_32FC3);
Mat blob = blobFromImage(img, 1.0, Size(), Scalar(), false, false);
std::vector<Mat> imgs = imagesFromBlob(blob);
normAssert(img, imgs[0]);
```

I don't think we need inverse mean because in general case blob can have any number of channels but `cv::Scalar` no more than 4 [by definition](https://docs.opencv.org/master/d1/da0/classcv_1_1Scalar__.html). This way the rest of post-processing arguments are useless too. I think interleaving is the most complicated procedure so it's better to leave other operations out of this function.

modules/dnn/src/dnn.cpp
```diff
@@ -191,6 +191,101 @@ void blobFromImages(InputArrayOfArrays images_, OutputArray blob_, double scalef
...
+        {
+            //store the current channel
+            cv::Mat currentChannel(width, height, CV_32F);
+
+            for (int w = 0; w < width; w++)
+            {
+                for (int h = 0; h < height; h++)
+                {
+                    int indx[4] = { i, c, h, w };
+                    currentChannel.at<float>(h, w) = blob_.at<float>(indx); //blobs store images in floating point precision
```
dkurt: That could be done much more easier like
```cpp
for (int i = 0; i < nbOfImages; ++i)
{
    std::vector<Mat> vectorOfChannels(nbOfChannels);
    for (int c = 0; c < nbOfChannels; ++c)
        vectorOfChannels[c] = slice(blob_, n, c);  // function from shape_utils.hpp
}
```

Pandinosaurus: Am I right if I assume you wanted to use `Mat getPlane(const Mat &m, int n, int cn)` instead of the quoted `Mat slice(const Mat &m, const _Range &r0, const _Range &r1)`?

dkurt: `slice` can manage it. There is a constructor of cv::Range for a single value, https://docs.opencv.org/master/d9/ddc/structcv_1_1dnn_1_1__Range.html#a5339946e5ddf17692f62956c28cbe623.

Pandinosaurus: I agree, but as far as I understand it, `slice` will output a 4D array (1,1,size,size) for each channel. This will result in a (1,c,size,size) array for each image using `cv::merge()`, where c is the number of channels. This is fine if we want to output a blob per image, but it does not allow us to directly use the images with the OpenCV functions made for 2D arrays (e.g. `cv::imshow()` and `cv::imwrite()`).

I propose to add a boolean flag to let the user choose if he prefers to get a 4D or a 2D array per image contained in a blob.

```c++
//slice channels
std::vector<Mat> vectorOfChannels(nbOfChannels);
for (int c = 0; c < nbOfChannels; ++c)
{
	if(get2DArrays_) vectorOfChannels[c] = getPlane(blob_, i, c); // function from shape_utils.hpp in dnn module, a plane is 2D array
	else vectorOfChannels[c] = slice(blob_, i, c); // function from shape_utils.hpp in dnn module, a slice is a 4D array
}//loop on channels
```

dkurt: @Pandinosaurus, Thanks, I see. You're completely right. We need to use `getPlane`:
```cpp
std::vector<int> blobSize(4);
blobSize[0] = 2;  // batch size
blobSize[1] = 3;  // channels
blobSize[2] = 4;  // height
blobSize[3] = 5;  // width
Mat blob(blobSize, CV_32F);

std::vector<Mat> imgs(blob.size[0]);
for (int n = 0; n < blob.size[0]; ++n)
{
    std::vector<Mat> channels(blob.size[1]);
    for (int c = 0; c < blob.size[1]; ++c)
    {
        channels[c] = getPlane(blob, n, c);
    }
    cv::merge(channels, imgs[n]);
}
```

vpisarev:
I'd suggest to pass the output `vector<Mat>` as the output parameter (OutputArrayOfArrays) rather than the return value. This way you would avoid memory reallocation on each call of the function. Also, I'm not sure if we need all those processing options like scaling, mean subtraction, normalization etc. It's obviously needed for the inverse conversion: blobFromImages, but I don't think we need it here.

modules/dnn/src/dnn.cpp
```diff
@@ -191,6 +191,37 @@ void blobFromImages(InputArrayOfArrays images_, OutputArray blob_, double scalef
...
+	std::vector<cv::Mat> images( blob_.size[0] );
+	for (int n = 0; n <  blob_.size[0]; ++n)
+	{
+		std::vector<Mat> vectorOfChannels(blob_.size[1]);
+		for (int c = 0; c < blob_.size[1]; ++c)
+		{
+			vectorOfChannels[c] = getPlane(blob_, n, c);
+		}
+		cv::merge(vectorOfChannels, images[n]);
+		CV_Assert(images[n].channels() == blob_.size[1]);
```
dkurt: I think the next three assertions isn't so useful.

vpisarev:
@Pandinosaurus, please, replace all tabs with 4 spaces. Also, move vectorOfChannels outside the outer loop

modules/dnn/src/dnn.cpp
```diff
@@ -191,6 +191,31 @@ void blobFromImages(InputArrayOfArrays images_, OutputArray blob_, double scalef
...
+
+    //A blob is a 4 dimensional matrix in floating point precision
+    //blob_[0] = batchSize = nbOfImages
+    //blob_[1] = nbOfChannels
+    //blob_[2] = height
+    //blob_[3] = width
+    CV_Assert(blob_.depth() == CV_32F);
+    CV_Assert(blob_.dims == 4);
+
+    images_.create(cv::Size(1, blob_.size[0]), blob_.depth());
```
dkurt: Please add a test for new function at https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_misc.cpp . I suggest you to generate several 3-channel images with `CV_32F` depth. Then create a blob using `blobFromImages` with no resize, mean subtraction, scale or channels swapping. Pass it to `imagesFromBlob` and check that output matches with origin images.

modules/dnn/test/test_misc.cpp
```diff
@@ -37,4 +37,24 @@ TEST(blobFromImage, allocated)
...
+        inputImgs[i] = Mat::ones(10, 10, CV_32FC3) * (i + 1);
+    }
+
+    cv::Mat blob = dnn::blobFromImages(inputImgs, 1., Size(), Scalar(), false, false);
+    std::vector<cv::Mat> outputImgs;
+    dnn::imagesFromBlob(blob, outputImgs);
+
+    for (int i = 0; i < nbOfImages; i++)
+    {
+        ASSERT_EQ(*inputImgs[i].data,*outputImgs[i].data);
```
dkurt: You compare two unsigned bytes here. Replace by `ASSERT_EQ(countNonZero(inputImgs[i] != outputImgs[i]), 0);`

modules/dnn/test/test_misc.cpp
```diff
@@ -37,4 +37,24 @@ TEST(blobFromImage, allocated)
...
 }
 
+TEST(imagesFromBlob, Regression)
+{
+    int nbOfImages = 3;
+
+    std::vector<cv::Mat> inputImgs(nbOfImages);
+    for (int i = 0; i < nbOfImages; i++)
+    {
+        inputImgs[i] = Mat::ones(10, 10, CV_32FC3) * (i + 1);
```
dkurt: Please use random generator like `cv::randu` to make test more robust.

dkurt:
:+1: Looks good to merge. @Pandinosaurus, thank you!

PR #10616, opened 2018-01-17
Fix #10615.

### LINKED ISSUE #10615: Conflicts of cv::InputArray::size()  and cv::InputArray::getSz() for std::array<cv::Mat, N>

- OpenCV => master

##### Detailed description

For `std::array<cv::Mat, N>`, 
 - `cv::InputArray::getSz()` returns a column vector representation, see [here][1]
```.cpp
template<std::size_t _Nm> inline
_InputArray::_InputArray(const std::array<Mat, _Nm>& arr)
{ init(STD_ARRAY_MAT + ACCESS_READ, arr.data(), Size(1, _Nm)); }
#endif
```
 - `cv::InputArray::size()` returns a row vector representation, see [here][2]

```.cpp
return sz.height==0 ? Size() : Size(sz.height, 1);
```
[2]: https://github.com/opencv/opencv/blob/master/modules/core/src/matrix.cpp#L1659
[1]: https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/mat.inl.hpp#L94

csukuangfj:
Fix #10615

### DIFF
--- a/modules/core/src/matrix.cpp
+++ b/modules/core/src/matrix.cpp
@@ -1656,7 +1656,7 @@ Size _InputArray::size(int i) const
     {
         const Mat* vv = (const Mat*)obj;
         if( i < 0 )
-            return sz.height==0 ? Size() : Size(sz.height, 1);
+            return sz.height==0 ? Size() : Size(1, sz.height);
         CV_Assert( i < sz.height );
 
         return vv[i].size();
--- a/modules/core/include/opencv2/core/matx.hpp
+++ b/modules/core/include/opencv2/core/matx.hpp
@@ -92,7 +92,7 @@ Except of the plain constructor which takes a list of elements, Matx can be init
     float values[] = { 1, 2, 3};
     Matx31f m(values);
 @endcode
-In case if C++11 features are avaliable, std::initializer_list can be also used to initialize Matx:
+In case if C++11 features are available, std::initializer_list can be also used to initialize Matx:
 @code{.cpp}
     Matx31f m = { 1, 2, 3};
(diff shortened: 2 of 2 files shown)


modules/core/src/matrix.cpp
```diff
@@ -1656,7 +1656,7 @@ Size _InputArray::size(int i) const
     {
         const Mat* vv = (const Mat*)obj;
         if( i < 0 )
-            return sz.height==0 ? Size() : Size(sz.height, 1);
+            return sz.height==0 ? Size() : Size(1, sz.height);
```
alalek: This fix is not consistent with another code. Check for `STD_VECTOR_MAT` or `STD_VECTOR_VECTOR` few lines above.

BTW, The most part of this code is not [covered](http://pullrequest.opencv.org/buildbot/export/opencv_releases/master_coverage-lin64-debug/20180117-000112_10342/coverage_html/opencv/modules/core/src/matrix.cpp.gcov.html) by tests (`std::array` requires C++11)

csukuangfj: Because the current implementation considers `std::vector<cv::Mat>` as a **row** of matrices, while `std::array<cv::Mat, N>` is treated as a **column** of matrices.

csukuangfj:
Maybe this pull request can be kept open until C++11 is enabled and test cases are
added for OpenCV.

vpisarev:
let's close it for now and maybe reopen later when we have C++ 11 enabled by default; the PR needs to be modified in any case to keep the consistent behaviour.

PR #10628, opened 2018-01-18
added cap field for VideoWriter (constructor, open, C interface) for …

### LINKED ISSUE #10623: VideoWriter support for 16-bit depth images and color-alpha color imags
- OpenCV => 3.4
- Compiler => Gcc

##### Detailed description

Allow VideoWriter to accept RGBA or 16-bit grayscale, in particular with ffmpeg backend. This requires changes in the VideoWriter constructor. 

Issue #10583 and associate PR address only the lossless aspect

eruffaldi:
(partially) resolves #10623 

If useful it can be completed by removing the assert in CvVideoWriter_FFMPEG_proxy plus creating some tests.

### This pullrequest changes
videoio module

VideoWriter interface adding cap field AFTER is_color. Cap field uses VideoCaptureModes.

### DIFF
--- a/modules/videoio/include/opencv2/videoio.hpp
+++ b/modules/videoio/include/opencv2/videoio.hpp
@@ -172,10 +172,13 @@ enum VideoCaptureProperties {
 @note Currently, these are supported through the libv4l backend only.
 */
 enum VideoCaptureModes {
+       CAP_MODE_AUTO = -1, 
        CAP_MODE_BGR  = 0, //!< BGR24 (default)
        CAP_MODE_RGB  = 1, //!< RGB24
        CAP_MODE_GRAY = 2, //!< Y8
-       CAP_MODE_YUYV = 3  //!< YUYV
+       CAP_MODE_YUYV = 3,  //!< YUYV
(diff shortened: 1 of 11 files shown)


vpisarev:
@eruffaldi, thanks for the contribution! We need to preserve binary compatibility in OpenCV 3.x, so we cannot add extra parameters to functions or classes. There is a bunch of other parameters of VideoWriter that we would like to add, such as bitrate etc. And it's better to introduce a universal way to add new video encoding parameters; but it should be done without breaking binary compatibility; see also #9192

eruffaldi:
I agree with your requirements.

An approach for keeping compatibility AND keeping the fact to invoke directly the constructor of VideoWriter without doing the sequence empty ctor, set and open, is to make the filename a URI that could even specify the requested API. 

ffmpeg://PATH?param=1....

or

resource://PATH?api=ffmpeg&...

asmorkalov:
cc @VadimLevin

asenyaev:

mshabunin:
* Branch 3.4 is obsolete
* Original issue (16-bit support in FFmpeg) is conisdered to be resolved (#22894)

PR #10629, opened 2018-01-18
Fix #10525

### LINKED ISSUE #10525: OCL perf test of MOG2 should use UMat as input
https://github.com/opencv/opencv/blob/f5862dacb681df36efded9200f77b965192ac1df/modules/video/src/bgfg_gaussmix2.cpp#L852
In bgfg_mog2, it needs UMat as input to enable OpenCL support,
but these two tests just take Mat as input.
https://github.com/opencv/opencv/blob/f5862dacb681df36efded9200f77b965192ac1df/modules/video/perf/opencl/perf_bgfg_mog2.cpp#L47
https://github.com/opencv/opencv/blob/f5862dacb681df36efded9200f77b965192ac1df/modules/video/test/ocl/test_bgfg_mog2.cpp#L67

GlueCrow:
resolves #10525
### This pullrequest changes
Do the UMat check on the output, not the input.
I am not sure if this is the best solution.

### DIFF
--- a/modules/video/src/bgfg_gaussmix2.cpp
+++ b/modules/video/src/bgfg_gaussmix2.cpp
@@ -849,7 +849,7 @@ void BackgroundSubtractorMOG2Impl::apply(InputArray _image, OutputArray _fgmask,
 #ifdef HAVE_OPENCL
     if (opencl_ON)
     {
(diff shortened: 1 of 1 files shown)


vpisarev:
the solution looks good to me :+1:

PR #10634, opened 2018-01-18
Update samples

sturkmen72:
### This pullrequest changes
updates samples.

Maybe some changes seems unnecessary but i believe doing such changes will be fine.


vpisarev:
looks good :+1:

Issue #10355, opened 2017-12-19
Build error: "cmd.exe" exited with code 1


Bodeeen:
##### System information (version)

- OpenCV => lastest from github
- Operating System / Platform => Windows 7 64 bit
- Compiler => Visual Studio 2017

##### Detailed description

Hi,

I'm trying to build the OpenCV libraries from source following these instructions https://docs.opencv.org/3.2.0/d3/d52/tutorial_windows_install.html. When I get to the point of actually building the solution in VS2017 a very large number of projects do not compile correctly and give the error in the title. Am I missing something stupidly obvious? The reason I'm not using the pre-built libraries is that I wanted to use the CUDA modules which, from what I understand, are not included in the pre-built version.

Thankful for any and all tips :)

catree:
I have successfully built OpenCV with CUDA9 recently and with VS2017 community. See this related [post](https://stackoverflow.com/q/45918446).

What you should have to change:

- in CMake, change the value of `CUDA_HOST_COMPILER` (should be by default `$(VCInstallDir)/bin`) by the path to `cl.exe`, something like: `C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.11.25503\bin\HostX64\x64\cl.exe` (**check this path on your computer**)
- in CMake, set `CUDA_NVCC_FLAGS` to `--cl-version=2017` (I don't remember if it is really needed or not)
- disable tests or perf_tests if you have some errors in these files or post the related errors here

Bodeeen:
Hi again,
thank you for your suggestions. Now the previous error is gone and the compilations seems to progess much better thatn earlier (though I didn't find the CUDA_NVCC_FLAGS in the CMake configuration?). Still though I'm having issues. When buiding the ALL_BUILD project, I get errors as shown in the attached screen-dumps.

![debug_x64](https://user-images.githubusercontent.com/14180271/34297527-e6710af0-e718-11e7-8bc6-bce1ee175f79.png)

If I still try to built the INSTALL project afterwards I get the same type of error. The "install" folder is created though and contains many seemengly correctly built libraries. But if I try to run the example programs, it complains that "The program can't start because cudart64_90.dll is missing from your computer." I'm guessing that many of the projects fail to build and that these libraries (eg. cudart64_90) are missing giving this final error,

Again thankful for any suggestions! :)

catree:
See also [here](https://github.com/opencv/opencv/issues/6716#issuecomment-323074308).

[...]

Issue #6716, opened 2016-06-22
std::tuple errors when Building OpenCV for Microsoft VS 2015 (x64)


oracle3001:
- OpenCV version: 3.x main branch
- Host OS: Windows 10
- Compiler & CMake: Visual Studio 2015 (update 1), Cuda 8RC & CMake 3.5

A large number of compiler errors related to std::tuple in lots of opencv_perf_cudaxxx... e.g opencv_perf_cudaimproc, opencv_perf_cudaarithm, opencv_perf_cudastereo, opencv_perf_cudalegacy, 

```
Severity	Code	Description	File	Project	Line	Suppression State
Error	C2610	'std::tuple<cv::Size,perf::`anonymous-namespace'::MatDepth>::tuple(const std::tuple<cv::Size,perf::`anonymous-namespace'::MatDepth> &)': is not a special member function which can be defaulted (compiling source file D:\Data\My Documents\GitHub\opencv\modules\cudaarithm\perf\perf_element_operations.cpp)	C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\tuple	opencv_perf_cudaarithm	328	



```

```
Severity	Code	Description	File	Project	Line	Suppression State
Error	C2535	'std::tuple<cv::Size,perf::`anonymous-namespace'::MatDepth,perf::`anonymous-namespace'::MatCn> &std::tuple<cv::Size,perf::`anonymous-namespace'::MatDepth,perf::`anonymous-namespace'::MatCn>::operator =(const std::tuple<cv::Size,perf::`anonymous-namespace'::MatDepth,perf::`anonymous-namespace'::MatCn> &)': member function already defined or declared (compiling source file D:\Data\My Documents\GitHub\opencv\modules\cudawarping\perf\perf_warping.cpp)	C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\tuple	opencv_perf_cudawarping	463	

```

pansk:
Hi, same problem here.
I also found there at least a third type of error:

Severity	Code	Description	Project	File	Line	Suppression State
[...]

PR #10637, opened 2018-01-18
Fix perf build with CUDA 9

catree:
This should fix build errors similar to this:

```
Error C2610 'std::tuple\<cv::Size,perf::\`anonymous-namespace'::MatDepth>::tuple(const std::tuple\<cv::Size,perf::\`anonymous-namespace'::MatDepth> &)': is not a special member function which can be defaulted (compiling source file modules\cudaarithm\perf\perf_element_operations.cpp)
```

Tested with:

- Microsoft Visual Studio Community 2017 Version 15.4.5

- Cuda compilation tools, release 9.1, V9.1.85

Related #6716 #10355 

```
force_builders=Custom
```

### DIFF
--- a/modules/cudaarithm/perf/perf_core.cpp
+++ b/modules/cudaarithm/perf/perf_core.cpp
@@ -51,6 +51,8 @@ using namespace perf;
 //////////////////////////////////////////////////////////////////////
 // Merge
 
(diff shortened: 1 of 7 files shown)


vpisarev:
:+1:

PR #10646, opened 2018-01-19
Add a new interface for hough transform

take1014:
There is an interface for Hough transform with image as input, but there is no interface with using set of points as input.                                                                                                     

 I thought that if we can use Hough transform with using set of points as input, we can develop image recognition more efficiently.

### This pullrequest changes
1:Add an interface for Hough transformation with set of points as input.
2:Add regression tests.

```
force_builders=Docs
```


alalek:
Please don't re-create PRs with the same source branch. There are no problems in Git that can't be resolved without re-opening new branch. Just use force flag with "git push".

BTW, Source branch name "master" doesn't allow to trigger automatic builds (so it is better to select proper branch names, like "hough").

take1014:
Thank you for your comment.

Sorry, I performed the wrong operation on GIT.

>BTW, Source branch name "master" doesn't allow to trigger automatic builds (so it is better to select >proper branch names, like "hough").
-> Now I created a branch named hough, derived from master.
　Should I request a pull request from 'hough' branch again?

alalek:
This is not a critical problem. We can do manual builds scheduling for 1-2 PRs. Make a "ping" via comments if we forget to do this.

take1014:
I merged branch 'master' of https://github.com/opencv/opencv .

Could you do manual build when you have time?
Although asking here may be wrong, how can I make all the builders green?

modules/imgproc/src/hough.cpp
```diff
@@ -1612,6 +1612,141 @@ void HoughCircles( InputArray _image, OutputArray _circles,
...
+
+            if( ((rho_max   - rho_min)   > 0) &&
+                ((theta_max - theta_min) > 0) &&
+                ((rho_step   > 0))            &&
+                ((theta_step > 0)) )
+            {
+                int rho_size = (int)((rho_max - rho_min) / rho_step);
+                int theta_size = (int)((TO_DEG(theta_max) - TO_DEG(theta_min)) / TO_DEG(theta_step));
+
+                int *plane = (int*)fastMalloc(sizeof(int) * rho_size * theta_size);
```
savuor: Rounding to zero can occur for `rho_size` and `theta_size`, need to check this before allocating memory.
Also it's better to use `AutoBuffer<int>` for memory allocation instead of malloc/free functions
(or even any kind of Mat to fill the data with zeros and avoid using `memset`)

savuor:
It'd be better to reuse the parts of code from `HoughLinesStandard` from `hough.cpp`.
It will make your function faster and will add missing features like nonmax filtering, sorting by confidence, etc.

savuor:
@vpisarev Do we need this new interface?

modules/imgproc/src/hough.cpp
take1014: Thank you for a review.

>t'd be better to reuse the parts of code from HoughLinesStandard from hough.cpp.
->I think so, too. Compared to my code, the HoughLinesStandard function is refined and proven. I reuse existing code and fix it.I should have reused the code from the beginning.

take1014: I fixed the code by reused the parts of code from HoughLinesStandard from hough.cpp.
Could you check this code when you have time?I know you are very busy, but I would appreciate hearing from you.

savuor:
The code compiles and works but I'd recommend you the following:
* choose shorter name (like `HoughLinesPointSet`)
* rewrite your function and `HoughLinesStandard` so that they don't duplicate each other's code but share it (through common function call for example)
* put your function before "Circle detection" section (line 886) where all Hough Lines implementations are placed (the end of file is for old interfaces)

take1014:
Thanks for the feedback. I am going to work on :
1.Rename function name.
2.Add common function for HoughLineStandard and this function.
3.I put this function before "Circle detection" section.

take1014:
>choose shorter name (like HoughLinesPointSet)
->I like this name, so I adopted this name. Thanks.

>rewrite your function and HoughLinesStandard so that they don't duplicate each other's code but share >it (through common function call for example)
->I added a common function so that the flow of existing processing does not change.
    Also I erased the memset in HoughLinesStandard and changed it to use Mat.

modules/imgproc/src/hough.cpp
```diff
@@ -883,6 +897,85 @@ void HoughLinesP(InputArray _image, OutputArray _lines,
...
+                    line.rho = static_cast<float>(min_rho) + r * (float)rho_step;
+                    line.angle = static_cast<float>(min_theta) + n * (float)theta_step;
+                    lines.push_back(Vec3d((double)accum[idx], (double)line.rho, (double)line.angle));
+                }
+
+                Mat(lines).copyTo(_lines);
+            }
+            else
+            {
+                CV_Error( Error::StsBadArg, "max must be greater than min, and step must be greater than 0" );
```
savuor: Please use `CV_Assert` instead of such if blocks. Also the message is redundant, `CV_Assert` will output almost the same in case of an error.
The same comment applies to 2 checks below.

modules/imgproc/test/test_houghlines.cpp
```diff
@@ -196,6 +219,33 @@ void BaseHoughLineTest::run_test(int type)
 #endif
 }
 
+void HoughLinesPointSetTest::run_test(void)
+{
+    Mat lines;
+    vector<Point2f> point;
```
savuor: Please add also a check for integer point coordinates (the values can be the same)

take1014:
Thanks for the feedback. I changed:
1.Changed to use CV_Assert without using If block.
2.Added integer test case.

savuor:
👍

modules/imgproc/test/test_houghlines.cpp
```diff
@@ -140,6 +140,29 @@ class ProbabilisticHoughLinesTest : public BaseHoughLineTest, public testing::Te
     }
 };
 
+typedef std::tr1::tuple<double, double, double, double> HoughLinesPointSetInput_t;
```
alalek: Don't use `std::tr1` here and below: https://github.com/opencv/opencv/wiki/Coding_Style_Guide#implementing-tests

take1014:
Thanks for the feedback. 
1.I solved the conflict.
2.I deleted "std :: tr1" and changed it to use "tuple"

modules/imgproc/test/test_houghlines.cpp
```diff
@@ -139,6 +139,29 @@ class ProbabilisticHoughLinesTest : public BaseHoughLineTest, public testing::Te
...
+protected:
+    void run_test();
+    double Rho;
+    double Theta;
+    double rhoMin, rhoMax, rhoStep;
+    double thetaMin, thetaMax, thetaStep;
+public:
+    HoughLinesPointSetTest()
+    {
+        rhoMin = std::tr1::get<0>(GetParam());
```
alalek: Here and **below**:
```.diff
-std::tr1::get
+get
```

take1014:
Thanks for the feedback.
I deleted std::tr1::get and changed it to use 'get'

savuor:
👍

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -2072,6 +2072,63 @@ CV_EXPORTS_W void HoughLinesP( InputArray image, OutputArray lines,
...
 
+/** @brief Finds lines in a set of points using the standard Hough transform.
+
+The function finds lines in a set of points using a modification of the Hough transform.
+
+Example: :
+@code
+    #include <opencv2/imgproc.hpp>
+    #include <opencv2/highgui.hpp>
+    #include <math.h>
```
alalek: Code in commends is not tested at all, and become dead/wrong very quickly. So we don't allow this.
Use snippets instead to check code compilation at least. Check for [this patch](https://github.com/opencv/opencv/pull/10512/files#diff-b6033f44ae355b1c40481eca8b593a89R200), snippets [location](https://github.com/opencv/opencv/tree/master/samples/cpp/tutorial_code/snippets).

take1014: I’m so sorry, I could not maintain the sample code at all.
I fixed sample code. Thank you for the feedback.

alalek: I mean that sample code should not be stored in comments. See this commit how to store function documentation samples: https://github.com/opencv/opencv/commit/11fa0094ffee2437ade42f99847d6c0485e0a360

take1014: I understood what you mean. I'm so sorry again. I work on it.
Can I add a new file for sample?

alalek: Yes, "imgproc_HoughLinesPointSet.cpp" should be fine.

take1014: Thank you for teaching me a lot.
1. Deleted sample code in comment.
2. Added snippets for HoughLinesPointSet.

alalek:
BTW, Function documentation preview is [here](http://pullrequest.opencv.org/buildbot/export/pr/10646/docs/dd/d1a/group__imgproc__feature.html#ga2858ef61b4e47d1919facac2152a160e).

PR #10660, opened 2018-01-22
Updating comments of imagproc_c.h file

### LINKED ISSUE #10652: Comments specifying about "cvWarpAffine"
In the OpenCV imagproc modules every functions description is present and at the last of their description a line is present which is something like **"@see "function-name"** but [here](https://github.com/opencv/opencv/blob/master/modules/imgproc/include/opencv2/imgproc/imgproc_c.h#L399)  function name following is different and name specifying through "@see" is different as here the line is **@see cvWarpAffine** but the function following is **cvGetQuadrangleSubPix**

victor-ludorum:
This PR fixes #10652 and updated the comments of [imagproc_c.h file](https://github.com/opencv/opencv/blob/master/modules/imgproc/include/opencv2/imgproc/imgproc_c.h#L399)

@alalek Would you mind to review it and suggest any necessary changes .


modules/imgproc/include/opencv2/imgproc/imgproc_c.h
```diff
@@ -396,7 +396,7 @@ CVAPI(void)  cvGetRectSubPix( const CvArr* src, CvArr* dst, CvPoint2D32f center
     matrixarr = ( a11  a12 | b1 )   dst(x,y) <- src(A[x y]' + b)
                 ( a21  a22 | b2 )   (bilinear interpolation is used to retrieve pixels
                                      with fractional coordinates)
-@see cvWarpAffine
+@see cv::getQuadrangleSubPix
```
terfendail: There is no `cv::getQuadrangleSubPix` function declared in OpenCV so it's a bit confusing to reference it in documentation

victor-ludorum: Thanks @terfendail . Actually I am referencing about this **"CVAPI(void)  cvGetQuadrangleSubPix"** function as in the module in the starting of every function there is declaration like **@see "function name"** which here is different as the function proceeding is **"CVAPI(void)  cvGetQuadrangleSubPix"** but here written is **cvWarpAffine** . Therefore , I have done it.  Please, suggest ..

PR #10661, opened 2018-01-22
core: fix kmeans multi-threading performance

alalek:
- Update perf tests (commit 1)
- Add nstripes (task granularity) parameter (commit 2)

```
allow_multiple_commits=1
```


alalek:
:+1:

PR #10663, opened 2018-01-22
hogsvm compatibility with python3

jmlich:
I was trying hogsvm tutorial with python3 and it wasn't working well. Here are necessary changes. 

I have added list() call to map() to be able run tutorial with python3.
```
-hogdata = [map(hog,row) for row in deskewed]
+    hogdata = [list(map(hog,row)) for row in deskewed]
```

I have also added if __main__ and corresponding whitespaces:
```
if __name__ == "__main__":
```

please use "git diff -w" to see changes without whitespaces.

### DIFF
--- a/samples/python/tutorial_code/ml/py_svm_opencv/hogsvm.py
+++ b/samples/python/tutorial_code/ml/py_svm_opencv/hogsvm.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 import cv2 as cv
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


alalek:
> if __name__ == "__main__"

This file is tutorial example, it is not expected to be imported anywhere as python module.

jmlich:
I thought the "main == __name__ " is good habit in python scripts. I am not going to modify pull request. Merge it if you like it. Reject it if you don't.

alalek:
Python3 fix works. Thank you :+1:
Unnecessary changes are reverted from this patch.

jmlich:
Thanks :thumbsup:

PR #10667, opened 2018-01-23
 calib3d: add stereoCalibrateExtended

paroj:
- allow CALIB_USE_EXTRINSIC_GUESS
- returns per view errors

TODO:
- [x] Test for new flag


catree:
How easy is it to return also the `rvecs` and `tvecs` for each image? See this [request](https://github.com/opencv/opencv/issues/7263#issuecomment-314168492).

paroj:
its certainly possible, but not a priority for me. Furthermore I dont think it makes sense as it makes the interface even clunkier (15 -> 17 params) and you can just use projectPoints to get the per view R, t.

alalek:
@paroj Thank you for the contribution!
Could you please add some test for this flag (there is no reason to test output accuracy, but at least that we don't have some assertion somewhere about unsupported flags and/or wrong sizes)?
Test case example:
```
... some valid non-singular inputs ...
stereoCalibrateExtended without CALIB_USE_EXTRINSIC_GUESS => res1
stereoCalibrateExtended with CALIB_USE_EXTRINSIC_GUESS and R/T from previous call => res2
... check something ...
```

PR #10669, opened 2018-01-23
Sitting DPI for TIFF files.

chrisestesaa01:
Writing tiff files now sets TIFFTAG_XRESOLUTION, TIFFTAG_YRESOLUTION, and TIFFTAG_RESOLUTIONUNIT. If these tags are not specified in params, they default to 300 and inches.


modules/imgcodecs/src/grfmt_tiff.cpp
```diff
@@ -804,6 +804,22 @@ bool  TiffEncoder::writeLibTiff( const Mat& img, const std::vector<int>& params)
         return false;
     }
 
+    //Get dpi.
+    int dpix = 300;
+    int dpiy = 300;
+    int resUnit = RESUNIT_INCH;
+    readParam(params, TIFFTAG_XRESOLUTION, dpix);
+    readParam(params, TIFFTAG_YRESOLUTION, dpiy);
+    readParam(params, TIFFTAG_RESOLUTIONUNIT, resUnit);
```
alalek: "params" are imwrite's values from this enum: https://github.com/opencv/opencv/blob/3.4.0/modules/imgcodecs/include/opencv2/imgcodecs.hpp#L81

Using values outside of mentioned enum just violates the current OpenCV `imwrite()` design (conflicts in enum values are possible so these hacks should not be allowed at all).

Also OpenCV public API should not rely on external "custom" headers, like "tiff.h".

Problem is here and for other "readParam()" calls. Related #361

vpisarev:
@chrisestesaa01, I agree with @alalek. Please, extend `ImwriteFlags` enumeration with IMWRITE_XRESOLUTION etc. and the convert them to the proper TIFF tags in the TIFF encoder.

modules/imgcodecs/src/grfmt_tiff.cpp
```diff
@@ -804,6 +804,36 @@ bool  TiffEncoder::writeLibTiff( const Mat& img, const std::vector<int>& params)
...
+    int dpiy = 300;
+
+    //Check for resolution inches.
+    int resUnit = RESUNIT_NONE;
+    readParam(params, IMWRITE_TIFF_RESUNIT_INCH, resUnit);
+
+    //Check for resolution centemeters.
+    if (resUnit == RESUNIT_NONE)
+    {
+        readParam(params, IMWRITE_TIFF_RESUNIT_CM, resUnit);
```
alalek: I don't see any test, so I'm ask here: how units selection works?
Both are read into the same variable and passed to backed without any changes.

modules/imgcodecs/src/grfmt_tiff.hpp
```diff
@@ -77,7 +77,10 @@ enum  TiffTag
     TIFF_TAG_SAMPLES_PER_PIXEL = 277,
     TIFF_TAG_ROWS_PER_STRIP = 278,
     TIFF_TAG_PLANAR_CONFIG = 284,
-    TIFF_TAG_COLOR_MAP = 320
+    TIFF_TAG_COLOR_MAP = 320,
+    TIFF_TAG_RESOLUTION_UNIT = 296,
+    TIFF_TAG_X_RESOLUTION = 282,
+    TIFF_TAG_Y_RESOLUTION = 283
```
alalek: These enums must go away completely (new and buggy old).
All these values are defined in "tiff.h". This header is private, so there is no sense to duplicate these unused values here.

saskatchewancatch:
I guess one more thing to keep in mind, when setting the x and y DPI values, because of the way `readParams` is written, we are limited to setting integer values. Further, when setting the TIFF tag for the x and y resolutions, it should be noted that the "TIFF type" for those tags is "RATIONAL", that is, it's two `long`s that specify a ratio. It would be a somewhat awkward undertaking to have `grfmt_tiff.cpp` be able to read in two integers for the x and y DPIs. I suggest just supporting integer DPIs for now (this is reasonable).

alalek:
Replaced by #11231

PR #10672, opened 2018-01-24
layer fusion patches for dnn module

pengli:
```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -1239,13 +1239,14 @@ struct Net::Impl
                     }
                 }
 
-                // For now,  OpenCL target only support fusion with activation of ReLU/ChannelsPReLU/Power
+                // For now, OpenCL target support fusion with activation of ReLU/ChannelsPReLU/Power/Tanh
                 if ( preferableTarget != DNN_TARGET_OPENCL ||
                         (preferableTarget == DNN_TARGET_OPENCL &&
                          nextData &&
-                        (!nextData->type.compare("ReLU") ||
-                         !nextData->type.compare("ChannelsPReLU") ||
-                         !nextData->type.compare("Power"))) )
+                        ((nextData->type == "ReLU") ||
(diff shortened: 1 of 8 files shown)


modules/dnn/include/opencv2/dnn/all_layers.hpp
```diff
@@ -473,6 +473,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         float epsilon;
 
         virtual void getScaleShift(Mat& scale, Mat& shift) const = 0;
+        virtual void getScaleShift(UMat& scale, UMat& shift) const = 0;
```
alalek: This call is not necessary. It is used during preparation stage, so there is no problem to use Mat instead.

Also
```
         umat_weight = weights_.getUMat(ACCESS_READ);
         umat_bias = bias_.getUMat(ACCESS_READ);
```

in batch_norm_layer.cpp should not be initialized unconditionally.

On systems with OpenCL without "zero-copy" support memory consumption would be too large (especially when OpenCL is not used).
It is better to initialize this due the first call of the "forward()" processing.

modules/dnn/src/dnn.cpp
```diff
@@ -1245,6 +1246,7 @@ struct Net::Impl
                          nextData &&
                         (!nextData->type.compare("ReLU") ||
                          !nextData->type.compare("ChannelsPReLU") ||
+                         !nextData->type.compare("TanH") ||
```
alalek: I believe this form `0 == nextData->type.compare("...")` looks more clear than negative `compare()` call

Please update the comment above too.

pengli: yes,  I changed to "==" operator to make it clear

modules/dnn/include/opencv2/dnn/all_layers.hpp
pengli: it makes sense to me.  code is updated accordingly

PR #10677, opened 2018-01-24
core(lapack): fix build issues related to 'extern "C"'

### LINKED ISSUE #10668: OpenBLAS support broken due to extern C linkage
- OpenCV => 3.4.0
- Compiler => GCC 5.4

##### Detailed description
Hi!

we build OpenCV with OpenBLAS 0.2.20 (also our own build) ...ah, up to patch #9955, LAPACK detection had been working properly.

Starting with the changes in patch #9955 in the 3.3.1 release (and 3.4.0 too, to which we're trying to upgrade to atm), the LAPACK cmake detection script outputs 

> -- LAPACK(OpenBLAS): LAPACK_LIBRARIES: /usr/local/lib/libopenblas.so
> -- LAPACK(OpenBLAS): Can't build LAPACK check code. This LAPACK version is not supported.

I was able to track it down to the extern "C" encapsulation of the OpenBLAS headers in the lapack proxy which seems similar to issue #10230. Now, if we're removing the `extern "C"` statement, everything works fine again. 

I'm just wondering if it this is due to some trouble with our own build version of OpenBLAS as #9955 was related to the OpenBLAS support in OpenCV, too. @alalek Why did you put the OpenBLAS headers in extern C linkage?

##### Workaround

Atm, I'd removed the extern C linkage from the detection script to build OpenCV with OpenBLAS...

Thanks in advance and regards

alalek:
resolves #10668

/cc @nightsparc

### DIFF
--- a/cmake/OpenCVFindLAPACK.cmake
+++ b/cmake/OpenCVFindLAPACK.cmake
@@ -31,7 +31,7 @@ macro(ocv_lapack_check)
   else()
     # adding proxy opencv_lapack.h header
     set(CBLAS_H_PROXY_PATH ${CMAKE_BINARY_DIR}/opencv_lapack.h)
-    if(APPLE)
+    if((APPLE OR OPENCV_SKIP_LAPACK_EXTERN_C) AND NOT OPENCV_FORCE_LAPACK_EXTERN_C)
         set(_lapack_include_str_extern_C "")
         set(_lapack_include_str_extern_C_end "")
     else()
--- a/cmake/checks/lapack_check.cpp
+++ b/cmake/checks/lapack_check.cpp
@@ -1,3 +1,4 @@
+#include <complex.h>
 #include "opencv_lapack.h"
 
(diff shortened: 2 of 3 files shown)


nightsparc:
Works, thx!

The `OPENCV_SKIP_LAPACK_EXTERN_C` and `OPENCV_FORCE_LAPACK_EXTERN_C` switches -
 how do you plan to set them? In the respective detection scripts for OpenBLAS, MKL, etc? Or should they be available for configuration in cmake?

alalek:
`OPENCV_SKIP_LAPACK_EXTERN_C` / `OPENCV_FORCE_LAPACK_EXTERN_C` are internal CMake parameters that can be passed additionally (via command-line or "Add entry" cmake-gui button) if something goes wrong (and `#include <complex.h>` is not enough).

nightsparc:
Ah ok.
Yes, I guess this will be enough then.

Thanks again!

PR #10678, opened 2018-01-24
Updated check of OutputArray dimensions to allow 1-dim recreation

### LINKED ISSUE #10552: calcHist can be called with const output parameter
- OpenCV: 3.3.0
- Operating System / Platform: Windows 10 Enterprise 64-Bit
- Compiler: Microsoft Visual Studio Professional 2015 (x86 Debug build)

##### Detailed description

cv::calcHist triggers a cv::Exception when provided a _const_ histogram:
```
(...)\opencv-3.3.0\modules\core\src\matrix.cpp:2515: error: (-215) CV_MAT_TYPE(mtype) == m.type() in function cv::debug_build_guard::_OutputArray::create
```
The issue can be reproduced with the C++ snippet below.

The code below should not build in the first place; it should raise a const correctness error at compile time.  
##### Steps to reproduce

    ```
        cv::Mat image(100, 100, CV_8U, 128);

        const int    channels[] = { 0 };
        const int    sizes[] = { 256 };
        const float  range[] = { 0.0f, 256.0f };
        const float* ranges[] = { range };
        const cv::Mat histogram; // Works if you remove 'const'

        cv::calcHist(&image, 1, channels, cv::noArray(), histogram, 1, sizes, ranges, true, false);
    ```

terfendail:
### This pullrequest changes
Updated check of OutputArray dimensions to allow 1-dim recreation
Closes #10552

### DIFF
--- a/modules/core/src/matrix_wrap.cpp
+++ b/modules/core/src/matrix_wrap.cpp
@@ -1277,6 +1277,14 @@ void _OutputArray::create(int _rows, int _cols, int mtype, int i, bool allowTran
 void _OutputArray::create(int d, const int* sizes, int mtype, int i,
                           bool allowTransposed, int fixedDepthMask) const
 {
+    int sizebuf[2];
+    if(d == 1)
+    {
+        d = 2;
+        sizebuf[0] = sizes[0];
+        sizebuf[1] = 1;
+        sizes = sizebuf;
+    }
     int k = kind();
     mtype = CV_MAT_TYPE(mtype);
 


alalek:
`cv::Mat` doesn't support 1D arrays (they are forced to 2D). Until this fix (or future change) using `OutputArray::create()` with dim=1 is very questionable and should be avoided.

vpisarev:
@terfendail, can you "patch" the input size once instead of fixing it in several places?

<pre>
int sizebuf[2];
if(d == 1)
{
     d = 2;
     sizebuf[0] = sizes[0];
     sizebuf[1] = 1;
     sizes = sizebuf;
}
</pre>

terfendail:
In #10552 matrix is created with zero size so it would be anyway impossible to write data to it. This patch allow to pass const matrix with proper size

alalek:
Should issue be closed anyway?

terfendail:
I think the issue could be closed. Since it's now possible to pass to calcHist const matrix of proper size and it is anyway impossible to raise type/size incompatibility error at compilation type due to specific of matrix implementation

alalek:
@vpisarev ping

vpisarev:
:+1:

PR #10685, opened 2018-01-25
Use Intel IPP for 8U/16U in cvtBGRtoGray

woodychow:
There are 8U and 16U in ippiColor2Gray/ippiRGB2Gray.

If there's a reason why IPP is limited to float32 in this function, could anyone please let me know?

### DIFF
--- a/modules/imgproc/src/color.cpp
+++ b/modules/imgproc/src/color.cpp
@@ -10140,29 +10140,31 @@ void cvtBGRtoGray(const uchar * src_data, size_t src_step,
 #if defined(HAVE_IPP) && IPP_VERSION_X100 >= 700
     CV_IPP_CHECK()
     {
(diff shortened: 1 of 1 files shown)


vpisarev:
@woodychow, thanks for the patch and for the question. I think, by looking at the results from our buildbot you now see the reason why we do not use IPP in those functions. Many pipelines that include color-to-gray conversion as the preprocessing step are very sensitive to the accuracy of this function. In fact, they demand bit-exact results. If you have +/-1 difference in the grayscale output that may lead to quite severe differences downstream the pipeline. So let me close this patch until IPP is updated to match OpenCV results 1:1

woodychow:
Yes I can now see why. Well, I am not sure if Intel will change their implementation to match the output of OpenCV but OK. Thanks!

PR #10688, opened 2018-01-25
solves issue #10687

### LINKED ISSUE #10687: calcbackproject exception in 3d
opencv 3.4-dev
VS 2017
Windows 10

Hi,

problem is relative to this [post ](http://answers.opencv.org/question/182814/backprojection-not-working-when-tried-with-bgr-instead-of-hsv/)in python but i can reproduce error in C++

Exception is : 
`OpenCV Error: Assertion failed (p[-1] <= 2) in cv::MatSize::operator (), file G:\Lib\opencv\modules\core\include\opencv2/core/mat.inl.hpp, line 1407
`

Step to reproduce in python

```
import numpy as np
import cv2 as cv
img = cv.imread('c:/lib/opencv/samples/data/fruits.jpg',cv.IMREAD_COLOR)
orange = img[300:380,150:220]
cv.imshow('fruits',img)
cv.imshow('orange',orange)
# calculating object histogram
c=[0,1,2]
r=[32,32,32]
rc=[  0, 256,0,256,0,256]
roihist = cv.calcHist([orange],c, None, r, rc )
# normalize histogram and apply backprojection
cv.normalize(roihist,roihist,0,255,cv.NORM_MINMAX)
dst = cv.calcBackProject([img],c,roihist,rc,1)
ret,thresh = cv.threshold(dst,10,255,0)
thresh = cv.merge((thresh,thresh,thresh))
cv.imshow('dst',dst)
cv.imshow('res',thresh)
cv.waitKey(0)
cv.destroyAllWindows()

```
and in C++

```
#include <opencv2/opencv.hpp> 
#include <iostream>

using namespace cv;
using namespace std;

{
        Mat img = imread("g:/lib/opencv/samples/data/fruits.jpg", IMREAD_COLOR);
    Mat orange = img(Rect(150,300, 70,80));
    imshow("fruits", img);
    imshow("orange", orange);
    int hBins = 32;
    int histSize[] = { hBins, hBins, hBins };
    int nbChannels = 3;
    int channels[] = { 0,1,2 };
    float h_range[] = { 0, 256 };
    const float* ranges[] = { h_range, h_range,h_range };
    cv::Mat histogramme;

    cv::calcHist(&orange, 1, channels, Mat(), histogramme, nbChannels, histSize, ranges, true, false);
    normalize(histogramme, histogramme, 0, 255, cv::NormTypes::NORM_MINMAX, -1, cv::Mat());
    cv::Mat backproj;
    Mat bk;
    cv::calcBackProject(&img, 1, channels, histogramme, bk, ranges, 1, true);
    imshow("result", bk);
    waitKey(0); // NO Problem it is good in 3d
    vector<Mat> x;
    vector<int> ch;
    vector<float> cranges;
    x.push_back(img);
    ch.push_back(0);
    ch.push_back(1);
    ch.push_back(2);
    cranges.push_back(0); cranges.push_back(256);
    cranges.push_back(0); cranges.push_back(256);
    cranges.push_back(0); cranges.push_back(256);
    calcBackProject(x, ch, histogramme, bk, cranges, 1); // Exception in 3d
    imshow("result", bk);
    waitKey(0);
   }

```

LaurentBerger:
resolves #10687

### DIFF
--- a/modules/imgproc/src/histogram.cpp
+++ b/modules/imgproc/src/histogram.cpp
@@ -2449,17 +2449,18 @@ void cv::calcBackProject( InputArrayOfArrays images, const std::vector<int>& cha
                           double scale )
 {
     CV_INSTRUMENT_REGION()
-
+    if (hist.dims() <= 2)
+    {
 #ifdef HAVE_OPENCL
-    Size histSize = hist.size();
-    bool _1D = histSize.height == 1 || histSize.width == 1;
-    size_t histdims = _1D ? 1 : hist.dims();
+        Size histSize = hist.size();
+        bool _1D = histSize.height == 1 || histSize.width == 1;
+        size_t histdims = _1D ? 1 : hist.dims();
 #endif
 
-    CV_OCL_RUN(dst.isUMat() && hist.type() == CV_32FC1 &&
-               histdims <= 2 && ranges.size() == histdims * 2 && histdims == channels.size(),
-               ocl_calcBackProject(images, channels, hist, dst, ranges, (float)scale, histdims))
-
+        CV_OCL_RUN(dst.isUMat() && hist.type() == CV_32FC1 &&
+            histdims <= 2 && ranges.size() == histdims * 2 && histdims == channels.size(),
+            ocl_calcBackProject(images, channels, hist, dst, ranges, (float)scale, histdims))
+    }
     Mat H0 = hist.getMat(), H;
     int hcn = H0.channels();
 


modules/imgproc/src/histogram.cpp
```diff
@@ -2449,17 +2449,18 @@ void cv::calcBackProject( InputArrayOfArrays images, const std::vector<int>& cha
                           double scale )
 {
     CV_INSTRUMENT_REGION()
-
+    if (hist.dims() != 3)
```
savuor: It's better to compare like that: `if(hist.dims() <= 2)` to avoid histograms of higher dimensions.

savuor:
:+1:

alalek:
BTW, New functionality should be covered by tests.

LaurentBerger:
@alalek why do you mean?  [calcBackProject ](https://github.com/opencv/opencv/blob/master/modules/imgproc/src/histogram.cpp#L2446)is not tested? or 3d calcbackproject?

alalek:
I mean that 3D case is not tested. We need some test to avoid future regressions.

PR #10689, opened 2018-01-25
Do not build protobuf without dnn

mshabunin:
### This pullrequest changes

- do not build _protobuf_ when _dnn_ module is disabled (via BUILD_LIST option)
- add BUILD_LIST to cmake cache so that it becomes available in cmake-gui


dkurt:
:+1:

modules/dnn/CMakeLists.txt
```diff
@@ -1,21 +1,17 @@
+set(the_description "Deep neural network module. It allows to load models from different frameworks and to make forward pass")
+ocv_add_module(dnn opencv_core opencv_imgproc WRAP python matlab java js)
+
 if(WINRT)
   ocv_module_disable(dnn)
```
alalek: It is too late to disable module after "ocv_add_module()".
The same problem is below.

dkurt: @alalek, for a second case we can add
```cmake
if(DEFINED BUILD_PROTOBUF AND NOT BUILD_PROTOBUF)
  include(${OpenCV_SOURCE_DIR}/cmake/OpenCVFindLibProtobuf.cmake)
  if(NOT Protobuf_FOUND)
    ocv_module_disable(opencv_dnn)
  endif()
endif()
```
before `ocv_add_module`. Because if `BUILD_PROTOBUF` is `false` we'll try to link to system protobuf. If `BUILD_PROTOBUF` is `true` we can build it from 3rdparty anyway.

mshabunin: What about `BUILD_${the_module}_INIT`?

alalek: I believe `EXCLUDE_FROM_ALL` for `add_library()` + install(OPTIONAL) should help with protobuf,

modules/dnn/CMakeLists.txt
```diff
@@ -2,26 +2,21 @@ if(WINRT)
...
+if(NOT HAVE_PROTOBUF)
   ocv_module_disable(opencv_dnn)
 endif()
 
 set(the_description "Deep neural network module. It allows to load models from different frameworks and to make forward pass")
+ocv_add_module(dnn opencv_core opencv_imgproc WRAP python matlab java js)
 
 ocv_add_dispatched_file("layers/layers_common" AVX AVX2 AVX512_SKX)
 
-ocv_add_module(dnn opencv_core opencv_imgproc WRAP python matlab java js)
```
alalek: Don't reorder `ocv_add_dispatched_file()` and `ocv_add_module()` (and especially `ocv_define_module()`)

modules/dnn/CMakeLists.txt
```diff
@@ -2,26 +2,21 @@ if(WINRT)
...
 ocv_warnings_disable(CMAKE_CXX_FLAGS -Wno-shadow -Wno-parentheses -Wmaybe-uninitialized -Wsign-promo
                                      -Wmissing-declarations -Wmissing-prototypes
 )
 ocv_warnings_disable(CMAKE_CXX_FLAGS /wd4701 /wd4100)
 
-include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src/ocl4dnn/include ${OPENCL_INCLUDE_DIRS})
+include_directories(src/ocl4dnn/include ${OPENCL_INCLUDE_DIRS})
```
alalek: Using absolute paths resolves some problems with opencv_world builds.

modules/dnn/CMakeLists.txt
```diff
@@ -59,26 +55,17 @@ ocv_warnings_disable(CMAKE_CXX_FLAGS
...
+  file(GLOB fw_srcs "misc/tensorflow/*.cc" "misc/caffe/opencv-caffe.pb.cc")
+  file(GLOB fw_hdrs "misc/tensorflow/*.h" "misc/caffe/opencv-caffe.pb.h")
 endif()
 
-ocv_source_group("Src\\protobuf" FILES ${Protobuf_SRCS} ${Protobuf_HDRS})
-ocv_module_include_directories(include ${Protobuf_INCLUDE_DIRS})
-
-ocv_glob_module_sources(${Protobuf_SRCS} ${Protobuf_HDRS} ${CBLAS_H_PROXY_PATH})
-ocv_create_module(${Protobuf_LIBRARIES} ${LAPACK_LIBRARIES})
+ocv_module_include_directories(${CMAKE_CURRENT_SOURCE_DIR}/misc/caffe ${CMAKE_CURRENT_SOURCE_DIR}/misc/tensorflow)
```
alalek: Including of folders with pre-generated conflict files should be avoided.

PR #10690, opened 2018-01-25
Fixed issue in ORB detection if firstLevel property is set above 0

### LINKED ISSUE #10561: `ORB::setFirstLevel(i>0)` causes a crash
- OpenCV => 3.4.0-dev
- Compiler => ALL

##### Detailed description
Cannot set the first level in the ORB detector to any value other than zero.
Setting it to any positive number `i` is supposed to make all octaves below `i` enlarged, rather than 
the usual shrinkage (similar to SIFT negative octaves).

##### Steps to reproduce
```.c++
vector<KeyPoint> kps;
Mat img = imread("any_image.png"), descs;
Ptr<ORB> orb = ORB::create();

orb->setFirstLevel(1); // <- Crash ahead!
orb->detectAndCompute(img, noArray(), kps, descs);
```

##### Actual result
CRASH!

terfendail:
resolves #10561

### This pullrequest changes
Fixed pyramid buffer size to allow retaining of upscaled images. Updated pyramid evaluation to perform image upscaling if necessary.

### DIFF
--- a/modules/features2d/include/opencv2/features2d.hpp
+++ b/modules/features2d/include/opencv2/features2d.hpp
@@ -304,10 +304,11 @@ public:
     will mean that to cover certain scale range you will need more pyramid levels and so the speed
     will suffer.
     @param nlevels The number of pyramid levels. The smallest level will have linear size equal to
-    input_image_linear_size/pow(scaleFactor, nlevels).
+    input_image_linear_size/pow(scaleFactor, nlevels - firstLevel).
     @param edgeThreshold This is size of the border where the features are not detected. It should
     roughly match the patchSize parameter.
-    @param firstLevel It should be 0 in the current implementation.
+    @param firstLevel The level of pytramid to put source image to. Previous layers are filled
+    with upscaled source image.
     @param WTA_K The number of points that produce each element of the oriented BRIEF descriptor. The
     default value 2 means the BRIEF where we take a random point pair and compare their brightnesses,
     so we get 0/1 response. Other possible values are 3 and 4. For example, 3 means that we take 3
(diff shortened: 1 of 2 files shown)


modules/features2d/include/opencv2/features2d.hpp
```diff
@@ -304,10 +304,11 @@ class CV_EXPORTS_W ORB : public Feature2D
     will mean that to cover certain scale range you will need more pyramid levels and so the speed
     will suffer.
     @param nlevels The number of pyramid levels. The smallest level will have linear size equal to
-    input_image_linear_size/pow(scaleFactor, nlevels).
+    input_image_linear_size/pow(scaleFactor, nlevels - firstLevel).
     @param edgeThreshold This is size of the border where the features are not detected. It should
     roughly match the patchSize parameter.
-    @param firstLevel It should be 0 in the current implementation.
+    @param firstLevel The level of pytramid to put source image to. Previous layers are filled
+    with upscaled source image.
```
alalek: please fix typo here

modules/features2d/src/orb.cpp
```diff
@@ -673,7 +673,7 @@ class ORB_Impl : public ORB
     void setEdgeThreshold(int edgeThreshold_) { edgeThreshold = edgeThreshold_; }
     int getEdgeThreshold() const { return edgeThreshold; }
 
-    void setFirstLevel(int firstLevel_) { firstLevel = firstLevel_; }
+    void setFirstLevel(int firstLevel_) { CV_Assert(firstLevel >= 0);  firstLevel = firstLevel_; }
```
alalek: Parameter name contains underscore: `firstLevel_`, assertion - doesn't.

vpisarev:
:+1:

modules/features2d/src/orb.cpp
alalek: @vpisarev This definitely should be fixed.

PR #10695, opened 2018-01-26
Add mask support to SimpleBlobDetector

mworchel:
### This pullrequest changes

Implementing mask handling in the SimpleBlobDetector in a similar way as AKAZE, KAZE, BRISK and AGAST features in the master branch. This fixes a (probably forgotten) TODO.

### DIFF
--- a/modules/features2d/src/blobdetector.cpp
+++ b/modules/features2d/src/blobdetector.cpp
@@ -272,9 +272,8 @@ void SimpleBlobDetector::findBlobs(const cv::Mat &image, const cv::Mat &binaryIm
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


mworchel:
@alalek Sorry to link you directly, but I saw you commenting on another issue that in some cases automatic builds are not triggered. I submitted this pull request to the 2.4 branch since I would consider it a small fix, but it seems like the required builds are not being triggered (and not even being queued looking at the build bot page).

Is there anything that can be done about it?

alalek:
Patch is fine (except missing tests). OpenCL builders are disabled for 2.4 branch.

Thank you!
:+1:

PR #10697, opened 2018-01-26
Use Intel TBB's task arena if possible

woodychow:
such that OpenCV's num threads setting only affects OpenCV.

```
buildworker:Custom=linux-3
```

### DIFF
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -328,7 +328,11 @@ namespace
 static int numThreads = -1;
 
 #if defined HAVE_TBB
-static tbb::task_scheduler_init tbbScheduler(tbb::task_scheduler_init::deferred);
+    #if TBB_INTERFACE_VERSION >= 8000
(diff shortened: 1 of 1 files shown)


modules/core/src/parallel.cpp
```diff
@@ -494,9 +519,17 @@ int cv::getNumThreads(void)
 
 #if defined HAVE_TBB
 
+#if TBB_INTERFACE_VERSION >= 9100
+    return tbbArena.max_concurrency();
+#elif TBB_INTERFACE_VERSION >= 8000
+    return numThreads > 0
+        ? numThreads
+        : tbb::task_scheduler_init::default_num_threads();
```
woodychow: Not sure what to do here as max_concurrency is only added in TBB2017 (which is 9100 I believe)
https://software.intel.com/en-us/articles/intel-threading-building-blocks-release-notes#2017

alalek:
@woodychow I replaced additional class to additional method. Please take a look.

woodychow:
@alalek it looks good. Thanks very much!

PR #10698, opened 2018-01-26
GaussianBlur performance test extension

terfendail:
**WIP**
### This pullrequest changes
Added GaussianBlur performance test for generic kernel case

### DIFF
--- a/modules/imgproc/perf/perf_blur.cpp
+++ b/modules/imgproc/perf/perf_blur.cpp
@@ -43,6 +43,9 @@ typedef perf::TestBaseWithParam<Size_MatType_BorderType3x3_t> Size_MatType_Borde
 typedef std::tr1::tuple<Size, MatType, BorderType> Size_MatType_BorderType_t;
 typedef perf::TestBaseWithParam<Size_MatType_BorderType_t> Size_MatType_BorderType;
 
+typedef std::tr1::tuple<Size, MatType, Size, BorderType> Size_MatType_KSize_BorderType_t;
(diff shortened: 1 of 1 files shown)


modules/imgproc/perf/perf_blur.cpp
```diff
@@ -230,3 +233,27 @@ PERF_TEST_P(Size_MatType_BorderType, blur5x5,
 
     SANITY_CHECK(dst, 1);
 }
+
+PERF_TEST_P(Size_MatType_KSize_BorderType, gaussianBlurMxN,
+    testing::Combine(
+        testing::Values(szODD, szQVGA, szVGA, sz720p),
+        testing::Values(CV_8UC1, CV_8UC4, CV_16UC1, CV_16SC1, CV_32FC1),
+        testing::Values(Size(3, 3), Size(5, 5), Size(7, 7), Size(11, 11)),
+        BorderType::all()
```
alalek: +320 tests.

Perhaps we don't want this really (we don't want garbage in performance reports).

- szODD = 127x61 . It is performance test, not an accuracy. Same note for 320x240.
- FullHD size is missing
- CV_8UC4 => CV_8UC3
- CV_16UC1, CV_16SC1 - FullHD or 720p test is enough, reduce size to single one
- Kernel size 11x11 - FullHD or 720p only, other for all sizes
- BorderType::all() - FullHD or 720p only, REPLICATE for all sizes

vpisarev:
yes, it looks like too much tests. let's close it. In general, it would be nice to bring the number of test cases down, especially the regression tests

Issue #10601, opened 2018-01-15
DNN module readNetFromTensorflow() - How to obtain text graph definition for retrained network?


HenrikPetersen:
- OpenCV => 3.4
- Operating System / Platform => Mac OS X
- Compiler => Python

I am trying to load a tensorflow MobileNet SSD model that was retrained using the Tensorflow Object Detection API. However, I am not able to find any documentation on how to create the text graph defintion that is supposed to be given to readNetFromTensorflow  together with the frozen graph weights. 

The example given [here](https://github.com/opencv/opencv/blob/master/samples/dnn/mobilenet_ssd_python.py) works with the text graph definition given [here](https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/ssd_mobilenet_v1_coco.pbtxt), but not with my retrained model (I changed the number of classes to number of retrained classes). The output is a bunch of random bboxes on top of each other.

I tried to use older config files for the training, but that gives the same result.  I searched the web for methods to convert .pb files to .pbtxt files, but none of them give the required  structure.

I also tried to only supply readNetFromTensorflow with  the frozen graph weights, but that resulted in an error.

I would be nice to have some documentation on how to produce the text graph definitions, or at least provided a link to the config file that was used to create the [one](https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/ssd_mobilenet_v1_coco.pbtxt) provided in the example. 

Thanks in advance,
Henrik

dkurt:
@HenrikPetersen, May I ask you to test PR https://github.com/opencv/opencv/pull/10676 using this graph definition: https://gist.github.com/dkurt/fb324b48d1c6230febd1718ee52f7291 . Change line `attr { key: "num_classes" value { i: 2 } }` appropriately (number of classes + 1 for background).

xbcReally:
I want to know the answer too...And I note that you @dkurt commit a new pbtxt file for ssd-mobilenet,adn I tried the new one but result in showing lots of bboxes with confidence 0.99...

dkurt:
@xbcReally, please use http://answers.opencv.org. Specify a config file used for training (see https://github.com/tensorflow/models/tree/master/research/object_detection/samples/configs). Or follow recommendations from issues related at https://github.com/opencv/opencv/pull/10676.

Issue #10431, opened 2017-12-27
Unknown layer type Cast in op ToFloat in function populateNet(Tensorflow detection api ssd+mobilenet)


kdouc:
I'm training a object detection model use tensorflow detection api. Now i want to use it in opencv 3.4.0(3.3.1).
I have test the https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/detection_model_zoo.md and use https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/ssd_mobilenet_v1_coco.pbtxt 
this well work well. But in my model, i have changed the size and others,my config is:https://github.com/kdouc/example/blob/master/pipeline.config
So,when i run my model,got 👍 
ToFloat:Cast(image_tensor)
DstT:[ ]
SrcT:[ ]
the same problem like https://github.com/opencv/opencv/issues/9830,
and i have try to use ssd_mobilent_v1_coco.pbtxt which from frozen.pb,but the same problems.
So is there any way to got a "pbtxt" like https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/ssd_mobilenet_v1_coco.pbtxt . Which don't include ToFloat node?
waiting for your help!

dkurt:
@kdouc, please share a model. It's hard to solve without ways to reproduce.

kdouc:
@dkurt Thanks,this is my model and files
https://www.dropbox.com/s/86h7vqgrx6arhx6/Dropbox%20%E4%BD%BF%E7%94%A8%E5%85%A5%E9%97%A8.pdf?dl=0

dkurt:
@kdouc, there is no problem for me:
```python
import cv2 as cv
net = cv.dnn.readNetFromTensorflow('frozen_inference_graph.pb',
                                   'ssd_mobilenet_v1_coco.pbtxt')
```
Model is imported fine. One and only issue is an updated aspect ratios. We need to match them to Caffe-like `PriorBox` layer attributes. I guess it's better to use http://answers.opencv.org for it and close this issue.

dkurt:
@kdouc, May I ask you to test your model with PR https://github.com/opencv/opencv/pull/10676?

Please use the following text graph for it: [ssd_mobilenet_v1_issue_10431.txt](https://github.com/opencv/opencv/files/1659830/ssd_mobilenet_v1_issue_10431.txt)

Proposals' widths and heights received by the following script:
```python
from math import sqrt

min_scale = 0.08
max_scale = 0.95
[...]

Issue #9977, opened 2017-10-31
hello!I want to know how to generate the ssd_mobilenet_v1_coco.pbtxt


barbecacov:
I have trained the ssd_mobilenet model for my own data,and got the frozen_inference_graph.pb,but when i want to load the model in Opencv(3.3.1),the ssd_mobilenet_v1_coco.pbtxt is needed.I dont know how to generate my own ssd_mobilenet_v1_coco.pbtxt.Thank you very much

barbecacov:
the official .pbtxt https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/ssd_mobilenet_v1_coco.pbtxt.But it didnt work in my own model for ssd_mobilenet_v1

dkurt:
@barbecacov, [TensorFlow Object Detection API](https://github.com/tensorflow/models/tree/master/research/object_detection) uses config files like [ssd_mobilenet_v1_coco.config](https://github.com/tensorflow/models/blob/master/research/object_detection/samples/configs/ssd_mobilenet_v1_coco.config) to define the model. What changes did you apply for your model?

dkurt:
@barbecacov, Have you tried to replace https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/ssd_mobilenet_v1_coco.pbtxt#L2222 to desired number? Note that it's a number of classes + 1. So single class detection means `attr { key: "num_classes" value { i: 2 } }` (background, class).

barbecacov:
@dkurt I just changed the num_classes to 1,and some files' path

barbecacov:
@dkurt Yes,I know it .But it cant work im my own model

barbecacov:
@dkurt I have tried attr { key: "num_classes" value { i: 2 } }.But another problem ocurred,Assertion failed (int(numPriors * _numClasses) == inputs[1][1]) in cv::dnn::DetectionOutputLayerImpl.

juventi:
i have the same question...

lhwcv:
@barbecacov  Have you solved this problem,  I have the similar problem with you. 
With  2 classes, I tried { key: "num_classes" value { i: 3} }   and using  mobilenet_ssd_python.py  in opencv/samples/dnn , then it not crashed, but the output very different with  tensorflow

mahsaayi:
[...]

PR #10701, opened 2018-01-26
Fix padding for average pooling from TensorFlow

dkurt:
### This pullrequest changes

* Test for Inception v2 SSD from https://github.com/tensorflow/models/tree/master/research/object_detection
* Fix average pooling normalization for layers from TensorFlow. In case of AVE pooling with padding Caffe and Torch averages by kernel area so zero padding is used. However TensorFlow divides on number of real elements.

Close issues with retrained MobileNet-SSD and Inception v2 SSD models from TensorFlow because now it's usage question and each model is tested. Wiki: https://github.com/opencv/opencv/wiki/TensorFlow-Object-Detection-API .

resolves https://github.com/opencv/opencv/issues/9977
resolves https://github.com/opencv/opencv/issues/10431
resolves https://github.com/opencv/opencv/issues/10601

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/433

### DIFF
--- a/modules/dnn/src/layers/pooling_layer.cpp
+++ b/modules/dnn/src/layers/pooling_layer.cpp
@@ -228,6 +228,7 @@ public:
         const Mat* src, *rois;
         Mat *dst, *mask;
         Size kernel, stride, pad;
+        String padMode;
         int nstripes;
         bool computeMaxIdx;
         std::vector<int> ofsbuf;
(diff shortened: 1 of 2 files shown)


modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -504,10 +505,11 @@ class PoolingLayerImpl : public PoolingLayer
                     {
                         int xstart = x0 * stride_w - pad_w;
                         int xend = min(xstart + kernel_w, inp_width + pad_w);
-                        int xdelta = xend - xstart;
                         xstart = max(xstart, 0);
                         xend = min(xend, inp_width);
-                        float inv_kernel_area = 1.f/(ydelta*xdelta);
+                        float inv_kernel_area =
+                            padMode == "SAME" ? (yend - ystart) * (xend - xstart) : kernel.area();
```
alalek: > padMode == "SAME"

Please avoid strings check inside of for loops.
Please declare `bool isSamePad` variable outsize of these loops.

dkurt: Thanks! Fixed.

PR #10707, opened 2018-01-27
perf reports: remove units from table cells

alalek:
- moved to table title
- can be restored via '--show_units' option
- fix microseconds: mks -> us


vpisarev:
:+1:

PR #10708, opened 2018-01-27
core: Remove raw intrinsics from arithmetic operations

seiko2plus:
#### Tasks
- [x]  Prepare universal intrinsics to expand (#11022)
- [ ]  Add universal intrinsics for AVX2 
- [ ]  Expand hal_intrin tests to support SIMD 256 and 512
- [ ]  Replace raw intrinsics with universal SIMD for arithmetic operations
- [ ]  Update docs
```
force_builders=Custom
```

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin.hpp
@@ -342,6 +342,39 @@ CV_CPU_OPTIMIZATION_HAL_NAMESPACE_END
 #define CV_SIMD128_64F 0
 #endif
 
+#ifndef CV_SIMD256
+#define CV_SIMD256 0
+#endif
+
+#ifndef CV_SIMD256_64F
+#define CV_SIMD256_64F 0
+#endif
+
+#ifndef CV_SIMD512
+#define CV_SIMD512 0
+#endif
+
+#ifndef CV_SIMD512_64F
+#define CV_SIMD512_64F 0
+#endif
+
+#if CV_SIMD512
+    #define CV_SIMD 1
+    #define CV_SIMD_64F CV_SIMD512_64F
+    #define CV_SIMD_WIDTH 64
+    #define OPENCV_HAL_SIMD_CALL_SFX(fnc, ...) v512_##fnc(__VA_ARGS__)
+#elif CV_SIMD256
+    #define CV_SIMD 1
+    #define CV_SIMD_64F CV_SIMD256_64F
+    #define CV_SIMD_WIDTH 32
+    #define OPENCV_HAL_SIMD_CALL_SFX(fnc, ...) v256_##fnc(__VA_ARGS__)
+#else
+    #define CV_SIMD CV_SIMD128
+    #define CV_SIMD_64F CV_SIMD128_64F
+    #define CV_SIMD_WIDTH 16
+    #define OPENCV_HAL_SIMD_CALL_SFX(fnc, ...) v_##fnc(__VA_ARGS__)
+#endif
+
 //! @}
 
 //==================================================================================================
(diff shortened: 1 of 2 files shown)


modules/core/include/opencv2/core/hal/intrin.hpp
```diff
@@ -329,11 +329,17 @@ CV_CPU_OPTIMIZATION_HAL_NAMESPACE_END
...
+#include "opencv2/core/hal/intrin_avx2.hpp"
+
+#endif
+
 //! @addtogroup core_hal_intrin
 //! @{
 
 #ifndef CV_SIMD128
-//! Set to 1 if current compiler supports vector extensions (NEON or SSE is enabled)
+//! Set to 1 if current compiler supports vector extensions (NEON or SSE or VSX is enabled)
```
xsacha: AVX?

seiko2plus: Just update doc for [Power VSX](https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/hal/intrin_vsx.hpp). Please consider this branch isn't ready for review yet, I made so many changes and still working on other things. I would expect to finish it in a week or two.

kinchungwong:
Can someone familiar with OpenCV - OpenVX integration comment on whether the prefix ```vx_```, such as ```vx_uint8```, can be used within the Wide Universal Intrinsics proposal namespace (```cv```, or ```cv::hal```, or ```cv::hal_baseline```), and does not affect OpenVX, which puts these typedefs under global namespace?

They are in different namespaces, but there are source code in OpenCV which will be exposed to both namespaces if ```HAVE_OPENVX``` is defined.

A quick check might be just compile with ```HAVE_OPENVX``` defined. 
*(I might be wrong since I'm a noob here.)*

The README on https://github.com/opencv/opencv/tree/master/3rdparty/openvx gives a basic usage of OpenVX, by including "ivx.hpp", which pulls in "VX/vx.h", "VX/vx_types.h" (both of which are in OpenVX proper, https://github.com/GPUOpen-ProfessionalCompute-Libraries/amdovx-core/blob/develop/openvx/include 

*(Sorry I'm a noob here, please pardon if this question doesn't make sense.)*

seiko2plus:
@kinchungwong, Aww My bad, I think you're right and I may have rushed into it.
Could you propose a different prefix? Can we just replace it with "xv_" or "v_x"?

i.e.
````
xv_uint8 vec1 = xv_load(ptr);
xv_uint8 vec2 = xv_load_halves(ptr1, ptr2);
Data<R> trait = XV_RegTrait<LaneType>::zero();
````
````
v_xuint8 vec1  = v_xload(ptr);
v_xuint8 vec2  = v_xload_halves(ptr1, ptr2);
Data<R> trait  = V_XRegTrait<LaneType>::zero();
````

Thank you !

**Comment by kinchungwong** [comment]
@seiko2plus

> "Simply to ***```v_uint8```*** " looks better to me , but we still needs to add suffix to loaders since, v_load_* taken by simd128.

I think @vpisarev already knew; which is why he suggested the simplest "v underscore" ***prefix*** in [OE-27](https://github.com/opencv/opencv/wiki/OE-27.-Wide-Universal-Intrinsics):

> It's suggested to rename all the vector types used in universal intrinsics to something size-agnostic. Currently we use v_uint8x16 etc. types for the vector types. They can be renamed to v_uint8xn (or simply to v_uint8).

I would prefer keeping the ```xn``` ***suffix***, to distinguish a vector (of any type) from a scalar, for mental reasons.

---

I would also look for ***another suffix*** (possibly a union or struct), to denote "array of underlying vector type", such as ```__m128i[4u]```, which can be used to emulate a 512-bit vector on SSE2.

---

> "Simply to v_uint8" looks better to me , but we still needs to add suffix to loaders since, ***```v_load_*```*** taken by simd128.

As for ```v_load_*```, I can mention a C++ programming trick. The idea is to create a dummy class that wraps the arguments (an untyped pointer), then add lots of overloaded conversion operators on it. By doing this, one can call a general function, and then call a specific overload by specifying the return type. This works around the issue of C++ typically not letting you choose an overload on return type. I don't know if it is good or bad taste according to @vpisarev , so I will wait for his comments first.

Sample code: https://github.com/kinchungwong/simd_style_helper/blob/master/simd_style_helper_loadstore.h#L168 

Sample code: (in case my repo is deleted)
```
class DummyLoader
{
private:
   const void* m_ptr;

public:
  DummyLoader(const void* ptr) : m_ptr(ptr) {}

  /* implicit */  operator v_elephant () const 
  { 
#if HAVE_BOAT
    return load_elephant_with_boat(m_ptr);
#elif HAVE_TRUCK
    return load_elephant_with_truck(m_ptr);
#endif
  }

  /* implicit */  operator v_bigcat () const 
  { 
#if HAVE_CAGE
    return load_bigcat_with_cage(m_ptr);
#else
   static_assert(false, "Bigcat not supported on this platform without CAGE enabled");
#endif
  }
};

template <typename ANIMAL>
ANIMAL load(const void* ptr)
{
   // calls implicit conversion operator
   // important reminder: this is NOT reinterpret-cast.
    return (ANIMAL)DummyLoader(ptr); 
}
```

---

Could someone from OpenCV, /cc @alalek @vpisarev @mshabunin , help coordinate the effort? E.g. have a central list of tasks and the name of people (*currently*) assigned, and some basic (*temporary*) design docs? A "touch base" time slot will also help.

---

Can CI catch this issue earlier? Is there at least one platform where ```HAVE_OPENVX``` is enabled (e.g. one with an AMD GPU; *this is not an advertisement*), and some OpenVX tests can be run on that build?

---

There's also need to coordinate the list of universal intrinsics (in particular the ```typedef```s, macros, etc), see PR #11169 , where /cc @tomoaki0705  has a question about ```v_load_deinterleave``` which needs an answer (and more importantly, need coordinate and central design docs).

---

We also need to make sure any modifications on algorithm implementations (e.g. in imgproc, imgwarp, objdetect hog, etc) do not run into conflicts with the [OE-15 bit-exactness initiative](https://github.com/opencv/opencv/wiki/OE-15.-Bit-Exactness), /cc @terfendail . I understand that #10708 does not touch algorithm implementations yet, but it might happen soon.

vpisarev:
thank you, @seiko2plus! the PR is superseded by #11953, where I merged all the latest changes (hopefully)

seiko2plus:
@vpisarev, My pleasure! Yes, #11953 It covers the latest changes.. thank you

PR #10711, opened 2018-01-28
cmake: update reporting of excluded dispatching files

alalek:
- Reduce number of messages in CMake logs.
- Reduce unnecessary compiled files (SSE4 optimizations with AVX baseline).


cmake/OpenCVUtils.cmake
```diff
@@ -1423,6 +1423,15 @@ macro(ocv_copyfiles_add_target target list_var comment_str)
   add_custom_target(${target} DEPENDS "${OPENCV_DEPHELPER}/${target}")
 endmacro()
 
+macro(ocv_get_smart_file_name output_var fpath)
+  if("${fpath}" MATCHES "^(${OpenCV_BINARY_DIR})([/\\]?)(.*)$")
```
mshabunin: I suggest using [is_subdir](https://github.com/opencv/opencv/blob/a0ce12f8ba670b930d2cbb6cd0edda888ca391de/cmake/OpenCVUtils.cmake#L120-L137) function to avoid possible conflicts in cases when path contains regexp special symbols.

cmake/OpenCVUtils.cmake
```diff
@@ -1423,6 +1423,15 @@ macro(ocv_copyfiles_add_target target list_var comment_str)
...
+    set(${output_var} "<BUILD>/${CMAKE_MATCH_3}")
```
mshabunin: `file(RELATIVE_PATH <variable> <directory> <file>)` can be used here instead of captured value.

PR #10724, opened 2018-01-30
fix faster_rcnn sample crashed at PoolingInvoker on Windows7(x64).

bygreencn:
fix faster_rcnn sample crashed at PoolingInvoker operator() of pooling_layer on Windows7(x64).
Change  ofsbuf from std::vector<int> to AutoBuffer<int>, since at some pooling_layer ofsbuf is not initialized and "const int* ofsptr = &ofsbuf[0];" will crash.


vpisarev:
@bygreencn, thanks for the patch! let's fix in a slightly different way, see below

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -287,7 +287,7 @@ class PoolingLayerImpl : public PoolingLayer
             bool compMaxIdx = computeMaxIdx;
 
 #if CV_SIMD128
-            const int* ofsptr = &ofsbuf[0];
+            const int* ofsptr = (const int*)ofsbuf;
```
vpisarev: `const int* ofsptr = ofsbuf.empty() ? 0 : &ofsbuf[0];` is the better solution

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -230,7 +230,7 @@ class PoolingLayerImpl : public PoolingLayer
         Size kernel, stride, pad;
         int nstripes;
         bool computeMaxIdx;
-        std::vector<int> ofsbuf;
+        AutoBuffer<int> ofsbuf;
```
vpisarev: let's keep `std::vector<>`

alalek: If we need some alignment requirement, we could use `AutoBuffer<int, 0>` to use OpenCV `fastAlloc()`.

modules/dnn/src/layers/pooling_layer.cpp
bygreencn: Perfect, I think it is OK.

modules/dnn/src/layers/pooling_layer.cpp
bygreencn: I think it is OK.

vpisarev:
@alalek, there is no any alignment requirements of ofsbuf/ofsptr. @bygreencn, please, change AutoBuffer back to std::vector and retrieve the data pointer in the way I suggested.

vpisarev:
:+1:

PR #10730, opened 2018-01-30
Faster-RCNN and RFCN tests

dkurt:
### This pullrequest changes

Tests for Faster-RCNN and RFCN networks.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/435


vpisarev:
great! :+1:

PR #10732, opened 2018-01-30
Introduces error throw for line drawn with THICKNESS=0

### LINKED ISSUE #10598: cv::line thickness 0 not supported
cv::line can only draw lines with an thickness of at least 1 as far as I can see from drawing.cpp (method ThickLine).

I tried to disable line drawing by setting the thickness to 0, but in vain. This should be either supported or at least documented, imo.

Thanks for the great lib anyways :)

Riyuzakii:
Resolves #10598 
Restricts creation of lines with zero thickness by throwing an exception on the occurance of such an event.
Does so by adding a test for the same.


modules/imgproc/test/test_drawing.cpp
```diff
@@ -750,5 +750,11 @@ TEST(Drawing, putText_no_garbage)
 }
 
 
+TEST(Drawing, line)
+{
+    Mat mat = Mat::zeros(Size(100,100), CV_8UC1);
+
+    ASSERT_THROW(line(mat, Point(1,1),Point(99,99),Scalar(255),0,8,0), cv::Exception);
```
terfendail: It would be better to use enum item `LINE_8` instead of fixed value 8 or omit two last parameters of `line()` function

terfendail:
You also have to fix java test `testLineMatPointPointScalarInt` at `modules\core\misc\java\test\CoreTest.java` since it try to draw line with deprecated now zero thickness

Riyuzakii:
@terfendail I made the changes you asked me to make, why the build still fails?

terfendail:
Java test still try to draw the line with zero thickness that is prohibited now. You should change the line thickness to 1 in this test

terfendail:
👍

Riyuzakii:
@terfendail Thanks for helping me!

PR #10741, opened 2018-01-31
added functionality to get frames per second in utility.hpp 

Bhanudutta:
### This pullrequest changes
1.added functions to utility.hpp for getting frames per second
2.created new file tickmeter.cpp for definitions of functions in TickMeter class
3.added example to use new added frames per second functionality. 

The pull request is based on a previous pull request https://github.com/opencv/opencv/pull/10586 and changes are made according to suggestions by @vpisarev


Bhanudutta:
Why no one is responding to the pull request , is any thing wrong with my commits ?

vpisarev:
@Bhanudutta, thank you for the contribution! Let me close this PR for now. It's the kind of functionality that is easy implement by hand. On the other hand, some of the proposed details look too specific. For example, why reset FPS after 10 frames? why not after 100? or after 5? why not using running average to compute FPS? should we exclude the initial few frames in order to let the system to properly initialize and "warp up"? etc. Feel free to open the feature request where all those things can be discussed.

Issue #10718, opened 2018-01-29
opencv dnn readNetFromCaffe error: "caffe.BatchNormParameter" has no field named "scale_filler"


akerimcapar:
##### System information (version)
- OpenCV => 3.3
- Operating System / Platform => Windows 10 / 64 Bit
- Compiler => Visual Studio 2015

##### Detailed description
I trained a custom caffe network with following batch normalization layer:

layer {
  name: "bn1"
  type: "BatchNorm"
  bottom: "conv1"
  top: "bn1"
  param {
    lr_mult: 1
    decay_mult: 0
  }
  param {
    lr_mult: 1
    decay_mult: 0
  }
  batch_norm_param {
    scale_filler {
      type: "constant"
      value: 1
    }
    bias_filler {
      type: "constant"
      value: 0
    }
  }
}
*****************
When I tried to load the caffe prototxt file with dnn::readNetFromCaffe function I get the following error:
[libprotobuf ERROR C:\opencv330\opencv\sources\3rdparty\protobuf\src\google\protobuf\text_format.cc:298] Error parsing text-format caffe.NetParameter: 63:18: Message type "caffe.BatchNormParameter" has no field named "scale_filler".
OpenCV Error: Unspecified error (FAILED: ReadProtoFromTextFile(param_file, param). Failed to parse NetParameter file: c:/deploy.prototxt) in cv::dnn::ReadNetParamsFromTextFileOrDie, file C:\opencv330\opencv\sources\modules\dnn\src\caffe\caffe_io.cpp, line 1137
Exception: C:\opencv330\opencv\sources\modules\dnn\src\caffe\caffe_io.cpp:1137: error: (-2) FAILED: ReadProtoFromTextFile(param_file, param). Failed to parse NetParameter file: c:/deploy.prototxt in function cv::dnn::ReadNetParamsFromTextFileOrDie

Can't load network by using the following files:
prototxt:   c:/deploy.prototxt
caffemodel: c://deploy.caffemodel
bvlc_googlenet.caffemodel can be downloaded here:
http://dl.caffe.berkeleyvision.org/bvlc_googlenet.caffemodel
*******************
I have to mention that readNetFromCaffe function works properly when I train the network without scale_filler and bias_filler parameters.

##### Steps to reproduce

* Train a caffe custom network with at least one batch normalization layer as in "Detailed description" part.
* get the prototxt and caffemodel files and call the file paths as parameters of dnn::readNetFromCaffe function

[...]

PR #10746, opened 2018-01-31
Fix Batch Normalization layer imported from NVIDIA Caffe.

dkurt:
### This pullrequest changes

NVIDIA Caffe (v 0.16) stores batch normalization's parameters in the order `mean, std, stdScale(is not used), weights, bias`. Before PR OpenCV worked only with `mean, std, stdScale`; `mean, std, weights`; `mean, std, bias` and `mean, std, weights, bias` blobs order.

resolves https://github.com/opencv/opencv/issues/10718


vpisarev:
:+1:

PR #10749, opened 2018-01-31
samples: remove orphan documentation snippet: SBM_Sample.cpp

### LINKED ISSUE #10736: SBM_Sample.cpp: Multiple issues

- OpenCV => 3.4.0
- Compiler => Visual Studio 2017 Express

##### Detailed description
Various problems with the coding of SBM_Sample.cpp.

Quoted from [here](http://answers.opencv.org/question/183414/sbm_samplecpp-is-this-disparity-map-correct/):
1. The default images should be changed from the rubberwhale images (optical flow test images) to rectified left and right images (e.g. aloeL.jpg and aloeR.jpg).
2. An appropriate minDisparity and numDisparities should be updated to match the new default images. (e.g. numDisparities = 16x10 instead of 16x5)
3. Program ignores its image parameters - if this is intentional, then remove mention of parameters. If it is not, then add and test support for them.
4. Any references to the example should be found and updated, and if there's a test that runs this program, the test should be updated as well.

##### Steps to reproduce
Look at the code of SBM_Sample.cpp

alalek:
closes #10736


vpisarev:
:+1:

PR #10751, opened 2018-01-31
cuda_stream: do not allocate GPU memory by default (fixes #8725)

### LINKED ISSUE #8725: Calling cv::cuda::Stream::Null() results in a stray cudaMalloc() call
- OpenCV => master
- Operating System / Platform => Windows 10 64-Bit , CUDA 8.0

##### Detailed description
If I call `cv::cuda::Stream::Null()` in my code, it seems it calls `cudaMalloc()` without freeing the allocated memory, resulting in a memory leak error when tested with the command line memory checker `cuda-memcheck --leak-check full`. 

##### Steps to reproduce
```
__global__ void hello()
{
	printf("%u: hello\n", threadIdx.x);
}
int main()
{
	hello<<<1, 16, 0, cv::cuda::StreamAccessor::getStream(cv::cuda::Stream::Null())>>>();
	CheckCudaError(cudaDeviceSynchronize());
	CheckCudaError(cudaDeviceReset());
}
```
If you profile above code with `nvprof` then it calls `cudaMalloc()` without any `cudaFree()` followed.

nglee:
Current implementation allocates some GPU memory when a `Stream` object is defined for the first time for a GPU device. The amount of memory allocated is 10MB times the number of GPU devices by default. The allocated amount can be configured beforehand via `setBufferPoolConfig`.

Usage of this allocated memory is described at OpenCV doc for `BufferPool`, which says:
> `BufferPool` utilizes `cuda::Stream`'s allocator to create new buffers. It is particularly useful when BufferPoolUsage is set to true, or a custom allocator is specified for the `cuda::Stream`, and you want to implement your own stream based functions utilizing the same underlying GPU memory management.

IMHO, however, users won't expect some additional memories to be allocated when they just use `Stream` objects for asynchronous API calls.

Also, deallocation of this memory is carried out by the destructor of a global object `DefaultDeviceInitializer`. Since it is global, the destructor might be called after CUDA context is destroyed. In such cases, issues like #8725 may arise.

```
force_builders=Custom
```

### DIFF
--- a/modules/core/src/cuda_stream.cpp
+++ b/modules/core/src/cuda_stream.cpp
@@ -552,7 +552,7 @@ Stream cv::cuda::StreamAccessor::wrapStream(cudaStream_t stream)
 
 namespace
 {
-    bool enableMemoryPool = true;
+    bool enableMemoryPool = false;
 
     class StackAllocator : public GpuMat::Allocator
     {


alalek:
Could you provide problem reproducer / regression tests (some CUDA test, for example [here](https://github.com/opencv/opencv/blob/3.4.0/modules/cudaarithm/test/test_buffer_pool.cpp))?

nglee:
Ok, I will do that.

Meanwhile, on my machine (with 1 GPU, GTX 1080), this commit reduces the number of failed tests of opencv_test_cudaarithm.exe from 2074 to 11.

Before this commit, a lot of tests failed with the following message:
> all CUDA-capable devices are busy or unavailable in function cv::cuda::setDevice

PR #10754, opened 2018-02-01
ocl4dnnGEMV in case of row_size < 4

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/10731

### DIFF
--- a/modules/dnn/src/ocl4dnn/src/math_functions.cpp
+++ b/modules/dnn/src/ocl4dnn/src/math_functions.cpp
@@ -451,23 +451,27 @@ bool ocl4dnnGEMV<float>(const CBLAS_TRANSPOSE TransA,
 
         uint row_size = M;
         uint col_size = N;
-        size_t localsize[] = { 128 };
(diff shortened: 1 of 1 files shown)


modules/dnn/src/ocl4dnn/src/math_functions.cpp
```diff
@@ -452,7 +452,7 @@ bool ocl4dnnGEMV<float>(const CBLAS_TRANSPOSE TransA,
         uint row_size = M;
         uint col_size = N;
         size_t localsize[] = { 128 };
-        size_t globalsize[] = { row_size / 4 * localsize[0] };
+        size_t globalsize[] = { std::max(row_size / 4, uint(1)) * localsize[0] };
```
alalek: My suggestion is to put this kernel call under condition `if (row_size >= 4)`.
`row_size` is not passed to kernel, so there is way to avoid memory corruption.

dkurt: @alalek, Thanks! Edited.

PR #10757, opened 2018-02-01
Histogram tests updated

savuor:
### This pullrequest changes

* Tests added for issue #10687 fixed in #10688 (3+ dimensions are checked now in all interfaces)
* Obsolete types (like `IplImage*` and `CvMat*`) are not used anymore

### DIFF
--- a/modules/imgproc/test/test_histograms.cpp
+++ b/modules/imgproc/test/test_histograms.cpp
@@ -828,6 +828,7 @@ int CV_ThreshHistTest::prepare_test_case( int test_case_idx )
             orig_nz_count = total_size;
 
             values = Mat( 1, total_size, CV_32F );
+            indices = Mat();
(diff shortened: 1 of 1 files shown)


modules/imgproc/test/test_histograms.cpp
```diff
@@ -1206,14 +1180,82 @@ int CV_CalcHistTest::prepare_test_case( int test_case_idx )
...
+            CV_Assert(it.ptr != NULL);
+            *(float*)cvPtrND(sparsemat, it.node()->idx, 0, -2) = *(const float*)it.ptr;
+        }
+    }
 }
 
 
 static void
-cvTsCalcHist( IplImage** _images, CvHistogram* hist, IplImage* _mask, int* channels )
+cvTsCalcHist( vector<Mat> images, CvHistogram* hist, Mat mask, vector<int> channels )
```
alalek: `const std::vector<Mat>&`
`const std::vector<int>&`

here and below

savuor: fixed

PR #10764, opened 2018-02-02
update CMakeList.txt

sturkmen72:
### This pullrequest changes
HAVE_WEBP is undefined in
https://github.com/opencv/opencv/blob/master/modules/highgui/src/window_w32.cpp#L1938

### DIFF
--- a/cmake/OpenCVFindWebP.cmake
+++ b/cmake/OpenCVFindWebP.cmake
@@ -23,7 +23,7 @@ else()
     FIND_LIBRARY(WEBP_LIBRARY NAMES webp)
     MARK_AS_ADVANCED(WEBP_LIBRARY)
 
-    # handle the QUIETLY and REQUIRED arguments and set WEBFOUND_FOUND to TRUE if
+    # handle the QUIETLY and REQUIRED arguments and set WEBP_FOUND to TRUE if
     # all listed variables are TRUE
     INCLUDE(${CMAKE_ROOT}/Modules/FindPackageHandleStandardArgs.cmake)
     FIND_PACKAGE_HANDLE_STANDARD_ARGS(WebP DEFAULT_MSG WEBP_LIBRARY WEBP_INCLUDE_DIR)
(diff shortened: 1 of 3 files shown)


modules/imgcodecs/CMakeLists.txt
```diff
@@ -24,7 +24,6 @@ if(HAVE_JPEG)
 endif()
 
 if(HAVE_WEBP)
-  add_definitions(-DHAVE_WEBP)
```
savuor: Maybe the build fails because of this?

sturkmen72:
i am not sure that the patch is perfect but i tried it in my local and the problem is resolved. ( i missed warnings )

sturkmen72:
changes in modules/highgui/CMakeLists.txt solved the main problem. other changes can be considered good or useless. @savuor what is your opinion?

savuor:
👍

cmake/OpenCVFindWebP.cmake
```diff
@@ -23,11 +23,12 @@ else()
...
 
-    # handle the QUIETLY and REQUIRED arguments and set WEBFOUND_FOUND to TRUE if
+    # handle the QUIETLY and REQUIRED arguments and set WEBP_FOUND to TRUE if
     # all listed variables are TRUE
     INCLUDE(${CMAKE_ROOT}/Modules/FindPackageHandleStandardArgs.cmake)
     FIND_PACKAGE_HANDLE_STANDARD_ARGS(WebP DEFAULT_MSG WEBP_LIBRARY WEBP_INCLUDE_DIR)
 
     SET(WEBP_LIBRARIES ${WEBP_LIBRARY})
     SET(WEBP_INCLUDE_DIRS ${WEBP_INCLUDE_DIR})
+    SET(WEBP_FOUND TRUE)
```
alalek: This is done by `FIND_PACKAGE_HANDLE_STANDARD_ARGS()` (if `WEBP_LIBRARY` / `WEBP_INCLUDE_DIR` are defined).
But variable name should be fixed (check for `FOUND_VAR` parameter).

cmake/OpenCVFindLibsGrfmt.cmake
```diff
@@ -111,7 +111,7 @@ if(WITH_WEBP AND NOT WEBP_FOUND
 
   add_subdirectory("${OpenCV_SOURCE_DIR}/3rdparty/libwebp")
   set(WEBP_INCLUDE_DIR "${${WEBP_LIBRARY}_SOURCE_DIR}/src")
-  set(HAVE_WEBP 1)
+  set(HAVE_WEBP YES)
```
alalek: These changes in this file are not necessary. Please revert them.

PR #10799, opened 2018-02-06
OpenCV face detection network using Inference Engine backend

dkurt:
### This pullrequest changes

best observed time at 300x300 input images:

| Default OpenCV backend | Inference Engine @ OpenCV |  Inference Engine with Model Optimizer, standalone application |
|-----|-----|-----|
| 14.57ms | 11.49ms (x1.27) |  11.8972ms | 

L2-normalization layer from the model has default backend implementation.


vpisarev:
:+1:

PR #10803, opened 2018-02-07
Update BufferReader documentation with some example code

nglee:
Updated documentation for BufferReader with some example code.


vpisarev:
probably, the sample should include some loop to demonstrate that the temporary GpuMat's are deallocated and the corresponding memory is automatically returned to the pool.

nglee:
@vpisarev I've added another sample code. If there are other things that need to be improved, please tell me.

vpisarev:
looks good! :+1:

PR #10806, opened 2018-02-07
build: eliminate calls of removed functionality from C++17

alalek:
Most part is deprecated since C++11

OpenEXR builtin sources are still unfixed. Need to disable them: `WITH_OPENEXR=OFF`.

MSVS flag: `/std:c++17`
GCC flag: `-std=c++17`


vpisarev:
great! :+1:

PR #10808, opened 2018-02-07
Add new Mat constructor

### LINKED ISSUE #10792: Feature: Add matrix size argument to initializer_list Mat constructor
OpenCV 3.3 introduced C++11 **initializer_list** constructors for `Mat`, a very welcome feature. However, these constructors have no size information (i.e. rows and columns), so new matrices need to call `reshape()` like this:

```.cpp
Mat_<double> A = Mat_<double>({1, 0, 0, 0, 1, 0, 0, 0, 1}).reshape(1, 3);
```

**Proposal**

I propose to add matrix **size** (rows, cols) to the constructor along with the initializer list so that matrices are fully initialized in the constructor. Construction would look like this: (compare with the above)

```.cpp
Mat_<double> A(3, 3, {1, 0, 0, 0, 1, 0, 0, 0, 1});
```

Alternatively, you could require **only the number of rows**, because columns will be known (`cols = list.size() / rows`):

```.cpp
Mat_<double> A(3, {1, 0, 0, 0, 1, 0, 0, 0, 1});
```

This syntax matches the *old* constructor taking a size and a pointer to an array `Mat(int rows, int cols, double* data)`).

**Possible implementation** (for the rows-only case)

```.cpp
template<typename _Tp, typename> inline
Mat::Mat(int rows, const std::initializer_list<_Tp> list)
    : flags(MAGIC_VAL | traits::Type<_Tp>::value | CV_MAT_CONT_FLAG), dims(2), rows((int)list.size()),
      cols(1), data(0), datastart(0), dataend(0), datalimit(0), allocator(0), u(0), size(&rows), step(0)
{
    if(list.size() == 0)
        return;
    Mat(rows, (int)list.size() / rows, traits::Type<_Tp>::value, (uchar*)list.begin()).copyTo(*this);
}
```

**Vectors** do not have this problem and are fine with the existing constructor.

take1014:
resolves #10792

### This pullrequest changes
1. Add new Mat constructor.


modules/core/include/opencv2/core/mat.hpp
```diff
@@ -2208,6 +2213,7 @@ template<typename _Tp> class Mat_ : public Mat
 
 #ifdef CV_CXX11
     Mat_(std::initializer_list<_Tp> values);
+    Mat_(const std::initializer_list<int> sizes, const std::initializer_list<_Tp> values);
```
alalek: explicit (both constructors)

modules/core/include/opencv2/core/mat.inl.hpp
```diff
@@ -584,6 +584,16 @@ Mat::Mat(const std::initializer_list<_Tp> list)
...
     Mat((int)list.size(), 1, traits::Type<_Tp>::value, (uchar*)list.begin()).copyTo(*this);
 }
+
+template<typename _Tp, typename> inline
+Mat::Mat(const std::initializer_list<int> sizes, const std::initializer_list<_Tp> list)
+    : flags(MAGIC_VAL | traits::Type<_Tp>::value | CV_MAT_CONT_FLAG), dims(2), rows((int)list.size()),
+      cols(1), data(0), datastart(0), dataend(0), datalimit(0), allocator(0), u(0), size(&rows), step(0)
+{
+    if(sizes.size() != 1 || *sizes.begin() == 0 || list.size() == 0)
+        return;
```
alalek: > `sizes.size() != 1`

Why? This should be allowed to create ND matrices.

> dims(2), rows((int)list.size()), cols(1),

Should be fixed. Initialize fields for an empty `Mat`.
It is better to use C++11 [Delegating constructor](http://en.cppreference.com/w/cpp/language/initializer_list) here.

Also this check should be added:
`sizes[0]*sizes[1]*..*sizes[dims-1] == list.total()`

> return;

No. This is an error => `CV_Error` / `CV_Assert`.

modules/core/test/test_mat.cpp
```diff
@@ -1719,17 +1719,23 @@ TEST(Mat, from_initializer_list)
 {
     Mat A({1.f, 2.f, 3.f});
     Mat_<float> B(3, 1); B << 1, 2, 3;
+    Mat_<float> C({1}, {1,2,3});
```
alalek: This should not work (or better to compile). "Sizes" declares matrix with single element, but it is initialized with 3 values.

modules/core/include/opencv2/core/mat.inl.hpp
```diff
@@ -584,6 +584,16 @@ Mat::Mat(const std::initializer_list<_Tp> list)
...
+    Mat((int)list.size() / *sizes.begin(), *sizes.begin(), traits::Type<_Tp>::value, (uchar*)list.begin()).copyTo(*this);
```
alalek: `(int)list.size() / *sizes.begin()`

This is not "dims" parameter.
"dims" is `sizes.size()`

modules/core/include/opencv2/core/mat.hpp
take1014: I fix line 2216 as follows:
explicit Mat_(const std::initializer_list<int> sizes, const std::initializer_list<_Tp> values);

modules/core/include/opencv2/core/mat.inl.hpp
take1014: I changed it to use Mat () as Delegating constructor.

take1014: And added check using CV_Assert.

modules/core/test/test_mat.cpp
take1014: I had tryed to use the following constructor for initialize.
->Mat (int rows, int cols, int type, const Scalar & s);
But, as you say, using the constructor below is correct.
->Mat (int ndims, const int * sizes, int type, const Scalar & s);
If I still do not understand what you say, I apologize.

modules/core/include/opencv2/core/mat.inl.hpp
```diff
@@ -584,6 +584,19 @@ Mat::Mat(const std::initializer_list<_Tp> list)
...
+
+template<typename _Tp, typename> inline
+Mat::Mat(const std::initializer_list<int> sizes, const std::initializer_list<_Tp> list)
+    : Mat()
+{
+    size_t size_total = 1;
+    int *sz = (int*)sizes.begin();
+    for(int i = 0; i < sizes.size(); i++)
+        size_total *= sz[i];
+    CV_Assert(list.size() != 0 || size_total != list.size());
```
alalek: This condition is always true for any on-empty list.
Also "size_total != list.size()" is wrong initially.

modules/core/include/opencv2/core/mat.inl.hpp
```diff
@@ -584,6 +584,19 @@ Mat::Mat(const std::initializer_list<_Tp> list)
         return;
     Mat((int)list.size(), 1, traits::Type<_Tp>::value, (uchar*)list.begin()).copyTo(*this);
 }
+
+template<typename _Tp, typename> inline
```
alalek: > template<typename _Tp, **typename**>

Why we need second "typename" here?

modules/core/include/opencv2/core/mat.inl.hpp
```diff
@@ -584,6 +584,19 @@ Mat::Mat(const std::initializer_list<_Tp> list)
...
+Mat::Mat(const std::initializer_list<int> sizes, const std::initializer_list<_Tp> list)
+    : Mat()
+{
+    size_t size_total = 1;
+    int *sz = (int*)sizes.begin();
+    for(int i = 0; i < sizes.size(); i++)
```
alalek:
```.diff
-int i = 0
+size_t i = 0
```

Because "sizes.size()" has size_t type (almost, http://en.cppreference.com/w/cpp/utility/initializer_list/size ).

Anyway it is better to use C++11 "foreach":

```
for (auto s : sizes)
{
    size_total *= s;
}
```

modules/core/include/opencv2/core/mat.inl.hpp
```diff
@@ -584,6 +584,19 @@ Mat::Mat(const std::initializer_list<_Tp> list)
...
+        size_total *= sz[i];
+    CV_Assert(list.size() != 0 || size_total != list.size());
+    flags |= (traits::Type<_Tp>::value | CV_MAT_CONT_FLAG);
```
alalek: Touching "flags" is wrong - we should not make Mat inconsistent.
Anyway this is not necessary - `copyTo()` will do it.

modules/core/test/test_mat.cpp
```diff
@@ -1719,17 +1719,23 @@ TEST(Mat, from_initializer_list)
 {
     Mat A({1.f, 2.f, 3.f});
     Mat_<float> B(3, 1); B << 1, 2, 3;
+    Mat_<float> C({3}, {1,2,3});
```
alalek: Original issue reports about creation of 2D Mat. I believe we should test them too.

modules/core/include/opencv2/core/mat.inl.hpp
take1014: I made a mistake. 
What I want to do was not to generate an assertion if the condition of "size_total == list.size()" is satisfied. I fix this code soon.

modules/core/include/opencv2/core/mat.inl.hpp
take1014: Actually, we don't need this second "typename". I fix two files to eraze them.

modules/core/include/opencv2/core/mat.inl.hpp
take1014: >for (auto s : sizes)
>{
>    size_total *= s;
>}
->I think that this code is most simplified and suitable.

vpisarev:
looks good! :+1:

Issue #10702, opened 2018-01-26
DNN: Access violation in deconv2d layer


philipp-fischer:
[minimal_net.py.txt](https://github.com/opencv/opencv/files/1668899/minimal_net.py.txt)

##### System information (version)
- OpenCV => 3.4 (Master Branch) 
- Operating System / Platform => Windows 64 Bit
- Compiler => Visual Studio 2015

##### Detailed description
Exception in deconvolution layer.
Easily reproducible with minimal example (see attachments) 
Access violation happens im "convolution_layer.cpp" line 1089.
(in the two for loops "val +=...") 

##### Steps to reproduce
Set up a minimal one-layer tensorflow network as in the attached file and export it.
Remove unneeded ops using the tf optimizer. 
Load the protobuf into opencv with readNetFromTensorflow and do a forward pass. 
It will crash. 
![20180126_200027](https://user-images.githubusercontent.com/12006597/35456341-61e96fa6-02d5-11e8-9a75-ad3f728a97f1.png)
[minimal_net.py.txt](https://github.com/opencv/opencv/files/1668900/minimal_net.py.txt)

[deconv.zip](https://github.com/opencv/opencv/files/1668978/deconv.zip)

akhandait:
@dkurt I would like to work on this issue, can you give me some initial directions to get started?

philipp-fischer:
Great. You can build open this example:

https://github.com/opencv/opencv/blob/master/samples/dnn/tf_inception.cpp

But you can really strip it down to
readNetFromTensorflow
SetInput and net.forward. 

I attached a zip file that contains the optimized_model.pb you can load.

dkurt:
@akhandait, as @philipp-fischer said firstly we need to reproduce a error. Download attached graph, try to run it using OpenCV.
Then we need to find a problem. There is a [generate_tf_models.py](https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/tensorflow/generate_tf_models.py) script to generate test data using TensorFlow. We already have one test for `conv2d_transpose` so we need to figure out what configuration leads the crush. I think you can start change strides and padding arguments then change the rest of parameters until you'll catch a bug.

akhandait:
@dkurt @philipp-fischer Thanks, I will get on it now.

philipp-fischer:
I had a look at the `generate_tf_models.py` script.
The conv2d_transpose layer there only uses strides of 1 and VALID padding.
[...]

PR #10850, opened 2018-02-12
Fix bugs in layers imported from TensorFlow

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/10702

* Fixed bugs in deconvolution layer imported from TensorFlow.
* import batch normalization with mean-variance normalization from TensorFlow.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/440


vpisarev:
thanks! :+1:

PR #10860, opened 2018-02-13
make tmpRow proper size before copyTo to avoid reallocated tmpCol

PalkoNazar:
### Fixed hierarchicalClustering function
Before this commit dists.row(minIdx) changed while dists.col(minIdx) didn't change (but it must)


alalek:
@PalkoNazar Could your provide some reproducer (code and data/images) for this problem detection?

vpisarev:
ok, looks like reasonable patch. Let's merge it in :+1:

PR #10861, opened 2018-02-14
Fix for CUDA_Arithm/Dft.Algorithm/0 test

nglee:
The previously failed test CUDA_Arithm/Dft.Algorithm/0 had an issue with setting flags for the DFT algorithm. After this fix, the test passes. This fix reduces the number of failed tests from 11 to 10.

Tested on GTX1080, Driver version 376.51.


vpisarev:
:+1:

PR #10868, opened 2018-02-14
Samples: reworked bgfg_bgsegm

mshabunin:
### This pullrequest changes

- use `<iostream>` instead of _printf_
- moved mask smoothing option to interactive loop
- minor code refactoring

### DIFF
--- a/samples/cpp/bgfg_segm.cpp
+++ b/samples/cpp/bgfg_segm.cpp
@@ -1,111 +1,113 @@
 #include "opencv2/core.hpp"
-#include <opencv2/core/utility.hpp>
 #include "opencv2/imgproc.hpp"
-#include "opencv2/video/background_segm.hpp"
+#include "opencv2/video.hpp"
 #include "opencv2/videoio.hpp"
 #include "opencv2/highgui.hpp"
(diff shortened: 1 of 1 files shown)


samples/cpp/bgfg_segm.cpp
```diff
@@ -1,111 +1,113 @@
...
-        return -1;
+        parser.printErrors();
+        return 1;
     }
 
-    namedWindow("image", WINDOW_NORMAL);
-    namedWindow("foreground mask", WINDOW_NORMAL);
-    namedWindow("foreground image", WINDOW_NORMAL);
-    namedWindow("mean background image", WINDOW_NORMAL);
+    VideoCapture cap(useCamera ? 0 : file.c_str());
```
alalek: > useCamera ? 0 : file.c_str()

This calls "VideoCapture((const char*)NULL)" instead of "VideoCapture((int)0)".
Consider using `.open()` method of `cv::Ptr<VideoCapture>` with different initializers.

mshabunin: Fixed, thank you!

PR #6000, opened 2016-01-20
CMake: fix python detection

alalek:
Python libs detection is broken on Ubuntu 14.04 (with **system** CMake 2.8.12.2) via `find_package(PythonLibs)` with specified `PYTHON_LIBRARY` and `PYTHON_INCLUDE_DIR`.
This problem cause a python detection failure on the second CMake pass where these variables are got from CMake cache.

Current solution is to bypass python search if python is already found.

PR #10873, opened 2018-02-15
try to detect PythonLibs every time CMake is executed if it is not found yet

andy-held:
When the python interpreter is found but the libraries are not, reconfiguring CMake will not search for them again, since PYTHONINTERP_FOUND is already true.

### This pullrequest changes

Also checking whether the libraries were found enables users to correct the cause for CMake not finding the libraries and reconfigure without having to delete or modify CMakeCache.txt by hand.


alalek:
See #6000 . Strong validation of this patch is required.
Consider configuring project during first CMake run.
Changing values on-the-fly doesn't work in many cases (and not tested), but we want that simple "cmake ." (with the same args) works.

BTW, `cmake -U*PYTHON* ...` can help to you.

PR #10676, opened 2018-01-24
PriorBox layer with explicit normalized sizes

dkurt:
### This pullrequest changes

Allow to specify explicit relative prior boxes' sizes.

related:
http://answers.opencv.org/question/180105/tensorflow-custom-model-in-opencv/
https://github.com/opencv/opencv/issues/10601
https://github.com/opencv/opencv/issues/10431

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/431

PR #10874, opened 2018-02-15
Use only absolute prior boxes explicit sizes. Remove scales attributes.

dkurt:
### This pullrequest changes

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/441

Experimented with PriorBox layer. Original SSD model generates proposals using bounding boxes sizes in pixels. Then normalize them by input image size.

Before PR, we put relative bounding boxes' sizes to proposals generator thinking that it helps with different input sizes (only TensorFlow models). In example, both relative and absolute proposals give the same results for 300x300 images:

* 300x300 input image, relative proposals sizes
```yaml
  attr {
    key: "width"
    value {
      tensor {
        dtype: DT_FLOAT
        tensor_shape {
          dim {
            size: 3
          }
        }
        float_val: 0.1
        float_val: 0.282843
        float_val: 0.141421
      }
    }
  }
```

* 300x300 input image, absolute proposals sizes
```yaml
  attr {
    key: "width"
    value {
      tensor {
        dtype: DT_FLOAT
        tensor_shape {
          dim {
            size: 3
          }
        }
        float_val: 30.0
        float_val: 84.8528137207
        float_val: 42.4264068604
      }
    }
  }
```

However we need to keep that sizes for any input dimensions:

* 512x512 input image, relative proposals sizes

This PR reverts changer from https://github.com/opencv/opencv/pull/10676 and use only absolute explicit widths and heights:

* 512x512 input image, absolute proposals sizes

Also removed `scales` attribute of prior box layer.

### DIFF
--- a/modules/dnn/src/layers/prior_box_layer.cpp
+++ b/modules/dnn/src/layers/prior_box_layer.cpp
@@ -179,7 +179,6 @@ public:
     }
 
     PriorBoxLayerImpl(const LayerParams &params)
-        : _boxWidth(0), _boxHeight(0)
     {
         setParamsFrom(params);
         _minSize = getParameter<float>(params, "min_size", 0, false, 0);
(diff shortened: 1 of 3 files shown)


modules/dnn/src/layers/prior_box_layer.cpp
```diff
@@ -280,6 +277,14 @@ class PriorBoxLayerImpl : public PriorBoxLayer
     }
 
 #ifdef HAVE_OPENCL
+    static inline void setKernelArg(ocl::Kernel& kernel, int id, const UMat& m)
+    {
+        if (!m.empty())
+            kernel.set(id, ocl::KernelArg::PtrReadOnly(m));
+        else
+            kernel.set(id, nullptr);
```
alalek: I believe we should avoid this approach.
It is better to use compile-time definitions and `#ifdef` blocks in OpenCL code instead of dummy parameters. It will produce more effective code without such hacks.

dkurt: @alalek, thanks! I'll manage it.

vpisarev:
looks great! :+1:

PR #10878, opened 2018-02-15
Improve #10808

take1014:
Improve #10808 

### This pullrequest changes
Improve #10808 
When I worked on the issue of # 10808, I got advice from alalek  and thought that it would be better to fix the existing constructor of Mat, too.

```
**WIP**
force_builders=Docs
```


alalek:
> Commits 11

You should reset your source branch on **updated** *upstream* "master" branch before adding changes.
Refer to Git/GitHub usage guidelines.

take1014:
>You should reset your source branch on updated upstream "master" branch before adding changes.
-> Thanks for the follow. I updated this branch with master.

alalek:
> Commits 15

Actually don't.
All merge commits should be eliminated from the patch.

Ideally there is should one commit per pull request. All "fixup" commits should be squashed into one commit.

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -987,8 +987,7 @@ class CV_EXPORTS Mat
 #ifdef CV_CXX11
     /** @overload
     */
-    template<typename _Tp, typename = typename std::enable_if<std::is_arithmetic<_Tp>::value>::type>
-    explicit Mat(const std::initializer_list<_Tp> list);
+    template<typename _Tp> explicit Mat(const std::initializer_list<_Tp> list);
```
alalek: Nope.
Check this PR: #9507

take1014: Thank you for teaching me a lot.
I didn't understand why this constructor was added, but I could understand it now.   I think I will close this pull request, what do you think?
>All merge commits should be eliminated from the patch.
-> I understood.

take1014: Prior to sending the pull request, I should have confirmed whether it was a problem and should receive advice.

PR #10880, opened 2018-02-15
add missing DNN header to opencv2/opencv.hpp

### LINKED ISSUE #10594: opencv2/dnn.hpp is not in opencv2/opencv.hpp
- OpenCV => 3.4-dev even with master and opencv release 3.4.0
- Compiler => Visual Studio 2015-17 cmake 3.10.1

##### Detailed description

Hi,

I want yo use dnn module. I think that I can include only  :  
`#include<opencv2/opencv.hpp>`
but that's not enough because there is  no #include "opencv2/dnn.hpp" in include\opencv2\opencv.hpp.
I check this in my build configuration and in https://github.com/opencv/opencv/releases/download/3.4.0/opencv-3.4.0-vc14_vc15.7z

Is it expected behaviour ?

StevenPuttemans:
resolves #10594

### DIFF
--- a/include/opencv2/opencv.hpp
+++ b/include/opencv2/opencv.hpp
@@ -58,6 +58,9 @@
 #ifdef HAVE_OPENCV_FEATURES2D
 #include "opencv2/features2d.hpp"
 #endif
+#ifdef HAVE_OPENCV_DNN
+#include "opencv2/dnn.hpp"
+#endif
 #ifdef HAVE_OPENCV_FLANN
 #include "opencv2/flann.hpp"
 #endif


terfendail:
Looks like it make sense to postpone this change due to DNN API instability

StevenPuttemans:
Actually in principle DNN module should be built without issues right? It is a default module and although a lot of changes happens, it is build by default?

terfendail:
It is included to default build. But at the moment dnn module changes don't affect header precompilation if one trying to use latest OpenCV master without dnn. Though I understand it is not a big issue.

vpisarev:
ok, let's put it in, it would not break any existing code :+1:

StevenPuttemans:
Ok thx!

PR #10889, opened 2018-02-17
add selu activation function

### LINKED ISSUE #10888: feature request: add SELU activation function
OpenCV supports ReLU, ELU, and several other activation functions. I would like to implement a relatively recently published activation function called SELU (Scaled Exponential Linear Unit, https://arxiv.org/abs/1706.02515) which has been shown to have several nice properties leading to better learning networks.

lakshayg:
resolves #10888

### This pullrequest changes
* adds the selu activation function

I have not yet added the test data and request the devs to guide me on how to do it.

### DIFF
--- a/modules/dnn/src/tensorflow/tf_importer.cpp
+++ b/modules/dnn/src/tensorflow/tf_importer.cpp
@@ -1553,6 +1553,14 @@ void TFImporter::populateNet(Net dstNet)
             layer_id[name] = id;
             connectToAllBlobs(layer_id, dstNet, parsePin(layer.input(0)), id, layer.input_size());
         }
+        else if (type == "Selu")
+        {
+            layerParams.set("scale", 1.0507009873554804934193349852946f);
+            layerParams.set("alpha", 1.6732632423543772848170429916717f);
+            int id = dstNet.addLayer(name, "SELU", layerParams);
+            layer_id[name] = id;
+            connectToAllBlobs(layer_id, dstNet, parsePin(layer.input(0)), id, layer.input_size());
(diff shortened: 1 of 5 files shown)


modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -1553,6 +1553,14 @@ void TFImporter::populateNet(Net dstNet)
             layer_id[name] = id;
             connectToAllBlobs(layer_id, dstNet, parsePin(layer.input(0)), id, layer.input_size());
         }
+        else if (type == "Selu")
+        {
+            layerParams.set("scale", 1.0507009873554804934193349852946f);
```
dkurt: Could you please provide an origin of these magic values? Are they results of some optimization problem that we can reproduce locally or they are just some irrational numbers that have too complicated formula? This way it'd be nice to put a reference to a some proof that they are optimal.

lakshayg:
I have put a link to the paper in the linked issue. Here it is again: https://arxiv.org/abs/1706.02515

dkurt:
@lakshayg, thank you for your contribution! Unfortunately, I did not find any strong proofs why this activation is so useful. I think we can add this layer later, when it'll be more popular. For now you can test custom layers creation from PR https://github.com/opencv/opencv/pull/11129.

lakshayg:
Sure, no problem.

dkurt:
Please look at custom layers registration: https://docs.opencv.org/master/dc/db1/tutorial_dnn_custom_layers.html.

PR #10891, opened 2018-02-17
ReLU6 layer ocl support

pengli:
include relu6 ocl kernel and layer fusion support

### DIFF
--- a/modules/dnn/src/opencl/conv_layer_spatial.cl
+++ b/modules/dnn/src/opencl/conv_layer_spatial.cl
@@ -48,19 +48,22 @@
 
 #if defined(FUSED_CONV_RELU)
 #define ACTIVATION_RELU_FUNCTION(x, c) ((Dtype)(x) > 0 ? (Dtype)(x) : ((Dtype)(x) * (Dtype)(negative_slope)))
-#define NEGATIVE_SLOPE_ARG Dtype negative_slope,
+#define FUSED_ARG Dtype negative_slope,
 #elif defined(FUSED_CONV_PRELU)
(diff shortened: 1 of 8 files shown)


modules/dnn/src/opencl/conv_layer_spatial.cl
```diff
@@ -49,18 +49,27 @@
...
 #define NEGATIVE_SLOPE_ARG Dtype power,
+#define RELU6_MAXVAL_ARG
 #elif defined(FUSED_CONV_TANH)
 #define ACTIVATION_RELU_FUNCTION(x, c) tanh(x)
 #define NEGATIVE_SLOPE_ARG
+#define RELU6_MAXVAL_ARG
+#elif defined(FUSED_CONV_RELU6)
+#define ACTIVATION_RELU_FUNCTION(x, c) (clamp((Dtype)(x), min_value, max_value))
+#define NEGATIVE_SLOPE_ARG Dtype min_value,
+#define RELU6_MAXVAL_ARG Dtype max_value,
```
alalek:
```.diff
-#define NEGATIVE_SLOPE_ARG Dtype min_value,
-#define RELU6_MAXVAL_ARG Dtype max_value,
+#define NEGATIVE_SLOPE_ARG
+#define RELU6_MAXVAL_ARG Dtype min_value, Dtype max_value,
```

or rename this args into "FUSED_ARGS" (`FUSED_CONV_POWER` case is not consistent too)

pengli: @alalek，thanks for the comment，patch is updated with this change

PR #10893, opened 2018-02-18
core: remove '-' from separators list of disabled CPU flags

alalek:
To allow runtime disabling of AVX512-SKX via 'OPENCV_CPU_DISABLE' parameter

```
docker_image:Linux x64 Debug=gstreamer:16.04
```


vpisarev:
:+1:

Issue #8938, opened 2017-06-18
Can `–default-stream per-thread`  be used with opencv ?


duchengyao:
##### System information (version)

- OpenCV => 3.2
- Operating System / Platform => Linux
- Compiler => Cmake

##### Detailed description

I use Muliptle streams in opencv, but blocked on the default stream, like this:
![](https://devblogs.nvidia.com/parallelforall/wp-content/uploads/2015/01/stream_legacy-624x178.png)

How can streams work like this:
![](https://devblogs.nvidia.com/parallelforall/wp-content/uploads/2015/01/stream_per_thread-624x168.png)

I found this can be solve by `nvcc –default-stream per-thread -o XXX` [(GPU Pro Tip: CUDA 7 Streams Simplify Concurrency)](https://devblogs.nvidia.com/parallelforall/gpu-pro-tip-cuda-7-streams-simplify-concurrency/). but how can `–default-stream per-thread`  be used with opencv ?

The same problem posted on devtalk.nvidia.com [https://devtalk.nvidia.com/default/topic/995591/muliptle-streams-don-t-speed-up-processing/?offset=5#5169503](https://devtalk.nvidia.com/default/topic/995591/muliptle-streams-don-t-speed-up-processing/?offset=5#5169503)

alalek:
Try [`CUDA_NVCC_FLAGS`](https://github.com/opencv/opencv/blob/3.2.0/cmake/FindCUDA.cmake#L86-L89) CMake option.

duchengyao:
Thanks.
I'm using `cmake gui`   but not find 'CUDA_NVCC_FLAGS' option

duchengyao:
I used `CUDA_NVCC_FLAGS` and default stream was gone.

```
	cuda::Stream stream0;
	cuda::Stream stream1;
	Mat src;
    src = imread(abspath("../data/aloeL-L.jpg"), IMREAD_GRAYSCALE);
    if (src.empty()) throw runtime_error("can't open ../data/aloeL.jpg");

    vector<KeyPoint> keypoints;


    cv::Ptr<cv::cuda::FastFeatureDetector> d_FAST = cv::cuda::FastFeatureDetector::create(20);
    cuda::GpuMat d_src=cuda::GpuMat(src.rows,src.cols,CV_8UC4);
    cuda::GpuMat d_src1=cuda::GpuMat(src.rows,src.cols,CV_8UC4);

    cuda::GpuMat d_keypoints;
[...]

PR #10906, opened 2018-02-20
cuda_fast : multi stream safety

nglee:
Currently, CUDA FAST feature detection algorithm is not safe if multiple CUDA streams run it in parallel. The reason behind this is that the device code access a global counter that is not private to each FAST algorithm instance. This PR fixes this issue by converting the global counter to private counter.

4 test cases for regression is also added. Some of these added tests fails without this fix. Which tests fail is nondeterminant, which means that we have race condition. Tested with GTX1080, driver version 376.51.

I've encountered this race condition when trying to solve issue https://github.com/opencv/opencv/issues/8938.

```
force_builders=Custom
```

### DIFF
--- a/modules/cudafeatures2d/test/test_features2d.cpp
+++ b/modules/cudafeatures2d/test/test_features2d.cpp
@@ -42,10 +42,6 @@
 
 #include "test_precomp.hpp"
 
-#ifdef _OPENMP
-#include <omp.h>
-#endif
-
 #ifdef HAVE_CUDA
 
 #include <cuda_runtime_api.h>
@@ -86,15 +82,7 @@ CUDA_TEST_P(FAST, Accuracy)
 
     if (!supportFeature(devInfo, cv::cuda::GLOBAL_ATOMICS))
     {
-        try
-        {
-            std::vector<cv::KeyPoint> keypoints;
-            fast->detect(loadMat(image), keypoints);
-        }
-        catch (const cv::Exception& e)
-        {
-            ASSERT_EQ(cv::Error::StsNotImplemented, e.code);
-        }
+        throw SkipTestException("CUDA device doesn't support global atomics");
     }
     else
     {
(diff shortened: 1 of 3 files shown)


modules/cudafeatures2d/test/test_features2d.cpp
```diff
@@ -102,6 +108,62 @@ CUDA_TEST_P(FAST, Accuracy)
...
+        cv::cuda::HostMem image(image_);
+
+        GpuMat d_keypoints[2];
+        std::vector<cv::KeyPoint> keypoints[2];
+        cv::Ptr<cv::cuda::FastFeatureDetector> d_fast[2];
+
+        d_fast[0] = cv::cuda::FastFeatureDetector::create(threshold, nonmaxSuppression);
+        d_fast[1] = cv::cuda::FastFeatureDetector::create(threshold, nonmaxSuppression);
+
+        #pragma omp parallel num_threads(2)
```
alalek: Please use `cv::parallel_for_(Range(0, 2), ...)` instead of OpenMP.

[Example](https://github.com/opencv/opencv/blob/3.4.0/modules/core/test/test_rand.cpp#L386-L420)

modules/cudafeatures2d/test/test_features2d.cpp
```diff
@@ -102,6 +108,62 @@ CUDA_TEST_P(FAST, Accuracy)
...
+#ifdef _OPENMP
+CUDA_TEST_P(FAST, Async)
+{
+    cv::Mat image_ = readImage("features2d/aloe.png", cv::IMREAD_GRAYSCALE);
+    ASSERT_FALSE(image_.empty());
+
+    if (!supportFeature(devInfo, cv::cuda::GLOBAL_ATOMICS))
+    {
+        try
+        {
```
alalek: Do we really need this code path?

Perhaps it is better to skip test:
```
if (!supportFeature(devInfo, cv::cuda::GLOBAL_ATOMICS))
    throw SkipTestException("CUDA device doesn't support global atomics");
```

modules/cudafeatures2d/test/test_features2d.cpp
nglee: Ok, I'll fix it in a few hours.

modules/cudafeatures2d/test/test_features2d.cpp
nglee: I think you are right.

PR #10916, opened 2018-02-21
videoio: add missing CAP_PROP_SAR_NUM/DEN propepties into .hpp file

alalek:
reuse av_guess_sample_aspect_ratio() if available in ffmpeg


vpisarev:
:+1:

PR #10919, opened 2018-02-22
Issues found by static analysis + misc changes

mshabunin:
### This pullrequest changes

- compilation on Windows using mingw-w64 + Ninja: one error and one warning fixed
- use C++11 types enabled recently in AVIContainer module
- static analysis found several problems: uninitialized class members, possible memory leak in TIFF handler, read without check in AVIContainer

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/videoio/src/container_avi.cpp
+++ b/modules/videoio/src/container_avi.cpp
@@ -7,66 +7,57 @@
 namespace cv
 {
 
-const unsigned int RIFF_CC = CV_FOURCC('R','I','F','F');
-const unsigned int LIST_CC = CV_FOURCC('L','I','S','T');
-const unsigned int HDRL_CC = CV_FOURCC('h','d','r','l');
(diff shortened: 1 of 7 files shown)


modules/videoio/src/container_avi.cpp
```diff
@@ -742,8 +734,8 @@ void BitStream::patchInt(int val, size_t pos)
     }
     else
     {
-        long fpos = ftell(m_f);
-        fseek(m_f, (long)pos, SEEK_SET);
+        int32_t fpos = ftell(m_f);
+        fseek(m_f, (int32_t)pos, SEEK_SET);
```
alalek: [ftell](http://www.cplusplus.com/reference/cstdio/ftell/) / [fseek](http://www.cplusplus.com/reference/cstdio/fseek/) use "long int" type (it is still 32-bit on Windows, even in x86-64 mode - so using of fgetpos/fsetpos may help).

Also we need `CV_Assert(pos < (1<<31));` check.

PR #10921, opened 2018-02-22
Updated warpAffine test to ensure bit-exactness for CV_8U

terfendail:
### This pullrequest changes
warpAffine accuracy test error tolerance set to zero for CV_8U


vpisarev:
@terfendail, indeed, the reference implementation in the test, as far as I remember, uses floating-point arithmetics, i.e. it's not bit-exact. I think, you need to create a bit-exact fixed-point reference implementation in the test and compare OpenCV function against it

vpisarev:
:+1:

PR #8009, opened 2017-01-14
CMake: Building Dynamic Framework on iOS

Legoless:
Resolves #6186

### This pullrequest changes

the python script and cmake scripts. It adds additional functionality to `build_framework.py` script to build a dynamic framework for iOS, for compatibility with Swift or Objective-C projects on iOS 8+.

Currently the script builds a static library and packages it up into a Framework bundle as defined in [Anatomy of Framework Bundles](https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/FrameworkAnatomy.html). As static libraries are fairly unsupported on iOS and cause multiple linking issues, especially when it comes to CocoaPods dependencies, a dynamic framework should be used instead.

Pull requests adds two parameters to the building script:
- `--dynamic` will build a Cocoa Touch dynamic framework target instead of static library packaged in a framework bundle with minimum deployment target of iOS 8.0.0.
- `--disable-bitcode` will disable bitcode during the build for either static or dynamic framework.

The default functionality of the script remains the same, so running will still generate the same static framework bundle:
`python opencv/platforms/ios/build_framework.py ios`

To build a dynamic framework, `--dynamic` parameter must be added:
`python opencv/platforms/ios/build_framework.py ios --dynamic`

In addition a new CocoaPod [podspec](https://github.com/Legoless/OpenCV-Dynamic) was prepared that can be used as a dependency.

`pod 'OpenCV-Dynamic'`

PR #10923, opened 2018-02-22
Fixed dynamic build script for iOS

benjohnde:
### Used environment
```
cmake version 3.10.2
opencv HEAD (5caf624), 3.3.0 and 3.4.0
```

### Encountered problems

1. App could not be uploaded via ItunesConnect as `UIDeviceFamily` was missing in the dynamic framework of opencv:

As described in #8009, there was a warning:

> warning: User supplied UIDeviceFamily key in the Info.plist will be overwritten. Please use the build setting TARGETED_DEVICE_FAMILY and remove UIDeviceFamily from your Info.plist.

Nonetheless, in the final `Info.plist` the entries regarding `UIDeviceFamily` were missing. Was reproducible on every Mac we tested that on.

2. The dynamic framework was built without bitcode being enabled:

It seems that at the current point of time we need to use `BITCODE_GENERATION_MODE=bitcode`. The only resource I found is https://stackoverflow.com/a/34965178. Sadly, no official documentation by Apple.

### This pullrequest changes

1. Readded UIDeviceFamily (was removed in beec247).
2. Added `BITCODE_GENERATION_MODE=bitcode` flag.

### DIFF
--- a/platforms/ios/Info.Dynamic.plist.in
+++ b/platforms/ios/Info.Dynamic.plist.in
@@ -24,5 +24,10 @@
     </array>
     <key>MinimumOSVersion</key>
     <string>8.0</string>
+    <key>UIDeviceFamily</key>
+    <array>
+        <integer>1</integer>
+        <integer>2</integer>
+    </array>
 </dict>
 </plist>
--- a/platforms/ios/build_framework.py
+++ b/platforms/ios/build_framework.py
@@ -150,6 +150,7 @@ class Builder:
             buildcmd += [
                 "IPHONEOS_DEPLOYMENT_TARGET=8.0",
                 "ONLY_ACTIVE_ARCH=NO",
+                "BITCODE_GENERATION_MODE=bitcode"
             ]
 
             for arch in archs:
--- a/platforms/ios/build_framework.py
+++ b/platforms/ios/build_framework.py
@@ -149,9 +149,11 @@ class Builder:
         if self.dynamic:
             buildcmd += [
                 "IPHONEOS_DEPLOYMENT_TARGET=8.0",
-                "ONLY_ACTIVE_ARCH=NO",
-                "BITCODE_GENERATION_MODE=bitcode"
+                "ONLY_ACTIVE_ARCH=NO"
             ]
+
+            if self.bitcodedisabled == False:
+                buildcmd.append("BITCODE_GENERATION_MODE=bitcode")
 
             for arch in archs:
                 buildcmd.append("-arch")
--- a/platforms/ios/build_framework.py
+++ b/platforms/ios/build_framework.py
@@ -149,10 +149,10 @@ class Builder:
         if self.dynamic:
             buildcmd += [
                 "IPHONEOS_DEPLOYMENT_TARGET=8.0",
-                "ONLY_ACTIVE_ARCH=NO"
+                "ONLY_ACTIVE_ARCH=NO",
             ]
 
-            if self.bitcodedisabled == False:
+            if not self.bitcodedisabled:
                 buildcmd.append("BITCODE_GENERATION_MODE=bitcode")
 
             for arch in archs:


benjohnde:
In addition: of course I recognized [build_framework.py#L102](https://github.com/Preventis/opencv/blob/01b60ab321d9ef7be5694640e17233afdf12e9c8/platforms/ios/build_framework.py#L102) and [build_framework.py#L103](https://github.com/Preventis/opencv/blob/01b60ab321d9ef7be5694640e17233afdf12e9c8/platforms/ios/build_framework.py#L103), same with [OpenCVUtils.cmake#L1234](https://github.com/opencv/opencv/blob/1dbdb1cfc9e38440c4495af08b08c8cd3ba7a367/cmake/OpenCVUtils.cmake#L1234).

benjohnde:
Sry missed the `self.bitcodedisabled` option. Was a bit in my own world as we never built without bitcode 🤔.

alalek:
@Legoless Could you take a look on this patch as experienced Apple user?

Legoless:
I think there was a reason I removed the `UIDeviceFamily` list, but cannot remember exactly why. As for the `BITCODE_GENERATION_MODE=bitcode` User Defined Setting is effectively the same as `-fembed-bitcode` flag. To be quite honest, I am not sure any of those changes are needed. I am currently deploying an app to iTunes Connect with the framework built by original script without any problems or warnings. Are you sure the problems you are having lie within OpenCV framework?

It is true that bitcode documentation is hard to find (and understand), but rather because most people are looking for it in the wrong places. It's not as much an 'Apple' thing, as it is a LLVM thing: https://llvm.org/docs/BitCodeFormat.html

benjohnde:
There may be another reason for having those problems regarding bitcode, but everytime I built the OpenCV Framework with the following lines (here for instance version 3.3.0):

```
git clone git@github.com:opencv/opencv.git
cd opencv
git checkout 3.3.0
python platforms/ios/build_framework.py ios \
  --dynamic \
  --without imgcodec --without videoio --without video --without calib3d --without features2d \
  --without objdetect --without dnn --without ml --without flann --without photo --without stitching \
  --without cudaarithm --without cudabgsegm --without cudacodec --without cudafeatures2d \
  --without cudafilters --without cudaimgproc --without cudalegacy --without cudaobjdetect \
  --without cudaoptflow --without cudastereo --without cudawarping --without cudev --without shape \
--without superres --without videostab --without viz
```

The bitcode flags are missing in the actual framework, you can easily inspect it via `otool -arch armv7 -l opencv2  | grep __LLVM`. I will try to reproduce it again and upload the error screen from Xcode.

In addition, I was not able to upload the app also because `UIDeviceFamily` was missing. Everytime we built the framework, we were forced to adjust the build script. I guess you removed those lines from the `Info.plist` because of the following warning:

> warning: User supplied UIDeviceFamily key in the Info.plist will be overwritten. Please use the build setting TARGETED_DEVICE_FAMILY and remove UIDeviceFamily from your Info.plist.

But can easily inspect the built framework: those lines are in fact missing.

Are you using the pre-built library from the OpenCV releases? Are you building the framework with the `dynamic` flag?

Legoless:
As far as I know the command you're using to check for bitcode is wrong. You should be looking for `segname __LLVM` using `otool`, not `bitcode`. See this link:
https://forums.developer.apple.com/thread/19775

I am using the dynamically built CocoaPods version - can't use the static one as a dynamic framework anyway.

benjohnde:
Well, I read about that here: https://stackoverflow.com/a/33105733. I guess the best solution would be to use `ebcutil`. I am going to clarify that tomorrow morning. I remember something like:

> " [...] was built without full bitcode [...]"

I am testing a bit at home right now but will give you a full-fledged report tomorrow.

benjohnde:
Hmmm, you are right, `otool -arch armv7 -l opencv2 | grep LLVM` seems to do the trick.

Anyway, I built the framework twice, first with my changes, second without. Both seems to have those segnames in it, but:

```
bjohn@Bens-MacBook-Pro ~/Desktop/opencv2.framework.with % ls -lah
total 158408
drwxr-xr-x   5 bjohn  staff   160B Feb 28 20:58 .
drwx------+  7 bjohn  staff   224B Feb 28 21:13 ..
drwxr-xr-x  14 bjohn  staff   448B Feb 28 20:52 Headers
-rw-r--r--   1 bjohn  staff   956B Feb 28 20:58 Info.plist
-rwxr-xr-x   1 bjohn  staff    77M Feb 28 20:58 opencv2
bjohn@Bens-MacBook-Pro ~/Desktop/opencv2.framework.without % ls -lah
total 70984
drwxr-xr-x   5 bjohn  staff   160B Feb 28 21:09 .
drwx------+  7 bjohn  staff   224B Feb 28 21:13 ..
drwxr-xr-x  14 bjohn  staff   448B Feb 28 21:05 Headers
-rw-r--r--   1 bjohn  staff   843B Feb 28 21:09 Info.plist
-rwxr-xr-x   1 bjohn  staff    35M Feb 28 21:09 opencv2
```

Both, exactly same build command was used. I find it a bit odd, that I resulted in twice the size with my variant, but what I definitely shows: my addition changes at least something in the build process.

IIRC it does make sense that the library has an increased size, as we are shipping more information to Apple. It is then decreased after the app thinning process.

I just found the answer on stackoverflow which led me to add the `BITCODE_GENERATION_MODE=bitcode` parameter to the build chain: https://stackoverflow.com/a/46122736.

benjohnde:
I am going to reproduce this at work tomorrow (archiving and submitting the app with the old variant) to give you the exact error message from Apple.

I don't know, maybe we are doing something strangely different from all the other users? I have never read about this problem before and I was kind of suprised that it still exists (at least on our side).

I wrote a gist for that https://gist.github.com/benjohnde/7155910ec3aa3dfec34d12461ab1d822 and I was motivated to do this PR as I received a notification about a fork. So at least someone had the same issue (well I guess it).

I drew the conclusion that most of the people are using non-dynamic frameworks and for that just picking the one which is attached to the release (which also worked for us in the first place). While trying to reduce app size, as we are using only a small subset, this error arose.

benjohnde:
Okay, reconstructing the first problem is really really easy. Just build the framework yourself with the dynamic flag and try to archive the app:

```
ld: bitcode bundle could not be generated because '/Users/bjohn/git/ios/Carthage/Build/iOS/opencv2.framework/opencv2' was built without full bitcode. All frameworks and dylibs for bitcode must be generated from Xcode Archive or Install build file '/Users/bjohn/git/ios/Carthage/Build/iOS/opencv2.framework/opencv2' for architecture arm64
```

As said, at least for us, the `BITCODE_GENERATION_MODE=bitcode` flag for dynamic mode is necessary.

benjohnde:
Okay @Legoless I did a bit of research. `UIDeviceFamily` is missing in the final `Info.plist`. Your change back then (beec247) was fully correct! The generated `Info.plist` in:

- `opencv/ios/build/build-iphonesimulator/lib/Release/opencv2.framework/Info.plist` and
- `opencv/ios/build/build-iphoneos/lib/Release/opencv2.framework/Info.plist`

does contain `UIDeviceFamily` due to `XCODE_ATTRIBUTE_TARGETED_DEVICE_FAMILY`.

But, it gets lost as we are not merging those generated files with the provided ones: [OpenCVGenInfoPlist.cmake#L6](https://github.com/opencv/opencv/blob/3.4.0/cmake/OpenCVGenInfoPlist.cmake#L6).

benjohnde:
The `UIDeviceFamily` was also reproducible but took me a while.

Uploading the IPA to the App Store works as you expected it @Legoless. My bad, the bug occurs if the `Info.plist` does not contain the `UIDeviceFamily` part and you want to export a build for development distribution, for instance using `Organizer` within Xcode. You will then see the following error:

grafik

If you deep dive further into this problem, you will see that this error occurs because parsing the `Info.plist` fails with the exception "No applicable devices found.". You could also adjust the ipatool here `/Applications/Xcode.app/Contents/Developer/usr/bin/ipatool#1106`.

Maybe this is an error of the ipatool in the first place. Nonetheless, the `UIDeviceFamily` part is missing and should be included in the final framework.

I suggest to use my proposed changes. Second option would be to drop both custom plist files and just use the generated ones.

benjohnde:
I just corrected the syntax to be more Pythonic. If you want to drop both custom plist files and use the generated one, I could fix and test that as well @Legoless @alalek. I would be pleased to do so. Just using the generated one for `iphoneos` is sufficient. Looking forward to hear from you guys.

PR #10928, opened 2018-02-23
Fix improper return code in case of timeout

### LINKED ISSUE #10927: cap_v4l Incorrect behavior on timeout
- OpenCV => 3.1
- Compiler => GCC

##### Detailed description

I use OpenCV to grab frames from See3CAM10_CUG camera in trigger mode. In this mode frames are captured only when external trigger is raised. So, time interval between frames can be large than 10s, the timeout specified in cap_v4l.cpp. The problem is that in case of timeout occurs return value of     cv2.VideoCapture.read() function is True. So it is not possible to determine is the frame really captured or  timeout occurs.

##### Steps to reproduce
Using python
```
cap = cv2.VideoCapture(device_id)
ret, frame = cap.read()
```

ret == True both in the case when new frame is grabbed properly and in the case when timeout occurs.

Mehanik:
resolves #10927

### This pullrequest changes

mainloop_v4l2 function returns -1 in case of timeout

### DIFF
--- a/modules/videoio/src/cap_v4l.cpp
+++ b/modules/videoio/src/cap_v4l.cpp
@@ -911,7 +911,7 @@ static int mainloop_v4l2(CvCaptureCAM_V4L* capture) {
             if(returnCode == -1)
                 return -1;
             if(returnCode == 1)
-                break;
+                return 1;
         }
     }
     return 0;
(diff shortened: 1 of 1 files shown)


Mehanik:
@alalek please check new solution.

> mainloop_v4l2() still able to return -1, 0, 1 values

This is not correct, it returns 1 in case of error and 0 both in case of timeout and correct capture. I changed return code to 1 in case when frame is captured correctly and handle it in mainloop_v4l2()'s caller function.

Issue #10658, opened 2018-01-21
Will opencv support C compiler after 3.4.0?


springkim:
- OpenCV => 3.4
- Operating System / Platform => Windows 10 64 Bit
- Compiler => Visual Studio 2015 , MinGW64(7.2.0)

Hi. 
I have a simple question.
opencv3.3.1 not support C compiler(e.g MinGW64, MSVC TC). 
But now i can compile without problem with 3.4.0.
Will opencv support C interface which means support C compiler after 3.4.0?

alalek:
#6585

Issue #10405, opened 2017-12-23
Compiling failiure using C APIs in C program caused by cvdef.h


qingswu:
I am compiling a C program using opencv's C API, while the problem is caused by
```.cpp
namespace cv { namespace debug_build_guard { } using namespace debug_build_guard; }
```
in cvdef.h introduced in revision f448d75

Some guard is required for C code.

##### System information (version)

- OpenCV => :grey_question:
- Operating System / Platform => :grey_question:
- Compiler => :grey_question:

alalek:
> C API

It is not supported anymore, OpenCV is C++ library with classes / exception / etc: #10246

> debug_build_guard

see #9161

Issue #6585, opened 2016-05-25
Opencv3 compilation issue with C API


jlguardi:
### Please state the information for your system
- OpenCV version: 3.1 and also dev version
- Host OS: Linux (Ubuntu 16.04-64 bits)
- Compiler & CMake: GCC 5.3.1 & CMake 3.5.1
### In which part of the OpenCV library you got the issue?

C-API
### Expected behaviour

Correct compilation
### Actual behaviour

Compilation error:

```
/tmp/ccDODlVK.o: In function `cvPointFrom32f':
main.c:(.text+0x54d): undefined reference to `cvRound'
main.c:(.text+0x563): undefined reference to `cvRound'
/tmp/ccDODlVK.o: In function `cvReadInt':
main.c:(.text+0xe68): undefined reference to `cvRound'
/tmp/ccDODlVK.o: In function `cvEllipseBox':
main.c:(.text+0x10d4): undefined reference to `cvRound'
main.c:(.text+0x10f6): undefined reference to `cvRound'
collect2: error: ld returned 1 exit status
```
### Code example to reproduce the issue / Steps to reproduce the issue

Please try to give a full example which will compile as is.

```
#include "opencv/cv.h"
int main()
{
    return 0;
}
```

Compilation line:

```
gcc -Wall main.c -o main -I/usr/local/include/opencv -I/usr/local/include -L/usr/local/lib  -lopencv_imgproc -lopencv_core -lm
```

alalek:
OpenCV is a C++ library now and actively uses C++ features, like exceptions and so on.
Direct support for C compilers has been dropped some years ago, so you should not create C applications using OpenCV.
"C-API" is not supported and should be removed totally (but we have a lack of resources to port this legacy C code to C++, so some of this code still exists right now). Also huge part of fresh OpenCV functionality is missing in "C-API".

There is no plan to fix this in OpenCV directly.

If you want to use OpenCV from C application code you need to find some implementation of C bindings which wraps OpenCV C++ functionality (al least these bindings should catch exceptions and return exit codes for user application).

daavve:
I had the same problem trying to use someone else code from OpenCV 2.4.

I ended up just installing OpenCV 2.4, and using that.

OliverBurgert:
This issue should be reflected in the official opencv website which says (2018-01-21) " It has C++, C, Python and Java interfaces"...

Issue #6221, opened 2016-03-08
OpenCV 3.1 x64 -- Incorrect code is generated when calling cvGetSize from a x64 C library.


rovarma:
### Please state the information for your system
- OpenCV version: 3.1
- Host OS: Windows 8
- Compiler & CMake: MSVC 2015 x64 & CMake 3.5

### In which part of the OpenCV library you got the issue?
- core

### Description
When using OpenCV 3.1 from a x64 C library, incorrect code is generated when calling `cvGetSize`. I have only seen this when compiling for x64 (I only have x64 projects), but I don't know whether the bug also appears when building for x86.

The reason for this bug is the fact that the `CvSize` struct is defined as follows:

```
typedef struct CvSize
{
    int width;
    int height;

#ifdef __cplusplus
    CvSize(int w = 0, int h = 0): width(w), height(h) {}
    template<typename _Tp>
    CvSize(const cv::Size_<_Tp>& sz): width(cv::saturate_cast<int>(sz.width)), height(cv::saturate_cast<int>(sz.height)) {}
    template<typename _Tp>
    operator cv::Size_<_Tp>() const { return cv::Size_<_Tp>(cv::saturate_cast<_Tp>(width), cv::saturate_cast<_Tp>(height)); }
#endif
} CvSize;
```

Since OpenCV itself is compiled as C++ (not as C), this means that `CvSize` has constructors (`#ifdef __cplusplus` == true), which makes it a user-defined type. This in turn leads the MSVC compiler to assume that functions returning this struct by value cannot return it in a register, which causes the following code to be generated for `cvGetSize` (note: this is just the function prologue):

```
// Returns the size of CvMat or IplImage
CV_IMPL CvSize
cvGetSize( const CvArr* arr )
{
00007FF763B006F0  mov         qword ptr [rsp+10h],rdx  
00007FF763B006F5  mov         qword ptr [rsp+8],rcx  
00007FF763B006FA  push        rdi  
00007FF763B006FB  sub         rsp,70h  
```

Note that up until this point, everything is still correct. `cvGetSize` is a function taking two arguments: the address where the return value of `cvGetSize` should be stored is passed in `rcx`, the `arr` argument is passed in `rdx`.

[...]

PR #10941, opened 2018-02-26
C API support for OpenCV 4.3.x

yoshimoto:
I wrote small patches to add C API support to 4.3.1

This will solve the following issues:
#10405, #6585,  #10658

Could you please consider to merge these fixes? 
There are many legacy program that still use C APIs.

### DIFF
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -480,7 +480,7 @@ using std::uint32_t;
 using std::int64_t;
 using std::uint64_t;
 }
-#else
+#elif __cplusplus
 #include <stdint.h>
 namespace cv {
 typedef ::int8_t int8_t;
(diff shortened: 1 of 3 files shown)


alalek:
> legacy program

These programs don't want OpenCV 3.x, because all new algorithms are written in C++.

List of problems is much bigger - there is one more C API issue: #6221 (It is better to break builds instead of wasting time during debugging of issues like this).

yoshimoto:
OK, I see. But, please understand there are some people who want use both of newer and legacy programs at the same time. 

Anyway, if C API is no longer supported by OpenCV 3.x, could you please announce that C API is deprecated, and update the documentations?

For example, https://opencv.org/ says "It has C++, C, Python and Java interfaces".
The API reference still mentions about C APIs. These inconsistencies will cause the confusions, I think.

And if possible, I think it better to add the following check into OpenCV's header files.
#ifndef __cplusplus
# error "C API is no longer supported in OpenCV 3.x. See .... for details."
#endif
I think this would save much time of us. ;-)

pbh5956:
So, do we have now a working C API or not ?

PR #10940, opened 2018-02-26
Fuse batch normalization TensorFlow subgraphs in runtime

dkurt:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/445

### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/10870
resolves https://github.com/opencv/opencv/issues/9894
resolves https://github.com/opencv/opencv/issues/11025

* Merge batch normalization and flatten TensorFlow subgraphs in runtime. Before model's import it fuses sets of nodes into a single alternative layer is known by importer.
* Batch normalization layer with no gamma parameter (no channel-wise multipliers)
* Squeeze operation for 4D blobs.

```
allow_multiple_commits=1
```

PR #10951, opened 2018-02-27
Import tf.reduce_mean operation as average pooling layer

dkurt:
### This pullrequest changes

A specific case of `tf.reduce_mean` layer from TensorFlow that can be interpret as an average pooling layer over all spatial dimensions (global averge pooling).

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/446

**WIP**: please do not merge


dkurt:
See https://github.com/opencv/opencv/pull/10940

PR #10959, opened 2018-02-28
cmake: add option to enable/disable OpenCL in DNN module

alalek:
- OpenCL is turned ON by default
- can be turned off via `cmake -DOPENCV_DNN_OPENCL=OFF <...>`

[Validated](http://pullrequest.opencv.org/buildbot/builders/precommit_opencl_linux/builds/14869/steps/test_dnn-ippicv-opencl/logs/stdio).

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/CMakeLists.txt
+++ b/modules/dnn/CMakeLists.txt
@@ -11,6 +11,14 @@ set(the_description "Deep neural network module. It allows to load models from d
 ocv_add_dispatched_file_force_all("layers/layers_common" AVX AVX2 AVX512_SKX)
 
 ocv_add_module(dnn opencv_core opencv_imgproc WRAP python matlab java js)
+
+ocv_option(OPENCV_DNN_OPENCL "Build with OpenCL support" HAVE_OPENCL)
+if(OPENCV_DNN_OPENCL AND HAVE_OPENCL)
+  add_definitions(-DCV_OCL4DNN=1)
+else()
+  ocv_cmake_hook_append(INIT_MODULE_SOURCES_opencv_dnn "${CMAKE_CURRENT_LIST_DIR}/cmake/hooks/INIT_MODULE_SOURCES_opencv_dnn.cmake")
+endif()
+
 ocv_warnings_disable(CMAKE_CXX_FLAGS -Wno-shadow -Wno-parentheses -Wmaybe-uninitialized -Wsign-promo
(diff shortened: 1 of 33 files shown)


modules/dnn/CMakeLists.txt
```diff
@@ -11,6 +11,14 @@ set(the_description "Deep neural network module. It allows to load models from d
 ocv_add_dispatched_file_force_all("layers/layers_common" AVX AVX2 AVX512_SKX)
 
 ocv_add_module(dnn opencv_core opencv_imgproc WRAP python matlab java js)
+
+ocv_option(OPENCV_DNN_OPENCL "Build with OpenCL support" HAVE_OPENCL)
+if(OPENCV_DNN_OPENCL AND HAVE_OPENCL)
+  add_definitions(-DCV_OCL4DNN=1)
+else()
+  ocv_cmake_hook_append(INIT_MODULE_SOURCES_opencv_dnn "${CMAKE_CURRENT_LIST_DIR}/cmake/hooks/INIT_MODULE_SOURCES_opencv_dnn.cmake")
```
mshabunin: I thought we will use hooks as points for external connections, but here this mechanism is used to extend functionality. In my opinion this approach complicates configuration scripts.

alalek: Existed hooks is not easy too: https://github.com/opencv/opencv/blob/3.4.1/cmake/OpenCVModule.cmake#L728-L737

mshabunin: Can't we do it like this?
```.cmake
...
ocv_glob_module_sources(${sources_options} SOURCES ${fw_srcs})
ocv_list_filterout(OPENCV_MODULE_${the_module}_SOURCES "ocl4dnn")
...
```
I understand that in this case hooks will see non-updated list of sources, but same problem can arise in your current implementation, if custom hook will be installed before ours - it will not get filtered list of sources too.

alalek: `OPENCV_MODULE_${the_module}_SOURCES` is global, i.e "cached" CMake variable.

There is no problem with order in hooks - and I don't want to introduce hooks priority yet =)
There is problem what "common" code should not know about details of this or another module. Module may reuse "common" logic/variables, but not vise-versa. Look on this hook as a some kind of "virtual" function or module interface.

alalek:
:+1:

PR #9293, opened 2017-08-02
core: add multi-argument CV_Assert

sovrasov:

PR #10978, opened 2018-03-03
core: add "check" macros

alalek:
Proposal basics only without massive code changes.

<cut>

<details>

<summary>Proposal details</summary>

* Name: Check macros *for passed parameters validation or internal checks*
* Status: **Draft**
* Platforms: **All**
* Complexity: N/A

## Introduction and Rationale

Current error reporting in OpenCV should be improved. There are a lot of complex "assertion" conditions, which are hard to understand without debugging (if debugging is possible and problem can be easily reproduced).
Informative error reporting should provide better experience for OpenCV users / developers.

## Proposed solution

Using of simple conditions with dumping of current values of checked parameters would significantly help to understand that wrong is happened.

## Impact on existing code, compatibility

None. Existed interfaces are preserved.

## Possible alternatives

Debugging

## References

- #9293

</details>

#### TODO list:
- `CV_CheckSizeEQ(value1, value2, helper_message)`
- `CV_SimilarMatSizeType(mat1, mat2, helper_message)`
- `CV_Check(v, v_test_condition, helper_message)

### DIFF
--- a/modules/core/include/opencv2/core/check.hpp
+++ b/modules/core/include/opencv2/core/check.hpp
@@ -1,0 +1,135 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef OPENCV_CORE_CHECK_HPP
+#define OPENCV_CORE_CHECK_HPP
+
+#include <opencv2/core/base.hpp>
+
+namespace cv {
+
+/** Returns string of cv::Mat depth value: CV_8U -> "CV_8U" or "<invalid depth>" */
+CV_EXPORTS const char* depthToString(int depth);
+
+/** Returns string of cv::Mat depth value: CV_8UC3 -> "CV_8UC3" or "<invalid type>" */
+CV_EXPORTS const String typeToString(int type);
+
+
+//! @cond IGNORED
+namespace detail {
+
+/** Returns string of cv::Mat depth value: CV_8U -> "CV_8U" or NULL */
+CV_EXPORTS const char* depthToString_(int depth);
+
+/** Returns string of cv::Mat depth value: CV_8UC3 -> "CV_8UC3" or cv::String() */
+CV_EXPORTS const cv::String typeToString_(int type);
+
+enum TestOp {
+  TEST_CUSTOM = 0,
+  TEST_EQ = 1,
+  TEST_NE = 2,
+  TEST_LE = 3,
+  TEST_LT = 4,
+  TEST_GE = 5,
+  TEST_GT = 6,
+  CV__LAST_TEST_OP
+};
+
+struct CheckContext {
+    const char* func;
+    const char* file;
+    int line;
+    enum TestOp testOp;
+    const char* message;
+    const char* p1_str;
+    const char* p2_str;
+};
+
+#ifndef CV__CHECK_FILENAME
+# define CV__CHECK_FILENAME __FILE__
+#endif
+
+#ifndef CV__CHECK_FUNCTION
+# if defined _MSC_VER
+#   define CV__CHECK_FUNCTION __FUNCSIG__
+# elif defined __GNUC__
+#   define CV__CHECK_FUNCTION __PRETTY_FUNCTION__
+# else
+#   define CV__CHECK_FUNCTION "<unknown>"
+# endif
+#endif
+
+#define CV__CHECK_LOCATION_VARNAME(id) CVAUX_CONCAT(CVAUX_CONCAT(__cv_check_, id), __LINE__)
+#define CV__DEFINE_CHECK_CONTEXT(id, message, testOp, p1_str, p2_str) \
+    static const cv::detail::CheckContext CV__CHECK_LOCATION_VARNAME(id) = \
+            { CV__CHECK_FUNCTION, CV__CHECK_FILENAME, __LINE__, testOp, message, p1_str, p2_str }
+
+CV_EXPORTS void CV_NORETURN check_failed_auto(const int v1, const int v2, const CheckContext& ctx);
+CV_EXPORTS void CV_NORETURN check_failed_auto(const float v1, const float v2, const CheckContext& ctx);
+CV_EXPORTS void CV_NORETURN check_failed_auto(const double v1, const double v2, const CheckContext& ctx);
+CV_EXPORTS void CV_NORETURN check_failed_MatDepth(const int v1, const int v2, const CheckContext& ctx);
+CV_EXPORTS void CV_NORETURN check_failed_MatType(const int v1, const int v2, const CheckContext& ctx);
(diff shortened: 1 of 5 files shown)


modules/core/include/opencv2/core/check.hpp
```diff
@@ -0,0 +1,129 @@
...
+namespace cv {
+
+//! @cond IGNORED
+namespace detail {
+
+/** Returns string of cv::Mat depth value: CV_8U -> "CV_8U" or "CV_InvalidDepth" */
+CV_EXPORTS const char* getMatDepth(int depth);
+
+/** Returns string of cv::Mat depth value: CV_8U -> "CV_8U" or NULL */
+CV_EXPORTS const char* getMatDepth_(int depth);
```
vpisarev: do we really need 2 functions?

Probably, the function shall be renamed to "depthToString()".

I would retain only the first function, and let it return `<invalid depth>` instead of `CV_InvalidDepth`, becase the latter is not a valid macro or whatever

modules/core/include/opencv2/core/check.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+/** Returns string of cv::Mat depth value: CV_8UC3 -> "CV_8UC3" or "CV_InvalidType" */
+CV_EXPORTS const cv::String getMatType(int type);
```
vpisarev: drop `cv::`

rename `getMatType` to `typeToString()`?

same question about 2 functions

modules/core/include/opencv2/core/check.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+#define CV__TEST_EQ(v1, v2) ((v1) == (v2))
+#define CV__TEST_NE(v1, v2) ((v1) != (v2))
+#define CV__TEST_LE(v1, v2) ((v1) <= (v2))
+#define CV__TEST_LT(v1, v2) ((v1) < (v2))
+#define CV__TEST_GE(v1, v2) ((v1) >= (v2))
+#define CV__TEST_GT(v1, v2) ((v1) > (v2))
+
+#define CV__CHECK(id, op, type, v1, v2, v1_str, v2_str, msg_str) if(CV__TEST_##op((v1), (v2))) ; else do { \
+    CV__DEFINE_CHECK_CONTEXT(id, msg_str, cv::detail::TEST_ ## op, v1_str, v2_str); \
```
vpisarev: why not hide CheckContext and let the `check_failed_*` functions to form it internally? This way we get rid of unnecessary restrictions, such as just 2 arguments in the check etc. Besides, the `do {} while (0)` trick won't be required because the else branch becomes just one function call

modules/core/include/opencv2/core/check.hpp
```diff
@@ -0,0 +1,129 @@
...
+#define CV_CheckEQ(v1, v2, msg)  CV__CHECK(_, EQ, auto, v1, v2, #v1, #v2, msg)
+#define CV_CheckNE(v1, v2, msg)  CV__CHECK(_, EQ, auto, v1, v2, #v1, #v2, msg)
+#define CV_CheckLE(v1, v2, msg)  CV__CHECK(_, EQ, auto, v1, v2, #v1, #v2, msg)
+#define CV_CheckLT(v1, v2, msg)  CV__CHECK(_, EQ, auto, v1, v2, #v1, #v2, msg)
+#define CV_CheckGE(v1, v2, msg)  CV__CHECK(_, EQ, auto, v1, v2, #v1, #v2, msg)
+#define CV_CheckGT(v1, v2, msg)  CV__CHECK(_, EQ, auto, v1, v2, #v1, #v2, msg)
+
+
+/// Check with additional "decoding" of type values in error message
+#define CV_CheckMatTypeEQ(t1, t2, msg)  CV__CHECK(_, EQ, MatType, t1, t2, #t1, #t2, msg)
```
vpisarev: I would drop `Mat` from the name.

modules/core/include/opencv2/core/check.hpp
```diff
@@ -0,0 +1,129 @@
...
+/// Check with additional "decoding" of depth values in error message
+#define CV_CheckMatDepthEQ(d1, d2, msg)  CV__CHECK(_, EQ, MatDepth, d1, d2, #d1, #d2, msg)
+
+#define CV_CheckMatChannelsEQ(c1, c2, msg)  CV__CHECK(_, EQ, MatChannels, c1, c2, #c1, #c2, msg)
+
+
+/// Example: type == CV_8UC1 || type == CV_8UC3
+#define CV_CheckMatType(t, test_expr, msg)  CV__CHECK_CUSTOM_TEST(_, MatType, t, (test_expr), #t, #test_expr, msg)
```
vpisarev: drop `Mat` from the name?

modules/core/include/opencv2/core/check.hpp
alalek: Renamed. First one function is moved into `cv::` namespace. Second stays in `cv::detail::`.

Error string is changed to `<invalid depth>`.
BTW, I would like to avoid angle brackets in messages, because many OpenCV users just copy-paste such messages into GitHub/other markdown editors "as is".

Example: it is here "<invalid depth>" <== (really here, try to edit this comment)

> Quote example: it is here "<invalid depth>" <== (really here, try to edit this comment)

```
Code block example: it is here "<invalid depth>" <== (really here, try to edit this comment)
```

BTW, Adding extra "spaces" allows to hack markdown editors: "< invalid depth >"

modules/core/include/opencv2/core/check.hpp
alalek: The main reason of adding of `CheckContext` is significantly reduce the size of generated code. This struct is "static const", so we don't need to fill it in in the code directly, compiler places them into "const data" segment.
Instead of passing 5-9 arguments to error reporting functions we need pass 2-3 args only.
The similar way is used in OpenCV tracer (`LocationStaticStorage`).

2 arguments should be enough for the most part of cases. Also it handles custom value checks (second "argument" is used to describe failed condition - instead of the name of the second value).

`do {} while(0)` get fixed (must wrap whole macro).

modules/core/include/opencv2/core/check.hpp
```diff
@@ -0,0 +1,129 @@
...
+    cv::detail::check_failed_ ## type((v), CV__CHECK_LOCATION_VARNAME(id)); \
+} while (0)
+
+} // namespace
+//! @endcond
+
+
+/// Supported values of these types: int, float, double
+#define CV_CheckEQ(v1, v2, msg)  CV__CHECK(_, EQ, auto, v1, v2, #v1, #v2, msg)
+#define CV_CheckNE(v1, v2, msg)  CV__CHECK(_, EQ, auto, v1, v2, #v1, #v2, msg)
```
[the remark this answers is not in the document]
alalek: Thanks! Fixed.
Tests with integers are added.

modules/core/include/opencv2/core/check.hpp
alalek: Dropped `Mat` from macros names

vpisarev:
:+1:

PR #10982, opened 2018-03-04
Include error code description into the message

native-api:
### This pullrequest changes

Adds the meaning of the error code to exception's text description.

Plain error code is not descriptive. This leads to confusion about the meaning of the error, e.g. https://stackoverflow.com/questions/48976002/python-opencv-error-215-in-videowriter-write .


terfendail:
👍

PR #10992, opened 2018-03-05
Parametric OpenCL deep learning tests

### LINKED ISSUE #10970: Wrong output when using DNN_TARGET_OPENCL with blobFromImages
**System information (version)**

- OpenCV => 3.4.1
- Operating System / Platform => Windows 10 (x64)
- Compiler => Visual Studio 2017 (v140)
- iGPU => Intel Iris Plus Graphics 650 (23.20.16.4901)

**Detailed description**

DNN's _net.forward("softmax")_ (using GoogleNet as a test case) produces wrong classification label and unbounded probability (i.e. >1) when _DNN_TARGET_OPENCL_ is paired with _blobFromImages_. I'm uncertain whether or not it is a driver issue as pointed out [here](https://github.com/01org/caffe/issues/63#issuecomment-325092900) because either _DNN_TARGET_OPENCL_ or _blobFromImages_ works fine in solitary (e.g. _DNN_TARGET_OPENCL_ + _blobFromImage_ or _DNN_TARGET_CPU_ + _blobFromImages_). Since Intel's CaffeCL doesn't have a problem with batch processing, it's highly unlikely that driver is the issue.

**Steps to reproduce**

Code:
```
...
//using fine-tuned GoogleNet with 6 classes.
...
net.setPreferableTarget(DNN_TARGET_OPENCL);
inputBlob = blobFromImages(imgs, 1.0f, Size(), Scalar(), false); 
net.setInput(inputBlob, "data"); 
prob = net.forward("softmax");
for (int i = 0; i < imgs.size(); ++i)
{
   int classId;
   double classProb;
   getMaxClass(prob.row(i), &classId, &classProb);
   std::vector<String> classNames = readClassNames();
   std::cout << "---------- Prediction for " << labels[i] << " ----------" << std::endl;
   std::cout << "Best class: #" << classId << " '" << classNames.at(classId) << "'" << std::endl;
   std::cout << "Probability: " << classProb * 100 << "%" << std::endl;
}
...
```

Output:
**//using _DNN_TARGET_OPENCL_** 
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 99.9999%
---------- Prediction for c6.jpg ----------
Best class: #3 'c4'
Probability: 297.147%
---------- Prediction for c6.jpg ----------
Best class: #0 'c1'
Probability: 214.173%
---------- Prediction for c6.jpg ----------
Best class: #3 'c4'
Probability: 246.445%
---------- Prediction for c6.jpg ----------
Best class: #4 'c5'
Probability: 377.985%
---------- Prediction for c6.jpg ----------
Best class: #0 'c1'
Probability: 131.051%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 150.404%
---------- Prediction for c6.jpg ----------
Best class: #2 'c3'
Probability: 177.241%
---------- Prediction for c6.jpg ----------
Best class: #4 'c5'
Probability: 81.1507%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6 '
Probability: 409.997%

**//using _DNN_TARGET_CPU_**
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 99.9998%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 100%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 100%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 100%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 100%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 99.9999%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 100%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 100%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 100%
---------- Prediction for c6.jpg ----------
Best class: #5 'c6'
Probability: 100%

dkurt:
### This pullrequest changes

* Replace `OCL_*` dnn tests to parametric tests.
* Fix SoftMax layer in case of OpenCL target and batch size changing.

resolves #10970

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/test/test_caffe_importer.cpp
+++ b/modules/dnn/test/test_caffe_importer.cpp
@@ -47,6 +47,21 @@
 
 namespace opencv_test { namespace {
 
+CV_ENUM(DNNTarget, DNN_TARGET_CPU, DNN_TARGET_OPENCL)
+static testing::internal::ParamGenerator<DNNTarget> availableBackends()
+{
+    static std::vector<DNNTarget> targets;
+    if (targets.empty())
+    {
+        targets.push_back(DNN_TARGET_CPU);
+#ifdef HAVE_OPENCL
+        if (cv::ocl::useOpenCL())
+            targets.push_back(DNN_TARGET_OPENCL);
+#endif
+    }
+    return testing::ValuesIn(targets);
+}
+
 template<typename TString>
 static std::string _tf(TString filename)
 {
@@ -83,10 +98,10 @@ TEST(Test_Caffe, read_googlenet)
     ASSERT_FALSE(net.empty());
 }
 
-typedef testing::TestWithParam<bool> Reproducibility_AlexNet;
+typedef testing::TestWithParam<tuple<bool, DNNTarget> > Reproducibility_AlexNet;
 TEST_P(Reproducibility_AlexNet, Accuracy)
 {
-    bool readFromMemory = GetParam();
+    bool readFromMemory = get<0>(GetParam());
     Net net;
     {
         const string proto = findDataFile("dnn/bvlc_alexnet.prototxt", false);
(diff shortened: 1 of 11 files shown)


modules/dnn/test/test_caffe_importer.cpp
```diff
@@ -47,6 +47,21 @@
 
 namespace opencv_test { namespace {
 
+CV_ENUM(DNNTarget, DNN_TARGET_CPU, DNN_TARGET_OPENCL)
+static testing::internal::ParamGenerator<DNNTarget> availableBackends()
```
alalek: > DNNTarget
> availableBackends

Looks like there is some mess with target/backend terminology.

modules/dnn/test/test_tf_importer.cpp
```diff
@@ -26,6 +26,21 @@ static std::string _tf(TString filename)
     return (getOpenCVExtraDir() + "/dnn/") + filename;
 }
 
+CV_ENUM(DNNTarget, DNN_TARGET_CPU, DNN_TARGET_OPENCL)
```
alalek: Multiple "CV_ENUM()" declarations under the same namespace::name should be avoided (to avoid cases with different values - linker will pick just one and will not warn about duplicate names because they are templates).
You should use anonymous namespace everywhere to avoid compiler mess with these symbols.
Or move this declaration and generator below into `test_precomp.hpp`.

modules/dnn/test/test_caffe_importer.cpp
dkurt: Thanks you! Replaced to `availableDnnTargets`.

modules/dnn/test/test_tf_importer.cpp
dkurt: Replaced all DNNTarget and DNNBackend from tests into `test_precomp.hpp`.

dkurt:
@alalek, I've modified the logic with OpenCL configs initialization since approval. The first solution was not correct in cases when `HAVE_OPENCL` is true but OpenCL target isn't used. So now we release them in `finalize` methods which is called after network is initialized.

PR #10996, opened 2018-03-05
fix incorrect relative path of opencl_kernels.cpp

adayoegi:
### This pullrequest changes

Fix build problem in branch 2.4 when building world module

OS: Ubuntu 16.04
compiler: g++ 5.4.0

#### Original problem
General cpp file are at \<module\>/src/\*.cpp.
But some opencl files such as opencl_kernels are auto genereated during compilation, they are instead at build/modules/\<module\>/\*.cpp
The problem occurs as their relative path are resolved incorrectly, leading to file not found error during make


alalek:
"opencv_world" doesn't work properly in 2.4 branch (and probably would not be fixed). This module should stay disabled by default.
Also development of experimental "ocl" module in 2.4 is abandoned since 2014.

adayoegi:
Did not know this. 
I will close this PR since it is only about building module world.

PR #10997, opened 2018-03-06
Optimize decolor in photo

shimat:
### This pullrequest changes
- Replaced `vector.push_back` into `vector.resize()` and `vector[i] = foo;` to improve performance.  In my environment it is accelerated by 30 to 50%.
- Removed redundant variable initialization

### DIFF
--- a/modules/photo/src/contrast_preserve.cpp
+++ b/modules/photo/src/contrast_preserve.cpp
@@ -52,6 +52,8 @@ using namespace cv;
 
 void cv::decolor(InputArray _src, OutputArray _dst, OutputArray _color_boost)
 {
+    CV_INSTRUMENT_REGION()
+
     Mat I = _src.getMat();
     _dst.create(I.size(), CV_8UC1);
     Mat dst = _dst.getMat();
--- a/modules/photo/src/contrast_preserve.cpp
+++ b/modules/photo/src/contrast_preserve.cpp
@@ -52,8 +52,6 @@ using namespace cv;
 
 void cv::decolor(InputArray _src, OutputArray _dst, OutputArray _color_boost)
 {
-    CV_INSTRUMENT_REGION()
-
     Mat I = _src.getMat();
     _dst.create(I.size(), CV_8UC1);
     Mat dst = _dst.getMat();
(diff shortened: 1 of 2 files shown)


vpisarev:
@shimat, thank you for the patch! It looks like besides the optimization itself you largely refactored the code including many stylistic and indentation changes. Could you please create a much smaller patch with only the optimization? And then add another patch with essential refactoring changes (no reindentation etc.)

modules/photo/src/contrast_preserve.hpp
```diff
@@ -81,350 +83,331 @@ int round_num(double a)
...
 {
-    int h=img.size().height;
-    int w=img.size().width;
+    const int h = img.rows;
+    const int w = img.cols;
 
-    vector <Mat> rgb_channel;
-    split(img,rgb_channel);
+    Mat rgb_channel[3];
+    split(img, rgb_channel);
```
alalek: > Mat rgb_channel[3];

This change is dangerous. There is no checks number of channels in this function.
RGBA/BGRA image will corrupt memory.

Also this change has minimal effect on performance.

The same note for other similar changes.

modules/photo/src/contrast_preserve.hpp
```diff
@@ -81,350 +83,331 @@ int round_num(double a)
...
-double Decolor::energyCalcu(vector <double> &Cg, vector < vector <double> > &polyGrad, vector <double> &wei)
+double Decolor::energyCalcu(const vector<double> &Cg, const vector<vector<double> > &polyGrad, vector<double> &wei) const
 {
     vector <double> energy;
     vector <double> temp;
     vector <double> temp1;
 
-    double val = 0.0;
-    for(unsigned int i=0;i< polyGrad[0].size();i++)
+    const vector<double>::size_type size = polyGrad[0].size();
```
alalek: size_t is enough:
```.diff
-vector<double>::size_type
+size_t
```

modules/photo/src/contrast_preserve.hpp
```diff
@@ -81,350 +83,331 @@ int round_num(double a)
...
+    energy.resize(size);
+    temp.resize(size);
+    temp1.resize(size);
+
+    for (unsigned int i = 0; i< polyGrad[0].size(); i++)
```
alalek: > unsigned int i
> .size()

Type should be "size_t" in these loops.

modules/photo/src/contrast_preserve.cpp
```diff
@@ -61,142 +61,138 @@ void cv::decolor(InputArray _src, OutputArray _dst, OutputArray _color_boost)
...
-    I.convertTo(img,CV_32FC3,1.0/255.0);
-
     // Initialization
+    Decolor obj;
     obj.init();
 
-    vector <double> Cg;
-    vector < vector <double> > polyGrad;
-    vector < vector < int > > comb;
+    Mat img = Mat(I.size(), CV_32FC3);
```
alalek: "Mat img;" is enough (without initialization, like changes below).

modules/photo/src/contrast_preserve.hpp
```diff
@@ -81,350 +83,331 @@ int round_num(double a)
...
-    for(unsigned int i=0;i<Rg.size();i++)
+    for (unsigned int i = 0; i<Rg.size(); i++)
     {
-        if(Rg[i] > level)
-            t1.push_back(1.0);
+        if (Rg[i] > level)
+            t1[i] = (1.0);
         else
-            t1.push_back(0.0);
+            t1[i] = (0.0);
```
alalek: Use this form:
```
t1[i] = (Rg[i] > level) ? 1.0 : 0.0;
```

modules/photo/src/contrast_preserve.hpp
```diff
@@ -81,350 +83,331 @@ int round_num(double a)
...
+        alf[i] -= tmp1[i] * tmp2[i] * tmp3[i];
 
-    Rg.clear(); Gg.clear(); Bg.clear();
-    t1.clear(); t2.clear(); t3.clear();
-    tmp1.clear(); tmp2.clear(); tmp3.clear();
+    /*
+    double sum = 0.0;
+    for (unsigned int i = 0; i<alf.size(); i++)
+    sum += abs(alf[i]);
+    sum = 100.0 * sum / alf.size();*/
```
alalek: "Dead code" in comments should be removed.

modules/photo/src/contrast_preserve.hpp
```diff
@@ -81,350 +83,331 @@ int round_num(double a)
...
-    for(int i=0;i<h;i++)
-        for(int j =0;j<w;j++)
-       {
-            if(Gray.at<float>(i,j) < minval)
-                minval = Gray.at<float>(i,j);
+    for (int i = 0; i<h; i++)
+        for (int j = 0; j<w; j++)
+        {
+            if (Gray.at<float>(i, j) < minval)
+                minval = Gray.at<float>(i, j);
```
alalek: It is shorter:
```
float v = Gray.at<float>(i, j);
minval = std::min(minval, v);
maxval = std::max(maxval, v);
```

Or better to use well optimized [`cv::minmaxLoc()`](https://docs.opencv.org/master/d2/de8/group__core__array.html#gab473bf2eb6d14ff97e89b355dac20707) instead of these loops.

shimat:
@vpisarev OK. I have removed refactoring diffs. I will create another patch for refactoring.

modules/photo/src/contrast_preserve.hpp
shimat: Thanks!  I will try this change in the next refactoring patch.

modules/photo/src/contrast_preserve.hpp
shimat: 👍   I will try this change in the next refactoring patch.

vpisarev:
looks great! thank you! :+1:

PR #11024, opened 2018-03-06
[tutorial_py_contours_more_function] Fixes python3 error int object is not iterable

sergiomb2:
AFAIK cnt[s] returns x,y  when cnt[s][0] is only x of contours , we need the tuple of cnt[s] and tuple (x) retur an error: 

    'int' object is not iterable

### DIFF
--- a/doc/py_tutorials/py_imgproc/py_contours/py_contours_more_functions/py_contours_more_functions.markdown
+++ b/doc/py_tutorials/py_imgproc/py_contours/py_contours_more_functions/py_contours_more_functions.markdown
@@ -47,9 +47,9 @@ defects = cv.convexityDefects(cnt,hull)
 
 for i in range(defects.shape[0]):
     s,e,f,d = defects[i,0]
-    start = tuple(cnt[s][0])
-    end = tuple(cnt[e][0])
-    far = tuple(cnt[f][0])
+    start = tuple(cnt[s])
+    end = tuple(cnt[e])
+    far = tuple(cnt[f])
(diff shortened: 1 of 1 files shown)


terfendail:
`findContours` produce `OutputArrayOfArrays` that maps internally to `std::vector<cv::Mat(n, 1,CV_32SC2)>` to represent `std::vector<std::vector<cv::Point>(n)>` since OpenCV doesn't support 1D matrices. So it became list of 3D `numpy.ndarray`s in python. Therefore one have to use `[i][0]` index to address particular point of the contour.

sergiomb2:
https://github.com/opencv/opencv/blob/master/doc/py_tutorials/py_imgproc/py_contours/py_contours_more_functions/py_contours_more_functions.markdown 

> So it became list of 3D

I think not it return in 2D points (it reads an image)

terfendail:
`findContours` returns `std::vector<cv::Mat(n, 1,CV_32SC2)>` so each contour is actually a matrix of 2D points rather than vector. While `std::vector<cv::Point>` could always be represented as single column matrix `cv::Mat(n, 1,CV_32SC2)` the opposite conversion isn't possible for arbitrary sized matrix. Python binding generator is a generic tool that doesn't know that output matrix will always have single row so it have to represent 2 channel matrix as three dimensional numpy array.

sergiomb2:
OK, the example is here [1] and works. 
I have to review my code , thanks for yours explanations and sorry about the noise . 

[1]
https://docs.opencv.org/3.4.1/d5/d45/tutorial_py_contours_more_functions.html

PR #11029, opened 2018-03-07
Fix PDB files install

db4:
### This pullrequest changes

This PR fixes broken commit https://github.com/opencv/opencv/commit/b06bd501445b027ac498bcb46dea0aa0461adc8e. MSVC .pdb files should now install correctly.

### DIFF
--- a/cmake/OpenCVUtils.cmake
+++ b/cmake/OpenCVUtils.cmake
@@ -953,7 +953,7 @@ function(ocv_install_target)
     set(${__package}_TARGETS "${${__package}_TARGETS}" CACHE INTERNAL "List of ${__package} targets")
   endif()
 
-  if(MSVS)
+  if(MSVC)
     if(NOT INSTALL_IGNORE_PDB AND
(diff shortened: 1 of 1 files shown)


alalek:
What CMake do you use for testing?

alalek:
Modern CMake [has](https://cmake.org/cmake/help/v3.10/manual/cmake-generator-expressions.7.html?highlight=PDB#informational-expressions) `$<TARGET_PDB_FILE:tgt>" - perhaps we should use this instead (there is no problem to install modern CMake version on Windows platforms - so warning would be enough).

db4:
> What CMake do you use for testing?

3.10.2

> Modern CMake has `$<TARGET_PDB_FILE:tgt>" - perhaps we should use this instead (there is no problem to install modern CMake version on Windows platforms - so warning would be enough).

Indeed, I've missed that. Should I rewrite PR to use `$<TARGET_PDB_FILE:tgt> instead?

alalek:
Replaced by #11276

PR #11041, opened 2018-03-09
Fix #10948

### LINKED ISSUE #10948: eigenNonSymmetric never returns
- OpenCV => master
- Compiler => mingw64
##### Steps to reproduce

    // 3x3 should be small enough ...
    Mat src(3,3,CV_32F, 0.0), evec, eval;
    eigenNonSymmetric(src, evec, eval);

this instead returns properly:

    Mat src(3,3,CV_32F, 0.5), evec, eval;
    eigenNonSymmetric(src, evec, eval);

take1014:
#10948 

### This pullrequest changes
Fixed a bug that caused an infinite loop if norm value was zero.

### DIFF
--- a/modules/video/src/bgfg_gaussmix2.cpp
+++ b/modules/video/src/bgfg_gaussmix2.cpp
@@ -849,7 +849,7 @@ void BackgroundSubtractorMOG2Impl::apply(InputArray _image, OutputArray _fgmask,
 #ifdef HAVE_OPENCL
     if (opencl_ON)
     {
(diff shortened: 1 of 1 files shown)


alalek:
Thank you for contribution!

Please remove unnecessary commits from the patch.
It would be nice to have regression test for this.

> if (norm == 0.0)

This condition may not work in some cases (like norm=1e-100). Perhaps we should use `if (fabs(norm) < FLT_EPSILON)` instead.

take1014:
I am sorry, I deleted all the commit history by mistake in the operation.…
Can I pull the request again?

PR #11042, opened 2018-03-09
LSD: support vector<Vec4i> lines

shimat:
### This pullrequest changes
- According to the LineSegmentDetector's comment, supported result types are `vector<Vec4f>` and `vector<Vec4i>` (https://github.com/opencv/opencv/blob/master/modules/imgproc/include/opencv2/imgproc.hpp#L1237). However `drawSegment()` does not support Vec4i lines. This patch resolves the issue.
- I added an test to check that `detect()` can also support Vec4i result.


vpisarev:
thanks! :+1:

PR #11047, opened 2018-03-09
V4L Buffer: Support CV_CAP_PROP_BUFFERSIZE in cap_v4l

### LINKED ISSUE #10671: Can't set Buffersize in opencv 3

- OpenCV => 3.2
-  Python

##### Detailed description

I tried to set the buffersize to minimize the delay in frames (there seems to be a delay of 100 ms in getting any frame) by using the following command in a python code:

cap = cv2.VideoCapture(0)

a=cap.get(cv2.CAP_PROP_BUFFERSIZE) # CV_CAP_PROP_BUFFERSIZE

print 'Buffersize', a

cap.set(cv2.CAP_PROP_BUFFERSIZE,1)

a=cap.get(cv2.CAP_PROP_BUFFERSIZE) # CV_CAP_PROP_BUFFERSIZE

But get the following error-- and then the code ignores the error and moves ahead with the processing. 

```
VIDEOIO ERROR: V4L2: setting property #38 is not supported
```

The main issue is to get rid of the 100 ms delay.

tlanclos:
resolves #10671 

### This pullrequest changes

Adds support for CV_CAP_PROP_BUFFERSIZE in cap_v4l to minimize frame delay


vpisarev:
:+1:

cassianocasagrande:
Hi folks, I'm having problem using 
cv.set(cv.CAP_PROP_BUFFERSIZE, nbuffers).
I'm getting error:
VIDEOIO ERROR: V4L: Property <unknown property string>(38) not supported by device
I'm not that familiar with V4L, and the OpenCV VideoCapture implementation, but I'll dig into the code to see if I find out what is happening.

mxmlnkn:
@cassianocasagrande 
This merge request fixes this. You'll have to use a version which has this change. For me, the error disappeared after upgrading from 3.4.0 to 4.1.1. I'm not entirely sure whether it works correctly though. It still looks like it's lagging 3 frames behind even though I set it to 1 for my webcam input.

PR #11049, opened 2018-03-10
Resolves #10948 again

### LINKED ISSUE #10948: eigenNonSymmetric never returns

take1014:
resolves #10948 

Fixed a bug that caused an infinite loop if norm value was zero.

I made a mistake in my operation, I deleted the pull request.
>alalek 
>if (norm == 0.0)
>This condition may not work in some cases (like norm=1e-100). 
>Perhaps we should use if (fabs(norm) < FLT_EPSILON) instead.
-> I think so, too. Since norm is an absolute value, I think that it is not necessary to perform abs conversion again.

### DIFF
--- a/modules/core/src/lda.cpp
+++ b/modules/core/src/lda.cpp
@@ -350,6 +350,9 @@ private:
             // Look for single small sub-diagonal element
             int l = n1;
             while (l > low) {
+                if (norm < FLT_EPSILON) {
+                    break;
+                }
                 s = std::abs(H[l - 1][l - 1]) + std::abs(H[l][l]);
                 if (s == 0.0) {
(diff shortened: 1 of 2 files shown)


take1014:
Thank you for fixing the test code. I think this is the best.

Issue #10985, opened 2018-03-05
[dnn] Ability to load FSSD pre trained caffe model with "Interp" layer 


fabito:
I'm trying to load a [FSSD: Feature Fusion Single Shot Multibox Detector](https://arxiv.org/pdf/1712.00960v2.pdf) pre trained [caffe model](https://github.com/lzx1413/CAFFE_SSD/tree/fssd#models). 
But I'm getting the following error:

```
[libprotobuf ERROR /io/opencv/3rdparty/protobuf/src/google/protobuf/text_format.cc:298] Error parsing text-format opencv_caffe.NetParameter: 751:16: Message type "opencv_caffe.LayerParameter" has no field named "interp_param".
OpenCV Error: Unspecified error (FAILED: ReadProtoFromTextFile(param_file, param). Failed to parse NetParameter file: /zoo/caffe-zoo/FSSD_512_COCO/deploy.prototxt) in ReadNetParamsFromTextFileOrDie, file /io/opencv/modules/dnn/src/caffe/caffe_io.cpp, line 1145
Traceback (most recent call last):
  File "vigilante/play.py", line 86, in <module>
    prototxt='/zoo/caffe-zoo/FSSD_512_COCO/deploy.prototxt')
  File "/usr/src/app/vigilante/object_detector.py", line 192, in __init__
    self.net = cv2.dnn.readNetFromCaffe(prototxt, caffemodel)
cv2.error: /io/opencv/modules/dnn/src/caffe/caffe_io.cpp:1145: error: (-2) FAILED: ReadProtoFromTextFile(param_file, param). Failed to parse NetParameter file: /zoo/caffe-zoo/FSSD_512_COCO/deploy.prototxt in function ReadNetParamsFromTextFileOrDie
```

See below the full `deploy.prototxt` for reference:

```
name: "VGG_COCO_SSD_FPN_NO33_WOP512x512_deploy"
input: "data"
input_shape {
  dim: 1
  dim: 3
  dim: 512
  dim: 512
}
layer {
  name: "conv1_1"
  type: "Convolution"
  bottom: "data"
  top: "conv1_1"
  param {
    lr_mult: 1.0
    decay_mult: 1.0
  }
  param {
    lr_mult: 2.0
    decay_mult: 0.0
  }
  convolution_param {
    num_output: 64
    pad: 1
    kernel_size: 3
    weight_filler {
      type: "xavier"
    }
    bias_filler {
      type: "constant"
      value: 0.0
    }
  }
[...]

PR #11050, opened 2018-03-10
Added ResizeBilinear op for tf

dmonterom:
### This pullrequest changes

Added ResizeBilinear op for tensorflow importer.

### DIFF
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -565,7 +565,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
     };
 
     /**
-     * @brief Resize input 4-dimensional blob by nearest neghbor or bilinear strategy.
+     * @brief Resize input 4-dimensional blob by nearest neighbor or bilinear strategy.
      *
      * Layer is used to support TensorFlow's resize_nearest_neighbor and resize_bilinear ops.
      */
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -565,14 +565,14 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
     };
 
     /**
-     * @brief Resize input 4-dimensional blob by nearest neighbor strategy.
+     * @brief Resize input 4-dimensional blob by nearest neghbor or bilinear strategy.
      *
-     * Layer is used to support TensorFlow's resize_nearest_neighbor op.
+     * Layer is used to support TensorFlow's resize_nearest_neighbor and resize_bilinear ops.
      */
-    class CV_EXPORTS ResizeNearestNeighborLayer : public Layer
+    class CV_EXPORTS ResizeLayer : public Layer
     {
     public:
-        static Ptr<ResizeNearestNeighborLayer> create(const LayerParams& params);
+        static Ptr<ResizeLayer> create(const LayerParams& params);
     };
 
     class CV_EXPORTS ProposalLayer : public Layer
--- a/modules/dnn/src/init.cpp
+++ b/modules/dnn/src/init.cpp
@@ -83,7 +83,7 @@ void initializeLayerFactory()
     CV_DNN_REGISTER_LAYER_CLASS(Concat,         ConcatLayer);
     CV_DNN_REGISTER_LAYER_CLASS(Reshape,        ReshapeLayer);
     CV_DNN_REGISTER_LAYER_CLASS(Flatten,        FlattenLayer);
-    CV_DNN_REGISTER_LAYER_CLASS(ResizeNearestNeighbor, ResizeNearestNeighborLayer);
+    CV_DNN_REGISTER_LAYER_CLASS(Resize,         ResizeLayer);
     CV_DNN_REGISTER_LAYER_CLASS(CropAndResize,  CropAndResizeLayer);
 
     CV_DNN_REGISTER_LAYER_CLASS(Convolution,    ConvolutionLayer);
--- a/modules/dnn/src/layers/resize_layer.cpp
+++ b/modules/dnn/src/layers/resize_layer.cpp
@@ -11,10 +11,10 @@
 
 namespace cv { namespace dnn {
 
-class ResizeNearestNeighborLayerImpl CV_FINAL : public ResizeNearestNeighborLayer
+class ResizeLayerImpl CV_FINAL : public ResizeLayer
 {
 public:
-    ResizeNearestNeighborLayerImpl(const LayerParams& params)
+    ResizeLayerImpl(const LayerParams& params)
     {
         setParamsFrom(params);
         CV_Assert(params.has("width") && params.has("height") || params.has("zoom_factor"));
@@ -22,9 +22,10 @@ public:
         outWidth = params.get<float>("width", 0);
         outHeight = params.get<float>("height", 0);
         zoomFactor = params.get<int>("zoom_factor", 1);
+        interpolation = params.get<int>("interpolation");
         alignCorners = params.get<bool>("align_corners", false);
         if (alignCorners)
-            CV_Error(Error::StsNotImplemented, "Nearest neighborhood resize with align_corners=true is not implemented");
+            CV_Error(Error::StsNotImplemented, "Resize with align_corners=true is not implemented");
     }
 
     bool getMemoryShapes(const std::vector<MatShape> &inputs,
@@ -78,7 +79,7 @@ public:
             for (size_t ch = 0; ch < inputs[0]->size[1]; ++ch)
             {
                 resize(getPlane(inp, n, ch), getPlane(out, n, ch),
-                       Size(outWidth, outHeight), 0, 0, INTER_NEAREST);
+                    Size(outWidth, outHeight), 0, 0, interpolation);
             }
         }
     }
@@ -103,14 +104,14 @@ public:
     }
 
 private:
-    int outWidth, outHeight, zoomFactor;
+    int outWidth, outHeight, zoomFactor, interpolation;
     bool alignCorners;
 };
 
 
-Ptr<ResizeNearestNeighborLayer> ResizeNearestNeighborLayer::create(const LayerParams& params)
+Ptr<ResizeLayer> ResizeLayer::create(const LayerParams& params)
 {
-    return Ptr<ResizeNearestNeighborLayer>(new ResizeNearestNeighborLayerImpl(params));
+    return Ptr<ResizeLayer>(new ResizeLayerImpl(params));
 }
 
 }  // namespace dnn
--- a/modules/dnn/src/layers/resize_layer.cpp
+++ b/modules/dnn/src/layers/resize_layer.cpp
@@ -30,9 +30,8 @@ public:
             zoomFactorWidth = params.get<int>("zoom_factor_x");
             zoomFactorHeight = params.get<int>("zoom_factor_y");
         }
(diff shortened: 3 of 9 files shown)


modules/dnn/include/opencv2/dnn/all_layers.hpp
```diff
@@ -575,6 +575,12 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         static Ptr<ResizeNearestNeighborLayer> create(const LayerParams& params);
     };
 
+    class CV_EXPORTS ResizeBilinearLayer : public Layer
```
dkurt: @dmonterom, what do you think about renaming `ResizeNearestNeighborLayer` to `ResizeLayer` and use interpolation strategy as an additional flag? Please add a test similar to nearest neighbor layer

https://github.com/opencv/opencv/blob/5b868ccd829975da5372bf330994553e176aee09/modules/dnn/test/test_tf_importer.cpp#L325-L328

https://github.com/opencv/opencv_extra/blob/b467369489529d8eb2e847c13b117ec4e253a638/testdata/dnn/tensorflow/generate_tf_models.py#L258-L262

dmonterom: Hi @dkurt, thats a good idea, I will change it as soon as possible. I will also add the new test soon.

dkurt: @dmonterom, thank you! It also can help us to resolve an issue https://github.com/opencv/opencv/issues/10985 with `Interp` layer from some of Caffe's forks smoother.

modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -1469,11 +1469,29 @@ void TFImporter::populateNet(Net dstNet)
 
             layerParams.set("height", outSize.at<int>(0, 0));
             layerParams.set("width", outSize.at<int>(0, 1));
+            layerParams.set("interpolation", 0);
```
dkurt: Enumerations `INTER_NEAREST` and `INTER_LINEAR` are also integers. You can just set them as arguments, `layerParams.set("interpolation", INTER_NEAREST);` and use `interpolation` like `resize(getPlane(inp, n, ch), getPlane(out, n, ch), Size(outWidth, outHeight), 0, 0, interpolation);`

modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -1469,11 +1469,29 @@ void TFImporter::populateNet(Net dstNet)
...
             if (hasLayerAttr(layer, "align_corners"))
                 layerParams.set("align_corners", getLayerAttr(layer, "align_corners").b());
 
-            int id = dstNet.addLayer(name, "ResizeNearestNeighbor", layerParams);
+            int id = dstNet.addLayer(name, "Resize", layerParams);
+            layer_id[name] = id;
+
+            connect(layer_id, dstNet, parsePin(layer.input(0)), id, 0);
+        }
+        else if (type == "ResizeBilinear")
```
dkurt: Body of that case is absolutely the same. Can you just extend the previous one by `else if (type == "ResizeNearestNeighbor" || type == "ResizeBilinear")` and vary "interpolation" flag?

modules/dnn/include/opencv2/dnn/all_layers.hpp
dmonterom: Hey @dkurt, I made the changes (I also tested the two layers and everything seems ok). If you agree with them i will replace the resize nearest neighbor test with a new one.

dkurt: @dmonterom please add `*_net`, `*_in` and `*_out` files generated by script as a PR to opencv_extra with modified version of the script.

modules/dnn/src/init.cpp
```diff
@@ -83,7 +83,7 @@ void initializeLayerFactory()
     CV_DNN_REGISTER_LAYER_CLASS(Concat,         ConcatLayer);
     CV_DNN_REGISTER_LAYER_CLASS(Reshape,        ReshapeLayer);
     CV_DNN_REGISTER_LAYER_CLASS(Flatten,        FlattenLayer);
-    CV_DNN_REGISTER_LAYER_CLASS(ResizeNearestNeighbor, ResizeNearestNeighborLayer);
+    CV_DNN_REGISTER_LAYER_CLASS(Resize, ResizeLayer);
```
dkurt: Please align `ResizeLayer` with the other layers names.

modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -1462,18 +1462,22 @@ void TFImporter::populateNet(Net dstNet)
...
-        else if (type == "ResizeNearestNeighbor")
+        else if (type == "ResizeNearestNeighbor" || type == "ResizeBilinear")
         {
             Mat outSize = getTensorContent(getConstBlob(layer, value_id, 1));
             CV_Assert(outSize.type() == CV_32SC1, outSize.total() == 2);
 
             layerParams.set("height", outSize.at<int>(0, 0));
             layerParams.set("width", outSize.at<int>(0, 1));
+            if (type == "ResizeNearestNeighbor")
+                layerParams.set("interpolation", 0);
```
dkurt: I think `INTER_NEAREST` or `INTER_LINEAR` should be here.

dmonterom: Hi, do you mean replacing the values by them? Then I should define them at the beginning or include imgproc.hpp, right?

dkurt: Why? They are just defined in OpenCV: https://docs.opencv.org/master/da/d54/group__imgproc__transform.html#ga5bb5a1fea74ea38e1a5445ca803ff121 . And please do not forget to add a test.

dmonterom:
Hi @dkurt, I have created a test for resize bilinear using the generator script, but when i run the test it fails:
 ```
error: Expected: (normL1) <= (l1), actual: 0.122411 vs 1e-05
error: Expected: (normInf) <= (lInf), actual: 1.06526 vs 0.0001
```
I have also regenerate the resize nearest one but it successes, so its not about my tensorflow version.
Maybe they are using different bilinear interpolation approaches?

**Comment by dkurt** [comment]
@dmonterom, that is why we need to test it. Perhaps the problem is in borders.Try to figure out what values are different. Test on small inputs like `1x5x5x1` and compare outputs.

**Comment by dmonterom** [comment]
@dkurt, I have been doing several tests with different sizes and they always fail. I found this information:
 [https://hackernoon.com/how-tensorflows-tf-image-resize-stole-60-days-of-my-life-aba5eb093f35](url)
[https://stackoverflow.com/questions/48628736/bilinear-interpolation-implentations-in-tensorflow-and-opencv?rq=1](url)
So it seems that there is a bug in the tensorflow resize op. Anyway, the actual implementation of the layer does the trick: I have imported blitznet with dnn (which has several resize_bilinear layers) and the results are very close to the ones I get with tensorflow.

**Comment by dkurt** [comment]
@dmonterom, it'd nice if you can attach the results of your experiments. We need to know if difference is only at boundaries or in corners or in a whole image. Thanks!

**Comment by dmonterom** [comment]
@dkurt, here is an example of blitznet with tensorflow and with OpenCV
harry-meghan-15
Here is the result with tensorflow:
tensorflow_result
And here with OpenCV:
opencv_result
As you can see, there is a small accuracy drop with OpenCV but the results are quite good.

**Comment by dkurt** [comment]
Thank you but I meant numerical experiments with resize of 5x5 inputs. We need to compare numbers first.

**Comment by dkurt** [comment]
@dmonterom, thank you for your work and research! It seems that TensorFlow's resize_bilinear is a quite different from OpenCV's INTER_LINEAR. Let me introduce you an experimental mechanic of custom layers. All that you need is just use changes from PR https://github.com/opencv/opencv/pull/11129 (work in progress for now). You may find a tutorial there.

File: modules/dnn/src/layers/resize_layer.cpp [hunk 6]
```diff
@@ -0,0 +1,172 @@
...
+            const float widthScale = static_cast<float>(inpWidth) / (outWidth);
+            const int numPlanes = inp.size[0] * inp.size[1];
+            CV_Assert(inp.isContinuous(), out.isContinuous());
+
+            for (int y = 0; y < outHeight; ++y)
+            {
+                float input_y = y * heightScale;
+                int y0 = static_cast<int>(input_y);
+                const float* inpData_row0 = (float*)inp.data + y0 * inpWidth;
+                const float* inpData_row1 = (y0 + 1 < inpHeight) ? (inpData_row0 + inpWidth) : inpData_row0;
```
alalek: Use `.ptr<float>(y0)`
and `.ptr<float>(std::min(y0 + 1, inpHeight - 1))`

modules/dnn/src/layers/resize_layer.cpp
```diff
@@ -0,0 +1,172 @@
...
+                for (int x = 0; x < outWidth; ++x)
+                {
+                    float input_x = x * widthScale;
+                    int x0 = static_cast<int>(input_x);
+                    int x1 = std::min(x0 + 1, inpWidth - 1);
+
+                    float* outData = (float*)out.data + y * outWidth + x;
```
alalek: `.ptr()`

modules/dnn/src/layers/resize_layer.cpp
```diff
@@ -0,0 +1,172 @@
...
+                    const float* inpData_row0_c = inpData_row0;
+                    const float* inpData_row1_c = inpData_row1;
+                    for (int c = 0; c < numPlanes; ++c)
+                    {
+                        *outData = inpData_row0_c[x0] +
+                            (input_y - y0) * (inpData_row1_c[x0] - inpData_row0_c[x0]) +
+                            (input_x - x0) * (inpData_row0_c[x1] - inpData_row0_c[x0] +
+                            (input_y - y0) * (inpData_row1_c[x1] - inpData_row0_c[x1] - inpData_row1_c[x0] + inpData_row0_c[x0]));
```
alalek: Why OpenCV `cv::resize()` with `BORDER_REPLICATE` doesn't work here? (answer: `0.5` pixel center)

Anyway, if you want to reuse OpenCV's `INTER_LINEAR` constant then it should have **the same behavior** as defined by this constant.

modules/dnn/src/layers/resize_layer.cpp
dkurt: Thanks for a review! Replaced it.

modules/dnn/src/layers/resize_layer.cpp
dkurt: @alalek, There are different topics that TenorFlow's resize_bilinear doesn't equal to OpenCV's one.

Replaced constants to string values.

vpisarev: @alalek, if we extend cv::resize in the master branch and add the optional floating-point shift, will that make cv::resize suitable for implementation of this layer?

alalek: Two different formulas are usually used for resize operation:

1. `x * widthScale`
2. Something like `(x + 0.5) * widthScale - 0.5`

The second formula is used in OpenVX, OpenCV's `INTER_LINEAR` resize (and probably generic `INTER_AREA`).
The first is used in OpenCV `INTER_NEAREST` (not accurate, but it is fast) / `INTER_AREA` (fast path).

Here, we need to use the first formula with "linear" interpolation (to match original framework operation).

vpisarev: @alalek, it means that cv::warpAffine can do that?

PR #11057, opened 2018-03-12
calib3d: avoid clang weird behavior

### LINKED ISSUE #11056: calib3d: some tests fails using clang on Raspberry Pi 3
- OpenCV => master ( c729fbd )
- Operating System / Platform => Debian (Raspbian) 9.3 / Raspberry Pi 3
- Compiler => GCC 4.9.2 / Clang 3.7.0

##### Detailed description
- The following tests fails from ```opencv_test_calib3d```
```
[==========] 89 tests from 46 test cases ran. (472539 ms total)
[  PASSED  ] 86 tests.
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] Calib3d_SolveP3P.accuracy
[  FAILED  ] Calib3d_SolvePnPRansac.accuracy
[  FAILED  ] Calib3d_SolvePnP.accuracy

 3 FAILED TESTS
```
- On the other hand, when building with GCC, this doesn't happen
```
[==========] 89 tests from 46 test cases ran. (410682 ms total)
[  PASSED  ] 89 tests.
```
- I reported similar behavior ( #9187 ) but this was based on unstable ```srand```, so it's independent for this case.
- This time, same code, same data, same platform, but different compiler.
- Since we have a reference from gcc, I traced back where the data differs and trace backed to [here](https://github.com/opencv/opencv/blob/c729fbd652b1ebfbb205ac90c303160573f40dd1/modules/calib3d/src/ap3p.cpp#L48) in [ap3p.cpp](https://github.com/opencv/opencv/blob/c729fbd652b1ebfbb205ac90c303160573f40dd1/modules/calib3d/src/ap3p.cpp)
```
complex<double> complex2 = 2 * q4 / sqrt_2m;
```
- I added some ```printf``` and found weird behvior of complex
```
GCC(expected behavior)   |Clang (weird behavior)
=========================|=========================
q4      :-0.00130684     |q4      :-0.00130684    
2 * q4  :(-0.00261369,0) |2 * q4  :(-0.00261369,0)
sqrt_2m :(0.00617747,0)  |sqrt_2m :(0.00617747,0) 
complex2:(-0.4231,0)     |complex2:(-0.00261369,0)
```
- It seems division of ```complex<double>``` has something wrong.
- From this part, the data was different, and concluded as a massive number of failure.
- Just to make sure, I wrote [a small code ](https://gist.github.com/tomoaki0705/6fc8430823d0dba49902ae0c68896717) to check the behaviour.
- Even for the [simplest following code](https://gist.github.com/tomoaki0705/6fc8430823d0dba49902ae0c68896717#file-main-cpp), the result differs between the compiler
```
#include <iostream>
#include <cmath>
#include <complex>

int main(int argc, char**argv)
{
    std::complex<double> a(-0.00261369, 0);
    std::complex<double> b(0.00617747, 0);

    std::complex<double> c = a / b;
    std::cout << "a" << a << std::endl;
    std::cout << "b" << b << std::endl;
    std::cout << "c" << c << std::endl;

    return 0;
}
```
- [result](https://gist.github.com/tomoaki0705/6fc8430823d0dba49902ae0c68896717#file-result-txt)
```
=========GCC=========
a(-0.00261369,0)
b(0.00617747,0)
c(-0.4231,0)
=========Clang=======
a(-0.00261369,0)
b(0.00617747,0)
c(-0.00261369,0)
=====GCC Version=====
g++ (Raspbian 4.9.2-10) 4.9.2
Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

=====Clang Version===
Raspbian clang version 3.7.0-4~bpo8+1 (tags/RELEASE_370/final) (based on LLVM 3.7.0)
Target: arm-unknown-linux-gnueabihf
Thread model: posix
```
- I also tried on wandbox but this issue did not reproduce. I didn't use any suspicious clang, I installed clang using ```apt```
- I added a workaround using ```ifdef``` in this code, and avoided the issue.
- I'll send PR later.

##### Steps to reproduce
- run ```opencv_test_calib3d``` on Raspbian on Raspberry Pi 3, using clang

tomoaki0705:
resolves #11056

### DIFF
--- a/modules/calib3d/src/ap3p.cpp
+++ b/modules/calib3d/src/ap3p.cpp
@@ -45,7 +45,13 @@ void solveQuartic(const double *factors, double *realRoots) {
     complex<double> sqrt_2m = sqrt(static_cast<complex<double> >(-2 * p4 / 3 + t));
     double B_4A = -a3 / (4 * a4);
     double complex1 = 4 * p4 / 3 + t;
+#if defined(__clang__) && defined(__arm__) && __clang_major__ == 3 && __clang_minor__ <= 7 && !defined(__ANDROID__)
+    // details: https://github.com/opencv/opencv/issues/11056
+    complex<double> complex2 = 2 * q4;
+    complex2 = complex<double>(complex2.real() / sqrt_2m.real(), 0);
+#else
     complex<double> complex2 = 2 * q4 / sqrt_2m;
+#endif
     double sqrt_2m_rh = sqrt_2m.real() / 2;
     double sqrt1 = sqrt(-(complex1 + complex2)).real() / 2;
     realRoots[0] = B_4A + sqrt_2m_rh + sqrt1;


modules/calib3d/src/ap3p.cpp
```diff
@@ -45,7 +45,12 @@ void solveQuartic(const double *factors, double *realRoots) {
     complex<double> sqrt_2m = sqrt(static_cast<complex<double> >(-2 * p4 / 3 + t));
     double B_4A = -a3 / (4 * a4);
     double complex1 = 4 * p4 / 3 + t;
+#if defined(__clang__)
+    complex<double> complex2 = 2 * q4;
+    complex2 = complex<double>(complex2.real() / sqrt_2m.real(), 0);
```
alalek: Can sqrt_2m have non-zero imagination? (from sqrt call of negative number)
Perhaps we need full formula for complex numbers division (and optimize it for zero `complex2.imag()`).

tomoaki0705: Thank you for nice comment.
All the test case from gtest doesn't fail to 0, but AFAIK, there's no guarantee on this point.
Let me think of that point.

tomoaki0705: As far as I understood, this function ```solveQuartic``` is calculating real roots from quartic equation.
Mathematically, there is a possibility that imaginary part becomes non-zero, (probably caused by wrong correspondence)
Should I add assertion before the ```sqrt``` or shall I keep as it is ?

tomoaki0705:
@alalek , if you don't mind, can I modify the commit by my self ?

alalek:
Sure.

tomoaki0705:
Thank you !

PR #11060, opened 2018-03-12
Thread-safe version of sparse function in cv::gpu::PyrLKOpticalFlow

juanecito:
Issue: The function sparse of cv::gpu::PyrLKOpticalFlow class isn't thread-safe.

Resolution: This modification permits to call this function from different CPU threads and to use CUDA texture resources in 5 parallel execution lines. Critical section is also managed in this function (TBB intel*). In the future a dense function thread-safe version can be made.

Best regards
Juan María Gómez López

*Fixed: (first version used posix threads)

```
force_builders=Custom
buildworker:Linux x64=linux-1
buildworker:Linux x64 Debug=linux-1
```

### DIFF
--- a/modules/gpu/src/cuda/pyrlk.cu
+++ b/modules/gpu/src/cuda/pyrlk.cu
@@ -60,12 +60,54 @@ namespace pyrlk
     __constant__ int c_halfWin_y;
     __constant__ int c_iters;
 
-    texture<float, cudaTextureType2D, cudaReadModeElementType> tex_If(false, cudaFilterModeLinear, cudaAddressModeClamp);
-    texture<float4, cudaTextureType2D, cudaReadModeElementType> tex_If4(false, cudaFilterModeLinear, cudaAddressModeClamp);
-    texture<uchar, cudaTextureType2D, cudaReadModeElementType> tex_Ib(false, cudaFilterModePoint, cudaAddressModeClamp);
+#define CUDA_CONSTANTS(index) \
+	__constant__ int c_winSize_x##index; \
+	__constant__ int c_winSize_y##index; \
+	__constant__ int c_halfWin_x##index; \
+	__constant__ int c_halfWin_y##index; \
+	__constant__ int c_iters##index;
 
-    texture<float, cudaTextureType2D, cudaReadModeElementType> tex_Jf(false, cudaFilterModeLinear, cudaAddressModeClamp);
-    texture<float4, cudaTextureType2D, cudaReadModeElementType> tex_Jf4(false, cudaFilterModeLinear, cudaAddressModeClamp);
+    CUDA_CONSTANTS(0)
+    CUDA_CONSTANTS(1)
+    CUDA_CONSTANTS(2)
+    CUDA_CONSTANTS(3)
+    CUDA_CONSTANTS(4)
+
+    template <int index> struct c_multi_winSize_x;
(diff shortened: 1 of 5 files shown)


alalek:
2.4 branch is in a deep maintenance mode, only bugfixes are accepted.

> pthread_mutex_t

OpenCV is cross-platform library, so this code breaks builds on Windows.

modules/gpu/src/cuda/pyrlk.cu
```diff
@@ -309,6 +402,198 @@ namespace pyrlk
...
+
+        __syncthreads();
+
+        A11 = smem1[0];
+        A12 = smem2[0];
+        A22 = smem3[0];
+
+        float D = A11 * A22 - A12 * A12;
+
+        if (D < numeric_limits<float>::epsilon())
```
alalek: Such checks should have this form: `(fabs(D) < ...)`

modules/gpu/src/pyrlk.cpp
```diff
@@ -40,6 +40,8 @@
 //
 //M*/
 
+#include <pthread.h>
+
 #include "precomp.hpp"
```
alalek: "precomp.hpp" must be the first include in .cpp file.

modules/gpu/src/pyrlk.cpp
```diff
@@ -98,7 +107,9 @@ namespace
     }
 }
 
-void cv::gpu::PyrLKOpticalFlow::sparse(const GpuMat& prevImg, const GpuMat& nextImg, const GpuMat& prevPts, GpuMat& nextPts, GpuMat& status, GpuMat* err)
+void cv::gpu::PyrLKOpticalFlow::sparse(const GpuMat& prevImg,
+						const GpuMat& nextImg, const GpuMat& prevPts,
+						GpuMat& nextPts, GpuMat& status, GpuMat* err)
```
alalek: Don't use "tabs"

juanecito: Ok, I'm going to change pthreads by c++11 std::mutex and std::condition_variable and add accuracy tests and performance tests

alalek: C++11 is not allowed for OpenCV 2.4/3.x - C++98 is required.

juanecito: I understand, so I can use TBB (tbb::mutex and condition_variable), and make compilable the thread-safe version only if there is TBB support
"#ifdef HAVE_TBB"
What do you think?
Thank you for your help

juanecito:
In second commit i have fixed tabs and pthread, and I have added performance and accuracy test

juanecito:
Hi alalek, should I do anything more to complete this pull-request?, is there anything to fix?

vpisarev:
@juanecito, sorry for delay with reply. As you can see, your pull request does not pass our buildbot, there are compile errors (see the right-most 'Custom' builder).

juanecito:
I think HAVE_TBB macro isn't defined in .cu files (missing in nvcc compiler flags). I've configured cmake  with -D WITH_TBB=ON option but the functions which are defined inside conditional-compilation block HAVE_TBB aren't compilated. I've added a fix including cvconfig.h in pyrlk.cu in another pullrequest (#11534 ). On the other hand i've added an example of thread-safe version of spase function in the same pullrequest (#11534)

PR #11064, opened 2018-03-13
cudaarithm: make the asynchronous call to NPP safe

### LINKED ISSUE #11063: cudaarithm: async call to NPP fails

- OpenCV => master 
- Operating System / Platform => Windows 10 Pro 64bit / CUDA 8.0
- Compiler => Visual Studio 2015 Win64
- Operating System / Platform => Ubuntu 64bit  / Jetson TX1 / CUDA 8.0
- Compiler => GCC 5.4.0

##### Detailed description
- test from ```opencv_test_cudaarithm``` fails
```
[==========] 10629 tests from 65 test cases ran. (34990 ms total)
[  PASSED  ] 10625 tests.
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] CUDA_Arithm/MeanStdDev.Async/0, where GetParam() = (NVIDIA Tegra X1, 128x128, whole matrix)
[  FAILED  ] CUDA_Arithm/MeanStdDev.Async/1, where GetParam() = (NVIDIA Tegra X1, 128x128, sub matrix)
[  FAILED  ] CUDA_Arithm/MeanStdDev.Async/2, where GetParam() = (NVIDIA Tegra X1, 113x113, whole matrix)
[  FAILED  ] CUDA_Arithm/MeanStdDev.Async/3, where GetParam() = (NVIDIA Tegra X1, 113x113, sub matrix)

 4 FAILED TESTS
```
- Also confirmed with Jetson TX2, Geforce GTX 650M, Geforce GTX 1060
- I cut out the minimum code from OpenCV and then the crash didn't reproduce.
- I googled up with the message "Microsoft C++ exception: NppStatus at memory location" which appeared on Visual Studio, and found [this post](https://devtalk.nvidia.com/default/topic/895161/gpu-accelerated-libraries/npp-row-and-column-filters/post/4746520/#4746520) from a nVidia person

> It seems that any time you issue a nppSetStream() call that in fact changes the underlying stream, you will need to issue a cudaDeviceSynchronize() call (first), before issuing the nppSetStream() call.

- This is exactly what is [happening inside OpenCV](https://github.com/opencv/opencv/blob/7e9578789b8f01cabf92d2c57691dbe65d6ee5df/modules/core/include/opencv2/core/private.cuda.hpp#L142)
- I added a macro to call ```cudaStreamSynchronize``` before ```nppSetStream``` and this solved the issue
- References
   - [Same situation still happens on CUDA 8.0](https://devtalk.nvidia.com/default/topic/999910/gpu-accelerated-libraries/npp-stream-crash/)
   - [Same explanation on NPP 9.1](http://docs.nvidia.com/cuda/npp/index.html)
> For this reason it is recommended that cudaDeviceSynchronize (or at least cudaStreamSynchronize) be called before making an nppSetStream call to change to a new stream ID. This will insure that any internal function calls that have not yet occurred will be completed using the current stream ID before it changes to a new ID.
   - [cudaStreamSynchronize](https://www.cs.cmu.edu/afs/cs/academic/class/15668-s11/www/cuda-doc/html/group__CUDART__STREAM_geb3b2f88b7c1cff8b67a998a3a41c179.html)

- I'll send a PR later

##### Steps to reproduce
- run ```opencv_test_cudaarithm --gtest_filter=*MeanStdDev*Async*```

tomoaki0705:
resolves #11063 

```
force_builders=Custom
allow_multiple_commits=1
```

### DIFF
--- a/modules/core/include/opencv2/core/private.cuda.hpp
+++ b/modules/core/include/opencv2/core/private.cuda.hpp
@@ -108,6 +108,8 @@ static inline void throw_no_cuda() { CV_Error(cv::Error::GpuNotSupported, "The l
 
 #else // HAVE_CUDA
 
+#define nppSafeSetStream(oldStream, newStream) { if(oldStream != newStream) { cudaStreamSynchronize(oldStream); nppSetStream(newStream); } }
+
 static inline void throw_no_cuda() { CV_Error(cv::Error::StsNotImplemented, "The called functionality is disabled for current build or platform"); }
 
 namespace cv { namespace cuda
@@ -139,13 +141,13 @@ namespace cv { namespace cuda
         inline explicit NppStreamHandler(Stream& newStream)
         {
             oldStream = nppGetStream();
-            nppSetStream(StreamAccessor::getStream(newStream));
+            nppSafeSetStream(oldStream, StreamAccessor::getStream(newStream));
         }
 
         inline explicit NppStreamHandler(cudaStream_t newStream)
         {
             oldStream = nppGetStream();
-            nppSetStream(newStream);
+            nppSafeSetStream(oldStream, newStream);
         }
 
         inline ~NppStreamHandler()
--- a/modules/cudaarithm/src/reductions.cpp
+++ b/modules/cudaarithm/src/reductions.cpp
@@ -157,8 +157,6 @@ void cv::cuda::meanStdDev(InputArray _src, OutputArray _dst, Stream& stream)
     BufferPool pool(stream);
     GpuMat buf = pool.getBuffer(1, bufSize, CV_8UC1);
 
-    NppStreamHandler h(StreamAccessor::getStream(stream));
-
     nppSafeCall( nppiMean_StdDev_8u_C1R(src.ptr<Npp8u>(), static_cast<int>(src.step), sz, buf.ptr<Npp8u>(), dst.ptr<Npp64f>(), dst.ptr<Npp64f>() + 1) );
 
     syncOutput(dst, _dst, stream);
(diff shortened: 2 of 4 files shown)


modules/core/include/opencv2/core/private.cuda.hpp
```diff
@@ -139,13 +141,13 @@ namespace cv { namespace cuda
         inline explicit NppStreamHandler(Stream& newStream)
         {
             oldStream = nppGetStream();
-            nppSetStream(StreamAccessor::getStream(newStream));
+            safeNppSetStream(oldStream, StreamAccessor::getStream(newStream));
```
alalek: Perhaps we should bailout if newStream equals to oldStream

modules/core/include/opencv2/core/private.cuda.hpp
```diff
@@ -139,13 +141,13 @@ namespace cv { namespace cuda
...
         }
 
         inline explicit NppStreamHandler(cudaStream_t newStream)
         {
             oldStream = nppGetStream();
-            nppSetStream(newStream);
+            nppSafeSetStream(oldStream, newStream);
         }
 
         inline ~NppStreamHandler()
```
nglee: Maybe the destructor should also use the macro.

tomoaki0705:
As I wrote in #11063, this patch is insufficient to avoid the crash.
I switched to ```cudaDeviceSynchronize```, but it didn't work.
I tried adding ```cudaDeviceSynchronize``` before the deconstructor (as @nglee suggested) but it didn't work.

Please wait merge until I find a solid fix for this.

nglee:
@tomoaki0705 
Hi, as I've commented above, I'd like to suggest changing `nppSetStream` to `nppSafeSetStream` in the deconstructor of `class NppStreamHandler`.

That is, from [here](https://github.com/opencv/opencv/blob/1ba25c5569fbbff135c65336cdc7665cd6593734/modules/core/include/opencv2/core/private.cuda.hpp#L153-L156),
```
inline ~NppStreamHandler()
{
    nppSetStream(oldStream);
}
```
we might want to change above code to:
```
inline ~NppStreamHandler()
{
    nppSafeSetStream(oldStream);
}
```
The npp kernels may still be running when the `NppStreamHandler` object is being destroyed, and we should wait for the kernel to finish before switching npp stream in such cases.

tomoaki0705:
Ahhh!!!! Very nice catch @nglee, 
At last, I could be released from this two month duration headache....

PR #11069, opened 2018-03-14
fix some samples colliding with std::beta

### LINKED ISSUE #11066: C++17 break
https://github.com/opencv/opencv/blob/master/samples/cpp/tutorial_code/ImgProc/changing_contrast_brightness_image/changing_contrast_brightness_image.cpp#L12 imports the contents of namespace std via `using namespace std`, and then defines a namespace-scope variable named `beta`. This collides with the function [`std::beta`](http://en.cppreference.com/w/cpp/numeric/special_math/beta) introduced in C++17.

berak:
resolves #11066 

replace `using namespace std` with selective imports like `using std::cout`


samples/cpp/cloning_gui.cpp
```diff
@@ -39,7 +39,11 @@
 #include <iostream>
 #include <stdlib.h>
 
-using namespace std;
+using std::cin;
```
alalek: Could you please add some one-line comment about avoiding using of `using namespace std;` which cause conflict in C++17 (to prevent future refactoring back to "using namespace std")

berak: sure.

PR #11072, opened 2018-03-14
core: use explicit for cv::AutoBuffer

alalek:
To avoid compilation of this code:
- `buf = 0;`

This code can be received after refactoring of 1D `cv::Mat` to `cv::AutoBuffer`.
- `cv_mat = 0` calls `setTo()`.
- `cv::AutoBuffer` calls `allocate(0)`, - this is wrong.


vpisarev:
:+1:

PR #11073, opened 2018-03-14
Better text contrast of pixel brightness values in Qt window

geordi:
This offsets text brightness of pixel brightness values
by offsetting it by 127 to the curent pixel value.
The text is now readable even if pixels are black.

### This pullrequest changes

This PR changes the way pixel brightness values are displayed in the Qt window.
At this point, brightness values are printed using black font, so values are unreadable for black/dark
pixels. This patch offsets text brightness by 127 to the current pixel value, so it's always readable.

Screenshots of before and after states:

before

after


alalek:
It would be nice to add two screenshots into PR description before/after this fix.

geordi:
I've updated the PR description with screenshots of before and after states.

PR #11075, opened 2018-03-14
Fix crash on exit in opencv_annotation

iluxave:
destroyWindow was called twice during completion of the
annotation procedure, resulting in a crash, and failure to write
an output file

### This pullrequest changes

This pull request drops extra call to destroyWindow in the opencv_annotate

### DIFF
--- a/apps/annotation/opencv_annotation.cpp
+++ b/apps/annotation/opencv_annotation.cpp
@@ -140,7 +140,6 @@ vector<Rect> get_annotations(Mat input_image)
         switch( key_pressed )
         {
         case 27:
-                destroyWindow(window_name);
                 stop = true;
                 break;
         case 99:


alalek:
What OS / GUI are you using?
Is it a crash (like SIGSEGV) or just uncatched C++ exception?

iluxave:
I'm on Linux. I tried both Qt5 and GTK. It's an unhandled exception:
OpenCV(3.4.1-dev) Error: Assertion failed (found && "Can't destroy non-registered window") in cvDestroyWindow, file /home/ilya/ai/opencv-build/opencv-git/modules/highgui/src/window_gtk.cpp, line 1257
  what():  OpenCV(3.4.1-dev) /home/ilya/ai/opencv-build/opencv-git/modules/highgui/src/window_gtk.cpp:1257: error: (-215) Assertion failed: found && "Can't destroy non-registered window" in function cvDestroyWindow

Aborted

With the change, the error is gone, and the annotations file gets generated properly

alalek:
Exception ("Assert failed") is expected behavior in this case (there is no crash, just unhandled exception).
Patch looks good to me :+1:

PR #11081, opened 2018-03-15
cmake: fix android examples dependencies

alalek:
`add_dependencies(${target})` doesn't control regeneration of .apk file, because "${target}" is a dummy CMake target.


alalek:
:+1:

PR #11083, opened 2018-03-15
Fix #9363

### LINKED ISSUE #9363: lsd.cpp debug
hello, I find that the codes attached here in lsd.cpp in opencv3.2 is useless. The codes aim to make the list starts by the larger bin, howerer ,it doesn't works.

    // Sort
    int idx = n_bins - 1;
    for(;idx > 0 && !range_s[idx]; --idx);
    coorlist* start = range_s[idx];
    coorlist* end = range_e[idx];
    if(start)
    {
        while(idx > 0)
        {
            --idx;
            if(range_s[idx])
            {
                end->next = range_s[idx];
                end = range_e[idx];
            }
        }
    }

take1014:
resolves #9363 

### This pullrequest changes
Fixed as list was not sorted as intended based on norm value.

### DIFF
--- a/modules/imgproc/src/lsd.cpp
+++ b/modules/imgproc/src/lsd.cpp
@@ -268,12 +268,6 @@ private:
 
     std::vector<normPoint> ordered_points;
 
-    // Compare norm
-    static inline bool compare_norm(const normPoint& n1, const normPoint& n2)
-    {
-        return (n1.norm > n2.norm);
-    }
-
     struct rect
     {
         double x1, y1, x2, y2;    // first and second point of the line segment
@@ -386,6 +380,13 @@ private:
  * @return      Whether the point is aligned.
  */
     bool isAligned(int x, int y, const double& theta, const double& prec) const;
+
+public:
+    // Compare norm
+    static inline bool compare_norm( const normPoint& n1, const normPoint& n2 )
+    {
+        return (n1.norm > n2.norm);
+    }
 };
 
 /////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


modules/imgproc/src/lsd.cpp
```diff
@@ -75,6 +75,12 @@ struct edge
     bool taken;
 };
 
+struct coorlist
+{
+    cv::Point2i p;
+    int norm;
```
alalek: Looks like this struct is misnamed ("list" part should be removed after removing "next" entry - something like PointEntry).

modules/imgproc/src/lsd.cpp
```diff
@@ -156,6 +162,12 @@ inline double log_gamma_lanczos(const double& x)
     }
     return a + log(b);
 }
+
+// Compare norm
+inline bool compare_norm(const coorlist& n1, const coorlist& n2)
```
alalek: You can put this in class `LineSegmentDetectorImpl` as **static** member function.

modules/imgproc/src/lsd.cpp
take1014: Thanks for your comment.
I think to change the structure name from "coorlist" to "coorPoint", what do you think?
In addition, since the member variable "list" does not already have a list structure, I'd like to change the name of "list" to "ordered_points", what do you think?

alalek: "coor" here is a just "typo" for "coordinate". "coordinatePoint" is redundant name.

> ordered_points

Sounds good.

modules/imgproc/src/lsd.cpp
take1014: I added this function in LineSegmentDetectorImpl class as static member function of public member.

PR #11091, opened 2018-03-16
dnn: add an openpose.cpp sample

berak:
### This pullrequest changes
add a corresponding cpp sample to the existing openpose.py

it also adds the ability, to use the openpose hand model (all it needs is another lookup table)

question: is `openpose_pose_mpi_faster_4_stages.prototxt` the correct one for MPI ? (i only have COCO)

### DIFF
--- a/samples/dnn/openpose.cpp
+++ b/samples/dnn/openpose.cpp
@@ -1,0 +1,168 @@
+//
+//  this sample demonstrates the use of pretrained openpose networks with opencv's dnn module.
+//
+//  it can be used for body pose detection, using either the COCO model(18 parts):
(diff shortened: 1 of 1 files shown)


dkurt:
@berak, That's cool!
[openpose_pose_mpi.prototxt](https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/openpose_pose_mpi.prototxt) and [openpose_pose_mpi_faster_4_stages.prototxt](https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/openpose_pose_mpi_faster_4_stages.prototxt) are both for the same `pose_iter_160000.caffemodel`. The second one just return results from early layers.

dkurt:
:+1:

PR #11092, opened 2018-03-16
Update for MSMF-based VideoCapture and VideoWriter

terfendail:
### This pullrequest changes
Fixed race condition and extended a set of supported formats for VideoCapture
Fixed top-bottom row ordering for VideoWriter

### DIFF
--- a/modules/videoio/src/cap_msmf.cpp
+++ b/modules/videoio/src/cap_msmf.cpp
@@ -505,11 +505,24 @@ private:
 class Media_Foundation
 {
 public:
-    virtual ~Media_Foundation(void);
-    static Media_Foundation& getInstance();
-    bool buildListOfDevices();
+    ~Media_Foundation(void) { CV_Assert(SUCCEEDED(MFShutdown())); }
+    static Media_Foundation& getInstance()
+    {
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1800)
+        //CXX11
+        static thread_local Media_Foundation instance;
+#else //__cplusplus >= 201103L || _MSC_VER >= 1800
+        //CXX98
+#ifdef _WIN32
+        static __declspec(thread) Media_Foundation instance;
+#else
+        static __thread Media_Foundation instance;
+#endif
+#endif
+        return instance;
+    }
 private:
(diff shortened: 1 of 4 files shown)


modules/videoio/src/cap_msmf.cpp
```diff
@@ -4023,261 +3976,147 @@ const GUID CvVideoWriter_MSMF::FourCC2GUID(int fourcc)
...
 {
-    if (!initiated)
+    if (initiated)
     {
-        return;
+        initiated = false;
+        sinkWriter->Finalize();
+        SafeRelease(&sinkWriter);
+        MFShutdown();
+        CoUninitialize();
```
alalek: > MFShutdown();
> CoUninitialize();

We should not call global cleanup/unload functions.

The same note is for other places.

modules/videoio/src/cap_msmf.cpp
```diff
@@ -4023,261 +3976,147 @@ const GUID CvVideoWriter_MSMF::FourCC2GUID(int fourcc)
...
+            BYTE* px = pData;
+            for (int rowIdx = 0; rowIdx < img->height; rowIdx++)
+            {
+                char* rowStart = img->imageData + rowIdx*img->widthStep;
+                for (int colIdx = 0; colIdx < img->width; colIdx++)
+                {
+                    *(px++) = rowStart[colIdx * img->nChannels + bIdx];
+                    *(px++) = rowStart[colIdx * img->nChannels + gIdx];
+                    *(px++) = rowStart[colIdx * img->nChannels + rIdx];
+                    *(px++) = 0;
```
alalek: Use cvtColor(BGR2BGRA / GRAY2BGRA) into **mapped** `cv::Mat` on buffer.

tomoaki0705:
Unfortunately, this PR breaks the build on Visual Studio 2013 and earlier.
Thread local storage is only supported on POD on VS2013.
https://blogs.msdn.microsoft.com/vcblog/2014/11/17/c111417-features-in-vs-2015-preview/
https://stackoverflow.com/questions/8772733/c11-thread-local-keyword-support-in-visual-studio-11/8773388#comment10946155_8773388

I wrote a small workaround, but I couldn't fix it totally.

PR #11093, opened 2018-03-16
code: add override/final specifiers

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1588

- available in C++11.
- `override` is a check useful during interfaces refactoring
- `final` allows better compiler optimization
- tests code are out of scope for now

Future TODOs:
- remove unnecessary "virtual" (need to fix/check bindings parsers)

```
allow_multiple_commits=1
docker_image:Linux x64 Debug=ubuntu:16.04
```


vpisarev:
:+1:

PR #11098, opened 2018-03-17
Import models from Intel's deep learning deployment toolkit

dkurt:
### This pullrequest changes
This PR enables OpenCV to run networks from [Intel's deep learning deployment toolkit](https://software.seek.intel.com/deep-learning-deployment). Comparing with Caffe/TensorFlow/Torch/Darknet importers, OpenCV does not parse intermediate representations from Model Optimizer but passes them directly to Inference Engine. So using OpenCV interface we can work with IRs from DLDT.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/447

```
allow_multiple_commits=1
```


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -339,6 +339,13 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
     public:
 
         CV_WRAP Net();  //!< Default constructor.
+        /** @brief Create a network from Intel's Model Optimizer intermediate representation.
+         *  @param[in] xml XML configuration file with network's topology.
+         *  @param[in] bin Binary file with trained weights.
+         *  Networks imported from Intel's Model Optimizer are lauched in Intel's Inference Engine
+         *  backend.
+         */
+        CV_WRAP Net(const String& xml, const String& bin);
```
alalek: There are a lot of frameworks uses scheme+weights pattern, so this "two strings" constructor would be very ambiguous, especially in Python or other bindings.
I believe it is better to follow existed pattern and add separate static method like `readDLDT(const String& xml, const String& bin)` instead.

dkurt: @alalek, Thanks! Replaced to `CV_WRAP static Net readFromModelOptimizer(const String& xml, const String& bin);`

modules/dnn/src/dnn.cpp
```diff
@@ -2846,10 +2887,22 @@ Net readNet(const String& _model, const String& _config, const String& _framewor
...
+    if (framework == "dldt" || modelExt == "bin" || configExt == "bin" ||
+                               modelExt == "xml" || configExt == "xml")
+    {
+        if (modelExt == "xml" || configExt == "bin")
+            std::swap(model, config);
+        return readNetFromModelOptimizer(config, model);
+    }
     CV_Error(Error::StsError, "Cannot determine an origin framework of files: " +
                               model + (config.empty() ? "" : ", " + config));
     return Net();
```
alalek: Using `CV_ErrorNoReturn()` allow to drop dead "return" statement.

modules/dnn/src/dnn.cpp
```diff
@@ -1912,6 +1913,46 @@ Net::Net() : impl(new Net::Impl)
 {
 }
 
+Net Net::readFromModelOptimizer(const String& xml, const String& bin)
+{
+    Net cvNet;
+#ifndef HAVE_INF_ENGINE
+    CV_Error(Error::StsError, "Build OpenCV with Inference Engine to enable loading models from Model Optimizer.");
```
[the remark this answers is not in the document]
dkurt: Fixed. Thank you!

modules/dnn/src/dnn.cpp
```diff
@@ -1912,6 +1913,46 @@ Net::Net() : impl(new Net::Impl)
...
+#else
+    InferenceEngine::CNNNetReader reader;
+    reader.ReadNetwork(xml);
+    reader.ReadWeights(bin);
+
+    InferenceEngine::CNNNetwork ieNet = reader.getNetwork();
+
+    std::vector<String> inputsNames;
+    for (auto& it : ieNet.getInputsInfo())
```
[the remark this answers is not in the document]
dkurt: Yes. So `ENABLE_CXX11=ON` is required.
https://github.com/opencv/opencv/wiki/Intel's-Deep-Learning-Inference-Engine-backend

PR #11101, opened 2018-03-17
fixed bug #11100 Integer overflow in kmeans

### LINKED ISSUE #11100: Integer overflow in kmeans
- OpenCV => 3.4.1
- Operating System / Platform => Windows 64 Bit / Windows 32 Bit

##### Detailed description

I switched to 3.4.1 from 3.3.0 and my code started failing with BOWKMeansTrainer on the same dataset that worked well in 3.4.1. Error - "OpenCV(3.4.1-dev) Error: Assertion failed (a >= 0) in cv::divUp"

It comes from new code from revision 90aac764dd6e8c77265b1b5d616b0a70d4c74a42 when it tries to calculate divUp(dims * N * K, CV_KMEANS_PARALLEL_GRANULARITY) where in my case:
dims = 64
N = 28493
K = 2000
And it results in integer overflow (-647863296). Since all multipliers are ints, compiler also calls integer version of divUp.

Changing could to (size_t)(dims * N * K) fixes the issue.

##### Steps to reproduce

Try to train some large BoW vocabulary.

lamantine:
resolves #11100 

### This pullrequest changes

Added explicit type conversion to kmeans code, because existing code had integer overflow as described in ticket.

### DIFF
--- a/modules/core/src/kmeans.cpp
+++ b/modules/core/src/kmeans.cpp
@@ -128,7 +128,7 @@ static void generateCentersPP(const Mat& data, Mat& _out_centers,
 
             parallel_for_(Range(0, N),
                           KMeansPPDistanceComputer(tdist2, data, dist, ci),
-                          divUp(dims * N, CV_KMEANS_PARALLEL_GRANULARITY));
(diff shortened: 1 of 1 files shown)


alalek:
There are two more similar "divUp" expressions, please change them too.

PR #11104, opened 2018-03-18
Reading Darknet from cv::FileStorage

asciian:
### This pullrequest changes
I added new constructor for DarknetImporter to load from std::istream.
Current constructor gets to two pathnames which are raw strings. Without std::filesystem of C++17, some pathname has problem for cross-platform development. This changes makes abstractly to load files.

<cut/>

```c++
std::ifstream cfg("path/to/cfg");
std::ifstream model("path/to/model");
Net net = readNetFromDarknet(cfg, model);
```
And more, by this changes we will be able to include resources into binary, like followings.

```c++
std::istringstream cfg("some configuration text");
std::ifstream model("path/to/model");
Net net = readNetFromDarknet(cfg, model);
```

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -644,6 +644,14 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
     */
     CV_EXPORTS_W Net readNetFromDarknet(const String &cfgFile, const String &darknetModel = String());
 
+   /** @brief Reads a network model stored in <a href="https://pjreddie.com/darknet/">Darknet</a> model files.
+    *  @param cfgFile      file node to the .cfg file with text description of the network architecture.
+    *  @param darknetModel file node to the .weights file with learned network.
+    *  @returns Network object that ready to do forward, throw an exception in failure cases.
+    *  @returns Net object.
+    */
+    CV_EXPORTS_W Net readNetFromDarknet(const FileNode &cfgFile, const FileNode &darknetModel = FileNode());
+
     /** @brief Reads a network model stored in <a href="http://caffe.berkeleyvision.org">Caffe</a> framework's format.
       * @param prototxt   path to the .prototxt file with text description of the network architecture.
       * @param caffeModel path to the .caffemodel file with learned network.
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -649,8 +649,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
      *  @param bufferModel A buffer contains a content of .weights file with learned network.
      *  @returns Net object.
      */
-    CV_EXPORTS_W Net readNetFromDarknet(const std::vector<char>& bufferCfg,
-                                        const std::vector<char>& bufferModel = std::vector<char>());
+    CV_EXPORTS_W Net readNetFromDarknet(const std::vector<uchar>& bufferCfg,
+                                        const std::vector<uchar>& bufferModel = std::vector<uchar>());
 
     /** @brief Reads a network model stored in <a href="https://pjreddie.com/darknet/">Darknet</a> model files.
      *  @param bufferCfg   A buffer contains a content of .cfg file with text description of the network architecture.
@@ -674,8 +674,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
       * @param bufferModel buffer containing the content of the .caffemodel file
       * @returns Net object.
       */
-    CV_EXPORTS_W Net readNetFromCaffe(const std::vector<char>& bufferProto,
-                                      const std::vector<char>& bufferModel = std::vector<char>());
+    CV_EXPORTS_W Net readNetFromCaffe(const std::vector<uchar>& bufferProto,
+                                      const std::vector<uchar>& bufferModel = std::vector<uchar>());
 
     /** @brief Reads a network model stored in Caffe model in memory.
       * @details This is an overloaded member function, provided for convenience.
@@ -703,8 +703,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
       * @param bufferConfig buffer containing the content of the pbtxt file
       * @returns Net object.
       */
-    CV_EXPORTS_W Net readNetFromTensorflow(const std::vector<char>& bufferModel,
-                                           const std::vector<char>& bufferConfig = std::vector<char>());
+    CV_EXPORTS_W Net readNetFromTensorflow(const std::vector<uchar>& bufferModel,
+                                           const std::vector<uchar>& bufferConfig = std::vector<uchar>());
 
     /** @brief Reads a network model stored in <a href="https://www.tensorflow.org/">TensorFlow</a> framework's format.
       * @details This is an overloaded member function, provided for convenience.
@@ -778,8 +778,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
       * @param[in] bufferConfig A buffer with a content of text file contains network configuration.
       * @returns Net object.
       */
-     CV_EXPORTS_W Net readNet(const String& framework, const std::vector<char>& bufferModel,
-                              const std::vector<char>& bufferConfig = std::vector<char>());
+     CV_EXPORTS_W Net readNet(const String& framework, const std::vector<uchar>& bufferModel,
+                              const std::vector<uchar>& bufferConfig = std::vector<uchar>());
 
     /** @brief Loads blob which was serialized as torch.Tensor object of Torch7 framework.
      *  @warning This function has the same limitations as readNetFromTorch().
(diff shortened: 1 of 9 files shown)


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -617,6 +617,21 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
     */
     CV_EXPORTS_W Net readNetFromDarknet(const String &cfgFile, const String &darknetModel = String());
 
+    /** @brief Reads a network model stored in <a href="https://pjreddie.com/darknet/">Darknet</a> model files.
+    *  @param cfgStream          stream to the .cfg file with text description of the network architecture.
+    *  @param darknetModelStream stream to the .weights file with learned network.
+    *  @returns Network object that ready to do forward, throw an exception in failure cases.
+    *  @returns Net object.
+    */
+    CV_EXPORTS_W Net readNetFromDarknet(std::istream &cfgStream, std::istream &darknetModelStream);
```
dkurt: Due `std::istream` has no wrapping policy we need to use `CV_EXPORTS` here.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -508,13 +502,9 @@ namespace cv {
                 return true;
             }
 
-
-            bool ReadDarknetFromWeightsFile(const char *darknetModel, NetParameter *net)
+            bool ReadDarknetFromWeightsStream(std::istream &darknetModelStream, NetParameter *net)
```
dkurt: Do you mind just rename `darknetModelStream` to `ifile` here?

modules/dnn/src/darknet/darknet_importer.cpp
```diff
@@ -179,7 +185,32 @@ class DarknetImporter
 
 Net readNetFromDarknet(const String &cfgFile, const String &darknetModel /*= String()*/)
 {
-    DarknetImporter darknetImporter(cfgFile.c_str(), darknetModel.c_str());
+    std::ifstream cfgStream(cfgFile.c_str());
+    CV_Assert(cfgStream.is_open());
```
dkurt: There were assertions include paths to files,
```
CV_Error(cv::Error::StsParseError, "Failed to parse NetParameter file: " + std::string(cfgFile));
```
I think it's useful to keep some kind of error message that prints actual paths.

modules/dnn/include/opencv2/dnn/dnn.hpp
asciian: OK, now I will to replace `std::istream` to IStream of 3rdparty. Can I add `3rdparty/openexr/IlmImf/ImfIO.h` ? Or what do I should use?

modules/dnn/src/darknet/darknet_importer.cpp
asciian: I think so too. I added it.

modules/dnn/src/darknet/darknet_io.cpp
asciian: Of course. I renamed.

modules/dnn/include/opencv2/dnn/dnn.hpp
asciian: I got your saying. I will change std::istream to the other buffered structure like #9994.

asciian:
@dkurt I added `readNetFromDarknet(FileNode &cfgFile, FileNode &darknetModel)` instread of the first request. This changes solves the topic of this request.

1. The parameters in the file which has specific pathname, can pass into FileStorage with std::string which is read via std::ifstream. 
2. FileStorage can have the data in memory. So we will be able to include resources into binary, as `const char*`.

modules/dnn/test/test_darknet_importer.cpp
```diff
@@ -66,6 +66,18 @@ TEST(Test_Darknet, read_yolo_voc)
...
 
+TEST(Test_Darknet, read_filestorage_yolo_voc)
+{
+    std::ifstream ifile(_tf("yolo-voc.cfg").c_str());
+    std::stringstream buffer;
+    buffer << " " << ifile.rdbuf(); // FIXME: FileStorage drops first character.
+    FileStorage ofs(".xml", FileStorage::WRITE | FileStorage::MEMORY);
+    ofs.write("cfgFile", buffer.str());
+    FileStorage ifs(ofs.releaseAndGetString(), FileStorage::READ | FileStorage::MEMORY | FileStorage::FORMAT_XML);
+    Net net = readNetFromDarknet(ifs["cfgFile"]);
```
dkurt: @asciian, please remind me why you decided to replace `std::istream` to `cv::FileNode`? The last one seems to be more complicated here.

modules/dnn/test/test_darknet_importer.cpp
```diff
@@ -66,6 +66,18 @@ TEST(Test_Darknet, read_yolo_voc)
...
+    ASSERT_FALSE(net.empty());
```
dkurt: I think we need to check an accuracy too. Could you make some of existing accuracy tests to be parametric similar to one of our tests for loading models from a buffer?

https://github.com/opencv/opencv/blob/1031dfe4e8b039955a738140e39316e14f18ae28/modules/dnn/test/test_caffe_importer.cpp#L85-L116

modules/dnn/test/test_darknet_importer.cpp
asciian: @dkurt Because `std::istream` couldn't become the arguments of exporting functions and knew no way to implements the topic of this issue without `cv::FileNode`. If you know that, Please tell me.

asciian: @dkurt Ah, I know another way. With raw string, we can implements that. But the way requires adding the function two arguments which are used for telling the length of content.

dkurt:
@asciian, I've reverted this changes back to `std::istream` because we cannot write a raw binary data (weights) into a text `.xml` file because it breaks a markup.

asciian:
I got it. BTW, I guess std::istream isn’t exported for Python and Java. Do you have any idea? We have the only way via FileNode, don’t you?

dkurt:
@asciian, I see your point. Please let me experiment a bit with a FileStorage and a binary data.

vpisarev:
@dkurt, filestorage supports base64-encoded data; I suggest to use that. Maybe embedded a Darknet model (or any deep learning model) into a FileStorage is not very good idea in general and we or users can implement so alternative options. E.g. extend our Darknet parser to read the net from memory, if it was the original intention.

dkurt:
@asciian, I've added methods which receives bytes buffers via `std::vector<char>`. In python you may use it in the following way:

```python
import cv2 as cv
import numpy as np

with open('yolov3.cfg', 'rt') as f:
    cfgFile = bytearray(f.read())
with open('yolov3.weights', 'rb') as f:
    weightsFile = bytearray(f.read())

inp = np.random.standard_normal([1, 3, 416, 416]).astype(np.float32)

net = cv.dnn.readNetFromDarknet(cfgFile, weightsFile)
net.setInput(inp)
out = net.forward()
```

asciian:
It looks good. However, as my comments, 
- cloud you tell me which I should use in JavaAPI? I'd like to try it in JavaAPI. 
- Now we have three way to read these networks, with string as filename, with bytearrays as content which requires 2 arguments, with string as buffers which requires 4 arguments. It very confuses users because there are almost string but they have different semantics. Would you like to redefine the interface for the development? If you don't mind, I'll open the new issue.

dkurt:
@asciian, Thank you. I've fixed that. Please check `dnn/misc/java/test/DnnTensorFlowTest.java` file for details:

```java
File modelFile = new File(modelFileName);
byte[] modelBuffer = new byte[ (int)modelFile.length() ];

try {
    FileInputStream fis = new FileInputStream(modelFile);
    fis.read(modelBuffer);
    fis.close();
} catch (IOException e) {
    System.out.println(e.getMessage());
}
net = Dnn.readNetFromTensorflow(new MatOfByte(modelBuffer));
```

asciian:
That's a bytearray. Thank you. I said that `std::vector<char>` looks like string. But `std::vector<uchar>` a completely different from it. Please don't mind the second comments.

modules/dnn/misc/java/test/DnnTensorFlowTest.java
```diff
@@ -77,31 +90,56 @@ public void testGetLayer() {
...
+        net.setInput(inputBlob, "input");
 
+        Mat result = new Mat();
+        try {
+            net.setPreferableBackend(Dnn.DNN_BACKEND_OPENCV);
+            result = net.forward("softmax2");
+        }
+        catch (Exception e) {
+            System.out.println(e.getMessage());
+            throw e;
```
alalek: In tests we can use this:
```
fail("DNN forward() failed: " + e.getMessage());
```

dkurt: Thank you! Fixed that.

PR #11108, opened 2018-03-18
Resolves #4303

### LINKED ISSUE #4303: ability to get Hough accumulator value
Transferred from http://code.opencv.org/issues/1409

```
|| ivansorokin - on 2011-10-07 15:02
|| Priority: Normal
|| Affected: None
|| Category: imgproc, video
|| Tracker: Feature
|| Difficulty: None
|| PR: None
|| Platform: None / None
```
## ability to get Hough accumulator value

```
Currently it is not possible to get Hough accumulator value (after stage 1 in icvHoughLinesStandard).

http://stackoverflow.com/questions/3265987/opencv-get-hough-accumulator-value

I also need this value. Currently I've just copy-paste this part of function from OpenCV, but I think it can be extracted in separate function.
```
##### Alexander Shishkov on 2012-02-12 20:26

```
-   Description changed from Currently it is not possible to get Hough
    accumulator value (after stage 1 in... to Currently it is not
    possible to get Hough accumulator value (after stage 1 in... More
```

take1014:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/460

resolves #4303

```
opencv_extra=hough4303
```

### DIFF
--- a/modules/imgproc/src/hough.cpp
+++ b/modules/imgproc/src/hough.cpp
@@ -105,17 +105,12 @@ array of (rho, theta) pairs. linesMax is the buffer size (number of pairs).
 Functions return the actual number of found lines.
 */
 static void
-HoughLinesStandard( InputArray src, OutputArray lines,
+HoughLinesStandard( InputArray src, OutputArray lines, int type,
                     float rho, float theta,
                     int threshold, int linesMax,
                     double min_theta, double max_theta )
 {
-    int type = CV_32FC2;
-    if( lines.fixedType() )
-    {
-        type = lines.type();
-        CV_Assert( type == CV_32FC2 || type == CV_32FC3 );
-    }
+    CV_CheckType(type, type == CV_32FC2 || type == CV_32FC3, "Internal error");
 
     Mat img = src.getMat();
 
@@ -123,6 +118,7 @@ HoughLinesStandard( InputArray src, OutputArray lines,
     float irho = 1 / rho;
 
     CV_Assert( img.type() == CV_8UC1 );
+    CV_Assert( linesMax > 0 );
 
     const uchar* image = img.ptr();
     int step = (int)img.step;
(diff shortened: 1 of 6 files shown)


alalek:
No need to add new enums / new overloads into public OpenCV API.
Just reuse `FIXED_TYPE` flag of OutputArray with type `cv::Mat_<T>` or `std::vector<T>`. See #11170 as example.
(Don't forget to add notes into documentation of existed interfaces about possible output formats).

modules/imgproc/src/hough.cpp
```diff
@@ -184,16 +199,36 @@ HoughLinesStandard( const Mat& img, float rho, float theta,
...
+    if ( type == CV_32FC2 )
     {
-        LinePolar line;
-        int idx = _sort_buf[i];
-        int n = cvFloor(idx*scale) - 1;
-        int r = idx - (n+1)*(numrho+2) - 1;
-        line.rho = (r - (numrho - 1)*0.5f) * rho;
-        line.angle = static_cast<float>(min_theta) + n * theta;
-        lines.push_back(Vec2f(line.rho, line.angle));
+        lines = Mat::zeros(linesMax, 1, CV_32FC2);
```
alalek: This doesn't update/propagate to original "OutputArray". It replaces Mat only.
You should pass OutputArray into this function and call ".create()" to change size.

take1014: Thank you for the follow. OK, I will fix it like "HoughCircles".

take1014: I modified to pass OutputArray on function interface.Please check it when you have time.

modules/imgproc/src/hough.cpp
```diff
@@ -105,51 +105,69 @@ array of (rho, theta) pairs. linesMax is the buffer size (number of pairs).
...
+
+    Mat img = image.getMat();
+
     int i, j;
     float irho = 1 / rho;
 
     CV_Assert( img.type() == CV_8UC1 );
 
-    const uchar* image = img.ptr();
+    const uchar* _image = img.ptr();
```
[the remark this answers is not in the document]
take1014: I changed the name of this variable because I changed the argument of this function to "InputArray image". However, it should not modifiy existing code as much as possible. I change the argument name to "InputArray src" and return this variable name to its original name.

modules/imgproc/test/test_houghlines.cpp
```diff
@@ -56,9 +56,15 @@ struct SimilarWith
 };
 
 template<>
-bool SimilarWith<Vec2f>::operator()(Vec2f other)
+bool SimilarWith<Vec2f>::operator()( Vec2f other )
 {
-    return std::abs(other[0] - value[0]) < rho_eps && std::abs(other[1] - value[1]) < theta_eps;
+    return std::abs( other[0] - value[0] ) < rho_eps && std::abs( other[1] - value[1] ) < theta_eps;
```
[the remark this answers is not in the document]
take1014: I'll fix it right now.

modules/imgproc/src/hough.cpp
```diff
@@ -105,51 +105,69 @@ array of (rho, theta) pairs. linesMax is the buffer size (number of pairs).
...
+            if (ok >= 0) {ok = CV_INSTRUMENT_FUN_IPP(ippiHoughLine_Region_8u32f_C1R, image, step, srcSize, (IppPointPolar*) &_lines[0], dstRoi, ipp_linesMax, &linesCount, delta, threshold, buffer);};
+            ippsFree(buffer);
+            if (ok >= 0)
+            {
+                lines.create(linesCount, 1, CV_32FC2);
+                Mat(linesCount, 1, CV_32FC2, &_lines[0]).copyTo(lines.getMat());
+                CV_IMPL_ADD(CV_IMPL_IPP);
+                return;
+            }
+            lines.assign(Mat(0,0,CV_32FC2));
```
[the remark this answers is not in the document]
take1014: Since the existing code contained the code for clearing the output,I corrected it accordingly, but it certainly is not necessary.

modules/imgproc/test/test_houghlines.cpp
```diff
@@ -262,6 +287,73 @@ void HoughLinesPointSetTest::run_test(void)
...
+    String test_case_name = format("lines_%s_%.0f_%.2f_%d_%d_%d", picture_name.c_str(), rhoStep, thetaStep,
+                                   threshold, minLineLength, maxGap);
+    test_case_name = getTestCaseName(test_case_name);
+
+    FileStorage fs(xml, FileStorage::READ);
+    FileNode node = fs[test_case_name];
+    if(node.empty())
+    {
+        fs.release();
+        fs.open(xml, FileStorage::APPEND);
```
alalek: 1. Writing into testdata must be disabled in default builds - put this code under "#ifdef GENERATE_DATA" condition.
Currently this test doesn't check anything, because it rewrites testdata (CI cleanups build folder before every build).
2. Test data should be properly added into opencv_extra repository.

take1014: 1. OK, I added "#ifdef GENERATE_DATA" between 326-331.
2. I'll add test data into opencv_extra repository.

modules/imgproc/src/hough.cpp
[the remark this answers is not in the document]
alalek: Problem here is, that empty Mat should not have any type information (OpenCV doesn't support standalone metadata).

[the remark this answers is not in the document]
take1014: I see, I'm thinking to fix it for the following, but what about it?
lines.assign (Mat ());
And since empty Mat should not have a type, I think that it should return an error.

modules/imgproc/test/test_houghlines.cpp
take1014: I added test data into opencv_extra. Please watch this pull request.
https://github.com/opencv/opencv_extra/pull/460

take1014:
@alalek 
I fixed the content that you pointed out. Could you confirm it when you have time?

alalek:
@take1014 Please take a look on updated patch.

take1014:
I fully agree with this update. It was a great learning experience for me. Thank you very much.

dan-masek:
> Just reuse `FIXED_TYPE` flag of OutputArray with type `cv::Mat_<T>` or `std::vector<T>`. See #11170 as example.

@alalek Wouldn't that prevent the feature being used via Python bindings?

Just ran into this in https://stackoverflow.com/questions/52276295/opencvpython-houghlines-accumulator-access-since-3-4-2 and as far as I can tell, there's no way of passing a fixed type OutputArray from Python side...

ocard:
I'd also like to ask for support for this functionality in Python.

PR #11112, opened 2018-03-19
cmake: don't add include <module>/src directory to avoid conflicts

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1585

during opencv_world builds:
- opencv_world is target which includes all cpp files into project
- headers from "/src" can have similar names (especialy precomp.hpp or "utils.hpp", etc)
- including header files from "/src" should be done via relative path


alalek:
:+1:

Issue #10871, opened 2018-02-15
OpenCL: some test seems too strict


tomoaki0705:
##### Abstract
- Some OpenCL tests fail on TinkerBoard
- Rather than the implementation, I think the threshold is too strict

##### System information (version)

- OpenCV => recent master ( 252e871a8 )
- Operating System / Platform => Debian 9.3 (32bit) Tinker Board
- Compiler => gcc 6.3.0
- OpenCL => 1.2

##### Detailed description
- I got some test failure on TinkerBoard
- Each of them were test on kernel of OpenCL
- [```opencv_test_video```](https://gist.github.com/tomoaki0705/e8553518a98a0bb1437593ee21a3c5f5)
```
[  PASSED  ] 55 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] OCL_Video/PyrLKOpticalFlow.Mat/4, where GetParam() = (21, 3)
```
- [```opencv_test_core```](https://gist.github.com/tomoaki0705/94d77ed5e7e7c84f2b33c440bfd3f50b)
```
[  PASSED  ] 10500 tests.
[  FAILED  ] 32 tests, listed below:
[  FAILED  ] OCL_Arithm/Mul.Mat_Scale/40, where GetParam() = (CV_32F, Channels(1), false)
(snip)
[  FAILED  ] OCL_Arithm/Mul.Mat_Scale/47, where GetParam() = (CV_32F, Channels(4), true)
[  FAILED  ] OCL_Arithm/AddWeighted.Mat/8, where GetParam() = (CV_8S, Channels(1), false)
(snip)
[  FAILED  ] OCL_Arithm/AddWeighted.Mat/39, where GetParam() = (CV_32S, Channels(4), true)
```
- [```opencv_test_imgproc```](https://gist.github.com/tomoaki0705/31f8528bfcc479b26410b977606e3dc1)
```
[  PASSED  ] 9585 tests.
[  FAILED  ] 110 tests, listed below:
[  FAILED  ] OCL_ImgProc/CvtColor.RGB2YCrCb/0, where GetParam() = (CV_8U, false)
(snip)
```
- Since these tests pass on other test platform on build farm, I suspected that the threshold is too tight.
[...]

PR #11113, opened 2018-03-19
ARM: let the test OCL_ImgprocWarp/Resize.Mat pass on ARM/Aarch64

tomoaki0705:
* disable carotenete when calling resize on ARM/Aarch64
  * loosen eps on Aarch64

### This pullrequest changes
detail explained in #10871 ([comment](https://github.com/opencv/opencv/issues/10871#issuecomment-374154954))

### DIFF
--- a/3rdparty/carotene/hal/tegra_hal.hpp
+++ b/3rdparty/carotene/hal/tegra_hal.hpp
@@ -1433,7 +1433,8 @@ inline int TEGRA_MORPHFREE(cvhalFilter2D *context)
 
 #define TEGRA_RESIZE(src_type, src_data, src_step, src_width, src_height, dst_data, dst_step, dst_width, dst_height, inv_scale_x, inv_scale_y, interpolation) \
 ( \
-    interpolation == CV_HAL_INTER_LINEAR ? \
+    /*bilinear interpolation disabled due to rounding accuracy issues*/ \
+    /*interpolation == CV_HAL_INTER_LINEAR ? \
         CV_MAT_DEPTH(src_type) == CV_8U && CAROTENE_NS::isResizeLinearOpenCVSupported(CAROTENE_NS::Size2D(src_width, src_height), CAROTENE_NS::Size2D(dst_width, dst_height), ((src_type >> CV_CN_SHIFT) + 1)) && \
         inv_scale_x > 0 && inv_scale_y > 0 && \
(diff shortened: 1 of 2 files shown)


3rdparty/carotene/hal/tegra_hal.hpp
```diff
@@ -1433,15 +1433,16 @@ inline int TEGRA_MORPHFREE(cvhalFilter2D *context)
 
 #define TEGRA_RESIZE(src_type, src_data, src_step, src_width, src_height, dst_data, dst_step, dst_width, dst_height, inv_scale_x, inv_scale_y, interpolation) \
 ( \
-    interpolation == CV_HAL_INTER_LINEAR ? \
+    /*nearest neighbour interpolation disabled due to rounding accuracy issues*/ \
+    /*interpolation == CV_HAL_INTER_LINEAR ? \
```
alalek: > nearest neighbour
> CV_HAL_INTER_LINEAR

Looks like this comment doesn't match to actual code.

@terfendail Does "bit-exact" changes affects this resize mode? or it should be a separate mode?

terfendail: "bit-exact" linear resize is implemented as separate mode `INTER_LINEAR_EXACT = 5` so it shouldn't affect `CV_HAL_INTER_LINEAR = 1` at least directly. However "bit-exact" change affect the mode used for internal resizes almost everywhere so it could affect the reference

tomoaki0705: Thanks @alalek , I updated the comment.

@terfendail , I understand that this doesn't need to be a bit-exact, so loosening the threshold is another choice (keep the implementation as is)

May be it's a choice to keep the current implementation since we have a "bit-exact" implementation.
How does that sound ?

terfendail: I think that it would be better to retain threshold as low as possible even for "bit-inexact" resize. So I also prefer to disable carotene for linear resize while the difference root cause is unclear.

tomoaki0705: Ii have a feeling that it's a rounding error.
Not the rounding of pixel value it self, but the pixel coordinate.

Bi-linear interpolation picks up 4 pixels, and computes a weighted average.  If we pick from different pixel due to the rounding difference of coordinate, the result pixel value could be differ more than 1.
That explains why the difference goes over threshold, and also why it happens rarely.


It might be worth of taking a different way of evaluation, such as number of pixels which goes over the threshold.

PR #11134, opened 2018-03-22
Fix center of rotation

not522:
### This pullrequest changes

The center of rotation in the tutorial is off the actual center.
The following code generates a strange result.
```
import cv2
import numpy as np

img = np.arange(9, dtype=np.float64).reshape(3, 3)
rows,cols = img.shape
M = cv2.getRotationMatrix2D((cols/2,rows/2),90,1)
dst = cv2.warpAffine(img,M,(cols,rows))
print(dst)
```
```
[[0. 0. 0.]
 [2. 5. 8.]
 [1. 4. 7.]]
```
After fixing, it generates a correct result.
```
import cv2
import numpy as np

img = np.arange(9, dtype=np.float64).reshape(3, 3)
rows,cols = img.shape
M = cv2.getRotationMatrix2D((cols/2.0-0.5,rows/2.0-0.5),90,1)
dst = cv2.warpAffine(img,M,(cols,rows))
print(dst)
```
```
[[2. 5. 8.]
 [1. 4. 7.]
 [0. 3. 6.]]
```

```
allow_multiple_commits=1
```


alalek:
What is python version used?
cols/2 is 1 in Python2 and "1.5" in Python3.
I believe we should force "integer div" operation in this sample: "cols // 2".

not522:
I use Python 3.6.2.
We need to "floating div" to specify the correct center.
The modified code works correctly in both versions, Python2 and Python3.

not522:
"Integer div" does not work correctly in the following code.
```
import cv2
import numpy as np

img = np.arange(16, dtype=np.float64).reshape(4, 4)
rows,cols = img.shape
M = cv2.getRotationMatrix2D((cols//2,rows//2),90,1)
dst = cv2.warpAffine(img,M,(cols,rows))
print(dst)
```
```
[[ 0.  0.  0.  0.]
 [ 3.  7. 11. 15.]
 [ 2.  6. 10. 14.]
 [ 1.  5.  9. 13.]]
```

alalek:
Right,

`(cols - 1) / 2.0` is correct center coordinate (the same as `(cols / 2.0 - 0.5)` but without strange "0.5").
"cols - 1" here is coordinate limit: 0, 1, ... , cols - 1

not522:
I feel both are tricky, but your suggestion is better.
I will rewrite the commit.

not522:
I forgot to regenerate the rotated image.
I cannot find out how to do it...

alalek:
Existed rotated image in tutorial looks good to me (1 pix difference is not critical for demo purposes). So this change is not necessary.

not522:
Oh, .png is wrong. I fix to .jpg.

not522:
The size was also wrong. Sorry for many mistakes.

PR #11146, opened 2018-03-23
Sets a higher limit on videoio's AVI container's maximum chunk size, …

### LINKED ISSUE #11126: Maximum chunk size assertion in videoio's AVI container implementation appears to be incorrect.

- OpenCV => 3.4.1
- Operating System / Platform => Ubuntu Linux 17.10 64-bit
- Compiler => clang-5.0

##### Detailed description

The AVI container implementation in the `videoio` module appears, to me, to have a bug. In file `modules/videoio/src/container_avi.cpp` the following function is defined:

```
std::vector<char> AVIReadContainer::readFrame(frame_iterator it)
{
    m_file_stream->seekg(it->first);

    RiffChunk chunk;
    *(m_file_stream) >> chunk;
    CV_Assert(chunk.m_size <= 0xFFFF);   // <--- ATW: Bug?

    std::vector<char> result;

    result.reserve(chunk.m_size);
    result.resize(chunk.m_size);

    m_file_stream->read(&(result[0]), chunk.m_size); // result.data() failed with MSVS2008

    return result;
}
```

Notice the assertion: `CV_Assert(chunk.m_size <= 0xFFFF);`. I was running into the issue where I was unable to open an MJPG video with the VideoCapture class after writing it with the VideoWriter class. The video seems perfectly fine, and when I remove this assertion the software works as expected.

Why is this assertion present? Has this chunk size limitation possibly been increased or removed in newer implementations? It seems this assertion either needs to be updated with a larger maximum chunk size or removed all together, but I'm not sure which. Alternatively, if this is a real limit then there may be a bug in the code that _writes_ the AVI videos, since it seems to be using chunk sizes that trigger this assertion when the videos are read back in.

##### Steps to reproduce

No code. You need to get (un)lucky and generate a video that includes large chunks. I can provide a video if helpful, but this seems like more of an API / specification technicality.

anthonytw:
…and adds an explanation of the assertion.

resolves #11126

### This pullrequest changes

The upper limit for the chunk size in the AVI video container implementation in `videoio` has been increased from 64KB (too small for practical applications) to 64MB. Also, an explanation of the assertion's function is included.

### DIFF
--- a/modules/videoio/src/container_avi.cpp
+++ b/modules/videoio/src/container_avi.cpp
@@ -511,7 +511,12 @@ std::vector<char> AVIReadContainer::readFrame(frame_iterator it)
 
     RiffChunk chunk;
     *(m_file_stream) >> chunk;
-    CV_Assert(chunk.m_size <= 0xFFFF);
+
+    // Assertion added to prevent complaints from static analysis tools
+    // as the chunk size is read from a file then used to allocate
+    // memory. 64MB was chosen arbitrarily as an upper bound but it may
+    // be useful to make it configurable.
+    CV_Assert(chunk.m_size <= 67108864);
 
     std::vector<char> result;
 


csaba85:
When this one will be available in release? I'm having the same issue, and just found this topic. I did everything the same as anthonytw did, I made an MJPG with video writer. Using Windows 10, and opencv 3.4.1

PR #11159, opened 2018-03-26
samples: avoid using of legacy C-like API

alalek:
- CV_RGB() macro is moved into opencv2/imgproc.hpp from imgproc_c.h
- samples/cpp/filestorage_base64.cpp is dropped


vpisarev:
thanks! :+1:

PR #11160, opened 2018-03-26
Solves #11156

### LINKED ISSUE #11156: Behaviour of FIleStorage::APPEND different from flags "a" in fopen
OpenCV => 3.4.1
Operating System / Platform => Windows 64 Bit
Compiler => Visual Studio 2017

My code 
```
    FileStorage f;
    f.open("test.yml", FileStorage::APPEND+ FileStorage::WRITE);
    f << "a" << 28.54;
    f.release();
    f.open("test.yml", FileStorage::READ);
    cout << " First read "<<float(f["a"]) << endl;
    f.release();
    f.open("test.yml", FileStorage::APPEND + FileStorage::WRITE);
    f << "b" << 323.54;
    f.release();
    f.open("test.yml", FileStorage::APPEND + FileStorage::WRITE);
    cout << " second read "<< float(f["a"]) << endl;
    cout << " second read " << float(f["b"]) << endl;
    f.release();

```
result :
```
 First read 28.54
 second read 0
 second read 0

```

Now with APPEND : 
```
    FileStorage f;
    f.open("test1.yml", FileStorage::APPEND);
    f << "a" << 28.54;
    f.release();
    f.open("test1.yml", FileStorage::READ);
    cout << " First read "<<float(f["a"]) << endl;
    f.release();

```
there is an exception : 
`OpenCV(3.4.1-dev) Error: Unknown error code -49 (Input file is empty) in cvOpenFileStorage, file g:\lib\opencv\modules\core\src\persistence_c.cpp, line 384
`
%YAML:1.0 is missing then file is [empty](https://github.com/opencv/opencv/blob/master/modules/core/src/persistence_c.cpp#L384)

I think that in APPEND mode signature must be append when file is empty

LaurentBerger:
Check file size when FileStorage::APPEND mode is used and add signature when size is empty

resolves #11156

### DIFF
--- a/modules/core/src/persistence_c.cpp
+++ b/modules/core/src/persistence_c.cpp
@@ -206,8 +206,12 @@ cvOpenFileStorage( const char* query, CvMemStorage* dststorage, int flags, const
         // and factor=4 for YAML ( as we use 4 bytes for non ASCII characters (e.g. \xAB))
         int buf_size = CV_FS_MAX_LEN*(fs->fmt == CV_STORAGE_FORMAT_XML ? 6 : 4) + 1024;
 
-        if( append )
+        if (append)
+        {
             fseek( fs->file, 0, SEEK_END );
+            if (ftell(fs->file) == 0)
+                append = false;
+        }
 
         fs->write_stack = cvCreateSeq( 0, sizeof(CvSeq), fs->fmt == CV_STORAGE_FORMAT_XML ?
                 sizeof(CvXMLStackRecord) : sizeof(int), fs->memstorage );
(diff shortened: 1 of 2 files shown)


alalek:
The similar problem exists for JSON format too (see line 319).

Perhaps it is better to check file size early and to drop "append" flag if file is empty. See line 209

It would be nice to have small test for this too.

vpisarev:
@LaurentBerger, I agree with @alalek; perhaps, the fix should be made universal - if the APPEND flag is specified, then check the file size and drop APPEND flag if the file is empty.

LaurentBerger:
Don't push I will write a test for APPEND mode

PR #11164, opened 2018-03-26
Fixed a couple of memory leaks in videoInput::getDevice() manifesting when more than one capture devices is present on the system

corleypc:
videoInput::getDevice() would leak an IMoniker object for each device in the device list before the requested device (that is, for devices with ids smaller than the requested).  And it would leak an IMoniker object whenever the requested device wasn't the last device in the list.


vpisarev:
:+1:

PR #11166, opened 2018-03-27
Fix build issue in stiching module when CUDA Toolkit is enabled

chaorunrun:
### This pullrequest changes

The patch changes CMakeLists.txt file to include 'opencv_cudaimgproc' module as a dependency of 'stiching' module. It resolves building issues when CUDA Toolkit is enabled for it. Without the patch, 'stiching' module is not built.


alalek:
Please update your code. Fix is already on the master branch: #11033

PR #11171, opened 2018-03-27
#11143 [FIX] Normalize node risk with sample weight sum

### LINKED ISSUE #11143: DecisionTree node risk should be normalized!?
I think node_risk should be normalized with sample weight sum / sample count. Otherwise if computed as absolute sum of squared error like it is now, it’s difficult to compare with anything because the absolute value always depends on the number of samples / sum of weights of samples in the node. 

https://github.com/opencv/opencv/blob/2dff9f4c6d1ae9982e253996fbdf6a289b108881/modules/ml/src/tree.cpp#L634

This is especially bad when used as a termination criterion like here:
https://github.com/opencv/opencv/blob/2dff9f4c6d1ae9982e253996fbdf6a289b108881/modules/ml/src/tree.cpp#L389

codingforfun:
### This pullrequest changes

This pullrequest resolves #11143. 

In case of regression trees, node risk is computed as sum of squared error. To get a meaningful value to compare with it needs to be normalized to the number of samples in the node (or more generally to the sum of sample weights in this node). Otherwise the sum of squared error is highly depended on the number of samples in the node and comparison with `regressionAccuracy` parameter is not very meaningful.

After normalization `node_risk` means in fact sample variance for all samples in the node, which makes much more sense and seams to be what was originally intended by the code given that node risk is later used as a split termination criteria by
```
sqrt(node.node_risk) < params.getRegressionAccuracy()
```

### DIFF
--- a/modules/ml/src/tree.cpp
+++ b/modules/ml/src/tree.cpp
@@ -632,6 +632,7 @@ void DTreesImpl::calcValue( int nidx, const vector<int>& _sidx )
         }
 
         node->node_risk = sum2 - (sum/sumw)*sum;
+        node->node_risk /= sumw;
         node->value = sum/sumw;
     }
 }


vpisarev:
thanks! :+1:

PR #11192, opened 2018-03-29
Fixed IPP based implementation of setTo() for infinity value

### LINKED ISSUE #11044: cv::Mat::setTo incorrectly handles values of infinity

- OpenCV => 3.3.0 with IPP enabled
- Compiler => GCC 5.4.0

##### Detailed description

Behavior of cv::Mat::setTo() when an infinite 32-bit floating point value is passed in for the value seems to have changed between OpenCV 2.4.9 and 3.3.0 with IPP enabled.

In 2.4.9 it behaves as expected, replacing the indicated values with inf, but in 3.3.0 it seems to replace them with apparently random very large floating point values.

##### Steps to reproduce

This can be shown with the following code:
```
cv::Mat a = cv::Mat::ones(2,2,CV_32FC1);
cv::Mat b = cv::Mat::ones(2,2,CV_8U);
a.setTo(std::numeric_limits<float>::infinity(), b);
```

Instead of infinite values as expected, 3.4028347e+38 gets inserted into each position of matrix a.

terfendail:
resolves #11044
### This pullrequest changes
IPP based implementation of setTo() use saturation for values outside of float range that is different from generic OpenCV implementation so IPP-based implementation disabled for values outside of the range

### DIFF
--- a/modules/core/src/copy.cpp
+++ b/modules/core/src/copy.cpp
@@ -478,6 +478,11 @@ static bool ipp_Mat_setTo_Mat(Mat &dst, Mat &_val, Mat &mask)
     if(dst.channels() > 4)
         return false;
 
+    if (dst.depth() == CV_32F)
+        for (int i = 0; i < (int)(_val.total()); i++)
+            if (_val.at<double>(i) < iwTypeGetMin(ipp32f) || _val.at<double>(i) > iwTypeGetMax(ipp32f))
+                return false;
+
     if(dst.dims <= 2)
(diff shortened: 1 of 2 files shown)


alalek:
Please add small regression test.

terfendail:
Regression test added

matt-attack:
Just looking at it this seems like a questionable fix as it adds substantial overhead by needing to first check every value, unless I'm missing something.

PR #11194, opened 2018-03-30
Fix #10557

### LINKED ISSUE #10557: convertFromVASurface() dest buffer overflow

- OpenCV => :grey_question: 3.4.0
- Operating System / Platform => :grey_question: Linux(Ubuntu)
- Compiler => :grey_question: 5.4

##### Detailed description

opencl kernel YUV2BGR_NV12_8u outputs 2x2 patch for each work item, 
but ocl_convert_nv12_to_bgr() is calling it on each pixel, which is doing redundant computation and more seriously: overflow the target buffer by one row.

a simple solution is reduce globalsize's cols by one to fix the buffer overflow bug, more appropriate solution would be half globalsize's rows and cols, and inside kernel multiple the id by 2

##### Steps to reproduce

I use media SDK to output to a VA surface and trying to convert it into BGR by passing surface ID and a cv::Mat into convertFromVASurface(), program will crash and complaining about memory, and everything works fine after I increased the rows of the Mat by 1.

ltqusst:
Fix overflow bugs in conversion from NV12 VA-surface/D3D11texture2D to OpenCL UMat

resolves #10557 

### This pullrequest changes

Fix bugs in x, y coordinate checks

### DIFF
--- a/modules/core/src/opencl/cvtclr_dx.cl
+++ b/modules/core/src/opencl/cvtclr_dx.cl
@@ -92,9 +92,9 @@ void YUV2BGR_NV12_8u(
     int x = get_global_id(0);
     int y = get_global_id(1);
 
-    if (x < cols)
+    if (x + 1 < cols)
     {
-        if (y < rows)
+        if (y + 1 < rows)
         {
             __global uchar* pDstRow1 = pBGR + mad24(y, bgrStep, mad24(x, NCHANNELS, 0));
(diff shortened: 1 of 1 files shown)


ltqusst:
Without this fix, cv::directx::convertFromD3D11Texture2D() and cv::va_intel::convertFromVASurface() will definitely crash, so please merge it if possible, Thanks a lot!

PR #11195, opened 2018-03-30
fix build issues with QT/OpenGL

alalek:
related #11093 #11189 

```
force_builders=Custom
docker_image:Custom=qt:16.04
buildworker:Custom=linux-2
```


alalek:
:+1:

PR #11196, opened 2018-03-30
#11189 Forget cv namespace.

Halo9Pan:
resolves #11189

### This pullrequest changes
Forget cv namespace in commit e28cc973bf5ad80988ea406468bd4f80386123ca, and this will make the build failed.

### DIFF
--- a/samples/directx/d3d10_interop.cpp
+++ b/samples/directx/d3d10_interop.cpp
@@ -135,7 +135,7 @@ public:
         if (!m_cap.read(m_frame_bgr))
             return -1;
 
-        cv::cvtColor(m_frame_bgr, m_frame_rgba, COLOR_BGR2RGBA);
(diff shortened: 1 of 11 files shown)


samples/directx/d3d10_interop.cpp
```diff
@@ -135,7 +135,7 @@ class D3D10WinApp : public D3DSample
         if (!m_cap.read(m_frame_bgr))
             return -1;
 
-        cv::cvtColor(m_frame_bgr, m_frame_rgba, COLOR_BGR2RGBA);
+        cv::cvtColor(m_frame_bgr, m_frame_rgba, cv::COLOR_BGR2RGBA);
```
alalek: This file doesn't have build problems.
"cv::" is not necessary in this file, see line 25: "using namespace cv;"

There is problem in `opengl_interop.cpp` only (with missing "using namespace cv;"), which violates existed guidelines for OpenCV samples.
+ WinRT samples.

So it is better to check/add "using namespace cv;" instead.

PR #11200, opened 2018-04-01
android: refactor Android SDK detection

alalek:
- properly detect Android SDK support of Ant/Gradle projects
- divide processing of And/Gradle based projects


cmake/android/OpenCVDetectAndroidSDK.cmake
```diff
@@ -0,0 +1,210 @@
...
+    endif()
+  endif()
+
+  if(NOT DEFINED ANDROID_PROJECTS_BUILD_TYPE)
+    if(ANDROID_PROJECTS_SUPPORT_ANT)
+      ocv_update(ANDROID_PROJECTS_BUILD_TYPE "ANT")
+    elseif(ANDROID_PROJECTS_SUPPORT_GRADLE)
+      ocv_update(ANDROID_PROJECTS_BUILD_TYPE "GRADLE")
+    else()
+      message(FATAL_ERROR "Android SDK: Can't build Android projects as requested by BUILD_ANDROID_PROJECTS=ON variable.\n"
```
mshabunin: I'd add "neither Ant nor Gradle has been found" comment to this error message to make it more clear.

alalek: Gradle (wrapper-downloader) is usually shipped as part of Android project. So we don't check existence of this and message gradle not found is not correct (because we must ship it).
Here we check support from Android SDK at first.

mshabunin: I've met this error and was confused at first. Then I looked through the code and found that I did not install _ant_ (I did not try to build with gradle). Other messages in this file provide more information regarding the root cause of the problem, but this one does not give any clues.

Issue #11206, opened 2018-04-02
Possible leaks in pycvSetMouseCallback, pycvCreateTrackbar, and pycvCreateButton


dan-masek:
Continuing from #11205 ... I'm reading through `modules/python/src2/cv2.cpp` and another thing struck me. The functions in question ([`pycvSetMouseCallback`](https://github.com/opencv/opencv/blob/3.4.1/modules/python/src2/cv2.cpp#L1572), [`pycvCreateTrackbar`](https://github.com/opencv/opencv/blob/3.4.1/modules/python/src2/cv2.cpp#L1592), and [`pycvCreateButton`](https://github.com/opencv/opencv/blob/3.4.1/modules/python/src2/cv2.cpp#L1633)) follow the same pattern, so I'll use `pycvSetMouseCallback` since it uses the fewest parameters.

The actual call of the C++ API function to set the callback looks like this:

    ERRWRAP2(setMouseCallback(name, OnMouse, Py_BuildValue("OO", on_mouse, param)));

In that statement, we call `Py_BuildValue` which allocates a new `PyObject*` -- now, what happens, when the user calls the function again? Aren't we leaking memory, since we never `DECREF` the previously allocated `PyObject`?

i.e. shouldn't there be something such as the following?

    static PyObject* last_mouse_cb_param = NULL;

    static PyObject *pycvSetMouseCallback(PyObject*, PyObject *args, PyObject *kw)
    {
        // ....
        if (last_mouse_cb_param) {
            Py_DECREF(last_mouse_cb_param);
            last_mouse_cb_param = NULL;
        }
        last_mouse_cb_param = Py_BuildValue("OO", on_mouse, param);
        ERRWRAP2(setMouseCallback(name, OnMouse, last_mouse_cb_param));
        Py_RETURN_NONE;
    }

alalek:
Looks good.

Please put "global" variable into the function itself. No need to keep it in global space.

dan-masek:
@alalek I wrote patches for this and #11205. Should I make new pull requests for those two branches?

alalek:
@dan-masek Yes, please create new Pull request.

dan-masek:
OK. Added oull request #11217 to solve this.

Issue #11205, opened 2018-04-01
Possible leak in Python OnChange and OnButtonChange?


dan-masek:
I've just been reading through `modules/python/src2/cv2.cpp` and noticed a slight discrepancy.

There are 3 static handlers - [`OnMouse`](https://github.com/opencv/opencv/blob/master/modules/python/src2/cv2.cpp#L1548), [`OnChange`](https://github.com/opencv/opencv/blob/master/modules/python/src2/cv2.cpp#L1578), and [`OnButtonChange`](https://github.com/opencv/opencv/blob/master/modules/python/src2/cv2.cpp#L1611). All three of them call the Python function as follows:

    PyObject *r = PyObject_Call(PyTuple_GetItem(o, 0), args, NULL);

Now,  [`OnMouse`](https://github.com/opencv/opencv/blob/master/modules/python/src2/cv2.cpp#L1548) continues in the following way:

    if (r == NULL)
        PyErr_Print();
    else
        Py_DECREF(r);

whereas [`OnChange`](https://github.com/opencv/opencv/blob/master/modules/python/src2/cv2.cpp#L1586), and [`OnButtonChange`](https://github.com/opencv/opencv/blob/master/modules/python/src2/cv2.cpp#L1628) don't do a `Py_DECREF(r);`.

That seems a bit odd -- shouldn't there a be DECREF in all 3 functions?

dan-masek:
Added oull request #11216  to solve this.

PR #11207, opened 2018-04-02
Add Python support for error message handlers.

dan-masek:
Based on [this StackOverflow question](https://stackoverflow.com/questions/49602465/how-to-stop-opencv-error-message-from-printing-in-python) I took an inspiration in the current `highgui` callback wrappers, and implemented a wrapper for `cv::redirectError`.

I attempted to address the concerns I've raised in issues #11205 and #11206, and also added support for resetting to a default handler.

### DIFF
--- a/modules/python/src2/cv2.cpp
+++ b/modules/python/src2/cv2.cpp
@@ -1535,6 +1535,61 @@ PyObject* pyopencv_from(const Moments& m)
 
 #include "pyopencv_custom_headers.h"
 
+static int OnError(int status, const char *func_name, const char *err_msg, const char *file_name, int line, void *userdata)
+{
+    PyGILState_STATE gstate;
+    gstate = PyGILState_Ensure();
+
+    PyObject *o = (PyObject*)userdata;
+    PyObject *args = Py_BuildValue("isssiO", status, func_name, err_msg, file_name, line, PyTuple_GetItem(o, 1));
+
+    PyObject *r = PyObject_Call(PyTuple_GetItem(o, 0), args, NULL);
+    if (r == NULL) {
+        PyErr_Print();
+    } else {
+        Py_DECREF(r);
+    }
+
+    Py_DECREF(args);
+    PyGILState_Release(gstate);
+
+    return 0; // The return value isn't used
+}
+
+// Keep track of the previous handler parameter, so we can decref it when no longer used
+static PyObject* last_on_error_param = NULL;
+
+static PyObject *pycvRedirectError(PyObject*, PyObject *args, PyObject *kw)
+{
+    const char *keywords[] = { "on_error", "param", NULL };
+    PyObject *on_error;
+    PyObject *param = NULL;
+
+    if (!PyArg_ParseTupleAndKeywords(args, kw, "O|O", (char**)keywords, &on_error, &param))
+        return NULL;
+
+    if ((on_error != Py_None) && !PyCallable_Check(on_error))  {
+        PyErr_SetString(PyExc_TypeError, "on_error must be callable");
+        return NULL;
+    }
+    if (param == NULL) {
+        param = Py_None;
+    }
+
+    if (last_on_error_param) {
+        Py_DECREF(last_on_error_param);
+        last_on_error_param = NULL;
+    }
+
+    if (on_error == Py_None) {
+        ERRWRAP2(redirectError(NULL));
+    } else {
+        last_on_error_param = Py_BuildValue("OO", on_error, param);
+        ERRWRAP2(redirectError(OnError, last_on_error_param));
+    }
+    Py_RETURN_NONE;
+}
+
 static void OnMouse(int event, int x, int y, int flags, void* param)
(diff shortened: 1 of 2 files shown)


alalek:

> redirectError(onError [, param])

"param" is not needed in Python. Language itself supports closures, so no need to pass "userData" or similar things.
(we plan to drop these "param" args from other functions with callbacks in the future releases).

alalek:
BTW, `redirectError` is C API legacy function.
Consider using Python "try/except" instead.

dan-masek:
"Consider using Python "try/except" instead."

That's not the point, setting a custom error handler doesn't stop the exceptions. The main intent is to stop the default error handler from printing into stderr even when you catch the exception.

See: https://github.com/opencv/opencv/blob/3.4.1/modules/core/src/system.cpp#L888

As far as I can see, this is the simplest way to solve it, given that the functionality already exists and just misses a Python binding.

dan-masek:
Implemented your suggestions.

Here is a Python script to show the functionality:

```
import cv2
from functools import partial

def verbose_error_handler(status, func_name, err_msg, file_name, line):
    print "Status = %d" % status
    print "Function = %s" % func_name
    print "Message = %s" % err_msg
    print "Location = %s(%d)" % (file_name, line)
    
def extended_error_handler(status, func_name, err_msg, file_name, line, userdata):
    verbose_error_handler(status, func_name, err_msg, file_name, line)
    print "User Data = %s" % userdata
    
def silent_error_handler(status, func_name, err_msg, file_name, line):
    pass
    

print "** Default handler"
try:
    cv2.imshow("", None) # This causes an assert
except cv2.error as e:
    pass

print "** Using verbose handler"
cv2.redirectError(verbose_error_handler)
try:
    cv2.imshow("", None) # This causes an assert
except cv2.error as e:
    pass
    
print "** Using extended handler"
cv2.redirectError(partial(extended_error_handler, userdata=42))
try:
    cv2.imshow("", None) # This causes an assert
except cv2.error as e:
    pass

print "** Using silent handler"
cv2.redirectError(silent_error_handler)
try:
    cv2.imshow("", None) # This causes an assert
except cv2.error as e:
    pass

print "** Back to default handler"
cv2.redirectError(None)
try:
    cv2.imshow("", None) # This causes an assert
except cv2.error as e:
    pass

```

Output:

```
** Default handler
OpenCV(3.4.1) Error: Assertion failed (size.width>0 && size.height>0) in cv::imshow, file F:\GitHub\opencv\modules\highgui\src\window.cpp, line 364
** Using verbose handler
Status = -215
Function = cv::imshow
Message = size.width>0 && size.height>0
Location = F:\GitHub\opencv\modules\highgui\src\window.cpp(364)
** Using extended handler
Status = -215
Function = cv::imshow
Message = size.width>0 && size.height>0
Location = F:\GitHub\opencv\modules\highgui\src\window.cpp(364)
User Data = 42
** Using silent handler
** Back to default handler
OpenCV(3.4.1) Error: Assertion failed (size.width>0 && size.height>0) in cv::imshow, file F:\GitHub\opencv\modules\highgui\src\window.cpp, line 364

```

alalek:
@dan-masek I added small test based on your code snippets.

dan-masek:
@alalek Good call. Thanks.

srihamat:
Sorry to disturbed, but since this change was merged into main OpenCV, why do I still cannot redirect error from opencv cpp raised error which I use python ? does someone already overwrite it back or make it's not working at all ?

sturkmen72:
it works on Python 3

```
import numpy as np
import cv2
from functools import partial

def verbose_error_handler(status, func_name, err_msg, file_name, line):
    print("Status = %d" % status)
    print("Function = %s" % func_name)
    print("Message = %s" % err_msg)
    print("Location = %s(%d)" % (file_name, line))
    
def extended_error_handler(status, func_name, err_msg, file_name, line, userdata):
    verbose_error_handler(status, func_name, err_msg, file_name, line)
    print("User Data = %s" % userdata)
    
def silent_error_handler(status, func_name, err_msg, file_name, line):
    pass
    

print("** Default handler")
try:
    cv2.imshow("", None) # This causes an assert
except cv2.error as e:
    pass

print("** Using verbose handler")
cv2.redirectError(verbose_error_handler)
try:
    cv2.imshow("", None) # This causes an assert
except cv2.error as e:
    pass
    
print("** Using extended handler")
cv2.redirectError(partial(extended_error_handler, userdata=42))
try:
    cv2.imshow("", None) # This causes an assert
except cv2.error as e:
    pass

print("** Using silent handler")
cv2.redirectError(silent_error_handler)
try:
    cv2.imshow("", None) # This causes an assert
except cv2.error as e:
    pass

print("** Back to default handler")
cv2.redirectError(None)
try:
    cv2.imshow("", None) # This causes an assert
except cv2.error as e:
    pass

image = np.zeros((200, 200, 3), dtype = np.uint8)
cv2.imshow("", image)
cv2.waitKey()
```

PR #11208, opened 2018-04-02
Fixed mask reduction in seamless_clone

### LINKED ISSUE #10627: seamless_cloning.cpp off-by-1
- OpenCV => master
- Compiler => Any

##### Detailed description

I think the seamless_cloning.cpp code is off by one pixel: In the loop, maxx and maxy are set as the last pixels that should be included (since they are white).

```
    for(int i=0;i<h;i++)
    {
        for(int j=0;j<w;j++)
        {
            if(gray.at<uchar>(i,j) == 255)
            {
                minx = std::min(minx,i);
                maxx = std::max(maxx,i);
                miny = std::min(miny,j);
                maxy = std::max(maxy,j);
            }
        }
    }
```
However, later-on, the length of the ROI is defined as  `int lenx = maxx - minx;`, thus excluding the` maxx` pixel.

##### Steps to reproduce

For example, having a pixel at x-positions 203 and 206 will generate a `lenx` of `206 - 203 = 3` pixels, while the actual length should be `4` _(pixels at 203, 204, 205 and 206)_.

terfendail:
resolves #10627
### This pullrequest changes
Fixed mask reduction in seamless_clone.
Renamed a few variables to make the code more obvious.

### DIFF
--- a/modules/photo/src/seamless_cloning.cpp
+++ b/modules/photo/src/seamless_cloning.cpp
@@ -75,26 +75,26 @@ void cv::seamlessClone(InputArray _src, InputArray _dst, InputArray _mask, Point
         {
             if(gray.at<uchar>(i,j) == 255)
             {
-                minx = std::min(minx,i);
-                maxx = std::max(maxx,i);
-                miny = std::min(miny,j);
-                maxy = std::max(maxy,j);
+                miny = std::min(miny,i);
+                maxy = std::max(maxy,i);
+                minx = std::min(minx,j);
+                maxx = std::max(maxx,j);
             }
         }
     }
 
-    int lenx = maxx - minx;
-    int leny = maxy - miny;
+    int lenx = maxx - minx + 1;
+    int leny = maxy - miny + 1;
 
-    int minxd = p.y - lenx/2;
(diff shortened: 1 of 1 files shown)


modules/photo/src/seamless_cloning.cpp
```diff
@@ -75,30 +75,30 @@ void cv::seamlessClone(InputArray _src, InputArray _dst, InputArray _mask, Point
...
 
-    Rect roi_d(minyd,minxd,leny,lenx);
-    Rect roi_s(miny,minx,leny,lenx);
+    Rect roi_d(minxd,minyd,lenx,leny);
+    Rect roi_s(minx,miny,lenx,leny);
 
     Mat destinationROI = dest(roi_d).clone();
 
-    Mat sourceROI = Mat::zeros(leny, lenx, src.type());
+    Mat sourceROI = Mat::zeros(lenx, leny, src.type());
```
alalek: Parameters order doesn't look correct:

> static MatExpr zeros(int rows, int cols, int type);

vpisarev:
looks good! :+1:

PR #11214, opened 2018-04-03
Resolves #9720

### LINKED ISSUE #9720: Problem with `MatExpr` division and multiplication

- OpenCV => 3.3.0

##### Detailed description

	Mat m1(1, 1, CV_32FC1);
	m1.at<float>(0) = 1.f;

	const float a = 0.1f;

	cout << (Mat)(m1.mul((a / m1))) << endl;
	cout << (Mat)(m1.mul((Mat)(a / m1))) << endl;

Both expressions `m1.mul((a / m1))` and `m1.mul((Mat)(a / m1))` are expected to give the same result (0.1), but only the second one works. the first gives 10.

take1014:
resolves #9720

・resolves #9720
・add test code later.

### DIFF
--- a/modules/core/src/matrix_expressions.cpp
+++ b/modules/core/src/matrix_expressions.cpp
@@ -420,7 +420,7 @@ void MatOp::multiply(const MatExpr& e1, const MatExpr& e2, MatExpr& res, double 
             {
                 op = '/';
                 m2 = e2.a;
-                scale = e2.alpha / scale;
+                scale *= e2.alpha;
             }
             else
                 e2.op->assign(e2, m2);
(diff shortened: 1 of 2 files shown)


modules/core/src/matrix_expressions.cpp
```diff
@@ -420,7 +420,7 @@ void MatOp::multiply(const MatExpr& e1, const MatExpr& e2, MatExpr& res, double
             {
                 op = '/';
                 m2 = e2.a;
-                scale /= e2.alpha;
+                scale = e2.alpha / scale;
```
terfendail: Shouldn't it be `scale *= e2.alpha;`
Otherwise the change will fail following evaluation: `(Mat)(mat.mul((a / mat), 2.))`

take1014: Thank you for the comment.
It certainly fails in the case :(Mat) (mat.mul ((a / mat), 2.)).
I will correct it soon.

PR #11222, opened 2018-04-04
Adds efficient sort algorithm to KNearest

lopespt:
### This pullrequest changes

Instead of doing bubble sort, this pullrequest uses the **std:sort**  O(n. log n) algorithm


lopespt:
The error above seems nothing to do with this pullrequest. It was a timeout error in another module (OpenCL).

PR #11227, opened 2018-04-04
add universal intrinsics for HSV2RGB_f

K-Shinotsuka:
I add the universal intrinsics for HSV2RGB_f().

```
force_builders=Custom
```

<cut>

<table border=1>
 <tr><th>dstcn</th><th>n</th><th>Average Clocks (original code)</th><th>Average Clocks (new code)</th><th>Faster Rate (compared with original code)</th></tr>
 <tr><td>3</td><td>24</td><td>888.0</td><td>850.4</td><td>104.4%</td></tr>
 <tr><td>3</td><td>48</td><td>1672.0</td><td>1653.2</td><td>101.1%</td></tr>
 <tr><td>3</td><td>72</td><td>2456.4</td><td>2455.0</td><td>100.0%</td></tr>
 <tr><td>4</td><td>24</td><td>957.6</td><td>866.8</td><td>110.4%</td></tr>
 <tr><td>4</td><td>48</td><td>1924.0</td><td>1654.6</td><td>116.3%</td></tr>
 <tr><td>4</td><td>72</td><td>2692.0</td><td>2453.6</td><td>109.7%</td></tr>
</table>

Measurement Conditions
OS : Ubuntu 17.10
Compiler : g++ (Ubuntu 7.2.0-8ubuntu3.2) 7.2.0
CPU : Intel(R) Core(TM)2 Duo CPU E8500 @ 3.16GHz
Compiler Option : -msse -msse2 -O3
Measurement Function: rdtsc()

<table border=1>
 <tr><th>dstcn</th><th>n</th><th>Average Clocks (original code)</th><th>Average Clocks (new code)</th><th>Faster Rate (compared with original code)</th></tr>
 <tr><td>3</td><td>24</td><td>4173.4</td><td>2591.8</td><td>161.0%</td></tr>
 <tr><td>3</td><td>48</td><td>15354.8</td><td>6392.4</td><td>240.2%</td></tr>
 <tr><td>3</td><td>72</td><td>36580.8</td><td>11599.0</td><td>315.4%</td></tr>
 <tr><td>4</td><td>24</td><td>40369.2</td><td>14211.0</td><td>284.1%</td></tr>
 <tr><td>4</td><td>48</td><td>49683.8</td><td>18132.8</td><td>274.0%</td></tr>
 <tr><td>4</td><td>72</td><td>66986.4</td><td>23033.4</td><td>290.8%</td></tr>
</table>

Measurement Conditions
OS : Android 6.0.1
Compiler : android-ndk-r10e ndk-build
CPU : Tegra K1 Dual Denver 2.3GHz
Additional Compiler Option : -O3 -mcpu=cortex-a8 -mfpu=neon -mfloat-abi=softfp -fPIC -march=armv7-a -ffunction-sections -funwind-tables -fstack-protector -fno-short-enums -fno-exceptions -fno-rtti
Measurement Function: clock_gettime()

```
force_builders=armv7,armv8
```

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin_sse.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_sse.hpp
@@ -1570,6 +1570,36 @@ inline void v_load_deinterleave(const unsigned* ptr, v_uint32x4& a, v_uint32x4& 
     v_transpose4x4(u0, u1, u2, u3, a, b, c, d);
 }
 
+inline void v_load_deinterleave(const float* ptr, v_float32x4& a, v_float32x4& b, v_float32x4& c)
+{
+    __m128 t0 = _mm_loadu_ps(ptr + 0);
+    __m128 t1 = _mm_loadu_ps(ptr + 4);
+    __m128 t2 = _mm_loadu_ps(ptr + 8);
+    __m128 at12 = _mm_shuffle_ps(t1, t2, 0x12);
+    a.val = _mm_shuffle_ps(t0, at12, 0x8c);
+    __m128 bt01 = _mm_shuffle_ps(t0, t1, 0x01);
+    __m128 bt12 = _mm_shuffle_ps(t1, t2, 0x23);
+    b.val = _mm_shuffle_ps(bt01, bt12, 0x88);
+    __m128 ct01 = _mm_shuffle_ps(t0, t1, 0x12);
+    c.val = _mm_shuffle_ps(ct01, t2, 0xc8);
+}
+
+inline void v_load_deinterleave(const float* ptr, v_float32x4& a, v_float32x4& b, v_float32x4& c, v_float32x4& d)
+{
+    __m128 t0 = _mm_loadu_ps(ptr +  0);
+    __m128 t1 = _mm_loadu_ps(ptr +  4);
+    __m128 t2 = _mm_loadu_ps(ptr +  8);
+    __m128 t3 = _mm_loadu_ps(ptr + 12);
+    __m128 t02lo = _mm_unpacklo_ps(t0, t2);
+    __m128 t13lo = _mm_unpacklo_ps(t1, t3);
+    __m128 t02hi = _mm_unpackhi_ps(t0, t2);
+    __m128 t13hi = _mm_unpackhi_ps(t1, t3);
+    a.val = _mm_unpacklo_ps(t02lo, t13lo);
+    b.val = _mm_unpackhi_ps(t02lo, t13lo);
+    c.val = _mm_unpacklo_ps(t02hi, t13hi);
+    d.val = _mm_unpackhi_ps(t02hi, t13hi);
+}
+
 inline void v_load_deinterleave(const uint64 *ptr, v_uint64x2& a, v_uint64x2& b, v_uint64x2& c)
 {
     __m128i t0 = _mm_loadu_si128((const __m128i*)ptr);
@@ -1796,6 +1826,41 @@ inline void v_store_interleave(float* ptr, const v_float32x4& a, const v_float32
     _mm_storeu_ps((ptr + 4), u1);
 }
 
+inline void v_store_interleave(float* ptr, const v_float32x4& a, const v_float32x4& b, const v_float32x4& c)
+{
+    __m128 u0 = _mm_shuffle_ps(a.val, b.val, 0x00);
+    __m128 u1 = _mm_shuffle_ps(c.val, a.val, 0x50);
+    __m128 v0 = _mm_shuffle_ps(u0, u1, 0x88);
+    __m128 u2 = _mm_shuffle_ps(b.val, c.val, 0x55);
+    __m128 u3 = _mm_shuffle_ps(a.val, b.val, 0xaa);
+    __m128 v1 = _mm_shuffle_ps(u2, u3, 0x88);
+    __m128 u4 = _mm_shuffle_ps(c.val, a.val, 0xfa);
+    __m128 u5 = _mm_shuffle_ps(b.val, c.val, 0xff);
+    __m128 v2 = _mm_shuffle_ps(u4, u5, 0x88);
+
+    _mm_storeu_ps(ptr + 0, v0);
+    _mm_storeu_ps(ptr + 4, v1);
+    _mm_storeu_ps(ptr + 8, v2);
+}
+
+inline void v_store_interleave(float* ptr, const v_float32x4& a, const v_float32x4& b,
+                               const v_float32x4& c, const v_float32x4& d)
+{
+    __m128 u0 = _mm_unpacklo_ps(a.val, c.val);
+    __m128 u1 = _mm_unpacklo_ps(b.val, d.val);
+    __m128 u2 = _mm_unpackhi_ps(a.val, c.val);
+    __m128 u3 = _mm_unpackhi_ps(b.val, d.val);
+    __m128 v0 = _mm_unpacklo_ps(u0, u1);
+    __m128 v1 = _mm_unpackhi_ps(u0, u1);
+    __m128 v2 = _mm_unpacklo_ps(u2, u3);
+    __m128 v3 = _mm_unpackhi_ps(u2, u3);
+
+    _mm_storeu_ps(ptr +  0, v0);
+    _mm_storeu_ps(ptr +  4, v1);
+    _mm_storeu_ps(ptr +  8, v2);
+    _mm_storeu_ps(ptr + 12, v3);
+}
+
 inline void v_store_interleave(uint64 *ptr, const v_uint64x2& a, const v_uint64x2& b, const v_uint64x2& c)
 {
     __m128i t0 = _mm_unpacklo_epi64(a.val, b.val);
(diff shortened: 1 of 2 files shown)


K-Shinotsuka:
@alalek
I'm sorry to bother you.
Please build this request in ARMv7, ARMv8, and Android pack.

modules/core/include/opencv2/core/hal/intrin_sse.hpp
```diff
@@ -1570,6 +1570,36 @@ inline void v_load_deinterleave(const unsigned* ptr, v_uint32x4& a, v_uint32x4&
...
+    __m128 t0 = _mm_loadu_ps(ptr + 0);
+    __m128 t1 = _mm_loadu_ps(ptr + 4);
+    __m128 t2 = _mm_loadu_ps(ptr + 8);
+    __m128 at12 = _mm_shuffle_ps(t1, t2, 0x12);
+    a.val = _mm_shuffle_ps(t0, at12, 0x8c);
+    __m128 bt01 = _mm_shuffle_ps(t0, t1, 0x01);
+    __m128 bt12 = _mm_shuffle_ps(t1, t2, 0x23);
+    b.val = _mm_shuffle_ps(bt01, bt12, 0x88);
+    __m128 ct01 = _mm_shuffle_ps(t0, t1, 0x12);
+    c.val = _mm_shuffle_ps(ct01, t2, 0xc8);
```
alalek: This implementation is quite different from `v_load_deinterleave()` for other 32-bit types (see `v_uint32x4` at line 1548)

Perhaps it is better to reuse exited function and load as `v_unit32x4` + `v_reinterpret_as_f32()`

kinchungwong: Just a reminder, there are these choices: 
 * UNPCKLPS / UNPCKHPS ( ```_mm_unpacklo_ps``` , ```_mm_unpackhi_ps``` )
 * The confusion coming from the difference between SHUFPS and PSHUFD can be resolved, if you pass in the same register into the two inputs of SHUFPS, i.e. ```_mm_shuffle_ps( same_reg, same_reg, shuffle_code )```
 * To emulate _mm_unpacklo_epi64 with 32-bit floating point lanes, you can define an inline function with hard-coded SHUFPS shuffle_code. To make it easier to remember, the inline function's name can spell out the shuffle code's four digits in base-four. (On x86 intrinsics headers, there is a macro, ```_MM_SHUFFLE``` (all capitalized), which converts the four permutation digits from base-four, into the 8-bit compile-time constant.)

K-Shinotsuka: Maybe your code is not faster than my code.

My code use _mm_shuffle_ps 7 times.
So latency is 7 and throughput is 7.

Your code (v_uint32x4 and v_reinterpret_as_f32) use
_mm_unpacklo_epi32 6times, _mm_unpackhi_epi64 6times.
So latency is 12 and throughput is 12.
Only IvyBridge, latency is 12 and throughput is 6.

Do I have to change my code?

modules/imgproc/src/color_hsv.cpp
```diff
@@ -134,159 +134,115 @@ struct HSV2RGB_f
...
+        v_h = v_h * v_scale;
+        v_float32x4 v_pre_sector = v_cvt_f32(v_trunc(v_h));
+        v_h = v_h - v_pre_sector;
+        v_float32x4 v_tab0 = v_v;
+        v_float32x4 v_one = v_float32x4(1.0f, 1.0f, 1.0f, 1.0f);
+        v_float32x4 v_tab1 = v_v * (v_one - v_s);
+        v_float32x4 v_tab2 = v_v * (v_one - (v_s * v_h));
+        v_float32x4 v_tab3 = v_v * (v_one - (v_s * (v_one - v_h)));
+
+        v_float32x4 v_six = v_float32x4(6.0f, 6.0f, 6.0f, 6.0f);
```
tomoaki0705: I suggest to use ```v_setall_f32```.  That's the "**DRY**"

Also, I think it's better to have ```1.0f/6.0f``` so you can replace the following ```_mm_div_ps``` to ```_mm_mul_ps``` for cost free.  Dividing is HEAVY.

modules/core/include/opencv2/core/hal/intrin_sse.hpp
K-Shinotsuka: @tomoaki0705 

>Changing the implementation of existing universal intrinsic affects much larger/wider than
> you think, (i.e. can you prove that your modification doesn't degrade any single function
> in entire OpenCV ?)

I can't prove that my modification doesn't degrade.
But wasted unpack code is not faster than my code.

>You're not the only one who reads this code. It's easy to read by keeping consistency
> among other 32bit deinterleave.

I don't think so. Many unpack code for 3 plane is hard to read.
But if you think easy to read many unpack code for 3 plane, I will change my code.

>About the performance, talking ONLY on latency is bit dangerous.
>Modern CPU has super scalar and out-of-order execution.
>This means latency could be hidden by other instruction on run-time.
>It's better to talk in seconds, not clocks.

Latency could be hidden by other instruction only at IvyBridge.
But latency cound not be hidden at other CPU. (ex. Haswell, Broadwell, Skylake)  
Please see intel intrinsics guide.

>Also, "average" varies on test case, so I recommend to use "median" value instead of average.
>Average could be affected by very slow/fast one time exceptional result, and median value is >much worth to trust.
>(That's one of the reason why OpenCV performance test measures the median as long as the >average)

1. I measure clocks 10 times.
2. I pickup 5 data at 1's result.
3. I calculate average 2's data.
So my measurement result is not include exceptional result.

>More than that, don't use magical 0x12 values for shuffle.
>Use _MM_SHUFFLE instead.

I don't think that Engineers that can write simd code are difficult to read shuffle's magic code.
But if you think hard to read, I will change my code.
I'm sorry to trouble to you, please write code using _MM_SHUFFLE.

modules/imgproc/src/color_hsv.cpp
K-Shinotsuka: @tomoaki0705 
Thank you for your advice.
I'll change my code that don't use divide intrinsics.
Is v_setall_f32 called "_mm_set~~" intrinsics?

modules/core/include/opencv2/core/hal/intrin_sse.hpp
tomoaki0705: About the throughput, your're right. it's my mistake.

> But wasted unpack code is not faster than my code.

In another word, then why don't you replace entire deinterleave/interleave by your approach ?
Then it'll keep the implementation in one place, and still it will speed up OpenCV, right ?
My concern is NOT ONLY about hard to read .  It's that having two different implementation doing same process is non-sense.

>  I measure clocks 10 times.
>   I pickup 5 data at 1's result.
>   I calculate average 2's data.
>   So my measurement result is not include exceptional result.

This is insane,,,,, which 5 data are you picking ?
More than that, what's the point doing everything in manual, when you already have implemented performance tests ?

> I don't think that Engineers that can write simd code are difficult to read shuffle's magic code.
But if you think hard to read, I will change my code.
> I'm sorry to trouble to you, please write code using _MM_SHUFFLE.

Look, ```_MM_SHUFFLE``` is used [here](https://github.com/opencv/opencv/blob/875b4e212a186afd9eb03af8b39d52b03a330c8d/modules/core/include/opencv2/core/hal/intrin_sse.hpp#L520), [here](https://github.com/opencv/opencv/blob/875b4e212a186afd9eb03af8b39d52b03a330c8d/modules/core/src/dxt.cpp#L391) and [here](https://github.com/opencv/opencv/blob/875b4e212a186afd9eb03af8b39d52b03a330c8d/modules/core/src/lapack.cpp#L959).

modules/imgproc/src/color_hsv.cpp
tomoaki0705: [I said look](https://github.com/opencv/opencv/blob/875b4e212a186afd9eb03af8b39d52b03a330c8d/modules/core/include/opencv2/core/hal/intrin_sse.hpp#L280),
and more than that, I'm not your google.

modules/core/include/opencv2/core/hal/intrin_sse.hpp
K-Shinotsuka: >In another word, then why don't you replace entire deinterleave/interleave by your approach ?
>Then it'll keep the implementation in one place, and still it will speed up OpenCV, right ?
>My concern is NOT ONLY about hard to read . It's that having two different implementation
> doing same process is non-sense.

I want alalek to decide the policy.
I say that my code is faster. Because I don't want to change my code.
Then if alalek say that you should change the code, I intend to change my code.
If alalek say that I adopt your code, I intend to keep my code.
I think that replacement for entire deinterleave/interleave code should be done after his decision.

> This is insane,,,,, which 5 data are you picking ?
>More than that, what's the point doing everything in manual, when you already have >implemented performance tests ?

Originally, don't you worry about that my measurement result include outlier?
I say my result doesn't include outlier.
Is it not enough that I measured the speed the before/after code at same condition?

>Look, _MM_SHUFFLE is used here, here and here.

I did't find the manual like "intel intrinsics guide".
So if you know the url for manual listed _MM_SHUFFLE, I want to be taught the url.
I am sorry that I have troubled you.

modules/imgproc/src/color_hsv.cpp
K-Shinotsuka: Sorry. I read intrin_cpp.hpp by mistake.
Certainly v_setall_f32 call _mm_set1_ps().
Thank you.

tomoaki0705:
What ever....
I don't have right to decide merge, and I don't have any kind of power to force you to follow my comment.
Just forget my comment, I'm out.

modules/core/include/opencv2/core/hal/intrin_sse.hpp
alalek: The policy is not mine. There is just a logic:
- we should avoid providing code duplication, or different implementations for exact same functionality (in this case it is function for deinterleaved loading of 32-bit values).
- it is better to reuse OpenCV performance tests system. If you want to propose alternative implementation, then we should be able to check regressions on different platforms.

Issue #10945, opened 2018-02-26
WImage3D not visible since OpenCV 3.4.1


dustsigns:
##### System information (version)

- OpenCV => 3.4.1
- Operating System / Platform => Ubuntu 16.04 with VTK 6.2
- Compiler => g++ 5.4.0

##### Detailed description

Images visualized with `cv::viz::WImage3D` are not visible in `cv::viz::Viz3d` windows in OpenCV 3.4.1, even though they were in 3.4.0.

##### Steps to reproduce

Consider the following minimum example:

```
#include <iostream>
#include <opencv2/viz.hpp>
#include <opencv2/highgui.hpp>

using namespace std;
using namespace cv;
using namespace viz;

int main(const int argc, const char * const argv[])
{
  if (argc != 2)
  {
    cerr << "Usage: " << argv[0] << " <input image>" << endl;
    return 1;
  }
  const Mat image = imread(argv[1]);
  assert(!image.empty());
  Viz3d visualization("Test");
  WImage3D viz_image(image, Size2d(1, 1), Vec3d(0, 0, 0), Vec3d(0, 0, -1), Vec3d(0, 1, 0));
  visualization.showWidget("Image", viz_image);
  visualization.spin();
}
```

This shows the image in version 3.4.0, but shows an empty window in 3.4.1. Note that the `WImage3D` constructor which only takes the image and a size as arguments works in both versions, so this might be related to the widget's position and thus the translation bug (#10944).

dustsigns:
Could this be categorized as a bug? The issue is also reproducible with 3.4.2 (the current master is affected by an issue that I will create shortly). It is also reproducible with newer versions of Ubuntu (18.04), VTK (7.1.1) and g++ (7.3.0), which strongly hints at it being an OpenCV issue (it works with OpenCV 3.4.0 in all setups).

alalek:
> it works with OpenCV 3.4.0

Try to run "git bisect" to detect commit which introduces problem.

dustsigns:
I think commit 28a78003f0b49336b8736b0ae131bcd0d3500b7e introduced this issue.

For others to reproduce:

[...]

PR #11230, opened 2018-04-04
Set translation in Viz3d::setViewerPose

### LINKED ISSUE #10944: Viz3d::setPose and Widget3d::setPose ignore translation
- OpenCV => 3.4.1 and later
- Operating System / Platform => Ubuntu 16.04 with VTK 6.2
- Compiler => g++ 5.4.0

##### Detailed description

Since OpenCV 3.4.1, `cv::viz::Viz3d::setPose()` ignores any translations. In 3.4.0 and prior, the behavior was correct in this regard.

##### Steps to reproduce

Consider the following minimum example:

```
#include <iostream>
#include <opencv2/viz.hpp>

using namespace std;
using namespace cv;
using namespace cv::viz;

int main()
{
  Viz3d window("Test");
  window.showWidget("Test object", WCone(1, 0.5, 100));
  window.spinOnce(1, true);
  const auto old_pose = window.getViewerPose();
  const auto old_rotation = old_pose.rvec();
  const auto old_translation = old_pose.translation();
  
  for (int i = 1; i <= 30; i++) 
  {
    cout << "Iteration " << i << endl;
    const Vec3d translation(old_translation[0], old_translation[1], old_translation[2] + i);
    const Affine3d pose(old_rotation, translation);
    window.setViewerPose(pose);
    window.spinOnce(500, true);
  }
  return 0;
}
```

When compiled with OpenCV 3.4.0, the cone moves away from the camera. Since 3.4.1, the cone does not move at all (but rotation works as expected). This seems to be a regression and also affects `cv::viz::Widget3d::setPose()`.

tkonolige:
### This pullrequest changes

`ResetCamera` was being called unnecessarily in `Viz3d::setViewerPose`. Removing the call makes `Viz3d::setViewPose` correctly set the translation of the viewer. This partially assesses #10944.

resolves #10944

### DIFF
--- a/modules/viz/src/vizimpl.cpp
+++ b/modules/viz/src/vizimpl.cpp
@@ -506,7 +506,6 @@ void cv::viz::Viz3d::VizImpl::setViewerPose(const Affine3d &pose)
     camera.SetViewUp(up_vec.val);
 
     renderer_->ResetCameraClippingRange();
-    renderer_->ResetCamera();
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////////


dustsigns:
This works for me, i.e., it solves #10944. However, it does not solve #10945. I am surprised by this, but maybe #10945 is another regression that is unrelated to `setViewerPose`.

PR #11237, opened 2018-04-05
cuda: refactor MemoryPool

alalek:
- make non-copyable (aligns inner mutex semantic to `std::mutex`)
- getMemoryPool() returns reference instead of pointer (NULL is not expected here)

```
force_builders=Custom
```


alalek:
:+1:

PR #11243, opened 2018-04-05
cuda: WITH_CUDA=OFF by default (need to enable it explicitly if required)

alalek:
```
force_builders=Custom
```


vpisarev:
:+1:

PR #11244, opened 2018-04-05
cuda: drop performance sample

alalek:
Regular OpenCV perf tests should be used instead

```
force_builders=Custom
```


vpisarev:
agree :+1:

PR #11245, opened 2018-04-05
cuda: drop samples for low-level CUDA API

### LINKED ISSUE #10967: CUDA: Drop samples which using "NVIDIA CUDA Driver API" (libcuda.so)
- driver_api_multi.cpp
- driver_api_stereo_multi.cpp

OpenCV itself (except CUVID module) and other samples use "NVIDIA CUDA Runtime API" (libcudart.so) only.

http://docs.nvidia.com/cuda/cuda-driver-api/driver-vs-runtime-api.html#driver-vs-runtime-api

Also these samples requires TBB which is disabled in default builds.

alalek:
Also these samples require TBB, so they are not tested during default OpenCV builds.

resolves #10967

```
force_builders=Custom
```


vpisarev:
ok, we are going to remove CUDA from the main repository anyway, so probably those samples can be put to opencv_contrib sometimes later :+1:

PR #11247, opened 2018-04-05
Solves issue #10506

### LINKED ISSUE #10506: python FileStorage cannot write integer values
- OpenCV => 3.4
- Compiler => vs2008 (opencv-python-contrib from PIP)

```
>>> fs = cv2.FileStorage("my.yml", 1)
>>> a = 3
>>> type(a)
<class 'int'>
>>> fs.write("a",a)
>>> fs.release()

%YAML:1.0
---
a: 3.
```

as of now, only double, String and InputArray are supported from the wrappers, an int will get silently converted to double, which will lead to interesting problems like [this one](https://github.com/opencv/opencv_contrib/issues/1504)

lopespt:
resolves #10506

possibly also [this](https://github.com/opencv/opencv_contrib/issues/1504)

### This pullrequest changes
New overload method **write**.

Now, there is no cast to double.

### DIFF
--- a/modules/core/include/opencv2/core/persistence.hpp
+++ b/modules/core/include/opencv2/core/persistence.hpp
@@ -438,6 +438,8 @@ public:
      * @param name Name of the written object
      * @param val Value of the written object
      */
+    CV_WRAP void write(const String& name, int val);
+    /// @overload
     CV_WRAP void write(const String& name, double val);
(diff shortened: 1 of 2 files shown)


vpisarev:
Thank you! All the tests pass, so the patch looks good! :+1:

PR #11252, opened 2018-04-06
Add CPU default extensions loading in IE dnn backend

sovrasov:
CPU extension are included in IE version 9354_icv and above. Some of popular custom layers (such as DetectionOutput) are located in this extension. Possible compatibility issues (build with IE packages for dlsdk, handling older versions of IE) are need to be investigated.

cc @dkurt


vpisarev:
@sovrasov, thank you! @dkurt, the patch looks good to me, but you are the best person to review this patch

PR #11253, opened 2018-04-06
Reduced direct TBB dependencies

mshabunin:
Coupled with opencv/opencv_contrib#1604

### This pullrequest changes

Removed TBB from all modules dependencies, only _core_ module depends on TBB now. All direct calls to TBB have been removed.

Notes:
* OpenNI 1&2 has been built with TBB support but not tested
* calcHist performance has decreased for 2d and 3d cases


vpisarev:
looks good! :+1:

PR #11254, opened 2018-04-06
MSMF-based VideoCapture reworked to use SourceReader

terfendail:
### This pullrequest changes
MSMF-based VideoCapture reworked to use SourceReader to support simultaneously reading of several files


vpisarev:
thanks! :+1:

PR #11262, opened 2018-04-07
imgproc: don't call .unlock() on non-holded mutex in houghCircles()

alalek:
Related message:
```
Assertion failed: (e == 0), function unlock, file /BuildRoot/Library/Caches/com.apple.xbs/Sources/libcxx/libcxx-307.5/src/mutex.cpp, line 102.
```


vpisarev:
:+1:

PR #11270, opened 2018-04-09
android: std::exception_ptr is not available on some platforms

alalek:
even with enabled C++11

Details: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58938

Error message:
```
/build/master_pack-android/opencv/modules/core/src/parallel.cpp:254:9: error: 'exception_ptr' in namespace 'std' does not name a type
         std::exception_ptr pException;
         ^
```


vpisarev:
:+1:

PR #11272, opened 2018-04-09
imgcodecs: exclude rle8.bmp from GDAL tests

### LINKED ISSUE #10118: GDAL test failure on Ubuntu 17.10
```
[ RUN      ] GDAL/Imgcodecs_FileMode.regression/3, where GetParam() = ("readwrite/rle8.bmp", 8)
ERROR 1: The BMP file is probably corrupted or too large. Image width = 480
/build/precommit_custom_linux/opencv/modules/imgcodecs/test/test_grfmt.cpp:59: Failure
Value of: single.empty()
  Actual: true
Expected: false
[  FAILED  ] GDAL/Imgcodecs_FileMode.regression/3, where GetParam() = ("readwrite/rle8.bmp", 8) (1 ms)
```

alalek:
GDAL message:
- ERROR 1: The BMP file is probably corrupted or too large. Image width = 480

GDAL fails due `GDALOpen()` call, so it can't be fixed in OpenCV.
- Ubuntu 14.06 / 16.04 works fine
- Ubuntu 17.10 / upcoming 18.04 - doesn't work

resolves #10118

### DIFF
--- a/modules/imgcodecs/test/test_grfmt.cpp
+++ b/modules/imgcodecs/test/test_grfmt.cpp
@@ -110,7 +110,8 @@ INSTANTIATE_TEST_CASE_P(All, Imgcodecs_FileMode,
 struct notForGDAL {
     bool operator()(const string &name) const {
         const string &ext = name.substr(name.size() - 3, 3);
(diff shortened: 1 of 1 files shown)


vpisarev:
:+1:

PR #11276, opened 2018-04-09
cmake: update PDB installation

alalek:
related #11029 

CMake PDB support is not complete (in 2017).
OpenCV has several fixes to handle PDB files for static libraries, like:
```
set_target_properties(<target>
    PROPERTIES
    COMPILE_PDB_NAME <name>
    COMPILE_PDB_NAME_DEBUG "<name>${OPENCV_DEBUG_POSTFIX}"
)
```


alalek:
:+1:

PR #11280, opened 2018-04-10
cmake: fix Ninja generator warning about pylintrc

alalek:
CMake generated files is a part of target depends list.
Details: `cmake --help-policy CMP0058`


alalek:
:+1:

PR #11285, opened 2018-04-10
Solves #6125

### LINKED ISSUE #6125: Negative values in normalize
I'm using Ubuntu 13.10 64 Bit, OpenCV master branch.

I get negative values in normalize() function for this example:

```cpp
    cv::Mat m = (cv::Mat_<float>(20,1)<<
     1888,
     1692,
     369,
     263,
     199,
     280,
     326,
     129,
     143,
     126,
     233,
     221,
     130,
     126,
     150,
     249,
     575,
     574,
     63,
     12);

    cv::normalize(m, m, 0, 400, NORM_MINMAX, CV_32F );
    std::cout << m <<std::endl;
```

The output is:

[400;
 358.20895;
 76.119408;
 53.518124;
 39.872066;
 57.142857;
 66.950958;
 24.946693;
 27.931768;
 24.307035;
 47.121532;
 44.562901;
 25.159914;
 24.307035;
 29.424307;
 50.533047;
 120.04265;
 119.82943;
 10.8742;
**-2.3841858e-07**]

take1014:
resolves #6125

### DIFF
--- a/modules/core/src/convert_scale.cpp
+++ b/modules/core/src/convert_scale.cpp
@@ -1885,26 +1885,34 @@ void cv::normalize( InputArray _src, InputOutputArray _dst, double a, double b,
     CV_INSTRUMENT_REGION()
 
     double scale = 1, shift = 0;
+    int type = _src.type(), depth = CV_MAT_DEPTH(type);
+
+    if( rtype < 0 )
+        rtype = _dst.fixedType() ? _dst.depth() : depth;
+
     if( norm_type == CV_MINMAX )
     {
         double smin = 0, smax = 0;
         double dmin = MIN( a, b ), dmax = MAX( a, b );
         minMaxIdx( _src, &smin, &smax, 0, 0, _mask );
         scale = (dmax - dmin)*(smax - smin > DBL_EPSILON ? 1./(smax - smin) : 0);
-        shift = dmin - smin*scale;
+        if( rtype == CV_32F )
+        {
+            scale = (float)scale;
+            shift = (float)dmin - (float)smin*scale;
+        }
+        else
+            shift = dmin - smin*scale;
     }
     else if( norm_type == CV_L2 || norm_type == CV_L1 || norm_type == CV_C )
     {
         scale = norm( _src, norm_type, _mask );
         scale = scale > DBL_EPSILON ? a/scale : 0.;
+        scale = rtype == CV_32F ? (float)scale : scale;
         shift = 0;
     }
     else
         CV_Error( CV_StsBadArg, "Unknown/unsupported norm type" );
-
-    int type = _src.type(), depth = CV_MAT_DEPTH(type);
-    if( rtype < 0 )
-        rtype = _dst.fixedType() ? _dst.depth() : depth;
 
     CV_OCL_RUN(_dst.isUMat(),
                ocl_normalize(_src, _dst, _mask, rtype, scale, shift))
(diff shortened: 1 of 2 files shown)


take1014:
Thank you for the follow.

>Consider learning how to run OpenCV performance tests and generate reports.
OK, I'll modifiy the code and do the performance test of core.

take1014:
@alalek 
I tried a variety of methods, but in order to deal with this issue outside of "convertTo", it also has nearly doubled the performance. (Although performance improved somewhat than first commit.)
I could modify the contents of "convertTo" to solve this issue, but considering the impact on other modules, I think that the benefit of this modification is small...
How do you think about this?

alalek:
1. "Regression" test doesn't fail without the "fix". Test must be fixed/replaced.
2. `normalize()` prepares scale/shift parameters as double values. But they are converted to float values during real calculation. Proper way is to prepare "float"-compatible parameters. For example:
```
        if (rtype == CV_32F)  // #6125
        {
            scale = (float)scale;
            shift = (float)dmin - (float)(smin*scale);  // calculate shift as "float" value
        }
```
3. Anyway this is just a workaround. There are many other similar places and floating-point accuracy/errors should be taken into account by application developers.

take1014:
@alalek 
I read the contents of "convertTo" and understood your words. I learned a lot about "float" and "double". Thank you.

take1014:
I revised the code with reference to your point. Can you check it when you have time?

take1014:
Ummm, there seems to be an error in the route of Ipp and ocl. I will check the detail of them.

modules/core/test/test_arithm.cpp
```diff
@@ -1918,6 +1918,31 @@ TEST(Normalize, regression_5876_inplace_change_type)
...
+        47.121532f, 44.562901f, 25.159914f, 24.307035f, 29.424307f,
+        50.533047f, 120.04265f, 119.82943f, 10.874201f  , 0.f
+    };
+
+    Mat src(Size(20, 1), CV_32F, initial_values);
+    Mat result(Size(20, 1), CV_32F, result_values);
+
+    normalize(src, src, 0, 400, NORM_MINMAX, CV_32F);
+
+    EXPECT_EQ(0, cvtest::norm(src, result, NORM_INF));
```
alalek: Perhaps it is better to check that values are "in range" 0..400 instead (as blames from the original issue).

There is no sense requiring bit-exact results for floating-point computations.

take1014: Thanks for the follow.
I modified the test code so that we can check whether normalized value exists between 0 to 400 using "EXPECT_GE" and "EXPECT_LE".

take1014:
Ummm...an exception seems to be thrown in the mac osx environment. 
It seems that I need to go get a Mac soon...

modules/core/src/convert_scale.cpp
```diff
@@ -1885,27 +1888,35 @@ void cv::normalize( InputArray _src, InputOutputArray _dst, double a, double b,
...
+            shift = (float)dmin - (float)(smin*scale);
+        }
+        else
+            shift = dmin - smin*scale;
     }
     else if( norm_type == CV_L2 || norm_type == CV_L1 || norm_type == CV_C )
     {
         scale = norm( _src, norm_type, _mask );
         scale = scale > DBL_EPSILON ? a/scale : 0.;
+        scale = rtype == CV_32F ? (float)scale : scale;
```
alalek: This change is not necessary.

PR #11286, opened 2018-04-10
Update the function rectangle

sturkmen72:
### This pullrequest changes
use of InputOutputArray instead of Mat

```
buildworker:Linux x64=linux-1
```

### DIFF
--- a/modules/imgproc/src/drawing.cpp
+++ b/modules/imgproc/src/drawing.cpp
@@ -1739,7 +1739,7 @@ PolyLine( Mat& img, const Point2l* v, int count, bool is_closed,
 /* ADDING A SET OF PREDEFINED MARKERS WHICH COULD BE USED TO HIGHLIGHT POSITIONS IN AN IMAGE */
 /* ----------------------------------------------------------------------------------------- */
 
-void drawMarker(Mat& img, Point position, const Scalar& color, int markerType, int markerSize, int thickness, int line_type)
+void drawMarker(InputOutputArray img, Point position, const Scalar& color, int markerType, int markerSize, int thickness, int line_type)
 {
     switch(markerType)
     {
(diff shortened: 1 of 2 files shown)


modules/imgproc/src/drawing.cpp
```diff
@@ -1739,63 +1739,63 @@ PolyLine( Mat& img, const Point2l* v, int count, bool is_closed,
...
+void drawMarker(InputOutputArray _img, Point position, const Scalar& color, int markerType, int markerSize, int thickness, int line_type)
 {
     switch(markerType)
     {
     // The cross marker case
     case MARKER_CROSS:
-        line(img, Point(position.x-(markerSize/2), position.y), Point(position.x+(markerSize/2), position.y), color, thickness, line_type);
-        line(img, Point(position.x, position.y-(markerSize/2)), Point(position.x, position.y+(markerSize/2)), color, thickness, line_type);
+        line(_img, Point(position.x-(markerSize/2), position.y), Point(position.x+(markerSize/2), position.y), color, thickness, line_type);
+        line(_img, Point(position.x, position.y-(markerSize/2)), Point(position.x, position.y+(markerSize/2)), color, thickness, line_type);
```
alalek: Why do we need parameter name change (`img` -> `_img`)?
It is better to avoid unnecessary massive renames.

sturkmen72: OK i can revert this changes. most of the drawing functions have _img as parameter name. this is the reason i renamed.

modules/imgproc/src/drawing.cpp
```diff
@@ -1875,15 +1875,14 @@ void rectangle( InputOutputArray _img, Point pt1, Point pt2,
 }
 
 
-void rectangle( Mat& img, Rect rec,
+void rectangle( InputOutputArray _img, Rect rec,
```
alalek: Please change name back here too: `_img` => `img`

modules/imgproc/src/drawing.cpp
```diff
@@ -1875,13 +1875,12 @@ void rectangle( InputOutputArray _img, Point pt1, Point pt2,
...
 
 
-void rectangle( Mat& img, Rect rec,
+void rectangle( InputOutputArray img, Rect rec,
                 const Scalar& color, int thickness,
                 int lineType, int shift )
 {
     CV_INSTRUMENT_REGION()
 
-    CV_Assert( 0 <= shift && shift <= XY_SHIFT );
```
alalek: Why is assertion removed?

sturkmen72: maybe i am wrong but i think no need because of the same assertion exist in called function

https://github.com/sturkmen72/opencv/blob/master/modules/imgproc/src/drawing.cpp#L1855

PR #11295, opened 2018-04-11
Parse repeated values of ConvolutionParameter

dkurt:
Fix hyper-parameters parsing in case of repeated `kernel_size` or `pad` or `stride` from Caffe's `convolution_param`.

https://github.com/opencv/opencv/blob/10ba6a93a6fee952fb7812b28989eb209d4f49a1/modules/dnn/src/caffe/opencv-caffe.proto#L727-L731

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/457

related: http://answers.opencv.org/question/187725/when-loading-caffe-vgg-16-i-get-error-215-idx-1-size-1-idx-0-idx-size-in-function-get/

```
```


vpisarev:
:+1:

PR #11314, opened 2018-04-16
Msmf camera

terfendail:
WIP
### This pullrequest changes
Reworked CameraCapture implementation to use SourceReader


alalek:
I believe this PR should go into 3.4 branch (WinRT code branch removal is not critical - this code was not used for WinRT 8+)

modules/videoio/src/cap_msmf.cpp
```diff
@@ -79,3840 +81,1706 @@
...
+template <class T>
+class ComPtr
 {
 public:
-    ~DPO(void);
-    static DPO& getInstance();
-    void printOut(const wchar_t *format, ...);
-    void setVerbose(bool state);
-    bool verbose;
+    ComPtr() throw()
```
alalek: throw() is [deprecated](http://en.cppreference.com/w/cpp/language/noexcept_spec).

PR #11321, opened 2018-04-16
Support for HEVC in cudacodec::VideoReader #6054

### LINKED ISSUE #6054: Support for HEVC and VP9 in cudacodec::VideoReader
Hello,

  Current version of cudacodec::VideoReader does not support hevc/vp9 decode, Nvidia cards with GM206 and above has VP that supports both decode and encode of HEVC.

PaulinoAranda:
resolves #6054

### DIFF
--- a/modules/cudacodec/src/video_decoder.cpp
+++ b/modules/cudacodec/src/video_decoder.cpp
@@ -63,6 +63,11 @@ void cv::cudacodec::detail::VideoDecoder::create(const FormatInfo& videoFormat)
               cudaVideoCodec_VC1   == _codec ||
               cudaVideoCodec_H264  == _codec ||
               cudaVideoCodec_JPEG  == _codec ||
+              cudaVideoCodec_H264_SVC == _codec ||
+              cudaVideoCodec_H264_MVC == _codec ||
+              cudaVideoCodec_HEVC == _codec ||
+              cudaVideoCodec_VP8 == _codec ||
+              cudaVideoCodec_VP9 == _codec ||
               cudaVideoCodec_YUV420== _codec ||
               cudaVideoCodec_YV12  == _codec ||
               cudaVideoCodec_NV12  == _codec ||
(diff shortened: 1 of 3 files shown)


modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -1707,7 +1707,7 @@ bool CvVideoWriter_FFMPEG::writeFrame( const unsigned char* data, int step, int
     const int CV_STEP_ALIGNMENT = 32;
     const size_t CV_SIMD_SIZE = 32;
     const size_t CV_PAGE_MASK = ~(4096 - 1);
-    const unsigned char* dataend = data + ((size_t)height * step);
+    const uchar* dataend = data + ((size_t)height * step);
```
alalek: Why is this line changed?

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -2665,13 +2668,18 @@ bool InputMediaStream_FFMPEG::open(const char* fileName, int* codec, int* chroma
...
+                break;
+
             default:
                 return false;
             };
 
             switch (enc->pix_fmt)
             {
             case AV_PIX_FMT_YUV420P:
+	    case AV_PIX_FMT_YUVJ420P:
```
alalek: Indentation should be fixed here (use spaces only).

modules/videoio/src/ffmpeg_codecs.hpp
```diff
@@ -78,6 +78,7 @@ typedef struct AVCodecTag {
 } AVCodecTag;
 
 #if (LIBAVCODEC_VERSION_INT <= AV_VERSION_INT(54, 51, 100))
+#define AV_CODEC_ID_HEVC CODEC_ID_HEVC
```
alalek: Perhaps `AV_CODEC_ID_HEVC` is not defined for all mentioned ffmpeg versions. Some distributions like CentOS likes to use very old versions of FFmpeg API.

We need proper check here and in other places.

For example, we could reuse `AV_CODEC_ID_H265` [definition](https://github.com/FFmpeg/FFmpeg/blob/release/3.4/libavcodec/avcodec.h#L395):
```
#ifdef AV_CODEC_ID_H265
... use AV_CODEC_ID_H265 ...
#endif
```

modules/cudacodec/src/video_decoder.cpp
```diff
@@ -84,7 +89,7 @@ void cv::cudacodec::detail::VideoDecoder::create(const FormatInfo& videoFormat)
     createInfo_.ulNumDecodeSurfaces = FrameQueue::MaximumSize;
 
     // Limit decode memory to 24MB (16M pixels at 4:2:0 = 24M bytes)
-    while (createInfo_.ulNumDecodeSurfaces * videoFormat.width * videoFormat.height > 16 * 1024 * 1024)
+    while (createInfo_.ulNumDecodeSurfaces * videoFormat.width * videoFormat.height > 16 * 4096 * 4096)
```
alalek: Comment above should be updated too.

vpisarev:
closing this down - no reaction from the submitter

PR #11332, opened 2018-04-17
core(hal): improve v_select() SSE4.1+

alalek:
Make `v_select()` eliment-wise (instead of bit-wise).
`v_select()` `mask` is restricted to these values only: `0` or `~0` (`0xff`/`0xffff`/etc)
`mask` in accuracy test is updated.

Requires build option: `cmake -DCPU_BASELINE=SSE4_1 ...`

Similar semantic is available in OpenCL: https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/select.html

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin_cpp.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_cpp.hpp
@@ -1039,13 +1039,16 @@ template<typename _Tp, int n> inline bool v_check_any(const v_reg<_Tp, n>& a)
     return false;
 }
 
-/** @brief Bitwise select
+/** @brief Per-element select (blend operation)
 
-Return value will be built by combining values a and b using the following scheme:
-If the i-th bit in _mask_ is 1
-    select i-th bit from _a_
-else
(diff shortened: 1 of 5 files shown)


modules/core/include/opencv2/core/hal/intrin_cpp.hpp
```diff
@@ -1055,8 +1058,7 @@ template<typename _Tp, int n> inline v_reg<_Tp, n> v_select(const v_reg<_Tp, n>&
     for( int i = 0; i < n; i++ )
     {
         int_type m = Traits::reinterpret_int(mask.s[i]);
-        c.s[i] =  Traits::reinterpret_from_int((Traits::reinterpret_int(a.s[i]) & m)
-                                             | (Traits::reinterpret_int(b.s[i]) & ~m));
+        c.s[i] = m ? a.s[i] : b.s[i];
```
alalek: BTW, This semantic is compatible with OpenCL: https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/select.html

PR #11338, opened 2018-04-18
core: fix FP16 conversion with CV_DISABLE_OPTIMIZATION option

alalek:
Reproducer:
```
cmake -DCPU_BASELINE=AVX2 -DCV_DISABLE_OPTIMIZATION=ON ...
```

This build configuration is used to measure "compiler" optimizations on C++ code without manual intrinsic/vectorized code.

Related: #8899

```
buildworker:Linux x64=linux-2
```


tomoaki0705:
I see no problem.

Thank you !

PR #11344, opened 2018-04-18
 Fix some C++ compiler warnings (case fallthrough, negation of string)

eecsninja:
### This pullrequest changes

Replace {!"string"} with {false && "string"}
Replace case statement fallthroughs with break statements

The warnings appear in gcc 7.3.0 with -Wimplicit-fallthrough

### DIFF
--- a/modules/core/include/opencv2/core/affine.hpp
+++ b/modules/core/include/opencv2/core/affine.hpp
@@ -445,7 +445,7 @@ void cv::Affine3<T>::rotation(const cv::Mat& data)
         rotation(_rvec);
     }
     else
-        CV_Assert(!"Input matrix can only be 3x3, 1x3 or 3x1");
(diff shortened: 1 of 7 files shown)


alalek:
BTW, Did you try to remove this [flag](https://github.com/opencv/opencv/blob/3.4/cmake/OpenCVCompilerOptions.cmake#L122)?

eecsninja:
No I didn't. I'm building OpenCV with another build system, alongside dozens of other open-source packages. That's why I'd like to fix the code instead of relying on suppressing warnings.

vpisarev:
@eecsninja, this is not very good fix for not very good coding style; I'd suggest to replace CV_Assert(!"msg") with CV_Error(Error::StsError, msg).

eecsninja:
Closing to redo based on vpisarev's comment.

PR #11350, opened 2018-04-19
python: fix support of UI callbacks

### LINKED ISSUE #11348: Regression: python: UI callbacks support is broken
Highgui callbacks are not an unique instances.
Different callbacks can be binded simultaneously to multiple OpenCV "windows" under different names.

Related: #11217 #11206 

Reproducer:
```.py

import cv2
import numpy as np

def test():
    frame = np.zeros([255,255], dtype=np.uint8)

    cv2.namedWindow('frame1')
    cv2.namedWindow('frame2')

    def on_mouse1(event,x,y,flags,params):
        print(1, x, y)

    cv2.setMouseCallback('frame1', on_mouse1)

    def on_mouse2(event,x,y,flags,params):
        print(2, x, y)

    cv2.setMouseCallback('frame2', on_mouse2)

    cv2.imshow('frame1',frame)
    cv2.imshow('frame2',frame)

    cv2.waitKey(0)

test()
```

alalek:
resolves #11348

### DIFF
--- a/modules/python/src2/cv2.cpp
+++ b/modules/python/src2/cv2.cpp
@@ -27,6 +27,7 @@
 
 #include "pycompat.hpp"
 
+#include <map>
 
 static PyObject* opencv_error = 0;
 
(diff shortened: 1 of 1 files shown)


vpisarev:
:+1:

PR #11351, opened 2018-04-19
Enable some tests for clDNN plugin from Intel's Inference Engine

dkurt:
### This pullrequest changes

Enable Intel Inference Engine's OpenCL fp16 target for some of models:

| | CPU | OpenCL GPU, fp32 | OpenCL GPU, fp16 |
|----|----|---|---|
| GoogLeNet | 8.88ms | 17.06ms | 12.50ms |
| ResNet-50 | 19.66ms | 42.60ms | 29.35ms |
| SqueezeNet v1.1 | 2.43ms | 5.66ms | 4.73ms |

CPU: Intel&reg; Core&trade; i7-6700K CPU @ 4.00GHz x 8
GPU: Intel&reg; HD Graphics 530 (Skylake GT2)

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -1255,6 +1255,15 @@ struct Net::Impl
                     if (weightableLayer->_biases)
                         weightableLayer->_biases = convertFp16(weightableLayer->_biases);
                 }
+                else
+                {
+                    for (const auto& weights : {"weights", "biases"})
+                    {
+                        auto it = ieNode->layer->blobs.find(weights);
+                        if (it != ieNode->layer->blobs.end())
+                            it->second = convertFp16(it->second);
+                    }
+                }
             }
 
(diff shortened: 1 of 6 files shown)


modules/dnn/src/dnn.cpp
```diff
@@ -1255,6 +1255,15 @@ struct Net::Impl
                     if (weightableLayer->_biases)
                         weightableLayer->_biases = convertFp16(weightableLayer->_biases);
                 }
+                else
+                {
+                    for (const auto& weights : {"weights", "biases"})
```
alalek: Does `HAVE_INF_ENGINE` require C++11 compilation mode?

dkurt: @alalek, Thanks! I added
```cpp
if(NOT HAVE_CXX11)
    ie_fail()
endif()
```
before we try to find Inference Engine.

Issue #11356, opened 2018-04-20
dnn module Unknown layer type Shape in conv2d_transpose/Shape in function populateNet


alexionby:
##### System information (version)
- OpenCV => 3.4.1
- Operating System / Platform => Ubuntu 17.10
- Compiler => python 3.6.5 Anaconda / gcc version 7.2.0 (Ubuntu 7.2.0-8ubuntu3.2)
- Tensorflow => 1.7.0

##### Detailed description
I tried to implement Unet with Transposed convolutions and faced some troubles with them.
`
error: OpenCV(3.4.1) /feedstock_root/build_artefacts/opencv_1520722599420/work/opencv-3.4.1/modules/dnn/src/tensorflow/tf_importer.cpp:1582: error: (-2) Unknown layer type Shape in op sequential_1/conv2d_transpose_1/Shape in function populateNet`

##### Steps to reproduce

```
# coding: utf-8

# In[10]:
import tensorflow as tf
import numpy as np


# In[11]:
tf.__version__


# In[12]:
x = tf.placeholder(tf.float32, shape=(None, 288, 288, 1))

model = tf.keras.Sequential([
    tf.keras.layers.Conv2D(filters=32, kernel_size=(3,3), activation='relu', padding='same', input_shape=(288,288,1)),
    tf.keras.layers.Conv2D(filters=32, kernel_size=(3,3), activation='relu', padding='same'),
    tf.keras.layers.MaxPool2D(padding='same'),
    tf.keras.layers.Conv2D(filters=64, kernel_size=(3,3), activation='relu', padding='same'),
    tf.keras.layers.Conv2D(filters=64, kernel_size=(3,3), activation='relu', padding='same'),
    tf.keras.layers.Conv2DTranspose(filters=32, kernel_size=(2,2), activation='relu', strides=2, padding='same'),
    tf.keras.layers.Conv2D(filters=32, kernel_size=(3,3), activation='relu', padding='same'),
    tf.keras.layers.Conv2D(filters=32, kernel_size=(3,3), activation='relu', padding='same'),
    tf.keras.layers.Conv2D(filters=1, kernel_size=(1,1), activation='sigmoid')
])


# In[24]:
img = np.random.random((1,288,288,1)).astype('float32')
mask = np.random.random((1,288,288,1)).astype('float32')


# In[14]:
input_img = tf.placeholder(tf.float32, shape=(None, 288, 288, 1), name='input_img')
output = model(input_img)


# In[15]:
[...]

PR #11362, opened 2018-04-20
Fuse deconvolution layer subgraphs from Keras

dkurt:
resolves https://github.com/opencv/opencv/issues/11356

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/463

```
```


naserpiltan:
Hi there
I have a same problem 
```dnn module Unknown layer type Shape in conv2d_transpose/Shape in function populateNet ```
which OpenCv version can fix this problem??

>

PR #11364, opened 2018-04-21
Allow using more types with LUT, findNonZero, and PSNR.

Tetragramm:
Allow using more types with LUT, findNonZero, and PSNR. 
Fix comment error in Rotate.

This is the first step in making sure all (or as many as I can get done) functions work with all the datatypes the can possibly accept.  This takes the functions in [core/operations on arrays](https://docs.opencv.org/3.4.1/d2/de8/group__core__array.html#gad327659ac03e5fd6894b90025e6900a7) and updates them. I believe that this is everything in core that needs to be updated, though if someone finds something else, just let me know.

```
buildworker:Linux x64=linux-1
```

### DIFF
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -1079,7 +1079,7 @@ and the rows and cols are switched for ROTATE_90_CLOCKWISE and ROTATE_90_COUNTER
 @param rotateCode an enum to specify how to rotate the array; see the enum #RotateFlags
 @sa transpose , repeat , completeSymm, flip, RotateFlags
 */
-CV_EXPORTS_W void rotate(InputArray src, OutputArray dst, RotateFlags rotateCode);
+CV_EXPORTS_W void rotate(InputArray src, OutputArray dst, int rotateCode);
 
 /** @brief Fills the output array with repeated copies of the input array.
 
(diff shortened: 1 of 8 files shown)


Tetragramm:
Added tests.
New LUT does not support 32S for precisely that reason.
As for binary size, is there a way to get that out of the build bot?  I can always do the comparison on my builds, but the compiler flags may be different.
Performance tests coming.

vpisarev:
that's a serious code expansion. Do we really need all this functionality? Can we make it more compact by taking into account the fact that for some of the types the code will actually be the same (e.g. findNonZero for 8u and 8s). Then, speaking of findNonZero, why not just use `findNonZero(myarray != 0., nz_locs_array)`, where myarray is single-channel array of an arbitrary depth?

Tetragramm:
It depends on whether you are worried about binary size or lines of code.  It would prevent the creation of two template specializations, but only 4 lines of code.  Even then, the template function isn't very large.

As to using != 0, that allocates and deallocates a Mat, and all the extra comparisons, reads, and writes.

I intend on continuing the expansion of data type support (particularly CV_16U) in other modules, and for that, having all the basic functions (LUT specifically) compatible is important.

vpisarev:
superseded by #11837; for LUTs, I think, we need some better solutions, like LUT with interpolation (e.g. pass the grid and the value range (e.g. 0 to 65535 range and 16x16x16 grid for 16-bit RGB to 16-bit RGB 3D transformation)). In fact, we already have something like that internally for some of our color conversion functions

PR #11376, opened 2018-04-23
core: CV_Error with set_terminate() on Windows

alalek:
To dump contents of the last OpenCV error.

Related #11357


alalek:
:+1:

Issue #10950, opened 2018-02-27
protobuf is built even when the modules needing it are turned off


dustsigns:
##### System information (version)
- OpenCV => 3.4.1 and master
- Operating System / Platform => Ubuntu 16.04
- Compiler => g++ 5.4.0

##### Detailed description

Since OpenCV 3.4.1, protobuf is built even when there are no modules enabled which require it. In 3.4.0, protobuf was not built if not needed.

##### Steps to reproduce

Consider the following build configuration on a system that does not have libprotobuf installed:

`cmake -D OPENCV_EXTRA_MODULES_PATH=../opencv_contrib-3.4.1/modules -D BUILD_LIST=calib3d,features2d,flann,highgui,imgcodecs,imgproc,ml,objdetect,sfm,stereo,video,videoio,viz,xfeatures2 -D BUILD_ITT=OFF -D WITH_1394=OFF -D WITH_GSTREAMER=OFF -D WITH_IPP=OFF -D WITH_JASPER=OFF -D WITH_WEBP=OFF -D WITH_OPENEXR=OFF -D WITH_PVAPI=OFF -D WITH_GIGEAPI=OFF -D WITH_QT=ON -D WITH_TIFF=OFF -D WITH_OPENCL=OFF -D BUILD_opencv_apps=OFF -D BUILD_DOCS=OFF -D BUILD_PACKAGE=OFF -D BUILD_PERF_TESTS=OFF -D BUILD_TESTS=OFF -D BUILD_FAT_JAVA_LIB=OFF -D ENABLE_PRECOMPILED_HEADERS=OFF -D CMAKE_BUILD_TYPE=RELEASE ../opencv-3.4.1/`

The output after configuration in OpenCV 3.4.1 is:
```
[..]
-- Caffe:   NO
-- Protobuf:   NO
-- Glog:   YES
[..]
-- Performing Test COMPILER_SUPPORTS_CXX11 - Success
-- The protocol buffer compiler is not found (PROTOBUF_PROTOC_EXECUTABLE='')
-- Module opencv_ovis disabled because OGRE3D was not found
[..]
--   Other third-party libraries:
--     Lapack:                      NO
--     Eigen:                       YES (ver 3.2.92)
--     Custom HAL:                  NO
--     Protobuf:                    build (3.5.1)
[..]
```

Note that protobuf is not found and forced to be built, even though it is not needed.

Trying to use the same configuration in OpenCV 3.4.0 gives the desired output, i.e., protobuf is not built:

[...]

PR #11377, opened 2018-04-23
protobuf build documenting: move the WITH_PROTOBUF option to the top …

Turim:
It seems somewhat unobvious that a user can configure protobuf within it's finder - my proposal is to move it to the top level.

P.S. IMHO may ease troubles like #10950 .


alalek:
Switched target branch to [3.4](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) and rebased this patch.

BTW, Next time please use proper name for the source branch instead of **master** (like `cmake_protobuf_options`).

PR #11380, opened 2018-04-23
Faster interleave deinterleave

vpisarev:
Improved performance of `v_load_deinterleave()` and `v_store_interleave()` intrinsics on 8-bit 3-channel data when using SSSE3 instruction set. `v_load_deinterleave()` is ~1.5x faster, `v_store_interleave()` is up to 2x faster.

Requires `-DCPU_BASELINE=SSSE3` (or `-DCPU_BASELINE=SSE4_1` or `-DCPU_BASELINE=SSE4_2`)


vpisarev:
:+1:

PR #11384, opened 2018-04-24
ml: refactor non-virtual methods

### LINKED ISSUE #8122: Non virtual methods in ML interface clases
There is a number of non virtual methods in interfaces. Such method have appeared because ABI compatibility checker prohibits changes in v-tables.
This causes creation of unnecessary implementations with implicit type casting:
```
CV_WRAP Mat getUncompressedSupportVectors() const; //supposed to be virtual
...
Mat SVM::getUncompressedSupportVectors() const
{
    const SVMImpl* this_ = dynamic_cast<const SVMImpl*>(this);
    if(!this_)
        CV_Error(Error::StsNotImplemented, "the class is not SVMImpl");
    return this_->getUncompressedSupportVectors_();
}
```

berak:
resolves #8122
### This pullrequest changes

make all SVM/ANN_MLP/TrainData/RTrees methods virtual and remove indirections.

this breaks ABI compatibility, and should go into upcoming 4.0


berak:
based on wrong branch , missing https://github.com/opencv/opencv/commit/2385a5870e1ad7b6fe689329c90a7d6e973acb4b

PR #11390, opened 2018-04-25
Deep learning text detection sample 

dkurt:
### This pullrequest changes

Added sample with deep learning text detection using EAST: An Efficient and Accurate Scene Text Detector.

paper: https://arxiv.org/abs/1704.03155v2
implementation: https://github.com/argman/EAST

* Move custom layers samples from `samples/cpp/tutorial_code/dnn/custom_layers.cpp` to `samples/dnn/custom_layers.hpp`
* Non-maximum suppression for rotated rectangles

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/467

res

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/src/nms.cpp
+++ b/modules/dnn/src/nms.cpp
@@ -8,6 +8,8 @@
 #include "precomp.hpp"
 #include "nms.inl.hpp"
 
+#include <opencv2/imgproc.hpp>
+
 namespace cv
 {
 namespace dnn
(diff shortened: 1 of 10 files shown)


modules/dnn/src/nms.cpp
```diff
@@ -28,6 +30,28 @@ void NMSBoxes(const std::vector<Rect>& bboxes, const std::vector<float>& scores,
     NMSFast_(bboxes, scores, score_threshold, nms_threshold, eta, top_k, indices, rectOverlap);
 }
 
+static inline float rotatedRectIOU(const RotatedRect& a, const RotatedRect& b)
+{
+    std::vector<Point2f> inter, hull;
+    int res = rotatedRectangleIntersection(a, b, inter);
```
vpisarev: I wonder if we can have a function like rotatedRectangleIntersectionArea(), which will compute area of the intersection without storing the coordinates explicitly. Another question - since a rotated rectangle is a convex polygon, its intersection is a convex polygon as well. Why is convexHull() called then?

samples/dnn/text_detection.cpp
```diff
@@ -0,0 +1,169 @@
...
+        double t = net.getPerfProfile(layersTimes) / freq;
+        std::string label = format("Inference time: %.2f ms", t);
+        putText(frame, label, Point(0, 15), FONT_HERSHEY_SIMPLEX, 0.5, Scalar(0, 255, 0));
+
+        imshow(kWinName, frame);
+    }
+    return 0;
+}
+
+void decode(const Mat& scores, const Mat& geometry, float scoreThresh,
```
vpisarev: probably, such a function should be put to objdetect module (or dnn module).

modules/dnn/src/nms.cpp
dkurt: @vpisarev, that's right but now `rotatedRectangleIntersection` returns just intersection points without contour orientation.
In example, 
```cpp
std::vector<Point2f> inter;
RotatedRect r1(Point2f(110.48, 272.86), Size2f(34.1097, 12.6564), 0);
RotatedRect r2(Point2f(109.976, 274.791), Size2f(36.023, 13.1669), 0);
rotatedRectangleIntersection(r1, r2, inter);
for (int i = 0; i < inter.size(); ++i)
    std::cout << inter[i].x << " " << inter[i].y << std::endl;
```
gives
```
93.4252 268.208
127.535 268.208
93.4252 279.188
127.535 279.188

2 ---- 3
|      |
0 ---- 1
```
So it just orient them. Otherwise `contourArea` returns `0`.

dkurt: @vpisarev, Done. `rotatedRectangleIntersection` returns an ordered set of points now.

samples/dnn/text_detection.cpp
dkurt: @vpisarev. The thing is that this decoding procedure is too specific for this network. In example, x4 final feature maps reduction comparing to input images:

```cpp
// Decode a prediction.
// Multiple by 4 because feature maps are 4 time less than input image.
float offsetX = x * 4.0f, offsetY = y * 4.0f;
...
```

dkurt:
@alalek, thank you!

vpisarev:
looks great, let's finally merge it! :+1:

yiran-THU:
@dkurt 
May I ask what is the speed of this implementation?

PR #11400, opened 2018-04-26
ipp: disable normL1(a, b, mask) 16UC3

### LINKED ISSUE #11399: core: IPP AVX2 bug in 16UC3 norm(a, b, mask)
Norm type: L1
IPP versions: IPPICV (2017u2), standalone (2017u2, 2018u2 and others)

AVX2 code - failed
SSE42 - pass

Reproducer (regression test): https://github.com/alalek/opencv/commits/ipp_norm_L1_16UC3_mask

```
modules/core/test/test_arithm.cpp:2102: Failure
      Expected: (double)20*cn
      Which is: 60
To be equal to: cv::norm(a, b, NORM_L1, mask)
      Which is: 6
```

alalek:
resolves #11399

Condition:
- AVX2
- sz.width < 16

```
allow_multiple_commits=1
```


lupustr3:
+1

alalek:
:+1:

alalek:
Looks like AVX512 case should be fixed too.

PR #11401, opened 2018-04-26
Optimize MultiBandBlender to run faster

exoson:
### This pullrequest changes
Blending using CUDA has been optimized quite a bit. New use case made possible. When blending multiple stitches from source images in which positions, ROIs and masks don't change in between stitches, e.g. when blending multiple video streams together from a set camera rig, some unnecessary recalculation and reinitialization is not done if not calling prepare in between the blends. This way one can call prepare once and after that call feed and blend over again if one doesn't want to change position, ROI or masks of the source images.
Another addition is that the blend function accepts GpuMat as destination InputOutputArray. If GpuMat is provided the images won't be downloaded from the GPU during blending and can be used for something else on the GPU right after.
On my computer with Intel i7-6700 Nvidia Quadro K620 blending two 512x512 images together is approximately 2.74 times faster after the commit compared to before the commit.
I can possibly provide an simple example code for the new use case if deemed appropriate.

```
force_builders=Custom
```

### DIFF
--- a/modules/stitching/include/opencv2/stitching/detail/blenders.hpp
+++ b/modules/stitching/include/opencv2/stitching/detail/blenders.hpp
@@ -145,6 +145,18 @@ private:
 #if defined(HAVE_OPENCV_CUDAARITHM) && defined(HAVE_OPENCV_CUDAWARPING)
     std::vector<cuda::GpuMat> gpu_dst_pyr_laplace_;
     std::vector<cuda::GpuMat> gpu_dst_band_weights_;
+    std::vector<Point> gpu_tl_points_;
+    std::vector<cuda::GpuMat> gpu_imgs_with_border_;
+    std::vector<std::vector<cuda::GpuMat> > gpu_weight_pyr_gauss_vec_;
+    std::vector<std::vector<cuda::GpuMat> > gpu_src_pyr_laplace_vec_;
+    std::vector<std::vector<cuda::GpuMat> > gpu_ups_;
+    cuda::GpuMat gpu_dst_mask_;
+    cuda::GpuMat gpu_mask_;
+    cuda::GpuMat gpu_img_;
+    cuda::GpuMat gpu_weight_map_;
+    cuda::GpuMat gpu_add_mask_;
+    int gpu_feed_idx_;
+    bool gpu_initialized_;
 #endif
 };
 
--- a/modules/stitching/src/blenders.cpp
+++ b/modules/stitching/src/blenders.cpp
@@ -221,6 +221,7 @@ MultiBandBlender::MultiBandBlender(int try_gpu, int num_bands, int weight_type)
 
 #if defined(HAVE_OPENCV_CUDAARITHM) && defined(HAVE_OPENCV_CUDAWARPING)
     can_use_gpu_ = try_gpu && cuda::getCudaEnabledDeviceCount();
+    gpu_feed_idx_ = 0;
 #else
     (void) try_gpu;
     can_use_gpu_ = false;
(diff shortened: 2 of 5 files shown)


alalek:
This patch should [go](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) into 3.4 branch. So, please:

modules/stitching/include/opencv2/stitching/detail/blenders.hpp
```diff
@@ -145,6 +145,18 @@ class CV_EXPORTS MultiBandBlender : public Blender
...
+    std::vector<std::vector<cuda::GpuMat>> gpu_weight_pyr_gauss_vec_;
+    std::vector<std::vector<cuda::GpuMat>> gpu_src_pyr_laplace_vec_;
+    std::vector<std::vector<cuda::GpuMat>> ups_;
+    cuda::GpuMat gpu_dst_mask_;
+    cuda::GpuMat gpu_mask_;
+    cuda::GpuMat gpu_img_;
+    cuda::GpuMat weight_map_;
+    cuda::GpuMat add_mask_;
+    int feed_idx;
+    int initialized;
```
alalek: Please use `gpu_` prefix for all of these vars.

modules/stitching/src/blenders.cpp
```diff
@@ -248,6 +249,15 @@ void MultiBandBlender::prepare(Rect dst_roi)
...
     if (can_use_gpu_)
     {
+        initialized = false;
+        feed_idx = 0;
+
+        tl_points_ = std::vector<Point>();
+        gpu_weight_pyr_gauss_vec_ = std::vector<std::vector<cuda::GpuMat>>();
+        gpu_src_pyr_laplace_vec_ = std::vector<std::vector<cuda::GpuMat>>();
+        ups_ = std::vector<std::vector<cuda::GpuMat>>();
+        imgs_with_border_ = std::vector<cuda::GpuMat>();
```
alalek: Why we need these "empty" assignments? Consider using `.clear()` instead.

modules/stitching/include/opencv2/stitching/detail/blenders.hpp
exoson: Done

modules/stitching/src/blenders.cpp
exoson: No reason I think. Changed to clear.

modules/stitching/include/opencv2/stitching/detail/blenders.hpp
```diff
@@ -145,6 +145,18 @@ class CV_EXPORTS MultiBandBlender : public Blender
 #if defined(HAVE_OPENCV_CUDAARITHM) && defined(HAVE_OPENCV_CUDAWARPING)
     std::vector<cuda::GpuMat> gpu_dst_pyr_laplace_;
     std::vector<cuda::GpuMat> gpu_dst_band_weights_;
+    std::vector<Point> gpu_tl_points_;
+    std::vector<cuda::GpuMat> gpu_imgs_with_border_;
+    std::vector<std::vector<cuda::GpuMat>> gpu_weight_pyr_gauss_vec_;
+    std::vector<std::vector<cuda::GpuMat>> gpu_src_pyr_laplace_vec_;
+    std::vector<std::vector<cuda::GpuMat>> gpu_ups_;
```
alalek: Please [add](http://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/392) space between `>>`: `>{space}>`

exoson: Done

PR #11410, opened 2018-04-27
core: disabled IPP AVX512 normL1(a, b, mask)

### LINKED ISSUE #11399: core: IPP AVX2 bug in 16UC3 norm(a, b, mask)

alalek:
for cv::Mat with type=16UC3 and width < 16

related #11400
resolves #11399

```
buildworker:Linux x64=linux-3
```


alalek:
:+1:

PR #11415, opened 2018-04-28
imgcodecs iface: add imquery() to get image size & decoder_type from the header

Turim:
As a possible follow-up due to #8511

### DIFF
--- a/modules/imgcodecs/include/opencv2/imgcodecs.hpp
+++ b/modules/imgcodecs/include/opencv2/imgcodecs.hpp
@@ -175,6 +175,36 @@ Currently, the following file formats are supported:
 */
 CV_EXPORTS_W Mat imread( const String& filename, int flags = IMREAD_COLOR );
 
+enum ImageDecoderType
+{
+	IMAGE_DECODER_BMP = 0,
+	IMAGE_DECODER_DICOM,
+	IMAGE_DECODER_EXR,
+	IMAGE_DECODER_GDAL,
+	IMAGE_DECODER_HDR,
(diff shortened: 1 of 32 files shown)


alalek:
Thank you for contribution!

> enum ImageDecoderType

IMHO, string is more consistent to return such information.
I believe this information can be used for debugging purposes (for example, to check for missing/disabled decoder).
Also, extended information can be provided:
- GDAL has internal decoder for BMP format, so "GDAL/BMP" can be used here. For OpenCV native decoded we could use just "BMP"
- JPEG2000 images can be decoded by libjasper (current backend) or openjpeg (not integrated yet).

Could you provide some examples/usecases for proposed enumeration?

Turim:
AFAIK enum's are pretty common for the imgcodecs interface.
I've tried to provide some samples via tests modification.
As for enums - for a user to be able to filter out images by the format also.
Like
I.
`
struct FileInfo
{
  int weight;
  int height;
  int fileType;
}

FileInfo GetFileInfo(filename)
{
    int decoderType;
    const Size size = imquery(filename, &decoderType);
    if (size.empty())
       throw runtime_error("imquery error");
    return FileInfo(size.width, size.height, decoderType);
}
`
II.
`
cv::Mat LoadImage(filename)
{
    int decoderType;
    const Size imageSize = imquery(filename, &decoderType)
    if (imageSize.empty())
       throw runtime_error("imquery failure");
   if (decoderType != IMAGE_DECODER_PNG)
      throw runtime_error("the file format has been filtered out");
   return imread(filename);
}
`

PR #11416, opened 2018-04-28
Add assertion for empty mat

take1014:
Empty Mat should not have any type information, so I added assertion for this.


vpisarev:
this is superseded by #11634

PR #11417, opened 2018-04-28
imgcodecs cmake: the option to customize supported formats list

Turim:
(WITH_HDR, WITH_SUNRASTER, WITH_PXM)

There were (and may be still exist) vulnerabilities within decoders.
The proposal is "if JPEG is could be excluded let's provide an option to exclude even more specific image formats."


alalek:

This patch should [go](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) into 3.4 branch. So, please:

cmake/OpenCVFindLibsGrfmt.cmake
```diff
@@ -218,14 +218,6 @@ if(WITH_GDAL)
     endif()
 endif()
 
-if (WITH_GDCM)
-  find_package(GDCM QUIET)
```
alalek: This block is still required.

vpisarev:
I think, at some point we need to normalize our CMake flags, 'cause now it's a real mess. As a part of this process, I suggest to use `WITH_IMGCODEC_*` notation, like `WITH_IMGCODEC_TIFF`

alalek:
@vpisarev This will be done as a separate patch.

cmake/OpenCVFindLibsGrfmt.cmake
```diff
@@ -218,14 +218,6 @@ if(WITH_GDAL)
...
-    ocv_clear_vars(GDCM_VERSION GDCM_LIBRARIES)
-  else()
-    set(HAVE_GDCM YES)
-    # include(${GDCM_USE_FILE})
-    set(GDCM_LIBRARIES gdcmMSFF) # GDCM does not set this variable for some reason
-  endif()
-endif()
+set(HAVE_HDR		WITH_HDR)
+set(HAVE_SUNRASTER	WITH_SUNRASTER)
+set(HAVE_PXM		WITH_PXM)
```
alalek: Modern (5+ years) CMake doesn't expand `WITH_` variables, so these `HAVE_` vars are always true.

modules/imgcodecs/src/loadsave.cpp
```diff
@@ -131,8 +131,10 @@ struct ImageCodecInitializer
         decoders.push_back( makePtr<BmpDecoder>() );
         encoders.push_back( makePtr<BmpEncoder>() );
 
+	#ifdef HAVE_HDR
```
alalek: No tabs please

Turim: I beg your pardon. (from what I see - was fixed already, thanks!)

PR #11420, opened 2018-04-28
cmake: update 'git describe' information

alalek:


alalek:
:+1:

cmake/OpenCVUtils.cmake
```diff
@@ -1620,3 +1620,40 @@ if(NOT CMAKE_VERSION VERSION_LESS 3.1)
...
+      execute_process(COMMAND "${GIT_EXECUTABLE}" describe --tags --always --dirty --match "[0-9].[0-9].[0-9]*" --exclude "[^-]*-cvsdk"
+        WORKING_DIRECTORY "${path}"
+        OUTPUT_VARIABLE ${var_name}
+        RESULT_VARIABLE GIT_RESULT
+        ERROR_QUIET
+        OUTPUT_STRIP_TRAILING_WHITESPACE
+      )
+      if(NOT GIT_RESULT EQUAL 0)  # --exclude is not supported by 'git'
+        # match only tags with complete OpenCV versions (ignores -alpha/-beta/-rc suffixes)
+        execute_process(COMMAND "${GIT_EXECUTABLE}" describe --tags --always --dirty --match "[0-9].[0-9]*[0-9]"
```
tomoaki0705: @alalek just to make sure about this line

On master as of today, after cmake, we get this output.
```
-- General configuration for OpenCV 4.0.0-pre =====================================
--   Version control:               3.4.3-844-g5459c11
```

I thought this is a bug because it's saying ```3.4.3``` on ```master```, it should be "[0-9].[0-9]**.**[0-9]**\***" here

Now reading the comment, I think 
 - this is an expected behavior
 - this will disappear when we get a official 4.0 release
Am I correct ?

alalek: This will be fixed after 4.0 release ("4.0.0" tag).

BTW, Ubuntu 18.04 shows something like this (newer git with `--exclude` support):
```
-- General configuration for OpenCV 4.0.0-pre =====================================
--   Version control:               4.0.0-beta-208-g75c87a5e5
```

tomoaki0705: Thank you.
It seems to upgrade to Ubuntu 18.04.

```
$ git --version
git version 2.7.4
$ lsb_release  -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.4 LTS
Release:        16.04
Codename:       xenial
```

PR #11425, opened 2018-04-29
FIX INF_ENGINE RELU ERROR

zuoshaobo:
fix bug that:
when use intel inference engine,the relu activation ：
max(0,x)+negative_slope×min(0,x)
in darknet ,the negative_slope is eq 0.1, but this value doesn't pass to inferenceengin.

### DIFF
--- a/modules/dnn/src/layers/elementwise_layers.cpp
+++ b/modules/dnn/src/layers/elementwise_layers.cpp
@@ -329,6 +329,7 @@ struct ReLUFunctor
     {
         lp.type = "ReLU";
         std::shared_ptr<InferenceEngine::ReLULayer> ieLayer(new InferenceEngine::ReLULayer(lp));
+	ieLayer->negative_slope = 0.1f;
         return ieLayer;
     }
 #endif  // HAVE_INF_ENGINE
--- a/modules/dnn/src/layers/elementwise_layers.cpp
+++ b/modules/dnn/src/layers/elementwise_layers.cpp
@@ -329,7 +329,7 @@ struct ReLUFunctor
     {
         lp.type = "ReLU";
         std::shared_ptr<InferenceEngine::ReLULayer> ieLayer(new InferenceEngine::ReLULayer(lp));
-	ieLayer->negative_slope = 0.1f;
+	ieLayer->negative_slope = slope;
         return ieLayer;
     }
 #endif  // HAVE_INF_ENGINE
--- a/modules/dnn/src/layers/elementwise_layers.cpp
+++ b/modules/dnn/src/layers/elementwise_layers.cpp
@@ -329,7 +329,7 @@ struct ReLUFunctor
     {
         lp.type = "ReLU";
         std::shared_ptr<InferenceEngine::ReLULayer> ieLayer(new InferenceEngine::ReLULayer(lp));
-	ieLayer->negative_slope = slope;
+        ieLayer->negative_slope = slope;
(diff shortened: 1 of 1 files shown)


modules/dnn/src/layers/elementwise_layers.cpp
```diff
@@ -329,6 +329,7 @@ struct ReLUFunctor
     {
         lp.type = "ReLU";
         std::shared_ptr<InferenceEngine::ReLULayer> ieLayer(new InferenceEngine::ReLULayer(lp));
+	ieLayer->negative_slope = 0.1f;
```
dkurt: > in darknet ,the negative_slope is eq 0.1, but this value doesn't pass to inferenceengin.

We cannot forget about the rest of supported frameworks so you need to use `slope` member:
```cpp
ieLayer->negative_slope = slope;
```

Anyway darknet's relu has no negative slope at all:

https://github.com/pjreddie/darknet/blob/8215a8864d4ad07e058acafd75b2c6ff6600b9e8/src/activations.h#L34

May I ask you to provide a source?

zuoshaobo: sorry,it is leakyrelu:
static inline float leaky_activate(float x){return (x>0) ? x : .1*x;}
https://github.com/pjreddie/darknet/blob/8215a8864d4ad07e058acafd75b2c6ff6600b9e8/src/activations.h#L38
as you see, the default value is 0.1

zuoshaobo: @dkurt ,But just use it by default backend, in gpu mode(intel inference engine),it's not set the slope value look:
defalut backend-> line 879
Ptr<ReLULayer> ReLULayer::create(const LayerParams& params)
{
    float negativeSlope = params.get<float>("negative_slope", 0.f);
    Ptr<ReLULayer> l(new ElementWiseLayer<ReLUFunctor>(ReLUFunctor(negativeSlope)));
    l->setParamsFrom(params);
    l->negativeSlope = negativeSlope;

    return l;
}
intel inference engine->line 328
#ifdef HAVE_INF_ENGINE
    InferenceEngine::CNNLayerPtr initInfEngine(InferenceEngine::LayerParams& lp)
    {
        lp.type = "ReLU";
        std::shared_ptr<InferenceEngine::ReLULayer> ieLayer(new InferenceEngine::ReLULayer(lp));
        return ieLayer;
    }
#endif  // HAVE_INF_ENGINE

dkurt: @zuoshaobo, that is why I asked you to replace
```cpp
ieLayer->negative_slope = 0.1f;
```
to 
```
ieLayer->negative_slope = slope;
```
both are correct for darknet networks but not for cases with `negative_slope != 0.1`.

zuoshaobo: @dkurt ok, i had  do it now,3q

dkurt: @zuoshaobo, Thank you! Could you please also add an extra space to align that line with the rest? If it possible, squash all the commits into one.

zuoshaobo:
@dkurt  
build error？
RemoveDirectory_1 Delete failed. ( 10 secs )
- no logs -

dkurt:
@zuoshaobo, that's OK. Let's just wait until buildbot starts work as usual.

PR #11431, opened 2018-05-01
Fix some issues with C API

### LINKED ISSUE #10963: vlc FTBFS with opencv-3.4.1
- OpenCV => 3.4.1
- Compiler => gcc-7-20180222 (a 7.3.1 snapshot)

vlc-3.0.1 fails to build against  opencv-3.4.1, because the C compiler produces and error at line 485 of /usr/include/opencv2/core/cvdef.h. The error message is:
           /usr/include/opencv2/core/cvdef.h:485:1: error: unknown type name 'namespace'  

This is because the C compiler is trying to compile C++ code.

##### Steps to reproduce
1. build and install opencv-3.4.1
2. configure vlc-3.0.1 to include opencv and then run make

##### Proposed fix
The following patch to opencv allows the build of vlc to complete:

--- opencv-3.4.1/modules/core/include/opencv2/core/cvdef.h.orig	2018-02-28 10:16:45.000320632 +0000
+++ opencv-3.4.1/modules/core/include/opencv2/core/cvdef.h	2018-02-28 10:34:30.852346927 +0000
@@ -454,6 +454,7 @@ Cv64suf;
 
 
 // Integer types portatibility
+#if defined __cplusplus
 #ifdef OPENCV_STDINT_HEADER
 #include OPENCV_STDINT_HEADER
 #else
@@ -494,6 +495,9 @@ typedef ::uint64_t uint64_t;
 }
 #endif
 #endif
+#else
+#include <stdint.h>
+#endif
 
 
 //! @}

With the patched opencv installed, I have also built the frei0r-plugin-1.6.0, gstreamer-plugins-bad-1.12.4 and LiVES-2.8.7, all of which use opencv (although I don't know whether they failed against an unpatched version)

pjreddie:
resolves #10963

### This pullrequest changes
two header files to be C compliant (not use namespaces) and window.cpp in highgui to have the correct return type. Compiles and passes all tests.

### DIFF
--- a/modules/highgui/src/window.cpp
+++ b/modules/highgui/src/window.cpp
@@ -165,7 +165,7 @@ CV_IMPL double cvGetWindowProperty(const char* name, int prop_id)
     }
 }
 
-cv::Rect cvGetWindowImageRect(const char* name)
+CvRect cvGetWindowImageRect(const char* name)
 {
     if (!name)
         return cv::Rect(-1, -1, -1, -1);
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -461,6 +461,7 @@ using std::uint64_t;
 }
 #else
 #include <stdint.h>
+#if defined __cplusplus
 namespace cv {
 typedef ::int8_t int8_t;
 typedef ::uint8_t uint8_t;
@@ -471,6 +472,7 @@ typedef ::uint32_t uint32_t;
 typedef ::int64_t int64_t;
 typedef ::uint64_t uint64_t;
 }
+#endif
 #endif
 #endif
 
--- a/modules/highgui/include/opencv2/highgui/highgui_c.h
+++ b/modules/highgui/include/opencv2/highgui/highgui_c.h
@@ -136,7 +136,7 @@ CVAPI(void) cvSetWindowProperty(const char* name, int prop_id, double prop_value
 CVAPI(double) cvGetWindowProperty(const char* name, int prop_id);
 
 /* Get window image rectangle coordinates, width and height */
-CVAPI(cv::Rect)cvGetWindowImageRect(const char* name);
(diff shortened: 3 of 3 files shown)


modules/highgui/src/window.cpp
```diff
@@ -165,7 +165,7 @@ CV_IMPL double cvGetWindowProperty(const char* name, int prop_id)
     }
 }
 
-cv::Rect cvGetWindowImageRect(const char* name)
+CvRect cvGetWindowImageRect(const char* name)
 {
     if (!name)
         return cv::Rect(-1, -1, -1, -1);
```
StevenPuttemans: You will need a CvRect here too ;)

pjreddie: Where? In the actual function? I think it works ok without changing the cv::Rect on line 171, it compiles fine. I imagine cv::Rect and CvRect are the same under the hood, one just isn't usable in C. But Let me know if I'm mistaken!

pjreddie:
I'm not sure why the checks haven't completed, does anyone know what builds it's waiting for or how I can make them happen?

StevenPuttemans:
@pjreddie I think this is just a warning that not all buildbots are ran, because http://pullrequest.opencv.org/#/summary/ clearly shows that all necessary builds completed. I guess it is just waiting for a dev to respond now!

alalek:
1. Wrong target branch is selected. master is for OpenCV 4.0 and all "C" stuff would be removed (at least from public API). You should select branch 3.4
2. `cvGetWindowImageRect()` should be removed completely. It has been added for latest release only, and it should not be here. Also replacing just return type is bad thing
3. Does it really compiles? We currently has no C samples to check this. Simple examples still doesn't link in non-release mode (due cvRound symbol, without `-O2`).
4. This compiler macro can workaround issue with "namespace":
    ```
    gcc -DOPENCV_STDINT_HEADER=\"stdint.h\" ...
    ```

alalek:
Replaced by #11605 (merged)

PR #11443, opened 2018-05-03
Fix Python string formatting

cclauss:
### This pullrequest changes
Invalid Python string formatting into valid Python string formatting.  The current lines will raise [NameError](https://docs.python.org/3/library/exceptions.html#NameError) on 's' instead of raising the expected __Fail()__.

Also removed a couple of superfluous trailing semicolons.  (Python != JavaScript)

Discovered via flake8 testing of https://github.com/opencv/opencv

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./platforms/android/build_sdk.py:21:59: F821 undefined name 's'
            raise Fail("Child was terminated by signal:" %s -retcode)
                                                          ^
./platforms/js/build_js.py:17:59: F821 undefined name 's'
            raise Fail("Child was terminated by signal:" %s -retcode)
                                                          ^
```

### DIFF
--- a/platforms/android/build_sdk.py
+++ b/platforms/android/build_sdk.py
@@ -18,7 +18,7 @@ def execute(cmd, shell=False):
         log.info('Executing: ' + ' '.join(cmd))
         retcode = subprocess.call(cmd, shell=shell)
         if retcode < 0:
-            raise Fail("Child was terminated by signal:" %s -retcode)
+            raise Fail("Child was terminated by signal: %s" % -retcode)
         elif retcode > 0:
             raise Fail("Child returned: %s" % retcode)
     except OSError as e:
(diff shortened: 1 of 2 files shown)


platforms/android/build_sdk.py
```diff
@@ -18,7 +18,7 @@ def execute(cmd, shell=False):
         log.info('Executing: ' + ' '.join(cmd))
         retcode = subprocess.call(cmd, shell=shell)
         if retcode < 0:
-            raise Fail("Child was terminated by signal:" %s -retcode)
```
dkurt: I think we need to keep a `-` sign here to print a positive error code.

cclauss: Change made.  I stuck with __%s__ to be consistent with the other error messages just below these lines.

PR #11462, opened 2018-05-04
Load CascasdeClassifier from cv::FileStorage.

edrosten:
### This pullrequest changes

CascadeClassifier currently allows loading from files using the load method. Internally this opens a FileStorage object and loads from that. This PR exposes that functionality to the public interface. The purpose is to allow the load() to work on all sources supported by FileStorage, for example in memory storage.

### DIFF
--- a/modules/core/include/opencv2/core/persistence.hpp
+++ b/modules/core/include/opencv2/core/persistence.hpp
@@ -95,6 +95,7 @@ the extension of the opened file, ".xml" for XML files, ".yml" or ".yaml" for YA
 JSON.
  */
 typedef struct CvFileStorage CvFileStorage;
+typedef struct CvMemStorage CvMemStorage;
 typedef struct CvFileNode CvFileNode;
 typedef struct CvMat CvMat;
 typedef struct CvMatND CvMatND;
@@ -1344,6 +1345,19 @@ inline FileNode FileNodeIterator::operator ->() const { return FileNode(fs, (con
 inline String::String(const FileNode& fn): cstr_(0), len_(0) { read(fn, *this, *this); }
 
 //! @endcond
+
+/** @brief Loads an object from a filestorage.
+
+This performs the same functionality as @cvLoad, but does it directly from a FileStorage object.
+@param filename File storage object
+@param memstorage Memory storage for dynamic structures, such as CvSeq or CvGraph . It is not used
+for matrices or images.
+@param name Optional object name. If it is NULL, the first top-level object in the storage will be
+loaded.
+@param real_name Optional output parameter that will contain the name of the loaded object
+(useful if name=NULL )
+ */
+CV_EXPORTS void* cvLoadFileStorage(FileStorage& filestorage, CvMemStorage* memstorage, const char* name, const char** real_name);
 
 
 CV_EXPORTS void cvStartWriteRawData_Base64(::CvFileStorage * fs, const char* name, int len, const char* dt);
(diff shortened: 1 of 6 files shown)


modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -241,6 +242,7 @@ class CV_EXPORTS_W CascadeClassifier
     traincascade application.
      */
     CV_WRAP bool load( const String& filename );
+    CV_WRAP bool load( FileStorage& filename );
```
terfendail: Could you please remove CV_WRAP here? Precommit build is failed due to inability to provide python bindings for FileStorage object.

modules/core/include/opencv2/core/persistence.hpp
```diff
@@ -1345,6 +1346,19 @@ inline String::String(const FileNode& fn): cstr_(0), len_(0) { read(fn, *this, *
 
 //! @endcond
 
+/** @brief Loads an object from a filestorage.
+
+This performs the same functionality as @cvLoad, but does it directly from a FileStorage object.
+@param filename File storage object
```
terfendail: This parameter should be named `filestorage`. Also could you please remove at sign before`cvLoad` in previous line? It looks like doxygen considered it as internal command.

modules/core/include/opencv2/core/persistence.hpp
```diff
@@ -1345,6 +1346,19 @@ inline String::String(const FileNode& fn): cstr_(0), len_(0) { read(fn, *this, *
...
+@param memstorage Memory storage for dynamic structures, such as CvSeq or CvGraph . It is not used
+for matrices or images.
+@param name Optional object name. If it is NULL, the first top-level object in the storage will be
+loaded.
+@param real_name Optional output parameter that will contain the name of the loaded object
+(useful if name=NULL )
+ */
+CV_EXPORTS void* cvLoadFileStorage(FileStorage& filestorage, CvMemStorage* memstorage, const char* name, const char** real_name);
```
alalek: Adding of new public entries for legacy C-like API must be avoided.
CvMemStorage and other "C API" things would be removed in OpenCV 4.0, so nobody would write proper tests for this.

modules/objdetect/src/cascadedetect.cpp
```diff
@@ -904,21 +904,24 @@ bool CascadeClassifierImpl::empty() const
...
     if( !fs.isOpened() )
         return false;
 
     if( read_(fs.getFirstTopLevelNode()) )
         return true;
 
-    fs.release();
-
-    oldCascade.reset((CvHaarClassifierCascade*)cvLoad(filename.c_str(), 0, 0, 0));
+    oldCascade.reset((CvHaarClassifierCascade*)cvLoadFileStorage(fs, 0, 0, 0));
```
alalek: Do we really need to support loading of old cascades by this way?

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -177,6 +177,7 @@ class CV_EXPORTS_W BaseCascadeClassifier : public Algorithm
     virtual ~BaseCascadeClassifier();
     virtual bool empty() const CV_OVERRIDE = 0;
     virtual bool load( const String& filename ) = 0;
+    virtual bool load( FileStorage& filestorage) = 0;
```
alalek: New functionality should be covered by tests.

vpisarev:
@edrosten, thank you for the contribution! As @alalek said, we are going to remove C-API in OpenCV 4.0 and the persistence functionality (XML/YAML I/O) will be rewritten using C++ structures. I suggest submit the revised PR after it's done

PR #11464, opened 2018-05-05
Naive proposal for solution to the problem in issue 11426

### LINKED ISSUE #11426: norm with NORM_HAMMING returns strange results

- OpenCV => 3.4.0

##### Detailed description

when using cv2.norm in python using NORM_HAMMING, the results are strange.
For instance the returned distance between 17 and 8 is 0, while clearly 17 = '00010001' and 8 = '00001000'
The real distance is 3.0

##### Steps to reproduce
    a = np.array([8], dtype=np.uint8)
    b = np.array([17], dtype=np.uint8)
    cv2.norm(a, b, cv2.NORM_HAMMING)
    0.0

saskatchewancatch:
### This pullrequest changes

Just a naive idea for a solution to the issue in #11426 . Feedback greatly needed and appreciated.

### DIFF
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -688,7 +688,7 @@ The type of norm to calculate is specified using #NormTypes.
 @param normType type of the norm (see #NormTypes).
 @param mask optional operation mask; it must have the same size as src1 and CV_8UC1 type.
 */
-CV_EXPORTS_W double norm(InputArray src1, InputArray src2,
+CV_EXPORTS_AS(norm2) double norm(InputArray src1, InputArray src2,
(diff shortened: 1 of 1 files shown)


saskatchewancatch:
Ugh, forgot that other language bindings also care about those EXPORT declarations.

saskatchewancatch:
@vpisarev, understood. I believe the logs are gone from the build system, but from my memory I believe what happened was that my proposal would have broken the Java API in the sense that the autogenerated functions were incompatible. I'll confirm this locally.

catree:
@saskatchewancatch 

I think that you can trigger a new build without adding modifications with:
`git commit --amend --no-edit && git push -f`

PR #11465, opened 2018-05-05
add universal intrinsics for HSV2RGB_b

K-Shinotsuka:
```
```

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin_neon.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_neon.hpp
@@ -1277,6 +1277,64 @@ OPENCV_HAL_IMPL_NEON_INTERLEAVED(float32x4, float, f32)
 OPENCV_HAL_IMPL_NEON_INTERLEAVED(float64x2, double, f64)
 #endif
 
+inline void v_load_deinterleave_q3(const uchar* ptr, v_uint8x16& a)
+{
+    uchar v1[] = { 0, 4, 12,  1,  5, 13,  2,  6};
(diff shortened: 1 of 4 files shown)


K-Shinotsuka:
@vpisarev 
Thank you for your advice.
I'll delete v_load_deinterleave_expand().
But I'll create v_load_deinterleave_q() or v_load_deinterleave_half().
And I'll use v_expand() and convert s32 to f32.

vpisarev:
@K-Shinotsuka, ok, that can be also the solution. Just a few small notes:

1. it would be better to add `v_load_deinterleave_low()` (that would load and deinterleave 8 pixels, producing v_uint16x8 x 3). It should have pretty good implementations for NEON and SSE. The problem is that PPC64 build will be broken unless you provide some default implementation for it as well.
2. if you want, you may add v_expand universal intrinsic that would convert u16 to f32 via conversion to u32, then reinterpretation to s32, then conversion fo f32. This intrinsic may be replicated 1:1 on each backend, because it can be implemented using existing universal intrinsics.
3. we have now the tradition to cover all important functionality, including the newly added universal intrinsics, with tests. So, you need to provide some regression tests for the newly added intrinsics.

vpisarev:
@K-Shinotsuka, it looks like the latest version of your patch does not pass the regression tests on ARM. Could you please look at it?

vpisarev:
superseded by #11637

K-Shinotsuka:
@vpisarev 
Ah!
I haven't finished it yet, because I'm so busy with work.

K-Shinotsuka:
Maybe I'll recommit it.

PR #11458, opened 2018-05-04
Update doc dnn yolo 3

cabelo:

PR #11468, opened 2018-05-05
Select the device (video capture) in segmentation.cpp

cabelo:


dkurt:
@cabelo, Could you please replace all the similar changes into a single PR? https://github.com/opencv/opencv/pull/11458 or https://github.com/opencv/opencv/pull/11467 or this one.

PR #11474, opened 2018-05-07
arm: fix build error of v_rotate_left

### LINKED ISSUE #11473: arm: build fails because of invalid parameter of v_rotate_left

- OpenCV => master ( after 8a767689cfbe6fec0e130d4efa2b726019f9c903 )
- Operating System / Platform => ODROID-XU4 Arm 32bit
- Compiler => GCC 5.4.0

##### Detailed description
- Building OpenCV master causes error on Arm platform
```
mkdir build
cd build
cmake -DBUILD_EXAMPLES:BOOL=ON -DCPU_BASELINE=FP16 -DNEON=ON .. 
make -j 4 all
<snip>
In file included from /home/odroid/opencv-fork/modules/core/include/opencv2/core/cv_cpu_dispatch.h:102:0,
                 from /home/odroid/opencv-fork/modules/core/include/opencv2/core/cvdef.h:202,
                 from /home/odroid/opencv-fork/modules/core/include/opencv2/core.hpp:52,
                 from /home/odroid/opencv-fork/modules/ts/include/opencv2/ts.hpp:10,
                 from /home/odroid/opencv-fork/modules/core/test/test_precomp.hpp:7,
                 from /home/odroid/opencv-fork/modules/core/test/test_intrin.cpp:4:
In function ‘int64x2_t vextq_s64(int64x2_t, int64x2_t, int)’,
    inlined from ‘opencv_test::hal::cpu_baseline::TheTest<R>& opencv_test::hal::cpu_baseline::TheTest<R>::test_rotate() [with int s = 0; R = cv::hal_baseline::v_int64x2]’ at /home/odroid/opencv-fork/modules/core/include/opencv2/core/hal/intrin_neon.hpp:802:269:
/usr/lib/gcc/arm-linux-gnueabihf/5/include/arm_neon.h:7583:59: error: lane 2 out of range 0 - 1
   return (int64x2_t)__builtin_neon_vextv2di (__a, __b, __c);
                                                           ^
```

The direct cause of the build error is 8a767689cfbe6fec0e130d4efa2b726019f9c903 , but the invalid parameter was passed to ```v_extq_***``` in ```v_rotate_left``` even before.
Following is the definition of the ```v_rotate_***``` in [intrin_neon.h](https://github.com/opencv/opencv/blob/000a13b6a35f65eb04e5c78721364406718f5d81/modules/core/include/opencv2/core/hal/intrin_neon.hpp#L784-L792)

```
#define OPENCV_HAL_IMPL_NEON_ROTATE_OP(_Tpvec, suffix) \
template<int n> inline _Tpvec v_rotate_right(const _Tpvec& a) \
{ return _Tpvec(vextq_##suffix(a.val, vdupq_n_##suffix(0), n)); } \
template<int n> inline _Tpvec v_rotate_left(const _Tpvec& a) \
{ return _Tpvec(vextq_##suffix(vdupq_n_##suffix(0), a.val, _Tpvec::nlanes - n)); } \
template<int n> inline _Tpvec v_rotate_right(const _Tpvec& a, const _Tpvec& b) \
{ return _Tpvec(vextq_##suffix(a.val, b.val, n)); } \
template<int n> inline _Tpvec v_rotate_left(const _Tpvec& a, const _Tpvec& b) \
{ return _Tpvec(vextq_##suffix(b.val, a.val, _Tpvec::nlanes - n)); }
```

Here, ```v_int32x4``` for instance, the 3rd parameter of ```v_extq_s32``` is 4, which is out of range of ```[0-3]```
```
template<int 0> inline v_int32x4 v_rotate_left(const v_int32x4& a)  { return v_int32x4(vextq_s32(vdupq_n_s32(0), a.val, v_int32x4::nlanes - 0)); }
```
- I'll send a PR later

##### Steps to reproduce
- described above

tomoaki0705:
resolves #11473


modules/core/include/opencv2/core/hal/intrin_neon.hpp
```diff
@@ -785,11 +785,11 @@ OPENCV_HAL_IMPL_NEON_SHIFT_OP(v_int64x2, s64, int64, s64)
 template<int n> inline _Tpvec v_rotate_right(const _Tpvec& a) \
 { return _Tpvec(vextq_##suffix(a.val, vdupq_n_##suffix(0), n)); } \
 template<int n> inline _Tpvec v_rotate_left(const _Tpvec& a) \
-{ return _Tpvec(vextq_##suffix(vdupq_n_##suffix(0), a.val, _Tpvec::nlanes - n)); } \
+{ return _Tpvec(vextq_##suffix(vdupq_n_##suffix(0), a.val, (_Tpvec::nlanes - 1) - n)); } \
 template<int n> inline _Tpvec v_rotate_right(const _Tpvec& a, const _Tpvec& b) \
 { return _Tpvec(vextq_##suffix(a.val, b.val, n)); } \
 template<int n> inline _Tpvec v_rotate_left(const _Tpvec& a, const _Tpvec& b) \
-{ return _Tpvec(vextq_##suffix(b.val, a.val, _Tpvec::nlanes - n)); }
+{ return _Tpvec(vextq_##suffix(b.val, a.val, (_Tpvec::nlanes - 1) - n)); }
```
terfendail: It looks like this will change the result. `v_rotate_left<_Tpvec::nlanes - 1>` will produce `b.val`, while it should combine parts of `a.val` and `b.val`

tomoaki0705: Right, let me get back with tested result.
I only confirmed that it let the build go through.

kinchungwong:
The android build (Android armeabi-v7a) on the build farm seems amiss.

[Buildbot Build 15004](http://pullrequest.opencv.org/buildbot/builders/precommit_android/builds/15004)

* It only compiles code, doesn't run the tests. This is presumably because Android NDK build happens on an x86-64 Linux machine, whereas running the tests would have required either an Android device or an emulator, the latter of which runs very slowly.
  * Unfortunately, when developing SIMD (intrinsics), not running the tests will make development more difficult.
* [NEON support was reported as not available in this build.](http://pullrequest.opencv.org/buildbot/builders/precommit_android/builds/15004/steps/cmake/logs/stdio) This seems to be caused by the automatic configuration tests [```/opencv/cmake/checks/cpu_neon.cpp```](https://github.com/opencv/opencv/blob/3.4/cmake/checks/cpu_neon.cpp), a small source file that is compiled by CMake and its result used by CMake to decide whether NEON was available, [reporting that NEON was not supported. See log.](http://pullrequest.opencv.org/buildbot/builders/precommit_android/builds/15004/steps/cmake/logs/CMakeError)
* Digging further, this seems to be due to flags passed into GCC. It is not sure whether the issue happens with OpenCV CMake or with Android NDK makefile. 
  * [NDK Reference 1: NEON Support](https://developer.android.com/ndk/guides/cpu-arm-neon)
  * [NDK Reference 2: Standalone Toolchains](https://developer.android.com/ndk/guides/standalone_toolchain)

Two phrases from NDK reference are highlighted:
* "Not all ARMv7-based Android devices support NEON, ..."
* "To have the NDK build all its source files with NEON support, include the following line in your module definition: ```LOCAL_ARM_NEON := true``` "

(The ```LOCAL_ARM_NEON``` refers to something to be added to the ```Android.mk```.)

mshabunin:
@kinchungwong , this Android automatic build is there for checking Android/Java specific features, not ARM platform compatibility. For ARM we have two separate manual builders: "ARMv7" and "ARMv8" which run a subset of tests on actual devices.

It is possible to build for Android ARMv7 with NEON by providing `-DANDROID_ABI='armeabi-v7a with NEON'` cmake option. NDK-specific options will not work because we build the library with cmake.

modules/core/include/opencv2/core/hal/intrin_neon.hpp
seiko2plus: Just try to add a special case if immediate constant value equal zero.
````
if (n == 0) return a;
````

**Comment by seiko2plus** [comment]
Seriously I don't see any reason to add test case if immediate constant value equal zero or out of range, I mean after all its "immediate operands", like the following one
https://github.com/opencv/opencv/blob/000a13b6a35f65eb04e5c78721364406718f5d81/modules/core/test/test_intrin.cpp#L38
what's the point for testing return of test_rotate<0>() or test_extract<0>() ??

In other word, I suggest remove any test case of immediate constant zero or out of range and replaced with assertion to check these situations just in case if compiler doesn't complain.

**Comment by tomoaki0705** [comment]
Hmmm, I'm not sure which way to go.
At least, [warning is happening from the specialized template function](http://pullrequest.opencv.org/buildbot/builders/precommit_armv7/builds/10812/steps/compile%20release/logs/stdio), 
```

In file included from /build/precommit_armv7/3.4/opencv/modules/core/include/opencv2/core/hal/intrin.hpp:320:0,
                 from /build/precommit_armv7/3.4/opencv/modules/core/src/precomp.hpp:85,
                 from /build/precommit_armv7/3.4/opencv/modules/core/src/cuda_stream.cpp:43:
/build/precommit_armv7/3.4/opencv/modules/core/include/opencv2/core/hal/intrin_neon.hpp:791:26: warning: unused parameter 'a' [-Wunused-parameter]
 template<> inline _Tpvec v_rotate_left<_Tpvec::nlanes>(const _Tpvec& a) \
                          ^
/build/precommit_armv7/3.4/opencv/modules/core/include/opencv2/core/hal/intrin_neon.hpp:802:1: note: in expansion of macro 'OPENCV_HAL_IMPL_NEON_ROTATE_OP'
 OPENCV_HAL_IMPL_NEON_ROTATE_OP(v_uint8x16, u8)
 ^
```

Probably I'll follow @seiko2plus comment.

PR #11481, opened 2018-05-08
Fixes 11480 : allow to force the video codec used to read videos

### LINKED ISSUE #11480: Allow to force the video codec when reading a video using FFMpeg backend
- OpenCV => 3.4.1

##### Detailed description
After trying unsuccessfully to use cuvid h264 decoder using the FFMpeg backend, I think that it could be useful to allow to force the video codec used to decode the video.

I see in the source code that the FFMpeg backend can already be configured using an environment variable named "OPENCV_FFMPEG_CAPTURE_OPTIONS", but this one cannot actually be used to force the codec.

So I propose to add an entry to this envvar that would be use to force the codec ; example :
export OPENCV_FFMPEG_CAPTURE_OPTIONS="video_codec;h264_cuvid"

dapicard:
Allow to force the video codec when reading a video using FFMpeg backend

resolves #11480

### DIFF
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -870,7 +870,12 @@ bool CvCapture_FFMPEG::open( const char* _filename )
             int enc_width = enc->width;
             int enc_height = enc->height;
 
-            AVCodec *codec = avcodec_find_decoder(enc->codec_id);
+            AVCodec *codec;
+            if(av_dict_get(dict, "video_codec", NULL, 0) == NULL) {
+                codec = avcodec_find_decoder(enc->codec_id);
+            } else {
+                codec = avcodec_find_decoder_by_name(av_dict_get(dict, "video_codec", NULL, 0)->value);
+            }
             if (!codec ||
 #if LIBAVCODEC_VERSION_INT >= ((53<<16)+(8<<8)+0)
                 avcodec_open2(enc, codec, NULL)


modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -870,7 +870,12 @@ bool CvCapture_FFMPEG::open( const char* _filename )
             int enc_width = enc->width;
             int enc_height = enc->height;
 
-            AVCodec *codec = avcodec_find_decoder(enc->codec_id);
+            AVCodec *codec;
+            if(av_dict_get(dict, "video_codec", NULL, 0) == NULL) {
+                codec = avcodec_find_decoder(enc->codec_id);
+            } else {
+                codec = avcodec_find_decoder_by_name(av_dict_get(dict, "video_codec", NULL, 0)->value);
```
alalek: What ffmpeg/avcodec versions are allowed to do this?

dapicard: To do what ? Find a decoder with its name ?
I will try to find this information.

alalek: We need to support multiple version of ffmpeg (at least actual versions).

> since the v0.5 (2009) 

It is fine. Thank you!

dapicard: My pleasure !

PR #11483, opened 2018-05-08
cuda_canny : multi stream safety

### LINKED ISSUE #11193: Problem using prioritized streams in CUDA canny edge detector

- OpenCV => Master

##### Detailed description

The CUDA canny edge detector (cudaimgproc/src/canny.cpp) has a bug on line 221, which is:
`        map_.setTo(Scalar::all(0));`
This call is ignoring the stream, which when using prioritized streams causes synchronization problems. The line should be:
`        map_.setTo(Scalar::all(0), stream);`

##### Steps to reproduce

The code is self-evident.

nglee:
Current implementation of CUDA Canny algorithm is not safe for concurrent execution with multiple streams for several reasons:

* There are several device variables that are declared as global which results in a race condition(`canny::counter`, `column_filter::c_kernel`, `row_filter::c_kernel`)
* Texture references `canny::tex_src` and `canny::tex_mag` are used and since texture references must be declared as static global variables, this results in a race condition

This PR fixes these issues by 1) converting global variables to class local variables, and 2) using texture objects instead of texture references when cc >= 3.0.

A new test suite `CUDA_ImgProc/Canny.Async` is introduced to check multi stream safety. This test fails if above issues are not handled. This test is disabled for cc < 3.0 since texture reference issue cannot be solved for them.

This PR also fixes #11193.

```
force_builders=Custom
```

### DIFF
--- a/modules/cudafilters/src/cuda/column_filter.hpp
+++ b/modules/cudafilters/src/cuda/column_filter.hpp
@@ -52,10 +52,8 @@ namespace column_filter
 {
     #define MAX_KERNEL_SIZE 32
 
-    __constant__ float c_kernel[MAX_KERNEL_SIZE];
-
     template <int KSIZE, typename T, typename D, typename B>
-    __global__ void linearColumnFilter(const PtrStepSz<T> src, PtrStep<D> dst, const int anchor, const B brd)
+    __global__ void linearColumnFilter(const PtrStepSz<T> src, PtrStep<D> dst, const float* kernel, const int anchor, const B brd)
     {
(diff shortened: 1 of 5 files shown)


nglee:
@alalek 
Hi, the Win64 OpenCL build machine says `RemoveDirectory_1 Delete failed.` without providing any detailed log. What should I do to pass this build check?

alalek:
@nglee It is sporadic failure, just ignore that. We usually re-trigger these builds manually or you can push "amended" commit into your branch to trigger all builds again.

PR #11489, opened 2018-05-09
core: fix ARM intrinsincs (v_rotate_left<0>)

alalek:
'0' is specific case (make no sence as a standalone operation),
but it can be useful in template-based programming.

reverts commit: a58c9d4d630926ed81beda1a8f7f8ec52165949e
related #11474


vpisarev:
:+1:

PR #11397, opened 2018-04-26
implementation for dnn fp16 ocl support

pengli:
This PR is for the feature "Adding FP16 path in DNN" at https://github.com/opencv/opencv/issues/11009

PR #11494, opened 2018-05-10
dnn: work around of the test failure of opencv_test_dnn

### LINKED ISSUE #11493: dnn: OCL implementation only works on Intel GPU

- OpenCV => recent master ( 9dc200525 )
- Operating System / Platform => Windows 10 64bit / Arm 32bit (Tinkerboard, ODROID-XU4)  / Arm 64bit (Firefly RK3399)
- Compiler => Visual Studio 2015 / GCC 6.3.0 / 5.4.0
- Device => Intel Core i7 8650U, NVIDIA GTX 1060, Arm Mali T860, Mali T760, Mali T628

##### Detailed description
- Two tests from ```opencv_test_dnn``` fail 
```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] Test_TensorFlow_layers.batch_norm/1, where GetParam() = DNN_TARGET_OPENCL
[  FAILED  ] Test_TensorFlow_layers.pooling/1, where GetParam() = DNN_TARGET_OPENCL

 2 FAILED TESTS
```
- The list of device and the test result is as following (device was chosen by environment variable ```OPENCV_OPENCL_DEVICE```)

|Device|OPENCV_OPENCL_DEVICE|Test result|
|---|:-:|:-:|
|Intel UHD Graphics 620|```:igpu```|PASS|
|Intel Corei7 8650U|```:cpu```|FAILURE|
|NVIDIA GTX 1060|```:dgpu```|FAILURE|
|Arm Mali T***|```:igpu```|FAILURE|

- I looked in the [source code](https://github.com/opencv/opencv/blob/9dc20052517ac580b8b0e079191442e8929f022e/modules/dnn/src/dnn.cpp), and there [are](https://github.com/opencv/opencv/blob/9dc20052517ac580b8b0e079191442e8929f022e/modules/dnn/src/layers/pooling_layer.cpp#L193) [lots](https://github.com/opencv/opencv/blob/9dc20052517ac580b8b0e079191442e8929f022e/modules/dnn/src/layers/mvn_layer.cpp#L253) [of](https://github.com/opencv/opencv/blob/9dc20052517ac580b8b0e079191442e8929f022e/modules/dnn/src/layers/elementwise_layers.cpp#L180) ```isIntel``` call in dnn module
https://github.com/opencv/opencv/blob/9dc20052517ac580b8b0e079191442e8929f022e/modules/dnn/src/layers/pooling_layer.cpp#L192-L194

- On the other hand, in [dnn.cpp](https://github.com/opencv/opencv/blob/9dc20052517ac580b8b0e079191442e8929f022e/modules/dnn/src/dnn.cpp), ```isIntel``` is not utilized correctly
https://github.com/opencv/opencv/blob/9dc20052517ac580b8b0e079191442e8929f022e/modules/dnn/src/dnn.cpp#L725
https://github.com/opencv/opencv/blob/9dc20052517ac580b8b0e079191442e8929f022e/modules/dnn/src/dnn.cpp#L1493
- Also, recent Corei7 series are capable of OpenCL, not on the iGPU, but on the CPU itself.
- In both Corei7 and HD graphics, ```isIntel()``` become true, but it seems that something wrong is happening when cpu version of OpenCL is triggered.
- I once tried removing all the ```isIntel``` but still, the result of OpenCL version didn't match with CPU version.
- As a result, OpenCL version implementation exists, but is only available on Intel HD graphics.
- I'll send a PR by inserting a ```isIntel``` every where, so that OpenCL implementation of dnn module will only correctly triggered on Intel GPU
- Probably, the correct way to go is to **remove** the ```isIntel```, but that was a mess, at least for now.
- I'll send a PR later

##### Steps to reproduce
- Build recent master with OpenCL activated
- Run ```opencv_test_dnn``` on NVIDIA GPU/Intel CPU/Arm GPU.

tomoaki0705:
related #11493 

### This pullrequest changes
  * fix Test_TensorFlow_layers.batch_norm
  * fix Test_TensorFlow_layers.pooling
  * let OpenCL kernel run only on Intel GPU
  * BTW, probably this PR will conflict with #11397.  I won't mind solving the conflict after #11397 is merged.

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -62,6 +62,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
 // this option is useful to run valgrind memory errors detection
 static bool DNN_DISABLE_MEMORY_OPTIMIZATIONS = utils::getConfigurationParameterBool("OPENCV_DNN_DISABLE_MEMORY_OPTIMIZATIONS", false);
 
+static bool DNN_OPENCL_ALLOW_ALL_DEVICES = utils::getConfigurationParameterBool("OPENCV_DNN_OPENCL_ALLOW_ALL_DEVICES", false);
+
 using std::vector;
 using std::map;
 using std::make_pair;
@@ -847,11 +849,21 @@ struct Net::Impl
 
         if (!netWasAllocated || this->blobsToKeep != blobsToKeep_)
         {
-#ifndef HAVE_OPENCL
             if (preferableBackend == DNN_BACKEND_DEFAULT && preferableTarget == DNN_TARGET_OPENCL)
+#ifndef HAVE_OPENCL
             {
-                CV_LOG_WARNING(NULL, "DNN: OpenCL target is not available in this OpenCV build, switching to CPU.")
+                CV_LOG_WARNING(NULL, "DNN: OpenCL target is not available in this OpenCV build, switching to CPU.");
                 preferableTarget = DNN_TARGET_CPU;
+            }
+#else
+            {
+                if (!DNN_OPENCL_ALLOW_ALL_DEVICES
+                    && !(ocl::Device::getDefault().isIntel() && ocl::Device::getDefault().type() == ocl::Device::TYPE_GPU) // Current implementation is only valid for Intel GPU (#11494)
+                    )
+                {
+                    CV_LOG_WARNING(NULL, "DNN: OpenCL target is not supported with current OpenCL device (tested with Intel GPUs only), switching to CPU.");
+                    preferableTarget = DNN_TARGET_CPU;
+                }
             }
 #endif
             clear();


alalek:
I tested this on AMD GPU device, but dnn tests still [fail](http://pullrequest.opencv.org/buildbot/builders/3_4-skx-lin64/builds/28).
Perhaps It is better to disable this early in some single place. For example, somewhere here: https://github.com/opencv/opencv/blob/b7a9d966d7b53f507becb3a44f75138976d94e8c/modules/dnn/src/dnn.cpp#L850-L856

Could you try something like this: https://github.com/alalek/opencv/commit/pr11494_proposal ?

tomoaki0705:
Thank you for the nice suggestion @alalek.
I followed your suggestion and pushed again.

modules/dnn/src/dnn.cpp
```diff
@@ -847,12 +849,22 @@ struct Net::Impl
...
                 preferableTarget = DNN_TARGET_CPU;
             }
+#else
+            {
+                if (!DNN_OPENCL_ALLOW_ALL_DEVICES
+                    && !(ocl::Device::getDefault().isIntel() && ocl::Device::getDefault().type() == ocl::Device::TYPE_GPU) // Current implementation is only valid for Intel GPU (#11494)
+                    )
+                {
+                    CV_LOG_WARNING(NULL, "DNN: OpenCL target is not supported with current OpenCL device (tested with Intel GPUs only), switching to CPU.")
+                        preferableTarget = DNN_TARGET_CPU;
```
alalek: Looks like there is an indentation mess (due missing semicolon ';' after `CV_LOG_WARNING()` - both cases)

alalek:
@tomoaki0705 Please remove mention of my login (or `@alalek` => `alalek`) from the commit message (GitHub sends tons of spam due that from OpenCV forks updates)

tomoaki0705:
Thanks for the review and sorry for the inconvenience, but that was good to know that I shouldn't write a mention in a log commit.
Thx

PR #11497, opened 2018-05-10
 3rdparty: added libjpeg-turbo (with deprecation of libjpeg)

alalek:
Repo: https://github.com/libjpeg-turbo/libjpeg-turbo.git
Tag: 1.5.3

SIMD optimizations are disabled (removed) due build issues.

Enable `BUILD_JPEG_TURBO_DISABLE=ON` to use libjpeg.

```
allow_multiple_commits=1
```


kinchungwong:
Just a curious question. All of the SIMD code in libjpeg-turbo are actually in the *.asm files in each folder. To build these assembly files, either NASM or YASM will be needed. 

If libjpeg-turbo is to be compiled on each user's computer, then either NASM or YASM will need to be accessible to CMake during configuration. 

Another approach is to invoke NASM or YASM using custom build actions defined for *.asm file type. In this approach, the Visual Studio solution projects need to contain a snippet that knows how to invoke NASM or YASM when the project is built.

There is also the question of whether NASM or YASM should be installed into user's computer as part of the CMake preparation. However this might not be popular with some (experienced) people. But not doing this means some (novice) users will be annoyed when their first encounter with OpenCV CMake fails, and they have to manually install NASM or YASM.

This is just a curious question; I'm sure you have decided how to solve it. I just want to know more about it.

alalek:
> users will be annoyed when their first encounter with OpenCV CMake fails, and they have to manually install NASM or YASM

We would like to avoid such cases.

Currently we put libjpeg-turbo without any SIMD `.asm` code which should work on almost all platforms without extra requirements.
In the future libjpeg-turbo SIMD will be enabled on supported (capable) configurations only (with proper detection of assembler support on build platform).

cmake/OpenCVFindLibsGrfmt.cmake
```diff
@@ -36,13 +36,38 @@ if(WITH_JPEG)
...
+      set(JPEG_INCLUDE_DIR "${${JPEG_LIBRARY}_SOURCE_DIR}")
+    endif()
   endif()
 
-  ocv_parse_header("${JPEG_INCLUDE_DIR}/jpeglib.h" JPEG_VERSION_LINES JPEG_LIB_VERSION)
+  if(NOT DEFINED JPEG_LIB_VERSION AND EXISTS "${JPEG_INCLUDE_DIR}/jpeglib.h")
+    ocv_parse_header("${JPEG_INCLUDE_DIR}/jpeglib.h" JPEG_VERSION_LINES JPEG_LIB_VERSION)
+  endif()
+  if(NOT DEFINED JPEG_LIB_VERSION AND EXISTS "${JPEG_INCLUDE_DIR}/jconfig.h")  # libjpeg-turbo stores version in a separate file
+    ocv_parse_header("${JPEG_INCLUDE_DIR}/jpeglib.h" JPEG_VERSION_LINES JPEG_LIB_VERSION)
```
mshabunin: Why is `jconfig.h` checked for existence but `jpeglib.h` is parsed?

Maybe it is better to make it in a loop, something like:
```.cmake
foreach(jpeg_dir "${JPEG_INCLUDE_DIR};${${JPEG_LIBRARY}_BINARY_DIR}")
foreach(jpeg_hdr "jpeglib.h;jconfig.h;jconfig-64.h;...")
  ocv_parse_header("${jpeg_dir}/${jpeg_hdr}" JPEG_VERSION_LINES JPEG_LIB_VERSION)
  if (JPEG_LIB_VERSION)
    break()
  endif()
endforeach()
endforeach()
```

alalek: Added macro to avoid copy-paste mistakes

PR #11499, opened 2018-05-10
Add assertion for GaussianBlur

take1014:
Using CV_32S, CV_8S as input for GaussianBlur, an error occurs in sepFilter2D. In using GaussianBlur, I think that we need to add assertions to explicitly prohibit the use of these types.


alalek:
Perhaps we don't need this really.
For example, `sepFilter()` gets improved and start supporting of new types. But there is some "standalone" irrelevant check which becomes inconsistent (because check is located in other place).

take1014:
@alalek  I understood. Given the expansion of future functions, it is not desirable to put assertions here as you say. Thank you so much, I learned a lot. I withdraw this pull request.

PR #11502, opened 2018-05-11
ml: fix internal state when changing the impl in KNearest

### LINKED ISSUE #11450: setAlgorithmType in KNearest unsets all parameters
- OpenCV => 3.2 / 3.4
- Operating System / Platform => Ubuntu 16.04 64 Bit
- Compiler => gcc 5.4.0

##### Detailed description

It took me some time to figure out why my trained kNN classifier seemed to forget all parameters I set until I saw that `setAlgorithmType` calls `initImpl` which instantiates the corresponding implementation with default parameters. In my opinion this should not be hidden in a setter function, so this should either happen on `create()` (which requires refactoring of the whole `Algorithm` interface) or simply when `train()` is called.

In any case this behaviour should be explicitly specified in the documantation!

##### Steps to reproduce

```.cpp

Ptr<KNearest> knn = KNearest::create();
knn->setDefaultK(3);
knn->setIsClassifier(false);
knn->setAlgorithmType(KNearest::Types::BRUTE_FORCE)

std::cout << knn->getDefaultK() << std::endl; // will print 10 -- should print 3
std::cout << knn->getIsClassifier() << std::endl; // will print 1 -- should print 0

// train knn with some data ...
```

berak:
resolves #11450

### This pullrequest changes

cache previous impl state vars, like defaultK, Emax, isClassifier, and set them again after creating a new impl object

### DIFF
--- a/modules/ml/src/knearest.cpp
+++ b/modules/ml/src/knearest.cpp
@@ -437,7 +437,16 @@ public:
     {
         if (val != BRUTE_FORCE && val != KDTREE)
             val = BRUTE_FORCE;
+
+        int k = getDefaultK();
+        int e = getEmax();
+        bool c = getIsClassifier();
+
         initImpl(val);
+
+        setDefaultK(k);
+        setEmax(e);
+        setIsClassifier(c);
     }
 
 public:


modules/ml/src/knearest.cpp
```diff
@@ -437,7 +437,16 @@ class KNearestImpl CV_FINAL : public KNearest
     {
         if (val != BRUTE_FORCE && val != KDTREE)
             val = BRUTE_FORCE;
+
+        int k = getDefaultK();
+        int e = getEmax();
+        int c = getIsClassifier();
```
alalek: `bool c`

PR #11505, opened 2018-05-11
handle huge matrices correctly

### LINKED ISSUE #10211: Very large image not supported (area > 2^32)
- OpenCV => 3.3.1
- Operating system => Windows 10 Pro (64bit)

**Related forum link**
[http://answers.opencv.org/question/178840/very-very-large-image-causes-issues-with-function-int-type-related/](http://answers.opencv.org/question/178840/very-very-large-image-causes-issues-with-function-int-type-related/)

**Detailed description**
Large images with an area greater than 2^32 cause crashes or do not work in some functions.
A simple example is the cv::threshold method with Triangle method flag:
The operation `size.width *= size.height;` is done (in file imgproc/src/thresh.cpp, line 1099), but in the case of large images, this overflows the "small" integer `int`.

Moreover, images with area greater than 2^30 cannot be opened (see imgcodecs/src/loadsave.cpp, line 65). As a workaround, I just modified the mentioned line with `#define CV_IO_MAX_IMAGE_PIXELS((uint64)1 << 40`, but this can't be a long term fix.

**Suggestions**
Maybe using other types (like int64 or size_t) for size-related (or maybe index-related) code may be a good idea.
Maybe this can be activated via a preprocessor directive.

vpisarev:
make sure that the matrix with more than `INT_MAX` elements is marked as non-continuous, and thus all the pixel-wise functions process it correctly (i.e. row-by-row, not as a single row, where integer overflow may occur when computing the total number of elements).

resolves #10211

```
force_builders=Custom
```

### DIFF
--- a/modules/core/include/opencv2/core/cuda.hpp
+++ b/modules/core/include/opencv2/core/cuda.hpp
@@ -305,6 +305,9 @@ public:
     //! returns true if GpuMat data is NULL
     bool empty() const;
 
+    //! internal use method: updates the continuity flag
+    void updateContinuityFlag();
+
     /*! includes several bit-fields:
     - the magic signature
     - continuity flag
--- a/modules/core/include/opencv2/core/mat.hpp
+++ b/modules/core/include/opencv2/core/mat.hpp
@@ -2092,6 +2092,9 @@ public:
     static MatAllocator* getDefaultAllocator();
     static void setDefaultAllocator(MatAllocator* allocator);
 
+    //! internal use method: updates the continuity flag
+    void updateContinuityFlag();
+
     //! interaction with UMat
     UMatData* u;
 
@@ -2564,6 +2567,9 @@ public:
     UMatUsageFlags usageFlags; // usage flags for allocator
     //! and the standard allocator
     static MatAllocator* getStdAllocator();
+
+    //! internal use method: updates the continuity flag
+    void updateContinuityFlag();
 
     // black-box container of UMat data
     UMatData* u;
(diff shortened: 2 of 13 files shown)


alalek:
BTW, Do we want to fix "continuous" / "contiguous" mess?

Memory block is contiguous.
Mathematic function is continuous.

Related StackOverflow questions:
- https://stackoverflow.com/questions/26998223/what-is-the-difference-between-contiguous-and-non-contiguous-arrays
- https://stackoverflow.com/questions/4059363/what-is-a-contiguous-memory-block
- https://stackoverflow.com/questions/28930465/what-is-the-difference-between-flatten-and-ravel-functions-in-numpy

vpisarev:
yes, definitely. And it's not a mess at all. On small images (like 20x20) this trick with
<pre>
Size size = mat.size();
if( mat.isContinuous() ) { size.width *= size.height; size.height = 1; }
</pre>
can accelerate processing by 10-20% in the case of simple functions. And it's put in many places of the library and in the user code as well. Obviously, this trick can only be done with `size.width * size.height` is not bigger than `INT_MAX`. And it's also obviously that it only needs to be checked during matrix constructions (just a few places in the library), not when you process the matrix (many places in the library and the user code). We attempted to implement such check in updateContinuityFlag() function, but 1) it contained a bug and 2) it was not used everywhere.

hrnr:
Do I understand correctly that huge arrays (matrix 1xN) are still not supported by cv::Mat? Or is it somehow solved too?

PR #11519, opened 2018-05-15
samples: openni_capture - use COLORMAP_JET for depth visualization

paroj:
instead of an ad-hoc implementation of it


terfendail:
Looks like the change will affect color representation used to display disparities. However the new version is clean and concise so it is fine for me

PR #11522, opened 2018-05-15
Allow usage of Vec6d in python-wrapped function signatures

fegorsch:
Especially useful for accepting or returning 6D-poses as a single
argument/return value.

### This pullrequest changes

Wrap `cv::Vec6d` so it automatically gets converted to/from a numpy array and thus can be used in classes and functions that should be accessible via Python.

### DIFF
--- a/modules/python/src2/cv2.cpp
+++ b/modules/python/src2/cv2.cpp
@@ -1059,6 +1059,19 @@ PyObject* pyopencv_from(const Point3f& p)
     return Py_BuildValue("(ddd)", p.x, p.y, p.z);
 }
 
+static bool pyopencv_to(PyObject* obj, Vec6d& v, ArgInfo info)
+{
+    (void)info;
+    if (!obj)
+        return true;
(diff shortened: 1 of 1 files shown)


alalek:
> automatically gets converted to/from a numpy array

numpy? I see Python tuples only (this is a problem, if you want to process tuples as numpy arrays - they are not compatible)

- Currently I don't see real usecases for this in OpenCV source code
- There are no tests for new functionality.

Missing refactoring of several places like this:
> typedef std::vector<Vec6f> vector_Vec6f;

Mentioned line is still alone.

Why Vec6d is added but Vec6f is ignored? This doesn't make code consistent.

fegorsch:
I underestimated the implications and necessary understanding of Python and Python wrapping. I see now that more thought and work is to be put into this.

As for the usecase: in a private project, I'd like to return a single 6D-vector consistently in C++ and Python instead of a separate 3D rotation and translation using output arguments and multiple return values respectively. I thought this was useful for others, too, even if there is no active usage, yet.

Eventually I'll retry after familiarizing myself more with testing, wrapping for Python and having a module actually use this feature, thanks for the input!

PR #11524, opened 2018-05-15
DNN: Only execute SMID part of convolutional layer if CV_SIMD128 is enabled

marting87:
After the SIMD instructions the non-SIMD part was still executed. The results of the SIMD part was overwritten.

Should lead to performance boost in when computing the conv layer on CPU only.

### DIFF
--- a/modules/dnn/src/layers/convolution_layer.cpp
+++ b/modules/dnn/src/layers/convolution_layer.cpp
@@ -482,10 +482,11 @@ public:
         bool useAVX;
         bool useAVX2;
         bool useAVX512;
+        bool hasSIMD;
 
         ParallelConv()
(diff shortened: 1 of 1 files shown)


alalek:
If SIMD128 compilation option is enabled, then OpenCV should not launch at all if there is no SSE/NEON/VSX support from the hardware.

Please describe in details your platform configuration and used build flags.

marting87:
Agree,` if( hasSIMD )` is not needed. 

I'm using an ARM Processor with NEON support. 

The issue is that the code encapsulated in `#if CV_SIMD128` is executed correctly, but after the `#endif` it is overwritten. Perhabs encapsulate this code in an `#else`?

PR #11551, opened 2018-05-20
Add arguments to dftFilter2D

### LINKED ISSUE #10683: filter2D does not utilize pixels outside ROI when kernel size > 11
**Description**

It is said
> ... now the filtering operations fully support the notion of image ROI, that is, pixels outside of the ROI but inside the image can be used in the filtering operations.

I found when calling `filter2D` to filter an ROI using a kernel larger than 11x11, the function does not utilize the pixels outside the ROI. Probably related to the fact that filter uses DCT instead of convolution beyond that size? I have tested 3.1, 3.2, 3.3, 3.4. This issue happens since 3.2. 3.1 does not have this issue.

**System information**
OpenCV 3.2, 3.3, 3.4
Ubuntu 16.04
g++ 5.4.0

**Steps to reproduce**

The following code produces the problem. It divides an image into blocks and filters each of them.

```c++

#include <iostream>
#include <vector>

#include <opencv2/opencv.hpp>

int main(void) {
  
  std::cout << "OpenCV version : " << CV_VERSION << std::endl;
  
  const char* input_file  = "roi_filter_test.jpg";
  const char* output_file = "roi_filter_test_output.jpg";
  
  cv::Mat image  = cv::imread(input_file);
  cv::Mat output = cv::Mat::zeros(image.size(), image.type());
  
  int BLOCK_SIZE = 64; // can be any
  
  std::vector<cv::Rect> regions;
  for (int r = 0; r < image.rows / BLOCK_SIZE; ++r)
    for (int c = 0; c < image.cols / BLOCK_SIZE; ++c)
      regions.emplace_back(c * BLOCK_SIZE, r * BLOCK_SIZE,
                           BLOCK_SIZE, BLOCK_SIZE);
      
  int ksize  = 12; // issue when > 11
  int ksize2 = ksize * ksize;
  cv::Mat kernel = cv::Mat::ones(ksize, ksize, CV_32F) / ksize2;
    
  for (const auto& region: regions) {
    cv::Mat roi_i(image,  region);
    cv::Mat roi_o(output, region);
    cv::filter2D(roi_i, roi_o, -1, kernel);
  }
  
  cv::imwrite(output_file, output); 
  
  return 0;
}

```
Input Image:
roi_filter_test

Result @ OpenCV 3.1
roi_filter_test_output_3_1

Result @ OpoenCV 3.2, 3.3, 3.4
roi_filter_test_output_3_4

take1014:
resolves #10683


alalek:
Please add some tests for these using of these args (width != full_width) - tests are required for new functionality.

take1014:
@alalek 
Thanks for the follow.
I added a test. Since I thought that "width" and "height" are not necessary as arguments of this function, I deleted them.

alalek:
Well done!
I have rebased patch on 3.4 branch and update test check condition (via "expected" values).

take1014:
Thanks for the follow. 
I agree this change.Thank you very much.

PR #11554, opened 2018-05-21
Add python doit scripts for simplifying opencv platform builds

anuraags:
### This pullrequest changes

I've added some python doit scripts (http://pydoit.org) which makes it easier to build opencv dependencies for the desktop, and mobile platform.


vpisarev:
@anuraags, thank you for the contribution! I do not think we are ready at the moment to integrate it into OpenCV. It looks like a niche tool that may be used by a small fraction of OpenCV users. But such tools arrive and disappear every few months. Perhaps, the scripts should be hosted and maintained separately for now and we'll see if this tool gains enough popularity to be put into the official OpenCV tree.

PR #10717, opened 2018-01-29
dnn patches for slice, eltwise  and mvn layer

pengli:
```
allow_multiple_commits=1
```

PR #11557, opened 2018-05-21
dnn: relax intel only ocl4 dnn

tomoaki0705:
related #11494 #11397 #10639 #10717

I realized that ocl4dnn kernel only works on Intel GPU.
I wrote a workaround in #11494 but the root of the cause was two points.
1. The kernel was using platform dependent behaviour of ```native_powr```
2. The fallback process when OpenCL failed on runtime, was doing wrong normalization

### Platform specific ```native_powr```
 * I confirmed that ```Test_TensorFlow_layers.batch_norm``` and ```Test_TensorFlow_layers.pooling``` passes only on Intel GPU
   * PASS : Intel GPU
   * FAILURE: Intel CPU, NVIDIA Geforce, Arm Mali
 * I digged in for a while and realized that after calculating the mean using OpenCL kernel, some of the elements became ```NaN```
 * I googled a lot and realized that [this line](https://github.com/opencv/opencv/blob/6aec71d7ee02955d781ef57126693ec674976076/modules/dnn/src/opencl/mvn.cl#L88) seems wrong

https://github.com/opencv/opencv/blob/6aec71d7ee02955d781ef57126693ec674976076/modules/dnn/src/opencl/mvn.cl#L88

 * This line is computing the square diff between the average value, but ```native_powr``` is showing a different behaviour based on platform
   * According to the references[1][2][3], it seems that ```native_powr``` requires the first parameter to be **greather than or equal to 0**
   * In this case, the first parameter can be a negative value, and thus, in some situation the computation value resulted in ```NaN```

> native_powr Computes x to the power of y, where x is ? 0. The range of x and y are implementation-defined. The maximum error is implementation-defined.

   * So probably, ```pown``` is the function we should use here.

> pown Computes x to the power of y, where y is an integer. 

### fallback process of normalization
 * Still, the test was failing on Arm Mali (Tinkerboard, ODROID-XU4, Firefly RK3399)
   * It seems that memory was not enough for running the kernel and it was falling back to the CPU version call
   * The fall back is inevidable, but I guess that fall back implementation was not verified correctly, and the normalization result was different from the others
   * I modified the code a bit, and now the test passes, but it's involves removing the following 0 fill from mvn_layer.cpp

https://github.com/opencv/opencv/blob/1060c0f439cbb8de44ff258d2ad3ab9c6b429331/modules/dnn/src/layers/mvn_layer.cpp#L274-L279

   * I'd like to confirm @pengli about this point, just in case if I'm going the wrong way.

### other topics
 * This PR isn't packed as a single commit, becasue there are other points to discuss
 * other ```native_powr```
   * in [mvn.cl](https://github.com/opencv/opencv/blob/master/modules/dnn/src/opencl/mvn.cl), there [are other call of ```native_powr```](https://github.com/opencv/opencv/blob/6aec71d7ee02955d781ef57126693ec674976076/modules/dnn/src/opencl/mvn.cl#L196-L199) which seems ```pown``` looks suitable

https://github.com/opencv/opencv/blob/6aec71d7ee02955d781ef57126693ec674976076/modules/dnn/src/opencl/mvn.cl#L196-L199

   * For this, I could make a separate PR
 * Also, I removed all the ```isIntel()``` call from the dnn module
   * If I understand correctly, these call were added because the kernel doesn't work on other platform than Intel GPU
   * Now all the test pass on 
     * Intel GPU (HD Graphics)
     * Intel CPU (Core i7)
     * NVIDIA GPU (Geforce GTX 1060)
     * Arm GPU (Mali T628, 760, T860)
   * So I guess ```isIntel``` is no more needed in this module
 

[1] https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/pow.html
[2] https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/pow.html
[3] https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/pow.html

### DIFF
--- a/modules/dnn/src/layers/mvn_layer.cpp
+++ b/modules/dnn/src/layers/mvn_layer.cpp
@@ -73,7 +73,7 @@ public:
 
     virtual bool tryFuse(Ptr<Layer>& top) CV_OVERRIDE
     {
-        if (preferableTarget == DNN_TARGET_OPENCL && !fuse_batch_norm)
+        if (!fuse_batch_norm)
         {
             top->getScaleShift(scale, shift);
             fuse_batch_norm = !scale.empty() || !shift.empty();
@@ -252,8 +252,7 @@ public:
         CV_TRACE_FUNCTION();
         CV_TRACE_ARG_VALUE(name, "name", name.c_str());
 
-        CV_OCL_RUN(IS_DNN_OPENCL_TARGET(preferableTarget) &&
-                   OCL_PERFORMANCE_CHECK(ocl::Device::getDefault().isIntel()),
+        CV_OCL_RUN(IS_DNN_OPENCL_TARGET(preferableTarget),
                    forward_ocl(inputs_arr, outputs_arr, internals_arr))
 
         Layer::forward_fallback(inputs_arr, outputs_arr, internals_arr);
@@ -274,25 +273,53 @@ public:
             for( i = 0; i < splitDim; i++ )
                 newRows *= inpBlob.size[i];
 
-            if (inpBlob.total() == newRows)
-            {
-                // MVN is applied to single values at an every row.
-                outBlob.setTo(0);
-                return;
-            }
-
             Mat inpMat = inpBlob.reshape(1, newRows);
             Mat outMat = outBlob.reshape(1, newRows);
+
+            if ( inpBlob.total() == newRows )
+            {
+                // MVN is applied to single values at an every row.
+                if (shift.empty())
+                {
+                    outBlob.setTo(0);
+                }
+                else
+                {
+                    for ( i = 0; i < newRows; i++ )
+                    {
+                        outMat.row(i).setTo(((float*)shift.data)[i]);
+                    }
+                }
+                return;
+            }
 
             Scalar mean, dev;
             for ( i = 0; i < newRows; i++)
             {
                 Mat inpRow = inpMat.row(i);
                 Mat outRow = outMat.row(i);
-
+                float weight = 1.f;
+                float bias = 0.f;
+                if (fuse_batch_norm)
+                {
+                    weight = i < scale.cols ? ((float*)scale.data)[i] : weight;
+                    bias = i < shift.cols ? ((float*)shift.data)[i] : bias;
+                }
                 cv::meanStdDev(inpRow, mean, (normVariance) ? dev : noArray());
                 double alpha = (normVariance) ? 1/(eps + dev[0]) : 1;
-                inpRow.convertTo(outRow, outRow.type(), alpha, -mean[0] * alpha);
+                double normalizationScale = 1.0;
+                double normalizationShift = 0.0;
+                if (fuse_batch_norm)
+                {
+                    normalizationScale = alpha * weight;
+                    normalizationShift = -mean[0] * normalizationScale + bias;
+                }
+                else
+                {
+                    normalizationScale = alpha;
+                    normalizationShift = -mean[0] * alpha;
+                }
+                inpRow.convertTo(outRow, outRow.type(), normalizationScale, normalizationShift);
             }
         }
     }
--- a/modules/dnn/test/test_tf_importer.cpp
+++ b/modules/dnn/test/test_tf_importer.cpp
@@ -160,10 +160,12 @@ TEST_P(Test_TensorFlow_layers, batch_norm)
 TEST_P(Test_TensorFlow_layers, pooling)
 {
     int targetId = GetParam();
+    cv::ocl::Device d = cv::ocl::Device::getDefault();
+    bool loosenFlag = targetId == DNN_TARGET_OPENCL && d.isIntel() && d.type() == cv::ocl::Device::TYPE_CPU;
     runTensorFlowNet("max_pool_even", targetId);
     runTensorFlowNet("max_pool_odd_valid", targetId);
     runTensorFlowNet("ave_pool_same", targetId);
-    runTensorFlowNet("max_pool_odd_same", targetId);
+    runTensorFlowNet("max_pool_odd_same", targetId, false, loosenFlag ? 3e-5 : 1e-5, loosenFlag ? 3e-4 : 1e-4);
     runTensorFlowNet("reduce_mean", targetId);  // an average pooling over all spatial dimensions.
 }
 
--- a/modules/dnn/src/layers/convolution_layer.cpp
+++ b/modules/dnn/src/layers/convolution_layer.cpp
@@ -966,8 +966,7 @@ public:
         CV_TRACE_FUNCTION();
         CV_TRACE_ARG_VALUE(name, "name", name.c_str());
 
-        CV_OCL_RUN(IS_DNN_OPENCL_TARGET(preferableTarget) &&
-                   OCL_PERFORMANCE_CHECK(ocl::Device::getDefault().isIntel()),
+        CV_OCL_RUN(IS_DNN_OPENCL_TARGET(preferableTarget),
                    forward_ocl(inputs_arr, outputs_arr, internals_arr))
 
         Layer::forward_fallback(inputs_arr, outputs_arr, internals_arr);
--- a/modules/dnn/src/layers/elementwise_layers.cpp
+++ b/modules/dnn/src/layers/elementwise_layers.cpp
@@ -176,8 +176,7 @@ public:
     {
         CV_TRACE_FUNCTION();
 
-        CV_OCL_RUN(IS_DNN_OPENCL_TARGET(this->preferableTarget) &&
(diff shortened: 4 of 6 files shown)


modules/dnn/src/layers/mvn_layer.cpp
```diff
@@ -289,10 +288,24 @@ class MVNLayerImpl CV_FINAL : public MVNLayer
...
+                    weight = i < scale.cols ? ((float*)scale.data)[i] : weight;
+                    bias = i < shift.cols ? ((float*)shift.data)[i] : bias;
+                }
                 cv::meanStdDev(inpRow, mean, (normVariance) ? dev : noArray());
                 double alpha = (normVariance) ? 1/(eps + dev[0]) : 1;
-                inpRow.convertTo(outRow, outRow.type(), alpha, -mean[0] * alpha);
+                inpRow.convertTo(outRow, outRow.type(), 1.0f, -mean[0]);
+                if (fuse_batch_norm)
+                {
+                    outRow.convertTo(outRow, outRow.type(), alpha*weight, bias);
```
alalek: It is better to fuse two sequential `convertTo()` calls into one:
```
inp.convertTo(tmp, type, a, b); // tmp = inp * a + b
tmp.convertTo(res, type, c, d); // res = tmp * c + d
// ==>
// res = tmp * c + d
//     = (inp * a + b) * c + d
//     = inp * (a * c) + (b * c + d)
inp.convertTo(res, type, a * c, b * c + d); 
```

modules/dnn/test/test_tf_importer.cpp
```diff
@@ -163,7 +163,7 @@ TEST_P(Test_TensorFlow_layers, pooling)
     runTensorFlowNet("max_pool_even", targetId);
     runTensorFlowNet("max_pool_odd_valid", targetId);
     runTensorFlowNet("ave_pool_same", targetId);
-    runTensorFlowNet("max_pool_odd_same", targetId);
+    runTensorFlowNet("max_pool_odd_same", targetId, false, 3e-5, 3e-4);
```
alalek: Defaults are: `double l1 = 1e-5, double lInf = 1e-4`

What configuration requires updated thresholds?

Perhaps we should preserve default values for:
- non-OpenCL targets (CPU code)
- for Intel GPU OpenCL

modules/dnn/src/layers/mvn_layer.cpp
```diff
@@ -274,12 +273,12 @@ class MVNLayerImpl CV_FINAL : public MVNLayer
...
-                // MVN is applied to single values at an every row.
-                outBlob.setTo(0);
-                return;
-            }
+            //if (inpBlob.total() == newRows)
+            //{
+            //    // MVN is applied to single values at an every row.
+            //    outBlob.setTo(0);
+            //    return;
+            //}
```
alalek: @dkurt Could you take a look on this?

modules/dnn/src/layers/mvn_layer.cpp
```diff
@@ -289,10 +288,28 @@ class MVNLayerImpl CV_FINAL : public MVNLayer
             {
                 Mat inpRow = inpMat.row(i);
                 Mat outRow = outMat.row(i);
-
+                float weight = 1.f;
+                float bias = 0.f;
+                if (fuse_batch_norm)
```
dkurt: We need to make it true at
https://github.com/opencv/opencv/blob/3dd916882a3001b4e65bc3173e23f0e97fe3b434/modules/dnn/src/layers/mvn_layer.cpp#L74-L83

alalek:
BTW, Please ignore DNN EAST test failure on "Linux Debug" configuration. Fix is in progress: #11563

tomoaki0705:
So, the situation is getting complicated.

* Removing the power function is now done, good.
https://github.com/opencv/opencv/blob/4fa673780280394e6e419641664c3292d6105eab/modules/dnn/src/opencl/mvn.cl#L92-L93
* The condition of loosening the threshold is written down, good.
https://github.com/opencv/opencv/blob/4fa673780280394e6e419641664c3292d6105eab/modules/dnn/test/test_tf_importer.cpp#L163-L168
* The scaling is now done once, good.
https://github.com/opencv/opencv/blob/4fa673780280394e6e419641664c3292d6105eab/modules/dnn/src/layers/mvn_layer.cpp#L312
* The 0 fill was my misunderstanding.  In some case there layer becomes single column vector, which has only single value in each row.
In such case standard deviation cannot be computed correctly, and that's probably why the 0 fill existed.


So, there are still three points to solve.
1. Comment from @dkurt
2. Test failure of Test_TensorFlow_fp16.tests on Intel GPU
3. Test failure of Test_TensorFlow_layers.batch_norm on Arm GPU

Some log of 2.
```
Current OpenCL device:
    Type = iGPU
    Name = Intel(R) UHD Graphics 620
    Version = OpenCL 2.1
    Driver version = 22.20.16.4840
(snip)
[==========] 261 tests from 52 test cases ran. (12757 ms total)
[  PASSED  ] 260 tests.
[  FAILED  ] Test_TensorFlow_fp16.tests/2, where GetParam() = DNN_TARGET_OPENCL_FP16
```


Some log of 3.
```
Current OpenCL device: 
    Type = iGPU
    Name = Mali-T760
(snip)
[==========] 261 tests from 52 test cases ran. (6898 ms total)
[  PASSED  ] 260 tests.
[  FAILED  ] Test_TensorFlow_layers.batch_norm/1, where GetParam() = DNN_TARGET_OPENCL
```

I knew the test of Test_TensorFlow_fp16 fails, so I was hoping I can make it as a separate PR, but now it seems they were somehow related, and I need to fix them both.
Let me dig a bit more.

alalek:
> Driver version = 22.20.16.4840

Try to upgrade graphics driver (we observed tests fails during FP16 integration on 23.20.16.4849): https://github.com/opencv/opencv/pull/11397#issuecomment-389098082

tomoaki0705:
@dkurt , may I ask the detail, please ?
1.1. On which commit are you talking about ?
1.2. On which test are you talking about ?

In the latest commit  4fa673780280394e6e419641664c3292d6105eab , I stored back the 0 fill, and now the test passes on Intel GPU, Intel CPU and NVIDIA GPU.
The line you pointed is from another commit 3dd9168, and I'm not sure if we are on the same page.

1.3. About the ```fuse_batch_norm``` flag, I did an experiment a bit, and I'm not sure if we really need it.
I tried another commit 0ba392adc6e2df0d3314467611800ece63038b46 and in here I added a code to write true in ```fuse_batch_norm``` but then the test ```batch_norm``` starts failing.

https://github.com/opencv/opencv/blob/0ba392adc6e2df0d3314467611800ece63038b46/modules/dnn/src/layers/mvn_layer.cpp#L82-L86


The current situation ( 4fa673780280394e6e419641664c3292d6105eab ) seems good to me.
2. seems to be driver related problem and 
3. will take more time to investigate, and it was failing before the PR, so the situation didn't turn bad.  IMHO it's probably better to separate this Arm GPU issue as a separate PR.

dkurt:
@tomoaki0705, I just wanted to say that `fuse_batch_norm` is set to true only for `DNN_TARGET_OPENCL` at `tryFuse` method. But these conditions are in `forward` method which is executed on CPU (`fuse_batch_norm` is always false there for now).

alalek:
BTW, OpenCL's "forward" is able to call CPU forward path too (as a fallback if something goes wrong or not suported).

dkurt:
@tomoaki0705, `mvn_batch_norm_1x1` is failed with fusion because it fills zeros but we need to add a bias too (`shift`). May I ask you to replace
```cpp
virtual bool tryFuse(Ptr<Layer>& top) CV_OVERRIDE
{
    if (preferableTarget == DNN_TARGET_OPENCL && !fuse_batch_norm)
    {
        top->getScaleShift(scale, shift);
        fuse_batch_norm = !scale.empty() || !shift.empty();
        return fuse_batch_norm;
    }
    return false;
}
```
on
```cpp
virtual bool tryFuse(Ptr<Layer>& top) CV_OVERRIDE
{
    if (!fuse_batch_norm)
    {
        top->getScaleShift(scale, shift);
        fuse_batch_norm = !scale.empty() || !shift.empty();
        return fuse_batch_norm;
    }
    return false;
}
```
and modify it something like this:

```cpp
Mat inpMat = inpBlob.reshape(1, newRows);
Mat outMat = outBlob.reshape(1, newRows);

if (inpBlob.total() == newRows)
{
    // MVN is applied to single values at an every row.
    if (shift.empty())
        outBlob.setTo(0);
    else
    {
        for ( i = 0; i < newRows; i++)
        {
            outBlob.row(i) *= ((float*)shift.data)[i];
        }
    }
    return;
}
```

tomoaki0705:
Thank you for nice review! I wouldn't be able to finish it by my self only.

PR #11575, opened 2018-05-23
dnn: fixed IE support on Windows

mshabunin:
### This pullrequest changes

Compilation was broken:
>modules\dnn\src\layers\batch_norm_layer.cpp(287): error C2398: Element '1': conversion from 'const int' to 'unsigned __int64' requires a narrowing conversion


alalek:
=> 3.4

PR #11576, opened 2018-05-23
avoid copyTo(outputarray.getMat()) coding pattern

alalek:
- leads to errors due 'const' modifier of getMat() method
- may be non-optimal with non-CPU data storage


alalek:
:+1:

PR #11582, opened 2018-05-23
interactive-calibration: use empty() method instead of checking dimension size

alalek:
Size accessor raises errors in debug mode with empty cv::Mat.


alalek:
:+1:

PR #11590, opened 2018-05-25
calib3d: chess board - properly detect/handle iCntMaxima=0 case

alalek:
Problem can be detected without `if (iCntMaxima == 0)` handling code:
- `cmake -DENABLE_GNU_STL_DEBUG=ON ...`
- or MSVC debug builds


vpisarev:
:+1:

PR #11594, opened 2018-05-26
 Add arguments for estimateRigidTransform

### LINKED ISSUE #4270: RANSAC parameters for estimateRigidTransform
Transferred from http://code.opencv.org/issues/936
```
|| Do Bi on 2011-03-08 20:26
|| Priority: Normal
|| Affected: None
|| Category: imgproc, video
|| Tracker: Feature
|| Difficulty: None
|| PR: None
|| Platform: None / None
```

RANSAC parameters for estimateRigidTransform
-----------
```
Hi,

I think, it would be cool, if RANSAC_MAX_ITERS, RANSAC_SIZE0 and RANSAC_GOOD_RATIO in cvEstimateRigidTransform were parameters and not hard coded, and if this flexibility would also be usable in cv::estimateRigidTransform.

Regards,
Dobias
```

History
-------

take1014:
resolves #4270

### DIFF
--- a/modules/video/test/test_estimaterigid.cpp
+++ b/modules/video/test/test_estimaterigid.cpp
@@ -172,4 +172,53 @@ void CV_RigidTransform_Test::run( int start_from )
 
 TEST(Video_RigidFlow, accuracy) { CV_RigidTransform_Test test; test.safe_run(); }
 
+class CV_RigidTransform_Test4270 : public cvtest::BaseTest
+{
+public:
+    CV_RigidTransform_Test4270();
+    ~CV_RigidTransform_Test4270();
+protected:
+    void run(int);
+};
+
+CV_RigidTransform_Test4270::CV_RigidTransform_Test4270() {}
+CV_RigidTransform_Test4270::~CV_RigidTransform_Test4270() {}
+
+void CV_RigidTransform_Test4270::run( int )
+{
+    Mat img;
+    Mat testImg = imread( string(ts->get_data_path()) + "shared/graffiti.png", 1);
+    double norm = 0.f;
+    if (testImg.empty())
+    {
+       ts->printf( ts->LOG, "test image can not be read");
+       ts->set_failed_test_info(cvtest::TS::FAIL_INVALID_TEST_DATA);
+       return;
+    }
+    pyrDown(testImg, img);
+
+    Mat aff = cv::getRotationMatrix2D(Point(img.cols/2, img.rows/2), 1, 0.99);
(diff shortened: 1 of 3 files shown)


hrnr:
there is already `estimateAffinePartial2D` and friends https://docs.opencv.org/master/d9/d0c/group__calib3d.html#gad767faff73e9cbd8b9d92b955b50062d

I think we should really get rid of estimateRigidTransform in OpenCV 4, or reimplement using `estimateAffinePartial2D` to avoid duplicate code.

modules/video/test/test_estimaterigid.cpp
```diff
@@ -172,4 +172,53 @@ void CV_RigidTransform_Test::run( int start_from )
...
+    ASSERT_ANY_THROW(estimateRigidTransform(img, rotated, true, 500, 2, 0.5f));
+    ASSERT_ANY_THROW(estimateRigidTransform(img, rotated, true, 500, 3, -1.5f));
+    ASSERT_ANY_THROW(estimateRigidTransform(img, rotated, true, 500, 3, 1.5f));
+
+
+    Mat aff_est1 = estimateRigidTransform(img, rotated, true);
+    Mat aff_est2 = estimateRigidTransform(img, rotated, true, 500, 3, 0.5f);
+    norm = cvtest::norm(aff_est1, aff_est2, NORM_INF);
+    EXPECT_DOUBLE_EQ(norm, 0.);
+    Mat aff_est3 = estimateRigidTransform(img, rotated, true, 100, 8, 0.1f);
```
terfendail: It looks like the test checks the result doesn't depend on RANSAC parameters rather than its correctness. Wouldn't it be better to check the result against `aff` matrix?

take1014: Thanks for the follow. 
Certainly, I need to fix this code as you pointed out.
However, as alalek and hrnr say, I do not modify the interface of estimateRigidTransform, because we need to unify the "posture estimation" algorithm to calib3d's function instead of video module's.

take1014:
@alalek, @hrnr, @terfendail 
Thanks for the comments.
 I agree with you.
I think  "estimateAffinePartial2D" should be reused instead of "estimateRigidTransform".
I try to replace the "posture estimation" algorithm used in the video module with that of calib3d's. What do you think?

hrnr:
maybe we should deprecate `estimateRigidTransform` in OpenCV 4. The RANSAC part could be reimplemented using calib3d right now.

PR #11596, opened 2018-05-26
videoio: backends priority list

alalek:
Current fixed priority is not good, especially in case of existence of two separate lists (legacy and new backend APIs).
This PR adds a controlled list of enabled backends.

- adds disabling of some backends during runtime (`OPENCV_VIDEOIO_PRIORITY_GSTREAMER=0`)
- allow to increase priorities of interested backends (`OPENCV_VIDEOIO_PRIORITY_LIST=backend1,backend2`)
- should not affect forced "apiPreference" (but disabling should work)
- unified way of capture creation:
    - try/catch
    - logging of tried backends

```
docker_image:Custom=gstreamer:14.04
docker_image:Linux x64 Debug=gstreamer:16.04
```


vpisarev:
:+1:

alalek:
BTW, Currently application dumps these lines when VideoCapture is opened (once per app launch):
```
[ INFO:0] VIDEOIO: Builtin backends(4): FFMPEG(1000); V4L2(990); CV_IMAGES(980); CV_MJPEG(970)
[ INFO:0] VIDEOIO: Available backends(4): FFMPEG(1000); V4L2(990); CV_IMAGES(980); CV_MJPEG(970)
[ INFO:0] VIDEOIO: Enabled backends(4, sorted by priority): FFMPEG(1000); V4L2(990); CV_IMAGES(980); CV_MJPEG(970)
```

Should we hide some (leave "enabled" line only) or all of these lines?

modules/videoio/src/videoio_c.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+#include "precomp.hpp"
+
+#include "videoio_registry.hpp"
+
+using namespace cv;
+
+// Legacy C-like API
+
+CV_IMPL CvCapture* cvCreateCameraCapture(int index)
```
terfendail: Previous version of `cvCreateCameraCapture` allows to specify preferred backend as `backend*100+cameraID`. It looks like updated version ignore provided backend.

modules/videoio/src/videoio_registry.cpp
```diff
@@ -0,0 +1,640 @@
...
+            CV_LOG_WARNING(NULL, cv::format("VIDEOIO(%s): result=%p isOpened=%d ...\n", \
+                #backend_func, icap.empty() ? NULL : icap.get(), icap.empty() ? -1: icap->isOpened())); \
+    } catch(const cv::Exception& e) { \
+        CV_LOG_ERROR(NULL, cv::format("VIDEOIO(%s): raised OpenCV exception:\n\n%s\n", #backend_func, e.what())); \
+    } catch (const std::exception& e) { \
+        CV_LOG_ERROR(NULL, cv::format("VIDEOIO(%s): raised C++ exception:\n\n%s\n", #backend_func, e.what())); \
+    } catch(...) { \
+        CV_LOG_ERROR(NULL, cv::format("VIDEOIO(%s): raised unknown C++ exception!\n\n", #backend_func)); \
+    } \
+    break; \
```
terfendail: IMHO it would be better to remove unconditional brake here and add it explicitly after every macro call. Explicit brake is already added after almost every `TRY_OPEN`, `TRY_OPEN_LEGACY`. So for a few places without explicit brake it could be unclear that processing will be stopped on first TRY.

alalek: "break" should be unconditional by design.
I added "break"s to make happy IDE (Eclipse CDT). Missing breaks are errors - I will re-check that again.

alalek: VFW shares **the same** api id with V4L/V4L2 (and can't be used simultaneously). Also there is no way to use multiple FIREWIRE backends - only one can be selected.

PR #11601, opened 2018-05-28
calib3d: chess board - replace to cv::findContours()

alalek:
Number of contours are same (in OpenCV tests), but order of contours is changed - this affects next parts of algorithms.

Legacy code are still here for debug purposes (wait for complete refactoring of legacy APIs).


vpisarev:
looks good to me! :+1:

PR #11603, opened 2018-05-28
Disable cvCreateVideoWriter_FFMPEG_proxy when HAVE_FFMPEG defined

### LINKED ISSUE #11602: AVI output file incorrect when HAVE_FFMPEG
For me, it appears since commit aeed43ec90d95dabde81cae01ae244c939fa487a the AVI output file generated by VideoWriter is incorrect/corrupt.

Looking into the code, this commit changed the VideoCapture/VideoWriter backend to C++ interface when HAVE_FFMPEG is defined. Changes to cap.cpp causes the FFMEG proxy function to be used for video capture and writing. Previous behaviour was to default to `createMotionJpegCapture(...)` and `createMotionJpegWriter(...)` even if HAVE_FFMPEG was defined, unless CAP_FFMPEG API preference was specifically passed during opening.

Unfortunately (for me at least) `cvCreateVideoWriter_FFMPEG_proxy(...)` produces incorrect AVI output files (I have not been able to test the capture variant), so `VideoWriter` does so too.

In my limited tests, QuickTime now attempts to convert the file before playing, VLC opens the file but reports a framerate of 600 (when I'd specified 20) and both miss at least the last frame during playback. Opening in Shotcut shows similar issues.

As a temporary fix I have produced a pull request that bypasses new HAVE_FFMPEG switch to the FFMPEG proxy function until it can be fixed (or otherwise confirmed as not the issue). I will continue to try and find the fault with the FFMPEG proxy functions themselves, since in principle these should indeed be used. I have no experience in these functions however, so help would be appreciated.

I am using the latest OpenCV from source, on macOS High Sierra (10.13.4). I am using FFMPEG 4.0 (though have not tried on lower versions, using the FFMPEG command line to create an AVI file itself had no issues).

Steps to reproduce (C++):

```
VideoWriter vw;
int fourcc_mjpeg = VideoWriter::fourcc('M', 'J', 'P', 'G');
vw.open("output.avi", fourcc_mjpeg, 20, frame_size, true);
...
vw << frame;
...
vw.release();
```

output.avi is different to the file produced by previous behaviour, and does not open/play correctly.

As an aside, a **possible workaround** without any code changes for now is to use the MKV container (by specifying output.mkv) instead, which still appears to work as before. Obviously if an AVI is required this is unsuitable.

---

**alalek** (2018-05-29): This issue is related to incomplete support of fresh FFmpeg 4.0.
Should be fixed in the next OpenCV release (OpenCV 3.4.2)

mattmyne:
resolves #11602

### This pullrequest changes

Prevents HAVE_FFMPEG compier define using cvCreateVideoCapture_FFMPEG_proxy(...) and cvCreateVideoWriter_FFMPEG_proxy(...), instead falling-back to the original behaviour of using createMotionJpegCapture(...) and createMotionJpegWriter(...) when no API preference specified.

### DIFF
--- a/modules/videoio/src/cap.cpp
+++ b/modules/videoio/src/cap.cpp
@@ -525,7 +525,7 @@ static Ptr<IVideoCapture> IVideoCapture_create(const String& filename, int apiPr
 {
     bool useAny = (apiPreference == CAP_ANY);
     Ptr<IVideoCapture> capture;
-#ifdef HAVE_FFMPEG
+#if 0 //#ifdef HAVE_FFMPEG
     if (useAny || apiPreference == CAP_FFMPEG)
     {
         capture = cvCreateFileCapture_FFMPEG_proxy(filename);
@@ -579,7 +579,7 @@ static Ptr<IVideoCapture> IVideoCapture_create(const String& filename, int apiPr
 static Ptr<IVideoWriter> IVideoWriter_create(const String& filename, int apiPreference, int _fourcc, double fps, Size frameSize, bool isColor)
 {
     Ptr<IVideoWriter> iwriter;
-#ifdef HAVE_FFMPEG
+#if 0 //#ifdef HAVE_FFMPEG
     if (apiPreference == CAP_FFMPEG || apiPreference == CAP_ANY)
     {
         iwriter = cvCreateVideoWriter_FFMPEG_proxy(filename, _fourcc, fps, frameSize, isColor);


alalek:
This patch can't be accepted, because you disable FFMPEG backend completely.

I believe it is better to configure preferred backends in runtime (via environment variables).
Or specify `apiPreference` for `.open()` method.

mattmyne:
I've modified the patch to still allow FFMPEG backend if apiPreference set, but still default to not using it when apiPreference is CAP_ANY (as was the previous behaviour).

alalek:
This would not work too: FFmpeg is default backend in OpenCV for many years (it must work via CAP_ANY)

mattmyne:
Prior to aeed43ec90d95dabde81cae01ae244c939fa487a (i.e. checking out your previous change to cap.cpp, ef2548778c173fe77495a0f41f3c6106c48949a0) my understanding is that `VideoWriter::open` calls `IVideoWriter_create` which in turn calls `createMotionJpegWriter` if CAP_ANY, regardless of HAVE_FFMPEG define. So FFMPEG proxy function was used when CAP_ANY anyway? Please let me know where I've gone wrong!

alalek:
Currently OpenCV have 2 priority lists to select backend:
- legacy (based on CvCapture)
- new C++ API (`IVideoCapture`)

So this approach was buggy initially. Any backend porting to new API provides mess and several issues like your one.

BTW,
`createMotionJpegWriter()` is designed as a fallback backend with lower priority. But it was added in C++ API list (that incorrectly increases its priority over "legacy API" FFmpeg / GStreamer - this was "fixed" recently due porting of FFmpeg/GStreamer API).

Summary,
you should use `apiPreference` to force selection of OpenCV MJPEG encoder.

mattmyne:
I agree, you should use apiPreference if you want to force OpenCV MJPEG encoder. Previously you'd get this encoder anyway if you requested MJPEG encoding, even if you have FFMPEG; I agree this was wrong.

Now that it has been fixed, and FFMPEG is used by default instead, the FFMPEG output file appears to be incorrectly formatted when specifying MJPEG encoding. This is my concern, apologies for not being clearer.

What my patch does is temporarily revert to the previous behaviour of using the fallback OpenCV MJPEG encoder which still works, until the FFMPEG proxy output is fixed downstream.

The worry is others that have FFMPEG installed (but don't specifically state CAP_FFMPEG preference) that are using VideoWriter to output AVI files with MJPEG encoding may now unexpectedly find their output files do not work anymore.

I have made another commit to my patch to further reduce the case of FFMPEG proxy bypassing to only VideoWriter using MJPEG encoding and AVI container and CAP_ANY preference (so even this can still be overridden with CAP_FFMPEG apiPreference). This solves the AVI and MJPEG issue for me, while retaining current behaviour in all other cases.

PR #11604, opened 2018-05-28
Disable ffmpeg proxy for VideoWriter when MJPEG AVI output

### LINKED ISSUE #11602: AVI output file incorrect when HAVE_FFMPEG

mattmyne:
resolves #11602

### This pullrequest changes

Updated previous pull request #11603 to make FFMPEG proxy bypass specifically for MJPEG encoded AVI output, when apiPreference is not CAP_FFMPEG. Still intended as temporary measure due to incorrect output for this type of file by FFMPEG proxy function, and restores previous use of fallback OpenCV `createMotionJpegWriter`.

### DIFF
--- a/modules/videoio/src/cap.cpp
+++ b/modules/videoio/src/cap.cpp
@@ -525,7 +525,7 @@ static Ptr<IVideoCapture> IVideoCapture_create(const String& filename, int apiPr
 {
     bool useAny = (apiPreference == CAP_ANY);
     Ptr<IVideoCapture> capture;
-#ifdef HAVE_FFMPEG
+#if 0 //#ifdef HAVE_FFMPEG
     if (useAny || apiPreference == CAP_FFMPEG)
     {
         capture = cvCreateFileCapture_FFMPEG_proxy(filename);
@@ -579,7 +579,7 @@ static Ptr<IVideoCapture> IVideoCapture_create(const String& filename, int apiPr
 static Ptr<IVideoWriter> IVideoWriter_create(const String& filename, int apiPreference, int _fourcc, double fps, Size frameSize, bool isColor)
 {
     Ptr<IVideoWriter> iwriter;
-#ifdef HAVE_FFMPEG
+#if 0 //#ifdef HAVE_FFMPEG
     if (apiPreference == CAP_FFMPEG || apiPreference == CAP_ANY)
     {
         iwriter = cvCreateVideoWriter_FFMPEG_proxy(filename, _fourcc, fps, frameSize, isColor);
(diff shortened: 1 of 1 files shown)


alalek:
Please stop.
You should not disable FFmpeg even for MJPEG .avi files to force OpenCV MJPEG encoder.
For this purpose `apiPreference` flag must be used.

FFmpeg backend works well but not with FFmpeg 4.0. We currently have tests with FFmpeg 3.3 / 3.4.

To downgrade homebrew package try something like this:
```
brew switch ffmpeg <something 3.x.y>
```

PR #11605, opened 2018-05-28
build: workaround issues with C compilation mode

alalek:
- cvdef.h + cvRound (double only)
- highgui_c.h

fixes build: #10963

**Note**: There is no official support for C compilation mode (no automatic builds, no tests). This API just "keep" names for old programs which should be compiled with C++ compiler.
Since OpenCV 2.4.x (>5 years).


alalek:
:+1:

PR #11608, opened 2018-05-28
Resize histSize if IPP is enabled

### LINKED ISSUE #11544: calcHist produces incorrect results with IPP
- OpenCV => 3.4.1
- Operating System / Platform => Ubuntu 18.04 64 Bit
- Compiler => gcc 7.3.0

##### Detailed description
cv::calcHist produces incorrect results for certain inputs, if the input images are CV_8U or CV_16U, and if IPP is enabled. The below code snippet reliably reveals the problem, which goes away if image type is changed to float, or if IPP is disabled. From testing other images, it appears as if each bucket index is incorrectly shifted up by one. In the example below, the max_val values are not that important, I found values 990-999 to produce incorrect results, while 1000 produces the correct result.

##### Steps to reproduce
```.cpp
    // Regardless of max value, the image is all zeros, so the histogram
    // should have all entries in the first bucket.
    for (float max_val : { 999, 1000 })
    {
        cv::Mat1w m = cv::Mat1w::zeros(10, 10);
        int n_images = 1;
        int channels[] = { 0 };
        cv::Mat mask;
        cv::MatND hist;
        int dims = 1;
        int hist_size[] = { 1000 };
        float range[] = { 0, max_val };
        const float* ranges[] = { range };
        cv::calcHist(&m, n_images, channels, mask, hist, dims, hist_size, ranges);
        for (int i = 0; i < 5; ++i)
        {
            printf("%d: %f\n", i, hist.at<float>(0, i));
        }
    }
```
Running this, the expected output should be the same for both max_val values, where all 100 elements are in the first bucket. However, with IPP enabled, I get the following:
```
0: 0.000000
1: 100.000000
2: 0.000000
3: 0.000000
4: 0.000000
0: 100.000000
1: 0.000000
2: 0.000000
3: 0.000000
4: 0.000000
```
Without IPP, result is as expected:
```
0: 100.000000
1: 0.000000
2: 0.000000
3: 0.000000
4: 0.000000
0: 100.000000
1: 0.000000
2: 0.000000
3: 0.000000
4: 0.000000
```

take1014:
resolves #11544

### DIFF
--- a/modules/imgproc/src/histogram.cpp
+++ b/modules/imgproc/src/histogram.cpp
@@ -821,6 +821,10 @@ static bool ipp_calchist(const Mat &image, Mat &hist, int histSize, const float*
         return false;
 #endif
 
+    // IPP_DISABLE_HISTOGRAM - https://github.com/opencv/opencv/issues/11544
+    if (uniform && (ranges[0][1] - ranges[0][0]) != histSize)
+        return false;
+
     Mat ihist = hist;
     if(accumulate)
         ihist.create(1, &histSize, CV_32S);
--- a/modules/imgproc/test/test_histograms.cpp
+++ b/modules/imgproc/test/test_histograms.cpp
@@ -1918,5 +1918,35 @@ TEST(Imgproc_Hist_CalcBackProject, accuracy) { CV_CalcBackProjectTest test; test
 TEST(Imgproc_Hist_CalcBackProjectPatch, accuracy) { CV_CalcBackProjectPatchTest test; test.safe_run(); }
 TEST(Imgproc_Hist_BayesianProb, accuracy) { CV_BayesianProbTest test; test.safe_run(); }
 
+TEST(Imgproc_Hist_Calc, calcHist_regression_11544)
+{
+    cv::Mat1w m = cv::Mat1w::zeros(10, 10);
+    int n_images = 1;
+    int channels[] = { 0 };
+    cv::Mat mask;
+    cv::MatND hist1, hist2;
+    cv::MatND hist1_opt, hist2_opt;
+    int dims = 1;
+    int hist_size[] = { 1000 };
+    float range1[] = { 0, 900 };
+    float range2[] = { 0, 1000 };
+    const float* ranges1[] = { range1 };
+    const float* ranges2[] = { range2 };
+
+    setUseOptimized(false);
+    cv::calcHist(&m, n_images, channels, mask, hist1, dims, hist_size, ranges1);
+    cv::calcHist(&m, n_images, channels, mask, hist2, dims, hist_size, ranges2);
+
+    setUseOptimized(true);
+    cv::calcHist(&m, n_images, channels, mask, hist1_opt, dims, hist_size, ranges1);
+    cv::calcHist(&m, n_images, channels, mask, hist2_opt, dims, hist_size, ranges2);
+
+    for(int i = 0; i < 1000; i++)
+    {
+        EXPECT_EQ(hist1.at<float>(i, 0), hist1_opt.at<float>(i, 0)) << i;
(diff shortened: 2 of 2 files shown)


alalek:

So, please:

take1014:
Thanks for the comments.

modules/imgproc/src/histogram.cpp
```diff
@@ -825,10 +825,14 @@ static bool ipp_calchist(const Mat &image, Mat &hist, int histSize, const float*
     if(accumulate)
         ihist.create(1, &histSize, CV_32S);
 
+    int _histSize = histSize;
+    if(histSize > ranges[0][1])
+        _histSize = (int)ranges[0][1];
```
alalek: 1. `histSize` and `ranges` are not dependent values:
    - histSize specifies number of bins (dst size)
    - 'ranges' specifies which src values put into bins.

2. This check ignores non-uniform case.

alalek: I believe we can use this check to disable IPP:
```
    // IPP_DISABLE_HISTOGRAM - https://github.com/opencv/opencv/issues/11544
    if (uniform && (ranges[0][1] - ranges[0][0]) != histSize)
        return false;
```

Please add this and remove other changes from this file.

modules/imgproc/test/test_histograms.cpp
```diff
@@ -1918,5 +1918,28 @@ TEST(Imgproc_Hist_CalcBackProject, accuracy) { CV_CalcBackProjectTest test; test
...
+    float range1[] = { 0, 999 };
+    float range2[] = { 0, 1000 };
+    const float* ranges1[] = { range1 };
+    const float* ranges2[] = { range2 };
+
+    cv::calcHist(&m1, n_images, channels, mask, hist1, dims, hist_size, ranges1);
+    cv::calcHist(&m2, n_images, channels, mask, hist2, dims, hist_size, ranges2);
+    for(int i = 0; i < 1000; i++)
+    {
+        EXPECT_EQ(hist1.at<float>(i, 0), hist2.at<float>(i, 0));
```
alalek: This check looks very strange. It is meaningless checking output similarity for calls with different input parameters.

I believe we should compare "optimized" vs "non-optimized" `calcHist()` calls here:
```
    cv::Mat1w m = cv::Mat1w::zeros(10, 10);
    int n_images = 1;
    int channels[] = { 0 };
    cv::Mat mask;
    cv::Mat hist1, hist2;
    cv::Mat hist1_opt, hist2_opt;
    int dims = 1;
    int hist_size[] = { 1000 };
    float range1[] = { 0, 900 };
    float range2[] = { 0, 1000 };
    const float* ranges1[] = { range1 };
    const float* ranges2[] = { range2 };

    setUseOptimized(false);
    cv::calcHist(&m, n_images, channels, mask, hist1, dims, hist_size, ranges1);
    cv::calcHist(&m, n_images, channels, mask, hist2, dims, hist_size, ranges2);
    setUseOptimized(true);
    cv::calcHist(&m, n_images, channels, mask, hist1_opt, dims, hist_size, ranges1);
    cv::calcHist(&m, n_images, channels, mask, hist2_opt, dims, hist_size, ranges2);
    for(int i = 0; i < 1000; i++)
    {
        EXPECT_EQ(hist1.at<float>(i, 0), hist1_opt.at<float>(i, 0)) << i;
        EXPECT_EQ(hist2.at<float>(i, 0), hist2_opt.at<float>(i, 0)) << i;
    }
```

modules/imgproc/src/histogram.cpp
take1014: I again confirmed all the logic of "calcHist" and understood what you are saying. 
I fix it as you say. It was a great learning experience for me. I am always grateful for your help.

PR #11610, opened 2018-05-28
StereoBM: fixed SIMD processing for fixed-type output arrays

savuor:
A check is not done before `findStereoCorrespondenceBM_SIMD` call whether outputting array should contain shorts or ints.

### This pullrequest changes

* A type of output disparity array is passed now through template parameter.
* Test modified to cover the case

### DIFF
--- a/modules/calib3d/src/stereobm.cpp
+++ b/modules/calib3d/src/stereobm.cpp
@@ -284,6 +284,37 @@ prefilterXSobel( const Mat& src, Mat& dst, int ftzero )
 static const int DISPARITY_SHIFT_16S = 4;
 static const int DISPARITY_SHIFT_32S = 8;
 
+template <typename T>
+struct dispShiftTemplate
+{ };
+
+template<>
+struct dispShiftTemplate<short>
+{
+    enum { value = DISPARITY_SHIFT_16S };
+};
+
+template<>
+struct dispShiftTemplate<int>
+{
+    enum { value = DISPARITY_SHIFT_32S };
+};
+
+template <typename T>
+inline T dispDescale(int /*v1*/, int /*v2*/, int /*d*/);
+
+template<>
+inline short dispDescale(int v1, int v2, int d)
+{
+    return (short)((v1*256 + (d != 0 ? v2*256/d : 0) + 15) >> 4);
+}
+
+template <>
+inline int dispDescale(int v1, int v2, int d)
+{
+    return (int)(v1*256 + (d != 0 ? v2*256/d : 0)); // no need to add 127, this will be converted to float
+}
+
 #if CV_SIMD128
 template <typename dType>
 static void findStereoCorrespondenceBM_SIMD( const Mat& left, const Mat& right,
@@ -303,7 +334,8 @@ static void findStereoCorrespondenceBM_SIMD( const Mat& left, const Mat& right,
     int ftzero = state.preFilterCap;
     int textureThreshold = state.textureThreshold;
     int uniquenessRatio = state.uniquenessRatio;
-    dType FILTERED = (dType)((mindisp - 1) << DISPARITY_SHIFT_16S);
+    const int disp_shift = dispShiftTemplate<dType>::value;
+    dType FILTERED = (dType)((mindisp - 1) << disp_shift);
 
     ushort *sad, *hsad0, *hsad, *hsad_sub;
     int *htext;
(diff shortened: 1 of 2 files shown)


alalek:

So, please:

savuor:
@alalek done

modules/calib3d/src/stereobm.cpp
```diff
@@ -1008,7 +1009,10 @@ struct FindStereoCorrespInvoker : public ParallelLoopBody
 #if CV_SIMD128
         if( useSIMD && useShorts )
         {
-            findStereoCorrespondenceBM_SIMD( left_i, right_i, disp_i, cost_i, *state, ptr, row0, rows - row1 );
+            if( disp_i.type() == CV_16S)
+                findStereoCorrespondenceBM_SIMD<short>( left_i, right_i, disp_i, cost_i, *state, ptr, row0, rows - row1 );
+            else
+                findStereoCorrespondenceBM_SIMD<int>( left_i, right_i, disp_i, cost_i, *state, ptr, row0, rows - row1 );
```
alalek: Please add some check into "else" branch:
- `CV_CheckTypeEQ(disp_i.type(), CV_32S, "Only 16S / 32S types are allowed for 'disparity' output")` or "Internal error" message
- or `CV_Assert()` if type check is performed before

modules/calib3d/test/test_stereomatching.cpp
```diff
@@ -791,6 +791,11 @@ class CV_StereoBMTest : public CV_StereoMatchingTest
         bm->compute( leftImg, rightImg, tempDisp );
         tempDisp.convertTo(leftDisp, CV_32F, 1./StereoMatcher::DISP_SCALE);
 
+        //check for fixed-type disparity data type
+        Mat_<float> fixedFloatDisp;
+        bm->compute( leftImg, rightImg, fixedFloatDisp );
+        EXPECT_TRUE(cvtest::norm(fixedFloatDisp, leftDisp, cv::NORM_L2) < (double)(leftDisp.total()));
```
alalek: 1. Use **relative** norm L2.
2. Use threshold something like 1e-3.

savuor: Threshold of such scale seems unreachable since it's a difference between float-point and short-int calculations. The typical measured (relative L2) error is 0.9375.

modules/calib3d/src/stereobm.cpp
savuor: done, see L1009

modules/calib3d/test/test_stereomatching.cpp
alalek: > The typical measured (relative L2) error is 0.9375

This means that compared data is almost **not** similar.
So you just can not apply this check to verify corectness.


Dump these values:
```
std::cout << "norm x256=" << cvtest::norm(fixedFloatDisp * 256, Mat_<float>(tempDisp), cv::NORM_L2 | cv::NORM_RELATIVE, mask) << std::endl;
std::cout << "norm x16=" << cvtest::norm(fixedFloatDisp * 16, Mat_<float>(tempDisp), cv::NORM_L2 | cv::NORM_RELATIVE, mask) << std::endl;
```

16 is `StereoMatcher::DISP_SCALE`, but what is 256? Bug?

vpisarev:
I have general question. Suppose that maxDisparity - minDisparity is 1024, which is very very big number for stereo correspondence. That's 10 bits. The extrapolation code adds 4 fractional bits. That's 14 bits in total, less than 16 bits for `CV_16S/CV_16U` output type. If `CV_32S` as the output type is added for convenience, can we just always produce 16-bit disparity map and then convert it to 32-bit in the end?

savuor:
@vpisarev sounds reasonable since there is a mess with `disparity_shift` for 16-bit and 32-bit output, this looks buggy.
I'll fix this and then measure if there's a difference between resulting disparities

modules/calib3d/src/stereobm.cpp
```diff
@@ -284,7 +284,39 @@ prefilterXSobel( const Mat& src, Mat& dst, int ftzero )
...
 static const int DISPARITY_SHIFT_32S = 8;
 
+template <typename T>
+struct dispShiftTemplate
+{ };
+
+template<>
+struct dispShiftTemplate<short>
+{
+    static const int value = DISPARITY_SHIFT_16S;
```
alalek: For C++98 "enum" is preferable for type traits:
```
template<> struct dispShiftTemplate<short> { enum { value = DISPARITY_SHIFT_16S }; };
```

modules/calib3d/src/stereobm.cpp
savuor: In fact, the type check is done in `FindStereoCorrespInvoker` constructor, no need to do it again

alalek: Purpose of assertion here (and in general of assertions) is to avoid errors in the future during code refactoring.
This "if" not complete, it assumes two values only.

For example, you want to add CV_32F type, adapt check somewhere (~50 lines from here), update other places, but don't change this block.
How many time would you waste for debugging, because program will not crash (sizeof(int) == sizeof(float)), just produce many NaNs?

PR #11623, opened 2018-05-30
don't set event callback for ocl kernels in dnn module

pengli:
The async notification of CL_COMPLETE event can slow down
GPU pipeline efficiency, in dnn module the input/output blobs
are alive all over the net inference period. So it is safe to
enqueue all the dnn kernels and not wait for CL_COMPLETE event.

### DIFF
--- a/modules/core/src/ocl.cpp
+++ b/modules/core/src/ocl.cpp
@@ -2755,6 +2755,7 @@ struct Kernel::Impl
         cl_program ph = (cl_program)prog.ptr();
         cl_int retval = 0;
         name = kname;
+        module = getProgInfo(prog);
         if (ph)
         {
             handle = clCreateKernel(ph, kname, &retval);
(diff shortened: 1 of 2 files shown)


alalek:
Could you share performance changes (via opencv_perf_dnn) and provide your platform configuration?

alalek:
OS: ?
OpenCL driver/runtime version: ? (Did you use Intel OpenCL NEO?)

pengli:
OS : ubuntu 16.04
OpenCL runtime : srb 5.0

alalek:
Please try NEO OpenCL runtime from here: https://github.com/intel/compute-runtime/releases
NEO is faster up to 20% on dnn tests than SRB 5.0 (i5-6600 / i7-6700, Linux, without this patch).

pengli:
with Neo, this event callback seems not a problem anymore, which is great. I will close this PR. Thanks for the info

PR #11635, opened 2018-05-31
Fix Inference Engine graphs with fused output layers

dkurt:
### This pullrequest changes

Request output blob by a base layer name but not the reference layer which can be fused.


vpisarev:
:+1:

PR #11637, opened 2018-05-31
Hsv2rgb univ intrin

vpisarev:
This is updated #11465 by @K-Shinotsuka; in this version no extra universal intrinsics are introduced

```
```


modules/core/include/opencv2/core/hal/intrin_sse.hpp
```diff
@@ -1773,19 +1773,18 @@ inline void v_load_deinterleave(const unsigned* ptr, v_uint32x4& a, v_uint32x4&
...
-
-    __m128 bt01 = _mm_shuffle_ps(t0, t1, _MM_SHUFFLE(0, 0, 0, 1));
-    __m128 bt12 = _mm_shuffle_ps(t1, t2, _MM_SHUFFLE(0, 2, 0, 3));
-    b.val = _mm_shuffle_ps(bt01, bt12, _MM_SHUFFLE(2, 0, 2, 0));
-
-    __m128 ct01 = _mm_shuffle_ps(t0, t1, _MM_SHUFFLE(0, 1, 0, 2));
-    c.val = _mm_shuffle_ps(ct01, t2, _MM_SHUFFLE(3, 0, 2, 0));
+    a.val = _mm_loadu_ps(ptr + 0);
+    b.val = _mm_loadu_ps(ptr + 4);
+    c.val = _mm_loadu_ps(ptr + 8);
```
alalek: Why this change is necessary?

Using non-local variables for caching doesn't look good. This may break optimizations (compiler must guarantee that "external" data was changed before any possible exception or signal/GP#0) - compiler usually generates terrible code for re-using of "non-local" variables.

"Inline" may help in this case, but similar coding practice should be avoided. Also, this change is not necessary.

vpisarev: oops, it looks like the change was in the original patch from and I missed it. It should be be retained as-is, of course.

Issue #11626, opened 2018-05-30
Task: GCC8 


alalek:
Fix/eliminate GCC 8 build warnings.

Build with warnings: http://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/495

GCC 8 is available in "fedora:28" docker image.

PR #11646, opened 2018-05-31
ts: support parameters with types from anonymous namespace

alalek:
related #11626

<cut>

Eliminates warnings like these:

> modules/ts/include/opencv2/ts/ts_gtest.h:11547:7: warning: 'testing::internal::ParameterizedTestFactory<opencv_test::ocl::OCL_ThreshFixture_Threshold_Threshold_Test>' has a field 'testing::internal::ParameterizedTestFactory<opencv_test::ocl::OCL_ThreshFixture_Threshold_Threshold_Test>::parameter_' whose type uses the anonymous namespace [-Wsubobject-linkage]

Make global variable for `WithParamInterface<T>::parameter_` due issues with using of anonymous namespace in `T`.

```
docker_image:Custom=fedora:28
```

### DIFF
--- a/modules/dnn/test/test_precomp.hpp
+++ b/modules/dnn/test/test_precomp.hpp
@@ -49,7 +49,7 @@
 #include "opencv2/dnn.hpp"
 #include "test_common.hpp"
 
-namespace opencv_test {
+namespace opencv_test { namespace {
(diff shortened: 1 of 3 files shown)


alalek:
:+1:

PR #11649, opened 2018-06-01
dnn: Fix output mismatch when forward dnn model contain [depthwise conv(group=1) + bn + prelu] 

rockzhan:
When loading dnn model contain [Depthwise Conv(group=1) + BN + Prelu], the output of Prelu will shift 2 position due to the default operation.


modules/dnn/src/layers/layers_common.simd.hpp
```diff
@@ -101,9 +101,18 @@ void fastConv( const float* weights, size_t wstep, const float* bias,
 
         if( relu )
         {
-            r0 = relu[i];
-            r1 = relu[i+1];
-            r2 = relu[i+2];
+            if( i+1 >= outCn )
```
dkurt: @rockzhan, Actually, it should works correct because of extra elements:

https://github.com/opencv/opencv/blob/0f298a42032798d2d83017030621fcba885a3372/modules/dnn/src/layers/convolution_layer.cpp#L1007
https://github.com/opencv/opencv/blob/0f298a42032798d2d83017030621fcba885a3372/modules/dnn/src/layers/convolution_layer.cpp#L1009

rockzhan: here the prelu weights add 2 extra elements at the end, but it should add 2 extra element to the front to handle 1 channel  situation.

rockzhan: opencv/modules/dnn/src/layers/layers_common.simd.hpp
Line 97 in a75840d 
97   outptr2 = outptr1 = outptr0;

and here:
104   r0 = relu[i];
105   r1 = relu[i+1];
106   r2 = relu[i+2];

It will multiply the 3rd value and output.

dkurt: @rockzhan, May I ask you to add a test? Build a network with a single convolution layer followed by PReLU? You may find examples at https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_layers.cpp.

rockzhan: sure, let me try

rockzhan: @dkurt Hi, I add a test in the latest commit 9fb15dc3c55cbcfae2dae9afbde04a35368820d7.

rockzhan: I create a caffe model with 2 layers  **[dwconv(group=1) + prelu]**, and manually set 
conv weights 3 channels as: 
> **weight[0]=-1 weight[1]=-2 weight[2]=-3 no_bias**,  <br>

set prelu weights as:  <br>

> **weight={1,2,3}**

Then create image size: **3x16x16** all pixel value is **1**
So the dwconv outout should be:
>1 1 1 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;                   -1 -1 -1
>1 1 1   **conv**    -1 -1 -1  **= -9** &nbsp; **,channel 1 output all -9**
>1 1 1 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;                   -1 -1 -1      
> &nbsp;&nbsp;&nbsp;**... channel 2 output all -18**
> &nbsp;&nbsp;&nbsp;**... channel 3 output all -27**  <br>

The expect prelu output should be:
> -9 x 1 &nbsp;&nbsp;= -9 &nbsp;&nbsp;  **,channel 1 output all -9**
> -18 x 2 = -36  **,channel 2 output all -36**
> -27 x 3 = -81  **,channel 3 output all -81**

But current opencv prelu output:
> -9 x 3 &nbsp;&nbsp;= -27 &nbsp;&nbsp;  **,channel 1 output all -27**
> -18 x 3 = -54  **,channel 2 output all -54**
> -27 x 3 = -81  **,channel 3 output all -81**

This will affect all models contain DepthwiseConv+Prelu block in OpenCV.

modules/dnn/test/test_layers.cpp
```diff
@@ -777,6 +777,50 @@ TEST(Layer_PriorBox, squares)
...
+        net = readNetFromCaffe(proto, model);
+        ASSERT_FALSE(net.empty());
+    }
+    Mat sample = imread(findDataFile("dnn/3_16_16_all_1.jpg", false));
+    ASSERT_TRUE(!sample.empty());
+
+    if (sample.channels() == 4)
+        cvtColor(sample, sample, COLOR_BGRA2BGR);
+
+    Mat in_blob = blobFromImage(sample, 1.0f, Size(16, 16), Scalar(), false, false);
```
dkurt: You can generate an input blob by creating a `cv::Mat` with a shape `1x3x16x16` and fill it by ones.
In example, 
```cpp
int shape[] = {1, 3, 16, 16};
Mat in_blob(4, &shape[0], CV_32FC1, Scalar(1));
```

modules/dnn/test/test_layers.cpp
```diff
@@ -777,6 +777,50 @@ TEST(Layer_PriorBox, squares)
...
+    // prelu       weight={1,2,3}
+    //   |
+    //   v
+    // output      out size 3x14x14  if right: out[0]=-9 out[0]=-36 out[0]=-81 
+    //             but current opencv output: out[0]=-27 out[0]=-54 out[0]=-81
+
+    Net net;
+    {
+        const string proto = findDataFile("dnn/conv_prelu.prototxt", false);
+        const string model = findDataFile("dnn/conv_prelu.caffemodel", false);
```
dkurt: Please build a model in runtime using `LayerParams` and `addLayerToPrev` methods. This way `.prototxt` and `.caffemodel` are not required.
https://github.com/opencv/opencv/blob/4ec456f0a0993457120b916b36bde1d6df93e37b/modules/dnn/test/test_layers.cpp#L581-L597

modules/dnn/test/test_layers.cpp
rockzhan: ok, let me modify it.

modules/dnn/test/test_layers.cpp
rockzhan: done.

rockzhan: @dkurt

modules/dnn/test/test_layers.cpp
```diff
@@ -777,6 +777,97 @@ TEST(Layer_PriorBox, squares)
...
+    net.addLayerToPrev(lp.name, lp.type, lp);
+
+    //layer 2: prelu
+    LayerParams lpr;
+    lpr.name = "dw_relu";
+    lpr.type = "PReLU";
+    std::vector<int> weightsShapep(2);
+    weightsShapep[0] = 1;  // #outChannels
+    weightsShapep[1] = 3;  // #inpChannels / group
+    Mat weightsp(weightsShapep, CV_32F, Scalar(1));
```
dkurt: Just create a `cv::Mat` with a single row and number of columns equals to number of output channels (3):
```cpp
Mat weightsp(1, 3, CV_32F, Scalar(1));
```
It's better to replace number of output channels to a special constant at the beginning of the test. Use it in `lp.set("num_output", 3);` too.

modules/dnn/src/layers/layers_common.simd.hpp
dkurt: Please make it in the same way as it is for bias values. Otherwise it doesn't work for number of output channels 6 and groups number 3, in example. I mean two `if` conditions `if( i+2 >= outCn )` and `if( i+1 >= outCn )`.

modules/dnn/test/test_layers.cpp
```diff
@@ -777,6 +777,97 @@ TEST(Layer_PriorBox, squares)
...
+    Mat out = net.forward();
+
+    const int numDetections = out.size[2];
+    for (int i = 0; i < numDetections; ++i)
+    {
+        float value = out.ptr<float>(0, 2, i)[2];
+        std::cout << value << " "<<std::endl;
+    }
+
+    ASSERT_EQ(out.ptr<float>(0, 0, 0)[2], -9);
```
dkurt: Please generate a target output and compare it with a received one by `normAssert` or make a `1x3x3x3` input instead of `1x3x16x16` so output will has a shape `1x3x1x1` which is easier to compare inside a loop over output channels.

modules/dnn/test/test_layers.cpp
```diff
@@ -777,6 +777,97 @@ TEST(Layer_PriorBox, squares)
...
+    Mat in_blob(4, &shape[0], CV_32FC1, Scalar(1));
+
+    net.setInput(in_blob);
+    Mat out = net.forward();
+
+    const int numDetections = out.size[2];
+    for (int i = 0; i < numDetections; ++i)
+    {
+        float value = out.ptr<float>(0, 2, i)[2];
+        std::cout << value << " "<<std::endl;
```
dkurt: Please remove debugging prints.

dkurt:
@rockzhan, this is a place you need to fix too:
https://github.com/opencv/opencv/blob/895e10c3177f76d59dcd1095913a28a4076376e4/modules/dnn/src/layers/convolution_layer.cpp#L738-L742
It's executed if optimizations are not available.

modules/dnn/test/test_layers.cpp
rockzhan: sure

alalek:

So, please:

modules/dnn/src/layers/layers_common.simd.hpp
rockzhan: I tested group = 2 or 3, the result is right. 
Because opencv make sure **outCn=1**, when use group convolution.
https://github.com/opencv/opencv/blob/b2cff44084b91c781d86b98bd2c99f90191a4ea1/modules/dnn/src/layers/convolution_layer.cpp#L513

Here

rockzhan: and here:
https://github.com/opencv/opencv/blob/b2cff44084b91c781d86b98bd2c99f90191a4ea1/modules/dnn/src/layers/convolution_layer.cpp#L712-L713

modules/dnn/test/test_layers.cpp
rockzhan: let me try

rockzhan:
I will try that. Do you know how to disable the optimizations? @dkurt

rockzhan:
Thanks, I have rebased it. @alalek

modules/dnn/src/layers/layers_common.simd.hpp
dkurt: @rockzhan, try with number of output channels == 6 and group 3.

rockzhan: yes I tested, output is right when use AVX2(number of output channels == 6 and group 3). and now I'm try to disable the optimizations.

dkurt: @rockzhan, `-DCV_DISABLE_OPTIMIZATION`

dkurt: > yes I tested, output is right when use AVX2(number of output channels == 6 and group 3). and now I'm try to disable the optimizations.

I just replaced 
```cpp
const int num_output = 3;
```
to
```cpp
const int num_output = 6;
```
and got
```
Expected: (normL1) <= (l1), actual: 8 vs 1e-05
Expected: (normInf) <= (lInf), actual: 32 vs 0.0001
```
Let's create a parametric test and check it on buildbot:
```cpp
typedef TestWithParam<int> Layer_Test_DWconv_Prelu;
TEST_P(Layer_Test_DWconv_Prelu, Accuracy)
{
    const int num_output = GetParam();
    ...
}
INSTANTIATE_TEST_CASE_P(/**/, Layer_Test_DWconv_Prelu, Values(3, 6));
```

rockzhan: >Expected: (normL1) <= (l1), actual: 8 vs 1e-05
>Expected: (normInf) <= (lInf), actual: 32 vs 0.0001

This has 2 reason:
1. previous commit (before 26953c8) not fix without the optimizations, and now fixed.
2. In the Test, the target is fixed before, not flexible, and now fixed.

I will add a parametric test soon.

modules/dnn/test/test_layers.cpp
```diff
@@ -777,6 +777,125 @@ TEST(Layer_PriorBox, squares)
...
+    //   |
+    //   v
+    // output      out size 3x14x14  if right: out[0]=-8 out[0]=-32 out[0]=-72
+    //             but current opencv output: out[0]=-24 out[0]=-48 out[0]=-72
+
+    const int num_input = 6;    //inpChannels
+    const int group = 3;        //outChannels=group when group>1
+    int num_output = 6;         //only need to set when group=1
+
+    num_output = group == 1 ? num_output : group;   //conv or group conv
```
dkurt: Please do not overcomplicate your test. I just asked you to check a case with number of input channels == 3, number of output channels == 6 and a group number 3 because it works in a wrong way for now.

rockzhan:
@dkurt I add a flexible parametric test.

modules/dnn/test/test_layers.cpp
rockzhan: sorry

dkurt:
@rockzhan, I modified a bit the test to show which set of parameters need to be tuned. Please see https://github.com/opencv/opencv/pull/11649/commits/b5b065aafac0d438e880e95ba15337d8e67013b1.

rockzhan:
I see, it's very helpful.

rockzhan:
maybe this value is no use when use group conv. 
> num_output

It determined by group. do you think ? @dkurt

rockzhan:
ok :sweat_smile:
then I change conv code to 
>            r0 = relu[i]; r1 = relu[i+1]; r2 = relu[i+2];
>            if( i+2 >= outCn )
>            {
>                r2 = r1;
>                if( i+1 >= outCn )
>                    r2 = r1 = r0;
>            }

and change the test back soon.

rockzhan:
you just want to see what will happen when outCn == 2, right? @dkurt

dkurt:
@rockzhan, exactly.

rockzhan:
good, do you think its a1eb776 ok now? @dkurt

PR #11650, opened 2018-06-01
Default deep learning backend

dkurt:
### This pullrequest changes

Intel's Inference Engine as a default backend is no preferable backend is specified.

Added manual `net.setPreferableBackend(DNN_BACKEND_DEFAULT)` in tests and samples.


vpisarev:
@dkurt, I think, the logic should be different. It's better if DNN_BACKEND_DEFAULT is a meta-value that means "some backend, used by default" instead of some concrete backend. That is, we need to add some DNN_BACKEND_CPP (or DNN_BACKEND_OPENCV), which means our own implementation of the layers. And then dnn.setPreferableBackend(DNN_BACKEND_DEFAULT) will set DNN_BACKEND_INFERENCE_ENGINE when OpenCV is built with Inference Engine and DNN_BACKEND_CPP otherwise. Maybe in the future, if/when we get support for other acceleration engines, some of them will be used by default on certain platforms.

dkurt:
@vpisarev, Replaced to `DNN_BACKEND_OPENCV` so `DNN_BACKEND_DEFAULT` is `DNN_BACKEND_INFERENCE_ENGINE` if OpenCV is built with it and is `DNN_BACKEND_OPENCV` otherwise.

samples/dnn/classification.cpp
```diff
@@ -21,12 +21,15 @@ const char* keys =
     "{ height      |   | Preprocess input image by resizing to a specific height. }"
     "{ rgb         |   | Indicate that model works with RGB input images instead BGR ones. }"
     "{ backend     | 0 | Choose one of computation backends: "
-                        "0: default C++ backend, "
+                        "0: automatically (by default), "
                         "1: Halide language (http://halide-lang.org/), "
-                        "2: Intel's Deep Learning Inference Engine (https://software.intel.com/openvino-toolkit)}"
+                        "2: Intel's Deep Learning Inference Engine (https://software.intel.com/openvino-toolkit), "
+                        "3: default implementation }"
```
alalek: > 0: automatically (by default)
> ...
> 3: default implementation

Perhaps "default" duplication should be avoided.
Can we change description to something like "3: OpenCV implementation"?

dkurt: Sure. Updated.

vpisarev:
great! :+1:

PR #11653, opened 2018-06-01
convertFp16: documentation edit

pauljurczak:
convertFp16 doesn't convert from FP16: 

> input array has to have type of CV_32F or CV_16S.

### DIFF
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -532,7 +532,7 @@ CV_EXPORTS_W void convertScaleAbs(InputArray src, OutputArray dst,
 
 /** @brief Converts an array to half precision floating number.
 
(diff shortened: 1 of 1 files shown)


alalek:
There is no OpenCV native support for `CV_FP16` type yet. Data storage is emulated via `CV_16S`.

pauljurczak:
What does it mean to convert from FP16 then? Input array of `CV_16S` and output array of `CV_16S`?

alalek:
FP32 <-> FP16 (via 16S storage).

pauljurczak:
So there are 2 use cases:

1. FP32 -> FP16
2. FP16 -> FP32

Right? If so, it is not clear from function description.

alalek:
Right.

Proposed patch is not correct anyway.

pauljurczak:
Of course. Now that I know what it does, I will redo the patch.

PR #11654, opened 2018-06-01
core: drop unnecessary duplicate check

### LINKED ISSUE #11648: remove unneeded check

- OpenCV => github master
- Operating System / Platform => ubuntu 1804
- Compiler => gcc 7

##### Detailed description

In  opencv/modules/core/include/opencv2/core/cvstd.inl.hpp:

> 

inline
String::String(const std::string& str)
    : cstr_(0), len_(0)
{
    if (!str.empty())
    {
        size_t len = str.size();
        if (len) memcpy(allocate(len), str.c_str(), len);
    }
}
the variable len can be removed.

> 
inline
String& String::operator = (const std::string& str)
{
    deallocate();
    if (!str.empty())
    {
        size_t len = str.size();
        if (len) memcpy(allocate(len), str.c_str(), len);
    }
    return *this;
}

same problem.
##### Steps to reproduce
see the code.

alalek:
resolves #11648

### DIFF
--- a/modules/core/include/opencv2/core/cvstd.inl.hpp
+++ b/modules/core/include/opencv2/core/cvstd.inl.hpp
@@ -77,11 +77,8 @@ inline
 String::String(const std::string& str)
     : cstr_(0), len_(0)
 {
-    if (!str.empty())
-    {
-        size_t len = str.size();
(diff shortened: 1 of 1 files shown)


vpisarev:
:+1:

PR #11657, opened 2018-06-01
Fix multiple networks with Intel's Inference Engine backend

dkurt:
### This pullrequest changes

* Use an updated API to create separate inference requests.
* Load unique plugin per device.


vpisarev:
:+1:

PR #11677, opened 2018-06-02
core:ppc Fix java CoreTest/testMahalanobis

### LINKED ISSUE #11676: core:ppc Java test fail: CoreTest/testMahalanobis
- OpenCV => 3.4/master
- Compiler => all gcc versions
- Platform => ppc64le GNU/Linux

##### Failed

```
[junit] Tests run: 147, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.685 sec
[junit] Test org.opencv.test.core.CoreTest FAILED
```

```
junit.framework.AssertionFailedError
at org.opencv.test.core.CoreTest.testMahalanobis(Unknown Source)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at org.opencv.test.OpenCVTestCase.runTest(Unknown Source)
at org.opencv.test.OpenCVTestCase.runBare(Unknown Source)
```
##### Build steps
https://ocv-power.imavr.com/#/builders/1/builds/3

seiko2plus:
resolves #11676 

````
force_builders=Custom
allow_multiple_commits=1
````

options for alternative buildbot for powerpc
https://ocv-power.imavr.com/#/opencv_pullrequests
````
pw_compilers=gcc-4.9, gcc-5, gcc-6, clang-4
# by default java test disabled due this issue
pw_disable_tests=rgbd, shape
#pw_cmake_definitions=OPENCV_EXTRA_MODULES_PATH:OFF
###
#pw_disable_perf_tests=*
#pw_filter_tests=core:*Mahalanobis*, java
````

### DIFF
--- a/modules/core/misc/java/test/CoreTest.java
+++ b/modules/core/misc/java/test/CoreTest.java
@@ -947,12 +947,16 @@ public class CoreTest extends OpenCVTestCase {
     }
 
     public void testMahalanobis() {
+        Mat src = new Mat(matSize, matSize, CvType.CV_32F);
+        Core.randu(src, -128, 128);
+
         Mat covar = new Mat(matSize, matSize, CvType.CV_32F);
         Mat mean = new Mat(1, matSize, CvType.CV_32F);
-        Core.calcCovarMatrix(grayRnd_32f, covar, mean, Core.COVAR_ROWS | Core.COVAR_NORMAL, CvType.CV_32F);
+        Core.calcCovarMatrix(src, covar, mean, Core.COVAR_ROWS | Core.COVAR_NORMAL, CvType.CV_32F);
         covar = covar.inv();
-        Mat line1 = grayRnd_32f.row(0);
-        Mat line2 = grayRnd_32f.row(1);
+
+        Mat line1 = src.row(0);
+        Mat line2 = src.row(1);
 
         double d = Core.Mahalanobis(line1, line1, covar);
 
--- a/modules/core/src/system.cpp
+++ b/modules/core/src/system.cpp
@@ -739,7 +739,6 @@ int64 getCPUTickCount(void)
 
 int64 getCPUTickCount(void)
 {
-    int64 result = 0;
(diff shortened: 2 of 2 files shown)


alalek:
I believe, we should avoid that hacks.
Floating-point numerical errors must be taken into account.
We should not block "fma" operations (actually, because they are often performed with better accuracy).

I believe test must be get fixed instead:
- comparison of floating-point results should be performed with tolerance. `assertEquals(0.0, d);` is incorrect check.
- check that input data is stable for testing algorithm (adding epsilon noise to input should not "break" result)

vpisarev:
@seiko2plus, I agree here with @alalek. We can introduce some kind of fma intrinsic and gradually add it in some accuracy-critical places, but we should not rely on it. Our tests should be written in such a way that they pass regardless of the hardware FMA availability.

vpisarev:
:+1:

seiko2plus:
Test fail on "assertTrue(d > 0.0);", because 'Mahalanobis' return NaN,
after trace sequence calls, I just found that inverse of covariance matrix return different result compared with disable cpu optimization and on x86 too.

so that lead me to check 'LU32f 'the algorithm just working fine under FMA and pass all other tests, but with this case FMA produce negative result due to invalid input, since its only stable with symmetric positive definite matrices.

however the current patch is decreasing boundaries just to avoid Mahalanobis to return NaN and pass the test.

vpisarev:
@seiko2plus, ok, let's merge it because the patch itself does not do any harm. The accuracy issues can be addressed in the further patches :+1:

seiko2plus:
@vpisarev Alright, thanks

PR #11692, opened 2018-06-05
core: set default logger level to 'warning'

alalek:
should hide unnecessary 'info' messages


alalek:
BTW, Should we have different settings for Debug/Release builds? For example:
- Release: "WARNINGS" level
- Debug: "INFO" level

mshabunin:
>Should we have different settings for Debug/Release builds?

@alalek , I think it will be great.

PR #11702, opened 2018-06-06
Fixed setting of frame size properties for MSMF-based VideoCapture

terfendail:
### This pullrequest changes
Fix setting of frame size properties for MSMF-based VideoCapture

### DIFF
--- a/modules/videoio/src/cap_msmf.cpp
+++ b/modules/videoio/src/cap_msmf.cpp
@@ -720,6 +720,7 @@ protected:
     MediaType nativeFormat;
     MediaType captureFormat;
     int outputFormat;
+    UINT32 requestedWidth, requestedHeight;
     bool convertFormat;
     UINT32 aspectN, aspectD;
     MFTIME duration;
@@ -741,12 +742,15 @@ CvCapture_MSMF::CvCapture_MSMF():
     videoFileSource(NULL),
     videoSample(NULL),
     outputFormat(CV_CAP_MODE_BGR),
+    requestedWidth(0),
+    requestedHeight(0),
     convertFormat(true),
     aspectN(1),
     aspectD(1),
     sampleTime(0),
     isOpen(false)
 {
+    configureHW(true);
 }
 CvCapture_MSMF::CvCapture_MSMF(int index) : CvCapture_MSMF() { open(index); }
 CvCapture_MSMF::CvCapture_MSMF(const cv::String& _filename) : CvCapture_MSMF() { open(_filename); }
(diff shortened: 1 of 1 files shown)


modules/videoio/src/cap_msmf.cpp
```diff
@@ -853,31 +857,28 @@ bool CvCapture_MSMF::configureOutput(UINT32 width, UINT32 height, double prefFra
...
-                    ((MT.width * MT.height) > (MTFallback.width * MTFallback.height)) ||
-                    (((MT.width * MT.height) == (MTFallback.width * MTFallback.height)) && getFramerate(MT) > getFramerate(MTFallback) && (prefFramerate == 0 || getFramerate(MT) <= prefFramerate)))
-                {
-                    dwStreamFallback = (int)dwStreamTest;
-                    MTFallback = MT;
-                }
-                if (MT.width == width && MT.height == height)
+                if (dwStreamBest < 0 ||
+                    (
+                        ((width > MT.width ? width - MT.width : width > 0 ? MT.width - width : 0) + (height > MT.height ? height - MT.height : height > 0 ? MT.height - height : 0)) <
```
vpisarev: the conditions look too complex. Can we simplify them?
e.g.

<pre>
int best_diff = INT_MAX;

for( ... ) {
...
int temp_width = width > 0 ? width : MT.width;
int temp_height = height > 0 ? height : MT.height;
int diff = abs(temp_width - MT.width) + abs(temp_height - MT.heighT);
if( diff < best_diff ) { best_width = MT.width; best_height = MT.height; best_diff = diff; }
}
</pre>

terfendail: I've refactored the code. Now it should be more readable

modules/videoio/src/cap_msmf.cpp
```diff
@@ -1218,13 +1218,13 @@ double CvCapture_MSMF::getProperty( int property_id ) const
         return aspectN;
     else if (property_id == CV_CAP_PROP_SAR_DEN)
         return aspectD;
+    else if (property_id == CV_CAP_PROP_FRAME_WIDTH)
+        return isOpen ? captureFormat.width : requestedWidth;
```
alalek: > isOpen ?

How we can create `cv::VideoCapture` where `isOpen == false`? In this case VideoCapture should trying the next backend from the list of enabled/available, and lost "this" instance.

How should we write MSMF test to test both branches of this condition?

terfendail:
Enabled `MODE_HW` by default and simplified set/get logic regarding `isOpen`

PR #11703, opened 2018-06-06
calib3d: chessboard detector - replace OpenCV C API

alalek:
Review link: https://github.com/opencv/opencv/pull/11703/files?utf8=%E2%9C%93&diff=split&w=1

```
docker_image:Custom=fedora:28
```


alalek:
:+1:

PR #11704, opened 2018-06-06
apps: catch() with "const reference"

alalek:
related #11626 

> apps/interactive-calibration/main.cpp:220:31: warning: catching polymorphic type 'class std::runtime_error' by value [-Wcatch-value=]

```
docker_image:Custom=fedora:28
```


vpisarev:
:+1:

PR #11705, opened 2018-06-06
samples: fix callback function type

alalek:
related #11626 

> samples/tapi/clahe.cpp:52:75: warning: cast between incompatible function types from 'void (*)(int)' to 'cv::TrackbarCallback' {aka 'void (*)(int, void*)'} [-Wcast-function-type]
> samples/tapi/clahe.cpp:53:77: warning: cast between incompatible function types from 'void (*)(int)' to 'cv::TrackbarCallback' {aka 'void (*)(int, void*)'} [-Wcast-function-type]

```
docker_image:Custom=fedora:28
```

### DIFF
--- a/samples/tapi/clahe.cpp
+++ b/samples/tapi/clahe.cpp
@@ -14,7 +14,7 @@ Ptr<CLAHE> pFilter;
 int tilesize;
 int cliplimit;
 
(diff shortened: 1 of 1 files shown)


alalek:
:+1:

PR #11719, opened 2018-06-08
core: align cv::AutoBuffer API with std::vector/std::array

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1662

- added `.data()` methods
- added `operator[] (int i)`
- extend checks support to generic and debug-only cases
- deprecate existed `operator* ()`

```
docker_image:Custom=fedora:28
allow_multiple_commits=1
```


vpisarev:
:+1:

PR #11730, opened 2018-06-08
videoio: update librealsense to API 2.0

paroj:
this enables the usage of current sensors, while dropping support for
legacy devices, see:
https://github.com/IntelRealSense/librealsense#overview

Given limited resources, and that the legacy sensors where not that
great, I think we should focus on v2.

### DIFF
--- a/modules/videoio/src/cap_librealsense.cpp
+++ b/modules/videoio/src/cap_librealsense.cpp
@@ -10,20 +10,19 @@
 namespace cv
 {
 
-VideoCapture_LibRealsense::VideoCapture_LibRealsense(int index)
+VideoCapture_LibRealsense::VideoCapture_LibRealsense(int) : mAlign(RS2_STREAM_COLOR)
 {
     try
     {
-        mDev = mContext.get_device(index);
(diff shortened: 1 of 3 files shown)


modules/videoio/src/cap_librealsense.cpp
```diff
@@ -10,20 +10,19 @@
...
-        mDev->start();
+        rs2::config config;
+        // Configure all streams to run at VGA resolution at default fps
+        config.enable_stream(RS2_STREAM_DEPTH, 640, 480, RS2_FORMAT_Z16);
+        config.enable_stream(RS2_STREAM_COLOR, 640, 480, RS2_FORMAT_BGR8);
+        config.enable_stream(RS2_STREAM_INFRARED, 640, 480, RS2_FORMAT_Y8);
+        mPipe.start();
     }
-    catch (rs::error&)
+    catch (rs2::error&)
```
alalek: Could you please "catch" by "const reference"? (here and below)

Link: http://www.riptutorial.com/cplusplus/example/9212/best-practice--throw-by-value--catch-by-const-reference

PR #11735, opened 2018-06-09
videoio(msmf): fix reopen confition in configureHW()

alalek:
`camid` value is always -1 after close().

### DIFF
--- a/modules/videoio/src/cap_msmf.cpp
+++ b/modules/videoio/src/cap_msmf.cpp
@@ -805,7 +805,7 @@ bool CvCapture_MSMF::configureHW(bool enable)
                     if (SUCCEEDED(D3DMgr->ResetDevice(D3DDev.Get(), mgrRToken)))
                     {
                         captureMode = MODE_HW;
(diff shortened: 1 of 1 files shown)


modules/videoio/src/cap_msmf.cpp
```diff
@@ -805,7 +805,7 @@ bool CvCapture_MSMF::configureHW(bool enable)
                     if (SUCCEEDED(D3DMgr->ResetDevice(D3DDev.Get(), mgrRToken)))
                     {
                         captureMode = MODE_HW;
-                        return reopen ? camid >= 0 ? open(prevcam) : open(prevfile.c_str()) : true;
+                        return reopen ? (prevfile.empty() ? open(prevcam) : open(prevfile.c_str())) : true;
```
terfendail: IMHO it would be better to check for `prevcam >= 0` to retain the state prior to `configureHW` if neither camera nor file was open. When VideoCapture tries to open file with empty name nothing will be opened. When VideoCapture tries to open camera with index -1 it will open first available camera.

alalek: updated

PR #11743, opened 2018-06-11
samples: don't call twice of VideoCapture::open()

alalek:
To avoid `EBUSY` errors in samples (some cameras/backends may require delay after `release()` call)


vpisarev:
:+1:

PR #11748, opened 2018-06-12
VideoCapture: add exception error mode similar to fstream

paroj:


paroj:
@alalek @mshabunin any objections to this? I would rebase this for 4.1.0

paroj:
rebased & testcase added

paroj:
@alalek anything else needed here?

modules/videoio/src/cap.cpp
```diff
@@ -112,10 +112,13 @@ bool VideoCapture::open(const String& filename, int apiPreference)
                             CV_LOG_WARNING(NULL, cv::format("VIDEOIO(%s): can't create capture", info.name));
                     }
                 } catch(const cv::Exception& e) {
+                    if(errorOnFail) throw;
```
alalek: This function iterates over all available backends, so re-throwing exceptions from here break normal control flow.
For example, FFmpeg backend can't propertly handle "/dev/video0", but V4L or GStreamer can.

Probably it make sense only if `apiPreference != CAP_ANY`.

modules/videoio/src/cap.cpp
```diff
@@ -126,6 +129,10 @@ bool VideoCapture::open(const String& filename, int apiPreference)
             }
         }
     }
+
+    if (errorOnFail)
+        CV_Error(Error::StsInternal, "");
```
alalek: Lets make exception messages useful - dump `filename` at least.

paroj:
- added context information where applicable 
- converted all caught exceptions to `StsError`
- only rethow inside open, if we are not looping

modules/videoio/src/cap.cpp
```diff
@@ -112,10 +112,13 @@ bool VideoCapture::open(const String& filename, int apiPreference)
                             CV_LOG_WARNING(NULL, cv::format("VIDEOIO(%s): can't create capture", info.name));
                     }
                 } catch(const cv::Exception& e) {
+                    if(errorOnFail && apiPreference != CAP_ANY) throw Exception(Error::StsError, e.err, e.func, e.file, e.line);
                     CV_LOG_ERROR(NULL, cv::format("VIDEOIO(%s): raised OpenCV exception:\n\n%s\n", info.name, e.what()));
```
alalek: > if(errorOnFail && apiPreference != CAP_ANY) throw Exception(...)
> if(errorOnFail && apiPreference != CAP_ANY) CV_Error(...)

Why?


```
if(errorOnFail && apiPreference != CAP_ANY) throw;
```

is enough here and below.

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -782,9 +782,15 @@ class CV_EXPORTS_W VideoCapture
      */
     CV_WRAP String getBackendName() const;
 
+    /**
+     * raise Error::StsError if a method is not successful
```
alalek: > `Sts::Error`

Probably we don't need to declare such limitations.

"Switches into exceptions mode - raise exception if some requested operation is failed" ?

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -782,9 +782,15 @@ class CV_EXPORTS_W VideoCapture
...
+     */
+    CV_WRAP void exceptions(bool enable) { errorOnFail = enable; }
```
alalek:
```
CV_WRAP void setExceptionsMode(bool enable) { errorOnFail = enable; }
CV_WRAP bool getExceptionsMode() { return errorOnFail; }
```

`errorOnFail` => `enableExceptions` / `allowExceptions`

paroj: tried to follow `fstream` here: https://en.cppreference.com/w/cpp/io/basic_ios/exceptions

but I guess using get/ set is more consistent here.

alalek:
```
CV_WRAP void exceptions(bool enable) { errorOnFail = enable; }
CV_WRAP bool exceptions() const { return errorOnFail; }
```

This work for C++ well, but OpenCV should support bindings like Python/Java.
So get/set is more consistent and straight-forward way to do that.

(we can use `CV_WRAP_AS(...)` here but it would lead for code inconsistency between C++ and Python/Java)

paroj:
@alalek 
- reverted the exception wrapping
- adapted variable/ method naming to your suggestions

anything else?

PR #11755, opened 2018-06-13
dnn: add runtime parameter OPENCV_DNN_BACKEND_DEFAULT

alalek:
to control `DNN_BACKEND_DEFAULT` enumeration value behavior


dkurt:
:+1:

PR #11764, opened 2018-06-14
videoio(msmf): cleanup unused methods from ComPtr wrapper

alalek:
Also use Win32 TRUE/FALSE in SetUINT32() calls for bool parameters.


terfendail:
Could `ComPtr` class be completely replaced with `cv::Ptr`?

alalek:
I see these issues with `cv::Ptr<>` replacing:
1. missing QueryInterface() method.
2. `cv::Ptr` requires different destructors for different types.

PR #11768, opened 2018-06-14
videoio: MSMF async reader for camera stream

alalek:
Synchronized reading from camera with heavy frame processing
provides bad effects (huge frame latency, processing frames from the past).
Generally, there is no way to process each frame and some frames will be dropped.
Allow preventive frame dropping to reduce lag of processed frames.

This mode is applied to cameras only (opened by 'index').

### DIFF
--- a/modules/videoio/src/cap_msmf.cpp
+++ b/modules/videoio/src/cap_msmf.cpp
@@ -93,6 +93,8 @@
 
 #include <comdef.h>
 
+#include <shlwapi.h>  // QISearch
+
 struct IMFMediaType;
 struct IMFActivate;
 struct IMFMediaSource;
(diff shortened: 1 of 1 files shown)


modules/videoio/src/cap_msmf.cpp
```diff
@@ -958,10 +1041,116 @@ bool CvCapture_MSMF::open(const cv::String& _filename)
...
+    if (dwResult == WAIT_TIMEOUT)
+    {
+        return E_PENDING;
+    }
+    else if (dwResult != WAIT_OBJECT_0)
+    {
+        return HRESULT_FROM_WIN32(GetLastError());
+    }
+
+    if (!bEOS)
```
terfendail: It looks like this check will always be true

vpisarev:
@terfendail checked the patch and it works well, thanks! :+1:

PR #11772, opened 2018-06-15
Added Size_::aspectRatio

PhilLab:
### This pullrequest changes
Aspect ratio computation is often needed when working with images. This is a convenience function to reduce the amount of code needed

## Things open for discussion:

1. Do we actually want to extend core type functionality like this?
2. Should we handle ``height == 0``, e.g. by returning -1 ?
3. Which (other) assertions do we want to impose (cf. ``area()``)
4. Should we handle larger types ``_TP`` where ``static_cast<double>(height)`` will fail?


alalek:
2. current "NaN" value is fine.
4. double is enough for this.

PR #11773, opened 2018-06-15
dnn(ocl): fix external / predefined builtin configuration behavior

alalek:
`OPENCV_OCL4DNN_FORCE_AUTO_TUNING` should ignore existed configuration from:
- builtin predefined configurations (for Intel OpenCL iGPUs)
- external configuration (via `OPENCV_OCL4DNN_CONFIG_PATH`)

Prefer external configuration over builtin.


alalek:
:+1:

PR #11790, opened 2018-06-19
dnn: add a coco labels file for yolov3

berak:
11 classes were dropped from the coco dataset, see arXiv:1405.0312v3 (Table2), so yolov3 has 80 classes, but [rcnn](https://github.com/opencv/opencv_extra/blob/master/testdata/dnn/faster_rcnn_inception_v2_coco_2018_01_28.pbtxt) still has the original 90, also fill blank names in object_detection_classes_coco.txt [from here](https://github.com/nightrome/cocostuff/blob/master/labels.md)

### DIFF
--- a/samples/data/dnn/object_detection_classes_yolov3.txt
+++ b/samples/data/dnn/object_detection_classes_yolov3.txt
@@ -9,7 +9,6 @@ truck
 boat
 traffic light
 fire hydrant
-street sign
 stop sign
 parking meter
 bench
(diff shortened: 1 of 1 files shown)


dkurt:
:+1: Looks good to me! Thank you!

I've tried Faster-RCNN models and it looks like that there is a bug with scores or we need different labels because a `person` is matched as a `toothbrush`.

dkurt:
Got it. Internally, an every predicted bounding box is associated with a vector of 90 scores (classes only) but not with a vector of 91 values (score of background + classes) as it is for SSD-based models.
screenshot from 2018-06-20 16-15-14
A replacement from
```python
classId = int(detection[1]) - 1  # Skip background label
```
to
```python
classId = int(detection[1]) 
```
helps (but it requires labels list with `background` for SSDs).

PR #11794, opened 2018-06-19
Kinect2 support added to OpenNI2 backend

savuor:
### This pullrequest changes
* added support for Kinect 2 drivers (`libfreenect2` and `kinect2` branch of OpenNI2)
* IR sensor support added
* checking available video modes added (before setting pre-defined one)
* less exceptions in case of unsupported options (for example getting baseline and focal length is used in case of disparity map generation only, there's no need to throw exception in other cases)
* a little refactoring

### DIFF
--- a/modules/videoio/src/cap_openni2.cpp
+++ b/modules/videoio/src/cap_openni2.cpp
@@ -175,9 +175,9 @@ protected:
     // Cameras settings:
     // TODO find in OpenNI function to convert z->disparity and remove fields "baseline" and depthFocalLength_VGA
     // Distance between IR projector and IR camera (in meters)
-    mutable double baseline;
+    double baseline;
     // Focal length for the IR camera in VGA resolution (in pixels)
-    mutable int depthFocalLength_VGA;
+    int depthFocalLength_VGA;
 
     // The value for shadow (occluded pixels)
     int shadowValue;
--- a/modules/videoio/src/cap_openni2.cpp
+++ b/modules/videoio/src/cap_openni2.cpp
@@ -70,6 +70,35 @@
 #include "PS1080.h"
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+static cv::Mutex initOpenNI2Mutex;
+
+struct OpenNI2Initializer
+{
+public:
+    static void init()
+    {
+        cv::AutoLock al(initOpenNI2Mutex);
+        static OpenNI2Initializer initializer;
+    }
+
+private:
+    OpenNI2Initializer()
+    {
+        // Initialize and configure the context.
+        openni::Status status = openni::OpenNI::initialize();
+        if (status != openni::STATUS_OK)
+        {
+            CV_Error(CV_StsError, std::string("Failed to initialize:") + openni::OpenNI::getExtendedError());
+        }
+    }
+
+    ~OpenNI2Initializer()
+    {
+        openni::OpenNI::shutdown();
+    }
+};
+
 class CvCapture_OpenNI2 : public CvCapture
 {
 public:
@@ -118,8 +147,8 @@ protected:
     IplImage* retrieveGrayImage();
     IplImage* retrieveIrImage();
 
-    openni::Status toggleStream(int stream, bool toggle);
-    bool readCamerasParams(bool throwException = false);
+    void toggleStream(int stream, bool toggle);
+    void readCamerasParams();
 
     double getDepthGeneratorProperty(int propIdx) const;
     bool setDepthGeneratorProperty(int propIdx, double propVal);
@@ -133,7 +162,6 @@ protected:
     // OpenNI context
     openni::Device device;
     bool isContextOpened;
-    openni::Recorder recorder;
 
     // Data generators with its metadata
     std::vector<openni::VideoStream> streams;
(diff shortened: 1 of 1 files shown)


modules/videoio/src/cap_openni2.cpp
```diff
@@ -191,72 +193,113 @@ openni::VideoMode CvCapture_OpenNI2::defaultStreamOutputMode(int stream)
...
+CvCapture_OpenNI2::~CvCapture_OpenNI2()
+{
+    for (int i = 0; i < CV_MAX_NUM_STREAMS; ++i)
+    {
+        streamFrames[i].release();
+        streams[i].stop();
+        streams[i].destroy();
+    }
+    device.close();
+    openni::OpenNI::shutdown();
```
alalek: > openni::OpenNI::shutdown();

How does it work with multiple VideoCapture instances?

IMHO, any system/library-wide shutdown calls must be avoided.

modules/videoio/src/cap_openni2.cpp
```diff
@@ -191,72 +193,113 @@ openni::VideoMode CvCapture_OpenNI2::defaultStreamOutputMode(int stream)
...
+    // default for Kinect2 camera
+    setProperty(CV_CAP_PROP_OPENNI2_MIRROR, 0.0);
 
     isContextOpened = true;
+}
 
-    setProperty(CV_CAP_PROP_OPENNI_REGISTRATION, 1.0);
+CvCapture_OpenNI2::~CvCapture_OpenNI2()
+{
+    for (int i = 0; i < CV_MAX_NUM_STREAMS; ++i)
```
alalek: > i < CV_MAX_NUM_STREAMS

It is `std::vector<>` now. So it is better to use .size() method instead of some "OpenCV" constant.

modules/videoio/src/cap_openni2.cpp
```diff
@@ -286,15 +329,27 @@ openni::Status CvCapture_OpenNI2::toggleStream(int stream, bool toggle)
...
-                                                 openni::OpenNI::getExtendedError()));
-                streams[stream].destroy();
-                return status;
+                if (vm[i].getPixelFormat() == dm.getPixelFormat() &&
+                    vm[i].getResolutionX() == dm.getResolutionX() &&
+                    vm[i].getResolutionY() == dm.getResolutionY() &&
+                    vm[i].getFps() == dm.getFps())
+                {
+                    // try to set default mode if available
+                    // by Jan 2011 xn::DepthGenerator supported VGA only
```
alalek: Does this comment still actual?

modules/videoio/src/cap_openni2.cpp
```diff
@@ -325,68 +380,14 @@ openni::Status CvCapture_OpenNI2::toggleStream(int stream, bool toggle)
...
 
-bool CvCapture_OpenNI2::readCamerasParams()
+bool CvCapture_OpenNI2::readCamerasParams(bool throwException)
 {
     double pixelSize = 0;
     if (streams[CV_DEPTH_STREAM].getProperty<double>(XN_STREAM_PROPERTY_ZERO_PLANE_PIXEL_SIZE, &pixelSize) != openni::STATUS_OK)
     {
-        CV_Error(CV_StsError, cv::format("CvCapture_OpenNI2::readCamerasParams : Could not read pixel size!\n"));
+        if (throwException)
+            CV_Error(CV_StsError, cv::format("CvCapture_OpenNI2::readCamerasParams : Could not read pixel size!\n"));
```
alalek: 1. Remove unnecessary `cv::format()`
2. Remove unnecessary `\n`

modules/videoio/src/cap_openni2.cpp
```diff
@@ -286,15 +329,27 @@ openni::Status CvCapture_OpenNI2::toggleStream(int stream, bool toggle)
...
+                    // try to set default mode if available
+                    // by Jan 2011 xn::DepthGenerator supported VGA only
+                    status = streams[stream].setVideoMode(defaultStreamOutputMode(stream));
+                    if (status != openni::STATUS_OK)
+                    {
+                        CV_Error(CV_StsError, cv::format("OpenCVKinect2 : Couldn't set %s stream output mode: %s\n",
+                                                         stream_names[stream],
+                                                         openni::OpenNI::getExtendedError()));
+                        streams[stream].destroy();
+                        return status;
```
alalek: Calling destroy() after CV_Error is just wrong.

modules/videoio/src/cap_openni2.cpp
```diff
@@ -191,72 +193,113 @@ openni::VideoMode CvCapture_OpenNI2::defaultStreamOutputMode(int stream)
...
         openni::OpenNI::shutdown();
         return;
     }
 
-    if (!readCamerasParams())
-    {
-        CV_Error(CV_StsError, cv::format("CvCapture_OpenNI2::CvCapture_OpenNI2 : Could not read cameras parameters\n"));
-        return;
-    }
+    readCamerasParams();
```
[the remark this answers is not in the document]
savuor: In fact, the only case we need to get camera params (baseline and focal lenght) is when generating disparity from depth map. Now I've fixed this, these params are read at disparity retrieving only.

modules/videoio/src/cap_openni2.cpp
savuor: Fixed (but no guarantee that this will work with multiple cameras since OpenNI's Kinect drivers don't support it)

modules/videoio/src/cap_openni2.cpp
savuor: I don't know about Xn and PrimeSense devices but OpenNI itself and its Kinect 2 drivers support a lot of resolutions
This comment is removed

vpisarev:
thanks! let's merge it in and fix the remaining issues (if any) afterwards :+1:

PR #11812, opened 2018-06-22
Implement Interp layer using Resize layer

dkurt:
### This pullrequest changes

Use Resize layer implementation to forward Interp layer from https://github.com/cdmh/deeplab-public


vpisarev:
:+1:

PR #11825, opened 2018-06-25
Detection QRCode

allnes:

### DIFF
--- a/modules/objdetect/src/qrcode.cpp
+++ b/modules/objdetect/src/qrcode.cpp
@@ -1,0 +1,760 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2017, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective owners.
+
+#include "precomp.hpp"
+#include "qrcode.hpp"
+// #include <opencv2/calib3d/calib3d.hpp>
+
+#include <limits>
+#include <cmath>
+#include <iostream>
+
+using namespace std;
+using namespace cv;
+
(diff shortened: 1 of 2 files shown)


vpisarev:
@allnes, thank you! see my comments below

modules/objdetect/src/qrcode.cpp
```diff
@@ -0,0 +1,760 @@
...
+
+#include "precomp.hpp"
+#include "qrcode.hpp"
+// #include <opencv2/calib3d/calib3d.hpp>
+
+#include <limits>
+#include <cmath>
+#include <iostream>
+
+using namespace std;
```
vpisarev: do not use `using namespace ...` inside OpenCV .cpp files. Put everything into `namespace cv { }`. Add `using std::vector;` etc. for each of the major type of function that you use many times; or just add `std::` for some functions and classes that you use just few times

modules/objdetect/src/qrcode.hpp
```diff
@@ -0,0 +1,56 @@
+// This file is part of OpenCV project.
```
vpisarev: remove this file from the patch. Put the class declaration right into the .cpp file. Put the external function declaration to the objdetect.hpp file.

modules/objdetect/src/qrcode.hpp
```diff
@@ -0,0 +1,56 @@
...
+// Copyright (C) 2017, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective owners.
+
+#ifndef QR_CODE_HPP_
+#define QR_CODE_HPP_
+
+using namespace std;
+using namespace cv;
+
+CV_EXPORTS void DetectionQRCode(InputArray in, OutputArray points,
```
vpisarev: Rename `DetectionQRCode` to `detectQRCode`. String decode should be passed by reference. What does the function return if there is no QR code in the image?

PR #11828, opened 2018-06-25
temporarily disabled OpenCL use in DNN module on Mac

vpisarev:
Did that since some of the tests fail. At some point, OpenCV DNN OpenCL branch will be expanded to support more generic OpenCL implementations; since OpenCL on Mac is a bit old and, e.g. does not support all the vendors extensions, like Intel extensions, some less sophisticated variant of kernels should be used there


alalek:
Perhaps it is better to disable this via CMake [here](https://github.com/opencv/opencv/blob/ab8022f74e65f6d4a33c8e64993adc87b0bfaa31/modules/dnn/CMakeLists.txt#L15).

vpisarev:
:+1:

PR #11829, opened 2018-06-25
Detect QR Code

allnes:


vpisarev:
let's merge it in and update then :+1:

PR #11836, opened 2018-06-26
Disable MSMF VideoCapture fallback to existing camera

### LINKED ISSUE #11780: Issue with VideoCapture when open an unavailable camera by index, the isOpen() funtction return true

- OpenCV =>4.00
- Operating System / Platform =>Windows 64 Bit
- Compiler =>Visual Studio 2017

##### Detailed description

I have compiled the OpenCV from source, but when I want to check how many USB or integrated cameras are available in the current machine. Before the `isOpen()` function returns `false ` when this index camera is not present so we I can stop the loop, but now `isOpen()` function still returns `true` even the camera is not present in the machine.

terfendail:
resolves #11780

### This pullrequest changes
Removed fallback to existing camera in case provided index is out of camera id range


vpisarev:
looks good :+1:

PR #11837, opened 2018-06-26
a part of PR #11364 (extended findNonZero & PSNR)

vpisarev:
Thanks to Tetragramm for #11364. While the extended LUT maybe too much for OpenCV (I'd prefer to have smaller, [bi- or tri-]linearly or [bi- or tri-]cubically interpolated LUTs), findNonZero and PSNR parts are much lighter additions and might be useful


vpisarev:
:+1:

PR #7153, opened 2016-08-23
new function filterHomographyDecompSolutionsByPointNormals

Vaesper:
### This pullrequest changes

This pull request adds a function that extends on the functionality in decomposeHomographyMat, making it possible to reduce the solution set from 4 to 2 possible solutions if additional information is present.

PR #11846, opened 2018-06-27
Filter homography decomp: updated PR #7153

vpisarev:
Filter homography decomposition solutions: updated PR #7153 by @Vaesper


vpisarev:
:+1:

PR #11852, opened 2018-06-28
Fix DLDT models import

dkurt:
### This pullrequest changes

* For models in Inference Engine's intermediate representation, set up output layers names and types. It's useful to determine output layers types in samples, i.e. object detection where postprocessing depends on it.

* Update object detection sample: apply NMS for an every type of outputs but not only for YOLOs (it's useful for models with multiple outputs or to regularize output for single models).


vpisarev:
:+1:

PR #11856, opened 2018-06-29
videoio(MSMF): remove "always true" check

alalek:
https://github.com/opencv/opencv/pull/11768#discussion_r198152112


vpisarev:
@terfendail, could you please review it?

alalek:
:+1:

PR #11862, opened 2018-06-30
WIP: HPX backend for OpenCV

Jakub-Golinowski:
### Introduction

My name is Jakub Golinowski and I am a student enrolled for Google Summer of Code 2018. I am working for Ste||ar Group on developing [HPX](https://github.com/STEllAR-GROUP/hpx) backend for parallelism in OpenCV. Here is the link to the GSoC project: https://summerofcode.withgoogle.com/projects/#5375652711104512. The main goal of the project is to allow better interoperability of HPX applications, which are using OpenCV. The primary use-case we have in mind is an application that is already using the HPX runtime environment and also performs computer vision operations using OpenCV. Including the HPX backend within the OpenCV will make such an application easier to implement and to control its parallelization behaviour using standard HPX approach.

So far there are 2 versions of HPX backend:

- HPX - primary version, described above, which assumes that user’s application starts the runtime before making calls to OpenCV functions that use cv::parallel_for_(). As mentioned in the introduction, this would be the preferred and hopefully only way to use the backend.
- HPX_STARTSTOP - back-up version of backend in which HPX runtime is started and stopped for each cv::parallel_for_() call. 

The backend was tested locally using small applications (they can be found here: https://github.com/Jakub-Golinowski/opencv_hpx_backend). The applications were behaving as expected.

### This pullrequest changes

- In the first commit (e2699a9) there are all source code changes required for backend functionality and proposition of cmake changes for HPX library inclusion.

- In the second commit (d4746dd) there is a proposition of changes required for OpenCV with HPX backend to enable the primary version of backend to pass the tests, however in this configuration there were linking errors encountered in the release mode build, which are not present in the debug mode build.

### Summary

This is still work in progress, but I would like to ask for feedback early especially on the CMake config. After that I will be able to run tests locally and make necessary fixes to backend if any will be required. 

Also, I would like to ask what would be required to have the HPX backend tested using OpenCV's CI? For now I added the cmake option WITH_HPX to choose HPX as the parallel backend and WITH_HPX_STARTSTOP to choose its back-up version (described above).

### DIFF
--- a/modules/core/test/test_main.cpp
+++ b/modules/core/test/test_main.cpp
@@ -3,4 +3,8 @@
 // of this distribution and at http://opencv.org/license.html.
 #include "test_precomp.hpp"
 
+#if defined(HAVE_HPX) && !defined(HPX_STARTSTOP)
+    #include <hpx/hpx_main.hpp>
+#endif
+
 CV_TEST_MAIN("cv")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,6 +252,8 @@ OCV_OPTION(WITH_WIN32UI        "Build with Win32 UI Backend support"         ON 
 OCV_OPTION(WITH_QUICKTIME      "Use QuickTime for Video I/O"                 OFF  IF APPLE )
 OCV_OPTION(WITH_QTKIT          "Use QTKit Video I/O backend"                 OFF  IF APPLE )
 OCV_OPTION(WITH_TBB            "Include Intel TBB support"                   OFF  IF (NOT IOS AND NOT WINRT) )
+OCV_OPTION(WITH_HPX            "Include Ste||ar Group HPX support"           OFF)
+OCV_OPTION(WITH_HPX_STARTSTOP  "Start-Stop HPX runtime for each backend call" OFF)
 OCV_OPTION(WITH_OPENMP         "Include OpenMP support"                      OFF)
 OCV_OPTION(WITH_PTHREADS_PF    "Use pthreads-based parallel_for"             ON   IF (NOT WIN32 OR MINGW) )
 OCV_OPTION(WITH_TIFF           "Include TIFF support"                        ON   IF (NOT IOS) )
@@ -1348,6 +1350,7 @@ endif()
 # Order is similar to CV_PARALLEL_FRAMEWORK in core/src/parallel.cpp
 ocv_build_features_string(parallel_status EXCLUSIVE
   IF HAVE_TBB THEN "TBB (ver ${TBB_VERSION_MAJOR}.${TBB_VERSION_MINOR} interface ${TBB_INTERFACE_VERSION})"
+  IF HAVE_HPX THEN "HPX (WITH_HPX_STARTSTOP=${WITH_HPX_STARTSTOP})"
   IF HAVE_OPENMP THEN "OpenMP"
   IF HAVE_GCD THEN "GCD"
   IF WINRT OR HAVE_CONCURRENCY THEN "Concurrency"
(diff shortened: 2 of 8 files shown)


modules/core/test/test_main.cpp
```diff
@@ -3,4 +3,8 @@
 // of this distribution and at http://opencv.org/license.html.
 #include "test_precomp.hpp"
 
+#if defined(HAVE_HPX) && !defined(HPX_STARTSTOP)
```
vpisarev: So, if OpenCV is built without HPX_STARTSTOP, this include shall be put into every single application that uses OpenCV? I believe, it's very big limitation. I would then find some other way to do that.

vpisarev:
@Jakub-Golinowski, thank you for the contribution. I wonder, if you can provide any benchmarks, e.g. comparing HPX against TBB or pthreads or OpenMP? Also, I found that w/o START_STOP option you need to put `#include "...hpx_main.h"` into every application, is this correct? If so, it's very big limitation. We shall think of some workaround. Also, we found that some OpenMP implementations have very aggressive timeout politics, for the sake of better performance in synthetic benchmarks. But in reality it means that CPU is 100% busy most of the time and the other parallel frameworks run out of CPU power at the same time. E.g. imagine that we have parallel TBB section, then parallel pthreads section, then parallel HPX section, then TBB again etc. How quickly will HPX threads fall asleep? Can the timeout be configured?

Jakub-Golinowski:
@hrnr it is really nice to hear your enthusiasm about the HPX backend! To answer your questions: HPX supports distributed computation so it would be possible to distribute algorithms among multiple worker machines, but current implementation is made analogously to the other backends and uses only local parallel_for()_. However, this is a very interesting idea which requires some work in order to figure out for example if it is necessary to back the cv::Mat data structure by HPX distributed data structures (like partitioned_vector).

As for the HPX_STARTSTOP, the static initialization is possible with HPX and can be considered as the alternative version of the HPX_STARTSTOP backend (current STARTSTOP version is just the simplest version of HPX backed that does not require any changes to user code).


@vpisarev thank you for your comments and questions. As for the benchmarks I was using the mandelbrot example from OpenCV documentation, the results for 4 cores can be found here:
https://github.com/Jakub-Golinowski/opencv_hpx_backend/blob/master/experiments/2018-06-17-22.33-all_over_workload_num_threasds_0-100_nstripes/images/Speed-up_as_function_of_image_size.png

As for the primary backend version (runtime assumed to be started by the user), it is true that user has to launch the HPX runtime explicitly and make calls to cv::parallel_for()_ loop from within the runtime - one way of achieving it is by putting `#include "...hpx_main.hpp"` into cpp file of his application that contains main() function, then all the user code is within the HPX runtime. We realize it might be a limitation to some applications but on the other hand this approach makes it easier for the use-case we have in mind. Since HPX is a library that is focused on parallelization by allowing the user to build a DAG of his work-flow instead of classic fork-join approach, the use-case we are thinking about is when user of HPX library wants to use functionality of OpenCV in his application as part of the above mentioned workflow DAG. In the above use-case using the backend in the primary version is far easier and more intuitive to the user. Whereas, any type of STARTSTOP backend in this use-case would make it really difficult to write such an application. Therefore, we propose two versions configurable by a build option.  The first version is for the user who is using HPX and adds OpenCV functionality and second (STARTSTOP) for the user who simply wants to use HPX backend for cv::parallel_for_().

As for the timeout policies (when the worker threads are put to sleep), they can also be aggressive in HPX but they are configurable so this aspect can be easily tuned. However, I am not sure if I understand the example with multiple backends, I thought that parallel backend is a compile-time option and there cannot be multiple backends in OpenCV? Or are we talking about more general case?

Jakub-Golinowski:
I deleted the branch to push the one with the rewritten history (rebase against master). Could you reopen this PR?

hrnr:
You can use force push to your branch `git push -f` to avoid this issue in the future. Never delete a branch with associated Github PR.  I believe no one can restore Github PRs with a deleted branch.

Jakub-Golinowski:
@hrnr thank you very much, I was mislead by the fact that this approach works for the gitlab.

I recreated the PR: https://github.com/opencv/opencv/pull/11897

Issue #11866, opened 2018-07-02
Enable more deep learning tests


dkurt:
##### System information (version)

- OpenCV => :grey_question:
- Operating System / Platform => :grey_question:
- Compiler => :grey_question:

##### Detailed description

The following tests are skipped using specific backend+target configurations. Classify them by categories (breaks with an assertion :negative_squared_cross_mark:, accuracy loss :beetle:), find a way to fix (as a bug solution or a feature request :star: ). Some of tests could be fixed due Inference Engine graphs building fix (it's assigned to @dkurt).

* DNN_BACKEND_OPENCV, DNN_TARGET_OPENCL_FP16

- [ ] Test_TensorFlow_layers.lstm :beetle: 
- [ ] Test_TensorFlow_layers.matmul (on [buildbot](http://pullrequest.opencv.org/buildbot/builders/precommit_opencl_linux)) :beetle: 
- [ ] Test_Caffe_layers.InnerProduct :beetle:
- [ ] Test_Caffe_layers.layer_prelu_fc :beetle: (unstable)
- [ ] Test_Caffe_layers.FasterRCNN_Proposal :beetle: 
- [x] Test_Caffe_layers.PriorBox_squares :beetle: 
- [x] Layer_Test_Halide/Convolution.Accuracy,([6, 4, 1], 5x6, 3x1, 1x1, 0x1, 1x1, false) :beetle: (unstable)
- [x] Layer_Test_Halide/Convolution.Accuracy,([6, 4, 1], 5x6, 3x1, 1x1, 0x1, 2x2, true) :beetle: (unstable)
- [x] Layer_Test_Halide/FullyConnected.Accuracy :negative_squared_cross_mark:
 
* DNN_BACKEND_OPENCV, DNN_TARGET_OPENCL

- [x] Test_Caffe_layers.PriorBox_squares :beetle: 

* DNN_BACKEND_INFERENCE_ENGINE, DNN_TARGET_CPU

- [ ] Test_Darknet_layers.shortcut :negative_squared_cross_mark:
- [x] Test_TensorFlow_layers.mvn_batch_norm :negative_squared_cross_mark:
- [ ] Test_TensorFlow_layers.reshape :negative_squared_cross_mark:
- [ ] Test_TensorFlow_layers.lstm :negative_squared_cross_mark:
- [ ] Test_TensorFlow_layers.split :negative_squared_cross_mark:
- [ ] Test_TensorFlow_layers.resize_nearest_neighbor :beetle: 
[...]

PR #11867, opened 2018-07-02
Enable per-layer deep learning tests for different backends

dkurt:
### This pullrequest changes

Enable more parametric tests varying deep learning backends and devices targets

see https://github.com/opencv/opencv/issues/11866 for skipped tests which should be fixed

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/484


vpisarev:
looks great! :+1:

PR #11880, opened 2018-07-03
videoio(MSMF): allow -1 index to open the first available camera

alalek:
-1 is used to search for first available/free camera.

Such behavior already exists in V4L backend.


vpisarev:
thanks! could you please update `VideoCapture::open()` (and the corresponding constructor) description and mention that some backends, such as V4L, MSMF, support -1 as the "magical" value

alalek:
Patch validation with 2 cameras are failed: required proper detection of "busy" cameras.
MSMF backend opens "used" camera, but fails on frame grabbing (it is too late to select another camera).

asmorkalov:
cc @VadimLevin

PR #11881, opened 2018-07-03
samples: videocapture_camera use VideoCapture with 0 index

alalek:
Not all backends support -1 index.

Related: #11880


alalek:
:+1:

Issue #11810, opened 2018-06-22
VideoCapture does not work with Video for Windows backend


Bleach665:
opencv - 64f2b1b  + contrib modules
Windows 10 Enterprise N 2016 LTSB 10.0.14393 Build 14393
VS 2017 15.7.3
CMake - 3.11.4.

Opencv builded without MSMF (for allow app to work on Windows N/KN).
I this code
```
VideoCapture cap(0);    
if(!cap.isOpened()) return;
while(true) { 
    cap >> frame;
    imshow("1", frame);
....
}
```
in frame only grey (RGB(205,205,205)) pixels in debug and black in release. 
Camera parameters are determinated fine.

Debug output:
` [ INFO:0] VIDEOIO: Enabled backends(5, sorted by priority): FFMPEG(1000); VFW(990); DSHOW(980); CV_IMAGES(970); CV_MJPEG(960)`

Also tested with different input engines:
if device index in VideoCapture is 
CAP_FFMPEG then cap.isOpened() return false;
CAP_VFW - described above black/grey frame
CAP_DSHOW - work fine
CAP_IMAGES and CAP_OPENCV_MJPEG - cap.isOpened() return false.

alalek:
Looks like VFW backend is broken (at least for cameras).
You can force DirectShow backend via environment variable: `OPENCV_VIDEOIO_PRIORITY_LIST=DSHOW`

vovka643:
At this moment VFW backend does not supported in any openCV branches.

PR #11882, opened 2018-07-03
videoio(VFW): lower priority (after DSHOW)

alalek:
Current implementation of VFW is not able to properly work with some cameras (black / grey frames are provided).

related #11810


alalek:
:+1:

PR #11883, opened 2018-07-03
core: FileStorage - add support for writing vector<String> with bindings

paroj:


alalek:
@paroj Could you please add some [test](https://github.com/opencv/opencv/blob/3165baa1f152878b49c984384aa61c430a686768/modules/python/test/test_persistence.py) for this feature?

modules/core/include/opencv2/core/persistence.hpp
```diff
@@ -444,7 +444,9 @@ class CV_EXPORTS_W FileStorage
     /// @overload
     CV_WRAP void write(const String& name, const String& val);
     /// @overload
-    CV_WRAP void write(const String& name, InputArray val);
+    CV_WRAP void write(const String& name, const Mat& val);
```
vpisarev: does it mean that we loose ability to write `UMat` or `std::vector<float>` into the FileStorage?
Can we just add the second signature `CV_WRAP void write(const String& name, const std::vector<String>& val);` without changing the first one?

paroj: for `UMat`, yes. Python tuples are converted into a Mat, so a `std::vector<float>` does not exists in the context of wrapping.

just adding the `vector<String>` version does not work because the implicit Mat to std::vector conversion:
```
pyopencv_generated_types.h:10611:41: error: call of overloaded ‘write(cv::String&, cv::Mat&)’ is ambiguous
```

vpisarev:
:+1:

paroj:
I was just writing a follow up PR to restore UMat support, but then I noticed that even in c++, there is only
```
void write( FileStorage& fs, const String& name, const Mat& value )
```
but no UMat variant, so probably we should leave it at that for consistency.

Issue #11015, opened 2018-03-06
OE-20 "Calibration Module" discussion


vpisarev:
the feature request about evolution proposal [OE-20](https://github.com/opencv/opencv/wiki/OE-20.-Module-Calibration)

paroj:
some more points:
* the different calibrators all have their own (unoptimized) Levenberg-Marquardt implementation.
* we also have several camera models (different distortion modelling)
  * ominidir
  * the calib3d one
  * calib3d::fisheye
  * calib3d::fisheye ex #6801
  * the latter 3 can probably folded into one, see e.g. http://april.eecs.umich.edu/pdfs/richardson2013iros.pdf (II)

catree:
Some potentially interesting links:
- [Accurate geometric camera calibration](https://github.com/puzzlepaint/camera_calibration)
- [Calibration Wizard: A Guidance System for Camera Calibration Based on Modelling Geometric and Corner Uncertainty](http://openaccess.thecvf.com/content_ICCV_2019/html/Peng_Calibration_Wizard_A_Guidance_System_for_Camera_Calibration_Based_on_ICCV_2019_paper.html)
- [Efficient pose selection for interactive camera calibration](http://publica.fraunhofer.de/documents/N-477259.html)
- the following issues:
  - [calibrateCamera does not enforce lens model constraints #15992](https://github.com/opencv/opencv/issues/15992)
  - [Ill-defined distortion model found by calibrateCamera #15577](https://github.com/opencv/opencv/issues/15577)

asmorkalov:
https://github.com/opencv/opencv/issues/22031

PR #11885, opened 2018-07-03
RFC: merge fisheye camera into default camera calibration API

paroj:
for #11015

- allow using default calib flags for fisheye functions
- forwarded the cameraModel flag where possible
- solvePnP will still fail with fisheye model in undistort points
  - as undistortPoints is in imgproc, while the fisheye version is in calib3d there is no forwarding
- the ccalib versions could be similarly integrated

### DIFF
--- a/modules/calib3d/include/opencv2/calib3d.hpp
+++ b/modules/calib3d/include/opencv2/calib3d.hpp
@@ -277,7 +277,17 @@ enum { CALIB_USE_INTRINSIC_GUESS = 0x00001,
        CALIB_ZERO_DISPARITY      = 0x00400,
        CALIB_USE_LU              = (1 << 17), //!< use LU instead of SVD decomposition for solving. much faster but potentially less precise
        CALIB_USE_EXTRINSIC_GUESS = (1 << 22), //!< for stereoCalibrate
+       // only for fisheye camera model
+       CALIB_FIX_SKEW              = 1 << 23,
+       CALIB_RECOMPUTE_EXTRINSIC   = 1 << 24,
+       CALIB_CHECK_COND            = 1 << 25
      };
+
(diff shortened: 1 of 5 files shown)


D-Alex:
A more object oriented solution would be to generate a class for different camera models and use the class to proxy the calls to the right underlying base functions. This could be also used for adding more complex camera models like refractive ones. In addition it would hide the complexity for manipulating cameras (move / rotate / map to scene / normalise / canonical / ...).

Also a class like this could be used to replace https://github.com/opencv/opencv/blob/b39cd06249213220e802bb64260727711d9fc98c/modules/viz/include/opencv2/viz/types.hpp#L158

paroj:
>  use the classes to proxy the calls to the right underlying base functions

this is probably the cleaner solution as it would alleviate passing on a consistent flags combination.

However I think (3D) move/ rotate should be kept outside to keep Camera focused on projecting things.

In the end it boils down on backwards compatible we want to be. Any opinions on that?

D-Alex:
Changing the pose of a pinhole camera is essentially just manipulating the underlying projection matrix with some helper methods and improves the usability for people not familiar with the 4x3 projection matrix.

I believe in the long run there will be more cameras in opencv than the currently implement ones and adding more flags for this does not really scale. In an ideal world application developers would just use a generic camera interface which gets proxied to the right opencv functions (opencv::fisheye, opencv::pinhole, ....). This would maintain the strength of opencv having base functions with near zero side effects while also giving a common interface for more demanding applications such as Sfm.

    camera = new FishEye(...)
    camera->calibrate
    camera->undistortPoints
    camera->estimatePose3D2D --> solvePnP
    camera->estimatePose2D2D --> fivePoint algorithm

PR #11899, opened 2018-07-05
Add missing constructors and operator=

### LINKED ISSUE #11898: Add missing constructors and assignment operators in type.hpp
Some types declared in types.hpp (Complex, Point2, Point3, Size, Rect, Scalar) miss some constructors and assignment operators.

The move constructor and assignment operator can especially benefit from being declared noexcept.
std::vector of these types can use move instead of copy during resize.
This may not be a big deal for these small types, but is important for OpenCV clients that create structures/classes containing Point2, Rect...

arnaudbrejeon:
resolves #11898

Add missing constructors and assignment operators to Complex, Point2, Point3, Size, Rect, Scalar.

```
force_builders=Custom
docker_image:Custom=fedora:28
```

### DIFF
--- a/modules/core/include/opencv2/core/types.hpp
+++ b/modules/core/include/opencv2/core/types.hpp
@@ -2145,11 +2145,7 @@ Scalar_<_Tp>::Scalar_(_Tp v0, _Tp v1, _Tp v2, _Tp v3)
 }
 
 template<typename _Tp> inline
-Scalar_<_Tp>::Scalar_(const Scalar_<_Tp>& s) {
-    this->val[0] = s.val[0];
-    this->val[1] = s.val[1];
-    this->val[2] = s.val[2];
-    this->val[3] = s.val[3];
+Scalar_<_Tp>::Scalar_(const Scalar_<_Tp>& s) : Vec<_Tp, 4>(s) {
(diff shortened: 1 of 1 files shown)


vpisarev:
@alalek, could you please review it?

modules/core/include/opencv2/core/types.hpp
```diff
@@ -163,10 +163,12 @@ template<typename _Tp> class Point_
     Point_();
     Point_(_Tp _x, _Tp _y);
     Point_(const Point_& pt);
+    Point_(Point_&& pt) noexcept;
     Point_(const Size_<_Tp>& sz);
     Point_(const Vec<_Tp, 2>& v);
 
     Point_& operator = (const Point_& pt);
+    Point_& operator = (Point_&& pt) noexcept;
```
alalek: Do we need overload with `const Point_&&`? Are there any benefits for std containers from that?
Should we replace this one with const?
Same questions are for constructor.

arnaudbrejeon: Although const rvalue reference exist, I've yet to see them in practical examples. I don't think containers take advantage of them.

PR #11901, opened 2018-07-06
cuda: fix build

alalek:
use cv::AutoBuffer::data() to get data pointer

related #11719 

```
```


alalek:
:+1:

PR #11920, opened 2018-07-09
Make TextFormat::Parse calls consistent.

egorpugin:
Google pb does not have (or removed in 3.x series) Parser::Parser(bool) constructor, so the current code fails to compile.
This PR replaces it with static Parse call like in line 1136 of that file.


alalek:
Thank you!

So, please:

Also please use single commit only for patch (no need to keep "fixup" commits in history, so please squash them).

egorpugin:
> OpenCV uses patched version of protobuf 

Is it stated somewhere?
I'm experimenting with custom builds (some C++ package manager). General assumption was that bundled 3rdparty deps exist only to help users to simplify build process, but this is not true for opencv now.
Explicit notice would be nice.

alalek:
@egorpugin Please take a look on #11923 . Most part of tested models still works well, but models with custom layers are not supported.

egorpugin:
Great! Thank you.

PR #11921, opened 2018-07-09
Update detect QRCode algorithm

allnes:
Along with this algorithm there are related data in opencv_extra:
PR link: [https://github.com/opencv/opencv_extra/pull/488](url)


modules/objdetect/test/test_qrcode.cpp
```diff
@@ -43,32 +43,32 @@
...
 {
     String root = cvtest::TS::ptr()->get_data_path() + "qrcode/";
-    // String cascades[] =
-    // {
-        // root + "haarcascade_frontalface_alt.xml",
-        // root + "lbpcascade_frontalface.xml",
-        // String()
-    // };
+    String dataset_config = cvtest::TS::ptr()->get_data_path() + "qrcode/dataset_config.json";
+    FileStorage file_config(dataset_config, FileStorage::READ);
```
alalek: Do we really need this file?
- don't create monster test cases. Use test parameters instead.
- one image - one test. It is too hard to debug what is wrong with 10th picture, because you should process first 9 before that.
- multiple parsing of summary file should be avoided. Configuration per file looks much better (`20110817_030.jpg` + `20110817_030.json`)
- Finally, image name + 4 points can be added into tests code directly.

I hope you collected these validation points in more accurate debug build with all disabled optimizations (like IPP).

allnes: @alalek @vpisarev  done.

allnes:
@alalek  @vpisarev please review me

modules/objdetect/test/test_qrcode.cpp
```diff
@@ -1,74 +1,98 @@
...
+            String test_image = images_list[index]["image_name"];
+            if (test_image == String(GetParam())) { break; }
+        }
+
+        for (int i = 0; i < 4; i++)
+        {
+            float x = images_list[index]["x"][i];
+            float y = images_list[index]["y"][i];
+            EXPECT_LT(abs(x - transform[i].x), 3);
+            EXPECT_LT(abs(y - transform[i].y), 3);
```
alalek: For floats use `fabs`.

It is better to use `EXPECT_NEAR(v1, v2, abs_error)` for that.

modules/objdetect/test/test_qrcode.cpp
```diff
@@ -1,74 +1,98 @@
...
+        EXPECT_TRUE(detectQRCode(src, transform));
+
+        FileNode images_list = file_config["test_images"];
+        int index = 0, images_count = static_cast<int>(images_list.size());
+        ASSERT_GT(images_count, 0);
+
+        for (; index < images_count; index++)
+        {
+            String test_image = images_list[index]["image_name"];
+            if (test_image == String(GetParam())) { break; }
```
alalek: "else" case is not properly handled.

modules/objdetect/test/test_qrcode.cpp
```diff
@@ -1,74 +1,98 @@
...
+            float y = images_list[index]["y"][i];
+            EXPECT_LT(abs(x - transform[i].x), 3);
+            EXPECT_LT(abs(y - transform[i].y), 3);
+        }
+        file_config.release();
+    }
+    else
+    {
+        std::cout << "SKIP_TEST: " << String(GetParam());
+        std::cout << " (not found dataset_config.json file)" << '\n';
```
alalek: We can run QRCode code itself anyway. At least to test for crashes like fresh bugreport.

modules/objdetect/test/test_qrcode.cpp
```diff
@@ -1,74 +1,98 @@
...
-    // }
+    String dataset_config = cvtest::TS::ptr()->get_data_path() + "qrcode/dataset_config.json";
+    FileStorage file_config(dataset_config, FileStorage::READ);
+
+    if (file_config.isOpened())
+    {
+        String image_path = root + String(GetParam());
+        std::vector<Point> transform;
+        Mat src = imread(image_path, IMREAD_GRAYSCALE);
+        EXPECT_TRUE(detectQRCode(src, transform));
```
alalek: What order of returned points? Is it well defined?

```.diff
-transform
+corners
```

modules/objdetect/test/test_qrcode.cpp
allnes: done, changed EXPECT_NEAR

modules/objdetect/test/test_qrcode.cpp
allnes: done, added "else"

modules/objdetect/test/test_qrcode.cpp
allnes: done, added check zero image (256x256)

modules/objdetect/test/test_qrcode.cpp
allnes: done, renamed "transform" and location points in a specific order

modules/objdetect/test/test_qrcode.cpp
alalek: This still doesn't work.
Just rename/remove some entry in your config and try again.

modules/objdetect/test/test_qrcode.cpp
alalek: Why we should check "zero" image multiple times?
You should check the passed image instead.

modules/objdetect/test/test_qrcode.cpp
allnes: Now the situation is handled when there is no picture in the json file list

modules/objdetect/test/test_qrcode.cpp
allnes: left just checking for QR code

vpisarev:
:+1:

PR #11923, opened 2018-07-09
dnn: allow to use external protobuf

alalek:
"custom layers" feature will not work properly in these builds.

related #11920

```
cmake -DBUILD_PROTOBUF=OFF -DPROTOBUF_UPDATE_FILES=ON ...
```


alalek:
:+1:

PR #11924, opened 2018-07-09
dnn(ocl4dnn): fix args for 'max_pool_forward' kernel

alalek:
Fix messages like this:
> C++ exception with description "OpenCV(3.4.2-dev) modules/core/src/ocl.cpp:2995: error: (-220:Unknown error code -220) OpenCL error CL_INVALID_ARG_INDEX (-49) during call: clSetKernelArg('max_pool_forward_float', arg_index=8, cl_mem=0x524d2d0) in function 'set'


alalek:
:+1:

Issue #11857, opened 2018-06-29
DNN(OpenCL): "test verification failed" for FP16 kernels


alalek:
Linux perf_dnn tests with enabled OpenCL (Intel OpenCL NEO):
```
[ RUN      ] DNNTestNetwork.SqueezeNet_v1_1/3, where GetParam() = (DNN_BACKEND_OPENCV, DNN_TARGET_OPENCL_FP16)
[ERROR:0] test verification failed @ image 0 group 0 out_ch 0 h 0 w 3 got 0.369629 expected 0.594727
[ERROR:0] test verification failed @ image 0 group 0 out_ch 0 h 0 w 3 got 0.369629 expected 0.594727
[ERROR:0] test verification failed @ image 0 group 0 out_ch 0 h 0 w 3 got 0.369629 expected 0.594727
[       OK ] DNNTestNetwork.SqueezeNet_v1_1/3 (278 ms)
...
[ RUN      ] DNNTestNetwork.SSD_VGG16/3, where GetParam() = (DNN_BACKEND_OPENCV, DNN_TARGET_OPENCL_FP16)
[ERROR:0] test verification failed @ image 0 group 0 out_ch 2 h 1 w 0 got 1.81836 expected 1.32812
[ERROR:0] test verification failed @ image 0 group 0 out_ch 2 h 1 w 0 got 1.81836 expected 1.32812
[ERROR:0] test verification failed @ image 0 group 0 out_ch 2 h 1 w 0 got 1.81836 expected 1.32812
[       OK ] DNNTestNetwork.SSD_VGG16/3 (4958 ms)
```

pengli:
This looks like a gemm like convolution kernel issue. I will look at it.

pengli:
PR https://github.com/opencv/opencv/pull/11927 is submitted to fix this issue

PR #11927, opened 2018-07-10
Fix gemmlike convolution input reading

pengli:
add zero padding if read position is bigger than input width

This patch fixes https://github.com/opencv/opencv/issues/11857

```
force_builders=Docs,linux,ocllinux,windows,ocl,macosx,oclmacosx,linuxNoOpt
```

### DIFF
--- a/modules/dnn/src/opencl/conv_layer_spatial.cl
+++ b/modules/dnn/src/opencl/conv_layer_spatial.cl
@@ -467,7 +467,7 @@ __kernel void Conv_Interleaved(GEMM_LIKE_KERNEL_ARGS)
         int saved_y = curr_y;
 #endif
         const __global Dtype *src0_read = src0
-          + aligned_input_size * global_z                            // batch offset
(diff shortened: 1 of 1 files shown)


alalek:
Thank you! Fix works, but there is OpenCL performance degradation 15-25% (i7-6700K).

Looks like, there is an issue with case of FP16 "half" type and KERNEL_WIDTH==3 - half3 is not properly handled. Other cases are good without fix.
I tried to recover OpenCL performance here: https://github.com/alalek/opencv/commit/pr11927
Please take a look.

Also there are other 3 similar places (search for "`(const __global Dtype_t*)src`" code on this [commit](https://github.com/opencv/opencv/blob/620292e208ffd4137eb3dbbff25e42a7bd93f7b8/modules/dnn/src/opencl/conv_layer_spatial.cl))

pengli:
@alalek , patch is updated,  it uses vload3 for vec3 input reading, it should be faster than old implementation.

PR #11935, opened 2018-07-10
Support for Matx read/write by FileStorage

### LINKED ISSUE #8148: FileStorage << cv::Matx not supported
- OpenCV => 3.1

##### Detailed description

The FileStorage class does not support writing of cv::Matx objects.

##### Steps to reproduce

    #include <stdlib.h>
    void main(int argc, char **argv)
    {
        cv::FileStorage fs("dummy.yml", cv::FileStorage::MEMORY);
        fs << "a" << cv::Matx33f::eye();
    }

does not compile because of ``error C2665: 'cv::write' : none of the 10 overloads could convert all the argument types``:

    1>E:\DEV\3rdparty.git\OpenCV\3.1.0\include\opencv2/core/persistence.hpp(1065): error C2665: 'cv::write' : none of the 10 overloads could convert all the argument types
    1>          E:\DEV\3rdparty.git\OpenCV\3.1.0\include\opencv2/core/persistence.hpp(974): could be 'void cv::write(cv::FileStorage &,const cv::String &,const cv::Range &)'
    1>          E:\DEV\3rdparty.git\OpenCV\3.1.0\include\opencv2/core/persistence.hpp(663): or       'void cv::write(cv::FileStorage &,const cv::String &,const std::vector<cv::DMatch,std::allocator<_Ty>> &)'
    1>          with
    1>          [
    1>              _Ty=cv::DMatch
    1>          ]
    1>          E:\DEV\3rdparty.git\OpenCV\3.1.0\include\opencv2/core/persistence.hpp(662): or       'void cv::write(cv::FileStorage &,const cv::String &,const std::vector<cv::KeyPoint,std::allocator<_Ty>> &)'
    1>          with
    1>          [
    1>              _Ty=cv::KeyPoint
    1>          ]
    1>          E:\DEV\3rdparty.git\OpenCV\3.1.0\include\opencv2/core/persistence.hpp(661): or       'void cv::write(cv::FileStorage &,const cv::String &,const cv::SparseMat &)'
    1>          E:\DEV\3rdparty.git\OpenCV\3.1.0\include\opencv2/core/persistence.hpp(660): or       'void cv::write(cv::FileStorage &,const cv::String &,const cv::Mat &)'
    1>          E:\DEV\3rdparty.git\OpenCV\3.1.0\include\opencv2/core/persistence.hpp(659): or       'void cv::write(cv::FileStorage &,const cv::String &,const cv::String &)'
    1>          E:\DEV\3rdparty.git\OpenCV\3.1.0\include\opencv2/core/persistence.hpp(658): or       'void cv::write(cv::FileStorage &,const cv::String &,double)'
    1>          E:\DEV\3rdparty.git\OpenCV\3.1.0\include\opencv2/core/persistence.hpp(657): or       'void cv::write(cv::FileStorage &,const cv::String &,float)'
    1>          E:\DEV\3rdparty.git\OpenCV\3.1.0\include\opencv2/core/persistence.hpp(656): or       'void cv::write(cv::FileStorage &,const cv::String &,int)'
    1>          while trying to match the argument list '(cv::FileStorage, cv::String, const cv::Matx<float,3,3>)'
    1>          OpenCVExperiments.cpp(46) : see reference to function template instantiation 'cv::FileStorage &cv::operator <<<cv::Matx<float,3,3>>(cv::FileStorage &,const _Tp &)' being compiled
    1>          with
    1>          [
    1>              _Tp=cv::Matx<float,3,3>
    1>          ]

### LINKED ISSUE #11934: Missing FileStorage read and write support for Matx
FileStorage does not directly support read or write of the Matx class.

##### Steps to reproduce:

```.cpp
Matx33d m(1, 2, 3, 4, 5, 6, 7, 8, 9);

FileStorage fsw("", FileStorage::WRITE | FileStorage::MEMORY | FileStorage::FORMAT_JSON);
fsw << "m" << m;
String content = fsw.releaseAndGetString();

FileStorage fsr(content, FileStorage::READ | FileStorage::MEMORY);
fsr["m"] >> m;
fsr.release();
```

..results in compiler errors similar to: ```...include/opencv2/core/persistence.hpp:1184:5: error: no matching function for call to 'write'``` and ```...include/opencv2/core/persistence.hpp:1240:5: error: no matching function for call to 'read'```

The current fix is to use temporary Mat instead (and appropriate use of constructors), though this is not immediately obvious from the error message(s) and is against the behaviour of almost all other "standard" data types.

A similar issue (considering only the write case) was #8148 and was closed as a "wontfix". However this may have been due in part to issues with the proposed PR #8164. I have produced a new PR which I hope resolves both the previous PR issues and the inconvenience for users such as myself who regularly make use of Matx and Filestorage. I'd appreciate a second look at this issue!

mattmyne:
resolves #11934
resolves #8148

### This pullrequest changes

Adds direct read and write support for the Matx class to FileStorage. Internally calls the read and write functions for the Mat class, so output format is identical and can be used interchangeably with Mat data types (so there will be no change to previous FileStorage output). If the read is empty, the default Matx (all zeros) is returned (similar to Vec), otherwise the data is passed to the Matx constructor. This may then throw an exception if the format or size are not correct.

### DIFF
--- a/modules/core/include/opencv2/core/persistence.hpp
+++ b/modules/core/include/opencv2/core/persistence.hpp
@@ -777,7 +777,7 @@ template<typename _Tp, int m, int n> static inline void read(const FileNode& nod
     Mat temp;
     read(node, temp); // read as a Mat class
 
-    if ( !temp.data || temp.dims > 2 || temp.rows != m || temp.cols != n || temp.channels() != 1 )
+    if (temp.empty())
         value = default_matx;
     else
         value = Matx<_Tp, m, n>(temp);
--- a/modules/core/include/opencv2/core/persistence.hpp
+++ b/modules/core/include/opencv2/core/persistence.hpp
@@ -772,6 +772,17 @@ template<typename _Tp, int cn> static inline void read(const FileNode& node, Vec
     value = temp.size() != cn ? default_value : Vec<_Tp, cn>(&temp[0]);
 }
 
+template<typename _Tp, int m, int n> static inline void read(const FileNode& node, Matx<_Tp, m, n>& value, const Matx<_Tp, m, n>& default_matx = Matx<_Tp, m, n>())
+{
+    Mat temp;
+    read(node, temp); // read as a Mat class
+
+    if ( !temp.data || temp.dims > 2 || temp.rows != m || temp.cols != n || temp.channels() != 1 )
+        value = default_matx;
+    else
+        value = Matx<_Tp, m, n>(temp);
+}
+
 template<typename _Tp> static inline void read(const FileNode& node, Scalar_<_Tp>& value, const Scalar_<_Tp>& default_value)
 {
     std::vector<_Tp> temp; FileNodeIterator it = node.begin(); it >> temp;
@@ -952,6 +963,12 @@ void write(FileStorage& fs, const Vec<_Tp, cn>& v )
         write(fs, v.val[i]);
 }
 
+template<typename _Tp, int m, int n> static inline
+void write(FileStorage& fs, const Matx<_Tp, m, n>& x )
+{
+    write(fs, Mat(x)); // write as a Mat class
+}
+
 template<typename _Tp> static inline
 void write(FileStorage& fs, const Scalar_<_Tp>& s )
 {
@@ -1015,6 +1032,12 @@ void write(FileStorage& fs, const String& name, const Vec<_Tp, cn>& v )
 {
     cv::internal::WriteStructContext ws(fs, name, FileNode::SEQ+FileNode::FLOW);
     write(fs, v);
+}
+
+template<typename _Tp, int m, int n> static inline
+void write(FileStorage& fs, const String& name, const Matx<_Tp, m, n>& x )
+{
+    write(fs, name, Mat(x)); // write as a Mat class
 }
 
 template<typename _Tp> static inline
--- a/modules/core/test/test_io.cpp
+++ b/modules/core/test/test_io.cpp
@@ -1095,33 +1095,55 @@ TEST(Core_InputOutput, filestorage_yaml_advanvced_type_heading)
 TEST(Core_InputOutput, filestorage_matx_io)
 {
     Matx33d matxTest(1.234, 2, 3, 4, 5, 6, 7, 8, 9.876);
+
+    FileStorage writer("", FileStorage::WRITE | FileStorage::MEMORY);
+    writer << "matxTest" << matxTest;
+    String content = writer.releaseAndGetString();
+
+    FileStorage reader(content, FileStorage::READ | FileStorage::MEMORY);
+    Matx33d matxTestRead;
+    reader["matxTest"] >> matxTestRead;
+    ASSERT_TRUE( cv::norm(matxTest, matxTestRead, NORM_INF) == 0 );
+
+    reader.release();
+}
+
+TEST(Core_InputOutput, filestorage_matx_io_size_mismatch)
+{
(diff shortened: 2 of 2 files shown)


alalek:
Just ignore - there are sporadic failures in this test.

modules/core/include/opencv2/core/persistence.hpp
```diff
@@ -767,6 +767,17 @@ template<typename _Tp, int cn> static inline void read(const FileNode& node, Vec
     value = temp.size() != cn ? default_value : Vec<_Tp, cn>(&temp[0]);
 }
 
+template<typename _Tp, int m, int n> static inline void read(const FileNode& node, Matx<_Tp, m, n>& value, const Matx<_Tp, m, n>& default_matx = Matx<_Tp, m, n>())
+{
+    Mat temp;
+    read(node, temp); // read as a Mat class
+
+    if ( !temp.data || temp.dims > 2 || temp.rows != m || temp.cols != n || temp.channels() != 1 )
+        value = default_matx;
```
alalek: 1. To catch errors early we should throw an error here.
Only empty Mat (missing file node) can replaced with default value.
Perhaps `Matx<_Tp, m, n>()` ctor is able to raise most part of errors, so condition should looks like:
```
if (temp.empty())
    value = default_matx;
else
    value = Matx<_Tp, m, n>(temp);
```

2. Missing type check in this condition. Not sure that we really want to cast bytes (8U) into floats implicitly.

modules/core/test/test_io.cpp
```diff
@@ -1092,6 +1092,38 @@ TEST(Core_InputOutput, filestorage_yaml_advanvced_type_heading)
...
+    Matx32d matxTestWrongSize(1, 2, 3, 4, 5, 6);
+    Mat normalMat = Mat::eye(3, 3, CV_64F);
+
+    std::vector<FileStorage::Mode> formats;
+    formats.push_back(FileStorage::Mode::FORMAT_JSON);
+    formats.push_back(FileStorage::Mode::FORMAT_XML);
+    formats.push_back(FileStorage::Mode::FORMAT_YAML);
+
+    for(size_t i = 0; i < formats.size(); i++)
+    {
```
alalek: It is better to create multiple tests instead of single monster. See Google Test manual about parameterized tests.

Or at least add `SCOPED_TRACE(cv::format("format=%d", (int)formats[i]));` into for body (to make error messages more useful).

modules/core/include/opencv2/core/persistence.hpp
mattmyne: 1. That's fine by me. I based the condition on the Vec read function (just above) which also returns default if the size does not match, but happy to have the Mat constructor throw an error in this condition if preferred.

2. I thought this would be handled sensibly by the Matx constructor, but I can't find the code that would run when a Mat is passed? Any suggestions welcome, even if just reject if Mat and Matx types are not the same (though this would seem a little harsh, i.e. float --> double).

modules/core/test/test_io.cpp
mattmyne: I can split the three inner tests (matxTest, matxTestWrongSize and normalMat) into separate tests. I can also drop the test for the three formats into just one (i.e. JSON) if it's not adding value, though other tests seem to test for all three in a single run.

mattmyne:
New commit to address issues from initial commit. Default Matx now only used if an empty read of the FileStorage request, otherwise the data is passed to the Matx constructor. This will throw an exception if the format or size are incorrect.

I've split the test function into smaller individual test functions to aid debugging, and included a test for an exception being thrown if the Mat read from the FileStorage container does not match the Matx expected size.

PR #11936, opened 2018-07-11
dnn: preserve name, type strings for ShuffleChannelLayer

berak:
### This pullrequest changes
call setParamsFrom, so name & type get properly copied from the importer params


dkurt:
:+1: Thanks! Looks like we need to replace `setParamsFrom` usage out of layers' implementations and call it for an every instance in a single place.

PR #11941, opened 2018-07-11
DNN(OCL): fix wrong usage of stalled .getMat() pointers

alalek:
Temporary object lifetime must be greater than pointer usage.


pengli:
@alalek , looks good to me 👍

PR #11951, opened 2018-07-12
cmake: allow to use external FindCUDA from modern CMake

alalek:
CMake 3.9.0+ is required.

OpenCV has own patched version of FindCUDA. It was necessary 5 years ago, but current implementation from CMake works well too.

Validated with Ubuntu 16.04 + CUDA SDK 8.0 + cmake 3.9.0 (official distributed binaries).

```
buildworker:Custom=linux-1
```


alalek:
:+1:

Issue #11022, opened 2018-03-06
OE-27 "Wide Universal Intrinsics" discussion


vpisarev:
the feature request about evolution proposal [OE-27](https://github.com/opencv/opencv/wiki/OE-27.-Wide-Universal-Intrinsics)

vpisarev:
@seiko2plus, I've added link to your #10708

seiko2plus:
@vpisarev, #10708 going to mapping all universal intrinsics to avx2, I also made some changes to make it friendly with OE-27.

pemmanuelviel:
@vpisarev I plan to port in OpenCV repo my assembly SSEx and intrinsic AVX-2 implementations of some distances for FLANN.
Are the wide universal instrinsics fully functional, in particular the ones for CV_SIMD256? Is there another doc than the one for "not wide" universal intrinsics? Thanks

alalek:
@pemmanuelviel There is page about [universal intrinsics](https://docs.opencv.org/3.4/df/d91/group__core__hal__intrin.html#details) in OpenCV Documentation.

pemmanuelviel:
@alalek Thank you for the link. This is the doc for the "not-wide" universal intrinsics I was mentioning.
But as I didn't see details on the 256 avx registers I would like to know if there is any doc on "wide" universal intrinsics, as well as if they are fully functional.
I would prefer porting directly on 256 bits wide universal intrinsics equivalent of AVX-2, than the equivalents of SSEx intrinsics. Going from SSEx intrinsics to AVX-x intrinsics on Intel architectures is not only a mater of registers size. Actually as both SSE-x and  AVX-x intrinsics are implemented in the way 
   C = A Op B, 
the assembly instructions for SSE-x mostly work only with two registers and take the form 
   A = A Op B
Having a single SSE-x intrinsic instruction mapping a sequence of several assembly instructions might be the reason why the performance difference between SIMD assembly and intrinsic code is quite noticeable with SSE-x.

terfendail:
Wide universal intrinsics are implemented for AVX2 and AVX512 architectures and are already used in core and impgproc modules.

Unfortunately there is no special documentation for wide universal intrinsics. However they were implemented in accordance with [OE-27](https://github.com/opencv/opencv/wiki/OE-27.-Wide-Universal-Intrinsics)

Actually there are just a few changes to universal intrinsics idea:
- vector types don't contain length anymore (e.g. v_uint8 instead of v_uint8x16)
[...]

PR #11953, opened 2018-07-12
Wide univ intrinsics

vpisarev:
this is reworked and updated #10708 by @seiko2plus:

#### Tasks
- [x] Prepare universal intrinsics to expand (#11022)
- [x] Add universal intrinsics for AVX2 (not very complete, but mostly usable already)
- [x] Expand hal_intrin tests to support SIMD 256
- [x] Convert math functions (sqrt, invsqrt, magnitude, fastatan, exp, log) to univ. intrin.
- [ ]  Replace raw intrinsics with universal SIMD for arithmetic operations
- [ ]  Update docs

```
force_builders=Custom
```


seiko2plus:
@vpisarev, Please could you combine the last changes that I made on "intrin_avx.hpp"?

modules/core/src/mathfuncs_core.simd.hpp
```diff
@@ -185,13 +110,60 @@ static inline void atanImpl(const T *Y, const T *X, T *angle, int len, bool angl
...
+    for( ; i < len; i += VECSZ*2 )
+    {
+        if( i + VECSZ*2 > len )
+        {
+            // if it's inplace operation, we cannot repeatedly process
+            // the tail for the second time, so we have to use the
+            // scalar code
+            if( i == 0 || angle == X || angle == Y )
+                break;
+            i = len - VECSZ*2;
```
alalek: if it's inplace operation and something is wrong then X/Y are already corrupted.

IMHO, vectorized loops in this file are malformed.
Proper design pattern is:
```
for( ; i <= len - BLOCK_SIZE; i += BLOCK_SIZE)
{
    // no more 'if's
    ...
}
for( ; i < len; i++ )
{
    ...
}
```

modules/core/src/mathfuncs_core.simd.hpp
```diff
@@ -185,13 +110,60 @@ static inline void atanImpl(const T *Y, const T *X, T *angle, int len, bool angl
...
+    float ybuf[BLKSZ], xbuf[BLKSZ], abuf[BLKSZ];
+    for( int i = 0; i < len; i += BLKSZ )
+    {
+        int j, blksz = std::min(BLKSZ, len - i);
+        for( j = 0; j < blksz; j++ )
+        {
+            ybuf[j] = (float)Y[i + j];
+            xbuf[j] = (float)X[i + j];
+        }
+        fastAtan32f(ybuf, xbuf, abuf, blksz, angleInDegrees);
```
alalek: `fastAtan32f()` is function with `CV_INSTRUMENT_REGION()` (required for OpenCV entry-points, but using this inside of "per-element" loops should be avoided).

Performance may be terrible, than computation in doubles.
Do we have performance tests for this?

modules/core/include/opencv2/core/hal/intrin.hpp
```diff
@@ -325,117 +145,196 @@ CV_CPU_OPTIMIZATION_HAL_NAMESPACE_END
 
 #else
 
+#define CV_SIMD128_CPP 1
 #include "opencv2/core/hal/intrin_cpp.hpp"
 
 #endif
 
+#if CV_AVX2
```
alalek: IMHO, this should support `CV_AVX` too.

modules/core/include/opencv2/core/hal/intrin.hpp
```diff
@@ -354,88 +167,175 @@ namespace cv {
...
-    static v_uint64x2 zero() { return v_setzero_u64(); }
-    static v_uint64x2 all(uint64 val) { return v_setall_u64(val); }
-};
+//==================================================================================================
 
-template <> struct V_RegTrait128<int64> {
-    typedef v_int64x2 reg;
-    static v_int64x2 zero() { return v_setzero_s64(); }
-    static v_int64x2 all(int64 val) { return v_setall_s64(val); }
+#define CV_INTRIN_DEFINE_WIDE_INTRIN(typ, vtyp, short_typ, prefix, loadsfx) \
```
alalek: It would be nice, for "internal" macros using `CV__` prefix (with double underscores).
- ASCII code of '_' is greater than A-Z, so IDE with auto-completion shows normal macros at first.
- These macros are filtered out from Doxygen [automatically](https://github.com/opencv/opencv/blob/3.4.2/doc/Doxyfile.in#L110)

modules/core/include/opencv2/core/hal/intrin_avx.hpp
```diff
@@ -0,0 +1,2016 @@
...
+
+template<int imm>
+inline v_uint8x32 v_rotate_right(const v_uint8x32& a)
+{
+    v_uint8x32 res;
+    // ESAC control[3] ? [127:0] = 0
+    __m256i swapz = _mm256_permute2x128_si256(a.val, a.val, _MM_SHUFFLE(2, 0, 0, 1));
+
+    if (imm == 0)
+        return a;
```
alalek: What is about definition of template specialization instead of `if`? At least we don't need "swapz" in this case.

modules/core/include/opencv2/core/hal/intrin_avx.hpp
```diff
@@ -0,0 +1,1679 @@
...
+    __m256 val;
+
+    explicit v_float32x8(__m256 v) : val(v) {}
+    v_float32x8(float v0, float v1, float v2, float v3,
+                float v4, float v5, float v6, float v7)
+    {
+        val = _mm256_setr_ps(v0, v1, v2, v3, v4, v5, v6, v7);
+    }
+    v_float32x8() : val(_mm256_setzero_ps()) {}
+    float get0() const { return _mm_cvtss_f32(_mm256_castps256_ps128(val)); }
```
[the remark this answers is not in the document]
vpisarev: `_mm256_cvtss_f32` is not supported by some compilers

modules/core/include/opencv2/core/hal/intrin.hpp
```diff
@@ -325,117 +145,196 @@ CV_CPU_OPTIMIZATION_HAL_NAMESPACE_END
...
-    typedef v_int16x8 int_reg;
-    typedef v_uint16x8 u_reg;
-    static v_uint16x8 zero() { return v_setzero_u16(); }
-    static v_uint16x8 all(ushort val) { return v_setall_u16(val); }
-};
+#define CV_INTRIN_DEFINE_WIDE_INTRIN(typ, vtyp, short_typ, prefix, loadsfx) \
+    inline vtyp vx_setall_##short_typ(typ v) { return prefix##_setall_##short_typ(v); } \
+    inline vtyp vx_setzero_##short_typ() { return prefix##_setzero_##short_typ(); } \
+    inline vtyp vx_##loadsfx(const typ* ptr) { return prefix##_##loadsfx(ptr); } \
+    inline vtyp vx_##loadsfx##_aligned(const typ* ptr) { return prefix##_##loadsfx##_aligned(ptr); } \
```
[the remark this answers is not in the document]
vpisarev: nope, sometimes it's load_f16

modules/core/include/opencv2/core/hal/intrin_avx.hpp
```diff
@@ -0,0 +1,1679 @@
...
+{
+    typedef double lane_type;
+    enum { nlanes = 4 };
+    __m256d val;
+
+    explicit v_float64x4(__m256d v) : val(v) {}
+    v_float64x4(double v0, double v1, double v2, double v3)
+    { val = _mm256_setr_pd(v0, v1, v2, v3); }
+    v_float64x4() : val(_mm256_setzero_pd()) {}
+    double get0() const { return _mm_cvtsd_f64(_mm256_castpd256_pd128(val)); }
```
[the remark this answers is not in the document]
vpisarev: same

modules/core/include/opencv2/core/hal/intrin.hpp
```diff
@@ -60,255 +60,70 @@
...
-#define CV_CPU_OPTIMIZATION_HAL_NAMESPACE hal_baseline
-#define CV_CPU_OPTIMIZATION_HAL_NAMESPACE_BEGIN namespace hal_baseline {
-#define CV_CPU_OPTIMIZATION_HAL_NAMESPACE_END }
-#endif
-
-
-CV_CPU_OPTIMIZATION_HAL_NAMESPACE_BEGIN
-CV_CPU_OPTIMIZATION_HAL_NAMESPACE_END
-using namespace CV_CPU_OPTIMIZATION_HAL_NAMESPACE;
-CV_CPU_OPTIMIZATION_HAL_NAMESPACE_BEGIN
```
alalek: Please restore this code block on this place (exact after "namespace cv {").

All HAL intrinsic types must be guarded, including "V_TypeTraits".

vpisarev:
@seiko2plus, thank you! I've merged your changes

PR #11963, opened 2018-07-13
Fix fully-connected layer in case of number of rows less than 4

dkurt:
### This pullrequest changes

Fixes OpenCL test cases (see https://github.com/opencv/opencv/issues/11866)

```
force_builders=Custom
buildworker:Custom=linux-2
test_opencl=ON
```

### DIFF
--- a/modules/dnn/src/ocl4dnn/src/math_functions.cpp
+++ b/modules/dnn/src/ocl4dnn/src/math_functions.cpp
@@ -612,7 +612,7 @@ bool ocl4dnnGEMV<float>(const CBLAS_TRANSPOSE TransA,
             ret = k.run(1, globalsize, localsize, false);
         }
 
-        if ((row_size % 4) != 0 && ret)
+        if (row_size < 4 || ((row_size % 4) != 0 && ret))
         {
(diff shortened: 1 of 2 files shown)


modules/dnn/src/layers/fully_connected_layer.cpp
```diff
@@ -299,11 +298,14 @@ class FullyConnectedLayerImpl CV_FINAL : public InnerProductLayer
...
             {
-                half_blobs.resize(umat_blobs.size());
                 for (int i = 0; i < umat_blobs.size(); i++)
                 {
                     if (!umat_blobs[i].empty())
-                        convertFp16(umat_blobs[i], half_blobs[i]);
+                    {
+                        UMat halfs;
+                        convertFp16(umat_blobs[i], halfs);
+                        umat_blobs[i] = halfs;
```
alalek: Perhaps we should preserve original blobs due possible network reconfiguration.

dkurt: @alalek, you're right. Fixed that.

PR #11964, opened 2018-07-13
AVI container fixes

### LINKED ISSUE #11841: cv2.VideoWriter fails with pos < (1u<<31) in function patchInt assertion

- OpenCV => 3.4.1.15
- Operating System / Platform => docker/lxc on ubuntu 16:04
- Compiler => did not used. opencv from pip

##### Detailed description
When I tried to build an avi file from number of frames opencv fails with assertions pos < (1u<<31) in function patchInt at /io/opencv/modules/videoio/src/container_avi.cpp:737.
It only happens if I try to do it inside container.
On host system there is no problem.

##### Steps to reproduce
1. create docker/lxc container
2. try to build an AVI video from frames which resulted size will be greater that 2^31 bytes 

code sample:
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-

import os,sys
import cv2

flp = "/tmp/files.txt"
vp = "/mnt/sde4/storagelayer/raw1598.avi"

fbp = "/mnt/sde4/storagelayer"

w = 5760
h = 3240
shape = (w, h)
fps = 30
fourcc = cv2.VideoWriter_fourcc(*'MJPG')

files = []

with open(flp,"r") as fl:
  files = [f.strip() for f in fl.readlines()]

vw = cv2.VideoWriter(vp, fourcc, fps, shape)

counter = 0
for fn in files:
  print("[{}] {}".format(counter, fn))
  fp = "{}{}{}".format(fbp, os.path.sep, fn)
  frame = cv2.imread(fp)
  print("  shape: {}".format(frame.shape))
  print("    sz: {}".format(os.path.getsize(fp)))
  sb = os.path.getsize(vp)
  print("  size before: {}".format(sb))
  print("    lg: {}, diff: {}".format(sb > (1 << 31), (1 << 31) - sb))
  if os.path.isfile(fp):
    vw.write(cv2.imread(fp))
  print("  size after: {}".format(os.path.getsize(vp)))
  counter += 1
vw.release()
```

logs example:
```
[1261] edafea25-0934-4f5f-b27c-2654ca531069.png
  shape: (3240, 5760, 3)
    sz: 11834116
  size before: 2145944300
    lg: False, diff: 1539348
  size after: 2147648320
[1262] 999663b3-9764-48ec-8685-65105070979d.png
  shape: (3240, 5760, 3)
    sz: 11841998
  size before: 2147648320
    lg: True, diff: -164672
OpenCV(3.4.1) Error: Assertion failed (pos < (1u<<31)) in patchInt, file /io/opencv/modules/videoio/src/container_avi.cpp, line 737
  File "vw.py", line 36, in <module>
    vw.write(cv2.imread(fp))
cv2.error: OpenCV(3.4.1) /io/opencv/modules/videoio/src/container_avi.cpp:737: error: (-215) pos < (1u<<31) in function patchInt

```

I have no clues what to check. Docker/lxc containers have no limits set.
Any suggestions, please?

mshabunin:
Resolves #11841

### This pullrequest changes

* use C++ streams for file I/O
* added check for integer overflows in some places

Tested by writing and reading a file near 4 Gb limit.

### DIFF
--- a/modules/videoio/src/container_avi.cpp
+++ b/modules/videoio/src/container_avi.cpp
@@ -3,9 +3,28 @@
 // of this distribution and at http://opencv.org/license.html.
 
 #include "opencv2/videoio/container_avi.private.hpp"
+#include <fstream>
+#include <limits>
+#include <typeinfo>
 
 namespace cv
 {
+
+// Utility function for safe integer conversions
+template <typename D, typename S>
+inline D safe_int_cast(S val)
+{
+    typedef std::numeric_limits<S> st;
+    typedef std::numeric_limits<D> dt;
+    CV_StaticAssert(st::is_integer && dt::is_integer, "Integer type is expected");
+    const bool in_range_r = (double)val <= (double)dt::max();
+    const bool in_range_l = (double)val >= (double)dt::min();
+    if (!in_range_r || !in_range_l)
+    {
+        CV_Error_(cv::Error::StsOutOfRange, ("Can not convert integer values (%s -> %s), value 0x%llx is out of range", typeid(S).name(), typeid(D).name(), val));
+    }
+    return static_cast<D>(val);
+}
 
 const uint32_t RIFF_CC = CV_FOURCC('R','I','F','F');
 const uint32_t LIST_CC = CV_FOURCC('L','I','S','T');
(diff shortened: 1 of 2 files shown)


modules/videoio/src/container_avi.cpp
```diff
@@ -3,10 +3,29 @@
...
+// Utility function for safe integer conversions
+template <typename D, typename S>
+inline D safe_int_cast(S val)
+{
+    typedef std::numeric_limits<S> st;
+    typedef std::numeric_limits<D> dt;
+    CV_StaticAssert(st::is_integer && dt::is_integer, "Integer type is expected");
+    const bool in_range_r = (double)val <= (double)dt::max();
+    const bool in_range_l = (double)val >= (double)dt::min();
+    if (!in_range_r || !in_range_l)
```
vpisarev: since we know the exact context, can we have more specific error message, e.g. "Too long AVI video; it can hold no more than 4Gb of data. Consider another type of container"

alalek: BTW, It would be nice to have similar function for other part of code too, at least `imgcodecs`.

PR #11967, opened 2018-07-13
Add Java and Python code for ML tutorials

catree:
### This pullrequest changes

I have added an overloaded `cv::PCACompute()` to be able to access to the eigenvalues. In C++, this can be access directly by member variables:
```
PCA pca_analysis(data_pts, Mat(), PCA::DATA_AS_ROW);
//pca_analysis.eigenvectors
//pca_analysis.eigenvalues
```
Add Java and Python code for the following tutorials:
- Introduction to Support Vector Machines
- Support Vector Machines for Non-Linearly Separable Data
- Introduction to Principal Component Analysis (PCA)


samples/cpp/tutorial_code/ml/non_linear_svms/non_linear_svms.cpp
```diff
@@ -26,13 +21,16 @@ int main()
...
+    const int NTRAINING_SAMPLES = 100;         // Number of training samples per class
+    const float FRAC_LINEAR_SEP = 0.9f;        // Fraction of samples which compose the linear separable part
+
     // Data for visual representation
     const int WIDTH = 512, HEIGHT = 512;
     Mat I = Mat::zeros(HEIGHT, WIDTH, CV_8UC3);
 
     //--------------------- 1. Set up training data randomly ---------------------------------------
-    Mat trainData(2*NTRAINING_SAMPLES, 2, CV_32FC1);
-    Mat labels   (2*NTRAINING_SAMPLES, 1, CV_32SC1);
```
alalek: > CV_32FC1

looks better to me in this case:
- `CV_32F` is just a data depth.
- `CV_32FC1` is complete cv::Mat data type (depth + nchannels)
So they are basically not the same things.

BTW, Currently `C1` types has the same direct mappings to data depths (same value).

catree: OK, I see. I will avoid this kind of changes.

PR #11969, opened 2018-07-13
core: eliminate 'if' logic from Matx::inv()/solve()

alalek:
- 'if' logic is moved into templates.
- removed unnecessary cv::Mat objects creation.
- fixed inv() test (invA * A == eye)
- added more Matx tests to cover all defined template specializations

[Current coverage of operations.hpp](http://pullrequest.opencv.org/buildbot/export/opencv_releases/3_4_coverage-lin64-debug/20180712-213645_88/coverage_html/opencv/modules/core/include/opencv2/core/operations.hpp.gcov.html).

```
buildworker:Custom=linux-3
```


vpisarev:
:+1:

PR #11972, opened 2018-07-14
support four channels in cv::findChessboardCorners

csukuangfj:
### This pullrequest changes
The original implementation missed the 4-channel case. This pullrequest fixes it.

### DIFF
--- a/modules/calib3d/src/calibinit.cpp
+++ b/modules/calib3d/src/calibinit.cpp
@@ -511,11 +511,14 @@ bool findChessboardCorners(InputArray image_, Size pattern_size,
 
     if (img.channels() != 1)
     {
-        cvtColor(img, img, COLOR_BGR2GRAY);
-    }
-    else
-    {
-        img.clone();
+        if (img.channels() == 3)
+        {
+            cvtColor(img, img, COLOR_BGR2GRAY);
+        }
+        else
+        {
+            cvtColor(img, img, COLOR_BGRA2GRAY);
+        }
     }
 
     int prev_sqr_size = 0;


alalek:
Do you have a test for this?

`COLOR_BGR2GRAY` has a trick to use `COLOR_BGRA2GRAY in case of 4 ch input (added for these cases).

vpisarev:
thanks you! indeed, looks like this patch is unecessary; cvtColor is smart enough to process 4-channel images as well in COLOR_BGR2GRAY mode

alalek:
@csukuangfj Almost right! But need to update this statement (line 581) with `.clone()` removal:
```
equalizeHist(img, img);
```

alalek:
BTW, GitHub doesn't allow to reopen this PR (because source branch has been changed during PR is closed).
Please open a new PR.

csukuangfj:
@alalek 
do you mean to add `img = img.clone()` just before this statement?

alalek:
> img = img.clone();
> img.clone();

I miss that :+1:

`img = img.clone()` just before `equalizeHist()` should work fine.

csukuangfj:
Underflow for `unsigned`
https://github.com/opencv/opencv/blob/5e31c82b5b71efdd4e9e119e73765d4a2b50457b/modules/core/include/opencv2/core/types.hpp#L1947-L1948

https://github.com/opencv/opencv/blob/5e31c82b5b71efdd4e9e119e73765d4a2b50457b/modules/core/include/opencv2/core/types.hpp#L1951

I think it is a bug.

PR #11977, opened 2018-07-14
fix #11976.

### LINKED ISSUE #11976: No debug version for CV_CheckTypeEQ

- OpenCV => latest

##### Detailed description

See the code
https://github.com/opencv/opencv/blob/fa66c6b797416292e2c6befb60b34b6ba0420499/modules/core/include/opencv2/core/check.hpp#L137-L153

The following macros
```.cpp
CV_DbgCheckTypeEQ
CV_DbgCheckDepthEQ
CV_DbgCheckChannelsEQ
CV_DbgCheckType
CV_DbgCheckDepth
```
are missing.

csukuangfj:
fix #11976.


alalek:
According to my comment in #11976 these macros are not needed really.
To clarify this in code, we could add some comment here with suggestion of using "Release" versions for "type" checks instead.

vpisarev:
ok, let's close it

PR #11984, opened 2018-07-16
Fixed issues found by static analysis (mostly DBZ)

mshabunin:
### This pullrequest changes

Added more checks for zero before division

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/calib3d/src/calibration.cpp
+++ b/modules/calib3d/src/calibration.cpp
@@ -2336,10 +2336,13 @@ void cvStereoRectify( const CvMat* _cameraMatrix1, const CvMat* _cameraMatrix2,
     _uu[2] = 1;
     cvCrossProduct(&uu, &t, &ww);
     nt = cvNorm(&t, 0, CV_L2);
+    CV_Assert(fabs(nt) > 0);
     nw = cvNorm(&ww, 0, CV_L2);
+    CV_Assert(fabs(nw) > 0);
     cvConvertScale(&ww, &ww, 1 / nw);
     cvCrossProduct(&t, &ww, &w3);
     nw = cvNorm(&w3, 0, CV_L2);
+    CV_Assert(fabs(nw) > 0);
     cvConvertScale(&w3, &w3, 1 / nw);
     _uu[2] = 0;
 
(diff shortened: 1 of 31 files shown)


modules/calib3d/src/calibration.cpp
```diff
@@ -2336,10 +2336,13 @@ void cvStereoRectify( const CvMat* _cameraMatrix1, const CvMat* _cameraMatrix2,
     _uu[2] = 1;
     cvCrossProduct(&uu, &t, &ww);
     nt = cvNorm(&t, 0, CV_L2);
+    CV_Assert(abs(nt) > 0);
```
alalek: > double nt, nw;

`fabs()` or `std::abs()`?

modules/imgproc/src/moments.cpp
```diff
@@ -495,6 +495,13 @@ static bool ocl_moments( InputArray _src, Moments& m, bool binary)
     const int TILE_SIZE = 32;
     const int K = 10;
 
+    Size sz = _src.getSz();
+    int xtiles = (sz.width + TILE_SIZE-1)/TILE_SIZE;
+    int ytiles = (sz.height + TILE_SIZE-1)/TILE_SIZE;
```
alalek: [`divUp()`](https://github.com/opencv/opencv/blob/3.4.2/modules/core/include/opencv2/core/utility.hpp#L479-L494)

modules/imgproc/src/morph.cpp
```diff
@@ -1615,6 +1615,7 @@ static bool ocl_morphSmall( InputArray _src, OutputArray _dst, InputArray _kerne
 
     int idxArg = kernel.set(0, ocl::KernelArg::PtrReadOnly(source));
     idxArg = kernel.set(idxArg, (int)source.step);
+    CV_Assert(source.elemSize() > 0);
```
alalek: Well, may be add this assertion ("dbg" version) into `.elemSize()` methods directly?

modules/core/src/arithm.cpp
```diff
@@ -1233,7 +1233,7 @@ void cv::compare(InputArray _src1, InputArray _src2, OutputArray _dst, int op)
     CV_Assert( op == CMP_LT || op == CMP_LE || op == CMP_EQ ||
                op == CMP_NE || op == CMP_GE || op == CMP_GT );
 
-    if(_src1.empty() && _src2.empty())
+    if(_src1.empty() || _src2.empty())
```
alalek: There is no sense to compare some non-empty Mat with empty other Mat.
```
if(_src1.empty() && _src2.empty())
{
    dst.release();
    return;
}
CV_Assert(!_src1.empty() && !_src2.empty());
```

I would just raise an error if some Mat is empty (see #8300).

modules/core/src/copy.cpp
```diff
@@ -411,6 +411,8 @@ Mat& Mat::operator = (const Scalar& s)
 {
     CV_INSTRUMENT_REGION()
 
+    if (empty()) return *this;
```
alalek: IMHO, there should be an exception. See [this code](https://github.com/opencv/opencv/pull/11984/commits/c473718bc24e1b6197696cb22a2ca46f2c6a1c3b#diff-2271efc7b6d06fd8b0b6d04a723898fcL92).

modules/core/src/rand.cpp
```diff
@@ -511,6 +511,8 @@ static RandnScaleFunc randnScaleTab[] =
 void RNG::fill( InputOutputArray _mat, int disttype,
                 InputArray _param1arg, InputArray _param2arg, bool saturateRange )
 {
+    if (_mat.empty())
+        return;
```
alalek: The same.
There is no sense to allow processing of empty data.

PR #11985, opened 2018-07-16
cmake: update DL IE detection to align with OpenVINO 2018 R2

alalek:
These CMake options are enough:
- `-DWITH_INF_ENGINE=ON`
- `-DIE_PLUGINS_PATH=lib/ubuntu_16.04/intel64`
- `-DENABLE_CXX11=ON` (OpenCV 3.4 only)

```
force_builders=Custom
buildworker:Custom=linux-1
test_opencl:Custom=ON
```


dkurt:
:+1:

PR #11990, opened 2018-07-16
core: CV_NODISCARD macro with semantic of [[nodiscard]] attr

alalek:
`[[nodiscard]]` is defined in C++17.
There is fallback alias for modern GCC / Clang compilers.

Merge after #11989.

<cut/>

[Validated](http://pullrequest.opencv.org/buildbot/builders/precommit_macosx/builds/16098).

```
/build/precommit_macosx/3.4/opencv/modules/calib3d/src/calibinit.cpp:518:9: warning: ignoring return value of function declared with 'warn_unused_result' attribute [-Wunused-result]
        img.clone();
```

```
docker_image:Custom=fedora:28
```

### DIFF
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -407,6 +407,24 @@ Cv64suf;
 
 
 /****************************************************************************************\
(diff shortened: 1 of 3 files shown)


alalek:
:+1:

PR #12002, opened 2018-07-18
imgproc(getPerspectiveTransform): add solveMethod parameter

alalek:
drop compatibility workaround via runtime parameter


alalek:
:+1:

PR #12004, opened 2018-07-18
Added strict checks for empty inputs to some core functions

mshabunin:
### This pullrequest changes

Some checks has been transformed to assertions and will throw an exception when empty inputs are passed to a function.


vpisarev:
@mshabunin, should it be put to 3.4? Looks like very useful patch for both branches

PR #12025, opened 2018-07-20
TF import : Allow a different input order for Mul+Maximum (ReLU)

### LINKED ISSUE #12024: tf_importer : Silent bug with a different input order for Mul+Maximum (ReLU)

- OpenCV => 3.4

##### Detailed description

Custom leaky ReLU implementations can have a different input order for the Maximum layer than the OpenCV import code expects, causing it to silently skip the ReLU layer in the network.

tf_importer.cpp expects the resulting graph to look like this ([tf_importer.cpp:1246-1258](https://github.com/opencv/opencv/blob/8b5f061dae9908222bf373e071ac3d6c314dafc1/modules/dnn/src/tensorflow/tf_importer.cpp#L1246) ) :
```.cpp
// Try to match with a LeakyRelu:
// node {
//   name: "LeakyRelu/mul"
//   op: "Mul"
//   input: "LeakyRelu/alpha"
//   input: "input"
// }
// node {
//   name: "LeakyRelu/Maximum"
//   op: "Maximum"
//   input: "LeakyRelu/mul"
//   input: "input"
// }
```
In this (normal) case, the importer will combine these two to a single ReLU node and redirect any input from the LeakyRelu/Maximum node to the new ReLU node LeakyRelu/mul. 

If the inputs of LeakyRelu/Maximum are the other way round, the importer will still combine the two nodes to a ReLU node named LeakyRelu/mul, but all inputs from LeakyRelu/Maximum will be redirected to input instead of LeakyRelu/mul.

The problem lies in the ExcludeLayer call ([tf_importer.cpp:1263](https://github.com/opencv/opencv/blob/8b5f061dae9908222bf373e071ac3d6c314dafc1/modules/dnn/src/tensorflow/tf_importer.cpp#L1263)) with input_blob_index set to 0, while LeakyRelu/mul could also be at input index 1.

I've seen more ExcludeLayer calls with input index 0, so this might not be the only instance of this issue.

##### Steps to reproduce
Load a graph with a ReLU as mentioned above using readNetFromTensorflow. 

Works fine :
```.py
tf.maximum(0.01*input, input, name="LeakyRelu")
```
Does not work (is effectively skipped in the network) :
```.py
tf.maximum(input, 0.01*input, name="LeakyRelu")
```

Triplesalt:
resolves #12024

This pullrequest adds a simple check for the input index of the Mul node in the Maximum node to solve the issue.

### DIFF
--- a/modules/dnn/test/test_tf_importer.cpp
+++ b/modules/dnn/test/test_tf_importer.cpp
@@ -230,6 +230,13 @@ TEST_P(Test_TensorFlow_layers, flatten)
     runTensorFlowNet("unfused_flatten_unknown_batch");
 }
 
+TEST_P(Test_TensorFlow_layers, leaky_relu)
+{
+    runTensorFlowNet("leaky_relu_order1");
+    runTensorFlowNet("leaky_relu_order2");
+    runTensorFlowNet("leaky_relu_order3");
+}
+
 TEST_P(Test_TensorFlow_layers, l2_normalize)
 {
     runTensorFlowNet("l2_normalize");
--- a/modules/dnn/src/tensorflow/tf_importer.cpp
+++ b/modules/dnn/src/tensorflow/tf_importer.cpp
@@ -1260,7 +1260,13 @@ void TFImporter::populateNet(Net dstNet)
                     if (!next_layers.empty())
                     {
                         int maximumLayerIdx = next_layers[0].second;
-                        ExcludeLayer(net, maximumLayerIdx, 0, false);
+
+                        CV_Assert(net.node(maximumLayerIdx).input_size() == 2);
+
+                        // The input from the Mul layer can also be at index 1.
+                        int mulInputIdx = (net.node(maximumLayerIdx).input(0) == name) ? 0 : 1;
+
+                        ExcludeLayer(net, maximumLayerIdx, mulInputIdx, false);
                         layers_to_ignore.insert(next_layers[0].first);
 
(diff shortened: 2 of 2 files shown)


dkurt:
@Triplesalt, nice catch, thank you! May I ask to add a test for it? You need to add a few lines into https://github.com/opencv/opencv_extra/blob/3.4/testdata/dnn/tensorflow/generate_tf_models.py and https://github.com/opencv/opencv/blob/3.4/modules/dnn/test/test_tf_importer.cpp.

like this ones: https://github.com/opencv/opencv_extra/pull/481/files#diff-c1e583866a1914ce3ebd7ec9ce5b4762 and https://github.com/opencv/opencv/pull/11840/files#diff-722c6bae2bedd98a5533bb14b97d4021 (feel free to comment the rest of tests to generate your only one).

modules/dnn/test/test_tf_importer.cpp
```diff
@@ -230,6 +230,13 @@ TEST_P(Test_TensorFlow_layers, flatten)
     runTensorFlowNet("unfused_flatten_unknown_batch");
 }
 
+TEST_P(Test_TensorFlow_layers, operand_order1)
```
dkurt: I think it should be named something like `TEST_P(Test_TensorFlow_layers, leaky_relu)` comparing to the name of operation.

dkurt:
:+1: Looks good for me. Thank you!
Could you please squash all the commits into one before we merged it?

PR #12050, opened 2018-07-24
dnn(ocl): avoid memory access violation

alalek:
Check should be always before memory access attempt.

```
```

<cut>

Pre-check via "if" is very slow (total performance drop ~2-5%).

Using `clamp()` / `min()` to fix reading indexes.


alalek:
Performance regressions are validated via these perf tests: #12126.
:+1:

PR #12056, opened 2018-07-24
core:test Expand hal_intrin tests to support SIMD256

seiko2plus:
Expand hal_intrin tests to support SIMD256

````
pw_compilers=gcc-4.9, gcc-5, gcc-6
````

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin.hpp
@@ -154,7 +154,7 @@ using namespace CV_CPU_OPTIMIZATION_HAL_NAMESPACE;
 // but some of AVX2 intrinsics get v256_ prefix instead of v_, e.g. v256_load() vs v_load().
 // Correspondingly, the wide intrinsics (which are mapped to the "widest"
 // available instruction set) will get vx_ prefix
-// (and will be mapped to v256_ counterparts) (e.g. vx_load() => v245_load())
+// (and will be mapped to v256_ counterparts) (e.g. vx_load() => v256_load())
 #if CV_AVX2
 
 #include "opencv2/core/hal/intrin_avx.hpp"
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: modules/core/test/test_intrin.avx2.cpp [hunk 1]
```diff
@@ -0,0 +1,43 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+#include "test_precomp.hpp"
+#include "test_intrin.simd.hpp"
+
+namespace opencv_test { namespace hal {
+using namespace CV_CPU_OPTIMIZATION_NAMESPACE;
+
+TEST(hal_intrin256, uint8x32)
```
alalek: It is "dead" test for non-AVX2 machines - any code from this file should not be called there.

Test declaration (via `TEST()`) and dispatching should be done outside of this function. See "FP16" test:
```
    CV_CPU_CALL_FP16_(test_hal_intrin_float16, ());
    throw SkipTestException("Unsupported hardware: FP16 is not available");
```

modules/core/include/opencv2/core/hal/intrin.hpp
```diff
@@ -316,7 +318,7 @@ template<typename _Tp> struct V_RegTraits
     CV_INTRIN_DEFINE_WIDE_INTRIN_ALL_TYPES(v256)
     CV_INTRIN_DEFINE_WIDE_INTRIN(double, v_float64, f64, v256, load)
     inline void vx_cleanup() { v256_cleanup(); }
-#elif CV_SIMD128
```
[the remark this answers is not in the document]
seiko2plus: IMO, we only need to fallback into dummy types and dummy intrinsics just for template programming and also dispatching simd128 intrin tests. 
However we need this for now to avoid compiling errors for non-simd optimizations.

modules/core/test/test_intrin.avx2.cpp
seiko2plus: Done, thank you !

seiko2plus: Accidentally rebase removed the last changes, I think its fine now .. thank you once again

PR #12059, opened 2018-07-25
Increase the performance

GermanAizek:
Increase the performance by removing the assignment to the params variable in the constructor (which is very bad) and using the implementation through the default parameters

### DIFF
--- a/modules/calib3d/src/stereobm.cpp
+++ b/modules/calib3d/src/stereobm.cpp
@@ -1089,15 +1089,9 @@ protected:
 class StereoBMImpl CV_FINAL : public StereoBM
 {
 public:
-    StereoBMImpl()
-    {
-        params = StereoBMParams();
-    }
+    StereoBMImpl() : params(StereoBMParams()) {}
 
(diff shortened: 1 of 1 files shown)


alalek:
> Increase the performance

Are there any results from performance tests?

---

As an "optimization" this patch should [go](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) into 3.4 branch. We will merge changes from 3.4 into master regularly (weekly/bi-weekly).

So, please:

vpisarev:
@GermanAizek, do you understand what StereoBM is doing and how many CPU clocks does it spend to produce each pixel of the output disparity map? I believe, it's comparable or more to the change you propose. Therefore, even for a tiny 160x100 disparity map it's 0.006% difference

GermanAizek:
@vpisarev, all the same it is optimization, and if there are many objects of class StereoBM, then the difference is greater. I also found a stereo similar in another code. If there are ways to improve performance, though small, why not take advantage of the opportunity?

PR #12063, opened 2018-07-25
further improvements in split & merge; started using non-temporary store instructions

vpisarev:
1. changed `static const __m128/256` to const `__m128/256` to avoid weird instructions and calls inserted by compiler when using `static const`.
2. added universal intrinsics that wrap `MOVNTPS` and other such (non-temporary or "no cache" store) instructions. `v_store_interleave()` and `v_store()` got the respective flags/overloaded variants
3. rewrote split & merge to use the "no cache" store instructions. It resulted in dramatic performance improvement when processing big arrays

`force_builders=Custom`
`docker_image:Custom=powerpc64le`


modules/core/include/opencv2/core/hal/intrin_avx.hpp
```diff
@@ -304,6 +304,17 @@ inline v_float16x16 v256_setall_f16(short val) { return v_float16x16(_mm256_set1
     { _mm256_storeu_si256((__m256i*)ptr, a.val); }                    \
     inline void v_store_aligned(_Tp* ptr, const _Tpvec& a)            \
     { _mm256_store_si256((__m256i*)ptr, a.val); }                     \
+    inline void v_store_aligned_nocache(_Tp* ptr, const _Tpvec& a)    \
+    { _mm256_stream_si256((__m256i*)ptr, a.val); }                    \
+    inline void v_store(_Tp* ptr, const _Tpvec& a, hal::StoreMode mode) \
```
alalek: > mode

I would like to avoid using this as a regular function parameter. Overwise somebody would try to pass this parameters dynamically via several function calls making code unreadable.
I already fixed sporadic crashes caused by improper usage of aligned memory access with unaligned pointers. Don't make debugging more harder.

Separate functions with different names are preferable (added).
Or at least template parameter instead of this regular.

vpisarev: @alalek, I would say, the function parameter makes some things easier to debug, not harder. You just temporarily modify those `v_store(ptr, reg, mode)` implementations to always use unaligned stores and see whether the crashes are eliminated. Also, each disassembly will stop at `movaps/movntps/...` instruction that attempted to write at unaligned address, regardless of whether this instruction is under condition or not. I would say, such function with embedded `if` makes some code more convenient to write and helps to avoid unnecessary code bloat.

vpisarev: it does not matter; it will show that it crashed inside v_store(ptr, reg, mode), not just in v_store(ptr, reg). Then you can try to modify the intrinsic to disable aligned and aligned_cache branches and recompile the code. If the crash is gone, the root cause if found. It's no different than putting if statements one level up.

PR #12064, opened 2018-07-25
core: Remove raw intrinsics from arithmetic operations

seiko2plus:
````
force_builders=Custom,armv7,armv8,linux,Linux AVX2
pw_compilers=gcc-4.9, gcc-5, gcc-6
disable_ipp=1
````

### CODE REVIEW

**Comment by vpisarev** [comment]
@seiko2plus, big thanks for the patch! May I ask you to make it a bit less cryptic? :) Frankly speaking, the proposed implementation is almost impossible to read, therefore its long term support is very difficult.

**Comment by seiko2plus** [comment]
@vpisarev, My apologies :), I'm going to rebase this patch into multiple commits and also improve readability.

**Comment by vpisarev** [comment]
@seiko2plus, thank you! To be more specific, I would like to ask you to avoid macros as much as possible. I know, we use them sometimes, e.g. for universal intrinsic implementation (which is difficult to avoid, because we map many different instructions to overloaded variants of the same inline function), but overall in the code we should try to use templates mostly for a generic code

File: modules/core/src/arithm.cpp [hunk 1]
```diff
@@ -2130,1013 +2130,4 @@ cvMaxS( const void* srcarr1, double value, void* dstarr )
 }
 
 
-
-namespace cv { namespace hal {
-
-//=======================================
-
-#if (ARITHM_USE_IPP == 1)
-static inline void fixSteps(int width, int height, size_t elemSize, size_t& step1, size_t& step2, size_t& step)
```
alalek: This massive code migration to another file "arithm.dispatch.cpp" is not required.
"dispatch.cpp" file is not required to enable "dispatched" optimizations.

Please move this part of code back (with added changes).

tompollok:
Im not very familiar with raw intrinsics/universal intrinsics, but does the switch improve the overall performance or whats the benefits?

modules/core/src/arithm.cpp
seiko2plus: done, thanks

modules/core/include/opencv2/core/hal/intrin_cpp.hpp
```diff
@@ -754,6 +755,30 @@ inline v_float64x2 v_absdiff(const v_float64x2& a, const v_float64x2& b)
     return c;
 }
 
+/** @brief Saturating absolute difference
+
+Returns \f$ saturate(|a - b|) \f$ .
+For 8-, 16-bit signed integer source types. */
```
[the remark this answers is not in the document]
seiko2plus: done but I had to remove absdiff[s](signed int) and reuse absdiff(signed int) by recasting it to handle overflow which gives the same result as removed code for all archs.

seiko2plus:
@tompollok, Sorry for the delayed response,
> Im not very familiar with raw intrinsics/universal intrinsics, but does the switch improve the overall performance or whats the benefits?

I'm not sure yet I still have to run performance tests over several archs but definitely this patch going to increase performance for Power8/VSX since VSX intrinsics wasn't implemented before only X86/IPP/SSE/AVX, ARM/NEON.
There are also additional benefits like supporting [CPU dispatching](https://github.com/opencv/opencv/wiki/CPU-optimizations-build-options) and replace raw intrinsics [SSE/AVX, NEON] with universal intrinsics which make [SIMD](https://en.wikipedia.org/wiki/SIMD) code hopefully more robust and clean.

seiko2plus:
@vpisarev, I'm sorry for delayed response :),  I'm almost done from this patch currently I'm going to final step which is analyses compiled objects and running performance tests.

Maybe I didn't literally follow your suggestions, I'm not sure if I did the right thing but I just tried to keep it simple without duplicating the same code as much as possible and without using macros only at "higher-level dispatcher functions " just like you said.

Please could you give me your final review? I have free time and I'm ready to accept any radical changes even I have to rewrite this patch from scratch!.
Thank you!

seiko2plus:
@vpisarev, please take into consideration that #12516 is required to merge first in order to successfully build this patch, also I combined the two patches at this repo [/seiko2plus/opencv/coreUnvintrinArithm2_test](https://github.com/seiko2plus/opencv/tree/coreUnvintrinArithm2_test) for test purposes.

PR #12068, opened 2018-07-26
imgproc: rectangle() overload with InputOutpuArray

### LINKED ISSUE #11204: Ability to write a rectangle on a UMAT
Hi,

It would be good to be able to write directly to UMAT using cv::rectangle instead of passing UMAT.getmat() beforehands.

THanks

berak:
resolves #11204

change CV_IN_OUT Mat& to InputOutputArray, so it also can accept UMat, GpuMat, etc.

(breaks ABI, so 4.0)


berak:
already fixed. (stale code here)

but it means: #11204 can be closed !

Issue #11008, opened 2018-03-06
OE-13 "Adding ONNX Importer in DNN" discussion


vpisarev:
the feature request about evolution proposal [OE-13](https://github.com/opencv/opencv/wiki/OE-13.-DNN-ONNX-Importer)

dkurt:
Solved by https://github.com/opencv/opencv/pull/12071

PR #12071, opened 2018-07-26
 Added ONNX parser with AlexNet support

l-bat:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/499

related #11008

```
force_builders=Custom
buildworker:Custom=linux-2
test_opencl:Custom=ON
```


modules/dnn/misc/onnx/opencv-onnx.pb.h
```diff
@@ -0,0 +1,5849 @@
...
+#include <string>
+
+#include <google/protobuf/stubs/common.h>
+
+#if GOOGLE_PROTOBUF_VERSION < 3005000
+#error This file was generated by a newer version of protoc which is
+#error incompatible with your Protocol Buffer headers.  Please update
+#error your headers.
+#endif
+#if 3005000 < GOOGLE_PROTOBUF_MIN_PROTOC_VERSION
```
dkurt: opencv-caffe.proto and TensorFlow's `.proto` files were compiled using protobuf compiler 3.5.1. Could you please regenerate it for ONNX with the same version?


https://github.com/opencv/opencv/blob/2f0fc920dd51a3627779dcc7458abdbb09c933e6/modules/dnn/misc/tensorflow/graph.pb.h#L16

modules/dnn/include/opencv2/dnn/dict.hpp
```diff
@@ -141,6 +141,9 @@ class CV_EXPORTS Dict
     template<typename T>
     const T &set(const String &key, const T &value);
 
+    //! Replace @p value for the @p key into the dictionary.
```
dkurt: Please update a comment comparing to method

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -749,6 +749,15 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
      */
     CV_EXPORTS_W Net readNetFromModelOptimizer(const String &xml, const String &bin);
 
+    /** @brief Reads a network model stored in <a href="https://github.com/onnx/models"> model files.
```
dkurt: Please fix doxygen warnings and keep a single `@returns` entry.

modules/dnn/include/opencv2/dnn/dnn.inl.hpp
```diff
@@ -364,6 +364,14 @@ inline const T &Dict::set(const String &key, const T &value)
     return value;
 }
 
+inline void Dict::erase(const String &key)
+{
+    _Dict::iterator it = dict.find(key);
+    if (it != dict.end()){
+        dict.erase(key);
```
dkurt: I think a single `dict.erase(key);` is enough here.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,262 @@
...
+    CV_Assert(!tensor_proto.raw_data().empty());
+    char* val = const_cast<char*>(tensor_proto.raw_data().c_str());
+
+    std::vector<int> sizes;
+    for (int i = 0; i < tensor_proto.dims_size(); i++) {
+            sizes.push_back(tensor_proto.dims(i));
+    }
+    Mat blob;
+    if (datatype == onnx::TensorProto_DataType_FLOAT) {
+        blob.create(sizes, CV_32FC1);
```
dkurt: `create` methods allocates a data but the following line replaces data pointer => memory leak.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,262 @@
...
+    tensor_proto = graph_proto.initializer(i);
+    Mat mat = getMatFromTensor(tensor_proto);
+    layers_weights.insert(std::make_pair(tensor_proto.name(), mat));
+  }
+  return layers_weights;
+}
+
+LayerParams ONNXImporter::getLayerParams(const onnx::NodeProto& node_proto)
+{
+  LayerParams lp;
```
dkurt: Please align indentations.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,262 @@
...
+    onnx::NodeProto node_proto;
+
+    for(int i = 0; i < layersSize; i++) {
+        node_proto = graph_proto.node(i);
+        layerParams = getLayerParams(node_proto);
+        if (node_proto.has_name() && !node_proto.name().empty()) {
+            layerParams.name = node_proto.name();
+        } else {
+            std::ostringstream s;
+            s << node_proto.op_type() << "_" << i;
```
dkurt: You may use `cv::format` here

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,262 @@
...
+            layerParams.type = "InnerProduct";
+            for (int i = 1; i < node_proto.input_size(); i++) {
+                weight = weights.find(node_proto.input(i));
+                CV_Assert(weight != weights.end());
+                layerParams.blobs.push_back(weight->second);
+            }
+            layerParams.set("num_output", layerParams.blobs[0].size[0]);
+            int numLayerWeights = node_proto.input_size() - 1;
+            if (numLayerWeights == 2) {
+                layerParams.set("bias_term", true);
```
dkurt: I think `layerParams.set("bias_term", layerParams.blobs.size() == 2);` is shorter.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,262 @@
...
+            } else {
+                layerParams.set("bias_term", false);
+            }
+        } else if (layer_type == "Reshape") {
+            layerParams.type = "Reshape";
+            for (int i = 1; i < node_proto.input_size(); i++) {
+                weight = weights.find(node_proto.input(i));
+                CV_Assert(weight != weights.end());
+                layerParams.set("dim", DictValue::arrayInt<int*>(
+                       weight->second.ptr<int>(),  weight->second.total() ));
```
dkurt: Please check that shape mar has a type CV_32SC1.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,262 @@
...
+            int numLayerWeights = node_proto.input_size() - 1;
+            if (numLayerWeights == 2) {
+                layerParams.set("bias_term", true);
+            } else {
+                layerParams.set("bias_term", false);
+            }
+        } else if (layer_type == "Reshape") {
+            layerParams.type = "Reshape";
+            for (int i = 1; i < node_proto.input_size(); i++) {
+                weight = weights.find(node_proto.input(i));
```
dkurt: We need to name it somehow differently because it's not weights mat for this layer. And we don't need a loop here because Reshape takes only input blob and shape, doesn't it?

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,262 @@
...
+}
+
+Mat readTensorFromONNX(const String& path)
+{
+    onnx::TensorProto tensor_proto = onnx::TensorProto();
+    std::fstream input(path.c_str(), std::ios::in | std::ios::binary);
+    if (!tensor_proto.ParseFromIstream(&input)) {
+        CV_Error(Error::StsUnsupportedFormat, "Failed to parse data");
+    }
+    CV_Assert(!tensor_proto.raw_data().empty());
```
dkurt: There is the same assertion at `getMatFromTensor`.

modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -0,0 +1,119 @@
...
+    return findDataFile(rootFolder + filename, false);
+}
+
+class Test_ONNX_layers : public DNNTestLayer
+{
+public:
+    void testLayerUsingONNXModels(const String& basename)
+    {
+        String modelsFolder = "models/";
+        String onnxmodel = _tf(modelsFolder + basename + ".onnx");
```
dkurt: I think we can remove `modelsFolder` variable:
String onnxmodel = _tf("models/"+ basename + ".onnx");

modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+        String dataFolder = "data/";
+        String inpfile = _tf(dataFolder + "input_" + basename + ".npy");
+        String outfile = _tf(dataFolder + "output_" + basename + ".npy");
```
dkurt: The same as above:
```cpp
String inpfile = _tf("data/input_" + basename + ".npy");
String outfile = _tf("data/output_" + basename + ".npy");
```

modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+        Net net = readNetFromONNX(onnxmodel);
+        ASSERT_FALSE(net.empty());
+
+        net.setPreferableBackend(backend);
+        net.setPreferableTarget(target);
+
+        net.setInput(inp);
+        Mat out = net.forward();
+std::cout << out.size << '\n';
```
dkurt: I think we can remove it.

modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -0,0 +1,119 @@
...
+    Net net;
+    const String model =  _tf("models/bvlc_alexnet.onnx");
+
+    net = readNetFromONNX(model);
+
+    ASSERT_FALSE(net.empty());
+
+    int targetId = get<1>(GetParam());
+
+    net.setPreferableBackend(DNN_BACKEND_OPENCV);
```
dkurt: Please use `target` and `backend` instead `targetId`, `backendId` (see DNNTestLayer  implementation)

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

dkurt:
@l-bat, check `caffe_io.hpp` to avoid warnings:
https://github.com/opencv/opencv/blob/1060c0f439cbb8de44ff258d2ad3ab9c6b429331/modules/dnn/src/caffe/caffe_io.hpp#L94-L101

modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+TEST_P(Test_ONNX_layers, Dropout)
+{
+    testLayerUsingONNXModels("dropout");
+}
+
+TEST_P(Test_ONNX_layers, Linear)
+{
+    testLayerUsingONNXModels("linear");
```
dkurt: It looks like we need to skip this test for some of test cases (OpenCV backend, OpenCL fp16 target).

l-bat: done

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -749,6 +749,18 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
      */
     CV_EXPORTS_W Net readNetFromModelOptimizer(const String &xml, const String &bin);
 
+    /** @brief Reads a network model stored in <a href="https://github.com/onnx/models">ONNX</a> model files.
```
dkurt: Let's replace a reference to https://onnx.ai/.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -749,6 +749,18 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+     *  @param onnxFile path to the .onnx file with text description of the network architecture.
+     *  @returns Network object that ready to do forward, throw an exception in failure cases.
+     */
+    CV_EXPORTS_W Net readNetFromONNX(const String &onnxFile);
+
+    /** @brief Creates 4-dimensional blob from .pb file.
```
dkurt: I think this method can read a general tensor as well but not only 4-dimensional.

modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -0,0 +1,116 @@
...
+        normAssert(ref, out, "", default_l1,  default_lInf);
+    }
+};
+
+TEST_P(Test_ONNX_layers, MaxPooling)
+{
+    testLayerUsingONNXModels("maxpooling");
+}
+
+TEST_P(Test_ONNX_layers, Convolution)
```
dkurt: Do you mind to unite some tests by layers?
In example, 
```cpp
TEST_P(Test_ONNX_layers, Convolution)
{
    testLayerUsingONNXModels("convolution");
    testLayerUsingONNXModels("convolution2");
}
```
Can you also rename "convolution2" to something more informative like "two_convolutions"?

dkurt:
:+1: Looks good to merge!
@alalek, can we squash all the commits into one authored by @l-bat?

alalek:
Commits here have been created using multiple credentials.
Some of them were with unknown by GitHub e-mail, perhaps you (@l-bat ) want to add e-mail to GitHub account (but do not make it primary or backup) here: https://github.com/settings/emails

Do you need history in this PR?
GitHub is able to squash commits during merge and leave history in PR as-is (but squashed commit uses author/e-mail from the first commit).

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,274 @@
...
+#ifdef HAVE_PROTOBUF
+
+#include <iostream>
+#include <fstream>
+#include <sstream>
+#include <algorithm>
+#include <string>
+#include <google/protobuf/message.h>
+#include <google/protobuf/text_format.h>
+#include <google/protobuf/io/zero_copy_stream_impl.h>
```
alalek: Do we need these explicit protobuf includes? Perhaps all of them are included via "opencv-onnx.pb.h".

modules/dnn/src/onnx/opencv-onnx.proto
```diff
@@ -0,0 +1,446 @@
...
+// WARNING: This file is automatically generated!  Please edit onnx.in.proto.
+//
+
+
+// Copyright (c) Facebook Inc. and Microsoft Corporation.
+// Licensed under the MIT license.
+
+syntax = "proto2";
+
+package onnx;
```
alalek: To avoid conflicts usually we need to rename package name too (see "opencv_caffe", #10139):
- something like `opencv_onnx`

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+#include "../precomp.hpp"
```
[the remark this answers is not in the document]
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,274 @@
...
+#endif
+
+
+#endif
+
+namespace cv {
+namespace dnn {
+CV__DNN_EXPERIMENTAL_NS_BEGIN
+
+#ifdef HAVE_PROTOBUF
```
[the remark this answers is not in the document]
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,274 @@
...
+using ::google::protobuf::Message;
+using ::google::protobuf::Descriptor;
+using ::google::protobuf::FieldDescriptor;
+using ::google::protobuf::Reflection;
+
+
+namespace
+{
+  class ONNXImporter
+  {
```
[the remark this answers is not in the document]
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,274 @@
...
+    if (datatype == onnx::TensorProto_DataType_FLOAT) {
+        Mat(sizes, CV_32FC1, val).copyTo(blob);
+    } else if (datatype == onnx::TensorProto_DataType_INT64) {
+        blob.create(sizes, CV_32SC1);
+
+        int64_t* src = reinterpret_cast<int64_t*>(val);
+        int32_t* dst = reinterpret_cast<int32_t*>(blob.data);
+
+        for (int i = 0; i < blob.total(); i++) {
+            dst[i] = static_cast<int32_t>(src[i]);
```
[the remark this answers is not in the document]
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,274 @@
...
+    onnx::NodeProto node_proto;
+
+    for(int i = 0; i < layersSize; i++) {
+        node_proto = graph_proto.node(i);
+        layerParams = getLayerParams(node_proto);
+        if (node_proto.has_name() && !node_proto.name().empty()) {
+            layerParams.name = node_proto.name();
+        } else {
+            layerParams.name = format("%s_%d", node_proto.op_type().c_str(), i);
+         }
```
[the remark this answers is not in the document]
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/opencv-onnx.proto
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,275 @@
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsuggest-override"
+#endif
+#include "opencv-onnx.pb.h"
+#if defined(__GNUC__) && __GNUC__ >= 5
+#pragma GCC diagnostic pop
+#endif
+
+
+#endif
```
alalek: Is this "endif" for `HAVE_PROTOBUF`?

This should be moved to the end of file.

Ideally, with "#else" branch with "Non Implemented" stubs for public API (but it is not used yet for caffe).

modules/dnn/src/onnx/opencv-onnx.proto
alalek: Need to re-generate "pb" files too.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,348 @@
...
+    std::string framework_name;
+    if (model_proto.has_producer_name()) {
+        framework_name = model_proto.producer_name();
+    }
+
+    std::map<std::string, int> layer_id;
+    std::map<std::string, int>::iterator layerId, currentId;
+
+    int count = 0;
+    for (int j = 0; j < graph_proto.input_size(); j++)
```
dkurt: Please add some comment that you register network's inputs here.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,348 @@
...
+                layerParams.blobs.push_back( getBlob(node_proto, constBlobs, j) );
+            }
+            layerParams.set("scale_bias", true);
+        }
+        else if (layer_type == "Gemm")
+        {
+            layerParams.type = "InnerProduct";
+            std::map<std::string, Mat>::iterator constBlob;
+            for (int j = 0; j < node_proto.input_size(); j++) {
+                if (layer_id.find(node_proto.input(j)) == layer_id.end()) {
```
dkurt: Can we consider that Gemm layer always has inputs in the following order?
```
input
weights
bias (optional)
```
This way we can start a loop from 1 and remove a condition that input is trained parameters but not an output from previous layers.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,348 @@
...
+                }
+            }
+            layerParams.set("num_output", layerParams.blobs[0].size[0]);
+            layerParams.set("bias_term", node_proto.input_size() == 3);  // A * B + bias
+        }
+        else if (layer_type == "Conv")
+        {
+            layerParams.type = "Convolution";
+            for (int j = 0; j < node_proto.input_size(); j++) {
+                if (layer_id.find(node_proto.input(j)) == layer_id.end()) {
```
dkurt: the same as above

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+         int id = dstNet.addLayer(layerParams.name, layerParams.type, layerParams);
+         layer_id.insert(std::make_pair(layerParams.name, id));
+
+         for (int j = 0; j < node_proto.input_size(); j++) {
+             layerId = layer_id.find(node_proto.input(j));
+             currentId = layer_id.find(layerParams.name);
+
+             if (layerId != layer_id.end()) {
+                 dstNet.connect(layerId->second, 0, currentId->second, j);
```
dkurt: Does `currentId->second` differ from `id` from `addLayer`?

l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,449 @@
...
+            CV_Error(Error::StsNotImplemented, "Unsupported attribute type");
+    }
+    return lp;
+}
+
+Mat ONNXImporter::getBlob(const opencv_onnx::NodeProto& node_proto,
+                    const std::map<std::string, Mat>& constBlobs, int index)
+{
+    std::map<std::string, Mat>::const_iterator constBlob;
+    constBlob = constBlobs.find(node_proto.input(index) );
```
dkurt: please add an assertion that `index < node_proto.input_size()`

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+    // create map with network inputs (without const blobs)
+    std::map<std::string, int> layer_id;
+    std::map<std::string, int>::iterator layerId;
+
+    int count = 0;
+    for (int j = 0; j < graph_proto.input_size(); j++)
+    {
+        if (constBlobs.find(graph_proto.input(j).name()) == constBlobs.end()) {
+            layer_id.insert(std::make_pair(graph_proto.input(j).name(), count));
```
dkurt: We need to create a test with a network which has more than one input. In example, element-wise summation of two inputs.
Please also add a comment above what this loop does.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+
+        if (layer_type == "MaxPool")
+        {
+            layerParams.type = "Pooling";
+            layerParams.set("pool", "MAX");
+            if (framework_name == "onnx-caffe2")
+                layerParams.set("ceil_mode", true);
+            else
+                layerParams.set("ceil_mode", false);
```
dkurt: As for me the following line looks better here and below.
```cpp
layerParams.set("ceil_mode", framework_name == "onnx-caffe2");
```

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,449 @@
...
+                }
+            }
+        }
+        else if (layer_type == "ImageScaler")
+        {
+            layerParams.type = "Scale";
+            if (layerParams.has("scale")) {
+                layerParams.blobs.push_back(
+                    //size of bias == number of input channels
+                    Mat(Size(1,  layerParams.get("bias").size()), CV_32FC1, layerParams.get<float>("scale")));
```
dkurt: What if `layerParams.has("bias")` is false?

modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -0,0 +1,362 @@
...
+    checkBackend(&inp, &ref);
+
+    net.setInput(inp);
+    ASSERT_FALSE(net.empty());
+    Mat out = net.forward();
+
+    normAssert(out, ref, "", default_l1,  default_lInf);
+}
+
+TEST_P(Test_ONNX_nets, ResNet101_DUC_HDC)
```
dkurt: It looks like we can use functionality of `testLayerUsingONNXModels` to test these models. Can you rename it more generally and replace similar testing code?

modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -0,0 +1,362 @@
...
+}
+
+TEST_P(Test_ONNX_layers, MaxPooling_Sigmoid)
+{
+    testLayerUsingONNXModels("maxpooling_sigmoid");
+}
+
+TEST_P(Test_ONNX_layers, Concatenation)
+{
+    testLayerUsingONNXModels("concatenation");
```
dkurt: This test is failed with `DNN_BACKEND_INFERENCE_ENGINE` and `DNN_TARGET_OPENCL`, `DNN_TARGET_OPENCL_FP16` targets. Can you add a skip exception?

l-bat:
I updated pull request and run script download_models.py from folder dnn/. All models downloaded successfully.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,582 @@
...
+            char* val = const_cast<char*>(tensor_proto.raw_data().c_str());
+            Mat(sizes, CV_32FC1, val).copyTo(blob);
+        }
+    }
+    else if (datatype == opencv_onnx::TensorProto_DataType_DOUBLE)
+    {
+        const ::google::protobuf::RepeatedField<double>& field = tensor_proto.double_data();
+        CV_Assert(!field.empty());
+        Mat(sizes, CV_64FC1, (void*)field.data()).copyTo(blob);
+        blob.convertTo(blob, CV_32FC1);
```
dkurt: It's better to call only `convertTo` here without in-place conversions:
```cpp
Mat(sizes, CV_64FC1, (void*)field.data()).convertTo(blob, CV_32FC1);
```

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,582 @@
...
+                }
+                dst[i] = saturate_cast<int32_t>(src[i]);
+            }
+        }
+        else
+        {
+            char* val = const_cast<char*>(tensor_proto.raw_data().c_str());
+            int64_t* src = reinterpret_cast<int64_t*>(val);
+
+            for (int i = 0; i < blob.total(); i++) {
```
dkurt: It looks like we can just get a pointer to `int64` data inside an if-else condition but move similar conversion code outside it.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,582 @@
...
+                lp.set(attribute_proto.name(), DictValue::arrayInt(
+                    &dst[0], attribute_proto.ints_size()));
+        }
+        else if (attribute_proto.has_t())
+        {
+            ::opencv_onnx::TensorProto tensor = attribute_proto.t();
+            Mat blob = getMatFromTensor(tensor);
+            lp.blobs.push_back(blob);
+        }
+        else if (attribute_proto.has_t() || attribute_proto.has_g() ||
```
dkurt: For now we can manage `attribute_proto.has_t()` (see above).

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,582 @@
...
+    std::map<std::string, LayerInfo> layer_id;
+    std::map<std::string, LayerInfo>::iterator layerId;
+    // fill map: push layer name, layer id and output id
+    int count = 0;
+    std::vector<String> netInputs;
+    for (int j = 0; j < graph_proto.input_size(); j++)
+    {
+        if (constBlobs.find(graph_proto.input(j).name()) == constBlobs.end()) {
+            netInputs.push_back(graph_proto.input(j).name());
+            layer_id.insert(std::make_pair(graph_proto.input(j).name(), LayerInfo(0, count)));
```
dkurt: `count` equals to `netInputs.size()`. So we can remove it.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,582 @@
...
+
+    // create map with network inputs (without const blobs)
+    std::map<std::string, LayerInfo> layer_id;
+    std::map<std::string, LayerInfo>::iterator layerId;
+    // fill map: push layer name, layer id and output id
+    int count = 0;
+    std::vector<String> netInputs;
+    for (int j = 0; j < graph_proto.input_size(); j++)
+    {
+        if (constBlobs.find(graph_proto.input(j).name()) == constBlobs.end()) {
```
dkurt: Please put `graph_proto.input(j).name())` to some temporal variable to remove duplications:
```
const std::string& name = graph_proto.input(j).name());
```

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: We use count to know id input. Max id = netInputs.size() -1, min = 0.

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,582 @@
...
+            else {
+                layerParams.type = "Shift";
+                layerParams.blobs.push_back(blob);
+            }
+        }
+        else if (layer_type == "Constant")
+        {
+            CV_Assert(node_proto.input_size() == 0);
+            CV_Assert(layerParams.blobs.size() == 1);
+            constBlobs.insert(std::make_pair(node_proto.output(0), layerParams.blobs[0]));
```
dkurt: `node_proto.output(0)` -> `layerParams.name`

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,582 @@
...
+        else if (layer_type == "Reshape")
+        {
+            Mat blob = getBlob(node_proto, constBlobs, 1);
+            CV_Assert(blob.type() == CV_32SC1);
+            layerParams.set("dim", DictValue::arrayInt<int*>(
+                        blob.ptr<int>(), blob.total() ));
+
+            if (layer_id.find(node_proto.input(0)) == layer_id.end()) {
+                Mat input = getBlob(node_proto, constBlobs, 0);
+                std::vector<int> dims;
```
dkurt: Try to cast a mat with shapes to std::vector:
```cpp
std::vector<int> dims = blob;
```

modules/dnn/src/onnx/onnx_importer.cpp
dkurt: But both `netInputs.size()` and `count` increase together in this loop.

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: done

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -0,0 +1,573 @@
...
+    std::map<std::string, LayerInfo> layer_id;
+    std::map<std::string, LayerInfo>::iterator layerId;
+    // fill map: push layer name, layer id and output id
+    std::vector<String> netInputs;
+    for (int j = 0, count = 0; j < graph_proto.input_size(); j++)
+    {
+        const std::string& name = graph_proto.input(j).name();
+        if (constBlobs.find(name) == constBlobs.end()) {
+            netInputs.push_back(name);
+            layer_id.insert(std::make_pair(name, LayerInfo(0, count++)));
```
dkurt: @l-bat, I still a bit confused about `count` variable. Do we really need it? Am I right that we use it to register network's inputs by pins? In example, (0, 0), (0, 1) etc. But at the same time we also push corresponding name to `netInputs`. That is why I wanted to ask you try the following changes:
```cpp
        if (constBlobs.find(name) == constBlobs.end()) {
            netInputs.push_back(name);
            layer_id.insert(std::make_pair(name, LayerInfo(0, netInputs.size() - 1)));
        }
```

l-bat: Yes, I understood.

vpisarev:
@l-bat, @dkurt, I suggest to merge it into the 3.4 or master branch in order to squeeze this important functionality into OpenCV 4.0 beta and then 4.0 final. The earlier users start testing it in real life, the better shape of the code we get by the final release!

dkurt:
@vpisarev, PR is ready to be merged.

PR #12088, opened 2018-07-28
core(ocl): don't expose exceptions from OpenCL callback

alalek:
to avoid silent crashes of OpenCL worker threads.


alalek:
:+1:

PR #12090, opened 2018-07-28
Change README.md URL's to https

jammm:


alalek:
It is a bad practice to start contribution into large projects from **unnecessary changes** into README.md file.
There are many of other places / issues which need fixes.

BTW, HTTPS for answers.opencv.org is not working yet (and we have an opened issue for this, not need to create a new one).

PR #12097, opened 2018-07-30
Support WITH_CUDA with clang compiler.

### LINKED ISSUE #11354: Errors when compiling OpenCV with Clang and CUDA with GCC
- OpenCV => 3.4.2
- Compiler => Clang 4.0.0 (OpenCV) + GCC 5.4 (CUDA)

##### Detailed description

1. When compiling OpenCV with CUDA enabled using the clang compiler, it tells you that clang is not supported for CUDA. This is incorrect information, as CUDA does support Clang.
See: https://github.com/opencv/opencv/blob/master/cmake/OpenCVDetectCUDA.cmake#L6
Worse still is that is using the incorrect compiler to make this decision. Instead of using CUDA_HOST_COMPILER (which was set to gcc), it came to this conclusion by using the project compiler.

2. When setting CPP flags for CUDA, it simply reuses the ones that were used for the project. However, some flags may not be compatible with the CUDA compiler. One instance of this is -Qunused-arguments, which GCC 5.4 failed on.
See: https://github.com/opencv/opencv/blob/master/cmake/OpenCVCompilerOptions.cmake#L139

##### Steps to reproduce

Compile OpenCV with clang but with CUDA_HOST_COMPILER=gcc

##### Fixed by
Removing the 3 lines at each link above resolved the issue and built working binaries.

xsacha:
resolves #11354
### This pullrequest changes
Allows compiling CUDA OpenCV with a clang compiler (which is a valid scenario supported by CUDA).
Allows compiling CUDA OpenCV with a clang compiler but using gcc for the cuda compiler (CUDA_HOST_COMPILER=/usr/bin/gcc). This is also an extremely common scenario as the latest versions of gcc and clang are not supported by Clang, so you often need to specify an older (usually gcc) compiler. Arch-based distros install gcc-5 with cuda.

Note: CUDA still throws warnings about -Wmissing-prototypes and -Wstrict-prototypes when compiling CUDA with gcc but OpenCV with clang. This is because they share CXX flags and there is no checks done to see what the cuda compiler is. Crucially, it compiles

```
force_builders=Custom
```

### DIFF
--- a/cmake/OpenCVDetectCUDA.cmake
+++ b/cmake/OpenCVDetectCUDA.cmake
@@ -3,7 +3,7 @@ if(WIN32 AND NOT MSVC)
   return()
 endif()
 
-if(NOT APPLE AND CV_CLANG)
+if(NOT UNIX AND CV_CLANG)
   message(STATUS "CUDA compilation is disabled (due to Clang unsupported on your platform).")
   return()
 endif()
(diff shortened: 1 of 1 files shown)


alalek:

So, please:

cmake/OpenCVDetectCUDA.cmake
```diff
@@ -188,6 +188,11 @@ if(CUDA_FOUND)
     foreach(var CMAKE_CXX_FLAGS CMAKE_CXX_FLAGS_RELEASE CMAKE_CXX_FLAGS_DEBUG)
       set(${var}_backup_in_cuda_compile_ "${${var}}")
 
+      # we remove -Winconsistent-missing-override and -Qunused-arguments
+      # just in case we are compiling CUDA with gcc but OpenCV with clang
+      string(REPLACE "-Winconsistent-missing-override" "" ${var} "${${var}}")
+      string(REPLACE "-Qunused-arguments" "" ${var} "${${var}}")
```
alalek: This should be guarded by `CV_CLANG` check.

xsacha:
Made those changes.

PR #12121, opened 2018-07-31
Fix in SSE2 optimizations of types conversion routines (32s->16u)

maver1:
Fix for SSE2 intrinsics problem in the part of saturation arithmetic processing during 32s->16u packed conversion - for big negative values less than -INT_MAX+32767 the sign of the numbers is lost due to overflow that leads to incorrect saturation to MAX value, instead of zero (e.g. convert32s8u(-2147465847) == 255).

CV_ENABLED_INTRINSICS settings shall be set to ON to reproduce the issue - it is not reproducible with disabled intrinsics.


vpisarev:
@maver1, thank you! However, your variant does not work properly for positive numbers that exceed 32767.

maver1:
@vpisarev, you're right. I've added missed parts.

vpisarev:
:+1:

PR #12136, opened 2018-08-02
Update Documentation

sturkmen72:
### This pullrequest changes

i will add some more link in [examples page](http://pullrequest.opencv.org/buildbot/export/pr/12136/docs/examples.html)


samples/cpp/squares.cpp
```diff
@@ -8,22 +8,18 @@
...
 
 #include <iostream>
-#include <math.h>
-#include <string.h>
 
 using namespace cv;
 using namespace std;
 
-static void help()
+static void help(char* arg)
```
alalek: `const char* programName` would be better.

sturkmen72:
my changes about adding paths to examples is experimental. it looks good on [example tab](http://pullrequest.opencv.org/buildbot/export/pr/12136/docs/examples.html) but look ugly on frequently used functions see [waitkey](http://pullrequest.opencv.org/buildbot/export/pr/12136/docs/d7/dfc/group__highgui.html#ga5628525ad33f52eab17feebcfba38bd7)

@alalek what is your opinion?

samples/cpp/connected_components.cpp
```diff
@@ -1,3 +1,7 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html
```
alalek: It would be nice to avoid adding of license header into sample files.

doc/tutorials/imgproc/basic_geometric_drawing/basic_geometric_drawing.markdown
```diff
@@ -82,20 +81,20 @@ Code
 
 @add_toggle_cpp
 -   This code is in your OpenCV sample folder. Otherwise you can grab it from
-    [here](https://raw.githubusercontent.com/opencv/opencv/3.4/samples/cpp/tutorial_code/core/Matrix/Drawing_1.cpp)
-    @include samples/cpp/tutorial_code/core/Matrix/Drawing_1.cpp
+    [here](https://raw.githubusercontent.com/opencv/opencv/3.4/samples/cpp/tutorial_code/ImgProc/basic_drawing/Drawing_1.cpp)
+    @include samples/cpp/tutorial_code/ImgProc/basic_drawing/Drawing_1.cpp
```
alalek: BTW, Such renames breaks direct links of already released documentation (for example, 3.4.2 points to old 3.4 path).

Currently I don't have right approach how to work with this:
- we don't want to update/rebuild old documentation
- fix tag 3.4 -> 3.4.2 doesn't make sense (file is already part of documentation via `@include` statements). Also we prefer to keep "version increment" commit simple as possible.

Suggestions are welcome.

sturkmen72:
@alalek what is your opinion about keeping the files moved in their folder and create new one temporarily until release of OpenCV 4.0 (as you see last commmits )

doc/tutorials/imgproc/basic_geometric_drawing/basic_geometric_drawing.markdown
mshabunin: I think we can add an alias to Doxyfile:
```
ALIASES               += github{1}="https://github.com/opencv/opencv/blob/@CMAKE_DOXYGEN_GIT_COMMIT@/\1'"
```
Then get current commit hash in docs/CMakeLists.txt:
```.cmake
if(GIT_FOUND)
  execute_process(COMMAND "${GIT_EXECUTABLE}" rev-parse HEAD
    WORKING_DIRECTORY "${OpenCV_SOURCE_DIR}"
    OUTPUT_VARIABLE CMAKE_DOXYGEN_GIT_COMMIT
    ERROR_QUIET
    OUTPUT_STRIP_TRAILING_WHITESPACE
  )
else()
  set(CMAKE_DOXYGEN_GIT_COMMIT "3.4")
endif()
```
And then use this alias:
```
**Downloadable code**: Click [here](@github{samples/cpp/tutorial_code/ImgProc/BasicLinearTransforms.cpp})
```

alalek: Problem here is that such links are usually provided by this wording: "the latest code can be found here".

Also, It is not necessary to link on the current code - it is already attached into documentation in general.

sturkmen72:
with last rebased commit
license headers removed from sample files.
tutorial files moved without keeping old copies.

Issue #10471, opened 2017-12-30
Find Solution...I can Change cross-bar Video Input in Capture Video By Writing multi Line Code And Rebuild the Videoio Dll.


Ehsan61:
##### System information (version)

- OpenCV => 3.3
- Operating System / Platform => Windows 32 Bit 7
- Compiler => Visual Studio 2012

- OpenCV => :grey_question: I found Solution...
- Operating System / Platform => :grey_question: Win 7
- Compiler => :grey_question: Visual Studio 2012

##### Detailed description

I Modify Coding of videoio library for Access to Crossbar of VideoInput for Change Input pin Correctly.
i write some line of code to send index of input pin to lower layer of videoio library and Rebiuld it to capable of feature.
by this feature you can change video input pin of capture card very nice & simply correctly in windows and Directshow backend.

wanghanmin:
Yes.  I added one prop and some code and it runs ok on my cv710 capture card.

Ehsan61:
Ok.tanx...it should be in new version of opencv...for example 3.5 or 3.6
release.

PR #12138, opened 2018-08-03
videoio:new crossbar setting support

wanghanmin:
Solved #issues/[10471](https://github.com/opencv/opencv/issues/10471)

This change resolve directx based video capture can not work on some capture cards (avermeida CV710, GC550 etc). These cards will use PhysConn_Video_SerialDigital or PhysConn_Video_YRYBY as crossbar input pin.

Also fixed one issue that pVih->rcSource did not change after  HEADER(pVih)->biWidth changed to attemptWidth/attemptHeight will cause can not find suitable MediaFormat

### This pullrequest changes
Changed videoio.hpp to add new VideoCapturePropertie: "CAP_CROSSBAR_INPIN_TYPE"  for clossbar input pin setting.
Changed cap_dshow.cpp to apply the CAP_CROSSBAR_INPIN_TYPE setting in capture workflow.
fixed issue in function "SetSizeAndSubtype" to avoid fail in new attemptWidth and attemptHeight parameters provide.


wanghanmin:
change code for missed one ";" when paste.

wanghanmin:
missed one ";" when paste code. Reworked to add it.

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -157,6 +157,7 @@ enum VideoCaptureProperties {
        CAP_PROP_FOCUS         =28,
        CAP_PROP_GUID          =29,
        CAP_PROP_ISO_SPEED     =30,
+       CAP_CROSSBAR_INPIN_TYPE      =31, //!<CrossBar input pin Setting
```
alalek: Please add enum values to the end of list only.

wanghanmin: I have changed enum value to the end of list. thanks

modules/videoio/test/test_camera.cpp
```diff
@@ -36,6 +36,30 @@ TEST(DISABLED_VideoIO_Camera, basic)
     printf("Processed %d frames on %.2f FPS\n", N, (N * cv::getTickFrequency()) / (time1 - time0 + 1));
 
     capture.release();
+    
+    //Following test if for capture device using PhysConn_Video_SerialDigital as crossbar input pin
```
alalek: Please create separate test:
```
}

TEST(DISABLED_VideoIO_Camera, dshow_avermedia_capture)
{
    ... your test code ...
```

PR #12142, opened 2018-08-03
dnn(perf): fix and merge Convolution tests

alalek:
- OpenCL tests didn't run any OpenCL kernels
- use real configuration from existed models
- changed dump format for DNN Backend/Target in tests

Configuration list is prepared by using of this patch: https://github.com/alalek/opencv/commit/dnn_dump_conv_kernels

```
allow_multiple_commits=1
buildworker:Custom=linux-2
```


modules/dnn/perf/perf_convolution.cpp
```diff
@@ -1,92 +1,680 @@
...
+    operator int() const { return val_; }
+    static ::testing::internal::ParamGenerator<ConvParamID> all()
+    {
+#if 0
+        enum { NUM = (int)CONV_LAST };
+#else
+        enum { NUM = (int)CONV_100 };
+#endif
+        ConvParamID v_[NUM]; for (int i = 0; i < NUM; ++i) { v_[i] = ConvParamID(i); } // reduce generated code size
+        return ::testing::ValuesIn(v
```
dkurt: I think we can skip OpenCL targets if OpenCL is not available to reduce duplications:
```cpp
static testing::internal::ParamGenerator<tuple<Backend, Target> > testBackendsAndTargets()
{
    static std::vector<tuple<Backend, Target> > targets;
    if (targets.empty())
    {
        targets.push_back(tuple<Backend, Target>(DNN_BACKEND_OPENCV, DNN_TARGET_CPU));
#ifdef HAVE_OPENCL
        if (cv::ocl::useOpenCL())
        {
            targets.push_back(tuple<Backend, Target>(DNN_BACKEND_OPENCV, DNN_TARGET_OPENCL));
            targets.push_back(tuple<Backend, Target>(DNN_BACKEND_OPENCV, DNN_TARGET_OPENCL_FP16));
        }
#endif
    }
    return testing::ValuesIn(targets);
}

INSTANTIATE_TEST_CASE_P(/**/, Conv, Combine(
    ConvParamID::all(),
    testBackendsAndTargets()
));
```

alalek: [Done](https://github.com/opencv/opencv/pull/12142/files#diff-c490d0b09b78b7645157b8d537cbbd31R238).

modules/dnn/test/test_precomp.hpp
```diff
@@ -95,22 +66,6 @@ static testing::internal::ParamGenerator<Target> availableDnnTargets()
     return testing::ValuesIn(targets);
 }
 
-static testing::internal::ParamGenerator<tuple<Backend, Target> > dnnBackendsAndTargets()
```
alalek: Moved into test_common.hpp (file is shared between accuracy and perf tests)

dkurt: @alalek, Could you please also replace `DNNBackend` and `DNNTarget` at https://github.com/opencv/opencv/blob/3.4/modules/dnn/perf/perf_net.cpp?

alalek: Done, these changes are placed as a separate commit.

modules/dnn/test/test_common.hpp
```diff
@@ -190,4 +231,45 @@ static inline bool readFileInMemory(const std::string& filename, std::string& co
     return true;
 }
 
+namespace opencv_test {
+
+using namespace cv::dnn;
+
+static testing::internal::ParamGenerator<tuple<Backend, Target> > dnnBackendsAndTargets(bool withInferenceEngine = true, bool withHalide = true)
```
dkurt: May be make `withHalide` is `false` by default? There are only two places where it's tested:

https://github.com/opencv/opencv/blob/3e027df5832ddb135f144d28b81c19bdf8a40133/modules/dnn/test/test_halide_layers.cpp#L47-L64

https://github.com/opencv/opencv/blob/3e027df5832ddb135f144d28b81c19bdf8a40133/modules/dnn/test/test_backends.cpp#L288-L301

Both have no `DNN_BACKEND_OPENCV, DNN_TARGET_CPU` because it is used as reference for outputs so I think we can apply this method something like this:
```cpp
dnnBackendsAndTargets(bool withInferenceEngine = true, bool withHalide = false, bool withCpuOCV = true)
```

alalek: Done.

PR #12147, opened 2018-08-04
add new chessboard detector

D-Alex:
The chessboar detector is based on the paper.
Accurate Detection and Localization of Checkerboard Corners for
Calibration Alexander Duda, Udo Frese
British Machine Vision Conference, o.A., 2018.

It utilizes point symmetry of checkerboard corners in combination with a
localized Radon transform approximated by box filters to achieve high
performance even on large images. Here, tests have shown that the
ability to localize checkerboard corners is close to the theoretical
limit of 1/100 of a pixel while being considerably less sensitive
to image noise than standard methods.

### This pullrequest changes

It adds a new method cv::findChessboardCorners2 which is analog to findChessboardCorners but usually outperforms it in respect to detection rate, accuracy and speed.

TODO:
- [ ] performance test
- [ ] add flags for setting detection profile (pre-process/scale space ...)
- [x] compiler warnings
- [x] accuracy test
- [x] fix dev mode - visualizing detection steps. 
- [x] fix documentation  
- [x] fix flann dependency 
- [x] fix chessboard normalization for asymmetric chessboards 
```
force_builders=linux,docs,windows,macosx
```


D-Alex:
I can forward the paper to maintainers interested in getting the pull request merged. The largest code section is to ensure integrity of the chessboard. However, the amount of code is similar to the one already implemented (BMVC2018 - is next month).

The comparison was made against findChessboardCorners with an additional call to cornerSubPix with an optimal window size. 

comparison

timing

tompollok:
Great work! Do you know anything about the runtime compared to the current implementation?

tompollok:
How fast does your implementation return in case there is no chessboard or only visible partially? The current implementation takes quite long, which happens a lot when doing stereocalibrations from an stereo image sequence.

D-Alex:
There are some parameters which can be tweaked to massivily speeding up chessboard detection.
* omit super-resolution which will result in some small rounding errors
* limit the scale space for the search by assuming a certain size of the chessboard
* limit the number of tries per scale
  * It pickes randomly points identified as symmetric corner and starts to grow a chessboard using a guided search based on the centerlines and a kd-tree.

In a worst case scenario (checkerboard size mismatch) here are some results for a 1280x960 image:
findChessboardCorner: 4,5 sec (without additional corner refinement)
findChessboardCorner2  5,3 sec (pull request - highest detection rate + scale space)
findChessboardCorner2  2.3 sec (omit super resolution)
findChessboardCorner2  1,68 sec (omit super resolution + limit scale scape)
findChessboardCorner2  0,8 sec (omit super resolution + limit scale + limit tests per scale)

The detection is mostly linear and it can very gradually adjusted to the available CPU budget. We could add different profiles as flag to allow for different scenarios (high-speed detection, hight accuracy detection , ...) without exposing all the parameters. In addition the detection is based on blur and rotate primitives which could be pushed to the GPU.

modules/calib3d/src/chessboard.cpp
```diff
@@ -0,0 +1,3116 @@
...
+//        return cv::Point2f(a.x*b,a.y*b);
+//    }
+//
+    void normalizePoints1D(cv::InputArray _points,cv::OutputArray _T,cv::OutputArray _new_points)
+    {
+        cv::Mat points = _points.getMat();
+        if(points.cols > 1 && points.rows == 1)
+            points = points.reshape(1,points.cols);
+        if(points.channels() != 1 )
+            throw std::runtime_error("findHomography: only data with one channel are supported");
```
alalek: It is better to use `CV_Error()` macro to raise errors. Or even one from `CV_Check()` macro

modules/calib3d/src/chessboard.cpp
```diff
@@ -0,0 +1,3116 @@
...
+
+#include "precomp.hpp"
+// #include "opencv2/imgproc.hpp"
+#include "opencv2/flann.hpp"
+#include "chessboard.hpp"
+
+using namespace std;
+namespace cv
+{
+    /////////////////////////////////////////////////////////////////////////////
```
alalek: Please avoid indentation in namespaces.

modules/calib3d/src/chessboard.cpp
```diff
@@ -0,0 +1,3116 @@
...
+    {
+        if(angle == 0)
+        {
+            out = img;
+            return;
+        }
+        else
+        {
+            cv::Mat m = cv::getRotationMatrix2D(cv::Point2f(img.cols*0.5,img.rows*0.5),angle/M_PI*180,1);
+            assert(m.type() == CV_64FC1);
```
alalek: Please don't use `assert()`. Use `CV_Assert()` instead or `CV_Check()` macro.

modules/calib3d/src/chessboard.hpp
```diff
@@ -0,0 +1,788 @@
...
+#define CHESSBOARD_HPP_
+
+#include "opencv2/core.hpp"
+#include "opencv2/features2d.hpp"
+#include <vector>
+#include <set>
+#include <map>
+
+namespace cv
+{
```
alalek: Keep public API as simple as possible.
All implementation details should be moved into "src" folder or separate ".private.hpp" headers under "cv::details" namespace or similar.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -841,6 +841,34 @@ square grouping and ordering algorithm fails.
...
+@param image Source chessboard view. It must be an 8-bit grayscale or color image.
+@param patternSize Number of inner corners per a chessboard row and column
+( patternSize = cvSize(points_per_row,points_per_colum) = cvSize(columns,rows) ).
+@param corners Output array of detected corners.
+@param flags operation flags for future improvements
+
+The function is analog to findchessboardCorners but uses a localized radon
+transformation approximated by box filters being more robust to all sort of
+noise, faster on larger images and is able to directly return the sub-pixel
+position of the internal chessboard corners. The Method is based on the paper
```
catree: I think you should add the bibtex reference of the paper [here](https://github.com/opencv/opencv/blob/master/doc/opencv.bib).

Then you can do:

> The method is based on the paper @cite Duda2018 "Accurate Detection and Localization of Checkerboard Corners for Calibration" by A. Duda and U. Frese, BMVC 2018...

`@cite` will add a link to the bibliography in the form `[20]` but not the text.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -841,6 +841,34 @@ square grouping and ordering algorithm fails.
 CV_EXPORTS_W bool findChessboardCorners( InputArray image, Size patternSize, OutputArray corners,
                                          int flags = CALIB_CB_ADAPTIVE_THRESH + CALIB_CB_NORMALIZE_IMAGE );
 
+/** @brief Finds the positions of internal corners of the chessboard using a sector based approach.
+
+@param image Source chessboard view. It must be an 8-bit grayscale or color image.
+@param patternSize Number of inner corners per a chessboard row and column
+( patternSize = cvSize(points_per_row,points_per_colum) = cvSize(columns,rows) ).
```
catree: `Size` instead of `cvSize`?

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -841,6 +841,34 @@ square grouping and ordering algorithm fails.
...
+@param flags operation flags for future improvements
+
+The function is analog to findchessboardCorners but uses a localized radon
+transformation approximated by box filters being more robust to all sort of
+noise, faster on larger images and is able to directly return the sub-pixel
+position of the internal chessboard corners. The Method is based on the paper
+"Accurate Detection and Localization of Checkerboard Corners for Calibration"
+by A. Duda and U. Frese, BMVC 2018 demonstrating that the returned sub-pixel
+positions are more accurate than the one returned by cornerSubPix allowing a
+preciser camera calibration for demanding applications.
```
catree: `precise` instead of `preciser`?

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -841,6 +841,34 @@ square grouping and ordering algorithm fails.
...
+
+@note The function requires a white boarder with roughly the same width as one
+of the checkerboard fields around the whole board to improve the detection in
+various environments. In addition, because of the localized radon
+transformation it is beneficial to use round corners for the field corners
+which are located on the outside of the board. The following figure illustrates
+a sample checkerboard optimized for the detection. However, any other checkerboard
+can be used as well.
+![Checkerboard](pics/checkerboard_radon.png)
```
catree: I think the image is missing?

modules/calib3d/include/opencv2/calib3d.hpp
D-Alex: Fair, copy paste error from findChessboardCorners. I would propose to change it as well:
https://github.com/opencv/opencv/blob/053259fd92c1736d3920c7d43cb0400c518c4a5e/modules/calib3d/include/opencv2/calib3d.hpp#L795

D-Alex:
Concerning opencv_flann, it could be replaced by a bruce force search using a std::map for only one coordinate in combination with an upper and lower bound iterator. It might be even faster because flann seems to be optimized for more dimensions. 

Would it be mandatory to remove the dependency to flann at this stage?

alalek:
> mandatory to remove the dependency to flann at this stage

It is not required.

Just add flann dependency here:
https://github.com/opencv/opencv/blob/master/modules/calib3d/CMakeLists.txt
(currently it is attached via features2d)

ljere:
Would it be possible to reorder the detected corners like the original findChessboardCorners() does? I'm attaching a screenshot with two image pairs: first row shows the result of the new  findChessboardCorners2() function and the 2nd row the chessboard detected with the current OpenCV findChessboardCorners().

findchessboards2

ljere:
@D-Alex The advantage of asymmetric (even x odd corners or vice versa) targets is, that the position of the each corner on the target is uniquely defined. This is essential for stereo, multi-camera calibration or any task involving solvePnP and/or multiple targets and/or multiple cameras.

Which point to choose as the first point isn't that important. But that first point and the order of the following ones should be consistent and independent of the target orientation (see 2nd row in the image above).

tompollok:
I see that there is a lot of great work going on here. So will there be flags that allow to enable fast detection or will there only be an default option for high accuracy?

D-Alex:
Ideally yes. A considerable amount of time goes for example into finding seed points to grow an initial checkerboard. Therefore, this could be shortcut by providing the old key points as guess for continues tracking applications.  Also using the GPU for the feature map calculation would be extremely beneficial. Therefore, I think there is a lot of room for further improvements.

modules/calib3d/src/chessboard.cpp
```diff
@@ -0,0 +1,3295 @@
...
+}
+
+size_t Chessboard::Board::rowCount()const
+{
+    return rows;
+}
+
+cv::Size Chessboard::Board::getSize()const
+{
+    return cv::Si
```
alalek: Looks like color -> gray conversion is performed in [findChessboardCorners2](https://github.com/opencv/opencv/pull/12147/files#diff-8df6e9bc9d383d0c31ba9d371e491352R3263)

alalek: Perhaps this step can be moved into `findChessboardCorners2()` and this behavior controlled by flags:
- no image preprocessing
- this minmax scale
- adaptiveThreshold()
- etc

modules/calib3d/src/chessboard.hpp
```diff
@@ -0,0 +1,811 @@
...
+ the use of this software, even if advised of the possibility of such damage.
+*/
+
+/**
+ * Bilaniuk, Olexa, Hamid Bazargani, and Robert Laganiere. "Fast Target
+ * Recognition on Mobile Devices: Revisiting Gaussian Elimination for the
+ * Estimation of Planar Homographies." In Computer Vision and Pattern
+ * Recognition Workshops (CVPRW), 2014 IEEE Conference on, pp. 119-125.
+ * IEEE, 2014.
+ */
```
alalek: Please remove unrelated comment.
Please use the short OpenCV license header.

modules/calib3d/src/chessboard.hpp
```diff
@@ -0,0 +1,811 @@
...
+    public:
+        struct CV_EXPORTS_W Parameters
+        {
+            CV_PROP_RW float strength;       //!< minimal strength of a valid junction in dB
+            CV_PROP_RW float resolution;     //!< angle resolution in radians
+            CV_PROP_RW int branches;         //!< the number of branches
+            CV_PROP_RW int min_scale;        //!< scale level [0..8]
+            CV_PROP_RW int max_scale;        //!< scale level [0..8]
+            CV_PROP_RW bool filter;          //!< post filter feature map to improve impulse response
+            CV_PROP_RW bool super_resolution; //!< up-sample
```
alalek: `CV_EXPORTS_W` / `CV_PROP_RW` / `CV_WRAP` are used for bindings only (like Python API).
This header is internal (not a part of public API), so these macros have no effect.

modules/calib3d/src/chessboard.cpp
```diff
@@ -0,0 +1,3295 @@
...
+        CV_Error(Error::StsBadArg, "points must have only one channel");
+
+    // calc centroid
+    double centroid= cv::mean(points)[0];
+
+    // shift origin to centroid
+    cv::Mat new_points = points-centroid;
+
+    // calc mean distance
+    double mean_dist = cv::mean(cv::abs(points))[0];
```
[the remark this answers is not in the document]
D-Alex: good catch :-)

modules/calib3d/src/chessboard.hpp
D-Alex: The API was used in combination with https://github.com/D-Alex/rbind to automatically generate a ruby binding compatible with the types wrapped by https://github.com/D-Alex/ropencv. Here, like you said, it makes no sense and I will remove them.

modules/calib3d/src/chessboard.cpp
alalek: GitHub shows wrong code patch (again =( ).

Correct code patch here is [here](https://github.com/opencv/opencv/pull/12147/files#diff-8df6e9bc9d383d0c31ba9d371e491352R3112):
```
    //TODO is this needed?
    double min,max;
    cv::minMaxLoc(gray,&min,&max);
    gray = (gray-min)*(255.0/(max-min));
```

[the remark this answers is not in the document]
alalek: Correct code patch here is [here](https://github.com/opencv/opencv/pull/12147/files#diff-8df6e9bc9d383d0c31ba9d371e491352R2715):
```
if(pt.x != pt.x)
```

Similar code is [here](https://github.com/opencv/opencv/pull/12147/files#diff-8df6e9bc9d383d0c31ba9d371e491352R2811):
```
    if(iter->x != iter->x)
```

alalek: [Code](https://github.com/opencv/opencv/pull/12147/files#diff-8df6e9bc9d383d0c31ba9d371e491352R3101):
```
    cv::Mat gray;
    switch(image.type())
    {
    case CV_8UC3:
        cv::cvtColor(image,gray,cv::COLOR_BGR2GRAY);
    case CV_8UC1:
        gray = image;
        break;
    default:
        CV_Error(Error::StsUnsupportedFormat,"unsupported color format");
    }
```

modules/calib3d/test/test_chesscorners.cpp
```diff
@@ -83,6 +83,10 @@ class CV_ChessboardDetectorTest : public cvtest::BaseTest
     void run(int);
     void run_batch(const string& filename);
     bool checkByGenerator();
+    bool checkByGeneratorHighAccurcy();
```
alalek: typo

modules/calib3d/src/chessboard.cpp
```diff
@@ -0,0 +1,3209 @@
...
+
+FastX::FastX(const Parameters &para)
+{
+    reconfigure(para);
+}
+
+void FastX::reconfigure(const Parameters &para)
+{
+    if(para.min_scale < 0 || para.min_scale > para.max_scale)
+        CV_Error(Error::StsBadArg,"invalid scale");
```
alalek:
```
CV_Check(para.min_scale, para.min_scale >= 0 && para.min_scale <= para.max_scale, "invalid scale");
```

modules/calib3d/src/chessboard.cpp
```diff
@@ -0,0 +1,3209 @@
...
+
+cv::Point2f Ellipse::getCenter()const
+{
+    return center;
+}
+
+void Ellipse::draw(cv::InputOutputArray img,const cv::Scalar &color)const
+{
+    cv::Mat image = img.getMat();
+    cv::ellipse(image,center,axes,360-angle/M_PI*180,0,360,color);
```
alalek: > cv::Mat image = img.getMat();

It is not necessary.
`ellipse()` accepts `InputOutputArray`.

modules/calib3d/src/chessboard.cpp
```diff
@@ -0,0 +1,3209 @@
...
+        const cv::Point2f &pt3,const cv::Point2f &pt4)
+{
+    cv::Point3f p1(pt2.x-pt1.x,pt2.y-pt1.y,0);
+    cv::Point3f p2(pt4.x-pt3.x,pt4.y-pt3.y,0);
+    return p1.cross(p2).z > 0;
+}
+
+static bool sortKeyPoint(const cv::KeyPoint &pt1,const cv::KeyPoint &pt2)
+{
+    return pt1.response >= pt2.response;
```
alalek: Equality should be dropped from here.
if `sortKeyPoint(a, b)` is true, then `sortKeyPoint(b, a)` must be a false (it is "less" comparison functions, not "less or equal").

modules/calib3d/src/chessboard.cpp
```diff
@@ -0,0 +1,3209 @@
...
+    if(dst.cols > 1 && dst.rows == 1)
+        dst = dst.reshape(1,dst.cols);
+    if(src.rows != dst.rows)
+        CV_Error(Error::StsBadArg, "size mismatch");
+    CV_CheckChannelsEQ(src.channels(), 1, "data with only one channel are supported");
+    CV_CheckChannelsEQ(dst.channels(), 1, "data with only one channel are supported");
+    if(src.rows < 3)
+        CV_Error(Error::StsBadArg, "at least three point pairs are needed");
+    if(src.type() !=  dst.type())
+        CV_Error(Error::StsBadArg,"src and dst must have the same type");
```
alalek:
```
CV_CheckTypeEQ(src.type(), dst.type(), "src and dst must have the same type");
```

modules/calib3d/src/chessboard.hpp
```diff
@@ -0,0 +1,803 @@
...
+  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
+
+  By downloading, copying, installing or using the software you agree to this license.
+  If you do not agree to this license, do not download, install,
+  copy or use the software.
+
+
+                          BSD 3-Clause License
+
+ Copyright (C) 2014, Olexa Bilaniuk, Hamid Bazargani & Robert Laganiere, all rights reserved.
```
alalek: Please use short license header:
```
```

modules/calib3d/src/chessboard.cpp
```diff
@@ -0,0 +1,3209 @@
...
+    std::vector<cv::Point2f> points;
+    cv::Point2f pt;
+    do
+    {
+        PointIter iter2(iter);
+        cv::Point2f *p0 = *iter2;
+        iter2.bottom();
+        cv::Point2f *p1 = *iter2;
+        iter2.bottom();
+        cv::Poin
```
alalek: Not used?

alalek: There is GitHub bug with code location, again =(

Code is [here](https://github.com/opencv/opencv/pull/12147/files#diff-8df6e9bc9d383d0c31ba9d371e491352R3173) (cv::findChessboardCorners2, line 3173):
```
std::vector<cv::Point2f> out_corners;
```

alalek:
@D-Alex Thank you for the contribution!

> findChessboardCorners2

Could you suggest some name suffix instead of "2" ?
- `findChessboardCorners_Duda` / `findChessboardCorners_Duda2018` (second one has some issues with possible improvements: we would like to avoid possible 2019, 2010 suffix)
- `findChessboardCorners_<some wording about algorithm: point symmetry and Radon transform>`

D-Alex:
The proposed method is sector based in comparison to the standard edge based methods. Therefore, findChessboardCornersSB could be something which might fit. Another option would be to try to implicitly encode the underlying method via the findChessboardCorners flags. However, I am not sure this is very user friendly.

D-Alex:
The paper describing the work is now available as free download:
http://bmvc2018.org/contents/papers/0508.pdf

alalek:
> findChessboardCornersSB

Thank you! Name looks good. Please update patch.

PR #12150, opened 2018-08-05
RFC: core: enable printf format warnings for cv::format

paroj:


alalek:
As you understand we can't merge builds with warnings/errors.

Unfortunately, cross-platform `printf` doesn't exist. Only `%s`, `%d`, `%f` and `%p` (debug only) works somehow.
Types like `size_t`, `int64`, etc requires some tricks / "verbose" type specifiers like `PRId64` (and the last one is missing on some platforms).
Perhaps we should cast large integers to `long long int` and using `%lld` specifier. This should cover large set of existed warnings.

BTW, there is my attempt to work on this: https://github.com/alalek/opencv/commit/gcc_printf_format_checks (it is not completed, 6 months).

paroj:
sorry, I should have added RFC to this.

Maybe if we target 4.0 instead, we can require MSVC 2015. It supports the cross-platform %zu for size_t and also %lld for int64 as mandated by C++11.

paroj:
I could fix all warnings in core as described above. However going over all remaining modules is too much to do for me alone..

PR #12160, opened 2018-08-06
Use secure URLs in README

splattael:
Saves a redirect from HTTP -> HTTPS each.

[ci skip]

```
force_builders=Docs
```


vpisarev:
thanks! :+1:

PR #12172, opened 2018-08-07
core: move const tables outside of dispatched code

alalek:
To avoid duplicates in binaries.

Size of libopencv_core.so (GCC 8.1.1):
- before: 14536616
- after: 14528232
- diff: ~8Kb


modules/core/src/mathfuncs.cpp
```diff
@@ -2023,4 +2024,382 @@ void cvSolvePoly(const CvMat* a, CvMat *r, int maxiter, int)
...
+    .67938098479579733801614338517538271844400,    .5069306930693069306930693069306930693069,
+    .68135922480790300781450241629499942064300,    .5059288537549407114624505928853754940711,
+    .68333355911162063645036823800182901322850,    .5049309664694280078895463510848126232742,
+    .68530400309891936760919861626462079584600,    .5039370078740157480314960629921259842520,
+    .68727057207096020619019327568821609020250,    .5029469548133595284872298624754420432220,
+    .68923328123880889251040571252815425395950,    .5019607843137254901960784313725490196078,
+    .69314718055994530941723212145818, 5.0e-01,
+};
+
+const double* getLogTab()
```
vpisarev: could you please rename it to `getLogTab64f()` and `getLogTab32f()`, respectively, in order to be more consistent with HAL notation?

modules/core/src/mathfuncs.cpp
```diff
@@ -2023,4 +2024,382 @@ void cvSolvePoly(const CvMat* a, CvMat *r, int maxiter, int)
...
+    static float CV_DECL_ALIGNED(64) expTab_f[EXPTAB_MASK+1];
+    static volatile bool expTab_f_initialized = false;
+    if (!expTab_f_initialized)
+    {
+        for( int j = 0; j <= EXPTAB_MASK; j++ )
+            expTab_f[j] = (float)expTab[j];
+        expTab_f_initialized = true;
+    }
+    return expTab_f;
+}
```
vpisarev: same, please, rename it to `getExpTab64f()`, `getExpTab32f()`

PR #12180, opened 2018-08-08
Rewrite the UMat Python wrapper

cv3d:
I prefer to merge #12303 first, but it does not really matter, as required commits are also included here.

As the title indicate, the current wrapper is very hacky and hard to maintain.

### This pullrequest changes
- Rewrote the UMat wrapper using the easy-binding macros, the mappable/phantom types, and the raw pointer wrapper, all introduced in #12269

```
force_builders=Docs,linux,ocllinux,linuxNoOpt,ocl,windows,macosx,oclmacosx,ios,android
allow_multiple_commits=1
```


cv3d:
@alalek It seems `ENABLE_PIC` is required for a position independent code.
Should not that be enabled by default in Linux-based test platforms?

alalek:
Not always. Some Linux platforms have no PIC support (embedded boards), some don't support `.so` files at all (PIC is used for `.so` creation in general, single firmware binary can be build without PIC requirement from static libs).

What is the problem?

cv3d:
@alalek I am not sure what to do with this error! Further details are [here](http://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/16891/steps/compile%20release/logs/stdio).
```
/usr/bin/ld: CMakeFiles/opencv_python2.dir/__/src2/cv2.cpp.o: relocation R_X86_64_PC32 against undefined symbol `_Z11pyopencv_toIjEbP7_objectRT_PKc' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
```

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -61,7 +61,7 @@ namespace cv
 //! @addtogroup core_basic
 //! @{
 
-enum { ACCESS_READ=1<<24, ACCESS_WRITE=1<<25,
+enum CV_EXPORTS_W_SIMPLE { ACCESS_READ=1<<24, ACCESS_WRITE=1<<25,
```
alalek: CV_EXPORTS_W_SIMPLE doesn't work for enums.

All these flags are already in Python: `cv.ACCESS_READ`

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -2408,7 +2411,7 @@ class CV_EXPORTS UMat
     //! assignment operators
     UMat& operator = (const UMat& m);
 
-    Mat getMat(int flags) const;
+    CV_WRAP_AS(get) Mat getMat(int flags = ACCESS_RW) const;
```
alalek: Result lifetime of this method must be restricted to local code scope.

Nearest Python alternative is "with" statement (enter / exit) methods.

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -2364,31 +2364,34 @@ typedef Mat_<Vec3d> Mat3d;
 typedef Mat_<Vec4d> Mat4d;
 
 /** @todo document */
-class CV_EXPORTS UMat
+class CV_EXPORTS_W UMat
```
alalek: This must be avoided.
We don't want to restrict C++ interface of this class and respect bindings restrictions. Like `cv::Mat` implement this using featutes of target language.

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -2364,31 +2364,34 @@ typedef Mat_<Vec3d> Mat3d;
...
+    CV_WRAP UMat(Size size, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
 
     //! constructs n-dimensional matrix
-    UMat(int ndims, const int* sizes, int type, UMatUsageFlags usageFlags = USAGE_DEFAULT);
-    UMat(int ndims, const int* sizes, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+    CV_WRAP UMat(int ndims, const int* sizes, int type, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+    CV_WRAP UMat(int ndims, const int* sizes, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+
+    //! constructs from a Mat
+    CV_WRAP UMat(const Mat& m, int accessFlags = ACCESS_RW, UMatUsageFlags usageFlags = USAGE_DEFAULT);
```
alalek: There is no such ctor, because "mapping" is preferable (cv::Mat::getUMat()).
In other cases: use "copyTo()".

Implementation is invalid, see note about lifetime requirement. Map/unmap operations is restricted to local code scope.

modules/python/test/test_umat.py
```diff
@@ -12,22 +12,15 @@ def test_umat_construct(self):
...
         data_um = cv.UMat(data)  # from ndarray
-        data_sub_um = cv.UMat(data_um, [128, 256], [128, 256])  # from UMat
+        data_sub_um = cv.UMat(data_um, (128, 256), (128, 256))  # from UMat
         data_dst_um = cv.UMat(128, 128, cv.CV_64F)  # from size/type
         # test continuous and submatrix flags
         assert data_um.isContinuous() and not data_um.isSubmatrix()
         assert not data_sub_um.isContinuous() and data_sub_um.isSubmatrix()
         # test operation on submatrix
-        cv.multiply(data_sub_um, 2., dst=data_dst_um)
+        cv.multiply(data_sub_um, cv.UMat((128, 128), cv.CV_64F, 2.), dst=data_dst_um)
```
alalek: this operation is multiply "by scalar".

modules/python/test/test_umat.py
```diff
@@ -12,22 +12,15 @@ def test_umat_construct(self):
...
         assert np.allclose(2. * data[128:256, 128:256], data_dst_um.get())
 
-    def test_umat_handle(self):
```
alalek: "Handle" was added to pass OpenCL buffer handles to other libraries like PyOpenCL.

It can't be removed.

modules/core/include/opencv2/core/mat.hpp
cv3d: @alalek Since `CV_EXPORTS_W` is an alias to `CV_EXPORTS`, C++ interface is not restricted whatsoever.

I cannot understand why you think this it needs to be revised, but if you insist, then there is actually no other way but to close the PR, since it fully depend on this exact line. What do you think?

cv3d: @alalek In fact, I believe that even `cv::Mat` needs to eventually be wrapped properly.
It is really a waste of processing power to convert the data to NumPy format back and forth before and after each function call. A better user control and performance gain can be achieved if the data stay in the C++ world (i.e. as `cv::Mat`) until the user explicitly require a conversion, similar to the current `UMat.get()` behavior.
You might think that Python is not about speed, but it is really becoming so. I hope you will consider your decision again..

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -2364,31 +2364,42 @@ typedef Mat_<Vec3d> Mat3d;
...
+#ifdef _BINDINGS_ONLY_CODE
+    //! wrapper-only constructor from a Mat
+    CV_WRAP_CALL(self->v.reset(new cv::UMat(m.getUMat(accessFlags, usageFlags)))) UMat(const Mat& m, int accessFlags = ACCESS_RW, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+
+    //! returns the OpenCL queue used by OpenCV UMat
+    CV_WRAP_CALL(retval = cv::ocl::Queue::getDefault().ptr()) static void* queue() const;
+
+    //! returns the OpenCL context used by OpenCV UMat
+    CV_WRAP_CALL(retval = cv::ocl::Context::getDefault().ptr()) static void* context() const;
+#endif
```
cv3d: This code does not change the interface, and provides the required wrapper-only functions that were introduced by the original wrapper. I hope you would not mind such idea, would you?

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -2408,7 +2419,7 @@ class CV_EXPORTS UMat
     //! assignment operators
     UMat& operator = (const UMat& m);
 
-    Mat getMat(int flags) const;
+    CV_WRAP_AS(get) Mat getMat(int flags CV_WRAP_DEFAULT(ACCESS_RW)) const;
```
cv3d: The interface is not changes, as `CV_WRAP_DEFAULT` affects only the wrapper. We can also utilize `CV_DEFAULT` instead to affect both the wrapper and the C++ interface, without affecting the C interface.

alalek:
Something like this I mean in my previous comment.
Lets leave OpenCV C++ core structures C++ only. We don't want to overcomplicate C++ interfaces. Lets keep C++ code clean.

Bindings relates stuff should go into bindings module. It will allow to catch bindings-specific issues outside of main C++ code.

cv3d:
@alalek So your only problem now is the `_BINDINGS_ONLY_CODE` block, or do you even consider even `CV_WRAP_DEFAULT` problematic as well?

If, by any chance, you mean a complete separation of the binding code and the C++ interface, then I wonder what is the meaning of the automatic binding generator? To ensure a complete separation, UMat was completely hand-written in an ugly and hard-to-maintain way with a lot of boilerplate code, but is that really necessary? I believe a good balance between automation and separation is needed.

alalek:
Problem has wide scope actually.
It is related not only to `cv::UMat`, but `cv::Mat` and other core things.

For example, there is no direct `cv::Mat` in Python. Instead of this we reusing Python numpy array to allow smooth transition and passing existed data into OpenCV library / using results from OpenCV into other Python libraries.
Similar specific things exist in Java (missing efficient per-pixel access) and other languages.

alalek:
> The idea in my mind is to introduce cv::Mat in Python

This change will be rejected.

cv3d:
> This change will be rejected.

Perhaps you will change your opinion after I push that PR. We will see about that ;)

@alalek Can you please revise this PR again?

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -2364,31 +2364,40 @@ typedef Mat_<Vec3d> Mat3d;
...
-    UMat(int ndims, const int* sizes, int type, UMatUsageFlags usageFlags = USAGE_DEFAULT);
-    UMat(int ndims, const int* sizes, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+    CV_WRAP UMat(int ndims, const int* sizes, int type, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+    CV_WRAP UMat(int ndims, const int* sizes, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+
+    //! Mat is mappable to UMat
+    CV_WRAP_MAPPABLE(Ptr<Mat>);
+
+    //! returns the OpenCL queue used by OpenCV UMat
+    CV_WRAP_PHANTOM(static void* queue());
```
cv3d: The new wrapper functionalities became elegant and clean..
@alalek I hope you would not have problems, do you?

cv3d:
@alalek @mshabunin Can you please consider this PR?

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -2364,31 +2364,40 @@ typedef Mat_<Vec3d> Mat3d;
...
-    UMat(int rows, int cols, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
-    UMat(Size size, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+    CV_WRAP UMat(int rows, int cols, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+    CV_WRAP UMat(Size size, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
 
     //! constructs n-dimensional matrix
-    UMat(int ndims, const int* sizes, int type, UMatUsageFlags usageFlags = USAGE_DEFAULT);
-    UMat(int ndims, const int* sizes, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+    CV_WRAP UMat(int ndims, const int* sizes, int type, UMatUsageFlags usageFlags = USAGE_DEFAULT);
+    CV_WRAP UMat(int ndims, const int* sizes, int type, const Scalar& s, UMatUsageFlags usageFlags = USAGE_DEFAULT);
```
alalek: > int ndims, const int* sizes

Completely unsafe API for Python/Java bindings.

cv3d:
@alalek 
> Declare bindings-only "shadow" class in `misc` directory.

If my understanding is correct, `misc/python/pyopencv*` files are **not parsed** but included into the binding code. Am I missing anything here, or do you expect/want me to develop the shadow mechanism thing?

> `.getMat()` must be used with "with" Python statement only. Including implementation of enter/exit methods.

Is that ensured by the current implementation? I noticed in the test file that `.get()` is utilized without "with" statement.

modules/core/misc/python/shadow_umat.hpp
```diff
@@ -0,0 +1,60 @@
+#ifdef HAVE_OPENCV_CORE
+
+#include "opencv2/core/mat.hpp"
```
alalek: > #ifdef HAVE_OPENCV_CORE

Check is not necessary.

> #include "opencv2/core/mat.hpp"

This file is not intended for processing by C++ compiler. Only bindings parser should handle this file.
Perhaps you can leave short comment message here about that.

modules/python/bindings/CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ foreach(m ${OPENCV_PYTHON_MODULES})
       list(APPEND opencv_hdrs "${hdr}")
     endif()
   endforeach()
+  file(GLOB hdr ${OPENCV_MODULE_${m}_LOCATION}/misc/python/shadow*.hpp)
+  list(APPEND opencv_hdrs ${hdr})
   file(GLOB userdef_hdrs ${OPENCV_MODULE_${m}_LOCATION}/misc/python/pyopencv*.hpp)
```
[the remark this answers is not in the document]
alalek: Looks good to me.

modules/core/misc/python/shadow_umat.hpp
cv3d: Done!

alalek:
> `.getMat()`

The semantic of this call is:
- create new mapped data
- until new mapped data is not released you can't use parent object (to avoid UB).
- releasing of nested object synchronizes data back to parent object (if there is `ACCESS_WRITE`)
- any exception (or other control flow break) without releasing of nested object, make parent object into invalid state until all nested objects are released in proper order.

> that .get() is utilized without "with" statement

Looks like this should be replaced to copyTo() call semantic (without `ACCESS_WRITE` flag) - to avoid "locking" of parent object.
Actually, it is already [done by the current implementation](https://github.com/opencv/opencv/blob/3.4.2/modules/python/src2/cv2.cpp#L562-L574) without using of `.getMat()`.

Python's `with` statement is used to properly control resources: https://www.python.org/dev/peps/pep-0343/

modules/python/src2/cv2.cpp
```diff
@@ -553,273 +554,32 @@ PyObject* pyopencv_from(const cv::Ptr<T>& p)
     return pyopencv_from(*p);
 }
 
-typedef struct {
-    PyObject_HEAD
-    UMat* um;
-} cv2_UMatWrapperObject;
+CV_PY_TO_CLASS(UMat);
+CV_PY_FROM_CLASS(UMat);
+CV_PY_TO_ENUM(UMatUsageFlags);
```
alalek: This code can be moved to `misc/python`.

modules/python/src2/cv2.cpp
```diff
@@ -553,273 +554,32 @@ PyObject* pyopencv_from(const cv::Ptr<T>& p)
...
-// UMatWrapper.isContinuous() - returns true if the matrix is a submatrix of another matrix
-static PyObject * UMatWrapper_isSubmatrix(PyObject* self_, PyObject * /*args*/)
+static bool cv_mappable_to(const Ptr<Mat>& src, Ptr<UMat>& dst)
 {
-    cv2_UMatWrapperObject* self = (cv2_UMatWrapperObject*)self_;
-    if (self == NULL)
-        return failmsgp("Incorrect type of self (must be 'cv2_UMatWrapperObject')");
-    return PyBool_FromLong(self->um->isSubmatrix());
+    dst.reset(new UMat(src->getUMat(ACCESS_RW)));
+    return true;
```
alalek: ".getUMat()" requires strong local scope processing with proper "release" process too.
Result should not be stored somewhere with unknown lifetime.

cv3d:
@alalek All done!

cv3d:
@alalek Kind reminder

modules/core/misc/python/pyopencv_umat.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+CV_PY_TO_CLASS(UMat);
+CV_PY_FROM_CLASS(UMat);
+CV_PY_TO_ENUM(UMatUsageFlags);
+
+static bool cv_mappable_to(const Ptr<Mat>& src, Ptr<UMat>& dst)
+{
+    //dst.reset(new UMat(src->getUMat(ACCESS_RW)));
+    dst.reset(new UMat());
+    dst->allocator = &g_numpyAllocator;
```
alalek: Why is allocator replacement necessary for `UMat`? It is not intended for processing via numpy .

cv3d: @alalek Now fixed

cv3d:
@alalek @pwuertz I installed `pyopencl` and `mako`, and run the provided example without any issues using the new UMat wrapper.

alalek:
Looks like, there is some new issue in regular nightly [builds](http://pullrequest.opencv.org/buildbot/builders/master-skx-lin64/builds/235/steps/test_python2-ippicv-opencl/logs/stdio
) on Skylake-X machine (CPU is without iGPU, only discrete GPU is installed).

cv3d:
@alalek I want to investigate further. Shall I use the following or what?
```
**WIP**
force_builders=skx_lin64
```

alalek:
```
**WIP**
test_modules=python2,python3
force_builders=Custom
buildworker:Custom=linux-3
docker_image:Custom=ubuntu:16.04
test_opencl:Custom=ON
```

Looks like it is sporadic failure, restarted build is [fine](http://pullrequest.opencv.org/buildbot/builders/master-skx-lin64/builds/236)

cv3d:
I think I got it! The error is because of `return NULL` in the definition of `UMat.get`:
```C++
    if(PyObject_Size(args) == 0 && (!kw || PyObject_Size(kw) == 0))
    {
        Mat retval;
        retval.allocator = &g_numpyAllocator;
        ERRWRAP2(_self_->copyTo(retval));
        return pyopencv_from(retval);
    }

    return NULL;
```
It seems the usage of `cv.UMat.get(umat)` would not ALWAYS work, but `umat.get()` should always do. Shall I investigate further, or just use change the test case to utilize `umat.get()`?

PR #12181, opened 2018-08-08
(2.4) gpu(sparse_multi): fix definition without TBB

alalek:
related #11060 

```
force_builders=Custom
```


alalek:
:+1:

PR #12189, opened 2018-08-09
cleanup: IPP Async (IPP_A)

alalek:
except header file with conversion routines (will be removed in OpenCV 4.0)


vpisarev:
let it go :+1:

PR #12192, opened 2018-08-10
Pfm

### LINKED ISSUE #5053: PFM image file format reading with imread
Transferred from http://code.opencv.org/issues/4453

```
|| Chloe LeGendre on 2015-06-30 15:00
|| Priority: Low
|| Affected: None
|| Category: highgui-images
|| Tracker: Feature
|| Difficulty: 
|| PR: 
|| Platform: None / None
```
## PFM image file format reading with imread

```
Add imread feature to read images of the format PFM (another HDR format).
(see: http://netpbm.sourceforge.net/doc/pfm.html).

http://stackoverflow.com/questions/29487925/opencv-how-to-read-pfm-files
```

pasbi:
resolves #5053

### Add pfm image file format support

Added pfm decoder and encoder
Added pfm codec test

[PFM](https://linux.die.net/man/5/pfm) is a simple image format. It stores 32-Bit float images and supports one or three channel, interpreted as gray scale or RGB.

Note that comments in the header (lines starting with '#') are not part of the standard, hence no comment is written and reading files with comments is not supported.

```
docker_image:Custom=fedora:28
```


modules/imgcodecs/src/grfmt_pfm.cpp
```diff
@@ -0,0 +1,314 @@
...
+  m_description = "Portable image format - float (*.pfm)";
+}
+
+PFMEncoder::~PFMEncoder()
+{
+}
+
+bool PFMEncoder::isFormatSupported(int depth) const
+{
+    return depth == CV_32FC1 || depth == CV_32FC3;
```
alalek: Depth doesn't contain information about number of channels. So `CV_32FC3` is a bad value here.
Valid depth values is [here](https://github.com/opencv/opencv/blob/3.4.2/modules/core/include/opencv2/core/hal/interface.h#L71-L77)  (C1 values are mapped to depth directly, i.e `CV_32FC1 == CV_32F`)

modules/imgcodecs/src/grfmt_pfm.cpp
```diff
@@ -0,0 +1,314 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
alalek: Please use short license header:
```
```

modules/imgcodecs/src/grfmt_pfm.cpp
```diff
@@ -0,0 +1,314 @@
...
+#include "grfmt_pfm.hpp"
+#include <iostream>
+
+#ifdef HAVE_IMGCODEC_PFM
+
+namespace {
+
+static_assert(sizeof(float) == 4, "float must be 32 bit.");
+
+bool is_little_endian()
```
alalek: It is better to make this check "compile time".

Something like this (snippet is from opencv_contrib):
```
#if (defined(_M_IX86) || defined(__i386__) || defined(__i386) || defined(i386)) \
    || (defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
# define OPENCV_PLATFORM_LITTLE_ENDIAN
#elif (defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)
    && defined(__GNUC__) && (__GNUC__>4 || (__GNUC__==4 && __GNUC_MINOR__>=3))
# define OPENCV_PLATFORM_BIG_ENDIAN
#endif
```


BTW, Currently we don't extremely care about big endian platforms.
As this feature is optional (can be disabled during CMake stage using `WITH_IMGCODEC_PFM=OFF` flag) you can implement here little-endian code only with some note about it.

modules/imgcodecs/src/grfmt_pfm.cpp
```diff
@@ -0,0 +1,314 @@
...
+    return false;
+  }
+
+  CV_Assert(abs(m_scale_factor) == 1.0);
+
+  for (int y = m_height - 1; y >= 0; --y) {
+    m_strm.getBytes(mat.ptr(y), m_width * mat.elemSize());
+    if (is_byte_order_swapped(m_scale_factor)) {
+      for (size_t i = 0; i < m_width * m_channels; ++i) {
+        swap_endianess(mat.ptr<float>(y)[i]);
```
alalek: Perhaps it is better to swap `uint32_t` instead of `float`.

modules/imgcodecs/src/grfmt_pfm.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+size_t PFMDecoder::signatureLength() const
+{
+    return 2;
+}
+
+bool PFMDecoder::checkSignature( const String& signature ) const
+{
+    return signature.size() >= 2 && signature[0] == 'P'
+                                 && ( signature[1] == 'f' || signature[1] == 'F' );
```
alalek: Could you please make it more robust. Add third symbol check for new line / space type. Something like [this](https://github.com/opencv/opencv/blob/3.4.2/modules/imgcodecs/src/grfmt_pam.cpp#L343-L353).

modules/imgcodecs/src/grfmt_pfm.cpp
```diff
@@ -0,0 +1,314 @@
...
+  if (m_channels == 3) {
+    mat = Mat3f(m_height, m_width);
+  } else if (m_channels == 1) {
+    mat = Mat1f(m_height, m_width);
+  } else {
+    CV_Error(Error::StsError, "Expected 1 or 3 channels.");
+    return false;
+  }
+
+  CV_Assert(abs(m_scale_factor) == 1.0);
```
alalek: How does this line correlate with line below:
```
mat /= abs(m_scale_factor);
```

modules/imgcodecs/src/grfmt_pfm.cpp
```diff
@@ -0,0 +1,314 @@
...
+{
+  if (!m_strm.isOpened()) {
+    CV_Error(Error::StsError, "Unexpected status in data stream");
+    return false;
+  }
+
+  if (m_channels == 3) {
+    mat = Mat3f(m_height, m_width);
+  } else if (m_channels == 1) {
+    mat = Mat1f(m_height, m_width);
```
alalek: `readData()` method should not re-create `Mat`.
You need to check/fit input data into passed Mat.

`m_type` should be properly filled.
or implement `.type()`method.

Also all used `imread()` calls (in tests) should be called with `IMREAD_UNCHANGED` flag (or `CV_LOAD_IMAGE_ANYDEPTH | CV_LOAD_IMAGE_ANYCOLOR`).

alalek:
@pasbi Currently all OpenCV image encoders should support 8U images.

In general there is no problem to write 8U values as floats.

```
Mat write_img;
if (img.depth() == CV_8U)
{
    img.convertTo(write_img, CV_32F, 1/255.0);  // result floats into range [0; 1]
}
else if (img.depth() == CV_32F)
{
    write_img = img;
}
else
{
    ... internal error (checks before that should exclude this code path) ...
}
```

BTW, Perhaps test code passed 8U image to `imwrite()`.

modules/imgcodecs/src/grfmt_pfm.cpp
alalek: See this comment: https://github.com/opencv/opencv/pull/12192#issuecomment-412051004

You can leave this "as is", but you should not pass 8U images in tests.

modules/imgcodecs/src/grfmt_pfm.cpp
pasbi: `swap_endianess` already works on `uint32_t` internally. But it can be simplified, though.

modules/imgcodecs/src/grfmt_pfm.cpp
pasbi: My mistake. I was not sure if the line should be `mat /= abs(m_scale_factor);` or `mat*/= abs(m_scale_factor);`, so I decided to drop the support for scale factors of magnitude other than 1.
Then I forgot to remove that line below.

What do you think? `*= scale`, `/=scale` or `assert(scale==1)`? The standard is pretty terse on this point.

alalek: Scale factor is not well defined in format specification. Looks like it is used as a standalone information / atrribute (OpenCV doesn't support extra attributes exporting via `imread`).

I believe we can ignore it now.

modules/imgcodecs/src/grfmt_pfm.cpp
pasbi: So if `IMREAD_UNCHANGED` is not passed to `imread`, the type changes silently to `CV_8UC1` or `CV_8UC3`, right?
But this will not work, as the image buffer will be too small (assuming 1 byte elements rather than 4 bytes). Hence it will crash or at least fail to read the image.
Is that what we want? Shouldn't no flags mean to load it as CV_32F1 or CV_32F3 in the PFM case?

alalek: > the type changes silently to CV_8UC1 or CV_8UC3

Right (8UC3 - if `IMREAD_COLOR` flag passed). This is default behavior of `cv::imread()` (independent of any used codec).

You should read into your buffer / Mat. And after reading, convert to passed `img`.

Or leave an exception. There is no convenient way to convert float number into int8.

modules/imgcodecs/test/test_grfmt.cpp
```diff
@@ -337,6 +342,32 @@ TEST(Imgcodecs_Pam, read_write)
 }
 #endif
 
+#ifdef HAVE_IMGCODEC_PFM
+TEST(Imgcodecs_Pfm, read_write)
+{
+  string folder = string(cvtest::TS::ptr()->get_data_path()) + "readwrite/";
```
alalek: > cvtest::TS::ptr()->get_data_path()

Use cvtest::findDataFile() instead.

modules/imgcodecs/test/test_grfmt.cpp
```diff
@@ -337,6 +342,32 @@ TEST(Imgcodecs_Pam, read_write)
...
+  string filepath = folder + "lena.pfm";
```
alalek: Lets reuse existed ["lena.pam"](https://github.com/opencv/opencv_extra/blob/3.4.2/testdata/highgui/readwrite/lena.pam) file as input (pfm files are really huge).

```
Mat img = imread(findDataFile("readwrite/lena.pam"));
ASSERT_FALSE(img.empty());
img.convertTo(img, CV_32F, 1/255.0f);
```

modules/imgcodecs/test/test_grfmt.cpp
```diff
@@ -337,6 +342,32 @@ TEST(Imgcodecs_Pam, read_write)
...
+
+  string writefile_no_param = cv::tempfile(".pfm");
+  EXPECT_NO_THROW(cv::imwrite(writefile_no_param, img));
+  cv::Mat reread_no_param = cv::imread(writefile_no_param, IMREAD_UNCHANGED);
+
+  EXPECT_EQ(0, cvtest::norm(reread, reread_no_param, NORM_INF));
+  EXPECT_EQ(0, cvtest::norm(img, reread, NORM_INF));
+
+  remove(writefile.c_str());
+  remove(writefile_no_param.c_str());
```
alalek: EXPECT_EQ(0, remove(tmp_file.c_str()));

alalek: Static analyzer tool coverity.com dislike if `remove()` call is added without check: https://scan.coverity.com/projects/opencv?tab=overview

Also this check allows to detect on some platforms if file handles are not properly closed and `remove()` returns errors. Or just misused/wrong file name.

`get_data_path()` is legacy code.

pasbi:
ok sounds good. Well-formed endianess detection at compile time seems to be dowdy.
The build still fails on android because `std::to_string` is not defined. Isn't the android build using c++11, too? I'll use something more portable anyway.

alalek:
C++11 is enabled. There is Android-specific issue with `std::to_string()`.

Current `std::stringstream` approach looks good (`std::ostringstream` should be better)

alalek:
@pfm There is already [definition](https://github.com/opencv/opencv/blame/3.4.2/cmake/templates/cvconfig.h.in#L200) for BIGENDIAN detection.

It can be used like this:
```
#ifndef WORDS_BIGENDIAN
    ... little endian code ...
#else
    ... big endian ...
#endif
```

Sorry for inconvenience.

pasbi:
I appreciate your efforts ;)

it should be done now. Should I 'Clean up "oops" commits before submitting'? How can I do that?

PR #12222, opened 2018-08-15
Increase precision for intersection calculation of RotatedRect

### LINKED ISSUE #12221: RotatedRect fails to calculate correct intersections due to precision
- OpenCV => master
- Compiler => gcc

##### Detailed description

OpenCV RotatedRect intersection calculation currently uses float for intermediate variables. In some edge cases, the limited precision causes assertion problems inside intersection code:
```
   CV_Assert( intersection.size() <= 8 );
```
The problem can be mitigated by doing all calculations in double precision as shown below:
```
      float A = -vec2[j].y;
      float B = vec2[j].x;
      float C = -(A * pts2[j].x + B * pts2[j].y);

      float s = A * x + B * y + C;

//Using float: xx = -1990.670654, yy = 1636.783203,A = 200.753662, B = -121.916260, C = 599184.937500, x = -2071.253906, y = 1504.090820, s = 0.000000

      double A = -vec2[j].y;
      double B = vec2[j].x;
      double C = -(A * pts2[j].x + B * pts2[j].y);

      double s = A * x + B * y + C;

//Using double: xx = -1990.670654, yy = 1636.783203,A = 200.753662, B = -121.916260, C = 599184.910076, x = -2071.253906, y = 1504.090820, s = -0.023922
```

##### Steps to reproduce

See above example.

NCBee:
resolves #12221

### This pullrequest changes

Precision of calculation for intersection of RotatedRect.

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/imgproc/src/intersection.cpp
+++ b/modules/imgproc/src/intersection.cpp
@@ -51,12 +51,13 @@ int rotatedRectangleIntersection( const RotatedRect& rect1, const RotatedRect& r
 {
     CV_INSTRUMENT_REGION()
 
-    const float samePointEps = 0.00001f; // used to test if two points are the same
+    // L2 metric
+    const float samePointEps = std::max(1e-16f, 1e-6f * (float)std::max(rect1.size.area(), rect2.size.area()));
 
     Point2f vec1[4], vec2[4];
     Point2f pts1[4], pts2[4];
 
-    std::vector <Point2f> intersection;
+    std::vector <Point2f> intersection; intersection.reserve(24);
 
     rect1.points(pts1);
     rect2.points(pts2);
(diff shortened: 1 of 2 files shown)


alalek:
Rebased onto 3.4 branch and added second pass with removal of closest points (if pass with `samePointEps` is not enough).

SuperIRabbit:
I think that the float-to-double switch of the lines @NCBee mentioned in https://github.com/opencv/opencv/issues/12221 is better to be applied besides this change.

The magnitude of input coordinates to cause the float underflow problem is not something very big - it's at 10^3 level which shows up frequently in our use case. Using double (in those 6 lines) can significantly reduce the chance of returning mathematically incorrect result, regardless of the returned number of vertices.

PR #12232, opened 2018-08-16
Avoid multi-inheritance in the cuda::features2d classes

### LINKED ISSUE #12230: cudafeatures2d cannot be wrapped for Python
- OpenCV => 4.0.0-pre (master branch)
- Operating System / Platform => Independent
- Compiler => Independent

##### Detailed description
Both classes of `cuda::ORB` and `cuda::FastFeatureDetector` utilizes multiple inheritance, thus it impossible to wrap them into Python, and perhaps other languages.

cv3d:
resolves #12230

```
force_builders=Custom
```

### DIFF
--- a/modules/cudafeatures2d/include/opencv2/cudafeatures2d.hpp
+++ b/modules/cudafeatures2d/include/opencv2/cudafeatures2d.hpp
@@ -374,7 +374,7 @@ public:
 
 /** @brief Abstract base class for CUDA asynchronous 2D image feature detectors and descriptor extractors.
  */
-class CV_EXPORTS Feature2DAsync
+class CV_EXPORTS Feature2DAsync : public cv::Feature2D
 {
 public:
     virtual ~Feature2DAsync();
@@ -423,7 +423,7 @@ public:
 
 /** @brief Wrapping class for feature detection using the FAST method.
  */
-class CV_EXPORTS FastFeatureDetector : public cv::FastFeatureDetector, public Feature2DAsync
+class CV_EXPORTS FastFeatureDetector : public Feature2DAsync
 {
 public:
     enum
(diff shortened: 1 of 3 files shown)


modules/cudafeatures2d/include/opencv2/cudafeatures2d.hpp
```diff
@@ -434,11 +434,17 @@ class CV_EXPORTS FastFeatureDetector : public cv::FastFeatureDetector, public Fe
 
         FEATURE_SIZE = 7
     };
+    enum
+    {
+        TYPE_5_8 = 0, TYPE_7_12 = 1, TYPE_9_16 = 2,
+        THRESHOLD = 10000, NONMAX_SUPPRESSION = 10001, FAST_N = 10002,
```
alalek: Code duplication should be avoided.

Does "using FastFeatureDetector::TYPE_5_8;" work here?

cv3d: I tried it, but it says `error C2602: 'TYPE_9_16' is not a member of a base class of 'cv::cuda::FastFeatureDetector'`.
We can either duplicating the enum or add `cv::` to the default value, thus become `cv::FastFeatureDetector::TYPE_9_16`. What do you think?

cv3d:
@alalek Although the motivation is to enable binding, this PR is basically to avoid using multi-inheritance, which is implicitly tested by the fact it can compile properly, thus I believe it is a complete PR, and does not need any further clutter from binding stuff. As for the bindings generator, I am planning to introduce it for `cuda::features2d` in #12234 with its tests and binding stuff after this gets merged.

cv3d:
@alalek I can split #12234 into 10 PRs or so (or 10 commit in same PR), each for a different module, with unit tests for each module in a separated file (or accumulated in the same file, in case of several commit & 1 PR). In that case, this PR can contain its binding code and unit tests as well. Would that make things better from your perspective?

alalek:
As you wish. Splitting make sense for completely independent tasks.
Currently, bindings from different modules depend on some common things like changes in "modules/core/misc/python/pyopencv_core.hpp". So keeping the single PR is easy way for now.

PR #12156, opened 2018-08-06
Support Python binding for CUDA functionalities

cv3d:
resolves #4524

### This pullrequest changes
- Wraps GpuMat in a clean way, in compare to UMat. Perhaps UMat wrapper needs to be rewritten following this PR.
- Provide the Python binding generator with automatic fail-over for CUDA functionalities (Similar to UMat and OpenCL). An additional effort to unify the naming shall lead to a great CUDA/OpenCL/CPU fail-over system.
- Exposes some CUDA functionalities to the Python binder
- Provides unit tests for some basic CUDA-based pythonic functionalities
- Minor cleanup of the binder generation code, to match the current coding style

Well, Python is getting more and more famous, and deserves more attention now.
Now that several Python libraries are already CUDA-enabled, especially in the field of vision and learning, it is really a necessity for OpenCV to catch up. This PR is intended as a backbone for such great change.

```
force_builders=Docs,linux,linux32,ocllinux,linuxNoOpt,ocl,windows,win32,macosx,oclmacosx,ios,android,Custom
docker_image:Custom=ubuntu-cuda:16.04
```

PR #12234, opened 2018-08-16
Extensive wrapping of CUDA functionalities for Python

cv3d:
### This pullrequest changes
- Introduce some macros for easy binding
- Enhance `cuda::Stream` and `cuda::GpuMat::Allocator` bindings introduced in #12156
- Wraps additional CUDA functionalities for Python
- Extend the CUDA unit tests

BTW: Currently 6 macros were introduced, but I still plan to reduce them - if possible - into 4. Furthermore, unit tests are still incomplete. However, this PR is huge enough, therefore I prefer to make extensive testing and macros optimization (or template conversion) in another PR.


cv3d:
@alalek I am stuck again with the [same error as of before](http://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/17028/steps/compile%20release/logs/stdio). If my understanding is correct, template specialization of `pyopencv_from` failed for some type, but I cannot see any clue in the error message about the type being instantiated to debug any further. Can you please help?

alalek:
Try [this](https://github.com/opencv/opencv/blob/3.4.2/modules/python/common.cmake#L16-L17) for debugging:
```.diff
-set(OPENCV_FORCE_PYTHON_LIBS OFF CACHE BOOL "")
-string(REPLACE "-Wl,--no-undefined" "" CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS}")
+set(OPENCV_FORCE_PYTHON_LIBS ON CACHE BOOL "")
```

cv3d:
I am still getting almost [the same error](http://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/17030/steps/compile%20release/logs/stdio), with no enough details aside from the function name `_Z13pyopencv_fromIjEP7_objectRKT_`..
```
[ 97%] Linking CXX shared module ../../lib/python3/cv2.cpython-35m-x86_64-linux-gnu.so
/usr/bin/ld: CMakeFiles/opencv_python3.dir/__/src2/cv2.cpp.o: relocation R_X86_64_PC32 against undefined symbol `_Z13pyopencv_fromIjEP7_objectRKT_' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
modules/python3/CMakeFiles/opencv_python3.dir/build.make:141: recipe for target 'lib/python3/cv2.cpython-35m-x86_64-linux-gnu.so' failed
```
@alalek Any further guidance is highly appreciated.

alalek:
Modern compiler emits more accurate messages about this problem - check [this build](http://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/813).

modules/cudacodec/include/opencv2/cudacodec.hpp
```diff
@@ -152,14 +152,14 @@ class CV_EXPORTS EncoderCallBack
     @param frameNumber
     @param picType Specify frame type (I-Frame, P-Frame or B-Frame).
      */
-    virtual void onBeginFrame(int frameNumber, PicType picType) = 0;
+    CV_WRAP virtual void onBeginFrame(int frameNumber, EncoderCallBack::PicType picType) = 0;
```
cv3d: The wrapper converts the arguments into statements, thus enum arguments results in maybe-uninitialized warnings as in [this build](http://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/814/steps/compile%20release/logs/warnings%20%284%29).

I wonder which resolution is better...
- Arguments should be converted to integers
- The generator should  do better conversion
- Maybe-uninitialized warnings of the generator code should be silenced

alalek: What is about forced initialization of these enum values in generator code?
```.diff
-EncoderCallBack::PicType picType;
+EncoderCallBack::PicType picType = (EncoderCallBack::PicType)0;
```

cv3d: The question is how to tell which argument is enum and which is not.

cv3d: I totally agree, and hope we can do that in the near future.
However, for the time being, I gave up on binding these functions to avoid their warnings.

modules/cudacodec/include/opencv2/cudacodec.hpp
```diff
@@ -152,14 +152,14 @@ class CV_EXPORTS EncoderCallBack
...
+    virtual void onBeginFrame(int frameNumber, EncoderCallBack::PicType picType) = 0;
 
     /** @brief Callback function signals that the encoding operation on the frame has finished.
 
     @param frameNumber
     @param picType Specify frame type (I-Frame, P-Frame or B-Frame).
      */
-    virtual void onEndFrame(int frameNumber, PicType picType) = 0;
+    virtual void onEndFrame(int frameNumber, EncoderCallBack::PicType picType) = 0;
 };
```
cv3d: Not wrapped due to enum-related warnings, but no harm in keeping `EncoderCallBack::` there.

modules/core/misc/python/pyopencv_core.hpp
```diff
@@ -4,51 +4,27 @@
...
+CV_PY_TO_ENUM(cuda::HostMem::AllocType);
+CV_PY_TO_ENUM(cuda::FeatureSet);
+
+CV_PY_FROM_CLASS(cuda::GpuMat);
+CV_PY_FROM_CLASS(cuda::Stream);
+CV_PY_FROM_CLASS(cuda::HostMem);
+
+CV_PY_FROM_CLASS_PTR(cuda::GpuMat::Allocator);
+
+CV_PY_FROM_ENUM(cuda::DeviceInfo::ComputeMode);
```
cv3d: It would be really great if we can simplify all the binding codes in a similar way.

modules/cudaimgproc/include/opencv2/cudaimgproc.hpp
```diff
@@ -145,7 +145,7 @@ E.g. Given an RGBA image, aDstOrder = [3,2,1,0] converts this to ABGR channel or
 
 The methods support arbitrary permutations of the original channels, including replication.
  */
-CV_EXPORTS_W void swapChannels(InputOutputArray image, const int dstOrder[4], Stream& stream = Stream::Null());
+CV_EXPORTS void swapChannels(InputOutputArray image, const int dstOrder[4], Stream& stream = Stream::Null());
```
cv3d: Unfortunately, the generator cannot handle arrays in the form of `int dstOrder[4]` properly, thus undo the wrapping..
This bad behavior needs to be addressed, hopefully in the near future.

modules/cudacodec/include/opencv2/cudacodec.hpp
cv3d: @alalek Today, I have supported enum-type detection. I hope you would consider merging this PR before it gets any bigger...

cv3d:
@alalek This PR, although length, is a very straight-forward.
Can you please consider merging it?

cv3d:
@alalek Is there a need to move the binding code of `GpuMat` to a shadow file, similar to #12180?

modules/core/include/opencv2/core/cuda.hpp
```diff
@@ -478,29 +478,29 @@ and the corresponding memory is automatically returned to the pool for later usa
...
 @endcode
  */
-class CV_EXPORTS BufferPool
+class CV_EXPORTS_W BufferPool
 {
 public:
 
     //! Gets the BufferPool for the given stream.
-    explicit BufferPool(Stream& stream);
+    CV_WRAP explicit BufferPool(Stream& stream);
```
alalek: Generated Python code:
```
    PyObject* pyobj_stream = NULL;
    Stream stream;

    const char* keywords[] = { "stream", NULL };
    if( PyArg_ParseTupleAndKeywords(args, kw, "O:BufferPool", (char**)keywords, &pyobj_stream) &&
        pyopencv_to(pyobj_stream, stream, ArgInfo("stream", 0)) )
    {
        new (&(self->v)) Ptr<cv::cuda::BufferPool>(); // init Ptr with placement new
        if(self) ERRWRAP2(self->v.reset(new cv::cuda::BufferPool(stream)));
        return 0;
    }
```

> Stream stream;

Constructor expects reference, not an another copy of object.

modules/core/include/opencv2/core/cuda.hpp
```diff
@@ -521,46 +521,46 @@ Its interface is also Mat-like but with additional memory type parameters.
...
     enum AllocType { PAGE_LOCKED = 1, SHARED = 2, WRITE_COMBINED = 4 };
 
-    static MatAllocator* getAllocator(AllocType alloc_type = PAGE_LOCKED);
+    static MatAllocator* getAllocator(HostMem::AllocType alloc_type = HostMem::AllocType::PAGE_LOCKED);
 
-    explicit HostMem(AllocType alloc_type = PAGE_LOCKED);
+    CV_WRAP explicit HostMem(HostMem::AllocType alloc_type = HostMem::AllocType::PAGE_LOCKED);
 
-    HostMem(const HostMem& m);
+    CV_WRAP HostMem(const HostMem& m);
```
alalek: Do we really want to wrap copy ctor?

Reference is lost here too.

modules/core/include/opencv2/core/cuda.hpp
```diff
@@ -569,18 +569,18 @@ class CV_EXPORTS HostMem
     hardware. Laptops often share video and CPU memory, so address spaces can be mapped, which
     eliminates an extra copy.
      */
-    GpuMat createGpuMatHeader() const;
+    CV_WRAP GpuMat createGpuMatHeader() const;
```
alalek: Perhaps it not usable in bindings due missing access to memory pointer.

modules/core/include/opencv2/core/cuda.hpp
```diff
@@ -569,18 +569,18 @@ class CV_EXPORTS HostMem
...
-    size_t step1() const;
-    Size size() const;
-    bool empty() const;
+    CV_WRAP bool isContinuous() const;
+    CV_WRAP size_t elemSize() const;
+    CV_WRAP size_t elemSize1() const;
+    CV_WRAP int type() const;
+    CV_WRAP int depth() const;
+    CV_WRAP int channels() const;
+    CV_WRAP size_t step1() const;
```
alalek: `step1` is wrapped, but `step` is not.

Same for GpuMat`

modules/core/include/opencv2/core/cuda.hpp
```diff
@@ -808,23 +808,23 @@ class CV_EXPORTS TargetArchs
...
      */
-    static bool has(int major, int minor);
-    static bool hasPtx(int major, int minor);
-    static bool hasBin(int major, int minor);
-
-    static bool hasEqualOrLessPtx(int major, int minor);
-    static bool hasEqualOrGreater(int major, int minor);
-    static bool hasEqualOrGreaterPtx(int major, int minor);
-    static bool hasEqualOrGreaterBin(int major, int minor);
+    CV_WRAP static bool has(int major, int minor);
```
alalek: "static" is not properly handled in Python bindings generator: `METH_STATIC` should be used.

cv3d: `METH_STATIC` is now utilized

modules/core/include/opencv2/core/cuda.hpp
cv3d: `step` is now wrapped in both classes.

modules/core/include/opencv2/core/cuda.hpp
cv3d: Wrapper removed

modules/core/include/opencv2/core/cuda.hpp
cv3d: You are right! Wrapper removed.

modules/core/include/opencv2/core/cuda.hpp
cv3d: Dropped wrapping it for the time being

cv3d:
@alalek I agree. Kindly proceed with merge.

PR #12237, opened 2018-08-16
Split fix

### LINKED ISSUE #12236: TrainData incorrectly split datasets where (rows == cols)
- OpenCV => 3.4

##### Detailed description

I understand that this is rare case, but it happened to me. My train data had number of samples (rows) equal to the number of features (cols). Attempt to retrieve samples from TrainData causes crash in TrainData::getSubVector(). It was trying to work with my data like they were ordered COL_SAMPLE.

##### Steps to reproduce

	cv::Mat test = cv::Mat::zeros(15, 15, CV_32FC1);
	cv::Mat labels = cv::Mat::zeros(15, 1, CV_32SC1);
	cv::Ptr<cv::ml::TrainData> train_data = cv::ml::TrainData::create(test, cv::ml::ROW_SAMPLE, labels);
	train_data->setTrainTestSplitRatio(0.9);
	train_data->getTestResponses();

lamantine:
resolves #12236 

Hopefully I correctly understood what original condition in line 82 means - it was used only in case when we're retrieving responses and they're organized by COL_SAMPLE.

### DIFF
--- a/modules/ml/src/data.cpp
+++ b/modules/ml/src/data.cpp
@@ -79,7 +79,7 @@ Mat TrainData::getSubVector(const Mat& vec, const Mat& idx)
 
     Mat subvec;
 
-    if( vec.cols == m )
+    if( (dims == 1) && (vec.cols == m) )
         subvec.create(dims, n, type);
     else
         subvec.create(n, dims, type);


alalek:
Replaced by #12242

PR #12241, opened 2018-08-17
Add batch mode CPU.

kopytjuk:
### This pullrequest changes

For darknet (region_layer) there was no batch-mode possible. Forwarding a batch (blobFromImages) had no effect because the network only calculated the first image.


gylns:
@kopytjuk why you close this PR? I also found this issue #12247.

kopytjuk:
@gylns 
There were several reasons:
- The tests defined for dnn module crashed after my MR because they expect the old output shape. 
- Additionally I will create a new MR when I implemented both the CPU and OPENCL functions.

Issue #10631, opened 2018-01-18
Missing estimation method, maxIteration and refinement option in cv::estimateAffine3D


csukuangfj:
##### System information (version)

- OpenCV => master

##### Detailed description
Both [cv::estimateAffine3D][1] and [cv::estimateAffine2D][2] **SHOULD**
have similar signatures, however,

 - `cv::estimateAffine2D` returns the model as the return value
 - `cv::estimateAffine3D` returns the model as a parameter

 - `cv::estimateAffine2D` supports RANSAC and LMeDS
 - `cv::estimateAffine3D` supports only RANSAC

 - `cv::estimateAffine2D` supports to specify the maximum number of iterations
 - `cv::estimateAffine3D` dose not support maxIter

 - `cv::estimateAffine2D` supports to refine the estimated model
 - `cv::estimateAffine3D` does not support refinement

[2]: https://docs.opencv.org/master/d9/d0c/group__calib3d.html#ga27865b1d26bac9ce91efaee83e94d4dd
[1]: https://docs.opencv.org/master/d9/d0c/group__calib3d.html#ga396afb6411b30770e56ab69548724715

hrnr:
After some discussion `cv::estimateAffine*2D` functions has been designed after findHomography (which should be probably called estimateHomography). We have considered this interface to be better.

We should add overload to `cv::estimateAffine3D` and rework it to match capabilities of `cv::estimateAffine*2D`. 

The interface to transform estimating functions is currently not consistent at all. My impression is that  except of the two `cv::estimateAffine*2D` no other pair of functions have the same signature and semantics. Some functions are even outside of `calib3d` see `cv::estimateRigidTransform` (also name is incorrect). This will only be worse as people are adding new overloads to control more parameters as in #10569.

It would be nice to agree on some ultimate interface to all those RANSAC-based estimators (and group them nicely in documentation). Maybe this is something to deal with in OpenCV 4?

PR #12248, opened 2018-08-18
video: remove duplicate RANSAC code

hrnr:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1724

Proposal to remove duplicate RANSAC code in video module in OpenCV 4.

In ideal world, we would just remove `estimateRigidTransform` as the RANSAC estimators are implemented in calib3d. To support transition I have kept at least the most used overload as deprecated.

Follow up of https://github.com/opencv/opencv/pull/10569#issuecomment-400312371 

Related #11594 #10631

### DIFF
--- a/modules/shape/CMakeLists.txt
+++ b/modules/shape/CMakeLists.txt
@@ -1,2 +1,2 @@
 set(the_description "Shape descriptors and matchers")
-ocv_define_module(shape opencv_core opencv_imgproc opencv_video WRAP python)
+ocv_define_module(shape opencv_core opencv_imgproc opencv_calib3d WRAP python)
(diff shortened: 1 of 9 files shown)


alalek:
> Will C API be removed from OpenCV 4

Yes.
We have plans to drop C-API from public headers and tests at least. Refactoring of internal implementations of OpenCV functions is not an easy task (and sometimes results are not 100% compatible).

> Should I just remove cvEstimateRigidTransform

Sure.

hrnr:
I have removed `cvEstimateRigidTransform` and supressed deprecation warnings in test. All is green now.

PR #12257, opened 2018-08-20
Update hog.cpp

sturkmen72:
### This pullrequest changes
makes openCL optimisations available when the input image is three channel UMat  by converting it to gray internally.

### DIFF
--- a/modules/objdetect/src/hog.cpp
+++ b/modules/objdetect/src/hog.cpp
@@ -2011,9 +2011,13 @@ static bool ocl_detectMultiScale(InputArray _img, std::vector<Rect> &found_locat
 {
     std::vector<Rect> all_candidates;
     std::vector<Point> locations;
-    UMat image_scale;
+    UMat image_scale, gray;
+
+    if(_img.channels() > 1)
+        cvtColor(_img, gray, COLOR_BGR2GRAY);
+    else _img.copyTo(gray);
+
     Size imgSize = _img.size();
-    image_scale.create(imgSize, _img.type());
 
     for (size_t i = 0; i<level_scale.size() ; i++)
     {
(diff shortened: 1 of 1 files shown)


alalek:
Is there similar code in CPU code branch?

alalek:
Yep, it looks strange. Could you add some [simple test](https://github.com/opencv/opencv/blob/3.4.2/modules/objdetect/test/test_cascadeandhog.cpp#L1318-L1346) with color image and compare results of CPU and OpenCL code path?

sturkmen72:
i inspired by https://github.com/opencv/opencv/blob/master/samples/tapi/hog.cpp

without patch color images fps is same ( about 10 on my system ) on CPU and  openCL mode.
gray images fps about 10 on CPU and about 18 on openCL mode.

with patch color image fps is about 18 on openCL mode.

hrnr:
Wait with updating stitching until #12288 goes in. It should fix the issues with wrapping enums.

alalek:
#12288 is for "master" only. This patch is targeted to 3.4.

hrnr:
I see. I just wanted to note that there has been some issues with wrapping Stitcher::create, see #10681. Wraping this might require more work in the python generator or just wait for #12288.

sturkmen72:
please ignore the changes about stitching ( just an experiment and will be rremoved)

sturkmen72:
there are many algorithms accepts 3 ( even 4 ) channels image and convert it to gray internally. So there is no difference on results when the algorithm called with gray or color image.

(as far as i understand ) in HOG code there is no conversion to gray when the input image is 3 channels. So there is small differences on results when the input is gray or color image.

although i prepared this patch i think leaving the code as is and making explanation on documentation will be better.

vpisarev:
@sturkmen72, thank you for the patch!
Following Navneet Dalal, the algorithm author, in OpenCV HOG operates on color images. For each pixel `(x,y)` and for each channel `c` (c=R, G, B) it computes spatial x- and y- derivatives (let's call them `dx(x,y,c)` and `dy(x,y,c)` and then it chooses the channel where the gradient magnitude is maximal: `c*(x,y) = arg max_c sqrt(dx(x,y,c)**2 + dy(x,y,c)**2)` and then it uses `dx(x,y,c*(x,y))` and `dy(x,y,c*(x,y))` as the gradient values at `(x,y)`.

So, I'm afraid, your patch is not quite correct. We should not convert color image to gray, but instead the proper channel selection (for each pixel) as in the CPU branch

PR #12258, opened 2018-08-20
Trace: fixed filename slash processing

savuor:
### This pullrequest changes

Fixed filename written to `OpenCVTrace.txt` file which breaks `trace_profiler.py` script


alalek:
Rebase to 3.4

PR #12264, opened 2018-08-21
Remove one of forward methods from dnn layers

dkurt:
### This pullrequest changes

Mark forward with `Mat*` from dnn as deprecated
```cpp
virtual void forward(std::vector<Mat*> &input, std::vector<Mat> &output, std::vector<Mat> &internals)
``` 

```
force_builders=Custom
buildworker:Custom=linux-2
test_opencl:Custom=ON
```

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -567,8 +567,11 @@ struct DataLayer : public Layer
         return false;
     }
 
-    void finalize(const std::vector<Mat*>&, std::vector<Mat>& outputs) CV_OVERRIDE
+    virtual void finalize(InputArrayOfArrays, OutputArrayOfArrays outputs_arr) CV_OVERRIDE
     {
+        std::vector<Mat> outputs;
+        outputs_arr.getMatVector(outputs);
+
         CV_Assert_N(outputs.size() == scaleFactors.size(), outputs.size() == means.size(),
                   inputsData.size() == outputs.size());
         skip = true;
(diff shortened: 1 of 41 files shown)


vpisarev:
:+1:

modules/dnn/src/dnn.cpp
```diff
@@ -430,19 +430,19 @@ struct DataLayer : public Layer
...
-    void forward(InputArrayOfArrays inputs, OutputArrayOfArrays outputs, OutputArrayOfArrays internals) CV_OVERRIDE
+    void forward(InputArrayOfArrays inputs_arr, OutputArrayOfArrays outputs_arr, OutputArrayOfArrays internals_arr) CV_OVERRIDE
     {
         CV_TRACE_FUNCTION();
         CV_TRACE_ARG_VALUE(name, "name", name.c_str());
 
         CV_OCL_RUN(IS_DNN_OPENCL_TARGET(preferableTarget),
-                   forward_ocl(inputs, outputs, internals));
+                   forward_ocl(inputs_arr, outputs_arr, internals_arr) ||
+                   forward_fallback(inputs_arr, outputs_arr, internals_arr));
```
alalek: `CV_OCL_RUN()` should contain OpenCL-related calls only.
It is just a "hook" for OpenCL processing.

Why is `forward_fallback` here? What is the real purpose of "fallback" call?

dkurt: `forward_fallback` calls Layer's forward method converting FP16 mats to FP32 and back to FP16 after `forward`. For now, `forward_fallback` calls forward with `std::vector<Mat*>` but in this PR we remove it.

In this PR an every forward looks like
```cpp
void forward(InputArrayOfArrays inputs_arr, OutputArrayOfArrays outputs_arr, OutputArrayOfArrays internals_arr) CV_OVERRIDE
{
    CV_TRACE_FUNCTION();
    CV_TRACE_ARG_VALUE(name, "name", name.c_str());

    CV_OCL_RUN(IS_DNN_OPENCL_TARGET(preferableTarget) &&
               forward_ocl(inputs_arr, outputs_arr, internals_arr))

    if (inputs_arr.depth() == CV_16S)
    {
        forward_fallback(inputs_arr, outputs_arr, internals_arr);
        return;
    }

    // Get Mats from input arrays and perform default CPU implementation
    // ...
}
```

PR #12274, opened 2018-08-22
viz: call "mapper->Update()" before and after SetInputData()

### LINKED ISSUE #10945: WImage3D not visible since OpenCV 3.4.1

alalek:
resolves #10945

### DIFF
--- a/modules/viz/src/widget.cpp
+++ b/modules/viz/src/widget.cpp
@@ -301,6 +301,7 @@ void cv::viz::Widget3D::applyTransform(const Affine3d &transform)
     vtkSmartPointer<vtkPolyDataMapper> mapper = vtkPolyDataMapper::SafeDownCast(actor->GetMapper());
     CV_Assert("Widget doesn't have a polydata mapper" && mapper);
 
+    mapper->Update(); // #10945
     VtkUtils::SetInputData(mapper, VtkUtils::TransformPolydata(mapper->GetInput(), transform));
     mapper->Update();
 }


csukuangfj:
@alalek 
sorry, I only have time at this weekend and I will have a look at it then.

alalek:
:+1:

PR #12278, opened 2018-08-22
Fix bug in distanceTransform

### LINKED ISSUE #12218: cv::distanceTransform() fail on a large image
- OpenCV => 3.4.2
- Compiler => gcc 8

Run
```.cpp

int main() {
  cv::Mat non_valid = cv::imread("surfaces_nonvalid.tif", cv::IMREAD_ANYDEPTH);
  cv::Mat dst;
  cv::Mat labels;
  cv::distanceTransform(non_valid, dst, labels, cv::DIST_L2, cv::DIST_MASK_3,
                        cv::DIST_LABEL_PIXEL);
  std::cout << labels.at<int>(0, 0) << std::endl; // <-- Error here: 0
  std::cout << labels.at<int>(50, 50) << std::endl; // <-- Error here: 0
  std::cout << labels.at<int>(5000, 5000) << std::endl; // Looks reasonable
  return 0;
}

```
on `non_valid` image from [surfaces_nonvalid.zip](https://github.com/opencv/opencv/files/2290343/surfaces_nonvalid.zip). The labels output has a lot of erroneous zero label values on the left part of the output as seen on this screenshot (left image: input non_valid image, right image: output labels image. black pixels `= 0`): 
screenshot from 2018-08-15 15-11-38

take1014:
resolve #12218

### DIFF
--- a/modules/imgproc/test/test_distancetransform.cpp
+++ b/modules/imgproc/test/test_distancetransform.cpp
@@ -285,26 +285,25 @@ TEST(Imgproc_DistanceTransform, accuracy) { CV_DisTransTest test; test.safe_run(
 
 BIGDATA_TEST(Imgproc_DistanceTransform, large_image_12218)
 {
-    int i,j;
-    Size size(20000, 8000);
-    Mat src(size, CV_8UC1, Scalar::all(128));
-    Mat ans(size, CV_32FC1, Scalar::all(8192));
-    Mat dst;
-    Mat labels;
+    const int lls_maxcnt = 449970000;   // labels's maximum count
+    const int lls_mincnt = 1;           // labels's minimum count
+    int i, j, nz;
+    Mat src(30000, 30000, CV_8UC1), dst, labels;
+    for( i = 0; i < src.rows; i++ )
+        for( j = 0; j < src.cols; j++ )
+            src.data[i*src.step + j] = ( j > (src.cols / 2)) ? 0 : 255;
 
     distanceTransform(src, dst, labels, cv::DIST_L2, cv::DIST_MASK_3, DIST_LABEL_PIXEL);
 
-    int* _ans = ans.ptr<int>();
-    int* _dst = dst.ptr<int>();
-    int ansstep = (int)(ans.step/sizeof(_ans[0]));
-    int dststep = (int)(dst.step/sizeof(_dst[0]));
+    Size size = labels.size();
     for( i = 0; i < size.height; i++ )
     {
-        float* d1 = (float*)(_dst + i*dststep);
-        float* d2 = (float*)(_ans + i*ansstep);
+        int* lls = labels.ptr<int>(i);
         for( j = 0; j < size.width; j++ )
-            EXPECT_FLOAT_EQ(d1[j], d2[j]);
+            lls[j] = (lls[j] - lls_mincnt) / (lls_maxcnt / 2);
     }
+    nz = cv::countNonZero(labels);
+    EXPECT_EQ(nz, (size.height*size.width / 2));
 }
 
 }} // namespace
--- a/modules/imgproc/test/test_distancetransform.cpp
+++ b/modules/imgproc/test/test_distancetransform.cpp
@@ -285,10 +285,10 @@ TEST(Imgproc_DistanceTransform, accuracy) { CV_DisTransTest test; test.safe_run(
 
 BIGDATA_TEST(Imgproc_DistanceTransform, large_image_12218)
 {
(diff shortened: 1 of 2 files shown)


modules/imgproc/test/test_distancetransform.cpp
```diff
@@ -283,4 +283,28 @@ void CV_DisTransTest::prepare_to_validation( int /*test_case_idx*/ )
...
+    distanceTransform(src, dst, labels, cv::DIST_L2, cv::DIST_MASK_3, DIST_LABEL_PIXEL);
+
+    int* _ans = ans.ptr<int>();
+    int* _dst = dst.ptr<int>();
+    int ansstep = (int)(ans.step/sizeof(_ans[0]));
+    int dststep = (int)(dst.step/sizeof(_dst[0]));
+    for( i = 0; i < size.height; i++ )
+    {
+        float* d1 = (float*)(_dst + i*dststep);
+        float* d2 = (float*)(_ans + i*ansstep);
```
alalek: Don't use manual pointer arithmetic.

At least there is mess of int / float types:
> _dst = dst.ptr<int>
> (float*)(_dst + i*dststep)

These "row" pointer should be received like this:
```
const float* d1 = dst.ptr<const float>(i);
const float* d2 = ans.ptr<const float>(i);
```

Anyway, this heavy loop is just a duplicate of existed norm function. So while check can be replaced to this:
```
EXPECT_LE(cvtest::norm(dst, ans, NORM_INF), 1e-3);
```

modules/imgproc/test/test_distancetransform.cpp
```diff
@@ -283,4 +283,28 @@ void CV_DisTransTest::prepare_to_validation( int /*test_case_idx*/ )
 
 TEST(Imgproc_DistanceTransform, accuracy) { CV_DisTransTest test; test.safe_run(); }
 
+BIGDATA_TEST(Imgproc_DistanceTransform, large_image_12218)
+{
+    int i,j;
+    Size size(20000, 8000);
+    Mat src(size, CV_8UC1, Scalar::all(128));
+    Mat ans(size, CV_32FC1, Scalar::all(8192));
```
alalek: > 8192

Magic number from algorithm implementation?

take1014: @alalek Thanks for your comments. This test is based on "testDistanceTransformWithLabels ()". I created this test with reference to imgproc / misc / java / test / ImgprocTest.java.

take1014: @alalek This test case is not suitable,  so I will fix it later.

vpisarev:
@take1014, thank you for the patch! let's revert the latest commit. 30000x30000 is way too much, even for 64-bit systems. Please, use a smaller image in the test

modules/imgproc/test/test_distancetransform.cpp
```diff
@@ -283,4 +283,27 @@ void CV_DisTransTest::prepare_to_validation( int /*test_case_idx*/ )
...
 
+BIGDATA_TEST(Imgproc_DistanceTransform, large_image_12218)
+{
+    const int lls_maxcnt = 79992000;   // labels's maximum count
+    const int lls_mincnt = 1;           // labels's minimum count
+    int i, j, nz;
+    Mat src(8000, 20000, CV_8UC1), dst, labels;
+    for( i = 0; i < src.rows; i++ )
+        for( j = 0; j < src.cols; j++ )
+            src.data[i*src.step + j] = ( j > (src.cols / 2)) ? 0 : 255;
```
alalek: Avoid address arithmetic. Use something like this:
```
src.at<uchar>(i, j) = ...
```

modules/imgproc/test/test_distancetransform.cpp
```diff
@@ -283,4 +283,27 @@ void CV_DisTransTest::prepare_to_validation( int /*test_case_idx*/ )
...
+
+    distanceTransform(src, dst, labels, cv::DIST_L2, cv::DIST_MASK_3, DIST_LABEL_PIXEL);
+
+    Size size = labels.size();
+    for( i = 0; i < size.height; i++ )
+    {
+        int* lls = labels.ptr<int>(i);
+        for( j = 0; j < size.width; j++ )
+            lls[j] = (lls[j] - lls_mincnt) / (lls_maxcnt / 2);
```
alalek: Element-wise `B = A * alpha + beta` operation can be performed via `cv::convertTo()` call.

vpisarev:
looks good to me, let's merge it in! :+1:

PR #12279, opened 2018-08-22
BGR2Gray 8u: IPP conformance

savuor:
### This pullrequest changes

BGR2Gray for 8u fixed to perform the same as IPP's code including nearest-even rounding


alalek:
Currently RGB -> GRAY result is equal to "Y" channel of RGB -> YCbCr / YUV conversions (luma component).

Not sure that we should introduce difference here.

savuor:
Indeed, need to fix RGB2Y** conversions
Looks like it's needed to fix RGB2Gray conversions for 16u as well (which may include rewriting them to new universal intrinsics)

vpisarev:
thanks! :+1:

PR #8736, opened 2017-05-15
cmake: enable c++ 11 by default

sovrasov:
**WIP**
Test

PR #12293, opened 2018-08-23
core: cleanup stl string replacement (drop implementation of cv::String)

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1739

`cv::String` is mapped to `std::string` directly.

```
typedef std::string String;
```

replaces #8736

```
allow_multiple_commits=1
```


vpisarev:
looks great!

PR #12288, opened 2018-08-23
int to proper enum type

cv3d:
TESTING... DO NOT MERGE!
<s>Merge with opencv/opencv_contrib#1730 and opencv/opencv_extra#515</s>

### This pullrequest changes
There are hundreds of unnamed enums, and they are being passed as int in the C++ interface.
In order to avoid passing a wrong value, and to maintain good practices, the hope is to name most of these enums, and change the int arguments into the proper enum types wherever possible.

### Newly introduced enums
- [x] `ElemType` for handling the depth and channels of matrix elements (Produced by `DataType<T>::type` and `traits::Type<T>::value`)
- [x] `ElemDepth` for handling the depth of matrix elements (Produced by `DataDepth<T>::type` and `traits::Depth<T>::value`)
- [x] `MagicFlag` for handling the mixed magic values of `ElemType`, `AccessFlag` and `_InputArray::KindFlag`

### Flag enums
- [x] `AccessFlag` for `unnamed` enum having `ACCESS_READ` as first member
- [x] [**Converted DrawMatchesFlags struct to enum struct**] DrawMatchesFlags::`unnamed` enum having `DEFAULT` as first member
- [x] `MemoryFlag` for UMatData::`unnamed` enum having `COPY_ON_MAP` as first member
- [x] `KindFlag` for _InputArray::`unnamed` enum having `KIND_SHIFT` as first member

### Mode enums
- [x] `DescriptorType` for AKAZE::`unnamed` enum having `DESCRIPTOR_KAZE_UPRIGHT` as first member
- [x] `DetectorType` for AgastFeatureDetector::`unnamed` enum having `AGAST_5_8` as first member
- [x] `FeatureType` for CvFeatureParams::`unnamed` enum having `HAAR` as first member
- [x] `DetectorType` for FastFeatureDetector::`unnamed` enum having `TYPE_5_8` as first member
- [x] `MatcherType` for DescriptorMatcher::`unnamed` enum having `FLANNBASED` as first member
- [x] `FormatType` for Formatter::`unnamed` enum having `FMT_DEFAULT` as first member
- [x] `HistogramNormType` for HOGDescriptor::`unnamed` enum having `L2Hys` as the only member
- [x] [**Converted to static const NormTypes**] for Hamming::`unnamed` enum having `normType` as the only member
- [x] [**Converted to static const NormTypes**] for SL2::`unnamed` enum having `normType` as the only member
[...]

PR #12303, opened 2018-08-24
Binding-generator improvements

cv3d:
With #12288, binding fails to cast named enums properly, and the C++11 enums are not recognized.
This PR fixes such issues, promoting a smooth transition to C++11 with proper enum interface (rather than just `int`s).

### This pullrequest changes
- Supports parsing `enum class` and `enum struct` with the correct scope
- Improves enum and consts hierarchical parsing
- Support enum arguments in Java and Python binders

<cut/>

```
allow_multiple_commits=1
docker_image:Docs=docs-js
```

### DIFF
--- a/modules/java/generator/gen_java.py
+++ b/modules/java/generator/gen_java.py
@@ -425,6 +425,10 @@ class JavaWrapperGenerator(object):
     def add_enum(self, decl): # [ "enum cname", "", [], [] ]
         enumname = decl[0].replace("enum ", "").strip()
         self.enum_types.append(enumname)
+        const_decls = decl[3]
+
+        for decl in const_decls:
+            self.add_const(decl)
 
     def add_func(self, decl):
         fi = FuncInfo(decl, namespaces=self.namespaces)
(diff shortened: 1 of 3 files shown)


cv3d:
I really wish this gets merged soon to keep #12288 clean.

@alalek @mshabunin Could one of you kindly review it for me?

cv3d:
@alalek Can you please consider this PR?

cv3d:
@alalek Please consider merging this PR.

cv3d:
> There are many functions/methods which accept references as parameters.
They expects the same object, not a new copied instance.

@alalek To faster grasp the intention here, could you please provide me with a sample diff fragment which captures the required change in the generated C++ code?

alalek:
[Here](https://github.com/opencv/opencv/pull/12234#discussion_r214004363). Calling of object copy constructor should be avoided.

Also references are often used as **output** parameter.
It is necessary to mark them with `CV_OUT` for bindings generators (if it is real output parameter).

alalek:
"References" problem requires some design work. Currently it doesn't cause real errors with "smart ptr" objects or with objects which are not stored in Python directly (like `std::vector<int>`).

PR #12305, opened 2018-08-25
Dnn example in google play

cabelo:


berak:
sorry for being negative here, but--

it seems, it seems, the main intent of this pr is boosting your own website.

cabelo:
Ok...

dkurt:
@cabelo, tutorials are used to describe novel algorithms and samples but not to demonstrate applications which source code is hidden from view. Let me close it for now.

PR #12341, opened 2018-08-29
countNonZero function reworked to use wide universal intrinsics

terfendail:
### This pullrequest changes
Replace custom SSE2 and NEON countNonZero implementatons with uiniversal intrinsics

TODO:
- [ ] performance results: SSE/AVX2 before and after this patch

```
force_builders=Custom,armv8
```


vpisarev:
looks great! :+1:

PR #12344, opened 2018-08-30
Replaced core module calls to universal intrinsics to wide version

terfendail:
### This pullrequest changes
Replaced core module calls to universal intrinsics to wide version


vpisarev:
@terfendail, thank you! It interferes with some other patches, like #12064 and my WIP patch with convert and convert+scale that I did not submit yet (but I'm adding `CV_16F` support there). So, could you please exclude `arithm*` and `convert*` parts for now?

vpisarev:
:+1:

PR #12346, opened 2018-08-30
Added option to fail on missing testdata

mshabunin:
### This pullrequest changes

Option `test_data_check` can be used to verify that no test cases have been skipped due to missing test data.

Example run without downloaded DL models:
```.sh
$ ./bin/opencv_test_dnn
...
[  PASSED  ] 1703 tests.
```
Same but with the new option enabled:
```.sh
$ ./bin/opencv_test_dnn  --test_require_data=true
...
[  PASSED  ] 1590 tests.                                                                                                  
[  FAILED  ] 113 tests, listed below: 
...
113 FAILED TESTS
```

### DIFF
--- a/modules/ts/src/ts.cpp
+++ b/modules/ts/src/ts.cpp
@@ -721,6 +721,7 @@ void checkIppStatus()
     }
 }
 
+static bool checkTestData = false;
 bool skipUnstableTests = false;
 bool runBigDataTests = false;
 int testThreads = 0;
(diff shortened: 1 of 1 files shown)


modules/ts/src/ts.cpp
```diff
@@ -733,6 +734,7 @@ void parseCustomOptions(int argc, char **argv)
         "{ test_threads       |-1       |the number of worker threads, if parallel execution is enabled}"
         "{ skip_unstable      |false    |skip unstable tests }"
         "{ test_bigdata       |false    |run BigData tests (>=2Gb) }"
+        "{ test_data_check    |false    |fail on missing non-required test data instead of skip}"
```
alalek: test_require_data

modules/ts/src/ts.cpp
```diff
@@ -723,6 +723,7 @@ void checkIppStatus()
 
 bool skipUnstableTests = false;
 bool runBigDataTests = false;
+bool checkTestData = false;
```
alalek: It is not defined in headers, so use "static".

Environment parameter is easy to integrate into build system, so this would nice to have:
```
static bool g_requireTestData = utils::getConfigurationParameterBool("OPENCV_TEST_REQUIRE_DATA", false);

...

if (parser.get<bool>("test_data_check"))
    g_requireTestData = true;
```

PR #12348, opened 2018-08-30
Fix bug in Subdiv2D::getTriangleList

### LINKED ISSUE #5788: Delaunay triangulation returns bad points
class Subdiv2D implements delaunay triangulation. 
One of constructors take a rect. Then you need to insert points. Then getTriangleList returns points of trinagles to you. But. Some points is not in the rect area. For example. An image 1280x1024. But you can take a point like this (-1117, 85). How can it be? I give a rect with (0, 0, width, heigh).
And. I have 2 images with faces. And each face have 69 points on it.
When I use triangulation, I take different count of trinagles of each face. Why? 
Bug maybe?

take1014:
resolve #5788

### DIFF
--- a/modules/imgproc/src/subdivision2d.cpp
+++ b/modules/imgproc/src/subdivision2d.cpp
@@ -758,6 +758,7 @@ void Subdiv2D::getTriangleList(std::vector<Vec6f>& triangleList) const
     triangleList.clear();
     int i, total = (int)(qedges.size()*4);
     std::vector<bool> edgemask(total, false);
+    Rect2f rect(topLeft.x, topLeft.y, bottomRight.x, bottomRight.y);
 
     for( i = 4; i < total; i += 2 )
     {
@@ -773,7 +774,8 @@ void Subdiv2D::getTriangleList(std::vector<Vec6f>& triangleList) const
         edge = getEdge(edge, NEXT_AROUND_LEFT);
         edgeOrg(edge, &c);
         edgemask[edge] = true;
-        triangleList.push_back(Vec6f(a.x, a.y, b.x, b.y, c.x, c.y));
+        if( rect.contains(a) && rect.contains(b) && rect.contains(c) )
+            triangleList.push_back(Vec6f(a.x, a.y, b.x, b.y, c.x, c.y));
     }
 }
 
--- a/modules/imgproc/test/test_subdivision2d.cpp
+++ b/modules/imgproc/test/test_subdivision2d.cpp
@@ -1,0 +1,53 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//M*/
+#include "test_precomp.hpp"
+
+namespace opencv_test { namespace {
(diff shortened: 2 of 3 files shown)


modules/imgproc/test/test_subdivision2d.cpp
```diff
@@ -0,0 +1,114 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
alalek: Please use short license header:
```
```

modules/imgproc/test/test_subdivision2d.cpp
```diff
@@ -0,0 +1,114 @@
...
+// or tort (including negligence or otherwise) arising in any way out of
+// the use of this software, even if advised of the possibility of such damage.
+//
+//M*/
+#include "test_precomp.hpp"
+
+namespace opencv_test { namespace {
+class CV_ImgprocSubdiv2D_getTriagnleListTest : public cvtest::BaseTest
+{
+public:
```
alalek: It is not necessary to use Legacy "BaseTest" class:
```
TEST(Imgproc_Subdiv2D_getTriangleList, regression_5788)
{
    ... put all test code here ...
}
```

modules/imgproc/test/test_subdivision2d.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    createTriangleTestData(pts);
+
+    subdiv.insert(pts);
+
+    std::vector<cv::Vec6f> triangles;
+    subdiv.getTriangleList(triangles);
+
+    int trig_cnt = 0;
+    for( auto it : triangles )
```
alalek: Use C++98

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -835,7 +835,7 @@ class CV_EXPORTS GeneralizedHoughBallard : public GeneralizedHough
     virtual int getVotesThreshold() const = 0;
 };
 
-//! Guil, N., González-Linares, J.M. and Zapata, E.L. (1999). Bidimensional shape detection using an invariant approach. Pattern Recognition 32 (6): 1025-1038.
+//! Guil, N., Gonzalez-Linares, J.M. and Zapata, E.L. (1999). Bidimensional shape detection using an invariant approach. Pattern Recognition 32 (6): 1025-1038.
 //! Detects position, translation and rotation
```
alalek: It would be nice to add reference in doc/opencv.bib file and use it here via `@cite NGuil1999`.

modules/imgproc/src/subdivision2d.cpp
```diff
@@ -852,6 +853,11 @@ void Subdiv2D::checkSubdiv() const
     }
 }
 
+bool Subdiv2D::isWithinRect(Point2f pt) const
+{
+    return (topLeft.x <= pt.x && pt.x <= bottomRight.x) && (topLeft.y <= pt.y && pt.y <= bottomRight.y);
```
alalek: Try to reuse existed functionality: `Point2f::inside()` or `Rect::contains()`

Note: right/bottom side of OpenCV rectangles is "open" (exclusive).

take1014:
@alalek  Thanks for your comment. I also modified the Java test. The Java test was incorrect and the coordinates of the outputted point were regarded as valid points even if it was outside the rectangle.

vpisarev:
thank you! :+1:

PR #12353, opened 2018-08-30
imgcodecs(webp): multiple fixes

alalek:
- don't reallocate passed 'img' (test is fixed - must use `IMREAD_UNCHANGED` / `IMREAD_ANYCOLOR`)
- avoid memory DDOS
- avoid reading of whole file during header processing
- avoid data access after allocated buffer during header processing (missing checks)
- use [WebPFree()](https://github.com/webmproject/libwebp/commit/822f113ebbe8810a1de0f11062faaf7edb0556e3) to free allocated buffers
- drop unused & undefined `.close()` method
- added checks for channels >= 5 in encoder

```
docker_image:Custom=fedora:28
```

### DIFF
--- a/modules/imgcodecs/src/grfmt_webp.cpp
+++ b/modules/imgcodecs/src/grfmt_webp.cpp
@@ -54,15 +54,21 @@
 
 #include "opencv2/imgproc.hpp"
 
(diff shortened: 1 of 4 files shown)


vpisarev:
:+1:

PR #12355, opened 2018-08-31
Update samples

sturkmen72:
### This pullrequest changes
some improvements on samples and documentation


sturkmen72:
ready to review.

both of directx_d3d11_interop.cpp and directx_d3d10_interop.cpp seems problematic about showed process time but i can't find a way to solve using cv:putText() maybe using print_info() as used in other samples will be better.

in directx_d3d11_interop.cpp when i press 3 my computer freezes ( with or without this patch)

PR #12357, opened 2018-08-31
fix some static analyzer warnings

cyyever:
### This pullrequest changes

some static analyzer warings.

```
docker_image:Docs=docs-js
```


alalek:
Thank you for contribution!

Please eliminate "merge" commit from the patch (rebase and squash commits).

It would be nice to have similar patch into 3.4 (current maintenance branch, but it is C++98).

cyyever:
I have fixed the commits,it is hard to change c++98 code without use of std::atomic.

PR #12364, opened 2018-08-31
Change default value of crop argument of blobFromImage

dkurt:
### This pullrequest changes

Replace default value of `crop` argument from true to false.
Replace default value of `swapRB` argument from true to false.

With this PR `blobFromImage` and `blobFromImages` just create a 4D blob from an image by default:

```python
# considering that img is a multichannel image with interleaved channels.
blob = img.transpose(2, 0, 1).reshape(1, img.shape[2], img.shape[0], img.shape[1])
```

Impacted applications:
* By default, Caffe framework works with BGR images ([source](https://github.com/BVLC/caffe/wiki/Image-Format:-BGR-not-RGB)) so `blobFromImage` without explicit `swapRB` flag for images from `cv::imread` now matches Caffe's models inputs. On the other hand, TensorFlow models work mostly with RGB images so we need to specify `swapRB=true` now.
* Without `crop=false` for object detection samples, `blobFromImage` cropped an input image keeping aspect ratios which gave wrong detections close to image borders (for object detection models from Caffe and TensorFlow).

```
force_builders=Custom,ocllinux
buildworker:Custom=linux-2
test_opencl:Custom=ON
```

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -46,9 +46,9 @@
 #include <opencv2/core.hpp>
 
 #if !defined CV_DOXYGEN && !defined CV_DNN_DONT_ADD_EXPERIMENTAL_NS
-#define CV__DNN_EXPERIMENTAL_NS_BEGIN namespace experimental_dnn_34_v8 {
+#define CV__DNN_EXPERIMENTAL_NS_BEGIN namespace experimental_dnn_34_v9 {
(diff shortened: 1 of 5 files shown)


vpisarev:
I'd change 3.4 as well, because it would be a very weird if 3.4 API is identical to 4.0 but the default parameter value is different.

alalek:
@dkurt Do we need some "version" definition for this change? (however DNN API is "experimental" in 3.x).

dkurt:
@alalek, added notes about most common applications which are affected by these changes.

vpisarev:
:+1:

PR #12372, opened 2018-09-01
Fix a bug in generating phantom non-static methods that does not have args

cv3d:
Related to #12180
### This pullrequest changes
Currently, phantom non-static methods which have no args have the signature:
```C++
    if(PyObject_Size(args) == 0 && (!kw || PyObject_Size(kw) == 0))
    {
        ERRWRAP2(retval = cv_class_phatom_method(_self_, args));
        return pyopencv_from(retval);
    }

    return NULL;
```
But the check `PyObject_Size(args) == 0 && (!kw || PyObject_Size(kw) == 0` is problematic, and it can return `NULL` in some platforms (Python 2 only?)

```
**WIP**
test_modules=python2,python3
force_builders=Custom
buildworker:Custom=linux-3
docker_image:Custom=ubuntu:16.04
test_opencl:Custom=ON
```


cv3d:
@alalek I still wonder whether we should really allow the behavior `cv.class.phatom_method(instance)` along with the standard member call `instance.phatom_method()`?
If we want to allow it, then it should be there for ALL methods, not just phantom ones.
But if we do not want to allow it, then let us just modify the unit test to use the standard member calling, and that is it.

cv3d:
@alalek I modified the unit test accordingly, with additional checks to make it more of a real unit test.

Issue #12337, opened 2018-08-29
solveLP() doesn't check the argument type


barisdemiroz:
##### System information (version)

- OpenCV => 3.0
- Operating System / Platform => Ubuntu 16.04 64-bit
- Compiler => gcc 4.9.4

##### Detailed description

In the documentation it says that `solveLP()` accepts only 64-bit floating point matrix for the last argument. But easy to overlook that and pass a 32-bit floating point matrix. There should be a precondition assert for that like in other functions. 

##### Steps to reproduce

```.cpp
Mat1f c,A,x;
// ... fill c and A
solveLP(c, A, x);
for (int i=0; i<x.rows; ++i)
    cout << x(i) << ", "; // observe wrong values
```

alalek:
> 3.0

This version is too old.
Please try latest code from 3.4 branch.

barisdemiroz:
`modules/core/src/lpsolver.cpp` is modified in 2014 (if you don't count the latest minor modification). If you look at [the implementation](https://github.com/opencv/opencv/blob/c1db75e0c7b1584b555242ec53fbef5f71d0c48f/modules/core/src/lpsolver.cpp#L93) you can see there is no assertion for the type of `z`.

But inspecting the code made me realize that `z.create()` is called and according to [the definition](https://docs.opencv.org/3.1.0/d3/d63/classcv_1_1Mat.html#a55ced2c8d844d683ea9a725c60037ad0) it dereferences the old matrix if the type is not compatible. So everything checks out in the documentation, it is safe to close this bug. But I still wish I would get an error  when it tries to change the underlying type of a `Mat1f`. This behavior is not very intuitive.

alalek:
This can't be fixed in the current interface with direct `cv::Mat` types.

"FixedType" attribute is available in `InputArray` only.

I added some checks to forbid Mat1f. Please take a look.
But they are not reliable (empty uninitialized `cv::Mat` has `CV_8UC1` type).

barisdemiroz:
Thank you very much. I wasn't expecting this to be fixed since it was a very fundamental problem. I am very happy to see the PR.

PR #12376, opened 2018-09-01
core: add solveLP type checks for output

alalek:
to forbid `Mat1f`

Checks are not reliable: empty uninitialized `cv::Mat` has `CV_8UC1` type

related #12337


vpisarev:
:+1:

PR #12377, opened 2018-09-02
core: wrap custom types via _RawArray (raw() call)

alalek:
- support passing of `std::vector<KeyPoint>` via InputArray

related #9457

```
allow_multiple_commits=1
force_builders=Custom
```


vpisarev:
great idea! I wonder if we can eliminate _RawArray type completely and instead just add the functions

<pre>
_InputArray rawIn(const std::vector&lt;...&gt;& arr);
_OutputArray rawOut(const std::vector&lt;...&gt;& arr);
</pre>

?

vpisarev:
looks great! :+1:

PR #12386, opened 2018-09-03
dnn: Remove a duplicated code snippet for flatten layer

wzw-intel:


alalek:
Please avoid re-creation of PRs. Submit updates inplace (perhaps push changes with "force" flag).

wzw-intel:
@alalek Ok, got it.

PR #12398, opened 2018-09-03
samples: eliminate cvconfig.h usage

alalek:
`cvconfig.h` is not a part of OpenCV public API.

```
#with_tbb=ON
```


vpisarev:
:+1:

Issue #11000, opened 2018-03-06
OE-1 "Maintaining OpenCV 1.x C API" discussion


vpisarev:
the feature request about evolution proposal [OE-1](https://github.com/opencv/opencv/wiki/OE-1.-Old-C-API)

kinchungwong:
/cc @alalek Re: https://github.com/opencv/opencv/issues/10246#issuecomment-350116500

> Possible solution here is FFI (https://en.wikipedia.org/wiki/Foreign_function_interface) bindings for C which should be able to emulate C++ features effectively.
> To implement this we must think about emulation of C++ exceptions in C, classes/methods/objects, etc, including some "runtime library" to support this.
> If you want strategic/political answer then you should ask

/cc @dkogan Re: https://github.com/opencv/opencv/issues/10246#issuecomment-350117100

>  Can you please update the docs, and at least the front webpage to not claim there's a C API anymore?

Dear Vadim, 

Please clarify whether a C-compatible FFI (foreign function interface) is within the scope of OE-1 or outside. Whether it is or not in OE-1, a C-compatible FFI needs to be on Evolution Proposals, because many open-source projects require the functionality of OpenCV. This issue also affects the relationship between OpenCV and the larger open-source community.

Some of them require OpenCV 1.x, but some of them may also require newer functionalities from 2.x or 3.x.

There may also be a need to ***either*** section off parts of [```core/cvdef.h```](https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/cvdef.h) for C compilers, ***or*** to poison off the entire file for C compilers (so that it's clear it won't build).

alalek:
This OE is about "OpenCV 1.x API" and about using of this API in other OpenCV parts.

C-compatible FFI is not covered here. Perhaps it would be another C-like API - at least internally (but compatibility headers/wrappers may be proposed).

sturkmen72:
i think functions like [cvCanny](https://github.com/opencv/opencv/blob/master/modules/imgproc/src/canny.cpp#L1124-L1132) can be cleaned up. i can work on this

PR #12401, opened 2018-09-03
cmake: update install paths (Linux)

alalek:
Improve support for side-by-side installations (multiple major versions).
Improve conformance with Linux/GNU installation layout.

Details:
- includes have one extra level: `/usr/include/opencv4/` (no changes in applications are required if CMake is used).
- CMake config is moved into "lib". It is incorrect to place it in "share" directory due direct paths to binaries.
- changed paths for "doc", "licenses", "java"
- use "opencv4" by default (except "lib" - no "opencv" in path)
- pkg-config file is renamed to `opencv4.pc`
- use lower-case "opencv" in paths

All paths are still configurable.

```
**WIP**
force_builders=linux,docs
test_modules=core
build_contrib:iOS=ON
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -373,8 +373,6 @@ endif()
 
 if(ANDROID OR WIN32)
   ocv_update(OPENCV_DOC_INSTALL_PATH doc)
-else()
-  ocv_update(OPENCV_DOC_INSTALL_PATH share/OpenCV/doc)
 endif()
 
 if(WIN32 AND CMAKE_HOST_SYSTEM_NAME MATCHES Windows)
@@ -414,8 +412,6 @@ if(ANDROID)
   ocv_update(OPENCV_TEST_DATA_INSTALL_PATH "sdk/etc/testdata")
 elseif(WIN32)
   ocv_update(OPENCV_TEST_DATA_INSTALL_PATH "testdata")
-else()
-  ocv_update(OPENCV_TEST_DATA_INSTALL_PATH "share/OpenCV/testdata")
 endif()
 
 if(ANDROID)
@@ -428,6 +424,7 @@ if(ANDROID)
   ocv_update(OPENCV_INCLUDE_INSTALL_PATH sdk/native/jni/include)
   ocv_update(OPENCV_SAMPLES_SRC_INSTALL_PATH samples/native)
   ocv_update(OPENCV_OTHER_INSTALL_PATH   sdk/etc)
+  ocv_update(OPENCV_LICENSES_INSTALL_PATH "${OPENCV_OTHER_INSTALL_PATH}/licenses")
 else()
   set(LIBRARY_OUTPUT_PATH                "${OpenCV_BINARY_DIR}/lib")
   ocv_update(3P_LIBRARY_OUTPUT_PATH      "${OpenCV_BINARY_DIR}/3rdparty/lib${LIB_SUFFIX}")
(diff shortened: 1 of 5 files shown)


mshabunin:
Maybe we should break `#include` compatibility. It does not look good: `<ROOT>/include/opencv4/opencv2/core.hpp`

alalek:
"opencv4" will not be used in source codes. Users still using "opencv2/core.hpp" (no any changes are required if using CMake)

> opencv4/opencv2/core.hpp

Right.
Currently I'm working on "proxy" headers in "opencv" directory (include/opencv4/opencv/core.hpp).
Currently they would just include "../opencv2/core.hpp" (relative path is required, just "opencv2/core.hpp" doesn't work due installations of OpenCV with other versions 2.4/3.x into system paths).

This will allow smooth transition from "opencv2". Both lines would have the same effect:
```
#include <opencv2/core.hpp>
#include <opencv/core.hpp>
```

Unarmed1000:
Is the impact of this change documented anywhere for code not using cmake?  
How can it be disabled so the headers install to their normal old location?

Is the recommended solution to do a  "ln -s /usr/local/include/opencv4/opencv2/ /usr/local/include/opencv2"?

We normally create docker images for building inside so being able to have  side-by-side installations is not something that we want to take advantage of and it appears to break all existing software that doesnt use cmake for building as the include path changed and there is no desire for us to update all our build files with a new include path for no good reason.

I tried looking at the documentation but did not find the 'breaking' change mentioned anywhere but maybe I missed it. I found multiple people that resorted to the "ln" solution which is why I wanted to ask if its the recommended way to deal with this.

alalek:
Recommended way is using of CMake's `find_package(OpenCV ...)`.

Unarmed1000:
I asked what is the recommended way for people **not** using cmake.

sergiomb2:
this commit https://github.com/opencv/opencv/commit/a95673287433fc810eda2d88b94bb234298c4cd5
for me is a disaster , why opencv.pc change to opencv4.pc ?

mshabunin:
@sergiomb2 , set the OPENCV_PC_FILE_NAME option to whatever you need.

sergiomb2:
thank you for reply mshabunin 

after move opencv4.pc to opencv.pc if not packages aren't found .
gmic don't compile (with opencv-4.1) [1] with :
CImg.h:424:10: fatal error: opencv2/opencv.hpp: No such file or directory

nomacs don't complie [3] with error: 'CV_DIST_C' was not declared in this scope , was C API deleted ?  
https://docs.opencv.org/4.1.0/df/d4e/group__imgproc__c.html is a clean page 

Thanks I have to go now 


[1]
https://copr-be.cloud.fedoraproject.org/results/sergiomb/opencv/fedora-28-x86_64/00912676-gmic/build.log.gz
https://copr-be.cloud.fedoraproject.org/results/sergiomb/opencv/fedora-28-x86_64/00912676-gmic/

[3]
https://copr-be.cloud.fedoraproject.org/results/sergiomb/opencv/fedora-rawhide-x86_64/00912674-nomacs/build.log.gz
https://copr-be.cloud.fedoraproject.org/results/sergiomb/opencv/fedora-28-x86_64/00912676-gmic/

mshabunin:
C-API was not deleted yet, but all inclusions have been removed from default headers, so now one needs to directly include "opencv2/imgproc/types_c.h" and other `*_c.h` headers:
https://github.com/opencv/opencv/blob/de977cc9c836dd954bc0a737b5aa80621a4f9c72/modules/imgproc/include/opencv2/imgproc/types_c.h#L593

Not sure why opencv.pc was not found by

alalek:
C API (from OpenCV 1.0) has been deprecated since OpenCV 2.4. It is not tested for years with C compiler.

Mentioned headers keep legacy C API constants only (for apps which are not migrated on C++ enums for some reasons).

Some part of legacy C API functions are gone (left functions/structures will be removed without notice during future OpenCV 4.x releases).

More info/links about C API is here: #11000

---

> opencv2/opencv.hpp

This all-in-one header is deprecated.
Consider using headers of OpenCV modules directly:
```
#include <opencv2/core.hpp>
```

PR #12424, opened 2018-09-05
build: don't use nullptr without C++11 guard

alalek:
C++98 is used in OpenCV 3.x.

```
```


alalek:
:+1:

PR #12431, opened 2018-09-05
fix rect() and edgemask treatment in SubDiv2D

take1014:
related #12348


alalek:
Almost right, but unchanged edgemask and resetted to false are different things actually.

See #12433 (with corresponding contrib PR which uses `SubDiv2D`)

take1014:
@alalek  Thanks for your comments. I agree with you, and close my pull request.

PR #12439, opened 2018-09-06
Use hal::normL2Sqr_ in base.hpp's templated normL2Sqr

woodychow:
On my computer with gcc 6.3, hal::normL2Sqr_ with AVX2 is much faster than the unrolled `_AccTp normL2Sqr(const _Tp* a, const _Tp* b, int n)` and the naive loop.

I am not sure if importing hal in base.hpp follows the design. If there's a better way, please let me know.


woodychow:
This branch should be based on 3.4. Will reopen soon

alalek:
Changing branch doesn't require to re-open PR. Make all changes "inplace".

PR #12440, opened 2018-09-06
Replace naive normL2Sqr with hal::normL2Sqr_ for better performance

woodychow:
On my computer with gcc 6.3, hal::normL2Sqr_ with AVX2 is much faster than the unrolled _AccTp normL2Sqr(const _Tp* a, const _Tp* b, int n) and the naive loop.

I am not sure if importing hal in base.hpp follows the design. If there's a better way, please let me know.

### DIFF
--- a/modules/core/src/batch_distance.cpp
+++ b/modules/core/src/batch_distance.cpp
@@ -5,6 +5,7 @@
 
 #include "precomp.hpp"
 #include "stat.hpp"
+#include <opencv2/core/hal/hal.hpp>
 
 namespace cv
 {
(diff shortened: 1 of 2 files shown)


alalek:
File "base.hpp" is part of OpenCV public API which is used in users code.
"hal.hpp" is private / internal API which doesn't work properly / effectively without special defines. These defines are missing during build of user applications.

Breaking API of existed public functions must be avoided.

Source code of OpenCV modules can use "hal::" routines.
So, my suggestion is to replace calls in kmeans.cpp to `hal::` calls directly (without touching of headers).

To check other similar places in OpenCV which uses "public" slow version, you can put public "normL2Sqr" under `#ifndef __OPENCV_BUILD` guard (disable this call in builds of OpenCV modules).

woodychow:
Thanks for your comment. I will update the branch soon.

woodychow:
Thanks!

PR #12450, opened 2018-09-06
Update HOGDescriptor

sturkmen72:
### This pullrequest changes
a crude attempt to replace Mat to InputArray

let me work on it deeper if the idea is correct.


modules/objdetect/src/hog.cpp
```diff
@@ -3665,7 +3671,7 @@ void HOGDescriptor::detectMultiScaleROI(const cv::Mat& img,
     Mutex mtx;
 
     parallel_for_(Range(0, (int)locations.size()),
-                  HOGConfInvoker(this, img, hitThreshold, Size(8, 8),
+                  HOGConfInvoker(this, img.getMat(), hitThreshold, Size(8, 8),
```
alalek: `.getMat()` returns temporary object which is passed by reference.
Usually it is stored by processing object as a reference too, but these temporary objects got destroyed.
Hopefully, this implementation is "not effective" and store object instead of const reference for source image - but it can be refactored and this code will become broken.

I would be better to avoid such usage and use local variable to store `cv::Mat`:
```
Mat img_ = img.getMat();
parallel_for_(...)
```

sturkmen72:
@alalek thank you for reviewing. let me work on this a bit more.

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -482,8 +482,8 @@ struct CV_EXPORTS_W HOGDescriptor
 
     /**@example samples/cpp/train_HOG.cpp
     */
-    /** @brief Computes HOG descriptors of given image.
-    @param img Matrix of the type CV_8U containing an image where HOG features will be calculated.
+    /** @brief Computes HOG features of given image.
```
sturkmen72: @alalek i am not sure about this change.
ready to review except it

Issue #7013, opened 2016-07-30
Set CAP_PROP_FOURCC does not work under Windows for USB camera


Lukse:
##### System information (version)
- OpenCV => 3.1.0
- Operating System / Platform => Windows 7 64 Bit
- Compiler => Python
##### Detailed description

Trying to switch camera output from YUY2 to MJPG but it does not work. MJPG mode is working in other software packages.
##### Steps to reproduce

> import cv2
> import sys
> import numpy as np
> import time
> 
> fourcc_names = ['CAP_PROP_POS_MSEC', 'CAP_PROP_POS_FRAMES', 'CAP_PROP_POS_AVI_RATIO', 'CAP_PROP_FRAME_WIDTH', 'CAP_PROP_FRAME_HEIGHT', 'CAP_PROP_FPS', 'CAP_PROP_FOURCC', 'CAP_PROP_FRAME_COUNT', 'CAP_PROP_FORMAT', 'CAP_PROP_MODE', 'CAP_PROP_BRIGHTNESS', 'CAP_PROP_CONTRAST', 'CAP_PROP_SATURATION', 'CAP_PROP_HUE', 'CAP_PROP_GAIN', 'CAP_PROP_EXPOSURE', 'CAP_PROP_CONVERT_RGB', 'CAP_PROP_WHITE_BALANCE_BLUE_U', 'CAP_PROP_RECTIFICATION', 'CAP_PROP_MONOCHROME', 'CAP_PROP_SHARPNESS', 'CAP_PROP_AUTO_EXPOSURE', 'CAP_PROP_GAMMA', 'CAP_PROP_TEMPERATURE', 'CAP_PROP_TRIGGER', 'CAP_PROP_TRIGGER_DELAY', 'CAP_PROP_WHITE_BALANCE_RED_V', 'CAP_PROP_ZOOM', 'CAP_PROP_FOCUS', 'CAP_PROP_GUID', 'CAP_PROP_ISO_SPEED', 'CAP_PROP_BACKLIGHT', 'CAP_PROP_PAN', 'CAP_PROP_TILT', 'CAP_PROP_ROLL', 'CAP_PROP_IRIS', 'CAP_PROP_SETTINGS', 'CAP_PROP_BUFFERSIZE', 'CAP_PROP_AUTOFOCUS']
> 
> def decode_fourcc(v):
>   v = int(v)
>   return "".join([chr((v >> 8 \* i) & 0xFF) for i in range(4)])
> 
> print cv2.**version**
[...]

PR #7196, opened 2016-08-29
Added support for V4L2_PIX_FMT_Y16

stbnps:
Resolves #7194

- Changed cap_v4l.cpp so that opencv tries to use V4L2_PIX_FMT_Y16.
- convert_rgb is turned off when using Y16  since it is a monochrome format.
- When using Y16 the module sets the image depth to IPL_DEPTH_16U.

PR #12454, opened 2018-09-07
Add support for changing fourcc and support mono formats (e.g. Y8, Y16) in DirectShow

jveitchmichaelis:
### This pullrequest changes

This PR adds support for Y16 cameras (and some others) using the DirectShow library on Windows. In the process it also fixes a bug where changing the FOURCC property has no effect. This code is based on a modified patch from e-Con systems, published here with their permission (see [here](https://www.e-consystems.com/blog/camera/accessing-see3cam-custom-format-with-opencv/)).

Given the discussion here: https://github.com/opencv/opencv/pull/7196 I've tried to keep the same behavior:

- Cameras still open in RGB24 mode by default
- If the user selects Y16 via `CAP_PROP_FOURCC`, DirectShow retrieves a 16-bit image, but it is implicitly converted to a grey RGB image and returned. I bit-shift the image for the conversion.
- If the user sets `CV_CAP_PROP_CONVERT_RGB` to `false`, and the camera is returning 16-bit images, a 16 bit Mat is returned. Ditto for single channel mono images (e.g. Y8)

I've tested this with a FLIR Boson in both RGB and Y16 modes.


jveitchmichaelis:
I believe this also fixes https://github.com/opencv/opencv/issues/7013

alalek:
Rebased onto 3.4 branch.
We will merge changes from 3.4 into master.

PR #12463, opened 2018-09-07
added basic support for CV_16F (the new datatype etc.)

vpisarev:
CV_USRTYPE1 is now equal to CV_16F, which may break some [rarely used] functionality. We'll see. The following functions and methods should support CV_16F:
 - [x] Mat::copyTo()
 - [x] Mat::setTo()
 - [x] Mat::convertTo()
 - [x] RNG::fill()
 - [x] split(), merge() and most other functions that only need to know the element size to process it, not the content.
 - [x] norm()
 - [ ] arithm and logic functions (do we need it?)
 - [ ] print() (need to test it)
 - [ ] persistence (XML, YAML, JSON I/O)

<cut/>

```
CPU_BASELINE:Custom=AVX2
```


modules/core/src/rand.cpp
```diff
@@ -511,8 +391,8 @@ static RandnScaleFunc randnScaleTab[] =
 void RNG::fill( InputOutputArray _mat, int disttype,
                 InputArray _param1arg, InputArray _param2arg, bool saturateRange )
 {
-    CV_Assert(!_mat.empty());
-
+    if (_mat.empty())
+        return;
```
alalek: Early error detection: #8300

Filling **empty** Mat with random numbers is not a normal use case of OpenCV.

modules/ts/src/ts_perf.cpp
```diff
@@ -2207,7 +2207,7 @@ void PrintTo(const MatType& t, ::std::ostream* os)
         case CV_32S: *os << "32S"; break;
         case CV_32F: *os << "32F"; break;
         case CV_64F: *os << "64F"; break;
-        case CV_USRTYPE1: *os << "USRTYPE1"; break;
+        case CV_USRTYPE1: *os << "16F"; break;
         default: *os << "INVALID_TYPE"; break;
```
alalek: It is time to replace this switch to something of:
```
depthToString(t)
typeToString(t)
```

modules/core/include/opencv2/core/hal/interface.h
```diff
@@ -76,6 +76,7 @@ typedef signed char schar;
 #define CV_32F  5
 #define CV_64F  6
 #define CV_USRTYPE1 7
+#define CV_16F  7
```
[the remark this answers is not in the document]
vpisarev: there are a few places where USRTYPE1 is still used, but need to be adjusted, e.g. Torch importer in DNN.

modules/ts/src/ts_perf.cpp
vpisarev: the naive change to typeToString(t) breaks the perf tests because of different naming conventions. Let's leave it as-is for now

vpisarev:
:+1:

modules/core/include/opencv2/core/hal/interface.h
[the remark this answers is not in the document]
alalek: New size of these elements are 2 bytes instead of previous 8.
This will lead to memory corruption.

/cc @dkurt

PR #12473, opened 2018-09-09
CMAKE: Improve python detection

cv3d:
I am not sure if the issue I went through is reproducible, but this change makes things more straightforward.

Basically I was experimenting with Emscripten on Windows, which affected my environment varibles heavily, and as a result `PYTHON2INTERP_FOUND` became set, while `PYTHON2_EXECUTABLE` is empty, thus broke building all python-based modules!
Anyway, what matters to us is that `PYTHON*_EXECUTABLE` is set, not `PYTHON*INTERP_FOUND`

### This pullrequest changes
- Improve detection of Python executable by checking for `PYTHON*_EXECUTABLE` rather than for `PYTHON*INTERP_FOUND`


cv3d:
@alalek I updated accordingly

I guess the issue I experienced is due to a dirty build directory.
However, I still believe merging this PR would prevent some potential issues.

alalek:
OK, this should not break anything.

Please rebase this patch onto 3.4 branch.

cv3d:
@alalek Thanks! I just rebased.

PR #12479, opened 2018-09-10
Update hog.cpp

sturkmen72:
### This pullrequest changes

fixes oclSvmDetector issue when the detector loaded by HOGDescriptor.load() method.

### DIFF
--- a/modules/objdetect/src/hog.cpp
+++ b/modules/objdetect/src/hog.cpp
@@ -163,8 +163,9 @@ bool HOGDescriptor::read(FileNode& obj)
     FileNode vecNode = obj["SVMDetector"];
     if( vecNode.isSeq() )
     {
-        vecNode >> svmDetector;
-        CV_Assert(checkDetectorSize());
+        std::vector<float> _svmDetector;
+        vecNode >> _svmDetector;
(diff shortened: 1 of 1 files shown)


modules/objdetect/src/hog.cpp
```diff
@@ -163,8 +163,10 @@ bool HOGDescriptor::read(FileNode& obj)
     FileNode vecNode = obj["SVMDetector"];
     if( vecNode.isSeq() )
     {
-        vecNode >> svmDetector;
+        std::vector<float> _svmDetector;
+        vecNode >> _svmDetector;
         CV_Assert(checkDetectorSize());
+        setSVMDetector(_svmDetector);
```
alalek: Looks like, `checkDetectorSize()` check should be after `setSVMDetector()`. Right?

sturkmen72: @alalek could you remove `CV_Assert(checkDetectorSize());` here because the same line in `setSVMDetector`

alalek: Yes, we can remove this check.

Kauldwish:
setSVMDetector This should work

PR #12482, opened 2018-09-10
Fix include paths when building with external protobuf

GeorgeMironov:
Typo I believe. No such property as `INTERFACE_INCLUDE_SYSTEM_DIRECTORIES` in cmake

### DIFF
--- a/cmake/OpenCVFindProtobuf.cmake
+++ b/cmake/OpenCVFindProtobuf.cmake
@@ -50,7 +50,8 @@ else()
       add_library(libprotobuf UNKNOWN IMPORTED)
       set_target_properties(libprotobuf PROPERTIES
         IMPORTED_LOCATION "${Protobuf_LIBRARY}"
-        INTERFACE_INCLUDE_SYSTEM_DIRECTORIES "${Protobuf_INCLUDE_DIR}"
+        INTERFACE_INCLUDE_DIRECTORIES "${Protobuf_INCLUDE_DIR}"
+        INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${Protobuf_INCLUDE_DIR}"
       )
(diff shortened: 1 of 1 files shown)


alalek:
`INTERFACE_INCLUDE_SYSTEM_DIRECTORIES` is available [since CMake 3.0](https://cmake.org/cmake/help/v3.0/prop_tgt/INTERFACE_INCLUDE_DIRECTORIES.html#prop_tgt:INTERFACE_INCLUDE_DIRECTORIES).
Change should be guarded with CMake version check (we don't want compiler warnings from external headers, so "SYSTEM" is required here)

GeorgeMironov:
@alalek You refer to `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES` i guess. So it's still mistyped.  
[Cmake docs](https://cmake.org/cmake/help/v3.4/prop_tgt/INTERFACE_SYSTEM_INCLUDE_DIRECTORIES.html) says that "adding directories to this property does not make those directories used during compilation. Adding directories to this property marks directories as SYSTEM which otherwise would be used in a non-SYSTEM manner". So I changed it set both `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES` and `INTERFACE_INCLUDE_DIRECTORIES`

alalek:
Right, correct name: `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES`
Available in [CMake 2.8.12](https://cmake.org/cmake/help/v2.8.12/cmake.html) (minimal CMake version required for OpenCV), so check/duplication is not required.

GeorgeMironov:
Duplication is still required, setting just `INTERFACE_SYSTEM_INCLUDE_DIRECTORIES` leads to a bunch of errors about missing headers. This flags just marks already included directories to be treated as `SYSTEM` without actually adding them to include paths.

alalek:
What CMake version do you use?

GeorgeMironov:
3.5.1

PR #12310, opened 2018-08-25
Chunks/enum interface

cv3d:
- Merge #12601 first to avoid breaking JS bindings
- Merge with opencv/opencv_contrib#1744

### This pullrequest changes
A small portion of a bigger project. Please refer to #12288 for details

```
force_builders=Custom,linux32,win32,oclmacosx,windows_vc15
docker_image:Custom=ubuntu-cuda:16.04
```

PR #12487, opened 2018-09-10
Refactor core module for type-safety

cv3d:
Merge with opencv/opencv_contrib#1768 <s>and opencv/opencv_extra#518</s>

### This pullrequest changes
- Provides a backward-compatible API based on enums for type-safty. Refer to #12288 for further details

### Utilized options/preprocessors:
- `CV_TYPE_SAFE_API` with CMAKE option `ENABLE_TYPE_SAFE_API` to enable enum-based type-safe API
- `CV_TYPE_COMPATIBLE_API` with CMAKE option `ENABLE_COMPATIBLE_API` to enable the overloaded int-based API. Although it is enabled by default and using it would raise deprecation warnings, it still recommended to disable it in the build farm to enforce good practices. `CV_COMPATIBLE_API` is only available when `ENABLE_TYPE_SAFE_API` is set.
- `CV_TRANSNATIONAL_API` utilized internally, and should be removed after migration completes.

```
force_builders=Custom,linux32,win32,windows_vc15,Android pack,ARMv7,ARMv8
docker_image:Docs=docs-js
```


cv3d:
@alalek @vpisarev @hrnr Although it is not complete yet, I hope you can give it a fast review to let me know if it is meeting your requirements, so that I handle your further concerns in it and on the rest of the library.

modules/core/include/opencv2/core/base.hpp
```diff
@@ -359,6 +377,18 @@ enum BorderTypes {
 #define CV_RETHROW() throw
 #endif
 
+#define CV_STR_HELPER(x) #x
+#define CV_STR(x) CV_STR_HELPER(x)
+
+#ifdef __GNUC__
+#define CV_COMPILE_TIME_WARNING(msg) CV_DO_PRAGMA(message(msg)) //TODO: use #warning(msg)
```
alalek: Use function attributes instead:
- GCC `__attribute__ ((deprecated))`: https://gcc.gnu.org/onlinedocs/gcc-5.4.0/gcc/Function-Attributes.html
- MSVC has https://msdn.microsoft.com/en-us/library/044swk7y.aspx

Both support text messages.

modules/core/include/opencv2/core/check.hpp
```diff
@@ -10,20 +10,48 @@
 namespace cv {
 
 /** Returns string of cv::Mat depth value: CV_8U -> "CV_8U" or "<invalid depth>" */
-CV_EXPORTS const char* depthToString(int depth);
+CV_EXPORTS const char* depthToString(ElemDepth depth);
```
alalek: This file is quite new.
Avoid compatibility guards here.

modules/core/include/opencv2/core/private.hpp
```diff
@@ -317,6 +317,14 @@ static inline IppDataType ippiGetDataType(int depth)
         (IppDataType)-1;
 }
 
+#if defined(CV_COMPATIBLE_API)
+static inline IppDataType ippiGetDataType(int depth)
```
alalek: It is "private" header. Avoid compatibility stuff here.

modules/core/perf/cuda/perf_gpumat.cpp
```diff
@@ -88,14 +88,14 @@ PERF_TEST_P(Sz_Depth_Cn, CUDA_GpuMat_SetTo,
 
 PERF_TEST_P(Sz_Depth_Cn, CUDA_GpuMat_SetToMasked,
             Combine(CUDA_TYPICAL_MAT_SIZES,
-                    Values(CV_8U, CV_16U, CV_32F, CV_64F),
+                    Values(CV_8U, CV_16U, CV_32FC1, CV_64F),
```
alalek: This looks strange. Is it a test?

Same changes are above/below.

modules/core/src/matrix_expressions.cpp
```diff
@@ -1669,7 +1693,22 @@ MatExpr Mat::zeros(int rows, int cols, int type)
     return e;
 }
 
-MatExpr Mat::zeros(Size size, int type)
+#if defined(CV_COMPATIBLE_API)
+MatExpr Mat::zeros(int rows, int cols, int type)
```
alalek: > matrix_expressions.cpp

Lets make compatibility parts as a header only (inline functions/methods). Only public headers.

modules/core/src/norm.cpp
```diff
@@ -630,7 +631,7 @@ double cv::norm( InputArray _src, int normType, InputArray _mask )
     normType &= NORM_TYPE_MASK;
     CV_Assert( normType == NORM_INF || normType == NORM_L1 ||
                normType == NORM_L2 || normType == NORM_L2SQR ||
-               ((normType == NORM_HAMMING || normType == NORM_HAMMING2) && _src.type() == CV_8U) );
+               ((normType == NORM_HAMMING || normType == NORM_HAMMING2) && _src.type() == CV_8UC1) );
```
alalek: Looks like bug is catched here from the original code.

Why not `_src.depth() == CV_8U` ?

/cc @vpisarev

modules/core/src/ocl.cpp
```diff
@@ -1980,7 +1980,7 @@ struct Context::Impl
             handle = NULL;
     }
 
-    Impl(int dtype0)
+    Impl(ElemType dtype0)
```
alalek: Revert changes here.
It is OpenCL device type.

modules/core/src/system.cpp
```diff
@@ -1851,7 +1851,7 @@ IntrumentationRegion::IntrumentationRegion(const char* funName, const char* file
             return;
         }
 
-        int depth = pTLS->pCurrentNode->getDepth();
+        ElemDepth depth = pTLS->pCurrentNode->getDepth();
```
alalek: Revert this.
It is functions calling depth.

modules/core/src/matrix_expressions.cpp
cv3d: `MatExpr` is incomplete type, thus needs declaration & body separation.
In fact, I have already tried inline definition, and also changing the return type to `Mat` and statically casting the embedded return as well, without any success..
Accordingly, I separated functions returning `MatExpr` as an exception

modules/core/include/opencv2/core/private.hpp
cv3d: It is utilized by the imgproc module though.. Is that really alright?

alalek:
@cv3d Today I'm out of keyboard until ~ 18:00 UTC - do not expect fast response from me.

cv3d:
@alalek I am getting undefined reference to some functions that are already defined 😕 such as in [this build](http://pullrequest.opencv.org/buildbot/builders/precommit_linux64_no_opt/builds/16812).
Any help is highly appreciated.

cv3d:
The current PR refactores the core code and tests it with the non-core code as a user code. Although there are still some failures, hopefully they will be fixed soon. 
> we can use non-core OpenCV code / tests as "user code", which we should not break.

> one of the acceptance criteria for me would be - users do not have to change all CV_32F to CV_32FC1 etc. 

@alalek @vpisarev @hrnr [Most of the tests are now passing](http://pullrequest.opencv.org/buildbot/builders/precommit-contrib_windows64/builds/12072) without touching most of the modules 😄

cv3d:
@alalek Thanks for your support so far. Another issue that I hope you can assist me with is the assertion `(!fixedType() || ((Mat*)obj)->type() == mtype)` failure which happens in `_OutputArray::create`.

I believe the bug is in the cv3d@1bcb004e, but I just cannot find it.. Can you please give it a look?

cv3d:
> This build step trying to build samples with "installed" OpenCV binaries + headers.

Is that even doable using the current legacy API? The problem is persisting as in [this build](http://pullrequest.opencv.org/buildbot/builders/precommit_linux64_no_opt/builds/16816) despite utilizing `static inline`...

modules/core/include/opencv2/core/base.hpp
cv3d: I plan to utilize this once the current compile issues go through

cv3d:
@alalek Never mind my previous request about some assertion failures, as it is already solved!
I was doomed with some virtual functions of int arguments, which cannot call the enum-type override!

alalek:
> virtual functions

Could you point on some examples? Perhaps we can skip compatibility stuff for them.

alalek:
Compatibility stuff is not needed in tests.

cv3d:
@alalek Then can we consider [this result](http://pullrequest.opencv.org/buildbot/builders/precommit_windows64/builds/17366) as sufficient proof that the core stuff would not affect the user code? If so, I can proceed and add the rest of the commits for the remaining modules, and avoid the overhead of writing a transnational API as well.

cv3d:
Or at least, while leaving the code of non-core modules as is, allow me to rewrite their unit tests to avoid compatibility stuff in tests.

cv3d:
I managed to make Windows compile and test all modules properly, but since I do not have access to [Linux ](http://pullrequest.opencv.org/buildbot/builders/precommit_opencl_linux/builds/17359)or [Mac](http://pullrequest.opencv.org/buildbot/builders/precommit_macosx/builds/17138), I find it difficult to debug them and find the reason for failing tests.

@alalek Can you please help?

alalek:
@cv3d We need to process OpenCV headers from "standalone samples" with the same flags.
Workaround is here: https://github.com/alalek/opencv/commit/pr12487 (but proper solution is to update generated OpenCVConfig.cmake - I will do it later)

cv3d:
This most probably will do it for the samples linkages, but how about the failing tests as in the provided links above?

modules/core/include/opencv2/core/base.hpp
```diff
@@ -57,6 +57,77 @@
...
 #include "opencv2/core/cvstd.hpp"
 
+#ifdef CV_COMPATIBLE_API
+
+namespace std
+{
+    CV_DEPRECATED
+    static inline const int& max(const ElemDepth& left, const int& right)
+    {
+        return max(static_cast<int>(left), right);
```
alalek: Return type is invalid here.
Never return reference on local objects.

Also "const reference" in arguments is useless and problematic for enum/int/other scalar types.

P.S. For better understanding, try to interpret reference as a syntax sugar for "pointers".

alalek:
Could we separate warnings into two types:
- compatibility for ElemDepth->ElemType: warning is disabled by default, can be enabled by extra C++ flag only (no need to extend CMake)
- compatibility via 'int': warnings are enabled by default, but can be disabled C++ flag (no need to change CMake)

```
#ifdef OPENCV_WARNING_ENABLE_DEPRECATED_ELEMDEPTH_ELEMTYPE_OVERLOAD
#define CV_DEPRECATED_ELEMDEPTH_PARAM(type) CV_DEPRECATED_MSG(CV_DEPRECATED_PARAM(ElemDepth, type, ElemType, type))
#else
#define CV_DEPRECATED_ELEMDEPTH_PARAM(type) /* nothing, default */
#endif

#ifndef OPENCV_WARNING_DISABLE_DEPRECATED_INT_ELEMTYPE_OVERLOAD
#define CV_DEPRECATED_ELEMDEPTH_INT_PARAM(type) CV_DEPRECATED_MSG(CV_DEPRECATED_PARAM(ElemDepth, type, ElemDepth, type))
#define CV_DEPRECATED_ELEMTYPE_INT_PARAM(type) CV_DEPRECATED_MSG(CV_DEPRECATED_PARAM(int, type, ElemType, type))
#else
#define CV_DEPRECATED_ELEMDEPTH_INT_PARAM(type) /* nothing */
#define CV_DEPRECATED_ELEMTYPE_INT_PARAM(type) /* nothing */
#endif
```


> workaround for Android pack

Working on it.

cv3d:
Sure! Will do that.

Do you plan to enable all warnings in the build farm?
If not, I am afraid all the efforts spent in this whole work will eventually go in vain..

alalek:
> Android pack

Please replace my commit by another one: https://github.com/alalek/opencv/commits/pr12487

modules/core/src/norm.cpp
cv3d: You will see hundreds of similar cases! There are even functions that receive type, although they just need the depth... I have actually made several decisions in utilizing depth instead of type (e.g. in all arithmetic operations, `convertTo`, `assign`, `assignTo`, `calcCovarMatrix`), and I really hop you will catch what I missed.

cv3d:
@alalek Thanks for your efforts!
However, with this workaround, CUDA cannot link the libraries 😕 
Can you please investigate?

cv3d:
I am sure you are aware of this, but let me stress on the fact that all functions/methods that are now receiving `ElemDepth` should be just fine.

However, if the function has an `ElemType` parameter then the question is: does it really need the channel information? Is not `ElemDepth` enough? If you find any such cases, please let me know, and I will be happy to change it.

vpisarev:
`ElemType` and `ElemDepth` should be made the same type

cv3d:
> users do not have to change all CV_32F to CV_32FC1 etc. And in general CV_... and CV_...C1 should be identical entities, not different

@vpisarev Currently, users do not have to change `CV_*` to `CV_*C1` and they are exchangable, and the comparison `CV_* == CV_*C1` holds true, thus I cannot understand why
> ElemType and ElemDepth should be made the same type

Are you implying that `CV_32FC1` should be considered `ElemDepth` as well?

alalek:
> ElemType and ElemDepth should be made the same type

There is a lot of code like this:
- `ddepth <= CV_32F ? CV_32F : CV_64F`
- `max(ddepth, CV_32F)`

OpenCV has well defined ordering of elements depth (actually it had, until CV_16F patch).

There is no simple way to define similar ordering for "complete" type with channels (CV_8UC2 vs CV_16SC1 - ... lets don't allow users to shoot foots).

C++ forbids sharing "values" between two non-class enums.

Another sample with type/depth mess:
```
auto type = src.type();
CV_Assert(type == CV_8U);
```

Do we want to avoid supporting CV_8UC3 here or does it a coding mistake and depth should be used here?

Additionally:

```
switch (depth)
{
    case CV_8U: ...
    case CV_8S: ...
    ...
    // compiler emit warnings if we missing some type depth (currently there are warnings for CV_16F in tests)
}
```

---

This patch avoids the most part of required code refactoring (8k changed lines => less that 100 changes outside of core module for both opencv + contrib codebases) and functions accepts depth instead of type and vise versa (however functions documentation forbid this). So impact on user code should be tolerable if code doesn't look like this: `UMat dst(1, 1, 0);`.

Warnings are configurable (will be disabled for user code and enabled for OpenCV itself for better code quality).

CMakeLists.txt
```diff
@@ -316,8 +316,8 @@ OCV_OPTION(INSTALL_TESTS            "Install accuracy and performance test binar
 
 # OpenCV build options
 # ===================================================
-OCV_OPTION(ENABLE_TYPE_SAFE_API       "Enable enum-based type-safe API? If disabled, int types will be utilized" ON)
-OCV_OPTION(ENABLE_COMPATIBLE_API      "Enable backwards compatibility with legacy API?"          ON IF ENABLE_TYPE_SAFE_API)
+OCV_OPTION(DISABLE_TYPE_SAFE_API      "Disable enum-based type-safe API? If disabled, 'int' types will be utilized" OFF)
+OCV_OPTION(DISABLE_TYPE_COMPATIBLE_API "Enable backwards compatibility with legacy API?"         ON IF DISABLE_TYPE_SAFE_API)
```
cv3d: Should not this be `DISABLE_TYPE_COMPATIBLE_API "..." OFF if NOT DISABLE_TYPE_SAFE_API`?

modules/core/include/opencv2/core/base.hpp
```diff
@@ -57,6 +57,172 @@
...
+    }
+
+    CV_DEPRECATED_MAX_DEPTH_INT_ATTR static inline int max(const int left, const ElemDepth right)
+    {
+        return max(left, static_cast<int>(right));
+    }
+
+    CV_DEPRECATED_MAX_DEPTH_DEPTH_ATTR static inline int max(const ElemDepth left, const ElemDepth right)
+    {
+        return max(static_cast<int>(left), static_cast<int>(right));
```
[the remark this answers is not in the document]
cv3d: > Do we really need to deprecate this (Depth, Depth) overload?

If we want to avoid `CV_16F > CV_32F` evaluation as true, then yes we do.

> CV_FP16 brokes existed ordering

We can manage that inside `CV_MAX_DEPTH` macro.

> (Type, Type) should be deprecated

I am not sure I have seen such comparisons, but it makes sense

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -463,14 +544,28 @@ class CV_EXPORTS MatAllocator
...
+    //virtual void allocate(int dims, const int* sizes, ElemType type, int*& refcount,
     //                      uchar*& datastart, uchar*& data, size_t* step) = 0;
     //virtual void deallocate(int* refcount, uchar* datastart, uchar* data) = 0;
-    virtual UMatData* allocate(int dims, const int* sizes, int type,
-                               void* data, size_t* step, int flags, UMatUsageFlags usageFlags) const = 0;
-    virtual bool allocate(UMatData* data, int accessflags, UMatUsageFlags usageFlags) const = 0;
+    virtual UMatData* allocate(int dims, const int* sizes, ElemType type,
+                               void* data, size_t* step, AccessFlag flags, UMatUsageFlags usageFlags) const = 0;
+#ifdef CV_COMPATIBLE_API
+    CV_DEPRECATED_INT_TO_ELEMTYPE_ATTR(type, type)
```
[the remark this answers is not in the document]
cv3d: Will convert that into `CV_TRANSNATIONAL_API`, and remove it after all factoring is done

cv3d:
@alalek Do we need your last three commits?

alalek:
Merge changes from "cmake: change type safe flags defaults" into your commits, other are just tests for green builds (plus one commit in contrib).

modules/core/include/opencv2/core/base.hpp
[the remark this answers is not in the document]
cv3d: I gave it another thought, and I realized that deprecation is about dropping support, while
> (Type, Type) 

was not supported to begin with. Therefore, allow me to skip this, otherwise, if we try to mark each stupid practice as a depreciated one, we will have thousands of lines to write.
I guess it is indispensable for reviewers to catch such abuses.

cv3d:
@alalek All done

cv3d:
> UMat dst(1, 1, 0);

@alalek When `CV_TYPE_COMPATIBLE_API` is set (default), even this ugly code shall work 😉

@vpisarev @mshabunin I have refactored the whole library including apps and samples.
With "Refactor *" PRs, OpenCV gets more robust, convenient, while maintaining high compatibility with previous versions with configurable deprecation level.

cv3d:
The following test must be added to the core tests, however, as you might already know, it is currently failing 😕 

```C++
    vector<ElemDepth> order { CV_8U, CV_8S, CV_16U, CV_16S, CV_16F, CV_32S, CV_32F, CV_64F };

    for (int i = 0; i < order.size(); ++i)
    {
        for (int j = 0; j < order.size(); ++j)
        {
            ASSERT_EQ(order[i] <  order[j], i <  j);
            ASSERT_EQ(order[i] <= order[j], i <= j);
            ASSERT_EQ(order[i] >  order[j], i >  j);
            ASSERT_EQ(order[i] >= order[j], i >= j);
        }
    }
```

Since `CV_MAX_DEPTH` and `CV_MIN_DEPTH` introduced in this PR are doing their job properly, my idea is to introduce the following operators:
```C++
static inline bool operator>=(const ElemDepth left, const ElemDepth right) 
{
    return left == CV_MAX_DEPTH(left, right);
}
static inline bool operator<=(const ElemDepth left, const ElemDepth right) 
{
    return left == CV_MIN_DEPTH(left, right);
}
static inline bool operator>(const ElemDepth left, const ElemDepth right) 
{
    return left != right && left >= right;
}
static inline bool operator<(const ElemDepth left, const ElemDepth right) 
{
    return left != right && left <= right;
}
```
but these operators causes ambiguity! Can you find any workaround?

Another idea is to make `ElemDepth` a class instead of an enum. What do you think?

vpisarev:
@cv3d, @alalek, well, first of all, in order to prove that users do not have to change `CV_32F` to `CV_32FC1` I'd like to ask you to remove all such changes in the patch.

Then, I understand that you provided some implicit conversions between ElemDepth and ElemType, but then why not just make them the same type?

@alalek, regarding the switch statement - good robust practice is to put "default" case regardless of whether all the cases are covered or not. It's future-proof decision that stops all the compiler complains.

max(), min() etc. - not a problem either. The enumeration values can be converted to int's and the result can be converted to the enumeration - easy. The same is about comparison operations. Often it's useful to have ordering operations just in order to be able to create std::map with ElemType keys or just get some consistent ordering. The actual ordering is often not that important (as soon as it's fixed and known to user).

Actually, instead of using min(), max() etc. I would rewrite the corresponding code so that it extracts bit-depth, sign presence and integer/floating-point flag from the type value and makes the decision based on these properties of the type.

In other words, I cannot find any reasons not to make those types identical. In my opinion, it will break much less code than the solution with 2 different types and implicit conversions between them.

cv3d:
> to remove all such changes in the patch

In this patch, the base type is already changed from `int` to either `ElemType` or `ElemDepth`. Once such change is made, it is not possible to assign and `int` to  `ElemType` or `ElemDepth`, or to assign a `CV_\d+[UFS]` to a `ElemType`, or `CV_\d+[UFS]C\d+` to `ElemDepth`, and that is exactly the point of type-safety. On the other hand, user code is still using `int` as the base type, therefore there is no immediate need for them to change anything, and **it will just work the same way the rest non-core OpenCV library did**. If you really insist, then I can reduce this particular PR to a minimal one, in which the base types remain `int`, but then the question is: what is the point of providing overloads that are not tested at all? Please note that this PR in its current situation provides a strong evidence that both the new API (`ElemType` & `ElemDepth`) and the legacy API (`int-based`) can work interchangeably without issues.

> why not just make them the same type

There are two types of implicit conversions: `int` to `ElemType`/`ElemDepth`, and `ElemDepth` to `ElemType`. The first type is meant to avoid breaking user code, while encouraging them to migrate from the type-unsafe `int` usage via a gentle compile-time warning. The second type (which I believe you are referring to) is again to mitigate breaking user code, and to improve the code quality. Here are some examples:
- `convertTo()` changes the depth of a matrix. However, there is inconsistency in calling it. Cases such as `mat.coverTo(dst, CV_32FC3)` or `mat.coverTo(dst, CV_MAKETYPE(depth, 3))` are very common even inside OpenCV itself, and inside the function you would notice things like `type = CV_MAT_DEPTH(type)` to strip the channel information. On the other hand, via `ElemDepth` and `ElemType` seperation, we would like to maintain clean code, in which beautiful and meaningful statements such as `mat.coverTo(dst, depth)` are utilized.
- `int type = ...; if (type == CV_32F) { ... }` These cases creates huge ambiguity and are hard-to-maintain. It is not clear if the intetion is to allow a specific depth (variable name is badly chosen?) or if single channel is the only allowed case.

Your question boils down to the basic nature of depth and type of a matrix. If they are made the same, then why does `.depth()` and `.type()` exist? Is not `.type()` enough? I bet convenience, clarity, and maintainability were the original motivation behind them, and so do our motivation behind this PR.

> max(), min() etc. - not a problem either

With all my respect, it is a problem. For instance, `max(CV_16F, CV_32F)` would result in `CV_16F`, and that is mostly not the user intent. We have deprecated such functions while correcting their behaviour for the time-being, and proposed to use `CV_MAX_DEPTH` or `CV_MIN_DEPTH` instead.

> The actual ordering is often not that important

I am sure I misunderstood, because your statement sounds to me like changing `CV_8U` value to 6  and `CV_32F` to 0 is not that important either.

> it will break much less code

It seems there is a conflict between our definitions of depth and type. Personally, I cannot say `depth is a type`, or `type is a depth`. But I can say `type consists of a depth and number of channels`.  If depth is a type, and type is a depth, then channel is a depth, and channel is a type as well by induction.
More importantly, to make such argument more concrete, could you please provide a scenario in which this PR would fail, while a single-type proposal would not?

alalek:
@vpisarev

> order to prove that users do not have to change CV_32F to CV_32FC1

Do you see in this patch changes outside of core module? in samples?
This is a prove that legacy code can be compiled without required massive changes.
Yes, users would get a set of warnings about their code. Warnings are still configurable to remove most part of them, see #10318 (green builds with this patch).
Anyway they can turn off type-safety completely (and continue to use 'int').

Prove has been here since 10:00 UTC (Friday, 14 Sep).

> them the same type

They are not the same entity.

1.
Try to `#define CV_MAT_DEPTH(type) (type)` for making them the same. Nothing will work.

2.
We can't declare them to the same, because the current OpenCV implementation and related documentation doesn't expect "type" instead of "depth".
Explicit documentation statements about expected depth values only:
```
$ grep -R '@param' ./modules | grep depth | wc -l
103
```
But it is documentation only.
Idea of the current patch is enabling of these existed "documentation-only" declaration via C++ code too.

---

> switch statement - good robust practice is to put "default" case regardless of whether all the cases are covered or not.

"Robust practice" to compile code only. But problem will be observed in runtime only (hopefully test coverage is enough) or via bug tracker report from a customer.
These warning are useful then you are working with enums and trying to extend them. These warnings are added into compilers (static code analyzers) for are reason.

P.S. CV_16F patch brokes "ordering" of depths, so a lot of existed code would not work as expected:
- `CV_32F < CV_64F` is true
- `CV_16F < CV_64F` is false (?!)

cv3d:
> CV_16F patch brokes "ordering"

I verified locally (to some extent) that `ElemDepth` can handle this issue if converted to `enum class`.
But I prefer to have some security that my future efforts will not go in vain, so I will wait until most of the open PRs are merged first.

vpisarev:
@alalec
> Do you see in this patch changes outside of core module? in samples?
> This is a prove that legacy code can be compiled without required massive changes.

first of all, I see over a thousand of warnings produced by PR itself, we did not even touch the user code yet. Can you or @cv3d get rid of them in order to continue the discussion?

> They are not the same entity.
ok, we planned to introduce CV_TYPE_RAW, CV_TYPE_UNDEFINED types or something like that. What's the corresponding `depth` of those types? How do I specify such type if the corresponding function now takes `ElemDepth` parameter?

> 
> Try to `#define CV_MAT_DEPTH(type) (type)` for making them the same. Nothing will work.
I do not have to; we can have CV_MAT_DEPTH(type) cv::typeDepth(type) where typeDepth is declared as

`namespace cv { ElemType typeDepth(ElemType type); }`

> 
> We can't declare them to the same, because the current OpenCV implementation and related documentation doesn't expect "type" instead of "depth".
> ```
> $ grep -R '@param' ./modules | grep depth | wc -l
> 103
> ```

Look at imgproc and other modules that take the desired output type argument. We use both depth and type interchangeably, without any strict rules. 

```
grep -R '@param' ./modules | grep type | wc -l
532
```

> 
> P.S. CV_16F patch brokes "ordering" of depths, so a lot of existed code would not work as expected:
> 
> * `CV_32F < CV_64F` is true
> * `CV_16F < CV_64F` is false (?!)
in your opinion, what should be the result of `CV_16F < CV_32S`?

cv3d:
@vpisarev I think you got confused by the warnings and thought low of this PR.
The main point of splitting the proposal into several PRs is to proof that each and every part can operate with user code. If you want to see the green lights, then how about giving #12288 a visit, which merges all the pieces together? BTW: It is no longer up-to-date.

Again, the proposal have several modes:
- You can switch all types to `int` (ElemDepth & ElemType are typedefed to int - they are exactly the same thing here)
- You can enable enums  (ElemDepth & ElemType are traditional enums, thus they still have int as their base type)
- You can enable/disable int overloads and implicit conversions to maintain compatibility with user code (enabled by default, hopefully will get disabled by OpenCV 5)
- You can control if none, some, or all deprecation warnings are raised.

vpisarev:
@cv3d, providing a million of options is not a solution to the problem, it's a workaround.

The headers should contain as few conditional compilation clauses as possible.

Please, create "all-green" PR. I want to see the possible damage to OpenCV itself and to the user code that the change brings. If it's too many changes, we'll just keep the status quo.

vpisarev:
@cv3d, I discussed this one more time with @alalek. The common agreement is that it makes sense to start with higher-level modules and replace anonymous enumeration with names ones there.

I mean, we should start with videoio, introduce named enumeration for CAP_PROP_, continue with features2d, refactor enumerations there etc.

And in the end, we can do the final step and try to introduce ElemType/ElemDepth.

The attempt to do everything in one huge PR is not very good, because it's almost impossible to review

vpisarev:
I mean, PRs #12550, #12551, #12553, #12555 should be closed. Because they attempt to change one the same thing, just in different places. Instead, let's try to refactor those modules w/o touching ElemType/ElemDepth at the moment

cv3d:
I can rename ElemType and ElemDepth to int in all open PRs, thus you can consider them all as just some kind of cleanup. This is to avoid wasting the effort (due to conflicts that I fix everyday), and to simplify the future steps once we arrive to ElemType and ElemDepth again.
@vpisarev Is that fair enough?

alalek:
@vpisarev Changes of "High" level enums are in another PR (before this one): #12310

cv3d:
My plan is to tackle all the unnamed enums(only 17 out of 107 are complete),
but it seems frustration is getting me now due to the tedious amendments ...

It is now up to @vpisarev, who seems to be the boss here, to let me to continue with my tasks, or to kick me away!

cv3d:
> create "all-green" PR. I want to see the possible damage to OpenCV itself and to the user code that the change brings.

@vpisarev All lights are green now

cv3d:
> it's almost impossible to review

I have reduced the size of PRs to something like half, and if you still think it is still huge, including this one, then I will split them further to make each one around 500 lines or so.

> users do not have to change all CV_32F to CV_32FC1 etc. And in general CV_... and CV_...C1 should be identical entities, not different. ... Otherwise, ok, I will definitely review the patch as soon as you finish it.

> create "all-green" PR

> ElemType and ElemDepth should be made the same type

> in the end, we can do the final step and try to introduce ElemType/ElemDepth

> should be closed

@vpisarev Let me remind you you initially asked that single-channel type should not be different from a depth, and I did that. Later, you asked for green lights, and I did that. Even by now, I do not see an upper limit to your decisions, but to throw away a month-worth of effort, is not that what you are asking for now? Now, if you really want them so badly to be the same even in case of multi-channels, then I will do that, given that this is your final change of mind.

vpisarev:
@cv3d,

well, if you look at the statistics of the closed PRs, you will see that we usually merge over 100 patches per month, over a thousand of patches per year. Most of the patches are not innocent, they fix a small bug, or bring some new optimization, or even add new functionality or new documentation. Such patches usually take little time to review and they get merged very fast.

Now the case of your patch is completely different. You started with very fundamental things in OpenCV that live there for 10 years already! You may not know, but I will tell you that many, many people still use OpenCV 2.4.x. I've heard about some big projects that only now consider migration to OpenCV 3.x, even though OpenCV 3.0 has been released 3 years ago. There is huge inertia in software.

If it ain't break, don't fix it. You said that you value compatibility, but I honestly do not see that. It's number #1 priority in our project. Between the radical patch that sort of improves API style and compatibility I would choose compatibility any time of day.

Say, if you come to Python community and suggest to change the language syntax "slightly", or come to Linux community and suggest to change a bit driver API that breaks backward compatibility - what would happen? Some PIPs in Python and some forks in Linux live separately for years.

OpenCV is 18 year-long project, we have hundreds of thousands of users, and even though we are quite liberal w.r.t patches and suggestions, we still have to carefully evaluate each patch that breaks compatibility. 1 month is nothing comparing to the years of support from our side and many man-years that OpenCV needs to spend to repair all the code that is broken because of this tiny patch. If the argument is that users will build OpenCV in "compatibility mode", then it looks like the patch is not up to some claims it makes - it does not make user's OpenCV-based code safer, it just makes OpenCV itself more type-safe (but we have many unit tests, valgrind and such, so OpenCV code itself is rather well-tested).

If you are brave enough (as one of the reviewers said) to submit such radical patches, please get enough patience and be ready to find compromises.

I value that you've updated the PR, thank you very much.

I still see that it's quite a lot of changes, including many cases of `CV_depth` to `CV_depthC1` changes, that I personally don't like, I think, users do not have to make such stupid changes in their code.

==

Here is the roadmap of this patch inclusion:

1. If you have time, please, prepare an alternative patch where there is no physically different type ElemDepth, just ElemType. Where there is zero changes from `CV_8U` to `CV_8UC1` etc. In your patch I still do not see it. We need to have some numerical estimate (basically, the number of patched lines or the patch size), how much more effort does it take to make ElemDepth and ElemType phyisically different types.

2. If you do not have time, then give me some time, I will create such patch, based on your patch, by myself.

3. After we have both patches and can compare them, there will be next step. I'm enthusiastic to see such a patch not just because of type safety. It's a smaller goal, in my opinion. The bigger goal is to phyisically separate elemType and array (Mat, UMat, etc.) flags. In your patch you use MagicFlag data structure, probably we need to consider 2 separate members: flags (or arrayflags to catch all the cases of improper use) and elemType. As soon as we have it, we will be able to add new ElemType members, like CV_TYPE_RAW, CV_32U, CV_64S, CV_TYPE_UNDEFINED, CV_TYPE_AUTO etc. That will be a real motivation for us and for users to make those changes - the set of supported data types will be extended.

We try to do it all by OpenCV 4.0. If not, in principle we can do it in 4.1. Or maybe we'll split `int flags` into `ElemType type` and `ArrayFlags arrayflags`, introduce `typedef int ElemType` `typedef int ArrayFlags` synonyms and leave the enumeration anonymous by 4.0 and then do this radical type-safe change in 4.1. We need to see, but first I need this alternative patch to be prepared (by you, @alalek or myself) to clearly see the difference in the impact.

cv3d:
@vpisarev I will prepare the `ElemType`-only patch. Hope that can be of any help!

cv3d:
> no physically different type ElemDepth, just ElemType

> zero changes from CV_8U to CV_8UC1

@vpisarev PR #12609 does what you request for. Please let me know if you still have further requirements.

> After we have both patches and can compare them

I have grouped the differences in a single commit for your convenience: cv3d@f72faed6b188937eaedf13749559109f492ba1e3
Anything else?

cv3d:
>  As soon as we have it, we will be able to add new ElemType members, like CV_TYPE_RAW, CV_32U, CV_64S, CV_TYPE_UNDEFINED, CV_TYPE_AUTO etc.

I also have the same goal in mind, but my approach is different. I plan to increase the capacity of `ElemDepth` enum from 8 to 16, thus be able to introduce these new types.
How about taking a look at my proposal at #12584?
Perhaps I am biased here, but I think it is cleaner, easier, and does not confuse users, in compare to merging-depth-and-type thing.

> If you are brave enough (as one of the reviewers said)

Darn! Is that why I am getting tortured now? It is not my fault.. I cannot control others 😢 

> please get enough patience and be ready to find compromises

please get enough faith/believe/trust and be ready to get amazed

cv3d:
> If it ain't break, don't fix it

We are not fixing it. We are introducing new feature: type-safety

> You said that you value compatibility, but I honestly do not see that

Besides your own opinion, what do you see exactly? What does this PR break?
Kindly be more technical and concrete.

> Where there is zero changes from CV_8U to CV_8UC1 etc. In your patch I still do not see it

I have reduced this PR to its minimal, in which there is zero changes from CV_8U to CV_8UC1 etc.
Please confirm, and kindly let me know if you still have any concrete concerns.

> the roadmap of this patch inclusion

After I prepared the comparison patch, I believe we are in the final bullet (although I cannot get why this PR shall wait for the new types anyway), but you also said
> there will be next step

It sounds like a recursive function call for me, but I do not see a termination criteria..

cv3d:
>  It's a smaller goal, in my opinion. The bigger goal is to phyisically separate elemType and array (Mat, UMat, etc.) flags.

@vpisarev I said it a week ago, and I will say it again: I am not convinced by your logic, and I cannot understand why this PR should wait..
In fact, I do not find any relation between the new types PR and this one.. Each one have different goal, and affects different components, with very small overlap.

BTW: Why all this silence?

cv3d:
>I'm enthusiastic to see such a patch

>As soon as we have it, we will be able to add new ElemType members

@vpisarev I believe you [meant this patch](https://github.com/opencv/opencv/pull/12730/commits/c61f6e54b75a0b744d26904d8a2cddbba9c29c7b). So, any chance to consider this PR as a separate and independent feature by now? I think I already demonstrated they are unrelated via my other design and new types PRs (i.e. #12729 & #12730).

Accordingly, I want to solve the conflicts in this PR in order to merge it, but first, I need your confirmation we will not have any further delays..

vpisarev:
experiment with ElemType-only patch is not finished yet (https://github.com/opencv/opencv/pull/12609). I left you multiple comments, but they are not addressed; instead, you continue to create tons of new PRs

cv3d:
@vpisarev My bad! I did not notice your reply 😕

PR #12497, opened 2018-09-11
dnn: remove raw SSE2 implementation

tomoaki0705:
### This pullrequest changes
 - remove raw SSE2 implementation
 - Unfortunately, neither test_dnn nor perf_dnn covers this part
 - Adding new new test case for these part seems really complicated task, so I'll just throw the PR

### DIFF
--- a/modules/dnn/src/layers/pooling_layer.cpp
+++ b/modules/dnn/src/layers/pooling_layer.cpp
@@ -488,19 +488,16 @@ public:
                                             max_val0 = v_max(max_val0, v0);
                                             max_val1 = v_max(max_val1, v1);
                                         }
-#if CV_SSE2
                                     else if( stride_w == 2 )
                                         for (int k = 0; k < kernel_w*kernel_h; k++)
                                         {
                                             int index = ofsptr[k];
-                                            v_float32x4 v00 = v_load(srcData1 + index), v01 = v_load(srcData1 + index + 4);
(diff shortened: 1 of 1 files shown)


vpisarev:
@tomoaki0705, thanks!

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -488,19 +488,18 @@ class PoolingLayerImpl CV_FINAL : public PoolingLayer
...
-#if CV_SSE2
                                     else if( stride_w == 2 )
                                         for (int k = 0; k < kernel_w*kernel_h; k++)
                                         {
                                             int index = ofsptr[k];
-                                            v_float32x4 v00 = v_load(srcData1 + index), v01 = v_load(srcData1 + index + 4);
-                                            v_float32x4 v0(_mm_shuffle_ps(v00.val, v01.val, _MM_SHUFFLE(2, 0, 2, 0)));
-                                            v_float32x4 v10 = v_load(srcData1 + index + 8), v11 = v_load(srcData1 + index + 12);
-                                            v_float32x4 v1(_mm_shuffle_ps(v10.val, v11.val, _MM_SHUFFLE(2, 0, 2, 0)));
+                                            v_int32x4 vs320, vs321, dummy;
```
vpisarev: why not use `v_float32x4`?

tomoaki0705: Nice catch !
I just found ```v_int32x4``` and didn't really look for ```v_float32x4```
Thank you !

vpisarev:
thanks! :+1:

PR #12502, opened 2018-09-11
dnn(test): fix failures of 32-bit builders

alalek:


modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -225,7 +229,11 @@ TEST_P(Test_ONNX_nets, VGG16)
     testONNXModels("vgg16", pb, l1, lInf);
 }
 
+#ifdef BUILD_32BIT
+TEST_P(Test_ONNX_nets, DISABLED_VGG16_bn)  // memory usage >2Gb
+#else
 TEST_P(Test_ONNX_nets, VGG16_bn)
+#endif
```
dkurt: Is it observed for both Linux and Windows configs? In example, there is a test which is skipped only for Win32 builds:
https://github.com/opencv/opencv/blob/10ae0c4364ecf1bebf3a80f7a6570142e4da7632/modules/dnn/test/test_caffe_importer.cpp#L154-L155

vpisarev:
`BUILD_32BIT` looks a bit weird name to me, can we have some `CV_` or `OPENCV_` prefix?

alalek:
My current assumption about `BIG_DATA_TEST()` is that it handles tests which use 4+Gb RAM.
Perhaps we can add additional macros for 1Gb, 2Gb, 4Gb, 7Gb, 15Gb memory limits.

alalek:
Lets merge it "as is" to make builds green again (patch is not large).
Future improvements can be added by next PRs.

PR #12508, opened 2018-09-12
Improve support of float16_t

cv3d:
Testing! do NOT merge
### This pullrequest changes
- Support some basic matrix operations using float16

### DIFF
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -597,6 +597,13 @@ protected:
     #endif
     }
 
+    template<typename T>
+    float16_t& operator =(const T& val)
+    {
+        *this = static_cast<float16_t>(val);
+        return *this;
+    }
+
     static float16_t fromBits(ushort b)
     {
         float16_t result;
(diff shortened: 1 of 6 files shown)


cv3d:
@alalek This PR is not complete yet, and perhaps I will not give it a lot of attention because I am not interested in float16, but I found it irritating that such support is not complete.
Mind to give it a look, and hopefully fix its glitch?

modules/core/include/opencv2/core/cvdef.h
```diff
@@ -597,6 +597,13 @@ class float16_t
     #endif
     }
 
+    template<typename T>
+    float16_t& operator =(const T& val)
```
alalek: There are two branches if `float16_t` implementation. They should have similar public interfaces.

Also conversion to float16 can be emulated and very slow, so we don't want to allow these automatic conversions for users - adding of this operator should be avoided.

FP16 support should be efficient.
There are no goals to push FP16 into all algorithms.

cv3d:
I see! I will close this PR then.

PR #12511, opened 2018-09-12
Rename tensorflow namespace

GeorgeMironov:
Preventing conflicts when linking both `opencv-dnn` and `tensorflow`. Same stuff as for Caffe

related #10139

```
allow_multiple_commits=1
```


cv3d:
How about just `tf` or `cv_tf`?

GeorgeMironov:
Used the same convention as for Caffe. It is not a part of public API so i don't think it really matters

alalek:
@GeorgeMironov How did you get `.pb.*` files?

PR #12526, opened 2018-09-13
Fixed bit-exact resize SIMD implementation for AVX2 baseline

terfendail:
### This pullrequest changes
Fixed bit-exact resize SIMD implementation for AVX2 baseline

```
force_builders=Custom,linux,macosx,windows,docs,iOS,ARMv8
CPU_BASELINE:Custom=AVX2
```


alalek:
Does it look like a compiler bug on Ubuntu 18.04?

There is [Fedora 28 (GCC 8.1.1) build](http://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/1154) without this patch.

What are new guidelines about using of `vxzeroupper` to avoid that?

/cc @vpisarev

PR #12530, opened 2018-09-13
Add semicolons after `CV_INSTRUMENT` macros

cv3d:
### This pullrequest changes
Very minor change, but it is a good practice. Without it, editors fails to indent the code properly.


cv3d:
BTW: I noticed that `hdr_parser.py` fails to parse headers containing macros without semicolon.
Since it is almost impossible to have a parser that understand any C++ file, instead, I suggest to enforce the semicolon via some linting test to avoid such cases.

alalek:
> hdr_parser.py

It supports very limited subset of C++ only.
Even "inline" functions can broke this parser.

PR #12538, opened 2018-09-14
Refactor imgproc module for type-safety

cv3d:
### This pullrequest changes
Extends the backward-compatible API introduced in #12487


cv3d:
@vpisarev @alalek As you already know, some of my PRs are there for a month or so, and I cannot keep the effort in solving conflicts every now and then in all of them anymore. Therefore, I thought to give it a try and simplify the review process by reducing the dependency on the core to the minimum. As you can see, the changed lines on this PR for example got reduced from `+5,370 −2,352` to `+1,411 −568`.

I am now going to do the same on all open PRs which are already conflicting, but in exchange, please consider the spent efforts and kindly help me focus on making new progress by merging them. Otherwise, my efforts will be spent on just maintaining and defending them, until I reach my limits and give up on it all.

PR #12546, opened 2018-09-15
Refactor test units & ts module for type-safety

cv3d:
Merge with opencv/opencv_contrib#1775 and opencv/opencv_extra#521

### This pullrequest changes
<s>Extends</s> Applies the backward-compatible API introduced in #12487 to unit tests

```
force_builders=Custom
```


cv3d:
@alalek I am quite annoyed that several methods does not utilize explicit `CV_OVERRIDE`, thus if the base virtual method signature is changed, they silently became overloaded without issuing any error, yet they never get called, and outcomes are, of course, catastrophic 😖 

Is there a way to enforce all methods that has the same name of a base method to explicitly specify either `CV_OVERRIDE` or `CV_OVERLOAD`?

cv3d:
Can you please restart the Win64 OpenCL builder?

vpisarev:
let's close this for now; too many irrelevant changes `CV_depth` to `CV_depthC1`

PR #12547, opened 2018-09-15
Refactor cuda (filters, imgproc, stereo) modules for type-safety

cv3d:
### This pullrequest changes
Extends the backward-compatible API introduced in #12487 

```
force_builders=Custom
```


cv3d:
@alalek Is it OK to group some small modules together, or shall I push each on on its own PR to proof that it is fully backwards compatible with the rest of the code?

cv3d:
@alalek If you are not sure, then splitting it is the safest bet.

BTW: I am not sure how to interpret all the silence, especially from pending reviewers..
Could you please emphasize on the situation?

PR #12548, opened 2018-09-15
videoio: added '.getBackendName()' method

alalek:
```
allow_multiple_commits=1
```

### DIFF
--- a/modules/videoio/include/opencv2/videoio.hpp
+++ b/modules/videoio/include/opencv2/videoio.hpp
@@ -808,6 +808,12 @@ public:
     */
     CV_WRAP virtual bool open(const String& filename, int apiPreference);
 
+    /** @brief Returns used backend API name
+
+     @note Stream should be opened.
+     */
+    CV_WRAP String getBackendName() const;
+
 protected:
     Ptr<CvCapture> cap;
     Ptr<IVideoCapture> icap;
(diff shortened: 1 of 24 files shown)


vpisarev:
@alalek, that's useful functionality! I wonder though, why not just add something like `CAP_PROP_BACKEND`  and retrieve it as property?

alalek:
String is preferable, but get returns "double" only (I prefer do not cast "const char*" to "double" and vice-versa).

"int" is not reliable due [mess with enum values](https://github.com/opencv/opencv/blob/3.4.3/modules/videoio/include/opencv2/videoio.hpp#L91-L93):
```
       CAP_VFW          = 200,          //!< Video For Windows (platform native)
       CAP_V4L          = 200,          //!< V4L/V4L2 capturing support via libv4l
       CAP_V4L2         = CAP_V4L,      //!< Same as CAP_V4L
```
so code, like this, looks right, but it doesn't work properly:
```
int backend = capture.get(CAP_PROP_BACKEND)
if (backend == CAP_V4L2)
{
    ... this code does not expect V4L here ...
    ... especially this code doesn't expect CAP_VFW on Windows platforms ...
}
```

Also "backend name" is mandatory "property".
".get()" is designed non-required optional extendable properties with unknown / untested quality.

vpisarev:
However, using String is quite error-prone. I would make getBackend() (or get(CAP_PROP_BACKEND) return the enumeration value or int and then a separate function to convert the integer code to string.

alalek:
In the future, I hope, we will add support for external plugin-based VideoCapture implementations (like OpenCL ICD, but with configurable priorities).

So "enum VideoCaptureAPIs" will become useless due absolutely custom names.

vpisarev:
:+1:

PR #12541, opened 2018-09-14
cmake: move Matlab scripts to opencv_contrib

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1774

PR #12549, opened 2018-09-15
cmake: split opencv_modules.hpp

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1776

~Merge after #12541~

New headers:
- opencv_modules.hpp
- opencv_contrib_modules.hpp

Notes:
- main purpose is avoiding mess between opencv/opencv_contrib packages.
- should resolve current issue of Ubuntu system packages (but still requires two builds)
- separated only headers; opencv_stiching still depends on opencv_xfeatures2d if OpenCV is built with contrib

```
```


mshabunin:
What if we add `HAVE_<module>` exported definition to all module targets in cmake package? So any app linked with these targets will recieve these defines. And usage of `opencv_modules.hpp` can be deprecated and recommended only for pkg-config users (should be removed from `opencv.hpp` then).

As I can see, [Ubuntu OpenCV package](https://packages.ubuntu.com/bionic/libopencv-dev) containing cmake files (libopencv-dev) depends on contrib development package (libopencv-contrib-dev). So it is no possible to use this header (via cmake) without having installed contrib modules headers.

alalek:
No sure if exploding of compiler command-line is a good idea.

> (libopencv-dev) depends on contrib development package (libopencv-contrib-dev)

This is because of this issue (but it is not the last problem). This dependency direction is not normal.

mshabunin:
>No sure if exploding of compiler command-line is a good idea.

What about adding single `HAVE_CONTRIB` definition to all contrib modules then?

mshabunin:
I think we don't need to merge this PR:
* it solves very minor Ubuntu packaging problem which can not even be reproduced
* it introduces new entity with fuzzy meaning and incomplete functionality: _OPENCV_PACKAGE_
* we don't test this functionality and probably won't do it
* it does not support builds with single-extra-module: `-DOPENCV_EXTRA_MODULES_PATH=../opencv_contrib/modules/plot` - _opencv_modules_contrib.hpp_ will not be generated and cmake will not add `OPENCV_PACKAGE_CONTRIB` definition
* it adds new build incompatibility point between _opencv_ and _opencv_contrib_

If we need to split the library into larger pieces (bundles/packages/metamodules) we have to develop more complete solution which can take in account dependencies between modules and 3rdparty libraries and can generate modular cmake packages.

alalek:
> it does not support builds with single-extra-module

This case is fixed via `target_compile_definitions()`


> we have to develop complete solution

"Developing all-at-once" activity has never been finished.

asmorkalov:
@alalek friendly reminder.

asmorkalov:
@alalek friendly reminder.

asmorkalov:
@alalek We deciced to close it, right?

asmorkalov:
@alalek Could I close the PR?

asenyaev:

PR #12559, opened 2018-09-17
Replace CV_USRTYPE1 for int64 to CV_32SC2 in Torch importer

dkurt:
```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/src/torch/torch_importer.cpp
+++ b/modules/dnn/src/torch/torch_importer.cpp
@@ -74,6 +74,18 @@ enum LuaType
     LEGACY_TYPE_RECUR_FUNCTION = 7
 };
 
+// We use OpenCV's types to manage CV_ELEM_SIZE.
+enum TorchType
+{
+    TYPE_DOUBLE = CV_64F,
+    TYPE_FLOAT  = CV_32F,
+    TYPE_BYTE   = CV_8U,
+    TYPE_CHAR   = CV_8S,
+    TYPE_SHORT  = CV_16S,
+    TYPE_INT    = CV_32S,
+    TYPE_LONG   = CV_32SC2
+};
+
 template<typename T>
 static String toString(const T &v)
 {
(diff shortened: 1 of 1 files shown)


modules/dnn/src/torch/torch_importer.cpp
```diff
@@ -257,7 +257,7 @@ struct TorchImporter
         case CV_32S:
             THFile_readIntRaw(file, (int*)storageMat.data, size);
             break;
-        case CV_USRTYPE1:
+        case CV_32SC2:
         {
             double *buf = storageMat.ptr<double>();
```
alalek: > `.ptr<double>()`

```
int64_t *buf = storageMat.ptr<int64_t>();
```

modules/dnn/src/torch/torch_importer.cpp
```diff
@@ -236,7 +236,7 @@ struct TorchImporter
     void readTorchStorage(int index, int type = -1)
     {
         long size = readLong();
-        Mat storageMat(1, size, (type != CV_USRTYPE1) ? type : CV_64F); //handle LongStorage as CV_64F Mat
+        Mat storageMat(1, size, (type != CV_32SC2) ? type : CV_64F); //handle LongStorage as CV_64F Mat
```
alalek: Check is not required here.

modules/dnn/src/torch/torch_importer.cpp
dkurt: @alalek, The thing is that it converts int64 data to doubles. I modified parser so now we use enum variables instead OpenCV's types.

modules/dnn/src/torch/torch_importer.cpp
```diff
@@ -236,36 +248,44 @@ struct TorchImporter
...
-        case CV_32S:
+        case TYPE_INT:
+            storageMat.create(1, size, CV_32S);
             THFile_readIntRaw(file, (int*)storageMat.data, size);
             break;
-        case CV_USRTYPE1:
+        case TYPE_LONG:
         {
+            storageMat.create(1, size, CV_64F);   //handle LongStorage as CV_64F Mat
             double *buf = storageMat.ptr<double>();
```
alalek: BTW, `.ptr<T>()` doesn't make type checks about `cv::Mat` elements like `.at<T>()` method do (in Debug mode).

Issue #12561, opened 2018-09-17
CUDA failing tests


cv3d:
##### System information (version)
- OpenCV => 4.0.0-pre
- Operating System / Platform => Windows 64 Bit
- Compiler => Visual Studio 2013 (Most likely independent from the compiler)
- WITH_CUDA

##### Detailed description
Several errors are related to threshold settings, while others are related to recent color conversion changes, and other stuff..
Here is a sample of the errors:
```
modules/cudev/test/test_color_cvt.cu(173): error: The max difference between matrices "dstb_gold" and "dstb" is 31 at (54, 84), which exceeds "2.0", where "dstb_gold" at (54, 84) evaluates to (8, 0, 11), "dstb" at (54, 84) evaluates to (34, 0, 42), "2.0" evaluates to 2
modules/cudev/test/test_color_cvt.cu(174): error: The max difference between matrices "dstb_gold" and "dstb" is 32 at (48, 118), which exceeds "2.0", where "dstb_gold" at (48, 118) evaluates to (13, 2, 0), "dstb" at (48, 118) evaluates to (45, 15, 0), "2.0" evaluates to 2
modules/cudev/test/test_color_cvt.cu(179): error: The max difference between matrices "dstb_gold" and "dstb" is 6 at (199, 86), which exceeds "2.0", where "dstb_gold" at (199, 86) evaluates to (0, 0, 3), "dstb" at (199, 86) evaluates to (2, 0, 9), "2.0" evaluates to 2
modules/cudev/test/test_color_cvt.cu(180): error: The max difference between matrices "dstb_gold" and "dstb" is 5 at (14, 118), which exceeds "2.0", where "dstb_gold" at (14, 118) evaluates to (1, 0, 0), "dstb" at (14, 118) evaluates to (6, 1, 0), "2.0" evaluates to 2
modules/cudaarithm/test/test_element_operations.cpp(2788): error: The max difference between matrices "x_gold" and "x" is 0.000106812 at (8, 52), which exceeds "1e-4", where "x_gold" at (8, 52) evaluates to (-150.702), "x" at (8, 52) evaluates to (-150.702), "1e-4" evaluates to 0.0001
[...]

PR #12563, opened 2018-09-17
Fix a bug in meanShiftSegmentation

cv3d:
related: #12561, #12268

### This pullrequest changes
My previous patch at #12268 was incomplete.

> Is just dst.copyTo(_dst); (without if) not enough?

> It seems it also works with copyTo(). Judging by the function code, I do not know how it can work though...

In fact it was working initially, but I got deceived by an already compiled test results (built, but forgot to install?), and thought the modifications worked as well, but it turned out it did not 😞 

```
force_builders=Custom
```

### DIFF
--- a/modules/core/src/copy.cpp
+++ b/modules/core/src/copy.cpp
@@ -238,6 +238,14 @@ void Mat::copyTo( OutputArray _dst ) const
 {
     CV_INSTRUMENT_REGION();
 
+#ifdef HAVE_CUDA
+    if (_dst.isGpuMat())
+    {
+        _dst.getGpuMat().upload(*this);
+        return;
+    }
+#endif
+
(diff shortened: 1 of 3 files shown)


cv3d:
@alalek Or shall I add the `if` case into the `copyTo()` function instead?

alalek:
> into the copyTo()

As you wish, probably under "#ifdef HAVE_CUDA" guard.

What is the current error/exception message?


(In case of `.copyTo()` fix please rebase your patch onto 3.4 branch)

cv3d:
Ok! Updating `copyTo()` instead makes it more convenient for developers.

> current error/exception message

It says "You should explicitly call download method ..."

cv3d:
@alalek All done! Your review is highly appreciated.

PR #12565, opened 2018-09-17
Remove isIntel check from deep learning layers

dkurt:
remove `OCL_PERFORMANCE_CHECK(ocl::Device::getDefault().isIntel())` from deep learning layers.

Is there a way to test this PR on non Intel GPUs?

```
force_builders=Custom
test_opencl:Custom=ON
buildworker:Custom=linux-3
```

### DIFF
--- a/modules/core/src/ocl.cpp
+++ b/modules/core/src/ocl.cpp
@@ -3078,7 +3078,7 @@ bool Kernel::run(int dims, size_t _globalsize[], size_t _localsize[],
             dims == 1 ? 64 : dims == 2 ? (i == 0 ? 256 : 8) : dims == 3 ? (8>>(int)(i>0)) : 1;
         CV_Assert( val > 0 );
         total *= _globalsize[i];
-        if (_globalsize[i] == 1)
+        if (_globalsize[i] == 1 && !_localsize)
             val = 1;
         globalsize[i] = divUp(_globalsize[i], (unsigned int)val) * val;
     }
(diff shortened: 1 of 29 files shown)


alalek:
> OCL_PERFORMANCE_CHECK(...)

Can be forced via `OPENCV_OPENCL_PERF_CHECK_BYPASS=1` variable.

> DNN checks for Intel iGPU

can be forced via `OPENCV_DNN_OPENCL_ALLOW_ALL_DEVICES=1`.

---

Both are available without OpenCV recompilation.

modules/core/src/ocl.cpp
```diff
@@ -3078,7 +3078,7 @@ bool Kernel::run(int dims, size_t _globalsize[], size_t _localsize[],
             dims == 1 ? 64 : dims == 2 ? (i == 0 ? 256 : 8) : dims == 3 ? (8>>(int)(i>0)) : 1;
         CV_Assert( val > 0 );
         total *= _globalsize[i];
-        if (_globalsize[i] == 1)
+        if (_globalsize[i] == 1 && !_localsize)
```
alalek: @dkurt What test case requires this change?

dkurt: @alalek, With gemm for small matrices.
https://github.com/opencv/opencv/blob/efceafcb5fa51142e14fc10b1e163e22888e8fa6/modules/core/test/ocl/test_gemm.cpp#L148-L161

PR #12568, opened 2018-09-17
Update seamless_cloning.cpp

sturkmen72:
### This pullrequest changes
clean up the code

### DIFF
--- a/modules/photo/src/seamless_cloning.cpp
+++ b/modules/photo/src/seamless_cloning.cpp
@@ -47,64 +47,45 @@
 using namespace std;
 using namespace cv;
 
+static Mat checkMask(InputArray _mask, Size size)
+{
+    Mat mask = _mask.getMat();
+    Mat gray;
+    if (mask.channels() == 3)
+        cvtColor(mask, gray, COLOR_BGR2GRAY);
+    else
+    {
+        if (mask.empty())
+            gray = Mat(size.height, size.width, CV_8UC1, Scalar(255));
+        else
+            mask.copyTo(gray);
+    }
+
+    return gray;
+}
+
 void cv::seamlessClone(InputArray _src, InputArray _dst, InputArray _mask, Point p, OutputArray _blend, int flags)
 {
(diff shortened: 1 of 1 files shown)


modules/photo/src/seamless_cloning.cpp
```diff
@@ -72,36 +72,12 @@ void cv::seamlessClone(InputArray _src, InputArray _dst, InputArray _mask, Point
...
-    int lenx = maxx - minx + 1;
-    int leny = maxy - miny + 1;
-
-    int minxd = p.x - lenx/2;
-    int minyd = p.y - leny/2;
-
-    Rect roi_d(minxd,minyd,lenx,leny);
-    Rect roi_s(minx,miny,lenx,leny);
+    Rect roi_s = boundingRect(gray);
+    Rect roi_d(p.x - roi_s.width / 2, p.y - roi_s.height / 2, roi_s.width, roi_s.height);
```
sturkmen72: is it better like
`Rect roi_d(p.x - roi_s.width * 0.5, p.y - roi_s.height * 0.5, roi_s.width, roi_s.height);`

modules/photo/src/seamless_cloning.cpp
```diff
@@ -72,36 +72,12 @@ void cv::seamlessClone(InputArray _src, InputArray _dst, InputArray _mask, Point
...
 
     Mat destinationROI = dest(roi_d).clone();
 
-    Mat sourceROI = Mat::zeros(leny, lenx, src.type());
+    Mat sourceROI;
     src(roi_s).copyTo(sourceROI,gray(roi_s));
```
alalek: `create()` in `copyTo()` doesn't initialize matrix values with zeros.

modules/photo/src/seamless_cloning.cpp
alalek: Rect is 'int'-based type. No need to use '0.5' here (also compiler should generate warning here due implicit cast `float -> int`).

PR #12581, opened 2018-09-18
SSE2 code in invert() replaced with universal intrinsics

terfendail:
### This pullrequest changes
SSE2 code in invert() replaced with universal intrinsics


vpisarev:
:+1:

PR #12582, opened 2018-09-18
Dockerfile for OpenCV

shivarajnaidu:
Dockerfile for OpenCV added in this pull request..

Kindly review and merge it in mainstream.. because it will be useful have one for this kind of amazing framework

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM ubuntu:18.04
+RUN apt update && apt upgrade -y
+RUN apt install wget git build-essential pkg-config cmake -y
+WORKDIR /root
+RUN git clone https://github.com/opencv/opencv.git --depth=1 && git clone https://github.com/opencv/opencv_contrib.git --depth=1
+RUN mkdir opencv/build
+WORKDIR /root/opencv/build
+RUN cmake ../ && make && make install && pkg-config --modversion opencv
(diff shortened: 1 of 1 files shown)


alalek:
No plans to add and/or maintain this "as is".
Consider moving this into OpenCV documentation (installation tutorials section).

Regarding docker file:
- missing recommended packages (like system libjpeg-dev, ffmpeg), check OpenCV tutorials
- pkg-config is deprecated and it is not supported. Use CMake.
- use multi stage docker files for build and production (nobody want image with 6Gb build folder)
- don't build under 'root'
- tests?
- OpenCL?
- Python?
- debug? development mode? volumes? usecase example?

P.S. We using docker in OpenCV CI actively.

shivarajnaidu:
Thank You for your quick response @alalek  

I don't have idea about building multistage OpenCV image.. I thought i would be useful if we can have some way to build docker image for opencv.. Can you please help me out on this.. to build Optimised docker image.. or if you already have one (as you said for CI builds you are using it) Please share those parts which would be very helpful for consumers to deploy opencv based applications..
if it is already available please share some docs realted to opencv-docker .

Thanks @alalek

shivarajnaidu:
Thank You Very much @mshabunin

vpisarev:
I think, this PR should be closed, because it's very incomplete. As suggested, this should be replaced with a tutorial (https://github.com/opencv/opencv/tree/master/doc/tutorials) on how to build and use OpenCV within Docker.

PR #12585, opened 2018-09-18
cuda: move CUDA modules to opencv_contrib

alalek:
**Merge with**: https://github.com/opencv/opencv_contrib/pull/1781

OpenCV 4.0+

The basic data structures (GpuMat etc.) will stay in the main repository.

```
```


vpisarev:
let's do it! :+1:

vmajor:
It does not compile on Ubuntu 18.04. Fails with the following message.

`n file included from /usr/include/host_config.h:50:0,
                 from /usr/include/cuda_runtime.h:78,
                 from <command-line>:0:
/usr/include/crt/host_config.h:121:2: error: #error -- unsupported GNU version! gcc versions later than 6 are not supported!
 #error -- unsupported GNU version! gcc versions later than 6 are not supported!
  ^~~~~
/media/1tb/opencv/opencv/modules/core/src/cuda/gpu_mat.cu:47:2: error: #error "opencv_cudev is required"
 #error "opencv_cudev is required"
  ^~~~~
`

PR #12587, opened 2018-09-19
ts: update valrind test filter

alalek:
- DNN: avoid very large models
- build filter per module
- fixed longTestFilter

[Validation build](http://pullrequest.opencv.org/buildbot/builders/3_4_valgrind-lin64-debug/builds/24)


alalek:
:+1:

PR #12589, opened 2018-09-19
Fix failure to request stddev of non-intrinsics

xoox:
Before this fix, the code would fail if only standard deviations of
extrinsic parameters are requested. While standard deviations matrix
should be computed if any set of standard deviations is requested. A
variable is added to represent this case.

### This pull request changes

It fixes a bug when only stddev of extrinsic parameters are requested.
To reproduce this bug, try to call `calibrateCamera` with
`stdDeviationsIntrinsics` set to `noArray()`, while
`stdDeviationsExtrinsics` set to a vector of `double`. After camera
calibration, check the elements of `stdDeviationsExtrinsics`.


vpisarev:
thanks! :+1:

PR #12592, opened 2018-09-19
dnn(test): test at least CPU code path

alalek:
```
force_builders=Custom
test_opencl:Custom=ON
buildworker:Custom=linux-2
```


dkurt:
It's actual for [test_backends.cpp](https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_backends.cpp). The main difference of these test is that backends are compared with CPU implementation twice (with flipped inputs by horizontal axis).
:+1: Looks good to merge!

PR #12604, opened 2018-09-21
convertTo/assignTo does not need a channel value

cv3d:
### This pullrequest changes
- Provide `convertTo()` and `assignTo()` with depth-only values


cv3d:
A small fraction of #12487
@vpisarev @alalek If you find it easier to review this way, then I can rewrite several parts the same way.

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -1218,20 +1218,20 @@ class CV_EXPORTS Mat
     \f[m(x,y) = saturate \_ cast<rType>( \alpha (*this)(x,y) +  \beta )\f]
     @param m output matrix; if it does not have a proper size or type before the operation, it is
     reallocated.
-    @param rtype desired output matrix type or, rather, the depth since the number of channels are the
-    same as the input has; if rtype is negative, the output matrix will have the same type as the input.
+    @param ddepth desired output matrix type or, rather, the depth since the number of channels are the
+    same as the input has; if ddepth is CV_DEPTH_AUTO, the output matrix will have the same type as the input.
```
alalek: Documentation already allows to using both type and depth in existed OpenCV versions.
So this change is questionable actually.

P.S. Allowing both is not a good decision for API if elements are same. I believe during smooth transition on type-safe enums there should be both overloads for ElemDepth and ElemType(second with additional check that `CV_MAT_CN(src.type())` == `CV_MAT_CN(rtype)`).

cv3d: Should not we update the documentation to state that depth is enough?

Issue #12569, opened 2018-09-17
Depth values ordering is broken


cv3d:
##### System information (version)
- OpenCV => 4.0.0-pre
- Operating System / Platform => All
- Compiler => All

##### Detailed description
After introducing `CV_16F`, assertions like `CV_Assert(CV_16F < CV_32F)` fails miserably...

##### Steps to reproduce
```C++
    vector<???> order { CV_8U, CV_8S, CV_16U, CV_16S, CV_16F, CV_32S, CV_32F, CV_64F };

    for (int i = 0; i < order.size(); ++i)
    {
        for (int j = 0; j < order.size(); ++j)
        {
            ASSERT_EQ(order[i] <  order[j], i <  j);
            ASSERT_EQ(order[i] <= order[j], i <= j);
            ASSERT_EQ(order[i] >  order[j], i >  j);
            ASSERT_EQ(order[i] >= order[j], i >= j);
        }
    }
```
At least among IEEE floats, in which `CV_16F < CV_32F` now returns false 😕

cv3d:
The current magic register is designed as:
```
FEDCBA9876543210
----------------
             ddd	depth
    nnnnnnnnn   	channels - 1
  r             	reserved (unset)
 c              	continuous flag
s               	submatrix flag
```

However, it is slightly inefficient, so allow me to redesign it as:
```
FEDCBA9876543210
----------------
            dddd	depth
        nnnn    	channels base - 1
     mmm        	channels shift
   r             	reserved (unset)
  1             	CV4 flag (set)
 c              	continuous flag
s               	submatrix flag
```
As you can see, we utilized the reserved bit 13 to differentiate the old and new formats.
Therefore, we will not break any user code, and we can ensure OpenCV to serve any future CV_128F possibilities as well

### Depth
16 different depth types are allowed, and the order is recovered again. Furthermore, we allow some discontinuity for future extensions, as:
```
CV_8U       = 0,
CV_8S       = 1,
CV_16U      = 3,
CV_16S      = 4,
CV_16F      = 5,
CV_32U      = 6,
CV_32S      = 7,
CV_32F      = 8,
CV_64F      = 11,
CV_USRTYPE1 = 15,
```

### Channels
By utilizing `channels = (nnnn + 1) << mmm`, we can have the following channels:
[...]

PR #12611, opened 2018-09-22
Clean CV_MAT_CN related boiler code

cv3d:
### This pullrequest changes
Cleanup for Carotene subproject


cv3d:
@alalek @mshabunin @vpisarev Help needed!

Can you please help in any of the following:
- fixing the Java boxFilter tests
- finding why ARMv7 is not passing the watershed tests
- introducing some test cases for the new types
- addressing any/all future work points listed in the PR description

modules/core/include/opencv2/core/hal/cvtype.h
```diff
@@ -0,0 +1,256 @@
...
+
+#define CV_8S        1
+#define CV_16S       3
+#define CV_32S       4
+#define CV_64S       11
+
+#define CV_16F       7
+#define CV_32F       5  /*avoid using this - order broken!*/
+#define CV_64F       6  /*avoid using this - order broken!*/
+#define CV_32F_      13 /*restore order among IEEE floats*/
```
vpisarev: since the agreement is to use CV_MAX_DEPTH() instead of max() and do not compare depths directly, we don't need these new duplicates; let's just retain `CV_32F` etc.

modules/core/include/opencv2/core/hal/cvtype.h
```diff
@@ -0,0 +1,256 @@
...
+#define CV_64F_      14 /*restore order among IEEE floats*/
+
+#define CV_8Q        17
```
vpisarev: what's that? fixed-point? there is no commonly used fixed-point types; the number of fractional bits varies a lot from one algorithm to another. Let's remove it.

cv3d: Please refer to:
- https://github.com/opencv/opencv/pull/12611/files#diff-39872d2057ab847614b5bb242a81207bR339
- https://github.com/opencv/opencv/pull/12611/files#diff-93b27ad7ebe8d837ba1a6bc86636e131R789

cv3d: Kindly let us discuss that in #12643

modules/core/include/opencv2/core/hal/cvtype.h
cv3d: @vpisarev If we are going to break numeric-based user-code for multi-channels, then why not to break numeric-based user-code for single-channels as well? It is a chance to avoid all the hacky workarounds, such as `CV_MAX_DEPTH` etc.

cv3d: @alalek @vpisarev @mshabunin Kindly consider that by this PR, user code will get broken in case of hard-coded multi-channel numeric values (e.g. `8` instead of `CV_8UC2`, or `CV_32FC2 + 8` instead of `CV_32FC3`), so why not to restore the order among depth types?

If we restore it, then will only add extra two cases in breaking user code in case of hard-coded single-channel numeric values (5 instead of `CV_32F` and 6 instead of `CV_64F`).. In my opinion, this is nothing comparable to the multi-channel case, and at any rate, users who hard code numeric values are expected to be very few to begin with.

What do you think?

cv3d:
@alalek @vpisarev Is [this failure](http://pullrequest.opencv.org/buildbot/builders/precommit_windows32/builds/11206/) related to my PR, or something else?

cv3d:
@alalek Could you please restart failed builds? The failures seems unrelated...

cv3d:
@vpisarev @alalek @mshabunin I feel extremely lucky today with all green lights, and 666 lines 😄 
Would you mind letting me know if you still require anything else for this PR inclusion?

modules/core/include/opencv2/core/hal/cvtype.h
cv3d: Any decision about Q formats?

modules/core/include/opencv2/core/hal/cvtype.h
cv3d: Any decision about swapping CV_16F with CV_32S or CV_16U?
Refer to https://github.com/opencv/opencv/issues/12569#issuecomment-425026127 for details

vpisarev:
@cv3d, your PR contains the following clause: "By merging this, you agree to properly attribute the contributor wherever you mention this feature!"

we are not ready to fulfil this requirement; we do our best to provide proper credits, but we do not promise to do it each time. either please remove it or close this PR.

PR #12612, opened 2018-09-22
samples: update winpack python samples launcher

alalek:
- accepts script parameter (allows drag & drop from 'explorer')
- use script dir instead of current dir (can launch samples from 'samples/dnn')
- added 'pause' to show error messages (about missing numpy) instead of instant closing


alalek:
3.4 branch =(

markhenryfisher:
merged and updated

PR #12615, opened 2018-09-22
findChessboardCornersSB: speed improvements

D-Alex:
findChessboardCornersSB: speed up using parallel_for and add support for flags

```
force_builders=linux,windows,macosx,docs
```

### DIFF
--- a/modules/calib3d/src/chessboard.cpp
+++ b/modules/calib3d/src/chessboard.cpp
@@ -3196,11 +3196,11 @@ bool cv::findChessboardCornersSB(cv::InputArray image_, cv::Size pattern_size,
         cv::equalizeHist(img,img);
         flags ^= CALIB_CB_NORMALIZE_IMAGE;
     }
-    if(flags & CALIB_CB_EXHAUSTING)
+    if(flags & CALIB_CB_EXHAUSTIVE)
     {
         para.max_tests = 100;
         para.max_points = std::max(500,pattern_size.width*pattern_size.height*2);
-        flags ^= CALIB_CB_EXHAUSTING;
+        flags ^= CALIB_CB_EXHAUSTIVE;
     }
     if(flags & CALIB_CB_ACCURACY)
     {
--- a/modules/calib3d/src/chessboard.cpp
+++ b/modules/calib3d/src/chessboard.cpp
@@ -623,45 +623,46 @@ void FastX::detectImpl(const cv::Mat& _gray_image,
     int num_scales = parameters.max_scale-parameters.min_scale+1;
     rotated_images.resize(num_scales);
     feature_maps.resize(num_scales);
-    for(int scale=parameters.min_scale;scale <= parameters.max_scale;++scale)
-    {
-        // calc images
-        // for each angle step
-        int scale_id = scale-parameters.min_scale;
-        cv::Mat rotated,filtered_h,filtered_v;
-        int diag = int(sqrt(gray_image.rows*gray_image.rows+gray_image.cols*gray_image.cols));
-        cv::Size size(diag,diag);
-        int num = int(0.5001*M_PI/parameters.resolution);
-        std::vector<cv::Mat> images;
-        images.resize(2*num);
-        int scale_size = int(1+pow(2.0,scale+1+super_res));
-        int scale_size2 = int((scale_size/10)*2+1);
-        for(int i=0;i<num;++i)
+    parallel_for_(Range(parameters.min_scale,parameters.max_scale+1),[&](const Range& range){
+        for(int scale=range.start;scale < range.end;++scale)
         {
-            float angle = parameters.resolution*i;
-            rotate(-angle,gray_image,size,rotated);
-            cv::blur(rotated,filtered_h,cv::Size(scale_size,scale_size2));
-            cv::blur(rotated,filtered_v,cv::Size(scale_size2,scale_size));
+            // calc images
+            // for each angle step
+            int scale_id = scale-parameters.min_scale;
+            cv::Mat rotated,filtered_h,filtered_v;
+            int diag = int(sqrt(gray_image.rows*gray_image.rows+gray_image.cols*gray_image.cols));
+            cv::Size size(diag,diag);
+            int num = int(0.5001*M_PI/parameters.resolution);
+            std::vector<cv::Mat> images;
+            images.resize(2*num);
+            int scale_size = int(1+pow(2.0,scale+1+super_res));
+            int scale_size2 = int((scale_size/10)*2+1);
+            for(int i=0;i<num;++i)
+            {
+                float angle = parameters.resolution*i;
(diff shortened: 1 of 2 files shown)


modules/calib3d/src/chessboard.cpp
```diff
@@ -3175,24 +3184,32 @@ bool cv::findChessboardCornersSB(cv::InputArray image_, cv::Size pattern_size,
...
-        break;
+        flags ^= CALIB_CB_EXHAUSTIVE;
+    }
+    if(flags & CALIB_CB_ACCURACY)
+    {
+        para.super_resolution = true;
+        flags ^= CALIB_CB_ACCURACY;
     }
+    if(flags)
+        CV_Error(Error::StsOutOfRange, "Invalid remaing flags " + std::to_string(flags));
```
alalek: GCC 4.8 (Android builder) has bad support for `std::to_string()`.

Please use this code:
```
CV_Error(Error::StsOutOfRange, cv::format("Invalid remaing flags %d", (int)flags));
```

modules/calib3d/src/chessboard.cpp
```diff
@@ -3070,60 +3068,71 @@ Chessboard::Board Chessboard::detectImpl(const Mat& gray,std::vector<cv::Mat> &f
...
-                        return Chessboard::Board();
+                    if(iter_boards->getSize().width*iter_boards->getSize().height < chessboard_size2.width*chessboard_size2.height)
+                        iter_boards->clear();
+                    else if(!parameters.larger)
+                        iter_boards->clear();
                 }
             }
+        });
+        // check if a good board was found
+        for(auto board : boards)
```
alalek: `const auto&` ?

tompollok:
@D-Alex did you check the chessboard detection rate of findChessboardCorners vs findChessboardCornersSB ?

I tested both functions and findChessboardCorners seems to outperform findChessboardCornersSB in terms of chessboard detection rate.

Check out this video: https://www.youtube.com/watch?v=ImpxJSIbBdk

I played the sequence twice. first with findChessboardCorners (red) and then again with findChessboardCorners SB (green)

I think about 30% of the chessboard detections are missing. I admit that shirt is not necessarily the best for doing calibrations, but findChessboardCorners had no problems with that.

tompollok:
I tested it with another sequence where i wear a neutral shirt and there i have the same or worse problem. In the conditions of the image below i have zero chessboard detections with or without cv::CALIB_CB_NORMALIZE_IMAGE being set.

grafik

D-Alex:
Thx for reporting problems with the detection rate and I am sure there is room for improvements. Have you tested it against the current version merged yesterday? 

I am asking because I changed MIN_RESPONSE_RATIO from 0.3F to= 0.1F which was added to speed up detections. However, this filters out points which are less than 30% of the strongest key points. And this is exactly the case in your images.

I tested your last image and it seems to work with the current version if CALIB_CB_NORMALIZE_IMAGE is added as flag. Attached is the repsonse map of the detector which looks very healthy.

feature_map_3

tompollok:
Thank you for your fast response. As mentioned i have only tested against 4.0.0-alpha. Do you know if it is possible to download master build artifacts for windows from the opencv buildservers?

alalek:
> download master build artifacts

We don't provide such builds on public.

D-Alex:
findChessboardCornersSB part of the alpha build is not really aware of the flags. This was added yesterday. Therefore, to work around the issue you can do the normalisation by yourself calling equalizeHist before calling findChessboardCornersSB. 

In addition you can provide 1 as flag to increase the number of tests. However, the alpha version does not take advantage of multiple CPUs for around half of the code base. Also you still have the hard coded 30% threshold which might be ok if you normalise your images.

tompollok:
Alright, thank you @D-Alex and @alalek for your replies.

tompollok:
cv::equalizeHist(mat, mat); before findChessboardCornersSB(...) improved the detection from 0% to about 50% for the sequence of the image of which you made a responsemap for me. So i guess ill have to recompile master with a lower MIN_RESPONSE_RATIO.

providing 1 as a flag for findChessboardCornersSB flags parameter crashes my application. Probably some assertion, but i didnt investigate.

D-Alex:
Yep - the assertion was also fixed when the flags were implemented. I forgot about it. If you upload your  sequence somewhere I can run it with the current version. Otherwise you would have to compile it using the current master.

tompollok:
I compiled master and it still crashes with 1  (CALIB_CB_ADAPTIVE_THRESH)

OpenCV(4.0.0-pre) C:\3rdparty\ocv4_master\opencv\modules\calib3d\src\chessboard.cpp:3211: error: (-211:One of arguments' values is out of range) Invalid remaing flags 1 in function 'cv::findChessboardCornersSB'

D-Alex:
Like stated in the header (I know it is hard to read when not compiled ;-)) it only supports:

CALIB_CB_NORMALIZE_IMAGE = 2
CALIB_CB_EXHAUSTING      = 16,
CALIB_CB_ACCURACY        = 32

tompollok:
I have seen the documentation but you mentioned "In addition you can provide 1 as flag to increase the number of tests." before. I also took a look at the code now and there i clearly see that 1 is not supported :-)

tompollok:
i guess you meant 16 instead of 1

tompollok:
flags = cv::CALIB_CB_NORMALIZE_IMAGE | cv::CALIB_CB_EXHAUSTIVE | cv::CALIB_CB_ACCURACY;

With flags as above the detection rate is much better but i still have many "easy" chessboards that aren't detected. If you want, i can provide you some of them.

Another question: What did you use to enable parallel_for_?

According to the link below a few options are available. Did you just enable openMP?
 
https://docs.opencv.org/trunk/d7/dff/tutorial_how_to_use_OpenCV_parallel_for_.html

The first precondition is to have OpenCV built with a parallel framework. In OpenCV 3.2, the following parallel frameworks are available in that order:

    1. Intel Threading Building Blocks (3rdparty library, should be explicitly enabled)
    2. C= Parallel C/C++ Programming Language Extension (3rdparty library, should be explicitly enabled)
    3. OpenMP (integrated to compiler, should be explicitly enabled)
    4. APPLE GCD (system wide, used automatically (APPLE only))
    5. Windows RT concurrency (system wide, used automatically (Windows RT only))
    6. Windows concurrency (part of runtime, used automatically (Windows only - MSVC++ >= 10))
    7. Pthreads (if available)

D-Alex:
Sure - I am happy to have a look why certain checkerboards are not detected. I am using a Mac therefore I would guess option 4.

tompollok:
Alright, i will send you some images during the next week.

With findChessboardCorners I used cornerSubPix to refine the corner locations. Am I assuming correctly that cornerSubPix would only degrade the corner locations from your algorithm?

PR #12623, opened 2018-09-23
Videoio 4

### LINKED ISSUE #7268: Consolidate VideoCapture methods vs apiPreference param
##### System information
- OpenCV => 3.1

It would be useful to consolidate `VideoCapture` constructors and `open` methods regards to `apiPreference` parameter. 

In 3.1 we have:

``` cpp
//from file
VideoCapture (const String &filename);
VideoCapture (const String &filename, int apiPreference);
open(const String & filename);
open(const String & filename, int apiPreference);
//from device
VideoCapture (int index)
open(int index)
```

where `index=camera_id + domain_offset (CAP_*)`

Proposal:

``` cpp
//from file
VideoCapture (const String &filename, int apiPreference=CAP_ANY);
open(const String & filename, int apiPreference=CAP_ANY);
//from device
VideoCapture (int index, int apiPreference=CAP_ANY);
open(int index, int apiPreference=CAP_ANY);
```
### Note

removing methods will break binary compatibility (ABI). Changes should be considered for next major release.
### Detailed description

**Capture from file:** Proposed changes just expose [the internal way](https://github.com/opencv/opencv/blob/3.1.0/modules/videoio/src/cap.cpp#L589)

**Capture from device:** Currently  `index=camera_id + domain_offset (CAP_*)` but we could have same scheme as capture from file just adding a bit of code in the `open` [method](https://github.com/opencv/opencv/blob/3.1.0/modules/videoio/src/cap.cpp#L592) to manage old style syntax calls index+pref... something like  this:

``` cpp
VideoCapture::VideoCapture(int index,int apiPreference)
{ open(index,apiPreference); }

bool VideoCapture::open(int index, int apiPreference)
{
    // manage old style calls index+pref
    // if both are present the apiPreference param will won
    int pref = (index / 100) * 100; //no pref => 0
    if (apiPreference == pref)
        pref = 0;
    else if (apiPreference == CAP_ANY)
        apiPreference = pref;
    index = index + apiPreference - pref;
    //old code
    if (isOpened()) release();
    icap = IVideoCapture_create(index);
    ...
```

even if changes in `IVideoCapture_create` and `cvCreateCameraCapture` should be considered

sturkmen72:
DO NOT MERGE! Work is still in progress...

### This pullrequest changes

resolves #7268


sturkmen72:
@alalek i made a mistake and lost commits. is there a way to recover lost commits?

alalek:
Here: https://github.com/alalek/opencv/commits/pr12623

sturkmen72:
thank you so much.

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -636,18 +629,19 @@ class CV_EXPORTS_W VideoCapture
...
+    CV_WRAP VideoCapture(const String& filename, int apiPreference = CAP_ANY);
 
     /** @overload
     @brief  Open a camera for video capturing
 
-    @param index camera_id + domain_offset (CAP_*) id of the video capturing device to open. To open default camera using default backend just pass 0.
-    Use a `domain_offset` to enforce a specific reader implementation if multiple are available like cv::CAP_FFMPEG or cv::CAP_IMAGES or cv::CAP_DSHOW.
-    e.g. to open Camera 1 using the MS Media Foundation API use `index = 1 + cv::CAP_MSMF`
+    @param index id of the video capturing device to open. To open default camera using default backend just pass 0.
+    (Usage of camera_id + domain_offset (CAP_*) is valid.in this case apiPreference is ignored)
```
sturkmen72: @alalek what is your opinion?

(Usage of camera_id + domain_offset (CAP_*) is valid.in this case apiPreference is ignored)

alalek: Using "domain offset" should be limited to default `apiPreference = CAP_ANY` case only.

PR #12626, opened 2018-09-24
Intel Realsense Device Example

cabelo:
rs

### This pullrequest changes

Example for capture in openCV data from the legacy Intel® RealSense™ F200, SR300, R200, LR200 and the ZR300 cameras.


alalek:
What does this demonstrate from OpenCV? Creation of `cv::Mat` from "user" data?

This is not enough for OpenCV sample.
Especially. there is RealSence backend for VideoCapture.

PR #12630, opened 2018-09-24
docs: DOXYGEN_BLACKLIST CMake variable for modules exclusion

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1792

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -14,6 +14,7 @@ if(DOXYGEN_FOUND)
   add_custom_target(doxygen)
 
   # not documented modules list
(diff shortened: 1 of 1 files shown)


alalek:
Main goal of this patch is excluding from documentation useless C++ code from bindings module.
But it is not complete/robust solution anyway - for example, it would be nice to include module's tutorials (if exist), but still exclude C++ internal code (no idea why it was placed as public headers in matlab).

So I prefer to avoid extra engineering work at this moment. This patch is simple enough.

P.S. `continue()` exists in modern CMake only. The same note about `IN_LIST`.

PR #12637, opened 2018-09-24
Fixes for instrumentation of IPP and OCL

savuor:
### This pullrequest changes

 * fixed warning about variable redefinition when both IPP and instrumentation are enabled
 * fixed segfault when no `funName` provided for `NodeData` struct
 * compilation fixed when both OCL and instrumentation are enabled

### DIFF
--- a/modules/core/src/ocl.cpp
+++ b/modules/core/src/ocl.cpp
@@ -3091,7 +3091,7 @@ bool Kernel::run(int dims, size_t _globalsize[], size_t _localsize[],
 bool Kernel::Impl::run(int dims, size_t globalsize[], size_t localsize[],
         bool sync, int64* timeNS, const Queue& q)
 {
-    CV_INSTRUMENT_REGION_OPENCL_RUN(name.c_str(););
+    CV_INSTRUMENT_REGION_OPENCL_RUN(name.c_str());
 
     if (!handle || isInProgress)
         return false;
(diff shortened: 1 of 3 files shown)


alalek:
=> 3.4 branch

savuor:
@alalek done

modules/core/src/system.cpp
```diff
@@ -1834,7 +1834,7 @@ FLAGS getFlags()
 #endif
 }
 
-NodeData::NodeData(const char* funName, const char* fileName, int lineNum, void* retAddress, bool alwaysExpand, cv::instr::TYPE instrType, cv::instr::IMPL implType)
+NodeData::NodeData(const cv::String funName, const char* fileName, int lineNum, void* retAddress, bool alwaysExpand, cv::instr::TYPE instrType, cv::instr::IMPL implType)
 {
     m_funName       = funName;
```
alalek: Instrumentation should have minimal performance impact.
`const char*` is enough in parameters (it designed for using of static const string literals only).
It is better to fix this "assignment" statement (`std::string` doesn't allow to pass NULL).
```
m_funName       = funName ? funName : "";
```

modules/core/src/ocl.cpp
```diff
@@ -3091,7 +3091,7 @@ bool Kernel::run(int dims, size_t _globalsize[], size_t _localsize[],
 bool Kernel::Impl::run(int dims, size_t globalsize[], size_t localsize[],
         bool sync, int64* timeNS, const Queue& q)
 {
-    CV_INSTRUMENT_REGION_OPENCL_RUN(name.c_str(););
+    CV_INSTRUMENT_REGION_OPENCL_RUN(name.c_str());
```
alalek: https://github.com/opencv/opencv/pull/12530/files#diff-ff67959a252bb213e2340f8557e16e84R3094

/cc @cv3d

cv3d: 😢 I am sorry!

PR #12647, opened 2018-09-25
build: cleanup js scripts

alalek:
- avoid unnecessary 2 messages for other build configurations
    ```
    -- ${EMSCRIPTEN_INCLUDE_DIR}/emscripten/bind.h` was not detected
    -- Module js disabled because the following dependencies are not found: Emscripten
    ```

```
force_builders=Docs,Custom
docker_image:Docs=docs-js
```


vpisarev:
:+1:

PR #12658, opened 2018-09-26
hide use of std::mutex from /clr compilation under Visual Studio

### LINKED ISSUE #11912: OpenCV 3.4.1 with CLR
- OpenCV => 3.4.1

##### Detailed description

I can not compile quite simple OpenCV project with CLR on Visual Studio 2017. It gives me "< mutex > is not supported when compiling with /clr or /clr:pure." error.

##### Steps to reproduce

- Start a new empty C++ project on Visual Studio 2017. 
- Try to read and show an image.
- Go to project properties -> Configuration Properties -> General -> Common Language Runtime Support -> Select Common Language Runtime Support (/CLR)
- Try to build.

chacha21:
C++11 `<mutex>` is not available when compiling with  /clr under Visual Studio, thus opencv cannot be easily included.
It is fixed by making a CEEMutex wrapper class, around an opaque implementation using std::mutex internally

resolves #11912

### DIFF
--- a/modules/core/include/opencv2/core/utility.hpp
+++ b/modules/core/include/opencv2/core/utility.hpp
@@ -676,32 +676,6 @@ void Mat::forEach_impl(const Functor& operation) {
 
 /////////////////////////// Synchronization Primitives ///////////////////////////////
 
-class CV_EXPORTS CEEMutex
-{
-  public:
-    CEEMutex(void);
-    ~CEEMutex();
-    CEEMutex(const CEEMutex&) = delete;
-    CEEMutex& operator=(const CEEMutex&) = delete;
-  public:
-    void lock(void);
-    void unlock(void);
-    bool try_lock(void) noexcept;
-  private:
-    void* impl;
-};
-
-class CV_EXPORTS CEELockGuard
-{
-  public:
-    explicit CEELockGuard(CEEMutex& _Mtx);
-    ~CEELockGuard();
-    CEELockGuard(const CEELockGuard&) = delete;
-    CEELockGuard& operator=(const CEELockGuard&) = delete;
-  private:
-    void* impl;
-};
-
 #if !defined(_M_CEE)
 typedef std::recursive_mutex Mutex;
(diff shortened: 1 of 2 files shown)


modules/core/include/opencv2/core/utility.hpp
```diff
@@ -674,8 +676,39 @@ void Mat::forEach_impl(const Functor& operation) {
...
+	  CEELockGuard& operator=(const CEELockGuard&) = delete;
+  private:
+    void* impl;
+};
+
+#if !defined(_M_CEE)
 typedef std::recursive_mutex Mutex;
 typedef std::lock_guard<cv::Mutex> AutoLock;
+#else
+typedef CEEMutex Mutex;
```
vpisarev: Can we just put custom definitions of Mutex and AutoLock here instead of introducing proxy classes CEEMutex and CEELockGuard?

chacha21: I don't think so. CEEMutex and CEELockGuard must be compiled anyway, so that the dll built without /clr will still work when linked to a /clr program including opencv headers

alalek: What API requires Mutex/AutoLock emulation?

chacha21: KMeansDistanceComputer

alalek: Looks like, mutex is misused there. Removed: #12660

chacha21: Indeed, when #12660 is commited, cv::Mutex can just be made unavailable from /clr builds (and CEEMutex/CEELockGuard can be removed)
The only drawback is that under a /clr build, the developer won't be able to just copy/paste valid code using cv::Mutex or cv::AuLock, and will have to rely on a custom implementation of his own

alalek: I believe CLR assumes using of own mechanics for synchronization instead of raw Mutex.

PR #12663, opened 2018-09-27
Update cv::findContours()

### LINKED ISSUE #8225: Update param type of input image in cv::findContours
#### System information (version)
- OpenCV => 3.2

#### Detailed description

Since opencv 3.2 source image is not modified by this function. See [master doc](http://docs.opencv.org/master/d3/dc0/group__imgproc__shape.html#ga17ed9f5d79ae97bd4c7cf18403e1689a) and #7516. The param type should reflect this change.

Now  `void cv::findContours( InputOutputArray _image, ...`
Better `void cv::findContours( InputArray _image, ...`

I know this would break ABI compatibility but may be in next release....

BTW it should exists great evidence of this update because `findContours`  is widely used, the 3.2 doc isn't aligned to the code and currently all users clone the input image while this is a complete of wast time.

sturkmen72:
resolves #8225

### DIFF
--- a/doc/py_tutorials/py_imgproc/py_contours/py_contour_features/py_contour_features.markdown
+++ b/doc/py_tutorials/py_imgproc/py_contours/py_contour_features/py_contour_features.markdown
@@ -23,7 +23,7 @@ import cv2 as cv
 
 img = cv.imread('star.jpg',0)
 ret,thresh = cv.threshold(img,127,255,0)
-im2,contours,hierarchy = cv.findContours(thresh, 1, 2)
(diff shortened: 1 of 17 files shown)


berak:
oh, seems, you've hit a "wasp nest" here ;(

since it's no more an InputOutputArray, it won't get returned from the python bindings any more, so it is:

     contours, hierarchy = cv.findContours(....)

a [lot of things need to be changed now](https://github.com/opencv/opencv/search?q=cv.findContours&unscoped_q=cv.findContours)

PR #12665, opened 2018-09-27
Restore depth ordering among IEEE floats

### LINKED ISSUE #12569: Depth values ordering is broken

cv3d:
resolves #12569

Introducing half-precision float is great, but it should not have higher value that its flat and double sisters.
This would unnecessarily complicate the system.

### This pullrequest changes
- Swap CV_16U and CV_16F values

Side effects should be be minimal and tolerable for the following reasons:
- It is rare to hard-code 2 instead of using CV_16U
- CV_16U is among the least famous depth values (compared to CV_8U or CV_32F etc)
- CV_16U and CV_16F both reserves two bytes
- CV_16F is just introduced recently


cv3d:
There are some failed tests that I just cannot bother to fix given the high possibility of rejecting this PR... I will re-discuss the issue instead.

PR #12666, opened 2018-09-27
cudev: fix test failure of cudev

tomoaki0705:
* follow the implementation of Luv2RGBfloat in imgproc/src/color_lab.cpp
  * related #12561

### This pullrequest changes
  * since the implementation of CPU side changed in #9226 , the ```opencv_test_cudev``` was failing
  * the difference was 30+ when the eps was 2.
  * still, it rarely happens, but when converting the Luv to RGB colorspace, there is a gap of gradation, and the error was happening around this gap.
  * expanding eps was one choice (as commented in #12561 ) but I managed to port the implementation of imgproc to cudev.
  * once this modification is approved, I'll open a PR for opencv_contrib, too.

```
force_builders=Custom
```


tomoaki0705:
Thank you alalek
I updated the PR

PR #12668, opened 2018-09-27
calib3d: findChessboardCornersSB() minor updates

alalek:
- avoid updating of input image during equalizeHist() call
- avoid for() with double variable (use 'int' instead)
- more CV_Check*() macros
- use Mat_<T>, Matx
- static for local variables

<cut/>

```
buildworker:Linux x64 Debug=linux-2
```

### DIFF
--- a/modules/calib3d/src/chessboard.cpp
+++ b/modules/calib3d/src/chessboard.cpp
@@ -10,7 +10,7 @@
 //#define CV_DETECTORS_CHESSBOARD_DEBUG
 #ifdef CV_DETECTORS_CHESSBOARD_DEBUG
 #include <opencv2/highgui.hpp>
-cv::Mat debug_image;
+static cv::Mat debug_image;
 #endif
 
 using namespace std;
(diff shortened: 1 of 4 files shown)


alalek:
/cc @D-Alex 

BTW, I'm observing strange results if `CALIB_CB_NORMALIZE_IMAGE` is "forced" in `findChessboardCornersSB()` (with and without other changes from this PR): https://github.com/alalek/opencv/commit/calib3d_chessboardsb_updates_test
- chessboard.cpp:3190: `flags |= CALIB_CB_NORMALIZE_IMAGE;`
- enable `#if 0` debug code in tests (see patch)
- Run: `./bin/opencv_test_calib3d --gtest_filter=*Calib3d_ChessboardDetector2.accuracy*`

alalek:
OpenCV 4.0 gold release is planned in the end of October.

This is non-default mode, looks like `equalizeHist()` introduces some artifacts. May be other pre-processing like `adaptiveThreshold()` should work better on these synthetic images.

Do you have some objectives about this patch?

modules/calib3d/src/chessboard.cpp
```diff
@@ -207,15 +202,16 @@ bool isPointOnLine(cv::Point2f l1,cv::Point2f l2,cv::Point2f pt,float min_angle)
...
 {
     cv::Rect image_rect(int(0.5*dist),int(0.5*dist),int(mat.cols-0.5*dist),int(mat.rows-0.5*dist));
     cv::Size size(int(0.5*dist),int(0.5*dist));
     int count = 0;
     cv::Mat patch1,patch2;
     cv::Point2f center1,center2;
-    for(double angle=0;angle <= CV_PI;angle+=CV_PI*0.1)
+    for (int angle_i = 0; angle_i < 10; angle_i++)
     {
+        double angle = angle_i * (CV_PI * 0.1);
```
alalek: @D-Alex Is this loop still correct? (`<=` condition has been replaced to `<` to get exact 10 "tests")

D-Alex: Looks ok. PI was already tested during the first iteration and should not be tested twice like in the previous implementation.

alalek:
:+1:

PR #12670, opened 2018-09-27
 imgproc: use Matx23d as result type of getRotationMatrix2D() call

alalek:
TODO:
- [x] validate/fix bindings

### DIFF
--- a/modules/calib3d/src/chessboard.cpp
+++ b/modules/calib3d/src/chessboard.cpp
@@ -282,9 +282,9 @@ void FastX::rotate(float angle,const cv::Mat &img,cv::Size size,cv::Mat &out)con
     }
     else
     {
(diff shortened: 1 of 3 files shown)


vpisarev:
@alalek, in my opinion, this is the example of function where breaking compatibility is both noticeable (and thus disturbing) and unnecessary.

I suggest to keep `getRotationMatrix2D()` as-is (same name, same API) and then add another function:

a) add overloaded `void getRotationMatrix2D(Point2f center, double angle, double scale, OutputArray R);` where you can pass `Mat`, `Matx23d`, `Matx23f` etc. This is the most flexible and my preferable option.
b) add `Matx23d getRotationMatrix2D_(Point2f center, double angle, double scale);`

Those few people who need this tiny improvement in performance in this case will easily replace one call with another. Those other 99.9% who find the current function just fine, won't have to do any changes to stay compatible with the latest OpenCV.

alalek:
Right!

Also this exercise shows how important to provide reliable interfaces from the beginning.

PR #12673, opened 2018-09-27
Fix build warnings

alalek:
Builds: [1](http://pullrequest.opencv.org/buildbot/builders/master_etc-fedora-tbb-lin64/builds/11) [2](http://pullrequest.opencv.org/buildbot/builders/master_opt-avx2_noICV-skl-lin64/builds/11)

Validated: [1](http://pullrequest.opencv.org/buildbot/builders/master_etc-fedora-tbb-lin64/builds/12) [2](http://pullrequest.opencv.org/buildbot/builders/master_opt-avx2_noICV-skl-lin64)

### DIFF
--- a/modules/core/include/opencv2/core/cuda.hpp
+++ b/modules/core/include/opencv2/core/cuda.hpp
@@ -134,8 +134,8 @@ public:
     CV_WRAP GpuMat(const GpuMat& m);
 
     //! constructor for GpuMat headers pointing to user-allocated data
-    CV_WRAP GpuMat(int rows, int cols, int type, void* data, size_t step = Mat::AUTO_STEP);
-    CV_WRAP GpuMat(Size size, int type, void* data, size_t step = Mat::AUTO_STEP);
+    GpuMat(int rows, int cols, int type, void* data, size_t step = Mat::AUTO_STEP);
+    GpuMat(Size size, int type, void* data, size_t step = Mat::AUTO_STEP);
 
     //! creates a GpuMat header for a part of the bigger matrix
     CV_WRAP GpuMat(const GpuMat& m, Range rowRange, Range colRange);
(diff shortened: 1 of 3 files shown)


modules/core/include/opencv2/core/cuda.hpp
```diff
@@ -134,8 +134,8 @@ class CV_EXPORTS_W GpuMat
     CV_WRAP GpuMat(const GpuMat& m);
 
     //! constructor for GpuMat headers pointing to user-allocated data
-    CV_WRAP GpuMat(int rows, int cols, int type, void* data, size_t step = Mat::AUTO_STEP);
-    CV_WRAP GpuMat(Size size, int type, void* data, size_t step = Mat::AUTO_STEP);
+    GpuMat(int rows, int cols, int type, void* data, size_t step = Mat::AUTO_STEP);
+    GpuMat(Size size, int type, void* data, size_t step = Mat::AUTO_STEP);
```
alalek: /cc @cv3d

cv3d: It is shame I caused such mess! Thanks for the cleanup.
BTW: It makes sense to disable args, while taking in consideration that we need to return `void` in order to support UMat wrapper.

vpisarev:
:+1:

pwuertz:
Just bumped into the assertion error for `void*` being a forbidden argument type. I'm trying to get interoperability between `GpuMat` and CUDA frameworks like Numba and CuPy.

With `void*` being forbidden, what is the preferred way for providing raw pointers to `CV_WRAP` functions?

I tried adding a helper method for passing `data` as `size_t` number, but `gen2.py` generally maps `size_t` to `PyArg_Parse...()` type `I` (32bit), thus not preserving 64bit values.

After fixing this manually, interoperability works like a charm. It just seems to be a matter choice for handling pointers in the binding generator.

PR #12693, opened 2018-09-29
Obtain fourcc when `AVStream` is network stream

peters:
### This pullrequest changes

The `codec_tag` is only available when opening a file from disk. If `AVStream` is a network stream then `fourcc` must be obtained using `codec_id`. I have tested the following scenarios:
1) Open a `.mp4` file and verify that `codec_tag` is returned (old behavior)
2) Open a `rtsp` stream and verify that `codec_fourcc` is returned (Tested with a MJPEG, H264 and H265 stream)

```
force_builders=Custom
docker_image:Custom=ffmpeg-master
```

### DIFF
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -48,6 +48,7 @@
 #include <algorithm>
 #include <limits>
 
+#define OPENCV_FOURCC(c1, c2, c3, c4) (((c1) & 255) + (((c2) & 255) << 8) + (((c3) & 255) << 16) + (((c4) & 255) << 24))
 #define CALC_FFMPEG_VERSION(a,b,c) ( a<<16 | b<<8 | c )
 
 #if defined _MSC_VER && _MSC_VER >= 1200
@@ -349,6 +350,41 @@ struct AVInterruptCallbackMetadata
     unsigned int timeout_after_ms;
     int timeout;
 };
+
+// https://github.com/opencv/opencv/pull/12693#issuecomment-426236731
+static
+inline const char* _opencv_avcodec_get_name(AVCodecID id)
+{
+#if LIBAVCODEC_VERSION_MICRO >= 100 \
+    && LIBAVCODEC_BUILD >= CALC_FFMPEG_VERSION(53, 47, 100)
+    return avcodec_get_name(id);
+#else
+    const AVCodecDescriptor *cd;
+    AVCodec *codec;
+
+    if (id == AV_CODEC_ID_NONE)
+    {
+        return "none";
+    }
+    cd = avcodec_descriptor_get(id);
+    if (cd)
+    {
+        return cd->name;
+    }
+    codec = avcodec_find_decoder(id);
+    if (codec)
+    {
+        return codec->name;
+    }
+    codec = avcodec_find_encoder(id);
+    if (codec)
+    {
+        return codec->name;
+    }
+
+    return "unknown_codec";
+#endif
+}
 
 static
 inline void _opencv_ffmpeg_free(void** ptr)
@@ -1121,6 +1157,10 @@ double CvCapture_FFMPEG::getProperty( int property_id ) const
 {
     if( !video_st ) return 0;
 
+    double codec_tag = 0;
+    AVCodecID codec_id = AV_CODEC_ID_NONE;
+    const char* codec_fourcc = NULL;
+
     switch( property_id )
     {
     case CV_FFMPEG_CAP_PROP_POS_MSEC:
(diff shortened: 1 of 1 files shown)


alalek:
@peters Please remove mention symbol from the commit message: "@alalek" - to avoid heavy spam from GitHub.

As a "bugfix" this patch should [go](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) into 3.4 branch. We will merge changes from 3.4 into master regularly (weekly/bi-weekly).

So, please:

peters:
@alalek Thanks for the detailed guide :) I've removed the (at) mention as you requested.

alalek:
Please remove mention from the **commit** message. PR description is fine (but still unnecessary, I'm subscribed on all notifications about opencv project).

peters:
There appears to be a problem with the build. The involved method (according to the blame history) has existed for at least six years. Any pointers @alalek?

alalek:
Failed builder is based on Ubuntu 14.04. This Ubuntu version uses libav instead of ffmpeg.

Unfortunately [`doc/APIchanges`](https://github.com/FFmpeg/FFmpeg/blob/master/doc/APIchanges) is empty about `avcodec_get_name()` call.

I can't find `avcodec_get_name()` call in libav: https://github.com/libav/libav/blob/master/libavcodec/avcodec.h
So it is ffmpeg-only thing.

FFmpeg added `avcodec_get_name()` in [this commit](https://github.com/FFmpeg/FFmpeg/commit/b3be9f4a88787d2d6692ffced187db677afdc726#diff-125aa97c502e43e14a581f5a77c59e8fR3440).
There is no special version bump on this commit. Also there is no "100" micro version (to diff libav / ffmpeg) - current avcodec version is [53.11.0](https://github.com/FFmpeg/FFmpeg/blob/b3be9f4a88787d2d6692ffced187db677afdc726/libavcodec/version.h).

The first commit with "100" and `avcodec_get_name()` is [here](https://github.com/FFmpeg/FFmpeg/commit/f0e90e00354f5705e132406a06d3360e2c00565d): .

Proper check to use this function is:
```
#if LIBAVCODEC_VERSION_MICRO >= 100 \  // FFmpeg (exclude libav)
    && LIBAVCODEC_BUILD >= CALC_FFMPEG_VERSION(53, 47, 100)
... use avcodec_get_name() ...
#else
... don't use
... return -1 ?
... or reuse fallback implementation:
    codec = avcodec_find_decoder(id);
    if (codec)
        codec_fourcc = codec->name;
#endif
```

BTW, What is about `codec_fourcc = video_st->codec->name;` ? Does it work?

peters:
@alalek Thanks for the detailed explanation. I added up with adding a inline helper method `_opencv_avcodec_get_name`. The build succeeds now, but there is a white space issue I can't explain.

peters:
It would be nice if we could require `ffmpeg 4.0` in the upcoming opencv `4.0` release. It would make it much easier maintaining the code. As it reads right now, it's very complex to reason about.

alalek:
[Here](http://pullrequest.opencv.org/buildbot/builders/precommit_docs/builds/17726/steps/whitespace%20opencv/logs/stdio) two "empty" lines with spaces:
```
modules/videoio/src/cap_ffmpeg_impl.hpp:1159: trailing whitespace.
+    
modules/videoio/src/cap_ffmpeg_impl.hpp:1193: trailing whitespace.
+        
```

Please remove spaces from these lines.


> require ffmpeg 4.0 in the upcoming opencv 4.0 release

We usually re-using system FFmpeg on Linux systems (or libav) to minimize OpenCV build prerequisites. Building it before OpenCV is not a real option.

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -1139,10 +1179,25 @@ double CvCapture_FFMPEG::getProperty( int property_id ) const
...
+            return codec_tag;
+        }
+
+        codec_fourcc = _opencv_avcodec_get_name(codec_id);
+        if(!codec_fourcc)
+        {
+            return codec_tag;
+        }
+
+        return (double) OPENCV_FOURCC(codec_fourcc[0], codec_fourcc[1], codec_fourcc[2], codec_fourcc[3]);
```
alalek: Please add `strlen(codec_fourcc) >= 4` check before this line.

peters: Done ✔️

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -1139,10 +1179,25 @@ double CvCapture_FFMPEG::getProperty( int property_id ) const
...
+        codec_tag = (double)video_st->codec.codec_tag;
 #endif
+
+        if(codec_tag || codec_id == AV_CODEC_ID_NONE)
+        {
+            return codec_tag;
+        }
+
+        codec_fourcc = _opencv_avcodec_get_name(codec_id);
+        if(!codec_fourcc || strlen(codec_fourcc) < 4 || strcmp(codec_fourcc, "unknown_codec") == 0)
```
peters: I also added a check that if "unknown_codec" is returned then we default to existing `codec_tag` which in this case is empty.

PR #12697, opened 2018-09-29
Faster Cocoa Window Drawing

tellowkrinkle:
### Faster Cocoa Window Drawing

This PR should speed up window drawing on retina Macs.  On my computer, displaying 1000 1080p images went from taking 130 seconds to 40, which still isn't great but is a lot better than before.

It also changes the default size of windows (and size of autosize windows) to the size of the image in pixels (instead of points), meaning that windows will default to 1/2 the size that they previously did on retina macs.

### DIFF
--- a/modules/highgui/src/window_cocoa.mm
+++ b/modules/highgui/src/window_cocoa.mm
@@ -83,6 +83,7 @@ static NSMutableDictionary *windows = nil;
 static bool wasInitialized = false;
 
 @interface CVView : NSView
+@property(retain) NSView *imageView;
 @property(retain) NSImage *image;
 @property int sliderHeight;
 - (void)setImageData:(CvArr *)arr;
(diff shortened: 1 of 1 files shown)


vpisarev:
@tellowkrinkle, thank you for the patch! Can you please eliminate extra whitespaces that our buildbot has found: http://pullrequest.opencv.org/buildbot/builders/precommit_docs/builds/17670 ?

vpisarev:
@tellowkrinkle, thank you very much for the patch! and sorry for a loong delay with review.

I've just checked it on macbook pro with retina display. Indeed, it works well and fast, and looks nice!

But there is issue - when the displayed image is small (e.g. try to run the squares or morphology2 or delaunay2 or convexhull examples), I would say very small.

I see 2 possible solutions of the problem:
1. very simple one - do not use `CV_WINDOW_AUTOSIZE` flag by default. For that I temporary patched cvShowImage() and cvNamedWindow() - set window flags to 0 instead of `CV_WINDOW_AUTOSIZE`. That let's user to adjust the window size and get a larger image. BTW, I've noticed one glitch with this solution. After you adjust the window size, there is 1-3 seconds delay until the window reacts on keys and slider moves.

2. more automatic one, but I frankly speaking don't know how to implement it, my knowledge of Cocoa is very sparse and obsolete. We could set some scale-factor for the image, depending on the display resolution (dpi) and the image size. E.g. in the case of small images (like 512x512) when they are displayed on macbook pro (or macbook 12), the scale factor would be ~2. Of course, in the case of large images, like fullhd or something like that, the scale factor should be 1.

Probably, we can even combine 1 and 2 - set the initial scale-factor depending on dpi and image size, and also let users to adjust the size.

vpisarev:
@tellowkrinkle, thank you; now it's really good! I have no more complains :+1:

PR #12703, opened 2018-09-30
dnn: Add a Vulkan based backend

wzw-intel:
This commit adds a new backend "DNN_BACKEND_VKCOM" and a
new target "DNN_TARGET_VULKAN". VKCOM means vulkan based
computation library.

This backend uses Vulkan API and SPIR-V shaders to do
the inference computation for layers. The layer types
that implemented in DNN_BACKEND_VKCOM include:
Conv, Concat, ReLU, LRN, PriorBox, Softmax, MaxPooling,
AvePooling, Permute

This is just a beginning work for Vulkan in OpenCV DNN,
more layer types will be supported and performance
tuning is on the way.

### **Build OpenCV DNN with Vulkan support** 
Due to usage of dynamic load mechanism, there is no compile-time dependency on Vulkan Loader Library or vendor driver. You just need add "-DWITH_VULKAN=ON" into cmake options to enable the DNN Vukand support.

### **Enable DNN Vulkan runtime support**

1. Install Vulkan Loader library
   You have 2 choices to install Vulkan Loader library:
   1.1 Install the library from Vulkan SDK
       Download Vulkan SDK from https://vulkan.lunarg.com/sdk/home#sdk/downloadConfirm/1.1.85.0/linux/vulkansdk-linux-x86_64-1.1.85.0.tar.gz
       'mkdir ~/vulkan_sdk'
       'tar zxf vulkansdk-linux-x86_64-1.1.85.0.tar.gz -C ~/vulkan_sdk'
       Your library is at "$HOME/vulkan_sdk/1.x.x.x/x86_64/libvulkan.so.1"

   1.2 Install the library by using apt1. 
        'sudo apt-get install libvulkan1'
        Your library will be loacated at  "/usr/lib/x86_64-linux-gnu/libvulkan.so.1"

2. Install vendor driver (For Intel and AMD GPU)
   Also 2 choices:
   2.1 Compiling and installing Mesa
        Ref https://www.mesa3d.org/install.html for HOWTO.
        Additionally, add option "--with-vulkan-drivers=intel" when doing configure to build intel Vulkan driver.
        "--with-vulkan-drivers=radeon" for AMD GPU.
   
   2.2 Install the library from apt
         'sudo apt-get install mesa-vulkan-drivers'

3. Install Vulkan utils (optional)
   If you have installed Vulkan SDK follow 1.1, you already got the utils.
   'source  $HOME/vulkan_sdk/1.x.x.x/setup-env.sh' to add the utils path into $PATH
   You can also install the vulkan utils by 'sudo apt-get install vulkan-utils'

4. Verify your Vulkan installation (optional)
    run 'vulkaninfo' and if you can see output string like "GPU id       : 0 (Intel(R) xxxxxx", that means Vulkan works.
 
### **Run test**
   Before run test, if you installed Vulkan Loader library as 1.1,
   you need 'export OPENCV_VULKAN_RUNTIME=$HOME/vulkan_sdk/1.x.x.x/x86_64/libvulkan.so.1'
   run '$OPENCV_BUILD_DIR/bin/opencv_test_dnn'
   The results with "VKCOM/VULKAN" suffix are for our backend. 

### **How to patch GLSL shaders**
The GLSL shader source code are in $OPENCV_SOURCE_DIR/module/dnn/src/vkcom/shader/, with suffix ".comp". 

Once you modify the shader source code or add new shaders, go into $OPENCV_SOURCE_DIR/module/dnn/src/vkcom/shader/ and run 'python spirv_generator.py' to compile them into SPIR-V format and attach the binary into .cpp
files. And rebuild the DNN module.

In order to run this script successfully, you need tool "glslangValidator" which is provied by Vulkan SDK.

```
force_builders=Custom
docker_image:Custom=ubuntu-vulkan:16.04
buildworker:Custom=linux-4
```


wzw-intel:
@vpisarev, thanks for your comments.

Shader source code are in the dnn/src/vkcom/shader directory and with suffix ".comp", but the vkcom backend will directly use the bianry format of shader instead of compiling the .comp on the runtime.

The reason is Vulkan driver don't cover the compiling from glsl source code to SPIR-V binary. If we want this functionality, 3 choices:
1. On the build time, use tool "glslangValidator", provided by Vulkan SDK, to make this compilation. This method introduce dependency on tool "glslangValidate".  In this PR, to avoid this denpendency, I provide the pre-built binaries in the source code.  

2. Use 3rd party helper library, like libshaderc (https://github.com/google/shaderc/tree/master/libshaderc). This introduce additional dependency.

3. Copy source code of 3rd party library into vkcom, so that we have no dependency on additional library and can build the shader on the runtime.

What is your opinion?

jljusten:
@wzw-intel: I don't work on opencv, so this might not be accepted by the project. But, maybe you could update opencv's cmake to attempt to detect if glslang is available. If it is available, it can enable vulkan support and use glslang to compile the spir-v source files. If it is not available, then the vulkan support could be disabled.

vpisarev:
@wzw-intel, we discussed it a little bit. The option 2 + option 3 seems to be preferable for us. That is, libshaderc (which has Apache 2 license and therefore it's fine to put it into OpenCV) can be put into opencv/3rdparty directory, get compiled and used.

The questions are - is this library enough? Since it's a compiler, does this library depends on LLVM? (I do not see such dependency in the CMakeLists.txt). In general, what's the scheme?

Ideally, we or our users should be able to make fixes directly in .comp so that the fixed files are automatically re-compiled and included into OpenCV DNN. Otherwise, the procedure of patching .comp files will be very complicated.

wzw-intel:
@vpisarev 
Thanks for your explanation for the "ideal situation with OpenCL acceleration".
I will cook a new commit like what the OpenCL acceleration do.

wzw-intel:
Hi @alalek, I pushed an update to fix can't find vulkan package issue.

You can follow below steps to test the vulkan DNN patch (take Ubuntu 16.04 as an example):
1. Install Vulkan SDK (driver loader)
Download Vulkan SDK from https://vulkan.lunarg.com/sdk/home#sdk/downloadConfirm/1.1.85.0/linux/vulkansdk-linux-x86_64-1.1.85.0.tar.gz
mkdir ~/vulkan_sdk
tar zxf vulkansdk-linux-x86_64-1.1.85.0.tar.gz -C ~/vulkan_sdk

2. Install mesa vulkan driver
sudo apt-get install mesa-vulkan-drivers

3. Build OpenCV with DNN vulkan backend and run the test
source ~/vulkan_sdk/1.1.85.0/setup-env.sh
cd $OPENCV_DIR
mkdir build && cd build
cmake -DWITH_VULKAN=ON ..
make -j8
./bin/opencv_test_dnn --gtest_filter=*VkCom*

As you can see, this version has dependency on Vulkan runtime (libvulkan.so) for both compile-time and run-time, and can't fallback to CPU path if can't detect Vulkan. I will fix these issues in next update by using dynamic load method.

wzw-intel:
Hi @vpisarev @alalek
I pushed the commits that fixed the dependency issue and shader source code patching issue.
I also update the PR message to add "HOWTO".
Please checkout it out. Any comments are appriciated.

cmake/OpenCVDetectVulkan.cmake
```diff
@@ -0,0 +1,17 @@
+set(VULKAN_INCLUDE_DIRS "${OpenCV_SOURCE_DIR}/modules/dnn/src/vkcom/" CACHE PATH "Vulkan include directory")
```
alalek: Please move 3rdparty headers (original from Khronos) into "<opencv>/3rdparty/include/vkcom".

cmake/OpenCVDetectVulkan.cmake
```diff
@@ -0,0 +1,17 @@
...
+      OUTPUT_VARIABLE TRY_OUT
+      )
+if(NOT ${VALID_VULKAN})
+  message(WARNING "Can't use Vulkan")
+  return()
+endif()
+
+set(HAVE_VULKAN 1)
+add_definitions(-DVK_NO_PROTOTYPES)
+include_directories(${VULKAN_INCLUDE_DIRS})
```
alalek: Perhaps we should not add this globally.

In `dnn/CMakeLists.txt`:
```
if(HAVE_VULKAN)
  add_definitions(-DVK_NO_PROTOTYPES)
  include_directories(${VULKAN_INCLUDE_DIRS})
endif()
```

modules/dnn/src/vkcom/shader/concat_spv.cpp
```diff
@@ -0,0 +1,83 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2017, Intel Corporation, all rights reserved.
```
alalek: Please fix copyright year.

wzw-intel:
Hi @alalek, I fixed the issues you mentioned in your review comments, plase check.

vpisarev:
@wzw-intel, big thanks! Excellent job!

modules/dnn/test/test_vkcom_layer.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+#include "test_precomp.hpp"
+
+namespace opencv_test { namespace {
+
+using namespace cv;
+using namespace cv::dnn;
+using namespace testing;
+
+static void test(Mat& input, Net& net, Backend backendId, Target targetId, bool skipCheck = false)
```
dkurt: I think it's better to add Vulkan target tests to existing ones. Please see
[test_backends.cpp](https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_backends.cpp) (whole networks)
[test_layers.cpp](https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_layers.cpp)
[test_halide_layers.cpp](https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_halide_layers.cpp) (there are not only Halide tests actually)

dkurt:
@wzw-intel, is there a way to choose Vulkan device because now it fallbacks to default CPU implementation for me.

wzw-intel:
> @wzw-intel, is there a way to choose Vulkan device because now it fallbacks to default CPU implementation for me.

Vulkan backend will choose the first available device.  Have you installed Vulkan runtime (loader library) and Vulkan driver? You can follow the steps descripted in the PR comments to setup your environments.

dkurt:
@wzw-intel, Thanks! It looks like accuracy tests use Vulkan backend properly. I just wanted to run performance tests and it fallbacks because of. Can you take at look on [perf_net.cpp](https://github.com/opencv/opencv/blob/master/modules/dnn/perf/perf_net.cpp)?
```cpp
bool isAvailable()
{
    return getContext() != NULL;
}
```

alalek:
[Test build](http://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/1462) dumps [this](http://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/1462/steps/perf_dnn-opencl/logs/stdio):

<details>

```
[ RUN      ] DNNTestNetwork.AlexNet/0, where GetParam() = OCV/CPU
[ INFO:0] VKCOM: init()
[ INFO:0] VKCOM: vkCreateInstance(&createInfo, NULL, &kInstance)
[ INFO:0] VKCOM: vkCreateDevice(kPhysicalDevice, &deviceCreateInfo, NULL, &device)
[ INFO:0] VKCOM: vkCreateCommandPool(device, &commandPoolCreateInfo, NULL, &cmd_pool)
[ INFO:0] VKCOM: init()
[ INFO:0] VKCOM: vkCreateDevice(kPhysicalDevice, &deviceCreateInfo, NULL, &device)
[ INFO:0] VKCOM: vkCreateCommandPool(device, &commandPoolCreateInfo, NULL, &cmd_pool)
Memory consumption:
    Weights(parameters): 233 Mb
    Blobs: 9 Mb
Calculation complexity: 2.33441 GFlops
[ PERFSTAT ]    (samples=1   mean=23.39   median=23.39   min=23.39   stddev=0.00 (0.0%))
[       OK ] DNNTestNetwork.AlexNet/0 (320 ms)
[ RUN      ] DNNTestNetwork.AlexNet/1, where GetParam() = OCV/OCL
[ INFO:0] VKCOM: init()
[ INFO:0] VKCOM: vkCreateInstance(&createInfo, NULL, &kInstance)
[ INFO:0] VKCOM: vkCreateDevice(kPhysicalDevice, &deviceCreateInfo, NULL, &device)
[ INFO:0] VKCOM: vkCreateCommandPool(device, &commandPoolCreateInfo, NULL, &cmd_pool)
[ INFO:0] VKCOM: init()
[ INFO:0] VKCOM: vkCreateDevice(kPhysicalDevice, &deviceCreateInfo, NULL, &device)
[ INFO:0] VKCOM: vkCreateCommandPool(device, &commandPoolCreateInfo, NULL, &cmd_pool)
[ INFO:0] Successfully initialized OpenCL cache directory: /build/.cache/opencv_opencl_cache_x64/
[ INFO:0] Preparing OpenCL cache configuration for context: Intel_R__Corporation--Intel_R__Gen9_HD_Graphics_NEO--18_31_11200
Memory consumption:
    Weights(parameters): 233 Mb
    Blobs: 9 Mb
Calculation complexity: 2.33441 GFlops
[ PERFSTAT ]    (samples=1   mean=19.14   median=19.14   min=19.14   stddev=0.00 (0.0%))
[       OK ] DNNTestNetwork.AlexNet/1 (327 ms)
[ RUN      ] DNNTestNetwork.AlexNet/2, where GetParam() = OCV/OCL_FP16
[ INFO:0] VKCOM: init()
[ INFO:0] VKCOM: vkCreateInstance(&createInfo, NULL, &kInstance)
[ INFO:0] VKCOM: vkCreateDevice(kPhysicalDevice, &deviceCreateInfo, NULL, &device)
[ INFO:0] VKCOM: vkCreateCommandPool(device, &commandPoolCreateInfo, NULL, &cmd_pool)
[ INFO:0] VKCOM: init()
[ INFO:0] VKCOM: vkCreateDevice(kPhysicalDevice, &deviceCreateInfo, NULL, &device)
[ INFO:0] VKCOM: vkCreateCommandPool(device, &commandPoolCreateInfo, NULL, &cmd_pool)
Memory consumption:
    Weights(parameters): 233 Mb
    Blobs: 9 Mb
Calculation complexity: 2.33441 GFlops
[ PERFSTAT ]    (samples=1   mean=13.63   median=13.63   min=13.63   stddev=0.00 (0.0%))
[       OK ] DNNTestNetwork.AlexNet/2 (392 ms)
[ RUN      ] DNNTestNetwork.AlexNet/3, where GetParam() = VKCOM/VULKAN
[ INFO:0] VKCOM: init()
[ INFO:0] VKCOM: vkCreateInstance(&createInfo, NULL, &kInstance)
[ INFO:0] VKCOM: vkCreateDevice(kPhysicalDevice, &deviceCreateInfo, NULL, &device)
[ INFO:0] VKCOM: vkCreateCommandPool(device, &commandPoolCreateInfo, NULL, &cmd_pool)
[ INFO:0] VKCOM: init()
[ INFO:0] VKCOM: vkCreateDevice(kPhysicalDevice, &deviceCreateInfo, NULL, &device)
[ INFO:0] VKCOM: vkCreateCommandPool(device, &commandPoolCreateInfo, NULL, &cmd_pool)
Memory consumption:
    Weights(parameters): 233 Mb
    Blobs: 9 Mb
Calculation complexity: 2.33441 GFlops
[ PERFSTAT ]    (samples=1   mean=30.58   median=30.58   min=30.58   stddev=0.00 (0.0%))
[       OK ] DNNTestNetwork.AlexNet/3 (372 ms)
```

</details>

Notes:
1. Vulkan initialization for non-Vulkan processing should be avoided. No need to allocate Vulkan resources if it will not be used.
2. As @dkurt mentioned, there is no extra calls in "vulkan" code path in perf_dnn tests.
3. Current error handling is questionable (just `CV_LOG_WARNING()`). For example, test_dnn fails due memory allocation error (it is limited to 2-3Gb on CI). Some fallback mechanism should be designed. In OpenCL if some kernel requires missing extension then it is able to fallback to CPU code path (see `OCL_RUN()`). Final result would be correct. If there is no switching fallback then C++ exception should be raised with proper cleanup of used memory.

modules/dnn/src/vkcom/src/vkcom.cpp
```diff
@@ -0,0 +1,403 @@
...
+{
+    VkDevice device;
+    VkQueue queue;
+    VkCommandPool cmd_pool;
+
+    VKCOM_CHECK_BOOL_RET_VAL(init(), false);
+    Context* ctx = getContext();
+    if (ctx)
+    {
+        ctx->ref++;
```
alalek: `ctx` is "thread local"?
If it is not, then the reference counted should be synchronized between threads properly (via `CV_XADD()` macro or use `std::atomic<int>` (OpenCV 4.0 requires C++11) ).

wzw-intel: Yes, it is thread local. 
It is for multiple callings on initPerThread(). e.g, multiple Net objects exsiting on a thread simultaneously. 

I am looking into the issues you and @dkurt mentioned. But I have short time on coding in these days and next week for I will be in biz travel. Responses and fixes may be a little bit late. Sorry.

dkurt:
Unfortunately I'm not able to run all the tests because of segfault:
```
[ RUN      ] Layer_Test_Halide/Deconvolution.Accuracy/198, where GetParam() = ([6, 9, 3], 5x6, 3x1, 0x1, 2x2, [1, 1, 0, 0], false, OCV/OCL)
[       OK ] Layer_Test_Halide/Deconvolution.Accuracy/198 (17 ms)
[ RUN      ] Layer_Test_Halide/Deconvolution.Accuracy/199, where GetParam() = ([6, 9, 3], 5x6, 3x1, 0x1, 2x2, [1, 1, 0, 0], false, OCV/OCL_FP16)
[       OK ] Layer_Test_Halide/Deconvolution.Accuracy/199 (28 ms)
[ RUN      ] Layer_Test_Halide/Deconvolution.Accuracy/200, where GetParam() = ([6, 9, 3], 5x6, 3x1, 0x1, 2x2, [1, 1, 0, 0], false, VKCOM/VULKAN)
[       OK ] Layer_Test_Halide/Deconvolution.Accuracy/200 (17 ms)
[ RUN      ] Layer_Test_Halide/Deconvolution.Accuracy/201, where GetParam() = ([6, 9, 3], 5x6, 3x1, 0x1, 2x2, [1, 1, 0, 0], true, OCV/OCL)
[       OK ] Layer_Test_Halide/Deconvolution.Accuracy/201 (28 ms)
[ RUN      ] Layer_Test_Halide/Deconvolution.Accuracy/202, where GetParam() = ([6, 9, 3], 5x6, 3x1, 0x1, 2x2, [1, 1, 0, 0], true, OCV/OCL_FP16)
[       OK ] Layer_Test_Halide/Deconvolution.Accuracy/202 (17 ms)
[ RUN      ] Layer_Test_Halide/Deconvolution.Accuracy/203, where GetParam() = ([6, 9, 3], 5x6, 3x1, 0x1, 2x2, [1, 1, 0, 0], true, VKCOM/VULKAN)
[ WARN:0] Vulkan check failed
Segmentation fault (core dumped)
```
It happens not for a specific test case but after some time.

PR #12711, opened 2018-09-30
dnn(ocl4dnn): drop weights_buf

alalek:
- avoid memory access violation during "prefetch" stage

No performance regressions:
- on i5-6600 (Fedora 28, Intel OpenCL NEO 18.30.11148)
- on i7-6700K (Fedora 28, Intel OpenCL NEO 18.33.11309)


vpisarev:
:+1:

PR #12716, opened 2018-10-01
Try to enable Android support for G-API module (do not merge!)

dmatveev:
Please don't merge this PR even if it is successful.

It is opened for testing purposes only

```
**WIP**
force_builders=linux,Android armeabi-v7a,windows,mac,docs,ios
```


dmatveev:
Seems Android NDK r10 (which is used on CI) has some header issues for C++ standard library which were fixed in r16 (according to some findings from StackOverflow).

Closing this issue so far as Android support is not a priority for now for G-API module (issue should be resolved after infrastructure update)

PR #12717, opened 2018-10-01
drop cvCheckChessboard with IplImage

tompollok:
### This pullrequest changes

drops cvCheckChessboard with IplImage

fixes documentation/comments (img vs src)


alalek:
It is a part of public legacy C API: https://github.com/opencv/opencv/blob/3.4.3/modules/calib3d/include/opencv2/calib3d/calib3d_c.h#L219

PR #12718, opened 2018-10-02
New types

cv3d:
DO NOT MERGE! Still testing ...

### This pullrequest changes
- Conversion between types
- SSE intrinsics for the new types
- Output formatters

```
**WIP**
force_builder=linux
```


modules/core/include/opencv2/core/hal/intrin_sse.hpp
```diff
@@ -368,6 +368,14 @@ inline v_uint8x16 v_pack_u(const v_int16x8& a, const v_int16x8& b)
...
 { _mm_storel_epi64((__m128i*)ptr, _mm_packus_epi16(a.val, a.val)); }
 
+// [a0 0 | b0 0]  [a1 0 | b1 0]
+inline v_uint32x4 v_pack_u(const v_int64x2& a, const v_int64x2& b)
+{
+    __m128i v0 = _mm_unpacklo_epi32(a.val, b.val); // a0 a1 0 0
+    __m128i v1 = _mm_unpackhi_epi32(a.val, b.val); // b0 b1 0 0
+    return v_uint32x4(_mm_unpacklo_epi32(v0, v1));
+}
+
```
cv3d: Need review!

modules/core/include/opencv2/core/hal/intrin_sse.hpp
```diff
@@ -2578,6 +2586,16 @@ inline v_float64x2 v_cvt_f64_high(const v_int32x4& a)
...
+inline v_float64x2 v_cvt_f64(const v_int64x2& a)
+{
+    return v_float64x2(_mm_cvtepi32_pd(a.val));
+}
+
+inline v_float64x2 v_cvt_f64_high(const v_int64x2& a)
+{
+    return v_float64x2(_mm_cvtepi32_pd(_mm_srli_si128(a.val, 8)));
+}
+
```
cv3d: Both functions need review

alalek:
Avoid optimizations (and touching intrinsics) at this stage of PR review.

vpisarev:
1. what's the difference between this PR and #12611? I think, we need to close one of the two
2. we do not need fixed-point types in the library at the moment. It does not add any new functionality, but it adds extra bulk (if we attempt to implement even basic operations for those types).

cv3d:
1. This PR utilizes #12611 to add higher-level operations. Extending #12611 in place would make it very enormous, besides it is only meant to redesign the register. You can think of this one as a testing one, and until you merge that one, I plan to just keep experimenting here.
2. Add extra blocks where? If you are referring to the convert functions in this PR, then such blocks would be there anyway, it just will be NULL instead, that is all.

Do you want me to remove the fixed point definitions from #12611?

vpisarev:
yes, please, remove fixed-point definitions from #12611. Also, please remove the words: "By merging this, you agree to properly attribute the contributor wherever you mention this feature!". We cannot fulfil this requirement, therefore we'd rather close that PR.

cv3d:
I can remove the word `wherever`, but I cannot accept a number near my name in the contributors list as a an acknowledgment in the change log. You do not think a merge number is sufficient, do you?

cv3d:
> We have certain practices that exist there for many years

Very convincing reply!

Anyway, I removed the notice because I do not want the effort to go in vain.
Hopefully you would think about the changelog issue seriously, a number would not differentiate a minor documentation update from a major change (given that you also squash the commits)!

PR #12723, opened 2018-10-02
2.4

Wrench24:
### This pullrequest changes

Reviewed and added extra empty line


berak:
hi, @Wrench24 just asking, -- *why* did you think, this would be useful (or nessecary) ?
(the 2.4 branch is no more developed since 5 years already, it's "frozen, and maintenance only")

but you must have had a *reason* to do so, which is ?
(this kind of "drive by shooting" happens like once a month, help us to avoid it in the future, please.)

PR #12739, opened 2018-10-04
Core: fix float16_t support for SparseMat::convertTo()

### LINKED ISSUE #12738: float16_t convertors support is incomplete
float16_t is supported in `cv::getConvertFunc()` and `cv::getConvertScaleFunc()`, but it is not supported in `cv::getConvertElem()`. Is that intentional (if so, mind to tell me why?), or was it accidentally skipped?

I understand float16_t is somehow special, but my impression is that, if `cv::getConvertFunc()` and `cv::getConvertScaleFunc()` are supportted, then `cv::getConvertElem()` should be as well.

cv3d:
resolves #12738


vpisarev:
I don't think we need this functionality; it increases the library size without adding any useful functionality

cv3d:
> it increases the library size

I wonder how many bytes it adds..

> without adding any useful functionality

I believe it is useful at least to maintain homogeneity and consistency.

vpisarev:
Let's close this for now. We'll reopen it when there is real-life case with half-precision sparse matrices.

PR #12740, opened 2018-10-04
Refactoring in CommandLineParser (use std::string and std::regex)

mshabunin:
### This pullrequest changes

* use std::string in CommandLineParser internals
* use std::regex for parsing
* separated `Impl` from interface object

**Note:** forbidden copy and assignment of CommandLineParser


mshabunin:
Looks like GCC <= 4.9 do not support `std::regex`

PR #12742, opened 2018-10-04
Removed ocl::KernelArg::Local()

savuor:
### This pullrequest changes

`ocl::KernelArg::Local()` has been removed since it's incorrect: it doesn't provide local mem size which makes kernel run to fail.

The desired effect of providing local memory argument can be achieved by the following constructor:
`ocl::KernelArg(ocl::KernelArg::LOCAL, nullptr, 0, 0, nullptr, localMemSize)`


alalek:
Enum value and implementation should be cleaned too.

savuor:
Decided not to remove the functuon but to add local mem size as an argument

PR #12751, opened 2018-10-05
imgproc: fix deconstructor crash at closing of OpenCL convert color

### LINKED ISSUE #12750: imgproc: program crashes after test of imgproc "OCL_ImgProc/CvtColor8u32f"

- OpenCV => 3.4 branch ( 0101fa789cde6c03bc58288a56e1d6ce7175cd48 )
- Operating System / Platform => Windows 10 64bit
- CUDA => 10.0

##### Summarized description
 - The ```opencv_test_imgproc``` passes, but when I ran it from target ```RUN_TESTS```, it says ```Segfault```

```
    Start 9: opencv_test_imgproc
1/1 Test #9: opencv_test_imgproc ..............***Exception: SegFault137.81 sec
```

- I ran ```opencv_test_imgproc``` separately, but all the test passes (!)
- To reproduce what gtest was seeing, I need to ran it in Debug mode
- When running ```opencv_test_imgproc``` with "Start debug" on Visual Studio 2015, the test passed but at the end, it did cause an ```Segfault``` at [ocl.cpp](https://github.com/opencv/opencv/blob/0101fa789cde6c03bc58288a56e1d6ce7175cd48/modules/core/src/ocl.cpp#L4328)
https://github.com/opencv/opencv/blob/0101fa789cde6c03bc58288a56e1d6ce7175cd48/modules/core/src/ocl.cpp#L4322-L4329

##### Detailed description
- I narrowed down the test case, and the test case was from ```OCL_ImgProc/CvtColor8u32f```
- Furthermore, it was related to test of either HSV/Lab/Luv conversion.
- Also, although I put a break point at where it causes ```Segfault```, it doesn't reproduce the situation.
- I have to exit from the [main routine call](https://github.com/opencv/opencv/blob/0101fa789cde6c03bc58288a56e1d6ce7175cd48/modules/imgproc/test/test_main.cpp) and it was happening after that exit call.  Mysterious...

- After watching the source code, I realized that some of the static ```UMat``` was used in a wrong scope, such like [here ](https://github.com/opencv/opencv/blob/0101fa789cde6c03bc58288a56e1d6ce7175cd48/modules/imgproc/src/color_hsv.cpp#L1483-L1486)
https://github.com/opencv/opencv/blob/0101fa789cde6c03bc58288a56e1d6ce7175cd48/modules/imgproc/src/color_hsv.cpp#L1483-L1486
- These ```UMat``` were passed as a [parameter of the OpenCL kernel](https://github.com/opencv/opencv/blob/0101fa789cde6c03bc58288a56e1d6ce7175cd48/modules/imgproc/src/color_hsv.cpp#L1516-L1521) but the scope [was closed before](https://github.com/opencv/opencv/blob/0101fa789cde6c03bc58288a56e1d6ce7175cd48/modules/imgproc/src/color_hsv.cpp#L1519) it was [actually used](https://github.com/opencv/opencv/blob/0101fa789cde6c03bc58288a56e1d6ce7175cd48/modules/imgproc/src/color_hsv.cpp#L1521)
https://github.com/opencv/opencv/blob/0101fa789cde6c03bc58288a56e1d6ce7175cd48/modules/imgproc/src/color_hsv.cpp#L1516-L1521
- Looking back the history, it seems that color convert was splitted in [early 2018](https://github.com/opencv/opencv/commit/0b900b54e54e8cbd19a65cfe69d20838f142e6b3), and before the split, the scope was aligned with the actual use.
- I moved the ```UMat``` declarations and also, I removed the ```static```
- Now the test passed without ```Segfault```

##### Steps to reproduce
- Build under environment explained above.
- Run with CUDA 10.0 and run the OpenCL kernel on dGPU 
- ```OPENCV_OPENCL_DEVICE=:dgpu opencv_test_imgproc --gtest_filter=OCL_ImgProc/CvtColor8u32f*``` 

- Also, there I'm not sure why this doesn't happen on ```CPU``` and ```igpu``` on my laptop.
- Still, the scope difference seems bad, so I think it has meaning to fix the scope.

tomoaki0705:
### This pullrequest changes
 *  resolves #12750 
 * still, I'm not that sure between the cause and the crash, so i'm open to discuss about "how" to fix

### DIFF
--- a/modules/core/src/ocl.cpp
+++ b/modules/core/src/ocl.cpp
@@ -4325,6 +4325,10 @@ public:
         CV_Assert(entry.clBuffer_ != NULL);
         LOG_BUFFER_POOL("OpenCL release buffer: %p, %lld (0x%llx) bytes\n",
                 entry.clBuffer_, (long long)entry.capacity_, (long long)entry.capacity_);
+#ifdef _WIN32
+        if (cv::__termination)
+            return;  // don't call any OpenCL runtime functions during termination, they may behave incorrectly during process termination
+#endif
         CV_OCL_DBG_CHECK(clReleaseMemObject(entry.clBuffer_));
     }
 };
@@ -4877,6 +4881,10 @@ public:
             }
             else
 #endif
+#ifdef _WIN32
+                if (cv::__termination)
+                    return;  // don't call any OpenCL runtime functions during termination, they may behave incorrectly during process termination
+#endif
             {
                 cl_int retval = clReleaseMemObject((cl_mem)u->handle);
                 CV_OCL_DBG_CHECK_RESULT(retval, cv::format("clReleaseMemObject(ptr=%p)", (void*)u->handle).c_str());
@@ -4894,6 +4902,10 @@ public:
         else
         {
             CV_Assert(u->origdata == NULL);
+#ifdef _WIN32
+            if (cv::__termination)
+                return;  // don't call any OpenCL runtime functions during termination, they may behave incorrectly during process termination
+#endif
             if(u->data && u->copyOnMap() && u->data != u->origdata)
             {
                 fastFree(u->data);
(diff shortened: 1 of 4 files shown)


alalek:
succeeded by #12791

PR #12757, opened 2018-10-06
Utilize macros to improve readability and extensibility

cv3d:
This should be useful when extending/maintaining current implementation (e.g. for the new types)


cv3d:
@alalek Could you please restart failed builds? They are unrelated!

cv3d:
@alalek Can you please check ARMv8, which is failing again!

cv3d:
@vpisarev Can you let me know why you think it is useless? Is the code before this patch understandable or easy to extend as it is after applying this patch? It seems to me you just want to practice your power, no more no less..

PR #12758, opened 2018-10-06
Utilize std::mt19937_64 generator

cv3d:
### This pullrequest changes
A thin wrapper for `std::mt19937_64` random generator


cv3d:
@alalek Help needed!
It seems Linux has a problem in following MT19937 standards, while Windows and Mac are just fine.
Do you happen to know any tweak for that?

alalek:
OpenCV pseudo random generator (usually used for testing purposes and for RANSAC-like algorithms):
- strong requirement of bit-exactness (between platforms, build flags, etc)
- thread-local
- fast
- statistics accuracy / strong requirements are tolerable.

---

No need to fix things which are **not broken**.

cv3d:
> No need to fix things which are not broken

We can say so for all upgrades to C++11 features, including smart pointers, threads, etc.

cv3d:
> No need to fix things which are not broken

@alalek It is broken, and I have updated the description with more convincing details.
This PR basically wants to leave true randomness to the experts who implemented the standard library, what is wrong with that?

cv3d:
@vpisarev He is talking about RNG implementation, and I am replacing the MT19937 implementation 😕

cv3d:
@alalek Have you read the updated description, and noticed that it is updating a different generator than what you thought initially? Do you also think this is a useless patch?

PR #12760, opened 2018-10-07
Replace NULL with nullptr

cv3d:
Merge with opencv/opencv_contrib#1814


alalek:
I don't want this patch. It breaks straightforward "merge-3.4" process.

NULL is not a problem here.
Problem is "0" for nullptr, but even this case is covered well by all compilers / code analyzers.

cv3d:
@alalek STL (Stephan T. Lavavej) has a famous quote about that:
> NULL is an abomination

Shall I change the base of this PR to 3.4 and define `nullptr` as an alias to `NULL`, in case the compilers would allow it?

vpisarev:
@cv3d, please, discuss such big patches privately or via submitted issues, before submitting them

cv3d:
@vpisarev I will do, but for the notice: I remember opening several issues, and even RFCing you, but I got no reply..

PR #12761, opened 2018-10-07
tabulation -> 4 spaces

cv3d:
Merge with opencv/opencv_contrib#1815

### DIFF
--- a/modules/calib3d/src/polynom_solver.cpp
+++ b/modules/calib3d/src/polynom_solver.cpp
@@ -33,7 +33,7 @@ int solve_deg3(double a, double b, double c, double d,
 {
   if (a == 0) {
     // Solve second order system
-    if (b == 0)	{
+    if (b == 0)    {
       // Solve first order system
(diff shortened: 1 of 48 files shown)


modules/calib3d/src/polynom_solver.cpp
```diff
@@ -33,7 +33,7 @@ int solve_deg3(double a, double b, double c, double d,
 {
   if (a == 0) {
     // Solve second order system
-    if (b == 0)	{
+    if (b == 0)    {
```
alalek: Replaced one wrong case to another.

vpisarev:
another useless patch

cv3d:
@vpisarev Why you think it is useless?
If so, why does the contribution guidelines asks as to avoid tabulation and utilize spaces?
I was editing the changes line by line, just to find that you closed the PR...

I am going to open another PR, and I hope you would either merge it, or convince me that it is useless , and I will close it myself.

PR #12765, opened 2018-10-07
TS/CUDA: Show device ID in case of multiple devices

cv3d:
When two installed CUDA-enabled devices have the same name, it is impossible to differentiate them..
As a matter of fact, it turned out that some failing tests reported in #12561 only occurs with a specific device, and without such information, debugging is very hard..

### This pullrequest changes
- Prints out the device ID if it is greater than zero

```
force_builders=Custom
```


tomoaki0705:
@vpisarev may I ask the reason why you closed this PR?
From my point of view, this PR has good enough reason, and good enough not to be closed immediately, at least discuissionable.

Is there a strong reason to close this PR?

Issue #5065, opened 2015-07-27
Ignoring the flags of solvePnP and silently executing EPNP


opencv-pushbot:
Transferred from http://code.opencv.org/issues/4472
```
|| Philipp Hasper on 2015-07-08 08:42
|| Priority: Normal
|| Affected: branch 'master' (3.0-dev)
|| Category: calibration, 3d
|| Tracker: Bug
|| Difficulty: 
|| PR: 
|| Platform: Any / Any
```

Ignoring the flags of solvePnP and silently executing EPNP
-----------
```
The way how uPnP and DLS have been disabled in [1] is highly counter-intuitive:

Just executing EPNP regardless of the given flag [2] is confusing. If somebody wants to evaluate the OpenCV version of uPnP, let him do that (and not silently executing epnp in the background). Imagine somebody wants to evaluate the available PnP solvers in OpenCV and finds out that it does not matter which one he chooses because the results are all the same.

So we have to
a) remove the upnp and dls flags completely or
b) leave them in and document the unstable behavior with a warning in the online documentation.

The latter option might lead to somebody improving the algorithms

[1]: https://github.com/Itseez/opencv/pull/3828 
[2]: https://github.com/vpisarev/opencv/blob/ca19ae8b5ae53afe3ca0084c55178b8d0796b28b/modules/calib3d/src/solvepnp.cpp#L66
```

History
-------

PR #12772, opened 2018-10-08
More accurate pinhole camera calibration with imperfect planar target

xoox:
**Merge with extra**: opencv/opencv_extra#528
### Problem to be solved

Checkerboard patterns on paper using off-the-shelf printers are the most
convenient calibration targets. And these printed patterns stuck on
roughly planar surfaces are the mostly used calibration structure world
wide. But in most cases printed patterns are inaccurate and the sticking
surfaces are not perfectly planar.

chessboard

For example, the square size of the above chessboard is set to 15mm
in the PDF file. The printed paper was pasted to a glass panel which was
not so perfect planar structure. We measured the distances between
combinations of p1, p2, p3 and p4 with a vernier caliper of 0.02mm
resolution (The uncertainties of the measurements are probably larger
than 0.1mm I think). The result are compared to hypothesis values. The
maximum error is around 1mm (about 3~4 pixels in our application).

distance | hypothesis | measured
---------|------------|---------
\|p1-p2\|| 225        | 224.14  
\|p1-p3\|| 135        | 135.35  
\|p1-p4\|| 262.393    | 261.85  
\|p2-p3\|| 262.393    | 262     
\|p2-p4\|| 135        | 135.13  
\|p3-p4\|| 225        | 224.73  

The inaccuracy of the checkerboard causes less precise and less stable
estimation of camera intrinsic parameters.

### New method integrated

This PR integrates an extension to the standard pinhole camera
calibration method in OpenCV. The method was proposed in the paper:
Klaus H. Strobl and Gerd Hirzinger. [More accurate pinhole camera
calibration with imperfect planar target](https://elib.dlr.de/71888/1/strobl_2011iccv.pdf). In 2011 IEEE International
Conference on Computer Vision (ICCV), pages 1068-1075, Barcelona, Spain,
Nov 2011. IEEE. And it was included in the camera calibration toolbox
[DLR CalDe and DLR CalLab](http://www.robotic.dlr.de/callab/).
In many common cases with inaccurate calibration plates, this method can
dramatically improve the precision of the estimated camera parameters.

In principle the method releases 3D coordinates of the object structure
when performing non-linear Levenberg-Marquardt optimization. The current
calibration process can be easily reinforced by adding extra
optimization parameters and extending the Jacobian matrix.

Before trying to integrate it into OpenCV, a standalone project
[calibrel](https://github.com/xoox/calibrel) was created to reimplement
and validate this method. [The results were double-checked](https://github.com/xoox/calibrel/issues/1#issuecomment-419775155) with [CalLab](http://www.robotic.dlr.de/callab/).
In addition to validating with stereo cameras in the paper, [we also
evaluated its precision using measured distances](https://github.com/xoox/calibrel/issues/2).

When preparing to add test code for this new method, [the existing data
for regression test is reordered](https://github.com/xoox/opencv_extra/blob/calib-release-object/testdata/cv/cameracalibration/calib1.dat) where the object points and image
feature points can be also used by the new calibration method (with
calibration flags changed). The results with releasing object points are
so diverse to standard methods. It may be caused by bad set of images or
inaccuracy of feature points. The comparison is shown in the following
table and further discussions are needed. Could anyone do some
comparisons between these two methods with regard to the same set of
images of an **_accurate_** checkerboard? Both cameras with larger and
smaller distortions are needed. Experiments can be conducted with
`samples/cpp/calibration.cpp` and
`samples/cpp/tutorial_code/calib3d/camera_calibration/camera_calibration.cpp`
in this PR.

parameter | standard   | release_object
----------|------------|---------------
fx        | 809.94872  | 790.93722
fy        | 806.56095  | 775.18624
cx        | 361.99611  | 331.43434
cy        | 214.39525  | 203.89052
k1        | -0.1282055 | -0.244456
k2        | -0.3944258 | 0.3578029
p1        | -0.0043549 | -0.002096
p2        | -0.0073207 | -0.001801
RMS       | 0.138892   | 0.0830433

I have no information of the camera, lens and checkerboard used in
`calib1.dat`. But after drawing its feature points on images, it seems those
images are probably not good candidates for camera calibration. Please
check the images on this PR opencv/opencv_extra#528. Also see [Dr.  Strobl](https://rmc.dlr.de/rm/en/staff/klaus.strobl/)'s
comments on good calibration images: https://github.com/xoox/calibrel/issues/1#issuecomment-419279616

If this PR is going to be squashed, it'd be better full commit messages
are kept instead of commit titles only, for that there are useful
information in the full commit messages.

In the similar fashion, this method can also be included into
`stereoCalibrate()`. But the speed is expected much slower. If the
intrinsic parameters can be estimated with high accuracy for each of the
cameras individually using `calibrateCamera()`, it is recommended to do
so and then pass CALIB_FIX_INTRINSIC flag to `stereoCalibrate()` along
with the computed intrinsic parameters. However, I'd integrate this
method into `stereoCalibrate()` if the necessity is confirmed.

#### cvCalibrateCamera4()

A new C interface is added which supports both standard calibration
method and the extended new calibration method.

#### calibrateCameraRO()

New C++ interfaces are added which support both standard calibration
method and the extended new calibration method.

#### Regression test code

The test code is modified to empower tests of both standard calibration
method and the new added one. The updated test data can be found here:
https://github.com/xoox/opencv_extra/tree/calib-release-object A PR opencv/opencv_extra#528 was
also created for the new test data.

#### Sample calibration code

Two calibration samples are updated to support both standard
calibration method and the extended new calibration method.

In conclusion, this PR is just a very straightforward extension to
current standard calibration code.

### DIFF
--- a/doc/tutorials/calib3d/table_of_content_calib3d.markdown
+++ b/doc/tutorials/calib3d/table_of_content_calib3d.markdown
@@ -13,7 +13,7 @@ Although we get most of our images in a 2D format they do come from a 3D world. 
 
 -   @subpage tutorial_camera_calibration
 
-    *Compatibility:* \> OpenCV 2.0
+    *Compatibility:* \> OpenCV 4.1
 
     *Author:* Bernát Gábor
 
--- a/doc/tutorials/calib3d/table_of_content_calib3d.markdown
+++ b/doc/tutorials/calib3d/table_of_content_calib3d.markdown
@@ -13,7 +13,7 @@ Although we get most of our images in a 2D format they do come from a 3D world. 
 
 -   @subpage tutorial_camera_calibration
 
-    *Compatibility:* \> OpenCV 4.1
+    *Compatibility:* \> OpenCV 4.0
 
     *Author:* Bernát Gábor
 
--- a/modules/calib3d/include/opencv2/calib3d.hpp
+++ b/modules/calib3d/include/opencv2/calib3d.hpp
@@ -992,7 +992,7 @@ together.
 the top-right corner point of the calibration board grid. This parameter is used only if
 CALIB_RELEASE_OBJECT is set. In other cases it can be set to -1 which is just ignored internally.
 According to \cite strobl2011iccv, two other points are also fixed. In this implementation,
-objectPoints[0].front and objectPoints[0].back are used. Accurate rvecs and tvecs are only possible
+objectPoints[0].front and objectPoints[0].back.z are used. Accurate rvecs and tvecs are only possible
 if coordinates of these three fixed points are accurate enough.
 @param cameraMatrix Output 3x3 floating-point camera matrix
 \f$A = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1}\f$ . If CV\_CALIB\_USE\_INTRINSIC\_GUESS
--- a/modules/calib3d/include/opencv2/calib3d.hpp
+++ b/modules/calib3d/include/opencv2/calib3d.hpp
@@ -1106,12 +1106,14 @@ camera parameters. Both the object-releasing method and standard method are supp
 function. Use the parameter **iFixedPoint** for method selection. In the internal implementation,
 calibrateCamera() is a wrapper for this function.
 
-@param objectPoints See calibrateCamera() for details. If the method of releasing object to be used,
+@param objectPoints Vector of vectors of calibration pattern points in the calibration pattern
+coordinate space. See calibrateCamera() for details. If the method of releasing object to be used,
 the identical calibration board must be used in each view and it must be fully visible, and all
 objectPoints[i] must be the same and all points should be roughly close to a plane. **The calibration
 target has to be rigid, or at least static if the camera (rather than the calibration target) is
 shifted for grabbing images.**
-@param imagePoints See calibrateCamera() for details.
+@param imagePoints Vector of vectors of the projections of calibration pattern points. See
+calibrateCamera() for details.
 @param imageSize Size of the image used only to initialize the intrinsic camera matrix.
 @param iFixedPoint The index of the 3D object point in objectPoints[0] to be fixed. It also acts as
 a switch for calibration method selection. If object-releasing method to be used, pass in the
--- a/modules/calib3d/include/opencv2/calib3d.hpp
+++ b/modules/calib3d/include/opencv2/calib3d.hpp
@@ -501,10 +501,11 @@ vector\<Point3f\> ), where N is the number of points in the view.
 4, 5, 8, 12 or 14 elements. If the vector is empty, the zero distortion coefficients are assumed.
 @param imagePoints Output array of image points, 2xN/Nx2 1-channel or 1xN/Nx1 2-channel, or
 vector\<Point2f\> .
-@param jacobian Optional output 2Nx(10+\<numDistCoeffs\>) jacobian matrix of derivatives of image
+@param jacobian Optional output 2Nx(10+\<numDistCoeffs\>+3N) jacobian matrix of derivatives of image
 points with respect to components of the rotation vector, translation vector, focal lengths,
-coordinates of the principal point and the distortion coefficients. In the old interface different
-components of the jacobian are returned via different output parameters.
+coordinates of the principal point, the distortion coefficients and coordinates of the object
+points. In the old interface different components of the jacobian are returned via different output
+parameters.
 @param aspectRatio Optional "fixed aspect ratio" parameter. If the parameter is not 0, the
 function assumes that the aspect ratio (*fx/fy*) is fixed and correspondingly adjusts the jacobian
 matrix.
--- a/modules/calib3d/include/opencv2/calib3d.hpp
+++ b/modules/calib3d/include/opencv2/calib3d.hpp
@@ -993,7 +993,8 @@ the top-right corner point of the calibration board grid. This parameter is used
 CALIB_RELEASE_OBJECT is set. In other cases it can be set to -1 which is just ignored internally.
 According to \cite strobl2011iccv, two other points are also fixed. In this implementation,
 objectPoints[0].front and objectPoints[0].back.z are used. Accurate rvecs and tvecs are only possible
-if coordinates of these three fixed points are accurate enough.
+if coordinates of these three fixed points are accurate enough. In theory, the three fixed points
+can be arbitrarily chosen as long as they are not collinear.
 @param cameraMatrix Output 3x3 floating-point camera matrix
 \f$A = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1}\f$ . If CV\_CALIB\_USE\_INTRINSIC\_GUESS
 and/or CALIB_FIX_ASPECT_RATIO are specified, some or all of fx, fy, cx, cy must be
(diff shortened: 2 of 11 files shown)


xoox:
The build failed because modified test data are needed. The data can be found here: opencv/opencv_extra#528
Is there any way to control the build bot to use a specific branch of opencv_extra?

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -1080,23 +1101,38 @@ The algorithm performs the following steps:
    findChessboardCorners, solvePnP, initCameraMatrix2D, stereoCalibrate, undistort
  */
 CV_EXPORTS_AS(calibrateCameraExtended) double calibrateCamera( InputArrayOfArrays objectPoints,
-                                     InputArrayOfArrays imagePoints, Size imageSize,
+                                     InputArrayOfArrays imagePoints, Size imageSize, int iFixedPoint,
```
alalek: Lets preserve existing interfaces and add new one (something like `calibrateCameraReleaseObject`).

> iFixedPoint

Looks like, this and other new parameters imply `flags += CALIB_RELEASE_OBJECT` automatically.

xoox: > Lets preserve existing interfaces and add new one (something like `calibrateCameraReleaseObject`).

That's my initial thought before starting. With regard to the fact that
they share most of the interface documents, I chose functions overload.
OK, I'll add new interfaces instead of overload old ones. How about the
name `calibrateCameraRO`? It's shorter and attractive to users, but with
the meaning less obvious. Since most part of the interface document are
shared by `calibrateCamera` and `calibrateCameraRO`,
`calibrateCameraRO`'s interface document will be simpler with most
explanations referenced to `calibrateCamera`.

The standard calibration method and object-releasing calibration method
will be implemented in `calibrateCameraRO`, and method is chosen based
on whether `iFixedPoint` is within a rational range. `calibrateCamera`
is then made a wrapper to `calibrateCameraRO`.

> 
> > iFixedPoint
> 
> Looks like, this and other new parameters imply `flags += CALIB_RELEASE_OBJECT` automatically.

Only `iFixedPoint` is input parameter, the other two are optional output
parameters. You are right. The method selection can be controlled with
`iFixedPoint`. I'll remove `CALIB_RELEASE_OBJECT` item and make the
method selection based on the value of `iFixedPoint`.

modules/calib3d/src/calibration.cpp
```diff
@@ -1716,8 +1862,47 @@ CV_IMPL double cvCalibrateCamera2( const CvMat* objectPoints,
...
+                    CvMat* rvecs, CvMat* tvecs, CvMat* newObjPoints, int flags, CvTermCriteria termCrit )
+{
+    // If iFixedPoint is out of rational range, fall back to standard method
+    if( iFixedPoint > npoints->data.i[0] - 2 )
+        iFixedPoint = -1;
+    // check object points. If not qualified, fall back to standard calibration.
+    int nimages = npoints->rows * npoints->cols;
+    int npstep = npoints->rows == 1 ? 1 : npoints->step / CV_ELEM_SIZE(npoints->type);
+    int i, ni;
+    if( iFixedPoint > 0 )
```
alalek: Why `iFixedPoint == 0` is out of scope? Is it allowed?

> The index of the 3D object point in objectPoints[0] to be fixed. Usually it is
the top-right corner point of the calibration board grid.

Index is zero-based in C++. Zero index should correspond to "top-left" corner.

xoox: Based on the cited paper, three 3D object points are chosen to be fixed. All x, y, z coordinates of point1 and point2 are fixed, and only z of point3 is fixed. In theory, these points could be arbitrarily chosen as long as they are non-collinear. but in practice, it will deliver more accurate results to select points that are far away from each other. In the internal implementation, the first (zero index, "top-left" corner) point is already selected as point1, see line 1651-1653. And the z coordinate of the last point ("bottom-right" corner) is also fixed as point3.z, see line 1657. Users have the freedom to select point2, and "top-right" or "bottom-left" point of the calibration board grid is recommended.

Use the convention in the sample C++ code, the first point ("top-left" corner) has the coordinates (0, 0, 0), "top-right" corner has (d, 0, 0) and the last point ("bottom-right" corner) has (x3, y3, 0). The accuracy of intrinsic parameters is free from these fixed points. But for accurate rvecs, tvecs and newObjPoints, the fixed coordinates must be accurate enough. In the above convention, only "d" should be accurately determined. "d" can be easily measured as the distance between the "top-left" and "top-right" corners.

There're samples on how to set the accurate "d". `samples/cpp/calibration.cpp` line 162 and `samples/cpp/tutorial_code/calib3d/camera_calibration/camera_calibration.cpp` line 561.

alalek: Thank you for detailed explanation.

Perhaps we should have documentation notes about this (specify valid / expected range of input values):
> In the internal implementation, the first (zero index, "top-left" corner) point is already selected as point1

> And the z coordinate of the last point ("bottom-right" corner) is also fixed as point3.z

Also add parameters checks into code like this:
```
CV_Assert(iFixedPoint != 0);
CV_Assert(iFixedPoint != numpoints - 1);
```
(not sure about second one)

xoox: The current document for `iFixedPoint` has the following information:

```
...
According to \cite strobl2011iccv, two other points are also fixed. In
this implementation, objectPoints[0].front and objectPoints[0].back.z
are used. Accurate rvecs and tvecs are only possible if coordinates of
these three fixed points are accurate enough.
...
```

I'd like to update it with the following text:

```
@param iFixedPoint The index of the 3D object point in objectPoints[0] to be fixed. Usually
the top-right corner point of the calibration board grid is recommended. For values out of the range
[1, objectPoints[0].size()-2], calibration will fall back to standard method. According to
\cite strobl2011iccv, two other points are also fixed. In this implementation, objectPoints[0].front
and objectPoints[0].back.z are used. Accurate rvecs, tvecs and newObjPoints are only possible if
coordinates of these three fixed points are accurate enough. If standard calibration method to be
used, just set this parameter to a negative value.
```

Is it acceptable?

------------------------------------------------------------------------

I think soft limit is a better experience for users, i.e. just falling back
to standard calibration method if `iFixedPoint < 1` or `iFixedPoint > numpoints - 2`.
If accuracy isn't the most important thing for some users, then fallback
scheme is acceptable for them. It seems assertions are a little bit
harsh for this parameter.

alalek: Before using, users would like to compare accuracy between old and new functions. What parameters are fit for their tasks.

I believe we should not confuse them with the same results (and silently disabled algorithm) if `iFixedPoint` is out of valid range for this algorithm (error message is a normal way to say that input values is out of expected range - you may use `CV_Check()` macros to add more accurate error message).

catree: I also agree that silently falling back to the classical `calibrateCamera` is not a good thing.

Another [example](https://github.com/opencv/opencv/issues/5065) that led users to report issues are the `SOLVEPNP_DLS` and `SOLVEPNP_UPNP` flags in `SolvePnP` function.

xoox: OK, the confusing might arise from the fact that we are using
`iFixedPoint` as a switch for calibration method selection. It'd be
better to not use "fall back" words. Just tell users to select new
method, pass in the parameter in the specified range. If standard method
is demanded, set it out of the range.

If new method is requested, but the `objectPoints` data are not qualified,
then `CV_Assert()` or `CV_error()` is called.

Thanks for your comments. I'll add more commits for further review.

doc/tutorials/calib3d/table_of_content_calib3d.markdown
```diff
@@ -13,7 +13,7 @@ Although we get most of our images in a 2D format they do come from a 3D world.
 
 -   @subpage tutorial_camera_calibration
 
-    *Compatibility:* \> OpenCV 2.0
+    *Compatibility:* \> OpenCV 4.1
```
xoox: If this feature could be released with 4.0, the compatibility would be updated to 4.0.

xoox: If this feature could be released with 4.0, the compatibility would be updated to 4.0.

alalek:
@xoox Well done!

Please update tutorial version to 4.0.

Squashed commit will look like this: https://github.com/alalek/opencv/commit/pr12772_r (if you want another commit message please add comment with updated version).

PR #12125, opened 2018-08-01
MobileNet-SSD v1 from TensorFlow with shared convolution weights

dkurt:
resolves #12114

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/501

This PR changes generation script adding new command line arguments. However we can easier extract them from a [config file](https://github.com/tensorflow/models/blob/master/research/object_detection/samples/configs/ssd_mobilenet_v1_ppn_shared_box_predictor_300x300_coco14_sync.config).

```
force_builders=Custom
docker_image:Custom=ubuntu-openvino:16.04
buildworker:Custom=linux-2
test_opencl:Custom=ON
```

PR #12775, opened 2018-10-08
Fix dnn caffe importer extract blobs from reused layers

radomsak:
### Fix bug when loading caffe network which reuses layers

We have been using OpenCV dnn module to run some of our caffe models (OpenCV version 3.4.1). After upgrading to 3.4.2 we noticed problems caused by #11461 

Our network reuses parameters from some of it's layers as described in our `deploy.prototxt` file (see the layer name in the lines highlighted by `HERE`):

```
layer {
  name: "conv1"  type: "Convolution"  bottom: "input_0"  top: "conv1_0"  <--- HERE
  param { name: "conv1_w"    lr_mult: 1    decay_mult: 1  }
  param { name: "conv1_b"   lr_mult: 2    decay_mult: 0  }
  convolution_param {    num_output: 48    kernel_size: 11    stride: 4
    weight_filler {      type: "gaussian"      std: 0.01    }
    bias_filler {      type: "constant"      value: 0    }
  }
}

...

layer {
  name: "conv1"  type: "Convolution"  bottom: "input_2"  top: "conv1_2"   <--- HERE
  param { name: "conv1_w"    lr_mult: 1    decay_mult: 1  }
  param { name: "conv1_b"   lr_mult: 2    decay_mult: 0  }
  convolution_param {    num_output: 48    kernel_size: 11    stride: 4
    weight_filler {      type: "gaussian"      std: 0.01    }
    bias_filler {      type: "constant"      value: 0    }
  }
}

```

The loop edited in `opencv/modules/dnn/src/caffe/caffe_importer.cpp` used to try to copy and clear the blob values as soon as it found matching layer name. Since our network has two layers with the same name, when trying to copy the blob values for the second `conv1` layer, the function would copy a cleared blob.

After the change the loop is searching for a layer with a matching name which does not have its blob values cleared.

Would you like me to write a separate unit test or include it into this pull request? 

Thank you very much for looking into this! Please let me know if you need any more details or if I could help somehow.

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/src/caffe/caffe_importer.cpp
+++ b/modules/dnn/src/caffe/caffe_importer.cpp
@@ -278,11 +278,13 @@ public:
         int li;
         for (li = 0; li != netBinary.layer_size(); li++)
         {
-            if (netBinary.layer(li).name() == name)
+            const caffe::LayerParameter& binLayer = netBinary.layer(li);
+            // Break if the layer name is the same and the blobs are not cleared
+            if (binLayer.name() == name && binLayer.blobs_size() != 0)
                 break;
         }
 
-        if (li == netBinary.layer_size() || netBinary.layer(li).blobs_size() == 0)
+        if (li == netBinary.layer_size())
             return;
 
         caffe::LayerParameter* binLayer = netBinary.mutable_layer(li);
--- a/modules/dnn/test/test_caffe_importer.cpp
+++ b/modules/dnn/test/test_caffe_importer.cpp
@@ -454,6 +454,28 @@ TEST(Test_Caffe, multiple_inputs)
     normAssert(out, first_image + second_image);
 }
 
+TEST(Test_Caffe, shared_weights)
+{
+  const string proto = findDataFile("dnn/layers/shared_weights.prototxt", false);
+  const string model = findDataFile("dnn/layers/shared_weights.caffemodel", false);
+
+  Net net = readNetFromCaffe(proto, model);
+
+  Mat input_1 = (Mat_<float>(2, 2) << 0., 2., 4., 6.);
+  Mat input_2 = (Mat_<float>(2, 2) << 1., 3., 5., 7.);
+
+  Mat blob_1 = blobFromImage(input_1);
+  Mat blob_2 = blobFromImage(input_2);
+
+  net.setInput(blob_1, "input_1");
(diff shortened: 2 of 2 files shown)


dkurt:
@radomsak, thank you!
We faced similar problem in https://github.com/opencv/opencv/pull/12125. There was a network from TensorFlow with shared convolution weights. Different layers are used a single weights blob.

However, according your changes, it looks like there are two copies of `conv1` weights and before PR parser tried to return released weights from the first layer but now it skips the first `conv1` and takes weights from the second one. Could you please do some research and compare binary weights data for both `conv1` entries?

radomsak:
@dkurt thanks a lot for quick reply!

I compared the binary data as you recommended. The first `conv1` data would be copied from the layer at `li==9`, the second one from `li==34`. 

I compared the weights data (caffe proto field `blobs`) and they are exactly the same. The only difference between the two `conv1` layers data were the values of `top` and `bottom` fields, which do not matter here.

Let me know if there is anything else I could do to help you

alalek:
As a bugfix this patch should [go](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) into 3.4 branch. We will merge changes from 3.4 into master regularly (weekly/bi-weekly).

So, please:

radomsak:
@alalek 

Thanks for the suggestion, I rebased the branch as you suggested

dkurt:
@radomsak, Got it, thank you!
So, let's protect you work against future changes by a test. Could you please create a simple two convolution layers network, initialize it's weights randomly by Caffe and put both `.prototxt` and `.caffemodel` files into https://github.com/opencv/opencv_extra/tree/master/testdata/dnn/layers.

Then in https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_caffe_importer.cpp add a test. You may use `TEST(Test_Caffe, multiple_inputs)` as an example.

Actually, we test importer's behavior so to exclude reference input/output data we can define a model as two sequential convolutions with 1x1 kernel initialized by a known value (see Caffe's fillers), in example (also do not forget for bias). So for a random input we know which output is expected.

radomsak:
@dkurt 

I added a test for my change. For the test files please refer to [opencv/opencv_extra#531](https://github.com/opencv/opencv_extra/pull/531)

Thanks!

PR #12636, opened 2018-09-24
samples: add winpack "drap & drop" build script

alalek:
```
force_builders=Docs
```

PR #12788, opened 2018-10-10
samples: add search for MSVS 2017 BuildTools

alalek:
relates #12636

Support configurations without full MSVS installations - where [BuildTools](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2017) is available only.


alalek:
:+1:

PR #12809, opened 2018-10-12
 canny edge detection in video

0xPrateek:
Detecting edges in video using canny edge detection algorithm.

### This pullrequest changes
This will add a new program to the open source describing how we can detect edges in a video using canny's edge detection algorithm.


alalek:
There is already canny sample: https://github.com/opencv/opencv/blob/3.4.3/samples/python/edge.py
Adding duplication of samples should be avoided.

PR #12354, opened 2018-08-30
core: add samples::findFile() function

alalek:
TODO:
- [x] update description
- [x] Winpack support: `cmake ... "-DOPENCV_INSTALL_DATA_DIR_RELATIVE=../../../../sources"`

```
tests_filter=*findFile*
```

PR #12813, opened 2018-10-12
Corrected data paths on many python tutorial code samples

swoodford:
### This pullrequest changes

I noticed many of the default data files in `samples/python/tutorial_code` had invalid paths to the data folder which would just throw an error if anyone tried to run the sample tutorial code without providing their own input files.  I believe I have found and corrected all of the data paths in this PR.
Thanks!

### DIFF
--- a/samples/python/tutorial_code/Histograms_Matching/histogram_calculation/calcHist_Demo.py
+++ b/samples/python/tutorial_code/Histograms_Matching/histogram_calculation/calcHist_Demo.py
@@ -6,7 +6,7 @@ import argparse
 
 ## [Load image]
 parser = argparse.ArgumentParser(description='Code for Histogram Calculation tutorial.')
-parser.add_argument('--input', help='Path to input image.', default='../data/lena.jpg')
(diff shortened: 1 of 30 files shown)


dkurt:
@swoodford, relative paths are relative. It all depends in which location you run python scripts.

swoodford:
@dkurt sure, but anyone looking at the sample code would not likely run `python samples/python/tutorial_code/Histograms_Matching/histogram_calculation/calcHist_Demo.py` from the root, but more likely would `cd samples/python/tutorial_code/Histograms_Matching/histogram_calculation/`, review the sample code, then run `python calcHist_Demo.py` which throws an error in nearly all of the code samples.

dkurt:
@swoodford, you're absolutely right. However it breaks samples for users who get used to run it by `python samples/python/tutorial_code/Histograms_Matching/histogram_calculation/calcHist_Demo.py` 😄 .

I think we need more complicated searching procedure like one proposed in https://github.com/opencv/opencv/pull/12354.

PR #12815, opened 2018-10-12
core: re-throw allocation exception if there is no fallback

### LINKED ISSUE #12812: No possibility to handle cv::OutOfMemoryError in cv::Mat::create() properly
When allocating a too large matrix / when having not enough RAM left
```
cv::Mat mat(1000000, 1000000, CV_8UC3, cv::Scalar::all(1));
```
The console output is
> OpenCV Error: Insufficient memory (Failed to allocate 2112827392 bytes) in cv::OutOfMemoryError, file C:\build\OpenCV\workspace\modules\core\src\alloc.cpp, line 52
OpenCV Error: Assertion failed (u != 0) in cv::Mat::create, file C:\build\OpenCV\workspace\modules\core\src\matrix.cpp, line 433

In the following code excerpt,
``a->allocate`` prints the OutOfMemoryError part to the console and throws an exception which is catched by ``CV_CATCH_ALL``. This one then fails in the assertion ``CV_Assert(u != 0); ``, printing the second console output. (btw, in my case ``a == a0``, so no second allocation was attempted)
https://github.com/opencv/opencv/blob/1ed9ff17e15ae0801da97167ef503ac88f294f77/modules/core/src/matrix.cpp#L358-L368

1. Bug: Why printing two error messages instead of "rethrowing"? Our application's exception handling just notices the fairly obscure ``u != 0`` instead of the actual cause
2. Strangeness: Why is the first allocation try catched, but the second one not? Actually, if the second allocation would've been attempted, the outOfMemoryException probably would've been propagated further
3. Feature: Although one can use ``cv::redirectError`` to set a custom error handler, it is almost impossible to synchronize the call of your own error callback with the area in your code you want to to error handling in. It is more for something like logging or showing a generic "your system is low on memory" message box to the user without the possibility to follow fall-backs for specific parts of your memory-intensive code. Of course, you could add try{}/catch around every single matrix constructor but this becomes quite hard to maintain

alalek:
resolves #12812

### DIFF
--- a/modules/core/src/matrix.cpp
+++ b/modules/core/src/matrix.cpp
@@ -355,15 +355,16 @@ void Mat::create(int d, const int* _sizes, int _type)
 #endif
         if(!a)
             a = a0;
-        CV_TRY
+        try
         {
             u = a->allocate(dims, size, _type, 0, step.p, 0, USAGE_DEFAULT);
             CV_Assert(u != 0);
         }
-        CV_CATCH_ALL
+        catch (...)
         {
-            if(a != a0)
-                u = a0->allocate(dims, size, _type, 0, step.p, 0, USAGE_DEFAULT);
+            if (a == a0)
+                throw;
+            u = a0->allocate(dims, size, _type, 0, step.p, 0, USAGE_DEFAULT);
             CV_Assert(u != 0);
         }
         CV_Assert( step[dims-1] == (size_t)CV_ELEM_SIZE(flags) );


PhilLab:
@alalek Yes, this then allows to check for outOfMemory in the application code with ``if (e.code == CV_StsNoMem)``

alalek:
:+1:

stiv-yakovenko:
HI! Can anyone be so kind to share code sample of handling out of memory error in opencv?

PR #12811, opened 2018-10-12
resolves #11283

take1014:
resolves #11283

PR #12823, opened 2018-10-13
change area() emptiness checks to empty()

tompollok:
relates #12811 

### This pullrequest changes

changes area() emptiness checks to use empty()

cases:
 .area() == 0 to empty()
 .area() != 0 to !empty()
 .area() > 0 to  !empty()

_src.size().area() > 0 to !_src.empty() //no need to create size here

if( r.area() ) to if( !r.empty() )


tompollok:
@alalek let me know if this is fine so i can then also prepare a pr for opencv-contrib

alalek:
Well done! Looks good to me.

Issue #8413, opened 2017-03-17
Element-wise division of CV_64F matrices outputs inf instead of 0


taketwo:
##### System information (version)

- OpenCV => 3.2
- Operating System / Platform => Ubuntu 16.04
- Compiler => GCC 5.4.0

##### Detailed description

The documentation of [`divide`](http://docs.opencv.org/master/d2/de8/group__core__array.html#ga1f96b569cac4c286642b34eff098138e) says that

> `dst(I) = saturate(src1(I) * scale / src2(I))`
> When `src2(I)` is zero, `dst(I)` will also be zero.

In other words, it should be safe to divide by zero. However, the behavior I observe with `CV_64F` matrices is not consistent and depends on the size of the matrices.

For small matrices (up to 3 elements) division outputs zeros, as expected. However, if the size of the matrices is 4 or more, division outputs `inf`s.

Other data types behave as expected.

##### Steps to reproduce

```.cpp
#include <iostream>
#include <opencv2/core.hpp>

int main(int argc, const char** argv)
{
  cv::Mat A = cv::Mat::ones(1, 4, CV_64FC1);
  cv::Mat B = cv::Mat::zeros(1, 4, CV_64FC1);
  std::cout << "A = " << A << std::endl;
  std::cout << "B = " << B << std::endl;
  std::cout << "A / B = " << cv::Mat(A / B) << std::endl;
  return 0;
}
```

Output is:

```
A = [1, 1, 1, 1]
B = [0, 0, 0, 0]
A / B = [inf, inf, inf, inf]
```

##### Additional info

```
$ cat /proc/cpuinfo
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 61
model name      : Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz
stepping        : 4
microcode       : 0x24
cpu MHz         : 2401.312
cache size      : 4096 KB
physical id     : 0
siblings        : 4
core id         : 0
cpu cores       : 2
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 20
wp              : yes
[...]

PR #12825, opened 2018-10-14
(3.4) divide() with zero values

alalek:
Added divide() zero values test.

relates #8413

---

OpenCL patch is not related to this patch directly, but it is necessary for 'master' patch (based on these commits).


vpisarev:
:+1:

Issue #11039, opened 2018-03-08
[evolution] Stitching Module


hrnr:
## Stitching Module

* Author: Jiri Horner
* Link: TBD
* Status: **Draft**
* Platforms: **All**
* Complexity: 1-2 man-weeks

## Introduction and Rationale

Stitching module has been around for quite some time and hasn't been refreshed to match changes in rest of the OpenCV.

Stitcher interface is widely used from Python, but lots of features can't be controlled from Python. User experience should be improved.

There is no interoperability with features2d module, all feature finders need to be wrapped manually for stitching module.

Dependencies on legacy CUDA stuff in stitching module let to some compilation problems recently.

Dependencies on xfeatures2d makes it impossible for distributions to provide patent-free stitching module while offering patented algorithms to use with  for users that need them.

## Proposed solution

* remove `detail::FeaturesFinder` and `detail::FeaturesMatcher` wrapper classes. Use `cv::Feature2D` and `cv::DescriptorMatcher` from features2d module. This will allow to use all features and descriptors with stitching modules. Move all estimation code from `detail::FeaturesMatcher` to `detail::Estimator`. This also solves problems with dependencies on xfeatures2d.

* allow user to explicitly set `detail::Estimator`

* consolidate creation methods for stitcher. Keep only `Stitcher::create`, wrap this method for Python. Drop `try_use_gpu` flag, rely on T-API or user explicitly setting CUDA classes to use.

* allow users to set FeaturesFinder, DescriptorMatcher, BundleAdjuster etc. from Python

* drop legacy usage of legacy Levenberg–Marquardt API from bundle adjusters. Replace with new LevMarq API.

* remove direct usage of CUDA from stitching module (OpenCL alternatives are already available). Move to CUDA modules where it makes sense.

## Impact on existing code, compatibility

Changes in Stitcher interface might require some changes in user code. We could introduce some wrapper classes and aliases to ease transition towards using new API with classes from features2d.

## Possible alternatives

Keep as is.

## References

https://github.com/opencv/opencv/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3A%22category%3A+gpu+%28cuda%29%22+stitching

cc @vpisarev

kinchungwong:
I'm looking for information on making the Stitcher class more usable (in C++) for:
* Large number of images (can't load all of them into memory for processing)
* Persisting to disk between steps (for example, save image features and close the application, then open the application again and load the image features to continue)
[...]

PR #12827, opened 2018-10-14
[evolution] Stitching for OpenCV 4.0

hrnr:
I have received a very little feedback on evolution proposal #11039. I have decided to come with some code, to start the discussion. Here are some minimal changes in the stitching module, that I think should go to OpenCV 4.0. Large parts of the evolution proposal are not implemented and I'm awaiting further feedback.

- [ ] remove `detail::FeaturesFinder` and `detail::FeaturesMatcher` wrapper classes. Use `cv::Feature2D` and `cv::DescriptorMatcher` from features2d module. This will allow to use all features and descriptors with stitching modules. Move all estimation code from `detail::FeaturesMatcher` to `detail::Estimator`. This also solves problems with dependencies on xfeatures2d.
- [x] allow user to explicitly set `detail::Estimator`
- [x] consolidate creation methods for stitcher. Keep only `Stitcher::create`, wrap this method for Python. Drop `try_use_gpu` flag, rely on T-API or user explicitly setting CUDA classes to use.
- [ ] allow users to set FeaturesFinder, DescriptorMatcher, BundleAdjuster etc. from Python
- [ ] drop legacy usage of legacy Levenberg–Marquardt API from bundle adjusters. Replace with new LevMarq API.
- [ ] remove direct usage of CUDA from stitching module (OpenCL alternatives are already available). Move to CUDA modules where it makes sense.

```
force_builders=linux,Linux OpenCL,windows,macosx,docs,Custom,Win64 OpenCL
**WIP**
```

### DIFF
--- a/modules/features2d/src/feature2d.cpp
+++ b/modules/features2d/src/feature2d.cpp
@@ -77,18 +77,18 @@ void Feature2D::detect( InputArrayOfArrays images,
 {
     CV_INSTRUMENT_REGION();
 
-    size_t nimages = images.total();
+    int nimages = (int)images.total();
 
     if (!masks.empty())
     {
-        CV_Assert(masks.total() == nimages);
+        CV_Assert(masks.total() == (size_t)nimages);
     }
 
     keypoints.resize(nimages);
 
     if (images.isMatVector())
     {
-       for (size_t i = 0; i < nimages; i++)
+       for (int i = 0; i < nimages; i++)
        {
            detect(images.getMat(i), keypoints[i], masks.empty() ? noArray() : masks.getMat(i));
        }
@@ -96,7 +96,7 @@ void Feature2D::detect( InputArrayOfArrays images,
     else
     {
         // assume UMats
-        for (size_t i = 0; i < nimages; i++)
+        for (int i = 0; i < nimages; i++)
         {
             detect(images.getUMat(i), keypoints[i], masks.empty() ? noArray() : masks.getUMat(i));
         }
@@ -134,15 +134,15 @@ void Feature2D::compute( InputArrayOfArrays images,
     if( !descriptors.needed() )
         return;
 
-    size_t nimages = images.total();
+    int nimages = (int)images.total();
 
-    CV_Assert( keypoints.size() == nimages );
+    CV_Assert( keypoints.size() == (size_t)nimages );
     // resize descriptors to appropriate size and compute
     if (descriptors.isMatVector())
     {
         vector<Mat>& vec = *(vector<Mat>*)descriptors.getObj();
         vec.resize(nimages);
-        for (size_t i = 0; i < nimages; i++)
+        for (int i = 0; i < nimages; i++)
         {
             compute(images.getMat(i), keypoints[i], vec[i]);
         }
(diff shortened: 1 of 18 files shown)


hrnr:
I have dropped `try_use_gpu` flag. OCL will be utilised automatically in enabled paths. To use CUDA paths user needs to explicitly set CUDA classes in stitching.

 What do you think about dropping `detail::FeaturesFinder`, `detail::FeaturesMatcher` and the wrapper code in favour of `cv::Feature2D`, `cv::DescriptorMatcher` and classes from features2d? Should I go forward with that? That will be a big change and quite a lot of work, so I want some approval before I continue.

alalek:
> favour of cv::Feature2D, cv::DescriptorMatcher and classes from features2d

Sounds good.

LaurentBerger:
May be too late but in this [example ](https://github.com/LaurentBerger/Traitement_d_images_et_de_videos_avec_OpenCV3/blob/master/ch5PanoCam/main.cpp)I can save all results of stitching process except ExposureCompensator. Is it possible to add a method to get ExposureCompensator result?

hrnr:
I don't get what you want to achieve. There is `ExposureCompensator::apply` to get the results of exposure compensation. What are you missing? Is there any other info that is not exposed?

LaurentBerger:
I need to investigate more but I think it is [gain_maps](https://github.com/opencv/opencv/blob/master/modules/stitching/src/exposure_compensate.cpp#L234-L237)
You can everything in yml file except results of those lines : 
```
       pp.algoCorrectExpo = ExposureCompensator::createDefault(pp.correctionExposition);
        pp.algoCorrectExpo->feed(pp.posCoin, imagesProjetees, pp.masqueCompo);

``` 
Example I have two webcam and I stitch images. I save everything . Then I can run my program without any initialize stitching (except ExposureCompensator)

hrnr:
I have rebased the brach due to merge conflict.

vpisarev:
:+1: can be merged as soon as buildbot becomes green

hrnr:
Thanks for the review. You are right, @alalek, `INTER_NEAREST` is the correct interpolation there. Fixed. Many thanks for fixing the issue in SURF.

Squashed fixup commits, rebased.

As you know I wanted a bigger scope for this, but I don't have much of a free time nowadays to put into this as I'm starting in a new job. At least, I'd love to see something similar done for `detail::FeaturesMatcher`, but I'm not sure if I find a free day for coding to take on that. Anyone from the community please feel free to take this further, I can help if you run into any issues, just mention me so I receive an email.

For the time being, I'm glad that at least this part goes in.

LaurentBerger:
@hrnr I can work on stitching module. I already wrote an[ example in C++](https://github.com/LaurentBerger/Traitement_d_images_et_de_videos_avec_OpenCV3/tree/master/ch5PanoCam). I need to translate this example in python. I need your PR. If something is missing I can make a new PR.

hrnr:
Sure go for it. My plan for `detail::FeaturesMatcher` was to replace it with `cv::DescriptorMatcher` everywhere, then delete almost all the code in the stitching module related to matches. One caveat there is that the transform estimation code (now sitting in the high-level Matchers) needs to moved to `detail::Estimator`, which will be solely responsible for the geometry model used. Apart from that it should be fairly straightforward job, requiring only small amount of the glue code to do the parallel matching - the code is already there, probably only needs to be moved around.

PR #12841, opened 2018-10-15
G-API: Introduce new `reshape()` API.

dmatveev:
This new method allows to compile a G-API graph one and then adjust it
to new input size(s). The support is determined by the underlying backends
capabilities - if a `GCompiled` object has been produced with backends
supporting `reshape()`, adjustments are made; an exception is thrown
otherwise.

Currently only Fluid backend supports reshape capability, and CPU (OpenCV)
backend throws. OpenCV backend reshaping will be added later.


dmatveev:
Closing, @rgarnov will reopen one

PR #12846, opened 2018-10-15
videoio: return -1 is incorrect for function with 'bool' result

### LINKED ISSUE #12834: Should be return false
https://github.com/opencv/opencv/blob/5115e5decbef657ceb234d21ad8e8e33a6c96ca4/modules/videoio/src/cap_v4l.cpp#L1746

alalek:
resolves #12834


alalek:
:+1:

PR #12848, opened 2018-10-15
core: update solveLP() interface

### LINKED ISSUE #12337: solveLP() doesn't check the argument type

alalek:
resolves #12337


alalek:
:+1:

PR #12849, opened 2018-10-15
Update test_camera.cpp

wanghanmin:
This property setting is for special H/W like avermedia cv710


wanghanmin:
Hi alalek, do you mean to create new test function or insert code after general test code in "TEST(DISABLED_VideoIO_Camera, basic)" ? And as test for camera, should we specify the resolution and format ?  If the setting not match the signal, the result maybe black or some still image as manufacturer preset.

alalek:
> after general test code

Right.

> should we specify the resolution and format ?

Usually VideoCapture backend should select some defaults which just works.
Also you can add separate test with properties settings, but please specify what camera H/W and configuration is used.

PR #12852, opened 2018-10-16
Bugfix/increase magic threshold for perf test

### LINKED ISSUE #12797: Test Failures in OpenCV 3 on NVIDIA Jetson TX2 (related to gcc-7 vs gcc < 7)
I built OpenCV 3.3.1 with OpenCV_Extra on NVIDIA's Jetson TX2 platform like the following with CUDA disabled:

    cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_PNG=OFF -DBUILD_TIFF=OFF -DBUILD_TBB=OFF -DBUILD_JPEG=OFF -DBUILD_JASPER=OFF -DBUILD_ZLIB=OFF -DBUILD_EXAMPLES=ON -DBUILD_opencv_java=OFF -DBUILD_opencv_python2=ON -DBUILD_opencv_python3=ON -DENABLE_PRECOMPILED_HEADERS=OFF -DWITH_CUDA=OFF -DWITH_OPENCL=OFF -DWITH_OPENMP=OFF -DWITH_FFMPEG=ON -DWITH_GSTREAMER=OFF -DWITH_GSTREAMER_0_10=OFF -DWITH_GTK=ON -DWITH_VTK=OFF -DWITH_TBB=ON -DWITH_1394=OFF -DWITH_OPENEXR=OFF -DINSTALL_C_EXAMPLES=ON -DINSTALL_TESTS=ON -DCPACK_GENERATOR_DEB=ON -DOPENCV_TEST_DATA_PATH=../opencv_extra/testdata ../opencv

And ran OpenCV test script, and found two test failures on 'Calib3d_Affine3f.accuracy' and 'match_bestOf2Nearest.bestOf2Nearest' testing like the following:

    [opencv_test_calib3d] RUN : /usr/bin/opencv_test_calib3d --perf_min_samples=1 --perf_force_samples=1 --gtest_output=xml:opencv_test_calib3d.xml
    [opencv_test_calib3d] CTEST_FULL_OUTPUT
    [opencv_test_calib3d] OpenCV version: 3.3.1
    [opencv_test_calib3d] OpenCV VCS version: 3.3.1
    [opencv_test_calib3d] Build type: release
    [opencv_test_calib3d] Parallel framework: tbb
    [opencv_test_calib3d] CPU features: neon fp16
    
    [opencv_test_calib3d] [ RUN      ] Calib3d_Affine3f.accuracy
    [opencv_test_calib3d] /home/nvidia/build-opencv/opencv/modules/calib3d/test/test_affine3.cpp:57: Failure
    [opencv_test_calib3d]       Expected: 0
    [opencv_test_calib3d] To be equal to: cvtest::norm(cv::Mat(affine.matrix, false).colRange(0, 3).rowRange(0, 3) != expected, cv::NORM_L2)
    [opencv_test_calib3d]       Which is: 441.673
    [opencv_test_calib3d] [  FAILED  ] Calib3d_Affine3f.accuracy (0 ms)
    
    [opencv_perf_stitching] RUN : /usr/bin/opencv_perf_stitching --perf_min_samples=1 --perf_force_samples=1 --gtest_output=xml:opencv_perf_stitching.xml
    [opencv_perf_stitching] Time compensation is 0
    [opencv_perf_stitching] CTEST_FULL_OUTPUT
    [opencv_perf_stitching] OpenCV version: 3.3.1
    [opencv_perf_stitching] OpenCV VCS version: 3.3.1
    [opencv_perf_stitching] Build type: release
    [opencv_perf_stitching] Parallel framework: tbb
    [opencv_perf_stitching] CPU features: neon fp16
    
    [opencv_perf_stitching] [----------] 1 test from match_bestOf2Nearest
    [opencv_perf_stitching] [ RUN      ] match_bestOf2Nearest.bestOf2Nearest/0
    [opencv_perf_stitching]  Expected: 
    [opencv_perf_stitching] [0.9970975582816909, 0.01136054503288174;
    [opencv_perf_stitching]  -0.002557125266879237, 1.02781673911756;
    [opencv_perf_stitching]  0.0002463026627945361, -1.679576661348132e-05]
    [opencv_perf_stitching]  Actual:
    [opencv_perf_stitching] [0.9986402872172184, -0.01796581492545124;
    [opencv_perf_stitching]  -0.00230781842425846, 1.031312571169278;
    [opencv_perf_stitching]  0.0002375978666932171, 9.189439617496881e-05]
    [opencv_perf_stitching] /home/nvidia/build-opencv/opencv/modules/ts/src/ts_perf.cpp:571: Failure
    [opencv_perf_stitching] Failed
    [opencv_perf_stitching]   Difference (=0.029326359958332979) between argument1 "R" and expected value is greater than 0.014999999999999999
    [opencv_perf_stitching] params    =       "orb"
    [opencv_perf_stitching] termination reason:  reached maximum number of iterations
    [opencv_perf_stitching] bytesIn   =      96896
    [opencv_perf_stitching] bytesOut  =          0
    [opencv_perf_stitching] samples   =          1
    [opencv_perf_stitching] outliers  =          0
    [opencv_perf_stitching] frequency = 1000000000
    [opencv_perf_stitching] min       =  306334005 = 306.33ms
    [opencv_perf_stitching] median    =  306334005 = 306.33ms
    [opencv_perf_stitching] gmean     =  306334005 = 306.33ms
    [opencv_perf_stitching] gstddev   = 0.00000000 = 0.00ms for 97% dispersion interval
    [opencv_perf_stitching] mean      =  306334005 = 306.33ms
    [opencv_perf_stitching] stddev    =          0 = 0.00ms
    [opencv_perf_stitching] [  FAILED  ] match_bestOf2Nearest.bestOf2Nearest/0, where GetParam() = "orb" (577 ms)
    [opencv_perf_stitching] [----------] 1 test from match_bestOf2Nearest (577 ms total)

I tried both Ubuntu 16.04-based image and Ubuntu 18.04-based image with different GCC versions. And below is the summary of my findings:
* The OpenCV test failure on  ‘Calib3d_Affine3f.accuracy’ was observed on different platforms w/ different images whenever the toolchain is based on gcc-7. And in any platform + image combination, this test always passes when gcc < 7 (i.e., lower than version 7, like gcc-6.4.0) is used.
* The OpenCV test failure on ‘match_bestOf2Nearest.bestOf2Nearest’ was persistently observed whatever image and whatever version of gcc used.

My observation is that the failure on  ‘Calib3d_Affine3f.accuracy’ test is because the test script compares the expected and the actual values with an equality sign (i.e., ==), which is to compare the exact values, not assuming any variations in floating-point operations in different platform with different compilers. With gcc < 7, the floating-point comparison test passes, but with gcc-7, it fails due to the two floating-point values are slightly different like 1e-17.

Questions:

 - What would be the rationale behind the comparison method in OpenCV test script (i.e., exact value comparison instead of compare with a tolerable margin)?
 - What's the ARM FPCR settings assumed in the testing? Interestingly, FPCR value on my workstation is 0x37f, while FPCR on Jetson TX2 is 0x0. Should I change FPCR on TX2 to some particular value?   
 - Why would the floating-point precision behavior be different when using gcc-7 compared to gcc < 7?
 - How to make the test passed with gcc-7 and moving forward?

The second failure case on ‘match_bestOf2Nearest.bestOf2Nearest’ is also related to the precision handling in floating-point operation in some of OpenCV functions. I found out that OpenCV functions sometimes cast the values between CV_32F and CV_64F internally. Assuming the CV_32F format is enough to handle the precision, this shouldn’t matter. For this case, the test script compares the expected and the actual values with a tolerable margin like an epsilon (instead of exact value equality testing). The observation is that the test fails depending on the random seed provided in the functions. The default random seed is set as '43' as in `cvflann::seed_random(42);`. But, if I change the random seed to a different one like '2', then the test passes. This implies that the pre-defined epsilon (which is 0.15 now) is not large enough to accommodate the new platform, or the floating-point operations on our CPU has variability larger than what is assumed in OpenCV test script.
 
Question:

 - What's the assumptions and/or rationale behind (1) the pre-defined threshold/epsilon like 0.15 and (2) the random seed value of 42? Can we change that?

nangchoo:
### This pullrequest resolves #12797 

There has been a persistent failure of ‘match_bestOf2Nearest.bestOf2Nearest’ test on NVIDIA Jetson TX2.
This PR increases the acceptable error margin in such a perf test to allow the variability in floating-point operations in newer platforms/compilers like Jetson TX2. The error margin is increased to 0.06 from 0.015.

### DIFF
--- a/modules/stitching/perf/perf_matchers.cpp
+++ b/modules/stitching/perf/perf_matchers.cpp
@@ -103,7 +103,7 @@ PERF_TEST_P( match, bestOf2Nearest, TEST_DETECTORS)
     Mat R (pairwise_matches.H, Range::all(), Range(0, 2));
     // separate transform matrix, use lower error on rotations
     SANITY_CHECK(dist, 1., ERROR_ABSOLUTE);
-    SANITY_CHECK(R, .015, ERROR_ABSOLUTE);
+    SANITY_CHECK(R, .06, ERROR_ABSOLUTE);
 }
 
 PERF_TEST_P( matchVector, bestOf2NearestVectorFeatures, testing::Combine(


alalek:
> Commits 2
> Merge pull request

Please remove unnecessary merge commit from the patch. It may break `git bisect`.

PR #12858, opened 2018-10-16
Catch exceptions by const-reference

janisozaur:
Exceptions caught by value incur needless cost in C++, most of them can
be caught by const-reference, especially as nearly none are actually
used. This could allow compiler generate a slightly more efficient code.


alalek:

So, please:

janisozaur:
Updated

tompollok:
For derived or custom exception types it makes sense, but for primitives like int?

janisozaur:
https://godbolt.org/z/5uQQvw

It might be over the top, but it doesn't cause any negative impact, even if the chance to generate any better code is slim.

PR #12870, opened 2018-10-18
G-API: Implement basic heterogeneity plugin for Fluid backend

dmatveev:
This PR introduces basic heterogeneity support for Fluid backend.

Now it can co-exist and run together with OpenCV plugin within a single graph so the graph becomes hybrid: executing both frame-by-frame (OpenCV) and row-by-row (Fluid) kernels.

<cut/>

For example, the following test case

    GAPIBigHeteroTest/GAPIBigHeteroTest.Test/120, where GetParam() = { 0, 0, 1, 1, 1, 1, 0, 0, 0 }

Produces the following graph which now works:

hetero_graph

There's a limitation that "output ROI" feature of Fluid backend can't work if other backends are available, it will be streamlined till the end of the year (not in the OpenCV 4.0 though)


dmatveev:
Thanks!

PR #12871, opened 2018-10-18
flann: Add documentation for Distance functor

### LINKED ISSUE #12847: Lack of documentation on `Distance` functor
- OpenCV => 3.4
- Operating System / Platform => N/A
- Compiler => N/A

##### Detailed description

In the flann module, class `GenericIndex` and function `hierarchicalClustering` are templated over a `Distance` functor. However, the [documentation](https://docs.opencv.org/3.4/dc/de5/group__flann.html) does not mention:

1. The list of available `Distance` functors (defined in [dist.h](https://github.com/opencv/opencv/blob/3.4/modules/flann/include/opencv2/flann/dist.h))

2. The specification of `Distance` in case a custom metric is needed. It can be found on page 7 of the [flann user manual](https://www.cs.ubc.ca/research/flann/uploads/FLANN/flann_manual-1.8.4.pdf).

UnderscoreAsterisk:
Resolves #12847

### This pullrequest changes
Added documentation for Distance functor in flann module. However, I am not sure if the "Detailed Description" section of `GenericIndex` is the most appropriate place for it. I would like some feedback in this regard.

### DIFF
--- a/modules/flann/include/opencv2/flann.hpp
+++ b/modules/flann/include/opencv2/flann.hpp
@@ -137,6 +137,22 @@ __cv::flann::HellingerDistance__ - The Hellinger distance functor.
 __cv::flann::ChiSquareDistance__ - The chi-square distance functor.
 
 __cv::flann::KL_Divergence__ - The Kullback-Leibler divergence functor.
+
+Although the provided implementations cover a vast range of cases, it is also possible to use
+a custom implementation. The distance functor is a class whose `operator()` computes the distance
+between two features. If the distance is also a kd-tree compatible distance, it should also provide an
+`accum_dist()` method that computes the distance between individual feature dimensions.
+
+In addition to `operator()` and `accum_dist()`, a distance functor should also define the
+`ElementType` and the `ResultType` as the types of the elements it operates on and the type of the
+result it computes. If a distance functor can be used as a kd-tree distance (meaning that the full
+distance between a pair of features can be accumulated from the partial distances between the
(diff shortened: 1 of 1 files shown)


modules/flann/include/opencv2/flann.hpp
```diff
@@ -103,6 +103,56 @@ using ::cvflann::KL_Divergence;
 
 /** @brief The FLANN nearest neighbor index class. This class is templated with the type of elements for which
 the index is built.
+
+`Distance` functor specifies the metric to be used to calculate the distance between two points.
+There are several `Distance` functors that are readily available:
+
+__cv::flann::L2_Simple__ - Squared Euclidean distance functor. This is the simpler, unrolled version.
```
alalek: Please use proper doxygen links instead. Try `#` prefix before symbol names.

UnderscoreAsterisk: Since the symbols are defined in the `cvflann` namespace, they only turn to links if I prefix them with `cvflan::` as in `cvflann::L2_Simple`. I did not use these links because I wasn't sure if `cvflann` namespace should be exposed.

alalek: Right. It seems that doxygen can't handle "using" statements.

This should work:
```
@link cvflann::L2_Simple cv::flann::L2_Simple @endlink
```

P.S. Direct usage of `cvflann` namespace should be avoided in user code (it is just a copy of original flann).

UnderscoreAsterisk: Thank you for the suggestion.

PR #12878, opened 2018-10-19
refactoring catching all exceptions as const ref

tompollok:
relates #12858

### This pullrequest changes

refactored codebase such that exception handling is always using const ref

Now the codebase should be the same everywhere, except the files in 3rdparty. I didnt know if im allowed to touch them.

```
allow_multiple_commits=1
```


tompollok:
@alalek are the CV_TRY and CV_CATCH macros still necessary or should their use disappear slowly from the codebase?

tompollok:
Thanks @alalek for merging and doing the work on CV_TRY
I have also added the same PR for opencv_contrib.

Issue #7989, opened 2017-01-10
Wrong HAMMING2 distance ?


iabdalkader:
##### System information (version)
- OpenCV => 2.4 and 3.2
- Operating System / Platform => Linux 64 Bit
- Compiler => GCC

##### Detailed description
For binary descriptors that output 2 bits per bin/element the `NORM_HAMMING2` distance should be used (especially for ORB with WTA_K== 3 or 4). The current [popcnt32/64](https://github.com/opencv/opencv/blob/2.4/modules/flann/include/opencv2/flann/dist.h#L520
) count every set bit, so for each different bin/element/2bits it counts 2 different elements instead of 1 as expected, this outputs the wrong distance, for example if an XOR'd byte result is say `11111111` the expected distance is 4, the distance with popcnt is 8, which could overlook a good match. 

I can send a PR to fix this issue, if I'm not missing anything please let me know and I'll send it.

StevenPuttemans:
> I can send a PR to fix this issue, if I'm not missing anything please let me know and I'll send it.

Go ahead! A developer will take a look at it and provide comments directly in the PR!

alalek:
Good catch!
But "flann" is external 3rdparty library. It is used partially only, and mentioned part of code is out of scope of OpenCV library interests.
Also this code is gone from upstream repository, so fixing this doesn't make a big sense.
Lets close it for now.

PR #12881, opened 2018-10-20
ORB: compute default norm by wta_k

huangqinjin:
relates #7989


huangqinjin:
Think for your hints @alalek .

PR #12886, opened 2018-10-21
resolved fatal error during initial compilation

Xenos5282:
fatal error is encountered when compiling OpenCV for Python which is 'fatal error hdf5.h no such file or directory'.


alalek:
All external 3rdparty dependencies should be optional.

It is better to fill issue with complete build information:
- used OpenCV version
- exact error message
- failed command line via `make VERBOSE=1` (without any `-j` option)
- CMake stdout (from clean build)
- files from build folder: CMakeCache.txt and CMakeVars.txt (as attachment or use "grep" for related problem keyword)
- include files `headers.txt` `pyopencv_custom_headers.h`

PR #12893, opened 2018-10-22
V4L-2: Added camera properties mostly the same as on MSMF.

aleksey-nikolaev:
I made this changes because when I tried to get the exposure property I got the `INF`. Also, all values for my cam on windows are not same on linux. 

### This pullrequest changes

1. Removed all conversion parameters to/from [0,1] range.
2. Added all known conversion from V4L2_CID_* to CV_CAP_PROP_*
3. Added selecting input channel by CV_CAP_PROP_MODE. Default value -1 the channels not changed.
4. Fixed use of video buffer index.
5. Removed extra memcopy for grab image.

Some code refactoring:
1. Removed device closing from autosetup_capture_mode_v4l2
2. Refactoring of Convert To RGB, Set/Get of properties, frame capture initialization.
3. Added cap_properties verbalization.
4. Removed extra code, like querying or enumerating all device parameters or an extra `{}`.

I tried to keep the code format unchanged. It will be nice to have `.clang-format` on the OpenCV project.

### DIFF
--- a/modules/videoio/src/cap_v4l.cpp
+++ b/modules/videoio/src/cap_v4l.cpp
@@ -263,6 +263,8 @@ struct buffer
 {
     void *  start;
     size_t  length;
+    // This is dequeued buffer. It used for to put it back in the queue. 
+    v4l2_buffer buffer;
 };
 
 struct CvCaptureCAM_V4L CV_FINAL : public CvCapture
@@ -440,10 +442,6 @@ static int autosetup_capture_mode_v4l2(CvCaptureCAM_V4L* capture) {
             return 0;
         }
     }
-
-    fprintf(stderr,
-            "VIDEOIO ERROR: V4L2: Pixel format of incoming image is unsupported by OpenCV\n");
-    icvCloseCAM_V4L(capture);
     return -1;
 }
 
@@ -549,13 +547,16 @@ static int _capture_V4L2 (CvCaptureCAM_V4L *capture)
     capture->form.type = V4L2_BUF_TYPE_VIDEO_CAPTURE;
 
     if (-1 == ioctl (capture->deviceHandle, VIDIOC_G_FMT, &capture->form)) {
-        fprintf( stderr, "VIDEOIO ERROR: V4L2: Could not obtain specifics of capture window.\n\n");
+        fprintf( stderr, "VIDEOIO ERROR: V4L2: Could not obtain specifics of capture window.\n");
         icvCloseCAM_V4L(capture);
         return -1;
     }
 
-    if (autosetup_capture_mode_v4l2(capture) == -1)
+    if (autosetup_capture_mode_v4l2(capture) == -1) {
+        fprintf(stderr, "VIDEOIO ERROR: V4L2: Pixel format of incoming image is unsupported by OpenCV\n");
+        icvCloseCAM_V4L(capture);
         return -1;
+    }
 
     /* try to set framerate */
     v4l2_set_fps(capture);
@@ -744,7 +745,7 @@ static int read_frame_v4l2(CvCaptureCAM_V4L* capture) {
             {
                 if (ioctl(capture->deviceHandle, VIDIOC_QBUF, &buf) == -1)
                 {
-                    return 0;
+                    return -1;
                 }
             }
             return 0;
@@ -758,20 +759,14 @@ static int read_frame_v4l2(CvCaptureCAM_V4L* capture) {
     }
 
     assert(buf.index < capture->req.count);
+    assert(capture->buffers[buf.index].length == buf.length);
 
-    memcpy(capture->buffers[MAX_V4L_BUFFERS].start,
-            capture->buffers[buf.index].start,
-            capture->buffers[MAX_V4L_BUFFERS].length );
-    capture->bufferIndex = MAX_V4L_BUFFERS;
-    //printf("got data in buff %d, len=%d, flags=0x%X, seq=%d, used=%d)\n",
-    //    buf.index, buf.length, buf.flags, buf.sequence, buf.bytesused);
+    //We shouldn't use this buffer in the queue while not capture image from it.
+    capture->buffers[buf.index].buffer = buf;
+    capture->bufferIndex = buf.index;
 
     //set timestamp in capture struct to be timestamp of most recent frame
     capture->timestamp = buf.timestamp;
-
-    if (-1 == ioctl (capture->deviceHandle, VIDIOC_QBUF, &buf))
-        perror ("VIDIOC_QBUF");
-
     return 1;
 }
 
@@ -1374,7 +1369,9 @@ static IplImage* icvRetrieveFrameCAM_V4L( CvCaptureCAM_V4L* capture, int) {
     }
 
     if(!capture->convert_rgb) {
-        capture->frame.imageData = (char*)capture->buffers[capture->bufferIndex].start;
+        capture->frame.imageData = (char *)capture->buffers[MAX_V4L_BUFFERS].start;
+        memcpy(capture->buffers[MAX_V4L_BUFFERS].start, capture->buffers[capture->bufferIndex].start,
+               std::min(capture->buffers[MAX_V4L_BUFFERS].length, capture->buffers[capture->bufferIndex].length));
         return &capture->frame;
     }
 
@@ -1440,13 +1437,13 @@ static IplImage* icvRetrieveFrameCAM_V4L( CvCaptureCAM_V4L* capture, int) {
     case V4L2_PIX_FMT_SGBRG8:
         sgbrg2rgb24(capture->form.fmt.pix.width,
                 capture->form.fmt.pix.height,
-                (unsigned char*)capture->buffers[(capture->bufferIndex+1) % capture->req.count].start,
+                (unsigned char*)capture->buffers[capture->bufferIndex].start,
                 (unsigned char*)capture->frame.imageData);
         break;
     case V4L2_PIX_FMT_RGB24:
         rgb24_to_rgb24(capture->form.fmt.pix.width,
                 capture->form.fmt.pix.height,
-                (unsigned char*)capture->buffers[(capture->bufferIndex+1) % capture->req.count].start,
+                (unsigned char*)capture->buffers[capture->bufferIndex].start,
                 (unsigned char*)capture->frame.imageData);
         break;
     case V4L2_PIX_FMT_Y16:
@@ -1847,7 +1844,11 @@ bool CvCaptureCAM_V4L::grabFrame()
 
 IplImage* CvCaptureCAM_V4L::retrieveFrame(int)
 {
-    return icvRetrieveFrameCAM_V4L( this, 0 );
+    IplImage *image = icvRetrieveFrameCAM_V4L(this, 0);
+    //Revert buffer to the queue 
+    if (-1 == ioctl(deviceHandle, VIDIOC_QBUF, &buffers[bufferIndex].buffer))
+        perror("VIDIOC_QBUF");
+    return image;
 }
 
 double CvCaptureCAM_V4L::getProperty( int propId ) const
--- a/modules/videoio/src/cap_v4l.cpp
+++ b/modules/videoio/src/cap_v4l.cpp
@@ -285,11 +285,12 @@ struct CvCaptureCAM_V4L CV_FINAL : public CvCapture
     bool convert_rgb;
     bool frame_allocated;
     bool returnFrame;
+    int channelNumber;
 
     /* V4L2 variables */
     buffer buffers[MAX_V4L_BUFFERS + 1];
-    v4l2_capability cap;
-    v4l2_input inp;
+    v4l2_capability capability;
+    v4l2_input videoInput;
     v4l2_format form;
     v4l2_crop crop;
     v4l2_cropcap cropcap;
@@ -299,9 +300,6 @@ struct CvCaptureCAM_V4L CV_FINAL : public CvCapture
 
     timeval timestamp;
 
-    /* V4L2 control variables */
-    Range focus, brightness, contrast, saturation, hue, gain, exposure;
-
     bool open(int _index);
     bool open(const char* deviceName);
 
@@ -309,31 +307,6 @@ struct CvCaptureCAM_V4L CV_FINAL : public CvCapture
     virtual bool setProperty(int, double) CV_OVERRIDE;
     virtual bool grabFrame() CV_OVERRIDE;
     virtual IplImage* retrieveFrame(int) CV_OVERRIDE;
-
-    Range getRange(int property_id) const {
-        switch (property_id) {
-        case CV_CAP_PROP_BRIGHTNESS:
-            return brightness;
-        case CV_CAP_PROP_CONTRAST:
-            return contrast;
-        case CV_CAP_PROP_SATURATION:
-            return saturation;
-        case CV_CAP_PROP_HUE:
-            return hue;
-        case CV_CAP_PROP_GAIN:
-            return gain;
-        case CV_CAP_PROP_EXPOSURE:
-            return exposure;
-        case CV_CAP_PROP_FOCUS:
-            return focus;
-        case CV_CAP_PROP_AUTOFOCUS:
-            return Range(0, 1);
-        case CV_CAP_PROP_AUTO_EXPOSURE:
-            return Range(0, 4);
-        default:
-            return Range(0, 255);
-        }
-    }
 
     virtual ~CvCaptureCAM_V4L();
 };
@@ -344,7 +317,7 @@ static bool icvGrabFrameCAM_V4L( CvCaptureCAM_V4L* capture );
 static IplImage* icvRetrieveFrameCAM_V4L( CvCaptureCAM_V4L* capture, int );
 
 static double icvGetPropertyCAM_V4L( const CvCaptureCAM_V4L* capture, int property_id );
-static int    icvSetPropertyCAM_V4L( CvCaptureCAM_V4L* capture, int property_id, double value );
+static int    icvSetPropertyCAM_V4L( CvCaptureCAM_V4L* capture, int property_id, int value );
 
 /***********************   Implementations  ***************************************/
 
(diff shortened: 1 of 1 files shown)


modules/videoio/src/cap_v4l.cpp
```diff
@@ -1880,7 +1830,11 @@ bool CvCaptureCAM_V4L::grabFrame()
 
 IplImage* CvCaptureCAM_V4L::retrieveFrame(int)
 {
-    return icvRetrieveFrameCAM_V4L( this, 0 );
+    IplImage *image = icvRetrieveFrameCAM_V4L(this, 0);
+    //Revert buffer to the queue
+    if (-1 == ioctl(deviceHandle, VIDIOC_QBUF, &buffers[bufferIndex].buffer))
+        perror("VIDIOC_QBUF");
```
alalek: Not sure that we really need this.

`.retrieve()` can be called several times (even with the same "index" parameter).

Next `.grab()` call should "return back" this buffer.

modules/videoio/src/cap_v4l.cpp
```diff
@@ -1533,48 +1415,157 @@ static IplImage* icvRetrieveFrameCAM_V4L( CvCaptureCAM_V4L* capture, int) {
...
+    }
+}
+
+static inline int capPropertyToV4L2(int prop)
+{
+    switch (prop) {
+    case CV_CAP_PROP_POS_MSEC:
+        return -1;
+    case CV_CAP_PROP_POS_FRAMES:
+        return -1;
```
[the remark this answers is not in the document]
aleksey-nikolaev: If it really obstruct, I remove it immediately.  I'm just not sure if there's no strict match for the parameters or something skipped, it's easier to find or add the skipped one in this form.

modules/videoio/src/cap_v4l.cpp
aleksey-nikolaev: We need to put buffer back to the queue as fast as we can. `Index` parameter is a channel `VideoCapture::retrieve(OutputArray image, int channel)` that has no sense for v4l capture, because changing channel lead to stop streaming. Ok we can call `.retrieve()` several times, I can fixed by send old `IplImage *image`, is it ok?

alalek: Also `.retrieve()` may not be called at all after `.grab()`.

aleksey-nikolaev: Done

modules/videoio/src/cap_v4l.cpp
```diff
@@ -273,8 +272,8 @@ struct CvCaptureCAM_V4L CV_FINAL : public CvCapture
     int bufferIndex;
     int FirstCapture;
     String deviceName;
+    mutable cv::Mutex mutex;
```
paroj: I do not think that locking should be introduced at a per-backend level. If desired it should be done at the VideoCapture toplevel.

modules/videoio/src/cap_v4l.cpp
```diff
@@ -285,71 +284,61 @@ struct CvCaptureCAM_V4L CV_FINAL : public CvCapture
...
-        case CV_CAP_PROP_AUTOFOCUS:
-            return Range(0, 1);
-        case CV_CAP_PROP_AUTO_EXPOSURE:
-            return Range(0, 4);
-        default:
-            return Range(0, 255);
-        }
-    }
-
+    CvCaptureCAM_V4L();
```
paroj: as you are rewriting this anyway, you can consider porting this to the new `IVideoCapture` API

paroj: unless you actually need this in 3.4 that is..

paroj: but then again you should leave parameter conversion as is - upgrading to 3.4.y should not needlessly break things

modules/videoio/src/cap_v4l.cpp
alalek: Agreed.
Extra synchronization should be avoided on backend level if it is not necessary (async callbacks).

modules/videoio/src/cap_v4l.cpp
aleksey-nikolaev: I remove parameter conversation because I can't set/get some of it in this case. I got INF for exposure. Look at stackoverflow - working with opencv-v4l is a pain. Most of parameters a missed. So, why you  want to leave  senseless parameter conversion when it so hard to use?
If you like, I rebase this to 4.0. Again, my goal wasn't rewrite v4l, I want to work with the camera in the same way on both linux and windows: the same parameters and the same synchronization.

modules/videoio/src/cap_v4l.cpp
aleksey-nikolaev: So, what is your purpose? Removing mutex/CriticalSection from all backend level and creating it at the VideoCapture level? Yes, this is good idea, but you don't have it now. For now, I want to get capture on the linux (v4l) same as on the windows (msmf).

paroj: you should not assume the backends to be threadsafe. Rather the msmf is threadsafe by coincidence as it needs the locks due to the internal async callbacks.
I would rather suggest introducing locks on application level so singlethreaded apps do not have to pay the penalty.

Locking on VideoCapture level should be a separate PR and probably would incur a separate discussion.

modules/videoio/src/cap_v4l.cpp
paroj: > So, why you want to leave senseless parameter conversion when it so hard to use?

backwards compatibility. There is working code that provides the parameters in the senseless format and it would break by this improvement.

Breaking on 4.0 should be fine though.

> Look at stackoverflow - working with opencv-v4l is a pain.

that is a little bit unspecific..

modules/videoio/src/cap_v4l.cpp
aleksey-nikolaev: "threadsafe by coincidence" - try to remove synchronization from cap_dshow, cap_msmf, cap_ffmpeg_impl (it have it's own wrapper :) ), cap_gstreamer, cap_openni2, cap_ffmpeg. Only for VideoCapture_WinRT you move mutex to VideoCapture. Why? Maybe the problem is that a backend synchronization is a backend specific? 

"you should not assume the backends to be threadsafe." Why? I configure and debug the capture on one platform and I have to do the same on the other, isn't it? Is it a good idea to force OpenCV users to write platform-specific wrappers for OpenCV objects?

"singlethreaded apps do not have to pay the penalty" I use cv::Mutex, if you say that it is slow than I could use something else or wrap it to `#ifdef USE_THREADSAFE_OPENCV_V4L`. Is the mutex really slower than converting to RGB or even memcopy of an image?

paroj: > try to remove synchronization from cap_dshow (..)

I was talking about VideoCapture level thread-safety there :confused: 

>  Is the mutex really slower than converting to RGB (..)

I agree that the overhead is likely neglectible. However I think if we make VideoCapture threadsafe, same reasoning also applies to the [Algorithm](https://docs.opencv.org/master/d3/d46/classcv_1_1Algorithm.html) classes.

alalek: Multi-threaded code is usually more complicated. So this is the code maintenance question at first. Also it is hard to test synchronization or reproduce issues.

If synchronization is not required to use library from single-threaded app (or can be successfully handled on caller level), then please avoid it here.

aleksey-nikolaev: I think, a thread-safety is a backend specific task. Let's holy war begin! :) Really, users don't care about synchronization destructor of backend object and getting parameter from a camera. Also, for some backends, getting parameters are thread-safe permanently and for other one you can't get anything while someone is already reading from it.

> same reasoning also applies to the Algorithm classes

Why not? Let user make a choice:  singlethread mode or thread-safe. For  singlethread you could make  cv::Mutex more simple. Or make cv::Mutex faster (depending on platform)

modules/videoio/src/cap_v4l.cpp
aleksey-nikolaev: ok, I will rebase it on 4.0 branch. Should I close this PR?

alalek: You can rebase this patch "inplace". No need to reopen PR.

modules/videoio/src/cap_v4l.cpp
```diff
@@ -437,6 +435,8 @@ static int autosetup_capture_mode_v4l2(CvCaptureCAM_V4L* capture) {
             V4L2_PIX_FMT_YUV411P,
             V4L2_PIX_FMT_YUYV,
             V4L2_PIX_FMT_UYVY,
+            V4L2_PIX_FMT_NV12,
+            V4L2_PIX_FMT_NV21,
```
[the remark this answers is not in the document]
aleksey-nikolaev: Because  icvRetrieveFrameCAM_V4L didn't have case for converting to rgb for this format

aleksey-nikolaev:
Unfortunately, I need this on 3.4. The main problem is a backward compatibility: [0,1] ranges for parameters. I could revert the ranges conversion and add the capture mode  or a build option for using real parameters instead of converted. Would that be an acceptable compromise?

aleksey-nikolaev:
@alalek , @paroj , I've already deleted the mutex and returned the parameter ranges, is that enough or am I missing something?

modules/videoio/src/cap_v4l.cpp
```diff
@@ -1585,312 +1533,361 @@ static inline __u32 capPropertyToV4L2(int prop) {
...
+    case CV_CAP_PROP_FRAME_WIDTH:
+        return form.fmt.pix.width;
+    case CV_CAP_PROP_FRAME_HEIGHT:
+        return form.fmt.pix.height;
+    case CV_CAP_PROP_FOURCC:
+        return palette;
+    case CV_CAP_PROP_FORMAT:
+        return CV_MAKETYPE(IPL2CV_DEPTH(frame.depth), frame.nChannels);
+    case CV_CAP_PROP_MODE:
+        return channelNumber;
```
alalek: There is some mess with this property (and there was mess before).

For example
- cap_libv4l:
    ```
    - Allow getting/setting CV_CAP_PROP_MODE. These values are supported:
        - CV_CAP_MODE_BGR  : BGR24 (default)
        - CV_CAP_MODE_RGB  : RGB24
        - CV_CAP_MODE_GRAY : Y8, extracted from YUV420
    ```
- DC1394: format presets via `dc1394_video_set_mode()`
- v4l (**patch must be backward compatible**) before this patch: "capture->palette" (like CV_CAP_PROP_FOURCC);
- MSMF: H/W acceleration (???)


There is similar property "for channels" in master branch: `CAP_CROSSBAR_INPIN_TYPE` (with wrong name!)

Lets introduce dedicated property `CAP_PROP_CHANNEL = 43` for that (I will replace `CAP_CROSSBAR_INPIN_TYPE` later).

aleksey-nikolaev: @alalek , I am sorry. I can't understand what do you want from me. Except CV_CAP_PROP_MODE, all "old" parameters used as always and as you says "there was mess before" but - it is **backward compatible** .
cap_libv4l and cap_v4l was not compatible, what should I do with CV_CAP_MODE_GRAY on cap_v4l? MSMF has same CV_CAP_MODE_* as cap_libv4l, but it is not documented. It is very unusual using device capture format as CV_CAP_PROP_MODE. I have device that send me YUYV or MJPEG. To get YUYV frame on windows (msmf) I should set CV_CAP_PROP_MODE to 3 and CV_CAP_PROP_CONVERT_RGB to 1 (WTF!!!), on linux (v4l) I just set CV_CAP_PROP_FOURCC to YUYV (nice!). Why libv4l and MSMF  used CV_CAP_PROP_MODE this way? CV_CAP_PROP_FOURCC - must set device output format.

I used CV_CAP_PROP_MODE for something device/backend special because it was not used for setting a value (v4l).

CAP_PROP_CHANNEL = 43 - one more than Answer to the Ultimate Question of Life, the Universe, and Everything. Wonderful! Should I insert it myself or it was an other PR?

As for me, CV_CAP_PROP_FORMAT (as docs says) is output format of result image, so if you want to get YUV instead of BGR (awful) just set CV_CAP_PROP_FORMAT. For now I configure capture for send me the frame without any transformation and convert it myself

alalek: Don't touch other backends. Lets process this one.

Actually there are two "processing" formats:
backend SDK -> (1) -> OpenCV -> (2) -> Application

Related properties:
- `CAP_PROP_FOURCC` usually works with (1). V4L stores this value in `capture->palette`
- `CAP_PROP_FORMAT` works with (2) "Format of the %Mat objects returned by VideoCapture::retrieve()"
- `CV_CAP_PROP_CONVERT_RGB` is early version of `CAP_PROP_FORMAT`. Some backends support only one property and don't support another, some of them support both.

(1) is capture/camera "native" format. Like MJPEG from here:
```
$ v4l2-ctl --list-formats
ioctl: VIDIOC_ENUM_FMT
	Index       : 0
	Type        : Video Capture
	Pixel Format: 'YUYV'
	Name        : YUYV 4:2:2

	Index       : 1
	Type        : Video Capture
	Pixel Format: 'MJPG' (compressed)
	Name        : Motion-JPEG
```
`cv::VideoCapture` is able to take this MJPEG stream and then converts it into BGR frames (also it is able to pass it "as is" via `CV_8UC1` buffer).

> cap_libv4l and cap_v4l was not compatible

`cap_libv4l` is going to be deprecated.


> CAP_PROP_CHANNEL = 43

Please add this into this PR and use this property to specify "channelNumber" (old `#define CHANNEL_NUMBER 1`)

---

Please drop `CV_` prefix (it is from legacy C API):
```.diff
-CV_CAP_PROP_...
+CAP_PROP_...
```

paroj: @alalek I think we really should not care about `CV_CAP_PROP_MODE`

as I described at lengths in #5485, it is just a bad (as in ambiguous) API.

alalek: @paroj I agree. So lets try avoid using of this ambiguous property.

aleksey-nikolaev: > * `CAP_PROP_FORMAT` works with (2) "Format of the %Mat objects returned by VideoCapture::retrieve()"

Are you sure you want to use the property as described? Because it's useless! I need the output format (after read/retrieve image), and what I get is the number of channels and the size of the component. And that it can be: BGR, RGB, YUV, Lab, HSV, HSL... or not?

Never mind, do as you wish.
I used `CAP_PROP_MODE` for setting backward compatibility mode (by default).
Is that enough to complete my PR?

modules/videoio/src/cap_v4l.cpp
```diff
@@ -285,149 +283,161 @@ struct CvCaptureCAM_V4L CV_FINAL : public CvCapture
     bool convert_rgb;
     bool frame_allocated;
     bool returnFrame;
+    // To select a video input set cv::CAP_PROP_MODE to channel number.
+    // If the new channel number is greater than 32768 or less than 0, then a video input will not change
+    int channelNumber;
```
alalek: CAP_PROP_MODE -> CAP_PROP_CHANNEL

modules/videoio/src/cap_v4l.cpp
```diff
@@ -1451,446 +1333,560 @@ static int sonix_decompress(int width, int height, unsigned char *inp, unsigned
...
-        }
-
-        switch (property_id) {
-        case CV_CAP_PROP_FRAME_WIDTH:
-            return form.fmt.pix.width;
-        case CV_CAP_PROP_FRAME_HEIGHT:
-            return form.fmt.pix.height;
-        case CV_CAP_PROP_FOURCC:
-        case CV_CAP_PROP_MODE:
-            return capture->palette;
```
alalek: Compatibility should be preserved for CV_CAP_PROP_MODE property.

modules/videoio/src/cap_v4l.cpp
```diff
@@ -285,149 +283,161 @@ struct CvCaptureCAM_V4L CV_FINAL : public CvCapture
...
     bool returnFrame;
+    // To select a video input set cv::CAP_PROP_MODE to channel number.
+    // If the new channel number is greater than 32768 or less than 0, then a video input will not change
+    int channelNumber;
+    // Backward compatibility for parameters. If set parameters will be converted to/from [0,1) range.
+    // To select real parameters mode set cv::CAP_PROP_MODE to 65536
+    // any other value greater than 32768 revert backward compatibility mode.
+    // Backward compatibility mode affects the following parameters:
+    // cv::CAP_PROP_*: BRIGHTNESS,CONTRAST,SATURATION,HUE,GAIN,EXPOSURE,FOCUS,AUTOFOCUS,AUTO_EXPOSURE.
+    bool compatibilityMode;
```
alalek: Comment is outdated.
`CAP_PROP_MODE` is already used, so behavior should not be changed.

Alternative variant via environment variables:
```
compatibilityMode = !utils::getConfigurationParameterBool("OPENCV_VIDEOIO_V4L_RANGE_NORMALIZED", false);
```

modules/videoio/src/cap_v4l.cpp
aleksey-nikolaev: +1

modules/videoio/src/cap_v4l.cpp
aleksey-nikolaev: +1

modules/videoio/src/cap_v4l.cpp
aleksey-nikolaev: `CAP_PROP_MODE` was used for get - not for set

modules/videoio/src/cap_v4l.cpp
```diff
@@ -769,7 +770,7 @@ bool CvCaptureCAM_V4L::open(const char* _deviceName)
     frame_allocated = false;
     deviceName = _deviceName;
     returnFrame = true;
-    compatibilityMode = utils::getConfigurationParameterBool("OPENCV_VIDEOIO_V4L_RANGE_NORMALIZED", true);
+    compatibilityMode = !utils::getConfigurationParameterBool("OPENCV_VIDEOIO_V4L_RANGE_NORMALIZED", false);
```
paroj: this was correct before: 
```
compatibilityMode <=>  "converted to/from [0,1) range" <=> OPENCV_VIDEOIO_V4L_RANGE_NORMALIZED <=> 0/1
```

maybe either rename either `compatibilityMode` or `OPENCV_VIDEOIO_V4L_RANGE_NORMALIZED` to match each other.

modules/videoio/src/cap_v4l.cpp
```diff
@@ -261,7 +261,12 @@ struct buffer
     size_t  length;
     // This is dequeued buffer. It used for to put it back in the queue.
     // The buffer is valid only if capture->bufferIndex >= 0
-    v4l2_buffer buffer;
+    v4l2_buffer v4l2_buf;
+
+    buffer() : start(NULL), length(0)
+    {
+        memset(&v4l2_buf, 0, sizeof(v4l2_buf));
```
aleksey-nikolaev: `v4l2_buf = v4l2_buffer()` this initialization style is used in this cpp file. Please, don't mix it.

aleksey-nikolaev: If you care about variable names, I could bring everything in this file to a single style, including renaming ambiguous entities

modules/videoio/src/cap_v4l.cpp
alalek: Thanks!
Renamed to `normalizePropRange`.

modules/videoio/src/cap_v4l.cpp
alalek: Renamed type instead.
Using `v4l2_buffer()` initializer.

alalek:
@paroj @aleksey-nikolaev I believe it should be fine now. Any comments?

aleksey-nikolaev:
code style maybe. And what about the comments at the top of the file, they're no longer relevant, are they?

washley:
I think this broke builds on kernels < 3.5 (at least that, maybe an even higher version is required). See https://github.com/opencv/opencv/issues/13335

modules/videoio/src/cap_v4l.cpp
```diff
@@ -1451,446 +1341,562 @@ static int sonix_decompress(int width, int height, unsigned char *inp, unsigned
...
-                capture->form.fmt.pix.height,
-                (unsigned char*)(capture->buffers[capture->bufferIndex].start),
-                (unsigned char*)capture->frame.imageData);
-        break;
+        cv::cvtColor(cv::Mat(imageSize.height * 3 / 2, imageSize.width, CV_8U, currentBuffer.start), destination,
+                     COLOR_YUV2BGR_IYUV);
+        return;
+    case V4L2_PIX_FMT_NV12:
+        cv::cvtColor(cv::Mat(imageSize.height * 3 / 2, imageSize.width, CV_8U, currentBuffer.start), destination,
+                     COLOR_YUV2RGB_NV12);
```
opencv-alalek, line 1392: > COLOR_YUV2RGB_NV12

Bug is introduced here. Should be BGR by default.

Found here: #26025 (to be backported to 4.x)

PR #12894, opened 2018-10-22
drop C-API sample code

alalek:


berak:
i found one more in https://github.com/opencv/opencv/blob/master/doc/tutorials/core/mat_operations.markdown#L202-206  /  https://github.com/opencv/opencv/blob/9c23f2f1a682faa9f0b2c2223a857c7d93ba65a6/samples/cpp/tutorial_code/core/mat_operations/mat_operations.cpp#L131-L139

maybe it can be added ?

PR #12896, opened 2018-10-23
cmake: OPENCV_GENERATE_PKGCONFIG

alalek:
Disable pkg-config file generation by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -340,7 +340,7 @@ OCV_OPTION(ENABLE_BUILD_HARDENING     "Enable hardening of the resulting binarie
 OCV_OPTION(ENABLE_LTO                 "Enable Link Time Optimization" OFF IF CV_GCC OR MSVC)
 OCV_OPTION(ENABLE_THIN_LTO            "Enable Thin LTO" OFF IF CV_CLANG)
 OCV_OPTION(GENERATE_ABI_DESCRIPTOR    "Generate XML file for abi_compliance_checker tool" OFF IF UNIX)
-OCV_OPTION(OPENCV_GENERATE_PKGCONFIG  "Generate .pc file for pkg-config build tool (deprecated)" ON IF (UNIX AND NOT MSVC AND NOT IOS AND NOT ANDROID) )
+OCV_OPTION(OPENCV_GENERATE_PKGCONFIG  "Generate .pc file for pkg-config build tool (deprecated)" OFF)
 OCV_OPTION(CV_ENABLE_INTRINSICS       "Use intrinsic-based optimized code" ON )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -337,6 +337,7 @@ OCV_OPTION(ENABLE_BUILD_HARDENING     "Enable hardening of the resulting binarie
 OCV_OPTION(ENABLE_LTO                 "Enable Link Time Optimization" OFF IF CV_GCC OR MSVC)
 OCV_OPTION(ENABLE_THIN_LTO            "Enable Thin LTO" OFF IF CV_CLANG)
 OCV_OPTION(GENERATE_ABI_DESCRIPTOR    "Generate XML file for abi_compliance_checker tool" OFF IF UNIX)
+OCV_OPTION(OPENCV_GENERATE_PKGCONFIG  "Generate .pc file for pkg-config build tool (deprecated)" OFF)
```
mshabunin: `GENERATE_PKGCONFIG`?

CMakeLists.txt
```diff
@@ -846,6 +847,7 @@ include(cmake/OpenCVGenHeaders.cmake)
 
 # Generate opencv.pc for pkg-config command
 if(NOT OPENCV_SKIP_PKGCONFIG_GENERATION
```
mshabunin: Can we remove `OPENCV_SKIP_PKGCONFIG_GENERATION`?

CMakeLists.txt
alalek: To avoid possible conflicts, I prefer to have `OPENCV_` prefix.
Other prefixes like `BUILD_` or `ENABLE_` are generic too.

CMakeLists.txt
alalek: It is still here for compatibility reasons.
However `OPENCV_GENERATE_PKGCONFIG` is disabled by default in master, so use case with enabling both options looks strange.

Perhaps we should maintain somehow list of deprecated build options, warn about them, and drop in the future.

alalek:
:+1:

sergiomb2:
Don't say that pkgconfig is deprecated please, all Fedora rpms packages can be driven by pkgconfig 
I got error just because package use [1]

[1] 
BuildRequires: pkgconfig(opencv)

mshabunin:
@sergiomb2 , it still can be generated and used for non-static builds. The mechanism has not been removed and fixes will be accepted. But it is very hard to generate correct pkg-config file for static builds with many dependencies.

sergiomb2:
> Don't say that pkgconfig is deprecated please, all Fedora rpms packages can be driven by pkgconfig
> I got error just because package use [1]
> 
> [1]
> BuildRequires: pkgconfig(opencv)

Correction in Fedora we also may use [1] 

[1] 
BuildRequires: cmake(opencv)

PR #12908, opened 2018-10-24
resolve #11855

### LINKED ISSUE #11855: BFmatcher with crossCheck=True doesn't cross check
- OpenCV => 3.4.0
- Operating System / Platform => Linux Ubuntu 16.04 X86 64 Bit
- Compiler => gcc 5.4

##### Detailed description

Running BFmatcher with crossCheck enabled doesn't really cross check. According to the documentation (https://docs.opencv.org/trunk/d3/da1/classcv_1_1BFMatcher.html#ac6418c6f87e0e12a88979ea57980c020)

> If crossCheck==true, then the knnMatch() method with k=1 will only return pairs (i,j) such that for i-th query descriptor the j-th descriptor in the matcher's collection is the nearest and vice versa, i.e. the BFMatcher will only return consistent pairs. 

I read this as, if *i* is closest to *j*, then *j* is also closest to *i*. I.e. this is symmetrical. However, this does not seem to be the case with the implemented algorithm. Also, if we switch the collections, i.e. queryDescriptors and trainDescriptors, we get different results, which is unexpected.

##### Steps to reproduce
To illustrate the first part of the issue, try the following
``` python
import numpy
import cv2

sources = numpy.array([[1, 1, 0], [1, 1, 1]], dtype=numpy.uint8)
targets = numpy.array([[1, 1, 1], [0, 0, 0]], dtype=numpy.uint8)
for si, source in enumerate(sources):
    for ti, target in enumerate(targets):
        print('%d -> %d: %d' % (si, ti, cv2.norm(source, target, cv2.NORM_HAMMING)))

bf = cv2.BFMatcher(cv2.NORM_HAMMING, crossCheck=True)
matches = bf.knnMatch(sources, targets, k=1)
print('-- crosscheck=True')
for match in matches:
    if match:
        print('%d -> %d: %f' % (match[0].queryIdx, match[0].trainIdx, match[0].distance))

bf_nocc = cv2.BFMatcher(cv2.NORM_HAMMING, crossCheck=False)
matches_nocc = bf.knnMatch(sources, targets, k=1)
print('-- crosscheck=False')
for match in matches_nocc:
    if match:
        print('%d -> %d: %f' % (match[0].queryIdx, match[0].trainIdx, match[0].distance))
```
This outputs
```
0 -> 0: 1
0 -> 1: 2
1 -> 0: 0
1 -> 1: 3
-- crossCheck=True
0 -> 1: 2.000000 <= Unexpected
1 -> 0: 0.000000
-- crossCheck=False
0 -> 0: 1.000000
1 -> 0: 0.000000
```
Here, the 0 -> 1 match when crossCheck=True is unexpected since source 0 is closer to target 0.

Also, if we switch the sources and targets arrays, we get
```
0 -> 0: 1
0 -> 1: 0
1 -> 0: 2
1 -> 1: 3
-- crossCheck=True
0 -> 1: 0.000000
-- crossCheck=False
0 -> 1: 0.000000
1 -> 0: 2.000000
```
Which is correct for crossCheck=True (as well as for crossCheck=False which I don't expect to be symmetrical).

Unless I have misunderstood how cross check is supposed to work, this appears to me to be an issue with the current implementation (there has been no updates on this algorithm since 3.4.0 as far as I can see)

For C++ sample code, see http://answers.opencv.org/question/194661/bfmatcher-with-crosscheck-doesnt-crosscheck/

alexevans:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/543

resolves #11855 

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/core/src/batch_distance.cpp
+++ b/modules/core/src/batch_distance.cpp
@@ -297,19 +297,21 @@ void cv::batchDistance( InputArray _src1, InputArray _src2,
         nidx = Scalar::all(-1);
     }
 
+
     if( crosscheck )
     {
         CV_Assert( K == 1 && update == 0 && mask.empty() );
         CV_Assert(!nidx.empty());
-        Mat tdist, tidx;
+        Mat tdist, tidx, sdist, sidx;
         batchDistance(src2, src1, tdist, dtype, tidx, normType, K, mask, 0, false);
+        batchDistance(src1, src2, sdist, dtype, sidx, normType, K, mask, 0, false);
 
         // if an idx-th element from src1 appeared to be the nearest to i-th element of src2,
         // we update the minimum mutual distance between idx-th element of src1 and the whole src2 set.
         // As a result, if nidx[idx] = i*, it means that idx-th element of src1 is the nearest
         // to i*-th element of src2 and i*-th element of src2 is the closest to idx-th element of src1.
         // If nidx[idx] = -1, it means that there is no such ideal couple for it in src2.
-        // This O(N) procedure is called cross-check and it helps to eliminate some false matches.
+        // This O(2N) procedure is called cross-check and it helps to eliminate some false matches.
         if( dtype == CV_32S )
         {
             for( int i = 0; i < tdist.rows; i++ )
@@ -334,6 +336,13 @@ void cv::batchDistance( InputArray _src1, InputArray _src2,
                     dist.at<float>(idx) = d;
                     nidx.at<int>(idx) = i + update;
                 }
+            }
+        }
+        for( int i = 0; i < sdist.rows; i++ )
+        {
+            if( tidx.at<int>(sidx.at<int>(i)) != i )
+            {
+                nidx.at<int>(i) = -1;
             }
         }
         return;
--- a/modules/features2d/test/test_matchers_algorithmic.cpp
+++ b/modules/features2d/test/test_matchers_algorithmic.cpp
@@ -558,4 +558,23 @@ TEST( Features2d_DMatch, read_write )
     ASSERT_NE( strstr(str.c_str(), "4.5"), (char*)0 );
 }
 
+
+TEST(Features2d_DMatch, issue_11855)
+{
+    Mat sources = (Mat_<uchar>(2, 3) << 1, 1, 0,
+                                        1, 1, 1);
+    Mat targets = (Mat_<uchar>(2, 3) << 1, 1, 1,
+                                        0, 0, 0);
+
+    Ptr<BFMatcher> bf = BFMatcher::create(NORM_HAMMING, true);
+    vector<vector<DMatch> > match;
+    bf->knnMatch(sources, targets, match, 1, noArray(), true);
+
+    ASSERT_EQ((size_t)1, match.size());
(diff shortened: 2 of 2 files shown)


alalek:
Please add test for this patch (code from answers doesn't expect empty entries).

alexevans:
@alalek 
Here's what I used, based on the tests in the original issue and answers post.

```cpp
#include <iostream>
#include <opencv2/core.hpp>
#include <opencv2/features2d.hpp>

using namespace cv;
using namespace std;

int main()
{
Mat targets = (Mat_<uchar>(2, 3) <<  1, 1, 0, 1, 1, 1);
Mat sources = (Mat_<uchar>(2, 3) << 1, 1, 1, 0,0,0);
cout << sources << "\n";
cout << targets << "\n";
cout << "SOURCES TARGETS\n";
    for (int i = 0; i < sources.rows; i++)
        for (int j = 0; j < targets.rows; j++)
            cout << i << " -> " << j << " = " << norm(sources.row(i), targets.row(j), NORM_HAMMING) << "\n";
    cout << "TARGETS SOURCES\n";
    for (int i = 0; i < targets.rows; i++)
        for (int j = 0; j < sources.rows; j++)
            cout << i << " -> " << j << " = " << norm(targets.row(i), sources.row(j), NORM_HAMMING) << "\n";

Ptr<BFMatcher> bf = BFMatcher::create(NORM_HAMMING, true);
vector<vector<DMatch>> match;
cout << "Match with crosscheck TRUE\n";
bf->knnMatch(sources, targets, match, 1);

for (auto m : match)
    if(m.size() > 0)
        cout << m[0].queryIdx << " -> " << m[0].trainIdx << " = " << m[0].distance << "\n";

Ptr<BFMatcher> bf2 = BFMatcher::create(NORM_HAMMING, false);
cout << "Match with crosscheck FALSE \n";
match.clear();
bf2->knnMatch(sources, targets, match, 1);

for (auto m : match)
    cout << m[0].queryIdx << " -> " << m[0].trainIdx << " = " << m[0].distance << "\n";
return 0;
}
```

```python
import numpy
import cv2

targets = numpy.array([[1, 1, 0], [1, 1, 1]], dtype=numpy.uint8)
sources = numpy.array([[1, 1, 1], [0, 0, 0]], dtype=numpy.uint8)
for si, source in enumerate(sources):
    for ti, target in enumerate(targets):
        print('%d -> %d: %d' % (si, ti, cv2.norm(source, target, cv2.NORM_HAMMING)))

bf = cv2.BFMatcher(cv2.NORM_HAMMING, crossCheck=True)
matches = bf.knnMatch(sources, targets, k=1)
print('-- crosscheck=True')
for match in matches:
    if match:
        print('%d -> %d: %f' % (match[0].queryIdx, match[0].trainIdx, match[0].distance))

bf_nocc = cv2.BFMatcher(cv2.NORM_HAMMING, crossCheck=False)
matches_nocc = bf_nocc.knnMatch(sources, targets, k=1)
print('-- crosscheck=False')
for match in matches_nocc:
    if match:
        print('%d -> %d: %f' % (match[0].queryIdx, match[0].trainIdx, match[0].distance))
```

PR #12916, opened 2018-10-24
fix matchTemplate for imgproc module

lqy123000:
### This pullrequest changes

Avoid rounding errors when using CV_8U image in matchTemplate.

The result of TM_CCOEFF_NORMED is 0/0 when all pixels have the same value in matched area. However, "wndSum2 - wndMean2 == 0" may become "wndSum2 - wndMean2 > 0" because of rounding errors. If "fabs(num) < t" is true at the same time, we may get a unexpected value. Therefore, I give a proper threshold for the denominator.

**Here is my test code:**

    Mat aim = imread("../images/aim2.png", IMREAD_GRAYSCALE);
    Mat templ = imread("../images/templ0.png", IMREAD_GRAYSCALE);
    Mat result;
    matchTemplate(aim, templ, result, TemplateMatchModes::TM_CCOEFF_NORMED);
    double minVal = 0.0f;
    double maxVal = 0.0f;
    Point minLoc;
    Point maxLoc;
    minMaxLoc(result, &minVal, &maxVal, &minLoc, &maxLoc);
    cout << maxLoc << endl;

A wrong result comes with the previous code. And right result is achieved with the modification.

aim2
aim2.png

templ0
templ0.png

result_wrong
wrong result

result_right
right result

### DIFF
--- a/modules/imgproc/src/templmatch.cpp
+++ b/modules/imgproc/src/templmatch.cpp
@@ -947,10 +947,11 @@ static void common_matchTemplate( Mat& img, Mat& templ, Mat& result, int method,
 
             if( isNormed )
             {
-                if (img.depth() == CV_8U && (wndSum2 - wndMean2) < invArea*0.5)
+                double diff2 = MAX(wndSum2 - wndMean2, 0);
+                if (diff2 <= std::min(0.5, 10 * FLT_EPSILON * wndSum2))
                     t = 0; // avoid rounding errors
                 else
-                    t = std::sqrt(MAX(wndSum2 - wndMean2, 0))*templNorm;
+                    t = std::sqrt(diff2)*templNorm;
 
                 if( fabs(num) < t )
                     num /= t;
--- a/modules/imgproc/src/templmatch.cpp
+++ b/modules/imgproc/src/templmatch.cpp
@@ -947,7 +947,11 @@ static void common_matchTemplate( Mat& img, Mat& templ, Mat& result, int method,
 
             if( isNormed )
             {
-                t = std::sqrt(MAX(wndSum2 - wndMean2,0))*templNorm;
+                if (img.depth() == CV_8U && (wndSum2 - wndMean2) < invArea*0.5)
+                    t = 0; // avoid rounding errors
+                else
+                    t = std::sqrt(MAX(wndSum2 - wndMean2, 0))*templNorm;
+
                 if( fabs(num) < t )
                     num /= t;
                 else if( fabs(num) < t*1.125 )


alalek:
Thank you for the report with reproducer.

1. `img.depth() == CV_8U` check is not necessary here. Problem is reproducible with other data types too (CV_32F).
2. threshold limit looks strange. Replaced.

Please check updated patch with your dataset.

lqy123000:
I forgot to give derivation of the threshold.
1. I agree that we should consider other types here, but I can not find a proper method to avoid rounding errors with CV_32F. So I only consider CV_8U since it is often used in image processing. Maybe someone else can give a universal method than a simple threshold.
2. Here is the reason for this "strange" threshold limit. When the depth of image is CV_8U, all pixels have integer value. So `sum` and `sqsum` are integers (integer in math not program). Define `wndMean2` as `raw_wndMean2` before it is multiplied by `invArea`. Then, we can deduce that `raw_wndMean2` is a integer and `wndSum2` is integer too. 
`wndSum2 - wndMean2 `
`= wndSum2 - raw_wndMean2 * invArea`
`= wndSum2 - raw_wndMean2 / (templ.rows * templ.cols)`
`= (wndSum2 * (templ.rows * templ.cols) - raw_wndMean2) / (templ.rows * templ.cols)`
`= (wndSum2 * (templ.rows * templ.cols) - raw_wndMean2) * invArea`
Note that multiplicand is integer and `wndSum2 - wndMean2 >= 0` (from the definition of standard deviation). so we can treat `(wndSum2 - wndMean2)` as `0` if `wndSum2 - wndMean2 < 1 * invArea`. Take account of rounding errors, I choose the balance threshold `0.5 * invArea`.
3. I can not find the mathematical meaning behind your condition. I do except a solution supporting CV_32F type, but it should have mathematical meaning. I construct a counterexample that your condition do not work. Here is the counterexample:
```
    Mat aim(120, 120, CV_8UC1, Scalar(255));
    aim.at<uchar>(0, 0) = 0;
    Mat templ(150, 150, CV_8UC1, Scalar(255));
    templ.at<uchar>(20, 20) = 0;
    Mat result;
    matchTemplate(aim, templ, result, TemplateMatchModes::TM_CCOEFF_NORMED);
    double minVal = 0.0f;
    double maxVal = 0.0f;
    Point minLoc;
    Point maxLoc;
    minMaxLoc(result, &minVal, &maxVal, &minLoc, &maxLoc);
    cout << maxLoc << endl;
```
The output with your condition is `[0, 0]`, but the true output is `[20, 20]`.

alalek:
3. Meaning of suggested check is "stddev(image window) <= mean(image window) * 1e-2" (squared).

BTW, https://en.wikipedia.org/wiki/Dimensional_analysis

lqy123000:
> 1. Meaning of suggested check is "stddev(image window) <= mean(image window) * 1e-2" (squared).
> 
> BTW, https://en.wikipedia.org/wiki/Dimensional_analysis

OK, I understand your condition now. But I still have some questions:
1. Consider TM_CCOEFF_NORMED (your condition seems not work on TM_CCORR_NORMED and TM_SQDIFF_NORMED). A important feature of TM_CCOEFF_NORMED is the result remain unchanged when we shift aim image or invert template and aim image at the same time, i.e. "remove" the influence of "mean". Using "mean" in condition will break it. Change code in the example I constructed from
`matchTemplate(aim, templ, result, TemplateMatchModes::TM_CCOEFF_NORMED);` 
to
`matchTemplate(~aim, ~templ, result, TemplateMatchModes::TM_CCOEFF_NORMED);`
the output changed from `[0, 0]` to `[20, 20]`. In fact, it is the same when using original code. 
2. How do you explain the example I constructed in last reply? It works well in original code, but fail when adding your condition. We can not make another problem when solve one.

lqy123000:
I find the threshold I used for CV_8U is a bit strict. A more appropriate threshold is `wndSum2 - wndMean2 < 0.5` which can be proved strictly. 
Your condition maybe more appropriate for CV_32F since it keeps scale invariant. But the pixel value of CV_8U is integer which can not be scaled too small. So translation invariance is more important for CV_8U. And use "mean" in condition will break translation invariance just like I described in last reply.

lqy123000:
I think this condition is all right.

PR #12935, opened 2018-10-25
python: eliminate DeprecationWarning: SO -> EXT_SUFFIX

alalek:
- https://python.readthedocs.io/en/stable/whatsnew/3.4.html
- The `sysconfig` key `SO` is deprecated, it has been replaced by `EXT_SUFFIX`

Message from CMake step (with Python 3.7):
```
DeprecationWarning: SO is deprecated, use EXT_SUFFIX
```

```
buildworker:Mac OpenCL=macosx-1
```


alalek:
:+1:

PR #12944, opened 2018-10-25
added threshold to g-api imgproc tests

dbudniko:
Accuracy thresholds and tolerance setup added in G-API module tests. 

```
buildworker:Mac=macosx-2
```


dmatveev:
@dbudniko it seems build failed because of recently merged #12947 (and Sobel/sobel was that merge conflict)

modules/gapi/perf/common/gapi_core_perf_tests.cpp
```diff
@@ -0,0 +1 @@
+#include "gapi_core_perf_tests_inl.hpp"
```
dmatveev: Please also put a license header in this file and include "test_precomp.hpp" (before any other includes)

modules/gapi/perf/common/gapi_core_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1777 @@
+#ifndef OPENCV_GAPI_CORE_PERF_TESTS_INL_HPP
```
dmatveev: Please put license header here as well

modules/gapi/perf/common/gapi_core_perf_tests.cpp
dbudniko: perf_precomp.hpp added

modules/gapi/perf/common/gapi_core_perf_tests_inl.hpp
dbudniko: Done

dmatveev:
@alalek thanks!
@dbudniko congrats!

PR #12948, opened 2018-10-25
Add drawFrameAxes function

catree:
### This pullrequest changes

Similar to `aruco::drawAxis`, it allows displaying the object frame from a pose estimation method.

image

Note that since `undistortPoints` has been moved into the calib3d module, `#include <opencv2/imgproc.hpp>` can be removed in `pose_from_homography.cpp` in master.

### DIFF
--- a/modules/calib3d/include/opencv2/calib3d.hpp
+++ b/modules/calib3d/include/opencv2/calib3d.hpp
@@ -859,6 +859,26 @@ found, or as colored corners connected with lines if the board was found.
 CV_EXPORTS_W void drawChessboardCorners( InputOutputArray image, Size patternSize,
                                          InputArray corners, bool patternWasFound );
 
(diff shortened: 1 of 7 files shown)


alalek:
Well done! Thank you :+1:

Some thoughts (no need to fix anything) are below:

IMHO, It make sense to extract such functions into separate header files (for example, `<module>/utility.hpp`):
- debug/development purpose
- sometimes functions need to be improved to handle User-specific tasks (change colors, add negative "dark" lines, etc)
- non-common, not "well" defined (like `cv::line()`), no compatibility is guaranteed
- functions are usually not covered by tests (due not "well" defined, also it is hard to break them unintentionally by changes in other parts)
- nothing else in modules sources depends on this (tests/samples are not "module sources")

Separate header approach:
- allow to adjust code of these helpers on User/Application side
- most part of debug code is not used, so it is not necessary to put it into pre-compiled binaries.

PR #12958, opened 2018-10-26
docs: fix document about camera distortion parameters

kamino410:
### This pullrequest changes

[The current documentation](https://docs.opencv.org/3.4/d9/d0c/group__calib3d.html) says that positive k1 causes barrel distortion and negative k1 causes pincushion distortion.
But it seems the opposite.

According to the formula, distorted point (x'', y'') moves toward the outside of the image when k1 > 0.
So it should cause not barrel distortion but pincushion distortion.


catree:
> According to the formula, distorted point (x'', y'') moves toward the outside of the image when k1 > 0.

I agree with that.

> So it should cause not barrel distortion but pincushion distortion.

Barrel and pincushion distortion (from [Matlab documentation](https://www.mathworks.com/help/vision/ug/camera-calibration.html)):

image

kamino410:
That's true.
But [Edmund Optics](https://www.edmundoptics.jp/resources/application-notes/imaging/distortion/) defines them as follows.
fig-2-d

At least, I'm sure that cameras whose k1 is negative take figures having barrel distortion because
* I tested `leftXX.jpg`s in [https://github.com/opencv/opencv/tree/master/samples/data](https://github.com/opencv/opencv/tree/master/samples/data) and got `-0.27437524 | -0.03096935 | 0.00089529 | -0.00022190 | 0.27721061` as distortion params.
* `undistort()` returns an image having pincushion distortion with k1 < 0.

paroj:
generated with k1=2, rest zero
cdist

catree:
@paroj Thanks for taking the time to check this.

PR #12962, opened 2018-10-26
videoio: add manual white-balance and implement it for V4L2

paroj:


alalek:
BTW, There is another ongoing large patch with V4L changes: #12893 .

PR #12964, opened 2018-10-26
AVX v256_cleanup() fixed

savuor:
### This pullrequest changes

`_mm256_zeroupper` is replaced by `_mm256_zeroall`

This should fix an incorrect behavior of `vx_cleanup` on AVX+ baselines on some compilers (for example GCC 7.3.0, Ubuntu 18.04).

### Details

`_mm256_zeroupper` unpredictably spoils local variables even when there's no other intrinsics used in the code.
`_mm256_zeroall` doesn't cause such errors at the same time having similar purpose and performance.


vpisarev:
@alalek, `vx_cleanup()` should be much cheaper (maybe even much cheaper) than CV_INSTRUMENT_REGION() call. So, I suggest to use `vx_cleanup()` explicitly where it's needed. Of course, CV_INSTRUMENT_REGION() may be changed as well, to play on the safe side. :+1:

PR #12977, opened 2018-10-28
python: bindings loader package

alalek:
Configures and loads OpenCV bindings extension including 3rdparty dependencies

Based on running Python specify:
- configure PYTHONPATH (via "sys.path")
- configure PATH / LD_LIBRARY_PATH (changes doesn't work in the same process)

TODO:
- [ ] RPATH support from standalone Python builds (via CMAKE_INSTALL_RPATH ?)

```
test_modules=python2,python3
```


alalek:
:+1:

hmaarrfk:
@alalek it seems that this is a new file that changes how we are supposed to import and/or install opencv for use with python. 

Do you have some docs that explain the new install flags/macros?

xref: https://github.com/conda-forge/opencv-feedstock/pull/132

alalek:
@hmaarrfk Yes, default behavior has been changed.

Old behavior can restored via these CMake variables:
- `OPENCV_SKIP_PYTHON_LOADER=ON`
- `OPENCV_PYTHON3_INSTALL_PATH=...site/dist-packages location...`

hmaarrfk:
Thanks that did the trick!

PR #12984, opened 2018-10-29
Allow load CascadeClassifier from buffer when it fail to open FileStorage from filename

campos537:
### This pullrequest changes

The goal of this pull request is to allow load the CascadeClassifier from a buffer instead of the filename when it try to open the FileStorage and fail


alalek:
Use `.read(const FileNode& node)` instead.

PR #12989, opened 2018-10-30
Fix reading of black-and-white (thresholded) TIFF images

venalil:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/538

I recently updated my local OpenCV version to 3.4.3 and found out that
I could not read my TIFF images related to my project. After debugging I
found out that there has been some static analysis fixes made
that accidentally have broken reading those black-and-white TIFF images.

Commit hash in which reading of mentioned TIFF images has been broken:
cbb1e867e5141412c62ff534def7f117e28e04e8

Basically the fix is to revert back to the same functionality that has been there before.
When black-and-white images are read bpp (bitspersample) is 1.
Without the case 1: this TiffDecoder::readHeader() function always return false.

### This pullrequest changes

PR fixes reading thresholded (black-and-white) TIFF images.

```
opencv_extra=3.4
```

### DIFF
--- a/modules/imgcodecs/src/grfmt_tiff.cpp
+++ b/modules/imgcodecs/src/grfmt_tiff.cpp
@@ -49,6 +49,7 @@
 
 #ifdef HAVE_TIFF
 #include "grfmt_tiff.hpp"
+#include <stdexcept>
 #include <limits>
 
 // TODO FIXIT Conflict declarations for common types like int64/uint64
@@ -254,8 +255,8 @@ bool TiffDecoder::readHeader()
             switch(bpp)
             {
                 case 1:
-                    result = true;
                     m_type = CV_MAKETYPE(CV_8U, photometric > 1 ? wanted_channels : 1);
+                    result = true;
                     break;
                 case 8:
                     m_type = CV_MAKETYPE(CV_8U, photometric > 1 ? wanted_channels : 1);
--- a/modules/imgcodecs/src/grfmt_tiff.cpp
+++ b/modules/imgcodecs/src/grfmt_tiff.cpp
@@ -255,6 +255,7 @@ bool TiffDecoder::readHeader()
             {
                 case 1:
                     result = true;
+                    m_type = CV_MAKETYPE(CV_8U, photometric > 1 ? wanted_channels : 1);
                     break;
                 case 8:
                     m_type = CV_MAKETYPE(CV_8U, photometric > 1 ? wanted_channels : 1);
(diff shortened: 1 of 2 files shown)


modules/imgcodecs/src/grfmt_tiff.cpp
```diff
@@ -253,6 +253,9 @@ bool TiffDecoder::readHeader()
             int wanted_channels = normalizeChannelsNumber(ncn);
             switch(bpp)
             {
+                case 1:
+                    result = true;
+                    break;
```
alalek: What is about "m_type" ?
It must be initialized in this method.

Also it would be nice to add proper error reporting via "default:" case.

alalek:
Could you please provide some sample image with mentioned problem (64x64 pixels is enough) ? It would be nice to use it in tests.

alalek:

So, please:

modules/imgcodecs/src/grfmt_tiff.cpp
venalil: It was not previously initialized so tbh I have not figured it out.

Yes I agree the default case should have some error message because now it silently failed and it was frustrating.

I will figure these things out when I can.

alalek: Actually this case was not properly supported (bpp=1 case works "by mistake").
Also there are no tests for this type of images (btw, imgcodecs test data is [here](https://github.com/opencv/opencv_extra/tree/3.4/testdata/highgui/readwrite)).

venalil: Yup I know it worked by mistake, just wanted to revert the changes that broke this functionality which the original committer was unwilling to do :)

So this will take some time on my behalf...

modules/imgcodecs/src/grfmt_tiff.cpp
```diff
@@ -269,6 +273,8 @@ bool TiffDecoder::readHeader()
                     m_type = CV_MAKETYPE(CV_64F, photometric > 1 ? 3 : 1);
                     result = true;
                     break;
+            default:
+                throw std::runtime_error("Invalid bitsperpixel value read from TIFF header! Must be 1, 8, 16, 32 or 64.");
```
alalek: Use `CV_Error()` for error reporting.

venalil: Test data: https://github.com/opencv/opencv_extra/pull/538

Ok I'll change that

alalek:
@venalil Well done!

I added test for image from opencv_extra and pushed into this PR.

PR #12994, opened 2018-10-31
fixed imshow on mac

vpisarev:
the patch fixes the recently updated implementation (#12697); sometimes only the window titlebar was shown, without the window content.


alalek:
Like mentioned original patch, this should also go into 3.4 branch first.

vpisarev:
done

vpisarev:
:+1:

PR #12999, opened 2018-10-31
CircleGridClusterFinder: map circle pattern width and height correctly

fegorsch:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/542

During the cluster-based detection of circle grids, the detected circle
pattern has to be mapped to 3D-points. When doing this the width (i.e.
more circles) and height (i.e. less circles) of the pattern need to
be identified in image coordinates.

Until now this was done by assuming that the shorter side in image
coordinates (length in pixels) corresponds to the height in 3D.
This assumption does not hold if we look at the pattern from
a perspective where the projection of the width is shorter
than the projection of the height. This in turn lead to misdetections in
although the circle pattern was clearly visible.

Instead count how many circles have been detected along two edges of the
projected quadrangle and use the one with more circles as width and the
one with less as height.

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/calib3d/src/circlesgrid.cpp
+++ b/modules/calib3d/src/circlesgrid.cpp
@@ -178,7 +178,7 @@ void CirclesGridClusterFinder::findGrid(const std::vector<cv::Point2f> &points, 
     if(outsideCorners.size() != outsideCornersCount)
       return;
   }
-  getSortedCorners(hull2f, corners, outsideCorners, sortedCorners);
+  getSortedCorners(hull2f, patternPoints, corners, outsideCorners, sortedCorners);
   if(sortedCorners.size() != cornersCount)
     return;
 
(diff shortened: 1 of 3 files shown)


alalek:

Do you have some images for regression test?

alalek:
As a "bugfix" this patch should [go](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) into 3.4 branch first. We will merge changes from 3.4 into master regularly (weekly/bi-weekly).

So, please:

fegorsch:
Rebased.

Unfortunately the flag combination `CIRCLES_GRID | CALIB_CB_CLUSTERING` is currently untested, so I can't just throw my data into the existing tests. Adding the flag combination as a branch to the existing tests also doesn't work, as it then fails on existing test data.

If this is a blocking issue for this PR, I'll have to get back to adapting the tests when I got time.

alalek:
Could you please use more fresh 3.4 branch? (current base is from May 2018)

Right, current tests doesn't [cover](http://pullrequest.opencv.org/buildbot/export/opencv_releases/3_4_coverage-lin64-debug/20181101-004602_198/coverage_html/opencv/modules/calib3d/src/circlesgrid.cpp.gcov.html) `if(!isAsymmetricGrid)` code branch.

Some simple separate test will be enough (`imread()`, call detector, check result with hardcoded "golden" values).
New test data can be placed into [opencv_extra](https://github.com/opencv/opencv_extra/tree/3.4/testdata/cv/cameracalibration) repository (please use the same branch name ("improve-circle-detector") for opencv_extra PR).

PR #13001, opened 2018-10-31
G-API: Added Standalone.cmake

rgarnov:
Added Standalone.cmake to simplify standalone build. (Standalone.cmake should be included into user CMakeLists.txt to add fluid target to the project)


modules/gapi/cmake/Standalone.cmake
```diff
@@ -0,0 +1,30 @@
+if (NOT TARGET ade )
```
alalek: Please use "lowercase" in file names: standalone.

Also you may put this file one level upper.

dmatveev:
> Also you may put this file one level upper.

IMHO it is okay to have this one under `/cmake` as we do have other cmake stuff in `/cmake`

PR #13008, opened 2018-11-01
G-API GPU-OpenCL backend

dbudniko:
T-API based G-APU GPU-OCL backend implmentation + tests (perf and accuracy). 

```
buildworker:Win64 OpenCL=windows-1
```


alalek:
Please make error reports more useful:
```
/build/precommit_opencl_linux/opencv/modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp:196: Failure
Value of: cmpF(out_mat_gapi, out_mat_ocv)
  Actual: false
Expected: true
```

It is unclear what is wrong here.
- What error tolerance is used?
- What condition from these sets of check doesn't pass?
- Do we have NaN values in results?

dmatveev:
> > It is unclear what is wrong here.
> What error tolerance is used?
> What condition from these sets of check doesn't pass?
> Do we have NaN values in results?

Hi Alexander @alalek! We were wondering what are existing OpenCV ts tools to compare results flexibly and didn't figure out if there's any (at least at the moment when Maria did the tests).

As a result, we ended up with a temporary solution like above. Definitely it should be improved and I assume it should be done on OpenCV-wide level.

Let us think more on our particular cases and figure out a plan, but it wouldn't happen in the scope of this MR (which brings a GPU backend, not a test suite)

alalek:
I will try to collect crash dump from Windows OpenCL builder.

modules/gapi/include/opencv2/gapi/gpu/ggpukernel.hpp
```diff
@@ -0,0 +1,230 @@
...
+{
+    // Forward-declare an internal class
+    class GGPUExecutable;
+} // namespace gimpl
+
+namespace gapi
+{
+namespace gpu
+{
+    GAPI_EXPORTS cv::gapi::GBackend backend();
```
alalek: > CV_EXPORTS
> GAPI_EXPORTS

I believe only one should be used.

dbudniko: Done

alalek:
Crash problem is still [here](http://pullrequest.opencv.org/buildbot/builders/precommit_opencl/builds/18305/steps/test_gapi-ippicv-opencl/logs/stdio):
```
[ RUN      ] SumTestGPU/SumTest.AccuracyTest/18, where GetParam() = (3, 1280x720, false, 0.5, { gapi.kernel_package })
[       OK ] SumTestGPU/SumTest.AccuracyTest/18 (5 ms)
[ RUN      ] SumTestGPU/SumTest.AccuracyTest/19, where GetParam() = (3, 1280x720, true, 0.5, { gapi.kernel_package })
Process returned: -1073740791
ERROR: some tests have failed
program finished with exit code -1073740791
```

PR #13013, opened 2018-11-01
java: add support for MatOfRotatedRect

berak:
add rudimentary support for MatOfRotatedRect, so `Dnn.NMSBoxesRotated()` can be used with the east text detection.

### DIFF
--- a/modules/core/misc/java/gen_dict.json
+++ b/modules/core/misc/java/gen_dict.json
@@ -765,6 +765,15 @@
             "v_type": "Mat",
             "j_import": "org.opencv.core.MatOfRect2d"
         },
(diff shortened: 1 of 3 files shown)


berak:
do we need additional conversion code [here](https://github.com/opencv/opencv/blob/master/modules/java/generator/src/java/org/opencv/utils/Converters.java) ?

(i already got that (+ tests), but was afraid to add it here, clutter or useful ?)

vpisarev:
it does not hurt, let's add it! thank you! :+1:

berak:
ok, i'll do that , and thanks for the quick merge ! ;)

PR #13027, opened 2018-11-02
Replaced strcpy_s, strcat_s with strcpy,strcat when the build is done

### LINKED ISSUE #13016: Replace the use of non-portable strcpy_s, strcat_s functions from window_w32.cpp
Hello,

- OpenCV => 4.0
- Compiler => mingw-w64 GCC 8.1.0

##### Detailed description

The use of *strcpy_s* and *strcat_s* in *window_w32.cpp* makes OpenCV impossible to build with GCC (on Windows) which doesn't have the optional *_s* functions from C11. I'm not an expert in C/C++ so I would suggest to replace the use of the above with *strncpy* and *strncat* which should be present in all compilers (MSVC, GCC, Clang).

##### Steps to reproduce

Try to build OpenCV 4.0 beta with mingw-w64 GCC 8.1 on Windows.

Thanks.

ijean:
Building OpenCV with MinGW-w64

resolves #13016 

### This pullrequest changes

Replaced the use *strcpy_s* and *strcat_s*, which are available only on MSVC, in file *window_w32.cpp* with *strcpy* and *strcat*. This was wrapped in a conditional compilation and take effect only when built with MinGW.

### DIFF
--- a/modules/highgui/src/window_w32.cpp
+++ b/modules/highgui/src/window_w32.cpp
@@ -307,9 +307,13 @@ icvLoadWindowPos( const char* name, CvRect& rect )
 {
     HKEY hkey;
     char szKey[1024];
+#if defined(__MINGW32__) || defined(__MINGW64__)
+    strcpy( szKey, icvWindowPosRootKey );
+    strcat( szKey, name );
+#else
     strcpy_s( szKey, 1024, icvWindowPosRootKey );
     strcat_s( szKey, 1024, name );
-
+#endif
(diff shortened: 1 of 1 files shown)


alalek:
We prefer to avoid making of multiple compatibility changes and spead them over codebase. Especially if we didn't test this build configuration.
Adding one more usage of these functions would break MinGW again.
So It is better to make MinGW compatible with MSVC behavior in a single place.
Consider using defines from the issue's comment.

ijean:
Thanks. Will do.

PR #13030, opened 2018-11-02
G-API Tutorial

dmatveev:
This is an initial version of G-API tutorial, based on porting anisotropic image segmentation sample to G-API. This tutorial is written to close G-API documentation gaps for OpenCV 4.0 release, and will be further extended with new chapters on writing own kernels (in the post-4.0 time frame).

In addition to the tutorial itself, this PR also contains the following changes:
* Original anisotropic image segmentation tutorial has been modified to use `CV_32F` instead of `CV_64F` - result didn't change, but now it is portable;
*  Two new kernels were added to G-API: `phase()` and `sqrt()`. Both are covered with documentation, accuracy tests, and implemented in OpenCV and Fluid backends;
* A number of issues has been fixed in Fluid backend (e.g. its standard kernels were not exported);
* `GKernelPackage` has been extended with new methods for finer-grain control over its contents;
* Fluid implementation of Sobel kernel was extended to support F32, though the complete work is done by @elatkin in #13042; 
* Fluid implementation of InRange kernel was extended to support F32;
* Other issues were reported and fixed separately by @AsyaPronina.


dmatveev:
@alalek note - now it is more or less ready for merge, and so I removed *WIP* and *force_builders* stuff from the description since there's a lot more changes rather than just a documentation update. Let tests be green!

dmatveev:
How can I assign @elatkin and @mlyashko to review this? :thinking:

elatkin:
Looks **OK** for me

dmatveev:
Hey Alexander @alalek !

Due to our old issue with Android NDK, G-API doesn't build there, and so tutorial build fails.

How can I specify that tutorial should be built if and only if `gapi` module is available?

Thanks,
D

alalek:
Can we place documentation under the GAPI module directory?

```.diff
-doc/tutorials/gapi/
+modules/gapi/tutorials/
````

```.diff
-samples/cpp/tutorial_code/gapi
+modules/gapi/samples/ (flat structure)
```

Other sources can be guarded:
```
#include "opencv2/opencv_modules.hpp"
#ifdef HAVE_OPENCV_GAPI
#include "opencv2/gapi/..."
...
#else
int main()
{
    std::cout << "This tutorial code requires G-API module to be run." << std::endl;
    return 0;
}
#endif
```

dmatveev:
@alalek New changes are not expected here, let's merge this if there's no issues

PR #13036, opened 2018-11-03
java: add converters, tests for MatOfRotatedRect

berak:
follow-up to #13013

### DIFF
--- a/modules/java/generator/src/java/org/opencv/utils/Converters.java
+++ b/modules/java/generator/src/java/org/opencv/utils/Converters.java
@@ -13,7 +13,9 @@ import org.opencv.core.MatOfPoint2f;
 import org.opencv.core.MatOfPoint3f;
 import org.opencv.core.Point;
 import org.opencv.core.Point3;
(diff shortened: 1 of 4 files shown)


modules/java/generator/src/java/org/opencv/utils/Converters.java
```diff
@@ -770,4 +772,41 @@ public static void Mat_to_vector_vector_char(Mat m, List<List<Byte>> llb) {
         }
         mats.clear();
     }
+
+    public static Mat vector_RotatedRect_to_Mat(List<RotatedRect> rs) {
+        Mat res;
+        int count = (rs != null) ? rs.size() : 0;
+        if (count > 0) {
+            res = new Mat(count, 1, CvType.CV_32FC5);
```
alalek: Can we use `CvType.CV_32FC(5)` directly instead of adding of new const?

berak: sure, that was ugly, anyway ;)

PR #13041, opened 2018-11-04
docs: remove NULL strings from highgui.hpp

berak:
since it is using std::string in opencv4,  NULL is no more a valid option

```
**WIP**
force_builders=linux,docs
```


alalek:
3rd parameter is `userData`. It can be NULL.

1st parameter is called "bar_name" instead of "winname".

berak:
ah right. userdata pointer changes reverted.

PR #13042, opened 2018-11-04
GAPI: Sobel 3x3 with FP32 input

elatkin:


modules/gapi/test/cpu/gapi_imgproc_tests_fluid.cpp
```diff
@@ -36,6 +36,24 @@ class AbsToleranceFluid : public Wrappable<AbsToleranceFluid>
...
+        cv::absdiff(in1, in2, diff);
+        a1 = cv::abs(in1);
+        a2 = cv::abs(in2);
+        cv::max(a1, a2, b);
+        cv::max(1, b, base);  // base = max{1, |in1|, |in2|}
+        return cv::countNonZero(diff > tolerance*base) == 0;
+    }
+private:
+    double tolerance;
+};
```
dmatveev: I assume this comparison operator needs to be extended to compare with tolerance only for F32 case, and remain exact for all other cases (like it used to be).

elatkin: I would rather think about splitting the test's instantiation into 2 parts:  
* with this new tolerance-driven acceptance criterion  for FP32 input,  
* and with stricter criterion (exact equality) for integral-type input.  

Please note that this code of Sobel 3x3 would meet the stricter acceptance  
criterion for an integral input even if output type is formally floating-point.

elatkin: @dmatveev  
Please, see the alternative MR https://github.com/opencv/opencv/pull/13052

elatkin:
Closing this PR as replaced with https://github.com/opencv/opencv/pull/13052

PR #13052, opened 2018-11-06
GAPI: Sobel 3x3 with FP32 input, v2

elatkin:


elatkin:
This pull request is the alternative to the https://github.com/opencv/opencv/pull/13042  

The difference is that this PR keeps the old test intact (for Sobel 3x3 with integral arguments)

PR #13054, opened 2018-11-06
highgui: fix waitKey() with closed windows

### LINKED ISSUE #13049: Win32 HighGUI Backend - waitKey() no longer exits when window is closed
- OpenCV: 3.4.2 (first occurrence), 3.4.3, 4.0.0-beta (standard distribution builds)
- Platform: Win10 x64 (using both 32bit and 64bit builds)
- Language: Python
----

In 3.4.1 and earlier, you'd get a -1 from `waitKey` when you closed the (single) open window.

Starting with 3.4.2, this is no longer the case. Closing the window leaves `waitKey` running, and since the window is no longer visible, it can't receive keyboard events, and hangs there.

There seems to be only one commit between 3.4.1 and 3.4.2 that touches relevant sources: https://github.com/opencv/opencv/commit/8620bd5a84b03933fe24a3d8ba50715b617837b0

I suspect the cause is due to removal of the test `hg_windows == 0`.

----
Script to reproduce:

```
import cv2
import numpy as np
img = np.ones((128,128), np.uint8)
cv2.imshow('', img)
cv2.waitKey()
```

alalek:
resolves #13049
relates #11714

### DIFF
--- a/modules/highgui/src/window_w32.cpp
+++ b/modules/highgui/src/window_w32.cpp
@@ -1976,13 +1976,15 @@ cvWaitKey( int delay )
         MSG message;
         int is_processed = 0;
 
-        if( delay <= 0 )
+        if (delay <= 0 && hg_windows != 0)
+        {
             GetMessage(&message, 0, 0, 0);
+        }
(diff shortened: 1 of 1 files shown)


alalek:
#13053

PR #13060, opened 2018-11-06
removed C API in the following modules: photo, video, imgcodecs, videoio

vpisarev:
n/t

```
force_builders=Custom
docker_image:Custom=fedora:28
```


modules/imgcodecs/include/opencv2/imgcodecs.hpp
```diff
@@ -53,6 +53,52 @@
   @}
 */
 
+/* duplicate of "ImreadModes" enumeration for better compatibility with OpenCV 3.x */
+enum
+{
+/* 8bit, color or not */
+    CV_LOAD_IMAGE_UNCHANGED  =-1,
```
alalek: Can we keep them in other header files? For example, `imgcodecs/c_api_compatibility.hpp` or current `imgcodecs.h` (under `__cplusplus`).

Main purpose of this to avoid using of these deprecated symbols:
- in OpenCV code itself
- in bindings parsers
- don't bloat documentation

vpisarev:
ok, let's merge it; then will create a separate header

PR #13063, opened 2018-11-07
Fix CMake find_package(OpenCV) for Windows when BUILD_SHARED_LIBS is not defined

xsacha:
### This pullrequest changes
Which library the find_package(OpenCV) loads when BUILD_SHARED_LIBS is not defined. This matches the standard practice as well as the library that is compiled by OpenCV when this variable is not set.

This has been broken since OpenCV 3.4.2.
Caused by https://github.com/opencv/opencv/commit/abce51fb7ba901ddf9480a127c727e514a951f90 which was matching the OpenCVDetectCXXCompiler.cmake code which has always had this incorrect logic.

### DIFF
--- a/cmake/OpenCVDetectCXXCompiler.cmake
+++ b/cmake/OpenCVDetectCXXCompiler.cmake
@@ -111,10 +111,10 @@ endif()
 # Similar code exists in OpenCVConfig.cmake
 if(NOT DEFINED OpenCV_STATIC)
   # look for global setting
-  if(NOT DEFINED BUILD_SHARED_LIBS OR BUILD_SHARED_LIBS)
-    set(OpenCV_STATIC OFF)
-  else()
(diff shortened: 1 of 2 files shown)


alalek:
1. Link time. Please verify link time with/without static link. Difference is dramatic.
2. Size of "empty" OpenCV app is more than 6Mb.
3. We don't ship staticlibs in Winpack anymore.
4. There are options to change this default behavior. So we don't see any problem here.
5. What is version of CRT assigned by CMake (`/MT` or `/MD`)?

So this patch breaks existed builds and provides inconvenient build time.

xsacha:
Not sure how any of that relates.
If you were building dynamic libraries, they will still be this way.

xsacha:
Closing for now to rework in to proper solution.

PR #13068, opened 2018-11-07
objdetect: 'const' correctness

### LINKED ISSUE #9891: detectMultiScale is not const
- OpenCV => 3.3
- Operating System / Platform => Ubuntu 16.04 x86-64
- Compiler => g++ 5.4.0

##### Detailed description

`CascadeClassifier::detectMultiScale` (https://docs.opencv.org/3.3.0/d1/de5/classcv_1_1CascadeClassifier.html#aaf8181cb63968136476ec4204ffca498) is not `const`, but it should be, in my opinion.

##### Steps to reproduce

In my opinion, it should be possible to build a function like the following, since detection should not alter the internal state of the cascade (as no training is happening):

```
static void FindFaces(const Mat &image, /*const*/ CascadeClassifier &classifier, vector<Rect> &faces) //detectMultiScale should be const, but is not
{
  classifier.detectMultiScale(image, faces /* optional parameters here */);
}
```

Maybe the method can even be made `const` even in the base class, `BaseCascadeClassifier` (https://docs.opencv.org/3.3.0/da/dd5/classcv_1_1BaseCascadeClassifier.html#a68623391e16daa264f089af99d8d4bbb)

alalek:
resolves #9891

### DIFF
--- a/modules/objdetect/include/opencv2/objdetect.hpp
+++ b/modules/objdetect/include/opencv2/objdetect.hpp
@@ -181,14 +181,14 @@ public:
                            CV_OUT std::vector<Rect>& objects,
                            double scaleFactor,
                            int minNeighbors, int flags,
-                           Size minSize, Size maxSize ) = 0;
+                           Size minSize, Size maxSize ) const = 0;
 
     virtual void detectMultiScale( InputArray image,
                            CV_OUT std::vector<Rect>& objects,
                            CV_OUT std::vector<int>& numDetections,
                            double scaleFactor,
                            int minNeighbors, int flags,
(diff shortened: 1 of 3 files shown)


vpisarev:
I left comment in @9891. In my opinion, detect should remain "non-const" method

vpisarev:
:-1:

asmorkalov:
The patch was discussed by @alalek @vladimir-dudnik @vpisarev and @asmorkalov offline. Formal constant correctness introduces optimization pitfalls and require not obvious solutions like thread local variables for state storage to support threads. It was decided to stay with existing interface.

PR #13070, opened 2018-11-07
GAPI (fluid): optimization of Sobel 3x3

elatkin:
```
force_builders=Linux AVX2
```


elatkin:
This PR optimizes Sobel 3x3 with "wide universal" intrinsics (CV_SIMD),  
and implements *static* dispatching to x86/SSE/AVX (or ARM/Neon, or...)  

Dynamic dispatcher implemented with separate PR https://github.com/opencv/opencv/pull/13133  
Yet it fails due to known bug in CV_SIMD dynamic dispatcher

elatkin:
CI failed at three configs:  
* 2 failures at OpenCL  
* 1 failure at ARMv7

OpenCL failure in stitching is known, fix is WIP by @alalek  

ARMv7 failure is hanging of OpenCV/Core tests

elatkin:
All test failures look fixed now.

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
```diff
@@ -476,15 +476,15 @@ PERF_TEST_P_(SobelPerfTest, TestPerformance)
...
+    auto out = cv::gapi::Sobel(in, dtype, dx, dy, kernSize);
     cv::GComputation c(in, out);
 
     // Warm-up graph engine:
     c.apply(in_mat1, out_mat_gapi, std::move(compile_args));
 
     TEST_CYCLE()
     {
-        c.apply(in_mat1, out_mat_gapi, std::move(compile_args));
+        c.apply(in_mat1, out_mat_gapi);
```
alalek: @dmatveev Is it correct change?

dmatveev: Yes and no. since `compile_args` are moved twice, the latter call to apply in fact passes an empty array - which has no effect (the underlying graph is not recompiled).

dmatveev: It works now but our implementation of `apply()` doesn't check contents of compile args so if we fix this problem, code like this will break. The issue is known and tracked internally alreday

modules/gapi/src/backends/fluid/gfluidimgproc.cpp
```diff
@@ -756,30 +761,14 @@ static void run_sobel(Buffer& dst,
...
     int chan  = dst.meta().chan;
 
-    for (int w=0; w < width; w++)
-    {
-        // TODO: make this cycle innermost
-        for (int c=0; c < chan; c++)
-        {
-            float sum=0;
+    GAPI_DbgAssert(ksize == 3);
+//  float buf[3][width * chan];
```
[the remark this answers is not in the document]
elatkin: This code would not work correctly if ksize is not 3; so yes, double-checking ksize looks good idea for me.  

The 2nd (commented) line reminds the dimensions of the *buf[] array. I'd leave it here for code readability.

PR #13072, opened 2018-11-07
Fixed pendantic warning

gineshidalgo99:
Resolves Pedantic warning of the kind: `extra ‘;’ [-Wpedantic]` (no open issue about it that I am aware of)

### This pullrequest changes

When enabled all warnings, it appears quite often a Pedantic warning of this kind:
```
/usr/local/include/opencv4/opencv2/features2d.hpp:1202:32: warning: extra ‘;’ [-Wpedantic]
 CV_ENUM_FLAGS(DrawMatchesFlags);
```
And analogously for every time `CV_ENUM_FLAGS` is used. This is quite annoying, so I have fixed it.

Thanks!

### DIFF
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -425,7 +425,7 @@ as well as exposing the C++11 enum class members for backwards compatibility
 
 @code
     // Provides operators required for flag enums
-    CV_ENUM_FLAGS(AccessFlag);
+    CV_ENUM_FLAGS(AccessFlag)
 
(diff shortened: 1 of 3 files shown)


gineshidalgo99:
The macro is a function declaration. Adding a `;` to a function is provoking the warning. That's why the macro itself can't be fixed, as it is a function.

Note: It does not only happen on `build/install`, it happens when I use externally the library, as those macros occur in the headers, not in the cpp files.

alalek:
> I use externally the library

GCC world uses `-isystem` to suppress warnings from external headers.

PR #13073, opened 2018-11-07
highgui: drop obsolete Carbon API

alalek:
https://en.wikipedia.org/wiki/Carbon_(API)

> In 2012, with the release of OS X 10.8 Mountain Lion, most Carbon APIs were considered deprecated.


alalek:
:+1:

PR #13074, opened 2018-11-07
videoio: drop obsolete backends

alalek:
- VFW
- QuickTime/QtKit
- Unicap - GPL, no active support: https://github.com/unicap/unicap
- DC1394 (1st version) / CMU1394


alalek:
:+1:

PR #13080, opened 2018-11-08
core: fix processing of vector-rows

### LINKED ISSUE #13078: Mat::convertTo  error
- OpenCV => 3.4.3
- Operating System / Platform => OpenSUSE Tumbleweed 20181105
- Compiler => gcc 8.2.1

##### Steps to reproduce
 ```.cpp
std::vector<float> floats = {1, 2, 3, 4, 5, 6, 7, 8};
std::vector<int> ints(floats.size() / 2);

cv::Mat(ints.size(), 1, CV_32FC1, floats.data(), sizeof(floats[0]) * 2)
            .convertTo(ints, CV_32S);
```

The code may cause segmentation fault or get error result `ints == [1, 0, 0, 0]`. But if I use
`std::array<int, 4>` for `ints`, then correct result `ints == [1, 3, 5, 7]` gets.

### PR DESCRIPTION
**Pull request by alalek**
resolves #13078

### DIFF
--- a/modules/core/src/arithm.cpp
+++ b/modules/core/src/arithm.cpp
@@ -199,6 +199,8 @@ static void binary_op( InputArray _src1, InputArray _src2, OutputArray _dst,
             func = tab[depth1];
 
         Mat src1 = psrc1->getMat(), src2 = psrc2->getMat(), dst = _dst.getMat();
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by huangqinjin** [comment]
```
std::vector<float> floats;
OutputArray a = floats;
a.create(2, 1, CV_32F);
Mat m = a.getMat(); 
```
`m` should be 2x1 but now is 1x2. Why not update the member `sz` in `create`?

PR #13085, opened 2018-11-08
significantly reduced OpenCV binary size by disabling IPP in some funcs

vpisarev:
the functions where IPP was disabled: *Sobel, Scharr, medianBlur, GaussianBlur, filter2D, mean, meanStdDev, sum, minMaxIdx, sort*.

`opencv_core` size is reduced by ~2Mb (19 down to 17), `opencv_imgproc` size is reduced by ~24Mb (45 down to 21); other modules did not change, but they are much smaller than core and imgproc.

The patch should not affect OpenCV performance much, since each of those functions is either very fast as-is, or optimized with SIMD instructions, or both. `cv::filter2D` uses IPP-optimized DFT, so the performance should not drop either.

### DIFF
--- a/modules/core/src/matrix_operations.cpp
+++ b/modules/core/src/matrix_operations.cpp
@@ -8,6 +8,11 @@
 #include "opencl_kernels_core.hpp"
 #include "precomp.hpp"
 
+#undef HAVE_IPP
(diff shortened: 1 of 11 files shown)


vpisarev:
:+1:

PR #13086, opened 2018-11-09
refined QRCodeDetector API for OpenCV 4.0

vpisarev:
n/t


alalek:
To avoid merge hell the huge part of these changes must go into 3.4 first (without API removal).

Issue #13087, opened 2018-11-09
G-API: OpenCL backend issues


alalek:
1. Windows-OpenCL (workarounded by #13094, similar problem with Canny, fixed by #13269)
    [Nightly build](http://pullrequest.opencv.org/buildbot/builders/master-win64-vc14/builds/10630):

    opencv_test_gapi / opencv_perf_gapi:

```
[ RUN      ] SumTestGPU/SumTest.AccuracyTest/18, where GetParam() = (3, 1280x720, false, 0.5, { gapi.kernel_package })
[       OK ] SumTestGPU/SumTest.AccuracyTest/18 (5 ms)
[ RUN      ] SumTestGPU/SumTest.AccuracyTest/19, where GetParam() = (3, 1280x720, true, 0.5, { gapi.kernel_package })
Process returned: -1073740791
ERROR: some tests have failed
program finished with exit code -1073740791
```

```
[ RUN      ] NormPerfTestGPU/NormPerfTest.TestPerformance/20, where GetParam() = (2, 128x128, 8UC1, 4, { gapi.kernel_package })
[ PERFSTAT ]    (samples=1   mean=0.19   median=0.19   min=0.19   stddev=0.00 (0.0%))
[       OK ] NormPerfTestGPU/NormPerfTest.TestPerformance/20 (13 ms)
[ RUN      ] NormPerfTestGPU/NormPerfTest.TestPerformance/21, where GetParam() = (2, 128x128, 8UC3, 4, { gapi.kernel_package })
Process returned: -1073740791
ERROR: some tests have failed
program finished with exit code -1073740791
```

2. Mac OpenCL
    [Nightly build](http://pullrequest.opencv.org/buildbot/builders/master-mac/builds/10629):
```
[ RUN      ] NormTestGPU/NormTest.AccuracyTest/NORM_L1_5_1280x720, where GetParam() = (2, 5, 1280x720, 0.04, { 32-byte object <26-67 61-70 69-2E 6B-65 72-6E 65-6C 5F-70 61-63 6B-61 67-65 00-00 00-00 80-72 E4-69 E1-7F 00-00> })
/build/master-mac/opencv/modules/gapi/test/common/gapi_core_tests_inl.hpp:805: Failure
Expected: (abs(out_norm[0] - out_norm_ocv[0])) <= (tolerance), actual: 0.269218 vs 0.04
[...]

PR #13094, opened 2018-11-09
OCL: core: run "reduce" kernels in synchronous mode

alalek:
Workaround to avoid OpenCL runtime crashing due large amount of kernel events.

relates #13087

<cut />

One of possible Linux crash point:
https://github.com/intel/compute-runtime/blob/18.33.11309/runtime/os_interface/linux/drm_buffer_object.cpp#L167

```
buildworker:Linux AVX2=linux-3
```


alalek:
:+1:

PR #13115, opened 2018-11-11
Fix missing intrin.hpp header in dis_flow.cpp

catree:
resolves opencv/opencv_contrib#1894

Geometric mean (ms)

|Name of Test|before|after|before vs after (x-factor)|
|---|:-:|:-:|:-:|
|perf::DenseOpticalFlow_DIS::("PRESET_FAST", 640x480)|14.602|13.442|1.09|
|perf::DenseOpticalFlow_DIS::("PRESET_FAST", 1280x720)|40.623|36.206|1.12|
|perf::DenseOpticalFlow_DIS::("PRESET_FAST", 1920x1080)|86.579|77.698|1.11|
|perf::DenseOpticalFlow_DIS::("PRESET_MEDIUM", 640x480)|53.288|41.514|1.28|
|perf::DenseOpticalFlow_DIS::("PRESET_MEDIUM", 1280x720)|177.925|137.496|1.29|
|perf::DenseOpticalFlow_DIS::("PRESET_MEDIUM", 1920x1080)|436.096|332.447|1.31|
|perf::DenseOpticalFlow_DIS::("PRESET_ULTRAFAST", 640x480)|7.564|6.485|1.17|
|perf::DenseOpticalFlow_DIS::("PRESET_ULTRAFAST", 1280x720)|23.024|19.151|1.20|
|perf::DenseOpticalFlow_DIS::("PRESET_ULTRAFAST", 1920x1080)|49.114|38.910|1.26|
|perf::OCL_DenseOpticalFlow_DIS::("PRESET_FAST", 640x480)|15.049|14.310|1.05|
|perf::OCL_DenseOpticalFlow_DIS::("PRESET_FAST", 1280x720)|40.428|36.169|1.12|
|perf::OCL_DenseOpticalFlow_DIS::("PRESET_FAST", 1920x1080)|84.386|76.711|1.10|
|perf::OCL_DenseOpticalFlow_DIS::("PRESET_MEDIUM", 640x480)|48.509|37.854|1.28|
|perf::OCL_DenseOpticalFlow_DIS::("PRESET_MEDIUM", 1280x720)|156.342|123.468|1.27|
|perf::OCL_DenseOpticalFlow_DIS::("PRESET_MEDIUM", 1920x1080)|390.009|288.881|1.35|
|perf::OCL_DenseOpticalFlow_DIS::("PRESET_ULTRAFAST", 640x480)|10.223|9.721|1.05|
|perf::OCL_DenseOpticalFlow_DIS::("PRESET_ULTRAFAST", 1280x720)|28.411|23.203|1.22|
|perf::OCL_DenseOpticalFlow_DIS::("PRESET_ULTRAFAST", 1920x1080)|58.171|48.353|1.20|

### DIFF
--- a/modules/video/src/dis_flow.cpp
+++ b/modules/video/src/dis_flow.cpp
@@ -41,6 +41,7 @@
 //M*/
 
 #include "precomp.hpp"
+#include "opencv2/core/hal/intrin.hpp"
 #include "opencl_kernels_video.hpp"
 
 using namespace std;
@@ -517,7 +518,7 @@ inline float processPatch(float &dst_dUx, float &dst_dUy, uchar *I0_ptr, uchar *
                           int I0_stride, int I1_stride, float w00, float w01, float w10, float w11, int patch_sz)
 {
     float SSD = 0.0f;
-#ifdef CV_SIMD128
+#if CV_SIMD128
     if (patch_sz == 8)
     {
         /* Variables to accumulate the sums */
@@ -569,7 +570,7 @@ inline float processPatch(float &dst_dUx, float &dst_dUy, uchar *I0_ptr, uchar *
                 dst_dUx += diff * I0x_ptr[i * I0_stride + j];
                 dst_dUy += diff * I0y_ptr[i * I0_stride + j];
             }
-#ifdef CV_SIMD128
+#if CV_SIMD128
     }
 #endif
     return SSD;
@@ -586,7 +587,7 @@ inline float processPatchMeanNorm(float &dst_dUx, float &dst_dUy, uchar *I0_ptr,
     float sum_I0x_mul = 0.0, sum_I0y_mul = 0.0;
     float n = (float)patch_sz * patch_sz;
 
-#ifdef CV_SIMD128
+#if CV_SIMD128
     if (patch_sz == 8)
     {
         /* Variables to accumulate the sums */
@@ -641,7 +642,7 @@ inline float processPatchMeanNorm(float &dst_dUx, float &dst_dUy, uchar *I0_ptr,
                 sum_I0x_mul += diff * I0x_ptr[i * I0_stride + j];
                 sum_I0y_mul += diff * I0y_ptr[i * I0_stride + j];
             }
-#ifdef CV_SIMD128
+#if CV_SIMD128
     }
 #endif
     dst_dUx = sum_I0x_mul - sum_diff * x_grad_sum / n;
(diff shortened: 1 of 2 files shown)


tompollok:
@catree @alalek  I checked also a little bit on this and thought about whether there could be also other algorithms in opencv that lack in the inclusion of  "opencv2/core/hal/intrin.hpp"

So in core module intrin.hpp is included via precomp.hpp
See: https://github.com/opencv/opencv/blob/master/modules/core/src/precomp.hpp

On the other hand in video module intrin.hpp is not included by precomp (every module has its own precomp.hpp for some reason)
https://github.com/opencv/opencv/blob/master/modules/video/src/precomp.hpp

I checked all modules of opencv and opencv_contrib and actually only the precomp.hpp of core includes  "opencv2/core/hal/intrin.hpp"
It is only included by specific algorithms cpp file.

Does it make sense to include by all other precomp.hpp of all other modules or should it specifically only be included by the actual cpp implemention files that make use of them? Because if we would do so then this kind of bug would never bother us anymore. I dont know it there is a significant impact on the compile time to not do it.

alalek:
> include by all other precomp.hpp

No

> be included by the actual cpp implemention files that make use of them

Yes


> this kind of bug would never bother us anymore

**Need to change**:

```.diff
-#ifdef CV_SIMD128
+#if CV_SIMD128
```

anywhere.

"ifdef" approach in "dis_optflow.cpp" and "variational_refinement.cpp " is invalid.

tompollok:
Thank you.

I added a PR for opencv_contrib 3.4 branch.

Am I assuming right that all cpp implementation files that use CV_AVX2, CV_SIMD, CV_VSX, CV_NEON, CV_SSE need an inclusion of #include "opencv2/core/hal/intrin.hpp" except for the core module implementations as they get it from precomp.hpp?

tompollok:
@alalek according to our discussion yesterday i went throu the source to check for more locations that might lack "opencv2/core/hal/intrin.hpp". Arent the following files also missing the inclusion of "opencv2/core/hal/intrin.hpp"?

https://github.com/opencv/opencv/blob/master/modules/dnn/src/layers/batch_norm_layer.cpp
https://github.com/opencv/opencv/blob/master/modules/dnn/src/layers/elementwise_layers.cpp
https://github.com/opencv/opencv/blob/master/modules/dnn/src/layers/fully_connected_layer.cpp

tompollok:
Thanks for answering and providing a pr. I was busy the last days.

PR #13120, opened 2018-11-12
Custom OpenCL G-API kernel example.

dbudniko:
### This pullrequest changes

This is example of custom OpenCL kernel for G-API module. The approach is very similar to other OpenCV modules. Fluid kernels approach is the reference. Functionality is added to G-API accuracy test application. Doesn't require G-API module rebuild.


dmatveev:
The main idea of custom kernels is that they are application specific - that means, no changes would be required in the library itself to support a new kernel.

For example, we do have a lot of test kernels for Fluid which exists in tests only. We only need to build test application itself to use (run) these kernels.

Same should apply to custom OpenCL kernels as well. Every application might have their own (which also could be IP), so rebuilding OpenCV to integrate any of them is not an option.

The design goal for this is to introduce (or extend) G-API GPU kernel API such way that utilizing a custom kernel in a graph would not require changes in the module itself.

How this could be done?

modules/gapi/CMakeLists.txt
```diff
@@ -87,6 +87,45 @@ ocv_add_dispatched_file(backends/fluid/gfluidimgproc_func SSE4_1 AVX2)
...
+  list(APPEND gapi_srcs ${cl_kernels} "${CMAKE_CURRENT_BINARY_DIR}/${OCL_NAME}.cpp" "${CMAKE_CURRENT_BINARY_DIR}/${OCL_NAME}.hpp")
+endif()
+
+file(GLOB cl_kernels_test
+     "${CMAKE_CURRENT_LIST_DIR}/test/opencl/*.cl"
+)
+if(cl_kernels_test)
+  set(OCL_NAME_TEST opencl_kernels_test_${name})
+  add_custom_command(
+    OUTPUT "${CMAKE_CURRENT_LIST_DIR}/test/${OCL_NAME_TEST}.cpp"  # don't add .hpp file here to optimize build process
```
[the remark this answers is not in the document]
dbudniko: Done. Writing in BIN now.

modules/gapi/test/opencl/symm7x7_test.cl
```diff
@@ -0,0 +1,302 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2018 Intel Corporation
+
+#if BORDER_REPLICATE
```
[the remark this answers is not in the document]
dbudniko: I just trying to demonstrate how to easaly reuse existing OpenCL kernel. This kernel is the part of OpenVX color copy pipeline. I just pick up this kernel as is and created G-API custom kernel example.

modules/gapi/CMakeLists.txt
```diff
@@ -98,14 +99,49 @@ ocv_module_include_directories("${CMAKE_CURRENT_LIST_DIR}/src")
...
+# This adds custom OpenCL kernel(s) to gapi accuracy tests
+# The similar approach can be used in custom application which uses OpenCV G-API
+  file(GLOB cl_kernels_test
+       "${CMAKE_CURRENT_LIST_DIR}/test/opencl/*.cl"
+  )
+  if(cl_kernels_test)
+    set(OCL_NAME_TEST opencl_kernels_test_${name})
+    add_custom_command(
+      OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${OCL_NAME_TEST}.cpp"  # don't add .hpp file here to optimize build process
+      COMMAND ${CMAKE_COMMAND} "-DMODULE_NAME=${name}" "-DCL_DIR=${CMAKE_CURRENT_LIST_DIR}/test/opencl" "-DOUTPUT=${CMAKE_CURRENT_BINARY_DIR}/${OCL_NAME_TEST}.cpp" -P "${OpenCV_SOURCE_DIR}/cmake/cl2cpp.cmake"
```
dmatveev: Not sure if custom application which is using _prebuilt OpenCV_ (as the most users are - e.g. downloaded as a binary package from distro's repository), it's unlikely there's `${OpenCV_SOURCE_DIR}` available in such installation.

I'd recommend to just reading a .cl file from disk - or putting its string version to the source manually (well, then the kernel should be simplier too).

@alalek would there be any issues with such approach? E.g. with respect to CI infrastructure

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -154,6 +154,8 @@ namespace imgproc {
             return in.withType(CV_8U, 1);
         }
     };
+
+
```
dmatveev: This change is not needed?

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -671,6 +673,7 @@ Output image must be 8-bit unsigned 3-channel image @ref CV_8UC3.
 GAPI_EXPORTS GMat YUV2RGB(const GMat& src);
 
 //! @} gapi_colorconvert
+
```
dmatveev: This change is not needed?

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
```diff
@@ -905,5 +905,7 @@ PERF_TEST_P_(YUV2BGRPerfTest, TestPerformance)
 
 //------------------------------------------------------------------------------
 
+
+
```
dmatveev: This change is not needed?

modules/gapi/test/common/gapi_core_tests_inl.hpp
```diff
@@ -684,8 +684,8 @@ TEST_P(SumTest, AccuracyTest)
     cv::Size sz_in = std::get<1>(param);
     auto tolerance = std::get<3>(param);
     auto compile_args = std::get<4>(param);
-    //initMatrixRandU(std::get<0>(param), sz_in, std::get<2>(param));
-    initMatsRandN(std::get<0>(param), sz_in, std::get<2>(param)); //TODO: workaround trying to fix SumTest failures
+    initMatrixRandU(std::get<0>(param), sz_in, std::get<2>(param));
+    //initMatsRandN(std::get<0>(param), sz_in, std::get<2>(param)); //TODO: workaround trying to fix SumTest failures
```
dmatveev: Does this change actually fix the problem with Sum?

modules/gapi/test/cpu/gapi_imgproc_tests_cpu.cpp
```diff
@@ -235,4 +235,5 @@ INSTANTIATE_TEST_CASE_P(YUV2BGRTestCPU, YUV2BGRTest,
 /*init output matrices or not*/ testing::Bool(),
                                 Values(cv::compile_args(IMGPROC_CPU))));
 
+
```
dmatveev: This change is not needed?

modules/gapi/test/gapi_gpu_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include "test_precomp.hpp"
+
+#include "opencv2/gapi/core.hpp"
+
+//#include "opencv2/gapi/fluid/gfluidbuffer.hpp"
+//#include "opencv2/gapi/fluid/gfluidkernel.hpp"
+
+ // FIXME: move these tests with priv() to internal suite
+//#include "backends/fluid/gfluidbuffer_priv.hpp"
```
dmatveev: These headers are not needed here?

modules/gapi/test/gapi_gpu_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+        cf[i] = coefficients[i];
+    }
+
+    cv::filter2D(in_mat, out_mat_ocv, CV_8UC1, kernel, anchor, delta, BORDER_REPLICATE);
+
+    //cv::imshow("Output OCV", out_mat_ocv);
+    //cv::imshow("Output GAPI", out_mat_gapi);
+    //cv::Mat diff = out_mat_ocv - out_mat_gapi;
+    //cv::imshow("DIFF", diff);
+    //int key = cv::waitKey(0);
```
dmatveev: This can be removed?

modules/gapi/test/gapi_gpu_test_kernels.cpp
```diff
@@ -0,0 +1,141 @@
...
+            cv::Mat kernel = cv::Mat(7, 7, CV_32FC1);
+            float* cf = kernel.ptr<float>();
+            for (int i = 0; i < 49; i++)
+            {
+                cf[i] = coefficients[i];
+            }
+
+            cv::filter2D(in, out, CV_8UC1, kernel, anchor, delta, cv::BORDER_REPLICATE);
+        }
+    }
```
dmatveev: I think this code branch can be moved to a separate function (in the same file) and then reused in (a) fall-back path here and (b) accuracy check in the test itself.

modules/gapi/test/gapi_gpu_test_kernels.cpp
```diff
@@ -0,0 +1,141 @@
...
+};
+
+
+cv::gapi::GKernelPackage gpuTestPackage = cv::gapi::kernels
+        <GGPUSymm7x7_test
+        >();
```
dmatveev: I also think we might move this stuff to a single file (gapi_gpu_test.cpp) - there's no technical need to have kernels defined in headers, and at the moment it is used in a single place only.

modules/gapi/test/gapi_gpu_test_kernels.hpp
```diff
@@ -0,0 +1,33 @@
...
+        };
+
+
+        extern cv::gapi::GKernelPackage gpuTestPackage;
+
+    } // namespace gapi_test_kernels
+#endif
+} // namespace cv
+
+#endif // GAPI_GPU_TEST_KERNELS_HPP
```
dmatveev: Same applies to this file - I think we could keep test kernel definitions and implementations at the place where we need it (if it is just about a single test)

modules/gapi/test/opencl/symm7x7_test.cl
```diff
@@ -0,0 +1,302 @@
...
+
+    float8 J = (float8)(a0_sum[0], a1_sum[0], a2_sum[0], b_sum[0].s0123, b_sum[0].s4) +
+               (float8)(b_sum[0].s3, b_sum[0].s4567, c0_sum[0], c1_sum[0], c2_sum[0]);
+    intermediate += J * (float)coeff[9];
+
+    intermediate *= SCALE;
+
+    vstore8(convert_uchar8_sat(intermediate), 0, (__global uchar*)(pdst));
+
+}
```
dmatveev: I assumes it expects a newline at the end of this file

modules/gapi/test/gapi_gpu_test_kernels.cpp
```diff
@@ -0,0 +1,141 @@
...
+            static const char * const borderMap[] = { "BORDER_CONSTANT", "BORDER_REPLICATE", "BORDER_UNDEFINED" };
+            std::string build_options = " -D BORDER_CONSTANT_VALUE=" + std::to_string(0) +
+                " -D " + borderMap[1] +
+                " -D SCALE=1.f/" + std::to_string(1 << shift) + ".f";
+
+
+            if (!kernel.create("symm_7x7_test", cv::ocl::gapi::symm7x7_test_oclsrc, build_options))
+            {
+                printf("symm_7x7_test OpenCL kernel creation failed with build_options = %s\n", build_options.c_str());
+            }
```
dmatveev: So would kernel here be compiled only once or on every execution (`run()`)? cc: @alalek

modules/gapi/test/gapi_gpu_test_kernels.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+            static const char * const borderMap[] = { "BORDER_CONSTANT", "BORDER_REPLICATE", "BORDER_UNDEFINED" };
+            std::string build_options = " -D BORDER_CONSTANT_VALUE=" + std::to_string(0) +
+                " -D " + borderMap[1] +
+                " -D SCALE=1.f/" + std::to_string(1 << shift) + ".f";
+
+
+            if (!kernel.create("symm_7x7_test", cv::ocl::gapi::symm7x7_test_oclsrc, build_options))
+            {
+                printf("symm_7x7_test OpenCL kernel creation failed with build_options = %s\n", build_options.c_str());
```
dmatveev: If it is failed, I assume execution should be cancelled

modules/gapi/test/gapi_gpu_test_kernels.cpp
```diff
@@ -0,0 +1,141 @@
...
+            idxArg = kernel.set(idxArg, cv::ocl::KernelArg::PtrWriteOnly(out));
+            idxArg = kernel.set(idxArg, (int)out.step);
+            idxArg = kernel.set(idxArg, (int)size.height);
+            idxArg = kernel.set(idxArg, (int)size.width);
+            idxArg = kernel.set(idxArg, (int)tile_y);
+            idxArg = kernel.set(idxArg, cv::ocl::KernelArg::PtrReadOnly(gKer));
+
+            if (!kernel.run(2, globalsize, NULL, false))
+            {
+                printf("symm_7x7_test OpenCL kernel run failed\n");
```
dmatveev: Same here?

modules/gapi/test/gapi_gpu_test_kernels.cpp
```diff
@@ -0,0 +1,141 @@
...
+            cv::Size size = in.size();
+            size_t globalsize[2] = { (size_t)size.width, (size_t)size.height };
+
+            //size_t maxWorkItemSizes[32];
+            //cv::ocl::Device::getDefault().maxWorkItemSizes(maxWorkItemSizes);
+
+            cv::ocl::Kernel kernel;
+
+            int coefficients[10] = { 1140, -118, 526, 290, -236, 64, -128, -5, -87, -7 };
+            int shift = 10;
```
dmatveev: So real kernels usually have this as parameters (of the kernel) - can it be made here so? Then it would be more "complete" example showing how almost arbitrary parameters can be passed from G-API to the underlying OpenCL kernel

modules/gapi/test/gapi_gpu_test_kernels.cpp
dbudniko: It will be compiled only once, I believe. Next time it will be retrieved from the cache.

modules/gapi/include/opencv2/gapi/imgproc.hpp
dbudniko: Done

modules/gapi/include/opencv2/gapi/imgproc.hpp
dbudniko: Done

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
dbudniko: Done

modules/gapi/test/common/gapi_core_tests_inl.hpp
dbudniko: No. Removed obsolete comment.

modules/gapi/test/cpu/gapi_imgproc_tests_cpu.cpp
dbudniko: Done

modules/gapi/test/gapi_gpu_test.cpp
dbudniko: Done

modules/gapi/test/gapi_gpu_test_kernels.cpp
dbudniko: Done

modules/gapi/test/gapi_gpu_test_kernels.cpp
dbudniko: Done

modules/gapi/test/opencl/symm7x7_test.cl
dbudniko: Done

modules/gapi/test/gapi_gpu_test_kernels.cpp
dbudniko: Done

modules/gapi/test/gapi_gpu_test_kernels.cpp
dbudniko: Done

modules/gapi/test/gapi_gpu_test_kernels.cpp
dbudniko: Done

modules/gapi/test/gapi_gpu_test_kernels.cpp
dbudniko: Done

modules/gapi/test/gapi_gpu_test_kernels.cpp
dbudniko: Done

modules/gapi/test/gapi_gpu_test_kernels.cpp
dbudniko: Done

modules/gapi/test/opencl/symm7x7_test.cl
dbudniko: Done

modules/gapi/test/gapi_gpu_test.cpp
dbudniko: Done

modules/gapi/test/gapi_gpu_test_kernels.cpp
dbudniko: Done. My intention was to separate kernel declarations/implementation in separate files to allow reusage in other applications out of the box. Kind of portability. Now everything is in the test source file gapi_gpu_test.cpp.

modules/gapi/test/gapi_gpu_test_kernels.hpp
dbudniko: Done

modules/gapi/CMakeLists.txt
dbudniko: Done

dmatveev: Cool

modules/gapi/test/gapi_gpu_test.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include "opencv2/gapi/gpu/ggpukernel.hpp"
+#include "opencl_kernels_test_gapi.hpp"
+
+
+namespace cv
+{
+
+#ifdef HAVE_OPENCL
+
+    static void reference_symm7x7_CPU(const cv::Mat& in, const cv::Mat& kernel_coeff, int shift, cv::Mat &out)
```
dmatveev: Note: OpenCV's convention is that _output object_ follows all _input objects_, then followed by _parameters_.

So this prototype would look like:
```diff
--- reference_symm7x7_CPU(const cv::Mat& in, const cv::Mat& kernel_coeff, int shift, cv::Mat &out)
+++ reference_symm7x7_CPU(const cv::Mat& in, cv::Mat &out, const cv::Mat& kernel_coeff, int shift)
```

dmatveev: It is not a critical issue though, just for a general alignment

modules/gapi/test/gapi_gpu_test.cpp
```diff
@@ -0,0 +1,207 @@
...
+            c_float[9], c_float[8], c_float[7], c_float[6], c_float[7], c_float[8], c_float[9],
+            c_float[8], c_float[5], c_float[4], c_float[3], c_float[4], c_float[5], c_float[8],
+            c_float[7], c_float[4], c_float[2], c_float[1], c_float[2], c_float[4], c_float[7],
+            c_float[6], c_float[3], c_float[1], c_float[0], c_float[1], c_float[3], c_float[6],
+            c_float[7], c_float[4], c_float[2], c_float[1], c_float[2], c_float[4], c_float[7],
+            c_float[8], c_float[5], c_float[4], c_float[3], c_float[4], c_float[5], c_float[8],
+            c_float[9], c_float[8], c_float[7], c_float[6], c_float[7], c_float[8], c_float[9]
+        };
+
+        cv::Mat kernel = cv::Mat(7, 7, CV_32FC1);
```
dmatveev: `cv::Mat` can be initialized directly from that array, given `stride = 7*sizeof(float)`

PR #13133, opened 2018-11-12
GAPI (fluid): Sobel 3x3 optimization: CV_SIMD dynamic dispatching

elatkin:
```
force_builders=Linux AVX2
```


elatkin:
This PR is addendum for https://github.com/opencv/opencv/pull/13070

modules/gapi/src/backends/fluid/gfluidimgproc_func.dispatch.cpp
```diff
@@ -0,0 +1,57 @@
...
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2018 Intel Corporation
+
+#if !defined(GAPI_STANDALONE)
+
+#include "gfluidimgproc_func.hpp"
+#include "gfluidimgproc_func.simd.hpp"
+#include "gfluidimgproc_func.simd_declarations.hpp"
```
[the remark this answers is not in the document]
elatkin: Yes, this trick seems to fix the CV_SIMD dynamic dispatching. Please see the next commit.

elatkin:
@alalek  
Could you please merge this PR?

modules/gapi/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ set(gapi_srcs
     src/backends/fluid/gfluidbuffer.cpp
     src/backends/fluid/gfluidbackend.cpp
     src/backends/fluid/gfluidimgproc.cpp
-    src/backends/fluid/gfluidimgproc_func.cpp
```
alalek: Looks like this file is not removed from sources: src/backends/fluid/gfluidimgproc_func.cpp

alalek:
Perhaps need to fix G-API "standalone" build mode after this patch.

modules/gapi/CMakeLists.txt
elatkin: Yes, I forgot removing gfluidimgproc_func.cpp, thanks for finding this problem!

elatkin: Removing this needless file with my next PR #13157

elatkin:
> Perhaps need to fix G-API "standalone" build mode after this patch.  

I have created the new issue to track this: 
#13158 G-API: establish regular testing of "standalone" build

PR #13151, opened 2018-11-13
videoio: drop VideoCaptureModes enum in favour of fourcc

### LINKED ISSUE #5485: videoio: enum CAP_MODE_* should be removed

paroj:
fixes  #5485

### DIFF
--- a/modules/videoio/include/opencv2/videoio.hpp
+++ b/modules/videoio/include/opencv2/videoio.hpp
@@ -178,17 +178,6 @@ enum VideoCaptureProperties {
 #endif
      };
 
-
-/** @brief Generic camera output modes identifier.
-@note Currently, these are supported through the libv4l backend only.
-*/
-enum VideoCaptureModes {
-       CAP_MODE_BGR  = 0, //!< BGR24 (default)
-       CAP_MODE_RGB  = 1, //!< RGB24
-       CAP_MODE_GRAY = 2, //!< Y8
-       CAP_MODE_YUYV = 3  //!< YUYV
-     };
-
 /** @brief %VideoWriter generic properties identifier.
  @sa VideoWriter::get(), VideoWriter::set()
 */
--- a/modules/videoio/include/opencv2/videoio/videoio_c.h
+++ b/modules/videoio/include/opencv2/videoio/videoio_c.h
@@ -452,16 +452,6 @@ enum
     CV_CAP_INTELPERC_GENERATORS_MASK = CV_CAP_INTELPERC_DEPTH_GENERATOR + CV_CAP_INTELPERC_IMAGE_GENERATOR
 };
 
-// Generic camera output modes.
-// Currently, these are supported through the libv4l interface only.
-enum
-{
-    CV_CAP_MODE_BGR  = 0, // BGR24 (default)
-    CV_CAP_MODE_RGB  = 1, // RGB24
-    CV_CAP_MODE_GRAY = 2, // Y8
-    CV_CAP_MODE_YUYV = 3  // YUYV
-};
-
 enum
 {
     // Data given from depth generator.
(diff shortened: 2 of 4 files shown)


alalek:
Please note this comment from [here](https://github.com/opencv/opencv/pull/12893#discussion_r230111555):

---

Actually there are two "processing" formats:
backend SDK -> (1) -> OpenCV -> (2) -> Application

Related properties:
- `CAP_PROP_FOURCC` usually works with (1). V4L stores this value in `capture->palette`
- `CAP_PROP_FORMAT` works with (2) "Format of the %Mat objects returned by VideoCapture::retrieve()"
- `CV_CAP_PROP_CONVERT_RGB` is early version of `CAP_PROP_FORMAT`. Some backends support only one property and don't support another, some of them support both.

`(1)` is capture/camera "native" format. Like MJPEG from here:
```
$ v4l2-ctl --list-formats
ioctl: VIDIOC_ENUM_FMT
	Index       : 0
	Type        : Video Capture
	Pixel Format: 'YUYV'
	Name        : YUYV 4:2:2

	Index       : 1
	Type        : Video Capture
	Pixel Format: 'MJPG' (compressed)
	Name        : Motion-JPEG
```
`cv::VideoCapture` is able to take this MJPEG stream and then converts it into BGR frames (also it is able to pass it "as is" via `CV_8UC1` buffer).

---

paroj:
yes, I am aware of that. However for (2) there is also `cvtColor`. So the context here is "just" optimization as in increasing throughput.

There I argue that e.g. `CAP_MODE_GRAY` is the wrong approach. Instead the interface should be more low-level to properly expose the hardware details.
The reason is that you can not consider (1) and (2) separate, but (2) also affects (1).

For instance in the MSMF backend: `CV_CAP_MODE_YUYV` > `MFVideoFormat_YUY2`. But for most cameras `MFVideoFormat_MJPG` (+decoding) would increase the frame-rate as the USB2 throughput is the bottleneck.

My point here is that the VideoCapture cannot know these details as they are hardware-depended and we should expose that to the user (as in allow (1) == (2)).

Therefore I consider `CAP_PROP_FORMAT` an inferior approach to `CV_CAP_PROP_CONVERT_RGB`. If anything I would replace it by something like: https://github.com/opencv/opencv/issues/5485#issuecomment-226723220

However it is not there yet, but maybe we can add it during the 4.0 cycle - but starting with VideoCaptureModes means that we have to keep it.

PR #13152, opened 2018-11-13
doc(tutorial_py_fourier_transform): Correct errors in tutorial

### LINKED ISSUE #13139: Bug in "Fourier Transform" Python tutorial
This is the page:
https://docs.opencv.org/4.0.0-beta/de/dbc/tutorial_py_fourier_transform.html

There are two bugs in consecutive lines in the second code block:

    crow,ccol = rows/2 , cols/2
    fshift[crow-30:crow+30, ccol-30:ccol+30] = 0

For Python 3, the first row should use `//` for integer division.

The second row does not create a symmetric filter. The right limit is not included in the range, therefore it should be:

    fshift[crow-30:crow+31, ccol-30:ccol+31] = 0

This creates a perfectly symmetric filter, meaning that the inverse transform will be real-valued, as it should, and the following line:

    img_back = np.abs(img_back)

can be replaced with:

    img_back = np.real(img_back)

which doesn't make negative filter results positive, and is faster to compute.

Note that, whenever the inverse transform yields complex values, then there is an error in the Fourier-domain processing (well, mostly, some processing is intended to yield complex values).

[This bug report resulted from [a question on Stack Overflow](https://stackoverflow.com/a/53267136/7328782).]

ssnover95:
for integer division and filter symmetry.

resolves #13139 

### This pullrequest changes
Changes the second snippet of example code lines 2, 3, and 6.
- Changes the calculation of `crow` and `ccol` to use integer division since they're used as indices for slicing.
- Changes the slicing on line 3 to have symmetry.
- Change call from `np.abs()` to `np.real()` as a symmetric filter from the previous change will cause a real value to be calculated in the inverse transform.

```
force_builders=linux,docs
```

### DIFF
--- a/doc/py_tutorials/py_imgproc/py_transforms/py_fourier_transform/py_fourier_transform.markdown
+++ b/doc/py_tutorials/py_imgproc/py_transforms/py_fourier_transform/py_fourier_transform.markdown
@@ -79,11 +79,11 @@ using **np.ifft2()** function. The result, again, will be a complex number. You 
 absolute value.
 @code{.py}
 rows, cols = img.shape
-crow,ccol = rows/2 , cols/2
-fshift[crow-30:crow+30, ccol-30:ccol+30] = 0
+crow,ccol = rows//2 , cols//2
+fshift[crow-30:crow+31, ccol-30:ccol+31] = 0
 f_ishift = np.fft.ifftshift(fshift)
 img_back = np.fft.ifft2(f_ishift)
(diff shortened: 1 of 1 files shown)


alalek:
Thank you!

So, please:

ssnover95:
Thanks for the complete instructions! I have never contributed to a project which maintains multiple major release branches so I wasn't sure what my base branch should be. I've updated my branch and PR with the changes.

PR #13168, opened 2018-11-14
dnn: reduce disabled warnings

alalek:
```
#test_modules=dnn
#docker_image:Custom=fedora:29
#buildworker:Custom=linux-4
buildworker:Custom=linux-1
```


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -223,7 +225,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
          * @overload
          * @deprecated Use Layer::finalize(InputArrayOfArrays, OutputArrayOfArrays) instead
          */
-        CV_DEPRECATED void finalize(const std::vector<Mat> &inputs, CV_OUT std::vector<Mat> &outputs);
+        CV_DEPRECATED_EXTERNAL
+        void finalize(const std::vector<Mat> &inputs, CV_OUT std::vector<Mat> &outputs);
```
dkurt: There are two more functions:
```cpp
        CV_DEPRECATED std::vector<Mat> finalize(const std::vector<Mat> &inputs);

        /** @brief Allocates layer and computes output.
         *  @deprecated This method will be removed in the future release.
         */
        CV_DEPRECATED CV_WRAP void run(const std::vector<Mat> &inputs, CV_OUT std::vector<Mat> &outputs,
                                       CV_IN_OUT std::vector<Mat> &internals);
```

PR #13174, opened 2018-11-15
GAPI (fluid): RGB to YUV optimization

elatkin:
force_builders=Linux AVX2


elatkin:
@savuor  
Could you please review?

elatkin:
Performance results on my desktop, i5-6600  

RGB to YUV (1 thread)  

| Image  | OpenCV | GAPI(new) | GAPI(old) |  
|----------|-------------|---------------|--------------|
| 1080p  | 7.57 ms  |  0.91  ms   | 26.7 ms    |  
|   720p  | 3.4   ms  | 0.415 ms   | 11.9 ms    |  
|   480p  | 1.14 ms  | 0.147 ms   | 7.11 ms    |  

YUV to RGB (1 thread)  

| Image   | OpenCV | GAPI(new) | GAPI(old) |  
|-----------|-------------|---------------|--------------|
| 1080p   | 7.45 ms  | 0.876 ms   | 26.4 ms    |  
|   720p   | 3.37 ms  | 0.4     ms   | 11.8 ms    |  
|   480p   | 1.11 ms  | 0.143 ms   | 3.96 ms    |  

So GAPI appears 8x times faster than OpenCV  

Speed-up with GAPI optimization is 30x times

elatkin:
> How is this compared to other libraries? For example: libyuv?
> 
> Also (unrelated to this PR): is there a reason why OpenCV does not attempt to use GAPI implementations? For example, if you are running cvtColor.

Good points. Thanks for both these questions!  

Wide universal intrinsics enable to cover all architectures with single reasonably compact piece of code.  

Yes, we could think of reusing such code for optimizing cvtColor. (@savuor could probably comment?)

elatkin:
@alalek  
Could you please merge this PR?

PR #13175, opened 2018-11-15
Fixed use of model.load instances in letter_recog.py

paddy74:
The load() function returns a new object, and as such does not use the one it is called on. This commit updates the uses of model.load in this program so it will work as intended and not throw an error.

### This pullrequest changes

Fixed use of model.load instances in letter_recog.py

### DIFF
--- a/samples/python/letter_recog.py
+++ b/samples/python/letter_recog.py
@@ -41,7 +41,7 @@ class LetterStatModel(object):
     train_ratio = 0.5
 
     def load(self, fn):
-        self.model.load(fn)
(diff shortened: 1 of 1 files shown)


samples/python/letter_recog.py
```diff
@@ -169,7 +169,7 @@ def predict(self, samples):
     if '--load' in args:
         fn = args['--load']
         print('loading model from %s ...' % fn)
-        model.load(fn)
+        model = model.load(fn)
```
alalek: This change is not correct. Reverted.

PR #13184, opened 2018-11-16
highgui: Qt - restore convertscale semantics

paroj:
broken in 11eafca3e2a4cbc62f1309d25db0ea3ed9a6ea8e

### DIFF
--- a/modules/highgui/src/window_QT.cpp
+++ b/modules/highgui/src/window_QT.cpp
@@ -2557,7 +2557,14 @@ void DefaultViewPort::updateImage(const CvArr* arr)
     nbChannelOriginImage = cvGetElemType(mat);
     CV_Assert(origin == 0);
     cv::Mat src = cv::cvarrToMat(mat), dst = cv::cvarrToMat(image2Draw_mat);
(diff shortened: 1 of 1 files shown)


modules/highgui/src/window_QT.cpp
```diff
@@ -2557,7 +2557,14 @@ void DefaultViewPort::updateImage(const CvArr* arr)
     nbChannelOriginImage = cvGetElemType(mat);
     CV_Assert(origin == 0);
     cv::Mat src = cv::cvarrToMat(mat), dst = cv::cvarrToMat(image2Draw_mat);
-    cv::cvtColor(src, dst, cv::COLOR_BGR2RGB, dst.channels());
+
+    cv::Mat tmp;
+    int src_depth = src.depth();
+    double scale = src_depth <= CV_8S ? 1 : src_depth <= CV_32S ? 1./256 : 255;
+    double shift = src_depth == CV_8S || src_depth == CV_16S ? 128 : 0;
+    cv::convertScaleAbs(src, tmp, scale, shift);
```
dan-masek: Shouldn't this just be called when scaling/shifting/conversion actually needs to be done? AFAICT in such cases the function still processes the whole image, which seems wasteful when unnecessary.

PR #13193, opened 2018-11-17
core: move dims() check to the top in copyMakeBorder()

### LINKED ISSUE #13180: cv2.copyMakeBorder segmentation fault
- operating system:
Ubuntu 16.04 (64-bit) in Docker 18.09.0.

- architecture (e.g. x86):
Intel® Core™ i5-7500 CPU @ 3.40GHz × 4.

- opencv-python version:
opencv-python==3.4.3.18

##### Detailed description

- Expected behavior:
`ValueError: passed image has wrong shape. Expected to have 3 dimensions, has 4.`
(or any other type of error with the similar description)

- Actual behavior:
`31807 segmentation fault (core dumped)`

- Description:
This issue was first opened in [opencv-python repository](https://github.com/skvark/opencv-python/issues/143). Since input data checks neither performed  on Python-level  nor on C-level usage of 4-D and greater tensors causes uninformative errors.

##### Steps to reproduce

```python
import cv2
import numpy as np

# Example image. 
# May also be an image loaded with opencv-python with np.expand_dims(img, 0) or similar.
img = np.zeros((1, 256, 256, 3), dtype=np.uint8)
cv2.copyMakeBorder(img, 1, 1, 1, 1, cv2.BORDER_CONSTANT, value=[0, 0, 0])
```

berak:
resolves #13180

since none of the implemented branches can handle more than 2 dimensions,
move the dims() check from the CV_OCL_RUN up, so it gets checked for all of them

### DIFF
--- a/modules/core/src/copy.cpp
+++ b/modules/core/src/copy.cpp
@@ -1186,9 +1186,9 @@ void cv::copyMakeBorder( InputArray _src, OutputArray _dst, int top, int bottom,
 {
     CV_INSTRUMENT_REGION();
 
-    CV_Assert( top >= 0 && bottom >= 0 && left >= 0 && right >= 0 );
+    CV_Assert( top >= 0 && bottom >= 0 && left >= 0 && right >= 0 && _src.dims() <= 2);
 
-    CV_OCL_RUN(_dst.isUMat() && _src.dims() <= 2,
+    CV_OCL_RUN(_dst.isUMat(),
                ocl_copyMakeBorder(_src, _dst, top, bottom, left, right, borderType, value))
 
     Mat src = _src.getMat();


modules/core/src/copy.cpp
```diff
@@ -1186,9 +1186,9 @@ void cv::copyMakeBorder( InputArray _src, OutputArray _dst, int top, int bottom,
 {
     CV_INSTRUMENT_REGION();
 
-    CV_Assert( top >= 0 && bottom >= 0 && left >= 0 && right >= 0 );
+    CV_Assert( top >= 0 && bottom >= 0 && left >= 0 && right >= 0 && _src.dims() <= 2);
```
alalek: BTW,

It is better to avoid complex assertion conditions, because they doesn't help debugging problems.
Consider using `CV_Check*()` macros instead.

PR #13195, opened 2018-11-17
move legacy C-API constants into separate files

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/1911

relates #13060


alalek:
:+1:

PR #13218, opened 2018-11-20
G-API: Fix STANDALONE build

dmatveev:
Recent merge of GPU plugin in G-API has broken it, this PR brings support back


dmatveev:
@dbudniko please review.

@alalek this is an urgent thing since it prevents bumping G-API in dldt to latest version (which is needed already)

dmatveev:
@alalek it's green, it's green!

dmatveev:
Thanks!

PR #13267, opened 2018-11-24
Python wrapper for detail

LaurentBerger:
Rewrite [stitching_detailed.cpp](https://github.com/opencv/opencv/blob/master/samples/cpp/stitching_detailed.cpp) in python


LaurentBerger:
@alalek Something is wrong in my last commit but I don't understand Opencl test failed for stitching
In commit (https://github.com/opencv/opencv/pull/13267/commits/d8416e36953ef8e093153fb0ca94c948a6d6fd14) and i following commit something is wrong : 
[diff.txt](https://github.com/opencv/opencv/files/2625732/diff.txt)

I add four methods (getMatGains, setMatGains, setUpdateGain getUpdateGain ) and a protected member updateGain

alalek:
Try to debug locally (built with opencv_contrib - SURF is there).

alalek:
GDB:
```
(gdb) frame 2
#2  0x00007ffff7aff456 in cv::detail::BlocksGainCompensator::apply (this=0x7a25c0, index=0, _image=...) at /home/alalek/projects/opencv/dev/modules/stitching/src/exposure_compensate.cpp:263
263	    if (gain_maps_[index].size() == _image.size())
(gdb) p gain_maps_
$1 = std::vector of length 0, capacity 0
```

LaurentBerger:
ABI the retrun ?

modules/stitching/include/opencv2/stitching/warpers.hpp
```diff
@@ -52,17 +53,86 @@ namespace cv {
...
     virtual Ptr<detail::RotationWarper> create(float scale) const = 0;
 };
 
+
+class CV_EXPORTS_W PyRotationWarper
+{
+    Ptr<detail::RotationWarper> rw;
+
+public:
+    CV_WRAP PyRotationWarper(String type, float scale);
```
alalek: I believe this should be moved out from public headers.

Could you create file similar to "shadow_umat.hpp"?

Do you have any thoughts about Java?

alalek:
My mistake - shadow doesn't work here, because there is not C++ class (like UMat).
Place this code as regular bindings only code in "misc" folder.

modules/python/src2/cv2.cpp
```diff
@@ -209,6 +210,9 @@ typedef std::vector<KeyPoint> vector_KeyPoint;
 typedef std::vector<Mat> vector_Mat;
 typedef std::vector<std::vector<Mat> > vector_vector_Mat;
 typedef std::vector<UMat> vector_UMat;
+typedef std::vector<detail::ImageFeatures> vector_ImageFeatures;
+typedef std::vector<detail::MatchesInfo> vector_MatchesInfo;
+typedef std::vector<detail::CameraParams> vector_CameraParams;
```
alalek: Only types from "core" module are allowed in cv2.cpp

Other types should be defined via "misc/python" headers.

modules/stitching/include/opencv2/stitching/detail/autocalib.hpp
```diff
@@ -76,7 +76,7 @@ void CV_EXPORTS estimateFocal(const std::vector<ImageFeatures> &features,
                               const std::vector<MatchesInfo> &pairwise_matches,
                               std::vector<double> &focals);
 
-bool CV_EXPORTS calibrateRotatingCamera(const std::vector<Mat> &Hs, Mat &K);
+bool CV_EXPORTS_W calibrateRotatingCamera(const std::vector<Mat> &Hs, Mat &K);
```
alalek: Use `CV_OUT` for output parameters.

LaurentBerger: OK

LaurentBerger:
Previous commit was wrong because PyRotationWraper was not exported in python

LaurentBerger:
@alalek What's  new?

modules/python/test/test_detail.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
```
alalek: Thank you!
Please move this test code into `test_stitching.py`.

PR #13289, opened 2018-11-27
Make lk_track window resizable

daniel-s-ingram:
I've always liked the ability to fullscreen/resize my OpenCV windows (it helps a lot with teaching), so I would be happy to go through and make all the sample windows resizable if that's something others would be interested in.


alalek:
I believe proper solution is adding support for "maximized" window's state. And allow this mode for `WINDOW_AUTOSIZE` (current default) too.

What OS / GUI backend are you using?

daniel-s-ingram:
I'm on Ubuntu 16.04.05 with GTK+ 3.18.9 as my OpenCV GUI backend. Can you clarify your comment? Do you mean that resizable windows should be the default and that something in the source should be changed to reflect this or that there's a more proper way to do what I'm doing without changing anything in the source?

vpisarev:
it works fine on macOS too; it's very minor but seemingly useful change; let's merge it :+1:

alalek:
Fixing just one sample doesn't look as consistent patch anyway.
Need to change OpenCV behavior instead.

asmorkalov:
Replaced by more generic feature request https://github.com/opencv/opencv/issues/15772

PR #13290, opened 2018-11-27
GAPI (fluid): Filter 2D optimization

elatkin:


modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
```diff
@@ -76,6 +77,26 @@ RUN_SEPFILTER3X3_IMPL( float,  float)
...
+void run_filter2d_3x3_impl(DST out[], const SRC *in[], int width, int chan, \
+                           const float kernel[], float scale, float delta);
+
+RUN_FILTER2D_3X3_IMPL(uchar , uchar )
+RUN_FILTER2D_3X3_IMPL(ushort, ushort)
+RUN_FILTER2D_3X3_IMPL( short,  short)
+RUN_FILTER2D_3X3_IMPL( float, uchar )
+RUN_FILTER2D_3X3_IMPL( float, ushort)
+RUN_FILTER2D_3X3_IMPL( float,  short)
+RUN_FILTER2D_3X3_IMPL( float,  float)
```
terfendail: Isn't it already defined in gfluidimgproc_func.hpp?

elatkin: *.simd.hpp defines run_filter2d_3x3_impl symbols in HW-specific name spaces  
(so we have cv::gapi::fluid::opt_SSE4_1::run_filter2d_3x3_impl for SSE 4.1, etc)

modules/gapi/src/backends/fluid/gfluidimgproc.cpp
```diff
@@ -1097,7 +1103,8 @@ GAPI_FLUID_KERNEL(GFluidFilter2D, cv::gapi::imgproc::GFilter2D, true)
 
         int k_rows = kernel.rows;
         int k_cols = kernel.cols;
-        const float *k = scratch.OutLine<float>(); // copy of kernel.data
+
+        float *k = scratch.OutLine<float>(); // copy of kernel.data
```
terfendail: Should const actually be removed here?

elatkin: Good point, thanks! I would restore `const` qualifier here.

elatkin:
@alalek  
Could you please merge?

PR #13332, opened 2018-11-29
DNN backends registry

mshabunin:
### This pullrequest changes

adds basic registry for DNN backends and targets

```
buildworker:Custom=linux-2
test_opencl:Custom=OFF
```

### DIFF
--- a/modules/dnn/test/test_backends.cpp
+++ b/modules/dnn/test/test_backends.cpp
@@ -292,6 +292,6 @@ TEST_P(DNNTestNetwork, FastNeuralStyle_eccv16)
     processNet("dnn/fast_neural_style_eccv16_starry_night.t7", "", inp, "", "", l1, lInf);
 }
 
-INSTANTIATE_TEST_CASE_P(/*nothing*/, DNNTestNetwork, dnnBackendsAndTargets(true, true, false));
+INSTANTIATE_TEST_CASE_P(/*nothing*/, DNNTestNetwork, dnnBackendsAndTargets());
 
 }} // namespace
--- a/modules/dnn/test/test_caffe_importer.cpp
+++ b/modules/dnn/test/test_caffe_importer.cpp
@@ -300,10 +300,11 @@ INSTANTIATE_TEST_CASE_P(/**/, Reproducibility_ResNet50,
 typedef testing::TestWithParam<Target> Reproducibility_SqueezeNet_v1_1;
 TEST_P(Reproducibility_SqueezeNet_v1_1, Accuracy)
 {
+    int targetId = GetParam();
+    if(targetId == DNN_TARGET_OPENCL_FP16)
+        throw SkipTestException("This test does not support FP16");
     Net net = readNetFromCaffe(findDataFile("dnn/squeezenet_v1.1.prototxt", false),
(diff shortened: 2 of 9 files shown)


modules/dnn/test/test_caffe_importer.cpp
```diff
@@ -324,7 +324,8 @@ TEST_P(Reproducibility_SqueezeNet_v1_1, Accuracy)
     Mat ref = blobFromNPY(_tf("squeezenet_v1.1_prob.npy"));
     normAssert(ref, out);
 }
-INSTANTIATE_TEST_CASE_P(/**/, Reproducibility_SqueezeNet_v1_1, availableDnnTargets());
+INSTANTIATE_TEST_CASE_P(/**/, Reproducibility_SqueezeNet_v1_1,
+    testing::ValuesIn(getAvailableTargets(DNN_BACKEND_OPENCV)));
```
dkurt: It enables FP16 tests so we can use common thresholds:

```cpp
double l1 = (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD) ? 4e-3 : 1e-5;
double lInf = (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD) ? 2e-2 : 1e-4;
```
It should fix the following tests:
```
Reproducibility_SqueezeNet_v1_1.Accuracy/1, where GetParam() = OCL_FP16
Reproducibility_GoogLeNet.Batching/1, where GetParam() = OCL_FP16
```
However the following ones doesn't pass:
```
Reproducibility_GoogLeNet.IntermediateBlobs/1, where GetParam() = OCL_FP16
Reproducibility_GoogLeNet.SeveralCalls/1, where GetParam() = OCL_FP16
```
I think it's OK if we skip them for now for OCL_FP16 target.

modules/dnn/test/test_backends.cpp
```diff
@@ -292,6 +292,6 @@ TEST_P(DNNTestNetwork, FastNeuralStyle_eccv16)
     processNet("dnn/fast_neural_style_eccv16_starry_night.t7", "", inp, "", "", l1, lInf);
 }
 
-INSTANTIATE_TEST_CASE_P(/*nothing*/, DNNTestNetwork, dnnBackendsAndTargets(true, true, false));
+INSTANTIATE_TEST_CASE_P(/*nothing*/, DNNTestNetwork, dnnBackendsAndTargets());
```
dkurt: I think we need to keep `false` for `withCpuOCV`

modules/dnn/test/test_common.hpp
```diff
@@ -237,49 +213,31 @@ namespace opencv_test {
 using namespace cv::dnn;
 
 static inline
-testing::internal::ParamGenerator<tuple<Backend, Target> > dnnBackendsAndTargets(
+testing::internal::ParamGenerator< tuple<Backend, Target> > dnnBackendsAndTargets(
         bool withInferenceEngine = true,
         bool withHalide = false,
         bool withCpuOCV = true
```
dkurt: `withCpuOCV` is ignored here. Some tests use default CPU implementation as reference so we don't compare it with itself.

alalek:
I will prepare fixup commit for that.

PR #13357, opened 2018-12-04
bugfix:Point reutrned by Rect::br()  is outside rect

sureiron:


alalek:
Refer to documentation.

PR #13360, opened 2018-12-04
Moved shadow UMat decls to separate file

### LINKED ISSUE #13248: Python bindings build fails when source directory named "opencv2"
- OpenCV => master
- Operating System / Platform => Ubuntu 18
- Compiler => default

##### Detailed description

When source directory is named `opecnv2` python bindings can not be built:
```
FAILED: <compiling> /work/opencv2/modules/python/src2/cv2.cpp
In file included from /work/opencv2/modules/python/src2/cv2.cpp:109:0:
modules/python_bindings_generator/pyopencv_generated_include.h:14:10: fatal error: opencv2/modules/core/misc/python/shadow_umat.hpp: No such file or directory
 #include "opencv2/modules/core/misc/python/shadow_umat.hpp"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
Directory names `opencv` and `opencv3` works well.

##### Steps to reproduce

```
git clone https://github.com/opencv/opencv opencv2
git -C opencv2 checkout master
mkdir build && pushd build
cmake -GNinja ../opencv2
ninja
```

mshabunin:
resolves #13248

### This pullrequest changes

* moved "shadow UMat decls" to separate Python file to avoid parsing it with hdr_parser
* removed dependency gen_java.py -> gen2.py


vpisarev:
@mshabunin, if I understand correctly, the patch gets rid of "stub" UMat declaration and replaces it with pre-parsed information about the class and the methods. Here we use some internal representation of the parsed API, produced by hdr_parser.py. I think, such internal representation is quite difficult to maintain. If we change the output representation of extend/change UMat API, we need to update this part as well. The extra header is much easier to support. In other words, I would not apply the patch in its current form

PR #13369, opened 2018-12-05
dnn: fixup available backends/targets

alalek:
relates #13332

```
buildworker:Custom=linux-2
test_opencl:Custom=OFF

docker_image:Linux AVX2=ubuntu-openvino:16.04
test_opencl:Linux AVX2=OFF
```


alalek:
:+1:

PR #13387, opened 2018-12-07
Minor fixes in IE backend tests

dkurt:
### This pullrequest changes

```
buildworker:Custom=linux-2
test_opencl:Custom=OFF
```

### DIFF
--- a/modules/dnn/test/test_layers.cpp
+++ b/modules/dnn/test/test_layers.cpp
@@ -137,7 +137,7 @@ TEST_P(Test_Caffe_layers, Convolution)
 
 TEST_P(Test_Caffe_layers, DeConvolution)
 {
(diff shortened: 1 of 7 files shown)


modules/dnn/test/test_layers.cpp
```diff
@@ -951,12 +956,15 @@ TEST(Layer_Test_Convolution_DLDT, setInput_uint8)
     for (int i = 0; i < 2; ++i)
     {
         Net net = readNet(_tf("layer_convolution.xml"), _tf("layer_convolution.bin"));
+        net.setPreferableTarget(targetId);
         net.setInput(inputs[i]);
```
alalek: setPreferableBackend?

dkurt: It turns automatically to `DNN_BACKEND_INFERENCE_ENGINE` for models in Intermediate Representation (.xml and .bin).

PR #13395, opened 2018-12-08
Add support for reading 4-channel BGRA frames from video

### LINKED ISSUE #13394: RGB frames for video files with Alpha channel support

- OpenCV => 3.4
- Operating System / Platform => Ubuntu
- Compiler => gcc

##### Detailed description

When reading a `.mov` file with alpha channel support, the `cv::VideoCapture` returns a BGR frame instead of BGRA frame. The issue was highlighted in [this SO post](https://stackoverflow.com/q/53552698/3051961). Sample file is also attached.
[sample_mov_file_with_alpha.zip](https://github.com/opencv/opencv/files/2659641/out.zip)

##### Steps to reproduce

```cpp
int main() {
    cv::VideoCapture cap("/path/to/file.mov");

    cv::Mat m;
    while(true) {
        cap.read(m);

        if (m.empty()) {
            break;
        }

        std::cout << m.channels() << std::endl;
    }
}
```

anmoluppal:
resolves #13394

### This pullrequest changes

Instead of assuming that the output frame of all video formats is a 3-channel `cv::Mat`, these changes check for the input video pixel format and chooses the relevant output format. This is a fix to this [issue](https://github.com/opencv/opencv/issues/13394)

### DIFF
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -497,6 +497,8 @@ struct CvCapture_FFMPEG
     double  r2d(AVRational r) const;
     int64_t dts_to_frame_number(int64_t dts);
     double  dts_to_sec(int64_t dts);
+    AVPixelFormat getDstPixelFormat(AVPixelFormat srcPixelFormat) const;
+    int getChannelCount(AVPixelFormat srcPixelFormat) const;
 
     AVFormatContext * ic;
     AVCodec         * avcodec;
@@ -1103,7 +1105,7 @@ bool CvCapture_FFMPEG::retrieveFrame(int, unsigned char** data, int* step, int* 
                 buffer_width, buffer_height,
                 video_st->codec->pix_fmt,
                 buffer_width, buffer_height,
-                AV_PIX_FMT_BGR24,
+                getDstPixelFormat(video_st->codec->pix_fmt),
                 SWS_BICUBIC,
                 NULL, NULL, NULL
                 );
@@ -1113,7 +1115,7 @@ bool CvCapture_FFMPEG::retrieveFrame(int, unsigned char** data, int* step, int* 
 
 #if USE_AV_FRAME_GET_BUFFER
         av_frame_unref(&rgb_picture);
-        rgb_picture.format = AV_PIX_FMT_BGR24;
+        rgb_picture.format = getDstPixelFormat(video_st->codec->pix_fmt);
         rgb_picture.width = buffer_width;
         rgb_picture.height = buffer_height;
         if (0 != av_frame_get_buffer(&rgb_picture, 32))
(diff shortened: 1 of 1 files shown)


vpisarev:
@anmoluppal, thanks a lot for the patch! As @alalek mentioned in the referenced issue, is it possible to retain 4-to-3 channel conversion behaviour by default and then add CAP_PROP_... option (e.g. `CAP_PROP_KEEP_ALPHA`) to enable 4-channel output?

alalek:
There is already such option. To get RGBA frames (in case of RGBA native video stream) we should pass `CAP_PROP_CONVERT_RGB=False`

HarrisonLeece:
For those who are smooth of brain, can I get a little more help on getting this working with my program? 

Do I pass CAP_PROP_CONVERT_RGB=False to the the VideoCapture instantiation as so?
```
VideoCapture cap(input_path, CAP_PROP_CONVERT_RGB=False);
```
or
```
cap.CAP_PROP_CONVERT_RGB=False
```
or
```
cap.get(CAP_PROP_FRAME_COUNT, CAP_PROP_CONVERT_RGB=False);
```
?

Thanks, my c++ wizards.

aCuria:
> For those who are smooth of brain, can I get a little more help on getting this working with my program?

videoCapture_.set(cv::CAP_PROP_CONVERT_RGB, 0);

Logix-Dev:
Correct me if I'm wrong, but this doesn't appear to be working. I'm trying to load a .avi video which I have rendered myself _with transparency_, and the alpha channel is verified to exist via OBS. And yet, if I try to load the video with OpenCV following the (sparse) advise above, and ask for the channel count, I get 3. The video renders with OpenGL just fine, but again, with no alpha channel. (Verified by displaying the alpha channel directly via the shader).

I'm using `OpenCV 4.6`, with `opencv_videoio_ffmpeg460_64.dll`.

The code:
```cpp
cv::VideoCapture m_Capture;
cv::Mat m_Frame;
```

```cpp
m_Capture = cv::VideoCapture(Filename);
m_Capture.set(cv::CAP_PROP_CONVERT_RGB, 0);
    
if (!m_Capture.isOpened())
{
    LogError("Failed to open the video file '" + Filename + "'");
    return false;
}
    
m_Capture >> m_Frame;
LogDebug("Loaded video '" + Filename + "' with " + std::to_string(m_Frame.channels()) + " channels");
```

The frames are being converted using:
```cpp
unsigned char* video_layer::ConvertFrame(cv::Mat& Image)
{
    cv::cvtColor(Image, Image, cv::COLOR_BGRA2RGBA);
    cv::transpose(Image, Image);
    cv::flip(Image, Image, 1);
    cv::flip(Image, Image, 0);
    
    return Image.data;
}
```

The output:
 > Loaded video '(filename).avi' with 3 channels

Any assistance in understanding the full proper usage would be greatly appreciated. And as stated before, the video I'm loading does have an alpha channel, and was rendered in RGBA format.

PR #13420, opened 2018-12-12
Remove MinCore_Downlevel, replace with Shlwapi

ThadHouse:
On windows 7, MinCore_Downlevel does not work correctly. However, the only API used was QISearch, which can be found in Shlwapi.

resolves #12010

This is a fix that should be backported to 3.4 as well. Should that be a separate PR, or done after merging?

PR #13421, opened 2018-12-12
Don't include DirectX when WITH_DIRECTX=OFF is specified

ThadHouse:
Portions of DirectX were removed, but others were not. WITH_DIRECTX=OFF should not include DirectX anywhere.

This mainly affected VideoIO, which included DirectX 11 even with the no directx flag set. 
https://github.com/opencv/opencv/blob/master/modules/videoio/src/cap_msmf.cpp#L84

This is a fix that should be backported to 3.4 as well. Should that be a separate PR, or done after merging?


ThadHouse:
I feel like `WITH_DIRECTX=OFF` should unconditionally mean there should be no DirectX linked anywhere in the library. It is confusing from a user perspective to have 2 flags to do the same things in 2 separate parts of the library.

I feel like a change that big also wouldn't be backported to 3.4, whereas this should be, as 3.3 and before I think didn't have this issue.

ThadHouse:
Per comments on https://github.com/opencv/opencv/pull/13420, I changed the base branch of this to 3.4 as it really is a 3.4 fix that should go in 4.0 as well.

alalek:
replaced by #13436

PR #13422, opened 2018-12-12
Split videoio dependencies search

mshabunin:
### This pullrequest changes

* Split `OpenCVFindLibsVideo.cmake` file into parts: `videoio/cmake/detect_<backend>.cmake`
* Removed corresponding defines from cvconfig.h:
    * Several tests depending on VideoCapture/VideoWriter support now will be skipped in runtime
* All dependencies to videoio module are added via cmake targets `ocv::3rdparty::<tgt>`, `VIDEOIO_LIBRARIES` has been removed 

#### Removed backends and leftovers:
* OpenNI v1 (use OpenNI v2 backend instead)
* QuickTime (OSX)
* Intel PerC (use RealSense backend instead)
* DC1394 v1
* GiGEVision SDK
* GStreamer 0.10 support

#### Platform specific notes:
* Windows: FFmpeg copying has been implemented as a custom target
* Windows: several explicitly linked libraries have been removed (e.g. avifil32 avicap32 winmm msvfw32)
* Apple: `add_apple_compiler_options` has been removed

#### Package search changes:
* Aravis SDK:
    * only version 0.6 is supported
    * detected with pkg-config
    * detected by `ARAVIS_ROOT` option or environment variable
    * **Windows verison have not been checked**
* AVFoundation / iOS:
    * suppressed deprecation warnings in source files (.mm)
    * iOS videoio support can be disabled by `WITH_CAP_IOS` option
* DC1394:
    * detected with pkg-config
    * detected by `DC1394_ROOT` option or environment variable
* DShow - similar detection
* FFmpeg - similar detection
* GPhoto - similar detection
* GStreamer:
    * detected with pkg-config
    * Windows version detection uses `GSTREAMER_DIR` cmake option or `GSTREAMER_ROOT`, `GSTREAMER_1_0_ROOT_X86_64`, `GSTREAMER_ROOT_X86_64`, `GSTREAMER_1_0_ROOT_X86`, `GSTREAMER_ROOT_X86` environment variables
* MediaSDK:
    * fixed detection of opensource version
* Media Foundation:
    * added `WITH_MSMF_DXVA` option to control HW acceleration (ported #13421)
    * added shlwapi link (ported #13420)
* OpenNI 2:
    * detected by `OPENNI2_DIR` cmake option and `OPENNI2_INCLUDE64`, `OPENNI2_LIB64`, `OPENNI2_REDIST64`, `OPENNI2_INCLUDE`, `OPENNI2_LIB`, `OPENNI2_REDIST` environment variables
    * detected in default system folders
    * **Windows verison have not been checked**
* PvAPI (is deprecated)
    * detected by `PVAPI_ROOT` option or environment variable
    * **Windows verison have not been checked**
* RealSense SDK:
    * uses find_package(realsense2)
    * detected by `LIBREALSENSE_INCLUDE`, `LIBREALSENSE_LIB` optons or environment variables
    * **Windows verison have not been checked**
* V4L - similar detection
* XIMEA:
    * detected by `XIMEA_ROOT` option or environment variable or in `/opt/XIMEA` directory
    * **has not been checked on OSX and Windows**
* XINE - similar detection

```
force_builders=Linux x64,Linux OpenCL,Linux AVX2,Win64,Win64 OpenCL,Mac,Mac OpenCL,Android armeabi-v7a,Linux x64 Debug,Docs,iOS,Linux32,Win32,Win64 MSVS2017,ARMv8,Android pack

#build_shared:Linux OpenCL=OFF
#build_world:Linux OpenCL=ON

#build_shared:Linux AVX2=OFF

#build_shared:Win64 OpenCL=OFF
#build_examples:Win64 OpenCL=OFF

#build_shared:Mac OpenCL=OFF

#build_world:Win32=ON

build_shared:Win64 MSVS2017=OFF
build_examples:Win64 MSVS2017=OFF
build_world:Win64 MSVS2017=ON
```


mshabunin:
Verified some basic dependency installation schemes on Linux (Ubuntu 18): https://github.com/mshabunin/videoio-build-verification

modules/videoio/cmake/init.cmake
```diff
@@ -0,0 +1,35 @@
+macro(add_backend backend_id cond_var)
+  if(${cond_var})
+    include("${CMAKE_CURRENT_LIST_DIR}/detect_${backend_id}.cmake")
+  endif()
+endmacro()
+
+macro(add_3p_target name inc link def)
+  add_library(ocv.3rdparty.${name} INTERFACE)
```
alalek: 1.
Please move this macro into common `OpenCVUtils.cmake` (this logic can be re-used for imgcodecs - if `BUILD_xxx=OFF`).

2.
Lets change name:
- add `ocv_` prefix
- it should be used for "external" targets
- something like this: `ocv_add_external_dependency_target()`

3.
In `BUILD_SHARED_LIBS=ON` configuration we should not export extra targets in OpenCVModules.cmake, because they are not used. Lets mark these targets for this configuration as `IMPORTED`.

modules/videoio/cmake/detect_xine.cmake
```diff
@@ -0,0 +1,12 @@
...
+  if(XINE_FOUND)
+    set(HAVE_XINE TRUE)
+  endif()
+endif()
+
+if(HAVE_XINE)
+  add_3p_target(xine "${XINE_INCLUDE_DIRS}" "${XINE_LIBRARIES}" "HAVE_XINE")
+endif()
+
+set(HAVE_XINE ${HAVE_XINE} PARENT_SCOPE)
```
alalek: > PARENT_SCOPE

It is better to remove that.

Unnecessary "function" scope should be removed (perhaps it is already done [here](https://github.com/opencv/opencv/pull/12541/files#diff-a8bb897ccface5bd5e7eff5548b73f77R299)).

modules/videoio/cmake/init.cmake
```diff
@@ -0,0 +1,35 @@
+macro(add_backend backend_id cond_var)
+  if(${cond_var})
+    include("${CMAKE_CURRENT_LIST_DIR}/detect_${backend_id}.cmake")
+  endif()
```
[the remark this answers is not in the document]
alalek: We should not change existed policy about `WITH_` / `HAVE_` variables.

modules/videoio/cmake/detect_xine.cmake
alalek: It is not about "status" text only.
It is about the similar process for all external dependencies, including zlib (which is used for opencv_core directly, and for libtiff dependency) and imgcodecs dependencies.

modules/videoio/cmake/init.cmake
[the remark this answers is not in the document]
mshabunin: I think the difference is not too big and the proposed policy is less error prone. For example, building two configurations in the same build directory:
* build OpenCV with `WITH_` disabled by default and `HAVE_` enabled via command-line - feature is turned ON
* build OpenCV with `WITH_` disabled by default and none additional variables set - feature is still turned ON, because `HAVE_` is stored in cache and overrides `WITH_` option.

In the method which I propose, `HAVE_` is internal variable and will not be stored in cache, thus `WITH_` will actually disable or enable corresponding feature. If you pass `HAVE_` via command line in one build and do not pass it in consecutive builds, the feature will still be disabled because target-creation code will not be executed.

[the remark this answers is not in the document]
alalek: I mean this:
```
if(WITH_${cond_var} OR HAVE_${cond_var})
    include(...)
endif()
```

and in `detect_xxx.cmake` (already done):

```
if(NOT HAVE_xxx)
  ... some auto-detection which work in most cases ...
endif()

... next auto detection attempt ...

if(HAVE_xxx)
  ... register dependencies with support of command-line vars: ...
  ... INCLUDE_DIRS, LIBRARIES, etc ...
endif()
```

modules/videoio/cmake/init.cmake
mshabunin: I don't want to extract it to OpenCVUtils.cmake yet because it has very basic interface which will be changed later.

cmake/OpenCVGenConfig.cmake
```diff
@@ -41,7 +41,7 @@ foreach(m ${OPENCV_MODULES_BUILD})
   endif()
 endforeach()
 
-export(TARGETS ${OpenCVModules_TARGETS} FILE "${CMAKE_BINARY_DIR}/OpenCVModules.cmake")
+export(TARGETS ${OpenCVModules_TARGETS} APPEND FILE "${CMAKE_BINARY_DIR}/OpenCVModules.cmake")
```
alalek: `APPEND` breaks Android pack build.
Multiple configurations are added to the same file during `build_sdk.py`.

BTW, "install" OpenCVModules doesn't have "APPEND"

mshabunin: Looks like we need to remove this file after each platform build. I'll fix the script.

mshabunin: `install(EXPORT ...` does not have `APPEND` parameter, I think it just takes the final file and copies it to the destination. `export(TARGETS ...` needs `APPEND` when called from several different places with same filename. Or we can try to use `export(EXPORT ...` variant which will export all targets added to our named export set - it needs to be called only once and does not need `APPEND` argument - I'll try this.

PR #13424, opened 2018-12-12
Add ability to build iOS and macOS frameworks with nonfree modules

### LINKED ISSUE #13417: Enabling non-free modules in iOS builds
I need to build OpenCV iOS framework with contrib and with non-free modules **enabled**. Currently, there's no possibility of passing additional CMake flags (like `-DOPENCV_ENABLE_NONFREE=ON`) to `build_framrwork.py`.

There are two proposed solutions to this:

1. Add a parameter `--enable_nonfree` to `build_framework.py` which would add the necessary flag to CMake invocation (would be off by default):

    ```sh
    ./platforms/ios/build_framework.py --enable_nonfree
   ```

   The advantage of this solution is that it's discoverable and pretty self-explanatory.

    The disadvantage is that it covers only one flag — if someone would want to map another CMake flag, they would need to modify build_framework once again.

3. Add a parameter `--cmake_flags` to `build_framework.py` which would allow to specify any flags that will be appended to CMake invocation (would be empty by default):

    ```sh
    ./platforms/ios/build_framework.py --cmake_flags="-DOPENCV_ENABLE_NONFREE=ON"
   ```

    The advantage of this solution is its versatility, users may pass any flag they want there.

    The disadvantage is that it might be misused — flags might be misspelled, conflict with the one used by the script and general number of build failures might increase because of that.

What's your opinion on this?

I will happily implement either solution according to your preference, or none of them if you have any objections or know a better way to handle the problem.

#### System information

- OpenCV: 4.0.0
- Platform: macOS 10.14.1
- Compiler: Xcode 10.1 10B61

akashivskyy:
This Pull Request changes `build_framework.py` script in `platforms/ios` and `osx` to allow specifying that `OpenCV.framework` should be built with nonfree modules enabled, via `--enable_nonfree` boolean flag. It's false by default, thus preserves backwards compatibility.

Resolves #13417.

### DIFF
--- a/platforms/ios/build_framework.py
+++ b/platforms/ios/build_framework.py
@@ -49,7 +49,7 @@ def getXCodeMajor():
         raise Exception("Failed to parse Xcode version")
 
 class Builder:
-    def __init__(self, opencv, contrib, dynamic, bitcodedisabled, exclude, targets):
+    def __init__(self, opencv, contrib, dynamic, bitcodedisabled, exclude, enablenonfree, targets):
         self.opencv = os.path.abspath(opencv)
         self.contrib = None
         if contrib:
@@ -61,6 +61,7 @@ class Builder:
         self.dynamic = dynamic
         self.bitcodedisabled = bitcodedisabled
         self.exclude = exclude
+        self.enablenonfree = enablenonfree
         self.targets = targets
 
     def getBD(self, parent, t):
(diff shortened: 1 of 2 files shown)


alalek:
Please update OSX framework too: https://github.com/opencv/opencv/blob/3.4.4/platforms/osx/build_framework.py#L43

PR #13428, opened 2018-12-12
G-API: Added "kernels" pass stage to compiler

rgarnov:
* Moved "resolve_kernels" and "check_islands_content" passes to separate "kernels" stage in compiler
* Removed unused opaque from cv::gimpl::Op

allow_multiple_commits=1

### DIFF
--- a/modules/gapi/src/backends/fluid/gfluidcore.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidcore.cpp
@@ -340,7 +340,7 @@ static void run_arithm_s3(uchar out[], const uchar in[], int width, const uchar 
         v_store_interleave(&out[3*w], x, y, z);
     }
 #endif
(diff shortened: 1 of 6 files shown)


modules/gapi/src/backends/fluid/gfluidcore.cpp
```diff
@@ -340,7 +340,7 @@ static void run_arithm_s3(uchar out[], const uchar in[], int width, const uchar
         v_store_interleave(&out[3*w], x, y, z);
     }
 #endif
-    UNUSED(v_op);
+    cv::util::suppress_unused_warning(v_op);
```
alalek: BTW, there is `CV_UNUSED` macro in OpenCV:
```
#define CV_UNUSED(name) (void)name
```

PR #12844, opened 2018-10-15
add support for latest Turing gpu and cuda 10

jasjuang:
### This pullrequest changes

Ran into problems with executing OpenCV on the latest Turing GPU we bought. This PR solves it.

```
docker_image:Custom=ubuntu-cuda:16.04
```

PR #13430, opened 2018-12-13
cuda: use correct CC value for Jetson Xavier

tomoaki0705:
### This pullrequest changes
  - This section is for fallback situation on Aarch64 + CUDA environment.
  - Currently, there are only 3 platfrom available on this planet.
     - Jetson TX1 (CC 5.3)
     - Jetson TX2 (CC 6.2)
     - Jetson Xaveir (CC **7.2**)
  - [ref1](https://devtalk.nvidia.com/default/topic/1039504)
  - [ref2](https://devtalk.nvidia.com/default/topic/1039596/)


alalek:
relates #12844

[Wiki](https://en.wikipedia.org/wiki/CUDA).

PR #13431, opened 2018-12-13
CUDA: detect the CUDA CC correctly

tomoaki0705:
### This pullrequest changes
  - fix syntax error
  - very similar to #13430 but 13430 exists both in 3.4/master
  - this bug only exists in master, so no choice but I had to create PR on master


tomoaki0705:
I should have realized it was already fixed.
This was duplicate of #13363 
Closing the PR

PR #13435, opened 2018-12-14
cmake(java): add OPENCV_JAVA_SOURCE_VERSION / OPENCV_JAVA_TARGET_VERSION

### LINKED ISSUE #13434: Not possible to specify Java version during Java build
With the current setup, it is not possible to set the Java build version from the Cmake build properties. This is something that would be very useful, especially with the quicker release cycle of Java.

The ANT build easily supports it with setting source and target properties in the javac block

An example of this can be found at
https://github.com/wpilibsuite/opencv/commit/5b6ade06e8af0ce7b91147024b4dd5c2d75a554a

Would this be something that would be accepted as a PR (Configurable in the build of course, with the default still being not specifying anything like currently happens).

alalek:
resolves #13434

Usage:
```
cmake -DOPENCV_JAVA_SOURCE_VERSION=1.5 -DOPENCV_JAVA_TARGET_VERSION=1.8 ...
cmake --build . --target opencv_java_jar
```

### DIFF
--- a/cmake/OpenCVDetectApacheAnt.cmake
+++ b/cmake/OpenCVDetectApacheAnt.cmake
@@ -1,3 +1,6 @@
+set(OPENCV_JAVA_SOURCE_VERSION "" CACHE STRING "Java source version (javac Ant target)")
+set(OPENCV_JAVA_TARGET_VERSION "" CACHE STRING "Java target version (javac Ant target)")
+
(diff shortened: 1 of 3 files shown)


alalek:
@ThadHouse Does this patch work for you?

ThadHouse:
Sorry, I haven't had a chance to test yet. It looks good to me, an I'll try and test it later today.

alalek:
:+1:

PR #13379, opened 2018-12-06
RGB to/from Gray rewritten to wide intrinsics

savuor:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/560

### This pullrequest changes

* All conversions in `color_rgb.cpp` rewritten to wide universal intrinsics
* Platform-specific code removed
* `mRGBA2RGBA`: saturation added

master branch has different RGB2Gray coefficients so there will be separate PR

<cut/>

### Performance

<details>
<summary>Baseline AVX2</summary>

|Name of Test|rgb initial avx2|rgb wide avx2|rgb wide avx2 vs rgb initial avx2 (x-factor)|
|---|:-:|:-:|:-:|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR2BGR555)|0.011|0.004|3.09|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR2BGR565)|0.011|0.003|3.19|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR2GRAY)|0.006|0.004|1.55|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR5552BGR)|0.007|0.003|2.20|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR5552BGRA)|0.008|0.003|2.48|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR5552GRAY)|0.003|0.003|0.86|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR5552RGB)|0.007|0.003|2.11|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR5552RGBA)|0.008|0.003|2.41|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR5652BGR)|0.007|0.003|2.31|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR5652BGRA)|0.008|0.003|2.63|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR5652GRAY)|0.003|0.003|0.84|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR5652RGB)|0.007|0.003|2.24|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR5652RGBA)|0.008|0.003|2.40|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGRA2BGR555)|0.005|0.004|1.05|
[...]

PR #13440, opened 2018-12-14
RGB to/from Gray rewritten to wide intrinsics (port to master)

savuor:
port of PR #13379 to master

### This pullrequest changes

See original PR description
The only difference is RGB2Gray integer coefficients set which are Q15 here (they were Q14 in 3.4 branch for backward compatibility).


alalek:
:+1:

PR #13447, opened 2018-12-15
videoio(dc1394): use lazy initialization on demand

### LINKED ISSUE #13445: videoio causes a secondary thread to start before main()
- OpenCV => 3.2
- Operating System / Platform => Fedora 26
- Compiler => GCC 7.3.1

##### Detailed description

When you link a program against the videoio library, the OS loader correctly runs videoio's global objects' constructors before passing control to main. However, one of those constructors **causes a secondary thread to be spawned** (see attached stack trace), which effectively makes your program multithreaded.

Furthermore, since the thread is started before any user code has yet had a chance to run, it becomes **impossible to control some of its attributes**. For example, let's take the mask of blocked signals, which is a per-thread attribute that is inherited from the parent thread: even if main() tries to block a signal as soon as it starts, it is too late and **the signal will still be delivered to the secondary thread because it wasn't blocked when the thread was started**.

This is the point where the thread is started (_modules/videoio/src/cap_dc1394_v2.cpp_, `CvDC1394::CvDC1394()`):
```
#0  0x00007ffff08f1644 in pthread_create@@GLIBC_2.2.5 () at /lib64/libpthread.so.0                                                                                                
#1  0x00007fffe392f9db in linux_udev_start_event_monitor () at /lib64/libusb-1.0.so.0                                                                                             
#2  0x00007fffe392e5d7 in op_init () at /lib64/libusb-1.0.so.0                                                                                                                    
#3  0x00007fffe3925dad in libusb_init () at /lib64/libusb-1.0.so.0                                                                                                                
#4  0x00007fffed1f2cdc in dc1394_usb_new () at /lib64/libdc1394.so.22                                                                                                             
#5  0x00007fffed1e1a79 in dc1394_new () at /lib64/libdc1394.so.22                                                                                                                 
#6  0x00007ffff263c329 in _GLOBAL__sub_I_cap_dc1394_v2.cpp () at /lib64/libopencv_videoio.so.3.2                                                                                  
#7  0x00007ffff7de7b33 in _dl_init () at /lib64/ld-linux-x86-64.so.2                                                                                                              
#8  0x00007ffff7dd7e0a in _dl_start_user () at /lib64/ld-linux-x86-64.so.2                                                                                                        
#9  0x0000000000000001 in  ()                                                                                                                                                     
#10 0x00007fffffffdd5c in  ()                                                                                                                                                     
#11 0x0000000000000000 in  ()
```

Please refer to next section for an example program that is unable to block signals because of this issue.

##### Steps to reproduce

I'm attaching an example program that tries, and fails, to block SIGUSR1. The real-life program that made me discover this issue actually tries to block SIGCHLD, in order to be able to receive it through [signalfd](http://man7.org/linux/man-pages/man2/signalfd.2.html).

[example-program.zip](https://github.com/opencv/opencv/files/2682339/example-program.zip)

The included `Makefile` compiles the same source file twice: one without linking to videoio, and one more time with videoio. If you run the two resulting executables and send the USR1 signal from another shell as requested, you will see that the one linked against videoio fails to block it:

```
$ make
g++ example.cpp -o example-without-opencv
g++ example.cpp -o example-with-opencv -Wl,--no-as-needed -lopencv_videoio

$ ./example-without-opencv 
Now run "kill -USR1 30574" to demonstrate the issue
^C

$ ./example-with-opencv 
Now run "kill -USR1 30638" to demonstrate the issue
In SIGUSR1 handler! This should NOT happen!!
^C
```

alalek:
resolves #13445

### DIFF
--- a/modules/videoio/src/cap_dc1394_v2.cpp
+++ b/modules/videoio/src/cap_dc1394_v2.cpp
@@ -192,7 +192,11 @@ CvDC1394::~CvDC1394()
     dc = 0;
 }
 
-static CvDC1394 dc1394;
+static CvDC1394& getDC1394()
+{
+    static CvDC1394 dc1394;
+    return dc1394;
+}
 
 class CvCaptureCAM_DC1394_v2_CPP : public CvCapture
 {
@@ -451,7 +455,7 @@ bool CvCaptureCAM_DC1394_v2_CPP::startCapture()
     code = dc1394_capture_setup(dcCam, nDMABufs, DC1394_CAPTURE_FLAGS_DEFAULT);
     if (code >= 0)
     {
-        FD_SET(dc1394_capture_get_fileno(dcCam), &dc1394.camFds);
+        FD_SET(dc1394_capture_get_fileno(dcCam), &getDC1394().camFds);
         dc1394_video_set_transmission(dcCam, DC1394_ON);
         if (cameraId == VIDERE)
         {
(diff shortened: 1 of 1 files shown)


alalek:
:+1:

PR #13449, opened 2018-12-15
Stitching: added functions to set warp interpolation mode

Tytan:
### This pullrequest changes

Add methods to select the interpolation mode used when warping images in the stitcher

### DIFF
--- a/modules/stitching/include/opencv2/stitching.hpp
+++ b/modules/stitching/include/opencv2/stitching.hpp
@@ -193,6 +193,9 @@ public:
     CV_WRAP bool waveCorrection() const { return do_wave_correct_; }
     CV_WRAP void setWaveCorrection(bool flag) { do_wave_correct_ = flag; }
 
+    CV_WRAP int interpolation() const { return interp_mode_; }
+    CV_WRAP void setInterpolation(int interp) { interp_mode_ = interp; }
+
     detail::WaveCorrectKind waveCorrectKind() const { return wave_correct_kind_; }
     void setWaveCorrectKind(detail::WaveCorrectKind kind) { wave_correct_kind_ = kind; }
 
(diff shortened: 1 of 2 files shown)


modules/stitching/include/opencv2/stitching.hpp
```diff
@@ -193,6 +193,9 @@ class CV_EXPORTS_W Stitcher
     CV_WRAP bool waveCorrection() const { return do_wave_correct_; }
     CV_WRAP void setWaveCorrection(bool flag) { do_wave_correct_ = flag; }
 
+    CV_WRAP int interpolation() const { return interp_mode_; }
+    CV_WRAP void setInterpolation(int interp) { interp_mode_ = interp; }
```
alalek: > int

It is better to use `InterpolationFlags` enum instead of raw "int".

modules/stitching/include/opencv2/stitching.hpp
```diff
@@ -193,6 +193,9 @@ class CV_EXPORTS_W Stitcher
     CV_WRAP bool waveCorrection() const { return do_wave_correct_; }
     CV_WRAP void setWaveCorrection(bool flag) { do_wave_correct_ = flag; }
 
+    CV_WRAP int interpolation() const { return interp_mode_; }
```
alalek: > interpolation

getInterpolation / getInterpolationFlags

modules/stitching/include/opencv2/stitching.hpp
Tytan: Indeed, this is fixed

modules/stitching/include/opencv2/stitching.hpp
Tytan: I replaced `interpolation` by `getInterpolationFlags` and `setInterpolation` by `setInterpolationFlags`, and renamed the attribute accordingly.
I notice that all other getters of the Stitcher are named `attribute()` and not `getAttribute()`, is this intentional ?

alalek: Looks like you are right!

This is not aligned with other OpenCV parts, but lets keep stitching API consistent.

Renamed to `interpolationFlags()`

PR #13451, opened 2018-12-16
Off-by-ones in windows/GDI window calculations

crackwitz:
### This pullrequest changes

When looking closely, there's a gray "gutter" around the displayed image. Since it's just 1 pixel wide, it never showed up on the bottom and right borders between image and window chrome.

That gutter is now visible properly. Perhaps it would make sense to remove it entirely (set to zero in `icvCalcWindowRect`). Previously that _would_ have cut off the last row and column of an image.

There has been no issue for this and it fixes no bugs (to my knowledge). This is purely cosmetic.


crackwitz:
I have no idea how to do this "properly". there is a lot of difference between master and 3.4. please bear with me.

crackwitz:
if there's anything further I should do, please let me know. this level of working with git is still beyond me.

PR #13453, opened 2018-12-16
fix for #4990: correct calculation of image coordinates from window coordinates

### LINKED ISSUE #4990: Mouse call back always give 0for x and y coordinates in Opengl window
Transferred from http://code.opencv.org/issues/4338

```
|| Laurent Berger on 2015-05-14 14:58
|| Priority: Normal
|| Category: highgui-gui
|| Tracker: Bug
|| Difficulty: Medium
|| PR: 
|| Platform: x64 / Windows
```
## Mouse call back always give 0for x and y coordinates in Opengl window

```
in file window_w32.cpp at line 1617 
window->on_mouse( event, pt.x*size.cx/MAX(rect.right - rect.left,1),
                                 pt.y*size.cy/MAX(rect.bottom - rect.top,1), flags,
                                 window->on_mouse_param );

size seems always (0,0) in a window when window->flags &cv::WINDOW_OPENGL is not zero
```

crackwitz:
resolves #4990

### This pullrequest changes

in `HighGUIProc` case `WM_MOUSEMOVE` as well as in `MainWindowProc` case `WM_MOUSEWHEEL`, window coordinates are turned into image coordinates. Unfortunately, the size of the image is stored nowhere explicitly and generally. It is determined by querying (`icvGetBitmapData`) the GDI bitmap object associated with the window. In the standard GDI case, that's the original image.

In the OpenGL case, that is instead the whole window (including chrome). The _actual_ image is only stored in `window->glDrawData`, which is a pointer to `cv::ogl::Texture2D` and to be used by the drawing callback to OpenGL.

I've added code to determine image size according to situation.

### DIFF
--- a/modules/highgui/src/window_w32.cpp
+++ b/modules/highgui/src/window_w32.cpp
@@ -66,6 +66,7 @@
 #include <functional>
 #include "opencv2/highgui.hpp"
 #include <GL/gl.h>
+#include "opencv2/core/opengl.hpp"
 #endif
 
 static const char* trackbar_text =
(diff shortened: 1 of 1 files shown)


modules/highgui/src/window_w32.cpp
```diff
@@ -1421,7 +1422,20 @@ MainWindowProc( HWND hwnd, UINT uMsg, WPARAM wParam, LPARAM lParam )
           GetClientRect( window->hwnd, &rect );
 
           SIZE size = {0,0};
-          icvGetBitmapData( window, &size, 0, 0 );
+#ifdef HAVE_OPENGL
+          if (window->useGl)
```
mshabunin: What about moving this if-block into `icvGetBitmapData`?

crackwitz: not my call either way. I chose not to because I don't fully comprehend the naming convention of `icvGetBitmapData` and related functions and they seem to not deal with OpenGL at all.

I would suggest not querying whatever bitmap is stored for whatever mode, but storing the size in `window` explicitly when bitmap data is set. then, none of these case distinctions have to be made.

alalek:

So, please:
- (**done**) change "base" branch of this PR: master => 3.4 (use "Edit" button near PR title)

Note: no needs to re-open PR or delete branch, apply changes "inplace".

PR #13467, opened 2018-12-18
features2d: fix retainBest() implementation

### LINKED ISSUE #12594: Incorrect iterator in KeyPointsFilter::retainBest
OpenCV => Source code on master
In modules/features2d/src/keypoint.cpp, function `KeyPointsFilter::retainBest` at line 80:
`  std::nth_element(keypoints.begin(), keypoints.begin() + n_points, keypoints.end(), ...
`

This will rearrange the list such that the first `n_points + 1` elements are greater than or equal to the subsequent points, ie `keypoints[0]` through `keypoints[n_points]`. However, the ambiguous response is taken from element `n_points - 1` and there is no guarantee of ordering within this region, so it may actually be quite different from the true ambiguous value at `keypoints[n_points]`.

When called on keypoint vectors that are identical but for ordering (as may happen when switching feature detector implementation from CPU to GPU), the results of `retainBest` may vary.

This only potentially affects the lowest valued keypoints that would otherwise be retained so probably not a big deal performance-wise, but it does manifest when testing different implementations of keypoint generators that return their values in different orders.

alalek:
resolves #12594

Thanks to @davep412 for reproducer and fix!

### DIFF
--- a/modules/features2d/src/keypoint.cpp
+++ b/modules/features2d/src/keypoint.cpp
@@ -77,7 +77,7 @@ void KeyPointsFilter::retainBest(std::vector<KeyPoint>& keypoints, int n_points)
             return;
         }
         //first use nth element to partition the keypoints into the best and worst.
-        std::nth_element(keypoints.begin(), keypoints.begin() + n_points, keypoints.end(), KeypointResponseGreater());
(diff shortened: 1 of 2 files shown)


alalek:
:+1:

PR #13473, opened 2018-12-18
find Protobuf in OpenCVConfig.cmake when OpenCV was build as static lib

hesmar:
When building OpenCV as static library, the target 'libprotobuf' is part of
the INTERFACE_LINK_LIBRARIES. Therefore, Protobuf should be found
by the OpenCVConfig.cmake if it was not found before and the
libprotobuf target has to be created.

### DIFF
--- a/cmake/OpenCVGenConfig.cmake
+++ b/cmake/OpenCVGenConfig.cmake
@@ -59,6 +59,13 @@ else()
   set(USE_IPPIW FALSE)
 endif()
 
+if(TARGET libprotobuf AND NOT BUILD_SHARED_LIBS AND NOT BUILD_PROTOBUF)
+  set(USE_PROTOBUF TRUE)
+  ocv_cmake_configure("${CMAKE_CURRENT_LIST_DIR}/templates/OpenCVConfig-Protobuf.cmake.in" PROTOBUF_CONFIGCMAKE @ONLY)
+else()
+  set(USE_PROTOBUF FALSE)
+endif()
+
 ocv_cmake_hook(PRE_CMAKE_CONFIG_BUILD)
 configure_file("${OpenCV_SOURCE_DIR}/cmake/templates/OpenCVConfig.cmake.in" "${CMAKE_BINARY_DIR}/OpenCVConfig.cmake" @ONLY)
 #support for version checking when finding opencv. find_package(OpenCV 2.3.1 EXACT) should now work.
@@ -77,6 +84,9 @@ endif()
 if(USE_IPPIW)
   file(RELATIVE_PATH IPPIW_INSTALL_PATH_RELATIVE_CONFIGCMAKE "${CMAKE_INSTALL_PREFIX}" "${IPPIW_INSTALL_PATH}")
   ocv_cmake_configure("${CMAKE_CURRENT_LIST_DIR}/templates/OpenCVConfig-IPPIW.cmake.in" IPPIW_CONFIGCMAKE @ONLY)
+endif()
+if(USE_PROTOBUF)
+  ocv_cmake_configure("${CMAKE_CURRENT_LIST_DIR}/templates/OpenCVConfig-Protobuf.cmake.in" PROTOBUF_CONFIGCMAKE @ONLY)
 endif()
 
 function(ocv_gen_config TMP_DIR NESTED_PATH ROOT_NAME)
--- a/cmake/templates/OpenCVConfig-Protobuf.cmake.in
+++ b/cmake/templates/OpenCVConfig-Protobuf.cmake.in
@@ -1,0 +1,56 @@
+set(OpenCV_Protobuf_VERSION "@Protobuf_VERSION@")
+
+function(get_protobuf_version version include)
+  file(STRINGS "${include}/google/protobuf/stubs/common.h" ver REGEX "#define GOOGLE_PROTOBUF_VERSION [0-9]+")
+  string(REGEX MATCHALL "[0-9]+" ver ${ver})
+  math(EXPR major "${ver} / 1000000")
+  math(EXPR minor "${ver} / 1000 % 1000")
+  math(EXPR patch "${ver} % 1000")
+  set(${version} "${major}.${minor}.${patch}" PARENT_SCOPE)
+endfunction()
+
+if(NOT Protobuf_FOUND)
+  find_package(Protobuf CONFIG QUIET)
+endif()
+
+if(NOT Protobuf_FOUND)
+  find_package(Protobuf MODULE QUIET)
+endif()
+
+# Backwards compatibility
+# Define camel case versions of input variables
+foreach(UPPER
+    PROTOBUF_FOUND
+    PROTOBUF_LIBRARY
+    PROTOBUF_INCLUDE_DIR
+    PROTOBUF_VERSION
+    )
+    if (DEFINED ${UPPER})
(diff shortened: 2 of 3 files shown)


alalek:
Builtin protobuf is included via OpenCVConfig (OpenCVModules).

**External** protobuf (your case) - likely "won't fix".
- Like many other external static libraries (static FFmpeg (which needs huge list of dependencies in default configuration, bz2 is not the worst case), LAPACK, IPP, MKL, ...).
- External shared libraries - the same - we don't want to control redistribution of external code.
- "Find" is not the worst problem here, more complex problem is redistribution of these static dependencies with their dependencies with their nested dependencies ...
- We are not going to transform OpenCVConfig into "package manager".
- There are too many options of external packages (which are hard to test). This configuration is not the default build, so it will be broken too quickly and frequently.

I believe we can add some "hooks" support in OpenCVConfig.cmake - check and `include()` extra `.cmake` files distributed within such static OpenCV builds with static dependencies.

hesmar:
> Builtin protobuf is included via OpenCVConfig (OpenCVModules).
> 
> **External** protobuf (your case) - likely "won't fix".
> 
>     * Like many other external static libraries (static FFmpeg (which needs huge list of dependencies in default configuration, bz2 is not the worst case), LAPACK, IPP, MKL, ...).
> 
>     * External shared libraries - the same - we don't want to control redistribution of external code.
> 
>     * "Find" is not the worst problem here, more complex problem is redistribution of these static dependencies with their dependencies with their nested dependencies ...
> 
>     * We are not going to transform OpenCVConfig into "package manager".
> 
>     * There are too many options of external packages (which are hard to test). This configuration is not the default build, so it will be broken too quickly and frequently.

I understand you concerns. Nevertheless, it is currently not possible to build an application with static OpenCV that was compiled with external protobuf. External protobuf is necessary when the application or other used libraries also use protobuf, because mixing different versions is not possible.
IMHO, the CMake config files of libraries should care about the dependencies, not the applications which use the library.

It is already done for CUDA, why not for other dependencies?
 
> I believe we can add some "hooks" support in OpenCVConfig.cmake - check and `include()` extra `.cmake` files distributed within such static OpenCV builds with static dependencies.

What is the difference to the current approach? Just the different file? If you like I could move it to a separate file which is included in the OpenCVConfig.cmake. Or do you think about something different?

alalek:
I mean additional optional files. But we don't want to maintain and/or put them into repository. Maintaining of these files is responsibility of binary package creator.
OpenCVConfig.cmake will just include these extra CMake files, which should "find" required dependencies for OpenCV.

hesmar:
> I mean additional optional files. But we don't want to maintain and/or put them into repository. Maintaining of these files is responsibility of binary package creator.
> OpenCVConfig.cmake will just include these extra CMake files, which should "find" required dependencies for OpenCV.

I still do not understand what you mean with "additional optional files". Maybe you can describe it a bit more detail? Are you talking about the library files itself? If yes, I didn't say that they have to be included into the repository. I am just saying that the the find_package (or [find_dependency()](https://cmake.org/cmake/help/v3.12/module/CMakeFindDependencyMacro.html)) call is missing and should be included to OpenCVConfig.cmake. If that call is inside the OpenCVConfig.cmake itself or inside an extra file (e.g. OpenCVFindDependencies.cmake or OpenCVFindProtobuf.cmake) which is included from the OpenCVConfig.cmake doesn't matter.

IMHO it should work as follows:
1. The application calls find_package(OpenCV) which includes the OpenCVConfig.cmake.
2. Inside the OpenCVConfig.cmake, dependencies of OpenCV (Cuda, Protobuf, etc.) should be found using find_dependency().
3. The Config files of the dependencies (eg. ProtobufConfig.cmake) maybe again include dependencies and so on...

Please have a look at this nice blog article: [https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/#import-your-targets-inside-your-configcmake](https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/#import-your-targets-inside-your-configcmake) There is also described why it is necessary to find the dependent packages inside the config file.

mshabunin:
@hesmar , `find_dependency` will use same parameters as parent call of `find_package(OpenCV)`, e.g. `REQUIRED`. But OpenCV is a modular library and user can link only to `opencv_core` module and it should be possible even if protobuf (which is dependency of `opencv_dnn` module) is not found.

mshabunin:
There can be a problem, because build-time environment can differ from usage-time environment. I don't think we can easily make static packages relocatable. So I suggest alternative variant, similar to current behaviour: we can create interface targets for each found dependency, set up all necessary properties, link them with modules using PRIVATE modifier and export/install them along with other modules.

In PR #13422 I tried to unify and separate all dependencies of videoio module and initially made them `INTERFACE IMPORTED` targets ([example](https://github.com/opencv/opencv/pull/13422/files#diff-147a2d6c0a28c306ecf5078d84c5eb84R58)). But this approach breaks static builds as in your case, so I tried to make them INTERFACE and export/install ([macro](https://github.com/opencv/opencv/pull/13422/files#diff-48102fe265e32fbd4e7a1054fb92a9cbR7), [usage](https://github.com/opencv/opencv/pull/13422/files#diff-d6bf9a4e7d80e9f28cb4953df07d4954R98)). Then `OpenCVModules.cmake` will contain new target definition:
```
# Create imported target ocv.3rdparty.gstreamer
add_library(ocv.3rdparty.gstreamer INTERFACE IMPORTED)

set_target_properties(ocv.3rdparty.gstreamer PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "HAVE_GSTREAMER"
  INTERFACE_INCLUDE_DIRECTORIES "/usr/include/gstreamer-1.0;/usr/include/glib-2.0;/usr/lib/x86_64-linux-gnu/glib-2.0/include;/usr/include/gstreamer-1.0;/usr/include/glib-2.0;/usr/lib/x86_64-linux-gnu/glib-2.0/include;/usr/include/gstreamer-1.0;/usr/include/glib-2.0;/usr/lib/x86_64-linux-gnu/glib-2.0/include;/usr/include/gstreamer-1.0;/usr/include/glib-2.0;/usr/lib/x86_64-linux-gnu/glib-2.0/include"
  INTERFACE_LINK_LIBRARIES "gstbase-1.0;gstreamer-1.0;gobject-2.0;glib-2.0;gstapp-1.0;gstbase-1.0;gstreamer-1.0;gobject-2.0;glib-2.0;gstriff-1.0;gstreamer-1.0;gobject-2.0;glib-2.0;gstpbutils-1.0;gstreamer-1.0;gobject-2.0;glib-2.0"
)
```
And it will be linked to videoio (only in case of static build):
```
# Create imported target opencv_videoio
add_library(opencv_videoio STATIC IMPORTED)

set_target_properties(opencv_videoio PROPERTIES
  INTERFACE_LINK_LIBRARIES "opencv_core;opencv_imgproc;opencv_imgcodecs;\$<LINK_ONLY:dl>;\$<LINK_ONLY:m>;\$<LINK_ONLY:pthread>;\$<LINK_ONLY:rt>;\$<LINK_ONLY:ippiw>;\$<LINK_ONLY:ippicv>;\$<LINK_ONLY:ocv.3rdparty.dc1394_2>;\$<LINK_ONLY:ocv.3rdparty.gstreamer>;\$<LINK_ONLY:ocv.3rdparty.v4l>"
)
```

Issue #13227, opened 2018-11-20
drawMatches with CV_8UC4 inputs draws on blank background


dan-masek:
Based on the code, seems relevant to all OpenCV versions/platforms.

When I call `cv::drawMatches` with `img1` and `img2` being BGRA images and `outImg` left empty, it doesn't complain. Since the documentation doesn't mention any restrictions on channel count, I assume this is a valid use-case. However, it fails to produce a background that's a composition of the two input images, as happens for 1 and 3 channel images.

The problem seems to be here: https://github.com/opencv/opencv/blob/3.4/modules/features2d/src/draw.cpp#L149

There's no check for invalid image depth as there is in [`drawKeypoints`](https://github.com/opencv/opencv/blob/3.4/modules/features2d/src/draw.cpp#L106), and the `copyTo` ends up reallocating the ROIs (so the master image is unmodified).

Also, IMHO `CV_8UC1` would seem more appropriate than just plain `CV_8U` when checking the `Mat` type (it's not entirely obvious that it means a 1-channel image, even though the enum values are the same).

**Example script:**

    import numpy as np
    import cv2

    imageLeft = cv2.cvtColor(cv2.imread('box.png'), cv2.COLOR_BGR2BGRA)
    imageRight = cv2.cvtColor(cv2.imread('box_in_scene.png'), cv2.COLOR_BGR2BGRA)

    orb = cv2.ORB_create()
    kpLeft, desLeft = orb.detectAndCompute(imageLeft, None)
    kpRight, desRight = orb.detectAndCompute(imageRight, None)

    bf = cv2.BFMatcher(cv2.NORM_HAMMING, crossCheck=True)
    matches = bf.match(desLeft, desRight)

    goodMatches = []
    for m in matches:
        if m.distance < 100:
            goodMatches.append(m)
            
    output = cv2.drawMatches(imageLeft, kpLeft, imageRight, kpRight, [goodMatches[0]], None, flags=0)

    cv2.imwrite('output.png', output)

**Actual output:**

![Actual output](https://user-images.githubusercontent.com/10824015/48800255-8f54fb80-ed09-11e8-9a1c-12efee75ebe3.png)

**Expected output:**

![Expected ouput](https://user-images.githubusercontent.com/10824015/48800264-9714a000-ed09-11e8-8298-9c4ee83a89e8.png)

dan-masek:
What would be the preferred solution to this? 
[...]

PR #13481, opened 2018-12-19
Fix drawMatches and drawKeypoints for BGRA input images

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/13227

### DIFF
--- a/modules/features2d/src/draw.cpp
+++ b/modules/features2d/src/draw.cpp
@@ -103,9 +103,13 @@ void drawKeypoints( InputArray image, const std::vector<KeyPoint>& keypoints, In
         {
             cvtColor( image, outImage, COLOR_GRAY2BGR );
         }
+        else if( image.type() == CV_8UC4 )
+        {
+            cvtColor( image, outImage, COLOR_BGRA2BGR );
+        }
         else
         {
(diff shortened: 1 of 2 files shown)


modules/features2d/src/draw.cpp
```diff
@@ -103,9 +103,13 @@ void drawKeypoints( InputArray image, const std::vector<KeyPoint>& keypoints, In
         {
             cvtColor( image, outImage, COLOR_GRAY2BGR );
         }
+        else if( image.type() == CV_8UC4 )
+        {
+            cvtColor( image, outImage, COLOR_BGRA2BGR );
```
alalek: It is not needed to force 3 channel images. For example, 4 channel "RGBA" is used on Android.

Update the first "copyTo" condition instead:
```
if (image.type() == CV_8UC3 || image.type() == CV_8UCU)
```

dkurt: If `drawKeypoints` can produce BGRA output images we can expect similar functionality from `drawMatches`. However `drawMatches` takes two inputs and we need to add extra logic to determine output depth.

alalek: We can add BGRA (4 channels) support into drawMatches too: https://github.com/alalek/opencv/commits/pr13481

PR #13482, opened 2018-12-19
Fix python bindings without stitching module

dkurt:
### This pullrequest changes

fix after https://github.com/opencv/opencv/pull/13267


LaurentBerger:
sorry I think that I misunderstood that it was not in core/misc/python but in stitching/misc/python

PR #13486, opened 2018-12-20
DNN: Openpose: Fixed samples

allnes:


samples/dnn/openpose.py
```diff
@@ -16,6 +16,7 @@
 parser.add_argument('--thr', default=0.1, type=float, help='Threshold value for pose parts heat map')
 parser.add_argument('--width', default=368, type=int, help='Resize input to specific width.')
 parser.add_argument('--height', default=368, type=int, help='Resize input to specific height.')
+parser.add_argument('--scale', default=0.003922, type=float, help='Scale for blob.')
```
dkurt: Please also remove redundant `To use Inference Engine backend, specify location of plugins:` at the beginning of this script.

samples/dnn/openpose.cpp
```diff
@@ -102,16 +104,17 @@ int main(int argc, char **argv)
...
         npairs = 17;
         nparts = 18; // skip background
     }
-    else if (nparts == 16)
+    else if (nparts == 16 || nparts == 44)
     {   // MPI body
         midx   = 1;
         npairs = 14;
+        nparts = 16;
     }
```
dkurt: I think it's better to add a flag `dataset` similar to python script and add an extra option 'HAND' to it's choices instead determining model by number of output channels.

samples/dnn/openpose.py
```diff
@@ -57,12 +59,15 @@
...
+    inp = cv.dnn.blobFromImage(frame, inScale, (inWidth, inHeight),
                               (0, 0, 0), swapRB=False, crop=False)
     net.setInput(inp)
     out = net.forward()
 
-    assert(len(BODY_PARTS) == out.shape[1])
+    shape_1 = out.shape[1]
+    if shape_1 == 57: shape_1 = 19
+    if shape_1 == 44: shape_1 = 16
+    assert(len(BODY_PARTS) == shape_1)
```
dkurt: `shape_1` is used only here. We check `assert(len(BODY_PARTS) == out.shape[1])` only to prevent out-of-bound slicing. Can we keep just `assert(len(BODY_PARTS) <= out.shape[1])`?

May I ask you also adapt python sample for HAND model?

allnes:
@dkurt done

PR #13488, opened 2018-12-20
videoio(v4l2): fix build due missing V4L2_CID_ISO_SENSITIVITY

### LINKED ISSUE #13335: Opencv 3.4.4 dependency on V4L2_CID_ISO_SENSITIVITY only available in kernel >= 3.5
Pull request https://github.com/opencv/opencv/pull/12893 added support for newer V4L camera properties, but these are only available on newer kernel versions (see https://www.kernel.org/doc/html/v4.14/media/uapi/v4l/hist-v4l2.html for, as an example, V4L2_CID_ISO_SENSITIVITY being introduced in 3.5).

I am not familiar with how Opencv typically handles this kind of scenario. Should these kinds of changes be behind preprocessor checks?

alalek:
resolves #13335

### DIFF
--- a/modules/videoio/src/cap_v4l.cpp
+++ b/modules/videoio/src/cap_v4l.cpp
@@ -237,6 +237,11 @@ make & enjoy!
 #include <sys/videoio.h>
 #endif
 
+// https://github.com/opencv/opencv/issues/13335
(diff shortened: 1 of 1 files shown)


washley:
Sorry I didn't have time to take a pass at fixing this myself, thanks for getting this out.

PR #13492, opened 2018-12-20
Fixed operations on Mat and Matx simultaneously

### LINKED ISSUE #13483: Addind Vec to Mat is incorrect
OpenCV 3.4.4

```
Mat mat = Mat::zeros(3, 1, CV_64F);
Vec3d vec(1, 2, 3);
cout << mat + vec << endl;
cout << mat + Mat(vec) << endl;
```

Output:
[1;
1;
1]
[1;
2;
3]

There is no runtime error, also I did not found any info in docs

terfendail:
resolves #13483

### This pullrequest changes
Fixed operations on Mat and Matx simultaneously


alalek:
Please add tests, including operations with `Scalar` (which is `Vec<>` which is `Matx<>`)

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -3597,6 +3597,10 @@ CV_EXPORTS MatExpr operator + (const Mat& m, const MatExpr& e);
 CV_EXPORTS MatExpr operator + (const MatExpr& e, const Scalar& s);
 CV_EXPORTS MatExpr operator + (const Scalar& s, const MatExpr& e);
 CV_EXPORTS MatExpr operator + (const MatExpr& e1, const MatExpr& e2);
+template<typename _Tp, int m, int n> inline
+CV_EXPORTS MatExpr operator + (const Mat& a, const Matx<_Tp, m, n>& b) { return a + Mat(b); }
```
alalek: > CV_EXPORTS

Remove for inlined functions.

> inline

Use "static inline" instead to avoid adding of new symbols into binaries.

terfendail: fixed

PR #13496, opened 2018-12-21
Internalize protobuf symbols 

peters:
### This pullrequest changes

Internalizes protobuf symbols in order for opencv to be statically linked on Ubuntu 18.04 / gcc 8.2.0.


alalek:
We already have these two flags in CMAKE_CXX_FLAGS: `-fvisibility=hidden -fvisibility-inlines-hidden`

peters:
@alalek Yes, it’s clearly stated in the Wiki but I’m building a fat .so of multiple static libraries and GCC refuses to mangle protobuf without this explicitly set. I’ve tried with a clean build (all sources are out of tree) and it does not work. Ideas?

PR #13503, opened 2018-12-21
Stitching: multi-feed on gain and block compensator

Tytan:
### This pullrequest changes

Add the possibility to feed images multiple times into the GainCompensator or BlocksGainComprensator. The method is taken from [this document](http://pages.cs.wisc.edu/~lizhang/courses/cs766-2007f/projects/mosaic/uyttendaele-cvpr01.pdf).

> We may then iterate and use the adjusted composite as the reference for the next  step. In our implementation we find the 3 iterations yields good results.

Note that I kept the default number of feeds to 1 to keep the default behavior as it was before this patch.

As shown by the perf test below, exposure compensation usually accounts for a small portion of the stitching time (sometimes it seems irrelevant due to variance...) and performing the step multiple times is not a huge problem.

```
Geometric mean (ms)

                  Name of Test                      stitching   
a123::stitchExposureCompMultiFeed::("blocks", 1)     629.115    
a123::stitchExposureCompMultiFeed::("blocks", 2)     646.658    
a123::stitchExposureCompMultiFeed::("blocks", 3)     687.371    
a123::stitchExposureCompMultiFeed::("blocks", 4)     684.875    
a123::stitchExposureCompMultiFeed::("blocks", 5)     798.196    
a123::stitchExposureCompMultiFeed::("gain", 1)       554.134    
a123::stitchExposureCompMultiFeed::("gain", 2)       630.136    
a123::stitchExposureCompMultiFeed::("gain", 3)       622.126    
a123::stitchExposureCompMultiFeed::("gain", 4)       649.410    
a123::stitchExposureCompMultiFeed::("gain", 5)       674.722
```


Tytan:
Had to push commit 2466a3b to make pullrequest.opencv.org happy about the ABI compatibility.
I believe opencv4 does not care about ABI change as long as the API is compatible, did I misunderstand something ? If so, feel free to remove 2466a3b.

alalek:
@Tytan "Source level" API checker is enabled for the master branch. But look like it doesn't properly track "default" argument values (this is false-positive).
Change is correct to make this checker happy.

PR #13504, opened 2018-12-21
stitcher result mask getter

Tytan:
Add the ability to retrieve the result mask of the stitcher.


yanzixiang:
can you add CV_WRAP to this,
and let other bindings to use this ?
and can you explain how to use this ?

Issue #13506, opened 2018-12-22
regression: Sobel is broken for 16S/16U


alalek:
Regression from #13392 can be observed with these inputs:
- dx=0
- dy=1
- kernelSize=5
- ddepth=CV_16U  (output is still CV_16S)
- input Mat CV_16UC1 (any random values):
```
[127, 138, 130, 102, 118, 97, 76, 84, 124, 90, 146, 63, 130, 87, 212, 85;
 164, 3, 51, 124, 151, 89, 154, 117, 36, 88, 116, 117, 180, 112, 147, 124;
 63, 50, 115, 103, 83, 148, 106, 79, 213, 106, 135, 53, 79, 106, 122, 112;
 218, 107, 81, 126, 78, 138, 85, 142, 151, 108, 104, 158, 155, 81, 112, 178;
 184, 96, 187, 148, 150, 112, 138, 162, 222, 146, 128, 49, 124, 46, 165, 104;
 119, 164, 77, 144, 186, 98, 106, 148, 155, 157, 160, 151, 156, 149, 43, 122;
 106, 155, 120, 132, 159, 115, 126, 188, 44, 79, 164, 201, 153, 97, 139, 133;
 133, 98, 111, 165, 66, 106, 131, 85, 176, 156, 67, 108, 142, 91, 74, 137]
```

Expected result:
```
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0;
 -1020, -796, -489, -469, -247, 317, 760, 1429, 1983, 1384, 254, -459, -899, -1197, -1172, -1058;
 2552, 2340, 1617, 591, 9, 96, 722, 1985, 2746, 1916, 676, 9, -635, -1115, -779, -380;
 3546, 3349, 2838, 2206, 1388, 669, 938, 1880, 2252, 1785, 1083, 606, 180, -298, -464, -418;
 816, 966, 1255, 1652, 1619, 924, 535, 288, 5, 601, 1581, 1870, 1520, 625, -627, -1260;
 -782, -610, -395, -267, -122, -42, -317, -1378, -2293, -1451, 596, 1870, 1679, 763, -69, -394;
 -882, -681, -463, -818, -1167, -732, -463, -1042, -1604, -1592, -1047, -334, -104, -117, 229, 512;
 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

Actual result:
```
[...]

PR #13392, opened 2018-12-07
Reworked separable filter to use wide universal intrinsics

terfendail:
### This pullrequest changes
Reworked separable filter to use wide universal intrinsics

TODO:
- [x] Performance reports for SSE2 / SSE4.2 / AVX2

```
allow_multiple_commits=1
```

PR #13507, opened 2018-12-22
imgproc: revert "Speedup filter2d by loop unrolling"

alalek:
Commit: 124011c3219b4be6b90aec4538b4560711deb3d2
PR: https://github.com/opencv/opencv/pull/13392

Sobel filter with 16S/16U datatype is broken.

relates #13506


alalek:
:+1:

PR #13532, opened 2018-12-26
Channels exposure compensators

Tytan:
### This pullrequest changes

Adds ChannelsCompensator and BlocksChannelsCompensator. The idea is to compensator each channel independantly. This is necessary if the images we want to stitch together do not have the same white balance.

For instance, I used three images of the "boatN.jpg" samples from opencv_extra, and modified the white balance of each of those. I then compared the result using the 4 compensators.

[Base images](https://ibb.co/y53HG7s)
1st one is original.
2nd one is colder.
3rd one is warmer.

[Stitching result](https://ibb.co/yVXzNfH)
1st one is `GainCompensator`
2nd one is `BlocksGainCompensator`
3rd one is `ChannelsCompensator`
4th one is `BlocksChannelsCompensator`

### Patch description
Commits e03ce35, 6ae003e and 833212e are quite simple.

5eb7d8a  brings some deep changes in the `BlocksGainCompensator`. To avoid code duplication I created a `BlockCompensator` which resembles the old `BlocksGainCompensator`. The difference is that is uses a templated `feed` method. The template argument determines which compensator to use. Then the class uses an overloaded `getGainMap` function to fill its gain map vector (as the gain format depends on the compensator used).

The final clases `BlocksGainCompensator` and `BlocksChannelsCompensator` simply inherit the generic `BlocksCompensator` and calls the right templated `feed`.

### DIFF
--- a/modules/stitching/include/opencv2/stitching/detail/exposure_compensate.hpp
+++ b/modules/stitching/include/opencv2/stitching/detail/exposure_compensate.hpp
@@ -110,6 +110,7 @@ intensities, see @cite BL07 and @cite WJ10 for details.
 class CV_EXPORTS_W GainCompensator : public ExposureCompensator
 {
 public:
+    // This Constructor only exists to make source level compatibility detector happy
     CV_WRAP GainCompensator()
             : GainCompensator(1) {}
     CV_WRAP GainCompensator(int nr_feeds)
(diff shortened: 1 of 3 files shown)


Tytan:
This is basically just an extension of the ExposureCompensators that exist in the stitching module. This still requires overlapping images to work and the goal is to find a common ground for exposition and white balance over overlapping regions, although it may not the ideal exposition or white balance overall. It is not intended to make the individual images look or feel "better", but to make sure they will merge more smoothly.

In the case of the stitcher, it is quite easy to implement as we have already determined overlapping region that we expect to look the same, if we want to find a common ground for exposure and white balance of individual photos outside of the stitching pipeline, the problem is much more complex.

Basically, the algorithm already exists in OpenCV, as it is just a gain on each channel. What this PR brings is the ability for the stitching pipeline to determine gains for each channel independently.

If I missed your point, please tell me :)

PR #13544, opened 2018-12-29
Fix a bug in cv :: merge when array of 3-channel mat is input

eightco:
The merge of the array of 3channel mat gives the wrong result on a particular input.<br/>
When an array of mat entering as input type If each Mat is not a single channel, but array of three or four ```Mat```'s of the three channels is input, the ```ipp_merge()``` call will result in an incorrect merge.<br/>
```ipp_merge ()``` seems to have been implemented assuming that the ```Mat``` channels are all 1 channel.
Therefore, we must move the ```ipp_merge()``` call below the ```if (! Allch1)``` branch to work properly.

#### Steps to Reproduce
```cpp
Mat src1(2, 2, CV_8UC3, Scalar::all(1));
Mat src2(2, 2, CV_8UC3, Scalar::all(2));
Mat src3(2, 2, CV_8UC3, Scalar::all(3));
Mat src_arr[] = { src1, src2, src3 };
Mat dst;
merge(src_arr, 3, dst);
cout << dst << endl;
```

### DIFF
--- a/modules/core/src/merge.cpp
+++ b/modules/core/src/merge.cpp
@@ -305,6 +305,8 @@ void cv::merge(const Mat* mv, size_t n, OutputArray _dst)
         return;
     }
 
+    CV_IPP_RUN(allch1, ipp_merge(mv, dst, (int)n));
+
     if( !allch1 )
     {
         AutoBuffer<int> pairs(cn*2);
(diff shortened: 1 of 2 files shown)


alalek:
Please add [simple test](https://github.com/opencv/opencv/blob/3.4.5/modules/core/test/test_mat.cpp#L1834-L1889) for the issue.

eightco:
Ok I added a simple test code for this issue.

PR #13550, opened 2018-12-31
fix mingw build direct show backend

### LINKED ISSUE #13549: build failed using mingw dshow module in master

- OpenCV => 4.0 (master)
- Operating System / Platform => Windows 7
- Compiler => MinGW GCC 8.1

##### Steps to reproduce

```
cmake.exe -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -G Ninja ..
```

rdhafidh:
resolves #13549

### This pullrequest changes

MinGW sprintf is would be defined as non stdio one. So include this later after dshow.h declaration, otherwise compile failure because mingw dshow sprintf has it's own declaration that is different from the standard.

### DIFF
--- a/modules/videoio/src/cap_dshow.cpp
+++ b/modules/videoio/src/cap_dshow.cpp
@@ -101,7 +101,12 @@ Thanks to:
 
 #include <tchar.h>
 #include <stdlib.h>
+#if !defined(__MINGW32__) && !defined(__MINGW64__)
+// MinGW sprintf is would be defined as non stdio one. So 
+// include this later after dshow.h declaration, otherwise compile failure 
+// because mingw dshow sprintf has it's own declaration that is different from the standard
 #include <stdio.h>
+#endif
 #include <math.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


vpisarev:
thanks! :+1:

modules/videoio/src/cap_dshow.cpp
```diff
@@ -101,7 +101,12 @@ Thanks to:
 
 #include <tchar.h>
 #include <stdlib.h>
+#if !defined(__MINGW32__) && !defined(__MINGW64__)
+// MinGW sprintf is would be defined as non stdio one. So 
+// include this later after dshow.h declaration, otherwise compile failure 
+// because mingw dshow sprintf has it's own declaration that is different from the standard
 #include <stdio.h>
```
alalek: Mentioned header has been included by precomp.hpp before that. So this condition looks useless.

Please provide:
- error message from compiler without patch
- used version of MinGW installation (with link)

PR #13553, opened 2018-12-31
Fix improper defining of CV_XADD when using Intel C++ on Mac

### LINKED ISSUE #13552: cvdef.h improperly defines CV_XADD when using Intel C++ Compiler on Mac

- OpenCV => 3.4.5 and 4.0.1
- Operating System / Platform => macOS 10.14 Mojave
- Compiler => Intel C++ 18.0

##### Detailed description

On Mac Intel C++ has a tendency to masquerade as both GCC and CLANG. This means that for whatever reason it defines both `__GNUC__` and `__clang__` despite neither being the current compiler. 

Because `__clang__` is defined, cvdef.h triggers [this define](https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/cvdef.h#L590).

```
#define CV_XADD(addr, delta) __c11_atomic_fetch_add((_Atomic(int)*)(addr), delta, __ATOMIC_ACQ_REL)
```
`__c11_atomic_fetch_add` is only available in CLANG (not Intel C++) so everything that uses CV_XADD fails to compile.

An easy way to fix this is to append `&& !defined __INTEL_COMPILER` to [this line](https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/cvdef.h#L588).

```diff
- #  if defined __clang__ && __clang_major__ >= 3 && !defined __ANDROID__ && !defined __EMSCRIPTEN__ && !defined(__CUDACC__) 
+ #  if defined __clang__ && __clang_major__ >= 3 && !defined __ANDROID__ && !defined __EMSCRIPTEN__ && !defined(__CUDACC__) && !defined __INTEL_COMPILER
```

##### Steps to reproduce

1. Add OpenCV include file

```
```

2. Try to compile

luctowers:
resolves #13552 

### This pullrequest changes

CV_XADD will not use `__c11_atomic_fetch_add` (clang exclusive) if `__INTEL_COMPILER` is defined.

### DIFF
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -463,7 +463,7 @@ Cv64suf;
 #ifdef CV_XADD
   // allow to use user-defined macro
 #elif defined __GNUC__ || defined __clang__
-#  if defined __clang__ && __clang_major__ >= 3 && !defined __ANDROID__ && !defined __EMSCRIPTEN__ && !defined(__CUDACC__)
+#  if defined __clang__ && __clang_major__ >= 3 && !defined __ANDROID__ && !defined __EMSCRIPTEN__ && !defined(__CUDACC__)  && !defined __INTEL_COMPILER
 #    ifdef __ATOMIC_ACQ_REL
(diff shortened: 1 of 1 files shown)


alalek:
Rebased onto [3.4 branch](https://github.com/opencv/opencv/wiki/Branches).

Please don't update "master" branch of your fork until this PR merge.

luctowers:
Gotcha, sorry

PR #13556, opened 2019-01-01
FIX:     ImportError: ERROR

Fernandez-Miguel-A:
FIX: 'ImportError: ERROR: recursion is detected during loading of 'cv2' binary extensions. Check OpenCV installation.'

with minimum of changed files: 
* "opencv\modules\python\package\cv2\__init__.py".
* "opencv\modules\python\common.cmake"


alalek:
Could you provide some details about your configuration?

alalek:
On clean machine (clean python + pip install numpy) this works well:
- open `<opencv_package_dir>/samples/_winpack_run_python_sample.cmd`.

Do you have other OpenCV binaries installed?

> then open python3.5

Steps?
How is `PYTHONPATH` configured?

Did you install cv2 via 'setup.py' (only "develop" mode works)? setup.py has been removed from 4.0.1.

Debug it!
```
import sys; sys.OpenCV_LOADER_DEBUG = True;
print(sys.path)
import cv2
```

Issue #13563, opened 2019-01-02
valgrind: perf_stitching issue


alalek:
[Nightly build](http://pullrequest.opencv.org/buildbot/builders/master_valgrind-lin64-debug/builds/10139)

relates #13532

alalek:
@qchateau Could you take a look on this?

qchateau:
I'll take a look at this probably this weekend

PR #13566, opened 2019-01-02
Fixed uninitialized value 

qchateau:
Fix issue #13563.

Uninitialized conditional branching was due to the `updateGain` attribute that was not initialized in the constructor of the class declaring it.
I fixed the behavior by initializing the attibute, and removed all `setUpdateGain(true)` that were called in various locations as they are not needed anymore.

EDIT: I could not rename the `updateGain` attribute to keep source compatibility as this attribute is only protected


alalek:
[Validated](http://pullrequest.opencv.org/buildbot/builders/master_valgrind-lin64-debug/builds/10140).

Issue #9688, opened 2017-09-22
Bug in operator x+=p when x is a scalar and p a row matrix


LaurentBerger:
Hi,

My code is : 

```
    Mat x(1,1,CV_64FC1);
    Mat p(10,4,CV_64FC1);
    x.setTo(0);
    p.setTo(Scalar::all(5));
    cout << " x " << x << endl;
    cout << " p " << p << endl;
    for (int i = 0; i < p.rows; i++)
    {
        x += p.row(i);
    }
    cout<<" x "<<x<<endl;
    }

```

Results are 

```
 x [0]
 p [5, 5, 5, 5;
 5, 5, 5, 5;
 5, 5, 5, 5;
 5, 5, 5, 5;
 5, 5, 5, 5;
 5, 5, 5, 5;
 5, 5, 5, 5;
 5, 5, 5, 5;
 5, 5, 5, 5;
 5, 5, 5, 5]
 x [-6.277438562204192e+66, -6.277438562204192e+66, -6.277438562204192e+66, -6.277438562204192e+66]
```

First I don't understand my code ! I copy this code [here](https://github.com/opencv/opencv_contrib/blob/master/modules/tracking/src/PFSolver.hpp#L58-L74). I  can say _std.rows is 1 (and _std.cols is 4 ). It is my example.
Why no exception is thrown ?

berak:
a [1x1] Mat will get treated like a Scalar value inside the addition. so, if you add it to a 4d vector, the scalar value is added to each elem of the 4d vector, here is a simplified version:

    Mat x(1,1,CV_64FC1, 1.0f);
    Mat p(1,4,CV_64FC1, 5.0f);
    cout << " x " << x << endl;
    cout << " p " << p << endl;
    x += p;
    cout<<" x "<<x<<endl;

    x [1]
    p [5, 5, 5, 5]
    x [6, 6, 6, 6]

it's arguable, if that is the *expected* behaviour, but apart from that, it looks all correct.

on the other hand, your result values look quite weird, i'd expect to see [50,50,50,50] as the result (that's what i get on win or linux)

LaurentBerger:
Thanks for your answer. 
I rebuilt opencv-opencv_contrib  (3.3.0-dev) in debug and release using world VS2015. My libs are good -> i use cmake

Using your example without opencl results in debug are

 x [1]
 p [5, 5, 5, 5]
[...]

Issue #8385, opened 2017-03-14
Strange behaviour of multiply 


LaurentBerger:
##### System information (version)
- OpenCV => 3.2-dev
- Operating System / Platform => Windows 10 64 Bit
- Compiler => Visual Studio 2015

##### Detailed description

My program is 

```
        Mat x(1, 2, CV_8UC1, Scalar::all(3)),y;
        cout<<x<<"\n";
        add(x, 2.0, x);
        cout << x << "\n";
        multiply(x, 2.1, y, 1, CV_32F);
        cout << y << "\n";
        multiply(x, x, x,1,CV_16UC1); // Line A to change
        cout << x << "\n";

```
results: 
[  3,   3]
[  5,   5]
[10.5, 10.5]
[42025, 42025] : why? 

Now instead of A I write
`multiply(x, x, x);`

[  3,   3]
[  5,   5]
[10.5, 10.5]
[ 25,  25] OK

instead of A I write 
`multiply(x, 2.1, x, 1, CV_32F);`

 results are
[  3,   3]
[  5,   5]
[10.5, 10.5]
[430.5, 430.5]

May be a clone is missing or an exception must be thrown when dts=src but type is changed

alalek:
Support for inplace parameters has some limitations. Current limitation works with the same size / the same type of src and dst matrix. Because any reallocation of dst will broke "inplaced" src in general.
There are only few algorithms that are inplace-safe, cvtColor has some fixes for this.
See #8352 #8341 #8149

LaurentBerger:
Is it possible to insert something like this ?
CV_Assert(src1.getObj() != dst.getObj() || type==-1);

raghav9-97:
I tried it out on OpenCV 3.3.0 and got results as follows when src=dst but type is changed to CV_32F
[ 3, 3]
[5, 5]
[10.5, 10.5]
[ 0, 0]
Is the bug fixed or it should throw an exception on encountering this situation?

aleksey-nikolaev:
Something wrong in https://github.com/opencv/opencv/blob/master/modules/core/src/matmul.cpp#L1558
This assert didn't rise even if I tried something like this `(cv::Mat_<double>)A * (cv::Mat_<int>)B`. In this case exception from cv::hal::gemm64f was rise. But if I try `(cv::Mat_<float>)A * (cv::Mat_<int>)B` - no exception and result is an uninitialized matrix (but correct size for result of A * B)

PR #13570, opened 2019-01-03
Fix bugs in arithm_op() for InputArray (src == dst) case.

eightco:
#### Fix bugs in ```arithm_op()``` for ```InputArray```  ```(src == dst)``` case.

It is a well-known issue that ```InputArray``` ```(src == dst)``` can cause problems in opencv.
So I try to fix this issue through this PR.

The cause of this problem is the recreate of ```dst``` by call ```create()```.

There are many ways to solve this problem, but most of them are difficult solutions.
Store Mat itself inside ```Input/OutputArray```, not the pointer to it, but it would noticeably increase the overhead  and also It is possible to generate a side effect bug on any code that uses ```InputArray```.

So it is difficult to solve the problem for every part that can be ```(src == dst)``` by modifying InputArray
In most cases, we can prevent the ```(src == dst)``` case with ```CV_Assert(src1.getObj() != dst.getObj())```
However, in the case of ```Mat```'s ```+=``` operations, we must allow ```src == dst```.

Change the internal state of ```InputArray``` to solve this problem is difficult and dangerous.
So I propose to make a temporary ```Mat``` for  ```src``` on the stack when ```src == dst``` at the beginning of the ```arithm_op``` function.
I think this method can avoid this problem with small overhead and the possibility of side effects being small.

**The reasons are as follows.**

Anyway, in the ```arithm_op``` function, ```src``` will be converted via ```getMat()``` or ```getUMat()``` to be used inside the operation function.
Therefore, if we create a temporary ```Mat``` (or ```UMat```) for ```src``` at the beginning of the function, the overhead would be small.

If ```src``` is ```Mat``` (or ```UMat```), it is only a ref count increment and more use of stack memory
Otherwise It may take time to convert to ```Mat```, but in the case of ```arithm_op``` function it must be converted by ```GetMat()``` (or ```GetUMat()```) function to be used inside this function anyway. 
So this is not an additional overhead.

After the above operation, even if ```dst``` is recreated in the case of ```src == dst```, ```psrc``` is safe because it refers to the temporary ```Mat```.
This temporary ```Mat``` is created on the stack and will disappear when the function ends.

The following issues can be solved with this modification.

issue https://github.com/opencv/opencv/issues/9688
```
    Mat x(1, 1, CV_64FC1);
    Mat p(10, 4, CV_64FC1);
    x.setTo(0);
    p.setTo(Scalar::all(5));
    for (int i = 0; i < p.rows; i++)
        x += p.row(i);
    ASSERT_DOUBLE_EQ(50.0, x.ptr<double>()[0]);
    ASSERT_DOUBLE_EQ(50.0, x.ptr<double>()[1]);
    ASSERT_DOUBLE_EQ(50.0, x.ptr<double>()[2]);
    ASSERT_DOUBLE_EQ(50.0, x.ptr<double>()[3]);
```

issue https://github.com/opencv/opencv/issues/8385
```
    Mat x(1, 2, CV_8UC1, Scalar::all(3)), y;
    cv::add(x, 2.0, x);
    cv::multiply(x, x, x, 1, CV_16UC1);
    EXPECT_EQ(25, (int)x.ptr<ushort>()[0]);
    EXPECT_EQ(25, (int)x.ptr<ushort>()[1]);
```

I want to solve this problem.
Please let me know if this PR has a problem or if there are other good solutions

### DIFF
--- a/modules/core/src/arithm.cpp
+++ b/modules/core/src/arithm.cpp
@@ -603,7 +603,41 @@ static void arithm_op(InputArray _src1, InputArray _src2, OutputArray _dst,
                       InputArray _mask, int dtype, BinaryFuncC* tab, bool muldiv=false,
                       void* usrdata=0, int oclop=-1 )
 {
-    const _InputArray *psrc1 = &_src1, *psrc2 = &_src2;
+    struct InputArray_backup
+    {
+        InputArray_backup(InputArray src, OutputArray dst)
+        {
+            if (src.getObj() == dst.getObj())
+            {
+                _InputArray::KindFlag k = src.kind();
+                if (k == _InputArray::KindFlag::UMAT)
+                {
+                    tmp_umat = src.getUMat();
+                    input_arr = _InputArray(tmp_umat);
+                }
+                else
+                {
+                    tmp_mat = src.getMat();
+                    input_arr = _InputArray(tmp_mat);
+                }
+            }
+            else
+            {
+                input_arr = src;
+            }
+        }
+        Mat tmp_mat;
+        UMat tmp_umat;
+        _InputArray input_arr;
+    };
+
+    const InputArray_backup src_backup1(_src1, _dst);
+    const InputArray_backup src_backup2(_src2, _dst);
+
+    CV_Assert(src_backup1.input_arr.getObj() != _dst.getObj());
+    CV_Assert(src_backup2.input_arr.getObj() != _dst.getObj());
+
+    const _InputArray *psrc1 = &src_backup1.input_arr, *psrc2 = &src_backup2.input_arr;
     int kind1 = psrc1->kind(), kind2 = psrc2->kind();
     bool haveMask = !_mask.empty();
     bool reallocate = false;
--- a/modules/core/src/arithm.cpp
+++ b/modules/core/src/arithm.cpp
@@ -609,8 +609,7 @@ static void arithm_op(InputArray _src1, InputArray _src2, OutputArray _dst,
         {
             if (src.getObj() == dst.getObj())
             {
-                _InputArray::KindFlag k = src.kind();
-                if (k == _InputArray::KindFlag::UMAT)
+                if (src.isUMat())
                 {
                     tmp_umat = src.getUMat();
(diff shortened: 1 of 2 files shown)


alalek:
Please add some [simple test](https://github.com/opencv/opencv/blob/3.4.5/modules/core/test/test_mat.cpp#L1949-L1963) based on your code.

modules/core/src/arithm.cpp
```diff
@@ -603,7 +603,41 @@ static void arithm_op(InputArray _src1, InputArray _src2, OutputArray _dst,
...
 {
-    const _InputArray *psrc1 = &_src1, *psrc2 = &_src2;
+    struct InputArray_backup
+    {
+        InputArray_backup(InputArray src, OutputArray dst)
+        {
+            if (src.getObj() == dst.getObj())
+            {
+                _InputArray::KindFlag k = src.kind();
+                if (k == _InputArray::KindFlag::UMAT)
```
alalek: Use this method of `InputArray`:
```
bool isUMat() const
```

alalek:
Actually there is more common problem: #8352
Not sure if such common problems should be fixed via local hacks.

Using of "inplace" operations with re-allocation of destination matrix is something strange operation (make no sense, because there is no real inplace processing).
However this approach is used quite often, especially as pre-processing (cvtColor(), filtering via GaussianBlur(), etc).

eightco:
#8341 #8149 are also the same cause, so it can be fixed by applying the same method as this PR.
but your comment also makes sense.
So there is way to throw an exception if recreation happens in an inplace operation.

However, if this local hacks does not fit your standard.
It is also good to reject and close this PR.

vpisarev:
@eightco, thank you for the contribution! We discussed this PR within the core team. Arithmetic operations are very cheap, so the implicit allocation of a temporary array is not very good. Especially given that in most cases those implicit operations work correctly. It's just that those 2 reported issues are about the case when the input (and the output at the same time) matrix is mis-interpreted as a scalar, and the code behaves incorrectly. I think, we should just detect such cases and throw an error instead of implicit allocation of a temporary array.

PR #13571, opened 2019-01-03
 js: allow building with aruco 

paroj:
```
allow_multiple_commits=1
```

### DIFF
--- a/modules/js/src/core_bindings.cpp
+++ b/modules/js/src/core_bindings.cpp
@@ -79,6 +79,9 @@ using namespace cv;
 using namespace dnn;
 #endif
 
+#ifdef HAVE_OPENCV_ARUCO
+using namespace aruco;
+#endif
 
 namespace binding_utils
 {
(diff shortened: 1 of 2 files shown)


alalek:
@paroj Thank you!
Could you please rebase patch onto [3.4 branch](https://github.com/opencv/opencv/wiki/Branches)?

paroj:
this will conflict because the photo module is not wrapped in 3.4 either. should I proceed anyway?

alalek:
OK, lets leave this in master only.
Will backport with photo module later.

daubaris:
Build with `opencv_contrib` fails:
```
opencv-4.1.0/build_wasm/modules/js/bindings.cpp:101:17: error: expected namespace name
using namespace aruco;
                ^
1 error generated.
ERROR:root:compiler frontend failed to generate LLVM bitcode, halting
```

Occurs when `opencv_contrib` modules building instructions are added to cmake.

PR #13586, opened 2019-01-06
Add Operator overload for multi-channel Mat with literal constant.

eightco:
#### Add Operator override for multi-channel ```Mat``` with literal constant.

When performing ```*=``` or ```/=``` operation on a multi-channel ```Mat``` with a literal constant, it is performed normally on all channels.

However, in the case of ```+=``` ```-=```, the operation result is applied to only one channel.
The reason is that literal constants are converted to ```Scalar``` datatypes, then they only apply to one channel.

So I added operator overrides for literal constants.

#### Simple test
```
    Mat mat1(2, 2, CV_8UC3, Scalar::all(1));
    mat1 += 3;
    mat1 -= 1;
    EXPECT_EQ(3, mat1.ptr<uchar>(0)[0]);
    EXPECT_EQ(3, mat1.ptr<uchar>(0)[1]);
    EXPECT_EQ(3, mat1.ptr<uchar>(0)[2]);
    mat1 &= 5;
    mat1 |= 12;
    mat1 ^= 6;
    EXPECT_EQ(11, mat1.ptr<uchar>(1)[0]);
    EXPECT_EQ(11, mat1.ptr<uchar>(1)[1]);
    EXPECT_EQ(11, mat1.ptr<uchar>(1)[2]);
```


alalek:
I'm prefer to forbid that cases at all - detect them and emit the error for multi-channel Mats with single constant.

For example, `CV_32FC2` is widely used for operation with complex numbers. So instead of adding `1` this patch adds `1 + 1i`.

eightco:
@alalek 
Okay. I understand.  you are right.

If so, should we prohibit all operator overloading when it is a literal constant with a multi-channel Mat?
But forbidding implicit casting to Scalar is also not easy.

I think it would be nice to add operator overloading for a literal constant and assert it if it is not for 1 channel.
Or just leave this PR close.

alalek:
> I think it would be nice to add operator overloading for a literal constant and assert it if it is not for 1 channel.

Right, adding overloads for "float" / "double" / "int" arguments should help avoiding of implicit `Scalar` constructors.

In tests you can use this:
`EXPECT_ANY_THROW(mat1 += 3);`

eightco:
In other tests, ```* =``` and ```/=``` are already used for basic types and multi-channel matrices.
Would it be okay to replace ```* =```, ```/=``` with ```convertTo``` in existing test code?

modules/dnn/test/test_tf_importer.cpp
```diff
@@ -40,7 +40,7 @@ TEST(Test_TensorFlow, read_inception)
     ASSERT_TRUE(!sample.empty());
     Mat input;
     resize(sample, input, Size(224, 224));
-    input -= 128; // mean sub
+    input -= Scalar(128); // mean sub
```
alalek: @dkurt, `Scalar::all(128)` ?

BTW, Test doesn't test inference result, only loading is tested here.

dkurt: Actually, I'm not sure about mean values for this model.

alalek: `Scalar(128)` is similar to `Scalar(128, 0, 0, 0)`
`Scalar::all(128)` is `Scalar(128, 128, 128, 128)`

dkurt: Yea, I just nor sure if we really need to pass [128, 128, 128] or [0, 0, 0]. From the origin repo, they use different value:
https://github.com/petewarden/tf_ios_makefile_example/blob/880c92d2a15d65b32319d43cca5987481069b574/RunModelViewController.mm#L175L209.
So I think we can replace it even to `Scalar::all(117)`.

alalek:
Operations `+=` and `-=` are not consistent with their base "+" / "-" [operations](https://github.com/opencv/opencv/blob/3.4.5/modules/core/include/opencv2/core/mat.hpp#L3592-L3608)

Also this patch breaks compatibility, so I'm going to revert it.

Issue #13576, opened 2019-01-04
dnn: edge_detection.py throws exception


berak:
##### System information (version)
- OpenCV => 3.4.5 / 3.4.3 (all from pip)
- Operating System / Platform => ubuntu 18
- Compiler => python3

##### Detailed description

[sample from here](https://github.com/opencv/opencv/blob/master/samples/dnn/edge_detection.py) is crashing in `net.forward()` : 

```
<ipython-input-1-f9cbd62f1c7f> in <module>()
     43 print(inp.shape)
     44 net.setInput(inp)
---> 45 out = net.forward()
     46 out = out[0, 0]
     47 out = cv.resize(out, (frame.shape[1], frame.shape[0]))

error: OpenCV(3.4.3) /io/opencv/modules/dnn/misc/python/pyopencv_dnn.hpp:163: error: (-213:The function/feature is not implemented) Failed to call "forward" method in function 'forward'

```

the error is quite misleading, the `CropLayer` 's forward() method IS implemented, we can even add some `print()`  statements there, to see that it is called. 

but then, something weird happens. whatever i try,  the result is invalid here:

https://github.com/opencv/opencv/blob/79e13be04c7c1f3f6d23b8caaeaa70a4a13947b7/modules/dnn/misc/python/pyopencv_dnn.hpp#L159-L163
##### Steps to reproduce

[similar findings here](http://answers.opencv.org/question/206255/cv2error-failed-to-call-forward-method-in-function-forward/)

data used:
```
wget http://vcl.ucsd.edu/hed/hed_pretrained_bsds.caffemodel
wget https://raw.githubusercontent.com/s9xie/hed/master/examples/hed/deploy.prototxt
wget http://answers.opencv.org/upfiles/15458197561802667.jpg
```

btw, [c++ version here](https://gist.github.com/berak/1c97115b3b343f1e9440bb22cf77fc76) is all fine ;)

dkurt:
@berak, Thank you! The following PR fixes that: https://github.com/opencv/opencv/pull/13600. However it's just a workaround and it'd be better to catch and print some python's error message.

```
Python 2.7.12 (default, Dec  4 2017, 14:50:18) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> a = 6
>>> b = 3
>>> c = a / b
>>> type(a), type(b), type(c)
(<type 'int'>, <type 'int'>, <type 'int'>)
```
but
```
[...]

PR #13600, opened 2019-01-09
Fix edge_detection.py sample for Python 3

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/13576

### DIFF
--- a/samples/dnn/edge_detection.py
+++ b/samples/dnn/edge_detection.py
@@ -27,8 +27,8 @@ class CropLayer(object):
         batchSize, numChannels = inputShape[0], inputShape[1]
         height, width = targetShape[2], targetShape[3]
 
-        self.ystart = (inputShape[2] - targetShape[2]) / 2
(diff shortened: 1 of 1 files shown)


samples/dnn/edge_detection.py
```diff
@@ -27,8 +27,8 @@ def getMemoryShapes(self, inputs):
         batchSize, numChannels = inputShape[0], inputShape[1]
         height, width = targetShape[2], targetShape[3]
 
-        self.ystart = (inputShape[2] - targetShape[2]) / 2
-        self.xstart = (inputShape[3] - targetShape[3]) / 2
+        self.ystart = int((inputShape[2] - targetShape[2]) / 2)
+        self.xstart = int((inputShape[3] - targetShape[3]) / 2)
```
alalek: Use integer division instead (if inputs are integers too):
`/` => `//`

PR #13601, opened 2019-01-09
Fix the bug in case determinant of rotation matrix is -1

### LINKED ISSUE #4978: decomposeHomographyMat does not return the expected decomposition
Transferred from http://code.opencv.org/issues/4307

```
|| Philippe Bouttefroy on 2015-05-01 18:52
|| Priority: Normal
|| Category: calibration, 3d
|| Tracker: Bug
|| Difficulty: 
|| PR: 
|| Platform: Any / Any
```
## decomposeHomographyMat does not return the expected decomposition

```
Hi,

I am integrating decomposeHomographyMat in our codebase. I noticed in my unit tests that the method returns incorrect results about 10-25% of the time. I have reproduced a case in the OpenCV unit tests framework and enclose the file here (test_homography_decomp.cpp). The data used does not seem to represent any specific configuration (or at least non that I could identify). Also, I have relaxed the threshold a LOT (from 0.001 to 1) to make sure that this is not due to a "tight threshold". I also copy/paste the test results at the end of this message, FYI.

The formula used to generate the homography is (from Malis paper): H = R + t^{T}*n/d, where R is the rotation, t is the translation, n is the plane normal and d is the distance of the plane to the origin. Then H is scaled so that H(2,2) = 1.

Could you please confirm the bug and let me know when a fix is available? Please let me know if you need any further information (I did not include any platform, etc. details because you should be able to reproduce on any platform). I can also generate more failure cases if you need.

Thanks,
Philippe

-----------------------------------------------
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Test with OpenCV original data:

***** VIDEOINPUT LIBRARY - 0.1995 - TFW07 *****

OpenCV version: 3.0.0-dev
OpenCV VCS version: unknown
Build type: debug
Parallel framework: ms-concurrency
CPU features: popcnt mmx sse sse2 sse3 ssse3 sse4.1 sse4.2
OpenCL Platforms:
    NVIDIA CUDA
        dGPU: GeForce GTX 970 (OpenCL 1.1 CUDA)
Current OpenCL device:
    Type = dGPU
    Name = GeForce GTX 970
    Version = OpenCL 1.1 CUDA
    Compute units = 13
    Max work group size = 1024
    Local memory size = 47 kB 1023 B
    Max memory allocation size = 1 GB
    Double support = Yes
    Host unified memory = No
    Has AMD Blas = No
    Has AMD Fft = No
    Preferred vector width char = 1
    Preferred vector width short = 1
    Preferred vector width int = 1
    Preferred vector width double = 1
Note: Google Test filter = *Calib3d_DecomposeHomography*
[==========] Running 1 test from 1 test case.
[----------] 1 test from Calib3d_DecomposeHomography
[ RUN      ] Calib3d_DecomposeHomography.regression
[       OK ] Calib3d_DecomposeHomography.regression (0 ms)
[----------] 1 test from Calib3d_DecomposeHomography (0 ms total)

[==========] 1 test from 1 test case ran. (0 ms total)
[  PASSED  ] 1 test.

++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Test with new data (in uploaded file):

***** VIDEOINPUT LIBRARY - 0.1995 - TFW07 *****

OpenCV version: 3.0.0-dev
OpenCV VCS version: unknown
Build type: debug
Parallel framework: ms-concurrency
CPU features: popcnt mmx sse sse2 sse3 ssse3 sse4.1 sse4.2
OpenCL Platforms:
    NVIDIA CUDA
        dGPU: GeForce GTX 970 (OpenCL 1.1 CUDA)
Current OpenCL device:
    Type = dGPU
    Name = GeForce GTX 970
    Version = OpenCL 1.1 CUDA
    Compute units = 13
    Max work group size = 1024
    Local memory size = 47 kB 1023 B
    Max memory allocation size = 1 GB
    Double support = Yes
    Host unified memory = No
    Has AMD Blas = No
    Has AMD Fft = No
    Preferred vector width char = 1
    Preferred vector width short = 1
    Preferred vector width int = 1
    Preferred vector width double = 1
Note: Google Test filter = *Calib3d_DecomposeHomography*
[==========] Running 1 test from 1 test case.
[----------] 1 test from Calib3d_DecomposeHomography
[ RUN      ] Calib3d_DecomposeHomography.regression
..\..\..\modules\calib3d\test\test_homography_decomp.cpp(78): error: Value of: containsValidMotion(rotations, translations, normals)
  Actual: false
Expected: true
[  FAILED  ] Calib3d_DecomposeHomography.regression (0 ms)
[----------] 1 test from Calib3d_DecomposeHomography (1 ms total)

[==========] 1 test from 1 test case ran. (2 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] Calib3d_DecomposeHomography.regression
```
##### Philippe Bouttefroy on 2015-05-02 01:31

```
The data that I provided is actually not consistent - I provided a homography with the translation denormalized but provided a normalized translation. The data below is consistent now:

        _K = Matx33d(1.0,   0.0,    0.0,
                     0.0,   1.0,    0.0,
                     0.0,   0.0,    1.0);

        _H = Matx33d(-0.102896, 0.270191,   -0.0031153,
                     0.0406387, 1.19569,    -0.0120456,
                     0.445351,  0.0410889,  1);

        //expected solution for the given homography and intrinsic matrices
        _R = Matx33d(0.692477,      0.657668,   -0.29656,
                     -0.615501,     0.753003,   0.232688,
                     0.376343,      0.0214022,  0.926233);

        _t = Vec3d(0.934629,
                   -0.775506,
                   -0.0368358);
        _n = Vec3d(-0.841608,
                   -0.43925,
                   0.314254);

        Matx33d h = _R + (_t * _n.t());
        double norm = h(2, 2);
        h = (1 / norm)*h;
        std::cout << h << std::endl;
```

JimZhou-001:
resolves #4978

### DIFF
--- a/modules/calib3d/src/homography_decomp.cpp
+++ b/modules/calib3d/src/homography_decomp.cpp
@@ -185,6 +185,10 @@ bool HomographyDecompZhang::findMotionFrom_tstar_n(const cv::Vec3d& tstar, const
     temp(1, 1) += 1.0;
     temp(2, 2) += 1.0;
     motion.R = getHnorm() * temp.inv();
+    if (cv::determinant(motion.R) < 0)
+    {
+        motion.R *= -1;
+    }
     motion.t = motion.R * tstar;
     motion.n = n;
     return passesSameSideOfPlaneConstraint(motion);
@@ -312,6 +316,10 @@ void HomographyDecompInria::findRmatFrom_tstar_n(const cv::Vec3d& tstar, const c
               0.0, 0.0, 1.0);
 
     R = getHnorm() * (I - (2/v) * tstar_m * n_m.t() );
+    if (cv::determinant(R) < 0)
+    {
+        R *= -1;
+    }
 }
 
 void HomographyDecompInria::decompose(std::vector<CameraMotion>& camMotions)
--- a/modules/calib3d/test/test_homography_decomp.cpp
+++ b/modules/calib3d/test/test_homography_decomp.cpp
@@ -134,4 +134,36 @@ private:
 
 TEST(Calib3d_DecomposeHomography, regression) { CV_HomographyDecompTest test; test.safe_run(); }
 
+
+TEST(Calib3d_DecomposeHomography, issue_4978)
+{
+    Matx33d K(
+        1.0,   0.0,    0.0,
+        0.0,   1.0,    0.0,
+        0.0,   0.0,    1.0
+    );
+
+    Matx33d H(
+        -0.102896, 0.270191,   -0.0031153,
+        0.0406387, 1.19569,    -0.0120456,
+        0.445351,  0.0410889,  1
+    );
+
+    vector<Mat> rotations;
+    vector<Mat> translations;
+    vector<Mat> normals;
+
+    decomposeHomographyMat(H, K, rotations, translations, normals);
+
+    ASSERT_GT(rotations.size(), (size_t)0u);
+    for (size_t i = 0; i < rotations.size(); i++)
+    {
+        // check: det(R) = 1
+        EXPECT_TRUE(std::fabs(cv::determinant(rotations[i]) - 1.0) < 0.01)
+            << "R: det=" << cv::determinant(rotations[0]) << std::endl << rotations[i] << std::endl
+            << "T:" << std::endl << translations[i] << std::endl;
+    }
+}
+
+
 }} // namespace


alalek:
I will rebase this patch.

alalek:
Done (updated branch in your GitHub fork). Please update your local git repository (on your machine).

Could  you capture problem inputs and create [simple test](https://github.com/opencv/opencv/blob/3.4.5/modules/calib3d/test/test_fundam.cpp#L1719-L1730) for this case (to avoid future regressions)?

alalek:
Added test

JimZhou-001:
@alalek Thank you for your kind help.

I recognize that it depends on whether and how we check the result furtherly.

As we know, homography can be induced by pure rotation or planar scenes. And in your test, you set translation as vector **0**. It seems that you cionsider pure rotation. I have a doubt whether these two kinds of homographies are consistent. Since in the case of planar scene, homography strongly relies on the plane in real world. But in image mosaicing (pure rotation), the homography is (theoretically) unique for two images.

Since we are discussing camera calibration and 3D reconstruction, it seems strange that the camera center never moves. (I'm not saying your test is wrong. It's just an example to check the result. I am consulting the differences between these two homographies.)

Since your test doesn't consider translation, here I give an example. I'm so sorry I didn't know the format of test and how to write code like that, so I didn't create the test. But we can discuss here with more freedom.

By the way, if the homography considers pure rotation, the estimated normal of the plane makes no sense. If the homography considers planar scene, the input of function findHomography should be coplanar points, right?

As I said at the beginning, we need to check the result, so the code before is not wrong. You just implement the algorithms of HomographyDecompInria and HomographyDecompZhang according to related references. The results of these methods should be checked by us manually. The reason why I say this is that I realize the simple check of det(**R**) > 0 (as I pulled request) is far from enough.

Suppose
    cv::Matx33d H;
    cv::Matx33d K;

    H << -0.0003348703671253786, 1.600035513782181e-05, 0.6118483492435596,
    -0.0006197791876426851, 1.176780051073813e-05, 0.7909747255691038,
    -1.578360198423081e-07, 7.725062124382988e-09, 0.000290482169257178;

    K << 2905.88, 0, 1416,
         0, 2905.88, 1064,
         0, 0, 1;

Using Zhang's method (you use Inria's as default) implemented in opencv (considering there may be difference between two intrinsic matrixes. But in this case, I set them the same), I get the following results:

Total number of solutions: 8

R =
[-0.9364946025241672, 0.3120714797227837, -0.1599664054464929;
 0.1521885764052158, 0.7726349432416537, 0.6163390963561511;
 0.3159374885396886, 0.5528531775804026, -0.771062168331385]
det(R)=
1
t=
[2.17194388281231;
 4.802225976544642;
 4.710798275733059]

R =
[1.21566372748587, -0.05669625349220311, -0.179458903335696;
 4.910668166876392, -0.04272037130885919, 0.5732406625282626;
 4.983822248521784, -0.2469789062604448, -0.8133400679460332]
det(R)=
0.00537154
t=
[0.01166667678070407;
 0.02579533418870028;
 0.02530422691718559]

R =
[1.21566372748587, -0.05669625349220311, -0.179458903335696;
 4.910668166876392, -0.04272037130885919, 0.5732406625282626;
 4.983822248521784, -0.2469789062604448, -0.8133400679460332]
det(R)=
0.00537154
t=
[-0.01166667678070407;
 -0.02579533418870028;
 -0.02530422691718559]

R =
[-0.9364946025241672, 0.3120714797227837, -0.1599664054464929;
 0.1521885764052158, 0.7726349432416537, 0.6163390963561511;
 0.3159374885396886, 0.5528531775804026, -0.771062168331385]
det(R)=
1
t=
[-2.17194388281231;
 -4.802225976544642;
 -4.710798275733059]

R =
[1.205355598255962, -0.05459292495324943, -0.1794144018133017;
 4.910986780251332, -0.03548683604340186, 0.5721812469516737;
 4.986385500086242, -0.2396315265538234, -0.8144920730750855]
det(R)=
0.00537154
t=
[0.00119940039434821;
 0.02593477835716218;
 0.02770854061742445]

R =
[0.9825332959057924, -0.07949774256686437, -0.1682510962077778;
 0.09287345319609841, -0.5740066896954943, 0.8135667408841405;
 -0.1612539739636958, -0.8149824716328578, -0.5565974548533852]
det(R)=
1
t=
[0.2232881221031006;
 4.828185803355463;
 5.158400839149721]

R =
[0.9825332959057924, -0.07949774256686437, -0.1682510962077778;
 0.09287345319609841, -0.5740066896954943, 0.8135667408841405;
 -0.1612539739636958, -0.8149824716328578, -0.5565974548533852]
det(R)=
1
t=
[-0.2232881221031006;
 -4.828185803355463;
 -5.158400839149721]

R =
[1.205355598255962, -0.05459292495324943, -0.1794144018133017;
 4.910986780251332, -0.03548683604340186, 0.5721812469516737;
 4.986385500086242, -0.2396315265538234, -0.8144920730750855]
det(R)=
0.00537154
t=
[-0.00119940039434821;
 -0.02593477835716218;
 -0.02770854061742445]

As you can see, det(**R**) may even be 0.00537154, which is larger than 0, too. Maybe it's because the homofgraphy itself is wrong, but as an example, it can reflect some facts of the code. So for more accurate results, we should write the code like this:
`    if (std::fabs(cv::determinant(motion.R) - 1) < 0.01)
    {
        //TODO
    }`

Actually, if we estimate homography from features (such as points and lines), the estimated rotation **R** and translation **t** should meet that (theoretically) all features are in front of both cameras. So we can check further. But simply choosing the pose with most features in the front is thoughtless. just for (**R**, **t**) and (**R**, -**t**), we may do it like this. In my test, there are still two sulutions to be checked after the steps above. Maybe we can check through the normal. Since the feature points are from the same plane, a line segment formed by any two of these points (theoretically) ought to be perpendicular to the normal. Then we may get the final one. I just analyze like this, but still don't have a try.

Of course, the funtion decomposeHomographyMat itself doesn't need to check by features. The above paragraph is just to discuss.

Last but not least, when I use 4 points to calculate homography and decompose it, the result (**R** and **t**) changes a lot even with one small pixel change of a point. So you provide robust methods such as RANSAC. I'm not so good at reading code. Could you tell me how you estimate homography robustly, randomly select 4 points every time and choose the best one, or estimate it based on so many points? I'd like to have a try in the former way, but don't know how to choose the best one (in detail).

Oh, could you tell me the right steps to pull request in detail (code) if I want to change it like what I said above?

Thank you so much.

PR #13608, opened 2019-01-10
DNN: Object detection sample: Fix bug with other size

allnes:
```
force_builders=Custom

test_opencl:Custom=ON
buildworker:Custom=linux-2

test_modules=dnn
test_bigdata:Custom=1
tests_filter=*
build_parallel_tests=1
```

**WIP**

### DIFF
--- a/modules/dnn/src/op_inf_engine.cpp
+++ b/modules/dnn/src/op_inf_engine.cpp
@@ -718,19 +718,33 @@ Mat infEngineBlobToMat(const InferenceEngine::Blob::Ptr& blob)
     return Mat(size, CV_32F, (void*)blob->buffer());
 }
 
-InfEngineBackendLayer::InfEngineBackendLayer(const InferenceEngine::DataPtr& output_)
-{
-    output = output_;
-}
-
 bool InfEngineBackendLayer::getMemoryShapes(const std::vector<MatShape> &inputs,
                                             const int requiredOutputs,
                                             std::vector<MatShape> &outputs,
                                             std::vector<MatShape> &internals) const
 {
-    std::vector<size_t> dims = output->dims;
-    std::vector<int> shape(dims.rbegin(), dims.rend());
-    outputs.assign(1, shape);
+    InferenceEngine::ICNNNetwork::InputShapes inShapes = t_net.getInputShapes();
+    InferenceEngine::ICNNNetwork::InputShapes::iterator itr;
+    bool equal_flag = true;
+    size_t i = 0;
+    for (itr = inShapes.begin(); itr != inShapes.end(); ++itr)
+    {
+        InferenceEngine::SizeVector currentInShape(inputs[i].begin(), inputs[i].end());
+        if (itr->second != currentInShape)
+        {
+            itr->second = currentInShape;
+            equal_flag = false;
+        }
+        i++;
+    }
+
+    if (!equal_flag)
+    {
+        InferenceEngine::CNNNetwork curr_t_net(t_net);
+        curr_t_net.reshape(inShapes);
+    }
+    std::vector<size_t> dims = t_net.getOutputsInfo()[name]->getDims();
+    outputs.push_back(MatShape(dims.begin(), dims.end()));
     return false;
 }
 
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -2600,7 +2600,7 @@ Net Net::readFromModelOptimizer(const String& xml, const String& bin)
     backendNode->net = Ptr<InfEngineBackendNet>(new InfEngineBackendNet(ieNet));
     for (auto& it : ieNet.getOutputsInfo())
     {
-        Ptr<Layer> cvLayer(new InfEngineBackendLayer(it.second));
+        Ptr<Layer> cvLayer(new InfEngineBackendLayer(ieNet));
         InferenceEngine::CNNLayerPtr ieLayer = ieNet.getLayerByName(it.first.c_str());
         CV_Assert(ieLayer);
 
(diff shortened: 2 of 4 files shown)


modules/dnn/src/op_inf_engine.cpp
```diff
@@ -545,6 +545,18 @@ void InfEngineBackendNet::initPlugin(InferenceEngine::ICNNNetwork& net)
...
+        for (itr = inShapes.begin(); itr != inShapes.end(); ++itr)
+        {
+            if(inpBlobs.find(itr->first) == inpBlobs.end())
+            {
+                CV_Error(Error::StsAssert, "BlobMap key not exist in Blob");
+            }
+            itr->second[2] = inpBlobs[itr->first]->dims()[1];
+            itr->second[3] = inpBlobs[itr->first]->dims()[0];
+        }
+        netOwner.reshape(inShapes);
```
dkurt: `netOwner` can be uninitialized in case of networks which are built in runtime (aren't loaded from `.xml` and `.bin` but from origin frameworks format). Please check `hasNetOwner` flag.

modules/dnn/src/op_inf_engine.cpp
```diff
@@ -545,6 +545,18 @@ void InfEngineBackendNet::initPlugin(InferenceEngine::ICNNNetwork& net)
...
+        InferenceEngine::ICNNNetwork::InputShapes inShapes = netOwner.getInputShapes();
+        InferenceEngine::ICNNNetwork::InputShapes::iterator itr;
+        for (itr = inShapes.begin(); itr != inShapes.end(); ++itr)
+        {
+            if(inpBlobs.find(itr->first) == inpBlobs.end())
+            {
+                CV_Error(Error::StsAssert, "BlobMap key not exist in Blob");
+            }
+            itr->second[2] = inpBlobs[itr->first]->dims()[1];
+            itr->second[3] = inpBlobs[itr->first]->dims()[0];
```
dkurt: I think it's better to replace all the dimensions. Please note that they should be reversed.

dkurt:
@allnes, thank you! It'd be nice to add some test for it. Please take a look on

https://github.com/opencv/opencv/blob/7e2ebecd52232d02e2d53f3e073c076be73e3470/modules/dnn/test/test_layers.cpp#L1016
as an example. You can use the same model but vary input sizes.

modules/dnn/src/op_inf_engine.cpp
dkurt: I think we can also check at first if we really need to call this procedure. I mean if network's shape are matched with input ones there is no need to use it.

modules/dnn/src/op_inf_engine.cpp
```diff
@@ -676,8 +676,31 @@ void InfEngineBackendNet::initPlugin(InferenceEngine::ICNNNetwork& net)
...
+                if(inpBlobs.find(itr->first) == inpBlobs.end())
+                {
+                    CV_Error(Error::StsAssert, "BlobMap key not exist in Blob");
+                }
+                InferenceEngine::SizeVector blobDims = inpBlobs[itr->first]->dims();
+                std::reverse(blobDims.begin(), blobDims.end());
+                if (!std::equal((itr->second).begin(), (itr->second).end(), blobDims.begin()))
+                {
+                    itr->second = blobDims;
+                    std::reverse(blobDims.begin(), blobDims.end());
```
dkurt: Why we need to reverse it here? It looks like `blobDims` is not used after.

modules/dnn/src/op_inf_engine.cpp
```diff
@@ -718,19 +741,53 @@ Mat infEngineBlobToMat(const InferenceEngine::Blob::Ptr& blob)
...
+            MatShape shape(dims.begin(), dims.end());
+            result_out.push_back(shape);
+        }
+        outputs.assign(result_out.begin(), result_out.end());
+    }
+    else
+    {
+        std::vector<size_t> dims = output->dims;
+        std::vector<int> shape(dims.rbegin(), dims.rend());
+        outputs.assign(1, shape);
```
dkurt: Can we remove `output` member completely and extract output shapes from CNNNetwork (calling reshape if needed).

allnes:
@dkurt please review me

modules/dnn/src/op_inf_engine.cpp
```diff
@@ -718,19 +740,44 @@ Mat infEngineBlobToMat(const InferenceEngine::Blob::Ptr& blob)
...
+        InferenceEngine::OutputsDataMap::iterator itr;
+        std::vector<MatShape> result_out;
+        for (itr = dataOut.begin(); itr != dataOut.end(); ++itr)
+        {
+            std::vector<size_t> dims = itr->second->getDims();
+            MatShape shape(dims.begin(), dims.end());
+            result_out.push_back(shape);
+        }
+        outputs.assign(result_out.begin(), result_out.end());
+    }
```
dkurt: `outputs` should contain a single output shape. We need to find it in `curr_t_net.getOutputsInfo()` by layer's name (check `name` Layer's member).

modules/dnn/src/op_inf_engine.cpp
```diff
@@ -164,6 +164,28 @@ void InfEngineBackendNet::init(int targetId)
...
+            }
+            InferenceEngine::SizeVector blobDims = inpBlobs[itr->first]->dims();
+            std::reverse(blobDims.begin(), blobDims.end());
+            if (!std::equal((itr->second).begin(), (itr->second).end(), blobDims.begin()))
+            {
+                itr->second = blobDims;
+                equal_flag = false;
+            }
+        }
+        if (!equal_flag) { cnn.reshape(inShapes); }
```
dkurt: Please check if this code is reachable.

modules/dnn/src/op_inf_engine.cpp
allnes: Where I get correctly layer's name?

dkurt: It's one of Layer class properties.

PR #13619, opened 2019-01-12
cmake: don't use LIB_SUFFIX with CMAKE_INSTALL_LIBDIR

### LINKED ISSUE #13595: An error occurred while setting LIB_SUFFIX=64

- OpenCV => :3.4
- Operating System / Platform => :CentOS-7.5 64Bit:
- Compiler => :GCC:

##### Detailed description
The RPM installation package under CENTOS was automatically added with LIB_SUFFIX=64 when use cmake. At this time, an error occurred and the shared library file directory became lib6464.
I think there is an error in line 595 of the CMakeLists.txt file, because the new version of cmake will automatically form CMAKE_INSTALL_LIBDIR according to the operating system environment, and the 64 - bit system will form lib64 without adding a LIB_SUFFIX it self

##### Steps to reproduce

cmake -DLIB_SUFFIX=64

alalek:
resolves #13595


alalek:
:+1:

Issue #11003, opened 2018-03-06
OE-11 "Adding Logging" discussion


vpisarev:
the feature request about evolution proposal [OE-11](https://github.com/opencv/opencv/wiki/OE-11.-Logging)

kinchungwong:
I have a list of basic opinions about what logging should do: https://gist.github.com/kinchungwong/ad3b3f7eaf34411f83bf9c393a925e14

I am prototyping some ways of addressing point 2 (enable/disable based on component, module, class, or source file name) and point 3 (some thread-level awareness and locality). But before I continue, I would like to confirm several things:

* glog is an example that shows that point 2 (log filtering by component) is possible, but its current approach of string-matching based on source file names (*.h, *.cc) is not satisfactory for OpenCV, because each module or component in OpenCV may span across many source files.
* point 3 (logging and threads) needs clarification. What settings need to be configurable "on per-thread basis"? How to do that globally, given that the main thread might not know the identities of the set of worker threads that will be launched to handle a parallel_for task? I interpret this requirement as providing thread-local configurability for a relevant subset of options.
* Regarding the past and future,
  * What aspects of the current OpenCV logging were found to be major pain points?
  * What aspects of glog were found to be major pain points?
  * Should glog be kept or removed? 
  * What aspects of ITT were found to be major pain points?
  * Should ITT be kept or removed?
  * OpenCV Core already has basic logging macros. Should these be kept, so as to minimize code changes needed in other modules that use them?
  * What are the restrictions regarding non-breaking of API if this task is changed to target OpenCV 4.1 instead?

alalek:
IMHO, Tracing / logging should be kept separately.
- Tracing is mostly about the performance.
- Logging is primary about errors reporting (including crash dumps) and about the current state of application:
  - missing file, cache
  - unsupported image codec
  - optimization branches are enabled/disabled
  - etc

---

ITT is just backend for performance tracing. It can be turned off during compilation / in runtime.

The similar things (backends) should be supported for logging:
- glog
- Android logging (adb logcat)
- ? Javascript logging (opencv.js)
- etc, including OpenCV's simple "stdout" logging should exist too.

OpenCV responsibility here is still "configuration" of logging topics/categories/tags/levels and probably basic formatting.
Logging backends are sinks with own formatting and with/without own filters configuration.
For example, "Network" / out of process sink support should be supported via some logging backend.

We prefer avoid "hard" bind of OpenCV logging API to "glog" only.
[...]

PR #13642, opened 2019-01-17
Logging revamp for 4.1.0 (#11003, OE-11)

kinchungwong:
does_not_resolve_yet #11003

### This pullrequest changes

Added class LogMeta and others.
Added setLogFilter, getLogFilter.
Added CV_LOGMETA.
Added CV_LOG_LOC, CV_LOG_THIS.
Modified CV_LOG_FATAL, CV_LOG_ERROR and others to use CV_LOGMETA.
Added defaultLogFilter as a default implementation.

<cut/>

### Known issues

Building with CV_LOG_STRIP_LEVEL set to (CV_LOG_LEVEL_VERBOSE + 1) will cause compile errors due to syntax errors in the following files:
- modules/core/src/ocl.cpp
- modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp

This is due to incorrect usage of the CV_LOG_VERBOSE macro, or trying to access non-existent C++ class members.
This commit does not fix these known issues; a separate PR will be submitted independently of Issue 11003.

### Progress

Checklist for core module changes
- [x] LogMeta class, and classes for pieces of metadata, such as LogLoc, LogObjInfo, LogModuleInfo
- [x] Global filter function that accept LogMeta class, and default implementation
- [x] CV_LOGMETA macro
- [x] Updated existing CV_LOG_(INFO, WARNING, etc) macros to use CV_LOGMETA
- [ ] Change backend function prototype (cv::utils::logging::internal::writeLogMessage) 
to accept LogMeta (or its data members) so that backends that can benefit from those 
data can use them programmatically without parsing from strings.
- [ ] Implement backend switchability
- [ ] Implement backend multicast
- [ ] Implement runtime configurable module-specific log level threshold
  - [ ] Add static functions to get and set threshold for specific modules
  - [ ] Update the default filter implementation to use that module-specific threshold
- [ ] Capture parallelized work thread-relationship information in cv::ParallelLoopBody executor implementations
- [ ] Implement string formatter switchability (for example fmtlib)
- [ ] Implement large matrix logging (frontend and backend work)
- [ ] Implement binary blob logging (frontend and backend work)

Checklist for required changes in every OpenCV module
- [ ] Define a macro for module name string that is visible to all code inside that module, without leaking outside
- [ ] Override (redefine) existing CV_LOG_(...) macros in each module to pass module name into LogMeta

Checklist for long-term improvement changes in every OpenCV module
- For modules that could use logging but don't have it, add some, strategically.
- For modules that use logging badly, fix those issues
- [ ] Where applicable, include LogObjInfo (CV_LOG_THIS) into LogMeta to help identify object instances
- Replace kludge, such as in stitching module

### DIFF
--- a/modules/core/include/opencv2/core/utils/logger.defines.hpp
+++ b/modules/core/include/opencv2/core/utils/logger.defines.hpp
@@ -19,4 +19,11 @@
 
 //! @}
 
+//! @addtogroup core_logging
+//! @{
+
+#define CV_LOG_LEVEL_CHECK_PARENT 128 //!< for use in scope-based filtering
+
+//! @}
+
 #endif // OPENCV_LOGGER_DEFINES_HPP
--- a/modules/core/include/opencv2/core/utils/logger.hpp
+++ b/modules/core/include/opencv2/core/utils/logger.hpp
@@ -34,12 +34,342 @@ enum LogLevel {
 #endif
 };
 
+#if 1 // TENTATIVE
+
+// Macros that expand to default declarations of special class methods
+
+// DC: Default (argumentless) constructor
+#define DECLARE_MEMBER_DEFAULTS_DC(ClassName) ClassName() = default;
+
+// CC: Copy constructor
+#define DECLARE_MEMBER_DEFAULTS_CC(ClassName) ClassName(const ClassName&) = default;
+
+// MC: Move constructor
+#define DECLARE_MEMBER_DEFAULTS_MC(ClassName) ClassName(ClassName&&) = default;
+
+// CA: Copy assignment operator
+#define DECLARE_MEMBER_DEFAULTS_CA(ClassName) ClassName& operator = (const ClassName&) = default;
+
+// MA: Move assignment operator
+#define DECLARE_MEMBER_DEFAULTS_MA(ClassName) ClassName& operator = (ClassName&&) = default;
+
+// ALL: All of above
+#define DECLARE_MEMBER_DEFAULTS_ALL(ClassName) \
+    DECLARE_MEMBER_DEFAULTS_DC(ClassName) \
+    DECLARE_MEMBER_DEFAULTS_CC(ClassName) \
+    DECLARE_MEMBER_DEFAULTS_MC(ClassName) \
+    DECLARE_MEMBER_DEFAULTS_CA(ClassName) \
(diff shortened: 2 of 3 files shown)


kinchungwong:
To prevent forced changes to all existing logging in application code, the new design must make sure existing log macros continue to work, even if their definitions are changed.

The existing log macros refer to:
- `CV_LOG_FATAL`
- `CV_LOG_ERROR`
- `CV_LOG_WARNING`
- `CV_LOG_INFO`
- `CV_LOG_DEBUG`
- `CV_LOG_VERBOSE`

However, that means new features cannot be tested on these macros. To test the new features, currently one must use the new macro, `CV_LOGMETA`.

Here are some examples of using `CV_LOGMETA`. Line breaks and indentation are added for readability.

Example 1

```cpp
CV_LOGMETA(
    NULL, 
    cv::utils::logging::LOG_LEVEL_DEBUG, 
    cv::utils::logging::LogModuleInfo("imgcodecs")
    << CV_LOG_LOC(),
    "findDecoder(" << filename << ")"
);
```

Example 2

```cpp
CV_LOGMETA(
    NULL, 
    cv::utils::logging::LOG_LEVEL_DEBUG, 
    cv::utils::logging::LogModuleInfo("imgcodecs") 
    << CV_LOG_LOC() 
    << CV_LOG_THIS(), 
    "Reading PNG header");
```

In these examples,
- `CV_LOG_LOC` captures the source file name, line number, and function name
- `CV_LOG_THIS` captures the type information and address of the "this" pointer, but only if inside a class non-static member function.
- LogModuleInfo captures the module name. Later, a macro will be added to each module to reduce clutter.

kinchungwong:
This document provides some of the underlying thinking about why certain design decisions are made. https://gist.github.com/kinchungwong/6d24bd3d45424197a4d1495237cd9be3

kinchungwong:
@alalek I probably need help writing a preprocessor check for GCC versions and C++ language versions (inside preprocessor code) in which `std::make_unique` is missing. `std::make_unique` is a C++14 thing, but it is available in some other compilers. I need to provide a replacement but only if there isn't one provided from the combination of compiler settings.

kinchungwong:
This PR is too far off from current design, not possible to contribute (merge) and/or improve. Closing PR.

PR #13644, opened 2019-01-17
dnn/samples: add googlenet to model zoo

berak:
it is used in [the tutorial](https://docs.opencv.org/4.0.0/d5/de7/tutorial_dnn_googlenet.html), and the mean/size/model, etc. information should be easy to find, and used with the "zoo" option.

### DIFF
--- a/samples/dnn/models.yml
+++ b/samples/dnn/models.yml
@@ -90,6 +90,18 @@ squeezenet:
   classes: "classification_classes_ILSVRC2012.txt"
   sample: "classification"
 
+# Googlenet from https://github.com/BVLC/caffe/tree/master/models/bvlc_googlenet
+googlenet:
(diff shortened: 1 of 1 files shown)


samples/dnn/models.yml
```diff
@@ -90,6 +90,18 @@ squeezenet:
   classes: "classification_classes_ILSVRC2012.txt"
   sample: "classification"
 
+# Googlenet from https://github.com/BVLC/caffe/tree/master/models/bvlc_googlenet
+googlenet:
+  model: "bvlc_googlenet.prototxt.caffemodel"
```
dkurt: `bvlc_googlenet.caffemodel` ?

berak: ofc. ;)

PR #13658, opened 2019-01-19
__shfl_up_sync with mask for CUDA >= 9

### LINKED ISSUE #13014: cuda blockScanInclusive hangs with RTX 2080

- OpenCV =>4.0.0 beta
- Cuda => 10.0
- Driver => 416.34
- GPU => RTX 2080

##### Detailed description

The example below crashes when using RTX 2080 but works with GTX 1080.

The example uses the opencv_contrib modules but the bug is in the files scan.hpp and warp_shuffle.hpp in opencv/modules/core/include/opencv2/cuda/

The crash is coming from the kernel hanging inside warpScanInclusive call to cv::cuda::device::shfl_up(iData, i) on line 191 of scan.hpp

On line 56 of warp_shuffle.hpp there is a #define for __shfl_up(x, y, z) to call __shfl_up_sync(0xFFFFFFFFU, x, y, z)

Nvidia must have changed the sync behavior with the new cards and the __shfl_up_sync fails because the mask is being passed to sync all warp threads when all warp threads are not active doe to the check on line 238 of scan.hpp

```.hpp

            if (tid < (tiNumScanThreads / OPENCV_CUDA_WARP_SIZE) )// not all threads active inside here
            {
                //grab top warp elements
                T val = s_Data[tid];
                //calculate exclusive scan and write back to shared memory
                s_Data[tid] = warpScanExclusive(val, s_Data, tid);
            }

```

Nvidia specifically says not to use 0xFFFFFFFF mask when updating legacy code.
[https://devblogs.nvidia.com/using-cuda-warp-level-primitives/](https://devblogs.nvidia.com/using-cuda-warp-level-primitives/)

I have a workaround for scan.hpp that works for me but I am not sure how to fix it to work more generally in the entire code base.

```.hpp

    template <typename T>
    __device__ T warpScanInclusive(T idata, volatile T* s_Data, unsigned int tid)
    {
    #if __CUDA_ARCH__ >= 300
        const unsigned int laneId = cv::cuda::device::Warp::laneId();

        unsigned mask = __activemask();// need to get mask for active threads

        // scan on shuffl functions
        #pragma unroll
        for (int i = 1; i <= (OPENCV_CUDA_WARP_SIZE / 2); i *= 2)
        {
            const T n = __shfl_up_sync(mask, idata, i, 32);//need to call with proper mask
            if (laneId >= i)
                  idata += n;
        }

        return idata;
    #else
        unsigned int pos = 2 * tid - (tid & (OPENCV_CUDA_WARP_SIZE - 1));
        s_Data[pos] = 0;
        pos += OPENCV_CUDA_WARP_SIZE;
        s_Data[pos] = idata;

        s_Data[pos] += s_Data[pos - 1];
        s_Data[pos] += s_Data[pos - 2];
        s_Data[pos] += s_Data[pos - 4];
        s_Data[pos] += s_Data[pos - 8];
        s_Data[pos] += s_Data[pos - 16];

        return s_Data[pos];
    #endif
    }

```

##### Steps to reproduce

Running the following code runs the cuda kernel that hangs when using a RTX 2080.

```.cpp

#include "opencv2\highgui.hpp"
#include "opencv2\cudaimgproc.hpp"

{
   cv::Ptr<cv::cuda::CLAHE> clahe = cv::cuda::createCLAHE();

   const cv::Size size(640, 480);

   cv::cuda::GpuMat srcDevice(size, CV_8UC1);
   cv::cuda::GpuMat dstDevice(size, CV_8UC1);

   cv::cuda::HostMem srcHost(size, CV_8UC1);
   cv::cuda::HostMem dstHost(size, CV_8UC1);

   cv::randu(srcHost, cv::Scalar::all(0.0), cv::Scalar::all(255.0));
   srcDevice.upload(srcHost);

   clahe->apply(srcDevice, dstDevice);

   dstDevice.download(dstHost);

   cv::imshow("src", srcHost);
   cv::imshow("dst", dstHost);

   cv::waitKey();

   return 0;
}

```

nglee:
For warp shuffle functions introduced from CUDA 9, giving full mask may cause the program to hang on some devices. This PR tries to fix this for `__shfl_up_sync` function. Without this PR, following tests hang with RTX 2080 Ti.

```
./bin/opencv_test_cudev --gtest_filter=*Integral*
./bin/opencv_test_cudaarithm --gtest_filter=*Integral*
./bin/opencv_test_cudaimgproc --gtest_filter=*CLAHE*
```

Resolves #13014 

```
force_builders=Custom
buildworker:Custom=linux-1
```

### DIFF
--- a/modules/cudaimgproc/src/cuda/clahe.cu
+++ b/modules/cudaimgproc/src/cuda/clahe.cu
@@ -42,15 +42,9 @@
 
 #if !defined CUDA_DISABLER
 
-#include "opencv2/core/cuda/common.hpp"
-#include "opencv2/core/cuda/functional.hpp"
-#include "opencv2/core/cuda/emulation.hpp"
-#include "opencv2/core/cuda/scan.hpp"
-#include "opencv2/core/cuda/reduce.hpp"
-#include "opencv2/core/cuda/saturate_cast.hpp"
+#include "opencv2/cudev.hpp"
 
-using namespace cv::cuda;
-using namespace cv::cuda::device;
+using namespace cv::cudev;
 
 namespace clahe
 {
@@ -73,7 +67,7 @@ namespace clahe
             for (int j = threadIdx.x; j < tileSize.x; j += blockDim.x)
             {
                 const int data = srcPtr[j];
-                Emulation::smem::atomicAdd(&smem[data], 1);
+                ::atomicAdd(&smem[data], 1);
             }
         }
 
@@ -96,7 +90,7 @@ namespace clahe
 
             // find number of overall clipped samples
 
-            reduce<256>(smem, clipped, tid, plus<int>());
+            blockReduce<256>(smem, clipped, tid, plus<int>());
 
             // broadcast evaluated value
 
@@ -128,10 +122,10 @@ namespace clahe
 
         calcLutKernel<<<grid, block, 0, stream>>>(src, lut, tileSize, tilesX, clipLimit, lutScale);
 
-        cudaSafeCall( cudaGetLastError() );
+        CV_CUDEV_SAFE_CALL( cudaGetLastError() );
 
         if (stream == 0)
-            cudaSafeCall( cudaDeviceSynchronize() );
+            CV_CUDEV_SAFE_CALL( cudaDeviceSynchronize() );
     }
 
     __global__ void transformKernel(const PtrStepSzb src, PtrStepb dst, const PtrStepb lut, const int2 tileSize, const int tilesX, const int tilesY)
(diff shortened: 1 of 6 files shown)


nglee:
Added tests for `BlockScanInclusive`, which is used by `CLAHE`, and found a bug in the function. The bug was that when the number of threads in the block was not a multiple of the warp size, the values for the last threads in the block were incorrectly calculated.

nglee:
I've tried to divide commits into meaningful units and force-pushed it.

nglee:
Following tests were performed:
```
opencv_test_cudev --gtest_filter=*BlockScan*
opencv_test_cudev --gtest_filter=*Integral*
opencv_test_cudaarithm --gtest_filter=*Integral*
opencv_test_cudaimgproc --gtest_filter=*CLAHE*
```

Verified on following test environments:
```
CUDA 10.0 on 2080 Ti
CUDA 10.0 on 1080
CUDA 8.0 on 1080
```

I think it is safe to merge now.

PR #13664, opened 2019-01-21
GAPI: Add version of descr_of function taking vector of Mat

smirnov-alexey:
This pullrequest adds descr_of function taking a vector of Mat objects

### DIFF
--- a/modules/gapi/include/opencv2/gapi/gmat.hpp
+++ b/modules/gapi/include/opencv2/gapi/gmat.hpp
@@ -133,6 +133,8 @@ static inline GMatDesc empty_gmat_desc() { return GMatDesc{-1,-1,{-1,-1}}; }
 class Mat;
 GAPI_EXPORTS GMatDesc descr_of(const cv::Mat &mat);
 GAPI_EXPORTS GMatDesc descr_of(const cv::UMat &mat);
+GAPI_EXPORTS std::vector<GMatDesc> descr_of(std::vector<cv::Mat> const& vec);
+GAPI_EXPORTS std::vector<GMatDesc> descr_of(std::vector<cv::UMat> const& vec);
 #endif // !defined(GAPI_STANDALONE)
 
 /** @} */
--- a/modules/gapi/include/opencv2/gapi/gmat.hpp
+++ b/modules/gapi/include/opencv2/gapi/gmat.hpp
@@ -133,8 +133,8 @@ static inline GMatDesc empty_gmat_desc() { return GMatDesc{-1,-1,{-1,-1}}; }
 class Mat;
 GAPI_EXPORTS GMatDesc descr_of(const cv::Mat &mat);
 GAPI_EXPORTS GMatDesc descr_of(const cv::UMat &mat);
-GAPI_EXPORTS std::vector<GMatDesc> descr_of(std::vector<cv::Mat> const& vec);
-GAPI_EXPORTS std::vector<GMatDesc> descr_of(std::vector<cv::UMat> const& vec);
+GAPI_EXPORTS std::vector<GMatDesc> descr_of(const std::vector<cv::Mat> &vec);
+GAPI_EXPORTS std::vector<GMatDesc> descr_of(const std::vector<cv::UMat> &vec);
 #endif // !defined(GAPI_STANDALONE)
 
 /** @} */
@@ -142,7 +142,7 @@ GAPI_EXPORTS std::vector<GMatDesc> descr_of(std::vector<cv::UMat> const& vec);
 namespace gapi { namespace own {
     class Mat;
     GAPI_EXPORTS GMatDesc descr_of(const Mat &mat);
-    GAPI_EXPORTS std::vector<GMatDesc> descr_of(std::vector<Mat> const& vec);
+    GAPI_EXPORTS std::vector<GMatDesc> descr_of(const std::vector<Mat> &vec);
 }}//gapi::own
 
 std::ostream& operator<<(std::ostream& os, const cv::GMatDesc &desc);
--- a/modules/gapi/include/opencv2/gapi/gmat.hpp
+++ b/modules/gapi/include/opencv2/gapi/gmat.hpp
@@ -140,6 +140,7 @@ GAPI_EXPORTS GMatDesc descr_of(const cv::UMat &mat);
 namespace gapi { namespace own {
     class Mat;
     GAPI_EXPORTS GMatDesc descr_of(const Mat &mat);
+    GAPI_EXPORTS std::vector<GMatDesc> descr_of(std::vector<Mat> const& vec);
 }}//gapi::own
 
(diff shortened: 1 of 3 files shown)


modules/gapi/test/gapi_desc_tests.cpp
```diff
@@ -40,6 +40,46 @@ TEST(GAPI_MetaDesc, MatDesc)
...
+    const auto desc2 = cv::descr_of(vec2);
+    EXPECT_EQ(CV_8U, desc2[0].depth);
+    EXPECT_EQ(3,       desc2[0].chan);
+    EXPECT_EQ(640,     desc2[0].size.width);
+    EXPECT_EQ(480,     desc2[0].size.height);
+}
+
+TEST(GAPI_MetaDesc, VecOwnMatDesc)
+{
+    using Mat = cv::gapi::own::Mat;
```
dmatveev: Oh no, please don't do so. At least name it `ownMat`

modules/gapi/include/opencv2/gapi/gmat.hpp
```diff
@@ -133,13 +133,16 @@ static inline GMatDesc empty_gmat_desc() { return GMatDesc{-1,-1,{-1,-1}}; }
...
+GAPI_EXPORTS std::vector<GMatDesc> descr_of(std::vector<cv::Mat> const& vec);
+GAPI_EXPORTS std::vector<GMatDesc> descr_of(std::vector<cv::UMat> const& vec);
 #endif // !defined(GAPI_STANDALONE)
 
 /** @} */
 
 namespace gapi { namespace own {
     class Mat;
     GAPI_EXPORTS GMatDesc descr_of(const Mat &mat);
+    GAPI_EXPORTS std::vector<GMatDesc> descr_of(std::vector<Mat> const& vec);
```
dmatveev: Did you check with Anton if overriding-in-a-namespace is ok for this purpose? W.r.t ADL and so on

modules/gapi/src/api/gmat.cpp
```diff
@@ -38,17 +38,45 @@ cv::GMatDesc cv::descr_of(const cv::Mat &mat)
 {
     return GMatDesc{mat.depth(), mat.channels(), {mat.cols, mat.rows}};
 }
+
 cv::GMatDesc cv::descr_of(const cv::UMat &mat)
 {
     return GMatDesc{ mat.depth(), mat.channels(),{ mat.cols, mat.rows } };
 }
+
+std::vector<cv::GMatDesc> cv::descr_of(std::vector<cv::Mat> const& vec)
```
dmatveev: the rest of the code is written with `const std::vector<cv::Mat> &` order, please align since now it differs

modules/gapi/src/api/gmat.cpp
```diff
@@ -38,17 +38,45 @@ cv::GMatDesc cv::descr_of(const cv::Mat &mat)
...
 
+std::vector<cv::GMatDesc> cv::gapi::own::descr_of(std::vector<cv::gapi::own::Mat> const& vec)
+{
+    std::vector<cv::GMatDesc> vec_descr;
+    for(auto& mat : vec){
+        vec_descr.emplace_back(descr_of(mat));
+    }
+    return vec_descr;
+}
+
```
dmatveev: Since this is a C++ file, I believe these three copy-pasted functions can be unified with a single template one

modules/gapi/src/api/gmat.cpp
```diff
@@ -33,22 +33,47 @@ const cv::GOrigin& cv::GMat::priv() const
     return *m_priv;
 }
 
+template <typename T> std::vector<cv::GMatDesc> vec_descr_of(const std::vector<T> &vec)
+{
+    std::vector<cv::GMatDesc> vec_descr;
```
anton-potapov: IMHO calling `vec_descr.reserve(vec.size())` will minimize number of memory allocations

modules/gapi/test/gapi_desc_tests.cpp
```diff
@@ -40,6 +40,45 @@ TEST(GAPI_MetaDesc, MatDesc)
...
+    EXPECT_EQ(1,       desc[0].chan);
+    EXPECT_EQ(320,     desc[0].size.width);
+    EXPECT_EQ(240,     desc[0].size.height);
+
+    EXPECT_EQ(CV_8U, desc[1].depth);
+    EXPECT_EQ(3,       desc[1].chan);
+    EXPECT_EQ(640,     desc[1].size.width);
+    EXPECT_EQ(480,     desc[1].size.height);
+}
+
```
anton-potapov: I would add an [ADL](https://en.cppreference.com/w/cpp/language/adl) test as well. 
I.e. something like this :

```c++
TEST(GAPI_MetaDesc, VecOwnMatDesc)
{
     cv::Mat m;
     auto desc = descr_of(m);
}
```
please note unqualified call to `descr_of`

modules/gapi/include/opencv2/gapi/gmat.hpp
```diff
@@ -133,13 +133,16 @@ static inline GMatDesc empty_gmat_desc() { return GMatDesc{-1,-1,{-1,-1}}; }
 class Mat;
 GAPI_EXPORTS GMatDesc descr_of(const cv::Mat &mat);
 GAPI_EXPORTS GMatDesc descr_of(const cv::UMat &mat);
+GAPI_EXPORTS std::vector<GMatDesc> descr_of(const std::vector<cv::Mat> &vec);
+GAPI_EXPORTS std::vector<GMatDesc> descr_of(const std::vector<cv::UMat> &vec);
 #endif // !defined(GAPI_STANDALONE)
```
dmatveev: @anton-potapov since you're here, I also wonder if this function should return `GMetaArgs` instead of `std::vector<GMatDesc>`. Was that actually supposed for DLDT use case?

modules/gapi/src/api/gmat.cpp
```diff
@@ -33,22 +33,47 @@ const cv::GOrigin& cv::GMat::priv() const
     return *m_priv;
 }
 
+template <typename T> std::vector<cv::GMatDesc> vec_descr_of(const std::vector<T> &vec)
+{
```
dmatveev: + please also make this function `static` or wrap with an anonymous `namespace`

anton-potapov: +1 for anonymous  `namespace`

modules/gapi/include/opencv2/gapi/gmat.hpp
anton-potapov: [InferenceEngine::PreprocEngine::preprocessWithGAPI](https://github.com/opencv/dldt/blob/30594bb3099b7874f372074e6c0cc82748344435/inference-engine/src/inference_engine/ie_preprocess_gapi.cpp#L335) expects  `std::vector<cv::GMetaArg>` (i.e. `GMetaArgs`) from the ad-hoc version

PR #8294, opened 2017-02-28
Fix stitching failure

alalek:
**Merge after:** #8293

Eliminate uncontrolled randomness in stitching multi-threaded code.

It still depends on initial RNG seed value, so provided results may be different. But results with same initial seed should be equal.

PR #13665, opened 2019-01-21
(2.4) flann: use OpenCV theRNG()

alalek:
std::rand() has no thread-safe guarantee.

backport commit: 147f3ebf0add6e281b5fc0a4d0c2826502ce88cd (#8294)


alalek:
:+1:

PR #13649, opened 2019-01-17
YUV/YCrCb conversions rewritten to wide intrinsics

savuor:
### This pullrequest changes

* All pixelwise conversions in `color_yuv.cpp` rewritten to wide universal intrinsics
* Platform-specific code removed
* Several YUV420 conversions merged using templates

<cut/>

### Performance

Note that although Yxx2RGB<uchar> was vectorized before, it didn't actually use vectorized code if source color space was YUV. This was because YUV coeffs didn't fit into 16 bits.
Now this problem is solved by altering the biggest coeff and by fixing the result after multiplication.

<details>
<summary>Baseline AVX2</summary>

|Name of Test|yuv initial avx2|yuv wide avx2|yuv wide avx2 vs yuv initial avx2 (x-factor)|
|---|:-:|:-:|:-:|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR2YCrCb)|0.009|0.009|0.98|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_BGR2YUV)|0.009|0.010|0.96|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_RGB2YCrCb)|0.009|0.009|0.98|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_RGB2YUV)|0.009|0.009|0.96|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_YCrCb2BGR)|0.010|0.009|1.07|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_YCrCb2RGB)|0.009|0.009|0.94|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_YUV2BGR)|0.054|0.010|5.59|
|cvtColor8u::Size_CvtMode::(127x61, COLOR_YUV2RGB)|0.054|0.009|5.81|
|cvtColor8u::Size_CvtMode::(127x61, CX_BGRA2YCrCb)|0.008|0.009|0.89|
|cvtColor8u::Size_CvtMode::(127x61, CX_BGRA2YUV)|0.008|0.010|0.83|
|cvtColor8u::Size_CvtMode::(127x61, CX_RGBA2YCrCb)|0.008|0.010|0.86|
|cvtColor8u::Size_CvtMode::(127x61, CX_RGBA2YUV)|0.008|0.010|0.84|
|cvtColor8u::Size_CvtMode::(127x61, CX_YCrCb2BGRA)|0.010|0.009|1.09|
[...]

PR #13666, opened 2019-01-21
YUV/YCrCb conversions rewritten to wide intrinsics (port to master)

savuor:
port of PR #13649

### This pullrequest changes

See the original PR description

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/imgproc/src/color_yuv.cpp
+++ b/modules/imgproc/src/color_yuv.cpp
@@ -123,7 +123,7 @@ struct RGB2YCrCb_f<float>
             }
 
             v_float32 y, cr, cb;
-            y = b*vc0 + g*vc1 + r*vc2;
+            y = v_fma(b, vc0, v_fma(g, vc1, r*vc2));
 
             if(bidx)
                 std::swap(r, b);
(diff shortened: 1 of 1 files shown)


modules/imgproc/src/color_yuv.cpp
```diff
@@ -85,193 +88,106 @@ struct RGB2YCrCb_f<float>
...
-            dst[i] = Y; dst[i+1+yuvOrder] = Cr; dst[i+2-yuvOrder] = Cb;
-        }
-    }
-    int srccn, blueIdx;
-    bool isCrCb;
-    float coeffs[5];
-    float32x4_t v_c0, v_c1, v_c2, v_c3, v_c4, v_delta;
-};
+            v_float32 y, cr, cb;
+            y = b*vc0 + g*vc1 + r*vc2;
```
terfendail: IMHO it would be better to use v_muladd() intrinsic here to ensure fusion of operation in case such opportunity is available.

savuor: Sure, this should give more precision so I'll add it
But according to my tests this has no (positive or negative) impact on performance

PR #13668, opened 2019-01-21
imgproc(test): RGB2YUV regression test

alalek:
relates #13649

Debug build is fine only (without SIMD optimizations)

```
allow_multiple_commits=1
```


savuor:
@alalek just investigated that, it's my fault: packing from s16 to u8 was done incorrectly
I've fixed that already in the port to master (PR #13666)

alalek:
@savuor Thank you :+1:

PR #13672, opened 2019-01-22
PyrDown: Fix bug #12961

### LINKED ISSUE #12961: Regression: image loading/processing on Android 32bit
- OpenCV => latest master or 3.4 branches, starting with this commit
    cc10e6b3441ccbe7344836a521fa6d4effa9a7dc

This commit does not have the regression:
    ad93ed720dfd1a69ecb60905777797e15ba7687f

So it's only a few commit between the two.

- Operating System / Platform => Android 32bit
- Compiler => NDK r18

##### Detailed description
Bus error for some specific jpg files.

Additional info: The issue is reproducible only on Android, all is fine on Intel-x86/gcc-6 and rasperry-pi3/raspian/g++-arm-linux-gnueabihf platforms.

We managed to reproduce the crash reliably with an image and the following code on Android (32bit NDK r18) compiled with -DENABLE_NEON=ON -DENABLE_VFPV3=ON

int main(int argc, char* argv[]) {
cv::Mat m = cv::imread(argv[1]);
cv::pyrDown(m, m);
return 0;}

Backtrace:
(cv::PyrDownVec_32s8u::operator()(int*, unsigned char, int, int) const+756)
(void cv::pyrDown_<cv::FixPtCast<unsigned char, 8>, cv::PyrDownVec_32s8u>(cv::Mat const&, cv::Mat&, int)+4232)
(cv::pyrDown(cv::_InputArray const&, cv::OutputArray const&, cv::Size const&, int)+660)

arnaudbrejeon:
resolves #12961 

### This pullrequest changes

Force ARM implementation to use an unaligned int pointer.

### DIFF
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -216,7 +216,7 @@ namespace cv { namespace debug_build_guard { } using namespace debug_build_guard
 #  define CV_ENABLE_UNROLLED 1
 #endif
 
-#if defined(__GNUC__) ||  defined(__clang__)
+#ifdef __GNUC__
 #  define CV_DECL_ALIGNED(x) __attribute__ ((aligned (x)))
 #elif defined _MSC_VER
 #  define CV_DECL_ALIGNED(x) __declspec(align(x))
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -216,7 +216,7 @@ namespace cv { namespace debug_build_guard { } using namespace debug_build_guard
 #  define CV_ENABLE_UNROLLED 1
 #endif
 
-#ifdef __GNUC__
+#if defined(__GNUC__) ||  defined(__clang__)
 #  define CV_DECL_ALIGNED(x) __attribute__ ((aligned (x)))
 #elif defined _MSC_VER
 #  define CV_DECL_ALIGNED(x) __declspec(align(x))
--- a/modules/imgproc/src/pyramids.cpp
+++ b/modules/imgproc/src/pyramids.cpp
@@ -112,7 +112,7 @@ struct PyrDownVec_32s8u
             v_rshr_pack_store<8>(dst + x, t0);
             x += v_uint16::nlanes;
         }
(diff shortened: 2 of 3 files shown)


modules/core/include/opencv2/core/cvdef.h
```diff
@@ -192,7 +192,7 @@ namespace cv { namespace debug_build_guard { } using namespace debug_build_guard
 #  define CV_ENABLE_UNROLLED 1
 #endif
 
-#ifdef __GNUC__
+#if defined(__GNUC__) ||  defined(__clang__)
```
alalek: As I know, Clang mimics as GCC 4.2.1, so it defines `__GNUC__` macro.

MacOSX:

```
$ clang -v
Apple LLVM version 10.0.0 (clang-1000.11.45.2)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
$ clang -x c /dev/null -dM -E | grep GNUC
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC_STDC_INLINE__ 1
#define __GNUC__ 4
```

Please, could you clarify what build configuration requires this change?

arnaudbrejeon: I didn't know clang defines __GNUC__, I will take out my change.

arnaudbrejeon:
I did a couple of quick measures on my iPad. I run 100 successive cv::pyrDown and measured the time in milliseconds. May not be the most accurate but it gives us an idea of the results.
I cannot compare directly both implementations because with the images supported by both implementations, the modified code will not run.

I tried to evaluate the speed on similar-size images.
**Current implementation:**
520x520, 1 channel image: 306 ms
521x527, 1 channel image: 356 ms

**New implementation** (is crashing  on current implementation):
521x521, 1 channel image: 333 ms

The new implementation is close to the current implementation for images of similar sizes. It makes me think the alignment code is behaving properly or at least doesn't drain performance much.
Let me know if you have other tests in mind.

alalek:
Thank you!

So, please:

PR #13687, opened 2019-01-24
G-API Fixed fluid heterogeneity support

rgarnov:
This patch fixes the issue when FluidData.internal flag incorrectly set to true for nodes which are slot nodes in terms of the IslandModel (nodes which connect different islands).
In the picture below Data1 is read by Sobel3x3 which requires 1-pixel border. Data1 is a slot node and its data will be full-image allocated by GExecutor. But without proposed fix FluidData.internal flag will be set to true for Data1 and later passes will assume what its data will be allocated as internal fluid ring buffer with border maintained and preallocated by fluid backend. In the end it will lead to a segfault as border will not be added to Data1 and Sobel3x3 will try to read border pixels.

```
    +--fluid_isl0----------------------+    +--nonfluid_isl0--+
    | Op -> Data1 -> Sobel3x3 -> Data2 | -> |   ....          |
    +---------:------------------------+    +----------:------+
              :                                        :
              :        +--fluid_isl1---------------------+
              :....... |   ....                          |
                       +---------------------------------+

```
* Changed FluidData.internal flag setting logic (now for all slot nodes it is set to false)
* Latency for all nodes with FluidData.internal == false is set to 0

### DIFF
--- a/modules/gapi/src/backends/fluid/gfluidbackend.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidbackend.cpp
@@ -930,9 +930,13 @@ namespace
                 {
                     // FIXME: ASSERT(DATA), ASSERT(FLUIDDATA)
                     auto &fd     = fg.metadata(out_data_node).get<FluidData>();
-                    fd.latency   = out_latency;
+                    // If fluid node is external, it will be bound to a real image without
+                    // fluid buffer allocation, so set its latency to 0 not to confuse later latency propagation.
+                    // Latency is used in fluid buffer allocation process and is not used by the scheduler
+                    // so latency doesn't affect the execution and setting it to 0 is legal
+                    fd.latency   = fd.internal ? out_latency : 0;
                     fd.lpi_write = fu.k.m_lpi;
-                    GModel::log(g, out_data_node, "Latency: " + std::to_string(out_latency));
+                    GModel::log(g, out_data_node, "Latency: " + std::to_string(fd.latency));
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -930,9 +930,11 @@ namespace
                 {
                     // FIXME: ASSERT(DATA), ASSERT(FLUIDDATA)
                     auto &fd     = fg.metadata(out_data_node).get<FluidData>();
-                    fd.latency   = out_latency;
+                    // If fluid node is external, it will be bound to a real image without
+                    // fluid buffer allocation, so set its latency to 0 not to confuse later latency propagation
```
dmatveev: And what if this data object is also read by its producing island? Then this data should be preserved here.

The case:
```
+-isl0--------------+
| OP -> Data1 -> OP |
+--------:----------+
         :
        +-isl1--------------+
        | OP -> Data2 -> OP :
        +-------------------+
```

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -1207,35 +1209,39 @@ void GFluidBackendImpl::addBackendPasses(ade::ExecutionEngineSetupContext &ectx)
...
+            switch (gim.metadata(nh).get<NodeKind>().k)
+            {
+            case NodeKind::ISLAND:
             {
                 const auto isl = gim.metadata(nh).get<FusedIsland>().object;
                 if (isl->backend() == cv::gapi::fluid::backend())
                 {
-                    // add FluidData to all data nodes inside island
+                    // Add FluidData to all data nodes inside island,
+                    // set internal = true if node is not a slot
```
dmatveev: "...is not a slot in terms of higher-level GIslandModel", I'd add for a better clarity.

But it is not necessary for Slot object to be island's start/end points only, (see my above comment)

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -1207,35 +1209,39 @@ void GFluidBackendImpl::addBackendPasses(ade::ExecutionEngineSetupContext &ectx)
...
-                    {
-                        if (!fg.metadata(io_node).contains<FluidData>())
-                            setFluidData(io_node, false);
-                    }
                 } // if (fluid backend)
-            } // if (ISLAND)
+            } break; // case::ISLAND
+            case NodeKind::SLOT:
+            {
+                // add FluidData to slot if it's read/written by fluid
```
dmatveev: please extend this comment with
```
// regardless if it is one fluid island (both writing to and reading from this object)
// or two distinct islands (both fluid)
```

modules/gapi/src/backends/fluid/gfluidbackend.cpp
dmatveev: Ok, I got that it seems to work

modules/gapi/src/backends/fluid/gfluidbackend.cpp
rgarnov: Done

modules/gapi/src/backends/fluid/gfluidbackend.cpp
rgarnov: Done

dmatveev:
@alalek I am not authorized to merge it :)

PR #13709, opened 2019-01-28
Add flushing to printing number of processed positive/negative samples so that user is informed of updates quicker

michal-kowalczyk:
### This pullrequest changes

Add flushing to printing number of processed positive/negative samples so that when running opencv_traincascade, the user is informed of updates quicker.
Without this change I got updates one per half of hour.

### DIFF
--- a/apps/traincascade/cascadeclassifier.cpp
+++ b/apps/traincascade/cascadeclassifier.cpp
@@ -340,7 +340,7 @@ int CvCascadeClassifier::fillPassedSamples( int first, int count, bool isPositiv
             if( predict( i ) == 1 )
             {
                 getcount++;
-                printf("%s current samples: %d\r", isPositive ? "POS":"NEG", getcount);
+                std::cout << (isPositive ? "POS" : "NEG") << " current samples: " << getcount << "\r" << std::flush;
                 break;
             }
         }
--- a/apps/traincascade/cascadeclassifier.cpp
+++ b/apps/traincascade/cascadeclassifier.cpp
@@ -340,7 +340,8 @@ int CvCascadeClassifier::fillPassedSamples( int first, int count, bool isPositiv
             if( predict( i ) == 1 )
             {
                 getcount++;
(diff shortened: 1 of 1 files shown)


mshabunin:
traincascade app is disabled on _master_ branch: https://github.com/opencv/opencv/blob/d0ba872e337516954fe431313b60154b4d484bc6/apps/CMakeLists.txt#L52

If you use version 3.4.x, you have to rebase this patch to the _3.4_ branch (see [branches](https://github.com/opencv/opencv/wiki/Branches) wiki page for instruction).

apps/traincascade/cascadeclassifier.cpp
```diff
@@ -340,7 +340,7 @@ int CvCascadeClassifier::fillPassedSamples( int first, int count, bool isPositiv
             if( predict( i ) == 1 )
             {
                 getcount++;
-                printf("%s current samples: %d\r", isPositive ? "POS":"NEG", getcount);
+                std::cout << (isPositive ? "POS" : "NEG") << " current samples: " << getcount << "\r" << std::flush;
```
alalek: Consider using this:
```
printf(...); fflush(stdout);
```

michal-kowalczyk: @alalek, I have no opportunity to test that C-style version now. Would you mind keeping C++ version here?
P.S.: thanks for resurrecting this change, I completely forgot about it.

alalek: BTW, Main concern here is to avoid changing of unrelated code (`printf()` statement).

These lines should do the same things:
```
fflush(stdout);
std::cout.flush();
std::cout << std::flush;
```

`fflush(stdout);` has been selected because it correlates with `printf()` (both are C-style).

michal-kowalczyk: I've reverted the change and use fflush as you suggested.

PR #13719, opened 2019-01-30
Fix Visual Studio 2017 compile error

philipyexushen:

### DIFF
--- a/modules/core/include/opencv2/core/cuda/detail/color_detail.hpp
+++ b/modules/core/include/opencv2/core/cuda/detail/color_detail.hpp
@@ -93,38 +93,38 @@ namespace cv { namespace cuda { namespace device
         }
 
         //constants for conversion from/to RGB and Gray, YUV, YCrCb according to BT.601
-        const float B2YF = 0.114f;
-        const float G2YF = 0.587f;
-        const float R2YF = 0.299f;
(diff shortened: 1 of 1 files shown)


mprevot:
I had a problem #13731 with the same point, however my fix (local) was to replace `__constant__` by `const` in color_detail.hpp and compilation succeeded; this seems to contradict this pull change.

mprevot:
@philipyexushen It is not clear in this context what you mean by "more rational":

> I think `__constant__` is more rational.

And what it involves in terms of code.

risa2000:
@mprevot I have just run into this when running the similar build as yours and I believe that @philipyexushen patch is the correct one.

`__constant__` is CUDA annotation which (expands into `__declspec(__constant__)`) marks "totally constant" memory on the device. So changing that to the standard C/C++ `const` will give it completely different meaning.

The problem seems with the fact that the original change which replaced inline values in the initialiser list by `const` variables seems to not be compatible with `__declspec(__constant__)` as it seems that `const` is "not constant enough" and may actually be initialized by the runtime. Making it `constexpr` just ensures that the values in the initialiser list are known at the compile time and the whole variable can be output by the compiler as `__declspec(__constant__)` at compile time as well.

Maybe the CUDA compiler got stricter error checking in the recent update.

Issue #13715, opened 2019-01-29
NC_OUT_OF_MEMORY in function 'initPlugin'


louisZL:
There are some errors when the backend is setted VPU:

E: [ncAPI] [    934946] ncGraphAllocate:1348	The graph file is bigger than the device memory
terminate called after throwing an instance of 'cv::Exception'
  what():  OpenCV(4.0.1-openvino) /home/jenkins/workspace/OpenCV/OpenVINO/build/opencv/modules/dnn/src/op_inf_engine.cpp:553: error: (-215:Assertion failed) Failed to initialize Inference Engine backend: Failed to allocate graph: NC_OUT_OF_MEMORY in function 'initPlugin'

dkurt:
@louisZL, Could you provide a reproducer please or at least specify which model is used?

louisZL:
@dkurt I installed OpenVINO R5.445 today, and my code run multi models at the same time (https://github.com/opencv/open_model_zoo/blob/2018/intel_models/index.md), like face-detection-adas-0001, person-detection-retail-0013, emotions-recognition-retail-0003 and facenet. When I run single model, it's ok.

dkurt:
@louisZL, And what is facenet? How much in bytes it takes?

louisZL:
@dkurt facenet is one face recognition model (https://github.com/davidsandberg/facenet), I use the OpenVINO model optimizer to transfer tf model. The .bin file size is about 46.9MB

louisZL:
@dkurt But when I use facenet model single, it's ok.

dkurt:
@ louisZL, Got it, thank you. Still could you please share some reference to facenet model? At least one with randomly initialized weights. And minimal code example would be also nice.

louisZL:
@dkurt I my code, I use two sub-thread, thread1 is for face detection, I use face-detection-adas-0001 model first, when I get the face rect, I use facenet to do recognition; thread2 is for person detection, I use person-detection-retail-0013 model. When I close each one thread, it runs ok. When I run two thread, got these errors.

And when I use CPU not VPU as the backend, two thread is ok.

dkurt:
> @dkurt I my code, I use two sub-thread, thread1 is for face detection, I use face-detection-adas-0001 model first, when I get the face rect, I use facenet to do recognition; thread2 is for person detection, I use person-detection-retail-0013 model.
> 
> When I close either one thread, it runs ok. When I run two thread, got these errors.
> 
[...]

Issue #13569, opened 2019-01-03
c/op_inf_engine.cpp:533: error: (-215:Assertion failed) Failed to initialize Inference Engine backend: The output blob size is not equal to the network output size: got 3 expecting 4 in function 'initPlugin'


liushuanpeng:
when i use opencv 4.0.0 and OpenVINO for intel engine opt, the error is:

**c/op_inf_engine.cpp:533: error: (-215:Assertion failed) Failed to initialize Inference Engine backend: The output blob size is not equal to the network output size: got 3 expecting 4 in function 'initPlugin'**

when i use breakpoint, i find the core function is:

**Mat prob = handler_->nets[modelIdx].forward(handler_->output_layer);**

who know the problem, please help!

liushuanpeng:
in opencv dnn,the function forward add mutex,the problem is gone.
eg:
  g_mutex.lock(); 
  Mat prob = handler_->nets[modelIdx].forward(handler_->output_layer);
  g_mutex.unlock();

who can give the certify,the function forward is not thread save?
the docment in opencv4.0.0 not find?
thanks!

liushuanpeng:
i create handler_ in every thread i used, make sure the handler_ is the same in one thread. but the problem is still exists.

alalek:
Need complete minimal reproducer:
- without external dependencies, OpenCV + OpenVINO/DLDT only
- specify exact versions of OpenCV / DLDT

Take a look on DNN module tests as reference for code/models.

dkurt:
@liushuanpeng, Do you still have this problem? If so, please provide minimal working example so we can reproduce it and either fix a bug or suggest some fixes.

dkurt:
Likely this issue similar to https://github.com/opencv/opencv/issues/13715. Both are caused by initialization of IE plugins in different threads (we keep them in the static map). I'll try to reproduce https://github.com/opencv/opencv/issues/13715 and fix it.

dkurt:
@liushuanpeng, please test the changes from https://github.com/opencv/opencv/pull/13738.

PR #13738, opened 2019-01-31
Add a mutex for shared Inference Engine plugins

dkurt:
resolves https://github.com/opencv/opencv/issues/13715
resolves https://github.com/opencv/opencv/issues/13569

```
force_builders=Custom

test_opencl:Custom=ON
buildworker:Custom=linux-2
#test_modules=dnn
```

### DIFF
--- a/modules/dnn/src/op_inf_engine.cpp
+++ b/modules/dnn/src/op_inf_engine.cpp
@@ -622,7 +622,11 @@ void InfEngineBackendNet::init(int targetId)
 
 #endif  // IE < R5
 
-static std::map<InferenceEngine::TargetDevice, InferenceEngine::InferenceEnginePluginPtr> sharedPlugins;
+static std::map<InferenceEngine::TargetDevice, InferenceEngine::InferenceEnginePluginPtr>& getSharedPlugins()
+{
+    static std::map<InferenceEngine::TargetDevice, InferenceEngine::InferenceEnginePluginPtr> sharedPlugins;
+    return sharedPlugins;
+}
 
 void InfEngineBackendNet::initPlugin(InferenceEngine::ICNNNetwork& net)
 {
@@ -630,6 +634,8 @@ void InfEngineBackendNet::initPlugin(InferenceEngine::ICNNNetwork& net)
 
     try
     {
+        AutoLock lock(getInitializationMutex());
+        auto& sharedPlugins = getSharedPlugins();
         auto pluginIt = sharedPlugins.find(targetDevice);
         if (pluginIt != sharedPlugins.end())
         {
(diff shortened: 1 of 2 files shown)


modules/dnn/src/op_inf_engine.cpp
```diff
@@ -622,6 +622,7 @@ void InfEngineBackendNet::init(int targetId)
 
 #endif  // IE < R5
 
+static Mutex sharedPluginsMutex;
 static std::map<InferenceEngine::TargetDevice, InferenceEngine::InferenceEnginePluginPtr> sharedPlugins;
```
alalek: It is better to wrap global "static" variables into static functions (to ensure initialization of them at least).

Example:
https://github.com/opencv/opencv/blob/3.4.5/modules/dnn/src/dnn.cpp#L94-L98

dkurt: @alalek, Is that Ok to use a mutex from `getInitializationMutex()`?

https://github.com/opencv/opencv/blob/e82e672a933c8bd713ff53ac9ca553ede169b566/modules/dnn/src/init.cpp#L52-L57

alalek: I believe you can reuse it (declared in dnn's precomp.hpp).

modules/dnn/test/test_layers.cpp
```diff
@@ -39,6 +39,10 @@
 //
 //M*/
 
+#ifdef HAVE_INF_ENGINE
+#include <thread>
+#endif
```
alalek: Please keep "test_precomp.hpp" as the first "include" statement.

modules/dnn/src/op_inf_engine.cpp
```diff
@@ -622,14 +622,20 @@ void InfEngineBackendNet::init(int targetId)
...
+    return sharedPlugins;
+}
 
 void InfEngineBackendNet::initPlugin(InferenceEngine::ICNNNetwork& net)
 {
     CV_Assert(!isInitialized());
 
     try
     {
+        AutoLock lock(getInitializationMutex());
```
shahzadlone: perhaps const if not modified.

dkurt: @shahzadlone, thanks for review! Could you please clarify it a bit?

shahzadlone: if lock isn't being modified could it perhaps be made AutoLock const?

alalek: This change doesn't fix (or change) anything in current code.

@shahzadlone Please report about real issues only.

Avoid reporting about "const" of local variables. They are usually not required and can be safely omitted (compilers are smart enough).
It make sense to report about "const" in parameters or about "const methods".

Issue #13724, opened 2019-01-30
A bug with PriorBox layer from IE backend


dkurt:
##### System information (version)

- OpenCV => 3.4x and master branches
- Operating System / Platform => all
- Compiler => all

##### Detailed description

Fix Inference Engine backend error for object detection sample:

* Default backend
```
python ../samples/dnn/object_detection.py ssd_caffe --width 500 --height 375 --backend 3 --input example.jpg
```
![out](https://user-images.githubusercontent.com/25801568/51986405-3b0c8b80-24b1-11e9-9676-9636550ea53e.jpg)

* IE backend
```
python ../samples/dnn/object_detection.py ssd_caffe --width 500 --height 375 --backend 2 --input example.jpg
```
causes
```
Traceback (most recent call last):
  File "../samples/dnn/object_detection.py", line 205, in <module>
    outs = net.forward(outNames)
cv2.error: OpenCV(3.4.5-dev) /home/dkurt/opencv/modules/dnn/src/layers/prior_box_layer.cpp:520: error: (-215:Assertion failed) _stepX == _stepY in function 'initInfEngine'
```

The problem starts from https://github.com/opencv/opencv/pull/13539. We need to specify both `step_h` and `step_w`.

Image:
![004545](https://user-images.githubusercontent.com/25801568/51986599-9f2f4f80-24b1-11e9-8886-f4483488e7dc.jpg)

PR #13746, opened 2019-02-01
DNN: IE Backend: Not work with other width-height

allnes:
Issue: [https://github.com/opencv/opencv/issues/13724](https://github.com/opencv/opencv/issues/13724)

### DIFF
--- a/modules/dnn/src/layers/prior_box_layer.cpp
+++ b/modules/dnn/src/layers/prior_box_layer.cpp
@@ -487,9 +487,7 @@ public:
         if (_explicitSizes)
         {
             InferenceEngine::Builder::PriorBoxClusteredLayer ieLayer(name);
-
-            CV_Assert(_stepX == _stepY);
-            ieLayer.setStep(_stepX);
+            ieLayer.setSteps({_stepY, _stepX});
 
             CV_CheckEQ(_offsetsX.size(), (size_t)1, ""); CV_CheckEQ(_offsetsY.size(), (size_t)1, ""); CV_CheckEQ(_offsetsX[0], _offsetsY[0], "");
             ieLayer.setOffset(_offsetsX[0]);
@@ -516,9 +514,6 @@ public:
             if (_maxSize > 0)
                 ieLayer.setMaxSize(_maxSize);
 
-            CV_Assert(_stepX == _stepY);
-            ieLayer.setStep(_stepX);
-
             CV_CheckEQ(_offsetsX.size(), (size_t)1, ""); CV_CheckEQ(_offsetsY.size(), (size_t)1, ""); CV_CheckEQ(_offsetsX[0], _offsetsY[0], "");
             ieLayer.setOffset(_offsetsX[0]);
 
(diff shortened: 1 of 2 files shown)


modules/dnn/src/layers/prior_box_layer.cpp
```diff
@@ -487,9 +487,7 @@ class PriorBoxLayerImpl CV_FINAL : public PriorBoxLayer
         if (_explicitSizes)
         {
             InferenceEngine::Builder::PriorBoxClusteredLayer ieLayer(name);
-
-            CV_Assert(_stepX == _stepY);
-            ieLayer.setStep(_stepX);
+            ieLayer.setSteps({_stepX, _stepY});
```
dkurt: According to it's implementation, we should pass `{_stepY, _stepX}` instead:

https://github.com/opencv/dldt/blob/17e66dc5a6631d630da454506902bd7c25d4170b/inference-engine/src/inference_engine/builders/ie_prior_box_clustered_layer.cpp#L83-L89

modules/dnn/src/layers/prior_box_layer.cpp
```diff
@@ -529,6 +526,12 @@ class PriorBoxLayerImpl CV_FINAL : public PriorBoxLayer
             if (!_aspectRatios.empty())
             {
                 l.getParameters()["aspect_ratio"] = _aspectRatios;
+                if (_stepX != _stepY)
+                {
+                    l.getParameters()["step"] = 0.0;
+                    l.getParameters()["step_h"] = _stepY;
+                    l.getParameters()["step_w"] = _stepX;
+                }
```
dkurt: Why this parameters under `if (!_aspectRatios.empty())` condition but not in the `else` section of `if (_stepX == _stepY)`?

samples/dnn/object_detection.cpp
```diff
@@ -153,51 +153,39 @@ void postprocess(Mat& frame, const std::vector<Mat>& outs, Net& net)
     std::vector<int> classIds;
     std::vector<float> confidences;
     std::vector<Rect> boxes;
-    if (net.getLayer(0)->outputNameToIndex("im_info") != -1)  // Faster-RCNN or R-FCN
+    if (outLayerType == "DetectionOutput")
```
dkurt: Please rebase a branch. These changes already merged.

Could you also add some test which covers this bug?

modules/dnn/test/test_ie_models.cpp
```diff
@@ -259,5 +259,46 @@ INSTANTIATE_TEST_CASE_P(/**/,
     Combine(testing::ValuesIn(getAvailableTargets(DNN_BACKEND_INFERENCE_ENGINE)), intelModels())
 );
 
+typedef TestWithParam< tuple< Size, Target > > IEBackend_Test;
+TEST_P(IEBackend_Test, Frame_Width_Height_Test)
```
dkurt: I think that it'd be better to replace this test to [test_backends.cpp](https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_backends.cpp).

allnes:
@dkurt done fix

modules/dnn/test/test_backends.cpp
```diff
@@ -162,6 +162,18 @@ TEST_P(DNNTestNetwork, MobileNet_SSD_Caffe)
                inp, "detection_out", "", diffScores);
 }
 
+TEST_P(DNNTestNetwork, MobileNet_SSD_Caffe_Different_Width_Height)
```
dkurt: This test is failed for me locally with `DNN_TARGET_MYRIAD`. May I ask you to take a look? The rest of test are OK.

PR #13708, opened 2019-01-28
YUV42x color conversions rewritten to wide intrinsics

savuor:
### This pullrequest changes

All 4:2:0 and 4:2:2 conversions from/to YUV in `color_yuv.cpp` rewritten to wide universal intrinsics

<cut/>

### Performance

Summary, performance boost (times) compared to original time

|Baseline|RGB to YUV42x|YUV42x to RGB, small|YUV42x to RGB, big|
|---|---|---|---|
|AVX2|2-3|8-10|13-18|
|AVX512SKX|2.5-3.5|9-11|11-18|
|SSE2|1|4-6|5.5-9.5|
|SSE4.1|2.5-3.5|9-10|11-18|

Geometric mean (ms):

<details>
<summary>Baseline AVX2</summary>

|Name of Test|yuv initial avx2|yuv wide 42x avx2|yuv wide 42x avx2 vs yuv initial avx2 (x-factor)|
|---|:-:|:-:|:-:|
|CvtColor::OCL_CvtColorFixture::(640x480, (COLOR_RGB2YUV_IYUV, 3, 1))|0.458|0.177|2.60|
|CvtColor::OCL_CvtColorFixture::(640x480, (COLOR_YUV2RGB_IYUV, 1, 3))|1.779|0.103|17.25|
|CvtColor::OCL_CvtColorFixture::(640x480, (COLOR_YUV2RGB_NV12, 1, 3))|1.809|0.103|17.48|
|CvtColor::OCL_CvtColorFixture::(640x480, (COLOR_YUV2RGB_YUY2, 2, 3))|2.814|0.214|13.17|
|CvtColor::OCL_CvtColorFixture::(1280x720, (COLOR_RGB2YUV_IYUV, 3, 1))|1.387|0.526|2.64|
|CvtColor::OCL_CvtColorFixture::(1280x720, (COLOR_YUV2RGB_IYUV, 1, 3))|5.378|0.311|17.29|
|CvtColor::OCL_CvtColorFixture::(1280x720, (COLOR_YUV2RGB_NV12, 1, 3))|5.430|0.302|17.96|
|CvtColor::OCL_CvtColorFixture::(1280x720, (COLOR_YUV2RGB_YUY2, 2, 3))|8.554|0.662|12.92|
|CvtColor::OCL_CvtColorFixture::(1920x1080, (COLOR_RGB2YUV_IYUV, 3, 1))|3.161|1.203|2.63|
[...]

PR #13757, opened 2019-02-04
YUV42x color conversions rewritten to wide intrinsics (port to master)

savuor:
port of #13708 

### This pullrequest changes

See the original PR description

### DIFF
--- a/modules/imgproc/src/color_yuv.cpp
+++ b/modules/imgproc/src/color_yuv.cpp
@@ -963,22 +963,22 @@ struct YCrCb2RGB_i<ushort>
 
 ///////////////////////////////////// YUV420 -> RGB /////////////////////////////////////
 
-const int ITUR_BT_601_CY = 1220542;
-const int ITUR_BT_601_CUB = 2116026;
-const int ITUR_BT_601_CUG = -409993;
-const int ITUR_BT_601_CVG = -852492;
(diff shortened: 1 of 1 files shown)


modules/imgproc/src/color_yuv.cpp
```diff
@@ -988,49 +988,146 @@ const int ITUR_BT_601_CBV = -74448;
 //G = (1220542(Y - 16) - 852492(V - 128) - 409993(U - 128) + (1 << 19)) >> 20
 //B = (1220542(Y - 16)                  + 2116026(U - 128) + (1 << 19)) >> 20
 
+static inline void uvToRGBuv(const uchar u, const uchar v, int& ruv, int& guv, int& buv)
+{
+    int uu, vv;
+    uu = int(u) - 128;
```
shahzadlone: why not ```int uu = int(u) - 128;```

modules/imgproc/src/color_yuv.cpp
```diff
@@ -988,49 +988,146 @@ const int ITUR_BT_601_CBV = -74448;
...
+    vv = int(v) - 128;
```
shahzadlone: same as above?

alalek:
Merged here: #13763 (to avoid duplication of similar changes)

Issue #9152, opened 2017-07-13
sometimes  camshfit get the  invalid width of  return (RotatedRect) 


vvck:
##### System information (version)
- OpenCV = 3.2:
- Operating System / Platform = win 10:
- Compiler = vs2013

##### Detailed description

sometimes when you use camshift to get the rectangle of object,you may get a invaild  value of the rectangle's width,the vaule is -1.#IND. Obviously,there is some wrong with it.

when i view the source code of camshift, i find the place  that the wrong may be happen.
file:  sources\modules\video\src\camshift.cpp
line:171
code:
       double width = std::sqrt( rotate_c * inv_m00 ) * 4;

if we put a negative number in sqrt , the return must be -1.#IND.
first **inv_m00** can't be a negative number, because m00 is determined in line 150,
so the only mistake should be **rotate_c** .
i can't find out the reason can cause this  situation,but if we can judge the rotate_c is a negative number may fix it.

sovrasov:
@vvck this bug is probably hard to reproduce since nobody report it before. Could you please share a test video?

vvck:
@sovrasov sorry, I tested the example used camera directly, so I don't have the video which use to reproduce the bug. 
as you say, this bug is hard to find, I will try my best to reproduce it ,then write the video file at same time.

vvck:
@sovrasov  the video had been saved, but It's size is up to 270MB. How can i send to you?  email or other ways?

sovrasov:
You can upload it to google drive.

vvck:
@sovrasov sorry， because of the GWF, I can,t visit google driver.

sturkmen72:
maybe this will be helpful to reproduce the bug

![bug](https://user-images.githubusercontent.com/12151541/31584749-c8af30c8-b1bc-11e7-819a-5df38eb141ee.png)

```
#include "opencv2/video/tracking.hpp"
#include "opencv2/imgproc.hpp"
#include "opencv2/highgui.hpp"

#include <iostream>

using namespace cv;
using namespace std;

int main( int /*argc*/, const char** /*argv*/ )
{
    Rect trackWindow(60, 73, 33, 25);
    Mat backproj = imread("12151541/31584749-c8af30c8-b1bc-11e7-819a-5df38eb141ee.png", -1);
    RotatedRect trackBox = CamShift(backproj, trackWindow,
        TermCriteria(TermCriteria::EPS | TermCriteria::COUNT, 10, 1));
[...]

PR #13760, opened 2019-02-04
fixed incorrect formulas in the algorithm camshift

AnastasiaaSenina:
Issue #9152
In the example of the post #9152 the calculated value of the variable "rotate_c" are negative. It can not be, because the formula by which it was calculated is the sum of squares. 
screenshot_1

In the counterexample, the variable is negative and very close to zero (due to inaccurate calculations), i think,  reasonable to add to avoid such cases theshold.

### DIFF
--- a/modules/video/src/camshift.cpp
+++ b/modules/video/src/camshift.cpp
@@ -167,6 +167,8 @@ cv::RotatedRect cv::CamShift( InputArray _probImage, Rect& window,
 
     double rotate_a = cs * cs * mu20 + 2 * cs * sn * mu11 + sn * sn * mu02;
     double rotate_c = sn * sn * mu20 - 2 * cs * sn * mu11 + cs * cs * mu02;
+    rotate_a = std::max(0.0, rotate_a);  // avoid negative result due calculation numeric errors
+    rotate_c = std::max(0.0, rotate_c);  // avoid negative result due calculation numeric errors
     double length = std::sqrt( rotate_a * inv_m00 ) * 4;
     double width = std::sqrt( rotate_c * inv_m00 ) * 4;
 


modules/video/src/camshift.cpp
```diff
@@ -105,7 +105,41 @@ int cv::meanShift( InputArray _probImage, Rect& window, TermCriteria criteria )
     window = cur_rect;
     return i;
 }
+static double calculateImax( const cv::Mat& _src,  double xc, double yc, double theta )
+{
 
+    double Imax = 0;
+    double cs = cos( theta );
+    double sn = sin( theta );
```
shahzadlone: Perhaps ```sn``` & ```cs``` should be ```const```

modules/video/src/camshift.cpp
```diff
@@ -105,7 +105,41 @@ int cv::meanShift( InputArray _probImage, Rect& window, TermCriteria criteria )
...
+        }
+    }
+    return Imax;
+}
+
+static double calculateImin( const cv::Mat& _src,  double xc, double yc, double theta )
+{
+
+    double Imin = 0;
+    double cs = cos( theta );
```
shahzadlone: same here as not modified.

modules/video/src/camshift.cpp
```diff
@@ -165,10 +199,19 @@ cv::RotatedRect cv::CamShift( InputArray _probImage, Rect& window,
     double cs = cos( theta );
     double sn = sin( theta );
 
-    double rotate_a = cs * cs * mu20 + 2 * cs * sn * mu11 + sn * sn * mu02;
-    double rotate_c = sn * sn * mu20 - 2 * cs * sn * mu11 + cs * cs * mu02;
+    int m_xc = m10/m00;
```
shahzadlone: also here

alalek:
Eliminated unnecessary changes (iostream/empty lines) and as a "bugfix" this patch is rebased onto [3.4 branch](https://github.com/opencv/opencv/wiki/Branches).

PR #13764, opened 2019-02-05
Add CV_16UC1 support for cuda::CLAHE

nglee:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/573

Due to size limit of shared memory, histogram is built on
the global memory for CV_16UC1 case.

The amount of memory needed for building histogram is:

    65536 * 4byte = 256KB

and shared memory limit is 48KB typically.

Added test cases for CV_16UC1 and various clip limits.
Added perf tests for CV_16UC1 on both CPU and CUDA code.

There was also a bug in CV_8UC1 case when redistributing
"residual" clipped pixels. Adding the test case where clip
limit is 5.0 exposes this bug.

Related : https://github.com/opencv/opencv_contrib/issues/1958

Performance test on 2080 Ti and CUDA 10 :

|   Type   | Image Size | Clip Limit |  CPU | CUDA |
|:--------:|:----------:|:----------:|:----:|:----:|
| CV_8UC1  | 1280x720   |          0 | 0.72 | 0.04 |
| CV_8UC1  | 1280x720   |         40 | 0.73 | 0.04 |
| CV_8UC1  | 1920x1080  |          0 | 1.59 | 0.08 |
| CV_8UC1  | 1920x1080  |         40 | 1.60 | 0.08 |
| CV_16UC1 | 1280x720   |          0 | 2.45 | 1.08 |
| CV_16UC1 | 1280x720   |         40 | 2.92 | 1.32 |
| CV_16UC1 | 1920x1080  |          0 | 4.24 | 1.78 |
| CV_16UC1 | 1920x1080  |         40 | 4.70 | 1.98 |

```
force_builders=Custom
buildworker:Custom=linux-1
```

### DIFF
--- a/modules/core/include/opencv2/core/cuda_types.hpp
+++ b/modules/core/include/opencv2/core/cuda_types.hpp
@@ -127,10 +127,12 @@ namespace cv
         };
 
         typedef PtrStepSz<unsigned char> PtrStepSzb;
+        typedef PtrStepSz<unsigned short> PtrStepSzus;
         typedef PtrStepSz<float> PtrStepSzf;
         typedef PtrStepSz<int> PtrStepSzi;
 
         typedef PtrStep<unsigned char> PtrStepb;
+        typedef PtrStep<unsigned short> PtrStepus;
         typedef PtrStep<float> PtrStepf;
         typedef PtrStep<int> PtrStepi;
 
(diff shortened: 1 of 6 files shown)


nglee:
@alalek How do we add regression data to perf tests? The builds fail with messages such as:

```
/build/precommit_linux64/3.4/opencv/modules/ts/src/ts_perf.cpp:625: Failure
Failed
  No regression data for dst argument, test node: Sz_ClipLimit_CLAHE--CLAHE---640x480--0--8UC1-
```

alalek:
Wiki page: https://github.com/opencv/opencv/wiki/HowToUsePerfTests#how-to-update-perf-data

Or use `SANITY_CHECK_NOTHING()`: accuracy checks in performance tests are usually not reliable.

Issue #13734, opened 2019-01-31
Bug in solvePnPRansac?


srikanthmg85:
I found an issue while using solvePnPRansac with useExtrensicGuess=True with other default parameters. With default parameters solvePnPRansac uses EPNP algorithm during RANSAC iterations and refinement is done using ITERATIVE algorithm. 

When useExtrensicGuess is set to True I found that rvec and tvec obtained from EPNP is not set to the best model obtained by RANSAC. It picks the model from last iteration for refinement which is not guaranteed to be good. Because of this issue final model obtained is bad. Has anyone faced this issue?

Issue #13405, opened 2018-12-10
solvePnP fails when useExtrinsicGuess is set true?


hunterlew:
I've tested solvePnP function setting useExtrinsicGuess true because I've already got a similar solution (R|t) and I just wanted to optimize that solution using LM. But it returned totally wrong answer! 
I've checked the source code in solvePnP.cpp and found that when useExtrinsicGuess is set truth, my provided R|t would not serve as input to LM solver.

alalek:
#12334

hunterlew:
@alalek why closed? I don't think it is the same problem as #12334

Issue #12334, opened 2018-08-29
The parameter types for cv::solvePnP() is misleading


csukuangfj:
##### System information (version)

- OpenCV => master

##### Detailed description
See the code
https://github.com/opencv/opencv/blob/ee1e1ce377aa61ddea47a6c2114f99951153bb4f/modules/calib3d/include/opencv2/calib3d.hpp#L692-L695

`rvec` and `tvec` are of type `OutputArray` indicating that they are used only for ouput.

But the code
https://github.com/opencv/opencv/blob/ee1e1ce377aa61ddea47a6c2114f99951153bb4f/modules/calib3d/src/solvepnp.cpp#L58
https://github.com/opencv/opencv/blob/ee1e1ce377aa61ddea47a6c2114f99951153bb4f/modules/calib3d/src/solvepnp.cpp#L71-L79
treats them as input when `useExtrinsicGuess==true`.

##### Suggestions
The prototype should be decalared as `InputOutputArray`.

alalek:
related issue with another function: #12012

asmorkalov:
Moved to 5.x as the changes break API.

Issue #4854, opened 2015-07-27
solvePnP: is not handling all degenerated cases


opencv-pushbot:
Transferred from http://code.opencv.org/issues/3984
```
|| Alexander  Duda on 2014-10-29 10:47
|| Priority: Normal
|| Affected: 2.4.9 (latest release)
|| Category: calibration, 3d
|| Tracker: Bug
|| Difficulty: 
|| PR: 
|| Platform: Any / Any
```

solvePnP: is not handling all degenerated cases
-----------
```
Depending on the point configuration and the selected method solvePnP returns wrong solutions
(deviation > 1-e3)

1000 Tests with four fixed scene points (A;B,C,D) and random camera positions

   case 1 right angle between ABC; all scene points are coplanar:
   SOLVEPNP_ITERATIVE  - 0    wrong solutions
   SOLVEPNP_P3P        - 445  wrong solutions
   SOLVEPNP_EPNP       - 0    wrong solutions
   SOLVEPNP_DLS        - 4    wrong solutions
   SOLVEPNP_UPNP       - 1000 wrong solutions

   case 2 right angle between ABC (D is not coplanar to ABC);
   SOLVEPNP_ITERATIVE  - 634  wrong solutions
   SOLVEPNP_P3P        - 397  wrong solutions
   SOLVEPNP_EPNP       - 884  wrong solutions
   SOLVEPNP_DLS        - 0    wrong solutions
   SOLVEPNP_UPNP       - 1000 wrong solutions

   case 3 (general configuration)
   SOLVEPNP_ITERATIVE  - 701  wrong solutions
   SOLVEPNP_P3P        - 0    wrong solutions
   SOLVEPNP_EPNP       - 613  wrong solutions
   SOLVEPNP_DLS        - 0    wrong solutions
   SOLVEPNP_UPNP       - 1000 wrong solutions

   case 4 (general configuration + 2 additional points)
   SOLVEPNP_ITERATIVE  - 0    wrong solutions
   SOLVEPNP_P3P        - 
   SOLVEPNP_EPNP       - 0    wrong solutions
   SOLVEPNP_DLS        - 0    wrong solutions
   SOLVEPNP_UPNP       - 6    wrong solutions

The main problem might be a lack of documentation because P3P is only handling the main case and should only be used in combination with RANSAC and environments having only a few right angles.
```

History
-------

carstenschwede:
Any updates on this one?

PR #13768, opened 2019-02-06
Enable SOLVEPNP_DLS and fix InputOutputArray for rvec and tvec

catree:
For now, just a test to see if with:
```
CV_EXPORTS_W bool solvePnP( InputArray objectPoints, InputArray imagePoints,
                            InputArray cameraMatrix, InputArray distCoeffs,
                            OutputArray rvec, OutputArray tvec,
                            bool useExtrinsicGuess = false, int flags = SOLVEPNP_ITERATIVE );
```
and `useExtrinsicGuess = true`, provided `rvec` and `tvec` are used or not.
On my computer, they are.

Related: #13405 #13734 #12334

```
**WIP**
force_builders=linux,docs,windows,mac
```

### DIFF
--- a/modules/calib3d/src/ptsetreg.cpp
+++ b/modules/calib3d/src/ptsetreg.cpp
@@ -89,8 +89,8 @@ public:
         const float* errptr = err.ptr<float>();
         uchar* maskptr = mask.ptr<uchar>();
         float t = (float)(thresh*thresh);
-        int i, n = (int)err.total(), nz = 0;
-        for( i = 0; i < n; i++ )
+        int n = (int)err.total(), nz = 0;
+        for( int i = 0; i < n; i++ )
         {
             int f = errptr[i] <= t;
             maskptr[i] = (uchar)f;
@@ -156,7 +156,7 @@ public:
         Mat m1 = _m1.getMat(), m2 = _m2.getMat();
         Mat err, mask, model, bestModel, ms1, ms2;
 
-        int iter, niters = MAX(maxIters, 1);
+        int niters = MAX(maxIters, 1);
         int d1 = m1.channels() > 1 ? m1.channels() : m1.cols;
         int d2 = m2.channels() > 1 ? m2.channels() : m2.cols;
         int count = m1.checkVector(d1), count2 = m2.checkVector(d2), maxGoodCount = 0;
@@ -193,9 +193,8 @@ public:
             return true;
         }
 
-        for( iter = 0; iter < niters; iter++ )
+        for( int iter = 0; iter < niters; iter++ )
         {
-            int i, nmodels;
             if( count > modelPoints )
             {
                 bool found = getSubset( m1, m2, ms1, ms2, rng, 10000 );
@@ -207,13 +206,13 @@ public:
                 }
             }
 
-            nmodels = cb->runKernel( ms1, ms2, model );
+            int nmodels = cb->runKernel( ms1, ms2, model );
             if( nmodels <= 0 )
                 continue;
             CV_Assert( model.rows % nmodels == 0 );
             Size modelSize(model.cols, model.rows/nmodels);
 
-            for( i = 0; i < nmodels; i++ )
+            for( int i = 0; i < nmodels; i++ )
             {
                 Mat model_i = model.rowRange( i*modelSize.height, (i+1)*modelSize.height );
                 int goodCount = findInliers( m1, m2, model_i, err, mask, threshold );
--- a/modules/calib3d/src/solvepnp.cpp
+++ b/modules/calib3d/src/solvepnp.cpp
@@ -82,7 +82,7 @@ void drawFrameAxes(InputOutputArray image, InputArray cameraMatrix, InputArray d
 
 bool solvePnP( InputArray _opoints, InputArray _ipoints,
                InputArray _cameraMatrix, InputArray _distCoeffs,
-               OutputArray _rvec, OutputArray _tvec, bool useExtrinsicGuess, int flags )
+               InputOutputArray _rvec, InputOutputArray _tvec, bool useExtrinsicGuess, int flags )
 {
     CV_INSTRUMENT_REGION();
 
@@ -118,21 +118,13 @@ bool solvePnP( InputArray _opoints, InputArray _ipoints,
     rvec = _rvec.getMat();
     tvec = _tvec.getMat();
 
-    //TODO:
-    if (useExtrinsicGuess)
-    {
-        std::cout << "useExtrinsicGuess" << std::endl;
-        std::cout << "rvec: " << rvec.t() << std::endl;
-        std::cout << "tvec: " << tvec.t() << std::endl;
-    }
-
     Mat cameraMatrix0 = _cameraMatrix.getMat();
     Mat distCoeffs0 = _distCoeffs.getMat();
     Mat cameraMatrix = Mat_<double>(cameraMatrix0);
     Mat distCoeffs = Mat_<double>(distCoeffs0);
     bool result = false;
 
-    if (flags == SOLVEPNP_EPNP || flags == SOLVEPNP_DLS || flags == SOLVEPNP_UPNP)
+    if (flags == SOLVEPNP_EPNP || flags == SOLVEPNP_UPNP) //UPnP is broken and is remaped to EPnP
     {
         Mat undistortedPoints;
         undistortPoints(ipoints, undistortedPoints, cameraMatrix, distCoeffs);
@@ -177,20 +169,19 @@ bool solvePnP( InputArray _opoints, InputArray _ipoints,
                                      &c_rvec, &c_tvec, useExtrinsicGuess );
         result = true;
     }
-    /*else if (flags == SOLVEPNP_DLS)
+    else if (flags == SOLVEPNP_DLS)
     {
         Mat undistortedPoints;
         undistortPoints(ipoints, undistortedPoints, cameraMatrix, distCoeffs);
 
         dls PnP(opoints, undistortedPoints);
 
-        Mat R, rvec = _rvec.getMat(), tvec = _tvec.getMat();
-        bool result = PnP.compute_pose(R, tvec);
+        Mat R;
+        result = PnP.compute_pose(R, tvec);
         if (result)
             Rodrigues(R, rvec);
-        return result;
     }
-    else if (flags == SOLVEPNP_UPNP)
+    /*else if (flags == SOLVEPNP_UPNP)
     {
         upnp PnP(cameraMatrix, opoints, ipoints);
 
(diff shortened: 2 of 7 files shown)


alalek:
OutputArray / InputOutputArray difference is mostly about bindings support (Python).

modules/calib3d/src/ptsetreg.cpp
```diff
@@ -207,24 +206,21 @@ class RANSACPointSetRegistrator : public PointSetRegistrator
                 }
             }
 
-            nmodels = cb->runKernel( ms1, ms2, model );
-            if( nmodels <= 0 )
+            int success = cb->runKernel( ms1, ms2, model );
```
catree: `cb->runKernel` simply returns if `solvePnP` succeeded or not:
https://github.com/opencv/opencv/blob/fcec053d59a8e30cbac7db571fcd448bfc98dd53/modules/calib3d/src/solvepnp.cpp#L215

modules/calib3d/src/solvepnp.cpp
```diff
@@ -259,20 +255,20 @@ class PnPRansacCallback CV_FINAL : public PointSetRegistrator::Callback
...
-                        OutputArray _rvec, OutputArray _tvec, bool useExtrinsicGuess,
+                        InputOutputArray _rvec, InputOutputArray _tvec, bool useExtrinsicGuess,
                         int iterationsCount, float reprojectionError, double confidence,
                         OutputArray _inliers, int flags)
 {
     CV_INSTRUMENT_REGION();
 
     Mat opoints0 = _opoints.getMat(), ipoints0 = _ipoints.getMat();
     Mat opoints, ipoints;
-    if( opoints0.depth() == CV_64F || !opoints0.isContinuous() )
```
catree: Always use double precision. I don't think there will be performance degradation and this avoid losing precision if inputs are in double.

catree:
> OutputArray / InputOutputArray difference is mostly about bindings support (Python).

I understand now. Thanks.

---

Try to enable `SOLVEPNP_DLS` method. It passes the tests on my computer.
See #4854 where `SOLVEPNP_DLS` is working.

Change `OutputArray` to `InputOutputArray`.

Try to clean a little the code.

These are some suggestions.

catree:
Benchmark [code](https://gist.github.com/catree/62d086714ab0b0ec71c38172a3389aa1) with methodology inspired by ["Very Fast Solution to the PnP Problem with Algebraic Outlier Rejection"](http://www.iri.upc.edu/files/scidoc/1579-Very-Fast-Solution-to-the-PnP-Problem-with-Algebraic-Outlier-Rejection.pdf), see also the [EPnP IJCV paper](https://icwww.epfl.ch/~lepetit/papers/lepetit_ijcv08.pdf) and related literature.

---

Using non centered object points.

- non planar, EPnP vs DLS vs Iterative:

figure_1

figure_2

figure_1

figure_2

DLT algorithm is [sensitive to noise](https://hal.inria.fr/inria-00548311/document), that's why the Iterative method performs badly when the Gaussian image noise is high.

- non planar, EPnP + VVS vs DLS + VVS vs Iterative:

figure_1

figure_2

figure_1

figure_2

---

- planar, EPnP vs DLS vs Iterative vs IPPE:

figure_1

figure_2

figure_1

figure_2

- planar, EPnP vs EPnP + VVS vs EPnP + LM

figure_1

figure_2

- planar, DLS vs DLS + VVS vs DLS + LM

figure_1

figure_2

- planar, EPnP + VVS vs DLS + VVS vs Iterative vs IPPE + VVS

figure_1

figure_2

---

UPnP code looks really broken, non planar case:

figure_1

figure_2

---

Eigen path is used for DLS for these tests. I don't know if results are the same if OpenCV is not built with Eigen.

DLS code should be cleaned and compared with reference DLS code that can be found [here](https://github.com/urbste/MLPnP_matlab_toolbox/tree/master/dls_pnp_matlab).

VVS or LM optimization should be used to refine the `solvePnP` pose. Therefore, there is no advantage to have the DLS method and I am closing this issue.

[Related](https://github.com/opencv/opencv/pull/6864#issuecomment-473450679), at some point either DLS code should be refactored and reactivated or `SOLVEPNP_DLS` and `SOLVEPNP_UPNP` should be removed from OpenCV in my opinion.

I wish the OpenCV organisation will have the money to hire full-time software developer and computer vision engineer for these kind of tasks. The code base is big and I wish to have better code quality, refactoring, redesigning, documentation. I wish also more users would look at the OpenCV documentation and if possible not outdated documentation ;)

tobycollins:
Did you have any idea of the computation speed of VVS versus Iterative?

tobycollins:
Another point is that iterative can only return 1 pose, so I imagine you were testing the planar case in situations that were not close to ambiguous? e.g. a small plane viewed far from the camera. In those cases you should expect to see the break-down of iterative w.r.t. accuracy.

catree:
> Did you have any idea of the computation speed of VVS versus Iterative?

I will report the computation time here when I will have some time.
I would expect:

- the computation time between refineLM and refineVVS to be approximately the same (in the VVS formulation, the expensive operation should be the pseudo-inverse computation of the interaction matrix)

Let's see what practically it is.

Some interesting things in my opinion, just for the record:

- compare the basin of convergence between the VVS and LM formulation, maybe there are already some insights in the literature
- compare the needed number of iterations for convergence between VVS and LM, check that the same criteria is used in both cases
- computation time of the refine step with respect the solvePnP step

---

> Another point is that iterative can only return 1 pose, so I imagine you were testing the planar case in situations that were not close to ambiguous? e.g. a small plane viewed far from the camera. In those cases you should expect to see the break-down of iterative w.r.t. accuracy.

Not really.

My initial intention were to test the DLS to confirm or not if the current implementation is broken or not (see #6864 and #4854).

I generate camera poses with:

```
    rvec = Matx31d(randGen.uniform(-1.0, 1.0), randGen.uniform(-1.0, 1.0), randGen.uniform(-1.0, 1.0));
    tvec = Matx31d(randGen.uniform(-1.0, 1.0), randGen.uniform(-1.0, 1.0), randGen.uniform(0.5, 2.0));
```

and I just check that all points are projected in front of the camera.

To generate the 3D object points:

```
  for (int i = 0; i < nbPoints; i++) {
    points.push_back(Point3d(randGen.uniform(0.0, 2*size),
                             randGen.uniform(0.0, 2*size),
                             planar ? 0.0 : randGen.uniform(0.0, 2*size)));
  }

  if (planar) {
    Matx31d rvec(randGen.uniform(-1.0, 1.0), randGen.uniform(-1.0, 1.0), randGen.uniform(-1.0, 1.0));
    Matx31d tvec(randGen.uniform(-size, size), randGen.uniform(-size, size), randGen.uniform(-size, size));
    Matx33d R;
    Rodrigues(rvec, R);

    for (size_t i = 0; i < points.size(); i++) {
      points[i] = transform(points[i], R, tvec);
    }
  }
```

So I generate random `X` and `Y` and `Z=0` coordinates. Then, I generate a random rotation and translation vector to have planar objects in a "general configuration".

I don't check the two solutions of the IPPE method neither I compute a ratio test to see if the two solutions are ambiguous or not.

For sure it is not the ideal test for IPPE.

---

The way I generate the 3D object points seems more difficult compared to the literature. For instance for `MLPnP`, the [Matlab code](https://github.com/urbste/MLPnP_matlab_toolbox/blob/ae9c1d3cc60c225932e318ae4947d538e65fc60a/main_ordinary_3d.m#L94-L101) is:

```
        % generate 3d coordinates in camera space
        Xc= [xrand(1,npt,[-2 2]); xrand(1,npt,[-2 2]); xrand(1,npt,[4 8])];
            
        t= mean(Xc,2);
        R= rodrigues(randn(3,1));
        XXw= inv(R)*(Xc-repmat(t,1,npt));
        
        % projection
        xx= [Xc(1,:)./Xc(3,:); Xc(2,:)./Xc(3,:)]*f;
```

Object points are uniformly distributed in the camera space and translation is at the centroid of the object points.

catree:
I have fixed a small bug in my code where I was doing vector resizing followed by `push_back`. So I was adding `nb_points` same points with coordinates `(0,0,0)`, then the `nb_points` generated 3D points. Benchmark code is [here](https://gist.github.com/catree/62d086714ab0b0ec71c38172a3389aa1).

---

Using non centered object points.

  - non planar, EPnP vs DLS vs Iterative (pose errors vs number of points):

figure_1

figure_2

The more points, the better is the precision.

figure_3

DLS method is the slowest method. EPnP is the fastest method.

  - non planar, EPnP vs DLS vs Iterative (pose errors vs Gaussian Image Noise, 30 points are used):

figure_1

figure_2

DLT algorithm is sensitive to noise. The more we increase the Gaussian image noise, the less precise we are.

figure_3

DLS is slow, EPnP is fast. Iterative takes more time when the noise is high, maybe because more iterations are needed to converge?

  - non planar, EPnP vs EPnP + VVS vs EPnP + LM (pose errors vs Gaussian Image Noise, 30 points are used):

figure_1

figure_2

VVS or LM non-linear minimization allow increasing the pose precision. IJCV EPnP has a Gauss-Newton refinement that can explain the small precision gain.

figure_3

Computation time will be higher when adding a pose refinement. It looks like the LM is faster to converge when the noise is small.

  - non planar, DLS vs DLS + VVS vs DLS + LM (pose errors vs Gaussian Image Noise, 30 points are used):

figure_1

figure_2

With DLS, precision gain is higher with VVS or LM than with EPnP. At the end, difference in precision between EPnP and DLS are erased when refining the pose with VVS or LM.

figure_3

With DLS, computation time is dominated by the main DLS method.

---

  - planar, EPnP vs DLS vs Iterative vs IPPE (pose errors vs number of points):

figure_1

figure_2

EPnP performs badly in the planar case. Since the translation is badly estimated, my guess is that the EPnP method is not very efficient in the planar case rather than a pose ambiguity case.

figure_3

EPnP and IPPE are the fastest and DLS is slow.

  - planar, EPnP vs DLS vs Iterative vs IPPE (pose errors vs Gaussian Image Noise, 30 points are used):

figure_1

figure_2

Accuracy in the 2D image coordinates is really important for having good precision in the planar case. Typically for tag pose estimation, corners accuracy is really important. Increasing the number of points, for instance by using the contour points is not a guarantee to improve the pose estimation in my opinion. Also, the square shape is most likely the best configuration for planar pose estimation, see [Insights into the robustness of control point configurations for homography and planar pose estimation](https://arxiv.org/abs/1803.03025).

  - planar, DLS + VVS vs DLS + LM vs Iterative vs IPPE + VVS (pose errors vs Gaussian Image Noise, 30 points are used):

figure_1

figure_2

Pose refinement with VVS or LM is highly recommended, they should perform more or less the same.

figure_3

Similar computation times are achieved between VVS and LM.

PR #13774, opened 2019-02-07
Runing tests on IE backend 

l-bat:
### This pullrequest changes
The following test now use IE backend:    
Normalize layer
* Test_Torch_layers.net_normalize
* Test_Torch_nets.OpenFace_accuracy

Resize layer
* Test_Darknet_nets.YOLOv3
* Test_Darknet_layers.upsample
* Test_Torch_layers.upsampling_nearest  

LRN layer
* Test_Caffe_layers.LRN

<cut/>

### Performance test results 
Processor: Intel® Core™ i5-6500 CPU @ 3.20GHz × 4
Graphics: Intel® HD Graphics 530 (Skylake GT2)  

| Name of Test                            | Time before changes (ms) | Time after changes (ms) |  x-factor after vs before |
|-----------------------------------------|--------------------------|-------------------------|---------------------------|
| OpenFace::DNNTestNetwork::DLIE/CPU      | 18.298                   | 17.791                  | 0.97                      |
|  OpenFace::DNNTestNetwork::DLIE/OCL     | 8.375                    | 8.051                   | 0.96                      |
| OpenFace::DNNTestNetwork::DLIE/OCL_FP16 | 8.778                    | 7.491                   | 0.85                      |
| YOLOv3::DNNTestNetwork::DLIE/MYRIAD     | 1946.705                 | 1957.809                | 1.01                      |

```
force_builders=Custom
test_opencl:Custom=ON
buildworker:Custom=linux-2
#test_modules=dnn
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/src/layers/normalize_bbox_layer.cpp
+++ b/modules/dnn/src/layers/normalize_bbox_layer.cpp
@@ -65,16 +65,12 @@ public:
     {
         if (backendId == DNN_BACKEND_INFERENCE_ENGINE)
         {
-            if (pnorm != 2)
+            if (!(pnorm == 2 || pnorm == 1)) {
                 return false;
-            if (!blobs.empty())
-                return true;
-            if (preferableTarget == DNN_TARGET_MYRIAD)
-                return !acrossSpatial;
-            return startAxis == 1 && (!acrossSpatial || endAxis > 1);
+            }
+            return preferableTarget == DNN_TARGET_MYRIAD ? !acrossSpatial : startAxis == 1;
         }
-        else
-            return backendId == DNN_BACKEND_OPENCV;
+        return backendId == DNN_BACKEND_OPENCV;
     }
 
     bool getMemoryShapes(const std::vector<MatShape> &inputs,
--- a/modules/dnn/src/layers/normalize_bbox_layer.cpp
+++ b/modules/dnn/src/layers/normalize_bbox_layer.cpp
@@ -65,9 +65,9 @@ public:
     {
         if (backendId == DNN_BACKEND_INFERENCE_ENGINE)
         {
-            if (pnorm != 2) {
+            if (pnorm != 2)
                 return false;
-            }
+
             return preferableTarget == DNN_TARGET_MYRIAD ? !acrossSpatial : startAxis == 1;
         }
         return backendId == DNN_BACKEND_OPENCV;
--- a/modules/dnn/src/layers/normalize_bbox_layer.cpp
+++ b/modules/dnn/src/layers/normalize_bbox_layer.cpp
@@ -65,7 +65,7 @@ public:
     {
         if (backendId == DNN_BACKEND_INFERENCE_ENGINE)
         {
-            if (!(pnorm == 2 || pnorm == 1)) {
+            if (pnorm != 2) {
                 return false;
             }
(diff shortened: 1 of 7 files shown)


dkurt:
It looks like we can enable OCL_FP16 target as well:
https://github.com/opencv/opencv/blob/b068d26fad1de316de4d6357e38afcf2193f6e32/modules/dnn/test/test_torch_importer.cpp#L280-L281

As here:
https://github.com/opencv/opencv/blob/b068d26fad1de316de4d6357e38afcf2193f6e32/modules/dnn/test/test_backends.cpp#L226-L240

P.S. it looks like we need to refactor framework-specific tests and `test_backends.cpp` because there are duplicated runs.

modules/dnn/src/layers/normalize_bbox_layer.cpp
```diff
@@ -65,16 +65,12 @@ class NormalizeBBoxLayerImpl CV_FINAL : public NormalizeBBoxLayer
     {
         if (backendId == DNN_BACKEND_INFERENCE_ENGINE)
         {
-            if (pnorm != 2)
+            if (!(pnorm == 2 || pnorm == 1)) {
```
dkurt: `pnorm` is not used for IE layer creation so I think we still need to allow only `pnorm == 2` for now.

modules/dnn/test/test_torch_importer.cpp
```diff
@@ -272,12 +272,12 @@ class Test_Torch_nets : public DNNTestLayer {};
...
 TEST_P(Test_Torch_nets, OpenFace_accuracy)
 {
-#if defined(INF_ENGINE_RELEASE) && (INF_ENGINE_RELEASE < 2018030000 || INF_ENGINE_RELEASE == 2018050000)
+#if defined(INF_ENGINE_RELEASE) && (INF_ENGINE_RELEASE < 2018030000)
     if (backend == DNN_BACKEND_INFERENCE_ENGINE && target == DNN_TARGET_MYRIAD)
         throw SkipTestException("");
 #endif
     checkBackend();
-    if (backend == DNN_BACKEND_INFERENCE_ENGINE && target == DNN_TARGET_OPENCL_FP16)
+    if (backend == DNN_BACKEND_INFERENCE_ENGINE && target == DNN_TARGET_OPENCL_FP16 && INF_ENGINE_RELEASE != 2018050000)
```
dkurt: Let's keep this line unchanged. We test only the latest version (R5)

modules/dnn/src/layers/normalize_bbox_layer.cpp
```diff
@@ -65,16 +65,12 @@ class NormalizeBBoxLayerImpl CV_FINAL : public NormalizeBBoxLayer
     {
         if (backendId == DNN_BACKEND_INFERENCE_ENGINE)
         {
-            if (pnorm != 2)
+            if (pnorm != 2) {
```
dkurt: I think it's better to remove brackets.

alalek:
Is it still **WIP**?

dkurt:
@alalek, I removed it.

l-bat:
Yes, I got result with `OMP_WAIT_POLICY=PASSIVE. ` Without this x-factor = 1.04

Issue #6002, opened 2016-01-20
Runtime assertions of calib3d cv::calibrateCamera() in some cases misleading.


Lirrec:
The runtime assertions of calibrateCamera() from the calib3d module may state misleading exception messages if one of the vectors given to calibrateCamera in the parameter imagePoints is empty. 
The exception is thrown in [collectCalibrationData() Line 3055](https://github.com/Itseez/opencv/blob/f9e2b036a502309c0da5abc5e711d4cd64eb8a38/modules/calib3d/src/calibration.cpp#L3055) and will look like this

```
OpenCV Error: Unsupported format or combination of formats (imagePoints1 should contain vector of vectors of points of type Point2f) in collectCalibrationData,
 file D:/dev/MINGW-packages/mingw-w64-opencv/src/opencv-3.1.0/modules/calib3d/src/calibration.cpp, line 3063
terminate called after throwing an instance of 'cv::Exception'
  what():  D:/dev/MINGW-packages/mingw-w64-opencv/src/opencv-3.1.0/modules/calib3d/src/calibration.cpp:3063:
error: (-210) imagePoints1 should contain vector of vectors of points of type Point2f in function collectCalibrationData
```

stating that the types/content of the vector is in an invalid format, while the real problem is that one of the vectors inside imagePoints is empty, causing the checkVector in Line [3053](https://github.com/Itseez/opencv/blob/f9e2b036a502309c0da5abc5e711d4cd64eb8a38/modules/calib3d/src/calibration.cpp#L3053) to fail.

Please add a separate error case and message that states that the vector is empty, before throwing wrong assertions about its contained type. The same case may also apply to the objectPoints Parameter.

A more helpfull ( and less missleading ) error message may save someone else some precious debugging time.

Adamwgoh:
+1 On this. Experienced the same situation. Thanks for putting this issue out.

micolsi:
+1. Please, I wasted 2 hours trying to find the error. Maybe, other users could save time not searching for the type Vector.

thangktran:
I'm working on a fix for this.

PR #13803, opened 2019-02-11
calib3d: Fix calibrateCamera() misleading error.

thangktran:
objectPoints and imagePoints are not checked whether they're empty and
cause checkVector() to fail, thus result in a wrong error message.

Fixes: https://github.com/opencv/opencv/issues/6002

### This pullrequest changes

- [x] build successfully locally
- [x] `make test` successfully

### DIFF
--- a/modules/calib3d/src/calibration.cpp
+++ b/modules/calib3d/src/calibration.cpp
@@ -3141,10 +3141,17 @@ static void collectCalibrationData( InputArrayOfArrays objectPoints,
 
     for( i = 0; i < nimages; i++ )
     {
-        ni = objectPoints.getMat(i).checkVector(3, CV_32F);
+        Mat objectPoint = objectPoints.getMat(i);
+        if (objectPoint.empty())
+            CV_Error(CV_StsBadSize, "objectPoints should not contain empty vector of vectors of points");
+        ni = objectPoint.checkVector(3, CV_32F);
         if( ni <= 0 )
             CV_Error(CV_StsUnsupportedFormat, "objectPoints should contain vector of vectors of points of type Point3f");
-        int ni1 = imagePoints1.getMat(i).checkVector(2, CV_32F);
+
+        Mat imagePoint1 = imagePoints1.getMat(i);
+        if (imagePoint1.empty())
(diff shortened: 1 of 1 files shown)


modules/calib3d/src/calibration.cpp
```diff
@@ -3141,11 +3141,18 @@ static void collectCalibrationData( InputArrayOfArrays objectPoints,
 
     for( i = 0; i < nimages; i++ )
     {
-        ni = objectPoints.getMat(i).checkVector(3, CV_32F);
-        if( ni <= 0 )
+        Mat objectPoint = objectPoints.getMat(i);
+        ni = objectPoint.checkVector(3, CV_32F);
+        if ( objectPoint.empty() )
+            CV_Error(CV_StsBadSize, "objectPoints should not contain empty vector of vectors of points");
+        else if( ni <= 0 )
```
alalek: BTW,
`CV_Error()` doesn't return (throws exception), so "else if" is not necessary here.

Something like this:
```
Mat objectPoint = objectPoints.getMat(i);
if (objectPoint.empty())
    CV_Error(...)
ni = objectPoint.checkVector(3, CV_32F);
if (ni <= 0)
    CV_Error(...)
```

thangktran: @alalek I think the `else if` makes more sense.
E.g. if the `objectPoint` is indeed empty, 2 errors will be printed:
1. `OpenCV Error: The input/output structure size is incorrect (objectPoints should not contain empty vector of vectors of points) in collectCalibrationData`
2. `OpenCV Error: Unsupported format or combination of formats (objectPoints should contain vector of vectors of points of type Point3f) in collectCalibrationData`

In this case i think it's better that only error no. 1 be printed.
What is your opinion about this?

alalek: > makes more sense

I understand your point from the code.
But it is correct in case if `CV_Error()` prints message, but `CV_Error()`

> throws exception

So, there is no way to see both messages and `else if` is not required here.

thangktran: @alalek You're right. I will fix it.

thangktran:
> Thank you

@alalek Thank you.

PR #13814, opened 2019-02-12
changed the negation condition

tsowa48:
condition optimization for "img = _img.getMat()"


vpisarev:
@tsowa48, thank you for the patch, but I would not call it an optimization; it's few more CPU cycles per image whereas the whole algorithm takes much more CPU cycles per each pixel (!) of the processed image, so even for 10x10 image the saved time is less than 1%, not to mention 320x240 or larger images. So, let's not do changes that are just cosmetic

PR #13837, opened 2019-02-15
New computeECC function, and updated findTransformECC function to make gaussian filtering optional

### LINKED ISSUE #12432: Function findTransformECC uses 'optional' gaussian blurr and there is no way to compute actual value of ECC
I am working on the problem of registering images from computer tomography onto each other. I am successfully using function findTransformECC for some data sets. However there are two issues with that function:

There is no direct function to compute value of 'enhanced correlation coefficient' when given two matrices. When one implements it i.e. in the way that
```
/** Function that computes enhanced corelation coefficient (5) from Georgies et.al. 2008.
*
*/
static double computeECC(cv::Mat i1, cv::Mat i2, cv::Mat masking)
{
        cv::Scalar mean1, mean2, s1, s2;
        cv::meanStdDev(i1, mean1, s1, masking);
        cv::meanStdDev(i2, mean2, s2, masking);
        cv::Mat p1, p2;
        cv::Mat i1_zeromean = cv::Mat::zeros(i1.rows, i1.cols, i1.type());
        cv::Mat i2_zeromean = cv::Mat::zeros(i2.rows, i2.cols, i2.type());
        cv::subtract(i1, mean1, i1_zeromean, masking);
        cv::subtract(i2, mean2, i2_zeromean, masking);
        double i1norm = std::sqrt(cv::countNonZero(masking)*s1.val[0]*s1.val[0]);
        double i2norm = std::sqrt(cv::countNonZero(masking)*s2.val[0]*s2.val[0]);
        return i1_zeromean.dot(i2_zeromean)/(i1norm*i2norm);
}
```

You then realize that the ECC from this function is not the same as ECC reported by the findTransformECC function. The reason why is the implementation in which, there is

```    
//gaussian filtering is optional
    src.convertTo(templateFloat, templateFloat.type());
    GaussianBlur(templateFloat, templateFloat, Size(5, 5), 0, 0);
```

See https://github.com/opencv/opencv/blob/d74b98c3d9264f9623b27a8d4e0fc92ac97b71d0/modules/video/src/ecc.cpp#L417

It would be really great to be able to turn off GaussianBlur or specify kernel size. This one fits all aproach might not work for all the applications. I also would like to have computeECC function implemented for easy check of the value of the ECC i.e. for debugging.

Best Vojtech.

amithjkamath:
resolves #12432

### This pullrequest changes

The issue tracked in https://github.com/opencv/opencv/issues/12432 is resolved with the changes here, by adding a new function called computeECC as suggested by the issue author, and, by adding another optional argument to findTransformECC to control the size of the Gaussian filter (set to default of 5, to maintain backward compatibility; set to 1 to do no filtering, like the issue author would like). Tests and documentation strings are included as well. 

This is my first pull request, and so pardon my tardiness.

### DIFF
--- a/modules/video/include/opencv2/video/tracking.hpp
+++ b/modules/video/include/opencv2/video/tracking.hpp
@@ -268,8 +268,8 @@ enum
 /** @brief Computes the Enhanced Correlation Coefficient value between two images @cite EP08 .
 
 @param templateImage single-channel template image; CV_8U or CV_32F array.
-@param inputImage single-channel input image which should be warped with the final warpMatrix in
-order to provide an image similar to templateImage, same type as templateImage.
+@param inputImage single-channel input image to be warped to provide an image similar to
+ templateImage, same type as templateImage.
 @param inputMask An optional mask to indicate valid values of inputImage.
 
 @sa
--- a/modules/video/include/opencv2/video/tracking.hpp
+++ b/modules/video/include/opencv2/video/tracking.hpp
@@ -276,7 +276,7 @@ order to provide an image similar to templateImage, same type as templateImage.
 findTransformECC
  */
 
-CV_EXPORTS_W double computeECC(const Mat& templateImage, const Mat& inputImage, const Mat& inputMask);
+CV_EXPORTS_W double computeECC(InputArray templateImage, InputArray inputImage, InputArray inputMask = noArray());
 
 /** @example samples/cpp/image_alignment.cpp
 An example using the image alignment ECC algorithm
@@ -334,9 +334,16 @@ an exception if algorithm does not converges.
 computeECC, estimateAffine2D, estimateAffinePartial2D, findHomography
  */
 CV_EXPORTS_W double findTransformECC( InputArray templateImage, InputArray inputImage,
-                                      InputOutputArray warpMatrix, int motionType = MOTION_AFFINE,
-                                      TermCriteria criteria = TermCriteria(TermCriteria::COUNT+TermCriteria::EPS, 50, 0.001),
-                                      InputArray inputMask = noArray(), int gaussFiltSize = 5);
+                                      InputOutputArray warpMatrix, int motionType,
+                                      TermCriteria criteria,
+                                      InputArray inputMask, int gaussFiltSize);
+
+/** @overload */
+CV_EXPORTS
+double findTransformECC(InputArray templateImage, InputArray inputImage,
+    InputOutputArray warpMatrix, int motionType = MOTION_AFFINE,
+    TermCriteria criteria = TermCriteria(TermCriteria::COUNT+TermCriteria::EPS, 50, 0.001),
+    InputArray inputMask = noArray());
 
 /** @example samples/cpp/kalman.cpp
 An example using the standard Kalman filter
--- a/modules/video/include/opencv2/video/tracking.hpp
+++ b/modules/video/include/opencv2/video/tracking.hpp
@@ -265,6 +265,19 @@ enum
     MOTION_HOMOGRAPHY  = 3
 };
 
+/** @brief Computes the Enhanced Correlation Coefficient value between two images @cite EP08 .
+
+@param templateImage single-channel template image; CV_8U or CV_32F array.
+@param inputImage single-channel input image which should be warped with the final warpMatrix in
+order to provide an image similar to templateImage, same type as templateImage.
+@param inputMask An optional mask to indicate valid values of inputImage.
+
+@sa
+findTransformECC
+ */
+
+CV_EXPORTS_W double computeECC(const Mat& templateImage, const Mat& inputImage, const Mat& inputMask);
+
 /** @example samples/cpp/image_alignment.cpp
 An example using the image alignment ECC algorithm
 */
@@ -273,7 +286,7 @@ An example using the image alignment ECC algorithm
 
 @param templateImage single-channel template image; CV_8U or CV_32F array.
 @param inputImage single-channel input image which should be warped with the final warpMatrix in
-order to provide an image similar to templateImage, same type as temlateImage.
+order to provide an image similar to templateImage, same type as templateImage.
 @param warpMatrix floating-point \f$2\times 3\f$ or \f$3\times 3\f$ mapping matrix (warp).
 @param motionType parameter, specifying the type of motion:
  -   **MOTION_TRANSLATION** sets a translational motion model; warpMatrix is \f$2\times 3\f$ with
@@ -317,12 +330,12 @@ sample image_alignment.cpp that demonstrates the use of the function. Note that 
 an exception if algorithm does not converges.
 
 @sa
-estimateAffine2D, estimateAffinePartial2D, findHomography
+computeECC, estimateAffine2D, estimateAffinePartial2D, findHomography
  */
 CV_EXPORTS_W double findTransformECC( InputArray templateImage, InputArray inputImage,
                                       InputOutputArray warpMatrix, int motionType = MOTION_AFFINE,
                                       TermCriteria criteria = TermCriteria(TermCriteria::COUNT+TermCriteria::EPS, 50, 0.001),
-                                      InputArray inputMask = noArray());
+                                      InputArray inputMask = noArray(), int gaussFiltSize = 5);
 
 /** @example samples/cpp/kalman.cpp
 An example using the standard Kalman filter
--- a/modules/video/include/opencv2/video/tracking.hpp
+++ b/modules/video/include/opencv2/video/tracking.hpp
@@ -303,6 +303,7 @@ criteria.epsilon defines the threshold of the increment in the correlation coeff
 iterations (a negative criteria.epsilon makes criteria.maxcount the only termination criterion).
 Default values are shown in the declaration above.
 @param inputMask An optional mask to indicate valid values of inputImage.
+@param gaussFiltSize An optional value indicating size of gaussian blur filter; (DEFAULT: 5)
 
 The function estimates the optimum transformation (warpMatrix) with respect to ECC criterion
 (@cite EP08), that is
--- a/modules/video/src/ecc.cpp
+++ b/modules/video/src/ecc.cpp
@@ -314,32 +314,27 @@ static void update_warping_matrix_ECC (Mat& map_matrix, const Mat& update, const
 */
 double cv::computeECC(InputArray templateImage, InputArray inputImage, InputArray inputMask)
 {
-    Mat src = templateImage.getMat();
-    Mat dst = inputImage.getMat();
-    Mat inputMaskMat = inputMask.getMat();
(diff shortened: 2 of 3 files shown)


amithk-uc:
@alalek : Apologies if you're not the person to get in touch about this, but the error here: https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/20152/steps/Compare%20ABI%20dumps/logs/report-html seems to be that the abi has changed. This is intentional, as I added a new optional argument to the function. I didn't find anything specific about handling such changes in the contribute page: could you please help and let me know if there's any other change I need to make to make this test pass? Local build passes (Ubuntu 18.04.2 LTS) and the tests pass too (build/opencv_test_video). Thanks!

modules/video/include/opencv2/video/tracking.hpp
```diff
@@ -317,12 +331,12 @@ sample image_alignment.cpp that demonstrates the use of the function. Note that
...
 
 @sa
-estimateAffine2D, estimateAffinePartial2D, findHomography
+computeECC, estimateAffine2D, estimateAffinePartial2D, findHomography
  */
 CV_EXPORTS_W double findTransformECC( InputArray templateImage, InputArray inputImage,
                                       InputOutputArray warpMatrix, int motionType = MOTION_AFFINE,
                                       TermCriteria criteria = TermCriteria(TermCriteria::COUNT+TermCriteria::EPS, 50, 0.001),
-                                      InputArray inputMask = noArray());
+                                      InputArray inputMask = noArray(), int gaussFiltSize = 5);
```
alalek: Need to keep original overload too.
Try this:

```
CV_EXPORTS_W
double findTransformECC(InputArray templateImage, InputArray inputImage,
    InputOutputArray warpMatrix, int motionType,
    TermCriteria criteria,
    InputArray inputMask, int gaussFiltSize);

/** @overload */
CV_EXPORTS
double findTransformECC(InputArray templateImage, InputArray inputImage,
    InputOutputArray warpMatrix, int motionType = MOTION_AFFINE,
    TermCriteria criteria = TermCriteria(TermCriteria::COUNT+TermCriteria::EPS, 50, 0.001),
    InputArray inputMask = noArray());

... .cpp ...

double findTransformECC(InputArray templateImage, InputArray inputImage,
    InputOutputArray warpMatrix, int motionType,
    TermCriteria criteria,
    InputArray inputMask)
{
    return findTransformECC(..., inputMask, 5)
}
```

modules/video/include/opencv2/video/tracking.hpp
```diff
@@ -265,6 +265,19 @@ enum
...
+@param templateImage single-channel template image; CV_8U or CV_32F array.
+@param inputImage single-channel input image which should be warped with the final warpMatrix in
+order to provide an image similar to templateImage, same type as templateImage.
+@param inputMask An optional mask to indicate valid values of inputImage.
+
+@sa
+findTransformECC
+ */
+
+CV_EXPORTS_W double computeECC(const Mat& templateImage, const Mat& inputImage, const Mat& inputMask);
```
alalek: 1. Use `InputArray` in public API instead of direct Mat.

2.
    > inputMask
    > An optional mask

    inputMask = noArray()

3.
    > which should be warped with the final warpMatrix in
order to provide

    What?
    Re-check parameters documentation and their limitations (CV_8U/32F is not correct).

modules/video/src/ecc.cpp
```diff
@@ -309,16 +309,40 @@ static void update_warping_matrix_ECC (Mat& map_matrix, const Mat& update, const
...
 
 
+/** Function that computes enhanced corelation coefficient from Georgios et.al. 2008 [5]
+*   See https://github.com/opencv/opencv/issues/12432
+*/
+double cv::computeECC(const Mat& templateImage, const Mat& inputImage, const Mat& inputMask)
+{
+        Scalar meanTemplate, sdTemplate;
+
+        meanStdDev(templateImage, meanTemplate, sdTemplate, inputMask);
```
alalek: Indentation is 4 spaces.

modules/video/test/test_ecc.cpp
```diff
@@ -96,6 +96,44 @@ double CV_ECC_BaseTest::computeRMS(const Mat& mat1, const Mat& mat2){
...
+protected:
+    void run(int);
+    bool testCompute(int);
+};
+
+CV_ECC_Test_Compute::CV_ECC_Test_Compute(){
+}
+
+bool CV_ECC_Test_Compute::testCompute(int)
+{
```
alalek: Lets avoid using of legacy testing framework.

Just use Google Test:
```
TEST(Video_ECC_Test_Compute, accuracy)
{
    Mat testImg = (Mat_<float>(3, 3) << 1, 0, 0, 1, 0, 0, 1, 0, 0);
    Mat warpedImage = (Mat_<float>(3, 3) << 0, 1, 0, 0, 1, 0, 0, 1, 0);
    Mat_<unsigned char> mask = Mat_<unsigned char>::ones(testImg.rows, testImg.cols);
    double ecc = computeECC(warpedImage, testImg, mask);

    EXPECT_NEAR(ecc, -0.5f, 1e-5f);
}
```

modules/video/src/ecc.cpp
amithk-uc: Done.

modules/video/test/test_ecc.cpp
amithk-uc: Done: Thanks for writing out the Google Test version.

modules/video/include/opencv2/video/tracking.hpp
amithk-uc: Fixed #1 and #2. Not sure I understand #3. I used the same description for the parameters as the existing findTransformECC function. As long as both images are of the same type, and are single channel, we should be good, right?

modules/video/include/opencv2/video/tracking.hpp
amithk-uc: Thanks for all the help! This resolves the previous build issue: https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/20174/steps/Compare%20ABI%20dumps/logs/report-html

modules/video/include/opencv2/video/tracking.hpp
alalek: Parameter description uses undefined "warpMatrix" (not used in this function at all).

amithk-uc: Ah: thank you again. Here's the latest version:

@param inputImage single-channel input image to be warped to provide an image similar to
 templateImage, same type as templateImage.

modules/video/src/ecc.cpp
```diff
@@ -309,16 +309,53 @@ static void update_warping_matrix_ECC (Mat& map_matrix, const Mat& update, const
 }
 
 
+/** Function that computes enhanced corelation coefficient from Georgios et.al. 2008
+*   See https://github.com/opencv/opencv/issues/12432
+*/
+double cv::computeECC(InputArray templateImage, InputArray inputImage, InputArray inputMask)
+{
+    Mat src = templateImage.getMat();
+    Mat dst = inputImage.getMat();
```
alalek: > dst = inputImage

This doesn't look consistent.

It is better to declare parameters with underscore `_` suffix (in .cpp file only, keep .hpp without this hack). For "Mat" use original names.

```
double cv::computeECC(InputArray templateImage_, InputArray inputImage_, InputArray inputMask)
{
    Mat templateImage = templateImage_.getMat();
    Mat inputImage = inputImage_.getMat();
```

Anyway, you don't need to extract `Mat` from `InputArray`. Just use "InputArray" directly:
- `.cols` -> `.cols()`
- `.rows` -> `.rows()`
- or `Mat::ones(src.rows, src.cols, ...)` -> `Mat::ones(templateImage.size(), ...)`

modules/video/src/ecc.cpp
```diff
@@ -309,16 +309,53 @@ static void update_warping_matrix_ECC (Mat& map_matrix, const Mat& update, const
...
+    Mat inputMaskMat = inputMask.getMat();
+
+    CV_Assert(!src.empty());
+    CV_Assert(!dst.empty());
+
+    if( ! (src.type()==dst.type()))
+        CV_Error( Error::StsUnmatchedFormats, "Both input images must have the same data type" );
+
+    if(inputMask.empty())
+        inputMaskMat = Mat::ones(src.rows, src.cols, CV_8U);
```
alalek: You don't need this.
Keep `inputMask = noArray()` - it is more faster.

Also cache `countNonZero(inputMask)` value (with override for the case of "no mask"):
`int active_pixels = inputMask.empty() ? templateImage.size().area() : countNonZero(inputMask);`

modules/video/src/ecc.cpp
amithjkamath: Thanks for your patient guidance. Done.

modules/video/src/ecc.cpp
amithjkamath: That makes sense to me. Thank you once again.

alalek:
Thank you for update!

It is better to put this patch into [3.4 branch](https://github.com/opencv/opencv/wiki/Branches) (there are no strong dependencies on "master" features, also it makes support easier).

Please squash and rebase your commits:

If you have any issues, let us know.

amithjkamath:
@alalek Don't know if I did it right, but I followed your instructions and I suppose the branch is set to 3.4 and the checks have passed. Please do let me know if there's anything else I can do. Thank you very much for your patience and reviews!

modules/video/src/ecc.cpp
mixmixmix: Is it possible that python wrapper requires inputMask to be explicitly passed?
```
    (cc, warp_matrix) = cv2.findTransformECC(im1_gray,im2_gray,warp_matrix, warp_mode, criteria)
TypeError: Required argument 'inputMask' (pos 6) not found
```

modules/video/include/opencv2/video/tracking.hpp
```diff
@@ -317,12 +331,19 @@ sample image_alignment.cpp that demonstrates the use of the function. Note that
...
 CV_EXPORTS_W double findTransformECC( InputArray templateImage, InputArray inputImage,
-                                      InputOutputArray warpMatrix, int motionType = MOTION_AFFINE,
-                                      TermCriteria criteria = TermCriteria(TermCriteria::COUNT+TermCriteria::EPS, 50, 0.001),
-                                      InputArray inputMask = noArray());
+                                      InputOutputArray warpMatrix, int motionType,
+                                      TermCriteria criteria,
+                                      InputArray inputMask, int gaussFiltSize);
+
+/** @overload */
+CV_EXPORTS
```
alalek: Perhaps we need to use `CV_EXPORTS_W` here.

/cc @mixmixmix

PR #13840, opened 2019-02-15
Making JavaCamera2View be consistent with JavaCameraView

jaewoosong:
Bug fix of `Interrupted while setCameraPreviewSize` and `width and height must be > 0` which is related to `JavaCamera2View`

### This pullrequest changes

`org.opencv.android.JavaCamera2View` crashes on certain phones such as Samsung Galaxy S7 and Alcatel A5A. (On contrary, `JavaCameraView` works fine.) It is because `CameraBridgeViewBase.CAMERA_ID_BACK` and `CameraBridgeViewBase.CAMERA_ID_FRONT` do not match correct values. This patch introduces two new `final int` variables to correctly detect cameras.

### Bug which happens with old code
```
java.lang.RuntimeException: Interrupted while setCameraPreviewSize.
        at org.opencv.android.JavaCamera2View.connectCamera(JavaCamera2View.java:333)
        at org.opencv.android.CameraBridgeViewBase.onEnterStartedState(CameraBridgeViewBase.java:360)
        at org.opencv.android.CameraBridgeViewBase.processEnterState(CameraBridgeViewBase.java:321)
        at org.opencv.android.CameraBridgeViewBase.checkCurrentState(CameraBridgeViewBase.java:313)
        at org.opencv.android.CameraBridgeViewBase.surfaceChanged(CameraBridgeViewBase.java:198)
        at android.view.SurfaceView.updateWindow(SurfaceView.java:725)
        at android.view.SurfaceView$3.onPreDraw(SurfaceView.java:180)
        at android.view.ViewTreeObserver.dispatchOnPreDraw(ViewTreeObserver.java:944)
        at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:2553)
        at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1462)
        at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:6965)
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:907)
        at android.view.Choreographer.doCallbacks(Choreographer.java:709)
        at android.view.Choreographer.doFrame(Choreographer.java:644)
        at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:893)
        at android.os.Handler.handleCallback(Handler.java:836)
        at android.os.Handler.dispatchMessage(Handler.java:103)
        at android.os.Looper.loop(Looper.java:203)
        at android.app.ActivityThread.main(ActivityThread.java:6247)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1063)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:924)
     Caused by: java.lang.IllegalArgumentException: width and height must be > 0
        at android.graphics.Bitmap.createBitmap(Bitmap.java:877)
        at android.graphics.Bitmap.createBitmap(Bitmap.java:856)
        at android.graphics.Bitmap.createBitmap(Bitmap.java:823)
        at org.opencv.android.CameraBridgeViewBase.AllocateCache(CameraBridgeViewBase.java:457)
        at org.opencv.android.JavaCamera2View.connectCamera(JavaCamera2View.java:322)
        at org.opencv.android.CameraBridgeViewBase.onEnterStartedState(CameraBridgeViewBase.java:360) 
        at org.opencv.android.CameraBridgeViewBase.processEnterState(CameraBridgeViewBase.java:321) 
        at org.opencv.android.CameraBridgeViewBase.checkCurrentState(CameraBridgeViewBase.java:313) 
        at org.opencv.android.CameraBridgeViewBase.surfaceChanged(CameraBridgeViewBase.java:198) 
        at android.view.SurfaceView.updateWindow(SurfaceView.java:725) 
        at android.view.SurfaceView$3.onPreDraw(SurfaceView.java:180) 
        at android.view.ViewTreeObserver.dispatchOnPreDraw(ViewTreeObserver.java:944) 
        at android.view.ViewRootImpl.performTraversals(ViewRootImpl.java:2553) 
        at android.view.ViewRootImpl.doTraversal(ViewRootImpl.java:1462) 
        at android.view.ViewRootImpl$TraversalRunnable.run(ViewRootImpl.java:6965) 
        at android.view.Choreographer$CallbackRecord.run(Choreographer.java:907) 
        at android.view.Choreographer.doCallbacks(Choreographer.java:709) 
        at android.view.Choreographer.doFrame(Choreographer.java:644) 
        at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:893) 
        at android.os.Handler.handleCallback(Handler.java:836) 
        at android.os.Handler.dispatchMessage(Handler.java:103) 
        at android.os.Looper.loop(Looper.java:203) 
        at android.app.ActivityThread.main(ActivityThread.java:6247) 
        at java.lang.reflect.Method.invoke(Native Method) 
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1063) 
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:924) 

```

### DIFF
--- a/modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
+++ b/modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
@@ -110,6 +110,15 @@ public class JavaCamera2View extends CameraBridgeViewBase {
             if (mCameraID != null) {
                 Log.i(LOGTAG, "Opening camera: " + mCameraID);
                 manager.openCamera(mCameraID, mStateCallback, mBackgroundHandler);
+            } else { // make JavaCamera2View behaves in the same way as JavaCameraView
+                Log.i(LOGTAG, "Trying to open camera with the value (" + mCameraIndex + ")");
+                if (mCameraIndex < camList.length) {
+                    mCameraID = camList[mCameraIndex];
+                    manager.openCamera(mCameraID, mStateCallback, mBackgroundHandler);
+                } else {
+                    // CAMERA_DISCONNECTED is used when the camera id is no longer valid
+                    throw new CameraAccessException(CameraAccessException.CAMERA_DISCONNECTED);
+                }
             }
             return true;
         } catch (CameraAccessException e) {


alalek:
> `CameraBridgeViewBase.CAMERA_ID_BACK` and `CameraBridgeViewBase.CAMERA_ID_FRONT` do not match correct values.

There values are OpenCV-specific (not mapped into any Android values).
How they do not match?

---

How many cameras are available? (check this via debugger breakpoint)
Are camera rights properly requested?

jaewoosong:
Hi! There were two cameras on testing phones. One on the front and the other on the back. I also confirmed that the two cameras are recognized correctly by using the debugger. The permission was also correctly granted. If I change JavaCamera2View to JavaCameraView, other things kept exactly same, then everything works correctly. But if I change to JavaCamera2View then the error occurs.

alalek:
What value of "mCameraIndex" is showed by debugger?

jaewoosong:
`mCameraIndex` is `0` in debugger.

jaewoosong:
I think I found the cause of the problem.

`android.hardware.Camera` takes integer inputs between `0` and `getNumberOfCameras()-1` ([API link](https://developer.android.com/reference/android/hardware/Camera.html#open(int))). Let's consider the following example code.

```java
private CameraBridgeViewBase mOpenCvCameraView;
mOpenCvCameraView.setCameraIndex(0);
```

If I use `JavaCameraView`, I will eventually reach `initializeCamera()` and arrive at [line 128](https://github.com/opencv/opencv/blob/b39cd06249213220e802bb64260727711d9fc98c/modules/java/generator/android/java/org/opencv/android/JavaCameraView.java#L128): `mCamera = Camera.open(localCameraIndex);` Then `android.hardware.camera` will successfully open the camera.

On the other hand, if I use `JavaCamera2View` the code will arrive at `initializeCamera()` at [line 86](https://github.com/opencv/opencv/blob/f1d9666a20263c0e8e2471ab8160fb97ac03c433/modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java#L86) but then will fail to find a camera.

So it seems the difference between `Camera` and `Camera2` APIs causes the problem. With my pull request, the users will be able to make a smooth transition between `JavaCameraView` and `JavaCamera2View` easily.

alalek:
> mOpenCvCameraView.setCameraIndex(0);

There are special OpenCV-specific values for this parameter:
https://github.com/opencv/opencv/blob/4.0.1/modules/java/generator/android/java/org/opencv/android/CameraBridgeViewBase.java#L53-L55 

Looks like, current implementation of JavaCamera2View supports these values only, without supporting of zero-based device-specific indexes.

jaewoosong:
> Looks like, current implementation of JavaCamera2View supports these values only, without supporting of zero-based device-specific indexes.

Yes. `JavaCameraView` and `android.hardware.camera` both support zero-based device-specific indexes but `JavaCamera2View` does not. So `JavaCameraView` and `JavaCamera2View` behave differently. I thought it would be good if they behave in same way.

jaewoosong:
I made a new commit. Compared to the previous commit, this new commit makes logic of `JavaCameraView` and `JavaCamera2View` consistent. (on contrast, previous commit was just a bypass)

There are three important final variables:
* `CAMERA_ID_ANY = -1;`
* `CAMERA_ID_BACK  = 99;`
* `CAMERA_ID_FRONT = 98;`

The logic of `JavaCameraView`'s `initializeCamera()` ([link](https://github.com/opencv/opencv/blob/b39cd06249213220e802bb64260727711d9fc98c/modules/java/generator/android/java/org/opencv/android/JavaCameraView.java#L69)) is four steps:
1. Condition block of `if (mCameraIndex == CAMERA_ID_ANY)`
2. Condition block of `if (mCameraIndex == CAMERA_ID_BACK)`
3. Condition block of `if (mCameraIndex == CAMERA_ID_FRONT)`
4. If all the above does not work, then try `Camera.open(localCameraIndex)` with `localCameraIndex` given by the user.

On contrary, the logic of `JavaCamera2View`'s `initializeCamera()` ([link](https://github.com/opencv/opencv/blob/f1d9666a20263c0e8e2471ab8160fb97ac03c433/modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java#L86)) is **missing the last step**:
1. Condition block of `if (mCameraIndex == CameraBridgeViewBase.CAMERA_ID_ANY)`
2. Condition block of `if ((mCameraIndex == CameraBridgeViewBase.CAMERA_ID_BACK &&
                            characteristics.get(CameraCharacteristics.LENS_FACING) == CameraCharacteristics.LENS_FACING_BACK) ||
                        (mCameraIndex == CameraBridgeViewBase.CAMERA_ID_FRONT &&
                            characteristics.get(CameraCharacteristics.LENS_FACING) == CameraCharacteristics.LENS_FACING_FRONT)
                    )`

So in this new commit I added a new step to `JavaCamera2View`:
```java
else { // this corresponds to the last step of JavaCameraView's logic
    Log.i(LOGTAG, "Trying to open camera with new open(" + mCameraIndex + ")");
    mCameraID = Integer.toString(mCameraIndex);
    manager.openCamera(mCameraID, mStateCallback, mBackgroundHandler);
}
```

This new commit makes the logic of `JavaCameraView` and `JavaCamera2View` consistent. It supports more diverse inputs while doing no harm to the existing code.

modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
```diff
@@ -110,6 +110,10 @@ protected boolean initializeCamera() {
             if (mCameraID != null) {
                 Log.i(LOGTAG, "Opening camera: " + mCameraID);
                 manager.openCamera(mCameraID, mStateCallback, mBackgroundHandler);
+            } else { // make JavaCamera2View's logic consistent with JavaCameraView's.
+                Log.i(LOGTAG, "Trying to open camera with new open(" + mCameraIndex + ")");
+                mCameraID = Integer.toString(mCameraIndex);
```
alalek: I believe it should be:
```
... check range (mCameraIndex) or lets exception be triggered ...
mCameraID = camList[mCameraIndex];
```

jaewoosong: I agree that there should be error-checking. To be consistent with `JavaCameraView`, I think we should check `if (Arrays.asList(camList).contains(Integer.toString(mCameraIndex)))`. How do you think about the following code?

```suggestion
Log.i(LOGTAG, "Trying to open camera with new open(" + mCameraIndex + ")");
mCameraID = Integer.toString(mCameraIndex);
if (Arrays.asList(camList).contains(mCameraID)) {
    manager.openCamera(mCameraID, mStateCallback, mBackgroundHandler);
} else {
    // CAMERA_DISCONNECTED is used when the camera id is no longer valid
    throw new CameraAccessException(CameraAccessException.CAMERA_DISCONNECTED);
}
```

jaewoosong:
Added an error-checking block and squashed the commits into one!

modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
alalek: > mCameraIndex

is "int", but

> mCameraID

is string. "Integers" string are not expected in mCameraID.

This is wrong in all meanings:

> Integer.toString(mCameraIndex)


Take a look on line 96.

jaewoosong: As you said, `mCameraIndex` is `int` and `mCameraID` is `String`. For example, `mCameraIndex` can be `0` while `mCameraID` can be `"0"`. `mCameraID` is a stringyfied version of camera index starting from 0. That's why I used `Integer.toString()` to convert `int` to `String`, for example `0` to `"0"`. Although it is used less often, `Integer.toString()` is an equivalent of `String.valueOf()` so that it can also be used for this purpose.

alalek: Again, it should be:
```
mCameraID = camList[mCameraIndex];
```

jaewoosong: I think this code is what you want:
```java
Log.i(LOGTAG, "Trying to open camera with the value (" + mCameraIndex + ")");
if (mCameraIndex < camList.length) {
    mCameraID = camList[mCameraIndex];
    manager.openCamera(mCameraID, mStateCallback, mBackgroundHandler);
} else {
    // CAMERA_DISCONNECTED is used when the camera id is no longer valid
    throw new CameraAccessException(CameraAccessException.CAMERA_DISCONNECTED);
}
```

Consider the following situation:
* There are 3 cameras: `camera-0`, `camera-1` and `camera-2`
* A problem occurred on`camera-0`, so that `camList == {"1", "2"}`
* A user manually set `setCameraIndex(1)`
* Then `JavaCameraView` will open `camera-1`; My previous commit will also open `camera-1`; but this new code will open `camList[1]`, which is `camera-2`

I hope this explanation to be clear. If you still want this new code, please tell me. I will commit the new code and squash the previous commits.

alalek: Right!

Java and C++ uses zero-based indexes. So,
- `0` opens the first available camera (`camList[0]`)
- `1` opens the second camera (`camList[1]`)
- etc

We should not assume that `camList == {"1", "2"}` or contains direct integers.

> setCameraIndex(1)
> will open `camera-1`

No, it should open the second available camera (second from **the list of available cameras**).

jaewoosong: I committed the new code and squashed all previous commits. After the automated build and checks it will become available. Thanks for your hard work on OpenCV! Many people around the world, including me, are benefiting from you.

jaewoosong:
The new code (which corresponds to the resolved conversation) is committed and all previous commits are squashed into one. Thanks for the hard work of OpenCV maintainers!

alalek:
Thank you!

So, please:

jaewoosong:
Hi! I rebased the commit according to your explanation.

PR #13841, opened 2019-02-15
Fix Inference Engine backend

dkurt:
```
force_builders=Custom
test_opencl:Custom=ON
buildworker:Custom=linux-2
test_modules=dnn
```

### DIFF
--- a/modules/dnn/src/op_inf_engine.cpp
+++ b/modules/dnn/src/op_inf_engine.cpp
@@ -18,6 +18,11 @@ namespace cv { namespace dnn {
 
 #ifdef HAVE_INF_ENGINE
 
+// For networks with input layer which has an empty name, IE generates a name id[some_number].
+// OpenCV lets users use an empty input name and to prevent unexpected naming,
+// we can use some predefined name.
+static std::string kDefaultInpLayerName = "empty_inp_layer_name";
+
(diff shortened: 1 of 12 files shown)


modules/dnn/src/op_inf_engine.cpp
```diff
@@ -167,12 +180,56 @@ void InfEngineBackendNet::init(int targetId)
...
     int id = netBuilder.addLayer(layer);
     const std::string& layerName = layer.getName();
     CV_Assert(layers.insert({layerName, id}).second);
     unconnectedLayersIds.insert(id);
+
+#if INF_ENGINE_VER_MAJOR_GT(INF_ENGINE_RELEASE_2018R5)
+    // By default, all the weights are connected to last ports ids.
+    for (int i = 0; i < blobsIds.size(); ++i)
+    {
+        netBuilder.connect((size_t)blobsIds[i], {(size_t)id, portIds[i]});
```
mshabunin: Here `portIds[i]` is an _int_, it should be converted to _size_t_ explicitly too.

dkurt: @mshabunin, Thanks! Fixed in https://github.com/opencv/opencv/pull/13887.

PR #13843, opened 2019-02-15
LineIterator witout a Mat

chacha21:
cv::LineIterator can be used without being attached to any cv::Mat, it only needs the size and type of data. An alternative constructor has been defined for that.
In that case, a LineIterator can no more be dereferenced with the * operator, but pos() still returns valid pixel positions.
It can be useful when LineIterator is just used to compute positions of pixels on a line, without requiring to build a Mat just for that.
Use case : with a dataset that would represent a huge image, pixel positions can be pre-computed before querying the dataset API.

### DIFF
--- a/modules/imgproc/include/opencv2/imgproc.hpp
+++ b/modules/imgproc/include/opencv2/imgproc.hpp
@@ -4697,7 +4697,7 @@ public:
       init(size, type, 0, CV_ELEM_SIZE(type)*size.width, pt1, pt2, connectivity, leftToRight);
     }
     void init(const Size& size, int type, uchar* data, size_t dataStep, Point pt1, Point pt2, int connectivity = 8, bool leftToRight = false);
-  
+
     /** @brief returns pointer to the current pixel
     */
     uchar* operator *();
--- a/modules/imgproc/include/opencv2/imgproc.hpp
+++ b/modules/imgproc/include/opencv2/imgproc.hpp
@@ -4689,7 +4689,15 @@ public:
     not to depend on the ordering of pt1 and pt2 parameters
     */
     LineIterator( const Mat& img, Point pt1, Point pt2,
-                  int connectivity = 8, bool leftToRight = false );
+                  int connectivity = 8, bool leftToRight = false ) {
+      init(img.size(), img.type(), (uchar*)img.ptr(), img.step1()*img.elemSize1(), pt1, pt2, connectivity, leftToRight);
+    }
+    LineIterator( const Size& size, int type, Point pt1, Point pt2,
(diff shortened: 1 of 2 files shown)


mshabunin:
I think this class should be completely redesigned, it should not "know" about `Mat` and data. I suggest implementing new safer class and deprecating the existing one.

vpisarev:
safer implementation might be nice to have, but as the intermediate solution this patch looks good enough to me :+1:

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4689,7 +4689,15 @@ class CV_EXPORTS LineIterator
     not to depend on the ordering of pt1 and pt2 parameters
     */
     LineIterator( const Mat& img, Point pt1, Point pt2,
-                  int connectivity = 8, bool leftToRight = false );
+                  int connectivity = 8, bool leftToRight = false ) {
+      init(img.size(), img.type(), (uchar*)img.ptr(), img.step1()*img.elemSize1(), pt1, pt2, connectivity, leftToRight);
+    }
+    LineIterator( const Size& size, int type, Point pt1, Point pt2,
+                  int connectivity = 8, bool leftToRight = false ) {
```
alalek: `type` is not needed for this interface at all.

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4718,7 +4726,7 @@ class CV_EXPORTS LineIterator
 inline
 uchar* LineIterator::operator *()
 {
-    return ptr;
+    return !ptr0 ? 0 : ptr;//when no Mat is attached, ptr is just a dummy address and should not be dereferenced
```
alalek: "Contact" of this call has been changed.
This patch allows "NULL" as a result. At least this would produces static analyzers issues which need to be resolved after this patch.

chacha21:
About the "type" not needed in the new interface :
This is just to ensure the best backward compatibility, and keep the old code that is based on address computations. One can mimic the behaviour of the LineIterator attached to a mat. 
Removing the "type" parameter would mean
-either change the implementation to not use addresses any more
-or default to CV_8UC1, but then again, why not mimic the old behaviour ?

About the operator *() returning 0
-I could return the address of a static variable large enough to hold a CV_64FC4
-I coudl return the raw "ptr" value, here again to mimic behaviour. Even a dummy address could be interesting as an offset information, but it seems dangerous.

alalek:
IMHO, separate specialized class would be better (without "operator *()" and without non needed pointers arithmetic).

PR #13847, opened 2019-02-16
Add JPEG2000 compression flag.

ghellwig:
I believe the title says it all.

Since this is my first PR to the OpenCV repository, I might have missed some required changes.
Therefore, your guidance is very appreciated!

Best,
Gregor

### DIFF
--- a/modules/imgcodecs/include/opencv2/imgcodecs.hpp
+++ b/modules/imgcodecs/include/opencv2/imgcodecs.hpp
@@ -94,7 +94,8 @@ enum ImwriteFlags {
        IMWRITE_PAM_TUPLETYPE       = 128,//!< For PAM, sets the TUPLETYPE field to the corresponding string value that is defined for the format
        IMWRITE_TIFF_RESUNIT = 256,//!< For TIFF, use to specify which DPI resolution unit to set; see libtiff documentation for valid values
        IMWRITE_TIFF_XDPI = 257,//!< For TIFF, use to specify the X direction DPI
-       IMWRITE_TIFF_YDPI = 258 //!< For TIFF, use to specify the Y direction DPI
+       IMWRITE_TIFF_YDPI = 258, //!< For TIFF, use to specify the Y direction DPI
+       IMWRITE_JPEG2000_COMPRESSION_X1000 = 272 //!< For JPEG2000, use to specify the target compression rate (multiplied by 1000). The value can be from 0 to 1000. Default is 1000.
      };
 
 enum ImwriteEXRTypeFlags {
--- a/modules/imgcodecs/src/grfmt_jpeg2000.cpp
+++ b/modules/imgcodecs/src/grfmt_jpeg2000.cpp
@@ -43,6 +43,7 @@
 #include "precomp.hpp"
 
 #ifdef HAVE_JASPER
+#include <sstream>
 
 #include "grfmt_jpeg2000.hpp"
 #include "opencv2/imgproc.hpp"
@@ -467,7 +468,7 @@ bool  Jpeg2KEncoder::isFormatSupported( int depth ) const
 }
 
 
-bool  Jpeg2KEncoder::write( const Mat& _img, const std::vector<int>& )
+bool  Jpeg2KEncoder::write( const Mat& _img, const std::vector<int>& params )
 {
     int width = _img.cols, height = _img.rows;
     int depth = _img.depth(), channels = _img.channels();
@@ -475,6 +476,18 @@ bool  Jpeg2KEncoder::write( const Mat& _img, const std::vector<int>& )
 
     if( channels > 3 || channels < 1 )
         return false;
+
+    CV_Assert(params.size() % 2 == 0);
+    double target_compression_rate = 1.0;
+    for( size_t i = 0; i < params.size(); i += 2 )
+    {
+        switch(params[i])
+        {
+        case cv::IMWRITE_JPEG2000_COMPRESSION_X1000:
+            target_compression_rate = std::min(std::max(params[i+1], 0), 1000) / 1000.0;
+            break;
+        }
+    }
 
     jas_image_cmptparm_t component_info[3];
     for( int i = 0; i < channels; i++ )
@@ -511,7 +524,10 @@ bool  Jpeg2KEncoder::write( const Mat& _img, const std::vector<int>& )
         jas_stream_t *stream = jas_stream_fopen( m_filename.c_str(), "wb" );
         if( stream )
         {
-            result = !jas_image_encode( img, stream, jas_image_strtofmt( (char*)"jp2" ), (char*)"" );
+            std::stringstream options;
+            options << "rate=" << target_compression_rate;
+
+            result = !jas_image_encode( img, stream, jas_image_strtofmt( (char*)"jp2" ), (char*)options.str().c_str() );
 
             jas_stream_close( stream );
         }


vpisarev:
thank you for the patch, see the comments below

modules/imgcodecs/src/grfmt_jpeg2000.cpp
```diff
@@ -476,6 +477,17 @@ bool  Jpeg2KEncoder::write( const Mat& _img, const std::vector<int>& )
     if( channels > 3 || channels < 1 )
         return false;
 
+    double target_compression_rate = 1.0;
+    for( size_t i = 0; i < params.size(); i += 2 )
```
vpisarev: there should be check that `params.size() % 2 == 0`

modules/imgcodecs/src/grfmt_jpeg2000.cpp
```diff
@@ -476,6 +477,17 @@ bool  Jpeg2KEncoder::write( const Mat& _img, const std::vector<int>& )
...
+    {
+        switch(params[i])
+        {
+        case cv::IMWRITE_JPEG2000_COMPRESSION:
+            target_compression_rate /= params[i+1];
```
vpisarev: handle the case when `params[i+1] <= 0`

modules/imgcodecs/include/opencv2/imgcodecs.hpp
```diff
@@ -94,7 +94,8 @@ enum ImwriteFlags {
        IMWRITE_PAM_TUPLETYPE       = 128,//!< For PAM, sets the TUPLETYPE field to the corresponding string value that is defined for the format
        IMWRITE_TIFF_RESUNIT = 256,//!< For TIFF, use to specify which DPI resolution unit to set; see libtiff documentation for valid values
        IMWRITE_TIFF_XDPI = 257,//!< For TIFF, use to specify the X direction DPI
-       IMWRITE_TIFF_YDPI = 258 //!< For TIFF, use to specify the Y direction DPI
+       IMWRITE_TIFF_YDPI = 258, //!< For TIFF, use to specify the Y direction DPI
+       IMWRITE_JPEG2000_COMPRESSION = 512 //!< For JPEG2000, use to specify the target compression rate. A higher value means a smaller size. Default is 1.
```
alalek: Initial increment was "16" between codecs: "(codec_index << 4) + prop_index"

Also we have plans to drop "Jasper" support completely (Jasper project is in abandoned state for several years) and replace in by OpenJPEG.
So, how is this property is mapped into OpenJPEG implementation?

modules/imgcodecs/src/grfmt_jpeg2000.cpp
ghellwig: Thank you for your comments!
Indeed a valid point!
I will add a check, if that's still required after resolving https://github.com/opencv/opencv/pull/13847#discussion_r258039971.

modules/imgcodecs/src/grfmt_jpeg2000.cpp
ghellwig: Very valid point!
Concerning the definition of "too big". I guess if 1/compression is numerically 0 is likely a good boundary, right?

modules/imgcodecs/include/opencv2/imgcodecs.hpp
ghellwig: Thank you for your feedback!
Concerning the increment: I was just guessing that the increment is a power of 2 because the previous values were 64,128, and 256.
I can adjust this to 272, if you prefer.

Concerning an OpenJPEG-based implementation: I have not used this library so far, i.e. I cannot directly answer this question.
If you like, I can do some research and propose a OpenJPEG-based implementation.
If you anyway want to drop JasPer, there is not really a point in merging this PR, right?

What do you think?

alalek: If it is Jasper specific behavior (without alternative in OpenJPEG) then it is better to avoid using such common name.

What is about this name?

`IMWRITE_JPEG2000_JASPER_COMPRESSION`

modules/imgcodecs/src/grfmt_jpeg2000.cpp
vpisarev: well, maybe not. Maybe 100 (or even 10) will already be such a big number that the produced image will have very low quality. For example, Photoshop uses the compression quality in from 0 to 12, some other editors use percentage (100% is the best quality, %50 is so-so, closer to 0% the image degrades very much). I think, it would be useful to have some guidance - how this quality parameter maps to the percentage or to the Photoshop scale.

ghellwig: Hm, a value of 100 would mean a target rate of 1%, i.e. depending on the image it might still be a possible value.
I would actually refrain from imposing an artificial upper limit there is no generally applicable limit for this parameter.
Maybe, one should point out in the documentation that it is a target, not a guaranteed value?

modules/imgcodecs/include/opencv2/imgcodecs.hpp
ghellwig: I think this name is fine.
However, I am wondering if it makes sense to introduce this enum, if it will be deprecated soon.
I will implement the changes later this week, if we agreed on the details.

modules/imgcodecs/src/grfmt_jpeg2000.cpp
```diff
@@ -476,6 +477,18 @@ bool  Jpeg2KEncoder::write( const Mat& _img, const std::vector<int>& )
...
         return false;
 
+    CV_Assert(params.size() % 2 == 0);
+    double target_compression_rate = 1.0;
+    for( size_t i = 0; i < params.size(); i += 2 )
+    {
+        switch(params[i])
+        {
+        case cv::IMWRITE_JPEG2000_JASPER_COMPRESSION:
+            target_compression_rate /= std::max(params[i+1], 1);
```
alalek: It seems that this haven't to behave well:
- OpenCV parameter value has "int" type
- valid parameter values are 1,2,3,4,...
- corresponding `target_compression_rate` are: 1.0, 0.5, 0.333, 0.25

As we can see target values set is very discreet.
Do we really need it in this way? (for example, we don't need 66%, 80% compression)

ghellwig: You are right that the values are discrete due to the `int` type.
If you prefer the possible values to be equidistant, I could change the implementation to something similar to the JPEG quality parameter (0-100), where one integer means 1% and 0 corresponds to the smallest possible compression rate.
In this case one could rename `COMPRESSION` to `QUALITY`.
What do you think?

alalek: Looks like JPEG2000 codecs may have both parameters "COMPRESSION" and "QUALITY":
- OpenJPEG compression tool: https://github.com/uclouvain/openjpeg/wiki/DocJ2KCodec (`-r` and `-q`)
- Jasper compression: https://www.ece.uvic.ca/~frodo/jasper/jasper.pdf (page 17 - example 3,4 for compression)
- Jasper quality - ?

and they are different (by behavior).

What is about these names:
- `IMWRITE_JPEG2000_QUALITY` - range 0-100 (if Jasper has this parameter)
- `IMWRITE_JPEG2000_COMPRESSION_X1000`, parameter value is multiplied by 1e-3

ghellwig: Thanks for your further investigation!
I scanned again the JasPer documentation and it seems that there's no quality parameter.
Therefore, my proposal would be the following:
 - introduce the parameter `IMWRITE_JPEG2000_COMPRESSION_X1000` proposed by you
 - do not introduce `IMWRITE_JPEG2000_QUALITY` now

Once JasPer is removed in favour of OpenJPEG, one can maintain `IMWRITE_JPEG2000_COMPRESSION_X1000` with the same meaning as with JasPer and add a new parameter `IMWRITE_JPEG2000_QUALITY` which would have a similar meaning as the JPEG quality parameter.
I believe this procedure would be the most future-proof version discussed so far.
What do you think?

alalek: Agreed!

ghellwig:
@alalek Looks like the build failure is unrelated to this PR, right?

PR #13849, opened 2019-02-18
Remove opencv_highgui as optional dependency of features2d module

tedsteiner:
The highgui module is added as an optional dependency of features2d.
However, this is only used in file blobdetector.cpp for a default-disabled debug option.
This makes it simple if someone someday wants to debug the blob detector.
However, this also means that any library downstream linking to the features2d module will also need to link to highgui and its dependencies.

### This pullrequest changes

[From my understanding] the current behavior is that if `BUILD_opencv_highgui` is enabled, then the `highgui` module is added as an optional dependency for the `features2d` module.
However, the only place `highgui` elements are used are in `blobdetector.cpp`, and only if `DEBUG_BLOB_DETECTOR` is manually defined.
And even if `DEBUG_BLOB_DETECTOR` is defined, all lines of code in the `DEBUG_BLOB_DETECTOR` code blocks would need to be uncommented.
These lines of debug code have been commented out in the repository for at least 8 years.

My guess is that these lines of debug code are very rarely, if ever, used, and only in special cases.
I would further guess that users who enable this debug tool would know how to enable highgui (or to read the fairly clear compile errors saying highgui not found). 

I think it makes sense to delete the debug code entirely, however, as this is my first OpenCV PR I don't want to overstep.

### Reasoning

A codebase that I manage benefits from the OpenCV features2d library.
A "core" portion of this codebase is used on embedded systems that do not support GUI tools.
A separate portion of this codebase allows GUI tools.
We recently ran into some issues where the features2d library caused strange build errors because it silently links against highgui due to CMake's modules system.
Once we tracked down the issue and understood it we were able to fix the issue for our systems.
However, it seems like linking features2d against highgui just because its' enabled doesn't serve any larger benefit except for in a very specific and rare use case, and sometimes can lead to system integration challenges.

### DIFF
--- a/modules/features2d/CMakeLists.txt
+++ b/modules/features2d/CMakeLists.txt
@@ -1,2 +1,2 @@
 set(the_description "2D Features Framework")
-ocv_define_module(features2d opencv_imgproc OPTIONAL opencv_flann opencv_highgui WRAP java python js)
+ocv_define_module(features2d opencv_imgproc OPTIONAL opencv_flann WRAP java python js)
--- a/modules/features2d/src/blobdetector.cpp
+++ b/modules/features2d/src/blobdetector.cpp
@@ -44,17 +44,6 @@
 #include <iterator>
 #include <limits>
 
-//#define DEBUG_BLOB_DETECTOR
-
-#ifdef DEBUG_BLOB_DETECTOR
-#  include "opencv2/opencv_modules.hpp"
-#  ifdef HAVE_OPENCV_HIGHGUI
-#    include "opencv2/highgui.hpp"
-#  else
-#    undef DEBUG_BLOB_DETECTOR
-#  endif
-#endif
-
 namespace cv
 {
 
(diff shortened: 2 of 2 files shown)


mshabunin:
I think it is better to remove debug code completely. 

Looks like _sfm_ module in _opencv_contrib_ uses _highgui_ as transitive dependency from _features2d_, so it should be explicitly listed in module dependencies here: https://github.com/opencv/opencv_contrib/blob/796853e057d08ae840e2d7d559775e8b701084a0/modules/sfm/CMakeLists.txt#L79-L85

Could you please create PR  to the _opencv_contrib_ repository using branch with same name (`features2d-without-highgui`), so that our CI could verify both changes together.

alalek:
It is not features2d-specific case only.
Similar debug code exists in other modules, like calib3d.

We need more robust/common solution for this.

tedsteiner:
I will update this soon with a matching branch in opencv_contrib and removing the debug code sections entirely.

When I dug in to where this is used in the `sfm` module, it looks to me like this is the only place: https://github.com/opencv/opencv_contrib/blob/796853e057d08ae840e2d7d559775e8b701084a0/modules/sfm/src/libmv_light/libmv/correspondence/nRobustViewMatching.cc#L21
It looks like this is included just for the `cv::imread` call: https://github.com/opencv/opencv_contrib/blob/796853e057d08ae840e2d7d559775e8b701084a0/modules/sfm/src/libmv_light/libmv/correspondence/nRobustViewMatching.cc#L55

I won't be able to test this until I get to another computer. However, I think the better fix is to remove the include line to highgui in the `sfm` module rather than to link `sfm` against `highgui` and link sfm against imgcodecs for the call to imread.

tedsteiner:
I removed the debug lines from `blobdetector.cpp` and made a partner pull request in opencv_contrib that should fix the build issues: https://github.com/opencv/opencv_contrib/pull/2007

tedsteiner:
Is there anything holding this back from being merged?

PR #13858, opened 2019-02-18
G-API Fluid nv12 support

rgarnov:
* Added NV12toRGB to FluidKernel::Kind enum to inform the engine of the special scheduling mechanic of the kernel (kernel consumes 2 lines from the first input, 1 line from the second, and produces 2 lines)
* Added appropriate NV12toRGBFluidAgent - the scheduler entity to support new kind of scheduling and line dependencies, added input port parameter to schedule related methods of FluidAgent (firstWindow, linesReadAndNextWindow)
* Extended FluidUnit structure to have a vector of line consumptions to support different line consumption requirements on different input ports
* Extended roi inference process to be aware of new kind of in/out roi dependency
* Added test kernel and tests

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/fluid/gfluidbuffer.hpp
+++ b/modules/gapi/include/opencv2/gapi/fluid/gfluidbuffer.hpp
@@ -53,6 +53,10 @@ public:
 
         inline const uint8_t* linePtr(int index) const
         {
+            // "out_of_window" check:
+            // user must not request the lines which are outside of specified kernel window
+            GAPI_DbgAssert(index >= -m_border_size
+                        && index <  -m_border_size + static_cast<int>(m_linePtrs.size()));
             return m_linePtrs[index + m_border_size];
         }
     };
--- a/modules/gapi/include/opencv2/gapi/fluid/gfluidkernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/fluid/gfluidkernel.hpp
@@ -49,7 +49,8 @@ public:
     enum class Kind
     {
         Filter,
-        Resize
+        Resize,
+        NV12toRGB
     };
 
     // This function is a generic "doWork" callback
--- a/modules/gapi/src/backends/fluid/gfluidbackend.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidbackend.cpp
@@ -67,7 +67,7 @@ namespace
         {
             GFluidModel fm(graph);
             auto fluid_impl = cv::util::any_cast<cv::GFluidKernel>(impl.opaque);
-            fm.metadata(op_node).set(cv::gimpl::FluidUnit{fluid_impl, {}, 0, 0, 0.0});
+            fm.metadata(op_node).set(cv::gimpl::FluidUnit{fluid_impl, {}, 0, {}, 0.0});
         }
 
         virtual EPtr compile(const ade::Graph &graph,
@@ -138,8 +138,8 @@ private:
 struct FluidFilterAgent : public FluidAgent
 {
 private:
-    virtual int firstWindow() const override;
-    virtual std::pair<int,int> linesReadAndnextWindow() const override;
+    virtual int firstWindow(std::size_t inPort) const override;
+    virtual std::pair<int,int> linesReadAndnextWindow(std::size_t inPort) const override;
     virtual void setRatio(double) override { /* nothing */ }
 public:
     using FluidAgent::FluidAgent;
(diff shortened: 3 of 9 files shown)


modules/gapi/include/opencv2/gapi/fluid/gfluidbuffer.hpp
```diff
@@ -53,6 +53,8 @@ class GAPI_EXPORTS View
 
         inline const uint8_t* linePtr(int index) const
         {
+            GAPI_DbgAssert(index >= -m_border_size
+                        && index <  -m_border_size + static_cast<int>(m_linePtrs.size()));
```
dmatveev: Please document this assert. Not clear to me how index and border size are related each other.

modules/gapi/include/opencv2/gapi/fluid/gfluidkernel.hpp
```diff
@@ -49,7 +49,8 @@ class GAPI_EXPORTS GFluidKernel
     enum class Kind
     {
         Filter,
-        Resize
+        Resize,
+        NV12toRGB
```
dmatveev: Note the enum is `kind`, not a particular conversion type.
This `Kind` tag affects the way we handle scheduling for kernel (input lines : output lines).

We may support YUV422 (interleaved) for output, and GRGB (Bayer) for input.

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -554,14 +583,15 @@ void cv::gimpl::GFluidExecutable::initBufferRois(std::vector<int>& readStarts,
...
-                const auto& in_data = m_gm.metadata(inNode).get<Data>();
+                const auto& in_node = in_edge->srcNode();
+                const auto& in_data = m_gm.metadata(in_node).get<Data>();
 
-                if (in_data.shape == GShape::GMAT && fg.metadata(inNode).contains<FluidData>())
+                if (in_data.shape == GShape::GMAT && fg.metadata(in_node).contains<FluidData>())
                 {
                     const auto& in_meta = util::get<GMatDesc>(in_data.meta);
-                    const auto& fd = fg.metadata(inNode).get<FluidData>();
+                    const auto& fd = fg.metadata(in_node).get<FluidData>();
```
dmatveev: So what was the reason to move from a loop over input nodes to a loop over input edges?

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -599,13 +629,29 @@ void cv::gimpl::GFluidExecutable::initBufferRois(std::vector<int>& readStarts,
...
                     cv::gapi::own::Rect produced = rois[m_id_map.at(data.rc)];
 
                     cv::gapi::own::Rect resized;
                     switch (fg.metadata(oh).get<FluidUnit>().k.m_kind)
                     {
-                    case GFluidKernel::Kind::Filter: resized = produced; break;
-                    case GFluidKernel::Kind::Resize: resized = adjResizeRoi(produced, in_meta.size, meta.size); break;
+                    case GFluidKernel::Kind::Filter:    resized = produced; break;
+                    case GFluidKernel::Kind::Resize:    resized = adjResizeRoi(produced, in_meta.size, meta.size); break;
+                    case GFluidKernel::Kind::NV12toRGB: resized = adjNV12Roi(produced, m_gm.metadata(in_edge).get<Input>().port); break;
```
dmatveev: Ok, I see now.

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -849,21 +896,35 @@ namespace
...
+                    const auto& d = g.metadata(in_edge->srcNode()).get<Data>();
+                    if (d.shape == cv::GShape::GMAT)
+                    {
+                        auto port = g.metadata(in_edge).get<Input>().port;
+                        fu.line_consumption[port] = maxLineConsumption(fu.k, in_h, out_h, fu.k.m_lpi, port);
 
-                fu.border_size = border_size;
-                fu.line_consumption = line_consumption;
+                        GModel::log(g, node, "Line consumption (port " + std::to_string(port) + "): "
+                                  + std::to_string(fu.line_consumption[port]));
```
dmatveev: Please fix the alignment here

modules/gapi/include/opencv2/gapi/fluid/gfluidbuffer.hpp
rgarnov: Done

modules/gapi/src/backends/fluid/gfluidbackend.cpp
rgarnov: Done

modules/gapi/include/opencv2/gapi/fluid/gfluidkernel.hpp
rgarnov: Interleaved case perfectly fits our recent engine as "line to line" kernel (user himself should interpret his pixel layout in his calculations). Bayer fits our engine too as "two lines to two lines" (2 lpi) kernel (Bayer mosaic with NxN pixels results in NxN rgb image, actually we get 3x number of pixels by interpolating the mosaic). So I propose to keep current name for nv12 kind of in/out lines scheduling as we do not have another precedents

PR #13860, opened 2019-02-19
G-API: Compile time checking kernels id

TolyaTalamanov:
### This pullrequest changes
Add static assert that all kernel id's are unique inside kernel package.
Kernels are compared by id


TolyaTalamanov:
@dmatveev @rgarnov

dmatveev:
Please start with a description what kind of check you actually introduce.

Does it check kernel _types_ which are implemented? Does it check their text _id_s to be unique?

It would be easier to review the code changes after that (knowing exactly what the goal is)

PR #13869, opened 2019-02-19
LineVirtualIterator

chacha21:
Proposal of LineVirtualIterator, an alternative to "LineIterator not attached to any mat".
This is basically the same implementation, replacing the address difference by a single "offset" variable. elemsize becomes irrelevant and considered to be 1. "step" is thus equal to size.width since no stride is expected.

related to #13843


modules/imgproc/src/drawing.cpp
```diff
@@ -248,6 +248,89 @@ void LineIterator::init(const Size& size, int type, uchar* data, size_t dataStep
...
+    if( (unsigned)pt1.x >= (unsigned)(size.width) ||
+        (unsigned)pt2.x >= (unsigned)(size.width) ||
+        (unsigned)pt1.y >= (unsigned)(size.height) ||
+        (unsigned)pt2.y >= (unsigned)(size.height) )
+    {
+        if( !clipLine( size, pt1, pt2 ) )
+        {
+            offset = 0;
+            err = plusDelta = minusDelta = plusStep = minusStep = count = 0;
+            step = 0;
```
alalek: This causes "div by zero" error in ".pos()": `(offset/step)`

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4757,6 +4757,73 @@ Point LineIterator::pos() const
...
+    LineVirtualIterator& operator ++();
+    /** @brief postfix increment operator (it++). shifts iterator to the next pixel
+    */
+    LineVirtualIterator operator ++(int);
+    /** @brief returns coordinates of the current pixel
+    */
+    Point pos() const;
+
+    int offset;
+    int step;
```
alalek: I believe "offset"/"step" is over complication here.
`Point` result is enough, especially in case of ROI (with step != width).

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4757,6 +4757,73 @@ Point LineIterator::pos() const
...
+
+    creates iterators for the line connecting pt1 and pt2
+    the line will be clipped on the image boundaries
+    the line is 8-connected or 4-connected
+    If leftToRight=true, then the iteration is always done
+    from the left-most point to the right most,
+    not to depend on the ordering of pt1 and pt2 parameters
+    */
+    LineVirtualIterator(const Size& size, Point pt1, Point pt2,
+                        int connectivity = 8, bool leftToRight = false );
```
alalek: > leftToRight

I believe direction should be specified by `pt1`/`pt2` instead of additional parameter.
Especially, in case `leftToRight = false` doesn't mean "right-to-left".

---

Upd: `forceLeftToRight` naming makes more sense.

modules/imgproc/src/drawing.cpp
```diff
@@ -248,6 +248,89 @@ void LineIterator::init(const Size& size, int type, uchar* data, size_t dataStep
...
+    dy ^= dx & s;
+    dx ^= dy & s;
+
+    bt_pix ^= istep & s;
+    istep ^= bt_pix & s;
+    bt_pix ^= istep & s;
+
+    if( connectivity == 8 )
+    {
+        assert( dx >= 0 && dy >= 0 );
```
alalek: "C" asserts must be avoided.

modules/imgproc/src/drawing.cpp
```diff
@@ -248,6 +248,89 @@ void LineIterator::init(const Size& size, int type, uchar* data, size_t dataStep
...
+    {
+        dx = (dx ^ s) - s;
+        bt_pix = (bt_pix ^ s) - s;
+    }
+
+    offset = static_cast<int>(pt1.y * istep + pt1.x);
+
+    s = dy < 0 ? -1 : 0;
+    dy = (dy ^ s) - s;
+    istep = (istep ^ s) - s;
```
alalek: > "undefined behavior"

Lets try to avoid that.
Compilers are smart enough (or at least this is not required in non "heavy" constructor calls).

chacha21:
some suggested improvements have been made in the new commit.
However, adding "accuracy tests" is a new step for me in my learning curve of OpenCV contributing, I still have to learn it.

modules/imgproc/src/drawing.cpp
chacha21: Not sure to understand this one. This is the LineIterator implementation; is there a problem here ?

chacha21: suggested improvements have been made in the new commit.
However, adding "accuracy tests" is a new step for me in my learning curve of OpenCV contributing, I still have to learn it.

chacha21:
I still did not updated the doc in the source code, I wait to know if the pull request is accepted

asmorkalov:
@chacha21 Do you have a chance to work on some tests for the new feature?

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4777,6 +4777,83 @@ Point LineIterator::pos() const
...
+      currentPosOffset = Point(aRect.x, aRect.y);
+      init(aRect.size(), pt1, pt2, connectivity, forceLeftToRight);
+    }
+    LineVirtualIterator(const Size& aSize, const Point& pt1, const Point& pt2,
+                        int connectivity = 8, bool forceLeftToRight = false ) {
+      init(aSize, pt1, pt2, connectivity, forceLeftToRight);
+    }
+    LineVirtualIterator(const Point& pt1, const Point& pt2,
+                        int connectivity = 8, bool forceLeftToRight = false );
+    void init(const Size& size, const Point& pt1, const Point& pt2,
```
VadimLevin: What is the purpose of public `init` method?

All initialization stuff can be done by constructors chaining.
Your main constructor is `LineVirtualIterator(const Size&, const Point&, const Point&, int, bool)`
```c++
struct A {
    A(int arg1, int arg2, int arg3) {
        std::cout << "Initialized with " << arg1 << ' ' << arg2 << ' ' << arg3 << '\n';
    }

    A(int arg1, int arg2) : A(arg1, arg2, 123) {}

    explicit A(int arg1) : A(arg1, 4) {}
};
```

Also note, that sometimes it is reasonable to use a ["named" constructor idiom](https://isocpp.org/wiki/faq/ctors#named-ctor-idiom) to make user understanding clear, what object is going to be created. (RVO eliminates copy and move constructor calls)

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4777,6 +4777,83 @@ Point LineIterator::pos() const
...
+//! @cond IGNORED
+
+// === LineVirtualIterator implementation ===
+
+inline
+LineVirtualIterator& LineVirtualIterator::operator ++()
+{
+    if (size.width)
+    {
+      int mask = err < 0 ? -1 : 0;
```
VadimLevin: `const`

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4777,6 +4777,83 @@ Point LineIterator::pos() const
...
+      err += minusDelta + (plusDelta & mask);
+      int offset = minusStep + (plusStep & mask);
```
VadimLevin: `const`

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4777,6 +4777,83 @@ Point LineIterator::pos() const
...
+      size_t flattenedCoord = (currentPos.y-currentPosOffset.y)*size.width+(currentPos.x-currentPosOffset.x)+offset;
```
VadimLevin: `const`

modules/imgproc/src/drawing.cpp
```diff
@@ -248,6 +248,97 @@ LineIterator::LineIterator(const Mat& img, Point pt1, Point pt2,
     this->elemSize = (int)bt_pix0;
 }
 
+LineVirtualIterator::LineVirtualIterator(const Point& pt1, const Point& pt2, int connectivity, bool forceLeftToRight)
+{
+  currentPosOffset = Point(std::min(pt1.x, pt2.x), std::min(pt1.y, pt2.y));
```
VadimLevin: Use `initializer_list` to initialize member

modules/imgproc/test/test_drawing.cpp
```diff
@@ -245,6 +248,46 @@ int CV_DrawingTest_CPP::checkLineIterator( Mat& img )
...
+           ts->printf( ts->LOG, "LineVirtualIterator works incorrect" );
+           ts->set_failed_test_info(cvtest::TS::FAIL_INVALID_OUTPUT);
+           break;
+        }
+        else
+        {
+            for(int i = 0; i < it.count; ++it, ++vit, i++ )
+            {
+                Point pIt = it.pos();
+                Point pVit = vit.pos();
```
VadimLevin: Can be removed

modules/imgproc/test/test_drawing.cpp
```diff
@@ -245,6 +248,46 @@ int CV_DrawingTest_CPP::checkLineIterator( Mat& img )
...
+        int width = randomGenerator.uniform(0, 512+1);
+        int height = randomGenerator.uniform(0, 512+1);
+        int x1 = randomGenerator.uniform(-512, 1024+1);
+        int y1 = randomGenerator.uniform(-512, 1024+1);
+        int x2 = randomGenerator.uniform(-512, 1024+1);
+        int y2 = randomGenerator.uniform(-512, 1024+1);
+        int channels = randomGenerator.uniform(1, 3+1);
+        Mat m(cv::Size(width, height), CV_MAKETYPE(8U, channels));
+        LineIterator it( m, Point(x1,y1), Point(x2, y2) );
+        LineVirtualIterator vit(Rect(0, 0, width, height), Point(x1,y1), Point(x2, y2) );
```
VadimLevin: This checks only the case when `currentPosOffset` is (0, 0), so can be reduced to constructor with the `Size` as the first argument. Consider to add case for non-zero offset too.

modules/imgproc/include/opencv2/imgproc.hpp
VadimLevin: As an alternative you give better names to parameters instead of `aSize`/`aRect`

VadimLevin:
Thank you for your work!
Please,  fix documentary string for `LineIterator` because you've changed the parameter name and add documentary strings for implemented functionality. But I suppose, last item should be done after @alalek approve code changes.

vpisarev:
@chacha21, thank you for the contribution! I think, in this class it makes sense to modify implementation of Bresenham algorithm to operate directly on the pixel coordinates. With current implementation it's so slow that we can basically use the original class, just add an extra constructor to it.

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4777,6 +4777,76 @@ Point LineIterator::pos() const
...
+LineVirtualIterator& LineVirtualIterator::operator ++()
+{
+    if (size.width)
+    {
+      const int mask = err < 0 ? -1 : 0;
+      err += minusDelta + (plusDelta & mask);
+      const int offset = minusStep + (plusStep & mask);
+      const size_t flattenedCoord = (currentPos.y - currentPosOffset.y) * size.width + (currentPos.x - currentPosOffset.x) + offset;
+      currentPos.y = static_cast<int>((flattenedCoord / size.width) + currentPosOffset.y);
+      currentPos.x = static_cast<int>((flattenedCoord % size.width) + currentPosOffset.x);
```
vpisarev: basically, I mean that this 2 lines should be removed; instead, you should update currentPos directly using the Bresenham algorithm

chacha21: The current implementation is rather far from standard Bresenham's algorithm descriptions, so it is not that straightfoward to rewrite differently all the cases in both 4 and 8 connectivity.

asmorkalov:
@chacha21 @vpisarev do you have consensus on the solution?

vpisarev:
@chacha21, @asmorkalov, I've updated the implementation, made it one class, not two. Hopefully, it's good enough to be merged

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4754,7 +4762,9 @@ LineIterator& LineIterator::operator ++()
 {
     int mask = err < 0 ? -1 : 0;
     err += minusDelta + (plusDelta & mask);
-    ptr += minusStep + (plusStep & mask);
+    p.x += minusShift + (plusShift & mask);
+    p.y += minusStep + (plusStep & mask);
+    ptr = (uchar*)ptr0 + (size_t)p.y*step + (size_t)p.x*elemSize;
```
alalek: OpenCV drawing functions are not optimized at all.
This change may bring unnecessary performance degradation to existed functionality.

---

It is impossible to run UBSAN on that code (adding offset to NULL pointer is an ERROR). Adding one more check is performance degradation again.

---

See comments in previous patch about that: #13843 (**mentioned** in PR description)

vpisarev: ok, I've eliminated the performance degradation and also made the code UBSAN compatible

vpisarev:
@alalek, can we merge it?

PR #13875, opened 2019-02-20
DNN: Optimization: Fuse Deconvolution -> BatchNorm

allnes:
```
force_builders=Custom
buildworker:Custom=linux-2
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
test_modules:Custom=dnn
```

### DIFF
--- a/modules/dnn/src/layers/convolution_layer.cpp
+++ b/modules/dnn/src/layers/convolution_layer.cpp
@@ -61,6 +61,8 @@ namespace dnn
 class BaseConvolutionLayerImpl : public ConvolutionLayer
 {
 public:
+    bool newWeightAndBias;
+    std::vector<double> weightsMultipliers;
     BaseConvolutionLayerImpl(const LayerParams &params)
     {
         setParamsFrom(params);
@@ -84,6 +86,8 @@ public:
         CV_Assert(numOutput % ngroups == 0);
         CV_Assert(adjustPad.width < stride.width &&
                   adjustPad.height < stride.height);
+
+        newWeightAndBias = false;
     }
 
     void finalize(InputArrayOfArrays inputs_arr, OutputArrayOfArrays outputs_arr) CV_OVERRIDE
@@ -133,6 +137,20 @@ public:
         (stride.height == 1 && stride.width == 1) &&
         (dilation.height == 1 && dilation.width == 1);
     }
+
+    virtual bool tryFuse(Ptr<Layer>& top) CV_OVERRIDE
+    {
+        Mat w, b;
+        top->getScaleShift(w, b);
+        if (!w.empty() || !b.empty())
+        {
+            fuseWeights(w, b);
+            return true;
+        }
+        return false;
+    }
+
+    virtual void fuseWeights(const Mat& w_, const Mat& b_) = 0;
 
     virtual void applyHalideScheduler(Ptr<BackendNode>& node,
                                       const std::vector<Mat*> &inputs,
@@ -184,11 +202,9 @@ class ConvolutionLayerImpl CV_FINAL : public BaseConvolutionLayerImpl
 public:
     enum { VEC_ALIGN = 8, DFT_TYPE = CV_32F };
     Mat weightsMat;
-    std::vector<double> weightsMultipliers;
     std::vector<float> biasvec;
     std::vector<float> reluslope;
     Ptr<ActivationLayer> activ;
-    bool newWeightAndBias;
     bool fusedBias;
 
 #ifdef HAVE_OPENCL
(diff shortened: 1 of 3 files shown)


modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -1245,6 +1246,52 @@ class DeConvolutionLayerImpl CV_FINAL : public BaseConvolutionLayerImpl
...
+    {
+        Mat w, b;
+        top->getScaleShift(w, b);
+        if (!w.empty() || !b.empty())
+        {
+            fuseWeights(w, b);
+            return true;
+        }
+        return false;
+    }
```
dkurt: This method looks similar to Convolution's `tryFuse`. May I ask you remove them both and define `tryFuse` at `BaseConvolutionLayerImpl`?

modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -1245,6 +1246,52 @@ class DeConvolutionLayerImpl CV_FINAL : public BaseConvolutionLayerImpl
...
+        CV_Assert_N(!weightsMat.empty(), w.empty() || outCn == w.total(), b.empty() || outCn == b.total());
+        if (!w.empty())
+        {
+            if (weightsMat.data == blobs[0].data) { weightsMat = weightsMat.clone(); }
+            Mat originWeights = weightsMat.clone();
+            for (int i = 0; i < outCn; ++i)
+            {
+                double wi = w.at<float>(i);
+                weightsMultipliers[i] *= wi;
+                cv::multiply(originWeights.row(i), weightsMultipliers[i], weightsMat.row(i));
```
dkurt: That looks confusing. `weightsMat` is initialized with as `weightsMat = (blobs[0].reshape(1, inpCn)).clone();` (with number of rows equals number of input channels) but in this loop we iterate over rows by number output channels.

modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -1245,6 +1246,52 @@ class DeConvolutionLayerImpl CV_FINAL : public BaseConvolutionLayerImpl
...
+                biasesMat.at<float>(i) *= wi;
+            }
+        }
+
+        if (!b.empty())
+        {
+            for (int i = 0; i < outCn; ++i) { biasesMat.at<float>(i) += b.at<float>(i); }
```
dkurt: Can we use `cv::add` or operator `+=` instead?

modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -1245,6 +1248,42 @@ class DeConvolutionLayerImpl CV_FINAL : public BaseConvolutionLayerImpl
...
+        {
+            transpose(blobs[0].reshape(1, blobs[0].size[0]), weightsMat);
+            biasesMat = hasBias() ? blobs[1].reshape(1, blobs[1].size[0])
+                                  : Mat::zeros(blobs[1].size[0], 1, CV_32F);
+        }
+    }
+
+    void fuseWeights(const Mat& w_, const Mat& b_) CV_OVERRIDE
+    {
+        const int outCn = weightsMat.size[0] / (blobs[0].size[2] * blobs[0].size[3]);
```
dkurt: You can use `numOutput` member.

modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -1245,6 +1248,42 @@ class DeConvolutionLayerImpl CV_FINAL : public BaseConvolutionLayerImpl
 
         pad.width = pad_l;
         pad.height = pad_t;
+
+        weightsMultipliers.assign(blobs[0].size[0], 1.0);
```
dkurt: Number of multipliers should be equal number of output channels (numOutput). However `blobs[0]` is a weights matrix with shape `IOHW` where `I` is a number of input channels and `O` is a number of output channels divided by group number. So it's better to use `numOutput` instead of raw shapes of weigths blob.

modules/dnn/test/test_backends.cpp
```diff
@@ -308,6 +308,12 @@ TEST_P(DNNTestNetwork, FastNeuralStyle_eccv16)
         (backend == DNN_BACKEND_OPENCV && target == DNN_TARGET_OPENCL_FP16) ||
         (backend == DNN_BACKEND_INFERENCE_ENGINE && target == DNN_TARGET_MYRIAD))
```
dkurt: please also check if `backend == DNN_BACKEND_OPENCV && target == DNN_TARGET_OPENCL_FP16` and `backend == DNN_BACKEND_INFERENCE_ENGINE && target == DNN_TARGET_MYRIAD` configurations pass this test.

PR #13878, opened 2019-02-20
good practise

LaurentBerger:
check imread return value in  many languages

### DIFF
--- a/doc/py_tutorials/py_gui/py_image_display/py_image_display.markdown
+++ b/doc/py_tutorials/py_gui/py_image_display/py_image_display.markdown
@@ -32,18 +32,9 @@ import cv2 as cv
 
 # Load an color image in grayscale
 img = cv.imread('messi5.jpg',0)
-if img is None:
-    print("File cannot be read. Check file path")
-    print("ファイルを読み込めません。 ファイルパスを確認する")
-    print("ບໍ່ສາມາດອ່ານໄຟລ໌ໄດ້. ກວດເບິ່ງເສັ້ນທາງໄຟລ໌")
-    print("No se puede leer el archivo. Verifique la ruta del archivo")
-    print("Le fichier image ne peut etre lu. Verifier le chemin du fichier")
-    print("Datei kann nicht gelesen werden. Dateipfad überprüfen")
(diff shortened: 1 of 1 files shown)


vpisarev:
@LaurentBerger, thank you for the patch! Don't you think it's not quite scalable solution? I mean, we have many, many places in OpenCV itself and in the samples where we report various errors. Why put the error message in different languages here and not everywhere else?

LaurentBerger:
I don't want to write in all languages everywhere but only  in first tutorial about imread. One way for users to check the return value of imread is to add these few lines: humor makes memorizing easier

dkurt:
I think it's better to keep a single code snippet modified. However I'm not sure which one is better.

And I'd like to suggest a kind of challenge. An every user proposes one's native language message. So after some time we'll see if OpenCV is international project.

And one more suggestion is to replace error message to success message. I.e.

```python
if not img is None:
  print("Image read successfully.")
```

You know, keep things positive :smile:

LaurentBerger:
Ok let's go

alalek:
... and silence on errors - this is not good actually.

IMHO, We should be silent on success and output meaningful messages in case of errors.

> print("File cannot be read. Check file path")

It is not complete. Path can be fine, but image codec is missing. Or file is broken or mode is not supported.
Exceptions can handle that (current "bool" result is not able to do that).

LaurentBerger:
> 
> > print("File cannot be read. Check file path")
> 
I follow @dkurt idea so there is no such line in my code : 
```
if img is not None:
    print("Image read successfully.")
    print("Félicitation l'image est bien lue.")
```

alalek:
I know - not all ideas are good for all cases, including my own =)

What is about this?

```
if img is not None:
    print("Image read successfully: " + str(img.shape))
else:
   print...
   exit(1)
... processing ...
```

LaurentBerger:
When there is a problem on answers.opencv.org forum with imread we cannot refer to doc. My first idea was message in all language because don't check  imread return value it's a noob idea. @dkurt idea is a good idea too. That's open source spirit.

Now if somebody got a better idea  he can make a PR 
waiting for https://github.com/opencv/opencv/wiki/GSoC_2019#idea-improve-and-expand-python-version-of-opencv

PR #13807, opened 2019-02-12
update libpng to 1.6.36

tailsu:
Sources copied from https://github.com/glennrp/libpng/archive/v1.6.36.tar.gz

PR #13901, opened 2019-02-23
Update libjpeg-turbo from version 1.5.3 to 2.0.2

reunanen:
Previous libjpeg-turbo update: #11497

<cut/>

```
allow_multiple_commits=1
```


alalek:
Take a look on comments from #13807 (most of them are applied to any 3rdparty updates)

reunanen:
> Take a look on comments from #13807 (most of them are applied to any 3rdparty updates)

Thanks, will do... at some point.

PR #13902, opened 2019-02-25
Add namespace cv:: for  consistency

insurgent92:


alalek:
Code is already under `namespace cv`.
So, `cv::` should be removed instead, but ...

Changes for changes must be avoided (they don't fix anything actually).
Lets don't introduce unnecessary "merging" hell.

If you want to fix something in OpenCV, fell free to pick something from here: https://github.com/opencv/opencv/issues

PR #13904, opened 2019-02-25
fix tf_text_graph_common tensor_content type bug

cftang0827:
### This pullrequest changes

```
Tensorflow version: 1.12.0
OpenCV version: 3.4.3
```

Fix the bug in creating tensorflow model's pbtxt file. 
The type of tensor.tensor_content should be `bytes` not `str` 

```
node.attr['value'].tensor.tensor_content = ""
--> 
node.attr['value'].tensor.tensor_content = bytes('', 'utf-8')
```

### DIFF
--- a/samples/dnn/tf_text_graph_common.py
+++ b/samples/dnn/tf_text_graph_common.py
@@ -324,6 +324,6 @@ def writeTextGraph(modelPath, outputPath, outNodes):
             for node in graph_def.node:
                 if node.op == 'Const':
                     if 'value' in node.attr and node.attr['value'].tensor.tensor_content:
-                        node.attr['value'].tensor.tensor_content = ''
+                        node.attr['value'].tensor.tensor_content = b''
 
         tf.train.write_graph(graph_def, "", outputPath, as_text=True)


dkurt:
@cftang0827, thank you! Could you please share a way how this bug affects import?

cftang0827:
Hi, actually it's in `opencv/samples/dnn/` , therefore, it would not affect import and cause any error.
However, if you want to use opencv to inference a tf-based model, you need to create a config file.

`https://github.com/opencv/opencv/wiki/TensorFlow-Object-Detection-API#generate-a-config-file`

That's the script that generate model related config file.
I also tested in opencv-4.0.0, there's no such problem because there's an implementation of `cv.dnn.writeTextGraph(modelPath, outputPath)`. However, if you use older version of opencv, the bug of type error would be showed up. 

```
def writeTextGraph(modelPath, outputPath, outNodes):
    try:
        import cv2 as cv

        cv.dnn.writeTextGraph(modelPath, outputPath)
    except:
        import tensorflow as tf
        from tensorflow.tools.graph_transforms import TransformGraph

        with tf.gfile.FastGFile(modelPath, 'rb') as f:
            graph_def = tf.GraphDef()
            graph_def.ParseFromString(f.read())

            graph_def = TransformGraph(graph_def, ['image_tensor'], outNodes, ['sort_by_execution_order'])

            for node in graph_def.node:
                if node.op == 'Const':
                    if 'value' in node.attr and node.attr['value'].tensor.tensor_content:
                        node.attr['value'].tensor.tensor_content = bytes('', 'utf-8')

        tf.train.write_graph(graph_def, "", outputPath, as_text=True)
```

Thanks.

dkurt:
@cftang0827, This code part depends on TensorFlow version only. In example, for TensorFlow `1.12.0-rc1` I have `<type 'str'>` for `type(node.attr['value'].tensor.tensor_content)`.

samples/dnn/tf_text_graph_common.py
```diff
@@ -324,6 +324,6 @@ def writeTextGraph(modelPath, outputPath, outNodes):
             for node in graph_def.node:
                 if node.op == 'Const':
                     if 'value' in node.attr and node.attr['value'].tensor.tensor_content:
-                        node.attr['value'].tensor.tensor_content = ''
+                        node.attr['value'].tensor.tensor_content = bytes('', 'utf-8')
```
dkurt: I'd like to propose a bit different solution:

```python
node.attr['value'].tensor.tensor_content = b''
```
I can now reproduce the problem (tested with different versions of TF and python).

dkurt:
@cftang0827, please change the base branch from 3.4 to master (Please do not close pull request but just create a new branch with the same name `dnn_tf_text_graph_common_bug` from 3.4 and then force push it to your fork).

samples/dnn/tf_text_graph_common.py
cftang0827: Thank you for your suggestion, and I think it's the better and simpler solution. Thank you.

Issue #13895, opened 2019-02-22
SSD object detector returns too large bboxes for non-square inputs


mahehu:
##### System information (version)
- OpenCV => 4.0.0.21 installed via pip
- Tensorflow => 1.12.
- Operating System / Platform => Ubuntu 18.04
- Compiler => Installed via pip
- Platform => Anaconda Python 3.6.8

##### Detailed description

An SSD object detector trained on Tensorflow object detection API gives incorrect bounding boxes with non-square images. With square images (in training and testing), the results are exactly the same as those of native Tensorflow. When training and testing images are not square, opencv returns larger bounding boxes than native TF.

For example, in the non-square [image](http://www.cs.tut.fi/~hehu/ttmmpp/out0.jpg), of size `hxw = 450x325` the bounding boxes are as follows (green = TF, yellow = opencv):

`OpenCV    : Left 0.1078883, Top 0.0987248, Width 0.7832990, Height 0.6393267`
`TensorFlow: Left 0.2059162, Top 0.0987247, Width 0.5874741, Height 0.6393268`

`top` and `height` coordinates are exactly the same, but `left` and `width` are not. Width appears to be exactly 33.3% larger with opencv than TF (which seems to be related to the input aspect ratio--verified with other models). 
 
##### Steps to reproduce

* Download our [face detection model](http://www.cs.tut.fi/~hehu/ttmmpp/ssdmobv1depth0.75_240x180.zip) trained on TF object detection api.
* Download the [example image](http://www.cs.tut.fi/~hehu/ttmmpp/0.jpg)
* Execute the following python code:
```
import cv2 as cv
import tensorflow as tf

cvNet = cv.dnn.readNetFromTensorflow('ssdmobv1depth0.75_240x180/frozen_inference_graph.pb',
                                     'ssdmobv1depth0.75_240x180/graph.pbtxt')

# Read the graph.
with tf.gfile.FastGFile('ssdmobv1depth0.75_240x180/frozen_inference_graph.pb', 'rb') as f:
    graph_def = tf.GraphDef()
    graph_def.ParseFromString(f.read())

with tf.Session() as sess:
    # Restore session
    sess.graph.as_default()
    tf.import_graph_def(graph_def, name='')

    img = cv.imread('0.jpg')
    rows = img.shape[0]
    cols = img.shape[1]

    # Run OpenCV
    cvNet.setInput(cv.dnn.blobFromImage(img, size=(240, 180), swapRB=True, crop=False))
[...]

PR #13906, opened 2019-02-25
Fix TensorFlow SSDs with non-square inputs

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/13895


alalek:
@dkurt Could you add some link with corresponding code in TF into description?

dkurt:
@alalek, Sure. Here it is:

https://github.com/tensorflow/models/blob/02a9969e94feb51966f9bacddc1836d811f8ce69/research/object_detection/anchor_generators/multiple_grid_anchor_generator.py#L211-L213

> base_anchor_size: base anchor size as [height, width].
        The height and width values are normalized to the minimum dimension of the
        input height and width, so that when the base anchor height equals the
        base anchor width, the resulting anchor is square even if the input image
        is not square.

PR #13909, opened 2019-02-25
OE-11 Logging revamp (#11003) for 4.2.x (draft (logging_20190220))

kinchungwong:
```
force_builders=linux,docs
```

### This pullrequest changes

Logging revamp for OpenCV 4.1.x, [OE-11](https://github.com/opencv/opencv/wiki/OE-11.-Logging), issue #11003

This PR replaces #13642 because the old PR was not salvageable.

Please keep all high-level or design-level discussions to the issue thread #11003 to keep the dialogue in one place.

Rebase and squash will happen, therefore keep in mind *code review comments might be lost*. Please put *persistent* comments (specific to the code changes in this PR and which are not high-level design issues) in this PR.

### Status

**Do not merge yet** - the PR contains numerous behavior changes that still need to be ironed out for backward compatibility. Also missing "meaningful" performance tests.

### Checklist 

- [x] Implement line-of-code information in logging output
  - **The logging output format is changed; this is an observable library behavior change.**
- [x] Implement runtime configurable tag-based log level threshold
- [ ] Improve startup-time log level configuration (parsing support for tags and levels)
  - *(startup-time refers to e.g. environment variable ```OPENCV_LOG_LEVEL```)*
- [x] Implement limited wildcard support for startup-time log level config for tags
  - [x] Fixed: should behave correctly whether environment config string is parsed and applied before or after the registration of tags.

### Regarding log config for tags at startup-time 

*(cross reference https://gist.github.com/kinchungwong/37b3eabea7cd7bef3056ca9763816db7)*

- Config string format is similar to **```logcat```**.
  - ```imgproc:D imgcodecs.*:V *:S```
- Actual tag names (in library code) can contain the period (```"."```)
  - *(Undecided)* "imgcodecs.tiff.decoder" or "imgcodecs.decoders.tiff"
- Wildcard support is limited to asterisk (```"*"```) and are **anchored to the "name parts" obtained from splitting tag names at the delimiter, the period (```"."```)**. 
- For example, if a tag name is ```"imgcodecs.tiff.decoder"```, the name parts are ```"imgcodecs"```, ```"tiff"```, and ```"decoder"```. It can be matched by:
  - its exact name, or 
  - its first name part, ```"imgcodecs*"```, by ending with an asterisk, or 
  - a name part that appears in any position, such as ```"*tiff*"``` or ```"*decoder*"```, by surrounding with a pair of asterisks.
- However, it will not match any of the following: ```"img*"```, ```"*code*"```, ```"*codecs"```, ```"*tiff.decoder*"```, ```"*coder"```, because they don't match anything aligned to the period delimiter.
- This is done because, in actual practice, unrestricted substring matching is not as useful for OpenCV.
- Users needing better tag filtering should redirect OpenCV log to a different filtering tool.

### DIFF
--- a/modules/core/include/opencv2/core/utils/logger.defines.hpp
+++ b/modules/core/include/opencv2/core/utils/logger.defines.hpp
@@ -17,6 +17,26 @@
 #define CV_LOG_LEVEL_DEBUG 5           //!< Debug message. Disabled in the "Release" build.
 #define CV_LOG_LEVEL_VERBOSE 6         //!< Verbose (trace) messages. Requires verbosity level. Disabled in the "Release" build.
 
+namespace cv {
+namespace utils {
+namespace logging {
+
+//! Supported logging levels and their semantic
+enum LogLevel {
+    LOG_LEVEL_SILENT = 0,              //!< for using in setLogVevel() call
+    LOG_LEVEL_FATAL = 1,               //!< Fatal (critical) error (unrecoverable internal error)
+    LOG_LEVEL_ERROR = 2,               //!< Error message
+    LOG_LEVEL_WARNING = 3,             //!< Warning message
+    LOG_LEVEL_INFO = 4,                //!< Info message
+    LOG_LEVEL_DEBUG = 5,               //!< Debug message. Disabled in the "Release" build.
+    LOG_LEVEL_VERBOSE = 6,             //!< Verbose (trace) messages. Requires verbosity level. Disabled in the "Release" build.
+#ifndef CV_DOXYGEN
+    ENUM_LOG_LEVEL_FORCE_INT = INT_MAX
+#endif
+};
+
+}}} // namespace
+
 //! @}
 
 #endif // OPENCV_LOGGER_DEFINES_HPP
--- a/modules/core/include/opencv2/core/utils/logger.hpp
+++ b/modules/core/include/opencv2/core/utils/logger.hpp
@@ -10,6 +10,7 @@
 #include <limits.h> // INT_MAX
 
 #include "logger.defines.hpp"
+#include "logtag.hpp"
 
 //! @addtogroup core_logging
 // This section describes OpenCV logging utilities.
@@ -20,20 +21,6 @@ namespace cv {
 namespace utils {
 namespace logging {
 
-//! Supported logging levels and their semantic
-enum LogLevel {
-    LOG_LEVEL_SILENT = 0,              //!< for using in setLogVevel() call
-    LOG_LEVEL_FATAL = 1,               //!< Fatal (critical) error (unrecoverable internal error)
-    LOG_LEVEL_ERROR = 2,               //!< Error message
-    LOG_LEVEL_WARNING = 3,             //!< Warning message
-    LOG_LEVEL_INFO = 4,                //!< Info message
-    LOG_LEVEL_DEBUG = 5,               //!< Debug message. Disabled in the "Release" build.
-    LOG_LEVEL_VERBOSE = 6,             //!< Verbose (trace) messages. Requires verbosity level. Disabled in the "Release" build.
-#ifndef CV_DOXYGEN
-    ENUM_LOG_LEVEL_FORCE_INT = INT_MAX
-#endif
-};
-
 /** Set global logging level
 @return previous logging level
 */
@@ -41,15 +28,19 @@ CV_EXPORTS LogLevel setLogLevel(LogLevel logLevel);
 /** Get global logging level */
 CV_EXPORTS LogLevel getLogLevel();
 
+/** Get global log tag */
+CV_EXPORTS cv::utils::logging::LogTag* getGlobalLogTag();
+
 namespace internal {
 /** Write log message */
 CV_EXPORTS void writeLogMessage(LogLevel logLevel, const char* message);
+CV_EXPORTS void writeLogMessageEx(LogLevel logLevel, const char* tag, const char* file, int line, const char* func, const char* message);
 } // namespace
 
 /**
  * \def CV_LOG_STRIP_LEVEL
  *
- * Define CV_LOG_STRIP_LEVEL=CV_LOG_LEVEL_[DEBUG|INFO|WARN|ERROR|FATAL|DISABLED] to compile out anything at that and before that logging level
+ * Define CV_LOG_STRIP_LEVEL=CV_LOG_LEVEL_[DEBUG|INFO|WARN|ERROR|FATAL|SILENT] to compile out anything at that and before that logging level
  */
 #ifndef CV_LOG_STRIP_LEVEL
 # if defined NDEBUG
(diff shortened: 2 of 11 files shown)


modules/core/src/utils/logtagconfigparser.cpp
```diff
@@ -0,0 +1,218 @@
...
+{
+    for (const auto& malformed : m_malformed)
+    {
+        func(malformed);
+    }
+}
+
+void LogTagConfigParser::segmentTokens()
+{
+	const size_t npos = std::string::npos;
```
kinchungwong: This will be fixed in the next commit.
```
/build/precommit_linux64_no_opt/opencv/modules/core/src/utils/logtagconfigparser.cpp: In member function 'void cv::utils::logging::LogTagConfigParser::segmentTokens()':
/build/precommit_linux64_no_opt/opencv/modules/core/src/utils/logtagconfigparser.cpp:55:15: warning: unused variable 'npos' [-Wunused-variable]
  const size_t npos = std::string::npos;
               ^
```

modules/core/src/utils/logtagconfigparser.hpp
```diff
@@ -0,0 +1,54 @@
+// This file is part of OpenCV project.
```
kinchungwong: Tabs in source files (logtagconfigparser.hpp, cpp) will be changed to whitespace in the next commit.
```
modules/core/src/utils/logtagconfigparser.cpp:18: tab in indent.
+	parse(input);
modules/core/src/utils/logtagconfigparser.cpp:27: tab in indent.
+	m_input = input;
modules/core/src/utils/logtagconfigparser.cpp:28: tab in indent.
+	segmentTokens();
modules/core/src/utils/logtagconfigparser.cpp:29: tab in indent.
+	return (m_malformed.empty());
modules/core/src/utils/logtagconfigparser.cpp:55: tab in indent.
...
```

modules/core/src/utils/logtagmanager.hpp
```diff
@@ -27,54 +32,99 @@ namespace logging {
...
-    using LockType = std::lock_guard<MutexType>;
+    static std::vector<std::string> splitNameParts(const std::string& fullName);
 
-    struct LogTagAndId
-    {
-        size_t id;
-        LogTag* value;
-    };
+private:
+    static constexpr char* m_globalName = "global";
```
kinchungwong: Should be ```static constexpr const char*```. Will be fixed on next commit.

```
/build/precommit_linux64/opencv/modules/core/src/utils/logtagmanager.hpp:118:43: warning: ISO C++ forbids converting a string constant to 'char*' [-Wwrite-strings]
```

modules/core/src/utils/logtagmanager.cpp
```diff
@@ -62,76 +82,205 @@ void LogTagManager::unassign(const std::string& name)
...
+    {
+        assignWildcardInfo(fullNamePair, namePart, isFirstNamePart);
+        isFirstNamePart = false;
+    }
+}
+
+void LogTagManager::assignWildcardInfo(const FullNamePair& fullNamePair, const std::string& namePart, bool isFirst)
+{
+    const auto memberIdAndPtr = fullNamePair.second.member;
+    const size_t id = memberIdAndPtr.id;
```
kinchungwong: Unused variable will be removed in next commit.
```
/build/precommit_linux64/opencv/modules/core/src/utils/logtagmanager.cpp:207:18: warning: unused variable 'id' [-Wunused-variable]
```

modules/core/test/test_logtagconfigparser.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+#include "test_precomp.hpp"
+#include <opencv2/core/utils/logger.hpp>
+#include "../src/utils/logtagmanager.hpp"
+#include "../src/utils/logtagconfigparser.hpp"
+
+// Because "LogTagConfigParser" isn't exported from "opencv_core", the only way
+// to perform test is to compile the source code into "opencv_test_core".
+// This workaround may cause step debugger breakpoints to work unreliably.
+#if 1
```
kinchungwong: Trailing whitespace will be removed on next commit.
```
modules/core/test/test_logtagconfigparser.cpp:13: trailing whitespace.
+#if 1 
```

alalek: Consider enabling of commit hook to avoid that: https://github.com/opencv/opencv/wiki/Linux_contribution_guide#git-configuration (step 2)

kinchungwong: Thanks, I forgot to do that after I set up the clone from scratch.

kinchungwong:
https://github.com/opencv/opencv/pull/13909/commits/b3f47cd7cbfef17d4c561522c166042604975a70 does not fix all problems of order dependency between log tag registration and config parsing/application. The next commit will focus on testing (formalizing) this aspect of LogTagManager and see how it can be fixed.

kinchungwong:
@alalek 

### About LogTag auto-registration, usage example, code impact

#### Commit ca9d0b1f7 

(timestamp 2019-03-06T04:29:40Z) 
Added LogTagAuto for auto registration in ctor

**This commit implements a class LogTagAuto** which is derived from LogTag and which 
auto-registers with LogTagManager at constructor time. The intended use is to
define static instances of LogTagAuto in namespace scope, and then refer to this
tag either explicitly (when calling CV_LOG_WITH_TAG), or implicitly (by re-defining
CV_LOGTAG_FALLBACK within a compilation unit (cpp file)).

----

#### Commit 0a50c6f00 

(timestamp 2019-03-06T04:33:30Z) 
EXCLUDE_ON_NEXT_REBASE demo logtag usage

This demo-only commit **shows how LogTagAuto and CV_LOGTAG_FALLBACK are used together.**

The demo change is applied to a limited number of files, in opencv_core:
- ocl.cpp
- datafile.cpp, 
- filesystem.cpp.

----

For each compilation unit (cpp file), the changes are:
- Initialize a static instance of LogTagAuto with instance name, the log tag string,
  the default log filtering threshold level in absence of any applicable configuration.
- The re-definition of the CV_LOGTAG_FALLBACK macro, which causes all the CV_LOG_*
  macro invocations that specify NULL as the tag argument to automatically pick up
  the LogTagAuto static variable that we just defined.

The minimum number of lines of change per compilation unit is 3:
- One line for LogTagAuto
- Two lines for re-defining a macro (```#undef``` and ```#define```). 
  - We can safely assume CV_LOGTAG_FALLBACK is always defined (as an identity function, in logger.hpp), so we don't need to check for existence.
  - Because this involves preprocessor directives, it is not possible to "automate" with preprocessor magic.

The use of the anonymous namespace for the LogTagAuto is optional, as it currently serves no purpose.

----

### Anticipation of introduced difficulties with respect to ongoing merging and backporting activities between 3.4 and master branches.

The insertion of LogTagAuto declaration is straightforward and could be easily scripted.
However, the nuance that require human insight and involvement is to determine the
line number on which the LogTagAuto declaration and macro redefinition shall be inserted.

It is possible to develop scripting tools that can temporarily remove and re-apply
these lines of code, but a human reviewer will need to provide some hints to this tool
so that those lines are always inserted at the correct place.

kinchungwong:
```
**WIP**
force_builders=linux,docs
```

kinchungwong:
Some non-critical commits have been taken off this branch (logging_20190220) and moved to dead-end branches:

- ```logging_20190220_demo_usage```
  - [(view dead-end branch on GitHub)](https://github.com/kinchungwong/opencv/commits/logging_20190220_demo_usage)
  - Modifies two files to show how the LogTag would be added.
  - Taken off PR because LogTag declaration details are subject to change.
- ```logging_20190220_docs```
  - [(view dead-end branch on GitHub)](https://github.com/kinchungwong/opencv/commits/logging_20190220_docs)
  - Tests docs build.
  - Taken off PR because implementation behavior are subject to change.
- ```logging_20190220_perf```
  - [(view dead-end branch on GitHub)](https://github.com/kinchungwong/opencv/commits/logging_20190220_perf)
  - Performance test for LogTagManager and for CV_LOG_WITH_TAG macro.
  - Taken off PR because performance testing methodology hasn't been agreed upon. 
  - However, one of the test was able to show a potential performance issue when many LogTags (several thousands) have the same name part (e.g. prefix, e.g. "core.*" or "imgproc.*") which is an ```O(N**2)``` bug but which normally isn't a concern because there normally wouldn't be thousands of LogTags. The estimated number of LogTags having same name part is lower, typically in the tens.

kinchungwong:
See spec revised 2019-03-20 https://github.com/kinchungwong/opencv_OE11_logging_11003_wiki/blob/master/spec_revised_20190320.md

modules/core/src/utils/logtagconfig.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+struct LogTagConfig
+{
+    std::string namePart;
+    LogLevel level;
+    bool isGlobal;
+    bool hasPrefixWildcard;
+    bool hasSuffixWildcard;
+
+    explicit LogTagConfig()
```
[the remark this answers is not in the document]
kinchungwong: Will fix.

modules/core/include/opencv2/core/utils/logger.defines.hpp
```diff
@@ -17,6 +17,26 @@
 #define CV_LOG_LEVEL_DEBUG 5           //!< Debug message. Disabled in the "Release" build.
 #define CV_LOG_LEVEL_VERBOSE 6         //!< Verbose (trace) messages. Requires verbosity level. Disabled in the "Release" build.
 
+namespace cv {
+namespace utils {
+namespace logging {
+
+//! Supported logging levels and their semantic
+enum LogLevel {
```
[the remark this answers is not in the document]
kinchungwong: Several other header files (log tag manager and log config parser) need to see the enum definition `LogLevel`. Because `LogLevel` is a traditional (non-strongly-typed) enum, and because some code refers to the members of LogLevel (the individual values), they need to include a header that contains the enum, but not the remainder of the content of logger.hpp (to avoid circular header dependency). Therefore, the enum definition `LogLevel` and its members are moved to `logger.defines.hpp` which is a convenient choice because it only contains "defines" (and by a stretch, the definition of a enum), and nothing else, which solves the difficulty described above.

modules/core/include/opencv2/core/utils/logger.hpp
```diff
@@ -20,36 +21,49 @@ namespace cv {
...
 /** Write log message */
 CV_EXPORTS void writeLogMessage(LogLevel logLevel, const char* message);
+
+/** Write log message */
+CV_EXPORTS void writeLogMessageEx(LogLevel logLevel, const char* tag, const char* file, int line, const char* func, const char* message);
+
 } // namespace
 
+struct CV_EXPORTS LogTagAuto
+    : public LogTag
```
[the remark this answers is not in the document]
kinchungwong: will do.

modules/core/include/opencv2/core/utils/logger.hpp
```diff
@@ -59,7 +73,85 @@ CV_EXPORTS void writeLogMessage(LogLevel logLevel, const char* message);
...
+// non-null. Do not re-define.
+#define CV_LOGTAG_GLOBAL cv::utils::logging::internal::getGlobalLogTag()
+
+#define CV_LOG_WITH_TAG(tag, msgLevel, ...) \
+    for(;;) { \
+        const auto cv_temp_msglevel = (cv::utils::logging::LogLevel)(msgLevel); \
+        if (cv_temp_msglevel >= (CV_LOG_STRIP_LEVEL)) break; \
+        auto cv_temp_logtagptr = CV_LOGTAG_PTR_CAST(CV_LOGTAG_EXPAND_NAME(tag)); \
+        if (!cv_temp_logtagptr) cv_temp_logtagptr = CV_LOGTAG_PTR_CAST(CV_LOGTAG_FALLBACK); \
+        if (!cv_temp_logtagptr) cv_temp_logtagptr = CV_LOGTAG_PTR_CAST(CV_LOGTAG_GLOBAL); \
```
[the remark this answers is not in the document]
kinchungwong: I wrote this in the most verbose way in order to make its internal workings obvious, for code review (design review) purpose. 

The current code looks like it is doing null-pointer-coalescing (something available in C# as an operator), but there can be more logic added to it in between the lines. For example, it may perform additional log stripping checks. If it turns out the flexibility is not needed, we can discuss ways to simplify the code.

I am open to proposals that take away one or more "log tag injection mechanisms" - 

1. tag argument passed into the macro 
  - I believe it is worth keeping
2. "re-definable" `CV_LOGTAG_FALLBACK` macro 
  - good for getting existing OpenCV code to provide log tags while minimizing scope of code changes
  - bad because may confuse many C++ analysis and refactoring tools. Some tools do not support/allow re-defining macros.
3. ultimate `CV_LOGTAG_GLOBAL` macro
  - alternative: we can simply combine this functionality with `CV_LOGTAG_FALLBACK`, in other words, the default value for `CV_LOGTAG_FALLBACK` would be the global log tag.
    - but if we want to accept this alternative (getting rid of `CV_LOGTAG_GLOBAL`), first we need accept that macros can be re-defined.

modules/core/include/opencv2/core/utils/logger.hpp
```diff
@@ -59,7 +73,85 @@ CV_EXPORTS void writeLogMessage(LogLevel logLevel, const char* message);
...
+            cv_temp_logstream.str().c_str()); \
+        break; \
+    }
+
+#define CV_LOG_FATAL(tag, ...)   CV_LOG_WITH_TAG(tag, cv::utils::logging::LOG_LEVEL_FATAL, __VA_ARGS__)
+#define CV_LOG_ERROR(tag, ...)   CV_LOG_WITH_TAG(tag, cv::utils::logging::LOG_LEVEL_ERROR, __VA_ARGS__)
+#define CV_LOG_WARNING(tag, ...) CV_LOG_WITH_TAG(tag, cv::utils::logging::LOG_LEVEL_WARNING, __VA_ARGS__)
+#define CV_LOG_INFO(tag, ...) CV_LOG_WITH_TAG(tag, cv::utils::logging::LOG_LEVEL_INFO, __VA_ARGS__)
+#define CV_LOG_DEBUG(tag, ...) CV_LOG_WITH_TAG(tag, cv::utils::logging::LOG_LEVEL_DEBUG, __VA_ARGS__)
+#define CV_LOG_VERBOSE(tag, v, ...) CV_LOG_WITH_TAG(tag, (cv::utils::logging::LOG_LEVEL_VERBOSE + (int)(v)), __VA_ARGS__)
```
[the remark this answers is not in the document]
kinchungwong: Options:

1. Combine "logLevel" and "v" into one value.
    - A hack, with unintended consequences. I'm okay with removing this hack, but first we need to choose one of the alternatives below.
1. Logging macro will use both "logLevel" and "v" for log filtering. 
    - Difficult. I haven't found a way to implement this. If anyone has idea, please let me know.
1. Logging macro will pass "v" into `writeLogMessageEx` function, without processing it.
    - Logging macro is not responsible for filtering based on value of "v"
    - If `writeLogMessageEx` wants to filter value of "v", user need to provide a user-defined backend.
    - Normally, `writeLogMessageEx` just print the value of "v" as part of log message.
1. Logging macro will convert value of "v" into a string and concatenate it to the logging message. `writeLogMessageEx` will not receive the value unless it parses the log message.

modules/core/src/utils/logtagconfigparser.cpp
```diff
@@ -0,0 +1,280 @@
...
+        {
+            return result;
+        }
+        if (!result.second)
+        {
+            return falseDontCare;
+        }
+        std::string upper = s;
+        std::transform(upper.begin(), upper.end(), upper.begin(),
+            [](char arg) -> char { return (char)toupper(arg); });
```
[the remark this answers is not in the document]
kinchungwong: The function 
```cpp
std::pair<LogLevel, bool> LogTagConfigParser::parseLogLevel(const std::string& s)
```
has been completely reworked to make it easier to understand and review for correctness. Even though the rewritten version may look like there are many calls to `toUpperCase`, these calls are mutually exclusive, so that any call to `parseLogLevel` will invoke `toUpperCase` at most once.

kinchungwong:
The [forced push on 2019-04-20 from 8bec94b to 107db4c](https://github.com/opencv/opencv/compare/8bec94b9622b58078e31d3048bdeceaba7a2eee7..107db4c617699db807bba56b8cb260e6f94c4039) resolves some code review issues. However, the following issues require additional discussions:

- Simplifying logging macro with null pointer coalescing operator technique, OR a complete redesign of the multiple log tag injection mechanism [(discussion_r274110590)](https://github.com/opencv/opencv/pull/13909#discussion_r274110590)
- Supporting the verbose sub-level (the "v" argument) without resorting to hacks with unintended consequences [(discussion_r274114449)](https://github.com/opencv/opencv/pull/13909#discussion_r274114449)

kinchungwong:
Thanks @alalek for working with this long development and review process. I will update the check-list to clarify what items are completed and what items remain to be implemented in the future. 

Also, I will start a new PR that will add log tags into existing OpenCV source code (possibly into contrib modules as well). This may be yet another long process. Hopefully I can schedule things so that I can work on multiple things in parallel.

PR #13910, opened 2019-02-25
Fix Issue_13908

### LINKED ISSUE #13908: calib3D : cannot allocate matrix in OpenCV_4.0.1
- OpenCV : 4.0.1
- Operating System / Platform  : Opensuse15
- Compiler => gcc (SUSE Linux) 7.3.1 20180323 [gcc-7-branch revision 258812]

##### Detailed description

When updating my codebase from OpenCV 3.2.0 to OpenCV 4.0.1, i see an opencv error (cannot allocate memory) while trying to calibrate cameras with some big input matrices while it wasn't reporting any error on previous version.

-----------------------------
 error: (-4:Insufficient memory) Failed to allocate 224616847152 bytes in function 'OutOfMemoryError'
-----------------------------

##### Steps to reproduce
Call cv::calibrateCamera with a total number of objectPoints greater than 300,000 .

x3lif:
resolves #13908

fix the regression of cv::calibrateCamera describe in Issue#13908 by allocating the matrix _Jo only when it seems requested

### DIFF
--- a/modules/calib3d/src/calibration.cpp
+++ b/modules/calib3d/src/calibration.cpp
@@ -1585,8 +1585,10 @@ static double cvCalibrateCamera2Internal( const CvMat* objectPoints,
     
     Mat _Ji( maxPoints*2, NINTRINSIC, CV_64FC1, Scalar(0));
     Mat _Je( maxPoints*2, 6, CV_64FC1 );
-    Mat _Jo = (solver.state == CvLevMarq::CALC_J) ? Mat( maxPoints*2, maxPoints*3, CV_64FC1, Scalar(0) ) : Mat();
-    Mat _err( maxPoints*2, 1, CV_64FC1 );
+    Mat _err( maxPoints*2, 1, CV_64FC1 );    
+    
+    const bool allocJo = (solver.state == CvLevMarq::CALC_J) || stdDevs;
+    Mat _Jo = allocJo ? Mat( maxPoints*2, maxPoints*3, CV_64FC1, Scalar(0) ) : Mat();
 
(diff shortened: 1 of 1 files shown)


modules/calib3d/src/calibration.cpp
```diff
@@ -1725,15 +1728,15 @@ static double cvCalibrateCamera2Internal( const CvMat* objectPoints,
             CvMat _dpdf = cvMat(_Ji.colRange(0, 2));
             CvMat _dpdc = cvMat(_Ji.colRange(2, 4));
             CvMat _dpdk = cvMat(_Ji.colRange(4, NINTRINSIC));
-            CvMat _dpdo = cvMat(_Jo.colRange(0, ni * 3));
+            CvMat _dpdo = _Jo.empty() ? CvMat() : cvMat(_Jo.colRange(0, ni * 3));
```
terfendail: Looks like all `_dpd*` variables are used only inside `cvProjectPoints2Internal` call.
Wouldn't it be better to create corresponding `cvMat` structures inside the call directly?

x3lif: i let these cvMat initialisation to respect the original the code.
you are right, i'll update this.

x3lif: Since `cvProjectPoints2Internal` takes **cvMat*** 
Making the creation of _dp* matrices at call site makes gcc angry : 

> ./src/github/opencv/modules/calib3d/src/calibration.cpp:1732:61: error: taking address of temporary [-fpermissive]

i'll move them under the if statement to reduce the scope of these variables.

PR #13922, opened 2019-02-27
G-API: Remove incorrect assert in NV12 to RGB/BGR kernels

smirnov-alexey:
This pullrequest removes incorrect asserts in NV12 to RGB/BGR color conversion kernels and add a little documentation on other asserts

### DIFF
--- a/modules/gapi/include/opencv2/gapi/imgproc.hpp
+++ b/modules/gapi/include/opencv2/gapi/imgproc.hpp
@@ -120,8 +120,7 @@ namespace imgproc {
             GAPI_Assert(in_uv.chan == 2);
             GAPI_Assert(in_y.depth == CV_8U);
             GAPI_Assert(in_uv.depth == CV_8U);
-            GAPI_Assert(in_uv.size.width % 2 == 0);
-            GAPI_Assert(in_uv.size.height % 2 == 0);
+            // UV size should be aligned with Y
             GAPI_Assert(in_y.size.width == 2 * in_uv.size.width);
             GAPI_Assert(in_y.size.height == 2 * in_uv.size.height);
             return in_y.withType(CV_8U, 3); // type will be CV_8UC3;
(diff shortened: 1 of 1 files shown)


modules/gapi/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+cmake_minimum_required(VERSION 3.3)
```
alalek: Why is 3.3?

OpenCV requires 3.5.1+ itself. Lower versions are not tested at all (however they may work).

modules/gapi/cmake/standalone.cmake
```diff
@@ -1,3 +1,8 @@
+if("${CMAKE_BUILD_TYPE}" STREQUAL "")
+    debug_message(STATUS "CMAKE_BUILD_TYPE not defined, 'Release' will be used")
```
alalek: 1.
What is `debug_message`?

2.
```
if(NOT DEFINED CMAKE_BUILD_TYPE)
```

dmatveev:
How CMAKE changes are related to NV12 assertion changes?

modules/gapi/cmake/standalone.cmake
smirnov-alexey: Yep, my bad. Will be fixed

smirnov-alexey:
@dmatveev It was some WIP stuff for testing standalone build. Will move cmake changes into another PR

modules/gapi/CMakeLists.txt
smirnov-alexey: IE uses G-API standalone build for its stuff but required version of cmake is 3.3 in IE. So, I think it's reasonable to put lower version in here

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -120,8 +120,11 @@ namespace imgproc {
             GAPI_Assert(in_uv.chan == 2);
             GAPI_Assert(in_y.depth == CV_8U);
             GAPI_Assert(in_uv.depth == CV_8U);
+            // We should check that uv width is even
+            // for the reason it must contain 2 pixels (u and v)
+            // for each output pixel
```
dmatveev: UV plane is a two-channel Mat. Regardless its width in pixels, every pixel _always_ contains two elements (U and V).

smirnov-alexey: You are right. uv.width assert is also wrong

PR #13932, opened 2019-02-28
Fix precision in tests for MyriadX

l-bat:
```
force_builders=Custom
buildworker:Custom=linux-2
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
test_modules:Custom=dnn

**WIP**
docker_image:Linux AVX2=ubuntu-openvino-2018r5:16.04
test_opencl:Linux AVX2=OFF
build_parallel_tests:Linux AVX2=1
test_bigdata:Linux AVX2=1
test_module:Linux AVX2=dnn,python2,python3,java
tests_filter:Linux AVX2=*
#tests_filter:Linux AVX2=*:-Test_Caffe_layers.Conv_Elu/1
```

### DIFF
--- a/modules/dnn/src/op_inf_engine.cpp
+++ b/modules/dnn/src/op_inf_engine.cpp
@@ -12,6 +12,7 @@
 #ifdef HAVE_INF_ENGINE
 #include <ie_extension.h>
 #include <ie_plugin_dispatcher.hpp>
+#include <vpu/vpu_plugin_config.hpp>
 #endif  // HAVE_INF_ENGINE
 
 namespace cv { namespace dnn {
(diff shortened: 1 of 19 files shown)


alalek:
There are [two failed tests](http://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/1848) with NCS 1.

modules/dnn/src/op_inf_engine.cpp
```diff
@@ -683,6 +683,80 @@ static std::map<InferenceEngine::TargetDevice, InferenceEngine::InferenceEngineP
...
+                                                 InferenceEngine::VPUConfigParams::VPU_2480}});
+        auto infRequest = netExec.CreateInferRequest();
+        infRequest.SetInput(inputBlobs);
+        infRequest.SetOutput(outputBlobs);
+        infRequest.Infer();
+    } catch(...) {
+        return false;
+    }
+    return true;
+}
```
dkurt: I believe that we can check it by less number of lines:
* Do we need to `CreateInferRequest` and subsequent steps or plugin throws an exception at `LoadNetwork`?
* Is it possible to create an empty `CNNNetwork` or simplify a builder? In example, keep only input and output layer.

modules/dnn/src/op_inf_engine.hpp
```diff
@@ -18,6 +18,7 @@
 #pragma GCC diagnostic ignored "-Wsuggest-override"
 #endif
 #include <inference_engine.hpp>
+#include <vpu/vpu_plugin_config.hpp>
```
dkurt: I think we can move it to a `.cpp` file.

modules/dnn/test/test_halide_layers.cpp
```diff
@@ -167,6 +167,13 @@ TEST_P(Deconvolution, Accuracy)
         throw SkipTestException("Test is disabled for OpenVINO 2018R4");
 #endif
 
+#if defined(INF_ENGINE_RELEASE) && INF_ENGINE_RELEASE >= 2018060000
+    if (backendId == DNN_BACKEND_INFERENCE_ENGINE && targetId == DNN_TARGET_MYRIAD &&
+        outChannels == 4 && group == 1 && kernel == Size(3, 1) && pad == Size(0, 1) &&
+        stride == Size(1, 1))
+        throw SkipTestException("Test is disabled for OpenVINO 2018R6");
+#endif
```
dkurt: Can we determine which hyperparameter causes error and skip problem layer is MyriadX is detected?

modules/dnn/src/op_inf_engine.cpp
alalek: Some notes:
- CreateInferRequest() call throws exception.
- empty `CNNNetwork` doesn't work (vpu plugin is crashed).

alalek:
@dkurt @mshabunin Please take a look on update of public header files.

modules/dnn/src/op_inf_engine.cpp
```diff
@@ -683,6 +689,61 @@ static std::map<InferenceEngine::TargetDevice, InferenceEngine::InferenceEngineP
...
+    builder.addLayer({InferenceEngine::PortInfo(clampId)},
+                      InferenceEngine::Builder::OutputLayer().setPort(InferenceEngine::Port({},
+                      InferenceEngine::Precision::FP16)));
+#endif
+
+    InferenceEngine::CNNNetwork cnn = InferenceEngine::CNNNetwork(
+                                      InferenceEngine::Builder::convertToICNNNetwork(builder.build()));
+
+    InferenceEngine::TargetDevice device = InferenceEngine::TargetDevice::eMYRIAD;
+    auto& sharedPlugins = getSharedPlugins();
```
alalek: `sharedPlugins` must be guarded by:
```
AutoLock lock(getInitializationMutex());
```

see #13738

PR #13939, opened 2019-03-01
Android build improvements

### LINKED ISSUE #13938: Android build should check well known locations for NDK/cmake/ninja
- OpenCV => latest

##### Detailed description
In order to ease the pain for someone getting started on Android the following build improvements would be beneficial:
* check well known locations under the Android SDK directory for NDK/cmake/ninja
* if ccache is not found then disable ccache in the build
* better log output to indicate the cause of build failures

I have a PR prepared to address this issue

komakai:
### This pullrequest changes platform/android/build_sdk.py

* look for NDK in "ndk-bundle" directory under the SDK directory if location not specified
* look for cmake and ninja in "cmake" directory under the SDK directory if not found in the path
* disable use of ccache if not found in the path

### DIFF
--- a/platforms/android/build_sdk.py
+++ b/platforms/android/build_sdk.py
@@ -48,6 +48,53 @@ def check_dir(d, create=False, clean=False):
             os.makedirs(d)
     return d
 
+def which(cmd, mode=os.F_OK | os.X_OK, path=None):
+
+    def _access_check(fn, mode):
+        return os.path.exists(fn) and os.access(fn, mode) and not os.path.isdir(fn)
+
+    # If we're given a path with a directory part, look it up directly
+    # rather than referring to PATH directories. This includes checking
+    # relative to the current directory, e.g. ./script
+    if os.path.dirname(cmd):
+        if _access_check(cmd, mode):
+            return cmd
+
+        return None
+
+    if path is None:
+        path = os.environ.get("PATH", os.defpath)
+    if not path:
+        return None
+
+    path = path.split(os.pathsep)
+
+    if sys.platform == "win32":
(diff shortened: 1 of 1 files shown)


platforms/android/build_sdk.py
```diff
@@ -48,6 +48,53 @@ def check_dir(d, create=False, clean=False):
...
+        if os.curdir not in path:
+            path.insert(0, os.curdir)
+
+            # need to consider PATHEXT on Windows.
+            pathext = os.environ.get("PATHEXT", "").split(os.pathsep)
+            if any(cmd.lower().endswith(ext.lower()) for ext in pathext):
+                files = [cmd]
+            else:
+                files = [cmd + ext for ext in pathext]
+    else:
```
alalek: Code is not complete.

I believe we can use something like this instead:
```
def check_executable(cmd)
    try:
        retcode = subprocess.call(cmd, shell=shell)
        if retcode < 0:
            return False
        return True
    except:
        return False

if not check_executable(['ninja', "--version"])
    ... fallback for manual searching ...
if not check_executable(['cmake', "--version"])
    ...
```

komakai:
> What about adding script parameter with force using of CMake/Ninja from Android SDK?

Might be a good idea - but probably something for a separate PR

platforms/android/build_sdk.py
komakai: Will add something like this and push a new commit

komakai: Pushed a new commit to address this.
I tested the changes on Windows and Mac and they seem to work well

PR #13943, opened 2019-03-01
G-API external backend development

rgarnov:
* Exported symbols required for external backend development
* Moved private part of gmodel.hpp to gmodel_priv.hpp, removed private Layout metadata from GModel::Graph (placed it in a special GModel::LayoutGraph). Added a public simplified overload of GModel::mkDataNode (the first one was placed into gmodel_priv.hpp)
* Moved HostCtor and ConstVal to gobjref.hpp
* Renamed gapi_priv.hpp/cpp to gorigin.hpp/cpp


modules/gapi/include/opencv2/gapi/backends/gbackend.hpp
```diff
@@ -19,7 +19,7 @@
 #include "opencv2/gapi/util/optional.hpp"
 #include "opencv2/gapi/own/scalar.hpp"
 
-#include "compiler/gmodel.hpp"
+#include "opencv2/gapi/compiler/gmodel.hpp"
```
dmatveev: should this file really be part of OpenCV public headers (included in the installation)?

I believe what we aligned on is that G-API backends are built with OpenCV as modules (with full access to G-API's internal headers)

modules/gapi/include/opencv2/gapi/compiler/gmodel.hpp
```diff
@@ -24,12 +24,8 @@
...
 #include "opencv2/gapi/garg.hpp"
 #include "opencv2/gapi/gkernel.hpp"
-#include "api/gapi_priv.hpp"   // GShape
-#include "api/gproto_priv.hpp" // origin_of
-#include "backends/common/gbackend.hpp"
-
-#include "compiler/gobjref.hpp"
-#include "compiler/gislandmodel.hpp"
+#include "opencv2/gapi/compiler/gobjref.hpp"
+#include "opencv2/gapi/compiler/gislandmodel.hpp"
```
dmatveev: Same applies to this

modules/gapi/include/opencv2/gapi/compiler/gmodel.hpp
```diff
@@ -198,50 +182,35 @@ namespace GModel
...
-    GMetaArgs collectOutputMeta(GModel::ConstGraph cg, ade::NodeHandle node);
+    GAPI_EXPORTS GMetaArgs collectInputMeta(GModel::ConstGraph cg, ade::NodeHandle node);
+    GAPI_EXPORTS GMetaArgs collectOutputMeta(GModel::ConstGraph cg, ade::NodeHandle node);
 
-    ade::EdgeHandle getInEdgeByPort(const GModel::ConstGraph& cg, const ade::NodeHandle& nh, std::size_t in_port);
+    GAPI_EXPORTS ade::EdgeHandle getInEdgeByPort(const GModel::ConstGraph& cg, const ade::NodeHandle& nh, std::size_t in_port);
 
     // Returns true if the given backend participates in the execution
-    bool isActive(const GModel::Graph &cg, const cv::gapi::GBackend &backend);
+    GAPI_EXPORTS bool isActive(const GModel::Graph &cg, const cv::gapi::GBackend &backend);
```
dmatveev: Once we refactor libopencv_gapi.so to be "framework-only" library, would that exports still be required? Is this reqiured for other possible backends linking with it?

dmatveev: Given this, I believe it is easier to make one exported class (`GModel`), rather than a ton of exported methods

modules/gapi/include/opencv2/gapi/backends/gbackend.hpp
rgarnov: Yes, you are correct. Reworked, now there are no changes on header export in this PR (only symbols and some cleanup and general header restructuring)

rgarnov:
@dbudniko

rgarnov:

PR #13957, opened 2019-03-03
python : Add rows cols and flags as umat member

LaurentBerger:
Add rows cols and flags as umat member

http://answers.opencv.org/question/187401/obtaining-opencv-umat-shape-in-python/

### DIFF
--- a/modules/core/misc/python/shadow_umat.hpp
+++ b/modules/core/misc/python/shadow_umat.hpp
@@ -54,6 +54,9 @@ public:
 
     // offset of the submatrix (or 0)
     CV_PROP_RW size_t offset;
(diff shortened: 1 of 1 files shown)


alalek:
1. Mat is not wrapped into Python. numpy arrays are used for handling ND matrix.
2. Python users use `.shape` to receive dimension information. But there is some issue with "channels" (channels are part of `.shape`)
3. UMat supports ND-arrays too (at least 3D), so cols/rows is not enough (for example, DNN blobs are 4D matrixes).

Perhaps more convenient way for Python users is adding numpy-compatible `.shape` for UMat too.

Another problem is Java bindings, but it is out of scope (not supported in Java).

PR #13985, opened 2019-03-06
Add ability to read thresh and nms_threshold from YOLO layer in YOLOV…

### LINKED ISSUE #13963: How to pass thresh and nms_threshold when loading YOLO v3?

- OpenCV => 3.4.4
- Operating System / Platform => Windows 7
- Compiler => n/a

##### Detailed description

There is no way to pass thresh and nms when working with YOLO v3, in the setYolo method here:
https://github.com/opencv/opencv/blob/master/modules/dnn/src/darknet/darknet_io.cpp#L374

So It always use 0.2 as default thresh:
`thresh = params.get<float>("thresh", 0.2);`
in https://github.com/opencv/opencv/blob/master/modules/dnn/src/layers/region_layer.cpp#L72

and 0.4 as default nms_threshold:
`nmsThreshold = params.get<float>("nms_threshold", 0.4);`
in https://github.com/opencv/opencv/blob/master/modules/dnn/src/layers/region_layer.cpp#L72

##### Steps to reproduce

`net = cv2.dnn.readNetFromDarknet(configPath, weightsPath)`

After read the net, it has 0.2 as default thresh and 0.4 as default nms_threshold, and we have no option to change that and so no way to evaluate a lower option.

Please kindly help.

liuyuns:
resolves #13963

### This pullrequest changes

Add the ability to read thresh and nms_threshold from YOLO layer of YOLOV3's cfg file.

Currently, the thresh is always 0.2, the nms_threshold is always 0.4, and there is no way to pass or change them.

With the change, we can add thresh and nms_threshold in cfg file to override the default values.

### DIFF
--- a/modules/dnn/src/darknet/darknet_io.cpp
+++ b/modules/dnn/src/darknet/darknet_io.cpp
@@ -371,7 +371,7 @@ namespace cv {
                     fused_layer_names.push_back(last_layer);
                 }
 
-                void setYolo(int classes, const std::vector<int>& mask, const std::vector<float>& anchors)
+                void setYolo(int classes, const std::vector<int>& mask, const std::vector<float>& anchors, float thresh, float nms_threshold)
                 {
                     cv::dnn::LayerParams region_param;
                     region_param.name = "Region-name";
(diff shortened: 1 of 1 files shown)


dkurt:
@liuyuns, please change a base branch from master to 3.4. Then it will be ported into the master branch as well.

liuyuns:
rebase and submit a new PR

PR #13988, opened 2019-03-06
videoio: gst: Fix gst assertion on null msg

raymanfx:
According to the gstreamer docs [1], the GstMessage pointer returned by
gst_bus_pop() is nullable, meaning NULL is a valid return value.

Previously, gst_is_missing_plugin_message would throw an assert when its
message object parameter would fail the GST_IS_MESSAGE macro check,
crashing the entire process (unless running in a try-catch block of course).

Instead of relying on valid messages, check if the message object itself is
valid before passing it to other gstreamer functions.

[1] https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/GstBus.html#gst-bus-pop

Signed-off-by: Christopher N. Hesse <raymanfx@gmail.com>

### DIFF
--- a/modules/videoio/src/cap_gstreamer.cpp
+++ b/modules/videoio/src/cap_gstreamer.cpp
@@ -1811,6 +1811,10 @@ void handleMessage(GstElement * pipeline)
 
     while(gst_bus_have_pending(bus)) {
         msg = gst_bus_pop(bus);
+        if (!msg || !GST_IS_MESSAGE(msg))
+        {
+            continue;
+        }
 
         //printf("\t\tGot %s message\n", GST_MESSAGE_TYPE_NAME(msg));
 


raymanfx:
FWIW I'm still not quite sure what exactly is causing this issue - or why no one else appears to be hitting it. I have an application that reads frames from a video file using VideoCapture. OpenCV is compiled with the Gstreamer backend enabled and I am also using gstreamer-vaapi plugins.

After some time (sometimes ~3h, sometimes ~10m) the application crashes because "gst_is_missing_plugin_message" asserts on the (invalid) message object pointer.
It's not really reproducible on my side though, I tried for more than a week now.

raymanfx:
FYI my application builds against OpenCV 3.4, not the master branch.
By looking at the code, this bug should at least affect those two branches in the same way though (3.4 and master).

alalek:
Thank you!
I believe this change doesn't not break anything.

So, please:

PR #13989, opened 2019-03-06
Make use of explicit output shape for the de-convolution layer.

kohei-us:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/582

This is relevant to cases where the net is loaded from an ONNX model
containing ConvTranspose layers, which may include the output_shape
attribute.

c.f. https://github.com/onnx/onnx/blob/master/docs/Operators.md#ConvTranspose

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/dnn/src/onnx/onnx_importer.cpp
+++ b/modules/dnn/src/onnx/onnx_importer.cpp
@@ -591,6 +591,37 @@ void ONNXImporter::populateNet(Net dstNet)
             }
             layerParams.set("num_output", layerParams.blobs[0].size[1] * layerParams.get<int>("group", 1));
             layerParams.set("bias_term", node_proto.input_size() == 3);
+
+            if (layerParams.has("output_shape"))
+            {
+                const DictValue& outShape = layerParams.get("output_shape");
+
+                if (outShape.size() != 4)
+                    CV_Error(Error::StsNotImplemented, "Output shape must have 4 elements.");
+
+                const int strideY = layerParams.get<int>("stride_h", 1);
+                const int strideX = layerParams.get<int>("stride_w", 1);
+                const int outH = outShape.getIntValue(2);
+                const int outW = outShape.getIntValue(3);
+
+                if (layerParams.get<String>("pad_mode") == "SAME")
+                {
+                    layerParams.set("adj_w", (outW - 1) % strideX);
+                    layerParams.set("adj_h", (outH - 1) % strideY);
+                }
+                else if (layerParams.get<String>("pad_mode") == "VALID")
+                {
+                    if (!layerParams.has("kernel_h") || !layerParams.has("kernel_w"))
+                        CV_Error(Error::StsNotImplemented,
+                                 "Required attributes 'kernel_h' and 'kernel_w' are not present.");
+
+                    int kernelH = layerParams.get<int>("kernel_h");
+                    int kernelW = layerParams.get<int>("kernel_w");
+
+                    layerParams.set("adj_w", (outW - kernelW) % strideX);
+                    layerParams.set("adj_h", (outH - kernelH) % strideY);
+                }
+            }
         }
         else if (layer_type == "Transpose")
         {
--- a/modules/dnn/test/test_onnx_importer.cpp
+++ b/modules/dnn/test/test_onnx_importer.cpp
@@ -73,6 +73,7 @@ TEST_P(Test_ONNX_layers, Deconvolution)
     testONNXModels("deconvolution");
     testONNXModels("two_deconvolution");
     testONNXModels("deconvolution_group");
+    testONNXModels("deconvolution_output_shape");
 }
 
 TEST_P(Test_ONNX_layers, Dropout)


dkurt:
@kohei-us, Can you add a test?

kohei-us:
@dkurt Sure, I can look into that.  Any suggestion as to where to add a new test?

kohei-us:
I guess modules/dnn/test/test_onnx_importer.cpp may be a good place for this (?)

dkurt:
@kohei-us, Yeah.

kohei-us:
I have another pull request on opencv_extra to add test case files.

https://github.com/opencv/opencv_extra/pull/582

kohei-us:
I also pushed a one-line change to run the test case.  Obviously it will depend on the change I made to the above referenced pull request...

dkurt:
@kohei-us, May I ask you about the state of this PR? Have your tried to simplify it considering my comment from https://github.com/opencv/opencv_extra/pull/582#issuecomment-471528501?

kohei-us:
@dkurt Sorry I got caught up with something else. I will take a look at this and try your suggestion early next week.

kohei-us:
@dkurt Ok.  I've made changes per your suggestion, and tested it locally to make sure it still addresses my issue.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -591,6 +591,21 @@ void ONNXImporter::populateNet(Net dstNet)
             }
             layerParams.set("num_output", layerParams.blobs[0].size[1] * layerParams.get<int>("group", 1));
             layerParams.set("bias_term", node_proto.input_size() == 3);
+
+            const DictValue* outShape = layerParams.ptr("output_shape");
+            if (outShape && outShape->size() == 4)
```
dkurt: * Please check if there is `"output_shape"` parameter by `layerParams.has("output_shape");`
* It's more convenient to use `DictValue outShape = layerParams.get("output_shape");`
* If `outShape.size() != 4` it's better to throw a `StsNotImplemented` exception because we don't know how to manage it.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -591,6 +591,21 @@ void ONNXImporter::populateNet(Net dstNet)
...
+            {
+                const int strideY = layerParams.get<int>("stride_h");
+                const int strideX = layerParams.get<int>("stride_w");
```
dkurt: Please use default values because `strides` might be optional:

```cpp
const int strideY = layerParams.get<int>("stride_h", 1);
const int strideX = layerParams.get<int>("stride_w", 1);
```

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -591,6 +591,21 @@ void ONNXImporter::populateNet(Net dstNet)
...
+                const int outH = outShape->getIntValue(2);
+                const int outW = outShape->getIntValue(3);
+
+                if (layerParams.get<String>("pad_mode") == "SAME")
+                {
+                    layerParams.set("adj_w", (outW - 1) % strideX);
+                    layerParams.set("adj_h", (outH - 1) % strideY);
+                }
```
dkurt: What about "VALID" mode? Please see https://github.com/opencv/opencv/blob/master/modules/dnn/src/tensorflow/tf_importer.cpp for a reference code.

modules/dnn/src/onnx/onnx_importer.cpp
kohei-us: Done.

modules/dnn/src/onnx/onnx_importer.cpp
kohei-us: This is done as well.

modules/dnn/src/onnx/onnx_importer.cpp
kohei-us: This is done too.

kohei-us:
@dkurt Let me know if my recent change addresses all of your comments.  Thanks!

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -591,6 +591,42 @@ void ONNXImporter::populateNet(Net dstNet)
...
+                else if (layerParams.get<String>("pad_mode") == "VALID")
+                {
+                    if (!layerParams.has("kernel_shape"))
+                        CV_Error(Error::StsNotImplemented,
+                                 "Kernel shape attribute is required for VALID pad mode.");
+
+                    const DictValue& kernelShape = layerParams.get("kernel_shape");
```
dkurt: It's better to use `kernel_h` and `kernel_w` (please also check that they are presented):
```
                    if (!layerParams.has("kernel_h") || !layerParams.has("kernel_w"))
                        CV_Error(Error::StsNotImplemented, "...");
```

kohei-us: Done and pushed.

dkurt:
@kohei-us, The last thing I wanted to ask you is to change the base branch from `master` to `3.4`. You can do it without closing this PR. See https://github.com/opencv/opencv/wiki/How_to_contribute#q2-i-was-asked-to-change-the-target-branch-for-my-pull-request-how-can-i-do-that for details.

kohei-us:
Done. Hopefully I've done this correctly.  I've done the same with opencv/opencv_extra#582 as well.

PR #14011, opened 2019-03-09
Add Android SDK debug build flag

### LINKED ISSUE #14010: Add Debug flag to Android SDK build
- OpenCV => Latest
- Operating System / Platform => Android

##### Detailed description
In order to be able to debug while working on the Android platform there should be a build flag "--debug" that builds the OpenCV binaries with debug information. (Additionally this needs to suppress any build stages that strip out debug info).

I have a PR prepared for this.

komakai:
### This pullrequest adds a flag "-debug" to platform/android/build_sdk.py which results in debuggable output binaries.

* add build flag "--debug"
* when flag is set then
  * set CMake build type to Debug
  * set CMake build flag BUILD_WITH_DEBUG_INFO to ON
  * invokes ninja with target "install" instead of "install/strip"
* suppresses stripping of debug information in the SDK gradle build
* fixes a bug in the DNN Java wrapper (missing LOG_TAG define)

### DIFF
--- a/modules/dnn/misc/java/src/cpp/dnn_converters.cpp
+++ b/modules/dnn/misc/java/src/cpp/dnn_converters.cpp
@@ -6,7 +6,7 @@
 
 #include "dnn_converters.hpp"
 
-#define LOG_TAG "DNN"
+#define LOG_TAG "org.opencv.dnn"
 
 void Mat_to_MatShape(cv::Mat& mat, MatShape& matshape)
 {
--- a/modules/dnn/misc/java/src/cpp/dnn_converters.cpp
+++ b/modules/dnn/misc/java/src/cpp/dnn_converters.cpp
@@ -6,6 +6,7 @@
 
 #include "dnn_converters.hpp"
 
+#define LOG_TAG "DNN"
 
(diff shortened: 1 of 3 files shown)


komakai:
Screenshot debugging OpenCV in Android Studio:
スクリーンショット 2019-03-09 14 58 34

modules/dnn/misc/java/src/cpp/dnn_converters.cpp
```diff
@@ -6,6 +6,7 @@
 
 #include "dnn_converters.hpp"
 
+#define LOG_TAG "DNN"
```
alalek: Used?
"org.opencv.dnn"

komakai: Used in `Mat_to_MatShape` which is not used anywhere

alalek: OK, please rename "org.opencv.dnn" (to keep tags consistent with other code)

bommo1:
Could you please merge this into the 3.4 branch? Thanks!

komakai:
@bommo1 the 3.4 Android build is totally different - there is no way to merge this.

Issue #14008, opened 2019-03-08
cv::merge() make some strange mistake when merge 3 or 4 pics.


2rintf:
### System information(version)
- OpenCV => 4.0.1-dev  
- Operating System / Platform => win10  
- Complier => VS2017  

### Detailed description  

Recently when i used `cv::merge()`, i found something strange. I try to merge **3 same pics**( 3-channels，BGR) into 1 pics(9-channels). And `cv::merge()` really give me a 9-channels pic. However, the returned pic is not complete, so it can't recovery correctly by using `cv::split()`. And the **same error** occured when i try to merge **4 pics**.   
In fact, the error **do not** occur **except the number of pic is 3 or 4**.(Not exact actually, I only have a try from 1 to 50 pics).  
Using the `ImageWatch` on VS2017, I found that the wrong merged pic have some rules.  
1. Only `1/3` cols of the wrong pic have value, the others are always 0.  
![20190308221110059_9916](https://user-images.githubusercontent.com/20776914/54035229-d224db80-41f3-11e9-87f0-3dad2b79e59b.png)
![20190308221137227_14866](https://user-images.githubusercontent.com/20776914/54035254-e36de800-41f3-11e9-9747-9883f1667314.png)
2. It seems like that the value of pixel are arranged in order of `BBB...GGG...RRR...` rather than `BGRBGRBGR...`, which caused the gray pic.(I also write some codes to confirm it. I successfully recovery the color although the pic is still wrong.)  
![20190308221417118_13171](https://user-images.githubusercontent.com/20776914/54035285-f7b1e500-41f3-11e9-8a45-a41aef07fee8.png)  

Is it a BUG? Or do I make some mistake?  

### step to reproduce  
PS: code for test  
```C
#include <iostream>
#include <string>
#include <opencv2/opencv.hpp>

using namespace std;
using namespace cv;

int main()
{
	cout << CV_VERSION << endl;
	//namedWindow("result", WINDOW_FREERATIO);
	//namedWindow("test", WINDOW_FREERATIO);

	Mat srcMat = imread("C:\\Users\\chen\\Desktop\\1.jpg");
	Mat mergeMat;
	vector<Mat> vM;
	vector<Mat> vM_out;
	vector<Mat> showPic;

	// merge 3 pics or 4 pics
	for (int i = 0; i < 3; i++)
	{
		vM.push_back(srcMat);
	}
	cv::merge(vM, mergeMat);// mergeMat here is actually wrong

[...]

PR #14032, opened 2019-03-11
fix ipp_merge

take1014:
resolove #14008


alalek:
Please start from test which fails on case from the original issue.

Currently if `allch1 == true`, then `n == cn`. So nothing has been changed (nothing fixed).

take1014:
@alalek Thanks for the follow.  Sorry, the revision I used to solve this problem seems to be old. (HEAD:d74b98c3d9264f9623b27a8d4e0fc92ac97b71d0)
This issue did not reproduce with the recent revision.(HEAD:55f02d8c66baf0747805f75762bdaca2e2d1985e)

image

PR #14035, opened 2019-03-12
add Y16/GRAY16_LE support to gstreamer

jveitchmichaelis:
### This pullrequest changes

* Adds Y16 (GRAY16_LE) support to Gstreamer.
* Adds support for `CAP_PROP_CONVERT_RGB`
* Adds a depth parameter to the gstreamer capture class
* Images are returned as `CV_8UC3` by default, unless `CAP_PROP_CONVERT_RGB` is set to false.

Tested with a FLIR Boson

### DIFF
--- a/modules/videoio/src/cap_gstreamer.cpp
+++ b/modules/videoio/src/cap_gstreamer.cpp
@@ -180,6 +180,8 @@ private:
     bool          isPosFramesEmulated;
     gint64        emulatedFrameNumber;
     bool          isOutputByteBuffer;
+    bool          shouldConvertRGB;
+    int           depth;
 
 public:
     GStreamerCapture();
@@ -218,7 +220,9 @@ GStreamerCapture::GStreamerCapture() :
     isPosFramesSupported(false),
     isPosFramesEmulated(false),
     emulatedFrameNumber(-1),
-    isOutputByteBuffer(false)
+    isOutputByteBuffer(false),
+    shouldConvertRGB(true),
+    depth(8)
 {
 }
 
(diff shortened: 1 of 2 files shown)


jveitchmichaelis:
Fixed trailing whitespace

mshabunin:
@jveitchmichaelis , could you please add test case here: https://github.com/opencv/opencv/blob/master/modules/videoio/test/test_gstreamer.cpp

Is this patch applicable to 3.4 branch too? If yes, please rebase your branch there and change pull request target to 3.4 branch. We will merge it to _master_ later (see https://github.com/opencv/opencv/wiki/Branches).

jveitchmichaelis:
No problem. I've rebased to 3.4, I'll add some tests too.

mshabunin:
> I added support for writing 16-bit frames with Gstreamer, but perhaps it makes sense to split that into another commit and leave this one only for capture.

I agree.

>At a minimum I can add a testcase like:
> ...
>Would that be sufficient?

Yes, at minimum. It would be better to test handling of `CV_CAP_PROP_CONVERT_RGB` property too and verify actual 16-bit image capturing. But we can add it in later PRs.

jveitchmichaelis:
I've added two test cases:

* An additional case for the existing suite that opens a GRAY16 pipeline and checks that there's something i it
* A test case to check the default behaviour for CONVERT_RGB

One issue that I've noticed is that CAP_GSTREAMER doesn't follow the convention(?) that capture pipelines should return 3-channel 8-bit images by default. Most of the outputs here (e.g. GRAY8, and all the YUV variants) return a single channel image and rely on the user to convert to colour.

joshdoe:
> I added support for writing 16-bit frames with Gstreamer, but perhaps it makes sense to split that into another commit and leave this one only for capture.

@jveitchmichaelis, we'd be happy to test a patch for writing, please create a PR when you get a chance. We looked at doing it, but it's not obvious how we could automatically determine a pipeline is Y16, seems like we'd need a property.

jveitchmichaelis:
@joshdoe sure, will do! It will probably be based on the output codec. Relatively few formats actually support 16 bit output per channel, so you could infer on that plus the pixel format I think.

joshdoe:
> @joshdoe sure, will do! It will probably be doable based on the output codec. Relatively few formats actually support 16 bit output per channel, so you could infer on that plus the pixel format I think.

Great, thanks. I recently added Y16 support to the GStreamer DirectShow source because I was testing a FLIR Boson thermal camera, but in this case we're using a Camera Link source (https://github.com/joshdoe/gst-plugins-vision).

joshdoe:
I've modified this to apply against master, as I couldn't easily get the 3.4 branch built. I removed the CONVERT_RGB bits as suggested by @mshabunin , and changed the tests a bit to check depth on all formats, and convert 16->8 for GRAY16_LE. My branch is [joshdoe/opencv/y16_gstreamer_master](https://github.com/joshdoe/opencv/tree/y16_gstreamer_master). I also rebased this to 3.4, but haven't been able to test it: [joshdoe/opencv/y16_gstreamer_3.4](https://github.com/joshdoe/opencv/tree/y16_gstreamer_3.4).

asmorkalov:
@jveitchmichaelis The PR diverged with 3.4. Could you rebase and fix conflicts. 
@mshabunin What is the PR status? Are there open items? Could we merge it after rebase and conflicts fix?

mshabunin:
@asmorkalov , I think #18694 is more compact and implements the same feature, so we can close this one. https://github.com/opencv/opencv/pull/18535 implements Y16 support in VideoWriter.

joshdoe:
I'll second the implementation in #18694 as it's cleaner, though I tested writing (from #18535) and reading for both of these implementations and they passed.

jveitchmichaelis:
Apologies, this feel through the cracks - feel free to go with whichever implementation you feel fits best.

Issue #13697, opened 2019-01-26
Segmentation fault when camera is released after disabling RGB conversion


jveitchmichaelis:
##### System information (version)
- OpenCV => Tested on 4.0.0 and 4.0.1
- Operating System / Platform => Linux 4.14.79-v7+ armv7l (Pi 3B+)
- Compiler => gcc

##### Detailed description

Capturing images from a camera on a Pi 3 B+. OpenCV is being run from within a Docker container, so the environment is reasonably clean. I have also replicated this when OpenCV is run on the Pi normally.

If I open the camera and capture an image, no problems. If I attempt to modify the camera settings, I can still capture images, but when the program quits (or in this case I manually release the camera), it segfaults.

gdb gets as far as releasing the V4L buffer and then dies:

``` bash
(gdb) bt full
#0  __GI___libc_free (mem=0x6e80c008) at malloc.c:2966
        ar_ptr = <optimized out>
        p = <optimized out>
        hook = 0x0
#1  0x75d58e04 in cv::CvCaptureCAM_V4L::releaseBuffers (this=0x5d3e8) at /code/opencv/modules/videoio/src/cap_v4l.cpp:1870
No locals.
#2  0x75d53050 in cv::CvCaptureCAM_V4L::~CvCaptureCAM_V4L (this=0x5d3e8, __in_chrg=<optimized out>)
    at /code/opencv/modules/videoio/src/cap_v4l.cpp:371
No locals.
#3  0x75d530b4 in cv::CvCaptureCAM_V4L::~CvCaptureCAM_V4L (this=0x5d3e8, __in_chrg=<optimized out>)
    at /code/opencv/modules/videoio/src/cap_v4l.cpp:374
No locals.
#4  0x75d2a47c in cvReleaseCapture (pcapture=0x7efffb88) at /code/opencv/modules/videoio/src/videoio_c.cpp:115
No locals.
#5  0x75d2b23c in cv::DefaultDeleter<CvCapture>::operator() (this=0x60074, obj=0x5d3e8)
    at /code/opencv/modules/videoio/src/cap.cpp:49
No locals.
#6  0x75d2e9d8 in std::_Sp_counted_deleter<CvCapture*, cv::DefaultDeleter<CvCapture>, std::allocator<void>, (__gnu_cxx::_Lock_policy)1>::_M_dispose (this=0x60050) at /usr/include/c++/6/bits/shared_ptr_base.h:464
No locals.
[...]

PR #14036, opened 2019-03-12
fix segfault when CONVERT_RGB is disabled with V4L backend

jveitchmichaelis:
resolves https://github.com/opencv/opencv/issues/13697

### This pullrequest changes

Fixes a bug where CAP_V4L would segfault after releasing a capture device. The error seemed to be caused by a frame not being released properly when CONVERT_RGB is disabled. Possibly it was marked as allocated when it wasn't, and then the segfault occurs when the destructor gets called.

I've tested switching between modes (before releasing the camera) and it works without any errors.

### DIFF
--- a/modules/videoio/src/cap_v4l.cpp
+++ b/modules/videoio/src/cap_v4l.cpp
@@ -1803,9 +1803,11 @@ bool CvCaptureCAM_V4L::setProperty( int property_id, double _value )
         if (bool(value)) {
             convert_rgb = convertableToRgb();
             return convert_rgb;
+        }else{
+            convert_rgb = false;
+            releaseFrame();
+            return true;
         }
-        convert_rgb = false;
-        return true;
     case cv::CAP_PROP_FOURCC:
     {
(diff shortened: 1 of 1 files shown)


modules/videoio/src/cap_v4l.cpp
```diff
@@ -1803,6 +1803,9 @@ bool CvCaptureCAM_V4L::setProperty( int property_id, double _value )
         if (bool(value)) {
             convert_rgb = convertableToRgb();
             return convert_rgb;
+        }else{
+            convert_rgb = false;
+            releaseFrame();
         }
         convert_rgb = false;
```
alalek: Lets keep code cleaner:
- remove duplicated `convert_rgb = false;`
- remove "else" body or move return into this "else" body.

jveitchmichaelis: Done and done.

jveitchmichaelis:
Ok, no problem, I've rebased to 3.4.

Issue #13952, opened 2019-03-02
OpenCV 4.0.1 + Cuda 10.1, failed to build?


jiapei100:
It looks it's still a kind of buggy? Did any of you have successfully built OpenCV 4.0.1 with Cuda 10.1?

```
....../opencv/modules/core/include/opencv2/core/cvstd_wrapper.hpp(45): error: incomplete type is not allowed
          detected during:
            instantiation of "cv::sfinae::has_parenthesis_operator<C, Ret, Args...>::check [with C=cv::DefaultDeleter<<unnamed>::LookUpTableImpl>, Ret=void, Args=<<unnamed>::LookUpTableImpl *>]" based on template argument <cv::DefaultDeleter<<unnamed>::LookUpTableImpl>> 
(49): here
            instantiation of class "cv::sfinae::has_parenthesis_operator<C, Ret, Args...> [with C=cv::DefaultDeleter<<unnamed>::LookUpTableImpl>, Ret=void, Args=<<unnamed>::LookUpTableImpl *>]" 
(138): here
            instantiation of "cv::Ptr<_Tp> cv::makePtr<_Tp,A1...>(const A1 &...) [with _Tp=<unnamed>::LookUpTableImpl, A1=<cv::_InputArray>]" 
....../opencv_contrib/modules/cudaarithm/src/cuda/lut.cu(207): here

....../opencv/modules/core/include/opencv2/core/cvstd_wrapper.hpp(49): error: template instantiation resulted in unexpected function type of "std::true_type (std::integral_constant<__nv_bool, false> *)" (the meaning of a name may have changed since the template declaration -- the type of the template is "std::true_type (std::is_same<std::decay<decltype((<expression>))>::type, void>::type *)")
          detected during:
            instantiation of "cv::sfinae::has_parenthesis_operator<C, Ret, Args...>::check [with C=cv::DefaultDeleter<<unnamed>::LookUpTableImpl>, Ret=void, Args=<<unnamed>::LookUpTableImpl *>]" based on template argument <cv::DefaultDeleter<<unnamed>::LookUpTableImpl>> 
(49): here
            instantiation of class "cv::sfinae::has_parenthesis_operator<C, Ret, Args...> [with C=cv::DefaultDeleter<<unnamed>::LookUpTableImpl>, Ret=void, Args=<<unnamed>::LookUpTableImpl *>]" 
(138): here
[...]

PR #14041, opened 2019-03-13
Extract Ptr-related code from lut.cu to new lut.cpp

nglee:
https://github.com/opencv/opencv/issues/13952#issuecomment-470260632

```
force_builders=Custom
buildworker:Custom=linux-1
```

### DIFF
--- a/modules/cudaarithm/src/cuda/lut.cu
+++ b/modules/cudaarithm/src/cuda/lut.cu
@@ -48,6 +48,8 @@
 
 #else
 
+#include "../lut.hpp"
+
 #include "opencv2/cudaarithm.hpp"
 #include "opencv2/cudev.hpp"
 #include "opencv2/core/private.cuda.hpp"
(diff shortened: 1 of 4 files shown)


modules/cudaarithm/src/cuda/lut.cu
```diff
@@ -48,31 +48,16 @@
 
 #else
 
-#include "opencv2/cudaarithm.hpp"
+#include "precomp.hpp"
```
alalek: > #include "precomp.hpp"

Please make this as a first "unconditional" include.

modules/cudaarithm/src/precomp.hpp
```diff
@@ -60,4 +60,24 @@
 #  include <cufft.h>
 #endif
 
+#include <cuda_runtime.h>
+
+namespace cv { namespace cuda {
+
+    class LookUpTableImpl : public LookUpTable
```
alalek: Please avoid indentation in namespaces

modules/cudaarithm/src/precomp.hpp
```diff
@@ -60,4 +60,24 @@
 #  include <cufft.h>
 #endif
 
+#include <cuda_runtime.h>
+
+namespace cv { namespace cuda {
```
alalek: It is better to place this code into lut.hpp (private header in "cudaarithm/src" folder).
To avoid storing of implementation details in `precomp.hpp` - lets keep here only common headers.

PR #14050, opened 2019-03-13
Changed eps checking condition

### LINKED ISSUE #13946: ANN_MLP TermCriteria::COUNT will still terminate after default epsilon
- OpenCV => 3.4.2 and 4.0.1 
- Operating System / Platform => Mac 32 Bit and 64 Bit
- Compiler => Eclipse

##### Detailed description

When training an ANN_MLP, if you specify to terminate training after (cv::TermCriteria::COUNT) it will still terminate with (cv::TermCriteria::COUNT + cv::TermCriteria::EPS) using the default epsilon value.

##### Steps to reproduce

int nEpochs;
double epsilon;
cv::Ptr<<cv::ml::ann_mlp>> network = cv::ml::ANN_MLP::create(); 
network->setTermCriteria(cv::TermCriteria(cv::TermCriteria::COUNT, nEpochs, epsilon)); 
network->train(…);

This should terminate after nEpochs, however if the error drops below the default epsilon value, training will also terminate.

The problem is first found here in source code, when defining epsilon:
https://github.com/opencv/opencv/blob/master/modules/ml/src/ann_mlp.cpp#L867

Then later here, when terminating the training loop:
https://github.com/opencv/opencv/blob/master/modules/ml/src/ann_mlp.cpp#L949

I found this error when doing some benchmarks. I noticed after a certain amount of epochs, the training time would stagnate (it should increase when you increase the number of epochs).

There is a work around, if you set term criteria to COUNT + EPS and define epsilon = 0:
network->setTermCriteria(cv::TermCriteria(cv::TermCriteria::COUNT + cv::TermCriteria::EPS, nEpochs, 0));

However, according to the documentation, just using cv::TermCriteria::COUNT should suffice.

npochhi:
resolves #13946 

Changed the epsilon assigning condition. Initially, epsilon was taking FLOAT_MIN value in case of only COUNT condition, however according to the relevant issue, epsilon checking shouldn't happen.

### DIFF
--- a/modules/ml/src/ann_mlp.cpp
+++ b/modules/ml/src/ann_mlp.cpp
@@ -971,7 +971,7 @@ public:
         int count = inputs.rows;
 
         int iter = -1, max_iter = termCrit.maxCount*count;
-        double epsilon = termCrit.epsilon*count;
+        double epsilon = (termCrit.type & CV_TERMCRIT_EPS) ? termCrit.epsilon*count : 0;
 
         int l_count = layer_count();
         int ivcount = layer_sizes[0];


modules/ml/src/ann_mlp.cpp
```diff
@@ -946,7 +946,7 @@ class ANN_MLPImpl CV_FINAL : public ANN_MLP
             if( idx == 0 )
             {
                 //printf("%d. E = %g\n", iter/count, E);
-                if( fabs(prev_E - E) < epsilon )
+                if( fabs(prev_E - E) < epsilon && termCrit.type & CV_TERMCRIT_EPS )
```
alalek: I believe it is better set epsilon to zero during initialization on line 902:
```
double epsilon = (termCrit.type & TermCriteria::EPS)
        ? termCrit.epsilon*count
        : 0;  // TermCriteria::COUNT
```

Also it is better to avoid deprecated C-style defines.

npochhi: I've changed the code accordingly, please check.

PR #14056, opened 2019-03-14
Adding laplace.py

adityak2920:
In some of the files in samples/python print(__doc__) statement is missing. This pull request is for printing document strings so that the user know the usage and the function of the script.

Added laplace.py python script for laplacian of image


alalek:
No need to re-open PRs. Push changes inplace.

PR #14067, opened 2019-03-15
DNN: Optimization fuse: Conv -> Concat -> BatchNorm

allnes:
**WIP**


dkurt:
@allnes, Looks like this PR a and https://github.com/opencv/opencv/pull/14325 looks very similar. Moreover this one looks broken because there is no Concat fusion. May I ask you to try to combine them  into a single one (closing one of them) with both Eltwise and Concat fusions?

PR #14078, opened 2019-03-16
python: discover tests from module/misc/python/test paths

alalek:
support python tests from opencv_contrib

Names must be unique (`test_<name>.py`).

TODOs:
- [ ] Check ARMv7 builder, add installation rules

<cut/>

```
docker_image:Custom=ubuntu-clang:18.04
```


cmake/OpenCVUtils.cmake
```diff
@@ -1768,3 +1768,19 @@ macro(ocv_git_describe var_name path)
...
 endmacro()
+
+function(ocv_update_file filepath content)
+  if(EXISTS "${filepath}")
+    file(READ "${filepath}" __content)
+  else()
+    set(__content "")
+  endif()
+  if("${__content}" STREQUAL "${content}")
+    if(";${ARGN};" MATCHES ";VERBOSE;")
```
mshabunin: It would be great to document this function:
```.cmake
# ocv_update_file(filepath content [VERBOSE])
#   write content to file 
#   will not change modification time in case when file already exists and content has not changed
```

modules/python/test/test.py
```diff
@@ -20,7 +25,35 @@
 basedir = os.path.abspath(os.path.dirname(__file__))
 
 def load_tests(loader, tests, pattern):
-    tests.addTests(loader.discover(basedir, pattern=os.environ.get('OPENCV_PYTEST_FILTER', 'test_') + '*.py'))
+    cwd = os.getcwd()
+    config_file = 'opencv_python_tests.cfg'
```
mshabunin: In some cases it can be convenient to run tests from directory other than `<build>`.

We could add new  parameter to test app...
```.sh
./opencv/modules/python/test/test.py --repo opencv --test_root opencv --test_root opencv_contrib
```
And discover tests using recursive directory walk based on provided paths.

If we had ability to detect enabled modules list in runtime built into binaries (python at least), we could skip or filter disabled tests dynamically, without extra config files.

alalek: Main purpose of this patch is enabling testing of opencv_contrib modules.

PR #14082, opened 2019-03-17
Added laplace.py in samples/python

adityak2920:
### This pullrequest changes

Added laplace.py in samples/python. This is the python script for laplace.cpp in samples/cpp.

### DIFF
--- a/samples/python/laplace.py
+++ b/samples/python/laplace.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+
+'''
+    This program demonstrates Laplace point/edge detection using
+    OpenCV function Laplacian()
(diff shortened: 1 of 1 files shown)


alalek:
> demonstrates Laplace point/edge detection

There is already `edge.py` demo, based on Canny algorithm.
Lets avoid suggesting users using of laplace filter for edge detection tasks.

adityak2920:
@alalek Ok, so should I close this PR.

adityak2920:
@alalek Actually I was working on an Idea of GSoC which is "Improve and expand python version of OpenCV", in this idea on of the expected outcomes is to add python code for samples written in C++ only.
So, should I close this PR?

PR #14096, opened 2019-03-18
Remove fullscreen android:theme from samples where it breaks the menu

oseiskar:
Many of the Android samples rely on an _options menu_ (see image below) to work properly. On newer devices & Android versions (or at least on my phone), the menu was permanently hidden by the fullscreen theme and there seemed to be no way for the user to make it visible. Consequently, most of the examples were dysfunctional.

<cut/>

### This pullrequest changes

This PR simply removes the Android theme `Theme.NoTitleBar.Fullscreen` in the samples that use the options menu, which seems to fix the issue.

image
_Figure: Options menu visible on top of the image (Android 8.0.0 on Google Pixel 2)_

I tested this by building OpenCV from source (on Debian Stretch) and installing the samples using `adb` as
```
export ANDROID_HOME=~/Android/Sdk
export ANDROID_NDK=~/Android/Sdk/ndk-bundle
export ANDROID_SDK=~/Android/Sdk

cd opencv/platforms/android
./build_sdk.py my-build

adb install my-build/o4a/opencv_android/tutorial-2-mixedprocessing/build/outputs/apk/debug/tutorial-2-mixedprocessing-arm64-v8a-debug.apk
```

### DIFF
--- a/samples/android/15-puzzle/AndroidManifest.xml
+++ b/samples/android/15-puzzle/AndroidManifest.xml
@@ -8,8 +8,7 @@
 
     <application
         android:icon="@drawable/icon"
-        android:label="@string/app_name"
-        android:theme="@android:style/Theme.NoTitleBar.Fullscreen" >
+        android:label="@string/app_name" >
 
         <activity
             android:name=".Puzzle15Activity"
--- a/samples/android/15-puzzle/gradle/AndroidManifest.xml
+++ b/samples/android/15-puzzle/gradle/AndroidManifest.xml
@@ -5,8 +5,7 @@
 
     <application
         android:icon="@drawable/icon"
-        android:label="@string/app_name"
-        android:theme="@android:style/Theme.NoTitleBar.Fullscreen" >
+        android:label="@string/app_name">
 
         <activity
             android:name=".Puzzle15Activity"
--- a/samples/android/camera-calibration/AndroidManifest.xml
+++ b/samples/android/camera-calibration/AndroidManifest.xml
@@ -6,8 +6,7 @@
 
     <application
         android:label="@string/app_name"
-        android:icon="@drawable/icon"
-        android:theme="@android:style/Theme.NoTitleBar.Fullscreen">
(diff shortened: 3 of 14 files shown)


alalek:
Does this change triggers additional frame resize? (what canvas resolution is used?)

oseiskar:
Hmm. I'm not completely sure if this is what you mean, but this does seem to change the camera preview resolution

### Before

Relevant part of the Logcat output (I added a log line to `onCameraViewStarted` and it appears once in the log)

```
...
... D/JavaCameraView: getSupportedPreviewSizes()
... D/JavaCameraView: Set preview size to 1440x1080
... D/JavaCameraView: startPreview
... D/JavaCameraView: Starting processing thread
... I/OCVSample::Activity: onCameraViewStarted(1440, 1080)
... D/JavaCameraView: Preview Frame received. Frame size: 2332800
... D/CameraBridge: mStretch value: 1.0
... (two previous lines repeat)
```


### After this PR

The different layout / theme causes a different preview image size (now with `mStretch` > 1). The `onCameraViewStarted` method is still called only once. Notice also the `called onCreateOptionsMenu` line which was missing from the previous log.

```
...
... D/JavaCameraView: getSupportedPreviewSizes()
... D/JavaCameraView: Set preview size to 1280x768
... D/JavaCameraView: startPreview
... D/JavaCameraView: Starting processing thread
... I/OCVSample::Activity: onCameraViewStarted(1280, 768)
... I/OCVSample::Activity: called onCreateOptionsMenu
... D/JavaCameraView: Preview Frame received. Frame size: 1474560
... D/CameraBridge: mStretch value: 1.1601562
... (two previous lines repeat)
```

alalek:
Yes, this is exactly what I'm asking for. Thank you!

1.
Changed camera frames resolution.

2.
> mStretch value: 1.0
vs
> mStretch value: 1.1601562

Canvas resize has been used.

---

Anyway, we need this fix (for enabling of full functionality of these Android samples).
Lets keep this comment with before/after changes here for reference.

PR #14101, opened 2019-03-19
core: fix fallback CV_CONSTEXPR definition

alalek:
keep it empty to avoid failures of these declarations:
```
static CV_CONSTEXPR const int var = ...;
```

### DIFF
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -691,7 +691,7 @@ __CV_ENUM_FLAGS_BITWISE_XOR_EQ   (EnumType, EnumType)                           
 #  endif
 #endif
 #ifndef CV_CONSTEXPR
-#  define CV_CONSTEXPR const
+#  define CV_CONSTEXPR
 #endif
 
 // Integer types portatibility


alalek:
/cc @cv3d @JoeHowse @Bleach665 @tomoaki0705 
Please take a look (if possible, check with this contrib patch: https://github.com/opencv/opencv_contrib/pull/2046).

JoeHowse:
I believe your change is necessary.

Before your change, we have a `CV_CONSTEXPR` macro that is defined as `constexpr` for most compilers but `const` for MSVS2013.  This leads to some contradictions.  For example, the preprocessor converts `CV_CONSTEXPR char *p = ...` to either `constexpr char *p = ...` (for most compilers) or `const char *p = ...` (for MSVS2013), but these do not have similar meanings. (Rather, `constexpr char *p = ...` and `char *p const = ...` would have similar meanings.)

Your change implies that authors should always specify both `CV_CONSTEXPR` and `const` together in order to make constness as consistent as possible for different compilers.  The people who are working on MSVC2013 compatibility should review existing code to ensure that this practice is actually being followed.  (I believe that currently, some existing code specifies `CV_CONSTEXPR` without `const`.)

cv3d:
I agree with @JoeHowse. Your change makes sense as both `const` and `constexpr` are not alternatives to each other.
Furthermore, they need to or can be utilized together in several cases.

```C++
constexpr const int val = 5;
constexpr const int *val_ptr = &val;
constexpr void func() const;
```

alalek:
@JoeHowse @cv3d Thank you :+1:

PR #14117, opened 2019-03-21
Postpone variable definitions

103yiran:
### This pullrequest changes

 variable `i` is only used int for loops, move it in.


modules/dnn/src/dnn.cpp
```diff
@@ -283,7 +283,7 @@ void blobFromImages(InputArrayOfArrays images_, OutputArray blob_, double scalef
         images[i] *= scalefactor;
     }
 
-    size_t i, nimages = images.size();
+    size_t nimages = images.size();
```
alalek: There is declaration of this variable below (line 290):

> Mat image;

Its scope can be reduced too.

PR #14150, opened 2019-03-26
G-API: GComputation doesn't work with output vector<cv::Mat>

TolyaTalamanov:
## Problem overview
Consider the following case:
```
TEST(GComputationTest, VectorAsOutputApply)
 {
      cv::GMat in;
      cv::GMat out[3];
      std::tie(out[0], out[1], out[2]) = cv::gapi::split3(in);

      cv::Mat in_mat(3, 3, CV_8UC3);
      cv::randu(in_mat, cv::Scalar::all(0), cv::Scalar::all(255));
      std::vector<cv::Mat> out_mats(3);
      std::vector<cv::Mat> ref_mats(3);

      cv::split(in_mat, ref_mats);

      cv::GComputation({in}, {out[0], out[1], out[2]}).apply({in_mat}, out_mats);

      EXPECT_EQ(cv::Size(3, 3), out_mats[0].size());
      for (const auto& it : ade::util::zip(ref_mats, out_mats))
      {
          const auto& ref_mat = std::get<0>(it);
          const auto& out_mat = std::get<1>(it);

          EXPECT_EQ(0, cv::countNonZero(ref_mat != out_mat));
      }
}
```
This will work fine if the output matrices are preallocated
```
for (auto& out_mat : out_mats)
{
    out_mats.create(in_mat.size(), CV_8UC1)
}
```
This is due to the fact that preallocated mat simply copy a pointer to memory, when copying, but when  
the matrix is empty this does not happen and a new mat is created when copying

Therefore, the data calculated in the copy of the mat does't reach to the output mat of the user

### DIFF
--- a/modules/gapi/src/api/gcomputation.cpp
+++ b/modules/gapi/src/api/gcomputation.cpp
@@ -159,16 +159,14 @@ void cv::GComputation::apply(cv::Mat in1, cv::Mat in2, cv::Scalar &out, GCompile
 }
 
 void cv::GComputation::apply(const std::vector<cv::Mat> &ins,
-                             const std::vector<cv::Mat> &outs,
+                             std::vector<cv::Mat>       &outs,
                              GCompileArgs &&args)
 {
     GRunArgs call_ins;
     GRunArgsP call_outs;
 
-    // Make a temporary copy of vector outs - cv::Mats are copies anyway
-    auto tmp = outs;
-    for (const cv::Mat &m : ins) { call_ins.emplace_back(m);   }
-    for (      cv::Mat &m : tmp) { call_outs.emplace_back(&m); }
+    for (const cv::Mat &m : ins)  { call_ins.emplace_back(m);   }
+    for (      cv::Mat &m : outs) { call_outs.emplace_back(&m); }
 
     apply(std::move(call_ins), std::move(call_outs), std::move(args));
 }
--- a/modules/gapi/src/api/gcomputation.cpp
+++ b/modules/gapi/src/api/gcomputation.cpp
@@ -159,7 +159,7 @@ void cv::GComputation::apply(cv::Mat in1, cv::Mat in2, cv::Scalar &out, GCompile
 }
 
 void cv::GComputation::apply(const std::vector<cv::Mat> &ins,
-                             std::vector<cv::Mat>       &outs,
+                             std::vector<cv::Mat> &outs,
                              GCompileArgs &&args)
 {
     GRunArgs call_ins;
--- a/modules/gapi/src/api/gcomputation.cpp
+++ b/modules/gapi/src/api/gcomputation.cpp
@@ -159,7 +159,7 @@ void cv::GComputation::apply(cv::Mat in1, cv::Mat in2, cv::Scalar &out, GCompile
 }
 
 void cv::GComputation::apply(const std::vector<cv::Mat> &ins,
-                             std::vector<cv::Mat> &outs,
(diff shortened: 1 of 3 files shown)


TolyaTalamanov:
@dmatveev

dmatveev:
Thanks, but
1) Please add a detailed problem description - was issue you actually fixed
2) Add a test when vector of Mats have Mats pre-allocated. Test that
   - if Mats have proper formats, Mats itself are not reallocated
   - if Mats have wrong format or insufficient size, Mats are reallocated (pointers changed)

dmatveev:
@TolyaTalamanov please find other reviewer in the team first. My review is some last resort.

Thanks,
D

P.S. No changes here since the last 20 days?

TolyaTalamanov:
@rgarnov

modules/gapi/src/api/gcomputation.cpp
```diff
@@ -159,16 +159,14 @@ void cv::GComputation::apply(cv::Mat in1, cv::Mat in2, cv::Scalar &out, GCompile
 }
 
 void cv::GComputation::apply(const std::vector<cv::Mat> &ins,
-                             const std::vector<cv::Mat> &outs,
+                             std::vector<cv::Mat>       &outs,
```
rgarnov: Misalignment

modules/gapi/test/gapi_gcomputation_tests.cpp
```diff
@@ -51,6 +52,41 @@ namespace opencv_test
...
+          cv::GMat out[3];
+          cv::Mat  in_mat;
+          cv::GComputation m_c;
+          std::vector<cv::Mat> ref_mats;
+
+          GComputationVectorMatsAsOutput() : in_mat(300, 300, CV_8UC3),
+          m_c([&](){
+                      std::tie(out[0], out[1], out[2]) = cv::gapi::split3(in);
+                      return cv::GComputation({in}, {out[0], out[1], out[2]});
+                  })
```
dmatveev: This is the most incorrect way to use lambda-based constructor.

It's whole idea was to keep GMat stuff/etc inside the lambda so not polluting the outside scope with extra variables.
Here you do the opposite.

Please move `in` and `out[3]` into the lambda itself. You don't need it in the class.

modules/gapi/src/api/gcomputation.cpp
dmatveev: +1

TolyaTalamanov: Fixed

modules/gapi/test/gapi_gcomputation_tests.cpp
TolyaTalamanov: Fixed

TolyaTalamanov:
@rgarnov @dmatveev

modules/gapi/src/api/gcomputation.cpp
```diff
@@ -159,16 +159,14 @@ void cv::GComputation::apply(cv::Mat in1, cv::Mat in2, cv::Scalar &out, GCompile
 }
 
 void cv::GComputation::apply(const std::vector<cv::Mat> &ins,
-                             const std::vector<cv::Mat> &outs,
+                             std::vector<cv::Mat> &outs,
```
rgarnov: Please be consistent with code alignment here and in the header (personally I prefer alignment from the header)

TolyaTalamanov: Fixed

PR #14151, opened 2019-03-26
G-API: own::Mat as output computation doesn't work

TolyaTalamanov:
## Problem overview
If you try to use own::Mat as an output parameter, an exception is thrown


TolyaTalamanov:
@dmatveev

dmatveev:
@TolyaTalamanov please find other reviewer in the team first. My review is some last resort.

Thanks,
D

TolyaTalamanov:
@rgarnov

modules/gapi/src/executor/gexecutor.cpp
```diff
@@ -155,15 +155,18 @@ void cv::gimpl::GExecutor::run(cv::gimpl::GRuntimeArgs &&args)
...
-#else
+            if (cv::util::holds_alternative<cv::Mat*>(args.outObjs.at(index)))
+            {
+                auto& out_mat   = *get<cv::Mat*>(args.outObjs.at(index));
+                createMat(desc, out_mat);
+                continue;
+            }
+#endif // !defined(GAPI_STANDALONE)
             // Building standalone - output buffer should always exist,
             // and _exact_ match our inferred metadata
```
dmatveev: So after #ifdef scope is changed, this comment is not correct anymore and meta is checked always.

modules/gapi/src/executor/gexecutor.cpp
```diff
@@ -155,15 +155,18 @@ void cv::gimpl::GExecutor::run(cv::gimpl::GRuntimeArgs &&args)
...
             // Building as part of OpenCV - follow OpenCV behavior
             // if output buffer is not enough to hold the result, reallocate it
-            auto& out_mat   = *get<cv::Mat*>(args.outObjs.at(index));
-            createMat(desc, out_mat);
-#else
+            if (cv::util::holds_alternative<cv::Mat*>(args.outObjs.at(index)))
+            {
+                auto& out_mat   = *get<cv::Mat*>(args.outObjs.at(index));
+                createMat(desc, out_mat);
+                continue;
```
dmatveev: this `continue`-under-`ifdef` is mind-breaking.

Please avoid constructions like this because it is VERY unclear how you first get `cv::Mat` as `args.outObjs.at(index)` and then, a few lines below, get `own::Mat*` from exactly the same object.

modules/gapi/src/executor/gexecutor.cpp
TolyaTalamanov: Fixed

modules/gapi/src/executor/gexecutor.cpp
TolyaTalamanov: Fixed

modules/gapi/src/executor/gexecutor.cpp
```diff
@@ -155,14 +155,23 @@ void cv::gimpl::GExecutor::run(cv::gimpl::GRuntimeArgs &&args)
...
-            auto& out_mat   = *get<cv::Mat*>(args.outObjs.at(index));
-            createMat(desc, out_mat);
+            if (cv::util::holds_alternative<cv::Mat*>(args.outObjs.at(index)))
+            {
+                auto& out_mat = *get<cv::Mat*>(args.outObjs.at(index));
+                createMat(desc, out_mat);
+            }
+            else
+            {
+                auto& out_mat = *get<cv::gapi::own::Mat*>(args.outObjs.at(index));
```
rgarnov: I'd still place these common lines for own::Mat in a lambda. Also this behavior of Ocv mode doesn't correlate with the comment above (on reallocation), please add a comment on own::Mat related behavior (never reallocated, checked to perfectly fit required meta)

TolyaTalamanov: Fixed

TolyaTalamanov:
@rgarnov

TolyaTalamanov:
@dmatveev

TolyaTalamanov:
@dmatveev

dmatveev:
OKAY 👍 thanks!

PR #14156, opened 2019-03-26
videoio: added bad parameters handling to VideoWriter

mshabunin:
### This pullrequest changes

We expect no throw behavior for incorrect VideoWriter parameters, added test to check this.

Changes in GStreamer backend:
* replaced `abort()` calls with error handling

Changes in AVFoundation VideoWriter backend:
* added support to UTF-8 filenames (not tested)
* removed fallback to mp4/h264 codec in case when extension or FourCC are not supported (supported extensions: `mov, qt, mp4, m4v`, supported FourCC codes: `JPEG, MJPEG, jpeg, mjpeg, H264, avc1`)
* added basic error handling and cleanup

Changes in MediaSDK backend:
* added soft `fps` check and modified

```
force_builders=Linux AVX2
docker_image:Linux AVX2=ubuntu:18.04
```

### DIFF
--- a/modules/videoio/src/cap_avfoundation_mac.mm
+++ b/modules/videoio/src/cap_avfoundation_mac.mm
@@ -45,6 +45,7 @@
 #include "precomp.hpp"
 #include "opencv2/imgproc.hpp"
 #include <stdio.h>
+#include <AvailabilityMacros.h>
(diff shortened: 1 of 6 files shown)


alalek:
We definitely should ban all "assert()" calls in OpenCV code, including tests.

PR #12988, opened 2018-10-30
Add python test for copyto with mask

LaurentBerger:
About PR #12980

PR #14175, opened 2019-03-28
core: python - test cv::copyTo with pre-allocated dst

paroj:


alalek:
Could you add some test code for this case? (like this: #12988)

paroj:
modified the existing test to copy into an existing array

modules/python/test/test_copytomask.py
```diff
@@ -24,12 +24,13 @@ def test_copytomask(self):
         valeurBGRSup = np.array([70, 70,255])
         maskRed = cv.inRange(img, valeurBGRinf, valeurBGRSup)
         #New binding
-        dstcv  = cv.copyTo(img,maskRed)
+        dstcv = np.full_like(img, 255)
+        cv.copyTo(img, dstcv, maskRed)
```
alalek: `dstcv = cv.copyTo(img, mask=maskRed)` ?

paroj: this will fail as dstcv will be allocated in copyTo.

paroj:
> Do you want to support copyTo() into ROI of numpy array?

```
cv.copyTo(img, dst[whatever], mask)
```

should work

alalek:
What is about adding of separate call for that: `copyInto()`?

P.S. Original `copyTo()` behavior is not very clean - there are too many "if" conditions.

/cc @LaurentBerger

paroj:
> What is about adding of separate call for that: copyInto()?

how would that be different from copyTo? IMO copyTo should rather fail then re-creating dst, like you sketched in 62ed6cdc742db2f3065f2c472dc8216970d130b9

alalek:
`copyInto()`: forbid allocation/re-allocation of dst

`copyTo()`: always allocate new dst. "mask" parameter is useless (not defined well in case of new allocation) and should be removed.

paroj:
> copyTo(): always allocate new dst.

which would make it clone().

Anyway, I think refactoring the original Mat API is out of scope here..

LaurentBerger:
Sorry I don't understand problem. I tried

```
import cv2 as cv
import numpy as np

img1=cv.imread("g:/lib/opencv/samples/data/baboon.jpg")
img2=cv.imread("g:/lib/opencv/samples/data/lena.jpg")
mask = cv.inRange(img1,(50,50,50),(100,250,255))
cv.copyTo(img1,dst=img2,mask=mask)
cv.imshow("test",img2)
cv.waitKey()

```

result is
image

paroj:
you are, I was just too stupid to figure that argument order is changed in python. (output is moved to end)

alalek:
> dst[whatever]

Does this case work well? (ROI filling)

paroj:
yes

alalek:
There are no checks of results values in these tests =(.

This doesn't work:
```
    def test_copyto_roi(self):
        N = 8
        M = 6
        src = np.full((N, M), 64, dtype=np.uint8)
        dst = np.full((N*2, M*2), 128, dtype=np.uint8)
        dst2 = cv.copyTo(src, dst[N:N+N, M/2:M/2+M])
        print(dst)
        print(dst2)
        print(dst[N:N+N, M/2:M/2+M])
        self.assertEqual(dst[0, 0], 128)
        self.assertEqual(dst[N, M/2], 64)
        self.assertEqual(dst[N, M], 64)
        self.assertEqual(dst[N*2-1, M*2-1], 128)
```

paroj:
use `dst2 = cv.copyTo(src, None, dst[N:N+N, M//2:M//2+M])`

you are making the same mistake, that I did initially :wink:

alalek:
Right! Change to '//' is necessary for Python3.
Actually I'm using Python2 at this moment and `print(... roi ...)` works well for me. Unfortunately problem is somewhere else.

Added output into my previous comment.

paroj:
> Right! Change to '//' is necessary for Python3.

sorry for putting two changes in one line. I was referring to the 2nd argument **None** (mask).

LaurentBerger:
roi filling : 

import cv2 as cv
import numpy as np

img1=cv.imread("g:/lib/opencv/samples/data/baboon.jpg")
img2=cv.imread("g:/lib/opencv/samples/data/lena.jpg")
mask = cv.inRange(img1,(50,50,50),(100,250,255))
cv.copyTo(img1[200:500,200:500],dst=img2[200:500,200:500],mask=mask[200:500,200:500])
cv.imshow("test",img2)
cv.waitKey()

image

alalek:
@paroj You are right again!
@LaurentBerger Thanks!

`copyInto()` would avoid such mistakes (at least warn about them due "dst" allocation/re-allocation), current interface is really error-prone.

PR #14193, opened 2019-03-30
3rdparty(zlib): prevent uninitialized use of state->check

alalek:
To unblock future checks by fuzzer with memory sanitizer.


alalek:
:+1:

PR #14199, opened 2019-03-31
OpenBSD build fix

devnexen:
required for close calls.


alalek:
I believe we can extend `#if` macro condition (line 97 of system.cpp):
```.diff
- ... || defined __FreeBSD__
+ ... || defined __FreeBSD__ || defined(__OpenBSD__)
```

Additionally you can take a look on other cases with `__FreeBSD__` mark (for example, "filesystem" features doesn't break build, but they are stay disabled).

PR #14217, opened 2019-04-01
build: fix world G-API tests

alalek:
Force using of relative includes to avoid mess in world builds.

```
allow_multiple_commits=1
```


dmatveev:
Awesome, thanks a lot, folks!

PR #14225, opened 2019-04-02
Add support for   tf.add_n operation

zuoshaobo:
tensorflow tf.add_n op has supported by opencv already.Just add a judgement code.


alalek:
Thanks for the fix!
It would be nice to have some test for this case.

PR #14226, opened 2019-04-02
QString => const QString&

universeroc:
Fix function argument type from QString to const QString& to avoid copy-constructor call in window_QT.h && window_QT.cpp

```
force_builders=Custom
docker_image:Custom=qt:16.04
```

### DIFF
--- a/modules/highgui/src/window_QT.cpp
+++ b/modules/highgui/src/window_QT.cpp
@@ -396,7 +396,7 @@ CV_IMPL void cvStopLoop()
 }
 
 
-static CvWindow* icvFindWindowByName(QString name)
+static CvWindow* icvFindWindowByName(const QString& name)
 {
     CvWindow* window = 0;
 
@@ -424,7 +424,7 @@ static CvWindow* icvFindWindowByName(QString name)
 }
 
 
-static CvBar* icvFindBarByName(QBoxLayout* layout, QString name_bar, typeBar type)
+static CvBar* icvFindBarByName(QBoxLayout* layout, const QString& name_bar, typeBar type)
 {
     if (!layout)
         return NULL;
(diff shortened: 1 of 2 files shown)


modules/highgui/src/window_QT.cpp
```diff
@@ -1633,7 +1633,7 @@ CvWinProperties::~CvWinProperties()
 // CvWindow
 
 
-CvWindow::CvWindow(QString name, int arg2)
+CvWindow::CvWindow(const QString& name, int arg2)
```
mshabunin: There is no corresponding change in the header.

universeroc: OK，I will check it again and build through locally

universeroc:
Yeah，Copy On Write could reduce memory，from my experience，user defined types should use reference or pointer to pass the value. It is one best practice of C++. Think of performance and intuitive design，so use const reference as possible as you can for function argument. Hope you agree with me :)

universeroc:
@mshabunin I've made the changes to make the compiler work. Also I've checked the Qt doc for example QJsonValue 's constructor: QJsonValue(const String& s) etc.

modules/highgui/src/window_QT.h
```diff
@@ -120,36 +120,36 @@ class GuiReceiver : public QObject
...
+    double isFullScreen(const QString& name);
+    double getPropWindow(const QString& name);
+    void setPropWindow(const QString& name, double flags );
+    void setWindowTitle(const QString& name, const QString& title);
+    double getWindowVisible(const QString& name);
+    double getRatioWindow(const QString& name);
+    void setRatioWindow(const QString& name, double arg2 );
+    void saveWindowParameters(const QString& name);
+    void loadWindowParameters(const QString& name);
+    void putText(void* arg1, const QString& text, QPoint org, void* font);
```
csukuangfj: const QPoint& org

asmorkalov:
@universeroc Do you have any progress on that? There are build failures on CI and unresolved comments.

universeroc:
> @universeroc Do you have any progress on that? There are build failures on CI and unresolved comments.

Sorry, I've not enough time to fix all of those compiler errors with all kinds of platforms, maybe later I'll contribute the code again. Now I think close it maybe is the best choice.

PR #14255, opened 2019-04-04
Deep networks visualization

### LINKED ISSUE #9452: Deep networks visualization
##### Proposal
Add functionality to visualizing topology of network.

Two ways are possible: console output and torch-like `graphviz` visualization.

- **Console output**
  - Preferable for small architectures or for selected parts of networks. It's also useful for testing because our unit tests are based on few-layers networks.
  - Might looks like ```std::cout << net << std::endl;``` similarly for `cv::Mat` printing. However single method is more flexible for managing different levels of detalization (hyperparameters, input/output shapes, backend, fusion etc.). Minimal level of detalization may includes just layer types and connections, though.
  - Connections between layers seems to be a key issue because it requires appropriate indentation.

- **graphviz**
  - Easier than console printing. We just write text file with `.dot` graph. In example, 
    ```dot
    strict digraph sample {
      node1->node2
      node1->node3
      node2->node3
    }
    ```
    may be rendered into .png:
    ```bash
    dot -Tpng graph.dot > graph.png
    ```
    simple_graph

  - It has dozens of settings. Torch and supposedly Caffe use is for visualization. 

**Misc**: this project looks pretty nice for me: http://ethereon.github.io/netscope/quickstart.html

l-bat:
Fixes #9452

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -389,12 +389,12 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
          *  @returns String with structure, hyperparameters, backend, target and fusion
          *  To see correct backend, target and fusion run after forward().
          */
-        CV_WRAP String dumpNet();
+        CV_WRAP String dump();
         /** @brief Dump net structure, hyperparameters, backend, target and fusion to dot file
          *  @param path   path to output file with .dot extension
-         *  @see dumpNet()
+         *  @see dump()
          */
-        CV_WRAP void dumpNetToFile(const String& path);
+        CV_WRAP void dumpToFile(const String& path);
         /** @brief Adds new layer to the net.
          *  @param name   unique name of the adding layer.
          *  @param type   typename of the adding layer (type must be registered in LayerRegister).
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -385,6 +385,11 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         /** Returns true if there are no layers in the network. */
         CV_WRAP bool empty() const;
 
+        /** @brief Dump net structure, hyperparameters, backend, target and fusion to dot file
+         *  @param path   path to output file with .dot extension
+         *  To see correct backend, target and fusion run after forward().
+         */
+        void dumpNet(String path);
         /** @brief Adds new layer to the net.
          *  @param name   unique name of the adding layer.
          *  @param type   typename of the adding layer (type must be registered in LayerRegister).
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -385,11 +385,16 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         /** Returns true if there are no layers in the network. */
         CV_WRAP bool empty() const;
 
-        /** @brief Dump net structure, hyperparameters, backend, target and fusion to dot file
-         *  @param path   path to output file with .dot extension
+        /** @brief Dump net to String
+         *  @returns String with structure, hyperparameters, backend, target and fusion
          *  To see correct backend, target and fusion run after forward().
          */
-        void dumpNet(String path);
+        CV_WRAP String dumpNet();
+        /** @brief Dump net structure, hyperparameters, backend, target and fusion to dot file
+         *  @param path   path to output file with .dot extension
+         *  @see dumpNet()
+         */
+        CV_WRAP void dumpNetToFile(const String& path);
         /** @brief Adds new layer to the net.
          *  @param name   unique name of the adding layer.
          *  @param type   typename of the adding layer (type must be registered in LayerRegister).
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -47,6 +47,7 @@
 #include <algorithm>
 #include <iostream>
 #include <sstream>
+#include <fstream>
 #include <iterator>
 #include <numeric>
 #include <opencv2/dnn/shape_utils.hpp>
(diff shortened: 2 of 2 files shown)


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -385,6 +385,11 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         /** Returns true if there are no layers in the network. */
         CV_WRAP bool empty() const;
 
+        /** @brief Dump net structure, hyperparameters, backend, target and fusion to dot file
+         *  @param path   path to output file with .dot extension
+         *  To see correct backend, target and fusion run after forward().
+         */
+        void dumpNet(String path);
```
alalek: For testing purposes it is better to dump into string as primary functionality.

Storing into files can be an additional method.

---

> String

Should be `const String&` for input parameters.

modules/dnn/src/dnn.cpp
```diff
@@ -2947,6 +2948,196 @@ int Net::getLayerId(const String &layer)
     return impl->getLayerId(layer);
 }
 
+void Net::dumpNet(String path) {
+    CV_Assert(!empty());
+
+    std::ofstream file(path);
```
alalek: To dump into string use:
```
std::ostringstream out;
```

modules/dnn/include/opencv2/dnn/dnn.hpp
l-bat: done

modules/dnn/src/dnn.cpp
l-bat: done

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -385,6 +385,16 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         /** Returns true if there are no layers in the network. */
         CV_WRAP bool empty() const;
 
+        /** @brief Dump net to String
+         *  @returns String with structure, hyperparameters, backend, target and fusion
+         *  To see correct backend, target and fusion run after forward().
+         */
+        CV_WRAP String dumpNet();
```
dkurt: Method `dump` is a `cv::Net`'s method. So `cv::Net::dumpNet`. I think that we can name them just `dump()` and `dumpToFile`.

modules/dnn/src/dnn.cpp
```diff
@@ -2947,6 +2948,204 @@ int Net::getLayerId(const String &layer)
...
+                     DictValue dilation = lp.get("dilation");
+                     out << "dilation (HxW): " << dilation << "x" << dilation << "\\l";
+                 } else if (lp.has("dilation_h") && lp.has("dilation_w")) {
+                     DictValue h = lp.get("dilation_h");
+                     DictValue w = lp.get("dilation_w");
+                     out << "dilation (HxW): " << h << "x" << w << "\\l";
+                 }
+                 if (lp.has("pad")) {
+                     DictValue pad = lp.get("pad");
+                     out << "pad (HxW): " << pad << "x" << pad << "x" << pad << "x" << pad << "\\l";
```
dkurt: `pad (LxTxRxB)`?

modules/dnn/include/opencv2/dnn/dnn.hpp
l-bat: done

modules/dnn/src/dnn.cpp
l-bat: done

modules/dnn/src/dnn.cpp
```diff
@@ -2947,6 +2948,204 @@ int Net::getLayerId(const String &layer)
...
+                 if (lp.has("global_pooling")) {
+                     out << "global_pooling: " << lp.get("global_pooling") << "\\l";
+                 }
+                 if (lp.has("group")) {
+                     out << "group: " << lp.get("group") << "\\l";
+                 }
+             }
+         }
+         if (!it->second.outputBlobs.empty())
+             out << "output shapes: " << it->second.outputBlobs[0].size << "\\l";
```
dkurt: I think that's it's better to name it just `output:`.

Please also iterate by `it->second.outputBlobs[0].size` and print dimensions delimited by "x". For now it adds extra whitespaces.

l-bat: Fix name, added whitespaces everywhere " x ".

modules/dnn/src/dnn.cpp
```diff
@@ -2947,6 +2948,204 @@ int Net::getLayerId(const String &layer)
...
+                 }
+             }
+         }
+         if (!it->second.outputBlobs.empty())
+             out << "output: " << it->second.outputBlobs[0].size << "\\l";
+
+         out << (!it->second.backendNodes[prefBackend].empty() ? backend : "OCV/");
+         int colorId = 0;
+         switch (it->second.layerInstance->preferableTarget) {
+             case DNN_TARGET_CPU: out << "CPU\\n"; colorId = (backend == "OCV/") ? 0 : 5; break;
```
dkurt: Please check `prefBackend` instead because current approach doesn't work:

image

l-bat: fixed

modules/dnn/src/dnn.cpp
```diff
@@ -2947,6 +2948,206 @@ int Net::getLayerId(const String &layer)
...
+         }
+         if (!it->second.outputBlobs.empty())
+             out << "output: " << it->second.outputBlobs[0].size << "\\l";
+
+         Ptr<BackendNode> layerBackend = it->second.backendNodes[prefBackend];
+         out << (!layerBackend.empty() ? backend : "OCV/");
+         int colorId = 0;
+         switch (it->second.layerInstance->preferableTarget) {
+             case DNN_TARGET_CPU: out << "CPU\\n"; colorId = (layerBackend.empty() ||
+                                                   layerBackend->backendId == DNN_BACKEND_OPENCV) ? 0 : 5; break;
```
dkurt: I think we can check just `layerBackend.empty()` because OpenCV's implementation is not wrapped as a `BackendNode`.

l-bat: done

dkurt: @l-bat, Have you pushed the changes?

alalek:
It would be nice to add some test case for that.
- load some model (AlexNet?)
- check that output is not empty
- check that some layer name is dumped
- check that some "fused" layer is not showed

PR #14260, opened 2019-04-04
samples: Add C++ implementation of digits.py

allanrodriguez:
Currently, only a Python implementation of the SVM and KNearest digit recognition sample exists. This pull request adds a C++ implementation.

### DIFF
--- a/samples/cpp/digits.cpp
+++ b/samples/cpp/digits.cpp
@@ -1,0 +1,375 @@
+#include "opencv2/core.hpp"
+#include "opencv2/highgui.hpp"
+#include "opencv2/imgcodecs.hpp"
+#include "opencv2/imgproc.hpp"
+#include "opencv2/ml.hpp"
+
+#include <algorithm>
+#include <iostream>
+#include <vector>
+
+using namespace cv;
+using namespace std;
+
+const int SZ = 20;  // size of each digit is SZ x SZ
+const int CLASS_N = 10;
+const char* DIGITS_FN = "digits.png";
+
+static void help()
+{
+    cout <<
+    "\n"
+    "SVM and KNearest digit recognition.\n"
+    "\n"
+    "Sample loads a dataset of handwritten digits from 'digits.png'.\n"
+    "Then it trains a SVM and KNearest classifiers on it and evaluates\n"
+    "their accuracy.\n"
+    "\n"
+    "Following preprocessing is applied to the dataset:\n"
+    " - Moment-based image deskew (see deskew())\n"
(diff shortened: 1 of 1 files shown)


samples/cpp/digits.cpp
```diff
@@ -0,0 +1,468 @@
...
+        for (int j = 0; j < M.cols; j++)
+        {
+            M.at<float>(i, j) = M_vals[i][j];
+        }
+    }
+
+    warpAffine(img, deskewed_img, M, Size(SZ, SZ), WARP_INVERSE_MAP | INTER_LINEAR);
+}
+
+class StatModel
```
berak: isn't this redundant (and misleading) ? 
there's already a `ml::StatModel`, a `ml::KNearest` and a `ml::SVM`

samples/cpp/digits.cpp
```diff
@@ -0,0 +1,468 @@
...
+        }
+
+        static Ptr<KNearest> create(int k = 3)
+        {
+            return makePtr<KNearest>(k);
+        }
+
+        void load(const char* fn)
+        {
+            _model->load<ml::KNearest>(fn);
```
berak: this won't work (same for SVM).
`Algorithm::load()` is a static method, and returns a new instance, 

[see here for a usage example](https://github.com/opencv/opencv/blob/master/samples/cpp/letter_recog.cpp#L88)

samples/cpp/digits.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+    vector<Mat> digits;
+    vector<int> labels;
+
+    load_digits(DIGITS_FN, digits, labels);
+
+    cout << "preprocessing..." << endl;
+
+    // shuffle digits
+    shuffle(digits, labels);
```
berak: not sure about KNearest, but the SVM will do internal shuffeling already

samples/cpp/digits.cpp
```diff
@@ -0,0 +1,468 @@
...
+        double sum = 0;
+
+        for (size_t i = 0; i < hist.size(); i++)
+        {
+            sum += hist[i];
+        }
+
+        for (size_t i = 0; i < hist.size(); i++)
+        {
+            hist[i] /= sum + eps;
```
berak: (sum+eps);   // '+' binds weaker than '/='

samples/cpp/digits.cpp
allanrodriguez: I copied the Python implementation of this sample, which wraps all three classes, but I agree that it's redundant.

PR #14232, opened 2019-04-02
Reworked v_popcount implementation

terfendail:
### This pullrequest changes
Reworked v_popcount implementation to provide number of bits in a single lane

TODO:
- [x] Perf results

```
allow_multiple_commits=1
force_builders=Linux AVX2,Custom
buildworker:Custom=linux-1,linux-2,linux-4
docker_image:Custom=powerpc64le
```

PR #14261, opened 2019-04-05
core:unv improve and split SSE into multiple files

seiko2plus:
- [x] types
- [x] memory
- [x] operators
- [x] arithmetic
- [x] math
- [x] reorder
- [ ] misc
- [x] conversion
- [x] interleave

```
force_builders_only=linux,windows,mac,docs,Linux32
build_contrib:Win64=OFF
```


modules/calib3d/src/stereosgbm.cpp
```diff
@@ -705,7 +705,7 @@ static void computeDisparitySGBM( const Mat& img1, const Mat& img2,
                                 L0 = L0 + v_load(Sp + d);
                                 v_store(Sp + d, L0);
 
-                                v_int16x8 mask = _minS > L0;
+                                v_int16x8 mask = v_int16x8::fromMask(_minS > L0);
```
alalek: Can we keep this code "as is" at least for SIMD128? (make it backward compatible, see opencv_contrib)

SIMD256/SIMD512/wide universal intrinsics - new "mask" type.

cmake/OpenCVCompilerOptimizations.cmake
```diff
@@ -177,6 +179,7 @@ if(X86 OR X86_64)
     ocv_update(CPU_FMA3_FORCE "") # Don't force other optimizations
     ocv_update(CPU_FP16_IMPLIES "AVX")
     ocv_update(CPU_FP16_FORCE "") # Don't force other optimizations
+    ocv_update(CPU_XOP_IMPLIES "AVX")
```
alalek: There is wiki page: https://en.wikipedia.org/wiki/XOP_instruction_set

`_mm_comge_epu32` can be emulated via SSE2 too. [Example](http://sseplus.sourceforge.net/group__emulated___s_s_e2.html#g17da283c41a855ae9c34c4782bbce9d9).

cmake/OpenCVCompilerOptimizations.cmake
```diff
@@ -81,6 +81,7 @@ ocv_optimization_process_obsolete_option(ENABLE_SSE41 SSE4_1 ON)
 ocv_optimization_process_obsolete_option(ENABLE_SSE42 SSE4_2 ON)
 ocv_optimization_process_obsolete_option(ENABLE_POPCNT POPCNT ON)
 ocv_optimization_process_obsolete_option(ENABLE_AVX AVX ON)
+ocv_optimization_process_obsolete_option(ENABLE_XOP XOP ON)
```
alalek: This is list of legacy options (`ENABLE_XOP` doesn't exist before).

modules/core/include/opencv2/core/cvdef.h
```diff
@@ -288,6 +290,7 @@ enum CpuFeatures {
     CPU_AVX_512VBMI     = 20,
     CPU_AVX_512VL       = 21,
 
+    CPU_XOP             = 99,
```
alalek: Could you add link with wiki page here?
```
CPU_XOP             = 99,          //!< https://en.wikipedia.org/wiki/XOP_instruction_set
```

cmake/OpenCVCompilerOptimizations.cmake
seiko2plus: There's no any AVX intrinsics involved only SSE. The reason of implies AVX its because all AMD CPUs that support XOP also support AVX and MSVC requires AVX too

modules/calib3d/src/stereosgbm.cpp
seiko2plus: I think we can't that will break dispatch headers and there's no need to workaround it since ::fromMask only reinterpret mask types for SIMD/128&256.
So we're not expecting any performance cost and of course I will provide performance testing results to make sure that everything is okay.

BTW I moved AVX512 compatibility commits into separate PR #14331 and this PR should be merge after #14232 and backport of #14331.

vpisarev:
Dear Sayed (@seiko2plus), this is a similar situation as with https://github.com/opencv/opencv/pull/14331. I suggest to close this PR and then break down the still relevant improvements into a smaller pieces (e.g. breaking down big files with lot's of SIMD optimizations is definitely a good idea) and submit new PRs that we could review and merge much quicker. What do you think?

seiko2plus:
@vpisarev, Totaly agree, I will try to submit the new PRs as soon as possible.

PR #14271, opened 2019-04-06
calib3d: don't change input data in estimateAffine2D/estimateAffinePartial2D()

### LINKED ISSUE #14259: cv::estimateAffinePartial2D should not reorder input point sets
- OpenCV => 3.4.5
- Compiler => GCC 5.4.0

##### Detailed description

`cv::estimateAffinePartial2D` rearranges both `from` and `to` point sets with outliers present (see `compressElems()`). This violates the expectation that `cv::InputArray` is const. Additionally, `cv::OutputArray inliers` corresponds to the original, non-reordered points. This is the expected behavior for this parameter.
 
##### Steps to reproduce
Via code inspection or run any example that includes an inlier pair after an outlier.

alalek:
resolves #14259

Thanks @llschloesser for the test case!


alalek:
:+1:

PR #14308, opened 2019-04-12
Fix compile error with external TBB and -DBUILD_SHARED_LIBS=OFF

m000:
Adds `IMPORTED` for the `TBB::tbb` alias target as otherwise cmake produces the following error:
```
CMake Error: install(EXPORT "OpenCVModules" ...) includes target "opencv_core" which requires target "tbb" that is not in the export set.
```

This problem occurs with the defaults used by homebrew on macOS, which compiles both static and shared versions of TBB and OpenCV. The problem should also occur on other platforms when `-DBUILD_SHARED_LIBS=OFF` is set and an external TBB library is used. However, I don't have the setup to verify this in practice.

A similar fix is implemented in #12741 for another module.

```
Xbuild_tbb=ON
Xwith_tbb=ON
Xbuild_shared=OFF
```


alalek:
Thank you!

So, please:

PR #14325, opened 2019-04-14
DNN: Optimization fuse: Conv -> Eltwise -> BatchNorm

allnes:


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -318,7 +318,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
          * @param[in] top Next layer to be fused.
          * @returns True if fusion was performed.
          */
-        virtual bool tryFuse(Ptr<Layer>& top);
+        virtual bool tryFuse(std::vector< Ptr<Layer> >& bottoms, Ptr<Layer>& top);
```
alalek: I believe "layer" should know own inputs ("bottoms") - directly or via `Network`.

modules/dnn/src/layers/eltwise_layer.cpp
```diff
@@ -121,6 +121,37 @@ class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
...
+        std::vector< Ptr<dnn::Layer> > fakeInputs;
+        bool isFused = bottoms[0]->tryFuse(fakeInputs, top);
+        if (isFused)
+        {
+            Mat w, b;
+            top->getScaleShift(w, b);
+            if (!w.empty())
+            {
+                LayerParams weights_lp;
+                weights_lp.blobs.push_back(w.clone());
```
dkurt: Can we emit cloning?

modules/dnn/test/test_layers.cpp
```diff
@@ -1452,5 +1452,88 @@ TEST(Layer_Test_Convolution, relu_fusion)
...
+
+    net.setInput(input);
+    net.setPreferableBackend(DNN_BACKEND_OPENCV);
+    Mat output = net.forward();
+
+    std::vector<double> out_vec = output.reshape(1, 9);
+    for (size_t i = 0; i < out_vec.size(); i++)
+    {
+        EXPECT_EQ(out_vec[i], 16);
+    }
```
dkurt: Please initialize a reference `cv::Mat` and compare with it by `normAssert`.

modules/dnn/test/test_layers.cpp
```diff
@@ -1453,4 +1453,86 @@ TEST(Layer_Test_Convolution, relu_fusion)
     normAssert(input, output);
 }
 
+// Check fuse convolution and relu wherein skip eltwise
+TEST(Layers_Test_Subnet, convolutionx2_eltwise_scalex2_relu)
+{
+    Net net;
+    int id_conv[2];
+    int weightsShape[] = {1, 1, 1, 1};
+    Mat weights(4, &weightsShape[0], CV_32F, Scalar(2));
```
alalek: Inputs are too weak, so testing coverage is very low.

allnes: @alalek please review test again)

modules/dnn/src/layers/eltwise_layer.cpp
```diff
@@ -121,6 +121,37 @@ class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
         return false;
     }
 
+    virtual bool tryFuse(std::vector< Ptr<dnn::Layer> >& bottoms, Ptr<dnn::Layer>& top) CV_OVERRIDE
+    {
+        std::string currLayerType = bottoms[0]->type;
+        for (int i = 1; i < bottoms.size(); i++)
+        {
+            // Either all input layers are fusing or nothing.
+            if (currLayerType != bottoms[i]->type) { return false; }
```
dkurt: I think that
```cpp
// Either all input layers are fusing or nothing.
for (int i = 1; i < bottoms.size(); i++)
{
    if (bottoms[0]->type != bottoms[i]->type)
        return false;
}
```
it more convenient.

modules/dnn/test/test_layers.cpp
```diff
@@ -1453,4 +1453,88 @@ TEST(Layer_Test_Convolution, relu_fusion)
...
+        lp.type = "ReLU";
+        lp.name = "testReLU";
+        net.addLayerToPrev(lp.name, lp.type, lp);
+    }
+
+    int sz[] = {1, 1, 3, 3};
+    Mat input(4, &sz[0], CV_32F, Scalar(1));
+
+    net.setInput(input);
+    net.setPreferableBackend(DNN_BACKEND_OPENCV);
```
dkurt: I think we also need to check other backends and targets.

modules/dnn/src/layers/eltwise_layer.cpp
allnes: @dkurt fone)

modules/dnn/test/test_layers.cpp
allnes: @dkurt fone)

modules/dnn/test/test_layers.cpp
```diff
@@ -1453,4 +1453,94 @@ TEST(Layer_Test_Convolution, relu_fusion)
...
+
+    {
+        LayerParams lp;
+        lp.set("kernel_size", 1);
+        lp.set("num_output", 1);
+        lp.set("bias_term", false);
+        lp.type = "Convolution";
+        lp.name = "testConv_2";
+
+        lp.blobs.push_back(Mat(4, &weightsShape[0], CV_32F, Scalar(1)));
```
dkurt: Let's do layers different. It's OK to initialize weights by constant number but choose different values.

modules/dnn/test/test_layers.cpp
```diff
@@ -1453,4 +1453,94 @@ TEST(Layer_Test_Convolution, relu_fusion)
...
+        lp.blobs.push_back(Mat(Size(1, 1), CV_32F, Scalar(1)));
+        net.addLayerToPrev(lp.name, lp.type, lp);
+    }
+
+    {
+        LayerParams lp;
+        lp.set("bias_term", false);
+        lp.type = "Scale";
+        lp.name = "testScale_2";
+        lp.blobs.push_back(Mat(Size(1, 1), CV_32F, Scalar(1)));
```
dkurt: The same. Please initialize them differently.

modules/dnn/test/test_layers.cpp
```diff
@@ -1453,4 +1453,94 @@ TEST(Layer_Test_Convolution, relu_fusion)
...
+
+    {
+        LayerParams lp;
+        lp.type = "ReLU";
+        lp.name = "testReLU";
+        net.addLayerToPrev(lp.name, lp.type, lp);
+    }
+
+    int sz[] = {1, 1, 11, 11};
+    Mat input(4, &sz[0], CV_32F), bias(4, &sz[0], CV_32F, Scalar(0));
```
dkurt: `bias` Mat is unused.

modules/dnn/test/test_layers.cpp
allnes: @dkurt done

modules/dnn/test/test_layers.cpp
allnes: @dkurt done

modules/dnn/test/test_layers.cpp
allnes: @dkurt done

modules/dnn/test/test_layers.cpp
dkurt: I think it's better to use constant initializers rather random ones. Just different ones.

modules/dnn/test/test_layers.cpp
```diff
@@ -1445,4 +1445,98 @@ TEST(Layer_Test_Convolution, relu_fusion)
...
+    {
+        LayerParams lp;
+        lp.type = "ReLU";
+        lp.name = "testReLU";
+        net.addLayerToPrev(lp.name, lp.type, lp);
+    }
+
+    int sz[] = {1, 1, 11, 11};
+    Mat input(4, &sz[0], CV_32F);
+    randu(input, -1.0, 1.0);
```
dkurt: If we have ReLU I think we can face situations with zeroed values. So please initialize input positively and all the scales as well.

modules/dnn/test/test_layers.cpp
allnes: @dkurt done

modules/dnn/test/test_layers.cpp
allnes: @dkurt done

modules/dnn/test/test_layers.cpp
```diff
@@ -1448,4 +1448,93 @@ TEST(Layer_Test_Convolution, relu_fusion)
...
+    int weightsShape[] = {1, 1, 1, 1};
+    {
+        LayerParams lp;
+        lp.set("kernel_size", 1);
+        lp.set("num_output", 1);
+        lp.set("bias_term", false);
+        lp.type = "Convolution";
+        lp.name = "testConv_1";
+
+        lp.blobs.push_back(Mat(4, &weightsShape[0], CV_32F, Scalar(-0.38)));
```
alalek: No.
Each value of passed Mat should be different.

Test must be reliable. Including inputs.
Swapping of any of two different values on input should change result.

asmorkalov:
@allnes Do you have plan to finish the PR in mean time? There are merge conflicts and open discussions.

allnes:
> @allnes Do you have plan to finish the PR in mean time? There are merge conflicts and open discussions.
@asmorkalov I will plan finish PR until to New Year.

asmorkalov:
cc @VadimLevin

modules/dnn/test/test_layers.cpp
allnes: @alalek corrected)

modules/dnn/src/layers/eltwise_layer.cpp
```diff
@@ -210,6 +210,37 @@ class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
         return false;
     }
 
+    virtual bool tryFuse(std::vector< Ptr<dnn::Layer> >& bottoms, Ptr<dnn::Layer>& top) CV_OVERRIDE
+    {
+        std::string currLayerType = bottoms[0]->type;
```
dkurt: unused variable?

modules/dnn/src/layers/eltwise_layer.cpp
```diff
@@ -210,6 +210,37 @@ class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
...
+            if (!w.empty())
+            {
+                LayerParams weights_lp;
+                weights_lp.blobs.push_back(w);
+                Ptr<Layer> weightsLayer = ScaleLayer::create(weights_lp);
+                weightsLayer->finalize(std::vector<Mat>(1), std::vector<Mat>(1));
+                for (size_t i = 1; i < bottoms.size(); i++)
+                {
+                    CV_Assert(bottoms[i]->tryFuse(bottoms, weightsLayer));
+                }
```
dkurt: Please add a comment why we added this part of fusion. I remember that there was a case but it'd be useful to provide here an example in comments.

modules/dnn/test/test_layers.cpp
```diff
@@ -1583,6 +1583,134 @@ TEST(Layer_Test_Convolution, relu_fusion)
...
+{
+    Net net;
+    RNG rng;
+    int id_conv[2];
+    int weightsShape[] = {1, 1, 3, 3};
+
+    Mat conv_mat_1(4, &weightsShape[0], CV_32F);
+    Mat conv_mat_2(4, &weightsShape[0], CV_32F);
+    rng.fill(conv_mat_1, RNG::UNIFORM, Scalar(0), Scalar(1));
+    rng.fill(conv_mat_2, RNG::UNIFORM, Scalar(0), Scalar(1));
```
dkurt: Why not `randu`?

modules/dnn/test/test_layers.cpp
```diff
@@ -1583,6 +1583,134 @@ TEST(Layer_Test_Convolution, relu_fusion)
...
+    );
+
+    const int pad = 1;
+    Mat result(
+        conv_out_1(cvtest::Rect(pad, pad, output_2d.size[0], output_2d.size[1])) +
+        conv_out_2(cvtest::Rect(pad, pad, output_2d.size[0], output_2d.size[1]))
+    );
+
+    result = result * scale_mat_1.at<float>(0, 0);
+    result = result * scale_mat_2.at<float>(0, 0);
```
dkurt: Maybe it's easier to create two networks but for one of them set `enableFusion(false)`?

modules/dnn/src/layers/eltwise_layer.cpp
allnes: @dkurt corrected.

modules/dnn/src/layers/eltwise_layer.cpp
allnes: @dkurt wrote.

modules/dnn/test/test_layers.cpp
allnes: @dkurt changed.

modules/dnn/test/test_layers.cpp
allnes: @dkurt changed.

allnes:
@alalek pls merge https://github.com/opencv/opencv_extra/pull/684 request, else test not work

modules/dnn/src/layers/eltwise_layer.cpp
```diff
@@ -210,6 +210,45 @@ class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
...
+                //        \           |           X
+                //          \         |         X
+                //            \       |       X
+                //              \     |     X
+                //                \   |   X
+                //                fuse layer
+                //  We need to check that every layer will be fused
+                for (size_t i = 1; i < bottoms.size(); i++)
+                {
+                    CV_Assert(bottoms[i]->tryFuse(bottoms, weightsLayer));
```
alalek: Do we really need `CV_Assert()` here? This is "try" function, not validation.

In case of assertion (non-supported case) network is left in non-consistent (broken) state.
Some transaction-based fusing should be implemented instead.

alalek: > bottoms[i]->tryFuse(bottoms, weightsLayer)
> bottoms[i]
> bottoms

Refer to documentation of `tryFuse()` method.
You passing array including layer itself as layer's inputs. Do you see logic error with recursion here?
Related comment is still [here](https://github.com/opencv/opencv/pull/14325/files#r275332184).

asmorkalov:
@allnes Do you have progress on the patch?

modules/dnn/src/layers/eltwise_layer.cpp
allnes: I don't know how used without change API, thats why we need to close PR, maybe.

asmorkalov:
@allnes There are merge conflicts still.

PR #14327, opened 2019-04-14
fix integer overflow for p.x when step is large

### LINKED ISSUE #14326: LineIterator pos() overflows when coordinates and/or base data type is large

- OpenCV => 3.4
- Compiler => gcc 5.4.0

##### Detailed description

The `pos()` function of `LineIterator` (from the `imgproc` module) returns incorrect result (specifically, incorrect x coordinate) under certain conditions due to integer overflow.

##### Steps to reproduce

```.cpp
// Pardon the weird example.. it's from a specific use case from my work which was how i discovered this bug
Mat testImg = Mat::zeros(21481, 72320, CV_8UC2);
LineIterator testIt(testImg, Point(63894,21163), Point(63895, 21162));
for(int i = 0; i < testIt.count; ++testIt, i++ ) {
  Point p = testIt.pos();
  cout << i << ": " << p << endl;
}
```

tomforge:
resolves #14326

### This pullrequest changes
Add a cast to `long` for integer arithmetric that may potentially overflow in function `pos()` of `cv::LineIterator`

### DIFF
--- a/modules/imgproc/include/opencv2/imgproc.hpp
+++ b/modules/imgproc/include/opencv2/imgproc.hpp
@@ -4903,7 +4903,7 @@ Point LineIterator::pos() const
 {
     Point p;
     p.y = (int)((ptr - ptr0)/step);
-    p.x = (int)(((ptr - ptr0) - p.y*step)/elemSize);
+    p.x = (int)(((ptr - ptr0) - (long)p.y*step)/elemSize);
     return p;
 }
 


modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -4903,7 +4903,7 @@ Point LineIterator::pos() const
 {
     Point p;
     p.y = (int)((ptr - ptr0)/step);
-    p.x = (int)(((ptr - ptr0) - p.y*step)/elemSize);
+    p.x = (int)(((ptr - ptr0) - (long)p.y*step)/elemSize);
```
alalek: `long` is still [32-bit on Win32](https://en.cppreference.com/w/cpp/language/types).

Consider using `int64` here (or `long long int`).

asmorkalov:
@tomforge Do you have any progress on that?

tomforge:
@asmorkalov No. I haven't been able to find time to do any extensive testing. Maybe we should close this first?

PR #14210, opened 2019-04-01
AVX512 wide universal intrinsics

terfendail:
relates #13891

Implementation of 512-bit wide universal intrinsics for AVX512
https://github.com/opencv/opencv/wiki/OE-29.-Adding-AVX512-Support

<cut/>

TODO:
- [x] extend test_intrin tests

```
force_builders=docs,linux,Custom
buildworker:Custom=linux-3
build_image:Custom=ubuntu:18.04
CPU_BASELINE:Custom=AVX512_SKX
build_image:Custom Win=msvs2019
CPU_BASELINE:Custom Win=AVX512_SKX
#build_image:Custom Mac=
CPU_BASELINE:Custom Mac=AVX512_SKX
```

Issue #13891, opened 2019-02-21
Add AVX512 support


vpisarev:
The discussion thread about https://github.com/opencv/opencv/wiki/OE-29.-Adding-AVX512-Support

PR #14331, opened 2019-04-15
 core:unv add compatibility with avx512

seiko2plus:
WIP

relates #13891,  #14210
Merge with opencv_contrib: opencv/opencv_contrib#2099

### This pullrequest changes

* implement new traits for universal vectors
    - V128_Traits<lane type || vector type>
    - V256_Traits<lane type || vector type>
    - V_Traits<lane type || vector type> for wide vectors
    - replace V_RegTraits with V_Traits
 * initialize mask types
 * v_select requires mask type
 * comparisons return mask types
 * replace v_pack_b with v_pack
 * add new two intrinsics ::fromMask and ::from to convert between mask and other types
 * add mask types to v_signmask
 * add mask types to v_expand_low, v_expand_high and v_expand
 * add mask types to logical operators
 * remove conflicts that caused by avx512 compatibility

Todo:
- [ ] AVX512
- [ ] cover tests
- [x] opencv_contrib
- [ ] backport

````
pw_compilers=clang-4, clang-5, clang-6, gcc-4.9,power9_ gcc-5, power9_gcc-6, power9_gcc-7
pw_with_opencv_contrib=branch:core_avx512_compatibility
````

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin.hpp
@@ -139,13 +139,15 @@ using namespace CV_CPU_OPTIMIZATION_HAL_NAMESPACE;
 #   undef CV_FP16
 #endif
 
-#if CV_SSE2 || CV_NEON || CV_VSX
-#define CV__SIMD_FORWARD 128
-#include "opencv2/core/hal/intrin_forward.hpp"
+#if !CV_SSE2 && !CV_NEON && !CV_VSX
+#define CV_SIMD128_CPP 1
+#include "opencv2/core/hal/intrin_cpp.hpp"
 #endif
 
-#if CV_SSE2
+#define CV__SIMD_FORWARD 128
+#include "opencv2/core/hal/intrin_forward.hpp"
 
+#if CV_SSE2
 #include "opencv2/core/hal/intrin_sse_em.hpp"
 #include "opencv2/core/hal/intrin_sse.hpp"
 
@@ -156,11 +158,6 @@ using namespace CV_CPU_OPTIMIZATION_HAL_NAMESPACE;
 #elif CV_VSX
 
 #include "opencv2/core/hal/intrin_vsx.hpp"
-
-#else
-
-#define CV_SIMD128_CPP 1
-#include "opencv2/core/hal/intrin_cpp.hpp"
 
 #endif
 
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

**Comment by seiko2plus** [comment]
@terfendail, could you please check this patch?

File: modules/core/include/opencv2/core/hal/intrin_cpp.hpp [hunk 1]
```diff
@@ -318,6 +325,31 @@ template<typename _Tp, int n> struct v_reg
     */
     _Tp get0() const { return s[0]; }
 
+    static v_reg<_Tp, n> fromMask(const v_reg<bool, n>& mask)
```
terfendail: Could it be defined as converting constructor? I think that could allow to rebuild existing code without v_mask related updates.
@seiko2plus @alalek WDYT?

modules/core/include/opencv2/core/hal/intrin.hpp
```diff
@@ -346,9 +298,15 @@ namespace CV__SIMD_NAMESPACE {
     typedef v_uint64x4   v_uint64;
     typedef v_int64x4    v_int64;
     typedef v_float32x8  v_float32;
+    typedef v_mask8x32   v_mask8;
+    typedef v_mask16x16  v_mask16;
+    typedef v_mask32x8   v_mask32;
+    typedef v_maskf32x8  v_maskf32;
```
terfendail: Is it necessary to retain special mask types for floating point types? AFAIK for x86 and NEON integers and floats use the same registers so type casting is used for syntactic reasons only.

vpisarev:
Dear Sayed (@seiko2plus), thank you very much for all your on-going contributions to OpenCV! Regarding this PR, we are going to close it as a part of our attempt to cleanup the queue of PRs. A lot of work has been done here, but at the same time one of our engineers, Vitaly (@terfendail) has also been working on AVX512 support. So it was some luck of communication. And now this PR is out-of-date a bit and it would be very difficult to completely repair it. I suggest we close this PR now and then, if it contains still-relevant improvements, break them down into smaller pieces and submit new PR's that are easier to review and merge quickly. One of the questions, that should be discussed, is what to do with AVX512 masks, which is the essential part of this instruction set? Probably here is the best place to do it: https://github.com/opencv/opencv/issues/13891. What do you think?

seiko2plus:
Hello Vadim(@vpisarev), I am pleased and honored to contribute to OpenCV and I hope to increase my contributions in the future.
I'm already familiar with this policy and respect it so it's okay to close it. It doesn't matter me or Vitaly at the end job is done in a good
way.
 I think gonna follow your advice, as usual, to re-implement this work into several patches also with extra improvements like
supporting KNL and sure I will discuss it first in #13891. currently, I'm trying to implement what we doing here for NumPy hopefully I gain more experience from this mission that it may help me to improve our SIMD interface not just for AVX512 but for other extensions too especially NEON and VSX.
Spasibo

modules/core/include/opencv2/core/hal/intrin.hpp
seiko2plus: No not necessary, actually the reason behind these special types is just because I was worried too much about the plenty of switching between integer and floating-point unit domains in old x86 processors which may occur delays but I guess compilers can handle it, so I think we can just reinterpreting the return of float-point comparisons to integer types.

modules/core/include/opencv2/core/hal/intrin_cpp.hpp
seiko2plus: I don't think so, that will break the type-safety, but I prefer to use converting constructors and operators together for each vector type only so we can avoid the use of sub access to 'val', it silly to keep typing ".val" in every single line. for example, take a look at the defined types in #14261.

PR #14137, opened 2019-03-24
Loadknearest

LaurentBerger:
Add Ptr<KNearest> KNearest::load and python binding

PR #14332, opened 2019-04-15
Add Ptr KNearest::load and python binding

LaurentBerger:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/597

Add Ptr KNearest::load and python binding

relates #14137

```
force_builders=linux,docs
test_modules=python2,python3
```

### DIFF
--- a/modules/ml/misc/python/test/test_knearest.py
+++ b/modules/ml/misc/python/test/test_knearest.py
@@ -1,0 +1,13 @@
+#!/usr/bin/env python
+import cv2 as cv
+
+from tests_common import NewOpenCVTests
+
+class knearest_test(NewOpenCVTests):
+    def test_load(self):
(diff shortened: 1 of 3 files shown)


modules/ml/misc/python/test/test_knearest.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
+import cv2 as cv
+
+from tests_common import NewOpenCVTests
+
+class knearest_test(NewOpenCVTests):
+    def test_load(self):
+        k_nearest = cv.ml.KNearest_load("opencv_ml_knn.xml")
```
alalek: > "opencv_ml_knn.xml"

Try this instead `self.find_file("ml/opencv_ml_knn.xml")`

PR #14338, opened 2019-04-15
build: avoid highgui dependency

alalek:
replaces #13849


alalek:
@tedsteiner Please take a look on this patch.

tedsteiner:
@alalek This accomplishes what I need from #13849, thanks.

PR #14340, opened 2019-04-15
Another sample invisibility effect

cabelo:
### This pullrequest changes

@alalek this is a example  (Invisibility) with OpenCV (using morphologyEx, bitwise_not, absdiff and another funcions)

Please. Sorry for the confusion in past, i found the problem. 

I  used  this example in this work: https://www.youtube.com/watch?v=h5savmDl5G0
opencv_01


cabelo:
@alalek Is this example bad?

dkurt:
Let's close this PR. It's cool sample but it does not demonstrates real application for OpenCV. It looks like background subtraction, but we already have samples for it.

PR #14342, opened 2019-04-16
Fix for JavaCamera2View not displaying colors correctly on some devices

### LINKED ISSUE #13526: Android javacamera2 green color in screen

- OpenCV => :4.0.1  Android pack

##### Detailed description

When using Opencv Javacamera2View , I get a green color on the camera preview. When i switch to JavaCameraView, the camera diplays well.

screenshot_20181225-162646

Victorlouisdg:
resolves #13526

### This pullrequest changes

Changed the way Images are converted into rgba Mats so that JavaCamera2View also shows correct colors when the Image contains I420 or YV12 data.

Converting Images to Mats now checks the properties of the Image Planes to choose the correct conversion method. Two main case can be discerned: 
1. The pixel stride of the chroma channels equals 2. This means the Image data is in NV21 format. This case was already supported and the code for the conversion was reused.
2. The pixel stride of the chroma channels equals 1. This means the chroma channels of the image are not interleaved. This case was not supported yet and caused the the colors of the image to be displayed incorrectly (the image was mostly green/purple as seen below). This was fixed by checking the pixel stride of the chroma channels, and when it equals 1, using code Imgproc.COLOR_YUV2RGBA_I420 for the conversion.

color_bug

Another thing I noticed is that in case of non-interleaved chroma channels, for certain resolutions, the row stride of the chroma channels did not equal the width of the chroma channels. For example, resolution 720x480 gave a chroma channel width of 360, but the row stride was 368 on my device, meaning there were 8 extra padding bytes that should not be copied into the Mat. This also caused artefacts (green lines and spread out colors). This was fixed by copying the chroma channel ByteBuffer contents row by row when the row stride does not equals the width of the channel.

stride_bug2

### DIFF
--- a/modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
+++ b/modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
@@ -181,17 +181,7 @@ public class JavaCamera2View extends CameraBridgeViewBase {
                     assert (planes.length == 3);
                     assert (image.getFormat() == mPreviewFormat);
 
-                    // see also https://developer.android.com/reference/android/graphics/ImageFormat.html#YUV_420_888
-                    // Y plane (0) non-interleaved => stride == 1; U/V plane interleaved => stride == 2
-                    assert (planes[0].getPixelStride() == 1);
-                    assert (planes[1].getPixelStride() == 2);
-                    assert (planes[2].getPixelStride() == 2);
-
-                    ByteBuffer y_plane = planes[0].getBuffer();
-                    ByteBuffer uv_plane = planes[1].getBuffer();
-                    Mat y_mat = new Mat(h, w, CvType.CV_8UC1, y_plane);
-                    Mat uv_mat = new Mat(h / 2, w / 2, CvType.CV_8UC2, uv_plane);
-                    JavaCamera2Frame tempFrame = new JavaCamera2Frame(y_mat, uv_mat, w, h);
+                    JavaCamera2Frame tempFrame = new JavaCamera2Frame(image);
                     deliverAndDrawFrame(tempFrame);
                     tempFrame.release();
                     image.close();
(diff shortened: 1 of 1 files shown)


floe:
Just had a quick look, pretty busy ATM - wouldn't it be cleaner to call through to a suitable native cvtColor* method? Byte-wise copying in Java is probably rather ineffective...

Victorlouisdg:
I agree the Byte-wise copying is a suboptimal solution, but currently I don't see any other solution for 2 reasons:
1. Android Images with [ImageFormat.YUV_420_888](https://developer.android.com/reference/android/graphics/ImageFormat.html#YUV_420_888) provide the 3 image planes (Y, U and V) separately. There is no guarantee that the underlying YUV data is contiguous in memory, which means we can't just read the data in either a single YUV Mat or in a Y Mat and a UV Mat. 
2. To make things even more complicated, the Y, U or V byte data can have padding (zero bytes) between the different rows of the image. For example, the data in the ByteBuffer of the U channel of a 8x8 pixels YUV_I420 image could look like this (where U represent a U byte and 0 a zero byte):
UUUU00
UUUU00
UUUU00
UUUU
In this case the width of the U channel would be 4, but the row stride would be 6. Doing the conversion with a Mat that uses this ByteBuffer would yield annoying artefacts on the results.

If we want to support all Android devices, the 2 reasons above are why we either need to copy bytes, or we need a cvtColor function that takes 3 Mats as input and takes into account possible row stride.

 In my opinion the best way to ensure that OpenCV works seamlessly with in Android in future would be to add a cvtColor function that handles all Android Images with format YUV_420_888. Android recommends using this generic YUV_420_888 format. The new cvtColor function would need to take as inputs:
- 3 seperate Mats for Y, U and V (because the byte arrays might not be contiguous)
- U and V pixel stride (to determine whether the U and V channels are interleaved)
- U and V row stride (so we can account for the possible padding bytes between rows)
- a destination Mat

Also note that for devices that provide NV21 YUV data, the code runs just as fast as before, the conversion code is unchanged in that case. The copying of bytes in this pull request will only happen on device that show incorrect colors with the current implementation. So even though the copying is an imperfect solution, I believe it is better than doing an incorrect conversion.

alalek:
@Victorlouisdg Thank you for clarification!

Possible ways for optimizations here:
a. Add separate "fast path" without manual copying, but with required pre-checks on input.
b. We can implement copying in JNI C++ code. Some bridge between Android Images and OpenCV cvtColor or cvtColorTwoPlane calls. There is [code](https://github.com/opencv/opencv/blob/3.4.6/modules/java/generator/src/cpp/utils.cpp#L26) with support for Android Bitmap.

alalek:
Well, looks like old code patch is almost still the same (without slow copying).
I believe we can merge this patch if there are no performance regressions.

Can we memoize already processed images (for case of sequential calls)?
- `.gray()` method in `mGray` field
- `.rgba()` in `mRgba`

```
public Mat gray() {
    if (mGray && !mGray.empty())
        returm mGray;
    ...
}
```


Please also note that `Mat()` constructor from `ByteBuffer` doesn't copy data. Also there is no reference back on `ByteBuffer`. Assumed that `Mat` scope is strictly less than `ByteBuffer`. So lifetime of derived objects must be properly re-checked again.

Victorlouisdg:
Ok I will add the memoizing to `JavaCameraView2`. Should this also be added to the `JavaCameraView`? And should the memoizing be added in this pull request or should a new one be opened?

How would you check the lifetime of the derived `Mat` objects and do we need to check this? Currently it is not valid to use a `CvCameraViewFrame` object outside the `onCameraFrame()` callback that delivered it (see [this comment](https://github.com/opencv/opencv/blob/e8ac7a34dde5d33a1c13a83a0068cb17d3eb261e/modules/java/generator/android/java/org/opencv/android/CameraBridgeViewBase.java#L178)). If a user wants to use a `Mat` object created by calling `.rgba()` or `.gray()` for a longer duration, he has to make his own deep copy of that `Mat`.

alalek:
New one PR is fine.

---

> check the lifetime

I don't suggest to add code for checking this in runtime. I mean we should review code more carefully on this scope.

```
JavaCamera2Frame tempFrame = new JavaCamera2Frame(image);
deliverAndDrawFrame(tempFrame);
```

Looks like this code part is fine.

Suspicious code is here:
```
ByteBuffer y_plane = planes[0].getBuffer();
mGray = new Mat(h, w, CvType.CV_8UC1, y_plane);
return mGray;
```

`ByteBuffer y_plane`'s "data" is not locked. And object itself can be released after "return".

Android documentation states that this buffer is "direct":

> Get a direct ByteBuffer containing the frame data.

Hopefully, its lifetime is not less than associated "Image".

PR #14368, opened 2019-04-19
Check if Inference Engine networks are fully supported by backend

dkurt:
### This pullrequest changes

Observed regression as `DNNTestNetwork.Inception_5h` test with LRN layer. Add more checks to be ensured that all the layers from networks are supported by IE backend

Also remove skips in case of missed test data.

<cut/>

```
force_builders=Custom,Linux AVX2
buildworker:Custom=linux-2
docker_image:Custom=ubuntu-openvino-2019r1:16.04
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
test_modules:Custom=dnn

docker_image:Linux AVX2=ubuntu-openvino-2018r5:16.04
test_opencl:Linux AVX2=OFF
build_parallel_tests:Linux AVX2=1
test_bigdata:Linux AVX2=1
test_module:Linux AVX2=dnn,python2,python3,java
tests_filter:Linux AVX2=*

buildworker:Mac OpenCL=macosx-1
build_image:Mac OpenCL=openvino-2019r1
test_module:Mac OpenCL=dnn,python2,python3,java
```

### DIFF
--- a/modules/dnn/src/layers/lrn_layer.cpp
+++ b/modules/dnn/src/layers/lrn_layer.cpp
@@ -91,7 +91,7 @@ public:
     virtual bool supportBackend(int backendId) CV_OVERRIDE
     {
         if (backendId == DNN_BACKEND_INFERENCE_ENGINE)
-            return bias == 1;
+            return bias == (int)bias;
         return backendId == DNN_BACKEND_OPENCV || backendId == DNN_BACKEND_HALIDE;
     }
 
--- a/modules/dnn/test/test_backends.cpp
+++ b/modules/dnn/test/test_backends.cpp
@@ -45,7 +45,7 @@ public:
         netDefault.setInput(inp);
         Mat outDefault = netDefault.forward(outputLayer).clone();
 
-        Net net = readNet(weights, proto);
+        net = readNet(weights, proto);
         net.setInput(inp);
         net.setPreferableBackend(backend);
         net.setPreferableTarget(target);
@@ -94,6 +94,8 @@ public:
         else
             normAssert(ref, out, msg, l1, lInf);
     }
+
+    Net net;
 };
 
 TEST_P(DNNTestNetwork, AlexNet)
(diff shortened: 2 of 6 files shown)


dkurt:
@alalek, It looks like ONNX's ShuffleNet test data is missed from buildbot. May I ask you to check it? Thanks!

https://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/2035/steps/test_dnn-opencl/logs/stdio

https://github.com/opencv/opencv_extra/blob/8869985f12606d02fecbfed5533744cd51e01609/testdata/dnn/download_models.py#L778-L800

alalek:
@dkurt Thanks! Test data has been updated on builders.

modules/dnn/src/layers/lrn_layer.cpp
```diff
@@ -91,7 +91,7 @@ class LRNLayerImpl CV_FINAL : public LRNLayer
     virtual bool supportBackend(int backendId) CV_OVERRIDE
     {
         if (backendId == DNN_BACKEND_INFERENCE_ENGINE)
-            return bias == 1;
```
csukuangfj: what is the problem with `bias == 1`?

alalek: `bias` is floating point number in general.
IE API expects integers only.

csukuangfj: @alalek 
thanks for the reply.

I cannot see the point of changing
`bias == 1`
to
`bias == (int)bias`

Why is it related to the constraint of IE API, as it is only a comparison?

alalek: Code looks good to me:
```
if (backendId == DNN_BACKEND_INFERENCE_ENGINE)
    return bias == (int)bias;  // only integers are supported by IE
... other backends ...
```

dkurt: @csukuangfj, Inference Engine may works with `bias == 1` and `bias == 2` and any other integers. However it cannot manage `bias == 1.1` in example. That is why the condition has been changed to `bias == (int)bias` which is similar to `bias == std::floor(bias)` that means bias is integer.

csukuangfj: @dkurt 
aha, i see. Thanks!

PR #14376, opened 2019-04-21
FIX: getting platformIDs in initializeContextFromD3D11Device

masa-iwm:
### This pullrequest changes
relates #13972

fix usage of clGetPlatformIDs in initializeContextFromD3D11Device

### DIFF
--- a/modules/core/src/directx.cpp
+++ b/modules/core/src/directx.cpp
@@ -256,6 +256,10 @@ Context& initializeContextFromD3D11Device(ID3D11Device* pD3D11Device)
         CV_Error(cv::Error::OpenCLInitError, "OpenCL: No available platforms");
 
     std::vector<cl_platform_id> platforms(numPlatforms);
+    status = clGetPlatformIDs(numPlatforms, &platforms[0], NULL);
+    if (status != CL_SUCCESS)
+        CV_Error(cv::Error::OpenCLInitError, "OpenCL: Can't get number of platforms");
+
     size_t exts_len;
     cv::AutoBuffer<char> extensions;
     bool is_support_cl_khr_d3d11_sharing = false;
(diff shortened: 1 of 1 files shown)


Mainvooid:
This is not a bug, but the error message in line 269 is wrong. 
`"OpenCL: Can't get number of platforms"` 
should be changed to
 `"OpenCL: Can't get platform ID"`
@masa-iwm Can you revise and resubmit it?

@alalek

masa-iwm:
@Mainvooid Really? I think that `platforms[i]` will always be the value of the first platform.
https://github.com/opencv/opencv/blob/5c0a98cfb63cbbf5c25dc5ad9ba2ece51c191097/modules/core/src/directx.cpp#L267
In fact, [samples/directx/d3d11_interop.cpp](https://github.com/opencv/opencv/blob/5c0a98cfb63cbbf5c25dc5ad9ba2ece51c191097/samples/directx/d3d11_interop.cpp) does not work property in my environment (OpenCL device No.0=Intel & No.1=NVIDIA, but primary DirectX device is NVIDIA), but it works correctly with my modified.

I think that your code does not specify platform index for clGetPlatformIDs in a for loop, and cannot specify platform index in `clGetPlatformIDs`.

@alalek

modules/core/src/directx.cpp
```diff
@@ -264,9 +268,6 @@ Context& initializeContextFromD3D11Device(ID3D11Device* pD3D11Device)
 #endif
     for (int i = 0; i < (int)numPlatforms; i++)
     {
-        status = clGetPlatformIDs(numPlatforms, &platforms[i], NULL);
```
alalek: `&platforms[i]`

Memory corruption was here if `numPlatforms` > 1

Mainvooid:
@masa-iwm 
I checked the documentation again, you are right. Since there is only one platform on my computer, I didn't find this error.
So it should be changed to the following code and placed in front.
```cpp
status = clGetPlatformIDs(numPlatforms, &platforms[0], NULL);
if (status != CL_SUCCESS)
    CV_Error(cv::Error::OpenCLInitError, "OpenCL: Can't get platforms");
```
After this change, we can get a list of OCL platforms found outside the for loop and then check required extensions for each platform within the for loop.

PR #14391, opened 2019-04-22
Fixed array indexing bugs in python hough lines tutorials

obroomhall:
### This pullrequest changes

The hough lines tutorials for python have a bug where they only display the first detected line, due to an array indexing bug. This pull request changes the array indexing so that all lines are displayed.


samples/python/tutorial_code/imgProc/hough_line_transform/hough_line_transform.py
```diff
@@ -6,8 +6,8 @@
 edges = cv.Canny(gray,50,150,apertureSize = 3)
 
 lines = cv.HoughLines(edges,1,np.pi/180,200)
-for line in lines:
-    rho,theta = line[0]
+for line in lines[0]:
```
alalek: This code draws one line only.

Please take a look on existed code from Python samples: https://github.com/opencv/opencv/blob/3.4.6/samples/python/houghlines.py

Probably this code should be dropped and tutorial should use Python sample directly.

PR #14402, opened 2019-04-23
G-API planar GMatDesc

rgarnov:
* Added `planar` flag to `GMatDesc`, `asPlanar()` and `asInterleaved()` combinators, integrated into framework, added tests

### DIFF
--- a/modules/gapi/include/opencv2/gapi/gmat.hpp
+++ b/modules/gapi/include/opencv2/gapi/gmat.hpp
@@ -46,6 +46,10 @@ private:
     std::shared_ptr<GOrigin> m_priv;
 };
 
+namespace gapi { namespace own {
+    class Mat;
+}}//gapi::own
+
 /** @} */
 
 /**
@@ -58,16 +62,28 @@ struct GAPI_EXPORTS GMatDesc
     int depth;
     int chan;
     cv::gapi::own::Size size; // NB.: no multi-dimensional cases covered yet
+    bool planar;
+
+    GMatDesc(int d, int c, cv::gapi::own::Size s, bool p = false)
+        : depth(d), chan(c), size(s), planar(p) {}
+
+    GMatDesc() : GMatDesc(-1, -1, {-1,-1}) {}
 
     inline bool operator== (const GMatDesc &rhs) const
     {
-        return depth == rhs.depth && chan == rhs.chan && size == rhs.size;
+        return depth == rhs.depth && chan == rhs.chan && size == rhs.size && planar == rhs.planar;
     }
 
     inline bool operator!= (const GMatDesc &rhs) const
     {
         return !(*this == rhs);
     }
+
+    // Checks if the passed mat can be described by this descriptor
+    // (it handles the case when
+    // 1-channel mat can be reinterpreted as is (1-channel mat)
+    // and as a 3-channel planar mat with height divided by 3)
+    bool canDescribe(const cv::gapi::own::Mat& mat) const;
 
     // Meta combinator: return a new GMatDesc which differs in size by delta
     // (all other fields are taken unchanged from this GMatDesc)
@@ -88,6 +104,8 @@ struct GAPI_EXPORTS GMatDesc
     {
         return withSize(to_own(sz));
     }
+
+    bool canDescribe(const cv::Mat& mat) const;
 #endif // !defined(GAPI_STANDALONE)
     // Meta combinator: return a new GMatDesc which differs in size by delta
     // (all other fields are taken unchanged from this GMatDesc)
@@ -125,6 +143,43 @@ struct GAPI_EXPORTS GMatDesc
         desc.chan = dchan;
         return desc;
     }
+
+    // Meta combinator: return a new GMatDesc with planar flag set
+    // (no size changes are performed, only channel interpretation is changed
+    // (interleaved -> planar)
+    GMatDesc asPlanar() const
+    {
+        GAPI_Assert(planar == false);
+        GMatDesc desc(*this);
+        desc.planar = true;
+        return desc;
+    }
+
+    // Meta combinator: return a new GMatDesc
+    // reinterpreting 1-channel input as planar image
+    // (size height is divided by plane number)
+    GMatDesc asPlanar(int planes) const
+    {
+        GAPI_Assert(planar == false);
+        GAPI_Assert(chan == 1);
+        GAPI_Assert(planes > 1);
+        GAPI_Assert(size.height % planes == 0);
+        GMatDesc desc(*this);
+        desc.size.height /=  planes;
+        desc.chan = planes;
+        return desc.asPlanar();
+    }
+
+    // Meta combinator: return a new GMatDesc with planar flag set to false
+    // (no size changes are performed, only channel interpretation is changed
+    // (planar -> interleaved)
+    GMatDesc asInterleaved() const
+    {
+        GAPI_Assert(planar == true);
+        GMatDesc desc(*this);
+        desc.planar = false;
+        return desc;
+    }
 };
 
 static inline GMatDesc empty_gmat_desc() { return GMatDesc{-1,-1,{-1,-1}}; }
@@ -138,7 +193,6 @@ GAPI_EXPORTS GMatDesc descr_of(const cv::UMat &mat);
 /** @} */
 
 namespace gapi { namespace own {
-    class Mat;
     GAPI_EXPORTS GMatDesc descr_of(const Mat &mat);
 }}//gapi::own
 
--- a/modules/gapi/include/opencv2/gapi/gproto.hpp
+++ b/modules/gapi/include/opencv2/gapi/gproto.hpp
@@ -112,8 +112,16 @@ GMetaArgs GAPI_EXPORTS descr_of(const GRunArgs &args);
 
 // Transform run-time operation result argument into metadata extracted from that argument
 // Used to compare the metadata, which generated at compile time with the metadata result operation in run time
-GMetaArg  GAPI_EXPORTS descr_of(const GRunArgP& argp);
+GMetaArg GAPI_EXPORTS descr_of(const GRunArgP& argp);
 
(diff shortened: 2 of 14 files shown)


modules/gapi/include/opencv2/gapi/gmat.hpp
```diff
@@ -58,17 +62,23 @@ struct GAPI_EXPORTS GMatDesc
     int depth;
     int chan;
     cv::gapi::own::Size size; // NB.: no multi-dimensional cases covered yet
+    bool planar;
+
+    GMatDesc(int d = {}, int c = {}, cv::gapi::own::Size s = {}, bool p = false)
+        : depth(d), chan(c), size(s), planar(p) {}
```
dmatveev: `int d = {}`?

modules/gapi/include/opencv2/gapi/gmat.hpp
```diff
@@ -58,17 +62,23 @@ struct GAPI_EXPORTS GMatDesc
...
-        return depth == rhs.depth && chan == rhs.chan && size == rhs.size;
+        return depth == rhs.depth && chan == rhs.chan && size == rhs.size && planar == rhs.planar;
     }
 
     inline bool operator!= (const GMatDesc &rhs) const
     {
         return !(*this == rhs);
     }
 
+    inline bool canDescribe(const cv::gapi::own::Mat& mat) const;
```
dmatveev: Please add a comment to this function

dmatveev: btw maybe `matches()` the right name?

rgarnov:
Currently CI is red due to ABI changes (GMatDesc constructors), one more good topic for discussion

modules/gapi/include/opencv2/gapi/gmat.hpp
```diff
@@ -138,12 +187,25 @@ GAPI_EXPORTS GMatDesc descr_of(const cv::UMat &mat);
...
+    return *this == desc;
+}
+
+#if !defined(GAPI_STANDALONE)
+bool GMatDesc::canDescribe(const cv::Mat& mat) const
+{
+    const auto desc = planar ? descr_of(mat).asPlanar(chan) : descr_of(mat);
+    return *this == desc;
+}
+#endif
```
dmatveev: Suggest to implement as a template with two instantiations (avoiding code duplication)

modules/gapi/include/opencv2/gapi/gproto.hpp
```diff
@@ -112,8 +112,14 @@ GMetaArgs GAPI_EXPORTS descr_of(const GRunArgs &args);
...
 // Used to compare the metadata, which generated at compile time with the metadata result operation in run time
-GMetaArg  GAPI_EXPORTS descr_of(const GRunArgP& argp);
+GMetaArg GAPI_EXPORTS descr_of(const GRunArgP& argp);
 
+// Says if run-time computation argument can be described by metadata
+bool GAPI_EXPORTS can_describe(const GMetaArg&  meta,  const GRunArg&  arg);
+bool GAPI_EXPORTS can_describe(const GMetaArgs& metas, const GRunArgs& args);
+
+// Says if run-time computation result argument can be described by metadata
+bool GAPI_EXPORTS can_describe(const GMetaArg&  meta,  const GRunArgP& argp);
```
dmatveev: Says -> checks

modules/gapi/src/api/gbackend.cpp
```diff
@@ -349,5 +349,24 @@ void writeBack(const Mag& mag, const RcDesc &rc, GRunArgP &g_arg, bool is_umat)
...
+
+#if !defined(GAPI_STANDALONE)
+void createMat(const cv::GMatDesc desc, cv::Mat& mat)
+{
+    const auto type = desc.planar ? desc.depth : CV_MAKETYPE(desc.depth, desc.chan);
+    const auto size = desc.planar ? cv::Size{desc.size.width, desc.size.height*desc.chan}
+                                  : cv::gapi::own::to_ocv(desc.size);
+    mat.create(size, type);
+}
+#endif
```
dmatveev: When do we need to create a cv::Mat? (internally).
I believe the only place is output argument allocation?

modules/gapi/src/api/gproto.cpp
```diff
@@ -132,6 +132,50 @@ cv::GMetaArg cv::descr_of(const cv::GRunArgP &argp)
...
+    }
+}
+
+bool cv::can_describe(const GMetaArgs &metas, const GRunArgs &args)
+{
+    return std::equal(metas.begin(), metas.end(), args.begin(),
+                      [](const GMetaArg& meta, const GRunArg& arg) {
+                          return can_describe(meta, arg);
+                      });
+}
```
dmatveev: So what is the reason behind these functions?

rgarnov:
@smirnov-alexey @TolyaTalamanov @dbudniko please take a look

modules/gapi/src/api/gbackend.cpp
rgarnov: It's also used for intermediate mat allocation in cpu and ocl backends

modules/gapi/src/api/gproto.cpp
rgarnov: They are used into compiler/backends to check if input/output arguments of the computation match inferred meta. It was a `descr_of(RunArgs&) == metas` comparison before, but now we should use `canDescribe` (or match) for such check for mats

modules/gapi/test/gapi_desc_tests.cpp
```diff
@@ -236,4 +236,62 @@ TEST(GAPI_MetaDesc, Typed_Compile_MatchMetaType_Mixed)
     EXPECT_NO_THROW(cc.compile(desc1, desc2));
 }
 
+TEST(GAPI_MetaDesc, Compare_Planar)
+{
+    const auto desc0 = cv::GMatDesc{CV_8U,3,{32,32},false};
```
dbudniko: Tests use square images almost everywhere. May be it makes sense to introduce rectangles {64, 32} instead to avoid potential width/height swap errors?

modules/gapi/test/gapi_planar_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+    cv::Size uv_sz(in_sz.width / 2, in_sz.height / 2);
+
+    cv::Mat y_mat  = cv::Mat(in_sz, CV_8UC1, in_mat.data);
+    cv::Mat uv_mat = cv::Mat(uv_sz, CV_8UC2, in_mat.data + in_mat.step1() * in_sz.height);
+
+    cv::Mat out_mat     = cv::Mat::zeros(out_sz.height*3, out_sz.width, CV_8UC1);
+    cv::Mat out_mat_ocv = cv::Mat::zeros(out_sz.height*3, out_sz.width, CV_8UC1);
+
+    cv::GMat inY, inUV;
+    auto out = GResize3p3p::on(GNV12toRGBp::on(inY, inUV), out_sz, interp);
```
dbudniko: Should we try to check BGR CSC as well? Is it redundant?

rgarnov: This is not a comprehensive testing of preproc-like pipeline with any possible configurations, it just tests that such kind of pipelines are supported by the framework with metadata inferred and kernels launched correctly. I believe that BGR csc is redundant in this test.

modules/gapi/test/gapi_desc_tests.cpp
rgarnov: Will change some tests to operate on non-square images

modules/gapi/include/opencv2/gapi/gproto.hpp
rgarnov: Done

modules/gapi/include/opencv2/gapi/gmat.hpp
rgarnov: Done

modules/gapi/include/opencv2/gapi/gmat.hpp
rgarnov: Added description

modules/gapi/include/opencv2/gapi/gmat.hpp
rgarnov: Added "invalid" default values

modules/gapi/src/api/gbackend.cpp
dmatveev: Thanks

modules/gapi/src/api/gproto.cpp
```diff
@@ -132,6 +132,50 @@ cv::GMetaArg cv::descr_of(const cv::GRunArgP &argp)
...
+
+bool cv::can_describe(const GMetaArg& meta, const GRunArg& arg)
+{
+    switch (arg.index())
+    {
+#if !defined(GAPI_STANDALONE)
+    case GRunArg::index_of<cv::Mat>():               return util::holds_alternative<GMatDesc>(meta) &&
+                                                            util::get<GMatDesc>(meta).canDescribe(util::get<cv::Mat>(arg));
+    case GRunArg::index_of<cv::UMat>():              return meta == cv::GMetaArg(descr_of(util::get<cv::Mat>(arg)));
```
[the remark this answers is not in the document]
rgarnov: Thanks for this catch! Updated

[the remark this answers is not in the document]
dmatveev: Seems this path was not tested. Did you add a test for that?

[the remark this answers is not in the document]
dbudniko: We discussed it with @rgarnov already. It will be separate task to enable planar mats in fluid and OCL backends testing.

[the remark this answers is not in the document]
rgarnov: Added tests for UMat, own::Mat and own::Scalar

PR #14403, opened 2019-04-23
video(DISOpticalFlow): use OpenCL shared mem

### LINKED ISSUE #14372: No NVIDIA OpenCL for DISOpticalFlow

- OpenCV => 4.1.0
- Compiler => gcc 7.3.0

##### Detailed description

I ran optical flow detection in an NVIDIA OpenCL Docker container. With `FarnebackOpticalFlow`, the GPU version is 9 times faster than the CPU version.  With `DISOpticalFlow`, there is no difference.  It looks like it is due to the test for `ocl::Device::getDefault().isIntel()`:

https://github.com/opencv/opencv/blob/master/modules/video/src/dis_flow.cpp#L1371

Is there a reason why the OpenCL implementation of `DISOpticalFlow` would not work on NVIDIA?

##### Steps to reproduce

If you like, I can provide a sample program. However, with the current source, `DISOpticalFlow` will not use OpenCL on NVIDIA.

alalek:
resolves #14372

Added reduction implementation via OpenCL shared memory (compatible with OpenCL 1.2).

TODO:
- [x] investigate/fix OpenCL processing on AMD devices on Linux: amdgpu 18.50(2766.4) and 17.50(2527.3)

<cut/>

```
allow_multiple_commits=1

force_builders=Custom
buildworker:Custom=linux-5
test_opencl:Custom=ON

build_image:Linux AVX2=ubuntu:18.04
buildworker:Linux AVX2=linux-3
test_opencl:Linux AVX2=ON
```

### DIFF
--- a/modules/video/perf/opencl/perf_dis_optflow.cpp
+++ b/modules/video/perf/opencl/perf_dis_optflow.cpp
@@ -37,10 +37,11 @@ OCL_PERF_TEST_P(DenseOpticalFlow_DIS, perf,
 
     Ptr<DenseOpticalFlow> algo = DISOpticalFlow::create(preset);
 
-    OCL_TEST_CYCLE_N(10)
+    PERF_SAMPLE_BEGIN()
     {
         algo->calc(frame1, frame2, flow);
     }
+    PERF_SAMPLE_END()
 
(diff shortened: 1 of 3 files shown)


alalek:
:+1:

PR #14408, opened 2019-04-24
Fix ONNX importer deconvolution

l-bat:
**Merge with extra:** opencv/opencv_extra#601


dkurt:
@l-bat, May I ask you to generate a test with PyTorch?

PR #14289, opened 2019-04-10
Changed scale condition in convertPointsFromHomogeneous()

manasbedmutha98:
### This pullrequest changes

Changes the condition for scale in `convertPointsFromHomogeneous()`

resolves #14242

PR #14411, opened 2019-04-24
Fix for Homogenous precision #14242

### LINKED ISSUE #14242: Homogenous precision
Can we have an `inf` at https://github.com/opencv/opencv/blob/master/modules/calib3d/src/fundam.cpp#L970 if  `1e-39>=sptr[i].z<=1e-45`?

ostarling:
resolves #14242

### DIFF
--- a/modules/calib3d/src/fundam.cpp
+++ b/modules/calib3d/src/fundam.cpp
@@ -909,6 +909,10 @@ void cv::computeCorrespondEpilines( InputArray _points, int whichImage,
     }
 }
 
+static inline float scaleFor(float x){
+    return (std::fabs(x) > std::numeric_limits<float>::epsilon()) ? 1.f/x : 1.f;
+}
+
 void cv::convertPointsFromHomogeneous( InputArray _src, OutputArray _dst )
 {
     CV_INSTRUMENT_REGION();
@@ -967,7 +971,7 @@ void cv::convertPointsFromHomogeneous( InputArray _src, OutputArray _dst )
             Point2f* dptr = dst.ptr<Point2f>();
             for( i = 0; i < npoints; i++ )
             {
-                float scale = sptr[i].z != 0.f ? 1.f/sptr[i].z : 1.f;
+                float scale = scaleFor(sptr[i].z);
                 dptr[i] = Point2f(sptr[i].x*scale, sptr[i].y*scale);
             }
         }
@@ -977,7 +981,7 @@ void cv::convertPointsFromHomogeneous( InputArray _src, OutputArray _dst )
             Point3f* dptr = dst.ptr<Point3f>();
             for( i = 0; i < npoints; i++ )
             {
-                float scale = sptr[i][3] != 0.f ? 1.f/sptr[i][3] : 1.f;
+                float scale = scaleFor(sptr[i][3]);
                 dptr[i] = Point3f(sptr[i][0]*scale, sptr[i][1]*scale, sptr[i][2]*scale);
             }
         }
@@ -990,7 +994,7 @@ void cv::convertPointsFromHomogeneous( InputArray _src, OutputArray _dst )
             Point2d* dptr = dst.ptr<Point2d>();
             for( i = 0; i < npoints; i++ )
             {
-                double scale = sptr[i].z != 0. ? 1./sptr[i].z : 1.;
+                double scale = scaleFor(sptr[i].z);
                 dptr[i] = Point2d(sptr[i].x*scale, sptr[i].y*scale);
             }
         }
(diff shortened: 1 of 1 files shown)


modules/calib3d/src/fundam.cpp
```diff
@@ -944,7 +944,7 @@ void cv::convertPointsFromHomogeneous( InputArray _src, OutputArray _dst )
             Point2f* dptr = dst.ptr<Point2f>();
             for( i = 0; i < npoints; i++ )
             {
-                float scale = sptr[i].z != 0 ? 1.f/sptr[i].z : 1.f;
+                float scale = (std::fabs(sptr[i].z) > FLT_EPSILON) ? 1.f/sptr[i].z : 1.f;
```
csukuangfj: I would suggest replacing
`FLT_EPSILON`
with
`std::numeric_limits<float>::epsilon()`.

csukuangfj:
duplicate #14289

ostarling:
Sure, I’ll do that. Do you have anything against turning these 6 or so similar changes to an inline function?

modules/calib3d/src/fundam.cpp
ostarling: Changed as suggested

modules/calib3d/src/fundam.cpp
```diff
@@ -944,7 +948,7 @@ void cv::convertPointsFromHomogeneous( InputArray _src, OutputArray _dst )
             Point2f* dptr = dst.ptr<Point2f>();
             for( i = 0; i < npoints; i++ )
             {
-                float scale = sptr[i].z != 0 ? 1.f/sptr[i].z : 1.f;
```
mshabunin: `sptr[i]` is `int` here, we can compare it with zero safely.

modules/calib3d/src/fundam.cpp
```diff
@@ -954,7 +958,7 @@ void cv::convertPointsFromHomogeneous( InputArray _src, OutputArray _dst )
             Point3f* dptr = dst.ptr<Point3f>();
             for( i = 0; i < npoints; i++ )
             {
-                float scale = sptr[i][3] != 0 ? 1.f/sptr[i][3] : 1.f;
```
mshabunin: `sptr[i]` is `int` here, we can compare it with zero safely.

modules/calib3d/src/fundam.cpp
thangktran: @mshabunin i believe `sptr[i]` is `float` from here
https://github.com/opencv/opencv/blob/master/modules/calib3d/src/fundam.cpp#L966

ostarling: Oh, yes, sorry. Will revert this and the other similar line back.

modules/calib3d/src/fundam.cpp
```diff
@@ -909,6 +909,10 @@ void cv::computeCorrespondEpilines( InputArray _points, int whichImage,
     }
 }
 
+inline float scaleFor(float x){
```
alalek: "static" should be used for local functions.

modules/calib3d/src/fundam.cpp
ostarling: reverted mistakenly changed lines where the value type was integer

modules/calib3d/src/fundam.cpp
ostarling: Made the function static

modules/calib3d/src/fundam.cpp
```diff
@@ -1000,7 +1004,7 @@ void cv::convertPointsFromHomogeneous( InputArray _src, OutputArray _dst )
             Point3d* dptr = dst.ptr<Point3d>();
             for( i = 0; i < npoints; i++ )
             {
-                double scale = sptr[i][3] != 0.f ? 1./sptr[i][3] : 1.;
+                double scale = scaleFor(sptr[i][3]);
```
thangktran: build failed due to conversion from `double` (`sptr[i][3]`) to `float` (parameter of `scaleFor()`).

ostarling: added an overloaded version of scaleFor for double

modules/calib3d/src/fundam.cpp
```diff
@@ -909,6 +909,14 @@ void cv::computeCorrespondEpilines( InputArray _points, int whichImage,
     }
 }
 
+static inline double scaleFor(double x){
+    return (std::fabs(x) > std::numeric_limits<double>::epsilon()) ? 1./x : 1.;
```
alalek: Double computations are usually truncated on float epsilon value too.
These "double" results are used for intermediate calculations only (for better precision).
But finally, they are converted into floats in functions which works with images (like `wapr()`).

ostarling: changed to <float>::epsilon()

bhack:
@alalek Sorry for cross-repository. What do you think about related issue in this PR using directly `1` as a scale factor. See https://github.com/arraiyopensource/kornia/pull/148#discussion_r288283863

bhack:
@alalek Have we a problem also in `cvProjectPoints2Internal`? https://github.com/opencv/opencv/blob/master/modules/calib3d/src/calibration.cpp#L786-L787
How are you clipping not visible (in the camera plane) points?

alalek:
Probably this code should be changed too.

bhack:
The new issue is at https://github.com/opencv/opencv/issues/14703

PR #14429, opened 2019-04-26
G-API fluid fix small rois

rgarnov:
* Fixed incorrect first iteration window setting for fluid FilterAgent: previously it was unconditionally set to Filter's lpi value which lead to incorrect work on small images/rois

```
allow_multiple_commits=1
```


rgarnov:
@smirnov-alexey please take a look

smirnov-alexey:
It's too late already :), yet it looks nice to me.

Issue #5206, opened 2015-08-16
Is it possible to use LMSolver  ? I f no can I change module source code of opencv_calib3d?


LaurentBerger:
In this post (http://answers.opencv.org/question/66169/how-can-i-use-lmsolver-to-solve-a-non-linear-system/) I try to use LMsolver to solve a non-linear system using Levenberg-Marquardt method. I don't find how to do and there is no answer. Withou any answer I think that's not possible to use it.

I have found my own answer in making a copy of class LMSolver in my own code and call method createLMSolver and then it works.

I propose then to change some file in module opencv_calib3d  to have a direct access to this class and to insert a small example how to use this class.

Module calib3d is not at all my module ((some authors are @CheeseWiz @taka-no-me @Dmitry-Me @Adil-Ibragimov @ilya-lavrenovilya-lavrenov @SpecLad) so I prefer asking if this job is usefull and if I can do it before starting work.

Waiting for your answer

Regards L. Berger

PhilLab:
@mshabunin: the label "question" does not fit this issue well. It is indeed phrased like a question but it is more a feature request or a "request permission to implement"
@LaurentBerger to discuss the change efficiently, you could **a)** Rephrase this issue's title and text to not get it mistaken as question (and therefore largely ignored) **and b)** shortly summarize your idea **and c)** trying to identify the author of the module and notify them by mentioning their Github user name

LaurentBerger:
@PhilLab As my english is not good at all I understand yours remarks. I have append some module's authors. About title I think that it is not too bad because may be someone know how to use LMSolver without changing source code. I have tried to improve it.

paroj:
it is probably not yet exported it lacks the possibility to fix parameters when compared to `CvLevMarq` - you could use the latter in the meantime though..

PR #14431, opened 2019-04-26
Add solvePnPRefineLM and solvePnPRefineVVS

catree:
related #14181 

This is a draft to see how adding a separate function to refine a pose after `solvePnP` would look:

- `solvePnPRefineLM` the classical Levenberg-Marquardt iterative minimization process
- `solvePnPRefineVVS` minimize the projection error using a virtual visual servoing scheme

---

`LMSolver` should be compared to the C implementation code for the Levenberg-Marquardt:
https://github.com/opencv/opencv/blob/7b8ce63730430edeaf26790020dbe23b6a8ddf81/modules/calib3d/src/calibration.cpp#L1221-L1248

Also, I have added the possibility to set the epsilon in the LM implementation.

### DIFF
--- a/modules/calib3d/src/solvepnp.cpp
+++ b/modules/calib3d/src/solvepnp.cpp
@@ -456,4 +456,271 @@ int solveP3P( InputArray _opoints, InputArray _ipoints,
     return solutions;
 }
 
+class SolvePnPRefineLMCallback CV_FINAL : public LMSolver::Callback
+{
+public:
+    SolvePnPRefineLMCallback(InputArray _opoints, InputArray _ipoints, InputArray _cameraMatrix, InputArray _distCoeffs)
+    {
+        objectPoints = _opoints.getMat();
+        imagePoints = _ipoints.getMat();
+        npoints = std::max(objectPoints.checkVector(3, CV_32F), objectPoints.checkVector(3, CV_64F));
+        imagePoints0 = imagePoints.reshape(1, npoints*2);
+        cameraMatrix = _cameraMatrix.getMat();
+        distCoeffs = _distCoeffs.getMat();
+    }
+
+    bool compute(InputArray _param, OutputArray _err, OutputArray _Jac) const CV_OVERRIDE
+    {
+         Mat param = _param.getMat();
+         _err.create(npoints*2, 1, CV_64FC1);
+
+         if(_Jac.needed())
+         {
+             _Jac.create(npoints*2, param.rows, CV_64FC1);
+         }
+
+         Mat rvec = param(Rect(0, 0, 1, 3)), tvec = param(Rect(0, 3, 1, 3));
+
+         Mat J, projectedPts;
+         projectPoints(objectPoints, rvec, tvec, cameraMatrix, distCoeffs, projectedPts, _Jac.needed() ? J : noArray());
+
+         if (_Jac.needed())
+         {
+             Mat Jac = _Jac.getMat();
+             for (int i = 0; i < Jac.rows; i++)
+             {
+                 for (int j = 0; j < Jac.cols; j++)
+                 {
+                     Jac.at<double>(i,j) = J.at<double>(i,j);
+                 }
+             }
+         }
+
+         Mat err = _err.getMat();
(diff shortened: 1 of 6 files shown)


modules/calib3d/src/solvepnp.cpp
```diff
@@ -456,4 +456,249 @@ int solveP3P( InputArray _opoints, InputArray _ipoints,
...
+                           TermCriteria _criteria=TermCriteria(TermCriteria::EPS+TermCriteria::COUNT, 20, FLT_EPSILON),
+                           double _vvslambda=1)
+{
+    CV_INSTRUMENT_REGION();
+
+    Mat opoints_ = _objectPoints.getMat(), ipoints_ = _imagePoints.getMat();
+    Mat opoints, ipoints;
+    opoints_.convertTo(opoints, CV_64F);
+    ipoints_.convertTo(ipoints, CV_64F);
+    int npoints = std::max(opoints.checkVector(3, CV_32F), opoints.checkVector(3, CV_64F));
```
alalek: After `opoints_.convertTo(opoints, CV_64F);` this check should always fail: `opoints.checkVector(3, CV_32F)`

modules/calib3d/src/solvepnp.cpp
```diff
@@ -456,4 +456,249 @@ int solveP3P( InputArray _opoints, InputArray _ipoints,
...
+    double wz = twist.at<double>(5,0);
+
+    Mat rvec = (Mat_<double>(3,1) << wx, wy, wz);
+    Mat R;
+    Rodrigues(rvec, R);
+
+    double theta = sqrt(wx*wx + wy*wy + wz*wz);
+    double sinc = abs(theta) < 1e-8 ? 1 : sin(theta) / theta;
+    double mcosc = (abs(theta) < 1e-8) ? 0.5 : (1-cos(theta)) / (theta*theta);
+    double msinc = (abs(theta) < 1e-8) ? (1/6.0) : (1-sinc) / (theta*theta);
```
alalek: Consider using of `std::fabs()` instead of `abs()`

modules/calib3d/src/solvepnp.cpp
```diff
@@ -456,4 +456,249 @@ int solveP3P( InputArray _opoints, InputArray _ipoints,
...
+{
+    //see Exponential Map in http://ethaneade.com/lie.pdf
+    double vx = twist.at<double>(0,0);
+    double vy = twist.at<double>(1,0);
+    double vz = twist.at<double>(2,0);
+    double wx = twist.at<double>(3,0);
+    double wy = twist.at<double>(4,0);
+    double wz = twist.at<double>(5,0);
+
+    Mat rvec = (Mat_<double>(3,1) << wx, wy, wz);
```
alalek: Matx31d rvec(wx, wy, wz);

catree:
@alalek 
Yes, there are two Levenberg-Marquardt implementations:
- one from the C-API and `SOLVEPNP_ITERATIVE` uses this one:
https://github.com/opencv/opencv/blob/faca45a7ea4db484244e6f177788fb404d0fa259/modules/calib3d/src/compat_ptsetreg.cpp#L53

- another public one that I have used for `solvePnPRefineLM`:
https://github.com/opencv/opencv/blob/faca45a7ea4db484244e6f177788fb404d0fa259/modules/calib3d/src/levmarq.cpp#L80

In the future, it would be good to drop the C-version and use `LMSolver` instead if:
- they provide the same results
- it is a one-to-one replacement (same features available) or implement the missing features

Related #5206

catree:
For me `solvePnPRefineLM` and `solvePnPRefineVVS` are fine but I am wondering if `solve` prefix could be misleading (these functions need an initial solution and are not a combination of `solvePnP` followed by a refinement method) for some users?
Otherwise `refinePnPLM` / `refinePnPWithLM` and `refinePnPVVS` / `refinePnPWithVVS` maybe?
With the class approach counterpart, `PnPSolver` and `PnPRefiner` are non-ambiguous and clear for me.

tobycollins:
I think the class approach is the good way to go. PnPSolver and PnPRefiner are clear names in my opinion.

PR #14448, opened 2019-04-29
G-API GMatP

rgarnov:
* Added GMatP type
* Changed planar tests to operate on GMatP

### DIFF
--- a/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
@@ -124,6 +124,10 @@ template<> struct get_in<cv::GMat>
 {
     static cv::Mat    get(GCPUContext &ctx, int idx) { return to_ocv(ctx.inMat(idx)); }
 };
+template<> struct get_in<cv::GMatP>
+{
+    static cv::Mat    get(GCPUContext &ctx, int idx) { return get_in<cv::GMat>::get(ctx, idx); }
+};
 template<> struct get_in<cv::GScalar>
 {
     static cv::Scalar get(GCPUContext &ctx, int idx) { return to_ocv(ctx.inVal(idx)); }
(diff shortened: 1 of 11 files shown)


modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -188,6 +192,14 @@ template<> struct get_out<cv::GMat>
         return {r};
     }
 };
+template<> struct get_out<cv::GMatP>
+{
+    static tracked_cv_mat get(GCPUContext &ctx, int idx)
+    {
+        auto& r = ctx.outMatR(idx);
+        return {r};
```
dmatveev: y not oneliner?

modules/gapi/include/opencv2/gapi/gmat.hpp
```diff
@@ -46,6 +46,19 @@ class GAPI_EXPORTS GMat
...
+public:
+    GMatP();                                 // Empty constructor
+    GMatP(const GNode &n, std::size_t out);  // Operation result constructor
+
+    GOrigin& priv();                        // Internal use only
+    const GOrigin& priv()  const;           // Internal use only
+
+private:
+    std::shared_ptr<GOrigin> m_priv;
+};
```
dmatveev: Could this be just subclassed from regular `cv::GMat` (with all fields reused)}

modules/gapi/include/opencv2/gapi/gtype_traits.hpp
```diff
@@ -47,6 +48,11 @@ namespace detail
         static constexpr const ArgKind kind = ArgKind::GMAT;
         static constexpr const GShape shape = GShape::GMAT;
     };
+    template<>           struct GTypeTraits<cv::GMatP>
+    {
+        static constexpr const ArgKind kind = ArgKind::GMATP;
```
dmatveev: Why does it need a special Kind enum?

rgarnov: It's required by GArg->GProtoArg conversion

modules/gapi/include/opencv2/gapi/gmat.hpp
rgarnov: Yes, will fix

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
rgarnov: Reworked

rgarnov:
The build is red due to GProtoArg variant extension with new type

PR #14452, opened 2019-04-29
imgproc: fix RemoveOverlaps() with empty input vector

### LINKED ISSUE #14432: HoughCircles sometimes returns [0,0,0]
- OpenCV => 4.0.1 (precompiled binary distributed by homebrew)
- Operating System / Platform => macOS Mojave 10.14.4
- Compiler => clang 10.0.0

##### Detailed description
When running HoughCircles on a series of images, it usually works well, but sometimes it will return a single value of [0,0,0] instead of the actual result. I receive an empty vector back on images where no circles are found, so I suspect that a circle was indeed detected, but there was an internal bug.  Unless there is an (undocumented) reason, HoughCircles should return the actual result instead of [0,0,0].

I am calling HoughCircles with a 480x640 CV_8UC1 image (either generated or loaded from a PNG using IMREAD_GRAYSCALE).  Full source code is attached, but the call to HoughCircles is as follows:
```HoughCircles(img, results, HOUGH_GRADIENT, 2, 5, 185, 100, 20, 100);```

The same issue is reported here:
https://stackoverflow.com/questions/53001840/opencv-houghcircles-occasionally-returning-0-0-0

This one seems similar also, except that it returns [10,0,0,0]:
https://stackoverflow.com/questions/49675692/opencv-houghcircles-returns-invalid-circle-parameters

##### Steps to reproduce

1. Download [hough_bug.cpp.txt](https://github.com/opencv/opencv/files/3129029/hough_bug.cpp.txt)
2. Rename the file to hough_bug.cpp
3. Compile hough_bug.cpp against OpenCV 4.0.1 to create a binary named `hough_bug`.
4. Run `./hough_bug` to generate a series of 1000 images, 26 of which demonstrate the issue (for me).  The random seed is fixed so the program will create the same series of images each time you run it.
5. Run `./hough_bug *.png` to test the images generated in the previous step. Again, 26 of the images will demonstrate the issue.  This proves that circle() and GaussianBlur() are not causing the issue.

The console output from an example run is attached:
[hough_bug_output.txt](https://github.com/opencv/opencv/files/3129000/hough_bug_output.txt)

This image demonstrates the bug - returning [0,0,0] instead of a list of valid circles.
000-bug

HoughCircles returns an empty list for this image, since no circles were found (this is expected with these parameters).
001-normal

HoughCircles returns a list of 2 circles that were detected in this image.
002-found-two-circles

alalek:
resolves #14432
relates #10232

### DIFF
--- a/modules/imgproc/src/hough.cpp
+++ b/modules/imgproc/src/hough.cpp
@@ -1355,6 +1355,8 @@ static void GetCircleCenters(const std::vector<int> &centers, std::vector<Vec4f>
 template<typename T>
 static void RemoveOverlaps(std::vector<T>& circles, float minDist)
 {
+    if (circles.size() <= 1u)
+        return;
     float minDist2 = minDist * minDist;
     size_t endIdx = 1;
     for (size_t i = 1; i < circles.size(); ++i)


alalek:
:+1:

Issue #14367, opened 2019-04-19
intersectionConvexConvex works incorrectly


aleksartamonov:
##### System information (version)
- OpenCV => 4.0
- Operating System / Platform => Mac
- Python 2.7.15 |Anaconda, Inc.| (default, May  1 2018, 18:37:05)
- IPython 5.8.0

##### Detailed description

If contour with points out of frame has 0 coordinates,  intersectConvexConvex does not works

##### Steps to reproduce

```
In [30]: frame
Out[30]:
array([[  0,   0],
       [740,   0],
       [740, 540],
       [  0, 540]])

In [31]: contour = np.array([[0,210],[-30,210],[-37,170],[ -7,172]])

In [32]: sq , chain  = cv2.intersectConvexConvex(frame, contour)

In [33]: chain
Out[33]:
array([[[  0., 210.]],

       [[-30., 210.]],

       [[-37., 170.]],

       [[ -7., 172.]]], dtype=float32)

In [34]: sq
Out[34]: 1163.0
```

thangktran:
I'm working on this issue.

alalek:
@aleksartamonov Could you try this patch: #14463 ?

thangktran:
@alalek how should we proceed with this issue?

PR #14463, opened 2019-04-30
imgproc: fixed intersectConvexConvex

thangktran:
Added checks for all of vertices from each contour instead of checking
only for the first vertex.

Fixes: https://github.com/opencv/opencv/issues/14367

- [x] built local successfully
- [x] `make test` successfully
- [x] added test

### DIFF
--- a/modules/imgproc/test/test_intersectconvexconvex.cpp
+++ b/modules/imgproc/test/test_intersectconvexconvex.cpp
@@ -1,0 +1,260 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include "test_precomp.hpp"
+
+namespace opencv_test { namespace {
+
+
+TEST(Imgproc_IntersectConvexConvex, no_intersection)
+{
+    std::vector<cv::Point> convex1;
+    convex1.push_back(cv::Point(290, 126));
+    convex1.push_back(cv::Point(284, 132));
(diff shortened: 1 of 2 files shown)


modules/imgproc/test/test_intersectconvexconvex.cpp
```diff
@@ -0,0 +1,301 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
```
alalek: Please add [short license header](https://github.com/opencv/opencv/wiki/Coding_Style_Guide#file-structure).

You may add authors after that (however authors are already in git history):
```
//
// @Authors
//      Thang Tran <TranKimThang279@gmail.com>
```

modules/imgproc/test/test_intersectconvexconvex.cpp
```diff
@@ -0,0 +1,301 @@
...
+    std::vector<cv::Point> convex2;
+    convex2.push_back(cv::Point(15, 0));
+    convex2.push_back(cv::Point(740, 0));
+    convex2.push_back(cv::Point(740, 540));
+    convex2.push_back(cv::Point(15, 540));
+
+    std::vector<cv::Point> intersection;
+    float area = cv::intersectConvexConvex(convex1, convex2, intersection);
+
+    std::vector<cv::Point> intersection_should;
```
alalek: `expected_intersection` ?

thangktran:
@alalek thank you for input.

PR #14491, opened 2019-05-06
Added support for clang build hardening

mshabunin:
### This pullrequest changes

Adds hardening support for clang and Apple clang builds.

**Note:** linker in XCode does not support `-z` option

### DIFF
--- a/cmake/OpenCVCompilerDefenses.cmake
+++ b/cmake/OpenCVCompilerDefenses.cmake
@@ -5,7 +5,7 @@
 # - https://wiki.debian.org/Hardening
 # - https://wiki.gentoo.org/wiki/Hardened/Toolchain
 # - https://docs.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks
-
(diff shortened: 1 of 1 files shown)


cmake/OpenCVCompilerDefenses.cmake
```diff
@@ -44,6 +44,12 @@ if(MSVC)
   if(NOT X86_64)
     set(OPENCV_LINKER_DEFENSES_FLAGS_COMMON "${OPENCV_LINKER_DEFENSES_FLAGS_COMMON} /safeseh")
   endif()
+elseif(CV_CLANG)
+  ocv_add_defense_compiler_flag("-fstack-protector-strong")
+  ocv_add_defense_compiler_flag_release("-D_FORTIFY_SOURCE=2")
```
alalek: Please add some relevant links with information into file header, like this:
https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Articles/BufferOverflows.html

PR #14524, opened 2019-05-09
Add support for loading ONNX model from in-memory buffer.

kohei-us:
This change allows ONNX model blob already loaded in-memory to be passed to the ONNX importer from which to initialize the cv::dnn::Net instance.  It accomplishes this by wrapping the in-memory buffer with std::streambuf interface which std::istream requires in order to read data from it.  Hopefully this functionality would be generally useful.

I'm open to renaming the top-level function to something more appropriate if the name I've given is not consistent with any sort of naming convention.

Thanks,

Kohei

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -868,6 +868,23 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
      */
     CV_EXPORTS_W Net readNetFromONNX(const String &onnxFile);
 
+    /** @brief Reads a network model from <a href="https://onnx.ai/">ONNX</a>
+     *         in-memory buffer.
+     *  @param buffer memory address of the first byte of the buffer.
+     *  @param sizeBuffer size of the buffer.
+     *  @returns Network object that ready to do forward, throw an exception
+     *        in failure cases.
+     */
+    CV_EXPORTS Net readNetFromONNX(const char* buffer, size_t sizeBuffer);
+
+    /** @brief Reads a network model from <a href="https://onnx.ai/">ONNX</a>
+     *         in-memory buffer.
+     *  @param buffer in-memory buffer that stores the ONNX model bytes.
+     *  @returns Network object that ready to do forward, throw an exception
+     *        in failure cases.
+     */
+    CV_EXPORTS_W Net readNetFromONNX(const std::vector<uchar>& buffer);
+
     /** @brief Creates blob from .pb file.
      *  @param path to the .pb file with input tensor.
      *  @returns Mat.
--- a/modules/dnn/src/onnx/onnx_importer.cpp
+++ b/modules/dnn/src/onnx/onnx_importer.cpp
@@ -57,6 +57,24 @@ public:
             CV_Error(Error::StsUnsupportedFormat, "Failed to parse onnx model");
     }
 
+    ONNXImporter(const char* buffer, size_t sizeBuffer)
+    {
+        struct _Buf : public std::streambuf
+        {
+            _Buf(const char* buffer, size_t sizeBuffer)
+            {
+                char* p = const_cast<char*>(buffer);
+                setg(p, p, p + sizeBuffer);
+            }
+        };
+
+        _Buf buf(buffer, sizeBuffer);
+        std::istream input(&buf);
+
+        if (!model_proto.ParseFromIstream(&input))
+            CV_Error(Error::StsUnsupportedFormat, "Failed to parse onnx model from in-memory byte array.");
+    }
+
     void populateNet(Net dstNet);
 };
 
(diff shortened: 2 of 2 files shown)


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -864,6 +864,15 @@ CV__DNN_INLINE_NS_BEGIN
...
     CV_EXPORTS_W Net readNetFromONNX(const String &onnxFile);
 
+    /** @brief Reads a network model <a href="https://onnx.ai/">ONNX</a> from
+     *         in-memory buffer.
+     *  @param buffer memory address of the first byte of the buffer.
+     *  @param sizeBuffer size of the buffer.
+     *  @returns Network object that ready to do forward, throw an exception
+     *        in failure cases.
+     */
+    CV_EXPORTS_W Net readNetFromONNXBytes(const char* buffer, size_t sizeBuffer);
```
dkurt: This method cannot be wrapped because there is no way to bind `const char*` buffer in different languages. So please replace to `CV_EXPORTS`. And keep the origin name `readNetFromONNX`.

For bindings, please define
```cpp
CV_EXPORTS_W Net readNetFromONNX(const std::vector<uchar>& buffer);
```

See Caffe importer for reference.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -15,6 +15,7 @@
 #include <string>
 #include <limits>
 #include <algorithm>
+#include <strstream>
```
dkurt: `<sstream>` ?

kohei-us: Yup.  I've dropped that line and verified that it still builds.

modules/dnn/include/opencv2/dnn/dnn.hpp
kohei-us: Thanks.  I've switched to using CV_EXPORTS, and renamed the function to drop the trailing "Bytes".

Unfortunately I cannot use std::vector<uchar> for my use case, since our ONNX model bytes are either embedded to the program binary, or memory-mapped in at run-time.  I could copy the bytes to std::vector<uchar>, but the point of this change is to avoid unnecessary memory-copy...

dkurt: @kohei-us, I mean you can define both methods with `const char*` and `std::vector<uchar>`. The second one might be used in Python and Java bindings.

dkurt: Or if you don't need Python and Java you can just keep `const char*` but with `CV_EXPORTS`, yes.

kohei-us: @dkurt Ah, I got it.  I've modified it to add an overloaded variant as you suggested.

alalek:

So, please:

kohei-us:
@alalek I've rebased my branch off of 3.4 and pushed it as you suggested.

kohei-us:
@alalek BTW, that git rebase command you shared above is awesome!  I've learned something new today. :-)

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -868,6 +868,23 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
      */
     CV_EXPORTS_W Net readNetFromONNX(const String &onnxFile);
 
+    /** @brief Reads a network model <a href="https://onnx.ai/">ONNX</a> from
```
dkurt: There is a wrong order: "<a href="https://onnx.ai/">ONNX</a> from" but not "from <a href="https://onnx.ai/">ONNX</a>"

kohei-us: @dkurt Good catch. Fixed.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -808,6 +826,19 @@ Net readNetFromONNX(const String& onnxFile)
...
+{
+    ONNXImporter onnxImporter(buffer, sizeBuffer);
+    Net net;
+    onnxImporter.populateNet(net);
+    return net;
+}
+
+Net readNetFromONNX(const std::vector<uchar>& buffer)
+{
+    return readNetFromONNX(reinterpret_cast<const char*>(buffer.data()), buffer.size());
```
dkurt: `std::vector::data` is a C++11 feature. @alalek, Do you have an idea why the tests are passed?

alalek: GCC uses GNU extensions by default (`-std=gnu++98`).
MSVS2015+ STL has `.data()` method too.

ANSI C++98 compilation mode is not tested at this moment (and may not work, at least in protobuf codes).

---

We already [using this method in OpenCV](https://github.com/opencv/opencv/blob/3.4.6/modules/core/src/matrix.cpp#L377-L380).

dkurt: Ok, I see, thank you! So I think we can merge this PR if you have no extra comments.

PR #14537, opened 2019-05-13
Fix dnn visualizer

l-bat:


modules/dnn/src/dnn.cpp
```diff
@@ -2980,6 +2980,18 @@ int Net::getLayerId(const String &layer)
...
 
+String parseLayerParams(const String& name, const LayerParams& lp) {
+    DictValue param = lp.get(name);
+    std::ostringstream out;
+    out << name << " ";
+    switch (param.size()) {
+        case 1: out << ": " << param << "\\l"; break;
+        case 2: out << "(HxW): " << param.get<int>(0) << " x " << param.get<int>(1) << "\\l"; break;
+        case 3: out << "(DxHxW): " << param.get<int>(0) << " x " << param.get<int>(1) << " x " << param.get<int>(2) << "\\l"; break;
+    }
```
alalek: `default:` ?

Lets don't try to process silently of unsupported cases.

dkurt: Maybe we need to use switch only for prefix but all the values print in the loop over `param.size()`?

modules/dnn/src/dnn.cpp
```diff
@@ -3066,38 +3083,45 @@ String Net::dump()
...
                  } else if (lp.has("dilation_h") && lp.has("dilation_w")) {
                      DictValue h = lp.get("dilation_h");
                      DictValue w = lp.get("dilation_w");
                      out << "dilation (HxW): " << h << " x " << w << "\\l";
                  }
                  if (lp.has("pad")) {
                      DictValue pad = lp.get("pad");
-                     out << "pad (LxTxRxB): " << pad << " x " << pad << " x " << pad << " x " << pad << "\\l";
+                     out << "pad ";
+                     switch (pad.size()) {
```
alalek: Unsupported case is ignored here too.

PR #14516, opened 2019-05-08
Async mode for dnn's object detection sample

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/13667

Camera with 60fps @ 1280x720 is used (Logitech C922 PRO)

| Target | `--async=0` | `--async=1` | `--async=2` | `--async=3` | `--async=4` |
|------|-----|-----|-----|----|---|
| NCS, USB 3.0 | 11.6 | 9.9 | 11.9 | 11.9 | 11.9 |
| NCS2, USB 3.0 | 26.43 | 19.9 | 39.5 | **53.2** | 48.6 |

CPU: Intel® Core™ i7-6700K CPU @ 4.00GHz × 8
NCS: Intel® Movidius™ Neural Compute Stick
NCS2: Intel® Neural Compute Stick 2

* All the numbers are frames per second which are processed by network (FPS):

```
$ python object_detection.py opencv_fd --target 3
```

```
allow_multiple_commits=1
buildworker:Custom=linux-2
docker_image:Custom=ubuntu-openvino:16.04
build_parallel_tests:Custom=1
test_modules:Custom=dnn
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
```

PR #14538, opened 2019-05-13
Asynchronous mode for C++ object detection sample

dkurt:
### This pullrequest changes

see https://github.com/opencv/opencv/pull/14516 for Python sample (will be ported into the master branch).

### DIFF
--- a/samples/dnn/object_detection.cpp
+++ b/samples/dnn/object_detection.cpp
@@ -1,5 +1,7 @@
 #include <fstream>
 #include <sstream>
+#include <thread>
+#include <queue>
 
 #include <opencv2/dnn.hpp>
 #include <opencv2/imgproc.hpp>
@@ -26,8 +28,9 @@ std::string keys =
                          "0: CPU target (by default), "
                          "1: OpenCL, "
                          "2: OpenCL fp16 (half-float precision), "
-                         "3: VPU }";
-
+                         "3: VPU }"
+    "{ async       | 0 | Number of asynchronous forwards at the same time. "
+                        "Choose 0 for synchronous mode. }";
 
 using namespace cv;
 using namespace dnn;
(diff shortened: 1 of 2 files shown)


samples/dnn/object_detection.cpp
```diff
@@ -104,44 +137,135 @@ int main(int argc, char** argv)
...
+    std::queue<Mat> processedFramesQueue;
+    QueueFPS<std::vector<Mat> > predictionsQueue;
+    std::thread processingThread([&](){
+        std::queue<std::future<Mat> > futureOutputs;
+        Mat blob;
+        while (process)
+        {
+            // Get a next frame
+            Mat frame;
+            framesQueueMutex.lock();
```
alalek: Use scope-based `std::shared_lock` or `cv::AutoLock` (with `cv::Mutex`).

samples/dnn/object_detection.cpp
```diff
@@ -104,44 +137,135 @@ int main(int argc, char** argv)
...
+    std::thread framesThread([&](){
+        Mat frame;
+        while (process)
         {
-            waitKey();
-            break;
+            cap >> frame;
+            if (!frame.empty())
+            {
+                framesQueueMutex.lock();
```
alalek: Use scope-based `std::shared_lock` or `cv::AutoLock` (with `cv::Mutex`).

samples/dnn/object_detection.cpp
```diff
@@ -104,44 +137,135 @@ int main(int argc, char** argv)
...
-    // Process frames.
-    Mat frame, blob;
-    while (waitKey(1) < 0)
-    {
-        cap >> frame;
-        if (frame.empty())
+    bool process = true;
+
+    // Frames capturing thread
+    std::mutex framesQueueMutex;
```
alalek: Mutex should go into `QueueFPS` class.

samples/dnn/object_detection.cpp
```diff
@@ -41,7 +44,36 @@ void drawPred(int classId, float conf, int left, int top, int right, int bottom,
...
+    QueueFPS() : counter(0) {}
+
+    void push(const T& entry)
+    {
+        std::queue<T>::push(entry);
+        counter += 1;
+        if (counter == 1)
+        {
+            tm.reset();
+            tm.start();
```
alalek: Add comment about skipping of "warmup" first frame.

samples/dnn/object_detection.cpp
dkurt: Replaced to `std::lock_guard`, thanks!

samples/dnn/object_detection.cpp
dkurt: Moved all the locks into the `QueueFPS`.

dkurt:
> Number of dropped frames?

Maybe percentage is even better.

dkurt:
> Number of dropped frames?

That's a good idea, thank you! Ratio of dropped might be computer by division of Camera FPS over Processed FPS but the number of skipped frames is useful to know.

alalek:
Please apply threading fixes from here: https://github.com/alalek/opencv/commit/pr14538_r

dkurt:
Merged to https://github.com/opencv/opencv/pull/14516.

@alalek, I'll apply the patch with threads now.

PR #14539, opened 2019-05-13
G-API Fixed fluid roi inference to always produce even sizes for nv12

rgarnov:
* Added an adjustment for roi produced by NV12toRGB to always be even

### DIFF
--- a/modules/gapi/src/backends/fluid/gfluidbackend.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidbackend.cpp
@@ -648,6 +648,7 @@ void cv::gimpl::GFluidExecutable::initBufferRois(std::vector<int>& readStarts,
 
                     cv::gapi::own::Rect produced = rois[m_id_map.at(data.rc)];
 
+                    // Apply resize-specific roi transformations
                     cv::gapi::own::Rect resized;
                     switch (fg.metadata(oh).get<FluidUnit>().k.m_kind)
                     {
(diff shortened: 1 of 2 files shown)


rgarnov:
@andrey-golubev please take a look

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -658,7 +658,22 @@ void cv::gimpl::GFluidExecutable::initBufferRois(std::vector<int>& readStarts,
...
 
                     int readStart = resized.y;
-                    cv::gapi::own::Rect roi = adjFilterRoi(resized, fd.border_size, in_meta.size.height);
+                    auto roi = adjFilterRoi(resized, fd.border_size, in_meta.size.height);
+
+                    // Extend required input roi to be even if it's produced by NV12toRGB
+                    if (!in_node->inNodes().empty()) {
+                        auto in_data_producer = in_node->inNodes().front();
+                        if (fg.metadata(in_data_producer).get<FluidUnit>().k.m_kind == GFluidKernel::Kind::NV12toRGB) {
+                            if (roi.y % 2 != 0) {
```
alalek: What should be even?
- base `.y`
- or `.height`

---

What is about `.width`?

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -658,7 +658,22 @@ void cv::gimpl::GFluidExecutable::initBufferRois(std::vector<int>& readStarts,
...
+                                roi.y--;
+                                roi.height++;
+                            }
+
+                            if (roi.height % 2 != 0) {
+                                roi.height++;
```
alalek: What is about a check with fitting into image?
Why is not to shrink `roi`?

modules/gapi/src/backends/fluid/gfluidbackend.cpp
rgarnov: Both .y and .height should be even, vertical slicing is not supported for now, so width is not taken into account

modules/gapi/src/backends/fluid/gfluidbackend.cpp
rgarnov: This is actually done later via adjFilterRoi call, I was thinking to clarify this step, will do

PR #14557, opened 2019-05-14
LAPACK: add support for complex numbers for MSVC

mshabunin:
### This pullrequest changes

MSVC does not support `float _Complex` type, instead it provides `_Fcomplex` and `_Dcomplex`. This patch fixes compilation with OpenBLAS. Previously we only used MKL on Windows.

```
force_builders=Custom Win
```

### DIFF
--- a/cmake/OpenCVFindLAPACK.cmake
+++ b/cmake/OpenCVFindLAPACK.cmake
@@ -31,26 +31,32 @@ macro(ocv_lapack_check)
   else()
     # adding proxy opencv_lapack.h header
     set(CBLAS_H_PROXY_PATH ${CMAKE_BINARY_DIR}/opencv_lapack.h)
-    if((APPLE OR OPENCV_SKIP_LAPACK_EXTERN_C) AND NOT OPENCV_FORCE_LAPACK_EXTERN_C)
-        set(_lapack_include_str_extern_C "")
-        set(_lapack_include_str_extern_C_end "")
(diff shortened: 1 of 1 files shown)


cmake/OpenCVFindLAPACK.cmake
```diff
@@ -38,7 +38,7 @@ macro(ocv_lapack_check)
         set(_lapack_include_str_extern_C "extern \"C\" {\n")
         set(_lapack_include_str_extern_C_end "}\n")
     endif()
-    set(_lapack_include_str "${_lapack_include_str_extern_C}\#include \"${OPENCV_CBLAS_H_PATH_${_lapack_impl}}\"")
+    set(_lapack_include_str "${_lapack_include_str_extern_C}\#ifdef _MSC_VER\n\#include <complex.h>\n\#define lapack_complex_float _Fcomplex\n\#define lapack_complex_double _Dcomplex\n\#endif\n\#include \"${OPENCV_CBLAS_H_PATH_${_lapack_impl}}\"")
```
alalek: Please split this line.
Escaping of `#` is not needed.
Example is [here](https://github.com/opencv/opencv/blob/3.4.6/cmake/OpenCVCompilerOptimizations.cmake#L785-L788).

---

It would be better to configure these defines via CMake with possibility of disabling them (if this would be fixed in OpenBLAS in the future).

PR #14558, opened 2019-05-14
cmake: enable threading in samples

alalek:
- moved code into separate file due issue with handling of CMake policies

For using of `#include <thread>` with `std::thread`.


alalek:
:+1:

PR #14561, opened 2019-05-14
calib3d: fix uninitialized fields in ctors

alalek:
```
>>>     CID 1444972:  Uninitialized members  (UNINIT_CTOR)
>>>     Non-static class member "epsf" is not initialized in this constructor nor in any functions that it calls.
83         LMSolverImpl() : maxIters(100) { init(); }
```


alalek:
:+1:

PR #14583, opened 2019-05-19
fix the bug, when k[4] is negative, icdist may be negative at the edg…

FanaticsKang:
### This pull request changes 
The `cvUndistortPointsInternal` function.
The k[4] is negative of my camera, the `icdist` will be negative at the edge of image. This cause the error like the point at right will be undistort to the left.
So the function will return when `icdist` < 0, and use the last compute result.


alalek:
Please capture your inputs and add some simple regression test for that case.

FanaticsKang:
Did you have a tutor for writing opencv test? I'm not familiar with your test style.

alalek:
@SachieKang Could you share some image/calibration parameters for regression test?

FanaticsKang:
I update a test, and #14800 is the test with original code.  You will see the error by running ./bin/opencv_test_imgproc.

PR #8883, opened 2017-06-10
Enable std::vector<T> support in java wrappers; refactor Java Wrapper Generator for better extensibility

abratchik:
resolves #8882
resolves #8906 
resolves #8908
 

### This pullrequest changes
/modules/java
- CMakeLists.txt (changed to support module-specific type dictionaries)

/modules/java/include
- java.hpp (added for storing common macro for type converters)

/modules/java/generator
- gen_java.py (refactored, simplified, added proper support of command line params, removed hardcode)
- src/cpp/common.h (modified, shared definitions moved to java.hpp)
- src/cpp/listconverters.hpp (added, converters for `std::vector<String>`)
- src/cpp/listconverters.cpp (added, converters for `std::vector<String>`)

/modules/calib3d/misc/java
- gen_dict.json (added, dictionaries for Calib3d module)

/modules/core/misc/java
- gen_dict.json (added, dictionaries for Core module)

/modules/highgui/misc/java
- gen_dict.json (added, dictionaries for Highgui module)

/modules/features2d/misc/java
 - gen_dict.json (added, dictionaries for Feature2d module)

/modules/imgproc/misc/java
 - gen_dict.json (added, dictionaries for Imgproc module)

/modules/video/misc/java
 - gen_dict.json (added, dictionaries for Video module)

/modules/videoio/misc/java
 - gen_dict.json (added, dictionaries for Videoio module)

Enabled external configuration of java wrapper dictionaries through <module>/misc/java/gen_dict.json file. This allows to define mappings between C/C++ and Java/JNI types, exclusions, manual functions etc. in each module thus eliminating the necessity to update gen_java.py for every new type required.  

Enabled support of `std::vector<T>` type in wrapped methods. Added convertors for `std::vector<cv::String>` type. These types are extensively used in dnn module.

#### Example/test case:
1. Enable java support for dnn module 
2. Build OpenCV with contrib and dnn enabled
3. Download  https://storage.googleapis.com/download.tensorflow.org/models/inception5h.zip. Unzip. 
4. Execute the following fragment:
```

        String modelFileName = "path/to/tensorflow_inception_graph.pb"; 
        Net net = Dnn.readNetFromTensorflow(modelFileName);
        System.out.println("Net loaded successfully");
        
        List<String> layertypes = new ArrayList();
        
        net.getLayerTypes(layertypes);
        
        for(int i=0; i<layertypes.size(); i++) {
            System.out.println("LayerType("+ i +") = " + layertypes.get(i));
        }
        
        
[...]

PR #14597, opened 2019-05-22
java: fix java missing_consts

utibenkei:
### This pullrequest changes

The values of CALIB_FIX_K1, CALIB_FIX_K2, CALIB_FIX_K3, and CALIB_FIX_K4 reset by gen_dict.json are incorrect. Correct values are output by removing unnecessary missing_consts blocks.

https://github.com/opencv/opencv/blob/master/modules/calib3d/include/opencv2/calib3d.hpp#L262-L265
https://github.com/opencv/opencv/blob/master/modules/calib3d/misc/java/gen_dict.json#L5-L19

### DIFF
--- a/modules/calib3d/misc/java/gen_dict.json
+++ b/modules/calib3d/misc/java/gen_dict.json
@@ -3,21 +3,6 @@
         "CirclesGridFinderParameters",
         "CirclesGridFinderParameters2"
     ],
-    "missing_consts" : {
-        "Calib3d": {
-            "public" : [
-                ["CALIB_USE_INTRINSIC_GUESS", "1"],
-                ["CALIB_RECOMPUTE_EXTRINSIC", "2"],
-                ["CALIB_CHECK_COND", "4"],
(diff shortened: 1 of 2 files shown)


alalek:
Constants overriding has been performed to avoid conflict with "same" constants from "fisheye" namespace. Bindings generator had/has issues with nested namespace handling.

- [calib3d constants](https://github.com/opencv/opencv/blob/4.1.0/modules/calib3d/include/opencv2/calib3d.hpp#L257-L267) (`cv` namespace)
- [fisheye constants](https://github.com/opencv/opencv/blob/4.1.0/modules/calib3d/include/opencv2/calib3d.hpp#L2743-L2746) (`cv::fisheye` namespace)

relates PR: #8883 (moved between files) #3249 (changed values)

---

Please add Java test for exact values of these constants (use values from the first set).
Try to remove them.
If removal fails then update values in bindings constants.

alalek:
Rebased onto [3.4 branch](https://github.com/opencv/opencv/wiki/Branches) and added Java test.

PR #14599, opened 2019-05-22
G-API aux kernels

rgarnov:
* Added `auxiliaryKernels` method of `GBackend::Priv`, intergated into compiler
* Also there is some small cleanup (gcpucore.hpp and gcpuimgproc.hpp are actually unused so were removed)
* Changed `GModel::orderedInputs` and `orderedOutputs` to get `ConstGraph&` instead of `Graph&`


modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -48,9 +48,17 @@ namespace
 {
     cv::gapi::GKernelPackage getKernelPackage(cv::GCompileArgs &args)
     {
+        auto withAuxKernels = [](const cv::gapi::GKernelPackage& pkg) {
+            cv::gapi::GKernelPackage aux_pkg;
+            for (const auto &b : pkg.backends()) {
+                aux_pkg = combine(aux_pkg, b.priv().auxiliaryKernels());
+            }
+            return combine(pkg, aux_pkg);
```
dmatveev: Looks like std::accumulate, you don't even need pass a lambda but can pass `&combine` directly.

dmatveev:
:+1:

PR #14600, opened 2019-05-22
Kernel using events from OpenCL.

avieira:
### This pullrequest changes

Currently in OpenCV, the available OpenCL integration still does not include several tools defined by the standard. Among these features missing, I needed a way to use events.
For one of my current project, I need several Queues and synchronization between these using events. Since I am also using the OpenCL Kernels available in OpenCV, I thought it could be a nice idea to implement this Event class and add a way to run kernels using these events.
Also, from my point of view concerning the current goals for OpenCV, I think that these events could be a good tool for future developments. 

All comments (especially nice ones 😄 ) are welcome!

### DIFF
--- a/modules/core/include/opencv2/core/ocl.hpp
+++ b/modules/core/include/opencv2/core/ocl.hpp
@@ -562,18 +562,36 @@ public:
         i = set(i, a6); i = set(i, a7); i = set(i, a8); i = set(i, a9); i = set(i, a10); i = set(i, a11);
         i = set(i, a12); i = set(i, a13); i = set(i, a14); set(i, a15); return *this;
     }
+
     /** @brief Run the OpenCL kernel.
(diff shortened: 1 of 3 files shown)


avieira:
Could someone help me with the build please? ^^"

catree:
@avieira 

In case you didn't see it, you can access the build logs by clicking on the red cross near the hash commit, or using [https://pullrequest.opencv.org/#/summary/opencv](https://pullrequest.opencv.org/#/summary/opencv).

avieira:
@catree I know, I just have no idea how to fix that! If someone could give me a clue...

asmorkalov:
Events support is very limited in current OpenCL implementations and there is no use-case for OpenCL events in library right now. Filed feature request https://github.com/opencv/opencv/issues/15933 with reference to patch to return to the PR later.

PR #14601, opened 2019-05-23
java: add CvType.CV_16F

### LINKED ISSUE #14211: OpenCV now supports the CV_16F type. However, CvType.CV_16F is missing from Java binding.
- OpenCV => 4.0
- Compiler => Eclipse

##### Detailed description

OpenCV now supports the CV_16F type. However, CvType.CV_16F is missing from Java binding.

asashour:
resolves #14211

### This pullrequest changes

- Adds CvType.CV_16F
- Deprecate `CV_USRTYPE1`, since it has the same value as `CV_16F`
- typeToString() is changed to return `CV_16F` instead of `CV_USRTYPE1`

### DIFF
--- a/modules/core/misc/java/src/java/core+CvType.java
+++ b/modules/core/misc/java/src/java/core+CvType.java
@@ -4,12 +4,20 @@ public final class CvType {
 
     // type depth constants
     public static final int
-            CV_8U = 0, CV_8S = 1,
-            CV_16U = 2, CV_16S = 3,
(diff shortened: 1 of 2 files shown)


asashour:
This change was done on `master`, but it should have been done against `3,4`.

Should I raise a new one against `3.4`?

alalek:
No, there is no `CV_16F` in `3.4`

PR #14606, opened 2019-05-23
java: generated code inline return

asashour:
### This pullrequest changes

The generated code should not have redundant `return;` statements, and whenever possible, directly return the created object.

Also, this change excludes all imports in `java.lang.` package.


alalek:
Some cleanup is needed:
- `Java_org_opencv_calib3d_StereoSGBM_setMode_10` and similar: `return;` is removed but lines with indentation spaces are left.

PR #14648, opened 2019-05-27
G-API Introduce GAPI_TRANSFORM initial interface

smirnov-alexey:
Description:

In utils.h were created 2 classes to separate kernels from transformations via "tag" in kernel package;
GAPI_TRANSFORM is implemented in a way similar to G_TYPED_KERNEL

```
```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
@@ -2,7 +2,7 @@
 // It is subject to the license terms in the LICENSE file found in the top-level directory
 // of this distribution and at http://opencv.org/license.html.
 //
-// Copyright (C) 2019 Intel Corporation
+// Copyright (C) 2018-2019 Intel Corporation
 
 
 #ifndef OPENCV_GAPI_GCPUKERNEL_HPP
--- a/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
@@ -2,7 +2,7 @@
 // It is subject to the license terms in the LICENSE file found in the top-level directory
 // of this distribution and at http://opencv.org/license.html.
 //
-// Copyright (C) 2018 Intel Corporation
+// Copyright (C) 2019 Intel Corporation
 
 
 #ifndef OPENCV_GAPI_GCPUKERNEL_HPP
--- a/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
@@ -19,6 +19,7 @@
 #include <opencv2/gapi/garg.hpp>
 #include <opencv2/gapi/own/convert.hpp> //to_ocv
 #include <opencv2/gapi/util/compiler_hints.hpp> //suppress_unused_warning
+#include <opencv2/gapi/util/util.hpp>
 
 // FIXME: namespace scheme for backends?
 namespace cv {
@@ -258,7 +259,8 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
 } // namespace detail
 
 template<class Impl, class K>
-class GCPUKernelImpl: public detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>
+class GCPUKernelImpl: public detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>,
+                      public cv::detail::KernelTag
 {
     using P = detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>;
 
--- a/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
@@ -259,7 +259,7 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
 } // namespace detail
 
 template<class Impl, class K>
-class GCPUKernelImpl: public detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>,
+class GCPUKernelImpl: public cv::detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>,
                       public cv::detail::KernelTag
 {
     using P = detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>;
--- a/modules/gapi/include/opencv2/gapi/gcommon.hpp
+++ b/modules/gapi/include/opencv2/gapi/gcommon.hpp
@@ -30,7 +30,7 @@ namespace detail
         static const char* tag() { return ""; };
     };
 
-    // These structures are tags which suppose to separate kernels and transformations
+    // These structures are tags which separate kernels and transformations
     struct KernelTag
     {};
     struct TransformTag
--- a/modules/gapi/include/opencv2/gapi/gcommon.hpp
+++ b/modules/gapi/include/opencv2/gapi/gcommon.hpp
@@ -29,6 +29,12 @@ namespace detail
     {
         static const char* tag() { return ""; };
     };
+
+    // These structures are tags which suppose to separate kernels and transformations
+    struct KernelTag
+    {};
+    struct TransformTag
+    {};
 }
 
 // This definition is here because it is reused by both public(?) and internal
--- a/modules/gapi/include/opencv2/gapi/gkernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/gkernel.hpp
@@ -2,7 +2,7 @@
 // It is subject to the license terms in the LICENSE file found in the top-level directory
 // of this distribution and at http://opencv.org/license.html.
 //
-// Copyright (C) 2019 Intel Corporation
+// Copyright (C) 2018-2019 Intel Corporation
 
 
 #ifndef OPENCV_GAPI_GKERNEL_HPP
@@ -302,7 +302,7 @@ namespace gapi {
      * implementation collections and graph transformations.
      *
      * GKernelPackage is a special container class which stores kernel
-     * _implementations_ and graph transformations. Objects of this class
+     * _implementations_ and graph _transformations_. Objects of this class
      * are created and passed to cv::GComputation::compile() to specify
      * which kernels to use and which transformations to apply in the
      * compiled graph. GKernelPackage may contain kernels of
@@ -347,8 +347,8 @@ namespace gapi {
 
         /// @private
         // Partial includes() specialization for kernels
-        template <typename KImpl, cv::detail::PackageObjectTag Val>
-        typename std::enable_if<(Val == cv::detail::PackageObjectTag::KERNEL), bool>::type
+        template <typename KImpl>
+        typename std::enable_if<(std::is_base_of<detail::KernelTag, KImpl>::value), bool>::type
         includesHelper() const {
             auto kernel_it = m_id_kernels.find(KImpl::API::id());
             return kernel_it != m_id_kernels.end() &&
@@ -357,8 +357,8 @@ namespace gapi {
 
         /// @private
         // Partial include() specialization for kernels
-        template <typename KImpl, cv::detail::PackageObjectTag Val>
-        typename std::enable_if<(Val == cv::detail::PackageObjectTag::KERNEL), void>::type
+        template <typename KImpl>
+        typename std::enable_if<(std::is_base_of<detail::KernelTag, KImpl>::value), void>::type
         includeHelper()
         {
             auto backend     = KImpl::backend();
@@ -371,13 +371,11 @@ namespace gapi {
 
         /// @private
         // Partial include() specialization for transformations
-        template <typename TImpl, cv::detail::PackageObjectTag Val>
-        typename std::enable_if<(Val == cv::detail::PackageObjectTag::TRANSFORMATION), void>::type
+        template <typename TImpl>
+        typename std::enable_if<(std::is_base_of<detail::TransformTag, TImpl>::value), void>::type
         includeHelper()
         {
-            auto transform_impl = GTransform{transformation<TImpl>()};
-
-            m_transformations.push_back(std::move(transform_impl));
+            m_transformations.emplace_back(TImpl::transformation());
         }
 
     public:
@@ -400,7 +398,7 @@ namespace gapi {
         template<typename KImpl>
         bool includes() const
         {
-            return includesHelper<KImpl, KImpl::object_entity()>();
+            return includesHelper<KImpl>();
         }
 
         /**
@@ -460,18 +458,11 @@ namespace gapi {
         /**
          * @brief Put a new kernel implementation or a new transformation
          * KImpl into the package.
-         *
-         * @param up unite policy to use. If the package has already
-         * implementation for this kernel (probably from another
-         * backend), and cv::unite_policy::KEEP is passed, the
-         * existing implementation remains in package; on
-         * cv::unite_policy::REPLACE all other existing
-         * implementations are first dropped from the package.
          */
         template<typename KImpl>
         void include()
         {
-            includeHelper<KImpl, KImpl::object_entity()>();
+            includeHelper<KImpl>();
         }
 
         /**
--- a/modules/gapi/include/opencv2/gapi/gkernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/gkernel.hpp
@@ -359,7 +359,7 @@ namespace gapi {
         // Partial include() specialization for kernels
         template <typename KImpl, cv::detail::PackageObjectTag Val>
         typename std::enable_if<(Val == cv::detail::PackageObjectTag::KERNEL), void>::type
-        includeHelper(const cv::unite_policy up)
+        includeHelper()
         {
             auto backend     = KImpl::backend();
             auto kernel_id   = KImpl::API::id();
@@ -373,9 +373,8 @@ namespace gapi {
         // Partial include() specialization for transformations
         template <typename TImpl, cv::detail::PackageObjectTag Val>
         typename std::enable_if<(Val == cv::detail::PackageObjectTag::TRANSFORMATION), void>::type
-        includeHelper(const cv::unite_policy up)
+        includeHelper()
         {
-            (void)up; // fix warning
             auto transform_impl = GTransform{transformation<TImpl>()};
 
             m_transformations.push_back(std::move(transform_impl));
@@ -472,7 +471,7 @@ namespace gapi {
         template<typename KImpl>
         void include()
         {
-            includeHelper<KImpl, KImpl::object_entity()>(up);
+            includeHelper<KImpl, KImpl::object_entity()>();
         }
 
         /**
@@ -515,6 +514,10 @@ namespace gapi {
      */
     template<typename... KK> GKernelPackage kernels()
     {
+        // FIXME: currently there is no check that transformations' signatures are unique
+        // and won't be any intersection in graph compilation stage
+        static_assert(detail::all_unique<typename KK::API...>::value, "Kernels API must be unique");
+
         GKernelPackage pkg;
 
         // For those who wonder - below is a trick to call a number of
@@ -523,8 +526,6 @@ namespace gapi {
         // Just note that `f(),a` always equals to `a` (with f() called!)
         // and parentheses are used to hide function call in the expanded sequence.
         // Leading 0 helps to handle case when KK is an empty list (kernels<>()).
-
-        static_assert(detail::all_unique<typename KK::API...>::value, "Kernels API must be unique");
         int unused[] = { 0, (pkg.include<KK>(), 0)... };
         cv::util::suppress_unused_warning(unused);
         return pkg;
--- a/modules/gapi/include/opencv2/gapi/gkernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/gkernel.hpp
@@ -388,12 +388,11 @@ namespace gapi {
         std::size_t size() const;
 
         /**
-         * @brief Returns total number of transformations
-         * in the package (across all backends included)
+         * @brief Returns vector of transformations included in the package
          *
-         * @return a number of transformations in the package
+         * @return vector of transformations included in the package
          */
-        std::size_t transform_size() const;
+        const std::vector<GTransform> &get_transformations() const;
 
         /**
          * @brief Test if a particular kernel _implementation_ KImpl is
--- a/modules/gapi/include/opencv2/gapi/gkernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/gkernel.hpp
@@ -2,7 +2,7 @@
 // It is subject to the license terms in the LICENSE file found in the top-level directory
 // of this distribution and at http://opencv.org/license.html.
 //
-// Copyright (C) 2018 Intel Corporation
+// Copyright (C) 2019 Intel Corporation
 
 
 #ifndef OPENCV_GAPI_GKERNEL_HPP
@@ -171,7 +171,7 @@ namespace detail
 // GKernelType and GKernelTypeM are base classes which implement typed ::on()
 // method based on kernel signature. GKernelTypeM stands for multiple-return-value kernels
 //
-// G_TYPED_KERNEL and G_TYPED_KERNEK_M macros inherit user classes from GKernelType and
+// G_TYPED_KERNEL and G_TYPED_KERNEL_M macros inherit user classes from GKernelType and
 // GKernelTypeM respectively.
 
 template<typename K, typename... R, typename... Args>
@@ -299,12 +299,13 @@ namespace gapi {
     // FIXME: Hide implementation
     /**
      * @brief A container class for heterogeneous kernel
-     * implementation collections.
+     * implementation collections and graph transformations.
      *
      * GKernelPackage is a special container class which stores kernel
-     * _implementations_. Objects of this class are created and passed
-     * to cv::GComputation::compile() to specify which kernels to use
-     * in the compiled graph. GKernelPackage may contain kernels of
+     * _implementations_ and graph transformations. Objects of this class
+     * are created and passed to cv::GComputation::compile() to specify
+     * which kernels to use and which transformations to apply in the
+     * compiled graph. GKernelPackage may contain kernels of
      * different backends, e.g. be heterogeneous.
      *
      * The most easy way to create a kernel package is to use function
@@ -316,7 +317,8 @@ namespace gapi {
      * with an empty package (created with the default constructor)
      * and then by populating it with kernels via call to
      * GKernelPackage::include(). Note this method is also a template
-     * one since G-API kernel implementations are _types_, not objects.
+     * one since G-API kernel and transformation implementations are _types_,
+     * not objects.
      *
      * Finally, two kernel packages can be combined into a new one
      * with function cv::gapi::combine().
@@ -343,15 +345,8 @@ namespace gapi {
         // Remove ALL implementations of the given API (identified by ID)
         void removeAPI(const std::string &id);
 
-    public:
-        /**
-         * @brief Returns total number of kernels in the package
-         * (across all backends included)
-         *
-         * @return a number of kernels in the package
-         */
-        std::size_t size() const;
-
+        /// @private
+        // Partial includes() specialization for kernels
         template <typename KImpl, cv::detail::PackageObjectTag Val>
         typename std::enable_if<(Val == cv::detail::PackageObjectTag::KERNEL), bool>::type
         includesHelper() const {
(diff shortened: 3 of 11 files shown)


modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -398,10 +413,7 @@ namespace gapi {
         template<typename KImpl>
         bool includes() const
         {
-            const auto set_iter = m_backend_kernels.find(KImpl::backend());
-            return (set_iter != m_backend_kernels.end())
-                ? (set_iter->second.count(KImpl::API::id()) > 0)
-                : false;
+            return includesHelper<KImpl, KImpl::object_entity()>();
```
smirnov-alexey: It is only supported for kernels for now. Probably transformations also require unique ids to find them in package.

modules/gapi/include/opencv2/gapi/util/util.hpp
```diff
@@ -84,6 +84,28 @@ namespace detail
     {
         static constexpr const std::size_t value = S;
     };
+
+    enum class PackageObjectTag {
```
smirnov-alexey: Potentially we can just derive from blank base class and use `std::is_base_of` when need to determine what type of object it is.

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -0,0 +1,202 @@
...
+    template <int... IIs, int... OIs>
+    static GArgs get_pattern_impl(const GArgs &in_args, detail::Seq<IIs...>, detail::Seq<OIs...>)
+    {
+        using R = std::tuple<typename TransType<Outs>::type...>;
+        const R r = K::pattern(get_in_args<Ins>(in_args, IIs)...);
+        return GArgs{GArg(std::get<OIs>(r))...};
+    }
+
+    template <int... IIs, int... OIs>
+    static GArgs get_substitute_impl(const GArgs &in_args, detail::Seq<IIs...>, detail::Seq<OIs...>)
```
smirnov-alexey: Pattern and substitute have the same signature. Need to figure out how to get rid of copy-paste.

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -387,6 +393,15 @@ namespace gapi {
          */
         std::size_t size() const;
 
+        template <typename KImpl, cv::detail::PackageObjectTag Val>
+        typename std::enable_if<(Val == cv::detail::PackageObjectTag::KERNEL), bool>::type
```
smirnov-alexey: Wondering if it is a good way to write like that...

smirnov-alexey:
Current structure description:
1) In utils.h were created 2 classes to separate kernels from transformations via "tag" in kernel package;
2) GAPI_TRANSFORM is implemented in a way similar to G_TYPED_KERNEL

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -2,7 +2,7 @@
 // It is subject to the license terms in the LICENSE file found in the top-level directory
 // of this distribution and at http://opencv.org/license.html.
 //
-// Copyright (C) 2018 Intel Corporation
+// Copyright (C) 2019 Intel Corporation
```
rgarnov: Should be `2018-2019`

rgarnov: (Applies to all changed headers)

modules/gapi/src/api/gkernel.cpp
```diff
@@ -42,12 +42,14 @@ void cv::gapi::GKernelPackage::removeAPI(const std::string &id)
...
 std::size_t cv::gapi::GKernelPackage::size() const
 {
-    return std::accumulate(m_backend_kernels.begin(),
+    std::size_t pkg_size = std::accumulate(m_backend_kernels.begin(),
                            m_backend_kernels.end(),
                            static_cast<std::size_t>(0u),
                            [](std::size_t acc, const M::value_type& v) {
                                return acc + v.second.size();
                            });
+    pkg_size += m_transformations.size();
```
rgarnov: I'm not sure if it's an expected behavior, this is kernel package after all)

smirnov-alexey: All the names and stuff are discussable

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
smirnov-alexey: Got it, thanks

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -0,0 +1,210 @@
...
+class GTransformImpl;
+
+template <typename K, typename R, typename... Args>
+class GTransformImpl<K, std::function<R(Args...)>> : public TransHelper<K, std::tuple<Args...>, R>,
+                                                     public cv::detail::TransformTag
+{
+public:
+    // FIXME: currently there is no check that transformations' signatures are unique
+    // and won't be any intersection in graph compilation stage
+    using API = K;
```
smirnov-alexey: This is kinda workaround. Need to think about 2 signatures having same signature but different kernels in use.

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -0,0 +1,210 @@
...
+    using type = cv::GArray<U>;
+};
+template <typename T>
+struct TransType
+{
+    using type = T;
+};
+
+template <typename T>
+using is_nongapi_type2 = std::is_same<T, typename TransType<T>::type>;
```
rgarnov: Why `is_nongapi_type ` from gkernel.hpp cannot be utulized here?

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -417,17 +458,20 @@ namespace gapi {
 
         // FIXME: No overwrites allowed?
         /**
-         * @brief Put a new kernel implementation KImpl into package.
+         * @brief Put a new kernel implementation or a new transformation
+         * KImpl into the package.
+         *
+         * @param up unite policy to use. If the package has already
```
rgarnov: There is no `up` param, please remove

modules/gapi/include/opencv2/gapi/gtransform.hpp
smirnov-alexey: Approach has been reconsidered

modules/gapi/test/gapi_transform_tests.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+namespace opencv_test
+{
+
+namespace
+{
+using GMat2 = std::tuple<GMat, GMat>;
+using GMat3 = std::tuple<GMat, GMat, GMat>;
+
+GAPI_TRANSFORM(my_transform, <cv::GMat(GMat2)>, "does nothing")
```
smirnov-alexey: Suppose all the specific information will be defined in parameters inside the class, and no non-G* types are passed in signature (otherwise there will be a compilation error)

smirnov-alexey:
@dmatveev

modules/gapi/include/opencv2/gapi/gtransform.hpp
andrey-golubev: something like this should work (and if we always operate with `GArg`/`GArgs` this may be a good candidate for universal utility):

```cpp
template<typename F, int... IIs, int... OIs>
static GArgs get_impl(const GArgs &in_args, detail::Seq<IIs...>, detail::Seq<OIs...>) {
        using R = std::tuple<Outs...>;
        const R r = F(in_args.at(IIs).template get<Ins>()...);
        return GArgs{GArg(std::get<OIs>(r))...};
}
```
usage:
```cpp
    static GArgs get_pattern(const GArgs &in_args)
    {
        return get_impl(K::pattern, in_args, typename detail::MkSeq<sizeof...(Ins)>::type(),
                                typename detail::MkSeq<sizeof...(Outs)>::type());
    }
    static GArgs get_substitute(const GArgs &in_args)
    {
        return get_impl(K::substitute, in_args, typename detail::MkSeq<sizeof...(Ins)>::type(),
                                   typename detail::MkSeq<sizeof...(Outs)>::type());
    }
```

Also, the whole thing looks like [invoke](https://en.cppreference.com/w/cpp/utility/functional/invoke) (since C++17, though), might be better to find good standard alternative

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -0,0 +1,161 @@
...
+public:
+    // FIXME: currently there is no check that transformations' signatures are unique
+    // and won't be any intersection in graph compilation stage
+    using API = K;
+};
+
+template <typename, typename>
+class GTransformImplM;
+
+template <typename K, typename... R, typename... Args>
```
andrey-golubev: hmm...

modules/gapi/src/api/gkernel.cpp
```diff
@@ -66,6 +66,9 @@ cv::gapi::GKernelPackage cv::gapi::combine(const GKernelPackage  &lhs,
                 result.m_id_kernels.emplace(kernel.first, kernel.second);
             }
         }
+        for (const auto &transforms : lhs.m_transformations){
+            result.m_transformations.push_back(transforms);
```
andrey-golubev: no collision detection  for transformations, is this a good thing? might be good point to scrub the vector with `std::unique` or something

modules/gapi/include/opencv2/gapi/gtransform.hpp
andrey-golubev: remark:
```cpp
static GArgs get_impl(F f, const GArgs &in_args, detail::Seq<IIs...>, detail::Seq<OIs...>) {
  const R r = f(in_args.at(IIs).template get<Ins>()...);
...
}
```

modules/gapi/include/opencv2/gapi/gkernel.hpp
smirnov-alexey: Yes, such containers will give us better performance but we need unique ids, which are not in agreement yet.

modules/gapi/include/opencv2/gapi/util/util.hpp
smirnov-alexey: Let's think about what if we extend this enum and how we are going to separate different entities in that case.

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -337,12 +345,47 @@ namespace gapi {
...
+            auto transform_impl = GTransform{transformation<TImpl>()};
+
+            m_transformations.push_back(std::move(transform_impl));
+        }
+
     public:
         /**
-         * @brief Returns total number of kernels in the package
-         * (across all backends included)
+         * @brief Returns total number of kernels and transformations
```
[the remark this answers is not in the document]
smirnov-alexey: We can handle all different types of sizes but like some comment above it's more about naming and stuff (size, size_t, size_k, ...). Please, provide an example.

modules/gapi/src/api/gkernel.cpp
smirnov-alexey: It would be tough to write all operator in a proper way, so maybe it's better to pick the easiest option considering our actual use case.

modules/gapi/include/opencv2/gapi/gkernel.hpp
andrey-golubev: let's think about `std::unordered_set` or something similar then: they do not require ids. anyway, I'm also not sure we need this:

@rgarnov @dmatveev what do you think?

modules/gapi/include/opencv2/gapi/gkernel.hpp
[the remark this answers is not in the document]
andrey-golubev: how do we use `size()` now? How your change may affect current behavior?

modules/gapi/include/opencv2/gapi/gtransform.hpp
smirnov-alexey: Done, thanks!

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
smirnov-alexey: Done!

modules/gapi/include/opencv2/gapi/gtransform.hpp
smirnov-alexey: Done! Please, resolve the conversation.

modules/gapi/include/opencv2/gapi/gkernel.hpp
smirnov-alexey: Removed. Please, verify.

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -296,12 +299,13 @@ namespace gapi {
...
     /**
      * @brief A container class for heterogeneous kernel
-     * implementation collections.
+     * implementation collections and graph transformations.
      *
      * GKernelPackage is a special container class which stores kernel
-     * _implementations_. Objects of this class are created and passed
-     * to cv::GComputation::compile() to specify which kernels to use
-     * in the compiled graph. GKernelPackage may contain kernels of
+     * _implementations_ and graph transformations. Objects of this class
```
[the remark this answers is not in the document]
smirnov-alexey: Done

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -0,0 +1,161 @@
...
+        return get_pattern_impl(in_args, typename detail::MkSeq<sizeof...(Ins)>::type());
+    }
+    static GArgs get_substitute(const GArgs &in_args)
+    {
+        return get_substitute_impl(in_args, typename detail::MkSeq<sizeof...(Ins)>::type());
+    }
+};
+
+template <typename K, typename... Ins, typename... Outs>
+struct TransHelper<K, std::tuple<Ins...>, std::tuple<Outs...>>
```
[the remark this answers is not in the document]
smirnov-alexey: Done, thanks!

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -0,0 +1,161 @@
...
+        using R = std::tuple<Outs...>;
+        const R r = K::pattern(in_args.at(IIs).template get<Ins>()...);
+        return GArgs{GArg(std::get<OIs>(r))...};
+    }
+
+    template <int... IIs, int... OIs>
+    static GArgs get_substitute_impl(const GArgs &in_args, detail::Seq<IIs...>, detail::Seq<OIs...>)
+    {
+        using R = std::tuple<Outs...>;
+        const R r = K::pattern(in_args.at(IIs).template get<Ins>()...);
```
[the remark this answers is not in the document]
smirnov-alexey: Done

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+struct GAPI_EXPORTS GTransform
+{
+    using F = std::function<GArgs(const GArgs &)>;
+
+    const char *description;
+    F pattern;
+    F substitute;
+
+    explicit GTransform(const char *d, const F &p, const F &s) : description(d), pattern(p), substitute(s){};
```
rgarnov: There is no need for `explicit` in this case

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+    static GArgs get_pattern(const GArgs &in_args)
+    {
+        return get_impl(K::pattern, in_args, typename detail::MkSeq<sizeof...(Ins)>::type(),
+                        typename detail::MkSeq<std::tuple_size<typename detail::tuple_wrap_helper<Out>::type>::value>::type());
+    }
+    static GArgs get_substitute(const GArgs &in_args)
+    {
+        return get_impl(K::substitute, in_args, typename detail::MkSeq<sizeof...(Ins)>::type(),
+                        typename detail::MkSeq<std::tuple_size<typename detail::tuple_wrap_helper<Out>::type>::value>::type());
```
rgarnov: `TransHelper` is already in `detail` namespace, all `detail::` stuff above can be removed

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -0,0 +1,98 @@
...
+    }
+};
+} // namespace detail
+
+template <typename, typename>
+class GTransformImpl;
+
+template <typename K, typename R, typename... Args>
+class GTransformImpl<K, std::function<R(Args...)>> : public detail::TransHelper<K, std::tuple<Args...>, R>,
+                                                     public cv::detail::TransformTag
```
rgarnov: Please be consistent, use `cv::detail` or `detail` for both parents

modules/gapi/include/opencv2/gapi/util/util.hpp
```diff
@@ -97,6 +97,36 @@ namespace detail
...
+    struct tuple_wrap_helper<std::tuple<Objs...>>
+    {
+        using type = std::tuple<Objs...>;
+        static type get(std::tuple<Objs...>&& objs) { return std::forward<std::tuple<Objs...>>(objs); }
+    };
+
+    enum class PackageObjectTag {
+        KERNEL,
+        TRANSFORMATION
+    };
```
rgarnov: Is not used anywhere

modules/gapi/include/opencv2/gapi/util/util.hpp
```diff
@@ -97,6 +97,36 @@ namespace detail
...
+
+    struct KernelTag
+    {
+    };
+
+    struct TransformTag
+    {
+    };
```
rgarnov: These structures must be moved somewhere else

smirnov-alexey: Moved to gcommon.h

modules/gapi/include/opencv2/gapi/util/util.hpp
smirnov-alexey: Removed

modules/gapi/include/opencv2/gapi/gtransform.hpp
smirnov-alexey: Done

modules/gapi/include/opencv2/gapi/gtransform.hpp
smirnov-alexey: Removed

modules/gapi/include/opencv2/gapi/gtransform.hpp
smirnov-alexey: Removed

modules/gapi/include/opencv2/gapi/gkernel.hpp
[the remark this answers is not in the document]
smirnov-alexey: Introduced a transform_size()

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -337,15 +345,56 @@ namespace gapi {
         // Remove ALL implementations of the given API (identified by ID)
         void removeAPI(const std::string &id);
 
+        /// @private
+        // Partial includes() specialization for kernels
+        template <typename KImpl>
+        typename std::enable_if<(std::is_base_of<detail::KernelTag, KImpl>::value), bool>::type
+        includesHelper() const {
```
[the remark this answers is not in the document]
andrey-golubev: if we don't need one, may be it's better to add some `static_assert` to the `includes()` function below so that we explicitly say that transformations are not checked this way?

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -0,0 +1,98 @@
...
+{
+
+template <typename, typename, typename>
+struct TransHelper;
+
+template <typename K, typename... Ins, typename Out>
+struct TransHelper<K, std::tuple<Ins...>, Out>
+{
+    template <typename Callable, int... IIs, int... OIs>
+    static GArgs get_impl(Callable f, const GArgs &in_args, Seq<IIs...>, Seq<OIs...>)
```
andrey-golubev: let's call it `invoke` or something, "get_impl" was a bad name I proposed without thinking about it too much. Now I think it doesn't make much sense. "invoke" sounds much better

modules/gapi/test/gapi_transform_tests.cpp
```diff
@@ -0,0 +1,114 @@
...
+namespace opencv_test
+{
+
+namespace
+{
+using GMat2 = std::tuple<GMat, GMat>;
+using GMat3 = std::tuple<GMat, GMat, GMat>;
+using GMat = cv::GMat;
+
+GAPI_TRANSFORM(my_transform, <GMat(GMat, GMat)>, "does nothing")
```
andrey-golubev: let's add one/two for single input argument as well:
`<GMat(GMat)>` or `<GMat2(GMat)>`

modules/gapi/include/opencv2/gapi/util/util.hpp
smirnov-alexey: Changed to be a blank class inheritance.

modules/gapi/include/opencv2/gapi/gkernel.hpp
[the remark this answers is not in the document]
smirnov-alexey: Now it returns the whole vector of transformations of which you can get size. Yet it's not the final proposal.

modules/gapi/include/opencv2/gapi/gcommon.hpp
```diff
@@ -29,6 +29,12 @@ namespace detail
     {
         static const char* tag() { return ""; };
     };
+
+    // These structures are tags which suppose to separate kernels and transformations
```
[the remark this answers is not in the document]
smirnov-alexey: Fixed, thanks!

modules/gapi/include/opencv2/gapi/gkernel.hpp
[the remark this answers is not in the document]
smirnov-alexey: Done, thanks!

modules/gapi/include/opencv2/gapi/gtransform.hpp
smirnov-alexey: Renamed

modules/gapi/test/gapi_transform_tests.cpp
smirnov-alexey: Done

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -337,29 +345,71 @@ namespace gapi {
...
          * @return true if there is such kernel, false otherwise.
          */
         template<typename KImpl>
         bool includes() const
         {
-            auto kernel_it = m_id_kernels.find(KImpl::API::id());
-            return kernel_it != m_id_kernels.end() &&
-                   kernel_it->second.first == KImpl::backend();
+            static_assert(!(std::is_base_of<detail::TransformTag, KImpl>::value),
+                          "includes() cannot be applied to transformations");
```
dmatveev: This check & message are not very robust.
What if later we extend our vocabulary with some new term (say, `Something` in addition to kernels and transformations) and forget about this method? Passing `Something` will raise an error that it is a transformation while it is in fact `Something`.

I'd check explicitly that `KImpl` has a kernel tag and fix text to `includes() can only be applied to kernels`.

modules/gapi/include/opencv2/gapi/gtransform.hpp
dmatveev: Why do you need `API` defined here BTW? Kernel package (which used that) should be already neutral to this, I believe.

modules/gapi/src/api/gkernel.cpp
dmatveev: What is collision for transformations? Having multiple transformations of the same type in the Kernel package? This can be solved by `all_unique` thing recently written by @TolyaTalamanov

modules/gapi/test/gapi_transform_tests.cpp
```diff
@@ -0,0 +1,148 @@
...
+    static GMat2 pattern(GMat)
+    {
+        return {};
+    };
+
+    static GMat2 substitute(GMat)
+    {
+        return {};
+    }
+};
```
dmatveev: `GMat`s are nice, what about `GScalar`s, `GArray<T>`s? :)

rgarnov: Will add

modules/gapi/include/opencv2/gapi/gkernel.hpp
rgarnov: Will fix

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -199,7 +201,8 @@ template<typename, typename> class GKernelType;
 
 template<typename K, typename R, typename... Args>
 class GKernelType<K, std::function<R(Args...)> >:
-        public detail::MetaHelper<K, std::tuple<Args...>, R >
+        public detail::MetaHelper<K, std::tuple<Args...>, R >,
+        public cv::detail::KernelTag
```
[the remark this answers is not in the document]
rgarnov: Will fix

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -170,12 +171,13 @@ namespace detail
...
 //
-// G_TYPED_KERNEL and G_TYPED_KERNEK_M macros inherit user classes from GKernelType and
+// G_TYPED_KERNEL and G_TYPED_KERNEL_M macros inherit user classes from GKernelType and
 // GKernelTypeM respectively.
 
 template<typename K, typename... R, typename... Args>
 class GKernelTypeM<K, std::function<std::tuple<R...>(Args...)> >:
-        public detail::MetaHelper<K, std::tuple<Args...>, std::tuple<R...> >
+        public detail::MetaHelper<K, std::tuple<Args...>, std::tuple<R...> >,
+        public cv::detail::KernelTag
```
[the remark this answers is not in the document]
rgarnov: Will fix

modules/gapi/include/opencv2/gapi/gtransform.hpp
rgarnov: `API` is checked on uniqueness via `all_unique` in kernels() method

modules/gapi/src/api/gkernel.cpp
rgarnov: This is already done this way (that's why we need `API` defined in transformations)

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -388,12 +388,11 @@ namespace gapi {
...
         /**
-         * @brief Returns total number of transformations
-         * in the package (across all backends included)
+         * @brief Returns vector of transformations included in the package
          *
-         * @return a number of transformations in the package
+         * @return vector of transformations included in the package
          */
-        std::size_t transform_size() const;
+        const std::vector<GTransform> &get_transformations() const;
```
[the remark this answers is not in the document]
rgarnov: done

rgarnov:
Updated

modules/gapi/include/opencv2/gapi/gkernel.hpp
dmatveev: Is it still relevant to look at?

dmatveev:
@andrey-golubev please resolve the discussions you feel resolved

modules/gapi/include/opencv2/gapi/gtransform.hpp
dmatveev: Oh, this is nice, thanks!

modules/gapi/src/api/gkernel.cpp
dmatveev: hmm, I believe there might be a corner case:
```c++
auto pkg = cv::gapi::kernels<TransformA>(); // all_unique<> passes the check
pkg.include<TransformA>(); // all_unique<> is not called at all!
```
I am not sure how to handle this, and we also can reach the same with `combine(...)`.
Let's file a bug and fix it later, there's no clear solution to this as I see it.

dmatveev: Still, what's wrong with having multiple similar transforms in the package? One of those duplicates will be called anyway. If there's nothing to transform, neither of them is called (also anyway). Not perfect, but not that awful too.

modules/gapi/test/gapi_transform_tests.cpp
dmatveev: And so it worked? :)

dmatveev: Ah I see, for the output type. OK

modules/gapi/include/opencv2/gapi/gkernel.hpp
andrey-golubev: @dmatveev no, I think this one is resolved (for me at least)

dmatveev:
Thanks!

dmatveev:
Thanks!

PR #14661, opened 2019-05-28
Enable more deconvolution layer configurations with IE backend

dkurt:
<cut/>

```
docker_image:Custom=ubuntu-openvino-2019r1:16.04
docker_image:Custom Win=openvino-2019r1
docker_image:Custom Mac=openvino-2019r1
test_modules=dnn,python2,python3,java

buildworker:Custom=linux-2
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/test/test_halide_layers.cpp
+++ b/modules/dnn/test/test_halide_layers.cpp
@@ -159,10 +159,6 @@ TEST_P(Deconvolution, Accuracy)
     Backend backendId = get<0>(get<7>(GetParam()));
     Target targetId = get<1>(get<7>(GetParam()));
 
-    if (backendId == DNN_BACKEND_INFERENCE_ENGINE && (targetId == DNN_TARGET_CPU || targetId == DNN_TARGET_MYRIAD) &&
-        dilation.width == 2 && dilation.height == 2)
-        throw SkipTestException("");
-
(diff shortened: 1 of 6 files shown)


modules/dnn/test/test_halide_layers.cpp
```diff
@@ -216,7 +212,7 @@ INSTANTIATE_TEST_CASE_P(Layer_Test_Halide, Deconvolution, Combine(
 /*in size*/  Values(Size(5, 6)),
 /*kernel*/   Values(Size(3, 1), Size(1, 3)),
 /*pad*/      Values(Size(1, 0), Size(0, 1)),
-/*dilation*/ Values(Size(1, 1), Size(2, 2)),
+/*dilation*/ Values(Size(1, 1)),
```
alalek: Why is this test case reduced?

dkurt: @alalek, I found that it's not implemented even in OpenCV backend:

there is dilation for convolution layer:
https://github.com/opencv/opencv/blob/db900af8c7a849539aff3941c2a788c190bb317b/modules/dnn/src/layers/convolution_layer.cpp#L287-L295

but not for deconvolution one:
https://github.com/opencv/opencv/blob/db900af8c7a849539aff3941c2a788c190bb317b/modules/dnn/src/layers/convolution_layer.cpp#L1287-L1301

PR #14678, opened 2019-05-30
Fix build issue on QNX platform

tedsteiner:
[QNX](http://blackberry.qnx.com/en/products/neutrino-rtos/index) is a POSIX-compliant real-time embedded operating system. In general OpenCV's core builds correctly with the QNX cross-compiler. However, there is a small conflict with the `gettimeofday` function.

The QNX implementation of the C standard library does not define the `timezone` struct, and the `tv` field of `gettimeofday` is not used. Documentation is available here: http://www.qnx.com/developers/docs/6.5.0/index.jsp?topic=%2Fcom.qnx.doc.neutrino_lib_ref%2Fg%2Fgettimeofday.html

Therefore, to maintain build compatibility with QNX we can define variable `tz` as an int rather than a `timezone`, which does not affect anything downstream because `tz` is unused but does fix the compile error.

### DIFF
--- a/modules/core/src/system.cpp
+++ b/modules/core/src/system.cpp
@@ -757,8 +757,12 @@ int64 getTickCount(void)
 #elif defined __MACH__ && defined __APPLE__
     return (int64)mach_absolute_time();
 #else
-    struct timeval tv;
+#if defined __QNX__
+    int tz;
+#else
     struct timezone tz;
+#endif
+    struct timeval tv;
     gettimeofday( &tv, &tz );
     return (int64)tv.tv_sec*1000000 + tv.tv_usec;
 #endif
--- a/modules/core/src/system.cpp
+++ b/modules/core/src/system.cpp
@@ -757,13 +757,8 @@ int64 getTickCount(void)
 #elif defined __MACH__ && defined __APPLE__
     return (int64)mach_absolute_time();
 #else
-#if defined __QNX__
-    int tz;
-#else
-    struct timezone tz;
-#endif
     struct timeval tv;
-    gettimeofday( &tv, &tz );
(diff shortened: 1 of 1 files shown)


modules/core/src/system.cpp
```diff
@@ -712,8 +712,12 @@ int64 getTickCount(void)
...
     return (int64)mach_absolute_time();
 #else
-    struct timeval tv;
+#if defined __QNX__
+    int tz;
+#else
     struct timezone tz;
+#endif
+    struct timeval tv;
     gettimeofday( &tv, &tz );
```
alalek: [Documentation](http://www.qnx.com/developers/docs/6.5.0/index.jsp?topic=%2Fcom.qnx.doc.neutrino_lib_ref%2Fg%2Fgettimeofday.html) says that "This pointer must be NULL or the behavior of gettimeofday() is unspecified" for the second `void*` parameter.
So it is better do not pass pointer on int value on stack.

For both Linux (validated) / QNX this code should work well:
```
    struct timeval tv;
    gettimeofday(&tv, NULL);
```

tedsteiner: Yes, definitely agree. I considered that but was trying to make the most minimal change possible without having access to every supported architecture. But I checked this in the Linux implementation documentation and saw that they said the use of the timezone is actually now deprecated and NULL should be passed in general. http://man7.org/linux/man-pages/man2/gettimeofday.2.html

Looking more closely at what `gettimeofday` does, what it's doing is setting the timezone rather than using the timezone. This is probably why a non-null pointer might have undefined behavior, because it might write data to an undefined location in memory. Good catch!

PR #14679, opened 2019-05-30
cmake: update ENABLE_FAST_MATH option

alalek:
```
ci_branch=fp_precise
force_builders=Custom Win
build_image:Custom Win=msvs2019-win32
```


alalek:
:+1:

PR #14682, opened 2019-05-31
Add Axpy layer

l-bat:
**Merge with extra:** opencv/opencv_extra#614

```
docker_image:Custom=ubuntu-openvino-2019r1.1:16.04
docker_image:Custom Win=openvino-2019r1
docker_image:Custom Mac=openvino-2019r1
test_modules=dnn,python2,python3,java

buildworker:Custom=linux-2
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/caffe/caffe_importer.cpp
+++ b/modules/dnn/src/caffe/caffe_importer.cpp
@@ -388,6 +388,33 @@ public:
                     layerParams.blobs[1].setTo(1);  // std
                 }
             }
+            else if (type == "Axpy")
+            {
+                CV_Assert_N(layer.bottom_size() == 3, layer.top_size() == 1);
+
+                std::string scaleName = name + "/scale";
+                int repetitions = layerCounter[scaleName]++;
+                if (repetitions) {
+                    scaleName += String("_") + toString(repetitions);
+                }
+
+                LayerParams scaleParams;
+                scaleParams.set("axis", 1);
+                scaleParams.set("has_bias", false);
+                int scaleId = dstNet.addLayer(scaleName, "Scale", scaleParams);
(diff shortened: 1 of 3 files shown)


modules/dnn/src/caffe/caffe_importer.cpp
```diff
@@ -388,6 +388,33 @@ class CaffeImporter
...
+                addOutput(layer, scaleId, 0);
+                net.mutable_layer(li)->set_bottom(0, layer.top(0));
+
+                LayerParams eltwiseParams;
+                eltwiseParams.set("operation", "sum");
+                int eltwiseId = dstNet.addLayer(name, "Eltwise", eltwiseParams);
+                addInput(layer.top(0), eltwiseId, 0, dstNet);
+                addInput(layer.bottom(1), eltwiseId, 1, dstNet);
+                addOutput(layer, eltwiseId, 0);
+                continue;
```
dkurt: If we replace Eltwise layer creation to just `type = "Eltwise"` (like in the next case), will it work?

modules/dnn/test/test_caffe_importer.cpp
```diff
@@ -109,6 +109,49 @@ TEST(Test_Caffe, read_googlenet)
     ASSERT_FALSE(net.empty());
 }
 
+TEST_P(Test_Caffe_nets, Axpy)
+{
+    if (backend == DNN_BACKEND_INFERENCE_ENGINE)
+        throw SkipTestException("");
```
dkurt: Can we enable IE backend? Initializing `ScaleShift` layer with variable inputs by one of IE layers?

PR #14686, opened 2019-05-31
doc: add parameter description

asashour:
Add javadoc parameter declaration.

This seems to prevent https://github.com/opencv/opencv_contrib/pull/2131. There log is [here](https://pullrequest.opencv.org/buildbot/builders/precommit-contrib_docs/builds/12485/steps/make%20doxygen/logs/warnings%20%281%29)

Although the error says "Parameter is documented more than once", the parameter is not duplicate, however there is another one named "probs", not sure if javadoc consider both "probs0" and "probs" the same, in which case we should rename one of them.

```
**WIP**
force_builders=Docs
```


alalek:
Move these changes into #14667

Currently combination of #14667 and https://github.com/opencv/opencv_contrib/pull/2131 is tested and you see their combined results.

PR #14689, opened 2019-05-31
Cancelation support fo Async GAPI

anton-potapov:
### Cancellation support for Async GAPI

```
```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/gasync_context.hpp
+++ b/modules/gapi/include/opencv2/gapi/gasync_context.hpp
@@ -1,0 +1,38 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2019 Intel Corporation
+
+#ifndef OPENCV_GAPI_GASYNC_CONTEXT_HPP
+#define OPENCV_GAPI_GASYNC_CONTEXT_HPP
+
+#if !defined(GAPI_STANDALONE)
+#  include <opencv2/core/cvdef.h>
+#else   // Without OpenCV
+#  include <opencv2/gapi/own/cvdefs.hpp>
+#endif // !defined(GAPI_STANDALONE)
+
+#include <opencv2/gapi/own/exports.hpp>
+
+namespace cv {
+namespace gapi{
+namespace wip {
+
+class GAPI_EXPORTS GAsyncContext{
+    std::atomic<bool> cancelation_requested = {false};
+public:
+    //returns true if it was a first request to cancel the context
+    bool cancel();
+    bool isCanceled() const;
+};
+
+class GAPI_EXPORTS GAsyncCanceled : public std::exception {
+public:
+    virtual const char* what() const noexcept CV_OVERRIDE;
+};
+} // namespace wip
+} // namespace gapi
+} // namespace cv
+
+#endif //OPENCV_GAPI_GASYNC_CONTEXT_HPP
--- a/modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
+++ b/modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
@@ -12,6 +12,7 @@
 #include <exception>        //for std::exception_ptr
 #include <functional>       //for std::function
 #include <opencv2/gapi/garg.hpp>
+#include <opencv2/gapi/own/exports.hpp>
 
 namespace cv {
     //fwd declaration
@@ -19,13 +20,17 @@ namespace cv {
 
 namespace gapi{
 namespace wip {
+    class GAsyncContext;
     //These functions asynchronously (i.e. probably on a separate thread of execution) call operator() member function of their first argument with copies of rest of arguments (except callback) passed in.
     //The difference between the function is the way to get the completion notification (via callback or a waiting on std::future object)
     //If exception is occurred during execution of apply it is transfered to the callback (via function parameter) or passed to future (and will be thrown on call to std::future::get)
     GAPI_EXPORTS void                async(GCompiled& gcmpld, std::function<void(std::exception_ptr)>&& callback, GRunArgs &&ins, GRunArgsP &&outs);
+    GAPI_EXPORTS void                async(GCompiled& gcmpld, std::function<void(std::exception_ptr)>&& callback, GRunArgs &&ins, GRunArgsP &&outs, GAsyncContext& ctx);
+
     GAPI_EXPORTS std::future<void>   async(GCompiled& gcmpld, GRunArgs &&ins, GRunArgsP &&outs);
-} // namespace gapi
+    GAPI_EXPORTS std::future<void>   async(GCompiled& gcmpld, GRunArgs &&ins, GRunArgsP &&outs, GAsyncContext& ctx);
 } // namespace wip
+} // namespace gapi
 } // namespace cv
 
 #endif // OPENCV_GAPI_GCOMPILED_ASYNC_HPP
--- a/modules/gapi/src/executor/gasync.cpp
+++ b/modules/gapi/src/executor/gasync.cpp
@@ -4,10 +4,12 @@
 //
 // Copyright (C) 2019 Intel Corporation
 
+
 #include <opencv2/gapi/gcomputation_async.hpp>
 #include <opencv2/gapi/gcomputation.hpp>
 #include <opencv2/gapi/gcompiled_async.hpp>
 #include <opencv2/gapi/gcompiled.hpp>
+#include <opencv2/gapi/gasync_context.hpp>
 
 #include <condition_variable>
 
@@ -19,11 +21,11 @@
 namespace {
     //This is a tool to move initialize captures of a lambda in C++11
     template<typename T>
-    struct move_through_copy{
+    struct copy_through_move{
        T value;
-       move_through_copy(T&& g) : value(std::move(g)) {}
-       move_through_copy(move_through_copy&&) = default;
-       move_through_copy(move_through_copy const& lhs) : move_through_copy(std::move(const_cast<move_through_copy&>(lhs))) {}
+       copy_through_move(T&& g) : value(std::move(g)) {}
+       copy_through_move(copy_through_move&&) = default;
+       copy_through_move(copy_through_move const& lhs) : copy_through_move(std::move(const_cast<copy_through_move&>(lhs))) {}
     };
 }
 
@@ -80,6 +82,7 @@ public:
                 }};
             }
         }
+
         std::unique_lock<std::mutex> lck{mtx};
         bool first_task = q.empty();
         q.push(std::move(t));
@@ -108,8 +111,12 @@ async_service the_ctx;
 }
 
 namespace {
-template<typename f_t>
-std::exception_ptr call_and_catch(f_t&& f){
+template<typename f_t, typename context_t>
+std::exception_ptr call_and_catch(f_t&& f, context_t&& ctx){
+    if (std::forward<context_t>(ctx).isCanceled()){
+        return std::make_exception_ptr(GAsyncCanceled{});
+    }
+
     std::exception_ptr eptr;
     try {
         std::forward<f_t>(f)();
@@ -120,15 +127,21 @@ std::exception_ptr call_and_catch(f_t&& f){
     return eptr;
 }
 
-template<typename f_t, typename callback_t>
-void call_with_callback(f_t&& f, callback_t&& cb){
-    auto eptr = call_and_catch(std::forward<f_t>(f));
+struct DummyContext {
+    bool isCanceled() const {
+        return false;
+    }
+};
+
+template<typename f_t, typename callback_t, typename context_t>
+void call_with_callback(f_t&& f, callback_t&& cb, context_t&& ctx){
+    auto eptr =  call_and_catch(std::forward<f_t>(f), std::forward<context_t>(ctx));
     std::forward<callback_t>(cb)(eptr);
 }
 
-template<typename f_t>
-void call_with_futute(f_t&& f, std::promise<void>& p){
-    auto eptr = call_and_catch(std::forward<f_t>(f));
+template<typename f_t, typename context_t>
+void call_with_future(f_t&& f, std::promise<void>& p, context_t&& ctx){
+    auto eptr =  call_and_catch(std::forward<f_t>(f), std::forward<context_t>(ctx));
     if (eptr){
         p.set_exception(eptr);
     }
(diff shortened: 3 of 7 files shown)


modules/gapi/include/opencv2/gapi/gasync_context.hpp
```diff
@@ -0,0 +1,43 @@
...
+};
+
+class MY__EXPORTS GAsyncCanceled : public std::exception {
+public:
+    virtual const char* what() const noexcept CV_OVERRIDE {
+        return "GAPI asynchronous operation was canceled";
+    }
+};
+} // namespace gapi
+} // namespace wip
```
dbudniko: please change order should be wip then gapi

modules/gapi/test/gapi_async_test.cpp
```diff
@@ -262,13 +341,57 @@ TYPED_TEST_P(stress, test){
...
+    }
+
+    size_t canceled = 0 ;
+    for (auto&& r : requests){
+        try {
+            r.wait_for_result();
+        }catch (cv::gapi::wip::GAsyncCanceled& e){
+//        }catch (std::exception& e){
+            //operation was canceled
+            std::cout<< "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!\n"
```
smirnov-alexey: This must be removed I guess.

modules/gapi/src/executor/gasync.cpp
```diff
@@ -138,33 +150,71 @@ void call_with_futute(f_t&& f, std::promise<void>& p){
...
+        call_with_future(apply_l, prms.value, DummyContext{});
+    };
+
+    impl::the_ctx.add_task(l);
+    return f;
+}
+
+void async_apply(GAsyncContext& ctx, GComputation& gcomp, std::function<void(std::exception_ptr)>&& callback, GRunArgs &&ins, GRunArgsP &&outs, GCompileArgs &&args){
+    //TODO: use copy_through_move for all args except gcomp
+    auto l = [=, &ctx]() mutable {
```
smirnov-alexey: Probably can be moved somewhere to prevent code duplication.

modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
```diff
@@ -19,13 +19,17 @@ namespace cv {
...
 namespace gapi{
 namespace wip {
+    class GAsyncContext;
     //These functions asynchronously (i.e. probably on a separate thread of execution) call operator() member function of their first argument with copies of rest of arguments (except callback) passed in.
     //The difference between the function is the way to get the completion notification (via callback or a waiting on std::future object)
     //If exception is occurred during execution of apply it is transfered to the callback (via function parameter) or passed to future (and will be thrown on call to std::future::get)
-    GAPI_EXPORTS void                async(GCompiled& gcmpld, std::function<void(std::exception_ptr)>&& callback, GRunArgs &&ins, GRunArgsP &&outs);
-    GAPI_EXPORTS std::future<void>   async(GCompiled& gcmpld, GRunArgs &&ins, GRunArgsP &&outs);
-} // namespace gapi
+    GAPI_EXPORTS void                async(GAsyncContext& ctx, GCompiled& gcmpld, std::function<void(std::exception_ptr)>&& callback, GRunArgs &&ins, GRunArgsP &&outs);
```
dbudniko: I recall some discussions about GAsyncContext moved to the end as optional argument.

modules/gapi/src/executor/gasync.cpp
```diff
@@ -19,11 +20,11 @@
 namespace {
     //This is a tool to move initialize captures of a lambda in C++11
     template<typename T>
-    struct move_through_copy{
+    struct copy_through_move{
```
dbudniko: Is it somehow related to cancelation or just naming fix?

modules/gapi/test/gapi_async_test.cpp
```diff
@@ -262,13 +341,57 @@ TYPED_TEST_P(stress, test){
...
+                    <<e.what() << std::endl;
+            ++canceled;
+        }catch (...)
+        {
+
+        }
+    }
+    ASSERT_GT(canceled, size_t(0));
```
dbudniko: Just want to check that I'm understanding this corectly. Should it be not just greater than 0? Should it be equal to 99?

modules/gapi/test/gapi_async_test.cpp
```diff
@@ -186,6 +253,12 @@ struct AsyncCompiled  : crtp_cast<crtp_final_t>{
         auto gcmpld = this->crtp_cast_(this)->compile();
         return cv::gapi::wip::async(gcmpld, std::forward<Args>(args)...);
     }
+
+    template<typename... Args>
+    auto async(cv::gapi::wip::GAsyncContext& ctx, Args&&... args) -> decltype(cv::gapi::wip::async(std::declval<cv::gapi::wip::GAsyncContext&>(), std::declval<cv::GCompiled&>(), std::forward<Args>(args)...)){
```
dbudniko: Could you please divide this long expression on several lines? It's hard to review. This is also true for several similar instances in this test file below.

modules/gapi/test/gapi_async_test.cpp
andrey-golubev: tend to agree: `ASSERT_EQ(canceled, num_tasks - 1)`? is such behavior guaranteed?

modules/gapi/src/executor/gasync.cpp
```diff
@@ -120,15 +126,21 @@ std::exception_ptr call_and_catch(f_t&& f){
...
+void call_with_callback(f_t&& f, callback_t&& cb, context_t&& ctx){
+    auto eptr =  call_and_catch(std::forward<f_t>(f), std::forward<context_t>(ctx));
     std::forward<callback_t>(cb)(eptr);
 }
 
-template<typename f_t>
-void call_with_futute(f_t&& f, std::promise<void>& p){
-    auto eptr = call_and_catch(std::forward<f_t>(f));
+template<typename f_t, typename context_t>
+void call_with_futute(f_t&& f, std::promise<void>& p, context_t&& ctx){
```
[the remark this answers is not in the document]
anton-potapov: fixed

modules/gapi/include/opencv2/gapi/gasync_context.hpp
anton-potapov: fixed

modules/gapi/src/executor/gasync.cpp
anton-potapov: removed

modules/gapi/src/executor/gasync.cpp
anton-potapov: not related at all

modules/gapi/test/gapi_async_test.cpp
dbudniko: There are some pecularities in multithread cases. As far as I understand (num_tasks - num_threads) is the most correct answer.

anton-potapov: in the __current__ implementation (with a single background thread) it will be 99 - but it is an implementation detail. 

So best we can do is to make sure that at least some requests were cancelled

modules/gapi/test/gapi_async_test.cpp
anton-potapov: split

modules/gapi/test/gapi_async_test.cpp
anton-potapov: removed

modules/gapi/src/executor/gasync.cpp
anton-potapov: added TODO for now

modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
anton-potapov: 1. Having a default value for context parameter (and having it last), means that than there os a need to specify it, previous arguments with default parameters (mostly `GCompileArgs` for `async_apply`) have to assigned explicitly  which IMHO is inconvenient
2. Changing parameters order may break existing code, so (without extra effort) it is not an option
3. As for new - name I have two counterquestions : Why? and For example ?

andrey-golubev: 1. I find it inconvenient to support 2 interfaces (with && without context parameter) as there clearly could've been 1, but I guess there's no other obvious choice
3. It makes sense only if we decide to put "context" last: usually a context is something that other stuff depends on, so it is passed as a first parameter and called, well, a context. If it's optional then it's not really a context (which seems to always exist by definition), it's rather an additional (meta?-)information that may or may not be present.

Anyway, I'm totally fine with naming it `GAsyncContext` as long as we have it as a first parameter, just thinking that if it's last && optional, context is not really the name it should have

alalek:
1. Please use options for G-API standalone mode basic validation (see PR description). For all G-API PRs.
2. Builds should be get fixed.

/cc @dmatveev

modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
dmatveev: +1 to make it last. It is a common vision now (GVA asks for the same)

modules/gapi/test/gapi_async_test.cpp
dmatveev: `size_t(0)` or `size_t(0u)`?

modules/gapi/include/opencv2/gapi/gasync_context.hpp
```diff
@@ -0,0 +1,37 @@
...
+// Copyright (C) 2019 Intel Corporation
+
+#ifndef OPENCV_GAPI_GASYNC_CONTEXT_HPP
+#define OPENCV_GAPI_GASYNC_CONTEXT_HPP
+
+#if !defined(GAPI_STANDALONE)
+#  include <opencv2/core/cvdef.h>
+#else   // Without OpenCV
+#  include <opencv2/gapi/own/cvdefs.hpp>
+#endif // !defined(GAPI_STANDALONE)
```
dmatveev: I thought our model is just "always include own/cvdefs.hpp`" and there we have that `#ifdef` check. Isn't it?

modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
dmatveev: From our past notes:

>    - Default contexts - will be one per operation

Please implement the agreed solution.

modules/gapi/include/opencv2/gapi/gasync_context.hpp
```diff
@@ -0,0 +1,37 @@
...
+namespace cv {
+namespace gapi{
+namespace wip {
+
+class GAPI_EXPORTS GAsyncContext{
+    std::atomic<bool> cancelation_requested = {false};
+public:
+    bool cancel();
+    bool isCanceled() const;
+};
```
dmatveev: Documentation, please. What does `bool` return value mean for `cancel()`?

We also discussed that the context destructor should do cancellation (see mail notes). Is it implemented?

modules/gapi/include/opencv2/gapi/gasync_context.hpp
anton-potapov: `<opencv2/gapi/own/cvdefs.hpp>` has no `#include <opencv2/core/cvdef.h>` and never had IIRC.

modules/gapi/include/opencv2/gapi/gasync_context.hpp
anton-potapov: Added a comment on returning `bool`. 

As for "context destructor should do cancellation" - no it is not implemented and most likely will not be implemented. (Well, long story short  - it is a __bad__ idea which does not allow to achieve desired behaviour - will write a separate e-mail)

modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
anton-potapov: In fact, It is implemented - then no context is specified - operation uses devoted "dummy" context which is not exposed in any way. So, there is no way to cancel all operations that did not specified contexts explicitly.

modules/gapi/include/opencv2/gapi/gasync_context.hpp
dmatveev: well, probably it is worth adding then :) Not in this PR, for sure.

modules/gapi/include/opencv2/gapi/gasync_context.hpp
dmatveev: okay, let's discuss what's remaining

anton-potapov:
@dmatveev , ping 
Made  `GAsyncContext` argument to be last one

modules/gapi/include/opencv2/gapi/gasync_context.hpp
dmatveev: It is not a doxygen-capable documentation, unfortunately.

Please open an issue to document this API part properly (with correct doxygen documentation and an API sample in it - see our other public headers)

modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
```diff
@@ -12,20 +12,25 @@
...
+    class GAsyncContext;
     //These functions asynchronously (i.e. probably on a separate thread of execution) call operator() member function of their first argument with copies of rest of arguments (except callback) passed in.
     //The difference between the function is the way to get the completion notification (via callback or a waiting on std::future object)
     //If exception is occurred during execution of apply it is transfered to the callback (via function parameter) or passed to future (and will be thrown on call to std::future::get)
     GAPI_EXPORTS void                async(GCompiled& gcmpld, std::function<void(std::exception_ptr)>&& callback, GRunArgs &&ins, GRunArgsP &&outs);
+    GAPI_EXPORTS void                async(GCompiled& gcmpld, std::function<void(std::exception_ptr)>&& callback, GRunArgs &&ins, GRunArgsP &&outs, GAsyncContext& ctx);
+
     GAPI_EXPORTS std::future<void>   async(GCompiled& gcmpld, GRunArgs &&ins, GRunArgsP &&outs);
-} // namespace gapi
+    GAPI_EXPORTS std::future<void>   async(GCompiled& gcmpld, GRunArgs &&ins, GRunArgsP &&outs, GAsyncContext& ctx);
```
andrey-golubev: I thought the idea was to have a default context so that there's no interface duplication just because there's a context. Something similar to this:
```suggestion
    GAPI_EXPORTS std::future<void>   async(GCompiled& gcmpld, GRunArgs &&ins, GRunArgsP &&outs, GAsyncContext& ctx = {});
```

If you don't like the idea of getting `std::atomic<bool>` for no reason, let's put `DummyContext` to the API and call it `DefaultContext`

anton-potapov: context is passed in by reference , so having a default value for it means (without adding extra complexity, such shared internal state to the `GAsyncContext` itself) having a __global__ context object visible to everyone, thus ability to cancel __all__ async operations without explicit context, which is, presumably,  a bad idea.

andrey-golubev: Hmm, you're right. I wonder, however, if we're actually going to modify the state of the context within the call. Most likely, context object is read-only for us, thus we can use `const GAsyncContext& ctx` which can be initialized with a temporary. Am I wrong?

My concern is in having `std::atomic<bool>` for no reason in case of default context

anton-potapov: Long story short - for "default" context no atomics are ever created

andrey-golubev: True

Consider the comment resolved

PR #14700, opened 2019-06-03
G-API: Render

TolyaTalamanov:
```
```


TolyaTalamanov:
@dmatveev @dbudniko

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -0,0 +1,53 @@
...
+namespace cv
+{
+namespace gapi
+{
+
+struct TextEvent
+{
+    std::string text;
+    float x;
+    float y;
```
dbudniko: Maybe cv::Point2f is better?

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -0,0 +1,53 @@
...
+    double fs;
+    cv::Scalar color;
+    int thick;
+    int lt;
+    bool bottom_left_origin;
+};
+
+struct RectEvent
+{
+    float x;
```
dbudniko: Again, may be it's better to use native OpenCV data structures?

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: I think it is better not to make an explicit dependence on OpenCV.
In this case it is better to delete cv::Scalar.

modules/gapi/test/gapi_render_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+        cv::putText(ref_mat, text, cv::Point(pos_x, pos_y), ff, fs, cv::Scalar{color[0], color[1], color[2]}, thick);
+        events.emplace_back(cv::gapi::TextEvent{text, pos_x, pos_y, ff, fs, color, thick, lt, false});
+    }
+
+    cv::gapi::render(out_mat, events);
+
+    EXPECT_EQ(0, cv::countNonZero(out_mat != ref_mat));
+}
+
+TEST_F(RenderTestFixture, Rectangle)
```
dbudniko: I would suggest to add the third test with both render events

dbudniko: And one more test with unsupported draw event.

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -0,0 +1,54 @@
...
+GAPI_EXPORTS void render(cv::Mat& bgrx, const std::vector<DrawEvent>& events);
+
+// FIXME Specify the signature for NV12 case
+GAPI_EXPORTS void render(cv::Mat& y_plane, cv::Mat& uv_plane , const std::vector<DrawEvent>& events);
+
+} // namespace gapi
+} // namespace cv
+
+
+#endif // OPENCV_GAPI_RENDER_HPP
```
dbudniko: I would expect more comments in this file.

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <opencv2/gapi/opencv_includes.hpp>
+#include <opencv2/gapi/util/variant.hpp>
+#include <opencv2/gapi/own/exports.hpp>
+#include <opencv2/gapi/own/scalar.hpp>
+
+namespace cv
+{
+namespace gapi
+{
+
```
dbudniko: Should it be in some new "osd" namespace?

modules/gapi/test/gapi_render_test.cpp
TolyaTalamanov: Added test

dbudniko: BTW can we avoid hardcoded values directly in the test code (like 5, 30, 40, 60) and make them google test parameters?

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -0,0 +1,53 @@
...
+struct TextEvent
+{
+    std::string text;
+    int x;
+    int y;
+    int ff;
+    double fs;
+    cv::gapi::own::Scalar color;
```
dbudniko: May be it make sence to align rith other PRs and introduce GAPI_STANDALONE macro to enable both "part of OpenCV" version which uses native data structures and standalone version with G-API own structures?

TolyaTalamanov:
@dmatveev

modules/gapi/include/opencv2/gapi/render.hpp
dmatveev: This part can depend on OpenCV anyway. There's no plan to expose it in Standalone mode.

modules/gapi/include/opencv2/gapi/render.hpp
dmatveev: +1. makes sense to document fields of our DrawEvents in the way how those are documented in the original OpenCV counterparts.

modules/gapi/include/opencv2/gapi/render.hpp
dmatveev: let it be `draw`. `cv::gapi::draw::Something`. Nice, isn't it? :)

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <opencv2/gapi/util/variant.hpp>
+#include <opencv2/gapi/own/exports.hpp>
+#include <opencv2/gapi/own/scalar.hpp>
+
+namespace cv
+{
+namespace gapi
+{
+
+struct TextEvent
```
dmatveev: Y user should care about events, etc.

Let it be just
```
cv::gapi::draw::Text{ text, ... }
```

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -0,0 +1,53 @@
...
+    int y;
+    int ff;
+    double fs;
+    cv::gapi::own::Scalar color;
+    int thick;
+    int lt;
+    bool bottom_left_origin;
+};
+
+struct RectEvent
```
dmatveev: `cv::gapi::draw::Rect{...}`

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -0,0 +1,53 @@
...
+    int y;
+    int widht;
+    int height;
+    cv::gapi::own::Scalar color;
+    int thick;
+    int lt;
+    int shift;
+};
+
+using DrawEvent  = util::variant<TextEvent, RectEvent>;
```
dmatveev: `cv::gapi::draw::Prim`[itive]. Prim is ok

dmatveev:
IMHO this is a nice thing to start with. Please align the API, add remaining tests for NV12, and then we can start merging it.

modules/gapi/include/opencv2/gapi/render.hpp
dmatveev: Or may be even `cv::gapi::wip::draw` - to make it solid with async API.

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Replace to cv::Scalar

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Fixed

dmatveev: There are warnings in Doxygen now: https://pullrequest.opencv.org/buildbot/builders/precommit_docs/builds/20842

Also, `@param` is a way to document function arguments, not structure fields.

This is how fields are documented in doxygen: https://wiki.scilab.org/Doxygen%20documentation%20Examples

And this is how it looks now in the generated documentation (see above build):

image

modules/gapi/test/gapi_render_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+    for (int i = 0; i < 5; ++i)
+    {
+        cv::Point point {30 + i * 60, 40 + i * 50};
+        cv::Rect  rect {point, size};
+
+        cv::rectangle(ref_mat, rect, color, thick);
+        cv::putText(ref_mat, text, point, ff, fs, color, thick);
+
+        prims.emplace_back(cv::gapi::wip::draw::Rect{rect, color, thick, lt, 0});
+        prims.emplace_back(cv::gapi::wip::draw::Text{text, point, ff, fs, color, thick, lt, false});
```
dmatveev: This is so nice, isn't it? 😍

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -0,0 +1,100 @@
...
+@param thick Thickness of lines that make up the rectangle. Negative values, like #FILLED,
+mean that the function has to draw a filled rectangle.
+@param lt Type of the line. See #LineTypes
+@param shift Number of fractional bits in the point coordinates.
+ */
+GAPI_EXPORTS void render(cv::Mat& bgr, const Prims& prims);
+GAPI_EXPORTS void render(cv::Mat& y_plane, cv::Mat& uv_plane , const Prims& prims);
+
+// FIXME only for tests
+GAPI_EXPORTS void splitNV12TwoPlane(const cv::Mat& yuv, cv::Mat& y_plane, cv::Mat& uv_plane);
```
dmatveev: Tests-only functions should go from the public API headers

dmatveev: Tests can safely include internal headers, BTW

dmatveev: I didn't catch if this function itself it tested (e.g. for the conversion equality), I belive it makes sense to do so (maybe added via FIXME later)

modules/gapi/test/gapi_render_test.cpp
TolyaTalamanov: Thanks for comment, it is really needed, but I plan to do it later.

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Fixed

TolyaTalamanov:
@dmatveev

dmatveev:
Congrats!

PR #14705, opened 2019-06-03
3rdparty: update libjpeg 9c

alalek:
https://www.ijg.org/

relates #10274

**Note**: libjpeg-turbo is used by default. libjpeg is deprecated (use for compatibility reasons only)


alalek:
:+1:

Issue #13713, opened 2019-01-29
LUT3D() implementation not available


bonfiley:
The documentation here mentions LUT3D() but its implementation is not available in the latest code base.
https://docs.opencv.org/4.0.1/d6/d91/group__gapi__transform.html

dmatveev:
Thanks for noticing! Seems the API was introduced, but the kernel itself was dropped from the plan.
Will remove the API entry.

dmatveev:
We just dropped this kernel from the API (no plan to implement it so far - an artifact) so the issue now can be closed. Thanks!

PR #14720, opened 2019-06-04
G-API: Removed LUT3D API entry

rgarnov:
Fix for https://github.com/opencv/opencv/issues/13713


rgarnov:
There is an expected CI failure at ABI check

PR #14724, opened 2019-06-05
Fix some typos

asashour:
```
force_builders=Docs
```


alalek:
> 3rdparty

Please don't touch code in these directories.
It is got from upstream directories, so all changes would be lost during next update. Remove changes from this location.
If you want to get that fixed then feel free to contribute into upstream repositories.

PR #14735, opened 2019-06-06
java: upgrade tests to JUnit 4

asashour:
Upgrades the java tests to JUnit 4:

- [X] use `@Test` annotation, so there is no need for mandating method name to start with `test`
- [X] add `@NotYetImplemented` annotation, which allows writing a complete "expected to fail" test, without commenting the test code.
- [X] use `Assume`, instead of programmatically skipping the test
- [ ] add a test to ensure all tests have `@Test` annotation, in case someone later forgets it
- [ ] have a single `OpenCVTestCase ` as hinted [here](https://answers.opencv.org/question/213918/opencvtestcase-refactoring/)
- [ ] (optional) rename all test methods to remove the initial `test` prefix
- [ ] **validate**: Android Java tests

```
**WIP**
force_builders_only=Win64,Android armeabi-v7a,Docs
```


asashour:
I couldn't run the Android tests locally.

PR #14741, opened 2019-06-06
G-API: Aligned include syntax for public includes

rgarnov:
Changed quotes to branches when including public headers

```
```


dmatveev:
@alalek so what is the right thing, `""` or `<>`?

alalek:
In general, exact behavior implementation is compiler implementation-defined (not covered by standard).
GCC with `"..."` [additionally looks](https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html) into directory of current processed file (.cpp or other header), so relative paths works in that case.


I'm trying to follow these rules:
- `<...>` is used with "external" headers (like C++ runtime or external library). Also used in samples/tests (because library is kind of external entity). Should not be used with "relative" paths.
- `"..."` is used with headers from the same project/module/directory or relative paths (like OpenCV's `precomp.hpp`).

(probably got from [here](https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html))

---

More info:
- https://stackoverflow.com/questions/21593/what-is-the-difference-between-include-filename-and-include-filename

dmatveev:
@alalek so in other words we should use `""` instead of `<>` for own/OpenCV header includes?

alalek:
There is no strong policy from my side. In general, both variants should work fine.

dmatveev:
ok, thanks, what about standalone mode? :)
https://github.com/opencv/opencv/wiki/Graph-API

dmatveev:
:+1:

dmatveev:
Wrong button, sorry!

PR #14742, opened 2019-06-06
Add files via upload (fps)

### LINKED ISSUE #12810: cv2 VideoCapture doesn't work correctly when fps and tbr are different
I have a video with 25 fps and 50 tbr, when i capture video the output is not as expected,
Observations:
 1.set CV_CAP_PROP_POS_FRAMES to some position and start to capture, the captured frame won't match the actual frame (at that time) pos/time when i watch the video using vlc player
2.The number of frames captured is less than the frames captured directly using ffmpeg. Many frames at the start are missing
You can download the video from the specified link,
https://drive.google.com/open?id=1RPvNcaiMTNWTA8Pi5MOMT1i07FmPiFXR

How to simulate:
1.Dump all the frames using video capture from the beginning without using set position.

handler = cv2.VideoCapture()
fd = handler.open(filename)
 ret,frame=handler.read()
index = 0
while frame is not None:
	cv2.imwrite(str(index)+".jpg",frame)
	ret,frame=handler.read()
	index+=1

2.Set frame position to 5840 and try to match the frame dumped now and the previous one without set.

handler = cv2.VideoCapture()
fd = handler.open(filename)
ret,frame=handler.read()
handler.set(cv2.CAP_PROP_POS_FRAMES, 5340)
ret,frame=handler.read()
cv2.imwrite("using_set_"+str(5340)+".jpg",frame)

Output  [5340.jpg] is not same as [using_set_5340.jpg]

### LINKED ISSUE #14715: VideoCapture - print(frame) gives array values as 0

- OpenCV => 4.1.0
- Operating System / Platform => Windows 64 bit

##### Detailed description

```
import cv2
cap = cv2.VideoCapture(0)
while True:
    ret, frame = cap.read()
    #print(ret)
    print(frame)
```

It prints the 3 channel array values as zero.
I want to take the array values real time and process them into my function. 
I know that ret is True and that my camera is reading the values because when I print frame.sum(), it gives me the sum of array values but for some reason I can't see it in an array form and hence my function reads the frame as a set of 0's. 
Any help about how I can read the array values would be very useful. 

Thanks!

Sushanti99:
Resolves #12810
Resolves #14715

### This pullrequest changes
Addition of code to:
* read frame values correctly incase 0 is being returned 
* find the fps while working with images/videos

### DIFF
--- a/samples/python/fps.py
+++ b/samples/python/fps.py
@@ -11,6 +11,8 @@ fps = cap.get(cv2.CAP_PROP_FPS)
 
 ref = 60
 
+#for images
+
 while True:
 	start = time.time()
 
@@ -25,5 +27,16 @@ while True:
 	if k==27:
 		break
 
+'''
+# for videos
+while True:
+    ret, frame = cap.read()
+    fps = video.get(cv2.CAP_PROP_FPS)
+    print(fps)
+    k=cv2.waitKey(0)
+	if k==27:
+		break
+'''
+        
 cap.release()
 cv2.destroyAllWindows()
--- a/samples/python/fps.py
+++ b/samples/python/fps.py
@@ -1,0 +1,29 @@
+import cv2
+import time
+
+# cv2.VideoCapture(0) gives return as True 
+# but then print(frame) gives me an array filles with zeros
+
+# cv2.VideoCapture(cv2.CAP_DSHOW) works just fine in this case
+
+cap = cv2.VideoCapture(cv2.CAP_DSHOW)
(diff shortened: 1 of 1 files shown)


alalek:
Separate sample for that is not needed.

There are already other samples with measurement of processing time (FPS is not applicable for dedicated processing steps):
- https://github.com/opencv/opencv/blob/4.1.0/samples/python/facedetect.py#L67
- https://github.com/opencv/opencv/blob/4.1.0/samples/python/video_threaded.py
- etc

PR #14746, opened 2019-06-06
Fixed git versioning for multiple work trees

PhilLab:
The version is now derived from a git repository if that git repository is rooted at the source directory. This is to prevent problems when OpenCV is included inside of another git repository.

Resolves https://github.com/microsoft/vcpkg/issues/6708

Same technique as this [AWS SDK for C++ fix](https://github.com/aws/aws-sdk-cpp/pull/425/files#diff-198ad692bfaa08553bc0dc7b00415241)

### DIFF
--- a/cmake/OpenCVUtils.cmake
+++ b/cmake/OpenCVUtils.cmake
@@ -1763,7 +1763,7 @@ endif()
 
 macro(ocv_git_describe var_name path)
   if(GIT_FOUND)
-    execute_process(COMMAND "${GIT_EXECUTABLE}" describe --tags --exact-match --dirty
+    execute_process(COMMAND "${GIT_EXECUTABLE}" --work-tree="${path}" describe --tags --exact-match --dirty
       WORKING_DIRECTORY "${path}"
       OUTPUT_VARIABLE ${var_name}
       RESULT_VARIABLE GIT_RESULT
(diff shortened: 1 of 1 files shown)


cmake/OpenCVUtils.cmake
```diff
@@ -1763,15 +1763,15 @@ endif()
 
 macro(ocv_git_describe var_name path)
   if(GIT_FOUND)
-    execute_process(COMMAND "${GIT_EXECUTABLE}" describe --tags --exact-match --dirty
+    execute_process(COMMAND "${GIT_EXECUTABLE}" --work-tree="${path}" describe --tags --exact-match --dirty
       WORKING_DIRECTORY "${path}"
```
alalek: Why is `WORKING_DIRECTORY` not enough?

cmake/OpenCVUtils.cmake
```diff
@@ -1763,15 +1763,15 @@ endif()
 
 macro(ocv_git_describe var_name path)
   if(GIT_FOUND)
-    execute_process(COMMAND "${GIT_EXECUTABLE}" describe --tags --exact-match --dirty
+    execute_process(COMMAND "${GIT_EXECUTABLE}" --work-tree="${path}" describe --tags --exact-match --dirty
```
alalek: "Rooting" is not expected and breaks opencv_contrib at least:

```
--   Extra modules:
--     Location (extra):            /build/precommit_linux64/opencv_contrib/modules
--     Version control (extra):     unknown
```

Consider passing `OPENCV_VCSVERSION` value via CMake (command line or upstream scripts).

cmake/OpenCVUtils.cmake
PhilLab: Thanks, so https://github.com/microsoft/vcpkg/issues/6708 is mostly a vcpckg issue and has to be solved there

PhilLab:
https://github.com/microsoft/vcpkg/issues/6708 is (mostly) a vcpkg problem and has to be fixed there. Closing PR

PR #14757, opened 2019-06-07
G-API: Introduce new approach to write accuracy tests

### LINKED ISSUE #14795: Follow-up on G-API tests refactoring
hmm. If this is all about specifying some enum, y not
1. Use CRTP and define some new macro to define tests (macro would pass the derived type to `TestWithParamBase` as template argument. Something like
```c++
#define GAPI_TEST(Class, Signature)   struct Class: public TestWithParamBase<Class, std::tuple Signature >
```
2. These macros become
```
#define USE_INTERPOLATION(X) constexpr int INTERPOLATION = X;
```
3. In the base class you just refer to interpolation via CRTP (or to a default value if interpolation is not defined)

_Originally posted by @dmatveev in https://github.com/opencv/opencv/pull/14757_

andrey-golubev:
### This pullrequest

- Introduces new approach to write G-API accuracy tests:
  - No longer need to initialize parameters in each test (now it is automatic) => focus on business logic not on boilerplate
  - Unified common parameter names across all tests with the flexibility to support any specific parameters
- Aligns core G-API tests with the new approach
- resolves #14795


andrey-golubev:
@dmatveev @rgarnov @anton-potapov @TolyaTalamanov @smirnov-alexey please review. suggestions are welcome

modules/gapi/test/common/gapi_core_tests.hpp
```diff
@@ -99,56 +108,197 @@ struct PrintNormCoreParams
...
+    USE_UNIFORM_INIT(BitwiseTest);
+};
+struct NotTest           : public TestWithParamBase<>
+{
+    USE_UNIFORM_INIT(NotTest);
+};
+struct SelectTest        : public TestWithParamBase<>
+{
+    USE_UNIFORM_INIT(SelectTest);
+};
```
dmatveev: I believe test definitions like this may be one-liners:
```c++
struct SelectTest        : public TestWithParamBase<> { USE_UNIFORM_INIT(SelectTest); };
```
So you get 4x less SLOC for the majority of tests. Tests with multiple magic spells inside may remain multi-line/

BTW why do you need to pass test name to the macro? it looks redundant.

modules/gapi/test/common/gapi_tests_common.hpp
```diff
@@ -117,6 +142,142 @@ using compare_f = std::function<bool(const cv::Mat &a, const cv::Mat &b)>;
...
+    {
+        return m_params;
+    }
+};
+
+#define USE_UNIFORM_INIT(class_name) \
+    class_name() : TestWithParamBase(cv::RNG::UNIFORM) {}
+
+#define USE_NORMAL_INIT(class_name) \
+    class_name() : TestWithParamBase(cv::RNG::NORMAL) {}
```
dmatveev: hmm. If this is all about specifying some enum, y not
1. Use CRTP and define some new macro to define tests (macro would pass the derived type to `TestWithParamBase` as template argument. Something like
```c++
#define GAPI_TEST(Class, Signature)   struct Class: public TestWithParamBase<Class, std::tuple Signature >
```
2. These macros become
```
#define USE_INTERPOLATION(X) constexpr int INTERPOLATION = X;
```
3. In the base class you just refer to interpolation via CRTP (or to a default value if interpolation is not defined)

andrey-golubev: I thought about this. Probably, it would've been even simpler if `TestWithParamBase` accepted `int` as a first template argument:
```cpp
struct Class: public TestWithParamBase<INTERPOLATION, std::tuple Signature >
```
Anyway, I propose to leave it out for now and add separately (in the following PR)

modules/gapi/test/common/gapi_core_tests.hpp
andrey-golubev: Done

modules/gapi/test/common/gapi_core_tests_inl.hpp
```diff
@@ -1355,22 +1148,17 @@ TEST_P(LUTTest, AccuracyTest)
...
 {
-    int type_mat = -1, depth_to = -1;
-    double alpha = 1.0, beta = 0.0;
-    cv::Size sz_in;
-    cv::GCompileArgs compile_args;
-    compare_f cmpF;
-
-    std::tie(type_mat, depth_to, sz_in, alpha, beta, cmpF, compile_args) = GetParam();
-
+    auto param = GetParam();
```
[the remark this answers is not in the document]
andrey-golubev: thanks! fixed

modules/gapi/test/common/gapi_tests_common.hpp
dmatveev: ok!

modules/gapi/test/common/gapi_tests_common.hpp
```diff
@@ -62,25 +76,36 @@ class TestFunctional
...
+            initOutMats(sz_in, dtype);
+        }
+    }
+
+    void initRand(int type, cv::Size sz_in, int dtype, bool createOutputMatrices,
+        int distributionType = cv::RNG::UNIFORM)
+    {
+        switch( distributionType )
+        {
+            case cv::RNG::UNIFORM: return initMatsRandU(type, sz_in, dtype, createOutputMatrices);
```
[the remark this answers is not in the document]
andrey-golubev: @rgarnov good point, I propose to think about it separately

andrey-golubev:
@dmatveev @rgarnov please review final version and suggest if anything else should be done now or can be added as a part of the merge requests to follow

dmatveev:
Generally I am fine but what we have is [now](https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/21741):
```
test_gapi test_gapi ; cases (tests): 296 (30072) ; passed: 30072 ; skipped: 1598 ; ( 5 mins, 33 secs )
```
While it was used to be: ([1](https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/21712), [2](https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/21764)):
```
test_gapi test_gapi ; cases (tests): 296 (30218) ; passed: 30218 ; skipped: 1598 ; ( 2 mins, 32 secs )
test_gapi test_gapi ; cases (tests): 296 (30218) ; passed: 30218 ; skipped: 1598 ; ( 2 mins, 59 secs )
```

So seems master has more tests than this branch (= 146 test cases dropped) and now it works 2x slower?

modules/gapi/test/common/gapi_core_tests.hpp
```diff
@@ -123,99 +123,99 @@ struct PrintNormCoreParams
...
-struct MulDoubleTest     : public TestWithParamBase<> { USE_UNIFORM_INIT(MulDoubleTest); };
-struct DivTest           : public TestWithParamBase<> { USE_UNIFORM_INIT(DivTest); };
-struct DivCTest          : public TestWithParamBase<> { USE_UNIFORM_INIT(DivCTest); };
-struct MeanTest          : public TestWithParamBase<> { USE_UNIFORM_INIT(MeanTest); };
-struct MaskTest          : public TestWithParamBase<> { USE_UNIFORM_INIT(MaskTest); };
-struct Polar2CartTest    : public TestWithParamBase<> { USE_UNIFORM_INIT(Polar2CartTest); };
-struct Cart2PolarTest    : public TestWithParamBase<> { USE_UNIFORM_INIT(Cart2PolarTest); };
+    USE_UNIFORM_INIT(MathOpTest, false);
+};
+struct MulDoubleTest     : public TestWithParamBase<> { USE_UNIFORM_INIT(MulDoubleTest, true); };
```
dmatveev:
```diff
--- USE_UNIFORM_INIT(MulDoubleTest, true)
+++ USE_UNIFORM_INIT2(MulDoubleTest)
```

modules/gapi/test/common/gapi_core_tests.hpp
```diff
@@ -123,99 +123,99 @@ struct PrintNormCoreParams
...
+struct DivCTest          : public TestWithParamBase<> { USE_UNIFORM_INIT(DivCTest, true); };
+struct MeanTest          : public TestWithParamBase<> { USE_UNIFORM_INIT(MeanTest, true); };
+struct MaskTest          : public TestWithParamBase<> { USE_UNIFORM_INIT(MaskTest, true); };
+struct Polar2CartTest    : public TestWithParamBase<> { USE_UNIFORM_INIT(Polar2CartTest, false); };
+struct Cart2PolarTest    : public TestWithParamBase<> { USE_UNIFORM_INIT(Cart2PolarTest, false); };
 struct CmpTest           : public TestWithParamBase<CmpTypes,bool>
 {
     DEFINE_SPECIFIC_PARAMS_2(opType, testWithScalar);
-    USE_UNIFORM_INIT(CmpTest);
+    USE_UNIFORM_INIT(CmpTest, false);
```
dmatveev: Very invasive

modules/gapi/test/common/gapi_core_tests.hpp
andrey-golubev: @dmatveev you're looking at the old version. please see the latest one

modules/gapi/test/common/gapi_core_tests.hpp
andrey-golubev: @dmatveev you're looking at the old version. please see the latest one

andrey-golubev:
@dmatveev @rgarnov please review again

modules/gapi/test/common/gapi_core_tests_inl.hpp
```diff
@@ -1442,27 +1216,14 @@ TEST_P(SqrtTest, AccuracyTest)
...
-    double a = 0 , b = 0;
-    int norm_type = 0;
-    bool createOut = 0;
-    cv::GCompileArgs compile_args;
-
-    std::tie(cmpF, type, sz, a, b, norm_type, ddepth, createOut, compile_args) = GetParam();
-    int dtype = CV_MAKETYPE(ddepth, CV_MAT_CN(type));
-
-    initMatsRandN(type, sz, dtype, createOut);
+    initMatrixRandN(type, sz, CV_MAKETYPE(ddepth, CV_MAT_CN(type)), createOutputMatrices);
```
rgarnov: Maybe it's worth to introduce initialization function which initializes the output with type which have the same number of channels as input? (I found 3 tests which can make use of that)

andrey-golubev: I don't think this is critical at the moment. also, this would likely break the signature pattern of other initialization function and I'm not sure that's a good thing.

I'll look into this however, I guess there should be a better abstraction (there are other cases with quite complex initialization), ideally, we must be able to initialize everything before we get to the test body (e.g. in constructor). But I didn't find a good-looking API for that apart from replacing the whole `GAPI_TEST_FIXTURE` macro with manual test fixture creation (basically, doing more or less the same as the macro does)

andrey-golubev: by the way, we can actually the following in test fixture definition:
```cpp
GAPI_TEST_FIXTURE(NormalizeTest, [this] (int, cv::Size, int, bool) { initMatrixRandN(type, sz, CV_MAKETYPE(ddepth, CV_MAT_CN(type)), createOutputMatrices); }, ...)
```
this will basically do the same thing as we do in the test body but we now leverage a custom init function. so I think this is probably a way to go for such cases. alas, I think the API now looks terribly gigantic, may be it makes sense to simplify it somehow

I propose to address this separately

andrey-golubev:
@alalek please merge

andrey-golubev:
Thanks!

PR #14760, opened 2019-06-07
Rewrite code in BGR2Gray functions so that auto vectorizer does a better job.

ChipKerchner:
### This pullrequest changes

The auto vectorizer misses some opportunities on the BGR2Gray functions.  Rewriting so that compiler does a better job - 2x speed up on Power8/9, 5% on x64.


ChipKerchner:
Will this commit be eventually pushed to master or do I have to create another pull request?

alalek:
It will be propagated into master on this week.

PR #14764, opened 2019-06-08
core(intrin): drop hasSIMD128 checks

alalek:
- use compile-time checks instead (`#if CV_SIMD128`)
- runtime checks are useless


alalek:
:+1:

PR #14779, opened 2019-06-11
core: avoid function type cast, make happy UBSAN

alalek:
Message from UBSAN:

<cut/>

> modules/core/src/convert.dispatch.cpp:213:9: runtime error: call to function cv::opt_AVX2::cvtScale32f(float const*, unsigned long, unsigned char const*, unsigned long, float*, unsigned long, cv::Size_<int>, double*) through pointer to incorrect function type 'void (*)(const unsigned char *, unsigned long, const unsigned char *, unsigned long, unsigned char *, unsigned long, cv::Size_<int>, void *)'
modules/core/src/convert_scale.simd.hpp:272: note: cv::opt_AVX2::cvtScale32f(float const*, unsigned long, unsigned char const*, unsigned long, float*, unsigned long, cv::Size_<int>, double*) defined here

```
```

### DIFF
--- a/modules/core/src/convert.dispatch.cpp
+++ b/modules/core/src/convert.dispatch.cpp
@@ -244,12 +244,12 @@ void convertFp16(InputArray _src, OutputArray _dst)
         }
         else
             ddepth =  CV_16S;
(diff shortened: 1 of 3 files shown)


alalek:
:+1:

Issue #14580, opened 2019-05-18
Higher level API for DL networks.


dkurt:
Basically, this is a GSoC project which has not been applied this year: 

We want to see some kind of higher level API which can manage DL networks. There are several criterias for now:

* Hide asynchronous inference management internally. User just choose the number of requests (what about heterogenous scenario? In example, use both GPU and VPU with specific number of requestst for each other).
* Models from https://github.com/opencv/opencv/blob/master/samples/dnn/models.yml should be downloaded automatically in samples. Can we also introduce some kind of new automatically generated classes definitions for them? I.e.:

```cpp
CV_EXPORTS_W YOLOv3 : public Model {
  YOLOv3(const std::string& model = "yolov3.weights", const std::string& config= "yolov3.cfg", int weight = 416, int height = 416, Scalar mean = Scalar(), float scale = 0.00392);
};
```
User do not need to care about input's preprocessing. We can do it automatically. The base class can find the files (already implemented) or download if there are no (need to do).

```cpp
class Model : public Net {
  Model(const std::string& model, const std::string& config = "", int weight = -1, int height = -1, Scalar mean = Scalar(), float scale = 1.0);

  // Overridden version which accepts frame and can make blob.
  void setInput();
};
```

* We can add some methods to simplify postprocessing (i.e. detect type of output layer for detection network as we do in object detection samples):
```cpp
void detect(std::vector<float>& confidences, std::vector<Rect>& boxes, float confThreshold, float nmsThreshold);
```

So instead doing `forward` with postprocessing and can just call `detect` and then manage the predictions (i.e. draw it on the frame).

*

skhadem:
Managing the postprocessing seems awesome. Could be cool to be able to call detect() and pass a flag, which would automatically visualize outputs on an image by drawing the BB or applying the segmentation mask.
I have been working to try and contribute by getting comfortable with the code base, and would love to try and take this on. Do you have any additional suggestions for getting started? I have read all the guides and so far have been just going through code.

dkurt:
@skhadem, I also think that this is a good idea. Yeah, we also wanted to add a camera management here. Something like

```cpp
void Model::setVideoCapture(VideoCapture& cap, bool async = true);

void Mat getFrame();
```

You can call `getFrame` and `setInput` or just `detect` and the Model can just check if VideoCapture is opened so it's a source for inputs. The thing is that the best efficiency might be achieved with separate camera thread (see https://github.com/opencv/opencv/pull/14516 for example).

[...]

Issue #14730, opened 2019-06-05
dot delimiters for Python after namespaces


dkurt:
Current python bindings are generated with constructors for classes with `'_'` symbol after namespaces. It might be nice to have more convenient `'.'` delimiters as well (just aliases). In example, for now we have
```python
net = cv.dnn_Net()
```  
But there is no
```python
net = cv.dnn.Net()
```
Note: we need to keep `dnn_Net` for compatibility.

We need it for GSoC's data augmentation module because for now we designed with @FadyEssam basic transformations in the `cv::augment` namespace and the only way to manage transformations is:
```python
t = cv.augment_FlipVertical()  # but cv.augment.FlipVertical() looks better
```

PR #14780, opened 2019-06-11
Add high level API

l-bat:
Related: #14580

```

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
```


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,34 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
                              CV_OUT std::vector<int>& indices,
                              const float eta = 1.f, const int top_k = 0);
 
+     class CV_EXPORTS_W_SIMPLE Model : public Net
```
alalek: > Model

Please don't introduce GOD entities.
It is better to cover specific cases, like detection and/or classification with better configuration. With name of `DetectionModel` or similar.

Link: https://en.wikipedia.org/wiki/God_object

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,34 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+     {
+     public:
+         Model(const std::string& model, const std::string& config = "", int width = -1,
+               int height = -1, Scalar mean = Scalar(), float scale = 1.0,
+               bool swapRB = true, bool crop = false);
+
+         void setInputWidth(int width);
```
alalek: Constructor has too many parameter.
This would cause the same problems, like `blobFromImage()`.

> void setInputWidth

Return self reference instead of `void`:
> Model& setInputWidth()

It would allow to build configuration chains with well specified parameters names:

```
m = Model(loaded_network)
    .setParameter1(param1)
    .setParameter2(param2)
;
```

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,34 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+         void setInputHeight(int height);
```
alalek: Don't split width and height. It is a bad practice. Use `setInputSize(const Size& sz)` instead.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,34 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+         void setInputMean(Scalar mean);
+         void setInputScale(float scale);
+         void setInputCrop(bool crop);
+         void setInputSwapRB(bool swapRB);
+         std::pair<int, float> classify(const Mat& frame);
+         void detect(Mat& frame, std::vector<int>& classIds, std::vector<float>& confidences,
+                     std::vector<Rect2d>& boxes, float confThreshold, float nmsThreshold = 0.0);
```
alalek: > Mat& frame

`const refererence` for Inputs.
In case of `Mat` it should be `InputArray`

Outputs should be specified with `CV_OUT` for bindings generator.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,265 @@
...
+}
+
+std::pair<int, float> Model::classify(const Mat& frame) {
+    Mat blob = blobFromImage(frame, scale_, Size(width_, height_), mean_, swapRB_, crop_);
+    setInput(blob);
+    Mat out = forward();
+    std::vector<float> result = out.reshape(1, 1);
+    std::vector<float>::iterator iter = std::max_element(result.begin(), result.end());
+
+    std::pair<int, float> prediction(iter - result.begin(), *iter);
```
dkurt: Please use `cv::minMaxLoc`.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,265 @@
...
+    return prediction;
+}
+
+void Model::detect(Mat& frame, std::vector<int>& classIds,
+                   std::vector<float>& confidences, std::vector<Rect2d>& boxes,
+                   float confThreshold, float nmsThreshold)
+{
+    Mat blob = blobFromImage(frame, scale_, Size(width_, height_), mean_, swapRB_, crop_);
+    setInput(blob);
```
dkurt: For Myriad target it's better to create uint8 blob and pass preprocessing parameters in `setInput`. So let's do the same in general case.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,265 @@
...
+    if (getLayer(0)->outputNameToIndex("im_info") != -1) {  // Faster-RCNN or R-FCN
+        resize(frame, frame, Size(width_, height_));
+        Mat imInfo = (Mat_<float>(1, 3) << frame.rows, frame.cols, 1.6f);
+        setInput(imInfo, "im_info");
+    }
+
+    std::vector<Mat> detections;
+    forward(detections);
```
dkurt: This method is https://docs.opencv.org/master/db/d30/classcv_1_1dnn_1_1Net.html#a8d248f50efec70dffa785de18a4b2170 but we want to use https://docs.opencv.org/master/db/d30/classcv_1_1dnn_1_1Net.html#adb34d7650e555264c7da3b47d967311b to get outputs from different layers (i.e. YOLOv3)

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,265 @@
...
+                }
+            }
+        }
+    } else if (lastLayer->type == "Region") {
+        for (int i = 0; i < detections.size(); ++i) {
+            for (int j = 0; j < detections[i].rows; ++j) {
+
+                Mat slice = detections[i].row(j).colRange(5, detections[i].cols);
+                std::vector<float> scores = slice.reshape(1, 1);
+                std::vector<float>::iterator classId = std::max_element(scores.begin(), scores.end());
```
dkurt: Please use `cv::minMaxLoc`. Otherwise there is extra memory copy at `std::vector<float> scores = slice.reshape(1, 1);`.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,265 @@
...
+    confidences.reserve(indices.size());
+    classIds.reserve(indices.size());
+
+    for (int idx : indices)
+    {
+        boxes.push_back(predBoxes[idx]);
+        confidences.push_back(predConf[idx]);
+        classIds.push_back(predClassIds[idx]);
+    }
+}
```
dkurt: Please avoid code duplication.

modules/dnn/include/opencv2/dnn/dnn.hpp
dkurt: As the advantage, we wanted to hide asynchronous inference mangement into the Model class. So would it be a good idea to keep `dnn::Model` with some method like `predict` which works in the way like `forward` but additionally performs preprocessing and can use VideoCapture as a source of inputs.

dkurt: @l-bat, Please create a separate `DetectionModel` class for object detection networks. It'll be inherited from `Model` and has just `detect` method. For `Model` we'll use something like `predict(const Mat& frame)`

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,194 @@
...
+    if (lastLayer->type == "DetectionOutput") {
+        for (int i = 0; i < detections.size(); ++i)
+        {
+            for (int j = 0; j < detections[i].size[2]; j++)
+            {
+                std::vector<Range> coord = { Range::all(), Range::all(), Range(j, j + 1), Range::all() };
+                Mat detect;
+                detections[i](coord).copyTo(detect);
+                std::vector<float> detection = detect.reshape(1, 1);
+                float conf = detection[2];
```
dkurt: Please use `float* data = detections[i].data` (as in https://github.com/opencv/opencv/blob/master/samples/dnn/object_detection.cpp sample). `std::vector<float> detection = detect.reshape(1, 1);` is an extra data copy.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,194 @@
...
+                if (conf > confThreshold)
+                {
```
dkurt: Please refactor it in this way:
```
if (conf < confThreshold)
    continue;

// process
```

modules/dnn/include/opencv2/dnn/dnn.hpp
dkurt: @l-bat, Please remove preprocessing params from constructors as @alalek pointed.

modules/dnn/include/opencv2/dnn/dnn.hpp
dkurt: I think we can also have something like void `setInputSize(int width, int height)`. Just more friendly for Python:

```python
model.setInputSize(width=300, height=300)
```
then
```python
model.setInputSize((300, 300))
```

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,194 @@
...
+        for (int i = 0; i < detections.size(); ++i) {
+            for (int j = 0; j < detections[i].rows; ++j) {
+
+                Mat slice = detections[i].row(j).colRange(5, detections[i].cols);
+
+                double max;
+                cv::Point maxLoc;
+                minMaxLoc(slice, nullptr, &max, nullptr, &maxLoc);
+
+                float conf  = static_cast<float>(max);
```
dkurt: That's confusing. Why not just use `conf` from the beginning?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,30 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+
+         Model& setInputSize(const Size& size);
+         Model& setInputMean(const Scalar& mean);
+         Model& setInputScale(float scale);
+         Model& setInputCrop(bool crop);
+         Model& setInputSwapRB(bool swapRB);
+         std::pair<int, float> classify(InputArray frame);
+         void detect(InputArray frame, CV_OUT std::vector<int>& classIds,
+                     CV_OUT std::vector<float>& confidences, CV_OUT std::vector<Rect2d>& boxes,
+                     float confThreshold, float nmsThreshold = 0.4, bool absoluteCoords = true);
```
dkurt: As one of the goals for this PR is to decrease number of actions made by user to use DL networks. Currently this method returns detections in relative coordinates, right? That means in the most of scenarios users should multiple them by frame's sizes. So in my point of view we need to return `cv::Rect` with pixel coords.

l-bat: if  `absoluteCoords = true` - absolute coordinates (default), else relative coordinates

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -999,27 +999,34 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
-         Model(const std::string& model, const std::string& config = "", const Size& size = {-1, -1},
-               const Scalar& mean = {}, float scale = 1.0, bool swapRB = true, bool crop = false);
-
-         Model(const Net& network, const Size& size = {-1, -1},
-               const Scalar& mean = {}, float scale = 1.0, bool swapRB = true, bool crop = false);
+         Model(const std::string& model, const std::string& config = "");
+         Model(const Net& network);
 
          Model& setInputSize(const Size& size);
+         Model& setInputSize(int height, int width);
```
dkurt: That might be the source of confusion because `cv::Size` has constructor with `width, height` order: https://docs.opencv.org/master/d6/d50/classcv_1_1Size__.html#a45c97e9a4930d73fde11c2acc5f371ac

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -999,27 +999,34 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
          Model& setInputMean(const Scalar& mean);
          Model& setInputScale(float scale);
          Model& setInputCrop(bool crop);
          Model& setInputSwapRB(bool swapRB);
          std::pair<int, float> classify(InputArray frame);
```
dkurt: And please move also `classify` to something like `ClassificationModel`.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -999,27 +999,34 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
 
+     class CV_EXPORTS_W_SIMPLE DetectionModel : public Model
+     {
+     public:
+         DetectionModel(const std::string& model, const std::string& config = "");
+         DetectionModel(const Net& network);
+
+         void detect(InputArray frame, CV_OUT std::vector<int>& classIds,
+                     CV_OUT std::vector<float>& confidences, CV_OUT std::vector<Rect2d>& boxes,
+                     float confThreshold, float nmsThreshold = 0.4, bool absoluteCoords = true);
```
dkurt: Is there default value for `confThreshold`? You may use one form samples.

modules/dnn/src/model.cpp
```diff
@@ -20,23 +20,26 @@ namespace dnn {
 struct Model::Impl {
     Size   size;
     Scalar mean;
-    float  scale;
-    bool   swapRB;
-    bool   crop;
+    float  scale = 1.0;
+    bool   swapRB = true;
```
dkurt: By default, preprocessing all the preprocessing stages should be disabled. So please turn `swapRB` to false by default.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,37 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+         Model& setInputSwapRB(bool swapRB);
+         std::pair<int, float> classify(InputArray frame);
+         std::vector<Mat> predict(const Mat& frame);
+
+     protected:
+         struct Impl;
+         Ptr<Impl> impl_;
+     };
+
+     class CV_EXPORTS_W_SIMPLE DetectionModel : public Model
```
dkurt: I think that we need to use `CV_EXPORTS_W` because `CV_EXPORTS_W_SIMPLE` is mostly used for some data structures but we have pretty complicated logic inside our new classes.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,210 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+// Copyright (C) 2018-2019, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective owners.
```
dkurt: Please use simplified version of the header without years:
```
```

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,210 @@
...
+std::vector<Mat> Model::predict(const Mat& frame) {
+    Mat blob;
+    blobFromImage(frame, blob, 1.0, impl_->size, Scalar(), impl_->swapRB, impl_->crop, CV_8U);
+    setInput(blob, "", impl_->scale, impl_->mean);
+
+    std::vector<String> outNames = getUnconnectedOutLayersNames();
+    std::vector<Mat> outs;
+    forward(outs, outNames);
+    return outs;
+}
```
dkurt: `blob` and `outNames` could be a part of `Impl`. BTW, what do you think about moving an entire method to Impl?

```cpp
std::vector<Mat> Model::predict(const Mat& frame) {
    impl->predict(frame);
}
```

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,212 @@
...
+struct Model::Impl {
+    Size   size;
+    Scalar mean;
+    float  scale = 1.0;
+    bool   swapRB = false;
+    bool   crop = false;
+    Mat    blob;
+    std::vector<String> outNames;
+
+    void predict(Net* net, const Mat& frame, std::vector<Mat> outs) {
```
dkurt:
```cpp
void predict(Net& net, const Mat& frame, std::vector<Mat>& outs)
```
is better

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,212 @@
...
+    return *this;
+}
+
+Model& Model::setInputSwapRB(bool swapRB) {
+    impl_->swapRB = swapRB;
+    return *this;
+}
+
+void Model::predict(InputArray frame, OutputArray outs) {
+    impl_->predict(this, frame.getMat(), outs.getMat());
```
dkurt: `outs` should be a vector of `cv::Mat`. So we need to use `OutputArrayOfArrays`

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+ClassificationModel::ClassificationModel(const std::string& model, const std::string& config)
+    : Model(model, config) {};
+
+ClassificationModel::ClassificationModel(const Net& network) : Model(network) {};
+
+std::pair<int, float> ClassificationModel::classify(InputArray frame) {
+    blobFromImage(frame, impl_->blob, 1.0, impl_->size, Scalar(), impl_->swapRB, impl_->crop, CV_8U);
+    setInput(impl_->blob, "", impl_->scale, impl_->mean);
+    Mat out = forward();
```
dkurt: We need to use `Impl::predict` instead.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
+                   CV_OUT std::vector<float>& confidences, CV_OUT std::vector<Rect2d>& boxes,
+                   float confThreshold, float nmsThreshold, bool absoluteCoords)
+{
+    int frameWidth  = frame.cols();
+    int frameHeight = frame.rows();
+
+    if (impl_->size.width <= 0) impl_->size.width = frame.cols();
+    if (impl_->size.height <= 0) impl_->size.height = frame.rows();
```
dkurt: We need to think about three options about input sizes:

1) blob size is specified
1) blob size == frame size
3) blob size is from model (not implemented yet).

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,45 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+
+     class CV_EXPORTS_W DetectionModel : public Model
+     {
+     public:
+         DetectionModel(const std::string& model, const std::string& config = "");
+         DetectionModel(const Net& network);
+
+         void detect(InputArray frame, CV_OUT std::vector<int>& classIds,
+                     CV_OUT std::vector<float>& confidences, CV_OUT std::vector<Rect2d>& boxes,
+                     float confThreshold = 0.5, float nmsThreshold = 0.4, bool absoluteCoords = true);
```
dkurt: Please add a Python test as well (see https://github.com/opencv/opencv/blob/3.4/modules/dnn/misc/python/test/test_dnn.py).

And additionally please invoke `cv.rectangle` to draw resulting detections on the frame so we can check what rectangles are wrapped correctly.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,212 @@
...
+namespace dnn {
+
+struct Model::Impl {
+    Size   size;
+    Scalar mean;
+    float  scale = 1.0;
+    bool   swapRB = false;
+    bool   crop = false;
+    Mat    blob;
+    std::vector<String> outNames;
```
dkurt: Maybe we can add a constructor which fills output layers names by `getUnconnectedOutLayersNames` once? Because now it's used every time.

dkurt: We can also store output blobs vector here.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,212 @@
...
+    boxes.reserve(indices.size());
+    confidences.reserve(indices.size());
+    classIds.reserve(indices.size());
+
+    for (int idx : indices)
+    {
+        boxes.push_back(predBoxes[idx]);
+        confidences.push_back(predConf[idx]);
+        classIds.push_back(predClassIds[idx]);
+    }
```
dkurt: We need to clean the vectors before or use `resize`.

modules/dnn/src/model.cpp
l-bat: Need to add assert?

modules/dnn/src/model.cpp
l-bat: done

modules/dnn/src/model.cpp
l-bat: done

modules/dnn/src/model.cpp
l-bat: done

modules/dnn/src/model.cpp
l-bat: done

modules/dnn/src/model.cpp
l-bat: done

modules/dnn/src/model.cpp
dkurt: For now `getUnconnectedOutLayersNames` is called at `predict` what means we call it every iteration. However we can just initialize all the output layers names in the Impl constructor.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,225 @@
...
+                minMaxLoc(scores, nullptr, &conf, nullptr, &classIdPoint);
+
+                if (conf < confThreshold)
+                    continue;
+
+                std::vector<Range> coord = {Range(j, j + 1), Range(0, 5)};
+                Mat coords;
+                detections[i](coord).copyTo(coords);
+
+                std::vector<float> bboxes = coords.reshape(1, 1);
```
dkurt: Please refactor "Region" output layer postprocessing as well. Use `float*` instead of vectors.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,45 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+         Ptr<Impl> impl_;
+     };
+
+     class CV_EXPORTS_W ClassificationModel : public Model
+     {
+     public:
+         ClassificationModel(const std::string& model, const std::string& config = "");
+         ClassificationModel(const Net& network);
+
+         std::vector<std::pair<int, float>> classify(InputArray frame);
```
dkurt: How we can interpret multiple outputs from classification model?

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,225 @@
...
+        setInput(imInfo, "im_info");
+    }
+
+    boxes.clear();
+    confidences.clear();
+    classIds.clear();
+
+    std::vector<Mat> detections;
+    impl_->outNames = getUnconnectedOutLayersNames();
+    forward(detections, impl_->outNames);
```
dkurt: We need to use `predict` method here.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,45 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
                              CV_OUT std::vector<int>& indices,
                              const float eta = 1.f, const int top_k = 0);
 
+     class CV_EXPORTS_W Model : public Net
```
dkurt: Please document new class and it's methods (see other classes in this header for doxygen syntax reference).

modules/dnn/src/model.cpp
dkurt: Not actually. I mean that it could be useful to have a feature with automatic input blob size detection from the model. However it won't work for all the models because some formats have no information about input sizes.

So I agree with you. Let's add an assertion if `width` or `height` weren't specified.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,45 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+         Model& setInputSize(int width, int height);
+         Model& setInputMean(const Scalar& mean);
+         Model& setInputScale(float scale);
+         Model& setInputCrop(bool crop);
+         Model& setInputSwapRB(bool swapRB);
+         void predict(InputArray frame, OutputArrayOfArrays outs);
+
+     protected:
+         struct Impl;
+         Ptr<Impl> impl_;
```
dkurt: Maybe it's better to name it as `impl`? Like for dnn::Net

modules/dnn/include/opencv2/dnn/dnn.hpp
l-bat: Impl::predict return std::vector<Mat>& outs.Do we need to check that vector contains 1 element?

modules/dnn/src/model.cpp
l-bat: Move to Impl::predict?
```
if (getLayer(0)->outputNameToIndex("im_info") != -1) {  // Faster-RCNN or R-FCN
    frameWidth  = impl_->size.width;
    frameHeight = impl_->size.height;
    Mat imInfo = (Mat_<float>(1, 3) << impl_->size.height, impl_->size.width, 1.6f);
    setInput(imInfo, "im_info");
}
```
How to change frame size: `frameWidth  = impl_->size.width;  frameHeight = impl_->size.height;`?

modules/dnn/src/model.cpp
l-bat: Maybe add 
```
if (size.width <= 0) size.width = frame.cols;
if (size.height <= 0) size.height = frame.rows;
```
to Impl::predict(Net& net, const Mat& frame, std::vector<Mat>& outs)?

modules/dnn/src/model.cpp
l-bat: Maybe initialize all the output layers names in Model constructor?

modules/dnn/src/model.cpp
l-bat: done

modules/dnn/include/opencv2/dnn/dnn.hpp
l-bat: done

modules/dnn/include/opencv2/dnn/dnn.hpp
l-bat: done

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,236 @@
...
+    Scalar mean;
+    float  scale = 1.0;
+    bool   swapRB = false;
+    bool   crop = false;
+    Mat    blob;
+    std::vector<String> outNames;
+
+    void predict(Net& net, const Mat& frame, std::vector<Mat>& outs) {
+        if (size.width <= 0) size.width = frame.cols;
+        if (size.height <= 0) size.height = frame.rows;
```
dkurt: Let's for now ask user to provide input blob size in any case. So please throw an assertion if input size is now specified.

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -53,6 +53,35 @@ def normAssertDetections(test, ref, out, confThreshold=0.0, scores_diff=1e-5, bo
...
+                    matched = True
+                    matchedRefBoxes[j] = True
+        if not matched:
+            errMsg += '\nUnmatched prediction: class %d score %f box %s' % (testClassId, testScore, box2str(testBox))
+
+    for i in range(len(refBoxes)):
+        if (not matchedRefBoxes[i]) and refScores[i] > confThreshold:
+            errMsg += '\nUnmatched reference: class %d score %f box %s' % (refClassIds[i], refScores[i], box2str(refBoxes[i]))
+    if errMsg:
+        test.fail(errMsg)
```
dkurt: Please avoid code duplication. Choose common parts with existing `normAssertDetections`. BTW there is no functions overloading in Python that means by adding second `def normAssertDetections` you replaces the first one.

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -135,6 +164,52 @@ def test_blobFromImage(self):
         target = target.transpose(2, 0, 1).reshape(1, 3, height, width)  # to NCHW
         normAssert(self, blob, target)
 
+    def test_model(self):
+        img_path = self.find_dnn_file("dnn/dog416.png")
+        weights = self.find_dnn_file("dnn/resnet50_rfcn_final.caffemodel")
+        config = self.find_dnn_file("dnn/rfcn_pascal_voc_resnet50.prototxt")
```
dkurt: Please choose lighter model. R-FCN is too heavy.

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -135,6 +164,52 @@ def test_blobFromImage(self):
...
+        frame = cv.imread(img_path)
+        model = cv.dnn.DetectionModel_create(weights, config)
```
dkurt: Can you check that `cv.dnn_DetectionModel(weights, config)` also works? I hope that with https://github.com/opencv/opencv/issues/14730 resolved we can even replace it to `cv.dnn.DetectionModel(weights, config)`.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,173 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+
+         /** @brief Set preprocessing parameters for frame.
+         *  @param[in] width New input width.
+         *  @param[in] height New input height.
+         *  @param[in] mean Scalar with mean values which are subtracted from channels.
+         *  @param[in] scale Multiplier for frame values.
+         *  @param[in] swapRB Flag which indicates that swap first and last channels.
+         *  @param[in] crop Flag which indicates whether image will be cropped after resize or not.
+         */
+         CV_WRAP void setParams(int width = -1, int height = -1, Scalar mean = Scalar(),
```
dkurt: Maybe `setInputParams`?

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -135,6 +164,52 @@ def test_blobFromImage(self):
...
+                    (132.093, 223.903, 206.984, 343.257))
+
+        normAssertDetections(self, refClassIds, refConfidences, refBoxes,
+                             classIds, confidences, boxes,
+                             confThreshold, scoreDiff, iouDiff)
+
+        frame = cv.resize(frame, size)
+        for box, conf, classId in zip(boxes, confidences, classIds):
+            left, top, width, height = box
+            left = int(left)
```
dkurt: I cannot think up an application where we really need bounding box coordinates in double precision. So for now we have a method which produces `cv::Rect2d` that means that [to draw](https://docs.opencv.org/master/d6/d6e/group__imgproc__draw.html#gac865734d137287c0afb7682ff7b3db23) these rectangle on the image or even [crop a ROI](https://docs.opencv.org/master/d3/d63/classcv_1_1Mat.html#a4272fce43d7405603fbb5e5f52580e57) we need to cast it to `cv::Rect` (which is `cv::Rect2i`).

So  think that we need to remove `absoluteCoords` flag at `Model::detect` and return bounding boxes in the absolute coordinates with integer data type. What do you think?

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -135,6 +164,52 @@ def test_blobFromImage(self):
...
+            assert(left + width < frame.shape[1])
+            assert(top + height < frame.shape[0])
+            cv.rectangle(frame, (left, top), (left + width, top + height), (0, 255, 0))
+
+            label = '%s: %.2f' % (classId, conf)
+
+            labelSize, baseLine = cv.getTextSize(label, cv.FONT_HERSHEY_SIMPLEX, 0.5, 1)
+            top = max(top, labelSize[1])
+            cv.rectangle(frame, (left, top - labelSize[1]), (left + labelSize[0], top + baseLine), (255, 255, 255), cv.FILLED)
+            cv.putText(frame, label, (left, top), cv.FONT_HERSHEY_SIMPLEX, 0.5, (0, 0, 0))
```
dkurt: Just the first `cv.rectangle` is enough. I just wanted to point that we need to cast coords to ints.

modules/dnn/misc/python/test/test_dnn.py
l-bat: `cv.dnn_DetectionModel(weights, config)` doesn't work.  
 `<class 'cv2.dnn_DetectionModel'> returned NULL without setting an error`
And `cv.dnn.DetectionModel(weights, config)` doesn't work too.  
`AttributeError: module 'cv2.dnn' has no attribute 'DetectionModel'`

modules/dnn/src/model.cpp
dkurt: @l-bat, also good. At least we need to do it one.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,229 @@
...
+            }
+        }
+    } else if (lastLayer->type == "Region") {
+        for (int i = 0; i < detections.size(); ++i) {
+            // Network produces output blob with a shape NxC where N is a number of
+            // detected objects and C is a number of classes + 4 where the first 4
+            // numbers are [center_x, center_y, width, height]
+            float* data = (float*)detections[i].data;
+            for (int j = 0; j < detections[i].rows; ++j, data += detections[i].cols) {
+
```
dkurt: Very strange line break. What about 
```cpp
for (int j = 0; j < detections[i].rows; ++j, data += detections[i].cols)
{
    // ...
```
?

Please also fix it in other places. We mostly use opening parentheses on separate lines.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,229 @@
...
+                int left    = centerX - width  / 2;
+                int top     = centerY - height / 2;
+
+                predClassIds.push_back(classIdPoint.x);
+                predConf.push_back(static_cast<float>(conf));
+                predBoxes.emplace_back(left, top, width, height);
+            }
+        }
+    } else {
+        CV_Error(Error::StsError, "Unknown output layer type: \"" + lastLayer->type + "\"");
```
dkurt: StsNotImplemented is better I guess.

modules/dnn/include/opencv2/dnn/dnn.hpp
dkurt: Sure. We can have multiple outputs from detection networks because there are could be more than one object and in example YOLOv3 produces outputs for different objects sizes. However there is one and only prediction for classification problem.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,229 @@
...
+    Size   size;
+    Scalar mean;
+    float  scale = 1.0;
+    bool   swapRB = false;
+    bool   crop = false;
+    Mat    blob;
+    std::vector<String> outNames;
+
+    void predict(Net& net, const Mat& frame, std::vector<Mat>& outs) {
+        CV_Assert(!size.empty());
```
dkurt: Let's make it more informative. Throw a text error message.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,171 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+          *  @param[in] crop Flag which indicates whether image will be cropped after resize or not.
+         */
+         Model& setInputCrop(bool crop);
+
+         /** @brief Set flag swapRB for frame.
+          *  @param[in] swapRB Flag which indicates that swap first and last channels.
+         */
+         Model& setInputSwapRB(bool swapRB);
+
+         /** @brief Set preprocessing parameters for frame.
```
dkurt: Please try to describe the order of preprocessing procedures. Ine example, by Latex formula like `blob(n, c, y, x) = scale * resize( frame(y, x, c) ) - mean(c) )`

modules/dnn/misc/python/test/test_dnn.py
dkurt: What if add `CV_WRAP` for `DetectionModel` constructors?

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -135,6 +143,41 @@ def test_blobFromImage(self):
...
+        weights = self.find_dnn_file("dnn/MobileNetSSD_deploy.caffemodel")
+        config = self.find_dnn_file("dnn/MobileNetSSD_deploy.prototxt")
+        frame = cv.imread(img_path)
+        model = cv.dnn.DetectionModel_create(weights, config)
+        size = (300, 300)
+        mean = (127.5, 127.5, 127.5)
+        scale = 1.0 / 127.5
+        model.setInputParams(size=size, mean=mean, scale=scale)
+
+        iouDiff = 1
```
dkurt: Intersection of Union is a value in range [0, 1]. Is that OK to have 1 diff?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,171 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+         *  @param[in] crop Flag which indicates whether image will be cropped after resize or not.
+         */
+         CV_WRAP void setInputParams(const Size& size = Size(), const Scalar& mean = Scalar(),
+                               float scale = 1.0, bool swapRB = false, bool crop = false);
+
+         /** @brief Given the @p input frame, create input blob, run net and return the output @p blobs.
+          *  @param[in]  frame  The input image.
+          *  @param[out] outs Allocated output blobs, which will store results of the computation.
+          */
+         CV_WRAP void predict(InputArray frame, CV_OUT OutputArrayOfArrays outs);
```
dkurt: For OutputArrays and other arrays you can emit CV_OUT: https://docs.opencv.org/master/da/d49/tutorial_py_bindings_basics.html

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,229 @@
...
+    return *this;
+}
+
+Model& Model::setInputSwapRB(bool swapRB) {
+    impl->swapRB = swapRB;
+    return *this;
+}
+
+void Model::setInputParams(const Size& size, const Scalar& mean,
+                           float scale, bool swapRB, bool crop) {
```
dkurt: Let's order the parameters as in blobFromImage: `scale, size, mean, swapRB, crop`.  We already have a separate method if user need to just set input size so we can keep the order.

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -135,6 +143,41 @@ def test_blobFromImage(self):
...
+                    (101.638, 189.152, 34.217, 138.234))
+
+        assertDetect(self, refClassIds, refConfidences, refBoxes,
+                             classIds, confidences, boxes,
+                             confThreshold, scoreDiff, iouDiff)
+
+        for box, conf, classId in zip(boxes, confidences, classIds):
+            left, top, width, height = box
+            assert(left > 0)
+            assert(top > 0)
```
dkurt: left and top might be zero as well, aren't they?

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+std::pair<int, float> ClassificationModel::classify(InputArray frame) {
+    std::vector<Mat> outs;
+    impl->predict(*this, frame.getMat(), outs);
+    CV_Assert(outs.size() == 1);
+
+    double conf;
+    cv::Point maxLoc;
+    minMaxLoc(outs[0], nullptr, &conf, nullptr, &maxLoc);
```
dkurt: I think that we need to reshape `outs[0]` because it might has more than two dimensions and `minMaxLoc` will fail:

```cpp
minMaxLoc(outs[0].reshape(1, 1), nullptr, &conf, nullptr, &maxLoc);
```

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+DetectionModel::DetectionModel(const String& model, const String& config)
+    : Model(model, config) {};
+
+DetectionModel::DetectionModel(const Net& network) : Model(network) {};
+
+Ptr<DetectionModel> DetectionModel::create(const String& model, const String& config) {
+    return makePtr<DetectionModel>(model, config);
+}
```
dkurt: Do we really need `create` methods?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,171 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+
+         /** @brief Set flag crop for frame.
+          *  @param[in] crop Flag which indicates whether image will be cropped after resize or not.
+         */
+         Model& setInputCrop(bool crop);
+
+         /** @brief Set flag swapRB for frame.
+          *  @param[in] swapRB Flag which indicates that swap first and last channels.
+         */
+         Model& setInputSwapRB(bool swapRB);
```
dkurt: Can we add `CV_WRAP` for these methods as well? I mean will it work in case of returned reference?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,171 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+          *  @param[in]  frame  The input image.
+          *  @param[out] classIds Class indexes in result detection.
+          *  @param[out] confidences A set of corresponding confidences.
+          *  @param[out] boxes A set of bounding boxes.
+          *  @param[in] confThreshold A threshold used to filter boxes by confidences.
+          *  @param[in] nmsThreshold A threshold used in non maximum suppression.
+          */
+         CV_WRAP void detect(InputArray frame, CV_OUT std::vector<int>& classIds,
+                     CV_OUT std::vector<float>& confidences, CV_OUT std::vector<Rect>& boxes,
+                     float confThreshold = 0.5, float nmsThreshold = 0.4);
```
dkurt: Please fix indentation here and in other similar cases.

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -21,16 +21,15 @@ def box2str(box):
...
+                 confThreshold=0.0, scores_diff=1e-5, boxes_iou_diff=1e-4):
     matchedRefBoxes = [False] * len(refBoxes)
     errMsg = ''
-    for i in range(len(refBoxes)):
-        testScore = testScores[i]
+    for i in range(len(testBoxes)):
+        try:
+            testScore = testScores[i]
+        except IndexError as e:
+            errMsg += '\nUnmatched reference: class %d score %f box %s' % (refClassIds[i], refScores[i], box2str(refBoxes[i]))
```
dkurt: That's looks strange. What actual error we want to catch here?

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -21,16 +21,15 @@ def box2str(box):
...
     return '[%f x %f from (%f, %f)]' % (width, height, left, top)
 
-def normAssertDetections(test, ref, out, confThreshold=0.0, scores_diff=1e-5, boxes_iou_diff=1e-4):
-    ref = np.array(ref, np.float32)
-    refClassIds, testClassIds = ref[:, 1], out[:, 1]
-    refScores, testScores = ref[:, 2], out[:, 2]
-    refBoxes, testBoxes = ref[:, 3:], out[:, 3:]
-
+def assertDetect(test, refClassIds, refScores, refBoxes, testClassIds, testScores, testBoxes,
+                 confThreshold=0.0, scores_diff=1e-5, boxes_iou_diff=1e-4):
```
dkurt: In my point of view, it's better now to keep the name as `normAssertDetections` with new set of parameters because this method is called once and we can add slicing there.

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -135,6 +143,41 @@ def test_blobFromImage(self):
...
+        mean = (127.5, 127.5, 127.5)
+        scale = 1.0 / 127.5
+        model.setInputParams(size=size, mean=mean, scale=scale)
+
+        iouDiff = 1
+        confThreshold = 0.000001
+        nmsThreshold = 0
+        scoreDiff = 1e-5
+
+        classIds, confidences, boxes = model.detect(frame, confThreshold, nmsThreshold)
```
dkurt: Please also test arguments names:

```python
classIds, confidences, boxes = model.detect(frame, confThreshold=0, nmsThreshold=0)
```

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,171 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+          *  @param[in]  frame  The input image.
+          *  @param[out] outs Allocated output blobs, which will store results of the computation.
+          */
+         CV_WRAP void predict(InputArray frame, CV_OUT OutputArrayOfArrays outs);
+
+         /** @brief The static method creates model from config and weights.
+         * @param[in] model Binary file contains trained weights.
+         * @param[in] config Text file contains network configuration.
+        */
+         CV_WRAP static Ptr<Model> create(const String& model, const String& config = "");
```
dkurt: I think that with wrapped constructors we can remove "create" methods.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,171 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+         CV_WRAP static Ptr<ClassificationModel> create(const String& model, const String& config = "");
+     };
+
+     /** @brief This class is presented high-level API for detection models.
+      *
+      * DetectionModel allows to set params for preprocessing input image.
+      * DetectionModel creates net from file with trained weights and config,
+      * sets preprocessing input, runs forward pass and return result detections.
+      */
+     class CV_EXPORTS_W DetectionModel : public Model
```
dkurt: I think for certain models such DetectionModel or ClassificationModel we need just brief summary.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,171 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+         CV_WRAP void classify(InputArray frame, CV_OUT int& classId, CV_OUT float& conf);
+
+         /** @brief The static method creates classification model from config and weights.
+         * @param[in] model Binary file contains trained weights.
+         * @param[in] config Text file contains network configuration.
+        */
+         CV_WRAP static Ptr<ClassificationModel> create(const String& model, const String& config = "");
+     };
+
+     /** @brief This class is presented high-level API for detection models.
```
dkurt: "is presented" -> "represents"
"detection" -> "object detection"

modules/dnn/include/opencv2/dnn/dnn.hpp
l-bat: I get error: `no matching function for call to ‘cv::dnn::experimental_dnn_34_v12::Model::Model()’
     Model retval;
`

dkurt: Ok, thanks. Let's fix it next time.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,256 @@
...
+                Mat scores = detections[i].row(j).colRange(5, detections[i].cols);
+                Point classIdPoint;
+                double conf;
+                minMaxLoc(scores, nullptr, &conf, nullptr, &classIdPoint);
+
+                if (conf < confThreshold)
+                    continue;
+
+                int centerX = std::max(0, std::min(static_cast<int>(data[0] * frameWidth), frameWidth - 1));
+                int centerY = std::max(0, std::min(static_cast<int>(data[1] * frameHeight), frameWidth - 1));
```
dkurt: I think it's enough to control just output `left` and `top` values.

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,256 @@
...
+                int width   = data[2] * frameWidth;
+                int height  = data[3] * frameHeight;
+
+                int left = std::max(0, std::min(centerX - width / 2, frameWidth - 1));
+                int top  = std::max(0, std::min(centerY - height / 2, frameHeight - 1));
+                width    = std::max(0, std::min(width, frameWidth - 1 - left));
```
dkurt: If `left == 0` and `width ==1` and `frameWidth == 1` we should have valid `left == 0` and `width == 1`. For now it assigns `width = min(1, 1 - 1 - 0) = 0`.

modules/dnn/test/test_model.cpp
```diff
@@ -0,0 +1,163 @@
...
+    std::string img_path = _tf("street.png");
+    std::string weights_file = _tf("MobileNetSSD_deploy.caffemodel");
+    std::string config_file = _tf("MobileNetSSD_deploy.prototxt");
+
+    Scalar mean = Scalar(127.5, 127.5, 127.5);
+    float scale = 1.0f / 127.5;
+    Size size{300, 300};
+
+    double scoreDiff = (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD) ? 1.5e-2 : 1e-5;
+    double iouDiff = 0.05;
```
dkurt: That's too big difference for FP32 computations.

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -136,6 +132,37 @@ def test_blobFromImage(self):
...
+        refConfidences = (0.9998, 0.8793)
+        refBoxes = ((329.351, 238.952, 85.334, 102.106),
+                    (101.638, 189.152, 34.217, 138.234))
+
+        normAssertDetections(self, refClassIds, refConfidences, refBoxes,
+                             classIds, confidences, boxes,confThreshold, scoreDiff, iouDiff)
+
+        for box, conf, classId in zip(boxes, confidences, classIds):
+            left, top, width, height = box
+            cv.rectangle(frame, (left, top), (left + width, top + height), (0, 255, 0))
```
dkurt: Can we use just `cv.rectangle(frame, box, (0, 255, 0))`?
source: https://docs.opencv.org/master/d6/d6e/group__imgproc__draw.html#gac865734d137287c0afb7682ff7b3db23

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -166,7 +193,13 @@ def test_face_detection(self):
             scoresDiff = 4e-3 if target in [cv.dnn.DNN_TARGET_OPENCL_FP16, cv.dnn.DNN_TARGET_MYRIAD] else 1e-5
             iouDiff = 2e-2 if target in [cv.dnn.DNN_TARGET_OPENCL_FP16, cv.dnn.DNN_TARGET_MYRIAD] else 1e-4
 
-            normAssertDetections(self, ref, out, 0.5, scoresDiff, iouDiff)
+            ref = np.array(ref, np.float32)
```
dkurt: What's an origin type of `ref`?

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -21,15 +21,11 @@ def box2str(box):
...
-    refClassIds, testClassIds = ref[:, 1], out[:, 1]
-    refScores, testScores = ref[:, 2], out[:, 2]
-    refBoxes, testBoxes = ref[:, 3:], out[:, 3:]
-
+def normAssertDetections(test, refClassIds, refScores, refBoxes, testClassIds, testScores, testBoxes,
+                 confThreshold=0.0, scores_diff=1e-5, boxes_iou_diff=1e-4):
     matchedRefBoxes = [False] * len(refBoxes)
     errMsg = ''
-    for i in range(len(refBoxes)):
+    for i in range(len(testBoxes)):
```
dkurt: Why we not iterate over reference boxes?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,157 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+         Ptr<Impl> impl;
+     };
+
+     /** @brief This class represents high-level API for classification models.
+      *
+      * ClassificationModel allows to set params for preprocessing input image.
+      * ClassificationModel creates net from file with trained weights and config,
+      * sets preprocessing input, runs forward pass and return top-1 prediction.
+      * For ClassificationModel AlexNet, SqueezeNet, VGG, Inception, ResNet
+      * classification networks are supported.
```
dkurt: We mentioned SSD, Faster R-CNN and YOLO networks because DetectionModel has different logic for them. For classification networks we have unified approach with argmax so we can emit this list because all the models are the same in terms of classification task.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -996,6 +996,157 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+         /** @overload */
+         CV_WRAP void classify(InputArray frame, CV_OUT int& classId, CV_OUT float& conf);
+     };
+
+     /** @brief This class represents high-level API for object detection networks.
+      *
+      * DetectionModel allows to set params for preprocessing input image.
+      * DetectionModel creates net from file with trained weights and config,
+      * sets preprocessing input, runs forward pass and return result detections.
+      * For DetectionModel SSD, Faster R-CNN, YOLO object detection networks are supported.
```
dkurt: Let's name them topologies or architectures.

modules/dnn/misc/python/test/test_dnn.py
l-bat: list

modules/dnn/misc/python/test/test_dnn.py
l-bat: if len(testBoxes) < len(refBoxes) we get `IndexError`

modules/dnn/misc/python/test/test_dnn.py
l-bat: It works only on the `master` branch.

modules/dnn/misc/python/test/test_dnn.py
dkurt: Ok, I see it.

modules/dnn/misc/python/test/test_dnn.py
dkurt: So we are going to merge this PR to master branch as well.

modules/dnn/test/test_model.cpp
```diff
@@ -0,0 +1,163 @@
...
+    std::string img_path = _tf("street.png");
+    std::string weights_file = _tf("MobileNetSSD_deploy.caffemodel");
+    std::string config_file = _tf("MobileNetSSD_deploy.prototxt");
+
+    Scalar mean = Scalar(127.5, 127.5, 127.5);
+    float scale = 1.0f / 127.5;
+    Size size{300, 300};
+
+    double scoreDiff = (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD) ? 1.5e-2 : 1e-5;
+    double iouDiff = 0.0098;
```
dkurt: Please check if we can use different IoU thresholds for FP32 and FP16 targets. In the origin test from https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_caffe_importer.cpp there were:
```cpp
    const float scores_diff = (targetId == DNN_TARGET_OPENCL_FP16 || targetId == DNN_TARGET_MYRIAD) ? 1.5e-2 : 1e-5;
    const float boxes_iou_diff = (targetId == DNN_TARGET_OPENCL_FP16 || targetId == DNN_TARGET_MYRIAD) ? 6.3e-2 : 1e-4;
```

modules/dnn/misc/python/test/test_dnn.py
dkurt: Oh, sorry. I agree with this change. In the origin C++ implementation it aslo iterates over test boxes: https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_common.impl.hpp

modules/dnn/src/model.cpp
```diff
@@ -0,0 +1,256 @@
...
+                int bottom = data[j + 6];
+                int width  = right  - left + 1;
+                int height = bottom - top + 1;
+
+                if (width * height <= 1)
+                {
+                    left   = data[j + 3] * frameWidth;
+                    top    = data[j + 4] * frameHeight;
+                    right  = data[j + 5] * frameWidth;
+                    bottom = data[j + 6] * frameHeight;
```
dkurt: @l-bat, in https://github.com/opencv/opencv/issues/14841 we figured out that it's more robust to multiply predicted values from DetectionOutput by `frameWidth - 1` and `frameHeight - 1` because it produces values in range [0, 1].

modules/dnn/test/test_model.cpp
```diff
@@ -0,0 +1,182 @@
...
+        refClassIds.emplace_back(ref.at<float>(i, 1));
+        refConfidences.emplace_back(ref.at<float>(i, 2));
+        int left   = ref.at<float>(i, 3) * (frameWidth - 1);
+        int top    = ref.at<float>(i, 4) * (frameHeight - 1);
+        int right  = ref.at<float>(i, 5) * (frameWidth - 1);
+        int bottom = ref.at<float>(i, 6) * (frameHeight - 1);
+        int width  = right  - left + 1;
+        int height = bottom - top + 1;
+        refBoxes.emplace_back(left, top, width, height);
+    }
```
dkurt: Please fix the rest of tests correspondingly. Perhaps, you need to write a separate method to extract boxes from reference blob.

modules/python/src2/cv2.cpp
```diff
@@ -1366,6 +1358,25 @@ template<> struct pyopencvVecConverter<RotatedRect>
...
+    pyopencvVecConverter<int>::to(obj, value, ArgInfo(name, 0));
+    r = Rect(value[0], value[1], value[2], value[3]);
+    return true;
+}
+
+template<>
+PyObject* pyopencv_from(const Rect& r)
+{
+    return Py_BuildValue("(iiii)", r.x, r.y, r.width, r.height);
+}
```
dkurt: Please move `pyopencv_from` to it's origin position.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -992,6 +992,155 @@ CV__DNN_INLINE_NS_BEGIN
...
+          *  @param[in]  frame  The input image.
+          *  @param[out] classIds Class indexes in result detection.
+          *  @param[out] confidences A set of corresponding confidences.
+          *  @param[out] boxes A set of bounding boxes.
+          *  @param[in] confThreshold A threshold used to filter boxes by confidences.
+          *  @param[in] nmsThreshold A threshold used in non maximum suppression.
+          */
+         CV_WRAP void detect(InputArray frame, CV_OUT std::vector<int>& classIds,
+                             CV_OUT std::vector<float>& confidences, CV_OUT std::vector<Rect>& boxes,
+                             float confThreshold = 0.5, float nmsThreshold = 0.4);
```
dkurt: Please use `0.5f` and `0.4f` or switch to doubles.

modules/dnn/test/test_model.cpp
```diff
@@ -0,0 +1,182 @@
...
+                         const Size& size = {-1, -1}, Scalar mean = Scalar(),
+                         double scale = 1.0, bool swapRB = false, bool crop = false)
+    {
+        checkBackend();
+
+        Mat frame = imread(imgPath);
+        DetectionModel model(weights, cfg);
+
+        model.setInputSize(size).setInputMean(mean).setInputScale(scale)
+             .setInputSwapRB(swapRB).setInputCrop(crop);
```
dkurt: I think that we also need to set backend and target?

modules/python/src2/cv2.cpp
```diff
@@ -1111,17 +1102,17 @@ template<typename _Tp> struct pyopencvVecConverter
...
                     return false;
                 data[j] = saturate_cast<_Cp>(v);
             }
-            else if( PyFloat_Check(item_ij))
+            else if( PyLong_Check(item_ij) || PyNumber_Check(item_ij))
             {
-                double v = PyFloat_AsDouble(item_ij);
-                if( PyErr_Occurred() )
+                int v = (int)PyLong_AsLong(item_ij);
+                if( v == -1 && PyErr_Occurred() )
```
dkurt: Can you please check the final version of changes? Is that necessary order of conditions?

dkurt:
@l-bat, The last action:

```
[ RUN      ] Test_Model.DetectionMobilenetSSD/1, where GetParam() = DLIE/MYRIAD
Unmatched prediction: class 7 score 1.000000 box [81 x 100 from (330, 239)]
/build/precommit_custom_linux/opencv/modules/dnn/test/test_common.impl.hpp:125: Failure
Value of: matched
  Actual: false
Expected: true
Unmatched reference: class 7 score 0.999835 box [85 x 102 from (328, 238)]
/build/precommit_custom_linux/opencv/modules/dnn/test/test_common.impl.hpp:135: Failure
Expected: (refScores[i]) <= (confThreshold), actual: 0.999835 vs 1.17549e-38
[  FAILED  ] Test_Model.DetectionMobilenetSSD/1, where GetParam() = DLIE/MYRIAD (468 ms)
```

bhack:
When do you plan to merge this? There is new a PR that needs this changes.

dkurt:
@bhack, We decided to merge it after coming release. This PR contains major changes and it's better to debug it between two OpenCV versions. 

You can rebase your branch with these changes and refer this PR.

bhack:
If you can push this branch in Opencv the  student could could open a PR on this branch.

dvd42:
Yes, as @bhack said, I would like to open a PR that depends this one. I have added a generation and segmentation module to the API and It would be great if I could get the PRs reviewed as soon as possible so I can get my changes reviewed

dkurt:
@dvd42, @bhack, there is no technical problem to do that:

```
git remote add l-bat https://github.com/l-bat/opencv
git fetch l-bat high_level_API
git rebase high_level_API your_branch
git push origin your_branch
```

bhack:
Is not that the case. The scope was to open a github PR with the minimal diff here. If not Diego will need to open on l-bat github  fork but probably nobody will review there.

mshabunin:
@bhack , @dkurt meant that Diego can make changes on top of commits from this PR and include it in theirs PR to the main repository (opencv/opencv). After this PR is merged, these commits will be hidden from Diego's PR.

```
Now:
  #14780 - commitA, commitB
  #<new PR> - commitA, commitB, commitX

After #14780 is merged:
  #14780 - closed
  #<new PR> - commitX
```

We can help with rebasing if you have any troubles.

bhack:
Yes I know but I meant can you review also if the PR have comittA and commitB in the diff and it is not so clean?
If you can push the branch in this repo it could be easier to review cause the diff will be minimized.

dkurt:
@bhack, Yes, it's not a problem.

dvd42:
I created 2 PRs, from my fork against opencv/master adding a [generation](https://github.com/opencv/opencv/pull/15081) and [segmentation](https://github.com/opencv/opencv/pull/15082) module

dvd42:
I created 1 PR for this module, from my fork against opencv/master adding a [keypoints](https://github.com/opencv/opencv/pull/15189) module. This module raises the problem that [G-API](https://github.com/opencv/opencv/pull/15137) is trying to fix by integrating detection and landmarking into one pipeline. Maybe we could do something similar for this high-level API, by concatenating modules?

dkurt:
@dvd42, Maybe yes but not now. We need to solve current issue first.

PR #14791, opened 2019-06-12
Small change to the Onnx Importer error handling

### LINKED ISSUE #14772: ONNX Graph Traversal and Unsupported Operation Handling
Hello I am a GSoC student working on an [OpenCV project](https://github.com/dvd42/opencv/wiki). This project requires importing multiple ONNX models using the OpenCV importer (readNetFromONNX). 
It might be the case that some of the models that have to be imported have some operations that are currently not supported in OpenCV _(e.g: acos, asin, etc)_. I am aware that OpenCV supports the creation of custom layers and one could implement any functionality in order to eventually export the model. However before starting to implement a custom layer for model **X** it could be very useful to know how many unsupported layers model **X** has, in order to perform a risk evaluation on the task of implementing this custom operation.  I am aware that you can look at the [importer code](https://github.com/opencv/opencv/blob/master/modules/dnn/src/onnx/onnx_importer.cpp#L385) to check every supported operation, however this is not very practical. 
Currently the importer will stop at the first unsupported operation and raise and exception. For example when running the following graph:

```
graph(%input.1 : Float(5, 10)
      %1 : Float(10, 10)
      %2 : Float(10)) {
      %3 : Float(5, 10) = onnx::Gemm[alpha=1, beta=1, transB=1](%input.1, %1, %2), scope: mymodel/Linear[linear]
  %4 : Float(5, 10) = onnx::Acos(%3), scope: mymodel
  %5 : Float(5, 10) = onnx::Gemm[alpha=1, beta=1, transB=1](%4, %1, %2), scope: mymodel/Linear[linear]
  %6 : Float(5, 10) = onnx::Gemm[alpha=1, beta=1, transB=1](%5, %1, %2), scope: mymodel/Linear[linear]
  %7 : Float(5, 10) = onnx::Asin(%6), scope: mymodel
  return (%7);
}
```

It will raise: 

` error: (-2:Unspecified error) Can't create layer "4" of type "Acos" in function 'getLayerInstance'`

But it wont tell you anything about layer 7 (Asin). 

So I would like to open a PR that fixes this minor issue by accumulating information about all the unsupported operations during the ONNX graph traversal and then throw an error at the end detailing each unsupported operation was found and what the operation is.

dvd42:
resolves #14772 

### This pullrequest changes

The way the **ONNX** unsupported operations are reported to the user. Instead of stopping at the first unsupported operation and rasing an exception, it will continue to run the graph until the end and report ALL of the unsupported operations at the end.


modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -840,27 +848,68 @@ void ONNXImporter::populateNet(Net dstNet)
...
+            {
+                // Compute shape of output blob for this layer.
+                Ptr<Layer> layer = dstNet.getLayer(id);
+                layer->getMemoryShapes(layerInpShapes, 0, layerOutShapes, layerInternalShapes);
+                CV_Assert(!layerOutShapes.empty());
+                outShapes[layerParams.name] = layerOutShapes[0];
+            }
+            catch(cv::Exception& e)
+            {
+                op_error.push_back(e.what());
```
alalek: I'm completely against that.
You loosing any useful information from C++ exception in that construction.
I didn't see any SDK/library whom uses exceptions in that way.

bhack: @alalek What design do you want without totally broke the importer?
 We want to give to the user full visibility on all the missing stuffs when he try to import an onnx model and not to let the user to die fixing a single op after  op after unsupport params etc.. We want to give him full visibility of the work that it is required to run that model.

dkurt: There are different levels where assertion appears. Most of them in layers implementation. As I initially proposed in https://github.com/opencv/opencv/issues/14772#issuecomment-500448114 we need to collect just names of unsupported layers. It might be done in in `setUp` method in dnn.cpp:

```
clear();

// Iterate over layers here and summarize the types of unsupported layers

allocateLayers(blobsToKeep_);
```

We couldn't keep all the error messages in a single message. But at least can report about unknown layers types once.

dvd42: I thought you proposed changing the [getLayerInstance](https://github.com/opencv/opencv/blob/6389303c1c3b7269aa805cc35f691dd20fc6dde0/modules/dnn/src/dnn.cpp#L525) function, which is the one that throws the error.

dkurt: @dvd42, Something like
```cpp
std::string errMsg = "";
for (MapIdToLayerData::iterator it = layers.begin(); it != layers.end(); ++it)
{
    LayerData &ld = it->second;
    try
    {
       ld.getLayerInstance()
    }
    catch(...)
    {
        errMsg += ld.type + " (" + ld.name + ");
    }
}
if (!errMsg.empty())
    CV_Error(Error::StsNotImplemented, "Unsupported layers types: " + errMsg);
```

`allocateLayers` is the first method which calls `getLayerInstance` for layers. So we can call it before to collect unsupported types and names.

dvd42: Thanks for the clarification. That looks very neat. Sorry for not understanding what you proposed at first. The dnn code can be a bit confusing for a newcomer. I will correct the implementation as per your suggestion.

dvd42: @dkurt I tried the solution you proposed and it didn't work. It seems like the readNetFromONNX runs fully without calling the setUp net function. And so the exception, as before gets thrown [here](https://github.com/opencv/opencv/blob/81fad9e2472af758130a15e0cd9f105bea9cc1f4/modules/dnn/src/onnx/onnx_importer.cpp#L860)

alalek: If exception is thrown in other place (not during instantiation of layer), then it can't be fixed via "custom layers" approach.

For example, there is no way to fix INT64 tensors due missing INT64 support in OpenCV.

PR #14800, opened 2019-06-14
Test request by #14583

FanaticsKang:
This is a test which is request by #14583.
The cvUndistortPointsInternal function.
The k[4] is negative of my camera, the icdist will be negative at the edge of image. This cause the error like the point at right will be undistort to the left.
So the function will return when icdist < 0, and use the last compute result.

I use my camera interior parameter matrix and distortion parameter. But the `icdist` will  `<0`, and cause error.

This pull request will be closed when #14583 is over.


FanaticsKang:
You will see the error by running ./bin/opencv_test_imgproc.

PR #14809, opened 2019-06-14
Increase CV_IO_MAX_IMAGE_PIXELS

### LINKED ISSUE #10211: Very large image not supported (area > 2^32)

jxu:
resolves #10211 again

### This pullrequest changes

Increase CV_IO_MAX_IMAGE_PIXELS to 2^40 since apparently #11505 fixes #10211. I have no idea what this change breaks, but I assume all tests passed because no one is testing huge images. So if this max is increased then big images need to be tested. 
While we're at it maybe CV_IO_MAX_IMAGE_WIDTH and CV_IO_MAX_IMAGE_HEIGHT should be increased too. 

I tried reading in a 50000x50000 PNG file and running simple operations: saving, GaussianBlur, boxFilter, Laplacian. I could not load a 100000x100000 PNG without running out of memory.

Perhaps a warning can be issued to users if they try loading images bigger than 2^30 (previous limit). I know Pillow errors but this error can be turned into a warning or suppressed without recompiling. 

<details>

50000_box
50000_blur

</details>

### DIFF
--- a/modules/imgcodecs/src/loadsave.cpp
+++ b/modules/imgcodecs/src/loadsave.cpp
@@ -63,7 +63,7 @@ namespace cv {
 static const size_t CV_IO_MAX_IMAGE_PARAMS = cv::utils::getConfigurationParameterSizeT("OPENCV_IO_MAX_IMAGE_PARAMS", 50);
 static const size_t CV_IO_MAX_IMAGE_WIDTH = utils::getConfigurationParameterSizeT("OPENCV_IO_MAX_IMAGE_WIDTH", 1 << 20);
 static const size_t CV_IO_MAX_IMAGE_HEIGHT = utils::getConfigurationParameterSizeT("OPENCV_IO_MAX_IMAGE_HEIGHT", 1 << 20);
-static const size_t CV_IO_MAX_IMAGE_PIXELS = utils::getConfigurationParameterSizeT("OPENCV_IO_MAX_IMAGE_PIXELS", 1 << 30);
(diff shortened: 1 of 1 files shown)


alalek:
This is done to avoid vulnerability (memory DOS).
It is not going to be increased by default. Current values are enough for the most part of applications.

If you want to play with huge images, then you should take a look on corresponding configuration parameters.

jxu:
Which config parameters? Does it require recompiling?

alalek:
It is runtime parameters, recompilation is not required (specify via environment variables).
You touching their default values, what it a bad idea.

LaurentBerger:
read https://answers.opencv.org/question/210482/read-and-display-image-of-large-size/

PR #14818, opened 2019-06-15
set appsink drop buffers

### LINKED ISSUE #5715: memory leak with videocapture, memory goes up with the function of read(image)
Hello! I am using opencv 3.0 beta with visual studio 2013 under windows 7 and 64bit. I downloaded opencv3.0 beta for windows from http://opencv.org/downloads.html. 
The following is the code I am using to capture the video from rtsp stream and then save some frames every several minutes. It has no problem to capture and read the stream. But I found, the function "read(InputArrray Image)" cause the memory leak. The memory goes up slowly, about one MB for two or three hours. But it keep going up. I am using it for severy months continually. so it is seriouly problem for me. The following is my code for test. Please help me find out where is my problem with these code. Thanks a lot!
int main()
{
    string fileNameSave;
    string URLName;

```
BOOL stop = FALSE;
BOOL saveFrame = TRUE;
BOOL got_frame = FALSE;
string fileSave = "saveFrame.jpg";
URLName = "rtsp://username:password@IP adress:554/h264/ch1/main/av_stream";
VideoCapture capture;
Mat frame;
capture.open(URLName);
if (!capture.isOpened()){
    cout << " channel 1st has error " << endl;
}
while (!stop){  
    got_frame = capture.read(frame); //here the memory keep goes up
    //save the frame every minute locally
    if (saveFrame == TRUE){
        if (got_frame){
            imwrite(fileSave, frame);
        }
    }
}
capture.release();
return 0;
```

}

tripzero:
If we don't drop the buffers, the queue may fill too much when
streaming.

resolves #5715

### DIFF
--- a/modules/videoio/src/cap_gstreamer.cpp
+++ b/modules/videoio/src/cap_gstreamer.cpp
@@ -763,7 +763,7 @@ bool GStreamerCapture::open(const String &filename_)
 
     //TODO: is 1 single buffer really high enough?
     gst_app_sink_set_max_buffers (GST_APP_SINK(sink), 1);
-//    gst_app_sink_set_drop (GST_APP_SINK(sink), stream);
+    gst_app_sink_set_drop (GST_APP_SINK(sink), TRUE);
     //do not emit signals: all calls will be synchronous and blocking
     gst_app_sink_set_emit_signals (GST_APP_SINK(sink), FALSE);
 //    gst_base_sink_set_sync(GST_BASE_SINK(sink), FALSE);


alalek:
Reading from regular video files should not drop frames.
It is critical for computer vision tests which may run under debug or heavy loaded machines / embedded platforms.

This mode may be acceptable for live streaming / IP cameras mode only.

So, the patch should have proper design, including:
- proper detection of that modes (if possible)
- property to control frames drop
- in the case of GStreamer, probably users can control frames dropping mode via manual pipeline configuration

tripzero:
custom/manual pipeline is an acceptable workaround.

PR #14820, opened 2019-06-17
G-API: Implement line and circle draw operations

TolyaTalamanov:
```
```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/render.hpp
+++ b/modules/gapi/include/opencv2/gapi/render.hpp
@@ -31,15 +31,14 @@ namespace draw
 struct Text
 {
     /*@{*/
-    std::string text;               /** < The text string to be drawn */
-    cv::Point   org;                /** < The bottom-left corner of the text string in the image */
-    int         ff;                 /** < The font type, see #HersheyFonts */
-    double      fs;                 /** < The font scale factor that is multiplied by the font-specific base size */
-    cv::Scalar  color;              /** < The text color */
-    int         thick;              /** < The thickness of the lines used to draw a text */
-    int         lt;                 /** < The line type. See #LineTypes */
-    bool        bottom_left_origin; /** < When true, the image data origin is at the bottom-left corner. Otherwise,
-                                          it is at the top-left corner. */
+    std::string text;               //!< The text string to be drawn
+    cv::Point   org;                //!< The bottom-left corner of the text string in the image
+    int         ff;                 //!< The font type, see #HersheyFonts
+    double      fs;                 //!< The font scale factor that is multiplied by the font-specific base size
+    cv::Scalar  color;              //!< The text color
+    int         thick;              //!< The thickness of the lines used to draw a text
+    int         lt;                 //!< The line type. See #LineTypes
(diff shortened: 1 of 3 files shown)


TolyaTalamanov:
@dmatveev

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -50,12 +50,41 @@ struct Rect
 {
     cv::Rect   rect;  /** Coordinates of the rectangle < */
     cv::Scalar color; /** The rectangle color or brightness (grayscale image) < */
```
dmatveev: I believe this `< ` thing is not required actually? Please take a look how other public structure fields are documented in OpenCV. When I checked the generated documentation for OSD the last time, I saw that `< ` there in the generated comments.

modules/gapi/include/opencv2/gapi/render.hpp
```diff
@@ -50,12 +50,41 @@ struct Rect
...
+    cv::Point  pt1;    /** The first point of the line segment  < */
+    cv::Point  pt2;    /** The second point of the line segment < */
+    cv::Scalar color;  /** The line color                       < */
+    int        thick;  /** The thickness of line                < */
+    int        lt;     /** The Type of the line. See #LineTypes < */
+    int        shift;  /** The number of fractional bits in the point coordinates */
+
+};
+
+using Prim  = util::variant<Text, Rect, Circle, Line>;
```
dmatveev: Let's make it
```
using Prim = util::variant
    < Text
    , Rect
    , Circle
    , Line
    >;
```
Here you are editing the same line again and again which makes harder to track changes in blame, in my case adding a primitive is a new line, and editing a primitive results in a single-line change (per primitive)

modules/gapi/src/api/render.cpp
```diff
@@ -28,7 +28,21 @@ void cv::gapi::wip::draw::render(cv::Mat& bgr, const Prims& prims)
                 break;
             }
 
-            default: util::throw_error(std::logic_error("Unsupported draw event"));
+            case Prim::index_of<Circle>():
+            {
+                auto c_p = cv::util::get<Circle>(p);
```
dmatveev: `const auto &`? (avoid possible extra copy?)

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/render.hpp
TolyaTalamanov: Fixed

modules/gapi/src/api/render.cpp
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/render.hpp
dmatveev: image

Much nicer now, thanks!

PR #14860, opened 2019-06-21
tensorflow support maxpoolgrad(MaxUnPool2d)

vonchenplus:
```

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
```

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/623

### DIFF
--- a/modules/dnn/src/layers/max_unpooling_layer.cpp
+++ b/modules/dnn/src/layers/max_unpooling_layer.cpp
@@ -43,12 +43,18 @@ public:
                          std::vector<MatShape> &outputs,
                          std::vector<MatShape> &internals) const CV_OVERRIDE
     {
-        CV_Assert(inputs.size() == 2);
+        CV_Assert(inputs.size() == 2 || inputs.size() == 3);
         CV_Assert(total(inputs[0]) == total(inputs[1]));
 
-        MatShape outShape = inputs[0];
-        outShape[2] = (outShape[2] - 1) * poolStride.height + poolKernel.height - 2 * poolPad.height;
-        outShape[3] = (outShape[3] - 1) * poolStride.width + poolKernel.width - 2 * poolPad.width;
+        MatShape outShape;
+        if (inputs.size() == 2)
+        {
+            outShape = inputs[0];
+            outShape[2] = (outShape[2] - 1) * poolStride.height + poolKernel.height - 2 * poolPad.height;
+            outShape[3] = (outShape[3] - 1) * poolStride.width + poolKernel.width - 2 * poolPad.width;
+        }
+        else
+            outShape = inputs[2];
 
         outputs.clear();
         outputs.push_back(outShape);
@@ -71,7 +77,7 @@ public:
         inputs_arr.getMatVector(inputs);
         outputs_arr.getMatVector(outputs);
 
-        CV_Assert(inputs.size() == 2);
+        CV_Assert(inputs.size() == 2 || inputs.size() == 3);
         Mat& input = inputs[0];
         Mat& indices = inputs[1];
 
--- a/modules/dnn/src/tensorflow/tf_importer.cpp
+++ b/modules/dnn/src/tensorflow/tf_importer.cpp
@@ -1370,6 +1370,24 @@ void TFImporter::populateNet(Net dstNet)
 
             connectToAllBlobs(layer_id, dstNet, parsePin(layer.input(0)), id, layer.input_size());
         }
+        else if (type == "MaxPoolGrad")
+        {
+            CV_Assert(layer.input_size() == 3);
+
+            layerParams.set("pool_k_h", 0);
+            layerParams.set("pool_k_w", 0);
+            layerParams.set("pool_stride_h", 0);
+            layerParams.set("pool_stride_w", 0);
+            layerParams.set("pool_pad_h", 0);
+            layerParams.set("pool_pad_w", 0);
+
+            int id = dstNet.addLayer(name, "MaxUnpool", layerParams);
+            layer_id[name] = id;
+
+            connect(layer_id, dstNet, parsePin(layer.input(2)), id, 0);
+            connect(layer_id, dstNet, parsePin(layer.input(1) + ":1"), id, 1);
+            connect(layer_id, dstNet, parsePin(layer.input(0)), id, 2);
+        }
         else if (type == "Placeholder")
         {
             if (!hasLayerAttr(layer, "dtype") ||
--- a/modules/dnn/test/test_tf_importer.cpp
+++ b/modules/dnn/test/test_tf_importer.cpp
@@ -218,6 +218,13 @@ TEST_P(Test_TensorFlow_layers, pooling)
     runTensorFlowNet("reduce_mean");  // an average pooling over all spatial dimensions.
 }
 
+TEST_P(Test_TensorFlow_layers, max_pool_grad)
+{
+    if (backend == DNN_BACKEND_INFERENCE_ENGINE)
+        applyTestTag(CV_TEST_TAG_DNN_SKIP_IE);
+    runTensorFlowNet("max_pool_grad");
+}
+
 // TODO: fix tests and replace to pooling
 TEST_P(Test_TensorFlow_layers, ave_pool_same)
 {


modules/dnn/src/layers/max_unpooling_layer.cpp
```diff
@@ -71,9 +78,63 @@ class MaxUnpoolLayerImpl CV_FINAL : public MaxUnpoolLayer
...
+            input = inputs[0];
+            indices = inputs[1];
+        }
+        else if (inputs.size() == 3)
+        {
+            input = inputs[2];
+
+            int kernel_area = poolKernel.height * poolKernel.width;
+            Mat maxpoolintp = inputs[0];
+            Mat maxpooloutp = inputs[1];
```
dkurt: May I ask you to just print these two matrices (with lower resolution input tensor)?
```cpp
std::cout << maxpoolintp.size << std::endl;
std::cout << maxpoolintp.reshape(1, 1) << std::endl;
std::cout << maxpooloutp.size << std::endl;
std::cout << maxpooloutpreshape(1, 1) << std::endl;
```

I think that one of them is constant and we can perform this layer using existing implementation.

modules/dnn/src/torch/torch_importer.cpp
```diff
@@ -1201,12 +1201,12 @@ struct TorchImporter
...
-                module->params.set("pool_stride_h", poolingLayer.second->params.get<int>("stride_h"));
-                module->params.set("pool_stride_w", poolingLayer.second->params.get<int>("stride_w"));
-                module->params.set("pool_pad_h", poolingLayer.second->params.get<int>("pad_h"));
-                module->params.set("pool_pad_w", poolingLayer.second->params.get<int>("pad_w"));
+                module->params.set("kernel_h", poolingLayer.second->params.get<int>("kernel_h"));
+                module->params.set("kernel_w", poolingLayer.second->params.get<int>("kernel_w"));
+                module->params.set("stride_h", poolingLayer.second->params.get<int>("stride_h"));
+                module->params.set("stride_w", poolingLayer.second->params.get<int>("stride_w"));
+                module->params.set("pad_h", poolingLayer.second->params.get<int>("pad_h"));
+                module->params.set("pad_w", poolingLayer.second->params.get<int>("pad_w"));
```
dkurt: Let's keep them as is. I think that there is no critical difference for now.

vonchenplus: tensorflow's setKSize use ‘’kernel_h‘’ and ‘’kernel_w‘’, setStrides use ‘’stride_h‘ and ’stride_w‘. This is somewhat different from the variable name of tensorflow.

modules/dnn/src/layers/max_unpooling_layer.cpp
vonchenplus: tensorflow's MaxUnpool need three inputs. The first parameter is the input of maxpool, the second paramter is the output of maxpool, the third paramter is the grad. 
for example:
_X_content = tf.placeholder(tf.float32, shape=(1, 256, 256, 3), name='X_content')
net = MaxPooling2D(pool_size=(2, 2), strides=(2, 2))(X_content)
net = MaxUnPooling2D(net.op, net)_
The first parameter is **net.op.input**(1, 256, 256, 3), the second paramter is **net.op.output**(1, 128, 128, 3) and the third paramter **net**(1, 128, 128, 3)

dkurt: > the second paramter is the output of maxpool

I think that this is not an output values but indices.

modules/dnn/src/torch/torch_importer.cpp
dkurt: `setKSize` sets parameters to `poolingLayer.second->params`. You can use any names inside the layer.

modules/dnn/src/layers/max_unpooling_layer.cpp
vonchenplus: we need use the first and second parameter to calc indices. This code works fine on my side.

modules/dnn/src/torch/torch_importer.cpp
vonchenplus: My intention is to reuse setKSize and setStrides method.

modules/dnn/src/layers/max_unpooling_layer.cpp
```diff
@@ -71,9 +78,63 @@ class MaxUnpoolLayerImpl CV_FINAL : public MaxUnpoolLayer
...
+                                    }
+                                }
+                                if (index == kernel_area)
+                                    indice_arr[i_c].at<float>(i_h / poolKernel.height, i_w / poolKernel.width) = -1;
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
dkurt: Let's try to trigger existing implementation. Usage of hardcoded thresholds is too unstable.

Out poolling layer can produce indices of maximum values if it has consumers to it's second pin (the first one are values). So you need to use `connect` to connect Unpooling layer to it's input (some layer) and the 1st pin of pooling layer. To be sure that you connected everything right I'd like to suggest to add one more layer in you network (i.e. 1x1 convolution):

```
input ---+---> pooling -> convolution -> unpooling
         |                                    ^
         +------------------------------------+
```

vonchenplus: OK, I wil try it.

vonchenplus: How can i connect to maxpool's maxid result?  
when i connect maxpool like this:
`connect(layer_id, dstNet, parsePin(pool_name), id, 0);`
the maxpool's output size will be set to 1 and computeMaxIdx be set to false.

modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -1366,6 +1366,43 @@ void TFImporter::populateNet(Net dstNet)
...
+
+            layerParams.set("pool", "max");
+
+            setKSize(layerParams, layer);
+            setStrides(layerParams, layer);
+
+            if (hasLayerAttr(layer, "padding"))
+            {
+                const std::string padding = getLayerAttr(layer, "padding").s();
+                layerParams.set("pad_mode", padding);
```
dkurt: Please do not set unused parameters: "pool", "pad_mode"

vonchenplus:
hello @dkurt, current implement is work well, but a little different from tensorflow.
for example:
Input size: (1, 3, 525, 700)
Tensorflow Output size is: (1, 3, 525, 700)
Our Output size is:(1, 3, 524, 700)
I think i should need to fix this problem, any solution?

dkurt:
@vonchenplus, For which parameters you have this shapes? Please share input resolution, kernel size, strides and paddings.

vonchenplus:
for example:
`inp = tf.placeholder(tf.float32, [1, 525, 700, 3], 'input')`
`pool = tf.layers.max_pooling2d(inp, pool_size=(2, 2), strides=(2, 2))`
`conv = tf.layers.conv2d(inputs=pool, filters=3, kernel_size=[1, 1], padding='VALID')`
`unpool = MaxUnPooling2D(pool.op, conv)`

_layer.input(0) == pool.op.input == [1, 525, 700, 3]
layer,input(1) == pool.op.input == [1, 262, 350, 3]
layer.input(2) == conv == [1, 262, 350, 3]
**tensorflow's** unpool size is [1, 525, 700, 3], **current implement** size is [1, 524, 700, 3]_

the first input has this shapes.

dkurt:
Can you please share text graph so we can see if there are useful parameters in TF graph. You can do it in Python:
```python
for node in graph_def.node:
  print(node)
```

vonchenplus:
hello @dkurt, this is maxpoolgrad node info.

**name: "MaxPoolGrad"
op: "MaxPoolGrad"
input: "X_content"
input: "max_pooling2d/MaxPool"
input: "conv2d/BiasAdd"
attr {
  key: "T"
  value {
    type: DT_FLOAT
  }
}
attr {
  key: "data_format"
  value {
    s: "NHWC"
  }
}
attr {
  key: "ksize"
  value {
    list {
      i: 1
      i: 2
      i: 2
      i: 1
    }
  }
}
attr {
  key: "padding"
  value {
    s: "VALID"
  }
}
attr {
  key: "strides"
  value {
    list {
      i: 1
      i: 2
      i: 2
      i: 1
    }
  }
}**

vonchenplus:
hello @dkurt , Is there any suggestion?

modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -1366,6 +1366,41 @@ void TFImporter::populateNet(Net dstNet)
...
+                else if (padding == "SAME")
+                {
+                    if (layerParams.has("kernel_h"))
+                    {
+                        layerParams.set("pad_h", layerParams.get("kernel_h"));
+                    }
+                    if (layerParams.has("kernel_w"))
+                    {
+                        layerParams.set("pad_w", layerParams.get("kernel_w"));
+                    }
```
dkurt: Looks like `kernel_h` and `kernel_w` are expected here?

modules/dnn/src/layers/max_unpooling_layer.cpp
```diff
@@ -43,13 +43,18 @@ class MaxUnpoolLayerImpl CV_FINAL : public MaxUnpoolLayer
...
 
         MatShape outShape = inputs[0];
         outShape[2] = (outShape[2] - 1) * poolStride.height + poolKernel.height - 2 * poolPad.height;
         outShape[3] = (outShape[3] - 1) * poolStride.width + poolKernel.width - 2 * poolPad.width;
 
+        if (inputs.size() == 3)
+        {
+            outShape = inputs[2];
+        }
+
```
dkurt: That's good decision. I also couldn't find any other alternatives.

So let's rewrite it in this way:

```cpp
MatShape outShape;
if (inputs.size() == 2)
{
  outShape = inputs[0];
  outShape[2] = (outShape[2] - 1) * poolStride.height + poolKernel.height - 2 * poolPad.height;
  outShape[3] = (outShape[3] - 1) * poolStride.width + poolKernel.width - 2 * poolPad.width;
}
else
  outShape = inputs[2];

outputs.clear();
outputs.push_back(outShape);
```

modules/dnn/src/layers/max_unpooling_layer.cpp
```diff
@@ -27,9 +27,9 @@ class MaxUnpoolLayerImpl CV_FINAL : public MaxUnpoolLayer
     MaxUnpoolLayerImpl(const LayerParams& params)
     {
         setParamsFrom(params);
-        poolKernel = Size(params.get<int>("pool_k_w"), params.get<int>("pool_k_h"));
-        poolPad = Size(params.get<int>("pool_pad_w"), params.get<int>("pool_pad_h"));
-        poolStride = Size(params.get<int>("pool_stride_w"), params.get<int>("pool_stride_h"));
+        poolKernel = Size(params.get<int>("kernel_w"), params.get<int>("kernel_h"));
```
dkurt: Test https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/22041 is failed:
```
C++ exception with description "OpenCV(3.4.7-pre) /build/precommit_linux64/3.4/opencv/modules/dnn/include/opencv2/dnn/dnn.inl.hpp:333: error: (-204:Requested object was not found) Required argument "kernel_h" not found into dictionary in function 'get'
" thrown in the test body.
```

Please check if we can always set kernel sizes at tf_importer. If not - we need some default value.

modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -1366,6 +1366,24 @@ void TFImporter::populateNet(Net dstNet)
...
+        else if (type == "MaxPoolGrad")
+        {
+            CV_Assert(layer.input_size() == 3);
+
+            layerParams.set("pool_k_h", 0);
+            layerParams.set("pool_k_w", 0);
+            layerParams.set("pool_stride_h", 0);
+            layerParams.set("pool_stride_w", 0);
+            layerParams.set("pool_pad_h", 0);
+            layerParams.set("pool_pad_w", 0);
```
dkurt: Wait a minute, you already printed text graph with `ksize` and `strides`. So we can use it?

https://github.com/opencv/opencv/pull/14860#issuecomment-505729393

modules/python/package/cv2/__init__.py
```diff
@@ -39,7 +39,7 @@ def bootstrap():
     l_vars = locals()
 
     if sys.version_info[:2] < (3, 0):
-        from cv2.load_config_py2 import exec_file_wrapper
+        from cv2. load_config_py2 import exec_file_wrapper
```
dkurt: Do we need this change?

vonchenplus: I do not changes these. This problem occurred after my rebase.

alalek:
As usual avoid using of merge commits in patches.
Use rebase instead (`git pull --rebase` or `git rebase -i updated_upstream_head`).

To fix that:
- save you work + `git branch -f saved_work`
- setup `upstream` remote by following [this GitHub guide](https://help.github.com/articles/configuring-a-remote-for-a-fork/)
- fetch fresh upstream code: `git fetch upstream`.
- rebase your commits onto actual 3.4 branch:
    `git rebase -i upstream/3.4`
    (check list of **your** commits, save and quit (Esc + "wq" + Enter)
- resolve conflicts if any. To continue run `git rebase --continue`.
- check commits and diffs (`gitk` is a nice tool for that: `gitk --all --date-order &`)
- check files with your saved copy from step 1.

Note: no needs to re-open PR, apply changes "inplace". Do NOT remove your fork or your branch in GitHub - this would break/close this PR completely.

modules/dnn/src/tensorflow/tf_importer.cpp
vonchenplus: No, i do not think so, the shape calculated using ksize and strides may not be correct.

vonchenplus:
hello @alalek, I fixed this problem, thinks.

modules/dnn/src/tensorflow/tf_importer.cpp
dkurt: Kernels and strides are used in forward pass. To calculate output shape we use `output[2]`.

vonchenplus: I use input1’s output2(maxpool's index) directly, so I don't need to recalculate it.

dkurt: Oh, yes, I sorry. `MaxUnpoolLayerImpl::forward` just doesn't use them at all.

dkurt:
Please add skip for Inference Engine backend tests (you may see other tests for reference):

failed tests: https://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/2327

vonchenplus:
Why does the test fail? the maxpool's index is not right? how do i skip the Inference Engine backend tests.

PR #14862, opened 2019-06-21
Add Ultra Wide Intrinsics implementation for BGR2Gray functions.

ChipKerchner:
Add Ultra Wide Intrinsics implementation for BGR2Gray functions. 
 2.6x faster on Power and 40% faster on Intel x64.

```
force_builders=linux,docs
```


ChipKerchner:
Ok, then there needs to be a feature request (?) to have these functions call the cvtColor APIs.

savuor:
I think there can be even a pull request which removes color conversions from `utils.cpp/hpp` and calls cvtColor instead

PR #14888, opened 2019-06-25
Sample app to demonstate nvof in opencv

vchiluka5:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/2165

Adds sample app and app which compares time taken by all OpenCV GPU based algorithms.
This change adds a sample app to demonstrate the use of the NVIDIA Optical flow interface class. It also adds a call to NvidiaHWOpticalFlow in the sample app from optical_flow.cpp which compares the time taken for calculating optical flow
using the existing OpenCV GPU based algorithms.

```
allow_multiple_commits=1
buildworker:Custom=linux-4
```


alalek:
Move sample to corresponding opencv_contrib PR: https://github.com/opencv/opencv_contrib/pull/2165
(see other contrib modules, like ximgproc or ovis)

vchiluka5:
I see that sample apps for other optical flow algorithms resides in samples/gpu. Is this not the correct place to add sample app for NVIDIA's optical flow also?

alalek:
Yes, these samples are here from OpenCV 2.x/3.x versions.
These samples are candidates for moving into opencv_contrib, like CUDA modules do (just need some time to revise these samples).

vchiluka5:
Okay. Can you please suggest appropriate location where should these samples be submitted? I dont see any samples for cpp in opencv_contrib branch

alalek:
New sample can be placed under `<opencv_contrib>/modules/cudaoptflow/samples`.

Samples from `samples/gpu` should be sorted by corresponding CUDA modules.
But some of them are very basic/generic (may be we should put them under `cudev` modules).

vchiluka5:
Okay Thanks.  Il move both files (nvidia_optical_flow.cpp and optical_flow.cpp) to <opencv_contrib>/modules/cudaoptflow/samples and update the PR

vchiluka5:
One more question i had here was we use CUDA driver API's which requires CUDA_CUDA_LIBRARY linked to opencv_world while building and compiling. After moving only samples(2 files) to opencv_contrib, Will this build/compile properly?

alalek:
Does sample using `CUDA_CUDA_LIBRARY` directly?

vchiluka5:
No. Sample app doesn't use it directly. CUDA Driver API's are used in <opencv_contrib>modules/cudaoptflow/src/nvidiaOpticalFlow.cpp

vchiluka5:
I tried compiling it without CUDA_CUDA_LIBRARY but world module gives linker errors. 
 error LNK2019: unresolved external symbol cuCtxGetCurrent referenced in function "public: __cdecl `anonymous namespace'::NvidiaOpticalFlowImpl::NvidiaOpticalFlowImpl(int,int,enum _NV_OF_PERF_LEVEL,bool,bool,bool,int)

vchiluka5:
Thanks @alalek  for submitting test patch which solved linker error.
New error from python bindings. Any suggestions?
https://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/2344/steps/compile%20release/logs/stdio

alalek:
@vchiluka5 Could you try applied workaround on your side?

vchiluka5:
> @vchiluka5 Could you try applied workaround on your side?

@alalek  Tried. Works perfectly. Compiles successfully.

PR #14895, opened 2019-06-26
dnn: configure plugin path for InferenceEngine

alalek:
<cut/>

```
force_builders_only=linux,docs,Custom,Custom Win,Custom Mac

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/op_inf_engine.cpp
+++ b/modules/dnn/src/op_inf_engine.cpp
@@ -409,6 +409,14 @@ void InfEngineBackendNet::initPlugin(InferenceEngine::ICNNNetwork& net)
                 enginePtr = dispatcher.getSuitablePlugin(targetDevice);
             sharedPlugins[targetDevice] = enginePtr;
 
+            std::vector<std::string> candidates;
+
+            std::string param_pluginPath = utils::getConfigurationParameterString("OPENCV_DNN_IE_EXTRA_PLUGIN_PATH", "");
+            if (!param_pluginPath.empty())
+            {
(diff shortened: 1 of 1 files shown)


modules/dnn/src/op_inf_engine.cpp
```diff
@@ -422,24 +430,36 @@ void InfEngineBackendNet::initPlugin(InferenceEngine::ICNNNetwork& net)
...
+                    enginePtr->AddExtension(extension, 0);
+                    CV_LOG_INFO(NULL, "DNN-IE: Loaded extension plugin: " << libName);
+                    found = true;
+                    break;
+                }
+                catch(...) {}
+            }
+            if (!found && !candidates.empty())
+            {
+                CV_LOG_WARNING(NULL, "DNN-IE: Can't load extension plugin (extra layers for some networks). Specify path via OPENCV_DNN_IE_EXTRA_PLUGIN_PATH parameter");
```
dkurt: > // Some of networks can work without a library of extra layers.

What about case when no extensions are found but it's not actually used? I mean cpu_extensions contains custom CPU layers but for some networks it's enough to use basic layers.

alalek: Currently this warning appear only if OpenVINO/IE is not configured properly.
Otherwise this message will be shown one time per model loading (which is tolerable and would help with next error messages).

alalek:
:+1:

PR #14898, opened 2019-06-26
Keep a local copy of `features` from `flann::GenericIndex ` constructor

### LINKED ISSUE #4376: FLANN Index params should be copied to avoid dangling pointer
Transferred from http://code.opencv.org/issues/2050

```
|| Marius Shekow on 2012-06-14 20:23
|| Priority: Normal
|| Affected: None
|| Category: documentation
|| Tracker: Bug
|| Difficulty: None
|| PR: None
|| Platform: None / None
```
## FLANN Index params should be copied to avoid dangling pointer

```
Hi,

in my program I was creating a flann index in one method (assigned to a member variable of type cv::flann::Index), let's say method* A()*, so that I can use it for searching in another method, say *B()*. Example:
<pre>
flannIndex = cv::flann::Index(myData32f, cv::flann::KDTreeIndexParams());
</pre>

However, searching for the k-NN in B() would cause an exception, because the index parameters are deleted at the point when A() exits. I.e.
<pre>
cv::flann::IndexParams::~IndexParams()
{
    delete &get_params(*this);
}
</pre>
is called, which causes a dangling pointer at:
<pre>
void runKnnSearch_(void* index, const Mat& query, Mat& indices, Mat& dists,
                  int knn, const SearchParams& params)
{
   [some code left out]
    ((IndexType*)index)->knnSearch(_query, _indices, _dists, knn,
                                  * (const ::cvflann::SearchParams&)get_params(params)*);
}
</pre>

The current way the FLANN wrapper is coded forces one to either create and use the index in the same method, or to create the index parameter at a safe place (e.g. member variable of your class) such that it is not deleted prematurely. It would be better instead, if the CTOR of cv::flann::Index actually copies the parameter!
```
##### Kirill Kornyakov on 2012-06-15 07:12

```
BTW, are you sure that this is an OpenCV bug, and not the FLANN itself? In the latter case you should contact original authors of the library...
-   Category set to flann
```
##### Marius Shekow on 2012-06-16 07:56

```
Well, the get_params() function essentially converts the cv:flann:*IndexParams to the flann-native index-parameters object. The deletion of the cv:flann:*IndexParams object when getting out of scope causes the dangling pointer, so I'd reckon it's a OpenCV problem.

Also, I'd like to note that it really seems to be only possible to define the cv::flann::Index object in one method but use it in another, if:
* You create the cv::flann::Index on the heap, not the stack
* The cv::Mat data that is indexed is available throughout the lifetime (again, why does the wrapper not internally store the cv::Mat object somewhere such that this is not necessary?)
* The IndexParam object is permanently available (as already noted)
```
##### Vadim Kantorov on 2012-08-01 12:37

```
Marius Shekow has a seemingly complete description.

I just bumped into this bug. Sth like:

@Mat features(...);
cv::flann::KDTreeIndexParams indexParams(5);
cv::flann::Index index(features, indexParams);
index.knnSearch(...);
@
works. While

cv::flann::Index BuildIndex(Mat features)
{
    cv::flann::KDTreeIndexParams indexParams(5);
    cv::flann::Index index(features, indexParams);
    return index;

}

@Mat features(...);
cv::flann::Index index = BuildIndex(features);
index.knnSearch(...);
@
does not because a good copy constructor of cv::flann is not implemented. It would be better to have it somewhere very clear in the documentation, or somehow to notify the user when they try to copy the Index object.
```
##### Andrey Kamaev on 2012-08-15 13:26

```
-   Assignee set to Vadim Pisarevsky
```
##### Andrey Kamaev on 2012-11-23 10:51

```
-   Category changed from flann to documentation
```
##### Vadim Kantorov on 2012-11-26 11:08

```
I guess it would be much better to provide a better wrapper interface for FLANN which will hold a reference to the "features" matrix (via a Mat object, not just a pointer to float array) and to the "params" object (via a shared ptr). The current behavior is confusing and fixing documentation should only be a partial solution from my point of view.
```

### LINKED ISSUE #12606: flann::Index does not copy features matrix
- OpenCV =>3.4.1
- Operating System / Platform =>Windows
- Compiler => VS15 x64

##### Detailed description
When i create a `cv::flann::Index`, it holds a pointer to the `features` matrix data. Calls to `radiusSearch` fail when the `features `matrix is destroyed or changed from outside. `cv::flann::Index` should hold a private copy of the `features` matrix instead.

##### Steps to reproduce

```.cpp
cv::Mat1f features(1, 1, 0.f);
cv::flann::Index flann_index(features, cv::flann::KDTreeIndexParams(1));
cv::Mat indices, dists;
cv::Mat1f query(1, 1, 0.f);
EXPECT_EQ(1, flann_index.radiusSearch(query, indices, dists, 1., 32)); //ok

EXPECT_EQ(features.u->refcount, 1);
features(0) = 100;
EXPECT_EQ(1, flann_index.radiusSearch(query, indices, dists, 1., 32)); //fails
```

UnderscoreAsterisk:
### This pullrequest changes
1. Keep a local copy of `features` matrix in `flann::GenericIndex` constructor. This fixes #4376 and #12606
2. Suppress some warnings in MSVC regarding implicit type conversions

### DIFF
--- a/modules/flann/include/opencv2/flann.hpp
+++ b/modules/flann/include/opencv2/flann.hpp
@@ -289,7 +289,7 @@ public:
 
         int veclen() const { return nnIndex->veclen(); }
 
-        int size() const { return nnIndex->size(); }
+        int size() const { return (int)nnIndex->size(); }
 
         ::cvflann::IndexParams getParameters() { return nnIndex->getParameters(); }
 
--- a/modules/flann/include/opencv2/flann.hpp
+++ b/modules/flann/include/opencv2/flann.hpp
@@ -297,6 +297,7 @@ public:
 
 private:
         ::cvflann::Index<Distance>* nnIndex;
+        Mat _dataset;
 };
 
 //! @cond IGNORED
@@ -312,10 +313,11 @@ private:
 
 template <typename Distance>
 GenericIndex<Distance>::GenericIndex(const Mat& dataset, const ::cvflann::IndexParams& params, Distance distance)
+: _dataset(dataset)
 {
     CV_Assert(dataset.type() == CvType<ElementType>::type());
     CV_Assert(dataset.isContinuous());
-    ::cvflann::Matrix<ElementType> m_dataset((ElementType*)dataset.ptr<ElementType>(0), dataset.rows, dataset.cols);
+    ::cvflann::Matrix<ElementType> m_dataset((ElementType*)_dataset.ptr<ElementType>(0), _dataset.rows, _dataset.cols);
 
     nnIndex = new ::cvflann::Index<Distance>(m_dataset, params, distance);
 
--- a/modules/flann/include/opencv2/flann/dist.h
+++ b/modules/flann/include/opencv2/flann/dist.h
@@ -114,7 +114,7 @@ struct L2_Simple
         ResultType result = ResultType();
         ResultType diff;
         for(size_t i = 0; i < size; ++i ) {
-            diff = *a++ - *b++;
+            diff = (ResultType)(*a++ - *b++);
             result += diff*diff;
         }
         return result;


savuor:
👍

alalek:

So, please:

PR #14899, opened 2019-06-26
 dnn: fix BNLL layer implementation

alalek:
https://github.com/BVLC/caffe/blame/1.0/src/caffe/layers/bnll_layer.cpp#L17

https://github.com/opencv/opencv/pull/14827#discussion_r296468934

Done:
- [x] run Halide tests
- [x] run OpenCL tests


alalek:
:+1:

PR #9893, opened 2017-10-20
Added BUILD_WHITELIST

mshabunin:
### This pullrequest changes

adds ability to build only modules which you really need:
```
> cmake -DBUILD_LIST=dnn ../opencv && make
> ls -1 lib
libopencv_core.so
libopencv_core.so.3.3
libopencv_core.so.3.3.0
libopencv_dnn.so
libopencv_dnn.so.3.3
libopencv_dnn.so.3.3.0
libopencv_imgproc.so
libopencv_imgproc.so.3.3
libopencv_imgproc.so.3.3.0
```
**Notes:**
- modules should be named as folders: _core_, _imgproc_, ...; not _opencv_core_
- modules can be delimited by space, comma, colon, semicolon: `-DBUILD_LIST=dnn,imgcodecs`; don't forget to escape shell's special characters `-DBUILD_LIST=dnn\;imgcodecs`
- to build a _world_ module:
    - add  _imgcodecs_, _videoio_, _highgui_ to the list: `-DBUILD_LIST=dnn,imgcodecs,videoio,highgui`
    - add `-DBUILD_opencv_world=ON` option
- to build tests add a _ts_ module: `-DBUILD_LIST=dnn,ts`
- to build samples add `-DBUILD_EXAMPLES=ON` option and enable dependent modules as follows:
    - _opencv/samples/cpp_ :arrow_right:  `flann,imgcodecs,videoio,highgui,ml,video,objdetect,photo,features2d,calib3d,stitching,videostab,shape`
    - _opencv/samples/dnn_ :arrow_right:  `dnn,imgcodecs,videoio,highgui`
    - _some_module/samples_ :arrow_right: `some_module,imgcodecs,videoio,highgui`
- to build wrappers add them to the list: `-DBUILD_LIST=dnn,imgcodecs,python3,java` (_java_ has hidden dependency on _imgcodecs_, it should be fixed)
- optional dependencies (`ocv_define_module(<module> <required_deps> OPTIONAL <optional_deps>`) are not taken in account so you should enable them manually
- tested with cmake 3.7.2

PR #14903, opened 2019-06-26
Adding Support for RTX64 Vision to OpenCV 3.4

owenmonahan:
Adding support for RTX64 Vision to OpenCV 3.4.  Primarily involves adding a new class derived from IVideoCapture to the VideoIO module.

previous PR: #13288


owenmonahan:
Hi,

Is there a timeline for how long it generally takes to get feedback on a PR?  I ask simply because I am eager to get this addition approved.  Thanks!

owenmonahan:
Thank you very much for the feedback.  Apologies for the confusion with the two PRs, I didn't mean to cause trouble.  I made a new PR, as in the original PR, you instructed us to implement IVideoCapture, instead of using the legacy code, as we were doing before.  We had to postpone doing this for a few months, so by the time I got to come back to this, I thought it would be best to start anew, especially since so much of the code and CMake was changing.  Again sorry for this, it won't happen again.

Though I understand your concerns about the maintainability of our code, going forward, we would still like to get our code into the true OpenCV repository (not just in our fork), if you are willing to allow it (once I make any and all changes/fixes you have requested).  I have attempted to keep all of our code #ifdefed out when in non-RTX64 builds of OpenCV.  Will this help to get our additions approved?

owenmonahan:
This will be difficult for you guys to build unfortunately, @mshabunin.  You would need to have access to our SDK.  If you would like to do this, I may be able to get you guys access to it.  Let me know!

mshabunin:
@owenmonahan , I agree with @alalek , this patch does not have value for majority of the OpenCV community and we can not maintain it. I can see that you want to add custom _VideoCapture_ backend and I propose the following approach:
* create a custom module with your backend, it is simple process - see our [coding guidelines](https://github.com/opencv/opencv/wiki/Coding_Style_Guide) for details. 
  * Take a look at the [fuzzy module](https://github.com/opencv/opencv_contrib/tree/master/modules/fuzzy) for example. 
  * You can add this module to your OpenCV build process by setting `OPENCV_EXTRA_MODULES_PATH` CMake option (it [accepts lists](https://github.com/opencv/opencv/pull/10758) too). 
  * You can limit modules being built by setting `OPENCV_BUILD_LIST` [option](https://github.com/opencv/opencv/pull/9893). 
  * Your custom build can be distributed with your SDK or as a standalone product and repository can be published on GitHub.
  * You can have a single class in you module - `cv::VideoCaptureRTX` which can be compatible with `cv::VideoCapture`.
* those little changes left in your patch after moving _VideoCapture_ backend to separate module should be rewritten in more generic manner (e.g. `#ifdef MY_PLATFORM` should be changed to `#ifdef HAVE_SOME_FEATURE`). Although, I believe it won't be needed at all, since most of options can be easily passed via CMake parameters.

owenmonahan:
@mshabunin , okay, we will go in a different direction then.  We will either do as you suggested, or do as @alalek suggested and maintain our own OpenCV fork.  Thank you very much for the input!  Should we close this PR for now?

mshabunin:
Let's close it for now. Feel free to reopen or create new if you come up with more generic and compact changes.

PR #14916, opened 2019-06-27
Avoid using v_signmask universal intrinsic and mark it as deprecated

terfendail:
### This pullrequest changes
Avoid using v_signmask universal intrinsic and mark it as deprecated

```
force_builders=Custom,Linux AVX2,ARMv7,ARMv8,Linux32,Win32
buildworker:Custom=linux-3
CPU_BASELINE:Custom Win=AVX512_SKX
```

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin_cpp.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_cpp.hpp
@@ -1098,7 +1098,7 @@ v_int32x4 r; // set to {0, 0, -1, -1}
 int idx = v_heading_zeros(r); // idx = 2
 @endcode
 */
-template <typename _Tp, int n> inline int v_find_negative(const v_reg<_Tp, n>& a)
+template <typename _Tp, int n> inline int v_scan_forward(const v_reg<_Tp, n>& a)
 {
     for (int i = 0; i < n; i++)
         if(V_TypeTraits<_Tp>::reinterpret_int(a.s[i]) < 0)
--- a/modules/core/include/opencv2/core/hal/intrin_cpp.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_cpp.hpp
@@ -1072,6 +1072,7 @@ template<typename _Tp, int n> inline typename V_TypeTraits< typename V_TypeTrait
 }
 
 /** @brief Get negative values mask
+@deprecated v_signmask depends on a lane count heavily and therefore isn't universal enough
 
 Returned value is a bit mask with bits set to 1 on places corresponding to negative packed values indexes.
 Example:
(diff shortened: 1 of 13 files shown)


modules/core/include/opencv2/core/hal/intrin_cpp.hpp
```diff
@@ -1088,6 +1089,23 @@ template<typename _Tp, int n> inline int v_signmask(const v_reg<_Tp, n>& a)
...
+/** @brief Get first negative lane index
+
+Returned value is an index of first negative lane (undefined for input of all positive values)
+Example:
+@code{.cpp}
+v_int32x4 r; // set to {0, 0, -1, -1}
+int idx = v_heading_zeros(r); // idx = 2
+@endcode
+*/
+template <typename _Tp, int n> inline int v_find_negative(const v_reg<_Tp, n>& a)
```
alalek: What is about these names?
- `v_find_non_zero()`
- `v_scan_non_zero()`
- `v_scan_forward()`
- `v_scan()`

modules/core/include/opencv2/core/hal/intrin_neon.hpp
```diff
@@ -1103,10 +1103,25 @@ inline int v_signmask(const v_uint64x2& a)
...
+inline int v_find_negative(const v_int8x16& a) { return trailingZeros32(v_signmask(a)); }
+inline int v_find_negative(const v_uint8x16& a) { return trailingZeros32(v_signmask(a)); }
+inline int v_find_negative(const v_int16x8& a) { return trailingZeros32(v_signmask(a)); }
+inline int v_find_negative(const v_uint16x8& a) { return trailingZeros32(v_signmask(a)); }
+inline int v_find_negative(const v_int32x4& a) { return trailingZeros32(v_signmask(a)); }
+inline int v_find_negative(const v_uint32x4& a) { return trailingZeros32(v_signmask(a)); }
+inline int v_find_negative(const v_float32x4& a) { return trailingZeros32(v_signmask(a)); }
+#if CV_SIMD128_64F
+inline int v_find_negative(const v_int64x2& a) { return trailingZeros32(v_signmask(a)); }
+inline int v_find_negative(const v_uint64x2& a) { return trailingZeros32(v_signmask(a)); }
```
alalek: It is strange that `CV_SIMD128_64F` blocks `int64` / `uint64` processing.

ChipKerchner:
Why are we deprecating v_signmask?  It's a useful instruction.

alalek:
There is "wide universal intrinsics (w.u.i.)" API over SIMD128 / SIMD256 / SIMD512 (depends on current compilation flags). So we are trying to unify APIs if possible (however it doesn't work well for reductions).

v_signmask() in case of SIMD512 doesn't fit well into 32-bit int (u8 x 64 elements).

ChipKerchner:
BTW, IMO v_signmask should return an 'unsigned int' and not an 'int'

PR #14917, opened 2019-06-27
G-API planar kernels

rgarnov:
```
```


rgarnov:
@dbudniko please take a look

dbudniko:
@rgarnov @alalek , please review the latest fixes.

modules/gapi/include/opencv2/gapi/core.hpp
```diff
@@ -1346,6 +1356,22 @@ enlarge an image, it will generally look best with cv::INTER_CUBIC (slow) or cv:
...
+
+Output image size will have the size dsize, the depth of output is the same as of src.
+
+@note Function textual ID is "org.opencv.core.transform.resizeP"
+
+@param src input image, must be of @ref CV_8UC3 type;
+@param dsize output image size;
+@param interpolation interpolation method, only cv::INTER_LINEAR is supported at the moment
+
+@sa  warpAffine, warpPerspective, remap
```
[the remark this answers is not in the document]
dbudniko: fixed

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -847,6 +876,36 @@ Output image must be 8-bit unsigned 2-channel image @ref CV_8UC2.
...
+The conventional ranges for Y, U, and V channel values are 0 to 255.
+
+Output image must be 8-bit unsigned planar 3-channel image @ref CV_8UC3.
+
+@note Function textual ID is "org.opencv.imgproc.colorconvert.nv12torgbp"
+
+@param src_y input image: 8-bit unsigned 1-channel image @ref CV_8UC1.
+@param src_uv input image: 8-bit unsigned 2-channel image @ref CV_8UC2.
+
+@sa YUV2RGB, NV12toBGRp, NV12toBGR
```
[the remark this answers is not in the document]
dbudniko: fixed

rgarnov:
Looks good

modules/gapi/include/opencv2/gapi/core.hpp
```diff
@@ -1342,10 +1352,26 @@ enlarge an image, it will generally look best with cv::INTER_CUBIC (slow) or cv:
...
 
+/** @brief Resizes a planar image.
+
+The function resizes the image src down to or up to the specified size.
+
+Output image size will have the size dsize, the depth of output is the same as of src.
+
+@note Function textual ID is "org.opencv.core.transform.resizeP"
+
+@param src input image, must be of @ref CV_8UC3 type;
```
alalek: > planar image
> CV_8UC3

`CV_8UC3` means interleaved image. So `planar` is source of confusion here.
planar is src1, src2 and src3 with `CV_8UC1` type.

What is the main difference between `resize` and `resizeP`? Missing `fx`, `fy`?
Do we really need this entry point?

---

How do you want to downscale image by 2 with odd source size?
9x9 => 4.5 x 4.5 as 4x4 - is not 2x downscaling without extra information.

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -847,6 +876,36 @@ Output image must be 8-bit unsigned 2-channel image @ref CV_8UC2.
 */
 GAPI_EXPORTS GMat RGB2YUV422(const GMat& src);
 
+/** @brief Converts an image from NV12 (YUV420p) color space to RGB.
+The function converts an input image from NV12 color space to RGB.
+The conventional ranges for Y, U, and V channel values are 0 to 255.
+
+Output image must be 8-bit unsigned planar 3-channel image @ref CV_8UC3.
```
alalek: > planar 3-channel

`CV_8UC3` is interleaved 3-channel.

modules/gapi/include/opencv2/gapi/core.hpp
rgarnov: The main difference is that `resizeP` has `GMatP` ("planar GMat") as input and as output so the engine can behave correctly.
This description definitely should be changed to reflect planar image memory layout (three planes laying in the memory contiguously), so the image height should be `plane_height`*`plane_number`, image type is `CV_8UC1` and so on. Maybe planar images deserve some special chapter?

dmatveev:
> Instead of playing with size, you can split planes completely:
> planeB, planeG, planeR

For sure, and this is how our preprocessing works now on CPU. We do use a "split3" thing with its output `Mat`s redirected properly to form a 3-channel buffer in the CHW layout.

`split3` with three distinct outputs has both advantages (e.g. we can compose the output plane order for free in a way we need) and disadvantages (it doesn't put any restrictions on its output memory chunks - so those could be three different bufferfs).

On some platform which we currently enable with G-API, the operations like color conversions and resize operate with planar images natively. However, that operations assume only a single data pointer (with a fixed offset between the planes), not three different ones. So from the API perspective it is safer to introduce "planar" image type there rather than simulate it with three data objects pointing to the same memory chunk

alalek:
3-channel color image:
- interleaved: 2D matrix `{height, width}` with `CV_8UC3` element type
- your concatenated planar: 3D matrix `{channels, height, width}` with `CV_8UC1` element type

Not sure why you need some strange `GMatP` workarounds with implicit number of channels (`plane_height*plane_number`).
Regular 3D matrix is able properly handle this layout with explicit number of channels in `size[0]`.

modules/gapi/include/opencv2/gapi/imgproc.hpp
dbudniko: I've upadted doxygen comments. I suggest to merge it as is to move forward.

dbudniko:
@dmatveev please approve. @alalek is waiting for it to merge.

dmatveev:
> Not sure why you need some strange GMatP workarounds with implicit number of channels (plane_height*plane_number).

It is mainly an API contract. All our existing kernels which work with `GMat`s in their API do not expect planar input, so we decided to forbid passing possibly-planar data there to avoid issues and confusion. `GMat/GMatP` clearly separates kernels on a type level.

dmatveev:
> Regular 3D matrix is able properly handle this layout with explicit number of channels in size[0].

This is a good point, BTW! Probably later we'll adjust API to use multi-dimensional matrices for the planar case.

dmatveev:
Thanks!

PR #14929, opened 2019-06-29
cmake: support rpath-link linker option

alalek:
- builds with CMAKE_SKIP_RPATH=ON should properly find project .so files from <build>/lib directory

Details are below.

<cut/>

```
cmake -DCMAKE_SKIP_RPATH=ON -DWITH_TBB=ON -DBUILD_TBB=ON ...
```
(`CMAKE_SKIP_RPATH` is widely used in OpenCV's cross-compilation toolchains - perhaps to avoid overriding of `LD_LIBRARY_PATH` settings)

Build produces `libtbb.so` in `<build>/lib`, `libopencv_core.so` in `<build>/lib`.
But build may fail linking of apps/tests (`opencv_version`) with message:
```
warning: libtbb.so, needed by ../../lib/libopencv_core.so.3.4.7, not found (try using -rpath or -rpath-link)
```
or even worse, it may grab libtbb.so from the system with wrong version (`-Wl,--verbose` output):
```
libtbb.so needed by lib/libopencv_core.so.3.4.7
attempt to open //usr/lib64/llvm5.0/lib/libtbb.so failed
attempt to open //usr/lib64/llvm6.0/lib/libtbb.so failed
...
attempt to open //usr/x86_64-redhat-linux/lib64/libtbb.so failed
found libtbb.so at //usr/lib64/libtbb.so
```
(Launch with `LD_LIBRARY_PATH` selects `libtbb.so` from `<build>/lib`)

Fix adds this linker parameter `-Wl,-rpath-link,"\$ORIGIN"` and linker is able to find the proper version of libtbb.so
```
libtbb.so needed by lib/libopencv_core.so.3.4.7
found libtbb.so at /home/alalek/projects/opencv/build/opencv/lib/libtbb.so
```

Other system .so files are searched in this way:
```
libz.so.1 needed by lib/libopencv_core.so.3.4.7
attempt to open /home/alalek/projects/opencv/build/opencv/lib/libz.so.1 failed
...
attempt to open //usr/x86_64-redhat-linux/lib64/libz.so.1 failed
found libz.so.1 at //usr/lib64/libz.so.1
```

---

Done:
- [x] Validated with CMake 2.8.12.2
- `$ORIGIN` may not work (with cross-compilation?)

Alternatives:
- CMake 3.14 has `CMAKE_BUILD_RPATH_USE_ORIGIN` [option](https://gitlab.kitware.com/cmake/cmake/merge_requests/2456)


alalek:
:+1:

mshabunin:
I don't think we need this fix:
* when user builds OpenCV with `CMAKE_SKIP_RPATH=ON`it is theirs responsibility to set up runtime/build environment (e.g. by setting `LD_LIBRARY_PATH=<opencv>/lib` variable)
* it uses private undocumented cmake variable: `CMAKE_PLATFORM_RUNTIME_PATH`
* there are options to disable RPATH separately for build and install configurations, which are not taken in account (`CMAKE_SKIP_INSTALL_RPATH`)
* some unrelated binaries can be affected by this patch: python/java library, videoio plugins

My suggestion is to recommend users either set up environment correctly when they remove RPATH from binaries, either correctly set all required options during build (`CMAKE_INSTALL_PREFIX`, `CMAKE_INSTALL_RPATH`, etc.)

PR #14930, opened 2019-06-29
Final fix for JS sample of dnn

dkurt:


alalek:
Re-create PR from your fork. User branches in upstream repository are forbidden.

PR #14931, opened 2019-06-29
Final fix for JS sample of dnn

dkurt:


dkurt:
@alalek, This sample can be found here: https://docs.opencv.org/master/d5/d86/tutorial_dnn_javascript.html

PR #14934, opened 2019-06-29
Explicitly default operator= for Vec<T, n>

### LINKED ISSUE #14933: Defaulted operator= for cv::Vec<T, N> should be made explicit
GCC9 issues many warnings like:
```
./opencv/modules/imgproc/src/templmatch.cpp: In function ‘bool cv::matchTemplate_CCOEFF(cv::InputArray, cv::InputArray, cv::OutputArray)’:
./opencv/modules/imgproc/src/templmatch.cpp:464:38: warning: implicitly-declared ‘constexpr cv::Vec<float, 4>& cv::Vec<float, 4>::operator=(const cv::Vec<float, 4>&)’ is deprecated [-Wdeprecated-copy]
  464 |         templ_sum = (Vec4f)mean(templ);
      |                                      ^
In file included from ./opencv/modules/core/include/opencv2/core.hpp:57,
                 from ./opencv/modules/imgproc/include/opencv2/imgproc.hpp:46,
                 from /./opencv/build/modules/imgproc/precomp.hpp:46:
./opencv/modules/core/include/opencv2/core/matx.hpp:1030:1: note: because ‘cv::Vec<float, 4>’ has user-provided ‘cv::Vec<_Tp, cn>::Vec(const cv::Vec<_Tp, cn>&) [with _Tp = float; int cn = 4]’
 1030 | Vec<_Tp, cn>::Vec(const Vec<_Tp, cn>& m)
      | ^~~~~~~~~~~~
```
openCV should follow the "rule of five", see e.g.
http://www.stroustrup.com/C++11FAQ.html#default

It is sufficient to explicitly tell the compiler to default the assignment operator, i.e.
```
    constexpr Vec<_Tp, cn>& operator=(const Vec<_Tp, cn>& rhs) = default;
``` 

- OpenCV => 3.4.6, 4.1.0
- Operating System / Platform => Linux 64 Bit, openSUSE Tumleweed
- Compiler => GCC 9 (gcc (SUSE Linux) 9.1.1 20190611 [gcc-9-branch revision 272147])

StefanBruens:
Due to the explicitly declared copy constructor Vec<T, n>::Vec(Vec <T,n>&)
GCC 9 warns if there is no assignment operator, as having one typically
requires the other (rule-of-three, constructor/desctructor/assginment).

As the values are just a plain array the default assignment operator does
the right thing. Tell the compiler explicitly to default it.

resolves #14933

### DIFF
--- a/modules/core/include/opencv2/core/matx.hpp
+++ b/modules/core/include/opencv2/core/matx.hpp
@@ -391,6 +391,10 @@ public:
     const _Tp& operator ()(int i) const;
     _Tp& operator ()(int i);
 
+#ifdef CV_CXX11
+    Vec<_Tp, cn>& operator=(const Vec<_Tp, cn>& rhs) = default;
+#endif
+
     Vec(const Matx<_Tp, cn, 1>& a, const Matx<_Tp, cn, 1>& b, Matx_AddOp);
     Vec(const Matx<_Tp, cn, 1>& a, const Matx<_Tp, cn, 1>& b, Matx_SubOp);
     template<typename _T2> Vec(const Matx<_Tp, cn, 1>& a, _T2 alpha, Matx_ScaleOp);


modules/core/include/opencv2/core/matx.hpp
```diff
@@ -391,6 +391,8 @@ template<typename _Tp, int cn> class Vec : public Matx<_Tp, cn, 1>
     const _Tp& operator ()(int i) const;
     _Tp& operator ()(int i);
 
+    constexpr Vec<_Tp, cn>& operator=(const Vec<_Tp, cn>& rhs) = default;
```
alalek: 1. `#ifdef CV_CXX11` (due `= default`)
2. Do we really need `constexpr` here? if so, please use `CV_CONSTEXPR` instead. IMHO, it is better to remove `contexpr` here completely (see Windows builder warnings)
3. What is about move assignment operator (`Vec<_Tp, cn>&& rhs`)?

StefanBruens: According to:
http://eel.is/c++draft/dcl.fct.def.default#3
> If a function is explicitly defaulted on its first declaration, it is implicitly considered to be constexpr if the implicit declaration would be.

As it *is* implicit constexpr, the "constexpr" can be removed without loss for the relevant cases (you can of course have a Vec<Foo, 1>, where Foo would lead to non-constexpr).

move assignments and constructors are implicitly "deleted" already, as there is a user-declared copy constructor. Both would be pointless for an array on the stack anyway (you can not efficiently move/swap these, not anymore than a memcpy would do).

PR #14945, opened 2019-07-01
G-API: clean up accuracy tests

andrey-golubev:
### This pullrequest

- Deletes createOutputMatrices flag
- Deletes custom printers for test fixtures. Introduces unified approach through `operator<<` overloading:
  - CV type is now properly printed: `CV_8U` printed as "CV_8U" instead of "0"
  - Enum values are also printed correctly: `NormTypes::NORM_INF` printed as "NORM_INF"
  - Comparison functions are pretty-printed: `AbsTolerance(1)` printed as "AbsTolerance(1)" instead of "compare_f"


andrey-golubev:
@dmatveev @rgarnov please review

modules/gapi/test/common/gapi_core_tests.hpp
```diff
@@ -37,11 +37,22 @@ const char *BitwiseOperations[] = {"And", "Or", "Xor"};
 const char *CompareOperations[] = {"CMP_EQ", "CMP_GT", "CMP_GE", "CMP_LT", "CMP_LE", "CMP_NE"};
 //corresponds to OpenCV
 const char *NormOperations[] = {"", "NORM_INF", "NORM_L1", "","NORM_L2"};
+// helper function that returns correct string representation for dtype
+const auto dtypeToStr = [] (int v) { return (v == -1) ? std::string("neg1") : std::to_string(v); };
```
dmatveev: maybe `"SAME"` is better than `"neg1"`?

modules/gapi/test/cpu/gapi_core_tests_cpu.cpp
```diff
@@ -11,7 +11,7 @@
 
 namespace
 {
-    #define CORE_CPU [] () { return cv::compile_args(cv::gapi::core::cpu::kernels()); }
+const auto getArgsCoreCpu = [] () { return cv::compile_args(cv::gapi::core::cpu::kernels()); };
```
dmatveev: I belive the old naming was better. It least you see clearly that this test is to run on CPU (Fluid, OCL, etc).
Now one should care what `getArgsCoreCpu` means.

andrey-golubev: reverted to previous version

modules/gapi/test/common/gapi_core_tests.hpp
andrey-golubev: deleted this stuff

andrey-golubev:
@dmatveev please see updated version

Regarding:
> Do we have a follow-up task on testing the output data allocation in our JIRA?

-- I plan to add it to JIRA and do separately

andrey-golubev:
@dmatveev added test on output data allocation - https://github.com/opencv/opencv/pull/14985

modules/gapi/test/common/gapi_core_tests.hpp
```diff
@@ -133,17 +77,17 @@ GAPI_TEST_FIXTURE(MinTest, initMatsRandU, <>, 0)
...
 GAPI_TEST_FIXTURE(IntegralTest, initNothing, <>, 0)
 GAPI_TEST_FIXTURE(ThresholdTest, initMatrixRandU, FIXTURE_API(int), 1, tt)
 GAPI_TEST_FIXTURE(ThresholdOTTest, initMatrixRandU, FIXTURE_API(int), 1, tt)
 GAPI_TEST_FIXTURE(InRangeTest, initMatrixRandU, <>, 0)
 GAPI_TEST_FIXTURE(Split3Test, initMatrixRandU, <>, 0)
 GAPI_TEST_FIXTURE(Split4Test, initMatrixRandU, <>, 0)
-GAPI_TEST_FIXTURE(ResizeTest, initNothing, FIXTURE_API(compare_f,int,cv::Size), 3,
+GAPI_TEST_FIXTURE(ResizeTest, initNothing, FIXTURE_API(CompareMats,int,cv::Size), 3,
     cmpF, interp, sz_out)
 GAPI_TEST_FIXTURE(ResizePTest, initNothing, FIXTURE_API(compare_f,int,cv::Size), 3,
```
rgarnov: There are few `compare_f`s appeared after recent merge (resizeP, nv12torgbP)

andrey-golubev: should be fixed now, thanks for pointing out!

andrey-golubev:
@dmatveev please look into this

andrey-golubev:
@dmatveev proper output allocation tests are introduced here - #15012

modules/gapi/test/common/gapi_tests_common.hpp
```diff
@@ -318,6 +354,11 @@ class Tolerance_FloatRel_IntAbs : public Wrappable<Tolerance_FloatRel_IntAbs>
             }
         }
     }
+    std::string desc() {
```
dmatveev: y desc?

I believe this method could take an `ostream&` as parameter and write its stuff there.

Then there's no matter if it is a std::cout our some stringstream - we use that stream API native to write our stuff out.

modules/gapi/test/cpu/gapi_core_tests_cpu.cpp
```diff
@@ -143,16 +128,14 @@ INSTANTIATE_TEST_CASE_P(Cart2PolarCPU, Cart2PolarTest,
...
 INSTANTIATE_TEST_CASE_P(PhaseCPU, PhaseTest,
                         Combine(Values(CV_32F, CV_32FC3),
                                 Values(cv::Size(1280, 720),
                                        cv::Size(640, 480),
                                        cv::Size(128, 128)),
-                                Values(SAME_TYPE),
-                                Values(true),
+                                Values(-1),
                                 Values(CORE_CPU),
                                 testing::Bool()));
```
dmatveev: what is this bool btw?

modules/gapi/test/gpu/gapi_operators_tests_gpu.cpp
```diff
@@ -58,17 +55,15 @@ INSTANTIATE_TEST_CASE_P(BitwiseOperatorTestGPU, MathOperatorMatScalarTest,
                                        cv::Size(640, 480),
                                        cv::Size(128, 128)),
                                 Values(-1),
-/*init output matrices or not*/ testing::Bool(),
                                 Values(CORE_GPU),
-                                Values(AbsExact().to_compare_f()),
+                                Values(AbsExact().to_compare_obj()),
```
dmatveev: not sure if this rename is necessary, `to_compare_f` could work fine (where `f` is for functor).

andrey-golubev: I don't want to break perf tests so this is the "compatibility" mode

modules/gapi/test/cpu/gapi_core_tests_cpu.cpp
andrey-golubev: whether the calculated angle is in degrees or in radians

modules/gapi/test/common/gapi_tests_common.hpp
andrey-golubev: makes sense, will update

andrey-golubev: I suggest to overload `operator<<` then. It seems to be the most robust option for such stuff + we get printability of comparators even without introduced `CompareF`

modules/gapi/test/cpu/gapi_core_tests_cpu.cpp
dmatveev: Ok, thanks. Hard to get it without a comment (not your fault though)

modules/gapi/test/common/gapi_tests_common.hpp
dmatveev: you can overload that `operator<<` using that `ostream&`-oriented method. Bingo!

dmatveev: Going further:

```c++
template<typename T> PP { const T &t; };
template<typename T> std::ostream& operator<< (std::ostream&os, const PP& pp) {
    pp.print_desc(os);
    return os;
}
```
So then you can dump a whatever tolerance with 
```c++
... << PP{tol} << ...
```
Probably it works even this way.

modules/gapi/test/gpu/gapi_operators_tests_gpu.cpp
dmatveev: ok!

modules/gapi/test/common/gapi_core_tests.hpp
```diff
@@ -30,91 +30,35 @@ enum bitwiseOp
...
-                    <<"_"<<sz.width
-                    <<"x"<<sz.height
-                    <<"_"<<AllParams::getCommon<3>(params);  // createOutputMatrices
-        return ss.str();
-   }
-};
+        CASE(ADD);
+        CASE(SUB);
+        CASE(MUL);
+        CASE(DIV);
```
dmatveev: missing default?

modules/gapi/test/common/gapi_core_tests.hpp
```diff
@@ -30,91 +30,35 @@ enum bitwiseOp
...
-                    <<"_"<<sz.width
-                    <<"x"<<sz.height;
-        return ss.str();
-   }
-};
+        CASE(AND);
+        CASE(OR);
+        CASE(XOR);
+        CASE(NOT);
+    }
```
dmatveev: missing default?

modules/gapi/test/common/gapi_tests_common.hpp
dmatveev: But you already did the operators - ok!

modules/gapi/test/common/gapi_core_tests.hpp
andrey-golubev: hmm, nothing bad should happen AFAIK. but nothing good either (except the warning). I agree with you, seems wise to add `default: throw () ...` just so that it's easier to see if some value is missing. will update

dmatveev: Let it be CV_Assert.

modules/gapi/test/cpu/gapi_core_tests_cpu.cpp
andrey-golubev: added comment

modules/gapi/test/common/gapi_core_tests.hpp
andrey-golubev: I think CV_Assert is more about internal OpenCV error-reporting (am I wrong?). Added `throw runtime_error()` for now, can redo with CV_Assert if required

andrey-golubev:
@dmatveev updated, please look

modules/gapi/test/common/gapi_core_tests.hpp
dmatveev: CV_Assert does exactly the same. Just use CV_Assert

dmatveev: Note we also have our own GAPI_Assert (which used to be more portable when we supported LeonOS directly).
```
GAPI_Assert(false && "your text");
```

dmatveev:
Thanks!

modules/gapi/test/common/gapi_core_tests.hpp
andrey-golubev: @dmatveev done with GAPI_Assert

andrey-golubev:
@alalek I believe this PR is done, please merge

andrey-golubev:
Thanks!

PR #14946, opened 2019-07-01
G-API: Use object instead of reference in extended lifetime case

andrey-golubev:
### This pullrequest changes

const lvalue reference to object since in the given expression a new object is returned from a function. 

Original code is a valid behavior since this is the case for extended lifetime of the temporary object  (const lvalue reference of the same type is bound to that temporary).

Anyway, the code is doubtful and requires the developer to check for the return type and provide some rationale to ensure the code is correct. Taking this into account, it's easier to replace reference with an object.

---
References: [Temporary Object Lifetime](https://en.cppreference.com/w/cpp/language/lifetime) - look at the exceptions from the rules


andrey-golubev:
@dmatveev please review

dmatveev:
It still could be (and must be) `const`.

`const` is a great hint that the value is never modified in the function and you don't need to track where it may be written to once it is defined as `const`.

andrey-golubev:
@dmatveev totally makes sense to add `const`, updated

andrey-golubev:
@alalek please merge!

PR #14952, opened 2019-07-01
G-API: GAPI_TRANSFORM internal functionality rework

smirnov-alexey:
This pullrequest changes internal functions signatures and changes some tests

### DIFF
--- a/modules/gapi/include/opencv2/gapi/garray.hpp
+++ b/modules/gapi/include/opencv2/gapi/garray.hpp
@@ -55,6 +55,12 @@ namespace detail
     class VectorRef;
     using ConstructVec = std::function<void(VectorRef&)>;
 
+    // This is the base struct for GArrayU type holder
+    struct TypeHintBase{virtual ~TypeHintBase() = default;};
+
+    // This class holds type of initial GArray to be checked from GArrayU
+    template <typename T>
+    struct TypeHint final : public TypeHintBase{};
 
     // This class strips type information from GArray<T> and makes it usable
     // in the G-API graph compiler (expression unrolling, graph generation, etc).
@@ -63,6 +69,9 @@ namespace detail
     {
     public:
         GArrayU(const GNode &n, std::size_t out); // Operation result constructor
+
+        template <typename T>
+        bool holds() const;                       // Check if was created from GArray<T>
 
         GOrigin& priv();                          // Internal use only
         const GOrigin& priv() const;              // Internal use only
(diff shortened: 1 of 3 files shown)


smirnov-alexey:
@rgarnov please, review. Need to figure out what to do with tests

smirnov-alexey:
Thanks, @dmatveev !

modules/gapi/test/gapi_transform_tests.cpp
```diff
@@ -103,81 +116,74 @@ TEST(KernelPackageTransform, Combine)
...
-{
-    transformTest<garr_in_gmat_out, GArray<int>, GMat>();
+    check(tr.pattern());
+    check(tr.substitute());
 }
 
-TEST(KernelPackageTransform, DISABLED_GArrayOut)
+TEST(KernelPackageTransform, garr_in_gmat_out)
 {
-    transformTest<gmat_in_garr_out, GMat, GArray<int>>();
```
rgarnov: I propose to make a common test body for 1 input -> 1 output transformations (like `transformTest` was) and to test all possible in-out type combinations (check for GArray and be extended via template specialization of the check function)

modules/gapi/include/opencv2/gapi/garray.hpp
```diff
@@ -73,7 +82,23 @@ namespace detail
 
         void setConstructFcn(ConstructVec &&cv);  // Store T-aware constructor
 
+        template <typename T>
+        void specifyType();                       // Store type of initial GArray<T>
+
         std::shared_ptr<GOrigin> m_priv;
+        std::shared_ptr<TypeHintBase> m_hint;
```
rgarnov: can it be `unique_ptr`?

rgarnov: And the name should reflect it's purpose (I mean should have `type` related words in it)

dmatveev: > can it be unique_ptr?

No, since GArrayU objects are copyable, and it is easier to share a type hint in this case.

modules/gapi/include/opencv2/gapi/gtransform.hpp
```diff
@@ -16,13 +16,16 @@
...
 
 struct GAPI_EXPORTS GTransform
 {
-    using F = std::function<GArgs(const GArgs &)>;
+    // FIXME: consider another simplified
+    // class instead of GComputation
+    using F = std::function<GComputation()>;
 
     std::string description;
     F pattern;
```
dmatveev: What is the reason to have this as a callback? Maybe store a cv::GComputation as-is instead?

modules/gapi/test/gapi_transform_tests.cpp
dmatveev: ?

dmatveev:
@rgarnov can you please check if your comments have been addressed?

modules/gapi/include/opencv2/gapi/gtransform.hpp
dmatveev: Will close after Alexey returns from a leave.

dmatveev:
Thanks!

modules/gapi/include/opencv2/gapi/gtransform.hpp
smirnov-alexey: Since this function doesn't operate with GArgs anymore there is no reason to have it as a callback. Can be easily changed if needed, just let me know.

Issue #11261, opened 2018-04-07
OpenCv 3.4.1 for Android: Bad JavaCamera2View performance 


ansanders:
#I recently upgraded to 3.4.1 and build it to support new `JavaCamera2View`. 
On my Nexus 7 (2013) I resolved a FPS decrease over time bug with `JavaCameraView`. 
`JavaCamera2View` now gives slow, but stable ~10 FPS . Not acceptable but maybe improvable. 
Performance on my Galaxy S8 differs much. ~30 FPS with `JavaCameraView`, without any processing. 
But only **2** FPS! with `JavaCamera2View`, which is not usable for any needs. 
In both test cases, `JavaCamera2View` has a very improved behaviour in resolution settings, scaling / zooming behaviour and image quality (removed some blurry and bad quality effects) and maybe that is some kind of the problem. 
As far as I can say the resolution, quality and autofocus behaviour is so good, that it drops the FPS to a minimum.  But that is just my "newbie" opinion. 
Buffer managment could also be a problem, but I can't imagine why the performance is better on an older Device than S8 (Nexus 7). 

really need to solve this issue because I'd love to use the new `JavaCamera2View` since the quality of camera is much better

ty in advance

##### System information (version)

- OpenCV => 3.4.1 with contrib
- Operating System / Platform => Ubuntu 16.04
- Compiler Settings => build_sdk.py with ndk-config.16 - downgraded android SDK 

#### Edit

the devices get really really hot after a few minutes of running a simple Java2CameraView

ansanders:
@floe let me know if u need more details about my setup or something to solve this nasty little bug

ansanders:
linked to [https://github.com/opencv/opencv/issues/11118](url) of **tipsmiller ** " replacing the Imgproc.COLOR_YUV2RGBA_NV21 value with Imgproc.COLOR_YUV420sp2BGR" worked for me too with the color behaviour bug

floe:
@ansanders thanks for the thorough analysis. I think the root cause is that `calcPreviewSize` is still basically the same code as for the original Camera1 API, however, that was much more limited in what preview sizes it supported (at best something up to 720p). Camera2 offers much wider range of preview sizes, up to the sensor resolution of the camera. I guess we need an extra parameter or somesuch to select a dedicated frame size...

tipsmiller:
[...]

PR #14981, opened 2019-07-04
android: JavaCamera2View use calculateCameraFrameSize() method

### LINKED ISSUE #14915: JavaCamera2View incorrect camera output caused by wrong preview size

- OpenCV => 4.1.0
- Operating System / Platform => Android API Level 28 (Patch: June 1st, 2019)
- Compiler => clang++ 7.0

##### Detailed description

Using `JavaCamera2View` will output a faulty camera image as shown below if the aspect ratio of the preview size, calculated by `calcPreviewSize`, does not match up with the actual screen aspect ratio.

<details>

device-2019-06-27-122401

</details>

For reference, the device I tested this on is a Xiaomi Mi A1 with the latest patches installed. It has a device screen size of 1920x1080. The calculated preview size, which is also being used in the image above, is 1440x1080. This results in an aspect ratio of 4:3 as opposed to the screen aspect ratio of 16:9. The highest preview size with the same aspect ratio as provided by the device is 1280x720. Here's a small snippet of the device's reported available preview sizes:

* 1440x1080 (automatically picked as "best")
* 1280x960
* 1280x720 (should be picked as "best")
* 1280x640
* 864x480
* ...

This behavior is due to `calcPreviewSize` assuming that the first output size provided by `getOutputSizes` [can be assumed as the "best" preview size from the start](https://github.com/opencv/opencv/blob/b95e93c20ab3df34c110385539952e5abcff3715/modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java#L264). Deleting the two lines before looping over all available preview sizes fixes this behavior. Just for reference, the image below is the same scene as above but with the two lines deleted.

<details>

device-2019-06-27-122256

</details>

I'm willing to submit a PR since this fix is just deleting two lines of code. However I'd like to see if someone is able to confirm this behavior.

##### Steps to reproduce

Haven't found a way to reproduce it yet because it seems to be an issue that heavily depends on the device (and I only have one device to test it on). However overriding the call to `calcPreviewSize` in `connectCamera` with dimensions smaller than the device's screen dimensions should evoke this behavior.

alalek:
from CameraBridgeViewBase (common base with JavaCameraView)

Both interfaces (JavaCameraView/JavaCamera2View) using the same method and the same approach.
If user want to tune calculation then it should override "View" class and override this method (`calculateCameraFrameSize()`)

resolves #14915

### DIFF
--- a/modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
+++ b/modules/java/generator/android-21/java/org/opencv/android/JavaCamera2View.java
@@ -2,6 +2,7 @@ package org.opencv.android;
 
 import java.nio.ByteBuffer;
 import java.util.Arrays;
+import java.util.List;
 
 import android.annotation.TargetApi;
 import android.content.Context;
@@ -24,6 +25,7 @@ import android.view.ViewGroup.LayoutParams;
 
 import org.opencv.core.CvType;
 import org.opencv.core.Mat;
+import org.opencv.core.Size;
 import org.opencv.imgproc.Imgproc;
 
 /**
@@ -248,6 +250,20 @@ public class JavaCamera2View extends CameraBridgeViewBase {
         }
     }
 
+    public static class JavaCameraSizeAccessor implements ListItemAccessor {
+        @Override
+        public int getWidth(Object obj) {
+            android.util.Size size = (android.util.Size)obj;
+            return size.getWidth();
+        }
+
+        @Override
+        public int getHeight(Object obj) {
+            android.util.Size size = (android.util.Size)obj;
+            return size.getHeight();
+        }
+    }
+
     boolean calcPreviewSize(final int width, final int height) {
         Log.i(LOGTAG, "calcPreviewSize: " + width + "x" + height);
         if (mCameraID == null) {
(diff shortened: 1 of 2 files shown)


DorJoogl:
Great effort. However it's missing the aspect ratio check from the old `JavaCamera2View`. Therefore [this line](https://github.com/opencv/opencv/blob/3998b41d6843217e639a3ba53389222ec0620351/modules/java/generator/android/java/org/opencv/android/CameraBridgeViewBase.java#L487) would have to look more like this.

```java
if (width >= calcWidth && height >= calcHeight && Math.abs(aspect - (float) width / height) <= 0.02) {
    calcWidth = (int) width;
    calcHeight = (int) height;
    Log.d(TAG, "passed");
}
```

`aspect` is defined as `float aspect = (float) maxAllowedWidth / maxAllowedHeight` prior to the outer loop. Tested it on my end and using `JavaCamera2View` now selects the correct preview size. It also doesn't break `JavaCameraView` for the sake of supporting the deprecated camera API.

This doesn't fix the broken previews on certain preview sizes I mentioned yet. If developers should override `calculateCameraFrameSize()` to adjust the preview size to their needs, then they might encounter the same issue with the faulty camera output.

I'm positive however that this change is very necessary and should be merged asap. As a first-time user of OpenCV4Android just a couple months ago I was a *bit* frustrated to see it not work out of the box. So I'm very glad this issue is getting resolved.

floe:
Can't test right now unfortunately, but looks good. Should this also be in 4.1.1?

alalek:
Purpose of this patch is to reuse existed code from both JavaCameraView/JavaCamera2View and provide the same selection for apps.

Next step is to refine frame selection itself:
- user callback instead of method overload
- aspect ratio (current approach is not reliable in some cases)
- try approach with "scoring" of each frame size and select the best one (will eliminate "select the first" workaround)

DorJoogl:
> aspect ratio (current approach is not reliable in some cases)

Just curious. What's unreliable with the current approach? I get the part about reusing old code but what's stopping us from adding the aspect ratio check to `CameraBridgeViewBase`? That'd somewhat fix the behavior in #14915 and in the recent comments of #11261.

alalek:
Example:
```
2019-07-08 20:33:16.083 15844-15844/org.opencv.samples.puzzle15 I/JavaCamera2View: calcPreviewSize: 1184x576
2019-07-08 20:33:23.070 15844-15844/org.opencv.samples.puzzle15 D/CameraBridge: trying size: 1600x1200  aspect: 1.3333333333333333  e_aspect:2.0555555555555554
2019-07-08 20:33:24.506 15844-15844/org.opencv.samples.puzzle15 D/CameraBridge: trying size: 1280x720  aspect: 1.7777777777777777  e_aspect:2.0555555555555554
2019-07-08 20:33:28.580 15844-15844/org.opencv.samples.puzzle15 D/CameraBridge: trying size: 960x720  aspect: 1.3333333333333333  e_aspect:2.0555555555555554
2019-07-08 20:33:30.887 15844-15844/org.opencv.samples.puzzle15 D/CameraBridge: trying size: 864x480  aspect: 1.8  e_aspect:2.0555555555555554
2019-07-08 20:33:35.285 15844-15844/org.opencv.samples.puzzle15 D/CameraBridge: trying size: 720x480  aspect: 1.5  e_aspect:2.0555555555555554
2019-07-08 20:33:35.290 15844-15844/org.opencv.samples.puzzle15 D/CameraBridge: trying size: 640x480  aspect: 1.3333333333333333  e_aspect:2.0555555555555554
2019-07-08 20:33:35.293 15844-15844/org.opencv.samples.puzzle15 D/CameraBridge: trying size: 480x320  aspect: 1.5  e_aspect:2.0555555555555554
2019-07-08 20:33:35.298 15844-15844/org.opencv.samples.puzzle15 D/CameraBridge: trying size: 320x240  aspect: 1.3333333333333333  e_aspect:2.0555555555555554
2019-07-08 20:33:35.303 15844-15844/org.opencv.samples.puzzle15 D/CameraBridge: trying size: 176x144  aspect: 1.2222222222222223  e_aspect:2.0555555555555554
2019-07-08 20:33:35.309 15844-15844/org.opencv.samples.puzzle15 I/JavaCamera2View: Selected preview size to 864x480
```

`0.02` aspect threshold is not reliable here.

BTW, device screen size is `1280x720`, but application area is much smaller (status bar, title bar, "home" button area)

PR #14994, opened 2019-07-08
WUI based implementation to initUndistortRectifyMap

terfendail:
### This pullrequest changes
WUI based implementation to initUndistortRectifyMap


terfendail:
Performance change for InitUndistortMap::Undistort

| Baseline | Reference time |  PR time | Speedup |
|------------------|----------------|----------|---------|
|SSE2|4.561|2.189|2.08|
|SSE3|4.560|2.195|2.08|
|SSE4_2|4.452|2.069|2.15|
|AVX2|0.930|0.860|1.08|

modules/imgproc/src/undistort.cpp
```diff
@@ -120,12 +135,129 @@ class initUndistortRectifyMapComputer : public cv::ParallelLoopBody
             else
                 CV_Assert(m1 != NULL);
 
-    #if CV_TRY_AVX2
+#if CV_TRY_AVX2
             if( useAVX2 )
                 j = cv::initUndistortRectifyMapLine_AVX(m1f, m2f, m1, m2,
```
alalek: What is about performance difference between w.u.i below and this "AVX" code branch?

terfendail: I suppose AVX2 branch should be executed for AVX2 baseline. Also I've tested performance against CPU_BASELINE=AVX2 and CPU_DISPATCH=AVX2 and got 15% performance improvement. So I think average performance gain is about 10% due to manual cycle unrolling and a few reused constant values.
I think it is reasonable to replace AVX2 branch with dynamic dispatching of w.u.i code. Does it make sense to make this change a part of the PR?

alalek: It make sense to avoid code duplication if we archiving similar performance.

PR #15010, opened 2019-07-10
support shape

vonchenplus:
hello @dkurt, I want to implement the eye operation of tensorflow, but I found that there are many basic operations that are not implemented, such as Shape, so I tried to submit this implementation, it is not feasible.

### DIFF
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -86,6 +86,15 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         static Ptr<Layer> create(const LayerParams &params);
     };
 
+    /**
+     * Shape layer 
+     */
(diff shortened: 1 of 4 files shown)


dkurt:
@vonchenplus, Hi!

Thanks for contribution!

`Shape` is not a deep learning layer. It mostly a supportive operation. Do you have a network with this op so OpenCV throws a error such "Shape layer is not implemented"?

vonchenplus:
@dkurt, yes, opencv throw a error such "Shape layer is not implemented". Such as the following code:
`X_content = tf.placeholder(tf.float32, shape=(1, None, None, 3))`
`content_t = tf.transpose(tf.squeeze(X_content, axis=[0]), (2, 0, 1))`
`Cc, Hc, Wc = tf.unstack(tf.shape(content_t))`
`eye = tf.eye(Cc,)`

vonchenplus:
hello @dkurt, any suggestions?

PR #15021, opened 2019-07-11
G-API: fix export specifier in standalone build

andrey-golubev:
### This pullrequest

fixes G-API symbol export specifier in case of G-API standalone mode

Originally, if someone decides to use G-API standalone and link G-API static library to a user's dynamic one: the symbols from our static lib are exported and placed into the symbol table of the user's dynamic library.

```
```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/own/exports.hpp
+++ b/modules/gapi/include/opencv2/gapi/own/exports.hpp
@@ -13,6 +13,19 @@
 #       define GAPI_EXPORTS CV_EXPORTS
 #   else
 #       define GAPI_EXPORTS
+
+#if 0  // Note: the following version currently is not needed for non-OpenCV build
+#       if defined _WIN32
+#           define GAPI_EXPORTS __declspec(dllexport)
+#       elif defined __GNUC__ && __GNUC__ >= 4
+#           define GAPI_EXPORTS __attribute__ ((visibility ("default")))
+#       endif
+
+#       ifndef GAPI_EXPORTS
+#           define GAPI_EXPORTS
+#       endif
+#endif
+
 #   endif
 
 #endif // OPENCV_GAPI_OWN_TYPES_HPP
--- a/modules/gapi/include/opencv2/gapi/own/exports.hpp
+++ b/modules/gapi/include/opencv2/gapi/own/exports.hpp
@@ -11,20 +11,8 @@
 #   if defined(__OPENCV_BUILD)
 #       include <opencv2/core/base.hpp>
 #       define GAPI_EXPORTS CV_EXPORTS
-
-#   elif !defined(GAPI_STANDALONE)
-#       if defined _WIN32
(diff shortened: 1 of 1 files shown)


alalek:
IMHO, `GAPI_STANDALONE` name should not be used to switch between static/shared builds.

Alternatives:
- allow to define `GAPI_EXPORTS` externally and wrap all current stuff under (`#ifndef GAPI_EXPORTS`. Add `-DGAPI_EXPORTS` for static builds
- introduce separate flag to control statis/shared (CMake usually adds `_EXPORTS` variant of this flag - see `DEFINE_SYMBOL CVAPI_EXPORTS` property)

andrey-golubev:
I believe `GAPI_STANDALONE` is more about building G-API independently of OpenCV (and as of now the static library is produced)

I see your point about `GAPI_EXPORTS` but it seems like this will be an additional compile definition just for the sake of the `GAPI_EXPORTS` which I doubt anyone (besides us) would update anyway.

Not sure I understand the version with `CVAPI_EXPORTS` that you propose, from the code it looks like that the difference is only in CMake's function to be used but the actual behavior/logic is the same, am I wrong?

alalek:
CMake's `DEFINE_SYMBOL` documentation is [here](https://cmake.org/cmake/help/v3.14/prop_tgt/DEFINE_SYMBOL.html).

andrey-golubev:
@rgarnov I simplified: looks like we don't need non-OpenCV part altogether, please look again

modules/gapi/include/opencv2/gapi/own/exports.hpp
```diff
@@ -11,18 +11,8 @@
...
-#           define GAPI_EXPORTS __declspec(dllexport)
-#       elif defined __GNUC__ && __GNUC__ >= 4
-#           define GAPI_EXPORTS __attribute__ ((visibility ("default")))
-#       endif
-
-#       ifndef GAPI_EXPORTS
-#           define GAPI_EXPORTS
-#       endif
-
+#       define GAPI_EXPORTS
```
dmatveev: Would IE still work with this? :)

dmatveev:
Meanwhile, if now we're making a change which is then will be 180*-reverted, y we do need to do it at all?

Maybe @andrey-golubev and @rgarnov go to IE folks and discuss moving IE's "fluid" to a separate .SO? :)

modules/gapi/include/opencv2/gapi/own/exports.hpp
andrey-golubev: yep, tested with it at first

andrey-golubev:
@dmatveev from my discussion with @rgarnov it seems like it's easier to do it this way right now and then see what's a proper way. good point about commenting out and not deleting, will do this

dmatveev:
@andrey-golubev ok!

andrey-golubev:
@dmatveev @alalek updated

andrey-golubev:
Thanks!

PR #15035, opened 2019-07-12
3.4

ChipKerchner:
This reduces the store gathering queue pressures.  The result is up to 20% faster histograms in the threshold algorithms.


ChipKerchner:
Need to remove old commits.  Closing and resubmitting.

PR #15040, opened 2019-07-13
Support headless WindowsCE: No getenv, UNICODE, wchar filesystem API

hugolm84:
### This pullrequest changes
This PR enables compilation for headless WINCE toolchains. WindowsCE only supports filesystem calls with wide characters, does not have getenv and operates on absolute filepaths only.
Compiling for WINCE should be done with the  `/p:CharacterSet=Unicode` flag.

Configuration that this PR supports:
```
-DBUILD_EXAMPLES=OFF -DBUILD_opencv_apps=OFF -DBUILD_opencv_calib3d=OFF -DBUILD_opencv_highgui=OFF -DBUILD_opencv_features2d=OFF -DBUILD_opencv_flann=OFF -DBUILD_opencv_ml=OFF -DBUILD_opencv_objdetect=OFF -DBUILD_opencv_photo=OFF -DBUILD_opencv_shape=OFF -DBUILD_opencv_stitching=OFF -DBUILD_opencv_superres=OFF -DBUILD_opencv_ts=OFF -DBUILD_opencv_video=OFF -DBUILD_opencv_videoio=OFF -DBUILD_opencv_videostab=OFF -DBUILD_opencv_dnn=OFF -DBUILD_opencv_java=OFF -DBUILD_opencv_python2=OFF -DBUILD_opencv_python3=OFF -DBUILD_opencv_java_bindings_generator=OFF -DBUILD_opencv_python_bindings_generator=OFF -DWITH_QT=OFF -DWITH_GTK=OFF -DWITH_QUIRC=OFF -DWITH_JASPER=OFF -DWITH_WEBP=OFF -DWITH_PROTOBUF=OFF -DBUILD_SHARED_LIBS=OFF -DWITH_OPENEXR=OFF -DCV_TRACE=OFF
```

### DIFF
--- a/modules/core/src/ocl.cpp
+++ b/modules/core/src/ocl.cpp
@@ -876,7 +876,12 @@ bool haveOpenCL()
     if (!g_isOpenCLInitialized)
     {
         CV_TRACE_REGION("Init_OpenCL_Runtime");
-        const char* envPath = getenv("OPENCV_OPENCL_RUNTIME");
+        const char* envPath =
+#ifdef NO_GETENV
+            NULL;
+#else
+            getenv("OPENCV_OPENCL_RUNTIME");
+#endif
         if (envPath)
         {
             if (cv::String(envPath) == "disabled")
@@ -1737,8 +1742,12 @@ static cl_device_id selectOpenCLDevice()
 {
     std::string platform, deviceName;
     std::vector<std::string> deviceTypes;
-
-    const char* configuration = getenv("OPENCV_OPENCL_DEVICE");
+    const char* configuration =
+#ifdef NO_GETENV
+        NULL;
+#else
+        getenv("OPENCV_OPENCL_DEVICE");
+#endif
     if (configuration &&
             (strcmp(configuration, "disabled") == 0 ||
              !parseOpenCLDeviceConfiguration(std::string(configuration), platform, deviceTypes, deviceName)
--- a/modules/core/src/system.cpp
+++ b/modules/core/src/system.cpp
@@ -914,15 +914,16 @@ String tempfile( const char* suffix )
     fname = String(aname);
     RoUninitialize();
 #elif _WIN32_WCE
-    LPWSTR temp_dir2[MAX_PATH+1] = {0};
-    LPWSTR temp_file[MAX_PATH+1] = {0};
+    const auto kMaxPathSize = MAX_PATH+1;
+    wchar_t temp_dir[kMaxPathSize] = {0};
+    wchar_t temp_file[kMaxPathSize] = {0};
 
-    ::GetTempPath(sizeof(temp_dir2), *temp_dir2);
+    ::GetTempPathW(kMaxPathSize, temp_dir);
 
-    if(0 != ::GetTempFileName(*temp_dir2, L"ocv", 0, *temp_file)) {
-        DeleteFile(*temp_file);
+    if(0 != ::GetTempFileNameW(temp_dir, L"ocv", 0, temp_file)) {
+        DeleteFileW(temp_file);
         char aname[MAX_PATH];
-        size_t copied = wcstombs(aname, *temp_file, MAX_PATH);
+        size_t copied = wcstombs(aname, temp_file, MAX_PATH);
         CV_Assert((copied != MAX_PATH) && (copied != (size_t)-1));
         fname = String(aname);
     }
--- a/modules/core/src/system.cpp
+++ b/modules/core/src/system.cpp
@@ -378,7 +378,7 @@ struct HWFeatures
 
     void initialize(void)
     {
-#ifndef WINRT
+#ifndef NO_GETENV
         if (getenv("OPENCV_DUMP_CONFIG"))
         {
             fprintf(stderr, "\nOpenCV build configuration is:\n%s\n",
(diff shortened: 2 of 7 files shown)


modules/core/src/utils/filesystem.cpp
```diff
@@ -88,10 +88,13 @@ cv::String join(const cv::String& base, const cv::String& path)
...
-#ifdef _WIN32
-    const char* result_str = _fullpath(NULL, path.c_str(), 0);
+    const char* result_str =
+#ifdef NO_GETENV
+        NULL;
+#elif defined(_WIN32)
+        _fullpath(NULL, path.c_str(), 0);
 #else
-    const char* result_str = realpath(path.c_str(), NULL);
+        realpath(path.c_str(), NULL);
```
alalek: This code is guarded by `OPENCV_HAVE_FILESYSTEM_SUPPORT`.
Perhaps it make sense to disable this support instead (see .hpp file).

> #ifdef NO_GETENV
> ...
> _fullpath

There is inconsistency with semantic.

modules/core/src/system.cpp
```diff
@@ -913,6 +913,19 @@ String tempfile( const char* suffix )
     CV_Assert((copied != MAX_PATH) && (copied != (size_t)-1));
     fname = String(aname);
     RoUninitialize();
+#elif _WIN32_WCE
+    LPWSTR temp_dir2[MAX_PATH+1] = {0};
+    LPWSTR temp_file[MAX_PATH+1] = {0};
```
alalek: Did you test that code?

`LPWSTR` is a pointer.

modules/core/src/system.cpp
```diff
@@ -1691,7 +1704,14 @@ class ThreadID {
         id(CV_XADD(&g_threadNum, 1))
     {
 #ifdef OPENCV_WITH_ITT
-        __itt_thread_set_name(cv::format("OpenCVThread-%03d", id).c_str());
+        std::string thread_name = cv::format("OpenCVThread-%03d", id);
+#if defined _WIN32 && defined _UNICODE
+        std::wstring ws_thread_name;
+        ws_thread_name.assign(thread_name.begin(), thread_name.end());
+        __itt_thread_set_name(ws_thread_name.c_str());
```
alalek: According to PR's description this code should be disabled:

> -DCV_TRACE=OFF

modules/core/src/ocl.cpp
```diff
@@ -876,7 +876,12 @@ bool haveOpenCL()
     if (!g_isOpenCLInitialized)
     {
         CV_TRACE_REGION("Init_OpenCL_Runtime");
-        const char* envPath = getenv("OPENCV_OPENCL_RUNTIME");
+        const char* envPath =
+#ifdef NO_GETENV
```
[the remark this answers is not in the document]
hugolm84: OpenCL is supported by for example the GuruCE iMX6 BSP (https://guruce.com/imx6). However `getenv` still don't exist, so I believe its more proper to ifdef for `getenv` as the define already exist for this purpose.
If you still feel its better to disable OpenCL in the configuration step, I can do that instead.

modules/core/src/system.cpp
hugolm84: You are correct, I have corrected the code.

modules/core/src/system.cpp
hugolm84: You are correct, I have removed this change

modules/core/src/utils/filesystem.cpp
hugolm84: Hm, Windows CE operates on absolute paths only. So converting a path to a canonical absolute path in this case is a noop.  That doesn't mean that there is no filesystem support. In this case, the path to be converted should be returned as is, just as the function does on L104. 

There is an inconsistency with the `OPENCV_HAVE_FILESYSTEM_SUPPORT`. It seems to disable filesystem functionality for WINRT by default, but still have #ifdefs in the code paths. Perhaps `OPENCV_HAVE_FILESYSTEM_SUPPORT` means that all functions have not been fully ported yet? In that case, I can revert this change and perhaps send in new PR's that address that separately.

hugolm84:
> It make sense to move part of configuration into external build scripts like [WinRT](https://github.com/opencv/opencv/tree/4.1.0/platforms/winrt) with minimal description of build prerequisites.
> 
> [JavaScript](https://github.com/opencv/opencv/tree/4.1.0/platforms/js) is another example.



> It make sense to move part of configuration into external build scripts like [WinRT](https://github.com/opencv/opencv/tree/4.1.0/platforms/winrt) with minimal description of build prerequisites.
> 
> [JavaScript](https://github.com/opencv/opencv/tree/4.1.0/platforms/js) is another example.

Ok, Perhaps 

> It make sense to move part of configuration into external build scripts like [WinRT](https://github.com/opencv/opencv/tree/4.1.0/platforms/winrt) with minimal description of build prerequisites.
> 
> [JavaScript](https://github.com/opencv/opencv/tree/4.1.0/platforms/js) is another example.

Hi @alalek Thank you for your comments! In the light of those, I have instead disabled OPEN_CL and OPENCV_HAVE_FILESYSTEM_SUPPORT. I will close this PR and open an new one with the more relevant changes.

PR #15048, opened 2019-07-15
Reduce store gathering pressures - speeds thresholds by up to 20%

ChipKerchner:
Reduce store gathering pressures - speeds thresholds by up to 20%.

### DIFF
--- a/modules/imgproc/src/thresh.cpp
+++ b/modules/imgproc/src/thresh.cpp
@@ -1160,7 +1160,7 @@ getThreshVal_Otsu_8u( const Mat& _src )
     const int N = 256;
     int i, j, h[N] = {0};
     #if CV_ENABLE_UNROLLED
-    int h2[3][N] = {0};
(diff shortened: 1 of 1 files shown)


modules/imgproc/src/thresh.cpp
```diff
@@ -1159,6 +1159,9 @@ getThreshVal_Otsu_8u( const Mat& _src )
 
     const int N = 256;
     int i, j, h[N] = {0};
+    #if CV_ENABLE_UNROLLED
+    int h2[3][N] = {0};
```
alalek: Could you please change name for this to `h_unrolled`?

```
int h_unrolled[3][N] = {};
```

(`= {}` initialization should make MacOSX builder happy)

ChipKerchner:
How or when does code like this get put into master?

alalek:
We have regular merges from 3.4 into master (weekly/bi-weekly).

Issue #14977, opened 2019-07-04
core: Persistence base64 encoding seems broken


alalek:
Test `Core_InputOutput.filestorage_base64_basic` doesn't process base64 data anymore.

broken by #13011

ganesh-k13:
Hey @alalek , is there something here I can take up?

asmorkalov:
https://github.com/opencv/opencv/pull/16542

asmorkalov:
https://github.com/opencv/opencv/pull/16543

asmorkalov:
Current status:
- BASE64 decoding is there for all formats.
- BASE64 encoding is not implemented in master at all. The flag is suppressed by implementation and `cv::FileStorage` always write file with list of values.

PR #15050, opened 2019-07-15
core(persistence): struct storage layout without alignment gaps

alalek:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/636

relates #14977

Align with OpenCV 4.x rewritten implementation.
Rework tests.

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/core/include/opencv2/core/persistence.hpp
+++ b/modules/core/include/opencv2/core/persistence.hpp
@@ -597,8 +597,8 @@ public:
     Usually it is more convenient to use operator `>>` instead of this method.
     @param fmt Specification of each array element. See @ref format_spec "format specification"
     @param vec Pointer to the destination array.
-    @param len Number of elements to read. If it is greater than number of remaining elements then all
-    of them will be read.
+    @param len Number of bytes to read (buffer size limit). If it is greater than number of
+               remaining elements then all of them will be read.
      */
     void readRaw( const String& fmt, uchar* vec, size_t len ) const;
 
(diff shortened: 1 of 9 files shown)


modules/core/src/persistence_base64.cpp
```diff
@@ -803,9 +811,28 @@ class BinaryToCvSeqConvertor
...
+                    static bool skip_message = cv::utils::getConfigurationParameterBool("OPENCV_PERSISTENCE_SKIP_PACKED_STRUCT_WARNING",
+#ifdef _DEBUG
+                            false
+#else
+                            true
+#endif
+                    );
+                    if (!skip_message)
+                    {
+                        CV_LOG_WARNING(NULL, "Binary converter: struct storage layout has been changed in OpenCV 3.4.7. Alignment gaps has been removed from the storage containers. "
```
mshabunin: Can we use `CV_LOG_DEBUG` and avoid introduction of new environment configuration parameter?

modules/core/include/opencv2/core/persistence.hpp
```diff
@@ -597,8 +597,8 @@ class CV_EXPORTS_W_SIMPLE FileNode
     Usually it is more convenient to use operator `>>` instead of this method.
     @param fmt Specification of each array element. See @ref format_spec "format specification"
     @param vec Pointer to the destination array.
-    @param len Number of elements to read. If it is greater than number of remaining elements then all
-    of them will be read.
+    @param len Number of bytes to read (buffer size limit). If it is greater than number of
```
mshabunin:
```suggestion
    @param len Number of bytes to read (buffer size limit). If it is greater than size of
```

modules/core/include/opencv2/core/persistence.hpp
```diff
@@ -668,11 +668,12 @@ class CV_EXPORTS FileNodeIterator
     Usually it is more convenient to use operator `>>` instead of this method.
     @param fmt Specification of each array element. See @ref format_spec "format specification"
     @param vec Pointer to the destination array.
-    @param maxCount Number of elements to read. If it is greater than number of remaining elements then
-    all of them will be read.
+    @param len Number of bytes to read (buffer size limit). If it is greater than number of
```
mshabunin:
```suggestion
    @param len Number of bytes to read (buffer size limit). If it is greater than size of
```

modules/core/src/persistence_base64.cpp
alalek: `DEBUG` messages are not compiled into release builds by default

Issue #15016, opened 2019-07-11
SegFault on cv::dnn::Net::dumpToFile call


JulienPlante:
##### System information (version)
- OpenCV => 4.1.0
- Operating System / Platform => `Linux raspberrypi 4.19.42-v7+ #1219 SMP Tue May 14 21:20:58 BST 2019 armv7l GNU/Linux`
- Compiler => `g++ (Raspbian 6.3.0-18+rpi1+deb9u1) 6.3.0 20170516`

##### Detailed description

Calling cv::dnn::Net::dumpToFile on a loaded Darknet network gives SegFault.

##### Steps to reproduce

```cpp
#include <opencv2/opencv.hpp>
#include <iostream>

using namespace cv;
using namespace dnn;
using namespace std;

int main(int argc, char* argv[])
{
    String configPath = "network.cfg";
    String modelPath = "network.weights";
    Net net = readNetFromDarknet(configPath, modelPath);
    cout << "Successfully loaded net, dumping to net.dot" << endl;
    net.dumpToFile("net.dot"); // SegFault
    return 0;
}
```

##### Edit : stack trace
```
$ gdb ./test_issue
GNU gdb (Raspbian 7.12-6) 7.12.0.20161007-git
...
(gdb) r
Starting program: /home/pi/Documents/yolo_objdetect/build/test_issue 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/arm-linux-gnueabihf/libthread_db.so.1".
Successfully loaded net, dumping to net.dot

Program received signal SIGSEGV, Segmentation fault.
0x75ee1a9c in cv::dnn::dnn4_v20190430::Net::dump[abi:cxx11]() () from /usr/local/lib/libopencv_dnn.so.4.1
(gdb) bt
#0  0x75ee1a9c in cv::dnn::dnn4_v20190430::Net::dump[abi:cxx11]() () from /usr/local/lib/libopencv_dnn.so.4.1
#1  0x75ee313c in cv::dnn::dnn4_v20190430::Net::dumpToFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /usr/local/lib/libopencv_dnn.so.4.1
#2  0x000117c8 in main (argc=1, argv=0x7efff034) at /home/pi/Documents/yolo_objdetect/src/test.cpp:14
(gdb) 

```

alalek:
> SegFault

or exception? (I don't see try-catch in your code)

[...]

PR #15057, opened 2019-07-16
Fix dumpToFile

l-bat:
Related: #15016
```

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/test/test_misc.cpp
+++ b/modules/dnn/test/test_misc.cpp
@@ -78,17 +78,25 @@ TEST(readNet, Regression)
     EXPECT_FALSE(net.empty());
 }
 
-TEST(dump, Regression)
+typedef testing::TestWithParam<tuple<Backend, Target> > dump;
+TEST_P(dump, Regression)
 {
+    const int backend  = get<0>(GetParam());
+    const int target   = get<1>(GetParam());
     Net net = readNet(findDataFile("dnn/squeezenet_v1.1.prototxt"),
-                      findDataFile("dnn/squeezenet_v1.1.caffemodel"));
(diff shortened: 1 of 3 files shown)


modules/dnn/test/test_misc.cpp
```diff
@@ -78,6 +78,18 @@ TEST(readNet, Regression)
     EXPECT_FALSE(net.empty());
 }
 
+TEST(dump, Regression)
+{
+    Net net = readNet(findDataFile("dnn/squeezenet_v1.1.prototxt"),
+                      findDataFile("dnn/squeezenet_v1.1.caffemodel"));
```
alalek: `.caffemodel`files are not in opencv_extra. So this test data is considered optional.
See #15052

modules/dnn/test/test_misc.cpp
```diff
@@ -78,6 +78,18 @@ TEST(readNet, Regression)
...
+
+    Mat input = blobFromImage(imread(findDataFile("dnn/googlenet_0.png")), 1.0f,
+                              Size(227,227), Scalar(), false, true);
+    net.setInput(input);
+    net.dump();
```
alalek: Test without check looks strange.

Lets check that result is not empty at least.

modules/dnn/test/test_misc.cpp
alalek: `dnn/squeezenet_v1.1.prototxt` is in opencv_extra. `false` is not needed there.
(issue is catched by `OPENCV_TEST_CHECK_OPTIONAL_DATA=1`)

PR #15077, opened 2019-07-18
calib3d: fix format in findChessboardCornersSB doxygen

lpea:
This pull request fixes a small formatting issue in the doxygen for cv::findChessboardCornersSB().


mshabunin:
@lpea , our CI can not test pull requests sent from _master_ and _3.4_ branches. You should open a new using different branch name with _3.4_ target branch.

See https://github.com/opencv/opencv/wiki/Branches#common-rules-for-all-branches

lpea:
@mshabunin Sorry about that, I'll change the branch name. However the function does not exist in 3.4 so this fix does not make any sense on the 3.4 branch. That's why the PR is based on master.

PR #15081, opened 2019-07-18
Generation module

dvd42:
**Merge with extra**: TBD

Adds a Generation module to the new high level DL API currently in [l-bat's fork](https://github.com/opencv/opencv/pull/14780)

```
opencv_extra=high_level_api
```

PR #15082, opened 2019-07-18
Segmentation module

dvd42:
Adds a Segmentation module to the new high level DL API currently in [l-bat's fork](https://github.com/opencv/opencv/pull/14780)

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/655

```
opencv_extra=segmentation-module_data

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*
```


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1104,6 +1104,36 @@ CV__DNN_INLINE_NS_BEGIN
...
+         /**
+          * @brief Create model from deep learning network.
+          * @param[in] network Net object.
+          */
+         CV_WRAP SegmentationModel(const Net& network);
+
+         /** @brief Given the @p input frame, create input blob, run net and returns the class prediction for each pixel
+          *  @param[in]  frame  The input image.
+          */
+         CV_WRAP Mat forward(InputArray frame);
```
dkurt: Please rename it to `segment`. Also use OutputArray for output: 

```cpp
CV_WRAP forward(InputArray frame, OutputArray mask);
```

modules/dnn/src/model.cpp
```diff
@@ -127,6 +127,49 @@ void ClassificationModel::classify(InputArray frame, int& classId, float& conf)
...
+
+SegmentationModel::SegmentationModel(const Net& network) : Model(network) {};
+
+Mat SegmentationModel::forward(InputArray frame){
+
+    std::vector<Mat> outs;
+    impl->predict(*this, frame.getMat(), outs);
+    CV_Assert(outs.size() == 1);
+    Mat score = outs[0];
+    const int batch = score.size[0];
```
dkurt: batch size should be `1` for a single frame, isn't it?

modules/dnn/test/test_model.cpp
```diff
@@ -202,6 +222,21 @@ TEST_P(Test_Model, DetectionMobilenetSSD)
                     scoreDiff, iouDiff, confThreshold, nmsThreshold, size, mean, scale);
 }
 
+TEST_P(Test_Model, Segmentation)
+{
+    std::string inp = _tf("dog416.png");
+    Net network = readNetFromONNX(_tf("segmentation.onnx"));
+    std::string exp = _tf("segmentation_exp.npy");
```
dkurt: Which model is used for test?

modules/dnn/src/model.cpp
dvd42: Yes, I was just trying to make it generic in case the input has batch_size 2. Should I ignore this case?

modules/dnn/test/test_model.cpp
dvd42: Its just a dummy, small model that I created.

dkurt: So why not to test existing model? Check ENet from Torch test.

modules/dnn/src/model.cpp
dkurt: This method receives a single frame so batch_size == 1.

modules/dnn/test/test_model.cpp
dvd42: @dkurt I get 404 error when downloading that model using the script in the opencv extra

dvd42: @dkurt  I changed the test data to use an fcn8 hope its ok now :)

dkurt:
@dvd42, please rebase this branch with current master.

dvd42:
@dkurt, done. I am guessing I should do the same for #15081 and #15189, right?

dkurt:
@dvd42, Yes. Please avoid "merge" commits - squash commits into one.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1104,6 +1104,37 @@ CV__DNN_INLINE_NS_BEGIN
...
+          * @param[in] network Net object.
+          */
+         CV_WRAP SegmentationModel(const Net& network);
+
+         /** @brief Given the @p input frame, create input blob, run net
+          *  @param[in]  frame  The input image.
+          *  @param[out] mask Allocated class prediction for each pixel
+          */
+         CV_WRAP void segment(InputArray frame, OutputArray mask);
+
```
dkurt: Please remove empty line

modules/dnn/src/model.cpp
```diff
@@ -127,6 +127,46 @@ void ClassificationModel::classify(InputArray frame, int& classId, float& conf)
     std::tie(classId, conf) = classify(frame);
 }
 
+SegmentationModel::SegmentationModel(const String& model, const String& config)
+    : Model(model, config) {};
+
+SegmentationModel::SegmentationModel(const Net& network) : Model(network) {};
+
+void SegmentationModel::segment(InputArray frame, OutputArray mask){
```
dkurt: Please move bracket
```
void SegmentationModel::segment(InputArray frame, OutputArray mask)
{
```

modules/dnn/src/model.cpp
```diff
@@ -127,6 +127,46 @@ void ClassificationModel::classify(InputArray frame, int& classId, float& conf)
...
+    for (int ch = 1; ch < chns; ch++)
+    {
+        for (int row = 0; row < rows; row++)
+        {
+            const float *ptrScore = score.ptr<float>(0, ch, row);
+            int *ptrMaxCl = maxCl.ptr<int>(row);
+            float *ptrMaxVal = maxVal.ptr<float>(row);
+            for (int col = 0; col < cols; col++)
+            {
+                if (ptrScore[col] >= ptrMaxVal[col])
```
dkurt: `>`

modules/dnn/src/model.cpp
```diff
@@ -127,6 +127,46 @@ void ClassificationModel::classify(InputArray frame, int& classId, float& conf)
...
+    impl->predict(*this, frame.getMat(), outs);
+    CV_Assert(outs.size() == 1);
+    Mat score = outs[0];
+
+    const int chns = score.size[1];
+    const int rows = score.size[2];
+    const int cols = score.size[3];
+
+    mask.create(rows, cols, CV_32S);
+    Mat maxCl = Mat::zeros(rows, cols, CV_32S);
```
dkurt: Why not just write into the `mask`?

modules/dnn/test/test_model.cpp
```diff
@@ -69,6 +69,26 @@ class Test_Model : public DNNTestLayer
         EXPECT_EQ(prediction.first, ref.first);
         ASSERT_NEAR(prediction.second, ref.second, norm);
     }
+
+    void testSegmentationModel(const Net& network, const std::string& inImgPath,
+                             const std::string& outImgPath, float norm,
+                             const Size& size = {-1, -1}, Scalar mean = Scalar(),
+                             double scale = 1.0, bool swapRB = false, bool crop = false)
```
dkurt: Please fix indentation

dvd42:
@dkurt, done

modules/dnn/test/test_model.cpp
```diff
@@ -7,68 +7,88 @@
 #include "npy_blob.hpp"
 namespace opencv_test { namespace {
 
-template<typename TString>
```
dkurt: Please avoid massive changes (looks like simple indentation)

dvd42: @dkurt, I am terribly sorry, did not notice that mess. I must have autoindented it by accident. Anyway, it is fixed now

dkurt:
@dvd42, Can you please squash all the commits into a single one?

dvd42:
@dkurt, done

modules/dnn/test/test_model.cpp
```diff
@@ -69,6 +69,26 @@ class Test_Model : public DNNTestLayer
...
+        Mat exp = blobFromNPY(outImgPath);
+
+        SegmentationModel model(network);
+        model.setInputSize(size).setInputMean(mean).setInputScale(scale)
+             .setInputSwapRB(swapRB).setInputCrop(crop);
+
+        model.segment(frame, mask);
+        exp.convertTo(exp, CV_8U);
+        normAssert(mask, exp, "", norm, norm);
+    }
```
dkurt: Looks like test is missed.

modules/dnn/src/model.cpp
```diff
@@ -127,6 +127,47 @@ void ClassificationModel::classify(InputArray frame, int& classId, float& conf)
...
+    CV_Assert(outs.size() == 1);
+    Mat score = outs[0];
+
+    const int chns = score.size[1];
+    const int rows = score.size[2];
+    const int cols = score.size[3];
+
+    mask.create(rows, cols, CV_8U);
+    Mat tmp_mask = mask.getMat();
+    tmp_mask = 0;
```
dkurt: Please use `setTo` and rename it to something like `classIds` because `tmp_mask` is not a good name for actual output.

dvd42: @dkurt, Done

modules/dnn/test/test_model.cpp
dkurt: What is the source of `testdata/dnn/segmentation_exp.npy`? For this model there is already reference data.

dvd42: Is the expected output of the model for that image, when ran in caffe. I generated it

dkurt: It's 1MB! :) You can just test the same image from [TEST(Reproducibility_FCN, Accuracy)](https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_caffe_importer.cpp).

dvd42: @dkurt The reference data in that example is just the output of the network, not the argmax over the channel dimension of the output. I am aware that I can just compute that after loading it, but what kind of test would that be. Testing the argmax code inside the module, against the same argmax code outside of it.

dkurt: For 500x500 CV_8UC1 mask? :confused: It must be even less using .png compression

dvd42: I can store it in a png file instead of an npy. Would that be OK?

dkurt: @dvd42, Even in NPY it should take about 250KB.

dvd42: @dkurt , done. Its 2.3kb now :)

dkurt: Just for experiment, can you try to run this model on lower input resolution? Will it produce maybe inaccurate but valid results?

dvd42: @dkurt, hey yeah I just tried it and the segmentation looks fine at half the resolution (256, 256)

modules/dnn/test/test_model.cpp
```diff
@@ -202,6 +221,21 @@ TEST_P(Test_Model, DetectionMobilenetSSD)
                     scoreDiff, iouDiff, confThreshold, nmsThreshold, size, mean, scale);
 }
 
+TEST_P(Test_Model, Segmentation)
+{
+    std::string inp = _tf("boat.jpg");
+    Net network = readNetFromCaffe(_tf("fcn8s-heavy-pascal.prototxt"), _tf("fcn8s-heavy-pascal.caffemodel"));
+    std::string exp = _tf("segmentation_exp.png");
```
dkurt: Please rewrite test so it'll initialize Model from weights and prototxt inside `testSegmentationModel`. Check existing tests for reference.

dvd42: @dkurt, done

modules/dnn/test/test_model.cpp
```diff
@@ -202,6 +221,22 @@ TEST_P(Test_Model, DetectionMobilenetSSD)
...
+{
+    std::string inp = _tf("dog416.png");
+    std::string weights_file = _tf("fcn8s-heavy-pascal.prototxt");
+    std::string config_file = _tf("fcn8s-heavy-pascal.caffemodel");
+    std::string exp = _tf("segmentation_exp.png");
+
+    Size size{512, 512};
+    float norm = 0;
+    double scale = 1.0;
+    Scalar mean = Scalar(127.5, 127.5, 127.5);
```
dkurt: Is this correct mean values? According to [models.yml](https://github.com/opencv/opencv/blob/master/samples/dnn/models.yml), there is no mean subtraction for FCN8s network and no Red<->Blue channels swap.

dvd42: I guess, I can change it, but does it really matter for test?

dkurt: @dvd42, It's very important. Tests as samples are used by developers as reference code. If we'll provide wrong usage in it - we'll mislead all the users.

dkurt: So the question is how you generated test data? You just run it once and collected output? Then how you can proof that the changes are correct?

Please use a sample to generate reference data for specific input size and input image: https://github.com/opencv/opencv/blob/master/samples/dnn/segmentation.py

dvd42: @dkurt Every time I make a change I generate new test data

dkurt: @dvd42, By sample?

dvd42: @dkurt What do you mean by sample?

dkurt: Run the following sample to get reference output data: https://github.com/opencv/opencv/blob/master/samples/dnn/segmentation.py:

```
python segmentation.py fcn8s --input /path/to/image.png --width xxx --height xxx
```

dvd42: @dkurt I just ran it and the output is **exactly** the same as my reference data. I dont see the problem.

dkurt: Please reduce input resolution as well. Model is pretty heavy. I think 128x128 would be enough (just check the segmentation mask).

dvd42: @dkurt, done

modules/dnn/test/test_model.cpp
dkurt: Thanks!

dkurt:
@alalek, It looks like buildbot cannot test this PR with different name of opencv_extra branch despite "Merge with extra". May I ask you to take a look? Thanks!

PR #15086, opened 2019-07-18
Bayer VNG Demosaicing Fix #2

### LINKED ISSUE #15011: Bayer VNG artifacts for specific edge orientations

- OpenCV => 3.2.0
- Operating System / Platform => Windows 10 64-bit / Python
- Compiler => MSC v.1900 64 bit (AMD64)

##### Detailed description

Using the VNG demosaicing in cvtColor( ) results in edge artifacts for edges running at orientations of 315 degrees (roughly NW -> SE).  You can see these artifacts in the image below created with the OpenCV VNG algorithm (top row), but not in an image created using a reference implementation of the VNG algorithm (in attached code vng.py) (bottom row).  Some color artifacts along edges are expected due to the nature of the interpolation, but the artifacts shown here on edges running from upper left to lower right are more severe than expected, and worse than the reference output.  Displaying a single channel (such as just the red channel shown on the right) makes it easier to see that pixels far from the edge are being modified in the OpenCV version.

edges315

For edges running along different orientations, such as 45 degrees (roughly SW -> NE), the OpenCV VNG implementation does not show artifacts, and appears to match the reference VNG implementation, as shown below.

edges45

Since the VNG algorithm computes gradients in the eight compass directions and there are separate blocks of code for processing each gradient direction, is it possible that there is a pointer error in the code for one of the gradient directions ?

##### Steps to reproduce

The examples above were generated by calling OpenCV from Python inside a Jupyter Notebook.
The input and output images for the examples above are attached, and can be duplicated by running the attached notebook "VNG_bug.ipynb"

[VNG_bug.zip](https://github.com/opencv/opencv/files/3378154/VNG_bug.zip)

dcouwenh:
Fixed calculation of Bs for non-green pixels.

resolves #15011

### This pullrequest changes

In file modules\imgproc\src\demosaicing.cpp, the Bs values for non-green cells were being calculated incorrectly in the NW and SE gradients. I fixed the math in both the SSE2 accelerated and non-accelerated code blocks.

### DIFF
--- a/modules/imgproc/perf/perf_cvt_color.cpp
+++ b/modules/imgproc/perf/perf_cvt_color.cpp
@@ -417,7 +417,7 @@ PERF_TEST_P(Size_CvtMode_Bayer, cvtColorBayer8u,
 
     TEST_CYCLE() cvtColor(src, dst, mode, ch.dcn);
 
-    SANITY_CHECK(dst, 1);
+    SANITY_CHECK_NOTHING();
 }
 
 typedef tuple<Size, CvtMode2> Size_CvtMode2_t;
--- a/modules/imgproc/src/demosaicing.cpp
+++ b/modules/imgproc/src/demosaicing.cpp
@@ -1153,14 +1153,14 @@ static void Bayer2RGB_VNG_8u( const Mat& srcmat, Mat& dstmat, int code )
                     {
                         Rs += srow[-bstep*2-2] + srow[0];
                         Gs += brow0[N6-1];
-                        Bs += srow[-bstep+1]*2;
+                        Bs += srow[-bstep-1]*2;
                         ng++;
                     }
                     if( gradSE < T )
                     {
                         Rs += srow[bstep*2+2] + srow[0];
                         Gs += brow2[N6+1];
-                        Bs += srow[-bstep+1]*2;
+                        Bs += srow[bstep+1]*2;
                         ng++;
                     }
                     R = srow[0];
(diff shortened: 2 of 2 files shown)


mshabunin:
Looks like there are failing tests:
```
Imgproc_ColorBayerVNG.regression : 
/build/precommit_linux64/3.4/opencv/modules/imgproc/test/test_color.cpp:1889: Failure
Expected equality of these values:
  0
  countNonZero(diff.reshape(1) > 1)
    Which is: 139600
Imgproc_ColorBayerVNG_Strict.regression : 
/build/precommit_linux64/3.4/opencv/modules/ts/src/ts.cpp:589: Failure
Failed

	failure reason: Bad accuracy
	test case #499
	seed: 6b745577e9e8e727
-----------------------------------
	SUM: 
Count non zero in absdiff: 41039

-----------------------------------
	CONSOLE: ............................................................
-----------------------------------
```
It seems that reference images should be regenerated:
* https://github.com/opencv/opencv_extra/tree/3.4/testdata/cv/cvtcolor - bayerVNG_gold.png
* https://github.com/opencv/opencv_extra/tree/3.4/testdata/cv/cvtcolor_strict - {bg,gb,rg,gr}lena.png
It looks like it is enough to remove these images and run tests: `opencv_test_imgproc --gtest_filter=*BayerVNG*`

Please note that you should create separate PR to the opencv/opencv_extra repository using same branch name as in this PR (_dcouwenh-vng-demosaicing-fix2_), then both PRs will be tested together.

dcouwenh:
@mshabunin , I have updated the test files in the opencv/opencv_extra repository to reflect the change to the bayer VNG demosaicing code in opencv/opencv, and created a pull request for the opencv_extra repository with the same name (dcouwenh-vng-demosaicing-fix2).

mshabunin:
@dcouwenh , looks like these files were used in another test: _opencv_perf_imgproc_. And now it is failing because it expects different values.

You need to replace this line: https://github.com/opencv/opencv/blob/ad092bf1ce369cd5d7311cfeebb9fb3ff4d0640b/modules/imgproc/perf/perf_cvt_color.cpp#L420
With `SANITY_CHECK_NOTHING();`

PR #15101, opened 2019-07-19
cmake: update initialization

alalek:
<cut/>

```
opencv_contrib=cmake_tesseract_cxx11_check
```


cmake/system/OpenCV_WinRT.cmake
```diff
@@ -0,0 +1,24 @@
+set(WINRT TRUE)
+
+add_definitions(-DWINRT -DNO_GETENV)
+
+# Making definitions available to other configurations and
+# to filter dependency restrictions at compile time.
+if(CMAKE_SYSTEM_NAME MATCHES WindowsPhone)
```
hugolm84: In light of previous discussion (https://github.com/opencv/opencv/pull/15059#discussion_r305446020), this should be changed to use the WINDOWS_STORE and WINDOWS_PHONE variables instead. https://cmake.org/cmake/help/latest/variable/WINDOWS_STORE.html https://cmake.org/cmake/help/latest/variable/WINDOWS_PHONE.html
These are set to true if the CMAKE_SYSTEM_NAME matches the specific system and available after project()

alalek:
:+1:

PR #15111, opened 2019-07-21
Remove unavoidable print of CV error

RealOrangeOne:
## This pullrequest changes

This PR removes a print which was impossible to hide. My use case is it's also printed in the Python bindings, and as brute-forcing the camera ids is the only reliable way to enumerate all cameras, it's annoying and noisy to print each time.

The return value covers whether the device exists. This might be better hidden behind a debug flag, but I couldn't work out how to do that nicely.

### DIFF
--- a/modules/videoio/src/cap_v4l.cpp
+++ b/modules/videoio/src/cap_v4l.cpp
@@ -788,13 +788,7 @@ bool CvCaptureCAM_V4L::open(int _index)
         name = cv::format("/dev/video%d", _index);
     }
 
-    /* Print the CameraNumber at the end of the string with a width of one character */
-    bool res = open(name.c_str());
-    if (!res)
-    {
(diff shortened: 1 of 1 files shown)


alalek:
1.
All `fprintf()` calls should be replaced with `CV_LOG_` macros (perhaps `CV_LOG_WARNING` / `CV_LOG_INFO`).

2.

So, please:

RealOrangeOne:
@alalek implemented your suggestions. Although in hindsight, i'm not sure whether keeping the log message. It's really not one which should be handed by OpenCV, more by the downstream user. 

I'm reasonably sure `WARNING` is the right level, but that's generally the default log level, so this likely won't actually solve my original issue!

mshabunin:
>this likely won't actually solve my original issue!

The message becomes avoidable. Set environment variable `OPENCV_LOG_LEVEL=ERROR` or `OPENCV_LOG_LEVEL=FATAL`.

Issue #15106, opened 2019-07-20
Strange output size in last convolution DNN face detector


lubagov:
in this commit, 03850008b9901ebe177b8f12efc47657ceab4fc4 by some reason, was changed padding in last convolution. After this, convolutions, conv8_2_h, conv9_2_h has output blob size 5x5. But, last detection levels should be 3x3 and 1x1, like in original SSD. When it detect big face 100 pixels side or more, i get strange effect:

![Frame_screenshot_20 07 2019](https://user-images.githubusercontent.com/6134203/61582874-fc3ec080-ab38-11e9-8e79-c2d4d1194e93.png)

When i remove this padding, and set to 0, it fine working:
![Frame_screenshot_20 07 2019_](https://user-images.githubusercontent.com/6134203/61582875-fc3ec080-ab38-11e9-952d-f7b4224c4c04.png)

for what, was this commit made?

dkurt:
1. Can you compare the model with and without padding with the same image (current issue description shows two different images)
2. There is a description of corresponding PR so you can understand the motivation: https://github.com/opencv/opencv/pull/10216

lubagov:
I can see this description, but it is trange description.
Really, each level SSD should detect higher size image, if add this padding all 3 blocks dtect image with 300/5=60 pixels (for 300x300 input), because blob output size:

layer:  conv6_1_h shape: (1, 128, 19, 19)
layer:  conv6_2_h shape: (1, 256, 10, 10)
layer:  conv7_1_h shape: (1, 64, 10, 10)
layer:  conv7_2_h shape: (1, 128, 5, 5)
layer:  conv8_1_h shape: (1, 64, 5, 5)
**layer:  conv8_2_h shape: (1, 128, 5, 5)
layer:  conv9_1_h shape: (1, 64, 5, 5)
layer:  conv9_2_h shape: (1, 128, 5, 5)**

after i add this padding back, shape sizes is:

layer:  conv7_1_h shape: (1, 64, 10, 10)
layer:  conv7_2_h shape: (1, 128, 5, 5)
layer:  conv8_1_h shape: (1, 64, 5, 5)
**layer:  conv8_2_h shape: (1, 128, 3, 3)
layer:  conv9_1_h shape: (1, 64, 3, 3)
layer:  conv9_2_h shape: (1, 128, 1, 1)**

Next image, received from Caffe samples, without OpenCV:
[...]

PR #10216, opened 2017-12-03
Extra paddings for face detection network

dkurt:
### This pullrequest changes

This patch lets run a face detection network on images with sizes smaller than 300x300 (i.e. 200x200) to improve efficiency. Extra paddings were added at two 3x3 convolution layers for it.

PR #15118, opened 2019-07-22
Fix false positives of face detection network for large faces

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/15106 after PR https://github.com/opencv/opencv/pull/10216

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/644

<cut/>

before:
out

with changes:
out_correct

### DIFF
--- a/modules/dnn/test/test_caffe_importer.cpp
+++ b/modules/dnn/test/test_caffe_importer.cpp
@@ -561,7 +561,7 @@ TEST(Test_Caffe, shared_weights)
 typedef testing::TestWithParam<tuple<std::string, Target> > opencv_face_detector;
 TEST_P(opencv_face_detector, Accuracy)
 {
-    std::string proto = findDataFile("dnn/opencv_face_detector.prototxt", false);
+    std::string proto = findDataFile("dnn/opencv_face_detector.prototxt");
     std::string model = findDataFile(get<0>(GetParam()), false);
     dnn::Target targetId = (dnn::Target)(int)get<1>(GetParam());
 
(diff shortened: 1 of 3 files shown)


dkurt:
@alalek, perhaps, we will need to update `opencv_face_detector.prototxt` on CI. Or it's better to replace tests to use an origin `deploy.prototxt`?

alalek:
Update file hash here too: https://github.com/opencv/opencv_extra/blob/3.4/testdata/dnn/download_models.py#L227-L229

And probably we need to change target branch to 3.4 in URL (I'll change that to master during 3.4 merging)

alalek:
In case of breaking of existed tests - add and use updated file under a new name, like `opencv_face_detector_20190722.prototxt`.

dkurt:
@alalek, please let's wait for @lubagov feedback before merge.

lubagov:
Red box, with original padding (pad: 0, on both), blu box, from this commit.

Frame_screenshot_23 07 2019

My opinion, `Conv_9 *` does not work at all, although `Conv_8 *` works. I think, if the resolution is less than 300x300, then the image needs to be scaled to 300x300. Or use protopxt, without the last one, or two SSD blocks. Depending on how much less resolution.

dkurt:
@lubagov, Thanks for quick response! May I ask you to share an origin image as well?

lubagov:
tstimg2
OpenCV_screenshot_23 07 2019_1
OpenCV_screenshot_23 07 2019_2

lubagov:
Please check, train.prototxt file also. Because it also has wrong padding, and wrong train network.

dkurt:
@lubagov, Will be fixed by https://github.com/opencv/opencv/pull/14938. Thanks!

PR #15122, opened 2019-07-22
fast_math.hpp performance improvements

pmur:
Optimize fast_math.hpp primitives for P8+ and GCC based systems.

Leverage compiler builtins to implement most rounding primitives. This should allow the compiler to choose more efficient instructions for the target architecture. PPC has dedicated rounding instructions. Hopefully this also carries over to other architectures too.

Notably, __builtin_lrint{,f} functions just call out to libm. Instead, include an inline solution similar to ARM.

This reduces the testing time by 200-300 seconds on POWER9.

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/core/include/opencv2/core/fast_math.hpp
+++ b/modules/core/include/opencv2/core/fast_math.hpp
@@ -74,7 +74,15 @@
 #  include "tegra_round.hpp"
 #endif
 
-#if defined __GNUC__ && defined __arm__ && (defined __ARM_PCS_VFP || defined __ARM_VFPV3__ || defined __ARM_NEON__) && !defined __SOFTFP__ && !defined(__CUDACC__)
+#if defined __PPC64__ && defined __GNUC__ && defined _ARCH_PWR8 && !defined (__CUDACC__)
+#  include <altivec.h>
+#endif
+
+#if ((defined _MSC_VER && defined _M_ARM) || defined CV_ICC || \
+        defined __GNUC__) && defined HAVE_TEGRA_OPTIMIZATION
+    #define CV_INLINE_ROUND_DBL(value) TEGRA_ROUND_DBL(value);
+    #define CV_INLINE_ROUND_FLT(value) TEGRA_ROUND_FLT(value);
+#elif defined __GNUC__ && defined __arm__ && (defined __ARM_PCS_VFP || defined __ARM_VFPV3__ || defined __ARM_NEON__) && !defined __SOFTFP__ && !defined(__CUDACC__)
     // 1. general scheme
     #define ARM_ROUND(_value, _asm_string) \
         int res; \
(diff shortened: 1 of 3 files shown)


alalek:
1. Optimization patches should [go](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) into [3.4 branch](https://github.com/opencv/opencv/wiki/Branches) first. We will merge changes from 3.4 into master regularly (weekly/bi-weekly).
2. Clang mimics for GCC.
Please try to use this condition (with extra macro to bypass this code path):
```
#if defined(__GNUC__) && !defined(__clang__) \
    && !defined(OPENCV_SKIP_FASTMATH_GCC_BUILTINS)
```
3. It would be nice to have simple perf tests for these functions.
4. You can check PowerPC build results here: https://ocv-power.imavr.com/#/opencv_pullrequests

modules/core/include/opencv2/core/fast_math.hpp
```diff
@@ -84,12 +92,56 @@
...
+       round and covert FP values. */
+    #define OPENCV_USE_FASTMATH_GCC_BUILTINS
+#endif
+
+/* Enable GCC builtin math functions if possible, desired, and available.
+   Note, not all math functions inline equally. E.g lrint will not inline
+   without the -fno-math-errno option. */
+#ifdef OPENCV_USE_FASTMATH_GCC_BUILTINS
+#  define _OPENCV_FASTMATH_ENABLE_GCC_MATH_BUILTINS ((defined __GNUC__ && !defined __clang__) \
+                                                     && defined OPENCV_USE_FASTMATH_GCC_BUILTINS)
```
alalek: Does it really works as expected?

> #if defined _OPENCV_FASTMATH_ENABLE_GCC_MATH_BUILTINS

modules/core/include/opencv2/core/fast_math.hpp
```diff
@@ -84,12 +92,56 @@
...
-        #define ARM_ROUND_DBL(value) ARM_ROUND(value, "vcvtr.s32.f64 %[temp], %[value] \n vmov %[res], %[temp]")
+        #define CV_INLINE_ROUND_DBL(value) ARM_ROUND(value, "vcvtr.s32.f64 %[temp], %[value] \n vmov %[res], %[temp]")
     #else
-        #define ARM_ROUND_DBL(value) ARM_ROUND(value, "vcvtr.s32.f64 %[temp], %P[value] \n vmov %[res], %[temp]")
+        #define CV_INLINE_ROUND_DBL(value) ARM_ROUND(value, "vcvtr.s32.f64 %[temp], %P[value] \n vmov %[res], %[temp]")
     #endif
     // 3. version for float
-    #define ARM_ROUND_FLT(value) ARM_ROUND(value, "vcvtr.s32.f32 %[temp], %[value]\n vmov %[res], %[temp]")
+    #define CV_INLINE_ROUND_FLT(value) ARM_ROUND(value, "vcvtr.s32.f32 %[temp], %[value]\n vmov %[res], %[temp]")
+#elif defined __PPC64__ && defined __GNUC__ && defined _ARCH_PWR8
```
alalek: Probably we need to apply `__CUDACC__` guard here too:
https://www.ibm.com/developerworks/community/blogs/fe313521-2e95-46f2-817d-44a4f27eba32/entry/how_to_develop_nvidia_cuda_applications_on_ibm_power8?lang=en

modules/core/perf/perf_cvround.cpp
```diff
@@ -5,41 +5,87 @@ namespace opencv_test
 using namespace perf;
 
 template <typename T>
-static void CvRoundMat(const cv::Mat & src, cv::Mat & dst)
+static void CvRoundMat(const cv::Mat & src, cv::Mat & dst, int (*round)(T))
```
alalek: Performance tests are degraded by itself due using of function pointer.

PR #15137, opened 2019-07-24
The facial landmarks via IE sample is added for dnn module

OrestChura:
### This pullrequest changes

- adds new sample that uses two models from IE: face detector (face-detection-adas-0001) and facial landmarks detector (facial-landmarks-35-adas-0002). The goal of the sample is to show how to inference IE models via dnn API. The sample detects described features on the source and then shows the result.
- added IE build flag check in samples/dnn/CMakeLists.txt as the sample requires IE.


OrestChura:
@dmatveev, @AsyaPronina, @andrey-golubev, @rgarnov, @dkurt Hello! Please, review my PR

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+//
+
+#include <opencv2/dnn.hpp>
+#include <opencv2/imgproc.hpp>
+#include <opencv2/highgui.hpp>
+using namespace cv;
+using namespace cv::dnn;
+
+#include <iostream>
+//#include "common.hpp"
```
AsyaPronina: Please remove

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+     "{ help  h              | |     Print the help message. }"
+     "{ facestruct           | |     Full path to a Face detection model structure file (for example, .xml file).}"
+     "{ faceweights          | |     Full path to a Face detection model weights file (for example, .bin file).}"
+     "{ landmstruct          | |     Full path to a facial Landmarks detection model structure file (for example, .xml file).}"
+     "{ landmweights         | |     Full path to a facial Landmarks detection model weights file (for example, .bin file).}"
+     "{ input                | |     Full path to an input image or a video file. Skip this argument to capture frames from a camera.}"
+     "{ b                    | |     The color to draw face rectangle and landmarks in format \"(b, g, r)\".}"
+     "{ g                    | |     The color to draw face rectangle and landmarks in format \"(b, g, r)\".}"
+     "{ r                    | |     The color to draw face rectangle and landmarks in format \"(b, g, r)\".}"
+     "{ faceframework        | |     Optional name of an origin framework of the facedetector. Detect it automatically if it does not set. }"
```
AsyaPronina: is not set

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+                                         "2: OpenCL fp16 (half-float precision), "
+                                         "3: VPU }"
+     "{ landmtarget          | 0 |   Choose one of target computation devices for landmarkdetector: "
+                                         "0: CPU target (by default), "
+                                         "1: OpenCL, "
+                                         "2: OpenCL fp16 (half-float precision), "
+                                         "3: VPU }"
+        );
+
+    parser.about("Use this script to run classification deep learning networks using OpenCV.");
```
AsyaPronina: About message need to be fixed

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+    const std::string landmbinPath = parser.get<std::string>("landmweights");
+    const Target landmTarget = parser.get<Target>("landmtarget");
+    const Backend landmBackend = parser.get<Backend>("landmbackend");
+
+    const Scalar color = Scalar(parser.get<int>("b"), parser.get<int>("g"), parser.get<int>("r"));
+
+    //Models' definition & initialization
+    Net faceNet = readNet(facexmlPath, facebinPath);
+    faceNet.setPreferableTarget(faceTarget);
+    faceNet.setPreferableBackend(faceBackend);
```
dkurt: There is only `DNN_INFERENCE_ENGINE_BACKEND` is supported fo IR models so you can emit `faceBackend` argument

dmatveev:
@AsyaPronina I asked to make an MR in our internal Gitlab, not here. Not sure if it makes sense to start merging it right now in any form. Please make an integration branch in Gitlab and move the MRs there (open against that integration branch).

dmatveev:
Still, if you think it is more useful and productive to continue here (and while a build is green), please continue here.

Not sure if any builder actually tests this code now, though (some custom builder configurations is required AFAIK).

samples/dnn/facial_landmarks_ie.cpp
OrestChura: Have I got it right that it has no sense even to ask the user about backends as there is only one of them available, so I can delete parser keys too?

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+    const unsigned int facerows = 384;
+
+    Net landmNet = readNet(landmxmlPath, landmbinPath);
+    landmNet.setPreferableTarget(landmTarget);
+    landmNet.setPreferableBackend(landmBackend);
+    const unsigned int landmcols = 60;
+    const unsigned int landmrows = 60;
+    const float bb_enlarge_coefficient = 1.0f;
+
+    const std::string win1 = "FaceLandmarkDetector";
```
AsyaPronina: win1 is not a good name in my opinion, why do you use 1?

OrestChura: Because there was another window some time ago.. My mistake, thank you

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
+// This file is part of OpenCV project.
```
AsyaPronina: Sorry for confuse but license header is not need in the samples folder

OrestChura: The strings 1-7 should be removed?

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    parser.about("Use this script to run classification deep learning networks using OpenCV.");
+    if (argc == 1 || parser.has("help"))
+    {
+        parser.printMessage();
+        return 0;
+    }
+
+    //Parsing input arguments
+    const std::string facexmlPath = parser.get<std::string>("facestruct");
```
AsyaPronina: If you use Camel case for variable naming please follow standard rules for this:
CamelCase Notation
CamelCase (sometimes called "CamelBack") notation is the practice of writing compound words or phrases in which the parts are joined without spaces, with each part's initial letter capitalized and the first letter in either upper or lower case. For example: SampleDataTable, computeExpr

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+    const std::string win1 = "FaceLandmarkDetector";
+    //TODO: rewrite color
+    const Scalar black(0,0,0);
+
+    //Input
+    VideoCapture cap;
+    if (parser.has("input"))
+        cap.open(parser.get<String>("input"));
+    else if (!cap.open(0))
+            {
```
AsyaPronina: Please align braces

samples/dnn/facial_landmarks_ie.cpp
OrestChura: There is the license header in, for example, colorization sample..

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+    while (waitKey(1) < 0)
+    {
+        cap >> img;
+        if (img.empty())
+        {
+           waitKey();
+           break;
+        }
+
+        Mat shwimg;         //Image to show
```
AsyaPronina: I suggest to stick to one naming style

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+        img.copyTo(shwimg);
+
+        //Preprocessing
+        Mat inputImg;
+        resize(img, inputImg, Size(facecols, facerows));
+
+        //Infering Facedetector
```
AsyaPronina: Face detector or FaceDetector

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+        faceNet.setInput(blobFromImage(inputImg));
+        Mat outFaces = faceNet.forward();
+
+        float* faceData = (float*)(outFaces.data);
+        for (size_t i = 0; i < outFaces.total(); i += faceObjectSize)
```
AsyaPronina: Very minor suggestion: if you use size_t it is better to use 0u

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+            //Face rectangle prediction processing
+            float confidence = faceData[i + 2];
+            if (confidence > faceConfThreshold)
+            {
+                int left = int(faceData[i + 3] * img.cols);
+                left = std::max(left, 0);
+                int top = int(faceData[i + 4] * img.rows);
+                top = std::max(top, 0);
+                int right  = int(faceData[i + 5] * img.cols);
+                right = std::min(right, img.cols-2);
```
AsyaPronina: Please stick to the one convention about using or not the spaces between operators and operands

samples/dnn/facial_landmarks_ie.cpp
```diff
@@ -0,0 +1,195 @@
...
+                int bottom = int(faceData[i + 6] * img.rows);
+                bottom = std::min(bottom, img.rows-2);
+                int width  = right - left + 1;
+                int height = bottom - top + 1;
+
+                rectangle(shwimg, Rect(left, top, width, height), color, 3);
+
+                //Postprocessing for landmarks
+                int max_of_sizes = std::max(width, height);
+                int add_width = int(max_of_sizes*bb_enlarge_coefficient) - width;
```
AsyaPronina: max_of_sizes and bb_enlarge_coefficient namings differ from Camel Case code convention you use in other code

AsyaPronina:
> @AsyaPronina I asked to make an MR in our internal Gitlab, not here. Not sure if it makes sense to start merging it right now in any form. Please make an integration branch in Gitlab and move the MRs there (open against that integration branch).

Okay will sort out with this today

bhack:
IMHO it is better to re-edit this with High Level API https://github.com/opencv/opencv/pull/14780 
@dvd42 is going to extend high level API proposal also for landmarks quite soon with a new PR.

dmatveev:
> IMHO it is better to re-edit this with High Level API #14780
> @dvd42 is going to extend high level API proposal also for landmarks quite soon with a new PR.

@bhack thanks for the pointer, but the idea of this is to trial #15090

bhack:
So what it means? Why high level API is not using G-API?

dmatveev:
> So what it means? Why high level API is not using G-API?

To be fair, I am not sure what the "high-level API" is. G-API has its own programming model so we're adding inference there now, and this one should be among the early samples.

dkurt:
@bhack, @dmatveev, It's just an abstraction over `dnn::Net` which let's to emit some preprocessing or postprocessing stuff. `dnn::Model` could as be as a node of G-API as use G-API internally so there is no conflict in two approaches.

dmatveev:
@dkurt thanks! Actually, efficient pipelining of preprocessing, inference, postprocessing, and other stuff is the reason why we add inference & streaming extensions in G-API. So what's the point and value-add of High-level API?

@bhack meanwhile, if you're interested, here's some [sample code](https://github.com/opencv/opencv/blob/014d18b6d54270444da85d6dcb0d1d3edf09b6f0/modules/gapi/test/infer/gapi_infer_ie_test.cpp) and some [more general](https://github.com/opencv/opencv/wiki/files/2018-12-24-GAPI_Overview.pdf) info on G-API.

dkurt:
@dmatveev, Due with this high-level API user do not use `blobFromImage`, we can use G-API as an optional dependency. It'll be transparent for user.

dvd42:
I created 1 PR for the DNN high_level api, adding a [keypoints](https://github.com/opencv/opencv/pull/15189) module. Maybe this chain of detection->landmarking could be implemented as part of said module by chaining the detection and keypointing module?

OrestChura:
> I created 1 PR for the DNN high_level api, adding a [keypoints](https://github.com/opencv/opencv/pull/15189) module. Maybe this chain of detection->landmarking could be implemented as part of said module by chaining the detection and keypointing module?

Thanks a lot! Have seen your PR, it looks reasonable to integrate our sample into your module, but is there any sense to add the InferenceEngine dependence into it? In my opinion, the sample is more intented to reflect the IE models usage in DNN API.

dvd42:
It seems that this won't be done yet as per [dkurts's](https://github.com/opencv/opencv/pull/14780#issuecomment-516391917) comment.

OrestChura:
The PR is closed as a similar sample exists already, so there's no need in this sample.

PR #15170, opened 2019-07-28
Update samples

sturkmen72:
### This pullrequest changes

updates samples

```
force_builders_only=linux,windows,docs
```

### DIFF
--- a/samples/cpp/peopledetect.cpp
+++ b/samples/cpp/peopledetect.cpp
@@ -31,9 +31,9 @@ public:
         // groupThreshold (set groupThreshold to 0 to turn off the grouping completely).
         vector<Rect> found;
         if (m == Default)
-            hog.detectMultiScale(img, found, 0, Size(8,8), Size(32,32), 1.05, 2, false);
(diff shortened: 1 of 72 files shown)


samples/cpp/peopledetect.cpp
```diff
@@ -5,7 +5,6 @@
 #include <opencv2/objdetect.hpp>
 #include <opencv2/highgui.hpp>
 #include <opencv2/imgproc.hpp>
-#include <opencv2/videoio.hpp>
```
alalek: Probably we should keep separate include of "videoio" module - `VideoCapture` is used below.
(due compatibility "videoio" headers are included by "highgui" header, but this doesn't look consistent - lets show to Users which modules are used)

sturkmen72: @alalek OK. let me add `#include <opencv2/videoio.hpp>` in all samples uses `VideoCapture`
please don't merge until i inform you.

alalek:
@sturkmen72 Could you please take a look on build failures?

sturkmen72:
@alalek i think the PR is ready.

PR #15188, opened 2019-07-30
Fix for building with older V4L2

mshabunin:
### This pullrequest changes

Added macro definition missing in older V4L2 headers.


alalek:
=> 3.4

PR #15190, opened 2019-07-30
G-API: verify y offset and height of Fluid output ROIs

andrey-golubev:
### This pullrequest

- Adds sanity checks for GFluidOutputRois
- Tests that invalid ROIs provided result in an exception throw


andrey-golubev:
Thanks!

PR #15199, opened 2019-07-31
Convert HOG from SSE SIMD to HAL - 35-45% faster on Power (VSX)

ChipKerchner:
Convert HOG from SSE SIMD to HAL - 35-45% faster on Power (VSX).

```
force_builders=ARMv8,Custom
```

### DIFF
--- a/modules/objdetect/src/hog.cpp
+++ b/modules/objdetect/src/hog.cpp
@@ -440,8 +440,8 @@ void HOGDescriptor::computeGradient(const Mat& img, Mat& grad, Mat& qangle,
 
             v_int32x4 mask0 = _hidx >> 31;
             v_int32x4 it0 = mask0 & _nbins;
-            mask0 = (_hidx < _nbins);
-            v_int32x4 it1 = ~mask0 & _nbins;
+            mask0 = (_hidx >= _nbins);
+            v_int32x4 it1 = mask0 & _nbins;
             _hidx += (it0 - it1);
 
             it0 = v_reinterpret_as_s32(v_pack(v_pack(_hidx, izero), v_reinterpret_as_s16(izero)));
--- a/modules/objdetect/src/hog.cpp
+++ b/modules/objdetect/src/hog.cpp
@@ -226,17 +226,6 @@ void HOGDescriptor::copyTo(HOGDescriptor& c) const
     c.signedGradient = signedGradient;
 }
 
-#if CV_NEON
-// replace of _mm_set_ps
-inline float32x4_t vsetq_f32(float f0, float f1, float f2, float f3)
-{
-    float32x4_t a = vdupq_n_f32(f0);
-    a = vsetq_lane_f32(f1, a, 1);
-    a = vsetq_lane_f32(f2, a, 2);
-    a = vsetq_lane_f32(f3, a, 3);
-    return a;
-}
-#endif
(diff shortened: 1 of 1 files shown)


mshabunin:
Shouldn't it be transformed differently?

Before:
```.cpp
#if CV_SSE2
...
#elif CV_NEON
...
#else
...
#endif
```
After:
```.cpp
#if CV_SIMD128 // or CV_SIMD for wide universal intrinsics
...
#else
...
#endif
```

ChipKerchner:
I could remove the NEON specific code and use the CV_SIMD128 for 3 platforms (SSE2, NEON, VSX) instead.

I just did NOT have a way of testing NEON.

alalek:
Yes, this is right way.
We would run tests on NEON-capable hardware.

modules/objdetect/src/hog.cpp
```diff
@@ -486,44 +487,42 @@ void HOGDescriptor::computeGradient(const Mat& img, Mat& grad, Mat& qangle,
...
-            it0 = _mm_unpacklo_epi8(it0, it1);
-
-            _mm_storel_epi64((__m128i*)(qanglePtr + x2), it0);
+            v_float32x4 _mag = v_load(dbuf + x + (width << 1));
+            v_float32x4 _angle = v_load(dbuf + x + width * 3);
+            _angle = (_angleScale * _angle) - fhalf;
+
+            v_int32x4 sign = (ione & v_reinterpret_as_s32(_angle < fzero));
+            v_int32x4 _hidx = v_trunc(_angle);
+            _hidx -= sign;
```
terfendail: I suppose `v_floor` intrinsic could be used here to compute `_hidx` instead of lines 502-504

modules/objdetect/src/hog.cpp
```diff
@@ -486,44 +487,42 @@ void HOGDescriptor::computeGradient(const Mat& img, Mat& grad, Mat& qangle,
...
+
+            v_float32x4 ft0 = _mag * (fone - _angle);
+            v_float32x4 ft1 = _mag * _angle;
+
+            v_store_interleave(gradPtr + x2, ft0, ft1);
+
+            v_int32x4 mask0 = _hidx >> 31;
+            v_int32x4 it0 = mask0 & _nbins;
+            mask0 = (_hidx < _nbins);
+            v_int32x4 it1 = ~mask0 & _nbins;
```
terfendail: I think the code will seem simpler if `>=` is used instead of `<` invertion

modules/objdetect/src/hog.cpp
```diff
@@ -663,31 +591,18 @@ void HOGCache::init(const HOGDescriptor* _descriptor,
...
-            _mm_storeu_ps(_di + i, t);
-        }
-    #elif CV_NEON
-        const int a[] = { 0, 1, 2, 3 };
-        int32x4_t idx = vld1q_s32(a);
-        float32x4_t _bw = vdupq_n_f32(bw), _bh = vdupq_n_f32(bh);
-        int32x4_t ifour = vdupq_n_s32(4);
+    #if CV_SIMD128
+        const float a[] = { 0.0, 1.0, 2.0, 3.0 };
+        v_float32x4 idx = v_load((float *)a);
```
alalek: `v_float32x4 idx(0.0f, 1.0f, 2.0f, 3.0f);` here and above (line 251).

Issue #15163, opened 2019-07-26
C++11 definition issue


neheb:
- OpenCV => 4.1.1
- Operating System / Platform = OpenWrt
- Compiler => GCC9 + uClibc-ng

uClibc-ng does not export several C++11 functions into the std namespace. As a result, I currently use this patch: 

```
diff --git a/modules/gapi/include/opencv2/gapi/core.hpp b/modules/gapi/include/opencv2/gapi/core.hpp
index a38d747..f01c0bd 100644
--- a/modules/gapi/include/opencv2/gapi/core.hpp
+++ b/modules/gapi/include/opencv2/gapi/core.hpp
@@ -392,8 +392,8 @@ namespace core {
             {
                 GAPI_Assert(fx != 0. && fy != 0.);
                 return in.withSize
-                    (Size(static_cast<int>(std::round(in.size.width  * fx)),
-                          static_cast<int>(std::round(in.size.height * fy))));
+                    (Size(static_cast<int>(round(in.size.width  * fx)),
+                          static_cast<int>(round(in.size.height * fy))));
             }
         }
     };
diff --git a/modules/gapi/include/opencv2/gapi/own/saturate.hpp b/modules/gapi/include/opencv2/gapi/own/saturate.hpp
index 207dcde..d716643 100644
--- a/modules/gapi/include/opencv2/gapi/own/saturate.hpp
+++ b/modules/gapi/include/opencv2/gapi/own/saturate.hpp
@@ -81,8 +81,8 @@ static inline DST saturate(SRC x, R round)
 // explicit suffix 'd' for double type
 inline double  ceild(double x) { return std::ceil(x); }
 inline double floord(double x) { return std::floor(x); }
-inline double roundd(double x) { return std::round(x); }
-inline double  rintd(double x) { return std::rint(x); }
+inline double roundd(double x) { return round(x); }
+inline double  rintd(double x) { return rint(x); }
 
 } //namespace own
 } //namespace gapi
diff --git a/modules/gapi/src/backends/fluid/gfluidcore.cpp b/modules/gapi/src/backends/fluid/gfluidcore.cpp
index 49d6d2e..e1821b8 100644
--- a/modules/gapi/src/backends/fluid/gfluidcore.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidcore.cpp
@@ -389,7 +389,7 @@ static void run_arithm_s1(uchar out[], const float in[], int width, const float
     cv::util::suppress_unused_warning(v_op);
[...]

PR #15211, opened 2019-08-03
G-API: Move several C++11 math functions out of std

neheb:
Some libc implementations like uClibc-ng do not place these functions in
std. Use the plain C versions.

resolves https://github.com/opencv/opencv/issues/15163

### DIFF
--- a/modules/gapi/src/backends/fluid/gfluidcore.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidcore.cpp
@@ -389,7 +389,7 @@ static void run_arithm_s1(uchar out[], const float in[], int width, const float 
     cv::util::suppress_unused_warning(v_op);
     for (; w < width; w++)
     {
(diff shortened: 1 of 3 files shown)


modules/gapi/src/backends/fluid/gfluidcore.cpp
```diff
@@ -1954,7 +1954,7 @@ GAPI_FLUID_KERNEL(GFluidCartToPolar, cv::gapi::core::GCartToPolar, false)
         {
             float x = in1[l];
             float y = in2[l];
-            float magnitude = std::hypot(y, x);
+            float magnitude = hypot(y, x);
```
elatkin: I think `hypotf` must be better here

elatkin:
@neheb  
This PR looks duplicated with my #15217  
Could you please review it?

PR #15223, opened 2019-08-04
Adding comment from source code to documentation

### LINKED ISSUE #14645: DetectionOutput layer doc
FasterRCnn use DetectionOutput as inference results. In source code it is written

[        // num() and channels() are 1.
        // Since the number of bboxes to be kept is unknown before nms, we manually
        // set it to maximal number of detections, [keep_top_k] parameter multiplied by batch size.
        // Each row is a 7 dimension std::vector, which stores
// [image_id, label, confidence, xmin, ymin, xmax, ymax]](https://github.com/LaurentBerger/opencv/blob/6c6dd16acf47da980f1f33ce578b02868f112b47/modules/dnn/src/layers/detection_output_layer.cpp#L221-L222)

Can we add this comment in [detectionOutput layer doc](https://docs.opencv.org/trunk/da/dbd/classcv_1_1dnn_1_1DetectionOutputLayer.html)?

kishdubey:
### This pullrequest changes
Adds a comment written in the source code for bool getMemoryShapes into the DetectionOutputLayer class documentation, as requested in the issue.

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -336,6 +336,11 @@ CV__DNN_INLINE_NS_BEGIN
          */
         virtual void unsetAttached();
 
+        /**
+         * num() and channels() are 1. 
+         * Since the number of bboxes to be kept is unknown before nms, we manually set it to maximal number of detections, [keep_top_k] parameter multiplied by batch size.
+         * Each row is a 7 dimension std::vector, which stores, [image_id, label, confidence, xmin, ymin, xmax, ymax]
+         */
         virtual bool getMemoryShapes(const std::vector<MatShape> &inputs,
                                      const int requiredOutputs,
                                      std::vector<MatShape> &outputs,


LaurentBerger:
issue  #14645

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -336,6 +336,11 @@ CV__DNN_INLINE_NS_BEGIN
...
         virtual void unsetAttached();
 
+        /**
+         * num() and channels() are 1. 
+         * Since the number of bboxes to be kept is unknown before nms, we manually set it to maximal number of detections, [keep_top_k] parameter multiplied by batch size.
+         * Each row is a 7 dimension std::vector, which stores, [image_id, label, confidence, xmin, ymin, xmax, ymax]
+         */
         virtual bool getMemoryShapes(const std::vector<MatShape> &inputs,
                                      const int requiredOutputs,
                                      std::vector<MatShape> &outputs,
```
dkurt: This is base class method. Most likely you wanted to change this file https://github.com/opencv/opencv/blob/master/modules/dnn/include/opencv2/dnn/all_layers.hpp?

alalek:
Source branch has been removed. PR is broken - can't test this patch on OpenCV CI.

PR #15228, opened 2019-08-05
fix exif bug

flyseaworld:
### This pullrequest changes

ExifReader::getNumDirEntry use the fixed offset value, which may parse error number of DirEntry in some cases.
Following is structure of TIFF header. We should get the offset through "Offset to first IFD" item.

image

get_thumbnail

### DIFF
--- a/modules/imgcodecs/src/exif.cpp
+++ b/modules/imgcodecs/src/exif.cpp
@@ -229,7 +229,7 @@ void ExifReader::parseExif()
 
     uint32_t offset = getStartOffset();
 
-    size_t numEntry = getNumDirEntry();
+    size_t numEntry = getNumDirEntry( offset );
 
     offset += 2; //go to start of tag fields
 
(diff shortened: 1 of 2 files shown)


alalek:
Details:

> Offset of IFD 4 0th IFD offset. If the TIFF header is followed immediately by the 0th IFD, it is written as 00000008.H.

http://www.digicamsoft.com/exif22/exif22/html/exif22_16.htm

flyseaworld:
> Details:
> 
> > Offset of IFD 4 0th IFD offset. If the TIFF header is followed immediately by the 0th IFD, it is written as 00000008.H.
> 
> http://www.digicamsoft.com/exif22/exif22/html/exif22_16.htm

It is only one common situation. Not all images have TIFF header followed by the oth IFD, such as the photo given above. Its offset of 0th IFD is 58 instead of fixed value 8. And the photo's orientation  is 6, but original version can't get the value.

PR #15230, opened 2019-08-05
in OpenCV.js added 'rotate' function and test sample

ilbalance:
```
force_builders=Custom
docker_image:Docs=docs-js
```

### DIFF
--- a/modules/js/test/test_imgproc.js
+++ b/modules/js/test/test_imgproc.js
@@ -944,13 +944,17 @@ QUnit.test('test_filter', function(assert) {
     //Rotate
     {
         let dst = new cv.Mat();
-        let src = cv.matFromArray(5, 2, cv.CV_8U, [1,2,3,4,5,6,7,8,9,0]);
+        let src = cv.matFromArray(3, 2, cv.CV_8U, [1,2,3,4,5,6]);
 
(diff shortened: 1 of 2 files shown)


modules/js/test/test_imgproc.js
```diff
@@ -941,4 +941,18 @@ QUnit.test('test_filter', function(assert) {
...
+    //Rotate
+    {
+        let dst = new cv.Mat();
+        let src = cv.matFromArray(5, 2, cv.CV_8U, [1,2,3,4,5,6,7,8,9,0]);
+
+        cv.rotate(src, dst, cv.ROTATE_90_CLOCKWISE);
+
+        size = dst.size();
+        assert.equal(size.height, 2, "ROTATE_HEIGHT");
+        assert.equal(size.width, 5, "ROTATE_WIGTH");
```
dkurt: I think it'd be nice to also add a test for data: take a look at `deepEqual` (there are examples in this file as well).

ilbalance: deepEqual is a function with no description. It used several times in test_imgproc.js file, but I do not understand, how it is works.
What do you mean - test for data? Test for created array?

PR #15237, opened 2019-08-05
Merge 3.4

alalek:
#14938 from LaurentBerger:face_clip
#15076
#15086
#15158 from dkurt:fix_tf_ssd_configs
#15168 from dkurt:dnn_onnx_15120
#15169 from dkurt:dnn_onnx_drop_vgg16
#15176 from dkurt:fix_14993
#15181 from berak:java_print_blob
#15188 from mshabunin:fix-v4l-old
#15195 from mshabunin:disable-pch
#15197 from mshabunin:restore-ipp-cut
#15202 from hugolm84:support_build_shared_for_wince
~~#15222 from hmaarrfk:verify_carbon~~
#15227 from tomoaki0705:fixFallThroughCarotene
#15228 from flyseaworld:exif/read_num_dir
#15233 from alalek:videoio_skip_msmf_tests
~~#15234 from alalek:backport~~
#15239 from alalek:videoio_skip_msmf_tests2

Previous "Merge 3.4": #15152

<cut/>

```
buildworker:Win64 OpenCL=windows-2
buildworker:Custom=linux-1,linux-2,linux-4
docker_image:Docs=docs-js
docker_image:Custom=javascript
#docker_image:Custom=powerpc64le
#docker_image:Custom=ubuntu-openvino:16.04
#buildworker:Custom=linux-2
#docker_image:Custom=ubuntu-vulkan:16.04
#buildworker:Custom=linux-4
#docker_image:Custom=fedora:28
#docker_image:Custom=ubuntu-cuda:16.04
build_image:Custom Mac=openvino-2019r2.0
build_image:Custom Win=openvino-2019r2.0
#build_image:Custom Win=msvs2017
#build_image:Custom Win=msvs2019
test_modules:Custom Mac=dnn,java,python3
```

PR #15241, opened 2019-08-05
Fix exif data failing to read on some images

Aldrog:
### This pullrequest changes

Fixes a bug in ExifReader causing it to fail on some specific jpegs (sample attached).

I do not have a deep knowledge of Exif structure, but exiv2 reads the number of entries in this manner and the patch fixes the problem for the attached image. I have also tested it with several other pictures taken with different cameras and didn't notice any regression.

Example

### DIFF
--- a/modules/imgcodecs/src/exif.cpp
+++ b/modules/imgcodecs/src/exif.cpp
@@ -229,7 +229,7 @@ void ExifReader::parseExif()
 
     uint32_t offset = getStartOffset();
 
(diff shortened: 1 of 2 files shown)


Aldrog:
Didn't notice it's already fixed in 3.4 branch by #15228. Is this PR still useful to fix the issue in 4.x?

alalek:
Thank you for contribution!
Mentioned patch has been merged into master branch here: #15237

PR #15242, opened 2019-08-05
Removing static linking of cuda library

vchiluka5:
Removes static linking of cuda library. This was added specifically since cuda driver APIs were used in opencv_contrib/modules/cudaoptflow/src/nvidiaOpticalFlow.cpp. Now we do Lazy loading of cuda.dll and hence static linking is no longer needed


vchiluka5:
Hi @alalek 
I see an exception in above build failure but i cant see anything wrt nvidia optical flow which failed.
Does this exception arised due to my PR?
Does some other module uses CUDA_CUDA_LIBRARY static Linking?
Can you please clarify the error/exception?

jsxyhelu:
sounds good!

PR #15246, opened 2019-08-06
cuda: add nvcc support of ccache

tomoaki0705:
# This pullrequest changes
 - This PR is a suggestion of feature.
 - ccache added support for nvcc in [3.4](https://ccache.dev/releasenotes.html#_ccache_3_4)
> Added support for NVCC compiler options --compiler-bindir/-ccbin, --output-directory/-odir and --libdevice-directory/-ldir.
 - Still, when we use this with CMake, [CMake 3.10 is required](https://gitlab.kitware.com/cmake/cmake/issues/16953).
 - But OpenCV 3.4 series rely on 2.8 and 4.0 series rely on 3.5.
 - So, I modified the CMake scripts to call nvcc via ccache

<cut/>

```
force_builders_only=Custom,linux,docs
buildworker:Custom=linux-4
```

### DIFF
--- a/cmake/FindCUDA.cmake
+++ b/cmake/FindCUDA.cmake
@@ -1531,7 +1531,7 @@ function(CUDA_LINK_SEPARABLE_COMPILATION_OBJECTS output_file cuda_target options
       add_custom_command(
         OUTPUT ${output_file}
         DEPENDS ${object_files}
-        COMMAND ${CUDA_NVCC_EXECUTABLE} ${nvcc_flags} -dlink ${object_files} -o ${output_file}
+        COMMAND ${OPENCV_NVCC_LAUNCHER} ${CUDA_NVCC_EXECUTABLE} ${nvcc_flags} -dlink ${object_files} -o ${output_file}
         ${flags}
         COMMENT "Building NVCC intermediate link file ${output_file_relative_path}"
         )
@@ -1540,7 +1540,7 @@ function(CUDA_LINK_SEPARABLE_COMPILATION_OBJECTS output_file cuda_target options
         TARGET ${cuda_target}
         PRE_LINK
         COMMAND ${CMAKE_COMMAND} -E echo "Building NVCC intermediate link file ${output_file_relative_path}"
-        COMMAND ${CUDA_NVCC_EXECUTABLE} ${nvcc_flags} ${flags} -dlink ${object_files} -o "${output_file}"
+        COMMAND ${OPENCV_NVCC_LAUNCHER} ${CUDA_NVCC_EXECUTABLE} ${nvcc_flags} ${flags} -dlink ${object_files} -o "${output_file}"
         )
     endif()
  endif()
--- a/cmake/OpenCVCompilerOptions.cmake
+++ b/cmake/OpenCVCompilerOptions.cmake
@@ -12,7 +12,9 @@ if(ENABLE_CCACHE AND NOT OPENCV_COMPILER_IS_CCACHE AND NOT CMAKE_GENERATOR MATCH
   # This works fine with Unix Makefiles and Ninja generators
   find_host_program(CCACHE_PROGRAM ccache)
   if(CCACHE_PROGRAM)
-    message(STATUS "Looking for ccache - found (${CCACHE_PROGRAM})")
+    execute_process(COMMAND ${CCACHE_PROGRAM} -V RESULT_VARIABLE CCACHE_RESULT OUTPUT_VARIABLE CCACHE_OUTPUT)
+    string(REGEX REPLACE "ccache version ([0-9]+\\.[0-9]+\\.?[0-9]*).*" "\\1" CCACHE_VERSION ${CCACHE_OUTPUT})
+    message(STATUS "Looking for ccache - found (${CCACHE_PROGRAM} : ${CCACHE_VERSION})")
     get_property(__OLD_RULE_LAUNCH_COMPILE GLOBAL PROPERTY RULE_LAUNCH_COMPILE)
     if(__OLD_RULE_LAUNCH_COMPILE)
       message(STATUS "Can't replace CMake compiler launcher")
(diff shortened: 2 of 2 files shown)


cmake/OpenCVCompilerOptions.cmake
```diff
@@ -29,6 +33,12 @@ if(ENABLE_CCACHE AND NOT OPENCV_COMPILER_IS_CCACHE AND NOT CMAKE_GENERATOR MATCH
         set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${__OLD_RULE_LAUNCH_COMPILE}")
       endif()
     endif()
+    if((NOT ${CCACHE_VERSION_MAJOR} LESS 3) AND (NOT ${CCACHE_VERSION_MINOR} LESS 4))
```
alalek: CMake has [builtin version comparison support](https://cmake.org/cmake/help/v3.14/command/if.html?highlight=VERSION_): `if(<variable|string> VERSION_LESS <variable|string>)`

So `CCACHE_VERSION_MAJOR|MINOR` extraction code is not needed.

cmake/FindCUDA/run_nvcc.cmake
```diff
@@ -149,7 +149,7 @@ macro(cuda_execute_process status command)
     execute_process(COMMAND ${CMAKE_COMMAND} -E echo ${cuda_execute_process_string})
   endif()
   # Run the command
-  execute_process(COMMAND ${ARGN} RESULT_VARIABLE CUDA_result )
+  execute_process(COMMAND ${OPENCV_NVCC_LAUNCHER} ${ARGN} RESULT_VARIABLE CUDA_result )
```
alalek: This file is a standalone build script.
Please check that this value is passed here properly (command line or `@CUDA_HOST_FLAGS@` above)

cmake/FindCUDA.cmake
```diff
@@ -1531,7 +1531,7 @@ function(CUDA_LINK_SEPARABLE_COMPILATION_OBJECTS output_file cuda_target options
       add_custom_command(
         OUTPUT ${output_file}
         DEPENDS ${object_files}
-        COMMAND ${CUDA_NVCC_EXECUTABLE} ${nvcc_flags} -dlink ${object_files} -o ${output_file}
+        COMMAND ${OPENCV_NVCC_LAUNCEHR} ${CUDA_NVCC_EXECUTABLE} ${nvcc_flags} -dlink ${object_files} -o ${output_file}
```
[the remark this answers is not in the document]
tomoaki0705: Oh my.  Thank you !!! fixed.

cmake/OpenCVCompilerOptions.cmake
tomoaki0705: Yes, that's much better ! Thank you !

tomoaki0705:
Thank you for the review.  I appreciate helpful review.

> Should we disable introduced OpenCV launcher in case of CMake >= 3.10 to avoid conflicts?

My understanding is NO.
[Here's a comment](https://gitlab.kitware.com/cmake/cmake/issues/16953#note_279455) from CMake gitlab.

> The project command accepts a list of languages to enable:

This means that ```.cu``` files are handled natively by CMake.
On the other hand, as far as I understand, ```.cu``` files in OpenCV are treated by custom process via CMake.
In another word, the build procedure doesn't change regardless the version of CMake.

I might be missing something, so I'm open for discussion.

cmake/FindCUDA/run_nvcc.cmake
tomoaki0705: I'm still struggling here.

tomoaki0705:
I think I have to close this PR
Reading the cmake scripts carefully, here's what I figured out.

1. Compiling with ```nvcc``` happens with ```run_nvcc.cmake``` script file, which I need to slip in the ```ccache``` command
2. Usually on a system, there are two ```run_nvcc.cmake``` on the system, one is [opencv/cmake/FindCUDA/run_nvcc.cmake](https://github.com/opencv/opencv/blob/d2c64b566bfc7339a93cb53c116603aabb61af2d/cmake/FindCUDA/run_nvcc.cmake) and the other one is ```/usr/share/cmake-xx/Module/FindCUDA/run_nvcc.cmake```.  The latter one is official CMake version, which was introduced in CMake 3.9.0 (according to the [comment](https://github.com/opencv/opencv/blob/d2c64b566bfc7339a93cb53c116603aabb61af2d/cmake/OpenCVDetectCUDA.cmake#L12)).
https://github.com/opencv/opencv/blob/d2c64b566bfc7339a93cb53c116603aabb61af2d/cmake/OpenCVDetectCUDA.cmake#L12
3. In OpenCV, just in case a user is using CMake older than 3.9.0, it calls cmake version ```run_nvcc.cmake```
https://github.com/opencv/opencv/blob/d2c64b566bfc7339a93cb53c116603aabb61af2d/cmake/OpenCVDetectCUDA.cmake#L17-L18
4. In Official ```run_cnvcc.cmake```, there is no space to slip in ```ccache``` even on the [latest CMake](https://gitlab.kitware.com/cmake/cmake/blob/52d9cd624fc0da9b26fd677d92e67dd033603337/Modules/FindCUDA/run_nvcc.cmake).
5. I can create PR for the OpenCV version of ```run_nvcc.cmake```, but for the official one, I have to work with CMake team to update it.

In addition, I confirmed that following simple CMakeLists.txt will launch nvcc via ccache.
```
project(MyProj C CXX CUDA)
find_program(CCACHE_PROGRAM ccache)
if(CCACHE_PROGRAM)
    set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE_PROGRAM}")
    set(CMAKE_CUDA_COMPILER_LAUNCHER "${CCACHE_PROGRAM}") # CMake 3.9+
endif()
add_executable(mycuda mycuda.cu main.cpp)
```
THIS is what the [issue](https://gitlab.kitware.com/cmake/cmake/issues/16953#note_279455) was saying.
In OpenCV, we compile each cu file for separate sm, so this approach is not taken.

It was my misunderstanding.

PR #15255, opened 2019-08-07
Rect intersection Improve to support only border overlap

### LINKED ISSUE #15254: Rect_ intersection improve to support only border overlap detect
##### Detailed description

template Rect_ union operator will return Rect(0,0,0,0) when two Rects have only border or point overlap, I think better to keep the overlap info, meaning may return a Rect who's width or/and height may be zero when there is only border or point overlap. 

##### Steps to reproduce

Rect(0, 0, 1, 1) & Rect(1, 0, 1, 1) will got Rect(0, 0, 0, 0) instead of Rect(1, 0, 0, 1)

iamgd67:
resolves #15254 
### This pullrequest changes

detect the situation when two Rect has only border or point overlap, and return the overlap info by a Rect who's width or/and height may be zero

### DIFF
--- a/modules/core/test/test_rect.cpp
+++ b/modules/core/test/test_rect.cpp
@@ -5,7 +5,7 @@
 
 namespace opencv_test { namespace {
 
-    TEST(Rect, rectUnionOperator) {        
+    TEST(Rect, rectUnionOperator) {
 
         Rect2f test = Rect();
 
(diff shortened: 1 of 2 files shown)


iamgd67:
win64 compile build fails because of test_videoio, looks like got nothing to do with this pull request

modules/core/test/test_rect.cpp
```diff
@@ -0,0 +1,20 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+#include "test_precomp.hpp"
+
+namespace opencv_test { namespace {
+
+    TEST(Rect, rectUnionOperator) {
```
dkurt: Intersection?

dkurt: I believe this test should be nearby similar ones.

iamgd67: it should be intersection, I will change it, both here and in the pr title and issue title, sorry for the language mistaken

alalek:
Closing, see issue's comment.

PR #15265, opened 2019-08-08
WINCE 8.0 requires ARMv7 Thumb2 and thus have NEON instructions

hugolm84:
### This pullrequest changes

This fixes a cv::exception being thrown when trying to use OpenCV and WINCE on ARMv7 Thumb2.

### DIFF
--- a/modules/core/src/system.cpp
+++ b/modules/core/src/system.cpp
@@ -554,7 +554,9 @@ struct HWFeatures
         have[CV_CPU_FP16] = true;
     #endif
     #endif
-
+    #if (defined(_WIN32_WCE) && _WIN32_WCE >= 0x800)
+        have[CV_CPU_NEON] = true;
+    #endif
     // there's no need to check VSX availability in runtime since it's always available on ppc64le CPUs
     have[CV_CPU_VSX] = (CV_VSX);
     // TODO: Check VSX3 availability in runtime for other platforms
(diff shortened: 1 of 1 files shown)


modules/core/src/system.cpp
```diff
@@ -554,7 +554,9 @@ struct HWFeatures
         have[CV_CPU_FP16] = true;
     #endif
     #endif
-
+    #if (defined(_WIN32_WCE) && _WIN32_WCE >= 0x800)
+        have[CV_CPU_NEON] = true;
```
alalek: Perhaps this should be under `defined __arm__` or `_M_ARM` condition.

hugolm84: Hi

The WINCE toolchain defines `/D_WIN32_WCE=0x800 /DUNDER_CE /DWINCE /DARM /D_ARM_ /DARMV7 /D_WINDOWS`.  I could add

```
#if defined __arm__ || defined _ARM_
#if defined _WIN32_WCE && _WIN32_WINCE >= 0x800
...
#elif __linux__
...
#elif
...
#endif
```

But feelt much more convoluted. Perhaps 

```
#if defined _ARM__ && defined _WIN32_WCE && _WIN32_WINCE >= 0x800
...
#endif
```
would suffice?

alalek: as you wish

hugolm84: Ok. I went with the _ARM_ solution

PR #15274, opened 2019-08-09
Convert lkpyramid from SSE SIMD to HAL - 90% faster on Power (VSX).

ChipKerchner:
Convert lkpyramid from SSE SIMD to HAL - 90% faster on Power (VSX).  Didn't convert NEON since it looks different from the SSE2.

### DIFF
--- a/modules/video/src/lkpyramid.cpp
+++ b/modules/video/src/lkpyramid.cpp
@@ -240,8 +240,8 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
         float A11, A12, A22;
 
 #if CV_SSE2 || CV_VSX
-        v_int32x4 qw0 = v_setall_s32(iw00 + (iw01 << 16));
-        v_int32x4 qw1 = v_setall_s32(iw10 + (iw11 << 16));
+        v_int16x8 qw0(iw00, iw01, iw00, iw01, iw00, iw01, iw00, iw01);
+        v_int16x8 qw1(iw10, iw11, iw10, iw11, iw10, iw11, iw10, iw11);
         v_int32x4 qdelta_d = v_setall_s32(1 << (W_BITS1-1));
         v_int32x4 qdelta = v_setall_s32(1 << (W_BITS1-5-1));
         v_float32x4 qA11 = v_setzero_f32(), qA12 = v_setzero_f32(), qA22 = v_setzero_f32();
@@ -288,8 +288,7 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
                 v_zip(v00, v01, t00, t01);
                 v_zip(v10, v11, t10, t11);
 
-                t0 = v_dotprod(t00, v_reinterpret_as_s16(qw0), qdelta) +
-                     v_dotprod(t10, v_reinterpret_as_s16(qw1));
+                t0 = v_dotprod(t00, qw0, qdelta) + v_dotprod(t10, qw1);
                 t0 = t0 >> (W_BITS1-5);
                 v_store_low(Iptr + x, v_pack(t0, t0));
 
@@ -301,8 +300,7 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
                 v_zip(v00, v01, t00, t01);
                 v_zip(v10, v11, t10, t11);
 
-                t0 = v_dotprod(t00, v_reinterpret_as_s16(qw0), qdelta_d) +
-                     v_dotprod(t10, v_reinterpret_as_s16(qw1));
+                t0 = v_dotprod(t00, qw0, qdelta_d) + v_dotprod(t10, qw1);
                 t1 = v_dotprod(t01, v_reinterpret_as_s16(qw0), qdelta_d) +
                      v_dotprod(t11, v_reinterpret_as_s16(qw1));
                 t0 = t0 >> W_BITS1;
@@ -310,8 +308,8 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
                 v00 = v_pack(t0, t1); // Ix0 Iy0 Ix1 Iy1 ...
                 v_store(dIptr, v00);
 
-                t0 = v_reinterpret_as_s32(v00) >> 16; // Iy0 Iy1 Iy2 Iy3
-                t1 = v_reinterpret_as_s32(v_reinterpret_as_u32(v00) << 16) >> 16; // Ix0 Ix1 Ix2 Ix3
+                v00 = v_reinterpret_as_s16(v_interleave_pairs(v_reinterpret_as_s32(v_interleave_pairs(v00))));
+                v_expand(v00, t1, t0);
 
                 v_float32x4 fy = v_cvt_f32(t0);
                 v_float32x4 fx = v_cvt_f32(t1);
@@ -426,13 +424,9 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
         }
 
 #if CV_SSE2 || CV_VSX
-        float CV_DECL_ALIGNED(16) A11buf[4], A12buf[4], A22buf[4];
-        v_store_aligned(A11buf, qA11);
-        v_store_aligned(A12buf, qA12);
-        v_store_aligned(A22buf, qA22);
-        iA11 += A11buf[0] + A11buf[1] + A11buf[2] + A11buf[3];
-        iA12 += A12buf[0] + A12buf[1] + A12buf[2] + A12buf[3];
-        iA22 += A22buf[0] + A22buf[1] + A22buf[2] + A22buf[3];
+        iA11 += v_reduce_sum(qA11);
+        iA12 += v_reduce_sum(qA12);
+        iA22 += v_reduce_sum(qA22);
 #endif
 
 #if CV_NEON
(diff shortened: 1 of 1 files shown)


modules/video/src/lkpyramid.cpp
```diff
@@ -419,11 +425,11 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
...
-#if CV_SSE2
+#if CV_SSE2 || CV_VSX
         float CV_DECL_ALIGNED(16) A11buf[4], A12buf[4], A22buf[4];
-        _mm_store_ps(A11buf, qA11);
-        _mm_store_ps(A12buf, qA12);
-        _mm_store_ps(A22buf, qA22);
+        v_store_aligned(A11buf, qA11);
+        v_store_aligned(A12buf, qA12);
+        v_store_aligned(A22buf, qA22);
         iA11 += A11buf[0] + A11buf[1] + A11buf[2] + A11buf[3];
```
alalek: Probably we should use [`v_reduce_sum()`](https://github.com/opencv/opencv/blob/4.1.1/modules/core/include/opencv2/core/hal/intrin_cpp.hpp#L1023-L1036) here instead of temporary buffer.

modules/video/src/lkpyramid.cpp
```diff
@@ -616,9 +622,9 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
                 }
             }
 
-#if CV_SSE2
+#if CV_SSE2 || CV_VSX
             float CV_DECL_ALIGNED(16) bbuf[4];
-            _mm_store_ps(bbuf, _mm_add_ps(qb0, qb1));
+            v_store_aligned(bbuf, qb0 + qb1);
             ib1 += bbuf[0] + bbuf[2];
             ib2 += bbuf[1] + bbuf[3];
```
alalek: @terfendail Could you suggest something here to avoid temporary buffer?

modules/video/src/lkpyramid.cpp
```diff
@@ -275,44 +274,51 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
...
+                     v_dotprod(t10, v_reinterpret_as_s16(qw1));
+                t1 = v_dotprod(t01, v_reinterpret_as_s16(qw0), qdelta_d) +
+                     v_dotprod(t11, v_reinterpret_as_s16(qw1));
+                t0 = t0 >> W_BITS1;
+                t1 = t1 >> W_BITS1;
+                v00 = v_pack(t0, t1); // Ix0 Iy0 Ix1 Iy1 ...
+                v_store(dIptr, v00);
+
+                t0 = v_reinterpret_as_s32(v00) >> 16; // Iy0 Iy1 Iy2 Iy3
+                t1 = v_reinterpret_as_s32(v_reinterpret_as_u32(v00) << 16) >> 16; // Ix0 Ix1 Ix2 Ix3
```
alalek: [`v_expand()`](https://github.com/opencv/opencv/blob/4.1.1/modules/core/include/opencv2/core/hal/intrin_cpp.hpp#L1158-L1175)?

modules/video/src/lkpyramid.cpp
```diff
@@ -239,13 +239,12 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
...
-#if CV_SSE2
-        __m128i qw0 = _mm_set1_epi32(iw00 + (iw01 << 16));
-        __m128i qw1 = _mm_set1_epi32(iw10 + (iw11 << 16));
-        __m128i z = _mm_setzero_si128();
-        __m128i qdelta_d = _mm_set1_epi32(1 << (W_BITS1-1));
-        __m128i qdelta = _mm_set1_epi32(1 << (W_BITS1-5-1));
-        __m128 qA11 = _mm_setzero_ps(), qA12 = _mm_setzero_ps(), qA22 = _mm_setzero_ps();
+#if CV_SSE2 || CV_VSX
+        v_int32x4 qw0 = v_setall_s32(iw00 + (iw01 << 16));
+        v_int32x4 qw1 = v_setall_s32(iw10 + (iw11 << 16));
```
alalek: Looks like this part of code assumes Little-endian platforms only.

And there are a lot of `v_reinterpret_as_s16(qw0)` statements below.

modules/video/src/lkpyramid.cpp
ChipKerchner: It is also deinterleaving the Ix and Iy components.

modules/video/src/lkpyramid.cpp
ChipKerchner: Changing to v_int16x8 with interleaved data.

modules/video/src/lkpyramid.cpp
ChipKerchner: Is there a HAL function(s) that deinterleaves register(s)?  I don't see any outside the v_load_interleave functions.

modules/video/src/lkpyramid.cpp
ChipKerchner: It could be done with

```
v_float32x4 qf0, qf1;
v_recombine(v_interleave_pairs(qb0 + qb1), v_setzero_f32(), qf0, qf1);
ib1 += v_reduce_sum(qf0);
ib2 += v_reduce_sum(qf1);
```

modules/video/src/lkpyramid.cpp
```diff
@@ -239,13 +239,12 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
...
         float A11, A12, A22;
 
-#if CV_SSE2
-        __m128i qw0 = _mm_set1_epi32(iw00 + (iw01 << 16));
-        __m128i qw1 = _mm_set1_epi32(iw10 + (iw11 << 16));
-        __m128i z = _mm_setzero_si128();
-        __m128i qdelta_d = _mm_set1_epi32(1 << (W_BITS1-1));
-        __m128i qdelta = _mm_set1_epi32(1 << (W_BITS1-5-1));
-        __m128 qA11 = _mm_setzero_ps(), qA12 = _mm_setzero_ps(), qA22 = _mm_setzero_ps();
+#if CV_SSE2 || CV_VSX
```
terfendail: Could you please use `#if CV_SIMD128 && !CV_NEON` here and below

modules/video/src/lkpyramid.cpp
```diff
@@ -275,44 +274,49 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
...
-                __m128 fy = _mm_cvtepi32_ps(t0);
-                __m128 fx = _mm_cvtepi32_ps(t1);
-
-                qA22 = _mm_add_ps(qA22, _mm_mul_ps(fy, fy));
-                qA12 = _mm_add_ps(qA12, _mm_mul_ps(fx, fy));
-                qA11 = _mm_add_ps(qA11, _mm_mul_ps(fx, fx));
+                v_int32x4 t0, t1;
+                v_int16x8 v00, v01, v10, v11, t00, t01, t10, t11;
+
+                v00 = v_reinterpret_as_s16(v_load_expand(src + x));
```
terfendail: There could be out of bound reading on the last iteration of the cycle. Probably it make sense to unroll the cycle to simultaneous processing of 8 elements.

modules/video/src/lkpyramid.cpp
```diff
@@ -275,44 +274,49 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
...
+                v00 = v_reinterpret_as_s16(v_load(dsrc));
+                v01 = v_reinterpret_as_s16(v_load(dsrc + cn2));
+                v10 = v_reinterpret_as_s16(v_load(dsrc + dstep));
+                v11 = v_reinterpret_as_s16(v_load(dsrc + dstep + cn2));
+
+                v_zip(v00, v01, t00, t01);
+                v_zip(v10, v11, t10, t11);
+
+                t0 = v_dotprod(t00, qw0, qdelta_d) + v_dotprod(t10, qw1);
+                t1 = v_dotprod(t01, v_reinterpret_as_s16(qw0), qdelta_d) +
```
terfendail: `v_reinterpret_as_s16` is unnecessary here

modules/video/src/lkpyramid.cpp
ChipKerchner: Unrolled to 8 elements.

modules/video/src/lkpyramid.cpp
```diff
@@ -616,11 +640,11 @@ void cv::detail::LKTrackerInvoker::operator()(const Range& range) const
...
             }
 
-#if CV_SSE2
-            float CV_DECL_ALIGNED(16) bbuf[4];
-            _mm_store_ps(bbuf, _mm_add_ps(qb0, qb1));
-            ib1 += bbuf[0] + bbuf[2];
-            ib2 += bbuf[1] + bbuf[3];
+#if CV_SIMD128 && !CV_NEON
+            v_float32x4 qf0, qf1;
+            v_recombine(v_interleave_pairs(qb0 + qb1), v_setzero_f32(), qf0, qf1);
```
terfendail: IMO it would be better to use `v_recombine(v_interleave_pairs(qb0), v_interleave_pairs(qb1), qf0, qf1);`. However I don't think it will essentially affect the performance.

terfendail: Addition will be performed as a part of following `v_reduce_sum`

PR #15279, opened 2019-08-11
jas_stream: Add definition for L_tmpnam if missing

neheb:
uClibc-ng for some reason does not provide a definition when some
deprecated APIs are disabled. Value taken from the musl libc.

### DIFF
--- a/3rdparty/libjasper/jas_stream.c
+++ b/3rdparty/libjasper/jas_stream.c
@@ -86,6 +86,10 @@
 #include <io.h>
 #endif
 
+#ifndef L_tmpnam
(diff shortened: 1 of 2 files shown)


alalek:
Hi!
We prefer to avoid patching of 3rdparty source code. It updates time-to-time, so such changes will be lost.

Is there some way to detect uClibc-ng by CMake?
To add this definition via CMakeLists.txt?

neheb:
It defines ```___UCLIBC___```

alalek:
@neheb Please verify updated patch with `uClibc-ng`.

neheb:
Works

PR #15282, opened 2019-08-12
G-API: fix compatibility with wingdi.h header on Windows systems

andrey-golubev:
### This pullrequest

- fixes compatibility issue between G-API and wingdi.h header on Windows systems
  - renames `cv::detail::ArgKind::OPAQUE` value to `GOPAQUE` to fix the conflict

---
__Issue__:
Compilation error occurs in G-API code if user includes both wingdi.h and gtype_traits.hpp. The latter is included in garg.hpp, gkernel.hpp, gtransform.hpp ("API" headers).

I imagine that the issue occurs when the order of includes is the following:
```cpp
#include "wingdi.h"  // defines OPAQUE value
#include "gtype_traits.hpp"  // uses ArgKind::OPAQUE
```
In that case opaque error description (hence the joke) is shown:
```bash
include/opencv2/gapi/gtype_traits.hpp(29): error: expected an identifier
         OPAQUE,       // Unknown, generic, opaque-to-GAPI data type - STATIC
         ^
include/opencv2/gapi/gtype_traits.hpp(44): error: expected an identifier
static constexpr const ArgKind kind = ArgKind::OPAQUE;
                                               ^
```
Instead of interpreting `ArgKind::OPAQUE` as a enum value, pre-processor likely substitutes a macro resulting in `ArgKind::<value defined by OPAQUE macro>` in the code and thus G-API code becomes invalid

---
According to the documentation:
https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-setbkmode
there's a defined value "OPAQUE". And in the source file of wingdi.h I see the following:
```cpp
/* Background Modes */
#define TRANSPARENT 1
#define OPAQUE 2
#define BKMODE_LAST 2
```
---
```bash
#build_gapi_standalone:Mac=ade-0.1.1f
```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/gtype_traits.hpp
+++ b/modules/gapi/include/opencv2/gapi/gtype_traits.hpp
@@ -26,9 +26,10 @@ namespace detail
     // a double dispatch
     enum class ArgKind: int
     {
-        GOPAQUE,      // Unknown, generic, opaque-to-GAPI data type - STATIC
+        OPAQUE_VAL,   // Unknown, generic, opaque-to-GAPI data type - STATIC
+                      // Note: OPAQUE is sometimes defined in Win sys headers
 #if !defined(OPAQUE) && !defined(CV_DOXYGEN)
-        OPAQUE = GOPAQUE,  // deprecated value used for compatibility only, use GOPAQUE instead
+        OPAQUE = OPAQUE_VAL,  // deprecated value used for compatibility, use OPAQUE_VAL instead
 #endif
         GOBJREF,      // <internal> reference to object
         GMAT,         // a cv::GMat
@@ -44,7 +45,7 @@ namespace detail
     template<typename T> struct GTypeTraits;
     template<typename T> struct GTypeTraits
     {
-        static constexpr const ArgKind kind = ArgKind::GOPAQUE;
+        static constexpr const ArgKind kind = ArgKind::OPAQUE_VAL;
     };
     template<>           struct GTypeTraits<cv::GMat>
     {
--- a/modules/gapi/include/opencv2/gapi/gtype_traits.hpp
+++ b/modules/gapi/include/opencv2/gapi/gtype_traits.hpp
@@ -27,6 +27,9 @@ namespace detail
     enum class ArgKind: int
     {
         GOPAQUE,      // Unknown, generic, opaque-to-GAPI data type - STATIC
+#if !defined(OPAQUE) && !defined(CV_DOXYGEN)
+        OPAQUE = GOPAQUE,  // deprecated value used for compatibility only, use GOPAQUE instead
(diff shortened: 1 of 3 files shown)


modules/gapi/include/opencv2/gapi/gtype_traits.hpp
```diff
@@ -26,7 +26,7 @@ namespace detail
     // a double dispatch
     enum class ArgKind: int
     {
-        OPAQUE,       // Unknown, generic, opaque-to-GAPI data type - STATIC
+        GOPAQUE,      // Unknown, generic, opaque-to-GAPI data type - STATIC
```
alalek: Add compatibility entry:
```
#if !defined(OPAQUE) && !defined(CV_DOXYGEN)
    OPAQUE = GOPAQUE,  // deprecated
#endif
```

andrey-golubev: thanks! by the way, why CV_DOXYGEN is required?

modules/gapi/include/opencv2/gapi/gtype_traits.hpp
```diff
@@ -26,7 +26,10 @@ namespace detail
...
```
dmatveev: `OPAQUE` is not `GOPAQUE`.

`OPAQUE` is every value which doesn't have a G-type for it.

`GOPAQUE` should be a `GOpaque<T>` thing for "manageable" user-defined types  in graph I/O. We still have that in our backlog :)

dmatveev: Proposed: `OPAQUE_VAL` or `OPAQUE_VALUE` (see what I had to do with `GModel::CONST` previously)

modules/gapi/include/opencv2/gapi/gtype_traits.hpp
andrey-golubev: ah, I think I got it. this way `OPAQUE` won't be seen in documentation

modules/gapi/include/opencv2/gapi/gtype_traits.hpp
andrey-golubev: agree, changed to `OPAQUE_VAL` to align with `Storage::CONST_VAL`

modules/gapi/include/opencv2/gapi/gtype_traits.hpp
andrey-golubev: updated

andrey-golubev:
@dmatveev addressed your comment, please review

andrey-golubev:
@alalek I guess this PR is merge-able

andrey-golubev:
Thanks!

PR #15283, opened 2019-08-12
G-API: Extend test suite on GAPI_TRANSFORM macro

smirnov-alexey:
This pullrequest adds more tests on GAPI_TRANSFORM macro

### DIFF
--- a/modules/gapi/test/gapi_transform_tests.cpp
+++ b/modules/gapi/test/gapi_transform_tests.cpp
@@ -23,6 +23,9 @@ using GMat3 = std::tuple<GMat, GMat, GMat>;
 using GScalar = cv::GScalar;
 template <typename T> using GArray = cv::GArray<T>;
 
+using ArrayT = int;
+using WrongArrayT = char;
+
 GAPI_TRANSFORM(gmat_in_gmat_out, <GMat(GMat)>, "gmat_in_gmat_out")
 {
     static GMat pattern(GMat) { return {}; }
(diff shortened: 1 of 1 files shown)


smirnov-alexey:
@andrey-golubev, @rgarnov, please, review.

modules/gapi/test/gapi_transform_tests.cpp
```diff
@@ -146,44 +164,102 @@ TEST(KernelPackageTransform, gmat_gsc_in_gmat_out)
...
+    typename std::enable_if<(cv::detail::GTypeTraits<InType>::kind != cv::detail::ArgKind::GARRAY &&
+                             cv::detail::GTypeTraits<OutType>::kind != cv::detail::ArgKind::GARRAY), void>::type
+    args_check(const cv::GComputation &comp)
+    {
+        const auto &p = comp.priv();
+        EXPECT_EQ(1u, p.m_ins.size());
+        EXPECT_EQ(1u, p.m_outs.size());
+        EXPECT_TRUE(ProtoContainsT<InType>(p.m_ins[0]));
+        EXPECT_TRUE(ProtoContainsT<OutType>(p.m_outs[0]));
+    }
```
rgarnov: Can all this checks be splitt into GArray / non-GArray check independent on input/output-ness? (So you can utilize such check for both input and output:)
```
template<typename Type>
typename std::enable_if<cv::detail::GTypeTraits<Type>::kind == cv::detail::ArgKind::GARRAY, void>::type
arg_check(const cv::GProtoArg &arg)
{
	// check array
}

// here should be a similar arg_check function for non-garray

// and args_check becomes something like
template<typename InType, typename OutType>   
args_check(const cv::GComputation &comp)
{
    const auto &p = comp.priv();
    EXPECT_EQ(1u, p.m_ins.size());
    EXPECT_EQ(1u, p.m_outs.size());

    arg_check<InType>(p.m_ins[0]);
    arg_check<OutType>(p.m_outs[0]);
}
```

PR #15285, opened 2019-08-12
 Added one more Preproc use case test for graph pattern matching

AsyaPronina:

### DIFF
--- a/modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
+++ b/modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
@@ -271,7 +271,7 @@ TEST(PatternMatching, TestMultiplePatternOuts)
     EXPECT_EQ(matching_test::V({dx_nh, dy_nh}), match.protoOuts());
 }
 
-TEST(PatternMatching, TestPreprocSplit3)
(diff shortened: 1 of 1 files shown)


modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
```diff
@@ -271,7 +271,7 @@ TEST(PatternMatching, TestMultiplePatternOuts)
     EXPECT_EQ(matching_test::V({dx_nh, dy_nh}), match.protoOuts());
 }
 
-TEST(PatternMatching, TestPreprocSplit3)
+TEST(PatternMatching, TestPrepResizeSplit3)
```
andrey-golubev: let's drop the "Prep" part everywhere
```suggestion
TEST(PatternMatching, TestResizeToNCHW)
```

AsyaPronina: Lets leave it, we may use --gtest_filter=*TestPrep*

andrey-golubev: we can use `--gtest_filter=*TestPrep*` as well with the new naming, but OK

Issue #9739, opened 2017-09-28
cv::cudacodec::VideoReader unable to Play rtsp stream 


pawan9:
##### System information (version)
- OpenCV => 3.3.0
- Operating System / Platform => Ubuntu 16.04, x86_64 
- Compiler => gcc version 5.4.1 20160904
- Cuda => 8.0
- Nvidia card => GTX 1080 Ti 

##### Detailed description

i am trying to play a rtsp stream using `cudacodec::VideoReader` 
###### Rtsp Stream Details ( from vlc )
![stream_details](https://user-images.githubusercontent.com/3811164/30963935-040eadcc-a46d-11e7-82b0-4970bc48756b.png)
this stream plays fine in vlc and `cv::VideoCapture` but when i try to play it in `cudacodec::VideoReader`  i get a **error** saying:

`OpenCV Error: Gpu API call (CUDA_ERROR_FILE_NOT_FOUND [Code = 301]) in CuvidVideoSource, file /home/sumit/Downloads/softwares/opencv/modules/cudacodec/src/cuvid_video_source.cpp, line 66
`
`Invalid UE golomb code`
`Invalid UE golomb code`
`OpenCV Error: Unsupported format or combination of formats (Unsupported video source) in FFmpegVideoSource, file /home/sumit/Downloads/softwares/opencv/modules/cudacodec/src/ffmpeg_video_source.cpp, line 110
terminate called after throwing an instance of 'cv::Exception'`
`  what():  /home/sumit/Downloads/softwares/opencv/modules/cudacodec/src/ffmpeg_video_source.cpp:110: error: (-210) Unsupported video source in function FFmpegVideoSource`

##### Steps to reproduce

```#include <iostream>
#include "opencv2/opencv_modules.hpp"

#if defined(HAVE_OPENCV_CUDACODEC)

#include <opencv2/core.hpp>
#include <opencv2/cudacodec.hpp>
#include <opencv2/highgui.hpp>

int main(int argc, const char* argv[])
{
    const std::string fname = "rtsp://admin:admin@192.168.1.13/media/video2";

    cv::namedWindow("GPU", cv::WINDOW_NORMAL);

    cv::cuda::GpuMat d_frame;
    cv::Ptr<cv::cudacodec::VideoReader> d_reader = cv::cudacodec::createVideoReader(fname);

    for (;;)
    {

        if (!d_reader->nextFrame(d_frame))
            break;

        cv::Mat frame;
        d_frame.download(frame);
[...]

PR #15290, opened 2019-08-13
Add retrieve encoded frame to VideoCapture

cudawarped:
### This pullrequest adds the functionality to retrieve the encoded bitstream for a grabbed video frame

Since opencv/opencv#14774 cudacodec has been broken because it needs the capacity to fall back on ffmpeg for parsing certain video files and rtsp streams.  This pull request adds the capacity to retrieve the encoded bit stream for each frame to the VideoCapture class as suggested in opencv/opencv_contrib#2180.

I have only built against the ffmpeg binaries (I am not sure how to build the plugin) I suspect that I may have broken the plugin functionality because I simply copied the existing retrieve method in `plugin_api.h` and `cap_ffmpeg.cpp`.

```
force_builders=linux,docs
buildworker:Custom=linux-1
build_image:Custom=fedora:31
```


cudawarped:
Hold this request, I am not that familiar with FFmpeg but it looks like the opencv [grab ](https://docs.opencv.org/4.1.1/d8/dfe/classcv_1_1VideoCapture.html#ae38c2a053d39d6b20c9c649e08ff0146)function which I am using to get the raw packet data [decodes ](https://github.com/opencv/opencv/blob/ddbd10c0019f3ee5f43b7902d47e7fc4303a6574/modules/videoio/src/cap_ffmpeg_impl.hpp#L1044) the frame as well.

mshabunin:
@cudawarped , I have several general comments/questions:

* do we really need class hierarchy here, can we make independent `VideoContainer` class which will use parts of same backend under the hood? For simplicity, we can make interface only for specific 
backend, i.e. omit probing and detection part and force users to always provide backend ID.
* as I understand not all media containers store one frame in one packet, so generally it is not possible to read "one frame" from demuxer. One should continuously read packets and pass them to decoder until it returns "I have decoded frame(s)" result. Consider this example: https://github.com/NVIDIA/video-sdk-samples/blob/aa3544dcea2fe63122e4feb83bf805ea40e58dbe/Samples/AppDecode/AppDec/AppDec.cpp#L51-L64 . I can imagine an interface similar to `std::istream` or `fread` with main method like this: `size_t readsome(uchar * buf, size_t sz)`.
* we need some tests for this feature, there are some video files here: https://github.com/opencv/opencv_extra/tree/master/testdata/highgui/video . I believe file big_buck_bunny.h264 is a raw stream extracted from one of containers (don't remember which one), so we can use it as reference data stream.
* it would be great to support reading several streams and returning metadata, but we can implement it later.

cudawarped:
>     * do we really need class hierarchy here, can we make independent `VideoContainer` class which will use parts of same backend under the hood? For simplicity, we can make interface only for specific
>       backend, i.e. omit probing and detection part and force users to always provide backend ID.

I would prefer not to mess with the class hierarchy as I have done if possible, but I couldn't think of a sensible way to use all the existing methods inside `VideoCapture` without inheriting from it.  Are you suggesting we make `VideoCapture` a member of `VideoContainer` as
`VideoCapture videoCapture;`
and expose additional methods in `VideoCapture `to allow `iCap `to be manipulated, e.g.
```
bool VideoCapture::grabEncoded()
{
    CV_INSTRUMENT_REGION();
    bool ret = !icap.empty() ? icap->grabEncodedFrame() : false;
    if (!ret && throwOnFail)
        CV_Error(Error::StsError, "");
    return ret;
}

bool VideoContainer::grab()
{
    return videoCapture.grabEncoded();
}
```
or to have a completely separate class VideoContainer with member
`Ptr<IVideoCapture> icap;`
and copy all the required method implementations from `VideoCapture `to `VideoContainer`, or something else?

mshabunin:
I was thinking about something like 
<details><summary>this</summary>

```.cpp
// videoio.hpp
class VideoContainer {
public:
    bool open(const std::string &, int);
    size_t read(uchar *, size_t);
    bool isOpened() const;
    // maybe something like: cv::Codec getCodec() const;
private:
    Ptr<IVideoCapture> icap;
};
```
```.cpp
// cap.cpp
bool VideoContainer::open(...) {
    // body similar to VideoCapture::open, some parts can be extracted to static function and reused
    icap = backend->createRawCapture(...);
}
size_t VideoContainer::read(...) {
    // ...
    icap->readRaw(...);
    // ...
}
```
```.cpp
// backend.hpp (.cpp)
Ptr<IVideoCapture> StaticBackend::createRawCapture(...) {
    auto cap = createCapture(...);
    if (cap && cap->setRaw(true)) return cap;
    return 0;
}
// TODO: implement PluginBackend::createRawCapture later
```
```.cpp
// cap_interface.hpp
class IVideoCapture {
    virtual bool setRaw(bool) { return false; }
    virtual size_t readRaw(...);
}
```
```.cpp
// cap_ffmpeg.cpp
class CvCapture_FFMPEG_proxy CV_FINAL : public cv::IVideoCapture {
 // implement setRaw and readRaw
}
```

</details>

Later we can rework internals and create hidden `IVideoContainer` or `IVideoSource`/`IRawVideoSource` classes with cleaner implementation. For now it is important to establish usable and convenient public interface (in videoio.hpp).

cudawarped:
@mshabunin I have added the suggested changes and have a few additional queries.

1. Do we need  a setRaw(...) method which can be turned on and off or would it be better to a) remove the method and pass a flag to the open method which performs this at initialization or b) have a setRaw() method which doesn't take any arguments and only turns on this capability?
2. I have not implemented any of the plugin parts because I am not sure how to build/test them can you advise?
3. The mp4 test file you suggested is h263 so to perform the testing I copied the h246 and h265 files to mp4 using

   > ffmpeg -i [big_buck_bunny.h264](https://github.com/opencv/opencv_extra/blob/master/testdata/highgui/video/big_buck_bunny.h264)  -codec copy big_buck_bunny_h264.mp4

   and
   > ffmpeg -i [big_buck_bunny.h265](https://github.com/opencv/opencv_extra/blob/master/testdata/highgui/video/big_buck_bunny.h265)  -codec copy big_buck_bunny_h265.mp4

    I am not familiar with h265 but I found that the [big_buck_bunny.h265](https://github.com/opencv/opencv_extra/blob/master/testdata/highgui/video/big_buck_bunny.h265) had a mixture of start code lengths, both 0x00 0x00 0x01 and 0x00 0x00 0x00 0x01 which is different to the h265 parsed from the big_buck_bunny_h265.mp4 with the [hevc_mp4toanexb](https://www.ffmpeg.org/ffmpeg-bitstream-filters.html#hevc_005fmp4toannexb) filter.  It may be an idea to add an additional file big_buck_bunny_annex.h265 converted using the hevc_mp4toanexb filter as

    > ffmpeg -i big_buck_bunny_h265.mp4 -codec copy -bsf:v hevc_mp4toannexb big_buck_bunny_h265_annex.h265

    Additionally in testing when ffmpeg reads from big_buck_bunny_h265_annex.h265 0x00 is included at the end of each read instead of the start.  This means that after the first read, each read from big_buck_bunny_h265.mp4 starts with 0x00 0x00 0x00 0x01 and each read from big_buck_bunny_h265_annex.h265 starts with 0x00 0x00 0x01, which I had to accommodate for in the test.
   Should I add the additional test files to the other repo or have you a better suggestion?

mshabunin:
1. Doing as a)-option describes would require many changes in all backends (_cap_interface.hpp_, `create*` functions), b)-option is fine.
2. I think we can postpone plugin support.
3. Why is it necessary to always add bitstream filter? Maybe it could be turned off by default?

Test files should be added to _opencv_extra_ repository using the same branch name as this PR (_ffmpeg_raw_retrieve_), then both PRs will be tested together.

cudawarped:
>     3\. Why is it necessary to always add bitstream filter? Maybe it could be turned off by default?

Good point, I am not sure that it is, my only use case for this PR is to decode with cudacodec where it is required.  Would you prefer a separate function to initialize this after creation (`useMp4ToAnexb()`) or a bool passed to read (`read(uchar** data, size_t* size, const bool mp4ToAnexB)`) which will initialize this the first time it is passed?

mshabunin:
OK, let's leave it enabled then. We will be able to extend interface later.

As for test data, I suggest replacing .h264 and .h265 with filtered streams and adding corresponding .mp4 files to leave the test simple. We can restore original files and add filtering test case later.

mshabunin:
@cudawarped , the test (1) should be dropped then.

cudawarped:
@mshabunin test dropped and redundant PR opencv/opencv_extra#673 adding additional test data closed.

cudawarped:
@mshabunin is there anything I can do to prepare this PR for merging so we can fix cudacodec?

mshabunin:
@cudawarped , sorry for delay, I'll take a look next week.
Did you get managed to make it work with the decoder? Do you have an example application?

cudawarped:
@mshabunin, no worries thanks for all your help.

> Did you get managed to make it work with the decoder?

The cudacodec decoder implementation from the latest commit to opencv/opencv_contrib#2180 should successfully decode frames retrieved using this [VideoContainer ](url)class and pass the included tests.

> Do you have an example application?

Are the test's sufficient?

Additionally I don't know if the ffmpeg plugin will compile against this.

cudawarped:
This change also applies the fixes #9739 with the change described [here](https://github.com/opencv/opencv/issues/9739#issuecomment-341885574) to the master branch.  I believe it has already been fixed on the 3.4 branch.

modules/videoio/src/cap.cpp
```diff
@@ -111,14 +105,17 @@ bool VideoCapture::open(const String& filename, int apiPreference)
                         if (param_VIDEOIO_DEBUG || param_VIDEOCAPTURE_DEBUG)
                             CV_LOG_WARNING(NULL, cv::format("VIDEOIO(%s): can't create capture", info.name));
                     }
-                } catch(const cv::Exception& e) {
-                    if(throwOnFail && apiPreference != CAP_ANY) throw;
+                }
```
mshabunin: Please restore formatting of this block.

modules/videoio/src/plugin_api.hpp
```diff
@@ -20,6 +20,7 @@ extern "C" {
 #endif
 
 typedef CvResult (CV_API_CALL *cv_videoio_retrieve_cb_t)(int stream_idx, unsigned const char* data, int step, int width, int height, int cn, void* userdata);
+typedef CvResult (CV_API_CALL* cv_videoio_retrieve_raw_cb_t)(unsigned const char* data, int step, void* userdata);
```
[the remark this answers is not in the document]
cudawarped: Does `Capture_retreive_raw` need to use a callback or can it just be defined as

`CvResult(CV_API_CALL *Capture_retreive_raw)(CvPluginCapture handle, uchar** data, size_t* size);`

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -174,11 +174,54 @@ enum VideoCaptureProperties {
...
+    VideoCodec_VC1,
+    VideoCodec_H264,
+    VideoCodec_JPEG,
+    VideoCodec_HEVC,
+    VideoCodec_VP8,
+    VideoCodec_VP9,
+    VideoCodec_NumCodecs,
+
+    // Uncompressed YUV
+    VideoCodec_YUV420 = (('I' << 24) | ('Y' << 16) | ('U' << 8) | ('V')),   // Y,U,V (4:2:0)
```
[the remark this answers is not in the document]
cudawarped: @mshabunin Those 

> Uncompressed YUV

are codecs returned by the nvcuvid video parser and can be removed, I incorrectly included them here.

I could use FOURCC codes for the remaining codecs but because there can be a many to one mapping from ffmpeg FOURCC codes to codec (h264, x264 etc.) it seemed more suitable to return the ffmpeg codec, do you agree?

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -1206,6 +1350,68 @@ double CvCapture_FFMPEG::getProperty( int property_id ) const
         return _opencv_ffmpeg_get_sample_aspect_ratio(ic->streams[video_stream]).num;
     case CAP_PROP_SAR_DEN:
         return _opencv_ffmpeg_get_sample_aspect_ratio(ic->streams[video_stream]).den;
+    case CAP_PROP_INT_CODEC:
```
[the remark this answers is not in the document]
cudawarped: Added simple tests checking both properties to the existing test

modules/videoio/src/cap_ffmpeg.cpp
```diff
@@ -380,7 +429,9 @@ static const OpenCV_VideoIO_Plugin_API_preview plugin_api_v0 =
     /*  9*/cv_writer_release,
     /* 10*/cv_writer_get_prop,
     /* 11*/cv_writer_set_prop,
-    /* 12*/cv_writer_write
+    /* 12*/cv_writer_write,
+    /* 13*/cv_capture_read_raw
+    /* 14*/cv_capture_set_raw
```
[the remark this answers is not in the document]
cudawarped: Now builds with -DVIDEOIO_PLUGIN_LIST=ffmpeg flag although the plugin code could still be incorrect.

modules/videoio/test/test_ffmpeg.cpp
```diff
@@ -95,6 +95,86 @@ TEST(videoio_ffmpeg, image)
...
+    if (!videoio_registry::hasBackend(api))
+        throw SkipTestException("Backend was not found");
+
+    const string path = get<1>(GetParam());
+    const string ext = get<2>(GetParam());
+    const string ext_raw = get<3>(GetParam());
+    const string codec = get<4>(GetParam());
+    const string pixelFormat = get<5>(GetParam());
+    const string fileName = path + "." + ext;
+    const string fileNameOut = cv::format("test_container_stream.%s", ext_raw.c_str());
```
mshabunin: `tempfile(...)`?

modules/videoio/test/test_ffmpeg.cpp
```diff
@@ -95,6 +95,86 @@ TEST(videoio_ffmpeg, image)
...
+            ASSERT_FALSE(file.fail());
+        }
+        ASSERT_GE(totalBytes, (size_t)4096) << "Encoded stream is too small";
+    }
+
+    std::cout << "Checking extracted video stream: " << fileNameOut << std::endl;
+
+    // Check decoded frames read from original media are equal to frames decoded from tmp file
+    {
+        VideoCapture capReference(findDataFile(fileName), CAP_FFMPEG);
```
mshabunin:
```suggestion
        VideoCapture capReference(findDataFile(fileName), api);
```

modules/videoio/test/test_ffmpeg.cpp
```diff
@@ -95,6 +95,86 @@ TEST(videoio_ffmpeg, image)
...
+        std::ofstream file(fileNameOut, ios::out | ios::trunc | std::ios::binary);
+        size_t totalBytes = 0;
+        Mat raw_data;
+        while (true)
+        {
+            container >> raw_data;
+            size_t size = raw_data.total();
+            if (raw_data.empty())
+                break;
+            CV_CheckTypeEQ(raw_data.type(), CV_8UC1, "");
```
mshabunin:
```suggestion
            ASSERT_EQ(raw_data.type(), CV_8UC1);
            ASSERT_TRUE(raw_data.isContinuous());
```

modules/videoio/test/test_ffmpeg.cpp
```diff
@@ -95,6 +95,86 @@ TEST(videoio_ffmpeg, image)
...
+        ASSERT_GE(totalBytes, (size_t)4096) << "Encoded stream is too small";
+    }
+
+    std::cout << "Checking extracted video stream: " << fileNameOut << std::endl;
+
+    // Check decoded frames read from original media are equal to frames decoded from tmp file
+    {
+        VideoCapture capReference(findDataFile(fileName), CAP_FFMPEG);
+        ASSERT_TRUE(capReference.isOpened());
+        VideoCapture capActual(fileNameOut.c_str(), CAP_FFMPEG);
```
mshabunin:
```suggestion
        VideoCapture capActual(fileNameOut.c_str(), api);
```

PR #15312, opened 2019-08-15
G-API: Support GArray as an input type in fluid

smirnov-alexey:
This pullrequest adds support for GArray as an input type in fluid

### DIFF
--- a/modules/gapi/src/backends/fluid/gfluidbackend.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidbackend.cpp
@@ -961,15 +961,9 @@ namespace
             {
                 std::set<int> in_hs, out_ws, out_hs;
 
-                //DELETE ME
-                bool has_garray = false;
-
                 for (const auto& in : node->inNodes())
                 {
                     const auto& d = g.metadata(in).get<Data>();
-                    if(d.shape == cv::GShape::GARRAY){
-                        has_garray = true;
-                    }
                     if (d.shape == cv::GShape::GMAT)
                     {
                         const auto& meta = cv::util::get<cv::GMatDesc>(d.meta);
@@ -990,10 +984,9 @@ namespace
 
                 auto &fu = fg.metadata(node).get<FluidUnit>();
 
-                GAPI_Assert(((out_ws.size() == 1 && out_hs.size() == 1) &&
+                GAPI_Assert((out_ws.size() == 1 && out_hs.size() == 1) &&
                             ((in_hs.size() == 1) ||
-                            ((in_hs.size() == 2) && fu.k.m_kind == cv::GFluidKernel::Kind::NV12toRGB))) ||
-                            has_garray);
+                            ((in_hs.size() == 2) && fu.k.m_kind == cv::GFluidKernel::Kind::NV12toRGB)));
 
                 const auto &op = g.metadata(node).get<Op>();
                 fu.line_consumption.resize(op.args.size(), 0);
@@ -1272,7 +1265,7 @@ void cv::gimpl::GFluidExecutable::packArg(cv::GArg &in_arg, const cv::GArg &op_a
         {
             in_arg = GArg(m_res.slot<cv::gapi::own::Scalar>()[ref.id]);
         }
-        if (ref.shape == GShape::GARRAY)
+        else if (ref.shape == GShape::GARRAY)
         {
             in_arg = GArg(m_res.slot<cv::detail::VectorRef>()[ref.id]);
         }
--- a/modules/gapi/src/backends/fluid/gfluidbackend.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidbackend.cpp
@@ -1229,6 +1229,7 @@ void cv::gimpl::GFluidExecutable::bindInArg(const cv::gimpl::RcDesc &rc, const G
     {
     case GShape::GMAT:    m_buffers[m_id_map.at(rc.id)].priv().bindTo(util::get<cv::gapi::own::Mat>(arg), true); break;
     case GShape::GSCALAR: m_res.slot<cv::gapi::own::Scalar>()[rc.id] = util::get<cv::gapi::own::Scalar>(arg); break;
+    case GShape::GARRAY:  m_res.slot<cv::detail::VectorRef>().at(rc.id) = util::get<cv::detail::VectorRef>(arg); break;
     default: util::throw_error(std::logic_error("Unsupported GShape type"));
     }
 }
--- a/modules/gapi/src/backends/fluid/gfluidbackend.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidbackend.cpp
@@ -961,9 +961,15 @@ namespace
             {
                 std::set<int> in_hs, out_ws, out_hs;
 
+                //DELETE ME
+                bool has_garray = false;
+
                 for (const auto& in : node->inNodes())
                 {
                     const auto& d = g.metadata(in).get<Data>();
+                    if(d.shape == cv::GShape::GARRAY){
+                        has_garray = true;
+                    }
                     if (d.shape == cv::GShape::GMAT)
                     {
                         const auto& meta = cv::util::get<cv::GMatDesc>(d.meta);
@@ -984,9 +990,10 @@ namespace
 
                 auto &fu = fg.metadata(node).get<FluidUnit>();
 
-                GAPI_Assert((out_ws.size() == 1 && out_hs.size() == 1) &&
+                GAPI_Assert(((out_ws.size() == 1 && out_hs.size() == 1) &&
                             ((in_hs.size() == 1) ||
-                            ((in_hs.size() == 2) && fu.k.m_kind == cv::GFluidKernel::Kind::NV12toRGB)));
+                            ((in_hs.size() == 2) && fu.k.m_kind == cv::GFluidKernel::Kind::NV12toRGB))) ||
+                            has_garray);
 
                 const auto &op = g.metadata(node).get<Op>();
                 fu.line_consumption.resize(op.args.size(), 0);
@@ -1229,7 +1236,7 @@ void cv::gimpl::GFluidExecutable::bindInArg(const cv::gimpl::RcDesc &rc, const G
     {
     case GShape::GMAT:    m_buffers[m_id_map.at(rc.id)].priv().bindTo(util::get<cv::gapi::own::Mat>(arg), true); break;
     case GShape::GSCALAR: m_res.slot<cv::gapi::own::Scalar>()[rc.id] = util::get<cv::gapi::own::Scalar>(arg); break;
-    case GShape::GARRAY:  m_res.slot<cv::detail::VectorRef>().at(rc.id) = util::get<cv::detail::VectorRef>(arg); break;
+    case GShape::GARRAY:  m_res.slot<cv::detail::VectorRef>()[rc.id] = util::get<cv::detail::VectorRef>(arg); break;
     default: util::throw_error(std::logic_error("Unsupported GShape type"));
     }
 }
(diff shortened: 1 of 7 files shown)


smirnov-alexey:
@rgarnov , @andrey-golubev , @dmatveev , please, review.

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -961,9 +961,15 @@ namespace
             {
                 std::set<int> in_hs, out_ws, out_hs;
 
+                //DELETE ME
+                bool has_garray = false;
```
smirnov-alexey: Need to figure out how to get rid of this

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -984,9 +990,10 @@ namespace
 
                 auto &fu = fg.metadata(node).get<FluidUnit>();
 
-                GAPI_Assert((out_ws.size() == 1 && out_hs.size() == 1) &&
+                GAPI_Assert(((out_ws.size() == 1 && out_hs.size() == 1) &&
```
smirnov-alexey: What does this assert stand for? I'm OK with a single output mat, but why it checks for exactly 1(2) input mats?

modules/gapi/test/gapi_fluid_test_kernels.cpp
```diff
@@ -448,6 +449,43 @@ GAPI_FLUID_KERNEL(FSum2MatsAndScalar, TSum2MatsAndScalar, false)
...
+    static void run(const std::vector<cv::Mat>    &arr,
+                          cv::gapi::fluid::Buffer &out)
+    {
+        std::vector<const uint8_t*> vec(arr.size());
+        for(unsigned int j = 0; j < arr.size(); ++j)
+        {
+            vec[j] = arr[j].ptr<uint8_t>();
+        }
+
+        static int row = 0;
```
smirnov-alexey: This `static` looks a little dirty

modules/gapi/test/gapi_fluid_test_kernels.hpp
```diff
@@ -112,6 +113,13 @@ G_TYPED_KERNEL_M(TSplit3_4lpi, <GMat3(GMat)>, "test.fluid.split3_4lpi") {
     }
 };
 
+G_TYPED_KERNEL(TSumArrayToMat, <GMat(GArray<cv::Mat>)>, "test.fluid.sum_array_to_mat")
+{
+    static GMatDesc outMeta(const cv::GArrayDesc&) {
+        return GMatDesc(CV_8U, 1, cv::Size(3, 3));
```
smirnov-alexey: Since we have no working GArrayDesc, there is no other ways to fill the Desc

modules/gapi/src/backends/fluid/gfluidbackend.cpp
smirnov-alexey: As discussed with Dima Matveev, there is no need to use fluid without mats (for a single row processing?). So, need to put static assert somewhere(probably macros) before this one, to check if we have at least 1 input mat in kernel.

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -961,9 +961,15 @@ namespace
...
 
+                //DELETE ME
+                bool has_garray = false;
+
                 for (const auto& in : node->inNodes())
                 {
                     const auto& d = g.metadata(in).get<Data>();
+                    if(d.shape == cv::GShape::GARRAY){
+                        has_garray = true;
+                    }
```
dmatveev: I believe you don't need to change anything here

dmatveev: I mean, this new code is not needed

modules/gapi/src/backends/fluid/gfluidbackend.cpp
dmatveev: Also you get this macro wrong.

It doesn't check how many input/output objects there, it checks if _dimensions_ of inputs/outputs are equal.

GArray doesn't (shouldn't) contribute here in any form.

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -1263,6 +1272,10 @@ void cv::gimpl::GFluidExecutable::packArg(cv::GArg &in_arg, const cv::GArg &op_a
         {
             in_arg = GArg(m_res.slot<cv::gapi::own::Scalar>()[ref.id]);
         }
+        if (ref.shape == GShape::GARRAY)
```
dmatveev: `else if`?

modules/gapi/test/gapi_fluid_test_kernels.cpp
```diff
@@ -448,6 +449,43 @@ GAPI_FLUID_KERNEL(FSum2MatsAndScalar, TSum2MatsAndScalar, false)
     }
 };
 
+GAPI_FLUID_KERNEL(FSumArrayToMat, TSumArrayToMat, false)
+{
+    static const int Window = 1;
+    static const int LPI    = 2;
+
+    static void run(const std::vector<cv::Mat>    &arr,
+                          cv::gapi::fluid::Buffer &out)
```
dmatveev: Does this work with fluid?

dmatveev: I believe it shouldn't :)

dmatveev: I'd keep this test here but mark it as DISABLED since we didn't design Fluid to handle cases like this.

modules/gapi/test/gapi_fluid_test_kernels.cpp
dmatveev: It shouldn't be here. Your `row` is `out.y()`

modules/gapi/test/gapi_fluid_test_kernels.hpp
dmatveev: I'd avoid testing cases like this.

Let's make `equalizeHistogram` kernel with GArray<T> input argument (a LUT). If it doesn't require more efforts

modules/gapi/src/backends/fluid/gfluidbackend.cpp
smirnov-alexey: Got it, thanks. Removed the logic

modules/gapi/test/gapi_fluid_test_kernels.cpp
smirnov-alexey: Changed the kernel

modules/gapi/test/gapi_fluid_test_kernels.hpp
smirnov-alexey: Done

modules/gapi/src/backends/fluid/gfluidbackend.cpp
smirnov-alexey: Removed

modules/gapi/src/backends/fluid/gfluidbackend.cpp
smirnov-alexey: Done, thanks

modules/gapi/test/gapi_fluid_test_kernels.cpp
smirnov-alexey: Removed that

modules/gapi/test/gapi_fluid_test.cpp
```diff
@@ -244,6 +244,45 @@ TEST(Fluid, Sum_2_Mats_And_Scalar)
...
+    cv::GComputation c(cv::GIn(mat, arr), cv::GOut(TEqualizeHist::on(mat, arr)));
+    cv::Mat in_mat(320, 480, CV_8UC1),
+            out_mat(320, 480, CV_8UC1),
+            ref_mat(320, 480, CV_8UC1);
+
+    cv::randu(in_mat, 200, 240);
+    std::vector<int> in_vec(256);
+
+    // Calculate normalized accumulated integral transformation array for gapi
+    for(int i = 0; i < in_mat.rows; ++i)
```
smirnov-alexey: Not the cleanest way to calculate a histogram array, but I didn't manage to find OCV functions for this

modules/gapi/test/gapi_fluid_test.cpp
```diff
@@ -244,6 +244,45 @@ TEST(Fluid, Sum_2_Mats_And_Scalar)
...
+        for(int j = 0; j < in_mat.cols; ++j)
+            ++in_vec[in_mat.at<uint8_t>(i, j)];
```
dmatveev: This is just cv::calcHist

dmatveev: Note, it should work with a regular std::vector for output

modules/gapi/test/gapi_fluid_test.cpp
```diff
@@ -244,6 +244,45 @@ TEST(Fluid, Sum_2_Mats_And_Scalar)
...
+    for(unsigned int i = 1; i < in_vec.size(); ++i)
+        in_vec[i]+=in_vec[i-1];
+
+    int min = 320*480;
+    for(unsigned int i = 0; i < in_vec.size(); ++i)
+        if(in_vec[i] != 0 && in_vec[i] < min)
+            min = in_vec[i];
+
+    for(auto & el : in_vec)
+        el = cvRound(((float)(el - min) / (float)(320*480 - min))*255);
```
dmatveev: Just comment what you do here (building a probability density function)

modules/gapi/test/gapi_fluid_test.cpp
```diff
@@ -244,6 +244,45 @@ TEST(Fluid, Sum_2_Mats_And_Scalar)
     EXPECT_EQ(0, cv::countNonZero(out_mat != ref_mat));
 }
 
+TEST(Fluid, EqualizeHist)
+{
+    cv::GMat mat;
+    cv::GArray<int> arr;
+
+    cv::GComputation c(cv::GIn(mat, arr), cv::GOut(TEqualizeHist::on(mat, arr)));
```
dmatveev: Also it would be a killer feature to test if the histogram itself was calculated in the graph.
You might need to write an extra OCV kernel to turn it into probability function though

modules/gapi/test/gapi_fluid_test_kernels.cpp
```diff
@@ -448,6 +449,30 @@ GAPI_FLUID_KERNEL(FSum2MatsAndScalar, TSum2MatsAndScalar, false)
...
+            //std::cout << "l=" << l << ": ";
+            for (int i = 0, w = mat.length(); i < w; i++)
+            {
+                //std::cout << std::setw(4) << int(in_row[i]);
+                out_row[i] = static_cast<uint8_t>(arr[in_row[i]]);
+            }
+            //std::cout << std::endl;
+        }
+    }
+};
```
dmatveev: Please drop the commented code

modules/gapi/test/gapi_fluid_test.cpp
smirnov-alexey: It doesn't work with vector and `calcHist` requires a lot variables to setup (convert single parameters into vectors). So, I think those 3 rows of code are much simpler

modules/gapi/test/gapi_fluid_test.cpp
smirnov-alexey: Where do you suggest to write the kernel? Just somewhere inside tests or add it to imgproc, for example?

modules/gapi/test/gapi_fluid_test.cpp
smirnov-alexey: Added, thanks

modules/gapi/test/gapi_fluid_test.cpp
smirnov-alexey: Done

modules/gapi/test/gapi_fluid_test_kernels.cpp
smirnov-alexey: Done

smirnov-alexey: Actually, some of the kernels there have commented prints

PR #15313, opened 2019-08-15
G-API: add transformation logic to GCompiler

andrey-golubev:
### This pullrequest changes

- Updates pattern matching functionality:
  - mapping of pattern to substitution (pt. 1)
  - applying substitution inside graph (pt. 2)
- Introduces 2 compiler passes (to `GCompiler`) (pt. 3):
  - check transformations pass (checks on correct transformations must be there)
  - apply transformations pass
- Adds high-level (integration) tests for pattern matching functionality

---
#### Description
1. Mapping of pattern to substitute (pt. 1):
This part is quite straightforward: given `pattern` inputs and `substitute` inputs just verify that they match correspondingly:
    - pattern inputs size == substitute inputs size
    - pattern input[i] has the same shape as substitute input[i]
The same is done for outputs

*Bottomline*: True enough, the matching is very basic and relies heavily on the user's ability to correctly align `pattern` with `substitute` (we just verify that everything is as it should be). I'm afraid, that we cannot do anything more than that just yet: all useful information is created after `inferMeta` pass of `GCompiler` if called for `pattern` and `substitute` (which is not done here and probably impossible in the general case)

2. Applying substitution (pt. 2):
This part relies heavily on existing functionality.
    - First, build `substitute` via `GModelBuilder` "inside" the main graph (a.k.a. fuse substitute into graph)
    - Second, according to `SubgraphMatch` information (pattern maps to graph, pattern maps to substitute), just "replace" OP nodes from main graph with corresponding OP nodes from substitute graph - this is done for all inputs and outputs
    - Third, erase old nodes from main graph + input/output DATA nodes of substitute (that is fused into main graph)

*Bottomline*: thus, the algorithm is quite simple. Just build substitute inside main graph, then fix connections so that data/op flow now goes through the substituted part

3. Compiler passes in `GCompiler` (pt. 3)
2 passes are added:
    1. "check_transformations" checks for endless loops that can occur if transformations are applied continuously. I doubt that I've covered all the possible cases of endless loops, so feel free to extend this part
    2. "apply_transformations" finds specified patterns in main graph and replaces them with substitutes (high-level API for patterns & substitutes specification - `GAPI_TRANSFORM`). This is done for as long as there are any patterns found (!)

---
#### TODOs:

- [x] unit tests for pt. 1
- [x] ~unit tests for pt. 2~ (properly tested via integration: much better semantically + can actually compare results (cv::Mats) after graphs execution)
- [x] integration tests for pt. 3 (need more corner cases?)
- [x] tests with multiple transformations applied at once (pt. 3 related)
- [x] ~split into multiple PRs~ (probably infeasible)
- [x] integrate compiler-level logic with `GTransform` API (separate effort) [added compiler pass => done]

```
```


andrey-golubev:
cc: @TolyaTalamanov

modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
```diff
@@ -994,4 +1002,293 @@ TEST(PatternMatching, adeSmokeTest)
...
+    compiled(input, output_transformed);
+
+    ASSERT_TRUE(AbsExact()(output_baseline, output_transformed));
+}
+
+using CompCreator = std::function<cv::GComputation()>;
+struct PatternMatchingHighLevel :
+    TestWithParam<std::tuple<CompCreator, CompCreator, CompCreator>> {};
+
+TEST_P(PatternMatchingHighLevel, UnifiedTest) {
```
andrey-golubev: Comment to reviewers: please look at this test first to understand the big picture

modules/gapi/src/compiler/passes/pattern_to_substitute_matching.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+namespace {
+using Graph = GModel::Graph;
+using Metadata = typename Graph::CMetadataT;
+using VisitedMatchings = std::list<std::pair<ade::NodeHandle, ade::NodeHandle>>;
+
+// Returns true if two DATA nodes are semantically and structurally identical:
+//  - both nodes have the same GShape
+//  - both nodes have the same storage
+//  - both nodes have the same meta
```
smirnov-alexey: Should be removed, since you check shape and storage only

modules/gapi/src/compiler/passes/pattern_to_substitute_matching.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+        // search optimization: push found nodes to the end
+        // FIXME: same iterator supported?
+        std::iter_swap(it, substituteNodes.begin() + (size - 1));
+        size--;
+    }
+    return matched;
+}
+
+SubgraphMatch::M matchDataNodes2(const Graph& pattern, const Graph& substitute,
```
smirnov-alexey: Need to choose version to use

modules/gapi/src/compiler/passes/pattern_to_substitute_matching.cpp
```diff
@@ -0,0 +1,176 @@
...
+    //---------------------------------------------------------------
+    // Match data nodes which start and end our pattern and substitute
+    const auto& patternDataInputs = patternP.in_nhs;
+    const auto& patternDataOutputs = patternP.out_nhs;
+
+    const auto& substituteDataInputs = substituteP.in_nhs;
+    const auto& substituteDataOutputs = substituteP.out_nhs;
+
+    // if number of data nodes doesn't match, abort
+    if (patternDataInputs.size() != substituteDataInputs.size()
```
smirnov-alexey: Looks like you can move the rest of the function below to another function, to prevent copypasting (both version of `findPatternToSubstituteMatch` got that).

modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
```diff
@@ -994,4 +1002,673 @@ TEST(PatternMatching, adeSmokeTest)
...
+        const cv::gimpl::SubgraphMatch::M& matched_nodes) {
+        // check that sizes are equal
+        EXPECT_EQ(p_nodes.size(), matched_nodes.size());
+        EXPECT_EQ(s_nodes.size(), matched_nodes.size());
+        for (size_t i = 0; i < p_nodes.size(); ++i) {
+            const auto& p_node = p_nodes[i];
+            const auto& s_node = s_nodes[i];
+
+            auto found = matched_nodes.find(p_node);
+            ASSERT_NE(matched_nodes.cend(), found);  // asssert - cannot de-reference end()
```
smirnov-alexey: Minor typo in the comment

modules/gapi/src/compiler/passes/pattern_to_substitute_matching.cpp
andrey-golubev: Updated this part, thanks!

modules/gapi/src/compiler/passes/pattern_to_substitute_matching.cpp
andrey-golubev: chosen `matchDataNodes2` and renamed

modules/gapi/src/compiler/passes/pattern_to_substitute_matching.cpp
andrey-golubev: deleted the 2nd method

modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
andrey-golubev: addressed, thanks

modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
```diff
@@ -994,4 +998,794 @@ TEST(PatternMatching, adeSmokeTest)
...
+        std::make_tuple(
+            [] () {
+                GMat in;
+                GMat out = cv::gapi::resize(in, cv::Size(100, 100));
+                return cv::GComputation(cv::GIn(in), cv::GOut(out));
+            },
+            cv::compile_args(
+                cv::gapi::kernels<EndlessLoopTransformChain2, EndlessLoopTransformChain1>())
+        )
+    ));
```
dmatveev: Test cases are nice, but please make it more "plain" (more cohesion, less coupling)

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,102 @@
...
+    if (matchedDataOutputs.empty()) {
+        return {};
+    }
+
+    //---------------------------------------------------------------
+    // Construct SubgraphMatch object
+    SubgraphMatch match;
+    match.inputDataNodes = std::move(matchedDataInputs);
+    match.outputDataNodes = std::move(matchedDataOutputs);
+    return match;
```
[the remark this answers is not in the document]
andrey-golubev: New logic. This part is matching of __pattern__ to __substitute__. Indeed it is simpler than finding a pattern in a __main graph__: we just need to match inputs && outputs (the matching itself is very simple && I'm not sure how we can make it more robust). I re-used `SubgraphMatch` here (introduced before & already in master) to reduce the amount of entities "in flight"

modules/gapi/src/compiler/passes/perform_substitution.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "pattern_matching.hpp"
+
+namespace cv { namespace gimpl {
+namespace {
+using Graph = GModel::Graph;
+
+template<typename It, typename Callable>
+void erase(Graph& g, It first, It last, Callable get) {
+    for (; first != last; ++first) {
+        ade::NodeHandle node = get(first);
```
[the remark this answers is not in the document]
andrey-golubev: let's rename then

modules/gapi/src/compiler/passes/perform_substitution.cpp
```diff
@@ -0,0 +1,72 @@
...
+        if (node == nullptr) continue;  // some nodes might already be erased
+        g.erase(node);
+    }
+}
+}  // anonymous namespace
+
+void performSubstitution(Graph& graph,
+    const SubgraphMatch& patternToGraph, const SubgraphMatch& patternToSubstitute) {
```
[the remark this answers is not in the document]
andrey-golubev: re-used existing class. it felt easier to use this way

modules/gapi/src/compiler/passes/perform_substitution.cpp
```diff
@@ -0,0 +1,72 @@
...
+    // substitute input nodes
+    for (const auto& inputNodePair : patternToGraph.inputDataNodes) {
+        // Note: we don't replace input DATA nodes here, only redirect their output edges
+        const auto& patternDataNode = inputNodePair.first;
+        const auto& graphDataNode = inputNodePair.second;
+        const auto& substituteDataNode = patternToSubstitute.inputDataNodes.at(patternDataNode);
+        GModel::redirectReaders(graph, substituteDataNode, graphDataNode);
```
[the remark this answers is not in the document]
andrey-golubev: readers: graph data node is INPUT node for pattern and substitute. so what we do here is replace the following OP nodes (from pattern) to OP nodes from substitute and this is achieved via `redirectReaders`

modules/gapi/src/compiler/passes/perform_substitution.cpp
```diff
@@ -0,0 +1,72 @@
...
+    for (const auto& outputNodePair : patternToGraph.outputDataNodes) {
+        // Note: we don't replace output DATA nodes here, only redirect their input edges
+        const auto& patternDataNode = outputNodePair.first;
+        const auto& graphDataNode = outputNodePair.second;
+        const auto& substituteDataNode = patternToSubstitute.outputDataNodes.at(patternDataNode);
+        for (auto e : graphDataNode->inEdges()) {
+            graph.erase(e);
+        }
+        GModel::redirectWriter(graph, substituteDataNode, graphDataNode);
+    }
```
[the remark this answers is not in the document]
andrey-golubev: writers: graphDataNode here is OUTPUT for pattern/substitute and so we need to connect "last level" OP nodes from substitute to it (without substituting the DATA node itself - otherwise, we get a new output that doesn't match to user's output data)

---
It's fused (2nd case): this way it is much simpler to implement both substitute nodes construction && replacement of pattern with substitute.

First case (I have it in commit history as well): we'd just need to traverse the substitute graph and re-create each node in main graph + connect them as in substitute as we go through

modules/gapi/src/compiler/passes/pattern_matching.hpp
```diff
@@ -36,12 +38,20 @@ namespace gimpl {
...
+            return    inputDataNodes.empty() && startOpNodes.empty()
+                   && finishOpNodes.empty() && outputDataNodes.empty()
+                   && inputTestDataNodes.empty() && outputTestDataNodes.empty();
+        }
+
         bool ok() const {
             return    !inputDataNodes.empty() && !startOpNodes.empty()
                    && !finishOpNodes.empty() && !outputDataNodes.empty()
                    && !inputTestDataNodes.empty() && !outputTestDataNodes.empty();
+        }
```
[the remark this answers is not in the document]
andrey-golubev: not really (I think): in `empty()` __everything__ is empty. so `!empty()` gives us that at __least one__ container is not empty -> and this is not the same as __all__ not empty (you can verify this with de Morgan's boolean logic rules)

modules/gapi/src/compiler/passes/perform_substitution.cpp
```diff
@@ -0,0 +1,72 @@
...
+    erase(graph, patternToGraph.internalLayers.begin(),
+        patternToGraph.internalLayers.end(), get_from_node);
+
+    // erase old finish OP nodes of __main graph__
+    erase(graph, patternToGraph.finishOpNodes.begin(),
+        patternToGraph.finishOpNodes.end(), get_from_pair);
+
+    // erase output data nodes of __substitute__
+    erase(graph, patternToSubstitute.outputDataNodes.begin(),
+        patternToSubstitute.outputDataNodes.end(), get_from_pair);
```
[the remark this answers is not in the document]
andrey-golubev: true, tried to follow standard library style, will update

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    // Store Computation's protocol in metadata
+    cv::gimpl::Protocol p;
+    std::tie(p.inputs, p.outputs, p.in_nhs, p.out_nhs) = proto_slots;
+    gm.metadata().set(p);
+
+    return pG;
+}
+
+bool transform(ade::Graph& main, const std::unique_ptr<ade::Graph>& patternG,
```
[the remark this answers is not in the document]
andrey-golubev: agree, will do

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    // 4. make substitution
+    performSubstitution(gm, match1, match2);
+    return true;
+}
+}  // anonymous namespace
+
+void checkTransformations(ade::passes::PassContext&,  // FIXME: context is unused here
+    const gapi::GKernelPackage& transformations,
+    std::vector<std::unique_ptr<ade::Graph>>& patterns)
```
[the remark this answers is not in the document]
andrey-golubev: I think it's actually the same (as far as I remember): if there's enough space on a single line place everything on one line, if not, go to the second, once second line is "full", go to the 3rd. anyway, I guess I'll align with other files

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,144 @@
...
+{
+    const auto& transforms = transformations.get_transformations();
+    const auto size = transforms.size();
+    if (0 == size) return;
```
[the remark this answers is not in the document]
andrey-golubev: should be `0u`. as a side note, I'm not sure who thought that this is a good idea to put extra `u`, `UL`, etc. (and report compile warnings otherwise!) for cases when we compare a variable of a very specific known type and a literal... to my mind, literal should be automatically cast to the type of the variable (disregarding the mutual order of literal and variable)

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,144 @@
...
+    cv::gimpl::GModelBuilder builder(g);
+    auto proto_slots = builder.put(c.priv().m_ins, c.priv().m_outs);
+
+    // Store Computation's protocol in metadata
+    cv::gimpl::Protocol p;
+    std::tie(p.inputs, p.outputs, p.in_nhs, p.out_nhs) = proto_slots;
+    gm.metadata().set(p);
+
+    return pG;
+}
```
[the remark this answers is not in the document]
andrey-golubev: matching graph (pattern). makes sense to rename. In general, this function should go to GModel I believe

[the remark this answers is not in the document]
andrey-golubev: oh wait, I also re-use it in check_transformations pass. this is generic and builds both versions

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,144 @@
...
+    for (size_t i = 0; i < size; ++i) {
+        auto& p = patterns[i];
+        if (p == nullptr) {
+            p = makeGraph(transforms[i].pattern());
+        }
+        for (size_t j = i; j < size; ++j) {
+            auto& s = substitutes[j];
+            if (s == nullptr) {
+                s = makeGraph(transforms[j].substitute());
+            }
```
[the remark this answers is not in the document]
andrey-golubev: I build both patterns and substitutes only once (there's a `s == nullptr` specifically for that). as for checking each pattern against each substitute: it's essential for cases when 2 transformations cancel each other, so we get an endless substitution loop in apply_transformations pass

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,144 @@
...
+            auto matchInSubstitute = findMatches(*p, *s);
+            if (!matchInSubstitute.empty()) {
+                throw std::runtime_error("Error: pattern detected inside substitute");
+            }
+        }
+    }
+}
+
+void applyTransformations(ade::passes::PassContext& ctx,
+    const gapi::GKernelPackage& transformations,
```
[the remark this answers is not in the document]
andrey-golubev: true, will rename to that

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,144 @@
...
+            bool transformationApplied = true;
+            while (transformationApplied)
+            {
+                transformationApplied = transform(ctx.graph, p, t.substitute());
+
+                // if at least one transformation happend, it is possible that other transforms will
+                // be applied in the next "round"
+                continueTransforming |= transformationApplied;
+            }
+        }
```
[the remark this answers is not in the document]
andrey-golubev: let's discuss separately, I'm not sure

andrey-golubev:
@dmatveev updated all but tests, please review

modules/gapi/src/compiler/passes/transformations.cpp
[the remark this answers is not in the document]
andrey-golubev: fixed

modules/gapi/src/compiler/passes/perform_substitution.cpp
```diff
@@ -0,0 +1,72 @@
...
+void erase(Graph& g, It first, It last, Callable get) {
+    for (; first != last; ++first) {
+        ade::NodeHandle node = get(first);
+        if (node == nullptr) continue;  // some nodes might already be erased
+        g.erase(node);
+    }
+}
+}  // anonymous namespace
+
+void performSubstitution(Graph& graph,
```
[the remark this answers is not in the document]
andrey-golubev: fixed

modules/gapi/src/compiler/passes/pattern_matching.cpp
```diff
@@ -560,7 +569,7 @@ cv::gimpl::findMatches(const cv::gimpl::GModel::Graph& patternGraph,
         // Create vector with the correctly ordered OUT data nodes in the test subgraph
         std::vector<ade::NodeHandle> outputTestDataNodes;
         for (const auto& patternOutNode : patternOutputDataNodes) {
-            outputTestDataNodes.push_back(outputApiMatch[patternOutNode]);
+            outputTestDataNodes.push_back(outputApiMatch.at(patternOutNode));
```
AsyaPronina: could you please consolidate these changes on the other code list when it is required?

modules/gapi/src/compiler/passes/pattern_matching.hpp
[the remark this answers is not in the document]
AsyaPronina: I think we can say that SubgraphMatch is empty if there is one empty container here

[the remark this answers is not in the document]
andrey-golubev: not sure: you can find there's a `partialOk` as well, which is true when only inputDataNodes && outputDataNodes are set (I use it in pattern to substitute matching)

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,144 @@
...
+    performSubstitution(gm, match1, match2);
+    return true;
+}
+}  // anonymous namespace
+
+void checkTransformations(ade::passes::PassContext&,  // FIXME: context is unused here
+    const gapi::GKernelPackage& transformations,
+    std::vector<std::unique_ptr<ade::Graph>>& patterns)
+{
+    const auto& transforms = transformations.get_transformations();
```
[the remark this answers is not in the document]
andrey-golubev: updated

modules/gapi/src/compiler/passes/pattern_matching.cpp
andrey-golubev: found only 1 different place, replaced there as well

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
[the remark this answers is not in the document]
dmatveev: >  Indeed it is simpler than finding a pattern in a main graph: we just need to match inputs && outputs 

Why do you think so?

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,106 @@
...
+        if (!compareDataNodes(pattern.metadata(pNode), substitute.metadata(sNode))) {
+            return {};
+        }
+        matched.insert({ pNode, sNode });
+    }
+    return matched;
+}
+}  // anonymous namespace
+
+SubgraphMatch matchPatternToSubstitute(const cv::gimpl::GModel::Graph& pattern,
```
AsyaPronina: > cv::gimpl::GModel::Graph - > Graph

modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
```diff
@@ -994,4 +998,1012 @@ TEST(PatternMatching, adeSmokeTest)
     EXPECT_EQ(2u, dst->inNodes().size());
 }
 
+// --------------------------------------------------------------------------------------
```
AsyaPronina: It is bad idea to expand this file with new entities. Please move to another test file

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,106 @@
...
+    const auto& patternDataOutputs = patternP.out_nhs;
+
+    const auto& substituteDataInputs = substituteP.in_nhs;
+    const auto& substituteDataOutputs = substituteP.out_nhs;
+
+    // if number of data nodes doesn't match, abort
+    if (patternDataInputs.size() != substituteDataInputs.size()
+        || patternDataOutputs.size() != substituteDataOutputs.size()) {
+        return {};
+    }
```
AsyaPronina: I have an question what do we intend to check here: whether user created a bad substitution or not?
I suppose there shall be check only on unique data nodes, because  Protocol::in_nhs can contain the copies of one node if this node is consumed more than once.
Checking only unique input data nodes can solve the next problems:

Pattern                   Substitute
   +--+  +--+          +--+
   |      |  |      |          |       |
   +--+  +--+          +--+
     |          |                 | |
     v         v                vv
   XXXX   XXXX         XXXX
  XX  XX XX  XX       XX  XX
   XXXX   XXXX         XXXX
      +         +             + +->---+
      v          v              v       |     |
   +--+   +--+         +--+   +--+
   |      |   |      |         |      |
   +--+   +--+         +--+

These two will  be matched but have not to.


Pattern                     Substitute
          +--+                    +--+
 +----+    +----+            |     |
 |        +--+        |           +-++
 |                        |               |
 |                        |               v
 |        XXXX       |            XXXX
+--> XX  XX<--+          XX  XX
          XXXX                    XXXX
             +                          +
             v                           v
          +-++                    +-++
           |     |                      |     |
          +--+                     +--+

These have to be matched but will not.

AsyaPronina: As I got Protocol::in_nhs may contain copies only if we pass copies to the cv::GIn as input API, could ypu confirm this?

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,106 @@
...
+
```
AsyaPronina: And also if check on size is the check for validity of user-defined substitution graph for this pattern, so, I have no idea how can we reject this pair of graphs with existent logic:
image

I also have no idea how to check does user intent to substitute given pattern this way or make a mistake.
I mean I am not sure if there any sense to check user-defined substitution graph validity, may be we can do things simpler and always assume that substitution graph is valid. So check on sizes may become assert statement as extra check for sure.

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // for each pattern input we must find a corresponding substitute input
+    auto matchedDataInputs = matchDataNodes(pattern, substitute, patternDataInputs,
+        substituteDataInputs);
+    // if nothing found, abort
+    if (matchedDataInputs.empty()) {
```
AsyaPronina: assert

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,106 @@
...
+        return {};
```
AsyaPronina: assert

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+    auto matchedDataOutputs = matchDataNodes(pattern, substitute, patternDataOutputs,
+        substituteDataOutputs);
+    // if nothing found, abort
+    if (matchedDataOutputs.empty()) {
+        return {};
```
AsyaPronina: assert

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,147 @@
...
+    auto match1 = findMatches(*patternG, gm);
+    if (!match1.ok()) {
+        return false;
+    }
+
+    // 2. build substitute graph inside the main graph
+    cv::gimpl::GModelBuilder builder(main);
+    const auto& proto_slots = builder.put(substitute.priv().m_ins, substitute.priv().m_outs);
+    Protocol p;
+    std::tie(p.inputs, p.outputs, p.in_nhs, p.out_nhs) = proto_slots;
```
AsyaPronina: Why not just add Protocol to metadata and reduce number of arguments in matchPatternToSubstitute?

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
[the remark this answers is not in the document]
AsyaPronina: Matching which is performed only for IN/OUT data nodes is enough because substitution logic here takes IN/OUT nodes from matched subgraph of the test graph and replace all other stuff of this matched subgraph with the substituton content.
I assume we shall not take care of substitution's content, it is on the user ownership.
If so, then we may say that such logic is kind of robust.

modules/gapi/src/compiler/passes/transformations.cpp
andrey-golubev: Because we build substitute __inside__ the main graph here. If we change its metadata we'll lose original protocol (we can remember it for a while, though) of the original big graph (which can be different, hence pattern is found in the middle). So not to mess with the metadata too much, it's seemed easier to just pass protocol as a  __distinct__ argument

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include "pattern_matching.hpp"
+
+#include <ade/util/zip_range.hpp>
+
+namespace cv { namespace gimpl {
+
+namespace {
+using Graph = GModel::Graph;
+using Metadata = typename Graph::CMetadataT;
+using VisitedMatchings = std::list<std::pair<ade::NodeHandle, ade::NodeHandle>>;
```
AsyaPronina: Extra:
using VisitedMatchings = std::list<std::pair<ade::NodeHandle, ade::NodeHandle>>;

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // compare shape
+    if (firstData.shape != secondData.shape) {
+        return false;
+    }
+    // compare storage
+    if (firstData.storage != secondData.storage) {
+        return false;
+    }
+
+    // FIXME: is it enough to only check shape && storage?
```
AsyaPronina: think yes

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                const Graph& substitute,
+                                const std::vector<ade::NodeHandle>& patternNodes,
+                                std::vector<ade::NodeHandle> substituteNodes)
+{
+    SubgraphMatch::M matched;
+    for (auto it : ade::util::zip(patternNodes, substituteNodes)) {
+        const auto& pNode = std::get<0>(it);
+        const auto& sNode = std::get<1>(it);
+        if (!compareDataNodes(pattern.metadata(pNode), substitute.metadata(sNode))) {
+            return {};
```
AsyaPronina: May be also assert?

modules/gapi/src/compiler/passes/transformations.cpp
AsyaPronina: Good

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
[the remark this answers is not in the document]
andrey-golubev: it's not quite robust to me, because in this setup we rely heavily on the user and if something incorrect is passed there, we'll get the error only on `GComputation::compile()` or worse - on `.apply()`. Ideally it'd be good to detect some semantic differences early on but I'm unsure how, so decided to leave it be like this for now

modules/gapi/src/compiler/passes/perform_substitution.cpp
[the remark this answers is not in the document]
andrey-golubev: renamed

modules/gapi/src/compiler/passes/perform_substitution.cpp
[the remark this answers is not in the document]
andrey-golubev: changed

modules/gapi/src/compiler/passes/transformations.cpp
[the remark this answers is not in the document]
andrey-golubev: changed this part to make clearer

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
andrey-golubev: As a matter of fact, I use assert later on for such stuff. I see your point (I think) and I tend to agree, As to "does user intend to substitute given pattern this way or makes a mistake" -> I guess, the best we can do now is just imagine that the user is __always right__ (let the compilation stage e.g. `inferMeta` to verify that)

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
andrey-golubev: I wonder if it's better to return and empty match at this level and decide what to do with this in the caller's scope: e.g. assert or throw an exception with a proper explanation. I think, in theory, we can benefit from having this function returning `{}` instead of always asserting. `return {}` gives us more freedom for how to use the result

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
andrey-golubev: To me it seems like:
1st pair of graphs: won't be matched - diff. number of ins (which makes sense)
2nd pair of graphs: will be matched if shape/storage are the same

I think that Protocol::in_nhs relates to the number of nodes passed to `GIn()`, so if it's the same node - fine, there will be as many as there are `GIn`s

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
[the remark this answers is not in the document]
andrey-golubev: IMHO, it is good enough for now to leave the verification of the inserted substitute to the `inferMeta` and related passes (that should fail if there are any op/data inconsistencies, etc.)

andrey-golubev:
@AsyaPronina addressed your feedback, please review

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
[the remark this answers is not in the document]
AsyaPronina: Haven't got last idea

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    // 3. match pattern ins/outs to substitute ins/outs
+    auto match2 = matchPatternToSubstitute(*patternG, main,
+        GModel::Graph(*patternG).metadata().get<Protocol>(), p);
+
+    // in theory, we should always be able to find a match here. if not, we're in half-completed
+    // state where some transformations are already applied - what can we do to handle the situation
+    // better?
```
AsyaPronina: This may be done before 2nd step as they are independent

modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
```diff
@@ -9,6 +9,10 @@
 
 #include <stdexcept>
 
+#include <opencv2/gapi/gtransform.hpp>
```
AsyaPronina: need to be removed

modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
```diff
@@ -9,6 +9,10 @@
...
+#include <opencv2/gapi/cpu/core.hpp>
```
AsyaPronina: and this

modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
```diff
@@ -9,6 +9,10 @@
...
+#include <opencv2/gapi/cpu/imgproc.hpp>
```
AsyaPronina: and this

modules/gapi/test/internal/gapi_int_match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,513 @@
+// This file is part of OpenCV project.
```
AsyaPronina: gapi_int_match_pattern_to_substitute.cpp -> gapi_int_match_pattern_to_substitute_test.cpp

modules/gapi/test/internal/gapi_int_perform_substitution.cpp
```diff
@@ -0,0 +1,593 @@
+// This file is part of OpenCV project.
```
AsyaPronina: gapi_int_perform_substitution.cpp -> gapi_int_perform_substitution_test.cpp

modules/gapi/src/compiler/passes/transformations.cpp
andrey-golubev: In general, I think assert is enough here because it throws an exception (which likely mean that current graph will be destroyed). So in the most common scenario, there's no problem if some transformations are applied already: an exception throw means we need to gracefully close the application or at most start anew from the beginning

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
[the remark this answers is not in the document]
andrey-golubev: the idea is: if user messed up the substitute (and instead of e.g. resize we have a matmul or something -> so the complete operation changed) then later steps like `inferMeta` will hopefully fail when there's a difference: we have orig graph, then invalid substitute, then orig graph again -> the invalid substitute triggers errors in steps when data descs do not match, etc. - so in theory, if we did the substitution to an "invalid" substitute then compilation procedure should be able to see it and report an error (though, very opaque). So we have a "last level" defence when doing `compile()`

modules/gapi/test/internal/gapi_int_pattern_matching_test.cpp
andrey-golubev: good job catching this, I forgot to checkout this file to master version - now it should be good

modules/gapi/test/internal/gapi_int_match_pattern_to_substitute.cpp
andrey-golubev: renamed

modules/gapi/test/internal/gapi_int_perform_substitution.cpp
andrey-golubev: renamed

andrey-golubev:
@AsyaPronina  addressed you last review comments

modules/gapi/src/compiler/passes/pattern_matching.hpp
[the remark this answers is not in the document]
andrey-golubev: removed `empty()` completely as well as `partialOk()`. now this should be less confusing

andrey-golubev:
@dmatveev please approve or at least verify that your comments are resolved

@smirnov-alexey please also review & resolve

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    // 2. build substitute graph inside the main graph
+    cv::gimpl::GModelBuilder builder(main);
+    const auto& proto_slots = builder.put(substitute.priv().m_ins, substitute.priv().m_outs);
+    Protocol p;
+    std::tie(p.inputs, p.outputs, p.in_nhs, p.out_nhs) = proto_slots;
+
+    // 3. match pattern ins/outs to substitute ins/outs
+    auto match2 = matchPatternToSubstitute(*patternG, main,
+        GModel::Graph(*patternG).metadata().get<Protocol>(), p);
```
dmatveev: I don't get it. Why do you search for something while you have your matched data already?

Are you sure you will get exactly those nodes which are matched?

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,159 @@
...
+        // iterate through every transformation and try to transform graph parts
+        for (auto it : ade::util::zip(ade::util::toRange(transforms), ade::util::toRange(patterns)))
+        {
+            const auto& t = std::get<0>(it);
+            auto& p = std::get<1>(it);  // Note: using pre-created graphs
+            GAPI_Assert(nullptr != p);
+
+            // if at least one transformation happend, it is possible that other transformations
+            // will be applied in the next "round"
+            canTransform |= tryToSubstitute(ctx.graph, p, t.substitute());
```
dmatveev: if substitution was successful here, this `for` loop continues while it should restart

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Returns matched pairs of {pattern node, substitute node}
+SubgraphMatch::M matchDataNodes(const Graph& pattern,
+                                const Graph& substitute,
+                                const std::vector<ade::NodeHandle>& patternNodes,
+                                std::vector<ade::NodeHandle> substituteNodes)
+{
+    SubgraphMatch::M matched;
+    for (auto it : ade::util::zip(patternNodes, substituteNodes)) {
```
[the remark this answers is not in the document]
andrey-golubev: yes

modules/gapi/src/compiler/passes/pattern_matching.cpp
```diff
@@ -171,14 +174,20 @@ std::size_t labelOf (const ade::NodeHandle& node, // reader node
...
-    // FIXME: Because of new changes which introduce existence of unused DATA nodes
-    // Try to rely on the nh Data::Storage::OUTPUT
-    return nh->outEdges().empty();
+// * OUTPUT must be explicitly checked via storage due to __unused__ pseudo-out nodes in graph
+inline bool IS_ENDPOINT(const cv::gimpl::GModel::Graph& graph, const ade::NodeHandle& nh) {
+    bool is_endpoint = nh->outEdges().empty();
+    const auto& meta = graph.metadata(nh);
+    if (meta.get<cv::gimpl::NodeType>().t == cv::gimpl::NodeType::DATA) {
+        is_endpoint &= (meta.get<cv::gimpl::Data>().storage == cv::gimpl::Data::Storage::OUTPUT);
+    }
```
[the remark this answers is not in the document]
andrey-golubev: oh, that's a good question! I'm not sure that it's even possible but if yes, well, I wasn't prepared for this :smile:
(a) && (b) give us __false__ together. But it seems correct that ENDPOINT, so the proper logic is to __always__ just check the storage type value (I wonder: can something similar be done for input nodes as well?):
1. is a DATA node
2. has `Storage::OUTPUT`

@AsyaPronina what was your original idea with `IS_ENDPOINT` and `outEdges().empty()` then?? it seems like current `master` code won't work with @dmatveev 's idea

modules/gapi/src/compiler/passes/match_pattern_to_substitute.cpp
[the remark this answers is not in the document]
andrey-golubev: no, this check is done separately, but I think you already saw that

modules/gapi/src/compiler/passes/transformations.cpp
andrey-golubev: Oh, found the transaction API. It seems like it's better to introduce it separately (and may be in a generic fashion): this PR is already big enough since we can't merge it for almost a month

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -133,6 +132,12 @@ cv::gimpl::GCompiler::GCompiler(const cv::GComputation &c,
 
     m_e.addPassStage("init");
     m_e.addPass("init", "check_cycles",  ade::passes::CheckCycles());
+    m_e.addPass("init", "check_transformations",
+        std::bind(passes::checkTransformations, _1, std::cref(m_all_kernels),
```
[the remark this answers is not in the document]
andrey-golubev: your approach sounds better, will change to that

andrey-golubev:
@dmatveev I believe your major comments are addressed

modules/gapi/src/compiler/passes/pattern_matching.cpp
```diff
@@ -54,8 +54,14 @@ bool compareDataNodes(const ade::NodeHandle& first, const std::vector<std::size_
...
+    const auto& firstData = firstMeta.get<cv::gimpl::Data>();
+    const auto& secondData = secondMeta.get<cv::gimpl::Data>();
+
+    if (firstData.shape != secondData.shape) {
+        return false;
+    }
+
+    if (firstData.storage != secondData.storage) {
         return false;
     }
```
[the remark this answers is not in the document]
AsyaPronina: Yes, may be..
I didn't realize your concern previous time
This shall fix bug with UNUSED data nodes, as comment below may show, but it seems we shall choose another way to fix it

modules/gapi/src/compiler/passes/pattern_matching.cpp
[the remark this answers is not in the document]
AsyaPronina: Hi Dmitry, thanks for the pointing this out! Yes, it is a problem, because pattern matching algorithm hopes that output protocol nodes have no output edges.
Hi Andrey, thank you for the comment! Yes, you are right current master is not prepared for this:) 
WIll think about possible options

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,135 @@
...
+    bool canTransform = true;
+    while (canTransform)
+    {
+        canTransform = false;
+
+        // iterate through every transformation and try to transform graph parts
+        for (auto it : ade::util::zip(ade::util::toRange(transforms), ade::util::toRange(patterns)))
+        {
+            const auto& t = std::get<0>(it);
+            auto& substituteP = std::get<1>(it);  // Note: using pre-created graphs
```
AsyaPronina: what does mean P here?

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,135 @@
...
+            GAPI_Assert(nullptr != substituteP);
+
+            // if at least one transformation happend, it is possible that other transformations
+            // will be applied in the next "round"
```
AsyaPronina: sounds not very easily understandable, suggest to re-phrase..
You may also add a point that you can apply 1 transformation as much times as it is possible in the graph w/o recursion to the description

modules/gapi/src/compiler/passes/transformations.cpp
andrey-golubev: oh, this should be called completely differently, thanks for catching!

modules/gapi/src/compiler/passes/transformations.cpp
andrey-golubev: tried to re-phrase

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -21,33 +21,55 @@
...
-// Generates ADE graph from expression-based computation
-std::unique_ptr<ade::Graph> makeGraph(const cv::GComputation& c)
+using Graph = GModel::Graph;
+using Metadata = typename Graph::CMetadataT;
+
+// Checks pairs of {pattern node, substitute node} and asserts if there are any incompatibilities
+void checkDataNodes(const Graph& pattern,
+                    const Graph& substitute,
+                    const std::vector<ade::NodeHandle>& patternNodes,
+                    std::vector<ade::NodeHandle> substituteNodes)
```
dmatveev: the last parameter should be `const &` as well?

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -102,6 +103,73 @@ namespace
...
+        for (size_t i = 0; i < size; ++i) {
+            auto& p = patterns[i];
+            for (size_t j = i; j < size; ++j) {
+                auto& s = substitutes[j];
+
+                auto matchInSubstitute = cv::gimpl::findMatches(*p, *s);
+                if (!empty(matchInSubstitute)) {
+                    std::stringstream ss;
+                    ss << "Error: pattern (from transformation #" << i << ") detected inside "
+                          "substitute (from transformation #" << j << ")";
```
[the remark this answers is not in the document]
andrey-golubev: ah, true. I forgot that there are names that I can use. Thanks!

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -102,6 +103,73 @@ namespace
...
+        // Store Computation's protocol in metadata
+        cv::gimpl::Protocol p;
+        std::tie(p.inputs, p.outputs, p.in_nhs, p.out_nhs) = proto_slots;
+        gm.metadata().set(p);
+
+        return pG;
+    }
+
+    void checkTransformations(const cv::gapi::GKernelPackage& pkg,
+                              std::vector<std::unique_ptr<ade::Graph>>& patterns) {
```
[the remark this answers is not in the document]
andrey-golubev: I think it'll be better designed if we pre-allocate, frankly. then in this part we could just check patterns against substitutes

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -102,6 +103,73 @@ namespace
...
+        const auto empty = [] (const cv::gimpl::SubgraphMatch& m) {
+            return m.inputDataNodes.empty() && m.startOpNodes.empty()
+                && m.finishOpNodes.empty() && m.outputDataNodes.empty()
+                && m.inputTestDataNodes.empty() && m.outputTestDataNodes.empty();
+        };
+        // 2. verify there are no patterns in substitutes
+        for (size_t i = 0; i < size; ++i) {
+            auto& p = patterns[i];
+            for (size_t j = i; j < size; ++j) {
+                auto& s = substitutes[j];
```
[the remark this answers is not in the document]
andrey-golubev: hmm, your argument is valid. I remember having similar idea but then decided not to change it. I wonder why... okay, if I won't come up with a decent answer to you and myself, let's check for `j = 0` && `j != i`

modules/gapi/src/compiler/passes/transformations.cpp
andrey-golubev: indeed. thanks!

modules/gapi/src/compiler/passes/transformations.cpp
dmatveev: I agree

andrey-golubev:
@dmatveev addressed you comments

Other tests became irrelevant: the logic that was tested (match pattern to substitute) is now part of substitution routine: which just assumes user is always correct and simply `ade::zip`s node handles from pattern with node handles from substitute (with some basic asserts). The only "exposed" part left of this logic is basically the `GAPI_TRANSFORM` macro itself which pretty much handles the same issues via the API. My initial idea was that we might want to extend the matching of pattern to substitute to something much more meaningful - that's where the tests might've helped.

modules/gapi/src/compiler/gcompiler.cpp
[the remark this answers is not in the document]
andrey-golubev: changed this part but I feel like this logic requires much more attention. @AsyaPronina please take a look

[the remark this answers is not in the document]
AsyaPronina: Hi, @andrey-golubev , @dmatveev!
I couldn't understand: why we prohibit one pattern to be in a substitute  of another pattern if this doesn't produce recursion?
The previous logic just allows to pattern from the right be in substitutes on the left, which doesn't produce recursion but, unfortunately, this logic also reduces some  possible cases where recursion couldn't happen and pattern from the left may be in substitute form the right.
 
How can we handle this case?

My suggestion is the following:
If we rely on the previous logic - it will allow all cases with pattern from right to be in substitute from left to be happened w/o recursion.
To handle case when pattern from the left is in substitute from the right we shall try to swap these two transformations to fit in our logic. If we can't swap that means that existence of left pattern in right substitute leads to recursion and we shall fail our check.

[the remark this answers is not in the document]
AsyaPronina: Hi, @andrey-golubev , @dmatveev!
I couldn't understand: why we prohibit one pattern to be in a substitute  of another pattern if this doesn't produce recursion?
The previous logic just allows to pattern from the right be in substitutes on the left, which doesn't produce recursion but, unfortunately, this logic also reduces some  possible cases where recursion couldn't happen and pattern from the left may be in substitute from the right.
 
How can we handle this case?

My suggestion is the following:
If we rely on the previous logic - it will allow all cases with pattern from right to be in substitute from left to be happened w/o recursion.
To handle case when pattern from the left is in substitute from the right we shall try to swap these two transformations to fit in our logic. If we can't swap that means that existence of left pattern in right substitute leads to recursion and we shall fail our check.

modules/gapi/test/internal/gapi_int_perform_substitution_test.cpp
```diff
@@ -0,0 +1,646 @@
...
+    ASSERT_NE(listener.counts.cend(), listener.counts.find(MyInterleavedResize::id()));
+    ASSERT_EQ(1u, listener.counts.at(MyInterleavedResize::id()));
+}
+
+TEST(PatternMatchingIntegrationBasic, OneNV12toBGRTransformationApplied)
+{
+    cv::Size in_sz(640, 480);
+    cv::Mat y(in_sz, CV_8UC1), uv(cv::Size(in_sz.width / 2, in_sz.height / 2), CV_8UC2);
+    cv::randu(y, cv::Scalar::all(0), cv::Scalar::all(100));
+    cv::randu(y, cv::Scalar::all(100), cv::Scalar::all(200));
```
AsyaPronina: uv?

andrey-golubev: Good catch! Fixed

modules/gapi/test/internal/gapi_int_perform_substitution_test.cpp
```diff
@@ -0,0 +1,646 @@
...
+    // Compare
+    ASSERT_TRUE(AbsExact()(orig_graph_output, transformed_graph_output));
+
+    // Custom verification via listener
+    ASSERT_EQ(2u, listener.counts.size());
+    // called in transformed graph:
+    ASSERT_NE(listener.counts.cend(), listener.counts.find(MyNV12toBGR::id()));
+    ASSERT_EQ(1u, listener.counts.at(MyNV12toBGR::id()));
+    ASSERT_NE(listener.counts.cend(), listener.counts.find(MyInterleavedResize::id()));
+    ASSERT_EQ(1u, listener.counts.at(MyInterleavedResize::id()));
```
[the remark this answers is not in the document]
andrey-golubev: Added FIXME to the part where "listener" is declared

modules/gapi/test/internal/gapi_int_perform_substitution_test.cpp
```diff
@@ -0,0 +1,646 @@
...
+        GMat bgr = cv::gapi::NV12toBGR(in1, in2);
+        GMat b, g, r;
+        std::tie(b, g, r) = cv::gapi::split3(bgr);
+        const auto resize = std::bind(&cv::gapi::resize, std::placeholders::_1,
+            cv::Size(100, 100), 0, 0, cv::INTER_AREA);
+        GMat tmp1 = cv::gapi::resize(bgr, cv::Size(90, 90));
+        GMat tmp2 = cv::gapi::bitwise_not(cv::gapi::merge3(resize(b), resize(g), resize(r)));
+        GMat out = cv::gapi::add(
+            cv::gapi::resize(cv::gapi::addC(tmp1, GScalar(10.0)), cv::Size(100, 100)),
+            tmp2);
```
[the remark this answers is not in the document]
andrey-golubev: thanks, improved!

modules/gapi/test/internal/gapi_int_perform_substitution_test.cpp
```diff
@@ -0,0 +1,646 @@
...
+
+    // Compare
+    ASSERT_TRUE(AbsExact()(orig_graph_output, transformed_graph_output));
+
+    // Custom verification via listener
+    ASSERT_EQ(1u, listener.counts.size());
+    // called in original graph:
+    ASSERT_NE(listener.counts.cend(), listener.counts.find(MyNV12toBGR::id()));
+    ASSERT_EQ(1u, listener.counts.at(MyNV12toBGR::id()));
+}
```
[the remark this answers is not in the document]
andrey-golubev: this test wasn't good enough, frankly. I used `MyNV12ToBGR` in __original__ graph && replaced it with `cv::gapi::NV12toBGR` via __transformation__.

Re-wrote this test so it's not confusing anymore. Now, My* kernels are always used in transformed graphs throughout the tests (hope I didn't miss some)

andrey-golubev:
@AsyaPronina updated as we discussed, please take a look

modules/gapi/src/compiler/passes/transformations.cpp
```diff
@@ -0,0 +1,140 @@
...
+                          const std::vector<std::unique_ptr<ade::Graph>>& patterns)
+{
+    const auto& transforms = pkg.get_transformations();
+    const auto size = transforms.size();
+    if (0u == size) return;
+    // Note: patterns are already generated at this point
+    GAPI_Assert(patterns.size() == transforms.size());
+
+    // transform as long as it is possible
+    // Note: check_transformations pass must handle endless loops and such
```
AsyaPronina: Just question: what does "and such" mean?
And check_transformations is not pass anymore..

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -103,6 +104,84 @@ namespace
...
+            auto& s = std::get<2>(it);
+
+            auto pattern_comp = t.pattern();
+            p = makeGraph(pattern_comp.priv().m_ins, pattern_comp.priv().m_outs);
+
+            auto substitute_comp = t.substitute();
+            s = makeGraph(substitute_comp.priv().m_ins, substitute_comp.priv().m_outs);
+        }
+    }
+
```
AsyaPronina: Where is bold TODO here about dynamic recursion check?

andrey-golubev: There's a FIXME above

modules/gapi/src/compiler/passes/transformations.cpp
andrey-golubev: True, will delete this part

modules/gapi/src/compiler/gcompiler.cpp
AsyaPronina: Where? There should be a bold to-do about that that recursion check is very simple and should be designed to be dynamic and with full coverage

modules/gapi/src/compiler/passes/transformations.cpp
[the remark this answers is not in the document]
dmatveev: Yes, but our arguments start all at the same column

modules/gapi/src/compiler/passes/pattern_matching.cpp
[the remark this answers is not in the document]
dmatveev: So what did you decide on this? Is it addressed?

modules/gapi/test/internal/gapi_int_perform_substitution_test.cpp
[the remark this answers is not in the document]
dmatveev: Ok, thanks!

modules/gapi/src/compiler/gcompiler.cpp
dmatveev: I see it below

andrey-golubev: @AsyaPronina ah, yes. thought it's a different piece of code. yep, the FIXME is below in `checkTransformations` (since this is the place where we validate the transformations that we have)

andrey-golubev:
@alalek I guess it's the time to merge. Please merge

andrey-golubev:
Thanks!

PR #15305, opened 2019-08-14
FIx misc. source and comment typos

luzpaz:
Found via `codespell -q 3 -S ./3rdparty,./modules -L amin,ang,atleast,dof,endwhile,hist,uint`

PR #15318, opened 2019-08-15
Fix modules/ typos

luzpaz:
Found using `codespell -q 3 -S ./3rdparty -L activ,amin,ang,atleast,childs,dof,endwhile,halfs,hist,iff,nd,od,uint`  
Follow-up to #15305


luzpaz:
CC @alalek

PR #15319, opened 2019-08-16
opencv.js : expose FS

cancerberoSgx:
expose FS in Module.FS so users can use the filesystem. This gives more flexibility and enable C/C++ APIs  that reference files. Also needed for supporting node.js and load images directly from local filesystem. (i'm working on supporting node.js too)

relates #15425

### DIFF
--- a/modules/js/src/helpers.js
+++ b/modules/js/src/helpers.js
@@ -38,6 +38,10 @@
 // the use of this software, even if advised of the possibility of such damage.
 //
 
+if (typeof Module.FS === 'undefined' && typeof FS !== 'undefined') {
+    Module.FS = FS;
+}
+
 Module['imread'] = function(imageSource) {
     var img = null;
     if (typeof imageSource === 'string') {


alalek:
Could you please add some usage example (tutorial code) how to use this patch?

cancerberoSgx:
@alalek I added an example here https://github.com/opencv/opencv/pull/15425/files#diff-73a1e2abb8562d484104b42b05a03fafR306 that makes extensive use of FS

However this not only applies to Node.js - accessing FS makes sense in lots of use cases also in the browser. mostly for loading data files for models. I think there's no tutorial  that explain how to do this - in tutorials that load data files , the JS code that do that is not present in the code snippet so the user needs to go to the html source itself in order to realize how to do it. 

I could modify one of these tutorials html like https://docs.opencv.org/master/d2/d99/tutorial_js_face_detection.html and include the code that loads the data file

Or alternative make a new tutorial about this in particular.

What do you think?

asmorkalov:
@cancerberoSgx As fo me, some dedicated example for FS is better than Face Detection modification.

asmorkalov:
@cancerberoSgx Just a friendly reminder. Do you plan to work on example in scope of the PR?

asmorkalov:
:+1:

alalek:

So, please:

PR #15338, opened 2019-08-19
QR-Code detector : multiple detection

rayonnant14:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/684


modules/objdetect/src/qrcode.cpp
```diff
@@ -216,264 +314,370 @@ vector<Point2f> QRDetect::separateVerticalLines(const vector<Vec3d> &list_lines)
...
-        vector<Point2f> list_area_pnt;
-        list_area_pnt.push_back(current_point);
-
-        vector<LineIterator> list_line_iter;
-        list_line_iter.push_back(LineIterator(bin_barcode, current_point, left_point));
-        list_line_iter.push_back(LineIterator(bin_barcode, current_point, central_point));
-        list_line_iter.push_back(LineIterator(bin_barcode, current_point, right_point));
 
-        for (size_t k = 0; k < list_line_iter.size(); k++)
+bool QRDetect::CheckPoints(Mat scr, vector<Point2f> triangle_points)
```
allnes: pls call function in the same style (checkPoints, not CheckPoints)

modules/objdetect/src/qrcode.cpp
```diff
@@ -216,264 +314,370 @@ vector<Point2f> QRDetect::separateVerticalLines(const vector<Vec3d> &list_lines)
...
+{
+      int count_b = 0;
+      int count_w = 0;
+      float x0 = triangle_points[0].x;
```
allnes: pls use `cv::Point`

modules/objdetect/src/qrcode.cpp
```diff
@@ -216,264 +314,370 @@ vector<Point2f> QRDetect::separateVerticalLines(const vector<Vec3d> &list_lines)
...
+   vector<vector<Point2f>> triangles;
+   size_t temp_num_points = 0;
+   bin_barcode_temp = bin_barcode;
+   struct {
+        bool operator()(Point2f a, Point2f b) const
         {
-            if (norm(localization_points[i] - localization_points[j]) < 10)
+            return ((sqrt(a.x * a.x + a.y * a.y)) < (sqrt(b.x * b.x + b.y * b.y)));
+        }
+    } CompareDistance;
```
allnes: same style

modules/objdetect/src/qrcode.cpp
```diff
@@ -216,264 +314,370 @@ vector<Point2f> QRDetect::separateVerticalLines(const vector<Vec3d> &list_lines)
...
-        Mat mask = Mat::zeros(bin_barcode.rows + 2, bin_barcode.cols + 2, CV_8UC1);
-        uint8_t next_pixel, future_pixel = 255;
-        int count_test_lines = 0, index = cvRound(localization_points[i].x);
-        for (; index < bin_barcode.cols - 1; index++)
-        {
-            next_pixel = bin_barcode.ptr<uint8_t>(cvRound(localization_points[i].y))[index + 1];
-            if (next_pixel == future_pixel)
+    CV_TRACE_FUNCTION();
+    size_t c = localization_points.size() - 1;
+      if (localization_points[c].size() != 3)
```
allnes: not on the same line as the line above

modules/objdetect/src/qrcode.cpp
```diff
@@ -762,6 +966,21 @@ inline double QRDetect::getCosVectors(Point2f a, Point2f b, Point2f c)
     return ((a - b).x * (c - b).x + (a - b).y * (c - b).y) / (norm(a - b) * norm(c - b));
 }
 
+bool QRDetect::NextSet(int *set, int n, int m)
```
allnes: same style

modules/objdetect/test/test_qrcode.cpp
```diff
@@ -119,19 +127,91 @@ INSTANTIATE_TEST_CASE_P(/**/, Objdetect_QRCode, testing::ValuesIn(qrcode_images_
...
+            << "Not found results for '" << name_current_image
+            << "' image in config file:" << dataset_config << std::endl
+            << "Re-run tests with enabled UPDATE_QRCODE_TEST_DATA macro to update test data."
+            << std::endl;
+    }
+}
+
+INSTANTIATE_TEST_CASE_P(/**/, Objdetect_Multiple_QRCode, testing::ValuesIn(mult_qrcode_images_name));
 
+//INSTANTIATE_TEST_CASE_P(/**/,Objdetect_Multiple_QRCode, testing::ValuesIn(mult_qrcodes_image_name));
```
allnes: pls remove comments

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -700,16 +700,15 @@ class CV_EXPORTS_W QRCodeDetector
      @param points Quadrangle vertices found by detect() method (or some other algorithm).
      @param straight_qrcode The optional output image containing rectified and binarized QR code
      */
-    CV_WRAP cv::String decode(InputArray img, InputArray points, OutputArray straight_qrcode = noArray());
 
+    CV_WRAP std::vector<std::string> decode(InputArray img, InputArrayOfArrays points, OutputArrayOfArrays straight_qrcode);
```
dkurt: Do we have python test for QR codes? Need to check that `std::vector<std::string>` works fine. Otherwise need to replace to `std::vector<cv::String>`

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -700,16 +700,15 @@ class CV_EXPORTS_W QRCodeDetector
...
     /** @brief Both detects and decodes QR code
 
      @param img grayscale or color (BGR) image containing QR code.
      @param points opiotnal output array of vertices of the found QR code quadrangle. Will be empty if not found.
      @param straight_qrcode The optional output image containing rectified and binarized QR code
      */
-    CV_WRAP cv::String detectAndDecode(InputArray img, OutputArray points=noArray(),
-                                        OutputArray straight_qrcode = noArray());
+     CV_WRAP std::vector<std::string> detectAndDecode(InputArray img, OutputArrayOfArrays points, OutputArrayOfArrays straight_qrcode);
```
dkurt: What about default values?

modules/objdetect/src/qrcode.cpp
```diff
@@ -8,6 +8,7 @@
 #include "precomp.hpp"
 #include "opencv2/objdetect.hpp"
 #include "opencv2/calib3d.hpp"
+#include "opencv2/imgproc/types_c.h"
```
dkurt: For which types it is used?

allnes: @rayonnant14, I don't see answer.

modules/objdetect/src/qrcode.cpp
```diff
@@ -27,31 +28,129 @@ class QRDetect
 public:
     void init(const Mat& src, double eps_vertical_ = 0.2, double eps_horizontal_ = 0.1);
     bool localization();
-    bool computeTransformationPoints();
+    bool computeTransformationPoints(size_t c);
```
allnes: pls rename variable for more detailed

modules/objdetect/src/qrcode.cpp
rayonnant14: @allnes @dkurt for parameters CV_RETR_TREE and CV_CHAIN_APPROX_SIMPLE in function findContours

dkurt: Please use just `RETR_TREE` and `CHAIN_APPROX_SIMPLE`

asmorkalov:
@rayonnant14, @allnes What is the PR status? Is it still WIP?

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -691,24 +691,25 @@ class CV_EXPORTS_W QRCodeDetector
      @param img grayscale or color (BGR) image containing (or not) QR code.
      @param points Output vector of vertices of the minimum-area quadrangle containing the code.
      */
-    CV_WRAP bool detect(InputArray img, OutputArrayOfArrays points) const;
+    CV_WRAP bool detect(InputArray img, OutputArrayOfArrays& points) const;
```
dkurt: Why & is used?

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -720,15 +721,16 @@ class CV_EXPORTS_W QRCodeDetector
     @param eps_x Epsilon neighborhood, which allows you to determine the horizontal pattern of the scheme 1:1:3:1:1 according to QR code standard.
     @param eps_y Epsilon neighborhood, which allows you to determine the vertical pattern of the scheme 1:1:3:1:1 according to QR code standard.
     */
-CV_EXPORTS bool detectQRCode(InputArray in, std::vector<Point> &points, double eps_x = 0.2, double eps_y = 0.1);
+CV_EXPORTS bool detectQRCode(InputArray in, std::vector<cv::Mat> &points, double eps_x = 0.2, double eps_y = 0.1);
```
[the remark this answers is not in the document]
rayonnant14: @dkurt Mat contains 4 points of qr-code bounding box. I use vector of Mat by the reason more than one qr-code on the image

modules/objdetect/include/opencv2/objdetect.hpp
rayonnant14: @dkurt fixed it

modules/objdetect/include/opencv2/objdetect.hpp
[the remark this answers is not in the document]
dkurt: I'd like to recommend to use `std::vector<std::vector<Point> >` instead.

BTW, it's even better to use `OutputArrayOfArrays`

modules/objdetect/src/qrcode.cpp
rayonnant14: @allnes fixed it in new commit

modules/objdetect/include/opencv2/objdetect.hpp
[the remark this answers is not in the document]
alalek: Existing `std::vector<Point>` doesn't limit you on 4 points only. You can push here 4,8,12,etc points.

dkurt:
Please try to apply this patch to resolve Java building error:
```patch
diff --git a/modules/java/generator/gen_java.py b/modules/java/generator/gen_java.py
index d98a98f..ea1b89e 100755
--- a/modules/java/generator/gen_java.py
+++ b/modules/java/generator/gen_java.py
@@ -914,7 +914,10 @@ class JavaWrapperGenerator(object):
                     c_epilogue.append("Mat* _retval_ = new Mat();")
                     c_epilogue.append(fi.ctype+"_to_Mat(_ret_val_vector_, *_retval_);")
                 else:
-                    c_epilogue.append("return " + fi.ctype + "_to_List(env, _ret_val_vector_);")
+                    if ret:
+                        c_epilogue.append("jobject _retval_ = " + fi.ctype + "_to_List(env, _ret_val_vector_);")
+                    else:
+                        c_epilogue.append("return " + fi.ctype + "_to_List(env, _ret_val_vector_);")
             if fi.classname:
                 if not fi.ctype: # c-tor
                     retval = fi.fullClass(isCPP=True) + "* _retval_ = "
```

modules/objdetect/src/qrcode.cpp
```diff
@@ -863,43 +863,1423 @@ struct QRCodeDetector::Impl
...
+      }
+
+      if ((tmp_num_points < 3) && (tmp_num_points >= 1))
+      {
+          const double min_side = std::min(bin_barcode_fullsize.size().width, bin_barcode_fullsize.size().height);
+          if(min_side > 512)
+          {
+            bin_barcode = tmp_shrinking;
+            purpose = SHRINKING;
+            coeff_expansion = min_side / 512.0;
```
allnes: please moved magic numbers in const variable and name its

modules/objdetect/src/qrcode.cpp
```diff
@@ -863,43 +863,1423 @@ struct QRCodeDetector::Impl
...
+     {
+         set_size[i] = int(0.5 * (true_points_group[i].size() - 2 ) * (true_points_group[i].size() - 1));
+     }
+     vector< vector< vector< int> > > all_points(true_points_group.size());
+     for(size_t i = 0; i < true_points_group.size(); i++)
+         all_points[i].resize(set_size[i]);
+     for(size_t i = 0; i < true_points_group.size(); i++)
+        for (int j = 0; j < set_size[i]; j++)
+             all_points[i][j].resize(3);
+     int p = 0;
```
allnes: what's mean variable `p`? please renamed variable, then other people understand it

modules/objdetect/src/qrcode.cpp
```diff
@@ -863,43 +863,1423 @@ struct QRCodeDetector::Impl
...
+          if ((localization_points[j].size() == 3) && (transformation_points[j].size() == 4))
+          {
+              for_copy_loc.push_back(localization_points[j]);
+              for_copy_trans.push_back(transformation_points[j]);
+          }
+      }
+      localization_points = for_copy_loc;
+      transformation_points = for_copy_trans;
+
+      if((num_points - tmp_num_points) == 1) q--;
```
allnes: with variable `q` too

modules/objdetect/src/qrcode.cpp
```diff
@@ -1246,4 +2609,202 @@ cv::String QRCodeDetector::detectAndDecode(InputArray in,
...
+        if(ok)
+        {
+            points_.create(int(points.size()), 1,  points_.fixedType() ? points_.type() : CV_32FC2);
+            for(size_t i = 0; i < points.size(); i++)
+            {
+                points_.create(4, 1, points_.fixedType() ? points_.type() : CV_32FC2, int(i), true);
+                Mat m = points_.getMat(int(i));
+                switch(points_.type())
+                {
+                    case 10:
```
allnes: please named points type

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -692,24 +692,26 @@ class CV_EXPORTS_W QRCodeDetector
...
+    CV_WRAP bool multipleDetect(InputArray img, OutputArrayOfArrays points) const;
 
     /** @brief Decodes QR code in image once it's found by the detect() method.
      Returns UTF8-encoded output string or empty string if the code cannot be decoded.
-
      @param img grayscale or color (BGR) image containing QR code.
      @param points Quadrangle vertices found by detect() method (or some other algorithm).
      @param straight_qrcode The optional output image containing rectified and binarized QR code
      */
     CV_WRAP cv::String decode(InputArray img, InputArray points, OutputArray straight_qrcode = noArray());
```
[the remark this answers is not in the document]
rayonnant14: @allnes  done

modules/objdetect/perf/perf_qrcode_pipeline.cpp
```diff
@@ -53,6 +53,59 @@ PERF_TEST_P_(Perf_Objdetect_QRCode, decode)
...
+{
+    const std::string name_current_image = GetParam();
+    const std::string root = "cv/qrcode/multiple/";
+
+    std::string image_path = findDataFile(root + name_current_image);
+    Mat src = imread(image_path);
+    std::vector<Mat> straight_barcode;
+    ASSERT_FALSE(src.empty()) << "Can't read image: " << image_path;
+
+    std::vector< std::vector< Point2f > > corners;
```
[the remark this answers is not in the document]
rayonnant14: @allnes corrected it

modules/objdetect/src/qrcode.cpp
```diff
@@ -112,7 +112,7 @@ vector<Vec3d> QRDetect::searchHorizontalLines()
         {
             if (bin_barcode_row[x] == future_pixel)
             {
-                future_pixel = 255 - future_pixel;
+                future_pixel = static_cast<int>(~future_pixel);
```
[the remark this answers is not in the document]
rayonnant14: @allnes correct it

modules/objdetect/src/qrcode.cpp
```diff
@@ -863,43 +863,1423 @@ struct QRCodeDetector::Impl
...
+    QRDetect qrdet;
+    qrdet.init(inarr, p->epsX, p->epsY);
+    if (!qrdet.localization()) { return false; }
+    if (!qrdet.computeTransformationPoints()) { return false; }
+    vector<Point2f> pnts2f = qrdet.getTransformationPoints();
+    Mat(pnts2f).convertTo(points, points.fixedType() ? points.type() : CV_32FC2);
+    return true;
+}
+
+bool detectQRCode(InputArray in, vector<Point> &points, double eps_x, double eps_y)
```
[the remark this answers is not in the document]
rayonnant14: @allnes done

modules/objdetect/src/qrcode.cpp
```diff
@@ -863,43 +863,1423 @@ struct QRCodeDetector::Impl
...
+      findContours(lineMask, contours, hierarchy, RETR_TREE, CHAIN_APPROX_SIMPLE);
+      if(contours.size() < 1) return false;
+      int count = 0;
+      for(size_t i = 0; i < all_localization_points.size(); i++)
+      {
+            double dist;
+            if((dist = pointPolygonTest(contours[0], all_localization_points[i], false)) > 0)
+            {
+
+               if((abs(all_localization_points[i].x - p0.x) > 3) && (abs(all_localization_points[i].x - p1.x) > 3)
```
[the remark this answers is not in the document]
rayonnant14: @allness done it

modules/objdetect/src/qrcode.cpp
rayonnant14: @allnes correct it

modules/objdetect/src/qrcode.cpp
rayonnant14: @allness done

allnes:
it looks good :+1:
@alalek @dkurt pls review again.

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -693,23 +693,43 @@ class CV_EXPORTS_W QRCodeDetector
      */
     CV_WRAP bool detect(InputArray img, OutputArray points) const;
 
+    /** @brief Detects QR codes in image and returns the vector of the quadrangle containing the codes.
+     @param img grayscale or color (BGR) image containing (or not) QR codes.
+     @param points Output vector of vector of vertices of the minimum-area quadrangle containing the codes.
+     */
+    CV_WRAP bool multipleDetect(InputArray img, OutputArrayOfArrays points) const;
```
alalek: detectMulti (like `detectMultiScale()` in objdetect)

From API point of view - it make sense to add `maxCount` or `minSize` parameters. Or add them as a class object properties.

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -693,23 +693,43 @@ class CV_EXPORTS_W QRCodeDetector
...
     CV_WRAP cv::String decode(InputArray img, InputArray points, OutputArray straight_qrcode = noArray());
 
-    /** @brief Both detects and decodes QR code
+    /** @brief Decodes QR codes in image once it's found by the detect() method.
+     Returns UTF8-encoded output vector of string or empty vector of string if the codes cannot be decoded.
+     @param img grayscale or color (BGR) image containing QR codes.
+     @param points vector of Quadrangle vertices found by detect() method (or some other algorithm).
+     @param straight_qrcode The optional output vector of images containing rectified and binarized QR codes
+     */
+    CV_WRAP std::vector<cv::String> multipleDecode(InputArray img, InputArrayOfArrays points, OutputArrayOfArrays straight_qrcode = noArray());
```
alalek: `decodeMulti`

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -693,23 +693,43 @@ class CV_EXPORTS_W QRCodeDetector
...
      */
     CV_WRAP cv::String detectAndDecode(InputArray img, OutputArray points=noArray(),
                                         OutputArray straight_qrcode = noArray());
+
+    /** @brief Both detects and decodes QR codes
+    @param img grayscale or color (BGR) image containing QR codes.
+    @param points optional output vector of arrays of vertices of the found QR code quadrangle. Will be empty if not found.
+    @param straight_qrcode The optional output vector of images containing rectified and binarized QR codes
+    */
+    CV_WRAP std::vector<cv::String> multipleDetectAndDecode(InputArray img, OutputArrayOfArrays points = noArray(),
```
alalek: `detectAndDecodeMulti`

with limitation parameters.

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -723,6 +743,15 @@ class CV_EXPORTS_W QRCodeDetector
...
 CV_EXPORTS bool detectQRCode(InputArray in, std::vector<Point> &points, double eps_x = 0.2, double eps_y = 0.1);
 
+/** @brief Detect QR codes in image and return vector of minimum area of quadrangle that describes QR codes.
+    @param in  Matrix of the type CV_8UC1 containing an image where QR codes are detected.
+    @param points Output vector of vector of vertices of a quadrangle of minimal area that describes QR codes.
+    @param eps_x Epsilon neighborhood, which allows you to determine the horizontal pattern of the scheme 1:1:3:1:1 according to QR code standard.
+    @param eps_y Epsilon neighborhood, which allows you to determine the vertical pattern of the scheme 1:1:3:1:1 according to QR code standard.
+    */
+CV_EXPORTS bool multipleDetectQRCode(InputArray in, std::vector< std::vector< Point > > &points,
+                                     double eps_x = 0.2, double eps_y = 0.1);
```
alalek: `detectQRCodeMulti` + parameters

modules/objdetect/src/qrcode.cpp
```diff
@@ -1245,5 +1228,1568 @@ cv::String QRCodeDetector::detectAndDecode(InputArray in,
     return decoded_info;
 }
 
+class  MultipleQRDetect
```
alalek: Why is inheritance not used here?

modules/objdetect/src/qrcode.cpp
```diff
@@ -1245,5 +1228,1568 @@ cv::String QRCodeDetector::detectAndDecode(InputArray in,
...
+  points.create((int)pnts2f.size(), 1, points.fixedType() ? points.type() : CV_32FC2);
+
+  for(int i = 0; i < (int)pnts2f.size(); i++)
+  {
+      points.create(4, 1, points.fixedType() ? points.type() : CV_32FC2, i, true);
+      Mat m = points.getMat(i);
+
+      switch(points.type())
+      {
+          case 10:
```
alalek: 10?

How do you plan to refactor code about OpenCV types?
Magic and hard-coded values must be avoided.

Consider using `.convertTo()`.

rayonnant14:
@alalek fixed
Look again, please

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -693,23 +693,43 @@ class CV_EXPORTS_W QRCodeDetector
...
     CV_WRAP cv::String decode(InputArray img, InputArray points, OutputArray straight_qrcode = noArray());
 
-    /** @brief Both detects and decodes QR code
+    /** @brief Decodes QR codes in image once it's found by the detect() method.
+     Returns UTF8-encoded output vector of string or empty vector of string if the codes cannot be decoded.
+     @param img grayscale or color (BGR) image containing QR codes.
+     @param points vector of Quadrangle vertices found by detect() method (or some other algorithm).
+     @param straight_qrcode The optional output vector of images containing rectified and binarized QR codes
+     */
+    CV_WRAP std::vector<cv::String> decodeMulti(InputArray img, InputArrayOfArrays points, OutputArrayOfArrays straight_qrcode = noArray());
```
alalek: We usually don't return `std::vector` from functions (it is very inefficient in C++98):
```
bool decodeMulti(InputArray img, CV_OUT std::vector<std::string>& decoded_info,
        InputArrayOfArrays points, OutputArrayOfArrays straight_qrcode = noArray());
```

`cv::String` => `std::string`

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -731,6 +760,15 @@ CV_EXPORTS bool detectQRCode(InputArray in, std::vector<Point> &points, double e
...
 CV_EXPORTS bool decodeQRCode(InputArray in, InputArray points, std::string &decoded_info, OutputArray straight_qrcode = noArray());
 
+/** @brief Decode QR codes in image and return text that is encrypted in QR code.
+    @param in  Matrix of the type CV_8UC1 containing an image where QR code are detected.
+    @param points Input vector of vector of vertices of a quadrangle of minimal area that describes QR codes.
+    @param decoded_info vector of String information that is encrypted in QR codes.
+    @param straight_qrcode vector of Matrixes of the type CV_8UC1 containing an binary straight QR codes.
+    */
+CV_EXPORTS bool decodeQRCodeMulti(InputArray in, InputArrayOfArrays points, std::vector<cv::String> &decoded_info,
+                             OutputArrayOfArrays straight_qrcode = noArray());
```
alalek: `cv::String` => `std::string`

modules/objdetect/src/qrcode.cpp
```diff
@@ -1245,5 +1228,1208 @@ cv::String QRCodeDetector::detectAndDecode(InputArray in,
...
+    uint8_t next_pixel;
+    vector<double> test_lines;
+    temp_length = 0;
+
+    for (size_t pnt = 0; pnt < list_lines.size(); pnt++)
+    {
+        const int x = cvRound(list_lines[pnt][0] + list_lines[pnt][2] * 0.5);
+        const int y = cvRound(list_lines[pnt][1]);
+
+        // --------------- Search vertical up-lines --------------- //
```
alalek: There is the the similar code in `QRDetect::separateVerticalLines()`.
What is the difference?
Why is the difference?

I try to describe problem which we should avoid by unifying and reusing the same code for features finding. We must have consistent behavior between `QRDetect` and `QRDetectMulti`:
- `QRDetectMulti` - no detections ==> `QRDetect` - no detection too and vice versa
- `QRDetectMulti` - has detections ==> `QRDetect` - one of these detections (largest)
- `QRDetect` - has detection ==> `QRDetectMulti` - has detections including this one

modules/objdetect/src/qrcode.cpp
```diff
@@ -1245,5 +1228,1208 @@ cv::String QRCodeDetector::detectAndDecode(InputArray in,
...
+               }
+            }
+       }
+       if (count == 0) return true;
+       else return false;
+}
+
+bool QRDetectMulti::compareSquare::operator()(vector<int> a, vector<int> b) const
+{
+  return 0.5 * abs((points[a[1]].x - points[a[0]].x) *
```
alalek: `0.5 * ` is not needed (on both sides of the condition)

modules/objdetect/src/qrcode.cpp
```diff
@@ -1245,5 +1228,1208 @@ cv::String QRCodeDetector::detectAndDecode(InputArray in,
...
+               {
+                  return false;
+               }
+            }
+       }
+       if (count == 0) return true;
+       else return false;
+}
+
+bool QRDetectMulti::compareSquare::operator()(vector<int> a, vector<int> b) const
```
alalek: > `vector<int> a`

Do we really need vector here? I see 3 values only.
vectors and other object must be passed by "const reference" by default.

modules/objdetect/src/qrcode.cpp
```diff
@@ -1245,5 +1228,1208 @@ cv::String QRCodeDetector::detectAndDecode(InputArray in,
...
+  (points[b[2]].y - points[b[0]].y) -
+  (points[b[2]].x - points[b[0]].x) *
+  (points[b[1]].y - points[b[0]].y));
+}
+
+bool QRDetectMulti::localization()
+{
+  CV_TRACE_FUNCTION();
+  vector<Point2f> tmp_localization_points;
+  int num_qrcodes;
```
alalek: Inconsistent indentation with other parts of the file.

modules/objdetect/src/qrcode.cpp
```diff
@@ -1245,5 +1228,1208 @@ cv::String QRCodeDetector::detectAndDecode(InputArray in,
...
+      if(((num_points - tmp_num_points) == 0) && (tmp_localization_points.size() == 0) && (true_points_group.size() == 1))
+      break;
+   }
+   if ((transformation_points.size() == 0) || (localization_points.size() == 0)) return false;
+      return true;
+}
+bool QRDetectMulti::computeTransformationPoints(size_t cur_ind)
+{
+      CV_TRACE_FUNCTION();
+      if (localization_points[cur_ind].size() != 3)
```
alalek: 6 spaces => 4 spaces

samples/cpp/qrcode_multi.cpp
```diff
@@ -0,0 +1,196 @@
+#include "opencv2/objdetect.hpp"
```
alalek: We don't need new sample.

Please update existing one:
- support both modes
- switch between modes by "space" key

modules/objdetect/include/opencv2/objdetect.hpp
rayonnant14: @alalek  std::vector<std::string> doesn't work fine in python and java
Maybe, it is better to use std::vector<cv::String> to avoid problems?

modules/objdetect/src/qrcode.cpp
rayonnant14: @alalek QRDetect::separateVerticalLines() finds the best coeff_epsilon by comparison of compactness, which kmeans returns (supposed that number of clusters is 3(1 qr-code).
If  there is more than one qr-code on image this method doesn't work correctly.
By this reason there is some changes in code in  QRDetectMulti::separateVerticalLines()
Maybe, it should be renamed to avoid the mess

samples/cpp/qrcode.cpp
```diff
@@ -88,42 +93,89 @@ int liveQRCodeDetect(const string& out_file)
...
+            string decode_info;
+            vector<Point> transform;
+            cap >> frame;
+            if (frame.empty())
+            {
+                cout << "End of video stream" << endl;
+                break;
+            }
+            cvtColor(frame, src, COLOR_BGR2GRAY);
+            total.start();
```
alalek: This code should be wrote once. The same note for lines 128-135.

Switch processing/drawing code only.

rayonnant14:
@alalek please, look

modules/objdetect/include/opencv2/objdetect.hpp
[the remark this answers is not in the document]
dkurt: Please keep the origin lines where it's possible: restore removed empty lines in example

modules/objdetect/include/opencv2/objdetect.hpp
dkurt: Do we really need two different epsilons? I mean according to QR codes standard, both horizontal and vertical directions should be equal, aren't they?

rayonnant14:
@alalek @dkurt please, look again

modules/objdetect/src/qrcode.cpp
```diff
@@ -1245,5 +1228,1206 @@ cv::String QRCodeDetector::detectAndDecode(InputArray in,
...
+{
+    CV_TRACE_FUNCTION();
+    double cos_angles[3], norm_triangl[3];
+
+    norm_triangl[0] = norm(local_point[1] - local_point[2]);
+    norm_triangl[1] = norm(local_point[0] - local_point[2]);
+    norm_triangl[2] = norm(local_point[1] - local_point[0]);
+
+    cos_angles[0] = (norm_triangl[1] * norm_triangl[1] + norm_triangl[2] * norm_triangl[2]
+                  -  norm_triangl[0] * norm_triangl[0]) / (2 * norm_triangl[1] * norm_triangl[2]);
```
alalek: You might use dot product instead of law of cosines in case of input vectors:
- https://en.wikipedia.org/wiki/Dot_product

modules/objdetect/src/qrcode.cpp
```diff
@@ -1245,5 +1228,1206 @@ cv::String QRCodeDetector::detectAndDecode(InputArray in,
...
+                   list_lines_y = separateVerticalLines(list_lines_x);
+                   if( list_lines_y.size() < 3 ) break;
+               }
+               else break;
+            }
+            vector<int> index_list_lines_y;
+            for(size_t i = 0; i < list_lines_y.size(); i++)
+                index_list_lines_y.push_back(-1);
+            num_points = 0;
+            for(size_t i = 0; i < list_lines_y.size() - 1; i++)
```
alalek: Consider adding high-level non-obvious comments into your code.
What and why are these loops doing here? What are main stages of this algorithm?

This method is ~400 lines. It definitely should be divided into well named (and well tested) parts/stages.

alalek:
Again, drawing functions should be avoided.

1.
`checkPointsInsideQuadrangle()` and `checkPointsInsideTriangle()` should not "draw" anything.
Take a look on `pointPolygonTest()` function.

2.
The similar note is about `checkPoints()`. Intensity check can be related for "bounding rectangle".

3.
another one `fillConvexPoly()` call in `findQRCodeContours()`.

4.
need to revise code using of `floodFill()` / `LineIterator()` functionality.

modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -705,11 +705,36 @@ class CV_EXPORTS_W QRCodeDetector
...
+     @param points optional output array of vertices of the found QR code quadrangle. Will be empty if not found.
      @param straight_qrcode The optional output image containing rectified and binarized QR code
      */
     CV_WRAP cv::String detectAndDecode(InputArray img, OutputArray points=noArray(),
                                         OutputArray straight_qrcode = noArray());
+    /** @brief Detects QR codes in image and returns the vector of the quadrangle containing the codes.
+     @param img grayscale or color (BGR) image containing (or not) QR codes.
+     @param points Output vector of vector of vertices of the minimum-area quadrangle containing the codes.
+     */
+    CV_WRAP bool detectMulti(InputArray img, OutputArrayOfArrays points) const;
```
alalek: > OutputArrayOfArrays
> InputArrayOfArrays
> std::vector< std::vector< Point > > corners;

Please avoid using of `*OfArrays` types and/or nested `std::vector`.
This is unnecessary for fixed size elements (4 points per QR code).

Lets store `points` as one "array" (without nesting):
- `std::vector< Point >` where each 4 points define QR code instance (`<0, 1, 2, 3>, <4, 5, 6 ,7> ...`).
- `Mat` - CV_32FC2 with one or multiple rows.
and pass them as `OutputArray`.

This is continued discussion from this comment: https://github.com/opencv/opencv/pull/15338#discussion_r339004770

samples/cpp/qrcode.cpp
```diff
@@ -89,41 +98,79 @@ int liveQRCodeDetect(const string& out_file)
...
-        vector<Point> transform;
+        Mat frame, src;
         cap >> frame;
         if (frame.empty())
         {
             cout << "End of video stream" << endl;
             break;
         }
+        if (!out_origin_file.empty())
+            imwrite(out_origin_file, frame); // write original frame
```
alalek: Saving each frame slowdowns demo significantly. Image I/O is not optimized (because it is for debug purposes here).

Original and processed frames should be saved by pressing key only (it was space ' '). Both at once.

modules/objdetect/misc/python/test/test_qrcode_detect.py
```diff
@@ -13,6 +13,21 @@
 class qrcode_detector_test(NewOpenCVTests):
     def test_detect_and_decode(self):
         img = cv.imread(os.path.join(self.extraTestDataPath, 'cv/qrcode/link_ocv.jpg'))
+        self.assertFalse(img == None)
```
alalek: > == None

In Python world checks should look like:
- `a is None`
- `a is not None`

rayonnant14: @alalek Which way is better img is None or np.shape(img) is () ?

PR #15341, opened 2019-08-20
issue #5769 fixed: cv::stereoRectify fails if given inliers mask of t…

### LINKED ISSUE #5769: cv::stereoRectify fails if given inliers mask of type vector<uchar>
The function `cv::recoverPose` fails when given an inliers mask of type `vector<uchar>`, which is a valid input/output array. If an equal `Mat`of type `CV_U8` is given instead, the function works fine. 

This is unfortunate since `cv::MatchesInfo`uses `vector<uchar>` to store the inliers mask.

I assume the error is due to a bogus assertion in the [corresponding method](https://github.com/Itseez/opencv/blob/master/modules/calib3d/src/five-point.cpp#L541):

`CV_Assert(mask.size() == mask1.size());`

`mask1` is always of type `Mat`, `mask` is the given input. In the case described here, `cv::Size` might be compared to a `size_t`, never returning `true`.

DiebBlue:
…ype vector<uchar>

resolves #5769

### This pullrequest changes
**original:** mask1, mask2, mask3, mask4 will be transposed but not the input mask -> so `CV_Assert(mask.size() == mask1.size())` causes an exception.
**fix:** the input mask will also be transposed like the masks 1 to 4, so that the sizes fit again and than it works.
(see also [https://github.com/opencv/opencv/issues/5769])

### DIFF
--- a/modules/calib3d/src/five-point.cpp
+++ b/modules/calib3d/src/five-point.cpp
@@ -570,7 +570,7 @@ int cv::recoverPose( InputArray E, InputArray _points1, InputArray _points2,
     // If _mask is given, then use it to filter outliers.
     if (!_mask.empty())
     {
-        Mat mask = _mask.getMat();
+        Mat mask = _mask.getMat().t();
         CV_Assert(mask.size() == mask1.size());
         bitwise_and(mask, mask1, mask1);
         bitwise_and(mask, mask2, mask2);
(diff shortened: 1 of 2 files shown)


alalek:
Lets continue working on "master" branch. I will rebase onto "3.4" before merge.

alalek:
OK. Please update PR target branch again after force push of local changes into fork (you should see on "commits" tab your commits only).

DiebBlue:
Ok, I work on it. :)

DiebBlue:
How is it going to continue here?

PR #15342, opened 2019-08-20
Fix Doc Feature Matching + Homography to find Objects's Code error

laugh12321:
### Doc Feature Matching + Homography to find Objects's  Code has error

```python
img1 = cv.imread('box.png',0)          # queryImage
....
h,w,d = img1.shape
```

Img1 is a grayscale image with only two color channels， but return three values.
And an error will accur after running.
```
ValueError: not enough values to unpack (expected 3, got 2)
```

Now changed `h, w = img1.shape`

```
force_builders=Docs
```


alalek:
Can't be tested/merged due source branch removal.

Reliable way (works with color images too):
```
h,w = img1.shape[:2]
```

PR #15351, opened 2019-08-20
Fix the detection of XIMEA on Windows (when it has been installed by …

philippefoubert:
Fix the detection of XIMEA on Windows (when it has been installed by another user with administrative privileges, for example).


modules/videoio/cmake/detect_ximea.cmake
```diff
@@ -1,6 +1,6 @@
 if(NOT HAVE_XIMEA)
   if(WIN32)
-    get_filename_component(regpath "[HKEY_CURRENT_USER\\Software\\XIMEA\\CamSupport\\API;Path]" ABSOLUTE)
+    get_filename_component(regpath "[HKEY_LOCAL_MACHINE\\SOFTWARE\\XIMEA\\API_SoftwarePackage;Path]" ABSOLUTE)
```
alalek: This change breaks behavior of this script.

Proper flow should be:
- try `HKEY_CURRENT_USER` key
- if empty, then try `HKEY_LOCAL_MACHINE`

alalek:
> 3.4 branch

Well, this code is moved between 3.4/master branches. So lets put this patch into master first (and then backport via separate PR if needed).

PR #15378, opened 2019-08-23
Document usage of accuracy tests model

andrey-golubev:
### This pullrequest 

extends documentation on how to use new accuracy tests model (in *.hpp files with test fixtures declaration)


andrey-golubev:
Thanks!

Issue #12194, opened 2018-08-10
OpenCV Memory Leak within C++ threads


arthurarj:
##### System information (version)

- OpenCV => 3.4.2
- Operating System / Platform => Windows 10
- Compiler => Visual Studio 2017

##### Detailed description

Hello there! I'm experiencing some **memory leak** on OpenCV when working with C++ threads. I've seen some similar problems around but not quite the same. Also, the ones I've found were not solved.

I've reduced my code down to the simplest test one can make in which the problem persists. 

What it does: 1000 threads are spawned and each one of them will:
- Create a 300x300 Mat image filled with zeros;
- Add the image with itself and store in a new Mat. This is done 100 times within the each thread.

Memory leaks as shown below:
<img src="https://user-images.githubusercontent.com/13644591/43962014-e1a76a80-9c8c-11e8-8886-f6ac6009c7d2.png" width="50%">

##### Steps to reproduce

<!-- to add code example fence it with triple backticks and optional file extension
    ```.cpp
    // C++ code example
    ``` or attach as .txt or .zip file
-->

```
void static thread_call(int id, int n_threads) {
    cv::Mat teste = cv::Mat(300, 300, CV_32FC1, cv::Scalar(0));	
    for (int i = 0; i < 100; i++) {
        cv::Mat result;
	cv::add(teste, teste, result);
    }
}

void static test() {
    int n_threads = 1000;
    for (int i = 0; i < 1000; i++) {
	//Fork
	std::thread *t = new std::thread[n_threads];
	for (int i = 0; i < n_threads; ++i)
		t[i] = std::thread(thread_call, i, n_threads);
	//Join
	for (int i = 0; i < n_threads; ++i)
		t[i].join();
	delete[] t;
    }
    return;
}

int main()
{
    test();
    return 0;
}

```
Any insights are welcome :)

alalek:
> 1000 threads

Avoid this if you want to use OpenCV. Consider using ThreadPool for async tasks.
At least, because massive thread creation / destruction is heavy task.

catree:
Should be related I think #9745 (for cross reference and additional information).

Issue #9745, opened 2017-09-29
Memory leak in every thread


FloopCZ:
- OpenCV => 3.3.0
- Operating System / Platform => Arch Linux (4.12.13-1-ARCH SMP PREEMPT x86_64 GNU/Linux)
- Compiler => gcc 7.2.0 and clang 5.0.0

##### Detailed description

It seems that a significant piece of memory is leaked in every thread.
The following code snippet consumes cca 100MB/s.

##### Steps to reproduce

```.cpp
#include <opencv2/opencv.hpp>
#include <future>
int main()
{
    while(true) {
        std::async(std::launch::async, []() { cv::Mat::zeros(1, 1, CV_32FC3); }).wait();
    }
}
```

alalek:
Uncontrolled spawning of threads that uses OpenCV should be avoided. This is a bad programming practice.
OpenCV widely uses TLS data.
Initialization of thread is quite expensive operation, so consider to use thread pools instead for better experience.

BTW, This code with 1M threads consumes 400Mb in peak:
```
    for (size_t i = 0; i < 1000000; i++)
    {
        std::async(std::launch::async, []() {
            std::cout << cv::utils::getThreadID() << std::endl; // init TLS structures
            cv::Mat::zeros(1, 1, CV_32FC3);
        }).wait();
    }
```

FloopCZ:
Thank you for your reply @alalek, I thought this was the case. :-)

It is still unfortunate that the TLS data are not freed on thread termination, because in long running multithreaded applications one constantly looses memory with every call to `std::async` (which I consider to be a pretty standard practice if I am not doing something like `parallel_for`). The provided example was quite extreme, of course, but in my original code, I have noticed the leak after about 12 hours of intensive computation. If there is a list of feature requests, please consider this one and feel free to close this issue.

flyingfalling:
@alalek 
Apologies to add on to a closed thread, but a simple question: Is there a way to compile opencv so that it does not use TLS? I took a quick look at the source and noticed mostly it is used in the mutexes sections and google's protobuf.

Of course spawning threads uncontrollably is bad, but many times on a supercomputer or cluster, one will want to try many different instances of a model that includes some opencv code, and restarting the entire program (including loading all shared library calls etc.) N times is much more expensive than having a simple wrapper that spawns a new thread for each model one wants to try.

[...]

PR #15390, opened 2019-08-24
try fix memory leakage when creating threads(under linux)

lijinpei:
This pr tries to solve memory leakage when using threads. This may relate to issue #9745 and #12194 . Ideally, it should work for both linux and windows, however, my observation is that is solves the problem under linux (but I still have a case in my mind which leaks memory but not solved by this issue), and under windows it now leaks slower than #9745 described.

### This pullrequest changes

This pr basically does two things:
1. Remember to delete each thread's ThreadData when thread exits. Ths windows version has already done this in DllMain, I add that for the linux version in pthread_key_create.
2. Remember to delete ThreadData's slots when thread exits. The windows version used not to do this, and I add this for both the windows and linux version. To properly delete a void*, I have to add a deleter pointer for it.

If I'm right, the following code won't leak under linux, and leaks slower than before under windows.
https://gist.github.com/lijinpei/f2f7ace623a76454f1381646a9ebac51

### DIFF
--- a/modules/core/include/opencv2/core/utility.hpp
+++ b/modules/core/include/opencv2/core/utility.hpp
@@ -702,21 +702,56 @@ typedef std::recursive_mutex Mutex;
 typedef std::lock_guard<cv::Mutex> AutoLock;
 #endif
 
+template <typename T>
+struct TLSDeleter {
+    static void delete_it(void*data) {
+        delete (T*)data;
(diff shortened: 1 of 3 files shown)


lijinpei:
@alalek Do you have some feed back on this? Will this pr be merged?

PR #15399, opened 2019-08-27
Added performance test for StereoBM algorithm

terfendail:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/670

Added performance test for StereoBM algorithm

### DIFF
--- a/modules/calib3d/perf/perf_stereosgbm.cpp
+++ b/modules/calib3d/perf/perf_stereosgbm.cpp
@@ -41,20 +41,18 @@ namespace opencv_test
 using namespace perf;
 using namespace testing;
 
-static void MakeArtificialExample(RNG rng, Mat& dst_left_view, Mat& dst_view);
+static void MakeArtificialExample(Mat& dst_left_view, Mat& dst_view);
 
 CV_ENUM(SGBMModes, StereoSGBM::MODE_SGBM, StereoSGBM::MODE_SGBM_3WAY, StereoSGBM::MODE_HH4);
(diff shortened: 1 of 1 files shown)


modules/calib3d/perf/perf_stereosgbm.cpp
```diff
@@ -78,6 +78,33 @@ PERF_TEST_P( TestStereoCorresp, DISABLED_TooLongInDebug_SGBM, Combine(Values(Siz
     SANITY_CHECK(dst, .01, ERROR_RELATIVE);
 }
 
+typedef tuple<Size, int> BMParams;
+typedef TestBaseWithParam<BMParams> TestStereoCorrespBM;
+
+PERF_TEST_P(TestStereoCorrespBM, BM, Combine(Values(Size(1280, 720), Size(640, 480)), Values(256, 128)))
+{
+    RNG rng(0);
```
alalek: This is not needed in general:
https://github.com/opencv/opencv/blob/4.1.1/modules/ts/src/ts.cpp#L758-L763

terfendail: Default seed is non zero, so this change will affect sanity result for some test cases. I could remove it for BM test only and regenerate new sanity data, but IMO it would be better to retain similar tests uniform. The other option(the one I actually prefer) is to move zero RNG initialization to the `MakeArtificialExample` function that produce input to make things more consistent. The third option is to remove RNG initialization completely and regenerate both BM and SGBM sanity data.

PR #15400, opened 2019-08-27
Fix for AVX2 implementation of v_check_any(), v_check_all() intrinsics

terfendail:
### This pullrequest changes
Fix for AVX2 implementation of v_check_any(), v_check_all() intrinsics


alalek:
BTW, It would be nice to add corresponding tests.

PR #15419, opened 2019-08-29
WIP: G-API: Internal IE flag check in public headers issue

OrestChura:
### This pullrequest changes

- HAVE_INF_ENGINE flag check in headers "infer/ie.hpp" and "infer/ie/util.hpp" is deleted
- headers in "infer/" and "infer/ie/" folders are included into gapi_ext_hdrs;
+ because of that a few #includes are required in the headers

OpenCV public API should not have features-based conditional compilation in general (because these internal flags are not available in user apps). Instead of using such condoitions, user should add checks into their app.

Also the headers for G-API inference should be put into gapi_ext_hdrs CMake varaible

### DIFF
--- a/modules/gapi/src/backends/ie/util.hpp
+++ b/modules/gapi/src/backends/ie/util.hpp
@@ -7,6 +7,8 @@
 #ifndef OPENCV_GAPI_INFER_IE_UTIL_HPP
 #define OPENCV_GAPI_INFER_IE_UTIL_HPP
 
+#ifdef HAVE_INF_ENGINE
+
 // NOTE: This file is not included by default in infer/ie.hpp
 // and won't be. infer/ie.hpp doesn't depend on IE headers itself.
 // This file does -- so needs to be included separately by those who care.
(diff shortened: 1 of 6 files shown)


OrestChura:
@AsyaPronina @dmatveev @alalek please review

alalek:
> GAPI_EXPORTS std::vector<int> to_ocv(const InferenceEngine::SizeVector &dims);
> InferenceEngine::SizeVector

3rdparty types are not allowed in OpenCV regular public API.
Move this code into separate header under "details"/"private" sub-directory.

OrestChura:
> > GAPI_EXPORTS std::vector to_ocv(const InferenceEngine::SizeVector &dims);
> > InferenceEngine::SizeVector
> 
> 3rdparty types are not allowed in OpenCV regular public API.
> Move this code into separate header under "details"/"private" sub-directory.

@dmatveev 

I have added  changes you mentioned to make the header private, but mb it's not the right decision as the features could be helpful for a user.

modules/gapi/src/backends/ie/util.hpp
```diff
@@ -27,5 +30,6 @@ GAPI_EXPORTS InferenceEngine::Blob::Ptr to_ie(cv::Mat &blob);
```
dmatveev: Since this file is internal now, please put a note here why these functions are `EXPORT`ed (to make them accessible by the test suite)

PR #15438, opened 2019-09-02
GAPI: make static analysis happy in Fluid Internal Parallelism

anton-potapov:
### This pullrequest make static analysis happy in GAPI module
Simplified code a bit to make static analysis tool able to understand it

@dmatveev @alalek  please take a look


dmatveev:
Thanks!

PR #15440, opened 2019-09-02
Adding all possible data type interactions to the perf tests since some use SIMD acceleration and others do not.

### LINKED ISSUE #15439: Not all data type combinations are being tested with the integral function performance tests
- OpenCV => 3.4
- Operating System / Platform => Linux 64bit
- Compiler => gcc

##### Detailed description

I'm trying to analyse the performance of the integral function for several parameter
combinations since some of those combinations use HAL instructions and others do
not. Running the performance tests I realised that only one combinations of matrix
size for square sum and sum matrices.

everton1984:
resolves #15439

### This pullrequest changes

This PR is changing modules/imgproc/perf/perf_integral.cpp if you look at sumpixels.cpp there are several possible combinations of matrix sizes for the integral function. The performance tests do not cover all possibilities even with some of them being optimised with HAL instructions so it's impossible to analyse improvements made for those combinations.

### DIFF
--- a/modules/imgproc/perf/perf_integral.cpp
+++ b/modules/imgproc/perf/perf_integral.cpp
@@ -6,19 +6,19 @@
 namespace opencv_test {
 
 enum PerfSqMatDepth{
-    CV_32S32S = 0,
-    CV_32S32F,
-    CV_32S64F,
-    CV_32F32F,
-    CV_32F64F,
-    CV_64F64F};
+    SZ_32S32S = 0,
+    SZ_32S32F,
+    SZ_32S64F,
+    SZ_32F32F,
+    SZ_32F64F,
+    SZ_64F64F};
 
-CV_ENUM(SqSizes, CV_32S32S, CV_32S32F, CV_32S64F, CV_32F32F, CV_32F64F, CV_64F64F);
+CV_ENUM(IntegralOutputDepths, SZ_32S32S, SZ_32S32F, SZ_32S64F, SZ_32F32F, SZ_32F64F, SZ_64F64F);
 
 typedef tuple<Size, MatType, MatDepth> Size_MatType_OutMatDepth_t;
 typedef perf::TestBaseWithParam<Size_MatType_OutMatDepth_t> Size_MatType_OutMatDepth;
 
-typedef tuple<Size, MatType, SqSizes> Size_MatType_OutMatDepthArray_t;
+typedef tuple<Size, MatType, IntegralOutputDepths> Size_MatType_OutMatDepthArray_t;
 typedef perf::TestBaseWithParam<Size_MatType_OutMatDepthArray_t> Size_MatType_OutMatDepthArray;
 
 PERF_TEST_P(Size_MatType_OutMatDepth, integral,
@@ -73,7 +73,7 @@ PERF_TEST_P(Size_MatType_OutMatDepthArray, DISABLED_integral_sqsum_full,
             testing::Combine(
                 testing::Values(TYPICAL_MAT_SIZES),
                 testing::Values(CV_8UC1, CV_8UC2, CV_8UC3, CV_8UC4),
-                testing::Values(0,1,2,3,4,5)
+                testing::Values(SZ_32S32S,SZ_32S32F,SZ_32S64F,SZ_32F32F,SZ_32F64F,SZ_64F64F)
                 )
             )
 {
@@ -92,8 +92,8 @@ PERF_TEST_P(Size_MatType_OutMatDepthArray, DISABLED_integral_sqsum_full,
 
     TEST_CYCLE() integral(src, sum, sqsum, sdepth, sqdepth);
 
-    SANITY_CHECK(sum, 1e-6);
-    SANITY_CHECK(sqsum, 1e-6);
+    SANITY_CHECK_NOTHING();
+    SANITY_CHECK_NOTHING();
 
 }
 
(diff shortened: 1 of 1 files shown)


modules/imgproc/perf/perf_integral.cpp
```diff
@@ -5,9 +5,22 @@
...
 
+enum PerfSqMatDepth{
+    CV_32S32S = 0,
+    CV_32S32F,
+    CV_32S64F,
+    CV_32F32F,
+    CV_32F64F,
+    CV_64F64F};
+
+CV_ENUM(SqSizes, CV_32S32S, CV_32S32F, CV_32S64F, CV_32F32F, CV_32F64F, CV_64F64F);
```
alalek: > SqSizes

This doesn't look as "sizes". Perhaps it should be `IntergralOutputDepths`

Also it is better to replace `CV_` prefix to avoid confusion with public OpenCV API.

modules/imgproc/perf/perf_integral.cpp
```diff
@@ -31,28 +44,57 @@ PERF_TEST_P(Size_MatType_OutMatDepth, integral,
...
+    SANITY_CHECK(sqsum, 1e-6);
+}
+
+int vals[6][2] = {{CV_32S, CV_32S}, {CV_32S, CV_32F}, {CV_32S, CV_64F}, {CV_32F, CV_32F}, {CV_32F, CV_64F}, {CV_64F, CV_64F}};
+PERF_TEST_P(Size_MatType_OutMatDepthArray, DISABLED_integral_sqsum_full,
             testing::Combine(
                 testing::Values(TYPICAL_MAT_SIZES),
                 testing::Values(CV_8UC1, CV_8UC2, CV_8UC3, CV_8UC4),
-                testing::Values(CV_32S, CV_32F, CV_64F)
+                testing::Values(0,1,2,3,4,5)
```
alalek: Avoid using of magic numbers. Use enum constants identifiers instead.

modules/imgproc/perf/perf_integral.cpp
```diff
@@ -31,28 +44,57 @@ PERF_TEST_P(Size_MatType_OutMatDepth, integral,
...
+    Mat sqsum(sz, sqdepth);
 
     declare.in(src, WARMUP_RNG).out(sum, sqsum);
     declare.time(100);
 
-    TEST_CYCLE() integral(src, sum, sqsum, sdepth);
+    TEST_CYCLE() integral(src, sum, sqsum, sdepth, sqdepth);
 
     SANITY_CHECK(sum, 1e-6);
     SANITY_CHECK(sqsum, 1e-6);
```
alalek: You don't have sanity data for this test (in opencv_extra repositry). Replace these 2 lines by `SANITY_CHECK_NOTHING();`

everton1984: Thanks for the feedback, just made the changes you requested.

modules/imgproc/perf/perf_integral.cpp
```diff
@@ -55,6 +68,35 @@ PERF_TEST_P(Size_MatType_OutMatDepth, integral_sqsum,
...
+    Mat sqsum(sz, sqdepth);
+
+    declare.in(src, WARMUP_RNG).out(sum, sqsum);
+    declare.time(100);
+
+    TEST_CYCLE() integral(src, sum, sqsum, sdepth, sqdepth);
+
+    SANITY_CHECK_NOTHING();
+    SANITY_CHECK_NOTHING();
+
```
alalek: One macro should be enough

modules/imgproc/perf/perf_integral.cpp
```diff
@@ -31,12 +44,12 @@ PERF_TEST_P(Size_MatType_OutMatDepth, integral,
...
-                testing::Values(CV_8UC1, CV_8UC2, CV_8UC3, CV_8UC4),
-                testing::Values(CV_32S, CV_32F, CV_64F)
-                )
-            )
+                testing::Combine(
+                    testing::Values(TYPICAL_MAT_SIZES),
+                    testing::Values(CV_8UC1, CV_8UC2, CV_8UC3, CV_8UC4),
+                    testing::Values(CV_32S, CV_32F, CV_64F)
+                    )
+        )
```
alalek: Lets avoid unrelated changes.

modules/imgproc/perf/perf_integral.cpp
```diff
@@ -5,9 +5,22 @@
...
 
+enum PerfSqMatDepth{
+    SZ_32S32S = 0,
+    SZ_32S32F,
+    SZ_32S64F,
+    SZ_32F32F,
+    SZ_32F64F,
+    SZ_64F64F};
+
+CV_ENUM(IntegralOutputDepths, SZ_32S32S, SZ_32S32F, SZ_32S64F, SZ_32F32F, SZ_32F64F, SZ_64F64F);
```
alalek: > SZ_

What is the meaning of this?
Depth != Size (size is usually associated with image size in pixels, not a "size" of type)

`DEPTH_32S_32S`?

modules/imgproc/perf/perf_integral.cpp
```diff
@@ -55,6 +68,35 @@ PERF_TEST_P(Size_MatType_OutMatDepth, integral_sqsum,
     SANITY_CHECK(sqsum, 1e-6);
 }
 
+int vals[6][2] = {{CV_32S, CV_32S}, {CV_32S, CV_32F}, {CV_32S, CV_64F}, {CV_32F, CV_32F}, {CV_32F, CV_64F}, {CV_64F, CV_64F}};
```
alalek: This should be close to related enumeration above.
This should be properly named.

This must be "static" or in anonymous namespace due its "local" usage (we don't want conflicts especially on such common name).

everton1984:
Changed the names per request, made the array static and moved the definition closer to the enum.

PR #15091, opened 2019-07-18
G-API: Fluid Internal Parallelism

anton-potapov:
### This pullrequest adds: 

- ability to use custom implementation of "parallel for" function for Fluid internal parallelism 

<cut/>

```
build_gapi_standalone:Linux x64=ade-0.1.1d
build_gapi_standalone:Win64=ade-0.1.1d
build_gapi_standalone:Mac=ade-0.1.1d
build_gapi_standalone:Linux x64 Debug=ade-0.1.1d
```

PR #15443, opened 2019-09-03
Fluid Internal Parallelism:

anton-potapov:
### G-API:  add comments on signature of custom parallel_for

as a follow -up for #15091 

@dmatveev please take a look

```
```


dmatveev:
Thanks but what about the another one (use `cv::paralllel_for` by default instead of own serial cycle)?

anton-potapov:
As for CMake : 
#15485

PR #15444, opened 2019-09-03
core(OpenCL): fix fft kernel compilation

alalek:
- error: variables in the local address space can only be declared in the outermost scope of a kernel function

[Build logs](http://pullrequest.opencv.org/buildbot/builders/3_4-win64-vc16/builds/94/steps/test_core-opencl/logs/stdio)


alalek:
:+1:

PR #15449, opened 2019-09-04
cmake: download tries, customizable download parameters

### LINKED ISSUE #13573: Short timeout in ocv_download causes build problems
ocv_download uses a [timeout of 600 seconds](https://github.com/opencv/opencv/blob/master/cmake/OpenCVDownload.cmake#L159), which is problematic when downloading large resources. This has already caused problems in the past (#11672, #10825), and is also a problem for me right now: Downloading face_landmark_model.dat fails due to the download speed barely reaching 50KiB/s (I don't know whether this is a problem of GitHub or my provider, I can only confirm it is not caused on my end, and that other downloads from GitHub are much faster).

This, together with the fact that failed downloads don't cause CMake to abort (deferring the errors to later steps of the build process), has been causing me quite a bit of trouble in the last few hours, which is why I suggest to remove the timeout completely.

mshabunin:
resolves #13573

### This pullrequest changes

Adds two paramters to customize cmake download process for unstable and slow networks:
* `OPENCV_DOWNLOAD_TRIES_LIST` - list of items representing download tries:
* `OPENCV_DOWNLOAD_PARAMS` - raw parameter string to be used in cmake's `file(DOWNLOAD ...)` call
* added directory creation to fallback `download_with_*.sh` scripts

<cut/>

Examples:

bash:
```.sh
cmake \
    -DOPENCV_DOWNLOAD_TRIES_LIST=1\;2\;3\;4\;5 \
    -DOPENCV_DOWNLOAD_PARAMS=INACTIVITY_TIMEOUT\;30\;TIMEOUT\;180\;SHOW_PROGRESS \
../opencv
```
cmd:
```.bat
cmake ^
    -DOPENCV_DOWNLOAD_TRIES_LIST=a;b;c ^
    -DOPENCV_DOWNLOAD_PARAMS=INACTIVITY_TIMEOUT;30;TIMEOUT;180;SHOW_PROGRESS ^
../opencv
```

Edit: renamed to `OPENCV_DOWNLOAD_TRIES` => `OPENCV_DOWNLOAD_TRIES_LIST`, marked both variables as advanced.

```
allow_multiple_commits=1
```

### DIFF
--- a/cmake/OpenCVDownload.cmake
+++ b/cmake/OpenCVDownload.cmake
@@ -204,8 +204,8 @@ For details please refer to the download log file:
 ${OPENCV_DOWNLOAD_LOG}
 ")
       # write helper scripts for failed downloads
-      file(APPEND "${OPENCV_DOWNLOAD_WITH_CURL}" "curl --output \"${CACHE_CANDIDATE}\" \"${DL_URL}\"\n")
-      file(APPEND "${OPENCV_DOWNLOAD_WITH_WGET}" "wget -O \"${CACHE_CANDIDATE}\" \"${DL_URL}\"\n")
+      file(APPEND "${OPENCV_DOWNLOAD_WITH_CURL}" "curl --create-dirs --output \"${CACHE_CANDIDATE}\" \"${DL_URL}\"\n")
+      file(APPEND "${OPENCV_DOWNLOAD_WITH_WGET}" "mkdir -p $(dirname ${CACHE_CANDIDATE}) && wget -O \"${CACHE_CANDIDATE}\" \"${DL_URL}\"\n")
       return()
     endif()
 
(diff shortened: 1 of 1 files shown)


alalek:
> -DOPENCV_DOWNLOAD_TRIES=1\;2\;3\;4\;5

This is confusing approach. This should be a single number.

Also it would be nice to support some value to turn off downloading completely (but use cache). Something like [this](https://github.com/microsoft/vcpkg/blob/a3e8197f2d7fa043f5c0dc5ccac3446fac6ebac2/ports/opencv4/0001-disable-downloading.patch).

mshabunin:
I agree, it is confusing but I don't think it is an issue:
* this functionality is for automated systems, not humans
* this approach can be less prone to errors, e.g. `TRIES=-8` will work too

Maybe we should mark these variables as `ADVANCED`?

It is possible to turn download off by providing an empty list: `TRIES=`
```.sh
cmake -DOPENCV_DOWNLOAD_TRIES= ../opencv
# a bunch of warnings is shown ...
bash download_with_curl.sh
cmake .
```

PR #15454, opened 2019-09-05
VideoCapture get property with range and default value support

### LINKED ISSUE #15219: where to get the value range of property of VideoCapture::set(id, value)?

- OpenCV => :grey_question:
- Compiler => :grey_question:

iamgd67:
**RFC**
resolves #15219
### This pullrequest changes

VideoCapture get properties support range and default value, add msmf and dshow backend implemention

### DIFF
--- a/modules/videoio/include/opencv2/videoio.hpp
+++ b/modules/videoio/include/opencv2/videoio.hpp
@@ -776,6 +776,17 @@ public:
     */
     CV_WRAP virtual double get(int propId) const;
 
+	/** @brief like get(int propId) but returns property range and default as well as value
+
+	@return true if success, false if querying a property that is not supported by the backend
+	used by the VideoCapture instance.
+
+	@note not all Property has a range and/or default value, if not , will set to min/max/defaultVal to -1,
+	see get(int propId) for more information
+	*/
+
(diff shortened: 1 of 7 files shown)


iamgd67:
@alalek are we waiting others opinion?

iamgd67:
@mshabunin could you please help review this pr? It has been opened for some days and not know what to continue.

mshabunin:
@iamgd67 , similar problem have been addressed differently in V4L backend:
https://github.com/opencv/opencv/blob/2fef9bc1d8ba47d253a034b9efcb26591d6011a6/modules/videoio/src/cap_v4l.cpp#L1779-L1791

VideoCapture can be switched to NORMALIZED_PROPERTIES mode and selected properties will accept and return values from range `[0, 1]`:
https://github.com/opencv/opencv/blob/2fef9bc1d8ba47d253a034b9efcb26591d6011a6/modules/videoio/src/cap_v4l.cpp#L323-L330

I think it is better to use similar approach in other backends.

asmorkalov:
@alalek @asmorkalov @vpisarev @vladimir-dudnik decided to close the PR on weekly meeting by the following reasons:
- The solution breaks compatibility and the same functionality can be achieved with several `cv::VideoCapture::get()` calls.
- The solution is proposed for one backend only and does not cover other cases.

PR #15467, opened 2019-09-05
dshow: Add ability to save directshow graph

elmsfu:
### This pullrequest changes

Adds ability to save the directshow graph by setting an environment variable providing a filename.


modules/videoio/src/cap_dshow.cpp
```diff
@@ -2946,6 +2946,22 @@ int videoInput::start(int deviceID, videoDevice *VD){
     DebugPrintOut("SETUP: Device is setup and ready to capture.\n\n");
     VD->readyToCapture = true;
 
+    // check for optional saving the direct show graph to a file
+    const char* graph_filename = getenv("OPENCV_DSHOW_SAVEGRAPH_FILENAME");
+    if (graph_filename) {
+        WCHAR* wfilename = new WCHAR[strlen(graph_filename) + 1];
```
alalek: > new

Manual memory management should be avoided in OpenCV code.
Use `std::vector<WCHAR>` or `cv::AutoBuffer<WCHAR>` instead.

modules/videoio/src/cap_dshow.cpp
```diff
@@ -2946,6 +2946,22 @@ int videoInput::start(int deviceID, videoDevice *VD){
...
+        mbstowcs(wfilename, graph_filename, strlen(graph_filename) + 1);
```
alalek: return value of `mbstowcs()` should be checked (asserted)

modules/videoio/src/cap_dshow.cpp
elmsfu: Done

modules/videoio/src/cap_dshow.cpp
elmsfu: Done

elmsfu:
Rebased to 3.4

modules/videoio/src/cap_dshow.cpp
```diff
@@ -2936,6 +2936,21 @@ int videoInput::start(int deviceID, videoDevice *VD){
     DebugPrintOut("SETUP: Device is setup and ready to capture.\n\n");
     VD->readyToCapture = true;
 
+    // check for optional saving the direct show graph to a file
+    const char* graph_filename = getenv("OPENCV_DSHOW_SAVEGRAPH_FILENAME");
+    if (graph_filename) {
+        std::vector<WCHAR> wfilename(strlen(graph_filename) + 1, L' ');
```
alalek: Perhaps `L'\0'` should be better (or just remove second argument).

modules/videoio/src/cap_dshow.cpp
```diff
@@ -2936,6 +2936,21 @@ int videoInput::start(int deviceID, videoDevice *VD){
...
+        size_t len = mbstowcs(&wfilename[0], graph_filename, strlen(graph_filename) + 1);
+        CV_Assert(len == strlen(graph_filename));
```
alalek: > strlen(graph_filename)

This expression has been used 3 times here. We may want to cache this value in a separate variable.

PR #15477, opened 2019-09-06
Remove old build java option

dab0bby:
### This pullrequest changes
Removes old `BUILD_opencv_java` option and updates documentation to the actual `BUILD_JAVA` option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1571,7 +1571,7 @@ if(FLAKE8_FOUND AND FLAKE8_EXECUTABLE)
 endif()
 
 # ========================== java ==========================
-if(BUILD_JAVA OR BUILD_opencv_java)
+if(BUILD_JAVA)
   status("")
   status("  Java:"            BUILD_FAT_JAVA_LIB  THEN "export all functions"                                      ELSE "")
   status("    ant:"           ANT_EXECUTABLE      THEN "${ANT_EXECUTABLE} (ver ${ANT_VERSION})"                    ELSE NO)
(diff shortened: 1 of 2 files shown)


dkurt:
As for me, we need to keep BUILD_opencv_java instead because it's more convinient for modules (like BUILD_opencv_python)

dab0bby:
@dkurt the only option that exists for building Java is here: https://github.com/opencv/opencv/blob/4de115c08cd751d9cf3a45f593e217a104ab8fa0/CMakeLists.txt#L438

I guesss the BUILD_opencv_java option is a leftover

CMakeLists.txt
```diff
@@ -1571,7 +1571,7 @@ if(FLAKE8_FOUND AND FLAKE8_EXECUTABLE)
 endif()
 
 # ========================== java ==========================
-if(BUILD_JAVA OR BUILD_opencv_java)
```
alalek: `BUILD_opencv_XXX` are widely used CMake variables (they are defined for all modules like `BUILD_opencv_highgui`).

`BUILD_JAVA=ON` doesn't not guarantee that all dependencies are found (JDK and Ant) and `BUILD_opencv_java=ON`

dkurt: So `BUILD_opencv_java` is more reliable option? I mean if `java` is OpenCV module it'd be better to keep convenient  `BUILD_opencv_java`.

alalek: `BUILD_JAVA` behaves like `WITH_XXX` options (in case of turning off it avoids unnecessary JDK/Ant searching).
`BUILD_opencv_java` stays here (as OpenCV modules variable).

PR #15486, opened 2019-09-09
Split assertion to make error message more specific

abrock:
I think having the assertion split into two separate assertions gives the programmer a better hint which of the matrices has the wrong size and speeds up debugging.


mshabunin:
@abrock , please create new pull request targeting branch _3.4_ with branch name different from _master_ or _3.4_. Read more here: https://github.com/opencv/opencv/wiki/Branches

PR #15489, opened 2019-09-09
CV_PAUSE: Include definition of _mm_pause.

eliteraspberries:
I tried to build OpenCV version 4.1.1 with minimal optimizations (CV_DISABLE_OPTIMIZATION=1) and I get the following error:

    [ 95%] Building CXX object modules/core/CMakeFiles/opencv_core.dir/src/parallel_impl.cpp.o
    /Users/.../opencv-4.1.1/modules/core/src/parallel_impl.cpp:368:21: error: use of
          undeclared identifier '_mm_pause'
                        CV_PAUSE(16);
                        ^
    /Users/.../opencv-4.1.1/modules/core/src/parallel_impl.cpp:45:79: note: 
          expanded from macro 'CV_PAUSE'
      ...do { for (int __delay = (v); __delay > 0; --__delay) { _mm_pause(); } } ...
                                                                ^

The `_mm_pause` intrinsic function is defined in the header file `emmintrin.h` (which comes with Clang). https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=pause&expand=4141

### DIFF
--- a/modules/core/src/parallel_impl.cpp
+++ b/modules/core/src/parallel_impl.cpp
@@ -42,6 +42,7 @@ DECLARE_CV_PAUSE
       static inline void cv_non_sse_mm_pause() { __asm__ __volatile__ ("rep; nop"); }
 #     define _mm_pause cv_non_sse_mm_pause
 #   endif
+#   include <emmintrin.h>
 #   define CV_PAUSE(v) do { for (int __delay = (v); __delay > 0; --__delay) { _mm_pause(); } } while (0)
 # elif defined __GNUC__ && defined __aarch64__
 #   define CV_PAUSE(v) do { for (int __delay = (v); __delay > 0; --__delay) { asm volatile("yield" ::: "memory"); } } while (0)


alalek:
Could you please dump compiler command line (`make VERBOSE=1`)?
Could you add `-H` compiler option and dump list of included files? (In my case `emmintrin.h` has been included via `fastmath.hpp` file)

TBD on GitHub.

eliteraspberries:
In your log it shows `emmintrin.h` is from GCC (`/usr/lib/gcc/x86_64-redhat-linux/8/include`).

I am using macOS, with Clang from pkgsrc. Note that `emmintrin.h` is part of Clang and not macOS or Xcode:

    $ find pkg -name emmintrin.h
    pkg/lib/clang/3.8.0/include/emmintrin.h
    $ find /usr/include -name emmintrin.h
    $ find $(xcrun --sdk macosx --show-sdk-path) -name emmintrin.h

The only headers that include emmintrin.h are cv_cpu_dispatch.h and fast_math.hpp.

In fast_math.hpp I see emmintrin.h is not included if `__APPLE__` is defined:

    #if ((defined _MSC_VER && defined _M_X64) || (defined __GNUC__ && defined __x86_64__ \
        && defined __SSE2__ && !defined __APPLE__)) && !defined(__CUDACC__)
    #include <emmintrin.h>
    #endif

Removing this condition also solves the problem.

Now I see this was fixed with commit 8a0b93bc4dd3183eb63ff6eed42120463878058c. Thank you.

PR #15491, opened 2019-09-10
Fixed universal intrinsic undistort() implementation

### LINKED ISSUE #15286: cv::undistort behaviour changed in new SIMD implementation
##### System information
- OpenCV => 4.1.1
- Operating System / Platform => Ubuntu 18.04 64bit
- Compiler => gcc 7.4.0

##### Detailed description

The SIMD_64F implementation of cv::undistort in undistort.simd.hpp' gives different results than the non-optimized version (and previous versions of OpenCV such as 4.1.0 and below)
For the example below the resulting image seems to be offset about half a pixel down and a smaller fraction of pixel left.

##### Steps to reproduce

Image source:
lena_512_24bpp

    ```.cpp
    // C++ code example
    double kmat_data[9] = {3217, 0, 1592, 0, 3217, 1201, 0, 0, 1};
    cv::Mat kmat = cv::Mat(3, 3, CV_64F, kmat_data);
    double dist_coeff_data[5] = { 0.04, -0.4, -0.01, 0.04, 0.7 };
    cv::Mat dist_coeffs = cv::Mat(5, 1, CV_64F, dist_coeff_data);
    cv::Mat img_undist, img = cv::read( "lena_512_24bpp.png" );

    cv::undistort( img, img_undist, kmat, dist_coeffs );
    ```

Disabling the `#CV_SIMD_64F` flags in `moduels/calib3d/src/undistort.simd.hpp' and recompiling results in a different image (seems to be offset by a fraction of a pixel in both dimensions)

##### Output of above code

Output in OpenCV 3.4.6 (same system)
cv_test 3 4 6

Output in OpenCV 4.1.1
cv_test 4 1 1

terfendail:
resolves #15286

### This pullrequest changes
Fixed universal intrinsic undistort() implementation

### DIFF
--- a/modules/calib3d/test/test_undistort.cpp
+++ b/modules/calib3d/test/test_undistort.cpp
@@ -1114,6 +1114,44 @@ TEST(Calib3d_UndistortPoints, outputShape)
     }
 }
 
+TEST(Imgproc_undistort, regression_15286)
+{
+    double kmat_data[9] = { 3217, 0, 1592, 0, 3217, 1201, 0, 0, 1 };
+    Mat kmat(3, 3, CV_64F, kmat_data);
+    double dist_coeff_data[5] = { 0.04, -0.4, -0.01, 0.04, 0.7 };
+    Mat dist_coeffs(5, 1, CV_64F, dist_coeff_data);
+
+    Mat img = Mat::zeros(512, 512, CV_8UC1);
+    img.at<uchar>(128, 128) = 255;
+    img.at<uchar>(128, 384) = 255;
+    img.at<uchar>(384, 384) = 255;
+    img.at<uchar>(384, 128) = 255;
+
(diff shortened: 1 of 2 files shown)


modules/imgproc/test/test_imgwarp.cpp
```diff
@@ -1909,5 +1909,43 @@ TEST(Imgproc_warpPolar, identity)
 #endif
 }
 
+TEST(Imgproc_undistort, regression_15286)
```
alalek: Please move this test into calib3d module: https://github.com/opencv/opencv/blob/3.4.7/modules/calib3d/test/test_undistort.cpp
(`undistort()` code is in `imgproc` module for legacy reasons only, in OpenCV 4.x it is moved to `calib3d`)

terfendail: Moved

PR #15494, opened 2019-09-10
Improving VSX performance of integral function

everton1984:
This pullrequest is adding support for the get function described in intrin_cpp.hpp line 322 for VSX and adapting the integral HAL implementations to use it. The way the integral functions are accessing a single element from the vector register require 3 instructions on VSX 2 of which are inside the loop. This new implementation requires only a single instruction.

```
force_builders=Linux AVX2,armv7,armv8,Custom,Win32,Linux32

build_image:Custom Win=msvs2017

#buildworker:Custom=linux-3
#build_image:Custom=ubuntu:18.04
#CPU_BASELINE:Custom=AVX512_SKX
#disable_ipp=ON

#buildworker:Custom=linux-1
#build_image:Custom=mips64el
#build_image:Custom=javascript-simd
```

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin_avx.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_avx.hpp
@@ -2290,21 +2290,72 @@ inline double v_extract_n(v_float64x4 v)
     return v_rotate_right<i>(v).get0();
 #endif
 }
-/*
-#define OPENCV_HAL_IMPL_AVX_EXTRACT_N(_Tpvec, _Tp) \
-template<int i> inline _Tp v_extract_n( _Tpvec v) { return v_rotate_right<i>(v).get0(); }
 
-OPENCV_HAL_IMPL_AVX_EXTRACT_N(v_uint8x32, uchar)
-OPENCV_HAL_IMPL_AVX_EXTRACT_N(v_int8x32, schar)
-OPENCV_HAL_IMPL_AVX_EXTRACT_N(v_uint16x16, ushort)
-OPENCV_HAL_IMPL_AVX_EXTRACT_N(v_int16x16, short)
-OPENCV_HAL_IMPL_AVX_EXTRACT_N(v_uint32x8, uint)
-OPENCV_HAL_IMPL_AVX_EXTRACT_N(v_int32x8, int)
-OPENCV_HAL_IMPL_AVX_EXTRACT_N(v_uint64x4, uint64)
-OPENCV_HAL_IMPL_AVX_EXTRACT_N(v_int64x4, int64)
-OPENCV_HAL_IMPL_AVX_EXTRACT_N(v_float32x8, float)
-OPENCV_HAL_IMPL_AVX_EXTRACT_N(v_float64x4, double)
-*/
+template<int i>
+inline v_int32x8 v_broadcast_element(v_int32x8 v)
+{
+    return v_int32x8(_mm256_shuffle_epi32(v.val,  _MM_SHUFFLE(i,i,i,i)));
+}
+
+template<int i>
+inline v_uint32x8 v_broadcast_element(v_uint32x8 v)
+{
+    return v_uint32x8(_mm256_shuffle_epi32(v.val, _MM_SHUFFLE(i,i,i,i)));
+}
+
+template<int i>
+inline v_int16x16 v_broadcast_element(v_int16x16 v)
+{
+    v_int16x16 t = v_int16x16(_mm256_shufflelo_epi16(v.val, _MM_SHUFFLE(i, i, i, i)));
+    return v_int16x16(_mm256_shufflehi_epi16(t.val, _MM_SHUFFLE(0, 0, 0, 0)));
+}
+
+template<int i>
+inline v_uint16x16 v_broadcast_element(v_uint16x16 v)
+{
+    v_uint16x16 t = v_uint16x16(_mm256_shufflelo_epi16(v.val, _MM_SHUFFLE(i, i, i, i)));
+    return v_uint16x16(_mm256_shufflehi_epi16(t.val, _MM_SHUFFLE(0, 0, 0, 0)));
+}
+
+template<int i>
+inline v_int8x32 v_broadcast_element(v_int8x32 v)
+{
+    static const __m256i perm = _mm256_set1_epi8((char)i);
+    return v_int8x32(_mm256_shuffle_epi8(v.val, perm));
+}
+
+template<int i>
+inline v_uint8x32 v_broadcast_element(v_uint8x32 v)
+{
+    static const __m256i perm = _mm256_set1_epi8((char)i);
+    return v_uint8x32(_mm256_shuffle_epi8(v.val, perm));
+}
+
+template<int i>
+inline v_int64x4 v_broadcast_element(v_int64x4 v)
+{
+    int64 t = v_extract_n<i>(v);
+    return v_int64x4(t,t,t,t);
+}
+
+template<int i>
+inline v_uint64x4 v_broadcast_element(v_uint64x4 v)
+{
+    uint64 t = v_extract_n<i>(v);
+    return v_uint64x4(t,t,t,t);
+}
+
+template<int i>
+inline v_float64x4 v_broadcast_element(v_float64x4 v)
+{
+    return v_float64x4(_mm256_shuffle_pd(v.val, v.val, _MM_SHUFFLE(i,i,i,i)));
+}
+
+template<int i>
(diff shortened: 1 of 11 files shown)


mshabunin:
I think it is better to implement `get()` method and corresponding tests for all intrinsics variants (x86, NEON, ...) and remove `CV_VSX` macros in the code.

everton1984:
I don't know if that's really a good idea, the implementation enhances performance for Power9 but I'm not really confident it would do the same for other platforms. Although I could try to simply implement 'get' with the shift approach summ_pixels was using. What do you think?

mshabunin:
Yes, I meant this. If it there is native intrinsic for _get_ - we should use it, otherwise rotate and _get0_. One thing is, this method should be template, because _v_rotate_ is template.

everton1984:
I tried to follow intrin_cpp's implementation approach
```
    _Tp get(const int i) const { return s[i]; }
```
just to keep consistency. Don't you think this is a better approach? I kinda fear implementing a template version would break consistency a bit.

mshabunin:
We can try this, but I doubt it will work with v_rotate, because underlying x86 and NEON intrinsics accept immediate value which must be known at compile time.

I've enabled ARM builders for this PR, so you can try to add a commit and see if it works with CI builds (https://pullrequest.opencv.org/#/summary/).

modules/core/include/opencv2/core/hal/intrin_vsx.hpp
```diff
@@ -38,6 +38,8 @@ struct v_uint8x16
     {}
     uchar get0() const
     { return vec_extract(val, 0); }
+    uchar get(const int& i) const
```
alalek: Should be templated:

```
template<int i>
uchar get() const;
{
    return v_rotate_right<i>(*this).get0();
}
```

with this implementation for all other non-VSX SIMD backends.

---

Also we can add a separate function (`vec_extract<i>(a)` ?) instead of member function.

everton1984: ok so what about removing the member functions and implementing something like this on VSX?

```
#define OPENCV_HAL_IMPL_VSX_EXTRACT(_Tpvec, _Tp) \
template<int i> inline _Tp v_extract(_Tpvec v) { return vec_extract(v.val, i); }

OPENCV_HAL_IMPL_VSX_EXTRACT(v_uint8x16, uchar)
OPENCV_HAL_IMPL_VSX_EXTRACT(v_int8x16, schar)
OPENCV_HAL_IMPL_VSX_EXTRACT(v_uint16x8, ushort)
OPENCV_HAL_IMPL_VSX_EXTRACT(v_int16x8, short)
OPENCV_HAL_IMPL_VSX_EXTRACT(v_uint32x4, uint)
OPENCV_HAL_IMPL_VSX_EXTRACT(v_int32x4, int)
OPENCV_HAL_IMPL_VSX_EXTRACT(v_uint64x2, uint64)
OPENCV_HAL_IMPL_VSX_EXTRACT(v_int64x2, int64)
OPENCV_HAL_IMPL_VSX_EXTRACT(v_float32x4, float)
OPENCV_HAL_IMPL_VSX_EXTRACT(v_float64x2, double)```

I think it might be better to name it v_extract instead of vec_extract first because vec_extract is the name of the intrinsic and secondly because other HAL functions usually start with v_ or vx_ instead of vec_. 

For the other archs we can use the rotate approach. What do you think?

**Comment by alalek** [comment]
BTW, Merge conflicts should be fixed through rebasing commits onto fresh upstream HEAD commit. It is nice to squash commits with rebasing.

**Comment by everton1984** [comment]
I'm planning to squash as soon as the tests are ok.

File: modules/imgproc/src/sumpixels.cpp [hunk 2]
```diff
@@ -147,7 +147,9 @@ struct Integral_SIMD<uchar, int, double>
                 v_expand(el8, el4l, el4h);
                 el4l += prev;
                 el4h += el4l;
-                prev = vx_setall_s32(v_rotate_right<v_int32::nlanes - 1>(el4h).get0());
+
+                prev = vx_setall_s32(v_extract_n<v_int32::nlanes - 1>(el4h));
```
alalek: Probably more specific intrinsic fits better in this case:
- something like `v_broadcast_element<v_int32::nlanes - 1>(el4h)`
- can be implemented through permutation/shuffle

alalek: Yes.

Plain C++ code:
```
/** Broadcast i-th element of vector
Scheme:
@code
{ v[0] v[1] v[2] ... v[SZ] } => { v[i], v[i], v[i] ... v[i] }
@endcode
Restriction: 0 <= i < nlanes (supported: TBD)
Supported types: TBD (lets start from `int32_t` - make sense after revieving of real intrinsic support)
 */
template<int i, typename _Tp, int n>
inline v_reg<_Tp, n> v_broadcast_element(const v_reg<_Tp, n>& a)
{
    return v_reg<_Tp, n>::all(a.s[i]);
}
```

Do you see any objections?

alalek:
Thank you for update!

I pushed commit with adding corresponding tests for new intrinsics.

Some code compilation is failed on Linux x64 (fixed / commented out), Windows (`_mm256_extract_epi64` issue), WASM/MSA backends (added stubs).

modules/core/include/opencv2/core/hal/intrin_avx.hpp
```diff
@@ -2195,6 +2239,146 @@ OPENCV_HAL_IMPL_AVX_EXTRACT(v_int64x4)
...
+
+template<int i>
+inline v_float32x8 v_broadcast_element(const v_float32x8 &a)
+{ return v_reinterpret_as_f32(v_broadcast_element<i>(v_reinterpret_as_u32(a))); }
+
+/*
+template<int i>
+inline v_uint32x8 v_broadcast_element(v_uint32x8 v)
+{
+    return v_uint32x8(_mm256_shuffle_epi32(v.val, _MM_SHUFFLE(i,i,i,i)));
```
alalek: I will remove commented out code in a separate commit tomorrow. Code is still available in the history of this PR.
@everton1984 Do you have any objections? (or time to update these extended cases too)

everton1984: @alalek Hi! Sorry.it took me so long to answer, I was on holiday. By extended cases you mean WASM/MSA?

alalek: I mean `v_broadcast_element` for types other than int32/uint32/float32 (see commented out code).

It is not about WASM/MSA.

everton1984: ok let me try to tackle it a bit, I'm not an Intel expert as you might have noticed :) but since there are tests written I might be able to explore. I'll post something as soon as I have it working.

alalek: These "other" cases are not tested or used, so we can remove these implementations if you don't have objections.

everton1984: TBH I don't see a problem, although I can see the need for at least a 16-bit version if someone wants to tackle vectorising the integral function for other datatypes in the future.

modules/core/include/opencv2/core/hal/intrin_vsx.hpp
```diff
@@ -1524,6 +1538,82 @@ OPENCV_HAL_IMPL_VSX_TRANSPOSE4x4(v_uint32x4, vec_uint4)
 OPENCV_HAL_IMPL_VSX_TRANSPOSE4x4(v_int32x4, vec_int4)
 OPENCV_HAL_IMPL_VSX_TRANSPOSE4x4(v_float32x4, vec_float4)
 
+template<int i>
+inline v_int8x16 v_broadcast_element(v_int8x16 v)
```
seiko2plus: I know its too late, but GCC will not able to optimize such permute like this one into `vsplt[b,h,w]` instructions, intrinsic `vec_splat()` must be used instead. in case of use `vec_splat()` with double word vector, the compiler will use the instruction `xxpermdi` A.K.A `vec_permi()` which is far way better than `vperm`.

alalek: Feel free to propose patch for this.

everton1984: @alalek I think I need to create another PR for this right?

everton1984: oh never mind, @seiko2plus already fixed it. Thanks.

PR #15519, opened 2019-09-13
Windows: partially restore install config generation

mshabunin:
### This pullrequest changes

Windows pack root cmake config script does not support custom layout, it is always assumed to be `$ARCH/$RUNTIME/lib`, thus we can not put cmake files for single configuration to `OPENCV_LIB_INSTALL_PATH`. Partially reverted changes made by #15418, root cmake config mechanism should be updated to support custom cmake files locations.

Also added `OPENCV_SKIP_CMAKE_ROOT_CONFIG` option to force flat cmake config layout just like on UNIX systems.


cmake/OpenCVInstallLayout.cmake
```diff
@@ -23,28 +23,28 @@ if(ANDROID)
 elseif(WIN32 AND CMAKE_HOST_SYSTEM_NAME MATCHES Windows)
 
   if(DEFINED OpenCV_RUNTIME AND DEFINED OpenCV_ARCH)
-    set(_prefix "${OpenCV_ARCH}/${OpenCV_RUNTIME}/")
+    ocv_update(OpenCV_INSTALL_BINARIES_PREFIX "${OpenCV_ARCH}/${OpenCV_RUNTIME}/")
```
alalek: Lets use uppercase variant for all new CMake configurable variables: `OPENCV_INSTALL_BINARIES_PREFIX`

PR #15537, opened 2019-09-18
Support nGraph API

l-bat:
Support new Inference Engine API. 
 Problem layers:
* LRN (problem DNNTestNetwork.Inception_5h, Test_ONNX_nets.ZFNet)
* Deconvolution (only works 2D without bias) 

<cut/>

Failed nets:
:heavy_check_mark: DNNTestNetwork.MobileNet_SSD_Caffe_Different_Width_Height (bad norm) 
:heavy_check_mark: DNNTestNetwork.OpenFace (bad norm) support Power
:heavy_check_mark: Test_Caffe_nets.FasterRCNN_zf (Proposal bug)
:heavy_check_mark: Test_TensorFlow_nets.opencv_face_detector_uint8 (blob not found) -> bad norm
:heavy_check_mark: Test_TensorFlow_nets.EAST_text_detection (maxpool incorrect shape + incorrect net) -> bad norm (support Power)
:interrobang: Test_Torch_nets.FastNeuralStyle_accuracy (DLDT)
:small_red_triangle_down: Test_Torch_nets.ENet_accuracy (bad norm)
:heavy_check_mark: Test_Torch_nets.OpenFace_accuracy (bad norm) support Power
:x: TEST_P(Test_Darknet_nets_async, Accuracy) (layer = nullptr sigseg Region doesn't supported)
After fuse:
:heavy_check_mark: Test_ONNX_nets.DenseNet121 (blob not found) 
:heavy_check_mark: Test_Darknet_nets.YOLOv3  ( bad norm -  DLDT)

:heavy_check_mark: Deconv3D fixed #15572 
:heavy_check_mark: Eltwise with coeffs
:heavy_check_mark: Pooling (default ceil TF)
:heavy_check_mark: Scale (BatchNorm)

Link with IE nGraph:
:small_red_triangle_down: Test_Caffe_nets.FasterRCNN_vgg16 
:heavy_check_mark:  Test_TensorFlow_nets.Faster_RCNN (topK DetectionOutput)

<cut/>

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -276,6 +276,9 @@ OCV_OPTION(WITH_HALIDE "Include Halide support" OFF
 OCV_OPTION(WITH_INF_ENGINE "Include Intel Inference Engine support" OFF
   VISIBLE_IF TRUE
   VERIFY INF_ENGINE_TARGET)
+OCV_OPTION(WITH_NGRAPH "Include nGraph support" WITH_INF_ENGINE
+  VISIBLE_IF TRUE
+  VERIFY TARGET ngraph::ngraph)
 OCV_OPTION(WITH_JASPER "Include JPEG2K support" ON
   VISIBLE_IF NOT IOS
   VERIFY HAVE_JASPER)
@@ -1489,10 +1492,35 @@ if(WITH_INF_ENGINE OR INF_ENGINE_TARGET)
     )
     get_target_property(_inc ${ie_target} INTERFACE_INCLUDE_DIRECTORIES)
     status("    Inference Engine:" "${__msg}")
-    status("                libs:" "${_lib}")
-    status("            includes:" "${_inc}")
+    status("        * libs:" "${_lib}")
+    status("        * includes:" "${_inc}")
   else()
     status("    Inference Engine:"     "NO")
+  endif()
+endif()
+if(WITH_NGRAPH OR HAVE_NGRAPH)
+  if(HAVE_NGRAPH)
+    set(__target ngraph::ngraph)
+    set(__msg "YES (${ngraph_VERSION})")
+    get_target_property(_lib ${__target} IMPORTED_LOCATION)
+    get_target_property(_lib_imp_rel ${__target} IMPORTED_IMPLIB_RELEASE)
+    get_target_property(_lib_imp_dbg ${__target} IMPORTED_IMPLIB_DEBUG)
+    get_target_property(_lib_rel ${__target} IMPORTED_LOCATION_RELEASE)
+    get_target_property(_lib_dbg ${__target} IMPORTED_LOCATION_DEBUG)
+    ocv_build_features_string(_lib
+      IF _lib THEN "${_lib}"
+      IF _lib_imp_rel AND _lib_imp_dbg THEN "${_lib_imp_rel} / ${_lib_imp_dbg}"
+      IF _lib_rel AND _lib_dbg THEN "${_lib_rel} / ${_lib_dbg}"
+      IF _lib_rel  THEN "${_lib_rel}"
+      IF _lib_dbg  THEN "${_lib_dbg}"
+      ELSE "unknown"
+    )
+    get_target_property(_inc ${__target} INTERFACE_INCLUDE_DIRECTORIES)
+    status("    nGraph:" "${__msg}")
+    status("        * libs:" "${_lib}")
+    status("        * includes:" "${_inc}")
+  else()
+    status("    nGraph:"     "NO")
   endif()
 endif()
 
--- a/cmake/OpenCVDetectInferenceEngine.cmake
+++ b/cmake/OpenCVDetectInferenceEngine.cmake
@@ -35,6 +35,15 @@ function(add_custom_ie_build _inc _lib _lib_rel _lib_dbg _msg)
     IMPORTED_IMPLIB_DEBUG "${_lib_dbg}"
     INTERFACE_INCLUDE_DIRECTORIES "${_inc}"
   )
+
+  find_library(ie_builder_custom_lib "inference_engine_nn_builder" PATHS "${INF_ENGINE_LIB_DIRS}" NO_DEFAULT_PATH)
+  if(EXISTS "${ie_builder_custom_lib}")
+    add_library(inference_engine_nn_builder UNKNOWN IMPORTED)
+    set_target_properties(inference_engine_nn_builder PROPERTIES
+      IMPORTED_LOCATION "${ie_builder_custom_lib}"
+    )
+  endif()
+
   if(NOT INF_ENGINE_RELEASE VERSION_GREATER "2018050000")
     find_library(INF_ENGINE_OMP_LIBRARY iomp5 PATHS "${INF_ENGINE_OMP_DIR}" NO_DEFAULT_PATH)
     if(NOT INF_ENGINE_OMP_LIBRARY)
@@ -44,7 +53,12 @@ function(add_custom_ie_build _inc _lib _lib_rel _lib_dbg _msg)
     endif()
   endif()
   set(INF_ENGINE_VERSION "Unknown" CACHE STRING "")
-  set(INF_ENGINE_TARGET inference_engine PARENT_SCOPE)
+  set(INF_ENGINE_TARGET inference_engine)
+  if(TARGET inference_engine_nn_builder)
+    list(APPEND INF_ENGINE_TARGET inference_engine_nn_builder)
+    set(_msg "${_msg}, with IE NN Builder API")
+  endif()
+  set(INF_ENGINE_TARGET "${INF_ENGINE_TARGET}" PARENT_SCOPE)
   message(STATUS "Detected InferenceEngine: ${_msg}")
 endfunction()
 
@@ -54,7 +68,7 @@ find_package(InferenceEngine QUIET)
 if(InferenceEngine_FOUND)
   set(INF_ENGINE_TARGET ${InferenceEngine_LIBRARIES})
   set(INF_ENGINE_VERSION "${InferenceEngine_VERSION}" CACHE STRING "")
-  message(STATUS "Detected InferenceEngine: cmake package")
+  message(STATUS "Detected InferenceEngine: cmake package (${InferenceEngine_VERSION})")
 endif()
 
 if(NOT INF_ENGINE_TARGET AND INF_ENGINE_LIB_DIRS AND INF_ENGINE_INCLUDE_DIRS)
@@ -93,4 +107,16 @@ if(INF_ENGINE_TARGET)
   set_target_properties(${INF_ENGINE_TARGET} PROPERTIES
     INTERFACE_COMPILE_DEFINITIONS "HAVE_INF_ENGINE=1;INF_ENGINE_RELEASE=${INF_ENGINE_RELEASE}"
   )
+endif()
+
+if(WITH_NGRAPH)
+  find_package(ngraph QUIET)
+  if(ngraph_FOUND)
+    ocv_assert(TARGET ngraph::ngraph)
+    if(INF_ENGINE_RELEASE VERSION_LESS "2019039999")
+      message(WARNING "nGraph is not tested with current InferenceEngine version: INF_ENGINE_RELEASE=${INF_ENGINE_RELEASE}")
+    endif()
+    message(STATUS "Detected ngraph: cmake package (${ngraph_VERSION})")
+    set(HAVE_NGRAPH ON)
+  endif()
 endif()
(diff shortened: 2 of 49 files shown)


CMakeLists.txt
```diff
@@ -1595,19 +1595,22 @@ endif()
 
 if(WITH_INF_ENGINE OR INF_ENGINE_TARGET)
   if(INF_ENGINE_TARGET)
+    list(GET INF_ENGINE_TARGET 0 ie_target)
```
alalek: Why is these changes here? They should be on the 3.4 branch already: #15376

Please rebase patch.

modules/dnn/src/layers/convolution_layer.cpp
```diff
@@ -528,6 +530,111 @@ class ConvolutionLayerImpl CV_FINAL : public BaseConvolutionLayerImpl
...
+                          ngraph::element::f16 : ngraph::element::f32;
+
+        std::vector<size_t> kernel_shape(&blobs[0].size[0], &blobs[0].size[0] + blobs[0].dims);
+        Mat halfs(1, blobs[0].total(), CV_16SC1);
+        if (precision == ngraph::element::f16) {
+            convertFp16(blobs[0], halfs);
+        }
+
+        auto ieWeights = std::make_shared<ngraph::op::Constant>(precision, kernel_shape,
+                         precision == ngraph::element::f16 ? halfs.data : blobs[0].data);
```
dkurt: Maybe we can create some method which can wrap OpenCV's Mat to nGraph's Constant op? Like `wrapToInfEngineBlob`. It will convert to FP16 is necessary.

modules/dnn/src/layers/fully_connected_layer.cpp
```diff
@@ -456,6 +459,44 @@ class FullyConnectedLayerImpl CV_FINAL : public InnerProductLayer
...
+#ifdef HAVE_INF_ENGINE
+    virtual Ptr<BackendNode> initNgraph(const std::vector<Ptr<BackendWrapper> >& inputs,
+                                        const std::vector<Ptr<BackendNode> >& nodes) CV_OVERRIDE
+    {
+        auto& ieInpNode = nodes[0].dynamicCast<InfEngineNgraphNode>()->node;
+        auto batch = ieInpNode->get_shape()[0];
+
+        std::vector<size_t> data = {(size_t)batch, (size_t)blobs[0].size[1]};
+        auto new_shape = std::make_shared<ngraph::op::Constant>(ngraph::element::i64, ngraph::Shape{2}, data.data());
+        auto inp = std::make_shared<ngraph::op::v1::Reshape>(ieInpNode, new_shape, true);
```
dkurt: Will it work with `batch size > 1`? Maybe we need reshape it to `batch x product of the rest of dims`

modules/dnn/src/layers/fully_connected_layer.cpp
```diff
@@ -456,6 +459,44 @@ class FullyConnectedLayerImpl CV_FINAL : public InnerProductLayer
...
+            auto bias_node = std::make_shared<ngraph::op::Constant>(ngraph::element::f32, ngraph::Shape{(size_t)blobs[1].size[1]}, blobs[1].data);
+            std::shared_ptr<ngraph::op::Add> fc;
+            if (precisionFP16) {
+                auto bias_fp16 = std::make_shared<ngraph::op::Convert>(bias_node, ngraph::element::f16);
+                fc = std::make_shared<ngraph::op::Add>(matmul, bias_fp16, ngraph::op::AutoBroadcastType::NUMPY);
+            } else {
+                fc = std::make_shared<ngraph::op::Add>(matmul, bias_node, ngraph::op::AutoBroadcastType::NUMPY);
+            }
+            return Ptr<BackendNode>(new InfEngineNgraphNode(fc));
+        }
```
dkurt: Please check FC layer stability without biases. It gives Segfault for me. Probably, we need a workaround with zeroed biases this way.

cmake/OpenCVDetectInferenceEngine.cmake
```diff
@@ -94,3 +94,15 @@ if(INF_ENGINE_TARGET)
...
+  if(INF_ENGINE_RELEASE VERSION_LESS "2019040000")
+    message(WARNING "nGraph is not tested with InferenceEngine versions prior 2019R4")
+  endif()
+  find_package(ngraph QUIET)
+  if(ngraph_FOUND)
+    ocv_assert(TARGET ngraph::ngraph)
+    message(STATUS "Detected ngraph: cmake package (${ngraph_VERSION})")
+    set(HAVE_NGRAPH ON)
+  endif()
+endif()
```
dkurt: Please also add the following changes to this file: https://github.com/opencv/opencv/commit/087824e57e11fd29ef1caa30334a387348403024#diff-da800785713bed07d12ee19660fa794a (please ignore the changes at modules/dnn/CMakeLists.txt).

It's necessary to use custom location of Inference Engine.

modules/dnn/src/dnn.cpp
```diff
@@ -729,14 +729,41 @@ struct DataLayer : public Layer
...
+                                           const std::vector<Ptr<BackendNode> >& nodes) CV_OVERRIDE
     {
 #ifdef HAVE_INF_ENGINE
         CV_CheckEQ(inputsData.size(), (size_t)1, "");
         CV_CheckEQ(inputsData[0].dims, 4, "");
         const size_t numChannels = inputsData[0].size[1];
         CV_Assert(numChannels <= 4);
 
+#ifdef HAVE_DNN_NGRAPH
+        if (DNN_BACKEND_INFERENCE_ENGINE_NGRAPH == getInferenceEngineBackendTypeParam())
```
alalek: 1. Using `getInferenceEngineBackendTypeParam()` is not correct - it might be changed on the fly. `== preferableBackend` should be used instead, but it is not exposed for Layers.
2. `#ifdef HAVE_DNN_NGRAPH` should be used.

modules/dnn/src/layers/batch_norm_layer.cpp
```diff
@@ -358,9 +358,9 @@ class BatchNormLayerImpl CV_FINAL : public BatchNormLayer
     virtual Ptr<BackendNode> initInfEngine(const std::vector<Ptr<BackendWrapper> >&,
                                            const std::vector<Ptr<BackendNode> >& nodes) CV_OVERRIDE
     {
-        if (DNN_BACKEND_INFERENCE_ENGINE_NGRAPH == getInferenceEngineBackendTypeParam())
-            return initNgraph(nodes);
-
+#ifdef HAVE_DNN_NGRAPH
+        return initNgraph(nodes);
+#endif  // HAVE_DNN_NGRAPH
         InferenceEngine::Builder::Layer ieLayer = InferenceEngine::Builder::ScaleShiftLayer(name);
```
alalek: How this can work without runtime check?
Please run tests locally.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -278,6 +286,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
 
         virtual Ptr<BackendNode> initInfEngine(const std::vector<Ptr<BackendWrapper> > &inputs);
 
+        virtual Ptr<BackendNode> initNgraph(const std::vector<Ptr<BackendWrapper> > &inputs, const std::vector<Ptr<BackendNode> >& nodes);
```
alalek: Lets keep it here for this first iteration (keep code separate for easy regressions investigations)

modules/dnn/src/dnn.cpp
```diff
@@ -2575,11 +2919,17 @@ Net Net::readFromModelOptimizer(const String& xml, const String& bin)
...
+
+#if INF_ENGINE_VER_MAJOR_LE(INF_ENGINE_RELEASE_2019R3)
     InferenceEngine::CNNNetReader reader;
     reader.ReadNetwork(xml);
     reader.ReadWeights(bin);
 
     InferenceEngine::CNNNetwork ieNet = reader.getNetwork();
+#else
+    InferenceEngine::Core& ie = getCore();
+    InferenceEngine::CNNNetwork ieNet = ie.ReadNetwork(xml, bin);
```
alalek: @dkurt Please review this part of code.

modules/dnn/src/ie_ngraph.cpp
```diff
@@ -0,0 +1,681 @@
...
+            break;
+        case DNN_TARGET_FPGA:
+            device_name = "FPGA";
+            break;
+        default:
+            CV_Error(Error::StsNotImplemented, "Unknown target");
+    };
+
+    if (cnn.getPrecision() != InferenceEngine::Precision::FP16 && targetId == DNN_TARGET_MYRIAD)
+        resetMyriadDevice();
```
alalek: Why is it here?

modules/dnn/src/ie_ngraph.cpp
```diff
@@ -0,0 +1,681 @@
...
+            itr->second = currentInShape;
+            equal_flag = false;
+        }
+        i++;
+    }
+
+    if (!equal_flag)
+    {
+        InferenceEngine::CNNNetwork curr_t_net(t_net);
+        curr_t_net.reshape(inShapes);
```
alalek: Why is ".get()" method with "const" logical modifier doing some network reshaping?

modules/dnn/src/dnn.cpp
dkurt: @l-bat, Can we keep `ReadNetwork` and `ReadWeights` for now? I think it'd be more stable considering coming https://github.com/opencv/opencv/pull/16015.

alalek: We can, but `CNNNetReader` is almost deprecated.

modules/dnn/src/ie_ngraph.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+    static Ptr<BackendWrapper> create(Ptr<BackendWrapper> wrapper);
+
+    virtual void copyToHost() CV_OVERRIDE;
+    virtual void setHostDirty() CV_OVERRIDE;
+
+    InferenceEngine::DataPtr dataPtr;
+    InferenceEngine::Blob::Ptr blob;
+    AsyncArray futureMat;
+};
```
dkurt: Can we merge `NgraphBackendWrapper` and Inference Engine's wrapper?

alalek: I believe it is better to "replace" it completely instead (after removal of NN Builder API).

I prefer to keep them separate - it least to guarantee that we don't break NN Builder API for now.

dkurt: Sure, thank you!

modules/dnn/src/ie_ngraph.cpp
dkurt: It looks like a workaround introduced at https://github.com/opencv/opencv/pull/13608 to dynamically reshape networks from OpenVINO's IRs in case of different input shape.

modules/dnn/src/layers/resize_layer.cpp
```diff
@@ -52,11 +57,15 @@ class ResizeLayerImpl : public ResizeLayer
...
     {
 #ifdef HAVE_INF_ENGINE
-        if (backendId == DNN_BACKEND_INFERENCE_ENGINE)
+        if (backendId == DNN_BACKEND_INFERENCE_ENGINE_NN_BUILDER_2019)
         {
             return (interpolation == "nearest" && scaleWidth == scaleHeight) ||
                    (interpolation == "bilinear");
         }
+        if (backendId == DNN_BACKEND_INFERENCE_ENGINE_NGRAPH) {
+            return (interpolation == "nearest" && scaleWidth == scaleHeight && scaleWidth == 0.5) ||
```
dkurt: That means only x2 nearest neighbor downsampling is supported?

modules/dnn/test/test_ie_models.cpp
```diff
@@ -310,10 +324,10 @@ TEST_P(DNNTestOpenVINO, models)
     std::map<std::string, cv::Mat> inputsMap;
     std::map<std::string, cv::Mat> ieOutputsMap, cvOutputsMap;
     // Single Myriad device cannot be shared across multiple processes.
-    if (target == DNN_TARGET_MYRIAD)
+    if (targetId == DNN_TARGET_MYRIAD)
         resetMyriadDevice();
-    runIE(target, xmlPath, binPath, inputsMap, ieOutputsMap);
-    runCV(target, xmlPath, binPath, inputsMap, cvOutputsMap);
+    runIE(targetId, xmlPath, binPath, inputsMap, ieOutputsMap);
+    runCV(backendId, targetId, xmlPath, binPath, inputsMap, cvOutputsMap);
```
dkurt: Why we iterate over backends? This tests is only for OpenVINO models that means only IE backend is supported.

modules/dnn/src/layers/resize_layer.cpp
l-bat: Yes

PR #15541, opened 2019-09-18
Added type deduced v_setall intrinsic

terfendail:
### This pullrequest changes
Added type deduced v_setall intrinsic

<cut/>

```
force_builders=Custom
docker_image:Custom=ubuntu-clang:18.04
```


terfendail:
v_setall_##<short_type_id>(type val) intrinsics are declared the way it is possible to deduce short_type_id from the input value type. So it could be convenient to use uniform named version.

alalek:
What is about a variant via template? (would help with proper forcing of used types)

terfendail:
Corresponding low-level intrinsics have different names for different types(at least for x86 SIMD) so it looks like it would be necessary to specialize the template for every type. Existing implementation will  IMO also force correctness of input type because it will define the output vector type that can't be implicitly casted to another vector type if there is a mismatch. The change make impossible(or essentially harder) to add implicit vector type casting, but I doubt whether we need such casting.

alalek:
Please add corresponding intrinsic test

alalek:
@terfendail Could you take a look on this patch: #15839 ?

asmorkalov:
@terfendail, @alalek  Do you have aby progress on the patch?

terfendail:
Replaced by #15839

PR #15544, opened 2019-09-19
Disable posix_memalign by default

### LINKED ISSUE #15526: memory leak / fragmentation issue

- OpenCV => 4.1.1
- Compiler => GCC 7.4.0

##### Detailed description

When creating large Mat objects followed by smaller allocations that are kept in a loop, resident memory keeps growing. This does not occur when the matrices are smaller or an allocation does not happen in between. If left to keep growing, the system will run out of memory and lock up, even though no more than 2000 small strings and 2 matrices should be in memory.

I've tried with OpenCV 3.4 as well, and the same thing happens

##### Steps to reproduce

compile the following, run it and keep an eye on memory.

```.cpp
#include <thread>
#include <utility>
#include <vector>
#include <opencv2/core/mat.hpp>

int main() {
  
    // minimum size for leak on my machine: 558995 x 1
    cv::Size size(600000, 5);
    cv::Mat tmp(size, CV_8UC3, cv::Scalar(0, 0, 0));

    std::vector<std::string> strings;

    for (size_t i = 0; i < 2000; ++i)
    {
      cv::Mat tmp2(size, CV_8UC3, cv::Scalar(0, 0, 0));
      tmp = tmp2;

      // needs to be suitably large, so it gets allocated on the heap
      std::string val("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa");
      strings.push_back(val);

      // gives you time to kill the process before it runs out of memory
      std::this_thread::sleep_for(std::chrono::milliseconds(10));
    }
}

```

update:

this appears to be an issue with the usage of posix_memalign: if I remove the checks for it in the CMakeLists.txt and remove the checks for it then recompile, the leak disappears. Strangely, just allocating with posix_memalign instead of using cv::Mat does not create the issue on its own, there seems to be some interaction with the other operations cv::Mat does when creating it.

mshabunin:
resolves #15526

### This pullrequest changes

Do not use posix_memalign due to possible problems with memory fragmentation in glibc.

Build CMake's option:
* `OPENCV_ENABLE_MEMALIGN` - enabled by default on *NIX

Runtime environment parameter:
* `OPENCV_ENABLE_MEMALIGN` - disabled by default (Linux + glibc)

valgrind, ASAN and other sanitizers runs must have `OPENCV_ENABLE_MEMALIGN=1` for reliable results.


alalek:
It should be enabled by default.
Custom alignment hacks block detection of out of buffer access via valgrind or sanitizers.

mshabunin:
There is [severe issue in glibc](https://sourceware.org/bugzilla/show_bug.cgi?id=14581) which affects image processing use cases: repeated allocation/deallocation of several large aligned buffers interleaved with smaller allocations cause memory exhaust.

See other examples in the related issue. I think we should disable it by default and enable only for dynamic analysis.

Reproducer from the issue attachments:
```
#include <stdlib.h>
#include <string.h>
#include <unistd.h>

#define ALIGNMENT        64               // 16 works, 32 fails, 64 fails, 128 works, 256 fails
#define LARGE_ALLOC_SIZE 16 * 1024 * 1024 // 1, 2, 4, 8, 16 fail, 32, 64, 128 work (all in MB)
#define SMALL_ALLOC_SIZE 25               // <= 24 works, > 24 breaks

void* work() {
    // Temporary buffers for the computation
    void* tmp1 = aligned_alloc(ALIGNMENT, LARGE_ALLOC_SIZE);
    void* tmp2 = aligned_alloc(ALIGNMENT, LARGE_ALLOC_SIZE);

    // Ensure memory is actually mapped
    memset(tmp1, 0, LARGE_ALLOC_SIZE);
    memset(tmp2, 0, LARGE_ALLOC_SIZE);

    // Buffer for the result
    void* result = malloc(SMALL_ALLOC_SIZE);

    free(tmp2);
    free(tmp1);

    return result;
}

int main() {
    for (;;) {
        usleep(1000);          // To make it easier to observe
        void* result = work(); // Normally the result is stored
    }
}
```
Graph of RSS/VSZ (~12 sec):
image

mshabunin:
For correct dynamic analysis results with custom alignment we can leverage [manual poisoning in ASAN](https://github.com/google/sanitizers/wiki/AddressSanitizerManualPoisoning) and [special macros for valgrind](http://valgrind.org/docs/manual/manual-core-adv.html#manual-core-adv.clientreq).

CMakeLists.txt
```diff
@@ -635,10 +636,12 @@ if(UNIX)
     set(HAVE_PTHREAD 1)
   endif()
 
-  CHECK_SYMBOL_EXISTS(posix_memalign stdlib.h HAVE_POSIX_MEMALIGN)
-  CHECK_INCLUDE_FILE(malloc.h HAVE_MALLOC_H)
-  if(HAVE_MALLOC_H)
-    CHECK_SYMBOL_EXISTS(memalign malloc.h HAVE_MEMALIGN)
+  if(ENABLE_ALIGNED_MALLOC)
```
alalek: Parameter name mismatch with `OCV_OPTION()`.

modules/core/src/alloc.cpp
```diff
@@ -81,6 +82,10 @@ cv::utils::AllocatorStatisticsInterface& getAllocatorStatistics()
     return allocator_stats;
 }
 
+#if defined HAVE_POSIX_MEMALIGN || defined HAVE_MEMALIGN
+static const bool useMemalign = cv::utils::getConfigurationParameterBool("OPENCV_ENABLE_MEMALIGN", false);
```
alalek: We should avoid "initialization order fiasco".

Variable can be used below without proper initialization and then crash in mismatched `fastFree()` calls.

PR #15555, opened 2019-09-20
Vectorize flipHoriz and flipVert functions.

ChipKerchner:
Vectorize flipHoriz and flipVert functions.

flipVert - 2x faster on VSX and 1.5x on x86
flipHoriz - up to 40x+faster on VSX and x86.

Also improved for NEON.

```
buildworker:Custom=linux-3
disable_ipp=ON
```

### DIFF
--- a/modules/core/src/copy.cpp
+++ b/modules/core/src/copy.cpp
@@ -576,9 +576,7 @@ CV_ALWAYS_INLINE v_uint8x16 v_load_mirror_1( const uchar* ptr )
     static const __m128i perm = _mm_setr_epi8(15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0);
     return v_uint8x16(_mm_shuffle_epi8(_mm_loadu_si128((__m128i*)(ptr)), perm));
 #elif CV_VSX
-    static const vec_uchar16 perm = {15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0};
-    vec_uchar16 vec = vsx_ld(0, ptr);
-    return v_uint8x16(vec_perm(vec, vec, perm));
+    return v_uint8x16(vec_uchar16_c(vec_revb(vsx_ld(0, (const unsigned __int128 *)ptr))));
 #elif CV_NEON
     uint8x16_t vec = vrev64q_u8(vld1q_u8(ptr));
     return v_uint8x16(vextq_u8(vec, vec, 8));
--- a/modules/core/src/copy.cpp
+++ b/modules/core/src/copy.cpp
@@ -585,14 +585,32 @@ template<typename V> CV_ALWAYS_INLINE void flipHoriz_single( const uchar* src, s
             v_store((T*)(dst + j - v_uint8x16::nlanes), t0);
             v_store((T*)(dst + i), t1);
         }
-        for ( ; i < width; i += sizeof(T), j -= sizeof(T) )
+        if (((size_t)src|(size_t)dst) % sizeof(T) == 0)
         {
-            T t0, t1;
+            for ( ; i < width; i += sizeof(T), j -= sizeof(T) )
+            {
+                T t0, t1;
 
-            t0 = *((T*)((uchar*)src + i));
-            t1 = *((T*)((uchar*)src + j - sizeof(T)));
-            *((T*)(dst + j - sizeof(T))) = t0;
-            *((T*)(dst + i)) = t1;
+                t0 = *((T*)((uchar*)src + i));
+                t1 = *((T*)((uchar*)src + j - sizeof(T)));
+                *((T*)(dst + j - sizeof(T))) = t0;
+                *((T*)(dst + i)) = t1;
+            }
+        }
+        else
+        {
+            for ( ; i < width; i += sizeof(T), j -= sizeof(T) )
+            {
+                for (int k = 0; k < (int)sizeof(T); k++)
+                {
+                    uchar t0, t1;
+
+                    t0 = *((uchar*)src + i + k);
+                    t1 = *((uchar*)src + j + k - sizeof(T));
+                    *(dst + j + k - sizeof(T)) = t0;
+                    *(dst + i + k) = t1;
+                }
+            }
         }
     }
 }
@@ -760,18 +778,18 @@ flipVert( const uchar* src0, size_t sstep, uchar* dst0, size_t dstep, Size size,
                                                   dst0 += dstep, dst1 -= dstep )
     {
         int i = 0;
+#if CV_SIMD
+        for( ; i <= size.width - (v_int32::nlanes * 4); i += v_int32::nlanes * 4 )
+        {
+            v_int32 t0 = vx_load((int*)(src0 + i));
+            v_int32 t1 = vx_load((int*)(src1 + i));
+            vx_store((int*)(dst0 + i), t1);
+            vx_store((int*)(dst1 + i), t0);
+        }
+#endif
+
         if( ((size_t)src0|(size_t)dst0|(size_t)src1|(size_t)dst1) % sizeof(int) == 0 )
         {
-#if CV_SIMD
-            for( ; i <= size.width - (v_int32::nlanes * 4); i += v_int32::nlanes * 4 )
-            {
-
-                v_int32 t0 = vx_load((int*)(src0 + i));
-                v_int32 t1 = vx_load((int*)(src1 + i));
-                vx_store((int*)(dst0 + i), t1);
-                vx_store((int*)(dst1 + i), t0);
-            }
-#else
             for( ; i <= size.width - 16; i += 16 )
             {
                 int t0 = ((int*)(src0 + i))[0];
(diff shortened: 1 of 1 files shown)


ChipKerchner:
Need someone to review the NEON code.

modules/core/src/copy.cpp
```diff
@@ -599,6 +867,26 @@ flipVert( const uchar* src0, size_t sstep, uchar* dst0, size_t dstep, Size size,
         int i = 0;
         if( ((size_t)src0|(size_t)dst0|(size_t)src1|(size_t)dst1) % sizeof(int) == 0 )
         {
+#if CV_SIMD
+#if CV_SIMD256 || CV_SIMD512
```
terfendail: Does tail processing make essential difference? It's possible to process half nlanes elements by `vx_load_low` `vx_store_low` in this case to reduce the tail size. Also check should be `CV_SIMD_WIDTH > 16` in this case to ensure everything will be fine if someday `CV_SIMD1024` appears

modules/core/src/copy.cpp
```diff
@@ -564,24 +564,292 @@ Mat& Mat::setTo(InputArray _value, InputArray _mask)
 }
 
 
+#if CV_SIMD
+CV_ALWAYS_INLINE v_uint8x16 v_load_mirror_1( const uchar* ptr );
```
terfendail: This functionality could be implemented via `v_lut` intrinsic. While I understand that shuffling should perform essentially faster and like the idea I prefer to retain platform specific functionality as low as possible. IMO it would be better to add new universal intrinsic. Something like `v_reverse`(that could internally fallback to `v_lut` for platforms that unable/not yet managed to implement this functionality).

modules/core/src/copy.cpp
ChipKerchner: Simplifying into native vx_load/vx_store case.

modules/core/src/copy.cpp
ChipKerchner: Using v_lut to implement this would significantly slow it down.  I will look into creating a new v_reverse function.

tomoaki0705:
I don't get why these `v_load_mirror` series goes in to universal intrinsic.
That's the place to write those wrapper functions.

ChipKerchner:
I've prepared a change for an universal intrinsics ```v_reverse``` that addresses the above requests. [15662](https://github.com/opencv/opencv/pull/15662)

ChipKerchner:
Could this PR be looked at and approved now that v_reverse works for all vector platforms?

modules/core/src/copy.cpp
```diff
@@ -564,24 +564,292 @@ Mat& Mat::setTo(InputArray _value, InputArray _mask)
...
+}
+
+CV_ALWAYS_INLINE v_uint8x16 v_load_mirror_8( const uchar* ptr )
+{
+#if CV_SSE2
+    return v_uint8x16(_mm_shuffle_epi32(_mm_loadu_si128((__m128i*)(ptr)), (2 << 0) | (3 << 2) | (0 << 4) | (1 << 6)));
+#elif CV_VSX
+    static const vec_uchar16 perm = {8, 9, 10, 11, 12, 13, 14, 15, 0, 1, 2, 3, 4, 5, 6, 7};
+    vec_uchar16 vec = vsx_ld(0, ptr);
+    return v_uint8x16(vec_perm(vec, vec, perm));
```
pmur: Is this the same as xxswapd (e.g vec_permxxdi(vec,vec,2))?

ChipKerchner: Outdated code.

modules/core/src/copy.cpp
```diff
@@ -599,6 +762,16 @@ flipVert( const uchar* src0, size_t sstep, uchar* dst0, size_t dstep, Size size,
         int i = 0;
         if( ((size_t)src0|(size_t)dst0|(size_t)src1|(size_t)dst1) % sizeof(int) == 0 )
         {
+#if CV_SIMD
+            for( ; i <= size.width - (v_int32::nlanes * 4); i += v_int32::nlanes * 4 )
```
alalek: SIMD code doesn't require alignment check above, so it can be placed before `if`.

mshabunin:
Looks like there is a problem with unaligned v_load on ARM platform:
```
#0  cv::hal_baseline::v_load (ptr=0x7effe83c) at ../opencv/modules/core/include/opencv2/core/hal/intrin_neon.hpp:1212
#1  0x76829e7e in cv::flipHoriz_single<cv::hal_baseline::v_uint64x2> (esz=8, size=..., dstep=32, dst=0x7effe83c "", 
    sstep=32, src=0x7effe83c "") at ../opencv/modules/core/src/copy.cpp:581
#2  cv::flipHoriz (src=0x7effe83c "", sstep=32, dst=0x7effe83c "", dstep=32, size=..., esz=8)
    at ../opencv/modules/core/src/copy.cpp:698
#3  0x7682b2f0 in cv::flip (_src=..., _dst=..., flip_mode=1) at ../opencv/modules/core/src/copy.cpp:977
#4  0x76bc2e72 in cv::intersectConvexConvex (_p1=..., _p2=..., _p12=..., handleNested=true)
    at ../opencv/modules/imgproc/src/geometry.cpp:530
#5  0x004ead06 in opencv_test::(anonymous namespace)::Imgproc_IntersectConvexConvex_intersection_1_Test::Body (
    this=0x8bb830) at ../opencv/modules/imgproc/test/test_intersectconvexconvex.cpp:161
#6  0x004eab5c in opencv_test::(anonymous namespace)::Imgproc_IntersectConvexConvex_intersection_1_Test::TestBody (
    this=0x8bb830) at ../opencv/modules/imgproc/test/test_intersectconvexconvex.cpp:146
<cut>
```

Can be reproduced by running this test case: `./bin/opencv_test_imgproc --gtest_filter=Imgproc_IntersectConvexConvex.intersection_1`

modules/core/src/copy.cpp
```diff
@@ -563,25 +563,206 @@ Mat& Mat::setTo(InputArray _value, InputArray _mask)
...
+    int i, j;
+
+    for( ; size.height--; src += sstep, dst += dstep )
+    {
+        for( i = 0, j = end; i < width_1; i += v_uint8x16::nlanes, j -= v_uint8x16::nlanes )
+        {
+            V t0, t1;
+
+            t0 = v_load((T*)((uchar*)src + i));
+            t1 = v_load((T*)((uchar*)src + j - v_uint8x16::nlanes));
```
alalek: > v_uint8x16::nlanes

Looks like, there is hardcoded type in `template<typename V>` function.

ChipKerchner: I am unable to reproduce it

mshabunin: Can be reproduced with ENABLE_BUILD_HARDENING option ON:
```
cmake \
-GNinja   \
-DCMAKE_TOOLCHAIN_FILE=../opencv/platforms/linux/arm-gnueabi.toolchain.cmake   \
-DCMAKE_BUILD_TYPE=Debug   \
-DBUILD_LIST=ts,imgproc   \
-DCMAKE_INSTALL_PREFIX=install   \
-DENABLE_NEON=ON   \
-DCPU_BASLINE=NEON \
-DENABLE_BUILD_HARDENING=ON   \
../opencv
```

ChipKerchner: Since ARM does NOT have an unaligned vector load would the best solution be?

```
diff --git a/modules/core/src/copy.cpp b/modules/core/src/copy.cpp
index 3f68a2555a..5e5f7ea025 100644
--- a/modules/core/src/copy.cpp
+++ b/modules/core/src/copy.cpp
@@ -574,7 +574,9 @@ template<typename V> CV_ALWAYS_INLINE void flipHoriz_single( const uchar* src, s

     for( ; size.height--; src += sstep, dst += dstep )
     {
-        for( i = 0, j = end; i < width_1; i += v_uint8x16::nlanes, j -= v_uint8x16::nlanes )
+        i = 0, j = end;
+#if !CV_NEON
+        for( ; i < width_1; i += v_uint8x16::nlanes, j -= v_uint8x16::nlanes )
         {
             V t0, t1;

@@ -585,6 +587,7 @@ template<typename V> CV_ALWAYS_INLINE void flipHoriz_single( const uchar* src, s
             v_store((T*)(dst + j - v_uint8x16::nlanes), t0);
             v_store((T*)(dst + i), t1);
         }
+#endif
         if (((size_t)src|(size_t)dst) % sizeof(T) == 0)
         {
             for ( ; i < width; i += sizeof(T), j -= sizeof(T) )
```

tomoaki0705:
I did some quick experiments. https://github.com/tomoaki0705/unalignedLoad
@ChipKerchner is right.  There is no "unaligned load instruction" in NEON.
Still, I figured out that there is a corner case in the combination of Armv7 + 64bit loading.

As @mshabunin  showed, the error comes when the load happens with `v_uint64x2`

```
#1  0x76829e7e in cv::flipHoriz_single<cv::hal_baseline::v_uint64x2>
```

Also, it happens on Armv7 (32bit) but not on Arm v8 (64bit)

So, when loading with unaligned address using v_uint64x2 on Armv7, `SIGBUS` raises.
I can't believe this, but as far as I tried, the result shows above result.

Armv8 Jetson TX1 (GCC 5.4.0) + uint64x2 + unaligned = success
Armv8 Jetson TX1 (GCC 5.4.0) + uint64x2 + aligned = success
Armv7 ODROID-XU4 (GCC 5.4.0) + uint64x2 + unaligned = failure
Armv7 ODROID-XU4 (GCC 5.4.0) + uint64x2 + aligned = success
Armv7 ODROID-XU4 (GCC 5.4.0) + uint32x4 + unaligned = success

How to fix, it's a difficult question.

Issue #15574, opened 2019-09-23
SIGABRT by cv::Exception on ocl.cpp:4908 -> CV_Assert(u->origdata == data);


TieSKey:
I'm having a lot of intermittent buy frequent aborts raised by this assertion:

https://github.com/opencv/opencv/blob/47007224445af3dce8dadb11174df14d81fd5a34/modules/core/src/ocl.cpp#L4908

inside a loop doing face detection and tracking. It usually happens on the tracking part, FaceEngine.hpp:135

I'm not sure if it's a real bug or a miss use on my side but some sort of race condition is involved on opencv/ocl library/driver side since it is not exactly 100% reproducible (it's somewhere around 70~50%) and depends on run modes and input timming.

You can find attached the outputs of clinfo, getBuildInformation() and a sample project.
[opencv-repro.tar.gz](https://github.com/opencv/opencv/files/3643700/opencv-repro.tar.gz)
[clinfo.txt](https://github.com/opencv/opencv/files/3643702/clinfo.txt)
[getBuildInformation.txt](https://github.com/opencv/opencv/files/3643703/getBuildInformation.txt)

To reproduce the error, run the sample using the "StaticFrameProvider" (as provided). Plz try several times alternating between "run" and "debug" modes if using an IDE (I'm using CLion). Although it fails too if make and run from the command line.

You can also change main.cpp line 36 with the "OpenCvFrameProvider" that uses opencv::VideoCapture. 
When using VideoCapture, the behavior becomes even more weird on an IDE, as launching the program in debug mode will most likely (but not always) result in an abort the first time a cv::Tracker is updated. Otoh if launch in "run" mode, it will only result in an abort if the very first frame contains a face but not if a face appears later in front of the camera.

I tried making copies and clones of the "colorFrame" variable (lenna image or frame from cv::VideoCapture over a webcam) before passing it to the Tracker but it doesn't help much.

alalek:
Mesa OpenCL implementation - known issue of `CL_MEM_USE_HOST_PTR` implementation: #6293

Try OpenCL runtime from AMDGPU package (from AMD site or from your system distribution)

TieSKey:
Thanks for the info.
Amdgpu only works on ubuntu 18.4 but if that is the only choice I'll have to downgrade. 

Have you identified other platforms/drivers where this happens?
Do you know if there's a bug report filled with mesa?

Have you considered adding a second string param to CV_Assert so you can add some information,  links to known issues or wiki entries, etc? 
[...]

Issue #6293, opened 2016-03-17
opencv_perf_objdetect "OCL_Cascade_*" tests fail since commit ad70ab40


sander1234567890:
When running opencv_perf_objdetect all haarcascade tests fail with the exception below.
This was introduced in commit ad70ab40, if i checkout 1 commit before that the tests run fine.

Time compensation is 0
CTEST_FULL_OUTPUT
OpenCV version: 3.1.0-dev
OpenCV VCS version: 3.1.0-315-gecf65a0-dirty
Build type: release
Parallel framework: pthreads
CPU features: mmx sse sse2 sse3
OpenCL Platforms:
    Clover
        iGPU: AMD TAHITI (DRM 2.43.0, LLVM 3.7.1) (OpenCL 1.1 MESA 11.1.2)
Current OpenCL device:
    Type = iGPU
    Name = AMD TAHITI (DRM 2.43.0, LLVM 3.7.1)
    Version = OpenCL 1.1 MESA 11.1.2
    Compute units = 28
    Max work group size = 256
    Local memory size = 32 kB
    Max memory allocation size = 256 MB
    Double support = Yes
    Host unified memory = Yes
    Has AMD Blas = No
    Has AMD Fft = No
    Preferred vector width char = 16
    Preferred vector width short = 8
    Preferred vector width int = 4
    Preferred vector width long = 2
    Preferred vector width float = 4
    Preferred vector width double = 2
[==========] Running 28 tests from 2 test cases.
[----------] Global test environment set-up.

[ RUN      ] OCL_Cascade_Image_MinSize_CascadeClassifier.CascadeClassifier/3
OpenCV Error: Assertion failed (u->origdata == data) in deallocate, file /usr/src/opencv/modules/core/src/ocl.cpp, line 4584
/usr/src/opencv/modules/ts/src/ts_perf.cpp:1721: Failure
Failed
Expected: PerfTestBody() doesn't throw an exception.
  Actual: it throws cv::Exception:
  /usr/src/opencv/modules/core/src/ocl.cpp:4584: error: (-215) u->origdata == data in function deallocate

params    = ("cv/cascadeandhog/cascades/haarcascade_frontalface_alt.xml", "cv/cascadeandhog/images/bttf301.png", 30)
termination reason:  unhandled exception
bytesIn   =     240950
bytesOut  =          0
samples   =         10 of 100
outliers  =          0
frequency = 1000000000
min       =   16528436 = 16.53ms
[...]

PR #15579, opened 2019-09-24
core(OpenCL): add flag OPENCV_OPENCL_ENABLE_MEM_USE_HOST_PTR

alalek:
to control CL_MEM_USE_HOST_PTR usage

work around for #6293 #15574


mshabunin:
Perhaps we should add a message mentioning new environment variables to `CV_Assert(u->origdata == data);`. Something like: "Try to set OPENCV_OPENCL_ENABLE_MEM_USE_HOST_PTR=0"

PR #15580, opened 2019-09-24
dnn: Allow LSTM layer to operate in reverse direction

smbz:
This can be used to produce a bidirectional LSTM by having two operating in parallel on the same input, but operating in different directions.

### This pullrequest changes

This PR lets cv::dnn::LSTMLayer interpret a new boolean parameter in the LayerParams called "reverse".  If it is true, the LSTM operates in the reverse direction to normal.

### DIFF
--- a/modules/dnn/test/test_layers.cpp
+++ b/modules/dnn/test/test_layers.cpp
@@ -489,6 +489,55 @@ TEST(Layer_RNN_Test_Accuracy_with_, CaffeRecurrent)
     normAssert(h_ref, output[0]);
 }
 
+TEST(Layer_LSTM_Test_Accuracy_, Reverse)
+{
+    // This handcrafted setup calculates (approximately) the prefix sum of the
+    // input, assuming the inputs are suitably small.
+    cv::Mat input(2, 1, CV_32FC1);
+    input.at<float>(0, 0) = 1e-5f;
+    input.at<float>(1, 0) = 2e-5f;
+
+    cv::Mat Wx(4, 1, CV_32FC1);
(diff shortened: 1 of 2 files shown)


dkurt:
@smbz, thanks for contribution!

Please change the target branch to 3.4 (we will cherry pick this commit to master after that).

smbz:
You're welcome!  Changed to 3.4 and squashed to a single commit (now that I've read the contribution guidelines properly :P )

modules/dnn/test/test_layers.cpp
```diff
@@ -489,6 +489,51 @@ TEST(Layer_RNN_Test_Accuracy_with_, CaffeRecurrent)
...
+    cv::Mat bias(4, 1, CV_32FC1);
+    bias.at<float>(0, 0) = 1e10f;  // Input gate - always allows input to c
+    bias.at<float>(1, 0) = 1e10f;  // Forget gate - never forget anything on c
+    bias.at<float>(2, 0) = 1e10f;  // Output gate - always output everything
+    bias.at<float>(3, 0) = 0.f;  // Update signal
+
+    LayerParams lp;
+    lp.set("reverse", true);
+    lp.set("use_timestamp_dim", true);
+    lp.blobs = std::vector<cv::Mat>{ Wh, Wx, bias };
```
alalek: OpenCV 3.4 code should be C++98 compatible.

```
lp.blobs.push_back(Wh);
lp.blobs.push_back(Wx);
lp.blobs.push_back(bias);
```

smbz:
I've just replaced the initializer lists and "auto"; I think everything else is C++98 compatible.

PR #15592, opened 2019-09-25
fix slow python sample code in "Changing the contrast and brightness of an image!"

qkiet:


alalek:
Did you read code comment?

> but we wanted to show you how to access the pixels :)

qkiet:
>Ah my mistake, I tried to run this sample and found it's too slow, so I decided to make a pull request. I forgot it mean to show underlying principle
> > Did you read code comment?

PR #15596, opened 2019-09-25
Improve copymask and implement SIMD extensions

pmur:
### This pullrequest vectorizes more copyMask variants

Improve masked copy performance by introducing the v_permute HAL to efficiently expand byte masks to non-native integer sizes (e.g 24b for 8uC3 masking) to implement the 8uC3 variant.

Likewise, add template specialization for 32b types which may not always implicitly vectorize.

```
buildworker:Custom=linux-3
disable_ipp=ON
```

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin_cpp.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_cpp.hpp
@@ -50,6 +50,8 @@
 #include <algorithm>
 #include "opencv2/core/saturate.hpp"
 
+#define CV_SIMD128_PERMUTE 2
+
 namespace cv
 {
 
@@ -219,6 +221,7 @@ Regular integers:
 |reverse            | x | x | x | x | x | x |
 |extract_n          | x | x | x | x | x | x |
 |broadcast_element  |   |   |   |   | x | x |
+|permute            | x |   |   |   |   |   |
 
 Big integers:
 
@@ -2543,6 +2546,59 @@ v_pack_store(float16_t* ptr, v_reg<float, V_TypeTraits<float>::nlanes128>& v)
     {
         ptr[i] = float16_t(v.s[i]);
     }
+}
+
+/** @brief Vector permute
+
+Shuffle, permute, or tabular lookup between two vectors from a control vector.
+
+Similar to v_permute(ctrl,a), however A and B are treated as a composite
+vector of 2 * a.nlanes, where A contains the lower lanes, and B the higher.
+
+A specific HAL implementation could extend this to select between 256 8 bit
+elements.
+
+Scheme:
+@code
+X = { A; B}
+{ X[[CTRL[0]], X[CTRL[1]], ... X[CTRL[15]] }
+where CTRL[x] < v_uint8 * 2
+@endcode
+
+@note this API is dependent on the compiler check CV_SIMD128_PERMUTE > 1
+      or CV_SIMD_PERMUTE > 1.
+*/
+inline v_uint8x16 v_permute(const v_uint8x16& ctrl, const v_uint8x16& a, const v_uint8x16& b)
+{
+    uint8_t p[ctrl.nlanes];
+    v_uint8x16 tbl[] = { a, b };
+    for( int i = 0; i < ctrl.nlanes; i++ )
+        p[i] = tbl[(ctrl.s[i] / (uint8_t)ctrl.nlanes) & 1].s[ctrl.s[i] % ctrl.nlanes];
+    return v_uint8x16(p);
+}
+
+/** @brief Vector permute
+
+Shuffle, permute, or tabular lookup within a vector from a control vector.
+
+This intrinsic allows arbitrary byte movement or duplication using the
+control vector CTRL to select any element from the source vector A.
+
+The compiler guard CV_SIMD128_PERMUTE, or more generally, CV_SIMD_PERMUTE represent
+the number of vectors which can be efficiently permuted from.
+
+Scheme:
+@code
+{ A[CTRL[0]], A[CTRL[1]], ... A[CTRL[15]] }
+where CTRL[x] < v_uint8
+@endcode
+
+@note this API is dependent on the compiler check CV_SIMD128_PERMUTE > 0
+      or CV_SIMD_PERMUTE > 0.
+*/
+inline v_uint8x16 v_permute(const v_uint8x16& ctrl, const v_uint8x16& a)
+{
+    return v_permute(ctrl, a, a);
 }
 
 inline void v_cleanup() {}
--- a/modules/core/src/copy.cpp
+++ b/modules/core/src/copy.cpp
@@ -94,14 +94,30 @@ copyMask_<uchar>(const uchar* _src, size_t sstep, const uchar* mask, size_t mste
         {
             v_uint8 v_zero = vx_setzero_u8();
 
-            for( ; x <= size.width - v_uint8::nlanes; x += v_uint8::nlanes )
+            for( ; x <= size.width - v_uint8::nlanes * 2; x += v_uint8::nlanes * 2)
             {
                 v_uint8 v_src   = vx_load(src  + x),
                         v_dst   = vx_load(dst  + x),
                         v_nmask = vx_load(mask + x) == v_zero;
 
+                v_uint8 v_src2   = vx_load(src  + x + v_uint8::nlanes),
+                        v_dst2   = vx_load(dst  + x + v_uint8::nlanes),
+                        v_nmask2 = vx_load(mask + x + v_uint8::nlanes) == v_zero;
+
+                v_dst = v_select(v_nmask, v_dst, v_src);
+                v_dst2 = v_select(v_nmask2, v_dst2, v_src2);
+                v_store(dst + x, v_dst);
+                v_store(dst + x + v_uint8::nlanes, v_dst2);
+            }
+
+            if( x <= size.width - v_uint8::nlanes)
+            {
+                v_uint8 v_src   = vx_load(src  + x),
+                        v_dst   = vx_load(dst  + x),
+                        v_nmask = vx_load(mask + x) == v_zero;
                 v_dst = v_select(v_nmask, v_dst, v_src);
                 v_store(dst + x, v_dst);
+                x += v_uint8::nlanes;
             }
         }
         vx_cleanup();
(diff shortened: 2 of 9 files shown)


modules/core/src/copy.cpp
```diff
@@ -93,15 +93,23 @@ copyMask_<uchar>(const uchar* _src, size_t sstep, const uchar* mask, size_t mste
         #if CV_SIMD
         {
             v_uint8 v_zero = vx_setzero_u8();
+            int stride = v_uint8x16::nlanes * 2;
+            int len0 = size.width & -stride;
 
-            for( ; x <= size.width - v_uint8::nlanes; x += v_uint8::nlanes )
+            for( ; x < len0; x += stride )
```
alalek: Please use straightforward condition: `x <= size.width - v_uint8x16::nlanes * 2`.

`& -stride` trick assumes that initial and all `x` values are multiplier of `stride` and stride is power of 2 (or we should add assertion for that).

modules/core/include/opencv2/core/hal/intrin_cpp.hpp
```diff
@@ -2371,6 +2374,26 @@ v_pack_store(float16_t* ptr, v_reg<float, V_TypeTraits<float>::nlanes128>& v)
     }
 }
 
+/** @brief Vector permute
+
+Shuffle, permute, or tabular lookup within a vector from a control vector.
+
+This intrinsic allows arbitrary byte movement or duplication using the
+control vector CTRL to select any element from the source vector A.
+
```
alalek: Please add comment like this:
```
@note Requires CV_SIMD128_PERMUTE / CV_SIMD_PERMUTE compilation check.
```

modules/core/test/test_intrin_utils.hpp
```diff
@@ -1139,6 +1139,37 @@ template<typename R> struct TheTest
...
+#if CV_SIMD_PERMUTE > 0
+        Data<R> dataA;
+        Data<R> dataB;
+
+        // We want 0 - nlanes, n
+        for (int i = 0; i < v_uint8::nlanes; i++ )
+            dataB.d[v_uint8::nlanes - 1 - i] = (LaneType)i;
+
+        // Use permute to reverse data
+        R a = v_permute(v_uint8(dataB), v_uint8(dataA));
```
terfendail: I think it make sense to include test for 2 vector permute as well to ensure compatibility if/when multi vector permute support will be extended to other platforms

modules/core/include/opencv2/core/hal/intrin_cpp.hpp
```diff
@@ -2371,6 +2374,26 @@ v_pack_store(float16_t* ptr, v_reg<float, V_TypeTraits<float>::nlanes128>& v)
...
+Scheme:
+@code
+{ A[CTRL[0]], A[CTRL[1]], ... A[CTRL[15]] }
+@endcode
+*/
+inline v_uint8x16 v_permute(const v_uint8x16& ctrl, const v_uint8x16& a)
```
terfendail: IMO 2-vector permute should be added here as well

modules/core/test/test_intrin_utils.hpp
```diff
@@ -1139,6 +1139,86 @@ template<typename R> struct TheTest
...
+#define _expand_7(val, wrap) _expand_6(val,wrap) , wrap(val[6])
+#define _expand_8(val, wrap) _expand_7(val,wrap) , wrap(val[7])
+#define _expand_9(val, wrap) _expand_8(val,wrap) , wrap(val[8])
+#define _expand_10(val, wrap) _expand_9(val,wrap) , wrap(val[9])
+#define _expand_11(val, wrap) _expand_10(val,wrap) , wrap(val[10])
+#define _expand_12(val, wrap) _expand_11(val,wrap) , wrap(val[11])
+#define _expand_13(val, wrap) _expand_12(val,wrap) , wrap(val[12])
+#define _expand_14(val, wrap) _expand_13(val,wrap) , wrap(val[13])
+#define _expand_15(val, wrap) _expand_14(val,wrap) , wrap(val[14])
+#define _expand_16(val, wrap) _expand_14(val,wrap) , wrap(val[15])
```
terfendail: Shouldn't it be `_expand_15` ?

modules/core/src/copy.cpp
```diff
@@ -94,14 +94,30 @@ copyMask_<uchar>(const uchar* _src, size_t sstep, const uchar* mask, size_t mste
...
             v_uint8 v_zero = vx_setzero_u8();
 
-            for( ; x <= size.width - v_uint8::nlanes; x += v_uint8::nlanes )
+            for( ; x <= size.width - v_uint8::nlanes * 2; x += v_uint8::nlanes * 2)
             {
                 v_uint8 v_src   = vx_load(src  + x),
                         v_dst   = vx_load(dst  + x),
                         v_nmask = vx_load(mask + x) == v_zero;
 
+                v_uint8 v_src2   = vx_load(src  + x + v_uint8x16::nlanes),
```
terfendail: `v_uint8::nlanes` should be used here

modules/core/src/copy.cpp
```diff
@@ -94,14 +94,30 @@ copyMask_<uchar>(const uchar* _src, size_t sstep, const uchar* mask, size_t mste
...
+                        v_dst2   = vx_load(dst  + x + v_uint8x16::nlanes),
+                        v_nmask2 = vx_load(mask + x + v_uint8x16::nlanes) == v_zero;
+
                 v_dst = v_select(v_nmask, v_dst, v_src);
+                v_dst2 = v_select(v_nmask2, v_dst2, v_src2);
                 v_store(dst + x, v_dst);
+                v_store(dst + x + v_uint8x16::nlanes, v_dst2);
```
terfendail: `v_uint8::nlanes` also

modules/core/src/copy.cpp
```diff
@@ -94,14 +94,30 @@ copyMask_<uchar>(const uchar* _src, size_t sstep, const uchar* mask, size_t mste
         {
             v_uint8 v_zero = vx_setzero_u8();
 
-            for( ; x <= size.width - v_uint8::nlanes; x += v_uint8::nlanes )
+            for( ; CV_SIMD_IS_LT_ALIGNED_UNROLLED(x, size.width, v_uint8, 2); x += v_uint8::nlanes * 2)
```
terfendail: I think it could be a really long story to find out a best way to unify SIMD loops. So since it's completely unrelated to copyMask improvement I prefer to retain this discussion out of the PR scope.
@alalek WDYT?

ChipKerchner:
Don't forget about the MSA and WASM instruction sets.

pmur:
@ChipKerchner The `CV_SIMD_PERMUTE` compiler guard will implicitly protect those architectures.

@terfendail I will drop the patch with loop test macros. It would be nice to have some common macros to iterate through vectors.

modules/core/src/copy.cpp
```diff
@@ -149,6 +165,130 @@ copyMask_<ushort>(const uchar* _src, size_t sstep, const uchar* mask, size_t mst
...
+                v_uint8 src1 = vx_load(src + x + v_uint8::nlanes);
+                v_uint8 src2 = vx_load(src + x + v_uint8::nlanes * 2);
+                v_uint8 dst0 = vx_load(dst + x);
+                v_uint8 dst1 = vx_load(dst + x + v_uint8::nlanes);
+                v_uint8 dst2 = vx_load(dst + x + v_uint8::nlanes * 2);
+
+                // Use a permute to "expand" a value of 0 or -1.
+                v_uint8 m24v0 = v_permute(p0, m8v);
+                v_uint8 m24v1 = v_permute(p1, m8v);
+                v_uint8 m24v2 = v_permute(p2, m8v);
```
alalek: As I mentioned in #15613 lets introduce more specific intrinsic for that.
- manipulation with control parameter in implementations doesn't looks good (it is not universal and generic) - this way is not optimal for processing of that task
- take a look on `v_zip()` - it interleaves two registers. Here we need interleaving of three registers.
- lets try to avoid conditional compilation
- we already have `v_store_interleave()`. Here we don't need "storing". We need just interleaving from registers into registers. But...
- but even interleaving would not work with masks in the future (see OE-29 - AVX512): there are plans to rework masks support completely (masks type will be a separate "abstract" type with reduced amount of operations).
- so, finally, it is better to introduce here more specific intrinsic: `void v_select3(mask, a1, b1, c1, a2, b2, c2, dst_a, dst_b, dst_c)` (merge these v_permute + v_select operations). It can be implemented with optimal performance for each architecture.

asmorkalov:
@pmur Friendly reminder.

modules/core/src/copy.cpp
pmur: I think updating this patch would be highly contingent on the choose of HAL API exposed for masks. I can only assume this would heavily bias towards AVX512. AFAIK composite types (e.g 8u3) don't map onto the HAL.

Maybe `v_memmove_triplet(dstp, srcp, maskp, ntriplets)`?

ChipKerchner: I'm looking at the OpenCV code and I can see uses for a generic v_permute (or v_shuffle) instruction(s) in other areas.  v_permute would need both single and double input registers (besides the control vector).  Without these instructions, work in these areas might not be able to be done.

pmur: I am still waiting on community feedback before refactoring this patchset.

I think v_permute is well defined for most architectures. I can drop copymask patch while consensus is being developed for masked operations. Other than my suggestion above, I lack insight into what a good HAL would look like for such operations.

alalek: > v_memmove_triplet

Variant with memory pointers make sense too. For example, skip data loading from source if `v_check_any(mask) == false` or `v_check_all(mask) == true` (blending is not needed).

> ntriplets

Number of interleaved channels?

> composite types (e.g 8u3) don't map onto the HAL

Right, there is no dedicated type for that.
Interleaved data for 3 channels is stored in 3 vector registers. In case of 128-bit int32:
- `{ R1 G1 B1 R2}    {G2 B2 R3 G3}    {B3 R4 G4 B4}`
- see also `v_load_deinterleave()`

terfendail:
<details>
<summary>Performance for SSE2 baseline</summary>

| Performance test | Reference time |  PR time | Speedup |
|------------------|----------------|----------|---------|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC1)|0.000|0.000|1.04|
|Mat_CopyToWithMask::Size_MatType::(127x61, 16UC1)|0.001|0.001|1.01|
|Mat_CopyToWithMask::Size_MatType::(127x61, 32SC1)|0.003|0.002|2.02|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC2)|0.001|0.001|1.00|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC3)|0.005|0.005|1.00|
|Mat_CopyToWithMask::Size_MatType::(127x61, 32FC4)|0.004|0.004|1.00|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC1)|0.231|0.229|1.01|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 16UC1)|0.450|0.437|1.03|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 32SC1)|1.114|1.057|1.05|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC2)|0.438|0.439|1.00|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC3)|1.367|1.325|1.03|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 32FC4)|4.399|4.384|1.00|
</details>
<details>
<summary>Performance for SSE3 baseline</summary>

| Performance test | Reference time |  PR time | Speedup |
|------------------|----------------|----------|---------|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC1)|0.000|0.000|1.08|
|Mat_CopyToWithMask::Size_MatType::(127x61, 16UC1)|0.001|0.001|1.01|
|Mat_CopyToWithMask::Size_MatType::(127x61, 32SC1)|0.003|0.002|2.02|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC2)|0.001|0.001|0.99|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC3)|0.005|0.005|1.00|
|Mat_CopyToWithMask::Size_MatType::(127x61, 32FC4)|0.004|0.004|1.00|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC1)|0.233|0.241|0.96|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 16UC1)|0.421|0.449|0.94|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 32SC1)|1.137|1.068|1.06|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC2)|0.420|0.446|0.94|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC3)|1.361|1.355|1.00|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 32FC4)|4.286|4.340|0.99|
</details>
<details>
<summary>Performance for SSE4_2 baseline</summary>

| Performance test | Reference time |  PR time | Speedup |
|------------------|----------------|----------|---------|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC1)|0.000|0.000|1.00|
|Mat_CopyToWithMask::Size_MatType::(127x61, 16UC1)|0.001|0.001|1.02|
|Mat_CopyToWithMask::Size_MatType::(127x61, 32SC1)|0.003|0.001|2.40|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC2)|0.001|0.001|1.01|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC3)|0.005|0.001|5.25|
|Mat_CopyToWithMask::Size_MatType::(127x61, 32FC4)|0.004|0.004|1.01|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC1)|0.235|0.235|1.00|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 16UC1)|0.451|0.428|1.05|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 32SC1)|1.086|1.021|1.06|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC2)|0.455|0.426|1.07|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC3)|1.393|0.673|2.07|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 32FC4)|4.405|4.261|1.03|
</details>
<details>
<summary>Performance for AVX2 baseline</summary>

| Performance test | Reference time |  PR time | Speedup |
|------------------|----------------|----------|---------|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC1)|0.001|0.000|1.40|
|Mat_CopyToWithMask::Size_MatType::(127x61, 16UC1)|0.001|0.001|1.04|
|Mat_CopyToWithMask::Size_MatType::(127x61, 32SC1)|0.003|0.002|2.02|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC2)|0.001|0.001|1.02|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC3)|0.005|0.001|4.34|
|Mat_CopyToWithMask::Size_MatType::(127x61, 32FC4)|0.004|0.004|1.00|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC1)|0.226|0.233|0.97|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 16UC1)|0.435|0.450|0.97|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 32SC1)|1.083|1.045|1.04|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC2)|0.436|0.444|0.98|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC3)|1.318|0.723|1.82|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 32FC4)|4.386|4.396|1.00|
</details>
<details>
<summary>Performance for AVX512 baseline</summary>

| Performance test | Reference time |  PR time | Speedup |
|------------------|----------------|----------|---------|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC1)|0.000|0.000|1.14|
|Mat_CopyToWithMask::Size_MatType::(127x61, 16UC1)|0.000|0.000|0.96|
|Mat_CopyToWithMask::Size_MatType::(127x61, 32SC1)|0.003|0.002|2.02|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC2)|0.000|0.000|1.00|
|Mat_CopyToWithMask::Size_MatType::(127x61, 8UC3)|0.005|0.001|5.71|
|Mat_CopyToWithMask::Size_MatType::(127x61, 32FC4)|0.004|0.004|0.99|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC1)|0.226|0.227|0.99|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 16UC1)|0.415|0.399|1.04|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 32SC1)|1.099|1.048|1.05|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC2)|0.409|0.403|1.02|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 8UC3)|1.300|0.674|1.93|
|Mat_CopyToWithMask::Size_MatType::(1920x1080, 32FC4)|4.381|4.363|1.00|
</details>

pmur:
I am not sure the interest/need is here for this PR anymore. It can be revisited if desired.

PR #15623, opened 2019-10-02
Use circular lut history buffer in computeGradient of HOG

ChipKerchner:
Use circular lut history buffer in computeGradient of HOG.

Reduce lut reads by 2/3 and is 33-52% faster.

### DIFF
--- a/modules/objdetect/src/hog.cpp
+++ b/modules/objdetect/src/hog.cpp
@@ -317,6 +317,63 @@ void HOGDescriptor::computeGradient(const Mat& img, Mat& grad, Mat& qangle,
         xmap += 1;
     }
 
+#if CV_SIMD128
+    typedef const uchar* const T;
+    float *lutPrev, *lutCurr, *lutNext;
+    {
+        y = 0;
+        const uchar* imgPtr  = img.ptr(ymap[y]);
+        const uchar* prevPtr = img.data + img.step*ymap[y-1];
+
+        lutPrev = lutBuf+widthP2*0;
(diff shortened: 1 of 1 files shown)


ChipKerchner:
Any traction on this PR?  It's been over 3 weeks.  It would be nice to have more PRs merged by the end of the month.

modules/objdetect/src/hog.cpp
```diff
@@ -337,43 +342,171 @@ void HOGDescriptor::computeGradient(const Mat& img, Mat& grad, Mat& qangle,
...
+                    v_float32x4 _dy2 = _dy20 - _dy21;
+
+                    v_store(lutPrev+x+widthP2*2+1, _dy20);
+                    v_store(lutNext+x+widthP2*2+1, _dy21);
+
+                    v_float32x4 _mag0 = (_dx0 * _dx0) + (_dy0 * _dy0);
+                    v_float32x4 _mag1 = (_dx1 * _dx1) + (_dy1 * _dy1);
+                    v_float32x4 _mag2 = (_dx2 * _dx2) + (_dy2 * _dy2);
+
+                    v_float32x4 mask = v_reinterpret_as_f32(_mag2 > _mag1);
```
alalek: Lets avoid code duplication.

Here (`(y == 0)`) initialize prefetched data only and then call generic code below (`y != 0`).

Perhaps, initialization code should be moved outside of `for( y = 0; y < gradsize.height; y++ )` loop.

PR #15638, opened 2019-10-04
OpenCV G-API Filter2D test improvement

AsyaPronina:
- Size parameter is changed from int to cv::Size type to allow rectangle kernels
- Kernel creation code is adopted for different kernel sizes to not create only white images on the output

### DIFF
--- a/modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
+++ b/modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
@@ -57,9 +57,23 @@ TEST_P(Filter2DTest, AccuracyTest)
     cv::Point anchor = {-1, -1};
     double delta = 0;
 
-    cv::Mat kernel = cv::Mat(kernSize, kernSize, CV_32FC1);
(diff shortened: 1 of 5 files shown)


modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
```diff
@@ -57,16 +57,31 @@ TEST_P(Filter2DTest, AccuracyTest)
     cv::Point anchor = {-1, -1};
     double delta = 0;
 
-    cv::Mat kernel = cv::Mat(kernSize, kernSize, CV_32FC1);
-    cv::Scalar kernMean = cv::Scalar(1.0);
-    cv::Scalar kernStddev = cv::Scalar(2.0/3);
+    cv::Mat kernel = cv::Mat(filterSize, CV_32FC1);
+    cv::Scalar kernMean, kernStddev;
+
+    const auto kernSize = filterSize.width * filterSize.height;
```
TolyaTalamanov: use `cv::Size::area` instead

TolyaTalamanov:
LGTM :+1:

PR #15639, opened 2019-10-04
dnn(test): fix optional test data

alalek:
Avoid failures of [these tests](https://ocv-power.imavr.com/#/builders/1/builds/4826).


alalek:
:+1:

PR #15650, opened 2019-10-05
Pip issue 15647

### LINKED ISSUE #15647: Redesign of OpenCV's PnP algorithms
The implementation of Perspective n Point (PnP) in OpenCV needs a significant improvement.
Algorithm-based design is necessary to provide a good interface for two major components of PnP:

- PnP Solvers: Algorithms that solve PnP without an initial estimate (also called closed-form solutions)
- PnP Refiners: Algorithms that take a solution from a PnP solver, and improve on the solution by minimizing a non-convex objective function, using e.g. Levenberg Marquardt

There are many PnP solvers and PnP refiners proposed in the literature, some of which are implemented in OpenCV. With this generic interface, it will be much easier to include new solvers and refiners, to benchmark different combinations of solvers & refiners, and to unit test.

Furthermore, different PnP solvers are able to work in some conditions and not others. For example, DLT requires 6 or more non-coplanar points, IPPE requires 4 or more co-planar points, and for the same problem, different algorithms may return different solutions. Finally, for some algorithms, there may exist special cases which cause the algorithm to fail. For example, all PnP algorithms will fail if the model points are co-linear (this is known as a **general degeneracy** in PnP).  Some PnP algorithms will fail if the model points are co-planar (e.g. DLT or OpenCV's EPnP implementation). Furthermore, there may exist special corner cases that cause an algorithm to fail that are either not known or not well documented. For example, OpenCV implementation of Zhang's method (for solving PnP with co-planar points) fails if the model points are 4 corners of a square. This is called an **artificial degeneracy** (i.e. when the problem can be solved, but the algorithm fails as a result of its design.) None of this is clearly documented in OpenCV, and it is the cause of a lot of confusion and inexplicable results. There are also no test cases to document and verify failure cases of algorithms.

tobycollins:
### This pullrequest changes
This pull request is designed to fix the problems discussed in #15647

1. A `PnPSolver` base class has been created according to the OpenCV algorithm-based design guidelines. This is derived by classes that implement a PnP solver algorithm, and also for specifying important properties of the PnP solver, implemented as virtual member functions. These include:

- `minPointNumber`: specifies the minimal number of points that can be handled by the solver (e.g. EPnP requires 4 and DLT requires 6)

- `maxPointNumber`: specifies the maximal number of points that can be handled by the solver (e.g. AP3P handles at most 4 points, while others can handle an arbitrary number of points)

- `requires3DObject`: specifies if the PnP solver requires non co-planar model points

- `requiresPlanarObject`: specifies if the PnP solver requires co-planar model points

- `requiresPlanarTagObject`: specifies if the PnP solver requires co-planar model points arranged in a square tag formation

- `artificalDegeneracyWarn`: warns if the object points & image points will cause an artificial degeneracy

2.  The `PnPSolver` class provides the function `geometryWarn` that tests the input object points and image points for whether they will will cause the solver to fail. This involves running basic tests (checking number of points, checking for co-linear points etc.) and also running specific checks for each algorithn, as implemented in `artificalDegeneracyWarn`.

3. All the currently used PnP solvers in OpenCV have been implemented as derived classes of `PnPSolver`: EPnP, DLT, DLS,  P3P, AP3P, Zhang, IPPE and IPPESquare. 

4. A `PnPRefiner` base class has bee created according to the OpenCV algorithm-based design guidelines. This is derived by classes that implements refinement algorithms. All the currently used refinement algorithms in OpenCV have been implemented as derived classes: C API levenberg Marquardt, C++ API Levenberg Marquardt and VVS.

5. The public interfaces: `SolvePnP` and `SolvePnPRansac` are unchanged. When these are called, `PnPSolver` and `PnPRefiner` objects are created according to the flags passed to `SolvePnP` and `SolvePnPRansac` respectively.

6. The public interface: `solvePnPGeneric` has been modified to accept algorithm pointers as arguments, providing a cleaner and generic use with any PnPSolver and PnPRefiner objects.

7. A new test set has been created (test_solve_generic.cpp) that does the following:

- Tests all combinations of `PnPSolver` and `PnPRefiner` objects for solver precision, using all combinations of precision and formats for object points, image points, intrinsics and distortion coefficients. This exhaustive test makes testing a new PnPSolver or PnPRefiner classes easy. 

- Tests all refiners using an imprecise initial pose estimate. 

- Documents corner cases that cause one or more `PnPSolvers` to fail (also called artificial degeneracies). This set can grow if more corner cases get included by the community.


alalek:
Please don't reopen PRs about the same/similar things.

tobycollins:
Sorry about the double PR.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -3270,14 +3270,1282 @@ optimization. It stays at the center or at a different location specified when C
...
+
+
+/** @brief P3P algorithm from @cite gao2003complete
+*/
+class CV_EXPORTS_W PnPSolverP3PComplete : public PnPSolver{
+public:
+
+/** @brief create creates a pointer to a new instance of this class
+
+    @param withGeometricTests set this true if you want to execute geometryWarn before solving pnp.
```
paroj: the refers to the non existing `geometryWarn` (probably `validProblem` ) function. Use @ref to get an buildbot error for this.
Also I would recommend using `@copydoc PnPSolver::PnPSolver` for the create functions to avoid copy & paste

paroj:
I think documenting the corner cases of the different algorithms alone is an invaluable contribution.

I just have some minor code-style suggestions.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -3270,14 +3270,1282 @@ optimization. It stays at the center or at a different location specified when C
...
+    @param ipoints Set of Image points. This must be a 1xN 2-channel or Nx1 2-channel matrix (double) where n is the number of points.
+    @return false if there is a problem, true otherwise
+    */
+    CV_WRAP bool validProblem(InputArray opoints,InputArray ipoints) const;
+
+/** @brief isPlanarTag Checks if the object points are in the correct format of a planar tag
+    @param _opoints
+    @return true if correct, false otherwise
+    */
+    CV_WRAP bool isPlanarTag(InputArray _opoints) const;
```
paroj: this does not use any members and thus can be static

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -3270,14 +3270,1282 @@ optimization. It stays at the center or at a different location specified when C
...
+
+    @returns Maximal number of points supported by algorithm. If no maximum, returns -1
+     */
+    CV_WRAP virtual int maxPointNumber() const = 0;
+
+/** @brief Indicates if the algorithm requires the object points to be 3D (i.e. not co-planar)
+
+    @returns true if required, false otherwise
+     */
+    CV_WRAP virtual bool requires3DObject() const = 0;
```
paroj: how about using a single `int getCaveats() const` function returning a mask like `PNP_CAVEAT_PLANAR | PNP_CAVEAT_TAG` instead of the three functions below?

The same mask could be also returned from `validProblem` to signal which check failed.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -3270,14 +3270,1282 @@ optimization. It stays at the center or at a different location specified when C
...
+    CV_WRAP bool isPlanarTag(InputArray _opoints) const;
+
+
+protected:
+
+/** @brief Makes a 3x3 identity intrinsic matrix
+
+    @returns Identity intrinsic matrix (3x3 single channel double)
+    */
+    CV_WRAP cv::Mat makeIdentityIntrinsic() const;
```
alalek: > protected:
> CV_WRAP

This should be removed from public API, so this can't be wrapped.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -3270,14 +3270,1282 @@ optimization. It stays at the center or at a different location specified when C
...
+
+/** @brief getNumberOfPoints Gets number of points in object matrix. Must be 1xN 3-channel or Nx1 3-channel matrix (double) where n is the number of points.
+
+    @param opoints Set of Object points. This must be a 1xN 3-channel or Nx1 3-channel matrix (double) where n is the number of points.
+    opoints are defined in object coordinates.
+    @return
+    */
+    CV_WRAP size_t getNumberOfPoints(InputArray opoints) const;
```
alalek: See `Mat::checkVector()`

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -3270,14 +3270,1282 @@ optimization. It stays at the center or at a different location specified when C
...
+    opoints are defined in object coordinates.
+    @param ipoints Set of Image points. This must be a 1xN 2-channel or Nx1 2-channel matrix (double) where n is the number of points.
+    ipoints are defined in normalized pixel coordinates. That is, they are defined by using a canonical intrinsic calibration matrix of K = eye(3).
+    You can convert a set of points defined in pixels to normalized pixel coordinates using @ref undistortPoints. This is convenient because it
+    eliminates the need for supplying the camera's intrinsics (calibration matrix and distortion parameters)
+    @param rvecs Vector of output rotation vectors (double) (see @ref Rodrigues ) that, together with tvecs, brings points from
+    the model coordinate system to the camera coordinate system.
+    @param tvecs Vector of output translation vectors (double)
+     */
+    CV_WRAP virtual void solve(InputArray opoints, InputArray ipoints, CV_OUT std::vector<Mat> & rvecs, CV_OUT std::vector<Mat> & tvecs) const = 0;
```
alalek: Why is this method protected? (I believe `solveProblem()` should be virtual instead)

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -3270,14 +3270,1282 @@ optimization. It stays at the center or at a different location specified when C
...
+    you can set this to false for a small speed improvement.
+    @return pointer
+    */
+    CV_WRAP static Ptr<PnPSolverP3PComplete> create(bool withGeometricTests = false);
+
+/** @brief PnPSolverP3P constructor
+
+    @param withGeometricTests flag
+    */
+    CV_WRAP PnPSolverP3PComplete(bool withGeometricTests);
```
alalek: `bool withGeometricTests = false` to be consistent with `create()` interface

asmorkalov:
@tobycollins Friendly reminder. Do you have a chance to fix notes?

asenyaev:

PR #15658, opened 2019-10-07
Cuda + OpenGL on ARM

tolysz:
### This pull request changes

There might be multiple ways of getting OpenCV compile on Tegra (NVIDIA Jetson) platform, but mainly they modify CUDA(8,9,10...) source code, this one fixes it for all installations. 
( https://devtalk.nvidia.com/default/topic/1007290/jetson-tx2/building-opencv-with-opengl-support-/post/5141945/#5141945 et al.).
This way is exactly the same as the one proposed but the code change happens in OpenCV.

```
opencv_contrib=
```

### DIFF
--- a/modules/core/src/opengl.cpp
+++ b/modules/core/src/opengl.cpp
@@ -45,6 +45,13 @@
 #ifdef HAVE_OPENGL
 #  include "gl_core_3_1.hpp"
 #  ifdef HAVE_CUDA
+#    if defined(__arm__) || defined(__aarch64__)
+#      warning Workaround for CUDA on ARM
+#      include <GL/gl.h>
+#      ifndef GL_VERSION
+#        define GL_VERSION
+#      endif
+#    endif
 #    include <cuda_gl_interop.h>
 #  endif
 #else // HAVE_OPENGL
--- a/modules/core/src/opengl.cpp
+++ b/modules/core/src/opengl.cpp
@@ -45,11 +45,11 @@
 #ifdef HAVE_OPENGL
 #  include "gl_core_3_1.hpp"
 #  ifdef HAVE_CUDA
-#    if defined(__arm__) || defined(__aarch64__)
-#      warning Workaround for CUDA on ARM
+#    if (defined(__arm__) || defined(__aarch64__)) \
+         && !defined(OPENCV_SKIP_CUDA_OPENGL_ARM_WORKAROUND)
 #      include <GL/gl.h>
 #      ifndef GL_VERSION
-#        define GL_VERSION
+#        define GL_VERSION 0x1F02
 #      endif
 #    endif
 #    include <cuda_gl_interop.h>


alalek:
> patch-1

Please use descriptive names next time.
This name conflicts with patch into `opencv_contrib` under the same "patch-1" but for different target branch (this is a CI feature for testing patches into multiple repositories at once).

modules/core/src/opengl.cpp
```diff
@@ -45,6 +45,13 @@
 #ifdef HAVE_OPENGL
 #  include "gl_core_3_1.hpp"
 #  ifdef HAVE_CUDA
+#    if defined(__arm__) || defined(__aarch64__)
+#      warning Workaround for CUDA on ARM
```
alalek: To make robust solution please add option to bypass this workaround:
```
if (defined(__arm__) || defined(__aarch64__)) \
    && !defined(OPENCV_SKIP_CUDA_OPENGL_ARM_WORKAROUND)
```

which can be applied via C++ CMake flags.

tolysz: Should this workaround be enabled or disabled by default? How should it called `FORCE_GL_CUDA_ARM` or `DISABLE_GL_CUDA_ARM` or???
The current change is derived from the patch for CUDAs to make OpenCV compile with it, and covers exactly the cases which CUDA currently rejects:
```diff --git a/cuda_gl_interop.h b/cuda_gl_interop.h
index 0f4aa17..e8c538c 100644
--- a/cuda_gl_interop.h
+++ b/cuda_gl_interop.h
@@ -59,13 +59,13 @@
 
 #else /* __APPLE__ */
 
-#if defined(__arm__) || defined(__aarch64__)
-#ifndef GL_VERSION
-#error Please include the appropriate gl headers before including cuda_gl_interop.h
-#endif
-#else
+//#if defined(__arm__) || defined(__aarch64__)
+//#ifndef GL_VERSION
+//#error Please include the appropriate gl headers before including cuda_gl_interop.h
+//#endif
+//#else
 #include <GL/gl.h>
-#endif
+//#endif
 
 #endif /* __APPLE__ */
```
re: `patch-1` it is an artifact from github editor, will try to rename them before making PRs

tolysz:
done, but I am not sure if I extended cmake in the right way... I tested it both ways.

modules/core/src/opengl.cpp
alalek: Thank you for update!

I don't see information about fixed CUDA versions in linked forum thread, so this workaround should be enabled by default (but probably we should eliminate `#warning` from here or disable that message under another "*_MESSAGE" definition flag).

Please change "default" value of `CL_VERSION` to `0x1F02`. I see this on my machine (Fedora):
```
/usr/include/GL/gl.h:#define GL_VERSION				0x1F02
/usr/include/GL/glcorearb.h:#define GL_VERSION                        0x1F02
```

I believe CMake changes are not necessary (option is very specific, lets keep it hidden).
Flag can be added via CMake for future fixed versions (if necessary - workaround is harmless without warning message and proper default value):
```
cmake ... -DOPENCV_EXTRA_FLAGS=-DOPENCV_SKIP_CUDA_OPENGL_ARM_WORKAROUND=1
```

asmorkalov:
The issue is still there with latest Jetpack and CUDA 10 I have on my Jetson Nano. @alalek Let's go ahead and merge the workaround. :+1:

PR #15510, opened 2019-09-12
 core: rework and optimize SIMD implementation of dotProd

seiko2plus:
relates #15506
merge with opencv_extra opencv/opencv_extra#674

### This pullrequest changes
   - add new universal intrinsics v_dotprod[int32], v_dotprod_expand[u&int8, u&int16, int32], v_cvt_f64(int64)
  - add a boolean param for all v_dotprod&_expand intrinsics that change the behavior of addition order between
    pairs in some platforms in order to reach the maximum optimization when the sum among all lanes is what only matters
  - fix clang build on ppc64le
  - support wide universal intrinsics for dotProd_32s
  - remove raw SIMD and activate universal intrinsics for dotProd_8
  - implement SIMD optimization for dotProd_s16&u16
  - extend performance test data types of dot prod
  - fix GCC VSX workaround of vec_mule and vec_mulo (in little-endian must be swapped)
  - optimize v_mul_expand(int32) on VSX

### TODO
- [x] performance tests
- [x] ppc64le clang coverage 
- [x]  cover msa & wasm
- [x]  update perf data

````
force_builders=Custom,armv7,armv8,linux
buildworker:Custom=linux-1
docker_image:Custom=mips64el
pw_compilers=gcc-4.9, power9_gcc-5, power9_gcc-6, power9_gcc-7, clang-4, clang-5, clang-6
pw_with_opencv_extra=branch:issue15506
# https://ocv-power.imavr.com/#/builders/3/builds/84
# imgproc test crash on clang 4, 5 . I disable it for now
pw_disable_tests=rgbd, shape, imgproc
````
#### Args used with opencv_perf_core
```
--gtest_filter=*dot* --perf_threads=1 --perf_min_samples=1000 --gtest_output=xml:[path]
```

### X86
<details>
<summary>CPU</summary>

```
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                8
On-line CPU(s) list:   0-7
Thread(s) per core:    2
Core(s) per socket:    4
Socket(s):             1
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 142
Model name:            Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
Stepping:              10
CPU MHz:               818.725
CPU max MHz:           4000.0000
CPU min MHz:           400.0000
BogoMIPS:              3984.00
Virtualization:        VT-x
L1d cache:             32K
[...]

PR #15661, opened 2019-10-08
build: update AVX2 check

alalek:
- _mm256_bslli_epi128() works in [GCC 4.9.3+](https://gcc.godbolt.org/z/LIZG-L) only
- Android NDK r10 doesn't support this instruction

relates #15510
fixes [nightly build](http://pullrequest.opencv.org/buildbot/builders/3_4_pack-android/builds/530).

```
force_builders=Android pack,Linux AVX2
```


terfendail:
Sorry for late response. Wouldn't it be better to replace calls to `_mm256_bslli_epi128` with calls to `_mm256_slli_si256`. According to the Intel Intrinsics Guide both have the same effect and are mapped to the same asm instruction, but the latter already exists in the code and probably doesn't cause the build failure.

PR #15676, opened 2019-10-09
Update ISSUE_TEMPLATE.md

Yakshchopra:


alalek:
Please avoid such kind of patches (like playing with wording, whitespaces, etc) which doesn't fix anything.
"Reporting" of feature requests doesn't look good anyway.

Issue #15677, opened 2019-10-09
cv::magnitudeSqr and cv::magnitude(InputArray xy,...) ?


chacha21:
cv::cuda::magnitude... functions have more options that cv::magnitude..., because :
-there is a cv::cuda::magnitudeSqr()
-cv::cuda::magnitude(...) can accept x and y as interleaved rather than two planes.

-cv::magnitudeSqr() would require the use of ippsPowerSpectr() rather than ippsMagnitude()
-cv::magnitude(xy) would require the use of "ipp...32/64fc" rather than "ipp...32/64f" functions
-the biggest part of the work would be in ocl/hal to support new functions like magnitudeSqr() magnitude32c().

Considering that work, would it be accepted as a pull request, or would it be rejected as not useful enough to justify more code ?

PR #15682, opened 2019-10-10
added cv::magnitude(InputArray xy...)

chacha21:
Issue #15677

cv::magnitude now support interlaved input data as 32FC2 or 64FC2
TODO:
ippicvlib needs an implementation of ippsMagnitude_32fc/ippsMagnitude_64fc
the OpenCL abstraction is beyond my understanding to support interleaved data in ocl_math_op

### DIFF
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -1593,6 +1593,7 @@ have the same size as x.
 @sa cartToPolar, polarToCart, phase, sqrt
 */
 CV_EXPORTS_W void magnitude(InputArray x, InputArray y, OutputArray magnitude);
+CV_EXPORTS_W void magnitude(InputArray xy, OutputArray magnitude);
 
 /** @brief Checks every element of an input array for invalid values.
 
--- a/modules/core/src/mathfuncs_core.dispatch.cpp
+++ b/modules/core/src/mathfuncs_core.dispatch.cpp
@@ -51,6 +51,19 @@ void magnitude32f(const float* x, const float* y, float* mag, int len)
         CV_CPU_DISPATCH_MODES_ALL);
 }
 
+void magnitude32fc(const float* xy, float* mag, int len)
+{
+    CV_INSTRUMENT_REGION();
+
+    CALL_HAL(magnitude32fc, cv_hal_magnitude32fc, xy, mag, len);
+    //TODO : use a new version of ippicv with ippsMagnitude_32fc implemented !
+    // SSE42 performance issues
+    //CV_IPP_RUN(IPP_VERSION_X100 > 201800 || cv::ipp::getIppTopFeatures() != ippCPUID_SSE42, CV_INSTRUMENT_FUN_IPP(ippsMagnitude_32fc, reinterpret_cast<const Ipp32fc*>(xy), mag, len) >= 0);
+
+    CV_CPU_DISPATCH(magnitude32fc, (xy, mag, len),
+        CV_CPU_DISPATCH_MODES_ALL);
+}
+
 void magnitude64f(const double* x, const double* y, double* mag, int len)
 {
     CV_INSTRUMENT_REGION();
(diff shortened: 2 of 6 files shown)


chacha21:
How can I ask for help for the OCL part ?

modules/core/include/opencv2/core.hpp
```diff
@@ -1595,6 +1595,7 @@ have the same size as x.
 @sa cartToPolar, polarToCart, phase, sqrt
 */
 CV_EXPORTS_W void magnitude(InputArray x, InputArray y, OutputArray magnitude);
+CV_EXPORTS_W void magnitude(InputArray xy, OutputArray magnitude);
```
alalek: There is conflict with Python bindings. Second variant with both variables doesn't not work (would select the first variant - output are optional parameter).

`magnitudeComplex()`?
'c' suffix below has meaning of 'complex' numbers.
But `xy` name has no sense. It is better to use `src` instead.

modules/core/src/mathfuncs_core.simd.hpp
```diff
@@ -224,6 +226,40 @@ void magnitude32f(const float* x, const float* y, float* mag, int len)
...
 
+void magnitude32fc(const float* xy, float* mag, int len)
+{
+    CV_INSTRUMENT_REGION();
+
+    int i = 0;
+
+#if CV_SIMD
+    const int VECSZ = v_float32::nlanes;
+    for( ; i < len; i += VECSZ*2 )
```
alalek: Why is loop unrolled? Are there benefits for that?

modules/core/include/opencv2/core.hpp
chacha21: I just followed the same naming convention as in cv::cuda;;magnitude()

modules/core/src/mathfuncs_core.simd.hpp
chacha21: I just took the cv::magnitude implementation.
I assumed that it was designed as a hint to allow specific compiler optimization on 256b registers.

asmorkalov:
@chacha21 Do you have any progress on the patch?

chacha21:
Still unresolved :
-alalek did not like the naming convention, that I just took from cv::cuda::magnitude() (see comments above). What should I do ?
-alalek asked about the loop unrolling : I am not expert enough to decide anything (see comments above).
-I do not know how to insert an OCL implementation in the current OCL abstraction for arithmetic operations. The target code would theoretically be very simple, if we could just access ".x" and ".y" from float2 variables, but I do not know how to push that in the current OCL workflow.

modules/core/src/mathfuncs_core.dispatch.cpp
```diff
@@ -51,6 +51,20 @@ void magnitude32f(const float* x, const float* y, float* mag, int len)
...
 }
 
+void magnitude32fc(const float* xy, float* mag, int len)
+{
+    CV_INSTRUMENT_REGION();
+
+    CALL_HAL(magnitude32fc, cv_hal_magnitude32fc, xy, mag, len);
+    #if defined(HAVE_IPP) && !defined(HAVE_IPP_ICV)//ippicv should implement ippsMagnitude_32fc
+    // SSE42 performance issues
+    CV_IPP_RUN(IPP_VERSION_X100 > 201800 || cv::ipp::getIppTopFeatures() != ippCPUID_SSE42, CV_INSTRUMENT_FUN_IPP(ippsMagnitude_32fc, reinterpret_cast<const Ipp32fc*>(xy), mag, len) >= 0);
```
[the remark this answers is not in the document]
chacha21: About the IPP performance, I can only test on my machine. There was a marginal improvement for a large number of loops on big images.
Is there a procedure to perform such tests ?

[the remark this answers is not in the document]
alalek: Collect performance through regular OpenCV performance tests + `summary.py` to compare results:
- You need to write corresponding test and select appropriate filter for "--gtest_filter" parameter (avoid running thousands of other tests)
- run and save results "with IPP": `--gtest_output=xml:with_ipp.xml`
- run and save results "without IPP" (can be disabled in runtime: `OPENCV_IPP=disabled`): : `--gtest_output=xml:without_ipp.xml`
- get report: `<path to ts/mics>summary.py without_ipp.xml with_ipp.xml`

chacha21:
Here are the test results, with and without IPP, not sure about the interpretation

[with_ipp.xml.txt](https://github.com/opencv/opencv/files/3977653/with_ipp.xml.txt)
[without_ipp.xml.txt](https://github.com/opencv/opencv/files/3977654/without_ipp.xml.txt)

vpisarev:
@chacha21, @alalek, in order to solve the problem with Python bindings, can we change the existing cv::magnitude(x, y, mag)?

cv::magnitude(xy, noArray(), mag) will automatically treat the first multi-channel array as an array of short vectors (x,y) (or more generally, in future, as an array of vectors (x,y,z,...)).

In this case in Python bindings, when you call `mag = cv2.magnitude(xy_numpy_arr, None)`, it will automatically call `cv::magnitude()` with empy `cv::Mat` 2nd parameter.

chacha21:
@vpisarev  @alalek 
You tell me, you are the best to decide which naming convention should be use

vpisarev:
@chacha21, yes, so I actually suggest not to export cv::magnitude(xy, mag) to Python/Java/etc (i.e. let's use `CV_EXPORTS` for it instead of `CV_EXPORTS_W`). Instead, we can modify the existing cv::magnitude(x, y, mag) to call cv::magnitude(xy, mag) if `y` is empty array.

asenyaev:

asenyaev:

PR #15684, opened 2019-10-10
G-API: G_ID_HELPER_BODY macro semicolon issue 

OrestChura:
### This pullrequest changes
- Unnecessary extra semicolon after member function definition has been removed from the G_ID_HELPER_BODY macro definition because that has been raising warnings.

```
 struct G_ID_HELPER_CLASS(Class)                                            \
        {                                                                   \
-           static constexpr const char * id() {return Id;};                \
+           static constexpr const char * id() {return Id;}                 \
        };     
```


OrestChura:
@dmatveev

PR #15699, opened 2019-10-14
G-API: Implement OpenCV render backend

TolyaTalamanov:


TolyaTalamanov:
@dmatveev

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -63,11 +65,13 @@ namespace
 
         static auto ocv_pkg =
 #if !defined(GAPI_STANDALONE)
-            combine(cv::gapi::core::cpu::kernels(),
-                    cv::gapi::imgproc::cpu::kernels());
+            combine(combine(cv::gapi::core::cpu::kernels(),
```
TolyaTalamanov: I think it's better than adding render kernel to opencv kernel package

modules/gapi/test/common/gapi_render_tests.hpp
```diff
@@ -78,7 +78,7 @@ class RenderWithParam : public TestWithParam<T>
     cv::Mat y_mat_ocv, uv_mat_ocv, y_mat_gapi, uv_mat_gapi, mat_ocv, mat_gapi;
 };
 
-using TestArgs = std::tuple<cv::Size,cv::gapi::wip::draw::Prims,cv::gapi::GKernelPackage>;
+using TestArgs = std::tuple<cv::Size,cv::gapi::wip::draw::Prims>;
```
TolyaTalamanov: I think there is no reason to add `GCompileArgs` to test parameters because we have only `use_freetype` but not testing it on CI. Maybe later.

modules/gapi/test/gapi_sample_pipelines.cpp
```diff
@@ -314,4 +316,20 @@ TEST(GAPI_Pipeline, CanUseOwnMatAsOutput)
     EXPECT_NO_THROW(comp.apply({in_own_mat}, {out_own_mat}));
 }
 
+TEST(GAPI_Pipeline, GDummyTest)
```
TolyaTalamanov: Remove it change

modules/gapi/include/opencv2/gapi/render/render.hpp
```diff
@@ -23,13 +23,16 @@ namespace cv
...
 {
 
-namespace ocv { GAPI_EXPORTS cv::gapi::GKernelPackage kernels(); }
-
 namespace wip
 {
 namespace draw
 {
 
+struct use_freetype
```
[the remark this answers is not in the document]
TolyaTalamanov: Added

modules/gapi/test/gapi_sample_pipelines.cpp
TolyaTalamanov: Removed

TolyaTalamanov:
@dmatveev

modules/gapi/src/backends/render/grenderocv.hpp
```diff
@@ -0,0 +1,46 @@
...
+    static KImpl kernel()               { return KImpl{Name::run};                 }
+};
+
+} // namespace ocv
+} // namespace render
+} // namespace gapi
+} // namespace cv
+
+#define GAPI_RENDER_OCV_KERNEL(Name, API) struct Name: public cv::gapi::render::ocv::GRenderKernelImpl<Name, API>, \
+                                                       public cv::detail::KernelTag
```
dmatveev: BTW, why not reuse the existing OCV_KERNEL's machinery to preserve kernel's entry points _typed_? The reason we unroll "call contexts" into specific call arguments is type safety - it is less error-prone this way.

modules/gapi/src/backends/render/grenderocvbackend.cpp
```diff
@@ -0,0 +1,141 @@
...
+                case NodeType::OP:
+                    if (this_nh == nullptr) {
+                        this_nh = nh;
+                    }
+                    else
+                        util::throw_error(std::logic_error("Multi-node rendering is not supported!"));
+                    break;
+                default: util::throw_error(std::logic_error("Unsupported NodeType type"));
+            }
+        }
```
dmatveev: I'd rewrite it with two `find_if`s. The first one would be used to obtain `this_nh` and the second one to check if that one was the only one.
Yes, IE backend written like here but it also handles DATA things

modules/gapi/src/compiler/gcompiler.cpp
dmatveev: Can you please write a FIXME on adding N-arg version of combine? Opening an internal issue would be nice as well

modules/gapi/test/common/gapi_render_tests.hpp
dmatveev: But how do we instantiate these tests with proper backend then?

modules/gapi/src/backends/render/grenderocv.hpp
```diff
@@ -0,0 +1,46 @@
...
+{
+namespace gapi
+{
+namespace render
+{
+namespace ocv
+{
+
+GAPI_EXPORTS cv::gapi::GBackend backend();
+GAPI_EXPORTS cv::gapi::GKernelPackage kernels();
```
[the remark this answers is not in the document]
TolyaTalamanov: User should't have access to it. It's internal stuff

modules/gapi/src/backends/render/grenderocv.hpp
TolyaTalamanov: Because `GCPUKernel` works with `<void(GCPUContext&)>` functors, but i want to expand it to `<void(GRenderContext&, BitmaskCreator&)>` for example

modules/gapi/src/backends/render/grenderocv.hpp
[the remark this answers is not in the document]
TolyaTalamanov: You are right !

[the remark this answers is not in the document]
TolyaTalamanov: What about export `cv::gapi::render::ocv::backend()` ? Why does the user need this function ?

modules/gapi/test/common/gapi_render_tests.hpp
TolyaTalamanov: I will add this in PR about integrate freetype

modules/gapi/src/compiler/gcompiler.cpp
TolyaTalamanov: Added FIXME

modules/gapi/include/opencv2/gapi/render/render.hpp
```diff
@@ -23,13 +23,21 @@ namespace cv
...
 namespace draw
 {
 
+/**
+ * A structure allows using freetype library for text rendering
+ */
+struct use_freetype
+{
+    /*@{*/
+    const char* path; //!< The path to font file (.ttf)
```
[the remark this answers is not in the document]
TolyaTalamanov: Replaced

modules/gapi/src/backends/render/grenderocvbackend.cpp
TolyaTalamanov: Fixed

modules/gapi/src/backends/render/grenderocv.hpp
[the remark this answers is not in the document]
TolyaTalamanov: Exported only `::kernels()`

modules/gapi/src/backends/render/grenderocv.hpp
TolyaTalamanov: Yes, fixed

TolyaTalamanov:
@dmatveev

modules/gapi/test/common/gapi_render_tests.hpp
TolyaTalamanov: - But how do we instantiate these tests with proper backend then?"
- It's not necessary we have a single render backend (I'm going to refactor tests in the near future)

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -98,6 +106,7 @@ class GAPI_EXPORTS GCPUContext
     std::unordered_map<std::size_t, GRunArgP> m_results;
 
     friend class gimpl::GCPUExecutable;
+    friend class gimpl::render::ocv::GRenderExecutable;
```
dmatveev: Is it still required here? Just because GCPUContext is reused in the implementation?

TolyaTalamanov: Yes

modules/gapi/include/opencv2/gapi/render/render.hpp
```diff
@@ -23,13 +23,21 @@ namespace cv
 namespace gapi
 {
 
-namespace ocv { GAPI_EXPORTS cv::gapi::GKernelPackage kernels(); }
```
alalek: Broken ABI.
Intentional?

TolyaTalamanov: Yes

PR #15701, opened 2019-10-14
core(alloc): force initialization of memalign flag

### LINKED ISSUE #15691: Segfault in multithreaded program in OpenCV 4.1.2

- OpenCV => 4.1.2
- Operating System / Platform => Gentoo Linux
- Compiler => gcc (Gentoo 8.3.0-r1 p1.1) 8.3.0

##### Detailed description

With OpenCV-4.1.2 I've started to get random segfaults in multi-threaded program. I've managed to reproduce the problem with the very basic application, 2 threads where each of them just creates a Mat.

##### Steps to reproduce
```c++
#include <opencv2/core.hpp>
#include <thread>

using namespace std;
using namespace cv;

void thread1() {
    Mat(10, 10, CV_16U, Scalar(0));
}

void thread2() {
    Mat(10, 10, CV_16U, Scalar(0));
}

int main() {
    thread th1(thread1);
    thread th2(thread2);
    th1.join();
    th2.join();
    return 0;
}
```

When compiled with:
```bash
g++ -g $(pkg-config --libs --cflags opencv4) -lpthread test-crash-412.cpp -o test
```

Will randomly crash or exit normally when run. To get a more stable repro, please run several times:
```bash
while true; do ./test; done
```

Running the same binary (without recompilation) against OpenCV-4.1.1 works fine without any crashes.

This is the stack trace I see:
```
Stack trace of thread 19341:
#0  0x00007f07d1dd9b50 cfree (libc.so.6)
#1  0x00007f07d256a5b9 n/a (/home/pro/projects/opencv-lib/opencv-4.1/install/lib64/libopencv_core.so.4.1.2)
#2  0x000055e3320da48c n/a (/tmp/test)
#3  0x000055e3320da2b8 n/a (/tmp/test)
#4  0x000055e3320da9ff n/a (/tmp/test)
#5  0x000055e3320da7f9 n/a (/tmp/test)
#6  0x000055e3320daf02 n/a (/tmp/test)
#7  0x000055e3320daec3 n/a (/tmp/test)
#8  0x000055e3320dae98 n/a (/tmp/test)
#9  0x00007f07d2159cae execute_native_thread_routine (libstdc++.so.6)
#10 0x00007f07d2288458 start_thread (libpthread.so.0)
#11 0x00007f07d1e546ef __clone (libc.so.6)
```

alalek:
resolves #15691

- before main() launch

relates #15544

### DIFF
--- a/modules/core/src/alloc.cpp
+++ b/modules/core/src/alloc.cpp
@@ -112,6 +112,13 @@ bool isAlignedAllocationEnabled()
     }
     return useMemalign;
 }
+// do not use variable directly, details: https://github.com/opencv/opencv/issues/15691
+static const bool g_force_initialization_memalign_flag
+#if defined __GNUC__
+    __attribute__((unused))
+#endif
+    = isAlignedAllocationEnabled();
+
 #endif
 
 #ifdef OPENCV_ALLOC_ENABLE_STATISTICS


mika-fischer:
@alalek We also ran into this issue.

I think this workaround is a bit clunky. The main problem is that `isAlignedAllocationEnabled` is not thread-safe.

Since c++11 is now required, you could simplify it to the following, which would make it thread-safe:
```cpp
static inline
bool isAlignedAllocationEnabled()
{
    static bool useMemalign = readMemoryAlignmentParameter();
    return useMemalign;
}
```
Then you wouldn't need the `g_force_initialization_memalign_flag` variable anymore.

PR #15704, opened 2019-10-15
G-API: avoid dynamic initialization in dispatched files

### LINKED ISSUE #15690: "Illegal instruction" for debug libraries on non-AVX CPU
##### System information
- OpenCV => 4.1.1, 4.1.2
- Operating System / Platform => Windows 8.1 64 Bit
- Compiler => Visual Studio 2019

##### Detailed description
When I tried to debug an OpenCV application using OpenCV 4.1.1, I got a runtime error in "glfluidimgproc_func.simd.hpp", line 879 - "Unhandled exception at 0x00007FFAA7D35862 (opencv_world412d.dll) in testApp.exe: 0xC000001D: Illegal instruction."

It seems that this is related to the CPU (Intel Core i5 M520) that doesn't support AVX. After some research, I found two StackOverflow posts with the exact same problem:

- https://stackoverflow.com/questions/58007793/using-cmake-to-make-a-library-without-sse-support-windows-version/58044006#58044006
- https://stackoverflow.com/questions/57959602/trouble-with-setting-up-opencv-4-1-1-on-visual-microsoft-studio-2019-illega

The system information given in the first one completely matches my own (VS 2019, Win 8.1, i5 M520), so the issue might as well be related to those. The second one is for Win 7, but an unspecified i5 CPU, too.

Note that release builds do work, but that isn't really helpful. Also, the first StackOverflow post offers a workaround by compiling from souce with an adjusted OpenCVCompilerOptimizations.cmake. Didn't try that myself yet, I'd prefer a bugfix or workaround that allows using the pre-built libraries.

##### Other things tried that didn't work

- Building 4.1.1 libraries myself (using the git-bash and cmake build [instructions here](https://docs.opencv.org/master/d3/d52/tutorial_windows_install.html))
- Using environment variable "OPENCV_CPU_DISABLE" set to "AVX2,AVX,FP16"
- Using the just released 4.1.2 pre-built libraries (from the Release section in this repository)

##### Steps to reproduce

- Downloading 4.1.1 or 4.1.2 pre-built libraries from the [releases section](https://github.com/opencv/opencv/releases) (opencv-4.1.x-vc14_vc15.exe)
- Using CMake to build the OpenCV samples (in opencv/sources/samples)
- Setting opencv_version as startup project
- Running as "Debug, x64"

##### Screenshot

image

alalek:
resolves #15690

Initialization code is executed without any dispatching guards for supported CPU instruction sets.

### DIFF
--- a/modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
+++ b/modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
@@ -876,11 +876,12 @@ void run_yuv2rgb_impl(uchar out[], const uchar in[], int width, const float coef
 // V' = (R' - Y')*0.877
 static const float coef[5] = {0.299f, 0.587f, 0.114f, 0.492f, 0.877f};
 
-static const ushort c0 = static_cast<ushort>(coef[0]*(1 << 16) + 0.5f);
-static const ushort c1 = static_cast<ushort>(coef[1]*(1 << 16) + 0.5f);
-static const ushort c2 = static_cast<ushort>(coef[2]*(1 << 16) + 0.5f);
-static const short c3 = static_cast<short>(coef[3]*(1 << 12) + 0.5f);
-static const short c4 = static_cast<short>(coef[4]*(1 << 12) + 0.5f);
+// don't use expressions (avoid any dynamic initialization): https://github.com/opencv/opencv/issues/15690
+static const ushort c0 = 19595;  // static_cast<ushort>(coef[0]*(1 << 16) + 0.5f);
+static const ushort c1 = 38470;  // static_cast<ushort>(coef[1]*(1 << 16) + 0.5f);
+static const ushort c2 = 7471;   // static_cast<ushort>(coef[2]*(1 << 16) + 0.5f);
+static const short c3 = 2015;    // static_cast<short>(coef[3]*(1 << 12) + 0.5f);
+static const short c4 = 3592;    // static_cast<short>(coef[4]*(1 << 12) + 0.5f);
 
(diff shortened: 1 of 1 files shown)


modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
```diff
@@ -876,11 +876,12 @@ void run_yuv2rgb_impl(uchar out[], const uchar in[], int width, const float coef
...
-static const ushort c1 = static_cast<ushort>(coef[1]*(1 << 16) + 0.5f);
-static const ushort c2 = static_cast<ushort>(coef[2]*(1 << 16) + 0.5f);
-static const short c3 = static_cast<short>(coef[3]*(1 << 12) + 0.5f);
-static const short c4 = static_cast<short>(coef[4]*(1 << 12) + 0.5f);
+// don't use expressions (avoid any dynamic initialization): https://github.com/opencv/opencv/issues/15690
+static const ushort c0 = 19595;  // static_cast<ushort>(coef[0]*(1 << 16) + 0.5f);
+static const ushort c1 = 38470;  // static_cast<ushort>(coef[1]*(1 << 16) + 0.5f);
+static const ushort c2 = 7471;   // static_cast<ushort>(coef[2]*(1 << 16) + 0.5f);
+static const short c3 = 2015;    // static_cast<short>(coef[3]*(1 << 12) + 0.5f);
+static const short c4 = 3592;    // static_cast<short>(coef[4]*(1 << 12) + 0.5f);
```
dmatveev: May this be solved with `constexpr` instead?

PR #15714, opened 2019-10-16
Support software triggering aravis cameras

neurodroid:
If an aravis camera is set up to be software triggered, a trigger needs to be explicitly sent using `arv_camera_software_trigger`, otherwise the camera will not grab any frames.

### This pullrequest changes
`modules/videoio/src/cap_aravis.cpp` to support aravis cameras that are set up to be software triggered. Without this change, only hardware-triggered cameras (i.e. cameras receiving a clock/TTL signal to one of their input lines) can grab frames.

```
allow_multiple_commits=1
```

### DIFF
--- a/modules/videoio/include/opencv2/videoio.hpp
+++ b/modules/videoio/include/opencv2/videoio.hpp
@@ -483,6 +483,16 @@ enum { CAP_PROP_XI_DOWNSAMPLING                                 = 400, //!< Chan
 
 //! @} XIMEA
 
+/** @name XIMEA Camera API
+*  @{
+*/
+
+//! Properties of cameras available through ARAVIS backend
+enum { CAP_PROP_ARAVIS_AUTOTRIGGER                              = 600 //!< Automatically trigger frame capture if camera is configured with software trigger
+};
+
+//! @} ARAVIS
+
 /** @name AVFoundation framework for iOS
     OS X Lion will have the same API
     @{
--- a/modules/videoio/src/cap_aravis.cpp
+++ b/modules/videoio/src/cap_aravis.cpp
@@ -148,6 +148,7 @@ protected:
     double          exposureCompensation;
     bool            autoGain;
     double          targetGrey;             // Target grey value (mid grey))
+    bool            softwareTriggered;      // Flag if the camera is software triggered
 
     gint64          *pixelFormats;
     guint           pixelFormatsCnt;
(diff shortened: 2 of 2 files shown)


ArkadiuszRaj:
Hi @neurodroid , which camera model you have in mind?
The code in reposiotry works with several cameras from different manufactures, to name few: TIS, PTGrey/FLIR, Basler.

neurodroid:
> Hi @neurodroid , which camera model you have in mind?
> The code in reposiotry works with several cameras from different manufactures, to name few: TIS, PTGrey/FLIR, Basler.

https://www.flir.com/products/chameleon3-usb3/?model=CM3-U3-13Y3M-CS

ArkadiuszRaj:
That is very strange as I have CM3-U3-31S4M and observe no problem with using this implementation of Aravis adapter. But I will take a look at the datasheet to say more

ArkadiuszRaj:
OK, I understand that instead of Line0 you need to set manual triggering.

I prefer then the configuration to select software or internal triggering, i.e. adapter code shall set TriggerSource register explicilty and follow that by optional software triggering you have already provided in patch.

asmorkalov:
@neurodroid Do you plan to work on option for `cv::VideoCapture` as @ArkadiuszRaj recommended in scope of this MR?

ArkadiuszRaj:
To be clear: I see no point in implementing software triggering that just happens after previous frame is taken from adapter library. I.e. this software triggering is a facility that shall be run in parallel to grabbing frames. For examaple to by in sync with some GPS-alike device and trigger frame acquisition exactly every 2s.

Having said that I am voting for enforcing the camera to use LINE_0 as trigger, i.e. to configure it properly when someone set camera to strange configuration outside of OpenCV application.

asmorkalov:
@neurodroid I appologise for very late response. The team was very busy with release activities before New Year holidays. Could you introduce Aravis camera specific flag like `CAP_PROP_ARAVIS_AUTOTRIGGER` and make the proposed behaviour optional. The flag can be set by `cv::VideoCapture::set` method. The default behaviour should be as it is now and auto triggering is disabled. Having the option you can enable required functionality and not break existing logic.

asmorkalov:
@alalek please take a look and merge.

asmorkalov:
@ArkadiuszRaj @neurodroid I added option to `cv::VideoCapture` that manages software trigger behavior. Original behavior without software trigger is set by default. Could you take a look on the patch and try with real camera?

asmorkalov:
@ArkadiuszRaj @neurodroid any concerns about my solution?

ArkadiuszRaj:
@asmorkalov  no comments form my side, looks fine.

asmorkalov:
:+1:

PR #15715, opened 2019-10-16
COMP: Enabling build with recent VTK version

adamrankin:
VTK_MAJOR_VERSION not found unless header is included

### This pullrequest changes

Adds include of vtkVersionMacros.h in precomp.hpp


alalek:
We have "Linux x64" builder based on Ubuntu 14.04 (EOL) which contains VTK 6.0 (see checks).

adamrankin:
Excellent, looks like it passed! Let me know if you need any information.

PR #15719, opened 2019-10-16
refactor: remove extraneous uses-sdk from Android Lib manifest

PiwwowPants:
# This pullrequest changes

The uses-sdk in the Android manifest for the android lib is deprecated and replaced by the declaration in the grade file for this same library. Unless it has a practical purpose, it should be removed. Otherwise it should at least be updated to match the gradle file. The linter throws error syncing gradle when uses-sdk is in an Android manifest.


alalek:
You looking for a wrong file:

Gradle variant is here: https://github.com/opencv/opencv/blob/master/modules/java/android_sdk/android_gradle_lib/AndroidManifest.xml

PiwwowPants:
>You looking for a wrong file: ...

@alalek i see. I saw the build.gradle.ln at the android_sdk level and assumed they’d both use it and that the gradle variant had other purposes.

PR #14492, opened 2019-05-06
Async API for GAPI

anton-potapov:
GAPI: fixup for https://github.com/opencv/opencv/pull/14346 

- minor cosmetic changes : comments, typos

@dbudniko please take a look

PR #14346, opened 2019-04-16
Async API for GAPI

anton-potapov:
- naive implementation as a starting point

### This pullrequest changes
- Adds a naive implementation of Asynchronous API for GAPI module

PR #15735, opened 2019-10-18
G-API: Doxygen documentatation for Async API

anton-potapov:
### This pullrequest adds doxygen documentation for GAPI Async API
(here  #14346 #14492 is the object of documentation )
@dmatveev, @dbudniko  please take a look

### DIFF
--- a/modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
+++ b/modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
@@ -21,18 +21,50 @@ namespace cv {
 namespace gapi{
 namespace wip {
     class GAsyncContext;
-    //These functions asynchronously (i.e. probably on a separate thread of execution) call operator() member function of their first argument with copies of rest of arguments (except callback) passed in.
-    //The difference between the function is the way to get the completion notification (via callback or a waiting on std::future object)
-    //If exception is occurred during execution of apply it is transferred to the callback (via function parameter) or passed to future (and will be thrown on call to std::future::get)
+    /**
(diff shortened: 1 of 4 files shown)


modules/gapi/include/opencv2/gapi/gcompiled_async.hpp
```diff
@@ -21,18 +21,50 @@ namespace cv {
...
+    N.B. :
+    Input arguments are copied on call to async function (actually on call to cv::gin) and thus do not have to outlive the actual completion of asynchronous activity.
+    While output arguments are "captured" by reference(pointer) and therefore _must_ outlive the asynchronous activity
+    (i.e. live at least until callback is called or future is unblocked)
+
+    @param gcmpld       Compiled computation (graph) to start asynchronously
+    @param callback     Callback to be called when execution of gcmpld is done
+    @param ins          Input parameters for gcmpld
+    @param outs         Output parameters for gcmpld
+    */
```
anton-potapov: @alalek , I see that doxygen erroneously links samples/dnn/object_detection.cpp. sample here. Probably due to fact that the sample has argument named async.  Is there any way to fix this ?

alalek: My suggestion is to rename variable (but not a parameter itself) `async` -> `asyncParam` in DNN sample.

anton-potapov: renamed parameter 'async' ->  'async_fwd' . @dkurt could you please take a look ?

PR #15488, opened 2019-09-09
Vectorize minMaxIdx functions

ChipKerchner:
Vectorize minMaxIdx functions.

minMaxIdx_8u & minMaxIdx_8s - 11.1x improvement on VSX and 8.6x speedup on x86.
minMaxIdx_16u & minMaxIdx_16s - 8.3x improvement on VSX and 7.5x speedup on x86.
minMaxIdx_32s - 5.1x improvement on VSX and 4.2x speedup on x86.
minMaxIdx_32f - 4.1x improvement on VSX and 3.2x speedup on x86.
minMaxIdx_64f - 1.6x improvement on VSX and 1.5x speedup on x86.

<cut/>

```
force_builders=Custom,ARMv7
build_image:Docs=docs-js
#buildworker:Custom=linux-1
#build_image:Custom=mips64el
#build_image:Custom=javascript-simd

buildworker:Custom=linux-3
build_image:Custom=ubuntu:18.04
CPU_BASELINE:Custom=AVX512_SKX
disable_ipp=ON
```

PR #15738, opened 2019-10-18
Fixing bug with comparison of v_int64x2 or v_uint64x2

ChipKerchner:
Casting v_int64x2 or v_uint64x2 to v_float64x2 and comparing does NOT work in all cases.  Rewrite using epi64 instructions - faster too.

Here is an example that does NOT work

v_int64x2 a = v_setall_s64(-1);
v_int64x2 b = v_setall_s64(-1);

if (v_reinterpret_as_f64(a) == v_reinterpret_as_f64(b))

// Always fails because reinterpreting it as a f64 produces a NaN.  NaN is never equal to any other number, not even itself.

<cut/>

```
buildworker:Custom=linux-3
disable_ipp=ON
```

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin_sse.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_sse.hpp
@@ -1042,7 +1042,7 @@ inline _Tpvec operator != (const _Tpvec& a, const _Tpvec& b) \
 #define OPENCV_HAL_IMPL_SSE_64BIT_CMP_OP(_Tpvec) \
 inline _Tpvec operator == (const _Tpvec& a, const _Tpvec& b) \
 { __m128i cmp = _mm_cmpeq_epi32(a.val, b.val); \
-  return _Tpvec(_mm_or_si128(cmp, _mm_shuffle_epi32(cmp, _MM_SHUFFLE(2, 3, 0, 1)))); } \
+  return _Tpvec(_mm_and_si128(cmp, _mm_shuffle_epi32(cmp, _MM_SHUFFLE(2, 3, 0, 1)))); } \
 inline _Tpvec operator != (const _Tpvec& a, const _Tpvec& b) \
 { return ~(a == b); }
 #endif
--- a/modules/core/include/opencv2/core/hal/intrin_sse.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_sse.hpp
@@ -1050,15 +1050,6 @@ inline _Tpvec operator != (const _Tpvec& a, const _Tpvec& b) \
 OPENCV_HAL_IMPL_SSE_64BIT_CMP_OP(v_uint64x2)
 OPENCV_HAL_IMPL_SSE_64BIT_CMP_OP(v_int64x2)
 
-#define OPENCV_HAL_IMPL_SSE_64BIT_CMP_OP(_Tpvec, cast) \
-inline _Tpvec operator == (const _Tpvec& a, const _Tpvec& b) \
-{ return cast(v_reinterpret_as_f64(a) == v_reinterpret_as_f64(b)); } \
-inline _Tpvec operator != (const _Tpvec& a, const _Tpvec& b) \
-{ return cast(v_reinterpret_as_f64(a) != v_reinterpret_as_f64(b)); }
-
-OPENCV_HAL_IMPL_SSE_64BIT_CMP_OP(v_uint64x2, v_reinterpret_as_u64)
-OPENCV_HAL_IMPL_SSE_64BIT_CMP_OP(v_int64x2, v_reinterpret_as_s64)
-
 inline v_float32x4 v_not_nan(const v_float32x4& a)
 { return v_float32x4(_mm_cmpord_ps(a.val, a.val)); }
 inline v_float64x2 v_not_nan(const v_float64x2& a)
--- a/modules/core/include/opencv2/core/hal/intrin_sse.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_sse.hpp
@@ -1032,6 +1032,24 @@ inline _Tpvec operator >= (const _Tpvec& a, const _Tpvec& b) \
 OPENCV_HAL_IMPL_SSE_FLT_CMP_OP(v_float32x4, ps)
 OPENCV_HAL_IMPL_SSE_FLT_CMP_OP(v_float64x2, pd)
 
+#if CV_SSE4_1
+#define OPENCV_HAL_IMPL_SSE_64BIT_CMP_OP(_Tpvec) \
+inline _Tpvec operator == (const _Tpvec& a, const _Tpvec& b) \
+{ return _Tpvec(_mm_cmpeq_epi64(a.val, b.val)); } \
+inline _Tpvec operator != (const _Tpvec& a, const _Tpvec& b) \
+{ return ~(a == b); }
+#else
+#define OPENCV_HAL_IMPL_SSE_64BIT_CMP_OP(_Tpvec) \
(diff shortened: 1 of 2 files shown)


ChipKerchner:
The WASM version needs updating as well.

modules/core/include/opencv2/core/hal/intrin_sse.hpp
```diff
@@ -1032,14 +1032,23 @@ inline _Tpvec operator >= (const _Tpvec& a, const _Tpvec& b) \
...
-{ return cast(v_reinterpret_as_f64(a) == v_reinterpret_as_f64(b)); } \
+{ return _Tpvec(_mm_cmpeq_epi64(a.val, b.val)); } \
 inline _Tpvec operator != (const _Tpvec& a, const _Tpvec& b) \
-{ return cast(v_reinterpret_as_f64(a) != v_reinterpret_as_f64(b)); }
+{ return ~(a == b); }
+#else
+#define OPENCV_HAL_IMPL_SSE_64BIT_CMP_OP(_Tpvec) \
+inline _Tpvec operator == (const _Tpvec& a, const _Tpvec& b) \
+{ __m128i cmp = _mm_cmpeq_epi32(a.val, b.val); \
+  return _Tpvec(_mm_or_si128(cmp, _mm_shuffle_epi32(cmp, _MM_SHUFFLE(2, 3, 0, 1)))); } \
```
alalek: > or

low 32-bits are equal "or" high 32-bits are equal?

Probably `operator !=` should be defined first through "or".

ChipKerchner: Actually if either low or upper are set, the whole 64-bit register needs to be set.  If not, functions like v_check_any/all will not work

ChipKerchner: Not quite sure what you are suggesting in your above comment.

ChipKerchner:
> The WASM version needs updating as well. I'm not sure I can do this.

It seems like the changes are something like this:

--- a/modules/core/include/opencv2/core/hal/intrin_wasm.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_wasm.hpp
@@ -2742,6 +2742,8 @@ OPENCV_HAL_IMPL_WASM_INIT_CMP_OP(v_uint16x8, u16x8, i16x8)
 OPENCV_HAL_IMPL_WASM_INIT_CMP_OP(v_int16x8, i16x8, i16x8)
 OPENCV_HAL_IMPL_WASM_INIT_CMP_OP(v_uint32x4, u32x4, i32x4)
 OPENCV_HAL_IMPL_WASM_INIT_CMP_OP(v_int32x4, i32x4, i32x4)
+OPENCV_HAL_IMPL_WASM_INIT_CMP_OP(v_uint64x2, u64x2, i64x2)
+OPENCV_HAL_IMPL_WASM_INIT_CMP_OP(v_int64x2, i64x2, i64x2)
 OPENCV_HAL_IMPL_WASM_INIT_CMP_OP(v_float32x4, f32x4, f32x4)

 #ifdef __wasm_unimplemented_simd128__
@@ -2762,15 +2764,6 @@ OPENCV_HAL_IMPL_INIT_FALLBACK_CMP_OP(v_float64x2, <=)
 OPENCV_HAL_IMPL_INIT_FALLBACK_CMP_OP(v_float64x2, >=)
 #endif

-#define OPENCV_HAL_IMPL_WASM_64BIT_CMP_OP(_Tpvec, cast) \
-inline _Tpvec operator == (const _Tpvec& a, const _Tpvec& b) \
-{ return cast(v_reinterpret_as_f64(a) == v_reinterpret_as_f64(b)); } \
-inline _Tpvec operator != (const _Tpvec& a, const _Tpvec& b) \
-{ return cast(v_reinterpret_as_f64(a) != v_reinterpret_as_f64(b)); }
-
-OPENCV_HAL_IMPL_WASM_64BIT_CMP_OP(v_uint64x2, v_reinterpret_as_u64)
-OPENCV_HAL_IMPL_WASM_64BIT_CMP_OP(v_int64x2, v_reinterpret_as_s64)
-
 inline v_float32x4 v_not_nan(const v_float32x4& a)
 {
     v128_t z = wasm_i32x4_splat(0x7fffffff);

modules/core/include/opencv2/core/hal/intrin_sse.hpp
alalek: Fix `_mm_or_si128` => `_mm_and_si128()` is enough.

Example:
```
v_int64x2 a = v_setall_s64(0x111223344);
v_int64x2 b = v_setall_s64(0x011223344);
```

alalek:
> WASM

Looks like there is no `wasm_i64x2_eq()` yet:
https://github.com/emscripten-core/emscripten/blob/incoming/system/include/wasm_simd128.h

Lets update WASM backend in a separate PR (test would help to catch error there too).

modules/core/include/opencv2/core/hal/intrin_sse.hpp
ChipKerchner: You're right.  It should be AND and not OR.

I'll add some test cases.

alalek:
> merge conflict

Your branch is based on very old commit from 2019-08-08
Please avoid moving code - put fresh code near the end of the file.

```
# add upstream (once)
git remote add upstream	https://github.com/opencv/opencv.git
# this is very useful for conflicts resolving: git config --global merge.conflictstyle diff3

# rebase current branch on latest code
git fetch -t upstream master:master 3.4:3.4
git rebase -i upstream/3.4
# in case of conflict: edit files, "git add <modified files>", "git rebase --continue"

# update with --force
git push origin <your_branch_name_like_bugInt64x2Comparison> -f

# In the future: starting of new bugfix branch:
git fetch -t upstream master:master 3.4:3.4
git checkout -B my_new_fix upstream/3.4
```

alalek:
Please ignore errors from Custom/AVX512 builder - they are not related to this PR.

modules/core/test/test_intrin_utils.hpp
```diff
@@ -1517,6 +1561,9 @@ void test_hal_intrin_uint64()
     TheTest<v_uint64>()
         .test_loadstore()
         .test_addsub()
+#if CV_SIMD_64F
+        .test_cmp64()
```
alalek: Perhaps we should avoid this check for `v_uint64` types.
@terfendail Could you take a look?

ChipKerchner: The problem is for NEON comparison of v_int64x2 and v_uint64x2 does NOT exist unless CV_SIMD_64.

alalek: You are right.
In theory, `CV_SIMD_64F` should guard using of vectors of `doubles` and their operations.
Additionally, it guards some 64u/64s operations and some operations with 32f (like div) on NEON.

Perhaps we should address that in a separate PR.

terfendail: On the first look It seems like there is no reason to guard 64u/64s for NEON. This intrinsics doesn't use specific 64f related instructions. Maybe it make sense to change NEON guard scope?

alalek: Underlying instructions are available on AARCH64 only. Removing guard breaks ARMv7 builds.

Probably we need dedicated macro for this functionality (btw, `==` can be emulated through 32-bit (like SSE2/3), but less/greater comparisons are not easy to re-implement).

ChipKerchner:
Thanks for the approval/merge.  This bug broke another PR's build. [15488](https://github.com/opencv/opencv/pull/15488)

How do I get 15488 to rebuild?

PR #15764, opened 2019-10-23
Convert demosaic functions to HAL (2.6x faster for VSX)

ChipKerchner:
Convert demosaic functions to HAL (2.6x faster for VSX).

Added bayerRGBA to vectorized functions (missing from x86) - plus added alpha to parameter list.

Leaving NEON code in since it has much better instructions for interleaving data - plus added HAL bayerRGB_EA for NEON.

```
force_builders=linux,docs,Custom,Linux AVX2,ARMv7
buildworker:Custom=linux-1
build_image:Custom=mips64el
```

### DIFF
--- a/modules/imgproc/src/demosaicing.cpp
+++ b/modules/imgproc/src/demosaicing.cpp
@@ -86,6 +86,7 @@
 
 
 #include "precomp.hpp"
+#include "opencv2/core/hal/intrin.hpp"
 
 #include <limits>
 
@@ -111,7 +112,7 @@ public:
         return 0;
     }
 
-    int bayer2RGBA(const T*, int, T*, int, int) const
+    int bayer2RGBA(const T*, int, T*, int, int, const T) const
     {
         return 0;
     }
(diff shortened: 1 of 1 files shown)


alalek:
Thank you for working on this!
Could you please take a look on compilation issues?

ChipKerchner:
> Could you please take a look on compilation issues?

I'm not sure why this compilation is taking and timing out after almost 2 hrs.  I've compiled it on gcc for x86, Power, and ARM without problem.

Any suggestions?

alalek:
@ChipKerchner Sorry, It is CI failure.

ChipKerchner:
> By the way, did you compare that NEON code performance vs universal intrinsics performance on the same platform?
> Although NEON intrinsics code is more compact and intuitive, the less code to support we have the better.

I have no way on comparing NEON code performance - only correctness.  If someone else can see if the NEON or the HAL code is faster on an ARM, please do so.  If HAL is faster, I'll eliminate the NEON code.

savuor:
👍

ChipKerchner:
Great!  I'll probably have a 2nd PR for the other sections of demosiacing at a later time.

Issue #15693, opened 2019-10-13
[Feature Request opencv-dnn] Swish Activation, Mish Activation


LukeAI:
Google's famous [Efficientnet](https://ai.googleblog.com/2019/05/efficientnet-improving-accuracy-and.html) uses the [Swish Activation function](https://arxiv.org/abs/1710.05941)
It is currently not possible to use efficientnet in opencv due to the absence of swish.

The newer [Mish](https://arxiv.org/abs/1908.08681) activation function reportedly outperforms swish and would be great to have as an option.

dkurt:
It's possible for now to enable this network with custom layers mechanism. Please read a tutorial https://docs.opencv.org/master/dc/db1/tutorial_dnn_custom_layers.html.

suab321321:
@dkurt @LukeAI I would like to help with this one

dkurt:
@suab321321, thanks! Feel free to open a PR

grassknoted:
Hi @dkurt seeing no progress, I was wondering if I could take this issue up? I had a look at the link:

> Please read a tutorial https://docs.opencv.org/master/dc/db1/tutorial_dnn_custom_layers.html.

I'm still unsure of where to add the Mish activation function. Could you just help me get started, please?

alalek:
Usage questions should go to Users OpenCV Q/A forum: http://answers.opencv.org
Custom layers are part of User code. Check "samples".

grassknoted:
Thanks for the reply @alalek ! 
Just to clarify, I would need to create an Mish function example in the `samples/` directory, right? Something like: `samples/python/mish.py`?

alalek:
Well, if you want to add this implementation OpenCV then check this patch: #10889 + add tests on data from original framework (1-2 layers, to ensure that it is parsed properly).

Not sure if this is really "good first issue".

dkurt:
As far as I know Swish is an activation layer so you need to implement it here: https://github.com/opencv/opencv/blob/master/modules/dnn/src/layers/elementwise_layers.cpp.

Also, you need to register a new name at https://github.com/opencv/opencv/blob/master/modules/dnn/src/init.cpp

and write some tests depends on origin framework.

grassknoted:
> Well, if you want to add this implementation OpenCV then check this patch: #10889 + add tests on data from original framework (1-2 layers, to ensure that it is parsed properly).
> 
> Not sure if this is really "good first issue".

[...]

PR #15808, opened 2019-10-30
Added Swish and Mish activations

thebhatman:
I have added the Swish and Mish activation functions. This resolves https://github.com/opencv/opencv/issues/15693

###This Pull request changes
[Feature addition] New activation functions Swish and Mish.

```
force_builders=Custom
buildworker:Custom=linux-4

```

### DIFF
--- a/modules/dnn/src/cuda/activations.cu
+++ b/modules/dnn/src/cuda/activations.cu
@@ -63,6 +63,44 @@ namespace cv { namespace dnn { namespace cuda4dnn  { namespace kernels {
         }
 
         template <class T, std::size_t N>
+        __global__ void swish_vec(Span<T> output, View<T> input) {
+            using vector_type = get_vector_type_t<T, N>;
+
+            auto output_vPtr = vector_type::get_pointer(output.data());
+            auto input_vPtr = vector_type::get_pointer(input.data());
+
+            for (auto i : grid_stride_range(output.size() / vector_type::size())) {
+                vector_type vec;
+                v_load(vec, input_vPtr[i]);
+                for (int j = 0; j < vector_type::size(); j++) {
+                    using device::sigmoid;
+                    vec.data[j] = vec.data[j] * sigmoid(vec.data[j]);
+                }
+                v_store(output_vPtr[i], vec);
+            }
+        }
+
+        template <class T, std::size_t N>
+        __global__ void mish_vec(Span<T> output, View<T> input) {
+            using vector_type = get_vector_type_t<T, N>;
+
+            auto output_vPtr = vector_type::get_pointer(output.data());
+            auto input_vPtr = vector_type::get_pointer(input.data());
+
+            for (auto i : grid_stride_range(output.size() / vector_type::size())) {
+                vector_type vec;
+                v_load(vec, input_vPtr[i]);
+                for (int j = 0; j < vector_type::size(); j++) {
+                    using device::tanh;
+                    using device::log;
+                    using device::exp;
+                    vec.data[j] = vec.data[j] * tanh(log(1.0f + exp(vec.data[j])));
+                }
+                v_store(output_vPtr[i], vec);
+            }
+        }
+
+        template <class T, std::size_t N>
         __global__ void sigmoid_vec(Span<T> output, View<T> input) {
             using vector_type = get_vector_type_t<T, N>;
 
--- a/modules/dnn/src/cuda/activations.cu
+++ b/modules/dnn/src/cuda/activations.cu
@@ -94,7 +94,7 @@ namespace cv { namespace dnn { namespace cuda4dnn  { namespace kernels {
                     using device::tanh;
                     using device::log;
                     using device::exp;
-                    vec.data[j] = vec.data[j] * tanh(log(1.0f + exp(vec.data[j])));
+                    vec.data[j] = vec.data[j] * tanh(log(T(1.0) + exp(vec.data[j])));
                 }
                 v_store(output_vPtr[i], vec);
             }
--- a/modules/dnn/src/cuda/activations.cu
+++ b/modules/dnn/src/cuda/activations.cu
@@ -279,6 +279,58 @@ namespace cv { namespace dnn { namespace cuda4dnn  { namespace kernels {
     template void tanh<float>(const Stream&, Span<float>, View<float>);
 
     template <class T, std::size_t N>
+    void launch_vectorized_swish(const Stream& stream, Span<T> output, View<T> input) {
+        CV_Assert(is_fully_aligned<T>(output, N));
+        CV_Assert(is_fully_aligned<T>(input, N));
+
+        auto kernel = raw::swish_vec<T, N>;
+        auto policy = make_policy(kernel, output.size() / N, 0, stream);
+        launch_kernel(kernel, policy, output, input);
+    }
+
+    template <class T>
+    void swish(const Stream& stream, Span<T> output, View<T> input) {
+        CV_Assert(input.size() == output.size());
+
+        if (is_fully_aligned<T>(output, 4) && is_fully_aligned<T>(input, 4)) {
+            launch_vectorized_swish<T, 4>(stream, output, input);
+        } else if (is_fully_aligned<T>(output, 2) && is_fully_aligned<T>(input, 2)) {
+            launch_vectorized_swish<T, 2>(stream, output, input);
+        } else {
+            launch_vectorized_swish<T, 1>(stream, output, input);
+        }
+    }
+
+    template void swish<__half>(const Stream&, Span<__half>, View<__half>);
+    template void swish<float>(const Stream&, Span<float>, View<float>);
+
+    template <class T, std::size_t N>
+    void launch_vectorized_mish(const Stream& stream, Span<T> output, View<T> input) {
+        CV_Assert(is_fully_aligned<T>(output, N));
+        CV_Assert(is_fully_aligned<T>(input, N));
+
+        auto kernel = raw::mish_vec<T, N>;
+        auto policy = make_policy(kernel, output.size() / N, 0, stream);
+        launch_kernel(kernel, policy, output, input);
+    }
+
+    template <class T>
+    void mish(const Stream& stream, Span<T> output, View<T> input) {
+        CV_Assert(input.size() == output.size());
+
+        if (is_fully_aligned<T>(output, 4) && is_fully_aligned<T>(input, 4)) {
+            launch_vectorized_mish<T, 4>(stream, output, input);
+        } else if (is_fully_aligned<T>(output, 2) && is_fully_aligned<T>(input, 2)) {
+            launch_vectorized_mish<T, 2>(stream, output, input);
+        } else {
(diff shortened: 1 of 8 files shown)


modules/dnn/src/layers/elementwise_layers.cpp
```diff
@@ -613,6 +613,186 @@ struct TanHFunctor
...
+        inps.getUMatVector(inputs);
+        outs.getUMatVector(outputs);
+        String buildopt = oclGetTMacro(inputs[0]);
+
+        for (size_t i = 0; i < inputs.size(); i++)
+        {
+            UMat& src = inputs[i];
+            UMat& dst = outputs[i];
+
+            ocl::Kernel kernel("MishForward", ocl::dnn::activations_oclsrc, buildopt);
```
dkurt: Missed kernel implementation?

dkurt:
Hi! Thanks for contribution.

Please check that all the backends are valid (OpenCL, CUDA, Intel's Inference Engine). Otherwise keep only default C++ implementation.

modules/dnn/src/cuda4dnn/kernels/activations.hpp
```diff
@@ -18,6 +18,12 @@ namespace cv { namespace dnn { namespace cuda4dnn { namespace kernels {
     template <class T>
     void tanh(const csl::Stream& stream, csl::Span<T> output, csl::View<T> input);
 
+    template <class T>
+    void swish(const csl::Stream& stream, csl::Span<T> output, csl::View<T> input);
+
+    template <class T>
+    void mish(const csl::Stream& stream, csl::Span<T> output, csl::View<T> input);
```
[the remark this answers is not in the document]
thebhatman: I have added the kernel implementation. Can you please help me with adding tests?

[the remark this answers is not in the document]
YashasSamaga: @thebhatman The CUDA kernels are called by a CPU function which selects the correct vectorized kernel based on the alignment.

You can have a look at [sigmoid](https://github.com/opencv/opencv/blob/85e848a0f73b69d99ae2e068f4987e2fc09c16fd/modules/dnn/src/cuda/activations.cu#L281-L302) example. The `sigmoid()` calls `launch_vectorized_sigmoid` which actually launches the kernel.

1. `SwishOp` calls `swish<T>(stream, output, input)`
2. `swish()` checks the alignment of `output` and `input` and calls `launch_vectorized_swish`
3. `launch_vectorized_swish` selects the correct kernel and launches it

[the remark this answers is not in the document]
thebhatman: @YashasSamaga Thank you! I have added the function for launching the kernel.

modules/dnn/src/cuda/activations.cu
```diff
@@ -62,6 +62,44 @@ namespace cv { namespace dnn { namespace cuda4dnn  { namespace kernels {
...
+            auto input_vPtr = vector_type::get_pointer(input.data());
+
+            for (auto i : grid_stride_range(output.size() / vector_type::size())) {
+                vector_type vec;
+                v_load(vec, input_vPtr[i]);
+                for (int j = 0; j < vector_type::size(); j++) {
+                    using device::tanh;
+                    using device::log;
+                    using device::exp;
+                    vec.data[j] = vec.data[j] * tanh(log(1.0f + exp(vec.data[j])));
```
YashasSamaga: Shouldn't `1.0f` be `T(1.0)`? This will enable some compiler optimizations in half precision kernels.

thebhatman: Yes. That makes sense. Thank you!

asmorkalov:
cc @VadimLevin

dkurt:
@thebhatman, Please add the tests for both layers.

thebhatman:
Yeah I will add the tests. Where exactly are the tests for activations? Should I define sample neural networks with these activations ? Or is there a test file with all activation functions such  as sigmoid and tanh?

dkurt:
@thebhatman, So due the activations are simple, you can test them without test data. So you can just create a test which applies all the backends and compares with reference data (which is computed inside the test). Take a look at [test_layers.cpp](https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_layers.cpp) and [test_halide_layers.cpp](https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_halide_layers.cpp). The only parameter you need to vary is target and backend.

thebhatman:
I have added the tests in `test_halide_layers.cpp` and they are passing. It seems like all tests in `test_layers.cpp` use Caffe Models to test layers by reading data from a file from opencv_extra. Is there something else to be done?

modules/dnn/test/test_halide_layers.cpp
```diff
@@ -583,7 +583,7 @@ TEST_P(NoParamActivation, Accuracy)
     testInPlaceActivation(lp, backendId, targetId);
 }
 INSTANTIATE_TEST_CASE_P(Layer_Test_Halide, NoParamActivation, Combine(
-/*type*/ Values("TanH", "Sigmoid", "AbsVal", "BNLL"),
+/*type*/ Values("TanH", "Sigmoid", "AbsVal", "BNLL", "Swish", "Mish"),
```
dkurt: Good choice! 😄

thebhatman:
Is this PR ready to be merged now?

YashasSamaga:
To enable the CUDA backend, you have to tick the following CMake options:
- `WITH_CUDA`
- `WITH_CUDNN`
- `OPENCV_DNN_CUDA`

The CUDA build is failing on CI. Please have a look at https://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/2912

The CI does not run tests on CUDA devices. Hence, you (or someone) has to run them on their PC to verify that the tests are passing. Let me know when you are done with the PR, I'll build and test the PR once.

thebhatman:
The error is because of the log function ` using device::log;`.

thebhatman:
Thanks @YashasSamaga . The CUDA builds are now passing. They were failing due to type mismatch errors. I resolved them by using `log1pexp` instead of using log and exp separately.

modules/dnn/test/test_halide_layers.cpp
YashasSamaga: The CUDA tests are passing for the Swish activation.

The test for `DNN_TARGET_CUDA` is passing for the Mish activation. The test for `DNN_TARGET_CUDA_FP16` is failing due.

```
[ RUN      ] Layer_Test_Halide/NoParamActivation.Accuracy/21, where GetParam() = ("Mish", CUDA/CUDA_FP16)
.../opencv/modules/dnn/test/test_common.impl.hpp:68: Failure
Expected: (normL1) <= (l1), actual: 0.0288897 vs 0.004
.../opencv/modules/dnn/test/test_common.impl.hpp:71: Failure
Expected: (normInf) <= (lInf), actual: 0.0909604 vs 0.02
```

You will have to increase the error tolerance for Mish activation when target is `DNN_TARGET_CUDA_FP16`. I can't think of an easy and correct way to do this. In the worst case, the test can be skipped for the FP16 target.

dkurt:
@thebhatman, there is no still OpenCL kernels implementation. Please add it or remove unused `applyOCL`.

modules/dnn/test/test_halide_layers.cpp
thebhatman: The default Thresholds are being called from here https://github.com/thebhatman/opencv/blob/f6221044661dedbf4ad729c8f24c2b72ff37aca0/modules/dnn/test/test_common.hpp#L113. I don't see a way to pass the thresholds as arguments without changing the prototype of `testInPlaceActivation` .

YashasSamaga: I think you should add new parameters (`l1` and `lInf`) to `test` and `testInPlaceActivation`. Their default values would be `0.0`.

Inside the `test` function:
- if `l1` and `lInf` arguments are zero, set them to the values given by `getDefaultThresholds`
- if `l1` and `lInf arguments are not zero, use them

@dkurt is this  ok?

dkurt: @YashasSamaga, yes, that would be fine. Thanks!

modules/dnn/src/layers/elementwise_layers.cpp
```diff
@@ -613,6 +613,186 @@ struct TanHFunctor
...
+    {
+        Halide::Var x("x"), y("y"), c("c"), n("n");
+        top(x, y, c, n) = input / (1.0f + exp(-input));
+    }
+#endif  // HAVE_HALIDE
+
+#ifdef HAVE_INF_ENGINE
+    InferenceEngine::Builder::Layer initInfEngineBuilderAPI()
+    {
+        return InferenceEngine::Builder::SwishLayer("");
```
dkurt: There is no `SwishLayer` in this namespace. Please remove IE backend support for now.

thebhatman: I have removed the InfEngine backend support now. Thank you!

dkurt: Thanks!

Sorry, I forgot to highlight that you need to implement it in the following way (see `BNLLFunctor`):

```cpp
#ifdef HAVE_INF_ENGINE
    InferenceEngine::Builder::Layer initInfEngineBuilderAPI()
    {
        CV_Error(Error::StsNotImplemented, "");
    }
#endif  // HAVE_INF_ENGINE
```

thebhatman: Done. Thanks!

modules/dnn/src/layers/elementwise_layers.cpp
```diff
@@ -613,6 +613,186 @@ struct TanHFunctor
...
+struct SwishFunctor
+{
+    typedef SwishLayer Layer;
+
+    bool supportBackend(int backendId, int)
+    {
+        return backendId == DNN_BACKEND_OPENCV ||
+               backendId == DNN_BACKEND_CUDA ||
+               backendId == DNN_BACKEND_HALIDE ||
+               backendId == DNN_BACKEND_INFERENCE_ENGINE;
```
dkurt: Please remove IE backend from this check and similar one in Mish layer.

thebhatman: I have removed it  now.

cansik:
Has this already been released? In 4.3.0, I still get the error that it is unsupported:

```
OpenCV(4.3.0) /Users/travis/build/bytedeco/javacpp-presets/opencv/cppbuild/macosx-x86_64/opencv-4.3.0/modules/dnn/src/darknet/darknet_io.cpp:821: error: (-212:Parsing error) Unsupported activation: mish in function 'ReadDarknetFromCfgStream'
```

dkurt:
@cansik, If I'm not mistaken, this fix for TensorFlow importer. Please track https://github.com/opencv/opencv/issues/17148

YashasSamaga:
@BlueNotesRobot It is already there on master.

PR #15812, opened 2019-10-30
Use FlsAlloc/FlsFree/FlsGetValue/FlsSetValue instead of TlsAlloc/TlsFree/TlsGetValue/TlsSetValue to implment TLS value cleanup when thread has been terminated on Windows Vista and above

yuriyluxriot:
- OpenCV => 4.1.2-dev
- Operating System / Platform => Windows 7 64-bit

##### Detailed description

Use FlsAlloc/FlsFree/FlsGetValue/FlsSetValue instead of TlsAlloc/TlsFree/TlsGetValue/TlsSetValue to implment TLS operations and **cleanup of value** when thread has been terminated on Windows Vista and above.
Fls-functions operates on fiber local storage (FLS) and FlsAlloc allows to specify application-defined callback function. If the FLS slot is in use, then callback is called on fiber deletion, **thread exit**, and when an FLS index is freed. For more information see https://docs.microsoft.com/en-us/windows/win32/api/fibersapi/nf-fibersapi-flsalloc

In orer to use Fls-functions need to define _WIN32_WINNT to 0x0600 or above.

### DIFF
--- a/modules/core/src/system.cpp
+++ b/modules/core/src/system.cpp
@@ -131,9 +131,10 @@ void* allocSingletonNewBuffer(size_t size) { return malloc(size); }
 #if (_WIN32_WINNT >= 0x0602)
   #include <synchapi.h>
 #endif
-#if (_WIN32_WINNT >= 0x0600)
+#if ((_WIN32_WINNT >= 0x0600) && !defined(CV_DISABLE_FLS)) || defined(CV_FORCE_FLS)
   #include <fibersapi.h>
-#endif // _WIN32_WINNT >= 0x0600
+  #define CV_USE_FLS
+#endif
 #undef small
 #undef min
 #undef max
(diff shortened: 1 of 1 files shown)


modules/core/src/system.cpp
```diff
@@ -1359,24 +1362,41 @@ void  TlsAbstraction::SetData(void *pData)
     tlsData = pData;
 }
 #else //WINRT
+static void WINAPI opencv_fls_destructor(void* pData);
```
alalek: > WINAPI.

Should be `NTAPI`. See definition of `PFLS_CALLBACK_FUNCTION` from Windows headers.
(basically their values are equal)

yuriyluxriot: Well, in my current configuration I use **static linking**. In this case there is no DllMain at all and TLS data never cleaned up (w/o using FLS).
DllMain( DLL_THREAD_DETACH/DLL_PROCESS_DETACH, ... ) is called and every TLS (FLS) data is released as expected in **dynamic linking**.
Important note: if DLL is unloaded while thread with allocated FLS is still running, then DllMain( DLL_PROCESS_DETACH, ... ) will be called **before** any call of FLS cleanup. In this case releaseThread() will free every FLS data and no access violation occurs.

modules/core/src/system.cpp
```diff
@@ -1611,7 +1631,12 @@ static void opencv_tls_destructor(void* pData)
 {
     getTlsStorage().releaseThread(pData);
 }
-#endif
+#else	// _WIN32
+static void WINAPI opencv_fls_destructor(void* pData)
```
alalek: Add `#ifdef CV_USE_FLS` guard to avoid unused static function warning.

modules/core/src/system.cpp
```diff
@@ -131,6 +131,9 @@ void* allocSingletonNewBuffer(size_t size) { return malloc(size); }
 #if (_WIN32_WINNT >= 0x0602)
   #include <synchapi.h>
 #endif
+#if (_WIN32_WINNT >= 0x0600)
+  #include <fibersapi.h>
+#endif // _WIN32_WINNT >= 0x0600
```
alalek: Could you please replace this part of code:

```
#if ((_WIN32_WINNT >= 0x0600) && !defined(CV_DISABLE_FLS)) || defined(CV_FORCE_FLS)
#include <fibersapi.h>
#define CV_USE_FLS
#endif
```

and use these guards below:

```.diff
-#if (_WIN32_WINNT < 0x0600)
+#ifndef CV_USE_FLS
```

alalek:
Do not merge.
I will rebase your commits.

PR #15814, opened 2019-10-30
Simplify OpenCL info dumping code

i-murzov:
Simplify OpenCL info dumping code:

* Reduce code nesting
* Drop redundant .c_str() calls

Just a cleanup. Makes line lengths more reasonable.

### DIFF
--- a/modules/core/include/opencv2/core/opencl/opencl_info.hpp
+++ b/modules/core/include/opencv2/core/opencl/opencl_info.hpp
@@ -64,33 +64,30 @@ static void dumpOpenCLInformation()
 
         std::vector<PlatformInfo> platforms;
         cv::ocl::getPlatfomsInfo(platforms);
-        if (platforms.size() > 0)
-        {
-            DUMP_MESSAGE_STDOUT("OpenCL Platforms: ");
-            for (size_t i = 0; i < platforms.size(); i++)
-            {
-                const PlatformInfo* platform = &platforms[i];
-                DUMP_MESSAGE_STDOUT("    " << platform->name().c_str());
-                Device current_device;
-                for (int j = 0; j < platform->deviceNumber(); j++)
-                {
-                    platform->getDevice(current_device, j);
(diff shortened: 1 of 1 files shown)


modules/core/include/opencv2/core/opencl/opencl_info.hpp
```diff
@@ -64,33 +64,30 @@ static void dumpOpenCLInformation()
...
+        for (size_t i = 0; i < platforms.size(); i++)
+        {
+            const PlatformInfo* platform = &platforms[i];
+            DUMP_MESSAGE_STDOUT("    " << platform->name());
+            Device current_device;
+            for (int j = 0; j < platform->deviceNumber(); j++)
+            {
+                platform->getDevice(current_device, j);
+                const char* deviceTypeStr = (current_device.type() == Device::TYPE_CPU) ? "CPU" :
+                    (current_device.type() == Device::TYPE_GPU ? current_device.hostUnifiedMemory() ? "iGPU" : "dGPU" : "unknown");
```
cbachhuber: These two lines would benefit of even further refactoring, consider extracting a function for getting deviceTypeStr

cbachhuber:
Thank you for this PR, good work!

modules/core/include/opencv2/core/opencl/opencl_info.hpp
i-murzov: I'd like to avoid touching that part, because device.type() looks odd to me. It's type is int and not cl_device_type. It's also different from Device::TYPE_* enum, because condition of having iGPU is  "device.type() == Device::TYPE_GPU && device.hostUnifiedMemory()" instead of "device.type() == Device::TYPE_IGPU".

cbachhuber: I just did this refactoring in [#15883](https://github.com/opencv/opencv/pull/15883). I would appreciate a review from you!

PR #15826, opened 2019-11-01
cmake: fix ITT define condition

alalek:
`BUILD_ITT=OFF` in case of using ITT from [VTune](https://software.intel.com/en-us/vtune/choose-download).

Example of CMake options (Linux x86-64):

<cut/>

```
cmake -DBUILD_ITT=OFF -DHAVE_ITT=ON \
    -DITT_INCLUDE_DIRS=${VTUNE_AMPLIFIER_2019_DIR}/include \
    -DITT_LIBRARIES=${VTUNE_AMPLIFIER_2019_DIR}/lib64/libittnotify.a ...
```


alalek:
:+1:

PR #15828, opened 2019-11-01
Convert moments in tile algorithms to HAL (1.3x faster for VSX).

ChipKerchner:
Convert moments in tile algorithms to HAL (1.3x faster for VSX).

```
force_builders=ARMv7,ARMv8,Custom
buildworker:Custom=linux-1
#build_image:Custom=javascript-simd
build_image:Custom=mips64el
```

### DIFF
--- a/modules/imgproc/src/moments.cpp
+++ b/modules/imgproc/src/moments.cpp
@@ -213,7 +213,7 @@ struct MomentsInTile_SIMD
     }
 };
 
-#if CV_SIMD128
+#if CV_SIMD128_64F
 
 template <>
 struct MomentsInTile_SIMD<uchar, int, int>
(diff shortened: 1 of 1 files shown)


ChipKerchner:
Found a "bug" in the NEON version of HAL.  Creating registers does NOT initialize them to zero like SSE and VSX HAL.  Wondering if this should be fixed for consistency.

modules/imgproc/src/moments.cpp
```diff
@@ -265,6 +265,65 @@ struct MomentsInTile_SIMD<uchar, int, int>
...
+            v_x22_lo += v_x22_hi;
+            v_float64x2 v_x0122_lo, v_x0122_hi;
+            v_recombine(v_reinterpret_as_f64(v_x01_lo), v_reinterpret_as_f64(v_x22_lo), v_x0122_lo, v_x0122_hi);
+            v_store(buf64, v_reinterpret_as_s64(v_x3));
+            v_store(buf, v_reinterpret_as_s32(v_x0122_lo) + v_reinterpret_as_s32(v_x0122_hi));
+
+            x0 = buf[0];
+            x1 = buf[1];
+            x2 = buf[2];
+            x3 = buf64[0] + buf64[1];
```
alalek: Perhaps we don't need to force 64F here (at least in such complex form).

This is post-processing, so performance here is not critical.

Could you try this approach:
```
x0 = v_reduce_sum(v_x0);
x1 = v_reduce_sum(v_x1);
x2 = v_reduce_sum(v_x2);
int64 CV_DECL_ALIGNED(16) buf64[2];  // avoid declarations as class fields
v_store_aligned(buf64, v_reinterpret_as_s64(v_x3));
x3 = buf64[0] + buf64[1];
```

PR #15840, opened 2019-11-03
G-API: Unify GKernelType and GKernelTypeM logic (+ GNetworkType)

andrey-golubev:
### This pullrequest

- Unifies:
  * `GKernelType` and `GKernelTypeM`
  * `detail::MetaHelper` logic
  * `GNetworkType` logic
- `#define G_KERNEL_TYPE_M G_KERNEL_TYPE`
- Introduces `return_type_helper` to support single/multiple return-value differences

---
**Bottom line**: no difference between `G_KERNEL_TYPE_M` and `G_KERNEL_TYPE` now

### DIFF
--- a/modules/gapi/include/opencv2/gapi/gkernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/gkernel.hpp
@@ -47,8 +47,6 @@ struct GAPI_EXPORTS GKernelImpl
     util::any         opaque;    // backend-specific opaque info
 };
 
-template<typename, typename> class GKernelTypeM;
-
 namespace detail
 {
     ////////////////////////////////////////////////////////////////////////////
@@ -108,31 +106,6 @@ namespace detail
         return in_args.at(idx).template get<T>();
     }
 
-    // Helper for return type definition. Provides the actual logic of how return type is defined.
-    template<bool, typename...>
-    struct return_type_helper_impl;
-
-    // Helper for return type definition. Generalizes type definition for single and multiple return
-    // values in the context of G-API kernels
-    template<typename... Ts>
-    struct return_type_helper
-    {
-        using type = typename
-            return_type_helper_impl<std::tuple_size<std::tuple<Ts...>>::value == 1, Ts...>::type;
-    };
-
-    template<typename T>
-    struct return_type_helper_impl<true, T>
-    {
-        using type = T;
-    };
-
-    template<typename... Ts>
-    struct return_type_helper_impl<false, Ts...>
-    {
-        using type = std::tuple<Ts...>;
-    };
-
     // 4. The MetaHelper itself: an entity which generates outMeta() call
     //    based on kernel signature, with arguments properly substituted.
     // FIXME: probably can be simplified with std::apply or analogue.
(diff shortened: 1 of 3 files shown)


andrey-golubev:
@alalek is there a certain build with `CV_DOXYGEN` and will it be triggered by default. I have a feeling that something might break there.

@dmatveev please take a look

andrey-golubev:
Seems like ABI is broken now. I have couple of ideas how I can fix that, will experiment with that later on

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -141,87 +139,54 @@ namespace detail
...
-
-template<typename K, typename... R, typename... Args>
-class GKernelTypeM<K, std::function<std::tuple<R...>(Args...)> >
-    : public detail::MetaHelper<K, std::tuple<Args...>, std::tuple<R...>>
-    , public detail::NoTag
-{
-    template<int... IIs>
-    static std::tuple<R...> yield(cv::GCall &call, detail::Seq<IIs...>)
+    template<typename K, typename... R, typename... Args>
+    class GKernelTypeImpl<K, std::function<std::tuple<R...>(Args...)> >
```
andrey-golubev: self-review:
can partially specialize to accept:
```suggestion
    class GKernelTypeImpl<K, std::function<R(Args...)> >
```
seems like `R` (single-return-value) is good enough: can use `tuple_wrap_helper` where required for simple and correct behavior

dmatveev:
@andrey-golubev thanks! Somehow I managed to get rid of this problem when doing Inference API - and I believe it required less efforts than I see in this diff. Can you take a look there? `G_API_NET` and relevant stuff.

andrey-golubev:
@dmatveev you did: it was done using a `GKernelMedium` (or something like this), where you had a single class that would specialize depending on `R` or `tuple<Rs...>` and then subclass corresponding `GKernelType`/`GKernelTypeM` specialization.
However, here I eliminated:
 - One `MetaHelper` specialization
 - `GKernelTypeM` class

so the outcome is: 1 class for Kernel-level functionality + 1 class for MetaHelper-level functionality, instead of redundant specializations (x2 classes) that only differ in whether they accept 1 return type or many
\+ added some utility class (in some sense similar to `tuple_wrap_helper`), reused also in `G_API_NET`-based stuff
\+ unfortunately, in this design, had to also add some higher-level class aka what you did in `G_API_NET`

---
I will look into how ABI is constructed (specifically, function signatures) in C++ and think if it's possible to do the change without breaking anything, then I guess it would make more sense. But since there are many different ABIs, different implementations may exist and I wonder: even if I would manage to fix problems in 1 specific ABI, others would also magically work :) probably not, sadly

modules/gapi/include/opencv2/gapi/gkernel.hpp
andrey-golubev: self-answer: not possible because we need to unpack single `R` into `R` instead of `tuple<R>` - this is done nicely in current approach

asmorkalov:
@dmatveev @smirnov-alexey Do you have any progress on the patch? There are some merge conflicts.

andrey-golubev:
@asmorkalov I have a simpler solution to this. just no time to actually make a patch

asmorkalov:
@andrey-golubev do you plan to open another patch for it? If so, let's close this PR.

andrey-golubev:
@asmorkalov yep, I think it would be easier with a new PR

PR #15842, opened 2019-11-04
Fix YAML parse

TH3CHARLie:
resolves #14130

### This pullrequest changes
This pullrequest changes `core/persistence_yml.cpp` to fix incorrect YAML parse behavior. Current 4.x fails to read multiple documents from single file.

PR #15848, opened 2019-11-05
core(test): add test for YAML parse multiple documents

alalek:
- added removal of temporary file

backport #15842 #15851 (test)


asmorkalov:
:+1:

PR #15849, opened 2019-11-05
fix misleading documentation

### LINKED ISSUE #7265: IPP: pow() with non-integral exponent does not follow documentation on negative numbers
- OpenCV => 3.1
- Compiler => g++ 4.9.3

##### Detailed description

When compiling OpenCV with `-DWITH_IPP=OFF`, there is (typically) no problem. When compiling OpenCV with `-DWITH_IPP=ON` however, taking a non-integral power of a negative number x results in `-nan` instead of the expected |x|^e, which is what should happen according to:
http://docs.opencv.org/3.1.0/d2/de8/group__core__array.html#gaf0d056b5bd1dc92500d6f6cf6bac41ef

##### Steps to reproduce

```
#include <opencv/cv.hpp>
using namespace cv;

main(int argc, char** argv) {
    Mat test(1,1,CV_32FC1);
    test = -1.0;
    pow(test,0.4,test);
    printf("result: %f\n",test.at<float>(0));
}
```

When compiled with `-DWITH_IPP=OFF`, you get `result: 1.0`
When compiled with `-DWITH_IPP=ON`, you get `result: -nan`

For exponent 0.5, both the IPP and non-IPP version seem to be broken (probably because it is handled as a special case), and they both result in `-nan`.

Suggested solution:
Update the documentation to say that the result of a negative number to a non-integral power is undefined.

Some background:
I was using `pow` to gamma-adjust an image, where the values were not necessarily between 0.0 and 1.0 but could be a bit outside that range because of previous processing steps. I was assuming that values below 0.0 would be clamped to 0.0 before taking the power (which would be the useful thing to do here), but instead they would actually become positive. So thanks to this bug in OpenCV, I discovered a bug in my code. I can't think of any cases where the documented behavior is actually what you want, so I think it's better to explicitly say that the result is undefined (after all, mathematically the result is not a real number), than it is to 'fix' the IPP implementation of `pow`.

pasbi:
resolves #7265

### This pullrequest changes
The documentation about the `pow`-function.

### DIFF
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -1466,10 +1466,7 @@ CV_EXPORTS_W void sqrt(InputArray src, OutputArray dst);
 The function cv::pow raises every element of the input array to power:
 \f[\texttt{dst} (I) =  \texttt{src}(I)^{power}\f]
 
-So, for a non-integer power exponent, the absolute values of input array
-elements are used. However, it is possible to get true values for
-negative values using some extra operations. In the example below,
-computing the 5th root of array src shows:
+In the example below, computing the 5th root of array src shows:
 @code{.cpp}
     Mat mask = src < 0;
     pow(src, 1./5, dst);
--- a/modules/core/include/opencv2/core.hpp
+++ b/modules/core/include/opencv2/core.hpp
@@ -1463,8 +1463,8 @@ CV_EXPORTS_W void sqrt(InputArray src, OutputArray dst);
 
 /** @brief Raises every array element to a power.
 
-The function cv::pow raises every element of the input array to power :
-\f[\texttt{dst} (I) =  \fork{\texttt{src}(I)^{power}}{if \(\texttt{power}\) is integer}{|\texttt{src}(I)|^{power}}{otherwise}\f]
+The function cv::pow raises every element of the input array to power:
+\f[\texttt{dst} (I) =  \texttt{src}(I)^{power}\f]
 
 So, for a non-integer power exponent, the absolute values of input array
 elements are used. However, it is possible to get true values for


alalek:
Wrong, because documentation still not corresponds to implementation (actually different implementations: generic C++, SIMD, IPP, OpenCL) and contains conflicts itself (read the next paragraph after your changes).

pasbi:
Thanks for the comment, I've updated that paragraph.
I don't know details about these different implementations you've mentioned, however I think that my proposed docs are more correct than what is on master now.

Please accept my PR or update the docs yourself, it's a nightmare to have wrong docs (for more than three years).
I don't insist in closing #7265.

alalek:
It make sense to add `@note` from original issue (until we have a consensus between implementations + special `sqrt()` cases):

> the result of a negative number to a non-integral power is undefined.

(probably zero values on input are undefined too)

i-murzov:
The example below doesn't make much sense without the comment removed.

pasbi:
```
>>> import cv2, numpy as np
>>> cv2.pow(np.array([-1.0]), 1.0)
array([[-1.]])
>>> cv2.pow(np.array([-1.0]), 0.5)
array([[nan]])
>>> cv2.pow(np.array([-1.0]), 1.2345)
array([[nan]])
>>> cv2.pow(np.array([0.0]), 0.0)
array([[1.]])
>>> cv2.pow(np.array([0.0]), 1.0)
array([[0.]])
>>> cv2.pow(np.array([0.0]), -1.0)
array([[inf]])
```

I'd say this is pretty much what one expects from high-school math. What am I missing?

alalek:
Example in documentation doesn't make sense if we declare "undefined behavior" for input <= 0.

PR #15852, opened 2019-11-05
GAPI Fluid: SIMD optimization for sep filters of 5x5 kernel size (gaussBlur)

akhakim:
### This pullrequest changes

### DIFF
--- a/modules/gapi/src/backends/fluid/gfluidimgproc.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidimgproc.cpp
@@ -599,6 +599,7 @@ static void run_sepfilter(Buffer& dst, const View& src,
 {
     constexpr int kMax = 11;
     GAPI_Assert(kxLen <= kMax && kyLen <= kMax);
+    GAPI_Assert(kxLen == kyLen);
 
     const SRC *in[kMax];
           DST *out;
@@ -624,6 +625,13 @@ static void run_sepfilter(Buffer& dst, const View& src,
 
         int border = xborder;
         run_sepfilter3x3_impl(out, in, width, chan, kx, ky, border, scale, delta, buf, y, y0);
+    }
+    else if (kxLen == 5 && kyLen == 5)
+    {
+        int y = dst.y();
+        int y0 = dst.priv().writeStart();
+
+        run_sepfilter5x5_impl(out, in, width, chan, kx, ky, xborder, scale, delta, buf, y, y0);
     }
     else
     {
@@ -788,7 +796,9 @@ GAPI_FLUID_KERNEL(GFluidGaussBlur, cv::gapi::imgproc::GGaussBlur, true)
                               Buffer&    dst,
                               Buffer&    scratch)
     {
-        int kxsize = ksize.width;
+        GAPI_Assert(ksize.height == ksize.width);
+        GAPI_Assert((ksize.height == 3) || (ksize.height == 5));
+        const int kxsize = ksize.width;
         int kysize = ksize.height;
 
         auto *kx = scratch.OutLine<float>(); // cached kernX data
@@ -801,7 +811,7 @@ GAPI_FLUID_KERNEL(GFluidGaussBlur, cv::gapi::imgproc::GGaussBlur, true)
         constexpr int buffSize = 5;
         GAPI_Assert(ksize.height <= buffSize);
 
-        float *buf[buffSize]{};
+        float *buf[buffSize] = { nullptr };
 
         buf[0] = ky + kysize;
         for (int i = 1; i < ksize.height; ++i)
--- a/modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
+++ b/modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
@@ -100,6 +100,23 @@ RUN_SEPFILTER3X3_IMPL( float,  float)
 
 #undef RUN_SEPFILTER3X3_IMPL
 
+#define RUN_SEPFILTER5x5_IMPL(DST, SRC)                                     \
+void run_sepfilter5x5_impl(DST out[], const SRC *in[], int width, int chan, \
+                           const float kx[], const float ky[], int border,  \
+                           float scale, float delta,                        \
+                           float *buf[], int y, int y0);
+
+RUN_SEPFILTER5x5_IMPL(uchar, uchar)
+RUN_SEPFILTER5x5_IMPL(short, uchar)
+RUN_SEPFILTER5x5_IMPL(float, uchar)
+RUN_SEPFILTER5x5_IMPL(ushort, ushort)
+RUN_SEPFILTER5x5_IMPL(short, ushort)
+RUN_SEPFILTER5x5_IMPL(float, ushort)
+RUN_SEPFILTER5x5_IMPL(short, short)
+RUN_SEPFILTER5x5_IMPL(float, short)
+RUN_SEPFILTER5x5_IMPL(float, float)
+
+#undef RUN_SEPFILTER5x5_IMPL
 //-------------------------
 //
 // Fluid kernels: Filter 2D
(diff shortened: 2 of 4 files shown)


akhakim:
@dmatveev

elatkin:
@akhakim  
Could you please clarify: how to try accuracy and performance tests for this new kernel?  
I have tried your branch on my PC, but it is not obvious from test names if 5x5 was tested.

modules/gapi/perf/cpu/gapi_imgproc_perf_tests_fluid.cpp
```diff
@@ -58,7 +58,7 @@ INSTANTIATE_TEST_CASE_P(BlurPerfTestFluid, BlurPerfTest,
 INSTANTIATE_TEST_CASE_P(GaussianBlurPerfTestFluid, GaussianBlurPerfTest,
     Combine(Values(ToleranceFilter(1e-3f, 0.01).to_compare_f()),
             Values(CV_8UC1, CV_8UC3, CV_16UC1, CV_16SC1, CV_32FC1),
-            Values(3),                                     // TODO: add size=5, when kernel is ready
+            Values(3, 5),
```
akhakim: @elatkin 
As you can see here, I add numeral 5 to instantiation's parameters of performance tests for GaussianBlur. So now all performance tests run for 3x3 and 5x5 kernel sizes according to our test API.

modules/gapi/test/cpu/gapi_imgproc_tests_fluid.cpp
```diff
@@ -106,7 +106,7 @@ INSTANTIATE_TEST_CASE_P(gaussBlurTestFluid, GaussianBlurTest,
                                 Values(-1),
                                 Values(IMGPROC_FLUID),
                                 Values(ToleranceFilter(1e-3f, 0.01).to_compare_obj()),
-                                Values(3))); // add kernel size=5 when implementation is ready
+                                Values(3, 5))); // add kernel size=5 when implementation is ready
```
akhakim: As you can see here, I also added numeral 5 to instantiation's parameters of accuracy tests for fluid's GaussianBlur. So now all accuracy tests run not only for 3x3 but also for 5x5 kernel size according to our test API.

modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
```diff
@@ -1464,18 +1481,576 @@ void run_sepfilter3x3_impl(DST out[], const SRC *in[], int width, int chan,  \
...
+        run_sepfilter5x5_any2char<noscale>(reinterpret_cast<uchar*>(out), in,
+            width, chan, kx, ky, border, scale, delta,
+            buf, y, y0);
+        return;
+    }
+#endif  // CV_SIMD
+
+    // reference code is quite fast for any-to-float case,
+    // but not for any-to-integral due to very slow rounding
+    run_sepfilter5x5_reference<noscale>(out, in, width, chan, kx, ky, border,
```
rgarnov: All this dispatcher logic can be unified for any kernel size if you add template parameter for kernel size. Moreover, I believe that almost all the code for 3x3 and 5x5 can be unified with such approach

akhakim: I will create a separate task for this according to the Dmitry's request.

rgarnov: Ok

smirnov-alexey:
Looks OK

modules/gapi/src/backends/fluid/gfluidimgproc.cpp
```diff
@@ -800,10 +808,28 @@ GAPI_FLUID_KERNEL(GFluidGaussBlur, cv::gapi::imgproc::GGaussBlur, true)
...
+        else
+            buffLen = ksize.height;
+
         buf[0] = ky + kysize;
-        buf[1] = buf[0] + length;
-        buf[2] = buf[1] + length;
+        for (int i = 1; i < buffLen; ++i)
+        {
+            buf[i] = buf[i-1] + length;
+        }
```
dmatveev: I believe this changes are coming from fix #15874 right? Let's wait for that PR to be merged so than this one would only contain optimization-relevant changes

modules/gapi/src/backends/fluid/gfluidimgproc.cpp
```diff
@@ -800,10 +808,28 @@ GAPI_FLUID_KERNEL(GFluidGaussBlur, cv::gapi::imgproc::GGaussBlur, true)
...
 
-        float *buf[3];
+        const int buffSize = 5;
+        float *buf[buffSize];
+
+        int buffLen = 0;
+        if (ksize.height >= buffSize)
+            buffLen = buffSize;
+        else
+            buffLen = ksize.height;
```
elatkin: Probably:  
```c++
GAPI_Assert(ksize.height <= buffSize);
int buffLen = ksize.height;
```

modules/gapi/src/backends/fluid/gfluidimgproc.cpp
```diff
@@ -800,10 +808,28 @@ GAPI_FLUID_KERNEL(GFluidGaussBlur, cv::gapi::imgproc::GGaussBlur, true)
...
+            buf[i] = buf[i-1] + length;
+        }
+
+        if (buffSize > buffLen)
+        {
+            for (int j = buffLen; j < buffSize; ++j)
+            {
+                buf[j] = nullptr;
+            }
+        }
```
elatkin: No need in this cycle if you initialize `buf` with zeros, e.g.:  
```c++
float *buf[buffSize] = {0};
```

modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
```diff
@@ -1464,18 +1481,576 @@ void run_sepfilter3x3_impl(DST out[], const SRC *in[], int width, int chan,  \
...
+    {
+        r[n] = (y - y0 + n) % 5;  // previous, this, next rows
+    }
+
+    const int length = width * chan;
+    const int shift = chan;
+
+    // horizontal pass
+
+    int k0 = (y == y0) ? 0 : 3;
```
elatkin: Probably:
```c++
int k0 = (y == y0) ? 0 : 3;
```
The idea of `k0` is to decide if you need to process all **5** rows in the buffer or just the last one.  
Here you process only the last one if the previous rows were already done at previous passes.

modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
```diff
@@ -1464,18 +1481,576 @@ void run_sepfilter3x3_impl(DST out[], const SRC *in[], int width, int chan,  \
...
```
elatkin: I think **4**, not **3**:
```c++
int k0 = (y == y0) ? 0 : 4;
```

modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
```diff
@@ -1464,18 +1481,576 @@ void run_sepfilter3x3_impl(DST out[], const SRC *in[], int width, int chan,  \
...
+
+    const int length = width * chan;
+    const int shift = chan;
+
+    // horizontal pass
+    // full horizontal pass is needed only if the very 1st row in ROI is handled;
+    // for 2nd and further rows, it's enough to convolve only the
+    // "next" row - as we can reuse buffers from previous calls to
+    // this kernel (Fluid does rows consequently: y=y0, y0+1, ...)
+    int k0 = (y == y0) ? 0 : 3;
```
elatkin: Probably **4**, not **3**:
```C++
int k0 = (y == y0) ? 0 : 4;
```

modules/gapi/src/backends/fluid/gfluidimgproc_func.simd.hpp
```diff
@@ -1464,18 +1481,576 @@ void run_sepfilter3x3_impl(DST out[], const SRC *in[], int width, int chan,  \
...
+    int shift = chan;
+
+    // horizontal pass
+
+    // full horizontal pass is needed only if very 1st row in ROI;
+    // for 2nd and further rows, it is enough to convolve only the
+    // "next" row - as we can reuse buffers from previous calls to
+    // this kernel (Fluid does rows consequently: y=y0, y0+1, ...)
+
+    int k0 = (y == y0) ? 0 : 3;
```
elatkin:
```c++
int k0 = (y == y0) ? 0 : kyLen-1;
```

PR #15855, opened 2019-11-06
imgproc: fix bit-exact GaussianBlur() / sepFilter2D()

### LINKED ISSUE #9863: cv2.GaussianBlur produces darker results than input
- OpenCV => 3.3.0 master
- Operating System / Platform => Ubuntu 16.04 / x86_64
- Compiler => gcc 5.4

##### Detailed description

`GaussianBlur` produces darker results than the input image.

Original Image:

gaus 0

```python
>>> cv2.imwrite('gaus.30.cv2.png', cv2.GaussianBlur(im, (151, 151), 30))
```

gaus 30 cv2

```
$ gm convert gaus.0.png -blur 75x30 gaus.30.gm.png
```

gaus 30 gm

Images should look the same, but this is obvious, that produced by OpenCV is much darker.

### Proof that results produced by OpenCV is wrong

In theory, Gaussian blur shouldn't change color distribution or lightness. So, lets calculate lightness of the images:

```python
# Using Pillow
from PIL import Image 

def lightness(im):
    hist = im.histogram()
    return [
        sum(i*v for i, v in enumerate(hist[:256])) / im.width / im.height,
        sum(i*v for i, v in enumerate(hist[256:512])) / im.width / im.height,
        sum(i*v for i, v in enumerate(hist[512:])) / im.width / im.height
    ]

# [59.76, 69.08, 40.2]
print(lighnes(Image.open('gaus.0.png')))

# [59.73, 69.18, 39.94]
print(lighnes(Image.open('gaus.30.gm.png')))

# [53.86, 62.26, 36.22]
print(lighnes(Image.open('gaus.30.cv2.png')))
```

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/2342
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/685

resolves #9863

- avoid kernels with bad approximation
- GaussiabBlur - apply error-diffusion approximation for kernel (8-bit fraction)

<cut/>

TODO:
- [x] fix test from main repository
- [x] fix tests from opencv_contrib

### DIFF
--- a/modules/imgproc/src/smooth.dispatch.cpp
+++ b/modules/imgproc/src/smooth.dispatch.cpp
@@ -43,6 +43,10 @@
 
 #include "precomp.hpp"
 
+#include <opencv2/core/utils/logger.hpp>
+
+#include <opencv2/core/utils/configuration.private.hpp>
+
 #include <vector>
 
 #include "opencv2/core/hal/intrin.hpp"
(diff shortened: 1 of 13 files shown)


alalek:
:+1:

modules/imgproc/src/smooth.dispatch.cpp
```diff
@@ -477,6 +584,19 @@ static bool ipp_GaussianBlur(InputArray _src, OutputArray _dst, Size ksize,
 }
 #endif
 
+template<typename T>
+static bool validateGaussianBlurKernel(std::vector<T>& kernel)
+{
+    softdouble validation_sum = softdouble::zero();
+    for (size_t i = 0; i < kernel.size(); i++)
+    {
+        validation_sum += softdouble((double)kernel[i]);
```
netbrain: error: conversion from ‘__gnu_cxx::__alloc_traits<std::allocator<cv::{anonymous}::ufixedpoint16> >::value_type {aka cv::{anonymous}::ufixedpoint16}’ to ‘double’ is ambiguous
         validation_sum += softdouble((double)kernel[i]);

mshabunin: @netbrain , what is your platform/compiler?

netbrain:
Linux x86_64

svobora:
@alalek I do not think you full realize what you did by this fix. Now the speed of GaussianBlur depends on the ability to create a precise filter. That certainly works for small kernels, but not for large ones where the corner coefficients diminish. If the filter cannot be created to be precise, the blur fallbacks to float processing, which results in very high performance hit. 

For CV_8U images, kernel_size > 10, and reasonable sigma, the _createBitExactKernel_32S_ always returns false. Then fallback to CV_32F computing is processed. But some of us are more interested in performance, not the precision of computation. For example, on embedded devices the slowdown is about 10 fold. 

Could you add an ignore flag to bitexact computation, such that the user may choose between precision (run CV_32S if possible, fallback to CV_32F) and performance (always run CV_32S version)? Right now we can propably bypass the issue by generating the kernels on our end and filtering directly. But it would be a nice feature to satisfy performance seekers.

asmorkalov:
@svobora Thanks for the issue report. Could you file bug about the problem as dedicated issue with this description and link to the patch?

PR #15874, opened 2019-11-08
GAPI Fluid: The run_sepfilter() has logic error into snippet for 5x5 …

akhakim:
…and larger kernels

### This pullrequest changes

### DIFF
--- a/modules/gapi/perf/cpu/gapi_imgproc_perf_tests_fluid.cpp
+++ b/modules/gapi/perf/cpu/gapi_imgproc_perf_tests_fluid.cpp
@@ -58,7 +58,7 @@ INSTANTIATE_TEST_CASE_P(BlurPerfTestFluid, BlurPerfTest,
 INSTANTIATE_TEST_CASE_P(GaussianBlurPerfTestFluid, GaussianBlurPerfTest,
     Combine(Values(ToleranceFilter(1e-3f, 0.01).to_compare_f()),
             Values(CV_8UC1, CV_8UC3, CV_16UC1, CV_16SC1, CV_32FC1),
-            Values(3),                                     // TODO: add size=5, when kernel is ready
+            Values(3, 5),
             Values(szVGA, sz720p, sz1080p),
             Values(cv::compile_args(IMGPROC_FLUID))));
 
--- a/modules/gapi/src/backends/fluid/gfluidimgproc.cpp
+++ b/modules/gapi/src/backends/fluid/gfluidimgproc.cpp
@@ -628,7 +628,7 @@ static void run_sepfilter(Buffer& dst, const View& src,
     else
     {
         int length = chan * width;
-        int xshift = chan * xborder;
+        int xshift = chan;
 
         // horizontal pass
 
@@ -788,8 +788,6 @@ GAPI_FLUID_KERNEL(GFluidGaussBlur, cv::gapi::imgproc::GGaussBlur, true)
                               Buffer&    dst,
                               Buffer&    scratch)
     {
-        GAPI_Assert(ksize.height == 3);
-
         int kxsize = ksize.width;
         int kysize = ksize.height;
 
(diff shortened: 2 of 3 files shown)


akhakim:
@dmatveev @elatkin

modules/gapi/src/backends/fluid/gfluidimgproc.cpp
```diff
@@ -788,8 +788,6 @@ GAPI_FLUID_KERNEL(GFluidGaussBlur, cv::gapi::imgproc::GGaussBlur, true)
                               Buffer&    dst,
                               Buffer&    scratch)
     {
-        GAPI_Assert(ksize.height == 3);
```
rgarnov: Complete removal of this assert tells that the kernel now can work with arbitrary kernel size, but there is a hardcoded buffSize = 5 below, so it will not work with kernel sizes > 5

modules/gapi/src/backends/fluid/gfluidimgproc.cpp
```diff
@@ -800,10 +798,28 @@ GAPI_FLUID_KERNEL(GFluidGaussBlur, cv::gapi::imgproc::GGaussBlur, true)
...
+            buf[i] = buf[i - 1] + length;
+        }
+
+        if (buffSize > buffLen)
+        {
+            for (int j = buffLen; j < buffSize; ++j)
+            {
+                buf[j] = nullptr;
+            }
+        }
```
rgarnov: Can be reduced to
```
        constexpr int buffSize = 5;
        GAPI_Assert(ksize.height <= buffSize);

        float *buf[buffSize]{};

        buf[0] = ky + kysize;
        for (int i = 1; i < ksize.height; ++i)
        {
            buf[i] = buf[i - 1] + length;
        }
```

modules/gapi/test/cpu/gapi_imgproc_tests_fluid.cpp
```diff
@@ -106,7 +106,7 @@ INSTANTIATE_TEST_CASE_P(gaussBlurTestFluid, GaussianBlurTest,
                                 Values(-1),
                                 Values(IMGPROC_FLUID),
                                 Values(ToleranceFilter(1e-3f, 0.01).to_compare_obj()),
-                                Values(3))); // add kernel size=5 when implementation is ready
+                                Values(3, 5)));
```
rgarnov: BTW can we extend kernel sizes of all of the kernels which utilize `run_sippfilter`? (not in this PR maybe)

modules/gapi/perf/cpu/gapi_imgproc_perf_tests_fluid.cpp
```diff
@@ -58,7 +58,7 @@ INSTANTIATE_TEST_CASE_P(BlurPerfTestFluid, BlurPerfTest,
 INSTANTIATE_TEST_CASE_P(GaussianBlurPerfTestFluid, GaussianBlurPerfTest,
     Combine(Values(ToleranceFilter(1e-3f, 0.01).to_compare_f()),
             Values(CV_8UC1, CV_8UC3, CV_16UC1, CV_16SC1, CV_32FC1),
-            Values(3),                                     // TODO: add size=5, when kernel is ready
+            Values(3, 5),
```
rgarnov: Not sure if we should measure performance of non-optimized kernels

modules/gapi/src/backends/fluid/gfluidimgproc.cpp
akhakim: Applied.

modules/gapi/src/backends/fluid/gfluidimgproc.cpp
akhakim: Applied.

modules/gapi/test/cpu/gapi_imgproc_tests_fluid.cpp
akhakim: Unfortunately we can't extend kernel sizes of all kernels which utilize run_sepfilter() since 5x5 kernel size supports still wasn't added to all kernels.

modules/gapi/perf/cpu/gapi_imgproc_perf_tests_fluid.cpp
akhakim: It was enabled to show that there is the error in pointer arithmetic in case 5x5 kernel size and larger.

PR #15888, opened 2019-11-11
Introducing the sample of Face Beautification algorithm implemented via Graph-API

OrestChura:
## This pullrequest changes

- A new G-API sample that processes real-time face beautification pipeline is added. This is a Face Beautification algorithm implemented via G-API.
The sample shows the usage of G-API applying to a real CV problem. The G-API features used in the sample:

  - Networks inference INSIDE G-API (used two IE models: face detector (face-detection-adas-0001) and facial landmarks detector (facial-landmarks-35-adas-0002))

  - Custom CPU (OpenCV) kernels implementation

  - Brand new Streaming-API to process a videostream source INSIDE G-API

- FIXME added in 'gcpukernel.hpp'

```

test_modules:Custom=gapi
test_modules:Custom Win=gapi
test_modules:Custom Mac=gapi

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*

```

### DIFF
--- a/modules/gapi/samples/face_beautification.cpp
+++ b/modules/gapi/samples/face_beautification.cpp
@@ -4,8 +4,6 @@
 //
 // Copyright (C) 2018-2019 Intel Corporation
 
-#if defined(HAVE_OPENCV_GAPI) && defined(HAVE_INF_ENGINE)
-
 #include <opencv2/gapi.hpp>
 #include <opencv2/gapi/core.hpp>
 #include <opencv2/gapi/imgproc.hpp>
@@ -749,7 +747,7 @@ int main(int argc, char** argv)
     // Now we are ready to compile the pipeline to a stream with specified
     //  kernels, networks and image format expected to process
     auto stream = pipeline.compileStreaming(cv::GMatDesc{CV_8U,3,
-                                                         cv::Size(640,480)},
+                                                         cv::Size(1280,720)},
                                             cv::compile_args(kernels,
                                                              networks));
     // Setting the source for the stream:
@@ -783,15 +781,3 @@ int main(int argc, char** argv)
     }
     return 0;
 }
-
-#else // HAVE_OPENCV_GAPI && HAVE_INF_ENGINE
-#include <iostream>
-
-int main(int,char**)
-{
-    std::cerr << "This tutorial code requires G-API module"
-                 " with Inference Engine backend to run"
-              << std::endl;
-    return 1;
-}
-#endif // HAVE_OPECV_GAPI && HAVE_INF_ENGINE
--- a/modules/gapi/samples/face_beautification.cpp
+++ b/modules/gapi/samples/face_beautification.cpp
@@ -4,6 +4,12 @@
 //
 // Copyright (C) 2018-2019 Intel Corporation
 
+#if defined(HAVE_OPENCV_GAPI) && defined(HAVE_INF_ENGINE)
+
+#if INF_ENGINE_RELEASE <= 2018050000
+#   error G-API IE module supports only OpenVINO IE >= 2019 R1
+#endif
+
 #include <opencv2/gapi.hpp>
 #include <opencv2/gapi/core.hpp>
 #include <opencv2/gapi/imgproc.hpp>
@@ -608,12 +614,6 @@ int main(int argc, char** argv)
         return 0;
     }
 
-#if defined(HAVE_OPENCV_GAPI) && defined(HAVE_INF_ENGINE)
-
-#if INF_ENGINE_RELEASE <= 2018050000
-#   error G-API IE module supports only OpenVINO IE >= 2019 R1
-#endif
-
     cv::namedWindow(config::kWinFaceBeautification, cv::WINDOW_NORMAL);
     cv::namedWindow(config::kWinInput,              cv::WINDOW_NORMAL);
 
@@ -784,6 +784,8 @@ int main(int argc, char** argv)
 #else
 #include <iostream>
 
+int main(int,char**)
+{
     std::cerr << "This tutorial code requires G-API module"
                  " with Inference Engine backend to run"
               << std::endl;
--- a/modules/gapi/samples/face_beautification.cpp
+++ b/modules/gapi/samples/face_beautification.cpp
@@ -1,0 +1,777 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level
+// directory of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2018-2019 Intel Corporation
+
+#include <opencv2/gapi.hpp>
+#include <opencv2/gapi/core.hpp>
+#include <opencv2/gapi/imgproc.hpp>
+#include <opencv2/gapi/fluid/core.hpp>
+#include <opencv2/gapi/infer.hpp>
+#include <opencv2/gapi/infer/ie.hpp>
+#include <opencv2/gapi/cpu/gcpukernel.hpp>
+#include "opencv2/gapi/streaming/cap.hpp"
+
+#include <opencv2/videoio.hpp>
+#include <opencv2/highgui.hpp>
+#include <iomanip>
+
+namespace config
+{
+constexpr char       kWinFaceBeautification[] = "FaceBeautificator";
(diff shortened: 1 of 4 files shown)


OrestChura:
@dmatveev @aDanPin

OrestChura:
@dmatveev please approve it to be merged

modules/gapi/samples/face_beautification.cpp
```diff
@@ -4,6 +4,12 @@
 //
 // Copyright (C) 2018-2019 Intel Corporation
 
+#if defined(HAVE_OPENCV_GAPI) && defined(HAVE_INF_ENGINE)
+
+#if INF_ENGINE_RELEASE <= 2018050000
+#   error G-API IE module supports only OpenVINO IE >= 2019 R1
+#endif
```
dmatveev: You don't need it here. G-API's public API doesn't depend on either version of OpenVINO. This is an internal check and the library won't build itself because of this.

modules/gapi/samples/face_beautification.cpp
```diff
@@ -0,0 +1,796 @@
...
+{
+    static cv::GMatDesc outMeta(cv::GMatDesc in, cv::GArrayDesc)
+    {
+        return in.withType(CV_8U, 1);
+    }
+};
+
+G_TYPED_KERNEL(GPolyLines, <cv::GMat(cv::GMat,cv::GArray<Contour>,bool,
+                                     cv::Scalar)>,
+               "custom.faceb12n.polyLines")
```
dmatveev: The formatting still looks weird. I think I will align it on my own once it merged.

modules/gapi/samples/face_beautification.cpp
```diff
@@ -0,0 +1,796 @@
...
+                          std::vector<Landmarks> &vctPtsFaceElems,
+                          std::vector<Contour>   &vctCntJaw)
+    {
+        // There are 35 landmarks given by the default detector for each face
+        //  in a frame; the first 18 of them are face elements (eyes, eyebrows,
+        //  a nose, a mouth) and the last 17 - a jaw contour. The detector gives
+        //  floating point values for landmarks' normed coordinates relatively
+        //  to an input ROI (not the original frame).
+        //  For more details please visit:
+// https://github.com/opencv/open_model_zoo/blob/master/intel_models/facial-landmarks-35-adas-0002
```
dmatveev: Alignment is broken here

modules/gapi/samples/face_beautification.cpp
```diff
@@ -0,0 +1,796 @@
...
+            // If there's no data, let UI refresh (and handle keypress)
+            if (cv::waitKey(1) >= 0) break;
+            else continue;
+        }
+        cv::imshow(config::kWinInput,              imgShow);
+        cv::imshow(config::kWinFaceBeautification, imgBeautif);
+    }
+    return 0;
+
+#else
```
dmatveev: if this `#else` stands for this if `#if defined(HAVE_OPENCV_GAPI) && defined(HAVE_INF_ENGINE)`, then
1. Please put the appropriate comment - it is hard to track otherwise
2. Breaking the code to have `}` hanging at the last line is wrong. You better declare two distinct `main` functions and cut those with preprocessor fully rather than making reader guess why this `main` is not finished here but new `main` is introduced a couple of lines later.

modules/gapi/samples/face_beautification.cpp
OrestChura: OK. I figured out how to align it better, I think..

modules/gapi/samples/face_beautification.cpp
OrestChura: Yes, it is true, thank you

modules/gapi/samples/face_beautification.cpp
OrestChura: solved

modules/gapi/samples/face_beautification.cpp
OrestChura: fixed

dmatveev:
But wait, please put the necessary build configuration in the description to make sure sample is built ok. Take lines from my Streaming PR for example.

OrestChura:
> But wait, please put the necessary build configuration in the description to make sure sample is built ok. Take lines from my Streaming PR for example.

Yes, it make sence. A rebasing is done to rerun a buildbot. All checks have passed, although there is a single warning "CMake Warning: Manually-specified variables (ENABLE_CXX11, INF_ENGINE_RELEASE, WITH_INF_ENGINE) were not used by the project" while building Custom Win:gapi_standalone, ade. Is it important?

dmatveev:
> "CMake Warning: Manually-specified variables (ENABLE_CXX11, INF_ENGINE_RELEASE, WITH_INF_ENGINE) were not used by the project" while building Custom Win:gapi_standalone, ade. Is it important?

No, it is expected. Thanks!

dmatveev:
👍

OrestChura:
Wonderful! Thank you!

modules/gapi/samples/face_beautification.cpp
```diff
@@ -0,0 +1,797 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level
+// directory of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2018-2019 Intel Corporation
+
+#if defined(HAVE_OPENCV_GAPI) && defined(HAVE_INF_ENGINE)
```
alalek: > `HAVE_OPENCV_GAPI`

Like other OpenCV modules it can be tested only after including of `<opencv2/opencv_modules.hpp>`.
Anyway, it is redundant and should be removed - this is G-API sample (located in `gapi/samples`).

---

> `HAVE_INF_ENGINE`

This check is much worse. There is no header with this definition (`cvconfig.h` has some build options, but this header is private and it is not allowed in samples).

And even these definitions would not help to build sample with IE, because IE is not in dependency list of this sample.

This definition should be passed through CMake scripts.
After `ocv_add_samples()` you need to add IE dependency:
```.diff
 ocv_add_samples()
+if(TARGET example_gapi_face_beautification AND INF_ENGINE_TARGET)
+  target_link_libraries(example_gapi_face_beautification PRIVATE ${INF_ENGINE_TARGET})
+endif()
```

See these [build logs](http://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/2879/steps/compile%20release/logs/stdio) of "Custom" builder (find "face_beautification" and check its compiler options)

dmatveev: @alalek in fact `HAVE_INF_ENGINE` is not even required since G-API builds its IA API part even when there's no IE present

alalek: Right, IE is not used by sample directly (IE headers are not used, linker dependency is not needed).

But some IE-related symbols are missing in G-API:
```
undefined reference to `cv::gapi::ie::backend()'
```
Consider adding fallback implementation of `cv::gapi::ie::backend()` with something like:
```
CV_Error(Error::StsNotImplemented, "This OpenCV build doesn't include InferenceEngine support");
```

and then remove this check from sample (make it runtime by capturing `StsNotImplemented` if needed).

OrestChura: @dmatveev As I understand, you meant changes from your PR #15753, specifically commits [1](https://github.com/opencv/opencv/pull/15753/commits/655663620b2a18a09d678cb43023d404f4099a2c) and [2](https://github.com/opencv/opencv/pull/15753/commits/451d21a53f2c1869ed316cce270fef82c5d3281a); would you mind if I use them?

dmatveev: Yes, I thought that was merged already, but it you're all right the problem is still here. Bless you to steal these changes into your PR.

OrestChura: Thank you!

dmatveev:
@OrestChura wait wait wait, why it is under `modules/gapi/samples`? We store only basic code snippets there. Tutorial code is stored under `/samples/cpp/tutorial_code/gapi`. Pleave move it there with the next PR

PR #15898, opened 2019-11-12
Updating window_installation guide

### LINKED ISSUE #10645: OpenCV_DIR wrong in documentation?
I'm using OpenCV 2.4.13.5.

The [installation documentation for Windows](https://docs.opencv.org/master/d3/d52/tutorial_windows_install.html) says to set OPENCV_DIR to D:\OpenCV\Build\[arch]\[compiler], but for find_package to work, CMake looks for OpenCVConfig.cmake in OPENCV_DIR.

Setting OPENCV_DIR to D:\OpenCV\Build works on my machine.
Note that [node-opencv](https://github.com/peterbraden/node-opencv) expects OPENCV_DIR to be set according to the documentation to work, which is a pain when using both.

sajarindider:
resolves #10645

### This pullrequest changes

Adding an alternative method of installing OpenCV by specifying directory in the `CMakeList.txt` file.

### DIFF
--- a/doc/tutorials/introduction/windows_install/windows_install.markdown
+++ b/doc/tutorials/introduction/windows_install/windows_install.markdown
@@ -355,6 +355,12 @@ libraries). If you do not need the support for some of these, you can just freel
 Set the OpenCV environment variable and add it to the systems path {#tutorial_windows_install_path}
 =================================================================
 
+@note
+  Instead of adding it to your path you can add the following command to your `CMakeLists.txt` file: 
+@code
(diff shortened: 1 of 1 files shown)


alalek:
It is bad idea to hard code paths in your scripts.

sajarindider:
Hey @alalek, sorry about that. I'm still fairly new to the project, perhaps you can provide some direction. What do you feel is the best way to correct the documentation to reflect the problem outlined in the issue?

PR #15906, opened 2019-11-13
G-API (Fluid) core support for I420

anton-potapov:
###  Extended NV12 support in Fluid Core engine to cover I420

As only difference of I420 compared to NV12 is separate planes for U and V channels (instead of interleaved ones for NV12), changes are relatively small 

What is done:
- extended cv::gimpl::GFluidExecutable::initBufferRois to handle both ports of I420 to RGB conversion
- renaming from  from NV12 to 4:2:0 to express more generic behaviour 

@rgarnov @dmatveev  please take a look

### DIFF
--- a/modules/gapi/include/opencv2/gapi/fluid/gfluidkernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/fluid/gfluidkernel.hpp
@@ -50,7 +50,7 @@ public:
     {
         Filter,
         Resize,
-        NV12toRGB
+        YUV420toRGB //Color conversion of 4:2:0 chroma sub-sampling formats (NV12, I420 ..etc) to RGB
     };
 
     // This function is a generic "doWork" callback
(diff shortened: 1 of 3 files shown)


modules/gapi/include/opencv2/gapi/fluid/gfluidkernel.hpp
```diff
@@ -50,7 +50,7 @@ class GAPI_EXPORTS GFluidKernel
     {
         Filter,
         Resize,
-        NV12toRGB
+        CC420toRGB //Color conversion of 4:2:0 chroma sub-sampling formats (NV12, I420 ..etc) to RGB
```
rgarnov: It's better to name it `YUV420toRGB`

anton-potapov: renamed

anton-potapov:
@alalek please merge it

PR #15915, opened 2019-11-15
Fix implicit conversion from array to scalar in python bindings

### LINKED ISSUE #11426: norm with NORM_HAMMING returns strange results

VadimLevin:
### This pullrequest changes

numpy array was implicitly converted by `PyArg_ParseTupleAndKeywords` format string. After removal possible conflicting types from `simple_argtype_mapping` `pyopencv_to` parsers for them are generated, where additional checks are done, so correct overload is chosen. 

resolves #11426 

  - Fix implicit conversion of numpy array with 1 element to scalar
  - Add tests for norm


asmorkalov:
@VadimLevin The patch introduce 1000+ compile warnings. Please take a look: https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/23624/steps/compile%20release/logs/warnings%20%281490%29

VadimLevin:
@alalek Can you have a look to this change?

modules/python/test/test_norm.py
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env python
+
+from itertools import product
+from functools import reduce
+
+import cv2 as cv
+import numpy as np
```
alalek: Please import `numpy` before `cv2` (there are some reports related to importing order).

modules/python/test/test_norm.py
```diff
@@ -0,0 +1,162 @@
...
+    cv.NORM_HAMMING: 'Hamming',
+    cv.NORM_HAMMING2: 'Hamming2'
+}
+
+
+def get_element_types(norm_type):
+    if norm_type in (cv.NORM_HAMMING, cv.NORM_HAMMING2):
+        return (np.uint8,)
+    else:
+        return (np.int16, np.int32, np.float32, np.float64)
```
alalek: uint8, int8, uint16

modules/python/test/test_norm.py
```diff
@@ -0,0 +1,162 @@
...
+from itertools import product
+from functools import reduce
+
+import cv2 as cv
+import numpy as np
+
+from tests_common import NewOpenCVTests
+
+
+def norm_inf(x, y=None):
```
alalek: It make sense to convert x/y to doubles before passing to numpy (issue can be observed with `uint8`):
```
x = x.astype(np.float64)
y = y if y is None else y.astype(np.float64)
```

(here and below, except hamming)

modules/python/test/test_norm.py
```diff
@@ -0,0 +1,162 @@
...
+def get_element_types(norm_type):
+    if norm_type in (cv.NORM_HAMMING, cv.NORM_HAMMING2):
+        return (np.uint8,)
+    else:
+        return (np.int16, np.int32, np.float32, np.float64)
+
+
+def generate_vector(size, dtype):
+    if np.issubdtype(dtype, np.integer):
+        return np.random.randint(0, 100, size).astype(dtype)
```
alalek: We need reproducible tests, so we need reproducible tests input.

`np.random.randint` generates different numbers on each test launch.

Fix `np.random` seed or use `cv.randu` / `cv.randn`

modules/python/test/test_norm.py
```diff
@@ -0,0 +1,162 @@
...
+    else:
+        return 10. + 12.5 * np.random.randn(size).astype(dtype)
+
+
+sizes = (1, 2, 3, 5, 7, 19, 21, 32)
```
alalek: May be add some 2D size? (like `(3,5)`)

modules/python/test/test_norm.py
```diff
@@ -0,0 +1,162 @@
...
+        for norm_type, norm in norm_type_under_test.items():
+            element_types = get_element_types(norm_type)
+            for size, element_type in product(sizes, element_types):
+                array = generate_vector(size, element_type)
+                expected = norm(array)
+                actual = cv.norm(array, norm_type)
+                self.assertAlmostEqual(
+                    expected, actual, places=2,
+                    msg='Array {0} and norm {1}'.format(array,
+                                                        norm_name[norm_type])
```
alalek: It make sense to dump `element_type` too.

modules/python/src2/gen2.py
```diff
@@ -172,18 +174,48 @@
...
-    "int": ("int", "i", "0"),
-    "float": ("float", "f", "0.f"),
-    "double": ("double", "d", "0"),
-    "c_string": ("char*", "s", '(char*)""')
+    "bool": ArgTypeInfo("bool", FormatStrings.unsigned_char, "0"),
+    "size_t": ArgTypeInfo("size_t", FormatStrings.unsigned_long_long, "0", True),
+    "int": ArgTypeInfo("int", FormatStrings.int, "0", True),
+    "float": ArgTypeInfo("float", FormatStrings.float, "0.f", True),
+    "double": ArgTypeInfo("double", FormatStrings.double, "0", True),
+    "c_string": ArgTypeInfo("char*", FormatStrings.string, '(char*)""')
```
alalek: Looks good to me.

modules/python/test/test_norm.py
```diff
@@ -0,0 +1,162 @@
...
+    if np.issubdtype(dtype, np.integer):
+        return np.random.randint(0, 100, size).astype(dtype)
+    else:
+        return 10. + 12.5 * np.random.randn(size).astype(dtype)
+
+
+sizes = (1, 2, 3, 5, 7, 19, 21, 32)
+
+
+class CVNormTests(NewOpenCVTests):
```
alalek: `CVNormTests` => `norm_test` (like other tests)

alalek:
This change may break users code, so we need to ensure that impact would be low.

So we need tests. We can start with bindings arguments handling/conversions.
We still don't have them :disappointed:

Initial part of tests is here: https://github.com/alalek/opencv/commit/pr15915_r
(some cases are failed with this patch - due bugs in `pyopencv_to<int>()` implementation)

modules/python/src2/cv2.cpp
```diff
@@ -579,13 +586,27 @@ template<>
...
 {
     CV_UNUSED(info);
-    if(!obj || obj == Py_None)
+    if (!obj)
+    {
         return true;
-    int _val = PyObject_IsTrue(obj);
-    if(_val < 0)
+    }
+    if (obj == Py_None)
```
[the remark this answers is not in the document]
VadimLevin: I've got your point and agree with it. I will add default value check for None as a separate PR.

modules/python/src2/cv2.cpp
```diff
@@ -18,6 +18,13 @@
 
 #if PY_MAJOR_VERSION < 3
 #undef CVPY_DYNAMIC_INIT
+#else
+#define CV_PYTHON_3 1
+#endif
+
+// Python definition alias to prevent define collisions
+#if defined(PY_LONG_LONG)
+#define CV_HAVE_LONG_LONG 1
```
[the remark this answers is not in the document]
VadimLevin: Yep, check was changed.

modules/python/src2/cv2.cpp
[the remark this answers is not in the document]
VadimLevin: I've got your point and agree with it. I'll add default value check as a separate PR.

modules/python/src2/cv2.cpp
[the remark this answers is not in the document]
VadimLevin: Thanks for catch it up. I've changed check for appropriate type. To be removed.

modules/python/src2/cv2.cpp
```diff
@@ -597,11 +696,56 @@ PyObject* pyopencv_from(const size_t& value)
...
+            value = PyInt_AsUnsignedLongLongMask(obj);
+    #endif
+        }
+#endif
+        else
+        {
+            const bool isParsed = parseNumpyScalar<size_t>(obj, value);
+            if (!isParsed) {
+                failmsg("Argument '%s' can not be safely parsed to 'size_t'", info.name);
+            }
```
alalek: Perhaps "return false;" is missing here.
Also it make sense to extend tests to cover this branch too.

modules/python/src2/cv2.cpp
```diff
@@ -159,6 +165,155 @@ catch (const cv::Exception &e) \
...
+
+template<class T>
+bool parseNumpyScalar(PyObject* obj, T& value)
+{
+    if (PyArray_CheckScalar(obj))
+    {
+        // According to the numpy documentation:
+        // There are 21 statically-defined PyArray_Descr objects for the built-in data-types
+        PyArray_Descr* to = getNumpyTypeDescriptor<T>();
+        if (canBeSafelyCasted<T>(obj, to))
```
alalek: > `canBeSafelyCasted<T>`

This looks over complicated. Code doesn't use all 21 types, because it is OpenCV code (I see only 3 specializations for `parseNumpyScalar` - `size_t`, `float`, `double`).

I suggest to remove "stdx" replacement and just write these specializations (or overloads).

VadimLevin: The underlying idea of this generalized approach is to reduce code repetition and have a set well tested functions as building blocks. Basically, it is possible to cover all primitive types by 3 general functions:

- For signed integral types;
- For unsigned integral types;
- For floating point types.

And 1 specialized function for `size_t`. 

After primitive all primitive types are covered by this set of functions, it is possible to cover other cases with their combination.
For instance,  `vector<type>` can be parsed as a sequence without fixed size, `cv::Size` can be parsed as a sequence with size 2. To parse a single element generalized function should be called.

If a new type (e.g. `int64_t` can not be parsed now) will be added, it automatically be covered by signed integral parser, so the contributor should provide only tests for it.

These functions are not introduced in the current PR, but will be.

If you are against this idea, I'm OK to rewrite this part with parsers specializations.

alalek: 1. Vector conversions is another story. I still didn't get problems with them.


2.
> If a new type

Set of types for bindings should be limited to cover other than PYthon languages too.

---

Lets keep patch as much simple as possible.

VadimLevin: @alalek 
Added some test to illustrate current problem with sequential types parsing (such as `vector` or `Size` (sequence of size 2)):
https://github.com/VadimLevin/opencv/commit/07bdc4ad6c6d86627d20cdb203c27d03e7ac8652

For `vector<int>` skipped test either fails with some inner error or perform narrowing conversion (`double` -> `int`) which we want to eliminate with strong type check.

For `cv::Size` it simply rejects everything except tuple.

So the main goal of the underlying idea, that I tried to describe in the previous comment, is elimination of confusing and non-intuitive conversion behavior.
Implementation with templates may be a bit complicated, but in my opinion has more advantages rather disadvantages.

asmorkalov:
@alalek @VadimLevin Do you have any conclusion on the patch?

asmorkalov:
@alalek Please take a look. Vadim reduced the patch as discussed offline.

Issue #15895, opened 2019-11-12
Drawing into a numpy view with a negative step in the channel dimension doesn't work


tailsu:
##### System information (version)

- OpenCV => opencv-contrib-python-headless==4.1.1.26
- Operating System / Platform => OS X

##### Detailed description

I create three 3-channel images and draw a circle into them. In the second case I use the step trick `A[:, :, ::-1]` to flip the R and B channels in the image, emulating `cv.cvtColor` with `COLOR_BGR2RGB`. I then try to draw into that array. The result, as seen in the screenshot, is that nothing gets drawn. In the third case I explicitly make a copy of the channel-flipped array (which is actually a view) before drawing, and it works.

![image](https://user-images.githubusercontent.com/730130/68665520-7a88fd80-0542-11ea-8cab-531c56921bc8.png)

Expected is that cv.circle works in all three cases.

##### Steps to reproduce

```python
import cv2 as cv

image1 = np.zeros((100, 100, 3), dtype='uint8')
cv.circle(image1, (50, 50), 10, (255, 255, 255), -1)

image2 = np.zeros((100, 100, 3), dtype='uint8')
image2 = image2[:, :, ::-1]
cv.circle(image2, (50, 50), 10, (255, 255, 255), -1)

image3 = np.zeros((100, 100, 3), dtype='uint8')
image3 = np.array(image3[:, :, ::-1])
cv.circle(image3, (50, 50), 10, (255, 255, 255), -1)

imshow(image1, image2, image3)
```

tailsu:
`imshow` is a custom function that wraps `ipywidgets.Image`, but I tried it with `cv.imwrite` as well and the result is the same, so it's not a visualization issue.

alalek:
There is lack of support in OpenCV for numpy on-the-fly reshaped arrays. OpenCV can't even copy data back if layout is not supported.

Workaround: use return value.
```.diff
-cv.circle(image2, (50, 50), 10, (255, 255, 255), -1)
+image2 = cv.circle(image2, (50, 50), 10, (255, 255, 255), -1)
```

```.py
image1 = 128 * np.ones((100, 100, 3), dtype='uint8')
cv.circle(image1, (50, 50), 10, (255, 255, 255), -1)

image2 = 128 * np.ones((100, 100, 3), dtype='uint8')
image2 = image2[:, :, ::-1]
cv.circle(image2, (50, 50), 10, (255, 255, 255), -1)
[...]

PR #15919, opened 2019-11-15
python: force using of ArgInfo

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/2353

Using just "name" is not enough, especially in case of `UMat` and "output" parameter.

relates #15895

### DIFF
--- a/modules/python/src2/cv2.cpp
+++ b/modules/python/src2/cv2.cpp
@@ -34,15 +34,31 @@
 #include "pycompat.hpp"
 #include <map>
 
+class ArgInfo
+{
+public:
+    const char * name;
+    bool outputarg;
+    // more fields may be added if necessary
+
+    ArgInfo(const char * name_, bool outputarg_)
+        : name(name_)
+        , outputarg(outputarg_) {}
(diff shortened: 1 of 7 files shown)


modules/python/src2/cv2.cpp
```diff
@@ -34,15 +34,31 @@
...
+    bool outputarg;
+    // more fields may be added if necessary
+
+    ArgInfo(const char * name_, bool outputarg_)
+        : name(name_)
+        , outputarg(outputarg_) {}
+
+private:
+    ArgInfo(const ArgInfo&); // = delete
+    ArgInfo& operator=(const ArgInfo&); // = delete
```
cbachhuber: Why don't you delete these publicly and explicitly?

alalek: "= delete" requires C++11 (OpenCV 3.4 still using C++98)

cbachhuber: Sorry, didn't see that. LGTM, then :+1:

PR #15925, opened 2019-11-17
core(test): extending tests with SIMD C++ emulation code (intrin_cpp.hpp)

alalek:
```
buildworker:Custom=linux-1
build_image:Custom=fedora:31
```


alalek:
:+1:

PR #15934, opened 2019-11-18
Remove hard coded meta data from compileStreaming()

aDanPin:
Its a first stap to remove hard coded meta data from compileStreaming().
Now all nedded graph passes replaced to setSource().
But it's still can't extract metadata from first frame and reshape on running.

### DIFF
--- a/modules/gapi/include/opencv2/gapi/gcompiled.hpp
+++ b/modules/gapi/include/opencv2/gapi/gcompiled.hpp
@@ -14,6 +14,8 @@
 #include <opencv2/gapi/own/assert.hpp>
 #include <opencv2/gapi/garg.hpp>
 
+#include <ade/execution_engine/execution_engine.hpp>
+
 namespace cv {
 
 // This class represents a compiled computation.
@@ -207,6 +209,9 @@ public:
      */
     // FIXME: Why it requires compile args?
     void reshape(const GMetaArgs& inMetas, const GCompileArgs& args);
+
+    static void setMetaData(ade::Graph &g, cv::GCompileArgs &args
+                        , const cv::GMetaArgs &metas);
 
 protected:
     /// @private
--- a/modules/gapi/src/compiler/gcompiled.cpp
+++ b/modules/gapi/src/compiler/gcompiled.cpp
@@ -14,6 +14,8 @@
 
 #include "compiler/gcompiled_priv.hpp"
 #include "backends/common/gbackend.hpp"
+#include "compiler/passes/passes.hpp"
+#include "compiler/gcompiler.hpp"
 
 // GCompiled private implementation ////////////////////////////////////////////
 void cv::GCompiled::Priv::setup(const GMetaArgs &_metaArgs,
@@ -153,4 +155,17 @@ bool cv::GCompiled::canReshape() const
 void cv::GCompiled::reshape(const GMetaArgs& inMetas, const GCompileArgs& args)
 {
     m_priv->reshape(inMetas, args);
+}
+
+void cv::GCompiled::setMetaData(ade::Graph &g, cv::GCompileArgs &args
+                                        , const cv::GMetaArgs &metas)
+{
+    auto pass_ctx = ade::passes::PassContext{g};
+    cv::gimpl::passes::initMeta(pass_ctx, metas);
+
+    cv::gimpl::passes::inferMeta(pass_ctx, true);
+    //compile islands for m_orig_graph
+    cv::gimpl::passes::storeResultingMeta(pass_ctx);
+    // Get compileArgs from m_ops??
+    cv::gimpl::GCompiler::compileIslands(g, args);
 }
--- a/modules/gapi/src/compiler/gcompiler.cpp
+++ b/modules/gapi/src/compiler/gcompiler.cpp
@@ -259,15 +259,16 @@ cv::gimpl::GCompiler::GCompiler(const cv::GComputation &c,
                                                       // (no compound backend present here)
     m_e.addPass("kernels", "check_islands_content", passes::checkIslandsContent);
 
-    m_e.addPassStage("meta");
-    m_e.addPass("meta", "initialize",   std::bind(passes::initMeta, _1, std::ref(m_metas)));
-    m_e.addPass("meta", "propagate",    std::bind(passes::inferMeta, _1, false));
-    m_e.addPass("meta", "finalize",     passes::storeResultingMeta);
-    // moved to another stage, FIXME: two dumps?
-    //    m_e.addPass("meta", "dump_dot",     passes::dumpDotStdout);
-
-    // Special stage for backend-specific transformations
-    // FIXME: document passes hierarchy and order for backend developers
+    if (/*!m_metas.empty()*/ false) {
+        m_e.addPassStage("meta");
+        m_e.addPass("meta", "initialize",   std::bind(passes::initMeta, _1, std::ref(m_metas)));
+        m_e.addPass("meta", "propagate",    std::bind(passes::inferMeta, _1, false));
+        m_e.addPass("meta", "finalize",     passes::storeResultingMeta);
+        // moved to another stage, FIXME: two dumps?
+        //    m_e.addPass("meta", "dump_dot",     passes::dumpDotStdout);
+    }
+        // Special stage for backend-specific transformations
+        // FIXME: document passes hierarchy and order for backend developers
     m_e.addPassStage("transform");
 
     m_e.addPassStage("exec");
@@ -366,13 +367,17 @@ cv::gimpl::GCompiler::GPtr cv::gimpl::GCompiler::generateGraph()
     return makeGraph(m_c.priv().m_ins, m_c.priv().m_outs);
 }
 
-void cv::gimpl::GCompiler::runPasses(ade::Graph &g)
+void cv::gimpl::GCompiler:: runPasses(ade::Graph &g)
 {
     m_e.runPasses(g);
     GAPI_LOG_INFO(NULL, "All compiler passes are successful");
 }
 
-void cv::gimpl::GCompiler::compileIslands(ade::Graph &g)
+void cv::gimpl::GCompiler::compileIslands(ade::Graph &g) {
+    compileIslands(g, m_args);
+}
+
+void cv::gimpl::GCompiler::compileIslands(ade::Graph &g, cv::GCompileArgs &args)
 {
     GModel::Graph gm(g);
     std::shared_ptr<ade::Graph> gptr(gm.metadata().get<IslandModel>().model);
@@ -383,7 +388,7 @@ void cv::gimpl::GCompiler::compileIslands(ade::Graph &g)
     ade::passes::TopologicalSort{}(pass_ctx);
 
     // Now compile islands
-    GIslandModel::compileIslands(gim, g, m_args);
+    GIslandModel::compileIslands(gim, g, args);
 }
 
 cv::GCompiled cv::gimpl::GCompiler::produceCompiled(GPtr &&pg)
@@ -419,7 +424,7 @@ cv::GStreamingCompiled cv::gimpl::GCompiler::produceStreamingCompiled(GPtr &&pg)
 {
     const auto &outMetas = GModel::ConstGraph(*pg).metadata()
         .get<OutputMeta>().outMeta;
-    std::unique_ptr<GStreamingExecutor> pE(new GStreamingExecutor(std::move(pg)));
+    std::unique_ptr<GStreamingExecutor> pE(new GStreamingExecutor(std::move(pg), m_args));
 
     GStreamingCompiled compiled;
     compiled.priv().setup(m_metas, outMetas, std::move(pE));
@@ -436,10 +441,11 @@ cv::GCompiled cv::gimpl::GCompiler::compile()
 
 cv::GStreamingCompiled cv::gimpl::GCompiler::compileStreaming()
 {
+  //  m_args
     // FIXME: self-note to DM: now keep these compile()/compileStreaming() in sync!
     std::unique_ptr<ade::Graph> pG = generateGraph();
     GModel::Graph(*pG).metadata().set(Streaming{});
     runPasses(*pG);
-    compileIslands(*pG);
+//    compileIslands(*pG);
     return produceStreamingCompiled(std::move(pG));
 }
--- a/modules/gapi/src/compiler/gcompiler.cpp
+++ b/modules/gapi/src/compiler/gcompiler.cpp
@@ -259,7 +259,7 @@ cv::gimpl::GCompiler::GCompiler(const cv::GComputation &c,
                                                       // (no compound backend present here)
     m_e.addPass("kernels", "check_islands_content", passes::checkIslandsContent);
 
-    if (!m_metas.empty()) 
+    if (!m_metas.empty())
     {
         m_e.addPassStage("meta");
         m_e.addPass("meta", "initialize",   std::bind(passes::initMeta, _1, std::ref(m_metas)));
--- a/modules/gapi/src/compiler/gcompiler.cpp
+++ b/modules/gapi/src/compiler/gcompiler.cpp
@@ -392,19 +392,6 @@ void cv::gimpl::GCompiler::compileIslands(ade::Graph &g, cv::GCompileArgs &args)
     GIslandModel::compileIslands(gim, g, args);
 }
 
-void cv::gimpl::GCompiler::setMetaData(ade::Graph &g, cv::GCompileArgs &args
-                                        , const cv::GMetaArgs &metas)
-{
-    auto pass_ctx = ade::passes::PassContext{g};
-    cv::gimpl::passes::initMeta(pass_ctx, metas);
-
-    cv::gimpl::passes::inferMeta(pass_ctx, true);
-    //compile islands for m_orig_graph
-    cv::gimpl::passes::storeResultingMeta(pass_ctx);
-    // Get compileArgs from m_ops??
-    cv::gimpl::GCompiler::compileIslands(g, args);
-}
-
 cv::GCompiled cv::gimpl::GCompiler::produceCompiled(GPtr &&pg)
 {
     // This is the final compilation step. Here:
--- a/modules/gapi/src/compiler/gcompiler.cpp
+++ b/modules/gapi/src/compiler/gcompiler.cpp
@@ -259,6 +259,7 @@ cv::gimpl::GCompiler::GCompiler(const cv::GComputation &c,
                                                       // (no compound backend present here)
     m_e.addPass("kernels", "check_islands_content", passes::checkIslandsContent);
 
+    //Input metas may be empty when a graph is compiled for streaming
     if (!m_metas.empty())
     {
         m_e.addPassStage("meta");
@@ -268,8 +269,8 @@ cv::gimpl::GCompiler::GCompiler(const cv::GComputation &c,
         // moved to another stage, FIXME: two dumps?
         //    m_e.addPass("meta", "dump_dot",     passes::dumpDotStdout);
     }
-        // Special stage for backend-specific transformations
-        // FIXME: document passes hierarchy and order for backend developers
+    // Special stage for backend-specific transformations
+    // FIXME: document passes hierarchy and order for backend developers
     m_e.addPassStage("transform");
 
     m_e.addPassStage("exec");
@@ -281,6 +282,9 @@ cv::gimpl::GCompiler::GCompiler(const cv::GComputation &c,
     // (even if it is not actually required to produce a GCompiled).
     // FIXME: add a better way to do that!
     m_e.addPass("exec", "add_streaming",    passes::addStreaming);
+    // Note: Must be called after addStreaming as addStreaming pass
+    // can possibly add new nodes to the IslandModel
+    m_e.addPass("exec", "sort_islands",     passes::topoSortIslands);
 
     if (dump_path.has_value())
     {
@@ -368,13 +372,14 @@ cv::gimpl::GCompiler::GPtr cv::gimpl::GCompiler::generateGraph()
     return makeGraph(m_c.priv().m_ins, m_c.priv().m_outs);
 }
 
-void cv::gimpl::GCompiler:: runPasses(ade::Graph &g)
+void cv::gimpl::GCompiler::runPasses(ade::Graph &g)
 {
     m_e.runPasses(g);
     GAPI_LOG_INFO(NULL, "All compiler passes are successful");
 }
 
-void cv::gimpl::GCompiler::compileIslands(ade::Graph &g) {
+void cv::gimpl::GCompiler::compileIslands(ade::Graph &g)
+{
     compileIslands(g, m_args);
 }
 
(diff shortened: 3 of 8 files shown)


aDanPin:
@rgarnov , @dmatveev  , @OrestChura 
Please revew.

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -259,15 +259,17 @@ cv::gimpl::GCompiler::GCompiler(const cv::GComputation &c,
...
+    {
+        m_e.addPassStage("meta");
+        m_e.addPass("meta", "initialize",   std::bind(passes::initMeta, _1, std::ref(m_metas)));
+        m_e.addPass("meta", "propagate",    std::bind(passes::inferMeta, _1, false));
+        m_e.addPass("meta", "finalize",     passes::storeResultingMeta);
+        // moved to another stage, FIXME: two dumps?
+        //    m_e.addPass("meta", "dump_dot",     passes::dumpDotStdout);
+    }
+        // Special stage for backend-specific transformations
+        // FIXME: document passes hierarchy and order for backend developers
```
dmatveev: Wrong indentation level here

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -259,15 +259,17 @@ cv::gimpl::GCompiler::GCompiler(const cv::GComputation &c,
...
-    m_e.addPassStage("meta");
-    m_e.addPass("meta", "initialize",   std::bind(passes::initMeta, _1, std::ref(m_metas)));
-    m_e.addPass("meta", "propagate",    std::bind(passes::inferMeta, _1, false));
-    m_e.addPass("meta", "finalize",     passes::storeResultingMeta);
-    // moved to another stage, FIXME: two dumps?
-    //    m_e.addPass("meta", "dump_dot",     passes::dumpDotStdout);
-
-    // Special stage for backend-specific transformations
-    // FIXME: document passes hierarchy and order for backend developers
+    if (!m_metas.empty())
```
dmatveev: Please put a comment here that it is a valid situation when a graph is compiled for streaming

dmatveev: ALSO: Please document in the GCompiler's class constructor comment that `metas` parameter now can be empty

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -366,13 +368,17 @@ cv::gimpl::GCompiler::GPtr cv::gimpl::GCompiler::generateGraph()
     return makeGraph(m_c.priv().m_ins, m_c.priv().m_outs);
 }
 
-void cv::gimpl::GCompiler::runPasses(ade::Graph &g)
+void cv::gimpl::GCompiler:: runPasses(ade::Graph &g)
```
dmatveev: We don't put spaces after `::` usually

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -366,13 +368,17 @@ cv::gimpl::GCompiler::GPtr cv::gimpl::GCompiler::generateGraph()
...
 {
     m_e.runPasses(g);
     GAPI_LOG_INFO(NULL, "All compiler passes are successful");
 }
 
-void cv::gimpl::GCompiler::compileIslands(ade::Graph &g)
+void cv::gimpl::GCompiler::compileIslands(ade::Graph &g) {
```
dmatveev: Please follow the same coding style as in the rest of this file. `{` should be on a new line

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -417,9 +423,9 @@ cv::GCompiled cv::gimpl::GCompiler::produceCompiled(GPtr &&pg)
 
 cv::GStreamingCompiled cv::gimpl::GCompiler::produceStreamingCompiled(GPtr &&pg)
 {
-    const auto &outMetas = GModel::ConstGraph(*pg).metadata()
-        .get<OutputMeta>().outMeta;
-    std::unique_ptr<GStreamingExecutor> pE(new GStreamingExecutor(std::move(pg)));
+    std::unique_ptr<GStreamingExecutor> pE(new GStreamingExecutor(std::move(pg), m_args, m_metas));
+
+    GMetaArgs outMetas{};
```
dmatveev: What's the reason to have empty `outMetas` here? if `compileStreaming()` has been called WITH meta (hope this case is still supported), then the above code is perfectly handled.

You need to handle both cases here.

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -438,8 +444,15 @@ cv::GStreamingCompiled cv::gimpl::GCompiler::compileStreaming()
...
+    GModel::Graph gm(*pG);
+
+    gm.metadata().set(Streaming{});
     runPasses(*pG);
-    compileIslands(*pG);
+
+    std::shared_ptr<ade::Graph> gptr(gm.metadata().get<IslandModel>().model);
+    // Run topological sort on GIslandModel first
+    auto pass_ctx = ade::passes::PassContext{*gptr};
+    ade::passes::TopologicalSort{}(pass_ctx);
```
dmatveev: Why do we need it here? Sorting the island model topologically should be the part of `runPasses()` - inst it?

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -395,11 +396,13 @@ void collectorThread(std::vector<Q*> in_queues,
 }
 } // anonymous namespace
 
-cv::gimpl::GStreamingExecutor::GStreamingExecutor(std::unique_ptr<ade::Graph> &&g_model)
+cv::gimpl::GStreamingExecutor::GStreamingExecutor(std::unique_ptr<ade::Graph> &&g_model, cv::GCompileArgs m_args, const GMetaArgs &in_metas)
     : m_orig_graph(std::move(g_model))
     , m_island_graph(GModel::Graph(*m_orig_graph).metadata()
                      .get<IslandModel>().model)
     , m_gim(*m_island_graph)
+    , comp_args(m_args)
```
dmatveev: please follow the same naming convention (`m_`) for all member variables.

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
         util::throw_error(std::logic_error("Only one video source is"
                                            " currently supported!"));
     }
+ 
+    // testing meta data is cv::GMatDesc{CV_8U,3,cv::Size{768,576}}
```
dmatveev: this needs to be removed

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+    if (m_metas.empty()) {
+        // extract meta from first frame
```
dmatveev: this code seem to be missing

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+    } else {
+        //const auto& metas = m_metas;
```
dmatveev: ???

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+    }
+    
+    const auto& metas = m_metas;
```
dmatveev: ???

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+    
+    if (/*wasFinished*/true) {
```
dmatveev: ???

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+        auto pass_ctx = ade::passes::PassContext{*m_orig_graph.get()};
+        cv::gimpl::passes::initMeta(pass_ctx, metas);
+        //does inferMeta needed??
```
dmatveev: Yes it is needed. How do you propagate the metadata information in the graph otherwise?

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+    
+    if (/*wasFinished*/true) {
+        auto pass_ctx = ade::passes::PassContext{*m_orig_graph.get()};
+        cv::gimpl::passes::initMeta(pass_ctx, metas);
+        //does inferMeta needed?? 
+        cv::gimpl::passes::inferMeta(pass_ctx, true);
+        //compile islands for m_orig_graph
+        cv::gimpl::passes::storeResultingMeta(pass_ctx);
+        // Get compileArgs from m_ops?? 
+        cv::gimpl::GCompiler::compileIslands(*m_orig_graph.get(), comp_args);
```
dmatveev: Please move this partial compilation code somewhere to `gcompiler.cpp`, maybe making it a static method of `GCompiled`. We don't need _this_ level of details here, it should be kept closer to the general compiler logic. This `GStreamingExecutor` should be just an _user_ of that logic.

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+                = gm.metadata(orig_data_nh).get<Data>();
+            if (orig_data_info.shape == GShape::GARRAY) {
+                // FIXME: GArray lost host constructor problem
+                GAPI_Assert(!cv::util::holds_alternative<cv::util::monostate>(orig_data_info.ctor));
+            }
+            vec.emplace_back(RcDesc{ orig_data_info.rc
+                                   , orig_data_info.shape
+                                   , orig_data_info.ctor});
+            metas.emplace_back(orig_data_info.meta);
+        };
```
dmatveev: Why do we need it here? Ins't it what is done in the constructor? What's the need to repeat this code here?

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+            input_rcs.reserve(op.nh->inNodes().size());
+            // in_constants.reserve(nh->inNodes().size()); // FIXME: Ugly
+            output_rcs.reserve(op.nh->outNodes().size());
+            output_metas.reserve(op.nh->outNodes().size());
+
+            for (auto in_slot_nh  : op.nh->inNodes())  xtract_in(in_slot_nh,  input_rcs);
+            for (auto out_slot_nh : op.nh->outNodes()) xtract_out(out_slot_nh, output_rcs, output_metas);
+
+            op.in_objects = input_rcs;
+            op.out_objects = output_rcs;
```
dmatveev: It seems you don't need to duplicate here what the constructor does, and all you need is to update `out_metas` fields only. Then this code can be simplified dramatically.

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+            op.out_metas = output_metas;
+        }
+
+        wasFinished = true;
```
dmatveev: `wasFinished` is a bad name for a variable in a player-like class. It actually reflects if the partial graph compilation has been completed or not.

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+    } else {
+        bool canReshape = true;
+        for (auto &op : m_ops) {
+            canReshape = op.isl_exec->canReshape();
+            if (!canReshape)
+                break;
+        }
```
dmatveev: please use `std::all_of` instead of this

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -553,6 +555,98 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+            auto& g = *m_orig_graph.get();
+            ade::passes::PassContext ctx{g};
+            passes::initMeta(ctx, metas);
+            passes::inferMeta(ctx, true);
+            // outMetas()?
+            for (auto &op : m_ops)
+                op.isl_exec->reshape(g, comp_args);
+
+        } else {
+            //???
```
dmatveev: Throw an error here.

modules/gapi/src/executor/gstreamingexecutor.hpp
```diff
@@ -79,6 +79,11 @@ class GStreamingExecutor final
 
     cv::gimpl::GIslandModel::Graph m_gim; // FIXME: make const?
 
+    cv::GCompileArgs comp_args;
+    bool wasFinished = false;
```
dmatveev: Please follow the same `m_` naming convention for member class variables as everywhere.

modules/gapi/src/compiler/gcompiler.cpp
aDanPin: Comment is added

aDanPin: Comment in gcompiler.hpp is added

modules/gapi/src/compiler/gcompiler.cpp
aDanPin: Fixed

modules/gapi/src/compiler/gcompiler.cpp
aDanPin: Fixed

modules/gapi/src/compiler/gcompiler.cpp
aDanPin: Now outMetas is 
`const auto &outMetas = GModel::ConstGraph(*pg).metadata().get<OutputMeta>().outMeta;`

modules/gapi/src/compiler/gcompiler.cpp
aDanPin: This code replaced to passes 
It moved to
`void topoSortIslands(ade::passes::PassContext &ctx);`
now

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: Fixed

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: Removed

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: Removed

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: This code is removed.
Cases with extract meta data and reshape will be implemented in future versions

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: This code is removed.
Cases with extract meta data and reshape will be implemented in future versions

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: Some code is deleted now.
Shod `xtract_in ` and `xtract_out` be implemented as a usual function??

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: Some code is deleted now

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: This code is removed.
Cases with extract meta data and reshape will be implemented in future versions

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: This code is removed.
Cases with extract meta data and reshape will be implemented in future versions

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: This code is removed.
Cases with extract meta data and reshape will be implemented in future versions

modules/gapi/src/executor/gstreamingexecutor.hpp
aDanPin: Fixed

modules/gapi/src/compiler/gcompiler.cpp
aDanPin: Fixed

modules/gapi/src/executor/gstreamingexecutor.cpp
aDanPin: Now its a static method of `GCompiled`
`cv::GCompiled::setMetaData(ade::Graph &g, cv::GCompileArgs &args, const cv::GMetaArgs &metas)`

modules/gapi/src/executor/gstreamingexecutor.cpp
dmatveev: You should add it, actually :)

modules/gapi/src/executor/gstreamingexecutor.cpp
dmatveev: But how it is supposed to work now?

modules/gapi/src/executor/gstreamingexecutor.cpp
dmatveev: They are needed in a single place only, so it doesn't make a lot of sense

modules/gapi/include/opencv2/gapi/gcompiled.hpp
```diff
@@ -14,6 +14,8 @@
 #include <opencv2/gapi/own/assert.hpp>
 #include <opencv2/gapi/garg.hpp>
 
+#include <ade/execution_engine/execution_engine.hpp>
```
dmatveev: This header it not needed here.

modules/gapi/include/opencv2/gapi/gcompiled.hpp
```diff
@@ -208,6 +210,9 @@ class GAPI_EXPORTS GCompiled
     // FIXME: Why it requires compile args?
     void reshape(const GMetaArgs& inMetas, const GCompileArgs& args);
 
+    static void setMetaData(ade::Graph &g, cv::GCompileArgs &args
+                        , const cv::GMetaArgs &metas);
```
dmatveev: Please fix formatting. Also it is not only about _set_ting, it also triggers a complex recursive process on inferring metadata in the whole graph. So I'd rename it to `inferMeta`

modules/gapi/src/compiler/gcompiled.cpp
```diff
@@ -154,3 +156,16 @@ void cv::GCompiled::reshape(const GMetaArgs& inMetas, const GCompileArgs& args)
...
+{
+    auto pass_ctx = ade::passes::PassContext{g};
+    cv::gimpl::passes::initMeta(pass_ctx, metas);
+
+    cv::gimpl::passes::inferMeta(pass_ctx, true);
+    //compile islands for m_orig_graph
+    cv::gimpl::passes::storeResultingMeta(pass_ctx);
+    // Get compileArgs from m_ops??
+    cv::gimpl::GCompiler::compileIslands(g, args);
+}
```
dmatveev: This file is a wrong place for this. `GCompiler`, not `GCompiled`, is the place for all compiler-related logic.

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -417,9 +423,10 @@ cv::GCompiled cv::gimpl::GCompiler::produceCompiled(GPtr &&pg)
 
 cv::GStreamingCompiled cv::gimpl::GCompiler::produceStreamingCompiled(GPtr &&pg)
 {
+    std::unique_ptr<GStreamingExecutor> pE(new GStreamingExecutor(std::move(pg), m_args, m_metas));
```
dmatveev: What's the reason to pass `m_metas` in constructor? Isn't it passed down in `setup()`?

dmatveev: Isn't your `m_metas` vector empty at this point?

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -417,9 +423,10 @@ cv::GCompiled cv::gimpl::GCompiler::produceCompiled(GPtr &&pg)
...
+
     const auto &outMetas = GModel::ConstGraph(*pg).metadata()
         .get<OutputMeta>().outMeta;
```
dmatveev: Your `outMeta` may be missing if you didn't run meta inference. This code will likely throw.

I asked to support _both_ cases here.

modules/gapi/src/compiler/gcompiler.hpp
```diff
@@ -35,6 +35,8 @@ class GAPI_EXPORTS GCompiler
     void validateOutProtoArgs();
 
 public:
+    // metas may be empty in case when graph compiling for streaming
+    // In this case graph get metas from first frame
```
dmatveev: from the

modules/gapi/src/compiler/passes/exec.cpp
```diff
@@ -638,4 +638,12 @@ void passes::syncIslandTags(ade::passes::PassContext &ctx)
     GIslandModel::Graph gim(*gptr);
     GIslandModel::syncIslandTags(gim, ctx.graph);
 }
+
+void passes::topoSortIslands(ade::passes::PassContext &ctx)
+{
+    GModel::Graph gm(ctx.graph);
+    std::shared_ptr<ade::Graph> gptr(gm.metadata().get<IslandModel>().model);
```
dmatveev: `auto gptr = gm.metadata().get<IslandModel>().model`?

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -554,6 +556,40 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
...
+    const auto& metas = m_metas;
+    cv::GCompiled::setMetaData(*m_orig_graph.get(), m_comp_args, metas);
+
+    GModel::Graph gm(*m_orig_graph);
+
+    auto xtract_out = [&](ade::NodeHandle slot_nh, std::vector<RcDesc> &vec, cv::GMetaArgs &metas) {
+        const auto orig_data_nh
+            = m_gim.metadata(slot_nh).get<DataSlot>().original_data_node;
+        const auto &orig_data_info
+            = gm.metadata(orig_data_nh).get<Data>();
```
dmatveev: Here you have enough space to write this code normally (one line per statement)

dmatveev:
Please push the next update when it's complete

PR #15935, opened 2019-11-18
Fixed own::Mat::empty() for non-md case

rgarnov:
* In standard (non multi dimensional) case `dims` in `own::Mat` are empty, so in current implementation `own::Mat::empty()` will always return `false` in non-md case


rgarnov:
Thanks!

PR #15955, opened 2019-11-20
Tests for argument conversion of Python bindings generator

VadimLevin:
Added tests for Python bindings generated for functions with elemental types as input arguments

Should be treated as a part of tests for [15915](https://github.com/opencv/opencv/pull/15915)

Bad cases, those are resolved by [15915](https://github.com/opencv/opencv/pull/15915) are added as skipped tests and should be enabled in the following PR.

### DIFF
--- a/modules/core/include/opencv2/core/bindings_utils.hpp
+++ b/modules/core/include/opencv2/core/bindings_utils.hpp
@@ -21,6 +21,44 @@ CV_EXPORTS_W String dumpInputOutputArray(InputOutputArray argument);
 CV_EXPORTS_W String dumpInputOutputArrayOfArrays(InputOutputArrayOfArrays argument);
 
 CV_WRAP static inline
+String dumpBool(bool argument)
+{
+    return (argument) ? String("Bool: True") : String("Bool: False");
+}
+
+CV_WRAP static inline
+String dumpInt(int argument)
+{
+    return cv::format("Int: %d", argument);
+}
+
+CV_WRAP static inline
+String dumpSizeT(size_t argument)
+{
+    std::ostringstream oss("size_t: ", std::ios::ate);
+    oss << argument;
+    return oss.str();
+}
+
+CV_WRAP static inline
+String dumpFloat(float argument)
+{
+    return cv::format("Float: %.2f", argument);
+}
+
+CV_WRAP static inline
+String dumpDouble(double argument)
+{
+    return cv::format("Double: %.2f", argument);
+}
+
+CV_WRAP static inline
+String dumpCString(const char* argument)
+{
+    return cv::format("String: %s", argument);
+}
+
+CV_WRAP static inline
 AsyncArray testAsyncArray(InputArray argument)
 {
     AsyncPromise p;
--- a/modules/python/test/test_misc.py
+++ b/modules/python/test/test_misc.py
@@ -1,38 +1,66 @@
 #!/usr/bin/env python
 from __future__ import print_function
 
+import ctypes
+from functools import partial
+
 import numpy as np
 import cv2 as cv
 
-from tests_common import NewOpenCVTests
+from tests_common import NewOpenCVTests, unittest
+
+
+def is_numeric(dtype):
+    return np.issubdtype(dtype, np.integer) or np.issubdtype(dtype, np.floating)
+
+
+def get_limits(dtype):
+    if not is_numeric(dtype):
+        return None, None
+
+    if np.issubdtype(dtype, np.integer):
+        info = np.iinfo(dtype)
+    else:
+        info = np.finfo(dtype)
+    return info.min, info.max
+
+
+def get_conversion_error_msg(value, expected, actual):
+    return 'Conversion "{}" of type "{}" failed\nExpected: "{}" vs Actual "{}"'.format(
+        value, type(value).__name__, expected, actual
+    )
+
+
+def get_no_exception_msg(value):
+    return 'Exception is not risen for {} of type {}'.format(value, type(value).__name__)
 
 class Bindings(NewOpenCVTests):
 
     def test_inheritance(self):
         bm = cv.StereoBM_create()
-        bm.getPreFilterCap() # from StereoBM
-        bm.getBlockSize() # from SteroMatcher
+        bm.getPreFilterCap()  # from StereoBM
+        bm.getBlockSize()  # from SteroMatcher
 
         boost = cv.ml.Boost_create()
-        boost.getBoostType() # from ml::Boost
-        boost.getMaxDepth() # from ml::DTrees
-        boost.isClassifier() # from ml::StatModel
-
+        boost.getBoostType()  # from ml::Boost
+        boost.getMaxDepth()  # from ml::DTrees
+        boost.isClassifier()  # from ml::StatModel
 
     def test_redirectError(self):
         try:
-            cv.imshow("", None) # This causes an assert
+            cv.imshow("", None)  # This causes an assert
             self.assertEqual("Dead code", 0)
         except cv.error as _e:
             pass
 
         handler_called = [False]
+
         def test_error_handler(status, func_name, err_msg, file_name, line):
             handler_called[0] = True
 
         cv.redirectError(test_error_handler)
         try:
-            cv.imshow("", None) # This causes an assert
+            cv.imshow("", None)  # This causes an assert
             self.assertEqual("Dead code", 0)
         except cv.error as _e:
             self.assertEqual(handler_called[0], True)
@@ -40,7 +68,7 @@ class Bindings(NewOpenCVTests):
 
         cv.redirectError(None)
         try:
-            cv.imshow("", None) # This causes an assert
+            cv.imshow("", None)  # This causes an assert
             self.assertEqual("Dead code", 0)
         except cv.error as _e:
             pass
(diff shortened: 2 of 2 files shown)


modules/core/include/opencv2/core/bindings_utils.hpp
```diff
@@ -20,6 +20,18 @@ CV_EXPORTS_W String dumpInputOutputArray(InputOutputArray argument);
 
 CV_EXPORTS_W String dumpInputOutputArrayOfArrays(InputOutputArrayOfArrays argument);
 
+CV_EXPORTS_W String dumpBool(bool argument);
```
alalek: Lets don't expose this as an exported API of core module.

Please use "static inline" for these simple test wrappers.

modules/core/include/opencv2/core/bindings_utils.hpp
```diff
@@ -20,6 +20,18 @@ CV_EXPORTS_W String dumpInputOutputArray(InputOutputArray argument);
...
+
+CV_EXPORTS_W String dumpInt(int argument);
+
+CV_EXPORTS_W String dumpSizet(size_t argument);
```
alalek:
```.diff
-dumpSizet()
+dumpSizeT()
```

modules/python/test/test_misc.py
```diff
@@ -1,46 +1,64 @@
...
+import ctypes
+
 import numpy as np
 import cv2 as cv
 
 from tests_common import NewOpenCVTests
 
+
+def is_numeric(dtype):
+    return np.issubdtype(dtype, np.integer) or np.issubdtype(dtype, np.float)
```
alalek: Python 3.7 warns about changing of this:
```.diff
-np.float
+np.floating
```

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,312 @@ class Arguments(NewOpenCVTests):
...
+            msg='Exception is not risen for tuple slice with 1 element'
+        ):
+            _ = cv.utils.dumpInt((1, 2, 12)[0:1])
+
+        for not_convertible in (1.2, np.float(4), float(3), np.double(45), 's',
+                                'str', np.array([1, 2]), (1,), [1, 2],
+                                min_int - 1, max_int + 1):
+            with self.assertRaises(
+                (TypeError, OverflowError),
+                msg='Exception is not risen for {}'.format(not_convertible)
```
alalek: Please dump "type" too.

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,312 @@ class Arguments(NewOpenCVTests):
...
+                             ))
+
+        with self.assertRaises(
+            TypeError,
+            msg='Exception is not risen for tuple slice with 1 element'
+        ):
+            _ = cv.utils.dumpInt((1, 2, 12)[0:1])
+
+        for not_convertible in (1.2, np.float(4), float(3), np.double(45), 's',
+                                'str', np.array([1, 2]), (1,), [1, 2],
```
alalek: It make sense to add these cases too:
```
#np.float32(5),
np.float64(6),
#np.array([11], dtype=np.uint8),
```

(commented out - they are failed now, but we probably want them to be fixed).

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,312 @@ class Arguments(NewOpenCVTests):
...
+        with self.assertRaises(
+            TypeError,
+            msg='Exception is not risen for tuple slice with 1 element'
+        ):
+            _ = cv.utils.dumpSizet((3, 12, 123)[0:1])
+
+        self.assertEqual(
+            cv.utils.dumpSizet(4 - max_size_t),
+            'size_t: 5',
+            msg='size_t does not overflow properly'
```
alalek: Not sure if we really want this check.

VadimLevin:
@alalek I've applied all your review comments and added new conversion tests. 
Everything is Ok on Linux and Windows machines, but on MacOs I observe very strange behavior,
`error: (-4:Insufficient memory) Failed to allocate 18446744073709551568 bytes in function 'OutOfMemoryError'
" is risen for conversion 64 of type int`
It is very odd, because 64 is a standard int, but it tries to allocate more than 2^64 bytes of memory to perform such conversion....

VadimLevin:
@alalek I resolved all issues with the MacOS build, can you proceed with review?

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,575 @@ class Arguments(NewOpenCVTests):
...
-        c = np.array([[[1,2],[3,4],[5,6]]], dtype='f')
+        c = np.array([[[1, 2], [3, 4], [5, 6]]], dtype='f')
         res4 = cv.utils.dumpInputArrayOfArrays([c, a, b])
         self.assertEqual(res4, "InputArrayOfArrays: empty()=false kind=0x00050000 flags=0x01050000 total(-1)=3 dims(-1)=1 size(-1)=3x1 type(0)=CV_32FC2 dims(0)=2 size(0)=3x1 type(0)=CV_32FC2")
 
+    def test_parse_to_bool(self):
+        for convertible in (True, False, 1, 0, 64, np.bool(1), np.uint8(0),
+                            np.int8(123), np.int16(11), np.int32(2),
+                            np.int64(1)):
+            expected = 'Bool: {}'.format(bool(convertible)).lower()
```
alalek: Could you please split this set into explicit true/false subsets?
- `convertible_true`
- `convertible_false`

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,575 @@ class Arguments(NewOpenCVTests):
...
+                                 ))
+
+        with self.assertRaises(
+            TypeError,
+            msg='Exception is not risen for tuple slice with 1 element'
+        ):
+            _ = cv.utils.dumpBool((True, False, False)[0:1])
+
+        for not_convertible in (1.2, np.float(2.3), 's', 'str', (1, 2),
+                                [1, 2], a, 1 + 1j, 2j, np.complex(1.1),
```
alalek: > 1 + 1j

consider using use straightforward `complex(1, 1)` ([`j` looks like a joke](https://stackoverflow.com/questions/24812444/why-are-complex-numbers-in-python-denoted-with-j-instead-of-i))

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,575 @@ class Arguments(NewOpenCVTests):
...
+                                np.complex(-0.5, 1.), np.complex(imag=3j)):
```
alalek: > np.complex(imag=3j)

I believe it should throw itself.
Please remove this case - there is already test for `<type 'complex'>`.

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,575 @@ class Arguments(NewOpenCVTests):
...
+                    not_convertible, type(not_convertible).__name__
+                )
+            ):
+                _ = cv.utils.dumpBool(not_convertible)
+
+    @unittest.skip('Wrong conversion behavior')
+    def test_parse_to_bool_extra(self):
+        _, max_size_t = get_limits(ctypes.c_size_t)
+        for convertible in (-1, max_size_t):
+            expected = 'Bool: {}'.format(bool(convertible)).lower()
```
alalek: Both are "true" - lets add a note or use `convertible_true`

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,575 @@ class Arguments(NewOpenCVTests):
...
+                (TypeError, OverflowError),
+                msg='Exception is not risen for {} of type {}'.format(
+                    not_convertible, type(not_convertible).__name__
+                )
+            ):
+                _ = cv.utils.dumpBool(not_convertible)
+
+    def test_parse_to_int(self):
+        min_int, max_int = get_limits(ctypes.c_int)
+        for convertible in (-10, -1, 2, True, False, int(43.2), np.uint8(15),
```
alalek: > True, False

Not sure that we really want them in this list (at least for type-safety and better overloads handling - we have bool+int parameters in a row).

/cc @mshabunin

modules/core/src/bindings_utils.cpp
```diff
@@ -141,5 +141,4 @@ CV_EXPORTS_W String dumpInputOutputArrayOfArrays(InputOutputArrayOfArrays argume
     }
     return ss.str();
 }
-
```
alalek: Please revert unnecessary change.

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,575 @@ class Arguments(NewOpenCVTests):
...
+                (TypeError, OverflowError),
+                msg='Exception is not risen for {} of type {}'.format(
+                    not_convertible, type(not_convertible).__name__
+                )
+            ):
+                _ = cv.utils.dumpSizeT(not_convertible)
+
+    def test_parse_to_float(self):
+        min_float, max_float = get_limits(ctypes.c_float)
+        for convertible in (2, True, False, -13, 1.24, float(32),
```
alalek: > float
> True, False

IMHO, should move to `non_convertible` list.

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,575 @@ class Arguments(NewOpenCVTests):
...
+            except Exception as e:
+                self.fail(
+                    '{} "{}" is risen for conversion {} of type {}'.format(
+                        type(e).__name__, e, nan,
+                        type(nan).__name__
+                    )
+                )
+
+        min_double, max_double = get_limits(ctypes.c_double)
+        for inf in (min_float * 10, max_float * 10, min_double, max_double):
```
alalek: Why is `* 10` needed?

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,575 @@ class Arguments(NewOpenCVTests):
...
+                msg='Exception is not risen for {} of type {}'.format(
+                    not_convertible, type(not_convertible).__name__
+                )
+            ):
+                _ = cv.utils.dumpFloat(not_convertible)
+
+    def test_parse_to_double(self):
+        min_float, max_float = get_limits(ctypes.c_float)
+        min_double, max_double = get_limits(ctypes.c_double)
+        for convertible in (2, True, False, -13, 1.24, np.float(32.45),
```
alalek: > True, False

the same

modules/python/test/test_misc.py
VadimLevin: Done

modules/python/test/test_misc.py
VadimLevin: Done

modules/python/test/test_misc.py
VadimLevin: I removed numpy complex alias.

`complex(imag=3j) == complex(-3)`

modules/python/test/test_misc.py
VadimLevin: I added boolean to convertible list to state implicit conversion behavior.

I think we should leave overload resolution to C++ backend. For example, converter will indicate if such conversion occurred.

modules/python/test/test_misc.py
VadimLevin: To check that if mantissa exceeds max value that can be held by float type it becomes inf.

Maybe this check is unnecessary, because max/min _double checks the same case.

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,538 @@ class Arguments(NewOpenCVTests):
...
+        b = np.array([[1, 2, 3], [4, 5, 6], [7, 8, 9]])
         res3 = cv.utils.dumpInputArrayOfArrays([a, b])
         self.assertEqual(res3, "InputArrayOfArrays: empty()=false kind=0x00050000 flags=0x01050000 total(-1)=2 dims(-1)=1 size(-1)=2x1 type(0)=CV_32SC1 dims(0)=2 size(0)=2x3 type(0)=CV_32SC1")
-        c = np.array([[[1,2],[3,4],[5,6]]], dtype='f')
+        c = np.array([[[1, 2], [3, 4], [5, 6]]], dtype='f')
         res4 = cv.utils.dumpInputArrayOfArrays([c, a, b])
         self.assertEqual(res4, "InputArrayOfArrays: empty()=false kind=0x00050000 flags=0x01050000 total(-1)=3 dims(-1)=1 size(-1)=3x1 type(0)=CV_32FC2 dims(0)=2 size(0)=3x1 type(0)=CV_32FC2")
 
+    def test_parse_to_bool(self):
+        def try_to_convert(value):
```
mshabunin: As I can see all `try_to_convert` functions are the same, can we make a single method which will accept function to be called?
```.py
def try_to_convert(self, fun, value):
    try:
        result = fun(value).lower()
    except Exception as e:
        ...
    else:
        return result
```

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,538 @@ class Arguments(NewOpenCVTests):
...
+            self.assertEqual('bool: false', actual,
+                             msg="Can't convert {0} of type {1}\n"
+                             "Expected: 'bool: false' vs Actual: {2}".format(
+                                 convertible_false,
+                                 type(convertible_false).__name__,
+                                 actual
+                             ))
+
+        a = np.array([1, 2, 0], dtype=np.bool)
+        for i in range(len(a)):
```
mshabunin: Is it necessary to check each value in the array, can it be moved to previous blocks?
```.py
for convertible_true in (..., np.array([1, 2, 0], dtype=np.bool)[1]):
...
for convertible_false in (..., np.array([1, 2, 0], dtype=np.bool)[2]):
```

Same comment for all following test cases.

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,538 @@ class Arguments(NewOpenCVTests):
...
+            actual = try_to_convert(a[i])
+            self.assertEqual(expected, actual,
+                             msg="Can't convert array element {0} of type {1}\n"
+                             "Expected: {2} vs Actual: {3}".format(
+                                 a[i],
+                                 type(a[i]).__name__,
+                                 expected, actual
+                             ))
+
+        with self.assertRaises(
```
mshabunin: Same with this case, can it be moved to the following block?

Same for other test cases.

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,538 @@ class Arguments(NewOpenCVTests):
...
+            TypeError,
+            msg='Exception is not risen for tuple slice with 1 element'
+        ):
+            _ = cv.utils.dumpBool((True, False, False)[0:1])
+
+        for not_convertible in (1.2, np.float(2.3), 's', 'str', (1, 2),
```
mshabunin: Perhaps it makes sense to split this test case: `test_parse_to_bool`, `test_not_parse_to_bool`?

Same for other test cases.

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,538 @@ class Arguments(NewOpenCVTests):
...
+                                complex(1.1), None):
+            with self.assertRaises(
+                (TypeError, OverflowError),
+                msg='Exception is not risen for {} of type {}'.format(
+                    not_convertible, type(not_convertible).__name__
+                )
+            ):
+                _ = cv.utils.dumpBool(not_convertible)
+
+    @unittest.skip('Wrong conversion behavior')
```
mshabunin: What is the plan for skipped tests: are we going to fix them? (is it possible?) Or maybe we should declare these cases unsupported and move them to negative tests?

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,538 @@ class Arguments(NewOpenCVTests):
...
+                        type(e).__name__, e, value, type(value).__name__
+                    )
+                )
+            else:
+                return result
+
+        for convertible_true in (True, 1, 64, np.bool(1), np.int8(123),
+                                 np.int16(11), np.int32(2), np.int64(1)):
+            actual = try_to_convert(convertible_true)
+            self.assertEqual('bool: true', actual,
```
mshabunin: We have 100 character line length limit recommendation in [the style guide](https://github.com/opencv/opencv/wiki/Coding_Style_Guide#file-structure). Look like you've used 80 character width restriction, I think in some cases the code would be more readable with relaxed requirement.

modules/python/test/test_misc.py
```diff
@@ -50,39 +68,538 @@ class Arguments(NewOpenCVTests):
...
+                             msg="Can't convert {0} of type {1}\n"
```
mshabunin: There is repeated pattern `"... {0} of type {1} ...".format(val, type(val).__name__)` everywhere in the code, I think it should be extracted to the utility function.

modules/python/test/test_misc.py
alalek: There is also "expected" part (+workarounds for Win32), so they are not the same.

modules/python/test/test_misc.py
alalek: Plan is to fix them with patch from here: #15915 (should be enabled there)

These tests purpose is ensuring that we don't break something else.

modules/python/test/test_misc.py
VadimLevin: If this check will be moved to "convertible" block, it won't be possible to determine if conversion error occurs when array was indexed or just normal check.

modules/python/test/test_misc.py
VadimLevin: Same reason as for array indexing.

modules/python/test/test_misc.py
mshabunin: It can be determined from current iteration index.

VadimLevin: Hmm, I think we can completly eliminate this check if we add array element type to appropriate list. E.G:
```python
type(np.array([1, 2], dtype=np.bool)[0]) # yields to np.bool_
type(np.array([1, 2], dtype=ctypes.c_size_t)[0]) # yields to np.uint64
type(np.array(['t1', 't2'])[0]) # yields to np.str_ which is np.str
```

modules/python/test/test_misc.py
VadimLevin: Tuple slice has type tuple than can be removed, because checked.

modules/python/test/test_misc.py
VadimLevin: Yes, thank you

modules/python/test/test_misc.py
VadimLevin: Done

modules/python/test/test_misc.py
VadimLevin: Sounds good, done.

modules/python/test/test_misc.py
```diff
@@ -1,88 +1,332 @@
...
+        for convertible in (max_size_t,):
+            expected = 'size_t: {0:d}'.format(convertible).lower()
+            actual = try_to_convert(convertible)
+            self.assertEqual(expected, actual,
+                             msg=get_conversion_error_msg(convertible, expected, actual))
+
+        for not_convertible in (np.array([123, ], dtype=np.uint8),):
+            with self.assertRaises(
+                (TypeError, OverflowError),
+                msg='Exception is not risen for {} of type {}'.format(
```
mshabunin: `msg=get_no_exception_msg(not_convertible)`

PR #15957, opened 2019-11-20
Fix cudacodec python

cudawarped:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/2362/files

### This pullrequest adds python bindings to cudacodec

Because cudacodec is in it's own namespace, python bindings are not currently generated for it.   This PR fixes this and adds tests to confirm the new functionality is working.

### DIFF
--- a/modules/python/src2/hdr_parser.py
+++ b/modules/python/src2/hdr_parser.py
@@ -911,9 +911,12 @@ class CppHeaderParser(object):
                         if stmt_type.startswith("enum"):
                             decls.append([stmt_type + " " + self.get_dotted_name(name), "", [], decl, None, ""])
                         else:
-                            decls.append(decl)
+                            if "cv.cudacodec.VideoReader.nextFrame" not in decl[0]:
+                                # Disable Mat decl for specific cudacodec function which can take zero args, to allow python call to
+                                # fall through to correct GpuMat decl
+                                decls.append(decl)
 
-                            if self._generate_gpumat_decls and "cv.cuda." in decl[0]:
+                            if self._generate_gpumat_decls and "cv.cuda" in decl[0]:
                                 # If function takes as one of arguments Mat or vector<Mat> - we want to create the
                                 # same declaration working with GpuMat (this is important for T-Api access)
                                 args = decl[3]
--- a/modules/python/src2/hdr_parser.py
+++ b/modules/python/src2/hdr_parser.py
@@ -86,6 +86,10 @@ class CppHeaderParser(object):
             modlist.append("/O")
             arg_str = arg_str.replace("CV_OUT", "")
 
+        if "CV_GPU" in arg_str:
+            modlist.append("/G")
+            arg_str = arg_str.replace("CV_GPU", "")
+
         if "CV_IN_OUT" in arg_str:
             modlist.append("/IO")
             arg_str = arg_str.replace("CV_IN_OUT", "")
@@ -911,7 +915,9 @@ class CppHeaderParser(object):
                         if stmt_type.startswith("enum"):
                             decls.append([stmt_type + " " + self.get_dotted_name(name), "", [], decl, None, ""])
                         else:
-                            decls.append(decl)
+                            has_gpu_mat = any(flag == '/G' for var in decl[3] for flag in var[3])
+                            if not has_gpu_mat:
+                                decls.append(decl)
 
                             if self._generate_gpumat_decls and "cv.cuda" in decl[0]:
                                 # If function takes as one of arguments Mat or vector<Mat> - we want to create the
--- a/modules/python/src2/hdr_parser.py
+++ b/modules/python/src2/hdr_parser.py
@@ -911,10 +911,7 @@ class CppHeaderParser(object):
                         if stmt_type.startswith("enum"):
                             decls.append([stmt_type + " " + self.get_dotted_name(name), "", [], decl, None, ""])
                         else:
-                            if "cv.cudacodec.VideoReader.nextFrame" not in decl[0]:
-                                # Disable Mat decl for specific cudacodec function which can take zero args, to allow python call to
-                                # fall through to correct GpuMat decl
-                                decls.append(decl)
+                            decls.append(decl)
 
                             if self._generate_gpumat_decls and "cv.cuda" in decl[0]:
                                 # If function takes as one of arguments Mat or vector<Mat> - we want to create the
--- a/modules/python/src2/hdr_parser.py
+++ b/modules/python/src2/hdr_parser.py
@@ -86,10 +86,6 @@ class CppHeaderParser(object):
             modlist.append("/O")
             arg_str = arg_str.replace("CV_OUT", "")
 
-        if "CV_GPU" in arg_str:
-            modlist.append("/G")
-            arg_str = arg_str.replace("CV_GPU", "")
-
         if "CV_IN_OUT" in arg_str:
             modlist.append("/IO")
             arg_str = arg_str.replace("CV_IN_OUT", "")
@@ -915,13 +911,11 @@ class CppHeaderParser(object):
                         if stmt_type.startswith("enum"):
                             decls.append([stmt_type + " " + self.get_dotted_name(name), "", [], decl, None, ""])
                         else:
-                            has_gpu_mat = any(flag == '/G' for var in decl[3] for flag in var[3])
-                            if not has_gpu_mat:
-                                decls.append(decl)
+                            decls.append(decl)
 
                             if self._generate_gpumat_decls and "cv.cuda" in decl[0]:
                                 # If function takes as one of arguments Mat or vector<Mat> - we want to create the
-                                # same declaration working with GpuMat (this is important for T-Api access)
+                                # same declaration working with GpuMat
                                 args = decl[3]
                                 has_mat = len(list(filter(lambda x: x[0] in {"Mat", "vector_Mat"}, args))) > 0
                                 if has_mat:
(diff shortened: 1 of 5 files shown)


modules/python/src2/hdr_parser.py
```diff
@@ -911,9 +911,12 @@ def parse(self, hname, wmode=True):
                         if stmt_type.startswith("enum"):
                             decls.append([stmt_type + " " + self.get_dotted_name(name), "", [], decl, None, ""])
                         else:
-                            decls.append(decl)
+                            if "cv.cudacodec.VideoReader.nextFrame" not in decl[0]:
+                                # Disable Mat decl for specific cudacodec function which can take zero args, to allow python call to
+                                # fall through to correct GpuMat decl
```
alalek: Try this instead:
```
ret, gpu_mat = .nextFrame(frame=cv.cuda_GpuMat())
```

or add separate wrapper for GPU-only (remove WRAP from `nextFrame(OutputArray)`):
```
CV_WRAP inline bool nextFrame(CV_OUT GpuMat& frame) { return nextFrame(OutputArray(frame); }
```

cudawarped: I have tried

> CV_WRAP inline bool nextFrame(CV_OUT GpuMat& frame, Stream& stream = Stream::Null()) { return nextFrame(OutputArray(frame), stream); }

because it needs to work with streams.  The problem is I can't figure out how to generate bindings so both 

> ret,gpu_mat = _reader.nextFrame()

and

> ret,_ = _reader.nextFrame(gpu_mat)

work.

The first case really needs to work because it is really useful to be able to interrogate the size of the expected output `GPUMat` when using the python interface.  If not then it can just silently not work because the size is wrong.

The second case definitely needs to work otherwise python will allocate a new `GPUMat` every time a video frame is read.

cudawarped: @alalek 
The latest changes to the python bindings generation code are to ensure that

> CV_WRAP inline bool nextFrame(CV_OUT GpuMat& frame, Stream& stream = Stream::Null()) { return nextFrame(frame, stream); }

will work with both

> ret,_ = _reader.nextFrame(gpu_mat)

and

> ret,gpu_mat = _reader.nextFrame()

alalek: Please run it locally - it doesn't work.

> test_cudacodec (test_cuda.cuda_test) ...
OpenCV(4.1.2-dev) Error: The function/feature is not implemented (getGpuMat is available only for cuda::GpuMat and cuda::HostMem) in getGpuMat, file modules/core/src/matrix_wrap.cpp, line 359
skipped 'NVCUVID is not installed'

(generated overloads are in this order: "Mat", "GpuMat", "UMat")

This works somehow:

```
ret,gpu_mat = _reader.nextFrame(cv.cuda_GpuMat())
self.assertTrue(ret)
print(type(gpu_mat))  # <== add check for this
#print(cv.utils.dumpInputArray(gpu_mat)) # - no support for GpuMat
```

cudawarped: Is that with the change to 

> CV_WRAP inline bool nextFrame(CV_OUT GpuMat& frame, Stream& stream = Stream::Null()) { return nextFrame(frame, stream); }

alalek: Got it.
Could you please create PR into opencv_contrib too (branch "fix_cudacodec_python")?

cudawarped: @alalek 
Would it make more sense to have an additional flag to signal that the `OutputArray` is a `GpuMat` instead of adding additional overides?  For example adding `CV_GPU`

` CV_WRAP virtual bool nextFrame(CV_GPU OutputArray frame, Stream &stream = Stream::Null()) = 0;`

and excluding the Mat bindings when it is present

```
if "CV_GPU" in arg_str:
    modlist.append("/G")
    arg_str = arg_str.replace("CV_GPU", "")
```

```
has_gpu_mat = any(flag == '/G' for var in decl[3] for flag in var[3])
if not has_gpu_mat:
    decls.append(decl)
```

I guess this could them be extended to work if there was a mixture of array types to a function.

alalek: > excluding the Mat bindings

Perhaps reordering should be better.

---

> `if "CV_GPU" in arg_str:`

Not sure if we need separate notation for that. Lets try to reuse this condition:
```
if self._generate_gpumat_decls and "cv.cuda" in decl[0]:
```

cudawarped: Can you expand on what you mean by reordering?  If we move the `Mat` bindings below `GpuMat ` then `GpuMat.download()` will fail.

alalek: > move the Mat bindings below GpuMat

Yes, I mean this. `Mat` overload below `GpuMat`

---

> then GpuMat.download() will fail

Is this for case with result of `Mat` type? (so we need to remove Mat completely)
This make sense for outputs (like `OutputArray`), but perhaps we should accept `InputArray` as numpy (`Mat`) array. Anyway, lets try with "excluding the Mat bindings".

cudawarped: Yes the problem is that `OutputArray` could be a `Mat` or `GpuMat` and currently there is no way to tell until the input is parsed.  This works fine if the python function has input arguments because the correct bindings will be chosen when this happens.  

Currently this also works when there are no input arguments because `GpuMat.download()` is the only function this applies to, so returning a `Mat` works.  Unfortunately this fails for `nextFrame()`.  Therefore if we reorder it will work for `nextFrame()` but fail for `GpuMat.download()`.

> This make sense for outputs (like OutputArray), but perhaps we should accept InputArray as numpy (Mat) array. Anyway, lets try with "excluding the Mat bindings".

I agree and I think that signaling that `OutputArray` is a `GpuMat` will allow input arrays which are a mixture of both types to be processed.  Currently I don't think there are any cuda functions which take a mixture of `Mat` and `GpuMat` but I may be wrong.

cudawarped:
> `or self.tp in ["GpuMat", "cuda::GpuMat"]`

If the function signature includes the cuda namespace, i.e.

`CV_WRAP inline bool nextFrame(CV_OUT cuda::GpuMat& frame, Stream &stream = Stream::Null()) { return nextFrame(OutputArray(frame), stream); }`

the compilation will fail because `self.tp==cuda_GpuMat`.

Is there a way to tell the type of the `InputArray`, without using an additional flag (`CV_GPU`), in order create python bindings for functions like `SURF_CUDA::uploadKeypoints(const std::vector<KeyPoint>& keypoints, GpuMat& keypointsGPU);` which take a mixture of host and device vectors?

alalek:
Just don't include `cuda::` for GpuMat in OpenCV API.

Mixture of parameters is not supported. Current overloads can handle all `GpuMat` or all `Mat` parameters, not mix of both. Trying to handle mix would explode generated code in 2**n times.
Anyway, `SURF_CUDA::uploadKeypoints()` should work (there is no `InputArray` at all).

cudawarped:
> Just don't include cuda:: for GpuMat in OpenCV API.

Fair enough, should it be documented somewhere, that cuda functions with only an `OutputArray` or `InputArray` need to be overloaded for the python bindings to work.  And that the `cuda` namespace should be excluded from the declaration?

Is `cuda::GpuMat` in

`or self.tp in ["GpuMat", "cuda::GpuMat"]`

redundant?

> Trying to handle mix would explode generated code in 2**n times.
> Anyway, SURF_CUDA::uploadKeypoints() should work (there is no InputArray at all).

This was a bad example but I would have thought the size of the generated code would be the same, because you would have a single wrapper for the function containing a mixture of input types?

cudawarped:
> "cuda::GpuMat" was added [here](https://github.com/opencv/opencv/pull/12156/files#diff-3de02a2e921c9580d1e703b0b0c4487aR407). If it don't work, then lets remove that.

Would you prefer
```
        return self.tp == "Mat" or self.tp == "vector_Mat" \
               or self.tp == "GpuMat" \
               or self.tp == "UMat" or self.tp == "vector_UMat" # or self.tp.startswith("vector")
```
or
```
       return self.tp in ["Mat", "vector_Mat", "GpuMat", "UMat", "vector_UMat"] 
```

cudawarped:
> Trying to handle mix would explode generated code in 2**n times

If size is an issue wouldn't it be better to add a flag to `download()`, e.g. `CV_MAT`

`CV_WRAP void download(CV_MAT OutputArray dst, Stream& stream) const;`

and then all of the CUDA function wrappers except `download()` can be defined for  `GPUMat` only removing the redundant `UMat` and `Mat` processing?

alalek:
Please avoid adding of CV_GPU/CV_MAT modifiers.
They look like hacks to hide implementation specifics (wrote InputArray/OutputArray, but really can't handle set of *Array storage types) and they have no future (because with this approach you would use 2**n super set of these modifiers or much worse `n!` due ordering issues)

> removing the redundant

Remove wrapping of generic `InputArray`, and just write "inline" specialized wrapper.
Or like in C++, use `frame=cv.cuda_GpuMat()` approach to specify exact wanted output type.

cudawarped:
> They look like hacks to hide implementation specifics (wrote InputArray/OutputArray, but really can't handle set of *Array storage types) and they have no future (because with this approach you would use 2**n super set of these modifiers or much worse `n!` due ordering issues)

Thank you for clarifying this, I have been completely confused because in most (probably all) cases in the cuda namespace arguments of type `InputArray`/`OutputArray` can only handle a single array type either `Mat` or `GpuMat` not a set of *Array storage types.

Therefore would it be correct to say the existing `nextFrame` function signature is wrong and that it should be

`CV_WRAP virtual bool nextFrame(CV_OUT GpuMat& frame, Stream &stream = Stream::Null()) = 0;`

If so should all cuda functions have `Input/OuputArray` types removed?

PR #15961, opened 2019-11-20
Fix type error in cudacodec python interfaces

nullptr-leo:
The Mat or vector<Mat> in functions under cv.cudacodec. should also be translated to GpuMat. e.g. cv.cudacodec.VideoReader.

### This pullrequest changes

When processing the python interface binding, translate Mat or vector<Mat> in interfaces under both cv.cuda. and cv.cudacodec. to GpuMat. Otherwise the TypeError excepton will be thrown when calling these functions.


alalek:
@nullptr-leo Please try to help here: #15957

PR #15966, opened 2019-11-21
Add checks for empty operands in Matrix expressions that

### LINKED ISSUE #15760: division with MatExpr
- OpenCV => 4.1.2-dev
- Operating System / Platform => win7 32bit
- Compiler => mingw64 8.1

assuming `X` a valid Mat and `Y` an empty/invalid one, the expression `X / Y` does not throw an exception, but calculates `1 / X` instead.

because here (e.b.data is invalid in our case):
https://github.com/opencv/opencv/blob/471b40040a304e92bbe4a3a53e81b56dd00b2c1a/modules/core/src/matrix_expressions.cpp#L1340-L1343

why does it have `e.a` and `e.alpha` in reverse ? 
shouldn't it have thrown `Sizes of input arguments do not match` , like it does in the `Y / X` case ?

##### Steps to reproduce
    Mat x(3,3,CV_32FC3,Scalar(3,2,1));
    Mat y; // invalid
    Mat r = x / y;
    cout << r << endl;

[0.33333334, 0.5, 1, 0.33333334, 0.5, 1, 0.33333334, 0.5, 1;
 0.33333334, 0.5, 1, 0.33333334, 0.5, 1, 0.33333334, 0.5, 1;
 0.33333334, 0.5, 1, 0.33333334, 0.5, 1, 0.33333334, 0.5, 1]

saskatchewancatch:
don't check properly.

More checks and tests coming soon.

resolves #15760

### DIFF
--- a/modules/core/include/opencv2/core/mat.hpp
+++ b/modules/core/include/opencv2/core/mat.hpp
@@ -3588,9 +3588,6 @@ public:
     Scalar s;
 };
 
-void checkOperandsExist(const Mat& a);
-void checkOperandsExist(const Mat& a, const Mat& b);
-
 //! @} core_basic
 
 //! @relates cv::MatExpr
--- a/modules/core/include/opencv2/core/mat.hpp
+++ b/modules/core/include/opencv2/core/mat.hpp
@@ -3588,6 +3588,9 @@ public:
     Scalar s;
 };
 
+void checkOperandsExist(const Mat& a);
+void checkOperandsExist(const Mat& a, Mat& b);
+
 //! @} core_basic
 
 //! @relates cv::MatExpr
--- a/modules/core/include/opencv2/core/mat.hpp
+++ b/modules/core/include/opencv2/core/mat.hpp
@@ -3589,7 +3589,7 @@ public:
 };
 
 void checkOperandsExist(const Mat& a);
-void checkOperandsExist(const Mat& a, Mat& b);
+void checkOperandsExist(const Mat& a, const Mat& b);
 
 //! @} core_basic
 
--- a/modules/calib3d/test/test_undistort_badarg.cpp
+++ b/modules/calib3d/test/test_undistort_badarg.cpp
@@ -270,7 +270,7 @@ void CV_UndistortPointsBadArgTest::run(int)
     cvReleaseMat(&temp);
 
     src_points = cv::Mat();
-    errcount += run_test_case( CV_StsAssert, "Input data matrix is not continuous" );
+    errcount += run_test_case( CV_StsBadArg, "Input data matrix is not continuous" );
     src_points = cv::cvarrToMat(&_src_points_orig);
     cvReleaseMat(&temp);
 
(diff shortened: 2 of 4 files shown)


saskatchewancatch:
I'll take care of the missing declarations.
[`Mat.hpp:L3592`](https://github.com/opencv/opencv/blob/3.4/modules/core/include/opencv2/core/mat.hpp#L3592) would be the place to declare `checkOperandsExist` if I am not mistaken.

saskatchewancatch:
Hmmm now why would those fail .... did we just find some places where MatExpr operations are being done on empty operands? Will take a look.

saskatchewancatch:
Test `Calib3d_UndistortPoints.badarg` is a Bad Args style test, and it is failing because of one of the sub-tests inside of it.

[This subtest in test_undistort_badarg.cpp](https://github.com/opencv/opencv/blob/3.4/modules/calib3d/test/test_undistort_badarg.cpp#L272) essentially checks that the input src matrix to `cv::undistortPoints` is not empty.

I see one of two things can be done:
1. That one sub-test may no longer be needed, in which case it can be removed
OR
2. In [cv::undistort undistort_dispatch.cpp](https://github.com/opencv/opencv/blob/3.4/modules/imgproc/src/undistort.dispatch.cpp#L448), adjust behaviour when src matrix is empty:
2.1 Return immediately because there is nothing to undistort??
2.2 CV_Error from here.

@sovrasov @alalek Any thoughts on this?

mshabunin:
Looks like the problem with the test is that it expects `CV_StsAssert`, but you throw `CV_StsBadArg`.

saskatchewancatch:
Indeed, it's expecting the wrong one exception. I was debugging the tests in that class to see if it could be improved. Seems fine as it is. Thank you.

saskatchewancatch:
@alalek The issue that this PR resolves (#15760) is for the 3.4.9 milestone. Maybe this PR could also be targeted for 3.4.9?

modules/core/include/opencv2/core/mat.hpp
```diff
@@ -3588,6 +3588,9 @@ class CV_EXPORTS MatExpr
     Scalar s;
 };
 
+void checkOperandsExist(const Mat& a);
+void checkOperandsExist(const Mat& a, const Mat& b);
+
```
alalek: These functions should not be a part of public OpenCV API.

Moreover, they are "local", so you should use "static" before definition instead.

saskatchewancatch: Thanks @alalek. I forgot about that. 

I suppose then this is best: Remove declarations from `mat.hpp` **and** add `static` qualifier to _checkOperandsExist_ in `matrix_expressions.cpp`. So _checkOperandsExist_ will _not_ be exposed outside `matrix_expressions.cpp`.

alalek: Right!

saskatchewancatch:
imgproc_perf test failing on Mac build, but the build farm's imgproc_perf stdio for that platform doesn't show which test failed or why.

alalek:
These failures are not related to this patch. See #16137

PR #15971, opened 2019-11-22
core(kmeans): bailout if can't select cluster center

alalek:
reported here: https://github.com/opencv/opencv_contrib/issues/2351


alalek:
:+1:

PR #15973, opened 2019-11-22
Fix infinite loop when trying to change state of the busy camera

### LINKED ISSUE #15782: cv2.VideoCapture(device) blocks indefinitely if device is in use

- OpenCV => 4.1.2
- Operating System / Platform => Arch Linux
- Compiler => GCC 9.2.0 

##### Detailed description

When creating a `cv2.VideoCapture` object, if the device is in use then the call will block until the device is released. This doesn't seem to the intended behavior, due to the existent of the `isOpened` method and multiple answers to questions which seem to indicate this was not always the behavior: [1](https://answers.opencv.org/question/16166/how-to-check-if-webcam-is-busy/) [2](https://answers.opencv.org/question/199259/how-to-check-if-camera-is-opened-by-any-application/) [3](https://stackoverflow.com/a/48049952) [4](https://stackoverflow.com/a/44786906) [5](https://stackoverflow.com/a/10472231) 

##### Steps to reproduce

```
Python 3.7.4 (default, Oct  4 2019, 06:57:26)
[GCC 9.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import cv2
>>> print(cv2.__version__)
4.1.2
>>> x = cv2.VideoCapture(0)
[ WARN:0] global /build/opencv/src/opencv-4.1.2/modules/videoio/src/cap_gstreamer.cpp (933) open OpenCV | GStreamer warning: Cannot query video position: status=0, value=-1, duration=-1
>>> y = cv2.VideoCapture(0)
[ WARN:0] global /build/opencv/src/opencv-4.1.2/modules/videoio/src/cap_gstreamer.cpp (1756) handleMessage OpenCV | GStreamer warning: Embedded video playback halted; module v4l2src1 reported: Device '/dev/video0' is busy
[ WARN:0] global /build/opencv/src/opencv-4.1.2/modules/videoio/src/cap_gstreamer.cpp (886) open OpenCV | GStreamer warning: unable to start pipeline
[ WARN:0] global /build/opencv/src/opencv-4.1.2/modules/videoio/src/cap_gstreamer.cpp (480) isPipelinePlaying OpenCV | GStreamer warning: GStreamer: pipeline have not been created

(blocks until original camera released)
```

VadimLevin:
According to the [paragraph](https://linuxtv.org/downloads/v4l-dvb-apis-new/uapi/v4l/open.html#multiple-opens) about multiple opens of the video device, filehandle becomes the owner of the device only after buffers allocations. Simple call to `open` does not affect theirs state, so it falls down to capture initialization, which will stuck on palette setup.

resolves #15782 

### This pullrequest changes
Add finite number of attempts in `tryIoctl` functions (10 by default). 
First function that tries to change state of the camera is `VIDIOC_S_FMT` and it fails with `errno EBUSY`.

### DIFF
--- a/modules/videoio/src/cap_v4l.cpp
+++ b/modules/videoio/src/cap_v4l.cpp
@@ -313,6 +313,7 @@ struct CvCaptureCAM_V4L CV_FINAL : public CvCapture
     bool convert_rgb;
     bool frame_allocated;
     bool returnFrame;
+
     // To select a video input set cv::CAP_PROP_CHANNEL to channel number.
     // If the new channel number is than 0, then a video input will not change
     int channelNumber;
@@ -353,7 +354,7 @@ struct CvCaptureCAM_V4L CV_FINAL : public CvCapture
     bool initCapture();
     bool streaming(bool startStream);
     bool setFps(int value);
-    bool tryIoctl(unsigned long ioctlCode, void *parameter) const;
+    bool tryIoctl(unsigned long ioctlCode, void *parameter, int attempts = 10) const;
     bool controlInfo(int property_id, __u32 &v4l2id, cv::Range &range) const;
     bool icvControl(__u32 v4l2id, int &value, bool isSet) const;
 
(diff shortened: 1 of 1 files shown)


modules/videoio/src/cap_v4l.cpp
```diff
@@ -406,10 +407,10 @@ bool CvCaptureCAM_V4L::try_palette_v4l2()
     form.fmt.pix.field       = V4L2_FIELD_ANY;
     form.fmt.pix.width       = width;
     form.fmt.pix.height      = height;
-
-    if (!tryIoctl(VIDIOC_S_FMT, &form))
+    if (!tryIoctl(VIDIOC_S_FMT, &form, 2))
```
alalek: What is about adding parameter to handle "EBUSY" case?

VadimLevin: Ok, I will add it as additional argument, before attempts count.
I would like to keep limited attempts to ensure that loop is finite.

PR #15978, opened 2019-11-23
videoio(v4l2): use logging, update handling of EBUSY, device closing

alalek:
- DEBUG logging compilation is enabled for all videoio backends
- eliminate output through perror(), stderr

### DIFF
--- a/modules/videoio/src/cap_v4l.cpp
+++ b/modules/videoio/src/cap_v4l.cpp
@@ -278,6 +278,32 @@ make & enjoy!
 
 namespace cv {
 
+static const char* decode_ioctl_code(unsigned long ioctlCode)
+{
+    switch (ioctlCode)
+    {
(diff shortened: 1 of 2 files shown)


modules/videoio/src/cap_v4l.cpp
```diff
@@ -863,37 +940,64 @@ bool CvCaptureCAM_V4L::read_frame_v4l2()
...
+            CV_LOG_INFO(NULL, "VIDEOIO(V4L2:" << deviceName << "): ioctl returns with errno=EBUSY");
             return false;
         }
-        if ((attempts > 0) && (--attempts == 0)) {
+        if (!(isBusy || errno == EAGAIN))
             return false;
-        }
 
-        if (!(isBusy || errno == EAGAIN))
+        if (--attempts == 0) {
```
VadimLevin: Maybe log about exceeding of attempts count should be added just like for busy device case?

PR #15988, opened 2019-11-25
Test create custom layer in python

l-bat:
```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/misc/python/test/test_dnn.py
+++ b/modules/dnn/misc/python/test/test_dnn.py
@@ -286,7 +286,7 @@ class dnn_test(NewOpenCVTests):
             src_shape = [1, 2, 5, 5]
             dst_shape = [1, 2, 3, 3]
             inp = np.arange(0, np.prod(src_shape), dtype=np.float32).reshape(src_shape)
-            roi = np.empty(dst_shape)
+            roi = np.empty(dst_shape, dtype=np.float32)
             net.setInput(inp, "input")
             net.setInput(roi, "roi")
             out = net.forward()
--- a/modules/dnn/misc/python/test/test_dnn.py
+++ b/modules/dnn/misc/python/test/test_dnn.py
@@ -280,13 +280,13 @@ class dnn_test(NewOpenCVTests):
         net = cv.dnn.readNetFromCaffe(str.encode(proto))
         src_shape = [1, 2, 5, 5]
         dst_shape = [1, 2, 3, 3]
-        inp = np.arange(0, np.prod(src_shape)).reshape(src_shape)
+        inp = np.arange(0, np.prod(src_shape), dtype=np.float32).reshape(src_shape)
         roi = np.empty(dst_shape)
         net.setInput(inp, "input")
         net.setInput(roi, "roi")
         out = net.forward()
         ref = inp[:, :, 1:4, 1:4]
-        assert(np.allclose(out, ref))
+        normAssert(self, out, ref)
 
 if __name__ == '__main__':
     NewOpenCVTests.bootstrap()
--- a/modules/dnn/misc/python/test/test_dnn.py
+++ b/modules/dnn/misc/python/test/test_dnn.py
@@ -248,32 +248,33 @@ class dnn_test(NewOpenCVTests):
                 return [inputs[0][:,:,self.ystart:self.yend,self.xstart:self.xend]]
 
         cv.dnn_registerLayer('CropCaffe', CropLayer)
-        proto = 'name: "TestCrop"\n' \
-        'input: "input"\n' \
-        'input_shape\n' \
-        '{\n' \
-        '    dim: 1\n' \
-        '    dim: 2\n' \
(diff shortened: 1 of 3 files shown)


modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -226,6 +226,67 @@ def test_async(self):
...
+        net = cv.dnn.readNetFromCaffe(str.encode(proto))
+        src_shape = [1, 2, 5, 5]
+        dst_shape = [1, 2, 3, 3]
+        inp = np.arange(0, np.prod(src_shape)).reshape(src_shape)
+        roi = np.empty(dst_shape)
+        net.setInput(inp, "input")
+        net.setInput(roi, "roi")
+        out = net.forward()
+        ref = inp[:, :, 1:4, 1:4]
+        assert(np.allclose(out, ref))
```
dkurt: Please use `normAssert`

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -226,6 +226,67 @@ def test_async(self):
...
+              bottom: "input"
+              bottom: "roi"
+              top: "Crop"
+            }'''
+
+        net = cv.dnn.readNetFromCaffe(str.encode(proto))
+        src_shape = [1, 2, 5, 5]
+        dst_shape = [1, 2, 3, 3]
+        inp = np.arange(0, np.prod(src_shape)).reshape(src_shape)
+        roi = np.empty(dst_shape)
```
dkurt: Please check that input data is `float32`.

modules/python/src2/cv2.cpp
```diff
@@ -392,6 +392,7 @@ static bool pyopencv_to(PyObject* o, Mat& m, const ArgInfo& info)
             oarr = (PyArrayObject*) o;
         }
         else {
+            CV_Assert(PyArray_ISCONTIGUOUS(oarr));
             oarr = PyArray_GETCONTIGUOUS(oarr);
```
dkurt: According to https://docs.scipy.org/doc/numpy-1.13.0/reference/c-api.array.html#c.PyArray_GETCONTIGUOUS it looks like `PyArray_GETCONTIGUOUS` should manage both cases.

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -226,6 +226,67 @@ def test_async(self):
...
+                self.xstart = (inputShape[3] - targetShape[3]) // 2
+                self.yend = self.ystart + height
+                self.xend = self.xstart + width
+                return [[batchSize, numChannels, height, width]]
+            def forward(self, inputs):
+                return [inputs[0][:,:,self.ystart:self.yend,self.xstart:self.xend]]
+
+        cv.dnn_registerLayer('Crop', CropLayer)
+        proto = '''
+            name: "TestCrop"
```
alalek: Please remove unnecessary indentation:
```
            proto = '''\
name: "TestCrop"
...
}'''
```

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -226,6 +226,67 @@ def test_async(self):
...
+                height, width = targetShape[2], targetShape[3]
+                self.ystart = (inputShape[2] - targetShape[2]) // 2
+                self.xstart = (inputShape[3] - targetShape[3]) // 2
+                self.yend = self.ystart + height
+                self.xend = self.xstart + width
+                return [[batchSize, numChannels, height, width]]
+            def forward(self, inputs):
+                return [inputs[0][:,:,self.ystart:self.yend,self.xstart:self.xend]]
+
+        cv.dnn_registerLayer('Crop', CropLayer)
```
alalek: What is happen in override attempt of existed layer?

dkurt: There is a stack of registered layers. So if there was a registered layer, new version will be used until dnn_unregister call. BTW, @l-bat, I think we should unregister it after the test similar to this sample - https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_layers.cpp#L1303

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -226,6 +226,73 @@ def test_async(self):
...
+        '}\n' \
+        'layer {\n' \
+        '  name: "Crop"\n' \
+        '  type: "CropCaffe"\n' \
+        '  bottom: "input"\n' \
+        '  bottom: "roi"\n' \
+        '  top: "Crop"\n' \
+        '} '
+
+        net = cv.dnn.readNetFromCaffe(str.encode(proto))
```
alalek: There are several functions with the same name. Code should call the right overload.
Try to wrap input into `bytearray(proto.encode())` (just string is a file name).

P.S. Revert multiline string back.

PR #15998, opened 2019-11-26
ts: count skipped tests via SkipTestException

alalek:
- apply tag 'skip_other'
- apply tag 'skip_bigdata' (legacy tests, new tests should use test tags with memory limits)


alalek:
Need to fix [this case](http://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/23772/steps/test_core/logs/stdio): 5 skipped tests -> 4 is reported only

PR #16008, opened 2019-11-27
Python module: replace config.py files by config.ini

tpetazzoni:
The OpenCV Python module currently installs and reads at runtime a config.py file and a config-X.Y.py file to retrieve some bits of configuration. This is done using the execfile() Python mechanism, which unfortunately only works with .py files, and not with .pyc files.

However, on some embedded systems, only .pyc files are installed: they are already byte-compiled, which optimizes the startup time, and installating both .pyc and .py files doubles the storage space, which is a limited resource on embedded systems.

In addition, loading .py file requires different code between Python 2 and Python 3.

To solve these issues, this pull request replaces the .py files by .ini files, which can trivially be parsed by the standard ConfigParser Python module.


modules/python/package/cv2/__init__.py
```diff
@@ -3,6 +3,7 @@
 '''
 import os
 import sys
+import configparser
```
mshabunin: In Python2 it is named `ConfigParser`.

tpetazzoni: Ah, yes, thank you, indeed.

VadimLevin:
Thanks for you contribution! 
@tpetazzoni Do you have any chance to proceed work on this PR?

asmorkalov:
@tpetazzoni any progress on the patch?

tpetazzoni:
@asmorkalov No sorry, not yet, I'll try to get to it.

alalek:
> on some embedded systems, only .pyc files are installed

Consider disabling OpenCV Python bindings loader in your configuration: `cmake -DOPENCV_SKIP_PYTHON_LOADER=ON ...`

Perhaps .ini files will not work for us. In the future we want to detect Python distributions in runtime, like Anaconda Python, Intel Python distribution, etc. One package + auto detection/configuration through .py files config.

Instead of .ini approach you may take a look in direction of changing of importing approach (use some "import" code instead of "execfile")

asmorkalov:
OpenCV team decided to close the PR as expected behavior is reachable with `-DOPENCV_SKIP_PYTHON_LOADER=ON` option in CMake. See alalek's comment above.

tpetazzoni:
Well, OPENCV_SKIP_PYTHON_LOADER may not be an option: some people do want the Python bindings of OpenCV available on their embedded system.

alalek:
`OPENCV_SKIP_PYTHON_LOADER` doesn't disable OpenCV Python bindings itself.

PR #16010, opened 2019-11-28
fix and enable tests for `DNN_TARGET_CUDA_FP16`

YashasSamaga:
### This pullrequest changes

- fix and enable tests for `DNN_TARGET_CUDA_FP16`

**Pending:**
- investigate FasterRCNN failures
- investigate deconvolution failures

Merge after #16045 

<cut/>

```
force_builders=Custom,linux,docs
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/dnn/test/test_backends.cpp
+++ b/modules/dnn/test/test_backends.cpp
@@ -203,7 +203,7 @@ TEST_P(DNNTestNetwork, MobileNet_SSD_Caffe_Different_Width_Height)
 #endif
     Mat sample = imread(findDataFile("dnn/street.png"));
     Mat inp = blobFromImage(sample, 1.0f / 127.5, Size(300, 560), Scalar(127.5, 127.5, 127.5), false);
-    float scoreDiff  = 0.0, iouDiff = 0.0;
+    float scoreDiff = 0.0, iouDiff = 0.0;
     if (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD)
     {
         scoreDiff = 0.029;
@@ -228,7 +228,7 @@ TEST_P(DNNTestNetwork, MobileNet_SSD_v1_TensorFlow)
     Mat sample = imread(findDataFile("dnn/street.png"));
     Mat inp = blobFromImage(sample, 1.0f, Size(300, 300), Scalar(), false);
     float detectionConfThresh = (target == DNN_TARGET_MYRIAD) ? 0.216 : 0.2;
-    float scoreDiff  = 0.0, iouDiff = 0.0;
+    float scoreDiff = 0.0, iouDiff = 0.0;
     if (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD)
     {
         scoreDiff = 0.095;
@@ -260,7 +260,7 @@ TEST_P(DNNTestNetwork, MobileNet_SSD_v1_TensorFlow_Different_Width_Height)
 
     Mat sample = imread(findDataFile("dnn/street.png"));
     Mat inp = blobFromImage(sample, 1.0f, Size(300, 560), Scalar(), false);
-    float scoreDiff  = 0.0, iouDiff = 0.0;
+    float scoreDiff = 0.0, iouDiff = 0.0;
     if (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD)
     {
         scoreDiff = 0.012;
@@ -284,7 +284,7 @@ TEST_P(DNNTestNetwork, MobileNet_SSD_v2_TensorFlow)
 
     Mat sample = imread(findDataFile("dnn/street.png"));
     Mat inp = blobFromImage(sample, 1.0f, Size(300, 300), Scalar(), false);
-    float scoreDiff  = 2e-5, iouDiff = 0.0;
+    float scoreDiff = 2e-5, iouDiff = 0.0;
     if (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD)
     {
         scoreDiff = 0.013;
@@ -308,7 +308,7 @@ TEST_P(DNNTestNetwork, SSD_VGG16)
         applyTestTag(CV_TEST_TAG_DNN_SKIP_HALIDE);  // TODO HALIDE_CPU
     Mat sample = imread(findDataFile("dnn/street.png"));
     Mat inp = blobFromImage(sample, 1.0f, Size(300, 300), Scalar(), false);
-    float scoreDiff  = 0.0, iouDiff = 0.0;
+    float scoreDiff = 0.0, iouDiff = 0.0;
     if (target == DNN_TARGET_OPENCL_FP16)
     {
         scoreDiff = 0.0325;
@@ -316,7 +316,7 @@ TEST_P(DNNTestNetwork, SSD_VGG16)
     else if (target == DNN_TARGET_MYRIAD)
     {
         scoreDiff = 0.0325;
-        iouDiff = 0.062;
+        iouDiff = 0.032;
     }
     else if (target == DNN_TARGET_CUDA_FP16)
     {
@@ -435,7 +435,7 @@ TEST_P(DNNTestNetwork, Inception_v2_SSD_TensorFlow)
         applyTestTag(CV_TEST_TAG_DNN_SKIP_HALIDE);
     Mat sample = imread(findDataFile("dnn/street.png"));
     Mat inp = blobFromImage(sample, 1.0f, Size(300, 300), Scalar(), false);
-    float scoreDiff  = 0.0, iouDiff = 0.0;
+    float scoreDiff = 0.0, iouDiff = 0.0;
     if (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD)
     {
         scoreDiff = 0.015;
--- a/modules/dnn/test/test_backends.cpp
+++ b/modules/dnn/test/test_backends.cpp
@@ -168,6 +168,8 @@ TEST_P(DNNTestNetwork, ENet)
         applyTestTag(CV_TEST_TAG_DNN_SKIP_IE_NGRAPH);
     if (backend == DNN_BACKEND_OPENCV && target == DNN_TARGET_OPENCL_FP16)
         applyTestTag(CV_TEST_TAG_DNN_SKIP_OPENCL_FP16);
+    if (backend == DNN_BACKEND_CUDA && target == DNN_TARGET_CUDA_FP16)
+        applyTestTag(CV_TEST_TAG_DNN_SKIP_CUDA_FP16);
     processNet("dnn/Enet-model-best.net", "", Size(512, 512), "l367_Deconvolution",
                target == DNN_TARGET_OPENCL ? "dnn/halide_scheduler_opencl_enet.yml" :
                                              "dnn/halide_scheduler_enet.yml",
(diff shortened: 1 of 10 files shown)


modules/dnn/test/test_caffe_importer.cpp
```diff
@@ -667,8 +699,17 @@ TEST_P(Test_Caffe_nets, RFCN)
...
+    float scoreDiff = default_l1, iouDiff = default_lInf;
+    if (backend == DNN_BACKEND_OPENCV && target == DNN_TARGET_OPENCL_FP16)
+    {
+        scoreDiff = 4e-3;
+        iouDiff = 8e-2;
+    }
+    if (target == DNN_TARGET_CUDA_FP16)
+    {
+        scoreDiff = 0.04;
+        iouDiff = 0.11;
```
dkurt: Please add a comment about the range of reference output values. It can help t understand if this threshold is acceptable or not.

modules/dnn/test/test_layers.cpp
```diff
@@ -141,7 +141,13 @@ TEST_P(Test_Caffe_layers, Convolution)
 
 TEST_P(Test_Caffe_layers, DeConvolution)
 {
-    testLayerUsingCaffeModels("layer_deconvolution", true, false);
+    double l1 = 0.0, lInf = 0.0;
+    if(target == DNN_TARGET_CUDA_FP16)
+    {
+        l1 = 4;
+        lInf = 60;
```
dkurt: Please also specify reference values range.

modules/dnn/test/test_layers.cpp
```diff
@@ -356,8 +362,14 @@ TEST_P(Test_Caffe_layers, Conv_Elu)
...
     net.setPreferableTarget(target);
     Mat out = net.forward();
-
-    normAssert(ref, out, "", default_l1, default_lInf);
+    
+    double l1 = default_l1, lInf = default_lInf;
+    if (target == DNN_TARGET_CUDA_FP16)
+    {
+        l1 = 2;
+        lInf = 3;
```
dkurt: Also seems too big. Please specify the range of reference values.

modules/dnn/test/test_model.cpp
```diff
@@ -166,11 +175,20 @@ TEST_P(Test_Model, DetectionOutput)
...
+    {
+        if (backend == DNN_BACKEND_OPENCV)
+            scoreDiff = 4e-3;
+        iouDiff = 1.8e-1;
+    }
+    else if (target == DNN_TARGET_CUDA_FP16)
+    {
+        scoreDiff = 4e-3;
+        iouDiff = 0.18;
+    }
```
dkurt: Can we merge these two conditions? Thresholds are the same.

modules/dnn/test/test_tf_importer.cpp
```diff
@@ -143,7 +143,9 @@ TEST_P(Test_TensorFlow_layers, Convolution3D)
 
 TEST_P(Test_TensorFlow_layers, padding)
 {
-    runTensorFlowNet("padding_valid");
+    runTensorFlowNet("padding_valid", false,
+        (target == DNN_TARGET_CUDA_FP16) ? 2 : 0.0,
+        (target == DNN_TARGET_CUDA_FP16) ? 2.1 : 0.0);
```
dkurt: Looks like a bug in the implementation. Too big difference for zero paddings.

modules/dnn/test/test_backends.cpp
```diff
@@ -199,8 +201,17 @@ TEST_P(DNNTestNetwork, MobileNet_SSD_Caffe_Different_Width_Height)
 #endif
     Mat sample = imread(findDataFile("dnn/street.png"));
     Mat inp = blobFromImage(sample, 1.0f / 127.5, Size(300, 560), Scalar(127.5, 127.5, 127.5), false);
-    float diffScores  = (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD) ? 0.029 : 0.0;
-    float diffSquares = (target == DNN_TARGET_OPENCL_FP16 || target == DNN_TARGET_MYRIAD) ? 0.09  : 0.0;
+    float diffScores  = 0.0, diffSquares = 0.0;
```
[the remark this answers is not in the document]
YashasSamaga: In detection networks, the `l1` parameter is the l1 threshold for scores and the `lInf` is the l1 threshold for the box IOU error. The names were already being used.

Hence, `lInf` doesn't make sense for the way it's used.

modules/dnn/test/test_tf_importer.cpp
YashasSamaga: The test network for `padding_valid` uses ELU activation. The activation is the source of the error (there is a test on Conv_Elu which also has high errors).

```
[ RUN      ] Test_TensorFlow_layers.padding/1, where GetParam() = CUDA/CUDA_FP16
padding valid:
[CUDA] Convolution conv2d_3/convolution
[CUDA] ELU conv2d_3/Elu
Ref. Range: -0.980401 0.123652
padding spatial:
[CUDA] Convolution conv2d/convolution
Ref. Range: -2.53209 1.58402
mirror_pad:
[CUDA] Padding MirrorPad
Ref. Range: -3.00851 3.45245
keras_pad_concat:
[CUDA] Convolution keras_pad_concat_conv/Conv2D
[CUDA] Const keras_pad_concat/Variable
[CUDA] Concat keras_pad_concat/concatenate/concat
Ref. Range: -1.31247 1.212
[       OK ] Test_TensorFlow_layers.padding/1 (32 ms)
```

modules/dnn/test/test_layers.cpp
YashasSamaga:
```
[ RUN      ] Test_Caffe_layers.Conv_Elu/1, where GetParam() = CUDA/CUDA_FP16
[CUDA] Convolution elu/Conv2D
[CUDA] ELU elu
Ref. Range: -0.255777 0.0283771
[       OK ] Test_Caffe_layers.Conv_Elu/1 (26 ms)
```

I  think it's just ELU being very inaccurate. The FP32 target uses a numerically stable `expm1` for the activation while the FP16 uses `exp(x) - 1`. I can try to write a numerically stable version for the half precision kernel but that will be considerably slower and would suffer from wrap divergence.

YashasSamaga: Bug fixed in #16045.

New thresholds:
```
    double l1 = 1e-5, lInf = 1e-4;
    if (target == DNN_TARGET_CUDA_FP16)
    {
        l1 = 0.0002;
        lInf = 0.0005;
    }
```

modules/dnn/test/test_tf_importer.cpp
YashasSamaga: Bug fixed in #16045. Custom thresholds removed.

modules/dnn/test/test_caffe_importer.cpp
YashasSamaga: The ranges for both scores and IOU is `[0, 1]`.

The thresholds have been updated:
```
scoreDiff = 0.0034;
iouDiff = 0.11;
```

The thresholds used by OCL_FP16 and MYRAID:
```
scoreDiff = 4e-3;
iouDiff = 8e-2;
```

The iouDiff threshold of `0.08` is close to `0.11`. I think there is no bug with CUDA_FP16 and the thresholds are correct.

alalek:
Please rebase commits on latest master

modules/dnn/test/test_layers.cpp
YashasSamaga: Is it ok if I leave deconvolution tests failing? There is a problem which has been there ever since the original CUDA backend had been merged.

dkurt: Why not skip them?

YashasSamaga: I thought broken stuff should be failing tests instead of being hidden. I'll skip then.

dkurt: So, we just trying to follow green tests strategy. To ignore skip tag there is a flags (maybe `--test_tag_enable` if I'm correct).

modules/dnn/test/test_torch_importer.cpp
```diff
@@ -112,16 +112,31 @@ class Test_Torch_layers : public DNNTestLayer
...
 {
     if (backend == DNN_BACKEND_OPENCV && target == DNN_TARGET_OPENCL_FP16)
         applyTestTag(CV_TEST_TAG_DNN_SKIP_OPENCL_FP16);
-    runTorchNet("net_pool_max", "", true);
+    double l1 = 0.0, lInf = 0.0;
+    if (target == DNN_TARGET_CUDA_FP16)
+    {
+        l1 = 0.2;
+        lInf = 121;
+    }
```
dkurt: It looks like a bug. Let's skip this test instead.

tompollok:
@YashasSamaga is there a way to check at runtime if FP16 is supported by the used nvidia graphicscard?

YashasSamaga:
@tompollok 
1. https://github.com/opencv/opencv/pull/16218#issuecomment-569517039
2. https://github.com/opencv/opencv/blob/dae1fc8889c13cb67f90b0e34b07d4d4b785ff0f/modules/dnn/src/dnn.cpp#L158-L173

tompollok:
@YashasSamaga  thank you! Wouldn't it make sense to expose a function, such that there is not need to copy pase it for checking in runtime or does DNN_TARGET_CUDA_FP16 automatically fallback on DNN_TARGET_CUDA if FP16 is not supported by the device?

cuixing158:
Why am I set to DNN_TARGET_CUDA_FP16, the inferencing speed is slower than DNN_TARGET_CUDA?
os: win10
opencv: opencv4.3.0, complie from source
gpu: gtx1050ti-4G
complier:VS2015

YashasSamaga:
@cuixing158 GTX 1050 Ti doesn't support half precision at 2x speed. FP16 is sixty four times slower than FP32.

You can check if a GPU supports high FP16 performanc here: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#arithmetic-instructions__throughput-native-arithmetic-instructions

cuixing158:
@YashasSamaga Thanks very much!

yiwenwan2008:
> @YashasSamaga Thanks very much!

are you able to use FP16 for some other GPUs to get the 2x speed (supporting half precision)?

yiwenwan2008:
> @cuixing158 GTX 1050 Ti doesn't support half precision at 2x speed. FP16 is sixty four times slower than FP32.
> 
> You can check if a GPU supports high FP16 performance here: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#arithmetic-instructions__throughput-native-arithmetic-instruction
SPECIFICATIONS
GPU Architecture NVIDIA Turing
NVIDIA Turing Tensor
Cores 320
NVIDIA CUDA® Cores 2,560
Single-Precision 8.1 TFLOPS
**Mixed-Precision
(FP16/FP32) 65 TFLOPS**
INT8 130 TOPS
INT4 260 TOPS
this is specs for Nvidia T4, why precision FP16/FP32 is mixed together? can we tell from the specs whether it support half precision to give 2x speed up?

YashasSamaga:
> Above is specs for Nvidia T4, why precision FP16/FP32 is mixed together? can we tell from the specs whether it support half precision to give 2x speed up?

Yes, T4 supports FP16 and gives a good performance. The speedup need not necessarily be 2x. It can be 1.5x or even 2.5x. I think mixed precision in your context refers to tensor cores which can multiple numbers in FP16 but accumulate with FP32.

yiwenwan2008:
Great and Thank you so much for your reply! Also what I need to do to turn on FP16?
By simply using DNN_TARGET_CUDA_FP16 option? Is there anything else I need to do? 
Thank you again for your help!

PR #16027, opened 2019-12-01
Adding build for ARM64 Windows10

shibayan:
### This pullrequest changes

- Adding ARM64 Windows 10 build support
  - Added `ARM64` archtecture
  - Use `_M_ARM64` preprocessor
  - Added default library for highgui (for ARM64 default library is less than x86/x64)
- I confirmed the build and execute with Visual Studio 2019 and Surface Pro X.

```
force_builders=Custom Win
build_image:Custom Win=msvs2019-arm64
buildworker:Custom Win=windows-1
build_contrib:Custom Win=OFF
```


alalek:
What arguments should be used for CMake run / flags enabled? (please share some best known configuration)

shibayan:
I used this CMake option to build and execute.

```
cmake -G "Visual Studio 16 2019" -A ARM64 -DCMAKE_SYSTEM_NAME=Windows -DCMAKE_SYSTEM_VERSION=10.0 \
      -DCMAKE_SYSTEM_PROCESSOR=ARM64 -DWITH_OPENCL=OFF -DWITH_FFMPEG=OFF -DWITH_CUDA=OFF \
      -DBUILD_EXAMPLES=ON -DBUILD_TESTS=ON C:\opencv

```

alalek:
Thank you for update!

What are arguments used for `vcvarsall.cmd` call?

Did you see these lines from CMake?
```
-- Performing Test HAVE_CPU_NEON_SUPPORT (check file: cmake/checks/cpu_neon.cpp)
-- Performing Test HAVE_CPU_NEON_SUPPORT - Failed
-- NEON is not supported by C++ compiler
```

If so, please try to fix check (ARM code without NEON optimizations is slow).

shibayan:
Thank you for review.

>What are arguments used for vcvarsall.cmd call?

Use `x64_arm64` argument

I'm trying to enable NEON, but the build doesn't pass.
Even I try to build with ARM (32bit) for UWP, an error will occur in the same part, so it seems to be another problem.

alalek:
Another problem here is that CMake is not running in cross-compilation mode (`CMAKE_CROSSCOMPILING` is no set).
Perhaps we need to prepare Windows ARM64 CMake toolchain file.

---

I tried to fix check_neon.cpp but got this error:

> fatal error C1189: #error:  This header is specific to ARM targets

However `/p:Platform=ARM64` parameter is used and compiler is "for ARM64".

<details>

```
Run Build Command(s):C:/Program Files (x86)/Microsoft Visual Studio/2019/BuildTools/MSBuild/Current/Bin/MSBuild.exe cmTC_928d6.vcxproj /p:Configuration=Debug /p:Platform=ARM64 /p:VisualStudioVersion=16.0 /v:m 
Microsoft (R) Build Engine version 16.0.461+g6ff56ef63c for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

  Microsoft (R) C/C++ Optimizing Compiler Version 19.20.27508.1 for ARM64
  Copyright (C) Microsoft Corporation.  All rights reserved.
  
  cpu_neon.cpp
  cl /c /Zi /W3 /WX- /diagnostics:classic /Od /Ob0 /Oy- /D WIN32 /D _WINDOWS /D "CMAKE_INTDIR=\"Debug\"" /D _ARM64_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE=1 /D _MBCS /Gm- /MDd /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /Fo"cmTC_928d6.dir\Debug\\" /Fd"cmTC_928d6.dir\Debug\vc142.pdb" /Gd /TP /analyze- /errorReport:queue E:\projects\opencv\cmake\checks\cpu_neon.cpp
  
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\VC\Tools\MSVC\14.20.27508\include\arm_neon.h(17): fatal error C1189: #error:  This header is specific to ARM targets [E:\build\opencv_arm64\CMakeFiles\CMakeTmp\cmTC_928d6.vcxproj]
```

</details>

shibayan:
Thanks you!

Attach `CMakeVars.txt` when the cmake command is executed. `CMAKE_CROSSCOMPILING` was set.
https://gist.github.com/shibayan/eec4209524d208cdfdea6a4eae45f949
(`CMAKE_CROSSCOMPILING` seems to be enabled by setting `CMAKE_SYSTEM_NAME` and `CMAKE_SYSTEM_VERSION`.)

Did you encounter a problem with a previously shared cmake command?

---

It was a build error when NEON was enabled, but I pushed the branch I worked on halfway.
https://github.com/shibayan/opencv/tree/arm64-neon-windows10

If you build after running cmake, you should get an error in `intrin_neon.hpp`.

```
C:\opencv\modules\core\include\opencv2/core/hal/intrin_neon.hpp(108,1): 
error C2084: function 'void cv::hal_baseline::_v128_unzip(const uint8x16_t &,const uint8x16_t &,uint8x16_t &,uint8x16_t &)' already has a body [C:\opencv\build-neon\modules\core\opencv_core.vcxproj]
```

alalek:
I added "custom" builder with ARM64 configuration. Please [take a look](http://pullrequest.opencv.org/#/summary/opencv).

Consider pushing changes with updates on source branch of this PR (`arm64-windows10`)

alalek:
Pushed NEON changes from branch "arm64-neon-windows10" here.

alalek:
> This header is specific to ARM targets

Resolved by upgrading MSVS 2019.

---

> function 'void cv::hal_baseline::_v128_unzip(const uint8x16_t &,const uint8x16_t &,uint8x16_t &,uint8x16_t &)' already has a body

Looks like a MSVC compiler bug which aliased "int8x16_t" and "uint8x16_t" and don't accept overloads.
MSVC versions: 19.23.28107.0 and 19.24.28314.0

This simple code:
```
void test_aliased_type(const uint8x16_t& a) { }
void test_aliased_type(const int8x16_t& a) { }
```

Emits these errors:
```
intrin_neon.hpp(71,44): error C2084: function 'void cv::hal_baseline::test_aliased_type(const uint8x16_t &)' already has a body
intrin_neon.hpp(70): message : see previous definition of 'test_aliased_type'
```

Any ideas how to workaround that? (without massive code refactoring)

shibayan:
In MSVC, `uint8x16_t` and `int8x16_t` are defined as aliases of `__n128`, whereas GCC is an alias of `__builtin_neon_uqi` and `__builtin_neon_qi`, so it seems that overloading can be defined.

I think this is a very difficult problem to solve. At least I couldn't find a workaround...

shibayan:
Bug fixes on the Visual Studio side will take time. The performance will drop because NEON cannot be used, but I would like to make the build pass for ARM / ARM64.
(I will continue to communicate with the Visual C++ team.)

PR #16015, opened 2019-11-28
Reading IR models from buffer

Quantizs:
resolves #15578

### This pullrequest changes

Added implementations to load IR models from buffer

PR #16034, opened 2019-12-02
Load IR model from buffer

### LINKED ISSUE #15578: Read ir model from buffer not implemented
- OpenCV => 4.1.1
- Operating System / Platform => Windows 64 Bit / Ubuntu 18.04 LTS
- Compiler => Visual Studio 2017 / clang6.0

We are trying to load an IR model from a memory buffer (loaded from a single binary file, not relevant) but `dnn::readNet()` for dldt from buffer is not implemented.

```cpp
 else if (framework == "dldt")
        CV_Error(Error::StsNotImplemented, "Reading Intel's Model Optimizer models from buffers");
```

My colleague is already working on submitting a pull request to implement this.

Quantizs:
resolves #15578

Added implementations to load IR models from buffer

<cut/>

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -384,13 +384,32 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         CV_WRAP Net();  //!< Default constructor.
         CV_WRAP ~Net(); //!< Destructor frees the net only if there aren't references to the net anymore.
 
-        /** @brief Create a network from Intel's Model Optimizer intermediate representation.
+        /** @brief Create a network from Intel's Model Optimizer intermediate representation (IR).
          *  @param[in] xml XML configuration file with network's topology.
          *  @param[in] bin Binary file with trained weights.
          *  Networks imported from Intel's Model Optimizer are launched in Intel's Inference Engine
          *  backend.
          */
         CV_WRAP static Net readFromModelOptimizer(const String& xml, const String& bin);
+
+        /** @brief Create a network from Intel's Model Optimizer in-memory buffers with intermediate representation (IR).
+         *  @param[in] bufferModelConfig buffer with model's configuration.
+         *  @param[in] bufferWeights buffer with model's trained weights.
+         *  @returns Net object.
+         */
+        CV_WRAP static
+        Net readFromModelOptimizer(const std::vector<uchar>& bufferModelConfig, const std::vector<uchar>& bufferWeights);
+
+        /** @brief Create a network from Intel's Model Optimizer in-memory buffers with intermediate representation (IR).
+         *  @param[in] bufferModelConfigPtr buffer pointer of model's configuration.
+         *  @param[in] bufferModelConfigSize buffer size of model's configuration.
+         *  @param[in] bufferWeightsPtr buffer pointer of model's trained weights.
+         *  @param[in] bufferWeightsSize buffer size of model's trained weights.
+         *  @returns Net object.
+         */
+        static
+        Net readFromModelOptimizer(const uchar* bufferModelConfigPtr, size_t bufferModelConfigSize,
+                                            const uchar* bufferWeightsPtr, size_t bufferWeightsSize);
 
         /** Returns true if there are no layers in the network. */
         CV_WRAP bool empty() const;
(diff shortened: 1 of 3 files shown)


modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -357,6 +357,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         virtual ~Layer();
     };
 
+    struct LayerData;
```
dkurt: This is an internal data structure.

Quantizs: https://github.com/opencv/opencv/pull/16015#discussion_r351744027
To create a seprate functon from the common parts in the two loading method (file and buffer):
- I have to pass an InferenceEngine::CNNNetReader which can not be included in the dnn.hpp, so i had to use static function
- I need to access the Impl struct of the Net class which is forward declarated in dnn.hpp and this where i can't find a clean solution.

As I see now I have these options here: 
1. Writing a getter to member impl which doesn't seems right because its a private member. 
2. The current solution: writing a getter which can get a layerData from the impl struct by id.
To do this a had to forward declarate the LayerData struct in the header. 
3. Writing setters to layerInstance and backendNodes:
        ld.layerInstance = cvLayer;
        ld.backendNodes[DNN_BACKEND_INFERENCE_ENGINE] = backendNode;
The problem here is the backendNode which is has an InfEngineBackendNet type whisch also can't be included in the dnn.hpp 

If there is an other way please help me to find it.

modules/dnn/src/dnn.cpp
```diff
@@ -4344,7 +4422,7 @@ Net readNet(const String& _framework, const std::vector<uchar>& bufferModel,
     else if (framework == "torch")
         CV_Error(Error::StsNotImplemented, "Reading Torch models from buffers");
     else if (framework == "dldt")
-        CV_Error(Error::StsNotImplemented, "Reading Intel's Model Optimizer models from buffers");
+        return readNetFromModelOptimizerBuffer(bufferConfig, bufferModel);
```
alalek: @dkurt Please verify parameters order.

dkurt: Looks fine because of origin

```cpp
CV_WRAP static Net readFromModelOptimizer(const String& xml, const String& bin);
```

modules/dnn/src/dnn.cpp
```diff
@@ -3001,9 +2995,93 @@ Net Net::readFromModelOptimizer(const String& xml, const String& bin)
...
+#ifndef HAVE_INF_ENGINE
+    CV_UNUSED(bufferModelConfigPtr); CV_UNUSED(bufferWeightsPtr);
+    CV_UNUSED(bufferModelConfigSize); CV_UNUSED(bufferModelConfigSize);
+    CV_Error(Error::StsError, "Build OpenCV with Inference Engine to enable loading models from Model Optimizer.");
+#else
+
+#if INF_ENGINE_VER_MAJOR_LE(INF_ENGINE_RELEASE_2019R3)
+    InferenceEngine::CNNNetReader reader;
+
+    InferenceEngine::CNNNetReader reader;
```
dkurt: please remove duplicated code

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -694,7 +713,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         CV_WRAP int64 getPerfProfile(CV_OUT std::vector<double>& timings);
 
     private:
-        struct Impl;
+        struct Impl; friend struct Impl;
```
dkurt: It looks like we can remove this as well.

alalek:
> readFromModelOptimizer
> readFromModelOptimizerBuffer
> readFromModelOptimizerBufferPtr

Any objections about these names?

dkurt:
@alalek, sorry. I missed this

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -384,14 +384,33 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
          */
         CV_WRAP static Net readFromModelOptimizer(const String& xml, const String& bin);
 
+        /** @brief Create a network from Intel's Model Optimizer in-memory buffers with intermediate representation (IR).
+         *  @param[in] bufferModelConfig buffer with model's configuration.
+         *  @param[in] bufferWeights buffer with model's trained weights.
+         *  @returns Net object.
+         */
+        CV_WRAP static
+        Net readFromModelOptimizerBuffer(const std::vector<uchar>& bufferModelConfig, const std::vector<uchar>& bufferWeights);
```
dkurt: All the versions should have the same name: readFromModelOptimizer and readNetFromModelOptimizer.

alalek: The same name is not good and should be avoided.
At least for bindings support.
"String" (bytes array in meaning of UTF-8 file path) and `vector<uchar>` (or other forms of bytes array) is similar things and bindings wrappers may convert between them.

`readFromModelOptimizer` should be called `readFromModelOptimizerFile` but it is too late.

---

What is about prefixes?
- "read" for files ("load" is also used, for example in objdetect module)
- "create" / "build" for buffers

dkurt: Here an example when bindings work correctly with Darknet model: https://github.com/opencv/opencv/pull/11104#issuecomment-403763020

alalek: I didn't say that they not working. My concern here is that overused overloading is not reliable.

Feel free to update patch (just keep eyes on documentation).

Quantizs:
Hi,
When will this feature be merged ?

dkurt:
@Quantizs, As we discussed with @alalek, I'd like to recommend to rename methods in the same way: `readNetFromModelOptimizer` because in Python they can work fine (with `bytearray`). So you can fix it.

modules/dnn/include/opencv2/dnn/dnn.hpp
dkurt: @Quantizs, please change `Net::readFromModelOptimizer*` as well.

PR #15739, opened 2019-10-18
Enable Eltwise layer with different numbers of inputs channels

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/15724

This PR enables Eltwise layer (sum, prod, max) with input tensors with different number of channels.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/679

<cut/>

```
Median (ms)

                       Name of Test                          3.4     pr        pr    
                                                                               vs    
                                                                              3.4    
                                                                           (x-factor)
AlexNet::DNNTestNetwork::OCV/CPU                           14.343  14.430     0.99   
DenseNet_121::DNNTestNetwork::OCV/CPU                      37.195  37.235     1.00   
EAST_text_detection::DNNTestNetwork::OCV/CPU               67.625  67.560     1.00   
ENet::DNNTestNetwork::OCV/CPU                              44.997  45.185     1.00   
FastNeuralStyle_eccv16::DNNTestNetwork::OCV/CPU            115.694 116.573    0.99   
GoogLeNet::DNNTestNetwork::OCV/CPU                         15.383  15.265     1.01   
Inception_5h::DNNTestNetwork::OCV/CPU                      16.909  16.771     1.01   
Inception_v2_Faster_RCNN::DNNTestNetwork::OCV/CPU          286.174 287.129    1.00   
Inception_v2_SSD_TensorFlow::DNNTestNetwork::OCV/CPU       42.374  42.402     1.00   
MobileNet_SSD_Caffe::DNNTestNetwork::OCV/CPU               19.031  19.038     1.00   
MobileNet_SSD_v1_TensorFlow::DNNTestNetwork::OCV/CPU       20.979  20.971     1.00   
MobileNet_SSD_v2_TensorFlow::DNNTestNetwork::OCV/CPU       29.889  29.921     1.00   
OpenFace::DNNTestNetwork::OCV/CPU                           3.887   3.916     0.99   
OpenPose_pose_mpi_faster_4_stages::DNNTestNetwork::OCV/CPU 602.822 605.825    1.00   
ResNet_50::DNNTestNetwork::OCV/CPU                         35.661  35.846     0.99   
SSD::DNNTestNetwork::OCV/CPU                               268.011 267.123    1.00   
SqueezeNet_v1_1::DNNTestNetwork::OCV/CPU                    3.860   3.913     0.99   
YOLOv3::DNNTestNetwork::OCV/CPU                            214.246 213.658    1.00   
[...]

Issue #15724, opened 2019-10-17
Missing Darknet neural network support: [shortcut] layers


VideoPlayerCode:
Ping @dkurt @liuyuns @dmonterom @vpisarev @arthur-williams @alalek @AlexeyAB (everyone who has ever worked on OpenCV's darknet module).

## New Neural Network Innovation!

There's an incredible new neural network innovation: YOLOv3-Tiny has been reduced to almost half CPU usage while retaining detection accuracy. They call it YOLOv3-Tiny-PRN. It's the new king of low-CPU usage networks! Here is the research and paper in progress (soon to be published): https://github.com/WongKinYiu/PartialResidualNetworks

Here's the performance difference on Darknet (it would be even faster in OpenCV):

![66791677-c76bbb00-ef27-11e9-93fe-842dab485bb1](https://user-images.githubusercontent.com/38923130/66974552-b2179d80-f09b-11e9-9a3d-b884e59fb9d6.png)

If you do a diff between yolov3-tiny-prn config and the original yolov3-tiny, you will see that the changes are that many layers have half the filter sizes, and there are many `[shortcut]` layers added instead.

You can also read their paper at the repo above.

The problem is that OpenCV's darknet loader lacks full support for `[shortcut]` layers. So nobody can enjoy this big PRN innovation on OpenCV's ultra-fast CPU neural network implementation...

## References

**Reference: Darknet's implementation of shortcut layers.**

- https://github.com/pjreddie/darknet/blob/b13f67bfdd87434e141af532cdb5dc1b8369aa3b/src/parser.c#L539-L556
- https://github.com/pjreddie/darknet/blob/f6d861736038da22c9eb0739dca84003c5a5e275/src/shortcut_layer.c#L9-L39

**Reference: OpenCV's incomplete implementation of shortcut layers.**

- https://github.com/opencv/opencv/blob/8c0b0714e76efef4a8ca2a7c410c60e55c5e9829/modules/dnn/src/darknet/darknet_io.cpp#L631-L641
- https://github.com/opencv/opencv/blob/8c0b0714e76efef4a8ca2a7c410c60e55c5e9829/modules/dnn/src/darknet/darknet_io.cpp#L412-L432
- We can see one big issue immediately: OpenCV's reading of the shortcut layers only reads the `from=` property. It totally ignores all other properties (mainly the important `activation=` property).
- As a sidenote: The PRN network uses `[shortcut]` with `activation=leaky` (leaky RELU).
[...]

PR #16063, opened 2019-12-04
support eltwise sum with different number of input channels in CUDA backend

YashasSamaga:
### This pullrequest changes:
- adds layer described in #15724 to the CUDA backend

**Timings (on GTX 1050):**

Model                     | Darknet     | OpenCV CUDA
------------------------ | ------------- | --------------------
YOLOv3 Tiny PRN  | 6.727ms   | **5.94ms**

**Pending:**
- enable FP16 tests (disabled by PR16010)

**References:**
#15724 
#15739 

<cut/>

```
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/dnn/src/cuda/shortcut.cu
+++ b/modules/dnn/src/cuda/shortcut.cu
@@ -41,7 +41,7 @@ namespace cv { namespace dnn { namespace cuda4dnn { namespace kernels {
                 const auto actual_idx = i * vector_type::size();
                 const auto b = actual_idx / batch_stride_input; /* `input` and `output` have the same shape */
                 const auto c = (actual_idx % batch_stride_input) / channel_stride;
-                const auto c_offset = (actual_idx % batch_stride_input) % channel_stride;
+                const auto c_offset = actual_idx % channel_stride;
 
                 vector_type vec_input;
                 v_load(vec_input, input_vPtr[i]);
--- a/modules/dnn/src/cuda/shortcut.cu
+++ b/modules/dnn/src/cuda/shortcut.cu
@@ -41,6 +41,7 @@ namespace cv { namespace dnn { namespace cuda4dnn { namespace kernels {
                 const auto actual_idx = i * vector_type::size();
                 const auto b = actual_idx / batch_stride_input; /* `input` and `output` have the same shape */
                 const auto c = (actual_idx % batch_stride_input) / channel_stride;
+                const auto c_offset = (actual_idx % batch_stride_input) % channel_stride;
 
                 vector_type vec_input;
                 v_load(vec_input, input_vPtr[i]);
@@ -53,7 +54,7 @@ namespace cv { namespace dnn { namespace cuda4dnn { namespace kernels {
                  * which have a corresponding channel in `from`.
                  */
                 if (c < c_from) {
-                    const auto from_actual_idx = b * batch_stride_from + c * channel_stride;
+                    const auto from_actual_idx = b * batch_stride_from + c * channel_stride + c_offset;
                     const auto from_vec_idx = from_actual_idx / vector_type::size();
 
                     vector_type vec_from;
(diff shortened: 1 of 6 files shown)


modules/dnn/src/cuda/shortcut.cu
```diff
@@ -0,0 +1,108 @@
...
+                 *
+                 * In this scheme, only some channels in the `output` differ from `input`. They differ in the channels
+                 * which have a corresponding channel in `from`.
+                 */
+                if (c < c_from) {
+                    const auto from_actual_idx = b * batch_stride_from + c * channel_stride;
+                    const auto from_vec_idx = from_actual_idx / vector_type::size();
+
+                    vector_type vec_from;
+                    v_load(vec_from, from_vPtr[from_vec_idx]);
```
YashasSamaga: If I change `from_vPtr[from_vec_idx]` to `from_vPtr[i]`, OCV CPU backend and the CUDA backend agree with their outputs. But I have doubts over which one is correct.

I used `input_shortcut_kernel` as reference for this PR:

https://github.com/AlexeyAB/darknet/blob/eac26226a7fc0a9da2b684a564f8f086eaf38390/src/blas_kernels.cu#L714-L729

```
int out_index = i*sample + w2*(j*sample + h2*(k + c2*b));
int add_index = i*stride + w1*(j*stride + h1*(k + c1*b));
out[out_index] = in[out_index] + add[add_index];
```
which is
```
int out_index = i + w*(j + h*(k + c2*b));
    int add_index = i + w*(j + h*(k + c1*b));
    out[out_index] = in[out_index] + add[add_index];
```
if the spatial dimensions are same.

As it can be seen, the values in the corresponding channels of both inputs are added. I can infer the same from https://github.com/opencv/opencv/pull/15739#issuecomment-544036806. This is what I think I have done in this PR.

The OpenCV CPU code does something different which I think is wrong and can potentially lead to segfaults.

The eltwise layer code in OCV backend assumes that the number of channels are same in all inputs. `globalDelta` is calculated [here](https://github.com/opencv/opencv/pull/15739/files#diff-7ac73ff12c29882cb913b6f09da2f82cR237). The same `globalDelta` is also used here for indexing another input [here for example](https://github.com/opencv/opencv/pull/15739/files#diff-7ac73ff12c29882cb913b6f09da2f82cR290). But two inputs have different number of channels and hence, the same `globalDelta` is not applicable. If the second input had lesser number of channels than for which `globalDelta` was calculated for, it could lead to an illegal memory access in the second input.

Is this a bug in the OCV backend or I have misunderstood what the shortcut layer does?

/cc @dkurt

YashasSamaga: I tried running the object detection sample on both variants. Unfortunately, both give reasonable outputs and hence it isn't obvious which one is correct.

alalek: Perhaps we need some test for this single layer only.

YashasSamaga: Silly me. The two variants are identical when batch size is one which was the case in the object detection sample. When I run a batch of 10 images, the CUDA and OCV backend agree on the first image only as expected.

YashasSamaga: I did some tests with two images in a batch.

https://gist.github.com/YashasSamaga/b895dc109d50cbbb4e171c5d8e64defd

The CUDA backend agrees with the darknet output while the CPU backend does not. So it's a bug in the OpenCV CPU backend.

alalek:
@YashasSamaga Please rebase this patch on latest master (there is conflicts after merging #16087).

modules/dnn/src/layers/eltwise_layer.cpp
```diff
@@ -466,6 +473,13 @@ class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
     {
         auto context = reinterpret_cast<csl::CSLContext*>(context_);
 
+        if(variableChannels)
```
alalek: `variableChannels` has been removed by #16088

Perhaps check here should looks like
```
channelsMode == ELTWISE_CHANNNELS_INPUT_0
    || channelsMode == ELTWISE_CHANNNELS_INPUT_0_TRUNCATE
```

and else branch should ensure that `channelsMode == ELTWISE_CHANNNELS_SAME`.

YashasSamaga:
I just noticed something weird. YOLOv3 has started using these shortcut kernels instead of the regular eltwise. Need to check.

PR #16066, opened 2019-12-05
G-API: Added G-API Overview slides & its source code

dmatveev:
- Sample code snippets are moved to separate files;
- Introduced a separate benchmark to measure Fluid/OpenCV
  performance;
- Added notes on API changes (it is still a 4.0, not a 4.2 talk!)

[gapi_overview.pdf](https://github.com/opencv/opencv/files/3926944/gapi_overview.pdf)


dmatveev:
OK. The file is originally stored at OpenCV Github Wiki already. BTW this is the first and the only ever public presentation on G-API

dmatveev:
[gapi_overview.pdf](https://github.com/opencv/opencv/files/3926944/gapi_overview.pdf)

@alalek attached. This presentation has been reviewed by @vpisarev already a year back

dmatveev:
@alalek also how can we omit .sty and .eps files from the "trailing whitespace" check? Not sure if they broke if I fix the files themselves, though :)

alalek:
> attached

Thanks.

---

> .sty and .eps files

Looks like they are external.
Could we add some instructions how to get them? (through `wget` for example)

---

Could you take a look on multiple "FIXME" lines?

dmatveev:
> Could we add some instructions how to get them? (through wget for example)

There's a README referring to the source of the template. https://github.com/matze/mtheme

.sty files are product of LaTeX processing so there's no easy way to just `wget` it, it also requires a LaTeX package

dmatveev:
> Could you take a look on multiple "FIXME" lines?

@alalek those FIXMEs are mainly about utilizing some features of org-mode and Beamer which I couldn't get working properly that last time. E.g. it can automatically break a long page to multiple slides preserving the common title, but I haven't managed to get that working that time.

There's nothing more left to address from the content / look perspective

dmatveev:
@alalek replaced precompiled *.sty files and Makefile with a download-n-build script. It is way more cleaner now

dmatveev:
@alalek can this be merged? Anything else required?

modules/gapi/doc/slides/get_sty.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+MTHEME_VER=2fa6084b9d34fec9d2d5470eb9a17d0bf712b6c8
```
alalek: Enable errors in the beginning of the script: `set -e`

.gitignore
```diff
@@ -21,5 +21,7 @@ bin/
 *.suo
 *.log
 *.tlog
+*.bbl
+*.blg
```
alalek: Please move this into dedicated `.gitignore` + add `*.sty` / `mtheme.sty` pattern:
- modules/gapi/doc/slides/.gitignore

modules/gapi/doc/slides/get_sty.sh
```diff
@@ -0,0 +1,21 @@
...
+
+    # Download template from Github
+    tmp_dir=$(mktemp -d)
+    wget -P "$tmp_dir" -c https://github.com/matze/mtheme/archive/${MTHEME_VER}.tar.gz
+    pushd "$tmp_dir"
+    tar -xzvf "$MTHEME_VER.tar.gz"
+    popd
+    make -C "$tmp_dir"/mtheme-"$MTHEME_VER"
+    cp   -v "$tmp_dir"/mtheme-"$MTHEME_VER"/*.sty "$MTHEME_DIR"
+    rm -r "$tmp_dir"
```
alalek: Let's make git happy in case of branch switching:
```
echo "*" > "$MTHEME_DIR/.gitignore"
```

dmatveev: nice

dmatveev: yeah but do you think it makes sense to put this `mtheme.sty/.gitignore` under a regular source control? I believe it will show as "untracked" otherwise

dmatveev: well, it works even without that

.gitignore
dmatveev: done

modules/gapi/doc/slides/get_sty.sh
dmatveev: done

dmatveev:
@alalek anything else?

dmatveev:
Awesome, thanks!

PR #16080, opened 2019-12-06
G-API: Mosaic handle corner cases

TolyaTalamanov:

### DIFF
--- a/modules/gapi/src/api/render_ocv.cpp
+++ b/modules/gapi/src/api/render_ocv.cpp
@@ -20,25 +20,53 @@ inline void mosaic(cv::Mat& mat, const cv::Rect &rect, int cellSz)
     auto intersection = mat_rect & rect;
 
     cv::Mat msc_roi = mat(intersection);
-    int crop_x = msc_roi.cols - msc_roi.cols % cellSz;
-    int crop_y = msc_roi.rows - msc_roi.rows % cellSz;
+
+    bool has_crop_x = false;
+    bool has_crop_y = false;
+
+    int cols = msc_roi.cols;
+    int rows = msc_roi.rows;
+
+    if (msc_roi.cols % cellSz != 0)
+    {
+        has_crop_x = true;
+        cols -= msc_roi.cols % cellSz;
+    }
+
+    if (msc_roi.rows % cellSz != 0)
+    {
+        has_crop_y = true;
+        rows -= msc_roi.rows % cellSz;
+    }
 
     cv::Mat cell_roi;
-    for(int i = 0; i < crop_y; i += cellSz ) {
-        for(int j = 0; j < crop_x; j += cellSz) {
+    for(int i = 0; i < rows; i += cellSz )
+    {
+        for(int j = 0; j < cols; j += cellSz)
+        {
             cell_roi = msc_roi(cv::Rect(j, i, cellSz, cellSz));
             cell_roi = cv::mean(cell_roi);
         }
-        cell_roi = msc_roi(cv::Rect(crop_x, i, msc_roi.cols - crop_x, cellSz));
-        cell_roi = cv::mean(cell_roi);
+        if (has_crop_x)
+        {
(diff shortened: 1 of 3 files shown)


TolyaTalamanov:
@dmatveev

TolyaTalamanov:
@smirnov-alexey @AsyaPronina

modules/gapi/src/api/render_ocv.cpp
```diff
@@ -16,16 +16,27 @@ namespace draw
...
     for(int i = 0; i < crop_y; i += cellSz ) {
         for(int j = 0; j < crop_x; j += cellSz) {
-            auto cell_roi = msc_roi(cv::Rect(j, i, cellSz, cellSz));
+            cell_roi = msc_roi(cv::Rect(j, i, cellSz, cellSz));
             cell_roi = cv::mean(cell_roi);
         }
     }
+
+    // Handle tail
+    cell_roi = msc_roi(cv::Rect(crop_x, 0, msc_roi.cols - crop_x, crop_y));
```
alalek: - Proper handling of "no tail" cases
- Are there two rectangles only for the tail processing?
- I believe you should not use rectangles greater than `cellSz x cellSz`. `0 .. crop_y` violates this condition.

TolyaTalamanov: Fixed

TolyaTalamanov:
@alalek Could you look at it again ?

modules/gapi/test/render/gapi_render_tests_ocv.cpp
```diff
@@ -493,15 +493,41 @@ INSTANTIATE_TEST_CASE_P(RenderNV12OCVTestFTextsImpl, RenderNV12OCVTestFTexts,
                             Values(cv::Scalar(0, 255, 0))));
 #endif // HAVE_FREETYPE
 
+// FIXME Implement a macros to instantiate the tests because BGR and NV12 have the same parameters
+
 INSTANTIATE_TEST_CASE_P(RenderBGROCVTestMosaicsImpl, RenderBGROCVTestMosaics,
                         Combine(Values(cv::Size(1280, 720)),
```
alalek: @dmatveev Consider avoiding using of HD-sized images in accuracy tests.

These tests are filtered from:
- Debug pre-commit build
- coverage
- valgrind

Only VGA and less sized images are run for these builder types.

---

BTW, If it is highly required to work with HD resolution, then don't make them parametrized.

modules/gapi/src/api/render_ocv.cpp
alalek: > Proper handling of "no tail" cases

Test still crashes in "Debug" build due extra checks (don't look on pre-commit, it doesn't care about 720p test cases).

modules/gapi/test/render/gapi_render_tests_ocv.cpp
alalek: @dmatveev Consider avoiding using of HD-sized images in accuracy tests of G-API.

These tests are filtered from:
- Debug pre-commit build
- coverage runs
- valgrind runs

Only VGA and less sized images are run for these builder types.

---

BTW, If it is highly required to work with HD+ resolution, then don't make these tests parametrized on image resolution.

dmatveev: ok we'll address it in our next tests refactoring round, thanks!

dmatveev:
@alalek This stability thing is required for the upcoming internal release, can we have it in 4.2 (which I believe will serve as a basis for that)?

alalek:
Current patch is buggy.
It throws exceptions in Debug builds (related tests are filtered out on public CI due HD resolution - try run them on local machine).

TolyaTalamanov:
@alalek Could you take a look ?

dmatveev:
Thanks!

TolyaTalamanov:
Great! Thanks.

PR #16082, opened 2019-12-06
add ROI pooling layer to the CUDA backend

YashasSamaga:
### This pullrequest changes
- adds ROI pooling layer to the CUDA backend

<cut/>

```
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/dnn/src/cuda4dnn/primitives/roi_pooling.hpp
+++ b/modules/dnn/src/cuda4dnn/primitives/roi_pooling.hpp
@@ -1,0 +1,52 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef OPENCV_DNN_SRC_CUDA4DNN_PRIMITIVES_ROI_POOLING_HPP
+#define OPENCV_DNN_SRC_CUDA4DNN_PRIMITIVES_ROI_POOLING_HPP
+
+#include "../../op_cuda.hpp"
+
+#include "../csl/stream.hpp"
+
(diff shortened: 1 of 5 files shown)


modules/dnn/src/cuda4dnn/primitives/roi_pooling.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include "../kernels/roi_pooling.hpp"
+
+#include <utility>
+
+namespace cv { namespace dnn { namespace cuda4dnn {
+
+    template <class T>
+    class ROIPoolingOp final : public CUDABackendNode {
+    public:
```
alalek: Consider avoiding indentation in "whole" file namespaces.

modules/dnn/src/cuda4dnn/primitives/roi_pooling.hpp
```diff
@@ -0,0 +1,52 @@
...
+        using wrapper_type = GetCUDABackendWrapperType<T>;
+
+        ROIPoolingOp(csl::Stream stream_, float spatial_scale)
```
alalek: Does `std::move()` used properly below?

YashasSamaga: Yes, I think. The streams are sharable RAII objects. The `stream_` argument is not a reference. Hence, a copy is created at the site where the constructor is invoked. This copy is moved into `stream` which is a member.

modules/dnn/src/cuda4dnn/primitives/roi_pooling.hpp
YashasSamaga: Should I push a commit now?

PR #16086, opened 2019-12-06
G-API(test): exclude 32F from bitwise_not case

alalek:
NaN result is not properly checked anyway


dmatveev:
@AsyaPronina @rgarnov can you please have a look on this?

PR #16088, opened 2019-12-06
dnn(eltwise): fix handling of different number of channels

alalek:
relates #15739

> The eltwise layer code in OCV backend assumes that the number of channels are same in all inputs.

Details: https://github.com/opencv/opencv/pull/16063/files#r353899705

<cut/>

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/darknet/darknet_io.cpp
+++ b/modules/dnn/src/darknet/darknet_io.cpp
@@ -425,6 +425,7 @@ namespace cv {
                     }
 
                     shortcut_param.set<std::string>("op", "sum");
+                    shortcut_param.set<std::string>("output_channels_mode", "input_0_truncate");
 
                     darknet::LayerParameter lp;
                     std::string layer_name = cv::format("shortcut_%d", layer_id);
--- a/modules/dnn/src/layers/eltwise_layer.cpp
+++ b/modules/dnn/src/layers/eltwise_layer.cpp
@@ -80,14 +80,14 @@ public:
         ELTWISE_CHANNNELS_USE_MAX,               //!< number of channels from inputs may be different,
                                                  //!< output's number of channels is equal to maximal number of input channels
                                                  //!< @note supported operation: `SUM`
-    } channelsMode;
+    } channelsModeInput;
 
-    bool variableChannels;                       //!< number of channels are different on input
-    /*size_t*/int outputChannels;
+
+    mutable OutputChannelsMode channelsMode;     //!< "optimized" channels mode (switch to ELTWISE_CHANNNELS_SAME if number of input channels are equal)
+    mutable /*size_t*/int outputChannels;
 
     EltwiseLayerImpl(const LayerParams& params)
-        : variableChannels(false)
-        , outputChannels(0)
+        : outputChannels(0)
     {
         setParamsFrom(params);
         op = SUM;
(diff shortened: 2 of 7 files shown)


modules/dnn/src/layers/eltwise_layer.cpp
```diff
@@ -97,6 +116,34 @@ class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
...
             }
         }
+
+        channelsMode = ELTWISE_CHANNNELS_SAME;
+        if (params.has("output_channels_mode"))
+        {
+            String v = toLowerCase(params.get<String>("output_channels_mode"));
+            if (v == "same")
+            {
+                channelsMode = ELTWISE_CHANNNELS_SAME;
```
dkurt: Maybe we can use "same" as a default value?

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -425,6 +425,7 @@ namespace cv {
                     }
 
                     shortcut_param.set<std::string>("op", "sum");
+                    shortcut_param.set<std::string>("output_channels_mode", "input_0_truncate");
```
dkurt: According to https://github.com/opencv/opencv/pull/15739#issuecomment-544036806, Darknet supports two modes of summation, `ELTWISE_CHANNNELS_INPUT_0` and `ELTWISE_CHANNNELS_INPUT_0_TRUNCATE` in our terms. Can we just have a single `ELTWISE_CHANNNELS_INPUT_0` that means output number of channels is the same as first input's number of channels.

There is a code which can swap inputs to have unified order of inputs: https://github.com/opencv/opencv/pull/15739/files#diff-7ac73ff12c29882cb913b6f09da2f82cR179-R185

modules/dnn/src/layers/eltwise_layer.cpp
```diff
@@ -137,194 +204,307 @@ class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
...
         variableChannels = false;
-        for (int i = 1; i < inputs.size(); ++i)
+        outputChannels = inputs[0].size[1];
+        for (size_t i = 1; i < inputs.size(); ++i)
         {
-            if (inputs[i].size[1] != inputs[0].size[1])
-            {
+            int input_channels = inputs[i].size[1];
+            if (outputChannels != input_channels)
                 variableChannels = true;
```
dkurt: May be it would be more efficient to switch `channelsMode` to `ELTWISE_CHANNNELS_SAME` here? This way we can remove `variableChannels` completely.

alalek: It is handled in this way in `EltwiseInvoker` below:
```
                if (!self.variableChannels)
                {
                    CV_Assert(srcs[i].size == dst.size);
                }
                else if (self.channelsMode == ELTWISE_CHANNNELS_INPUT_0)
                ...
```

if `self.variableChannels == false` then `channelsMode` is not checked anymore.
So it is efficient.

---

> switch channelsMode to ELTWISE_CHANNNELS_SAME

I would like to avoid changing of input parameters. Lets do not "remove" extra checks for next network reshapes.

modules/dnn/src/layers/eltwise_layer.cpp
```diff
@@ -137,194 +204,307 @@ class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
...
-                    if (!p.coeffs.empty())
-                        std::swap(p.coeffs[j - 1], p.coeffs[j]);
+                    CV_Assert(srcs[i].size == dst.size);
+                }
+                else if (self.channelsMode == ELTWISE_CHANNNELS_INPUT_0)
+                {
+                    if (i == 0)
+                        CV_Assert(srcs[0].size == dst.size);
+                    CV_Assert(self.outputChannels >= p.srcNumChannels[i]);
+                }
```
dkurt: It seems to me that we need to sort the inputs here as well: 2, 4, 3 channels, in example

alalek: BTW, processing doesn't require sorted inputs. Result is correct anyway, but may be less efficient.

Updated code.

modules/dnn/test/test_darknet_importer.cpp
```diff
@@ -115,6 +116,42 @@ class Test_Darknet_layers : public DNNTestLayer
         net.setInput(inp);
         Mat out = net.forward();
         normAssert(out, ref, "", default_l1, default_lInf);
+
+        if (inp.size[0] == 1)  // test handling of batch size
```
dkurt: Can we add a boolean flag to skip this test? Looks like there is a problem with Myriad target and `testDarknetLayer("shortcut");`:

```
[ RUN      ] Test_Darknet_layers.shortcut/1, where GetParam() = DLIE/MYRIAD
/build/precommit_custom_linux/3.4/opencv/modules/dnn/test/test_darknet_importer.cpp:137: Failure
Expected equality of these values:
  0
  cv::norm(out2(ranges0), out2(ranges1), NORM_INF)
    Which is: 4.29004
Batch result is not equal: shortcut
Google Test trace:
/build/precommit_custom_linux/3.4/opencv/modules/dnn/test/test_darknet_importer.cpp:122: batch size 2
/build/precommit_custom_linux/3.4/opencv/modules/dnn/test/test_darknet_importer.cpp:102: shortcut
/build/precommit_custom_linux/3.4/opencv/modules/dnn/test/test_common.impl.hpp:66: Failure
Expected: (normL1) <= (l1), actual: 1.97884 vs 0.004
Google Test trace:
/build/precommit_custom_linux/3.4/opencv/modules/dnn/test/test_darknet_importer.cpp:122: batch size 2
/build/precommit_custom_linux/3.4/opencv/modules/dnn/test/test_darknet_importer.cpp:102: shortcut
/build/precommit_custom_linux/3.4/opencv/modules/dnn/test/test_common.impl.hpp:69: Failure
Expected: (normInf) <= (lInf), actual: 4.29028 vs 0.02
Google Test trace:
/build/precommit_custom_linux/3.4/opencv/modules/dnn/test/test_darknet_importer.cpp:122: batch size 2
/build/precommit_custom_linux/3.4/opencv/modules/dnn/test/test_darknet_importer.cpp:102: shortcut
[  FAILED  ] Test_Darknet_layers.shortcut/1, where GetParam() = DLIE/MYRIAD (8220 ms)
```

PR #16090, opened 2019-12-06
cuda_mog2_issue_5296

### LINKED ISSUE #5296: Multiple MOG2_GPU objects share the same parameter values
Hi,

While using MOG2_GPU class, I realized that if you have two MOG2_GPU objets whenever you modify any parameter of one of the objects, the other is affected too.

Taking a look at OpenCV code, I think that this is related to the declaration of MOG2_GPU parameters in file "opencv/modules/cudabgsegm/src/cuda/mog2.cu", where MOG2_GPU variables are allocated only once, independently from the number of MOG2_GPU objects . As a consequence, each object that calls the loadConstants method, overwrites the variables values affecting every MOG2_GPU object.

I think that this implementation is similar to other gpu classes, so I suppose the same situation takes place with these classes. 

My question is whether there is any plan to modify this behavior or this was implemented intentionally and there will be no modification in the future.

jeffeDurand:
resolves #5296

### This pullrequest changes

These modifications allow a cv::cuda::BackgroundSubtractorMOG2 instance to be unaffected by another created with different parameters.

```
force_builders=Custom,docs
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/cudabgsegm/src/cuda/mog2.hpp
+++ b/modules/cudabgsegm/src/cuda/mog2.hpp
@@ -1,0 +1,37 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#ifndef OPENCV_CUDA_MOG2_H
+#define OPENCV_CUDA_MOG2_H
+
+#include "opencv2/core/cuda.hpp"
+
+struct CUstream_st;
+typedef struct CUstream_st *cudaStream_t;
+
+namespace cv { namespace cuda {
+
+class Stream;
+
+namespace device { namespace mog2 {
+
+typedef struct
+{
+    float Tb_;
+    float TB_;
+    float Tg_;
+    float varInit_;
+    float varMin_;
+    float varMax_;
+    float tau_;
+    int nmixtures_;
+    unsigned char shadowVal_;
+} Constants;
+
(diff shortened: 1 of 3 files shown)


modules/cudabgsegm/src/cuda/mog2.hpp
```diff
@@ -0,0 +1,27 @@
+#ifndef OPENCV_CUDA_MOG2_H
```
alalek: Please add [short license header](https://github.com/opencv/opencv/wiki/Coding_Style_Guide#file-structure).

modules/cudabgsegm/src/mog2.cpp
```diff
@@ -41,209 +41,222 @@
...
-    {
-        void loadConstants(int nmixtures, float Tb, float TB, float Tg, float varInit, float varMin, float varMax, float tau, unsigned char shadowVal);
-        void mog2_gpu(PtrStepSzb frame, int cn, PtrStepSzb fgmask, PtrStepSzb modesUsed, PtrStepSzf weight, PtrStepSzf variance, PtrStepSzb mean, float alphaT, float prune, bool detectShadows, cudaStream_t stream);
-        void getBackgroundImage2_gpu(int cn, PtrStepSzb modesUsed, PtrStepSzf weight, PtrStepSzb mean, PtrStepSzb dst, cudaStream_t stream);
-    }
-}}}
+namespace mog2
+{
+void mog2_gpu(PtrStepSzb frame, int cn, PtrStepSzb fgmask, PtrStepSzb modesUsed, PtrStepSzf weight, PtrStepSzf variance, PtrStepSzb mean, float alphaT, float prune, bool detectShadows, const Constants *const constants, cudaStream_t stream);
+void getBackgroundImage2_gpu(int cn, PtrStepSzb modesUsed, PtrStepSzf weight, PtrStepSzb mean, PtrStepSzb dst, const Constants *const constants, cudaStream_t stream);
```
alalek: Perhaps these definitions should be moved into created "cuda/mog2.hpp" header.

PR #16094, opened 2019-12-08
Add eps error checking for approxPolyDP to allow sensible values only

### LINKED ISSUE #16053: approxPolyDP OOMs

- OpenCV => 3.4
- Compiler => x86_64-linux-gnu-g++-8

##### Detailed description

For certain contours, if approxPolyDP is passed an epsilon of NaN, then it OOMs.

This may simply be a request for input validation in approxPolyDP, because I believe an epsilon of NaN is meaningless.

##### Steps to reproduce

Here is a fairly minimal repro. Note that this uses the same contour as in the repro of #16051. You may need to tweak the headers/namespaces.
```
#include "third_party/OpenCVX/include/opencv2/imgproc/imgproc.hpp"
int main(int argc, char **argv) {
  std::vector<cv::Point> contour(
      {cv::Point(2085, 1415), cv::Point(2087, 1415), cv::Point(2089, 1414),
       cv::Point(2089, 1414), cv::Point(2087, 1412)});
  float epsilon = NAN;
  std::vector<cv::Point> approx;
  cv::approxPolyDP(contour, approx, epsilon,
                    true /* closed curve */);
  // The process will OOM before getting here.
}
```

saskatchewancatch:
for epsilon value of Douglas-Peucker algorithm.

resolves #16053

### DIFF
--- a/modules/imgproc/test/test_approxpoly.cpp
+++ b/modules/imgproc/test/test_approxpoly.cpp
@@ -363,10 +363,13 @@ TEST(Imgproc_ApproxPoly, bad_epsilon)
     inputPoints.push_back(Point2f(0.0f, 0.0f));
     std::vector<Point2f> outputPoints;
 
-    double eps = INFINITY;
+    double eps = std::numeric_limits<double>::infinity();
     ASSERT_ANY_THROW(approxPolyDP(inputPoints, outputPoints, eps, false));
 
     eps = 9e99;
+    ASSERT_ANY_THROW(approxPolyDP(inputPoints, outputPoints, eps, false));
+
+    eps = -1e-6;
     ASSERT_ANY_THROW(approxPolyDP(inputPoints, outputPoints, eps, false));
 
     eps = NAN;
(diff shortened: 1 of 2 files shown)


modules/imgproc/test/test_approxpoly.cpp
```diff
@@ -355,4 +355,22 @@ void CV_ApproxPolyTest::run( int /*start_from*/ )
...
 
+//Tests to make sure that unreasonable epsilon (error)
+//values never get passed to the Douglas-Peucker algorithm.
+TEST(Imgproc_ApproxPoly, bad_epsilon)
+{
+    std::vector<Point2f> inputPoints;
+    inputPoints.push_back(Point2f(0.0f, 0.0f));
+    std::vector<Point2f> outputPoints;
+
+    double eps = INFINITY;
```
alalek: Please use `std::numeric_limits<double>::infinity()` instead of C macros.

modules/imgproc/test/test_approxpoly.cpp
```diff
@@ -355,4 +355,22 @@ void CV_ApproxPolyTest::run( int /*start_from*/ )
...
+    ASSERT_ANY_THROW(approxPolyDP(inputPoints, outputPoints, eps, false));
+
+    eps = 9e99;
+    ASSERT_ANY_THROW(approxPolyDP(inputPoints, outputPoints, eps, false));
+
+    eps = NAN;
+    ASSERT_ANY_THROW(approxPolyDP(inputPoints, outputPoints, eps, false));
```
alalek: Please add case with negative epsilon (`-1e-6`).

PR #15950, opened 2019-11-19
ffmpeg/4.x: update FFmpeg plugin

alalek:
**Merge with 3rdparty**: https://github.com/opencv/opencv_3rdparty/pull/48
relates #15468

```
force_builders=Win32,Custom Win
buildworker:Win64 OpenCL=windows-2
```

PR #16103, opened 2019-12-09
videoio(plugins): relax version check for FFmpeg wrapper on Windows

alalek:
Should fix messages like these:
> [ERROR:0] global C:\build\precommit_windows64\opencv\modules\videoio\src\backend_plugin.cpp (215) cv::impl::PluginBackend::PluginBackend
> Video I/O: wrong OpenCV version used by plugin 'FFmpeg OpenCV Video I/O plugin': 4.1, OpenCV version is '4.2.0-pre'

relates #16078 #15950


alalek:
:+1:

PR #16104, opened 2019-12-09
calib3d: clarify error messages in collectCalibrationData()

### LINKED ISSUE #16095: Unintuitive exception message in C++ (calibrateCamera: assertion failed, ni == ni1)
OpenCV version 4.1.1
on Linux openSUSE Tumbleweed
Compiler is g++

##### Detailed description
I get this error message when trying to call calibrateCamera. I searched on it in some foruns and found out that it can mean that either:

- my vectors of object points and image points are not the same size (in my case, they are, I did check that)

- the type is wrong (in this case, I have no idea what to do).

Anyway, this message is really not intuitive and aside from some help with my situation I'd appreciate if this exception could have a better error message.
```
  what():  OpenCV(4.1.1) /home/abuild/rpmbuild/BUILD/opencv-4.1.1/modules/calib3d/src/calibration.cpp:3349: error: (-215:Assertion failed) ni == ni1 in function 'collectCalibrationData'
```

##### Steps to reproduce

My module can be seen here: https://github.com/torresguilherme/bakeneko/blob/master/src/calibration/calibration.cpp

thanks in advance for your work and the assistance

alalek:
resolves #16095

### DIFF
--- a/modules/calib3d/src/calibration.cpp
+++ b/modules/calib3d/src/calibration.cpp
@@ -3115,28 +3115,30 @@ static void collectCalibrationData( InputArrayOfArrays objectPoints,
                                     Mat& npoints )
 {
     int nimages = (int)objectPoints.total();
-    int i, j = 0, ni = 0, total = 0;
-    CV_Assert(nimages > 0 && nimages == (int)imagePoints1.total() &&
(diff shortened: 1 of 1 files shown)


alalek:
:+1:

PR #16120, opened 2019-12-10
python: fix issue with bindings loading on Python 3.8

alalek:
DLL loading process is changed: https://github.com/python/cpython/pull/12302
`PATH` env variable is not used anymore for DLL searching.

```
allow_multiple_commits=1
```


alalek:
Validated with [Winpack](http://pullrequest.opencv.org/buildbot/builders/3_4_winpack_controller/builds/614).

PR #16125, opened 2019-12-11
core: provide safe implementations of CV_XADD() only

alalek:
relates #16123


zchrissirhcz:
Hi, @alalek 
I'm late here for this issue. I wonder if the modified line, missing return type? i.e.

Current:
```c++
CV_INLINE CV_XADD(int* addr, int delta) { int tmp = *addr; *addr += delta; return tmp; }
```
Should be:
```
CV_INLINE int CV_XADD(int* addr, int delta) { int tmp = *addr; *addr += delta; return tmp; }
```

https://github.com/opencv/opencv/blob/1b64851fa88700281f58812e22eff56607eb1dd1/modules/core/include/opencv2/core/cvdef.h#L557

PR #16139, opened 2019-12-12
core: workaround flipHoriz() alignment issues

alalek:


tomoaki0705:
Nice and clean code!
However, this issue is limited to Arm v7 + uint64 combination.
I'd vote to work inside NEON universal intrinsic rather than working in `flipHoriz` part.
The issue became visible in `flipHoriz`, but could happen anywhere using 64bit integer `v_load`
It's a dirty hack, but how about working with [`v_reinterpret_as_u64`](https://github.com/opencv/opencv/blob/92b98888370128143a7d9180198a9336c2271a69/modules/core/include/opencv2/core/hal/intrin_neon.hpp#L331) in `intrin_neon.hpp`?

tomoaki0705:
I don't mean to disturb you, but alalek, could you have a look on my PR #16152 plz ?

PR #16160, opened 2019-12-13
optimize cvCeil and cvFloor in fase_math.hpp

giantpanpan:


alalek:
Why do you think that this is optimal? Optimized against that part of code?

PR #16161, opened 2019-12-14
cuda4dnn(concat): write outputs from previous layers directly into concat's output

YashasSamaga:
### This pullrequest changes
- optimizes out concat layer by forcing the layers providing input to the concat to directly write to the corresponding section of concat's output

<cut/>

```
force_builders=Custom,docs
buildworker:Custom=linux-4

test_opencl:Custom Win=OFF
```

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -2450,7 +2450,9 @@ struct Net::Impl
                      ld.layerInstance->type != "Concat")) )
                     continue;
 
-                if (preferableBackend == DNN_BACKEND_CUDA && IS_DNN_CUDA_TARGET(preferableTarget) && ld.layerInstance->type != "Convolution")
+                if (preferableBackend == DNN_BACKEND_CUDA && IS_DNN_CUDA_TARGET(preferableTarget)
+                    && ld.layerInstance->type != "Convolution"
+                    && ld.layerInstance->type != "Concat")
                     continue;
 
                 while (nextData)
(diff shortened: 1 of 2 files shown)


alalek:
Do you have any suggestions how to test that (avoid future regressions)?

YashasSamaga:
Oops, I should have rebased after PR#16092 was merged and enable more `Concat` fusions. Sorry.

modules/dnn/src/dnn.cpp
```diff
@@ -2639,6 +2641,22 @@ struct Net::Impl
 
                         if(inp_i_data->skip || inp_i_data->consumers.size() != 1)
                             break;
+#ifdef HAVE_CUDA
+                        if (preferableBackend == DNN_BACKEND_CUDA &&
+                            (inp_i_data->layerInstance->supportBackend(DNN_BACKEND_CUDA) == false || inp_i_data->layerInstance->type == "Region"))
```
YashasSamaga: Here is an opportunity for a future regression. If somebody adds a layer which computes the final output on CPU (like region layer) but they forget to add an entry here and also does not test what would happen if their layer appears before a concat (who would even think about a test like this), this will throw off assertions in `CUDABackendWrapper`.
Possible solutions:
- have a test which would connect every layer that exists in OCV DNN to a concat
- use a whitelist instead of a blacklist
  - allow concat fusion for few types of input layers like Convolution, Pooling, etc. (this is what OCL is doing currently)

alalek:
Can we track applied fusing through checking of `Net::dump()` output? (using some test networks)

YashasSamaga:
The `Net::dump()`  output appear to be identical with and without concat fusion. I don't know if this is intended.

modules/dnn/src/dnn.cpp
asmorkalov: let's use whitelist. Stability has higher priority than performance optimization for rare cases.

PR #16164, opened 2019-12-14
[WIP] cuda4dnn(conv): fuse eltwise with convolutions

YashasSamaga:
### This pullrequest changes
- fuses eltwise with convolution for three topologies

<cut/>

### activation(conv + eltwise)
Examples: ResNet

```
--------------               ---------------
| Bias Layer |               | Convolution |
--------------               ---------------
      |                             |
      |       ----------------      |
      --------| Eltwise Sum  |-------
              ----------------
                     |
              ----------------
              |  Activation  |                 
              ----------------
                     |
```

### eltwise + activation(conv)
Examples: YOLOv3

```
--------------               ---------------
| Bias Layer |               | Convolution |
--------------               ---------------
      |                             |
      |                      ----------------
      |                      |  Activation  |                 
      |                      ----------------
      |                             |
      |       ----------------      |
      --------| Eltwise Sum  |-------
              ----------------
                     |
             
```

### conv + eltwise
```
--------------               ---------------
| Bias Layer |               | Convolution |
--------------               ---------------
      |                             |
      |       ----------------      |
      --------| Eltwise Sum  |-------
              ----------------
                     |
           
```

**Pending:**
- make a test network comprising all activations and all topologies for testing the entire PR

---

```
force_builders=Custom,docs
buildworker:Custom=linux-4

test_opencl:Custom Win=OFF
```

### DIFF
--- a/modules/dnn/src/layers/eltwise_layer.cpp
+++ b/modules/dnn/src/layers/eltwise_layer.cpp
@@ -64,13 +64,6 @@ namespace dnn
 class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
 {
 public:
-    enum EltwiseOp
-    {
-        PROD = 0,
-        SUM = 1,
-        MAX = 2,
-        DIV = 3
-    } op;
     std::vector<float> coeffs;
 
     enum OutputChannelsMode
--- a/modules/dnn/src/layers/eltwise_layer.cpp
+++ b/modules/dnn/src/layers/eltwise_layer.cpp
@@ -64,8 +64,6 @@ namespace dnn
 class EltwiseLayerImpl CV_FINAL : public EltwiseLayer
 {
 public:
-    std::vector<float> coeffs;
-
     enum OutputChannelsMode
     {
         ELTWISE_CHANNNELS_SAME = 0,              //!< number of channels from inputs must be the same and equal to output's number of channels
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -518,6 +518,8 @@ CV__DNN_INLINE_NS_BEGIN
     class CV_EXPORTS EltwiseLayer : public Layer
     {
     public:
+        std::vector<float> coeffs;
+
         enum EltwiseOp
         {
             PROD = 0,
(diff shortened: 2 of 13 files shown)


modules/dnn/src/layers/eltwise_layer.cpp
```diff
@@ -64,15 +64,6 @@ namespace dnn
...
-    enum EltwiseOp
-    {
-        PROD = 0,
-        SUM = 1,
-        MAX = 2,
-        DIV = 3
-    } op;
-    std::vector<float> coeffs;
-
     enum OutputChannelsMode
```
YashasSamaga: I have moved `EltwiseOp`, `op` and `coeffs` to `EltwiseLayer` (which is a part of the public headers). This will alter the layout of `EltwiseLayer` and hence is not ABI compatible. Is this ok?

alalek: Public API of DNN layers is `LayerParams`.

alalek:
> Note: first 8 commits are from PR16092 (GitHub doesn't seem to handle squashed merges well)

Please rebase on latest master and remove these 8 commits (these commits should go away completely from this PR). Try to rebase again:
```
git checkout -B pr_16164 bc3f5d8fbb4b3e04fae6fa8e1bf9d40de368ac38
#git remote add upstream https://github.com/opencv/opencv
git fetch upstream
git rebase -i upstream/master
# remove 8 lines ('d' + 'd' in Vim)
# Save+quit (Esc+wq)
```

In case of troubles:
```
git rebase --abort
```

modules/dnn/src/layers/eltwise_layer.cpp
alalek: > all_layers.hpp

Remove changes from there. This header is part of public API.

If you need to access to EltwiseLayerImpl, then create internal `modules/dnn/src/layers/eltwise_layer.hpp`

YashasSamaga: I need access to just `op` and `coeffs`. The CUDA backend can fuse eltwise if:
- the operation is `SUM`
- has no coefficients
- no variable channels

The variable channels is easy to check as the input shapes are available in `fuseLayers()`. I moved `op` and `coeffs` to `EltwiseLayer` (which  is in `all_layers.hpp`) so that they are accessible in `fuseLayers`.

I tried to see what the OCL code was doing to check the conditions but I couldn't find any such checks. Maybe I can use the same technique as OCL code to get those properties.

asmorkalov:
@YashasSamaga Do you have a chance to finish the PR? CI bot reports some issues on all platforms.

YashasSamaga:
I haven't been able to find where the OCL backend performs the checks on the eltwise operation and variable channels before fusing. I haven't tested but this could be a bug.

PR #16165, opened 2019-12-14
Mac os install

### LINKED ISSUE #13822: Updating installation instructions, adding package manager installation and macOS
https://docs.opencv.org/3.4.5/df/d65/tutorial_table_of_content_introduction.html

This link contains links to installation instructions for Linux, Windows and iOS.

A lot of macOS users will know that they can probably use the instructions for Linux (they can right? I'm not 100% sure) - but it's not safe to assume that they all will. Also, the Linux instructions do not mention that they can work on macOS, too (they only say`The following steps have been tested for Ubuntu 10.04 but should work with other distros as well.`). 

I don't have too much knowledge about OpenCV, but I feel like those instructions (for Linux, for example) should be updated. The fact that they mention testing on Ubuntu 10.04 is a hint towards that. They are also 100 lines long (if counting a few whitespace-only lines) and never mention installing via something like `apt-get opencv`. Is there a reason there isn't a package manager installation? I found instructions to install with `brew` on macOS, could that just be added there?

sajarindider:
resolves #13822

### This pullrequest adds 
a macOS installation guide based on the current linux installation guide.

### DIFF
--- a/doc/tutorials/introduction/macOS_install/macOS_install.markdown
+++ b/doc/tutorials/introduction/macOS_install/macOS_install.markdown
@@ -1,0 +1,110 @@
+Installation in MacOS {#tutorial_macOS_install}
+=====================
+
+The following steps have been tested for MacOX (Mavericks) but should work with other versions as well.
+
+Required Packages
+-----------------
+
+-   CMake 2.8.7 or higher
+-   Git
+-   Python 2.6 or later and Numpy 1.5 or later
+
+This tutorial will assume you have [Python](https://docs.python.org/3/using/mac.html),
+[Numpy](https://docs.scipy.org/doc/numpy-1.10.1/user/install.html) and
+[Git](https://www.atlassian.com/git/tutorials/install-git) installed on your machine.
+
+Installing CMake
+----------------
+-# Find the version for your system and download CMake from their release's [page](https://cmake.org/download/)
+
+-# Install the dmg package and launch it from Applications. That will give you the UI app of CMake
+
+-# From the CMake app window, choose menu Tools --> Install For Command Line Use.
+
+-# Install folder will be /usr/bin/ by default, submit it by choosing Install command line links.
+
+-# Test that it works by running
+    @code{.bash}
+    cmake --version
+    @endcode
+
+Getting OpenCV Source Code
+--------------------------
+
+You can use the latest stable OpenCV version or you can grab the latest snapshot from our [Git
+repository](https://github.com/opencv/opencv.git).
+
+### Getting the Latest Stable OpenCV Version
+
+-   Go to our [downloads page](http://opencv.org/releases.html).
+-   Download the source archive and unpack it.
+
+### Getting the Cutting-edge OpenCV from the Git Repository
+
+Launch Git client and clone [OpenCV repository](http://github.com/opencv/opencv). If you need
+modules from [OpenCV contrib repository](http://github.com/opencv/opencv_contrib) then clone it as well.
+
+For example
+@code{.bash}
+cd ~/<my_working_directory>
+git clone https://github.com/opencv/opencv.git
+git clone https://github.com/opencv/opencv_contrib.git
+@endcode
+Building OpenCV from Source Using CMake
+---------------------------------------
+
+-#  Create a temporary directory, which we denote as \<cmake_build_dir\>, where you want to put
+    the generated Makefiles, project files as well the object files and output binaries and enter
+    there.
+
+    For example
+    @code{.bash}
+    cd ~/opencv
+    mkdir build
+    cd build
+    @endcode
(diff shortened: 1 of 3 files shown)


doc/tutorials/introduction/macOS_install/macOS_install.markdown
```diff
@@ -0,0 +1,110 @@
...
+Required Packages
+-----------------
+
+-   CMake 2.8.7 or higher
+-   Git
+-   Python 2.6 or later and Numpy 1.5 or later
+
+This tutorial will assume you have [Python](https://docs.python.org/3/using/mac.html),
+[Numpy](https://docs.scipy.org/doc/numpy-1.10.1/user/install.html) and
+[Git](https://www.atlassian.com/git/tutorials/install-git) installed on your machine.
```
alalek: OS X already has Python (2.7).
Git has been install with Xcode (at least with xcode command-line tools).

doc/tutorials/introduction/macOS_install/macOS_install.markdown
```diff
@@ -0,0 +1,110 @@
...
+    @code{.bash}
+    cmake -D CMAKE_BUILD_TYPE=Release -D CMAKE_INSTALL_PREFIX=/usr/local ..
+    @endcode
+    or cmake-gui
+
+    -   set full path to OpenCV source code, e.g. /home/user/opencv
+    -   set full path to \<cmake_build_dir\>, e.g. /home/user/opencv/build
+    -   set optional parameters
+    -   run: �Configure�
+    -   run: �Generate�
```
alalek: > “

Please use ASCII quotes: "

doc/tutorials/introduction/macOS_install/macOS_install.markdown
```diff
@@ -0,0 +1,110 @@
...
+    make -j7 # runs 7 jobs in parallel
+    @endcode
+
+-# Finally, install using the following command:
+    @code{.bash}
+    make install
+    @endcode
+@note
+You can also use a package manager like [Homebrew](https://brew.sh/)
+or [pip](https://pip.pypa.io/en/stable/) to install official releases of OpenCV only (Not the cutting edge).
```
alalek: > official

Releases through pip or Homebrew are not "official" (in meaning, OpenCV team don't maintain them). "to install releases" should be ok.

doc/tutorials/introduction/macOS_install/macOS_install.markdown
```diff
@@ -0,0 +1,110 @@
...
+    -   build type: `CMAKE_BUILD_TYPE=Release\Debug`
+    -   to build with modules from opencv_contrib set OPENCV_EXTRA_MODULES_PATH to \<path to
+        opencv_contrib/modules/\>
+    -   set BUILD_DOCS for building documents
+    -   set BUILD_EXAMPLES to build all examples
+
+-#  [optional] Building python. Set the following python parameters:
+    -   PYTHON2(3)_EXECUTABLE = \<path to python\>
+    -   PYTHON_INCLUDE_DIR = /usr/include/python\<version\>
+    -   PYTHON2(3)_NUMPY_INCLUDE_DIRS =
```
alalek: > PYTHON2(3)_EXECUTABLE 

I would suggest using `PYTHON3_` prefix here (for all cases) and add note about how to specify Python2 through "PYTHON2_".

BTW, https://pythonclock.org/

doc/tutorials/introduction/macOS_install/macOS_install.markdown
```diff
@@ -0,0 +1,110 @@
...
+@code{.bash}
+cd ~/<my_working_directory>
+git clone https://github.com/opencv/opencv.git
+git clone https://github.com/opencv/opencv_contrib.git
+@endcode
+Building OpenCV from Source Using CMake
+---------------------------------------
+
+-#  Create a temporary directory, which we denote as \<cmake_build_dir\>, where you want to put
+    the generated Makefiles, project files as well the object files and output binaries and enter
```
alalek: > Makefiles

On OSX it makes sense to generate Xcode project instead of Makefiles.

doc/tutorials/introduction/macOS_install/macOS_install.markdown
```diff
@@ -0,0 +1,110 @@
...
+-#  Build. From build directory execute *make*, it is recommended to do this in several threads
+
+    For example
+    @code{.bash}
+    make -j7 # runs 7 jobs in parallel
+    @endcode
+
+-# Finally, install using the following command:
+    @code{.bash}
+    make install
```
alalek: It is not necessary to "install" OpenCV to use it. OpenCV can be integrated into other CMake projects through `OpencCV_DIR` pointing on build directory.

PR #16184, opened 2019-12-17
Added a helper function that will list all available Intel devices

neFAST:
compatible with OpenVino inference engine (cpu, gpus, movidius, etc.)

### This pullrequest changes

new cv::dnn::listInferenceEngineDevices() which is a wrapper on InferenceEngine::Core::GetAvailableDevices()

### DIFF
--- a/modules/dnn/include/opencv2/dnn/utils/inference_engine.hpp
+++ b/modules/dnn/include/opencv2/dnn/utils/inference_engine.hpp
@@ -42,6 +42,10 @@ CV_EXPORTS_W cv::String setInferenceEngineBackendType(const cv::String& newBacke
  */
 CV_EXPORTS_W void resetMyriadDevice();
 
+/** @brief List available devices
+ * The devices are returned as { "CPU", "GPU", "FPGA.0", "FPGA.1", "MYRIAD" }
+ */
+CV_EXPORTS_W std::vector<std::string> listInferenceEngineDevices();
 
 /* Values for 'OPENCV_DNN_IE_VPU_TYPE' parameter */
 #define CV_DNN_INFERENCE_ENGINE_VPU_TYPE_UNSPECIFIED ""
--- a/modules/dnn/src/op_inf_engine.cpp
+++ b/modules/dnn/src/op_inf_engine.cpp
@@ -1045,6 +1045,20 @@ void forwardInfEngine(const std::vector<Ptr<BackendWrapper> >& outBlobsWrappers,
 
 CV__DNN_INLINE_NS_BEGIN
 
+std::vector<std::string> listInferenceEngineDevices()
+{
+    std::vector<std::string> availableDevices;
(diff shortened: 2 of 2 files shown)


dkurt:
Thanks!

Can this method test the devices? I mean which lists it will show in the following scenarios:
* When there is a device but there is no plugin available (i.e. CPU with renamed MKLDNNPlugin)
* When there is a plugin but there is no device (i.e. MYRIAD)
* When there is a device and plugin but it's unavailable (i.e. with MYRIAD which is used by different process)
* When there is no plugin and there is no device (i.e. ARM CPU as in case of Raspberry Pi).

neFAST:
> Thanks!
> 
> Can this method test the devices? I mean which lists it will show in the following scenarios:
> 
> * When there is a device but there is no plugin available (i.e. CPU with renamed MKLDNNPlugin)
- [x] this is handled by OpenVino: no plugin means it won't show up in the list of available devices
> * When there is a plugin but there is no device (i.e. MYRIAD)
- [ ] this is handled by OpenVino: no device means it won't show up in the list of available devices
> * When there is a device and plugin but it's unavailable (i.e. with MYRIAD which is used by different process)
- [x] I thinks this is not handled but I'm unable to reproduce this right now
> * When there is no plugin and there is no device (i.e. ARM CPU as in case of Raspberry Pi).
- [x] handled as well, no plugin -> not in the list

dkurt:
Can this or similar method determine Myriad 2 or Myriad X?

Can you also experiment replacing checkIETarget from dnn.cpp using this new function? Thanks!

neFAST:
Well I did not know I could callcv::dnn::getAvailableBackends()
In op_inf_engine.cpp there's a isMyriadX() function

dkurt:
@neFAST, May I ask you if [getAvailableTargets(DNN_BACKEND_INFERENCE_ENGINE)](https://docs.opencv.org/master/d6/d0f/group__dnn.html#ga0534d8b21218317927530afe2d019495) can match your expectations?

neFAST:
@dkurt Do you mean I should replace calls to `bool checkIETarget(Target target)` https://github.com/opencv/opencv/blob/master/modules/dnn/src/dnn.cpp#L110 with my solution?
Now that you showed me that I can use getAvailableTargets(), my PR doesn't make sense anymore

dkurt:
@neFAST, You still can help us adding a wrapper for `getAvailableTargets` because for now it's C++ only. So you can do it in this PR (or it's even better to reopen it to 3.4 branch with a different branch name):
```
git checkout 3.4
git checkout -b branch_name
```

neFAST:
Sorry to be slow. You want me to reopen a PR on 3.4 with nothing more than the diff you posted just above?

dkurt:
@neFAST, Yes, if you don't mind.

neFAST:
Ok. I close this PR as a duplicate of #16232

PR #16190, opened 2019-12-18
Videocapture skip frames

pletessier:
### This pullrequest changes

**Add the possibility to skip frames** in VideoCapture by using the existing `OPENCV_FFMPEG_CAPTURE_OPTIONS` environment variable and the `AVCodecContext::skip_frame` method.

**AVDiscard options are :** 
AVDISCARD_NONE : discard nothing
AVDISCARD_DEFAULT : discard useless packets like 0 size packets in avi
AVDISCARD_NONREF : discard all non reference
AVDISCARD_BIDIR : discard all bidirectional frames
AVDISCARD_NONKEY : discard all frames except keyframes
AVDISCARD_ALL : discard all

For example, if you want VideoCapture to capture only keyframes, you can set the env var :
`OPENCV_FFMPEG_CAPTURE_OPTIONS=avdiscard;nonkey`

### DIFF
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -966,6 +966,29 @@ bool CvCapture_FFMPEG::open( const char* _filename )
         enc->thread_count = get_number_of_cpus();
 //#endif
 
+#if LIBAVCODEC_BUILD >= CALC_FFMPEG_VERSION(52, 123, 0)
+        AVDictionaryEntry* avdiscard_entry = av_dict_get(dict, "avdiscard", NULL, 0);
+
+        if (avdiscard_entry != 0) {
+            if(strcmp(avdiscard_entry->value, "all") == 0)
+                enc->skip_frame = AVDISCARD_ALL;
+            else if (strcmp(avdiscard_entry->value, "bidir") == 0)
+                enc->skip_frame = AVDISCARD_BIDIR;
+            else if (strcmp(avdiscard_entry->value, "default") == 0)
+                enc->skip_frame = AVDISCARD_DEFAULT;
+            else if (strcmp(avdiscard_entry->value, "none") == 0)
+                enc->skip_frame = AVDISCARD_NONE;
+#if LIBAVCODEC_BUILD >= CALC_FFMPEG_VERSION(54, 59, 100)
+            else if (strcmp(avdiscard_entry->value, "nonintra") == 0)
+                enc->skip_frame = AVDISCARD_NONINTRA;
+#endif
+            else if (strcmp(avdiscard_entry->value, "nonkey") == 0)
+                enc->skip_frame = AVDISCARD_NONKEY;
+            else if (strcmp(avdiscard_entry->value, "nonref") == 0)
+                enc->skip_frame = AVDISCARD_NONREF;
+        }
+#endif
+
 #if LIBAVFORMAT_BUILD < CALC_FFMPEG_VERSION(53, 2, 0)
 #define AVMEDIA_TYPE_VIDEO CODEC_TYPE_VIDEO
 #endif
--- a/modules/videoio/test/test_ffmpeg.cpp
+++ b/modules/videoio/test/test_ffmpeg.cpp
@@ -303,6 +303,52 @@ const videoio_container_params_t videoio_container_params[] =
 
 INSTANTIATE_TEST_CASE_P(/**/, videoio_container, testing::ValuesIn(videoio_container_params));
 
+#ifndef _WIN32  // details: https://github.com/opencv/opencv/pull/16190
+
+typedef tuple<string, string, int> videoio_skip_params_t;
+typedef testing::TestWithParam< videoio_skip_params_t > videoio_skip;
+
+TEST_P(videoio_skip, read)
+{
+#if CV_VERSION_MAJOR >= 4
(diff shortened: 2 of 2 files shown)


asmorkalov:
@pletessier Thanks for the PR. The changes looks good for me. Could you add simple test for the feature? The simplest solution I see is to use some existing video in opencv_extra, restrict FFmpeg provide key frames only with `AVDISCARD_NONKEY` option and check if `cv::VideoCaptrue` returns expected amount of frames. Also please, guard the test with FFMPEG specific macro to not break test for other backends.

pletessier:
Hi @asmorkalov, 
Thanks for taking this PR.
I wrote a test that works pretty well, but I don't know how to set the `OPENCV_FFMPEG_CAPTURE_OPTIONS` env var only for this single test. 

Setting it in my shell before to run the test is breaking some tests using FFMpeg. That was expected.
Setting it in source code with the `setenv` method is breaking tests also, but only after my new one.
I tried to unset the var with the same `setenv` method and a empty value, or using the `unsetenv` method. It doesn't change anything.

Is there another approach I could try ?

asmorkalov:
`OPENCV_FFMPEG_CAPTURE_OPTIONS` is read at https://github.com/opencv/opencv/blob/master/modules/videoio/src/cap_ffmpeg_impl.hpp#L911. The option is read every time when VideoCapture object is created and opened, so unset should resolve side effects. Please push your test code to discuss it.

pletessier:
@asmorkalov  Starting from a clean env, it works fine now !

pletessier:
@asmorkalov Windows tests are failing. Probably because of https://github.com/opencv/opencv/pull/9292#issuecomment-457217883.

I can't test on Windows. I am not even sure that the `_WIN32` variable is defined...

modules/videoio/test/test_ffmpeg.cpp
```diff
@@ -180,6 +180,49 @@ const videoio_container_params_t videoio_container_params[] =
...
+    #ifdef _WIN32
+        _putenv_s("OPENCV_FFMPEG_CAPTURE_OPTIONS", "");
+    #else
+        setenv("OPENCV_FFMPEG_CAPTURE_OPTIONS", "", 1);
+    #endif
+}
+
+const videoio_skip_params_t videoio_skip_params[] =
+{
+    videoio_skip_params_t(CAP_FFMPEG, "video/big_buck_bunny", "mp4", "avdiscard;nonkey", 11)
```
asmorkalov: The test is FFMPEG specific, so no need to push backend type as parameter. Also file name and extension are always used together, no need to pass them as separate parameters and then concatenate in test body.

asmorkalov:
Hello @pletessier , I found out the failure reason. OpenCV links FFMPEG using wrapper code to prevent license conflict. When the library is built on Unix and wrapper code is built together with the library as we do not distribute FFMPEG itself. On Windows the wrapper is built as dedicated library and is loaded on the go, if FFMPEG is present in the distribution. Wrapper is built and updated manually and CI just uses old version of the wrapper lib without your skip option.

asmorkalov:
:+1

asmorkalov:
@pletessier Please clean up test code and we can merge the path and update pre-built libs after it.

asmorkalov:
:+1:

alalek:
You need to skip Windows tests for now (until FFmpeg wrapper got new update) - we can't merge with red builders.

Add this guard for test code:
```
#ifndef _WIN32  // details: https://github.com/opencv/opencv/pull/16190
...
#endif
```

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -967,6 +967,25 @@ bool CvCapture_FFMPEG::open( const char* _filename )
...
+            else if (strcmp(avdiscard_entry->value, "default") == 0)
+                enc->skip_frame = AVDISCARD_DEFAULT;
+            else if (strcmp(avdiscard_entry->value, "none") == 0)
+                enc->skip_frame = AVDISCARD_NONE;
+            else if (strcmp(avdiscard_entry->value, "nonintra") == 0)
+                enc->skip_frame = AVDISCARD_NONINTRA;
+            else if (strcmp(avdiscard_entry->value, "nonkey") == 0)
+                enc->skip_frame = AVDISCARD_NONKEY;
+            else if (strcmp(avdiscard_entry->value, "nonref") == 0)
+                enc->skip_frame = AVDISCARD_NONREF;
```
alalek: What FFmpeg / libav versions support that? (at least compilation)

Need to properly guard code through "#if"s.

BTW, access to `OPENCV_FFMPEG_CAPTURE_OPTIONS` is guarded by
```
#if LIBAVFORMAT_BUILD >= CALC_FFMPEG_VERSION(52, 111, 0)
```

pletessier: @alalek All options (except `AVDISCARD_NONINTRA`) are in use since ffmpeg 0.7 (libavcodec 52.123.0)

First use of `AVDISCARD_NONINTRA` is in version 2.3 (libavcodec 55.69.100).

I added some checks on LIBAVCODEC_BUILD version in my last commit, and squashed previous ones.

alalek:
Rebased on 3.4 branch and "disabled" test:
- currently this is not way to detect LIBAV or old FFmpeg versions from the test code (to skip / workaround)
- test can be launched locally using "--gtest_also_run_disabled_tests" option.

PR #16218, opened 2019-12-21
Enable cuda4dnn on hardware without support for __half

JulienMaille:
ie. hardware with compute capability < 5.3

It compiles/link fine and I was able to run some inference on my Geforce 960!
Right now I limited support to CC 5.2+ but I suppose we can go lower, what do you think? 5.0? 4.0?

<cut/>

@YashasSamaga said
>Some checks in dnn.cpp to identify use of DNN_TARGET_CUDA_FP16 when half precision is disabled.

But compute capability has to be queried at runtime on the selected device in order to tell if FP16 is supported or not, correct?

```
force_builders=Custom,docs
buildworker:Custom=linux-4

test_opencl:Custom Win=OFF
```


YashasSamaga:
The `DNN_TARGET_CUDA_FP16` option must exist in the enumeration irrespective of whether that target is supported or not. This is required to maintain ABI compatibility. Hence, it's possible for a user who has built the module without FP16 support to set the target to `DNN_TARGET_CUDA_FP16`. Now since this target isn't supported on the device, there should be some error or a warning and then switch to `DNN_TARGET_CUDA`.

The capability would be known at compile-time because you decide at compile-time whether the half precision kernels would be instantiated or not.

The CUDA backend would create computation nodes for each supported layer in `initCUDABackend()`. This would invoke `initCUDA()` method on every layer which will create and return a node. This node is constructed using a helper template called `make_cuda_node` which automatically instantiates the correct node template based on the target.

The backend nodes take the form given below where `T` is `float` or `half` depending on the target.

```
template <class T>
class SomeComputeNode : public CUDABackendNode;
```

Here is what `make_cuda_node` does:

https://github.com/opencv/opencv/blob/89d3f95a8eea50acbfb4b8db380d5a4dc8a98173/modules/dnn/src/op_cuda.hpp#L169-L201

This will attempt to instantiate `half` backend nodes which in turn will attempt to invoke half precision kernels. Now since the half-precision kernels were not instantiated, this should lead to a load of linker errors.

I am confused how you have managed to built.

modules/dnn/CMakeLists.txt
```diff
@@ -95,8 +95,8 @@ if(OPENCV_DNN_CUDA AND HAVE_CUDA AND HAVE_CUBLAS AND HAVE_CUDNN)
   set(CC_LIST ${CUDA_ARCH_BIN})
   separate_arguments(CC_LIST)
   foreach(cc ${CC_LIST})
-    if(cc VERSION_LESS 5.3)
-      message(FATAL_ERROR "CUDA backend for DNN module requires CC 5.3 or higher. Please remove unsupported architectures from CUDA_ARCH_BIN option.")
+    if(cc VERSION_LESS 5.2)
+      message(FATAL_ERROR "CUDA backend for DNN module requires CC 5.2 or higher. Please remove unsupported architectures from CUDA_ARCH_BIN option.")
```
YashasSamaga: We can get rid of this check or drop it to 3.0.

neFAST:
> The capability would be known at compile-time because you decide at compile-time whether the half precision kernels would be instantiated or not.
> I am confused how you have managed to build. I'll check in a few hours.

What I did is make sure nvcc doesn't compile the __half related code when the compiling for a CC<5.3
gcc/msvc is still compile all the code handling __half.

I ran my CMAKE with `CUDA_ARCH_BIN="5.2 5.3 6.0 6.1 7.0 7.5"` so in the end I still support __half when it is available.

YashasSamaga:
Can you try with just `5.2` in `CUDA_ARCH_BIN`?

neFAST:
Sure, what do you expect?

neFAST:
> Can you try with just `5.2` in `CUDA_ARCH_BIN`?

I just did, it compiles and works

neFAST:
@asmorkalov I updated this PR to handle latest changes in .cu files. Let me know if and how I can help.

YashasSamaga:
There is a new transpose kernel in permute.cu which was added along with the copy kernel.

All the FP16 tests fail when there is no FP16 support:
```
[ RUN      ] Test_ONNX_nets.ResNet50v1/1, where GetParam() = CUDA/CUDA_FP16
unknown file: Failure
C++ exception with description "OpenCV(4.2.0-dev) /home/yashas/Desktop/gsoc/opencv/modules/dnn/src/cuda/execution.hpp:52: error: (-217:Gpu API call) invalid device function in function 'make_policy'
" thrown in the test body.
[  FAILED  ] Test_ONNX_nets.ResNet50v1/1, where GetParam() = CUDA/CUDA_FP16 (271 ms)
```

These should be disabled at runtime if possible as they are not really failures.

neFAST:
Do you know how to test compute capability at runtime?

YashasSamaga:
You might have to use [`cudaDeviceGetAttribute`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gb22e8256592b836df9a9cc36c9db7151) and obtain the attributes corresponding to `cudaDevAttrComputeCapabilityMajor` and `cudaDevAttrComputeCapabilityMinor`. You have to put them together `major.mino` to get the compute capability.

neFAST:
In that case (built for 7.5 but ran on 6.x) can't you have just in time compilation?

YashasSamaga:
CUDA provides two options:
- you can build binaries for various architectures
- you can generate PTX for various virtual architectures

The CUDA runtime generates the binary for the device from the PTX (generated by the compiler). This generation incurs a cost at runtime but the generated binary is cached. If the binaries are pre-built at compile-time, you can avoid this initialization cost.

Currently, these are the only two mechanisms supported by the CUDA backend. Runtime compilation is something I have planned for future (other ideas can be found [here](https://gist.github.com/YashasSamaga/3735b68381d9f6f7e4f2ccdad0f7d8e4)). It's quite complex and non-trival to implement JIT especially with the current template based kernels.

neFAST:
> All the FP16 tests fail when there is no FP16 support (used `CUDA_ARCH_BIN` as `5.2`). An example failure:
> These should be disabled at runtime if possible as they are not really failures.

Can you try again with latest commit? BTW I must be stupid but I can't find how to compile and run tests. I have replaced `-D_BUILD_TESTS=OFF` with `-D_BUILD_TESTS=ON` but I don't see them in the generated solution.

alalek:
```.diff
--D_BUILD_TESTS=ON
+-DBUILD_TESTS=ON
```

neFAST:
@alalek sorry this is just a typo in my comment, but not in my command line.

alalek:
Build `opencv_test_dnn` target.
Run `./bin/opencv_test_dnn` binary.
Also you should specify environment variables for tests:
- `OPENCV_TEST_DATA_PATH=<opencv_extra>/testdata` (clone "opencv_extra" repository)
- and optionally `OPENCV_DNN_TEST_DATA_PATH` (need to download 5+Gb)

YashasSamaga:
@neFast You need to have the ts module (which in turn requires videoio module) to run the tests.

JulienMaille:
I confirm the runtime check works and doesn't show FP16 target on my Geforce 960

JulienMaille:
@YashasSamaga Do you confirm you now pass the tests?
I've been looking at the code and there's a lot of stuff I don't understand like this:
https://github.com/opencv/opencv/blob/43a91f82fe634c9ba8abc89ef24decd4d508a0dd/modules/dnn/test/test_common.impl.hpp#L410

YashasSamaga:
@JulienMaille I don't own a device with CC 5.2 or below. I have to borrow it from someone. I think it's sufficient if you could upload the output of `opencv_test_dnn` and `opencv_perf_dnn`.

Those are tags which are used to mark the tests. They are specifically skip tags which cause the tests which which are marked with any of them to be skipped.

JulienMaille:
@YashasSamaga what I don't understand with the code I've linked is that it looks like if cuda is present then we set the flag to skip cuda tests.

YashasSamaga:
@JulienMaille Registering skip tags and applying skip tags are different. Applying a skip tag is what causes the test to be skipped.

YashasSamaga:
@JulienMaille Please rebase onto master. #16230 added new half-precision kernels. These kernels are used in FP16 target only.

JulienMaille:
Done, rebasing was enough. Do you think this can be merged soon?

YashasSamaga:
`fp16_to_fp32` and `fp32_to_fp16` needs checks. You can find them in `fp_conversions.cu`.

modules/dnn/src/dnn.cpp
```diff
@@ -135,6 +135,20 @@ class BackendRegistry
     }
 #endif
 
+    static inline bool deviceSupportsFp16() {
```
YashasSamaga: The `inline` isn't required. The function name is not very clear. This function only checks for FP16 support in CUDA capable devices but the name suggests that it checks for any device.

JulienMaille:
> `fp16_to_fp32` and `fp32_to_fp16` needs checks. You can find them in `fp_conversions.cu`.

Runtime checks? It compiles without any extra check with CC5.2

modules/dnn/src/dnn.cpp
JulienMaille: Something like `currentCudaDeviceSupportsFp16()`?

asmorkalov:
@JulienMaille CI bot reports build error:
```
/build/precommit_linux64/opencv/modules/dnn/src/dnn.cpp:138:2: error: #endif without #if
 #endif
```

asmorkalov:
@JulienMaille Please take a look on CI again. Build is still broken: https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/24320/steps/compile%20release/logs/stdio

JulienMaille:
@asmorkalov Forgot to remove an extra #endif, it is fixed now

JulienMaille:
@YashasSamaga probably a stupid question, but does cudnn module relies on cublas? (I'm really suprised by the size of the dll that have to be redistributed and trying to squeeze out anything useless)

YashasSamaga:
@JulienMaille cuDNN does not require cuBLAS but the CUDA backend requires cuBLAS for GEMM.

asmorkalov:
:+1:

modules/dnn/src/cuda/atomics.hpp
```diff
@@ -8,7 +8,7 @@
 #include <cuda_runtime.h>
 #include <cuda_fp16.h>
 
-#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 700
+#if !defined(__CUDA_ARCH__) || __CUDA_ARCH__ >= 700 || __CUDA_ARCH__ < 530
```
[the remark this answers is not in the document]
JulienMaille: Is it better now?

asmorkalov:
:+1: @alalek Please take a look and merge.

modules/dnn/CMakeLists.txt
```diff
@@ -95,8 +95,8 @@ if(OPENCV_DNN_CUDA AND HAVE_CUDA AND HAVE_CUBLAS AND HAVE_CUDNN)
   set(CC_LIST ${CUDA_ARCH_BIN})
   separate_arguments(CC_LIST)
   foreach(cc ${CC_LIST})
-    if(cc VERSION_LESS 5.3)
-      message(FATAL_ERROR "CUDA backend for DNN module requires CC 5.3 or higher. Please remove unsupported architectures from CUDA_ARCH_BIN option.")
+    if(cc VERSION_LESS 3.0)
+      message(FATAL_ERROR "CUDA backend for DNN module requires CC 3.0 or higher. Please remove unsupported architectures from CUDA_ARCH_BIN option.")
```
alalek: Please add this CMake hint to error message:
> "or disable OPENCV_DNN_CUDA=OFF"

JulienMaille: done

jiapei100:
Clearly, there is a one trivial mistake, I cannot PR for now. But, please refer to: https://github.com/opencv/opencv/compare/master...jiapei100:master#diff-c4c2826fcd60836eb106c34d1cdc6c74

JulienMaille:
@alalek jiape is right, the #endif was including a float operation
Correction here, shall I rebase and create a new PR?
https://github.com/JulienMaille/opencv/commit/3d3ed03fd404eff227afe9d17070ccc5f55b6835

alalek:
@JulienMaille Sure! Feel free to prepare new PR with fix (add `relates #16218` into description to add GitHub cross-link).

sl1pkn07:
please backport (or upport) to opencv 4.2 tag

greetings

for example, my hardware setup don't have cuda 5.3 feature

~~~
--     NVIDIA GPU arch:             30 35 37 50 52 60 61 70 75
~~~

(Nvidia 2060RTX)

ibrewster:
In the first post it was theorized that you might be able to go lower than architecture 5.2. To quote:

> but I suppose we can go lower, what do you think? 5.0? 4.0?

Was this ever investigated? I have a GPU with architecture 5.0, and would love to be able to try this out, if it might work.

PR #16223, opened 2019-12-23
Add human parsing demo

l-bat:
**Merge with extra:** opencv/opencv_extra#695
```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*

allow_multiple_commits=1
```
**Example:**
Input  / Output

### DIFF
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -250,7 +250,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         std::vector<size_t> pads_begin, pads_end;
         CV_DEPRECATED_EXTERNAL Size kernel, stride, pad;
         CV_DEPRECATED_EXTERNAL int pad_l, pad_t, pad_r, pad_b;
-        CV_DEPRECATED_EXTERNAL bool globalPooling; //!< Flag is true if at least one of the axes is global pooled.
+        bool globalPooling; //!< Flag is true if at least one of the axes is global pooled.
         std::vector<bool> isGlobalPooling;
         bool computeMaxIdx;
         String padMode;
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -251,7 +251,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         CV_DEPRECATED_EXTERNAL Size kernel, stride, pad;
         CV_DEPRECATED_EXTERNAL int pad_l, pad_t, pad_r, pad_b;
         bool globalPooling;
-        int global_axis;
+        std::vector<bool> isGlobalPooling;
         bool computeMaxIdx;
         String padMode;
         bool ceilMode;
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -251,6 +251,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         CV_DEPRECATED_EXTERNAL Size kernel, stride, pad;
         CV_DEPRECATED_EXTERNAL int pad_l, pad_t, pad_r, pad_b;
         bool globalPooling;
+        int global_axis;
         bool computeMaxIdx;
         String padMode;
         bool ceilMode;
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -250,7 +250,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         std::vector<size_t> pads_begin, pads_end;
         CV_DEPRECATED_EXTERNAL Size kernel, stride, pad;
         CV_DEPRECATED_EXTERNAL int pad_l, pad_t, pad_r, pad_b;
-        bool globalPooling;
+        CV_DEPRECATED_EXTERNAL bool globalPooling;
         std::vector<bool> isGlobalPooling;
         bool computeMaxIdx;
         String padMode;
--- a/modules/dnn/include/opencv2/dnn/all_layers.hpp
+++ b/modules/dnn/include/opencv2/dnn/all_layers.hpp
@@ -250,7 +250,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         std::vector<size_t> pads_begin, pads_end;
         CV_DEPRECATED_EXTERNAL Size kernel, stride, pad;
         CV_DEPRECATED_EXTERNAL int pad_l, pad_t, pad_r, pad_b;
-        CV_DEPRECATED_EXTERNAL bool globalPooling;
+        CV_DEPRECATED_EXTERNAL bool globalPooling; //!< Flag is true if at least one of the axes is global pooled.
         std::vector<bool> isGlobalPooling;
         bool computeMaxIdx;
         String padMode;
--- a/modules/dnn/src/layers/layers_common.cpp
+++ b/modules/dnn/src/layers/layers_common.cpp
@@ -144,14 +144,26 @@ void getStrideAndPadding(const LayerParams &params, std::vector<size_t>& pads_be
 }
 }
 
-void getPoolingKernelParams(const LayerParams &params, std::vector<size_t>& kernel, bool &globalPooling,
+void getPoolingKernelParams(const LayerParams &params, std::vector<size_t>& kernel, std::vector<bool>& globalPooling,
                             std::vector<size_t>& pads_begin, std::vector<size_t>& pads_end,
                             std::vector<size_t>& strides, cv::String &padMode)
 {
-    globalPooling = params.has("global_pooling") &&
-                    params.get<bool>("global_pooling");
+    bool is_global = params.get<bool>("global_pooling", false);
+    globalPooling = std::vector<bool>(3, is_global);
+    if (params.has("global_d"))
+    {
+        globalPooling[0] = params.get<bool>("global_d");
+    }
+    else if (params.has("global_h"))
+    {
+        globalPooling[1] = params.get<bool>("global_h");
+    }
+    else if (params.has("global_w"))
+    {
+        globalPooling[2] = params.get<bool>("global_w");
+    }
 
-    if (globalPooling)
+    if (is_global)
     {
         util::getStrideAndPadding(params, pads_begin, pads_end, strides, padMode);
         if(params.has("kernel_h") || params.has("kernel_w") || params.has("kernel_size"))
--- a/modules/dnn/src/layers/layers_common.cpp
+++ b/modules/dnn/src/layers/layers_common.cpp
@@ -149,13 +149,10 @@ void getPoolingKernelParams(const LayerParams &params, std::vector<size_t>& kern
                             std::vector<size_t>& strides, cv::String &padMode)
 {
     bool is_global = params.get<bool>("global_pooling", false);
-    globalPooling = std::vector<bool>(3, is_global);
-    if (params.has("global_pooling_d"))
-        globalPooling[0] = params.get<bool>("global_pooling_d");
-    if (params.has("global_pooling_h"))
-        globalPooling[1] = params.get<bool>("global_pooling_h");
-    if (params.has("global_pooling_w"))
-        globalPooling[2] = params.get<bool>("global_pooling_w");
+    globalPooling.resize(3);
+    globalPooling[0] = params.get<bool>("global_pooling_d", is_global);
+    globalPooling[1] = params.get<bool>("global_pooling_h", is_global);
+    globalPooling[2] = params.get<bool>("global_pooling_w", is_global);
 
     is_global = globalPooling[0] || globalPooling[1] || globalPooling[2];
     if (!is_global)
@@ -172,25 +169,18 @@ void getPoolingKernelParams(const LayerParams &params, std::vector<size_t>& kern
             CV_Error(cv::Error::StsBadArg, "In global_pooling mode, kernel_size (or kernel_h and kernel_w) cannot be specified");
         }
 
-        kernel.resize(3, 1);
+        kernel.resize(3);
         pads_begin.resize(3, 0);
         pads_end.resize(3, 0);
         strides.resize(3, 1);
-        if (params.has("kernel_d"))
-            kernel[0] = params.get<int>("kernel_d");
-        if (params.has("kernel_h"))
-            kernel[1] = params.get<int>("kernel_h");
-        if (params.has("kernel_w"))
-            kernel[2] = params.get<int>("kernel_w");
+        kernel[0] = params.get<int>("kernel_d", 1);
+        kernel[1] = params.get<int>("kernel_h", 1);
+        kernel[2] = params.get<int>("kernel_w", 1);
 
-        if (params.has("pad_t"))
-            pads_begin[1] = params.get<int>("pad_t");
-        if (params.has("pad_l"))
-            pads_begin[2] = params.get<int>("pad_l");
-        if (params.has("pad_b"))
-            pads_end[1] = params.get<int>("pad_b");
-        if (params.has("pad_r"))
-            pads_end[2] = params.get<int>("pad_r");
+        pads_begin[1] = params.get<int>("pad_t", 0);
+        pads_begin[2] = params.get<int>("pad_l", 0);
+        pads_end[1] = params.get<int>("pad_b", 0);
+        pads_end[2] = params.get<int>("pad_r", 0);
         if (params.has("pad_h")) {
             pads_begin[1] = params.get<int>("pad_h");
             pads_end[1] = params.get<int>("pad_h");
@@ -219,10 +209,8 @@ void getPoolingKernelParams(const LayerParams &params, std::vector<size_t>& kern
             }
         }
 
-        if (params.has("stride_h"))
-            strides[1] = params.get<int>("stride_h");
-        if (params.has("stride_w"))
-            strides[2] = params.get<int>("stride_w");
+        strides[1] = params.get<int>("stride_h", 1);
+        strides[2] = params.get<int>("stride_w", 1);
         if (params.has("stride")) {
             DictValue param = params.get("stride");
             for (int i = param.size() - 1, j = strides.size() - 1; i >= 0; i--, j--) {
--- a/modules/dnn/src/layers/layers_common.cpp
+++ b/modules/dnn/src/layers/layers_common.cpp
@@ -154,8 +154,7 @@ void getPoolingKernelParams(const LayerParams &params, std::vector<size_t>& kern
     globalPooling[1] = params.get<bool>("global_pooling_h", is_global);
     globalPooling[2] = params.get<bool>("global_pooling_w", is_global);
 
-    is_global = globalPooling[0] || globalPooling[1] || globalPooling[2];
-    if (is_global)
+    if (globalPooling[0] || globalPooling[1] || globalPooling[2])
     {
         util::getStrideAndPadding(params, pads_begin, pads_end, strides, padMode);
         if ((globalPooling[0] && params.has("kernel_d")) ||
--- a/modules/dnn/src/layers/layers_common.cpp
+++ b/modules/dnn/src/layers/layers_common.cpp
@@ -157,26 +157,85 @@ void getPoolingKernelParams(const LayerParams &params, std::vector<size_t>& kern
     if (params.has("global_pooling_w"))
         globalPooling[2] = params.get<bool>("global_pooling_w");
 
-    if (is_global)
-    {
-        util::getStrideAndPadding(params, pads_begin, pads_end, strides, padMode);
-        if(params.has("kernel_h") || params.has("kernel_w") || params.has("kernel_size"))
-        {
-            CV_Error(cv::Error::StsBadArg, "In global_pooling mode, kernel_size (or kernel_h and kernel_w) cannot be specified");
-        }
-        for (int i = 0; i < pads_begin.size(); i++) {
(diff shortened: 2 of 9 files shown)


modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -124,6 +124,7 @@ class PoolingLayerImpl CV_FINAL : public PoolingLayer
             CV_Error(Error::StsBadArg, "Cannot determine pooling type");
         setParamsFrom(params);
         ceilMode = params.get<bool>("ceil_mode", true);
+        global_axis = params.get<int>("global_axis", -1);
```
dkurt: I'd like to recommend to introduce `global_pooling_w` and `global_pooling_h` as there are for `kernel` and `stride`.

dkurt: Maybe just a vector like `std::vector<bool> isGlobalPooling`

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -122,8 +122,17 @@ class PoolingLayerImpl CV_FINAL : public PoolingLayer
         }
         else
             CV_Error(Error::StsBadArg, "Cannot determine pooling type");
+
         setParamsFrom(params);
         ceilMode = params.get<bool>("ceil_mode", true);
+        if (params.has("is_global_pooling"))
```
dkurt: There is already "global_pooling" flag so let's just use it in the same way as we use vector of kernel sizes or strides.

modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -1935,34 +1935,131 @@ void TFImporter::populateNet(Net dstNet)
...
+                layer_id[reshape_name] = id;
+                connect(layer_id, dstNet, parsePin(layer.input(i)), id, 0);
+            }
+
+            layerParams.set("axis", dim);
+            int id = dstNet.addLayer(name, "Concat", layerParams);
+            layer_id[name] = id;
+
+            for (int li = 0; li < num; li++) {
+                Pin inp = parsePin(reshape_names[li]);
```
dkurt: I think we can just keep corresponding input layer ids instead of names

modules/dnn/src/layers/layers_common.cpp
```diff
@@ -144,14 +144,26 @@ void getStrideAndPadding(const LayerParams &params, std::vector<size_t>& pads_be
...
-void getPoolingKernelParams(const LayerParams &params, std::vector<size_t>& kernel, bool &globalPooling,
+void getPoolingKernelParams(const LayerParams &params, std::vector<size_t>& kernel, std::vector<bool>& globalPooling,
                             std::vector<size_t>& pads_begin, std::vector<size_t>& pads_end,
                             std::vector<size_t>& strides, cv::String &padMode)
 {
-    globalPooling = params.has("global_pooling") &&
-                    params.get<bool>("global_pooling");
+    bool is_global = params.get<bool>("global_pooling", false);
+    globalPooling = std::vector<bool>(3, is_global);
+    if (params.has("global_d"))
```
dkurt: `global_pooling_d`? (The same for `global_pooling_w` and `global_pooling_h`)

modules/dnn/src/layers/layers_common.cpp
```diff
@@ -150,18 +150,12 @@ void getPoolingKernelParams(const LayerParams &params, std::vector<size_t>& kern
...
-    else if (params.has("global_w"))
-    {
-        globalPooling[2] = params.get<bool>("global_w");
-    }
+    if (params.has("global_pooling_d"))
+        globalPooling[0] = params.get<bool>("global_pooling_d");
+    else if (params.has("global_pooling_h"))
+        globalPooling[1] = params.get<bool>("global_pooling_h");
+    else if (params.has("global_pooling_w"))
+        globalPooling[2] = params.get<bool>("global_pooling_w");
```
dkurt: What if we have two of the flags simultaneously? I think we need to have just `if`s here.

l-bat: Thanks! 🙈

modules/dnn/src/layers/layers_common.cpp
```diff
@@ -144,14 +144,20 @@ void getStrideAndPadding(const LayerParams &params, std::vector<size_t>& pads_be
...
+    bool is_global = params.get<bool>("global_pooling", false);
+    globalPooling = std::vector<bool>(3, is_global);
+    if (params.has("global_pooling_d"))
+        globalPooling[0] = params.get<bool>("global_pooling_d");
+    if (params.has("global_pooling_h"))
+        globalPooling[1] = params.get<bool>("global_pooling_h");
+    if (params.has("global_pooling_w"))
+        globalPooling[2] = params.get<bool>("global_pooling_w");
+
+    if (is_global)
```
dkurt: Due global pooling across depth is not tested yet, to match previous behavior it's better to check 
```cpp
if (globalPooling[1] || globalPooling[2])
```

asmorkalov:
@dkurt Could you look at it again?

samples/dnn/human_parsing.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    input = preprocess(image_path)
+    input_h, input_w = input.shape[2:]
+    output = run_net(input, model_path, backend, target)
+    grayscale_out = postprocess(output, (input_w, input_h))
+    segmentation = decode_labels(grayscale_out)
+    return segmentation
+
+
+if __name__ == '__main__':
```
dkurt: Please move ArgumentParser under `f __name__ == '__main__':`

modules/dnn/src/layers/layers_common.cpp
```diff
@@ -144,33 +144,86 @@ void getStrideAndPadding(const LayerParams &params, std::vector<size_t>& pads_be
...
+        pads_end.resize(3, 0);
+        strides.resize(3, 1);
+        kernel[0] = params.get<int>("kernel_d", 1);
+        kernel[1] = params.get<int>("kernel_h", 1);
+        kernel[2] = params.get<int>("kernel_w", 1);
+
+        pads_begin[1] = params.get<int>("pad_t", 0);
+        pads_begin[2] = params.get<int>("pad_l", 0);
+        pads_end[1] = params.get<int>("pad_b", 0);
+        pads_end[2] = params.get<int>("pad_r", 0);
```
dkurt: Let's try once again to simplify it considering the following rules:
1) Paddings. User should specify either `pad` or `pad_w` && `pad_h` (getParameter) or `pad_l` && `pad_t` && `pad_r` && `pad_b` (current 3.4 works in this way.). So in case of global pooling by a specific axis, all the other padding will be specified anyway.
2) Strides similar as (1) but with `stride` or `stride_w` && `stride_h`
3) check only parameters which conflict with each other:
```
(!params.has("global_pooling") || !params.has("kernel_size")) &&
(!params.has("global_pooling_d") || !params.has("kernel_size_d")) &&
(!params.has("global_pooling_h") || !params.has("kernel_size_h")) &&
(!params.has("global_pooling_w") || !params.has("kernel_size_w"))
```

modules/dnn/src/layers/layers_common.cpp
```diff
@@ -144,34 +144,46 @@ void getStrideAndPadding(const LayerParams &params, std::vector<size_t>& pads_be
...
+        for (int i = 0, j = globalPooling.size() - strides.size(); i < strides.size(); i++, j++) {
+            if (strides[i] != 1 && globalPooling[j])
                 CV_Error(cv::Error::StsBadArg, "In global_pooling mode, strides must be = 1");
         }
     }
-    else
-    {
-        util::getKernelSize(params, kernel);
-        util::getStrideAndPadding(params, pads_begin, pads_end, strides, padMode, kernel.size());
-    }
```
dkurt: Let's keep `if(is_global) { ... } else { ... }`

modules/dnn/src/layers/layers_common.cpp
```diff
@@ -144,26 +144,38 @@ void getStrideAndPadding(const LayerParams &params, std::vector<size_t>& pads_be
...
+            params.has("kernel_size")) {
             CV_Error(cv::Error::StsBadArg, "In global_pooling mode, kernel_size (or kernel_h and kernel_w) cannot be specified");
         }
-        for (int i = 0; i < pads_begin.size(); i++) {
-            if (pads_begin[i] != 0 || pads_end[i] != 0)
+
+        kernel.resize(3);
+        kernel[0] = params.get<int>("kernel_d", 1);
+        kernel[1] = params.get<int>("kernel_h", 1);
+        kernel[2] = params.get<int>("kernel_w", 1);
```
dkurt: Do we still need this here?

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -95,7 +96,8 @@ class PoolingLayerImpl CV_FINAL : public PoolingLayer
             else
                 CV_Error(Error::StsBadArg, "Unknown pooling type \"" + pool + "\"");
 
-            getPoolingKernelParams(params, kernel_size, globalPooling, pads_begin, pads_end, strides, padMode);
+            getPoolingKernelParams(params, kernel_size, isGlobalPooling, pads_begin, pads_end, strides, padMode);
+            globalPooling = std::accumulate(isGlobalPooling.begin(), isGlobalPooling.end(), 0) == 3;
```
dkurt: let's compare with `isGlobalPooling.size()` instead of hardcoded value

dkurt: Shall we check that at least one of the flags is true instead? As we do above: `is_global = globalPooling[0] || globalPooling[1] || globalPooling[2];`

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -146,12 +148,21 @@ class PoolingLayerImpl CV_FINAL : public PoolingLayer
...
             out.push_back(outputs[0].size[i]);
         }
-        if (globalPooling) {
-            kernel = Size(inp[1], inp[0]);
-            kernel_size = std::vector<size_t>(inp.begin(), inp.end());
+
+        if (kernel_size.size() > inp.size()) {
+            kernel_size.erase(kernel_size.begin());
         }
+        kernel_size.resize(out.size());
```
dkurt: I think we do not need this due in the next loop we can iterate over input shape and change `kernel_size`:

```cpp
        CV_Assert(isGlobalPooling.size() == kernel_size.size());
        std::vector<size_t> finalKernel;
        for (int i = 0; i < inp.size(); i++)
        {
            int idx = isGlobalPooling.size() - inp.size() + i;
            finalKernel.push_back(isGlobalPooling[idx] ? inp[i] : kernel_size[idx]);
        }
        kernel_size = finalKernel;
```

l-bat: Always  `isGlobalPooling.size() = 3`. If pooling type is `ROI` then kernel_size is empty.

modules/dnn/include/opencv2/dnn/all_layers.hpp
```diff
@@ -250,7 +250,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         std::vector<size_t> pads_begin, pads_end;
         CV_DEPRECATED_EXTERNAL Size kernel, stride, pad;
         CV_DEPRECATED_EXTERNAL int pad_l, pad_t, pad_r, pad_b;
-        bool globalPooling;
+        CV_DEPRECATED_EXTERNAL bool globalPooling; //!< Flag is true if at least one of the axes is global pooled.
```
dkurt: Due we are going to use this flag as well, let's remove `CV_DEPRECATED_EXTERNAL`

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -995,26 +1004,29 @@ virtual Ptr<BackendNode> initNgraph(const std::vector<Ptr<BackendWrapper> >& inp
...
-        if (globalPooling)
-        {
-            outShape.push_back(1);
-            outShape.push_back(1);
+        std::vector<size_t> local_kernel = kernel_size.empty() ?
+                                           std::vector<size_t>(inpShape.begin(), inpShape.end()) : kernel_size;
+
+        for (int i = 0, j = local_kernel.size() - inpShape.size(); i < inpShape.size(); i++, j++) {
+            if (isGlobalPooling[j])
+                local_kernel[j] = inpShape[i];
```
dkurt: Can we move it under `if (globalPooling)` so that will guarantee that kernel is not empty?

modules/dnn/src/tensorflow/tf_importer.cpp
```diff
@@ -1932,37 +1932,151 @@ void TFImporter::populateNet(Net dstNet)
...
+                layer_id[reshape_name] = id;
+                connect(layer_id, dstNet, parsePin(layer.input(i)), id, 0);
+            }
+
+            layerParams.set("axis", dim);
+            int id = dstNet.addLayer(name, "Concat", layerParams);
+            layer_id[name] = id;
+
+            for (int li = 0; li < num; li++)
+                connect(layer_id, dstNet, Pin(reshape_names[li]), id, li);
```
dkurt: Maybe it's better to use `dstNet`'s connect directly?

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -146,12 +148,19 @@ class PoolingLayerImpl CV_FINAL : public PoolingLayer
...
         }
+
         if (globalPooling) {
-            kernel = Size(inp[1], inp[0]);
-            kernel_size = std::vector<size_t>(inp.begin(), inp.end());
-        }
+            std::vector<size_t> finalKernel;
+            for (int i = 0; i < inp.size(); i++) {
+                int idx = isGlobalPooling.size() - inp.size() + i;
+                finalKernel.push_back(isGlobalPooling[idx] ? inp[i] : kernel_size[idx]);
```
dkurt: Let's add assert before the loop that
```cpp
isGlobalPooling.size() == kernel_size.size()
```

modules/dnn/src/layers/pooling_layer.cpp
dkurt: Let's reuse the loop above:

```cpp
            std::vector<size_t> finalKernel;
            if (globalPooling)
              for (int i = 0; i < inp.size(); i++) {
                  int idx = isGlobalPooling.size() - inp.size() + i;
                     finalKernel.push_back(isGlobalPooling[idx] ? inp[i] : kernel_size[idx]);
              }
            else
              finalKernel = kernel_size;
```

so in the code below we can use it without offsets:
```cpp
for (int i = 0; i < finalKernel.size(); i++) {
    float dst = (float)(inpShape[i] + pads_begin[i] + pads_end[i] - finalKernel[i]) / strides[i];
```

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -995,26 +1004,29 @@ virtual Ptr<BackendNode> initNgraph(const std::vector<Ptr<BackendWrapper> >& inp
...
-                float dst = (float)(inpShape[i] + pads_begin[i] + pads_end[i] - kernel_size[i]) / strides[i];
+            for (int i = 0, j = local_kernel.size() - inpShape.size(); i < inpShape.size(); i++, j++) {
+                float dst = (float)(inpShape[i] + pads_begin[i] + pads_end[i] - local_kernel[j]) / strides[i];
                 outShape.push_back(1 + (ceilMode ? ceil(dst) : floor(dst)));
             }
 
             // If we have padding, ensure that the last pooling starts strictly
             // inside the image (instead of at the padding); otherwise clip the last.
-            for (int i = 0; i < pads_end.size(); i++) {
+            for (int i = 0, j = local_kernel.size() - inpShape.size(); i < inpShape.size(); i++, j++) {
```
dkurt: please check that we still need this change

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -1037,6 +1049,7 @@ virtual Ptr<BackendNode> initNgraph(const std::vector<Ptr<BackendWrapper> >& inp
             outShape[0] = inputs[1][0];  // Number of proposals;
             outShape[1] = psRoiOutChannels;
         }
+
```
dkurt: please remove extra newline here and in one more place before

modules/dnn/src/layers/slice_layer.cpp
```diff
@@ -114,7 +114,8 @@ class SliceLayerImpl : public SliceLayer
     virtual bool supportBackend(int backendId) CV_OVERRIDE
     {
         return backendId == DNN_BACKEND_OPENCV ||
-               ((backendId == DNN_BACKEND_INFERENCE_ENGINE_NN_BUILDER_2019 || backendId == DNN_BACKEND_INFERENCE_ENGINE_NGRAPH) &&
+               (backendId == DNN_BACKEND_INFERENCE_ENGINE_NGRAPH && sliceRanges.size() == 1) ||
```
dkurt: Did we have a test with `sliceRanges.size() > 1` before?

l-bat: Yes, Test_Caffe_layers.Reshape_Split_Slice: sliceRanges.size() = 3

dkurt: But why we need this condition in this PR?

l-bat: We enable case with ` sliceRanges.size() =1` and with any` sliceRanges[0].size()`  (not just `sliceRanges[0].size() == 4`)

dkurt: I mean this PR introduces limited usage of Slice layer with nGraph because of new mode of Mean layer?

dkurt: Oh I got it (just need to open an entire file to see). Thanks!

modules/dnn/src/layers/pooling_layer.cpp
```diff
@@ -995,26 +1003,30 @@ virtual Ptr<BackendNode> initNgraph(const std::vector<Ptr<BackendWrapper> >& inp
...
-                float dst = (float)(inpShape[i] + pads_begin[i] + pads_end[i] - kernel_size[i]) / strides[i];
+            for (int i = 0; i < inpShape.size(); i++) {
+                float dst = (float)(inpShape[i] + pads_begin[i] + pads_end[i] - local_kernel[i]) / strides[i];
                 outShape.push_back(1 + (ceilMode ? ceil(dst) : floor(dst)));
             }
 
             // If we have padding, ensure that the last pooling starts strictly
             // inside the image (instead of at the padding); otherwise clip the last.
-            for (int i = 0; i < pads_end.size(); i++) {
+            for (int i = 0; i < inpShape.size(); i++) {
```
dkurt: Why not `pads_end.size()`?

samples/dnn/human_parsing.py
```diff
@@ -0,0 +1,173 @@
...
+    parser = argparse.ArgumentParser(description='Use this script to run human parsing using JPPNet',
+                                     formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('--input', '-i', help='Path to input image.')
+    parser.add_argument('--model', '-m', required=True, help='Path to pb model
+                        (https://drive.google.com/open?id=1XHvo111Gj1ZGoNUJt4Y4OsShrt_eUT34).')
```
dkurt: Please check that script still running. It seems to me that this string is invalid.

You may also replace the link to this one: https://www.dropbox.com/s/qag9vzambhhkvxr/lip_jppnet_384.pb?dl=0

samples/dnn/human_parsing.py
```diff
@@ -0,0 +1,172 @@
...
+
+    output = parse_human(args.input, args.model, args.backend, args.target)
+    winName = 'Deep learning human parsing in OpenCV'
+    cv.namedWindow(winName, cv.WINDOW_AUTOSIZE)
+    cv.imshow(winName, output)
+    cv.waitKey()
+
+
+# To get original .meta pre-trained model download https://drive.google.com/file/d/1BFVXgeln-bek8TCbRjN6utPAgRE0LJZg/view
+# For correct convert .meta to .pb model download original repository https://github.com/Engineering-Course/LIP_JPPNet
```
alalek: Please move this information in the beginning of the file and reformat it as "doc" string:
```
#!/usr/bin/env python
'''
... doc string
        goes
            here ...
'''
```

samples/dnn/human_parsing.py
```diff
@@ -0,0 +1,172 @@
+import argparse
+import cv2 as cv
+import numpy as np
```
alalek: Lets `import numpy` first

modules/dnn/include/opencv2/dnn/all_layers.hpp
```diff
@@ -250,7 +250,8 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
         std::vector<size_t> pads_begin, pads_end;
         CV_DEPRECATED_EXTERNAL Size kernel, stride, pad;
         CV_DEPRECATED_EXTERNAL int pad_l, pad_t, pad_r, pad_b;
-        bool globalPooling;
+        bool globalPooling; //!< Flag is true if at least one of the axes is global pooled.
+        std::vector<bool> isGlobalPooling;
```
alalek: Please increase version for DNN namespace `CV__DNN_EXPERIMENTAL_NS`.

samples/dnn/human_parsing.py
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env python
+'''
+    You can download the converted pb model from https://www.dropbox.com/s/qag9vzambhhkvxr/lip_jppnet_384.pb?dl=0
+    or convert the model yourself.
+
+    Follow these steps if you want to convert the original model yourself:
```
alalek: Why do we need indentation here?

PR #16225, opened 2019-12-23
dnn: small fix in pose estimation model

### LINKED ISSUE #16204: dnn: human keypoint onnx model seems broken
- OpenCV => 4.1.2
- Operating System / Platform => Windows 64 Bit / ubuntu18
- Compiler => mingw64 8.3 / python

the [lightweight-human-pose model in opencv_extra](https://github.com/opencv/opencv_extra/blob/4fdb88f83fe39f969e9945f531045a5a0f51b827/testdata/dnn/download_models.py#L783-L787)  looks broken, the last (output) layer:

Clipboard01

looks like it only contains the 38 paf maps , but not the required heatmaps. (so the KeypointModel code goes haywire, and the results are all bogus)

there should be 19 heatmaps, and optionally 38 paf maps, like below:

Clipboard02

the [newly added opencv Keypoints model](https://github.com/opencv/opencv/pull/15189/files#diff-6de938636906f03a3c475c556f5c8878R166) tries to loop over all 38 of them, and results look quite bad ;( also those 14 (!) of 38 that made it over the threshold were  wrongly saved to the test expectation npy.

there is also an issue here:
https://github.com/opencv/opencv/pull/15189/files#diff-6de938636906f03a3c475c556f5c8878R180
keypoints with low confidence must not be skipped, else we change the expected order.

so, imho, we should

* upload a new,corrected model, [based on daniil osokins script](https://github.com/Daniil-Osokin/lightweight-human-pose-estimation.pytorch/blob/master/scripts/convert_to_onnx.py#L13)  (this also produced the correct output above)
* correct the dnn model code: use `net.forward("stage_1_output_1_heatmaps")` and don't skip low prb kps
* record new test keypoints and upload to extra
* update the test code, 

see also: https://github.com/Daniil-Osokin/lightweight-human-pose-estimation.pytorch/issues/58

[noisy colab](https://colab.research.google.com/drive/186N1ZF5UrCCD2lKMWjQYAAK 35DV9Geje)

ps: imho it's not a problem with the dnn importer, the heatmaps are simply not in the current onnx file (same dot graph from onnx.tools)

berak:
resolves #16204 , if merged with https://github.com/opencv/opencv_extra/pull/692

insert (-1,-1) keypoints in case of low probability, instead of skipping them.

### DIFF
--- a/modules/dnn/test/test_model.cpp
+++ b/modules/dnn/test/test_model.cpp
@@ -273,9 +273,15 @@ TEST_P(Test_Model, Keypoints_pose)
 #endif
 
     Mat inp = imread(_tf("pose.png"));
-    std::string weights = _tf("onnx/models/lightweight_pose_estimation.onnx", false);
-    Mat exp = blobFromNPY(_tf("keypoints_exp.npy"));
-
+    std::string weights = _tf("onnx/models/lightweight_pose_estimation_201912.onnx", false);
+    float kpdata[] = {
+        237.65625f, 78.25f, 237.65625f, 136.9375f,
+        190.125f, 136.9375f, 142.59375f, 195.625f, 79.21875f, 176.0625f, 285.1875f, 117.375f,
+        348.5625f, 195.625f, 396.09375f, 176.0625f, 205.96875f, 313.0f, 205.96875f, 430.375f,
+        205.96875f, 528.1875f, 269.34375f, 293.4375f, 253.5f, 430.375f, 237.65625f, 528.1875f,
+        221.8125f, 58.6875f, 253.5f, 58.6875f, 205.96875f, 78.25f, 253.5f, 58.6875f
+    };
+    Mat exp(18, 2, CV_32FC1, kpdata);
 
     Size size{256, 256};
     float norm = 1e-4;
--- a/modules/dnn/src/model.cpp
+++ b/modules/dnn/src/model.cpp
@@ -177,8 +177,8 @@ std::vector<Point2f> KeypointsModel::estimate(InputArray frame, float thresh)
                 p = maxLoc;
                 p.x *= (float)frameWidth / width;
                 p.y *= (float)frameHeight / height;
-                points.push_back(p);
             }
+            points.push_back(p);
         }
     }
     // Otherwise the output is a vector of keypoints and we can just return it


asmorkalov:
@dkurt Could you look at the patch?

asmorkalov:
@berak @dkurt Do you have any updates?

berak:
https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/24507/steps/test_dnn/logs/stdio

> [ RUN      ] Test_Model.Keypoints_pose/0, where GetParam() = OCV/CPU
> exp [2 x 18]
> out [2 x 37]
> unknown file: Failure
> C++ exception with description "OpenCV(4.2.0-dev) /build/precommit_linux64/opencv/modules/ts/src/ts_func.cpp:1387: error: (-215:Assertion failed) src1.size == src2.size in function 'norm'
> " thrown in the test body.
> 

out has the wrong shape, it seems to use the old / wrong onnx model (which still uses PAF not heatmaps)?

at which time is download_models.py run ?

it's pulling in the correct changes from opencv_extra  (at least i think so):
https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/24507/steps/Merge%20opencv%20with%20test%20branch/logs/stdio

modules/dnn/test/test_model.cpp
```diff
@@ -273,7 +273,7 @@ TEST_P(Test_Model, Keypoints_pose)
 #endif
 
     Mat inp = imread(_tf("pose.png"));
-    std::string weights = _tf("onnx/models/lightweight_pose_estimation.onnx", false);
+    std::string weights = _tf("lightweight_pose_estimation_201912.onnx", false);
```
alalek: Keep `onnx/models`.

Model is not updated yet through `download_models.py` - this not fully automated process.
ETA 10-20min - need to scan SHA of several Gb (on all build machines).

berak:
thanks for all your help, and apologies for causing so much trouble.

which is the correct path now ?

        std::string weights = _tf("onnx/models/lightweight_pose_estimation.onnx", false);
or: 
        std::string weights = _tf("lightweight_pose_estimation.onnx");

berak:
thanks for all your help !

alalek:
Looks like there is no fix in implementation - this patch replaces test model only.
Is old model lost completely? In meaning should we test both versions (do we really want that)?

/cc @dkurt

modules/dnn/test/test_model.cpp
```diff
@@ -273,7 +273,7 @@ TEST_P(Test_Model, Keypoints_pose)
 #endif
 
     Mat inp = imread(_tf("pose.png"));
-    std::string weights = _tf("onnx/models/lightweight_pose_estimation.onnx", false);
+    std::string weights = _tf("onnx/models/lightweight_pose_estimation_201912.onnx", false);
```
dkurt: So may I ask you to highlight once again the difference comparing previous version?

berak:
* changes in extra:
old model (wrong output, PAF instead of heatmaps)

changed to correct output (19 heatmaps) :
heatmap
using [daniil osokins script here](https://github.com/Daniil-Osokin/lightweight-human-pose-estimation.pytorch/blob/master/scripts/convert_to_onnx.py#L13) 
also the resp. test data had to be changed to reflect the changed model.

* changes here:



landmarks are ordered. we cannot simply skip keypoints with low probability.

----
[another colab example demontrating the corrected model](https://colab.research.google.com/drive/1hGrvL2HeZDfsYNE46Bbib6Lkl2r-NpNA)

dkurt:
@berak, what about including PAFs into the model? I mean to have 19 heatmaps + 38 PAFs channels. I believe that one day we could improve [openpose samples](https://github.com/opencv/opencv/blob/master/samples/dnn/openpose.py) to support them as well. Now we slice first 19 heatmaps:

https://github.com/opencv/opencv/blob/32aba5e64ba40b372f02b5fab18c9632f763bb75/samples/dnn/openpose.py#L87-L92

berak:
@dkurt , concatenating sounds like a good idea, i'll try that. 

where would the nPoints information come from here ? 
https://github.com/opencv/opencv/blob/2ced568d3449c0878b1d25b03e7103ab18ce9d0f/modules/dnn/src/model.cpp#L156

dkurt:
@berak, as for me we merged `KeypointsModel` to early without enough samples. Some of landmarks model produce coordinates directly but not heatmaps. So let's focus on the samples and if they work - that's fine. I just wanted to recommend to include PAFs so if we will support them as well - we won't need to upload a model again.

berak:
so, i [modified the script](https://github.com/berak/lightweight-human-pose-estimation.pytorch/blob/patch-1/models/witth_mobilenet.py#L124) and exported a new [model with concatenated heat + paf maps](https://drive.google.com/open?id=1SZiz1AvZ85R6Q4_tGqFQiVO9IhMpcqgk), so it has same output structure as in the caffemodel.

[output from here](https://www.learnopencv.com/multi-person-pose-estimation-in-opencv-using-openpose):
group

but the problem with the KeypointsModel remains:  once are more maps in the output, than needed here, we need to know "where to stop", the number of actual keypoints must be hardcoded (and somehow "cleverly deduced" [from the model file](https://github.com/opencv/opencv/blob/27edba495a38cff10ecd2171cb9c014436b224de/samples/dnn/openpose.cpp#L87-L91)  -- a bummer, imho.)

maybe we should close this for now (as it seems slightly too early), and restart the discussion, once @dvd42 finds the time to write actual tutorial for this ?

asmorkalov:
@dkurt @berak do you have conclusion on the patch?

modules/dnn/test/test_model.cpp
```diff
@@ -299,7 +305,43 @@ TEST_P(Test_Model, Keypoints_face)
...
+        0.4858782f ,  0.8166637f ,        0.64733803f,  0.8720045f ,
+        0.75894463f,  0.96210575f,        0.6322932f ,  1.1599172f ,
+        0.5222149f ,  1.282458f  ,        0.3378732f ,  1.2986604f ,
+        0.19809961f,  1.2634348f ,        0.02682695f,  1.1636689f ,
+       -0.11073969f,  0.9403825f ,        0.22035715f,  0.9148466f ,
+        0.38391942f,  0.9214617f ,        0.4866761f ,  0.93515944f,
+        0.7235508f ,  0.9525098f ,        0.49401033f,  1.1250464f ,
+        0.3664117f ,  1.1356674f ,        0.18867175f,  1.0959036f
+    };
+    Mat exp(68, 2, CV_32FC1, kpdata);
```
dkurt: Let's keep `facial_keypoints_exp.npy`

berak: ok, i'll do that.

should i change to the new model, containing concatenated heat&paf maps, and add some code to the model like:

```
        // slice output, if it's the lightweight_human_pose one, or the openpose model with original prototxt
        if (nPoints == 57)
        {
            nPoints = 19;
        }
```

or keep it as is (model w. 19 heatmaps, code unchanged) ?

dkurt: I think it's better to keep as is for now. It's easier for user to process output.

berak: ok, done, thanks for the quick reply.

PR #16226, opened 2019-12-23
cuda4dnn(permute): improve reduction logic and add transpose kernel

YashasSamaga:
### This pullrequest changes
- upgrades the reduction logic to:
  - eliminate singleton axes
  - merge adjacent axes which do not change relative order
- do raw copy if the rank is reduced to one or if the order does not mandate permutation

It adds a new very efficient (on par with raw memory copy) transpose kernel which can be used when the reduction logic leads to rank two tensors.

In almost all cases, the permutation operation reduces to a transpose. Transpose is several times faster than permute kernel.

<cut/>

```
force_builders=Custom,docs
buildworker:Custom=linux-4

test_opencl:Custom Win=OFF
```

### DIFF
--- a/modules/dnn/src/cuda/fill_copy.cu
+++ b/modules/dnn/src/cuda/fill_copy.cu
@@ -21,12 +21,23 @@ namespace cv { namespace dnn { namespace cuda4dnn { namespace kernels {
         template <class T, std::size_t N>
         __global__ void fill_vec(Span<T> output, T value) {
             using vector_type = get_vector_type_t<T, N>;
-
             auto output_vPtr = vector_type::get_pointer(output.data());
             for (auto i : grid_stride_range(output.size() / vector_type::size())) {
                 vector_type vec;
                 for (int j = 0; j < vector_type::size(); j++)
                     vec.data[j] = value;
+                v_store(output_vPtr[i], vec);
+            }
+        }
+
+        template <class T, std::size_t N>
+        __global__ void  copy_vec(Span<T> output, View<T> input) {
+            using vector_type = get_vector_type_t<T, N>;
+            auto input_vPtr = vector_type::get_pointer(input.data());
+            auto output_vPtr = vector_type::get_pointer(output.data());
+            for (auto i : grid_stride_range(output.size() / vector_type::size())) {
+                vector_type vec;
+                v_load(vec, input_vPtr[i]);
                 v_store(output_vPtr[i], vec);
             }
         }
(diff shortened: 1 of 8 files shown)


YashasSamaga:
Name                                | Description                             | Status
----------------------------------- | --------------------------------------- | -------------------
smarter kernel code | get rid of massive code duplication in CUDA files using templates, functors, etc. |
shared memory API | <ul><li>block_load, block_store</li><li>fast dynamic array indexing (much faster than local memory which the compiler prefers); abstract away the low-level details by some class `dynamic_shared_memory_array`</li></ul> |
L1 LDG API | <ul><li>currently requires `__ldg` intrinsic but this can be abstracted away with something nicer</li><li>`__ldg` can load constant data into L1 cache which is otherwise unused (normal global loads are cached in L2)</li></ul>  |
enable 128-bit loads on half | <ul><li>very easy if `constexpr if` was there</li><li>need to add 
enable 256-bit loads on float | same as above |
enable 256-bit loads on half | same as above |
vector<half, 8> type and invoke vec8 kernels for half precision ONLY</li></ul> |
fuse BN, pooling and activation | requires dumping cuDNN pooling kernels and writing new kernels | 
no fallback message for skipped layers | |
implement data layer (input layer) | it's faster to perform normalization/standardization on GPU than in `blobFromImage` (this can potentially be fused with other layers) |
fuse backwards for some layers | conv can fuse with a power layer which appears before it? |
inner production fusion | <ul><li>fuse bias and activation</li><li>look into NVIDIA/CUTLASS library</li></ul> |
eltwise fusion | fuse activation with eltwise |
forwardAsync (not a priority) | similar to InferenceEngine API |
GpuMat as input and output | GpuMat doesn't support arbitrary dimensions which is a problem |
reorder relu to come after max_pool | some models perform ReLU before max_pool while performing ReLU after max_pool would give the same results with better performance  |
inplace redundant permute | <ul><li>many models have permute layers which can be reduced to a reshape</li><li>update getMemoryShapes to identify these cases and return true (for inplace and eventually skip the permute entirely)</li></ul> |
avoid redundant transfers | ReshapeOp copies data if the input ptr and output ptr are not same. To check if the pointers are same, it uses the `getView`/`getSpan` methods which wrongly mutates the dirty state and causes spurious transfers |
skip NOP layers based on ReshapeOp  | <ul><li>required to enable more fusion</li><li>some fusions depend on reshape such as ChannelsPReLU</li></ul>
overlap data transfers with computation | 40% speedup in some models like YOLO |
record graph and reuse | <ul><li>records forward pass using CUDA graph API and uses it for future forward passes</li><li>makes dumping operations to GPU extremely fast (will save a bit of CPU overall)</li></ul> | 
kernel tuning | select best kernels by profiling at run-time |
NVRTC integration | compile kernels from source at runtime (better performance as some variables can be replaced with constants which in turn could enable mor compiler optimizations  |
use BackendNode for fallbacks | allows fallbacks to be integrated into the CUDA stream pipeline |
fuse conv + bias + eltwise + relu using cudnn | faster as it's done with conv contrary to the current fusion schemes where only bias and activations are fused
profile and optimize half-precision kernels | half-precision kernels currently fully rely on compiler to make the optimizations but from the generated PTX, we can see that the compiler does less than 60% of the possible optimizations; rest have to be done manually
compute prior box on CPU on init | precompute priors and perform transfer in forward pass |
layer options for CUDA backend |<ul><li>allows end-user to fine-tune</li><li>for example, force CPU fallback if inputs and outputs are provided and consumed by CPU layers and there is no benefit from GPU (like slice layer in Inception v2 Mask RCNN)</li></ul> 
mixed-precision options | <ul><li>allows user to tweak accuracy and performance</li><li>fairly easy to implement in the CUDA backend if the DNN backbone code gets some base code for handling mixed-precision</li></ul> |
quantization | |
UFF importer? | |
TensorRT integration | |

PR #16236, opened 2019-12-25
core: fix simd emulator code

alalek:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/2403

Usage:

```
cmake -DOPENCV_EXTRA_FLAGS="-DCV_FORCE_SIMD128_CPP=1" ...
```

<cut/>

```
force_builders=Custom
buildworker:Custom=linux-1
build_image:Custom=simd-emulator

#test_timeout:Linux x64 Debug=1200
#test_maxtime:Linux x64 Debug=14400
```

### DIFF
--- a/modules/core/include/opencv2/core/hal/intrin_cpp.hpp
+++ b/modules/core/include/opencv2/core/hal/intrin_cpp.hpp
@@ -50,6 +50,14 @@
 #include <algorithm>
 #include "opencv2/core/saturate.hpp"
 
+//! @cond IGNORED
+#define CV_SIMD128_CPP 1
+#if defined(CV_FORCE_SIMD128_CPP) || defined(CV_DOXYGEN)
+#define CV_SIMD128 1
+#define CV_SIMD128_64F 1
+#endif
+//! @endcond
+
 namespace cv
 {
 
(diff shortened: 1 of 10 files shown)


modules/core/include/opencv2/core/hal/intrin_cpp.hpp
```diff
@@ -1659,22 +1773,19 @@ Pointer __should__ be aligned by 16-byte boundary. */
 template<typename _Tp, int n>
 inline void v_store_aligned(_Tp* ptr, const v_reg<_Tp, n>& a)
 {
-    for( int i = 0; i < n; i++ )
-        ptr[i] = a.s[i];
+    v_store(ptr, a);
```
terfendail: May be it make sense to add alignment check here?

modules/core/src/arithm.simd.hpp
```diff
@@ -1593,7 +1593,7 @@ struct op_div_f
 {
     static inline Tvec r(const Tvec& a, const Tvec& b)
     {
-        const Tvec v_zero = Tvec();
+        const Tvec v_zero = vx_setall<typename Tvec::lane_type>(0);
```
terfendail: IMO it make sense to add vx_setzero template as well and use it here instead

alalek: Will be added as a separate PR

modules/core/src/mathfuncs_core.simd.hpp
```diff
@@ -725,7 +725,7 @@ void log32f( const float *_x, float *y, int n )
 
         yf0 = v_fma(v_cvt_f32(yi0), vln2, yf0);
 
-        v_float32 delta = v_reinterpret_as_f32(h0 == vx_setall_s32(510)) & vshift;
+        v_float32 delta = v_select(v_reinterpret_as_f32(h0 == vx_setall_s32(510)), vshift, vx_setall<float>(0));
```
terfendail: There is a possibility that this version will be slower on SSE2 and SSE3. However I think the line could be optimized by a compiler.

PR #16242, opened 2019-12-27
Add size check for Rodrigues() function 

catree:
### This pullrequest changes

When calling the `Rodrigues` function with a `4x4` (or any wrong size) matrix, the function does nothing and lets the destination matrix unitialized. This leads to this kind of confusion / issue: [SO issue](https://stackoverflow.com/questions/58969553)

---

Which strategy should be adopted to deal with bad arguments (maybe related #8300)?
Here the shape of the input source matrix is wrong, so it makes sense in my opinion to throw an error.
Otherwise, at least a debug mesage or trace log should be somehow recorded.

---

Also, unfortunately [people](https://stackoverflow.com/a/59398256) are still looking at, refering to the 2.4 documentation.

### DIFF
--- a/modules/calib3d/src/calibration.cpp
+++ b/modules/calib3d/src/calibration.cpp
@@ -250,8 +250,6 @@ CV_IMPL void cvComposeRT( const CvMat* _rvec1, const CvMat* _tvec1,
 
 CV_IMPL int cvRodrigues2( const CvMat* src, CvMat* dst, CvMat* jacobian )
 {
-    int depth, elem_size;
-    int i, k;
     double J[27] = {0};
     CvMat matJ = cvMat( 3, 9, CV_64F, J );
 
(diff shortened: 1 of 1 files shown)


modules/calib3d/src/calibration.cpp
```diff
@@ -3219,6 +3221,9 @@ void cv::Rodrigues(InputArray _src, OutputArray _dst, OutputArray _jacobian)
     CV_INSTRUMENT_REGION();
 
     Mat src = _src.getMat();
+    CV_Check(src.rows, (src.rows == 1 && src.cols == 3) || (src.rows == 3 && src.cols == 1) || (src.rows == 3 && src.cols == 3),
```
catree: It is
```
#define CV_Check(v, test_expr, msg)
```
Here, somehow it would have been better to have multiple `v` to print the values of `src.rows` and `src.cols` or have a version without `v` maybe: `#define CV_Check(test_expr, msg)`?

Alternative is to use `CV_Assert((src.rows == 1 && src.cols == 3) || (src.rows == 3 && src.cols == 1) || (src.rows == 3 && src.cols == 3));` 

But I think it is better to have an error message and I hope that most of the users will actually read the full message error.

PR #16245, opened 2019-12-27
VideoBackendRegistry refactoring

VadimLevin:
- Changed raw `for` loops to STL algorithms call or `for-each`
  - Applied library code style
  - Changed `NULL` to `nullptr`
  - Moved dshow and realsense includes under the `#ifdef` macros

### DIFF
--- a/modules/videoio/src/videoio_registry.cpp
+++ b/modules/videoio/src/videoio_registry.cpp
@@ -8,43 +8,50 @@
 
 #include "opencv2/videoio/registry.hpp"
 
-#include "cap_librealsense.hpp"
-#include "cap_dshow.hpp"
+#ifdef HAVE_DSHOW
+    #include "cap_dshow.hpp"
+#endif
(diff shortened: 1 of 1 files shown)


alalek:
Won't merge:
- **patch doesn't fix anything** and doesn't reduce code complexity (except may be `getAvailableCaptureApi()` approach).
- there is code for similar functionality in 3.4 branch. Different code in branches introduces "merge hell". As contribution guides states, if patch is applicable then it should go into maintenance branch first.
- currently there is no automatic tests for this functionality. And no plans to add automatic tests for that. So manual check is required - no free time for that.
- `CV_LOG_INFO(NULL, ...)` is reserved (`NULL` is designed as a part of logging tag identifier, but not added yet).
- `NULL`/`nullptr` was discussed multiple times in opencv/opencv_contrib repositories. Don't touch legacy code without fixing anything.

PR #16246, opened 2019-12-27
VideoIO classes refactoring

VadimLevin:
Changes:
- Added `explicit` to `VideoCapture` constructors with 2
    arguments, 1 of them has default value
- Applied library code style
- Introduced 2 debug macros to improve readability of the code

### DIFF
--- a/modules/videoio/include/opencv2/videoio.hpp
+++ b/modules/videoio/include/opencv2/videoio.hpp
@@ -628,7 +628,7 @@ public:
     implementation if multiple are available: e.g. cv::CAP_FFMPEG or cv::CAP_IMAGES or cv::CAP_DSHOW.
     @sa The list of supported API backends cv::VideoCaptureAPIs
     */
-    CV_WRAP VideoCapture(const String& filename, int apiPreference = CAP_ANY);
+    CV_WRAP explicit VideoCapture(const String& filename, int apiPreference = CAP_ANY);
 
     /** @overload
     @brief  Opens a camera for video capturing
@@ -640,7 +640,7 @@ public:
 
     @sa The list of supported API backends cv::VideoCaptureAPIs
     */
-    CV_WRAP VideoCapture(int index, int apiPreference = CAP_ANY);
+    CV_WRAP explicit VideoCapture(int index, int apiPreference = CAP_ANY);
 
     /** @brief Default destructor
 
--- a/modules/videoio/src/cap.cpp
+++ b/modules/videoio/src/cap.cpp
@@ -50,6 +50,17 @@ static bool param_VIDEOIO_DEBUG = utils::getConfigurationParameterBool("OPENCV_V
 static bool param_VIDEOCAPTURE_DEBUG = utils::getConfigurationParameterBool("OPENCV_VIDEOCAPTURE_DEBUG", false);
 static bool param_VIDEOWRITER_DEBUG = utils::getConfigurationParameterBool("OPENCV_VIDEOWRITER_DEBUG", false);
 
+#define CV_CAPTURE_LOG_DEBUG(tag, ...)                   \
+    if (param_VIDEOIO_DEBUG || param_VIDEOCAPTURE_DEBUG) \
+    {                                                    \
+        CV_LOG_WARNING(nullptr, __VA_ARGS__);            \
+    }
+
+#define CV_WRITER_LOG_DEBUG(tag, ...)                   \
+    if (param_VIDEOIO_DEBUG || param_VIDEOWRITER_DEBUG) \
+    {                                                   \
+        CV_LOG_WARNING(nullptr, __VA_ARGS__)            \
+    }
 
 void DefaultDeleter<CvCapture>::operator ()(CvCapture* obj) const { cvReleaseCapture(&obj); }
 void DefaultDeleter<CvVideoWriter>::operator ()(CvVideoWriter* obj) const { cvReleaseVideoWriter(&obj); }
@@ -80,7 +91,10 @@ bool VideoCapture::open(const String& filename, int apiPreference)
 {
     CV_TRACE_FUNCTION();
 
-    if (isOpened()) release();
+    if (isOpened())
+    {
+        release();
+    }
 
     const std::vector<VideoBackendInfo> backends = cv::videoio_registry::getAvailableBackends_CaptureByFilename();
     for (size_t i = 0; i < backends.size(); i++)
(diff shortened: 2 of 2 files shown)


VadimLevin:
This change breaks ABI compatibility. @alalek what do you think about it?

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -628,7 +628,7 @@ class CV_EXPORTS_W VideoCapture
     implementation if multiple are available: e.g. cv::CAP_FFMPEG or cv::CAP_IMAGES or cv::CAP_DSHOW.
     @sa The list of supported API backends cv::VideoCaptureAPIs
     */
-    CV_WRAP VideoCapture(const String& filename, int apiPreference = CAP_ANY);
+    CV_WRAP explicit VideoCapture(const String& filename, int apiPreference = CAP_ANY);
```
vpisarev: this is useful change

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -640,13 +640,13 @@ class CV_EXPORTS_W VideoCapture
 
     @sa The list of supported API backends cv::VideoCaptureAPIs
     */
-    CV_WRAP VideoCapture(int index, int apiPreference = CAP_ANY);
+    CV_WRAP explicit VideoCapture(int index, int apiPreference = CAP_ANY);
```
vpisarev: useful

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -640,13 +640,13 @@ class CV_EXPORTS_W VideoCapture
...
 
     /** @brief Default destructor
 
     The method first calls VideoCapture::release to close the already opened file or camera.
     */
-    virtual ~VideoCapture();
+    ~VideoCapture();
```
vpisarev: no; please, undo it

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -657,7 +657,7 @@ class CV_EXPORTS_W VideoCapture
 
     The method first calls VideoCapture::release to close the already opened file or camera.
      */
-    CV_WRAP virtual bool open(const String& filename, int apiPreference = CAP_ANY);
+    CV_WRAP bool open(const String& filename, int apiPreference = CAP_ANY);
```
vpisarev: please, restore all the virtual. no reason to change it

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -791,10 +791,10 @@ class CV_EXPORTS_W VideoCapture
      *
      * methods raise exceptions if not successful instead of returning an error code
      */
-    CV_WRAP void setExceptionMode(bool enable) { throwOnFail = enable; }
+    CV_WRAP void setExceptionMode(bool enable) CV_NOEXCEPT { throwOnFail = enable; }
```
vpisarev: is noexcept really important here? let's undo it

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -791,10 +791,10 @@ class CV_EXPORTS_W VideoCapture
...
 
     /// query if exception mode is active
-    CV_WRAP bool getExceptionMode() { return throwOnFail; }
+    CV_WRAP bool getExceptionMode() const CV_NOEXCEPT { return throwOnFail; }
```
vpisarev: same

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -820,7 +820,7 @@ class CV_EXPORTS_W VideoCapture
 protected:
     Ptr<CvCapture> cap;
     Ptr<IVideoCapture> icap;
-    bool throwOnFail;
+    bool throwOnFail { false };
```
vpisarev: it's initialized in the constructor. why initialize it here?

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -872,21 +872,21 @@ class CV_EXPORTS_W VideoWriter
       (eg. FFMPEG FFV1, Huffman HFYU, Lagarith LAGS, etc...)
     - If FFMPEG is enabled, using `codec=0; fps=0;` you can create an uncompressed (raw) video file.
     */
-    CV_WRAP VideoWriter(const String& filename, int fourcc, double fps,
-                Size frameSize, bool isColor = true);
+    CV_WRAP VideoWriter(const String& filename, int fourcc, double fps, const Size& frameSize,
+                        bool isColor = true);
```
vpisarev: let's undo it. no enough reason to break compatibility

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -872,21 +872,21 @@ class CV_EXPORTS_W VideoWriter
...
 
     /** @overload
     The `apiPreference` parameter allows to specify API backends to use. Can be used to enforce a specific reader implementation
     if multiple are available: e.g. cv::CAP_FFMPEG or cv::CAP_GSTREAMER.
      */
     CV_WRAP VideoWriter(const String& filename, int apiPreference, int fourcc, double fps,
-                Size frameSize, bool isColor = true);
+                        const Size& frameSize, bool isColor = true);
```
vpisarev: same

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -872,21 +872,21 @@ class CV_EXPORTS_W VideoWriter
...
 
     /** @brief Default destructor
 
     The method first calls VideoWriter::release to close the already opened file.
     */
-    virtual ~VideoWriter();
+    ~VideoWriter();
```
vpisarev: please, undo

modules/videoio/src/cap.cpp
```diff
@@ -50,21 +50,34 @@ static bool param_VIDEOIO_DEBUG = utils::getConfigurationParameterBool("OPENCV_V
...
+{
+    cvReleaseVideoWriter(&obj);
+}
 
 
-VideoCapture::VideoCapture() : throwOnFail(false)
-{}
+VideoCapture::VideoCapture() = default;
 
-VideoCapture::VideoCapture(const String& filename, int apiPreference) : throwOnFail(false)
```
vpisarev: let's restore it

vpisarev:
@VadimLevin, if I remember correctly, the goal was to add some profiles to video writer to enhance functionality. Here I just see many changes and absolutely no steps towards the goal that we stated. I have no principal objections against the changes in .cpp, as long as it passes all the tests. But besides adding "explicit" to the constructor, please, undo all the changes in header files, because a questionably better style is too high price for the lost compatibility.

VadimLevin:
@vpisarev Yes, I separated work into several independent steps and this was one of them.
`virtual` methods in this class may lead to unexpected behavior for users, those want to inherited their classes from `VideoWriter` and `VideoCapture` because `open` and `release` are called in constructor and destructor respectively. If inheritance is not supposed to be, so there is no reasons to leave virtual methods in these classes at all (I try to follow the rule [C.132: Don’t make a function virtual without reason][https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c132-dont-make-a-function-virtual-without-reason] from Cpp Core Guidelines)

asmorkalov:
@vpisarev @alalek Please take a look again.
@VadimLevin Please update PR description after scope change.

asmorkalov:
@vpisarev friendly reminder.

PR #16248, opened 2019-12-27
Add python tests for cudaobjdetect, cudawarping and cudaarithm modules

cudawarped:
**Merge with contrib**: https://github.com/opencv/opencv_contrib/pull/2396

Adds python tests for bindings added in opencv/opencv_contrib#2396.

The test for `cv.cuda.warpAffine` does not produce the same output as `cv.warpAffine` for all rotations.  Because of this the [C++ tests](https://github.com/opencv/opencv_contrib/blob/65abc7090dedc84bbedec4dfd143f0340e52114f/modules/cudawarping/test/test_warp_affine.cpp#L114) tests have there own host implementation `warpAffineImpl` which seems like overkill in python.  I have therefore only added tests for a rotation which is identical to show verify functionality.


alalek:
Thank you for contribution!

Please move module-specific tests into modules itself (`misc/python/test`).
See example: https://github.com/opencv/opencv/tree/master/modules/dnn/misc/python/test

In test_cuda.py lets keep basic/common tests only.

There is no goal to check "math accuracy" in Python tests. Purpose of Python tests is checking that Python bindings works (pass parameters and fetch results).

cudawarped:
Would you like me to move all of the tests except
`test_cuda_upload_download()`
to their own module locations?

alalek:
Yes.
Leave basic test here.
Please move module-specific tests to module's locations.

cudawarped:
All module specific tests have been moved to the contrib repo.

asmorkalov:
The patch looks good to me. We can merge it right after mege in opencv_contrib repo.

PR #16251, opened 2019-12-28
Allow controlling the random number generator for RTrees training

pwuertz:
Currently the random number generator for training `cv::ml::RTrees` is seeded with a hardcoded value, which ensures that every forest trained with the same data yields identical results.

This PR adds a method for modifying the training seed per instance in order to allow training new or
additional solutions if desired.

Ultimately, this enables parallelized training / evaluation via forest subdivision. I.e. use `N` threads with `M` trees each instead of training a single `N*M` wide forest.

There are other benefits to this as well, like determining the variance of a model for a given set of hyperparameters.

**Update:** PR modified to make use of global `theRNG()` for `RTrees` training, making random forests random by default. Previous deterministic behavior is achieved by calling `cv::setRNGSeed` before training.

### DIFF
--- a/modules/ml/src/rtrees.cpp
+++ b/modules/ml/src/rtrees.cpp
@@ -90,12 +90,12 @@ public:
         CV_TRACE_FUNCTION();
         DTreesImpl::clear();
         oobError = 0.;
-        rng = RNG((uint64)-1);
     }
 
     const vector<int>& getActiveVars() CV_OVERRIDE
     {
         CV_TRACE_FUNCTION();
+        RNG &rng = theRNG();
         int i, nvars = (int)allVars.size(), m = (int)activeVars.size();
         for( i = 0; i < nvars; i++ )
         {
@@ -134,6 +134,7 @@ public:
     bool train( const Ptr<TrainData>& trainData, int flags ) CV_OVERRIDE
     {
         CV_TRACE_FUNCTION();
+        RNG &rng = theRNG();
         CV_Assert(!trainData.empty());
         startTraining(trainData, flags);
         int treeidx, ntrees = (rparams.termCrit.type & TermCriteria::COUNT) != 0 ?
@@ -424,7 +425,6 @@ public:
     double oobError;
     vector<float> varImportance;
     vector<int> allVars, activeVars;
-    RNG rng;
 };
 
 


pwuertz:
Ok, so the way `RTrees` and `RTreesImpl` are designed there is no way of adding something without breaking the ABI. This is ok when switching from, say 4.2 to 4.3, right?

alalek:
> uint64_t seed

Please follow the existed API. Pass `RNG` instead of `seed`. See `setAnnealEnergyRNG()`.

Also there is `theRNG()` (thread-local object). Perhaps it make sense to use it instead of custom per-algorithm).

modules/ml/src/rtrees.cpp
```diff
@@ -425,6 +425,7 @@ class DTreesImplForRTrees CV_FINAL : public DTreesImpl
     vector<float> varImportance;
     vector<int> allVars, activeVars;
     RNG rng;
+    RNG initRng = RNG((uint64) - 1);
```
alalek: What is about this: https://github.com/alalek/opencv/commit/pr16251_r ?
Lets re-use global `theRNG()` (it allows to save / restore RNG state too).

```
cv::theRNG() = RNG(your train seed);
... call train() ...
```

pwuertz:
> Pass `RNG` instead of `seed`. See `setAnnealEnergyRNG()`.

Applied the suggested change. The method now uses RNG instances instead of RNG states.

This makes the method inaccessible for Python though :/.

Also I think this sort of hides the true intent of the function - setting a fixed, specific seed for deterministic training.

> Please follow the existed API.

`cv::setRNGSeed` seems to be a prominent API example using integer seed values and is closely related to the use-case of the function I suggested. Also I'd argue that setting integer seeds for quasi-random processes is a more common API pattern.

> Also there is `theRNG()` (thread-local object). Perhaps it make sense to use it instead of custom per-algorithm).

I think the current idea of having a custom, per-algorithm RNG is very handy. It makes training reproducible by default and independent from other parts of your application.
Sharing a global or thread local RNG for training would break this behavior.

With a per-algorithm seed-setter the current behavior is maintained, yet it enables use-cases where (deterministic) randomness is required (batched or continuous training).

modules/ml/src/rtrees.cpp
pwuertz: Personally, I'd be ok with this. But it would break the default behavior of `RTrees`.
People might be relying on deterministic `RTrees` results.
Honestly, I was actually surprised that OpenCV does this ^^.

pwuertz:
I think your solution (shared RNG) looks way better. You just have to decide if dropping the deterministic default behavior is acceptable.

It shouldn't be a massive shock to anyone, random forests are called "random" for a reason ;).

alalek:
Thank you! Looks good to me.

-----

So, please:

pwuertz:
Thanks for your help!

PR #16254, opened 2019-12-29
cuda4dnn(conv): use cudnn for conv + bias + relu fusion

YashasSamaga:
### This pullrequest changes
cuDNN supports conv + bias + relu fusion in some configurations. This PR enables this cuDNN optimized path whenever possible.

**Device:** GTX 1050

Model                               | without this patch | with this patch
--------------------------------- | ----------------------- | -------------------
Inception v2 MaskRCNN (1024x1024) | 187.6ms               | **180.3ms**


tompollok:
Which card was that? 1050/1060?

YashasSamaga:
@tompollok GTX 1050 and note that I have increased the input image size. It's the same image which was being used earlier but it has been resized to 1024 x 1024.

PR #16199, opened 2019-12-19
Compute gradient x

sajarindider:
resolves #16168

### This pull request adds 
cvtColor function to the computeGradient functions to make it more efficient.

PR #16285, opened 2020-01-04
Fixed small inefficiency in seamless clone init

### LINKED ISSUE #16186: There seems to be a bug in modules/calib3d/src/stereobm.cpp at line 685.

- OpenCV => 4.1.2
- Operating System / Platform => ubuntu 16.04 64 bit
- Compiler => gcc 5.4.0

##### Detailed description

There may be a bug in modules/calib3d/src/stereobm.cpp at line 685.

At line 1161 `cost` is defined as
```
cost.create( left0.size(), CV_16S );
```
while at line 685 :
```
int* costptr = cost.data ? cost.ptr<int>() + lofs + x : &costbuf;
``` 
Shoud not it be `short* costptr`?

stringtheory1234:
Resolves #16186

### DIFF
--- a/modules/photo/src/seamless_cloning_impl.cpp
+++ b/modules/photo/src/seamless_cloning_impl.cpp
@@ -58,10 +58,9 @@ void Cloning::computeGradientX( const Mat &img, Mat &gx)
     else if (img.channels() == 1)
     {
         Mat tmp[3];
-        for(int chan = 0 ; chan < 3 ; ++chan)
-        {
-            filter2D(img, tmp[chan], CV_32F, kernel);
-        }
+        filter2D(img, tmp[0], CV_32F, kernel);
+        tmp[1]=tmp[0];
+        tmp[2]=tmp[0];
         merge(tmp, 3, gx);
     }
 }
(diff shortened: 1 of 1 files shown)


asmorkalov:
@stringtheory1234 Please organize the PR according to contribution guide: https://github.com/opencv/opencv/wiki/How_to_contribute

asmorkalov:
Duplicates https://github.com/opencv/opencv/pull/16199. Closed in favor of original solution.

asmorkalov:
Alternative solution uses cvtColor instead of copy. GRAY2BGR conversion not just copy values to 3 channels, but multiply it with coefficient, so the alternative patch is not equivalent.

asmorkalov:
:+1:

alalek:
> GRAY2BGR conversion not just copy values to 3 channels, but multiply it with coefficient

This is NOT true.
I hope you can find related code (search for `Gray2RGB` implementation).

P.S. Use`~~` to mark ~~statements~~ incorrect.

asmorkalov:
@alalek My fault, I screwed up and looked inside another piece of code (another window). Yes, cvtColor is more efficient for this.

Issue #15244, opened 2019-08-06
cv::dnn::readNetFromONNX  pnet load error


xinyi61:
- OpenCV => 4.1.1
- Operating System / Platform => centos x86_64
- Compiler => gcc
##### 
When I load the onnx model (converted from pytorch ) using cv::dnn::readNetFromONNX, it will get the "terminate called after throwing an instance of 'cv::Exception'
  what():  OpenCV(4.1.1) /opt/notebook_files/usefull/opencv-4.1.1/modules/dnn/src/dnn.cpp:525: error: (-2:Unspecified error) Can't create layer "23" of type "Exp" in function 'getLayerInstance'"

the script i use to convert pytorch to onnx is:
example = torch.rand(1, 3, 12, 12, requires_grad=True).to(config.DEVICES)
torch_out = torch.onnx._export(pnet,
                                                     example, 
                                                     "pnet_model_final.onnx",
                                                      export_params=True)   

[pnet_Weights.pth](https://github.com/xinyi61/xinyi61.github.TEST/blob/master/pnet_Weights.pth)

[pnet_model_final.onnx](https://github.com/xinyi61/xinyi61.github.TEST/blob/master/pnet_model_final.onnx)

adamhrv:
Similar problem `Can't create layer "649" of type "Exp" in function 'getLayerInstance'`. I'm trying to load the ONNX Yolo object detection model from https://github.com/ultralytics/yolov3
```
net = cv.dnn.readNetFromONNX('yolov3.onnx')
cv2.error: OpenCV(4.1.1) ../modules/dnn/src/dnn.cpp:525: error: (-2:Unspecified error) Can't create layer "649" of type "Exp" in function 'getLayerInstance'
```

Steps to repeat error:
```
import cv2 as cv
net = cv.dnn.readNetFromONNX('yolov3.onnx')
```
Download ONNX model from: https://drive.google.com/drive/folders/1uxgUBemJVw9wZsdpboYbzUN4bcRhsuAI
Using OpenCV 4.1.1 via `conda install -c conda-forge opencv`

ThanosRoidis:
Any luck with that? I get the same error when loading an exponential layer from tensorflow and, as crazy as it sounds, I am starting to think that  `tf.exp()` is not supported yet or, at the very least, it can not be a final output layer...

ansonku:
meet the same issue. any comments ?

dkurt:
[...]

PR #16291, opened 2020-01-06
ONNX graphs simplifier

dkurt:
### This pullrequest changes

resolves https://github.com/opencv/opencv/issues/15244

Share graphs simplification logic from TensorFlow importer to ONNX importer.

**Merge with extra**: https://github.com/opencv/opencv_extra/pull/694

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/graph_simplifier.cpp
+++ b/modules/dnn/src/graph_simplifier.cpp
@@ -1,0 +1,207 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+// Copyright (C) 2020, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective owners.
+
+#include "precomp.hpp"
+
+#include "graph_simplifier.hpp"
+
+#include <queue>
+
+namespace cv { namespace dnn {
+
+Subgraph::~Subgraph() {}
+
+int Subgraph::addNodeToMatch(const std::string& op, int input_0, int input_1,
+                             int input_2, int input_3)
+{
+    int nodeInputs[] = {input_0, input_1, input_2, input_3};
+    int numInputs = 0;
+    for (int i = 0; i < 4; ++i)
+    {
+        numInputs += (int)(nodeInputs[i] != -1);
+    }
+    return addNodeToMatch(op, std::vector<int>(&nodeInputs[0], &nodeInputs[0] + numInputs));
+}
+
+int Subgraph::addNodeToMatch(const std::string& op, const std::vector<int>& inputs_)
+{
+    for (int i = 0; i < inputs_.size(); ++i)
+    {
+        CV_Assert(inputs_[i] < (int)nodes.size());
+    }
+    nodes.push_back(op);
+    inputs.push_back(inputs_);
+    return nodes.size() - 1;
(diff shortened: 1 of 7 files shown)


modules/dnn/src/graph_simplifier.cpp
```diff
@@ -0,0 +1,205 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+// Copyright (C) 2020, Intel Corporation, all rights reserved.
+// Third party copyrights are property of their respective owners.
+
+#include "graph_simplifier.hpp"
```
alalek: `precomp.hpp` should be first

modules/dnn/src/graph_simplifier.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+private:
+    std::vector<std::string> nodes;         // Nodes to be matched in the origin graph.
+    std::vector<std::vector<int> > inputs;  // Connections of an every node to it's inputs.
+
+    std::string fusedNodeOp;           // Operation name of resulting fused node.
+    std::vector<int> fusedNodeInputs;  // Inputs of fused node.
+};
+
+void simplifySubgraphs(Ptr<ImportGraphWrapper> net,
```
alalek: `const Ptr` because you don't modify `Ptr` itself (`const` says nothing about modification of its contents)

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -0,0 +1,157 @@
...
+private:
+    int axis;
+};
+
+void simplifySubgraphs(opencv_onnx::GraphProto& net)
+{
+    std::vector<Ptr<Subgraph> > subgraphs;
+    subgraphs.push_back(Ptr<Subgraph>(new SoftMaxSubgraph()));
+
+    simplifySubgraphs(Ptr<ImportGraphWrapper>(new ONNXGraphWrapper(net)), subgraphs);
```
alalek: `makePtr` for these 2 cases

dkurt: cannot do it for `ONNXGraphWrapper` due it uses non-const reference to delete nodes from the `net`. But `makePtr` works with `const` arguments:

https://github.com/opencv/opencv/blob/b58b2c5ae8b857e4e897a4a8d9265993202ecebc/modules/core/include/opencv2/core/cvstd_wrapper.hpp#L25-L26

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+private:
+    int numInputs;
+    opencv_onnx::GraphProto& net;
+};
+
+class SoftMaxSubgraph : public Subgraph
+{
+public:
+    SoftMaxSubgraph()
```
mshabunin: `axis` is not initialized here.

dkurt: Thanks! Fixed in https://github.com/opencv/opencv/pull/16353

PR #16302, opened 2020-01-08
Fix build on non-Linux glibc-based systems

sthibaul:
dl functions are provided by all glibc-based systems (GNU/Linux, but
also GNU/Hurd, GNU/kFreeBSD)


asmorkalov:
:+1:

PR #16309, opened 2020-01-08
Added resize flag handlers for imread and imdecode | Issue #16203

### LINKED ISSUE #16203: imdecode ignores IMREAD_REDUCED_* flags
Passing any of the `IMREAD_REDUCED_` flags, e.g. `IMREAD_REDUCED_GRAYSCALE_2`, to the flags parameter of `imdecode()` has no effect. I notice the `imread_()` function includes this code:

    int scale_denom = 1;
    if( flags > IMREAD_LOAD_GDAL )
    {
    if( flags & IMREAD_REDUCED_GRAYSCALE_2 )
        scale_denom = 2;
    else if( flags & IMREAD_REDUCED_GRAYSCALE_4 )
        scale_denom = 4;
    else if( flags & IMREAD_REDUCED_GRAYSCALE_8 )
        scale_denom = 8;
    }

    /// set the scale_denom in the driver
    decoder->setScale( scale_denom );

This code is absent in `imdecode()` (and `imdecode_`).

For those that aren't familiar with that flag: The expected effect is that the image is resized by the stated factor if it is a JPEG. This happens for "free" due to the way the JPEG encoding works, so it is much faster than opening the image at full resolution and then doing a separate resize.

ganesh-k13:
# This pullrequest changes

1. Added flag handler for imdecode which was the issue
2. Take care of color resize cases
3. Port changes to imread as well.

resolves #16203

```
force_builders=Win32
```

### DIFF
--- a/modules/imgcodecs/src/loadsave.cpp
+++ b/modules/imgcodecs/src/loadsave.cpp
@@ -423,12 +423,12 @@ imread_( const String& filename, int flags, int hdrtype, Mat* mat=0 )
     int scale_denom = 1;
     if( flags > IMREAD_LOAD_GDAL )
     {
-    if( flags & IMREAD_REDUCED_GRAYSCALE_2 )
-        scale_denom = 2;
-    else if( flags & IMREAD_REDUCED_GRAYSCALE_4 )
-        scale_denom = 4;
-    else if( flags & IMREAD_REDUCED_GRAYSCALE_8 )
-        scale_denom = 8;
+        if( flags & IMREAD_REDUCED_GRAYSCALE_2 )
+            scale_denom = 2;
+        else if( flags & IMREAD_REDUCED_GRAYSCALE_4 )
+            scale_denom = 4;
+        else if( flags & IMREAD_REDUCED_GRAYSCALE_8 )
+            scale_denom = 8;
     }
 
     /// set the scale_denom in the driver
@@ -766,6 +766,20 @@ imdecode_( const Mat& buf, int flags, int hdrtype, Mat* mat=0 )
     if( !decoder )
         return 0;
 
+    int scale_denom = 1;
+    if( flags > IMREAD_LOAD_GDAL )
+    {
+        if( flags & IMREAD_REDUCED_GRAYSCALE_2 )
+            scale_denom = 2;
+        else if( flags & IMREAD_REDUCED_GRAYSCALE_4 )
+            scale_denom = 4;
+        else if( flags & IMREAD_REDUCED_GRAYSCALE_8 )
+            scale_denom = 8;
+    }
+
+    /// set the scale_denom in the driver
+    decoder->setScale( scale_denom );
+
     if( !decoder->setSource(buf_row) )
     {
         filename = tempfile();
@@ -861,7 +875,7 @@ imdecode_( const Mat& buf, int flags, int hdrtype, Mat* mat=0 )
     {
         std::cerr << "imdecode_('" << filename << "'): can't read data: unknown exception" << std::endl << std::flush;
     }
-    decoder.release();
+
     if (!filename.empty())
     {
         if (0 != remove(filename.c_str()))
@@ -878,6 +892,13 @@ imdecode_( const Mat& buf, int flags, int hdrtype, Mat* mat=0 )
             mat->release();
         return 0;
     }
+
+    if( decoder->setScale( scale_denom ) > 1 ) // if decoder is JpegDecoder then decoder->setScale always returns 1
+    {
+        resize( *mat, *mat, Size( size.width / scale_denom, size.height / scale_denom ), 0, 0, INTER_LINEAR_EXACT);
+    }
+
+    decoder.release();
 
     return hdrtype == LOAD_CVMAT ? (void*)matrix :
         hdrtype == LOAD_IMAGE ? (void*)image : (void*)mat;
--- a/modules/imgcodecs/test/test_read_write.cpp
+++ b/modules/imgcodecs/test/test_read_write.cpp
@@ -5,38 +5,43 @@
 
 namespace opencv_test { namespace {
 
-/* Flag_Resize_Data = <file_name, resize_flag_and_dims> */
-typedef tuple< string, tuple < int, int, int > > Flag_Resize_Data;
+/* < <file_name, image_size>, <imread mode, scale> > */
+typedef tuple< tuple<string, Size>, tuple<ImreadModes, int> > Imgcodecs_Resize_t;
 
-typedef testing::TestWithParam< Flag_Resize_Data > Imgcodecs_Resize;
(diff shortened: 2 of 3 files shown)


ganesh-k13:
Hi @sturkmen72, I have added the handlers as discussed. In addition to this I noticed that color resize was not taking place in case of imread, I fixed that as well.

sturkmen72:
i believe your addition is not correct ( or unnecessary)

sturkmen72:
also see https://github.com/opencv/opencv/blob/master/modules/imgcodecs/src/loadsave.cpp#L494 should be added in imdecode

ganesh-k13:
Hi @sturkmen72 thanks for the response, so the addition to imread is not needed? Or even imdecode is not necessary? As mentioned in last comment, do I add that to imdecode?

modules/imgcodecs/src/loadsave.cpp
```diff
@@ -421,14 +421,11 @@ imread_( const String& filename, int flags, int hdrtype, Mat* mat=0 )
...
     {
-    if( flags & IMREAD_REDUCED_GRAYSCALE_2 )
-        scale_denom = 2;
-    else if( flags & IMREAD_REDUCED_GRAYSCALE_4 )
-        scale_denom = 4;
-    else if( flags & IMREAD_REDUCED_GRAYSCALE_8 )
-        scale_denom = 8;
+        /// Make it an even number as color flags are odd
+        /// We scale the image based on resize factor
+        scale_denom = ( flags % 2 == 0 ) ? flags >> 3 : ( flags - 1 ) >> 3;
```
alalek: Not readable.
Not maintainable (after enum refactoring code may become broken).

ganesh-k13: Thanks for the review @alalek, do I add the if conditions for greyscale and color, then it'll be readable, also I add these conditions to imread as well right?

modules/imgcodecs/src/loadsave.cpp
```diff
@@ -421,14 +421,14 @@ imread_( const String& filename, int flags, int hdrtype, Mat* mat=0 )
     }
 
     int scale_denom = 1;
-    if( flags > IMREAD_LOAD_GDAL )
+    if( ( flags > IMREAD_LOAD_GDAL ) & ( flags < IMREAD_IGNORE_ORIENTATION ) )
```
sturkmen72: if( flags > IMREAD_LOAD_GDAL ) is good here ( to make usage like `IMREAD_REDUCED_GRAYSCALE_4 | IMREAD_IGNORE_ORIENTATION`  acceptable )
I recommend that you do not make any changes to the `imread ` function

ganesh-k13: Sure @sturkmen72, I'll remove it.

modules/imgcodecs/src/loadsave.cpp
```diff
@@ -766,6 +766,20 @@ imdecode_( const Mat& buf, int flags, int hdrtype, Mat* mat=0 )
     if( !decoder )
         return 0;
 
+    int scale_denom = 1;
+    if( ( flags > IMREAD_LOAD_GDAL ) & ( flags < IMREAD_IGNORE_ORIENTATION ) )
```
sturkmen72: if( flags > IMREAD_LOAD_GDAL )

ganesh-k13: Oh sorry, my bad, skipped this, thanks.

modules/imgcodecs/test/test_read_write.cpp
```diff
@@ -5,6 +5,151 @@
 
 namespace opencv_test { namespace {
 
+TEST(Imgcodecs_Image, imread_reduce_flags)
```
alalek: Please add `#ifdef HAVE_JPEG` guard for these tests (to avoid test failures in some configurations).

ganesh-k13: Sure @alalek, will do

sturkmen72:
`imread `manually resizes other formats than jpeg see https://github.com/opencv/opencv/blob/master/modules/imgcodecs/src/loadsave.cpp#L494-L497

ganesh-k13:
@sturkmen72, thanks for that, shall I add that to imdecode as well as you showed earlier?

alalek:
@ganesh-k13 Please fix warnings from Win32 builder (add `static_cast<size_t>()`)

sturkmen72:
@alalek let me do some tests before merge

ganesh-k13:
> `imread `manually resizes other formats than jpeg see https://github.com/opencv/opencv/blob/master/modules/imgcodecs/src/loadsave.cpp#L494-L497

@sturkmen72, do we make this in a different PR, or shall I add it to `imdecode` now?

ganesh-k13:
Hey @sturkmen72 / @alalek, I'm getting cores if we add that if conditions and try other formats, if that if is not there then resize does not happen but no crash. It involves some work, can I raise a different PR for that, might involve more work. Please advise.

sturkmen72:
you can add 
```
    if( decoder->setScale( scale_denom ) > 1 ) // if decoder is JpegDecoder then decoder->setScale always returns 1
    {
        resize( mat, mat, Size( size.width / scale_denom, size.height / scale_denom ), 0, 0, INTER_LINEAR_EXACT);
    }
   decoder.release();
``` 
end of imdecode_  
( delete `decoder.release();` which located a few lines upper )

ganesh-k13:
Ohh thanks, I didn't see it was getting freed. I'll add that and test.

ganesh-k13:
@sturkmen72 thanks for the fix, it works. @alalek I have added a case for the same by resizing a PNG to hit that code path.

ganesh-k13:
@sturkmen72 / @alalek, any updates on this one? any changes needed?

ganesh-k13:
Thanks for the link @asmorkalov, I'll make those changes

alalek:
Parametrized tests look like overkill here. Duplicated code and checks may be moved into common function, like [this](https://github.com/opencv/opencv/blob/4.2.0/modules/core/test/test_eigen.cpp#L390-L455) (important note, that return value should be 'void').

ganesh-k13:
Hi @alalek , @sturkmen72 , @asmorkalov, we can see that imread and imdecode have duplicate code in both test and src. I would suggest we make it a different PR as we have digressed from the initial issue. I can raise a new issue and PR for that.

ganesh-k13:
Hi @alalek , @asmorkalov , I went ahead with refactor of tests anyways. I went with `Value-Parameterized Tests` as I wanted to try something new and it seems very well suited for our case. Please merge this PR if changes are satisfactory.

asmorkalov:
@ganesh-k13 Well done! Please squash your commits to have clean history and we can merge the patch. @alalek What do you think?

PR #16322, opened 2020-01-11
Add package for cmake project

ruffsl:
This PR simply adds a package.xml used for build tools like catkin or colcon, allowing for the essential build dependencies to be readily identified and when scheduling build order from dependency graphs.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>libopencv-dev</name>
+  <version>4.2.0</version>
+  <description>
+    Open Source Computer Vision Library.
+  </description>
+  <maintainer email="opencv@opencv.org">opencv</maintainer>
+  <license>BSD</license>
+
+  <buildtool_depend>cmake</buildtool_depend>
+  
+  <build_depend>eigen</build_depend>
+
+  <export>
+    <build_type>cmake</build_type>
+  </export>
+</package>


package.xml
```diff
@@ -0,0 +1,19 @@
...
+  <version>4.2.0</version>
+  <description>
+    Open Source Computer Vision Library.
+  </description>
+  <maintainer email="opencv@opencv.org">opencv</maintainer>
+  <license>BSD</license>
+
+  <buildtool_depend>cmake</buildtool_depend>
+  
+  <build_depend>eigen</build_depend>
```
alalek: Is this list really complete?

asmorkalov:
@ruffsl friendly reminder

mikaelarguedas:
@ruffsl If it's useful, we used the following `package.xml` files to package opencv in ROS Lunar in the past: https://github.com/ros-gbp/opencv3-release/blob/3ffcbb05e149676d9f92a5d66c3f4b5da42962d7/lunar/package.xml (this was for opencv 3.3.1 at the time)

ruffsl:
> Some guide/link how to use that would be extremely useful.

Sorry, here is some more context as per the motivation and rationale for include a package manifest for summarizing the package depencies, licensing, etc:

Package Manifest Format Three Specification:
https://www.ros.org/reps/rep-0149.html

> Platform-specific files should be placed under the "platforms" directory

This package manifest is os or platform agnostic, we use them in ROS for cmake packages built on windows, linux, macos.. and is necessarily located in the same directory as the root `CMakeLists.txt` for the project. But perhaps we could iterate on alternative idea for its location. Perhaps a fancy cmake file placed in its platforms

package.xml
ruffsl: As @mikaelarguedas https://github.com/opencv/opencv/pull/16322#issuecomment-575364146 pointed out, we could specify more of the optional dependencies, I just wasn't sure to what extent  we'd like to include them explicitly as required, though the one linked seems like a more practical useful build for general users.

alalek:
I believe, dedicated place/repository for package manager configuration is preferable, than pushing these configurations into upstream reporsitory.

Mixing upstream source code with packaging details of specific package manager doesn't look as a good idea. All package managers and/or their build configurations have own specific stuff.

Refer to multi-platform [homebrew](https://github.com/Homebrew/linuxbrew-core/blob/master/Formula/opencv.rb), [vcpkg](https://github.com/microsoft/vcpkg/tree/master/ports/opencv4), [conda](https://github.com/conda-forge/opencv-feedstock) package managers or even Linux distros ([Ubuntu](https://launchpad.net/ubuntu/+source/opencv/) / [Fedora](https://src.fedoraproject.org/rpms/opencv/tree/master) / [Arch](https://git.archlinux.org/svntogit/packages.git/tree/trunk?h=packages/opencv) / etc).

---

> `<package format="2">`

`REP-0149` says about `format="3"`. It is hard to support multiple versions in upstream repository.

---

/cc @vrabaud as contributor of OpenCV and https://github.com/ros-gbp/opencv3-release repository.

PR #16325, opened 2020-01-11
core(SIMD): fix NEON alignmnet issue

alalek:


modules/core/include/opencv2/core/hal/intrin_neon.hpp
```diff
@@ -2129,10 +2129,12 @@ inline v_float32x4 v_lut(const float* tab, const int* idx)
 }
 inline v_float32x4 v_lut_pairs(const float* tab, const int* idx)
 {
+    typedef uint64 CV_DECL_ALIGNED(1) unaligned_uint64;
```
terfendail: There are a few places in the file that use this approach. Maybe it make sense to define this unaligned types outside of specific functions?

alalek: I believe, it is not critical enough for now.
BTW, This one may use `CV_DECL_ALIGNED(4)` (if code would compile)

terfendail: OK. That's really not a big issue.

PR #16334, opened 2020-01-13
ml/python: fix digits samples

berak:
fix loading serialized SVM / KNearest and findContours() usage

### DIFF
--- a/samples/python/digits_video.py
+++ b/samples/python/digits_video.py
@@ -1,4 +1,12 @@
 #!/usr/bin/env python
+'''
+Digit recognition from video.
+
+Run digits.py before, to train and save the SVM.
+
+Usage:
+  digits_video.py [{camera_id|video_file}]
+'''
 
 # Python 2/3 compatibility
 from __future__ import print_function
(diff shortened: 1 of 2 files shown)


samples/python/digits_video.py
```diff
@@ -41,7 +45,7 @@ def main():
 
         bin = cv.adaptiveThreshold(gray, 255, cv.ADAPTIVE_THRESH_MEAN_C, cv.THRESH_BINARY_INV, 31, 10)
         bin = cv.medianBlur(bin, 3)
-        _, contours, heirs = cv.findContours( bin.copy(), cv.RETR_CCOMP, cv.CHAIN_APPROX_SIMPLE)
+        contours, heirs = cv.findContours( bin.copy(), cv.RETR_CCOMP, cv.CHAIN_APPROX_SIMPLE)
```
alalek: OpenCV 3.4.x still uses 3 outputs (due compatibility reasons).

Please apply this fix on master branch (if applicable).

berak: ah, right, will do.

alalek: I mean separate patch. Keep this for 3.4.

BTW, patch for master is not needed ([code](https://github.com/opencv/opencv/blob/4.2.0/samples/python/digits_video.py#L44)).

alalek:
> +22 -12

Both patches are similar, so patch onto 3.4 is enough (we will merge these changes into "master" by regular "merge 3.4" process).

Closing this.

berak:
yea, right. thanks !

PR #16118, opened 2019-12-10
G-API: GOpaque implementation

smirnov-alexey:
This pullrequest adds a new type in G-API - GOpaque which is designed to store a single custom user type.

Currently it's almost copypasted GArray except other internal types are not related to vector.

<cut/>

```
build_gapi_standalone:Linux x64=ade-0.1.1f
build_gapi_standalone:Win64=ade-0.1.1f
build_gapi_standalone:Mac=ade-0.1.1f
build_gapi_standalone:Linux x64 Debug=ade-0.1.1f

force_builders=Custom,Custom Win,ARMv7
buildworker:Custom=linux-1
#build_image:Custom=ubuntu-clang:18.04
build_image:Custom=centos:7
build_contrib:Custom=OFF
```

PR #16338, opened 2020-01-13
cmake: support OPENCV_ABI_SKIP_MODULES_LIST

alalek:
To regenerate ABI/API snapshot of OpenCV 4.2.0 without G-API module

relates #16118

### DIFF
--- a/cmake/OpenCVGenABI.cmake
+++ b/cmake/OpenCVGenABI.cmake
@@ -2,6 +2,8 @@ if (NOT GENERATE_ABI_DESCRIPTOR)
   return()
 endif()
 
+set(OPENCV_ABI_SKIP_MODULES_LIST "" CACHE STRING "List of modules to exclude from ABI checker")
+
 set(filename "opencv_abi.xml")
 set(path1 "${CMAKE_BINARY_DIR}/${filename}")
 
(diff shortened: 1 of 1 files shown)


dmatveev:
As the opposite, is there any way to mark problematic (for ABI checker) parts of code to skip the check, but to keep it for the rest of G-API?

dmatveev:
```diff
--- _ZN2cv6detail10MetaHelperINS_4gapi4core8GAbsDiffESt5tupleIJNS_4GMatES6_EES6_E10getOutMetaERKSt6vectorINS_4util7variantIJNSA_9monostateENS_8GMatDescENS_11GScalarDescENS_10GArrayDescEEEESaISG_EERKS9_INS_4GArgESaISL_EE
+++ _ZN2cv6detail10MetaHelperINS_4gapi4core8GAbsDiffESt5tupleIJNS_4GMatES6_EES6_E10getOutMetaERKSt6vectorINS_4util7variantIJNSA_9monostateENS_8GMatDescENS_11GScalarDescENS_10GArrayDescENS_11GOpaqueDescEEEESaISH_EERKS9_INS_4GArgESaISM_EE
```

This and other ~180 hits are purely mangling stuff, but source-level compatibility is preserved

alalek:
Used tool "abi-compliance-checker" is not very good on source-level compatibility checks. It is designed as ABI checker at first. API (source) check list is just relaxed subset of ABI checks.

There are too many false positive hits on GAPI (and non G-API) patches.
Lets try to "validate" source-level compatibility through compilation of corresponding tests and samples. G-API has good coverage (than other OpenCV code), so serious issues should not pass.

dmatveev:
I think this particular ABI false alarm can be fixed, though the fix will generate 180 hits again :)

alalek:
:+1:

PR #16342, opened 2020-01-13
Convolution padding parsing modified

### LINKED ISSUE #16259: darknet importer: padding is parsed incorrectly for convolutional layers
- OpenCV => 4.2.0
- Compiler => g++ 7.4

##### Detailed description

It appears that the `pad` value in darknet configuration files indicates the padding mode and not the padding value itself. I think `pad` being one implies same padding and `pad` being zero implies manual padding (which is specified in the `padding` parameter).

https://github.com/AlexeyAB/darknet/blob/dcfeea30f195e0ca1210d580cac8b91b6beaf3f7/src/parser.c#L180-L182

The OpenCV darknet parser treats the pad value as the actual padding value. There is also a comment which shows that the original author of the parser was stunned by the weird `pad=1` and sets it to zero. It's actually zero because `pad=1` would imply a padding of `kernel_size//2` which would anyway give zero.

https://github.com/opencv/opencv/blob/7b12cbd7dcbb025c31e7630c9ae709493a8f3d62/modules/dnn/src/darknet/darknet_io.cpp#L566-L567

ashishkrshrivastava:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/697

### This pullrequest modifies the parsing of 'padding' for convolution layers in darknet importer.
resolves #16259

```
opencv_extra=opencv_extra_modified
```

### DIFF
--- a/modules/dnn/src/darknet/darknet_io.cpp
+++ b/modules/dnn/src/darknet/darknet_io.cpp
@@ -556,6 +556,7 @@ namespace cv {
                     {
                         int kernel_size = getParam<int>(layer_params, "size", -1);
                         int pad = getParam<int>(layer_params, "pad", 0);
+                        int padding = getParam<int>(layer_params, "padding", 0);
                         int stride = getParam<int>(layer_params, "stride", 1);
                         int filters = getParam<int>(layer_params, "filters", -1);
                         bool batch_normalize = getParam<int>(layer_params, "batch_normalize", 0) == 1;
@@ -563,13 +564,13 @@ namespace cv {
                         if (flipped == 1)
                             CV_Error(cv::Error::StsNotImplemented, "Transpose the convolutional weights is not implemented");
 
-                        // correct the strange value of pad=1 for kernel_size=1 in the Darknet cfg-file
-                        if (kernel_size < 3) pad = 0;
+                        if (pad)
+                            padding = kernel_size/2;
 
                         CV_Assert(kernel_size > 0 && filters > 0);
                         CV_Assert(current_channels > 0);
 
-                        setParams.setConvolution(kernel_size, pad, stride, filters, current_channels,
+                        setParams.setConvolution(kernel_size, padding, stride, filters, current_channels,
                             batch_normalize);
 
                         current_channels = filters;
--- a/modules/dnn/src/darknet/darknet_io.cpp
+++ b/modules/dnn/src/darknet/darknet_io.cpp
@@ -565,7 +565,7 @@ namespace cv {
                             CV_Error(cv::Error::StsNotImplemented, "Transpose the convolutional weights is not implemented");
 
                         if (pad)
-                            padding = kernel_size/2;
+                            padding = kernel_size / 2;
 
                         CV_Assert(kernel_size > 0 && filters > 0);
                         CV_Assert(current_channels > 0);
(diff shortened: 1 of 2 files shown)


alalek:
Thank you for contribution!

Could you please try to add testcase + generate simple model for that? (see issue's comment)

ashishkrshrivastava:
@alalek you mean adding a model in .cfg file and test data for convolution layer in https://github.com/opencv/opencv_extra/tree/master/testdata/dnn/darknet module so that it can be tested using generate_darknet_models.py. Am I right?

alalek:
Yes, some small models with `pad`/`padding` parameters (1 layer with random weights should be enough).

ashishkrshrivastava:
@alalek, I have pushed a PR for a 2-layer convolution layer model. one with pad = 1(SAME) and one with manual padding. I have also added test data. https://github.com/opencv/opencv_extra/pull/697

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -556,20 +556,21 @@ namespace cv {
                     {
                         int kernel_size = getParam<int>(layer_params, "size", -1);
                         int pad = getParam<int>(layer_params, "pad", 0);
+                        int padding = getParam<int>(layer_params, "padding", 0);
```
dkurt: What if `padding > 0` ?

ashishkrshrivastava: if pad=1 then padding will be assigned (kernel_size/2). if pad=0 or pad is not mentioned then padding will be manual i.e. whatever mentioned in model and if padding and pad both are not mentioned then padding will be assigned 0.

dkurt:
Please add a corresponding test at https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_onnx_importer.cpp and change target branch to 3.4

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -556,20 +556,21 @@ namespace cv {
...
                         int filters = getParam<int>(layer_params, "filters", -1);
                         bool batch_normalize = getParam<int>(layer_params, "batch_normalize", 0) == 1;
                         int flipped = getParam<int>(layer_params, "flipped", 0);
                         if (flipped == 1)
                             CV_Error(cv::Error::StsNotImplemented, "Transpose the convolutional weights is not implemented");
 
-                        // correct the strange value of pad=1 for kernel_size=1 in the Darknet cfg-file
-                        if (kernel_size < 3) pad = 0;
+                        if (pad)
+                            padding = kernel_size/2;
```
dkurt: Please keep extra whitespace before and after `/`

ashishkrshrivastava:
> Please add a corresponding test at https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_onnx_importer.cpp and change target branch to 3.4

@dkurt shouldn't I add test at 'test_darknet_importer.cpp' rather than 'test_onnx_importer.cpp'

dkurt:
@ashishkrshrivastava, sure, I'm sorry for misleading.

ashishkrshrivastava:
> Please add a corresponding test at https://github.com/opencv/opencv/blob/master/modules/dnn/test/test_onnx_importer.cpp and change target branch to 3.4

@dkurt, you mean to work on branch 3.4 instead of master ?

dkurt:
@ashishkrshrivastava, right. We'll port to master after the merge.

ashishkrshrivastava:
Created a new PR #16358 for the same issue. Closing this PR.

PR #16350, opened 2020-01-15
Fix lint issues in Android samples

hannesa2:
It fixes lint errors and warnings in Android samples.
Because library uses functions which requires camera permission, it makes more sense to move permissions to library instead of apply it to each sample separat

### DIFF
--- a/samples/android/15-puzzle/src/org/opencv/samples/puzzle15/Puzzle15Activity.java
+++ b/samples/android/15-puzzle/src/org/opencv/samples/puzzle15/Puzzle15Activity.java
@@ -22,7 +22,7 @@ import java.util.List;
 
 public class Puzzle15Activity extends CameraActivity implements CvCameraViewListener, View.OnTouchListener {
 
-    private static final String  TAG = "Sample::Puzzle15::Activity";
+    private static final String  TAG = "Puzzle15::Activity";
 
     private CameraBridgeViewBase mOpenCvCameraView;
     private Puzzle15Processor    mPuzzle15;
--- a/samples/android/camera-calibration/src/org/opencv/samples/cameracalibration/CalibrationResult.java
+++ b/samples/android/camera-calibration/src/org/opencv/samples/cameracalibration/CalibrationResult.java
@@ -8,7 +8,7 @@ import android.content.SharedPreferences;
 import android.util.Log;
 
 public abstract class CalibrationResult {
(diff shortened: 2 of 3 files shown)


asmorkalov:
@hannesa2 I think that's bad idea to move permissions to library project. I understand that most popular case is application with camera as input, but OpenCV can be used for apps without camera, e.g. Gallery analysis or for apps for alternative camera configuration. I propose to presume current layout.

hannesa2:
@asmorkalov In my point of view it's not that bad. If you run `./gradlew lint` you will run into an error too because library uses camera features. 
Please have a view on the other side: You use the library and have to specify the permissions by your own everytime. It's a source of error on runtime (not build time). 
To add permission in AndroidManifest is one thing, but request it from the user is an other thing.
Your gallery analysis example does not need to request it from the user.

Anyway, I add this is an own commit to easy remove it again, because I expect such doubt.
Please confirm the removement of this commit and I will do immediate

asmorkalov:
@alalek what do you think?

hannesa2:
> BTW, "camera" is not single required permission to run apps, there is also "storage" access permissions in some samples.

But "camera" is the only permission which causes a lint issue here.
I removed the commit with the camera permission

asmorkalov:
:+1:

alalek:
> But "camera" is the only permission which causes a lint issue here.

Is it possible to have separate manifest for lint tool? or configure lint tool to skip manifest issues?

alalek:

So, please:

hannesa2:
> or configure lint tool to skip manifest issues

I can/will ignore them.

But step by step.

What I want to do, is a error free lint and targetSdk 29
For this several steps are needed.

1. this PR
2. ignore remaining lint errors (as less as possible)
3. targetSdk 29
4. for this Android X is needed.

but I'm not sure if it works for non Gradle 3.4 as well

hannesa2:
@alalek 
I changed the base to 3.4.
What's your change-base-short-tutorial would make perfect, is the missing command how to generate a Android (3.4) release localy on my machine for test purpose.
How to do so ?

alalek:
OpenCV 3.4 requires Java 6, so some changes doesn't fit well.
I will backport them separately.

Please revert "rebase" back:
```
git checkout -B fixLintIssuesAndroid 4057e81b7649d8d814908157478456c6889271cd
git push origin fixLintIssuesAndroid --force
```

PR #16352, opened 2020-01-15
add mobilenet-objdetect to Android samples

hannesa2:
Exported `opencv-X.X.X-android-sdk.zip` from https://sourceforge.net/projects/opencvlibrary/files does not includes `mobilenet-objdetect` and `tutorial-4-opencl` example.

<cut/>

image

I'm not sure if this is the right place and this pull request solves the add of `mobilenet-objdetect`. 
And why `tutorial-4-opencl` is missing too ?


dkurt:
This sample won't work because it requires additional data (the network).

alalek:
> push them by adb to application's data directory

Some Android configurations may disallow that.

---

I believe we should start with preparing "gradle" file which bundles models as APK resources (assume some location of model on local disk or download them itself from internet (with MD5 checking)).

hannesa2:
Ok, I see it's a bigger task.
The question about `tutorial-4-opencl` is sill open

PR #16354, opened 2020-01-15
Wrap custom OpenCV layers to try-catch

dkurt:
### This pullrequest changes

```cpp
ie.AddExtension(extension, "CPU");
```
is wrapped to try-catch so we need to do the same for
```cpp
ie.AddExtension(std::make_shared<InfEngineExtension>(), "CPU");
```

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/ie_ngraph.cpp
+++ b/modules/dnn/src/ie_ngraph.cpp
@@ -323,7 +323,14 @@ void InfEngineNgraphNet::initPlugin(InferenceEngine::CNNNetwork& net)
             }
             // Some of networks can work without a library of extra layers.
             // OpenCV fallbacks as extensions.
-            ie.AddExtension(std::make_shared<InfEngineExtension>(), "CPU");
+            try
+            {
+                ie.AddExtension(std::make_shared<InfEngineExtension>(), "CPU");
+            }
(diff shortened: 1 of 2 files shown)


modules/dnn/src/ie_ngraph.cpp
```diff
@@ -323,7 +323,11 @@ void InfEngineNgraphNet::initPlugin(InferenceEngine::CNNNetwork& net)
             }
             // Some of networks can work without a library of extra layers.
             // OpenCV fallbacks as extensions.
-            ie.AddExtension(std::make_shared<InfEngineExtension>(), "CPU");
+            try
+            {
+                ie.AddExtension(std::make_shared<InfEngineExtension>(), "CPU");
+            }
+            catch(...) {}
```
alalek: > catch(...) {}

Please add some logging:

```
catch(const std::exception& e)
{
    CV_LOG_INFO(NULL, "DNN-IE: Can't register OpenCV custom layers extension: " << e.what());
}
```

dkurt: done

PR #16358, opened 2020-01-15
Convolutional padding parsing modified

### LINKED ISSUE #16259: darknet importer: padding is parsed incorrectly for convolutional layers

ashishkrshrivastava:
**Merge with extra**:  https://github.com/opencv/opencv_extra/pull/698

This PR modifies the parsing of 'padding' for convolutional layers in darknet importer.

resolves #16259

```
opencv_extra=3.4modified
```

### DIFF
--- a/modules/dnn/src/darknet/darknet_io.cpp
+++ b/modules/dnn/src/darknet/darknet_io.cpp
@@ -556,6 +556,7 @@ namespace cv {
                     {
                         int kernel_size = getParam<int>(layer_params, "size", -1);
                         int pad = getParam<int>(layer_params, "pad", 0);
+                        int padding = getParam<int>(layer_params, "padding", 0);
                         int stride = getParam<int>(layer_params, "stride", 1);
                         int filters = getParam<int>(layer_params, "filters", -1);
                         bool batch_normalize = getParam<int>(layer_params, "batch_normalize", 0) == 1;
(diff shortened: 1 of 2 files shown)


modules/dnn/test/test_darknet_importer.cpp
```diff
@@ -106,7 +106,7 @@ class Test_Darknet_layers : public DNNTestLayer
         std::string cfg = findDataFile("dnn/darknet/" + name + ".cfg");
         std::string model = "";
         if (hasWeights)
-            model = findDataFile("dnn/darknet/" + name + ".weights", false);
+            model = findDataFile("dnn/darknet/" + name + ".weights", required);
```
alalek: Lets remove unclear `required` parameter (with meaning of `requiredWeights`).
Just remove `", false"` from here (weights are not "optional" testdata - they are stored in "opencv_extra").

dkurt: I agree, lest just emit the last argument (so make it `true` by default value):

```
            model = findDataFile("dnn/darknet/" + name + ".weights");
```

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -556,20 +556,21 @@ namespace cv {
...
                         int stride = getParam<int>(layer_params, "stride", 1);
                         int filters = getParam<int>(layer_params, "filters", -1);
                         bool batch_normalize = getParam<int>(layer_params, "batch_normalize", 0) == 1;
                         int flipped = getParam<int>(layer_params, "flipped", 0);
                         if (flipped == 1)
                             CV_Error(cv::Error::StsNotImplemented, "Transpose the convolutional weights is not implemented");
 
-                        // correct the strange value of pad=1 for kernel_size=1 in the Darknet cfg-file
-                        if (kernel_size < 3) pad = 0;
+                        if (pad == 1)
```
dkurt: What if user specified non zero value of `pad` different from `1`? So let's keep `if (pad)`

ashishkrshrivastava:
> Well done!

Thanks

PR #16372, opened 2020-01-17
(3.4) fix lint errors and warnings ins Android samples

alalek:
backport #16350

```
force_builders=Android pack
```


alalek:
:+1:

PR #16375, opened 2020-01-17
Reduce LLC loads, stores and multiplies on MulTransposed - 8% faster

ChipKerchner:
Reduce LLC loads, stores and multiplies by 2x on MulTransposed - 8% faster

### DIFF
--- a/modules/core/src/matmul.simd.hpp
+++ b/modules/core/src/matmul.simd.hpp
@@ -2042,6 +2042,20 @@ MahalanobisImplFunc getMahalanobisImplFunc(int depth)
 *                                        MulTransposed                                   *
 \****************************************************************************************/
 
+#ifdef CV_SIMD_64F
+template<typename T, typename U>
+struct is_same 
+{
+    static const bool value = false; 
+};
+
+template<typename T>
+struct is_same<T,T>  //specialization
+{ 
+   static const bool value = true; 
+};
+#endif
+
 template<typename sT, typename dT> static void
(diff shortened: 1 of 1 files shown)


ChipKerchner:
Last failure doesn't seem to be related to my check-in

modules/core/src/matmul.simd.hpp
```diff
@@ -2086,22 +2104,41 @@ MulTransposedR(const Mat& srcmat, const Mat& dstmat, const Mat& deltamat, double
 
             for( j = i; j <= size.width - 4; j += 4 )
             {
-                double s0 = 0, s1 = 0, s2 = 0, s3 = 0;
-                const sT *tsrc = src + j;
+#if CV_SIMD_64F
+                if (is_same<sT, double>::value && is_same<dT, double>::value)
```
alalek: Please try type traits from OpenCV:
```
DataType<sT>::depth == CV_64F && DataType<dT>::depth == CV_64F
```

ChipKerchner: Done

terfendail:
The change looks fine for me.
However I can't reproduce the performance gain. There is no dedicated performance test for the function, so I've used accuracy test evaluation time. For SSE3 baseline I've got performance gain of about 3 per cent, for SSE4_2 baseline I've got performance degradation of 2-3 per cent. Both results look like random fluctuation rather than stable performance change.

ChipKerchner:
The performance gains were measured on a Power9 VSX system.  It is possible that the gains only show up on a non-Intel platform since there are different stalls related to memory access for this platform.

PR #16377, opened 2020-01-17
dnn(onnx): handle unaligned access in ONNX importer

### LINKED ISSUE #16373: Bus fault when importing ONNX model on Android
- OpenCV => 1a74de1f9c7a3e1c8409a8b8e20c1f992f614e4a (4.2)
- Operating System / Platform => Android 4.4.4 amv7l
- Compiler => Android NDK r18b (clang, Android API level = 16)
- Optmization => -DCMAKE_BUILD_TYPE=Release (-Oz)

##### Detailed description

I'm getting a SIGBUS when loading an ONNX model on Android (armv7l), this problem only occurs when building with optimizations enabled, -O0 doesn't trigger the issue. I've nailed down the problem to these specific lines:

https://github.com/opencv/opencv/blob/d85e67d3ecf425f0ffbf643e7fa8323e289bfff0/modules/dnn/src/onnx/onnx_importer.cpp#L149-L153

The problem is that the pointer returned by std::string::c_str() doesn't have any guarantee that it is 8 bytes aligned, and sure enough when inspecting with GDB the ```val``` pointer isn't aligned, thus ```src```.

I've workaround this by allocating a intermediate buffer and copying the data pointed by ```val```:

```c++
{
    char* val = const_cast<char*>(tensor_proto.raw_data().c_str());
    int size = tensor_proto.raw_data().size();
    int64_t* src = new int64_t[size];
    memcpy(src, val, size);
    convertInt64ToInt32(src, dst, blob.total());
    delete[] src;
}
```

I don't known if this is an good solution, please let me known.

##### Steps to reproduce

Build for Android armv7l, NDK r18b with optimizations. Call ```cv::dnn::readNetFromONNX()``` with a valid ONNX file.

Thanks,
Augusto.

alalek:
resolves #16373

### DIFF
--- a/modules/dnn/src/onnx/onnx_importer.cpp
+++ b/modules/dnn/src/onnx/onnx_importer.cpp
@@ -147,8 +147,18 @@ Mat getMatFromTensor(opencv_onnx::TensorProto& tensor_proto)
         }
         else
         {
-            char* val = const_cast<char*>(tensor_proto.raw_data().c_str());
-            int64_t* src = reinterpret_cast<int64_t*>(val);
+            const char* val = tensor_proto.raw_data().c_str();
+            // Aligned pointer is required: https://github.com/opencv/opencv/issues/16373
+            // this doesn't work: typedef int64_t CV_DECL_ALIGNED(1) unaligned_int64_t;
+            AutoBuffer<int64_t, 16> aligned_val;
+            if (!isAligned<sizeof(int64_t)>(val))
+            {
+                size_t sz = tensor_proto.raw_data().size();
+                aligned_val.allocate(divUp(sz, sizeof(int64_t)));
+                memcpy(aligned_val.data(), val, sz);
+                val = (const char*)aligned_val.data();
+            }
+            const int64_t* src = reinterpret_cast<const int64_t*>(val);
             convertInt64ToInt32(src, dst, blob.total());
         }
     }


modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -147,8 +147,18 @@ Mat getMatFromTensor(opencv_onnx::TensorProto& tensor_proto)
...
+            // this doesn't work: typedef int64_t CV_DECL_ALIGNED(1) unaligned_int64_t;
+            AutoBuffer<int64_t, 16> aligned_val;
+            if (!isAligned<sizeof(int64_t)>(val))
+            {
+                size_t sz = tensor_proto.raw_data().size();
+                aligned_val.allocate(divUp(sz, sizeof(int64_t)));
+                memcpy(aligned_val.data(), val, sz);
+                val = (const char*)aligned_val.data();
+            }
+            const int64_t* src = reinterpret_cast<const int64_t*>(val);
```
dkurt: Shall we do the same for floats data?

https://github.com/opencv/opencv/blob/55d54b56bfaf32991a45520855e46ee67940f37c/modules/dnn/src/onnx/onnx_importer.cpp#L129-L130

alalek: `Mat::copyTo()` uses type-agnostic `memcpy()` internally.
- `.copyTo()` is safe
- `.convertTo()` is not safe.

PR #15929, opened 2019-11-17
Update seamless_cloning.cpp

sturkmen72:
### This pullrequest changes

resolves #15927

PR #16392, opened 2020-01-20
No further issue regarding channels

### LINKED ISSUE #15927: could we include 4-channels images, too? i.e. 4D alpha images?
https://github.com/opencv/opencv/blob/1acadd363b0d0ffcdabac8af3196cb65bef426b1/modules/photo/src/seamless_cloning.cpp#L54
reason:  all images with 4 channels (alpha channels) are required to be converted to 3 channels on the client side, then back to 3 channels just for API compatibility purposes.

Lalit-garg:
resolves #15927


alalek:
These is already fix for that #15929
It will be ported into "master" during next "Merge-3.4" activity.

Lalit-garg:
Hi alalek,
Basically I am a newbie,and wanted to contribute in open source ,to increase skills.
Can you suggest something for me doing in OpenCv.

PR #16409, opened 2020-01-22
G-API: Fix linkage in standalone mode

smirnov-alexey:
This PR fixes linkage of G-API library to some target in standalone mode.

The solution is pretty simple, yet maybe not so trim. Alternatively, we can move definitions and usage of functions from `gstreaming.cpp` under `#if !defined` in other files.
```
```


smirnov-alexey:
@rgarnov, @dmatveev, please, review

dmatveev:
Seems I forgot to remove that. Thanks!

Issue #15403, opened 2019-08-27
Object detection sample code does not work with YOLOv3 model


rgov:
- OpenCV => 4.1.0
- Operating System / Platform => macOS 10.14.6
- Compiler => Python 3

##### Detailed description

The `object_detection.py` script barfs when using YOLOv3 weights and config.

##### Steps to reproduce

Download weights and config:

```
wget https://pjreddie.com/media/files/yolov3.weights  
wget https://raw.githubusercontent.com/pjreddie/darknet/master/cfg/yolov3.cfg
```

Run the example code:

```
$ python object_detection.py --model yolov3.weights --config yolov3.cfg
Traceback (most recent call last):
  File "object_detection.py", line 71, in <module>
    net = cv.dnn.readNet(cv.samples.findFile(args.model), cv.samples.findFile(args.config), args.framework)
cv2.error: OpenCV(4.1.0) /Users/travis/build/skvark/opencv-python/opencv/modules/dnn/src/darknet/darknet_io.cpp:509: error: (-215:Assertion failed) separator_index < line.size() in function 'ReadDarknetFromCfgStream'
```

ManzarIMalik:
@rgov, I am having this code error in object detection sample code file. How did you fix this?

`if args.async: line 234
                ^
SyntaxError: invalid syntax`

rgov:
@ManzarIMalik My guess is that you are running Python 3.7 or newer, where the `async` keyword is reserved. You will have to change the variable name used by the sample code, or use an earlier version of the sample code (a few commits back) that does not provide the async feature.

Please post back if that works.

ManzarIMalik:
@rgov, Yes it works now. Thank you :)

zhoujinhai:
@rgov， Hello, I am having this error, Have you ever met?
```
 File "object_detection.py", line 262, in processingThreadBody
   outs = net.forward(outNames)

cv2.error: OpenCV(4.1.0) C:\projects\opencv-python\opencv\modules\dnn\src\layers
\concat_layer.cpp:95: error: (-201:Incorrect size of input array) Inconsistent s
hape for ConcatLayer in function 'cv::dnn::ConcatLayerImpl::getMemoryShapes'
```

zhoujinhai:
@rgov
The code is run, but the result is bad, What parameters do you need to adjust?
```python
python detection.py --input 000035.jpg --model yolov3.weights --config yolov3.cfg --width 416 --height 416 --classes coco_classes.txt
```

[...]

PR #16414, opened 2020-01-23
samples(dnn): avoid 'async' keyword (Python 3.7+)

alalek:
```
File "object_detection.py", line 234
    if args.async:
                ^
SyntaxError: invalid syntax
```

https://github.com/opencv/opencv/issues/15403#issuecomment-526911163


asmorkalov:
:+1:

PR #7705, opened 2016-11-22
Improvement of sequential connected components Wu's algorithm and pro…

MicheleCancilla:
### This pullrequest changes

With this Pull Request we provide the following improvements on “cv::connectedComponents” and on “cv::connectedComponentsWithStats” functions: 
<ul>
<li>We improve the speed of Wu’s algorithm (sequential version) providing a better implementation, both for 4 and 8-way connectivity.</li>
<li>We provide a parallel CPU implementation of both Wu and Grana’s algorithms that are the ones currently included in the OpenCV library. These parallel versions are automatically chose by the functions when the TBB variable is defined. The performance are shown in the charts below where acronyms are used to distinguish the different implementation/version of the algorithms. An explanation of the datasets used and how the tests were performed can be found <a href="https://github.com/prittt/YACCLAB/blob/master/README.md">here</a>. </li>
</ul>
To conclude, this pull request reply to RFC #7270 and take into account the issue #7607 providing the same solution proposed by @sokian with pull request #7608 that is open yet. 
<br/><br/>
<b>SAUF_OLD:</b> is the old implementation of the sequential version of the Wu’s algorithm;
<br/>
<b>SAUF_NEW:</b> is the implementation of the sequential version of the Wu’s algorithm proposed with this pull request;
<br/>
<b>SAUF_PAR:</b> is the parallel version of the Wu’s algorithm introduced by this pull request; 
<br/>
<b>BBDT_SEQ:</b> is the current sequential version of the Grana’s algorithm (currently included in the OpenCV);
<br/>
<b>BBDT_PAR:</b> is the parallel version of the Grana’s algorithm introduced by this pull request;
<br/>
<br/>
<table border="0">
<tr><p align="center">Comparison of  Wu’s algorithms with 8-way connectivity (without computing statistics)</p></tr>
<tr>
 <td>3DPeSfigure1</td>
 <td>Medicalfigure2</td>
 <td>Tobacco800figure3</td>
</tr>
</table><br/>
<table border="0">
<tr><p align="center">Comparison of  Wu’s algorithms with 4-way connectivity (without computing statistics)</p></tr>
<tr>
 <td>3DPeSfigure1</td>
 <td>Medicalfigure2</td>
 <td>Tobacco800figure3</td>
</tr>
</table><br/>
<table border="0">
<tr><p align="center">Comparison of  Grana’s algorithms without computing statistics</p></tr>
<tr>
 <td>3DPeSfigure1</td>
 <td>Medicalfigure2</td>
[...]

PR #16415, opened 2020-01-23
Fix bug 16410 and add test

### LINKED ISSUE #16410: Crash in connectedComponentsWithStats
OpenCV 4.1.1
OS: OSX
Compiler: clang

I found a crash in `connectedComponentsWithStats` on very specific images. It is a bit difficult reproduce because it depends on the number of threads (16 on my machine).
See the code below to reproduce systematically.

From my understanding, the problem comes from the colliding labels.
I will try to provide a fix this week.

```
cv::Mat createCrashMat(int numThreads) {
    const int h = numThreads * 4 * 2 + 8;
    const double nParallelStripes = std::max(1, std::min(h / 2, numThreads * 4));
    const int w = 4;

    const int nstripes = cvRound(nParallelStripes <= 0 ? h : MIN(MAX(nParallelStripes, 1.), h));
    const cv::Range stripeRange(0, nstripes);
    const cv::Range wholeRange(0, h);

    cv::Mat m(h, w, CV_8U);
    m = 0;

    // Look for a range that starts with odd value and ends with even value
    cv::Range bugRange;
    for (int s = stripeRange.start; s < stripeRange.end; s++) {
        cv::Range sr(s, s + 1);
        cv::Range r;
        r.start = (int) (wholeRange.start +
                         ((uint64) sr.start * (wholeRange.end - wholeRange.start) + nstripes / 2) / nstripes);
        r.end = sr.end >= nstripes ?
                    wholeRange.end :
                    (int) (wholeRange.start +
                           ((uint64) sr.end * (wholeRange.end - wholeRange.start) + nstripes / 2) / nstripes);

        if (r.start > 0 && r.start % 2 == 1 && r.end % 2 == 0 && r.end >= r.start + 2) {
            bugRange = r;
            break;
        }
    }

    if (bugRange.empty()) {
        std::cout << "Could not find a buggy range\n";
        return m;
    }

    // Fill in bug Range
    for (int x = 1; x < w; x++) {
        m.at<char>(bugRange.start - 1, x) = 1;
    }

    m.at<char>(bugRange.start + 0, 0) = 1;
    m.at<char>(bugRange.start + 0, 1) = 1;
    m.at<char>(bugRange.start + 0, 3) = 1;
    m.at<char>(bugRange.start + 1, 1) = 1;
    m.at<char>(bugRange.start + 2, 1) = 1;
    m.at<char>(bugRange.start + 2, 3) = 1;
    m.at<char>(bugRange.start + 3, 0) = 1;
    m.at<char>(bugRange.start + 3, 1) = 1;

    return m;
}
```

arnaudbrejeon:
resolves #16410

Here is what is happening. In order to do the parallelization, the image is split in stripes and labelling is done on each stripe. In order for stripes to be independent, each stripe starting at height `y` and with width `w` gets a startLabel defined as:
`startLabel = (y+1)/2 * (w+1)/2 + 1`

The formula comes from the fact that for an image of height `h` and width `w`,  an upper bound of the number of labels is: `(h+1)/2 * (w+1)/2 + 1`

The startLabels are too close to each other in some cases and create labels overlapping. Here is why:
let's say a stripe starts in `y0` and finishes in `y1`, we can write
`y0 = 2*y + 1` and `y1 = 2*y + 1 + h`
The maximum number of labels is `(h+1)/2 * (w+1)/2 + 1`
The number of labels for the stripe is: 
`( (2*y+1+h+1)/2 * (w+1)/2 + 1  ) - ( (2*y+1+1)/2 * (w+1)/2 + 1 )` which is equal to
`((h+2)/2 - 1) * (w+1)/2 = (h/2) * (w+1)/2`

If `h` is odd, then the number of allowed labels in the stripe is not large enough.

In order to make sure there is enough labels for each stripe, I suggest we use the formula:
`startLabel = (y+2)/2 * (w+1)/2 + 1`

### DIFF
--- a/modules/imgproc/src/connectedcomponents.cpp
+++ b/modules/imgproc/src/connectedcomponents.cpp
@@ -271,6 +271,11 @@ namespace cv{
     template<typename LabelT, typename PixelT, typename StatsOp = NoOp >
     struct LabelingWuParallel{
 
+        template <typename LT>
+        static LT stripeFirstLabel8Connectivity(int y, int w) {
+            return LT((y+2)/2) * LT((w+1)/2) + 1;
+        }
+
         class FirstScan8Connectivity : public cv::ParallelLoopBody{
             const cv::Mat& img_;
             cv::Mat& imgLabels_;
@@ -288,7 +293,7 @@ namespace cv{
                 int r = range.start;
                 chunksSizeAndLabels_[r] = range.end;
 
-                LabelT label = LabelT((r + 1) / 2)  * LabelT((imgLabels_.cols + 1) / 2) + 1;
+                LabelT label = stripeFirstLabel8Connectivity<LabelT>(r, imgLabels_.cols);
 
                 const LabelT firstLabel = label;
                 const int w = img_.cols;
@@ -615,7 +620,7 @@ namespace cv{
                 mergeLabels8Connectivity(imgLabels, P, chunksSizeAndLabels);
 
                 for (int i = 0; i < h; i = chunksSizeAndLabels[i]){
-                    flattenL(P, int((i + 1) / 2) * int((w + 1) / 2) + 1, chunksSizeAndLabels[i + 1], nLabels);
+                    flattenL(P, stripeFirstLabel8Connectivity<int>(i, w), chunksSizeAndLabels[i + 1], nLabels);
                 }
             }
             else{
--- a/modules/imgproc/src/connectedcomponents.cpp
+++ b/modules/imgproc/src/connectedcomponents.cpp
@@ -615,7 +615,7 @@ namespace cv{
 
             //Array used to store info and labeled pixel by each thread.
             //Different threads affect different memory location of chunksSizeAndLabels
-            std::vector<int> chunksSizeAndLabels(h);
+            std::vector<int> chunksSizeAndLabels(roundUp(h, 2));
 
             //Tree of labels
             std::vector<LabelT> P_(Plength, 0);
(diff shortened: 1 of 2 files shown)


asmorkalov:
Hello @arnaudbrejeon thanks for the bug report and fix!

alalek:
@arnaudbrejeon Well done! Thanks for the reproducer!

Reference on original patch: #8789 (#7705).


> If h is odd

Your guess seems correct.

Perhaps "odd" case is not expected at all by processing code:
- `chunksSizeAndLabels_[r] = range.end;`
- `chunksSizeAndLabels_[startR + 1] = label - firstLabel;`

and finally:
- `const double nParallelStripes = std::max(1, std::min(h / 2, getNumThreads()*4));`
- with `L.rows / nThreads >= 2` limit:
```
    //Run parallel labeling only if the rows of the image are at least twice the number of available threads
    const bool is_parallel = currentParallelFramework != NULL && nThreads > 1 && L.rows / nThreads >= 2;
```

Last statements trying to guarantee that subtask size is not less than 2 rows (or may be even "even").
Anyway this approach is wrong because `nStripes` parameter of `parallel_for()` is just a hint and it is not accurate at all (like any result of floating-point calculations).

There is similar problem with 4 connectivity stripes: `(r * imgLabels_.cols + 1) / 2 + 1`
- r=3, cols=3 => 6
- r=4, cols=3 => 7
- r=6, cols=3 => 10
- `1 0 1` => 2 labels are required for 3 cols x 1 row
- `1 0 1 | 0 1 0 | 1 0 1` => up to 5 labels for 3 x 3 stripe (but 10-6=4)

---

Q: Why are we observing crash instead of incorrect result?

A: Uninitialized array with garbage after the first element:
```
            //Tree of labels
            LabelT *P = (LabelT *)cv::fastMalloc(Plength * sizeof(LabelT));
            //First label is for background
            P[0] = 0;
```

---

There is one more similar place in `LabelingGranaParallel` (8 connectivity):
```
LabelT label = LabelT((r + 1) / 2)  * LabelT((imgLabels_.cols + 1) / 2) + 1;
```

---

Additional commits will be pushed shortly.

PR #16421, opened 2020-01-24
Local memory pool experiment

mshabunin:
### This pullrequest changes

Introduces `cv::util::BufferArea` class automating creation of structured memory areas for local memory buffers.
<cut>

There are two modes: _safe_ and _unsafe_. In safe mode each allocation is performed immediately, in unsafe mode (default) single large buffer will be allocated during the call to commit and all pointers will be set to appropriate locations inside this buffer. Safe mode can be forced at runtime by constructor parameter for one specific object or using an environment variable for all objects (`OPENCV_BUFFER_AREA_ALWAYS_SAFE`). Safe mode will also be forced for sanitizer-enabled builds.

Single buffer split into several smaller ones is used in several algorithms to reduce memory allocations for better performance. However this approach does not allow to catch OOB access using valgrind or address sanitizer and is prone to errors in address arithmetic calculations. Proposed class allows fast but unsafe mode by default and slower but safer mode for instrumentation and debugging. 

Example of use:
```.cpp
int * buf1 = 0;
double * buf2 = 0;
cv::util::BufferArea area;
area.allocate(buf1, 200); // buf1 = new int[200];
area.allocate(buf2, 1000, 64); // buf2 = new double[1000]; - aligned by 64
area.commit();
// NOTE: area stores references to buf1 and buf2 and effectively owns them
```

Example of old code doing the same work :
```.cpp
char * buf = malloc(200 * sizeof(int) + 1000 * sizeof(double) + some_alignment_reserve);
int * buf1 = (int*)alignPtr(buf, sizeof(int));
double * buf2 = (double*)alignPtr((char*)buf1 + 200 * sizeof(int), 64);
//...
free(buf);
```

This pull request also proposes refactoring of StereoBM algorithm using introduced class. **Note:** buffers aligning to 64 bytes has been disabled during refactoring.

TODO list:

- [x] Test
- [x] Documentation
- [x] Use in real algorithm

```
buildworker:Custom=linux-1
build_image:Custom=centos:7
```

### DIFF
--- a/modules/calib3d/src/stereobm.cpp
+++ b/modules/calib3d/src/stereobm.cpp
@@ -48,8 +48,10 @@
 #include "precomp.hpp"
 #include <stdio.h>
 #include <limits>
+#include <vector>
 #include "opencl_kernels_calib3d.hpp"
 #include "opencv2/core/hal/intrin.hpp"
+#include "opencv2/core/utils/buffer_area.private.hpp"
 
 namespace cv
 {
@@ -85,6 +87,19 @@ struct StereoBMParams
     Rect roi1, roi2;
     int disp12MaxDiff;
     int dispType;
+
+    inline bool useShorts() const
+    {
+        return preFilterCap <= 31 && SADWindowSize <= 21;
+    }
+    inline bool useFilterSpeckles() const
+    {
+        return speckleRange >= 0 && speckleWindowSize > 0;
+    }
+    inline bool useNormPrefilter() const
+    {
+        return preFilterType == StereoBM::PREFILTER_NORMALIZED_RESPONSE;
+    }
 };
 
 #ifdef HAVE_OPENCL
@@ -110,10 +125,10 @@ static bool ocl_prefilter_norm(InputArray _input, OutputArray _output, int winsi
 }
 #endif
 
-static void prefilterNorm( const Mat& src, Mat& dst, int winsize, int ftzero, uchar* buf )
+static void prefilterNorm( const Mat& src, Mat& dst, int winsize, int ftzero, int *buf )
 {
     int x, y, wsz2 = winsize/2;
-    int* vsum = (int*)alignPtr(buf + (wsz2 + 1)*sizeof(vsum[0]), 32);
+    int* vsum = buf + (wsz2 + 1);
     int scale_g = winsize*winsize/8, scale_s = (1024 + scale_g)/(scale_g*2);
     const int OFS = 256*5, TABSZ = OFS*2 + 256;
     uchar tab[TABSZ];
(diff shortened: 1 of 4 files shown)


dzhura:
stereoBM throws segfault when the input images are of size approx. 6000 by 9000 pixels. It caused by overflow of bufSize at the line [1211](https://github.com/opencv/opencv/blob/5d01821d9f4d57921dad82045cc9819cfabcfe15/modules/calib3d/src/stereobm.cpp#L1211)

Does this patch solves the issue?

mshabunin:
@dzhura , I think it can solve the issue, especially if safe mode will be used for local buffer. Though we will not know until we try it. Do you have minimal reproducer app?

dzhura:
@mshabunin

```
#include <opencv2/calib3d.hpp>

int main()
{
	cv::Mat left(8500, 6000, CV_8UC1, cv::Scalar(0));
	cv::Mat right(8500, 6000, CV_8UC1, cv::Scalar(0));

	cv::Ptr<cv::StereoBM> matcher = cv::StereoBM::create(256, 31);

	cv::Mat disparities;
	matcher->compute(left, right, disparities); // kaboom

	return 0;
}
```

There is no segfalut when the number of rows is 8000 or less

mshabunin:
Your sample does not crash with this patch (on 64-bit system). Valgrind does not show errors too.

dzhura:
Thank you! I look forward to this patch to be merged

modules/core/src/local_pool.cpp
```diff
@@ -0,0 +1,55 @@
...
+#define LOCAL_POOL_DEFAULT_MODE false
+#endif
+
+static bool CV_LOCAL_POOL_OVERRIDE_SAFE_MODE =
+    cv::utils::getConfigurationParameterBool("OPENCV_LOCAL_POOL_ALWAYS_SAFE", LOCAL_POOL_DEFAULT_MODE);
+
+namespace cv { namespace utils {
+
+LocalPool::LocalPool(bool safe_)
+    : safe(safe_ || CV_LOCAL_POOL_OVERRIDE_SAFE_MODE),
```
alalek: We need a way to turn off this mode using environment (may be useful if problem can't be reproduced with switched ON mode).

modules/core/include/opencv2/core/utils/local_pool.private.hpp
```diff
@@ -0,0 +1,176 @@
...
+    @note In safe mode allocation will be performed immediatly.
+    */
+    template <typename T>
+    void allocate(T*&ptr, size_t count, size_t alignment = sizeof(T))
+    {
+        CV_Assert(ptr == NULL);
+        CV_Assert(count > 0);
+        CV_Assert(alignment > 0);
+        CV_Assert(alignment % sizeof(T) == 0);
+        blocks.push_back(new Block<T>(ptr, count, alignment));
```
alalek: > new

Main purpose of original approach is to call "new" only once.
Current approach calls "new" several times in all modes.

modules/core/src/local_pool.cpp
mshabunin: We assume that nobody will create an object with safe mode enabled `LocalPool::LocalPool(true)`, it is available mainly for testing purpose.

modules/core/include/opencv2/core/utils/local_pool.private.hpp
```diff
@@ -0,0 +1,176 @@
...
+        size_t count;
+        size_t alignment;
+    };
+
+public:
+    /** @brief Class constructor.
+
+    @param safe_ Enable _safe_ operation mode, each allocation will be performed independently.
+    */
+    LocalPool(bool safe_ = false);
```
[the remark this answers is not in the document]
mshabunin: I think this name is quite descriptive in context of instrumentation and debugging.

PR #16436, opened 2020-01-26
dnn(darknet-importer): add grouped convolutions, sigmoid, swish, scale_channels

### LINKED ISSUE #15987: Feature-request: EfficientNet as backbone for Yolo v3 Detector:
@dkurt Hi,

Feature-request: support **EfficientNet** as backbone for Yolo v3 Detector (this is just part of the changes for newer versions of Yolo):

* EfficientNetB0-Yolo - 45.5% mAP@0.5 - 3.7 BFLOPS - 1.5x less flops than Yolov3-tiny, but +12.1 higher mAP@0.5
* YOLOv3-tiny - 33.1% mAP@0.5 - 5.56 BFLOPS
* YOLOv3-416 - 55.3% mAP@0.5  - 65.86 BFLOPS

EfficientNetb0-Yolo:
* cfg-file: https://github.com/AlexeyAB/darknet/blob/master/cfg/enet-coco.cfg
* weights file: https://drive.google.com/file/d/1FlHeQjWEQVJt0ay1PVsiuuMzmtNyv36m/view
* coco.names file: https://raw.githubusercontent.com/AlexeyAB/darknet/master/cfg/coco.names

`./darknet detector test cfg/coco.data enet-coco.cfg enetb0-coco_final.weights dog.jpg`

----

* **45.5% mAP@0.5** on MS COCO: https://github.com/WongKinYiu/PartialResidualNetworks#partial-residual-networks
* **3.7 BFlops** (1.85B FMA in terms of google flops = **1.85 BFlops**)
*  2019 ICCV Workshop on Low-Power Computer Vision: https://github.com/WongKinYiu/PartialResidualNetworks/blob/master/pdf/iccvw-prn.pdf

----

EfficientNet B0 - B7:
* paper: https://arxiv.org/abs/1905.11946v2

| params | flops |
|-|-|
|  | |

----

Required changes:

* swish-activation: https://github.com/AlexeyAB/darknet/blob/dd4dcf20b4886f4cfedb82e6d205a93c3a4c1034/cfg/enet-coco.cfg#L65

* groups for Grouped and Depth-wise convolution: https://github.com/AlexeyAB/darknet/blob/dd4dcf20b4886f4cfedb82e6d205a93c3a4c1034/cfg/enet-coco.cfg#L48

* `[avgpool]` - Global average pooling layer - for squeeze and excitation blocks (as in MobileNet v3): https://github.com/AlexeyAB/darknet/blob/dd4dcf20b4886f4cfedb82e6d205a93c3a4c1034/cfg/enet-coco.cfg#L112

* `[scale_channels]` - layer for squeeze and excitation blocks - it just multiply each of WxH element in channel-C in one layer (size WxHxC) by value from channel-C in another layer (size: 1x1xC) - source code: https://github.com/AlexeyAB/darknet/blob/dd4dcf20b4886f4cfedb82e6d205a93c3a4c1034/src/scale_channels_layer.c#L61-L74
https://github.com/AlexeyAB/darknet/blob/dd4dcf20b4886f4cfedb82e6d205a93c3a4c1034/cfg/enet-coco.cfg#L129-L130

image

YashasSamaga:
resolves #15987

**Merge with extra:** https://github.com/opencv/opencv_extra/pull/711

### This pull request changes
- adds support for grouped convolutions
- adds new activations
  - sigmoid
  - swish
- adds support for scale channels

<cut/>

**Prediction using OpenCV CPU backend:**

Screenshot from 2020-03-01 22-47-12

**Darknet Reference:**
```
tvmonitor: 44%
dog: 78%
cat: 27%
bicycle: 58%
truck: 89%
car: 33%
```

predictions

<cut/>

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/darknet/darknet_io.cpp
+++ b/modules/dnn/src/darknet/darknet_io.cpp
@@ -217,24 +217,6 @@ namespace cv {
                     fused_layer_names.push_back(last_layer);
                 }
 
-                void setDropout()
-                {
-                    cv::dnn::LayerParams dropout_param;
-                    dropout_param.name = "Dropout-name";
-                    dropout_param.type = "Dropout";
-                    darknet::LayerParameter lp;
-
-                    std::string layer_name = cv::format("dropout_%d", layer_id);
-                    lp.layer_name = layer_name;
-                    lp.layer_type = dropout_param.type;
-                    lp.layerParams = dropout_param;
-                    lp.bottom_indexes.push_back(last_layer);
-                    last_layer = layer_name;
-                    net->layers.push_back(lp);
-                    layer_id++;
-                    fused_layer_names.push_back(last_layer);
-                }
-
                 void setReLU()
                 {
                     cv::dnn::LayerParams activation_param;
@@ -695,6 +677,7 @@ namespace cv {
 
                         CV_Assert(kernel_size > 0 && filters > 0);
                         CV_Assert(tensor_shape[0] > 0);
+                        CV_Assert(tensor_shape[0] % groups == 0);
 
                         setParams.setConvolution(kernel_size, padding, stride, filters, tensor_shape[0],
                             groups, batch_normalize);
@@ -718,10 +701,6 @@ namespace cv {
                         tensor_shape[0] = output;
                         tensor_shape[1] = 1;
                         tensor_shape[2] = 1;
-                    }
-                    else if (layer_type == "dropout")
-                    {
-                        setParams.setDropout();
                     }
                     else if (layer_type == "maxpool")
                     {
@@ -937,7 +916,6 @@ namespace cv {
                             int sizes_weights[] = { filters, total(tensor_shape) };
                             weightsBlob.create(2, sizes_weights, CV_32F);
                         }
-
                         CV_Assert(weightsBlob.isContinuous());
 
                         cv::Mat meanData_mat(1, filters, CV_32F);	// mean
--- a/modules/dnn/src/darknet/darknet_io.cpp
+++ b/modules/dnn/src/darknet/darknet_io.cpp
@@ -217,49 +217,30 @@ namespace cv {
                     fused_layer_names.push_back(last_layer);
                 }
 
-                void setReLU()
+                void setActivation(String type)
                 {
                     cv::dnn::LayerParams activation_param;
-                    activation_param.set<float>("negative_slope", 0.1f);
-                    activation_param.name = "ReLU-name";
-                    activation_param.type = "ReLU";
+                    if (type == "relu")
+                    {
+                        activation_param.set<float>("negative_slope", 0.1f);
+                        activation_param.type = "ReLU";
+                    }
+                    else if (type == "swish")
+                    {
+                        activation_param.type = "Swish";
+                    }
+                    else if (type == "logistic")
+                    {
+                        activation_param.type = "Sigmoid";
+                    }
+                    else
+                    {
+                        CV_Error(cv::Error::StsParseError, "Unsupported activation: " + type);
+                    }
+
+                    std::string layer_name = cv::format("%s_%d", type.c_str(), layer_id);
 
                     darknet::LayerParameter lp;
-                    std::string layer_name = cv::format("relu_%d", layer_id);
-                    lp.layer_name = layer_name;
-                    lp.layer_type = activation_param.type;
-                    lp.layerParams = activation_param;
-                    lp.bottom_indexes.push_back(last_layer);
-                    last_layer = layer_name;
-                    net->layers.push_back(lp);
-
-                    fused_layer_names.back() = last_layer;
-                }
-
-                void setSwish()
-                {
-                    cv::dnn::LayerParams activation_param;
-                    activation_param.type = "Swish";
-
-                    darknet::LayerParameter lp;
-                    std::string layer_name = cv::format("swish_%d", layer_id);
-                    lp.layer_name = layer_name;
-                    lp.layer_type = activation_param.type;
-                    lp.layerParams = activation_param;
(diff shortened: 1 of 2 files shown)


dkurt:
Please add single layer tests for each of newly added layers. You may take a look at https://github.com/opencv/opencv/pull/16358 as a reference.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -191,6 +211,42 @@ namespace cv {
...
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+
+                    fused_layer_names.back() = last_layer;
+                }
+
+                void setLogistic()
+                {
+                    cv::dnn::LayerParams activation_param;
+                    activation_param.name = "Sigmoid-name";
```
dkurt: Please check if we really need to specify names here. Due they are not unique, maybe they are not used at all?

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -172,6 +174,24 @@ namespace cv {
...
+                    std::string layer_name = cv::format("dropout_%d", layer_id);
+                    lp.layer_name = layer_name;
+                    lp.layer_type = dropout_param.type;
+                    lp.layerParams = dropout_param;
+                    lp.bottom_indexes.push_back(last_layer);
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+                    layer_id++;
+                    fused_layer_names.push_back(last_layer);
+                }
```
dkurt: @YashasSamaga, thanks for PR! Do you mind to let @ashishkrshrivastava to integrate Dropout layer with https://github.com/opencv/opencv/pull/16438? So the rest of layers will be yours :)

YashasSamaga: That'll be great. I am anyway having trouble finding free time to complete the PR. I have been very busy for the last two weeks. It'll be until next Wednesday before I can make progress here.

YashasSamaga:
@dkurt For the convolution and activation tests, should I create a new cfg files for swish, logistic and grouped convolution OR add more layers to existing `convolutional.cfg`?

dkurt:
@YashasSamaga, I think the best way if to check existing tests and replace some of duplicated activation to new ones.

modules/dnn/src/darknet/darknet_io.cpp
YashasSamaga: They are not used.

The actual name and type are set in `darknet::LayerParameter`. This name and type is used in `darknet_importer.cpp` to add layers using `Net::addLayer`. `Net::addLayer` invokes `LayerData` constructor which accepts name, type and `Net::LayerParams`. In the constructor, it overwrites the `name` and `type` field of `Net::LayerParams` with the name and type that were passed as arguments.

Hence, these names `LayerType-name` which are used through the darknet importer are never used.

YashasSamaga:
`scale_channels` does not work with batches.

https://github.com/opencv/opencv/blob/3efa78311a837b7a06d81f2f359ca98e6d26609a/modules/dnn/src/layers/scale_layer.cpp#L86-L94

`weights` is reshaped (assuming that batch size is one?).

https://github.com/opencv/opencv/blob/3efa78311a837b7a06d81f2f359ca98e6d26609a/modules/dnn/src/layers/scale_layer.cpp#L98-L104

This code ignores the batch dimension and tries to match `numWeights` (i.e. `weights.total()`) to a subrange in the input tensor.

Example:
```
input shape: 2 3 3 3
weights input: 2 3 1 1
```

`weights` is reshaped to `6 1`. The matching code tries to look for a subrange in the input (`2 3 3 3`) which totals to 6 but it fails to find any.

I can tweak the `ScaleLayer` to make it behave like Darknet's scale channels but I fear that it would break things and other backends may not be setup to behave like Darknet's scale channels.

Or I can add a new layer which behaves like Darknet's scale channels.

dkurt:
@YashasSamaga, let's open a new issue for that observation. So for Scale layer there will be a new test with batch size 1 but for Swish and Logistic - let's try to include them to existing ones.

modules/dnn/src/darknet/darknet_io.cpp
YashasSamaga: @dkurt Should I remove them? If yes, for all the layers which aren't part of this PR or only the layers which are part of this PR?

dkurt: Maybe just for newly created layers, thanks!

YashasSamaga:
I have used a different branch name in opencv_extra PR. How can the CI be informed about it?

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -234,6 +236,40 @@ namespace cv {
...
+                    lp.layer_type = activation_param.type;
+                    lp.layerParams = activation_param;
+                    lp.bottom_indexes.push_back(last_layer);
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+
+                    fused_layer_names.back() = last_layer;
+                }
+
+                void setLogistic()
```
l-bat: Can you combine `setLogistic()` with `setSwish()` to avoid code duplication?
For example, `setActivation(String type)`.

YashasSamaga: I have merged ReLU also.

alalek:
Need to adjust Inference Engine checks.
@l-bat Please suggest required changes.

l-bat:
Please, rebase with updated 3.4. PR  #16738 fixed your tests.

PR #16438, opened 2020-01-26
Added parsing for Connected and Dropout. Modified maxpool padding

### LINKED ISSUE #16437: Darknet io: Need parsing of "Connected" and "Dropout" layers and setting of MaxPool padding parameters correctly.

- OpenCV => 4.2.0
- Compiler => g++ 7.4.0

##### Detailed description

**1** Loading model using `cv2.dnn.readNetFromDarknet(alexnet.cfg, alexnet.weights)` gives error `cv2.error: OpenCV(4.2.0-dev) /home/ashish/opencv/modules/dnn/src/darknet/darknet_io.cpp:800: error: (-212:Parsing error) Unknown layer type:  in function 'ReadDarknetFromCfgStream'. `

Getting this error because Parsing for "Connected" layer and "dropout" is not added. 

**2** Padding for Maxpool is incorrect. [Darknet ](https://github.com/pjreddie/darknet/blob/61c9d02ec461e30d55762ec7669d6a1d3c356fb2/src/parser.c#L483) maxpool parser takes `padding` parameter for manual padding. if not mentioned, then take `kernel_size - 1` as padding.  Size after parsing becomes `new size = original size + padding` which is different from opencv. Hence padding needs attention.

https://github.com/pjreddie/darknet/blob/61c9d02ec461e30d55762ec7669d6a1d3c356fb2/src/parser.c#L487

##### Steps to reproduce

`cv2.dnn.readNetFromDarknet(alexnet.cfg, alexnet.weights)`

Alexnet.cfg - https://github.com/pjreddie/darknet/blob/master/cfg/alexnet.cfg
Alexnet.weights - https://pjreddie.com/media/files/alexnet.weights

ashishkrshrivastava:
**Merge with extra**: [opencv/opencv_extra#700](https://github.com/opencv/opencv_extra/pull/700)
resolves #16437

### This pullrequest changes
1. Added parsing for Connected layer and Dropout.
2. Modifies padding parsing for Maxpool layer. 

`opencv_extra=opnecvc_extra`

### DIFF
--- a/modules/dnn/src/darknet/darknet_io.cpp
+++ b/modules/dnn/src/darknet/darknet_io.cpp
@@ -67,6 +67,7 @@
 //M*/
 
 #include "../precomp.hpp"
+#include <opencv2/dnn/shape_utils.hpp>
 
 #include <iostream>
 #include <fstream>
@@ -109,6 +110,26 @@ namespace cv {
                     params.blobs = blobs;
                 }
 
+                void setBatchNorm()
+                {
+                    cv::dnn::LayerParams bn_param;
+
+                    bn_param.name = "BatchNorm-name";
+                    bn_param.type = "BatchNorm";
+                    bn_param.set<bool>("has_weight", true);
+                    bn_param.set<bool>("has_bias", true);
+                    bn_param.set<float>("eps", 1E-6);	// .000001f in Darknet Yolo
+
+                    darknet::LayerParameter lp;
+                    std::string layer_name = cv::format("bn_%d", layer_id);
+                    lp.layer_name = layer_name;
+                    lp.layer_type = bn_param.type;
+                    lp.layerParams = bn_param;
+                    lp.bottom_indexes.push_back(last_layer);
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+                }
+
                 cv::dnn::LayerParams getParamConvolution(int kernel, int pad,
                     int stride, int filters_num)
                 {
@@ -149,24 +170,46 @@ namespace cv {
                     net->layers.push_back(lp);
 
                     if (use_batch_normalize)
-                    {
-                        cv::dnn::LayerParams bn_param;
+                        setBatchNorm();
 
-                        bn_param.name = "BatchNorm-name";
-                        bn_param.type = "BatchNorm";
-                        bn_param.set<bool>("has_weight", true);
-                        bn_param.set<bool>("has_bias", true);
-                        bn_param.set<float>("eps", 1E-6);	// .000001f in Darknet Yolo
+                    layer_id++;
+                    fused_layer_names.push_back(last_layer);
+                }
 
-                        darknet::LayerParameter lp;
-                        std::string layer_name = cv::format("bn_%d", layer_id);
-                        lp.layer_name = layer_name;
-                        lp.layer_type = bn_param.type;
-                        lp.layerParams = bn_param;
-                        lp.bottom_indexes.push_back(last_layer);
-                        last_layer = layer_name;
-                        net->layers.push_back(lp);
+                cv::dnn::LayerParams getParamFullyConnected(int output)
+                {
+                    cv::dnn::LayerParams params;
+                    params.name = "FullyConnected-name";
+                    params.type = "InnerProduct";
+
+                    params.set<bool>("bias_term", false);	// true only if(BatchNorm == false)
+                    params.set<int>("num_output", output);
+
+                    return params;
+                }
+
+                void setFullyConnected(int output, int use_batch_normalize)
+                {
+                    cv::dnn::LayerParams fullyconnected_param =
+                        getParamFullyConnected(output);
+
+                    darknet::LayerParameter lp;
+                    std::string layer_name = cv::format("fullyConnected_%d", layer_id);
+
+                    // use BIAS in any case
+                    if (!use_batch_normalize) {
+                        fullyconnected_param.set<bool>("bias_term", true);
                     }
+
+                    lp.layer_name = layer_name;
+                    lp.layer_type = fullyconnected_param.type;
+                    lp.layerParams = fullyconnected_param;
+                    lp.bottom_indexes.push_back(last_layer);
+                    last_layer = layer_name;
+                    net->layers.push_back(lp);
+
+                    if (use_batch_normalize)
+                        setBatchNorm();
 
                     layer_id++;
                     fused_layer_names.push_back(last_layer);
@@ -191,18 +234,21 @@ namespace cv {
                     fused_layer_names.back() = last_layer;
                 }
 
-                void setMaxpool(size_t kernel, size_t pad, size_t stride)
+                void setMaxpool(int kernel, int pad, int stride)
                 {
                     cv::dnn::LayerParams maxpool_param;
                     maxpool_param.set<cv::String>("pool", "max");
                     maxpool_param.set<int>("kernel_size", kernel);
-                    maxpool_param.set<int>("pad", pad);
+                    maxpool_param.set<int>("pad_l", floor((float)pad / 2));
+                    maxpool_param.set<int>("pad_r", ceil((float)pad / 2));
+                    maxpool_param.set<int>("pad_t", floor((float)pad / 2));
+                    maxpool_param.set<int>("pad_b", ceil((float)pad / 2));
+                    maxpool_param.set<bool>("ceil_mode", false);
                     maxpool_param.set<int>("stride", stride);
-                    maxpool_param.set<cv::String>("pad_mode", "SAME");
                     maxpool_param.name = "Pooling-name";
                     maxpool_param.type = "Pooling";
-                    darknet::LayerParameter lp;
 
+                    darknet::LayerParameter lp;
                     std::string layer_name = cv::format("pool_%d", layer_id);
                     lp.layer_name = layer_name;
                     lp.layer_type = maxpool_param.type;
@@ -539,7 +585,10 @@ namespace cv {
                 net->channels = getParam(net_params, "channels", 3);
                 CV_Assert(net->width > 0 && net->height > 0 && net->channels > 0);
 
-                int current_channels = net->channels;
+                MatShape tensor_shape(3);
+                tensor_shape[0] = net->channels;
+                tensor_shape[1] = net->width;
+                tensor_shape[2] = net->height;
                 net->out_channels_vec.resize(net->layers_cfg.size());
 
                 layers_counter = -1;
@@ -568,23 +617,46 @@ namespace cv {
                             padding = kernel_size / 2;
 
                         CV_Assert(kernel_size > 0 && filters > 0);
-                        CV_Assert(current_channels > 0);
+                        CV_Assert(tensor_shape[0] > 0);
 
-                        setParams.setConvolution(kernel_size, padding, stride, filters, current_channels,
+                        setParams.setConvolution(kernel_size, padding, stride, filters, tensor_shape[0],
                             batch_normalize);
 
-                        current_channels = filters;
+                        tensor_shape[0] = filters;
+                        tensor_shape[1] = (tensor_shape[1] - kernel_size + 2 * padding) / stride + 1;
+                        tensor_shape[2] = (tensor_shape[2] - kernel_size + 2 * padding) / stride + 1;
+                    }
+                    else if (layer_type == "connected")
+                    {
+                        int output = getParam<int>(layer_params, "output", 1);
+                        bool batch_normalize = getParam<int>(layer_params, "batch_normalize", 0) == 1;
+
+                        CV_Assert(output > 0);
+
+                        setParams.setFullyConnected(output, batch_normalize);
+
+                        if(layers_counter && tensor_shape[1] > 1)
+                            net->out_channels_vec[layers_counter-1] = total(tensor_shape);
+
+                        tensor_shape[0] = output;
+                        tensor_shape[1] = 1;
+                        tensor_shape[2] = 1;
                     }
                     else if (layer_type == "maxpool")
                     {
                         int kernel_size = getParam<int>(layer_params, "size", 2);
                         int stride = getParam<int>(layer_params, "stride", 2);
-                        int pad = getParam<int>(layer_params, "pad", 0);
-                        setParams.setMaxpool(kernel_size, pad, stride);
+                        int padding = getParam<int>(layer_params, "padding", kernel_size - 1);
+                        setParams.setMaxpool(kernel_size, padding, stride);
+
+                        tensor_shape[1] = (tensor_shape[1] - kernel_size + padding) / stride + 1;
+                        tensor_shape[2] = (tensor_shape[2] - kernel_size + padding) / stride + 1;
                     }
                     else if (layer_type == "avgpool")
                     {
                         setParams.setAvgpool();
+                        tensor_shape[1] = 1;
+                        tensor_shape[2] = 1;
                     }
                     else if (layer_type == "softmax")
                     {
@@ -599,10 +671,10 @@ namespace cv {
                         CV_Assert(!bottom_layers.empty());
                         std::vector<int> layers_vec = getNumbers<int>(bottom_layers);
 
-                        current_channels = 0;
+                        tensor_shape[0] = 0;
                         for (size_t k = 0; k < layers_vec.size(); ++k) {
                             layers_vec[k] = layers_vec[k] >= 0 ? layers_vec[k] : (layers_vec[k] + layers_counter);
-                            current_channels += net->out_channels_vec[layers_vec[k]];
+                            tensor_shape[0] += net->out_channels_vec[layers_vec[k]];
                         }
 
                         if (layers_vec.size() == 1)
@@ -610,10 +682,16 @@ namespace cv {
                         else
                             setParams.setConcat(layers_vec.size(), layers_vec.data());
                     }
+                    else if (layer_type == "dropout")
+                    {
+                        setParams.setIdentity(layers_counter-1);
+                    }
                     else if (layer_type == "reorg")
                     {
                         int stride = getParam<int>(layer_params, "stride", 2);
-                        current_channels = current_channels * (stride*stride);
+                        tensor_shape[0] = tensor_shape[0] * (stride * stride);
+                        tensor_shape[1] = tensor_shape[1] / stride;
+                        tensor_shape[2] = tensor_shape[2] / stride;
 
                         setParams.setReorg(stride);
                     }
@@ -653,6 +731,8 @@ namespace cv {
                     {
                         int scaleFactor = getParam<int>(layer_params, "stride", 1);
                         setParams.setUpsample(scaleFactor);
+                        tensor_shape[1] = tensor_shape[1] * scaleFactor;
+                        tensor_shape[2] = tensor_shape[2] * scaleFactor;
                     }
                     else if (layer_type == "yolo")
                     {
@@ -686,7 +766,7 @@ namespace cv {
                     else if (activation != "linear")
                         CV_Error(cv::Error::StsParseError, "Unsupported activation: " + activation);
 
-                    net->out_channels_vec[layers_counter] = current_channels;
+                    net->out_channels_vec[layers_counter] = tensor_shape[0];
                 }
 
                 return true;
@@ -712,7 +792,10 @@ namespace cv {
                 if(transpose)
                     CV_Error(cv::Error::StsNotImplemented, "Transpose the weights (except for convolutional) is not implemented");
 
-                int current_channels = net->channels;
+                MatShape tensor_shape(3);
+                tensor_shape[0] = net->channels;
+                tensor_shape[1] = net->width;
+                tensor_shape[2] = net->height;
                 int cv_layers_counter = -1;
                 int darknet_layers_counter = -1;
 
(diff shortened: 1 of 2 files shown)


modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -574,17 +647,47 @@ namespace cv {
...
+                        current_height = 0;
+                        current_total = output;
+
                     }
                     else if (layer_type == "maxpool")
                     {
                         int kernel_size = getParam<int>(layer_params, "size", 2);
                         int stride = getParam<int>(layer_params, "stride", 2);
-                        int pad = getParam<int>(layer_params, "pad", 0);
+                        int pad = getParam<int>(layer_params, "padding", -1);
```
dkurt: It seems to me it should work in the same way as this https://github.com/opencv/opencv/pull/16358

ashishkrshrivastava: @dkurt thanks for your review. 
Can you please check output height and width formulae in case of maxpool and convolution layer. They seem different:
For Maxpool:
https://github.com/pjreddie/darknet/blob/61c9d02ec461e30d55762ec7669d6a1d3c356fb2/src/maxpool_layer.c#L30-L31

For Convolution:
https://github.com/pjreddie/darknet/blob/61c9d02ec461e30d55762ec7669d6a1d3c356fb2/src/convolutional_layer.c#L66-L74

ashishkrshrivastava: I checked output shape of maxpool layer on (width=6, height=6, channels=3) input image if **Padding is used same as Convolution layer**
```
[net]
batch=1
channels=3
height=6
width=6

[maxpool]
size=3
stride=2
padding=0
```
Output using Darknet : `2 x 2 x 3` (height x width x channels) `VALID` padding.
Output using OpenCV : `3 x 3 x 3` (height x width x channels) `SAME` padding.

In Darknet while parsing maxpool, there is no parameter `pad`,  `padding` parameter is used for MANUAL padding that makes output shape
```
l.out_w = (w + padding - size)/stride + 1;
l.out_h = (h + padding - size)/stride + 1;
```

**Please note**: 2*Padding is not there, unlike convolution padding.
`SAME` padding is applied when `padding` is not mentioned in model.
 
**Explanation of changes made for MaxPool padding:**
Hence I explicitly added 

- `pad_mode = SAME` in case no mention of `padding` in model. 
- `pad_mode = VALID` If found `padding = 0`
- `pad = (padding + 1) / 2` in case padding is not 0.

Please correct me if I missed something.

dkurt: @ashishkrshrivastava, According to this, `pad` is not used by Darknet but `padding` is parsed:

https://github.com/pjreddie/darknet/blob/b13f67bfdd87434e141af532cdb5dc1b8369aa3b/src/parser.c#L487

I think that it's better to use `size - 1` (`kernel_size - 1) as a default padding size similarly to Darknet. So at least it's more intuitive than `0` and `-1`

ashishkrshrivastava: @dkurt, using `int pad = getParam<int>(layer_params, "padding", kernel_size -1 );` wont solve the problem. 
Still if I use it then, 
padding will be set to `maxpool_param.set<int>("pad", padding / 2);` and this will give wrong result in some cases like 

1. `kernel_size = 2` , padding = 1 (kernel_size - 1) which makes pad = 0 (padding / 2).
2. `padding = 1`, pad = 0 (padding / 2)

whenever padding will be odd, results will be wrong. we can only handle even padding this way.

ashishkrshrivastava: There is one more issue, `padding = 0` does not work same as `VALID` padding. The reason is ceilmode is set true by default:

https://github.com/opencv/opencv/blob/5d01821d9f4d57921dad82045cc9819cfabcfe15/modules/dnn/src/layers/pooling_layer.cpp#L136

which makes output:

https://github.com/opencv/opencv/blob/5d01821d9f4d57921dad82045cc9819cfabcfe15/modules/dnn/src/layers/pooling_layer.cpp#L1165-L1166

on the other hand, output for` VALID` padding is:

https://github.com/opencv/opencv/blob/426482e05b9db0e891539d01177dc5be120af370/modules/dnn/src/layers/layers_common.cpp#L203-L207

Hence, setting ceilmode = false by default and setting padding = kernel_size - 1 (default) will work only for odd kernel_size i.e. even padding.

ashishkrshrivastava:
@dkurt , I have set Maxpool, it is working properly in all cases. 
Connected layer is working fine and building successfully for all but  linux OpenCL (OCV/OCL_FP16).  Can you please suggest anything so that I can fix that.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -505,16 +565,24 @@ namespace cv {
                         }
                         else {
                             // read section
+                            isDropout = false;
                             read_net = false;
-                            ++layers_counter;
                             const size_t layer_type_size = line.find("]") - 1;
                             CV_Assert(layer_type_size < line.size());
                             std::string layer_type = line.substr(1, layer_type_size);
+                            if(layer_type == "dropout")
```
dkurt: please include single `dropout` layer into your `connected` config file. So we could test both of them

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -540,7 +608,11 @@ namespace cv {
                 CV_Assert(net->width > 0 && net->height > 0 && net->channels > 0);
 
                 int current_channels = net->channels;
+                int current_width = net->width;
+                int current_height = net->height;
+                int current_total = current_height * current_width * current_channels;
```
dkurt: Do we really need to track `current_total`? We can  just calculate it once for connected layer.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -574,17 +646,42 @@ namespace cv {
                             batch_normalize);
 
                         current_channels = filters;
+                        current_width = (current_width - kernel_size + 2 * padding) / stride + 1;
+                        current_height = (current_height - kernel_size + 2 * padding) / stride + 1;
+
```
dkurt: please remove extra line break

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -773,6 +881,51 @@ namespace cv {
...
+                        cv::Mat weightsData_mat(1, output, CV_32F);// scale
+                        cv::Mat biasData_mat(1, output, CV_32F);	// bias
+
+                        ifile.read(reinterpret_cast<char *>(biasData_mat.ptr<float>()), sizeof(float)*output);
+                        if (use_batch_normalize) {
+                            ifile.read(reinterpret_cast<char *>(weightsData_mat.ptr<float>()), sizeof(float)*output);
+                            ifile.read(reinterpret_cast<char *>(meanData_mat.ptr<float>()), sizeof(float)*output);
+                            ifile.read(reinterpret_cast<char *>(stdData_mat.ptr<float>()), sizeof(float)*output);
+                        }
+                        ifile.read(reinterpret_cast<char *>(weightsBlob.ptr<float>()), sizeof(float)*weights_size);
```
dkurt: Please move it to separate function due there are multiple entries with this BN layer

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: Done !

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: @dkurt , Are you asking not to use extra space i.e. out_total_vec ?

https://github.com/opencv/opencv/blob/b9e10746e810d4128521dac48431e771c075e2bf/modules/dnn/src/darknet/darknet_io.hpp#L98

Then we will have to store output size in `out_channel_vec` of previous layer.
```
 else if (layer_type == "connected")
 {
    int output = getParam<int>(layer_params, "output", 1);

    CV_Assert(output > 0);

    setParams.setFullyConnected(output, batch_normalize);

    if(layers_counter > 0)
    {
       if(current_width)
           net->out_channels_vec[layers_counter-1] = current_channels * current_width * current_height;
       else
           net->out_channels_vec[layers_counter-1] = current_channels;
    }
    current_channels = output;
    current_width = 0;
    current_height = 0;
}
```
No need to store `current_total`
This would work fine too but would be little ambiguous to store it in out_channel_vec.

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: Done !

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -126,7 +146,6 @@ namespace cv {
                     return params;
                 }
 
-
```
dkurt: please avoid extra changes here

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -191,18 +232,21 @@ namespace cv {
...
                 {
                     cv::dnn::LayerParams maxpool_param;
                     maxpool_param.set<cv::String>("pool", "max");
                     maxpool_param.set<int>("kernel_size", kernel);
-                    maxpool_param.set<int>("pad", pad);
+                    maxpool_param.set<int>("pad_l", floor((float)pad / 2));
+                    maxpool_param.set<int>("pad_r", ceil((float)pad / 2));
+                    maxpool_param.set<int>("pad_t", floor((float)pad / 2));
+                    maxpool_param.set<int>("pad_b", ceil((float)pad / 2));
+                    maxpool_param.set<bool>("ceil_mode", false);
```
dkurt: Do we have tests for both odd and even `pad` values? May I ask you to check?

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -488,9 +532,53 @@ namespace cv {
...
+                    bn_blobs.push_back(biasData_mat);
+                    blobs.push_back(bn_blobs);
+                }
+                return blobs;
+            }
+
             bool ReadDarknetFromCfgStream(std::istream &ifile, NetParameter *net)
             {
                 bool read_net = false;
+                bool isDropout = false;
```
dkurt: Maybe it's better to process Dropout as Identoty layer (see `layer_type == "route"`)

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -488,9 +532,53 @@ namespace cv {
...
+                if (use_batch_normalize) {
+                    std::vector<cv::Mat> bn_blobs;
+                    bn_blobs.push_back(meanData_mat);
+                    bn_blobs.push_back(stdData_mat);
+                    bn_blobs.push_back(weightsData_mat);
+                    bn_blobs.push_back(biasData_mat);
+                    blobs.push_back(bn_blobs);
+                }
+                return blobs;
+            }
```
dkurt: Looks like it is not so much better. Do you mind to try to unite `convolutional` and `connected` at the previous position?

```cpp
                for (it_type i = net->layers_cfg.begin(); i != net->layers_cfg.end(); ++i) {
                    ++darknet_layers_counter;
                    ++cv_layers_counter;
                    std::map<std::string, std::string> &layer_params = i->second;
                    std::string layer_type = layer_params["type"];

                    if (layer_type == "convolutional" || layer_type == "connected")
                    {
```
than using a different logic for number of weights and their shape

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -574,17 +672,40 @@ namespace cv {
...
+                        CV_Assert(output > 0);
+
+                        setParams.setFullyConnected(output, batch_normalize);
+
+                        if(layers_counter && current_width)
+                            net->out_channels_vec[layers_counter-1] = current_channels * current_width * current_height;
+
+                        current_channels = output;
+                        current_width = 0;
+                        current_height = 0;
```
dkurt: What if network will have two subsequent `connected` layers? A product will be zero.
Maybe it's better to have `MatShape` instead of `current_channels`, `current_width` and  `current_height` with last layer's output shape. This way it can be variable size.

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: I have checked for both even and odd padding. Should I add Maxpool.cfg having two or three maxpool with both even and odd padding so that it can cover all variations ?

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: Sorry, couldn't understand this.

> A product will be zero.

https://github.com/opencv/opencv/blob/c15b33ef80e0fc811b9c9d285d8db8211659da25/modules/dnn/src/darknet/darknet_io.cpp#L687-L690

Using this if statement we are checking the condition that the current layer should not be the first layer of model `layer_counter != 0` and last layer was not connected layer `current_width != 0`. Only then we are doing product and updating that product in previous layer's (`layers_counter-1`) `out_channel_vec`.
`net->out_channels_vec[layers_counter-1] = current_channels * current_width * current_height;`

If current connected is not the first connected layer, it will simply update `out_channel_vec` by `current_channels` (i.e. `output` of current connected layer)
https://github.com/opencv/opencv/blob/c15b33ef80e0fc811b9c9d285d8db8211659da25/modules/dnn/src/darknet/darknet_io.cpp#L690
https://github.com/opencv/opencv/blob/c15b33ef80e0fc811b9c9d285d8db8211659da25/modules/dnn/src/darknet/darknet_io.cpp#L814

For test, [connected.cfg](https://github.com/opencv/opencv_extra/blob/c3db6200e6c87e42f192f45a61f604ae6562b918/testdata/dnn/darknet/connected.cfg) has two connected layers.

dkurt: Got it. I just wanted to solve two problems at the same time - remove this condition because it's too specific and remove all the calculations of width/height from onnx_importer but replace them to layers' `getMemoryShapes`. So if you could now just introduce MatShape and use it, we will easier integrate `getMemoryShapes` in separate PR.

dkurt: So for now it's OK to compute width and height. But If you can replace it by MatShape - it'd be simpler in the future.

ashishkrshrivastava: sure.

modules/dnn/src/darknet/darknet_io.cpp
dkurt: I think max pooling with kernel size 3 and padding 2 will be enough. There are some tests with that configuration from other frameworks as well.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -149,25 +169,47 @@ namespace cv {
...
-                        lp.layer_name = layer_name;
-                        lp.layer_type = bn_param.type;
-                        lp.layerParams = bn_param;
-                        lp.bottom_indexes.push_back(last_layer);
-                        last_layer = layer_name;
-                        net->layers.push_back(lp);
+                        setBatchNorm();
+
+                    layer_id++;
+                    fused_layer_names.push_back(last_layer);
```
dkurt: Looks like this is extra changes? Or not?

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -539,7 +584,7 @@ namespace cv {
                 net->channels = getParam(net_params, "channels", 3);
                 CV_Assert(net->width > 0 && net->height > 0 && net->channels > 0);
 
-                int current_channels = net->channels;
+                MatShape tensor_shape{ net->channels, net->width, net->height };
```
dkurt: Please replace to `MatShape tensor_shape = shape(net->channels, net->width, net->height);` to keep it compatible with C++98.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -725,42 +801,64 @@ namespace cv {
...
                     std::string layer_type = layer_params["type"];
 
-                    if (layer_type == "convolutional")
+                    if (layer_type == "convolutional" || layer_type == "connected")
                     {
-                        int kernel_size = getParam<int>(layer_params, "size", -1);
-                        int filters = getParam<int>(layer_params, "filters", -1);
-                        bool use_batch_normalize = getParam<int>(layer_params, "batch_normalize", 0) == 1;
+                        size_t weights_size;
+                        int nums;
```
dkurt: Let's rename it to something like `numOutChannels`

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -725,42 +801,64 @@ namespace cv {
...
+                        bool use_batch_normalize;
+                        cv::Mat weightsBlob;
+                        if(layer_type == "convolutional")
+                        {
+                            int kernel_size = getParam<int>(layer_params, "size", -1);
+                            int filters = getParam<int>(layer_params, "filters", -1);
```
dkurt: we can do `nums =  getParam<int>(layer_params, "filters", -1);` directly

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -725,42 +801,64 @@ namespace cv {
...
+                            CV_Assert(current_channels > 0);
+
+                            weights_size = filters * current_channels * kernel_size * kernel_size;
+                            int sizes_weights[] = { filters, current_channels, kernel_size, kernel_size };
+                            weightsBlob.create(4, sizes_weights, CV_32F);
+                            nums = filters;
+                        }
+                        else
+                        {
+                            int output = getParam<int>(layer_params, "output", 1);
```
dkurt: `nums = = getParam<int>(layer_params, "output", 1);`

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -725,42 +801,64 @@ namespace cv {
...
-                        CV_Assert(kernel_size > 0 && filters > 0);
-                        CV_Assert(current_channels > 0);
+                            CV_Assert(output>0);
 
-                        size_t const weights_size = filters * current_channels * kernel_size * kernel_size;
-                        int sizes_weights[] = { filters, current_channels, kernel_size, kernel_size };
-                        cv::Mat weightsBlob;
-                        weightsBlob.create(4, sizes_weights, CV_32F);
+                            if(darknet_layers_counter == 0)
+                                current_channels = current_channels * net->height * net->width;
```
dkurt: Looks very tricky. Can we just produce MatShape values?

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: No, I didn't do any change here.

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: Sure, let me change that.

modules/dnn/src/darknet/darknet_io.cpp
dkurt: Oh, got it, sorry!

modules/dnn/src/darknet/darknet_io.cpp
dkurt: Let's use `numOutChannels`. We don't need `filters` here

dkurt: On the other side, let's rename `numOutChannels` to `filters`, yes. This way we will have less changes below. Sorry for inconvenience :)

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -782,7 +883,10 @@ namespace cv {
                     if(activation == "leaky")
                         ++cv_layers_counter;  // For ReLU
 
-                    current_channels = net->out_channels_vec[darknet_layers_counter];
+                    if(!darknet_layers_counter)
```
dkurt: What this `if` condition means? I think that we need to do this only in case of `layer_type == "connected"`

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -568,23 +617,46 @@ namespace cv {
...
+                    else if (layer_type == "connected")
+                    {
+                        int output = getParam<int>(layer_params, "output", 1);
+                        bool batch_normalize = getParam<int>(layer_params, "batch_normalize", 0) == 1;
+
+                        CV_Assert(output > 0);
+
+                        setParams.setFullyConnected(output, batch_normalize);
+
+                        if(layers_counter && tensor_shape[1] > 1)
```
dkurt: What this `if` condition means?

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -568,23 +617,46 @@ namespace cv {
...
+                            net->out_channels_vec[layers_counter-1] = total(tensor_shape);
+
+                        tensor_shape[0] = output;
+                        tensor_shape[1] = 1;
+                        tensor_shape[2] = 1;
```
dkurt: Looks like we need to resize `tensor_shape` to a single element here.

modules/dnn/src/darknet/darknet_io.cpp
```diff
@@ -191,18 +234,21 @@ namespace cv {
                     fused_layer_names.back() = last_layer;
                 }
 
-                void setMaxpool(size_t kernel, size_t pad, size_t stride)
+                void setMaxpool(int kernel, int pad, int stride)
```
dkurt: Why not to use `size_t`?

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: > Let's use `numOutChannels`. We don't need `filters` here

ohh Sorry, Couldn't notice that. I will remove this line, 
https://github.com/opencv/opencv/blob/24932c781ea872cf542536402793b43a8d8c0216/modules/dnn/src/darknet/darknet_io.cpp#L821

ashishkrshrivastava: > On the other side, let's rename `numOutChannels` to `filters`, yes. This way we will have less changes below. Sorry for inconvenience :)

you want me to rename `numOutChannels` to `filters` everywhere ? Just wanted to clarify once.

ashishkrshrivastava: No problem. let me change that.

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava:
```
if(layers_counter && tensor_shape[1] > 1)
            net->out_channels_vec[layers_counter-1] = total(tensor_shape);
```
If `Connected` is first layer of model - `layers_counter = zero` so wanted to avoid this situation `net->out_channels_vec[ -1 ] = total(tensor_shape);`

`tensor_shape[1] > 1` It makes sure that current connected layer if first connected layer after conv or other layer, just like

```
1. [conv] 
2. [conv]
3. [connected]
4. [connected]
```
so it only applies for `3. [connected] ` 
For `4. [connected]` and later on layers, no need to use `net->out_channels_vec[layers_counter-1] = total(tensor_shape);`. 

For all other layers, this will handle.
https://github.com/opencv/opencv/blob/24932c781ea872cf542536402793b43a8d8c0216/modules/dnn/src/darknet/darknet_io.cpp#L769

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: If Connected is the first layer of model, only then we need `total(tensor_shape)` i.e. net->channels * net->width * net->height. 
For all other cases, we can get input size in `tensor_shape[0]` from https://github.com/opencv/opencv/blob/24932c781ea872cf542536402793b43a8d8c0216/modules/dnn/src/darknet/darknet_io.cpp#L889
That's why after first layer, tensor_shape is resized to 1. otherwise, it will hurt here
https://github.com/opencv/opencv/blob/24932c781ea872cf542536402793b43a8d8c0216/modules/dnn/src/darknet/darknet_io.cpp#L838-L839

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: That can be done, but I wanted to act it more like size of ** 1 * 1 * output ** instead of just **output**,.
Considering the case, if any other layer comes connected lets say `upsample` then width and height can simply be resized. If I resize it to single element, then for later layer we will have to check tensor_shape shape and then push_back.
I am not sure if this thing kinda model is there or not, but thought there was no harm doing that. 
I can change it single element size if you think that it will be better that way.

modules/dnn/src/darknet/darknet_io.cpp
ashishkrshrivastava: We can use both of them. I did that because 
https://github.com/opencv/opencv/blob/24932c781ea872cf542536402793b43a8d8c0216/modules/dnn/src/darknet/darknet_io.cpp#L645-L650
here we are taking inputs in `int` type variable.

modules/dnn/src/darknet/darknet_io.cpp
dkurt: Ok, thanks! So let's fix that and we are going to merge PR.

ashishkrshrivastava: Done !

ashishkrshrivastava:
> +1 Thanks for your patience and valuable contribution!

Thank you, It wasn't possible without your help.

PR #16447, opened 2020-01-28
fix spelling errors

hjung4:


alalek:
Thank you for your effort.
Spelling fixes for "3rdparty" code is not accepted, because upgrading on next version lost all these fixes.
It is better to fix code in upstream projects instead.

PR #16452, opened 2020-01-28
objdetect(QR): don't introduce deprecated API, compatibility code

alalek:
relates #15338

### DIFF
--- a/modules/objdetect/include/opencv2/objdetect.hpp
+++ b/modules/objdetect/include/opencv2/objdetect.hpp
@@ -743,6 +743,43 @@ public:
             OutputArrayOfArrays straight_qrcode = noArray()
     ) const;
 
+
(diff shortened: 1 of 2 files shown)


modules/objdetect/include/opencv2/objdetect.hpp
```diff
@@ -743,6 +743,43 @@ class CV_EXPORTS_W QRCodeDetector
             OutputArrayOfArrays straight_qrcode = noArray()
     ) const;
 
+
+#ifndef CV_DOXYGEN  // COMPATIBILITY
```
mshabunin: In most cases we use doxygen directives to ignore block of code:
```.cpp
//! @cond IGNORED 
...
//! @endcond
```

alalek: Lets keep it as is - there is no benefits in that replacement.

PR #16457, opened 2020-01-28
Fixed getCPUCount | Issue #16268

### LINKED ISSUE #16183: Memory leak on K8S

- OpenCV => 4.1.0.25
- Operating System / Platform => ubuntu 18.04 image on k8s (host: centos 7.6)
- Docker version 19.03.4, build 9013bf583a
- K8S v1.15.3
- Compiler => installed with python pip

##### Detailed description

I have a tornado web service container using `cv2.imdecode`, and it causes memory leaking when I deployed the image to k8s, but it would not happen on the same machine with native docker rather than k8s.

I can reproduce the problem with the below code.

I used gdb `info threads` to find out that there were a lot of threads not collected:

```
(gdb) info threads
  Id   Target Id         Frame
* 1    Thread 0x7f78a8376580 (LWP 5991) "python" 0x00007f78a7c85603 in select () at ../sysdeps/unix/syscall-template.S:84
  2    Thread 0x7f789ed93700 (LWP 5992) "python" pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
  3    Thread 0x7f789e592700 (LWP 5993) "python" pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
  4    Thread 0x7f789dd91700 (LWP 5994) "python" pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
  5    Thread 0x7f7899590700 (LWP 5995) "python" pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
  6    Thread 0x7f7896d8f700 (LWP 5996) "python" pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
  7    Thread 0x7f789458e700 (LWP 5997) "python" pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
  8    Thread 0x7f7891d8d700 (LWP 5998) "python" pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
  9    Thread 0x7f788f58c700 (LWP 5999) "python" pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
...more
```

when I tried to look into the thread, it seems my gdb is not well set:

```
(gdb) thread 5
[Switching to thread 5 (Thread 0x7fe775e0a700 (LWP 6052))]
#0  pthread_cond_wait@@GLIBC_2.3.2 () at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
185     ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S: No such file or directory.
```

The comparison I run with the native docker,  it just display:

```
(gdb) info threads
No threads.
```
It proves that threads were not properly collected on k8s.

##### Steps to reproduce

```python
import time
import cv2
import numpy as np

if __name__ == "__main__":
    with open("test.jpeg", "rb") as f:
        image = f.read()
        image_arr = np.fromstring(image, np.uint8)
        image_dec = cv2.imdecode(image_arr, cv2.IMREAD_COLOR)
        print(f"Image shape is {image_dec.shape}")

    while True:
        print("Sleep start")
        time.sleep(3)
        print("Sleep end")
```

### LINKED ISSUE #16268: cv::getCPUCount returns number of CPUs instead of available CPUs on Linux
On a system using the kernel argument `isolcpu`, or when using `taskset` to limit the CPUs available to the process OpenCV continues to allocate threads for all cores on the system even if they are not available to it. This causes OpenCV to create far more threads then is optimal for the target system.

https://github.com/opencv/opencv/blob/7b12cbd7dcbb025c31e7630c9ae709493a8f3d62/modules/core/src/parallel.cpp#L816

To obtain the actual number of cores available to the process one should use `sched_getaffinity` as such:

```
#include <sched.h>

int getCPUs()
{
  cpu_set_t cpu_set;
  sched_getaffinity(0, sizeof(cpu_set), &cpu_set);
  return CPU_COUNT(&cpu_set);
}
```

To workaround this bad behavior one can do the following:

```
cpu_set_t cpu_set;
sched_getaffinity(0, sizeof(cpu_set_t), &cpu_set);
cv::setNumThreads(std::max(1, CPU_COUNT(&cpu_set)-1));
```

ganesh-k13:
#### This pullrequest changes

Fixed getCPUCount to account for cgroups.
Added common code to parse CPU files into a generic function for android and Linux
Tested on Linux VMs, docker images, Kubernetes pods.

resolves #16268
resolves #16183

### DIFF
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -62,6 +62,7 @@
     #if defined __ANDROID__
         #include <sys/sysconf.h>
         #include <sys/syscall.h>
+        #include <sched.h>
     #elif defined __APPLE__
         #include <sys/sysctl.h>
     #endif
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -754,7 +754,7 @@ int cv::getThreadNum(void)
 #endif
 }
 
-#if defined CV_LINUX_GROUPS
+#if defined CV_LINUX_GROUPS || defined __ANDROID__
 static inline std::string getFileContents(const char *filename)
 {
     std::ifstream ifs(filename);
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -743,7 +743,7 @@ int cv::getThreadNum(void)
 #endif
 }
 
-#if (defined __linux__ || defined __GLIBC__ || defined __HAIKU__ || defined __EMSCRIPTEN__) && !defined __ANDROID__
+#if defined __linux__ || defined __GLIBC__ || defined __HAIKU__ || defined __EMSCRIPTEN__ || defined __ANDROID__
 static inline char* getFileContents(const char *filename)
 {
    FILE* fd = fopen(filename, "r");
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -767,7 +767,7 @@ static inline std::string getFileContents(const char *filename)
 }
 #endif
 
-#if defined CV_LINUX_GROUPS
+#if defined CV_LINUX_GROUPS || defined __ANDROID__
 static inline int getNumberOfCPUsImpl(const char *filename)
 {
    std::string file_contents = getFileContents(filename);
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -61,6 +61,7 @@
     #include <fstream>
     #if defined __ANDROID__
         #include <sys/sysconf.h>
+        #include <sys/syscall.h>
     #elif defined __APPLE__
         #include <sys/sysctl.h>
     #endif
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -849,7 +849,7 @@ int cv::getNumberOfCPUs(void)
 #endif
 
     return concurentThreadsSupported == 0 ? (unsigned)sysinfo.dwNumberOfProcessors : (concurentThreadsSupported < (unsigned)sysinfo.dwNumberOfProcessors ? concurentThreadsSupported : (unsigned)sysinfo.dwNumberOfProcessors);
-#elif defined CV_HAVE_CGROUP
+#elif defined CV_HAVE_CGROUP && !defined __ANDROID__
     unsigned ncpus = concurentThreadsSupported;
     static unsigned ncpus_impl = (unsigned)getNumberOfCPUsImpl("/sys/fs/cgroup/cpuset/cpuset.cpus");
     ncpus = min_non_zero(ncpus, ncpus_impl);
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -743,7 +743,7 @@ int cv::getThreadNum(void)
 #endif
 }
 
-#ifdef __ANDROID__
+#if defined __ANDROID__ || defined __linux__ || defined __GLIBC__ || defined __HAIKU__ || defined __EMSCRIPTEN__
 static inline int getNumberOfCPUsImpl(const char *filename)
 {
    FILE* cpuPossible = fopen(filename, "r");
@@ -808,9 +808,10 @@ int cv::getNumberOfCPUs(void)
     #if defined __ANDROID__
         static unsigned ncpus_impl = (unsigned)getNumberOfCPUsImpl("/sys/devices/system/cpu/possible");
     #else /* Linux */
-        static unsigned ncpus_impl = getNumberOfCPUsImpl("/sys/fs/cgroup/cpuset/cpuset.cpus");
+        static unsigned ncpus_impl = (unsigned)getNumberOfCPUsImpl("/sys/fs/cgroup/cpuset/cpuset.cpus");
 
         if(concurentThreadsSupported == 0) { /* somehow hardware_concurrency failed or c++ version < 11 */
+            cpu_set_t cpu_set;
             sched_getaffinity(0, sizeof(cpu_set), &cpu_set);
             concurentThreadsSupported = CPU_COUNT(&cpu_set);
         }
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -45,6 +45,10 @@
 #include <opencv2/core/utils/configuration.private.hpp>
 #include <opencv2/core/utils/trace.private.hpp>
 
+#define min_non_zero(val_1, val_2)                                                    \
+   (((val_1 > 0) && (val_2 > 0) && (val_1 > val_2)) || ((val_1 == 0) && (val_2 > 0))) \
+   ? val_2 : val_1
+
 #if defined _WIN32 || defined WINCE
     #include <windows.h>
     #undef small
@@ -84,9 +88,6 @@
 
 #if defined __linux__ || defined __GLIBC__ || defined __HAIKU__ || defined __EMSCRIPTEN__
    #define CV_HAVE_CGROUP
-   #define min_non_zero(val_1, val_2)                                                    \
-      (((val_1 > 0) && (val_2 > 0) && (val_1 > val_2)) || ((val_1 == 0) && (val_2 > 0))) \
-      ? val_2 : val_1
 #endif
 
 /* IMPORTANT: always use the same order of defines
@@ -847,8 +848,8 @@ int cv::getNumberOfCPUs(void)
 #else
     GetSystemInfo( &sysinfo );
 #endif
-
-    return concurentThreadsSupported == 0 ? (unsigned)sysinfo.dwNumberOfProcessors : (concurentThreadsSupported < (unsigned)sysinfo.dwNumberOfProcessors ? concurentThreadsSupported : (unsigned)sysinfo.dwNumberOfProcessors);
+    unsigned ncpus = min_non_zero(concurentThreadsSupported, sysinfo.dwNumberOfProcessors);
+    return ncpus != 0 ? ncpus : 1;
 #elif defined CV_HAVE_CGROUP && !defined __ANDROID__
     unsigned ncpus = concurentThreadsSupported;
     static unsigned ncpus_impl = (unsigned)getNumberOfCPUsImpl("/sys/fs/cgroup/cpuset/cpuset.cpus");
@@ -859,9 +860,10 @@ int cv::getNumberOfCPUs(void)
 
     #if defined _GNU_SOURCE
         cpu_set_t cpu_set;
-        sched_getaffinity(0, sizeof(cpu_set), &cpu_set);
-        unsigned cpu_count_cpu_set = CPU_COUNT(&cpu_set);
-        ncpus = min_non_zero(ncpus, cpu_count_cpu_set);
+        if(!sched_getaffinity(0, sizeof(cpu_set), &cpu_set)) {
+            unsigned cpu_count_cpu_set = CPU_COUNT(&cpu_set);
+            ncpus = min_non_zero(ncpus, cpu_count_cpu_set);
+        }
     #endif
 
     static unsigned cpu_count_sysconf = (unsigned)sysconf( _SC_NPROCESSORS_ONLN );
@@ -895,9 +897,10 @@ int cv::getNumberOfCPUs(void)
             numCPU = 1;
     }
 
-    return  concurentThreadsSupported == 0 ? (unsigned)numCPU : (concurentThreadsSupported < (unsigned)numCPU ? concurentThreadsSupported : (unsigned)numCPU);
+    unsigned ncpus = min_non_zero(concurentThreadsSupported, numCPU);
+    return ncpus != 0 ? ncpus : 1;
 #else
-    return 1;
+    return concurentThreadsSupported != 0 ? concurentThreadsSupported : 1;
 #endif
 }
 
--- a/modules/core/src/parallel.cpp
+++ b/modules/core/src/parallel.cpp
@@ -743,19 +743,29 @@ int cv::getThreadNum(void)
 #endif
 }
 
+static inline char* getFileContents(const char *filename)
+{
+   FILE* fd = fopen(filename, "r");
+   if(!fd)
+       return (char*)NULL;
+
+   const off_t fsize = fseek(fd, 0, SEEK_END);
+   fseek(fd, 0, SEEK_SET);
+   char* buf = new char[fsize];
+   const char* pbuf = fgets(buf, sizeof(buf), fd);
+   fclose(fd);
+
+   return (char*)pbuf; /* Typecast if NULL */
+}
+
 #if defined __ANDROID__ || defined __linux__ || defined __GLIBC__ || defined __HAIKU__ || defined __EMSCRIPTEN__
 static inline int getNumberOfCPUsImpl(const char *filename)
 {
-   FILE* cpuPossible = fopen(filename, "r");
-   if(!cpuPossible)
-       return 1;
+   char *file_contents = getFileContents(filename);
+   if(!file_contents)
+       return 0;
 
-   char buf[2000]; //big enough for 1000 CPUs in worst possible configuration
-   char* pbuf = fgets(buf, sizeof(buf), cpuPossible);
-   fclose(cpuPossible);
-   if(!pbuf)
-      return 0;
-
+   char *pbuf = file_contents;
    //parse string of form "0-1,3,5-7,10,13-15"
    int cpusAvailable = 0;
 
@@ -779,7 +789,30 @@ static inline int getNumberOfCPUsImpl(const char *filename)
       }
 
    }
+   delete[] file_contents;
    return cpusAvailable;
+}
+#endif
+
+#if defined __linux__ || defined __GLIBC__ || defined __HAIKU__ || defined __EMSCRIPTEN__
+static inline unsigned getNumberOfCPUsCFS(void)
+{
+    char *file_contents = getFileContents("/sys/fs/cgroup/cpu/cpu.cfs_quota_us");
+    std::stringstream ss_period(file_contents);
+    static int cfs_quota;
+    ss_period >> cfs_quota;
+    delete[] file_contents;
+
+    if(cfs_quota < 1) /* cfs_quota must not be 0 or negative */
+        return 0;
+
+    file_contents = getFileContents("/sys/fs/cgroup/cpu/cpu.cfs_period_us");
+    std::stringstream ss_quota(file_contents);
+    static int cfs_period;
+    ss_quota >> cfs_period;
+    delete[] file_contents;
+
+    return (unsigned) cfs_quota/cfs_period;
 }
 #endif
 
@@ -808,7 +841,15 @@ int cv::getNumberOfCPUs(void)
     #if defined __ANDROID__
         static unsigned ncpus_impl = (unsigned)getNumberOfCPUsImpl("/sys/devices/system/cpu/possible");
     #else /* Linux */
-        static unsigned ncpus_impl = (unsigned)getNumberOfCPUsImpl("/sys/fs/cgroup/cpuset/cpuset.cpus");
+        unsigned ncpus_impl_cpuset = (unsigned)getNumberOfCPUsImpl("/sys/fs/cgroup/cpuset/cpuset.cpus");
+        unsigned ncpus_impl_cfs = getNumberOfCPUsCFS();
+        static unsigned ncpus_impl;
+        if(ncpus_impl_cpuset == 0 && ncpus_impl_cfs == 0)
+            ncpus_impl = 1;
+        else if(!ncpus_impl_cpuset != !ncpus_impl_cfs)          /* One of them is zero */
+            ncpus_impl = ncpus_impl_cpuset | ncpus_impl_cfs;    /* Take non zero value */
+        else                                                    /* Both non zero, so take min*/
+            ncpus_impl = ncpus_impl_cpuset < ncpus_impl_cfs ? ncpus_impl_cpuset : ncpus_impl_cfs;
 
         if(concurentThreadsSupported == 0) { /* somehow hardware_concurrency failed or c++ version < 11 */
             cpu_set_t cpu_set;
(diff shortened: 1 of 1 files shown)


ganesh-k13:
Hey @alalek, I do not have the environments of the failing builds, any idea how to solve?

ganesh-k13:
In `org.opencv.test.core.CoreTest.testGetNumberOfCPUUs(CoreTest.java:637)` this line:
https://github.com/opencv/opencv/blob/223790e6fc5fb09d62b10d5616c2902fdc223579/modules/core/misc/java/test/CoreTest.java#L637

I'm not sure if it is true.

ganesh-k13:
Checks pass now, open to suggestions.

alalek:
Thank you for contribution!

Could you please squash commits and rebase onto 3.4 branch?

ganesh-k13:
Thanks @alalek, done.

modules/core/src/parallel.cpp
```diff
@@ -779,8 +773,51 @@ static inline int getNumberOfCPUsImpl()
...
+      return 1;
+
+   return getNumberOfCPUsImplFromFileStr(pbuf);
+}
+#endif
+
+#if (defined __linux__ || defined __GLIBC__ || defined __HAIKU__ || defined __EMSCRIPTEN__) && !defined __ANDROID__
+static inline unsigned getNumberOfCPUsImplLinux()
+{
+   FILE* cpuPossible = fopen("/sys/fs/cgroup/cpuset/cpuset.cpus", "r");
```
mshabunin: This is cgroup v1 files, for v2 it will be different:
http://man7.org/linux/man-pages/man7/cgroups.7.html
https://wiki.archlinux.org/index.php/cgroups

Shouldn't we first determine cgroup of current process and then use it in the path?

mshabunin: Does HAIKU and EMSCRIPTEN have cgroups at all?

ganesh-k13: Thanks for the review @mshabunin, As per [this](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#mounting), they are backward compatible and we get the info we need from v1 itself. Open to change if required as I do not know much about it. 


> Does HAIKU and EMSCRIPTEN have cgroups at all?

I am not sure about this, but if the file does not exist, we handle that case by returning 0. Again open to removing it if needed, did not find much data on this.

modules/core/src/parallel.cpp
```diff
@@ -779,8 +773,51 @@ static inline int getNumberOfCPUsImpl()
...
+   if(!cpuPossible)
+       return 0;
+
+   char buf[2000]; /* big enough for 1000 CPUs in worst possible configuration */
+   char *contents = fgets(buf, sizeof(buf), cpuPossible);
+   fclose(cpuPossible);
+
+   int cpusAvailable = strchr(contents, '-') == NULL ? atoi(contents) : getNumberOfCPUsImplFromFileStr(contents); /* In dockers, this file is a single number, can be (, -) seperated as well*/
```
mshabunin: > In dockers, this file is a single number

It can be anything:
```
docker run -it  --cpuset-cpus 1-2,5 ubuntu:18.04
# cat /sys/fs/cgroup/cpuset/cpuset.cpus 
1-2,5
```

modules/core/src/parallel.cpp
```diff
@@ -789,12 +826,31 @@ int cv::getNumberOfCPUs(void)
...
+    }
+
+    /* Cases:
+     * 1. Both are zero return 1
+     * 2. Logical xor of ncpus_cgroup and ncpus_cpu_set is True:
+     *  2.1. ncpus_cgroup is 0, return ncpus_cpu_set
+     *  2.2. ncpus_cpu_set is 0, return ncpus_cgroup
+     * 3. Return min of zoth
+     */
+    static unsigned ncpus = ncpus_cgroup == 0 && ncpus_cpu_set == 0 ? 1 : (!ncpus_cgroup != !ncpus_cpu_set ? ncpus_cgroup | ncpus_cpu_set : (ncpus_cgroup < ncpus_cpu_set ? ncpus_cgroup : ncpus_cpu_set));
```
mshabunin: Please refactor very long lines like this one. We have soft limitation of 100 columns.

modules/core/src/parallel.cpp
mshabunin: I think it is better to restore previous behavior for HAIKU and EMSCRIPTEN cases.

modules/core/src/parallel.cpp
ganesh-k13: Yeah, I noticed two types:
1. Just `3` or `4`, etc,  it can be interpreted as CPU 4 or 4 CPUs. If one number it means 4 CPUs
2. Next is `1-2,5`. here I just realized I look only for `-`, I'll look for `,` also, thanks

modules/core/src/parallel.cpp
ganesh-k13: Sure

modules/core/src/parallel.cpp
mshabunin: I've rebooted with `cgroup_no_v1=all` option and this path is not available anymore:
```
$ ls /sys/fs/cgroup/
blkio        cpuset   freezer  memory            perf_event  rdma     unified
cpu,cpuacct  devices  hugetlb  net_cls,net_prio  pids        systemd
```
I think we should have fallback to previous method if it is not readable.

modules/core/src/parallel.cpp
mshabunin: I thought just `3` or `4` means one CPU with this id.

modules/core/src/parallel.cpp
ganesh-k13: Oh cool, sounds good. I'll remove HAIKU and EMSCRIPTEN. I'll return 0 if the file is absent and handle.

modules/core/src/parallel.cpp
ganesh-k13: Exactly what I thought too, you can try alpine, it gives count, surprised me as well.

ganesh-k13:
```
docker run -v $PWD:/var/tmp --cpuset-cpus=3 -it alpine sh
/ # cat /sys/fs/cgroup/cpuset/cpuset.cpus
3
```

```
docker run -v $PWD:/var/tmp --cpuset-cpus=4 -it alpine sh
/ # cat /sys/fs/cgroup/cpuset/cpuset.cpus
4
```

mshabunin: What does it show? You set cpuset-cpus and get exactly what you've passed.
```
      --cpus decimal                   Number of CPUs
      --cpuset-cpus string             CPUs in which to allow execution (0-3, 0,1)
```
Docker image does not matter, kernel from the host system is used in containers.

ganesh-k13: Ohh I get what you are saying, ok I thought that was number of CPUs, thanks for the explanation, I'll fix that and the documentation error.

modules/core/src/parallel.cpp
ganesh-k13: I haven't removed Haiku and emscripten to avoid extra # ifs and a possibility of the file existing, but handled it logically.

ganesh-k13:
@mshabunin, I have combined some of the android and linux code to reduce duplicates, I have also fixed the logic of cpus-sets by reusing android code.

mshabunin:
Looks like this approach not work with `--cpus` Docker option, because it does not change cpuset-cpus. As I understand K8S mainly use this option to limit cpu load.

Perhaps we need to handle this case too by dividing cfs_quota to cfs_period and rounding the value, because this parameter supports fractional numbers.
```
$ docker run --cpus 2 -it ubuntu:18.04
# cat /sys/fs/cgroup/cpu/cpu.cfs_period_us 
100000
# cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us 
200000

$ docker run --cpus 1.3 -it ubuntu:18.04
# cat /sys/fs/cgroup/cpu/cpu.cfs_period_us
100000
# cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us
130000
```

ganesh-k13:
Oh, thanks @mshabunin for finding this, I'll read up on this and get back to you

ganesh-k13:
@mshabunin, I've handled `--cpus` case. I tested it on docker and a k8s pod as well.

modules/core/src/parallel.cpp
```diff
@@ -789,12 +838,37 @@ int cv::getNumberOfCPUs(void)
...
+        if(ncpus_impl_cpuset == 0 && ncpus_impl_cfs == 0)
+            ncpus_impl = 1;
+        else if(!ncpus_impl_cpuset != !ncpus_impl_cfs)          /* One of them is zero */
+            ncpus_impl = ncpus_impl_cpuset | ncpus_impl_cfs;    /* Take non zero value */
+        else                                                    /* Both non zero, so take min*/
+            ncpus_impl = ncpus_impl_cpuset < ncpus_impl_cfs ? ncpus_impl_cpuset : ncpus_impl_cfs;
+
+        if(concurentThreadsSupported == 0) { /* somehow hardware_concurrency failed or c++ version < 11 */
+            cpu_set_t cpu_set;
+            sched_getaffinity(0, sizeof(cpu_set), &cpu_set);
```
mshabunin: Looks like this API is only available on Linux and only in case when `_GNU_SOURCE` is enabled: https://linux.die.net/man/2/sched_getaffinity

Function call should be checked for errors.

On my platform `std::thread::hardware_concurrency` returns 8, thus affinity check is effectively disabled.

modules/core/src/parallel.cpp
```diff
@@ -775,12 +791,45 @@ static inline int getNumberOfCPUsImpl()
...
+    ss_quota >> cfs_period;
+    delete[] file_contents;
+
+    return (unsigned) cfs_quota/cfs_period;
 }
 #endif
 
 int cv::getNumberOfCPUs(void)
 {
+#if __cplusplus >= 201103L || (defined(_MSC_VER) && _MSC_VER >= 1900)
```
mshabunin: Wouldn't `CV_CXX11` macro work here?

ganesh-k13:
Thanks for the review @mshabunin, I'll address them. As for priority, I think we can take minimum most value if we have many options as that is always the most probable value.

mshabunin:
@ganesh-k13 , yes minimum should be good. Maybe something like this will be more clear (variables and functions should have meaningful names ofc):
```.cpp
int result = 0;

#if CV_HAVE_METHOD_1
static const int m1 = getCPUS_method1();
if (result > 0 && m1 > 0 && result > m1)
    result = m1;
#endif

#if CV_HAVE_METHOD_2
...
#endif

return result > 0 ? result : 1;
```

ganesh-k13:
Yes, that is much clearer @mshabunin, I'll do like that. Thanks.

modules/core/src/parallel.cpp
ganesh-k13: I seem to have missed this, I'll add it

ganesh-k13:
@mshabunin, can you please review the changes.

modules/core/src/parallel.cpp
```diff
@@ -816,9 +897,10 @@ int cv::getNumberOfCPUs(void)
             numCPU = 1;
     }
 
-    return (int)numCPU;
+    unsigned ncpus = min_non_zero(concurentThreadsSupported, (unsigned)numCPU);
+    return ncpus != 0 ? ncpus : 1;
 #else
-    return 1;
+    return concurentThreadsSupported != 0 ? concurentThreadsSupported : 1;
```
alalek: > return ncpus != 0 ? ncpus : 1;
> return ncpus != 0 ? ncpus : 1;
> return ncpus != 0 ? ncpus : 1;
> return ncpus != 0 ? ncpus : 1;
> return concurentThreadsSupported != 0 ? concurentThreadsSupported : 1;

There are 5 similar returns above.
Please make return once.
Rename `concurentThreadsSupported` => `ncpus` to reduce code.

modules/core/src/parallel.cpp
```diff
@@ -775,26 +796,86 @@ static inline int getNumberOfCPUsImpl()
...
+#endif
+
+#if defined CV_HAVE_CGROUP
+static inline unsigned getNumberOfCPUsCFS(void)
+{
+    std::string file_contents = getFileContents("/sys/fs/cgroup/cpu/cpu.cfs_quota_us");
+    if(file_contents.empty())
+        return 0;
+
+    std::stringstream ss_period(file_contents);
```
alalek:
```.diff
-std::stringstream
+std::istringstream
```

modules/core/src/parallel.cpp
```diff
@@ -77,6 +86,10 @@
     #define HAVE_CONCURRENCY
 #endif
 
+#if defined __linux__ || defined __GLIBC__ || defined __HAIKU__ || defined __EMSCRIPTEN__
+   #define CV_HAVE_CGROUP
```
alalek: `__EMSCRIPTEN__` doesn't have cgroups.

modules/core/src/parallel.cpp
```diff
@@ -45,6 +45,10 @@
 #include <opencv2/core/utils/configuration.private.hpp>
 #include <opencv2/core/utils/trace.private.hpp>
 
+#define min_non_zero(val_1, val_2)                                                    \
+   (((val_1 > 0) && (val_2 > 0) && (val_1 > val_2)) || ((val_1 == 0) && (val_2 > 0))) \
+   ? val_2 : val_1
```
alalek: Please implement this as `static inline` function near `getNumberOfCPUs()` (line 830)

modules/core/src/parallel.cpp
```diff
@@ -775,26 +796,86 @@ static inline int getNumberOfCPUsImpl()
...
+        return 0;
+
+    std::stringstream ss_quota(file_contents);
+    int cfs_period;
+    ss_quota >> cfs_period;
+
+    if(cfs_period < 1)
+        return 0;
+
+    return (unsigned) cfs_quota/cfs_period;
```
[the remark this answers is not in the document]
ganesh-k13: This gets taken care of in the return, if 0.

modules/core/src/parallel.cpp
ganesh-k13: Oh ok, thanks, will do

modules/core/src/parallel.cpp
ganesh-k13: Can you suggest another name :) ? I think it's easier if we include  __EMSCRIPTEN__

modules/core/src/parallel.cpp
ganesh-k13: Sure

ganesh-k13:
Thanks @alalek, I'll address them.

modules/core/src/parallel.cpp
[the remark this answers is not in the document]
alalek: I'm about the case with CPU=0.5

modules/core/src/parallel.cpp
```diff
@@ -775,26 +796,86 @@ static inline int getNumberOfCPUsImpl()
...
 
-    return (int)sysinfo.dwNumberOfProcessors;
+    return ncpus != 0 ? ncpus : 1;
 #elif defined __ANDROID__
-    static int ncpus = getNumberOfCPUsImpl();
-    return ncpus;
-#elif defined __linux__ || defined __GLIBC__ || defined __HAIKU__ || defined __EMSCRIPTEN__
-    return (int)sysconf( _SC_NPROCESSORS_ONLN );
+    unsigned ncpus = concurentThreadsSupported;
+    static unsigned ncpus_impl = (unsigned)getNumberOfCPUsImpl("/sys/devices/system/cpu/possible");
```
alalek: > /sys/devices/system/cpu/possible

This is valid on regular Linux too.

modules/core/src/parallel.cpp
```diff
@@ -775,26 +796,86 @@ static inline int getNumberOfCPUsImpl()
...
+    #if defined _GNU_SOURCE
+        cpu_set_t cpu_set;
+        if(!sched_getaffinity(0, sizeof(cpu_set), &cpu_set)) {
+            unsigned cpu_count_cpu_set = CPU_COUNT(&cpu_set);
+            ncpus = min_non_zero(ncpus, cpu_count_cpu_set);
+        }
+    #endif
+
+    static unsigned cpu_count_sysconf = (unsigned)sysconf( _SC_NPROCESSORS_ONLN );
+    ncpus = min_non_zero(ncpus, cpu_count_sysconf);
```
alalek: Just split this block on 4 cases through sequence (moving "return" to the end would allow that).

`sched_getaffinity()` is applicable on Android too.

modules/core/src/parallel.cpp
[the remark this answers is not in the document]
ganesh-k13: Oh, I assumed we take floor so we return 0. Should I take ceil?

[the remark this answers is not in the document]
alalek: Rounding with truncation is fine except range (0, 1). I believe this should work better:
```
max(1u, (unsigned)cfs_quota/cfs_period);
```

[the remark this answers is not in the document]
ganesh-k13: Oh ok that makes sense, 0.5 makes 1 the valid count, since we take min_non_zero, we might end up taking a bigger value like 8 from other methods cause this is 0. Thanks for explaining.

modules/core/src/parallel.cpp
ganesh-k13: @alalek, can I check cgroups for android as well?

ganesh-k13:
Ok so Andriod is just hard when it comes to `sched_getaffinity`, seems like it's not directly supported due to missing `cpu_set_t`. Any pointers on how to proceed and should we even include this in Andriod code path? Open to other review comments as well. @alalek.

asmorkalov:
@ganesh-k13 I propose to preserve original implementation for Andoird and do not use cgroups option there.

ganesh-k13:
Sure @asmorkalov, will do.

ganesh-k13:
@alalek / @asmorkalov / @mshabunin , can you please review?

modules/core/src/parallel.cpp
```diff
@@ -77,6 +84,14 @@
     #define HAVE_CONCURRENCY
 #endif
 
+#if (defined __linux__ || defined __GLIBC__ || defined __HAIKU__ || defined __EMSCRIPTEN__) && !defined __ANDROID__
+   #define CV_LINUX_GROUPS
+#endif
+
+#if (defined __linux__ || defined __GLIBC__ || defined __HAIKU__) && !defined __ANDROID__
```
asmorkalov: HAIKU is not Linux distribution, it's OSS BeOS implementation with it's own kernel and it does not support CGroups. GLIBC is not requirement for it too. I propose to stay with condition `#if (defined __linux__ && !defined __ANDROID__)`

modules/core/src/parallel.cpp
```diff
@@ -775,26 +798,102 @@ static inline int getNumberOfCPUsImpl()
       }
 
    }
-   return cpusAvailable ? cpusAvailable : 1;
+   return cpusAvailable;
+}
+#endif
+
+#if defined CV_LINUX_GROUPS
```
asmorkalov: It should be `CV_HAVE_CGROUPS`

modules/core/src/parallel.cpp
```diff
@@ -775,26 +798,102 @@ static inline int getNumberOfCPUsImpl()
...
+#elif defined CV_LINUX_GROUPS
+    #if defined CV_HAVE_CGROUPS
+        static unsigned ncpus_impl_cpuset = (unsigned)getNumberOfCPUsImpl("/sys/fs/cgroup/cpuset/cpuset.cpus");
+        ncpus = minNonZero(ncpus, ncpus_impl_cpuset);
+
+        static unsigned ncpus_impl_cfs = getNumberOfCPUsCFS();
+        ncpus = minNonZero(ncpus, ncpus_impl_cfs);
+    #endif
+
+    static unsigned ncpus_impl_devices = (unsigned)getNumberOfCPUsImpl("/sys/devices/system/cpu/possible");
```
alalek: BTW, Why is `/possible` here? (I know that it was here before patch)

See https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-devices-system-cpu:

> online: cpus that are online and being scheduled.
> possible: cpus that have been allocated resources and can be brought online if they are present.

I believe we should use `/online` instead:
```
# cat /sys/devices/system/cpu/online /sys/devices/system/cpu/possible 
0-3
0-3
root ~# echo 0 > /sys/devices/system/cpu/cpu3/online
root ~# cat /sys/devices/system/cpu/online /sys/devices/system/cpu/possible 
0-2
0-3
```

or even both: https://devtalk.nvidia.com/default/topic/1042662/jetson-tx1/-sys-devices-system-cpu-cpu-online-missing-in-the-latest-jetpack-l4t-release/ and select minimal as regular

ganesh-k13: Thanks for the info, I'll read up more on it.

modules/core/src/parallel.cpp
ganesh-k13: Yeah, I think I'll restructure the way we handle Linux, probably just split up the Linux group as there are many uncommon blocks

modules/core/src/parallel.cpp
ganesh-k13: This is very true, `/possible` is showing the hardware count, 128 in my case over just `6` for the VM. I saw a few more options and `online` seems to be right. But Android also has CGroups apparently, Can I add it to CGroups list?

alalek: > Android
> Can I add it to CGroups list?

I believe - yes. It is harmless (just a check for missing files in worst case)

ganesh-k13:
@alalek and @asmorkalov, I have made some changes. Open to any changes, this was just my first thoughts on handling Linux and Andriod based stuff. Let me know what needs to be done.

modules/core/src/parallel.cpp
```diff
@@ -77,6 +84,17 @@
     #define HAVE_CONCURRENCY
 #endif
 
+#if defined __linux__ || defined __GLIBC__ || defined __EMSCRIPTEN__ || defined __HAIKU__
+   #define CV_CPU_GROUPS_1
+   #if defined __ANDROID__
```
asmorkalov: You can remove the condition, it's already handled in line 94.

Issue #14281, opened 2019-04-08
OpenCV 4.0.1 Crashes on armv7


YuraIU02:
##### System information (version)

- OpenCV => 4.0.1:
- Operating System / Platform => Android 8.0 armeabi-v7a:
- Compiler => :grey_question: clang NDK 19, 18, 17, 16

##### Detailed description
I am trying to use opencv-4.0.1-android-sdk.zip in latest Android studio. For Armv8 everything is ok. Armv7 crashes when I am trying to call compute function.

Maybe I need any special arguments in build.gradle file with Armv7 build?

I tried NDK 19, 18, 17, 16, and "-DANDROID_STL=c++_shared" flags, it does not helps. 

Here is Logcat:

signal 7 (SIGBUS), code 1 (BUS_ADRALN)
...
/lib/arm/libopencv_java4.so (_ZN2cv3hal11sepFilter2DEiiiPhjS1_jiiiiiiS1_iS1_iiidi+474)
/lib/arm/libopencv_java4.so (_ZN2cv11sepFilter2DERKNS_11_InputArrayERKNS_12_OutputArrayEiS2_S2_NS_6Point_IiEEdi+428)
/lib/arm/libopencv_java4.so (_ZN2cv12GaussianBlurERKNS_11_InputArrayERKNS_12_OutputArrayENS_5Size_IiEEddi+432)
/lib/arm/libopencv_java4.so
/lib/arm/libopencv_java4.so (_ZN2cv9Feature2D7computeERKNS_11_InputArrayERNSt6__ndk16vectorINS_8KeyPointENS4_9allocatorIS6_EEEERKNS_12_OutputArrayE+76)
...
##### Steps to reproduce
call Feature2D.compute or cv::buildOpticalFlowPyramid

alalek:
probably relates #12961

alalek:
Could you try the latest release: [OpenCV 4.1.0](https://github.com/opencv/opencv/releases/tag/4.1.0) ?

Palladinium:
I'm having the same issue with armv7 devices.
Unlike @YuraIU02, I'm building my own `.so` which links to the static prebuilt binaries.
Updating to 4.1.0 did not help.

Here are stack trace excerpts from versions running both 4.0.1 and 4.1.0:

[stacktrace_4.0.1.txt](https://github.com/opencv/opencv/files/3250378/stacktrace_4.0.1.txt)
[...]

PR #16463, opened 2020-01-29
core: CV_STRONG_ALIGNMENT macro

alalek:
Should be used to guard unsafe type casts of pointers.
Some platforms requires that.

relates #16373 #16152 #16139 #14281 #12961

### DIFF
--- a/modules/core/include/opencv2/core/cvdef.h
+++ b/modules/core/include/opencv2/core/cvdef.h
@@ -335,6 +335,13 @@ enum CpuFeatures {
 
 #include "cv_cpu_dispatch.h"
 
+#if !defined(CV_STRONG_ALIGNMENT) && defined(__arm__) && !(defined(__aarch64__) || defined(_M_ARM64))
+// int*, int64* should be propertly aligned pointers on ARMv7
+#define CV_STRONG_ALIGNMENT 1
+#endif
+#if !defined(CV_STRONG_ALIGNMENT)
+#define CV_STRONG_ALIGNMENT 0
+#endif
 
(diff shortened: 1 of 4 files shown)


vpisarev:
@alalek, we have `v_load_aligned()` to load from aligned addresses. `v_load()` _must_ be able to handle unaligned data.

alalek:
@vpisarev This is not about SIMD only - check #16139 and related issue.
Also there is no way to propagate alignment requirements through template.

> v_load_aligned() to load from aligned addresses

... aligned on all 128 / 256 bits. This requirement is much stronger than alignment on lane(base) type.

vpisarev:
@alalek, oh, I see, I misunderstood the code. But can we replace CV_Assert() with CV_DbgAssert()? Since adding the check to each load and store operation is quite significant overhead.

alalek:
> replace CV_Assert() with CV_DbgAssert()

It is `intrin_cpp.hpp` file. This SIMD implementation is not used by default:
- tests for "intrin_cpp" are enabled 2.5 months ago: #15925
- for other OpenCV code it is enabled 1 month ago: #16236 (through C++ flag)

There are [weekly "SIMD emulator" builders](http://pullrequest.opencv.org/buildbot/grid?tag=etc&width=7) which run `intrin_cpp` code through all tests (to ensure that SIMD functions are used properly).
Replacing to `CV_DbgAssert()` would require to switch these builders in debug mode - this is much slower than current checks (+emulated SIMD code) in release mode. So I would like to avoid that.

alalek:
:+1:

Issue #16480, opened 2020-02-02
Android manifests are duplicated in samples


hannesa2:
During prepare a  pull request, I found the `AndroidManifest.xml`'s sample are there **twice**.

<img width="775" alt="image" src="https://user-images.githubusercontent.com/3314607/73610071-f8095a80-45d3-11ea-9d8f-8b180204cbf9.png">

What's the purpose of this, or is one of them just an outdated relict ?
Who knows it ?

If it's a pointless double code, I vote for a deletion to reduce confusion. But which one ?

## Eclipse ?
<img width="528" alt="image" src="https://user-images.githubusercontent.com/3314607/73610117-5df5e200-45d4-11ea-9f9e-8900f179aa58.png">
<img width="706" alt="image" src="https://user-images.githubusercontent.com/3314607/73610191-fa1fe900-45d4-11ea-99cb-151387dc4ec2.png">

## Gradle
<img width="590" alt="image" src="https://user-images.githubusercontent.com/3314607/73610128-70701b80-45d4-11ea-91c6-99ccae43feb8.png">
<img width="693" alt="image" src="https://user-images.githubusercontent.com/3314607/73610176-e2e0fb80-45d4-11ea-952f-c59d7a69d5ad.png">

## Compare a sample
<img width="1358" alt="image" src="https://user-images.githubusercontent.com/3314607/73610238-70bce680-45d5-11ea-93d2-91feefc7608c.png">

For me it looks like the one with **/**gradle**/AndroidManifest.xml is the right one

alalek:
relates #13386

Feel free to remove outdated files which supports abandoned "Eclipse ADT":
- SO topic: https://stackoverflow.com/questions/42958186/is-eclipse-adt-officially-abandoned
- old link: http://developer.android.com/tools/sdk/eclipse-adt.html
from "master" branch.

PR #16481, opened 2020-02-02
delete old AndroidManifest.xml

hannesa2:
It resolves https://github.com/opencv/opencv/issues/16480 

I found the AndroidManifest.xml's sample are there twice, a newer was introduced during switch to Gradle.
That's why I delete the old one to reduce confusion


hannesa2:
I see here https://pullrequest.opencv.org/buildbot/builders/precommit_docs/builds/23638/steps/make%20doxygen/logs/warnings%20%281%29 `mobilenet-objdetect` needs still both AndroidManifest.xml's 
That is not necessary, but I've no clue how to solve it

alalek:
@hannesa2 Refer to related tutorial page: https://docs.opencv.org/4.2.0/d0/d6c/tutorial_dnn_android.html

Path should be fixed (missing `gradle/` prefix):
```
@include android/mobilenet-objdetect/AndroidManifest.xml
```

PR #16488, opened 2020-02-03
trying to fix handling file storages with extremely long lines

### LINKED ISSUE #11061: Large JSON can't be loaded while smaller ones can
- OpenCV => :grey_question: 3.4.1-dev
- Operating System / Platform => :grey_question: Ubuntu16.04.4lts
- Compiler => :grey_question: g++ 5.4.0

##### Detailed description

# The FileStorage could not handle large JSON file. 

OpenCV(3.4.1-dev) Error: Parsing error (../models/svm_0.json(0): Unexpected character) in icvJSONParseSeq, file /home/king-kong/my-repo/action_detect/opencv/modules/core/src/persistence_json.cpp, line 460
error: OpenCV(3.4.1-dev) /home/king-kong/my-repo/action_detect/opencv/modules/core/src/persistence_json.cpp:460: error: (-212) Parsing error: ../models/svm_0.json(0): Unexpected character in function icvJSONParseSeq

OpenCV(3.4.1-dev) Error: Unknown error code -49 (Input file is empty) in cvOpenFileStorage, file /home/king-kong/my-repo/action_detect/opencv/modules/core/src/persistence_c.cpp, line 384
error: OpenCV(3.4.1-dev) /home/king-kong/my-repo/action_detect/opencv/modules/core/src/persistence_c.cpp:384: error: (-49) Unknown error code -49: Input file is empty in function cvOpenFileStorage

And Interestingly, after I split the json file into smaller pieces, the FileStorage is able to handle the json file. 

##### Steps to reproduce

I have the json file and the code here, https://github.com/326623/test/tree/3884c5cd0231abbdc953d10361a0453f4ab4b8f2

Thanks in adanced!

vpisarev:
resolves #11061

### DIFF
--- a/modules/core/src/persistence.cpp
+++ b/modules/core/src/persistence.cpp
@@ -606,12 +606,11 @@ public:
                     int last_occurrence = -1;
                     xml_buf_size = MIN(xml_buf_size, int(file_size));
                     fseek( file, -xml_buf_size, SEEK_END );
-                    std::vector<char> xml_buf_(xml_buf_size+2);
                     // find the last occurrence of </opencv_storage>
                     for(;;)
                     {
                         int line_offset = (int)ftell( file );
-                        const char* ptr0 = this->gets(&xml_buf_[0], xml_buf_size );
+                        const char* ptr0 = this->gets(xml_buf_size);
                         const char* ptr = NULL;
                         if( !ptr0 )
                             break;
@@ -693,8 +692,8 @@ public:
         }
         else
         {
-            const size_t buf_size0 = 1 << 20;
-            size_t buf_size = buf_size0;
+            const size_t buf_size0 = 1 << 15;
+            buffer.resize(buf_size0);
             if( mem_mode )
             {
                 strbuf = (char*)filename_or_buf;
@@ -704,8 +703,7 @@ public:
             const char* yaml_signature = "%YAML";
             const char* json_signature = "{";
             const char* xml_signature  = "<?xml";
-            char buf[16];
-            this->gets( buf, sizeof(buf)-2 );
+            char* buf = this->gets(16);
             char* bufPtr = cv_skip_BOM(buf);
             size_t bufOffset = bufPtr - buf;
 
@@ -720,23 +718,8 @@ public:
             else
                 CV_Error(CV_BADARG_ERR, "Unsupported file storage format");
 
-            if( !isGZ )
-            {
-                if( !mem_mode )
-                {
-                    fseek( file, 0, SEEK_END );
-                    buf_size = ftell( file );
-                }
-                else
-                    buf_size = strbufsize;
-                buf_size = MIN( buf_size, buf_size0 );
-                buf_size = MAX( buf_size, (size_t)(CV_FS_MAX_LEN*6 + 1024) );
-            }
             rewind();
             strbufpos = bufOffset;
-
-            buffer.reserve(buf_size + 256);
-            buffer.resize(buf_size);
             bufofs = 0;
 
             try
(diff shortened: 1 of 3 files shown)


modules/core/src/persistence.cpp
```diff
@@ -809,56 +792,70 @@ class FileStorage::Impl : public FileStorage_API
...
         {
-            char* ptr = gzgets( gzfile, str, maxCount );
-            if (ptr && maxCount > 256 && !(flags & cv::FileStorage::BASE64))
-            {
-                size_t sz = strnlen(ptr, maxCount);
-                CV_Assert(sz < (size_t)(maxCount - 1) && "OpenCV persistence doesn't support very long lines");
-            }
-            return ptr;
+            int count = (int)std::min(buffer.size() - ofs - 16, maxCount);
+            char* ptr = getsFromFile( &buffer[ofs], count+1 );
```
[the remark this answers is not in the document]
vpisarev: fgets() requires "count+1", because it adds '\0' in the end. but sometimes we need to add '\n' to the end of line. so we might need at least count + 2. "+16" has been added just in case, it's definitely sufficient

modules/core/src/persistence.cpp
```diff
@@ -809,56 +792,70 @@ class FileStorage::Impl : public FileStorage_API
...
+            ofs += delta;
+            maxCount -= delta;
+            if( ptr[delta-1] == '\n' || maxCount == 0 )
+                break;
+            if( delta == count )
+                buffer.resize((size_t)(buffer.size()*1.5));
         }
-#endif
-        CV_Error(CV_StsError, "The storage is not opened");
+        return ofs > 0 ? &buffer[0] : 0;
```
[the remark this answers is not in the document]
vpisarev: yes. I'm sure. when ".gets()" is called by the XML/YAML/JSON parser, it returns the pointer to the read line. The parsers process the line and then call ".gets()" another time to get the next line and so on. Of course, they do not store pointers to the middle of the previously read line, because they consume all the content. ".resize()" can be safely called inside ".gets()".

vpisarev:
@alalek, the idea is obvious, and the patch is not that big. persistence functions are tested extensively by the test system itself. I suggest to integrate the patch now. 99% sure that it works correctly. If we find any problems with it, with can always revert the patch, because it's very local, touches just one file (+ adds the test that can be disabled).

modules/core/src/persistence_json.cpp
```diff
@@ -858,8 +858,6 @@ class JSONParser : public FileStorageParser
             CV_PARSE_ERROR_CPP( "left-brace of top level is missing" );
         }
 
-        if( !ptr || !*ptr )
-            CV_PARSE_ERROR_CPP( "Unexpected End-Of-File" );
```
alalek: I believe `ptr` should be reassigned above instead:
```
ptr = parseMap(ptr, ...);
...
ptr = parseSeq(ptr, ...);
```

vpisarev: I believe, those 2 lines just left out by mistake after refactoring. If `ptr` is reassigned, all the tests fail. Simply remove them.

modules/core/src/persistence.cpp
```diff
@@ -693,8 +692,8 @@ class FileStorage::Impl : public FileStorage_API
         }
         else
         {
-            const size_t buf_size0 = 1 << 20;
-            size_t buf_size = buf_size0;
+            const size_t buf_size0 = 40;
+            buffer.resize(buf_size0);
```
alalek: Do we left this "as is" for merge?

vpisarev: @alalek, I suggest to merge it as-is. It will help to catch issues (if any) with this patch sooner rather than later.

vpisarev:
@alalek, please, do not merge it! I think, I found some problems with this implementation. Wait a bit.

PR #16491, opened 2020-02-03
fixed issue with Mat reshape in decodeMulti()

rayonnant14:
solved problem with incorrect processing of transformation points in method QRCodeDetector::decodeMulti()

relates #15338

### DIFF
--- a/modules/objdetect/test/test_qrcode.cpp
+++ b/modules/objdetect/test/test_qrcode.cpp
@@ -485,8 +485,11 @@ TEST(Objdetect_QRCode_decodeMulti, decode_regression_16491)
 {
 #ifdef HAVE_QUIRC
     Mat zero_image = Mat::zeros(256, 256, CV_8UC1);
-    std::vector<Point> vec_corners{Point(16, 16), Point(128, 16), Point(128, 128), Point(16, 128),
-                                   Point(16, 16), Point(128, 16), Point(128, 128), Point(16, 128)};
+    Point corners_[] = {Point(16, 16), Point(128, 16), Point(128, 128), Point(16, 128),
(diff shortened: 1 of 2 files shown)


asmorkalov:
@rayonnant14 Could you add a test for the resolved bug?

alalek:
Add simple "bad" test (~10 lines): `TEST(Objdetect_QRCode_Multi, decode_regression_16491)`
Use `EXPECT_NO_THROW(qrcode.decodeMulti(...));` as check.
Don't load any images. Don't check accuracy.
Hardcode points. Use zero input image if possible.

modules/objdetect/test/test_qrcode.cpp
```diff
@@ -481,6 +481,23 @@ INSTANTIATE_TEST_CASE_P(/**/, Objdetect_QRCode_Close, testing::ValuesIn(qrcode_i
 INSTANTIATE_TEST_CASE_P(/**/, Objdetect_QRCode_Monitor, testing::ValuesIn(qrcode_images_monitor));
 INSTANTIATE_TEST_CASE_P(/**/, Objdetect_QRCode_Multi, testing::ValuesIn(qrcode_images_multiple));
 
+TEST(Objdetect_QRCode_decodeMulti, decode_regression_16491)
+{
+#ifdef HAVE_QUIRC
+    Mat zero_image = Mat::zeros(256, 256, CV_8UC1);
+    std::vector<Point> vec_corners{Point(16, 16), Point(128, 16), Point(128, 128), Point(16, 128),
+                                   Point(16, 16), Point(128, 16), Point(128, 128), Point(16, 128)};
```
alalek: C++11 is not allowed in OpenCV tests on 3.4 branch by default.
Please use `Points corners_[] = {...};` array + `vector::assign()`

PR #16503, opened 2020-02-04
fixes #16500

### LINKED ISSUE #16500: should add cuda dll path via os.add_dll_directory on windows

- OpenCV => master
- Operating System / Platform =>win10
- Compiler => vs2019
- CUDA 10.2
- Python 3.8

##### Detailed description

I compiled opencv from source with cuda and python supports and encountered "import cv2 ,dll failed to load " error, after spent a lot of time solving this problem, I found that dll loading policy has changed in python3.8 such that PATH and the current working directory are no longer used.
Please see [this note](https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew) 

##### Steps to reproduce

1. Compiler opencv under python3.8 on windows.
2. Can't load cv2 even when C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.2\bin is in PATH
3. Copy all cuda DLL to C:\Python38 then it works, which shows that python3.8 doesn't use PATH when loading DLLs.
4. Now remove all cuda DLLs  from  C:\Python38 and add C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v10.2\bin  via os.add_dll_directory in cv2/__init__.py ,then it will work too.

cyyever:
resolves #16500


modules/python/python_loader.cmake
```diff
@@ -48,6 +48,9 @@ if(NOT OpenCV_FOUND)  # Ignore "standalone" builds of Python bindings
   else()
     list(APPEND CMAKE_PYTHON_BINARIES_PATH "'${LIBRARY_OUTPUT_PATH}'")
   endif()
+  if (DEFINED CUDA_TOOLKIT_ROOT_DIR)
+    list(APPEND CMAKE_PYTHON_BINARIES_PATH "'${CUDA_TOOLKIT_ROOT_DIR}/bin'")
```
alalek: 1. It is CUDA specific, so it should be guarded by `if(HAVE_CUDA)`
2. It is Windows only, so add `if(WIN32)` guard too
3. It is unlikely that we want path from the "build" machine. Perhaps python loader should reuse target machine environment. Something like:
```
os.environ.get('SOME_CUDA_ROOT_DIR_ENV_PARAMETER', `${CUDA_TOOLKIT_ROOT_DIR}`) + '/bin'
```

modules/python/python_loader.cmake
```diff
@@ -58,7 +58,13 @@ if(NOT OpenCV_FOUND)  # Ignore "standalone" builds of Python bindings
     else()
       list(APPEND CMAKE_PYTHON_BINARIES_INSTALL_PATH "os.path.join(${CMAKE_PYTHON_EXTENSION_INSTALL_PATH_BASE}, '${OPENCV_LIB_INSTALL_PATH}')")
     endif()
-    string(REPLACE ";" ",\n    " CMAKE_PYTHON_BINARIES_PATH "${CMAKE_PYTHON_BINARIES_INSTALL_PATH}")
+    set(CMAKE_PYTHON_BINARIES_PATH "${CMAKE_PYTHON_BINARIES_INSTALL_PATH}")
+    if (WIN32 AND WITH_CUDA)
```
alalek: Please use `HAVE_CUDA` instead of `WITH_CUDA`

Issue #12342, opened 2018-08-30
imread() didn't close the image file sometime


akmalhakimi1991:
I have a program where I read an image file (e.g. from a webcam), crop a face (if any) and save the cropped face into a new image file. The program will then delete the original image file. Here's the code: 
  
    import cv2
    import os
    import datetime

    cam = cv2.VideoCapture(0)
    while True:
        imagenames = []
        imagename = 'face'+str(datetime.datetime.now().strftime('%Y%m%d_%H%M%S'))+'.jpg'
        ret, frame = cam.read()
        cv2.imwrite(imagename, frame)
        
        img = cv2.imread(imagename)
        face_cascade = cv2.CascadeClassifier("haarcascade_frontalface_default.xml")
        gray = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
        faces = face_cascade.detectMultiScale(gray, 1.3, 5)
        
        n = 0
        for (x,y,w,h) in faces:
            img = cv2.rectangle(img,(x,y),(x+w,y+h),(255,0,0),2)
            crop_img = img[y:y+h, x:x+w]

            try:
                imagenames.append(str(n)+'_'+imagename)
                cv2.imwrite("img"+str(n)+'_'+imagename,crop_img)
                n=n+1
            except Exception, e:
                print e
                pass

        os.remove(imagename)  
  
  
Problem is, sometime the image is not properly closed. This cause the program to throw this error:
`The process cannot access the file because it is being used by another process`  
  
opencv-python version: 3.4.2.17
OS: Windows 7
Python version: 2.7.15

alalek:
Please provide minimal and complete reproducer (eliminate CascadeClassifier and camera access at least).

akmalhakimi1991:
This code will occasionally reproduce the error:  
  
    import cv2
    import os
    import datetime

    while True:
        imagename_src = "src_img.jpg"
        imagename_dest = 'to_be_deleted_'+str(datetime.datetime.now().strftime('%Y%m%d_%H%M%S'))+'.jpg'
        img = cv2.imread(imagename_src)
        cv2.imwrite(imagename_dest, img)
        img2 = cv2.imread(imagename_dest)

        os.remove(imagename_dest)

ganeshtata:
@akmalhakimi1991 Could you post the entire stacktrace?

karlphillip:
I would like to investigate this.

karlphillip:
[...]

PR #16509, opened 2020-02-05
fixes #15155

### LINKED ISSUE #15155: cv2.imwrite does not report error if user does not have permission to write to a directory

- OpenCV => 3.*
- Operating System / Platform => Windows x64

##### Detailed description

In python 3.6.8 using the cv2 library, when using cv2.imwrite as:

> cv2.imwrite('path/to/image/myi_image.jpg', my_image)

cv2 does not report an error if the user running the script does not have write permissions to the directory '/path/to/image' as in the above code segment.

imwrite just returns false and the error is extremely hard to debug due to no stderr output.

In addition, the existence of the directory is also not validated.

omasaht:
resolves #15155

### This pull request adds changes to modules/imgcodecs/src/loadsave.cpp
imwrite_ now describes permission denied error when a user does not have write permission in directory.

<cut/>

Example:
```python
import cv2
import numpy as np

img = 255*np.ones((100,100),dtype=np.uint8)

#writing in permissible directory
ret = cv2.imwrite('myImage.png',img)
print('imwrite returned: ',ret)

#writing in non permissible directory
ret = cv2.imwrite('/myImage.png',img)
print('imwrite returned: ',ret)

```

Code Output:
```
imwrite returned:  True
[ WARN:0] imwrite_('/myImage.png'): can't open file for writing: permission denied
imwrite returned:  False
```

### DIFF
--- a/modules/imgcodecs/src/loadsave.cpp
+++ b/modules/imgcodecs/src/loadsave.cpp
@@ -51,6 +51,8 @@
 #undef max
 #include <iostream>
 #include <fstream>
+#include <cerrno>
+#include <opencv2/core/utils/logger.hpp>
 #include <opencv2/core/utils/configuration.private.hpp>
 
 
@@ -718,6 +720,23 @@ static bool imwrite_( const String& filename, const std::vector<Mat>& img_vec,
             code = encoder->write( write_vec[0], params );
         else
             code = encoder->writemulti( write_vec, params ); //to be implemented
+
+        if (!code)
+        {
+            FILE* f = fopen( filename.c_str(), "wb" );
+            if ( !f )
+            {
+                if (errno == EACCES)
+                {
+                    CV_LOG_WARNING(NULL, "imwrite_('" << filename << "'): can't open file for writing: permission denied");
+                }
+            }
+            else
+            {
+                fclose(f);
+                remove(filename.c_str());
+            }
+        }
     }
     catch (const cv::Exception& e)
     {


dkurt:
This patch changes key behavior of imwrite - it returns boolean flag to indicate success of operation. Nobody expect exceptions here.

omasaht:
@dkurt Output of imwrite will still be a boolean, however, permission denied error message will simply be printed on stdcerr in case of exception as it is being catched in try { } catch { } block. This patch simply describes the permission denied error so it will be easier to debug for end user. I do notice that imwrite does not always return a boolean i.e. if invalid file extension is specified: cv2.imwrite('test.abc',img), it will throw an exception and exit code. I'm sorry if I misunderstood your comment, I'm open to suggestions.

modules/imgcodecs/src/loadsave.cpp
```diff
@@ -714,6 +715,19 @@ static bool imwrite_( const String& filename, const std::vector<Mat>& img_vec,
     bool code = false;
     try
     {
+        FILE* f = fopen( filename.c_str(), "wb" );
+        if (f == 0)
+        {
+            if (errno == EACCES)
+            {
+                CV_Error( Error::StsError, "permission denied" );
```
alalek: Please use logging:
```
CV_LOG_WARNING(NULL, "imwrite(): Can't open file for writing: " << filename);
return false;
```

modules/imgcodecs/src/loadsave.cpp
```diff
@@ -714,6 +715,19 @@ static bool imwrite_( const String& filename, const std::vector<Mat>& img_vec,
...
+            }
+        }
+        else
+        {
+            fclose(f);
+            remove(filename.c_str());
```
alalek: There is some chance that we may get problem like this again: #12342
See general note.

dkurt:
@omasaht, thanks, got it! With code example it's clear now.

modules/imgcodecs/src/loadsave.cpp
omasaht: I have updated my code accordingly. I did initially consider trying for file access in all encoders however, as you've said, the implementations of encoder.write are different. This could potentially raise more issues in the process and may not be so elegant. Thus I've opted for second suggestion.

modules/imgcodecs/src/loadsave.cpp
omasaht: I believe this issue is not something caused by OpenCV itself but rather an inherent could-be problem that can occur in some OS's like Windows which don't allow mutual sharing of file. In the unlikely scenario that the program lands on this else condition, it will at worse create a temporary file that is not removed, imwrite will still not crash as remove will simply return non zero. I think this could occur if another process from OS accesses the temporary file in between fclose() and remove() function. I will try to generate this occurrence by building opencv without the condition imposed by if( !f ) and share my findings. Additionally, I could further add warning log if fclose() and remove() do not return 0.

PR #16514, opened 2020-02-05
WIP: Basic RMat implementation

rgarnov:
* Added basic RMat implementation
* Added integration to GExecutor: maps RMat, saves mapped gapi::own::mat into the magazine and passes own::mat to backends for execution, so backends currently can't work with RMat directly (actually there are no changes which affect backends' point of view)

<cut/>


modules/gapi/include/opencv2/gapi/own/rmat.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace cv { namespace gapi { namespace own {
+
+class HostMat
+{
+public:
+    virtual ~HostMat() = default;
+    virtual Mat getMat() const = 0;
+    virtual void flush() const = 0;
+};
```
dmatveev: What this `HostMat` stands for and should we really expose it?

modules/gapi/include/opencv2/gapi/own/rmat.hpp
```diff
@@ -0,0 +1,52 @@
...
+        virtual ~Impl() = default;
+        virtual GMatDesc desc() const = 0;
+        virtual HostMatP access(AccessType) const = 0;
+    };
+    using ImplP = std::shared_ptr<Impl>;
+
+    RMat() = default;
+    RMat(ImplP impl) : m_impl(impl) {}
+    GMatDesc desc() const { return m_impl->desc(); }
+    HostMatP access(AccessType a) const { return m_impl->access(a); }
```
dmatveev: Why it is not a cv::Mat?

modules/gapi/include/opencv2/gapi/own/rmat.hpp
```diff
@@ -0,0 +1,52 @@
...
+    ImplP impl() const { return m_impl; }
+
+private:
+    ImplP m_impl;
+};
```
dmatveev: Documentation for this class is required

modules/gapi/src/api/gbackend.cpp
```diff
@@ -106,6 +106,15 @@ namespace magazine {
...
 // FIXME implement the below functions with visit()?
 
+static void bindRMat(Mag& mag, const RcDesc& rc, const cv::gapi::own::RMat& rmat)
+{
+    // Note: HostMatP should be kept in the magazine
+    auto& hmat = mag.template slot<cv::gapi::own::RMat::HostMatP>()[rc.id];
+    auto&  mat = mag.template slot<cv::gapi::own::Mat>()[rc.id];
+    hmat = rmat.access(cv::gapi::own::RMat::AccessType::R);
+    mat = hmat->getMat();
+}
```
dmatveev: What is the reason to do this? It makes the memory ownership more tangled

modules/gapi/src/backends/common/gbackend.hpp
```diff
@@ -46,9 +46,11 @@ namespace magazine {
 
 } // namespace magazine
 #if !defined(GAPI_STANDALONE)
-using Mag = magazine::Class<cv::gapi::own::Mat, cv::UMat, cv::gapi::own::Scalar, cv::detail::VectorRef, cv::detail::OpaqueRef>;
+using Mag = magazine::Class<cv::gapi::own::Mat, cv::UMat, cv::gapi::own::Scalar, cv::detail::VectorRef, cv::detail::OpaqueRef,
+                            cv::gapi::own::RMat::HostMatP>;
 #else
-using Mag = magazine::Class<cv::gapi::own::Mat, cv::gapi::own::Scalar, cv::detail::VectorRef, cv::detail::OpaqueRef>;
+using Mag = magazine::Class<cv::gapi::own::Mat, cv::gapi::own::Scalar, cv::detail::VectorRef, cv::detail::OpaqueRef,
+                            cv::gapi::own::RMat::HostMatP>;
```
dmatveev: So why it stores `HostMat` instead of `RMat`??

modules/gapi/test/own/rmat_tests.cpp
```diff
@@ -0,0 +1,178 @@
...
+    virtual GMatDesc desc() const override { return cv::descr_of(m_mat); }
+};
+
+TEST(RMat, Test)
+{
+    cv::Size sz{8, 8};
+    cv::Mat mat(sz, CV_8UC1);
+    cv::randu(mat, cv::Scalar::all(127), cv::Scalar::all(40));
+
+    RMat rmat(std::make_shared<RMatImplCpuWRef>(mat));
```
dmatveev: The API to create and wrap external memory to RMat needs to be more friendly.

modules/gapi/test/own/rmat_tests.cpp
```diff
@@ -0,0 +1,178 @@
...
+    auto omat = rmat.access(RMat::AccessType::RW)->getMat();
+    auto matFromDevice = cv::gapi::own::to_ocv(omat);
+    EXPECT_TRUE(cv::descr_of(mat) == cv::descr_of(matFromDevice));
+    EXPECT_EQ(0, cv::countNonZero(abs(mat-matFromDevice)));
```
dmatveev: PLEASE start using cv::norm(INF) instead of this

asmorkalov:
@rgarnov friendly reminder

modules/gapi/include/opencv2/gapi/own/rmat.hpp
rgarnov: `HostMat` is a wrapper of a memory which is mapped to a host and can be accessed by the host. All this methods can be moved to `RMat` not to introduce new entities. Done it this way to make `RMat` as lightweight as possible: if we don't need to map memory to host and just want to describe some workload on some remote memory, we can do it without `HostMat` at all and use only `RMat`s current interface

modules/gapi/include/opencv2/gapi/own/rmat.hpp
rgarnov: We need some sort of `flush` method to sync data with remote device but we can't extend cv::Mat with that

modules/gapi/test/own/rmat_tests.cpp
rgarnov: Will fix, thanks!

modules/gapi/test/own/rmat_tests.cpp
rgarnov: I can propose only some template `make_rmat` which call `std::make_shared` inside:
```
auto rmat = cv::make_rmat<RMatImplCpuWRef>(mat);
```

rgarnov: Just to clarify: there always user defined constructor for `Impl` class, so it's not required cv::Mat to be in the signature, it can be some another handle or so

modules/gapi/src/backends/common/gbackend.hpp
rgarnov: I didn't touch executor's interaction with backends and islands, they still know nothing about RMat and work as they work previously. Executor accesses RMat, extracts own::Mat from it and pass extracted own::Mat to islands

modules/gapi/include/opencv2/gapi/own/rmat.hpp
dmatveev: I'd clearly express that we have a method to obtain data and a method to upload data.
The adapter could also indicate if the obtained data is a copy (then there's no sense in writing to this memory) or a mapping (then this memory is writable). That's it :)

-----

Maybe handling this at user code side is not very useful though. Need to prototype and see how it looks like

modules/gapi/src/backends/common/gbackend.hpp
dmatveev: They should start supporting it natively, this is the key.

modules/gapi/test/own/rmat_tests.cpp
dmatveev: May be, but this `make_rmat` is STL-ish not OpenCV-ish :)

dmatveev: ... it applies to our existing code as well

modules/gapi/test/own/rmat_tests.cpp
dmatveev: Please fix :)

asmorkalov:
@rgarnov @dmatveev Do you have a chance to finish the PR?

dmatveev:
@asmorkalov it must happen soon

PR #16525, opened 2020-02-07
changed layerBlobs in dnn.cpp and added DaSiamRPN tracker

ieliz:
Changed layerBlobs in dnn.cpp, cause - seg. faults while using OpenCV with Python 3 (cv.dnn).
Added Python 3 script with DaSiamRPN tracker. All related links inside script. Tracker still WIP.

**WIP**


asmorkalov:
Hello @ieliz thanks for the contribution. Could you provide more details on the issue you've fixed: - conditions when it appears or some reproducer. The best solution is test case that indicates the bug without fix. Also, please pay attention on CI status.

ieliz:
PR reopened :https://github.com/opencv/opencv/pull/16554

PR #16549, opened 2020-02-10
fix wrong memory allocation

rayonnant14:
There was mistake in use new() function without delete() function. I fixed it using vector<int> instead dynamic array


alalek:
[Valgrind build](http://pullrequest.opencv.org/buildbot/builders/3_4_valgrind-lin64-debug/builds/135).

PR #16555, opened 2020-02-11
Add "Privacy masking camera" sample for the upcoming posting

dmatveev:

The feature is a sample and will be documented separately, planned for posting on "opencv.org". Same applies to the "related work".

```

test_modules:Custom=gapi
test_modules:Custom Win=gapi
test_modules:Custom Mac=gapi

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*

```


modules/gapi/samples/privacy_masking_camera.cpp
```diff
@@ -0,0 +1,215 @@
...
+    "This is an OpenCV-based version of Privacy Masking Camera example";
+const std::string keys =
+    "{ h help |   | print this help message }"
+    "{ input  |   | Path to an input video file }"
+    "{ platm  |   | IE vehicle/license plate detection model IR }"
+    "{ platw  |   | IE vehicle/license plate detection model weights }"
+    "{ platd  |   | IE vehicle/license plate detection model device }"
+    "{ facem  |   | IE face detection model IR }"
+    "{ facew  |   | IE face detection model weights }"
+    "{ faced  |   | IE face detection model device }"
```
alalek: It would be nice to have default values for model-related parameters and/or some instructions how to get them

modules/gapi/samples/privacy_masking_camera.cpp
```diff
@@ -0,0 +1,215 @@
...
+GAPI_OCV_KERNEL(OCVParseSSD, ParseSSD) {
+    static void run(const cv::Mat &in_ssd_result,
+                    const cv::Mat &in_frame,
+                    const int      filter_label,
+                    std::vector<cv::Rect> &out_objects) {
+        const auto &in_ssd_dims = in_ssd_result.size;
+        CV_Assert(in_ssd_dims.dims() == 4u);
+
+        const int MAX_PROPOSALS = in_ssd_dims[2];
+        const int OBJECT_SIZE   = in_ssd_dims[3];
```
alalek: `OBJECT_SIZE >= 7` or overrun below (strong equality for SSD output)

modules/gapi/samples/privacy_masking_camera.cpp
```diff
@@ -0,0 +1,215 @@
...
+            const float confidence = data[i * OBJECT_SIZE + 2];
+            const float rc_left    = data[i * OBJECT_SIZE + 3];
+            const float rc_top     = data[i * OBJECT_SIZE + 4];
+            const float rc_right   = data[i * OBJECT_SIZE + 5];
+            const float rc_bottom  = data[i * OBJECT_SIZE + 6];
+
+            if (image_id < 0.f) {
+                break;    // marks end-of-detections
+            }
+            if (confidence < 0.5f) {
```
alalek: BTW, It would be nice to configure confidence threshold too (in the future)

modules/gapi/samples/privacy_masking_camera.cpp
```diff
@@ -0,0 +1,215 @@
...
+    }
+};
+
+GAPI_OCV_KERNEL(OCVToMosaic, ToMosaic) {
+    static void run(const std::vector<cv::Rect> &in_plate_rcs,
+                    const std::vector<cv::Rect> &in_face_rcs,
+                          std::vector<cv::gapi::wip::draw::Prim> &out_prims) {
+        out_prims.clear();
+        const int BLOCK_SIZE = 24;
+        const auto cvt = [](cv::Rect rc) {
```
alalek: const reference?

modules/gapi/samples/privacy_masking_camera.cpp
```diff
@@ -0,0 +1,215 @@
...
+            // Align the mosaic region to mosaic block size
```
alalek: Perhaps this mosaic bug has been fixed.
Anyway, It make sense to limit minimal size (expand to BLOCK_SIZE).

modules/gapi/samples/privacy_masking_camera.cpp
```diff
@@ -0,0 +1,215 @@
...
+#include <opencv2/gapi/streaming/cap.hpp>
+#include <opencv2/highgui.hpp>
+
+const std::string about =
+    "This is an OpenCV-based version of Privacy Masking Camera example";
+const std::string keys =
+    "{ h help |   | print this help message }"
+    "{ input  |   | Path to an input video file }"
+    "{ platm  |   | IE vehicle/license plate detection model IR }"
+    "{ platw  |   | IE vehicle/license plate detection model weights }"
```
dkurt: It's convenient for OpenVINO samples to specify just path to `.xml` and keep both `.xml` and `.bin` with same same name in the same directory. So probably we can just estimate `.bin` path from `.xml` and then only `plat` flag is needed.

modules/gapi/samples/privacy_masking_camera.cpp
```diff
@@ -0,0 +1,215 @@
...
+    TS      now() const { return std::chrono::high_resolution_clock::now(); }
+    double  tick() const { return std::chrono::duration_cast<MS>(now() - started).count(); }
+    Elapsed elapsed() const { return Elapsed{tick()}; }
+    double  fps(std::size_t n) const { return static_cast<double>(n) / (tick() / 1000.); }
+};
+std::ostream& operator<<(std::ostream &os, const Avg::Elapsed &e) {
+    os << e.mm << ':' << (e.ss - 60*e.mm);
+    return os;
+}
+} // namespace
```
dkurt: Why not [TickMeter](https://docs.opencv.org/master/d9/d6f/classcv_1_1TickMeter.html)?

modules/gapi/samples/privacy_masking_camera.cpp
```diff
@@ -0,0 +1,215 @@
...
+    }
+    const std::string input = cmd.get<std::string>("input");
+    const bool no_show = cmd.get<bool>("noshow");
+    const bool run_trad = cmd.get<bool>("trad");
+
+    cv::GMat in;
+    cv::GMat blob_plates = cv::gapi::infer<custom::VehLicDetector>(in);
+    cv::GMat blob_faces  = cv::gapi::infer<custom::FaceDetector>(in);
+    cv::GArray<cv::Rect> rc_plates = custom::ParseSSD::on(blob_plates, in, 2);
+    cv::GArray<cv::Rect> rc_faces  = custom::ParseSSD::on(blob_faces, in, -1);
```
alalek: > 2
> -1

Consider extracting of magic numbers and give names for them:
```
const int ssd_label_plate = 2;
const int ssd_label_face = -1;
```

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: It is not about bug, it is more about the visual appearance :)

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: Ok will do

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: Never mind we have a TickMeter :D Ok will port

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: Would assert be enough here?

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: I plan to add a new G-API operation to parse the SSD output, and yes it will have this among its parameters

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: -1 is not a face label BTW, it assumes no filtering is required (a single-class detector).
Maybe an inline comment would be enough?

modules/gapi/samples/privacy_masking_camera.cpp
```diff
@@ -0,0 +1,215 @@
...
+            exec(in_frame, out_frame);
+            if (!no_show) {
+                cv::imshow("Out", out_frame);
+                cv::waitKey(1);
+            }
+            frames++;
+        } while (cap.read(in_frame));
+    } else {
+        auto pipeline = graph.compileStreaming(cv::compile_args(kernels, networks));
+        pipeline.setSource(cv::gapi::wip::make_src<cv::gapi::wip::GCaptureSource>(input));
```
dkurt: Can it work with `cv::VideoCapture`? I mean, if I have a source which should be preconfigured. In example Logitech C920. It works with 30FPS by default but for achieve 60FPS we need to set JPEG compressing, specific frame size (1280x720 if I'm not mistaken) and specific video backend (V4L2).

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: I remember a couple of DNN samples do that already, do we have a function for this somewhere in utils?

modules/gapi/samples/privacy_masking_camera.cpp
alalek: Yes, I mean assertion check.

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: `GCaptureSource` is built atop of `cv::VideoCapture`. I just did not want to expose `VideoCapture` itself to simplify the resource management (and make it less error-prone). May be what we might actually need is to support more `VideoCapture` parameters in the `GCaptureSource` constructor.

Note this API is still marked `wip` so

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: Not sure what may be defaults for model paths, here you mean defaults for devices?

alalek: model ~~paths~~ names at first
devices - `CPU` by default can be used too (with example of possible values in parameter's description: CPU/GPU/VPU/etc)

Goal: get working sample with minimal command line (perhaps with copying of model files...).

Also you may add some examples into `about` string.

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: Added my own

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: Kept path so far

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: Done

dmatveev:
@alalek @dkurt thanks for the review, all done. Anything else?

modules/gapi/samples/privacy_masking_camera.cpp
alalek: Use `name` of the model by default (which just works if models are already in current working directory).
User may override them with own path.

Main idea of that is showing what User should specify here.

modules/gapi/samples/privacy_masking_camera.cpp
```diff
@@ -0,0 +1,216 @@
...
+    "This is an OpenCV-based version of Privacy Masking Camera example";
+const std::string keys =
+    "{ h help |     | print this help message }"
+    "{ input  |     | Path to an input video file }"
+    "{ platm  |     | IE vehicle/license plate detection model (.xml) }"
+    "{ platd  | CPU | IE vehicle/license plate detection model device (for OpenVINO IE, e.g. CPU, GPU, VPU, ...) }"
+    "{ facem  |     | IE face detection model (.xml) }"
+    "{ faced  | CPU | IE face detection model device (for OpenVINO IE, e.g. CPU, GPU, VPU, ...) }"
+    "{ trad   |     | Run processing in traditional (non-pipelined) way }"
+    "{ noshow |     | Don't display UI (improves performance) }";
```
alalek: Please use "false" default value for bool parameters.

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: What is the name of the model?

alalek: Some name of model which can be used with this sample (which knows label=2), like `vehicle-license-plate-detection-barrier-0106` from OMZ.

dmatveev: So then you want this sample to append .xml automatically?
I found this counter-intuitive if the model is placed in some other directory. Then the "model name" effectively becomes a model path

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: Done

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: Renamed its description to "Path to <..>"

alalek: > append .xml automatically

No.

---

Path of which model user should write here?
How many models users should try before they run demo successfully?
Select one that works successfully with this sample and write here as "default" (filename only).

> or some instructions how to get them

This is that I want to see here. "Model name" (from OMZ) is an instruction.

dmatveev: Ok so now I've finally got it :)

dkurt:
@alalek, it looks like @kangzixiang is out of organization but approved this PR :)

modules/gapi/samples/privacy_masking_camera.cpp
dmatveev: @alalek done

alalek: What is a purpose of strings below?

```
G_API_NET(VehLicDetector, <cv::GMat(cv::GMat)>, "vehicle-license-plate-detector");
G_API_NET(FaceDetector,   <cv::GMat(cv::GMat)>,                  "face-detector");
```

Model names?
If they are just internal names of operations, then my suggestion is:

```
    "{ platm  | vehicle-license-plate-detection-barrier-0106.xml | Path to OpenVINO IE model .xml file (plate detection) }"
    "{ platd  | CPU   | Target device for plate detection model (e.g. CPU, GPU, VPU, ...) }"
    "{ facem  | face-detection-adas-0001.xml | Path to OpenVINO IE model .xml file (face detection) }"
    "{ faced  | CPU   | Target device for face detection model (e.g. CPU, GPU, VPU, ...) }"
```

dmatveev: > If they are just internal names of operations

This

> then my suggestion is

OK

dmatveev: @alalek done

dmatveev:
@alalek thanks!

sklautau:
Where can I find the documentation in order to use it, please?

dmatveev:
> Where can I find the documentation in order to use it, please?

@sklautau  The publishing is pending... If you need help with running this demo, you can ask questions here and I will answer

PR #16563, opened 2020-02-12
Add samples code for tutorials (Load, Modify, and Save an Image)

sunwoo2:
I found that there is no source code link and repository in OpenCV Tutorials/Introduction to OpenCV/ Load, Modify, and Save an Image. So i've added it.

https://docs.opencv.org/3.4.0/db/d64/tutorial_load_save_image.html

```
```

### DIFF
--- a/samples/cpp/tutorial_code/introduction/modify_image/modify_image.cpp
+++ b/samples/cpp/tutorial_code/introduction/modify_image/modify_image.cpp
@@ -1,0 +1,55 @@
+//! [includes]
+#include <opencv2/opencv.hpp>
+//! [includes]
+
+//! [namespace]
+using namespace cv;
+//! [namespace]
+
+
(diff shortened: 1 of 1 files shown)


dkurt:
Am I right that you wanted to replace it to code snippets from file?

That means you need to modify https://github.com/opencv/opencv/blob/3.4/doc/tutorials/introduction/load_save_image/load_save_image.markdown as well

sunwoo2:
I couldn't see 'doc' folder well enough because I concentrate 'samples' folder and docs homepage.
thank you for your feedback. I think the github address you linked can maintain current state.
 
-----Original Message-----

asmorkalov:
@sunwoo2 Do you have progress on the patch?

sunwoo2:
@asmorkalov No.

PR #16565, opened 2020-02-12
update video.py

goddl717:
I find some error that sample window exit button doesn't work, so i add commet " Warning:You can't exit using exit button. ".


alalek:
There is no "exit" button.
System UI drawing close window button. In case of multiple windows it should not trigger automatic exit.

Use `ESC` to exit from application loop.

Issue #14713, opened 2019-06-04
Assertion failed (inputs.size()) in cv::dnn::dnn4_v20190122::Layer::getMemoryShapes


Aeroxander:
##### System information (version)

- OpenCV => 4.1.0 C++
- Operating System / Platform => Mac OS High Sierra (also had this issue on Windows 10)
- Compiler => XCode

##### Detailed description

So I have 3 PyTorch models where 2 of them give me trouble when loading in ONNX, the inference works fine in PyTorch and the conversion is all ok too.

So in the first model I get this error when reading the ONNX model in OpenCV:

> Assertion failed (inputs.size()) in cv::dnn::dnn4_v20190122::Layer::getMemoryShapese

With the other model I get this error:

> Blob 64 not found in const blobs

This has all something to do with the input where in the first error the inputs.size() is 0 and in the second model the blob isn't big enough somehow. 
I don't know which framework or action I did I can pinpoint the error to..

##### Steps to reproduce
Read the decoder ONNX model

##### Standalone project
https://github.com/Aeroxander/decodererror

alalek:
> keep them both as a global variable

Due static initialization order fiasco it is a bad practice in general.

Please provide:
- complete standalone reproducer (without extra dependencies outside of OpenCV)
- and stack trace of your crashes (to ensure that problem is the same)

Aeroxander:
@alalek My mistake, I actually get an error just for that 1 wrong ONNX model, it just can't parse it and gives an "Can't read ONNX model file" error.

As the first one works fine, I suppose the conversion to my ONNX model went fine, although something may have not converted properly. Here is some info of the model:

format: ONNX v4
producer: pytorch 1.1
imports: ai.onnx v9
name: torch-jit-export

Should I try out ONNX v5 that may be more compatible with converting from PyTorch to ONNX, or how can I discover what's wrong with my model (could still something be wrong with ONNX)?

Here is an visualized Netron picture of it if that helps:

![decoder4Kr31](https://user-images.githubusercontent.com/11189629/58924188-b5ffe000-86f7-11e9-8f9d-6cc52ab5970e.png)

Aeroxander:
Alright, so I'm just trying everything out right now, but the model that doesn't give any errors also doesn't give any models if I try to convert the ONNX model to a Tensorflow model, but the other one gives this error:

[...]

PR #16573, opened 2020-02-13
ONNX: upsample subgraph fusion added

### LINKED ISSUE #16323: (-215:Assertion failed) inputs.size() in function cv::dnn::dnn4_v20191202::Layer::getMemoryShapes

- OpenCV => 4.2.0-pre
- Compiler => Visual Studio 2016

##### Detailed description

When im trying to load ONNX model im getting error:

F:\\OpenCV\\opencv-master\\modules\\dnn\\src\\dnn.cpp:4479: error: (-215:Assertion failed) inputs.size() in function 'cv::dnn::dnn4_v20191202::Layer::getMemoryShapes'\n" ...}

#### Model architecture:
HRNet
#### Model was exported from PyTorch to ONNX with following code:
```C++
    model.load_state_dict(torch.load(args.model_file))
    data = torch.zeros((1, 3, 288, 384)).cuda()
    model.eval()
    model.cuda()
    model.float()

    torch.onnx.export(model, data, 'test.onnx')
```
#### Model properties and first input layer:
image

#### About error
Looks like error appears in onnx_importer.cpp at line
```C++
        Ptr<Layer> layer = dstNet.getLayer(id);
        layer->getMemoryShapes(layerInpShapes, 0, layerOutShapes, layerInternalShapes);
```
on layer id = 113

##### Steps to reproduce:

1. [Upload_model](https://drive.google.com/open?id=1qQ5oFBk8UjzK0FAYLBaCYcGYfKMm5AXB)
2. load in:
```C++
auto network = readNetFromONNX("test.onnx");
```

### LINKED ISSUE #16475: error when load onnx model
I am trying to load a onnx model which is converted from pytorch. 

`import cv2`
`model = cv2.dnn.readNetFromONNX('tmp.onnx')`

cv2.error: OpenCV(4.1.2) /io/opencv/modules/dnn/src/dnn.cpp:525: error: (-2:Unspecified error) Can't create layer "166" of type "Cast" in function 'getLayerInstance'

here is my model
[https://www.dropbox.com/s/sg0n1kdpbbio848/tmp.onnx?dl=0](url)

anyone have idea how to fix this problem? Thanks a lot.

ashishkrshrivastava:
**Merge with extra**: opencv/opencv_extra#707
resolves #16323
resolves #16475 
resolves [#14713](https://github.com/opencv/opencv/issues/14713)

` opencv_extra=opencvonnx`

- [x]  I agree to contribute to the project under OpenCV (BSD) License.

### This PR adds

Subgraph fusion for `upsample` layer when exported from pytorch 1.3
Screenshot from 2020-02-13 17-43-16

If model is exported using pytorch 1.2, it does not produce two extra `cast` layers after `Gather` layer. 
Screenshot from 2020-02-13 17-46-06

The additions I made are according to latest version of Pytorch. 

```
allow_multiple_commits=1
```


ashishkrshrivastava:
> There is accuracy test, performance test and test data in opencv_extra repository, if applicable
> Patch to opencv_extra has the same branch name.

@dkurt , do both opencv and opencv_extra branches need to have same name ?

```
unknown file: Failure
C++ exception with description "OpenCV(3.4.9-dev) /build/precommit_macosx/3.4/opencv/modules/ts/src/ts.cpp:1052: error: (-2:Unspecified error) OpenCV tests: Can't find required data file: dnn/onnx/models/upsample_unfused.onnx in function 'findData'
" thrown in the test body.
```

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,9 +145,92 @@ class SoftMaxSubgraph : public Subgraph
...
+            opencv_onnx::NodeProto* node = upsample.dynamicCast<ONNXNodeWrapper>()->node;
+
+            CV_Assert(node->op_type() == "Upsample");
+
+            for (int i = 0; i < node->attribute_size(); i++)
+            {
+                opencv_onnx::AttributeProto attr = node->attribute(i);
+                if (attr.name() != "mode")
+                    continue;
+                mode = attr.s();
```
dkurt: Am I right that only Upsample node has "mode" attribute?

ashishkrshrivastava: yeah.

dkurt:
@ashishkrshrivastava, by default yes. If you already have one - add it's name to PR description

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -974,6 +974,38 @@ void ONNXImporter::populateNet(Net dstNet)
...
+                layerParams.set("zoom_factor_y", B1.at<int>(0));
+                layerParams.set("zoom_factor_x", B2.at<int>(0));
+            }
+            else if(indices2.at<int>(0) == 2 && indices1.at<int>(0) == 3)
+            {
+                layerParams.set("zoom_factor_y", B2.at<int>(0));
+                layerParams.set("zoom_factor_x", B1.at<int>(0));
+            }
+            replaceLayerParam(layerParams, "mode", "interpolation");
+        }
```
dkurt: Can we make a fused layer to match existing implementation?

https://github.com/opencv/opencv/blob/221564d12c6183f2d15515e38a7d2d0f0c4bdc12/modules/dnn/src/onnx/onnx_importer.cpp#L958

ashishkrshrivastava: `Scales` are attribute in `Upsample` layer. 
But inputs `B` of `mul `layers in unfused subgraph. So either we need to extract them in graph_simplifier.cpp or in onnx_importer.cpp (current implementation).

dkurt: Yeah, I mean we can extract the exact values from them at finalize step and put to scales

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
dkurt: I think we can remove this check because if Upsample node has no mode attribute - it's a problem of Upsample node but not a criteria of subgraph matching

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,9 +145,91 @@ class SoftMaxSubgraph : public Subgraph
...
+    }
+
+    virtual void finalize(const Ptr<ImportGraphWrapper>&,
+                          const Ptr<ImportNodeWrapper>& fusedNode,
+                          std::vector<Ptr<ImportNodeWrapper> >&) CV_OVERRIDE
+    {
+        opencv_onnx::NodeProto* node = fusedNode.dynamicCast<ONNXNodeWrapper>()->node;
+        opencv_onnx::AttributeProto* attr = node->add_attribute();
+        attr->set_name("mode");
+        attr->set_s(mode);
```
dkurt: It seems to me that fused node is a last node in subgraph (Upsample) so it should already has this attribute

ashishkrshrivastava: yeah, you were right. Thanks.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +145,161 @@ class SoftMaxSubgraph : public Subgraph
...
+    }
+
+    virtual bool match(const Ptr<ImportGraphWrapper>& net, int nodeId,
+                       std::vector<int>& matchedNodesIds,
+                       std::vector<int>& targetNodesIds) CV_OVERRIDE
+    {
+        if (Subgraph::match(net, nodeId, matchedNodesIds, targetNodesIds))
+            return true;
+        return false;
+    }
```
dkurt: We can remove this override now

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +145,161 @@ class SoftMaxSubgraph : public Subgraph
...
+        {
+            opencv_onnx::TensorProto tensor_proto = input_nodes[i]->attribute(0).t();
+            nodes_attributes[i] = getMatFromTensor(tensor_proto);
+        }
+        CV_Assert(nodes_attributes[0].total() == 2 && nodes_attributes[1].total() == 1 && nodes_attributes[2].total() == 1);
+        attr->set_name("scales");
+        attr->add_ints((int)nodes_attributes[0].at<float>(0));
+        attr->add_ints((int)nodes_attributes[0].at<float>(1));
+        attr->add_ints((int)nodes_attributes[1].at<float>(0));
+        attr->add_ints((int)nodes_attributes[2].at<float>(0));
```
dkurt: Let's match our fused layer to "Caffe2" line Upsample:

https://github.com/opencv/opencv/blob/f9442aec16484ea322b30f55479462f78f7f13df/modules/dnn/src/onnx/onnx_importer.cpp#L881-L899

So you will only need to set `height_scale` and `width_scale`.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: Done !

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: Done !

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +145,153 @@ class SoftMaxSubgraph : public Subgraph
...
+        input_nodes[0] = inputs[5].dynamicCast<ONNXNodeWrapper>()->node; //inpIndex
+        input_nodes[1] = inputs[2].dynamicCast<ONNXNodeWrapper>()->node; //B1
+        input_nodes[2] = inputs[4].dynamicCast<ONNXNodeWrapper>()->node; //B2
+        for(int i = 0; i < 3; i++)
+        {
+            opencv_onnx::TensorProto tensor_proto = input_nodes[i]->attribute(0).t();
+            nodes_attributes[i] = getMatFromTensor(tensor_proto);
+        }
+        CV_Assert((int)nodes_attributes[0].at<float>(0) == 1 && (int)nodes_attributes[0].at<float>(1) == 1 &&
+            nodes_attributes[1].total() == 1 && nodes_attributes[2].total() == 1);
```
dkurt: I'd like to recommend to simplify it.

`nodes_attributes[0]` is never used and I believe that we can extract values from B1 and B2 without `getMatFromTensor` (maybe not, depends of field type). Some thing like `B1->float(0)` or similar.

ashishkrshrivastava:
```
if (datatype == opencv_onnx::TensorProto_DataType_FLOAT) {

        if (!tensor_proto.float_data().empty()) {
            const ::google::protobuf::RepeatedField<float> field = tensor_proto.float_data();
            Mat(sizes, CV_32FC1, (void*)field.data()).copyTo(blob);
        }
        else {
            char* val = const_cast<char*>(tensor_proto.raw_data().c_str());
            Mat(sizes, CV_32FC1, val).copyTo(blob);
        }
    }
```
B1 falls under second category, i.e `raw_data()` so I couldn't find any simple function to extract it. Please recommend I missed it.

ashishkrshrivastava: `nodes_attribute[0]` has scales of batch and channels. [batch, channels, Height, width]. I understand that it is always [1,1, scale_height, scale_width]. But I still kept it to check in case they are not 1. 
Not sure to keep them or remove them. What do you say?

dkurt: Ok, then let's rename them properly:

```
setFusedNode("Upsample", input, ?, scaleH, ?, scaleW, scaleNC);
```
BTW, what is a size of `nodes_attributes[0]`?

ashishkrshrivastava: @dkurt, can I solve two issues in this PR. issue [16475](https://github.com/opencv/opencv/issues/16475) has similar subgraph of Resize.
Screenshot from 2020-02-16 18-57-25
while current subgraph of upsample
Screenshot from 2020-02-13 17-43-16
and we may find some other variations in future. So, I am thinking of creating a layer of this part of subgraph 
Screenshot from 2020-02-16 19-00-03

which is common in both and will use this fused as a part of both subgraphs.

ashishkrshrivastava: > Ok, then let's rename them properly:
> 
> ```
> setFusedNode("Upsample", input, ?, scaleH, ?, scaleW, scaleNC);
> ```
> 
> BTW, what is a size of `nodes_attributes[0]`?

`nodes_attributes[0]` has [1.0, 1.0].

ashishkrshrivastava: Both subgraphs are obtained from same pytorch layer `Upsample` but with different versions. version of subgraph of current issue is 9 whereas vesion of subgraph of issue [16475](https://github.com/opencv/opencv/issues/16475) is 11.

dkurt: @ashishkrshrivastava, if you don't mind, let's enable these two subgraph in this PR. This way it could help to see how they similar.

ashishkrshrivastava: @dkurt , I am not able to produce subgraph same as in model in [16475](https://github.com/opencv/opencv/issues/16475). I am getting this subgraph with the same code which is little different.
Screenshot from 2020-02-16 22-22-31

I have asked the way that model has exported in comments there. Will work on that issue later. 
Actually, slight variation in pytorch version or onnx version are producing different subgraphs. Even sometimes sequential and functional way of pytorch gives different subgraph. 
Please continue with this one only.

dkurt: Ok, got it, thanks!

So let's finish this PR first just not to keep your waiting

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +145,153 @@ class SoftMaxSubgraph : public Subgraph
...
+                          std::vector<Ptr<ImportNodeWrapper> >& inputs) CV_OVERRIDE
+    {
+        opencv_onnx::NodeProto* node = fusedNode.dynamicCast<ONNXNodeWrapper>()->node;
+        CV_Assert(node->input_size() == 4);
+
+        std::vector<opencv_onnx::NodeProto*> input_nodes(3);
+        std::vector<cv::Mat> nodes_attributes(3);
+        input_nodes[0] = inputs[3].dynamicCast<ONNXNodeWrapper>()->node; //scaleNC
+        input_nodes[1] = inputs[1].dynamicCast<ONNXNodeWrapper>()->node; //scaleH
+        input_nodes[2] = inputs[2].dynamicCast<ONNXNodeWrapper>()->node; //scaleW
```
dkurt: I think that we need to check which scale relates to width and which to height (might be swapped). So the axes values are in Gather nodes. I think it's easier to extract values from them at `match` stage and then set in `finalize`. BTW, we can do the same for scales and perform assertions for batch and channels indices.

ashishkrshrivastava:
```
virtual bool match(const Ptr<ImportGraphWrapper>& net, int nodeId,
                       std::vector<int>& matchedNodesIds,
                       std::vector<int>& targetNodesIds)
```
We are extrating values from `Costant` nodes.
And while putting ids in `matchedNodeIds/targetNodeIds` constant nodes were skipped. 
https://github.com/opencv/opencv/blob/f3237fdc6e7eb3a0260e0942af27b9fadc9a11ef/modules/dnn/src/graph_simplifier.cpp#L114-L123
That is why I added them to ` inputs` of fused node so that we can extract the values later. I am not sure how to get id of constant node in match function.

ashishkrshrivastava: Let me add check for axes swap.

ashishkrshrivastava: @dkurt , what to do about this ?
```
Total tests: 8605
Tests failed: 0
Tests passed: 8605
Exception has been thrown. Please see stdio log.
```
https://pullrequest.opencv.org/buildbot/builders/precommit_macosx/builds/23190/steps/test_imgproc/logs/tests%20summary

alalek: @ashishkrshrivastava Please ignore MacOSX builder failures. They are not related to this PR.

ashishkrshrivastava: okay.

dkurt: @ashishkrshrivastava, Maybe we can combine it in this way?
```cpp
if (!Subgraph::match(net, nodeId, matchedNodesIds, targetNodesIds))
    return false;

// Get Constant nodes from "net" using "matchedNodesIds"

return true;
```

in example, from TensorFlow's graphs fusion:

https://github.com/opencv/opencv/blob/4cb9faf6c9c012d603217bc09313d325bb9fb04a/modules/dnn/src/tensorflow/tf_graph_simplifier.cpp#L254-L263

So you can extract scales from there and keep only actual inputs for the fused node

ashishkrshrivastava: ohh, okay. got it.

ashishkrshrivastava:
@dkurt 
```
node = onnx.helper.make_node(
    'Resize',
    inputs=['X', 'roi', 'scales'],
    outputs=['Y'],
    mode='cubic',
)

data = np.array([[[
    [1, 2, 3, 4],
    [5, 6, 7, 8],
    [9, 10, 11, 12],
    [13, 14, 15, 16],
]]], dtype=np.float32)

roi = np.array([], dtype=np.float32)
scales = np.array([1.0, 1.0, 0.8, 0.8], dtype=np.float32)

# [[[[ 1.47119141  2.78125     4.08251953]
#    [ 6.71142578  8.02148438  9.32275391]
#    [11.91650391 13.2265625  14.52783203]]]]
output = interpolate_nd(
    data, cubic_coeffs, scale_factors=scales).astype(np.float32)

expect(node, inputs=[data, roi, scales], outputs=[output],
```
Here, In ONNX resize node example ROI is empty, and hence ONNX model contain an empty constant node. I want your suggestion regarding this empty constant node.
It need to be handled otherwise it would create problem here, 
```
Mat getMatFromTensor(opencv_onnx::TensorProto& tensor_proto)
{
    CV_Assert(!tensor_proto.raw_data().empty() || !tensor_proto.float_data().empty()
                    || !tensor_proto.double_data().empty() || !tensor_proto.int64_data().empty());
```
I think we can add some dummy data or can delete the node. Deleting node is complex.

alalek:
> Conflicting files

Please rebase patch on latest commits from 3.4 branch:
```
git fetch updstream
git rebase -i upstream/3.4
```

( looks like it conflicts with #16472 - also please split related "Resize" test on smaller parts )

ashishkrshrivastava:
> > Conflicting files
> 
> Please rebase patch on latest commits from 3.4 branch:
> 
> ```
> git fetch updstream
> git rebase -i upstream/3.4
> ```
> 
> ( looks like it conflicts with #16472 - also please split related "Resize" test on smaller parts )

Okay, thanks. Let me do that.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +145,282 @@ class SoftMaxSubgraph : public Subgraph
...
+                          std::vector<Ptr<ImportNodeWrapper> >& inputs) CV_OVERRIDE
+    {
+        opencv_onnx::NodeProto* node = fusedNode.dynamicCast<ONNXNodeWrapper>()->node;
+        opencv_onnx::AttributeProto* attr_height = node->add_attribute();
+        attr_height->set_name("height_scale");
+        attr_height->set_i(scaleH);
+        opencv_onnx::AttributeProto* attr_width = node->add_attribute();
+        attr_width->set_name("width_scale");
+        attr_width->set_i(scaleW);
+    }
```
dkurt: Looks like finalize part is the same for all the cases. What do you think about creating a base class which implements `finalize` in this way:
```cpp
    virtual void finalize(const Ptr<ImportGraphWrapper>&,
                          const Ptr<ImportNodeWrapper>& fusedNode,
                          std::vector<Ptr<ImportNodeWrapper> >& inputs) CV_OVERRIDE
    {
        opencv_onnx::NodeProto* node = fusedNode.dynamicCast<ONNXNodeWrapper>()->node;
        opencv_onnx::AttributeProto* attr = node->add_attribute();
        attr->set_name("scales");
        attr->add_ints(scaleH);
        attr->add_ints(scaleW);
    }
```
so all the subgraphs will just override `match` and have different constructors?

modules/dnn/src/graph_simplifier.cpp
```diff
@@ -110,13 +111,15 @@ bool Subgraph::match(const Ptr<ImportGraphWrapper>& net, int nodeId,
             if (nodes[inputNodes[j]].empty())  // Unknown input node type.
                 continue;
             nodeId = getInputNodeId(net, node, j);
+            if (nodeId == -1)
+                return false;
```
dkurt: That's strange. this method trying to find `node.input[j]` in the network `net`. When it can cannot find a node?

modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -316,6 +316,12 @@ TEST_P(Test_ONNX_layers, Resize)
     testONNXModels("resize_bilinear");
 }
 
+TEST_P(Test_ONNX_layers, ResizeUnfused)
+{
+    testONNXModels("upsample_unfused");
+    testONNXModels("resize_nearest_unfused");
+}
```
dkurt: Do you have test data for all the three subgraphs? Or just for two?

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: Yeah, I was thinking the same. just wanted to have your view. I will add that.

modules/dnn/test/test_onnx_importer.cpp
ashishkrshrivastava: I have test for only two i.e.` UpsampleSubgraph` and `ResizeSubgraph` and first one is getting tested in both tests. Actually these two are subgraph of upsample from differnet opset version (9 and 11). First (`ExtractScalesSubgraph`) is common in both, Actually `ExtractScalesSubgraph` will be part of each subgraph of upsample/Resize from any opset version.

modules/dnn/src/graph_simplifier.cpp
ashishkrshrivastava: Screenshot from 2020-02-20 07-48-37

```
names of all nodes are :
name - tensor   Type - 
name - 1        Type - Split. 
name - 3        Type - Concat 
```
And the input names of `Concat` are 1, 2 and there is no node with name 2.

Since `ExtractScalesSubgraph` fuse this subgraph to node `Concat` ang in split test subgraph Concat layer is present, it will try to match with 'Concat' layer of `ExtractScalesSubgraph`.
Screenshot from 2020-02-16 19-00-03
And there is no node with name `2` in split test subgraph. Split layer has output `1` and `2` but it is named as `1`. 

Before program was stopping at line 75.
https://github.com/opencv/opencv/blob/4b0d3316f6e1437c1fee98934f6036e96d1d321b/modules/dnn/src/graph_simplifier.cpp#L70-L75

but in my view it shouldn't stop. It must simply move on. 
https://github.com/opencv/opencv/blob/23ab3cd59045c0d583a9f233b2f3885184101c9a/modules/dnn/src/graph_simplifier.cpp#L70-L76
https://github.com/opencv/opencv/blob/23ab3cd59045c0d583a9f233b2f3885184101c9a/modules/dnn/src/graph_simplifier.cpp#L113-L115

So, in conclusion, this problem would arise if some node giving more than one output to any layer taking those as inputs. just in case of `split` and `concat`  in split test subgraph. 
Pls correct me if i am wrong.

dkurt: Can you please specify what is Split test? Looks like in this PR there is no subgraphs with Split layer?

ashishkrshrivastava: ohh sorry. I am talking about these tests. 
https://github.com/opencv/opencv/blob/4d14307f5d4aaabe9af29e72c17102cce50d883a/modules/dnn/test/test_onnx_importer.cpp#L392-L401
and the image shared is of split_1.
Screenshot from 2020-02-20 07-48-37

ashishkrshrivastava: These test were failing because of above mentioned reason.

dkurt: Got it, thanks!
So maybe we can modify `getInputNodeId` in this way?

Due we build patterns without output indices, so we can return a node which produces one of the outputs.

In this example with Split,

```
         0
       +--->
split -+     concat
       +--->
         1
```

`getInputNodeId(net, concatNode, 0) -> Split`
and
`getInputNodeId(net, concatNode, 1) -> Split`

So this method checks all the output names for nodes and if it matches - return it.

ashishkrshrivastava: But there in no function like `getNumOutputs()` or `getOutputName()` in `class ImportNodeWrapper
` so I think to get output names we need to add these two functions in `class ImportNodeWrapper` ?

dkurt: Agree, `int getNumOutputs()` and add an index for `getOutputName(int idx = 0)`

ashishkrshrivastava: @dkurt , is it fine to add 
```
    virtual int getNumOutputs() const
    {
        return 0;
    }
    virtual std::string getOutputName(int idx = 0) const
    {
        return "";
    }
```
instead of 

   ```
virtual int getNumOutputs() const = 0;
virtual std::string getOutputName(int idx) const = 0;
```
in graph_simplifier.hpp. 

Otherwise, we need to modify `class TFNodeWrapper : public ImportNodeWrapper`
in tf_graph_simplifier.cpp

**Comment by ashishkrshrivastava** [comment]
@dkurt , I just encountered one more problem while loading
`net = cv2.dnn.readNetFromONNX("/home/ashish/Downloads/Object Detection & Image Segmentation/yolov3/yolov3.onnx")`

It attempted to match `SoftmaxSubgraph` ( div node ).

`cv2.error: OpenCV(4.2.0-dev) /home/ashish/opencv/modules/dnn/src/graph_simplifier.cpp:84: error: (-212:Parsing error) Input node with name const_fold_opt__570 not found in function 'getInputNodeId'`

Here, `const_fold_opt__570` is initializer.
Screenshot from 2020-02-21 15-53-22

So, it is neither node or output of some node, I guess. What do you suggest about it?

File: modules/dnn/src/graph_simplifier.cpp [hunk 11]
```diff
@@ -63,14 +63,23 @@ int Subgraph::getInputNodeId(const Ptr<ImportGraphWrapper>& net,
...
+    std::string outName = name ;
     // If operation produces several tensors, they are specified by index
     // after ':' character. In example, "input:0".
-    name = name.substr(0, name.rfind(':'));
+    std:: string nodeName = name.substr(0, name.rfind(':'));
     const int numNodes = net->getNumNodes();
     for (int i = 0; i < numNodes; ++i)
     {
-        if (net->getNodeName(i) == name)
+        if (net->getNodeName(i) == nodeName)
```
dkurt: Node name can be not output name. So let's check only outputs names (let's TF graphs return `getNodeName` there)

dkurt:
@ashishkrshrivastava, I'd like to recommend to move it to separate issue / PR if you do not mind.

ashishkrshrivastava:
> @dkurt , I just encountered a problem while loading
> `net = cv2.dnn.readNetFromONNX("/home/ashish/Downloads/Object Detection & Image Segmentation/yolov3/yolov3.onnx")`
> Here is the [model](https://github.com/onnx/models/tree/master/vision/object_detection_segmentation/yolov3).
> It attempted to match `SoftmaxSubgraph` ( div node ).
> 
> `cv2.error: OpenCV(4.2.0-dev) /home/ashish/opencv/modules/dnn/src/graph_simplifier.cpp:84: error: (-212:Parsing error) Input node with name const_fold_opt__570 not found in function 'getInputNodeId'`
> 
> Here, `const_fold_opt__570` is initializer.
> Screenshot from 2020-02-21 15-53-22
> 
> So, it is neither node nor output of some node, I guess. What do you suggest about it?

@dkurt, Are you talking about moving only this issue to another issue/PR or everything ?

dkurt:
Yeah. So if this PR can already fix two mentioned issues (#16323 and #16475) - let's finish it. Or they also require this SoftMax fusion?

ashishkrshrivastava:
okay. for these two issues we dont need to worry about softamx fusion.

dkurt:
@ashishkrshrivastava, thank you!

modules/dnn/src/graph_simplifier.cpp
ashishkrshrivastava:
```
int Subgraph::getInputNodeId(const Ptr<ImportGraphWrapper>& net,
                             const Ptr<ImportNodeWrapper>& node,
                             int inpId)
{
    CV_Assert(inpId < node->getNumInputs());
    std::string name = node->getInputName(inpId);
    // If operation produces several tensors, they are specified by index
    // after ':' character. In example, "input:0".
    std::string name = name.substr(0, name.rfind(':'));
    const int numNodes = net->getNumNodes();
    for (int i = 0; i < numNodes; ++i)
    {
        if (net->getNodeName(i) == name)
            return i;
        Ptr<ImportNodeWrapper> node = net->getNode(i);
        const int numOutputs = node->getNumOutputs();
        for (int j = 0; j < numOutputs; j++)
        {
            if (node->getOutputName(j) == name)
                return i;
        }
    }
    CV_Error(Error::StsParseError, "Input node with name " + name + " not found");
}
```
@dkurt , is this okay ?

dkurt: I mean due we search input for some layer - it should be an output of another one, so I expected something like
```cpp
    for (int i = 0; i < numNodes; ++i)
    {
        Ptr<ImportNodeWrapper> node = net->getNode(i);
        const int numOutputs = node->getNumOutputs();
        for (int j = 0; j < numOutputs; j++)
        {
            if (node->getOutputName(j) == name)
                return i;
        }
    }
```

ashishkrshrivastava: for TF graphs, below statement will work
```
if (net->getNodeName(i) == name)
       return i;
```
for onnx graphs, we need to check both node name and all output names. Only checking output names wouldnt work. 

 ```
if (net->getNodeName(i) == name)
       return i;
Ptr<ImportNodeWrapper> node = net->getNode(i);
const int numOutputs = node->getNumOutputs();
for (int j = 0; j < numOutputs; j++)
{
      if (node->getOutputName(j) == name)
          return i;
}
```

Comment (dkurt) [reply modules/dnn/src/graph_simplifier.cpp]: So for TF we can just override it
```cpp
int getNumOutputs() {
  return 1;
}

std::string getOutputName(int i) {
  return name;
}
```

ashishkrshrivastava: > I mean due we search input for some layer - it should be an output of another one, so I expected something like
> 
> ```c++
>     for (int i = 0; i < numNodes; ++i)
>     {
>         Ptr<ImportNodeWrapper> node = net->getNode(i);
>         const int numOutputs = node->getNumOutputs();
>         for (int j = 0; j < numOutputs; ++j)
>         {
>             if (node->getOutputName(j) == name)
>                 return i;
>         }
>     }
> ```

Okay, I got it. 
the problem with this codeis that sometimes some nodes have no outputs, like these nodes
Screenshot from 2020-02-21 19-51-29

```
> number of nodes are 3 
> Node name : 0 number of outputs of node are : 0 having name 
> Node name : 1 number of outputs of node are : 0 having name 
> Node name : 2 number of outputs of node are : 1 having name 2 
```

there is no output of node 0 and 1. 
thats why we need to check both node and output names

dkurt: Yeah, but due it's ONNX specific problem - we need to keep it separately from common part. So for TF `getOutputName() === getNodeName` but for ONNX we can check
```cpp
getOutputName(int idx) {
  if (idx == 0)
    return name;
  else 
    output(idx)
}
```

ashishkrshrivastava: @dkurt 

```
int Subgraph::getInputNodeId(const Ptr<ImportGraphWrapper>& net,
                             const Ptr<ImportNodeWrapper>& node,
                             int inpId){
            ...
            Ptr<ImportNodeWrapper> node = net->getNode(i);
            const int numOutputs = node->getNumOutputs();
            ...
}
```

node is nullptr here, so it cannot be acessed using `class ImportNodeWrapper`. 

```
getOutputName(int idx) {
  if (idx == 0)
    return node->name;
  else 
   node->output(idx)
}
```
so defining getOutputName inside ImportNodeWrapper like this wouldnt work.
Node name can be acessed by `net->getNodeName(i)`

ashishkrshrivastava: **tf_graph_simplifier**
```
virtual int getNumOutputs() const CV_OVERRIDE
{
        return 1;
}
std::string getOutputName(int i) const CV_OVERRIDE
{
      return node->name();
}
```
**onnx_graph_simplifier**
```
virtual int getNumOutputs() const CV_OVERRIDE
{
    return node ? node->output_size() : 0;
}
virtual std::string getOutputName(int idx) const CV_OVERRIDE
{
    CV_Assert_N(node, idx < node->output_size());
     return node->output(idx);
}
```
**graph_simplifier**
```
int Subgraph::getInputNodeId(const Ptr<ImportGraphWrapper>& net,
                             const Ptr<ImportNodeWrapper>& node,
                             int inpId)
{
    CV_Assert(inpId < node->getNumInputs());
    std::string name = node->getInputName(inpId);
    // If operation produces several tensors, they are specified by index
    // after ':' character. In example, "input:0".
    name = name.substr(0, name.rfind(':'));
    const int numNodes = net->getNumNodes();
    for (int i = 0; i < numNodes; ++i)
    {
        Ptr<ImportNodeWrapper> node = net->getNode(i);
        int numOutputs = node->getNumOutputs();
        if (numOutputs == 0 && net->getNodeName(i) == name)
            return i;
        for (int j = 0; j < numOutputs; j++)
        {
            if (node->getOutputName(j) == name)
                return i;
        }
    }
    CV_Error(Error::StsParseError, "Input node with name " + name + " not found");
}
```
What about this implementation ?

ashishkrshrivastava:
@dkurt , In issue [16475](https://github.com/opencv/opencv/issues/16475) Unet model is obtained using Pytorch version (1.3) while modifications in this PR are according to Pytorch version (1.4). should I make changes according to version 1.3 or it is fine to continue with version 1.4(latest) ? I am able to load Unet exported using version 1.4 with current modifications.

modules/dnn/src/graph_simplifier.cpp
```diff
@@ -69,8 +69,15 @@ int Subgraph::getInputNodeId(const Ptr<ImportGraphWrapper>& net,
     const int numNodes = net->getNumNodes();
     for (int i = 0; i < numNodes; ++i)
     {
-        if (net->getNodeName(i) == name)
+        Ptr<ImportNodeWrapper> node = net->getNode(i);
+        const int numOutputs = node->getNumOutputs();
+        if (numOutputs == 0 && net->getNodeName(i) == name)
```
dkurt: Can we avoid checking `getNodeName` here? Which tests are failed if we test only `getOutputName` and why?

dkurt:
@ashishkrshrivastava, If this PR can fix the issues - let's focus on current state. We need to be user oriented and if one have 1.4 version - we need to fix it here.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +156,266 @@ class SoftMaxSubgraph : public Subgraph
...
+    virtual bool match(const Ptr<ImportGraphWrapper>& net, int nodeId,
+                       std::vector<int>& matchedNodesIds,
+                       std::vector<int>& targetNodesIds) CV_OVERRIDE
+    {
+        if (FuseUpsampleSubgraph::match(net, nodeId, matchedNodesIds, targetNodesIds))
+        {
+            Ptr<ImportNodeWrapper> scales = net->getNode(matchedNodesIds[0]);
+            opencv_onnx::NodeProto* node = scales.dynamicCast<ONNXNodeWrapper>()->node;
+            opencv_onnx::AttributeProto attr = node->attribute(1);
+            CV_Assert(attr.name() == "scales");
```
dkurt: Let's not force this name - we know that scales are located in this node and the name should not block as to extract them.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +156,266 @@ class SoftMaxSubgraph : public Subgraph
...
+            CV_Assert(attr.ints_size() == 4);
+            scaleW = attr.ints(2);
+            scaleH = attr.ints(3);
+            //constant node id after matchedNodesIds removal.
+            emptyNodeId = matchedNodesIds.front();
```
dkurt: Why we need this? The node will be removed anyway or not?

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +156,266 @@ class SoftMaxSubgraph : public Subgraph
...
+{
+public:
+    FuseUpsampleSubgraph() : scaleH(1), scaleW(1), emptyNodeId(-1) {}
+
+    virtual bool match(const Ptr<ImportGraphWrapper>& net, int nodeId,
+                    std::vector<int>& matchedNodesIds,
+                    std::vector<int>& targetNodesIds) CV_OVERRIDE
+    {
+        return Subgraph::match(net, nodeId, matchedNodesIds, targetNodesIds);
+    }
```
dkurt: Do not override this method if it just calls partent's implementation

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +156,266 @@ class SoftMaxSubgraph : public Subgraph
...
+
+    void finalize(const Ptr<ImportGraphWrapper>& net,
+                          const Ptr<ImportNodeWrapper>& fusedNode,
+                          std::vector<Ptr<ImportNodeWrapper> >& inputs) CV_OVERRIDE
```
dkurt: Please fix indentation

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +156,266 @@ class SoftMaxSubgraph : public Subgraph
...
+        opencv_onnx::AttributeProto* attr = node->add_attribute();
+        attr->set_name("scales");
+        attr->add_ints(1);
+        attr->add_ints(1);
+        attr->add_ints(scaleH);
+        attr->add_ints(scaleW);
+        //remove empty constant node
+        if(emptyNodeId != -1)
+            net->removeNode(emptyNodeId);
+    }
```
dkurt: Please set just `height_scale` and `width_scale` attributes

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +156,266 @@ class SoftMaxSubgraph : public Subgraph
...
+            std::vector<cv::Mat> nodes_attributes;
+            for (int i = matchedNodesIds.front(); i < matchedNodesIds.back(); i++)
+            {
+                if (std::find(matchedNodesIds.begin(), matchedNodesIds.end(), i) == matchedNodesIds.end())
+                {
+                    opencv_onnx::NodeProto* constant_node = net->getNode(i).dynamicCast<ONNXNodeWrapper>()->node;
+                    opencv_onnx::TensorProto tensor_proto = constant_node->attribute(0).t();
+                    nodes_attributes.push_back(getMatFromTensor(tensor_proto));
+                }
+            }
```
dkurt: What this loop actually does?

modules/dnn/src/graph_simplifier.cpp
ashishkrshrivastava: yeah, we can.
Only three tests are failing. 
Reduce_L2
https://github.com/opencv/opencv/blob/8eafddf211ec27091a3096eb50964e04d49b36cb/modules/dnn/test/test_onnx_importer.cpp#L412
Unfused_softmax
https://github.com/opencv/opencv/blob/8eafddf211ec27091a3096eb50964e04d49b36cb/modules/dnn/test/test_onnx_importer.cpp#L438
Div
https://github.com/opencv/opencv/blob/8eafddf211ec27091a3096eb50964e04d49b36cb/modules/dnn/test/test_onnx_importer.cpp#L366
The problem will occur if any node of subgraph that we are trying to match has input from graph's input layer. 
**Solution** 
If we add some layer between our test model node and graph's input. then it will never try to match with graph's input layer.

For example, this will not work
```
class Div(nn.Module):
    def __init__(self):
        super(Div, self).__init__()
    def forward(self, a, b):
        return a/b
a = Variable(torch.randn(1, 3, 2, 4))
b = Variable(torch.randn(1, 3, 2, 4))
model = Div()
save_data_and_model_multy_inputs("division", model, a, b)
```
Screenshot from 2020-02-23 07-19-50

But this model will work
```
a = Variable(torch.randn(1, 3, 2, 2))
b = Variable(torch.randn(1, 3, 2, 2))
class DivExp(nn.Module):
    def __init__(self):
        super(DivExp, self).__init__()
    def forward(self, x1, x2):
        flatten1 = x1.flatten().view(1,-1)
        flatten2 = x2.flatten().view(1,-1)
        return flatten1/flatten2
divexp = DivExp()
save_data_and_model_multy_inputs("divexp", divexp, a , b)
```
Screenshot from 2020-02-23 07-20-28

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: We need to remove this node. because this is empty constant node. `replace` function will not remove this.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: okay

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: Okay, I will change that.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: Sequence of our subgraph is

`Shape Constant Gather Cast Constant Mul Cast Floor Shape Constant Gather Cast Constant Mul Cast Floor Unsqueeze Unsqueeze`

While matchedNodesIds contain vector ids of nodes 

`Shape Gather Cast Mul Cast Floor Shape Gather Cast Mul Cast Floor Unsqueeze Unsqueeze`
i.e. without constant. 
so this loop is searching  for consant nodes that are missing from matchedNodeIds because we need to extract data from those constant nodes.

modules/dnn/src/graph_simplifier.cpp
ashishkrshrivastava: @dkurt should I make some changes in these tests [`Reduce_L2`, `Div `and `Softamx_unfused`] ?

modules/dnn/src/graph_simplifier.cpp
```diff
@@ -69,8 +69,14 @@ int Subgraph::getInputNodeId(const Ptr<ImportGraphWrapper>& net,
...
-        if (net->getNodeName(i) == name)
-            return i;
+        Ptr<ImportNodeWrapper> node = net->getNode(i);
+        const int numOutputs = node->getNumOutputs();
+        for (int j = 0; j < numOutputs; j++)
+        {
+            std::string outName = (node->getOutputName(j) != "graph_input") ? node->getOutputName(j) : net->getNodeName(i);
+            if (outName  == name)
+                return i;
+        }
```
dkurt: Please take a look at commit https://github.com/dkurt/opencv/commit/c39ea60140c247a90d8ccf43c9c9d4b25e2a0d6a

Now I see the problem. Let's check outputs names from network rather than nodes. Sorry for misleading!

ashishkrshrivastava: ohk, got it. I have added you suggested changes. Thank you so much.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -145,13 +154,258 @@ class SoftMaxSubgraph : public Subgraph
...
+        int mul2 = addNodeToMatch("Mul", castG2, scaleW);
+        int castM2 = addNodeToMatch("Cast", mul2);
+        int floor2 = addNodeToMatch("Floor", castM2);
+
+        int unsqueeze1 = addNodeToMatch("Unsqueeze", floor1);
+        int unsqueeze2 = addNodeToMatch("Unsqueeze", floor2);
+        addNodeToMatch("Concat", unsqueeze1, unsqueeze2);
+
+        setFusedNode("Scales", input);
+    }
```
dkurt: Please take a look at the changes from https://github.com/dkurt/opencv/commit/15132e85b3f19782b1f40019877196a1331c34ca

I just refactored this part to be more object oriented. So to  fuse each graph we will have not a sequence of fusions but a single pattern.

There are also some improvements which let to manage empty Constant nodes or other stuff.

ashishkrshrivastava: @dkurt I have gone through all of your changes, this is much better way of doing fusion. My changes were really redundant. I have added your suggested changes. I will keep in mind to minimize and optimize code and make it object oriented as much as possible from now onwards. Thanks for your help.

dkurt: @ashishkrshrivastava, Anyway it's really interesting idea to combine multiple patters, really. I believe it can be useful somewhere but in this sample it's better to keep them separately for now.

ashishkrshrivastava: Sure.

BadMachine:
@ashishkrshrivastava excuse me, can u explain how can i use your fix in OpenCV 4.2.0 ?

dkurt:
@BadMachine, you need to build OpenCV from source. The fix has been merged after OpenCV 4.2.0

BadMachine:
@dkurt Thanks for your answer!
But source code now build with errors...
image
PS: I know it should be in issue section, just for u know

dkurt:
@BadMachine, Please do not use closed PRs for bug reports - open a new issue including build flags and error logs (but not a screenshot).

PR #16577, opened 2020-02-13
G-API: Fix. own::Mat successfully creates a {-1, -1} mat.

Volskig:
### own::Mat successfully created a mat with negative dimensions. 
Method `create(...)` throw exception in this case now.

cv::Mat checks the dimensions in `setSize(...)`. If the dimension is less then 0, it throws exception. own::Mat checks `size.height` and `size.width` in `create(...)` and throws exception if dimension is less then 0.

Added test for this case.

### DIFF
--- a/modules/gapi/include/opencv2/gapi/own/mat.hpp
+++ b/modules/gapi/include/opencv2/gapi/own/mat.hpp
@@ -230,6 +230,7 @@ namespace cv { namespace gapi { namespace own {
         */
         void create(Size _size, int _type)
         {
+            GAPI_Assert(_size.height >= 0 && _size.width >= 0);
             if (_size != Size{cols, rows} )
             {
                 Mat tmp{_size.height, _size.width, _type, nullptr};
--- a/modules/gapi/test/own/mat_tests.cpp
+++ b/modules/gapi/test/own/mat_tests.cpp
@@ -588,4 +588,23 @@ TEST(OwnMat, ROIView)
         << to_ocv(roi_view) << std::endl
         << expected_cv_mat  << std::endl;
 }
+
+TEST(OwnMat, CreateWithNegativeDims)
+{
+    Mat own_mat;
+    ASSERT_ANY_THROW(own_mat.create(cv::Size{-1, -1}, CV_8U));
+}
+
+TEST(OwnMat, CreateWithNegativeWidth)
+{
+    Mat own_mat;
+    ASSERT_ANY_THROW(own_mat.create(cv::Size{-1, 1}, CV_8U));
+}
+
+TEST(OwnMat, CreateWithNegativeHeight)
+{
+    Mat own_mat;
+    ASSERT_ANY_THROW(own_mat.create(cv::Size{1, -1}, CV_8U));
+}
+
 } // namespace opencv_test


modules/gapi/include/opencv2/gapi/own/mat.hpp
```diff
@@ -205,6 +205,7 @@ namespace cv { namespace gapi { namespace own {
         */
         void create(Size _size, int _type)
         {
+            GAPI_Assert(_size.height >= 0 && _size.width >= 0);
```
dkurt: Why 0 values are included?

Volskig: Because `CV_Assert( s >= 0 )` is used in cv::Mat in `create(...)` in `setSize(...)`.

cv::Mat successfully creates Size{0, 0} mat.

It is a similar check for own::Mat.

asmorkalov:
LGTM! @dmatveev Please merge the patch.

modules/gapi/test/own/mat_tests.cpp
```diff
@@ -384,4 +384,21 @@ TEST(OwnMat, ROIView)
...
+    EXPECT_THROW(ocv_mat.create(cv::Size(-1,-1), CV_8U), std::exception);
+
+    Mat own_mat_t1, own_mat_t2, own_mat_t3;
+    EXPECT_THROW(own_mat_t1.create(cv::Size{-1, -1}, CV_8U), std::exception);
+    EXPECT_THROW(own_mat_t2.create(cv::Size{ 1, -1}, CV_8U), std::exception);
+    EXPECT_THROW(own_mat_t3.create(cv::Size{-1,  1}, CV_8U), std::exception);
+
+    // if Size{cols, rows} of create method is equal to Size{cols, rows} of Mat case
+    Mat own_mat_t4(-1, -1, CV_8U, nullptr);
+    EXPECT_THROW(own_mat_t4.create(-1, -1, CV_8U), std::exception);
```
dmatveev: What if just one of dimensions is negative? Can I create a Mat with zero height?

modules/gapi/include/opencv2/gapi/own/mat.hpp
dmatveev: so `create(cv::Size(0,0))` is valid?

modules/gapi/test/own/mat_tests.cpp
Volskig: You can't create own::Mat if just one of dimensions is negative.

And you can create own::Mat with zero height.

modules/gapi/include/opencv2/gapi/own/mat.hpp
Volskig: Yes. For now.

modules/gapi/test/own/mat_tests.cpp
dmatveev: > You can't create own::Mat if just one of dimensions is negative.

My point was that this needs to be demonstrated by tests

> And you can create own::Mat with zero height.

What's the reason in a Mat with zero height? Can we also create a cv::Mat with zero height?

Volskig: We can create cv::Mat with zero dimensions. We can use constructor or create method for this.
But cv::Mat with zero dimensions isn't use anywhere. So create own::Mat with zero dimensions doesn't make sense.

Should we throw an exception in own::Mat's constructor for this case ( rows or cols <= 0 )? It destroys the own::Mat with zero-height creation bug.

Volskig:
@dmatveev

modules/gapi/test/own/mat_tests.cpp
dmatveev: ok got it

modules/gapi/test/own/mat_tests.cpp
```diff
@@ -384,4 +384,37 @@ TEST(OwnMat, ROIView)
     << to_ocv(roi_view) << std::endl
     << expected_cv_mat  << std::endl;
 }
+
+TEST(OwnMat, CreateOwnMat)
```
dmatveev: Just figured out that the title of this test doesn't reflect the meaning of this test.

Also you can split this test into two or three, highlighting what aspects do you verify with this test.

Volskig: Fixed. Divided test into three. Changed names for tests.

Volskig:
@dmatveev

Volskig:
@TolyaTalamanov

asmorkalov:
@Volskig @dmatveev Do you have any progress here?

modules/gapi/test/own/mat_tests.cpp
```diff
@@ -384,4 +384,32 @@ TEST(OwnMat, ROIView)
     << to_ocv(roi_view) << std::endl
     << expected_cv_mat  << std::endl;
 }
+
+TEST(OwnMat, CreateNegEmptyMat)
+{
+    // We use create method for empty own::Mat
+    Mat own_mat_t1, own_mat_t2, own_mat_t3;
+    EXPECT_THROW(own_mat_t1.create(cv::Size{-1, -1}, CV_8U), std::exception);
```
TolyaTalamanov: std::exception is base class it coverage all exceptions, better to use `EXPECT_ANY_THROW` if you don't want to clarify exception type

modules/gapi/test/own/mat_tests.cpp
```diff
@@ -384,4 +384,32 @@ TEST(OwnMat, ROIView)
...
+    EXPECT_THROW(own_mat_t2.create(cv::Size{ 1, -1}, CV_8U), std::exception);
+    EXPECT_THROW(own_mat_t3.create(cv::Size{-1,  1}, CV_8U), std::exception);
+}
+
+TEST(OwnMat, CreateNegExistMat)
```
TolyaTalamanov: Split this into small tests for specific cases instead of one large test with a bunch of checks
This way you will see that comments will be unnecessary

dmatveev: +1

modules/gapi/test/own/mat_tests.cpp
Volskig: Fixed.

modules/gapi/test/own/mat_tests.cpp
Volskig: @TolyaTalamanov I made this tests easier. Is split necessary?

modules/gapi/test/own/mat_tests.cpp
```diff
@@ -384,4 +384,23 @@ TEST(OwnMat, ROIView)
...
+{
+    Mat own_mat_t1, own_mat_t2, own_mat_t3;
+    EXPECT_ANY_THROW(own_mat_t1.create(cv::Size{-1, -1}, CV_8U));
+    EXPECT_ANY_THROW(own_mat_t2.create(cv::Size{ 1, -1}, CV_8U));
+    EXPECT_ANY_THROW(own_mat_t3.create(cv::Size{-1,  1}, CV_8U));
+}
+
+TEST(OwnMat, ExistCreateWithNegDims)
+{
+    Mat own_mat_t1(-1, -1, CV_8U, nullptr);
```
TolyaTalamanov: Is it really necessary to pass `nullptr` here ?

Volskig: own::Mat's constructor is `Mat(int, int, int, void* _data, size_t)`. I should put something in `_data`.

TolyaTalamanov: Ok

TolyaTalamanov:
LGTM 👍 
Thanks!

Issue #14521, opened 2019-05-09
integer overflow in convexHull


andreatta-deltamax:
##### System information (version)

- OpenCV => 3.4.3
- Operating System / Platform => Windows 64 Bit
- Compiler => Visual Studio 2015

##### Detailed description

I was testing convexHull alternative algorithms, and I had different results unsing convexHull with Point and Point2f with the same data.
I looks like there is a potential overflow in 
`template<typename _Tp>
static int Sklansky_( Point_<_Tp>** array, int start, int end, int* stack, int nsign, int sign2 )`

convhull.cpp:82
` _Tp convexity = ay*bx - ax*by; // if >0 then convex angle`

##### Steps to reproduce

```
  std::vector<cv::Point> points;
  std::vector<cv::Point2f> pointsf;

  points.push_back(cv::Point(14763, 2890));
  points.push_back(cv::Point(14388, 72088));
  points.push_back(cv::Point(62810, 72274));
  points.push_back(cv::Point(63166, 3945));
  points.push_back(cv::Point(56782, 3945));
  points.push_back(cv::Point(56763, 3077));
  points.push_back(cv::Point(34666, 2965));
  points.push_back(cv::Point(34547, 2953));
  points.push_back(cv::Point(34508, 2866));
  points.push_back(cv::Point(34429, 2965));
  
  for (auto it = points.begin(); it != points.end(); ++it) pointsf.push_back(*it);

  std::vector<int> hull;
  std::vector<int> hullf;

  cv::convexHull(points, hull, false, false);
  cv::convexHull(pointsf, hullf, false, false);

  if (hull != hullf) throw std::runtime_error("error!");

```

vpisarev:
@andreatta-deltamax, thank you! fixed

Issue #4539, opened 2015-07-27
Indices from convexHull not monotonic


opencv-pushbot:
Transferred from http://code.opencv.org/issues/2922

```
|| Kirt Lillywhite on 2013-03-26 20:47
|| Priority: Normal
|| Affected: 2.4.4
|| Category: imgproc, video
|| Tracker: Bug
|| Difficulty: None
|| PR: 
|| Platform: None / None
```
## Indices from convexHull not monotonic

```
From a segmented image I find the contour and then try to find the convexity defects from that contour. The largest convexity defect is wrong. Looking through the opencv code I think the reason is the indices from convexHull are in decreasing order except in one spot. This causes the convexityDefects function to mess up. The attached code shows the problem. I'm using 2.4.4 compiled from source on Archlinux x86-64. During compilation I turned off CUDA and enable QT, everything else is set to the default options.
```
##### Kirt Lillywhite on 2013-03-26 20:56

```
-   File mask.png added
```

kfrance:
I submitted this bug originally 4 years ago. I recently updated opencv and code with a fix to deal with this bug no longer worked. It appears that the bug has been solved in 3.2.0.

hannarud:
Hi! I think the issue still exists (or please clarify if it's not) - I use OpenCV 3.4.1, Python interface.

I guess the main matter is that `cv2.convexHull` returns the hull points ordered differently than in original contour. After that when distances are computed for defects, some of them pass through too many points (instead of checking in-between hull points only). Not sure in this though. Can anyone help?

My reproducible code sample:

```python
import cv2
import numpy as np
import matplotlib.pyplot as plt

contour = np.array([[[26,  9]],
       [[25, 10]],
       [[24, 10]],
       [[23, 10]],
       [[22, 10]],
       [[21, 10]],
       [[20, 11]],
       [[19, 11]],
       [[18, 11]],
       [[17, 12]],
       [[17, 13]],
       [[18, 14]],
       [[18, 15]],
       [[18, 16]],
       [[18, 17]],
       [[19, 18]],
       [[19, 19]],
       [[20, 20]],
       [[21, 21]],
       [[21, 22]],
       [[22, 23]],
       [[22, 24]],
       [[23, 25]],
       [[23, 26]],
       [[24, 27]],
       [[25, 28]],
       [[26, 29]],
       [[27, 30]],
       [[27, 31]],
       [[28, 32]],
       [[29, 32]],
       [[30, 33]],
       [[31, 34]],
[...]

PR #16594, opened 2020-02-16
fixed the ordering of contour convex hull points

vpisarev:
Fixes issue #14521.
Partially fixes the issue #4539.

1. if possible, the convex hull indices are "rotated" to form a monotonous sequence, decreasing or increasing.
2. added the check to convexityDefects to check that the indices are monotonous; if not, an exception is thrown
3. added the test from #4539 discussion with a workaround (to eliminate the contour self-intersection). With this workaround everything works well.

### DIFF
--- a/modules/imgproc/src/convhull.cpp
+++ b/modules/imgproc/src/convhull.cpp
@@ -122,7 +122,13 @@ template<typename _Tp>
 struct CHullCmpPoints
 {
     bool operator()(const Point_<_Tp>* p1, const Point_<_Tp>* p2) const
-    { return p1->x < p2->x || (p1->x == p2->x && p1->y < p2->y); }
+    {
+        if( p1->x != p2->x )
+            return p1->x < p2->x;
+        if( p1->y != p2->y )
+            return p1->y < p2->y;
+        return p1 < p2;
+    }
 };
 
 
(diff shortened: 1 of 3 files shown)


vpisarev:
@alalek, can you please, rerun tests on mac. something strange happened there, just timeout

modules/imgproc/test/test_convhull.cpp
```diff
@@ -2154,5 +2154,142 @@ TEST(Imgproc_FitLine, regression_4903)
...
+    };
+    int npoints = (int)(sizeof(contour)/sizeof(contour[0][0])/2);
+    int scale = 20;
+    Mat contour_(1, npoints, CV_32SC2, contour);
+    contour_ *= (double)scale;
+
+    Mat canvas_gray(Size(60*scale, 45*scale), CV_8U, Scalar::all(0));
+    fillConvexPoly(canvas_gray, contour_.ptr<Point>(), npoints, Scalar(255, 255, 255));
+
+#if 1 // try to eliminate the self-intersection in one way or another
```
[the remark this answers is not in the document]
vpisarev: ok, I added the code to test the original contour

modules/imgproc/src/convhull.cpp
```diff
@@ -299,12 +344,22 @@ void convexityDefects( InputArray _points, InputArray _hull, OutputArray _defect
...
     {
         int hnext = hptr[rev_orientation ? hpoints - i - 1 : i];
         CV_Assert( 0 <= hnext && hnext < npoints );
 
         Point pt0 = ptr[hcurr], pt1 = ptr[hnext];
+        if( increasing_idx < 0 )
+            increasing_idx = !(hcurr < hnext);
+        else if( increasing_idx != (hcurr < hnext))
+        {
+            CV_Error(Error::StsBadArg,
```
bommo1, line 359: This change sometimes breaks the ERFilter.
https://github.com/opencv/opencv_contrib/blob/master/modules/text/src/erfilter.cpp#L2837

alalek: Please provide complete minimal reproducer and fill an issue in opencv_contrib.

PR #16597, opened 2020-02-17
Fixes matrix expression bug which happens when Scalar in addition pas…

### LINKED ISSUE #16538: Problem for Simple Alpha Blending

- OpenCV => 4.2.0
- Compiler => Visual Studio 2019

##### Detailed description

If running following the code for alpha blending, it output following results.  

```cpp
#include <opencv2/opencv.hpp>

int main( int argc, char* argv[] )
{
    // Read Image
    cv::Mat src = cv::imread( "lena.jpg" );

    // Alpha Blend with Blue, Green, and Red
    double alpha = 0.6;
    double beta  = 1.0 - alpha;
    cv::Mat blend_b = alpha * src + beta * cv::Scalar( 255, 0, 0 ); // Blue (issue case!)
    cv::Mat blend_g = alpha * src + beta * cv::Scalar( 0, 255, 0 ); // Green
    cv::Mat blend_r = alpha * src + beta * cv::Scalar( 0, 0, 255 ); // Red

    // Show Image
    cv::imshow( "src"  , src     );
    cv::imshow( "blue" , blend_b );
    cv::imshow( "green", blend_g );
    cv::imshow( "red"  , blend_r );
    cv::waitKey( 0 );

    return 0;
}
```

The output of alpha blending with green and red is correct result.  
These turned greenish color and reddish color.  
But, the output of alpha blending with blue has become whitish color.  
It is not expected output. It should turn to bluish color.  

output1

If changes blend color to close blue <code>cv::Scalar(255, 1, 0)</code> (just add one to green), it turn correctly to bluish color.

```cpp
cv::Mat blend_b = alpha * src + beta * cv::Scalar( 255, 1, 0 ); // Close Blue
```
output2

##### Related issues

This seems to be related to this issue #14738.  
I think this is due to the difference between calculation results of cv::Mat and cv::MatExpr.  
Therefore, I know that this issue can be avoided by changing to following code.  

```patch
- cv::Mat blend_b = alpha * src + beta * cv::Scalar( 255, 0, 0 ); // Blue
+ cv::Mat blend_b = cv::Mat( alpha * src ) + beta * cv::Scalar( 255, 0, 0 ); // Blue
```

But, I think this is a bug of corner case. It should be fixed.
What do you think? Thanks,

saskatchewancatch:
…ses isReal test
Fixes #16538 
Unit test is under work. Coming shortly.

### DIFF
--- a/modules/core/src/matrix_expressions.cpp
+++ b/modules/core/src/matrix_expressions.cpp
@@ -1312,9 +1312,12 @@ void MatOp_AddEx::assign(const MatExpr& e, Mat& m, int _type) const
                 cv::add(dst, e.s, dst);
         }
         else
-            cv::addWeighted(e.a, e.alpha, e.b, e.beta, e.s[0], dst);
+        {
+            cv::addWeighted(e.a, e.alpha, e.b, e.beta, 0, dst);
+            cv::add(dst, e.s, dst);
+        }
     }
-    else if( e.s.isReal() && (dst.data != m.data || fabs(e.alpha) != 1))
+    else if( e.a.channels() == 1 && e.s.isReal() && (dst.data != m.data || fabs(e.alpha) != 1))
     {
         e.a.convertTo(m, _type, e.alpha, e.s[0]);
         return;
--- a/modules/core/src/matrix_expressions.cpp
+++ b/modules/core/src/matrix_expressions.cpp
@@ -10,6 +10,7 @@
 // */
 
 #include "precomp.hpp"
+#include <iostream>
 
 namespace cv
 {
@@ -1329,7 +1330,10 @@ void MatOp_AddEx::assign(const MatExpr& e, Mat& m, int _type) const
     else
     {
         e.a.convertTo(dst, e.a.type(), e.alpha);
-        cv::add(dst, e.s, dst);
+        if (e.a.channels() == 1)
+            cv::add(dst, e.s, dst);
+        else
+            cv::add(dst, e.s[0], dst);
     }
 
     if( dst.data != m.data )
--- a/modules/core/src/matrix_expressions.cpp
+++ b/modules/core/src/matrix_expressions.cpp
@@ -10,7 +10,6 @@
 // */
 
 #include "precomp.hpp"
-#include <iostream>
 
 namespace cv
 {
--- a/modules/core/src/matrix_expressions.cpp
+++ b/modules/core/src/matrix_expressions.cpp
@@ -1329,10 +1329,7 @@ void MatOp_AddEx::assign(const MatExpr& e, Mat& m, int _type) const
     else
     {
         e.a.convertTo(dst, e.a.type(), e.alpha);
-        if (e.a.channels() == 1)
-            cv::add(dst, e.s, dst);
-        else
-            cv::add(dst, e.s[0], dst);
+        cv::add(dst, e.s, dst);
     }
 
     if( dst.data != m.data )
--- a/modules/core/test/test_operations.cpp
+++ b/modules/core/test/test_operations.cpp
@@ -1490,22 +1490,43 @@ TEST(Core_MatExpr, empty_check_15760)
 TEST(Core_MatExpr, issue_16538)
 {
     Mat multiChannel_2x2__1_2_3(Size(2, 2), CV_8UC3, Scalar(1, 2, 3));
+    Mat multiChannel_2x2__1_2_3_4(Size(2, 2), CV_8UC4, Scalar(1, 2, 3, 4));
     Mat multiChannel_2x2__6_2_3(Size(2, 2), CV_8UC3, Scalar(6, 2, 3));
-    Mat multiChannel_2x2__7_9_11(Size(2, 2), CV_8UC3, Scalar(7, 9, 11));
+    Mat multiChannel_2x2__7_4_6(Size(2, 2), CV_8UC3, Scalar(7, 4, 6));
     Mat multiChannel_2x2__11_7_3(Size(2, 2), CV_8UC3, Scalar(11, 7, 3));
+    Mat multiChannel_2x2__6_7_8_9(Size(2, 2), CV_8UC4, Scalar(6, 7, 8, 9));
+    Mat multiChannel_2x2__7_9_11_13(Size(2, 2), CV_8UC4, Scalar(7, 9, 11, 13));
 
-
+    Scalar s5 = Scalar(5);
+    Scalar sa5 = Scalar::all(5);
+    Scalar s5_0_0_0= Scalar(5, 0, 0, 0);
 
     Mat res;
-    res = multiChannel_2x2__1_2_3 + 5;
+    res = multiChannel_2x2__1_2_3 + s5;
     //Single multi-channel matrix plus single-channel scalar.
     EXPECT_DOUBLE_EQ(cvtest::norm(multiChannel_2x2__6_2_3, res, NORM_INF), 0.0);
+    res = multiChannel_2x2__1_2_3 + s5_0_0_0;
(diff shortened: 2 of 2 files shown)


saskatchewancatch:
I missed the superres test on my VM, seems it was disabled because of long-ness.

Running on another machine, this part of the change is causing the failure:
```diff
-    else if( e.s.isReal() && (dst.data != m.data || fabs(e.alpha) != 1))
+    else if( e.a.channels() == 1 && e.s.isReal() && (dst.data != m.data || fabs(e.alpha) != 1))
     {
         e.a.convertTo(m, _type, e.alpha, e.s[0]);
         return;
```

I _suspect_ problem may be somewhere [here where SSIM is calculated.](https://github.com/opencv/opencv/blob/3.4/modules/superres/test/test_superres.cpp#L155)

asmorkalov:
@saskatchewancatch Could add test for the bug to prove the solution?

alalek:
Some code for tests:
```
    Mat m0(Size(5, 3), CV_8UC3, Scalar(1,2,3));

    {
        Mat m = m0 + 5;
        cout << m << endl;
    }
    {
        Mat m = 2 * m0 + 5;
        cout << m << endl;
    }
    {
        Mat m = m0 + 5 * Scalar(2, 1, 0);
        cout << m << endl;
    }
#if 0 // can't compile
    {
        Mat m = 2 * m0 + 5 * Scalar(2, 1, 0);
        cout << m << endl;
    }
#endif
```

<details>

<summary>Current 3.4.9/4.2.0 output</summary>

```
[  6,   2,   3,   6,   2,   3,   6,   2,   3,   6,   2,   3,   6,   2,   3;
   6,   2,   3,   6,   2,   3,   6,   2,   3,   6,   2,   3,   6,   2,   3;
   6,   2,   3,   6,   2,   3,   6,   2,   3,   6,   2,   3,   6,   2,   3]
[  7,   9,  11,   7,   9,  11,   7,   9,  11,   7,   9,  11,   7,   9,  11;
   7,   9,  11,   7,   9,  11,   7,   9,  11,   7,   9,  11,   7,   9,  11;
   7,   9,  11,   7,   9,  11,   7,   9,  11,   7,   9,  11,   7,   9,  11]
[ 11,   7,   3,  11,   7,   3,  11,   7,   3,  11,   7,   3,  11,   7,   3;
  11,   7,   3,  11,   7,   3,  11,   7,   3,  11,   7,   3,  11,   7,   3;
  11,   7,   3,  11,   7,   3,  11,   7,   3,  11,   7,   3,  11,   7,   3]
```

</details>

saskatchewancatch:
`core/test/test_operations.cpp`'s `CV_OperationsTest::TestMat()` is a good place to put the test. Thoughts?

I propose integrating alalek's suggestion like such:

```diff
diff --git a/modules/core/test/test_operations.cpp b/modules/core/test/test_operations.cpp
index dd91617537..d2ff870559 100644
--- a/modules/core/test/test_operations.cpp
+++ b/modules/core/test/test_operations.cpp
@@ -292,6 +292,16 @@ bool CV_OperationsTest::TestMat()
         CHECK_DIFF(4.0 * (maskMat1 | maskMat1), maskMat4);
         CHECK_DIFF((maskMat4 | maskMat4)/4.0, maskMat1);

+        //Multi-channel Scalar addition tests
+        Mat multiChannel_2x2__1_2_3(Size(2, 2), CV_8UC3, Scalar(1, 2, 3));
+        Mat multiChannel_2x2__6_2_3(Size(2, 2), CV_8UC3, Scalar(6, 2, 3));
+        Mat multiChannel_2x2__7_9_11(Size(2, 2), CV_8UC3, Scalar(7, 9, 11));
+        Mat multiChannel_2x2__11_7_3(Size(2, 2), CV_8UC3, Scalar(11, 7, 3));
+
+        CHECK_DIFF(multiChannel_2x2__6_2_3, multiChannel_2x2__1_2_3 + 5);
+        CHECK_DIFF(multiChannel_2x2__7_9_11, 2 * multiChannel_2x2__1_2_3 + 5);
+        CHECK_DIFF(multiChannel_2x2__11_7_3, multiChannel_2x2__1_2_3 + 5 * Scalar(2, 1, 0));
+
 #if !MSVC_OLD
         CHECK_DIFF(2.0 * (maskMat1 * 2.0) , maskMat4);
 #endif
```

alalek:
> test_operations.cpp

Right.

It is better to create separate test case (it is hard to maintain or debug huge tests), like:
```
TEST(Core_MatExpr, issue_16538)
{
    ...
}
```

saskatchewancatch:
Thank you for test suggestions @alalek 

Between the tests you provided and the ones that already exist in test_operations.cpp, the optimization branches that were touched in this PR are tested.

modules/core/src/matrix_expressions.cpp
```diff
@@ -1326,7 +1329,10 @@ void MatOp_AddEx::assign(const MatExpr& e, Mat& m, int _type) const
     else
     {
         e.a.convertTo(dst, e.a.type(), e.alpha);
-        cv::add(dst, e.s, dst);
+        if (e.a.channels() == 1)
+            cv::add(dst, e.s, dst);
+        else
+            cv::add(dst, e.s[0], dst);
```
alalek: Both code branches are equal to:
```
cv::add(dst, e.s[0], dst);
```

so proposed condition doesn't make sense.

The real problem is how input "double" is stored through `Scalar`:
- `5` is `Scalar(5, 0, 0, 0)`
- or `5` is `Scalar:all(5)`
- also tests should properly handle `+ Scalar(5, 0, 0, 0)` and `Scalar:all(5)`

`.isReal()` check is not correct if we try to use `Scalar::all(5)` approach.
Probably we should raise exceptions on operations like `multichannel_mat + just scalar`.

saskatchewancatch: @alalek Thanks for review, to address your feedback:

- Yes, that if/else is redundant. I will just replace with `cv::add(dst, e.s, dst);`

- Raising exception in case of `mulitchannel_mat + just_scalar` is fine imo but I think it's better handled in a separate PR, where `isReal()` issue can be fully sorted out. That may take longer as the solution cannot be implemented inside matrix_expressions.cpp.

- As for this PR, I will change the tests above so that the `just scalar` is changed to the Scalar form: ex) `res = multiChannel_2x2__1_2_3 + Scalar(5, 0, 0);`, and then add the tests Scalar::all test you suggested.

This PR at least will fix OP's alpha blending issue and strengthen against regressions with the added tests. Is this fine?

alalek: Patch is fine.
But it explodes another problem: `multiChannel + 5.0` result is not well defined.
Need to take a look on this in details.

saskatchewancatch:
The superres test is the problem. In the test, a multichannel matrix is being added to a double constant. As per issue [16739](https://github.com/opencv/opencv/issues/16739), and above discussion, this is a grey area. I propose this adjustment for `test_superres.cpp` to make the addition with the pure scalar explicit.


```diff
diff --git a/modules/superres/test/test_superres.cpp b/modules/superres/test/test_superres.cpp
index b0d4122f6e..b98258484f 100644
--- a/modules/superres/test/test_superres.cpp
+++ b/modules/superres/test/test_superres.cpp
@@ -154,8 +154,8 @@ void DegradeFrameSource::reset()

 double MSSIM(cv::InputArray _i1, cv::InputArray _i2)
 {
-    const double C1 = 6.5025;
-    const double C2 = 58.5225;
+    const Scalar C1(6.5025, 6.5025, 6.5025);
+    const Scalar C2(58.5225, 58.5225, 58.5225);

     const int depth = CV_32F;
```

This makes all tests pass. 
Until we figure out how to deal with issue 16739, this seems sane to me.

alalek:
> The superres test is the problem

Right.
Unfortunately, similar **users** code may exist too, but there is no way to fix that (and probably properly want about that).

asenyaev:

PR #16604, opened 2020-02-17
G-API: Zero-height mat is cause of crash

Volskig:
### Zero-height mat is cause of crash
     cv::GMat in;
     cv::GMat a, b, c, d;
     std::tie(a, b, c, d) = cv::gapi::split4(in);
     cv::GMat out = cv::gapi::merge3(a, b, c);   

     // Zero-height in_mat
     cv::Mat in_mat(cv::Size(8, 0), CV_8UC4);
     cv::Mat out_mat(cv::Size(8, 8), CV_8UC3);

     cv::GComputation comp(cv::GIn(in), cv::GOut(out));
     comp.apply(cv::gin(in_mat), cv::gout(out_mat),
         cv::compile_args(cv::gapi::core::fluid::kernels()));

It crashed in `physIdx(int logIdx)` from gfluidbuffer_priv.h ( Fluid backend ), because we use `(logIdx - m_roi.y) % m_data.rows`, when m_data.rows == 0 ( Segmentation fault ).

Added GAPI_Assert for this case.

### DIFF
--- a/modules/gapi/include/opencv2/gapi/gproto.hpp
+++ b/modules/gapi/include/opencv2/gapi/gproto.hpp
@@ -126,6 +126,9 @@ bool GAPI_EXPORTS can_describe(const GMetaArgs& metas, const GRunArgs& args);
 // coincides with output arguments passed to computation in cpu and ocl backends
 bool GAPI_EXPORTS can_describe(const GMetaArg&  meta,  const GRunArgP& argp);
 
+// Checks dimensions of input Mat for inequality to zero.
+bool GAPI_EXPORTS nonzero_dims_in_inmat(const GMetaArgs& metas);
+
 } // namespace cv
 
 #endif // OPENCV_GAPI_GPROTO_HPP
--- a/modules/gapi/include/opencv2/gapi/gproto.hpp
+++ b/modules/gapi/include/opencv2/gapi/gproto.hpp
@@ -127,8 +127,8 @@ bool GAPI_EXPORTS can_describe(const GMetaArgs& metas, const GRunArgs& args);
 bool GAPI_EXPORTS can_describe(const GMetaArg&  meta,  const GRunArgP& argp);
 
 // Validates input arguments
-bool GAPI_EXPORTS validate_input_arg(const GRunArg& arg);
-bool GAPI_EXPORTS validate_input_arg(const GRunArgs& args);
+void GAPI_EXPORTS validate_input_arg(const GRunArg& arg);
+void GAPI_EXPORTS validate_input_args(const GRunArgs& args);
 
 } // namespace cv
 
--- a/modules/gapi/include/opencv2/gapi/gproto.hpp
+++ b/modules/gapi/include/opencv2/gapi/gproto.hpp
@@ -126,8 +126,9 @@ bool GAPI_EXPORTS can_describe(const GMetaArgs& metas, const GRunArgs& args);
 // coincides with output arguments passed to computation in cpu and ocl backends
 bool GAPI_EXPORTS can_describe(const GMetaArg&  meta,  const GRunArgP& argp);
 
-// Checks dimensions of input Mat for inequality to zero.
-bool GAPI_EXPORTS nonzero_dims_in_inmat(const GMetaArgs& metas);
+// Validates input arguments
+bool GAPI_EXPORTS validate_input_arg(const GRunArg& arg);
+bool GAPI_EXPORTS validate_input_arg(const GRunArgs& args);
 
 } // namespace cv
 
--- a/modules/gapi/src/api/gproto.cpp
+++ b/modules/gapi/src/api/gproto.cpp
@@ -197,18 +197,43 @@ bool cv::can_describe(const GMetaArgs &metas, const GRunArgs &args)
                      });
 }
 
-bool cv::nonzero_dims_in_inmat(const GMetaArgs& metas)
+bool cv::validate_input_arg(const GRunArg& arg)
 {
-    bool state = true;
-    for (const auto& current_meta : metas)
+    // FIXME: It checks only Mat's argument
+    switch (arg.index())
     {
-        if (current_meta.index() == cv::GMetaArg::index_of<cv::GMatDesc>())
-        {
-            const auto& desc_mat = cv::util::get<cv::GMatDesc>(current_meta);
-            state &= (desc_mat.size.height != 0 && desc_mat.size.width != 0);
-        }
+#if !defined(GAPI_STANDALONE)
+    case GRunArg::index_of<cv::Mat>():
+    {
+        const auto desc = descr_of(util::get<cv::Mat>(arg));
+        GAPI_Assert(desc.size.height != 0 && desc.size.width != 0 && "incorrect dimensions of cv::Mat!"); return true;
     }
-    return state;
+    case GRunArg::index_of<cv::UMat>():
+    {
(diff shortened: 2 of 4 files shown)


asmorkalov:
@Volskig @dmatveev any progress on the patch?

dmatveev:
@asmorkalov yes

Volskig:
@dmatveev Does GMatDesc get from `cv::GMetaArg` correctly?

modules/gapi/src/api/gproto.cpp
```diff
@@ -197,6 +197,20 @@ bool cv::can_describe(const GMetaArgs &metas, const GRunArgs &args)
...
+    bool state = true;
+    for (const auto& current_meta : metas)
+    {
+        if (current_meta.index() == cv::GMetaArg::index_of<cv::GMatDesc>())
+        {
+            const auto& desc_mat = cv::util::get<cv::GMatDesc>(current_meta);
+            state &= (desc_mat.size.height != 0 && desc_mat.size.width != 0);
+        }
+    }
+    return state;
```
dmatveev: There's no need to check every argument, you can return `false` early once you find a match.

ALSO: `std::any_of`

modules/gapi/src/api/gproto.cpp
```diff
@@ -197,6 +197,20 @@ bool cv::can_describe(const GMetaArgs &metas, const GRunArgs &args)
                      });
 }
 
+bool cv::nonzero_dims_in_inmat(const GMetaArgs& metas)
+{
+    bool state = true;
+    for (const auto& current_meta : metas)
+    {
+        if (current_meta.index() == cv::GMetaArg::index_of<cv::GMatDesc>())
```
dmatveev: what you actually need here is `holds_alternative`.

modules/gapi/src/api/gproto.cpp
dmatveev: Also I'd refactor the code in other form. Lets have a function like `validate_input_arg` which itself takes a `GRunArg` and checks if the parameter is valid based on its type. If anything wrong for, say, Mat, you just throw an exception.

dmatveev: And yes, please start with `GRunArgs` not `GMetaArgs`.

Volskig: Fixed. Another form of validation. Is it?

modules/gapi/include/opencv2/gapi/gproto.hpp
```diff
@@ -126,6 +126,10 @@ bool GAPI_EXPORTS can_describe(const GMetaArgs& metas, const GRunArgs& args);
 // coincides with output arguments passed to computation in cpu and ocl backends
 bool GAPI_EXPORTS can_describe(const GMetaArg&  meta,  const GRunArgP& argp);
 
+// Validates input arguments
+bool GAPI_EXPORTS validate_input_arg(const GRunArg& arg);
+bool GAPI_EXPORTS validate_input_arg(const GRunArgs& args);
```
dmatveev: What's the reason in `bool` return type here if the function throws?

dmatveev: for a `GRunArgs` version I'd name it `.._args`

modules/gapi/src/api/gproto.cpp
```diff
@@ -197,6 +197,45 @@ bool cv::can_describe(const GMetaArgs &metas, const GRunArgs &args)
...
+bool cv::validate_input_arg(const GRunArg& arg)
+{
+    // FIXME: It checks only Mat's argument
+    switch (arg.index())
+    {
+#if !defined(GAPI_STANDALONE)
+    case GRunArg::index_of<cv::Mat>():
+    {
+        const auto desc = descr_of(util::get<cv::Mat>(arg));
+        GAPI_Assert(desc.size.height != 0 && desc.size.width != 0 && "incorrect dimensions of cv::Mat!"); return true;
```
dmatveev: you don't need `return` here. And please don't forget about `break` once you refactor

modules/gapi/src/api/gproto.cpp
```diff
@@ -197,6 +197,45 @@ bool cv::can_describe(const GMetaArgs &metas, const GRunArgs &args)
...
+    case GRunArg::index_of<cv::gapi::own::Mat>():
+    {
+        const auto desc = descr_of(util::get<cv::gapi::own::Mat>(arg));
+        GAPI_Assert(desc.size.height != 0 && desc.size.width != 0 && "incorrect dimensions of own::Mat!"); return true;
+    }
+    case GRunArg::index_of<cv::gapi::own::Scalar>(): return true;
+    case GRunArg::index_of<cv::detail::VectorRef>(): return true;
+    case GRunArg::index_of<cv::detail::OpaqueRef>(): return true;
+    case GRunArg::index_of<cv::gapi::wip::IStreamSource::Ptr>(): return true;
+    default: util::throw_error(std::logic_error("Unsupported GRunArg type"));
```
dmatveev: Actually, if a type doesn't require any validation, you can just do
```c++
default:
    // No extra handling
    break;
```

modules/gapi/src/api/gproto.cpp
```diff
@@ -197,6 +197,45 @@ bool cv::can_describe(const GMetaArgs &metas, const GRunArgs &args)
...
+    }
+}
+
+bool cv::validate_input_arg(const GRunArgs& args)
+{
+    return std::any_of(args.begin(), args.end(),
+                      [](const GRunArg& arg) {
+                          return validate_input_arg(arg);
+                      });
```
dmatveev: You don't need lambda here. Just make `validate_input_arg` the 3rd argument to `any_of`.

dmatveev: Since we're going to `throw`, I'd just run it via `for_each`.

modules/gapi/src/compiler/gcompiled.cpp
```diff
@@ -56,6 +56,11 @@ void cv::GCompiled::Priv::checkArgs(const cv::gimpl::GRuntimeArgs &args) const
                                            "for different metadata!"));
         // FIXME: Add details on what is actually wrong
     }
+    if (!validate_input_arg(args.inObjs))
+    {
+        util::throw_error(std::logic_error("This object was compiled "
+                                           "for incorrect input arguments!"));
```
dmatveev: This text is wrong. The object has been compiled already, here you enter the execution.

modules/gapi/src/api/gproto.cpp
Volskig: std::for_each() or range based for(auto el : cont)?

modules/gapi/include/opencv2/gapi/gproto.hpp
Volskig: Fixed

Volskig: Fixed

modules/gapi/src/api/gproto.cpp
Volskig: Fixed

modules/gapi/src/api/gproto.cpp
Volskig: Fixed

modules/gapi/src/compiler/gcompiled.cpp
Volskig: Removed

PR #16608, opened 2020-02-17
fixed several problems when running tests on Mac

vpisarev:
* failures in ocl_pyrUp
* failures in ocl_flip when processing CV_16UC3 or CV_16SC3 data
* some basic UMat tests; `Mat => UMat => Mat` chain did not work properly on Discrete video cards.
* One of histogram badarg tests tried to access the 2nd element of 1-element input array. Most of the time it worked fine, because the function reported the expected error because of some other bad arg, but sometimes it crashed on histogram allocation when the size was valid but huge.

```
force_builders=Custom Mac,Custom
test_opencl:Custom Mac=ON

buildworker:Custom=linux-5
test_opencl:Custom=ON
```

### DIFF
--- a/modules/core/src/ocl.cpp
+++ b/modules/core/src/ocl.cpp
@@ -4706,8 +4706,6 @@ public:
             getBestFlags(ctx, accessFlags, usageFlags, createFlags, flags0);
 
             bool copyOnMap = (flags0 & UMatData::COPY_ON_MAP) != 0;
-            if( copyOnMap )
-                accessFlags &= ~ACCESS_FAST;
 
             cl_context ctx_handle = (cl_context)ctx.ptr();
             int allocatorFlags = 0;
@@ -4768,6 +4766,9 @@ public:
             else
 #endif
             {
+                if( copyOnMap )
+                    accessFlags &= ~ACCESS_FAST;
+
                 tempUMatFlags = UMatData::TEMP_UMAT;
                 if (!copyOnMap
                     && CV_OPENCL_ENABLE_MEM_USE_HOST_PTR
--- a/modules/core/src/ocl.cpp
+++ b/modules/core/src/ocl.cpp
@@ -4705,6 +4705,10 @@ public:
             int createFlags = 0, flags0 = 0;
             getBestFlags(ctx, accessFlags, usageFlags, createFlags, flags0);
 
+            bool copyOnMap = (flags0 & UMatData::COPY_ON_MAP) != 0;
+            if( copyOnMap )
+                accessFlags &= ~ACCESS_FAST;
+
             cl_context ctx_handle = (cl_context)ctx.ptr();
             int allocatorFlags = 0;
             int tempUMatFlags = 0;
@@ -4765,7 +4769,8 @@ public:
 #endif
             {
                 tempUMatFlags = UMatData::TEMP_UMAT;
-                if (CV_OPENCL_ENABLE_MEM_USE_HOST_PTR
+                if (!copyOnMap
+                    && CV_OPENCL_ENABLE_MEM_USE_HOST_PTR
                     // There are OpenCL runtime issues for less aligned data
                     && (CV_OPENCL_ALIGNMENT_MEM_USE_HOST_PTR != 0
                         && u->origdata == cv::alignPtr(u->origdata, (int)CV_OPENCL_ALIGNMENT_MEM_USE_HOST_PTR))
@@ -4793,7 +4798,7 @@ public:
             u->handle = handle;
             u->prevAllocator = u->currAllocator;
             u->currAllocator = this;
-            u->flags |= tempUMatFlags;
+            u->flags |= tempUMatFlags | flags0;
             u->allocatorFlags_ = allocatorFlags;
         }
         if(accessFlags & ACCESS_WRITE)
--- a/modules/core/src/ocl.cpp
+++ b/modules/core/src/ocl.cpp
@@ -4770,8 +4770,11 @@ public:
                     accessFlags &= ~ACCESS_FAST;
 
                 tempUMatFlags = UMatData::TEMP_UMAT;
-                if (!copyOnMap
-                    && CV_OPENCL_ENABLE_MEM_USE_HOST_PTR
+                if (
+                #ifdef __APPLE__
+                    !copyOnMap &&
+                #endif
+                    CV_OPENCL_ENABLE_MEM_USE_HOST_PTR
                     // There are OpenCL runtime issues for less aligned data
                     && (CV_OPENCL_ALIGNMENT_MEM_USE_HOST_PTR != 0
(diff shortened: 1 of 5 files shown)


vpisarev:
@alalek, could you please add the magic comment to enable OpenCL on Mac?

modules/core/src/ocl.cpp
```diff
@@ -4705,6 +4705,10 @@ class OpenCLAllocator CV_FINAL : public MatAllocator
             int createFlags = 0, flags0 = 0;
             getBestFlags(ctx, accessFlags, usageFlags, createFlags, flags0);
 
+            bool copyOnMap = (flags0 & UMatData::COPY_ON_MAP) != 0;
+            if( copyOnMap )
+                accessFlags &= ~ACCESS_FAST;
```
alalek: Do you remember which test triggers that?

I have [forced](https://github.com/opencv/opencv/commit/d134b78d4c9031b1e566b7a8c9df6c9073c4b6a5) `hostUnifiedMemory = false`, but [nothing](http://pullrequest.opencv.org/buildbot/builders/precommit_custom_mac/builds/535) is failed on Macs.

modules/core/src/opencl/flip.cl
```diff
@@ -42,10 +42,20 @@
...
 #else
 #define loadpix(addr) vload3(0, (__global const T1 *)(addr))
 #define storepix(val, addr) vstore3(val, 0, (__global T1 *)(addr))
+#define storepix_2(val0, val1, addr0, addr1) \
+    ((__global T1 *)(addr0))[0] = val0.x; \
+    ((__global T1 *)(addr1))[0] = val1.x; \
+    ((__global T1 *)(addr0))[1] = val0.y; \
+    ((__global T1 *)(addr1))[1] = val1.y; \
+    ((__global T1 *)(addr0))[2] = val0.z; \
+    ((__global T1 *)(addr1))[2] = val1.z
```
alalek: This looks strange and may affect performance (especially in interleaved mode).
Could you please limit this approach for CV_16U/CV_16S only?

modules/core/src/ocl.cpp
vpisarev: @alalek, the test is "UMat.SyncTemp". When you put some condition and nothing is failed, it does not mean that there is no bug in the logic. Before the patch the code attempts to do Map/Unmap even on machines without unified (read "shared") GPU-CPU memory. Somehow modern drivers on Windows/Linux implement this map unmap by doing implicit copying. On Mac it does not work well. On your macs everything works well probably because there is just iGPU there, no dGPU. I have 2 macs with dGPU, and they just freeze on "UMat.SyncTemp" test for infinite amount of time. After the patch everything works well.

alalek: > On Mac it does not work well.

So, where should we put `#ifdef __APPLE__`?

Dropping `ACCESS_FAST` flag here does not look like a good idea.
AFAIK, it was designed to fail fast if any data copying is required.

Please change in `ocl.cpp`:
```.diff
-#define CV_OPENCL_TRACE_CHECK                    0
+#define CV_OPENCL_TRACE_CHECK                    1
```

and collect stdout dump on problematic test (e.g, which requires this fix).

---

BTW, I can't find any place where `ACCESS_FAST` is enabled (including tests).
[GitHub results](https://github.com/opencv/opencv/search?q=ACCESS_FAST&type=Code): 1 declaration + 2 checks.

vpisarev: With the tracing turned on sometimes it hangs on unmap:
```
OpenCV(OpenCL:0): clCreateBuffer(CL_MEM_USE_HOST_PTR|createFlags, sz=100, origdata=0x7fa0d5278680) => 0x7fa0d5277ea0
OpenCV(OpenCL:0): clCreateBuffer() => 0x7fa0d5277ea0
OpenCV(OpenCL:0): handle = clCreateCommandQueue_pfn(ch, dh, props, &retval)
OpenCV(OpenCL:0): clEnqueueReadBuffer(q, handle=0x7fa0d5277ea0, CL_TRUE, 0, sz=100, data=0x7fa0d5278680, 0, 0, 0)
OpenCV(OpenCL:0): clEnqueueWriteBuffer(q, handle=0x7fa0d5277ea0, CL_TRUE, 0, sz=100, data=0x7fa0d5278680, 0, 0, 0)
[ INFO:0] Successfully initialized OpenCL cache directory: /var/folders/ll/ddsp7fq10lvb2lf9v91x6ks40000gn/T/opencv/3.4.x-dev/opencl_cache/
[ INFO:0] Preparing OpenCL cache configuration for context: 32-bit--AMD--AMD_Radeon_HD_-_FirePro_D700_Compute_Engine--1_2__Jan_13_2020_20_12_40_
OpenCV(OpenCL:0): clBuildProgram(binary: core/copyset)
OpenCV(OpenCL:0): result = clGetProgramBuildInfo_pfn(handle, devices[0], 0x1181, sizeof(build_status), &build_status, &retsz)
OpenCV(OpenCL:0): clCreateKernel('set')
OpenCV(OpenCL:0): clSetKernelArg('set', arg_index=0, cl_mem=0x7fa0d5277ea0)
OpenCV(OpenCL:0): clSetKernelArg('set', arg_index=1, step_value=10)
OpenCV(OpenCL:0): clSetKernelArg('set', arg_index=2, offset_value=0)
OpenCV(OpenCL:0): clSetKernelArg('set', arg_index=3, rows_value=10)
OpenCV(OpenCL:0): clSetKernelArg('set', arg_index=4, cols_value=5)
OpenCV(OpenCL:0): clSetKernelArg('set', arg_index=5, size=2, obj=0x7ffeea546ea0)
OpenCV(OpenCL:0): clFinish_pfn(qq)
OpenCV(OpenCL:0): clReleaseKernel_pfn(handle)
OpenCV(OpenCL:0): clEnqueueReadBuffer(q, handle=0x7fa0d5277ea0, CL_TRUE, 0, sz=100, data=0x7fa0d5278680, 0, 0, 0)
OpenCV(OpenCL:0): clEnqueueWriteBuffer(q, handle=0x7fa0d5277ea0, CL_TRUE, 0, sz=100, data=0x7fa0d5278680, 0, 0, 0)
OpenCV(OpenCL:0): clEnqueueMapBuffer(handle=0x7fa0d5277ea0, sz=100) => 0x7fa0d5278680
OpenCV(OpenCL:0): clEnqueueUnmapMemObject(handle=0x7fa0d5277ea0, data=0x7fa0d5278680, [sz=100])
```

sometimes it crashes on the OpenCL kernel (setTo) release,
sometimes it crashes at some other random places.

if the code that removes ACCESS_FAST flag fails, no any error/exception happens. The code returns UMat with OpenCL handle=NULL. Then, obviously, OpenCL kernel refuses to run and then it automatically falls back to CPU implementation. But I'm almost sure that this is much worse and error-prone behaviour than the real data copy. I moved ACCESS_FAST remove code into non-SVM branch, because ACCESS_FAST is really for SVM.

PR #16614, opened 2020-02-18
added estimateTranslation3D to calib3d/ptsetreg

GFleishman:
I've done my best to follow the PR guidelines, but this is my first PR to OpenCV, so I anticipate some back and forth; I'm grateful for your time.

Currently the RANSAC filter in calib3D/ptsetreg.cpp only permits full affine transformations. I have an application (and others may as well) that requires limiting the transform to translation only.

The new function follows the API and implementation structure for estimateAffine3D exactly, only the implementations of runKernel, computeError, and checkSubset are modified (and are considerably simpler) for the new class - users can call estimateTranslation3D as a drop in replacement for estimateAffine3D whenever it is appropriate.

I've built successfully on my system and implemented tests. estimateTranslation3D will be strictly faster than estimateAffine3D as it does not require SVD.

I believe I have documented the new code properly, as I simply copied the documentation strings from estimateAffine3D and modified them to be consistent with translation only. Please advise if I'm missing anything.

- [NA] There is reference to original bug report and related work
- [X] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
- [X] The feature is well documented and sample code can be built with the project CMake

### DIFF
--- a/modules/calib3d/src/ptsetreg.cpp
+++ b/modules/calib3d/src/ptsetreg.cpp
@@ -519,7 +519,7 @@ public:
  *      t2
  *      t3
  */
-class Translation3DEstimatorCallback : public PointSetRegistrator::Callback
+class Translation3DEstimatorCallback CV_FINAL : public PointSetRegistrator::Callback
 {
 public:
     int runKernel( InputArray _m1, InputArray _m2, OutputArray _model ) const CV_OVERRIDE
@@ -529,8 +529,7 @@ public:
         const Point3f* from = m1.ptr<Point3f>();
         const Point3f* to   = m2.ptr<Point3f>();
 
-        double buf[3] = {0, 0, 0};
-        Matx13d T(buf);
+        Matx13d T;
 
         // The optimal translation is the mean of the pointwise displacements
         for(int i = 0; i < 4; i++)
--- a/modules/calib3d/src/ptsetreg.cpp
+++ b/modules/calib3d/src/ptsetreg.cpp
@@ -505,6 +505,89 @@ public:
     }
 };
 
+
+/*
+ * Compute
+ *  x      X     t1
+ *  y  =   Y  +  t2
+ *  z      Z     t3
+ *
+ *  - every element in _m1 contains (X,Y,Z), which are called source points
+ *  - every element in _m2 contains (x,y,z), which are called destination points
+ *  - _model is of size 3x1, which contains
+ *      t1
+ *      t2
+ *      t3
+ */
+class Translation3DEstimatorCallback : public PointSetRegistrator::Callback
+{
+public:
+    int runKernel( InputArray _m1, InputArray _m2, OutputArray _model ) const CV_OVERRIDE
+    {
+
+        Mat m1 = _m1.getMat(), m2 = _m2.getMat();
+        const Point3f* from = m1.ptr<Point3f>();
+        const Point3f* to   = m2.ptr<Point3f>();
+
+        double buf[3];
+        Mat T(3, 1, CV_64F, &buf[0]);
+        double* Tptr = T.ptr<double>();
+        T = Scalar::all(0);
+
+        // The optimal translation is the mean of the pointwise displacements
+        for(int i = 0; i < 4; i++)
+        {
+            const Point3f& f = from[i];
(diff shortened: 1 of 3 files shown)


GFleishman:
I believe the build error is because of warnings only. The warnings are because checkSubset in my implementation of Translation3DEstimatorCallback does not use the input arguments (i.e. unused arguments errors). But checkSubset must be implemented with these arguments to be consistent with objects which use Translation3DEstimatorCallback (namely RANSACPointSetRegistrator). Not sure of the best way to resolve this.

asmorkalov:
@GFleishman Good job! Could you add tests for the solution you introduced?

GFleishman:
@asmorkalov Yep! I saw the tags last week but have been very busy. I think just copying and renaming the existing tests for estimateAffine3D will probably work. I may get to this today.

GFleishman:
@asmorkalov Ok! I've added `test_translation3d_estimator.cpp` to `modules/calib3d/test`
It compiles just fine, and when I execute the binary `opencv_test_calib3d` the new tests are included in the run and pass with OK. It was a very straight forward mapping of `test_affine3d_estimator.cpp`
Please let me know if there is anything else I should do to get this PR merged :)

asmorkalov:
@GFleishman Good job! Thanks for tests.

asmorkalov:
@vpisarev could you look on the PR?

vpisarev:
:+1:

vpisarev:
Great job, thank you!

modules/calib3d/src/ptsetreg.cpp
```diff
@@ -505,6 +505,93 @@ class Affine3DEstimatorCallback : public PointSetRegistrator::Callback
...
+    int runKernel( InputArray _m1, InputArray _m2, OutputArray _model ) const CV_OVERRIDE
+    {
+
+        Mat m1 = _m1.getMat(), m2 = _m2.getMat();
+        const Point3f* from = m1.ptr<Point3f>();
+        const Point3f* to   = m2.ptr<Point3f>();
+
+        double buf[3];
+        Mat T(3, 1, CV_64F, &buf[0]);
+        double* Tptr = T.ptr<double>();
```
alalek: Consider using `Matx13d` instead.

```.diff
-T.copyTo(_model);
+Mat(T, false).copyTo(_model);
```

modules/calib3d/src/ptsetreg.cpp
```diff
@@ -505,6 +505,93 @@ class Affine3DEstimatorCallback : public PointSetRegistrator::Callback
...
+            const Point3f& f = from[i];
+            const Point3f& t = to[i];
+
+            Tptr[0] = Tptr[0] + t.x - f.x;
+            Tptr[1] = Tptr[1] + t.y - f.y;
+            Tptr[2] = Tptr[2] + t.z - f.z;
+        }
+        Tptr[0] = Tptr[0] / 4;
+        Tptr[1] = Tptr[1] / 4;
+        Tptr[2] = Tptr[2] / 4;
```
alalek: `T *= (1.0f / 4);` ?

modules/calib3d/test/test_translation3d_estimator.cpp
```diff
@@ -0,0 +1,211 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
+//
+//  IMPORTANT: READ BEFORE DOWNLOADING, COPYING, INSTALLING OR USING.
```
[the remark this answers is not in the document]
GFleishman: Done!

modules/calib3d/src/ptsetreg.cpp
GFleishman: Done! Also required modifying `test_translation3d_estimator` to use the new data type

modules/calib3d/src/ptsetreg.cpp
GFleishman: Done! More concise 👍

GFleishman:
Hi guys - what is still required for merging?

alalek:
OpenCV code base is in code freeze state due release preparation. This patch will be merged after 3.4.10 release (targeted to 3.4.11).

PR #16616, opened 2020-02-18
dnn: fix processing of input shapes

alalek:
resolves #16618

- ~~importer(caffe): fix `net.input_dim` handling~~ (moved to #16617)
- importers: avoid using of `setInput()` => `setInputShape()`
- setInput(): shape limitation check (partial)

<cut/>

```
force_builders=Custom,Custom Win,Custom Mac
build_image:Custom=ubuntu-openvino-2020.1.0:16.04
build_image:Custom Win=openvino-2020.1.0
build_image:Custom Mac=openvino-2020.1.0

test_modules:Custom=dnn,python2,python3,java
test_modules:Custom Win=dnn,python2,python3,java
test_modules:Custom Mac=dnn,python2,python3,java

buildworker:Custom=linux-1
# disabled due high memory usage: test_opencl:Custom=ON
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

PR #16617, opened 2020-02-19
dnn(caffe): fix net.input_dim handling in Caffe importer

alalek:
resolves https://github.com/opencv/opencv_contrib/issues/2426
relates #15203

subset of #16616

<cut/>

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/caffe/caffe_importer.cpp
+++ b/modules/dnn/src/caffe/caffe_importer.cpp
@@ -75,6 +75,17 @@ static cv::String toString(const T &v)
     return ss.str();
 }
 
+static inline
+MatShape parseBlobShape(const caffe::BlobShape& _input_shape)
+{
+    MatShape shape;
+    for (int i = 0; i < _input_shape.dim_size(); i++)
+    {
+        shape.push_back((int)_input_shape.dim(i));
+    }
+    return shape;
+}
+
 class CaffeImporter
 {
     caffe::NetParameter net;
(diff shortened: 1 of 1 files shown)


modules/dnn/src/caffe/caffe_importer.cpp
```diff
@@ -334,12 +342,49 @@ class CaffeImporter
...
             {
                 addedBlobs.push_back(BlobNote(net.input(inNum), 0, inNum));
                 netInputs[inNum] = net.input(inNum);
             }
+
+            if (net.input_dim_size() > 0)  // deprecated in Caffe proto
+            {
+                int net_input_dim_size = net.input_dim_size();
+                CV_Check(net_input_dim_size, net_input_dim_size % 4 == 0, "");
+                CV_CheckEQ(net_input_dim_size, net_input_size * 4, "");
```
dkurt: Maybe just just check that number of `input_dim` is divided by number of inputs?
```cpp
CV_Check(net_input_dim_size, net_input_dim_size % net_input_size == 0, "");
```
So we could handle multiple inputs but with different number of dimensions (considering that it is the same for each input).

We have an example with different number of shapes (`faster_rcnn_vgg16.prototxt`). Looks like that's why `input_dim` is deprecated in Caffe.
```yaml
input: "data"
input_shape {
  dim: 1
  dim: 3
  dim: 224
  dim: 224
}

input: "im_info"
input_shape {
  dim: 1
  dim: 3
}
```

alalek: This code block is about "input_dim".
It is strictly restricted to 4 elements, see [proto file comment](https://github.com/opencv/opencv/blob/3.4.9/modules/dnn/src/caffe/opencv-caffe.proto#L227-L231).

"input_shape" doesn't have such limitations (related code is below).

dkurt: Got it, thanks!

PR #16625, opened 2020-02-20
findChessboardCornersSB improvements

D-Alex:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/713

**Improvements:**

* UMat for blur + rotate resulting in a speedup of around 2X on an i7
* support for boards larger than specified allowing to cover full FOV
* support for markers moving the origin into the center of the board
* support for measuring target quality
* fixing couple of issues preventing a detection of the board

The main change is for supporting boards that are larger than the FOV of
the camera and have their origin in the board center. This allows
building OEM calibration targets similar to the one from intel real
sense utilizing corner points as close as possible to the image border. 

Furthermore, a method was added to estimate the sharpness of the detected chessboard. This allows to filter out unsharp calibration images or images subject to motion blur, which would otherwise corrupt the calibration result.

The improved detector is the base for the article posted here:
https://medium.com/@hey_duda/the-magic-behind-camera-calibration-8596b7ddcd71

```
opencv_extra=master
allow_multiple_commits=1
```


tompollok:
Hi @D-Alex,

as always, thanks for your great work in this field!

Just one question: So what i do is that i typically take images from multiple chessboard poses with a fixed chessboard size. In your images from the blog you use a large chessboard patterns and find a "dynamically" sized chessboard. How do you handle that? Dont you have to give chessboards with the same dimensions into the calibration routine? Or can you just give every chessboard its own dynamically adjusted 3D corrdinates?

Also: i see that you said that blurred images should be filtered out and you added a function estimateChessboardSharpness to estimate the sharpness for each board. Doesnt it make sense to also provide a funtion that returns the angle to the camera axis, so that the ones orthogonal are automatically removed?

tompollok:
Regarding "fixing couple of issues preventing a detection of the board": would you mind explaining which cases have been improved?

D-Alex:
@tompollok for a mono camera, you can use a different coordinate system for each chessboard and feed different sized chessboards to the calibration routine as long as the object z-coordinates are zero and all boards are rectangular. 

However, in the case of a stereo system, you need the same coordinate system on the target, which is solved by adding little circles to mark the origin of the board visible in all calibration images.

Concerning the orthogonal case. Little noise can completely change the orientation of the board. This is visible by a jumping overlayed pose indicator. They can only be removed after calibration because the calibration process itself will determine the pose.  Usually, I try to avoid them by using a fixed target in V-configuration. The calibration process is as follows:

For each of the two poses:
- find first board and mask it
- find second board 
- calculate sharpness and brightness
- discard image if unsharp or overexposed (tell the user to adjust camera)

Concerning the detection rate:
- The algorithm works by calculating a feature map with a high response value at locations where possible chessboard corners are.
- The best points are stored in a kd-tree 
- Based on these points, the algorithm tries to grow multiple chessboards using cross-ratio constraints (the largest one wins). 
- Here several issues were solved, leading to a wrong search area in the image for the next row/col of the chessboard preventing the growth of the chessboard under challenging scenarios such as high perspective distortion.

D-Alex:
It looks like "Compare ABI dumps" failed because findChessboardSB has now an additional optional parameter for forwarding the actual board size in the case a larger board was found including some meta data.

What would be the best approach for fixing this and allowing an existing method to have a new optional parameter?

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -1257,7 +1277,40 @@ a sample checkerboard optimized for the detection. However, any other checkerboa
 can be used as well.
 ![Checkerboard](pics/checkerboard_radon.png)
  */
-CV_EXPORTS_W bool findChessboardCornersSB(InputArray image,Size patternSize, OutputArray corners,int flags=0);
+CV_EXPORTS_W bool findChessboardCornersSB(InputArray image,Size patternSize, OutputArray corners,
+                                          int flags=0,OutputArray meta= noArray());
```
alalek: Try this to pass ABI check:
```
CV_EXPORTS_AS(findChessboardCornersSBWithMeta)
//or CV_EXPORTS to bypass Python/Java bindings
bool findChessboardCornersSB(InputArray image, Size patternSize, OutputArray corners,
                             int flags, OutputArray meta);
/** @overload */
CV_EXPORTS_W static inline
bool findChessboardCornersSB(InputArray image, Size patternSize, OutputArray corners,
                             int flags = 0)
{
    rerturn findChessboardCornersSB(image, patternSize, corners, flags, noArray());
}
```

alalek: > CV_EXPORTS_W static inline

Please remove `static` or `CV_EXPORTS_W`

D-Alex:
The required additional meta data for testing estimateChessboardSharpness:
https://github.com/opencv/opencv_extra/pull/713

The test is performed for chessboards found by findChessboardCorners and findChessboardCornersSB. In the case the sharpness is unknown for a given image, the test is bypassed for the particular image.

alalek:
@D-Alex Thank you for the contribution!

Do you have a chance to look on current tests failures?
On next week we are going to start preparations for the next releases, so OpenCV repo will be moved in "code freeze" stage in 2 weeks (for 1-2 weeks until final release).

D-Alex:
@alalek - I should be able to fix all suggestions over this weekend.

modules/calib3d/src/chessboard.cpp
```diff
@@ -611,37 +703,41 @@ void FastX::detectImpl(const cv::Mat& _gray_image,
     CV_CheckTypeEQ(_gray_image.type(), CV_8UC1, "Unsupported image type");
 
     // up-sample if needed
-    cv::Mat gray_image;
-    int super_res = int(parameters.super_resolution);
+    cv::UMat gray_image;
```
mshabunin: I though we want to get rid of all implicit GPU calls in algorithms' internals like we did with CascadeClassifier.

Issue #16002, opened 2019-11-26
DNN Torch import is missing several layers


cyberbeat:
##### System information (version)
- OpenCV => master (4.2.x)
- Operating System / Platform => Linux 64 Bit
- Compiler => gcc9

##### Detailed description

I try to import a dnn model from torch, and several layers cannot be created.

The model is for deblurring images:
https://arxiv.org/abs/1612.02177

This is the code:
https://github.com/SeungjunNah/DeepDeblur_release
with links to the model-files:
https://drive.google.com/file/d/1Z8dV6KuubfOKj4ganEjxymhyMoXoydfo/view?usp=sharing

The missing torch layers are:
- PixelShuffle
- Copy
- AddConstant
- ParallelTable
- FlattenTable
- NarrowTable
- SelectTable

I think I got Copy and AddConstant simply implemented via custom layer (remap to opencv layers). I also tried PixelShuffle, I am not sure, if I got it right. But the table layers seem to be difficult with custom layers in opencv?

czgdp1807:
I would like to take up this issue.

czgdp1807:
@cyberbeat Can you please share the CPP/Python  code which you used to import the `.t7` files. I tried to find `ParallelTable`, `FlattenTable`, `SelectTable` in PyTorch but they didn't appear to be there.
```python
>>> dir(torch.nn.FlattenTable)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'torch.nn' has no attribute 'FlattenTable'
>>> dir(torch.nn.SelectTable)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'torch.nn' has no attribute 'SelectTable'
>>> dir(torch.nn.PixelShuffle)
[...]

PR #16631, opened 2020-02-21
[WIP] Added Missing Layers for TorchImporter

czgdp1807:
resolves https://github.com/opencv/opencv/issues/16002

Other than the above checklist, the following is to be done for completing this PR,

- [ ] Add CopyLayer
- [ ] Add PixelShuffleLayer
- [ ] Add AddConstantLayer
- [ ] Add ParallelTable
- [ ] Add FlattenTable
- [ ] Add NarrowTable
- [ ] Add SelectTable
- [ ] Add Tests for each one of the above layers.
- [ ] Squash all commits into 1


czgdp1807:
I have added constructors for the missing layers. I will now add `getMemoryShapes` for each of the added layers. Please review the interface and let me know if anything is to be changed. Thank you.

modules/dnn/include/opencv2/dnn/all_layers.hpp
```diff
@@ -648,6 +648,62 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
 
+    class CV_EXPORTS CopyLayer : public Layer
+    {
+        public:
+
+            String intype, outtype;
+            bool forceCopy, dontCast;
+
+            static Ptr<Layer> create(const LayerParams& params);
+    };
```
dkurt: I think that Copy can be mapped to something that already exists (Identity in example)

czgdp1807: Sure, I will try to find an alternative for it and will make a fix there.

modules/dnn/include/opencv2/dnn/all_layers.hpp
```diff
@@ -648,6 +648,62 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
...
+
+    class CV_EXPORTS ParallelTableLayer : public Layer
+    {
+        public:
+
+            static Ptr<Layer> create(const LayerParams& params);
+    };
```
dkurt: ParallelTable is not a layer but a way Torch splits the topology on branches. Maybe we just need to process it like all other `*Table` entries?

https://github.com/opencv/opencv/blob/ec43292e1ea9da963e67427505b4113750829c3e/modules/dnn/src/torch/torch_importer.cpp#L501

modules/dnn/src/dnn.cpp
```diff
@@ -4531,7 +4531,6 @@ Ptr<Layer> LayerFactory::createLayerInstance(const String &type, LayerParams& pa
     cv::AutoLock lock(getLayerFactoryMutex());
     String type_ = type.toLowerCase();
     LayerFactory_Impl::const_iterator it = getLayerFactoryImpl().find(type_);
-
```
dkurt: please avoid extra changes (adding or removing new  empty lines)

modules/dnn/include/opencv2/dnn/all_layers.hpp
czgdp1807: I looked at all the layers currently present in `layers/` and observed that `Identity` layer is not present. In addition, the implementation [here](https://github.com/torch/nn/blob/master/Copy.lua) suggests that it casts the input from `intype` to `outtype`, may be identity will not be able to do that from what I can figure out from the implementation of Identity layer available [here](https://github.com/torch/nn/blob/master/Identity.lua).
Please let me know your views.

modules/dnn/src/dnn.cpp
czgdp1807: Done locally, will push in the next commits.

modules/dnn/include/opencv2/dnn/all_layers.hpp
czgdp1807: I will try the above approach locally.

modules/dnn/include/opencv2/dnn/all_layers.hpp
dkurt: There is:

https://github.com/opencv/opencv/blob/4e3c09b4766d14569ae5ef75e52e57d3c18416cb/modules/dnn/src/init.cpp#L115

By default, OpenCV work with FP32 tensors data

czgdp1807: May be then mapping Copy to Identity will be a nice option. Thanks for the help. :-)

czgdp1807: It works for the model shared in the issue. Resolving.

czgdp1807: For most use cases it will work, though I am not sure, what will happen if, the `intype` and `outtype` are different, may be something like, `intype` is FP and `outtype` is `int`, where `Copy` will be helpful to truncate the decimal points.

czgdp1807:
Closing. Thanks for the help.

PR #16635, opened 2020-02-21
intrin: fixed int64->double conversion for AVX-512

mshabunin:
related #15510

Wrong intrinsics were used in AVX-512 without DQ fallback implementation.

```
buildworker:Custom=linux-3
disable_ipp=ON
```


modules/core/include/opencv2/core/hal/intrin_avx512.hpp
```diff
@@ -1553,13 +1553,13 @@ inline v_float64x8 v_cvt_f64(const v_int64x8& v)
     return v_float64x8(_mm512_cvtepi64_pd(v.val));
```
seiko2plus: My bad forget to remove it, however, no need to emulate it, since the minimum support for AVX512 support is SKX features, which always enables  `AVX512DQ`.

PR #16643, opened 2020-02-21
Add CIFuzz action

Leo-Neat:
OSS-Fuzz was wondering if opencv would use CIFuzz, a new feature being rolled out 
in OSS-Fuzz. It allows for pull requests to be fuzzed in CI using fuzz targets enabled 
in OSS-Fuzz. Documentation for this feature can be found [here](https://google.github.io/oss-fuzz/getting-started/continuous-integration/). 
Let us know if you have any questions!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,23 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'opencv'
+        dry-run: false
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'opencv'
+        fuzz-seconds: 600
+        dry-run: false
+    - name: Upload Crash
+      uses: actions/upload-artifact@v1
+      if: failure()
+      with:
+        name: artifacts
+        path: ./out/artifacts


dkurt:
> It allows for pull requests to be fuzzed in CI using fuzz targets enabled
in OSS-Fuzz. 

Thanks for proposal! How log it takes to run fuzz tests?

alalek:
There is no completion time applicable for Fuzz tests. They may run "forever" (expect some special cases).

> Determines how long CIFuzz spends fuzzing your project in seconds. The default is 600 seconds. The GitHub Actions max run time is 21600 seconds (6 hours).

https://google.github.io/oss-fuzz/getting-started/continuous-integration/

dkurt:
@alalek, does it mean that they are not deterministic at every run? I do think that it is not the best quality for pull requests validation.

Leo-Neat:
@dkurt It is deterministic every run. It runs the fuzzers for a specified amout of time e.g. 600 seconds. If a crash is detected in that time period the test reads as failure, if not it reads as passing. This is done in order to prevent vulnerabilities being submitted to the codebase.

dkurt:
@Leo-Neat that means there is a way to determine which code should be fuzzed depends on the changes in PR?

Leo-Neat:
@dkurt We are currently developing a feature that uses code paths to determine the affected fuzzers and only run them. This feature is not out yet but should be soon. In its current state every fuzzer is run for an equal amount of time.

dkurt:
@Leo-Neat, Got it, thanks! I'm just not sure that we really need it as a part of pull requests validation. Maybe nightly or weekly? The thing is that OpenCV has thousands of methods that means 600 seconds can cover only a part of them. So for some of the pull requests this test may be failed even if it's not connected to it's changes - just a new method has been added to this 600 seconds of fuzzing.

Leo-Neat:
@dkurt CIFuzz will not report a bug unless it was introduced since the last OSS-Fuzz build. Once we have the code coverage feature it should have pretty much no false positives and catch a lot of mistakes. I will re-submit when the feature Is implemented.

PR #16644, opened 2020-02-22
Using argv[0] represent binary executable files' name in help() function in sample codes instead of placeholder name.

MoonChasing:

**In samples/cpp dictory, the edge.cpp has misspelling. It shoule be**
```C++
static void help()
{
    printf("\nThis sample demonstrates Canny edge detection\n"
           "Call:\n"
           "    ./edge [image_name -- Default is fruits.jpg]\n\n");
}
```


samples/cpp/edge.cpp
```diff
@@ -43,7 +43,7 @@ static void help()
 {
     printf("\nThis sample demonstrates Canny edge detection\n"
            "Call:\n"
-           "    /.edge [image_name -- Default is fruits.jpg]\n\n");
+           "    ./edge [image_name -- Default is fruits.jpg]\n\n");
```
alalek: It is still wrong, because names of binaries are differ (has `example_` prefix).

MoonChasing: Okay,  I get it. And I wonder if it's worth creating a PR. If it's not worth it, I won't bother you. Thanks.

alalek: Take a look on other samples, like this:
https://github.com/opencv/opencv/blob/102ef39a27bfc277d82045e5c5e0d4b1251294df/samples/cpp/inpaint.cpp#L16

PR #16650, opened 2020-02-23
Darknet_io: Parsing for cost layer added

### LINKED ISSUE #16623: Darknet importer: unable to import clasification configuration file darknet19.cfg
Using one of the cfg file from darknet repo. (darknet-19 for classification not detection).

https://github.com/AlexeyAB/darknet/blob/master/cfg/darknet19.cfg

opencv/modules/dnn/src/darknet/darknet_io.cpp:678: error: (-212:Parsing error) Unknown layer type: sse in function 'ReadDarknetFromCfgStream'

Seems to be in cost layer...

[cost]
type=sse

ashishkrshrivastava:
Merge with extra: [opencv/opencv_extra#714](https://github.com/opencv/opencv_extra/pull/714)
resolves #16623  

`opencv_extra=cvdarknet`

### This PR adds
Parsing for `[cost]` layer in darknet_io. Modifications are made to treat `[cost] `layer as `identity` layer.

### DIFF
--- a/modules/dnn/src/darknet/darknet_io.cpp
+++ b/modules/dnn/src/darknet/darknet_io.cpp
@@ -556,7 +556,7 @@ namespace cv {
                             const size_t layer_type_size = line.find("]") - 1;
                             CV_Assert(layer_type_size < line.size());
                             std::string layer_type = line.substr(1, layer_type_size);
-                            net->layers_cfg[layers_counter]["type"] = layer_type;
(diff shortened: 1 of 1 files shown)


dkurt:
@ashishkrshrivastava, Please add this layer to one of the existing `.cfg` files from tests so we can cover it.

ashishkrshrivastava:
@dkurt , I have added `[cost] `in `connected.cfg` as you suggested.

PR #16651, opened 2020-02-23
Using argv[0] represent binary executable files' name in help() function in sample codes instead of cpp files' name.

MoonChasing:
In some sample files, when display usage, it will use the cpp files' name as binary executable files' name.  However, it should be actual file name. So I use the `argv[0]`  instead of cpp files' name.

For example,  sample files code as follows.
```C++
static void help()
{
    printf("\nThis sample demonstrates Canny edge detection\n"
           "Call:\n"
           "    /.edge [image_name -- Default is fruits.jpg]\n\n");
}
```

Now, I update it to:
```C++
static void help(char** argv)
{
    printf("\nThis sample demonstrates Canny edge detection\n"
           "Call:\n"
           "    %s [image_name -- Default is fruits.jpg]\n\n", argv[0]);
}
```

### DIFF
--- a/samples/cpp/3calibration.cpp
+++ b/samples/cpp/3calibration.cpp
@@ -17,10 +17,10 @@ using namespace std;
 
 enum { DETECTION = 0, CAPTURING = 1, CALIBRATED = 2 };
 
-static void help()
+static void help(char** argv)
 {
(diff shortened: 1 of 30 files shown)


MoonChasing:
Sorry for my terrible git operation.

PR #16653, opened 2020-02-23
core(MatExpr): drop from InputArray

alalek:
Rework `MatExpr` implementation in `InputArray`:
- avoid multiple expression evaluations
- avoid issues with reduced support of `InputArray::EXPR`

<cut/>

TODO:
- [x] tests
- [ ] (backlog) drop EXPR kind from InputArray implementation

```
allow_multiple_commits=1
```


alalek:
:+1:

PR #16658, opened 2020-02-24
cuda4dnn(activations, eltwise, scale_shift): refactor to reduce code duplication

YashasSamaga:

Variadic templates and many metaprogramming features were introduced in CUDA 7. A lot of code was duplicated in the GSoC PR in an attempt to support older versions of CUDA. Since the built-in half-precision operators which are currently used require CUDA 10, there is no point in having this massively duplicated code anymore.

```
force_builders=Custom
buildworker:Custom=linux-4
```


alalek:
@YashasSamaga Thank you for the cleanup! Did you check for performance regressions?

YashasSamaga:
@alalek The new kernels generate identical assembly. Hence, there should be no change in performance.

PR #16663, opened 2020-02-25
Fixing typo in "-DOPENCV_"

flaviostutz:
I've fixed a lot of problems because people are using -DOPENCV_ in OpenCV CMake tasks and pkg file is not generated. These errors are very difficult to find.

Now I just realized that the root of those problems seems to be in the documentation.

see #13154

### DIFF
--- a/doc/tutorials/introduction/linux_install/linux_install.markdown
+++ b/doc/tutorials/introduction/linux_install/linux_install.markdown
@@ -98,7 +98,7 @@ Building OpenCV from Source Using CMake
         will be statically linked with OpenCV and can take a lot of memory.
 
 -#  [optional] Generate pkg-config info
-    -   Add this flag when running CMake: `-DOPENCV_GENERATE_PKGCONFIG=ON`
+    -   Add this flag when running CMake: `-D OPENCV_GENERATE_PKGCONFIG=ON`
     -   Will generate the .pc file for pkg-config and install it.
     -   Useful if not using CMake in projects that use OpenCV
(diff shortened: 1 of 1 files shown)


doc/tutorials/introduction/linux_install/linux_install.markdown
```diff
@@ -98,7 +98,7 @@ Building OpenCV from Source Using CMake
         will be statically linked with OpenCV and can take a lot of memory.
 
 -#  [optional] Generate pkg-config info
-    -   Add this flag when running CMake: `-DOPENCV_GENERATE_PKGCONFIG=ON`
+    -   Add this flag when running CMake: `-D OPENCV_GENERATE_PKGCONFIG=ON`
```
saskatchewancatch: Interesting. I thought the space between the D and option token was optional for CMake when specified on the CLI. When I use -DOPENCV_GENERATE_PKGCONFIG=On as part of the initial cmake configuration command, and open up the CMakeCache in the cmake-gui, it shows the option as set. Subsequently running cmake -DOPENCV_GENERATE_PKGCONFIG=Off disables pkgconfig generation in cmake-gui .

alalek:
This is not a typo.

CMake accepts both `-D<var>=<value` and `-D<space><var>=<value>`.

You can find several discussions about that.

flaviostutz:
Ok. Maybe in my fixes I made something else that solved the problem along with this supposed “fix”. Thanks for the clarification.

Issue #15897, opened 2019-11-12
Feature Request: 3+ dimension GpuMat


Llcoolsouder:
##### Detailed description

To my knowledge, OpenCV still has no implementation for GpuMats with more than 2 dimensions.
This has been a feature of `cv::Mat` for a long time, but for some reason it has never been applied to `GpuMat`, which creates a disconnect when transferring algorithms from CPU implementations to GPU implementations.

This is an example of a `cv::Mat` function that creates a multi-dimensional matrix:
https://docs.opencv.org/master/d3/d63/classcv_1_1Mat.html#a156df5a1326dd5c30b187b0e721a5f57

Issue #16433, opened 2020-01-25
GpuMat as input/output to cv::dnn::Net


YashasSamaga:
### System information (version)
- OpenCV => 4.2.0

##### APIs that need changes

Suitable APIs for `GpuMat` input:
1. `void setInput(InputArray blob, const String& name = "", double scalefactor = 1.0, const Scalar& mean = Scalar());`

Suitable APIs for `GpuMat` output:
1. `void forward(OutputArrayOfArrays outputBlobs, const String& outputName = String());`
2. `void forward(OutputArrayOfArrays outputBlobs, const std::vector<String>& outBlobNames);`

Modifications to support `GpuMat`:

1. `void blobFromImage(InputArray image, OutputArray blob, double scalefactor=1.0, const Size& size = Size(), const Scalar& mean = Scalar(), bool swapRB=false, bool crop=false, int ddepth=CV_32F);`

Might also require a new API that would allow a stream to be specified (required if `blobFromImage` needs to be performed asynchronously).

The CUDA backend uses fully packed tensors everywhere and expects a fully packed tensor as input. `GpuMat` given as input might be using a 2D memory layout which may not be contiguous. `blobFromImage` must return a contiguous copy of the input.

The outputs of the CUDA backend are also in fully packed tensors. Hence, for further optimal use of the output (in the case it's an image) would require a transformation to 2D memory layout.

##### Problems:
1. `GpuMat` cannot store an arbitrary number of dimensions. How can the batch of images be passed?

alalek:
@YashasSamaga Thank you for collecting overview of this problem! Good job!

---

> GpuMat cannot store an arbitrary number of dimensions

Right, current GpuMat implementation supports 2D layout only.
Problem is that existed CUDA API code expects 2D layout only - need to avoid massive refactoring of code (there are tons of user code which we are not able to fix anyway - compatibility problem).

---

> new API that would allow a stream to be specified

Consider using of dedicated `cuda` (`cv:dnn::cuda`) namespace.
Also we can introduce thread-local "default" stream API.

czgdp1807:
@alalek @YashasSamaga Is this issue still open for a PR? I would like to work on it.

> 1. `GpuMat` cannot store an arbitrary number of dimensions. How can the batch of images be passed?

Though I am very new to this side of OpenCV. However, may be we can use row major order at the time of indexing containers with arbitrary number of dimensions and in the memory we can store these containers as linear arrays. Please let me know if I have misinterpreted something wrong here.

[...]

PR #16666, opened 2020-02-25
[WIP] Add GpuMatND with arbitrary dimension support

czgdp1807:

This PR aims to add support for arbitrary dimension support as a first step to support the APIs given in https://github.com/opencv/opencv/issues/16433
As of creating this PR only interface is added and comments on the same is awaited.

```
force_builders_only=docs,Custom
buildworker:Custom=linux-4
build_image:Custom=ubuntu-cuda:16.04
```


modules/core/include/opencv2/core/cuda.hpp
```diff
@@ -337,6 +337,195 @@ class CV_EXPORTS GpuMat
...
+
+    //! default allocator
+    static Allocator* defaultAllocator();
+    static void setDefaultAllocator(Allocator* allocator);
+
+    //! default constructor
+    explicit GpuMatND(Allocator* allocator = defaultAllocator());
+
+    //! constructs GpuMatND of the specified shape and type
+    GpuMatND(int* shape, int type, Allocator* allocator = defaultAllocator());
```
czgdp1807: We can use variable number of arguments as well here using the APIs in `cstdarg` but in other libraries like TF, the shape is passed as a list, so passing it as an array is better and avoids complications of variable number of arguments.

alalek: > int* shape

This is unsafe API - missing `ndims`.

Consider targeting to "master" branch with C++11 support.
Use `std::array` and or `std::vector<int>` for passing ND array layouts.

modules/core/include/opencv2/core/cuda.hpp
```diff
@@ -337,6 +337,195 @@ class CV_EXPORTS GpuMat
...
+    GpuMatND(int* shape, int type, Scalar s, Allocator* allocator = defaultAllocator());
+
+    //! copy constructor
+    GpuMatND(const GpuMatND& m);
+
+    //! constructor for GpuMatND headers pointing to user-allocated data
+    GpuMatND(int* shape, int type, void* data, size_t step = Mat::AUTO_STEP);
+
+    //! builds GpuMatND from host memory (Blocking call)
+    explicit GpuMatND(InputArray arr, Allocator* allocator = defaultAllocator());
```
alalek: Host memory is `Mat` and not `InputArray` (which is some kind of wrapper over multiple types, but it doesn't store any data).

czgdp1807: Something similar is used in `GpuMat`, so do I need to make changes there as well?

czgdp1807:
> There should be interoperability with 2D `GpuMat`

I will try to keep this in mind while implementing the interface.

modules/core/include/opencv2/core/cuda.hpp
czgdp1807: > Consider targeting to "master" branch with C++11 support.

Done.

> Use `std::array` and or `std::vector<int>` for passing ND array layouts.

I have used `std::vector<int>`.

czgdp1807:
@YashasSamaga @alalek Please let me know if any change is to be made in the interface.

czgdp1807:
I hope I can start working on the implementation now. As no more comments are received on the interface. Probably we can minor changes during writing the implementation. Thanks.

Llcoolsouder:
Closes #15897

modules/core/include/opencv2/core/cuda.hpp
```diff
@@ -337,6 +337,195 @@ class CV_EXPORTS_W GpuMat
...
+    //! assignment operators
+    GpuMatND& operator =(const GpuMatND& m);
+
+    //! allocates new GpuMatND data unless the GpuMatND already has specified size and type
+    void create(std::vector<int> shape, int type);
+
+    //! decreases reference counter, deallocate the data when reference counter reaches 0
+    void release();
+
+    //! swaps with other smart pointer
```
YashasSamaga: implementation detail spilled into public API?

modules/core/include/opencv2/core/cuda.hpp
```diff
@@ -337,6 +337,195 @@ class CV_EXPORTS_W GpuMat
...
+    //! template version of the above method
+    template<typename _Tp> _Tp* ptr(int y = 0);
+    template<typename _Tp> const _Tp* ptr(int y = 0) const;
+
+    template <typename _Tp> operator PtrStepSz<_Tp>() const;
+    template <typename _Tp> operator PtrStep<_Tp>() const;
+
+    //! creates alternative GpuMatND header for the same data, with different
+    //! number of channels and/or different number of rows
+    GpuMatND reshape(int* new_shape) const;
```
YashasSamaga: how will you figure out the number of dimensions?

YashasSamaga:
@Llcoolsouder But this won't allocate 3D pitched memory.

@czgdp1807 I am not sure if a generic `GpuMatND` should be implemented or if a proxy object (like the one described in #16433) must be used.

The `GpuMat` interface was for C++98. I think it is better to have a C++11 interface (what's the point of having C++98 interface when OpenCV has already moved to C++11). The only problem would be that the interface would not be compatible with the `GpuMat`'s interface. It's just my opinion.

modules/core/include/opencv2/core/cuda.hpp
czgdp1807: Ah! I forgot to replace this with `vector<int> new_shape`. Thanks for pointing out.

czgdp1807:
> I think it is better to have a C++11 interface (what's the point of having C++98 interface when OpenCV has already moved to C++11).

Actually, I am not aware about which parts of the interface of `GpuMat` that are there due to C++98. For example, there is something called `Allocator`. Why don't we use `static` methods to reserve space in the memory and call them in the initializer list of constructor? Please let me know if we can do this and which functions can we remove from `GpuMatND`.

czgdp1807:
Hi,
Any thoughts on the ideas given in https://github.com/opencv/opencv/pull/16666#issuecomment-595111320 or can I proceed with the current interface given in the diff?
Please let me know.
Thanks.

nglee:
Is anyone still working on this? If no one's currently looking into it, I'd like to follow up on what's been discussed so far and work on this.

YashasSamaga:
Bumping. Is there any chance of this being completed soon? I think `GpuMat` is currently the only way to pass GPU memory to `InputArray`/`OutputArray`. This is very limiting. And not all GPU data are 2D images.

czgdp1807:
Hi, 

First of all apologies for not responding. It's been a long time since I updated this PR. If anyone's interested, please feel free to take this over and complete it. 

Apologies again. Stay safe.

asmorkalov:
Replaced with https://github.com/opencv/opencv/pull/19259

PR #16691, opened 2020-02-28
WIP: Localization of the ffmpeg compatibility macros

VadimLevin:
The general idea of this patch is to localize deprecated functionality from the FFMPEG libraries into the `compat` functions and make code cleaner and easier to maintain.

Some of the parts and workflows are updated according to the FFMPEG recommended functionality e.g. `avcodec_close` is changed to `avcodec_free_context`.

<cut/>

```
force_builders=Custom
build_image:Custom=ffmpeg-master
```

### DIFF
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -385,6 +385,17 @@ typedef AVBSFContext BitStreamFilterContext;
 typedef AVBitStreamFilterContext BitStreamFilterContext;
 #endif
 
+static inline void avCodecFreeContext(AVStream* stream) {
+    // functional is added CALC_FFMPEG_VERSION(55, 52, 0), but
+    // Memory leaks were fixed at
+    // https://github.com/FFmpeg/FFmpeg/commit/345cfd04d093d9fdec81ea3531e73b1f5c1b6a6c
+#if LIBAVCODEC_BUILD > CALC_FFMPEG_VERSION(56, 13, 100)
+    avcodec_free_context(&(stream->codec));
+#else
+    avcodec_close(stream->codec);
+#endif
+}
+
 static inline void avImageFillArrays(AVFrame* frame, uint8_t* pictureBuf, enum AVPixelFormat pix_fmt,
                                      int width, int height, int align = 1)
 {
@@ -459,7 +470,7 @@ static inline void avFormatFreeContext(AVFormatContext* formatContext)
 #endif
 }
 
-static inline AVStream* avFormatNewStream(AVFormatContext* formatContext, const AVCodec* codec = NULL)
+static inline AVStream* avFormatNewStream(AVFormatContext* formatContext, AVCodec* codec = NULL)
 {
 #if LIBAVFORMAT_BUILD >= CALC_FFMPEG_VERSION(53, 10, 0)
     return avformat_new_stream(formatContext, codec);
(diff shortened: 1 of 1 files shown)


modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -2281,186 +2423,71 @@ bool CvVideoWriter_FFMPEG::open( const char * filename, int fourcc,
...
+        };
+        if ((codec_id = av_codec_get_id(fallback_tags, fourcc)) == CV_CODEC(CODEC_ID_NONE))
         {
             fflush(stdout);
-            fprintf(stderr, "OpenCV: FFMPEG: tag 0x%08x/'%c%c%c%c' is not found (format '%s / %s')'\n",
-                    fourcc, CV_TAG_TO_PRINTABLE_CHAR4(fourcc),
-                    fmt->name, fmt->long_name);
+            fprintf(stderr,
+                    "OpenCV: FFMPEG: tag 0x%08x/'%c%c%c%c' is not found (format '%s / %s')'\n",
+                    fourcc, CV_TAG_TO_PRINTABLE_CHAR4(fourcc), fmt->name, fmt->long_name);
```
alalek: Please avoid playing with whitespaces.

Patch is huge (>100 lines), so don't make it worse for review.

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -3195,7 +3104,7 @@ bool InputMediaStream_FFMPEG::read(unsigned char** data, int* size, int* endOfFi
 
         if (pkt_.stream_index != video_stream_id_)
         {
-            _opencv_ffmpeg_av_packet_unref(&pkt_);
+            compat::avPacketUnref(&pkt_);
```
alalek: Please keep original function identifier in names (of the latest modern FFmpeg version).:
- original: av_packet_unref
- good compatibility wrapper name: `_opencv_ffmpeg_av_packet_unref`
- avoid that: `compat::avPacketUnref`

Approach:
```
... define _opencv_ffmpeg_av_packet_unref ...

#define av_packet_unref   av_packet_unref_must_be_wrapped  // guard

... other code without av_packet_unref() ...
```

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -1143,45 +1297,20 @@ bool CvCapture_FFMPEG::processRawPacket()
...
-#if LIBAVFORMAT_BUILD > 4628
-        AVCodecContext* ctx = ic->streams[video_stream]->codec;
-#else
-        AVCodecContext* ctx = &ic->streams[video_stream]->codec;
-#endif
-        int err = av_bitstream_filter_filter(bsfc, ctx, NULL, &packet_filtered.data,
-            &packet_filtered.size, packet.data, packet.size, packet_filtered.flags & AV_PKT_FLAG_KEY);
-        if (err < 0)
+        if (!compat::avBitstreamFilterFilter(bsfc, ic->streams[video_stream]->codec, packet,
+                                             packet_filtered))
```
alalek: Avoid extraction of one time used code into functions.

modules/videoio/src/cap_ffmpeg_impl.hpp
VadimLevin: I've tried to place all functions to `compatibility` namespace with pascal case naming (keeping the original naming).
But I've got your point and change it to what you suggest. Thank you.

modules/videoio/src/cap_ffmpeg_impl.hpp
VadimLevin: I have enabled autoformat, sorry.

modules/videoio/src/cap_ffmpeg_impl.hpp
VadimLevin: I tried to remove versions `#ifdef`'s from the actual functions of the capture and writer, to keep them simple and clean. What is the matter to do not do it?

VadimLevin:
> Main question here is what changes are really necessary?

The general points are: 
* Make functions simple and clean with removing `#ifdef`'s from the body of the functions, those do actual work. 
* Remove old functionality ([FFmpeg libs with versions < 0.5](https://github.com/FFmpeg/FFmpeg/tree/release/0.5))
* Fix wrong encoder/decoder initializations/freeing (e.g. do not use [`avcodec_get_context_defaults3`](https://github.com/FFmpeg/FFmpeg/commit/9b211c43dc5f2e618f204c4a7fd184eb2ea51f02))
* Add missing memory deallocations

Should I split this into several PR, can you suggest what is more important?

PR #16693, opened 2020-02-28
G-API: Include <imgproc.hpp> to <render.hpp>

Volskig:
### Added #include<imgproc.hpp> to <render.hpp>
Integer parameters to Drawing primitives require definitions from `imgproc.hpp` which is not included by default.

`#include <opencv2/imgproc.hpp>` already exists in `render.hpp`

Removed `#include <opencv2/imgproc.hpp>`  from `render.cpp`

### DIFF
--- a/modules/gapi/src/api/render.cpp
+++ b/modules/gapi/src/api/render.cpp
@@ -2,7 +2,6 @@
 
 #include <stdexcept>
 
-#include <opencv2/imgproc.hpp>
 #include <opencv2/gapi/render/render.hpp>
 #include <opencv2/gapi/own/assert.hpp>
 


alalek:
@Volskig Remove this from the patch history:

> Merge pull request #

modules/gapi/src/api/render.cpp
```diff
@@ -2,7 +2,6 @@
 
 #include <stdexcept>
 
-#include <opencv2/imgproc.hpp>
```
TolyaTalamanov: Could you add this header to render.hpp ?

Volskig: I can, but `render.cpp` include `render.hpp` from `gapi/opencv_gapi/include/opencv2/gapi/`. 

This `render.hpp` is just a shortcut to `render/render.hpp`.

`#include <opencv2/imgproc.hpp>` already exist in `render/render.hpp`.

TolyaTalamanov:
LGTM 👍

Volskig:
@alalek 1 commit now

PR #16704, opened 2020-03-01
core(logger): add CV_LOG_ONCE_xxx() CV_LOG_IF_xxx() macros

alalek:
Added macros:
- `CV_LOG_ONCE_xxx(tag, ...message)`
- `CV_LOG_IF_xxx(tag, logging_condition, ...message)`

### DIFF
--- a/modules/core/test/test_utils.cpp
+++ b/modules/core/test/test_utils.cpp
@@ -291,7 +291,7 @@ TEST(CommandLineParser, testScalar)
 }
 
 
-TEST(Logger, message)
+TEST(Logger, DISABLED_message)
 {
     int id = 42;
     CV_LOG_VERBOSE(NULL, 0, "Verbose message: " << id);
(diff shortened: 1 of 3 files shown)


modules/core/test/test_utils.cpp
```diff
@@ -287,6 +290,53 @@ TEST(CommandLineParser, testScalar)
...
 }
 
+
+TEST(Logger, message)
+{
+    int id = 42;
+    CV_LOG_VERBOSE(NULL, 0, "Verbose message: " << id);
+    CV_LOG_VERBOSE(NULL, 1, "Verbose message: " << id);
+    CV_LOG_DEBUG(NULL, "Debug message: " << id);
+    CV_LOG_INFO(NULL, "Info message: " << id);
```
VadimLevin: Google Test framework has 2 helper function to capture output: `testing::internal::CaptureStdout` and `testing::internal::CaptureStderr`. I would like to suggest to use the following pattern to check that message actually was logged:
```c++
    testing::internal::CaptureStdout();
    CV_LOG_INFO(NULL, "Info message: " << id);
    EXPECT_TRUE(testing::internal::GetCapturedStdout().find_first_of("Info message: 42") != std::string::npos);
```

alalek: Thank you for the suggestion.

Actually I want to keep these tests as disabled for now - check compilation only (and run them in manual mode if we have corresponding changes).

alalek:
:+1:

PR #16709, opened 2020-03-01
ONNX: Add support for Style trasfer models, ONNX model zoo

### LINKED ISSUE #16707: Unable to load Style transfer models from ONNX Model Zoo

- OpenCV => 3.4
- Operating System / Platform => Ubuntu 18.

##### Detailed description
Problem : Loading any model from  ONNX model zoo / Image Manipulation / [Fast Neural style transfer](https://github.com/onnx/models/tree/master/vision/style_transfer/fast_neural_style) for example , Candy model ([direct link](https://github.com/onnx/models/blob/master/vision/style_transfer/fast_neural_style/models/opset9/candy.tar.gz))  gives error - 

`cv2.error: OpenCV(4.2.0-dev) /home/usr/opencv/modules/dnn/src/dnn.cpp:4600: error: (-215:Assertion failed) inputs.size() in function 'getMemoryShapes'`

##### Steps to reproduce
```
import cv2
net = cv2.dnn.readNetFromONNX("/Path_to_model/candy.onnx")
```

ashishkrshrivastava:
resolves #16707 

**Merge with extra**: [opencv/opencv_extra#719](https://github.com/opencv/opencv_extra/pull/719)

`opencv_extra=cvonnx`

### This PR adds :
1. Gather-Cast subgraph fusion
2. Mul-Cast subgraph fusion

### Detailed Description
Different Pytorch versions create different subgraphs, according to what I have observed till now, the difference is only because of presence or absence of cast layer after Gather and Mul layer. The idea is to fuse Gather-Cast and Mul-Cast . Hence presence or absence of Cast layer wouldn't make any affect on loading of model.

Below attached, three different subgraphs obtained using same code but different pytorch versions. Producer of Fast style transfer models is Pytorch 1.1, which is middle subgraph in attached image . PR [#16573](https://github.com/opencv/opencv/pull/16573) added support for only first subgraph. This PR adds support for all three subgraphs. 

Screenshot from 2020-03-01 20-28-02

### Merging these changes 
would allow loading of all [Style trasfer models](https://github.com/onnx/models/tree/master/vision/style_transfer/fast_neural_style) of ONNX model Zoo which includes:
- Mosaic 
- Candy 
- Rain Princess 
- Udnie 
- Pointilism

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/graph_simplifier.cpp
+++ b/modules/dnn/src/graph_simplifier.cpp
@@ -194,15 +194,14 @@ void simplifySubgraphs(const Ptr<ImportGraphWrapper>& net,
 {
     int numNodes = net->getNumNodes();
     std::vector<int> matchedNodesIds, targetNodesIds;
-    for (int i = 0; i < numNodes; ++i)
+    for (int j = 0; j < patterns.size(); ++j)
     {
-        for (int j = 0; j < patterns.size(); ++j)
+        for (int i = 0; i < numNodes; ++i)
         {
             if (patterns[j]->match(net, i, matchedNodesIds, targetNodesIds))
             {
                 patterns[j]->replace(net, matchedNodesIds, targetNodesIds);
                 numNodes -= matchedNodesIds.size() - 1;  // #matchedNodes removed and one added.
-                break;
             }
         }
     }
--- a/modules/dnn/src/onnx/onnx_graph_simplifier.cpp
+++ b/modules/dnn/src/onnx/onnx_graph_simplifier.cpp
@@ -154,6 +154,32 @@ private:
     int axis;
 };
 
+class GatherCastSubgraph : public Subgraph
+{
+public:
+    GatherCastSubgraph()
+    {
+        int input = addNodeToMatch("");
+        int index = addNodeToMatch("Constant");
+        int gather = addNodeToMatch("Gather", input, index);
+        addNodeToMatch("Cast", gather);
+        setFusedNode("Gather", input, index);
+    }
+};
+
+class MulCastSubgraph : public Subgraph
+{
+public:
+    MulCastSubgraph()
+    {
+        int input = addNodeToMatch("");
+        int scaleNode = addNodeToMatch("Constant");
+        int mul = addNodeToMatch("Mul", input, scaleNode);
+        addNodeToMatch("Cast", mul);
+        setFusedNode("Mul", input, scaleNode);
+    }
+};
+
 class ExtractScalesSubgraph : public Subgraph
 {
 public:
@@ -164,20 +190,16 @@ public:
         int indexH = addNodeToMatch("Constant");
         int shape1 = addNodeToMatch("Shape", input);
         int gather1 = addNodeToMatch("Gather", shape1, indexH);
-        int castG1 = addNodeToMatch("Cast", gather1);
         scaleHNode = addNodeToMatch("Constant");
-        int mul1 = addNodeToMatch("Mul", castG1, scaleHNode);
-        int castM1 = addNodeToMatch("Cast", mul1);
-        int floor1 = addNodeToMatch("Floor", castM1);
+        int mul1 = addNodeToMatch("Mul", gather1, scaleHNode);
+        int floor1 = addNodeToMatch("Floor", mul1);
 
         int indexW = addNodeToMatch("Constant");
         int shape2 = addNodeToMatch("Shape", input);
         int gather2 = addNodeToMatch("Gather", shape2, indexW);
-        int castG2 = addNodeToMatch("Cast", gather2);
         scaleWNode = addNodeToMatch("Constant");
-        int mul2 = addNodeToMatch("Mul", castG2, scaleWNode);
-        int castM2 = addNodeToMatch("Cast", mul2);
-        int floor2 = addNodeToMatch("Floor", castM2);
+        int mul2 = addNodeToMatch("Mul", gather2, scaleWNode);
+        int floor2 = addNodeToMatch("Floor", mul2);
 
         int unsqueeze1 = addNodeToMatch("Unsqueeze", floor1);
         int unsqueeze2 = addNodeToMatch("Unsqueeze", floor2);
@@ -190,19 +212,23 @@ public:
     {
         opencv_onnx::NodeProto* constant_node = inputs[1].dynamicCast<ONNXNodeWrapper>()->node;
         opencv_onnx::TensorProto tensor_proto = constant_node->attribute(0).t();
-        float scaleW = getMatFromTensor(tensor_proto).at<float>(0);
+        Mat scaleW = getMatFromTensor(tensor_proto);
+        CV_Assert(scaleW.total() == 1);
+        scaleW.convertTo(scaleW, CV_32F);
 
         constant_node = inputs[2].dynamicCast<ONNXNodeWrapper>()->node;
         tensor_proto = constant_node->attribute(0).t();
-        float scaleH = getMatFromTensor(tensor_proto).at<float>(0);
+        Mat scaleH = getMatFromTensor(tensor_proto);
+        CV_Assert(scaleH.total() == 1);
+        scaleH.convertTo(scaleH, CV_32F);
 
         opencv_onnx::NodeProto* node = fusedNode.dynamicCast<ONNXNodeWrapper>()->node;
         opencv_onnx::AttributeProto* attrH = node->add_attribute();
         attrH->set_name("height_scale");
-        attrH->set_i(scaleH);
+        attrH->set_i(scaleH.at<float>(0));
         opencv_onnx::AttributeProto* attrW = node->add_attribute();
         attrW->set_name("width_scale");
-        attrW->set_i(scaleW);
+        attrW->set_i(scaleW.at<float>(0));
 
         node->mutable_input()->DeleteSubrange(1, 2);  // Remove two last inputs
     }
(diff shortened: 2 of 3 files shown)


modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -154,6 +154,32 @@ class SoftMaxSubgraph : public Subgraph
...
 
+class FuseGatherCastSubgraph : public Subgraph
+{
+public:
+    FuseGatherCastSubgraph()
+    {
+        int input = addNodeToMatch("");
+        int index = addNodeToMatch("Constant");
+        int gather = addNodeToMatch("Gather", input, index);
+        addNodeToMatch("Cast", gather);
```
dkurt: What if we will just add an entry for `Cast` layer at https://github.com/opencv/opencv/blob/master/modules/dnn/src/onnx/onnx_importer.cpp?

ashishkrshrivastava: But here how can we use that, I mean we are trying to fuse upsample/resize subgraph.

ashishkrshrivastava: Do you mean adding cast layer implementation in [elementwise_layer](https://github.com/opencv/opencv/blob/master/modules/dnn/src/layers/elementwise_layers.cpp) ?

dkurt: Got it, thanks!

So it still related to https://github.com/opencv/opencv/issues/16323 and https://github.com/opencv/opencv/issues/14713 which were not actually resolved by https://github.com/opencv/opencv/issues/16323.

I tried to open https://github.com/opencv/opencv/pull/16710 but it seems like it's better to have more generic solution for all three combinations, yes.

So if you don't mind, let's move a test from PyTorch 1.2. to your PR (see https://github.com/opencv/opencv_extra/pull/716).

I'll keep only Mul layer related changes.

modules/dnn/src/graph_simplifier.cpp
```diff
@@ -202,6 +202,7 @@ void simplifySubgraphs(const Ptr<ImportGraphWrapper>& net,
             {
                 patterns[j]->replace(net, matchedNodesIds, targetNodesIds);
                 numNodes -= matchedNodesIds.size() - 1;  // #matchedNodes removed and one added.
+                i -= matchedNodesIds.size() - 1;
```
dkurt: I don't think that it works as expected.

Am I right that after fusion of Gather+Cost or Mul+Const layers you want to run matching for Upsample layer? the problem is that it's patter start a bit earlier (includes at least Shape node) that means we cannot shift enough back to match Upsample after Gather.

So maybe we need to make the first loop over patterns but internal one by nodes? This way we can guaratnee that each following patter in our list will start from the beginning of graph. But we need to benchmark the performance of graphs fusion logic - how much it will grow up.

dkurt: Just tested two approaches for graphs matching on Mask R-CNN network (1401 nodes, 17 TensorFlow patterns to match)

1) current
```cpp
    for (int i = 0; i < numNodes; ++i)
    {
        for (int j = 0; j < patterns.size(); ++j)
        {
            if (patterns[j]->match(net, i, matchedNodesIds, targetNodesIds))
            {
                patterns[j]->replace(net, matchedNodesIds, targetNodesIds);
                numNodes -= matchedNodesIds.size() - 1;  // #matchedNodes removed and one added.
                break;
            }
        }
    }
```
21.3925ms

2) Modified:
```cpp
    for (int j = 0; j < patterns.size(); ++j)
    {
        for (int i = 0; i < numNodes; ++i)
        {
            if (patterns[j]->match(net, i, matchedNodesIds, targetNodesIds))
            {
                patterns[j]->replace(net, matchedNodesIds, targetNodesIds);
                numNodes -= matchedNodesIds.size() - 1;  // #matchedNodes removed and one added.
            }
        }
    }
```
22.6679ms

So I think you can freely try to use it for your experiment without any significant efficiency gaps.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -154,6 +154,32 @@ class SoftMaxSubgraph : public Subgraph
...
+        setFusedNode("Gather", input, index);
+    }
+};
+
+class FuseMulCastSubgraph : public Subgraph
```
dkurt: Just `MulCastSubgraph` would be OK

modules/dnn/src/graph_simplifier.cpp
ashishkrshrivastava: Below graph is not a part of any test, I have exported this graph just to show why I was adding ` i -= matchedNodesIds.size() - 1;`
Screenshot from 2020-03-02 08-17-33
Net nodes before graph simplification

```
Conv Constant Shape Gather Cast Constant Mul Cast Floor Constant Shape Gather Cast Constant Mul Cast Floor Unsqueeze Unsqueeze Concat Constant Cast Shape Slice Cast Div Concat Upsample 
Conv Constant Shape Gather Cast Constant Mul Cast Floor Constant Shape Gather Cast Constant Mul Cast Floor Unsqueeze Unsqueeze Concat Constant Cast Shape Slice Cast Div Concat Upsample
```
Net nodes before graph simplification

if `i -= matchedNodesIds.size() - 1;` is **not added** (not desired)
```
Conv Constant Constant Constant Constant Constant Upsample 
Conv Constant Shape Gather Cast Constant Mul Cast Floor Constant Shape Gather Cast Constant Mul Cast Floor Unsqueeze Unsqueeze Concat Constant Cast Shape Slice Cast Div Concat Upsample
```
so here, after fusion at first `upsample` node, nodes ids changed but` i` was still the same, hence it missed `Cast` after `Gather` and `Mul` nodes because ids of these nodes became less than `i` which in turn results in graph mismatch at second `upsample `node so it didn't fuse. 

if `i -= matchedNodesIds.size() - 1;` is **added** (desired)
```
Conv Constant Constant Constant Constant Constant Upsample 
Conv Constant Constant Constant Constant Constant Upsample
```

In my view, if graph would be like 
```
                                  | Unfused_upsample |
                                           |
                                  | Unfused_softmax |
```
then it would miss unfused_softmax fusion after unfused_upsample fusion if i is not changed. 
Please correct me if I misinterpreted something.

ashishkrshrivastava: Thanks, This modified version is working perfectly fine
```cpp
for (int j = 0; j < patterns.size(); ++j)
    {
        for (int i = 0; i < numNodes; ++i)
        {
            if (patterns[j]->match(net, i, matchedNodesIds, targetNodesIds))
            {
                patterns[j]->replace(net, matchedNodesIds, targetNodesIds);
                numNodes -= matchedNodesIds.size() - 1;  // #matchedNodes removed and one added.
            }
        }
    }
```

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: Sure, I will add the test. 
So you want me to remove Gather-Cast related changes and only keep Mul-Cast related changes ?

ashishkrshrivastava: yeah, you are right, [#16323](https://github.com/opencv/opencv/issues/16323) and [#14713](https://github.com/opencv/opencv/issues/14713) were not completely resolved by [#16573](https://github.com/opencv/opencv/pull/16573) Those models would have worked if exported using recent Pytorch version. Actually, i thought that PR was getting complex so didn't add these modifications there.
So I intend to make those changes generic for all pytorch versions with this PR.

modules/dnn/src/graph_simplifier.cpp
dkurt: So will you consider this change in this PR? I believe that this way we can avoid `i -= matchedNodesIds.size() - 1;`

ashishkrshrivastava: yeah, I will add this change here.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: So should I remove changes related to `GatherCastSubgraph `and keep changes for `MulCastSubgraph` ?

ashishkrshrivastava:
@dkurt,  So you want me to remove Gather-Cast related changes and only keep Mul-Cast related changes ?

modules/dnn/src/graph_simplifier.cpp
ashishkrshrivastava: > So if you don't mind, let's move a test from PyTorch 1.2. to your PR (see opencv/opencv_extra#716).

@dkurt, here you wanted me to Push a PR to Opencv_extra adding your suggested change to that new PR?

dkurt:
@ashishkrshrivastava, no, please keep both Gather-Cast and Mul-Cast but remove `i -= matchedNodesIds.size() - 1;` from the core of graphs fusion logic.

modules/dnn/src/graph_simplifier.cpp
dkurt: Yes, they should be bonded, thanks!

dkurt:
Please resolve merge conflicts (rebase with updated 3.4)

ashishkrshrivastava:
>  Thank you!

Thank you so much :+1:

PR #16740, opened 2020-03-05
Fix building with ToT libc++

m-gupta:
ToT libc++ (LLVM) no longer includes `<sstream>`
as part of `<complex>` which breaks building opencv.
Include `<sstream>` header explcitly to fix this.

<cut/>


m-gupta:
Done, please check.

asmorkalov:
:+1:

PR #16744, opened 2020-03-06
core: don't allow reallocation in add/div/sub/bitwise aug operators

alalek:
example: https://github.com/opencv/opencv/issues/9688#issuecomment-331400000

Operators with "inplace" arguments should not re-allocate output arrays.


LaurentBerger:
read https://github.com/opencv/opencv/issues/9688#issuecomment-596205662

alalek:
:+1:

PR #16751, opened 2020-03-06
core: fix coverity issues

alalek:
- extra assertion checks
- "specialize" generic norm implementation through simple loop


alalek:
:+1:

PR #16759, opened 2020-03-07
Add getter for mCamera in JavaCameraView class

StidOfficial:

- [X] There is reference to original bug report and related work
- [X] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
- [X] The feature is well documented and sample code can be built with the project CMake

Example usage :
```
private JavaCameraView cameraView;
...
Camera.Parameters params = cameraView.getCamera().getParameters();
params.setFlashMode(Camera.Parameters.FLASH_MODE_TORCH);
cameraView.getCamera().setParameters(params);
```

### DIFF
--- a/modules/java/generator/android/java/org/opencv/android/JavaCameraView.java
+++ b/modules/java/generator/android/java/org/opencv/android/JavaCameraView.java
@@ -65,6 +65,10 @@ public class JavaCameraView extends CameraBridgeViewBase implements PreviewCallb
     public JavaCameraView(Context context, AttributeSet attrs) {
         super(context, attrs);
     }
+    
+    public Camera getCamera() {
+        return mCamera;
+    }
 
     protected boolean initializeCamera(int width, int height) {
         Log.d(TAG, "Initialize java camera");


asmorkalov:
Could you provide some reason why you want to expose camera object? The protect ed camera object guarantees that application has predictable lifecycle and camera settings are done in proper state. You can use inheritance and implement own class for specific behaviour, there is no need to modify original code.

StidOfficial:
It's more complex to create a extends class :
```
public class JavaCameraViewAdvanced extends JavaCameraView {

    public JavaCameraViewAdvanced(Context context, int cameraId) {
        super(context, cameraId);
    }

    public JavaCameraViewAdvanced(Context context, AttributeSet attrs) {
        super(context, attrs);
    }

    public void enableTorch() {
        Camera.Parameters parameters = mCamera.getParameters();
        parameters.setFlashMode(Camera.Parameters.FLASH_MODE_TORCH);
        mCamera.setParameters(parameters);
    }

    public void disableTorch() {
        Camera.Parameters parameters = mCamera.getParameters();
        parameters.setFlashMode(Camera.Parameters.FLASH_MODE_OFF);
        mCamera.setParameters(parameters);
    }
}
```
Remplace JavaCameraView component in xml layout and use it in normal code juste for enable/disable flash torch.
Than add getter and use it in normal code.

asmorkalov:
Theoretically yes, but in practice even with flash you need to check  flash mode availability (there is no flash for frontal camera), handle view events (turn it of, when camera view is closed), etc. So the amount of extra code lines for real use case is very small.

asmorkalov:
@alalek What do you think about the patch?

asmorkalov:
The solution was discussed with @alalek @VadimLevin @vpisarev and the team decided to reject the patch. There are several reasons for that:
- It's not safe to expose camera for manipulations, because particular camera options should be sent during particular camera stages. User has to handle the events and inheritance is treated as better solution, see discussion above.
- There are `JavaCamera2View` and it should be consistent.

PR #16766, opened 2020-03-10
feature: Extend VideoWriter to accept vector of parameters

VadimLevin:
- Added additional constructor and `open` method for `VideoWriter`
   those accept a vector of parameters
 - Moved the actual implementation of the `VideoWriter::open` to the general method
   which accepts a vector of parameters
 - Propagate parsed parameters map up to actual video backend construction

```
build_image:Custom=centos:7
buildworker:Custom=linux-1

force_builders=Custom Win
```

### DIFF
--- a/modules/videoio/include/opencv2/videoio.hpp
+++ b/modules/videoio/include/opencv2/videoio.hpp
@@ -191,7 +191,7 @@ enum VideoWriterProperties {
   VIDEOWRITER_PROP_FRAMEBYTES = 2, //!< (Read-only): Size of just encoded video frame. Note that the encoding order may be different from representation order.
   VIDEOWRITER_PROP_NSTRIPES = 3,   //!< Number of stripes for parallel encoding. -1 for auto detection.
   VIDEOWRITER_PROP_IS_COLOR = 4    //!< If it is not zero, the encoder will expect and encode color frames, otherwise it
-                                   //!< will work with grayscale frames (the flag is ignored on MSMF and Intel MFX backends)
+                                   //!< will work with grayscale frames.
 };
 
 //! @} videoio_flags_base
@@ -878,7 +878,7 @@ public:
     @param fps Framerate of the created video stream.
     @param frameSize Size of the video frames.
     @param isColor If it is not zero, the encoder will expect and encode color frames, otherwise it
-    will work with grayscale frames (the flag is ignored on MSMF and Intel MFX backends).
+    will work with grayscale frames.
 
     @b Tips:
     - With some backends `fourcc=-1` pops up the codec selection dialog from the system.
--- a/modules/videoio/include/opencv2/videoio.hpp
+++ b/modules/videoio/include/opencv2/videoio.hpp
@@ -191,7 +191,7 @@ enum VideoWriterProperties {
   VIDEOWRITER_PROP_FRAMEBYTES = 2, //!< (Read-only): Size of just encoded video frame. Note that the encoding order may be different from representation order.
   VIDEOWRITER_PROP_NSTRIPES = 3,   //!< Number of stripes for parallel encoding. -1 for auto detection.
   VIDEOWRITER_PROP_IS_COLOR = 4    //!< If it is not zero, the encoder will expect and encode color frames, otherwise it
-                                   //!< will work with grayscale frames (the flag is currently supported on Windows only)
+                                   //!< will work with grayscale frames (the flag is ignored on MSMF and Intel MFX backends)
 };
 
 //! @} videoio_flags_base
@@ -878,7 +878,7 @@ public:
     @param fps Framerate of the created video stream.
     @param frameSize Size of the video frames.
     @param isColor If it is not zero, the encoder will expect and encode color frames, otherwise it
-    will work with grayscale frames (the flag is currently supported on Windows only).
+    will work with grayscale frames (the flag is ignored on MSMF and Intel MFX backends).
 
     @b Tips:
     - With some backends `fourcc=-1` pops up the codec selection dialog from the system.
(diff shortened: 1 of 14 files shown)


modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -894,6 +896,19 @@ class CV_EXPORTS_W VideoWriter
     CV_WRAP VideoWriter(const String& filename, int apiPreference, int fourcc, double fps,
                 Size frameSize, bool isColor = true);
 
+    /** @overload
+     *
+     * @param params Encoder parameters encoded as pairs (paramId_1, paramValue_1, paramId_2, paramValue_2, ... .)
```
Kaustubh72: * @param params Encoder parameters encoded as pairs (paramId_1, paramValue_1, paramId_2, paramValue_2,paramId_3, paramValue_3,... )

asmorkalov:
@vpisarev @alalek  Can we merge the patch?

modules/videoio/src/cap.cpp
```diff
@@ -628,4 +658,37 @@ int VideoWriter::fourcc(char c1, char c2, char c3, char c4)
     return (c1 & 255) + ((c2 & 255) << 8) + ((c3 & 255) << 16) + ((c4 & 255) << 24);
 }
 
+namespace detail {
+static void fillRequiredVideoWriterParametersWithDefaults(VideoWriterParameters& parameters)
```
alalek, line 662: Not sure that this is correct approach.

Backends may have own specific parameters.
So one function is not enough.
Also it is not enough one such function per backend - for example, file container / codec tuning parameters.

So it is better to replace this by stuff by similar to Python `.get(key, default_value)` method:
```
int get(std::map<int, int>& map, int key, const int defaultValue);

// or more generic (not needed for now)
template<...>
getKeyOrDefault(std::map<TKey, TValue>& map, const TKey& key, const TValue defaultValue)
```

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -188,7 +188,9 @@ enum VideoCaptureProperties {
 enum VideoWriterProperties {
   VIDEOWRITER_PROP_QUALITY = 1,    //!< Current quality (0..100%) of the encoded videostream. Can be adjusted dynamically in some codecs.
   VIDEOWRITER_PROP_FRAMEBYTES = 2, //!< (Read-only): Size of just encoded video frame. Note that the encoding order may be different from representation order.
-  VIDEOWRITER_PROP_NSTRIPES = 3    //!< Number of stripes for parallel encoding. -1 for auto detection.
+  VIDEOWRITER_PROP_NSTRIPES = 3,   //!< Number of stripes for parallel encoding. -1 for auto detection.
+  VIDEOWRITER_PROP_IS_COLOR = 4    //!< If it is not zero, the encoder will expect and encode color frames, otherwise it
+                                   //!< will work with grayscale frames (the flag is currently supported on Windows only)
```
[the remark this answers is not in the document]
VadimLevin: This statement comes from the `isColor` parameter description

modules/videoio/src/cap.cpp
VadimLevin: The goal of this function is to be a configuration point for required *common* parameters, e.g. `isColor`, to share 1 default value for various backends.

I like your way of getting codec/backend specific parameters and will use it in further.

Maybe better name (e.g. `fillCommonRequiredWriterParameters`) should remove such confusion or it's better to remove it anyway?

modules/videoio/include/opencv2/videoio.hpp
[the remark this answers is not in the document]
VadimLevin: I rechecked `isColor` parameter usage. It is completely ignored on MSMF and Intel MFX backends, so I updated the description.

modules/videoio/include/opencv2/videoio.hpp
```diff
@@ -876,7 +878,7 @@ class CV_EXPORTS_W VideoWriter
     @param fps Framerate of the created video stream.
     @param frameSize Size of the video frames.
     @param isColor If it is not zero, the encoder will expect and encode color frames, otherwise it
-    will work with grayscale frames (the flag is currently supported on Windows only).
+    will work with grayscale frames (the flag is ignored on MSMF and Intel MFX backends).
```
alalek, line 881: > the flag is ignored

It is better to keep here common information only.
Backend-specific information should be moved on dedicated pages (there is huge amount of implementation details)

VadimLevin: Does `(the flag may be ignored by some backends. For more information refer to backend implementation)` sound better?

alalek: I would prefer to keep this info in a single place (to avoid manual maintenance in different places).

See here:
https://github.com/opencv/opencv/blame/4.3.0/modules/videoio/doc/videoio_overview.markdown#L27

I believe the header of properties enumeration should be enough.

VadimLevin: Thank you, got it!

jomade:
Hi! 
With these changes - is it really possible to set VIDEOWRITER_PROP_IS_COLOR =0? I mean, the parameter list accepts only the "identifier" VIDEOWRITER_PROP_IS_COLOR but no value to that property. I think one should integrate a VIDEOWRITER_PROP_IS_GRAYSCALE or sth like that to be able to actually use this flag.

alalek:
Enum's documentation clearly states which values are accepted.
Passed values are pairs of `int`s.
Also check implementation of `bool VideoWriter::open` on this PR's diff.

jomade:
Sorry, I didn't check the first doxygen-comment of the overloaded function. And from the API with `const std::vector<int>&` params it was not clear to me that this would be pairs. I would have expected sth like `const std::vector<std::pair<int, int>>`. Sorry for the noise.

asmorkalov:
@jomade The std::verctor<int> can be easily wrapped to Python, Java and other supported language bindings. It's the main reson why, it's not vector of pair, map or other structures.

Issue #16339, opened 2020-01-13
Julia Bindings for OpenCV


archit120:
## Bindings and Tests for the Julia Language

* Author: Archit Rungta / @archit120
* Link: [Raised Issue](https://github.com/opencv/opencv/issues/16233)
* Status: **Draft** 
* Platforms: **All** 
* Complexity: 2-3 Man-Months

## Introduction and Rationale

Julia is a popular dynamic language used for scientific and numerical computing, with performance comparable to traditional statically-typed languages. It has powerful libraries for many different tasks such as FluxMl and Zygote. However, the JuliaImages library lacks a lot of functionality. 

There are two existing OpenCV wrappers: https://github.com/JuliaOpenCV and https://github.com/maxruby/OpenCV.jl. Both of these do not work with current version of Julia and have been abandoned for multiple years now. The only way of using OpenCV from Julia right now is to call the Python bindings for OpenCV using PyCall.jl

This is aimed at creating bindings for Julia which are similar in scope and ease-of-use to OpenCV's Python bindings. Furthermore, test files similar to Python should be created.

## Proposed solution

The solution is to build wrapping for Julia similar to how they are constructed for Python. Ref [1]. Julia has a library CxxWrap.jl which provides the needed functionality of creating Julia wrappers in C++

This wrapper will then call the C++ OpenCV code. All functionality of OpenCV would then be exposed to be called from Julia code.

## Impact on existing code, compatibility

The changes will not affect any other part of OpenCV's functionality. A seperate folder for Julia bindings will contain all the related code. Only a flag for make configuration has to be added which can be off by default.

## Possible alternatives

Julia supports an alternative form of wrapping using Cxx.jl Unlike CxxWrap this requires wrapper code to be written in Julia. Another alternative is to build the Julia bindings on top of Python. This method, however, is not as preferable because of extra dependencies and overheads. 

## References

1. [Description of How Python Bindings Work](https://docs.opencv.org/4.1.0/da/d49/tutorial_py_bindings_basics.html)
2. [Some Work on CxxWrap based binding](https://github.com/clynamen/opencv)

archit120:
I want to work on this for GSoC 2020.

nmaxwell:
You could also consider using the xtensor family [1-5]. The basic idea would be to write an xtensor-based C++ wrapper around opencv, then get python, Julia, and R bindings, almost for free. If/when the xtensor folks support Matlab bindings, you would then get those too.

[...]

PR #16767, opened 2020-03-10
Add some docstrings to gen2.py

archit120:
Some documentation for python bindings generator file gen2.py. 

Original issue: https://github.com/opencv/opencv/issues/16339

### DIFF
--- a/modules/python/src2/gen2.py
+++ b/modules/python/src2/gen2.py
@@ -87,6 +87,9 @@ template<> bool pyopencv_to(PyObject* src, ${cname}& dst, const ArgInfo& info);
 
 """)
 
+"""
+    Template to add code for mapping fields in CV_EXPORTS_W_MAP classes
+"""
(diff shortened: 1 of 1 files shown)


VadimLevin:
@archit120 Thank you for the contribution! I think docstring should contain an example of usage describing expected behavior like it is done for libraries like NumPy or SciPy. 
This info not only demonstrates how this function can be called but can be used with [pydoctest](https://docs.python.org/3.6/library/doctest.html) or be converted to actual test cases.
Can you update your docs with it and add appropriate tests for python bindings generator?

asmorkalov:
@archit120 The patch was discussed on OpenCV team weekly meeting and @alalek @asmorkalov @VadimLevin @vladimir-dudnik @vpisarev decided to reject it because it brings small additional value and introduce extra conflicts during 3.4 merge to master.

PR #16770, opened 2020-03-10
pre: OpenCV 3.4.10 (version++)

alalek:
- Android Manager internal version is not increased (stuck on 3.49)

previous PR with version bump: #16071


alalek:
:+1:

PR #16772, opened 2020-03-10
Added g-api render performance tests

aDanPin:
### Added:

 Performance tests render G-API

### Description:

G-API render perforamance tests. These tests are very similar to G-API unit render tests.

 - [ ] Patch to opencv_extra has the same branch name.

```
force_builders=Custom,ARMv7

Xbuild_image:Custom=centos:7
build_image:Custom=ubuntu-gapi-freetype:16.04
buildworker:Custom=linux-1
```


aDanPin:
@TolyaTalamanov

modules/gapi/perf/common/gapi_render_perf_tests.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+  //------------------------------------------------------------------------------
+
+    class RenderTestTexts : public TestPerfParams<tuple<compare_f, MatType, cv::Size, cv::GCompileArgs>> {};
+    class RenderTestRects : public TestPerfParams<tuple<compare_f, MatType, cv::Size, cv::GCompileArgs>> {};
+    class RenderTestCircles : public TestPerfParams<tuple<compare_f, MatType, cv::Size, cv::GCompileArgs>> {};
+    class RenderTestLines : public TestPerfParams<tuple<compare_f, MatType, cv::Size, cv::GCompileArgs>> {};
+    class RenderTestMosaics : public TestPerfParams<tuple<compare_f, MatType, cv::Size, cv::GCompileArgs>> {};
+    class RenderTestImages : public TestPerfParams<tuple<compare_f, MatType, cv::Size, cv::GCompileArgs>> {};
+    class RenderTestPolylines : public TestPerfParams<tuple<compare_f, MatType, cv::Size, cv::GCompileArgs>> {};
```
TolyaTalamanov: Could you add to test with multiple primitives.
For example:
n - `Rect`
n - `Mosaic` inside of `Rects`
m - `Text`
k - `Image`

n, m, k - number of each primitive (`Hint`: you can make it as a test parameters )

modules/gapi/perf/common/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,918 @@
+// This file is part of OpenCV project.
```
TolyaTalamanov: You don't need this file in the `common` part, because it can only be used by one backend
You can put the test definition and its instance together in the same file (as for unit tests)

modules/gapi/perf/common/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,918 @@
...
+
+    cv::Scalar cvtBGRToYUVC(const cv::Scalar& bgr)
+    {
+        double y = bgr[2] *  0.299000 + bgr[1] *  0.587000 + bgr[0] *  0.114000;
+        double u = bgr[2] * -0.168736 + bgr[1] * -0.331264 + bgr[0] *  0.500000 + 128;
+        double v = bgr[2] *  0.500000 + bgr[1] * -0.418688 + bgr[0] * -0.081312 + 128;
+        return {y, u, v};
+    }
+
+    void drawMosaicRef(const cv::Mat& mat, const cv::Rect &rect, int cellSz)
```
TolyaTalamanov: Put a comment about code duplication here

modules/gapi/perf/common/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,918 @@
...
+            }
+            if (has_crop_x)
+            {
+                cell_roi = msc_roi(cv::Rect(cols, rows, msc_roi.cols - cols, msc_roi.rows - rows));
+                cell_roi = cv::mean(cell_roi);
+            }
+        }
+    }
+
+    void blendImageRef(cv::Mat& mat, const cv::Point& org, const cv::Mat& img, const cv::Mat& alpha)
```
TolyaTalamanov: And here

modules/gapi/perf/common/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,918 @@
...
+        roi32f += img32f;
+
+        roi32f.convertTo(roi, CV_8U, 255.0);
+    }
+
+  }
+
+PERF_TEST_P_(RenderTestTexts, RenderTextsPerformanceBGROCVTest)
+{
+    std::string text("SomeText");
```
TolyaTalamanov: I think this should be a test parameter

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "../common/gapi_render_perf_tests.hpp"
+#include <opencv2/gapi/render.hpp>
+
+#define RENDER_CPU cv::gapi::render::ocv::kernels()
+
+namespace opencv_test
+{
+
+
+INSTANTIATE_TEST_CASE_P(RenderTestTextsCPU_8U, RenderTestTexts,
```
TolyaTalamanov: What about NV12 instances ?

TolyaTalamanov: Why is called `*_8U` ?

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
```diff
@@ -0,0 +1,59 @@
...
+    Combine(Values(AbsExact().to_compare_obj()),
+        Values(CV_8UC1, CV_8UC3),
+        Values(szVGA, sz720p, sz1080p, cv::Size(12, 12)),
+        Values(cv::compile_args(RENDER_CPU))));
+
+INSTANTIATE_TEST_CASE_P(RenderTestRectsCPU_8U, RenderTestRects,
+    Combine(Values(AbsExact().to_compare_f()),
+        Values(CV_8UC1, CV_8UC3),
```
TolyaTalamanov: We use only `RGB` and `NV12` formats for `render`
So there is no need to pass the matrix type here

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
```diff
@@ -0,0 +1,59 @@
...
+        Values(szVGA, sz720p, sz1080p),
+        Values(cv::compile_args(RENDER_CPU))));
+
+INSTANTIATE_TEST_CASE_P(RenderTestCirclesCPU_8U, RenderTestCircles,
+    Combine(Values(AbsExact().to_compare_f()),
```
TolyaTalamanov: I think the comparator is also unnecessary here we don 't use it in unit tests

dmatveev:
Please ping me here once Anatoliy completes the review.

P.S. An MR description is missing here.

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
```diff
@@ -0,0 +1,59 @@
...
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+
+#include "../perf_precomp.hpp"
+#include "../common/gapi_render_perf_tests.hpp"
+#include <opencv2/gapi/render.hpp>
+
+#define RENDER_CPU cv::gapi::render::ocv::kernels()
```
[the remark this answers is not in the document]
aDanPin: Fixed

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace opencv_test
+{
+
+
+INSTANTIATE_TEST_CASE_P(RenderTestTextsCPU_8U, RenderTestTexts,
+    Combine(Values(AbsExact().to_compare_obj()),
+        Values(CV_8UC1, CV_8UC3),
+        Values(szVGA, sz720p, sz1080p, cv::Size(12, 12)),
+        Values(cv::compile_args(RENDER_CPU))));
```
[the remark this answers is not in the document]
aDanPin: Fixed

modules/gapi/perf/common/gapi_render_perf_tests.hpp
aDanPin: I did this test, but now it is red in the case of a 640x480 matrix. I'm working on it.
But you can look at this test, did I understand you correctly?

modules/gapi/perf/common/gapi_render_perf_tests_inl.hpp
aDanPin: Done

modules/gapi/perf/common/gapi_render_perf_tests_inl.hpp
aDanPin: Done

modules/gapi/perf/common/gapi_render_perf_tests_inl.hpp
aDanPin: Done

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
aDanPin: We can use a common instance for NV12 and BGR tests cases, aren't we?

aDanPin: Fixed

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
aDanPin: Done

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
aDanPin: I got you. But now the comparator is the second parameter that we passe to the tests. I failed to create tests with one parameter. So I left the comparator as a test parameter. Is it ok?

dmatveev:
Where is da MR description?

aDanPin:
@TolyaTalamanov

aDanPin:
@TolyaTalamanov

modules/gapi/perf/common/gapi_render_perf_tests.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <string>
+
+namespace opencv_test
+{
+
+  using namespace perf;
+
+  //------------------------------------------------------------------------------
+
+    class RenderTestTexts : public TestPerfParams<tuple<std::string, cv::Size, compare_f>> {};
```
TolyaTalamanov: Alignment is broken

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1018 @@
...
+  INSTANTIATE_TEST_CASE_P(RenderTestPolyItems, RenderTestPolyItems,
+      Combine(Values(szVGA, sz720p, sz1080p),
+              Values(50),
+              Values(50),
+              Values(50),
+              Values(AbsExact().to_compare_f())));
+
+PERF_TEST_P_(RenderTestTexts, RenderTextsPerformanceBGROCVTest)
+{
+    cv::Point org(200, 200);
```
TolyaTalamanov: All parameters should be as tests parameters

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1018 @@
...
+    {
+        EXPECT_TRUE(cmpF(gapi_mat, ref_mat));
+    }
+
+    SANITY_CHECK_NOTHING();
+}
+
+PERF_TEST_P_(RenderTestTexts, RenderTextsPerformanceNV12OCVTest)
+{
+    cv::Point org(200, 200);
```
TolyaTalamanov: And here

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1018 @@
...
+        EXPECT_TRUE(cmpF(y_out_gapi_mat,  y_ref_mat));
+        EXPECT_TRUE(cmpF(uv_out_gapi_mat, uv_ref_mat));
+    }
+
+    SANITY_CHECK_NOTHING();
+}
+
+PERF_TEST_P_(RenderTestRects, RenderRectsPerformanceBGROCVTest)
+{
+    cv::Rect rect(100, 100, 200, 200);
```
TolyaTalamanov: And here

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1018 @@
...
+    {
+        EXPECT_TRUE(cmpF(gapi_mat, ref_mat));
+    }
+
+    SANITY_CHECK_NOTHING();
+}
+
+PERF_TEST_P_(RenderTestRects, RenderRectsPerformanceNV12OCVTest)
+{
+    cv::Rect rect(100, 100, 200, 200);
```
TolyaTalamanov: And here

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1018 @@
...
+        EXPECT_TRUE(cmpF(y_out_gapi_mat,  y_ref_mat));
+        EXPECT_TRUE(cmpF(uv_out_gapi_mat, uv_ref_mat));
+    }
+
+    SANITY_CHECK_NOTHING();
+}
+
+PERF_TEST_P_(RenderTestCircles, RenderCirclesPerformanceBGROCVTest)
+{
+    cv::Point center(cv::Point(100, 100));
```
TolyaTalamanov: And here

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1018 @@
...
+        EXPECT_TRUE(cmpF(gapi_mat, ref_mat));
+    }
+
+    SANITY_CHECK_NOTHING();
+}
+
+
+PERF_TEST_P_(RenderTestCircles, RenderCirclesPerformanceNV12OCVTest)
+{
+    cv::Point center(cv::Point(100, 100));
```
TolyaTalamanov: Here and everything below

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1018 @@
...
+
+        roi32f.convertTo(roi, CV_8U, 255.0);
+    }
+
+  }
+
+  INSTANTIATE_TEST_CASE_P(RenderTestTexts, RenderTestTexts,
+      Combine(Values(std::string("Some text")),
+              Values(szVGA, sz720p, sz1080p),
+              Values(AbsExact().to_compare_f())));
```
TolyaTalamanov: Remove comparator from here

modules/gapi/perf/common/gapi_render_perf_tests.hpp
```diff
@@ -0,0 +1,36 @@
...
+    class RenderTestRects : public TestPerfParams<tuple<cv::Size, compare_f>> {};
```
TolyaTalamanov: Remove compare_f from here

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,9 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+
+#include "../perf_precomp.hpp"
```
TolyaTalamanov: I think, there isn't any reason for create seperate *.hpp file for render perfomance tests

TolyaTalamanov: You can move to *.cpp file

dmatveev:
@aDanPin when you're ready to push an update?

modules/gapi/perf/common/gapi_render_perf_tests.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <opencv2/gapi/imgproc.hpp>
+
+#include <string>
+
+namespace opencv_test
+{
+
+  using namespace perf;
+
+  //------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
aDanPin: Removed

modules/gapi/perf/common/gapi_render_perf_tests.hpp
aDanPin: Is now it ok?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Fixed

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Fixed

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Fixed

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Fixed

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Fixed

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Fixed

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Removed

modules/gapi/perf/common/gapi_render_perf_tests.hpp
aDanPin: Removed

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: I think I did not understand you correctly
Could you clarify your point

modules/gapi/perf/common/gapi_render_perf_tests.hpp
aDanPin: This test is done now. 

But!
Test framework allows to specify only 10 test parameters
Therefore, I cannot specify all the parameters for the image, text, mosaic, rectangle in the test instance.
All these parameters are announced in the body of the test.

aDanPin:
@TolyaTalamanov

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1144 @@
...
+#include "../common/gapi_render_perf_tests.hpp"
+
+#include "../../src/api/render_priv.hpp"
+
+namespace opencv_test
+{
+  using namespace perf;
+
+  namespace {
+    void create_rand_mats(const cv::Size& size, MatType type, cv::Mat& ref_mat, cv::Mat& gapi_mat) {
```
alalek, line 22: Avoid indentation in namespaces.
```
namespace abc {

...
...
...

}  // namespace
```

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1144 @@
...
+
+        // YUV -> NV12
+        cv::gapi::wip::draw::cvtYUVToNV12(yuv, y_ref_mat, uv_ref_mat);
+    }
+
+    // Comparison //////////////////////////////////////////////////////////////
+    {
+        EXPECT_EQ(0, cv::norm(y_out_gapi_mat,  y_ref_mat));
+        EXPECT_EQ(0, cv::norm(uv_out_gapi_mat, uv_ref_mat));
+    }
```
alalek, line 837: It is bad idea to check accuracy in performance tests.
Minimal checks should be applied, like result is not empty and/or is not black(zeroed) completely.

asmorkalov:
@TolyaTalamanov @dmatveev What is the PR status? Can it be merged?

modules/gapi/perf/common/gapi_render_perf_tests.cpp
TolyaTalamanov: You can move all stuff from this file to corresponding cpp file

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1144 @@
...
+  using namespace perf;
+
+  namespace {
+    void create_rand_mats(const cv::Size& size, MatType type, cv::Mat& ref_mat, cv::Mat& gapi_mat) {
+        ref_mat.create(size, type);
+        cv::randu(ref_mat, cv::Scalar::all(0), cv::Scalar::all(255));
+        ref_mat.copyTo(gapi_mat);
+    };
+
+    // TODO: remove this code duplication with /gapi/test/common/gapi_render_tests.hpp
```
TolyaTalamanov, line 28: I think it's better to use `FIXME` instead of `TODO`

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
TolyaTalamanov: So, we can remove all reference code from here, because we check accuracy via unit tests. 
@dmatveev What do you think ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1144 @@
...
+
+PERF_TEST_P_(RenderTestPolylines, RenderPolylinesPerformanceBGROCVTest)
+{
+    std::vector<cv::Point> points;
+    cv::Scalar color;
+    int thick;
+    int lt;
+    int shift;
+    MatType type;
+
```
TolyaTalamanov, line 850: Maybe remove this space here and put it line below ? It would be better I think

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1144 @@
...
+    prims.emplace_back(cv::gapi::wip::draw::Poly{points, color, thick, lt, shift});
+
+    cv::GMat in;
+    cv::GArray<cv::gapi::wip::draw::Prim> arr;
+
+    cv::GComputation comp(cv::GIn(in, arr),
+                          cv::GOut(cv::gapi::wip::draw::render3ch(in, arr)));
+
+    // Warm-up graph engine:
+    comp.apply(gin(gapi_mat, prims), gout(gapi_mat));
```
TolyaTalamanov, line 868: BTW,  `Computation` have `compile` method for warmp-up, don't it ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1144 @@
...
+PERF_TEST_P_(RenderTestPolyItems, RenderPolylinesPerformanceBGROCVTest)
+{
+    int thick = 2;
+    int lt = LINE_8;
+    cv::Scalar color(100, 50, 150);
+    MatType type = CV_8UC3;
+
+    cv::Size sz;
+    int n, m, k;
+    std::tie(sz, n, m, k) = GetParam();
```
TolyaTalamanov, line 960: I hope you can make more clearly names for this

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1144 @@
...
+
+  namespace {
+    void create_rand_mats(const cv::Size& size, MatType type, cv::Mat& ref_mat, cv::Mat& gapi_mat) {
+        ref_mat.create(size, type);
+        cv::randu(ref_mat, cv::Scalar::all(0), cv::Scalar::all(255));
+        ref_mat.copyTo(gapi_mat);
+    };
+
+    // TODO: remove this code duplication with /gapi/test/common/gapi_render_tests.hpp
+    cv::Scalar cvtBGRToYUVC(const cv::Scalar& bgr)
```
TolyaTalamanov, line 29: What about `FText` primitve ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
TolyaTalamanov, line 22: Can we reuse function from `TestPerfParams` to init mats ?

asmorkalov:
@aDanPin your turn.

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1235 @@
...
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
+    {
+        cv::putText(ref_mat, text, org, ff, fs, color, thick, lt, blo);
+    }
+
+    // Comparison //////////////////////////////////////////////////////////////
+    {
+        EXPECT_EQ(0, cv::norm(gapi_mat, ref_mat));
```
alalek, line 326: Force using of `NORM_INF` in bit-exact checks for better diagnostic messages.

Anyway, [this item](https://github.com/opencv/opencv/pull/16772#discussion_r404431219) is still open about accuracy checks in performance tests.

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
dmatveev: Actually we have followed @elatkin's vast experience in this question. We need to make sure the result is still valid when optimizing our code

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Done

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Fixed

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: In this case these are similar things
Internally, a `cv::GCompiled` object is created for the given input format configuration, which then is executed on the input data immediately. `cv::GComputation` caches compiled objects produced within `apply()` -- if this method would be called next time with the same input parameters (image formats, image resolution, etc), the underlying compiled graph will be reused without recompilation. If new metadata doesn't match the cached one, the underlying compiled graph is regenerated.

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Fixed

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Fixed

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
aDanPin: Yes we can
But in this case we will have to use mats named in_mat1 and in_mat2 instead y_in_gapi_mat y_out_gapi_mat
I think better way to use mats named y_in_gapi_mat y_out_gapi_mat for better code understanding

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Is now its ok?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
dmatveev: `compile()` is not a warm-up.

"Warm-up" assumes a full run so image buffers which are allocated are actually allocated and are written to. On some systems (e.g. Linux) the memory allocation is lazy and happens on actual access but not the `new`/`malloc` thing itself.

TolyaTalamanov: Got it, thanks!

asmorkalov:
@aDanPin friendly reminder.

aDanPin:
@TolyaTalamanov , please review

asmorkalov:
@TolyaTalamanov @dmatveev Friendly reminder.

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1238 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+#include "../perf_precomp.hpp"
+
+#include <iostream>
```
TolyaTalamanov, line 9: I'm not sure that it's really should be here

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1238 @@
...
+
+    cv::multiply(img32f, img32f_w, img32f);
+    cv::multiply(roi32f, roi32f_w, roi32f);
+    roi32f += img32f;
+
+    roi32f.convertTo(roi, CV_8U, 255.0);
+}
+
+#ifdef HAVE_FREETYPE
+static std::wstring to_wstring(const char* bytes)
```
TolyaTalamanov, line 118: Put comment about code duplication

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1238 @@
...
+    std::wstring text;
+    cv::Size sz;
+    cv::Point org;
+    int fh;
+    cv::Scalar color;
+    MatType type;
+
+    std::tie(text ,sz ,org ,fh ,color, type) = GetParam();
+
+    cv::Mat gapi_mat = cv::Mat(sz, type);
```
TolyaTalamanov, line 221: Why don't you use `create_rand_mats` here

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1238 @@
...
+    int ff, thick, lt;
+    double fs;
+    cv::Scalar color;
+    bool blo;
+    MatType type;
+    std::string text;
+    cv::Size sz;
+    std::tie(text, sz, org, ff, fs, color, thick, lt, blo, type) = GetParam();
+
+    cv::Mat gapi_mat = cv::Mat(sz, type);
```
TolyaTalamanov, line 300: I should use `create_rand_mats` for every test cases not only for `NV12`

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1238 @@
...
+    cv::Point org;
+    int fh;
+    cv::Scalar color;
+    MatType type;
+
+    std::tie(text ,sz ,org ,fh ,color, type) = GetParam();
+
+    cv::Mat y_ref_mat, uv_ref_mat;
+
+    cv::Mat y_in_gapi_mat(sz, CV_8UC1), uv_in_gapi_mat(sz / 2, CV_8UC2),
```
TolyaTalamanov, line 258: First `Mat` you init from outside of `create_rand_mats` but second `Mat` you init inside of `create_rand_mat`.  I don't think that it's clear code. You can init both of mats inside of function or create them outside then fill using this function. It's up to you.

modules/gapi/perf/common/gapi_render_perf_tests.hpp
```diff
@@ -0,0 +1,56 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+
+#ifndef OPENCV_GAPI_RENDER_PERF_TESTS_HPP
```
TolyaTalamanov, line 8: Why do we need this file ?

Can we move all stuff from here to corresponding *.cpp file ?

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
TolyaTalamanov: Yes, could you fix it for unit tests as well, please ?

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
TolyaTalamanov: Not done

modules/gapi/perf/cpu/gapi_render_perf_tests_cpu.cpp
TolyaTalamanov: As I can see, you've already managed it

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1238 @@
...
+
+    // G-API code //////////////////////////////////////////////////////////////
+    cv::gapi::wip::draw::Prims prims;
+
+    // Rects
+    cv::Rect rect(200, 200, 200, 200);
+    int shift = 0;
+
+    for (int i = 0; i < rects_num; ++i)
+        prims.emplace_back(cv::gapi::wip::draw::Rect(rect, color, thick, lt, shift));
```
TolyaTalamanov, line 1161: Looks like you use the same rect for every iteration

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1238 @@
...
+    create_rand_mats(sz / 2, CV_8UC2, uv_ref_mat, uv_in_gapi_mat);
+
+    // G-API code //////////////////////////////////////////////////////////////
+    cv::gapi::wip::draw::Prims prims;
+
+    // Rects
+    cv::Rect rect(200, 200, 200, 200);
+    int shift = 0;
+
+    for (int i = 0; i < rects_num; ++i)
```
TolyaTalamanov, line 1160: It's better to use brackets even if `for` consist of only one line

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1238 @@
...
+
+    // Text
+    std::string text = "Some text";
+    cv::Point org(200, 200);
+    int ff = FONT_HERSHEY_SIMPLEX;
+    double fs = 2.0;
+    bool blo = false;
+
+    for (int i = 0; i < text_num; ++i)
+        prims.emplace_back(cv::gapi::wip::draw::Text(text, org, ff, fs, color, thick, lt, blo));
```
TolyaTalamanov, line 1179: The same problem with `Text`

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1238 @@
...
+    cv::Rect rect_img(0, 0, 50, 50);
+    double transparency = 1.0;
+
+    cv::Mat img(rect_img.size(), CV_8UC3, color);
+    cv::Mat alpha(rect_img.size(), CV_32FC1, transparency);
+    auto tl = rect_img.tl();
+    cv::Point org_img = {tl.x, tl.y + rect_img.size().height};
+
+    for (int i = 0; i < image_num; ++i)
+        prims.emplace_back(cv::gapi::wip::draw::Image({org_img, img, alpha}));
```
TolyaTalamanov, line 1191: The same problem with `Image`

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Removed

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Done

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Fixed

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Fixed

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Fixed

modules/gapi/perf/common/gapi_render_perf_tests.hpp
aDanPin: Removed

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Fixed

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Fixed

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Fixed

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Yes it is
Please suggest how to avoid this

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1281 @@
...
+        {
+            cell_roi = msc_roi(cv::Rect(j, rows, cellSz, msc_roi.rows - rows));
+            cell_roi = cv::mean(cell_roi);
+        }
+        if (has_crop_x)
+        {
+            cell_roi = msc_roi(cv::Rect(cols, rows, msc_roi.cols - cols, msc_roi.rows - rows));
+            cell_roi = cv::mean(cell_roi);
+        }
+    }
```
dmatveev, line 123: @TolyaTalamanov I thought there workarounds are not needed for the Mosiac anymore, is it correct?

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1281 @@
...
+}
+
+// FIXME: remove this code duplication with /gapi/test/common/gapi_render_tests.hpp
+void blendImageRef(cv::Mat &mat, const cv::Point &org, const cv::Mat &img, const cv::Mat &alpha)
```
dmatveev, line 127: If this function is in `common` already, why it is copied here? Why we can't reuse the old original one?

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1281 @@
...
+
+#ifdef HAVE_FREETYPE
+PERF_TEST_P_(RenderTestFTexts, RenderFTextsPerformanceBGROCVTest)
+{
+    std::wstring text;
+    cv::Size sz;
+    cv::Point org;
+    int fh;
+    cv::Scalar color;
+    MatType type;
```
dmatveev, line 248: Please use the default initializers.

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1281 @@
...
+    prims.emplace_back(cv::gapi::wip::draw::FText{text, org, fh, color});
+
+    cv::GMat in;
+    cv::GArray<cv::gapi::wip::draw::Prim> arr;
+
+    cv::GComputation comp(cv::GIn(in, arr),
+                          cv::GOut(cv::gapi::wip::draw::render3ch(in, arr)));
+
+    // Warm-up graph engine:
+    comp.apply(gin(gapi_mat, prims), gout(gapi_mat));
```
dmatveev, line 266: The problem with this tests now is that the tests can't be instantiated with another backend. Please follow the general model of having `compile_args` as one of test's parameters.

We need to reuse these tests on other platforms.

Also, since we need to do this, this file needs to be a header file not a .cpp, right?

Have a look on this: https://github.com/opencv/opencv/blob/3a46321fef901b8635cf5040a1f1764ffeb86db7/modules/gapi/perf/common/gapi_core_perf_tests.hpp

...And this: https://github.com/opencv/opencv/blob/3a46321fef901b8635cf5040a1f1764ffeb86db7/modules/gapi/perf/common/gapi_core_perf_tests_inl.hpp

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1281 @@
...
+
+    TEST_CYCLE()
+    {
+        comp.apply(gin(gapi_mat, prims), gout(gapi_mat));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
+    {
+        cv::putText(ref_mat, text, org, ff, fs, color, thick, lt, blo);
+    }
```
dmatveev, line 355: There's no need in the extra `{}` here...

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1281 @@
...
+
+    // Comparison //////////////////////////////////////////////////////////////
+    {
+        EXPECT_EQ(0, cv::norm(gapi_mat, ref_mat, NORM_INF));
+    }
```
dmatveev, line 360: ...and here.

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1281 @@
...
+
+PERF_TEST_P_(RenderTestCircles, RenderCirclesPerformanceBGROCVTest)
+{
+    cv::Point center;
+    int radius;
+    cv::Scalar color;
+    int thick;
+    int lt;
+    int shift;
+    MatType type;
```
dmatveev, line 539: Default initializers for scalar types are missing again (yes it is written to with `tie` but static analysis tools we use can't into it).

modules/gapi/perf/common/gapi_render_perf_tests.cpp
TolyaTalamanov: @dmatveev Let me explain, in case `render` we implement tests for every backend separately, because there is no chance to implement the same reference versions

modules/gapi/perf/common/gapi_render_perf_tests.cpp
TolyaTalamanov: We've already discussed it, so can we share stuff between unit & perf tests ?

modules/gapi/perf/common/gapi_render_perf_tests.cpp
dmatveev: @TolyaTalamanov then it shouldn't be common.

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Done

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Done

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Done

modules/gapi/perf/common/gapi_render_perf_tests.cpp
aDanPin: Done

modules/gapi/perf/common/gapi_render_perf_tests.cpp
dmatveev: You can calculate x/y/width/height based on `i` to make `n` different rectangles.

modules/gapi/perf/common/gapi_render_perf_tests.cpp
```diff
@@ -0,0 +1,1257 @@
...
+#ifdef HAVE_FREETYPE
+#include <codecvt>
+#endif // HAVE_FREETYPE
+
+namespace opencv_test
+{
+  using namespace perf;
+
+  #ifdef HAVE_FREETYPE
+  class RenderTestFTexts : public TestPerfParams<tuple<std::wstring, cv::Size, cv::Point,
```
aDanPin, line 22: Remove #ifdef's

aDanPin: Do same thing for unit tests

modules/gapi/perf/common/gapi_render_perf_tests.cpp
TolyaTalamanov: Currently case with fractional mosaic is handled inside opencv backend.
I've opened a ticket to handle it in general case for every backend

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+    comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+               cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+                   cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 375: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+
+    // Warm-up graph engine:
+    comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 426: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+    comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+               cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+                   cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 477: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+
+    // Warm-up graph engine:
+    comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 529: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+    comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+               cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+                   cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 581: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+
+    // Warm-up graph engine:
+    comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 634: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+    comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+               cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+                   cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 687: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+
+    // Warm-up graph engine:
+    comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 736: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+    comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+               cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+                   cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 786: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+
+    // Warm-up graph engine:
+    comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 841: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+    comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+               cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+                   cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 895: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+
+    // Warm-up graph engine:
+    comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 948: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+    comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+               cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+                   cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 1000: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+
+    // Warm-up graph engine:
+    comp.apply(gin(gapi_mat, prims), gout(gapi_out_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(gin(gapi_mat, prims), gout(gapi_out_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 1097: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+    comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+               cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(cv::gin(y_in_gapi_mat, uv_in_gapi_mat, prims),
+                   cv::gout(y_out_gapi_mat, uv_out_gapi_mat), std::move(comp_args));
+    }
+
+    // OpenCV code //////////////////////////////////////////////////////////////
```
TolyaTalamanov, line 1190: We decided to eliminate code with the accuracy check, didn't we ?

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+        prims.emplace_back(cv::gapi::wip::draw::Rect(rect, color, thick, lt, shift));
+    }
+
+    // Mosaic
+    cv::Rect mos(200, 200, 200, 200);
+    int cellsz = 25;
+    int decim = 0;
+
+    for (int i = 0; i < rects_num; ++i) {
+        prims.emplace_back(cv::gapi::wip::draw::Mosaic(mos, cellsz, decim));
```
TolyaTalamanov, line 1056: You draw a mosaic in the same place

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+
+    // Text
+    std::string text = "Some text";
+    cv::Point org(200, 200);
+    int ff = FONT_HERSHEY_SIMPLEX;
+    double fs = 2.0;
+    bool blo = false;
+
+    for (int i = 0; i < text_num; ++i) {
+        prims.emplace_back(cv::gapi::wip::draw::Text(text, org, ff, fs, color, thick, lt, blo));
```
TolyaTalamanov, line 1067: The same for text

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,1216 @@
...
+    cv::Rect rect_img(100, 100, 100, 100);
+    double transparency = 1.0;
+
+    cv::Mat img(rect_img.size(), CV_8UC3, color);
+    cv::Mat alpha(rect_img.size(), CV_32FC1, transparency);
+    auto tl = rect_img.tl();
+    cv::Point org_img = {tl.x, tl.y + rect_img.size().height};
+
+    for (int i = 0; i < image_num; ++i) {
+        prims.emplace_back(cv::gapi::wip::draw::Image(org_img, img, alpha));
```
TolyaTalamanov, line 1080: The same for image

TolyaTalamanov:
BTW please take a look how to implemented performance tests for cpu backend. Tests implementations place in `common` folder but instances place in separate `cpp` file for every backend

modules/gapi/perf/common/gapi_render_perf_tests.cpp
TolyaTalamanov: There isn't changes for unit tests. But ok, let's do it in separate PR

TolyaTalamanov:
LGTM 👍 

@dmatveev please take a look and go merge it !!!

dmatveev:
@TolyaTalamanov are you sure? If I find something, would you fix that?

modules/gapi/perf/common/gapi_render_perf_tests.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include "../../test/common/gapi_tests_common.hpp"
+#include <opencv2/gapi/render/render.hpp>
+#include <opencv2/gapi/util/compiler_hints.hpp> //suppress_unused_warning
+
+namespace opencv_test
+{
+
+// FIXME: remove this code duplication with gapi/test/render/gapi_render_tests_ocv.cpp
+static inline std::wstring to_wstring(const char* bytes)
```
TolyaTalamanov, line 22: Wrap to namespace

modules/gapi/perf/common/gapi_render_perf_tests.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#ifndef OPENCV_GAPI_RENDER_PERF_TESTS_HPP
+#define OPENCV_GAPI_RENDER_PERF_TESTS_HPP
+
+
+#include <codecvt>
+
+#include "../../test/common/gapi_tests_common.hpp"
+#include <opencv2/gapi/render/render.hpp>
+#include <opencv2/gapi/util/compiler_hints.hpp> //suppress_unused_warning
```
TolyaTalamanov, line 16: Remove

modules/gapi/perf/render/gapi_render_perf_tests_ocv.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+
+#ifdef HAVE_FREETYPE
+INSTANTIATE_TEST_CASE_P(RenderTestFTexts, RenderTestFTexts,
+                        Combine(Values(to_wstring("\xe4\xbd\xa0\xe5\xa5\xbd")),
+                                Values(szVGA, sz720p, sz1080p),
+                                Values(cv::Point(50, 50)),
+                                Values(60),
+                                Values(cv::Scalar(200, 100, 25)),
+                                Values(CV_8UC3),
```
TolyaTalamanov, line 23: Remove

modules/gapi/perf/common/gapi_render_perf_tests.hpp
```diff
@@ -13,39 +13,40 @@
...
 namespace opencv_test
 {
 
+namespace {
 // FIXME: remove this code duplication with gapi/test/render/gapi_render_tests_ocv.cpp
-static inline std::wstring to_wstring(const char* bytes)
-{
-    std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> converter;
-    return converter.from_bytes(bytes);
+  inline std::wstring to_wstring(const char* bytes)
```
TolyaTalamanov, line 22: `inline` isn't necessary

modules/gapi/perf/common/gapi_render_perf_tests_inl.hpp
```diff
@@ -0,0 +1,827 @@
...
+
+    cv::GComputation comp(cv::GIn(in, arr),
+                          cv::GOut(cv::gapi::wip::draw::render3ch(in, arr)));
+
+    // Warm-up graph engine:
+    comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
+
+    TEST_CYCLE()
+    {
+        comp.apply(gin(gapi_mat, prims), gout(gapi_mat), std::move(comp_args));
```
TolyaTalamanov, line 52: `comp_args` has been already moved.   
Check test for `CPU`:  https://github.com/opencv/opencv/blob/master/modules/gapi/perf/common/gapi_core_perf_tests_inl.hpp#L42

TolyaTalamanov:
@dmatveev

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
dmatveev: So after a long-long discussion we decided to go with basic checks only.
According to @TolyaTalamanov , our different (private) OSD backends can never converge to the same accuracy, so in unit tests we test those separately (unlike we do with the rest of our regular kernels).

In order to reuse the performance tests for OSD, we relax this here now.

modules/gapi/perf/common/gapi_render_perf_tests.hpp
dmatveev: Not a big deal though

modules/gapi/perf/render/gapi_render_perf_tests_inl.hpp
TolyaTalamanov: I was sure that we've removed those checks !

@aDanPin Can u comment it ?

elatkin: May "vast experience" includes the cases when people used to test and report great performance of their software, while the software produced completely wrong results.  

What if your software (the backend you test) gets completely broken, say, does not start at all? Performance would look great, but the output would be completely useless.

dmatveev:
cc: @smirnov-alexey

dmatveev:
@alalek it seems this thing is ready for merge -- fyi

modules/gapi/perf/common/gapi_render_perf_tests.hpp
```diff
@@ -0,0 +1,55 @@
...
+      std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> converter;
+      return converter.from_bytes(bytes);
+  }
+}
+
+  using namespace perf;
+
+  class RenderTestFTexts : public TestPerfParams<tuple<std::wstring, cv::Size, cv::Point,
+                                                       int, cv::Scalar, cv::GCompileArgs>> {};
+  class RenderTestTexts : public TestPerfParams<tuple<std::string, cv::Size, cv::Point,
```
alalek, line 34: Avoid indentation in namespaces (especially whole-file namespaces)

modules/gapi/perf/common/gapi_render_perf_tests.hpp
```diff
@@ -0,0 +1,55 @@
...
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+
+#ifndef OPENCV_GAPI_RENDER_PERF_TESTS_HPP
+#define OPENCV_GAPI_RENDER_PERF_TESTS_HPP
+
+
+#include <codecvt>
```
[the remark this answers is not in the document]
TolyaTalamanov: Why does it work for unit tests ?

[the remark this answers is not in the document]
TolyaTalamanov: FreeType job isn't validated on GCC less 5+ ?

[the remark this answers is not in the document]
alalek: Due to `#ifdef HAVE_FREETYPE` => this image is Ubuntu 16.04 only.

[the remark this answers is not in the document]
TolyaTalamanov: Yes, but what is the correct solution ? How should `w_string` should passed into the test ?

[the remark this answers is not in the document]
alalek: Why do you can't pass final value without calling extra conversion helpers?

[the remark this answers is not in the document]
TolyaTalamanov: Like this: `str = L"hello, world!` ?

PR #16774, opened 2020-03-10
core: update CPUs detection

alalek:
relates #16457

- cache value, evaluate once
- better support for MINGW
- anything in 'cv' namespace
- test: dump number of active threads


alalek:
:+1:

PR #16786, opened 2020-03-11
core: adjust type of allocator_stats counter, allow to disable

### LINKED ISSUE #16398: opencv mips1 : atomic<long long> not supported
On MIPS1 version 1, atomic only supports 32 bit types.
So in cmake/checks/atomic_check.cpp and modules/core/include/opencv2/core/utils/allocator_stats.impl.hpp, atomic<long long> won't compile. Do we really need long long types for the statistics? If not, you will fix opencv for mips

alalek:
resolves #16398


alalek:
:+1:

PR #16717, opened 2020-03-03
G-API: cv::gapi::goodFeaturesToTrack() implementation

OrestChura:
### goodFeaturesToTrack() kernel is included in the standard CPU G-API kernel pack now
 - there are performance and accuracy tests
 - additional comparison functions and fixtures implemented

### Pull Request Readiness Checklist
- [x] I agree to contribute to the project under OpenCV (BSD) License.
- [x] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
- [x] The PR is proposed to proper branch
- [x] There is accuracy test, performance test _and test data in opencv_extra repository, if applicable_
      _Patch to opencv_extra has the same branch name._
- [x] The feature is well documented _and sample code can be built with the project CMake_

```
build_image:Custom=ubuntu-clang:18.04
build_gapi_standalone:Custom=ade-0.1.1f

force_builders=Custom,Custom Win,Custom Mac
Xbuild_image:Custom=ubuntu-openvino-2019r2.0:16.04
build_image:Custom Win=openvino-2019r2.0
build_image:Custom Mac=openvino-2019r2.0

test_modules:Custom=gapi
test_modules:Custom Win=gapi
test_modules:Custom Mac=gapi

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*

build_gapi_standalone:Linux x64=ade-0.1.1f
build_gapi_standalone:Win64=ade-0.1.1f
Xbuild_gapi_standalone:Mac=ade-0.1.1f
build_gapi_standalone:Linux x64 Debug=ade-0.1.1f
build_gapi_standalone:Custom Win=ade-0.1.1f

```

PR #16788, opened 2020-03-12
G-API: gapi::calcOpticalFlowPyrLK() Implementation

OrestChura:
### calcOpticalFlowPyrLK() kernel is included in the standard CPU G-API kernel pack now
 - two implementations with different input (GMat's or GArray<GMat>'s) to support features to build / not to build optical flow pyramids inside the kernel
 - opencv_gapi is linked with opencv_video now
 - there are performance and accuracy tests (implemented actively using cv::video perf test for optFlowPyrLK)

```

test_modules:Custom=gapi
test_modules:Custom Win=gapi
test_modules:Custom Mac=gapi

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*

```

### DIFF
--- a/modules/gapi/CMakeLists.txt
+++ b/modules/gapi/CMakeLists.txt
@@ -15,7 +15,12 @@ endif()
 
 set(the_description "OpenCV G-API Core Module")
 
-ocv_add_module(gapi opencv_imgproc)
+ocv_add_module(gapi
+    REQUIRED
+      opencv_imgproc
+    OPTIONAL
+      opencv_video
+)
 
 if(MSVC)
   # Disable obsollete warning C4503 popping up on MSVC <<2017
@@ -55,6 +60,7 @@ set(gapi_srcs
     src/api/operators.cpp
     src/api/kernels_core.cpp
     src/api/kernels_imgproc.cpp
+    src/api/kernels_video.cpp
     src/api/render.cpp
     src/api/render_ocv.cpp
     src/api/ginfer.cpp
@@ -87,6 +93,7 @@ set(gapi_srcs
     src/backends/cpu/gcpubackend.cpp
     src/backends/cpu/gcpukernel.cpp
     src/backends/cpu/gcpuimgproc.cpp
+    src/backends/cpu/gcpuvideo.cpp
     src/backends/cpu/gcpucore.cpp
 
     # Fluid Backend (also built-in, FIXME:move away)
--- a/modules/gapi/include/opencv2/gapi/cpu/video.hpp
+++ b/modules/gapi/include/opencv2/gapi/cpu/video.hpp
@@ -1,0 +1,25 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+#ifndef OPENCV_GAPI_CPU_VIDEO_API_HPP
+#define OPENCV_GAPI_CPU_VIDEO_API_HPP
+
+#include <opencv2/gapi/gkernel.hpp> // GKernelPackage
+
+namespace cv {
+namespace gapi {
+namespace video {
+namespace cpu {
+
+GAPI_EXPORTS GKernelPackage kernels();
+
+} // namespace cpu
+} // namespace video
+} // namespace gapi
+} // namespace cv
+
+
+#endif // OPENCV_GAPI_CPU_VIDEO_API_HPP
--- a/modules/gapi/include/opencv2/gapi/imgproc.hpp
+++ b/modules/gapi/include/opencv2/gapi/imgproc.hpp
@@ -261,7 +261,6 @@ namespace imgproc {
 
 } //namespace imgproc
 
-
 //! @addtogroup gapi_filters
 //! @{
 /** @brief Applies a separable linear filter to a matrix(image).
--- a/modules/gapi/include/opencv2/gapi/video.hpp
+++ b/modules/gapi/include/opencv2/gapi/video.hpp
@@ -1,0 +1,135 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+#ifndef OPENCV_GAPI_VIDEO_HPP
+#define OPENCV_GAPI_VIDEO_HPP
+
+#include <utility> // std::tuple
+
+#include <opencv2/gapi/gkernel.hpp>
+
+
+/** \defgroup gapi_video G-API Video processing functionality
+ */
+
+namespace cv { namespace gapi {
+namespace  video
+{
+    using GOptFlowLKOutput = std::tuple<cv::GArray<cv::Point2f>,
+                                        cv::GArray<uchar>,
+                                        cv::GArray<float>>;
+
+    G_TYPED_KERNEL(GCalcOptFlowLK,
+                   <GOptFlowLKOutput(GMat,GMat,cv::GArray<cv::Point2f>,cv::GArray<cv::Point2f>,Size,
+                                     int,TermCriteria,int,double)>,
+                   "org.opencv.video.calcOpticalFlowPyrLK")
+    {
+        static std::tuple<GArrayDesc,GArrayDesc,GArrayDesc> outMeta(GMatDesc,GMatDesc,GArrayDesc,
+                                                                    GArrayDesc,const Size&,int,
+                                                                    const TermCriteria&,int,double)
+        {
+            return std::make_tuple(empty_array_desc(), empty_array_desc(), empty_array_desc());
+        }
+
+    };
+
+    G_TYPED_KERNEL(GCalcOptFlowLKForPyr,
+                   <GOptFlowLKOutput(cv::GArray<cv::GMat>,cv::GArray<cv::GMat>,
+                                     cv::GArray<cv::Point2f>,cv::GArray<cv::Point2f>,Size,int,
+                                     TermCriteria,int,double)>,
+                   "org.opencv.video.calcOpticalFlowPyrLKForPyr")
+    {
+        static std::tuple<GArrayDesc,GArrayDesc,GArrayDesc> outMeta(GArrayDesc,GArrayDesc,
+                                                                    GArrayDesc,GArrayDesc,
+                                                                    const Size&,int,
+                                                                    const TermCriteria&,int,double)
+        {
+            return std::make_tuple(empty_array_desc(), empty_array_desc(), empty_array_desc());
+        }
+    };
+} //namespace video
+
+//! @addtogroup gapi_video
+//! @{
+/** @brief Calculates an optical flow for a sparse feature set using the iterative Lucas-Kanade
+method with pyramids.
+
+See @cite Bouguet00 .
+
+@note Function textual ID is "org.opencv.video.calcOpticalFlowPyrLK"
+
+@param prevImg first 8-bit input image (GMat) or pyramid (GArray<GMat>) constructed by
+buildOpticalFlowPyramid.
+@param nextImg second input image (GMat) or pyramid (GArray<GMat>) of the same size and the same
+type as prevImg.
+@param prevPts GArray of 2D points for which the flow needs to be found; point coordinates must be
+single-precision floating-point numbers.
+@param predPts GArray of 2D points initial for the flow search; make sense only when
+OPTFLOW_USE_INITIAL_FLOW flag is passed; in that case the vector must have the same size as in
+the input.
+@param winSize size of the search window at each pyramid level.
+@param maxLevel 0-based maximal pyramid level number; if set to 0, pyramids are not used (single
+level), if set to 1, two levels are used, and so on; if pyramids are passed to input then
+algorithm will use as many levels as pyramids have but no more than maxLevel.
+@param criteria parameter, specifying the termination criteria of the iterative search algorithm
+(after the specified maximum number of iterations criteria.maxCount or when the search window
+moves by less than criteria.epsilon).
+@param flags operation flags:
+ -   **OPTFLOW_USE_INITIAL_FLOW** uses initial estimations, stored in nextPts; if the flag is
+     not set, then prevPts is copied to nextPts and is considered the initial estimate.
+ -   **OPTFLOW_LK_GET_MIN_EIGENVALS** use minimum eigen values as an error measure (see
+     minEigThreshold description); if the flag is not set, then L1 distance between patches
+     around the original and a moved point, divided by number of pixels in a window, is used as a
+     error measure.
+@param minEigThresh the algorithm calculates the minimum eigen value of a 2x2 normal matrix of
+optical flow equations (this matrix is called a spatial gradient matrix in @cite Bouguet00), divided
+by number of pixels in a window; if this value is less than minEigThreshold, then a corresponding
+feature is filtered out and its flow is not processed, so it allows to remove bad points and get a
+performance boost.
+
+@return GArray of 2D points (with single-precision floating-point coordinates)
+containing the calculated new positions of input features in the second image.
+@return status GArray (of unsigned chars); each element of the vector is set to 1 if
+the flow for the corresponding features has been found, otherwise, it is set to 0.
+@return GArray of errors (doubles); each element of the vector is set to an error for the
+corresponding feature, type of the error measure can be set in flags parameter; if the flow wasn't
+found then the error is not defined (use the status parameter to find such cases).
+ */
+GAPI_EXPORTS std::tuple<GArray<Point2f>, GArray<uchar>, GArray<float>>
+calcOpticalFlowPyrLK(const GMat            &prevImg,
+                     const GMat            &nextImg,
+                     const GArray<Point2f> &prevPts,
+                     const GArray<Point2f> &predPts,
+                     const Size            &winSize      = Size(21, 21),
+                           int              maxLevel     = 3,
+                     const TermCriteria    &criteria     = TermCriteria(TermCriteria::COUNT |
+                                                                        TermCriteria::EPS,
+                                                                        30, 0.01),
+                           int              flags        = 0,
+                           double           minEigThresh = 1e-4);
+
+/**
+@overload
+@note Function textual ID is "org.opencv.video.calcOpticalFlowPyrLKForPyr"
+*/
+GAPI_EXPORTS std::tuple<GArray<Point2f>, GArray<uchar>, GArray<float>>
+calcOpticalFlowPyrLK(const GArray<GMat>    &prevPyr,
+                     const GArray<GMat>    &nextPyr,
+                     const GArray<Point2f> &prevPts,
+                     const GArray<Point2f> &predPts,
+                     const Size            &winSize      = Size(21, 21),
+                           int              maxLevel     = 3,
+                     const TermCriteria    &criteria     = TermCriteria(TermCriteria::COUNT |
+                                                                        TermCriteria::EPS,
+                                                                        30, 0.01),
+                           int              flags        = 0,
+                           double           minEigThresh = 1e-4);
+
+//! @} gapi_video
+} //namespace gapi
+} //namespace cv
+
+#endif // OPENCV_GAPI_VIDEO_HPP
--- a/modules/gapi/perf/common/gapi_imgproc_perf_tests.hpp
+++ b/modules/gapi/perf/common/gapi_imgproc_perf_tests.hpp
@@ -49,5 +49,6 @@ class YUV2BGRPerfTest : public TestPerfParams<tuple<compare_f, cv::Size, cv::GCo
 class RGB2HSVPerfTest : public TestPerfParams<tuple<compare_f, cv::Size, cv::GCompileArgs>> {};
 class BayerGR2RGBPerfTest : public TestPerfParams<tuple<compare_f, cv::Size, cv::GCompileArgs>> {};
 class RGB2YUV422PerfTest  : public TestPerfParams<tuple<compare_f, cv::Size, cv::GCompileArgs>> {};
-}
+} // opencv_test
+
 #endif //OPENCV_GAPI_IMGPROC_PERF_TESTS_HPP
--- a/modules/gapi/perf/common/gapi_video_perf_tests_inl.hpp
+++ b/modules/gapi/perf/common/gapi_video_perf_tests_inl.hpp
@@ -1,0 +1,90 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+#ifndef OPENCV_GAPI_VIDEO_PERF_TESTS_INL_HPP
+#define OPENCV_GAPI_VIDEO_PERF_TESTS_INL_HPP
+
+#include <iostream>
+
+#include "gapi_video_perf_tests.hpp"
+
+namespace opencv_test
+{
+
+  using namespace perf;
+
+//------------------------------------------------------------------------------
+
+PERF_TEST_P_(OptFlowLKPerfTest, TestPerformance)
+{
+    std::vector<cv::Point2f> outPtsOCV,    outPtsGAPI,    inPts;
+    std::vector<uchar>       outStatusOCV, outStatusGAPI;
+    std::vector<float>       outErrOCV,    outErrGAPI;
+
+    OptFlowLKTestParams params;
+    std::tie(params.fileNamePattern, params.channels,
+             params.pointsNum, params.winSize, params.criteria,
+             params.compileArgs) = GetParam();
+
+    OptFlowLKTestOutput outOCV  { outPtsOCV,  outStatusOCV,  outErrOCV };
+    OptFlowLKTestOutput outGAPI { outPtsGAPI, outStatusGAPI, outErrGAPI };
+
+    cv::GComputation c = runOCVnGAPIOptFlowLK(*this, inPts, params, outOCV, outGAPI);
+
+    declare.in(in_mat1, in_mat2, inPts).out(outPtsGAPI, outStatusGAPI, outErrGAPI);
+
+    TEST_CYCLE()
+    {
+        c.apply(cv::gin(in_mat1, in_mat2, inPts, std::vector<cv::Point2f>{ }),
+                cv::gout(outPtsGAPI, outStatusGAPI, outErrGAPI));
+    }
+
+    // Comparison //////////////////////////////////////////////////////////////
+    compareOutputsOptFlow(outOCV, outGAPI);
+
+    SANITY_CHECK_NOTHING();
+}
+
+//------------------------------------------------------------------------------
+
+PERF_TEST_P_(OptFlowLKForPyrPerfTest, TestPerformance)
+{
+    std::vector<cv::Mat>     inPyr1, inPyr2;
+    std::vector<cv::Point2f> outPtsOCV,    outPtsGAPI,    inPts;
+    std::vector<uchar>       outStatusOCV, outStatusGAPI;
+    std::vector<float>       outErrOCV,    outErrGAPI;
+
+    bool withDeriv = false;
+    OptFlowLKTestParams params;
+    std::tie(params.fileNamePattern, params.channels,
+             params.pointsNum, params.winSize, params.criteria,
+             withDeriv, params.compileArgs) = GetParam();
+
+    OptFlowLKTestInput<std::vector<cv::Mat>> in { inPyr1, inPyr2, inPts };
+    OptFlowLKTestOutput outOCV  { outPtsOCV,  outStatusOCV,  outErrOCV };
+    OptFlowLKTestOutput outGAPI { outPtsGAPI, outStatusGAPI, outErrGAPI };
+
+    cv::GComputation c = runOCVnGAPIOptFlowLKForPyr(*this, in, params, withDeriv, outOCV, outGAPI);
+
+    declare.in(inPyr1, inPyr2, inPts).out(outPtsGAPI, outStatusGAPI, outErrGAPI);
+
+    TEST_CYCLE()
+    {
+        c.apply(cv::gin(inPyr1, inPyr2, inPts, std::vector<cv::Point2f>{ }),
+                cv::gout(outPtsGAPI, outStatusGAPI, outErrGAPI));
+    }
+
+    // Comparison //////////////////////////////////////////////////////////////
+    compareOutputsOptFlow(outOCV, outGAPI);
+
+    SANITY_CHECK_NOTHING();
+}
+
+//------------------------------------------------------------------------------
+
+} // opencv_test
+
+#endif // OPENCV_GAPI_VIDEO_PERF_TESTS_INL_HPP
--- a/modules/gapi/perf/cpu/gapi_imgproc_perf_tests_cpu.cpp
+++ b/modules/gapi/perf/cpu/gapi_imgproc_perf_tests_cpu.cpp
@@ -220,4 +220,4 @@ INSTANTIATE_TEST_CASE_P(RGB2YUV422PerfTestCPU, RGB2YUV422PerfTest,
         Combine(Values(ToleranceColor(1e-3).to_compare_f()),
             Values(szVGA, sz720p, sz1080p),
             Values(cv::compile_args(IMGPROC_CPU))));
-}
+} // opencv_test
--- a/modules/gapi/perf/cpu/gapi_video_perf_tests_cpu.cpp
+++ b/modules/gapi/perf/cpu/gapi_video_perf_tests_cpu.cpp
@@ -1,0 +1,64 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+
+#include "../perf_precomp.hpp"
+
+#include "../common/gapi_video_perf_tests.hpp"
+#include <opencv2/gapi/cpu/video.hpp>
+
+namespace
+{
+#define VIDEO_CPU cv::gapi::video::cpu::kernels()
+
+#ifdef HAVE_OPENCV_VIDEO
+#define WITH_VIDEO(X) X
+#else
+#define WITH_VIDEO(X) DISABLED_##X
+#endif // HAVE_OPENCV_VIDEO
+
+#define INSTANTIATE_TEST_CASE_MACRO_P(prefix, test_case_name, generator, ...) \
+    INSTANTIATE_TEST_CASE_P(prefix, test_case_name, generator, __VA_ARGS__)
+} // namespace
+
+
+namespace opencv_test
+{
+INSTANTIATE_TEST_CASE_MACRO_P(WITH_VIDEO(OptFlowLKPerfTestCPU), OptFlowLKPerfTest,
+                              Combine(Values("cv/optflow/rock_%01d.bmp",
+                                             "cv/optflow/frames/1080p_%02d.png"),
+                                      Values(1, 3, 4),
+                                      Values(std::make_tuple(9, 9), std::make_tuple(15, 15)),
+                                      Values(7, 11),
+                                      Values(cv::TermCriteria(cv::TermCriteria::COUNT |
+                                                              cv::TermCriteria::EPS,
+                                                              30, 0.01)),
+                                      Values(cv::compile_args(VIDEO_CPU))));
+
+INSTANTIATE_TEST_CASE_MACRO_P(WITH_VIDEO(OptFlowLKForPyrPerfTestCPU), OptFlowLKForPyrPerfTest,
+                              Combine(Values("cv/optflow/rock_%01d.bmp",
+                                             "cv/optflow/frames/1080p_%02d.png"),
+                                      Values(1, 3, 4),
+                                      Values(std::make_tuple(9, 9), std::make_tuple(15, 15)),
+                                      Values(7, 11),
+                                      Values(cv::TermCriteria(cv::TermCriteria::COUNT |
+                                                              cv::TermCriteria::EPS,
+                                                              30, 0.01)),
+                                      Values(true, false),
+                                      Values(cv::compile_args(VIDEO_CPU))));
+
+INSTANTIATE_TEST_CASE_MACRO_P(WITH_VIDEO(OptFlowLKInternalPerfTestCPU),
+                              OptFlowLKForPyrPerfTest,
+                              Combine(Values("cv/optflow/rock_%01d.bmp"),
+                                      Values(1),
+                                      Values(std::make_tuple(10, 10)),
+                                      Values(15),
+                                      Values(cv::TermCriteria(cv::TermCriteria::COUNT |
+                                                              cv::TermCriteria::EPS,
+                                                              21, 0.05)),
+                                      Values(true),
+                                      Values(cv::compile_args(VIDEO_CPU))));
+} // opencv_test
--- a/modules/gapi/src/api/kernels_video.cpp
+++ b/modules/gapi/src/api/kernels_video.cpp
@@ -1,0 +1,45 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+
+#include "precomp.hpp"
+
+#include <opencv2/gapi/video.hpp>
+
+namespace cv { namespace gapi {
+using namespace video;
+
+
+GOptFlowLKOutput calcOpticalFlowPyrLK(const GMat                    &prevImg,
+                                      const GMat                    &nextImg,
+                                      const cv::GArray<cv::Point2f> &prevPts,
+                                      const cv::GArray<cv::Point2f> &predPts,
+                                      const Size                    &winSize,
+                                            int                      maxLevel,
+                                      const TermCriteria            &criteria,
+                                            int                      flags,
+                                            double                   minEigThresh)
+{
+    return GCalcOptFlowLK::on(prevImg, nextImg, prevPts, predPts, winSize, maxLevel,
+                              criteria, flags, minEigThresh);
+}
+
+GOptFlowLKOutput calcOpticalFlowPyrLK(const cv::GArray<cv::GMat>    &prevPyr,
+                                      const cv::GArray<cv::GMat>    &nextPyr,
+                                      const cv::GArray<cv::Point2f> &prevPts,
+                                      const cv::GArray<cv::Point2f> &predPts,
+                                      const Size                    &winSize,
+                                            int                      maxLevel,
+                                      const TermCriteria            &criteria,
+                                            int                      flags,
+                                            double                   minEigThresh)
+{
+    return GCalcOptFlowLKForPyr::on(prevPyr, nextPyr, prevPts, predPts, winSize, maxLevel,
+                                    criteria, flags, minEigThresh);
+}
+
+} //namespace gapi
+} //namespace cv
--- a/modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
+++ b/modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
@@ -784,7 +784,6 @@ TEST_P(RGB2YUV422Test, AccuracyTest)
         EXPECT_EQ(out_mat_gapi.size(), sz);
     }
 }
-
 } // opencv_test
 
 #endif //OPENCV_GAPI_IMGPROC_TESTS_INL_HPP
--- a/modules/gapi/test/common/gapi_tests_common.hpp
+++ b/modules/gapi/test/common/gapi_tests_common.hpp
@@ -196,24 +196,31 @@ public:
         }
     }
 
+    void initMatsFromImages(int channels, const std::string& pattern, int imgNum)
+    {
+        initTestDataPath();
+        GAPI_Assert(channels == 1 || channels == 3 || channels == 4);
+        const int flags = (channels == 1) ? cv::IMREAD_GRAYSCALE : cv::IMREAD_COLOR;
+
+        cv::Mat m1 = cv::imread(findDataFile(cv::format(pattern.c_str(), imgNum)), flags);
+        cv::Mat m2 = cv::imread(findDataFile(cv::format(pattern.c_str(), imgNum + 1)), flags);
+        if (channels == 4)
+        {
+            cvtColor(m1, in_mat1, cv::COLOR_BGR2BGRA);
+            cvtColor(m2, in_mat2, cv::COLOR_BGR2BGRA);
+        }
+        else
+        {
+            std::tie(in_mat1, in_mat2) = std::make_tuple(m1, m2);
+        }
+    }
+
     // empty function intended to show that nothing is to be initialized via TestFunctional methods
     void initNothing(int, cv::Size, int, bool = true) {}
 };
 
 template<class T>
-class TestParams: public TestFunctional, public TestWithParam<T>{};
-
-template<class T>
 class TestPerfParams: public TestFunctional, public perf::TestBaseWithParam<T>{};
-
-using compare_f = std::function<bool(const cv::Mat &a, const cv::Mat &b)>;
-
-using compare_scalar_f = std::function<bool(const cv::Scalar &a, const cv::Scalar &b)>;
-
-template<typename Elem>
-using compare_vector_f = std::function<bool(const std::vector<Elem> &a,
-                                            const std::vector<Elem> &b)>;
-
 
 // FIXME: re-use MatType. current problem: "special values" interpreted incorrectly (-1 is printed
 //        as 16FC512)
@@ -367,6 +374,14 @@ struct TestWithParamsSpecific : public TestWithParamsBase<ParamsSpecific<Specifi
 // Wrapper for test fixture API. Use to specify multiple types.
 // Example: FIXTURE_API(int, bool) expands to <int, bool>
 #define FIXTURE_API(...) <__VA_ARGS__>
+
+
+using compare_f = std::function<bool(const cv::Mat &a, const cv::Mat &b)>;
+using compare_scalar_f = std::function<bool(const cv::Scalar &a, const cv::Scalar &b)>;
+
+template<typename Elem>
+using compare_vector_f = std::function<bool(const std::vector<Elem> &a,
+                                            const std::vector<Elem> &b)>;
 
 template<typename T1, typename T2>
 struct CompareF
--- a/modules/gapi/test/common/gapi_video_tests.cpp
+++ b/modules/gapi/test/common/gapi_video_tests.cpp
@@ -1,0 +1,9 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+
+#include "../test_precomp.hpp"
+#include "gapi_video_tests_inl.hpp"
--- a/modules/gapi/test/common/gapi_video_tests_common.hpp
+++ b/modules/gapi/test/common/gapi_video_tests_common.hpp
@@ -1,0 +1,249 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+#ifndef OPENCV_GAPI_VIDEO_TESTS_COMMON_HPP
+#define OPENCV_GAPI_VIDEO_TESTS_COMMON_HPP
+
+#include "gapi_tests_common.hpp"
+#include "../../include/opencv2/gapi/video.hpp"
+
+#ifdef HAVE_OPENCV_VIDEO
+#include <opencv2/video.hpp>
+#endif // HAVE_OPENCV_VIDEO
+
+
+
+namespace opencv_test
+{
+namespace
+{
+inline void initTrackingPointsArray(std::vector<cv::Point2f>& points, int width, int height,
+                                    int nPointsX, int nPointsY)
+{
+    if (nPointsX > width || nPointsY > height)
+    {
+        FAIL() << "Specified points number is too big";
+    }
+
+    int stepX = width  / nPointsX;
+    int stepY = height / nPointsY;
+
+
+    points.clear();
+    GAPI_Assert((nPointsX >= 0) && (nPointsY) >= 0);
+    points.reserve(static_cast<size_t>(nPointsX * nPointsY));
+
+    for (int x = stepX / 2; x < width; x += stepX)
+    {
+        for (int y = stepY / 2; y < height; y += stepY)
+        {
+            Point2f pt(static_cast<float>(x), static_cast<float>(y));
+            points.push_back(pt);
+        }
+    }
+}
+
+template<typename Type>
+struct OptFlowLKTestInput
+{
+    Type& prevData;
+    Type& nextData;
+    std::vector<cv::Point2f>& prevPoints;
+};
+
+struct OptFlowLKTestOutput
+{
+    std::vector<cv::Point2f> &nextPoints;
+    std::vector<uchar>       &statuses;
+    std::vector<float>       &errors;
+};
+
+struct OptFlowLKTestParams
+{
+    OptFlowLKTestParams(): fileNamePattern(""), format(1), channels(0), pointsNum{0, 0},
+                           winSize(0), maxLevel(3), minEigThreshold(1e-4), flags(0) { }
+
+    OptFlowLKTestParams(const std::string& namePat, int chans,
+                        const std::tuple<int,int>& ptsNum, int winSz,
+                        const cv::TermCriteria& crit, const cv::GCompileArgs& compArgs,
+                        int flgs = 0, int fmt = 1, int maxLvl = 3, double minEigThresh = 1e-4):
+
+                        fileNamePattern(namePat), format(fmt), channels(chans),
+                        pointsNum(ptsNum), winSize(winSz), maxLevel(maxLvl),
+                        criteria(crit), minEigThreshold(minEigThresh), compileArgs(compArgs),
+                        flags(flgs) { }
+
+    std::string fileNamePattern   = "";
+    int format                    = 1;
+    int channels                  = 0;
+    std::tuple<int,int> pointsNum = std::make_tuple(0, 0);
+    int winSize                   = 0;
+    int maxLevel                  = 3;
+    cv::TermCriteria criteria;
+    double minEigThreshold        = 1e-4;
+    cv::GCompileArgs compileArgs;
+    int flags                     = 0;
+};
+
+#ifdef HAVE_OPENCV_VIDEO
+
+template<typename GType, typename Type>
+cv::GComputation runOCVnGAPIOptFlowLK(OptFlowLKTestInput<Type>& in,
+                                      int width, int height,
+                                      const OptFlowLKTestParams& params,
+                                      OptFlowLKTestOutput& ocvOut,
+                                      OptFlowLKTestOutput& gapiOut)
+{
+
+    int nPointsX = 0, nPointsY = 0;
+    std::tie(nPointsX, nPointsY) = params.pointsNum;
+
+    initTrackingPointsArray(in.prevPoints, width, height, nPointsX, nPointsY);
+
+    cv::Size winSize(params.winSize, params.winSize);
+
+    // OpenCV code /////////////////////////////////////////////////////////////
+    {
+        cv::calcOpticalFlowPyrLK(in.prevData, in.nextData, in.prevPoints,
+                                 ocvOut.nextPoints, ocvOut.statuses, ocvOut.errors,
+                                 winSize, params.maxLevel, params.criteria,
+                                 params.flags, params.minEigThreshold);
+    }
+
+    // G-API code //////////////////////////////////////////////////////////////
+    {
+        GType               inPrev,  inNext;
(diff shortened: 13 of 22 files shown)


OrestChura:
There are contents from https://github.com/opencv/opencv/pull/16717 used here; as far as that PR is merged, I rebase/cut all the unnecessary features

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -102,6 +109,20 @@ namespace imgproc {
...
+            return std::make_tuple(empty_array_desc(), empty_array_desc(), empty_array_desc());
+        }
+    };
+
+    G_TYPED_KERNEL(GCalcOptFlowPyrLK, <GArrOutLK(GArrayMat,GArrayMat,GArrayPoint2f,GArrayPoint2f,cv::Size,int,cv::TermCriteria,int,double)>,
+                   "org.opencv.imgproc.optFlowPyrLK") {
+        static std::tuple<GArrayDesc, GArrayDesc, GArrayDesc> outMeta(GArrayDesc, GArrayDesc, GArrayDesc, GArrayDesc, const cv::Size&, int, const cv::TermCriteria&, int, double) {
+            return std::make_tuple(empty_array_desc(), empty_array_desc(), empty_array_desc());
+        }
+    };
```
OrestChura: There are two different kernels (almost the same except input GMats/GArrays< GMat >) because cv::calcOptFlow doesn't call pyramid building if we pass to it vector< Mat >, so we can't just invent casting GMat -> GArray< GMat > without functionality lost

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
```diff
@@ -622,6 +641,203 @@ PERF_TEST_P_(CannyPerfTest, TestPerformance)
 
 //------------------------------------------------------------------------------
 
+PERF_TEST_P_(OptFlowLKPerfTest, TestPerformance)
```
OrestChura: Used https://github.com/opencv/opencv/blob/master/modules/video/perf/perf_optflowpyrlk.cpp as a basis/example

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
```diff
@@ -180,6 +180,26 @@ GAPI_OCV_KERNEL(GCPUCanny, cv::gapi::imgproc::GCanny)
     }
 };
 
+GAPI_OCV_KERNEL(GCPUCalcOptFlowLK, cv::gapi::imgproc::GCalcOptFlowLK)
+{
+    static void run(const cv::Mat& prevImg, const cv::Mat& nextImg, const std::vector<cv::Point2f>& prevPts, const std::vector<cv::Point2f>& predPts, const cv::Size& winSize, int maxLevel, const cv::TermCriteria& criteria, int flags, double minEigThresh, std::vector<cv::Point2f>& outPts, std::vector<uchar>& status, std::vector<double>& err)
+    {
+        if (flags & cv::OPTFLOW_USE_INITIAL_FLOW)
+            outPts = std::vector<cv::Point2f>(predPts);
+        cv::calcOpticalFlowPyrLK(prevImg, nextImg, prevPts, outPts, status, err, winSize, maxLevel, criteria, flags, minEigThresh);
```
OrestChura: It is planned that outPts will be null-initialized if the condition isn't passed. Should this be done explicitly or outputs from kernels are initialized by nulls automatically?

OrestChura:
@AsyaPronina, @TolyaTalamanov, pls review

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
AsyaPronina: I think that outputs from kernels are not initialized with nulls automatically and you should initialized them explicitly.

AsyaPronina: @TolyaTalamanov , please correct me if I am wrong

modules/gapi/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ endif()
 
 set(the_description "OpenCV G-API Core Module")
 
-ocv_add_module(gapi opencv_imgproc)
+ocv_add_module(gapi opencv_imgproc opencv_video)
```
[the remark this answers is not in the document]
OrestChura: [cv::calcOpticalFlowPyrLK()](https://docs.opencv.org/4.2.0/dc/d6b/group__video__track.html#ga473e4b886d0bcc6b65831eb88ed93323) function belongs to video module. Should video be linked therefore?

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -102,6 +109,20 @@ namespace imgproc {
         }
     };
 
+    G_TYPED_KERNEL(GCalcOptFlowLK, <GArrOutLK(GMat,GMat,GArrayPoint2f,GArrayPoint2f,cv::Size,int,cv::TermCriteria,int,double)>,
```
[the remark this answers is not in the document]
OrestChura: As I can see, Andrey unified these classes to avoid different macros usage: https://github.com/opencv/opencv/pull/16496. So, is it necessary to use _M modification?

modules/gapi/CMakeLists.txt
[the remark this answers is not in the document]
AsyaPronina: If so, yes, it should.

modules/gapi/include/opencv2/gapi/imgproc.hpp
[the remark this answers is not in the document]
AsyaPronina: Ough, okay, you are right, there is no reason to use _M modification.

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -30,6 +31,12 @@ namespace imgproc {
     using GMat2 = std::tuple<GMat,GMat>;
     using GMat3 = std::tuple<GMat,GMat,GMat>; // FIXME: how to avoid this?
 
+    using GArrayPoint2f = cv::GArray<cv::Point2f>;
+    using GArrayUchar   = cv::GArray<uchar>;
+    using GArrayFloat  = cv::GArray<float>;
+    using GArrayMat     = cv::GArray<cv::GMat>;
```
OrestChura: GArrayGMat

modules/gapi/include/opencv2/gapi/imgproc.hpp
OrestChura: cv::Size -> Size, etc.

modules/gapi/perf/cpu/gapi_imgproc_perf_tests_cpu.cpp
```diff
@@ -134,6 +134,31 @@ INSTANTIATE_TEST_CASE_P(CannyPerfTestCPU, CannyPerfTest,
         Values(true, false),
         Values(cv::compile_args(IMGPROC_CPU))));
 
+INSTANTIATE_TEST_CASE_P(OptFlowLKPerfTestCPU, OptFlowLKPerfTest,
```
AsyaPronina: All stuff you've added here should be moved to the separate file which is dedicated to test G-API operations for OpenCV *video* module on CPU : modules/gapi/perf/cpu/gapi_video_perf_tests_cpu.cpp (which you should create).

modules/gapi/src/api/kernels_imgproc.cpp
```diff
@@ -97,6 +97,22 @@ GMat Canny(const GMat& src, double thr1, double thr2, int apertureSize, bool l2g
     return imgproc::GCanny::on(src, thr1, thr2, apertureSize, l2gradient);
 }
 
+std::tuple<GArray<Point2f>, GArray<uchar>, GArray<float>> calcOpticalFlowPyrLK(const GMat& prevImg,
```
AsyaPronina: All stuff you've added here should be moved to the separate file which is dedicated for definition of G-API operations wrappers for OpenCV video module functions:  modules/gapi/src/api/kernels_video.cpp (which you should create).

modules/gapi/src/api/kernels_imgproc.cpp
```diff
@@ -97,6 +97,22 @@ GMat Canny(const GMat& src, double thr1, double thr2, int apertureSize, bool l2g
...
+    const GMat& nextImg, const GArray<Point2f>& prevPts, const GArray<Point2f>& predPts, const Size& winSize,
+    int maxLevel, const TermCriteria& criteria, int flags, double minEigThresh)
```
AsyaPronina: Please replace return type and all GArray<*> types with the types you defined in modules/gapi/include/opencv2/gapi/imgproc.hpp.

modules/gapi/src/api/kernels_imgproc.cpp
```diff
@@ -97,6 +97,22 @@ GMat Canny(const GMat& src, double thr1, double thr2, int apertureSize, bool l2g
...
+{
+    return imgproc::GCalcOptFlowLK::on(prevImg, nextImg, prevPts, predPts, winSize, maxLevel,
+                                          criteria, flags, minEigThresh);
+}
+
+std::tuple<GArray<Point2f>, GArray<uchar>, GArray<float>> calcOpticalFlowPyrLK(const GArray<GMat>& prevPyr,
+    const GArray<GMat>& nextPyr, const GArray<Point2f>& prevPts, const GArray<Point2f>& predPts, const Size& winSize,
+    int maxLevel, const TermCriteria& criteria, int flags, double minEigThresh)
```
AsyaPronina: Please replace return type and all GArray<*> types with the types you defined in modules/gapi/include/opencv2/gapi/imgproc.hpp.

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
```diff
@@ -180,6 +180,26 @@ GAPI_OCV_KERNEL(GCPUCanny, cv::gapi::imgproc::GCanny)
     }
 };
 
+GAPI_OCV_KERNEL(GCPUCalcOptFlowLK, cv::gapi::imgproc::GCalcOptFlowLK)
```
AsyaPronina: All stuff you've added here should be moved to the separate file which is dedicated for G-API kernels on CPU for OpenCV video module functions: modules/gapi/src/backends/cpu/gcpuvideo.cpp (which you should create).

modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
```diff
@@ -50,6 +50,25 @@ namespace
             rgb2yuyv(in_line_p, out_line_p, in.cols);
         }
     }
+
+    void FormTrackingPointsArray(vector<Point2f>& points, int width, int height, int nPointsX, int nPointsY)
```
AsyaPronina: Please name the function from the lower case letter.
Can this function somehow be shared between accuracy and performance tests?

modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
```diff
@@ -380,6 +399,152 @@ TEST_P(CannyTest, AccuracyTest)
     }
 }
 
+TEST_P(OptFlowLKTest, AccuracyTest)
```
AsyaPronina: This test duplicates 50% of code of the performance test. Can common stuff be somehow shared between?
May be initialization function can made as a test parameter and be shared among the instantiations and so on?

modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
```diff
@@ -380,6 +399,152 @@ TEST_P(CannyTest, AccuracyTest)
...
+    }
+    // Comparison //////////////////////////////////////////////////////////////
+    {
+        EXPECT_TRUE(std::get<0>(cmpFs)(out_vec_pts_gapi, out_vec_pts_ocv) &&
+                    std::get<1>(cmpFs)(out_vec_status_gapi, out_vec_status_ocv) &&
+                    std::get<2>(cmpFs)(out_vec_err_gapi, out_vec_err_ocv));
+    }
+}
+
+TEST_P(OptFlowPyrLKTest, AccuracyTest)
```
AsyaPronina: The same comment about code duplicate.

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -642,6 +663,66 @@ L2gradient=false ).
...
+     minEigThreshold description); if the flag is not set, then L1 distance between patches
+     around the original and a moved point, divided by number of pixels in a window, is used as a
+     error measure.
+@param minEigThresh the algorithm calculates the minimum eigen value of a 2x2 normal matrix of
+optical flow equations (this matrix is called a spatial gradient matrix in @cite Bouguet00), divided
+by number of pixels in a window; if this value is less than minEigThreshold, then a corresponding
+feature is filtered out and its flow is not processed, so it allows to remove bad points and get a
+performance boost.
+
+@note Function returns back:
```
OrestChura: @return instead

modules/gapi/test/common/gapi_tests_common.hpp
```diff
@@ -272,6 +350,13 @@ struct TestWithParamBase : TestFunctional,
         Fixture() { InitF(type, sz, dtype); } \
     };
 
+#define GAPI_TEST_FIXTURE_NO_INIT(Fixture, API, Number, ...) \
```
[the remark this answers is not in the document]
AsyaPronina: But if you want to get rid of not-always-needed "common parameters" also (as I got from your explanations), so this is a good feature to add. But I would suggest to think on a more better way to implement such logic with avoiding the reproduction of existing classes. May be we can somehow change existing classes to make them scalable to support such feature? If no, or have no time for it, then I am ok with changes but I would suggest to rename macro and all accompanying stuff to make them reflect the new feature purpose.

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
OrestChura: As all the output vectors are initialized inside the function cv::calcOpticalFlowPyrLK(), it is not necessary here

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -102,6 +109,20 @@ namespace imgproc {
         }
     };
 
+    G_TYPED_KERNEL(GCalcOptFlowLK, <GArrOutLK(GMat,GMat,GArrayPoint2f,GArrayPoint2f,cv::Size,int,cv::TermCriteria,int,double)>,
+                   "org.opencv.imgproc.optFlowLK") {
+        static std::tuple<GArrayDesc, GArrayDesc, GArrayDesc> outMeta(GMatDesc, GMatDesc, GArrayDesc, GArrayDesc, const cv::Size&, int, const cv::TermCriteria&, int, double) {
```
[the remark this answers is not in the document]
AsyaPronina: Sorry, it is not enum, so it worth to preserve const &

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -102,6 +109,20 @@ namespace imgproc {
...
+            return std::make_tuple(empty_array_desc(), empty_array_desc(), empty_array_desc());
+        }
+    };
+
+    G_TYPED_KERNEL(GCalcOptFlowPyrLK, <GArrOutLK(GArrayMat,GArrayMat,GArrayPoint2f,GArrayPoint2f,cv::Size,int,cv::TermCriteria,int,double)>,
+                   "org.opencv.imgproc.optFlowPyrLK") {
+        static std::tuple<GArrayDesc, GArrayDesc, GArrayDesc> outMeta(GArrayDesc, GArrayDesc, GArrayDesc, GArrayDesc, const cv::Size&, int, const cv::TermCriteria&, int, double) {
```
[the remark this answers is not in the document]
AsyaPronina: Sorry, it is not enum, so it worth to preserve const &

modules/gapi/include/opencv2/gapi/imgproc.hpp
OrestChura: fixed

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
```diff
@@ -622,6 +641,201 @@ PERF_TEST_P_(CannyPerfTest, TestPerformance)
...
+        default:
+            FAIL() << "Unexpected number of channels: " << channels;
+    }
+
+    int nPointsX = std::min(std::get<0>(nPoints), in_mat1.cols);
+    int nPointsY = std::min(std::get<1>(nPoints), in_mat1.rows);
+
+    std::vector<cv::Point2f> in_vec_pts, out_vec_pts_ocv, out_vec_pts_gapi;
+    std::vector<uchar> out_vec_status_ocv, out_vec_status_gapi;
+    std::vector<float> out_vec_err_ocv, out_vec_err_gapi;
```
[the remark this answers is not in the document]
OrestChura: got it, thank you:)

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -642,6 +663,66 @@ L2gradient=false ).
...
+ -  output vector of errors; each element of the vector is set to an error for the
+corresponding feature, type of the error measure can be set in flags parameter; if the flow wasn't
+found then the error is not defined (use the status parameter to find such cases).
+ */
+GAPI_EXPORTS std::tuple<GArray<Point2f>, GArray<uchar>, GArray<float>> calcOpticalFlowPyrLK(
+    const GMat& prevImg, const GMat& nextImg, const GArray<Point2f>& prevPts,
+    const GArray<Point2f>& predPts = GArray<Point2f>(), const Size& winSize = Size(21, 21),
+    int maxLevel = 3,
+    const TermCriteria& criteria = TermCriteria(TermCriteria::COUNT + TermCriteria::EPS, 30, 0.01),
+    int flags = 0, double minEigThresh = 1e-4);
```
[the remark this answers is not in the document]
AsyaPronina: Discussed: Orest convinced that changes are not needed because these "typedefs" will be reflected in documentation otherwise.

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -642,6 +663,66 @@ L2gradient=false ).
...
+
+GAPI_EXPORTS std::tuple<GArray<Point2f>, GArray<uchar>, GArray<float>> calcOpticalFlowPyrLK(
+    const GArray<GMat>& prevPyr, const GArray<GMat>& nextPyr, const GArray<Point2f>& prevPts,
+    const GArray<Point2f>& predPts = GArray<Point2f>(), const Size& winSize = Size(21, 21),
+    int maxLevel = 3,
+    const TermCriteria& criteria = TermCriteria(TermCriteria::COUNT + TermCriteria::EPS, 30, 0.01),
+    int flags = 0, double minEigThresh = 1e-4);
```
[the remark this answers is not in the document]
AsyaPronina: Discussed: Orest convinced that changes are not needed because these "typedefs" will be reflected in documentation otherwise.

AsyaPronina:
Please rebase on the latest master.

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -253,6 +260,35 @@ namespace imgproc {
...
 }
 
+// TODO: move to the separate file modules/gapi/include/opencv2/gapi/video.hpp
+namespace  video {
+    using GPoint2fArray        = cv::GArray<cv::Point2f>;
+    using GUcharArray          = cv::GArray<uchar>;
+    using GFloatArray          = cv::GArray<float>;
+    using GGMatArray           = cv::GArray<cv::GMat>;
+
+    using GOptFlowLKOutput     = std::tuple<GPoint2fArray, GUcharArray, GFloatArray>;
```
dmatveev: Can out G_TYPED_KERNEL macro work without this?

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -253,6 +260,35 @@ namespace imgproc {
...
+    using GOptFlowLKOutputDesc = std::tuple<GArrayDesc, GArrayDesc, GArrayDesc>;
+
+    G_TYPED_KERNEL(GCalcOptFlowLK, <GOptFlowLKOutput(GMat,GMat,GPoint2fArray,GPoint2fArray,
+                                                     Size,int,TermCriteria,int,double)>,
+                   "org.opencv.video.calcOpticalFlowLK") {
```
dmatveev: let's not confuse people, it always works with Pyramids and the OpenCV kernel is always `-PyrLK`.

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -943,6 +979,73 @@ image type is @ref CV_8UC1.
...
+
+// TODO: move to the separate file modules/gapi/include/opencv2/gapi/video.hpp
+//! @addtogroup gapi_video
+//! @{
+/** @brief Calculates an optical flow for a sparse feature set using the iterative Lucas-Kanade
+method with pyramids.
+
+See @cite Bouguet00 . The function is parallelized with the TBB library.
+
+@note Function textual ID is "org.opencv.video.calcOpticalFlowLK"
```
dmatveev: Needs to be fixed once you address the kernel name problem

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -943,6 +979,73 @@ image type is @ref CV_8UC1.
...
 
 //! @} gapi_colorconvert
+
+// TODO: move to the separate file modules/gapi/include/opencv2/gapi/video.hpp
+//! @addtogroup gapi_video
+//! @{
+/** @brief Calculates an optical flow for a sparse feature set using the iterative Lucas-Kanade
+method with pyramids.
+
+See @cite Bouguet00 . The function is parallelized with the TBB library.
```
dmatveev: This TBB node applies to kernel but not the Operation. What if we run this on GPU? There's no TBB then.

While in OpenCV there's almost a clear link between the API function and its implementation, in G-API an Operation (what you define here) may be implemented in a dozen of possible ways (and for a dozen of possible targets).

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -943,6 +979,73 @@ image type is @ref CV_8UC1.
...
+corresponding feature, type of the error measure can be set in flags parameter; if the flow wasn't
+found then the error is not defined (use the status parameter to find such cases).
+ */
+GAPI_EXPORTS std::tuple<GArray<Point2f>, GArray<uchar>, GArray<float>> calcOpticalFlowPyrLK(
+        const GMat& prevImg, const GMat& nextImg,
+        const GArray<Point2f>& prevPts, const GArray<Point2f>& predPts,
+        const Size& winSize = Size(21, 21), int maxLevel = 3,
+        const TermCriteria& criteria = TermCriteria(TermCriteria::COUNT + TermCriteria::EPS,
+                                                    30, 0.01),
+        int flags = 0, double minEigThresh = 1e-4);
```
dmatveev: I'd align this code a different way. There's lots of parameters, so why not make every parameter standing on its own line?

```c++
GAPI_EXPORTS std::tuple<GArray<Point2f>, GArray<uchar>, GArray<float>>
calcOpticalFlowPyrLK(const GMat& prevImg,
                     const GMat& nextImg,
                     ...);
```

modules/gapi/perf/common/gapi_imgproc_perf_tests.hpp
```diff
@@ -46,5 +46,22 @@ class YUV2BGRPerfTest : public TestPerfParams<tuple<compare_f, cv::Size, cv::GCo
 class RGB2HSVPerfTest : public TestPerfParams<tuple<compare_f, cv::Size, cv::GCompileArgs>> {};
 class BayerGR2RGBPerfTest : public TestPerfParams<tuple<compare_f, cv::Size, cv::GCompileArgs>> {};
 class RGB2YUV422PerfTest  : public TestPerfParams<tuple<compare_f, cv::Size, cv::GCompileArgs>> {};
-}
+} // opencv_test
+
+// TODO: move to the separate file modules/gapi/perf/common/gapi_video_perf_tests.hpp
```
dmatveev: Why not do that here

modules/gapi/perf/common/gapi_imgproc_perf_tests.hpp
```diff
@@ -46,5 +46,22 @@ class YUV2BGRPerfTest : public TestPerfParams<tuple<compare_f, cv::Size, cv::GCo
...
+#include "../../test/common/gapi_video_tests_common.hpp"
+namespace opencv_test
+{
+class OptFlowLKPerfTest : public TestPerfParams<tuple<tuple<compare_vector_f<Point2f>,
+                                                            compare_vector_f<uchar>,
+                                                            compare_vector_f<float>>,
```
dmatveev: I am not sure if we need three different comparison functions. In the end, it is still a result of a single operation.

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
```diff
@@ -1072,5 +1072,91 @@ PERF_TEST_P_(RGB2YUV422PerfTest, TestPerformance)
...
+    auto outTupleOCV  = std::make_tuple(std::ref(outPtsOCV), std::ref(outStatusOCV),
+                                        std::ref(outErrOCV));
+    auto outTupleGAPI = std::make_tuple(std::ref(outPtsGAPI), std::ref(outStatusGAPI),
+                                        std::ref(outErrGAPI));
+
+    cv::GComputation c = runOCVnGAPIOptFlowLK(*this, { get<1>(GetParam()), format,
+                                                       get<2>(GetParam()), get<3>(GetParam()),
+                                                       get<4>(GetParam()), maxLevel,
+                                                       get<5>(GetParam()), flags, minEigThreshold,
+                                                       get<6>(GetParam()) },
```
dmatveev: I recommend to use `tie()` instead. It is more type-safe than this.

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
```diff
@@ -26,6 +26,12 @@ namespace {
         }
         return in;
     }
+
+// TODO: move to the separate file modules/gapi/src/backends/cpu/gcpuvideo.cpp
+    using Point2fVector = std::vector<cv::Point2f>;
+    using UcharVector   = std::vector<uchar>;
+    using FloatVector   = std::vector<float>;
+    using MatVector     = std::vector<cv::Mat>;
```
dmatveev: Not sure if this is required here.

modules/gapi/test/common/gapi_imgproc_tests.hpp
```diff
@@ -72,4 +72,25 @@ GAPI_TEST_FIXTURE(BayerGR2RGBTest, initMatrixRandN, FIXTURE_API(CompareMats), 1,
 GAPI_TEST_FIXTURE(RGB2YUV422Test, initMatrixRandN, FIXTURE_API(CompareMats), 1, cmpF)
 } // opencv_test
 
+// TODO: move to the separate file modules/gapi/test/common/gapi_video_tests.hpp
```
dmatveev: Its time!

modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
```diff
@@ -757,6 +757,59 @@ TEST_P(RGB2YUV422Test, AccuracyTest)
...
+
+    auto outTupleOCV  = std::make_tuple(std::ref(outPtsOCV), std::ref(outStatusOCV),
+                                        std::ref(outErrOCV));
+    auto outTupleGAPI = std::make_tuple(std::ref(outPtsGAPI), std::ref(outStatusGAPI),
+                                        std::ref(outErrGAPI));
+
+    runOCVnGAPIOptFlowLK(*this, { fileNamePattern, format, channels, pointsNum,
+                                  winSize, maxLevel, criteria, flags, minEigThreshold,
+                                  getCompileArgs() },
+                         inPts, outTupleOCV, outTupleGAPI);
```
dmatveev: I don't quite get why do we need this, btw.

modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
```diff
@@ -757,6 +757,59 @@ TEST_P(RGB2YUV422Test, AccuracyTest)
...
+
+    {
+        EXPECT_TRUE(std::get<0>(cmpFs)(outPtsGAPI,    outPtsOCV)    &&
+                    std::get<1>(cmpFs)(outStatusGAPI, outStatusOCV) &&
+                    std::get<2>(cmpFs)(outErrGAPI,    outErrOCV));
```
dmatveev: The failure message will be very informative. :)

modules/gapi/test/common/gapi_tests_common.hpp
```diff
@@ -155,12 +155,58 @@ class TestFunctional
         }
     }
 
+    void initMatsFromImages(int channels, const std::string& fileNamePattern, int format)
+    {
+        initTestDataPath();
+        switch (channels)
+        {
+            case 1:
+                in_mat1 = cv::imread(perf::TestBase::getDataPath(cv::format(fileNamePattern.c_str(),
```
dmatveev: sometimes `using` really helps

modules/gapi/test/common/gapi_tests_common.hpp
```diff
@@ -155,12 +155,58 @@ class TestFunctional
...
+        {
+            // Since G-API has no own test data (yet), it is taken from the common space
+            const char* testDataPath = getenv("OPENCV_TEST_DATA_PATH");
+            GAPI_Assert(testDataPath != nullptr &&
+            "OPENCV_TEST_DATA_PATH environment variable is either not set or set incorrectly.");
+
+            cvtest::addDataSearchPath(testDataPath);
+            initialized = true;
+        }
+#endif // WINRT
```
dmatveev: So now we need people to have `opencv_extra` repo available by default. Not sure if this is a good idea. 99% of our tests don't need that.

modules/gapi/test/common/gapi_video_tests_common.hpp
```diff
@@ -0,0 +1,179 @@
...
+    int format;
+    int channels;
+    const std::tuple<size_t,size_t>& pointsNum;
+    int winSize;
+    int maxLevel;
+    const cv::TermCriteria& criteria;
+    int flags;
+    double minEigThreshold;
+    const cv::GCompileArgs& compileArgs;
+};
```
dmatveev: KlocWork may be unhappy without the  default initializers.

modules/gapi/test/common/gapi_video_tests_common.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+    OptFlowLKTestParams updatedTestParams(testParams);
+    updatedTestParams.maxLevel = maxLevel;
+
+    return runOCVnGAPIOptFlowLK<cgv::GGMatArray>(static_cast<size_t>(testInst.in_mat1.cols),
+                                                 static_cast<size_t>(testInst.in_mat1.rows),
+                                                 updatedTestParams,
+                                                 inTuple, outOCVTuple, outGAPITuple);
+}
+} // namespace
```
dmatveev: I believe this code may be more compact. I assume we save already by reusing this function in every test, but still this function itself looks scary.

modules/gapi/include/opencv2/gapi/imgproc.hpp
OrestChura: What do you mean? There shouldn't be this _using_?

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
OrestChura: I can't agree: if we use tie() here, we will have to predefine all the variables, but actually we don't really need them

modules/gapi/CMakeLists.txt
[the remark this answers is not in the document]
alalek:
```
ocv_add_module(gapi
    REQUIRED
      opencv_core
      opencv_imgproc
    OPTIONAL
      opencv_video
)
```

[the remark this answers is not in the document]
dmatveev: OpenCV already have this :) cool

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
dmatveev: Explained offline

dmatveev:
Any updates here ready for review?

modules/gapi/CMakeLists.txt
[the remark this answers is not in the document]
OrestChura: Thank you! Applied

modules/gapi/include/opencv2/gapi/cpu/imgproc.hpp
```diff
@@ -20,6 +20,15 @@ GAPI_EXPORTS GKernelPackage kernels();
 
 } // namespace cpu
 } // namespace imgproc
+
+// TODO: move to the separate file modules/gapi/include/opencv2/gapi/cpu/video.hpp
```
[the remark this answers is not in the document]
OrestChura: OK

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -10,6 +10,9 @@
 
 #include <opencv2/imgproc.hpp>
 
+// TODO:  move to the separate file modules/gapi/include/opencv2/gapi/video.hpp
+#include <opencv2/video.hpp>
```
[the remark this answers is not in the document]
OrestChura: All includes are checked to minimize dependency

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -253,6 +260,35 @@ namespace imgproc {
...
 
+// TODO: move to the separate file modules/gapi/include/opencv2/gapi/video.hpp
+namespace  video {
+    using GPoint2fArray        = cv::GArray<cv::Point2f>;
+    using GUcharArray          = cv::GArray<uchar>;
+    using GFloatArray          = cv::GArray<float>;
+    using GGMatArray           = cv::GArray<cv::GMat>;
+
+    using GOptFlowLKOutput     = std::tuple<GPoint2fArray, GUcharArray, GFloatArray>;
+    using GOptFlowLKOutputDesc = std::tuple<GArrayDesc, GArrayDesc, GArrayDesc>;
```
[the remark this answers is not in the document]
OrestChura: Agree :)

modules/gapi/include/opencv2/gapi/imgproc.hpp
OrestChura: Leaved as it was to reduse the use of long tuples in some places

modules/gapi/include/opencv2/gapi/imgproc.hpp
OrestChura: ok

modules/gapi/include/opencv2/gapi/imgproc.hpp
OrestChura: ok

modules/gapi/include/opencv2/gapi/imgproc.hpp
OrestChura: Agree, cut it

modules/gapi/include/opencv2/gapi/imgproc.hpp
OrestChura: aligned here and in some other places

modules/gapi/perf/common/gapi_imgproc_perf_tests.hpp
OrestChura: implemented, I hope it looks good

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
OrestChura: Removed

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
OrestChura: OK, the use of get<> is removed

modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
OrestChura: Agree, my bad

modules/gapi/test/common/gapi_tests_common.hpp
OrestChura: _lambda_ used, I hope it looks good enough

modules/gapi/test/common/gapi_tests_common.hpp
OrestChura: The function isn't called by default, only by user. Should I leave it as is or write a comment to use it wisely not to create problems?

modules/gapi/test/common/gapi_video_tests_common.hpp
OrestChura: default initializations added; for some variables in functions' bodies, too

modules/gapi/test/common/gapi_video_tests_common.hpp
OrestChura: A try has been done.

modules/gapi/perf/common/gapi_video_perf_tests_inl.hpp
```diff
@@ -0,0 +1,105 @@
...
+#include "gapi_video_perf_tests.hpp"
+
+namespace opencv_test
+{
+
+  using namespace perf;
+
+//------------------------------------------------------------------------------
+
+#ifdef HAVE_OPENCV_VIDEO
```
dmatveev, line 21: Does the body of this file really depend on the OPENCV_VIDEO?

As far as I can see, it is not

modules/gapi/perf/cpu/gapi_video_perf_tests_cpu.cpp
```diff
@@ -0,0 +1,57 @@
...
+                        Combine(Values(AbsExactForOptFlow().to_compare_f()),
+                                Values("cv/optflow/rock_%01d.bmp"),
+                                Values(1),
+                                Values(std::make_tuple(10, 10)),
+                                Values(15),
+                                Values(cv::TermCriteria(cv::TermCriteria::COUNT |
+                                                        cv::TermCriteria::EPS,
+                                                        21, 0.05)),
+                                Values(true),
+                                Values(cv::compile_args(VIDEO_CPU))));
```
dmatveev, line 56: And this code actually depends on the presense of OPENCV_VIDEO module.

If this module is not built, these tests shouldn't be instantiated (or disabled with a macro as @Volskig did that)

Here I see tests are instantiated always so in some (pretty weird) configurations tests will fail.

modules/gapi/test/common/gapi_tests_common.hpp
```diff
@@ -155,20 +155,60 @@ class TestFunctional
...
+                break;
+            case 4:
+                cvtColor(cv::imread(getDataPath(cv::format(fileNamePattern.c_str(), format))),
+                         in_mat1, cv::COLOR_BGR2BGRA);
+                cvtColor(cv::imread(getDataPath(cv::format(fileNamePattern.c_str(), format + 1))),
+                         in_mat2, cv::COLOR_BGR2BGRA);
+                break;
+            default:
+                FAIL() << "Unexpected number of channels: " << channels;
+        }
```
dmatveev, line 185: This code can be simpler:
```c++
GAPI_Assert(channels == 1 || channels == 3 || channels == 4);
const int flags = (channels == 1) ? cv::IMREAD_GRAYSCALE : cv::IMREAD_COLOR;
cv::Mat m1 = cv::imread(getDataPath(cv::format(fileNamePattern.c_str(), format);
cv::Mat m2 = cv::imread(getDataPath(cv::format(fileNamePattern.c_str(), format + 1);
if (channels == 4) {
    cvtColor(m1, in_mat1, cv::COLOR_BGR2BGRA);
    cvtColor(m2, in_mat2, cv::COLOR_BGR2BGRA);
} else {
    std::tie(in_mat1, in_mat2) = std::make_tuple(m1, m2);
}
```

BTW why do we need this `format + 1` thing?

modules/gapi/test/common/gapi_tests_common.hpp
dmatveev: Better to move this code out of here

modules/gapi/test/common/gapi_video_tests_common.hpp
dmatveev: `int flags = -1`;

modules/gapi/test/common/gapi_video_tests_common.hpp
```diff
@@ -0,0 +1,239 @@
...
+        break;
+    };
+
+    return os << criteria.maxCount << "; " << criteria.epsilon <<"}";
+}
+} // namespace
+
+#ifdef HAVE_OPENCV_VIDEO
+
+#include <opencv2/video.hpp>
```
dmatveev, line 134: Better to have all includes written at the top of the file (ok, with an extra ifdef)

modules/gapi/test/common/gapi_video_tests_common.hpp
```diff
@@ -0,0 +1,239 @@
...
+    OptFlowLKTestParams updatedParams(params);
+    updatedParams.maxLevel = cv::buildOpticalFlowPyramid(testInst.in_mat1, in.prevData,
+                                                         winSize, params.maxLevel, withDeriv);
+    updatedParams.maxLevel = cv::buildOpticalFlowPyramid(testInst.in_mat2, in.nextData,
+                                                         winSize, params.maxLevel, withDeriv);
+
+
+    return runOCVnGAPIOptFlowLK<cv::GArray<cv::GMat>>(in,
+                                                      static_cast<size_t>(testInst.in_mat1.cols),
+                                                      static_cast<size_t>(testInst.in_mat1.rows),
```
dmatveev, line 229: either use checked_cast from ADE or make parameters integers.

modules/gapi/test/common/gapi_video_tests_common.hpp
OrestChura: The `flags` variable is not reinitialized by tests as it's value is 0 for all the tests (it is not a test parameter). So it should always be equal to 0. It is just a default parameter flags=0 for `calcOpticalFlow()`.

modules/gapi/perf/cpu/gapi_video_perf_tests_cpu.cpp
```diff
@@ -0,0 +1,57 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2018 Intel Corporation
```
AsyaPronina, line 5: 2020

modules/gapi/test/common/gapi_tests_common.hpp
```diff
@@ -297,6 +401,18 @@ struct CompareF
 
 using CompareMats = CompareF<cv::Mat, cv::Mat>;
 using CompareScalars = CompareF<cv::Scalar, cv::Scalar>;
+struct CompareForOptFlow {};
```
AsyaPronina, line 404: I think this function shouldn't behere and should be in gapi_video_tests_common.hpp

modules/gapi/test/common/gapi_video_tests.cpp
```diff
@@ -0,0 +1,9 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2018 Intel Corporation
```
AsyaPronina, line 5: 2020

modules/gapi/test/common/gapi_tests_common.hpp
OrestChura: Unfortunately, the following structure uses this struct in its constructor so I have to declare this here

OrestChura: But I think I've figured out how to handle it. Will change this

modules/gapi/test/common/gapi_tests_common.hpp
OrestChura: Agree with the simplification, it's nice. Thanks!
Formatting is convenient here: it is supposed to that we are loading two files with the same names but the number in them like "optFlow01.jpg" and "optFlow02.jpg".
Ex., two instances of the test can take `fileNamePattern` = "cv/optflow/rock_%01d.bmp" and `format` = 1, `format` = 2. Processing the first instance, we initialize `in_mat1` and `in_mat2` with "cv/optflow/rock_1.bmp" and "cv/optflow/rock_2.bmp". Processing the second one, they are "cv/optflow/rock_2.bmp" and "cv/optflow/rock_3.bmp".
It can be useful

dmatveev: Ok, I see it now.

modules/gapi/test/common/gapi_tests_common.hpp
OrestChura: Will it be better if I move it to the unnamed namespace in this file (str. 22)? I need this function not only in tests of `video::` kernels, but also in tests `imgproc::goodFeatures`, so I want to preserve the function in this file

modules/gapi/test/common/gapi_video_tests_common.hpp
OrestChura: Is it good to link  perf_tests  with `ade` to use this?

modules/gapi/perf/common/gapi_video_perf_tests_inl.hpp
OrestChura: Yes, it does, but indirectly: the function `runOCVnGAPIOptFlowLK()` from `gapi_video_tests_common.hpp` is not defined when `opencv_video` is disabled. Should I add the definition of it (`runOCVnGAPIOptFlowLK(...) {}`) if `opencv_video` is disabled?

OrestChura: Pushed my option

modules/gapi/perf/cpu/gapi_video_perf_tests_cpu.cpp
OrestChura: disabled by #ifdef macro (DISABLED_ would turn off it always)

modules/gapi/test/common/gapi_video_tests_common.hpp
OrestChura: fixed

modules/gapi/perf/cpu/gapi_video_perf_tests_cpu.cpp
OrestChura: fixed

modules/gapi/test/common/gapi_tests_common.hpp
OrestChura: fixed

modules/gapi/test/common/gapi_video_tests.cpp
OrestChura: fixed

modules/gapi/test/common/gapi_tests_common.hpp
dmatveev: anonymous namespace in a header doesn't really solve _this_ problem (but for sure solves a number of others)

modules/gapi/test/common/gapi_video_tests_common.hpp
dmatveev: Then just put 0 there.

modules/gapi/perf/common/gapi_video_perf_tests_inl.hpp
dmatveev: Ok

modules/gapi/test/common/gapi_video_tests_common.hpp
dmatveev: Well, no, then better fix the types

modules/gapi/perf/common/gapi_video_perf_tests_inl.hpp
```diff
@@ -0,0 +1,96 @@
...
+    TEST_CYCLE()
+    {
+        c.apply(cv::gin(in_mat1, in_mat2, inPts, std::vector<cv::Point2f>{ }),
+                cv::gout(outPtsGAPI, outStatusGAPI, outErrGAPI));
+    }
+
+    // Comparison //////////////////////////////////////////////////////////////
+    EXPECT_TRUE(static_cast<compare_vector_forOptFlow<cv::Point2f>>(cmpF)(outPtsGAPI, outPtsOCV));
+    EXPECT_TRUE(static_cast<compare_vector_forOptFlow<uchar>>(cmpF)(outStatusGAPI, outStatusOCV));
+    EXPECT_TRUE(static_cast<compare_vector_forOptFlow<float>>(cmpF)(outErrGAPI,    outErrOCV));
```
dmatveev, line 49: I don't really get what happens here.

You're casting `cmpF` to three different types??

Why do you have three comparison functions here?

dmatveev: GTest is composable and it doesn't need all its `EXPECT*`/`ASSERT*` macros be called from the test body.

You can call these macros in any function (e.g. your comparison one) and then call it from your test body.

EXPECT_TRUE is not very informative as it is, when arrays or images are compared.

modules/gapi/perf/cpu/gapi_video_perf_tests_cpu.cpp
dmatveev: The idea was to put `DISABLED_` optionally.

```c++
#ifdef HAVE_OPENCV_VIDEO
#define WITH_VIDEO(X) X
#else
#define WITH_VIDEO(X) DISABLED_##X 
#endif // HAVE_OPENCV_VIDEO

INSTANTIATE_TEST_CASE_P(WITH_VIDEO(OptFlowLKInternalPerfTestCPU), ...
```

modules/gapi/test/common/gapi_video_tests_common.hpp
dmatveev: ...and please initilize the other `int`/`size_t` fields.

modules/gapi/test/common/gapi_video_tests_common.hpp
```diff
@@ -0,0 +1,289 @@
...
+
+#else // !HAVE_OPENCV_VIDEO
+
+inline cv::GComputation runOCVnGAPIOptFlowLK(TestFunctional&,
+                                             std::vector<cv::Point2f>&,
+                                             const OptFlowLKTestParams&,
+                                             OptFlowLKTestOutput&,
+                                             OptFlowLKTestOutput&)
+{
+    return cv::GComputation(cv::GMat(), cv::GMat());
```
dmatveev, line 270: Just put an assert here. This function shouldn't be called in this configuration.

modules/gapi/test/common/gapi_video_tests_common.hpp
```diff
@@ -0,0 +1,289 @@
...
+}
+
+inline cv::GComputation runOCVnGAPIOptFlowLKForPyr(TestFunctional&,
+                                                   OptFlowLKTestInput<std::vector<cv::Mat>>&,
+                                                   const OptFlowLKTestParams&,
+                                                   bool,
+                                                   OptFlowLKTestOutput&,
+                                                   OptFlowLKTestOutput&)
+{
+    return cv::GComputation(cv::GMat(), cv::GMat());
```
dmatveev, line 280: Just put an assert here. This function shouldn't be called in this configuration.

dmatveev:
Almost!!! Almost...

modules/gapi/test/common/gapi_video_tests_common.hpp
OrestChura: done

modules/gapi/perf/cpu/gapi_video_perf_tests_cpu.cpp
OrestChura: Oh, now I get it. Done
Also anticipate questions: `INSTANTIATE_TEST_CASE_P()` macro concatenates the arguments in its body, therefore prescan to seek macros in arguments does not occur; that's why I have to use another macro which calls gtest-macro with macro in the first argument to expand it properly

modules/gapi/test/common/gapi_video_tests_common.hpp
OrestChura: as I can see it's impossible to fix the types totally: on the one hand, the function got `int`'s - from `Mat.cols` and `Mat.rows` members; on the other - `std::vector.resize()` takes `size_t`...
not to resize vector before filling in cycle is bad, I think
can we leave the only `static_cast` (on line 48 this file) and check Assert before it?

modules/gapi/perf/common/gapi_video_perf_tests_inl.hpp
OrestChura: I refused to cast cmpF, it looked weird; the comparison function which is called different templates is implemented instead.
As for EXPECT_TRUE, we are comparing on bitexact here, so I think it's OK not to be very informative..

modules/gapi/test/common/gapi_video_tests_common.hpp
OrestChura: I hope there wouldn't be an error

alalek:
Rebase is required to resolve merge conflicts.

modules/gapi/perf/cpu/gapi_video_perf_tests_cpu.cpp
dmatveev: Ok, it also works, thanks!

modules/gapi/perf/common/gapi_video_perf_tests_inl.hpp
dmatveev: So what is `cmpF` here and how do you cast it to the different kinds of functions?

Why not finally make `cmpF` of six arguments?

OrestChura: This place is rewritten now. Yes, you are right, I've probably over-complicated it. In the end, after Asya's review, I sorted it out and then implemented just what you've said - the function with all 6 vectors to compare (but it calls vector2vector compare function 3 times). Now it looks much better (for me)
Is it good just not to give cmpF to test instances as a parameter?  the case is too custom and also there is no need to check different comparisons (eg Exact and Tolerant ones) at the same time

modules/gapi/test/common/gapi_video_tests_common.hpp
```diff
@@ -0,0 +1,258 @@
...
+    double minEigThreshold        = 1e-4;
+    cv::GCompileArgs compileArgs;
+    int flags                     = 0;
+};
+} // namespace
+} // namespace opencv_test
+
+namespace
+{
+inline std::ostream& operator<<(std::ostream& os, const cv::TermCriteria& criteria)
```
AsyaPronina, line 95: May be it is better to move this operator at the bottom of the file?

modules/gapi/include/opencv2/gapi/cpu/video.hpp
```diff
@@ -0,0 +1,25 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+#ifndef OPENCV_GAPI_CPU_VIDEO_API_HPP
```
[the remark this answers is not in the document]
OrestChura: I wrote it based on an example in [modules/gapi/include/opencv2/gapi/cpu/imgproc.hpp](https://github.com/opencv/opencv/blob/master/modules/gapi/include/opencv2/gapi/cpu/imgproc.hpp)

modules/gapi/include/opencv2/gapi/video.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+#ifndef OPENCV_GAPI_VIDEO_HPP
+#define OPENCV_GAPI_VIDEO_HPP
+
+#include <utility> // std::tuple
+
+#include <opencv2/gapi/gkernel.hpp>
+
+
+/** \defgroup gapi_video G-API Optical Flow processing functionality
```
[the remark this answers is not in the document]
OrestChura: Yes, I forgot to fix it, it was just an initial version. Would be "G-API Video processing functionality"

modules/gapi/include/opencv2/gapi/video.hpp
```diff
@@ -0,0 +1,135 @@
...
+        static std::tuple<GArrayDesc,GArrayDesc,GArrayDesc> outMeta(GMatDesc,GMatDesc,GArrayDesc,
+                                                                    GArrayDesc,const Size&,int,
+                                                                    const TermCriteria&,int,double)
+        {
+            return std::make_tuple(empty_array_desc(), empty_array_desc(), empty_array_desc());
+        }
+
+    };
+
+    G_TYPED_KERNEL(GCalcOptFlowLKForPyr,
```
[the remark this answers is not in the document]
OrestChura: I don't think it is very important, I've just fixed the places which will be used and seen by users (documentation and string API), and this place is internal

modules/gapi/test/common/gapi_video_tests_common.hpp
OrestChura: Yes, and also there is the mess with namespaces, 'll fix it

modules/gapi/src/api/kernels_video.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include "precomp.hpp"
+
+#include <opencv2/gapi/video.hpp>
+
+namespace cv { namespace gapi {
+namespace
+{
+    using namespace video;
+} // namespace
```
alalek, line 16: Why is anonymous namespace required here?

modules/gapi/include/opencv2/gapi/video.hpp
```diff
@@ -0,0 +1,135 @@
...
+//! @{
+/** @brief Calculates an optical flow for a sparse feature set using the iterative Lucas-Kanade
+method with pyramids.
+
+See @cite Bouguet00 .
+
+@note Function textual ID is "org.opencv.video.calcOpticalFlowPyrLK"
+
+@param prevImg first 8-bit input image (GMat) or pyramid (GArray<GMat>) constructed by
+buildOpticalFlowPyramid.
```
alalek, line 65: > or pyramid (GArray<GMat>) constructed by buildOpticalFlowPyramid

Which one? I don't see `buildOpticalFlowPyramid` in GAPI API.

modules/gapi/test/common/gapi_tests_common.hpp
```diff
@@ -196,25 +196,32 @@ class TestFunctional
...
     }
 
+    void initMatsFromImages(int channels, const std::string& pattern, int imgNum)
+    {
+        initTestDataPath();
+        GAPI_Assert(channels == 1 || channels == 3 || channels == 4);
+        const int flags = (channels == 1) ? cv::IMREAD_GRAYSCALE : cv::IMREAD_COLOR;
+
+        cv::Mat m1 = cv::imread(findDataFile(cv::format(pattern.c_str(), imgNum)), flags);
+        cv::Mat m2 = cv::imread(findDataFile(cv::format(pattern.c_str(), imgNum + 1)), flags);
```
alalek, line 206: Avoid "cv::" prefixes in OpenCV code (if they are not required).

modules/gapi/test/test_precomp.hpp
```diff
@@ -16,8 +16,9 @@
 #include <opencv2/ts.hpp>
 
 #include <opencv2/gapi.hpp>
-#include <opencv2/gapi/imgproc.hpp>
 #include <opencv2/gapi/core.hpp>
+#include <opencv2/gapi/imgproc.hpp>
+#include <opencv2/gapi/video.hpp>
```
alalek, line 21: `#ifdef HAVE_OPENCV_VIDEO` is required here.

Try to build OpenCV with disabled video module:
```
cmake -DBUILD_opencv_video=OFF ...
```

modules/gapi/include/opencv2/gapi/video.hpp
OrestChura: This kernel will be implemented later based on cv::buildOpticalFlowPyramid(); although, it works even now if you build pyramids with cv:: function and give them to this kernel (as it is implemented in tests)

modules/gapi/test/common/gapi_tests_common.hpp
OrestChura: Is it a general rule? I thought this increases readability..

modules/gapi/test/test_precomp.hpp
OrestChura: Yeah, I've just checked it now (just in case, because I've already build it before): it compiles without opencv_video as there is a workaround on that case. There is no dependency on opencv_video in `opencv2/gapi/video.hpp` because it is a declaration of kernels, and the CPU(OpenCV) implementation (in [modules/gapi/src/backends/cpu/gcpuvideo.cpp](https://github.com/opencv/opencv/pull/16788/files#diff-baf3285f744f48afaa6774d88c1c4ba1) is protected by `#ifdef`

AsyaPronina: opencv2/gapi/video.hpp file doesn't contain information which depends on video module, as I got. It just contains G-API operations interfaces.

modules/gapi/src/api/kernels_video.cpp
OrestChura: Ih, it is a mistake, the file is .cpp so there is no need in this.. fixed

modules/gapi/test/test_precomp.hpp
alalek: Sorry, my mistake - read as `opencv2/video.hpp`

PR #16795, opened 2020-03-12
Added a sample file for qrcode detection in python

jshiwam:
<cut/>

```
```


samples/python/qrcode.py
```diff
@@ -0,0 +1,139 @@
...
+o_f_ext=''
+
+#Default values of input parameters
+parser = argparse.ArgumentParser(description='This program detects the QR-codes input images using OpenCV Library.')
+parser.add_argument('-i','--input',help="input image path (default input file path is 'opencv_extra/testdata/cv/qrcode/multiple/*_qrcodes.png",default="../../../opencv_extra/testdata/cv/qrcode/multiple/*_qrcodes.png",metavar="")
+parser.add_argument('-d','--detect',help="detect QR code only (skip decoding) (value True/False default value is False)",metavar='',type=str2bool,default=False)
+parser.add_argument('-m','--multi',help="use detect for multiple qr-codes (value True/False default value is True)",metavar="",type=str2bool,default=True)
+parser.add_argument('-o','--out',help="path to result file (default output filename is qr_code.png)",metavar="",default="qr_code.png")
+args = parser.parse_args()
+print(args)
```
alalek: "global" code is not allowed in OpenCV Python samples (see "demo.py", it loads these files as modules).
Move such code into the `main()` function.

samples/python/qrcode.py
```diff
@@ -0,0 +1,139 @@
...
+import numpy as np
+import sys
+from glob import glob
+PY3 = sys.version_info[0] == 3
+import argparse
+
+if PY3:
+    xrange = range
+
+def str2bool(v):
```
alalek: Use `action='store_true'` in `.add_argument()` instead. Refer to other code in OpenCV repository.

samples/python/qrcode.py
```diff
@@ -0,0 +1,139 @@
...
+        cv.imwrite(outfile,result)
+
+    print("Press any key to exit ...")
+    cv.waitKey(0)
+    print("Exit")
+
+def main(argv):
+    global o_f_name
+    global o_f_ext
+    if(args.out!=''):
```
alalek: `if` in Python doesn't require brackets

samples/python/qrcode.py
```diff
@@ -0,0 +1,139 @@
...
+
+def drawQRCodeResults(result,points,decode_info,fps):
+    n=len(points)
+    if(type(decode_info) is str):
+        decode_info=[decode_info]
+    if(n>0):
+        for i in range(n):
+            cnt=np.array(points[i]).reshape((-1,1,2)).astype(np.int32)
+            drawQRCodeContours(result,cnt)
+            print ('QR[',i,']@',cnt.reshape(1,-1).tolist(),": ",end="")
```
alalek: Please use "Pythonic" code. See "string format" examples: https://docs.python.org/2.7/library/string.html#format-examples (you may replace `.format` with ` % ` too - see `drawFPS()` above).

jshiwam: Done!

samples/python/qrcode.py
alalek: > Move such code into the main() function

Not resolved.

jshiwam: I have updated the changes please check

samples/python/qrcode.py
```diff
@@ -0,0 +1,157 @@
...
+        default="../../../opencv_extra/testdata/cv/qrcode/multiple/*_qrcodes.png",
+        metavar="")
+    parser.add_argument(
+        '-d',
+        '--detect',
+        help="detect QR code only (skip decoding) (default value is False)",
+        action='store_true')
+    parser.add_argument(
+        '-m',
+        '--multi',
```
asmorkalov, line 133: I propose to run detectMulti by default and inverse the option like for detection case. The QR code images in OpenCV extra contains several QR codes per image and the sample started with default options produces garbage.

samples/python/qrcode.py
```diff
@@ -0,0 +1,157 @@
...
+        description='This program detects the QR-codes input images using OpenCV Library.')
+    parser.add_argument(
+        '-i',
+        '--input',
+        help="input image path (default input file path is 'opencv_extra/testdata/cv/qrcode/multiple/*_qrcodes.png",
+        default="../../../opencv_extra/testdata/cv/qrcode/multiple/*_qrcodes.png",
+        metavar="")
+    parser.add_argument(
+        '-d',
+        '--detect',
```
asmorkalov, line 128: Sample executed with default options and just -d shows garbage as detection.
qr_garbadge

jshiwam: I have enabled the multiple detection by default and used the -m tag to disable the multiple detection so that single qrcodes can be detected and decoded. Now running on the opencv_extra dataset with default parameters doesn't give garbage values. 
Please let me know if some other changes are to be made in this.

alalek:
> I have enabled the multiple detection by default and used the -m tag to disable the multiple detection so that single qrcodes can be detected and decoded. 

Please align behavior with C++ QR code sample.

Code should preserve behavior and should do the same things.
Also critical parts of code itself should look similar.
We should minimize confusions in general.

jshiwam:
> > I have enabled the multiple detection by default and used the -m tag to disable the multiple detection so that single qrcodes can be detected and decoded.
> 
> Please align behavior with C++ QR code sample.
> 
> Code should preserve behavior and should do the same things.
> Also critical parts of code itself should look similar.
> We should minimize confusions in general.



> > I have enabled the multiple detection by default and used the -m tag to disable the multiple detection so that single qrcodes can be detected and decoded.
> 
> Please align behavior with C++ QR code sample.
> 
> Code should preserve behavior and should do the same things.
> Also critical parts of code itself should look similar.
> We should minimize confusions in general.

Now I have removed the ambiguity in behaviour and implemented the detection from camera as well.

PR #16800, opened 2020-03-12
opencv_version: dump available DNN targets

alalek:
**RFC**
<cut/>
**WIP**


mshabunin:
I think it would be better to load _dnn_ module dynamically or extract it to separate application, because otherwise it will block users from getting opencv configuration in some cases (e.g. with InferenceEngine).

alalek:
I thought about `opencv_version_dnn`, but it looks strange.

PR #16802, opened 2020-03-12
Scale_layer: enable scale layer to handle inputs in batch

### LINKED ISSUE #13626: dnn: assertion failure with multiple images
- OpenCV => 4.0.1-dev
- Compiler => mingw64 7.2.0
##### Detailed description
using the[ face_detector_uint8.pb](https://github.com/opencv/opencv_3rdparty/raw/8033c2bc31b3256f0d461c919ecc01c2428ca03b/opencv_face_detector_uint8.pb) model with multiple images (using dnn::blobFromImages), i get :
```
OpenCV(4.0.1-dev) Error: Assertion failed (start <= (int)shape.size() && end <= (int)shape.size() && start <= end) in total, file C:/p/opencv/modules/dnn/include/opencv
2/dnn/shape_utils.hpp, line 161
```

debugging it, i found, that in the 2nd ScaleLayer, the `endAxis` exceeds the size of the `inpShape`:

```
Thread 1 hit Breakpoint 9, cv::dnn::ScaleLayerImpl::forward (this=0xa5c1090, inputs_arr=..., outputs_arr=..., internals_arr=...)
    at C:\p\opencv\modules\dnn\src\layers\scale_layer.cpp:94
94              CV_Assert(total(inpShape, axis, endAxis) == numWeights);
(gdb) p inpShape
$16 = std::vector of length 4, capacity 4 = {3, 32, 24, 32}
(gdb) p axis
$17 = 0
(gdb) p endAxis
$18 = 5 // this is 2 with one image, and 5 with 3 of them !
```
thus the assert fails here: 

https://github.com/opencv/opencv/blob/7e2ebecd52232d02e2d53f3e073c076be73e3470/modules/dnn/include/opencv2/dnn/shape_utils.hpp#L160-L161

##### Steps to reproduce
    ```
    dnn::Net net = dnn::readNetFromTensorflow("opencv_face_detector_uint8.pb","opencv_face_detector.pbtxt");
    Mat f1=imread("../img/h1.png");
    Mat f2=imread("../img/h2.png");
    Mat f3=imread("../img/h3.png");
    vector<Mat> vf {f1, f2, f3};
    Mat blob = dnn::blobFromImages(vf, 1, Size(128,96), Scalar(104, 177, 123, 0), false, false);
    net.setInput(blob);
    Mat res = net.forward("detection_out");
    ```

### LINKED ISSUE #16630: dnn(Scale): errornous handling of batch inputs in untrainable scale mode
- OpenCV => 3.4

##### Detailed description

The untrainable mode in ScaleLayer uses the second input as scale factors for the first input.
```
outputs[0] = inputs[0] * inputs[1] // elementwise multiplication with broadcasting
```

The code which performs the above operation appears to assume that the batch size is one (but there is no assertion to enforce the requirement).

https://github.com/opencv/opencv/blob/3efa78311a837b7a06d81f2f359ca98e6d26609a/modules/dnn/src/layers/scale_layer.cpp#L86-L94

`weights` is reshaped (assuming that batch size is one?).

https://github.com/opencv/opencv/blob/3efa78311a837b7a06d81f2f359ca98e6d26609a/modules/dnn/src/layers/scale_layer.cpp#L98-L104

This code attempts to look for broadcasting options.  It does so by attempting to match `numWeights` (which is `weights.total()`) to a subrange in the input tensor starting from `axis`. This logic will fail to find a match and depending on input shapes will either lead to incorrect outputs or tip-off downstream assertions 

**Example:**
```
input shape: 2 3 3 3
weights input: 2 3 1 1
```

The shapes are compatible for an elementwise operation. The current logic will fail to see it.

`weights` is reshaped to `6 1`. The matching code tries to look for a subrange in the input `(2 3 3 3)` which totals to 6 but there is no such range.

This particular example triggers an assertion.

```
(-215:Assertion failed) start <= (int)shape.size() && end <= (int)shape.size() && start <= end in function 'total'
```

##### Steps to reproduce
A test for PR16436 (not merged yet) can reproduce it:

https://github.com/opencv/opencv/blob/a42268678eac90fb7d5f09d16e3a2db7167d2641/modules/dnn/test/test_darknet_importer.cpp#L547-L548

Change `testDarknetLayer("scale_channels", false, false);` to `testDarknetLayer("scale_channels", false, true);`

### LINKED ISSUE #17063: OpenCV DNN - Assertion Failed on Batch
- OpenCV => 4.3.0-dev
- Compiler => Python 3.6.8

##### Detailed description

Using OpenCV DNN with an EfficientNetB0 saved using Tensorflow I can input a batch of 1 image but it breaks while doing the forward when I input a batch higher than that. The error is:

> Traceback (most recent call last):
>   File "efficientnet_batch_error.py", line 18, in <module>
>     output = model.forward(output_layers)
> cv2.error: OpenCV(4.3.0-dev) /home/computer/workspace/opencv/modules/dnn/include/opencv2/dnn/shape_utils.hpp:171: error: (-215:Assertion failed) start <= (int)shape.size() && end <= (int)shape.size() && start <= end in function 'total'

##### Steps to reproduce

```
from cv2 import dnn
import numpy as np


model_bytes = open('b0.pb', 'rb').read()
model = dnn.readNetFromTensorflow(model_bytes)
del model_bytes
ln = model.getLayerNames()
output_layers = [ln[i[0] - 1] for i in model.getUnconnectedOutLayers()]
output_layers = output_layers[-1]

# This works
input_images = np.random.random((1, 3, 224, 224))
model.setInput(input_images)
output = model.forward(output_layers)
print(output.shape)

# This breaks
input_images = np.random.random((32, 3, 224, 224))
model.setInput(input_images)
output = model.forward(output_layers)
print(output.shape)
```

The model can be found [here](https://drive.google.com/open?id=1xnV2uydgaUffVTCwBhz51qN2qcZKunVa).

ashishkrshrivastava:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/729

resolves #16630 
resolves #17063
resolves #13626

```
opencv_extra=scale_batchsize
```

### This PR Modifies
Handling of inputs in batch for scale layer. 

Changes are done considering two cases:
1 - Weights and Biases are constants, in this case dimension could be either [1, C, H, W] or [C, H, W].
2 - Weights and Biases are non-constants ( say input_1 ), i.e. they are calculated at runtime. so their shapes will be [N, C, H, W] where N will be same as batch size of input_0.

So Basically changes are added for Point 2. i.e. untrainable parameters/ scales with batch size greater than 1. Before scale layer was only supporting Point 1. i.e. trainable weights, ignoring the fact that scales can have batchsize > 1.

```
build_image:Custom=ubuntu-openvino-2020.3.0:16.04

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*Layer_Test*:*cale*
```

### DIFF
--- a/modules/dnn/src/layers/scale_layer.cpp
+++ b/modules/dnn/src/layers/scale_layer.cpp
@@ -78,64 +78,95 @@ public:
         Mat &outBlob = outputs[0];
         // There is a mode when we multiply a first blob by a second one
         // instead of trainable weights.
-        Mat weights = hasWeights ? (blobs.empty() ? inputs[1] : blobs[0]).reshape(1, 1) : Mat();;
+        Mat weights = hasWeights ? (blobs.empty() ? inputs[1] : blobs[0]).reshape(1, 1) : Mat();
         Mat bias = hasBias ? (blobs.empty() ? inputs[1] : blobs.back()).reshape(1, 1) : Mat();
 
         MatShape inpShape = shape(inpBlob);
+        const int numSamples = inpShape[0];
+        const int numWeightsSamples = (hasWeights || hasBias) && blobs.empty() ? numSamples : 1;
         const int numWeights = !weights.empty() ? weights.total() : bias.total();
+        const int numWeightsPerSample = numWeights / numWeightsSamples;
         CV_Assert(numWeights != 0);
         if (hasWeights && hasBias)
             CV_CheckEQ(weights.total(), bias.total(), "Incompatible weights/bias blobs");
 
-        int endAxis;
+        axis = clamp(axis , inpBlob.dims) ? clamp(axis , inpBlob.dims) : 1;
+        CV_Assert(axis > 0 && axis < inpBlob.dims);
+
+        int endAxis = 1;
         for (endAxis = axis + 1; endAxis <= inpBlob.dims; ++endAxis)
         {
-            if (total(inpShape, axis, endAxis) == numWeights)
+            if (total(inpShape, axis, endAxis) == numWeightsPerSample)
                 break;
         }
-        CV_Assert(total(inpShape, axis, endAxis) == numWeights);
+        endAxis = (numWeightsPerSample == 1) ? 1 : endAxis;
+
+        CV_Assert(total(inpShape, axis, endAxis) == numWeightsPerSample);
         CV_Assert(!hasBias || numWeights == bias.total());
         CV_CheckTypeEQ(inpBlob.type(), CV_32FC1, ""); CV_CheckTypeEQ(outBlob.type(), CV_32FC1, "");
 
-        int numSlices = total(inpShape, 0, axis);
+        int numSlices = total(inpShape, 1, axis);
         float* inpData = (float*)inpBlob.data;
         float* outData = (float*)outBlob.data;
 
+        float* weightsData = !weights.empty() ? (float*)weights.data : 0;
+        float* biasesData = hasBias ? (float*)bias.data : 0;
         if (endAxis != inpBlob.dims)
         {
-            float* weightsData = !weights.empty() ? (float*)weights.data : 0;
-            float* biasesData = hasBias ? (float*)bias.data : 0;
-            int spatialSize = total(inpShape, endAxis);  // spatialSize != 1
-            for (int i = 0; i < numSlices; ++i)
+            for (int inp_sample = 0; inp_sample < numSamples; inp_sample++)
             {
-                for (int j = 0; j < numWeights; ++j)
+                int spatialSize = total(inpShape, endAxis);  // spatialSize != 1
+                for (int i = 0; i < numSlices; ++i)
                 {
-                    float w = weightsData ? weightsData[j] : 1;
-                    float b = biasesData ? biasesData[j] : 0;
-                    Mat inpSlice(1, spatialSize, CV_32F, inpData);
-                    Mat outSlice(1, spatialSize, CV_32F, outData);
-                    inpSlice.convertTo(outSlice, CV_32F, w, b);
-                    inpData += spatialSize;
-                    outData += spatialSize;
+                    for (int j = 0; j < numWeightsPerSample; ++j)
+                    {
+                        float w = weightsData ? weightsData[j] : 1;
+                        float b = biasesData ? biasesData[j] : 0;
+                        Mat inpSlice(1, spatialSize, CV_32F, inpData);
+                        Mat outSlice(1, spatialSize, CV_32F, outData);
+                        inpSlice.convertTo(outSlice, CV_32F, w, b);
+                        inpData += spatialSize;
+                        outData += spatialSize;
+                    }
+                }
+                if (numWeightsSamples > 1)
+                {
+                    weightsData += weightsData ? numWeightsPerSample : 0;
+                    biasesData += biasesData ? numWeightsPerSample : 0;
                 }
             }
         }
         else
         {
-            for (int i = 0; i < numSlices; ++i)
+            for (int inp_sample = 0; inp_sample < numSamples; inp_sample++)
             {
-                Mat inpSlice(1, numWeights, CV_32F, inpData);
-                Mat outSlice(1, numWeights, CV_32F, outData);
-                if (!weights.empty())
+                for (int i = 0; i < numSlices; ++i)
                 {
-                    multiply(inpSlice, weights, outSlice);
-                    if (hasBias)
-                        add(outSlice, bias, outSlice);
+                    Mat inpSlice(1, numWeightsPerSample, CV_32F, inpData);
+                    Mat outSlice(1, numWeightsPerSample, CV_32F, outData);
+                    if (!weights.empty())
+                    {
+                        Mat weightsSlice(1, numWeightsPerSample, CV_32F, weightsData);
+                        multiply(inpSlice, weightsSlice, outSlice);
+                        if (hasBias)
+                        {
+                            Mat biasesSlice(1, numWeightsPerSample, CV_32F, biasesData);
+                            add(outSlice, biasesSlice, outSlice);
+                        }
+                    }
+                    else if (hasBias)
+                    {
+                        Mat biasesSlice(1, numWeightsPerSample, CV_32F, biasesData);
+                        add(inpSlice, biasesSlice, outSlice);
+                    }
+                    inpData += numWeightsPerSample;
+                    outData += numWeightsPerSample;
                 }
-                else if (hasBias)
-                    add(inpSlice, bias, outSlice);
-                inpData += numWeights;
-                outData += numWeights;
+                if (numWeightsSamples > 1)
+                {
+                    weightsData += numWeightsPerSample;
+                    biasesData += numWeightsPerSample;
+                }
             }
         }
     }
--- a/modules/dnn/src/layers/scale_layer.cpp
+++ b/modules/dnn/src/layers/scale_layer.cpp
@@ -101,7 +101,6 @@ public:
         }
         endAxis = (numWeightsPerSample == 1) ? 1 : endAxis;
 
-        CV_Assert(total(inpShape, axis, endAxis) == numWeightsPerSample);
         CV_Assert(!hasBias || numWeights == bias.total());
         CV_CheckTypeEQ(inpBlob.type(), CV_32FC1, ""); CV_CheckTypeEQ(outBlob.type(), CV_32FC1, "");
 
@@ -115,7 +114,7 @@ public:
         {
             for (int inp_sample = 0; inp_sample < numSamples; inp_sample++)
             {
-                int spatialSize = total(inpShape, endAxis);  // spatialSize != 1
+                int spatialSize = total(inpShape, (endAxis > axis ? endAxis : axis));  // spatialSize != 1
                 for (int i = 0; i < numSlices; ++i)
                 {
                     for (int j = 0; j < numWeightsPerSample; ++j)
@@ -164,8 +163,8 @@ public:
                 }
                 if (numWeightsSamples > 1)
                 {
-                    weightsData += numWeightsPerSample;
-                    biasesData += numWeightsPerSample;
+                    weightsData += weightsData ? numWeightsPerSample : 0;
+                    biasesData += biasesData ? numWeightsPerSample : 0;
                 }
             }
         }
--- a/modules/dnn/test/test_layers.cpp
+++ b/modules/dnn/test/test_layers.cpp
@@ -714,10 +714,76 @@ TEST_P(Scale_untrainable, Accuracy)
 INSTANTIATE_TEST_CASE_P(Layer_Test, Scale_untrainable, Combine(
 /*input size*/   Values(Vec4i(2, 3, 4, 5)),
 /*axis, #dims*/  Values(Vec2i(0, 1), Vec2i(0, 2), Vec2i(0, 3), Vec2i(0, 4),
-                                     Vec2i(1, 2), Vec2i(1, 3), Vec2i(1, 4),
-                                     Vec2i(2, 2), Vec2i(2, 3),
-                                     Vec2i(3, 2)),
+                        Vec2i(1, 1), Vec2i(1, 2), Vec2i(1, 3), Vec2i(1, 4),
+                                     Vec2i(2, 1), Vec2i(2, 2), Vec2i(2, 3),
+                                                  Vec2i(3, 1), Vec2i(3, 2)),
 /*conv fusion*/  testing::Bool()
+));
+
+typedef testing::TestWithParam<tuple<Vec4i, Vec2i, bool> > Scale_trainable;
+TEST_P(Scale_trainable, Accuracy)
+{
+    Vec4i inpShapeVec = get<0>(GetParam());
+    int axis = get<1>(GetParam())[0];
+    int weightsDims = get<1>(GetParam())[1];
+    bool isbatchDim = get<2>(GetParam());
+    int batchDim = int(isbatchDim);
+    const int inpShape[] = {inpShapeVec[0], inpShapeVec[1], inpShapeVec[2], inpShapeVec[3]};
+    int totalPerSample = inpShapeVec[1] * inpShapeVec[2] * inpShapeVec[3];
+    int Shape_weights[4];
+
+    // Create a network with one inputs. Scale layer multiplies a first input with weights blob
+    Net net;
+    LayerParams lp;
+    lp.type = "Scale";
+    lp.name = "testLayer";
+    lp.set("axis", axis);
+
+    Mat input(4, inpShape, CV_32F);
+    // batch size is 1
+    Shape_weights[0] = 1;
+    axis = axis == 0 ? 1 : axis;
+
+    for (int dim = batchDim; dim < weightsDims + batchDim; dim++)
+        Shape_weights[dim] = inpShape[axis + dim - batchDim];
+    Mat weights(weightsDims + batchDim, Shape_weights, CV_32F);
+    int totalWeightsPerSample = weights.total();
+    randu(input, -1, 1);
+    randu(weights, -1, 1);
+
+    lp.blobs.push_back(weights);
+    net.addLayerToPrev(lp.name, lp.type, lp);
+
+    net.setInput(input);
+    net.setPreferableBackend(DNN_BACKEND_OPENCV);
+    Mat out = net.forward();
+
+    Mat ref(input.dims, input.size, CV_32F);
+    float* inpData = (float*)input.data;
+    float* refData = (float*)ref.data;
+    float* weightsData = (float*)weights.data;
+    int spatialSize = 1;
+    for (int i = axis + weightsDims; i < 4; ++i)
+        spatialSize *= inpShape[i];
+    for (int num_samples = 0; num_samples < inpShape[0] ;num_samples++)
+    {
+        for (int i = 0; i < totalPerSample; ++i)
+        {
+            int index = i + num_samples * totalPerSample;
+            float w = weightsData[(i / spatialSize) % totalWeightsPerSample];
+            refData[index] = inpData[index] * w;
+        }
+    }
+    normAssert(out, ref);
+}
+
+INSTANTIATE_TEST_CASE_P(Layer_Test, Scale_trainable, Combine(
+/*input size*/   Values(Vec4i(2, 3, 4, 5)),
+/*axis, #dims*/  Values(Vec2i(0, 1), Vec2i(0, 2), Vec2i(0, 3),
+                                     Vec2i(1, 1), Vec2i(1, 2), Vec2i(1, 3),
+                                                  Vec2i(2, 1), Vec2i(2, 2),
+                                                               Vec2i(3, 1)),
+/*batchsize*/    testing::Bool()
 ));
 
 typedef testing::TestWithParam<tuple<Vec4i, Vec4i, int, int, int> > Crop;
--- a/modules/dnn/test/test_layers.cpp
+++ b/modules/dnn/test/test_layers.cpp
@@ -639,6 +639,8 @@ TEST_P(Scale_untrainable, Accuracy)
     int weightsDims = get<1>(GetParam())[1];
     bool testFusion = get<2>(GetParam());
     const int inpShape[] = {inpShapeVec[0], inpShapeVec[1], inpShapeVec[2], inpShapeVec[3]};
+    int totalPerSample = inpShapeVec[1] * inpShapeVec[2] * inpShapeVec[3];
+    int Shape_weights[4];
 
     // Create a network with two inputs. Scale layer multiplies a first input to
     // a second one. See http://caffe.berkeleyvision.org/tutorial/layers/scale.html
@@ -672,7 +674,12 @@ TEST_P(Scale_untrainable, Accuracy)
     net.connect(0, 1, id, 1);
 
     Mat input(4, inpShape, CV_32F);
-    Mat weights(weightsDims, &inpShape[axis], CV_32F);
+    Shape_weights[0] = inpShape[0];
+    axis = axis == 0 ? 1 : axis;
+    for (int dim = 1; dim < weightsDims; dim++)
+        Shape_weights[dim] = inpShape[axis+dim-1];
+    Mat weights(weightsDims, Shape_weights, CV_32F);
+    int totalWeightsPerSample = weights.total() / Shape_weights[0];
     randu(input, -1, 1);
     randu(weights, -1, 1);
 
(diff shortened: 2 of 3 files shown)


YashasSamaga:
There is a `scale_channels` test which was added recently but batch testing was disabled due to this bug.

https://github.com/opencv/opencv/blob/3a99ebb446cec957fcbbb221933e0cd2a568dd8e/modules/dnn/test/test_darknet_importer.cpp#L557-L558

You have to change the last argument to `true` to enable the test on batches.

ashishkrshrivastava:
>  You have to change the last argument to true to enable the test on batches.

@YashasSamaga thanks, Done !

modules/dnn/test/test_layers.cpp
```diff
@@ -690,22 +697,27 @@ TEST_P(Scale_untrainable, Accuracy)
...
 
 INSTANTIATE_TEST_CASE_P(Layer_Test, Scale_untrainable, Combine(
 /*input size*/   Values(Vec4i(2, 3, 4, 5)),
 /*axis, #dims*/  Values(Vec2i(0, 1), Vec2i(0, 2), Vec2i(0, 3), Vec2i(0, 4),
-                                     Vec2i(1, 1), Vec2i(1, 2), Vec2i(1, 3),
-                                                  Vec2i(2, 1), Vec2i(2, 2),
-                                                               Vec2i(3, 1)),
+                                     Vec2i(1, 2), Vec2i(1, 3), Vec2i(1, 4),
+                                     Vec2i(2, 2), Vec2i(2, 3),
+                                     Vec2i(3, 2)),
```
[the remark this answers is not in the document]
ashishkrshrivastava: Sorry, couldn't understand.

> And just add one more parameter for batch size.

by this you mean `axis, dims, batchsize` ? 
Batchsize is always 2 in these tests and original parameters are still there. `axis, dims`.
I just changed dims by adding 1 because previously batchsize wasn't considered in dims.

modules/dnn/src/layers/scale_layer.cpp
```diff
@@ -78,64 +78,94 @@ class ScaleLayerImpl CV_FINAL : public ScaleLayer
...
-        Mat weights = hasWeights ? (blobs.empty() ? inputs[1] : blobs[0]).reshape(1, 1) : Mat();;
-        Mat bias = hasBias ? (blobs.empty() ? inputs[1] : blobs.back()).reshape(1, 1) : Mat();
-
         MatShape inpShape = shape(inpBlob);
+        const int numSamples = inpShape[0];
+
+        const int numWeightsSamples = (hasWeights || hasBias) && blobs.empty() ? numSamples : 1;
+
+        Mat weights = hasWeights ? (blobs.empty() ? inputs[1] : blobs[0]).reshape(1 , 1) : Mat();
+        Mat bias = hasBias ? (blobs.empty() ? inputs[1] : blobs.back()).reshape(1 , 1) : Mat();
```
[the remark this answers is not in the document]
ashishkrshrivastava: No, we dont need really. just wanted to initialize scalar variable above Mat variables. I will change that.

modules/dnn/test/test_layers.cpp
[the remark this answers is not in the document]
dkurt: So why do we need to modify this test? Can you please clarify a bit

modules/dnn/src/layers/scale_layer.cpp
```diff
@@ -29,7 +29,7 @@ class ScaleLayerImpl CV_FINAL : public ScaleLayer
     {
         setParamsFrom(params);
         hasBias = params.get<bool>("bias_term", false);
-        axis = params.get<int>("axis", 1);
+        axis = (params.get<int>("axis", 1) > 0) ? params.get<int>("axis", 1) : 1;
```
YashasSamaga: Is this correct? Why can't axis be negative?

modules/dnn/test/test_layers.cpp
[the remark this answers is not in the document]
ashishkrshrivastava:
```cpp
/*axis, #dims*/  Values(Vec2i(0, 1), Vec2i(0, 2), Vec2i(0, 3), Vec2i(0, 4),
                                     Vec2i(1, 1), Vec2i(1, 2), Vec2i(1, 3),
                                                  Vec2i(2, 1), Vec2i(2, 2),
                                                               Vec2i(3, 1)),
```

shape of input [ N, C, H, W ] 

`Vec2i(0, 1) - axis = 0, dim = 1 [ N ]` fine
`Vec2i(0, 2) - axis = 0, dim = 2 [ N, C ]` fine
`Vec2i(0, 3) - axis = 0, dim = 2 [ N, C, H ]` fine
`Vec2i(0, 4) - axis = 0, dim = 2 [ N, C, H, W ]` fine
`Vec2i(1, 1) - axis = 1, dim = 1 [ N ]` , seems incorrect 
`Vec2i(1, 2) - axis = 1, dim = 2 [ N, C ]` - fine
`Vec2i(1, 3) - axis = 1, dim = 3 [ N, C, H ]` - fine
`Vec2i(2, 1) - axis = 2, dim = 1 [ N ]`, seems incorrect 
`Vec2i(2, 2) - axis = 2, dim = 2 [ N, H ]`- fine, can be broadcasted as  [ N, 1, H ] 
`Vec2i(3, 1) - axis = 3, dim = 1 [ N ]`- seems incorrect

So I removed these three tests [ (1, 1), (2, 1), (3, 1)]. and added three more test that are possible [ (1, 4), (2, 3), (3, 2) ]:
`Vec2i(1, 4) - axis = 1, dim = 4 [ N, C, H, W ]`- fine
`Vec2i(2, 3) - axis = 2, dim = 3 [ N, C, H ]`- fine
`Vec2i(3, 2) - axis = 3, dim = 2 [ N, W ]`- fine, can be broadcasted as [ N, 1, 1, W ]

```cpp

/*axis, #dims*/  Values(Vec2i(0, 1), Vec2i(0, 2), Vec2i(0, 3), Vec2i(0, 4),
                                     Vec2i(1, 2), Vec2i(1, 3), Vec2i(1, 4),
                                     Vec2i(2, 2), Vec2i(2, 3),
                                     Vec2i(3, 2)),
```

modules/dnn/src/layers/scale_layer.cpp
ashishkrshrivastava: Good point, Thanks.

ashishkrshrivastava: @YashasSamaga, please have a look now.

YashasSamaga: I think you're looking for this:

https://github.com/opencv/opencv/blob/ca23c0e630ba9c4fc147c4913d6238f5de3bbcd9/modules/dnn/src/layers/fully_connected_layer.cpp#L118

ashishkrshrivastava: @YashasSamaga not exactly, but yeah

```
axis = clamp(axis , inpBlob.dims) ? clamp(axis , inpBlob.dims) : 1;
CV_Assert(axis > 0 && axis < inpBlob.dims);
```
this would work. 

actually here in this loop, `numWeightsPerSample` is total weights excluding batchsize, so we need to check with axis >= 1.
```
for (endAxis = axis + 1; endAxis <= inpBlob.dims; ++endAxis)
        {
            if (total(inpShape, axis, endAxis) == numWeightsPerSample)
                break;
        }
endAxis = (numWeightsPerSample == 1) ? 1 : endAxis;
```

modules/dnn/test/test_layers.cpp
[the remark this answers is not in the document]
dkurt: Can you please add a separate test only for new functionality? If I'm not mistaken, we try so solve batching in Scale Layer. So while
```
1 x C x H x W
1 x C
```
work can we test
```
N x H x W x C
1 x C
```
and
```
N x C x H x W
N x C
```
?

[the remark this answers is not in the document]
ashishkrshrivastava: sure. they work. Let me add tests for them.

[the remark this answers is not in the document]
ashishkrshrivastava: https://github.com/opencv/opencv/blob/2842116ebd717356ecc8f222284161ae5f0cb4a9/modules/dnn/test/test_layers.cpp#L634
these tests will cover 
```
N x C x H x W
N x C
```
https://github.com/opencv/opencv/blob/2842116ebd717356ecc8f222284161ae5f0cb4a9/modules/dnn/test/test_layers.cpp#L723
and these will cover 
```
N x H x W x C
1 x C
```

ashishkrshrivastava:
@dkurt  do you suggest any changes in this PR. Actually changes in this PR are needed to proceed PR [#16750](https://github.com/opencv/opencv/pull/16750)

asmorkalov:
@dkurt Do you have any comments?

modules/dnn/test/test_layers.cpp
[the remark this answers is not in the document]
dkurt: Please keep existing test unchanged.

modules/dnn/src/layers/scale_layer.cpp
```diff
@@ -78,64 +78,95 @@ class ScaleLayerImpl CV_FINAL : public ScaleLayer
...
         // instead of trainable weights.
-        Mat weights = hasWeights ? (blobs.empty() ? inputs[1] : blobs[0]).reshape(1, 1) : Mat();;
+        Mat weights = hasWeights ? (blobs.empty() ? inputs[1] : blobs[0]).reshape(1, 1) : Mat();
         Mat bias = hasBias ? (blobs.empty() ? inputs[1] : blobs.back()).reshape(1, 1) : Mat();
 
         MatShape inpShape = shape(inpBlob);
+        const int numSamples = inpShape[0];
+        const int numWeightsSamples = (hasWeights || hasBias) && blobs.empty() ? numSamples : 1;
         const int numWeights = !weights.empty() ? weights.total() : bias.total();
+        const int numWeightsPerSample = numWeights / numWeightsSamples;
```
dkurt, line 88: Do we have real example when weights have batch dimension? I mean when we have different NCHW input and NC weights?

dkurt:
@ashishkrshrivastava, thanks for the patch! I still would like to recommend to write in the description what's new in this PR. It seems to me a bit overcomplicated for now and I wanted to clarify which configurations were supported before and starting from this PR.

modules/dnn/src/layers/scale_layer.cpp
YashasSamaga: The scaling parameters can be computed during inference (i.e. untrainable mode). This situation will lead to different weights for different batch items.

EfficientNetB0-YOLO is one example which does it. (https://github.com/opencv/opencv/pull/16436#issuecomment-585732681)

modules/dnn/src/layers/scale_layer.cpp
```diff
@@ -78,64 +78,95 @@ class ScaleLayerImpl CV_FINAL : public ScaleLayer
         Mat &outBlob = outputs[0];
         // There is a mode when we multiply a first blob by a second one
         // instead of trainable weights.
-        Mat weights = hasWeights ? (blobs.empty() ? inputs[1] : blobs[0]).reshape(1, 1) : Mat();;
+        Mat weights = hasWeights ? (blobs.empty() ? inputs[1] : blobs[0]).reshape(1, 1) : Mat();
         Mat bias = hasBias ? (blobs.empty() ? inputs[1] : blobs.back()).reshape(1, 1) : Mat();
 
         MatShape inpShape = shape(inpBlob);
+        const int numSamples = inpShape[0];
+        const int numWeightsSamples = (hasWeights || hasBias) && blobs.empty() ? numSamples : 1;
```
ashishkrshrivastava, line 86: Here, I assumed that if weights are non-constants, then batch size could be N for input NCHW. i.e. for each sample weight is different ( example- multiplication of two inputs ) and if weights are constants then batch size is equal to 1, assuming it will be same for all samples.

modules/dnn/src/layers/scale_layer.cpp
```diff
@@ -78,64 +78,95 @@ class ScaleLayerImpl CV_FINAL : public ScaleLayer
...
         const int numWeights = !weights.empty() ? weights.total() : bias.total();
+        const int numWeightsPerSample = numWeights / numWeightsSamples;
         CV_Assert(numWeights != 0);
         if (hasWeights && hasBias)
             CV_CheckEQ(weights.total(), bias.total(), "Incompatible weights/bias blobs");
 
-        int endAxis;
+        axis = clamp(axis , inpBlob.dims) ? clamp(axis , inpBlob.dims) : 1;
+        CV_Assert(axis > 0 && axis < inpBlob.dims);
```
ashishkrshrivastava, line 94: if axis is 0, then it is assigned as 1. since broadcasting cannot be done at N for input dim NCHW.

modules/dnn/src/layers/scale_layer.cpp
```diff
@@ -78,64 +78,95 @@ class ScaleLayerImpl CV_FINAL : public ScaleLayer
...
         {
-            if (total(inpShape, axis, endAxis) == numWeights)
+            if (total(inpShape, axis, endAxis) == numWeightsPerSample)
                 break;
         }
-        CV_Assert(total(inpShape, axis, endAxis) == numWeights);
+        endAxis = (numWeightsPerSample == 1) ? 1 : endAxis;
+
+        CV_Assert(total(inpShape, axis, endAxis) == numWeightsPerSample);
         CV_Assert(!hasBias || numWeights == bias.total());
```
ashishkrshrivastava, line 105: Checking broadcasting options. if size of weights is either N or 1, then endAxis=1. 
ex - NCHW ( input dims ), N/1 ( weights ). then endAxis = 1.
otherwise calculate endAxis from condition. 
`if (total(inpShape, axis, endAxis) == numWeightsPerSample)`

modules/dnn/src/layers/scale_layer.cpp
```diff
@@ -78,64 +78,95 @@ class ScaleLayerImpl CV_FINAL : public ScaleLayer
...
         CV_CheckTypeEQ(inpBlob.type(), CV_32FC1, ""); CV_CheckTypeEQ(outBlob.type(), CV_32FC1, "");
 
-        int numSlices = total(inpShape, 0, axis);
+        int numSlices = total(inpShape, 1, axis);
         float* inpData = (float*)inpBlob.data;
```
ashishkrshrivastava, line 109: axis 0 is not considered while calculating numSlices because an extra loop for number of sample in a batch is applied at Line 116.

modules/dnn/src/layers/scale_layer.cpp
ashishkrshrivastava: issue #17063 also encountered problem in EfficientNetBO ( tensorflow )

modules/dnn/test/test_layers.cpp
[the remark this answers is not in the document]
ashishkrshrivastava: @dkurt Done!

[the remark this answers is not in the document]
ashishkrshrivastava: I didn't remove any but added some more tests.

modules/dnn/test/test_layers.cpp
```diff
@@ -690,23 +697,94 @@ TEST_P(Scale_untrainable, Accuracy)
...
     }
     normAssert(out, ref);
 }
 
 INSTANTIATE_TEST_CASE_P(Layer_Test, Scale_untrainable, Combine(
 /*input size*/   Values(Vec4i(2, 3, 4, 5)),
 /*axis, #dims*/  Values(Vec2i(0, 1), Vec2i(0, 2), Vec2i(0, 3), Vec2i(0, 4),
+                        Vec2i(1, 1), Vec2i(1, 2), Vec2i(1, 3), Vec2i(1, 4),
+                                     Vec2i(2, 1), Vec2i(2, 2), Vec2i(2, 3),
+                                                  Vec2i(3, 1), Vec2i(3, 2)),
```
ashishkrshrivastava, line 719: These tests are for scales that are computed at runtime. So their batchsize should be equal to Batchsize of Input0. So for all of them Scales are of size [N,C,H,W] / [N,C,H] / [N,H,W] / [N,C] / [N,H] / [N,W] / [N]

Extra added tests are 
Vec2i(1,4)
Vec2i(2,3) 
Vec2i(3,2)

modules/dnn/test/test_layers.cpp
```diff
@@ -690,23 +697,94 @@ TEST_P(Scale_untrainable, Accuracy)
...
+}
+
+INSTANTIATE_TEST_CASE_P(Layer_Test, Scale_trainable, Combine(
+/*input size*/   Values(Vec4i(2, 3, 4, 5)),
+/*axis, #dims*/  Values(Vec2i(0, 1), Vec2i(0, 2), Vec2i(0, 3),
                                      Vec2i(1, 1), Vec2i(1, 2), Vec2i(1, 3),
                                                   Vec2i(2, 1), Vec2i(2, 2),
                                                                Vec2i(3, 1)),
-/*conv fusion*/  testing::Bool()
+/*batchsize*/    testing::Bool()
```
ashishkrshrivastava, line 787: Scale trainable have weights as constants, 
Case 1: so for dim <= 4. shape would be like [1,C,H,W] / [1,C,H] / [1,H,W] / [1,C] / [1,H] / [1,W] / [1]  and 
Case 2: for dim<=3, shape would be [C,H,W] / [C,H] / [H,W] / [C] / [H] / [W].
One more argument `batchsize` is added.
batchsize = True for Case 1
batchsize = False for Case 2

-------------------------------------------Batchsize=True---------------------------------------Batchsize=False------------------------------
Vec2i(0,1)-----------------------------------[1,C]-----------------------------------------------------[C]-------------------------------------------
Vec2i(0,2)-----------------------------------[1,C,H]--------------------------------------------------[C,H]----------------------------------------
Vec2i(0,3)-----------------------------------[1,C,H,W]----------------------------------------------[C,H,W]------------------------------------
**XX** Vec2i(0,4)------------------------------[1,C,H,W,_]----------------------------------------------[C,H,W,_]----------------------------
Vec2i(1,1)-----------------------------------[1,C]-----------------------------------------------------[C]-------------------------------------------
Vec2i(1,2)-----------------------------------[1,C,H]--------------------------------------------------[C,H]----------------------------------------
Vec2i(1,3)-----------------------------------[1,C,H,W]----------------------------------------------[C,H,W]-------------------------------------
Vec2i(2,1)-----------------------------------[1,H]-----------------------------------------------------[H]-------------------------------------------
Vec2i(2,2)-----------------------------------[1,C,H]--------------------------------------------------[C,H]-----------------------------------------
Vec2i(3,1)-----------------------------------[1,H]-----------------------------------------------------[H]--------------------------------------------

So all of the cases are there but Vec2i(0,4)

asmorkalov:
@ashishkrshrivastava @dkurt Do you have conclusion on the patch? Is it ready?

dkurt:
Yeah, sorry for delay. Just wanted to make some local experiments.

dkurt:
@ashishkrshrivastava, thanks for the complete solution for these three issues! It's good to see that it works but I wanted to propose smaller some steps to solve them one-by-one and probably it will help us to simplify this PR and additionally fix other bugs/weak points.

1. issue https://github.com/opencv/opencv/issues/13626 can be solved by a single line patch:
```patch
@@ -303,7 +303,7 @@ Ptr<Layer> ShiftLayer::create(const LayerParams& params)
     scaleParams.type = "Scale";
     scaleParams.blobs = params.blobs;
     scaleParams.set("bias_term", true);
-    scaleParams.set("axis", 0);
+    scaleParams.set("axis", 1);
```

<details>
<summary>validation code</summary>

```python
import numpy as np
import cv2 as cv

inp0 = cv.dnn.blobFromImage(cv.imread('/home/dkurt/Pictures/image.jpg'), size=(300, 300), mean=(104, 177, 123))
inp1 = cv.dnn.blobFromImage(cv.imread('/home/dkurt/Pictures/image2.jpg'), size=(300, 300), mean=(104, 177, 123))

batch = np.concatenate([inp0, inp1])

net = cv.dnn.readNet('opencv_face_detector_uint8.pb', 'opencv_face_detector.pbtxt')
net.setInput(batch)
out = net.forward()
print(out.shape)

net = cv.dnn.readNet('opencv_face_detector_uint8.pb', 'opencv_face_detector.pbtxt')
net.setInput(inp0)
out0 = net.forward()
net.setInput(inp1)
out1 = net.forward()

out = out.reshape(-1, 7)[:,1:]  # remove batch id column
out0 = out0.reshape(-1, 7)[:,1:]
out1 = out1.reshape(-1, 7)[:,1:]

num0 = 0
for det in out0:
    if det[0] != 0:  # exclude zeroed predictions at the bottom
        num0 += 1
    else:
        break
print(np.max(np.abs(out[:num0] - out0[:num0])))

num1 = 0
for det in out1:
    if det[0] != 0:  # exclude zeroed predictions at the bottom
        num1 += 1
    else:
        break
print(np.max(np.abs(out[num0:(num0+num1)] - out1[:num1])))
```

</details>

The only problem is failed test `runTensorFlowNet("shift_reshape_no_reorder");`:
```python
inp = tf.placeholder(tf.float32, [2, 3, 4], 'input')
shift = tf.Variable(tf.random_normal([2, 3, 4]), name='shift')
shifted = tf.add(inp, shift, name='shifted')
reshape = tf.reshape(shifted, [4, 3, 2], 'reshaped')
save(inp, reshape, 'shift_reshape_no_reorder')
```

asmorkalov:
@dkurt @ashishkrshrivastava Have you driven to conclusion with the patch?

ashishkrshrivastava:
> 1. issue #13626 can be solved by a single line patch:
> 
> ```diff
> @@ -303,7 +303,7 @@ Ptr<Layer> ShiftLayer::create(const LayerParams& params)
>      scaleParams.type = "Scale";
>      scaleParams.blobs = params.blobs;
>      scaleParams.set("bias_term", true);
> -    scaleParams.set("axis", 0);
> +    scaleParams.set("axis", 1);
> ```
> 
> validation code
> The only problem is failed test `runTensorFlowNet("shift_reshape_no_reorder");`:

As you change axis default to 1.
Broadcasting checking will change 
```cpp
1. for (endAxis = axis + 1; endAxis <= inpBlob.dims; ++endAxis)
2.        {
3.            if (total(inpShape, axis, endAxis) == numWeights)
4.                break;
5.        }
```
at **line 3**
I guess, that is why it is failing.

> but I wanted to propose some smaller steps to solve them one-by-one and probably it will help us to simplify this PR and additionally fix other bugs/weak points.

@dkurt you are right that the changes are complex. I have given the explanations above as reviews. 
I think all the issues has same problem. I have checked how to solve them one-by-one but I think the solution cant be breakdown into parts.

asmorkalov:
@dkurt Friendly reminder.

asmorkalov:
@dkurt Reminder.

asmorkalov:
@dkurt Reminder.

asmorkalov:
@dkurt, @ashishkrshrivastava  Do you have conclusion on the patch?

liuxingbaoyu:
Hello, are there any updates?
@dkurt @alalek Sorry to interrupt.

asmorkalov:
@dkurt @ashishkrshrivastava Friendly reminder.

asmorkalov:
@dkurt Friendly reminder.

asmorkalov:
@dkurt Friendly reminder.

ashishkrshrivastava:
@dkurt @asmorkalov are there any improvements needed or should I close this PR so that if anyone wants to work on this issue can contribute?

alalek (commented):
@ashishkrshrivastava Thank you for the contribution!

Could you please add some link on used specification (reference code) for this layer?
Part with axis 0=>1 should is not clear to me yet.

asenyaev:

ashishkrshrivastava:
@alalek , without changing tests these modifications won't work. I found that both tests and implementation were incorrect so I modified both. Maybe the way I was approaching this issue was wrong,  but these modifications were solving all issues related to this. I am closing this PR so that anyone else can work on it.

PR #16803, opened 2020-03-12
G-API: Implement NV12toGray

TolyaTalamanov:

```

test_modules:Custom=gapi
test_modules:Custom Win=gapi
test_modules:Custom Mac=gapi

buildworker:Custom=linux-1
test_opencl:Custom=ON
test_bigdata:Custom=1
test_filter:Custom=*

```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/imgproc.hpp
+++ b/modules/gapi/include/opencv2/gapi/imgproc.hpp
@@ -222,6 +222,21 @@ namespace imgproc {
         }
     };
 
+    G_TYPED_KERNEL(GNV12toGray, <GMat(GMat,GMat)>, "org.opencv.colorconvert.imgproc.nv12togray") {
+        static GMatDesc outMeta(GMatDesc inY, GMatDesc inUV) {
+            GAPI_Assert(inY.depth   == CV_8U);
+            GAPI_Assert(inUV.depth  == CV_8U);
+            GAPI_Assert(inY.chan    == 1);
+            GAPI_Assert(inY.planar  == false);
+            GAPI_Assert(inUV.chan   == 2);
+            GAPI_Assert(inUV.planar == false);
+
+            GAPI_Assert(inY.size.width  == 2 * inUV.size.width);
+            GAPI_Assert(inY.size.height == 2 * inUV.size.height);
+            return inY.withType(CV_8U, 1);
+        }
+    };
+
     G_TYPED_KERNEL(GNV12toBGRp, <GMatP(GMat,GMat)>, "org.opencv.colorconvert.imgproc.nv12tobgrp") {
         static GMatDesc outMeta(GMatDesc inY, GMatDesc inUV) {
             GAPI_Assert(inY.depth == CV_8U);
@@ -818,6 +833,21 @@ Output image must be 8-bit unsigned 3-channel image @ref CV_8UC3.
 @sa YUV2RGB, NV12toBGR
 */
 GAPI_EXPORTS GMat NV12toRGB(const GMat& src_y, const GMat& src_uv);
+
+/** @brief Converts an image from NV12 (YUV420p) color space to gray-scaled.
+The function converts an input image from NV12 color space to gray-scaled.
+The conventional ranges for Y, U, and V channel values are 0 to 255.
+
+Output image must be 8-bit unsigned 1-channel image @ref CV_8UC1.
+
+@note Function textual ID is "org.opencv.imgproc.colorconvert.nv12togray"
+
+@param src_y input image: 8-bit unsigned 1-channel image @ref CV_8UC1.
+@param src_uv input image: 8-bit unsigned 2-channel image @ref CV_8UC2.
+
+@sa YUV2RGB, NV12toBGR
+*/
+GAPI_EXPORTS GMat NV12toGray(const GMat& src_y, const GMat& src_uv);
 
 /** @brief Converts an image from NV12 (YUV420p) color space to BGR.
 The function converts an input image from NV12 color space to RGB.
--- a/modules/gapi/include/opencv2/gapi/imgproc.hpp
+++ b/modules/gapi/include/opencv2/gapi/imgproc.hpp
@@ -222,17 +222,6 @@ namespace imgproc {
         }
     };
 
-    G_TYPED_KERNEL(GYUV2Gray, <GMat(GMat)>, "org.opencv.colorconvert.imgproc.yuvtogray") {
-        static GMatDesc outMeta(GMatDesc in) {
-            GAPI_Assert(in.depth  == CV_8U);
-            GAPI_Assert(in.planar == false);
-            GAPI_Assert(in.size.width  % 2 == 0);
-            GAPI_Assert(in.size.height % 3 == 0);
-
-            return {CV_8U, 1, Size{in.size.width, in.size.height - (in.size.height / 3)}, false};
-        }
-    };
-
     G_TYPED_KERNEL(GNV12toGray, <GMat(GMat,GMat)>, "org.opencv.colorconvert.imgproc.nv12togray") {
         static GMatDesc outMeta(GMatDesc inY, GMatDesc inUV) {
             GAPI_Assert(inY.depth   == CV_8U);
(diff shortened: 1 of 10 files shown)


modules/gapi/src/backends/cpu/gcpuimgproc.cpp
```diff
@@ -335,6 +335,17 @@ GAPI_OCV_KERNEL(GCPUNV12toRGBp, cv::gapi::imgproc::GNV12toRGBp)
...
 };
 
+GAPI_OCV_KERNEL(GCPUNV12toGray, cv::gapi::imgproc::GNV12toGray)
+{
+    static void run(const cv::Mat& in_y, const cv::Mat& in_uv, cv::Mat& out)
+    {
+        cv::Mat yuv_planar;
+        cv::Mat uv_planar(in_y.rows / 2, in_y.cols, CV_8UC1, in_uv.data);
+        cv::vconcat(in_y, uv_planar, yuv_planar);
+        cv::cvtColor(yuv_planar, out, cv::COLOR_YUV2GRAY_NV12);
```
TolyaTalamanov: OpenCV has got many conversion codes from various YUV formats to gray but it's just alias to `COLOR_YUV2GRAY_420`:

```
  cv::COLOR_YUV2GRAY_NV21 = COLOR_YUV2GRAY_420,
  cv::COLOR_YUV2GRAY_NV12 = COLOR_YUV2GRAY_420,
  cv::COLOR_YUV2GRAY_YV12 = COLOR_YUV2GRAY_420,
  cv::COLOR_YUV2GRAY_IYUV = COLOR_YUV2GRAY_420,
  cv::COLOR_YUV2GRAY_I420 = COLOR_YUV2GRAY_420,
  cv::COLOR_YUV420sp2GRAY = COLOR_YUV2GRAY_420,
  cv::COLOR_YUV420p2GRAY  = COLOR_YUV2GRAY_420
```

`cvtColor` function requires a planar YUV format for this conversion

To create a planar Mat we simply concatenate 2 planes vertically using the `cv::vconcat` function

But we can 't just concatenate 2 planes, because the UV mat is stored in `CV_8U2` format and has a width half that of Y mat. So just create a wrapper with correct format and sizes for `cv::vconcat` function


Anyway, we copy data when concatenating planes, I think it makes sense to have an analog of this function that accepts planar YUV as input. 

@dmatveev What do you think ? Is it requires in this PR ?

TolyaTalamanov:
@dmatveev

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
dmatveev: It is not required in this PR I believe. Thanks,D

dmatveev: However, in order to better manage this temporary memory, I believe it can become a compound kernel.

TolyaTalamanov: Done

TolyaTalamanov:
@dmatveev

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
```diff
@@ -335,6 +345,29 @@ GAPI_OCV_KERNEL(GCPUNV12toRGBp, cv::gapi::imgproc::GNV12toRGBp)
...
+    static cv::GMatDesc outMeta(cv::GMatDesc y, cv::GMatDesc uv) {
+        return {CV_8U, 1, cv::gapi::own::Size{y.size.width, y.size.height + uv.size.height}, false};
+    }
+};
+
+GAPI_OCV_KERNEL(GCPUConcatYUVPlanes, GConcatYUVPlanes)
+{
+    static void run(const cv::Mat& in_y, const cv::Mat& in_uv, cv::Mat& out)
+    {
+        cv::Mat uv_planar(in_uv.rows, in_uv.cols * 2, CV_8UC1, in_uv.data);
```
TolyaTalamanov: So we can 't do this via a `cv::gapi::vertConcat` call, because we need to do a uv conversion before that, which can't be a separate kernel

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
```diff
@@ -335,6 +345,29 @@ GAPI_OCV_KERNEL(GCPUNV12toRGBp, cv::gapi::imgproc::GNV12toRGBp)
     }
 };
 
+G_TYPED_KERNEL(GConcatYUVPlanes, <cv::GMat(cv::GMat, cv::GMat)>, "concatyuvplanes") {
```
TolyaTalamanov: I don't think the user needs this operation

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -222,6 +222,32 @@ namespace imgproc {
...
+    G_TYPED_KERNEL(GYUV2Gray, <GMat(GMat)>, "org.opencv.colorconvert.imgproc.yuvtogray") {
+        static GMatDesc outMeta(GMatDesc in) {
+            GAPI_Assert(in.depth  == CV_8U);
+            GAPI_Assert(in.planar == false);
+            GAPI_Assert(in.size.width  % 2 == 0);
+            GAPI_Assert(in.size.height % 3 == 0);
+
+            return {CV_8U, 1, Size{in.size.width, in.size.height - (in.size.height / 3)}, false};
+        }
+    };
```
dmatveev: I am not sure if we need to have this very specific kernel defined in a public header -- since it serves as an implementation detail only

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -222,6 +222,32 @@ namespace imgproc {
...
+            GAPI_Assert(inY.chan    == 1);
+            GAPI_Assert(inY.planar  == false);
+            GAPI_Assert(inUV.chan   == 2);
+            GAPI_Assert(inUV.planar == false);
+
+            GAPI_Assert(inY.size.width  == 2 * inUV.size.width);
+            GAPI_Assert(inY.size.height == 2 * inUV.size.height);
+            return inY.withType(CV_8U, 1);
+        }
+    };
```
dmatveev: And this one is ok to have here - it is public

modules/gapi/include/opencv2/gapi/imgproc.hpp
dmatveev: I don't get why width must be odd and height must be even. Why?

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
dmatveev: User doesn't need  it

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
```diff
@@ -335,6 +345,29 @@ GAPI_OCV_KERNEL(GCPUNV12toRGBp, cv::gapi::imgproc::GNV12toRGBp)
...
+    }
+};
+
+GAPI_COMPOUND_KERNEL(GCPUNV12toGray, cv::gapi::imgproc::GNV12toGray)
+{
+    static cv::GMat expand(cv::GMat y, cv::GMat uv)
+    {
+        return cv::gapi::YUV2Gray(GConcatYUVPlanes::on(y, uv));
+    }
+};
```
dmatveev: You don't even need this in the public header.

modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
```diff
@@ -456,6 +456,29 @@ TEST_P(YUV2RGBTest, AccuracyTest)
     }
 }
 
+TEST_P(YUV2GrayTest, AccuracyTest)
+{
+    // G-API code //////////////////////////////////////////////////////////////
+    cv::GMat in;
+    auto out = cv::gapi::YUV2Gray(in);
+
+    cv::Mat in_mat(cv::Size(sz.width, static_cast<int>(sz.height * 1.5)), CV_8UC1);
```
dmatveev: I believe this function is misaligned with https://docs.opencv.org/4.2.0/dc/d38/group__gapi__colorconvert.html#gad322eb34ae0d3ca84e4666fd5351bec9 which takes an interleaved YUV444 image.

If you've had I420 in mind, then the kernel needs to be renamed appropriately and have 3 inputs for every plane.

modules/gapi/include/opencv2/gapi/imgproc.hpp
dmatveev: So now I've got that it is not a building block of a compound kernel but in fact an I420 conversion. Please rename and comment it appropriately as I stated in the test comments

dmatveev: oh no, it is still the building block for NV12Gray. So then you don't need it in public here.

dmatveev:
Please also open an internal ticket to support NV12toGray in Fluid.

modules/gapi/include/opencv2/gapi/imgproc.hpp
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/imgproc.hpp
TolyaTalamanov: Done

modules/gapi/src/backends/cpu/gcpuimgproc.cpp
TolyaTalamanov: Done

modules/gapi/test/common/gapi_imgproc_tests_inl.hpp
TolyaTalamanov: Done

TolyaTalamanov:
@dmatveev

Issue #16811, opened 2020-03-14
DNN custom layer memory leak


Pajsch:
##### System information (version)

- OpenCV => 4.2.0
- Operating System / Platform => Ubuntu 18.04.4
- Compiler => usr/bin/c++  (ver 7.4.0), /usr/bin/cc
- Python 3.6.9

##### Detailed description
I am running the example in:
opencv/samples/dnn/edge_detection.py

When running the example with a sufficiently long video, the memory usage of python is steadily increasing as long as it is running. An increase in memory usage happens at each execution of net.forward() .
 
This behavior seems to be due to the custom definition of CropLayer. The memory usage does not increase, when using the standard CropLayer by commenting out the line
cv.dnn_registerLayer('Crop', CropLayer)

But this leads to a wrong cropping and removes part of each frame for this model. 

Thank you for your work!

##### Steps to reproduce

Run: opencv/samples/dnn/edge_detection.py for a sufficiently long time and observe memory usage (I did it in the System Monitor).

python3 edge_detection.py --input ./videos/video.avi --prototxt ./hed_model/deploy.prototxt --caffemodel ./hed_model/hed_pretrained_bsds.caffemodel

##### Issue submission checklist

 - [x ] I report the issue, it's not a question
   
 - [x ] I checked the problem with documentation, FAQ, open issues,
       answers.opencv.org, Stack Overflow, etc and have not found solution
   
 - [ x] I updated to latest OpenCV version and the issue is still there
   
 - [ ] There is reproducer code and related data files: videos, images, onnx, etc

dkurt:
@Pajsch, thanks for this catch! Please test the changes from https://github.com/opencv/opencv/pull/16820.

Pajsch:
@dkurt With your changes the issue is fixed for me. Amazing work, thank you.

dkurt:
@Pajsch, thank you!

closing an issue

PR #16820, opened 2020-03-16
Fix memory leak in Python custom dnn layers

dkurt:
resolves https://github.com/opencv/opencv/issues/16811


ndvbd:
Was this problem fixed in opencv 4? If so in what version?

dkurt:
@ndvbd, with OpenCV 4.3.0: https://github.com/opencv/opencv/releases/tag/4.3.0 Released after several weeks after fix has been merged.

ndvbd:
Thanks @dkurt ! Do you know if the memory leak was only in the python wrapper, or also when using the DNN forward from C++?

dkurt:
@ndvbd, the issue connected to only Python bindings and corresponding solution solves it in the code related to Python API. So, it was only Python problem if you use custom layers. If there is no custom layers - there were no leaks.

PR #16822, opened 2020-03-16
more robust error handling

### LINKED ISSUE #16823: FileStorage does not handle properly an empty yml file
Under OpenCV 4.2.*, using cv::FileStorage to open an (invalid) empty yml file (0 bytes) results in resource leaking by not closing files properly.
With current 4.3.0 dev, there is even a crash by nullptr dereference when trying to check the BOM.

---

relates #16488

chacha21:
Fixed #16823 (OOB crash and resource leaking for invalid empty yml file)

### DIFF
--- a/modules/core/src/persistence.cpp
+++ b/modules/core/src/persistence.cpp
@@ -704,10 +704,15 @@ public:
             const char* json_signature = "{";
             const char* xml_signature  = "<?xml";
             char* buf = this->gets(16);
-            char* bufPtr = !buf ? "" : cv_skip_BOM(buf);
-            size_t bufOffset = bufPtr - buf;
+            char* bufPtr = !buf ? 0 : cv_skip_BOM(buf);
+            size_t bufOffset = (bufPtr - buf);
 
-            if(strncmp( bufPtr, yaml_signature, strlen(yaml_signature) ) == 0)
+            if(!bufPtr)
+            {
+                closeFile();
+                CV_Error(CV_BADARG_ERR, "Input file is invalid");
+            }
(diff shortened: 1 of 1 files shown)


modules/core/src/persistence.cpp
```diff
@@ -704,19 +704,30 @@ class FileStorage::Impl : public FileStorage_API
...
-            char* bufPtr = cv_skip_BOM(buf);
-            size_t bufOffset = bufPtr - buf;
+            char* bufPtr = !buf ? 0 : cv_skip_BOM(buf);
+            size_t bufOffset = (bufPtr - buf);
 
-            if(strncmp( bufPtr, yaml_signature, strlen(yaml_signature) ) == 0)
+            if(!bufPtr)
+            {
+                closeFile();
+                CV_Error(CV_BADARG_ERR, "Input file is invalid");
```
alalek: > closeFile();

Lets try to handle that in caller code instead (in single place).

chacha21: I am afraid of regressions by trying to refactor error handling too much.
-the current FileStorage::~FileStorage destructor does not call closeFile(), and I don't know why
-the current FileStorage::open() has some try{...}catch(...){is_opened = true;release();throw;} (line 774) that rethrow an exception, but still marks the files as open

perhaps a robust solution to avoid resource leaking could be to use local FILE* variables in the FileStorage::open(), with some RAII pattern (to automatically call some release handlers in case of exceptions), and moving the local FILE* to this->file/this->gzFile only on success.

chacha21: the unique_ptr<> won't be a practical solution, because open() needs to call puts() and rewind() that require file and gzfile members to be initialized

PR #16824, opened 2020-03-16
Fixs several problems found by static analysis

mshabunin:
Note: I'll backport to 3.4. manually


modules/calib3d/src/chessboard.cpp
```diff
@@ -3369,7 +3371,7 @@ cv::Scalar Chessboard::Board::calcEdgeSharpness(cv::InputArray _img,float rise_d
                 value += patch.at<uint8_t>(0,0);
                 ++count2;
             }
-            values.push_back(uint8_t(value/count2));
```
alalek: @D-Alex Please take a look on changes in this file.

PR #16828, opened 2020-03-16
dnn: expose only float variant of NMSBoxes for bindings 

paroj:
the float variant seems to be always shadowed by the int version as
Rect2d is implicitly convertible to Rect

### DIFF
--- a/modules/dnn/misc/python/test/test_dnn.py
+++ b/modules/dnn/misc/python/test/test_dnn.py
@@ -230,6 +230,12 @@ class dnn_test(NewOpenCVTests):
                 self.assertTrue(ret)
                 normAssert(self, refs[i], result, 'Index: %d' % i, 1e-10)
 
+    def test_nms(self):
+        confs = (1, 1)
+        rects = ((0, 0, 0.4, 0.4), (0, 0, 0.2, 0.4)) # 0.5 overlap
+
+        self.assertTrue(all(cv.dnn.NMSBoxes(rects, confs, 0, 0.6).ravel() == (0, 1)))
+
     def test_custom_layer(self):
         class CropLayer(object):
             def __init__(self, params, blobs):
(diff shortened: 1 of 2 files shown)


alalek:
Thank you for contribution!
Could you please add simple test for this?

paroj:
not sure what to test. `CV_EXPORTS_AS`? A failure test that `cv2.NMSBoxes` silently swallows `Rect2d`?

VadimLevin:
> not sure what to test. `CV_EXPORTS_AS`? A failure test that `cv2.NMSBoxes` silently swallows `Rect2d`?

Hello, @paroj, please add a test that function is exported and correct overload is called from python.

Also, there is a more general solution, that does not require to add the alias for function. If you look how `Rect` argument is parsed in  the `modules/python/src2/cv2.cpp`, you can see that `PyArg_ParseTuple` is used. You can change it to use strict type match with `pyopencv_to` calls to parse `float`/`double`/`int` parts of the argument as it was done for arrays (ref #15915)

alalek:
> You can change it to use strict type match with

Strict check may break user code.

Flexible check requires reworking of overloads handling in python bindings.

alalek:
Thank you! Test looks good to me.

Could you please rebase patch onto 3.4 branch?

paroj:
I changed the fix to just always hide the int variant. This should be fine, as the vector of boxes was always copied from tuple and the computation was done in double.

modules/dnn/misc/python/test/test_dnn.py
```diff
@@ -230,6 +230,12 @@ def test_async(self):
                 self.assertTrue(ret)
                 normAssert(self, refs[i], result, 'Index: %d' % i, 1e-10)
 
+    def test_nms(self):
+        confs = (1, 1)
+        rects = ((0, 0, 0.4, 0.4), (0, 0, 0.2, 0.4)) # 0.5 overlap
```
alalek: Does case with integers (multiple by 10) work as expected after change?

paroj: yes

PR #16832, opened 2020-03-17
3.4 tengine android

liqi-c:

```
force_builders=ARMv7,ARMv8
```


cmake/OpenCVFindTengine.cmake
```diff
@@ -36,10 +36,29 @@ IF(OPENCV_LIBTENGINE_ROOT_DIR)
...
+	IF(ANDROID)
+		SET(NDK_REVISION_REGEX "^Pkg\\.Desc = Android NDK\nPkg\\.Revision = ([0-9]+)\\.([0-9]+)\\.([0-9]+)?")
+		IF(CMAKE_TOOLCHAIN_FILE)
+			string(REGEX REPLACE "(.+)/build/cmake/android.toolchain.cmake" "\\1"  PARSED_ANDROID_NDK ${CMAKE_TOOLCHAIN_FILE})
+			IF( NOT EXISTS "${PARSED_ANDROID_NDK}/source.properties" )
+				SET(PARSED_ANDROID_NDK "${ANDROID_NDK}")
+			ENDIF()
+		ENDIF()
+		file(READ "${PARSED_ANDROID_NDK}/source.properties" PARSED_NDK_SOURCE_PROPERTIES)
+		IF(PARSED_NDK_SOURCE_PROPERTIES MATCHES "${NDK_REVISION_REGEX}")
```
alalek: Why is `ANDROID_NDK_REVISION` not work for you?

liqi-c: ANDROID_NDK_REVISION  not work for ndkr10 and ndkr14...it's empty .

alalek: check for missing values is still a check.

I believe we should support modern NDKs (for example, r16 is 2 years old, r18 drops gcc).

---

We don't want to duplicate here complex logic from Android NDK.

If user wants support for non-modern NDK versions then there is no problem to add definition of this variable through command line manually.

Also you can add `_SKIP_` / `_FORCE_` helper variables ([example](https://github.com/opencv/opencv/blob/c48473dfe42f447a6dd8a5c7b524e5aa80aa46c7/cmake/OpenCVFindLibsPerf.cmake#L29-L34)).

liqi-c: Hi ,alalek , I modify  as you suggested , use ANDROID_NDK_REVISION  to judgement . 
 but  _SKIP_ / _FORCE_  i don't know how to use . If necessary hope can  give us more detailed suggestions ,thanks

CMakeLists.txt
```diff
@@ -432,8 +432,8 @@ OCV_OPTION(WITH_IMGCODEC_PXM "Include PNM (PBM,PGM,PPM) and PAM formats support"
 OCV_OPTION(WITH_QUIRC "Include library QR-code decoding" ON
   VISIBLE_IF TRUE
   VERIFY HAVE_QUIRC)
-OCV_OPTION(WITH_TENGINE "Include Arm Inference Tengine support" OFF
-  VISIBLE_IF (ARM OR AARCH64) AND UNIX AND NOT ANDROID AND NOT IOS
+OCV_OPTION(WITH_TENGINE "Include Arm Inference Tengine support" ON
```
alalek: @liqi-c @vpisarev 
Is it a testing mode for PR?

liqi-c: I'm sorry , i close the test .

Issue #16568, opened 2020-02-13
Faster-RCNN Caffe Inference with OpenCV


enjoykcc456:
Hi, 

I have run into an error while trying to make inference with Faster-RCNN caffemodel using OpenCV (4.2.0) following this https://github.com/opencv/opencv/issues/10476.

The following is the code i use:
```
import cv2
import numpy as np

MODEL_PATH = '3c773411-a963-41b1-ae24-707545e9f601.caffemodel'
PROTOTXT_PATH = 'test.prototxt'
IMAGE_PATH = '3c773411-a963-41b1-ae24-707545e9f601.jpg'

kInpWidth = 800
kInpHeight = 600
net = cv2.dnn.readNetFromCaffe(PROTOTXT_PATH, MODEL_PATH)

img = cv2.imread(IMAGE_PATH)
img = cv2.resize(img, (kInpWidth, kInpHeight))
blob = cv2.dnn.blobFromImage(img, 1.0, None, (102.9801, 115.9465, 122.7717), False, False)
print("Input:", blob.shape, blob.dtype)

imInfo = np.array([600, 800, 1.6]).reshape([1,3]).astype('float32')
print("imInfo:", imInfo.shape, imInfo.dtype)

net.setInput(blob,"data")
net.setInput(imInfo, "im_info")

detections = net.forward()
```
The prototxt of the model:
```
name: "VGG_ILSVRC_16_layers"
input: "data"
input: "im_info"
input_shape {
  dim: 1
  dim: 3
  dim: 224
  dim: 224
}
input_shape {
  dim: 1
  dim: 3
}
layer {
  name: "conv1_1"
  type: "Convolution"
  bottom: "data"
  top: "conv1_1"
  param {
    lr_mult: 0.0
    decay_mult: 0.0
  }
  param {
    lr_mult: 0.0
    decay_mult: 0.0
  }
  convolution_param {
    num_output: 64
    pad: 1
    kernel_size: 3
  }
}
And more...
```
The error that i ran into:
`detections = net.forward()
cv2.error: OpenCV(4.2.0) /io/opencv/modules/dnn/src/dnn.cpp:562: error: (-2:Unspecified error) Can't create layer "proposal" of type "Python" in function 'getLayerInstance'`

May i know what is the problem and how can i make it works?
Thanks in advance

dkurt:
[...]

PR #16835, opened 2020-03-17
[WIP] cuda4dnn: release and relock page-locked memory when tensor is resized

YashasSamaga:
The host memory is page-locked to improve host-device memory transfer performance. #16648 
fixed a bug where the device memory is resized if the previously allocated memory isn't sufficient. However, it did not re-register the corresponding host memory.

Original report: https://github.com/opencv/opencv/issues/16568#issuecomment-590202635

```
force_builders_only=docs,Custom
buildworker:Custom=linux-4
build_image:Custom=ubuntu-cuda:16.04
```


YashasSamaga:
I think the host memory should be page-locked if and only if it's involved in a memory transfer. Currently, all the host memory associated with some device tensor is page-locked irrespective of whether it's involved in a memory transfer or not.

Perhaps can use lazy initialization in the first forward pass to lock the memory if it's involve in a memory transfer.

PR #16836, opened 2020-03-17
stitching: don't use nonfree by default

alalek:
nonfree code is disabled by default.


mshabunin:
BTW, there are some mentions of SIFT in C++ and Java tests and an optional dependency on _xfeatures2d_ in `stitching/CMakeLists.txt`. We should test it in builds without opencv_contrib.

alalek:
> builds without opencv_contrib

"Linux Debug" should handle that.

PR #16845, opened 2020-03-18
Update types_c.h

Binjian:
it seems that "CV__ENABLE_C_API_CTORS" was no where defined. Occasionally it led to failed conversion  of some old c style structures to the c++ ones in my projects.


alalek:
It is not enabled intentionally. If you need that then you should enable it externally through `CXXFLAGS`.

BTW, C API is not supported anymore (since OpenCV 2.4.0).

Binjian:
Yes. Seems to work by adding the macro CV__ENABLE_C_API_CTORS before the header where it's required. 

Thanks for the feedback!

PR #16849, opened 2020-03-19
G-API utils - fix compilation error in variant::operator=

anton-potapov:
### Description
This patch fixes compilation during (perfectly valid) assignment of [lvalue reference](https://en.cppreference.com/w/cpp/language/reference) to a `variant`. 
The error is something like this : 
> error: static assertion failed: Type not found
             static_assert(std::is_same<Target, First>::value, "Type not found");

Root cause of the problem is that during argument type deduction for `emplate<class T> varaiant::operator=(T&&)`  and lvalue ref argument  `T` is matched as a ref (e.g. `int&`). Then it is look up in the total types list of the variant (which fails) : 
>required from ‘constexpr const size_t cv::util::type_list_index<__int&__, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::value’

```
```


dmatveev:
Please add STANDALONE magic spell here in the MR text

PR #16858, opened 2020-03-20
core: fix .begin()/.end() of empty Mat

### LINKED ISSUE #16855: Calling .begin() or .end() on an empty Matrix leads to a division by zero exception.
- OpenCV => 3.1, 4.2

##### Detailed description
Calling .begin() or .end() on an empty matrix leads to a crash. This is supprising. Especially, when you are using colRange (or rowRange) to calculate a subrange you want to process. If for some reason the range is empty (this happend for us in some edge cases), then an empty Mat is returned, which again leads to a crash when calling bagin (or end) on it.

##### Steps to reproduce
```.cpp
#include <opencv2/core.hpp>

cv::Range calcRange();

void main()
{
    cv::Mat m;
    m.begin(); // <- crash
    
    cv::Mat1b b(5, 5);
    b.colRange(calcRange()).begin() // <- crash when calcRange returns an empty Range
}
 ```

alalek:
resolves #16855

### DIFF
--- a/modules/core/include/opencv2/core/mat.inl.hpp
+++ b/modules/core/include/opencv2/core/mat.inl.hpp
@@ -1304,6 +1304,8 @@ const _Tp& Mat::at(const Vec<int, n>& idx) const
 template<typename _Tp> inline
 MatConstIterator_<_Tp> Mat::begin() const
 {
+    if (empty())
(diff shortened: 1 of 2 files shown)


alalek:
:+1:

PR #16860, opened 2020-03-20
Closes 16292: calib3d(docs): misleading/ wrong documentation for RT

### LINKED ISSUE #16292: calib3d(docs): misleading/ wrong documentation for RT
* I ~do volunteer~ would like to work on this issue.
* I ran some tests with simulation data to check my assumptions, see below. 

##### Detailed description
Many functions within the calib3d module compute or rely on the extrinsics R and T in the camera model. The rigid transformation given by (R, T) describe the change of coordinates for a point `\mathbf{p}` in object coordinates to its representation in camera coordinates: `\mathbf{p}^{cam} = \mathbf{R} \mathbf{p}^{obj} + \mathbf{t}`.

Unfortunately, there are several problems in connection with this in the documentation. 

###### Wrong Documentation
* The documentation of `calibrateCamera` states "rotation vector together with (...) translation vector (...) brings the calibration pattern from the model coordinate space (...) to the *world coordinate space*". It should rather say *brings to camera coordinates*. 

* The documentation of `stereoCalibrate` states "(...) the position and orientation of the second camera relative to the first camera. This is what the described function does. It computes ( R, T ) (...) " However, (R, T) describe the change of coordinates from cam1 to cam2, and not the position of cam2. There is a duality, see below.

* The documentations of `projectPoints`, `recoverPose`, `decomposeEssentialMat`, and `decomposeHomographyMat` are rather unexplicit.

###### Unexplicit Documentation
* The documentation for `calibrateHandEye` is explicit about the meaning of (R, T) and its direction, e.g. `R_gripper2base`. The docs on the camera models are not so explicit. 
* The `T` vector is easy to inter. However, one needs to know, that `T` is given in the camera's coordinate system. 
* The documentation on `reprojectImage3D` and `triangulatePoints` does not state the coordinate system in which the points are returned. This seems to be an issue for [many](https://answers.opencv.org/question/118966/is-cvtriangulatepoints-returning-3d-points-in-world-coordinate-system/) people.

###### Duality
The problems with the documentation wouldn't be problems, if not for the duality of poses and change of coordinates. Whereas OpenCV uses (R, T) as a change of coordinates, i.e. it relates the same vector in two different coordinate systems, other frameworks like VTK and Blender use (R, T) as the camera's pose (orientation + position), i.e. it transforms a point to generate a new point, both in the same coordinate system `\mathbf{p}_{new}^{world} = \mathbf{R} \mathbf{p}_{old}^{world} + \mathbf{t}`. These two concepts are closely related, as one is the inverse to the other, see [this](http://www.boris-belousov.net/2016/05/31/change-of-basis/). 

##### Summary
The topic of change of coordinates in OpenCV is not hard to grasp but extremely error-prone. One needs to be careful to use points in their correct coordinate representation and not mix notations and concepts. Rigorous and explicit documentation will help people to speed up their development and also consolidate their understanding of the matter. 

I also believe that a section in the modules docs or a tutorial on working with a change of coordinates will be helpful. 

##### Naming of Camera Matrix
The module's detailed description and some functions' documentation denote the camera matrix with A, whereas others use K. We should pick one and change the other ones accordingly. Furthermore, some functions show the camera matrix's form while others don't. This should be unified. The same is true for the distortion coefficients. 

##### Code I Used for Testing
I set up (virtual) cameras and calibration patterns projected points and used `solvePnP`, `CalibrateCamera` and `StereoCalibrate` to solve for extrinsics. Eventually, I compared the found extrinsic parameters to the ground truth and print the norm of the max. deviation (over all elements in R, T). The results prints values in order `e-07`. 

See complete code [here](https://gist.github.com/FischerGundlach/b6b383e32e977ca0bc0de7121d316845). This code might also transform into a tutorial? 

##### Tasks
- [x] Fork + Clone, set up local build and testing pipeline 
- [ ] Find references on concepts change of coordinates/ transformation
- [x] List all occurrences in the current module documentation that need attention
- [x] Change Occurences in
   - [x] Into + Detailed Description
   - [x] `calibrateCamera`
   - [x] `stereoCalibrate`
   - [x] `stereoRectify`
   - [x] `projectPoints`
   - [x] `recoverPose`
   - [x] `decomposeEssentialMat`
   - [x] `decomposeHomographyMat` 
   - [x] `reprojectImage3D`
   - [x] `triangulatePoints`
- [x] Only correct spelling, grammar, line length etc. if I touched thed function's documentation anyway.
- [x] [Make it a good pull request](https://github.com/opencv/opencv/wiki/How_to_contribute#making-a-good-pull-request)

FischerGundlach:
This PR closes #16292 calib3d(docs): misleading/ wrong documentation for RT

This PR introduces fixes to the documentation of the calib3d module. In addition to improving the documentation, I did minor spelling, grammar, and style guide (line length mostly) work BUT restricted these to those function's documentation that I needed to touch for this PR. 

```
force_builders_only=Docs,linux
```

### DIFF
--- a/modules/calib3d/include/opencv2/calib3d.hpp
+++ b/modules/calib3d/include/opencv2/calib3d.hpp
@@ -51,17 +51,125 @@
 /**
   @defgroup calib3d Camera Calibration and 3D Reconstruction
 
-The functions in this section use a so-called pinhole camera model. In this model, a scene view is
-formed by projecting 3D points into the image plane using a perspective transformation.
+The functions in this section use a so-called pinhole camera model. The view of a scene
+is obtained by projecting a scene's 3D point \f$P_w\f$ into the image plane using a perspective
+transformation which forms the corresponding pixel \f$p\f$. Both \f$P_w\f$ and \f$p\f$ are
+represented in homogeneous coordinates, i.e. as 3D and 2D homogeneous vector respectively. You will
+find a brief introduction to projective geometry, homogeneous vectors and homogeneous
+transformations at the end of this section's introduction. For more succinct notation, we often drop
+the 'homogeneous' and say vector instead of homogeneous vector.
 
-\f[s  \; m' = A [R|t] M'\f]
+The distortion-free projective transformation given by a  pinhole camera model is shown below.
 
-or
+\f[s \; p = A \begin{bmatrix} R|t \end{bmatrix} P_w,\f]
 
-\f[s  \vecthree{u}{v}{1} = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1}
+where \f$P_w\f$ is a 3D point expressed with respect to the world coordinate system,
+\f$p\f$ is a 2D pixel in the image plane, \f$A\f$ is the intrinsic camera matrix,
+\f$R\f$ and \f$t\f$ are the rotation and translation that describe the change of coordinates from
+world to camera coordinate systems (or camera frame) and \f$s\f$ is the projective transformation's
+arbitrary scaling and not part of the camera model.
+
+The intrinsic camera matrix \f$A\f$ (notation used as in @cite Zhang2000 and also generally notated
+as \f$K\f$) projects 3D points given in the camera coordinate system to 2D pixel coordinates, i.e.
+
+\f[p = A P_c.\f]
+
+The camera matrix \f$A\f$ is composed of the focal lengths \f$f_x\f$ and \f$f_y\f$, which are
+expressed in pixel units, and the principal point \f$(c_x, c_y)\f$, that is usually close to the
+image center:
+
+\f[A = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1},\f]
+
+and thus
+
+\f[s \vecthree{u}{v}{1} = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1} \vecthree{X_c}{Y_c}{Z_c}.\f]
+
+The matrix of intrinsic parameters does not depend on the scene viewed. So, once estimated, it can
+be re-used as long as the focal length is fixed (in case of a zoom lens). Thus, if an image from the
+camera is scaled by a factor, all of these parameters need to be scaled (multiplied/divided,
+respectively) by the same factor.
+
+The joint rotation-translation matrix \f$[R|t]\f$ is the matrix product of a projective
+transformation and a homogeneous transformation. The 3-by-4 projective transformation maps 3D points
+represented in camera coordinates to 2D poins in the image plane and represented in normalized
+camera coordinates \f$x' = X_c / Z_c\f$ and \f$y' = Y_c / Z_c\f$:
+
+\f[Z_c \begin{bmatrix}
+x' \\
+y' \\
+1
+\end{bmatrix} = \begin{bmatrix}
+1 & 0 & 0 & 0 \\
+0 & 1 & 0 & 0 \\
+0 & 0 & 1 & 0
+\end{bmatrix}
 \begin{bmatrix}
-r_{11} & r_{12} & r_{13} & t_x  \\
-r_{21} & r_{22} & r_{23} & t_y  \\
+X_c \\
+Y_c \\
+Z_c \\
+1
+\end{bmatrix}.\f]
+
+The homogeneous transformation is encoded by the extrinsic parameters \f$R\f$ and \f$t\f$ and
+represents the change of basis from world coordinate system \f$w\f$ to the camera coordinate sytem
+\f$c\f$. Thus, given the representation of the point \f$P\f$ in world coordinates, \f$P_w\f$, we
+obtain \f$P\f$'s representation in the camera coordinate system, \f$P_c\f$, by
+
+\f[P_c = \begin{bmatrix}
+R & t \\
+0 & 1
+\end{bmatrix} P_w,\f]
+
+This homogeneous transformation is composed out of \f$R\f$, a 3-by-3 rotation matrix, and \f$t\f$, a
+3-by-1 translation vector:
+
+\f[\begin{bmatrix}
+R & t \\
+0 & 1
+\end{bmatrix} = \begin{bmatrix}
+r_{11} & r_{12} & r_{13} & t_x \\
+r_{21} & r_{22} & r_{23} & t_y \\
+r_{31} & r_{32} & r_{33} & t_z \\
+0 & 0 & 0 & 1
+\end{bmatrix},
+\f]
+
+and therefore
+
+\f[\begin{bmatrix}
+X_c \\
+Y_c \\
+Z_c \\
+1
+\end{bmatrix} = \begin{bmatrix}
+r_{11} & r_{12} & r_{13} & t_x \\
+r_{21} & r_{22} & r_{23} & t_y \\
+r_{31} & r_{32} & r_{33} & t_z \\
+0 & 0 & 0 & 1
+\end{bmatrix}
+\begin{bmatrix}
+X_w \\
+Y_w \\
+Z_w \\
+1
+\end{bmatrix}.\f]
+
+Combining the projective transformation and the homogeneous transformation, we obtain the projective
+transformation that maps 3D points in world coordinates into 2D points in the image plane and in
+normalized camera coordinates:
+
+\f[Z_c \begin{bmatrix}
+x' \\
+y' \\
+1
+\end{bmatrix} = \begin{bmatrix} R|t \end{bmatrix} \begin{bmatrix}
+X_w \\
+Y_w \\
+Z_w \\
+1
+\end{bmatrix} = \begin{bmatrix}
+r_{11} & r_{12} & r_{13} & t_x \\
+r_{21} & r_{22} & r_{23} & t_y \\
 r_{31} & r_{32} & r_{33} & t_z
 \end{bmatrix}
 \begin{bmatrix}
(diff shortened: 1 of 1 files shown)


FischerGundlach:
Is there the possibility to squash before merging or do I need to do it locally and push a squashed final version?

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -2192,7 +2362,7 @@ CV_EXPORTS Mat findFundamentalMat( InputArray points1, InputArray points2,
 @param points1 Array of N (N \>= 5) 2D points from the first image. The point coordinates should
 be floating-point (single or double precision).
 @param points2 Array of the second image points of the same size and format as points1 .
-@param cameraMatrix Camera matrix \f$K = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1}\f$ .
+@param cameraMatrix Camera matrix \f$A = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1}\f$ .
```
alalek: Do we really need to rename this matrix?

It is already used in samples, like [here](https://github.com/opencv/opencv/blob/4.2.0/samples/cpp/stitching_detailed.cpp#L562). Also there are [wiki](https://en.wikipedia.org/wiki/Camera_resectioning), external [articles](http://ksimek.github.io/2013/08/13/intrinsic/).

I believe there is higher chance to find documentation about "K" camera matrix rather than about "A" (due false positive cases).

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -51,80 +51,181 @@
 /**
   @defgroup calib3d Camera Calibration and 3D Reconstruction
 
-The functions in this section use a so-called pinhole camera model. In this model, a scene view is
-formed by projecting 3D points into the image plane using a perspective transformation.
+The functions in this module use a so-called pinhole camera model. The view of a scene
```
alalek: > in this module use a so-called pinhole camera model

`fisheye` camera model is in this "module" too (in terms of OpenCV modules). Can we keep "section" here?

alalek:
["Live" patched documentation page](http://pullrequest.opencv.org/buildbot/export/pr/16860/docs/d9/d0c/group__calib3d.html#details).

/cc @catree Could you please take a look on this update? Thank you!

modules/calib3d/include/opencv2/calib3d.hpp
FischerGundlach: +1

modules/calib3d/include/opencv2/calib3d.hpp
FischerGundlach: Yes, I fully agree. Maybe one day we should take the courage and change the name in the section's introduction and accordingly in its functions. maybe :)

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -51,80 +51,165 @@
...
 
-or
+where \f$P_w\f$ is a 3D point expressed with respect to the world coordinate system,
+\f$p\f$ is a 2D pixel in the image plane, \f$A\f$ is the intrinsic camera matrix,
+\f$R\f$ and \f$t\f$ are the rotation and translation that describe the change of coordinates from
+world to camera coordinate systems and \f$s\f$ is the projective transformation's arbitrary scaling
+and not part of the camera model.
 
-\f[s  \vecthree{u}{v}{1} = \vecthreethree{f_x}{0}{c_x}{0}{f_y}{c_y}{0}{0}{1}
+The intrinsic camera matrix \f$A\f$ projects 3D points given in the camera coordinate system to 2D
```
catree:
```suggestion
The intrinsic camera matrix \f$A\f$ (notation used as in @cite Zhang2000 and also generally noted \f$K\f$) projects 3D points given in the camera coordinate system to 2D
```

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -1442,44 +1539,48 @@ CV_EXPORTS_W bool findCirclesGrid( InputArray image, Size patternSize,
...
+describtion above.
+@param stdDeviationsIntrinsics Output vector of standard deviations estimated for intrinsic
+parameters. Order of deviations values:
 \f$(f_x, f_y, c_x, c_y, k_1, k_2, p_1, p_2, k_3, k_4, k_5, k_6 , s_1, s_2, s_3,
  s_4, \tau_x, \tau_y)\f$ If one of parameters is not estimated, it's deviation is equals to zero.
-@param stdDeviationsExtrinsics Output vector of standard deviations estimated for extrinsic parameters.
- Order of deviations values: \f$(R_1, T_1, \dotsc , R_M, T_M)\f$ where M is number of pattern views,
- \f$R_i, T_i\f$ are concatenated 1x3 vectors.
+@param stdDeviationsExtrinsics Output vector of standard deviations estimated for extrinsic
+parameters. Order of deviations values: \f$(R_0, T_0, \dotsc , R_M, T_M)\f$ where M is number of
```
catree:
```suggestion
parameters. Order of deviations values: \f$(R_0, T_0, \dotsc , R_{M-1}, T_{M-1})\f$ where M is number of
```

I feel like it is overcomplicated and would just use `\f$(R_1, T_1, \dotsc , R_M, T_M)\f$`.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -1738,37 +1847,56 @@ the optimization. If CALIB_USE_INTRINSIC_GUESS is set, the coefficient from the
...
+coordinate system when given the point's coordinate representation in the first camera's coordinate
+system:
+
+\f[\begin{bmatrix}
+X_2 \\
+Y_2 \\
+Z_2 \\
+1
+\end{bmatrix} = \begin{bmatrix}
+R|T
```
catree:
```suggestion
R & T\\ 
0 & 1
```

Since the points are expressed in homogeneous coordinates.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -2485,11 +2668,20 @@ The function transforms a single-channel disparity map to a 3-channel image repr
 surface. That is, for each pixel (x,y) and the corresponding disparity d=disparity(x,y) , it
 computes:
 
-\f[\begin{array}{l} [X \; Y \; Z \; W]^T =  \texttt{Q} *[x \; y \; \texttt{disparity} (x,y) \; 1]^T  \\ \texttt{\_3dImage} (x,y) = (X/W, \; Y/W, \; Z/W) \end{array}\f]
```
catree: Missing from the original doc: how to get the 3D point by dividing by `W`.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -1738,37 +1847,56 @@ the optimization. If CALIB_USE_INTRINSIC_GUESS is set, the coefficient from the
...
+the poses of an object relative to the first camera and to the second camera,
+( \f$R_1\f$,\f$T_1\f$ ) and (\f$R_2\f$,\f$T_2\f$), respectively, for a stereo camera where the
+relative position and orientation between the two cameras are fixed, then those poses definitely
+relate to each other. This means, if the relative position and orientation (\f$R\f$,\f$T\f$) of the
+two cameras is known, it is possible to compute (\f$R_2\f$,\f$T_2\f$) when (\f$R_1\f$,\f$T_1\f$) is
+given. This is what the described function does. It computes (\f$R\f$,\f$T\f$) such that:
 
 \f[R_2=R*R_1\f]
-\f[T_2=R*T_1 + T,\f]
+\f[T_2=R*T_1 + T.\f]
```
catree: Not related to this specific change. I have already seen this before.

```suggestion
\f[T_2=R \times T_1 + T.\f]
```

`*` means usually the convolution operation.

The same changes should be done everywhere, so it is better to keep it like this.

catree:
This is great 👍 

---

I would also advice to use `K` for the camera intrinsic matrix. `A` is used in the [Zhang paper](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/tr98-71.pdf) and most of the original terminology in the OpenCV doc is taken I believe from there.
But generally, I think `K` is used:
- [ROS CameraInfo](http://docs.ros.org/api/sensor_msgs/html/msg/CameraInfo.html)
- or in various courses: [Robot Vision, Peter Corke](http://cvpr.cecs.anu.edu.au/files/petercorkekioloa2012.pdf), [Camera models  and calibration (ETHZ)](https://www.cvg.ethz.ch/teaching/compvis/2011/lecture/vision03.pdf)

---

Just some thoughts.
I feel like the notion of homogeneous coordinates, homogeneous transform is missing from the doc while it is an important notion in robotics and in computer vision.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -51,80 +51,165 @@
...
 
-The functions in this section use a so-called pinhole camera model. In this model, a scene view is
-formed by projecting 3D points into the image plane using a perspective transformation.
+The functions in this section use a so-called pinhole camera model. The view of a scene
+is obtained by projecting a scene's 3D point \f$P_w\f$ into the image plane which forms
+the corresponding pixel \f$p\f$. The distortion-free projective transformation given by
+such a camera model is shown below.
 
-\f[s  \; m' = A [R|t] M'\f]
+\f[s \; p = A \begin{bmatrix} R|t \end{bmatrix} P_w,\f]
```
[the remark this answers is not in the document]
FischerGundlach: No worries, my OCD wanted the same. Unfortunately, the rendering is quite unpleasant.
Screenshot from 2020-03-22 10-56-02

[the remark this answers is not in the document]
FischerGundlach: image

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -51,80 +51,165 @@
 /**
   @defgroup calib3d Camera Calibration and 3D Reconstruction
 
-The functions in this section use a so-called pinhole camera model. In this model, a scene view is
-formed by projecting 3D points into the image plane using a perspective transformation.
+The functions in this section use a so-called pinhole camera model. The view of a scene
+is obtained by projecting a scene's 3D point \f$P_w\f$ into the image plane which forms
```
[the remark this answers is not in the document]
FischerGundlach: +1

modules/calib3d/include/opencv2/calib3d.hpp
FischerGundlach: +1

modules/calib3d/include/opencv2/calib3d.hpp
FischerGundlach: my idea for this enumeration was that both c++ and python use 0-based-numbering. 

I moved this enumeration from the original rvecs description, as it didn't fit in well at that point. 
>  ... calibration pattern in the k-th pattern view (k=0.. M -1)

FischerGundlach: *though yes, it needs to be M-1 anyways

modules/calib3d/include/opencv2/calib3d.hpp
FischerGundlach: +1

modules/calib3d/include/opencv2/calib3d.hpp
FischerGundlach: now that I look at it, shouldn't we simply do:

```
\f[R_2=R R_1\f]
\f[T_2=R T_1 + T.\f]
```

FischerGundlach: ?

FischerGundlach:
@catree @alalek Thank you for your feedback and time!

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -181,8 +336,62 @@ pattern (every view is described by several 3D-2D point correspondences).
 -   Estimate the relative position and orientation of the stereo camera "heads" and compute the
 *rectification* transformation that makes the camera optical axes parallel.
 
+<B> Homogeneous Coordinates </B><br>
```
FischerGundlach: I added a short part on homogeneous coordinates and sprinkle in the words projective transformation, homogeneous transformation and homogeneous coordinates in the description of the camera model.

modules/calib3d/include/opencv2/calib3d.hpp
FischerGundlach: I added some parts on this mapping.

alalek:
@catree Thank you for the review!

@FischerGundlach Could you please squash your commits into one? (I would like to backport this patch onto 3.4 branch after merge)

FischerGundlach:
> Could you please squash your commits into one? (I would like to backport this patch onto 3.4 branch after merge)

sure, do you need it asap? otherwise, I would squash everything once all discussions are resolved.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -51,80 +51,165 @@
...
+r_{31} & r_{32} & r_{33} & t_z
+\end{bmatrix}
+\begin{bmatrix}
+X_w \\
+Y_w \\
+Z_w \\
+1
+\end{bmatrix}.\f]
+
+Without loss of generality, we will say that the matrix of extrinsic parameters gives the change
```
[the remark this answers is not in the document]
FischerGundlach: I agree that mentioning calibrateCamera here is misleading. I added a part on homogeneous transformationns

```As mentioned, by using homogeneous coordinates we can express any change of basis parameterized by
\f$R\f$ and \f$t\f$ as a linear transformation, e.g. for the change of basis from coordinate system
0 to coordinate system 1 becomes:

\f[P_1 = R P_0 + t \rightarrow P_{h_1} = \begin{bmatrix}
R & t \\
0 & 1
\end{bmatrix} P_{h_0}.\f]
```

further down in the documentation and removed the reference to calibrateCamera. I think it is clear from calibrateCamera's documentation.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -51,80 +51,223 @@
...
+
+The joint rotation-translation matrix \f$[R|t]\f$ is the matrix product of a projective
+transformation and a homogeneous transformation. The 3-by-4 projective transformation maps 3D points
+represented in camera coordinates to 2D points in the image plane and represented in camera
+coordinates:
+
+\f[\begin{bmatrix}
+X_c \\
+Y_c \\
+Z_c
```
catree:
```suggestion
transformation and a homogeneous transformation. The 3-by-4 projective transformation maps 3D points
represented in camera coordinates to 2D points in the image plane and represented in normalized camera
coordinates:

\f[\begin{bmatrix}
x' \\
y' \\
1
```

I think it is a little bit confusing to have in both side almost the "same thing" for a beginner, and this would introduce normalized camera coordinates: x' = X_c / Z_c ; y' = Y_c / Z_c

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -181,8 +330,63 @@ pattern (every view is described by several 3D-2D point correspondences).
...
+\end{bmatrix} \rightarrow \begin{bmatrix}
+X / W \\
+Y / W
+\end{bmatrix},\f]
+
+if \f$W \ne 0\f$.
+
+Due to this mapping, all multiples \f$k P_h\f$, for \f$k \ne 0\f$, of a homogeneous point represent
+the same point \f$P_h\f$. An intuitive understanding of this property is, that under a projective
+transformation all multiples of \f$P_h\f$ are mapped to the same point. This is the physical
```
catree:
```suggestion
the same point \f$P_h\f$. An intuitive understanding of this property is that under a projective
transformation, all multiples of \f$P_h\f$ are mapped to the same point. This is the physical
```

modules/calib3d/include/opencv2/calib3d.hpp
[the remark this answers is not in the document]
catree: OK, I see.

modules/calib3d/include/opencv2/calib3d.hpp
catree: Yes, it is simpler like that.

modules/calib3d/include/opencv2/calib3d.hpp
```diff
@@ -181,8 +330,63 @@ pattern (every view is described by several 3D-2D point correspondences).
...
+Y \\
+Z
+\end{bmatrix} \rightarrow \begin{bmatrix}
+X \\
+Y \\
+Z \\
+1
+\end{bmatrix}.\f]
+
+For the inverse mapping \f$P_h \rightarrow P\f$, one devides all elements of the homogeneous vector
```
catree:
```suggestion
For the inverse mapping \f$P_h \rightarrow P\f$, one divides all elements of the homogeneous vector
```

FischerGundlach:
@alalek @catree All discussions resolved and squashed.

PR #16864, opened 2020-03-21
imgproc: add src.empty() checks in filter operations

### LINKED ISSUE #16857: cv::blur on an empty Submatrix leads to an infinite loop
- OpenCV => 4.2

##### Detailed description
When calling cv::blur on an empty subranges an infinite loop is caused in cv::borderInterpolate. This will probably also occur with other functions that use FilterEngine internally. This did not happen with  OpenCV 3.1.

##### Steps to reproduce
```.cpp
int main()
{
  cv::Mat1b m(1,1), o;
  cv::blur(m.colRange(cv::Range()), o, {3, 3});
}
```

alalek:
resolves #16857

Throws exception on empty input - see #8300

opencv_contrib tests are good :+1:


alalek:
:+1:

Issue #11019, opened 2018-03-06
OE-24 "Revised HighGUI Module" discussion


vpisarev:
the feature request about evolution proposal [OE-24](https://github.com/opencv/opencv/wiki/OE-24.-Module-HighGUI)

paroj:
[imgui](https://github.com/ocornut/imgui) gives you both OpenGL buttons as well as (optionally embedded) font loading.
There are example backends for OpenGL as well as for DirectX, but why bother with DirectX?

catree:
[GLFW](http://www.glfw.org/) + imgui similar to the approach used by [librealsense (realsense-viewer)](https://github.com/IntelRealSense/librealsense/tree/v2.10.1/tools/realsense-viewer) maybe?

LaurentBerger:
What do we want to do with highgui module? menu, button, slider, file browser

LaurentBerger:
do we want a highui module for all platforms? android, ios, linux, windows

LaurentBerger:
Must we use a third party lib ?

[wikipedia list ](https://en.wikipedia.org/wiki/Nana_(C%2B%2B_library))and previous post

[gtkmm ](https://www.gtkmm.org/fr/index.html)
[FLTK](http://www.fltk.org/index.php)
[FOX toolkit](http://www.fox-toolkit.org/)
[Juce](https://juce.com/)
[Qt](https://www1.qt.io/developers/)
[wxWidgets ](https://www.wxwidgets.org/)
[Ultimate++](https://www.ultimatepp.org/)
[openscenegraph](www.openscenegraph.org/)
[imgui](https://github.com/ocornut/imgui)
[GLFW](http://www.glfw.org/)

ebachard:
I confirm ImGui works very well at least on Windows and Linux. Should work on Mac OS X two.

As example, I'm mixing OpenCV + ImGui + OpenGL + SDL2 + NativeFileDialog  + AweSomeFont  an more in miniDart ( see details here : **https://github.com/ebachard/miniDart** ).

Last, there are other projects who worth a try: 
- cvui : **https://github.com/Dovyski/cvui**
- canvascv : **https://github.com/sagi-z/CanvasCV**

vpisarev:
ImGui is definitely the first thing to look it; probably, it's too much for us, but the license should let us extract some essential parts out of it.

@LaurentBerger@
* menu - no
* button - yes
* slider - already there
* file browser - no
* on the other items - they are actually listed in the proposal

[...]

PR #16880, opened 2020-03-22
Added Dashed Rectangle feature in Drawing.cpp

VisheshKhosla:

### DIFF
--- a/modules/imgproc/src/drawing.cpp
+++ b/modules/imgproc/src/drawing.cpp
@@ -1893,19 +1893,19 @@ void dash_rectangle(InputOutputArray _img, Point pt1, Point pt2,
     while (startX + 2 < endX) {
         Point p1(startX, pt1.y);
         Point p2(startX + 2, pt1.y);
-        line(_img, p1, p2, color, thickness, lineType = 8, shift = 0);
+        line(_img, p1, p2, color, thickness, lineType = lineType, shift = shift);
         Point p3(startX, pt2.y);
         Point p4(startX + 2, pt2.y);
-        line(_img, p3, p4, color, thickness, lineType = 8, shift = 0);
+        line(_img, p3, p4, color, thickness, lineType = lineType, shift = shift);
         startX += gap+2;
     }
     while (startY + 2 < endY) {
         Point p1(pt1.x, startY);
         Point p2(pt1.x, startY + 2);
-        line(_img, p1, p2, color, thickness, lineType = 8, shift = 0);
+        line(_img, p1, p2, color, thickness, lineType = lineType, shift = shift);
         Point p3(pt2.x, startY);
         Point p4(pt2.x, startY + 2);
-        line(_img, p3, p4, color, thickness, lineType = 8, shift = 0);
+        line(_img, p3, p4, color, thickness, lineType = lineType, shift = shift);
         startY += gap + 2;
 
     }
@@ -1914,10 +1914,10 @@ void dash_rectangle(InputOutputArray _img, Point pt1, Point pt2,
     Point v3(endX, startY);
     Point v4(endX, endY);
     // Plotting the vertices
-    line(_img, v1, v1, color, thickness, lineType = 8, shift = 0);
-    line(_img, v2, v2, color, thickness, lineType = 8, shift = 0);
-    line(_img, v3, v3, color, thickness, lineType = 8, shift = 0);
-    line(_img, v4, v4, color, thickness, lineType = 8, shift = 0);
+    line(_img, v1, v1, color, thickness, lineType = lineType, shift = shift);
+    line(_img, v2, v2, color, thickness, lineType = lineType, shift = shift);
+    line(_img, v3, v3, color, thickness, lineType = lineType, shift = shift);
+    line(_img, v4, v4, color, thickness, lineType = lineType, shift = shift);
 }
 
 
--- a/modules/imgproc/src/drawing.cpp
+++ b/modules/imgproc/src/drawing.cpp
@@ -1099,6 +1099,7 @@ static inline void ICV_HLINE(uchar* ptr, int xl, int xr, const void* color, int 
 }
 //end ICV_HLINE()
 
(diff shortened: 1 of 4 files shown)


VisheshKhosla:
void dash_rectangle(InputOutputArray _img, Point pt1, Point pt2,
    const Scalar& color, int thickness = 1,
    int lineType = 8, int shift = 0,int gap=5) 
Where the gap is the distance between the dashes.

VisheshKhosla:
It shows that  docs have failed the merging process.Can i know more about this?

alalek:
> docs have failed the merging process

This is not true. It is not merging failure.

---

> know more about this

Yes, read https://github.com/opencv/opencv/wiki/How_to_contribute till the end.

---

> Merge pull request #

These commits will not be accepted. Remove them.

modules/imgproc/src/drawing.cpp
```diff
@@ -1879,6 +1880,45 @@ void rectangle( InputOutputArray img, Rect rec,
         rectangle( img, rec.tl(), rec.br() - Point(1<<shift,1<<shift),
                    color, thickness, lineType, shift );
 }
+void dash_rectangle(InputOutputArray _img, Point pt1, Point pt2,
+    const Scalar& color, int thickness = 1,
+    int lineType = 8, int shift = 0,int gap=5) {
```
dkurt: Please use enum values https://docs.opencv.org/master/d6/d6e/group__imgproc__draw.html#gaf076ef45de481ac96e0ab3dc2c29a777 instead of int `8`

modules/imgproc/src/drawing.cpp
```diff
@@ -1879,6 +1880,45 @@ void rectangle( InputOutputArray img, Rect rec,
...
+    }
+    Point v1(startX, startY);
+    Point v2(startX, endY);
+    Point v3(endX, startY);
+    Point v4(endX, endY);
+    // Plotting the vertices
+    line(_img, v1, v1, color, thickness, lineType = 8, shift = 0);
+    line(_img, v2, v2, color, thickness, lineType = 8, shift = 0);
+    line(_img, v3, v3, color, thickness, lineType = 8, shift = 0);
+    line(_img, v4, v4, color, thickness, lineType = 8, shift = 0);
```
dkurt: Drawing points as lines is not a good idea due it's not convenient if this line has length or not. So nothing can be drawn as well. I would like to recommend to replace to filled rectangle

modules/imgproc/src/drawing.cpp
```diff
@@ -1879,6 +1880,45 @@ void rectangle( InputOutputArray img, Rect rec,
...
+    thickness = 1;
+    int startX = pt1.x;
+    int endX = pt2.x;
+    int startY = pt1.y;
+    int endY = pt2.y;
+
+    while (startX + 2 < endX) {
+        Point p1(startX, pt1.y);
+        Point p2(startX + 2, pt1.y);
+        line(_img, p1, p2, color, thickness, lineType = 8, shift = 0);
```
asmorkalov, line 1896: `lineType` should be the same as the function parameter.

asmorkalov: The same for `shift` I suspect.

modules/imgproc/src/drawing.cpp
```diff
@@ -1879,6 +1880,45 @@ void rectangle( InputOutputArray img, Rect rec,
...
+        Point p3(startX, pt2.y);
+        Point p4(startX + 2, pt2.y);
+        line(_img, p3, p4, color, thickness, lineType = 8, shift = 0);
```
asmorkalov, line 1899: `lineType`, `shift`

modules/imgproc/src/drawing.cpp
```diff
@@ -1879,6 +1880,45 @@ void rectangle( InputOutputArray img, Rect rec,
...
+        startX += gap+2;
+    }
+    while (startY + 2 < endY) {
+        Point p1(pt1.x, startY);
+        Point p2(pt1.x, startY + 2);
+        line(_img, p1, p2, color, thickness, lineType = 8, shift = 0);
```
asmorkalov, line 1905: `lineType`, `shift`

modules/imgproc/src/drawing.cpp
```diff
@@ -1879,6 +1880,45 @@ void rectangle( InputOutputArray img, Rect rec,
...
+        Point p3(pt2.x, startY);
+        Point p4(pt2.x, startY + 2);
+        line(_img, p3, p4, color, thickness, lineType = 8, shift = 0);
```
asmorkalov, line 1908: `lineType`, `shift`

asmorkalov:
@VisheshKhosla Could you also add dashed rectangle case to `samples/cpp/drawing.cpp` and `samples/python/drawing.py` to demonstrate the new functionality.

asmorkalov:
@VisheshKhosla do you plan to finish the PR in mean time?

VisheshKhosla:
> @VisheshKhosla do you plan to finish the PR in mean time?

Yes sure,i am really sorry for the delay.

modules/imgproc/src/drawing.cpp
VisheshKhosla: I have changed it now.

asmorkalov:
@vpisarev What is your opinion?

vpisarev:
@VisheshKhosla, thank you for the contribution! I think, we need to carefully design and then implement this functionality with extended drawing primitives. Just adding functions one-by-one is not a good idea, that will result in a bloated and inconsistent API. Please, look at the following evolution proposal: https://github.com/opencv/opencv/wiki/OE-24.-Module-HighGUI. I've added there items about dashed figures. Let me close this PR for now, and you are welcome to the discussion thread about the evolution proposal (https://github.com/opencv/opencv/issues/11019). Someone needs to compose a detailed API proposal, then it should be discussed and then we should start working on the implementation.

PR #16886, opened 2020-03-23
G-API: DISABLED_ in tests instead #if 0

mpashchenkov:
### DISABLED_ in tests instead #if 0

`#if 0` isn't the right way to disable tests. `DISABLED_` is the right one (in gtest). 

<cut/>

Added:
- `DISABLED_` for tests under `#if 0`
- tests in `gapi_core_test_fluid.cpp` from CPU's tests from `#if 0`.
- GMatDesc's tests with own::Mat

Removed:
- `#if 0` in tests

```
```


modules/gapi/test/cpu/gapi_core_tests_fluid.cpp
```diff
@@ -285,260 +285,4 @@ INSTANTIATE_TEST_CASE_P(CopyTestFluid, CopyTest,
                                        cv::Size(128, 128)),
                                 Values(-1),
                                 Values(CORE_FLUID)));
-
-//----------------------------------------------------------------------
-// FIXME: Clean-up test configurations which are enabled already
```
Volskig: @anton-potapov  Tests with the same name located in `gapi_core_test_cpu.cpp`.  I deleted them from here. Is it right way? Please review.

Volskig:
@elatkin please review.

modules/gapi/test/cpu/gapi_core_tests_fluid.cpp
elatkin: @Volskig  
I think the FIXME comment here assumes double-checking each test on the case-by case basis: if the corresponding ...TestFluid case is already enabled. If you have done this case-by-case check, then I think it is OK to simply remove the redundant cases here.

modules/gapi/test/cpu/gapi_core_tests_fluid.cpp
```diff
@@ -285,260 +285,4 @@ INSTANTIATE_TEST_CASE_P(CopyTestFluid, CopyTest,
...
-                                Values(cv::Size(1280, 720),
-                                       cv::Size(640, 480),
-                                       cv::Size(128, 128))));
-INSTANTIATE_TEST_CASE_P(ConcatVertTestCPU, ConcatVertTest,
-                        Combine(Values(CV_8UC1, CV_16UC1, CV_16SC1),
-                                Values(cv::Size(1280, 720),
-                                       cv::Size(640, 480),
-                                       cv::Size(128, 128))));
-
-//----------------------------------------------------------------------
```
dmatveev: So do we really have all these tests running with Fluid now?

modules/gapi/test/cpu/gapi_core_tests_fluid.cpp
dmatveev: +1

modules/gapi/test/cpu/gapi_core_tests_fluid.cpp
Volskig: Added unused tests. Fixed.

modules/gapi/test/cpu/gapi_core_tests_fluid.cpp
```diff
@@ -119,6 +148,24 @@ INSTANTIATE_TEST_CASE_P(AddWeightedTestFluid, AddWeightedTest,
...
+                                Values(AbsToleranceScalar(1e-5).to_compare_obj()),
+                                Values(NORM_INF, NORM_L1, NORM_L2)));
+
+INSTANTIATE_TEST_CASE_P(IntegralTestFluid, IntegralTest,
+                        Combine(Values(CV_8UC1, CV_16UC1, CV_16SC1),
+                                Values(cv::Size(1280, 720),
+                                       cv::Size(640, 480),
+                                       cv::Size(128, 128)),
+                                Values(-1),
+                                Values(CORE_FLUID)));
```
dmatveev, line 167: We don't have Integral kernel in Fluid.

And when this kernel is not found, G-API takes the OpenCV version by default.

We need to forbid G-API taking the default kernel package when running accuracy test - otherwise, we can't guarantee that a proper kernel is used.

Please turn all `CORE_FLUID`, `IMGPROC_FLUID`, and other macros for all backend's accuracy tests to use `cv:;gapi::use_only` [modifier](https://docs.opencv.org/4.2.0/d6/d25/structcv_1_1gapi_1_1use__only.html)

dmatveev: Please also note - this `use_only` should apply to accuracy tests only! Other functional and unit tests can still rely on the default kernel package.

Volskig: Fixed. 
Added DISABLED_ for tests with unsupported kernels.

dmatveev: Cool! Thanks

PR #16887, opened 2020-03-23
ONNX: Add support for Resnet_backbone (Torchvision)

### LINKED ISSUE #16876: Loading ONNX model producing error

- OpenCV => 3.4.9

##### Detailed description

Loading ONNX model producing error

cv2.error: OpenCV(3.4.10-pre) /home/Jarvis/opencv/modules/dnn/src/dnn.cpp:574: error: (-2:Unspecified error) Can't create layer "291" of type "Sqrt" in function 'getLayerInstance'

##### Steps to reproduce

    import cv2
    net = cv2.dnn.readNet('resnet_backbone.onnx')

Here is the code to convert the model

```
import torch
from torch.autograd import Variable
import torch.nn as nn
from torchvision.models.detection.backbone_utils import *

model = resnet_fpn_backbone('resnet50', True)
x = Variable(torch.randn( 1, 3, 256, 256))
model.eval()
torch.onnx.export(model, x, "resnet_backbone.onnx")
```

Here is the link to the [model](https://drive.google.com/open?id=1xGBQ0aNsG3XD9p7G3D8QVPBkndH_f48P)

ashishkrshrivastava:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/734

resolves #16876

```
opencv_extra=fasterrcnn
```

### This PR adds
Support to resnet_backbone by adding 
1. **fusion of Unfused Frozen BatchNorm.**
2. **fusion of unfused Upsample with two inputs.**

Unfused Version of batchnorm. Latest version of pytorch ( 1.4 )
Screenshot from 2020-03-28 14-24-46
Below for are node properties of Reshape nodes
Screenshot from 2020-03-28 14-30-07
Screenshot from 2020-03-28 14-30-22
Screenshot from 2020-03-28 14-30-34
Screenshot from 2020-03-28 14-31-53

Below attached Fused version of Batchnorm to show comparison between unfused and fused batchnorm.
Initializers inputs of Reshape of above unfused subgraph acts same as Initializer inputs of Fused batchnorm.
Screenshot from 2020-03-28 14-36-06

**Initializers** are not considered as nodes.

One more example of **Initializer** is Weight and bias of Convolution node

Screenshot from 2020-03-28 14-52-34

And Below mentioned **Upsample** subgraph with two inputs.

Screenshot from 2020-04-05 14-53-17

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/onnx/onnx_graph_simplifier.cpp
+++ b/modules/dnn/src/onnx/onnx_graph_simplifier.cpp
@@ -382,6 +382,36 @@ public:
     }
 };
 
+class BatchNormalizationSubgraph : public Subgraph
+{
+public:
+    BatchNormalizationSubgraph()
+    {
+        int input = addNodeToMatch("");
+        int data1 = addNodeToMatch("Constant");
+        int data2 = addNodeToMatch("Constant");
+        int data3 = addNodeToMatch("Constant");
+        int data4 = addNodeToMatch("Constant");
+        int shape1 = addNodeToMatch("Constant");
+        int reshape1 = addNodeToMatch("Reshape", data1, shape1);
+        int shape2 = addNodeToMatch("Constant");
+        int reshape2 = addNodeToMatch("Reshape", data2, shape2);
+        int shape3 = addNodeToMatch("Constant");
+        int reshape3 = addNodeToMatch("Reshape", data3, shape3);
+        int shape4 = addNodeToMatch("Constant");
+        int reshape4 = addNodeToMatch("Reshape", data4, shape4);
+        int sqrtNode = addNodeToMatch("Sqrt", reshape3);
+        int A = addNodeToMatch("Constant");
+        int divNode = addNodeToMatch("Div", A, sqrtNode);
+        int mul1 = addNodeToMatch("Mul", reshape1, divNode);
+        int mul2 = addNodeToMatch("Mul", reshape4, mul1);
+        int sub = addNodeToMatch("Sub", reshape2, mul2);
+        int mul3 = addNodeToMatch("Mul", input, mul1);
+        addNodeToMatch("Add", mul3, sub);
+        setFusedNode("BatchNormalization", input, data1, data2, data4 ,data3);
+    }
+};
+
 void simplifySubgraphs(opencv_onnx::GraphProto& net)
 {
     std::vector<Ptr<Subgraph> > subgraphs;
@@ -394,6 +424,7 @@ void simplifySubgraphs(opencv_onnx::GraphProto& net)
     subgraphs.push_back(makePtr<NormalizeSubgraph1>());
     subgraphs.push_back(makePtr<NormalizeSubgraph2>());
     subgraphs.push_back(makePtr<NormalizeSubgraph3>());
+    subgraphs.push_back(makePtr<BatchNormalizationSubgraph>());
 
     simplifySubgraphs(Ptr<ImportGraphWrapper>(new ONNXGraphWrapper(net)), subgraphs);
 }
--- a/modules/dnn/src/onnx/onnx_importer.cpp
+++ b/modules/dnn/src/onnx/onnx_importer.cpp
@@ -309,11 +309,30 @@ static void addConstant(const std::string& name,
     outShapes.insert(std::make_pair(name, shape(blob)));
 }
 
+void addConstantNodesForInitializers(opencv_onnx::GraphProto& graph_proto)
+{
+    int num_initializers = graph_proto.initializer_size();
+    for (int id = 0; id < num_initializers; id++)
+    {
+        opencv_onnx::TensorProto initializer = graph_proto.initializer(id);
+        opencv_onnx::NodeProto* constant_node = graph_proto.add_node();
+        constant_node->set_op_type("Constant");
+        constant_node->set_name(initializer.name());
+        constant_node->add_output(initializer.name());
+        opencv_onnx::AttributeProto* value = constant_node->add_attribute();
+        opencv_onnx::TensorProto* tensor = initializer.New();
+        tensor->CopyFrom(initializer);
+        releaseONNXTensor(initializer);
+        value->set_allocated_t(tensor);
+    }
+}
+
 void ONNXImporter::populateNet(Net dstNet)
 {
     CV_Assert(model_proto.has_graph());
     opencv_onnx::GraphProto graph_proto = model_proto.graph();
 
+    addConstantNodesForInitializers(graph_proto);
     simplifySubgraphs(graph_proto);
 
     std::map<std::string, Mat> constBlobs = getGraphTensors(graph_proto);
--- a/modules/dnn/test/test_onnx_importer.cpp
+++ b/modules/dnn/test/test_onnx_importer.cpp
@@ -290,6 +290,15 @@ TEST_P(Test_ONNX_layers, BatchNormalization3D)
     testONNXModels("batch_norm_3d");
 }
 
+TEST_P(Test_ONNX_layers, BatchNormalizationUnfused)
+{
+    if (backend == DNN_BACKEND_INFERENCE_ENGINE_NN_BUILDER_2019)
+        applyTestTag(CV_TEST_TAG_DNN_SKIP_IE_NN_BUILDER);
+    if (backend == DNN_BACKEND_INFERENCE_ENGINE_NGRAPH)
+        applyTestTag(CV_TEST_TAG_DNN_SKIP_IE_NGRAPH);
+    testONNXModels("frozenBatchNorm2d");
+}
+
 TEST_P(Test_ONNX_layers, Transpose)
 {
     if (backend == DNN_BACKEND_INFERENCE_ENGINE_NN_BUILDER_2019)
@@ -370,6 +379,16 @@ TEST_P(Test_ONNX_layers, ResizeUnfused)
     testONNXModels("resize_nearest_unfused_opset11_torch1.4");
     testONNXModels("resize_nearest_unfused_opset11_torch1.3");
     testONNXModels("resize_bilinear_unfused_opset11_torch1.4");
+}
+
+TEST_P(Test_ONNX_layers, ResizeUnfusedTwoInputs)
+{
+    if (backend == DNN_BACKEND_INFERENCE_ENGINE_NN_BUILDER_2019)
+        applyTestTag(CV_TEST_TAG_DNN_SKIP_IE_NN_BUILDER);
+    if (backend == DNN_BACKEND_INFERENCE_ENGINE_NGRAPH)
+        applyTestTag(CV_TEST_TAG_DNN_SKIP_IE_NGRAPH);
+    testONNXModels("upsample_unfused_two_inputs_opset9_torch1.4", npy, 0, 0, false, true, 2);
+    testONNXModels("upsample_unfused_two_inputs_opset11_torch1.4", npy, 0, 0, false, true, 2);
 }
 
 TEST_P(Test_ONNX_layers, MultyInputs)


alalek:
@ashishkrshrivastava Could you please rebase opencv_extra patch (to resolve conflict)?

ashishkrshrivastava:
@alalek I was really really busy for last 2 days. I will complete it by today.

modules/dnn/src/graph_simplifier.cpp
```diff
@@ -111,7 +111,7 @@ bool Subgraph::match(const Ptr<ImportGraphWrapper>& net, int nodeId,
 
         for (int j = 0; j < inputNodes.size(); ++j)
         {
-            if (nodes[inputNodes[j]].empty())  // Unknown input node type.
+            if (nodes[inputNodes[j]].empty() || nodes[inputNodes[j]] == "Initializer")  // Unknown input node type.
```
dkurt: Can you please add some note in PRs description what is Initializer and from which version of PyTorch it appears?

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -290,9 +294,77 @@ class ResizeSubgraph2 : public ExtractScalesSubgraph
...
+
+        int slice = addNodeToMatch("Slice", shape);
+        int cast2 = addNodeToMatch("Cast", slice);
+
+        int div = addNodeToMatch("Div", cast1, cast2);
+        int concat = addNodeToMatch("Concat", constant, div);
+
+        addNodeToMatch("Upsample", input1, concat);
+        setFusedNode("Upsample", input1, input2);
+    }
```
dkurt: Can we inherit it from `ExtractScalesSubgraph`?

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -1116,6 +1116,17 @@ void ONNXImporter::populateNet(Net dstNet)
                 layerParams.set("zoom_factor_y", scales.getIntValue(2));
                 layerParams.set("zoom_factor_x", scales.getIntValue(3));
             }
+            else if(node_proto.input_size() == 2 && constBlobs.find(node_proto.input(1)) == constBlobs.end())
```
dkurt: This way i think it's not critical if second input will be non-constant. That way it works as "resize first input as second".

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -1116,6 +1116,17 @@ void ONNXImporter::populateNet(Net dstNet)
...
+            {
+                // upsample with two inputs
+                shapeIt = outShapes.find(node_proto.input(0));
+                MatShape inpShape1 = shapeIt->second;
+                shapeIt = outShapes.find(node_proto.input(1));
+                MatShape inpShape2 = shapeIt->second;
+                layerParams.set("zoom_factor_y", inpShape2[2]/inpShape1[2]);
+                layerParams.set("zoom_factor_x", inpShape2[3]/inpShape1[3]);
```
dkurt: Please check that number of dimensions equals 4 before getting it:
```
CV_CheckEQ(inpShape1.size(), 4, "Only resize for 4D tensors is supported.")
CV_CheckEQ(inpShape2.size(), 4, "Only resize for 4D tensors is supported.")
```

modules/dnn/src/graph_simplifier.cpp
dkurt: Can you please also add some tests for new functionality?

ashishkrshrivastava: I have added test for **FrozenBatcnNorm**.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: I think no. this subgraph is quite different than `ExtractScalesSubgraph`. The main thiing is scales are not constant input of `Mul` nodes unlike `ExtractScalesSubgraph`.

modules/dnn/src/onnx/onnx_importer.cpp
ashishkrshrivastava: yeah, you are right. Second input need not to be non-constant. I have changed it. Please have a look now.

modules/dnn/src/graph_simplifier.cpp
ashishkrshrivastava: > I have added test for **FrozenBatcnNorm** and **UpsampleUnfused** with two inputs

dkurt: Can you please check if the Initializer nodes will remain if before export, `model.eval()` is used? It seems to me that model in training state.

modules/dnn/src/onnx/onnx_importer.cpp
dkurt: Can you please fix it?

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -1116,6 +1116,19 @@ void ONNXImporter::populateNet(Net dstNet)
                 layerParams.set("zoom_factor_y", scales.getIntValue(2));
                 layerParams.set("zoom_factor_x", scales.getIntValue(3));
             }
+            else if(node_proto.input_size() == 2 && (outShapes.find(node_proto.input(1))->second).size() == 4)
```
dkurt, line 1119: What if second condition is false? Do we have working branch below?

dkurt: please add a whitespace after `if`

dkurt:
@ashishkrshrivastava, thanks for keep working on that! I'm wondering if it's possible to make calculations over constants during import and then just extend Resize parsing, what do you think? Because it looks like that we are trying for now very specific thing but can make subgraph to calculate target size. Can you please check which nodes from the following resize subgraph are computed properly in runtime and which are not?

image

ashishkrshrivastava:
@dkurt, we just need some changes in onnx_importer.cpp in order to run this version of upsample/Resize subgraph. No need of fusion. 
The changes are mentioned below:

```cpp
else if (layer_type == "Resize")
        {
            for (int i = 1; i < node_proto.input_size(); i++)
                CV_Assert(layer_id.find(node_proto.input(i)) == layer_id.end());

            String interp_mode = layerParams.get<String>("coordinate_transformation_mode");
            CV_Assert_N(interp_mode != "tf_crop_and_resize",
                        interp_mode != "tf_half_pixel_for_nn");

            layerParams.set("align_corners", interp_mode == "align_corners");
            Mat shapes = getBlob(node_proto, constBlobs, node_proto.input_size() - 1);
            CV_CheckEQ(shapes.size[0], 4, "");
            CV_CheckEQ(shapes.size[1], 1, "");
            CV_CheckTypeEQ(shapes.depth(), CV_32S, "");
            int height = shapes.at<int>(2);
            int width  = shapes.at<int>(3);
            if (node_proto.input_size() == 3)
            {
                shapeIt = outShapes.find(node_proto.input(0));
                CV_Assert(shapeIt != outShapes.end());
                MatShape scales = shapeIt->second;
                height *= scales[2];
                width  *= scales[3];
            }
            layerParams.set("width", width);
            layerParams.set("height", height);

            if (layerParams.get<String>("mode") == "linear") {
                layerParams.set("mode", interp_mode == "pytorch_half_pixel" ?
                                        "opencv_linear" : "bilinear");
            }
            replaceLayerParam(layerParams, "mode", "interpolation");
        }
        else if (layer_type == "Upsample")
        {
            layerParams.type = "Resize";
            if (layerParams.has("scales"))
            {
                // Pytorch layer
                DictValue scales = layerParams.get("scales");
                CV_Assert(scales.size() == 4);
                layerParams.set("zoom_factor_y", scales.getIntValue(2));
                layerParams.set("zoom_factor_x", scales.getIntValue(3));
            }
            else
            {
                // Caffe2 layer
                replaceLayerParam(layerParams, "height_scale", "zoom_factor_y");
                replaceLayerParam(layerParams, "width_scale", "zoom_factor_x");
            }
            replaceLayerParam(layerParams, "mode", "interpolation");

            if (node_proto.input_size() == 2 || (layerParams.get<String>("interpolation") == "linear" && framework_name == "pytorch")) {
                layerParams.type = "Resize";
                Mat scales = getBlob(node_proto, constBlobs, 1);
                CV_Assert(scales.total() == 4);
                if (layerParams.get<String>("interpolation") == "linear" && framework_name == "pytorch")
                    layerParams.set("interpolation", "opencv_linear");
                layerParams.set("zoom_factor_y", scales.at<float>(2));
                layerParams.set("zoom_factor_x", scales.at<float>(3));
            }
        }
```

Change 1. 
```cpp
CV_Assert_N(interp_mode != "tf_crop_and_resize",
                        interp_mode != "tf_half_pixel_for_nn");
```
from
```cpp
CV_Assert_N(interp_mode != "tf_crop_and_resize", iinterp != "assymetric",
                        interp_mode != "tf_half_pixel_for_nn");
```

change 2..
```cpp
if (node_proto.input_size() == 2 || (layerParams.get<String>("interpolation") == "linear" && framework_name == "pytorch")) {
                layerParams.type = "Resize";
                Mat scales = getBlob(node_proto, constBlobs, 1);
                CV_Assert(scales.total() == 4);
                if (layerParams.get<String>("interpolation") == "linear" && framework_name == "pytorch")
                    layerParams.set("interpolation", "opencv_linear");
                layerParams.set("zoom_factor_y", scales.at<float>(2));
                layerParams.set("zoom_factor_x", scales.at<float>(3));
            }
```
from 
```cpp
if (layerParams.get<String>("interpolation") == "linear" && framework_name == "pytorch") {
                layerParams.type = "Resize";
                Mat scales = getBlob(node_proto, constBlobs, 1);
                CV_Assert(scales.total() == 4);
                layerParams.set("interpolation", "opencv_linear");
                layerParams.set("zoom_factor_y", scales.at<float>(2));
                layerParams.set("zoom_factor_x", scales.at<float>(3));
            }
```

> Can you please check which nodes from the following resize subgraph are computed properly in runtime and which are not?

So each node is working properly but resize and upsample, adding above changes would make that work too.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -290,9 +294,104 @@ class ResizeSubgraph2 : public ExtractScalesSubgraph
...
+        int A = addNodeToMatch("Constant");
+        int div = addNodeToMatch("Div", A, sqrt);
+        int mul1 = addNodeToMatch("Mul", reshape1, div);
+        int mul2 = addNodeToMatch("Mul", reshape4, mul1);
+        int sub = addNodeToMatch("Sub", reshape2, mul2);
+        int mul3 = addNodeToMatch("Mul", input, mul1);
+        addNodeToMatch("Add", mul3, sub);
+        setFusedNode("BatchNormalization", input, data1, data2, data4 ,data3);
+    }
+};
```
dkurt, line 387: I'd like to recommend alternative solution which will work for other cases with `"Initializer"` nodes (i.e. as this resize https://github.com/daquexian/onnx-simplifier/issues/34). Before the fusion, create copies of the Initializer contents as regular Constant nodes. (i.e. as the first graph nodes).

ashishkrshrivastava: Yeah, that can also be done. I try.

ashishkrshrivastava: @dkurt I just want to make sure that you mean before simplify subgraph,
for each node: 
replace initializer input by constant node (created by copying initializer tensor data) input so that number of inputs for each node will be same as before.

dkurt: That's right! Otherwise we will have to duplicate existing subgraphs fusion to manage both Constant and Initializer while then look similar as for me.

ashishkrshrivastava: @dkurt I have made the suggested changes. Just added constant nodes with the same output id as initializers so that Graph matching can be done without matching initializers.

modules/dnn/test/test_onnx_importer.cpp
```diff
@@ -290,6 +290,15 @@ TEST_P(Test_ONNX_layers, BatchNormalization3D)
     testONNXModels("batch_norm_3d");
 }
 
+TEST_P(Test_ONNX_layers, BatchNormalizationUnfused)
+{
+    if (backend == DNN_BACKEND_INFERENCE_ENGINE_NN_BUILDER_2019)
+        applyTestTag(CV_TEST_TAG_DNN_SKIP_IE_NN_BUILDER);
+    if (backend == DNN_BACKEND_INFERENCE_ENGINE_NGRAPH)
+        applyTestTag(CV_TEST_TAG_DNN_SKIP_IE_NGRAPH);
```
dkurt, line 298: please remove skip tags if the test passes

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -1384,31 +1406,46 @@ void ONNXImporter::populateNet(Net dstNet)
...
+
+                layerParams.set("align_corners", interp_mode == "align_corners");
+                if (layerParams.get<String>("mode") == "linear")
+                {
+                    layerParams.set("mode", interp_mode == "pytorch_half_pixel" ?
+                                            "opencv_linear" : "bilinear");
+                }
+            }
             layerParams.type = "Resize";
+            //scale as attribute
```
dkurt, line 1423: I think that this comment is not necessary

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -1384,31 +1406,46 @@ void ONNXImporter::populateNet(Net dstNet)
...
             if (layerParams.has("scales"))
             {
-                // Pytorch layer
+                // Pytorch
```
dkurt, line 1426: please revert

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -1384,31 +1406,46 @@ void ONNXImporter::populateNet(Net dstNet)
...
             {
                 // Caffe2 layer
                 replaceLayerParam(layerParams, "height_scale", "zoom_factor_y");
                 replaceLayerParam(layerParams, "width_scale", "zoom_factor_x");
             }
-            replaceLayerParam(layerParams, "mode", "interpolation");
-
-            if (layerParams.get<String>("interpolation") == "linear" && framework_name == "pytorch") {
-                layerParams.type = "Resize";
+            //scale as input
```
dkurt, line 1438: redundant comment

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -300,11 +306,28 @@ void ONNXImporter::addLayer(Net& dstNet, LayerParams& layerParams,
...
+    for (int id = 0; id < num_initializers; id++)
+    {
+        opencv_onnx::TensorProto initializer = graph_proto.initializer(id);
+        opencv_onnx::NodeProto* constant_node = graph_proto.add_node();
+        constant_node->set_op_type("Constant");
+        constant_node->set_name(initializer.name());
+        constant_node->add_output(initializer.name());
+        opencv_onnx::AttributeProto* value = constant_node->add_attribute();
+        opencv_onnx::TensorProto* tensor = value->add_tensors();
+        tensor->CopyFrom(initializer);
```
dkurt, line 321: Let's also remove initialized nodes in the loop after they copied to the constant node

modules/dnn/src/onnx/onnx_graph_simplifier.hpp
```diff
@@ -27,10 +27,8 @@ void simplifySubgraphs(opencv_onnx::GraphProto& net);
 template<typename T1, typename T2>
 void convertInt64ToInt32(const T1& src, T2& dst, int size)
 {
-    for (int i = 0; i < size; i++) {
-        if (src[i] < std::numeric_limits<int32_t>::min() || src[i] > std::numeric_limits<int32_t>::max()) {
-            CV_Error(Error::StsOutOfRange, "Input is out of OpenCV 32S range");
-        }
+    for (int i = 0; i < size; i++)
+    {
```
dkurt, line 31: Is it necessary for this PR?

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -357,9 +357,40 @@ class ResizeSubgraph2 : public ExtractScalesSubgraph
...
+        int reshape1 = addNodeToMatch("Reshape", data1, shape1);
+        int shape2 = addNodeToMatch("Constant");
+        int reshape2 = addNodeToMatch("Reshape", data2, shape2);
+        int shape3 = addNodeToMatch("Constant");
+        int reshape3 = addNodeToMatch("Reshape", data3, shape3);
+        int shape4 = addNodeToMatch("Constant");
+        int reshape4 = addNodeToMatch("Reshape", data4, shape4);
+        int sqrt = addNodeToMatch("Sqrt", reshape3);
+        int A = addNodeToMatch("Constant");
+        int div = addNodeToMatch("Div", A, sqrt);
```
dkurt, line 380: please replace `int sqrt` and `int div` to something like `sqrtNode` and `divNode` to avoid conflicts with std namespace.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
dkurt: Thanks!

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -357,9 +357,40 @@ class ResizeSubgraph2 : public ExtractScalesSubgraph
...
+        int mul3 = addNodeToMatch("Mul", input, mul1);
+        addNodeToMatch("Add", mul3, sub);
+        setFusedNode("BatchNormalization", input, data1, data2, data4 ,data3);
+    }
+};
+
 void simplifySubgraphs(opencv_onnx::GraphProto& net)
 {
     std::vector<Ptr<Subgraph> > subgraphs;
+    subgraphs.push_back(makePtr<BatchNormalizationSubgraph>());
```
dkurt, line 393: Why should it go first? Let's add to the end of the list if there is no conflicts with other subgraphs.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -300,11 +306,28 @@ void ONNXImporter::addLayer(Net& dstNet, LayerParams& layerParams,
...
+    int num_initializers = graph_proto.initializer_size();
+    for (int id = 0; id < num_initializers; id++)
+    {
+        opencv_onnx::TensorProto initializer = graph_proto.initializer(id);
+        opencv_onnx::NodeProto* constant_node = graph_proto.add_node();
+        constant_node->set_op_type("Constant");
+        constant_node->set_name(initializer.name());
+        constant_node->add_output(initializer.name());
+        opencv_onnx::AttributeProto* value = constant_node->add_attribute();
+        opencv_onnx::TensorProto* tensor = value->add_tensors();
```
dkurt, line 320: Maybe let's copy it to `t` field instead of `tensors`?

modules/dnn/src/onnx/onnx_graph_simplifier.hpp
ashishkrshrivastava: yeah, tests were failing without this change.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: sure.

modules/dnn/src/onnx/onnx_importer.cpp
ashishkrshrivastava: @dkurt t I was trying to do so but couldn't find any function for that so I had to copy it to `tensor`. Maybe I missed it. I am checking it, please tell me if you have that in your mind.

modules/dnn/src/onnx/onnx_importer.cpp
ashishkrshrivastava: changes are removed. resolving conversation.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
ashishkrshrivastava: subgraph fusion for that subgraph is removed. resolving conversation.

modules/dnn/src/onnx/onnx_importer.cpp
ashishkrshrivastava: If I remove intializers inside this loop. then this function would be useless because it wont be able to fetch initializer data.
https://github.com/opencv/opencv/blob/c9c60783df11f4434d9b6afcd3ff00f5518682a7/modules/dnn/src/onnx/onnx_importer.cpp#L133-L147
which is being called here
https://github.com/opencv/opencv/blob/c9c60783df11f4434d9b6afcd3ff00f5518682a7/modules/dnn/src/onnx/onnx_importer.cpp#L330-L334

So we will have to fetch data from `constant` nodes, means we need to create a function which first go through all of the nodes and fetch data from all `constant` nodes. 
Am I right ?

modules/dnn/src/onnx/onnx_importer.cpp
dkurt: There is `set_allocated_t` in https://github.com/opencv/opencv/blob/master/modules/dnn/misc/onnx/opencv-onnx.pb.h

modules/dnn/src/onnx/onnx_importer.cpp
ashishkrshrivastava: we aren't using these line in the code now. may I resolve the conversation ?

modules/dnn/src/onnx/onnx_importer.cpp
dkurt: Let's see, it `set_allocated_t` will work - we can keep as is.

ashishkrshrivastava: @dkurt , please have a look at the changes.

ashishkrshrivastava: Setting tensor directly from `initializer` wasn't working
`value->set_allocated_t(initializer);`
It was giving me Segmentation core error so I created a new tensor and set that to value attribute. So it is copied tensor. 
```
opencv_onnx::TensorProto* tensor = initializer.New();
tensor->CopyFrom(initializer);
```

modules/dnn/src/onnx/onnx_importer.cpp
ashishkrshrivastava: Done! thanks

modules/dnn/src/onnx/onnx_importer.cpp
dkurt: Can you experiment if we get `mutable_initializer` and then set it without copying by `set_allocated_t` everything will be fine or it will crash? It depends on ownership policy in case of `set_allocated_t` - will protobuf release the object twice or not.

ashishkrshrivastava: I have already tried it. Its not working.

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -1393,22 +1423,23 @@ void ONNXImporter::populateNet(Net dstNet)
...
+            else
+            {
+                // scales as input
                 Mat scales = getBlob(node_proto, constBlobs, 1);
                 CV_Assert(scales.total() == 4);
-                layerParams.set("interpolation", "opencv_linear");
                 layerParams.set("zoom_factor_y", scales.at<float>(2));
                 layerParams.set("zoom_factor_x", scales.at<float>(3));
+                if (layerParams.get<String>("mode") == "linear" && framework_name == "pytorch")
+                    layerParams.set("mode", "opencv_linear");
```
dkurt, line 1440: I think that this logic should be independent from scaling source location. Let's move out of if-else scope

modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -1355,8 +1373,7 @@ void ONNXImporter::populateNet(Net dstNet)
                 CV_Assert(layer_id.find(node_proto.input(i)) == layer_id.end());
 
             String interp_mode = layerParams.get<String>("coordinate_transformation_mode");
-            CV_Assert_N(interp_mode != "tf_crop_and_resize", interp_mode != "asymmetric",
-                        interp_mode != "tf_half_pixel_for_nn");
+            CV_Assert_N(interp_mode != "tf_crop_and_resize", interp_mode != "tf_half_pixel_for_nn");
```
dkurt, line 1376: I think that it works only for nearest but not bilinear upsampling.

modules/dnn/src/onnx/onnx_importer.cpp
dkurt: Got it, thanks!
So can you try now right after `tensor->CopyFrom(initializer);` clear the origin tensor and remove it from the graph? For memory optimization.

modules/dnn/src/onnx/onnx_importer.cpp
ashishkrshrivastava: using mode `bilinear` creater the graph with attributes.
Screenshot from 2020-04-12 19-46-21

modules/dnn/src/onnx/onnx_importer.cpp
ashishkrshrivastava:
```
tensor->CopyFrom(initializer);
releaseONNXTensor(initializer);
value->set_allocated_t(tensor);
```

where releaseONNXTensor is
```
void releaseONNXTensor(opencv_onnx::TensorProto& tensor_proto)
{
    if (!tensor_proto.raw_data().empty()) {
        delete tensor_proto.release_raw_data();
    }
}
```
I think you meant the same.

modules/dnn/src/onnx/onnx_importer.cpp
dkurt: Thank, ok, let's keep as is.

modules/dnn/src/onnx/onnx_importer.cpp
dkurt: Something like that but we have no raw_data in this case

ashishkrshrivastava: all the initializers have data type as raw_data. I have checked data types. 
```
bias: raw data

running_mean: raw data

running_var: raw data

weight: raw data

```
Do you think we should modify this function for all data_types or a new function should be created to check and delete all data_types and let this function as it is ?

modules/dnn/test/test_onnx_importer.cpp
ashishkrshrivastava: @dkurt I have tried removing skip tags, tests are getting failed for some combinations of backend and target. So should I add skip tags for those specific combinations or it is fine the way it is right now ?

modules/dnn/src/onnx/onnx_importer.cpp
ashishkrshrivastava: Done !

modules/dnn/src/onnx/onnx_importer.cpp
dkurt: No, that's not necessary. We can just call method `clear_initializer`.

ashishkrshrivastava: i think we cannot use
`graph_proto.clear_initializer()` before calling `getGraphTensors(graph_proto)`.

dkurt:
Please take a look at merge artifacts for opencv_extra PR: https://github.com/opencv/opencv_extra/pull/734/files#r407673861.

ashishkrshrivastava:
> Please take a look at merge artifacts for opencv_extra PR: https://github.com/opencv/opencv_extra/pull/734/files#r407673861.

Done !

JulienMaille:
@ashishkrshrivastava Right now I need to export my Unet/Resnet with opset=9, other wise the Upsample block is split into many cast/concat blocks like described here (which won't load in OpenCV): https://github.com/pytorch/pytorch/issues/27376#issue-502757936
Does your PR tackles this? Isn't the problem on pytorch's side?

ashishkrshrivastava:
@julienMaille support for opset 9 and opset 11 for unfused Upsample and Resize are already added. check #16709 . But those subgraphs has scaling for both H, W. The example you shared has only scale for H only. So subgraph is different accordingly. Still in my view that subgraph will also work because in case of no fusion each node has to work properly as data flows through each node and OpenCV currently supports all those operator nodes. I would love to add subgraph fusion for such subgraphs, if needed. But for now, it will work without fusion.

PR #16895, opened 2020-03-24
core(persistence): fix resource leaks - force closing files

### LINKED ISSUE #16823: FileStorage does not handle properly an empty yml file

alalek:
resolves #16823
closes #16822

### DIFF
--- a/modules/core/test/test_io.cpp
+++ b/modules/core/test/test_io.cpp
@@ -1711,4 +1711,65 @@ TEST(Core_InputOutput, FileStorage_JSON_VeryLongLines)
     }
 }
 
+TEST(Core_InputOutput, FileStorage_empty_16823)
+{
+    std::string fname = tempfile("test_fs_empty.yml");
+    {
+        // create empty file
+        std::ofstream f(fname.c_str(), std::ios::out);
+    }
+
+    try
+    {
+        FileStorage fs(fname, FileStorage::READ);
+        ADD_FAILURE() << "Exception must be thrown for empty file.";
+    }
+    catch (const cv::Exception&)
+    {
+        // expected way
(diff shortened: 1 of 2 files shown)


modules/core/test/test_io.cpp
```diff
@@ -1711,4 +1711,34 @@ TEST(Core_InputOutput, FileStorage_JSON_VeryLongLines)
     }
 }
 
+TEST(Core_InputOutput, FileStorage_empty_16823)
+{
+    std::string fname = "test_empty.yml";
```
mshabunin:
```suggestion
    std::string fname = tempfile("test_empty.yml");
```

modules/core/test/test_io.cpp
```diff
@@ -1711,4 +1711,34 @@ TEST(Core_InputOutput, FileStorage_JSON_VeryLongLines)
...
+    {
+        // create empty file
+        std::ofstream f(fname.c_str(), std::ios::out);
+    }
+
+    try
+    {
+        FileStorage fs(fname, FileStorage::READ);
+        ADD_FAILURE() << "Failed: Exception throws on empty input. Actual it doesn't.";
```
mshabunin:
```suggestion
        ADD_FAILURE() << "Exception must be thrown for empty input.";
```

mshabunin:
Maybe it will be even better to use `EXPECT_THROW(..., cv::Exception)`?

alalek:
@chacha21 updated with handling of proposed use case

PR #16900, opened 2020-03-24
cuda4dnn(conv): autotuning for convolution

YashasSamaga:
**Statistics as of May 14th 2020:**

Devices:

- GTX 1050 Mobile
- GTX 1080 Ti
- RTX 2080 Ti

[benchmarks without this PR](https://gist.github.com/YashasSamaga/4640594be2580d85b80f5c662339ff4b)

[benchmarks with this PR](https://gist.github.com/YashasSamaga/be44243671b4796977b475d02f41ef80)

[autotuning algorithm selections](https://gist.github.com/YashasSamaga/963d628a10711c8b5bfd6d8da2a5fee0)

[convolution configurations](https://gist.github.com/YashasSamaga/b048ee7084e52d9c9a3dd41f15663e09)

---

**NOTE:** Not up-to-date.

**Device:** GTX 1050

Model                                       | without this patch | with this patch
---------------------------------------- | ----------------------- | ---------------------
MobileNet SSD Coco v1          | 55ms                    |  **7ms**
MobileNet SSD Coco v2          | 84ms                    | **11ms**
OpenPose pose MPI                | 156ms                 | **111ms**
EfficientNet B0 YOLOv3          | 121ms                  | **15ms**
FastNeuralStyle Stary Night    | 26ms                    | **22ms**
Inception v2 Mask RCNN        | 180ms                  | **152ms**

Every model which uses depthwise convolutions has got an improvement.

Most of Mask RCNN improvement comes from a single convolution layer where the algorithm chosen by heuristics took 40ms whereas the best algorithm took just 20ms. 

---

**Pending:**
- insanely big initialization times

```
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/dnn/src/cuda4dnn/csl/cudnn/convolution.hpp
+++ b/modules/dnn/src/cuda4dnn/csl/cudnn/convolution.hpp
@@ -137,6 +137,23 @@ namespace cv { namespace dnn { namespace cuda4dnn { namespace csl { namespace cu
         cudnnFilterDescriptor_t descriptor;
     };
 
+    enum class MathType {
+        DEFAULT_MATH,
+        TENSOR_OP_MATH,
+        TENSOR_OP_MATH_ALLOW_CONVERSION
+    };
+
+    namespace detail {
+        inline cudnnMathType_t get_cudnn_math_type(MathType type) {
+            switch(type) {
+                case MathType::DEFAULT_MATH: return CUDNN_DEFAULT_MATH;
+                case MathType::TENSOR_OP_MATH: return CUDNN_TENSOR_OP_MATH;
+                case MathType::TENSOR_OP_MATH_ALLOW_CONVERSION: return CUDNN_TENSOR_OP_MATH_ALLOW_CONVERSION;
+            }
+            return {};
+        }
+    };
+
     /** describes a convolution operation
      *
      * @tparam  T   type of element participating in convolution
@@ -165,9 +182,11 @@ namespace cv { namespace dnn { namespace cuda4dnn { namespace csl { namespace cu
             const SequenceContainer& zero_padding,
             const SequenceContainer& stride,
             const SequenceContainer& dilation,
-            std::size_t group_count)
+            std::size_t group_count,
+            MathType math_type
+        )
         {
-            constructor(zero_padding, stride, dilation, group_count);
+            constructor(zero_padding, stride, dilation, group_count, math_type);
         }
 
         ~ConvolutionDescriptor() noexcept {
@@ -192,7 +211,8 @@ namespace cv { namespace dnn { namespace cuda4dnn { namespace csl { namespace cu
             const SequenceContainer& zero_padding,
             const SequenceContainer& stride,
             const SequenceContainer& dilation,
-            std::size_t group_count)
+            std::size_t group_count,
+            MathType math_type)
         {
             CV_Assert(zero_padding.size() == stride.size());
             CV_Assert(zero_padding.size() == dilation.size());
@@ -225,8 +245,7 @@ namespace cv { namespace dnn { namespace cuda4dnn { namespace csl { namespace cu
                     );
                 }
                 CUDA4DNN_CHECK_CUDNN(cudnnSetConvolutionGroupCount(descriptor, group_count));
-                if (std::is_same<T, half>::value)
-                    CUDA4DNN_CHECK_CUDNN(cudnnSetConvolutionMathType(descriptor, CUDNN_TENSOR_OP_MATH));
+                CUDA4DNN_CHECK_CUDNN(cudnnSetConvolutionMathType(descriptor, detail::get_cudnn_math_type(math_type)));
             } catch (...) {
                 /* cudnnDestroyConvolutionDescriptor will not fail for a valid descriptor object */
                 CUDA4DNN_CHECK_CUDNN(cudnnDestroyConvolutionDescriptor(descriptor));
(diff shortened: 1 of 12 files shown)


tompollok:
Great Job @YashasSamaga

YashasSamaga:
The CUDA backend now performs autotuning on its own which appears have resolved the performance degradation problems. Moreover, it also includes fused convolutions while autotuning. This has improved the performance further.

### Problem:
The initialization time has gone up significantly. It's in several seconds for many models and few of them take two digit seconds. Mask RCNN takes 26s for initialization!

There are eight algorithms available for convolution. These are the combinations which are tried while autotuning:

1. **cuDNN default math fused:**
    - all eight algorithms are tried
    - cuDNN does convolution, bias addition and activation together
2. **cuDNN default math unfused:**
    - all eight algorithms are tried
    - cuDNN does the convolution
    - bias addition and activation are carried out by cuda4dnn kernels (bias and activation are fused together)
3. **cuDNN tensor core fused:**
    - two algorithms are tried
    - cuDNN does convolution, bias addition and activation together
4. **cuDNN tensor core unfused**
    - two algorithms are tried
    - cuDNN does the convolution
    - bias addition and activation are carried out by cuda4dnn kernels (bias and activation are fused together)

In the worst case, 20 combinations will be tried for every convolution layer. There are few algorithms, particularly FFT based algorithms, which request insane amounts of workspace. For example, the FFT tiling algorithm requests a workspace of 8GB for convolving on a 200KB image. These memory allocations take seconds!

The FFT algorithms are feebly used (only once in my testbed consisting of 26 models) as they are too slow for small filters and small batches.

**Why try unfused cuDNN combinations when fused cuDNN combinations are available?**
Because fused cuDNN can sometimes be slower than unfused cuDNN convolution with separate bias_activation step. For example, if fused cuDNN operations are forced whenever they are available, MobileNet to take 37ms instead of just 7ms!

tompollok:
Would it make sense to add a file next to the model file that persists the tuned configuration in case the same hardware is used to save expensive autotuning when just processing a single image on request, because sometimes several models have to be applied to a single image for some tasks.

YashasSamaga:
That would require discussions on the storage format, API and a lot of other things (and quite a bit of work).

An easier temporary solution might be to make autotuning optional (opt-in feature, i.e. disabled by default). This again will require new API (or we can use a layer param of the conv layer).

alalek:
Some similar approach exists in OpenCL (ocl4dnn) backend. It includes:
- rules to define convolution [configuration](https://github.com/opencv/opencv/blob/4.2.0/modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp#L631-L643).
- [default "pre-tuned" values](https://github.com/opencv/opencv/blob/4.2.0/modules/dnn/src/ocl4dnn/include/default_kernel_config.hpp) for some cases (target H/W is selected through number of execution units (EU) - autotuning is available for Intel iGPUs).
- multi-level storage of tuned configurations: ["default"](https://github.com/opencv/opencv/blob/4.2.0/modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp#L1973-L1974), ["in-memory"](https://github.com/opencv/opencv/blob/4.2.0/modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp#L1976), ["on disk"](https://github.com/opencv/opencv/blob/4.2.0/modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp#L2045).
- tuning configuration [flags](https://github.com/opencv/opencv/blob/4.2.0/modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp#L204-L233). Auto-tuning is disabled by default, but can be enabled by request or loaded from the disk.
- some [magic](https://github.com/opencv/opencv/blob/4.2.0/modules/dnn/src/ocl4dnn/src/ocl4dnn_conv_spatial.cpp#L1665-L1695) how candidates are generated for tuning process.

YashasSamaga:
What about having an API for loading, saving and optimizing model cofigurations that is exposed through `cv::dnn::Net`? This seems more user-friendly than using environment variables. It also allows having unoptimized and optimized models to be used simultaneously (without having to modify the environment variables at runtime). It's also easier to extend in the future.

Both OpenCL and CUDA could use the same API instead of having individual backend-specific APIs.

I think only three functions are required (and maybe user-friendly overloads):
```
void Net::loadOptimizedConfiguration(const std::string& config);
void Net::saveOptimizedConfiguration(std::string& config);
void Net::runOptimizer();
```

The configuration could be stored in protobuf format. The format can be kept as an internal format.

tompollok:
The API proposal sounds reasonable to me as it makes the use of optimized vs non optimized versions dynamically selectable at runtime and transparent to the users. I would also vote against the use environment variables.

YashasSamaga:
cuDNN 8 has a brand new API which offers better autotuning capabilities. I will make a PR supporting cuDNN 8's new backend API and then a PR for autotuning.

lorenzolightsgdwarf:
@YashasSamaga thanks for the work! On my net (a custom version of SSD MobileNet v2) I haven't seen any improvement. I'm using Cuda 10.2 and Cudnn 7.6.2. Is there any flag/environment variable that enables the autotuning? Or is there something that disables it? Bests!

YashasSamaga:
> @YashasSamaga thanks for the work! On my net (a custom version of SSD MobileNet v2) I haven't seen any improvement. I'm using Cuda 10.2 and Cudnn 7.6.2. Is there any flag/environment variable that enables the autotuning? Or is there something that disables it? Bests!

It's always enabled in this PR. This PR hasn't been merged into master. Therefore, you need to build this PR to use the autotuning facility.

Related: https://github.com/opencv/opencv/issues/20966

PR #16906, opened 2020-03-25
Remove NN Builder API

l-bat:

```

Xtest_modules:Custom=dnn,python2,python3,java

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```


dkurt:
@l-bat, I think that we also should remove these releases support, right?
```
#define INF_ENGINE_RELEASE_2018R5 2018050000
#define INF_ENGINE_RELEASE_2019R1 2019010000
#define INF_ENGINE_RELEASE_2019R2 2019020000
#define INF_ENGINE_RELEASE_2019R3 2019030000
```

modules/dnn/src/layers/fully_connected_layer.cpp
```diff
@@ -43,7 +43,8 @@
 #include "../precomp.hpp"
 #include "layers_common.hpp"
 #include "../op_halide.hpp"
-#include "../op_inf_engine.hpp"
+// #include "../op_inf_engine.hpp"
```
dkurt: Please remove commented code

modules/dnn/test/test_backends.cpp
```diff
@@ -78,17 +78,6 @@ class DNNTestNetwork : public DNNTestLayer
     {
         if (outputLayer == "detection_out")
         {
-            if (backend == DNN_BACKEND_INFERENCE_ENGINE_NN_BUILDER_2019)
```
dkurt: Can you please check why nGraph path works correctly this way? I thought outputs from both APIs should be the same.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -284,9 +279,7 @@ CV__DNN_EXPERIMENTAL_NS_BEGIN
          */
         virtual Ptr<BackendNode> initHalide(const std::vector<Ptr<BackendWrapper> > &inputs);
 
-        virtual Ptr<BackendNode> initInfEngine(const std::vector<Ptr<BackendWrapper> > &inputs);
-
-        virtual Ptr<BackendNode> initNgraph(const std::vector<Ptr<BackendWrapper> > &inputs, const std::vector<Ptr<BackendNode> >& nodes);
+        virtual Ptr<BackendNode> initNgraph(const std::vector<Ptr<BackendNode> >& nodes);
```
dkurt: What do you think, it's better to rename it to initial `initInfEngine`? But let's discuss before change :)

modules/dnn/src/dnn.cpp
```diff
@@ -588,8 +535,7 @@ struct DataLayer : public Layer
 
     virtual bool supportBackend(int backendId) CV_OVERRIDE
     {
-        return backendId == DNN_BACKEND_OPENCV ||
-               (backendId == DNN_BACKEND_INFERENCE_ENGINE_NN_BUILDER_2019 && inputsData.size() == 1);
+        return backendId == DNN_BACKEND_OPENCV;
```
dkurt: We need to support this layer with nGraph as for me

modules/dnn/src/dnn.cpp
```diff
@@ -1471,20 +1366,12 @@ struct Net::Impl
...
-#else
-            CV_Assert(false && "This OpenCV version is built without Inference Engine NN Builder API support");
-#endif
-        }
-        else if (preferableBackend == DNN_BACKEND_INFERENCE_ENGINE_NGRAPH)
+        else if (preferableBackend == DNN_BACKEND_INFERENCE_ENGINE)
         {
-#ifdef HAVE_DNN_NGRAPH
+#ifdef HAVE_INF_ENGINE
             initNgraphBackend(blobsToKeep_);
```
dkurt: Maybe let's rename `initNgraphBackend` to `initInfEngineBackend`

modules/dnn/src/dnn.cpp
l-bat: Ddo we need `inputData.size () == 1`condition?

modules/dnn/test/test_backends.cpp
l-bat: It strange, but nGraph works in both cases

modules/dnn/src/dnn.cpp
dkurt: At least match DNN Builder implementation

asmorkalov:
@l-bat @dkurt Do you plan to work on the patch in mean time? There are merge conflicts.

alalek:
Lets hold this activity for now.

JulienMaille:
@alalek is there a way to compile current `master` with legacy IE disabled (and avoid linking with `inference_engine_legacy.dll`)?
Thanks

PR #16911, opened 2020-03-26
G-API utils - make variant converting constructor and assignment operator properly forward it's argument

anton-potapov:
This patch fixes usage of [Universal (Forwarding ) reference](https://en.cppreference.com/w/cpp/language/reference) in the converting constructor and assignment operator of `variant`.
 
<cut/>

In c++,  function template argument such  as `T&&`  is not always a _rvalue  reference_. If `T` is function template parameter, then `T&&` is a universal reference , which can be bound to both _rvalue  reference_ and _lvalue  reference_  . I.e. in the following declaration  `t` is the universal reference: 
```c++
template <typename T> 
voif f(T&& t);
```
So, in order to have a template function with templated parameter accepting only _rvalue references_, some extra effort is needed . One possible way is to add an extra overload which is a better match for _lvalue references_ .  For `variant` converting constructor it is:
```c++
template<typename T> explicit variant::variant(T& t);
```

```
force_builders=ARMv7,Custom
build_image:Custom=centos:7
buildworker:Custom=linux-1

build_gapi_
```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/util/variant.hpp
+++ b/modules/gapi/include/opencv2/gapi/util/variant.hpp
@@ -36,11 +36,14 @@ namespace util
             static const constexpr std::size_t value = I;
         };
 
+        template< bool B, class T = void >
+        using enable_if_t = typename std::enable_if<B,T>::type;
 
-        template<class T, class U, class V> using are_different =
-            std::enable_if<!std::is_same<typename std::decay<T>::type,
-                                         typename std::decay<U>::type>::value,
-                           V>;
+        template<class T, class U, class V = void>
+        using are_different_t = enable_if_t<
+                !std::is_same<typename std::decay<T>::type,
+                              typename std::decay<U>::type>::value,
+                 V>;
     }
 
     template<typename Target, typename... Types>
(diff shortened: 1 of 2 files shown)


alalek:
@anton-potapov This patch breaks builds on old compilers (like GCC 4.8.5 from CentOS 7). Please take a look.

dmatveev:
@alalek @anton-potapov 

Can the failure we see there in the very first GPU kernel be related to this CMake option I see in the [cmake log](https://pullrequest.opencv.org/buildbot/builders/precommit_custom_linux/builds/3602/steps/cmake/logs/stdio): `-DWITH_OPENCL=OFF`

Our tests should for OpenCL (GPU) should be skipped in this case, for some reason it didn't happen

anton-potapov:
@dmatveev @alalek , it seems that this is broken for a pretty long time. 

I took 5d01821d9f4d57921dad82045cc9819cfabcfe15 and see the issue is reproduced.  So this failures are not related to the patch. ( I will create a ticket for this).

alalek:
> this failures are not related to the patch

Right. See nightly [master](http://pullrequest.opencv.org/buildbot/grid?tag=scheduled-master&width=7) builds and click on "[centos](http://pullrequest.opencv.org/buildbot/builders/master_etc-centos-lin64)" builder.

dmatveev:
Nice done, thanks! 👍

dmatveev:
@anton-potapov can you please check if this issue also apply to our implementation of `optional`?

PR #6801, opened 2016-07-08
Support for super-fisheye lenses (i.e., opening angles above 180 degr…

vpisarev:
This PR adds support for super-fisheye lenses (i.e., lenses with opening angles above 180 degrees).

Essentially, the idea is to compute `theta` as `atan2(sqrt(x*x+y*y),z)` (where `[x,y,z]` is the point relative to the camera, with z pointing forward).

It is downward-compatible for points in front of the camera. Points behind the camera (i.e., negative z coordinates) gave bogus results previously and now only fail if they are exactly behind the camera.
The computation of the Jacobians should be slightly faster and more accurate, especially for z close to zero.

Other functions in the fisheye module would need to be ported as well and the documentation would also need to be updated. But before I proceed, are there any opinions on this?

PR #16926, opened 2020-03-28
Fisheye model extended with additional plane to support FOV > 90 in projectPoints

dmitry42nd:

- [y] I agree to contribute to the project under OpenCV (BSD) License.
- [y] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
- [y] The PR is proposed to proper branch
- [y] There is reference to original bug report and related work 
**see: http://www.ee.oulu.fi/~jkannala/publications/tpami2006.pdf**
- [y] There is accuracy test, performance test and test data in opencv_extra repository, if applicable. Patch to opencv_extra has the same branch name.
**opencv_extra PR: https://github.com/opencv/opencv_extra/pull/747**
- [n] The feature is well documented and sample code can be built with the project CMake

Context: OpenCV 3.4.3-4.2.0, Ubuntu 18.04, Cmake ~3.10
Documentation in progress

---
-According paper theta_d is actually tan_theta_d and that restrictive condition had no sense in undistortPoints();
-model extended on additional plane to support FOV > 90 in projectPoints()

### DIFF
--- a/modules/calib3d/src/fisheye.cpp
+++ b/modules/calib3d/src/fisheye.cpp
@@ -416,6 +416,8 @@ void cv::fisheye::undistortPoints( InputArray distorted, OutputArray undistorted
                     break;
             }
 
+            theta = min(theta, CV_PI/2.);
+                
             scale = std::tan(theta) / theta_d;
         }
 
--- a/modules/calib3d/src/fisheye.cpp
+++ b/modules/calib3d/src/fisheye.cpp
@@ -127,15 +127,27 @@ void cv::fisheye::projectPoints(InputArray objectPoints, OutputArray imagePoints
     {
         Vec3d Xi = objectPoints.depth() == CV_32F ? (Vec3d)Xf[i] : Xd[i];
         Vec3d Y = aff*Xi;
-        if (fabs(Y[2]) < DBL_MIN)
-            Y[2] = 1;
-        Vec2d x(Y[0]/Y[2], Y[1]/Y[2]);
+
+        if(fabs(Y[2]) < 1e-8)
+        {
+            if(Y[2] > 0)
+                Y[2] = 1e-8;
+            else
+                Y[2] = -1e-8;
+        }
+
+        Vec2d x(Y[0]/fabs(Y[2]), Y[1]/fabs(Y[2]));
 
         double r2 = x.dot(x);
         double r = std::sqrt(r2);
 
         // Angle of the incoming ray:
-        double theta = atan(r);
+        double theta;
+
+        if(Y[2] > 0)
+            theta = atan2(r, 1);
+        else
+            theta = atan2(1, r) + M_PI/2;
 
         double theta2 = theta*theta, theta3 = theta2*theta, theta4 = theta2*theta2, theta5 = theta4*theta,
                 theta6 = theta3*theta3, theta7 = theta6*theta, theta8 = theta4*theta4, theta9 = theta8*theta;
@@ -170,12 +182,12 @@ void cv::fisheye::projectPoints(InputArray objectPoints, OutputArray imagePoints
 
             //Vec2d x(Y[0]/Y[2], Y[1]/Y[2]);
             Vec3d dxdom[2];
-            dxdom[0] = (1.0/Y[2]) * dYdom[0] - x[0]/Y[2] * dYdom[2];
-            dxdom[1] = (1.0/Y[2]) * dYdom[1] - x[1]/Y[2] * dYdom[2];
+            dxdom[0] = (1.0/fabs(Y[2])) * dYdom[0] - x[0]/fabs(Y[2]) * dYdom[2];
+            dxdom[1] = (1.0/fabs(Y[2])) * dYdom[1] - x[1]/fabs(Y[2]) * dYdom[2];
 
             Vec3d dxdT[2];
-            dxdT[0]  = (1.0/Y[2]) * dYdT[0] - x[0]/Y[2] * dYdT[2];
-            dxdT[1]  = (1.0/Y[2]) * dYdT[1] - x[1]/Y[2] * dYdT[2];
+            dxdT[0]  = (1.0/fabs(Y[2])) * dYdT[0] - x[0]/fabs(Y[2]) * dYdT[2];
+            dxdT[1]  = (1.0/fabs(Y[2])) * dYdT[1] - x[1]/fabs(Y[2]) * dYdT[2];
 
             //double r2 = x.dot(x);
             Vec3d dr2dom = 2 * x[0] * dxdom[0] + 2 * x[1] * dxdom[1];
@@ -188,7 +200,12 @@ void cv::fisheye::projectPoints(InputArray objectPoints, OutputArray imagePoints
 
             // Angle of the incoming ray:
             //double theta = atan(r);
-            double dthetadr = 1.0/(1+r2);
+            double dthetadr;
+            if(Y[2] > 0)
+                dthetadr = 1/(1+r2);
+            else
+                dthetadr = -1/(1+r2);
+
             Vec3d dthetadom = dthetadr * drdom;
             Vec3d dthetadT  = dthetadr *  drdT;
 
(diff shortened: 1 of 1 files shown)


asmorkalov:
@dmitry42nd Do you have a chance to finish the patch? Please pay attention to test failures on CI. Could you also provide reproducer/test for the fixed issue.

asmorkalov:
@dmitry42nd friendly reminder.

sergregory:
@dmitry42nd could you please provide a test set or a synthetic dataset showing the feasibility of the implemented approach? The main question with regard to the super-large FOVs (>180 degrees) that I have whether we able to calibrate the camera under these assumptions or not? I mean, from the top of my head, it seems that with most of the projection models we will have rays coming from the directions that are symmetric around the +/- 90 degrees mapped to the same point on the image. 

Looking at the code, I have the following concerns:
- Limiting `Y[2]` to the range `[-1e-8; \infty]` seems strange - if we should work with the FOV > 180 degrees, `Y[2]` can be significantly less than `-1e-8`
- I'm not sure if the rest of the algorithm can work with the FOV > 180 degrees. As I understand it, it relays on the `r = f \threta` projection model which is not applicable for this case.

modules/calib3d/src/fisheye.cpp
```diff
@@ -127,15 +127,27 @@ void cv::fisheye::projectPoints(InputArray objectPoints, OutputArray imagePoints
...
 
         double r2 = x.dot(x);
         double r = std::sqrt(r2);
 
         // Angle of the incoming ray:
-        double theta = atan(r);
+        double theta;
+
+        if(Y[2] > 0)
+            theta = atan2(r, 1);
```
paroj, line 148: this seems wrong to me. See https://april.eecs.umich.edu/media/pdfs/richardson2013iros.pdf Section II and the previous attempt on this https://github.com/opencv/opencv/pull/6801

dmitry42nd:
@sergregory 
Some of present-day fisheye cameras has horizontal FOV > 180 degrees (195-200 degrees) and project it on single plane with fisheye distortion. Take a look here: http://www.ee.oulu.fi/~jkannala/publications/tpami2006.pdf, beginning of the part IV

>     * Limiting `Y[2]` to the range `[-1e-8; \infty]` seems strange - if we should work with the FOV > 180 degrees, `Y[2]` can be significantly less than `-1e-8`

Its attempt to avoid devision by zero. Shall I replace '-1e-8' with -DBL_MIN?(-std::numeric_limits<double>::epsilon() would be perfect, but not sure if it breaks code style)

>     * I'm not sure if the rest of the algorithm can work with the FOV > 180 degrees. As I understand it, it relays on the `r = f \theta` projection model which is not applicable for this case.

 `r = f \theta_d`, actually. Theta distorted (by fisheye lense) is always < 180 degrees, it's true.

modules/calib3d/src/fisheye.cpp
dmitry42nd: The main concern i see here is how mach atan2() precise? Is it works well with `DBL_MIN` or `1e-8`

dmitry42nd:
@asmorkalov 
Modifications do not touch OpenCL parts of  OpenCV. 
I'm creating pull request to opencv_extra with dataset and code example right now.

modules/calib3d/src/fisheye.cpp
```diff
@@ -127,15 +127,27 @@ void cv::fisheye::projectPoints(InputArray objectPoints, OutputArray imagePoints
     {
         Vec3d Xi = objectPoints.depth() == CV_32F ? (Vec3d)Xf[i] : Xd[i];
         Vec3d Y = aff*Xi;
-        if (fabs(Y[2]) < DBL_MIN)
-            Y[2] = 1;
-        Vec2d x(Y[0]/Y[2], Y[1]/Y[2]);
+
+        if(fabs(Y[2]) < 1e-8)
```
dmitry42nd, line 131: @sergregory check if Y[2] close to zero (When theta_d is close to 90 degrees +-1e-8)

dmitry42nd:
### cv::fisheye::undistortPoints()

#### Abbreviations and terms
 - `CS` - coordinate system
 - `theta_d` - distorted theta, angle between optical axis and incoming distorted by lens light ray (restricted with PI/2)
 - `theta` - undistorted theta, angle  between optical axis and incoming  light ray **before** lens distortion (may be greater than PI/2)
#### What actually theta_d is.

                          |
                          +  <- pw
                      .   |
                  .       |
    optical -> +----------+  <- projective plane origin
    center          1     |
                          |
                          |
                          |

    The optical center and fisheye image projective plane (aka camera sensor)

Let's take a look on `pw` (see code diff `line 395`):
    
    Vec2d pw((pi[0] - c[0])/f[0], (pi[1] - c[1])/f[1]);      // world point

More precisely, it is the point coordinates on the projective plane, which are measured in focals (**1** unit = **1** focal length). Distance between `projective plane origin` and  `opical center` is **1** focal length. So (`line 399`):

    double theta_d = sqrt(pw[0]*pw[0] + pw[1]*pw[1])

is actually `distance` between `projective plane origin` to `pw`. From another side `distance` / **1**  equal to tangent of theta distorted, (**1** is focal length). So, correct naming here shall be:

    double tan_theta_d = sqrt(pw[0]*pw[0] + pw[1]*pw[1])

Next. This restriction (`line 387 in origin`):

    theta_d = min(max(-CV_PI/2., theta_d), CV_PI/2.);

causes incorrect undistortion for points, which **tangent** of `theta_d` is greater than PI/2. 
It makes no sense. First, because tan units are compared with radians. Second, because `theta_d` itself will be always less than PI/2. **Tangent** of `theta_d` may be legally equal to 1.685, (for example, when angle is 89 deg), but it is greater than `PI/2` = 1.570...! 
So this restriction may be relaxed.

More correct way would be to check if **undistorted** `theta` itself is greater than PI/2.

Currently (in both origin and fix), not all fisheye image points may be correctly undistorted with cv::fisheye::undistortPoints(), without changing API. 
Points can not be projected on 2D plane here, when `theta` is greater than PI/2, because it can not be described correctly in 2D Euclidean space. Possible solution is to add additional dimension and return 3D points (x, y, 1) in case of theta <PI/2 and (x, y, -1) in case of >PI/2 (still uncertainty in `theta`=PI/2). Another possible solution is spherical coordinates.

**But purpose of this PR** is to extend correct behavior only for points, which `theta` is less than PI/2. 

### cv::fisheye::projectPoints()

Some 3D real world points may be projected into "`theta` > PI/2" fisheye image zone. In this case Z coodrinate of the world point `Y[2]` is less than 0 (in camera CS), because it is placed behind the projective plane (beautiful property of >180 FOV fisheye cameras).
Here, using the same ideas, sign of Z coordinate of 3D point (`Y[2]`) is handled to compute `theta` angle properly, which allows to compute distorted `theta_d` later (fix do not touch this part) and project points on fisheye image correctly, when proper camera intrinsics and 3D points positions are known.

Example of code with provided fisheye images are presented here: 
* https://github.com/opencv/opencv_extra/pull/747
* https://github.com/CogentEmbedded/opencv_extra/tree/fisheye-over-90-degree-dsoldatov/testdata/calib3d

asmorkalov:
@vpisarev Could you handle it?

vpisarev:
@dmitry42nd, thank you for the contribution! Unfortunately, we have no extensive regression tests for the fisheye camera calibration, but it's clear that your patch may affect the correctness of existing functions.

We discussed it with the core team.

In order to accept the patch, we would like you to make the following changes:
1. make a separate namespace cv::super_fisheye, where you need to copy all the functions that are altered by your patch, and put the changes there instead of the original functions. This will also give us more flexibility if/when we decide to make some further modifications (like the one you mentioned in the "possible improvement" part of your comment).
2. provide at least one regression test for each newly added function. Preferably, a real dataset should be used, but you can also use a synthetic one. If you generate it on fly, make sure that the generated data is always the same (use cv::RNG with fixed seed).

vpisarev:
@dmitry42nd, let's close it for now. When the revised patch with requested changes is ready, please, submit the new pull request.

PR #16961, opened 2020-04-02
QRDetectMulti : different return value bug fix

rayonnant14:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/746

Fixed bug with skip some localization points. Due to this bug some QR-Codes couldn't be found and also the method detectMulti() could return a different number of found QR-codes 

### DIFF
--- a/modules/objdetect/test/test_qrcode.cpp
+++ b/modules/objdetect/test/test_qrcode.cpp
@@ -23,7 +23,8 @@ std::string qrcode_images_monitor[] = {
 };
 std::string qrcode_images_multiple[] = {
   "2_qrcodes.png", "3_close_qrcodes.png", "3_qrcodes.png", "4_qrcodes.png",
-   "5_qrcodes.png", "6_qrcodes.png", "7_qrcodes.png", "8_close_qrcodes.png"
+  "5_qrcodes.png", "6_qrcodes.png", "7_qrcodes.png", "8_close_qrcodes.png",
+  "14_qrcodes.png"
 };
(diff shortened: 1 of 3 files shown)


asmorkalov:
@rayonnant14 Please pay attention to failed test Perf_Objdetect_QRCode_Multi.detectMulti/7 on CI. Also could you add test case that failed before the patch, but works properly with patch to reproduce regression and test the fix.

asmorkalov:
@rayonnant14 Friendly reminder.

alalek:
6+Mb image in lossless .png format is too large. Please try to use JPEG format instead (algorithm should be stable enough to small noise or compression artifacts).
Size should not be more than 300-500Kb per mega pixel.

modules/objdetect/test/test_qrcode.cpp
```diff
@@ -501,6 +500,23 @@ TEST(Objdetect_QRCode_decodeMulti, decode_regression_16491)
...
+    std::string image_path = findDataFile(root + name_current_image);
+    Mat src = imread(image_path);
+    ASSERT_FALSE(src.empty()) << "Can't read image: " << image_path;
+
+    QRCodeDetector qrcode;
+    std::vector<Point> corners;
+    EXPECT_TRUE(qrcode.detectMulti(src, corners));
+    ASSERT_FALSE(corners.empty());
+    size_t expect_corners_size = 36;
+    EXPECT_EQ(corners.size(), expect_corners_size);
```
alalek, line 517: GoogleTests's `_EQ` args order is `(expected, actual)`.

PR #16963, opened 2020-04-02
Added GFrame class

rgarnov:
## Summary

* Added a GFrame class to introduce an api entity for image buffer with some auxiliary information

```
```

### DIFF
--- a/modules/gapi/test/gapi_frame_tests.cpp
+++ b/modules/gapi/test/gapi_frame_tests.cpp
@@ -1,0 +1,61 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+#include "test_precomp.hpp"
+
+#include <opencv2/gapi/cpu/gcpukernel.hpp>
+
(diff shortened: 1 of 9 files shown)


modules/gapi/test/gapi_frame_tests.cpp
```diff
@@ -0,0 +1,61 @@
...
+{
+    cv::GFrame in;
+    auto out = GBlurFrame::on(in);
+    cv::GComputation c(cv::GIn(in), cv::GOut(out));
+
+    auto pkg = cv::gapi::kernels<OCVBlurFrame>();
+    c.apply(cv::gin(in_mat), cv::gout(out_mat), cv::compile_args(pkg));
+
+    check();
+}
```
anton-potapov, line 59: IMHO this only test case does not show why ever `GFrame` abstraction was invented and how it differs from `GMat`

dmatveev: This is a long-long story...

dmatveev: But the idea is to distinguish the data coming from media (e.g. sensors or decoders) from an arbitrary image data -- and make it at a type level. No more details can be exposed here, unfortunately

rgarnov:

PR #16965, opened 2020-04-02
Introduce a new abstraction layer over island input/output data

rgarnov:
- Port GExecutor to this new interface

```
```

### DIFF
--- a/modules/gapi/src/backends/fluid/gfluidbackend.hpp
+++ b/modules/gapi/src/backends/fluid/gfluidbackend.hpp
@@ -153,6 +153,8 @@ public:
     virtual void run(std::vector<InObj>  &&input_objs,
                      std::vector<OutObj> &&output_objs) override;
 
+    using GIslandExecutable::run; // (IInput&, IOutput&) version
+
     void run(std::vector<InObj>  &input_objs,
              std::vector<OutObj> &output_objs);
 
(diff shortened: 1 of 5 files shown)


dmatveev:
I believe it really waits for other MRs to be merged first (e.g. on fixing the Opaque support).

...Or not?

modules/gapi/src/executor/gexecutor.cpp
```diff
@@ -117,11 +117,36 @@ void cv::gimpl::GExecutor::initResource(const ade::NodeHandle &orig_nh)
         // Constructed on Reset, do nothing here
         break;
 
+    // FIXME: What about GOPAQUE here??
```
dmatveev, line 120: ? :)

rgarnov: Added in https://github.com/opencv/opencv/pull/16964, removed this FIXME

modules/gapi/src/backends/fluid/gfluidbackend.hpp
```diff
@@ -153,6 +153,8 @@ class GFluidExecutable final: public GIslandExecutable
     virtual void run(std::vector<InObj>  &&input_objs,
                      std::vector<OutObj> &&output_objs) override;
 
+    using GIslandExecutable::run; // (IInput&, IOutput&) version
```
dmatveev, line 156: Is this required?

rgarnov: Yes, there were warnings on android and mac that another (non virtial) run below hides this method

rgarnov:

modules/gapi/src/backends/fluid/gfluidbackend.hpp
dmatveev: ok I see now. Thanks!

PR #16980, opened 2020-04-04
Incorrect PREDEFINED setting.

albert-github:
The PREDEFINED setting for  had a space between the define name and the equal sign and this is not allowed, it results in the warning:
```
error: Illegal PREDEFINED format '=', no define name specified
```
according to the documentation explicitly states that no space is allowed:
> The PREDEFINED tag can be used to specify one or more macro names that are defined before the preprocessor is started (similar to the -D option of e.g. gcc). The argument of the tag is a list of macros of the form: name or name=definition (no spaces). If the definition and the "=" are omitted, "=1" is assumed. To prevent a macro definition from being undefined via #undef or recursively expanded use the := operator instead of the = operator.

### DIFF
--- a/doc/Doxyfile.in
+++ b/doc/Doxyfile.in
@@ -241,7 +241,7 @@ PREDEFINED             = __cplusplus=1 \
                          CV_WRAP= \
                          CV_WRAP_AS(x)= \
                          CV_CDECL= \
-                         CV_Func = \
+                         CV_Func= \
                          CV_DO_PRAGMA(x)= \
                          CV_SUPPRESS_DEPRECATED_START= \
                          CV_SUPPRESS_DEPRECATED_END= \


alalek:
Thank you for contribution!

So, please:

albert-github:
After a lot of searching I hope that the new branch is now correct, otherwise I'm sorry but I have no idea how this would work. I just use a few basic git commands (and like to keep it like it is).
I have my personal  opinion about the friendliness etc. of git (won't share it here).

When it is not in the correct branch and you want this patch:
- please apply the fix yourself
  **or**
- provide an exact way of all the commands to be applied

alalek:
Pushed rebased commit.

albert-github:
Actually the credits should go to https://fossies.org where a brute force doxygen run is run over all the source code of a big number of projects.
From https://fossies.org I can see also some extra log files (there is NO intention to share these files) and in these I can see some problems (and try to solve them either in doxygen or report them to the projects).

PR #17006, opened 2020-04-07
Remove OpenCV comment in .ppm writer

tpoisonooo:
remove comment in .ppm which is **NOT** compatible with old OpenCV version.
we have to rewrite .ppm parser in C/C++ code for this comment.


asmorkalov:
@tpoisonooo @alalek I rebased the patch to 3.4 and it's ready to be merged after CI run.

3rdparty/openexr/IlmImf/ImfAttribute.cpp
```diff
@@ -63,7 +63,7 @@ Attribute::~Attribute () {}
 
 namespace {
 
-struct NameCompare: std::binary_function <const char *, const char *, bool>
+struct NameCompare
```
alalek, line 66: @asmorkalov Wrong commit is pushed

asmorkalov: My fault. I restored original commit, thanks to Github, it stores rewritten commits too.

PR #17009, opened 2020-04-07
videoio: added plug-in support to MSMF backend

mshabunin:
* enabled by cmake option `VIDEOIO_PLUGIN_LIST=msmf`
* ~~`OPENCV_VIDEOIO_PLUGIN_PATH` environment variable should be set correctly~~

<cut/>

```
force_builders=Custom Win

allow_multiple_commits=1
```


modules/videoio/test/test_video_io.cpp
```diff
@@ -323,10 +323,7 @@ static const VideoCaptureAPIs backend_params[] = {
    CAP_AVFOUNDATION,
 #endif
 
-#ifdef HAVE_MSMF
```
alalek, line 326: Should we keep `#ifdef _WIN32` here?

alalek:
"Custom Win" builder with enabled MSMF plugin:

1.

> /OUT:"C:\build\precommit_custom_windows\build\lib\Release\opencv_videoio_msmf430_64.dll"

Need to change/add `RUNTIME_OUTPUT_DIRECTORY` target property:

```.diff
-build\lib\Release\
+build\bin\Release\
```

---

2. build warning:

> modules\videoio\src\cap_msmf.cpp(1783,105): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data

alalek:
The first problem is still here:

> /OUT:"C:\build\precommit_custom_windows\build\lib\Release\opencv_videoio_msmf430_64.dll"

PR #17023, opened 2020-04-09
Remove std::binary_function as it's deprecated in C++11

poelmanc:
std::binary_function was deprecated with C++11 and removed in C++17. It provided just two typedefs which in this case were unused.


asmorkalov:
@poelmanc @alalek I rebased the patch to 3.4 and it's ready to be merged after CI run.

PR #17028, opened 2020-04-10
Update OpenCVFindVA_INTEL.cmake

xipingyan:
When set env VA_INTEL_IOCL_ROOT, "if($ENV{VA_INTEL_IOCL_ROOT})" don't work.
    My modification as follow.
    
    -    if($ENV{VA_INTEL_IOCL_ROOT})
    +    if(DEFINED ENV{VA_INTEL_IOCL_ROOT})
    
    Refer: https://cmake.org/cmake/help/latest/variable/ENV.html

### DIFF
--- a/cmake/OpenCVFindVA_INTEL.cmake
+++ b/cmake/OpenCVFindVA_INTEL.cmake
@@ -5,10 +5,9 @@
 # VA_INTEL_IOCL_ROOT - root of Intel OCL installation
 
 if(UNIX AND NOT ANDROID)
-    if(DEFINED ENV{VA_INTEL_IOCL_ROOT})
-        set(VA_INTEL_IOCL_ROOT $ENV{VA_INTEL_IOCL_ROOT})
-    else()
-        set(VA_INTEL_IOCL_ROOT "/opt/intel/opencl")
+    ocv_check_environment_variables(VA_INTEL_IOCL_ROOT)
+    if(NOT DEFINED VA_INTEL_IOCL_ROOT)
+      set(VA_INTEL_IOCL_ROOT "/opt/intel/opencl")
     endif()
 
     find_path(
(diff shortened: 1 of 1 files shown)


alalek:
Which case doesn't work?

xipingyan:
I need use intel_va. So after I installed intel_sdk_for_opencl_applications_2020.0.270.tar.gz, it's default install path is not "/opt/intel/opencl".  So when I exported "VA_INTEL_IOCL_ROOT=/opt/intel/system_studio_2020/opencl/SDK", and then excute cmake,  always tip: 

CMake Warning at cmake/OpenCVFindVA_INTEL.cmake:29 (message):
  Intel OpenCL installation is not found.

After debug, I found this line "if($ENV{VA_INTEL_IOCL_ROOT})" don't work, so I modify it based on cmake doc:https://cmake.org/cmake/help/latest/variable/ENV.html

cmake/OpenCVFindVA_INTEL.cmake
```diff
@@ -5,7 +5,7 @@
 # VA_INTEL_IOCL_ROOT - root of Intel OCL installation
 
 if(UNIX AND NOT ANDROID)
-    if($ENV{VA_INTEL_IOCL_ROOT})
+    if(DEFINED ENV{VA_INTEL_IOCL_ROOT})
         set(VA_INTEL_IOCL_ROOT $ENV{VA_INTEL_IOCL_ROOT})
     else()
         set(VA_INTEL_IOCL_ROOT "/opt/intel/opencl")
```
alalek, line 11: Also this code breaks priority of sources for variable values:
- cmake default ("/opt/intel/opencl" - lowest)
- environment
- cmake variable in command line (high priority)

So, please use this code snippet:
```
    ocv_check_environment_variables(VA_INTEL_IOCL_ROOT)
    if(NOT DEFINED VA_INTEL_IOCL_ROOT)
      set(VA_INTEL_IOCL_ROOT "/opt/intel/opencl")
    endif()
```

xipingyan:
Please help me check whether if I had finished your request.

xipingyan:
Great, I am very glad to contribute OpenCV. Thanks.

PR #17029, opened 2020-04-10
Switch to new nGraph API

l-bat:

```
Xbuild_image:Custom=ubuntu-openvino-2019r3.0:16.04

Xtest_modules:Custom=dnn,python2,python3,java

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/ie_ngraph.hpp
+++ b/modules/dnn/src/ie_ngraph.hpp
@@ -91,16 +91,18 @@ class InfEngineNgraphNode : public BackendNode
 {
 public:
     InfEngineNgraphNode(const std::vector<Ptr<BackendNode> >& nodes, Ptr<Layer>& layer,
-                        std::vector<Mat*>& inputs, std::vector<Mat>& outputs,
+                        std::vector<Mat*>& inputs, std::vector<Mat>& outMats,
                         std::vector<Mat>& internals);
 
     InfEngineNgraphNode(std::shared_ptr<ngraph::Node>&& _node);
     InfEngineNgraphNode(std::shared_ptr<ngraph::Node>& _node);
 
     void setName(const std::string& name);
+    ngraph::Output<ngraph::Node> GetOidOutput() const;
 
     // Inference Engine network object that allows to obtain the outputs of this layer.
     std::shared_ptr<ngraph::Node> node;
+    int oid;
     Ptr<InfEngineNgraphNet> net;
     Ptr<dnn::Layer> cvLayer;
 };
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -2101,7 +2101,8 @@ struct Net::Impl
                     auto ieInpNode = inputNodes[i].dynamicCast<InfEngineNgraphNode>();
                     CV_Assert(oid < ieInpNode->node->get_output_size());
 #if INF_ENGINE_VER_MAJOR_GT(2020020000)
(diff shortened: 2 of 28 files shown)


modules/dnn/src/ie_ngraph.hpp
```diff
@@ -91,16 +91,18 @@ class InfEngineNgraphNode : public BackendNode
...
                         std::vector<Mat>& internals);
 
     InfEngineNgraphNode(std::shared_ptr<ngraph::Node>&& _node);
     InfEngineNgraphNode(std::shared_ptr<ngraph::Node>& _node);
 
     void setName(const std::string& name);
+    ngraph::Output<ngraph::Node> GetOidOutput() const;
 
     // Inference Engine network object that allows to obtain the outputs of this layer.
     std::shared_ptr<ngraph::Node> node;
```
dkurt, line 104: Maybe we just can replace `node` type to `ngraph::Output<ngraph::Node>`?

l-bat: We need to store `ngraph::OutputVector` and `oid`. And we have some logic with nodes to release them after creating ngraph_function.

dkurt: Which logic? Can we use fields of ngraph::Output instead? Can we use Node* from ngraph::Output where it's not possible?
My main goal is to reduce number of changed files. Each layer returns ngraph::Output and std::shapred_ptr<Node> can be casted to each other explicitly (i.e. each layer accepts ngraph::Output as input but we pass shared ptr).

l-bat: Each layer returns `ngraph::OutputVector`. And we need to get specific` ngraph::Output` inside layer. Logic with setting names for node. If store OutputVector, then we have problems with same names.

dkurt: But not to drop `std::shared_ptr<Node>`?

l-bat: To store node name

dkurt: So maybe we can use `get_node()`  to get Node* if needed?

https://github.com/NervanaSystems/ngraph/blob/0af332260d9dd44a5722a1664f48781cef4c6456/src/ngraph/node_output.hpp#L73

l-bat: How it helps to reduce number of changed files? I think if we store `ngraph::OutputVector` instead of `std::shared_ptr<Node>` this increases number of changed lines. Used get_node() is less convenient than a node directly.

dkurt: In example, can we keep
```patch
-    auto& ieInpNode = nodes[0].dynamicCast<InfEngineNgraphNode>()->node;
+    auto ieInpNode = nodes[0].dynamicCast<InfEngineNgraphNode>()->GetOidOutput();
```
is `node` will be `ngraph::Output<ngraph::Node>`?

```
ngraph::Output<ngraph::Node> InfEngineNgraphNode::GetOidOutput() const {
    return node->output(oid);
}
```

l-bat: 1. We need to store ngraph::OutputVector.
2. Inside layers we need get current ngraph::Output<ngraph::Node>

PR #17033, opened 2020-04-11
Update exit status code

divmadan:
### Description

Not a big change, just changed exit status code for errors to `exit(-1)`, to maintain uniformity across the script.


alalek:
Unnecessary change with no value (doesn't fix any real problem).
Closing.

P.S. `-1` exit code is bad for cross-platform apps.

Issue #16598, opened 2020-02-17
Decode with multiple GPUs: How can I specify whichi gpu to use in cv::VideoCapture()?


AlbertHuyb:
##### System information (version)

- OpenCV => :grey_question:
- Operating System / Platform => :grey_question:
- Compiler => :grey_question:

PR #17064, opened 2020-04-14
GPU device selection for FFmpeg backend

stolyarchuk:
This PR adds the the ability to specify GPU device for VideoCapture FFmpeg decoding. 
Currently only CUDA decoders are supported. i.e. all *_cuvid ffmpeg codecs. Under the hood it explicitly initializes ffmpeg hw context with the selected gpu device id.

To use non-default gpu device one have to add "hwaccel_device" argument to OPENCV_FFMPEG_CAPTURE_OPTIONS environment variable
(eg.: "hwaccel_device;1|video_codec;h264_cuvid")

resolves opencv/opencv#16598


modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -511,10 +517,14 @@ struct CvCapture_FFMPEG
 
     AVPacket          packet;
     Image_FFMPEG      frame;
-    struct SwsContext *img_convert_ctx;
-
+    SwsContext      * img_convert_ctx;
     int64_t frame_number, first_frame_number;
 
+#if LIBAVUTIL_BUILD >= CALC_FFMPEG_VERSION(56,14,100)
```
mshabunin, line 523: I believe all new preprocessor checks should be shorthanded to something like `#ifdef ENABLE_FFMPEG_HW`

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -887,6 +903,22 @@ class InternalFFMpegRegister
     }
 };
 
+#if LIBAVUTIL_BUILD >= CALC_FFMPEG_VERSION(56,14,100)
+int CvCapture_FFMPEG::init_hw(AVCodecContext *ctx, const char *hwaccel_device)
+{
+    int err = 0;
+
+    if ((err = av_hwdevice_ctx_create(&hw_device_ctx, AV_HWDEVICE_TYPE_CUDA, hwaccel_device,
```
mshabunin, line 911: What about other HW decoders: VDPAU, VAAPI, DXVA2?

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -77,6 +77,11 @@ extern "C" {
 
 #include <libavutil/mathematics.h>
 
+#if LIBAVUTIL_BUILD >= CALC_FFMPEG_VERSION(56,14,100)
```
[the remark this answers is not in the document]
stolyarchuk: Thank you for pointing me to this doc file. I will have a deeper look into it and resolve this in the next update.

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -1006,6 +1038,19 @@ bool CvCapture_FFMPEG::open( const char* _filename )
             } else {
                 codec = avcodec_find_decoder_by_name(av_dict_get(dict, "video_codec", NULL, 0)->value);
             }
+#if LIBAVUTIL_BUILD >= CALC_FFMPEG_VERSION(56,14,100)
+            if (av_dict_get(dict, "hwaccel_device", NULL, 0) != NULL)
+            {
+                const char *hwaccel_device =
+                    av_dict_get(dict, "hwaccel_device", NULL, 0)->value;
```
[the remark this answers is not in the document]
stolyarchuk: i'm going to implement all other hwaccel methods as @mshabunin suggested, thus   "hwaccel_device" would cover not only cuda devices.

modules/videoio/src/cap_ffmpeg_impl.hpp
```diff
@@ -1006,6 +1038,19 @@ bool CvCapture_FFMPEG::open( const char* _filename )
...
+
+                if (init_hw(enc, hwaccel_device) < 0)
+                {
+                    av_buffer_unref(&hw_device_ctx);
```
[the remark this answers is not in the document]
stolyarchuk: seems it is unnecessary at all, as hw_device_ctx is to be set to NULL if av_hwdevice_ctx_create call fails.

modules/videoio/src/cap_ffmpeg_impl.hpp
stolyarchuk: Thank you for suggestion. Should I shorthand this directive to smth like ENABLE_FFMPEG_HW_SELECTION, because ffmpeg hw can be initialized implicitly with the default device?

modules/videoio/src/cap_ffmpeg_impl.hpp
stolyarchuk: Will be implemented

modules/videoio/src/cap_ffmpeg_impl.hpp
mshabunin: Yes, sounds good.

asmorkalov:
@stolyarchuk Do you have a chance to resolve comments?

stolyarchuk:
Yes. Unfortunately i m not very familiar with the ffmpeg api and a little busy with my primary work. Seems like i have to implement new ffmpeg send/receive api and not to break opencv private things. Anyway this ugly patch look like being working.

Issue #9096, opened 2017-07-05
Nearest neighbor interpolation does not give expected results


gerhardneuhold:
##### System information (version)

- OpenCV => 2.4.9
- Operating System / Platform => Ubuntu 16.04
- Compiler => gcc 5.4.0

##### Detailed description

Nearest neighbor interpolation using cv2.resize does not give expected results.

##### Steps to reproduce
```
import cv2
import numpy as np

a = np.array([[0, 1, 2, 3, 4]], dtype=np.uint8)
print 'nearest', cv2.resize(a, dsize=(3, 1), interpolation=cv2.INTER_NEAREST)
```
gives
```
nearest [[0 1 3]]
```
expected
```
nearest [[0 2 4]]
```

sergiud:
How did you come up with expected result?

The downsampling factor is 3/5 (in horizontal direction) meaning destination columns 0, 1, 2 are mapped to source columns with indices 0, 1 * 5/3 and 2 * 5/3 (i.e., 0, 1.67 and 3.33). After rounding half up (probably what you expect), the values of the corresponding columns are 0, 2, 3.

Either way, it seems OpenCV rounds towards zero instead of rounding half away from zero (not verified).

gerhardneuhold:
You can compute the coordinates as following:
```
np.floor((np.arange(3) + 0.5) / 3. * 5.)
```
to get the expected result. PIL works accordingly:
```
from PIL import Image
print np.asarray(Image.fromarray(a).resize((3,1)))
```
gives:
```
[[0 2 4]]
```

sergiud:
Why do you shift by 0.5 before transforming the coordinates? Do you assume the center of the pixel to be at .5 fraction of the pixel coordinate? I don't know much about conventions used PIL.

`np.round((np.arange(3)) / 3. * 5.)` produces `[[0, 2, 3]]`.

gerhardneuhold:
Exactly, treating pixels as squares with center at 0.5.

gerhardneuhold:
@sergiud The issue is that opencv's `resize` function and your coordinate example lead to a slight shift of the pixels to the right/bottom of the image in addition to scaling, hence not only performs scaling but also a translation of the pixels. 

@alalek What is the reason for the `question (invalid)` label?

sergiud:
@gerhardneuhold I agree.

alalek:
[...]

Issue #10146, opened 2017-11-23
Error in resize method when using nearest neighbour interpolation


Kirill888:
##### System information (version)
- OpenCV => master
- Operating System / Platform => all
- Compiler => all

##### Detailed description

When resizing image with nearest neighbour source image pixel coordinates are computed incorrectly. For example shrink a 5x5 image down to 1x1, you should get center pixel value in the result, instead top-left pixel value is present in the result.

Incorrect code is here:

https://github.com/opencv/opencv/blob/981009ac1f06106244ac52b16a20a4dc337ad816/modules/imgproc/src/resize.cpp#L235
Should be instead:
```.cpp
   int sx = cvFloor(x*ifx + (ifx-1)*0.5);
```
and here:

https://github.com/opencv/opencv/blob/981009ac1f06106244ac52b16a20a4dc337ad816/modules/imgproc/src/resize.cpp#L144
Should be instead:
```.cpp
   int sy = std::min(cvFloor(y*ify + (ify-1)*0.5), ssize.height-1);
```

There are probably more optimized versions (maybe AVX version) that are also using wrong math, I didn't look for them.

I have a more detailed explanation with equations in here:

https://gist.github.com/Kirill888/6b187e65f088971a48d078df72d27d32

I'm aware of this issue #9096, it's the same problem, but I thought I'll create new issue anyway to discuss the solution for it.

alalek:
> nearest neighbour

Is designed be fast as much as possible with some adequate result. Speed has higher priority in NN mode.
Usually used to quickly resize image down by scales up to 2.

Consider to use resize area if you want "more accurate" results.


BTW, proposed formula doesn't work well with downscaling with even factors (2,4,6).

Try this scenario in your experiments to see differences:
- "B" = downscale "A" by N; "C" = upscale "B" by N; compare result with original: "A" vs "C". Use odd, even and non-integer scales
- upscale + downscale
Check pixels near image borders.


> Should be instead

This formula is used in OpenVX (see #10130). We know this.

Kirill888:
> BTW, proposed formula doesn't work well with downscaling with even factors (2,4,6).

Formula is correct regardless of the scaling factor (like you said standard implementation in OpenVX is using it, see link below). But yes result will look ugly in this special, since lookup pixel coordinates will fall exactly on pixel boundaries in the source image. Using NN for downscaling by factor of 2 is silly, since area is just as fast in this special case, maybe faster and is much better visually.

[...]

PR #17068, opened 2020-04-14
PIL

AlexanderShashkin:
I would like to add a possibility to resize in a "PIL nearest neighbor style".
I've started with the issue
https://github.com/opencv/opencv/issues/9096
Than try it by myself and see differencies:
image
Left is a original binary mask, middle is a PIL resize and right is OpenCV resize.
Both libraries were launched with nearest neighbor algorithm.


asmorkalov:
@AlexanderShashkin Thanks for the contribution. Could you describe the purpose of PIL-style resize? What is your use-case? Also please take a look CI bot failures.

dkurt:
We definitely know that there is difference between OpenCV and PIL bilinear resize. Is that the same for NN interpolation? I think there are difference only with specific configurations like non integer scale factor (i.e. 10x11 to 13x14) or with downsampling. So please provide an example when OpenCV NN resize does not match PIL's one. Thanks!

/cc @l-bat

asmorkalov:
@AlexanderShashkin thanks for detailed description.

AlexanderShashkin:
> We definitely know that there is difference between OpenCV and PIL bilinear resize. Is that the same for NN interpolation? I think there are difference only with specific configurations like non integer scale factor (i.e. 10x11 to 13x14) or with downsampling. So please provide an example when OpenCV NN resize does not match PIL's one. Thanks!
> 
> /cc @l-bat



> @AlexanderShashkin Thanks for the contribution. Could you describe the purpose of PIL-style resize? What is your use-case? Also please take a look CI bot failures.

Sorry for bot failures, this is because of test file that I included to show the difference. I've deleted it  for now and than I will add test.

modules/imgproc/include/opencv2/imgproc.hpp
```diff
@@ -261,6 +261,8 @@ enum InterpolationFlags{
     /** flag, fills all of the destination image pixels. If some of them correspond to outliers in the
     source image, they are set to zero */
     WARP_FILL_OUTLIERS   = 8,
+    /** PIL nearest neighbor interpolation */
+    INTER_NEAREST_PIL = 9,
```
mshabunin, line 265: It would be nice to have slightly more verbose description in the documentation, something like `differs from INTER_NEAREST in the way it calculates pixel coordinates: center of the square instead of top left corner. For example, pixel (0, 0) will have coordinates (0.5, 0.5) in this mode.`

mshabunin: Won't this new enum value interfere with flags? `INTER_NEAREST_PIL = 9 // 1001` will imply `WARP_FILL_OUTLIERS = 8 // 1000` flag set. Maybe it is better to add new flag: `WARP_SQUARE_PIXELS` or `WRAP_SHIFT_COORDINATES` with value `32`?

AlexanderShashkin:
> We definitely know that there is difference between OpenCV and PIL bilinear resize. Is that the same for NN interpolation? I think there are difference only with specific configurations like non integer scale factor (i.e. 10x11 to 13x14) or with downsampling. So please provide an example when OpenCV NN resize does not match PIL's one. Thanks!
> 
> /cc @l-bat



> @AlexanderShashkin Thanks for the contribution. Could you describe the purpose of PIL-style resize? What is your use-case? Also please take a look CI bot failures.

Sorry for bot failures, this is because of test file that I included to show the difference. I've deleted it  for now and than I will add test.

> We definitely know that there is difference between OpenCV and PIL bilinear resize. Is that the same for NN interpolation? I think there are difference only with specific configurations like non integer scale factor (i.e. 10x11 to 13x14) or with downsampling. So please provide an example when OpenCV NN resize does not match PIL's one. Thanks!
> 
> /cc @l-bat

Yes, these are really differencies in NN algorithm as well. I've added pictures in a descriprion of this PR. The original binary mask ias 256x256 and than it was downscaled and upscaled with thae factor of 16.
I've implemented NN algorithm as I've found in PIL sources (in Python) that binary images are resized only with NN algorithm no matter which flag was given to the function (bilinear or whatever):
if self.mode in ("1", "P"):
            resample = NEAREST
Binary images is "1" and for them only NEAREST in PIL

Now my solution works only for grayscale and squared images, I'm going to support not only squared and add test. 
Could you please advice, which tests and docs will be enough?

modules/imgproc/src/resize.cpp
```diff
@@ -1104,6 +1104,79 @@ resizeNN( const Mat& src, Mat& dst, double fx, double fy )
...
+            int yi;
+            if (yo < 0)
+                yi = -1;
+            else
+                yi = (int)yo;
+            if(yi >= 0 && yi < ssize.height)
+            {
+                for( x = 0; x < dsize.width; x++ )
+                {
+                    dst.at<char>(y, x) = src.at<char>(yi, x_ofs[x]);
```
mshabunin, line 1134: You support only `CV_8UC1`. What about other types of `Mat`? You should at least add `Not implemented` assertion.

modules/imgproc/src/resize.cpp
```diff
@@ -3723,6 +3796,11 @@ void resize(int src_type,
         return;
     }
 
+    if (interpolation == INTER_NEAREST_PIL) {
+      resizeNNPIL(src, dst, inv_scale_x, inv_scale_y);
```
mshabunin, line 3800: This functionality must be tested.

samples/cpp/resize_pil.cpp
```diff
@@ -0,0 +1,32 @@
+#include <opencv2/opencv.hpp>
+//Include file for every supported OpenCV function
+
+using namespace cv;
+
+int main(int argc, char** argv)
+{
+  Mat image;
+  image = imread("d:\\mask.bmp", IMREAD_GRAYSCALE);  // Read the file
```
mshabunin, line 9: I think it is better to use synthetic image, maybe gradient?

samples/cpp/resize_pil.cpp
```diff
@@ -0,0 +1,32 @@
...
+  resize(image, target_image, Size(16, 16), 0, 0, INTER_NEAREST_PIL);
+  for (int row = 0; row < target_image.rows; ++row) {
+      for (int col = 0; col < target_image.cols; ++col) {
+        std::cout << abs((int) target_image.at<char>(row, col)) << ' ';
+      }
+      std::cout << "\n";
+  }
+  resize(target_image, image, Size(256, 256), 0, 0, INTER_NEAREST_PIL);
+
+  imshow("Target image", image);
```
mshabunin, line 29: Perhaps it makes sense to show results of both `NEAREST_PIL` and `NEAREST` to show difference, otherwise the sample does not have any value. Or, even better, all modes and flags should be visualized with `lena.jpg` and/or some synthetic images - that would be a great example.

dkurt: Maybe it's even better to have Python sample instead of C++ which can run both  PIL and OpenCV and compare the difference between, INTER_NEAREST, potential INTER_NEAREST_PIL and PIL?

modules/imgproc/src/resize.cpp
AlexanderShashkin: I've found in resize.cpp similar situation:
if (interpolation == INTER_LINEAR_EXACT && (_src.depth() == CV_32F || _src.depth() == CV_64F))
        interpolation = INTER_LINEAR; // If depth isn't supported fallback to generic resize

and add my code:
if (interpolation == INTER_NEAREST_PIL && _src.depth() != CV_8UC1)
        interpolation = INTER_NEAREST; // INTER_NEAREST_PIL is supported only for grayscale images

If image is not a mask (binary) or grayscale  generic INTER_NEAREST will be used. I suppose it could be better than Not implemented assertion. What do you think?
In case NotImplemented is a demand - I'll do it.

modules/imgproc/src/resize.cpp
AlexanderShashkin: I've added tests in test_resize_pil.cpp

samples/cpp/resize_pil.cpp
AlexanderShashkin: Yes, I've added  tests with gradient.png file

samples/cpp/resize_pil.cpp
AlexanderShashkin: Yes, sure. I've created resize_pil.py in which I've demonstrated:
1. Difference between cv.INTER_NEAREST and PIL NEAREST
2. Equivalence betwen cv.INTER_NEAREST_PIL and PIL NEAREST 
I've used lena.jpg

samples/cpp/resize_pil.cpp
AlexanderShashkin: @mshabunin, I've added test with gradient.png but my tests haven't passed due to absence of this file in test folder. Here is my code:

_const string root = cvtest::TS::ptr()->get_data_path();
const string filename = root + "gradient.png";
Mat img;
ASSERT_NO_THROW(img = imread(filename, IMREAD_GRAYSCALE));
ASSERT_FALSE(img.empty());_ 

and the error is:
_Value of: img.empty()
  Actual: true
Expected: false_

Should I add tests with another image or it is possible to add gradient.png in test data folder as well?

mshabunin: I meant that you can generate source image on the fly, like [here](https://github.com/opencv/opencv/issues/9096#issuecomment-333673380).

AlexanderShashkin: OK, I've done it in test_resize_pil.cpp

modules/imgproc/src/resize.cpp
mshabunin: Case with `INTER_LINEAR_EXACT` is slightly different, it is same algorithm as `INTER_LINEAR` but with stable calculation method which produces same results on all platforms. While `INTER_NEAREST_PIL` is separate algorithm and can produce completely different results so the assertion is more appropriate.

samples/python/resize_pil.py
```diff
@@ -0,0 +1,59 @@
...
+    except:
+        fn = 'lena.jpg'
+
+    img_cv_lena = cv.imread(cv.samples.findFile(fn), cv.IMREAD_GRAYSCALE)
+
+    if img_cv_lena is None:
+        print('Failed to load image file:', fn)
+        sys.exit(1)
+
+    img_pil_lena = Image.fromarray(np.array(img_cv_lena))
```
dkurt, line 33: `np.array` is extra here

samples/python/resize_pil.py
```diff
@@ -0,0 +1,59 @@
...
+
+    img_cv_lena_resized = cv.resize(img_cv_lena, (231,322), interpolation=cv.INTER_NEAREST)
+    img_pil_lena_resized = img_pil_lena.resize((231,322), Image.NEAREST)
+    assert (np.array(img_cv_lena_resized) != np.array(img_pil_lena_resized)).any()
```
dkurt, line 37: Why do we need assert?

samples/python/resize_pil.py
```diff
@@ -0,0 +1,59 @@
...
+import sys
+
+from PIL import Image
+
+def main():
+
+    try:
+        fn = sys.argv[1]
+    except:
+        fn = 'lena.jpg'
```
dkurt, line 25: Please use ArgumentParser

modules/imgproc/src/resize.cpp
AlexanderShashkin: OK, I've added 
if (interpolation == INTER_NEAREST_PIL && _src.depth() != CV_8UC1)        
        CV_Error(Error::StsNotImplemented, "INTER_NEAREST_PIL supports only CV_8UC1");

samples/python/resize_pil.py
AlexanderShashkin: OK, removed

samples/python/resize_pil.py
AlexanderShashkin: I wanted to note that cv.INTER_NEAREST differs from  Image.NEAREST

samples/python/resize_pil.py
AlexanderShashkin: OK, I've done it

samples/python/resize_pil.py
dkurt: But it will stop script execution here with `AssertionError` and traceback

samples/python/resize_pil.py
```diff
@@ -0,0 +1,59 @@
...
+from PIL import Image
+import argparse
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("fname", nargs='?', help="picture filename to test resize", default='lena.jpg')
+    args = parser.parse_args()
+    fn = args.fname
+
+    img_cv_lena = cv.imread(cv.samples.findFile(fn), cv.IMREAD_GRAYSCALE)
```
dkurt, line 27: as for me, `_lena` suffix is extra. What if user have chosen different image? Looks confusing.

samples/python/resize_pil.py
```diff
@@ -0,0 +1,59 @@
...
+    img_cv_lena_resized = cv.resize(img_cv_lena, (231,322), interpolation=cv.INTER_NEAREST_PIL)
+    img_pil_lena_resized = img_pil_lena.resize((231,322), Image.NEAREST)
+    assert (np.array(img_cv_lena_resized) == np.array(img_pil_lena_resized)).all()
+    print('cv.INTER_NEAREST_PIL is equal to Image.NEAREST')
+    resized_images = np.hstack((img_cv_lena_resized, img_pil_lena_resized))
+    cv.imshow('LEft is OpenCV INTER_NEAREST_PIL, right is PIL Image.NEAREST', resized_images)
+
+    cv.waitKey()
+
+    print('Done')
```
dkurt, line 53: Is that necessary?

samples/python/resize_pil.py
AlexanderShashkin: Yes, I've removed it

samples/python/resize_pil.py
AlexanderShashkin: I tested it before and I've tested it once again. 
assert (np.array(img_cv_lena_resized) **!=** np.array(img_pil_lena_resized)).**any()**
As cv.INTER_NEAREST really differs from Image.NEAREST we really have difference at least in in one pixel. So it isn't stop, goes further and print 
_cv.INTER_NEAREST is not equal to Image.NEAREST_

catree:
I don't think using `INTER_NEAREST_PIL` naming is a good idea. From the [mentioned issue](https://github.com/opencv/opencv/issues/9096), the issue is more general and comes from the OpenCV choice to use top-left pixel coordinates. So instead of `PIL` suffix, something related to the pixel coordinate convention with the corresponding info in the doc?

Looks like center-pixel coordinate is theoretically better since it does not introduce shift in image. But unfortunately the problem is more general since top-left pixel coordinate is the convention in OpenCV.

This PR can be a first step, but it should be done in a way that the transition should be seamless if in the future center-pixel coordinate convention can be used for image resizing, warping, ...
I mean, if for instance using CMake option (again just an example) to choose the pixel coordinate convention is the chosen approach in the future, having `INTER_NEAREST_PIL` or `INTER_NEAREST_CENTER_PIX_CONVENTION` can become useless.

AlexanderShashkin:
> This PR can be a first step, but it should be done in a way that the transition should be seamless if in the future center-pixel coordinate convention can be used for image resizing, warping, ...
I suppose that we could start from this small step to help people, who try reproducing PIL results  with OpenCV library. After that we can expect feedback from users and understand which way (CMake option or some flag) is preferable for  next steps.

samples/python/resize_pil.py
AlexanderShashkin: Dear Colleagues, any questions\suggestions so far?

AlexanderShashkin:
> @AlexanderShashkin thanks for detailed description.

@asmorkalov , could you please remove tags needs docs and need tests as I've added them both

asmorkalov:
@AlexanderShashkin good job! done.

asmorkalov:
@mshabunin @alalek Are you ready to merge the PR?

mshabunin:
I agree with @catree , we need to think about better naming for this mode. Maybe it should be another flag?

catree:
Yes, probably another flag would be better. To be able to do something like:
```
dst = cv.resize(img, new_size, \
    interpolation=cv.INTER_NEAREST + cv.CENTER_PIXEL_COORDINATES)
dst = cv.warpAffine(img, M, new_size, \
    interpolation=cv.INTER_LINEAR_EXACT + \
    cv.CENTER_PIXEL_COORDINATES + cv.WARP_FILL_OUTLIERS)
//same thing for cv.warpPerspective, cv.remap, ....?
```

Not sure, but I think pixel coordinates convention should matter also for `getRotationMatrix2D()`, and probably some other functions.

---

Comparison between OpenCV, Pillow, scikit-images and Matlab NN interpolation mode.

```
#!/usr/bin/env python
# -*- coding: utf-8 -*-

# Python 2/3 compatibility

import numpy as np
import cv2 as cv
print('OpenCV:', cv.__version__)
import PIL
from PIL import Image
print('PIL:', PIL.__version__)
import skimage
from skimage.transform import resize
print('scikit-image:', skimage.__version__)

def resize_compare(size, new_size):
    img = np.arange(size[0]*size[1], dtype=np.uint8).reshape(size)
    print('\nimg:\n', img)

    # OpenCV
    img_resize_OpenCV = cv.resize(img, (new_size[1], new_size [0]), interpolation=cv.INTER_NEAREST)
    print('\nOpenCV:\n', img_resize_OpenCV)

    # Pillow
    im = Image.fromarray(img)
    img_resize_PIL = im.resize((new_size[1], new_size [0]), resample=Image.NEAREST)
    print('\nPIL:\n', np.array(img_resize_PIL))

    # scikit-image
    img_resize_sk = resize(img, new_size, order=0, preserve_range=True).astype(np.uint8)
    print('\nskimage:\n', img_resize_sk)

size = (3,5)
new_size = (5,7)
resize_compare(size, new_size)

size = (2,3)
new_size = (4,6)
resize_compare(size, new_size)
```

```
clearvars;
clc;

iw = 5;
ih = 3;
ow = 7;
oh = 5;

img = reshape(uint8(0:iw*ih-1), [iw, ih])'
img_resize = imresize(img, [oh, ow], 'nearest')

iw = 3;
ih = 2;
ow = iw*2;
oh = ih*2;

img = reshape(uint8(0:iw*ih-1), [iw, ih])'
img_resize = imresize(img, [oh, ow], 'nearest')
```

Results:

```
OpenCV: 4.3.0-dev
PIL: 7.1.2
scikit-image: 0.14.2

img:
 [[ 0  1  2  3  4]
 [ 5  6  7  8  9]
 [10 11 12 13 14]]

OpenCV:
 [[ 0  0  1  2  2  3  4]
 [ 0  0  1  2  2  3  4]
 [ 5  5  6  7  7  8  9]
 [ 5  5  6  7  7  8  9]
 [10 10 11 12 12 13 14]]

PIL:
 [[ 0  1  1  2  3  3  4]
 [ 0  1  1  2  3  3  4]
 [ 5  6  6  7  8  8  9]
 [10 11 11 12 13 13 14]
 [10 11 11 12 13 13 14]]

skimage:
 [[ 0  1  1  2  3  3  4]
 [ 0  1  1  2  3  3  4]
 [ 5  6  6  7  8  8  9]
 [10 11 11 12 13 13 14]
 [10 11 11 12 13 13 14]]

img:
 [[0 1 2]
 [3 4 5]]

OpenCV:
 [[0 0 1 1 2 2]
 [0 0 1 1 2 2]
 [3 3 4 4 5 5]
 [3 3 4 4 5 5]]

PIL:
 [[0 0 1 1 2 2]
 [0 0 1 1 2 2]
 [3 3 4 4 5 5]
 [3 3 4 4 5 5]]

skimage:
 [[0 0 1 1 2 2]
 [0 0 1 1 2 2]
 [3 3 4 4 5 5]
 [3 3 4 4 5 5]]
```

```
img =

  3×5 uint8 matrix

    0    1    2    3    4
    5    6    7    8    9
   10   11   12   13   14


img_resize =

  5×7 uint8 matrix

    0    1    1    2    3    3    4
    0    1    1    2    3    3    4
    5    6    6    7    8    8    9
   10   11   11   12   13   13   14
   10   11   11   12   13   13   14


img =

  2×3 uint8 matrix

   0   1   2
   3   4   5


img_resize =

  4×6 uint8 matrix

   0   0   1   1   2   2
   0   0   1   1   2   2
   3   3   4   4   5   5
   3   3   4   4   5   5
```

Looks like output of Pillow, scikit-images and Matlab match.

homm:
PIL has the same problems with arithmetic which you try to fix and is abandoned for a long time. The geometry for affine transformations was fixed in Pillow. Please consider change the name of the PR and correct all comments to prevent misleading.

AlexanderShashkin:
> Yes, probably another flag would be better. To be able to do something like:
> 
> ```
> dst = cv.resize(img, new_size, \
>     interpolation=cv.INTER_NEAREST + cv.CENTER_PIXEL_COORDINATES)


> I agree with @catree , we need to think about better naming for this mode. Maybe it should be another flag?

@mshabunin , what do you think about cv.CENTER_PIXEL_COORDINATES? If it is accepted, I can change the flag and tests for resize.
@catree has also mentioned some other functions, probably we could start thinking about them as well

AlexanderShashkin:
> PIL has the same problems with arithmetic which you try to fix and is abandoned for a long time. The geometry for affine transformations was fixed in Pillow. Please consider change the name of the PR and correct all comments to prevent misleading.

@homm yes, we're definetely talking about Pillow. I've seen your comment to the issue
https://github.com/opencv/opencv/issues/10146
and reproduced your example from https://stackoverflow.com/questions/49764662/nearest-neighbor-image-resizing-wrong-using-scipy-and-pil/49784207
INTER_NEAREST_PIL gives exactly the same result, so I've resized
{1, 2, 3,
 4, 5, 6,
 7, 8, 9}
to
[  1,   1,   1,   2,   2,   2,   3,   3,   3;
   1,   1,   1,   2,   2,   2,   3,   3,   3;
   1,   1,   1,   2,   2,   2,   3,   3,   3;
   4,   4,   4,   5,   5,   5,   6,   6,   6;
   4,   4,   4,   5,   5,   5,   6,   6,   6;
   4,   4,   4,   5,   5,   5,   6,   6,   6;
   7,   7,   7,   8,   8,   8,   9,   9,   9;
   7,   7,   7,   8,   8,   8,   9,   9,   9;
   7,   7,   7,   8,   8,   8,   9,   9,   9]

asmorkalov:
@mshabunin @dkurt Friendly reminder. I re-triggered CI build to update status.

mshabunin:
@asmorkalov , we need design decision on adding a flag for shifted pixel coordinates mode and its naming (see https://github.com/opencv/opencv/pull/17068#issuecomment-629847301). Implementations for other resize modes and same flag for other functions (e.g. warpAffine) can be added later (should it be named differently for each function? or should it be placed in separate enum?).

modules/imgproc/src/resize.cpp
vpisarev: does not look very fast.
```
const char* srcrow = src.ptr<char>(yi);
char* dstrow = dst.ptr<char>(y);
for(x = 0; x < dsize.width; x++)
    dstrow[x] = srcrow[x_ofs[x]];
```
should be a bit better

vpisarev:
@AlexanderShashkin, thank you for the patch!
@dkurt, @mshabunin, @asmorkalov, my vote is for INTER_NEAREST_PIL. It would be nice to support color images and speedup the inner loop, but maybe the patch can be integrated and then the code can be polished further.

homm:
> my vote is for INTER_NEAREST_PIL

Again, PIL originally has the same issue. It was fixed only in Pillow

AlexanderShashkin:
> @AlexanderShashkin, thank you for the patch!
> @dkurt, @mshabunin, @asmorkalov, my vote is for INTER_NEAREST_PIL. It would be nice to support color images and speedup the inner loop, but maybe the patch can be integrated and then the code can be polished further.

@vpisarev , thanx! I can add a support for color images when naming is approved

AlexanderShashkin:
> > my vote is for INTER_NEAREST_PIL
> 
> Again, PIL originally has the same issue. It was fixed only in Pillow

yes, sure, as I've mentioned before, we're definetely talking about Pillow and there is no that issue in my INTER_NEAREST_PIL implementation

homm:
> and there is no that issue in my INTER_NEAREST_PIL implementation

And you still call it PIL…

catree:
**Please disregard my comments in this thread above.**

From my point of view:

---

About adding a flag for pixel center coordinates:

- unfortunately not gonna happen given the amount of work and how everything is possibly entangled between C++, SIMD code, OCL, Cuda, resize code and related functions like `warpAffine`, ... (too much work to make everything consistent)
- did a quick test between Pillow, scikit-image and Matlab and none of these can produce a consistent result
- maybe all of these methods use center pixel coordinates and the difference comes from rounding method?
- **I am definitively not qualified to speak about these resize and pixel coordinates issues**

---

About using `INTER_NEAREST_PILLOW`. I just strongly disagree to name it using the name of a library. Moreover when the NN resize method is trivial.
I can only propose `INTER_NEAREST_2` instead, and in the documentation add something like: "produce same results than PIL, scikit-image or Matlab like NN resize method".

Speaking about documentation, it would be great to add the formula to iterate over the pixels. From the code I have no idea how it works. I believe the correct formula to take into account pixel center coordinates in the **naive implementation** is the following:
```
u_dst = (u_src + 0.5) * scale_u - 0.5;
v_dst = (v_src + 0.5) * scale_v - 0.5;
```

Plus the rounding / border methods.
This would be very useful for learning purpose for OpenCV users.

---

About this PR:

- honestly **I see no urge to have this PR merged**,
- especially if only grayscale is supported
- I see some for loop variables declared outside. For me this is useless and not recommended in C++ but I might be wrong. Since I am using github code to learn best practices, and I believe other people too, it would be great if **more efforts toward code quality** would be made.

---

**Finally my last words.**

Whatever solution, flag name is chosen, from this statement:
> once a code is inside into OpenCV, it stays for years

**Please check that the user will not be able to shoot himself in the foot**:
- e.g. the enum value `INTER_NEAREST_PIL = 6` for correct use in combination with other flags
- support of this flag with (possibly) related functions like other warping functions
- different mat type

alalek:
To exclude mess with common "interpolation" enum (shared between other warp functions) and avoid creation of God functions (like cvtColor) we should:
- use different separate name, like `resizePillow()`
- which has own interpolation/filter [modes](https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.resize) (`PIL.Image.NEAREST`, `PIL.Image.BOX`, `PIL.Image.BILINEAR`, `PIL.Image.HAMMING`, `PIL.Image.BICUBIC` or `PIL.Image.LANCZOS`)

Unification of resize pixel centers in 0 / 0.5 should be done as a separate process (through OE).

vpisarev:
@AlexanderShashkin, @alalek, @homm,

I played a bit with it and think and we need just one INTER_NEAREST that works well enough in all cases. Whether it will be a universal formula or some adaptive one, depending on the combination of the input and output size – we need to see.

Here are some partial cases where the behaviour is either obvious or just standard convention that people rely on:

1. 2x decimation. The formula should return even points, starting with 0: [0, 2, 4, ...]. The suggested formula gives [1, 3, 5, ...], which could be also fine, but it's incompatible with the previous behaviour, and as far as I remember, for some algorithms it's crucial difference. The formula needs to be corrected, or special formula should be used in this case.
2. decimation n to 1 where n is odd. I think, in this case most reasonable behaviour is to choose the middle of each segment of n pixels. The new formula does it correctly.
3. decimation n*2-1 to n. This is the case in the reported issue. The new formula does it correctly, the old one is incorrect.
4. decimation n*2+1 to n. The new formula seems to do it well.
5. zoom 1 to n where n is integer. The formula should, of course, give segments of equal size. The new formula provides it.

Also, preferably the code should not use floating-point arithmetics to give the same results on any hardware.

I found that the following algorithm (in Python) satisfies all conditions (1-5):

```
sw = 5
dw0 = 3
fx = (dw0+0.)/sw
dw = int(round(sw*fx))
shift = 16
delta = 1 << (shift-1)
ifx = (sw*(1 << shift) + dw//2)//dw
ifx0 = ifx//2-1

print ([(ifx0+ifx*x) >> shift for x in range(0, dw)])
```

where `sw` is input image size (we consider width, height is processed in the same way), `fx` is the supplied scale factor, `dw0` is the desired image size, from which user computed fx and passed it to the function. We first find `dw` – the output image size, and after that use solely fixed-point arithmetics.
The formula is basically equivalent to the proposed, with the only small difference how we compute the initial offset. We subtract a tiny 1/65535 from it, and this is enough to give [0, 2, 4, ...] sequence in the case of 2x decimation. This formula also gives correct results for the 3x5 => 5x7 image upscaling sample.

Maybe in general that `ifx0` needs to be corrected, depending on the combination of `sw` and `dw`.

Probably, this should be put into the current implementation of INTER_NEAREST (after OpenCV 4.4 release).

asenyaev:

asenyaev:

vpisarev:
@asmorkalov, probably, this PR can finally be closed after #28661 is merged

PR #17071, opened 2020-04-15
TickMeter: added FPS and AvgTime, improved docs, reformatted

mshabunin:
- New `TickMeter` methods: `getFPS`, `getAvgTimeSec` and `getAvgTimeMilli`
- Extracted code snippets from documentation
- Minor formatting fixes

### DIFF
--- a/modules/core/include/opencv2/core/utility.hpp
+++ b/modules/core/include/opencv2/core/utility.hpp
@@ -282,107 +282,98 @@ CV_EXPORTS_W double getTickFrequency();
 
 The class computes passing time by counting the number of ticks per second. That is, the following code computes the
 execution time in seconds:
-@code
-TickMeter tm;
-tm.start();
-// do something ...
-tm.stop();
-std::cout << tm.getTimeSec();
-@endcode
+@snippet snippets/core_various.cpp TickMeter_total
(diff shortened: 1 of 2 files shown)


modules/core/include/opencv2/core/utility.hpp
```diff
@@ -282,107 +282,100 @@ CV_EXPORTS_W double getTickFrequency();
...
+            return 0.;
+        return getTimeSec() / counter;
+    }
+
+    //! returns average time in milliseconds
+    CV_WRAP double getAvgTimeMilli() const
+    {
+        if (counter <= 0)
+            return 0.;
+        return getTimeMilli() / counter;
```
alalek, line 370: Lets reduce logic:
```
return getAvgTimeSec() * 1e3;
```

mshabunin, line 370:
```suggestion
        return getAvgTimeSec() * 1e3 / counter;
```

alalek: `if` should be removed too.

mshabunin: 🤦

PR #17077, opened 2020-04-15
core(matrix): Negative values checks

i386x:
Add checks that prevents indexing an array by negative values (issues were discovered by static code analysis with coverity).

### DIFF
--- a/modules/core/src/matrix_wrap.cpp
+++ b/modules/core/src/matrix_wrap.cpp
@@ -947,7 +947,7 @@ bool _InputArray::isContinuous(int i) const
     if( k == STD_ARRAY_MAT )
     {
         const Mat* vv = (const Mat*)obj;
-        CV_Assert(i > 0 && i < sz.height);
+        CV_Assert(i >= 0 && i < sz.height);
         return vv[i].isContinuous();
     }
 
@@ -981,21 +981,21 @@ bool _InputArray::isSubmatrix(int i) const
     if( k == STD_VECTOR_MAT )
     {
         const std::vector<Mat>& vv = *(const std::vector<Mat>*)obj;
-        CV_Assert((size_t)i < vv.size());
+        CV_Assert(i >= 0 && (size_t)i < vv.size());
         return vv[i].isSubmatrix();
     }
 
     if( k == STD_ARRAY_MAT )
     {
         const Mat* vv = (const Mat*)obj;
-        CV_Assert(i < sz.height);
+        CV_Assert(i >= 0 && i < sz.height);
         return vv[i].isSubmatrix();
     }
 
     if( k == STD_VECTOR_UMAT )
     {
         const std::vector<UMat>& vv = *(const std::vector<UMat>*)obj;
-        CV_Assert((size_t)i < vv.size());
+        CV_Assert(i >= 0 && (size_t)i < vv.size());
         return vv[i].isSubmatrix();
     }
 
@@ -1026,9 +1026,7 @@ size_t _InputArray::offset(int i) const
     if( k == STD_VECTOR_MAT )
     {
         const std::vector<Mat>& vv = *(const std::vector<Mat>*)obj;
-        if( i < 0 )
-            return 1;
-        CV_Assert( i < (int)vv.size() );
+        CV_Assert( i >= 0 && i < (int)vv.size() );
 
         return (size_t)(vv[i].ptr() - vv[i].datastart);
     }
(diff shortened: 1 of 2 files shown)


modules/core/src/matrix_wrap.cpp
```diff
@@ -1119,6 +1125,8 @@ size_t _InputArray::step(int i) const
     if (k == STD_VECTOR_CUDA_GPU_MAT)
     {
         const std::vector<cuda::GpuMat>& vv = *(const std::vector<cuda::GpuMat>*)obj;
+        if( i < 0 )
+            return 1;
```
dkurt, line 1129: Is that correct? Why not assertion?

modules/core/src/matrix_wrap.cpp
```diff
@@ -981,21 +981,21 @@ bool _InputArray::isSubmatrix(int i) const
...
-        CV_Assert((size_t)i < vv.size());
+        CV_Assert(i >= 0 && (size_t)i < vv.size());
         return vv[i].isSubmatrix();
     }
 
     if( k == STD_ARRAY_MAT )
     {
         const Mat* vv = (const Mat*)obj;
-        CV_Assert(i < sz.height);
+        CV_Assert(i > 0 && i < sz.height);
```
dkurt, line 991: Why >= somewhere and > here?

i386x: The same assertion is used also in `_InputArray::isContinuous` for `STD_ARRAY_MAT` case above. I am not sure whether `>` is intentional for `STD_ARRAY_MAT` case or it is a typo. Maybe `>=` should be used in both places?

modules/core/src/matrix_wrap.cpp
i386x: It is used at several similar places, like in conditional that handles `STD_ARRAY_MAT`. As `step` (and `offset`) are declared with `-1` to be a default value of `i`, my guess, based on [this](https://github.com/opencv/opencv/blob/3.4.10/modules/core/src/matmul.dispatch.cpp#L83), is that `if ( i < 0 ) return 1;` is preferred over assertion to prevent `mat.step()` and `mat.offset()` calls from failing.

asmorkalov:
@i386x Do have a chance to review notes?

i386x:
Hi @asmorkalov, I am afraid I not fully understand what notes do you have in your mind. Can you please be more specific?

asmorkalov:
@i386x I messed you replies above. Some of cases you touched are not well covered with tests. Could you add some cases for the fixed bug? @alalek could you provide some details on test coverage?

asmorkalov:
@386x well done! Thanks for the contributed test.

modules/core/src/matrix_wrap.cpp
vpisarev: maybe, it should be `sizeof(cuda::GpuMat>)`?

vpisarev:
:+1:

vpisarev:
@alalek, the pull request now passes all the checks, I suggest to merge it in

modules/core/src/matrix_wrap.cpp
```diff
@@ -1045,6 +1045,8 @@ size_t _InputArray::offset(int i) const
     if( k == STD_VECTOR_UMAT )
     {
         const std::vector<UMat>& vv = *(const std::vector<UMat>*)obj;
+        if( i < 0 )
+            return 1;
```
alalek, line 1049: Why is 1 here for offset? What is logic here?

I know that this is a copy-paste, but we should not copy mistakes.

IMHO we should ban this method call with negative argument value (through `CV_Assert()`).

asmorkalov:
@386x Friendly reminder.

modules/core/src/matrix_wrap.cpp
```diff
@@ -981,21 +981,21 @@ bool _InputArray::isSubmatrix(int i) const
     if( k == STD_VECTOR_MAT )
     {
         const std::vector<Mat>& vv = *(const std::vector<Mat>*)obj;
-        CV_Assert((size_t)i < vv.size());
+        CV_Assert(i >= 0 && (size_t)i < vv.size());
```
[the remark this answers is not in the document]
i386x: Oh, ok. Should I revert it than and write a note to comment?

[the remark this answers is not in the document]
alalek: I believe we can keep this as is.

modules/core/src/matrix_wrap.cpp
i386x: > Why is 1 here for offset? What is logic here?
I tried to find some design documents about `_InputArray`, but without success.
> I know that this is a copy-paste, but we should not copy mistakes.
> 
> IMHO we should ban this method call with negative argument value (through `CV_Assert()`).
Agree. The current behavior is a bit inconsistent (some kind of matrices raise an error and some return 1). I will fix the patch and tests.

asmorkalov:
@i386x Friendly reminder.

asmorkalov:
@i386x Do you have chance to finish the patch in meantime?

i386x:
@asmorkalov Sorry, I totally forgot about it. I am now waiting on build and tests to be finished on my machine and then I will submit the updated patch.

alalek (approved):
Well done! Thank you :+1:

vpisarev (approved):

PR #17088, opened 2020-04-16
G-API: Laplacian and bilateralFilter standard kernels

mpashchenkov:
### Laplacian and bilateralFilter standard kernels

Added:
- `Laplacian` kernel for CPU and GPU backends;
- Accuracy tests for `Laplacian` kernel (CPU, GPU);
- Performance tests for `Laplacian` kernel (CPU, GPU);

- `bilateralFilter` kernel for CPU and GPU backends;
- Accuracy tests for `bilateralFilter` kernel (CPU, GPU);
- Performance tests for `bilateralFilter` kernel (CPU, GPU);

Kernels call relevant cv:: functions.

```
```


mpashchenkov:
@OrestChura  Please review.

modules/gapi/test/common/gapi_imgproc_tests.hpp
```diff
@@ -50,6 +50,11 @@ GAPI_TEST_FIXTURE(SobelTest, initMatrixRandN, FIXTURE_API(CompareMats,int,int,in
     cmpF, kernSize, dx, dy)
 GAPI_TEST_FIXTURE(SobelXYTest, initMatrixRandN, FIXTURE_API(CompareMats,int,int,int,int), 5,
     cmpF, kernSize, order, border_type, border_val)
+GAPI_TEST_FIXTURE(LaplacianTest, initMatrixRandN, FIXTURE_API(CompareMats,int),
+                  2, cmpF, kernSize)
```
OrestChura, line 54: It would be better to use the same variable names everywhere: the G-API operation wrapper, kernel's run method and every other place where you expect variables reflecting the same entities (e.g: test parameters). This will give user opportunity to perceive the code in different places more easily. (I mean, e.g., kernSize -> kSize here)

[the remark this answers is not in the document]
OrestChura: I won't write this further; it's up to you whether to accept this suggestion

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
```diff
@@ -579,6 +579,91 @@ PERF_TEST_P_(SobelXYPerfTest, TestPerformance)
 
 //------------------------------------------------------------------------------
 
+PERF_TEST_P_(LaplacianPerfTest, TestPerformance)
+{
+    compare_f cmpF;
+    MatType type = 0;
+    int kernSize = 0, dtype = 0;
+    cv::Size sz;
+    cv::GCompileArgs compile_args;
```
OrestChura, line 588: Very minor comment: I'd rather use the mixedCase naming notation even if compile_args variable usually named differently as you're using it, e.g., in kernSize. But it is only my subjective opinion.

modules/gapi/perf/common/gapi_imgproc_perf_tests.hpp
```diff
@@ -32,6 +32,13 @@ class DilatePerfTest : public TestPerfParams<tuple<compare_f, MatType,int,cv::Si
 class Dilate3x3PerfTest : public TestPerfParams<tuple<compare_f, MatType,cv::Size,int, cv::GCompileArgs>> {};
 class SobelPerfTest : public TestPerfParams<tuple<compare_f, MatType,int,cv::Size,int,int,int, cv::GCompileArgs>> {};
 class SobelXYPerfTest : public TestPerfParams<tuple<compare_f, MatType,int,cv::Size,int,int, cv::GCompileArgs>> {};
+class LaplacianPerfTest : public TestPerfParams<tuple<compare_f, MatType,int,
+                                                      cv::Size,int,
```
OrestChura, line 36: mb you can carry the inheritance on the next line to polish the appearance?

modules/gapi/perf/cpu/gapi_imgproc_perf_tests_cpu.cpp
```diff
@@ -124,6 +124,25 @@ INSTANTIATE_TEST_CASE_P(SobelPerfTestCPU32F, SobelPerfTest,
...
+                                Values(CV_8UC1, CV_8UC3, CV_16SC1),
+                                Values(3, 5),
+                                Values(szVGA, sz720p, sz1080p),
+                                Values(-1, CV_16S, CV_32F),
+                                Values(cv::compile_args(IMGPROC_CPU))));
+
+INSTANTIATE_TEST_CASE_P(BilateralFilterPerfTestCPU, BilateralFilterPerfTest,
+                        Combine(Values(AbsExact().to_compare_f()),
+                                Values(CV_8UC1, CV_8UC3),
+                                Values(-1, CV_16S, CV_32F),
```
OrestChura, line 138: Is it necessary to define this parameter with 3 different values? As I can see it's not used in the test body. Is it important to check the kernel's behaviour with different output initializations?

modules/gapi/test/cpu/gapi_imgproc_tests_cpu.cpp
```diff
@@ -176,6 +176,27 @@ INSTANTIATE_TEST_CASE_P(SobelXYTestCPU32F, SobelXYTest,
...
+                                Values(-1, CV_16S, CV_32F),
+                                Values(IMGPROC_CPU),
+                                Values(AbsExact().to_compare_obj()),
+                                Values(3, 5)));
+
+INSTANTIATE_TEST_CASE_P(BilateralFilterTestCPU, BilateralFilterTest,
+                        Combine(Values(CV_8UC1, CV_8UC3),
+                                Values(cv::Size(1280, 720),
+                                       cv::Size(640, 480)),
+                                Values(-1, CV_16S, CV_32F),
```
OrestChura, line 192: The same comment as in perf_tests

modules/gapi/test/cpu/gapi_imgproc_tests_cpu.cpp
```diff
@@ -176,6 +176,27 @@ INSTANTIATE_TEST_CASE_P(SobelXYTestCPU32F, SobelXYTest,
...
                                 Values(0, 1, 255)));
 
+INSTANTIATE_TEST_CASE_P(LaplacianTestCPU, LaplacianTest,
+                        Combine(Values(CV_8UC1, CV_8UC3, CV_16SC1),
+                                Values(cv::Size(1280, 720),
+                                       cv::Size(640, 480)),
+                                Values(-1, CV_16S, CV_32F),
+                                Values(IMGPROC_CPU),
+                                Values(AbsExact().to_compare_obj()),
+                                Values(3, 5)));
```
OrestChura, line 186: Are you sure you shouldn't check more parameters of Laplacian? You can see what guys in imgproc_tests and perf_tests check if you haven't yet. If you sure it's enough then OK

OrestChura: Mb it's not needed in gapi_perf_tests, but in accuracy tests we must become confident that the kernel works quite similarly to how OpenCV function does with (almost) all range of parameters. Especially for GPU implementation.

Check for bilateral as well, please

modules/gapi/include/opencv2/gapi/imgproc.hpp
```diff
@@ -672,7 +672,6 @@ with the following \f$3 \times 3\f$ aperture:
 @note Function textual ID is "org.opencv.imgproc.filters.laplacian"
 
 @param src Source image.
-@param dst Destination image of the same size and the same number of channels as src .
```
OrestChura: Oh, and actually if you do want to specify what your kernel returns - you can use the @return which looks like this:
prscr_return

modules/gapi/src/backends/ocl/goclimgproc.cpp
```diff
@@ -147,6 +147,25 @@ GAPI_OCL_KERNEL(GOCLSobel, cv::gapi::imgproc::GSobel)
...
+    }
+};
+
+GAPI_OCL_KERNEL(GOCLBilateralFilter, cv::gapi::imgproc::GBilateralFilter)
+{
+    static void run(const cv::UMat& in, int ddepth, double sigmaColor,
+                    double sigmaSpace, int borderType, cv::UMat &out)
+    {
+        cv::bilateralFilter(in, out, ddepth, sigmaColor, sigmaSpace,
+                            borderType);
```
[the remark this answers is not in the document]
mpashchenkov: Fixed.

modules/gapi/test/common/gapi_imgproc_tests.hpp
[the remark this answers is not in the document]
mpashchenkov: Fixed.

mpashchenkov: Other tests contain the kernSize parameter, and their functions get kSize. I prefer standardization.

modules/gapi/perf/common/gapi_imgproc_perf_tests_inl.hpp
mpashchenkov: I prefer standardization.

modules/gapi/perf/common/gapi_imgproc_perf_tests.hpp
mpashchenkov: Mb. What do you think about "global alignment"?

modules/gapi/perf/cpu/gapi_imgproc_perf_tests_cpu.cpp
mpashchenkov: Removed. Tests with parameters as in OCV.

modules/gapi/test/cpu/gapi_imgproc_tests_cpu.cpp
mpashchenkov: Fixed.

modules/gapi/test/cpu/gapi_imgproc_tests_cpu.cpp
mpashchenkov: Extended.

modules/gapi/include/opencv2/gapi/imgproc.hpp
mpashchenkov: Fixed.

modules/gapi/test/common/gapi_imgproc_tests.hpp
OrestChura: I see. OK

modules/gapi/perf/common/gapi_imgproc_perf_tests.hpp
OrestChura: I don't really know.. Now your new lines are too long, as you can see. I'd prefer you to realign a little, but, honestly, it's not so important. Generally speaking, your global alignment looks good)

modules/gapi/perf/cpu/gapi_imgproc_perf_tests_cpu.cpp
```diff
@@ -126,21 +126,22 @@ INSTANTIATE_TEST_CASE_P(SobelPerfTestCPU32F, SobelPerfTest,
...
 INSTANTIATE_TEST_CASE_P(LaplacianPerfTestCPU, LaplacianPerfTest,
                         Combine(Values(AbsExact().to_compare_f()),
-                                Values(CV_8UC1, CV_8UC3, CV_16SC1),
-                                Values(3, 5),
+                                Values(CV_8UC1, CV_8UC3, CV_16UC1, CV_16SC1),
+                                Values(1, 3, 5),
                                 Values(szVGA, sz720p, sz1080p),
-                                Values(-1, CV_16S, CV_32F),
+                                Values(-1),
+                                Values(1.0),
```
OrestChura, line 133: So, as I get it right, you added one more parameter but it's actually not changing during tests (and in GPU tests it's the same). Maybe you could just hardcode it in the test body?

modules/gapi/perf/cpu/gapi_imgproc_perf_tests_cpu.cpp
OrestChura: Actually, I meant only bilateral tests. As I can see, Laplacian uses the dtype argument as it can return back a Mat with types differ from input ones. Do they check this functionality in imgproc tests? If don't, then OK

OrestChura: And, I'm sorry, should have mentioned: it is important to check almost all the functionality/range of parameters in our tests; but on the other hand there shouldn't be too many tests.. If you think all the checked parameters configuration is important, then leave them, otherwise drop them to reduce the testing time

mpashchenkov:
@AsyaPronina Please review.

modules/gapi/perf/cpu/gapi_imgproc_perf_tests_cpu.cpp
OrestChura: Decided to leave it as is

modules/gapi/perf/cpu/gapi_imgproc_perf_tests_cpu.cpp
mpashchenkov: Removed single parameters from tests. Parameters in perf_tests are cut. Parameters in accuracy tests as in OCV.

mpashchenkov:
@dmatveev  Please review.

modules/gapi/test/cpu/gapi_imgproc_tests_cpu.cpp
```diff
@@ -176,6 +176,29 @@ INSTANTIATE_TEST_CASE_P(SobelXYTestCPU32F, SobelXYTest,
...
+INSTANTIATE_TEST_CASE_P(BilateralFilterTestCPU, BilateralFilterTest,
+                        Combine(Values(CV_32FC1, CV_32FC3, CV_8UC1, CV_8UC3),
+                                Values(cv::Size(1280, 720),
+                                       cv::Size(640, 480)),
+                                Values(-1),
+                                Values(IMGPROC_CPU),
+                                Values(AbsExact().to_compare_obj()),
+                                Values(3, 5, 9),
+                                Values(20, 100),
+                                Values(10, 40),
```
alalek, line 199: > 288 tests from BilateralFilterTestCPU/BilateralFilterTest

Do we really need all of these test cases and the same set for GPU?
99% of test cases has zero code coverage increase.

The similar note is about perf tests.

modules/gapi/perf/gpu/gapi_imgproc_perf_tests_gpu.cpp
```diff
@@ -128,20 +128,20 @@ INSTANTIATE_TEST_CASE_P(SobelPerfTestGPU32F, SobelPerfTest,
 
 INSTANTIATE_TEST_CASE_P(LaplacianPerfTestGPU, LaplacianPerfTest,
                         Combine(Values(ToleranceFilter(1e-4f, 0.01).to_compare_f()),
-                                Values(CV_8UC1, CV_8UC3, CV_16UC1, CV_16SC1),
-                                Values(3, 5),
+                                Values(CV_32FC1),
```
OrestChura, line 131: I don't think it's good if types checked on CPU differ from ones on GPU

mpashchenkov: Fixed.

PR #17102, opened 2020-04-18
Flexible inputs for OpenVINO IR models

dkurt:
Support flexible inputs for OpenVINO IR models. We can work with non-default IR input resolutions but not with flexible inputs which may change between runs.

<cut/>

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/op_inf_engine.cpp
+++ b/modules/dnn/src/op_inf_engine.cpp
@@ -891,6 +891,13 @@ bool InfEngineBackendNet::isInitialized()
 #endif
 }
 
+void InfEngineBackendNet::reset()
+{
+    allBlobs.clear();
+    infRequests.clear();
+    isInit = false;
+}
+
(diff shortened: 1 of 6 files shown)


modules/dnn/src/op_inf_engine.cpp
```diff
@@ -891,6 +891,13 @@ bool InfEngineBackendNet::isInitialized()
 #endif
 }
 
+void InfEngineBackendNet::reset()
+{
+    allBlobs.clear();
+    infRequests.clear();
```
alalek, line 897: What is about multiple async requests with different shapes?

dkurt: That's good point but I think we can cover it separately because even using native Inference Engine API it's tricky. After each network reshaping it's required to load is into the plugin (`LoadNetwork`) then manage asynchronous requests from `ExecutableNetwork::CreateInferRequest`.

PR #17109, opened 2020-04-19
cmake: handle VIDEOIO_PLUGIN_LIST=all

alalek:


mshabunin:
I actually wanted to make separate options for each plug-in: `OPENCV_VIDEOIO_PLUGIN_MSMF=ON`, `OPENCV_VIDEOIO_PLUGIN_MFX=ON`, etc.

alalek:
It is for simple CI configuration to test all supported plugins (current and future).
Other variant can be added separately - but we need to define priorities of used variables.

PR #17122, opened 2020-04-21
videoio: load debug versions of plug-ins in debug builds

mshabunin:
- added suffix to debug versions of videoio plug-ins
- trying to load plug-ins with `d` suffix in debug builds

Note: FFmpeg plug-in does not work because it does not have debug version, but the file can be renamed or forced via environment and then it will work (more or less)

Validated debug builds:
- [Win64](http://pullrequest.opencv.org/buildbot/builders/precommit_opencl/builds/24118)
- [Win32](http://pullrequest.opencv.org/buildbot/builders/precommit_windows32/builds/12399)

<cut/>

```
# no plugins in Custom builds (except forced FFmpeg)
build_debug:Custom Win=ON
test_modules:Custom Win=videoio

Xbuild_debug:Win64 OpenCL=ON
Xtest_modules:Win64 OpenCL=videoio

build_debug:Win32=ON
test_modules:Win32=videoio
**WIP**
```

### DIFF
--- a/modules/videoio/CMakeLists.txt
+++ b/modules/videoio/CMakeLists.txt
@@ -31,6 +31,10 @@ file(GLOB videoio_ext_hdrs
   "${CMAKE_CURRENT_LIST_DIR}/include/opencv2/${name}/*.h"
   "${CMAKE_CURRENT_LIST_DIR}/include/opencv2/${name}/legacy/*.h")
 
+if(OPENCV_DEBUG_POSTFIX)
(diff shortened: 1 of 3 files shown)


alalek:
FFmpeg plugin on Windows is DEBUG-agnostic. Duplication of this file is not an option.

mshabunin:
How should we load plugins in debug builds then? Load only files with `d` suffix or try two variants with `d` and without `d`?

alalek:
I believe check for both variants in "Debug" mode should work well.
Also it would be nice to propagate CMake's `OPENCV_DEBUG_POSTFIX` variable into C++ code.

alalek:
Plugin loading is broken in Release builds now (see Win64 OpenCL test logs):

```.diff
-#ifdef DEBUG_POSTFIX
+#if defined(_DEBUG) && defined(DEBUG_POSTFIX)
```

PR #17133, opened 2020-04-22
GAPI: fix build, update IE handling, add OPENCV_GAPI_INF_ENGINE

alalek:
relates #16995

<cut/>

```
Xforce_builders=Custom
buildworker:Custom=linux-1
build_image:Custom=plaidml2
test_modules:Custom=gapi
test_filter:Custom=*ML* 

Xtest_modules:Custom=gapi
test_modules:Custom Win=gapi
test_modules:Custom Mac=gapi

Xbuildworker:Custom=linux-1
Xtest_opencl:Custom=OFF
Xtest_bigdata:Custom=1
Xtest_filter:Custom=*

test_bigdata:Custom Win=1
test_filter:Custom Win=*
```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/own/convert.hpp
+++ b/modules/gapi/include/opencv2/gapi/own/convert.hpp
@@ -25,23 +25,27 @@ namespace cv
         return result;
     }
 
-           cv::gapi::own::Mat to_own(Mat&&) = delete;
+    cv::gapi::own::Mat to_own(Mat&&) = delete;
 
     inline cv::gapi::own::Mat to_own(Mat const& m) {
         return (m.dims == 2)
             ?  cv::gapi::own::Mat{m.rows, m.cols, m.type(), m.data, m.step}
             :  cv::gapi::own::Mat{to_own<int>(m.size), m.type(), m.data};
     };
+
 namespace gapi
 {
 namespace own
 {
+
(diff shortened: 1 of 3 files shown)


alalek:
:+1:

modules/gapi/include/opencv2/gapi/own/convert.hpp
```diff
@@ -25,23 +25,27 @@ namespace cv
         return result;
     }
 
-           cv::gapi::own::Mat to_own(Mat&&) = delete;
+    cv::gapi::own::Mat to_own(Mat&&) = delete;
```
dmatveev, line 28: ...it was intentionally aligned that way, to make all `to_own` start at the same column

alalek: ...got it.

I have a general suggestion to put "inline", "static", "CV_EXPORTS*" and other modifiers and attributes on a separate line. So same column start would be preserved without playing with indentation.

The case below:

```.diff
-    inline cv::Mat to_ocv(Mat const& m) {
+    inline
+    cv::Mat to_ocv(Mat const& m) {
         return m.dims.empty()
             ? cv::Mat{m.rows, m.cols, m.type(), m.data, m.step}
             : cv::Mat{m.dims, m.type(), m.data};
     }

     cv::Mat to_ocv(Mat&&) = delete;
```

PR #17134, opened 2020-04-22
dnn(IE): avoid conversion to legacy CNNNetworkImpl

alalek:
Avoid calling `CNNNetworkNGraphImpl::convertToCNNNetworkImpl()` during loading of IRv10.

<cut/>

Scope: tests `DNNTestOpenVINO.models`
Message of failing on MYRIAD with IRv10:
```
[ RUN      ] DNNTestOpenVINO.models/22, where GetParam() = (NGRAPH/MYRIAD, "person-detection-retail-0013")
Exception message: OpenCV(3.4.10-dev) /build/precommit_custom_linux/3.4/opencv/modules/dnn/src/ie_ngraph.cpp:612: error: (-2:Unspecified error)
Failed to initialize Inference Engine backend (device = MYRIAD):
Unexpected CNNNetwork format: it was converted to deprecated format prior plugin's call
in function 'initPlugin'
...
```

```
Xbuild_image:Custom=ubuntu-openvino-2019r3.0:16.04
#build_image:Custom Win=openvino-2019r3.0

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*

test_bigdata:Custom Win=1
test_filter:Custom Win=*
```

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -3157,6 +3157,8 @@ Net Net::Impl::createNetworkFromModelOptimizer(InferenceEngine::CNNNetwork& ieNe
 {
     CV_TRACE_FUNCTION();
 
+    CV_TRACE_REGION("register_inputs");
+
     std::vector<String> inputsNames;
     std::vector<MatShape> inp_shapes;
     for (auto& it : ieNet.getInputsInfo())
@@ -3174,6 +3176,8 @@ Net Net::Impl::createNetworkFromModelOptimizer(InferenceEngine::CNNNetwork& ieNe
     {
         cvNet.setInputShape(inputsNames[inp_id], inp_shapes[inp_id]);
     }
+
+    CV_TRACE_REGION_NEXT("backendNode");
 
     Ptr<BackendNode> backendNode;
 #ifdef HAVE_DNN_NGRAPH
@@ -3195,8 +3199,26 @@ Net Net::Impl::createNetworkFromModelOptimizer(InferenceEngine::CNNNetwork& ieNe
         CV_Error(Error::StsNotImplemented, "This OpenCV version is built without Inference Engine NN Builder API support");
 #endif
     }
+
+    CV_TRACE_REGION_NEXT("register_outputs");
+
+#ifdef HAVE_DNN_NGRAPH
+    auto ngraphFunction = ieNet.getFunction();
+#if INF_ENGINE_VER_MAJOR_LT(INF_ENGINE_RELEASE_2020_2)
+    std::list< std::shared_ptr<ngraph::Node> > ngraphOperations;
+#else
+    std::vector< std::shared_ptr<ngraph::Node> > ngraphOperations;
+#endif
+    if (ngraphFunction)
+    {
+        ngraphOperations = ngraphFunction->get_ops();
+    }
+#endif
+
     for (auto& it : ieNet.getOutputsInfo())
     {
+        CV_TRACE_REGION("output");
+
         LayerParams lp;
         int lid = cvNet.addLayer(it.first, "", lp);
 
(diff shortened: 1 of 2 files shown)


modules/dnn/src/dnn.cpp
```diff
@@ -3205,15 +3227,51 @@ Net Net::Impl::createNetworkFromModelOptimizer(InferenceEngine::CNNNetwork& ieNe
...
+                            {
+                                int num_classes = detOutput->get_attrs().num_classes;
+                                ld.params.set("num_classes", num_classes);
+                                CV_LOG_INFO(NULL, "DNN/IE: num_classes=" << num_classes << " (" << outputName << " @ " << typeName << ")");
+                            }
+                            else
+                            {
+                                CV_LOG_WARNING(NULL, "DNN/IE: Can't extract layer information: '" << outputName << "'");
+                            }
+                        }
```
alalek, line 3258: @dkurt Do we need something like this? (extraction of some parameters)

dkurt:
But why only some of the models are failed?

alalek:
IRv10 + VPU plugin 2020.2

dkurt:
@alalek, I think that the problem in the deprecated [getLayerByName](https://docs.openvinotoolkit.org/latest/classInferenceEngine_1_1ICNNNetwork.html#a9661a132ab8eabe950d41ecd53221c0a) which converts to old representation. May I ask to try https://github.com/dkurt/opencv/commit/fa64494e7b59c44371d651f871bb36397fe5b2d7?

modules/dnn/src/dnn.cpp
dkurt: Not sure that we need this. DetectionOutput returns just indices of detected classes. Do we use the number of classes somewhere in the tests?

modules/dnn/src/dnn.cpp
```diff
@@ -3205,15 +3227,51 @@ Net Net::Impl::createNetworkFromModelOptimizer(InferenceEngine::CNNNetwork& ieNe
...
+            cvLayer->name = outputName;
+            cvLayer->type = "_unknown_";
 
-            InferenceEngine::CNNLayerPtr ieLayer = ieNet.getLayerByName(it.first.c_str());
-            CV_Assert(ieLayer);
+            if (ngraphFunction)
+            {
+                CV_TRACE_REGION("ngraph_function");
+                bool found = false;
+                for (const auto op : ngraphOperations)
```
dkurt, line 3239: Can we remove this variable? Just
```cpp
for (const auto& op : ngraphFunction->get_ops())
```

alalek: `get_ops()` is not trivial. So it is better to cache its result while using in a loop body.

l-bat: Can we cache this right before loop?
```
auto ngraphOperations = ngraphFunction->get_ops();
for (const auto op : ngraphOperations)
```

dkurt: Probably, we have to change it here as well: https://github.com/opencv/opencv/blob/bfcc136dc7bd2e246a67be0a39e2922aacc85b4c/modules/dnn/src/ie_ngraph.cpp#L596?

alalek:
```
auto ngraphOperations = ngraphFunction->get_ops();
for (const auto op : ngraphOperations)
```

is the same as

```
for (const auto op : ngraphFunction->get_ops())
```

---

Currently it is cached before this parent loop:
```
for (auto& it : ieNet.getOutputsInfo())
```

---

> we have to change it here as well:

There is no parent loop in that case. Caching can't help in that case.

modules/dnn/src/dnn.cpp
alalek: Unused functionality is removed.

modules/dnn/src/dnn.cpp
dkurt: Now I see it, thanks!

dkurt: One more question here is shall we use
```
for (const auto& op : ngraphOperations)
```
instead of
```cpp
for (const auto op : ngraphOperations)
```
?

alalek: Replaced to `const auto&`.

PR #17145, opened 2020-04-24
Added initial version of cmake toolchain for RISC-V architecture

asmorkalov:


asmorkalov:
@alalek I removed the controversial lines. It looks like the same should be done for ARM toolchain files too.

asmorkalov:
@vpisarev Do you have any comments on the PR?

vpisarev:
@asmorkalov, good start! :+1:

PR #17170, opened 2020-04-28
TBB executor for GAPI

anton-potapov:
### This Merge Request adds and opt in TBB based executor to GAPI ###
- opt in for now (default off)
- only non-streaming mode (for now)
- The executor uses priority queue based logic for ready to execution tasks

## TBB executor description
TBB executor engine operate with graph of tasks represented via `cv::gimpl::parallel::tile_node`  instances.   

### Dependencies 
For every task (`tile_node` ) there might be a number of input dependencies (i.e. tasks that should be executed before). This number is represented twice in the  `tile_node`  structure. They are `dependencies`  and `dependency_count`. First one describes total number (i.e. input edges in the graph before execution) and does not change during  execution, while second one describes current number of unresolved input dependencies. Once `dependency_count` drops to zero, task is ready for execution as all it's dependencies are satisfied  

### Dependents 
Every task that directly depends on the current one is listed in the `tile_node::dependees` member. The engine decrees  `dependency_count` of these tasks right after the completion of the current task.

### Sync Vs. Async payload 
Task's body in enclosed in either `task_body` or `async_task_body`  members depending on task's nature (and `async ` flag). While in both cases the engine calls passed in function there are two differences:    
 - async one accept the callback (from the engine), which should be called when asynchronous  activity started by this function completes 
- While for synchronous payloads engine treat the task completed after the function returns, for asynchronous ones  task treated completed once the callback is called. 

### Execution and Priorities
Once all input dependencies for the task are satisfied (i.e. it's `dependency_count` drops to zero) it is treated as ready for execution and is put into the "ready" queue.  The queue (`tbb::concurrent_priority_queue`) sorts all task according to their priorities.   

### Mapping to TBB
TBB uses `tbb::task` derivatives to describe a piece of work. They can be linked together into a tree describing tasks dependencies. Every TBB task has a `refcount` field describing unsatisfied dependencies. Once it drops to zero task is ready for execution (i.e. can be `spawn`ed). Dependency tasks are called child tasks (i.e. parent depends on it's childs). Every child task on creation increases it's parent's `refcount`. After child task is executed TBB task scheduler automatically decrease parent's `refcount`.   Root of tasks tree is used only to call it's [`tbb::task::wait_for_all`](https://www.threadingbuildingblocks.org/docs/help/reference/task_scheduler/synchronization_task.html) method to wait for the CPU work to complete.  `wait_for_all` waits for  root `refcount` value to become `1`  as a sign that all work is done. 

The executor engine creates a `root` task (derived from `tbb::task`) - a root of the task's tree.  Once new `tile_node` item is pushed to the ready queue, new `root`'s child task is created. Body of every child task does the following

1. Pops ready for execution `tile_node`  instance from the priority queue
2. Execute it's body 
3. Decrease `dependency_count` for every dependent `tile_node` instance (via `tile_node::dependees` member)
4. Per every `tile_node`  with `dependency_count` dropped to zero, extra (`tbb`) child task is created. 

### Asynchronicity work and TBB active waiting
While waiting for the  root task, TBB does not put to sleep calling thread. Instead it keep searching for `spawn`ed (or `enqueue`d) TBB tasks to do, assuming it somehow related with the root task. 

How ether, in case of async work (like GPU or non TBB CPU work) there may be no such. In this case calling thread constantly spinning, wasting CPU time.  To overcome this more elaborated scheme then plain calling to `tbb::task::wait_for_all` is needed.   

### tile_node graph creation and re-usage
5. `dependency_count` is reset to `dependencies` value for easier reuse of `tile_node` 

```
force_builders=ARMv7,Custom
build_image:Custom=centos:7
buildworker:Custom=linux-1

```


asmorkalov:
@anton-potapov Friendly reminder. Please take a look on CI failures.

modules/gapi/src/executor/gapi_tbb_executor.cpp
```diff
@@ -0,0 +1,355 @@
...
+   for (size_t i=0; i<count; i++)
+   {
+       spawn_no_assert(root, body);
+   }
+}
+
+
+}}}  // namespace cv::gimpl::parallel
+
+void cv::gimpl::parallel::execute(tbb::concurrent_priority_queue<tile_node* , tile_node_indirect_priority_comparator> & q)
```
dmatveev, line 108: Please reformat this code to keep line width limit. 80 is desired, ~110 is maximum.

modules/gapi/src/executor/gapi_tbb_executor.cpp
```diff
@@ -0,0 +1,355 @@
...
+
+    //Specify tbb::task_group_context::concurrent_wait in the traits to ask TBB scheduler not to change ref_count of the task we wait on (root) when wait is complete.
+    //As the traits is last argument explicitly specify (default) value for first argument
+    tbb::task_group_context m_context = {tbb::task_group_context::bound, tbb::task_group_context::default_traits | tbb::task_group_context::concurrent_wait};
+
+
+    root.reset(new (tbb::task::allocate_root(m_context)) tbb::empty_task);
+    root->set_ref_count(1); //required by wait_for_all, as it waits until counter drops to 1
+
+#if 0
```
dmatveev, line 123: Do we need this here?

modules/gapi/src/executor/gapi_tbb_executor.cpp
```diff
@@ -0,0 +1,355 @@
...
+                 return r < counter_limits_t::max() && !counter_limits_t::is_signed;
+             };
+             ASSERT(is_not_wrapped_around(active_async_tasks));
+             //TODO: checks performance gains of minimizing number of call to notify (i.e. only if this is the last one async node or there are new tbb task to execute)
+             if ((active_async_tasks == 0) || (wake_master == wake_tbb_master::yes) )//was the last or there is the new TBB tasks to execute
+             {
+                 ITT_AUTO_TRACE_GUARD(ittTbbUnlockMasterThread);
+                //Wile decrement of async_tasks_t::count is atomic it might be done after waiting thread checked it value but _before_ it actually start waiting on the condition variable.
+                //So, lock acquire is needed to guarantee that current condition check (if any) in waiting thread (possibly ran in parallel to async_tasks_t::count decrement above) is completed _before_
+                //signal is issued. Therefore when notify_one is called, waiting thread is either sleeping on the condition variable or running a new check which is guaranteed to pick the new value and return
```
dmatveev, line 194: Noooooooooooooo waaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaay

modules/gapi/src/executor/gapi_tbb_executor.hpp
```diff
@@ -0,0 +1,81 @@
...
+};
+
+struct async_tag {};
+constexpr async_tag async;
+
+struct tile_node {
+    //place in totally ordered queue of tasks to execute. Inverse to priority, i.e. lower index means higher priority
+    size_t                                                                              total_order_index = 0;
+    std::function<void()>                                                               task_body;
+    std::function<void(std::function<void()> callback, size_t total_order_index)>       async_task_body;
```
dmatveev, line 44: I'd give these types names like we do in the rest of our code.

modules/gapi/src/executor/gexecutor.cpp
```diff
@@ -228,14 +300,16 @@ void cv::gimpl::GExecutor::run(cv::gimpl::GRuntimeArgs &&args)
...
     {
-        // (5)
-        Input i{m_res, op.in_objects};
-        Output o{m_res, op.out_objects};
-        op.isl_exec->run(i, o);
+        // (5) and  (6)
+        run_op(op, m_res);
     }
-
+#endif
```
dmatveev, line 312: Instead of modifying the existing class, can we just put a new one on side? Then there's no changes, just additions to.

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
```diff
@@ -166,7 +166,7 @@ TEST(StatefulKernel, InvalidReallocatingKernel)
     const auto pkg = cv::gapi::kernels<GOCVStInvalidResize>();
     cv::GComputation comp(cv::GIn(in), cv::GOut(out));
 
-    EXPECT_THROW(comp.apply(in_mat, out_mat, cv::compile_args(pkg)), std::logic_error);
+    GAPI_EXPECT_THROW(comp.apply(in_mat, out_mat, cv::compile_args(pkg)), std::logic_error);
```
dmatveev, line 169: But why?

modules/gapi/test/test_precomp.hpp
```diff
@@ -34,4 +34,15 @@ static inline void countNonZero_is_forbidden_in_tests_use_norm_instead() {}
...
+namespace opencv_test {
+    bool isExactExceptionPropogationEnabled();
+}
+
+#define GAPI_EXPECT_THROW(expr, expected_exception)          \
+if (opencv_test::isExactExceptionPropogationEnabled()) {     \
+    EXPECT_THROW(expr, expected_exception);                  \
+} else {                                                     \
+    EXPECT_ANY_THROW(expr);                                  \
+}                                                            \
```
dmatveev, line 46: This shouldn't be in precomp, I believe

modules/gapi/src/executor/gapi_tbb_executor.cpp
```diff
@@ -0,0 +1,355 @@
...
+               }
+            }
+            while(!tbb_work_done() || !async_work_done());
+
+            ASSERT(tbb_work_done() && async_work_done() && "Graph is still running?");
+//        }
+//    );
+
+    LOG_INFO(NULL, "Done. Executed " <<executed<<" tasks");
+}
```
dmatveev, line 354: 200-LOC functions shouldn't exist.

asmorkalov:
@anton-potapov Friendly reminder. How does the PR correlates with https://github.com/opencv/opencv/pull/17851 ?

dmatveev:
@anton-potapov  probably we should close this one?

asmorkalov:
@anton-potapov Friendly reminder.

asmorkalov:
@anton-potapov Friendly reminder.

anton-potapov:
> 
> 
> @anton-potapov Friendly reminder. How does the PR correlates with #17851 ?

This patch was split into parts to speed up integration.  #17851 is the first part - the executor itself. The actual usage it the next part.

So this pull request can be closed.

PR #17196, opened 2020-05-01
core(MatExpr): fix warning in case of e.s == (0, 0, 0, 0)

alalek:
relates #16948

Test case: `fisheyeTest.jacobians`


alalek:
:+1:

PR #17200, opened 2020-05-01
cuda4dnn: optimizations for swish, mish, sigmoid, region, resize based ops, transpose, identity-conv fusion

YashasSamaga:

**Mish (+ swish and sigmoid):**

This activation is used in YOLOv4. Surprisingly, the mish kernels are compute-bound instead of being bandwidth-bound. The first convolution in YOLOv4 takes less time than the fused bias mish activation step. 

Mish is a composition of several functions: tanh(log(1 + exp(x))). Hence, it is very likely that there is a very good approximation or simplification to the mish function (clearly the log isn't necessary since tanh is a composition over e^x). This PR introduces a fast numerically stable implementation of the mish activation.

Details of the approximation can be found [here](https://cs.stackexchange.com/a/125052/66162).

Operation                                              | without this PR | with this PR
--------------------------------------------------- | -------------------- | -----------------
biasN_mish (YOLOv4 conv_1)              | 1.35ms             | 980us
biasN_mish (YOLOv4 conv_2)             | 742us                | 496us

Based on similar reasoning, good fast approximations to swish and sigmoid were also added.

**Identity is fused with convolution layer:**

Activations would not be fused with convolution if an identity layer is present between them. This happens in the case of Mask RCNN: the final sigmoid operation is not fused with the preceding convolution operation.

Operation                                              | without fusion   | with fusion
--------------------------------------------------- | -------------------- | ---------------------------------
biasN_sigmoid (Mask RCNN)               |               368us | 206us

This kernel is compute-bound if the sigmoid approximation mentioned above is not used. The timings reported here are with the approximation.

**Others:**

Operation                                              | without this PR | with this PR
--------------------------------------------------- | -------------------- | -----------------
transpose (YOLOv4 permute1)             | 195us              | 154us
region_box                                            | 62us                  | 56us
region_sigmoid_class_score                | 100.4us            | 72us 
crop_and_resize                                   | 2.98ms             | 2.72ms
resize_bilinear                                      | 1.77ms             | 1.65ms 

<cut/>

```
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/dnn/src/cuda/functors.hpp
+++ b/modules/dnn/src/cuda/functors.hpp
@@ -40,16 +40,28 @@ struct swish_functor {
 template <class T>
 struct mish_functor {
     __device__ T operator()(T value) {
+        using csl::device::tanh;
+        using csl::device::log1pexp;
+        return value * tanh(log1pexp(value));
+    }
+};
+
+template <>
+struct mish_functor<float> {
+    __device__ float operator()(float value) {
         // f(x) = x * tanh(log1pexp(x));
         using csl::device::fast_divide;
         using csl::device::fast_exp;
 
         auto e = fast_exp(value);
-        if (value <= static_cast<T>(-8))
+        if (value <= -18.0f)
             return value * e;
 
-        auto p = static_cast<T>(1) + e;
-        return value - static_cast<T>(2) * fast_divide(value, p * p + static_cast<T>(1));
+        auto n = e * e + 2 * e;
+        if (value <= -5.0f)
+            return value * fast_divide(n, n + 2);
+
+        return value - 2 * fast_divide(value, n + 2);
     }
 };
 
--- a/modules/dnn/src/cuda/functors.hpp
+++ b/modules/dnn/src/cuda/functors.hpp
@@ -30,25 +30,34 @@ struct tanh_functor {
 template <class T>
 struct swish_functor {
     __device__ T operator()(T value) {
-        using csl::device::sigmoid;
-        return value * sigmoid(value);
+        // f(x) = x * sigmoid(x)
+        using csl::device::fast_divide;
+        using csl::device::fast_exp;
+        return fast_divide(value, static_cast<T>(1) + fast_exp(-value));
     }
 };
 
 template <class T>
 struct mish_functor {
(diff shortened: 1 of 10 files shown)


modules/dnn/src/cuda/functors.hpp
```diff
@@ -44,11 +46,30 @@ struct mish_functor {
...
+    __device__ float operator()(float value) {
+        // f(x) = x * tanh(log1pexp(x));
+        using csl::device::fast_divide;
+        using csl::device::fast_exp;
+
+        auto e = fast_exp(value);
+        if (value <= -18.0f)
+            return value * e;
+
+        auto n = e * e + 2 * e;
```
tompollok, line 60: n = e(e+2)

YashasSamaga: `e` is `exp(value)`. For negative values below some threshold, `e` becomes really small that adding it to `2` will give back `2`. This will cause you to lose significant digits. `e * e` is close to `2 * e` in the small negative numbers range `[-1, -15]` and hence you won't suffer from the same problem.

While this is algebraically same, it's not the same when working with finite-precision numbers.

tompollok: Oh your comment disappeared somehow. I agree that what you say makes sense. So do you think that creating specific taylor series function for some functions like mish would even more boost speed?

YashasSamaga: I deleted it because it was wrong. `e * e + 2 * e` and `e * (e + 2)` are different but the error  was only a few ULPs in the final expression and it happened at the very edge. I have to do more rigirous checks. It's negligible improvement though.

I don't think Taylor series would help. It would still involve `tanh`, `log`, `exp`, etc. It's possible to make a rational approximation to mish but that computationally more expensive than what's currently in the PR. And it improves the accuracy negligibly, so it's not worth it.

YashasSamaga: `e * e + 2 * e` leads to `FMA(e, e, FADD(e, e))`
`e * (e + 2)` leads to `FMUL(e, FADD(e + 2))`

PR #17223, opened 2020-05-05
Update table of content for imgproc module tutorials to reflect languages used

dmallia17:
This adds the Languages field for the "Random generator and text with OpenCV" tutorial and updates the Languages field for the "Anisotropic image segmentation by a gradient structure tensor" tutorial.

On a related note: I have not been able to locate an equivalent for the C++ RNG class in Python, but I believe this functionality can be achieved with NumPy (among other options). If I have missed the equivalent or if NumPy is an acceptable alternative, I would very much like to try expanding the "Random generator and text with OpenCV" tutorial to include Python. Thank you!

### DIFF
--- a/doc/tutorials/imgproc/table_of_content_imgproc.markdown
+++ b/doc/tutorials/imgproc/table_of_content_imgproc.markdown
@@ -15,6 +15,8 @@ In this section you will learn about the image processing (manipulation) functio
 
 -   @subpage tutorial_random_generator_and_text
 
+    *Languages:* C++
+
     *Compatibility:* \> OpenCV 2.0
 
     *Author:* Ana Huamán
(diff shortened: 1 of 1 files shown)


asmorkalov:
@dmallia17 Functions like `theRNG` that returns default OpenCV random generator are not wrapped into Python bindings and you need to use alternatives. Numpy `np.random` is good choice. Some OpenCV algorithms are stochastic by design and use OpenCV RNG implementation internally. You still can manage their parameters in Python with calls like `setRNGSeed`.

asmorkalov:
@dmallia17 Could you squash the commits to have clean library history?

dmallia17:
@asmorkalov I believe that should do it - please correct me if I'm wrong. (First time having to squash commits for a PR!)

Also thank you very much for the explanation! I did spot `setRNGSeed` but wasn't sure of its usage. I am hopeful that I will be able to tackle expanding that tutorial very soon.

dmallia17:
@asmorkalov Thank you! After looking around a bit further, I have found a couple of other tutorial modules where the table of contents should be updated. Would it be best to make a pull request for each table of contents or update all of the ones I spot at once?

alalek:
@dmallia17 Thank you for contribution! Feel free to create single Pull Request with similar changes.

dmallia17:
@alalek  Thank you very much - will do!

PR #17224, opened 2020-05-06
Fixed indexing in prefilter

### LINKED ISSUE #17201: Possible code error in stereobm.cpp in calib3d module
Not sure what's the protocol for reporting possible code errors, but in stereobm.cpp line 382:
  area.allocate(prefilter[0], width + params.preFilterSize + 2);
which is inside a for loop:
 for (size_t i = 0; i < 2; ++i)
Shouldn't prefilter[0] be iterated over i : prefilter[i]  ?

ganesh-k13:
### Pull Request Changes:

1. Fixes Possible code error in stereobm.cpp in calib3d module
2. Added UT for BufferBM usage in StereoBM

resolves #17201

### DIFF
--- a/modules/calib3d/src/stereobm.cpp
+++ b/modules/calib3d/src/stereobm.cpp
@@ -348,7 +348,7 @@ public:
         cbuf0(nstripes, NULL),
         sad_short(nstripes, NULL),
         hsad_short(nstripes, NULL),
-        prefilter{ 0, 0 }
+        prefilter()
     {
         const int wsz = params.SADWindowSize;
         const int ndisp = params.numDisparities;
--- a/modules/calib3d/src/stereobm.cpp
+++ b/modules/calib3d/src/stereobm.cpp
@@ -347,7 +347,8 @@ public:
         htext(nstripes, NULL),
         cbuf0(nstripes, NULL),
         sad_short(nstripes, NULL),
-        hsad_short(nstripes, NULL)
+        hsad_short(nstripes, NULL),
+        prefilter{ 0, 0 }
     {
         const int wsz = params.SADWindowSize;
         const int ndisp = params.numDisparities;
--- a/modules/calib3d/src/stereobm.cpp
+++ b/modules/calib3d/src/stereobm.cpp
@@ -379,7 +379,7 @@ public:
         if (params.useNormPrefilter())
         {
             for (size_t i = 0; i < 2; ++i)
-                area.allocate(prefilter[0], width + params.preFilterSize + 2);
+                area.allocate(prefilter[i], width + params.preFilterSize + 2);
         }
         area.commit();
 
--- a/modules/calib3d/test/test_stereomatching.cpp
+++ b/modules/calib3d/test/test_stereomatching.cpp
@@ -809,6 +809,52 @@ protected:
     }
 };
 
+TEST(Calib3d_StereoBM, regression) { CV_StereoBMTest test; test.safe_run(); }
+
+typedef tuple < int, int, int > BufferBM_Params_t;
+
+typedef testing::TestWithParam< BufferBM_Params_t > Calib3d_StereoBM_BufferBM;
+
+const int preFilters[] =
+{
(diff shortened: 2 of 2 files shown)


ganesh-k13:
@alalek, is a unit test required to catch this?

asmorkalov:
@ganesh-k13 Thanks for the patch! Yes, unit test is very appreciated.

ganesh-k13:
Sure @asmorkalov, I'll add it. I'm new to this, so I'll read a bit before adding.

ganesh-k13:
Hi @mshabunin, I have a basic doubt. 
I tried the fix in a sample file:
```
int main() {
     Mat leftImg = imread("im2.png", 0);
     Mat rightImg = imread("im2.png", 0);
     Mat leftDisp;
     Ptr<StereoBM> bm = StereoBM::create(16,9);
     bm->setPreFilterType(StereoBM::PREFILTER_NORMALIZED_RESPONSE);
     bm->compute( leftImg, rightImg, leftDisp);
     return 0;
 }
```

I see no crash. 
But if I do same in a Test(`calib3d/test/test_stereomatching.cpp `):
```
TEST(Calib3d_StereoBM, regression17201)
 {
     String path = cvtest::TS::ptr()->get_data_path() + "cv/stereomatching/datasets/teddy/";
     Mat leftImg = imread(path + "im2.png", 0);
     ASSERT_FALSE(leftImg.empty());
     Mat rightImg = imread(path + "im2.png", 0);
     ASSERT_FALSE(rightImg.empty());
     Mat leftDisp;
     {
         Ptr<StereoBM> bm = StereoBM::create(16, 9);
         bm->setPreFilterType(StereoBM::PREFILTER_NORMALIZED_RESPONSE);
         bm->compute( leftImg, rightImg, leftDisp);
     }
 }
```
`[1]    17033 segmentation fault (core dumped)  bin/opencv_test_calib3d --gtest_filter=Calib3d_StereoBM.regression17201` almost as if change is not there.

Is there anything for this difference in behavior between normal code and gtest code?

asmorkalov:
@ganesh-k13 could you put your sample code to OpenCV samples and build OpenCV with `-DBUILD_EXAMPLES=ON`. All samples are built with the same compiler options and it should help to reproduce the issue.

ganesh-k13:
Thanks @asmorkalov, I'll try that, I think it is some small config mistake from my end. I'll try some more stuff or post in answers.opencv.

mshabunin:
@ganesh-k13 , prefilter pointers must be initialized with 0, add something like
```.cpp
prefilter[0] = 0;
prefilter[1] = 0;
```
at the beginning of the `BufferBM` constructor.

ganesh-k13:
Thanks, @mshabunin, that fixed the issue. 
I had a doubt regarding:https://github.com/opencv/opencv/blob/09799402f912dc4c08b759ff628842cd160756ce/modules/calib3d/test/test_stereomatching.cpp#L795-L798

How was the value of 0.005+DBL_EPSILON arrived at?

I was planning to add ` bm->setPreFilterType(StereoBM::PREFILTER_NORMALIZED_RESPONSE);` before compute there(or once again after PREFILTER_XSOBEL is used). Is this ok or I'll add a separate test?

mshabunin:
Looks like this tests does not verify the output, it only checks that integer output does not differ from float output and this error margin is an arbitrary value.

Yes, I think you can add another filter to the same test, or you can make [parameterized test](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#value-parameterized-tests) (`TEST_P(...)`).

ganesh-k13:
Thanks, @mshabunin , I have added cases, the aim of the cases is to trigger different mem allocs in BufferBM. So I added cases with combinations of `preFilterCap`, `SADWindowSize` and `preFilterType`. I also moved Calib3d_StereoBM, regression up because it was under Calib3d_StereoSBM

ganesh-k13:
@alalek , @mshabunin , @asmorkalov any update on this?

PR #17228, opened 2020-05-06
CAP_MFX: surface pool with timeout, more parameters

mshabunin:
Added environment variables to control MFX videoio backend:
- `OPENCV_VIDEOIO_MFX_IMPL` - preferred implementation, number, see mfxIMPL documentation
- `OPENCV_VIDEOIO_MFX_POOL_TIMEOUT` - surface pool wait timeout, seconds, default value is 1
- `OPENCV_VIDEOIO_MFX_EXTRA_SURFACE_NUM` - extra surfaces in surface pool, number, default value is 0
- `OPENCV_VIDEOIO_MFX_BITRATE_DIVISOR` - bitrate calculation divisor, number, default value is 500, `bitrate = pixel_num * fps/ divisor`
- `OPENCV_VIDEOIO_MFX_WRITER_TIMEOUT` - encoder timeout, seconds, default value is 1

### DIFF
--- a/modules/videoio/src/cap_mfx_common.cpp
+++ b/modules/videoio/src/cap_mfx_common.cpp
@@ -14,10 +14,30 @@
 using namespace std;
 using namespace cv;
 
+static mfxIMPL getImpl()
+{
+    static const size_t res = utils::getConfigurationParameterSizeT("OPENCV_VIDEOIO_MFX_IMPL", MFX_IMPL_AUTO_ANY);
+    return (mfxIMPL)res;
+}
+
+static size_t getExtraSurfaceNum()
+{
+    static const size_t res = cv::utils::getConfigurationParameterSizeT("OPENCV_VIDEOIO_MFX_EXTRA_SURFACE_NUM", 1);
+    return res;
+}
+
+static size_t getPoolTimeoutSec()
+{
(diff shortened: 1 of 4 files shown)


modules/videoio/src/cap_mfx_common.cpp
```diff
@@ -14,10 +14,12 @@
 using namespace std;
 using namespace cv;
 
+static size_t IMPL = utils::getConfigurationParameterSizeT("OPENCV_VIDEOIO_MFX_IMPL", MFX_IMPL_AUTO_ANY);
```
alalek, line 17: Lets avoid exploding of global static initialization tables.
Use lazy on-demand initialization instead. See `static const String getBuildExtraOptions()` from ocl.cpp

mshabunin: Should I repeat scheme with `initialized` flag or use simple static initialization like shown below?
```.cpp
size_t getValue()
{
    static const size_t value = initValue();
    return value;
}
```

alalek: Something like this is enough (sample code prints "overloaded" message):
```
size_t getValue()
{
    static const size_t value = utils::getConfigurationParameterSizeT("OPENCV_VIDEOIO_MFX_IMPL", MFX_IMPL_AUTO_ANY);
    return value;
}
```

PR #17233, opened 2020-05-07
Added ONNX BatchNorm subgraph

l-bat:
**Merge with extra:** opencv/opencv_extra#755

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

image

### DIFF
--- a/modules/dnn/src/onnx/onnx_graph_simplifier.cpp
+++ b/modules/dnn/src/onnx/onnx_graph_simplifier.cpp
@@ -61,27 +61,28 @@ public:
     ONNXGraphWrapper(opencv_onnx::GraphProto& _net) : net(_net)
     {
         numInputs = net.input_size();
+        numInitializers = net.initializer_size();
     }
 
     virtual Ptr<ImportNodeWrapper> getNode(int idx) const CV_OVERRIDE
     {
         opencv_onnx::NodeProto* node = 0;
-        if (idx >= numInputs)
-            node = net.mutable_node(idx - numInputs);
+        if (idx >= numInputs + numInitializers)
+            node = net.mutable_node(idx - numInputs - numInitializers);
         return makePtr<ONNXNodeWrapper>(node);
     }
 
     virtual int getNumNodes() const CV_OVERRIDE
     {
-        return numInputs + net.node_size();
+        return numInputs + numInitializers + net.node_size();
     }
 
     virtual int getNumOutputs(int nodeId) const CV_OVERRIDE
     {
-        if (nodeId < numInputs)
+        if (nodeId < numInputs + numInitializers)
             return 1;
         else
-            return net.node(nodeId - numInputs).output_size();
+            return net.node(nodeId - numInputs - numInitializers).output_size();
     }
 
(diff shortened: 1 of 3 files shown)


modules/dnn/src/onnx/onnx_importer.cpp
```diff
@@ -378,6 +377,22 @@ void ONNXImporter::populateNet(Net dstNet)
...
+        std::string name = graph_proto.input(idx).name();
+        if (std::find(netInputs.begin(), netInputs.end(), name) == netInputs.end())
+        {
+            graph_proto.mutable_input()->DeleteSubrange(idx, 1);
+        }
+        else
+        {
+            ++idx;
+        }
+    }
```
dkurt, line 393: Why not to remove only `Initializer` nodes at `addConstantNodesForInitializers`?

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
```diff
@@ -382,33 +382,63 @@ class ResizeSubgraph2 : public ExtractScalesSubgraph
...
+class BatchNormalizationSubgraph1 : public BatchNormalizationSubgraphBase
+{
+public:
+    BatchNormalizationSubgraph1()
+    {
+        int reshape1 = addNodeToMatch("Reshape", weight, shape1);
+        int reshape2 = addNodeToMatch("Reshape", bias, shape2);
         int shape3 = addNodeToMatch("Constant");
-        int reshape3 = addNodeToMatch("Reshape", data3, shape3);
+        int reshape3 = addNodeToMatch("Reshape", var, shape3);
```
dkurt, line 411: I think it's simpler to have a boolean variable which indicates if there are Reshape over contant nodes or not

modules/dnn/src/onnx/onnx_importer.cpp
l-bat: `Initializers `overlap with inputs (not constant nodes). I think  removing `Initializers` is useless because we don't use initializers in `simplifySubgraphs`, only inputs and nodes.

dkurt: Because we have a procedure which replaces Initializers to Constants. That means other code which is related to this procedure should be also located there.

modules/dnn/src/onnx/onnx_graph_simplifier.cpp
l-bat: In `BatchNormalizationSubgraph1` reshape used only for constant nodes, but in `BatchNormalizationSubgraph2` reshape used after `Mul` and `Sub` operations.

dkurt:
@l-bat, please try to extend https://github.com/dkurt/opencv/commit/08c066a5ea21e53e29a78ce6eecdad19ea8498b2 considering scenario when initializers included in the inputs. This patch is something that @ashishkrshrivastava proposed once but agnostic input nodes (might be any type):
```cpp
int input = addNodeToMatch("");
int data1 = addNodeToMatch("");
int data2 = addNodeToMatch("");
int data3 = addNodeToMatch("");
int data4 = addNodeToMatch("");
```

Issue #16306, opened 2020-01-08
dnn(permute): reduce permute to reshape (or transpose) whenever possible and work inplace


YashasSamaga:
##### System information (version)
- OpenCV => 4.2.0
- Operating System / Platform => Ubuntu 18.04
- Compiler => GCC 7.4

##### Detailed description

Vast majority of permute operations can be reduced to a reshape or a 2d-transpose. The reshape operation can be skipped entirely and the transpose operation can be efficiently performed inplace.

**Stats for single image inference:**

Model                 | Total Permute Layers | Reduced to transpose | Reduced to reshape
---------------------- | ---------------------------- | ----------------------------- | ---------------------------
MobileNet SSD  | 12                               | 10                                | 2
YOLOv3             | 3                                  | 3                                  | 0
Inception v2 Faster RCNN | 3                   | 2                                 | 1
Inception v2 Mask RCNN | 3                   | 2                                 | 1

Currently, the permute layer operates inplace when the permute order is identity order. This logic can be upgraded to include the logic used here:
1. https://github.com/opencv/opencv/blob/1f2b2c52422813188bc5cbcc5d312c89e41786b2/modules/dnn/src/cuda/permute.cu#L173-L186
2. https://github.com/opencv/opencv/blob/1f2b2c52422813188bc5cbcc5d312c89e41786b2/modules/dnn/src/cuda/permute.cu#L217-L231

This logic would allow the permute operation to be skipped entirely when it reduces to a reshape and perform inplace when it reduces to a 2d-transpose.

The reshape optimization is easier to implement. The logic can be implemented in `getMemoryShapes`. The same logic can be implemented in `op_permute.cpp` for the Vulkan backend. It's fairly easy to implement the same in the CUDA backend. IE nodes created in `BlankLayer` can be used when permute reduces to a reshape (or maybe IE's PermuteLayer/TransposeOp could handle it on its own).

The inplace transpose optimization (as all backends may not support it) would require inplace to be triggered for specific backends (which I believe the current DNN module doesn't support).

tompollok:
Whats the expected performance gain?

YashasSamaga:
@tompollok Not much. The CUDA backend already reduces permute to 2d-transpose and reshape (which in turn leads to raw memory copy). If other backends support the optimization, it will allow the permute layers which reduce to reshape to be fully skipped and permutes that reduce to a transpose to be performed inplace.

YashasSamaga:
[...]

PR #17236, opened 2020-05-07
dnn: Implement graph (gapi) version of forward functions

Yougmark:
Unroll the for loop of layers' forward function calls, using the gapi
support for lambda functions provided in the CPU backend.

Add a test that retrieves per-layer results.

### DIFF
--- a/modules/dnn/src/dnn.cpp
+++ b/modules/dnn/src/dnn.cpp
@@ -64,6 +64,9 @@
 
 #include <opencv2/core/cuda.hpp>
 
+#include <opencv2/gapi/core.hpp>
+#include <opencv2/gapi/cpu/gcpukernel.hpp>
+
 namespace cv {
 namespace dnn {
 CV__DNN_INLINE_NS_BEGIN
@@ -1140,6 +1143,14 @@ static Ptr<BackendWrapper> wrapMat(int backendId, int targetId, cv::Mat& m)
 }
 
 static int g_networkId = 0;
+
+G_TYPED_KERNEL(GLayer, <GMat(GMat, cv::Ptr<LayerData>)>, "org.opencv.dnn.layer")
+{
+    static GMatDesc outMeta(GMatDesc in, cv::Ptr<LayerData> )
+    {
+        return in;
+    }
+};
 
 struct Net::Impl
 {
@@ -3164,6 +3175,57 @@ struct Net::Impl
         ld.flag = 1;
     }
 
+    void forwardToLayerGraph(LayerData &lastld, bool clearFlags = true)
+    {
+        CV_TRACE_FUNCTION();
+
+        if (clearFlags)
+        {
+            MapIdToLayerData::iterator it;
+            for (it = layers.begin(); it != layers.end(); it++)
+                it->second.flag = 0;
+        }
+
+        //already was forwarded
+        if (lastld.flag)
+            return;
+
+        // G-API code
+        std::vector<cv::GMat> outs(layers.size() + 1);
+        MapIdToLayerData::iterator it = layers.begin();
+        int i = 0;
+        while (1)
+        {
+            LayerData &ld = it->second;
+            outs[i + 1] = GLayer::on(outs[i], makePtr<LayerData>(ld));
+            it++;
+            if (it == layers.end() || (it->second.id > lastld.id))
+                break;
+            i++;
+        }
+        cv::GComputation comp(outs[0], outs[i + 1]);
+
+        // a trivial Mat is used here because empty Mat is not accepted by GAPI;
+        // empty Mat will trigger exception when creating internal mat.
+        cv::Mat in = Mat({1, 1, 1, 1}, CV_32F);
+        cv::Mat out;
+        auto impl = cv::gapi::cpu::ocv_kernel<GLayer>(
+                    [=](const cv::Mat& in,
+                            cv::Ptr<LayerData> ld,
+                            cv::Mat& out)
+                    {
+                        if (!ld->flag)
+                            this->forwardLayer(*ld);
+                    });
+        auto pkg = cv::gapi::kernels(impl);
+        comp.apply(in, out, cv::compile_args(pkg));
+
+#ifdef HAVE_CUDA
+        if (preferableBackend == DNN_BACKEND_CUDA)
+            cudaInfo->context.stream.synchronize();
+#endif
+    }
+
     void forwardToLayer(LayerData &ld, bool clearFlags = true)
     {
         CV_TRACE_FUNCTION();
@@ -3836,6 +3898,33 @@ void Net::forward(OutputArrayOfArrays outputBlobs, const String& outputName)
     }
 }
 
+void Net::forwardGraph(OutputArrayOfArrays outputBlobs,
+                       const std::vector<String>& outBlobNames)
+{
+    CV_TRACE_FUNCTION();
+
+    std::vector<LayerPin> pins;
+    for (int i = 0; i < outBlobNames.size(); i++)
+    {
+        pins.push_back(impl->getPinByAlias(outBlobNames[i]));
+    }
+
+    impl->setUpNet(pins);
+
+    LayerPin out = impl->getLatestLayerPin(pins);
+
+    impl->forwardToLayerGraph(impl->getLayerData(out.lid));
+
+    std::vector<Mat> matvec;
+    for (int i = 0; i < pins.size(); i++)
+    {
+        matvec.push_back(impl->getBlob(pins[i]));
+    }
+
+    std::vector<Mat> & outputvec = *(std::vector<Mat> *)outputBlobs.getObj();
+    outputvec = matvec;
+}
+
 void Net::forward(OutputArrayOfArrays outputBlobs,
                   const std::vector<String>& outBlobNames)
 {
@@ -3861,6 +3950,35 @@ void Net::forward(OutputArrayOfArrays outputBlobs,
 
     std::vector<Mat> & outputvec = *(std::vector<Mat> *)outputBlobs.getObj();
     outputvec = matvec;
+}
+
+void Net::forwardGraph(std::vector<std::vector<Mat> >& outputBlobs,
+                       const std::vector<String>& outBlobNames)
+{
+    CV_TRACE_FUNCTION();
+
+    std::vector<LayerPin> pins;
+    for (int i = 0; i < outBlobNames.size(); i++)
+    {
+        pins.push_back(impl->getPinByAlias(outBlobNames[i]));
+    }
+
+    impl->setUpNet(pins);
+
+    LayerPin out = impl->getLatestLayerPin(pins);
+
+    impl->forwardToLayerGraph(impl->getLayerData(out.lid));
+
+    outputBlobs.resize(outBlobNames.size());
+    for (int i = 0; i < outBlobNames.size(); i++)
+    {
+        std::vector<LayerPin> lp = impl->getLayerOutPins(outBlobNames[i]);
+        outputBlobs[i].resize(lp.size());
+        for (int j = 0; j < lp.size(); j++)
+        {
+            outputBlobs[i][j] = impl->getBlob(lp[j]);
+        }
+    }
 }
 
 void Net::forward(std::vector<std::vector<Mat> >& outputBlobs,
--- a/modules/dnn/CMakeLists.txt
+++ b/modules/dnn/CMakeLists.txt
@@ -10,7 +10,7 @@ set(the_description "Deep neural network module. It allows to load models from d
 
 ocv_add_dispatched_file_force_all("layers/layers_common" AVX AVX2 AVX512_SKX)
 
-ocv_add_module(dnn opencv_core opencv_imgproc WRAP python java js)
+ocv_add_module(dnn opencv_core opencv_imgproc opencv_gapi WRAP python java js)
 
 ocv_option(OPENCV_DNN_OPENCL "Build with OpenCL support" HAVE_OPENCL AND NOT APPLE)
 if(HAVE_TENGINE)
--- a/modules/dnn/include/opencv2/dnn/dnn.hpp
+++ b/modules/dnn/include/opencv2/dnn/dnn.hpp
@@ -521,12 +521,26 @@ CV__DNN_INLINE_NS_BEGIN
          */
         CV_WRAP void forward(OutputArrayOfArrays outputBlobs, const String& outputName = String());
 
+        /** @brief Runs forward pass using GAPI to compute outputs of layers listed in @p outBlobNames.
+         *  @param outputBlobs contains blobs for first outputs of specified layers.
+         *  @param outBlobNames names for layers which outputs are needed to get
+         */
+        CV_WRAP void forwardGraph(OutputArrayOfArrays outputBlobs,
+                                  const std::vector<String>& outBlobNames);
+
         /** @brief Runs forward pass to compute outputs of layers listed in @p outBlobNames.
          *  @param outputBlobs contains blobs for first outputs of specified layers.
          *  @param outBlobNames names for layers which outputs are needed to get
          */
         CV_WRAP void forward(OutputArrayOfArrays outputBlobs,
                              const std::vector<String>& outBlobNames);
+
+        /** @brief Runs forward pass using GAPI to compute outputs of layers listed in @p outBlobNames.
+         *  @param outputBlobs contains all output blobs for each layer specified in @p outBlobNames.
+         *  @param outBlobNames names for layers which outputs are needed to get
+         */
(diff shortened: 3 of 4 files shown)


Yougmark:
```
57>C:\build\precommit_windows64\opencv\modules\dnn\src\dnn.cpp(355): warning C4244: 'argument': conversion from 'double' to 'int', possible loss of data [C:\build\precommit_windows64\build\modules\dnn\opencv_dnn.vcxproj]
```
This warning is not triggerred by my patch.  Thoughts?

dkurt:
Hi!
Can  you please describe which benefits of using G-API we have? Maybe it make sense to make G-API optional and if it's available use always without separate method?

dmatveev:
This is cool :)

modules/dnn/src/dnn.cpp
```diff
@@ -1141,6 +1144,14 @@ static Ptr<BackendWrapper> wrapMat(int backendId, int targetId, cv::Mat& m)
 
 static int g_networkId = 0;
 
+G_TYPED_KERNEL(GLayer, <GMat(GMat, cv::Ptr<LayerData>)>, "org.opencv.dnn.layer")
+{
+    static GMatDesc outMeta(GMatDesc in, cv::Ptr<LayerData> )
+    {
+        return in;
```
dmatveev, line 1151: Not sure if this method is correct...

The goal of `outMeta` is to let G-API know how to allocate the internal buffers to hold the results.

If your `GLayer` operation is a generic one, I assume the output dimensions change for DNN operators thorough the graph. I am wondering how it works now, but it seems you need to update the return value `GMatDesc` based on dimensions of the input (`in`) and details about the layer (`LayerData`)

modules/dnn/src/dnn.cpp
```diff
@@ -3164,6 +3175,57 @@ struct Net::Impl
...
+        int i = 0;
+        while (1)
+        {
+            LayerData &ld = it->second;
+            outs[i + 1] = GLayer::on(outs[i], makePtr<LayerData>(ld));
+            it++;
+            if (it == layers.end() || (it->second.id > lastld.id))
+                break;
+            i++;
+        }
```
dmatveev, line 3205: Is it, hmm, correct?

What I see here is that we're building a linear pipeline of operators, but I assume real DL graphs are more complex than this?

modules/dnn/src/dnn.cpp
```diff
@@ -3164,6 +3175,57 @@ struct Net::Impl
...
+        cv::Mat in = Mat({1, 1, 1, 1}, CV_32F);
+        cv::Mat out;
+        auto impl = cv::gapi::cpu::ocv_kernel<GLayer>(
+                    [=](const cv::Mat& in,
+                            cv::Ptr<LayerData> ld,
+                            cv::Mat& out)
+                    {
+                        if (!ld->flag)
+                            this->forwardLayer(*ld);
+                    });
```
dmatveev, line 3219: 👍 . @TolyaTalamanov must be happy now. :D

Yougmark:
> Hi!
> Can you please describe which benefits of using G-API in dnn we have? Maybe it make sense to make G-API optional and if it's available use always without separate method?

Hi,

That's a good question.  I am thinking about this patch as the first step in supporting graph scheduling of DNN in OpenCV (I'll talk more about this below).  Graph scheduling helps improve throughput performance through pipelining and parallel execution, which I'm not sure if G-API has implemented (I haven't looked into `GStreamingExecutor` yet).  So I think the answer to the first question is that using G-API in DNN has the potential of throughput improvement through graph scheduling.  In particular, this is more applicable to DNNs that do not rely on historical results; for example, object detection DNNs.

I have concerns regarding always using G-API when it's available.   I suppose G-API comes with overheads, compared to the for loop that iterates over layers' forward function calls.  That's why I didn't replace the original `Net::forward` function, besides the intention of making it easier for getting this PR accepted :).

More on supporting graph scheduling for DNN using G-API:
This for-loop-unrolling approach is not my original plan.  I was thinking about adding `dnn` APIs in `gapi`, as with `imgproc`.  I wanted to start with a CUDA backend because I have a NVIDIA GPU.  However, the `CUDABackendNode` in `dnn` module is not exported, thus preventing the usage of it in `gapi` module.  Also, to build a DNN using the existing import functions (such as `readNetFromDarknet`) then causes circular dependencies between `dnn` and `gapi` modules.  I figured this plan (of having a `dnn` API and a CUDA backend in `gapi`) won't work without modifying the `dnn` module by a lot.

Therefore, I switched to this simpler approach as a start point.  I'm happy to hear and would appreciate any feedback from the team.

Thanks,
Ming

modules/dnn/src/dnn.cpp
Yougmark: This is not relying on G-API to pass data between layers, yet.  I agree that it is better implemented that way.  Please see my comments on the PR for further explanation of why I am not able to do that.

This implementation relies on the `Net::Impl` itself to pass the data (or data pointers) using its internal data structures.   This `GLayer` operation only wraps the `forward` function call.  I hope this answer your question.

modules/dnn/src/dnn.cpp
Yougmark: I tested the existing networks in OpenCV by running `opencv_dnn_test`, and it seems okay so far --- tests of all networks pass.  I agree that more complex networks shouldn't be supported by this linear pipeline, but the original `forward` function calls layers' `forward` function as well.  I suppose this `dnn` module hasn't extended to those more complex networks yet.

dkurt:
@Yougmark, thanks! Can you please start with some POC and show the improvement in throughput? You may try to use perf tests engine: https://github.com/opencv/opencv/wiki/HowToWritePerfTests + https://github.com/opencv/opencv/wiki/HowToUsePerfTests

Yougmark:
> @Yougmark, thanks! Can you please start with some POC and show the improvement in throughput? You may try to use perf tests engine: https://github.com/opencv/opencv/wiki/HowToWritePerfTests + https://github.com/opencv/opencv/wiki/HowToUsePerfTests

This patch doesn't improve throughput.  Sorry if my previous reply causes any confusion.

modules/dnn/src/dnn.cpp
YashasSamaga: The DNN module supports complex networks. Since the graphs are an acyclic directed graph, you can always find a topological sorting. This linear pipeline is just a topological sorting of the computational graph.

Yougmark: If cyclic subgraphs are treated as single nodes, this should indeed take care of all kinds of NNs.  I wonder how it is implemented; are layers fused into one?

You are right: in most cases, concurrent kernel execution (CKE) is negligible.  Here I mention a few cases that CKE matters.  High-end GPUs (e.g., TITAN X with 80 SMs) might still be underutilized. Also, when kernels are exiting, part of GPUs can be idle.  In addition, pipelined execution using G-API can enable concurrency between CPU and GPU; for example, the CPU part that fetches new frames can be overlapped with GPU execution, although this isn't about CKE.

CUDA graphs are interesting.  It seems to me that it would be natural to add G-API graph compiler that transforms G-API graphs into CUDA graphs.  However, this is a far-future work to do as you mentioned that cuDNN doesn't support CUDA graphs until maybe 8.0.

Anyway, the pipelined execution of DNN, using G-API, needs further implementation.  I think I will continue the implementation.  This is my first time working on adding features into OpenCV, so any interest or guidance from the team (reviewers @dmatveev and @dkurt) can be big motivation and help for me.

A side note: I experimented graph scheduling with Darknet and have seen promising throughput improvement; that's why I think it worths (and is even better) to implement the combination of DNN+G-API in OpenCV.  However, this patch itself is not enough to show throughput benefits.

dmatveev:
@Yougmark thanks for your extended reply. Please ask questions if you need help implementing your backends with G-API.

Yougmark:
> @Yougmark thanks for your extended reply. Please ask questions if you need help implementing your backends with G-API.

@dmatveev Thanks!  I'm looking into `GStreamingExecutor`.  It seems to support pipelined execution but at a one-island-per-stage granularity.  For now, I still want to use GCPUBackend for decomposing DNNs into G-API graphs (using this patch), and to enable pipelined execution at a one-layer-per-stage (i.e., one-node-per-stage) granularity, I think I need to implement a new executor.  Is this a fine way to move forward?  Or would you suggest a different approach for decomposing DNN graphs or enabling finer-grained pipelined execution?

YashasSamaga:
The discussions in #16306 might be relevant.

modules/dnn/src/dnn.cpp
YashasSamaga: The cycles in RNN can be expended as N repetitive steps. For efficiency purposes, the RNN (and all its derivatives) are more efficiently implemented as a single node.

Currently, the CUDA backend only dumps operations into a stream (a pipeline) in `forwardLayer`, i.e. `forwardLayer` returns immediately after dumping the operations. All the waiting for the pipeline to finish execution happens at the end of the `forwardToLayer`. CUDA runtime optimizes the stream execution by minimizing the launch overhead.

Currently, the pipeline is linear. I have [attempted in the past](https://github.com/YashasSamaga/opencv/tree/cuda4dnn-concurrent) to construct a graph using streams and events and let the CUDA runtime handle the scheduling of this graph. The overheads lead to mostly no improvement or a net loss. My implementation is pathetic and it was an attempt to have a proof of concept prototype. I eventually compared CUDA graphs with a graph implemented using streams and events. The [CUDA graph-based implementation](https://github.com/YashasSamaga/opencv/tree/cuda4dnn-graph) fares better (with proxy kernels for cuDNN) than the manually constructed graph but unfortunately, cuDNN has issues with CUDA graphs.

dmatveev:
@Yougmark I see you're pretty familiar with our framework internals and understand the concept of islands and their execution pretty well. That's cool, I really appreciate it!

Let me summarize it a bit. We have two graph models, a `GModel` and a `GIslandModel`:
* `GModel` is the compiler's primary internal representation of the graph, we use it to validate the graph, understand its structure, assign it to backends, and so on.
* `GIslandModel` is a projection from `GModel` where we fuse `GModel` nodes into clusters ("islands") based on their backend affinity.

This structure is heterogeneous by default and we have two levels of execution:
1. The lower level, where backends execute their subgraphs (islands). Every backend is in charge in providing its own execution logic under the hood.
2. The upper level, where graph of islands (and so, backends) is executed. The upper level is backend-neutral. G-API implements common logic to schedule graph of islands, and backends are asked to run their parts in their own manner when G-API asks them.

There are two modes of execution defined at (2): the regular mode (the product of `compile()`), and the streaming mode (the product of `compileStreaming()`).

The regular mode is so-far serial, but @anton-potapov has his TBB-based executor coming in #17170. This executor will enable *graph-level parallelism* for the regular (`GExecutor`) thing. *graph-level parallelism* makes sense when we have branches in the graph which can be scheduled in parallel, which applies to an island model perfectly if there's more than one island. This parallelism wouldn't give much if our graph is a linear (`a->b->c`).

The streaming mode is threaded as you see, and implements *graph-level pipelining*. The pipelining makes most sense when different stages of your pipeline can run in parallel (that is, those are heterogeneous too). It shows improvement even for linear structures like `a->b->c` if and only if `a` and `b` and `c` can run in parallel with no effect on each other (e.g., decoding with a hardware decoder, then running inference on GPU, then running post-processing/analysis/etc on CPU).

And here comes the main question - what kind of performance effect you really want to achieve on your target system?

modules/dnn/src/dnn.cpp
Yougmark: I read your code in your `cuda4dnn-concurrent` branch.  If I understand correctly, per-layer streams are created to handle asynchronous pipelined execution, using events to enforce dependencies between layers as you mentioned.  I have a couple of questions. Would the processing of different inputs be serialized as each layer has only one event recorder? For example, the processing of image (x+1) must wait for the processing of image (x) to finish at each layer.  Another question is about `forwardAsync`, is the returned `AyncArray` used to pull for results asynchronously?  I'm not familiar with `AsyncPromise`.  Thanks!

Can I see your evaluation results if you have some already?

I suppose using events supports generic networks including non-linear ones.  But
I wonder if an alternative approach using per-input streams would work better for linear networks as events can be avoided.  Then for non-linear networks, a graph scheduling framework like G-API is required to enforce nodes/layers dependencies.

Yougmark:
Thanks for your summary, @dmatveev!  It is very helpful.  To answer your question,
> And here comes the main question - what kind of performance effect you really want to achieve on your target system?

I want to improve the throughput of neural networks that handle multiple data streams, by exploring parallelism on GPUs through executing multiple kernels simultaneously, and intra-node parallelism on CPUs through multi-threading.  So I need finer-grained pipelined execution between nodes, compared to the existing `GStreamingExecutor` that supports pipelined execution between islands.  #17170 seems very relevant, and I'll look into that.

Another goal of mine is not as relevant: I also want to use OpenCV in real-time systems that require managing computation workloads as real-time tasks, using SCHED_DEADLINE scheduler in Linux or other real-time operating systems (RTOSs).  One G-API node per real-time task is the granularity I want.  That takes at least one thread to represent each node.

These two goals are orthogonal, and I think the first might be more interesting to the OpenCV team.  I'm gonna work on the executor for a while.  Do you use an email list or chat app like Slack where I can ask questions during the development?

dmatveev:
> I want to improve the throughput of neural networks that handle multiple data streams, by exploring parallelism on GPUs through executing multiple kernels simultaneously, and intra-node parallelism on CPUs through multi-threading. So I need finer-grained pipelined execution between nodes, compared to the existing GStreamingExecutor that supports pipelined execution between islands. #17170 seems very relevant, and I'll look into that.

Thanks, it is more clear now. Sounds challenging though :) Can modern GPUs execute multiple kernels simultaneously? I'm just not following the SOTA there. `GIslandExecutor` still could work for this, if we forbid fusing the relevant (DNN) operations to the same `island`. It is not supported out of the box, but the island fusion pass can be extended easily for that. Here comes some downsides, though:
1. `GIslandExecutor` will create its own thread for every island. If you don't let network fuse into a single island, you'll get as many threads as layers your network has and it will obviously die of oversubsription
2. Our islands interoperate with CPU data so far (that is, `cv::Mat`, `std::vector<>`, and so on) what may also kill the performance if you're running a discrete GPU. This model assumes our data is always "in place" and is never "remote". The support on exchanging the remote data between islands is WIP (you might find some `RMat` PRs open here). This problem can be work-arounded if use start exchanging with `GOpaque<>`s of whatever remote data between your "virtual" nodes, and e.g. keep the memory handles within, but changing this affects the API of every operation and is not quite portable (if you put a platform-specific handle into a `GOpaque<>`.

The proper and so, the G-API-ish way to do that is to write a new backend and map the graph structure to your own execution model. Then the TBB executor could really help, but you'll need to adapt it _within_ your backend (by default it will be used at a higher level, to schedule islands). Having the stuff consolidated within your own backend also solves the memory management problem.

> Another goal of mine is not as relevant: I also want to use OpenCV in real-time systems that require managing computation workloads as real-time tasks, using SCHED_DEADLINE scheduler in Linux or other real-time operating systems (RTOSs). One G-API node per real-time task is the granularity I want. That takes at least one thread to represent each node.

That's really interesting :) Please keep us informed with your progress. While this was not the goal for G-API to support, it sounds as mapping to the G-API model and nature pretty well.

Yougmark:
@dmatveev  Thanks for your suggestions!  I agree that a backend for my execution model is the way to go.  I'm not familiar with TBB, though, and after skimming about TBB, I have a couple of questions: 
- does TBB supports non-busy (suspension-based) waiting, e.g., is the `pop` function of `concurrent_bounded_queue` or `concurrent_queue` non-busy and waits for the signal of new data? 
- can I have persistent threads managed by OS schedulers, instead of tasks managed by the task scheduler in TBB?

If not, I might need to implement non-busy waiting and persistent threads on my own.

Regarding the question about the simultaneous execution on GPU, yes, you can look up concurrent kernel execution, which is supported since NVIDIA Fermi architecture, and similar concurrency is supported on AMD GPUs as well.

dmatveev:
@Yougmark sorry I've missed your comment.

@anton-potapov  can you have a look on the questions?

asmorkalov:
@Yougmark Could you rebase the patch to fix merge conflicts.
@anton-potapov  Friendly reminder.

asmorkalov:
@Yougmark @anton-potapov Do you have chance to finish the PR? Is it still relevant?

asmorkalov:
@Yougmark @dmatveev @dkurt Is the patch still relevant. Do you plan to work on it in mean time?

dkurt:
Hi! Still there is no clear understanding if this implementation can give permormance/memory optimizations.

Yougmark:
Closing this PR for now.

PR #17250, opened 2020-05-09
Restore LineSegmentDetector LSD & avoid license conflict

ruelj2:
Relative to [#2524](https://github.com/opencv/opencv_contrib/issues/2524#issue-615242133)
Original issue with license conflict: [#2016](https://github.com/opencv/opencv_contrib/issues/2016)

### DIFF
--- a/modules/imgproc/include/opencv2/imgproc.hpp
+++ b/modules/imgproc/include/opencv2/imgproc.hpp
@@ -1248,12 +1248,14 @@ protected:
 //! @addtogroup imgproc_feature
 //! @{
 
+/** @example samples/cpp/lsd_lines.cpp
+An example using the LineSegmentDetector
+\image html building_lsd.png "Sample output image" width=434 height=300
+*/
+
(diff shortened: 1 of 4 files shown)


asmorkalov:
@vpisarev could you look at it?

vpisarev:
@ruelj2, thank you for the contribution! It's much better, but still I can find many many similarities, e.g. in the rect_improve, like variable names, comments etc. That is, the code does not seem to be written from scratch, but rather a slightly refactored original implementation

ruelj2:
Thanks for the review! As the [original issue with license conflict](https://github.com/opencv/opencv_contrib/issues/2016) stated, the only code that seemed to be conflicting was the "Compare doubles by relative error" which is the only thing I changed from the original code.

As mentioned [here](https://github.com/opencv/opencv_contrib/issues/2524#issuecomment-626290398), it now seems that the whole code should be refactored from scratch. Working on that, closing the PR.

PR #17254, opened 2020-05-10
add a DeepGreen colormap

jsxyhelu:
### This pullrequest changes
I implemented the colormap "DeepGreen"
And, I add cv::COLORMAP_DEEPGREEN as the flag of cv::applyColorMap.
I checked this feature using the following code.
```
#include <opencv2/core.hpp>
#include <opencv2/imgproc.hpp>
#include <opencv2/imgcodecs.hpp>

int main(int argc, const char* argv[])
{
    cv::Mat disp(cv::Size(256, 30), CV_8UC1, cv::Scalar(0));
    for(int y = 0; y < disp.rows; y++)
    {
        for(int x = 0; x < disp.cols; x++)
        {
            disp.at<uchar>(y, x) = x;
        }
    }
    cv::Mat deepgreen;
    cv::applyColorMap(disp, deepgreen, cv::COLORMAP_DEEPGREEN);
    cv::imwrite("colorscale_deepgreen.jpg", deepgreen);
    return 0;
}
```
colorscale_deepgreen

### Pull Request Readiness Checklist

See details at https://github.com/opencv/opencv/wiki/How_to_contribute#making-a-good-pull-request

- [x] I agree to contribute to the project under OpenCV (BSD) License.
- [x] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
- [x] The PR is proposed to proper branch

PR #17260, opened 2020-05-11
Implementation of colormap "DeepGreen"

jsxyhelu:
### This pullrequest changes
from  [#17254](https://github.com/opencv/opencv/pull/17254)
I need to implement the "false color" transformation of the green theme. Of the templates currently provided, only COLORMAP_SUMMER is the closest, but its color is too light and does not look very clear,in my opinion. So I combined the two existing templates of COLORMAP_OCEAN and COLORMAP_SUMMER to build a new colormap "DeepGreen". It can achieve significant pseudo-color transformation of the green theme.
Also, I add cv::COLORMAP_DEEPGREEN as the flag of cv::applyColorMap.

I checked this feature using the following code.
```
#include <opencv2/core.hpp>
#include <opencv2/imgcodecs.hpp>

int main(int argc, const char* argv[])
{
    cv::Mat disp(cv::Size(256, 30), CV_8UC1, cv::Scalar(0));
    for(int y = 0; y < disp.rows; y++)
    {
        for(int x = 0; x < disp.cols; x++)
        {
            disp.at<uchar>(y, x) = x;
        }
    }
    cv::Mat deepgreen;
    cv::applyColorMap(disp, deepgreen, cv::COLORMAP_DEEPGREEN);
    cv::imwrite("colorscale_deepgreen.jpg", deepgreen);
    return 0;
}
```
colorscale_deepgreen


asmorkalov:
Thanks for the contribution. Could you add some tests for the new feature or extend the existing one?

jsxyhelu:
@asmorkalov This pullrequest hase been changed。

PR #17275, opened 2020-05-13
fix bug for CHECK bpp issue

shijiayuan:
**Merge with extra**: https://github.com/opencv/opencv_extra/pull/758

For the tiff file which has 1 bpp, it will fail after calling CV_TIFF_CHECK_CALL. But TIFFGetField for bpp failed will not influence the final image decoding result. I have attached a tiff file for this case.(https://drive.google.com/file/d/1NurdgX8QF_7Nig-k6n_BFDY52_pY8Yob/view?usp=sharing)
Too many CHECK will result regression.

<cut/>

```
opencv_extra=master
```

### DIFF
--- a/modules/imgcodecs/test/test_tiff.cpp
+++ b/modules/imgcodecs/test/test_tiff.cpp
@@ -333,6 +333,17 @@ TEST(Imgcodecs_Tiff, decode_black_and_write_image_pr12989_default)
     EXPECT_EQ(CV_8UC3, img.type()) << cv::typeToString(img.type());
 }
 
+TEST(Imgcodecs_Tiff, decode_black_and_write_image_pr12989_default)
+{
+    const string filename = cvtest::findDataFile("readwrite/bitsperpixel1_min.tiff");
+    cv::Mat img;
+    ASSERT_NO_THROW(img = cv::imread(filename));  // by default image type is CV_8UC3
+    ASSERT_FALSE(img.empty());
+    EXPECT_EQ(64, img.cols);
+    EXPECT_EQ(64, img.rows);
+    EXPECT_EQ(CV_8UC3, img.type()) << cv::typeToString(img.type());
+}
+
 #endif
 
 }} // namespace
--- a/modules/imgcodecs/test/test_tiff.cpp
+++ b/modules/imgcodecs/test/test_tiff.cpp
@@ -308,7 +308,7 @@ TEST(Imgcodecs_Tiff, imdecode_no_exception_temporary_file_removed)
 }
 
 
-TEST(Imgcodecs_Tiff, decode_black_and_write_image_pr12989)
+TEST(Imgcodecs_Tiff, decode_black_and_write_image_pr12989_grayscale)
 {
     const string filename = cvtest::findDataFile("readwrite/bitsperpixel1.tiff");
     cv::Mat img;
@@ -333,7 +333,7 @@ TEST(Imgcodecs_Tiff, decode_black_and_write_image_pr12989_default)
     EXPECT_EQ(CV_8UC3, img.type()) << cv::typeToString(img.type());
 }
 
-TEST(Imgcodecs_Tiff, decode_black_and_write_image_pr17275)
+TEST(Imgcodecs_Tiff, decode_black_and_write_image_pr17275_grayscale)
 {
     const string filename = cvtest::findDataFile("readwrite/bitsperpixel1_min.tiff");
     cv::Mat img;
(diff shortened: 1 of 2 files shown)


alalek:
Change doesn't look consistent:
- there are multiple places reading of `TIFFTAG_BITSPERSAMPLE` value. Changed one place only.
- default value is not correct and using "8" value.
- `CV_TIFF_CHECK_CALL` is about security and vulnerabilities at first. Affected code paths must be analysed.

What is the source of mentioned image? Is it correct in terms of TIFF format? Does it generated using original TIFF library?

shijiayuan:
> Change doesn't look consistent:
> 
> * there are multiple places reading of `TIFFTAG_BITSPERSAMPLE` value. Changed one place only.
> * default value is not correct and using "8" value.
> * `CV_TIFF_CHECK_CALL` is about security and vulnerabilities at first. Affected code paths must be analysed.
> 
> What is the source of mentioned image? Is it correct in terms of TIFF format? Does it generated using original TIFF library?

I

shijiayuan:
> Change doesn't look consistent:
> 
> * there are multiple places reading of `TIFFTAG_BITSPERSAMPLE` value. Changed one place only.
> * default value is not correct and using "8" value.
> * `CV_TIFF_CHECK_CALL` is about security and vulnerabilities at first. Affected code paths must be analysed.
> 
> What is the source of mentioned image? Is it correct in terms of TIFF format? Does it generated using original TIFF library?

Hi,
I have removed the check for the TIFFTAG_BITSPERSAMPLE reading, my tiff file's bpp is 1, but will get 0 after running TIFFGetField for bpp. The tiff file can be opened in both windows and mac system, you can find it here:[https://drive.google.com/file/d/1NurdgX8QF_7Nig-k6n_BFDY52_pY8Yob/view](url). It is a standard tiff filed and it shouldn't be rejected by opencv tiff decoder. CV_TIFF_CHECK_CALL for bpp would cause regression for tiff files.

shijiayuan:
I also have question: how could I rebuild my branch? Since there is interruption for the build.

shijiayuan:
For the pdf(http://www.exif.org/TIFF6.pdf) in opencv/3rdparty/libtiff/tiff.h, You can see Section3 of Part1, there is an introduction of Bilevel Images. For this kind of tiff image, BitsPerSample filed is not required(page 21). CV_TIFF_CHECK_CALL for bpp is not necessary because it is not a required field for all tiff files.

shijiayuan:
image
My tiff images are bilevel image, which do not have BitsPerSample filed. It shouldn't be check for this kind of image.

mshabunin:
@shijiayuan , is there a test for this functionality? If not it should be added along with bi-level test image (goes to opencv_extra).

shijiayuan:
> @shijiayuan , is there a test for this functionality? If not it should be added along with bi-level test image (goes to opencv_extra).
Hi @mshabunin , I have added the bilevel test image in opencv_extra:
https://github.com/opencv/opencv_extra/pull/758/files

mshabunin:
@shijiayuan , oh looks like there is bi-level tiff already (`bitsperpixel1.tiff`). Please check whether it participates in any tests and has or has no BPS tag.

shijiayuan:
> @shijiayuan , oh looks like there is bi-level tiff already (`bitsperpixel1.tiff`). Please check whether it participates in any tests and has or has no BPS tag.

@mshabunin , It looks like the  (`bitsperpixel1.tiff`) image is not standard bi-level image, which has BitPerSample field but my tiff image doesn't have this filed.
image

mshabunin:
So I suggest creating a copy of that file with required fields only (as per specification), e.g. `bitsperpixel1_min.tiff` and adding it to the the test.

shijiayuan:
> So I suggest creating a copy of that file with required fields only (as per specification), e.g. `bitsperpixel1_min.tiff` and adding it to the the test.

Sure, thanks for the suggestion, I have uploaded the bitsperpixel1_min.tiff file in https://github.com/opencv/opencv_extra/pull/758/files

shijiayuan:
@alalek  Could you pls to send me instructions on how to add test on it? Thanks a lot!

alalek:
@shijiayuan Please use [this test code](https://github.com/opencv/opencv/blob/3.4.10/modules/imgcodecs/test/test_png.cpp#L53-L94) as a reference.

shijiayuan:
@alalek  Test added, please help to review it, thanks a lot!

modules/imgcodecs/test/test_tiff.cpp
```diff
@@ -333,6 +333,31 @@ TEST(Imgcodecs_Tiff, decode_black_and_write_image_pr12989_default)
     EXPECT_EQ(CV_8UC3, img.type()) << cv::typeToString(img.type());
 }
 
+TEST(Imgcodecs_Tiff, decode_black_and_write_image_pr12989)
```
mshabunin, line 336: Please rename new tests, they are duplicated.
Did you run these tests locally?

shijiayuan: Sorry for that, I have updated the name and ran these tests locally. Thanks!

shijiayuan:
> Thank you for contribution 👍

Thanks a lot! Really appreciate!

PR #17281, opened 2020-05-13
Determine SSD input shape

l-bat:


samples/dnn/tf_text_graph_ssd.py
```diff
@@ -234,6 +234,9 @@ def to_remove(name, op):
 
     # Connect input node to the first layer
     assert(graph_def.node[0].op == 'Placeholder')
+    input_shape = graph_def.node[0].attr['shape']['shape'][0]['dim']
```
dkurt, line 237: Is that possible that model does not have such field? In example, after `optimize_for_inference_lib` or `TransformGraph`?

PR #17313, opened 2020-05-17
Revise and expand kNN Python tutorials

dmallia17:
This pull request covers a number of revisions and small additions to the kNN tutorials for OpenCV-Python. I would be very happy to do the same for the other Python ML tutorials!

Please note that I am not sure why some sentences were split across multiple lines. While it does not seem to reflect on Github, this can be seen by viewing the *original unaltered* files (and other ML tutorials) in VSCode and Vim.

### DIFF
--- a/doc/py_tutorials/py_ml/py_knn/py_knn_opencv/py_knn_opencv.markdown
+++ b/doc/py_tutorials/py_ml/py_knn/py_knn_opencv/py_knn_opencv.markdown
@@ -4,20 +4,20 @@ OCR of Hand-written Data using kNN {#tutorial_py_knn_opencv}
 Goal
 ----
 
-In this chapter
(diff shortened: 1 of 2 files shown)


dmallia17:
Just committed again to correct the broken NPTEL link I had missed earlier. I can, of course, squash the commits if approved, but was not sure if this had already been seen and wanted to make the additional change clear.

alalek:
> why some sentences were split across multiple lines.

It is better to avoid starting a new large sentence in the middle of the line.
Start them from the beginning.
There is no problem to split sentence on multiple lines.
Limit number of symbols to 100-120 per line.

Markdown format merges lines into a single paragraph (empty dedicated line is required to start new paragraph).

dmallia17:
@alalek Ah that makes sense - thank you very much for the explanation and for accepting the pr!

PR #17331, opened 2020-05-20
Objdetect: QR decoder: Update 3rd party

allnes:
Update [quirc library](https://github.com/dlbeer/quirc) to v1.1 version

```
allow_multiple_commits=1
```


asmorkalov:
@allnes Please add link/reference to quirc library version that you used for update.

Issue #17246, opened 2020-05-09
cv::dnn::NMSBoxes bug


dzywater:
OpenCV => 4.3
Operating System / Platform => Windows 64 Bit
Compiler => Visual Studio 2015

Question description:
NMSBoxes:  should do NMS for each class not for all classes. The input params which do not contain class indexes could reflect this issue.

dkurt:
Hi! Please take a look at conversation https://github.com/opencv/opencv/issues/17111.

Issue #17111, opened 2020-04-20
Using opencv to call the darknet version of yolo results are inconsistent


Hiwyl:
`idxs = cv2.dnn.NMSBoxes(boxes, confidences, confident, threshold)`
You can see that the source code is NMS between categories，there is no incoming category-related information.The NMS of object detection is based on the category, so this is the essential reason for the difference between the two results.Therefore the following changes should be made，with python:
`def NMSBoxes_fix(boxes, confidences, confidence_thre, nms_thre, class_id):`
     `class_id_set = set(class_id)`  
      `result = [] ` 
      `for cls in class_id_set: `
        `cls_boxes = []  `
        `cls_confidences = []  `
      `  indices = [i for i, c in enumerate(class_id) if c == cls] `
        `for i in indices:`
            `cls_boxes.append(boxes[i])`
           ` cls_confidences.append(confidences[i]) `
       ` idxs = cv2.dnn.NMSBoxes(cls_boxes, cls_confidences, confidence_thre, nms_thre) `
       ` for i in idxs:  `
          `  result.append([indices[i[0]]])   `
   ` return np.array(result)`

dkurt:
Hi!

Can you please provide a source of information why NMS should be applied for each class but not for all the detections? Let's assume we have overlapped boxes with high IoU for the same object but two classes,

![image](https://user-images.githubusercontent.com/25801568/79720695-b7bf9b80-82e9-11ea-8998-acba926a20d7.png)

Current approach will keep the highest score object and reject all others detections which have high IoU with selected box but lower score.

Can you provide a reproducer with incorrect results?

Hiwyl:
Hi!
This is the result of darknet，the result is in line with my expectation.
obj and obj1 are the two labels I want to recognize.
![image](https://user-images.githubusercontent.com/44716550/79722525-ca9c9500-8316-11ea-9270-a7137cf6d62f.png)

but,when I did NMS with opencv, the results were very different:
![image](https://user-images.githubusercontent.com/44716550/79724317-e5bcd400-8319-11ea-889d-1458689320f1.png)

The recognition results are as follows:
![image](https://user-images.githubusercontent.com/44716550/79724548-5532c380-831a-11ea-93c0-e357e82d32f5.png)

you can see,this inter-category based NMS suppresses boxes larger than the iou threshold，and these are the boxes that I need to recognize.

[...]

PR #17332, opened 2020-05-20
Fixed NMSBoxes bug

l-bat:
resolves: opencv/opencv#17246, opencv/opencv#17111

<details>
<summary>Hypothesis: NMS over multiple output layers is not needed (i.e. YOLOv3, YOLOv4)</summary>

**Hypothesis**: NMS over multiple output layers is not needed (i.e. YOLOv3, YOLOv4)
**Resuls**: NMS is required

Output without NMS in sample - 10 detections. Some of detections have an IoU approximately equal 1.
no_nms

Output with NMS in sample - 7 detections.
nms

</details>

### DIFF
--- a/samples/dnn/object_detection.cpp
+++ b/samples/dnn/object_detection.cpp
@@ -1,6 +1,5 @@
 #include <fstream>
 #include <sstream>
-#include <numeric>
 
 #include <opencv2/dnn.hpp>
 #include <opencv2/imgproc.hpp>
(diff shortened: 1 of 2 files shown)


samples/dnn/object_detection.py
```diff
@@ -163,9 +160,23 @@ def drawPred(classId, conf, left, top, right, bottom):
         print('Unknown output layer type: ' + lastLayer.type)
         exit()
 
-    indices = cv.dnn.NMSBoxes(boxes, confidences, confThreshold, nmsThreshold)
+    indices = []
+    if lastLayer.type == 'Region' and len(outNames) > 1:
```
dkurt, line 164: why only for Region?

samples/dnn/object_detection.py
```diff
@@ -163,9 +160,23 @@ def drawPred(classId, conf, left, top, right, bottom):
...
+        classIds = np.array(classIds)
+        boxes = np.array(boxes)
+        confidences = np.array(confidences)
+        unique_classes = set(classIds)
+        for cl in unique_classes:
+            class_indices = np.where(classIds == cl)[0]
+            conf = confidences[class_indices]
+            box  = boxes[class_indices].tolist()
+            nms_indices = cv.dnn.NMSBoxes(box, conf, confThreshold, nmsThreshold)
```
dkurt, line 173: Can you please check if we really need NMS for YOLOv3? I mean if NMS in 3 Region layers is not enough.

samples/dnn/object_detection.py
```diff
@@ -163,9 +160,23 @@ def drawPred(classId, conf, left, top, right, bottom):
         print('Unknown output layer type: ' + lastLayer.type)
         exit()
 
-    indices = cv.dnn.NMSBoxes(boxes, confidences, confThreshold, nmsThreshold)
+    indices = []
+    if lastLayer.type == 'Region' and len(outNames) > 1:
+        classIds = np.array(classIds)
+        boxes = np.array(boxes)
+        confidences = np.array(confidences)
```
dkurt, line 167: Is that necessary?

dkurt:
Please modify C++ sample as well.

samples/dnn/object_detection.py
l-bat: Yes, because `class_indices` may be a list. Onle `np.array` support a list of indices.

samples/dnn/object_detection.py
l-bat: Added proof to description

samples/dnn/object_detection.py
dkurt: But what about boxes and confidences?

l-bat: We need np.array(boxes) to get boxes[list_indices]

samples/dnn/object_detection.cpp
```diff
@@ -397,7 +398,37 @@ void postprocess(Mat& frame, const std::vector<Mat>& outs, Net& net)
         CV_Error(Error::StsNotImplemented, "Unknown output layer type: " + outLayerType);
 
     std::vector<int> indices;
-    NMSBoxes(boxes, confidences, confThreshold, nmsThreshold, indices);
+    if (outLayers.size() > 1)
+    {
+        std::map<int, std::vector<int>> class2indices;
+        for (int i = 0; i < classIds.size(); i++)
+        {
+            class2indices[classIds[i]].push_back(i);
```
dkurt, line 406: I'd like to recommend to filter detections by `confThreshold` here.

samples/dnn/object_detection.cpp
```diff
@@ -397,7 +398,37 @@ void postprocess(Mat& frame, const std::vector<Mat>& outs, Net& net)
...
+            std::vector<int> class_indices = it->second;
+            for (size_t i = 0; i < class_indices.size(); i++)
+            {
+                local_boxes.push_back(boxes[class_indices[i]]);
+                local_confidences.push_back(confidences[class_indices[i]]);
+            }
+            NMSBoxes(local_boxes, local_confidences, confThreshold, nmsThreshold, nms_indices);
+            for (int j = 0; j < nms_indices.size(); j++)
+            {
+                indices.push_back(class_indices[nms_indices[j]]);
```
dkurt, line 422: If we will filter boxes by confidence above, I think it's better to just collect all the boxes into a vector and then override origin `boxes`. So we will avoid `iota`

samples/dnn/object_detection.cpp
```diff
@@ -396,13 +396,48 @@ void postprocess(Mat& frame, const std::vector<Mat>& outs, Net& net)
...
+                nmsConfidences.push_back(local_confidences[idx]);
+                nmsClassIds.push_back(it->first);
+            }
+        }
+        boxes = nmsBoxes;
+        classIds = nmsClassIds;
+        confidences = nmsConfidences;
+    }
+
+    for (size_t i = 0; i < boxes.size(); ++i)
```
dkurt, line 437: it'd be perfect to name the iterator as `idx` (so two lines below will be unchanged :) )

PR #17346, opened 2020-05-21
added bash command to create shared library in the system lib path /usr/local/lib

nilsjohnson:
In order to allow code to run after installation, we need to add opencv as a shared library to the system library path.

- [ x] I agree to contribute to the project under OpenCV (BSD) License.
- [x ] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
- [ x] The PR is proposed to proper branch


alalek:
It is a bad idea in general. There too many "notes", including many Linux distros already do this thing, others don't have that for a reason - due to garbage in `/usr/local` which may break system apps.

---

To try OpenCV you do NOT need to install it.
Read CMake manual about `find_package` and use `find_package(OpenCV)` with your apps.

Issue #17319, opened 2020-05-18
Implement 17 Photoshop blending modes 


jsxyhelu:
##### System information (version)
- OpenCV => 4.3:
- Operating System / Platform => ALL:
- Compiler => ALL:

##### Detailed description

Inspired by GIMP and [Formulas for Photoshop blending modes,](http://www.deepskycolors.com/archivo/2010/04/21/formulas-for-Photoshop-blending-modes.html) I'd like to implement  the following 17  Photoshop blending modes。

![捕获](https://user-images.githubusercontent.com/16490259/82269839-792f0680-99a5-11ea-810e-b8f1c1c452e4.JPG)

I wonder this code should be put under opencv / photo module or opencv_contrib / xphoto module
？Any other suggestions or requirements? I'd like to use this issue to track progress.

##### Issue submission checklist

 - [ ] I report the issue, it's not a question
   
 - [x] I checked the problem with documentation, FAQ, open issues,
       answers.opencv.org, Stack Overflow, etc and have not found solution
   
 - [x] I updated to latest OpenCV version and the issue is still there
   
 - [ ] There is reproducer code and related data files: videos, images, onnx, etc

tweenietomatoes:
nice title

jsxyhelu:
I will make a PR later.We can discuss there.

pwnorbitals:
This sounds great @jsxyhelu, are you having any success ? :)

PR #17347, opened 2020-05-22
Add layer model blending

jsxyhelu:
### The PR has been Updated!
from [https://github.com/opencv/opencv/issues/17319](url)
Inspired by GIMP 、https://en.wikipedia.org/wiki/Blend_modes and https://stackoverflow.com/questions/5919663/how-does-photoshop-blend-two-images-together, I'd like to implement the following 17 Photoshop blending modes.

<cut/>

82269839-792f0680-99a5-11ea-810e-b8f1c1c452e4

An example using the layerModelBlending technique has also been given!
https://github.com/opencv/opencv/pull/17347/files#diff-e174f4325b3bc9b66c1b068639ffec30
for detail .

results on lena and it's gaussblur


asmorkalov:
@jsxyhelu Thanks for the contribution. Could you add simple test for the new functionality?

asmorkalov:
Example code could be useful too.

modules/photo/src/layerModelBlending.cpp
```diff
@@ -0,0 +1,201 @@
...
+    if (fabsf(a) > SAFE_DIV_MIN)
+    {
+        result = a / b;
+        result = CLAMP(result, -SAFE_DIV_MAX, SAFE_DIV_MAX);
+    }
+    return result;
+}
+CV_EXPORTS_W void layerModelBlending(InputArray _target, InputArray _blend, OutputArray _dst, int flag)
+{
+    Mat target = _target.getMat();
```
mshabunin, line 74: * This function can work only with CV_32FC3 arrays, but there is no guard against other types.
* There is no check of target and blend size equality
* Perhaps you should call `_dst.create()` to allocate memory if `_dst` does not exist or have different size

modules/photo/src/layerModelBlending.cpp
```diff
@@ -0,0 +1,201 @@
...
+    Mat blend = _blend.getMat();
+    Mat dst = _dst.getMat();
+
+    for (int index_row = 0; index_row < target.rows; index_row++)
+        for (int index_col = 0; index_col < target.cols; index_col++)
+            for (int index_c = 0; index_c < 3; index_c++)
+                switch (flag)
```
mshabunin, line 81: I believe per-pixel loop with a switch on every iteration is not efficient way of doing it. Is it possible to use existing operations to implement blending, e.g. [addWeighted](https://docs.opencv.org/master/d2/de8/group__core__array.html#gafafb2513349db3bcff51f54ee5592a19)?

modules/photo/include/opencv2/photo.hpp
```diff
@@ -851,6 +851,45 @@ CV_EXPORTS_W void stylization(InputArray src, OutputArray dst, float sigma_s = 6
 
 //! @} photo_render
 
+//! @addtogroup layerModelBlending
+//! @{
+
+//! layerModelBlending algorithm flags
+enum
```
mshabunin, line 858: It is better to use SnakeCase naming style for enum values to avoid conflicts with preprocessor definitions. `enum` should be named, so it can be referred in documentation and used in function signature instead of `int`.

modules/photo/include/opencv2/photo.hpp
```diff
@@ -851,6 +851,45 @@ CV_EXPORTS_W void stylization(InputArray src, OutputArray dst, float sigma_s = 6
...
+    VIVID_LIGHT = 12,          //(Blend > 0.5) * (1 - (1-Target) / (2*(Blend-0.5))) +(Blend <= 0.5) * (Target / (1-2*Blend))
+    LINEAR_LIGHT = 13,         //(Blend > 0.5) * (Target + 2*(Blend-0.5)) +(Blend <= 0.5) * (Target + 2*Blend - 1)
+    PIN_LIGHT = 14,            //(Blend > 0.5) * (max(Target,2*(Blend-0.5))) +(Blend <= 0.5) * (min(Target,2*Blend)))
+    DIFFERENCE = 15,           //| Target - Blend |         
+    EXCLUSION = 16,            //0.5 - 2*(Target-0.5)*(Blend-0.5)
+    DIVIDE = 17                //Target/Blend
+
+};
+
+/** @17 Photoshop blending modes Inspired by GIMP and
```
mshabunin, line 880: First line of documentation should be a short description of a function.

modules/photo/include/opencv2/photo.hpp
```diff
@@ -851,6 +851,45 @@ CV_EXPORTS_W void stylization(InputArray src, OutputArray dst, float sigma_s = 6
...
+
+@param target and blend Input SAME_SIZE 8-bit 3-channel image.
+@param dst Output 8-bit 3-channel image.
+@param flags layerModelBlending algorithm : CV::DARKEN、CV::MULTIPY、CV::COLOR_BURN 、CV::LINEAR_BRUN
+、CV::LIGHTEN 、CV::SCREEN 、CV::COLOR_DODGE 、CV::LINEAR_DODGE 、CV::OVERLAY、CV::SOFT_LIGHT、CV::HARD_LIGHT、
+CV::VIVID_LIGHT、CV::LINEAR_LIGHT、CV::PIN_LIGHT、CV::DIFFERENCE、CV::EXCLUSION 、CV::DIVIDE 
+*/
+CV_EXPORTS_W void layerModelBlending(InputArray target, InputArray blend, OutputArray dst, int flag=1);
+
+//! @} addtogroup layerModelBlending
```
mshabunin, line 891: It is not necessary to create a doc group with single function.

modules/photo/include/opencv2/photo.hpp
```diff
@@ -851,6 +851,45 @@ CV_EXPORTS_W void stylization(InputArray src, OutputArray dst, float sigma_s = 6
...
+
+};
+
+/** @17 Photoshop blending modes Inspired by GIMP and
+http://www.deepskycolors.com/archivo/2010/04/21/formulas-for-Photoshop-blending-modes.html.
+
+@param target and blend Input SAME_SIZE 8-bit 3-channel image.
+@param dst Output 8-bit 3-channel image.
+@param flags layerModelBlending algorithm : CV::DARKEN、CV::MULTIPY、CV::COLOR_BURN 、CV::LINEAR_BRUN
+、CV::LIGHTEN 、CV::SCREEN 、CV::COLOR_DODGE 、CV::LINEAR_DODGE 、CV::OVERLAY、CV::SOFT_LIGHT、CV::HARD_LIGHT、
```
mshabunin, line 886: Looks like documentation contains non-ASCII symbols.

modules/photo/src/layerModelBlending.cpp
```diff
@@ -0,0 +1,201 @@
+/*M///////////////////////////////////////////////////////////////////////////////////////
```
mshabunin, line 1: It is better to use modern 3-line license header.

jsxyhelu: yes,modern 3-line license header looks better.

modules/photo/include/opencv2/photo.hpp
jsxyhelu: Yes, I should guarantee a unified code style.

modules/photo/include/opencv2/photo.hpp
jsxyhelu: delete non-ASCII yet

modules/photo/include/opencv2/photo.hpp
jsxyhelu: yes,I missed @brief

modules/photo/include/opencv2/photo.hpp
jsxyhelu: Can it not be added? Let me see

modules/photo/src/layerModelBlending.cpp
jsxyhelu: “addWeighted” is very worth learning, and I also know there are some methods to improve the speed loop。However that will make the code more difficult to read, and Less intuitive。 I am not sure if it is scientific to do so?@mshabunin

modules/photo/src/layerModelBlending.cpp
jsxyhelu: yes,you are right,working at it.

modules/photo/src/layerModelBlending.cpp
mshabunin: We try to focus on efficiency, if you want to demonstrate library usage it is better to create a sample application or tutorial. There is tutorial about pixel-wise operations: https://docs.opencv.org/master/db/da5/tutorial_how_to_scan_images.html and one about blending: https://docs.opencv.org/master/d5/dc4/tutorial_adding_images.html . C++ samples are here: https://github.com/opencv/opencv/tree/master/samples/cpp

jsxyhelu: > tutorial
The PR has been updated.
I can barely write sample code that meets the quality requirements, but the tutorial is relatively rusty, especially my existing editor does not recognize the markdown format on github very well, I hope to give some knowledge. I very much hope that my code can be integrated into OpenCV, and through this process to grow my relevant experience.3Q

jsxyhelu: A supporting demo has been uploaded, and the relevant parts of the comments have also been modified. Thanks for suggestions!

jsxyhelu:
> Example code could be useful too.


A supporting demo has been uploaded, and the relevant parts of the comments have also been modified. Thanks for suggestions!

jsxyhelu:
> Example code could be useful too.

@asmorkalov example has been added

mshabunin:
@jsxyhelu , some of my previous comments have not been addressed.

You have to choose on of following forms for your contribution:
* a tutorial or sample application which demonstrates how OpenCV functions can be used in real world task - a function is not needed, all methods should be implemented in the application so that users can read it and copy snippets of code
* a function implementing an algorithm in an efficient way, convenient to use, covered with tests - sample application is not needed for a single function with simple arguments

jsxyhelu:
> @jsxyhelu , some of my previous comments have not been addressed.
> 
> You have to choose on of following forms for your contribution:
> 
> * a tutorial or sample application which demonstrates how OpenCV functions can be used in real world task - a function is not needed, all methods should be implemented in the application so that users can read it and copy snippets of code
> * a function implementing an algorithm in an efficient way, convenient to use, covered with tests - sample application is not needed for a single function with simple arguments

cool!work at it now.

mshabunin:
1. You should move the `switch` statement to the top level, CPUs don't like conditional statements inside the loop body:
    ```.cpp
    switch(mode)
    {
    case A: for(...) for(...) { /* do something */ } break;
    ...
    }
    ```
    Use existing functions where possible, e.g. [multiply](https://docs.opencv.org/master/d2/de8/group__core__array.html#ga979d898a58d7f61c53003e162e7ad89f) for `Multiply` mode.
    Further optimization steps include vectorization and parallelization, but these can be done in a separate PR later.
2. I think function should accept 8UC3 arrays as input, formulas should be modified accordingly, e.g. "inverted" should mean `255 - X` instead of `1 - X`. Check [Wikipedia](https://en.wikipedia.org/wiki/Blend_modes) page and its references for more detailed descriptions.
3. Buildbot does not execute your function, because it is not used anywhere. To add it to buildbot runs it should be tested, see [contribution guide](https://github.com/opencv/opencv/wiki/Coding_Style_Guide#implementing-tests) for details. Tests should check good cases (all parameters are valid) as well as bad cases (some parameters are invalid, corner cases, overflows, etc.). You can use existing images for input (e.g. lena.jpg) and/or generate synthetic images (e.g. gradients). Output should be checked for correctness in some way.

jsxyhelu:
The PR has been improved
1、Have move the switch statement to the top level, it works fine;
2、Use existing functions such as Multiply, Add, abs, etc.
3、I still think CV_32FC3 is better（just in https://en.wikipedia.org/wiki/Blend_modes,it says
"In the formulas shown on this page, values go from 0.0  to 1.0");
@mshabunin

jsxyhelu:
Have tested it and will continue to do more tests~ Hopping to be merged!

mshabunin:
@jsxyhelu , most images used as input for blending will be in 8UC3 format and conversion to and from 32F will require additional processing as well as additional memory (4x the input image size!). I provided a link to Wikipedia page because it contains not only formulas for [0; 1] images, but also a text description and useful links. [This one](https://stackoverflow.com/questions/5919663/how-does-photoshop-blend-two-images-together) for example has formulas for 8U images.

jsxyhelu:
@mshabunin  https://stackoverflow.com/questions/5919663/how-does-photoshop-blend-two-images-together is great,many thanks.
I modified the program so that it handles the format of cv_8uc3 by default; at the same time, the demo is modified to superimpose the gaussblur of lena and lena, which can highlight the processing effect of the program.

modules/photo/src/layerModelBlending.cpp
```diff
@@ -8,37 +8,32 @@
...
-{
-    float result = 0.0f;
-    if (fabsf(a) > SAFE_DIV_MIN)
-    {
-        result = a / b;
-        result = CLAMP(result, -SAFE_DIV_MAX, SAFE_DIV_MAX);
-    }
-    return result;
-}
+#define LayerModelBlend_Darken(A,B)     ((uchar)((B > A) ? A:B))
```
mshabunin, line 11: If you want to copy code from StackOverflow, you should give a credit (e.g. comment with a link).

jsxyhelu:
1、Although I added this stackoverflow link in the main body of the PR, it was not added in the eye-catching position of the code. Your comment is right，Have fix it, thanks!
2、I can make sure that the codes I submitted in the CV_32FC3 format are correct. During the conversion to CV_8UC3, I made a detailed comparison and selection, and indeed the stackoverflow code is valid.
3、However, the stackoverflow code cannot be mixed with the layers that come with PS. (such as the lack of ‘Divide’, the addition of ‘Phoenix’, etc.), I still organize the code according to the 17 modes of PS.

I have a question. 
Earlier you mentioned that CV_32FC3 is much larger than CV_8UC3 in size. This is indeed the case. So is the CV_32FC3 much slower in terms of computing speed? I did some tests and found that the difference is not obvious? Is it optimized for CV_32FC3 in OpenCV or other reasons?

In addition, I used the "DIVIDE" method to achieve a good "scan enhancement" effect, but this should be another PR.

Thanks for your review and 
Longing for your guidance!
@mshabunin

jsxyhelu:
now,just win64 buildbot is error,reports" fatal error LNK1180: insufficient disk space to complete link "
en~~seems there is somewrong with buildbot?

jsxyhelu:
now,just win64 buildbot is error,reports" fatal error LNK1180: insufficient disk space to complete link "
en~~seems there is somewrong with buildbot?

alalek:
@jsxyhelu Working on fixing infrastructure. Sorry for inconvinience.

jsxyhelu:
Suggestions for revision are welcome! 
Where can I continue to improve? 
Or do I need to do those specific tests to prove that the function is correct?
@alalek @mshabunin

jsxyhelu:
Suggestions for revision are welcome! 
Where can I continue to improve? 
Or do I need to do those specific tests to prove that the function is correct?
@alalek @mshabunin

jsxyhelu:
yes,it is time to close this PR.

mshabunin:
@jsxyhelu , why did you close the PR?

You should've add a test which would call the function and check if results are correct.

jsxyhelu:
@mshabunin I close this PR because I did not kown what to do next.
I have add a demo,how to add a test?
please give me a tutorial and I will make it well.

mshabunin:
@jsxyhelu , we don't have a dedicated test writing tutorial. I recommend checking following materials:
* Example PR with a fixed bug and added test (photo module): https://github.com/opencv/opencv/pull/14729/files
* Recommendations from Coding Style Guide: https://github.com/opencv/opencv/wiki/Coding_Style_Guide#implementing-tests (specifically [Simple tests](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#simple-tests) section)
* Getting started with Googletest: https://github.com/google/googletest/blob/master/googletest/docs/primer.md

If you'll reopen the PR or create a new one, I can add a test skeleton to your branch.

PR #17349, opened 2020-05-22
cuda4dnn: do not create temporary unused handles

YashasSamaga:
Layers that require cuDNN or cuBLAS capabilities have handles as member object. A useless handle is created when they are default constructed. They are going to be replaced by the actual handle later. Handle creation is costly. Hence, all the temporary handles waste resources.

Each `cv::dnn::Net` object now creates exactly one stream, one cuDNN handle and one cuBLAS handle. 

This PR makes default constructed handle objects not create any handle. A special constructor must be used to create a handle.

Model                      | init time (master) | init time (with this PR)
-------------------------- | ---------------------- | -----------------------------
ResNet 50               | 260ms                 | 170ms                        
Enet Cityscapes      | 170ms                 | 30ms
YOLOv4                  | 630ms                 | 440ms
Inception v2 Mask RCNN            | 520ms                 | 340ms

**Others:**
- drop support for default streams (for better error detection)
- do  not allow moving empty handles into other handles (for better error detection)
- fix handle leak (no bugs yet since the bugged move constructor isn't being used)
- avoid creating shared pointers for empty UniqueHandle objects

```
force_builders=Custom
buildworker:Custom=linux-4
buildworker:iOS=macosx-2
```

### DIFF
--- a/modules/dnn/src/cuda4dnn/csl/cublas.hpp
+++ b/modules/dnn/src/cuda4dnn/csl/cublas.hpp
@@ -52,22 +52,30 @@ namespace cv { namespace dnn { namespace cuda4dnn { namespace csl { namespace cu
         }
     }
 
-    /** noncopyable cuBLAS smart handle
+    /** non-copyable cuBLAS smart handle
      *
      * UniqueHandle is a smart non-sharable wrapper for cuBLAS handle which ensures that the handle
-     * is destroyed after use. The handle can be associated with a CUDA stream by specifying the
-     * stream during construction. By default, the handle is associated with the default stream.
+     * is destroyed after use. The handle must always be associated with a non-default stream. The stream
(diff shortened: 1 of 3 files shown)


YashasSamaga:
I had planned on fixing `getPerfProfile` for the CUDA backend. Currently, the layerwise timings returned by `getPerfProfile` is the time taken for dumping the layer operation into the CUDA stream. It does not tell how much time the operation actually took.

The only way to allow proper `getPerfProfile` result is to place events before and after every layer in the CUDA stream. But this comes at a cost which appears to be on average ~1% (need to redo calculations on idle system) on the inference time. I am not sure what to do.

Many samples and tests use `getPerfProfile`. They all lead to wrong results when the CUDA backend is used.

PR #17350, opened 2020-05-22
Enable stateful kernels in G-API OCV Backend

AsyaPronina:

```
buildworker:iOS=macosx-2

force_builders=ARMv7,Custom
build_image:Custom=centos:7
buildworker:Custom=linux-1

```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
@@ -17,6 +17,7 @@
 #include <opencv2/gapi/gcommon.hpp>
 #include <opencv2/gapi/gkernel.hpp>
 #include <opencv2/gapi/garg.hpp>
+#include <opencv2/gapi/gmetaarg.hpp>
 #include <opencv2/gapi/util/compiler_hints.hpp> //suppress_unused_warning
 #include <opencv2/gapi/util/util.hpp>
 
@@ -109,11 +110,17 @@ public:
         return outOpaqueRef(output).wref<T>();
     }
 
+    GArg state()
+    {
+        return m_state;
+    }
+
 protected:
     detail::VectorRef& outVecRef(int output);
     detail::OpaqueRef& outOpaqueRef(int output);
 
     std::vector<GArg> m_args;
+    GArg m_state;
 
     //FIXME: avoid conversion of arguments from internal representation to OpenCV one on each call
     //to OCV kernel. (This can be achieved by a two single time conversions in GCPUExecutable::run,
@@ -127,16 +134,18 @@ protected:
 class GAPI_EXPORTS GCPUKernel
 {
 public:
-    // This function is kernel's execution entry point (does the processing work)
-    using F = std::function<void(GCPUContext &)>;
+    // This function is a kernel's execution entry point (does the processing work)
+    using RunF = std::function<void(GCPUContext &)>;
+    // This function is a stateful kernel's setup routine (configures state)
+    using SetupF = std::function<void(const GMetaArgs &, const GArgs &, GArg &)>;
 
     GCPUKernel();
-    explicit GCPUKernel(const F& f);
+    GCPUKernel(const RunF& runF, const SetupF& setupF = nullptr);
 
-    void apply(GCPUContext &ctx);
+    RunF m_runF = nullptr;
+    SetupF m_setupF = nullptr;
 
-protected:
-    F m_f;
+    bool m_isStateful = false;
 };
 
 // FIXME: This is an ugly ad-hoc implementation. TODO: refactor
@@ -269,12 +278,38 @@ template<typename U> struct get_out<cv::GOpaque<U>>
     }
 };
 
+template<typename, typename>
+struct OCVSetupHelper;
+
+template<typename Impl, typename... Ins>
+struct OCVSetupHelper<Impl, std::tuple<Ins...>>
+{
+    template<int... IIs>
+    static void setup_impl(const GMetaArgs &metaArgs, const GArgs &args, GArg &state,
+                           detail::Seq<IIs...>)
+    {
+        // TODO: unique_ptr <-> shared_ptr conversion ?
+        // To check: Conversion is possible only if the state which should be passed to
+        // 'setup' user callback isn't required to have previous value
+        std::shared_ptr<typename Impl::State> stPtr;
+        Impl::setup(detail::get_in_meta<Ins>(metaArgs, args, IIs)..., stPtr);
+        state = GArg(stPtr);
+    }
+
+    static void setup(const GMetaArgs &metaArgs, const GArgs &args, GArg& state)
+    {
+        setup_impl(metaArgs, args, state,
+                   typename detail::MkSeq<sizeof...(Ins)>::type());
+    }
+};
+
+// OCVCallHelper is a helper class to call stateless OCV kernels and OCV kernel functors.
 template<typename, typename, typename>
 struct OCVCallHelper;
 
 // FIXME: probably can be simplified with std::apply or analogue.
 template<typename Impl, typename... Ins, typename... Outs>
-struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
+struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...>>
 {
     template<typename... Inputs>
     struct call_and_postprocess
@@ -302,19 +337,16 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
         //by comparing it's state (data ptr) before and after the call.
         //This is done by converting each output Mat into tracked_cv_mat object, and binding
         //them to parameters of ad-hoc function
-        //Convert own::Scalar to cv::Scalar before call kernel and run kernel
-        //convert cv::Scalar to own::Scalar after call kernel and write back results
         call_and_postprocess<decltype(get_in<Ins>::get(ctx, IIs))...>
-                                      ::call(get_in<Ins>::get(ctx, IIs)...,
-                                             get_out<Outs>::get(ctx, OIs)...);
+            ::call(get_in<Ins>::get(ctx, IIs)..., get_out<Outs>::get(ctx, OIs)...);
     }
 
     template<int... IIs, int... OIs>
-    static void call_impl(cv::GCPUContext &ctx, Impl& impl, detail::Seq<IIs...>, detail::Seq<OIs...>)
+    static void call_impl(cv::GCPUContext &ctx, Impl& impl,
+                          detail::Seq<IIs...>, detail::Seq<OIs...>)
     {
-        call_and_postprocess<decltype(cv::detail::get_in<Ins>::get(ctx, IIs))...>
-                                      ::call(impl, cv::detail::get_in<Ins>::get(ctx, IIs)...,
-                                                   cv::detail::get_out<Outs>::get(ctx, OIs)...);
+        call_and_postprocess<decltype(get_in<Ins>::get(ctx, IIs))...>
+            ::call(impl, get_in<Ins>::get(ctx, IIs)..., get_out<Outs>::get(ctx, OIs)...);
     }
 
     static void call(GCPUContext &ctx)
@@ -335,22 +367,77 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
     }
 };
 
+// OCVStCallHelper is a helper class to call stateful OCV kernels.
+template<typename, typename, typename>
+struct OCVStCallHelper;
+
+template<typename Impl, typename... Ins, typename... Outs>
+struct OCVStCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...>> :
+    OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...>>
+{
+    template<typename... Inputs>
+    struct call_and_postprocess
+    {
+        template<typename... Outputs>
+        static void call(typename Impl::State& st, Inputs&&... ins, Outputs&&... outs)
+        {
+            Impl::run(std::forward<Inputs>(ins)..., outs..., st);
+            postprocess(outs...);
+        }
+    };
+
+    template<int... IIs, int... OIs>
+    static void call_impl(GCPUContext &ctx, detail::Seq<IIs...>, detail::Seq<OIs...>)
+    {
+        auto& st = *ctx.state().get<std::shared_ptr<typename Impl::State>>();
+        call_and_postprocess<decltype(get_in<Ins>::get(ctx, IIs))...>
+            ::call(st, get_in<Ins>::get(ctx, IIs)..., get_out<Outs>::get(ctx, OIs)...);
+    }
+
+    static void call(GCPUContext &ctx)
+    {
+        call_impl(ctx,
+                  typename detail::MkSeq<sizeof...(Ins)>::type(),
+                  typename detail::MkSeq<sizeof...(Outs)>::type());
+    }
+};
+
 } // namespace detail
 
 template<class Impl, class K>
-class GCPUKernelImpl: public cv::detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>,
-                      public cv::detail::KernelTag
+class GCPUKernelImpl: public cv::detail::KernelTag
 {
-    using P = detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>;
+    using CallHelper = detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>;
 
 public:
     using API = K;
 
-    static cv::gapi::GBackend backend()  { return cv::gapi::cpu::backend(); }
-    static cv::GCPUKernel     kernel()   { return GCPUKernel(&P::call);     }
+    static cv::gapi::GBackend backend() { return cv::gapi::cpu::backend(); }
+    static cv::GCPUKernel      kernel() { return GCPUKernel(&CallHelper::call); }
+};
+
+template<class Impl, class K, class S>
+class GCPUStKernelImpl: public cv::detail::KernelTag
+{
+    using StSetupHelper = detail::OCVSetupHelper<Impl, typename K::InArgs>;
+    using StCallHelper  = detail::OCVStCallHelper<Impl, typename K::InArgs, typename K::OutArgs>;
+
+public:
+    using API = K;
+    using State = S;
+
+    static cv::gapi::GBackend backend() { return cv::gapi::cpu::backend(); }
+    static cv::GCPUKernel     kernel()  { return GCPUKernel(&StCallHelper::call,
+                                                            &StSetupHelper::setup); }
 };
 
 #define GAPI_OCV_KERNEL(Name, API) struct Name: public cv::GCPUKernelImpl<Name, API>
+
+// TODO: Reuse Anatoliy's logic for support of types with commas in macro.
+//       Retrieve the common part from Anatoliy's logic to the separate place.
+#define GAPI_OCV_KERNEL_ST(Name, API, State)                  \
+    struct Name:public cv::GCPUStKernelImpl<Name, API, State> \
+
 
 class gapi::cpu::GOCVFunctor : public gapi::GFunctor
 {
--- a/modules/gapi/include/opencv2/gapi/gcompiled.hpp
+++ b/modules/gapi/include/opencv2/gapi/gcompiled.hpp
@@ -208,6 +208,19 @@ public:
     // FIXME: Why it requires compile args?
     void reshape(const GMetaArgs& inMetas, const GCompileArgs& args);
 
+    /**
+     * @brief Prepare inner kernels states for a new video-stream.
+     *
+     * GCompiled objects may be used to process video streams frame by frame.
+     * In this case, a GCompiled is called on every image frame individually.
+     * Starting OpenCV 4.4, some kernels in the graph may have their internal
+     * states (see GAPI_OCV_KERNEL_ST for the OpenCV backend).
+     * In this case, if user starts processing another video stream with
+     * this GCompiled, this method needs to be called to let kernels re-initialize
+     * their internal states to a new video stream.
+     */
+    void prepareForNewStream();
+
 protected:
     /// @private
     std::shared_ptr<Priv> m_priv;
--- a/modules/gapi/src/backends/cpu/gcpubackend.cpp
+++ b/modules/gapi/src/backends/cpu/gcpubackend.cpp
@@ -33,13 +33,13 @@
 //
 // If not, we need to introduce that!
 using GCPUModel = ade::TypedGraph
-    < cv::gimpl::Unit
+    < cv::gimpl::CPUUnit
     , cv::gimpl::Protocol
     >;
 
 // FIXME: Same issue with Typed and ConstTyped
 using GConstGCPUModel = ade::ConstTypedGraph
-    < cv::gimpl::Unit
+    < cv::gimpl::CPUUnit
     , cv::gimpl::Protocol
     >;
 
@@ -53,7 +53,7 @@ namespace
         {
             GCPUModel gm(graph);
             auto cpu_impl = cv::util::any_cast<cv::GCPUKernel>(impl.opaque);
-            gm.metadata(op_node).set(cv::gimpl::Unit{cpu_impl});
+            gm.metadata(op_node).set(cv::gimpl::CPUUnit{cpu_impl});
         }
 
         virtual EPtr compile(const ade::Graph &graph,
@@ -78,11 +78,23 @@ cv::gimpl::GCPUExecutable::GCPUExecutable(const ade::Graph &g,
 {
     // Convert list of operations (which is topologically sorted already)
     // into an execution script.
+    GConstGCPUModel gcm(m_g);
     for (auto &nh : nodes)
     {
         switch (m_gm.metadata(nh).get<NodeType>().t)
         {
-        case NodeType::OP: m_script.push_back({nh, GModel::collectOutputMeta(m_gm, nh)}); break;
+        case NodeType::OP:
+        {
+            m_script.push_back({nh, GModel::collectOutputMeta(m_gm, nh)});
+
+            // If kernel is stateful then prepare storage for its state.
+            GCPUKernel k = gcm.metadata(nh).get<CPUUnit>().k;
+            if (k.m_isStateful)
+            {
+                m_nodesToStates[nh] = GArg{ };
+            }
+            break;
+        }
         case NodeType::DATA:
         {
             m_dataNodes.push_back(nh);
@@ -104,6 +116,9 @@ cv::gimpl::GCPUExecutable::GCPUExecutable(const ade::Graph &g,
         default: util::throw_error(std::logic_error("Unsupported NodeType type"));
         }
     }
+
+    // For each stateful kernel call 'setup' user callback to initialize state.
+    setupKernelStates();
 }
 
 // FIXME: Document what it does
@@ -140,6 +155,26 @@ cv::GArg cv::gimpl::GCPUExecutable::packArg(const GArg &arg)
     }
 }
 
+void cv::gimpl::GCPUExecutable::setupKernelStates()
+{
+    GConstGCPUModel gcm(m_g);
+    for (auto& nodeToState : m_nodesToStates)
+    {
+        auto& kernelNode = nodeToState.first;
+        auto& kernelState = nodeToState.second;
+
+        const GCPUKernel& kernel = gcm.metadata(kernelNode).get<CPUUnit>().k;
+        kernel.m_setupF(GModel::collectInputMeta(m_gm, kernelNode),
+                        m_gm.metadata(kernelNode).get<Op>().args,
+                        kernelState);
+    }
+}
+
+void cv::gimpl::GCPUExecutable::handleNewStream()
+{
+    m_newStreamStarted = true;
+}
+
 void cv::gimpl::GCPUExecutable::run(std::vector<InObj>  &&input_objs,
                                     std::vector<OutObj> &&output_objs)
 {
@@ -167,6 +202,14 @@ void cv::gimpl::GCPUExecutable::run(std::vector<InObj>  &&input_objs,
         }
     }
 
+    // In case if new video-stream happens - for each stateful kernel
+    // call 'setup' user callback to re-initialize state.
+    if (m_newStreamStarted)
+    {
+        setupKernelStates();
+        m_newStreamStarted = false;
+    }
+
     // OpenCV backend execution is not a rocket science at all.
     // Simply invoke our kernels in the proper order.
     GConstGCPUModel gcm(m_g);
@@ -176,7 +219,7 @@ void cv::gimpl::GCPUExecutable::run(std::vector<InObj>  &&input_objs,
 
         // Obtain our real execution unit
         // TODO: Should kernels be copyable?
-        GCPUKernel k = gcm.metadata(op_info.nh).get<Unit>().k;
+        GCPUKernel k = gcm.metadata(op_info.nh).get<CPUUnit>().k;
 
         // Initialize kernel's execution context:
         // - Input parameters
@@ -185,8 +228,8 @@ void cv::gimpl::GCPUExecutable::run(std::vector<InObj>  &&input_objs,
 
         using namespace std::placeholders;
         ade::util::transform(op.args,
-                          std::back_inserter(context.m_args),
-                          std::bind(&GCPUExecutable::packArg, this, _1));
+                             std::back_inserter(context.m_args),
+                             std::bind(&GCPUExecutable::packArg, this, _1));
 
         // - Output parameters.
         // FIXME: pre-allocate internal Mats, etc, according to the known meta
@@ -198,8 +241,14 @@ void cv::gimpl::GCPUExecutable::run(std::vector<InObj>  &&input_objs,
             context.m_results[out_port] = magazine::getObjPtr(m_res, out_desc);
         }
 
+        // For stateful kernel add state to its execution context
+        if (k.m_isStateful)
+        {
+            context.m_state = m_nodesToStates.at(op_info.nh);
+        }
+
         // Now trigger the executable unit
-        k.apply(context);
+        k.m_runF(context);
 
         //As Kernels are forbidden to allocate memory for (Mat) outputs,
         //this code seems redundant, at least for Mats
--- a/modules/gapi/src/backends/cpu/gcpubackend.hpp
+++ b/modules/gapi/src/backends/cpu/gcpubackend.hpp
@@ -23,7 +23,7 @@
 
 namespace cv { namespace gimpl {
 
-struct Unit
+struct CPUUnit
 {
     static const char *name() { return "HostKernel"; }
     GCPUKernel k;
@@ -48,6 +48,13 @@ class GCPUExecutable final: public GIslandExecutable
     // Actual data of all resources in graph (both internal and external)
     Mag m_res;
     GArg packArg(const GArg &arg);
+    void setupKernelStates();
+
+    // TODO: Check that it is thread-safe
+    std::unordered_map<ade::NodeHandle, GArg,
+                       ade::HandleHasher<ade::Node>> m_nodesToStates;
+
+    bool m_newStreamStarted = false;
 
 public:
     GCPUExecutable(const ade::Graph                   &graph,
@@ -61,6 +68,8 @@ public:
         // should be dropped.
         util::throw_error(std::logic_error("GCPUExecutable::reshape() should never be called"));
     }
+
+    virtual void handleNewStream() override;
 
     virtual void run(std::vector<InObj>  &&input_objs,
                      std::vector<OutObj> &&output_objs) override;
--- a/modules/gapi/src/backends/cpu/gcpukernel.cpp
+++ b/modules/gapi/src/backends/cpu/gcpukernel.cpp
@@ -45,13 +45,7 @@ cv::GCPUKernel::GCPUKernel()
 {
 }
 
-cv::GCPUKernel::GCPUKernel(const GCPUKernel::F &f)
-    : m_f(f)
+cv::GCPUKernel::GCPUKernel(const GCPUKernel::RunF &runF, const GCPUKernel::SetupF &setupF)
+    : m_runF(runF), m_setupF(setupF), m_isStateful(m_setupF != nullptr)
 {
-}
-
-void cv::GCPUKernel::apply(GCPUContext &ctx)
-{
-    GAPI_Assert(m_f);
-    m_f(ctx);
 }
--- a/modules/gapi/src/executor/gexecutor.cpp
+++ b/modules/gapi/src/executor/gexecutor.cpp
@@ -265,3 +265,11 @@ void cv::gimpl::GExecutor::reshape(const GMetaArgs& inMetas, const GCompileArgs&
     passes::inferMeta(ctx, true);
     m_ops[0].isl_exec->reshape(g, args);
 }
+
+void cv::gimpl::GExecutor::prepareForNewStream()
+{
+    for (auto &op : m_ops)
+    {
+        op.isl_exec->handleNewStream();
+    }
+}
--- a/modules/gapi/src/executor/gstreamingexecutor.cpp
+++ b/modules/gapi/src/executor/gstreamingexecutor.cpp
@@ -800,6 +800,7 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
             }
         }
     };
+    bool islandsRecompiled = false;
     const auto new_meta = cv::descr_of(ins); // 0
     if (gm.metadata().contains<OriginalInputMeta>()) // (1)
     {
@@ -821,6 +822,8 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
             }
             update_int_metas(); // (7)
             m_reshapable = util::make_optional(is_reshapable);
+
+            islandsRecompiled = true;
         }
         else // (8)
         {
@@ -929,7 +932,15 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
         for (auto &&out_eh : op.nh->outNodes()) {
             out_queues.push_back(reader_queues(*m_island_graph, out_eh));
         }
-        op.isl_exec->handleNewStream();
+
+        // If Island Executable is recompiled, all its stuff including internal kernel states
+        // are recreated and re-initialized automatically.
+        // But if not, we should notify Island Executable about new started stream to let it update
+        // its internal variables.
+        if (!islandsRecompiled)
+        {
+            op.isl_exec->handleNewStream();
+        }
 
         m_threads.emplace_back(islandActorThread,
                                op.in_objects,
--- a/modules/gapi/test/cpu/gapi_ocv_stateful_kernel_test_utils.hpp
+++ b/modules/gapi/test/cpu/gapi_ocv_stateful_kernel_test_utils.hpp
@@ -1,0 +1,47 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+//
+// Copyright (C) 2020 Intel Corporation
+
+#ifndef OPENCV_GAPI_OCV_STATEFUL_KERNEL_TESTS_UTILS_HPP
+#define OPENCV_GAPI_OCV_STATEFUL_KERNEL_TESTS_UTILS_HPP
+
+#include "../test_precomp.hpp"
+
+// TODO: Reuse Anatoliy's logic for support of types with commas in macro.
+//       Retrieve the common part from Anatoliy's logic to the separate place.
+#define DEFINE_INITIALIZER(Name, StateType, ...) \
+struct Name                                      \
+{                                                \
+    static StateType value()                     \
+    {                                            \
+       return __VA_ARGS__;                       \
+    }                                            \
+}                                                \
+
+namespace opencv_test
+{
+namespace
+{
+struct UserStruct
+{
+    UserStruct() = default;
+    UserStruct(short myShortVal, float myFloatVal):
+    _myShortVal(myShortVal),
+    _myFloatVal(myFloatVal) { }
+
+    bool operator==(const UserStruct& rhs) const
+    {
+        return ((_myShortVal == rhs._myShortVal) &&
+                (_myFloatVal == rhs._myFloatVal));
+    }
+
+private:
+    short _myShortVal;
+    float _myFloatVal;
+};
+} // anonymous namespace
+} // opencv_test
+
+#endif // OPENCV_GAPI_OCV_STATEFUL_KERNEL_TESTS_UTILS_HPP
--- a/modules/gapi/src/backends/ocl/goclbackend.cpp
+++ b/modules/gapi/src/backends/ocl/goclbackend.cpp
@@ -33,13 +33,13 @@
 //
 // If not, we need to introduce that!
 using GOCLModel = ade::TypedGraph
-    < cv::gimpl::Unit
+    < cv::gimpl::OCLUnit
     , cv::gimpl::Protocol
     >;
 
 // FIXME: Same issue with Typed and ConstTyped
 using GConstGOCLModel = ade::ConstTypedGraph
-    < cv::gimpl::Unit
+    < cv::gimpl::OCLUnit
     , cv::gimpl::Protocol
     >;
 
@@ -53,7 +53,7 @@ namespace
         {
             GOCLModel gm(graph);
             auto ocl_impl = cv::util::any_cast<cv::GOCLKernel>(impl.opaque);
-            gm.metadata(op_node).set(cv::gimpl::Unit{ocl_impl});
+            gm.metadata(op_node).set(cv::gimpl::OCLUnit{ocl_impl});
         }
 
         virtual EPtr compile(const ade::Graph &graph,
@@ -198,7 +198,7 @@ void cv::gimpl::GOCLExecutable::run(std::vector<InObj>  &&input_objs,
 
         // Obtain our real execution unit
         // TODO: Should kernels be copyable?
-        GOCLKernel k = gcm.metadata(op_info.nh).get<Unit>().k;
+        GOCLKernel k = gcm.metadata(op_info.nh).get<OCLUnit>().k;
 
         // Initialize kernel's execution context:
         // - Input parameters
--- a/modules/gapi/src/backends/ocl/goclbackend.hpp
+++ b/modules/gapi/src/backends/ocl/goclbackend.hpp
@@ -23,7 +23,7 @@
 
 namespace cv { namespace gimpl {
 
-struct Unit
+struct OCLUnit
 {
     static const char *name() { return "OCLKernel"; }
     GOCLKernel k;
--- a/modules/gapi/src/backends/render/grenderocvbackend.cpp
+++ b/modules/gapi/src/backends/render/grenderocvbackend.cpp
@@ -93,7 +93,7 @@ void cv::gimpl::render::ocv::GRenderExecutable::run(std::vector<InObj>  &&input_
 
     context.m_args.emplace_back(m_ftpr.get());
 
-    k.apply(context);
+    k.m_runF(context);
 
     for (auto &it : output_objs) magazine::writeBack(m_res, it.first, it.second);
 }
--- a/modules/gapi/src/compiler/gcompiled.cpp
+++ b/modules/gapi/src/compiler/gcompiled.cpp
@@ -72,6 +72,12 @@ void cv::GCompiled::Priv::reshape(const GMetaArgs& inMetas, const GCompileArgs& 
     m_metas = inMetas;
 }
 
+void cv::GCompiled::Priv::prepareForNewStream()
+{
+    GAPI_Assert(m_exec);
+    m_exec->prepareForNewStream();
+}
+
 const cv::gimpl::GModel::Graph& cv::GCompiled::Priv::model() const
 {
     GAPI_Assert(nullptr != m_exec);
@@ -154,4 +160,9 @@ bool cv::GCompiled::canReshape() const
 void cv::GCompiled::reshape(const GMetaArgs& inMetas, const GCompileArgs& args)
 {
     m_priv->reshape(inMetas, args);
+}
+
+void cv::GCompiled::prepareForNewStream()
+{
+    m_priv->prepareForNewStream();
 }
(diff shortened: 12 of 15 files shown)


modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -176,17 +227,28 @@ void cv::gimpl::GCPUExecutable::run(std::vector<InObj>  &&input_objs,
...
+            else
+            {
+                fillContextArgs(op.args, context.m_args);
+                context.m_statePtr = &m_nodesToStates[op_info.nh];
+            }
+        }
+        else
+        {
+            fillContextArgs(op.args, context.m_args);
+        }
```
dmatveev, line 251: This is somewhat complicated.

Instead, you could check your flag once (before the loop), and if you need to reset the state, then walk through the _affected_ nodes only and call the callback.

Please bring that old 6 lines of code back here.

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -224,5 +285,10 @@ void cv::gimpl::GCPUExecutable::run(std::vector<InObj>  &&input_objs,
         }
     } // for(m_script)
 
+    if (m_needKernelsSetup)
+    {
+        m_needKernelsSetup = false;
+    }
```
dmatveev, line 291: Again, all this logic about handling the reset condition can be moved to a separate function. You don't need to spread these details everywhere.

modules/gapi/src/compiler/gislandmodel.hpp
```diff
@@ -123,6 +123,7 @@ class GAPI_EXPORTS GIslandExecutable
     virtual bool canReshape() const = 0;
     virtual void reshape(ade::Graph& g, const GCompileArgs& args) = 0;
 
+    virtual void prepareForNewStream() { }; // do nothing here by default
```
dmatveev, line 126: 😿 
The comment below explains the thing.

modules/gapi/src/executor/gexecutor.cpp
```diff
@@ -265,3 +265,11 @@ void cv::gimpl::GExecutor::reshape(const GMetaArgs& inMetas, const GCompileArgs&
...
     m_ops[0].isl_exec->reshape(g, args);
 }
+
+void cv::gimpl::GExecutor::prepareForNewStream()
+{
+    for (auto &op : m_ops)
+    {
+        op.isl_exec->prepareForNewStream();
+    }
+}
```
dmatveev, line 275: "no newline at the end of file"

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
```diff
@@ -0,0 +1,46 @@
+// This file is part of OpenCV project.
```
dmatveev, line 1: RE file name. There is no such thing as `gapi_statefulkernel`.

It is not a G-API kernel but an OpenCV Backend Kernel.

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
```diff
@@ -0,0 +1,46 @@
...
+};
+
+#define DEFINE_INITIALIZER(Name, StateType, ...)                   \
+struct Name                                                        \
+{                                                                  \
+    static typename get_first_param<void(StateType)>::type value() \
+    {                                                              \
+       return __VA_ARGS__;                                         \
+    }                                                              \
+}                                                                  \
```
dmatveev, line 23: ????

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+namespace
+{
+class UserClass
+{
+public:
+    UserClass() = default;
+    UserClass(short myShortVal, float myFloatVal):
+    _myShortVal(myShortVal),
+    _myFloatVal(myFloatVal) { }
```
dmatveev, line 33: This could be a struct?

modules/gapi/test/gapi_statefulkernel_tests.cpp
```diff
@@ -0,0 +1,201 @@
+// This file is part of OpenCV project.
```
dmatveev, line 1: This test must come to the OCV (CPU) backend directory.

Also please change `statefulkernel` to `stateful_kernel`

dmatveev, line 1: This test must come to the OCV (CPU) backend directory.

Also please change `statefulkernel` to `stateful_kernel`

modules/gapi/test/gapi_statefulkernel_tests.cpp
```diff
@@ -0,0 +1,201 @@
...
+    GMat in;
+    GOpaque<int> out = GCountCalls::on(in);
+    cv::GComputation comp(cv::GIn(in), cv::GOut(out));
+
+    const auto pkg = cv::gapi::kernels<GOCVCountCalls>();
+
+    // Compilation of G-API expression
+    auto callsCounter = comp.compile(cv::descr_of(dummyIn), cv::compile_args(pkg));
+
+    // Streaming
```
dmatveev, line 108: You and me understand what do you mean by `Streaming` here. Others won't.

modules/gapi/test/gapi_statefulkernel_tests.cpp
```diff
@@ -0,0 +1,201 @@
...
+{
+    using StateT = typename std::tuple_element<0, Tuple>::type;
+    using InitT  = typename std::tuple_element<1, Tuple>::type;
+
+    G_TYPED_KERNEL(GReturnState, <cv::GOpaque<StateT>(GMat)>, "org.opencv.test.return_state")
+    {
+        static GOpaqueDesc outMeta(GMatDesc /* in */) { return empty_gopaque_desc(); }
+    };
+
+    GAPI_OCV_STATEFUL_KERNEL(GOCVReturnState, GReturnState, StateT)
```
dmatveev, line 143: So our kernels can be templates? this is cool :)

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
```diff
@@ -0,0 +1,46 @@
...
+        return ((_myShortVal == rhs._myShortVal) &&
+                (_myFloatVal == rhs._myFloatVal));
+    }
+
+private:
+    short _myShortVal;
+    float _myFloatVal;
+};
+} // anonymous namespace
+} // opencv_test
```
dmatveev, line 46: !

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -109,11 +116,21 @@ class GAPI_EXPORTS GCPUContext
...
 protected:
     detail::VectorRef& outVecRef(int output);
     detail::OpaqueRef& outOpaqueRef(int output);
 
-    std::vector<GArg> m_args;
+    std::vector<GMetaArg> m_metaArgs { };
+    std::vector<GArg> m_args { };
+    // Need to use optional reference for sure what isn't allowed in C++17:(
+    // But optional references exist in Boost..
+    GArg* m_statePtr = nullptr;
```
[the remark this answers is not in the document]
AsyaPronina: I thought we don't want GArg with shared_ptr to be copied here?

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -333,25 +387,60 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
...
     using P = detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>;
 
 public:
     using API = K;
+    using State = S;
 
     static cv::gapi::GBackend backend()  { return cv::gapi::cpu::backend(); }
-    static cv::GCPUKernel     kernel()   { return GCPUKernel(&P::call);     }
+
+    template<typename Q = S>
```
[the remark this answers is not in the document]
AsyaPronina: Now I can't, this is how std::enable_if with usual (non-template) method of template class can be used

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -333,25 +387,60 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
...
+    template<int... IIs>
+    static void setup_impl(cv::GCPUContext& ctx, detail::Seq<IIs...>)
+    {
+        setup_impl(ctx.state(), ctx.inMetaArg<Ins>(IIs)...);
+    }
+
+    static void setup(cv::GCPUContext& ctx)
+    {
+        setup_impl(ctx, typename detail::MkSeq<sizeof...(Ins)>::type());
+    }
```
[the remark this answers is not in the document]
AsyaPronina: What about run method? Should we hide GArg state into GCPUContext or not?

modules/gapi/src/backends/cpu/gcpubackend.cpp
AsyaPronina: I walked away from implementation with handling of only affected nodes because thought in the way that runtime information may be needed for setup. If last was the right so we need to retrieve the newest runtime data before the setup call.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -333,25 +387,60 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
...
+
+    template<typename Q = S>
+    static
+    typename std::enable_if<!std::is_void<Q>::value, cv::GCPUKernel>::type
+    kernel()   { return GCPUKernel(&P::setup, &P::call); }
 };
 
 #define GAPI_OCV_KERNEL(Name, API) struct Name: public cv::GCPUKernelImpl<Name, API>
 
+#define GAPI_OCV_STATEFUL_KERNEL(Name, API, State) struct Name: public cv::GCPUKernelImpl<Name, API, State>
```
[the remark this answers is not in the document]
AsyaPronina: Maybe better to GAPI_OCV_ST_KERNEL?

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -333,25 +387,60 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
...
 public:
     using API = K;
+    using State = S;
 
     static cv::gapi::GBackend backend()  { return cv::gapi::cpu::backend(); }
-    static cv::GCPUKernel     kernel()   { return GCPUKernel(&P::call);     }
+
+    template<typename Q = S>
+    static
+    typename std::enable_if<std::is_void<Q>::value, cv::GCPUKernel>::type
```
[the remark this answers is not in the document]
AsyaPronina: I've removed has_state for now because this meta-check is not needed in OCVCallHelper anymore

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -78,11 +78,26 @@ cv::gimpl::GCPUExecutable::GCPUExecutable(const ade::Graph &g,
...
+            m_script.push_back({nh, GModel::collectOutputMeta(m_gm, nh)});
+
+            GCPUKernel k = gcm.metadata(nh).get<Unit>().k;    
+            if (k.m_hasState)
+            {
+                // Initial state setup
+                m_nodesToStates[nh] = GArg { };
+
+                GCPUContext context;
+                setupKernelState(nh, context);
```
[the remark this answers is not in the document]
AsyaPronina: GCPUContext isn't used anymore as information holder for GCPUExecutable and Kernel callbacks communications.
Now all arguments are directly passed from GCPUExecutable right in callbacks.

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -140,6 +155,42 @@ cv::GArg cv::gimpl::GCPUExecutable::packArg(const GArg &arg)
...
 
+void cv::gimpl::GCPUExecutable::fillContextArgs(const GArgs &args, GArgs &ctxArgs)
+{
+    ctxArgs.reserve(args.size());
+
+    using namespace std::placeholders;
+    ade::util::transform(args,
+                         std::back_inserter(ctxArgs),
+                         std::bind(&GCPUExecutable::packArg, this, _1));
+}
```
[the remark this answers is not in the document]
AsyaPronina: Removed

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -127,16 +144,17 @@ class GAPI_EXPORTS GCPUContext
 class GAPI_EXPORTS GCPUKernel
 {
 public:
-    // This function is kernel's execution entry point (does the processing work)
```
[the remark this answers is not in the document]
AsyaPronina: Fixed.

modules/gapi/src/backends/cpu/gcpubackend.hpp
```diff
@@ -62,6 +70,9 @@ class GCPUExecutable final: public GIslandExecutable
         util::throw_error(std::logic_error("GCPUExecutable::reshape() should never be called"));
     }
 
+    virtual void prepareForNewStream() override;
+    virtual void handleNewStream() override;
```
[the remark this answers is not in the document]
AsyaPronina: Removed

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -109,11 +116,21 @@ class GAPI_EXPORTS GCPUContext
...
+        return *m_statePtr;
+    }
+
 protected:
     detail::VectorRef& outVecRef(int output);
     detail::OpaqueRef& outOpaqueRef(int output);
 
-    std::vector<GArg> m_args;
+    std::vector<GMetaArg> m_metaArgs { };
+    std::vector<GArg> m_args { };
```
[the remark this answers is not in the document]
AsyaPronina: These entries are now removed

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -109,11 +116,21 @@ class GAPI_EXPORTS GCPUContext
         return outOpaqueRef(output).wref<T>();
     }
 
+    GArg& state()
+    {
+        GAPI_Assert(m_statePtr);
+        return *m_statePtr;
+    }
```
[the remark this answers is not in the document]
AsyaPronina: This method doesn't exist anymore

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -90,6 +91,12 @@ namespace cpu
 class GAPI_EXPORTS GCPUContext
 {
 public:
+    template<typename T>
+    auto inMetaArg(int input) -> decltype(detail::get_in_meta<T>(std::declval<GMetaArgs>(),
+                                                                 std::declval<GArgs>(),
+                                                                0))
+    { return detail::get_in_meta<T>(m_metaArgs, m_args, input); }
```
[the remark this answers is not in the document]
AsyaPronina: Yes, this method is removed

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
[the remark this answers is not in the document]
AsyaPronina: Btw, state isn't held by GCPUContext anymore

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -269,21 +287,46 @@ template<typename U> struct get_out<cv::GOpaque<U>>
...
 template<typename Impl, typename... Ins, typename... Outs>
-struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
+struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...>>
 {
     template<typename... Inputs>
     struct call_and_postprocess
     {
         template<typename... Outputs>
-        static void call(Inputs&&... ins, Outputs&&... outs)
+        static void call(GArg state, Inputs&&... ins, Outputs&&... outs)
```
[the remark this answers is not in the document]
AsyaPronina: Fixed.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -333,25 +387,60 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
                   typename detail::MkSeq<sizeof...(Ins)>::type(),
                   typename detail::MkSeq<sizeof...(Outs)>::type());
     }
+
+    template<typename... InMetaArgs>
+    static void setup_impl(GArg& state, InMetaArgs... metaArgs)
+    {
+        // unique_ptr <-> shared_ptr conversion ?
+        // To check: Conversion is possible only if the state which should be passes to
```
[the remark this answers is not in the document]
AsyaPronina: Fixed.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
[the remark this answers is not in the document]
AsyaPronina: Thanks a lot! Fixed.
OCVSetupHelper is created, state, metaargs and args are passed directly from GCPUExecutable to OCVSetupHelper::setup callback

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -295,8 +338,19 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
         }
     };
```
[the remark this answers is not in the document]
AsyaPronina: I've removed call_and_postprocess structure but preserved postprocess helper function to call it after Impl::run. Should I remove postprocess(...) call?
I've returned GCPUContext back to its original state and now call_impl versions are just overload of each other, when one requires GCPUContext and state and other one just GCPUContext.

modules/gapi/include/opencv2/gapi/gcompiled.hpp
```diff
@@ -208,6 +208,11 @@ class GAPI_EXPORTS GCompiled
     // FIXME: Why it requires compile args?
     void reshape(const GMetaArgs& inMetas, const GCompileArgs& args);
 
+    /**
+     * @brief Reset for a new stream.
+     */
```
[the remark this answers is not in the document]
AsyaPronina: Fixed.

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -140,6 +155,42 @@ cv::GArg cv::gimpl::GCPUExecutable::packArg(const GArg &arg)
...
+    k.m_setupF(context);
+}
+
+void cv::gimpl::GCPUExecutable::prepareForNewStream()
+{
+    // For future: If we will work with state which should be filled with runtime infromation,
+    //             we should setup our states in runtime ("run") method and here just set the flag
+    m_needKernelsSetup = true;
+}
+
```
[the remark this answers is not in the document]
AsyaPronina: Fixed.

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -140,6 +155,42 @@ cv::GArg cv::gimpl::GCPUExecutable::packArg(const GArg &arg)
...
+                         std::bind(&GCPUExecutable::packArg, this, _1));
+}
+
+void cv::gimpl::GCPUExecutable::setupKernelState(const ade::NodeHandle &kernelNode, GCPUContext& context)
+{
+    GConstGCPUModel gcm(m_g);
+    GCPUKernel k = gcm.metadata(kernelNode).get<Unit>().k;
+
+    context.m_metaArgs = GModel::collectInputMeta(m_gm, kernelNode);
+    fillContextArgs(m_gm.metadata(kernelNode).get<Op>().args, context.m_args);
```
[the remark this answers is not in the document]
AsyaPronina: Thanks a lot! Fixed.

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -140,6 +155,42 @@ cv::GArg cv::gimpl::GCPUExecutable::packArg(const GArg &arg)
...
+{
+    // For future: If we will work with state which should be filled with runtime infromation,
+    //             we should setup our states in runtime ("run") method and here just set the flag
+    m_needKernelsSetup = true;
+}
+
+void cv::gimpl::GCPUExecutable::handleNewStream()
+{
+    // For future: If we will work with state which should be filled with runtime infromation,
+    //             we should setup our states in runtime ("run") method and here just set the flag
```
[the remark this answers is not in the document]
AsyaPronina: Removed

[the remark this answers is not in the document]
AsyaPronina: Removed

modules/gapi/src/backends/cpu/gcpubackend.cpp
AsyaPronina: Implemented handling of only affected nodes and moved this logic upper in the "run" code.

modules/gapi/src/backends/cpu/gcpubackend.cpp
AsyaPronina: I moved this logic upper in the "run" code. Is this acceptable solution or better to move out logic into a separate function?

modules/gapi/src/compiler/gislandmodel.hpp
AsyaPronina: This method is removed from GIslandExecutable.

modules/gapi/src/executor/gexecutor.cpp
AsyaPronina: Thanks! Fixed.

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
AsyaPronina: Renamed to gapi_ocv_kernel_st_test_utils.hpp (name is derived from macro name for OCV Stateful kernel).

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
AsyaPronina: It is needed for easier creation of initializers for different state types.
Particularly get_first_param logic is responsible for possibility of such construction with commas:
DEFINE_INITIALIZER(StdArrayValue, (std::array<int, 3>), { 1, 2, 3 });

modules/gapi/test/gapi_statefulkernel_tests.cpp
AsyaPronina: Moved to cpu directory and renamed to gapi_ocv_kernel_st_tests.cpp ((name is derived from macro name for OCV Stateful kernel).

AsyaPronina: Moved to cpu directory and renamed to gapi_ocv_kernel_st_tests.cpp ((name is derived from macro name for OCV Stateful kernel).

modules/gapi/test/gapi_statefulkernel_tests.cpp
AsyaPronina: Fixed.

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
AsyaPronina: Fixed, thanks!

modules/gapi/test/gapi_statefulkernel_tests.cpp
AsyaPronina: Yes, I also was pleasantly surprised:)

AsyaPronina:
Hi Dmitry, Anatoly (@dmatveev @TolyaTalamanov), could you please check the updates?

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -127,16 +128,28 @@ class GAPI_EXPORTS GCPUContext
...
+    // FIXME: Workaround to hide dangerous bug:
+    //        If we add callbacks here as std::function-s then we will get
+    //        into segfault with "stack smashing detected" error for GPU tests.
+    //        Segfault will be happened on GOCLBackendImpl::unpackKernel exit.
+    //        But, if we comment last statement in GOCLBackendImpl::unpackKernel
+    //        so this issue will not be reproducible.
+    // This function is a stateful kernel's setup routine (configures state)
+    using SetupF = void(*)(GMetaArgs, GArgs, GArg &);
+    // This function is a stateful kernel's execution entry point (does the processing work)
+    using StRunF = void(*)(GCPUContext &, GArg &);
```
dmatveev, line 142: Those should remain be std::functions. The problem is somewhere else. Also, the SetupF should take its `GMetaArgs` and `GArgs` as `const &`.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -127,16 +128,28 @@ class GAPI_EXPORTS GCPUContext
...
 
     GCPUKernel();
-    explicit GCPUKernel(const F& f);
+    explicit GCPUKernel(const RunF& runF);
+    explicit GCPUKernel(const SetupF& setupF, const StRunF& stRunF);
```
dmatveev, line 146: There's no point in `explicit` for this constructor.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
dmatveev: Also, please drop a dedicated StRunF. Why use it at all? `GCPUContext` should be enough.

Let me say it this way. The main reason to wrap everything into a structure like `GCPUContext` is a possible ABI compatibility. While we can extend the information we pass through it, the ABI won't change. In your case it changes already

dmatveev: ...we don't strictly follow the ABI compatibility now, it is mostly FYI why use contexts instead of actual arguments.

dmatveev: ...we don't strictly follow the ABI compatibility now, it is mostly FYI why use contexts instead of actual arguments.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -269,31 +282,52 @@ template<typename U> struct get_out<cv::GOpaque<U>>
...
 };
 
+template<typename, typename>
+struct OCVSetupHelper;
+
+template<typename Impl, typename... Ins>
+struct OCVSetupHelper<Impl, std::tuple<Ins...>>
+{
+    template<int... IIs>
+    static void setup_impl(GMetaArgs metaArgs, GArgs args, GArg &state, detail::Seq<IIs...>)
```
dmatveev, line 292: const&

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -269,31 +282,52 @@ template<typename U> struct get_out<cv::GOpaque<U>>
...
+    {
+        // unique_ptr <-> shared_ptr conversion ?
+        // To check: Conversion is possible only if the state which should be passed to
+        // 'setup' user callback isn't required to have previous value
+        std::shared_ptr<typename Impl::State> stPtr;
+        Impl::setup(detail::get_in_meta<Ins>(metaArgs, args, IIs)...,
+                    stPtr);
```
dmatveev, line 299: This can remain one-liner. Your previous comment is longer than this call to `setup()`.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -269,31 +282,52 @@ template<typename U> struct get_out<cv::GOpaque<U>>
...
+        state = GArg(stPtr);
+    }
+
+    static void setup(GMetaArgs metaArgs, GArgs args, GArg& state)
```
dmatveev, line 303: `const&`

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -269,31 +282,52 @@ template<typename U> struct get_out<cv::GOpaque<U>>
...
-        template<typename... Outputs>
-        static void call(Inputs&&... ins, Outputs&&... outs)
-        {
-            //not using a std::forward on outs is deliberate in order to
-            //cause compilation error, by trying to bind rvalue references to lvalue references
-            Impl::run(std::forward<Inputs>(ins)..., outs...);
-            postprocess(outs...);
-        }
+        ::cv::detail::postprocess(outs...);
+    }
```
dmatveev, line 321: What's the reason to have this?

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -269,31 +282,52 @@ template<typename U> struct get_out<cv::GOpaque<U>>
...
-        template<typename... Outputs>
-        static void call(Impl& impl, Inputs&&... ins, Outputs&&... outs)
-        {
-            impl(std::forward<Inputs>(ins)..., outs...);
-        }
-    };
+    template<int... IIs, int... OIs>
+    static void call_impl(GCPUContext &ctx, GArg &state, detail::Seq<IIs...>, detail::Seq<OIs...>)
+    {
+        auto& unpackedState = *state.get<std::shared_ptr<typename Impl::State>>();
```
dmatveev, line 326: this `st` instead of `unpacketState` the next line can be one-liner too.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -304,17 +338,16 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
         //them to parameters of ad-hoc function
         //Convert own::Scalar to cv::Scalar before call kernel and run kernel
         //convert cv::Scalar to own::Scalar after call kernel and write back results
-        call_and_postprocess<decltype(get_in<Ins>::get(ctx, IIs))...>
-                                      ::call(get_in<Ins>::get(ctx, IIs)...,
-                                             get_out<Outs>::get(ctx, OIs)...);
+        Impl::run(get_in<Ins>::get(ctx, IIs)...,
+                  get_out<Outs>::get(ctx, OIs)...);
```
dmatveev, line 342: one-liner

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -304,17 +338,16 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
...
     }
 
     template<int... IIs, int... OIs>
     static void call_impl(cv::GCPUContext &ctx, Impl& impl, detail::Seq<IIs...>, detail::Seq<OIs...>)
     {
-        call_and_postprocess<decltype(cv::detail::get_in<Ins>::get(ctx, IIs))...>
-                                      ::call(impl, cv::detail::get_in<Ins>::get(ctx, IIs)...,
-                                                   cv::detail::get_out<Outs>::get(ctx, OIs)...);
+        impl(cv::detail::get_in<Ins>::get(ctx, IIs)...,
+             cv::detail::get_out<Outs>::get(ctx, OIs)...);
```
dmatveev, line 350: one-liner

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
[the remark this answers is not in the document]
dmatveev: ok

modules/gapi/include/opencv2/gapi/gcompiled.hpp
```diff
@@ -208,6 +208,14 @@ class GAPI_EXPORTS GCompiled
     // FIXME: Why it requires compile args?
     void reshape(const GMetaArgs& inMetas, const GCompileArgs& args);
 
+    /**
+     * @brief Prepare inner kernels states for a new stream.
```
dmatveev, line 212: It will be great if this comments also explains what the `stream` is.

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -78,11 +78,25 @@ cv::gimpl::GCPUExecutable::GCPUExecutable(const ade::Graph &g,
...
+        {
+            m_script.push_back({nh, GModel::collectOutputMeta(m_gm, nh)});
+
+            GCPUKernel k = gcm.metadata(nh).get<Unit>().k;
+            if (k.m_isStateful)
+            {
+                // Initial state setup
+                m_nodesToStates[nh] = GArg { };
+
+                setupKernelState(nh, m_nodesToStates[nh]);
```
dmatveev, line 96: Two map look-ups instead of one. Please rework.

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -140,6 +154,22 @@ cv::GArg cv::gimpl::GCPUExecutable::packArg(const GArg &arg)
...
+void cv::gimpl::GCPUExecutable::setupKernelState(const ade::NodeHandle &kernelNode,
+                                                 GArg &state)
+{
+    GConstGCPUModel gcm(m_g);
+    GCPUKernel k = gcm.metadata(kernelNode).get<Unit>().k;
+
+    k.m_setupF(GModel::collectInputMeta(m_gm, kernelNode),
+               m_gm.metadata(kernelNode).get<Op>().args,
+               state);
+}
```
dmatveev, line 166: I think it doesn't make much sense here in a function that handles only one kernel, and rewrite it to call this callback for _all_ affected nodes. Then the caller code becomes more clear.

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -199,7 +241,14 @@ void cv::gimpl::GCPUExecutable::run(std::vector<InObj>  &&input_objs,
...
         // Now trigger the executable unit
-        k.apply(context);
+        if (k.m_isStateful)
+        {
+            k.m_stRunF(context, m_nodesToStates[op_info.nh]);
+        }
+        else
+        {
+            k.m_runF(context);
+        }
```
dmatveev, line 251: Instead, you can optionally put a state into context and call the same old entry point.

modules/gapi/src/backends/cpu/gcpubackend.cpp
AsyaPronina: This function is used twice in this code

modules/gapi/src/backends/cpu/gcpubackend.cpp
AsyaPronina: This was done previously when I tried to hide all information for communication between GCPUExecutable and GCPUKernel in GCPUContext.
The reason why I haven't just hide state in GCPUContext is in inconsistent(only from my personal opinion) GCPUKernel interface - ```m_runF``` callback hides all information in GCPUContext for purpose of making stateless kernel call and statefull kernel call similar, when ```m_setupF```  shows all internals (or part of internals - GArgs) out.

That's why I want either to hide all information in contexts or expose all information outside.
I've chosen last option because contexts bring more code lines to fill them.
Also if I will think from GCPUExecutable point of view, then I:
1) accept graph of nodes, where each op node knows whether it is statefull or not;
2) if node is stateful it creates and holds state for it;
3) during run it checks if node is stateful and calls stateful callback
My brain found it as transparent contract between GCPUBackend and GCPUKernel where GCPUBackend doesn't know about some GCPUKernel inner details likes a requirement to fill or not to fill GCPUContext with state.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
dmatveev: Also please put `postprocess` here. I wrote already about that old bug.

modules/gapi/src/backends/cpu/gcpubackend.cpp
dmatveev: The last time you did it wrong, since you reused the same context object for two different types of callbacks (a compile-time one and a run-time one).

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -199,7 +240,12 @@ void cv::gimpl::GCPUExecutable::run(std::vector<InObj>  &&input_objs,
         }
 
         // Now trigger the executable unit
-        k.apply(context);
+        if (k.m_isStateful)
+        {
+            context.m_state = m_nodesToStates[op_info.nh];
```
dmatveev, line 245: Use `.at` here instead of `[]`. Missing a pre-initialized state for a stateful kernel should be fatal.

modules/gapi/src/backends/cpu/gcpukernel.cpp
```diff
@@ -45,13 +45,7 @@ cv::GCPUKernel::GCPUKernel()
...
+    : m_runF(runF), m_setupF(setupF), m_isStateful(m_setupF != nullptr)
 {
-}
-
-void cv::GCPUKernel::apply(GCPUContext &ctx)
-{
-    GAPI_Assert(m_f);
-    m_f(ctx);
-}
+}
```
dmatveev, line 51: No newline

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: The reason is in that that cv::detail::postprocess accepts arguments by reference, so arguments should be lvalues

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: okay, no problem, I thought about another Anatoliy's feature - OSD Render and not Kernel as lambda

modules/gapi/src/backends/cpu/gcpubackend.cpp
AsyaPronina: Thanks a lot!!

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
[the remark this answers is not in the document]
AsyaPronina: Fixed.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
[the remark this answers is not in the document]
AsyaPronina: Fixed.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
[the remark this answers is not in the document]
AsyaPronina: Added state into GCPUContext and differentiated two call_impl versions with std::enable_if SFINAE check.

modules/gapi/src/backends/cpu/gcpubackend.cpp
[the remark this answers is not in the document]
AsyaPronina: Returned state into GCPUContext, but GCPUContext is not used anymore for setup call.

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
AsyaPronina: Fixed.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: Fixed.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: Made it is implicit

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: Fixed.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: Fixed.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: Fixed.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: Fixed.

modules/gapi/include/opencv2/gapi/gcompiled.hpp
AsyaPronina: Please check if new explanation is enough?

modules/gapi/src/backends/cpu/gcpubackend.cpp
AsyaPronina: Fixed, thanks!

modules/gapi/src/backends/cpu/gcpubackend.cpp
AsyaPronina: Done.

modules/gapi/src/backends/cpu/gcpubackend.cpp
AsyaPronina: Done.

modules/gapi/src/backends/cpu/gcpukernel.cpp
AsyaPronina: Thanks! Fixed.

AsyaPronina: Thanks, fixed!

AsyaPronina:
Hi Dmitry (@dmatveev), I've fixed all comments, could you please check the updates?

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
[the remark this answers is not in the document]
dmatveev: not sure if it is better, the `_ST` thing is more regular than some difference-in-the-middle

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
dmatveev: HERE better to put "stateful" instead of just "st". Looking at the directory listing, go figure out what this file is about?

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
dmatveev: Sounds over-designish, but lets have a look how it looks like in tests.

modules/gapi/test/gapi_statefulkernel_tests.cpp
dmatveev: File names should be descriptive. The point of _ST in the macro was that people have to type it when write their kernels. File names are more read than written.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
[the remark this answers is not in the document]
AsyaPronina: Okay.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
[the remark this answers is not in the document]
AsyaPronina: Added state into GCPUContext and differentiated two call_impl versions with different holding classes.

modules/gapi/test/gapi_statefulkernel_test_utils.hpp
AsyaPronina: Fixed.

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_test_utils.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#include "../test_precomp.hpp"
+
+// TODO: Reuse Anatoliy's logic for support of types with commas in macro.
+//       Retrieve the common part from Anatoliy's logic to the separate place.
+namespace opencv_test
+{
+template<typename T> struct get_first_param;
+template<typename R, typename P1> struct get_first_param<R(P1)> {
+  typedef P1 type;
```
TolyaTalamanov, line 18: `using` ?

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -929,7 +930,11 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
         for (auto &&out_eh : op.nh->outNodes()) {
             out_queues.push_back(reader_queues(*m_island_graph, out_eh));
         }
-        op.isl_exec->handleNewStream();
+
+        if (containsOriginalMeta)
+        {
+            op.isl_exec->handleNewStream();
+        }
```
dmatveev, line 937: Not sure if this is correct.

Then, calling this method only on `containsOriginalMeta` is also hard to explain. There's a little logic for a reader to understand why do we need to call this at this point.

The logic remains the same - internal states are initialized automatically when Island Executables are recompiled, so no need to call `handleNewStream()` there. It only makes sense to call it when we don't recompile the streams and update meta.

It also needs to be called when the new stream's meta doesn't change and there's no need to reshape, but the stream is still new.

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -800,8 +800,9 @@ void cv::gimpl::GStreamingExecutor::setSource(GRunArgs &&ins)
             }
         }
     };
+    bool containsOriginalMeta = gm.metadata().contains<OriginalInputMeta>();
```
dmatveev, line 803: Not sure if this flag is required, see below.

modules/gapi/src/executor/gstreamingexecutor.cpp
dmatveev: Summarizing, you need to call this method here only when your Islands WERE NOT recompiled. `containsOriginalMeta` doesn't capture this in all cases. Please introduce an explicit flag indicating your islands were recompiled by the code above, and run this call only if they weren't.

modules/gapi/include/opencv2/gapi/gcompiled.hpp
AsyaPronina: Fixed with your help, thanks a lot! Just added small corrections to your text, please check if you agree with them

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_test_utils.hpp
AsyaPronina: Fixed!

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -269,32 +278,43 @@ template<typename U> struct get_out<cv::GOpaque<U>>
...
+struct OCVSetupHelper;
+
+template<typename Impl, typename... Ins>
+struct OCVSetupHelper<Impl, std::tuple<Ins...>>
+{
+    template<int... IIs>
+    static void setup_impl(const GMetaArgs &metaArgs, const GArgs &args, GArg &state,
+                           detail::Seq<IIs...>)
+    {
+        // unique_ptr <-> shared_ptr conversion ?
```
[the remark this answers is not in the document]
AsyaPronina: Yes, thanks, it was my bad, fixed!

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -335,23 +356,65 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
     }
 };
 
+template<typename, typename, typename>
+struct OCVStCallHelper;
```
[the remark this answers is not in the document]
AsyaPronina: Please check comments on OCVCallHelper and OCVStCallHelper

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -335,23 +356,65 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
...
 
-template<class Impl, class K>
+template<class Impl, class K, class S = void>
 class GCPUKernelImpl: public cv::detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>,
                       public cv::detail::KernelTag
 {
-    using P = detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>;
+    using StSetupHelper = detail::OCVSetupHelper<Impl, typename K::InArgs>;
+    using CallHelper = detail::OCVCallHelper<Impl, typename K::InArgs, typename K::OutArgs>;
+    using StCallHelper = detail::OCVStCallHelper<Impl, typename K::InArgs, typename K::OutArgs>;
```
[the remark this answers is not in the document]
AsyaPronina: Done.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -335,23 +356,65 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
...
+    typename std::enable_if<!std::is_void<Q>::value, cv::GCPUKernel>
+    ::type kernel() { return GCPUKernel(&StCallHelper::call, &StSetupHelper::setup); }
 };
 
 #define GAPI_OCV_KERNEL(Name, API) struct Name: public cv::GCPUKernelImpl<Name, API>
 
+// TODO: Reuse Anatoliy's logic for support of types with commas in macro.
+//       Retrieve the common part from Anatoliy's logic to the separate place.
+#define GAPI_OCV_KERNEL_ST(Name, API, State)                \
+    struct Name:public cv::GCPUKernelImpl<Name, API, State> \
```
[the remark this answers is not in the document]
AsyaPronina: Done.

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -140,6 +154,30 @@ cv::GArg cv::gimpl::GCPUExecutable::packArg(const GArg &arg)
...
+void cv::gimpl::GCPUExecutable::setupKernelState(const ade::NodeHandle &kernelNode,
+                                                 GArg &state)
+{
+    GConstGCPUModel gcm(m_g);
+    GCPUKernel k = gcm.metadata(kernelNode).get<CPUUnit>().k;
+
+    k.m_setupF(GModel::collectInputMeta(m_gm, kernelNode),
+               m_gm.metadata(kernelNode).get<Op>().args,
+               state);
+}
```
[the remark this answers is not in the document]
AsyaPronina: Done.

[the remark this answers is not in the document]
AsyaPronina: Done.

modules/gapi/src/executor/gstreamingexecutor.cpp
AsyaPronina: Fixed, thanks a lot!

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -335,23 +370,80 @@ struct OCVCallHelper<Impl, std::tuple<Ins...>, std::tuple<Outs...> >
     }
 };
 
+// OCVStCallHelper is a helper class for call of stateful OCV kernels.
```
dmatveev, line 370:
```diff
- for call of stateful OCV kernels.
+ to call stateful OCV kernels.
```

dmatveev: same applies to other class comments.

modules/gapi/include/opencv2/gapi/gcompiled.hpp
```diff
@@ -208,6 +208,19 @@ class GAPI_EXPORTS GCompiled
...
 
+    /**
+     * @brief Prepare inner kernels states for a new video-stream.
+     *
+     * GCompiled objects may be used to process video streams frame by frame.
+     * In this case, a GCompiled is called on every image frame individually.
+     * Starting OpenCV 4.4, some kernels in the graph may have their internal
+     * states (see GAPI_OCV_KERNEL_ST for the OpenCV backend).
+     * In this case, if user starts processing another video stream with this
+     * GCompiled, method above needs to be called to let kernels re-initialize
```
dmatveev, line 219: "Method above" is reshape here. Why not just "this method"?

modules/gapi/src/backends/cpu/gcpubackend.cpp
```diff
@@ -140,6 +155,26 @@ cv::GArg cv::gimpl::GCPUExecutable::packArg(const GArg &arg)
...
 
+void cv::gimpl::GCPUExecutable::setupKernelStates()
+{
+    GConstGCPUModel gcm(m_g);
+    for (auto& nodeToState : m_nodesToStates)
+    {
+        auto& kernelNode = nodeToState.first;
+        auto& kernelState = nodeToState.second;
+
+        GCPUKernel kernel = gcm.metadata(kernelNode).get<CPUUnit>().k;
```
dmatveev, line 166: &

modules/gapi/src/executor/gstreamingexecutor.cpp
dmatveev: Ideally some comment should occur there to describe this new logic. You can reuse my text above.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: Done.

modules/gapi/include/opencv2/gapi/gcompiled.hpp
AsyaPronina: Rewritten to "this method". What is about "re-initialize states to new stream" - should here be "for" instead of "to" or it doesn't matter actually here?

modules/gapi/src/backends/cpu/gcpubackend.cpp
dmatveev: yes

PR #17363, opened 2020-05-24
cuda4dnn(conv): fuse eltwise with convolutions

YashasSamaga:
## Changes:

- use `device::size_type` and `device::index_type` in `Span`
- refactor functors
  - each functor has a `Params` member type (replacement for varidiac templates)
- fuses eltwise with convolution for three topologies (shown below)
- cuDNN fused convolution when actiation is ReLU and is applied after the eltwise operation

<cut/>

**activation(conv + eltwise)**

Examples: ResNet, ...

```
--------------               ---------------
| Bias Layer |               | Convolution |
--------------               ---------------
      |                             |
      |       ----------------      |
      --------| Eltwise Sum  |-------
              ----------------
                     |
              ----------------
              |  Activation  |                 
              ----------------
                     |
```

**eltwise + activation(conv)**
Examples: YOLOv3, YOLOv4, ....

```
--------------               ---------------
| Bias Layer |               | Convolution |
--------------               ---------------
      |                             |
      |                      ----------------
      |                      |  Activation  |                 
      |                      ----------------
      |                             |
      |       ----------------      |
      --------| Eltwise Sum  |-------
              ----------------
                     |
             
```

**conv + eltwise**
```
--------------               ---------------
| Bias Layer |               | Convolution |
--------------               ---------------
      |                             |
      |       ----------------      |
      --------| Eltwise Sum  |-------
              ----------------
                     |
           
```

---

### Benchmarks

**Device:** RTX 2080 Ti

Model        | Mode          | Time (without this PR)  | Time (with this PR)
-------------- | --------------- | ------------------------------ | --------------------------
YOLOv3    | FP32          | 17.44ms                        | 16.92ms
YOLOv3    | FP16          | 7.87ms                          | 7.71ms
YOLOv4    | FP32          | 18.01ms                        | 17.64ms
YOLOv4    | FP16          | 9.90ms                          | 9.68ms

608  x 608 input for YOLOv3 and YOLOv4.

Timings for YOLOv4 does not include NMS.

**Darknet Reference:**

Model        | Mode          | Time
-------------- | --------------- | -----------
YOLOv4    | FP32          | 24.8ms
YOLOv4    | FP16          | 20.6ms

Average of five runs of timings reported by `./darknet detector test cfg/coco.data cfg/yolov4.cfg yolov4.weights data/dog.jpg
`

---

**Pending:**
- need some test which checks all possible fusion combinations

```
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/dnn/src/cuda/activation_eltwise.cu
+++ b/modules/dnn/src/cuda/activation_eltwise.cu
@@ -1,0 +1,121 @@
+// This file is part of OpenCV project.
+// It is subject to the license terms in the LICENSE file found in the top-level directory
+// of this distribution and at http://opencv.org/license.html.
+
+#include <cuda_runtime.h>
+#include <cuda_fp16.h>
+
+#include "functors.hpp"
+#include "vector_traits.hpp"
+#include "grid_stride_range.hpp"
+#include "execution.hpp"
+
+#include "../cuda4dnn/csl/stream.hpp"
+#include "../cuda4dnn/csl/span.hpp"
+
+using namespace cv::dnn::cuda4dnn::csl;
+using namespace cv::dnn::cuda4dnn::csl::device;
+
+namespace cv { namespace dnn { namespace cuda4dnn { namespace kernels {
+
+namespace raw {
+
+    template <class T, class ActivationOp, class EltwiseOp, std::size_t N>
+    __global__ void generic_op_eltwise_op_inplace_vec(Span<T> inplace_output, View<T> eltwise, const typename ActivationOp::Params act_params, const typename EltwiseOp::Params eltwise_params) {
+        using vector_type = get_vector_type_t<T, N>;
+
+        auto inplace_output_vPtr = vector_type::get_pointer(inplace_output.data());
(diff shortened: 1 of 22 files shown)


YashasSamaga:
There is a small performance regression (~8% for ResNet 50) on some devices in FP16 mode. cuDNN's fused convolution is slower than unfused convolution. Since cuDNN is closed source, it's not possible to tell in what situations or devices fused convolution is slower.

**Possible options:**
1. Ignore the regression and let #16900 handle it
2. Disable cuDNN fused convolution for FP16 here and enable it in #16900

alalek:
Add an option to control this fusing, enable by default. Something like:
```
utils::getConfigurationParameterBool("OPENCV_DNN_CUDA_CONVOLUTION_FUSE", true)
```

YashasSamaga:
It turns out that fusing bias addition or eltwise with tensor core based convolutions is slower than the unfused tensor core convolution. Only half precision target uses tensor core based convolutions.

So now the fusion is only enabled in fp16 target for devices with CC lower than 7.0.

PR #17377, opened 2020-05-25
pkg-config: Drop broken include path

jtojnar:
The legacy `includedir` was removed in 4.0 (https://github.com/opencv/opencv/pull/12477) but it was not removed from the `.pc` file.

The non-existing path causes issues when trying to link against OpenCV using pkg-config in CMake.

### DIFF
--- a/cmake/templates/opencv-XXX.pc.in
+++ b/cmake/templates/opencv-XXX.pc.in
@@ -3,12 +3,11 @@
 prefix=@prefix@
 exec_prefix=@exec_prefix@
 libdir=@libdir@
-includedir_old=@includedir@/opencv
-includedir_new=@includedir@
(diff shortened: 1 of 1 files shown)


jtojnar:
For completeness, this is the CMake error:

```
CMake Error in CMakeLists.txt:
  Imported target "CImg::CImg" includes non-existent path

    "/nix/store/l4ggl5zbavhd4arcdi7ba4ibr6far22k-opencv-4.3.0/include/opencv4/opencv"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

PR #17388, opened 2020-05-26
dnn: update network dump code, include ngraph serialization

alalek:
relates #17382

<cut/>

```

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*
```

### DIFF
--- a/modules/dnn/src/ie_ngraph.cpp
+++ b/modules/dnn/src/ie_ngraph.cpp
@@ -6,6 +6,9 @@
 // Third party copyrights are property of their respective owners.
 
 #include "precomp.hpp"
+
+#include <fstream>
+
 #include "ie_ngraph.hpp"
 
 #include <opencv2/dnn/shape_utils.hpp>
@@ -21,6 +24,8 @@
 namespace cv { namespace dnn {
 
 #ifdef HAVE_DNN_NGRAPH
+
+static bool DNN_IE_SERIALIZE = utils::getConfigurationParameterBool("OPENCV_DNN_IE_SERIALIZE", false);
 
 // For networks with input layer which has an empty name, IE generates a name id[some_number].
 // OpenCV lets users use an empty input name and to prevent unexpected naming,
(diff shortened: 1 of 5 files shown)


modules/dnn/src/ie_ngraph.cpp
```diff
@@ -440,9 +448,27 @@ void InfEngineNgraphNet::init(Target targetId)
             ngraph_function->validate_nodes_and_infer_types();
         }
         cnn = InferenceEngine::CNNNetwork(ngraph_function);
-#ifdef _DEBUG  // TODO
-        //cnn.serialize("/tmp/cnn.xml", "/tmp/cnn.bin");
+
+        if (DNN_IE_SERIALIZE)
+        {
+#ifndef OPENCV_DNN_DISABLE_NETWORK_AUTO_DUMP
```
dkurt, line 454: Where this is defined? And what is default value?

alalek: It is not defined anywhere in OpenCV, but can be passed through build flags.
Purpose is custom builds which want to turn off network dumps.

dkurt: I just though that `AUTO_DUMP` means to dump networks by default, without flags. If it's build flag, that means that if `OPENCV_DNN_DISABLE_NETWORK_AUTO_DUMP` is not specified but `OPENCV_DNN_IE_SERIALIZE` is ON we cannot dump the network?

modules/dnn/src/ie_ngraph.cpp
```diff
@@ -440,9 +448,27 @@ void InfEngineNgraphNet::init(Target targetId)
...
+            std::string dumpFileNameBase = netImpl_.getDumpFileNameBase();
+            try
+            {
+                cnn.serialize(dumpFileNameBase + "_ngraph.xml", dumpFileNameBase + "_ngraph.bin");
```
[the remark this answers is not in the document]
dkurt: Looks like a bug in Inference Engine. Have you checked an accuracy after a single `serialize`? Maybe it breaks even inference.

modules/dnn/src/ie_ngraph.cpp
alalek: Naming is aligned with [this code](https://github.com/opencv/opencv/blame/master/modules/dnn/src/dnn.cpp#L3416-L3428).

Idea is compile-time blocker to dump/serialize network.

dkurt: So, basically my motivation is to have this option available in OpenCV from OpenVINO distribution. Sometimes to enable models in OpenVINO it's helpful to have this method. Will this build flag be available in OpenVINO build by default or we can miss it?

alalek: This code is enabled by default, including OpenVINO builds.

dkurt: Oh, I see now, sorry. `#ifndef` and `DISABLE`. Thanks!

dkurt:
Faced the following error:
```
  what():  OpenCV(3.4.10-dev) /home/dkurt/opencv/modules/core/src/system.cpp:2026: error: (-5:Bad argument) Invalid value for parameter OPENCV_DNN_IE_SERIALIZE: ./model.xml in function 'read'

Aborted (core dumped)
```
when
```
export OPENCV_DNN_IE_SERIALIZE=./model.xml
```

PR #17424, opened 2020-05-29
dnn: add a human parsing cpp sample

berak:

should this go into 3.4 or master ?

### DIFF
--- a/samples/dnn/human_parsing.cpp
+++ b/samples/dnn/human_parsing.cpp
@@ -1,0 +1,104 @@
+//
+// this sample demonstrates parsing (segmenting) human body parts from an image using opencv's dnn,
+// based on https://github.com/Engineering-Course/LIP_JPPNet
+//
+// get the pretrained model from: https://www.dropbox.com/s/qag9vzambhhkvxr/lip_jppnet_384.pb?dl=0
+//
+
+#include <opencv2/dnn.hpp>
+#include <opencv2/highgui.hpp>
+#include <opencv2/imgproc.hpp>
+using namespace cv;
+
+
+static Mat parse_human(const Mat &image, const std::string &model, int backend=dnn::DNN_BACKEND_DEFAULT, int target=dnn::DNN_TARGET_CPU) {
+    // this network expects an image and a flipped copy as input
+    Mat flipped;
+    flip(image, flipped, 1);
+    std::vector<Mat> batch;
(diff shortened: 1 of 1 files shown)


samples/dnn/human_parsing.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include <opencv2/core.hpp>
+#include <opencv2/dnn.hpp>
+#include <opencv2/highgui.hpp>
+#include <opencv2/imgproc.hpp>
+#include <opencv2/imgcodecs.hpp>
+using namespace cv;
+
+static Mat parse_human(const std::string &image, const std::string &model, int backend=dnn::DNN_BACKEND_DEFAULT, int target=dnn::DNN_TARGET_CPU) {
+    // this network expects an image and a flipped copy
+    Mat input = imread(image);
```
l-bat, line 17: I think it’s better to read image in `main()` because we want to use `parse_human()` in other samples

samples/dnn/human_parsing.cpp
```diff
@@ -0,0 +1,100 @@
...
+// based on https://github.com/Engineering-Course/LIP_JPPNet
+//
+// get the pretrained model from: https://www.dropbox.com/s/qag9vzambhhkvxr/lip_jppnet_384.pb?dl=0
+//
+
+#include <opencv2/core.hpp>
+#include <opencv2/dnn.hpp>
+#include <opencv2/highgui.hpp>
+#include <opencv2/imgproc.hpp>
+#include <opencv2/imgcodecs.hpp>
```
l-bat, line 12: Why do we need to include `opencv2/core.hpp` and `opencv2/imgcodecs.hpp`?

samples/dnn/human_parsing.cpp
```diff
@@ -0,0 +1,100 @@
...
+        return 0;
+    }
+    std::string model = parser.get<std::string>("model");
+    std::string image = parser.get<std::string>("image");
+    int backend = parser.get<int>("backend");
+    int target = parser.get<int>("target");
+
+    Mat segm = parse_human(image, model, backend, target);
+
+    imwrite("human.png", segm);
```
l-bat, line 96: `imshow` is enough, `imwrite`  is not needed

l-bat:
I think this PR should go into 3.4

PR #17430, opened 2020-05-29
videoio(ffmpeg): fix handling of AVERROR_EOF

### LINKED ISSUE #17257: Problems with correctly reading video frames
- OpenCV => 4.2
- opencv-contrib-python 4.2.0.34
- opencv-python         4.2.0.34

##### Detailed description

Video is 06.933 sec long with 25 FPS, what makes 173 frames (observed in video player)

However :

```
cap = cv2.VideoCapture(video_path)
cap.get(cv2.CAP_PROP_FRAME_COUNT)
```
gives 172 frames. And while counting manual:
```
while(cap.isOpened()):
        ret, frame = cap.read()
        if ret == True:
            frame_index += 1
        else: 
            break
```

it can't go any further than 168 frame

Sadly I can't share the video file

alalek:
decoder should be properly flushed after that.

resolves #17257

Windows wrapper will be updated later (~2 weeks).

### DIFF
--- a/modules/videoio/src/cap_ffmpeg_impl.hpp
+++ b/modules/videoio/src/cap_ffmpeg_impl.hpp
@@ -1215,9 +1215,20 @@ bool CvCapture_FFMPEG::grabFrame()
 #endif
 
         int ret = av_read_frame(ic, &packet);
-        if (ret == AVERROR(EAGAIN)) continue;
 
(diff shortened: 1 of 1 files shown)


xuguozhi:
Thanks~ This PR saves me

PR #17431, opened 2020-05-30
Added VTK 9 support

mshabunin:
**Merge with contrib**: opencv/opencv_contrib#2549

~Note: the code builds but does not work, additional investigation required~


cmake/OpenCVDetectVTK.cmake
```diff
@@ -1,3 +1,21 @@
...
+    IOGeometry
+    IOPLY
+    InteractionStyle
+    RenderingCore
+    RenderingLOD
+    RenderingOpenGL2
+    NO_MODULE)
+if(VTK_FOUND)
+  set(HAVE_VTK ON)
+  return()
```
alalek, line 16: I believe we still need `try_compile()` test with `checks/vtk_test.cpp` - see below.

mshabunin: Done

PR #17436, opened 2020-05-30
fixed #17044

### LINKED ISSUE #17044: python code in tutorial doesn't work

https://answers.opencv.org/question/228751/python-code-in-tutorial-file-input-and-output-using-xml-and-yaml-files/

vpisarev:
1. fixed Python part of the tutorial about using OpenCV XML-YAML-JSON I/O functionality from C++ and Python.
2. added startWriteStruct() and endWriteStruct() methods to FileStorage
3. modifed FileStorage::write() methods to make them work well inside sequences, not only mappings.

it seems that in master branch just the tutorial should be fixed and CV_WRAP should be added to already existing startWriteStruct() and endWriteStruct() declarations in FileStorage class. Then everything works out of the box. But I can check it as soon as the patch gets merged into master and submit another master-specific patch, if needed.

- [x] There is accuracy test, performance test and test data in opencv_extra repository, if applicable patch to opencv_extra has the same branch name.


alalek:
It would be nice to have Python automatic test (which writes FileStorage into string and check its content).

vpisarev:
done

PR #17438, opened 2020-05-31
core: fix builds with eigen helper header

### LINKED ISSUE #17366: Eigen headers not included in eigen.hpp

- OpenCV => 4.3.0_2
- Operating System / Platform => macOS 10.15
- Compiler => clang

##### Detailed description

When `#include`ing `core/eigen.hpp`, compilation fails, unless `<Eigen/Dense>` or something is included before. Otherwise, compilation fails, because 
```
/usr/local/Cellar/opencv/4.3.0_2/include/opencv4/opencv2/core/eigen.hpp:63:22: error: use of undeclared identifier 'Eigen'
void eigen2cv( const Eigen::Matrix<_Tp, _rows, _cols, _options, _maxRows, _maxCols>& src, OutputArray dst )
```

Am I missing something here? Is this expected? In that case, this should go into the documentation. Otherwise, `#include <Eigen/Dense>` should be added to https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/eigen.hpp

alalek:
- fixes #17366
- [fixes](http://pullrequest.opencv.org/buildbot/builders/3_4_noICV-lin64-clang/builds/595) `warning: 'const' qualifier on reference type 'cv::InputArray' (aka 'const cv::_InputArray &') has no effect`
- [fixes](https://ocv-power.imavr.com/#/builders/9/builds/8727) requirement of C++11: `error: 'array' in namespace 'std' does not name a template type`

relates #17320

```
force_builders=Custom
buildworker:Custom=linux-1
build_image:Custom=ubuntu-clang:18.04
```

### DIFF
--- a/modules/core/include/opencv2/core/eigen.hpp
+++ b/modules/core/include/opencv2/core/eigen.hpp
@@ -45,9 +45,14 @@
 #ifndef OPENCV_CORE_EIGEN_HPP
 #define OPENCV_CORE_EIGEN_HPP
 
+#ifndef EIGEN_WORLD_VERSION
+#error "Wrong usage of OpenCV's Eigen utility header. Include Eigen's headers first. See https://github.com/opencv/opencv/issues/17366"
+#endif
+
 #include "opencv2/core.hpp"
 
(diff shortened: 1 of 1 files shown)


asmorkalov:
:+1:

PR #17454, opened 2020-06-02
fix instable fisheye undistortPoints

creinders:
Fisheye undistorting points is unstable in configurations with larger distortion coefficients. Lets assume a standard situation of undistorting  images and redistorting it with new fisheye distortion coefficients.

<cut/>

The following test image from Cityscapes datasets is used
image

### Current Implementation

The following settings show that the current implementation is instable for larger distortion coefficients.

Distortion Parameters 1: [3.0, 0.0, 0.0, 0.0]
distorted_original_A

Distortion Parameters 2: [9.0, 0.0, 0.0, 0.0]
distorted_original_B

### Improvement A: Filter noise

The current implementation shows a lot of noise outside the actual image. To fix this, the optimization can be improved by checking the convergence and returning an invalid point otherwise.

Results of improvement A:
Distortion Parameters 1: [3.0, 0.0, 0.0, 0.0]
distorted_imp1_A
Distortion Parameters 2: [9.0, 0.0, 0.0, 0.0]
distorted_imp1_B

### Improvement B: Check flipped theta

As visible, the noise disappeared but another problem is visible: Inverted patches of the original image appear in the distorted images. During the optimization, theta_fix is very large. Generally, theta is monotonously increasing or decreasing depending on its sign. In the mentioned areas, theta converges due to the symmetry, however, on the opposite of the camera center. So we can check whether theta has changed the sign during the optimization

Results of improvement B (without improvement A)
Distortion Parameters 1: [3.0, 0.0, 0.0, 0.0]
distorted_imp2_A

Distortion Parameters 2: [9.0, 0.0, 0.0, 0.0]
distorted_imp2_B

### Combination of A and B

Combining both improvements allows for removing all the mentioned artifacts:
Distortion Parameters 1: [3.0, 0.0, 0.0, 0.0]
distorted_imp12_A

Distortion Parameters 2: [9.0, 0.0, 0.0, 0.0]
distorted_imp12_B

###  Sample Code in Python

``` python
import os
import numpy as np
import cv2


def undistort2distort(
        img_src,
        K_src,
        dist_coeffs_src,
        K_dst, dist_coeffs_dst,
):
    # Generate an array of ids
    width = img_src.shape[1]
    height = img_src.shape[0]

    iv, jv = np.meshgrid(range(0, width), range(0, height), sparse=False, indexing='xy')

    iv = iv.flatten()
    jv = jv.flatten()

    # for iP in range(1, iv.length):
    npts = iv.shape[0]

    # convert to Nx1x2 (fisheye)
    img_px_pts = np.zeros((npts, 1, 2), dtype=np.float32)
    img_px_pts[:, 0, 0] = iv
    img_px_pts[:, 0, 1] = jv

    # undistort destination points
    undPts = cv2.fisheye.undistortPoints(img_px_pts, K_dst, dist_coeffs_dst)

    # distort destination points to the source
    distPts = cv2.fisheye.distortPoints(undPts, K_src, dist_coeffs_src)

    # build a map
    x_pts = np.reshape(distPts[:, 0, 0], (height, width))
    y_pts = np.reshape(distPts[:, 0, 1], (height, width))

    img_dst = cv2.remap(img_src, x_pts, y_pts, interpolation=cv2.INTER_LINEAR, borderMode=cv2.BORDER_CONSTANT)

    return img_dst


# Calibration data

K_src = np.array([[2245, 0.00000000e+00, 2048 / 2],
                  [0.00000000e+00, 2245, 1024 / 2],
                  [0.00000000e+00, 0.00000000e+00, 1.00000000e+00]])
dist_coeffs_src = np.zeros((1, 4))

K_dst = K_src.copy()
dist_coeffs_dst = np.array([[-3.0, 0.0, 0.0, 0.0]])  # setting 1
# dist_coeffs_dst = np.array([[-9.0, 0.0, 0.0, 0.0]])  # setting 2
# dist_coeffs_dst = np.array([[-20.0, 0.0, 0.0, 0.0]])  # setting 3

# image to distort
img = cv2.imread('image.png')

# distort
img_distorted = undistort2distort(img, K_src, dist_coeffs_src, K_dst, dist_coeffs_dst)

# save
output_image_name = os.path.join("distorted.jpg")
cv2.imwrite(output_image_name, img_distorted)
```

```
force_builders=linux,docs,Win64,Mac
allow_multiple_commits=1
```

### DIFF
--- a/modules/calib3d/src/fisheye.cpp
+++ b/modules/calib3d/src/fisheye.cpp
@@ -411,8 +411,8 @@ void cv::fisheye::undistortPoints( InputArray distorted, OutputArray undistorted
             }
 
             scale = std::tan(theta) / theta_d;
-        } 
-        else 
+        }
+        else
         {
             converged = true;
         }
--- a/modules/calib3d/src/fisheye.cpp
+++ b/modules/calib3d/src/fisheye.cpp
@@ -377,8 +377,6 @@ void cv::fisheye::undistortPoints( InputArray distorted, OutputArray undistorted
         Vec2d pi = sdepth == CV_32F ? (Vec2d)srcf[i] : srcd[i];  // image point
         Vec2d pw((pi[0] - c[0])/f[0], (pi[1] - c[1])/f[1]);      // world point
 
-        double scale = 1.0;
-
         double theta_d = sqrt(pw[0]*pw[0] + pw[1]*pw[1]);
 
         // the current camera model is only valid up to 180 FOV
@@ -386,12 +384,15 @@ void cv::fisheye::undistortPoints( InputArray distorted, OutputArray undistorted
         // clip values so we still get plausible results for super fisheye images > 180 grad
         theta_d = min(max(-CV_PI/2., theta_d), CV_PI/2.);
 
+        bool converged = false;
+        double theta = theta_d;
+
         if (theta_d > 1e-8)
         {
             // compensate distortion iteratively
-            double theta = theta_d;
 
             const double EPS = 1e-8; // or std::numeric_limits<double>::epsilon();
+
             for (int j = 0; j < 10; j++)
             {
                 double theta2 = theta*theta, theta4 = theta2*theta2, theta6 = theta4*theta2, theta8 = theta6*theta2;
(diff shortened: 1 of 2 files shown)


asmorkalov:
@creinders Thanks for the patch. Please take a look on CI bot log, it reports test failure:
https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/26307/steps/test_calib3d/logs/stdio

asmorkalov:
Could you also add test for the fixed case?

creinders:
> @creinders Thanks for the patch. Please take a look on CI bot log, it reports test failure:
> https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/26307/steps/test_calib3d/logs/stdio

I will look into it.

> Could you also add test for the fixed case?

Sure. I will get back to you the next days

vpisarev:
@creinders, thank you for the patch, it's definitely an improvement vs the existing implementation!

One thing worries me a bit though. If some points are unstable, you return QNAN as the coordinates in this case. It might be fine, but I'm not sure how much it slows down cv::remap and whether it's convenient to deal with such numbers.

One of the option to solve this problem is to return some fixed large negative coordinates (which can represented bit-exact in FP32, e.g. -1000000). Another option is to add extra `OutputArray mask` parameter that (if it's not noArray()) will contain non-zero U8/S8 values in the case of valid data.

creinders:
@vpisarev I was also thinking about this and didn't know whether it is good practice to return a very large number. One the other side, it could be easier than handling nans.

I like the ideas with an additional output mask. However, does that break all existing implementations especially in Python because two instead of one output is expected?

creinders:
I fixed a small bug and added a test. Additionally -1000000 is returned for invalid points instead of nans as suggested.

vpisarev:
thank you, @creinders, looks good to me! :+1:

alalek:
@creinders Please take a look on build issues.

creinders:
> @creinders Please take a look on build issues.

Ok, let me have a look.

Issue #17453, opened 2020-06-02
countNonZero returns wrong counts for unit8_t arrays with large dimensions (IPP)


D-Alex:
- OpenCV => 4.3
- Operating System / Platform =>Mac OS Catalina + OpenCL + IPP
- Compiler => clang

##### Detailed description

countNonZero is returning wrong counts for large unit8_t arrays. 

The issue seems to be related to ippiCountInRange_8u_C1R which is called behind the scene returning a wrong inverse count.

##### Steps to reproduce

```.cpp
std::cout << cv::countNonZero(cv::Mat::zeros(1,524190-1,CV_8UC1));     // returns 0
std::cout << cv::countNonZero(cv::Mat::zeros(1,524190,CV_8UC1));        // returns 65536
std::cout << cv::countNonZero(cv::Mat::zeros(1,524190+1,CV_8UC1));    // returns 65536
```

mshabunin:
Looks like this problem can be reproduced only on OSX and Win32. Perhaps we can disable IPP path for these platforms temporarily.

diablodale:
Hi all. I can reproduce this exact fail using the current 4.5.x master branch e726ff3296b9e382371a778ad807d37a2e49b979 with VS2019 and a **64-bit** compile target.
I think the #ifdef added in the PR is too narrow to fully resolve this issue.
Test case `Core/CountNonZeroBig./1, where GetParam() = (0, 524190x1)` fails.

#### Cause

https://github.com/opencv/opencv/blob/e726ff3296b9e382371a778ad807d37a2e49b979/modules/core/src/count_non_zero.dispatch.cpp#L65

That code is looking for Apple *or* (Microsoft MSVC compiler with 32-bit compile target). `_M_IX86` is *not* defined for Windows 64-bit compile target. https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160

The original report above is broader. It is all of Windows. Or perhaps all of MSVC compiling for Windows. I don't see anything in the bug report distinguishing that it is only Windows 32-bit compile targets. @mshabunin, you investigated the affected groups. Is this true?

#### Recommendation

[...]

PR #17455, opened 2020-06-02
Added countNonZero test for big arrays

mshabunin:
related #17453

<cut/>

```
buildworker:Custom=linux-3
Xdisable_ipp=ON
```

### DIFF
--- a/modules/core/src/count_non_zero.dispatch.cpp
+++ b/modules/core/src/count_non_zero.dispatch.cpp
@@ -62,6 +62,12 @@ static bool ipp_countNonZero( Mat &src, int &res )
 {
     CV_INSTRUMENT_REGION_IPP();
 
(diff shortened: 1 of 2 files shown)


vpisarev:
@mshabunin, as we adding this test, I would also add:

```
EXPECT_EQ(524190-1, cv::countNonZero(cv::Mat::ones(1, 524190 - 1, CV_8UC1)));
EXPECT_EQ(524190, cv::countNonZero(cv::Mat::ones(1, 524190, CV_8UC1)));
EXPECT_EQ(524190+1, cv::countNonZero(cv::Mat::ones(1, 524190 + 1, CV_8UC1)));
```

which would be another good check for SIMD implementation (including the pure code in OpenCV w/o IPP), since we are also counting intermediate results in 16-bit arithmetics.

PR #17458, opened 2020-06-02
Added a copy constructor and an overloaded assignment operator to FileStorage

### LINKED ISSUE #17412: FileStorage access segfaults after assignment (assignment operator invalidates object)

- OpenCV => 4.3.0 (homebrew)
- Operating System / Platform => macos 10.15
- Compiler => Clang 10.0.1

##### Detailed description

When a `cv::FileStorage` object is assigned, the object cannot be used and causes a segfault with every access.

##### Steps to reproduce

**CMakeLists.txt**
```
find_package(OpenCV 4.0 REQUIRED)
set(CMAKE_CXX_STANDARD 14)
cmake_minimum_required(VERSION 3.15.0)
project(test)
add_executable(test test.cpp)
include_directories(${OpenCV_INCLUDE_DIRS})
target_link_libraries(test ${OpenCV_LIBRARIES})
```

**test.cpp**
```
#include <iostream>
#include <opencv2/core/persistence.hpp>

int main(int argc, char *argv[]) {
  // crash:
  cv::FileStorage fs;
  fs = cv::FileStorage("test.yml",  cv::FileStorage::READ);
  // no crash
  /* cv::FileStorage fs("test.yml", cv::FileStorage::READ); */
  std::string s;
  fs["string"] >> s;
  std::cout << "read value '" << s << "'" << std::endl;
  return 0;
}

```

**test.yml**
```
%YAML:1.0
string: "wat"

```

The issue has been raised [here](https://answers.opencv.org/question/225841/420-filestorage-crash-on-getnodeptr-segmentation-fault/), but so far as I can see, no bug report submitted
 - [x] I updated to latest OpenCV version and the issue is still there. I checked with current master, the problem persists.

EricFlorin:
tl:dr -> Fixes #17412 

Hello everyone,

This is going to be my first pull request, so I'm sorry if I missed something in my submission. I appreciate any feedback as this is my first time contributing to an open source project.

Quick summary of issue #17412: a `cv::FileStorage` object that was assigned by another `cv::FileStorage` object was unusable as it was causing segfaults every time an access occurs.

I look over the issue replication code provided in #17412, and I noticed that `fs` was being assigned to a temporary `cv::FileStorage` object. I went through the execution path of the assignment using my debugger, and noticed that `fs` was not being assigned a deep copy of the temporary `cv::FileStorage` object.

Since `fs` was not given a deep copy of the temporary `cv::FileStorage` object, it loses access to the file reference in the temporary `cv::FileStorage` object after the temporary object is destroyed upon assignment completion.

I fixed the issue by adding a copy constructor and an overloaded assignment operator to `cv::FileStorage` that creates deep copies of `cv::FileStorage` objects. I feel that there is probably a better way of fixing this issue, but this is what I came up with for now.

Thank you for reading!
Eric

### DIFF
--- a/modules/core/include/opencv2/core/persistence.hpp
+++ b/modules/core/include/opencv2/core/persistence.hpp
@@ -340,6 +340,12 @@ public:
      */
     CV_WRAP FileStorage(const String& filename, int flags, const String& encoding=String());
 
+    /** @brief Copy constructor for FileStorage
+
+    This copy constructor creates a deep copy the passed in FileStorage object
+    */
+    CV_WRAP FileStorage(const FileStorage& other);
+
     //! the destructor. calls release()
     virtual ~FileStorage();
 
@@ -398,6 +404,12 @@ public:
      */
     FileNode operator[](const String& nodename) const;
 
+    /** @brief Overloaded assignment operator for FileStorage objects
+    @param other FileStorage object on right hand side of assignment
+    @returns A reference to a FileStorage object
+    */
+    FileStorage& operator=(const FileStorage& other);
+
     /** @overload */
     CV_WRAP_AS(getNode) FileNode operator[](const char* nodename) const;
 
@@ -455,6 +467,8 @@ public:
 
     class Impl;
     Ptr<Impl> p;
+    // Added encoding field to have a complete deep copy
+    std::string encoding;
 };
 
 /** @brief File Storage Node class.
--- a/modules/core/include/opencv2/core/persistence.hpp
+++ b/modules/core/include/opencv2/core/persistence.hpp
@@ -342,7 +342,7 @@ public:
 
     /** @brief Copy constructor for FileStorage
 
-    Copy constructor that creates deep copies of passed in cv::FileStorage objects. Based of the full
+    Copy constructor that creates deep copies of passed in cv::FileStorage objects. Based off the full
     constructor.
     */
     FileStorage(const FileStorage& other);
@@ -465,8 +465,8 @@ public:
 
     int state;
     std::string elname;
-    // Added encoding field to have a complete deep copy
-    std::string encoding;
+    // Added file_encoding field to have a complete deep copy
+    std::string file_encoding;
 
     class Impl;
     Ptr<Impl> p;
--- a/modules/core/include/opencv2/core/persistence.hpp
+++ b/modules/core/include/opencv2/core/persistence.hpp
@@ -342,9 +342,10 @@ public:
 
     /** @brief Copy constructor for FileStorage
 
-    This copy constructor creates a deep copy the passed in FileStorage object
+    Copy constructor that creates deep copies of passed in cv::FileStorage objects. Based of the full
+    constructor.
     */
-    CV_WRAP FileStorage(const FileStorage& other);
+    FileStorage(const FileStorage& other);
 
     //! the destructor. calls release()
     virtual ~FileStorage();
@@ -464,11 +465,11 @@ public:
 
     int state;
     std::string elname;
+    // Added encoding field to have a complete deep copy
+    std::string encoding;
 
     class Impl;
     Ptr<Impl> p;
-    // Added encoding field to have a complete deep copy
-    std::string encoding;
 };
 
 /** @brief File Storage Node class.
--- a/modules/core/src/persistence.cpp
+++ b/modules/core/src/persistence.cpp
@@ -1792,10 +1792,47 @@ FileStorage::FileStorage()
 FileStorage::FileStorage(const String& filename, int flags, const String& encoding)
     : state(0)
 {
+    // Assigns the encoding field with the one passed in
+    this->encoding = encoding;
+
     p = makePtr<FileStorage::Impl>(this);
     bool ok = p->open(filename.c_str(), flags, encoding.c_str());
     if(ok)
         state = FileStorage::NAME_EXPECTED + FileStorage::INSIDE_MAP;
+}
+
+FileStorage::FileStorage(const FileStorage& other) : state(0)
+{
+    // Assign this->encoding to other.encoding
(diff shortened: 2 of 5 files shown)


asmorkalov:
@EricFlorin Thanks for the patch! Please take a look on CI status, buildbot reports compiler warnings:
```
/build/precommit_linux64/opencv/modules/core/src/persistence.cpp:1792:83: warning: declaration of 'encoding' shadows a member of 'cv::FileStorage' [-Wshadow]
/build/precommit_linux64/opencv/modules/core/src/persistence.cpp:1862:81: warning: declaration of 'encoding' shadows a member of 'cv::FileStorage' [-Wshadow]
```

asmorkalov:
Could you also add simple test for the new functions to ensure stability for future.

EricFlorin:
Hey @asmorkalov! I wanted to say thanks for the review!

I also want to ask about what kinds of tests you may be looking for when it comes to the copy and assignment operators for `cv::FileStorage`? I was thinking of using a sample XML/YAML/JSON file and just doing different types of assignment operations with them, but I am not sure if that will be enough.

asmorkalov:
XML/JSON/YAML formats handling is done by different `cv::FileStorage` backends and your fixes work in the same way for all formats. You can test only one of them. The reproducer code in the ticket is enough, just replace final cout by assert.

asmorkalov:
@EricFlorin Please take a look on CI status: https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/26297/steps/compile%20release/logs/warnings%20%282%29. CI bot reports build warnings:
```
/build/precommit_linux64/opencv/modules/core/src/persistence.cpp:1792:83: warning: declaration of 'encoding' shadows a member of 'cv::FileStorage' [-Wshadow]
/build/precommit_linux64/opencv/modules/core/src/persistence.cpp:1862:81: warning: declaration of 'encoding' shadows a member of 'cv::FileStorage' [-Wshadow]
```

EricFlorin:
Hey @asmorkalov,

I got around to fixing the warnings on my side (we will see what the CI bot says). I also modified the reproducer code to act as a test, as well as create my own test script to test different types of cases. The scripts and support files are [in this repository](https://github.com/EricFlorin/OpenCV-PR-17458-Tests) that I created.
- SimpleTest.cpp is the modified reproducer code
- ExtensiveNewFunctionalityTests.cpp is my custom test script

If I am missing a case in my custom test script, please let me know and I will add it!

modules/core/include/opencv2/core/persistence.hpp
```diff
@@ -340,6 +340,13 @@ class CV_EXPORTS_W FileStorage
      */
     CV_WRAP FileStorage(const String& filename, int flags, const String& encoding=String());
 
+    /** @brief Copy constructor for FileStorage
+
+    Copy constructor that creates deep copies of passed in cv::FileStorage objects. Based off the full
+    constructor.
```
alalek, line 346: Almost all non-trivial OpenCV classes are designed as a smart pointers (this behavior is aligned with other language bindings):
- assignment operation means increasing object lifetime (through refcounter)
- optional deep-copy is through `.clone()` operations

Deep copy is over-complicated thing for `FileStorage` (especially in case of write mode), so let avoid that completely.

asmorkalov:
@EricFlorin Do you have chance to implement proposal from @alalek

EricFlorin:
Hey @asmorkalov and @alalek,

Sorry for the long silence this week as I was having final exams.

I did look at the proposal made by @alalek and I did roll back the deep-copy implementation on my end. However, I am stuck on coming up with a solution using smart pointers.

Here's what I found so far using the reproducer code given in #17412 :
1. The temporary `cv::FileStorage` object is created. When the temporary `cv::FileStorage` object is created, `p` (a shared pointer) is initialize a new `cv::FileStorage::Impl` object. The `cv::FileStorage::Impl` object contains an vector of `cv::FileNode` called `roots`. Each `cv::FileNode` in `roots `contains a constant `cv::FileStorage` pointer, whose address is that of the temporary cv::FileStorage object.
2. The `fs` `cv::Filestorage` object is assigned the temporary `cv::FileStorage` object. The assignment causes `fs.p` to be assigned to the temporary `cv::FileStorage` object's `p`.
3. After assignment is complete, the temporary `cv::FileStorage` object is destroyed. The destruction does not affect `fs.p` since `p` is a shared pointer. However, since the `p` in the temporary `cv:FileStorage` object is emptied via the destructor, any `cv::FileNode` with a pointer to the address of the temporary `cv::FileStorage` object will have access to an empty `p`.
4. From there, `fs["string"] >> s` is called. The segfault occurs after `res = p->roots[i][key];` is called in `FileNode FileStorage::operator [](const std::string& key) const`. In the case of the reproducer code, this statement is uses the `cv::FileNode` located at `roots[0]`. The constant `cv::FileStorage` pointer in this `cv::FileNode` points to a `cv::FileStorage` object whose `p` is empty. The "contents" that `p` contained was attempted to be accessed in `const uchar* FileNode::ptr() const` after `IsMap()` was called in `FileNode FileNode::operator[](const std::string& nodename) const`; which failed since `p` is empty.

From there, I am not sure what should I do. I was thinking of having the `const FileStorage*` in `cv::FileStorage` be a `cv::Ptr<FileStorage>`, but I feel that it will not fix the problem as the `p` in the temporary `cv::FileStorage` object will still be deleted.

Any thoughts?

vpisarev:
@EricFlorin, thank you for the detailed research. I fixed the problem in https://github.com/opencv/opencv/pull/17841. The solution was simple – just remove "p.release()" from FileStorage destructor, because otherwise the smart pointer was released twice

EricFlorin:
Thanks for letting me know @vpisarev! I was wondering what the error could be.

PR #17459, opened 2020-06-02
added Apache 2 license and a separate Copyright file for major contributors

vpisarev:
Text of the license is a verbatim copy of http://www.apache.org/licenses/LICENSE-2.0.txt.

Copyrights are extracted from the current BSD license. 3rd-party copyright clause is separated by empty line to leave some space for more major contributors and yet to keep this 3rd-party copyright clause noticeable.

Apache 2 license should be the default license for all new contributions (and eventually for the whole OpenCV).

- [-] There is reference to original bug report and related work
- [-] There is accuracy test, performance test and test data in opencv_extra repository, if applicable
- [-] The feature is well documented and sample code can be built with the project CMake


alalek:
Two concurrent license files in a repository doesn't look as good idea. It is too confusing, because there is no plan to provide dual license.

asmorkalov:
@vpisarev @alalek We decided to add notes to Readme document to make license status and process obvious for users and contributors.

vpisarev:
@alalek, in any case, there will be some transition period, even if it lasts for 1 week. But I expect it to be a bit longer. As @asmorkalov said, at first we need to change the template for OpenCV pull requests, so that all new contributions are supplied under Apache 2 license. Next, we publish at opencv.org announcement that we are migrating to Apache 2 license. The next step will be removal of LICENSE and renaming LICENSE.APACHE2 to LICENSE. Because Apache 2 license does not have convenient place to insert copyrights, and because we will unlikely change the license itself often, I've added a separate copyright file.

vpisarev:
following our discussion, I'll resubmit the PR later

Issue #17470, opened 2020-06-04
G-API streaming test fails, if videoio module built without suitable backend


asmorkalov:
##### System information (version)
- OpenCV => 4.3
- Operating System / Platform => Linux, Android (cross build)

##### Detailed description

Streaming test does not check VideoCapture status and fails with exception "Could't grab the very first frame in function prep; thrown in the test body." The issue is relevant first of all for cross-compilation.

##### Issue submission checklist

 - [ ] I report the issue, it's not a question
   
 - [ ] I checked the problem with documentation, FAQ, open issues,
       answers.opencv.org, Stack Overflow, etc and have not found solution
   
 - [ ] I updated to latest OpenCV version and the issue is still there
   
 - [ ] There is reproducer code and related data files: videos, images, onnx, etc

asmorkalov:
@dmatveev I added check for cases when `cv::VideoCapture` is used directly in https://github.com/opencv/opencv/pull/17471, but it's not possible with G-API `GCaptureSource`. Test report with failed tests is attached.
[gapi_20200603-131810.zip](https://github.com/opencv/opencv/files/4728217/gapi_20200603-131810.zip)

dmatveev:
@asmorkalov would `findDataFile(.., false)` help?

asmorkalov:
The issue is related not to files search, but build configuration. FFmpeg and other video decoding back ends are not available during cross compilation. findDataFile can find file, but VideoCapture raises exception and test fails. I fixed only cases where VideoCapture is created in the test itself, but not hidden inside of G-API module (GCaptureSource).

PR #17471, opened 2020-06-04
Skip some GAPI tests if VideoCapture is not capable to playback video

asmorkalov:
Relates to https://github.com/opencv/opencv/issues/17470


dmatveev:
Thanks! Should it be `findDataFile(..., false)` instead?

I also added an instruction on running tests, can you please check if it works for you? https://github.com/opencv/opencv/wiki/Graph-API

asmorkalov:
@dmatveev thanks for the instruction. The issue is related not to files search, but build configuration. FFmpeg and other video decoding back ends are not available during cross compilation. `findDataFile` can find file, but VideoCapture raises exception and test fails. I fixed only cases where VideoCapture is created in the test itself, but not hidden inside of G-API module.

dmatveev:
@asmorkalov got it now, thanks!

PR #17478, opened 2020-06-05
CUDA: drop support of old CC

### LINKED ISSUE #16649: [Proposal] Remove code for old compute capability
Now, OpenCV has code for old compute capability.

- https://github.com/opencv/opencv/blob/4.2.0/modules/core/include/opencv2/core/cuda/simd_functions.hpp#L95-L103
- https://github.com/opencv/opencv/blob/4.2.0/modules/core/include/opencv2/core/cuda/saturate_cast.hpp#L105-L107
- https://github.com/opencv/opencv/blob/4.2.0/modules/core/include/opencv2/core/cuda/scan.hpp#L196-L207

But, the latest CUDA Toolkit will stop to support old compute capability.  
https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#deprecated-features

> * Support for the following compute capabilities are deprecated in CUDA 10.2. Note that support for these compute capabilities may be removed in a future release of CUDA. 
>   * sm_35 (Kepler)
>   * sm_37 (Kepler)
>   * sm_50 (Maxwell)

So, I think that OpenCV had better to remove code for old compute capability.
It's also effective in terms of maintenance.

tomoaki0705:
resolves #16649
  * update minimum requirement of CUDA to 8.0
  * remove deprecated version checks

Still, this PR is open for discussion.
Supporting different CUDA based on branch is one solution.
Updating CUDA 8.0 is also another point to discuss.
It's still possible to build for old CC, by specifying `CUDA_ARCH_BIN` manually at cmake stage

<cut/>

```
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/cmake/OpenCVDetectCUDA.cmake
+++ b/cmake/OpenCVDetectCUDA.cmake
@@ -78,7 +78,12 @@ if(CUDA_FOUND)
 
   message(STATUS "CUDA detected: " ${CUDA_VERSION})
 
-  set(_generations "Fermi" "Kepler" "Maxwell" "Pascal" "Volta" "Turing" "Ampere")
+  OCV_OPTION(CUDA_ENABLE_DEPRECATED_GENERATION "Enable deprecated generations in the list" OFF)
+  set(_generations "Maxwell" "Pascal" "Volta" "Turing" "Ampere")
+  if(CUDA_ENABLE_DEPRECATED_GENERATION)
+    set(_generations "Fermi" "${_generations}")
+    set(_generations "Kepler" "${_generations}")
+  endif()
   set(_arch_fermi   "2.0")
   set(_arch_kepler  "3.0;3.5;3.7")
   set(_arch_maxwell "5.0;5.2")
@@ -199,10 +204,6 @@ if(CUDA_FOUND)
     endif()
   endmacro()
 
-  macro(ocv_wipeout_deprecated _arch_bin_list)
-    string(REPLACE "2.1" "2.1(2.0)" ${_arch_bin_list} "${${_arch_bin_list}}")
-  endmacro()
-
   set(__cuda_arch_ptx "")
   if(CUDA_GENERATION STREQUAL "Fermi")
     set(__cuda_arch_bin ${_arch_fermi})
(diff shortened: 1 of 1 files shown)


tomoaki0705:
There are many aspects to discuss on this PR.
 - master/3.4 branch
   - May be we can keep the support situation for 3.4, so porting this branch only to master branch makes sense
 - CUDA version
   - dropping support of CC is not equal to minimum CUDA version
   - Choosing 8.0 was a topic not discussed on #16649 
 - Jetson TK1 support
   - If I remember correctly, one of the build machine was based on Jetson TK1.
   - Dropping TK1 might be too early ?
   - If we upgrade to 8.0, user can't enable CUDA feature on TK1, since it only has old CUDA 6.5
 - sm_50
   - Already, there are notice about dropping support sm_50
   - Even we wipe out from the default list, user can specify the CC with `CUDA_ARCH_BIN` option, so there is a way to use it.

asmorkalov:
@tomoaki0705 Thanks for the started discussion you touched the same items that I want to bring. I discussed the proposal with @alalek and @vpisarev  and outcome is the following:

*Key ideas*
- It's not good idea to drop support in minor releases and it's better to do it with major change. We propose to presume support on current level till 5.0.
- OpenCV 3.4 branch is merged to master weekly or be-weekly and we ask people to propose bug fixes to 3.4. Major changes in build infrastructure leads to merge conflicts and it's better to introduce changes to 3.4 first.
- It's reasonable to drop CC 1.x and 2.x, but 3.x is still in use and it'll be great to support Jetson TK1.

*Technical proposal*
- Exclude not CCs from `CUDA_ARCH_BIN` and `CUDA_ARCH_PTX` defaults. and propose only modern arches.
- Introduce extra cmake flag like `ENABLE_DEPRICATED_CUDA_CC" to turn them on or some solution like that to mark old stuff deprecated and enable it with extra options.
- Stay with CUDA 6.5 as baseline, see TK1.
- Enable new functionality for modern CUDA/CC only, if it's needed for DNN.

tomoaki0705:
Sorry for the late reply.
I now understand the key idea of what you meant.
Let me update the PR

tomoaki0705:
@asmorkalov , I've updated the PR.
Please let me know if this PR is aligned with your thought.

PR #17492, opened 2020-06-06
add Memory barriers

cyyever:
<cut/>

```
force_builders=Custom
build_image:Custom=centos:7
buildworker:Custom=linux-1
```

### DIFF
--- a/modules/core/src/precomp.hpp
+++ b/modules/core/src/precomp.hpp
@@ -364,15 +364,8 @@ bool __termination;  // skip some cleanups, because process is terminating
 
 cv::Mutex& getInitializationMutex();
 
-// TODO Memory barriers?
 #define CV_SINGLETON_LAZY_INIT_(TYPE, INITIALIZER, RET_VALUE) \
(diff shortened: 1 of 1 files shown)


modules/core/src/precomp.hpp
```diff
@@ -364,13 +365,12 @@ bool __termination;  // skip some cleanups, because process is terminating
...
-    if (instance == NULL) \
+    static std::atomic<TYPE*> instance; \
+    if (instance == nullptr) \
     { \
         cv::AutoLock lock(cv::getInitializationMutex()); \
-        if (instance == NULL) \
+        if (instance == nullptr) \
             instance = INITIALIZER; \
     } \
     return RET_VALUE;
```
alalek, line 376: Perhaps, in C++11 mode this can be simplified (without thread-safety tricks):
```
static TYPE* const instance = INITIALIZER;
return RET_VALUE;
```

(static global initialization is guarded by C++11 compiler/runtime)

See also https://stackoverflow.com/questions/8102125/is-local-static-variable-initialization-thread-safe-in-c11

Issue #17464, opened 2020-06-03
Python bindings generator fails for argument with name "args"


dkurt:
Because of `PyObject* args`

/cc @TolyaTalamanov

PR #17493, opened 2020-06-06
G-API: Python bindings

TolyaTalamanov:
## Overview
This patch is a first part of work to make it possible to use G-API from python
Going to split this feature on separate PR's to minimize diff
Rough plan might look like this:
- [x]  **Basic functionality**
- [ ] Kernels (core/imgproc)
- [ ] Streaming API
- [ ] Serialization
- [ ] Infer API
- [ ] Python API to implement custom operations

<cut/>

### OpenCV wrapper changes
OpenCV has already had mechanism to wrap code into python
It consist of two parts:
* Header parser (`modules/python/src2/hdr_parser.py`)
* Code generator (`modules/python/src2/gen2.py`)

#### Problems
While using this mechanism, the following problems were found:
1. Impossible to wrap function if one of arguments has name `args`, because this name used in generated code    
**Solution**: Rename `args` to `py_args`, I think it makes sense to add assert that arguments names don't match with internals names used by code generator, but I didn't add in this PR. Let's to consider this separately.

2. Impossible to wrap function if one of arguments is `rvalue`, because parser doesn't recognize this pattern: `foo(Obj&& obj)`
**Solution**: Extend parser

3. Impossible to wrap function if there is a such default initialization: `foo(Obj&& obj = { })`
**Solution**: Extend parser

4. Impossible to wrap `enum class Enum : int`, because parser doesn't recognize inheritance syntax for `enum`
BTW it also doesn't recognize `enum class` but since to `class` will be parsed after `enum` it works fine, make sense to handle `enum class` separately  
**Solution**: Extend parser

5. Currently, the code gen only adds `namespace` where a function resides, but this means that the inputs/outputs and function must be in the same `namespace`

Example:
https://github.com/opencv/opencv/blob/3e32247deacccb4eec064f263577c7b4833eb4ea/modules/gapi/include/opencv2/gapi/ocl/core.hpp#L14-L20

It would be error if it didn't contains `cv::gapi::` before `GKernelPackage`, because code gen would add only `cv::gapi::core::ocl` namespace and `GKernelPackage` wouldn't be found

**Solution** Just explicitly add `cv::gapi::` for such cases.

#### Needed changes
1. To communicate with python wrapper there is a special API expressed by a list of macros, but since G-API can work in a standalone mode it should contains an own macros, so parser should know about these macros as well. 
Were added following macros:
* `GAPI_EXPORT_W`
* `GAPI_EXPORT_W_SIMPLE`
* `GAPI_WRAP`

2. Parser shouldn't process headers which needed only for standalone mode, so just skip header if it contains `GAPI_STANDALONE` macro

### G-API changes
1. Were added macros to communicate with python wrapper
2. Wrapped several entities to validate basic functionality:
3. Added default `ctor` for `GCompileArg` & overload for `compile_args(GKernelPackage pkg)`
4. Added special folder `gapi/misc/python` which should contain python specific things for G-API

### Testing
I think it makes sense to test every operation
I hope I can find a good testing strategy in futher PR's

### Buildbot configuration

```

force_builders=ARMv7,Custom
build_image:Custom=centos:7
buildworker:Custom=linux-1
```


modules/python/src2/hdr_parser.py
```diff
@@ -699,7 +707,7 @@ def parse_stmt(self, stmt, end_token, mat="Mat", docstring=""):
                     decl[1] = ": " + ", ".join([self.get_dotted_name(b).replace(".","::") for b in bases])
                 return stmt_type, classname, True, decl
 
-            if stmt.startswith("class") or stmt.startswith("struct"):
+            if (stmt.startswith("class") or stmt.startswith("struct")) and not stmt.startswith('enum'):
```
TolyaTalamanov, line 710: It's necessary to handle `enum class` there

TolyaTalamanov:
@alalek Could you take a look on changes for python wrapper ?

modules/python/src2/hdr_parser.py
dkurt: Maybe it may help: https://github.com/opencv/opencv/pull/13471

modules/gapi/CMakeLists.txt
```diff
@@ -24,6 +24,8 @@ ocv_add_module(gapi
       opencv_imgproc
     OPTIONAL
       opencv_video
+      WRAP
```
dkurt, line 27: Please align WRAP with OPTIONAL

modules/gapi/include/opencv2/gapi/own/exports.hpp
```diff
@@ -8,11 +8,18 @@
...
 #define OPENCV_GAPI_OWN_TYPES_HPP
 
+#   define GAPI_WRAP
+
 #   if defined(__OPENCV_BUILD)
 #       include <opencv2/core/base.hpp>
 #       define GAPI_EXPORTS CV_EXPORTS
+        /* special informative macros for wrapper generators */
+#       define GAPI_EXPORTS_W_SIMPLE CV_EXPORTS
+#       define GAPI_EXPORTS_W CV_EXPORTS
```
dkurt, line 18: Maybe it's more safe to define in this way?
```
#       define GAPI_EXPORTS_W_SIMPLE CV_EXPORTS_W_SIMPLE
#       define GAPI_EXPORTS_W CV_EXPORTS_W
```

modules/python/src2/hdr_parser.py
TolyaTalamanov: Great, I'll try it

modules/gapi/include/opencv2/gapi/own/exports.hpp
TolyaTalamanov: I thinks so, thanks

modules/gapi/misc/python/test/test_gapi_core.py
```diff
@@ -0,0 +1,42 @@
...
+        in1 = np.random.randint(0, 100, sz).astype(np.uint8)
+        in2 = np.random.randint(0, 100, sz).astype(np.uint8)
+
+        # OpenCV
+        expected = in1 + in2
+
+        # G-API
+        g_in1 = cv.GMat()
+        g_in2 = cv.GMat()
+        g_out = cv.gapi.add(g_in1, g_in2)
```
dkurt, line 32: I think that it's better to show in tests the best codestyle to work with G-API in Python. So maybe we need `from cv2 import gapi` and use as
```python
g_out = gapi.add(g_in1, g_in2)
```

modules/gapi/misc/python/test/test_gapi_core.py
```diff
@@ -0,0 +1,42 @@
...
+        comp = cv.GComputation(g_in1, g_in2, g_out)
+
+        for pkg in pkgs:
+            actual = comp.apply(in1, in2, args=cv.compile_args(pkg))
+            # Comparison
+            np.testing.assert_array_equal(expected, actual)
```
dkurt, line 38: Please use asserts from base class

TolyaTalamanov: Yea, but is there any way to compare two `numpy` array ?

modules/gapi/src/api/gkernel.cpp
```diff
@@ -98,3 +98,8 @@ std::vector<cv::gapi::GBackend> cv::gapi::GKernelPackage::backends() const
 
     return std::vector<cv::gapi::GBackend>(unique_set.begin(), unique_set.end());
 }
+
+cv::GCompileArgs cv::compile_args(cv::gapi::GKernelPackage pkg)
+{
+    return cv::GCompileArgs{ cv::GCompileArg(std::move(pkg)) };
+}
```
dkurt, line 105: Just for note - there is a way not to add it to main library - you just can mark in header as `#if 0` and move implementation to `pyopencv_gapi.hpp`

modules/python/src2/gen2.py
```diff
@@ -27,14 +27,14 @@
 gen_template_call_constructor_prelude = Template("""new (&(self->v)) Ptr<$cname>(); // init Ptr with placement new
         if(self) """)
 
-gen_template_call_constructor = Template("""self->v.reset(new ${cname}${args})""")
+gen_template_call_constructor = Template("""self->v.reset(new ${cname}${py_args})""")
```
dkurt, line 30: Am I right that this PR will solve https://github.com/opencv/opencv/issues/17464?

modules/gapi/misc/python/test/test_gapi_core.py
dkurt: `cv.norm(a, b, cv.NORM_INF)` in example

modules/python/src2/gen2.py
TolyaTalamanov: Exactly, but it can be improved by adding check that internal names doesn't match with names from user

modules/python/src2/hdr_parser.py
TolyaTalamanov: Fixed

modules/gapi/CMakeLists.txt
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/own/exports.hpp
TolyaTalamanov: Fixed

modules/gapi/misc/python/test/test_gapi_core.py
TolyaTalamanov: It's better to leave `cv.gapi.add` because it's a little bit strange when it use `cv.GMat` and `gapi.add`

modules/gapi/src/api/gkernel.cpp
TolyaTalamanov: Cool! Fixed

modules/gapi/misc/python/test/test_gapi_core.py
TolyaTalamanov: Fixed

TolyaTalamanov:
@dkurt Could you check it, please ?

modules/gapi/include/opencv2/gapi/fluid/gfluidbuffer.hpp
```diff
@@ -80,6 +78,7 @@ class GAPI_EXPORTS View
     Priv& priv();               // internal use only
     const Priv& priv() const;   // internal use only
 
+    View();
```
dmatveev, line 81: How did this affect the python binding? I thought we don't wrap it now, right?

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -698,6 +698,11 @@ namespace detail
         static const char* tag() { return "gapi.use_only"; }
     };
 } // namespace detail
+
+#ifdef OPENCV_BINDINGS_PARSER
+GAPI_EXPORTS_W GCompileArgs compile_args(cv::gapi::GKernelPackage pkg);
+#endif // OPENCV_BINDINGS_PARSER
```
dmatveev, line 704: Is it ok to have such conditionals in the public API headers? cc: @alalek

modules/gapi/misc/python/pyopencv_gapi.hpp
```diff
@@ -0,0 +1,21 @@
+using gapi_GKernelPackage = cv::gapi::GKernelPackage;
```
dmatveev, line 1: If this file can have comments, having a comments here would help

dmatveev: also, does it need the include guards, license header, or something?

modules/gapi/include/opencv2/gapi/gkernel.hpp
```diff
@@ -698,6 +698,11 @@ namespace detail
         static const char* tag() { return "gapi.use_only"; }
     };
 } // namespace detail
+
+#ifdef OPENCV_BINDINGS_PARSER
```
dkurt, line 702: Should be `#if OPENCV_BINDINGS_PARSER`

modules/gapi/include/opencv2/gapi/fluid/gfluidbuffer.hpp
TolyaTalamanov: These changes should fix compilation errors on window we faced when compiled g-api with python module

modules/gapi/misc/python/pyopencv_gapi.hpp
TolyaTalamanov: It's a good question     
I suppose this file used only by python wrapper and it isn't necessary to have include guards    
Check this: https://github.com/opencv/opencv/blob/master/modules/dnn/misc/python/pyopencv_dnn.hpp

@dkurt Could you comment it ?

modules/gapi/include/opencv2/gapi/gkernel.hpp
TolyaTalamanov: Are u sure ?     
`/build/precommit_linux64/opencv/modules/gapi/include/opencv2/gapi/gkernel.hpp:702:5: warning: "OPENCV_BINDINGS_PARSER" is not defined [-Wundef]`

TolyaTalamanov: Reverted

modules/gapi/misc/python/pyopencv_gapi.hpp
TolyaTalamanov: Added comment

modules/gapi/misc/python/pyopencv_gapi.hpp
```diff
@@ -0,0 +1,30 @@
+// This file is supposed to use for things necessary for supporting the G-API python API
```
TolyaTalamanov, line 1: To support

modules/gapi/misc/python/pyopencv_gapi.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+template<>
+PyObject* pyopencv_from(const std::vector<GCompileArg>& value)
+{
+    return pyopencv_from_generic_vec(value);
+}
+
+// NB: Used only by python wrapper
+// G-API have template implementation for compile_args which can't be exported into python, due python doesn't work with templates
+// so it's a implementation for specific type
```
TolyaTalamanov, line 23: an

TolyaTalamanov:
@alalek Could you check, please ?

modules/python/src2/hdr_parser.py
```diff
@@ -260,7 +268,7 @@ def parse_class_decl(self, decl_str):
             modlist.append("=" + macro_arg)
             l = l[:npos] + l[npos3+1:]
 
-        l = self.batch_replace(l, [("CV_EXPORTS_W", ""), ("CV_EXPORTS", ""), ("public virtual ", " "), ("public ", " "), ("::", ".")]).strip()
+        l = self.batch_replace(l, [("CV_EXPORTS_W", ""), ("CV_EXPORTS", ""), ("GAPI_EXPORTS_W", ""), ("GAPI_EXPORTS", ""), ("public virtual ", " "), ("public ", " "), ("::", ".")]).strip()
```
alalek, line 271: To keep code maintainable/mergable it is better to put G-API specifics on a separate single line (because they are just aliases):
```
# G-API specific aliases
l = self.batch_replace(l, [
        ("GAPI_EXPORTS", "CV_EXPORTS"),
        ("GAPI_EXPORTS_W", "CV_EXPORTS_W"),
        ...
        ('defined(GAPI_STANDALONE)', '0'),
    ])

... existed unchanged OpenCV code ...
```

dmatveev: > args should be compile_args in .apply() methods.

Yep but we have a function with the same name. :)

modules/gapi/include/opencv2/gapi/gkernel.hpp
dkurt: @TolyaTalamanov , please move to something like `misc/python/shadow_gapi.hpp` (see https://github.com/opencv/opencv/blob/master/modules/python/bindings/CMakeLists.txt and https://github.com/opencv/opencv/blob/master/modules/core/misc/python/shadow_umat.hpp)

modules/gapi/include/opencv2/gapi/own/exports.hpp
```diff
@@ -8,11 +8,18 @@
 #ifndef OPENCV_GAPI_OWN_TYPES_HPP
 #define OPENCV_GAPI_OWN_TYPES_HPP
 
+#   define GAPI_WRAP
```
dkurt, line 11: Shall it be mapped to `CV_WRAP` in case of `#   if defined(__OPENCV_BUILD)` ? Similar to `GAPI_EXPORTS_*`

modules/python/src2/hdr_parser.py
dkurt: @TolyaTalamanov , please consider https://github.com/opencv/opencv/pull/17493/files#r447131187 (it will make it easier to backport changes between master and 3.4.

modules/python/src2/hdr_parser.py
```diff
@@ -821,7 +831,7 @@ def parse(self, hname, wmode=True):
                     continue
                 state = SCAN
                 l = re.sub(r'//(.+)?', '', l).strip()  # drop // comment
-                if l == '#if 0' or l == '#if defined(__OPENCV_BUILD)' or l == '#ifdef __OPENCV_BUILD':
+                if l == '#if 0' or l == '#if defined(__OPENCV_BUILD)' or l == '#ifdef __OPENCV_BUILD' or l == '#if defined(GAPI_STANDALONE)':
```
dkurt, line 834: For maintenance purpose:
```python
if l == '#if 0' or \
   l == '#if defined(GAPI_STANDALONE)' or \
   l == '#if defined(__OPENCV_BUILD)' or \
   l == '#ifdef __OPENCV_BUILD':
```

modules/python/src2/hdr_parser.py
```diff
@@ -708,13 +716,15 @@ def parse_stmt(self, stmt, end_token, mat="Mat", docstring=""):
                         print("Error at %s:%d" % (self.hname, self.lineno))
                         exit(1)
                     decl = []
-                    if ("CV_EXPORTS_W" in stmt) or ("CV_EXPORTS_AS" in stmt) or (not self.wrap_mode):# and ("CV_EXPORTS" in stmt)):
+                    if ("CV_EXPORTS_W" in stmt) or ("GAPI_EXPORTS_W" in stmt) or ("CV_EXPORTS_AS" in stmt) or (not self.wrap_mode):# and ("CV_EXPORTS" in stmt)):
```
dkurt, line 719:
```python
if ("CV_EXPORTS_W" in stmt) or \
   ("GAPI_EXPORTS_W" in stmt) or \
   ("CV_EXPORTS_AS" in stmt) or \
   (not self.wrap_mode):# and ("CV_EXPORTS" in stmt)):
```

modules/gapi/include/opencv2/gapi/gcomputation.hpp
```diff
@@ -288,7 +288,7 @@ class GAPI_EXPORTS GComputation
      * @param args compilation arguments for underlying compilation
      * process.
      */
-    void apply(cv::Mat in1, cv::Mat in2, cv::Mat &out, GCompileArgs &&args = {}); // Binary overload
+    GAPI_WRAP void apply(cv::Mat in1, cv::Mat in2, CV_OUT cv::Mat &out, GCompileArgs &&args = {}); // Binary overload
```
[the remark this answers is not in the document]
dkurt: @TolyaTalamanov , can you please comment on this?

modules/gapi/include/opencv2/gapi/gkernel.hpp
TolyaTalamanov: Done

modules/python/src2/hdr_parser.py
TolyaTalamanov: Done

modules/gapi/include/opencv2/gapi/own/exports.hpp
TolyaTalamanov: Yes, fixed

modules/python/src2/hdr_parser.py
TolyaTalamanov: Fixed

modules/python/src2/hdr_parser.py
TolyaTalamanov: Fixed

modules/gapi/include/opencv2/gapi/gcomputation.hpp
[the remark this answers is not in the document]
TolyaTalamanov: My suggestion is to do this in the next PR to simplify diff

TolyaTalamanov:
@alalek Can it be merged ?

TolyaTalamanov:
@mshabunin

dmatveev:
@TolyaTalamanov is back! 🥳

TolyaTalamanov:
@mshabunin Errors don't seem to be related to my changes, is it possible to rebuild CI ?

TolyaTalamanov:
> @mshabunin Errors don't seem to be related to my changes, is it possible to rebuild CI ?

Thanks, CI passed

dmatveev:
Fantastic!

PR #17497, opened 2020-06-08
videoio: fix plugins build

alalek:
Fix embedding of version information through .rc [extra file](https://github.com/opencv/opencv_3rdparty/blame/ffmpeg/master/ffmpeg/build_videoio_plugin.sh#L35).


alalek:
:+1:

Issue #17329, opened 2020-05-20
How can I find the essential matrix between two different cameras?


themightyoarfish:
This is a question about the API, not really a bug report.

The documentation on [`findEssentialMat()`](https://docs.opencv.org/3.4/d9/d0c/group__calib3d.html#ga13f7e34de8fa516a686a56af1196247f) states
> Note that this function assumes that points1 and points2 are feature points from cameras with the same camera matrix. 

But this is a very strong restriction. Is there a way to solve this problem between two distinct cameras?
Or would it be possible to `undistortPoints()` the points to normalized image coordinates with identity camera matrices for both cameras?

tompollok:
I think you could extract it estimating the Fundamental Matrix and then using the formula to extract the essential matrix.

E = (K')^t F K

Other option could be to undistort the points with 

cv::undistortPoints(inputDistortedPoints, outputUndistortedPoints, cameraMatrix, distCoeffs, R=cv::noArray(), P=cv::noArray());

P – New camera matrix (3x3) or new projection matrix (3x4). P1 or P2 computed by stereoRectify() can be passed here. If the matrix is empty, the identity new camera matrix is used.

So i guess you could just put the identity camera matrix into findEssentialMat().

themightyoarfish:
I think I'd like to submit a PR to the documentation that makes it clear how it can be done, so I'll keep this open for now.

chadrs2:
Use the stereoCalibrate() function

PR #17551, opened 2020-06-14
[WIP] add findEssentialMat for two different cameras

tompollok:
Adds findEssentialMatrix for two cameras with different matrices. Uses the undistortion workaround discussed in  #17329

related to #17329

```
force_builders=linux,docs
```

### DIFF
--- a/modules/calib3d/include/opencv2/calib3d.hpp
+++ b/modules/calib3d/include/opencv2/calib3d.hpp
@@ -2521,6 +2521,57 @@ CV_EXPORTS_W Mat findEssentialMat( InputArray points1, InputArray points2,
                                  double focal = 1.0, Point2d pp = Point2d(0, 0),
                                  int method = RANSAC, double prob = 0.999,
                                  double threshold = 1.0, OutputArray mask = noArray() );
+    
+/** @brief Calculates an essential matrix from the corresponding points in two images from potentially two different cameras.
+
+@param points1 Array of N (N \>= 5) 2D points from the first image. The point coordinates should
+be floating-point (single or double precision).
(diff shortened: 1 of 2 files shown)


tompollok:
Should i rebase to 3.4?

Can i also add an recoverPose overload that only takes the points and matrices as input and returns R and t?

alalek:
Thank you for contribution!

3.4 is not necessary here.

Please add tests for new functionality.

tompollok:
There is an issue with the threshold parameter in this implementation. Due to the fact, that i undistort the points, they are in normalized coordinates. When passing them to the findEssentialMatrix function i have to scale the threshold parameter. But given the fact that it can be two different cameras with different intrinsics, i dont know whats a good way to scale it.

So what could be potentially done is to scale thresh by dividing by fx of one of the cameras? Doesnt sound perfect to me. Any other ideas?

tompollok:
If using magsac instead of opencv RANSAC, then the thresh could be ommited completely if i understand correctly.
https://github.com/danini/magsac

Is there any information which RANSAC will be implemented by the GSOC? It seems like opencv RANSAC is really needing some work towards state of the art.

https://local-features-tutorial.github.io/pdfs/Local_features_from_paper_to_practice.pdf
p14
grafik

alalek:
@tompollok BTW, There is RANSAC sub-topic in OE-20 (#11015) - good place for your findings.

vpisarev:
@tompollok, thank you for the contribution. I could not figure out how to edit your PR in-place, because adding a test required changing another file, not from the list. So I created new PR with your commits preserved: https://github.com/opencv/opencv/pull/17816

PR #17556, opened 2020-06-15
cuda optflow TVL1 : run safely in async mode

### LINKED ISSUE #16013: Corrupted optical flow using cuda::DenseOpticalFlow asynchronously in multithreaded environment

- OpenCV => 4.0
- Cuda => 10.1

##### Detailed description

I will duplicate the problem described here: 
https://stackoverflow.com/questions/59075944/why-do-i-get-corrupted-results-if-i-run-gpu-dense-optical-flow-algorithm-concurr

> While using OpenCV CUDA dense optical flow in parallel I noticed that sometimes I get corrupted optical flow results, though I run it on different `cuda::GpuMat`s and in separate `cuda::Stream`s with separate Algorithm instances.
> 
> After some experiments with the code I found out that if I protect `DenseOpticalFlow::calc()` call with `mutex` or run only one thread, I always get correct (not corrupted) results.
> 
> I reduced my code to minimum reproducible example where I run optical flow algorithm on same input images multiple times in multiple threads

Furthermore, I checked that calling `cuda::Stream::waitForCompletion()` before calling `DenseOpticalFlow::calc()` asynchronously still causes the corruption.

However, if i use default stream `cuda::Stream::Null()` or use only `DenseOpticalFlow::calc()` synchronously, no corruption occurs.

Seems like using `DenseOpticalFlow::calc()` asynchronously causes data races.

##### Steps to reproduce

To reproduce the problem, you can use the example provided by the link above.

nglee:
Fixes #16013

We are going to use texture object, rather than texture reference.
It allows us to run the algorithm asynchronously.
This is similar to what's been done in #11483

<cut/>

```
allow_multiple_commits=1
force_builders=Custom
buildworker:Custom=linux-4
```

### DIFF
--- a/modules/core/include/opencv2/core/cuda/common.hpp
+++ b/modules/core/include/opencv2/core/cuda/common.hpp
@@ -101,6 +101,20 @@ namespace cv { namespace cuda
             cudaChannelFormatDesc desc = cudaCreateChannelDesc<T>();
             cudaSafeCall( cudaBindTexture2D(0, tex, img.ptr(), &desc, img.cols, img.rows, img.step) );
         }
+
+        template<class T> inline void createTextureObjectPitch2D(cudaTextureObject_t* tex, PtrStepSz<T>& img, const cudaTextureDesc& texDesc)
+        {
+            cudaResourceDesc resDesc;
+            memset(&resDesc, 0, sizeof(resDesc));
+            resDesc.resType = cudaResourceTypePitch2D;
+            resDesc.res.pitch2D.devPtr = static_cast<void*>(img.ptr());
+            resDesc.res.pitch2D.height = img.rows;
+            resDesc.res.pitch2D.width = img.cols;
+            resDesc.res.pitch2D.pitchInBytes = img.step;
+            resDesc.res.pitch2D.desc = cudaCreateChannelDesc<T>();
+
+            cudaSafeCall( cudaCreateTextureObject(tex, &resDesc, &texDesc, NULL) );
+        }
     }
 }}
 
--- a/modules/cudaimgproc/src/cuda/canny.cu
+++ b/modules/cudaimgproc/src/cuda/canny.cu
@@ -90,53 +90,47 @@ namespace cv { namespace cuda { namespace device
 
 namespace canny
 {
-    texture<uchar, cudaTextureType2D, cudaReadModeElementType> tex_src(false, cudaFilterModePoint, cudaAddressModeClamp);
     struct SrcTex
     {
-        int xoff;
-        int yoff;
+        virtual ~SrcTex() {}
+
         __host__ SrcTex(int _xoff, int _yoff) : xoff(_xoff), yoff(_yoff) {}
 
-        __device__ __forceinline__ int operator ()(int y, int x) const
+        __device__ __forceinline__ virtual int operator ()(int y, int x) const = 0;
+
+        int xoff;
+        int yoff;
+    };
+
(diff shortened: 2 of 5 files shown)


alalek:
@nglee Thank you for the contribution!

Could you please take a look on this error message (from Custom builder)?
```
/build/precommit_custom_linux/3.4/opencv/modules/cudaimgproc/src/cuda/canny.cu:93:8: error: 'struct canny::SrcTex' has virtual functions and accessible non-virtual destructor [-Werror=non-virtual-dtor]
     struct SrcTex
        ^~~~~~
/build/precommit_custom_linux/3.4/opencv/modules/cudaimgproc/src/cuda/canny.cu:104:8: error: base class 'struct canny::SrcTex' has accessible non-virtual destructor [-Werror=non-virtual-dtor]
     struct SrcTexRef : SrcTex
        ^~~~~~~~~
/build/precommit_custom_linux/3.4/opencv/modules/cudaimgproc/src/cuda/canny.cu:104:8: error: 'struct canny::SrcTexRef' has virtual functions and accessible non-virtual destructor [-Werror=non-virtual-dtor]
/build/precommit_custom_linux/3.4/opencv/modules/cudaimgproc/src/cuda/canny.cu:114:8: error: base class 'struct canny::SrcTex' has accessible non-virtual destructor [-Werror=non-virtual-dtor]
     struct SrcTexObj : SrcTex
        ^~~~~~~~~
/build/precommit_custom_linux/3.4/opencv/modules/cudaimgproc/src/cuda/canny.cu:114:8: error: 'struct canny::SrcTexObj' has virtual functions and accessible non-virtual destructor [-Werror=non-virtual-dtor]
cc1plus: some warnings being treated as errors
```

(perhaps it is enough to define destructor of the base class as `virtual`)

nglee:
@alalek Build error fixed!

nglee:
It is ready to be merged. I've also tested with OP's test code in #16013.

tomoaki0705:
Hi @nglee , I see build error on Jetson TX1 and TX2.

```
/opencv-fork/modules/cudaimgproc/src/cuda/canny.cu(131): error: namespace "std" has no member "enable_if_t"

/opencv-fork/modules/cudaimgproc/src/cuda/canny.cu(131): error: expected a "," or ">"
```

The direct cause come from [this line](https://github.com/opencv/opencv/blame/9755ab160d94f4ad2a672653cf657ad421d46837/modules/cudaimgproc/src/cuda/canny.cu#L131)

https://github.com/opencv/opencv/blame/9755ab160d94f4ad2a672653cf657ad421d46837/modules/cudaimgproc/src/cuda/canny.cu#L131

We only enable C++11 feature on master, not on 3.4, but it seems `enable_if_t` is a [c++14 feature](https://en.cppreference.com/w/cpp/types/enable_if)
> template< bool B, class T = void > using enable_if_t = typename enable_if<B,T>::type; |   | (since C++14)

I appreciate if you could have a look on this.
Still it's unknown why I didn't face this on Jetson Xavier and Jetson Nano.

nglee:
@tomoaki0705 

Thank you for investigation. Would you please check your g++ version?
> The default mode is C++98 for GCC versions prior to 6.1, and C++14 for GCC 6.1 and above. ([Link](https://www3.ntu.edu.sg/home/ehchua/programming/cpp/gcc_make.html))

To enable c++14 via cmake, maybe we can try this flag: `-DCMAKE_CXX_STANDARD=14`.

Just removing the lines with `enable_if_t` would also work.

tomoaki0705:
Jetson TX1 GCC 5.4.0
Jetson TX2 GCC 5.4.0
Jetson Nano GCC 7.5.0
Jetson Xavier GCC 7.5.0
That makes sense.

PR #17570, opened 2020-06-17
[GSoC] High Level API and Samples for Scene Text Detection and Recognition

HannibalAPE:
**Merge with extra**: opencv/opencv_extra#773
High-Level API and Samples for Scene Text Detection and Recognition
This is my project in GSoC 2020: OpenCV Text/Digit detection & recognition projects.

Short Video: https://drive.google.com/file/d/1IlGpRRhPCifC9TRzuhq0_G1P6MkP33BJ/view?usp=sharing

For more information:
https://github.com/HannibalAPE/opencv/blob/text_det_recog_demo/doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown

<cut/>

TODO LIST:

- Scene Text Recognition:
- [x] Samples
- [x] High-Level API

- Scene Text Detection:
- [x] Samples
- [x] High-Level API

- Scene Text Spotting:
- [x] Samples

- Document:
- [x] Tutorials

- [x] I agree to contribute to the project under OpenCV (Apache 2) License.
      But the patch to opencv_extra does not have the same branch name. see as follows.

```
opencv_extra=text_det_rec

build_image:Custom=centos:7
buildworker:Custom=linux-1
```


samples/dnn/scene_text_recognition.cpp
```diff
@@ -0,0 +1,114 @@
...
+        for (uint j = 0; j < vocabulary.length() + 1; j++) {
+            float score = prediction.at<float>(i, j);
+            if (maxScore < score) {
+                maxScore = score;
+                maxLoc = j;
+            }
+        }
+        if (maxLoc > 0) {
+            char currentChar = vocabulary[maxLoc - 1];
+            if (currentChar != decodeSeq[-1])
```
berak, line 108: `decodeSeq[-1]` 

`-1` is illegal, maybe use `decodeSeq.back()` instead ?

samples/dnn/scene_text_detection.cpp
```diff
@@ -0,0 +1,242 @@
...
+#include <opencv2/dnn/dnn.hpp>
+
+using namespace cv;
+using namespace cv::dnn;
+
+std::string keys =
+        "{ help  h                          | | Print help message. }"
+        "{ inputImage i                     | | Path to an input image. Skip this argument to capture frames from a camera. }"
+        "{ device d                         | 0 | camera device number. }"
+        "{ modelPath mp                     | | Path to a binary .onnx file contains trained DB detector model.}"
```
berak, line 30: just curious -- where do we find the model (link) ?

samples/dnn/scene_text_recognition.cpp
HannibalAPE: Fix it.
I got the right result with index -1, so I thought a negative index may be supported now. :(
Thank you.

samples/dnn/scene_text_detection.cpp
HannibalAPE: Maybe I can upload models to Google Drive.
I am not sure about it, and I will ask my mentor.

modules/dnn/test/test_model.cpp
```diff
@@ -370,6 +389,21 @@ TEST_P(Test_Model, Segmentation)
     testSegmentationModel(weights_file, config_file, inp, exp, norm, size, mean, scale, swapRB);
 }
 
+TEST_P(Test_Model, SceneTextRec)
+{
+    std::string imgPath = _tf("welcome.png");
```
bhack, line 394: Where is this file?

HannibalAPE: I have already put the data into opencv/opencv_extra, but it has not been merged yet.
https://github.com/opencv/opencv_extra/pull/773
The name of image has been changed into "text_rec_test.png", and I will push again when the data is ready.
Thanks for your review.

bhack: Please push the new name. You don't need to wait the merge. See [this step](https://pullrequest.opencv.org/buildbot/builders/precommit_linux64/builds/26892/steps/Merge%20opencv_extra%20with%20test%20branch/logs/stdio) in the build pipeline.

bhack:
@vpisarev I suppose that we could not have a single PR looking at the TODO list in the description.

samples/dnn/scene_text_detection.cpp
```diff
@@ -0,0 +1,380 @@
...
+        static const std::string winNameGT = "GT";
+        namedWindow(winNameGT, WINDOW_NORMAL);
+
+        String testImgPath;
+        while (std::getline(testList, testImgPath)) {
+            String imgPath = evalDataPath + "/test_images/" + testImgPath;
+            // std::cout << "Image Path: " << imgPath << std::endl;
+
+            Mat frame = imread(imgPath, 1);
+            CV_Assert(!frame.empty());
```
bhack, line 111: You could use something like https://github.com/opencv/opencv/blob/73959fed4581556d8e6b770987c819adfacceb87/modules/video/perf/perf_bgfg_mog2.cpp#L30

samples/dnn/scene_text_detection.cpp
```diff
@@ -0,0 +1,380 @@
...
+            polylines(frame, detectionResults, true, Scalar(0, 255, 0), 2);
+            imshow(winName, frame);
+
+            // load groundtruth
+            String imgName = testImgPath.substr(0, testImgPath.length() - 4);
+            String gtPath = evalDataPath + "/test_gts/" + imgName + ".txt";
+            std::cout << gtPath << std::endl;
+            std::ifstream gtFile;
+            gtFile.open(gtPath);
+            CV_Assert(gtFile.is_open());
```
bhack, line 155: Same here

samples/dnn/scene_text_detection.cpp
HannibalAPE: I feel confused that this file is only a sample for scene text detection, not an API. Is there anything related to the test?
Why should I change CV_Assert into SkipTestException?
Thank you.

modules/dnn/test/test_model.cpp
HannibalAPE: Thank you.
Sadly, I still get an error in "Linux x64 Debug", but I find that the test_dnn in "Linux x64" passed. I am not sure about the difference between these two tests.

Is there any detailed log information containing which line throws out the error.
I can only get "error: (-215: Assertion failed) dims <= 2 in function 'at' thrown in the test body." in https://pullrequest.opencv.org/buildbot/builders/precommit_linux64_no_opt/builds/24676/steps/test_dnn/logs/stdio

I think I have tested the API successfully, and you can see more information in https://github.com/HannibalAPE/opencv/blob/text_det_recog_demo/doc/tutorials/dnn/dnn_scene_text_det_and_rec/scene_text_recognition.markdown

samples/dnn/scene_text_detection.cpp
bhack: Yes check only for test.

modules/dnn/test/test_model.cpp
alalek: > I still get an error in "Linux x64 Debug", but I find that the test_dnn in "Linux x64" passed. I am not sure about the difference between these two tests.

Obviously, it is "Debug" mode (with extra checks).

HannibalAPE: @alalek Thanks for your reminder.

vpisarev:
@HannibalAPE, I'm now trying to run the code. None of the models you provided can be read with the version of OpenCV that is in text_det_recog_demo. BTW, hash sum for onnx/models/DB_IC15_resnet50.onnx does not match as well when I run download_models.py script. Could you please check that the models are imported correctly. Make sure that in CMake the version of protobuf from OpenCV is used, not the system version.

HannibalAPE:
> @HannibalAPE, I'm now trying to run the code. None of the models you provided can be read with the version of OpenCV that is in text_det_recog_demo. BTW, hash sum for onnx/models/DB_IC15_resnet50.onnx does not match as well when I run download_models.py script. Could you please check that the models are imported correctly. Make sure that in CMake the version of protobuf from OpenCV is used, not the system version.

@vpisarev 
1. I have updated the sha of DB_IC15_resnet50.onnx. I forgot to update it when I updated the model. sry.
2. I compile it again, and I am sure that my models can be imported.
I checked the protobuf version in the CMAKE.
```
  Other third-party libraries:
    Lapack:                      YES (/usr/lib/x86_64-linux-gnu/liblapack.so /usr/lib/x86_64-linux-gnu/libcblas.so /usr/lib/x86_64-linux-gnu/libatlas.so)
    Eigen:                       NO
    Custom HAL:                  NO
    Protobuf:                    build (3.5.1)
```
Is it not right? 
When I compile it, only the ippicv_2020_lnx_intel64_20191018_general.tgz downloading is failed. Dose it matter?

HannibalAPE:
@vpisarev Hi Vadim, I have tested it on other Linux-based systems (e.g. Ubuntu 18.04 and 16.04). On both of them, my models can be imported and output the right results. Can you provide some information about your issue, then I can reproduce it.

vpisarev:
@HannibalAPE, thank you. I've downloaded the latest DB_IC15_resnet50.onnx and it works well. It's noticeably slower than EAST detector, but the results are definitely better!

However, with the model DB_TD500_resnet50.onnx it still crashes with the following message:

```
[ERROR:0] global /Users/vpisarev/work/opencv/modules/dnn/src/dnn.cpp (3272) getLayerShapesRecursively OPENCV/DNN: [Reshape]:(712): getMemoryShapes() throws exception. inputs=1 outputs=1/1 blobs=0
[ERROR:0] global /Users/vpisarev/work/opencv/modules/dnn/src/dnn.cpp (3275) getLayerShapesRecursively     input[0] = [ 1 1 736 1280 ]
[ERROR:0] global /Users/vpisarev/work/opencv/modules/dnn/src/dnn.cpp (3279) getLayerShapesRecursively     output[0] = [ ]
[ERROR:0] global /Users/vpisarev/work/opencv/modules/dnn/src/dnn.cpp (3285) getLayerShapesRecursively Exception message: OpenCV(4.4.0-pre) /Users/vpisarev/work/opencv/modules/dnn/src/layers/reshape_layer.cpp:113: error: (-215:Assertion failed) total(srcShape, srcRange.start, srcRange.end) == maskTotal in function 'computeShapeByReshapeMask'

libc++abi.dylib: terminating with uncaught exception of type cv::Exception: OpenCV(4.4.0-pre) /Users/vpisarev/work/opencv/modules/dnn/src/layers/reshape_layer.cpp:113: error: (-215:Assertion failed) total(srcShape, srcRange.start, srcRange.end) == maskTotal in function 'computeShapeByReshapeMask'

Abort trap: 6
```

I'm using macOS 10.15.6, xcode 11.6. Will try it on Linux tomorrow or on Friday.

Some other complains:

As I said, the speed is not that good. I tried to play with "--inputWidth" and "--inputHeight" parameters.

1. First of all, it's inconvenient that you need to set both these fields. I would compute "height" out of "width" for example if just "width" is set and vice versa.
2. Ok, when I specify both parameters, inputWidth, inputHeight, the app crashes again (with DB_IC15_resnet50.onnx model):

```
V16M:Release vpisarev$ ./example_dnn_scene_text_detection --mp=DB_IC15_resnet50.onnx --inputWidth=640 --inputHeight=368
[ERROR:0] global /Users/vpisarev/work/opencv/modules/dnn/src/dnn.cpp (3272) getLayerShapesRecursively OPENCV/DNN: [Eltwise]:(551): getMemoryShapes() throws exception. inputs=2 outputs=0/1 blobs=0
[ERROR:0] global /Users/vpisarev/work/opencv/modules/dnn/src/dnn.cpp (3275) getLayerShapesRecursively     input[0] = [ 1 256 24 40 ]
[ERROR:0] global /Users/vpisarev/work/opencv/modules/dnn/src/dnn.cpp (3275) getLayerShapesRecursively     input[1] = [ 1 256 23 40 ]
[ERROR:0] global /Users/vpisarev/work/opencv/modules/dnn/src/dnn.cpp (3285) getLayerShapesRecursively Exception message: OpenCV(4.4.0-pre) /Users/vpisarev/work/opencv/modules/dnn/src/layers/eltwise_layer.cpp:216: error: (-215:Assertion failed) inputs[0][j] == inputs[i][j] in function 'getMemoryShapes'

libc++abi.dylib: terminating with uncaught exception of type cv::Exception: OpenCV(4.4.0-pre) /Users/vpisarev/work/opencv/modules/dnn/src/layers/eltwise_layer.cpp:216: error: (-215:Assertion failed) inputs[0][j] == inputs[i][j] in function 'getMemoryShapes'
```
can you please check if inputWidth/inputHeight work at all?

==

The text recognition sample says it supports live video capture from camera, but if I run it without image, it complains that the image is not set. From the code I can conclude that it does not support live video capture. Can you modify scene_text_detection sample to support recognition as well? It would be very useful demonstration on how to use detection and recognition together.

HannibalAPE:
> However, with the model DB_TD500_resnet50.onnx it still crashes with the following message:

@vpisarev It may be caused by the wrong input size. For DB_TD500_resnet50.onnx, the image shape should be set to 736x736, which is mentioned in both the tutorial and the sample.

> when I specify both parameters, inputWidth, inputHeight, the app crashes again

Currently, it only supports a predefined shape. I will try to update it to support dynamic shape.

> It would be very useful demonstration on how to use detection and recognition together.

This sample and some new tutorials will be pushed in this weak.

doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown
```diff
@@ -0,0 +1,366 @@
...
+Text Detection:
+url: https://drive.google.com/uc?export=dowload&id=149tAhIcvfCYeyufRoZ9tmc2mZDKE_XrF
+sha: ced3c03fb7f8d9608169a913acf7e7b93e07109b
+```
+
+## Example for Text Recognition
+
+```cpp
+#include <iostream>
+#include <fstream>
```
alalek, line 119: Create standalone .cpp file for example code and embed snippets into documentation only.

You can adopt sample file below or create new one.

doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown
```diff
@@ -0,0 +1,366 @@
...
+
+    model.setInputParams(scale, inputSize, mean);
+
+    // more decoding method will be supported in future
+    String decodeType = "CTC-greedy";
+
+    // Load a cropped text line image
+    // you can find cropped images for testing in "Images for Testing"
+    int rgb = 1; // This should be changed according to the model input requirement.
+    Mat image = imread("path/to/text_rec_test.png", rgb);
```
alalek, line 159: > 1

Avoid magic numbers.
`IMREAD_COLOR`

doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown
```diff
@@ -0,0 +1,366 @@
...
+    CV_Assert(vocFile.is_open());
+    String vocLine;
+    std::vector<String> vocabulary;
+    while (std::getline(vocFile, vocLine)) {
+        vocabulary.push_back(vocLine);
+    }
+
+    // Normalization parameters
+    double scale = 1.0 / 127.5;
+    Scalar mean = Scalar(127.5);
```
alalek, line 146: `127.5, 0, 0, 0`

or

`127.5, 127.5, 127.5, 127.5`

?

doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown
```diff
@@ -0,0 +1,366 @@
...
+    model.recognize(image, decodeType, vocabulary, recResults);
+    std::cout << recResults[0] << std::endl;
+
+    return 0;
+}
+
+```
+Input image:
+
+![Picture example](https://github.com/HannibalAPE/images/blob/master/opencv/text_rec_test.png)
```
alalek, line 172: External links on images are not allowed.
Make image as a part of documentation tree instead (but please avoid adding images with bad license terms)

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,111 @@ CV__DNN_INLINE_NS_BEGIN
...
+          * @param[in] decodeType: The decoding method of translating the network output into string.
+          * @param[in] vocabulary: The vocabulary for decoding.
+          * @param[out] results: A set of text recognition results.
+          * @param[in] roiPolygons: A set of text detection results.
+          */
+         CV_WRAP void recognize(InputArray frame, String decodeType, const std::vector<String> & vocabulary,
+                                CV_OUT std::vector<String> & results,
+                                const std::vector<std::vector<Point>> & roiPolygons = std::vector<std::vector<Point>>());
+
+     private:
```
alalek, line 1320: Exposing of private API is not allowed in OpenCV Public headers.
You can use PImpl pattern with static `create()` or static local functions.

alalek (commented):
Thank you for contribution!

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: I have changed the private functions into static local functions in the dnn/src/model.cpp
Thank you for your help, and please check whether it is ok.

doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown
HannibalAPE: I am not sure about your confusion here.
When I specify the `mean` parameter to `Scalar(127.5)`, all the values of the input image will subtract the mean (127.5) in the normalization step.

alalek:
> doc/tutorials/dnn/dnn_text_spotting/detect_test1.png
> doc/tutorials/dnn/dnn_text_spotting/detect_test2.png

PNG lossless format is not necessary here (files are large for real world images). Try to reduce image size using the `.jpg` format

alalek:
> doc/tutorials/dnn/dnn_text_spotting/text_recognition.cpp

Sample code for tutorials should to into samples/cpp/tutorial_code/...

doc/tutorials/dnn/table_of_content_dnn.markdown
```diff
@@ -79,4 +79,14 @@ Deep Neural Networks (dnn module) {#tutorial_table_of_content_dnn}
...
 
-    In this tutorial you will learn how to use opencv_dnn module using custom OCR models.
\ No newline at end of file
+    In this tutorial you will learn how to use opencv_dnn module using custom OCR models.
+
+-   @subpage tutorial_dnn_text_spotting
+
+    *Languages:* C++
+
+    *Compatibility:* \> OpenCV 4.3
```
alalek, line 88: 4.5 ?

samples/dnn/scene_text_detection.cpp
```diff
@@ -0,0 +1,167 @@
...
+        // Create a window for showing groundtruth
+        static const std::string winNameGT = "GT";
+        namedWindow(winNameGT, WINDOW_NORMAL);
+
+        String testImgPath;
+        while (std::getline(testList, testImgPath)) {
+            String imgPath = evalDataPath + "/test_images/" + testImgPath;
+            // std::cout << "Image Path: " << imgPath << std::endl;
+
+            Mat frame = imread(imgPath, 1);
```
alalek, line 103: > 1

Please avoid using of magic numbers.
Use `IMREAD_COLOR` instead.

samples/dnn/scene_text_detection.cpp
```diff
@@ -0,0 +1,167 @@
...
+            }
+            polylines(src, gts, true, Scalar(0, 255, 0), 2);
+            imshow(winNameGT, src);
+
+            waitKey();
+        }
+    } else {
+        // Open an image file
+        CV_Assert(parser.has("inputImage"));
+        Mat frame = imread(parser.get<String>("inputImage"));
```
alalek, line 154: Please use `samples::findFile()` for better file searching experience.

alalek (commented):
Looks good to me!

Please take a look on left comments

alalek:
@dkurt Please take look on public API proposals.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,83 @@ CV__DNN_INLINE_NS_BEGIN
...
+         /**
+          * @brief Create model from deep learning network.
+          * @param[in] network Net object.
+          */
+         CV_WRAP TextRecognitionModel(const Net& network);
+
+         /**
+          * @brief Given the @p input frame, create input blob, run net and return recognition result.
+          * @param[in] frame: The input image.
+          * @param[in] decodeType: The decoding method of translating the network output into string.
```
dkurt, line 1311: Which are possible values?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,83 @@ CV__DNN_INLINE_NS_BEGIN
...
+
+         /**
+          * @brief Given the @p input frame, create input blob, run net and return detection results.
+          * @param[in] frame: The input image.
+          * @param[out] results: A set of text detection results.
+          * @param[in] outputType: The output type: {0: multi-oriented quadrilateral; 1: polygon}
+          * @param[in] binThresh: The threshold of the binary map.
+          * @param[in] polyThresh: The threshold of text polygons.
+          * @param[in] unclipRatio: The unclip ratio of the detected text region.
+          * @param[in] maxCandidates: The max number of the output polygons.
```
dkurt, line 1354: Why we need to limit this number?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,83 @@ CV__DNN_INLINE_NS_BEGIN
...
+         CV_WRAP TextDetectionModel(const Net& network);
+
+         /**
+          * @brief Given the @p input frame, create input blob, run net and return detection results.
+          * @param[in] frame: The input image.
+          * @param[out] results: A set of text detection results.
+          * @param[in] outputType: The output type: {0: multi-oriented quadrilateral; 1: polygon}
+          * @param[in] binThresh: The threshold of the binary map.
+          * @param[in] polyThresh: The threshold of text polygons.
+          * @param[in] unclipRatio: The unclip ratio of the detected text region.
```
dkurt, line 1353: What does unclip ratio?

modules/dnn/test/test_model.cpp
```diff
@@ -389,6 +428,44 @@ TEST_P(Test_Model, Segmentation)
...
+    Scalar mean = Scalar(122.67891434, 116.66876762, 104.00698793);
+
+    float binThresh = 0.3;
+    float polyThresh = 0.5;
+    uint maxCandidates = 200;
+    int outputType = 0;
+    double unclipRatio = 2.0;
+
+    testTextDetectionModel(weightPath, "", imgPath, gt, outputType, binThresh, polyThresh, maxCandidates, unclipRatio, size, mean, scale);
+}
```
dkurt, line 467: Please test existing EAST model as well. Does it fit current API?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,83 @@ CV__DNN_INLINE_NS_BEGIN
...
+          * @brief Given the @p input frame, create input blob, run net and return recognition result.
+          * @param[in] frame: The input image.
+          * @param[in] decodeType: The decoding method of translating the network output into string.
+          * @param[in] vocabulary: The vocabulary for decoding.
+          * @param[out] results: A set of text recognition results.
+          * @param[in] roiPolygons: A set of text detection results.
+          */
+         CV_WRAP void recognize(InputArray frame, String decodeType, const std::vector<String> & vocabulary,
+                                CV_OUT std::vector<String> & results,
+                                const std::vector<std::vector<Point>> & roiPolygons = std::vector<std::vector<Point>>());
```
dkurt, line 1318: No need to put vocabulary at recognition step. Add a method such `setVocabulary`

modules/dnn/test/test_model.cpp
dkurt: https://github.com/opencv/opencv/blob/b45273eccb01323f5615b8900fd35407b1148939/modules/dnn/test/test_tf_importer.cpp#L817

samples/dnn/scene_text_spotting.cpp
```diff
@@ -0,0 +1,125 @@
...
+        "{ inputHeight ih                   |736| image height of the model input. }"
+        "{ inputWidth iw                    |1280| image width of the model input. }"
+        "{ RGBInput rgb                     |0| 0: imread with flags=IMREAD_GRAYSCALE; 1: imread with flags=IMREAD_COLOR. }"
+        "{ binaryThreshold bt               |0.3| Confidence threshold of the binary map. }"
+        "{ polygonThreshold pt              |0.5| Confidence threshold of polygons. }"
+        "{ maxCandidate max                 |200| Max candidates of polygons. }"
+        "{ outputType type                  |0| The output type of detected text: 0: multi-oriented quadrilateral; 1: polygon }"
+        "{ unclipRatio ratio                |2.0| unclip ratio. }"
+        "{ vocabularyPath vp                | | Path to benchmarks for evaluation. "
+            "Download links are provided in doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown}";
```
dkurt, line 27: Is that possible to rewrite https://github.com/opencv/opencv/blob/master/samples/dnn/text_detection.cpp instead? This way we can proof that new API is a good replacement for old approach.

samples/cpp/tutorial_code/dnn/text_spotting/text_spotting.cpp
```diff
@@ -0,0 +1,82 @@
+#include <iostream>
```
dkurt, line 1: Ad for me, too much samples with the same meaning - high level API should be as flexible as possible

doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown
berak: `Scalar(127.5)` only initializes the 1st element, so that would be (127.5, 0, 0, 0)

samples/cpp/tutorial_code/dnn/text_spotting/text_spotting.cpp
HannibalAPE: I have deleted the samples in samples/cpp/tutorial_code/dnn/text_spotting, and only embed snippets in the tutorial.

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: I have modified the `TextRecognitionModel` class, please check it again.

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: It only supports "CTC-greedy" now, and I will add more decoding methods in the future, such as Beam Searching.

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: We can get good detection results with these parameters, for more information you can refer to the [paper](https://arxiv.org/abs/1911.08947) and [code](https://github.com/MhLiao/DB).

HannibalAPE: In some difficult cases, the output map of the network is full of small noise, and `maxCandidates` can avoid wasting inference time.

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: This detection method is to detect the kernel of each text instance, so we should "expand" the kernel into the final result by the 'unclipRatio'.

samples/dnn/scene_text_spotting.cpp
HannibalAPE: The methods of scene text detection can be divided into (1) regression-based method and (2) segmentation-based method.
The original method (EAST) in text_detection.cpp is a regression-based method, while the new method is a segmentation-based method. Different post-processing is needed in the inference for them.

The methods of scene text recognition can be divided into (1) CTC-based method and (2) Attention-based method.
All of the samples adopts the CTC-based method (CRNN) for recognition because the current DNN module can not support some layers. In the following months, I will put some issues about them which are quite needed.

As you see, this is the first version of APIs for text detection and recognition. I will keep working on it and add more methods to make it as flexible as possible in the future.

doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown
HannibalAPE: Thank you.
I have tested it and changed it to `Scalar(127.5, 127.5, 127.5)`.

samples/dnn/scene_text_detection.cpp
HannibalAPE: I have updated all `imread` with `samples::findFile()`

doc/tutorials/dnn/table_of_content_dnn.markdown
HannibalAPE: I have changed it to 4.5

samples/dnn/scene_text_detection.cpp
HannibalAPE: I will check all my code to avoid it, thank you.

modules/dnn/include/opencv2/dnn/dnn.hpp
dkurt: User should know the options so it must be documented or we can add enum

modules/dnn/include/opencv2/dnn/dnn.hpp
dkurt: I mean that this line is then converted to documentation so users should know how this parameter changes the behavior

samples/dnn/scene_text_spotting.cpp
dkurt: I just worry that we introduce an API which is not generic. Initial idea for the High Level API to help users start fast. Initially, we designed it for object detection networks which also have different post-processing depends on network family.

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: I have updated the parameter description for a better understanding.

samples/dnn/scene_text_spotting.cpp
HannibalAPE: I understand it, and I also have the aspiration to bring the latest topologies and algorithms to OpenCV.
As I discussed with the mentors before GSoC, we agreed on the current APIs to be the first version. I think it is a work in progress, and it is needed to be updated with more algorithms in the future.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,96 @@ CV__DNN_INLINE_NS_BEGIN
...
+     protected:
+         std::vector<String> vocabulary;
+
+         /**
+          * @brief decode the output of the network.
+          * @param[in] prediction: The output of the network.
+          * @param[in] decodeType: The decoding method of translating the network output into string.
+          * @returns a string holding the decoded recognition result
+          */
+         CV_WRAP String decode(Mat prediction, String decodeType);
```
dkurt, line 1332: `const Mat& prediction, const String& decodeType`

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,96 @@ CV__DNN_INLINE_NS_BEGIN
...
+         CV_WRAP void setVocabulary(const std::vector<String> & voc);
+
+         /**
+          * @brief Given the @p input frame, create input blob, run net and return recognition result.
+          * @param[in] frame: The input image.
+          * @param[in] decodeType: The decoding method of translating the network output into string: {'CTC-greedy': greedy decoding for the output of CTC-based methods}
+          * @param[out] results: A set of text recognition results.
+          * @param[in] roiPolygons: (Optional) A set of text detection results. If it is not empty, all the ROIs will be cropped as the network inputs.
+          */
+         CV_WRAP void recognize(InputArray frame, String decodeType, CV_OUT std::vector<String> & results,
```
dkurt, line 1321: `const String&`

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,96 @@ CV__DNN_INLINE_NS_BEGIN
...
+                                const std::vector<std::vector<Point>> & roiPolygons = std::vector<std::vector<Point>>());
```
dkurt, line 1322: `= {}`

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,96 @@ CV__DNN_INLINE_NS_BEGIN
...
+         CV_WRAP TextDetectionModel(const Net& network);
+
+         /**
+          * @brief Given the @p input frame, create input blob, run net and return detection results. For more information about the hyper-parameters, please refer to https://github.com/MhLiao/DB.
+          * @param[in] frame: The input image.
+          * @param[out] results: A set of text detection results.
+          * @param[in] outputType: The output shape type: {0: multi-oriented quadrilateral; 1: polygon}
+          * @param[in] binThresh: The threshold of the binary map. It can be set in [0, 1.0], and is usually set to 0.3 in the official code.
+          * @param[in] polyThresh: The threshold of text polygons. It can be set in [0, 1.0], and is usually set to 0.5, 0.6, and 0.7 in the official code.
+          * @param[in] unclipRatio: The unclip ratio of the detected text region, which determines the output size. It is set to 2.0 in the official code.
```
dkurt, line 1366: what does it mean "official code"?

modules/dnn/src/model.cpp
```diff
@@ -401,4 +401,242 @@ void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
...
+TextRecognitionModel::TextRecognitionModel(const String& model, const String& config)
+    : Model(model, config) {}
+
+TextRecognitionModel::TextRecognitionModel(const Net& network) : Model(network) {}
+
+void TextRecognitionModel::setVocabulary(const std::vector<String> & voc)
+{
+    this->vocabulary.clear();
+    for (uint i = 0; i < voc.size(); i++) {
+        this->vocabulary.push_back(voc.at(i));
```
dkurt, line 413: do not use `this->` for class fields

modules/dnn/src/model.cpp
```diff
@@ -401,4 +401,242 @@ void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
...
+    results.clear();
+
+    std::vector<Mat> outs;
+    uint roiSize = roiPolygons.size();
+    if (roiSize == 0) {
+        impl->predict(*this, frame.getMat(), outs);
+        CV_Assert(outs.size() == 1);
+        Mat prediction = outs[0];
+        String result = decode(prediction, decodeType);
+        results.push_back(result);
```
dkurt, line 429: please reduce code duplication - create a vector of inputs and process it after

modules/dnn/src/model.cpp
```diff
@@ -401,4 +401,242 @@ void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
...
+            for (uint j = 0; j < roiPolygons[i].size(); j++) {
+                if (xmin > roiPolygons[i][j].x) xmin = roiPolygons[i][j].x;
+                if (xmax < roiPolygons[i][j].x) xmax = roiPolygons[i][j].x;
+                if (ymin > roiPolygons[i][j].y) ymin = roiPolygons[i][j].y;
+                if (ymax < roiPolygons[i][j].y) ymax = roiPolygons[i][j].y;
+            }
+            if (xmin < 0) xmin = 0;
+            if (ymin < 0) ymin = 0;
+            if (xmax > input.cols) xmax = input.cols - 1;
+            if (ymax > input.rows) ymax = input.rows - 1;
```
dkurt, line 445: Just use `std::min` and `std::max`

samples/dnn/scene_text_spotting.cpp
dkurt: > we agreed on the current APIs to be the first version
we cannot merge something draft in `opencv/opencv`. API should be scalable and at least cover all the known topologies. There are multiple text detection and recognition network already supported by OpenCV (see the notes from https://github.com/opencv/opencv/blob/master/samples/dnn/text_detection.cpp). So introduced API should manage them.

HannibalAPE: Ok, I will add those topologies to my API in this month. 
I have to prepare for a test this week.

modules/dnn/test/test_model.cpp
HannibalAPE: The API can support EAST now, and there is a code testing for EAST in test_model.cpp.
`TEST_P(Test_Model, TextDetectionByEAST)`

I have applied my APIs to samples/dnn/text_detection.cpp and tested it successfully with the previous models.
But, I am not sure about how to change the file name, because the text_detection.cpp and scene_text_spotting.cpp do the same thing with different algorithms. Do you have any suggestions?

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: I have checked all my codes to update some input parameters with `const &`.

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: I have put the official code's URL of this algorithm at the top of these comments.

modules/dnn/src/model.cpp
HannibalAPE: I have updated it. Please check whether I get your mind. Thank you.

modules/dnn/src/model.cpp
```diff
@@ -401,4 +407,310 @@ void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
...
+            scores.push_back(score);
+        }
+    }
+
+    // Apply non-maximum suppression procedure.
+    std::vector<int> indices;
+    NMSBoxes(boxes, scores, confThreshold, nmsThreshold, indices);
+
+    // Re-scale
+    Point2f ratio((float)frame.cols() / impl->size.width, (float)frame.rows() / impl->size.width);
```
berak, line 612: shouldn't it be:

    (float)frame.rows() / impl->size.height;

HannibalAPE: Yes, I am sorry about it. The input width and height are the same, so I did not find it.

vpisarev:
:+1:

samples/dnn/text_detection.cpp
```diff
@@ -2,22 +2,14 @@
     Text detection model: https://github.com/argman/EAST
     Download link: https://www.dropbox.com/s/r2ingd0l3zt8hxs/frozen_east_text_detection.tar.gz?dl=1
 
-    CRNN Text recognition model taken from here: https://github.com/meijieru/crnn.pytorch
-    How to convert from pb to onnx:
-    Using classes from here: https://github.com/meijieru/crnn.pytorch/blob/master/models/crnn.py
```
alalek, line 7: Why is this information dropped?

alalek (changes requested):
Discussion is required

samples/dnn/text_detection.cpp
HannibalAPE: The previous models are taken from this link, but I also provide some other models.
How about put this information into the tutorial in detail? or just leave it here?
Do you have any suggestions?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,130 @@ CV__DNN_INLINE_NS_BEGIN
...
+          * @param[in] frame: The input image.
+          * @param[in] decodeType: The decoding method of translating the network output into string: {'CTC-greedy': greedy decoding for the output of CTC-based methods}
+          * @param[out] results: A set of text recognition results.
+          * @param[in] roiPolygons: (Optional) A set of text detection results. If it is not empty, all the ROIs will be cropped as the network inputs.
+          */
+         CV_WRAP void recognize(InputArray frame, const String& decodeType, CV_OUT std::vector<String>& results,
+                                const std::vector<std::vector<Point>>& roiPolygons = {});
+
+     protected:
+         std::vector<String> vocabulary;
```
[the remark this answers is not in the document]
HannibalAPE: I feel confused here, because the `vocabulary` is a member of the `TextRecognitionModel` class.
If both of the private and protected members are not allowed, I should define such details as static local variants/functions?
For a better understanding, can you show me any proper implementation manners in other existing files?

samples/dnn/scene_text_detection.cpp
```diff
@@ -0,0 +1,167 @@
...
+ *      Minghui Liao    <Huazhong University of Science and Technology>
+ *      Zhaoyi Wan      <Megvii>
+ *      Cong Yao        <Megvii>
+ *      Kai Chen        <Shanghai Jiao Tong University>
+ *      Xiang Bai       <Huazhong University of Science and Technology>, Corresponding author
+ *
+ * This script is written by Wenqing Zhang <Huazhong University of Science and Technology>.
+ *
+ * This code has been contributed to OpenCV under the terms of Apache 2 license:
+ * https://www.apche.org/licenses/LICENSE-2.0
```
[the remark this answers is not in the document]
HannibalAPE: sry, I will check it again.

modules/dnn/include/opencv2/dnn/dnn.hpp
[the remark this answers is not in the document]
dkurt: Something like this: https://github.com/opencv/opencv/blob/c3e8a82c9c0b18432627fe100db332b42d64b8d3/modules/dnn/src/model.cpp#L16

modules/dnn/src/model.cpp
```diff
@@ -37,7 +37,13 @@ struct Model::Impl
             Mat imInfo = (Mat_<float>(1, 3) << size.height, size.width, 1.6f);
             net.setInput(imInfo, "im_info");
         }
-        net.forward(outs, outNames);
+
+        if (outputNames.empty()) {
+            net.forward(outs, outNames);
+        } else {
+            net.forward(outs, outputNames);
+        }
```
dkurt, line 45: Why we need this?

modules/dnn/src/model.cpp
```diff
@@ -401,4 +407,310 @@ void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
...
+    : Model(model, config) {}
+
+TextRecognitionModel::TextRecognitionModel(const Net& network) : Model(network) {}
+
+void TextRecognitionModel::setVocabulary(const std::vector<String>& voc)
+{
+    vocabulary.clear();
+    for (uint i = 0; i < voc.size(); i++) {
+        vocabulary.push_back(voc.at(i));
+    }
```
dkurt, line 420: Please use `stl` - resize and insert

modules/dnn/src/model.cpp
HannibalAPE: I wrote it to keep consistent with the previous `text_deteciton.cpp`, and it gives users a choice to specify the output names.
Maybe it is not necessary, and I have removed it.

samples/dnn/text_detection.cpp
HannibalAPE: I have added the necessary information to my tutorial.

modules/dnn/include/opencv2/dnn/dnn.hpp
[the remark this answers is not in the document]
HannibalAPE: I have re-implemented my API in this way, please check it. 
Thank you for your help.

samples/dnn/scene_text_detection.cpp
```diff
@@ -0,0 +1,167 @@
+/*
+ * Real-time Scene Text Detection with Differentiable Binarization
+ * Copyright (C) 2020, The patent is owned by <Huazhong University of Science and Technology>
```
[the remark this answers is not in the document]
HannibalAPE: Do I need to add some other information?
Pattent Name: 一种基于可微分二值化的实时文本检测方法
Publication Number: CN110781967A

modules/dnn/src/model.cpp
HannibalAPE: We need to set the `vocabulary` with the input `voc`, so I think `assign` is a better choice than `insert`.
Please check it and let me know if it is not good. Thank you for your help.

samples/dnn/scene_text_detection.cpp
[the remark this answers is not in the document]
HannibalAPE: @vpisarev Please take a look on it.
Thank you for your suggestions in advance.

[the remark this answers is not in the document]
HannibalAPE: @alalek I have talked with Vadim by email, and I have removed the patent information.

samples/dnn/text_detection.cpp
HannibalAPE: I have put these information back to it.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,101 @@ CV__DNN_INLINE_NS_BEGIN
...
+
+         /**
+          * @brief Given the @p input frame, create input blob, run net and return recognition result.
+          * @param[in] frame: The input image.
+          * @param[in] decodeType: The decoding method of translating the network output into string: {'CTC-greedy': greedy decoding for the output of CTC-based methods}
+          * @param[out] results: A set of text recognition results.
+          * @param[in] roiPolygons: (Optional) A set of text detection results. If it is not empty, all the ROIs will be cropped as the network inputs.
+          */
+         CV_WRAP void recognize(InputArray frame, const String& decodeType, CV_OUT std::vector<String>& results,
+                                const std::vector<std::vector<Point>>& roiPolygons = {});
```
alalek, line 1322: API looks over-complicated for common class.
Not sure that we really need to support custom **contours** as part of such basic API.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1281,6 +1281,101 @@ CV__DNN_INLINE_NS_BEGIN
...
+         CV_WRAP void setVocabulary(const std::vector<String>& inputVoc);
+
+         /**
+          * @brief Given the @p input frame, create input blob, run net and return recognition result.
+          * @param[in] frame: The input image.
+          * @param[in] decodeType: The decoding method of translating the network output into string: {'CTC-greedy': greedy decoding for the output of CTC-based methods}
+          * @param[out] results: A set of text recognition results.
+          * @param[in] roiPolygons: (Optional) A set of text detection results. If it is not empty, all the ROIs will be cropped as the network inputs.
+          */
+         CV_WRAP void recognize(InputArray frame, const String& decodeType, CV_OUT std::vector<String>& results,
```
alalek, line 1321: > decodeType

This is algorithm / model parameter.
(it is not intended to be changed between multiple "recognize" calls for different frames).

Must be moved into constructor / `::create()` methods and `.get` (may be `.set`) properties.

---

/cc @dkurt @l-bat @vpisarev

modules/dnn/include/opencv2/dnn/dnn.hpp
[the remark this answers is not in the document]
alalek: > Ptr<Voc> voc;

Please keep only one "implementation" pointer (it is already in Model::impl).

(perhaps "Bridge" design pattern should be applied here)

samples/dnn/scene_text_detection.cpp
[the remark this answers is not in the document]
alalek: If there is a reference/link on patented DNN model in this patch then it is still better to provide such type of information near the corresponding link (instead of whole sample).

alalek:
@HannibalAPE If it is not clear how to process comments about API changes, then let us know (we will prepare updates on our side)

modules/dnn/src/model.cpp
```diff
@@ -401,4 +402,322 @@ void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
...
+
+    Mat mask = Mat::zeros(ymax - ymin + 1, xmax - xmin + 1, CV_8U);
+    std::vector<Point> roiContour;
+    for (size_t i = 0; i < contour.size(); i++) {
+        Point pt = Point(contour[i].x - xmin, contour[i].y - ymin);
+        roiContour.push_back(pt);
+    }
+    std::vector<std::vector<Point>> roiContours = {roiContour};
+    fillPoly(mask, roiContours, Scalar(1));
+    double score = mean(binROI, mask).val[0];
```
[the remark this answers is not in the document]
HannibalAPE: Is there any better solutions? The official post-processing uses this function to do it
> https://github.com/MhLiao/DB/blob/f08c44faa1b4790039381203a8079d73a77494b4/structure/representers/seg_detector_representer.py#L203

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: The idea of adding this optional parameter is proposed by Vadim in the last months.
I think it can be another choice for users?

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: I moved `decodeType` into `Voc`, but I have noticed that you propose to update cv::dnn::Model class in https://github.com/opencv/opencv/pull/18711.
Hence, I need to wait for that pr to be merged, and update my codes?

HannibalAPE:
> @HannibalAPE If it is not clear how to process comments about API changes, then let us know (we will prepare updates on our side)

@alalek 
If I understand your meaning rightly, I think I need your help.
Before that, I need to squash the previous commits, right?

modules/dnn/src/model.cpp
alalek: Just `vocabulary = voc;` should work fine.

modules/dnn/include/opencv2/dnn/dnn.hpp
alalek: Main concern here:
- this contours parameter is optional for User, but it is not optional for extending implementation for supporting of other DL models which can't work with contours.

alalek:
> I need to squash the previous commits

Not necessary. As you wish.

Finally, commits will be squashed during merge.

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: ok, I understant that, and have removed it.

HannibalAPE:
@alalek 
I have no idea about how to update the `struct Voc` without modification to `struct Impl`.
I tried to define `TextRecognition::Voc` Inherited from `Model::Impl`, but I do not how to apply "Bridge" design pattern here.
Do you have any suggestions about it? Is there any example?
Thank you.

alalek (commented):
@HannibalAPE Thank you for contribution!
I will update introduced public APIs and push them here till the end of this week.

---

**Update**: Still work is in progress.

samples/dnn/scene_text_recognition.cpp
```diff
@@ -52,7 +52,7 @@ int main(int argc, char** argv)
     // Load vocabulary
     CV_Assert(!vocPath.empty());
     std::ifstream vocFile;
-    vocFile.open(vocPath);
+    vocFile.open(samples::findFile(vocPath));
```
HannibalAPE, line 55: I wonder whether I need to add `samples::findFile()` whenever I open files?
Do I need to change `TextRecognitionModel recognizer(modelPath)` into `TextRecognitionModel recognizer(samples::findFile(modelPath))` ?

HannibalAPE (commented):
@alalek Thank you! I learn a lot with your help.

samples/dnn/scene_text_recognition.cpp
alalek: `samples::findFile()` helps to use file by name (instead of full path) from `<opencv>/samples/data` location.

Model file is not stored there (and no plans to put it there due its size).

modules/dnn/src/model.cpp
```diff
@@ -487,4 +489,669 @@ void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
...
+    void setMaxCandidates(int maxCandidates_) { maxCandidates = maxCandidates_; }
+    int getMaxCandidates() const { return maxCandidates; }
+
+
+    virtual
+    std::vector<cv::RotatedRect> detect(InputArray frame, CV_OUT std::vector<float>& confidences) CV_OVERRIDE
+    {
+        CV_TRACE_FUNCTION();
+        std::vector< std::vector<Point> > contours = detectTextContours(frame);
+        confidences = std::vector<float>(contours.size(), 1.0f);
```
alalek, line 930: Is there any `confidences` scoring in DB detection algo?

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1309,6 +1309,227 @@ CV__DNN_INLINE_NS_BEGIN
...
+
+    /** @brief Performs detection
+     *
+     * Given the input @p frame, prepare network input, run network inference, post-process network output and return result detections.
+     *
+     * @param[in] frame The input image.
+     * @return array with text detection results
+     */
+    CV_WRAP
+    std::vector< std::vector<Point> > detectTextContours(InputArray frame) const;
```
alalek, line 1530: There is an issue with ObjC binding generator with contours handling (`std::vector< std::vector<Point> >` type).
I will take a look on this on next week.

alalek (commented):
@HannibalAPE 

I pushed updated public API for text detection / recognition tasks. Please take a look and check samples / documentation (probably I miss something).

modules/dnn/src/model.cpp
HannibalAPE: https://github.com/HannibalAPE/opencv/blob/f59aa6d4ae76e08f199e56ae94235295e0380076/modules/dnn/src/model.cpp#L972

You can regard the return of `contourScore()` as a kind of `confidence`, but it is only used to filter some bad detection results. It is not the same meaning as those in general object detection algorithms.

HannibalAPE:
@alalek Since the onnx importer has supported models with dynamic input shape, I will try to generate new models of DB and update them in this week. I will also check samples and tutorials in this week.

samples/dnn/scene_text_spotting.cpp
```diff
@@ -0,0 +1,163 @@
...
+        } else {
+            recInput = frame;
+        }
+        std::vector< std::vector<Point> > contours;
+        for (uint i = 0; i < detResults.size(); i++)
+        {
+            const auto& box = detResults[i];
+
+            Point2f vertices[4] = {};
+            box.points(vertices);
```
HannibalAPE, line 120: I think this function can not sort the points into the right order (bottomLeft, topLeft, topRight, bottomRight) in some cases.
> -i=path/to/evaluation_data_det/IC15/test_images/img_5.jpg
-dmp=path/to/DB_IC15_resnet50.onnx
-iw=1280 -ih=736 -rmp=path/to/crnn_cs.onnx
 -vp=path/to/alphabet_94.txt
-rgb=1

By the above setting, you can reproduce the errors. 
You can find the test image [here](https://drive.google.com/uc?export=dowload&id=149tAhIcvfCYeyufRoZ9tmc2mZDKE_XrF)

modules/dnn/src/model.cpp
```diff
@@ -488,307 +489,669 @@ void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
...
+    void setMaxCandidates(int maxCandidates_) { maxCandidates = maxCandidates_; }
+    int getMaxCandidates() const { return maxCandidates; }
 
-        // Unclip
-        std::vector<Point> approx;
-        RotatedRect box = minAreaRect(contourScaled);
-        Point2f vertex[4];
-        box.points(vertex);
-        for (int j = 0; j < 4; j++) {
-            approx.push_back(Point((int)(vertex[3-j].x), (int)(vertex[3-j].y)));
```
alalek: > 3-j

Please check this part of code too (do we really need to reorder points?)

samples/dnn/scene_text_spotting.cpp
alalek: `minAreaRect()` may return boxes with angle -90.

Please check latest commit with workarounds.

modules/dnn/src/model.cpp
HannibalAPE: I have removed it, and updated the `rotateVec` in `unclip`.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1309,6 +1309,227 @@ CV__DNN_INLINE_NS_BEGIN
...
+    std::vector<cv::RotatedRect> detect(InputArray frame, CV_OUT std::vector<float>& confidences) const;
+};
+
+/** @brief This class represents high-level API for text detection DL networks compatible with EAST model.
+ *
+ * Configurable parameters:
+ * - (float) confThreshold - used to filter boxes by confidences, default: 0.5f
+ * - (float) nmsThreshold - used in non maximum suppression, default: 0.0f
+ */
+class CV_EXPORTS_W_SIMPLE TextDetectionModel_EAST : public TextDetectionModel
```
dkurt, line 1427: Why we need to derive implementations for algorithms? Is seems to me that more optimal introduce a parameter which can indicate EAST or DB model.

alalek: There are different returned results (type).
Also there is different set of configuration parameters. It is really hard to properly handle with checks both/multiple sets in one instance. We should not create GOD classes.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1309,6 +1309,227 @@ CV__DNN_INLINE_NS_BEGIN
...
+public:
+    /** @brief Performs detection
+     *
+     * Given the input @p frame, prepare network input, run network inference, post-process network output and return result detections.
+     *
+     * @param[in] frame the input image
+     * @return array with text detection results
+     */
+    CV_WRAP
+    std::vector<cv::RotatedRect> detect(InputArray frame) const;
```
alalek, line 1407: Actually `RotatedRect` doesn't work well with perspective transformations (like [this one](https://www.pngwing.com/en/free-png-pdprp)).

Perhaps we need 4 points in API with strong order (bl, tl, tr, br - according to `targetVertices`) which should be used with `getPerspectiveTransform()` to get more accurate results.

I will try to update API on this week.

HannibalAPE: I can not open the above link. Can you share it with google drive?

alalek: It is similar to side boxes of cube from [here](https://en.wikipedia.org/wiki/Perspective_(graphical)#/media/File:Perspective1.jpg).

alalek: However, DL models may not work with perspective transformations (EAST output doesn't know anything about that) and they just detect rotated text.

any thoughts?

HannibalAPE: It depends on different DL models. Some methods can output irregular quadrilaterals.

From my side, the perspective transformation is a temporary replacement, or just a choice for the four-point outputs. (rotated boxes and irregular quadrilaterals)

There is a popular and fast text recognition algorithm `ASTER` which adopts `Thin Plate Spline Transformation` in its rectification network. (like [this](https://github.com/ayumiymk/aster.pytorch/blob/master/overview.png)).
I am not sure whether the TPS transformation is implemented in OpenCV, maybe not?
You can refer to [this](https://github.com/ayumiymk/aster.pytorch/blob/be670046c775b54de79766208f0c59321ae1eccf/lib/models/tps_spatial_transformer.py#L98).

By the way, ASTER is also an algorithm from our lab, and we are glad to contribute it to OpenCV.
However, there are some things to do before.
1) support TPS transformation
2) update LSTM in modules/dnn/src/onnx/onnx_importer.cpp
we need to set [these parameters](https://github.com/opencv/opencv/blob/ef0eed8d3c63b64bacc755dc4ffe5358a2cf30a7/modules/dnn/src/onnx/onnx_importer.cpp#L904) non-zero, but it is not supported now.
3) support [GRU layer](https://pytorch.org/docs/stable/generated/torch.nn.GRU.html?highlight=gru#torch.nn.GRU)
4) ...

I suppose to work on it after this PR.

doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown
```diff
@@ -0,0 +1,314 @@
...
+- [samples/dnn/scene_text_recognition.cpp](https://github.com/opencv/opencv/blob/master/samples/dnn/scene_text_recognition.cpp)
+- [samples/dnn/scene_text_detection.cpp](https://github.com/opencv/opencv/blob/master/samples/dnn/scene_text_detection.cpp)
+- [samples/dnn/text_detection.cpp](https://github.com/opencv/opencv/blob/master/samples/dnn/text_detection.cpp)
+- [samples/dnn/scene_text_spotting.cpp](https://github.com/opencv/opencv/blob/master/samples/dnn/scene_text_spotting.cpp)
+
+#### Test with an image
+Examples:
+```bash
+example_dnn_scene_text_recognition -mp=path/to/crnn_cs.onnx -i=path/to/an/image -rgb=1 -vp=/path/to/alphabet_94.txt
+example_dnn_scene_text_detection -mp=path/to/DB_TD500_resnet50.onnx -i=path/to/an/image -ih=736 -iw=736
```
alalek, line 289: > -mp=path/to/DB_TD500_resnet50.onnx
> -ih=736 -iw=736

Please check model parameters here (and above near model download links).
This set performs better: `-ih=736 -iw=1280` (on `IC15/test_images/img_5.jpg`)

BTW, It make sense to put some defaults into `TextDetectionModel_DB` ctor.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1309,6 +1309,255 @@ CV__DNN_INLINE_NS_BEGIN
...
+
+    /** @brief Performs detection
+     *
+     * Given the input @p frame, prepare network input, run network inference, post-process network output and return result detections.
+     *
+     * Each result is quadrangle's 4 points in this order:
+     * - bottom-left
+     * - top-left
+     * - top-right
+     * - bottom-right
```
alalek, line 1408: Added strong requirements for results to avoid "points reordering" in sample code.

modules/dnn/include/opencv2/dnn/dnn.hpp
```diff
@@ -1309,6 +1309,255 @@ CV__DNN_INLINE_NS_BEGIN
...
+     * @note If DL model doesn't support that kind of output then result may be derived from detectTextRectangles() output.
+     *
+     * @param[in] frame The input image
+     * @param[out] detections array with detections' quadrangles (4 points per result)
+     * @param[out] confidences array with detection confidences
+     */
+    CV_WRAP
+    void detect(
+            InputArray frame,
+            CV_OUT std::vector< std::vector<Point> >& detections,
```
alalek, line 1421: Any thoughts about `Point` vs `Point2f`?

modules/dnn/src/model.cpp
```diff
@@ -487,4 +489,792 @@ void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
...
+                    box.angle += 90;
+                else if (box.angle > 0)
+                    box.angle -= 90;
+            }
+
+            Point2f vertex[4];
+            box.points(vertex);  // order: bl, tl, tr, br
+            std::vector<Point2f> approx = { vertex[3], vertex[0], vertex[1], vertex[2] };  // TODO unclip doesn't preserve order?
+            std::vector<Point2f> polygon;
+            unclip(approx, polygon, unclipRatio);
```
alalek, line 1119: @HannibalAPE Could you please check preserving order of vertexes in `unclip()` call?

alalek (commented):
Updated `TextDetectionModel` API:
- added quadrangle support with strong requirement for order of returned points
- dropped `.detectTextCountours()` from `TextDetectionModel_DB` (replaced by quadrangles).

doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown
HannibalAPE: Actually, DB_TD500_resnet50.onnx and DB_IC15_resnet50.onnx are prepared for different datasets (i.e. TD500 and IC15) respectively, which is tricky to perform the best performance in each benchmark. The recommended settings do not cover the above case (use DB_TD500_resnet50.onnx on IC15 images).

If it is needed, I can train a new model with different datasets together. What is your opinion?

alalek: Thanks for explanation. That makes sense.

> I can train a new model

This can be an activity after this PR merge.

modules/dnn/include/opencv2/dnn/dnn.hpp
HannibalAPE: I think `Point` is okay.

modules/dnn/src/model.cpp
```diff
@@ -487,4 +489,778 @@ void DetectionModel::detect(InputArray frame, CV_OUT std::vector<int>& classIds,
...
+        initNet(network);
+    }
+
+    void setConfidenceThreshold(float confThreshold_) { confThreshold = confThreshold_; }
+    float getConfidenceThreshold() const { return confThreshold; }
+
+    void setNMSThreshold(float nmsThreshold_) { nmsThreshold = nmsThreshold_; }
+    float getNMSThreshold() const { return nmsThreshold; }
+
+    // TODO: According to article EAST supports quadrangles output: https://arxiv.org/pdf/1704.03155.pdf
```
HannibalAPE, line 816: You can find that the performance of QUAD is worse than RBOX in Table. 3 and the authors do not provide official code and models.
Some good re-implementations of EAST only support RBOX.
TF: https://github.com/argman/EAST
PyTorch: https://github.com/songdejia/EAST

alalek (approved):
LGTM :+1:

@HannibalAPE @bhack Thank you for contribution!

HannibalAPE:
Thank you. I really appreciate your help. @alalek @vpisarev @dkurt @bhack @berak

doc/tutorials/dnn/dnn_text_spotting/dnn_text_spotting.markdown
xulihang: Is there a tutorial on training custom models and converting existing models to onnx?

PR #17576, opened 2020-06-18
videoio(ximea): fix the detection of the XIMEA library

philippefoubert:
The location of the XIMEA library is not the same in all versions of this software.
The search path is updated to take into account the different locations identified.

### DIFF
--- a/modules/videoio/cmake/detect_ximea.cmake
+++ b/modules/videoio/cmake/detect_ximea.cmake
@@ -6,10 +6,10 @@ if(NOT HAVE_XIMEA)
     endif()
   endif()
   if(X86_64)
-    set(lib_dir "x64")
+    set(lib_dir "API/x64" "API/64bit")
     set(lib_suffix "64")
   else()
-    set(lib_dir "x86")
+    set(lib_dir "API/x86" "API/32bit")
     set(lib_suffix "32")
   endif()
   find_path(XIMEA_INCLUDE "xiApi.h"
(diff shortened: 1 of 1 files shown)


modules/videoio/cmake/detect_ximea.cmake
```diff
@@ -6,10 +6,12 @@ if(NOT HAVE_XIMEA)
     endif()
   endif()
   if(X86_64)
-    set(lib_dir "x64")
+    set(lib_dir1 "API/x64")
+    set(lib_dir2 "API/64bit")
     set(lib_suffix "64")
   else()
-    set(lib_dir "x86")
+    set(lib_dir1 "API/x86")
```
asmorkalov, line 13: I think `if(X86)` check should be added. OpenCV supports ARM, PowerPC, etc.

philippefoubert: Perhaps. Do you have the opportunity to check the ximea folder tree on an ARM board for example? I have only been able to test it on X86 boards using Windows or Linux...
However i have not changed the way it was previously done and if the architecture is not x86, the folder will not be found and it should not be a problem.

asmorkalov: Also I propose to use CMake list instead of lib_dirXXX. It's native interface for `FIND_LIBRARY` and easy to extend in future.

asmorkalov:
Looks good to me :+1: Please squash commits and we can merge then.

Issue #17546, opened 2020-06-14
[imgproc] Compilation error caused by color_yuv.simd.hpp


philippefoubert:
##### System information (version)
- OpenCV => 4.4.0
- Operating System / Platform => Windows 64 Bit
- Compiler => mingw32

##### Detailed description

The symbol "scr1" is already defined in the mingw-w64 runtime package in "mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\x86_64-w64-mingw32\include\dlgs.h" (included through "windows.h" for example).

##### Steps to reproduce

The compilation fails.

##### Several ways to solve this issue

1. The ugliest one would be to add at the beginning of this file:
```
#if defined scr1
#  undef scr1
#endif  // scr1
```

2.  A better one should be to avoid including the "windows.h" file before "color_yuv.simd.hpp"

3. An alternative could be the renaming of variables.

The guilty mingw32 "dlgs.h" file:
[dlgs_h.txt](https://github.com/opencv/opencv/files/4776326/dlgs_h.txt)

alalek:
Thank you for report!

Option 1 with workaround looks good to me.
Option 3 is acceptable too (these variables are local): `scr1` => `v_scr1`.

---

Included headers can be captured by adding `-H` option into the failed command (`make VERBOSE=1` should show failed command)

---

relates https://github.com/opencv/opencv_contrib/issues/2568

philippefoubert:
cf. PR #17577

PR #17577, opened 2020-06-18
imgproc(color_yuv.simd): Fix the build using MinGW

philippefoubert:
The symbol "scr1" is already defined in the mingw-w64 runtime package in "mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\x86_64-w64-mingw32\include\dlgs.h" (included through "windows.h" for example).
=> renaming of variables to avoid this conflict: `scr1 => v_scr1`

resolves #17546

PR #17579, opened 2020-06-18
fix Windows buils, color_yuv.simd.

Bleach665:
fix Windows build, color_yuv.simd.

resolve
https://github.com/opencv/opencv/issues/17548;
also maybe https://github.com/opencv/opencv/issues/17546


alalek:
Fixed here: #17577

PR #17581, opened 2020-06-18
CUDA: fix build error on Jetson TX1 and TX2

tomoaki0705:
relates #17556 

 * After #17556 , the build on Jetson TX1 and TX2 fails
 * enable_if_t and is_base_of is c++14 feature
 * override is c++11 feature

```
opencv/modules/cudaimgproc/src/cuda/canny.cu(110): warning: "override" and "final" are C++11 features
opencv/modules/cudaimgproc/src/cuda/canny.cu(131): error: namespace "std" has no member "enable_if_t"
```

### DIFF
--- a/modules/cudaimgproc/src/cuda/canny.cu
+++ b/modules/cudaimgproc/src/cuda/canny.cu
@@ -107,7 +107,7 @@ namespace canny
     {
         __host__ SrcTexRef(int _xoff, int _yoff) : SrcTex(_xoff, _yoff) {}
 
-        __device__ __forceinline__ int operator ()(int y, int x) const override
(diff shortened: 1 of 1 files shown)


nglee:
I think latest JetPack for TX1 and TX2 supports Ubuntu 18.04, which, in turn, is bundled with g++ 7.4.0.

To support older versions of g++, where c++14 standard is not a default option, can't we just add relevant flag for nvcc, or g++?

asmorkalov:
@nglee OpenCV team  maintains 3.4 as C++03 compatible branch. C++11 is default for master.

nglee:
@asmorkalov OK, I get it. Thank you.

PR #17584, opened 2020-06-18
Enable state initialization params via compile_args

AsyaPronina:
<cut/>

```
**WIP**
build_image:Linux AVX2=ubuntu-clang:18.04

Xbuild_image:Custom=centos:7
Xbuildworker:Custom=linux-1
Xbuild_gapi_standalone:Custom=ade-0.1.1f

build_image:Custom=ubuntu-openvino-2020.3.0:16.04
build_image:Custom Win=openvino-2020.3.0
build_image:Custom Mac=openvino-2020.3.0

test_modules:Custom=gapi
test_modules:Custom Win=gapi
test_modules:Custom Mac=gapi

buildworker:Custom=linux-1
test_opencl:Custom=OFF
test_bigdata:Custom=1
test_filter:Custom=*

Xforce_builders=Custom
Xbuildworker:Custom=linux-1
Xbuild_image:Custom=plaidml2
Xtest_modules:Custom=gapi
Xtest_filter:Custom=*ML*
```

### DIFF
--- a/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
+++ b/modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
@@ -2,7 +2,7 @@
 // It is subject to the license terms in the LICENSE file found in the top-level directory
 // of this distribution and at http://opencv.org/license.html.
 //
-// Copyright (C) 2018-2019 Intel Corporation
+// Copyright (C) 2018-2020 Intel Corporation
 
 
 #ifndef OPENCV_GAPI_GCPUKERNEL_HPP
@@ -137,7 +137,8 @@ public:
     // This function is a kernel's execution entry point (does the processing work)
     using RunF = std::function<void(GCPUContext &)>;
     // This function is a stateful kernel's setup routine (configures state)
-    using SetupF = std::function<void(const GMetaArgs &, const GArgs &, GArg &)>;
+    using SetupF = std::function<void(const GMetaArgs &, const GArgs &,
+                                      GArg &, const GCompileArgs &)>;
 
     GCPUKernel();
     GCPUKernel(const RunF& runF, const SetupF& setupF = nullptr);
@@ -284,21 +285,56 @@ struct OCVSetupHelper;
 template<typename Impl, typename... Ins>
 struct OCVSetupHelper<Impl, std::tuple<Ins...>>
 {
+    // Using 'auto' return type and 'decltype' specifier in both 'setup_impl' versions
+    // to check existence of required 'Impl::setup' functions.
+    // While 'decltype' specifier accepts expression we pass expression with 'comma-operator'
+    // where first operand of comma-operator is call attempt to desired 'Impl::setup' and
+    // the second operand is 'void()' expression.
+    //
+    // SFINAE for 'Impl::setup' which accepts compile arguments.
     template<int... IIs>
-    static void setup_impl(const GMetaArgs &metaArgs, const GArgs &args, GArg &state,
-                           detail::Seq<IIs...>)
+    static auto setup_impl(const GMetaArgs &metaArgs, const GArgs &args,
+                           GArg &state, const GCompileArgs &compileArgs,
+                           detail::Seq<IIs...>) ->
+        decltype(Impl::setup(detail::get_in_meta<Ins>(metaArgs, args, IIs)...,
+                             std::declval<typename std::add_lvalue_reference<
+                                              std::shared_ptr<typename Impl::State>
+                                                                            >::type
+                                         >(),
+                            compileArgs)
+                 , void())
     {
         // TODO: unique_ptr <-> shared_ptr conversion ?
         // To check: Conversion is possible only if the state which should be passed to
         // 'setup' user callback isn't required to have previous value
         std::shared_ptr<typename Impl::State> stPtr;
+        Impl::setup(detail::get_in_meta<Ins>(metaArgs, args, IIs)..., stPtr, compileArgs);
+        state = GArg(stPtr);
+    }
+
+    // SFINAE for 'Impl::setup' which doesn't accept compile arguments.
+    template<int... IIs>
+    static auto setup_impl(const GMetaArgs &metaArgs, const GArgs &args,
+                           GArg &state, const GCompileArgs &/* compileArgs */,
+                           detail::Seq<IIs...>) ->
+        decltype(Impl::setup(detail::get_in_meta<Ins>(metaArgs, args, IIs)...,
+                             std::declval<typename std::add_lvalue_reference<
+                                              std::shared_ptr<typename Impl::State>
+                                                                            >::type
+                                         >()
+                            )
+                 , void())
+    {
+        // The same comment as in 'setup' above.
+        std::shared_ptr<typename Impl::State> stPtr;
         Impl::setup(detail::get_in_meta<Ins>(metaArgs, args, IIs)..., stPtr);
         state = GArg(stPtr);
     }
 
-    static void setup(const GMetaArgs &metaArgs, const GArgs &args, GArg& state)
+    static void setup(const GMetaArgs &metaArgs, const GArgs &args,
+                      GArg& state, const GCompileArgs &compileArgs)
     {
-        setup_impl(metaArgs, args, state,
+        setup_impl(metaArgs, args, state, compileArgs,
                    typename detail::MkSeq<sizeof...(Ins)>::type());
     }
 };
@@ -435,8 +471,8 @@ public:
 
 // TODO: Reuse Anatoliy's logic for support of types with commas in macro.
 //       Retrieve the common part from Anatoliy's logic to the separate place.
-#define GAPI_OCV_KERNEL_ST(Name, API, State)                  \
-    struct Name:public cv::GCPUStKernelImpl<Name, API, State> \
+#define GAPI_OCV_KERNEL_ST(Name, API, State)                   \
+    struct Name: public cv::GCPUStKernelImpl<Name, API, State> \
 
 
 class gapi::cpu::GOCVFunctor : public gapi::GFunctor
--- a/modules/gapi/include/opencv2/gapi/gcommon.hpp
+++ b/modules/gapi/include/opencv2/gapi/gcommon.hpp
@@ -2,7 +2,7 @@
 // It is subject to the license terms in the LICENSE file found in the top-level directory
 // of this distribution and at http://opencv.org/license.html.
 //
-// Copyright (C) 2018 Intel Corporation
+// Copyright (C) 2018-2020 Intel Corporation
 
 
 #ifndef OPENCV_GAPI_GCOMMON_HPP
@@ -15,6 +15,7 @@
 #include <opencv2/gapi/opencv_includes.hpp>
 
 #include <opencv2/gapi/util/any.hpp>
+#include <opencv2/gapi/util/optional.hpp>
 #include <opencv2/gapi/own/exports.hpp>
 #include <opencv2/gapi/own/assert.hpp>
 
@@ -31,7 +32,7 @@ namespace detail
     {};
     struct TransformTag
     {};
-}
+} // namespace detail
 
 // This definition is here because it is reused by both public(?) and internal
 // modules. Keeping it here wouldn't expose public details (e.g., API-level)
@@ -52,7 +53,8 @@ struct GCompileArg;
 namespace detail {
     template<typename T>
     using is_compile_arg = std::is_same<GCompileArg, typename std::decay<T>::type>;
-}
+} // namespace detail
+
 // CompileArg is an unified interface over backend-specific compilation
 // information
 // FIXME: Move to a separate file?
@@ -121,13 +123,33 @@ private:
 using GCompileArgs = std::vector<GCompileArg>;
 
 /**
- * Wraps a list of arguments (a parameter pack) into a vector of
- * compilation arguments (cv::GCompileArg).
+ * @brief Wraps a list of arguments (a parameter pack) into a vector of
+ *        compilation arguments (cv::GCompileArg).
  */
 template<typename... Ts> GCompileArgs compile_args(Ts&&... args)
 {
     return GCompileArgs{ GCompileArg(args)... };
 }
+
+/**
+ * @brief Retrieves particular compilation argument by its type from
+ *        cv::GCompileArgs
+ */
+namespace gapi
+{
+template<typename T>
+inline cv::util::optional<T> getCompileArg(const cv::GCompileArgs &args)
+{
+    for (auto &compile_arg : args)
+    {
+        if (compile_arg.tag == cv::detail::CompileArgTag<T>::tag())
+        {
+            return cv::util::optional<T>(compile_arg.get<T>());
+        }
+    }
+    return cv::util::optional<T>();
+}
+} // namespace gapi
 
 /**
  * @brief Ask G-API to dump compiled graph in Graphviz format under
--- a/modules/gapi/src/backends/cpu/gcpubackend.hpp
+++ b/modules/gapi/src/backends/cpu/gcpubackend.hpp
@@ -2,7 +2,7 @@
 // It is subject to the license terms in the LICENSE file found in the top-level directory
 // of this distribution and at http://opencv.org/license.html.
 //
-// Copyright (C) 2018 Intel Corporation
+// Copyright (C) 2018-2020 Intel Corporation
 
 
 #ifndef OPENCV_GAPI_GCPUBACKEND_HPP
@@ -33,6 +33,7 @@ class GCPUExecutable final: public GIslandExecutable
 {
     const ade::Graph &m_g;
     GModel::ConstGraph m_gm;
+    const cv::GCompileArgs m_compileArgs;
 
     struct OperationInfo
     {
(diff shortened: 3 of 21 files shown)


modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -137,7 +137,8 @@ class GAPI_EXPORTS GCPUKernel
     // This function is a kernel's execution entry point (does the processing work)
     using RunF = std::function<void(GCPUContext &)>;
     // This function is a stateful kernel's setup routine (configures state)
-    using SetupF = std::function<void(const GMetaArgs &, const GArgs &, GArg &)>;
+    using SetupF = std::function<void(const GMetaArgs &, const GArgs &,
+                                      const GCompileArgs &, GArg &)>;
```
dmatveev, line 141: So now you should understand the point of Context object. :)

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -285,20 +286,22 @@ template<typename Impl, typename... Ins>
...
+    static void setup_impl(const GMetaArgs &metaArgs, const GArgs &args,
+                           const GCompileArgs &compileArgs, GArg &state,
                            detail::Seq<IIs...>)
     {
         // TODO: unique_ptr <-> shared_ptr conversion ?
         // To check: Conversion is possible only if the state which should be passed to
         // 'setup' user callback isn't required to have previous value
         std::shared_ptr<typename Impl::State> stPtr;
-        Impl::setup(detail::get_in_meta<Ins>(metaArgs, args, IIs)..., stPtr);
+        Impl::setup(detail::get_in_meta<Ins>(metaArgs, args, IIs)..., compileArgs, stPtr);
```
dmatveev, line 297: I'd put compile arguments after the state pointer.

ALSO:

I'd allow to let people write both versions of setup -- with compile args or not.
The extended version (with compile args) is the callback triggered by runtime.
The reduced (current master) version can be called from that default implementation.

Properly implementing this will require SFINAE checks which you can implement as a separate PR.
So far we could go with a less correct implementation where one callback just calls another by default.

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
```diff
@@ -285,20 +286,22 @@ template<typename Impl, typename... Ins>
...
         state = GArg(stPtr);
     }
 
-    static void setup(const GMetaArgs &metaArgs, const GArgs &args, GArg& state)
+    static void setup(const GMetaArgs &metaArgs, const GArgs &args,
+                      const GCompileArgs &compileArgs, GArg& state)
     {
-        setup_impl(metaArgs, args, state,
+        setup_impl(metaArgs, args, compileArgs, state,
```
dmatveev, line 304: ...with Context, we could leave all these methods unmodified.

dmatveev: Hopefully we'll not extend it anymore.

modules/gapi/include/opencv2/gapi/gcommon.hpp
```diff
@@ -121,14 +122,31 @@ struct GAPI_EXPORTS GCompileArg
...
+{
+    for (auto &compile_arg : args)
+    {
+        if (compile_arg.tag == cv::detail::CompileArgTag<T>::tag())
+        {
+            return cv::util::optional<T>(compile_arg.get<T>());
+        }
+    }
+    return cv::util::optional<T>();
+}
```
dmatveev, line 148: Once this function becomes public, we'll need to put more tests on it

modules/gapi/src/backends/cpu/gcpubackend.hpp
```diff
@@ -33,6 +33,7 @@ class GCPUExecutable final: public GIslandExecutable
 {
     const ade::Graph &m_g;
     GModel::ConstGraph m_gm;
+    cv::GCompileArgs m_compileArgs;
```
dmatveev, line 36: I think it can be `const` as well.

Having `const` is a good way to let people reading this code know that this variable won't change once initialized - so will not cause any side effects going further.

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -86,15 +86,15 @@ namespace
                     return gim.metadata(nh).get<NodeKind>().k == NodeKind::ISLAND;
                 });
 
-            const auto out_rois = cv::gimpl::getCompileArg<cv::GFluidOutputRois>(args);
+            const auto out_rois = cv::getCompileArg<cv::GFluidOutputRois>(args);
```
dmatveev, line 89: looking at this, maybe this `getCompileArg<>` thing should be under `cv::gapi`

modules/gapi/src/compiler/gcompiler.cpp
```diff
@@ -452,7 +452,8 @@ cv::GStreamingCompiled cv::gimpl::GCompiler::produceStreamingCompiled(GPtr &&pg)
         outMetas = GModel::ConstGraph(*pg).metadata().get<OutputMeta>().outMeta;
     }
 
-    std::unique_ptr<GStreamingExecutor> pE(new GStreamingExecutor(std::move(pg)));
+    std::unique_ptr<GStreamingExecutor> pE(new GStreamingExecutor(std::move(pg),
+                                                                  m_args));
```
dmatveev, line 456: So why do we pass `m_args` to the Streaming Executor but don't pass those to the `GExecutor` in the `produceCompiled`?

Why do we need to pass *compile arguments* to *executor* at all?

modules/gapi/src/executor/gstreamingexecutor.cpp
```diff
@@ -603,10 +603,12 @@ void collectorThread(std::vector<Q*> in_queues,
 }
 } // anonymous namespace
 
-cv::gimpl::GStreamingExecutor::GStreamingExecutor(std::unique_ptr<ade::Graph> &&g_model)
+cv::gimpl::GStreamingExecutor::GStreamingExecutor(std::unique_ptr<ade::Graph> &&g_model,
+                                                  const GCompileArgs &comp_args)
     : m_orig_graph(std::move(g_model))
     , m_island_graph(GModel::Graph(*m_orig_graph).metadata()
                      .get<IslandModel>().model)
+    , m_comp_args(comp_args)
```
dmatveev, line 611: this is not used, right?

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
```diff
@@ -8,6 +8,30 @@
...
 #include <opencv2/gapi/streaming/cap.hpp>
 
+#include <opencv2/core.hpp>
+#include <opencv2/core/cvstd.hpp>
+#include <opencv2/video.hpp>
+
+
+namespace opencv_test
+{
+    struct SubstractBackgroundStateParams
```
dmatveev, line 18: `SubtractBackground` is uncommon name for the feature, usually it is `BgSub` or something like that.

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
```diff
@@ -37,7 +61,8 @@ namespace
 
     GAPI_OCV_KERNEL_ST(GOCVCountCalls, GCountCalls, int)
     {
-        static void setup(const cv::GMatDesc &/* in */, std::shared_ptr<int> &state)
+        static void setup(const cv::GMatDesc &/* in */, const cv::GCompileArgs &/* compileArgs */,
+                          std::shared_ptr<int> &state)
```
dmatveev, line 65: Maybe having compile args before state is good though

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
```diff
@@ -85,6 +112,36 @@ namespace
...
+    GAPI_OCV_KERNEL_ST(GOCVSubstractBackground, GSubstractBackground, cv::BackgroundSubtractor)
+    {
+        static void setup(const cv::GMatDesc &/* desc */,
+                          const cv::GCompileArgs &compileArgs,
+                          std::shared_ptr<BackgroundSubtractor> &state)
+        {
+            auto sbParams = getCompileArg<SubstractBackgroundStateParams>(compileArgs)
+                                .value_or(SubstractBackgroundStateParams { });
+
+            Ptr<BackgroundSubtractor> model;
```
dmatveev, line 130: Please don't use `cv::Ptr`

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
```diff
@@ -85,6 +112,36 @@ namespace
...
+            if (sbParams.method == "knn")
+                model = createBackgroundSubtractorKNN();
+            else if (sbParams.method == "mog2")
+                model = createBackgroundSubtractorMOG2();
```
dmatveev, line 134: ...else?

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
```diff
@@ -85,6 +112,36 @@ namespace
...
+                          const cv::GCompileArgs &compileArgs,
+                          std::shared_ptr<BackgroundSubtractor> &state)
+        {
+            auto sbParams = getCompileArg<SubstractBackgroundStateParams>(compileArgs)
+                                .value_or(SubstractBackgroundStateParams { });
+
+            Ptr<BackgroundSubtractor> model;
+            if (sbParams.method == "knn")
+                model = createBackgroundSubtractorKNN();
+            else if (sbParams.method == "mog2")
```
dmatveev, line 133: I know it is here for testing purposes only, but in this particular case, it could be an enum defined directly as operation argument at `G_API_OP` level.

We don't put the resize interpolation type as a `compile_arg` :)

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
```diff
@@ -167,7 +224,114 @@ TEST(StatefulKernel, InvalidReallocatingKernel)
...
+{
+    void compareBackSubResults(const cv::Mat &actual, const cv::Mat &expected,
+                                       const int diffPercent)
+    {
+        GAPI_Assert(actual.size() == expected.size());
+        int allowedNumDiffPixels = actual.size().area() * diffPercent / 100;
+
+        cv::Mat diff;
+        cv::absdiff(actual, expected, diff);
+        #undef countNonZero
```
dmatveev, line 239: This `#undef` has an effect for the rest of the file, unfortunately.

@alalek if we really need countNonZero, what should we do?

@AsyaPronina instead you can calculate a histogram over absdiff.

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
```diff
@@ -167,7 +224,114 @@ TEST(StatefulKernel, InvalidReallocatingKernel)
...
+    }
+} // anonymous namespace
+
+TEST(StatefulKernel, StateIsInitViaCompArgs)
+{
+    cv::Mat frame(1080, 1920, CV_8UC3),
+            gapiForeground,
+            ocvForeground;
+
+    cv::randu(frame, cv::Scalar(0, 0, 0), cv::Scalar(255, 255, 255));
```
dmatveev, line 252: Y not play a file instead? `opencv_extra` has some.

dmatveev: This, again, can be done separately if we decide to have Background Subtractor at the API level.

...And there is already a demand in it. It may be a next task to do

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
```diff
@@ -167,7 +224,114 @@ TEST(StatefulKernel, InvalidReallocatingKernel)
...
+
+    // G-API code
+    cv::GMat in;
+    cv::GMat out = GSubstractBackground::on(in);
+    cv::GComputation c(cv::GIn(in), cv::GOut(out));
+
+    const auto pkg = cv::gapi::kernels<GOCVSubstractBackground>();
+
+    auto gapiBackSub = c.compile(cv::descr_of(frame),
+                                 cv::compile_args(pkg, SubstractBackgroundStateParams { "knn" }));
```
dmatveev, line 262: Really, in this case

```c++
cv::GMat out = cv::gapi::backgroundSubtract(in, cv::gapi::BG::KNN);
```

is better

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
dmatveev: Ok I see it below. Great!

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: Yes, got you!

modules/gapi/src/compiler/gcompiler.cpp
AsyaPronina: Discussed in IM, comment about compile arguments added before GStreamingExecutor constructor definition

modules/gapi/src/executor/gstreamingexecutor.cpp
AsyaPronina: Discussed in IM

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
AsyaPronina: Yes, agree with you!

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: Fixed, please check if such implementation you can accept current implementation

modules/gapi/include/opencv2/gapi/cpu/gcpukernel.hpp
AsyaPronina: Got you

modules/gapi/include/opencv2/gapi/gcommon.hpp
AsyaPronina: Added, please check

modules/gapi/src/backends/cpu/gcpubackend.hpp
AsyaPronina: Thanks! Fixed.

modules/gapi/src/backends/fluid/gfluidbackend.cpp
AsyaPronina: Fixed

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
AsyaPronina: Renamed to `BackSub` to comply with used names for substractor in the test

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
AsyaPronina: Fixed

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
AsyaPronina: GAPI_Assert should fail in other cases

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
AsyaPronina: Calculated via histogram

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
AsyaPronina: Yes

modules/gapi/include/opencv2/gapi/gcommon.hpp
dmatveev: Thanks

modules/gapi/src/backends/fluid/gfluidbackend.cpp
```diff
@@ -86,15 +86,15 @@ namespace
                     return gim.metadata(nh).get<NodeKind>().k == NodeKind::ISLAND;
                 });
 
-            const auto out_rois = cv::gimpl::getCompileArg<cv::GFluidOutputRois>(args);
```
dmatveev, line 89: So the old name is not available anymore?

I'd keep the old function `cv::gimpl`, it could be implemented via the new `cv::gapi` one.

If the `cv::gimpl` has gone, we have all our private backends broken now.

dmatveev: I see this issue fixed, thanks

dmatveev:
Please remember about our PR guidelines, at least Standalone needs to be tested.

https://github.com/opencv/opencv/wiki/Graph-API

AsyaPronina:
> Copyright years are usually bumped as a separate activity, but anyway thanks for that.
> 
> 👍



> Please remember about our PR guidelines, at least Standalone needs to be tested.
> 
> https://github.com/opencv/opencv/wiki/Graph-API

Hi Dmitry, thanks a lot!
Added build for standalone G-API mode + PlaidML, as gplaidmlbackend.cpp file was affected.

modules/gapi/test/cpu/gapi_ocv_stateful_kernel_tests.cpp
```diff
@@ -85,6 +112,34 @@ namespace
...
+         static GMatDesc outMeta(GMatDesc in) { return in.withType(CV_8U, 1); }
+    };
+
+    GAPI_OCV_KERNEL_ST(GOCVBackSub, GBackSub, cv::BackgroundSubtractor)
+    {
+        static void setup(const cv::GMatDesc &/* desc */,
+                          std::shared_ptr<BackgroundSubtractor> &state,
+                          const cv::GCompileArgs &compileArgs)
+        {
+            auto sbParams = cv::gapi::getCompileArg<BackSubStateParams>(compileArgs)
```
TolyaTalamanov, line 127: Cool 👍

PR #17607, opened 2020-06-21
Fix: error in the dimension used for computeMinMax

pemmanuelviel:
Instead of using the current dimension for which we just got a big span,
we were computing Min and Max for the previous dimension stored in cutfeat
(and using 0 instead of the dimension indice for the very first dimension
with "span > (1-eps)max_span")

### DIFF
--- a/modules/flann/include/opencv2/flann/kdtree_single_index.h
+++ b/modules/flann/include/opencv2/flann/kdtree_single_index.h
@@ -461,7 +461,7 @@ private:
             DistanceType span = bbox[i].high-bbox[i].low;
             if (span>(DistanceType)((1-EPS)*max_span)) {
                 ElementType min_elem, max_elem;
-                computeMinMax(ind, count, cutfeat, min_elem, max_elem);
+                computeMinMax(ind, count, (int)i, min_elem, max_elem);
                 DistanceType spread = (DistanceType)(max_elem-min_elem);
                 if (spread>max_spread) {
                     cutfeat = (int)i;


asmorkalov:
@pemmanuelviel Thanks for the patch. Could you provide small test that proofs your fix?

asmorkalov:
@pemmanuelviel Also it looks like the same issue is present in 3.4 branch too. This patch should [go](https://github.com/opencv/opencv/wiki/How_to_contribute#faq) into [3.4 branch](https://github.com/opencv/opencv/wiki/Branches) first. We will merge changes from 3.4 into master regularly (weekly/bi-weekly).

So, please:

Note: no need to re-open PR, apply changes "inplace".

vpisarev:
@pemmanuelviel, thank you, looks good to me! :+1:
I think, since it's the change in a single file, maybe it would be enough just to change the target branch from master to 3.4 right in your pull request without any git commands

pemmanuelviel:
Hi @vpisarev ,
Just changing the target branch is not enough as merging will not work.
I will backport this change and some others in 3.4 following @asmorkalov advices after I've finished fixing the issues spotted by the buildfarm on my different submitted branches.

pemmanuelviel:
@asmorkalov To ensure each commit is fine on 3.4 I've created new branches (replacing pev-xxx by pev--xxx) and cherry-picked each commit. These new branches are currently in pull-request.

pemmanuelviel:
BTW, how can I associate labels to my PR? Can't find where in the gui...

asmorkalov:
@pemmanuelviel Only maintainers add labels to PRs.

PR #17617, opened 2020-06-22
core(logger): complete initialization of logger structures

alalek:
- for using of logging functions from global destructors

relates #13909


alalek:
:+1:

PR #17618, opened 2020-06-22
Added/Fixed testcases for SIFT

Yosshi999:
* The testset in perf_feature2d.cpp was a superset of perf_sift.cpp, so I merged it.
* Added an exact regression test for a bit-exact version which will be implemented in the near future. The testdata is: https://github.com/opencv/opencv_extra/pull/768
* Resolved two TODOs in scale invariance tests.


modules/features2d/test/test_sift.cpp
```diff
@@ -0,0 +1,107 @@
...
+{
+public:
+    CV_SiftTest();
+    ~CV_SiftTest();
+protected:
+    void run(int);
+};
+
+CV_SiftTest::CV_SiftTest() {}
+CV_SiftTest::~CV_SiftTest() {}
```
alalek, line 19: Avoid using of legacy test classes.
Use GoogleTest directly.

modules/features2d/test/test_sift.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+void CV_SiftTest::run( int )
+{
+    Mat image = imread(string(ts->get_data_path()) + "features2d/tsukuba.png");
+    string xml = string(ts->get_data_path()) + "sift/result.xml.gz";
```
alalek, line 24: Consider using of `cvtest::findDataFile()` for searching testdata.

modules/features2d/test/test_sift.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    if (image.empty())
+    {
+        ts->printf( cvtest::TS::LOG, "No image.\n" );
+        ts->set_failed_test_info( cvtest::TS::FAIL_INVALID_TEST_DATA );
+        return;
+    }
```
alalek, line 31: => `ASSERT_FALSE(image.empty());`

modules/features2d/test/test_sift.cpp
```diff
@@ -0,0 +1,107 @@
...
+    cvtColor(image, gray, COLOR_BGR2GRAY);
+
+    vector<KeyPoint> calcKeypoints;
+    Mat calcDescriptors;
+    Ptr<SIFT> sift = cv::SIFT::create();
+    sift->detectAndCompute(gray, Mat(), calcKeypoints, calcDescriptors, false);
+
+    FileStorage fs(xml, FileStorage::READ);
+    if (!fs.isOpened())
+    {
```
alalek, line 43: Implicit data re-generation is dangerous.

```
#ifndef GENERATE_DATA
    ASSERT_TRUE(fs.isOpened()) << xml;
    ... read ...
#else
    ... creating XML ...
#endif // GENERATE_DATA
```

Commented `//#define GENERATE_DATA` in the beginning of the file (see test_houghlines.cpp as example)

modules/features2d/test/test_sift.cpp
```diff
@@ -0,0 +1,107 @@
...
+    fs.release();
+
+    // Compare the number of keypoints
+    if (validKeypoints.size() != calcKeypoints.size())
+    {
+        ts->printf( cvtest::TS::LOG, "Bad keypoints count (validCount = %d, calcCount = %d).\n",
+                    validKeypoints.size(), calcKeypoints.size() );
+        ts->set_failed_test_info(cvtest::TS::FAIL_MISMATCH);
+        return;
+    }
```
alalek, line 75: `ASSERT_EQ(expected, actual) << optional_message;`

terfendail:
Looks like the new test is too strict. Since the implementation at the moment use floating point values and therefore isn't bit-exact by nature there could be small differences depending on compiler optimization options and AVX2 availability.
I think that precision requirement could be relaxed a bit at the moment

Yosshi999:
The relaxed test already exists. (https://github.com/opencv/opencv/blob/3.4/modules/features2d/test/test_descriptors_regression.cpp#L345)

As we discussed in https://github.com/terfendail/opencv/issues/3#issuecomment-647376389 and https://github.com/terfendail/opencv/issues/3#issuecomment-647422882, I will withdraw 2d09bfb (with opencv/opencv_extra#768) and re-create a PR when a bit-exact version is implemented.

PR #17627, opened 2020-06-23
Pev fix trees descent

pemmanuelviel:


pemmanuelviel:
Branch addressing the descent that is done till the last tree even if criteria for not accepting anymore points are met in previous tree.
So now, either we save processing time by not descending the other trees when the condition is met, or if we decide however to descend all trees like before, setting explore_all_trees to true, we look for better points in the leaf .

pemmanuelviel:
@vpisarev Hi Vadim, as you were already the one reviewing my changes on FLANN in 2013 I write you directly. I have the comparator of ABI dumps complaining about the new parameter with default value I have added to the interface of SearchParams. 
SearchParams::SearchParams [C1] ( int checks, float eps, bool sorted )  1  --> SearchParams::SearchParams [C1] ( int checks, float eps, bool sorted, bool explore_all_trees )
As it makes sense to leave the choice to the user between similar quality at faster speed or better quality at same speed than previously, we need this new option.
I'm not familiar on how to address this issue. How do I need to proceed?
Thanks

PR #17636, opened 2020-06-23
support indexed color images on iOS

okamotoR:
UIImageToMat() fail to convert indexed color image now.
This is because CGBitmapContextCreate() in UIImageToMat() don't support indexed color spaces.
I added if block for indexed color spaces.

CGBitmapContextCreate
https://developer.apple.com/documentation/coregraphics/1455939-cgbitmapcontextcreate?language=objc

- [○] I agree to contribute to the project under OpenCV (BSD) License.
- [○] To the best of my knowledge, the proposed patch is not based on a code under GPL or other license that is incompatible with OpenCV
- [○] The PR is proposed to proper branch
- [○] The feature is well documented and sample code can be built with the project CMake

### DIFF
--- a/modules/imgcodecs/src/ios_conversions.mm
+++ b/modules/imgcodecs/src/ios_conversions.mm
@@ -112,6 +112,21 @@ void UIImageToMat(const UIImage* image,
                                            m.step[0], colorSpace,
                                            bitmapInfo);
     }
+    else if (CGColorSpaceGetModel(colorSpace) == kCGColorSpaceModelIndexed)
+    {
+        // CGBitmapContextCreate() does not support indexed color spaces.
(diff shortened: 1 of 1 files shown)


okamotoR:
Thank you for reviewing.

I created UIImage on swift from https://en.wikipedia.org/wiki/Indexed_color.
Then I used UIImagePickerController.

I copy/pasted from not kCGColorSpaceModelMonochrome block but else block 
because CGColorSpaceCreateDeviceRGB() make RGBColorSpace.

I think this function has no problem now for other color spaces.

komakai:
@okamotoR ありがとう！ 
I was able to test the code and can confirm it is working.

okamotoR:
Thank you very much.

PR #17642, opened 2020-06-23
Pev  fixes and clean

pemmanuelviel:

### DIFF
--- a/modules/flann/include/opencv2/flann/lsh_table.h
+++ b/modules/flann/include/opencv2/flann/lsh_table.h
@@ -245,7 +245,7 @@ public:
     {
         std::cerr << "LSH is not implemented for that type" << std::endl;
         assert(0);
-        return static_cast<size_t>(-1);
+        return 0;
     }
 
     /** Get statistics about the table
--- a/modules/flann/include/opencv2/flann/lsh_table.h
+++ b/modules/flann/include/opencv2/flann/lsh_table.h
@@ -245,7 +245,7 @@ public:
     {
         std::cerr << "LSH is not implemented for that type" << std::endl;
         assert(0);
-        return -1;
+        return static_cast<size_t>(-1);
     }
 
     /** Get statistics about the table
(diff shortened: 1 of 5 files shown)


pemmanuelviel:
Asides from some fixes I've made parsing the code, I've also ported some from the original FLANN repo from Marius Muja.

modules/flann/include/opencv2/flann/lsh_table.h
```diff
@@ -245,7 +245,7 @@ class LshTable
     {
         std::cerr << "LSH is not implemented for that type" << std::endl;
         assert(0);
-        return 1;
+        return static_cast<size_t>(-1);
```
dkurt, line 248: I'm not sure that this is reliable for 32bit OS - the values are different

pemmanuelviel: Errors return -1 and not 1 in FLANN. As the return type is an unsigned integer, value needed to be casted.
However, as an error value for a key, 0 should do the work to help the user to figure out what is going on.

modules/flann/include/opencv2/flann/lsh_table.h
```diff
@@ -245,7 +245,7 @@ class LshTable
     {
         std::cerr << "LSH is not implemented for that type" << std::endl;
         assert(0);
-        return 1;
+        return 0;
```
alalek, line 248: Please use `CV_Assert(0 && "LSH is not implemented for that type");` which triggers an exception (need to check for possible memory leaks in caller code or properly use RAII).

BTW, C-like `assert()` is not allowed and should be removed.

pemmanuelviel: Hi Alexander,
Thanks for your feedback. Well, this is not the only C-like stuff we should get rid of, like the old style casts. 
In a general way, I try to stick to the coding style of the code I'm bringing changes to in order not to be too intrusive. For FLANN it's a mater of where to put the cursor. 
So, in another branch, I'm going to parse FLANN for replacing all the assert() that are there, it can't harm. Just to let you know, just in core you already have more than a hundred of them.
For the context my intend was first to disclose quickly some work I did some years ago in order to bring also a new feature. Seeing several bugs while parsing the code I have however decided to fix them first, to the extent I can afford...

PR #17643, opened 2020-06-23
Pev  new flann demo

pemmanuelviel:
<cut/>


pemmanuelviel:
The only example currently available for FLANN demonstrates how to use the based matcher.
Using other modes requires some work to figure out at what level to access FLANN functionalities (Index_, GenericIndex, miniflann, etc.) and it appears on forums it's not straightforward for many.
This new sample addresses this problem.
It will also ease the work of potential contributors to FLANN by saving them the burden to write a program to test their modifications in the lib.

samples/cpp/tutorial_code/features2D/feature_flann_matcher/SURF_FLANN_search_dataset_Demo.cpp
```diff
@@ -0,0 +1,270 @@
+// SURF_FLANN_search_dataset_Demo.cpp
```
alalek, line 1: > tutorial_code

This code must belong to some tutorial (it should be mentioned by some tutorial).

pemmanuelviel: Moved to sample/cpp/

samples/cpp/flann_search_dataset.cpp
```diff
@@ -0,0 +1,264 @@
...
+    //-- Show detected matches
+    imshow("Good Matches", img_matches );
+
+    while(1)
+    {
+        int k = waitKey();
+        if (k == 27) // Esc key
+            break;
+    }
+    delete index;
```
alalek, line 262: > delete

Lets avoid using of manual memory management.

Please use `std::shared_ptr` approach instead (through `cv::Ptr` / `makePtr` emulation for C++98).

PR #17644, opened 2020-06-23
Pev  fix trees descent

pemmanuelviel:


pemmanuelviel:
Branch addressing the descent that is done till the last tree even if criteria for not accepting anymore points are met in previous tree.
So now, either we save processing time by not descending the other trees when the condition is met, or if we decide however to descend all trees like before, setting explore_all_trees to true, we look for better points in the leaf .

pemmanuelviel:
/!\ Having SearchParams changing from 3 to 4 arguments, even if the fourth one has default value, can be considered as an interface break on 3.4. Do we prefer to keep it only for master?

